PR #7709, opened 2022-04-24
prov/efa: refactor message protocol selection

wzamazon:
a series of patches that refactor the protocol selection of message

### DIFF
--- a/prov/efa/src/rxr/rxr_msg.c
+++ b/prov/efa/src/rxr/rxr_msg.c
@@ -55,157 +55,107 @@
  *  Send function
  */
 
+
 /**
- *   Utility functions used by both non-tagged and tagged send.
+ * @brief select a two-sided protocol for the send operation when send buffer is on cuda memory
+ *
+ * @param[in]		rxr_ep		endpoint
+ * @param[in]		tx_entry	contains information of the send operation
+ * @return		the RTM packet type of the two-sided protocol. Four
+ *                      types of protocol can be used: eager, medium, longcts, longread.
+ *                      Each protocol has tagged/non-tagged version. Some protocols has a DC version.
+ *
  */
 static inline
-ssize_t rxr_msg_post_cuda_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
+int rxr_msg_select_rtm_for_cuda(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
 {
-	int err, tagged;
-	struct rdm_peer *peer;
-	int pkt_type, max_eager_data_size;
+	int tagged;
+	int eager_rtm, medium_rtm, longcts_rtm, longread_rtm;
+	int eager_rtm_max_data_size;
 	bool delivery_complete_requested;
 
-	assert(RXR_EAGER_MSGRTM_PKT + 1 == RXR_EAGER_TAGRTM_PKT);
-	assert(RXR_LONGREAD_MSGRTM_PKT + 1 == RXR_LONGREAD_TAGRTM_PKT);
-	assert(RXR_DC_EAGER_MSGRTM_PKT + 1 == RXR_DC_EAGER_TAGRTM_PKT);
-
+	assert(tx_entry->op == ofi_op_tagged || tx_entry->op == ofi_op_msg);
 	tagged = (tx_entry->op == ofi_op_tagged);
 	assert(tagged == 0 || tagged == 1);
 
 	delivery_complete_requested = tx_entry->fi_flags & FI_DELIVERY_COMPLETE;
-	/*
-	 * Todo: use information in handshake packet to determine whether
-	 * the receiver supports gdrcopy.
-	 */
-	if (tx_entry->total_len == 0 || cuda_is_gdrcopy_enabled()) {
-		pkt_type = delivery_complete_requested ? RXR_DC_EAGER_MSGRTM_PKT : RXR_EAGER_MSGRTM_PKT;
-		max_eager_data_size = rxr_pkt_req_max_data_size(rxr_ep,
-								tx_entry->addr,
-								pkt_type + tagged,
-								tx_entry->fi_flags, 0);
 
-		max_eager_data_size = MIN(max_eager_data_size, rxr_env.efa_max_gdrcopy_msg_size);
+	eager_rtm = (delivery_complete_requested) ? RXR_DC_EAGER_MSGRTM_PKT + tagged
+						  : RXR_EAGER_MSGRTM_PKT + tagged;
 
-		if (tx_entry->total_len <= max_eager_data_size) {
-			return rxr_pkt_post_ctrl(rxr_ep, RXR_TX_ENTRY, tx_entry,
-						 pkt_type + tagged, 0, 0);
-		}
+	medium_rtm = (delivery_complete_requested) ? RXR_DC_MEDIUM_MSGRTM_PKT + tagged
+						   : RXR_MEDIUM_MSGRTM_PKT + tagged;
 
-		if (tx_entry->total_len <= rxr_env.efa_max_gdrcopy_msg_size) {
-			if (tx_entry->total_len <= rxr_env.efa_max_medium_msg_size) {
-				pkt_type = delivery_complete_requested ? RXR_DC_MEDIUM_MSGRTM_PKT : RXR_MEDIUM_MSGRTM_PKT;
-			} else {
-				pkt_type = delivery_complete_requested ? RXR_DC_LONGCTS_MSGRTM_PKT : RXR_LONGCTS_MSGRTM_PKT;
-			}
+	longcts_rtm = (delivery_complete_requested) ? RXR_DC_LONGCTS_MSGRTM_PKT + tagged
+						    : RXR_LONGCTS_MSGRTM_PKT + tagged;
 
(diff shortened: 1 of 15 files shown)


prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -527,6 +528,9 @@ ssize_t rxr_pkt_init_medium_msgrtm(struct rxr_ep *ep,
 	if (ret)
 		return ret;
```
shijin-aws, line 538: Could u refresh my memory why this mr registration is tried for medium msg transfer only?

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -325,77 +320,91 @@ ssize_t rxr_pkt_post_ctrl_once(struct rxr_ep *rxr_ep, int entry_type, void *x_en
...
+ *
+ * Some REQ packet types must be queued.
+ *
+ * @param[in]   rxr_ep          endpoint
+ * @param[in]   x_entry         pointer to rxr_op_entry. (either a tx_entry or an rx_entry)
+ * @param[in]   pkt_type        packet type.
+ * @param[in]   inject          send control packet via inject or not.
+ * @return      On success return 0, otherwise return a negative libfabric error code. Typical error:
+ */
+ssize_t rxr_pkt_post_req(struct rxr_ep *ep, struct rxr_op_entry *tx_entry, int req_type, bool inject, uint64_t flags)
```
shijin-aws, line 372: what are the typical errors?

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -55,157 +55,85 @@
  *  Send function
  */
 
-/**
- *   Utility functions used by both non-tagged and tagged send.
- */
 static inline
-ssize_t rxr_msg_post_cuda_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
+int rxr_msg_select_rtm_for_cuda(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
```
shijin-aws, line 59: Could you add a docstring for this function?

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -328,13 +287,19 @@ ssize_t rxr_msg_generic_send(struct fid_ep *ep, const struct fi_msg *msg,
...
 	tx_entry = rxr_ep_alloc_tx_entry(rxr_ep, msg, op, tag, flags);
-
 	if (OFI_UNLIKELY(!tx_entry)) {
 		err = -FI_EAGAIN;
 		rxr_ep_progress_internal(rxr_ep);
 		goto out;
 	}
 
+	efa_ep = container_of(rxr_ep->rdm_ep, struct efa_ep, util_ep.ep_fid);
+	ret = efa_ep_use_p2p(efa_ep, tx_entry->desc[0]);
```
shijin-aws, line 297: can you rename `ret` to `use_p2p`, similarly for the one used in `rxr_msg_post_rtm`.

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -94,6 +94,14 @@ struct rxr_req_inf REQ_INF_LIST[] = {
 	[RXR_COMPARE_RTA_PKT] = {4, sizeof(struct rxr_rta_hdr), 0},
 };
 
+bool rxr_pkt_req_supported_by_peer(int req_type, struct rdm_peer *peer)
+{
+	assert(peer->flags & RXR_PEER_HANDSHAKE_RECEIVED);
+
+	int extra_id = REQ_INF_LIST[req_type].protover - 4;
```
shijin-aws, line 101: Not quite follow this line. Could you explain how `extra_id` is calculated?

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -55,157 +55,85 @@
...
+		return longcts_rtm;
 	}
 
-	return rxr_pkt_post_ctrl(rxr_ep, RXR_TX_ENTRY, tx_entry,
-				 RXR_LONGREAD_MSGRTM_PKT + tagged, 0, 0);
+	return longread_rtm;
 }
 
-ssize_t rxr_msg_post_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
+int rxr_msg_select_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry, int use_p2p)
```
shijin-aws, line 104: Could u add a docstring for this function?

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -216,95 +144,126 @@ ssize_t rxr_msg_post_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
...
-		 * It is possible for the remote endpoint to support RDMA read,
-		 * but not p2p transfers between efa and neuron. That scenario
-		 * will cause a fatal error; if we want to catch this we will
-		 * need to extend the handshake packet to report device p2p
-		 * support.
-		 */
-		ret = rxr_ep_determine_rdma_support(rxr_ep, tx_entry->addr, peer);
-		if (ret < 0)
-			return ret;
+	if (efa_mr_is_neuron(tx_entry->desc[0]))
```
shijin-aws, line 185: Not related to this refactor but i am wondering if we should add a `assert(use_p2p)` for neuron.

prov/efa/src/rxr/rxr_pkt_type_req.c
wzamazon: It is not only for medium, but also for long-cts. These two protocols use this function because they can use memory registration opportunistically. e.g. they can use MR but does not rely on MR. Other protocols is not like that. Eager does not use MR at all. Long read rely on MR. Will update the comment to make it more clear.

wzamazon:
bot:aws:retest

prov/efa/src/rxr/rxr_pkt_type_req.c
wzamazon: Updated comment in new revision.

prov/efa/src/rxr/rxr_pkt_cmd.c
wzamazon: Removed this not. There is no typical error for this function.

prov/efa/src/rxr/rxr_msg.c
wzamazon: Added in new revision.

prov/efa/src/rxr/rxr_msg.c
wzamazon: Fixed in new revision. though I think we need to discussion whether check `use_p2p` on send path is necessary.

prov/efa/src/rxr/rxr_msg.c
wzamazon: Added in new revision.

prov/efa/src/rxr/rxr_msg.c
wzamazon: Yes, though I think we need to discuss whether it is necessary to check `use_p2p` on send path.

prov/efa/src/rxr/rxr_msg.c
shijin-aws: Thx, the current implementation looks good to me.

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -45,61 +45,52 @@
...
- *
- * Baseline features will always be version 4 features, baseline and
- * not have a ex_feature_flag.
- *
- * Each extra feature will be assign a version and an ex_feature_flag.
- * Each extra feature will correspond to 1 or more REQ packet types.
- */
 static const
 struct rxr_req_inf REQ_INF_LIST[] = {
 	/* rtm header */
```
shijin-aws, line 55: is `extra_info_id` always 0?

wzamazon: No.

The No. 65th's extra feature's extra_info_id will be 1.
Similiarly, the No. 129th extra feature's extra_info_id will be 2.

shijin-aws: I see, so `extra_info_id` is the offset of the 64 bit feature flag in the `extra_info` array. When we have more than 64 features we need to use the second element in `extra_info` to check it, and so on.

shijin-aws (approved):

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: But in that case what will be the 65th feature bit look like? `BIT_ULL(65)` or `BIT_ULL(1)`?

wzamazon: 65th feature will has `extra_info_id` being 1, and flag being `BIT_ULL(0)`.

`BIT_ULL(65)` does not exist because the flag is 64 bits

shijin-aws: Thanks, that makes sense.

PR #7714, opened 2022-04-28
prov/rxm: Fix possible use of event data after free

shefty:
From a valgrind report:

==3590290== at 0xCA3AD30: rxm_process_connect (rxm_conn.c:508)
ERROR: daos_engine:0 ==3590290== by 0xCA3C054: rxm_handle_event (rxm_conn.c:816)
==3590290== by 0xCA3C614: rxm_cm_progress (rxm_conn.c:915)
==3590290== by 0x6031179: start_thread (in /usr/lib64/libpthread-2.28.so)
==3590290== by 0x6E55DC2: clone (in /usr/lib64/libc-2.28.so)
==3590290== Address 0x38d04c6c8 is 8 bytes inside a block of size 18,864 free'd
==3590290== at 0x4C38A03: free (vg_replace_malloc.c:755)
==3590290== by 0xCBDE02C: tcpx_ep_close (tcpx_ep.c:624)
==3590290== by 0xCA3845B: fi_close (fabric.h:612)
==3590290== by 0xCA39538: rxm_close_conn (rxm_conn.c:88)
==3590290== by 0xCA3B08A: rxm_process_reject (rxm_conn.c:566)
==3590290== by 0xCA3BF95: rxm_handle_error (rxm_conn.c:798)
==3590290== by 0xCA3C193: rxm_conn_progress (rxm_conn.c:841)
==3590290== by 0xCA3ACAC: rxm_get_conn (rxm_conn.c:476)

In rxm_cm_progress(), the cm thread can retrieve an event for
a msg ep.  However, another thread could be trying to send over
that ep.  If the ep is not connected, that thread will try to
progress the ep (rxm_conn_progress in the above trace).  That
thread could find another even for the same msg ep.  In the
processing of that event, the msg ep may be freed.

At this point, the cm thread starts processing the event that
it read, which can result in it accessing a freed ep or rxm_conn.
Valgrind properly caught this during a DAOS stress test where
connections were suddenly aborted at the peer.

To avoid this, we need to ensure that event processing is serialize
with respect to closing the ep.  To handle this, we change the
rxm_cm_progress to only act on event data that it reads while
holding the rdm ep lock.  We change the fi_eq_sread() call to use
the FI_PEEK flag.  That allows the call to block and return only
once we have an event to process.  We re-read the eq after acquiring
the ep lock.  This avoids races with another thread processing events.
The msg ep is closed holding the ep lock, which will remove any
queued events on the eq.

An alternate (more efficient) solution is to update the fi_eq_sread() calls to handle a null buf parameter, similar to fi_cq_read().  The tcp provider already supports this, but the verbs provider does not.  This patch is the simpler fix to try to hit the v1.15.0 release.  The update to the verbs provider is slightly more involved.

### DIFF
--- a/prov/rxm/src/rxm_conn.c
+++ b/prov/rxm/src/rxm_conn.c
@@ -904,16 +904,30 @@ static void *rxm_cm_progress(void *arg)
 	while (ep->do_progress) {
 		ofi_ep_lock_release(&ep->util_ep);
 
+		/* We must retrieve any event after we acquire the ep lock.
+		 * Otherwise, we could obtain an event for a msg ep that
+		 * another thread could be closing.  If we try to process that
+		 * event, we can access freed memory.  So, we use FI_PEEK here
+		 * to wait until an event is ready, then read any event after
+		 * we hold the ep lock.  Because closing an ep will free any
+		 * events queued on the eq, the event we find here may be gone
+		 * by the time we call read below.  This is what we want as it
+		 * avoids processing the stale event.
+		 */
 		ret = fi_eq_sread(ep->msg_eq, &event, &cm_entry,
-				  sizeof(cm_entry), -1, 0);
+				  sizeof(cm_entry), -1, FI_PEEK);
 
 		ofi_ep_lock_acquire(&ep->util_ep);
+		if (ret > 0) {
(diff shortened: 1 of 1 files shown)


chien-intel (commented):
Looks like a good fix to me

PR #7716, opened 2022-04-28
contrib/intel/jenkins: Add OneCCL GPU tests to CI

Juee14Desai:
- Add OneCCL examples and functional test suites in GPU mode
to libfabric CI.
- Tests run with tcp provider only. There is an issue with
OFI MPI init hook while running tests with psm3 provider.
This needs to be investigated separately.
- One of the functional tests (allreduce_test) is not added
as it is resulting in a segmentation fault. OneCCL team
needs to be consulted for this issue.
- Takes 1 min 42 secs to complete.

Signed-off-by: Juee Himalbhai Desai <juee.himalbhai.desai@intel.com>

### DIFF
--- a/contrib/intel/jenkins/run.py
+++ b/contrib/intel/jenkins/run.py
@@ -157,5 +157,27 @@ def oneccltest(core, hosts, mode, util):
               .format(runoneccltest.testname))
     print("-------------------------------------------------------------------")
 
+def oneccltestgpu(core, hosts, mode, util):
+
+    runoneccltestgpu = tests.OneCCLTestsGPU(jobname=jbname,buildno=bno,
+                                         testname="oneccl GPU test", core_prov=core,
+                                         fabric=fab, hosts=hosts,
+                                         ofi_build_mode=mode, util_prov=util)
+
+    print("-------------------------------------------------------------------")
+    if (runoneccltestgpu.execute_condn):
+        print("Running oneCCL GPU examples test for {}-{}-{}" \
+              .format(core, util, fab))
+        runoneccltestgpu.execute_cmd('examples')
+
+        print("---------------------------------------------------------------")
+        print("Running oneCCL GPU functional test for {}-{}-{}" \
+              .format(core, util, fab))
+        runoneccltestgpu.execute_cmd('functional')
+    else:
+        print("Skipping {} as execute condition fails" \
+              .format(runoneccltestgpu.testname))
+    print("-------------------------------------------------------------------")
+
 if __name__ == "__main__":
     pass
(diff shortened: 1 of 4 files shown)


fabtests/ubertest/uber.c
```diff
@@ -676,7 +677,12 @@ int main(int argc, char **argv)
 	char *service = "2710";
 	opts = INIT_OPTS;
 	int ret, op;
+/*	volatile int i = 10;
 
+	while (i == 10)
+	{
+	}
+*/
```
shefty, line 685: Please revert this change.

shefty (changes requested):
Revert uber.c changes.

shefty:
Can you please rebase?  @zachdworkin - can you check these changes?

contrib/intel/jenkins/run.py
```diff
@@ -157,5 +157,27 @@ def oneccltest(core, hosts, mode, util):
...
+    runoneccltestgpu = tests.OneCCLTestsGPU(jobname=jbname,buildno=bno,
+                                         testname="oneccl GPU test", core_prov=core,
+                                         fabric=fab, hosts=hosts,
+                                         ofi_build_mode=mode, util_prov=util)
+
+    print("-------------------------------------------------------------------")
+    if (runoneccltestgpu.execute_condn):
+        print("Running oneCCL GPU examples test for {}-{}-{}" \
+              .format(core, util, fab))
+        runoneccltestgpu.execute_cmd("examples")
```
zachdworkin, line 171: This is nitpicky but can you change all of your "single word" strings aka strings that contain no spaces to use the single quoted 'string' instead of the double quoted "string" to match the other files?

contrib/intel/jenkins/tests.py
```diff
@@ -651,3 +651,78 @@ def execute_cmd(self, oneccl_test):
...
+                                }
+
+    @property
+    def cmd(self):
+        return "{}/run_oneccl_gpu.sh ".format(ci_site_config.testpath)
+
+    def options(self, oneccl_test_gpu):
+        opts = "-n {n} ".format(n=self.n)
+        opts += "-ppn {ppn} ".format(ppn=self.ppn)
+        opts += "-hosts {server},{client} ".format(server=self.server,
```
zachdworkin, line 702: Do these -hosts need to contain server-interface,client-interface instead of just self.server,self.client?

If they need it: the other tests use common.get_node_name(self.server, self.nw_interface_map) to format this for you.

zachdworkin (commented):
Overall this looks good. I am worried about not explicitly stating the interfaces you need since this is running tcp on the ze node which will default to not using any specific interface.

As of now the 'ze' fabric will tell you it doen't care about an interface which will mean using common.get_node_name will return server-,client- which is not what you want. That function will need to change to correctly return server,client if the interface value is none.

Juee14Desai:
Will take care of the server, client interfaces in a separate commit once ib interface is available on ze nodes.

PR #7717, opened 2022-04-29
prov/efa: info refactor part2

wzamazon:
keep refactoring getinfo code path.


prov/efa/src/efa_prov_info.c
```diff
@@ -384,13 +371,19 @@ int efa_prov_info_alloc(struct fi_info **info,
 	if (!fi)
 		return -FI_ENOMEM;
 
-	fi->caps		= ep_dom->caps;
+	if (ep_type != FI_EP_RDM && ep_type != FI_EP_DGRAM) {
+		EFA_WARN(FI_LOG_DOMAIN, "Unsupported endpoint type: %d\n",
+			 ep_type);
```
shijin-aws, line 493: Should we fail here?

prov/efa/src/efa_ep.c
```diff
@@ -622,68 +615,64 @@ static struct fi_ops_atomic efa_ep_atomic_ops = {
 	.compwritevalid = fi_no_atomic_compwritevalid,
 };
 
-int efa_ep_open(struct fid_domain *domain_fid, struct fi_info *info,
```
shijin-aws, line 625: since you modified it, could u add a docstring.

prov/efa/src/efa_prov_info.c
```diff
@@ -64,13 +64,43 @@
 
 #define EFA_NO_DEFAULT -1
```
shijin-aws, line 65: This commit has no commit message, could you add it?

prov/efa/src/efa_prov_info.c
wzamazon: Yes, we should. In new revision, I returned `-FI_ENODATA` here.

prov/efa/src/efa_ep.c
wzamazon: I plan to refactor this function next.

I will move the rdm part of this function to `rxr_endpoing`, then rename it`efa_dgram_ep_open`. So I would prefer to  add docstring at that time.

prov/efa/src/efa_prov_info.c
wzamazon: Added in new revision.

prov/efa/src/efa_prov_info.c
```diff
@@ -259,118 +439,55 @@ static int efa_alloc_fid_nic(struct fi_info *fi, struct efa_device *device)
...
+		prov_info->caps			|= FI_HMEM;
+		prov_info->tx_attr->caps		|= FI_HMEM;
+		prov_info->rx_attr->caps		|= FI_HMEM;
+		prov_info->domain_attr->mr_mode	|= FI_MR_HMEM;
 	}
-
-	return 0;
 }
+#else
+void efa_prov_info_set_hmem_flags(struct fi_info *prov_info)
```
shijin-aws, line 460: Why not have the `#if` inside the function?

shijin-aws (approved):

prov/efa/src/efa_prov_info.c
wzamazon: `#if` outside function is libfabric's preference.

Issue #7715, opened 2022-04-28
prov/opx: provider is reported even when OPA hw is not present


j-xiong:
**Describe the bug**
On machines w/o OPA hardware, fi_getinfo() still returns opx as a valid provider. 

**To Reproduce**
Run `fi_info -p opx` on systems w/o OPA hw and get the result:
```
provider: opx
    fabric: OPX-100
    domain: hfi1
    version: 115.0
    type: FI_EP_UNSPEC
    protocol: FI_PROTO_OPX
```
In addition, the opx provider is listed before the verbs provider so applications that use the first returned provider may end up not working properly.

**Expected behavior**
fi_getinfo returns -61

**Output**
n/a

**Environment:**
Linux

**Additional context**
Comment from DAOS issue https://github.com/daos-stack/libfabric/pull/56:

> [@frostedcmos](https://github.com/frostedcmos) , Lei discovered that `--disable-opx` is required on a cluster with Mellanox / IB for ofi to properly return that verbs is a provider that we can use. this is why i was having issues with using 1.15rc3 on thee io500 cluster. i just updated with ofi build without changing the configure options. I don't know if this is really the intention, but why would one need to explicitly disable opx on an IB cluster and why would it not report that verbs is an available provider? we don't even have opa on those nodes. sounds like a bug to me, but i don't know whether we should hold this patch for it.

timothom64:
Yea this isn't right.  Will fix

timothom64:
I will try

timothom64:
In CI testing now

timothom64:
Created pr

PR #7719, opened 2022-04-30
prov/opx: Do not return opx as a provider on systems without hardware

timothom64:
Fix bug in opx provider where it was returning itself as
a provider choice on systems with no Omnipath hardware.

ofiwg/libfabric Github Issue #7715

Signed-off-by: Tim Thompson <tim.thompson@cornelisnetworks.com>


shefty:
Thanks!

shefty:
Picked for v1.15.x -- in the v1.15.0 PR

wzamazon (approved):

PR #7722, opened 2022-05-03
fabtests/efa: do not run efa rnr test with strict mode

shijin-aws:
Older efa kernel driver does not support RNR retry capability
and the test will return FI_ENOSYS. Disable the strict mode for
this test explicitly to mark it as skipped in this case.

Test info

`./runfabtests --strict -b -p <bin_path> efa <server_ip> <client_ip>`

Tested on efa instances that have efa kernel support rnr_retry, `test_rnr*` tests passed

Tested on efa instances that have efa kernel not support rnr_retry, `test_rnr*` tests skipped

### DIFF
--- a/fabtests/pytest/efa/test_rnr.py
+++ b/fabtests/pytest/efa/test_rnr.py
@@ -1,4 +1,5 @@
 import pytest
+import copy
 
 @pytest.mark.functional
 def test_rnr_read_cq_error(cmdline_args):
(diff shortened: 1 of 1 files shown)


shijin-aws:
AWS CI failed due to DSO build make failed. Will push a new revision to fix it.

shijin-aws:
@wzamazon yes please

fabtests/prov/efa/Makefile.include
```diff
@@ -51,3 +51,5 @@ prov_efa_src_fi_efa_rnr_queue_resend_LDADD = libfabtests.la
 prov_efa_src_fi_efa_info_test_SOURCES = \
 	prov/efa/src/efa_info_test.c
 prov_efa_src_fi_efa_info_test_LDADD = libfabtests.la
+
+nobase_dist_config_DATA += pytest/efa/efa_device_caps.c
```
wzamazon, line 55: Do you need some configure to rdma-core? since this program explicitly uses rdma-core functions.

shefty:
Intel CI is failing tests on the psm3 provider.  We're looking into it, but don't let that CI hold you up on any PRs that doesn't touch anything in core in around that provider.

fabtests/prov/efa/Makefile.include
shijin-aws: I switched to turn off the strict mode for rnr test in the latest revision, please review again.

wzamazon:
This is a pure EFA change, so ignore intel CI error

PR #7723, opened 2022-05-03
prov/rxm: Cast bit-field to avoid signed behavior on windows

dshinaberry:
The windows msvc compiler interprets a 2-bit wide bit field enum as a signed
value. Explicitly cast to unsigned and mask off any remaining sign bits to
obtain behavior from msvc that is equivalent to gcc.

On-behalf-of: https://github.com/cmru-ps <dshinaberry@mru.medical.canon>

### DIFF
--- a/prov/rxm/src/rxm.h
+++ b/prov/rxm/src/rxm.h
@@ -442,7 +442,7 @@ union rxm_sar_ctrl_data {
 static inline enum rxm_sar_seg_type
 rxm_sar_get_seg_type(struct ofi_ctrl_hdr *ctrl_hdr)
 {
-	return ((union rxm_sar_ctrl_data *)&(ctrl_hdr->ctrl_data))->seg_type;
+	return (uint8_t)((union rxm_sar_ctrl_data *)&(ctrl_hdr->ctrl_data))->seg_type & 0x3;
 }
 
 static inline void


dshinaberry:
I don't have visibility into the output from the failed CI build. Can someone point out what I broke? Looks like maybe this is going to happen to all the PRs I created. Sorry. ;-(

j-xiong:
**psm3**
```
06:56:02      ci14.cluster:pid337295.fi_ubertest: unexpected send completion on mlx5_0 port 1 opcode 136 QP 17851
06:56:02      ci14.cluster:pid337295: Assertion failure at prov/psm3/psm3/hal_verbs/verbs_ep.c:1103: b == buf
```

**verbs;ofi_rxm**
```
07:19:03  # OSU MPI Latency Test v5.7
07:19:03  # Size          Latency (us)
07:29:10  Abort(271176463) on node 0 (rank 0 in comm 0): Fatal error in PMPI_Send: Other MPI error, error stack:
07:29:10  PMPI_Send(159)...................: MPI_Send(buf=0x7f10f4565000, count=0, MPI_CHAR, dest=1, tag=1, MPI_COMM_WORLD) failed
07:29:10  MPID_Send(882)...................: 
07:29:10  MPIDI_send_unsafe(190)...........: 
07:29:10  MPIDI_OFI_send_lightweight(44)...: 
07:29:10  MPIDI_OFI_inject_handler_vci(677): OFI tagged inject failed (ofi_impl.h:677:MPIDI_OFI_inject_handler_vci:Cannot allocate memory)
```
```
07:29:26  # OSU MPI Multi-threaded Latency Test v5.7
07:29:26  # Size          Latency (us)
07:40:02  
07:40:02  ===================================================================================
07:40:02  =   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
07:40:02  =   RANK 0 PID 185706 RUNNING AT ci1
07:40:02  =   KILLED BY SIGNAL: 9 (Killed)
07:40:02  ===================================================================================
```

dshinaberry:
Seems like a more general issue and not connected to the change in the PR then.

prov/rxm/src/rxm.h
```diff
@@ -433,7 +433,7 @@ union rxm_sar_ctrl_data {
 			RXM_SAR_SEG_FIRST	= 1,
 			RXM_SAR_SEG_MIDDLE	= 2,
 			RXM_SAR_SEG_LAST	= 3,
-		} seg_type : 2;
+		} seg_type : 3;
```
shefty, line 436: Gurgle.  Can we agree that an enum in a protocol header is goofy?  The rxm headers look broken for ever trying to support mixed endian communication.

Do we know where these bits land before and after the change?  We need to maintain wire compatibility between this and older versions of libfabric.  I'm inclined to replace this enum with a uint8_t seg_type, redefine RXM_SAR_SEG_XXX so that it aligns on the same bits, and add uint8_t resv[3] to cover the gap.  That would help us to eventually support mixed endianness.

shefty: Btw, if we can confirm that the bits are set the same before and after this change, I'm okay merging this and marking fixing this as a TODO

dshinaberry: What if instead of changing the size of the bit field, we change
```
static inline enum rxm_sar_seg_type
rxm_sar_get_seg_type(struct ofi_ctrl_hdr *ctrl_hdr)
{
	return ((union rxm_sar_ctrl_data *)&(ctrl_hdr->ctrl_data))->seg_type;
}
```
to be
```
static inline enum rxm_sar_seg_type
rxm_sar_get_seg_type(struct ofi_ctrl_hdr *ctrl_hdr)
{
	return (uint8_t)((union rxm_sar_ctrl_data *)&(ctrl_hdr->ctrl_data))->seg_type & 0x3;
}
```
This is maybe a more direct response to the crux of the issue, which is that the MSVC compiler treats those 2 bits as a signed value. So, we explicitly tell it to treat it as unsigned and then lop off any sign bits.

shefty: I think the 2 functions you listed are what we want.  What I also want to do is remove the enum.  In theory the enum should convert to an int, which I'm going to assume is 32-bits on whatever platforms we're building on.  So, we could replace the enum with uint32_t.  We need to verify that the same bits are being set before and after the changes (which I think they are).

shefty: Unless we add checks for byte and bitfield ordering, IMO, it's much easier to drop the bit fields and just set the bits that we want.  Then standard byte swapping routines could be added in the future to make the protocol work between different endian systems.

dshinaberry: I think I'm getting confused about what you want now versus what is for a potential future. And also compatibility with previous versions of libfabric versus compatibility between different endian systems.

If I understand the assembly language correctly, on x86_64 it appears the compilers (gcc and msvc) are using the 2 lsb of the LSB of the 64-bit `union rxm_sar_ctrl_data`, i.e. the lowest byte in memory. On power64 gcc it appears to be using the 2 msb of the MSB of the 64-bit `union rxm_sar_ctrl_data`, i.e. the lowest byte in memory. So, I don't see how there can be any compatibility between different endian systems with this structure layout.

If we do something like
```
enum rxm_sar_seg_type2 {
	RXM_SAR_SEG2_FIRST	= 1,
	RXM_SAR_SEG2_MIDDLE	= 2,
	RXM_SAR_SEG2_LAST	= 3,
};
union rxm_sar_ctrl_data2 {
    struct {
        uint32_t seg_type;
    	uint32_t offset;
    };
	uint64_t align;
};
```
Then on x86_64, it is still using the same bits, i.e. the 2 lsb of the LSB. And on power64 it uses the 2 lsb of the 4th byte, i.e. MSB-3 of the 64-bit union. So a htonl(seg_type) of the x86 would put the bits at the right spot for power64. But this wouldn't be backward compatible for a big endian system; only a little endian system.

So, I don't see how we can change anything about the structure and still maintain backward compatibility with previous versions.

I know you have a broader scope about this than I do as I just want to get this working correctly on windows. I submit that casting the bitfield enum to uint8_t and then masking off the sign bits with `& 0x3` as I suggested above will be backward compatible for non-windows and add windows compatibility.

shefty: I don't see a way to fix compatibility between different endian systems without breaking one of them.  That's a future fix to me.

What I'm looking for now is to fix this for windows without changing the wire format at all, so that the next libfabric version (v1.16) can still talk with an older version (v1.15).  This assumes we're speaking the same endianness on both sides.  If I'm following your comment, then we *conceptually* end up with this today:
```
#if LITTLE_ENDIAN
enum {
    RXM_SAR_SEG_FIRST = 0x00000001,
    RXM_SAR_SEG_MIDDLE = 0x00000002,
    RXM_SAR_SEG_LAST = 0x00000003,
};
#else
enum {
    RXM_SAR_SEG_FIRST = 0x4000000,
    RXM_SAR_SEG_MIDDLE = 0x80000000,
    RXM_SAR_SEG_LAST = 0xC0000000,
};
#endif

struct {
    uint32_t seg_type;
    uint32_t offset;
};
#endif
```
Is this correct?

shefty: Adding to my previous comment, by changing the bit field from 2 to 3, does that shift the values by 1 bit in the big endian case?  That's my concern.  I'm struggling to find if the C language defines if bit fields must map to specific bits, or if that's left up to the compiler to decide.

dshinaberry: In response to your previous comment, I believe that is correct.

I do understand your concern about that now. That's why I suggested the alternative where we don't touch the structure definition at all, leave the bit field at 2 and instead change the implementation of rxm_sar_get_seg_type to:

```
static inline enum rxm_sar_seg_type
rxm_sar_get_seg_type(struct ofi_ctrl_hdr *ctrl_hdr)
{
	return (uint8_t)((union rxm_sar_ctrl_data *)&(ctrl_hdr->ctrl_data))->seg_type & 0x3;
}
```

I don't think there was any issue with setting the field. It was just that when msvc extracted those 2 bits, it treated them as a 2 bit int rather than a 2 bit uint. So casting it to an unsigned and masking off all but the 2 bits we are interested in should make windows happy.

shefty: Ah, got it.  Yes, if we can cast away the warning, I'm okay with that.

PR #7724, opened 2022-05-03
prov/util: Try IPv4 before IPv6 addresses when starting name server

dshinaberry:
On Windows, getaddrinfo returned IPv6 addresses before IPv4 addresses leading
to the name server being started on an IPv6 address. However, name server
clients were attempting to connect to an IPv4 address for the name server.
Trying IPv4 addresses before IPv6 addresses leads to more deterministic
behavior.

### DIFF
--- a/prov/util/src/util_ns.c
+++ b/prov/util/src/util_ns.c
@@ -122,8 +122,8 @@ static int util_ns_map_del(struct util_ns *ns, void *service_in,
 	int ret = -FI_ENOENT;
 	void *service, *name;
 
-        it = rbtFind(ns->map, service_in);
-        if (it) {
+	it = rbtFind(ns->map, service_in);
+	if (it) {
 		rbtKeyValue(ns->map, it, &service, &name);
 		if (memcmp(name, name_in, ns->name_len))
 			return ret;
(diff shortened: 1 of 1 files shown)


prov/util/src/util_ns.c
```diff
@@ -260,23 +260,30 @@ static int util_ns_listen(struct util_ns *ns)
...
-			continue;
-
-		(void) setsockopt(ns->listen_sock, SOL_SOCKET, SO_REUSEADDR,
-				  (void *) &n, sizeof(n));
-		ret = bind(ns->listen_sock, p->ai_addr, (socklen_t) p->ai_addrlen);
-		if (!ret)
-			break;
+	// Linux seems to return IPv4 addresses before IPv6 ones and Windows does
+	// the opposite. So, try IPv4 addresses first followed by IPv6 addresses.
+	for (f = AF_INET; f < AF_INET6; f += (AF_INET6 - AF_INET)) {
```
shefty, line 265: Rather than having this questionable for loop, I would rather pass in the address family into util_ns_listen() and use that as a hint for getaddrinfo().  Then call util_ns_listen() once for each address family if the first call fails.

shefty (approved):

PR #7725, opened 2022-05-03
core/pollfds: Fix behavior of element deletion from middle of array

dshinaberry:
ofi_pollfds_do_del was not correctly handling the deletion of an element from
the middle of the array leading to an assert being triggered when a subsequent
addition to the array was made.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1592,8 +1592,8 @@ int ofi_pollfds_del(struct ofi_pollfds *pfds, int fd)
 static void ofi_pollfds_do_del(struct ofi_pollfds *pfds,
 			       struct ofi_pollfds_work_item *item)
 {
-	struct ofi_pollfds_ctx *ctx, *swap_ctx;
-	struct pollfd *swap_pfd;
+	struct ofi_pollfds_ctx *ctx, *last_ctx;
+	struct pollfd *last_pfd;
 
 	ctx = ofi_pollfds_get_ctx(pfds, item->fd);
 	if (!ctx)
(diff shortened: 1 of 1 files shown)


src/common.c
```diff
@@ -1602,19 +1602,22 @@ static void ofi_pollfds_do_del(struct ofi_pollfds *pfds,
...
-		assert(swap_ctx);
-		swap_ctx->index = ctx->index;
-		pfds->fds[swap_ctx->index] = *swap_pfd;
-
-		swap_pfd->fd = INVALID_SOCKET;
-		swap_pfd->events = 0;
-		swap_pfd->revents = 0;
-	}
+		last_ctx->index = ctx->index;
+		pfds->ctx[ctx->index] = *last_ctx;
```
shefty, line 1610: This looks like the relevant change.  Can you describe the problem in more detail in the commit message?  The code below the if statement mostly appears preventative.

src/common.c
```diff
@@ -1602,19 +1602,22 @@ static void ofi_pollfds_do_del(struct ofi_pollfds *pfds,
...
+		pfds->fds[ctx->index] = *last_pfd;
+	}
+	last_pfd->fd = INVALID_SOCKET;
+	last_pfd->events = 0;
+	last_pfd->revents = 0;
+	last_ctx->context = 0;
```
shefty, line 1616: Please use NULL for pointer assignments, rather than 0.

src/common.c
dshinaberry: Preventative, but required. When adding an element there is an assert that the element it is about to use is not already in use `assert(ctx->index == -1)` Tripping that assert is what eventually led me here.

In addition to not copying the ofi_pollfds_ctx from the last position to the position being deleted, the setting of `ctx->index = -1` was not correct when the if statement had been taken, because it would obliterate the very element that it had just moved.

I'll make the 0 a NULL and update the commit message to reflect what I just described.

src/common.c
```diff
@@ -1602,19 +1602,22 @@ static void ofi_pollfds_do_del(struct ofi_pollfds *pfds,
 	if (ofi_poll_fairness && ctx->hot_index >= 0)
 		ofi_pollfds_coolfd(pfds, item->fd);
 
+	last_pfd = &pfds->fds[pfds->nfds - 1];
+	last_ctx = ofi_pollfds_get_ctx(pfds, last_pfd->fd);
```
shefty, line 1606: Btw, I used swap_ctx rather than last_ctx, because this may not be the last context.

dshinaberry: Doesn't it have to be the last [used] context by definition? `[pfds->nfds-1]`
I left the 'used' out of the rename. I can switch it back to swap if you prefer.

src/common.c
shefty: On linux, this works as follows.  The fds[] is a compact array of file descriptors.  The ctx[] is a sparse array of context values.  The value of any fd is used as a direct index into the ctx[] array.  That provides a mapping where the fd is stored in the fds[] array.  So, the ctx[] structures are never swapped, only the fds[] entry changes.  So, the changes below where last_ctx is modified would break linux as there's no swapping occurring.  The change you added to line 1610 ends up being unnecessary.  I think we may need to pull ofi_pollfds_do_del() back out into OS specific implementations, or maybe introduce a new OS specific swap function.

dshinaberry: Hmm...Guess I was a little too focused on windows and didn't appreciate the amount of nuance here.

shefty: That's okay, I was focused on linux mostly when I wrote the code.  :)
But I think that for linux, we need to swap the fds[] entries only.  For windows, we either need to swap both or update the windows code to allow for unused ctx[] entries.  The latter might be possible by updating ofi_pollfds_alloc_ctx() to return the first unused ctx[] entry.

shefty:
@dshinaberry - Please see #7734 for a possible fix and let me know if that one works for you.  That fix updates the windows specific ofi_pollfds_get/alloc_ctx calls.

shefty:
#7734 isn't correct.  I need to update the get_ctx() call.

shefty:
#7734 was updated

dshinaberry:
@shefty - #7734 appears to work fine in my windows test workspace.

shefty:
@dshinaberry Thanks - if you're okay with that change, I'll merge it once the windows CI passes and close this PR.

dshinaberry:
@shefty - Sounds good to me.

PR #7726, opened 2022-05-03
prov/verbs_nd: Various fixes for test failures and minor feature additions

dshinaberry:
- Move checking of pending asynchronous callbacks into the overlapped structure
  and add locking to prevent race conditions between rdma/ibv functions and
  asynchronous completion callbacks.

- Add support to ibv_post_send for opcodes IBV_WR_RDMA_READ/WRITE.

- Improve disconnect handling by better distinguishing between host initiated
  disconnect and peer initiated disconnect and removing some race conditions.

### DIFF
--- a/prov/verbs/include/windows/verbs_nd.h
+++ b/prov/verbs/include/windows/verbs_nd.h
@@ -116,7 +116,7 @@ struct nd_infrastructure {
 extern struct nd_infrastructure nd_infra;
 
 static inline struct nd_adapter *
-get_adapter_by_context(struct ibv_context *context)
+nd_get_adapter_by_context(struct ibv_context *context)
 {
 	for (int i = 0; i < nd_infra.adapters.count; ++i) {
 		if (nd_infra.adapters.context_list[i] == context) {
@@ -127,7 +127,7 @@ get_adapter_by_context(struct ibv_context *context)
 }
 
 static inline struct ibv_context *
-get_context_by_device(struct ibv_device *device)
+nd_get_context_by_device(struct ibv_device *device)
 {
 	for (int i = 0; i < nd_infra.adapters.count; ++i) {
 		if (nd_infra.adapters.device_list[i] == device) {
@@ -145,7 +145,12 @@ struct nd_event_base {
 	OVERLAPPED ov;
 	nd_event_handler event_cb;
 	nd_error_handler error_cb;
+	ofi_mutex_t lock;
+	pthread_cond_t cond;
+	uint32_t cb_pending;
 };
+
+HRESULT nd_cancel_pending(struct nd_event_base *event, IND2Overlapped *ov);
 
 void CALLBACK nd_io_cb(DWORD error, DWORD bytes, LPOVERLAPPED ov);
 
@@ -165,8 +170,8 @@ void nd_cm_connreq_error(struct nd_event_base *base, DWORD bytes, DWORD error);
 //   IND2Connector::CompleteConnect()
 //   IND2Connector::Disconnect()
 //   IND2Connector::NotifyDisconnect()
-enum nd_cm_state {
-	ND_CM_IDLE,
+enum nd_cm_event_type {
+	ND_CM_NONE,
 	ND_CM_CONNECT,
 	ND_CM_ACCEPT,
 	ND_CM_COMPLETE,
@@ -176,7 +181,7 @@ enum nd_cm_state {
 
 struct nd_cm_connect_event {
 	struct nd_event_base base;
-	enum nd_cm_state state;
+	enum nd_cm_event_type type;
 	IND2Connector *connector;
 	struct rdma_cm_id *id;
 	struct rdma_event_channel *channel;
@@ -188,9 +193,6 @@ struct nd_cm_connect_event {
 		uint8_t initiator_depth;
 	} param;
 };
-
-void nd_get_connection_data(IND2Connector *connector,
-			    struct nd_cm_event *entry_nd);
 
 void nd_cm_connect_ack(struct nd_event_base *base, DWORD bytes);
 void nd_cm_connect_nack(struct nd_event_base *base, DWORD bytes, DWORD error);
@@ -207,7 +209,7 @@ struct nd_cm_id {
 	struct nd_cm_connect_event connect_event;
 	struct nd_cm_connect_event peer_event;
 	IND2Listener *listener;
-    struct nd_cm_listen_event listen_event;
+	struct nd_cm_listen_event listen_event;
 	ofi_atomic32_t cm_events_pending;
 };
 
@@ -225,8 +227,6 @@ struct nd_cq {
 	struct ibv_cq cq;
 	HANDLE ov_file;
 	IND2CompletionQueue *nd2cq;
-	ofi_mutex_t lock;
-	bool notify_pending;
 	struct nd_event_base notification;
 	struct dlist_entry entry;
 };
--- a/prov/verbs/src/windows/verbs_nd_ibv.c
+++ b/prov/verbs/src/windows/verbs_nd_ibv.c
@@ -54,7 +54,7 @@ void ibv_free_device_list(struct ibv_device **list)
 struct ibv_context *ibv_open_device(struct ibv_device *device)
 {
 	VRB_TRACE(FI_LOG_FABRIC, "\n");
-	return get_context_by_device(device);
+	return nd_get_context_by_device(device);
 }
 
 int ibv_close_device(struct ibv_context *context)
@@ -85,7 +85,7 @@ int ibv_query_device(struct ibv_context *context,
 		return errno;
 	}
 
-	adapter = get_adapter_by_context(context);
+	adapter = nd_get_adapter_by_context(context);
 	if (!adapter) {
 		errno = ENODEV;
 		return errno;
@@ -139,7 +139,7 @@ int ___ibv_query_port(struct ibv_context *context, uint8_t port_num,
 		return errno;
 	}
 
-	adapter = get_adapter_by_context(context);
+	adapter = nd_get_adapter_by_context(context);
 	if (!adapter) {
 		errno = ENODEV;
 		return errno;
@@ -168,8 +168,8 @@ int ibv_query_gid(struct ibv_context *context, uint8_t port_num, int index,
 {
 	VRB_TRACE(FI_LOG_FABRIC, "\n");
 
-	gid->global.subnet_prefix =
-		0x80fe; // little endian representation of link-local address prefix.
+	// little endian representation of link-local address prefix.
+	gid->global.subnet_prefix = 0x80fe;
 	gid->global.interface_id = 0;
 
 	return 0;
@@ -292,7 +292,7 @@ struct ibv_cq *ibv_create_cq(struct ibv_context *context, int cqe,
 		return NULL;
 	}
 
-	adapter = get_adapter_by_context(context);
+	adapter = nd_get_adapter_by_context(context);
 	if (!adapter) {
 		errno = ENODEV;
 		return NULL;
@@ -307,9 +307,10 @@ struct ibv_cq *ibv_create_cq(struct ibv_context *context, int cqe,
 	if (channel)
 		++channel->refcnt;
 
+	ofi_mutex_init(&cq_nd->notification.lock);
+	pthread_cond_init(&cq_nd->notification.cond, 0);
 	cq_nd->notification.event_cb = nd_cq_notify_event;
 	cq_nd->notification.error_cb = nd_cq_notify_error;
-	ofi_mutex_init(&cq_nd->lock);
 
 	cq = &cq_nd->cq;
 	cq->context = context;
@@ -352,8 +353,6 @@ err1:
 int ibv_destroy_cq(struct ibv_cq *cq)
 {
 	struct nd_cq *cq_nd;
-	HRESULT hr;
-	int delay = 0;
 
 	VRB_TRACE(FI_LOG_FABRIC, "\n");
 
@@ -363,29 +362,10 @@ int ibv_destroy_cq(struct ibv_cq *cq)
 	}
 
 	cq_nd = container_of(cq, struct nd_cq, cq);
-	ofi_mutex_lock(&cq_nd->lock);
-	if (cq_nd->notify_pending) {
-		hr = cq_nd->nd2cq->lpVtbl->CancelOverlappedRequests(
-			cq_nd->nd2cq);
-		if (FAILED(hr)) {
-			errno = hresult2fi(hr);
-			return errno;
-		}
-		ofi_mutex_unlock(&cq_nd->lock);
-		pthread_yield();
-
-		ofi_mutex_lock(&cq_nd->lock);
-		while (true) {
-			if (cq_nd->notify_pending) {
-				ofi_mutex_unlock(&cq_nd->lock);
-				usleep(1 << delay++);
-				ofi_mutex_lock(&cq_nd->lock);
-			} else {
-				break;
-			}
-		}
+	if (nd_cancel_pending(&cq_nd->notification,
+			      (IND2Overlapped *)cq_nd->nd2cq)) {
+		return errno;
 	}
-	ofi_mutex_unlock(&cq_nd->lock);
 
 	if (!CloseHandle(cq_nd->ov_file)) {
 		errno = GetLastError();
@@ -403,7 +383,8 @@ int ibv_destroy_cq(struct ibv_cq *cq)
 	if (cq_nd->cq.channel)
 		--cq_nd->cq.channel->refcnt;
 
-	ofi_mutex_destroy(&cq_nd->lock);
+	pthread_cond_destroy(&cq_nd->notification.cond);
+	ofi_mutex_destroy(&cq_nd->notification.lock);
 	free(cq_nd);
 
 	return 0;
@@ -437,7 +418,8 @@ int ibv_poll_cq(struct ibv_cq *cq, int num_entries, struct ibv_wc *wc)
 		wc[num_results].opcode = result.RequestType;
 		FI_LOG(&vrb_prov, result.Status ? FI_LOG_WARN : FI_LOG_DEBUG,
 		       FI_LOG_CQ,
-		       "ibv_poll_cq: context=0x%016llx, numBytes=%d, status=0x%08lx, request=0x%08lx\n",
+		       "ibv_poll_cq: context=0x%016llx, numBytes=%d, "
+		       "status=0x%08lx, request=0x%08lx\n",
 		       (uint64_t)result.RequestContext, result.BytesTransferred,
 		       result.Status, result.RequestType);
 	}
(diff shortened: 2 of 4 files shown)


prov/verbs/src/windows/verbs_nd_ibv.c
```diff
@@ -791,11 +783,25 @@ int ibv_post_recv(struct ibv_qp *qp, struct ibv_recv_wr *wr,
...
+	for (int i = 0; i < wr->num_sge; ++i) {
+		length += wr->sg_list[i].length;
+	}
+
+	if (length <= qp_nd->max_inline_data) {
+		return ND_OP_FLAG_INLINE;
+	}
+
+	return 0;
+}
```
shefty, line 799: This function shouldn't be needed.  The libfabric API returns the max_inject_size, which aligns with the max_inline_data size for verbs.  Callers should have already checked their message sizes and used either an inject call or the FI_INJECT flag to leverage this feature.  Adding this call may duplicate checks that were already made.  You may just need to check for the IBV_SEND_INLINE flag having been set.

shefty, line 799: This function shouldn't be needed.  The libfabric API returns the max_inject_size, which aligns with the max_inline_data size for verbs.  Callers should have already checked their message sizes and used either an inject call or the FI_INJECT flag to leverage this feature.  Adding this call may duplicate checks that were already made.

prov/verbs/src/windows/verbs_nd_ibv.c
```diff
@@ -806,34 +812,61 @@ int ibv_post_send(struct ibv_qp *qp, struct ibv_send_wr *wr,
...
+			break;
+
+		case IBV_WR_RDMA_READ:
+			hr = qp_nd->nd2qp->lpVtbl->Read(
+				qp_nd->nd2qp, (void *)wr->wr_id,
+				(ND2_SGE *)wr->sg_list, wr->num_sge,
+				wr->wr.rdma.remote_addr, wr->wr.rdma.rkey,
+				flags);
+			VRB_DBG(FI_LOG_EP_DATA,
+				"IND2QueuePair::Read: hr=0x%08lx, nsge=%d, addr=0x%016llx, raddr=0x%016llx, rkey=0x%08lx, lkey=0x%08lx\n",
```
shefty, line 840: line lengths are getting long

shefty, line 840: line lengths are getting long

prov/verbs/src/windows/verbs_nd_ov.c
```diff
@@ -313,29 +325,48 @@ void nd_cm_connect_ack(struct nd_event_base *base, DWORD bytes)
 		if (!cm_event)
 			cm_event = nd_allocate_cm_event(
 				id, RDMA_CM_EVENT_ESTABLISHED);
+		break;
```
shefty, line 328: Th first part of this comment applies to code 8 lines up, only I can't comment there.  The comment above says "// Intentional fall through."  I know some compilers will recognize "/* fall through */" as a comment to avoid generating a warning about a missing break statement.  I don't know what the windows compiler recognizes for this.

The second part of this comment is why did we fall through?  The comment under case ND_CM_ACCEPT looks like it's backwards.  We have an cm_event if we fell through.  And in falling through, we don't actually do anything but break.

shefty, line 328: This comment applies to code 8 lines up, only I can't comment there.  The comment above says "// Intentional fall through."  I know some compilers will recognize "/* fall through */" as a comment to avoid generating a warning about a missing break statement.  I don't know what the windows compiler recognizes for this.

prov/verbs/src/windows/verbs_nd_ov.c
```diff
@@ -313,29 +325,48 @@ void nd_cm_connect_ack(struct nd_event_base *base, DWORD bytes)
...
+	VRB_DBG(FI_LOG_EQ, "EQ SET event:%p %p %d\n", cm_event->event.id,
+		cm_event->event.listen_id, cm_event->event.event);
+	dlistfd_insert_tail(&cm_event->entry, &ch_nd->q);
+
+	// The call to NotifyDisconnect needs to happen after we insert the
+	// CM event into the queue in order to avoid a race condition where
+	// a disconnect is virtually immediate and that callback interrupts
+	// this processing leading to an RDMA_CM_EVENT_DISCONNECTED coming
+	// before its associated RDMA_CM_EVENT_ESTABLISHED.
+	switch (event->state) {
```
shefty, line 348: I don't understand what's happening here.  There's a switch statement above this that checks for event->state.  AFAICT, there's no change to event, then we have another switch statement here on event->state.  Why are there 2 switch statements?

shefty, line 348: I don't understand what's happening here.  There's a switch statement above this that checks for event->state.  AFAICT, there's no change to event, then we have another switch statement here on event->state.  Why are there 2 switch statements?

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -168,15 +171,64 @@ int rdma_destroy_id(struct rdma_cm_id *id)
 	}
 
 	id_nd = container_of(id, struct nd_cm_id, id);
+	ofi_mutex_lock(&id_nd->connect_event.base.lock);
+	ofi_mutex_lock(&id_nd->peer_event.base.lock);
+	ofi_mutex_lock(&id_nd->listen_event.base.lock);
```
shefty, line 176: Please document this lock ordering where they are defined.

shefty, line 176: Please document this lock ordering where they are defined.

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -168,15 +171,64 @@ int rdma_destroy_id(struct rdma_cm_id *id)
...
+	    id_nd->peer_event.base.cb_pending)
+		id_nd->connector->lpVtbl->CancelOverlappedRequests(
+			id_nd->connector);
+
+	if (id_nd->listen_event.base.cb_pending)
+		id_nd->listener->lpVtbl->CancelOverlappedRequests(
+			id_nd->listener);
+
 	while (compare = ofi_atomic_get32(&id_nd->cm_events_pending)) {
 		WaitOnAddress(&id_nd->cm_events_pending, &compare,
```
shefty, line 187: Waiting on an event for an infinite time while holding 3 mutexes gives the impression of bad form.  Why do we need to hold these locks?  We're destroying them a little further down this function anyway.

shefty, line 187: Waiting on an event for an infinite time while holding 3 mutexes gives the impression of bad form.  Why do we need to hold these locks?

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -185,11 +237,21 @@ int rdma_destroy_id(struct rdma_cm_id *id)
...
 	// Should also destroy any associated SRQ here.
+
+	if (id_nd->ov_file != INVALID_HANDLE_VALUE) {
+		ret = CloseHandle(id_nd->ov_file);
+		assert(ret);
+	}
+
+	ofi_mutex_destroy(&id_nd->connect_event.base.lock);
+	ofi_mutex_destroy(&id_nd->peer_event.base.lock);
+	ofi_mutex_destroy(&id_nd->listen_event.base.lock);
```
shefty, line 254: This function is getting long.  Please break up the functionality into multiple calls.

shefty, line 254: This function is getting long.  Please break up the functionality into multiple calls.

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -168,15 +171,64 @@ int rdma_destroy_id(struct rdma_cm_id *id)
...
+		id_nd->peer_event.base.cb_pending ||
+		id_nd->listen_event.base.cb_pending) {
+		ofi_mutex_unlock(&id_nd->connect_event.base.lock);
+		ofi_mutex_unlock(&id_nd->peer_event.base.lock);
+		ofi_mutex_unlock(&id_nd->listen_event.base.lock);
+		pthread_yield();
+		ofi_mutex_lock(&id_nd->connect_event.base.lock);
+		ofi_mutex_lock(&id_nd->peer_event.base.lock);
+		ofi_mutex_lock(&id_nd->listen_event.base.lock);
+	}
```
shefty, line 201: Okay, I would look at creating a function that takes whatever object peer/connect/listen event is.  Have that function take any lock, check for pending callbacks, cancel the overlap request, wait for the cancel, etc.  Then just call it 3 times, once for each event.  That should eliminate the nested locking and shorten this function.

shefty, line 201: Okay, I would look at creating a function that takes whatever object peer/connect/listen event is.  Have that function take any lock, check for pending callbacks, cancel the overlap request, wait for the cancel, etc.  Then just call it 3 times, once for each event.  That should eliminate the nested locking and shorten this function.

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -168,15 +171,64 @@ int rdma_destroy_id(struct rdma_cm_id *id)
...
+					if (entry_nd->event.param.conn
+						    .private_data) {
+						free((void *)entry_nd->event
+							     .param.conn
+							     .private_data);
+					}
+					free(&entry_nd->event);
+				}
+			}
+		}
```
shefty, line 229: I would move the contents of this if statement into its own function as some sort of event queue cleanup routine.  That would avoid such deep indentation.

shefty, line 229: I would move the contents of this if statement into its own function as some sort of event queue cleanup routine.  That would avoid such deep indentation.

prov/verbs/src/windows/verbs_nd_ibv.c
dshinaberry: Didn't realize that there was a flag for this. That's much tidier.

prov/verbs/src/windows/verbs_nd_ibv.c
dshinaberry: Hmm, looks like clang-format doesn't like to split a string literal.

prov/verbs/src/windows/verbs_nd_ov.c
dshinaberry: There are 2 switch statements because what happens in the first switch prepares cm_event->entry to be placed into ch_nd->q at line 341. The second switch makes a call to NotifyDisconnect and that call MUST happen after the event is inserted into the queue for the reason I tried to explain in the comment at lines 343-347.

Obviously I failed at making that clear. Do you have a suggestion for how I can make this more understandable?

shefty: Please restructure to have a single switch statement.  We're talking about a single line of real code being added to the case statements.  Also, I'm guessing nd_allocate_cm_event() can fail, in which case we should handle that failure.

As a separate comment, it seems odd that we set the peer_event state to DISCONNECTED below as part of handling a connect ack.  That's non-intuitive.  It makes more sense to set the state after we received the notification of the disconnect.  As it is, it appears to be a predicted future state, not a current one as one would expect.

dshinaberry:
I think I've addressed the crux of all your comments. I tidied up the cb_pending code by adding a condition variable and using a common function to deal with cancelling overlapped requests and waiting for the callback to complete. One hiccup was that the ofi_wait_cond wants the raw mutex, which is wrapped when ENABLE_DEBUG is set. For now I just used #if ENABLE_DEBUG to deal with it. I'm not sure what your thoughts are about this. Perhaps we want to rework ofi_wait_cond to take a ofi_mutex_t instead of pthred_mutext_t.

prov/verbs/include/windows/verbs_nd.h
```diff
@@ -145,8 +145,38 @@ struct nd_event_base {
 	OVERLAPPED ov;
 	nd_event_handler event_cb;
 	nd_error_handler error_cb;
+	ofi_mutex_t lock;
+    pthread_cond_t cond;
```
shefty, line 149: nit: indentation is off

prov/verbs/include/windows/verbs_nd.h
```diff
@@ -145,8 +145,38 @@ struct nd_event_base {
...
+	uint32_t cb_pending;
 };
 
+static inline HRESULT cancel_pending(struct nd_event_base *event,
```
shefty, line 153: This should just be a standard call in some source file, no need for static inline.  Also, please try to add a prefix to all calls/structures/globals related to this provider

prov/verbs/src/windows/verbs_nd_ibv.c
```diff
@@ -363,29 +362,9 @@ int ibv_destroy_cq(struct ibv_cq *cq)
...
-				ofi_mutex_lock(&cq_nd->lock);
-			} else {
-				break;
-			}
-		}
-	}
-	ofi_mutex_unlock(&cq_nd->lock);
+    if (cancel_pending(&cq_nd->notification, (IND2Overlapped*)cq_nd->nd2cq)) {
+        return errno;
+    }
```
shefty, line 367: indentation, code uses tabs, with a tab being 8 spaces (but as a tab character, not spaces).

prov/verbs/src/windows/verbs_nd_ibv.c
```diff
@@ -806,34 +788,68 @@ int ibv_post_send(struct ibv_qp *qp, struct ibv_send_wr *wr,
...
+					      0;
+			hr = qp_nd->nd2qp->lpVtbl->Write(
+				qp_nd->nd2qp, (void *)wr->wr_id,
+				(ND2_SGE *)wr->sg_list, wr->num_sge,
+				wr->wr.rdma.remote_addr, wr->wr.rdma.rkey,
+				flags);
+			VRB_DBG(FI_LOG_EP_DATA,
+				"IND2QueuePair::Write: hr=0x%08lx, nsge=%d, "
+				"addr=0x%016llx, raddr=0x%016llx, "
+                "rkey=0x%08lx, lkey=0x%08lx\n",
```
shefty, line 838: indentation

prov/verbs/src/windows/verbs_nd_ibv.c
```diff
@@ -806,34 +788,68 @@ int ibv_post_send(struct ibv_qp *qp, struct ibv_send_wr *wr,
...
+		case IBV_WR_RDMA_READ:
+			hr = qp_nd->nd2qp->lpVtbl->Read(
+				qp_nd->nd2qp, (void *)wr->wr_id,
+				(ND2_SGE *)wr->sg_list, wr->num_sge,
+				wr->wr.rdma.remote_addr, wr->wr.rdma.rkey,
+				flags);
+			VRB_DBG(FI_LOG_EP_DATA,
+				"IND2QueuePair::Read: hr=0x%08lx, nsge=%d, "
+				"addr=0x%016llx, raddr=0x%016llx, "
+                "rkey=0x%08lx, lkey=0x%08lx\n",
```
shefty, line 820: indentation

prov/verbs/src/windows/verbs_nd_ov.c
```diff
@@ -40,21 +40,44 @@ void CALLBACK nd_io_cb(DWORD error, DWORD bytes, LPOVERLAPPED ov)
...
+	struct nd_cm_event *cm_event;
+
+	// TODO: Find an alternative to memory allocation in
+	// asynchronous completion routines
+	cm_event = calloc(1, sizeof(*cm_event));
+	assert(cm_event);
+	cm_event->event.id = id;
+	cm_event->event.listen_id = NULL;
+	cm_event->event.event = type;
+	cm_event->event.status = 0;
```
shefty, line 75: nit: calloc() zeroes memory, so no need to init values to 0/NULL.

prov/verbs/src/windows/verbs_nd_ov.c
```diff
@@ -84,34 +106,45 @@ void nd_get_connection_data(IND2Connector *connector,
...
 				FI_LOG_EP_CTRL,
 				"Failed to allocate memory for connection data.\n");
 		}
 	} else {
-		entry_nd->event.param.conn.private_data = NULL;
+		event->event.param.conn.private_data = NULL;
 	}
 }
 
+static void insert_cm_event(struct rdma_event_channel *channel,
```
shefty, line 133: Please add nd prefix

prov/verbs/src/windows/verbs_nd_ov.c
```diff
@@ -306,85 +321,72 @@ void nd_cm_connect_ack(struct nd_event_base *base, DWORD bytes)
...
-	struct nd_event_channel *ch_nd;
 
 	VRB_TRACE(FI_LOG_FABRIC, "\n");
 
 	event = container_of(base, struct nd_cm_connect_event, base);
 
 	if (error == ND_CANCELED) {
 		return;
-	} else if ((event->state == ND_CM_DISCONNECTED) &&
+	} else if ((event->type == ND_CM_DISCONNECTED) &&
```
shefty, line 357: nit: the if () statement returns, so the else can be dropped here.  This is just a new if statement.

prov/verbs/src/windows/verbs_nd_ov.c
```diff
@@ -306,85 +321,72 @@ void nd_cm_connect_ack(struct nd_event_base *base, DWORD bytes)
...
-	entry_nd->event.listen_id = NULL;
-
-	if ((event->state == ND_CM_CONNECT) || (event->state == ND_CM_ACCEPT)) {
-		entry_nd->event.event = RDMA_CM_EVENT_REJECTED;
-		entry_nd->event.status = ECONNREFUSED;
+	switch (event->type) {
+	case ND_CM_CONNECT:
+	case ND_CM_ACCEPT:
+		cm_event =
+			nd_allocate_cm_event(event->id, RDMA_CM_EVENT_REJECTED);
```
shefty, line 367: nit: This would look better splitting the line after "id," rather than after the "=", similar to how you wrote it below.

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -136,23 +136,51 @@ int rdma_create_id(struct rdma_event_channel *channel, struct rdma_cm_id **id,
...
 	return 0;
 }
 
+static void remove_cm_id_events_from_channel(struct rdma_cm_id *id)
+{
+	struct nd_event_channel *ch_nd =
+		container_of(id->channel, struct nd_event_channel, channel);
+	if (!dlistfd_empty(&ch_nd->q)) {
+		struct nd_cm_event *entry_nd;
+		struct dlist_entry *tmp;
```
shefty, line 168: Please declare variables at the top of functions.  It makes them easier to find.  This function would be easier to read as:
```
dlist_foreach_container_safe() {
    ...
}
```
no dlistfd_empty() check.  The foreach should check for an empty queue.

prov/verbs/src/windows/verbs_nd_rdma.c
```diff
@@ -136,23 +136,51 @@ int rdma_create_id(struct rdma_event_channel *channel, struct rdma_cm_id **id,
...
+		dlist_foreach_container_safe(&ch_nd->q.list, struct nd_cm_event,
+					     entry_nd, entry, tmp)
+		{
+			if (entry_nd->event.id == id) {
+				dlistfd_remove(&entry_nd->entry, &ch_nd->q);
+				if (entry_nd->event.param.conn.private_data) {
```
shefty, line 174: free() accepts a null parameter, no need to check

shefty (commented):
See comments -- thanks, looks better.  Comments are nits.  I also noticed that github posted my previous comments twice.  I have no idea why it did that.

shefty:
For ofi_wait_cond(), what would make sense is to add a wrapper call in ofi_lock.h, e.g. ofi_pthread_wait_cond(), that takes the ofi_mutex_t as input.  There would be different implementations for the debug and non-debug builds there, similar to what you had.

shefty:
See #7740 for a proposed fix for the ofi_wait_cond() issue.

shefty:
Thanks - this is looking really good.

PR #7727, opened 2022-05-03
fabtests: Build verbs related tests on windows

dshinaberry:
- Adds most of the tests handled by linux to the msvc project files
- Adds build commands to the windows nmake makefile
- Add/Modify windows specific include files to provide declarations/definitions
  used by the newly built tests

### DIFF
--- a/fabtests/include/windows/osd.h
+++ b/fabtests/include/windows/osd.h
@@ -46,6 +46,7 @@ struct iovec
 #define strdup _strdup
 #define strcasecmp _stricmp
 #define strncasecmp _strnicmp
+#define alloca _alloca
 #define SHUT_RDWR SD_BOTH
 #define CLOCK_MONOTONIC	1
 
--- a/fabtests/include/windows/unistd.h
+++ b/fabtests/include/windows/unistd.h
@@ -1,3 +1,10 @@
 #pragma once
 
+#include <math.h>
+
 #define sleep(x) Sleep(x * 1000)
+
+static inline void usleep(DWORD waitTime)
+{
+    Sleep(ceil(waitTime/1000.0));
+}
--- a/fabtests/Makefile.win
+++ b/fabtests/Makefile.win
@@ -49,7 +49,7 @@ basedeps = common\hmem.c common\shared.c \
 	common\hmem_neuron.c
 
 includes = /Iinclude /Iinclude\windows /I..\include /FIft_osd.h \
-	/Iinclude\windows\getopt
+	/Iinclude\windows\getopt /Imultinode\include
 defines = /DGETOPT_STATIC
 libs = ..\$(arch)\$(config)\libfabric.lib Ws2_32.lib
 
@@ -57,7 +57,7 @@ CFLAGS = $(CFLAGS) $(includes) $(defines)
 
 #all: complex functional
 
-all: benchmarks functional unit
+all: benchmarks functional unit multinode
 
 {benchmarks}.c{$(outdir)}.exe:
 	if not exist $(@D) mkdir $(@D)
@@ -72,15 +72,24 @@ all: benchmarks functional unit
 	$(CC) /Fe$@ $** $(baseincludes) $(CFLAGS) $(libs)
 
 
-benchmarks: $(outdir)\msg_pingpong.exe $(outdir)\rdm_cntr_pingpong.exe \
-	$(outdir)\rdm_pingpong.exe $(outdir)\rdm_tagged_pingpong.exe
+benchmarks: $(outdir)\dgram_pingpong.exe $(outdir)\msg_bw.exe \
+	$(outdir)\msg_pingpong.exe $(outdir)\rdm_cntr_pingpong.exe \
+	$(outdir)\rdm_pingpong.exe $(outdir)\rdm_tagged_bw.exe \
+	$(outdir)\rdm_tagged_pingpong.exe $(outdir)\rma_bw.exe 
 
-functional: $(outdir)\cq_data.exe $(outdir)\dgram.exe $(outdir)\dgram_waitset.exe $(outdir)\msg.exe \
-	$(outdir)\msg_epoll.exe $(outdir)\msg_sockets.exe \
-	$(outdir)\poll.exe $(outdir)\rdm.exe $(outdir)\rdm_rma_event.exe $(outdir)\rdm_rma_trigger.exe \
-	$(outdir)\rdm_tagged_peek.exe $(outdir)\scalable_ep.exe $(outdir)\msg_inject.exe $(outdir)\bw.exe
+functional: $(outdir)\av_xfer.exe $(outdir)\bw.exe $(outdir)\cm_data.exe $(outdir)\cq_data.exe \
+	$(outdir)\dgram.exe $(outdir)\dgram_waitset.exe $(outdir)\msg.exe $(outdir)\msg_epoll.exe \
+	$(outdir)\msg_inject.exe $(outdir)\msg_sockets.exe $(outdir)\multi_mr.exe \
+	$(outdir)\multi_ep.exe $(outdir)\multi_recv.exe $(outdir)\poll.exe $(outdir)\rdm.exe \
+	$(outdir)\rdm_atomic.exe $(outdir)\rdm_multi_client.exe $(outdir)\rdm_rma_event.exe \
+	$(outdir)\rdm_rma_trigger.exe $(outdir)\rdm_shared_av.exe $(outdir)\rdm_tagged_peek.exe \
+	$(outdir)\recv_cancel.exe $(outdir)\scalable_ep.exe $(outdir)\shared_ctx.exe \
+	$(outdir)\unexpected_msg.exe
 
-unit: $(outdir)\av_test.exe $(outdir)\dom_test.exe $(outdir)\eq_test.exe
+unit: $(outdir)\av_test.exe $(outdir)\cntr_test.exe $(outdir)\cq_test.exe $(outdir)\dom_test.exe \
+	$(outdir)\eq_test.exe $(outdir)\getinfo_test.exe $(outdir)\mr_test.exe
+
+multinode: $(outdir)\multinode.exe $(outdir)\multinode_coll.exe
 
 complex: $(outdir)\complex.exe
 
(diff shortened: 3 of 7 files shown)


fabtests/benchmarks/dgram_pingpong.c
```diff
@@ -100,7 +100,7 @@ int main(int argc, char **argv)
 			if (!ft_parse_long_opts(op, optarg))
 				continue;
 			ft_parse_benchmark_opts(op, optarg);
-			ft_parseinfo(op, optarg, hints, &opts);
+			ft_parseinfo(op, optarg, &hints, &opts);
```
shefty, line 103: There's a lot of changes in this single patch.  This change should be in a separate patch.  Why was this made to pass in the address of a pointer, rather than the pointer?  That's non-intuitive and suggests the function has a possible side effect.

fabtests/common/shared.c
```diff
@@ -747,6 +747,12 @@ int ft_init(void)
 	tx_cq_cntr = 0;
 	rx_cq_cntr = 0;
 
+	int ret = ft_startup();
```
shefty, line 750: Please declare variables at the top of the function and not inline with calls.  That makes them hard to find.

fabtests/common/shared.c
```diff
@@ -790,7 +796,7 @@ int ft_init_oob(void)
 			return ret;
 		}
 
-		close(listen_sock);
+		ft_close_fd(listen_sock);
```
shefty, line 799: All of these changes can be in a separate patch.

fabtests/common/shared.c
```diff
@@ -2983,9 +2994,31 @@ void ft_longopts_usage()
...
+
+	if (*ptr) {
+		free(*ptr);
+		*ptr = NULL;
+	}
+
+	fi_freeinfo(hints);
+
+	return consolidated_hints;
+}
```
shefty, line 3015: This flow is odd.  It would be easier to follow if the apps simply freed the memory that it attaches to the hints, versus calling libfabric to allocate and copy the structure, freeing it here, and returning the copy back.  Nothing in this call path suggests that this will happen, making this fragile.  If hints weren't allocated earlier using fi_allocinfo() or if the app stored a pointer to the original hints, this flow breaks.

fabtests/common/shared.c
```diff
@@ -3458,13 +3494,13 @@ int ft_sock_send(int fd, void *msg, size_t len)
 	ssize_t ret, err = 0;
 
 	for (sent = 0; sent < len; ) {
-		ret = send(fd, ((char *) msg) + sent, len - sent, 0);
+		ret = ofi_send_socket(fd, ((char *) msg) + sent, len - sent, 0);
```
shefty, line 3497: I would create a single patch to update all socket calls to being portable.

fabtests/common/windows/getopt.c
```diff
@@ -245,7 +245,7 @@ static opt_type check_for_long_option(const struct option *longopts, int *longin
 		*arg = str + i + 1; /* set potential option name substring length */
 
 	/* detecting the nearest option name to string */
-	int index = 1;
+	int index = 0;
```
shefty, line 248: This looks like a fix that should be a separate patch.

fabtests/functional/shared_ctx.c
```diff
@@ -67,20 +67,22 @@ static int get_dupinfo(void)
...
+	 * is used only for non-MSG endpoints. */
 
 	hints_dup = fi_dupinfo(hints);
 	if (!hints_dup)
 		return -FI_ENOMEM;
 
-	free(hints_dup->src_addr);
-	free(hints_dup->dest_addr);
+	src_addr = hints_dup->src_addr;
+	dest_addr = hints_dup->dest_addr;
```
shefty, line 85: This change should be moved to a separate patch, so that the reason for the change can be captured.  Should also add a  comment here referencing the problem (pointing to the more descriptive comment about the issue).

fabtests/include/windows/osd.h
```diff
@@ -46,6 +46,7 @@ struct iovec
 #define strdup _strdup
 #define strcasecmp _stricmp
 #define strncasecmp _strnicmp
+#define alloca _alloca
```
shefty, line 49: We should look at creating symbolic links from libfabric/fabtests/include/windows to libfabric/include/windows, so we only have a single copy of these portability files.  It's probably not quite that easy, as we'd also need the implementation source files linked in some cases.

shefty (changes requested):

fabtests/common/shared.c
dshinaberry: If only it were easy and straightforward to implement that.

Several places in shared.c will modify the global hints structure and the apps won't necessarily know about it, which makes it difficult for an app to know what to free before calling fi_freeinfo or ft_free_res, which will call fi_freeinfo for the app. Then there is the oddness of hints being a global, but then not treated that way by being used as a parameter to lots of functions. And all those functions are modifying hints as a side-effect despite the fact that hints is passed by value, so I think perhaps passing &hints actually makes it more clear that side effects are happening. Several apps tinker directly with hints on their own. And some call fi_dupinfo, which then makes all the memory of the returned fi_info pointer owned by libfabric and not the app.

Given all the various and sundry ways that hints was being used, it seemed like perhaps the most straightforward way of dealing with it was acknowledging that the complicating factor was the app adding its own pointer to part of the hints and making this an up front detail. If you have a hints fi_info pointer and you want to add an app owned pointer to it, add the pointer, then call ft_consolidate_hints with that pointer and it will return a new pointer that is safe to call fi_freeinfo with as well as freeing the app owned pointer.

> Nothing in this call path suggests that this will happen, making this fragile.

I would submit that a function taking a fi_info* and returning an fi_info* is pretty suggestive that perhaps an old for new swap is happening.
> If hints weren't allocated earlier using fi_allocinfo() ..., this flow breaks.

Does anyone really handcraft a fi_info hints structure and not just start with a call fi_allocinfo()? All the tests I saw had a line like `hints = fi_allocinfo()` near the beginning of main(). But your core point is definitely correct that if an app did create all its own bits of an fi_info struct, it wouldn't be correct to then pass that to fi_freeinfo.
> ...or if the app stored a pointer to the original hints, this flow breaks.

As I mentioned above, my idea was to sort of change the paradigm so that apps that needed to add app owned pieces to the hints would always add them then immediately swap that for a new structure that was safe to pass to fi_freeinfo and free the app owned data. So, under the new paradigm, it wouldn't be valid for an app to store the pointer to the original hints.

I hope none of that came off as being too defensive. I was trying hard to obtain the desired result and give windows what it needed with minimal code impact, which, I'll admit, isn't always the best way to good code. Sometimes I think that windows is the reason we can't have nice things. ;-)

I'm certainly open to any other ideas that you might have.

dshinaberry:
I'll split this into several different PRs that are more focused as you suggested.

fabtests/common/shared.c
shefty: I was thinking of something like this:
```
/* We need to free any data that we allocated before freeing the
 * hints.  Windows doesn't like it when a library frees memory that
 * was allocated by the application.
 */
void ft_freehints(struct fi_info *hints) 
{
    if (!hints)
        return;

    if (hints->src_addr) {
        free(hints->src_addr);
        hints->src_addr = NULL;
        hints->src_addrlen = 0;
    }
    if (hints->dest_addr) {
        ...
    }
    fi_freeinfo(hints);
}
```
Then update all locations where fi_freeinfo(hints) is called to call this instead.  The fi_allocinfo() call allocates the fi_info structure and all attribute substructures.  There are only a couple places where the hints would point to application allocated memory.  The src/dest addresses are two.  The auth_key might be the only other place.  Everything else is allocated by libfabric.

To dup hints, I would follow a similar flow.  I.e. create an ft_duphints(struct fi_info *hints) and have that perform the duplication.  This could be used in the shared_ctx test where we dup the hints to better isolate what's going on.

As a side note, fabtests does a poor job with this, but we could use const to indicate what parameters are not supposed to change.

dshinaberry:
I repurposed this PR to just be the addition of tests to windows msvc/nmake project files. I can close this and open a new PR if you'd rather.

fabtests/include/windows/unistd.h
```diff
@@ -1,3 +1,16 @@
...
+	LARGE_INTEGER perfCnt, start, now;
+
+	QueryPerformanceFrequency(&perfCnt);
+	QueryPerformanceCounter(&start);
+
+	do {
+		QueryPerformanceCounter((LARGE_INTEGER*)&now);
+	} while ((now.QuadPart - start.QuadPart) /
+		(float)perfCnt.QuadPart * 1000 * 1000 < waitTime);
+}
```
shefty, line 16: This isn't really sleeping, so may not relinquish the cpu to the other thread for processing.  Can we just divide the time by 1000, rounding up, and call Sleep()?  We don't really care if usleep takes a full ms, since we can't really control that.

shefty:
Note that the CI builds failed for window.

fabtests/include/windows/unistd.h
dshinaberry: Your observation is correct, but I pulled this directly from include/windows/unistd.h assuming that what was good for libfabric was good for fabtests.

Would you like a PR to change this in the libfabric include as well?

dshinaberry:
I think the CI failure was just due to the fact that the other PRs hadn't been merged into main. We'll see if the CI works as expected this time around.

fabtests/include/windows/unistd.h
shefty: Yes, please, fixing libfabric would be great.

dshinaberry:
Looks like the windows CI passed this time. I assume the linux CI build failure is due to the issues that have been ongoing.

fabtests/include/windows/unistd.h
dshinaberry: I'll create a PR for the same code in libfabric once you've approved the changes here.

shefty:
Intel CI failure is unrelated:
```
- name:   fi_av_xfer -e rdm -p "udp"
07:50:05    timestamp: 20220517-144958+0000
07:50:05    result: Fail
07:50:05    time:   301
07:50:05    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-7727/4/reg/bin/fi_av_xfer -e rdm -p "udp"   -s ci7-eth2
07:50:05    server_stdout: |
07:50:05      AV address removal: FAIL
07:50:05      bind: Address already in use
```

shefty:
thanks!

PR #7733, opened 2022-05-04
fabtests/sockets: Make socket calls portable

dshinaberry:
- Add calls to ft_startup where needed to ensure WinSock is started on windows
- Change close calls to portable ft_close_fd calls
- Change send/recv calls to portable ofi_[send/recv]_socket calls
- Use ofi_sockerr instead of errno to obtain failure codes from socket calls
- Add necessary OS dependent functions to osd.h to achieve above items


shefty:
CI failures are unrelated

PR #7738, opened 2022-05-05
contrib/intel/jenkins: disable failing tests [verbs osu, psm3 ubertest]

zachdworkin:
Verbs osu tests are hanging and causing seg faults. Suspected issue is mlnx 5.5 needs to upgrade to mlnx 5.6 for the RDMA core.
psm3 ubertest is failing also due to the suspected 5.5->5.6 issue.


shefty:
Thank you!

PR #7745, opened 2022-05-09
prov/tcp: read cm error data when receiving connreq response

ooststep:


prov/rxm/src/rxm_conn.c
```diff
@@ -561,8 +561,12 @@ rxm_process_reject(struct rxm_conn *conn, struct fi_eq_err_entry *entry)
 		break;
 	case RXM_CM_CONNECTING:
 		rxm_close_conn(conn);
-		if (reason != RXM_REJECT_EALREADY)
+		if (reason != RXM_REJECT_EALREADY) {
 			rxm_free_conn(conn);
+		}
+		else {
+			FI_INFO(&rxm_prov, FI_LOG_EP_CTRL, "rejected, already connected\n");
+		}
```
shefty, line 569: Please fix formatting.  In this case, braces are not needed.  If they were, they should be on the same line as the else.

prov/tcp/src/tcpx_conn_mgr.c
```diff
@@ -325,6 +325,16 @@ static void tcpx_cm_recv_resp(struct util_wait *wait,
 		if (ret == -FI_EAGAIN)
 			return;
 
+		err_entry.err = -ret;
+		if(cm_ctx->cm_data_sz){
```
shefty, line 329: add space after if and before {

prov/tcp/src/tcpx_conn_mgr.c
```diff
@@ -325,6 +325,16 @@ static void tcpx_cm_recv_resp(struct util_wait *wait,
...
+			err_entry.err_data = calloc(1, cm_ctx->cm_data_sz);
+			if (!err_entry.err_data) {
+				goto err1;
+			}
+			memcpy(err_entry.err_data, cm_ctx->msg.data, cm_ctx->cm_data_sz);
```
shefty, line 334: Can use mem_dup() here instead of calloc + memcpy

prov/tcp/src/tcpx_ep.c
```diff
@@ -285,10 +285,25 @@ static void tcpx_ep_flush_all_queues(struct tcpx_ep *ep)
...
+/*
+ * when sending data with UTIL_FLAG_ERROR, the util eq
+ * will store the err_entry->err_data and free it prior
+ * to the fi_eq_readerr after this data has been read.
+ *
+ * callers of tcpx_ep_disable_with_data should not
+ * free err_entry->err_data
+ */
+void tcpx_ep_disable_with_data(struct tcpx_ep *ep,
+			struct fi_eq_err_entry *err_entry)
```
shefty, line 303: This call ends up being difficult to maintain because of unexpected behavior..  We pass in struct fi_eq_err_entry, but only expect it to be partially initialized, with the call finishing the initialization.  What's expected to be set and what's modified is not conveyed anywhere.  It would be cleaner to modify tcpx_ep_disable() directly to be:
```
void tcpx_ep_disable(struct tcpx_ep *ep, int cm_err, void *err_data, size_t size)
```
Where no err_data is available, the caller can pass in NULL,  tcpx_ep_disable() can duplication err_data as needed.

shefty (changes requested):

prov/tcp/src/tcpx_ep.c
```diff
@@ -284,7 +284,7 @@ static void tcpx_ep_flush_all_queues(struct tcpx_ep *ep)
 	ofi_bsock_discard(&ep->bsock);
 }
 
-void tcpx_ep_disable(struct tcpx_ep *ep, int cm_err)
+void tcpx_ep_disable(struct tcpx_ep *ep, int cm_err, void* err_data, size_t err_data_size)
```
shefty, line 287: Check line length here and in header file

shefty: I base splitting on the line focused on keeping it readable.  For example, I might do:
```
static struct ofi_some_structure *
foo(struct blah *blahblah, int param2, size_t param3, ...)
```
if the function type is long.  But if the function type is short (void, int, etc.) I'd wrap on one of the params, maybe even based on what the param itself is.
```
void tcpx_ep_disable(struct tcpx_ep *ep, int cm_err,
        void *err_data, size_t err_data_size);
```
except aligned better than I can do in this markdown.  In this case, I would wrap to keep err_data and err_data_size together, since they're treated as a pair.  But honestly splitting after err_data would be fine.  It doesn't help to add a break after the void as that doesn't impact the line length enough.

shefty (approved):

shefty:
@ooststep - Please rebase this PR to pick up the latest Intel CI changes.

PR #7746, opened 2022-05-09
fabtests/windows: Free hints memory in module that allocated it

dshinaberry:
On windows, memory allocated by a module must be freed by that same module
because each module contains its own copy of CRT heap data structures. Any
application owned memory placed into the hints data structure must be freed
by the application before calling fi_freeinfo.

### DIFF
--- a/fabtests/unit/getinfo_test.c
+++ b/fabtests/unit/getinfo_test.c
@@ -780,6 +780,15 @@ static int getinfo_unit_test(char *node, char *service, uint64_t flags,
 			break;
 	}
 out:
+	/* If init was called, then there is a chance that the hints were
+	 * modified so that the application now owns some of the memory.
+	 * At the moment, only invalid_dom does this and the domain name
+	 * is the only application owned memory. Free the application owned
+	 * memory so that fi_freeinfo only frees memory that it owns. */
+	if (init) {
+		free(test_hints->domain_attr->name);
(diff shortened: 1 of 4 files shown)


fabtests/unit/getinfo_test.c
```diff
@@ -780,6 +780,15 @@ static int getinfo_unit_test(char *node, char *service, uint64_t flags,
...
 out:
+	/* If init was called, then there is a chance that the hints were
+	 * modified so that the application now owns some of the memory.
+	 * At the moment, only invalid_dom does this and the domain name
+	 * is the only application owned memory. Free the application owned
+	 * memory so that fi_freeinfo only frees memory that it owns. */
+	if (init) {
+		free(test_hints->domain_attr->name);
+		test_hints->domain_attr->name = NULL;
+	}
```
shefty, line 791: Why not just call ft_freehints() here?

dshinaberry: Because test_hints is created from a call to fi_dupinfo, which makes libfabric the owner of all of it. Then a call to init might make the further change to the domain name.
So the application needs to free the domain name and then everything else needs to be freed by libfabric, thus the call to fi_freeinfo(test_hints) at line 792.

I can refine the comment to be more explicit if you'd like.

shefty: Thanks - comment is fine.

PR #7748, opened 2022-05-09
fabtests/runfabtests.cmd: Rewrite to be more like runfabtests.sh

dshinaberry:
Brings a significant amount of the functionality of runfabtests.sh to
runfabtests.cmd in order to run tests on the verbs provider on windows
in a similar fashion to linux.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -53,10 +53,50 @@ before_test:
   - cd fabtests
   - msbuild fabtests.sln
 
-test_script:
-  - set PATH=%CD%\x64\%CONFIGURATION%;%PATH%
-  - set FI_PROVIDER=sockets
-  - scripts\runfabtests.cmd
+for:
+-
+  matrix:
+    only:
+      - image: Visual Studio 2017
+        configuration: Debug-v140
+      - image: Visual Studio 2017
+        configuration: Release-v140
+
+  test_script:
+    - scripts\runfabtests.cmd -v -v -v -p %CD%\x64\%CONFIGURATION% -t unit sockets
+
+-
+  matrix:
+    only:
+      - image: Visual Studio 2017
+        configuration: Debug-v141
+      - image: Visual Studio 2017
+        configuration: Release-v141
+
+  test_script:
+    - scripts\runfabtests.cmd -v -v -v -p %CD%\x64\%CONFIGURATION% -t functional sockets
+
+-
+  matrix:
+    only:
+      - image: Visual Studio 2019
+        configuration: Debug-v141
+      - image: Visual Studio 2019
+        configuration: Release-v141
+
+  test_script:
+    - scripts\runfabtests.cmd -v -v -v -p %CD%\x64\%CONFIGURATION% -C "-I 10" -L "-I 10" -t standard sockets
+
+-
+  matrix:
+    only:
+      - image: Visual Studio 2019
+        configuration: Debug-v142
+      - image: Visual Studio 2019
+        configuration: Release-v142
+
+  test_script:
+    - scripts\runfabtests.cmd -v -v -v -p %CD%\x64\%CONFIGURATION% -C "-I 10" -L "-I 10" -t multinode sockets
 
 skip_commits:
   files:
--- a/fabtests/functional/rdm_multi_client.c
+++ b/fabtests/functional/rdm_multi_client.c
@@ -221,13 +221,17 @@ int main(int argc, char **argv)
 
 	if (opts.dst_addr) {
 		for (i = 0; i < opts.num_connections; i++) {
-			save = fi_dupinfo(hints);
 			printf("Starting client: %d\n", i);
 			ret = run_client(i, address_reuse);
 			if (ret) {
 				FT_PRINTERR("run_client", -ret);
 				goto out;
 			}
+			// Reuse hints for each iteration without using fi_dupinfo
+			// because that would complicate memory ownership between the
+			// application and the library, which Windows doesn't like.
+			save = hints;
+			hints = NULL;
 			ft_free_res();
 			hints = save;
 		}
--- a/fabtests/functional/unexpected_msg.c
+++ b/fabtests/functional/unexpected_msg.c
@@ -50,7 +50,6 @@
 
 
 static size_t concurrent_msgs = 5;
-static size_t num_iters = 600;
 static bool send_data = false;
 
 
@@ -129,7 +128,7 @@ static int run_test_loop(void)
 	char *op_buf;
 	int i, j;
 
-	for (i = 0; i < num_iters; i++) {
+	for (i = 0; i < opts.iterations; i++) {
 		for (j = 0; j < concurrent_msgs; j++) {
 			op_buf = get_tx_buf(j);
 			if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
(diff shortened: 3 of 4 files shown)


dshinaberry:
I'm not sure how you'd like to modify .appveyor.yml to invoke the new version of runfabtests.cmd. I've been doing it without setting the FI_PROVIDER environment variable and using a command line of `runfabtests.cmd -v -v -v -p C:\Users\dshinaberry\repos\libfabric\fabtests\x86\Release-v141 -t all verbs 192.168.2.118 192.168.2.118`.

shefty:
With runfabtests.sh, we pass in the provider name as an argument.  That may be possible here, and just call the script once for each provider.  We also have an exclude file to skip tests that do not run, usually because of missing features.  We'll eventually want this to run against sockets (for backwards testing only), tcp, and maybe udp at some point.

dshinaberry:
This PR will not likely pass CI until #7727 is merged. I'll kick this again after that. Would you like me to make changes to appveyor.xml as part of this PR or as a follow on?

I didn't tackle the exclude file or complex test functionality as I felt that was just too much to take on.

I have run both the verbs and sockets providers through this script with no problems, but haven't tried tcp or udp. I can give them a go and if they pass without issues add them to the appveyor.yml file as well if you'd like.

dshinaberry:
Something I hadn't thought about until just now is whether the Windows CI machine has access to hardware for the verbs tests to actually use and pass on. Is that already in place or can it be relatively easily added from hardware you have access to? If not, I can see what my management has to say about potentially providing some of our equipment to be used by the project for CI purposes.

shefty:
We don't need an exclude file as long as all tests pass, or we can somehow skip problematic tests.  I don't have a preference on how to change the appveyor file, except that we don't break the CI.  If we can add tcp testing to appveyor, that would be a help, since that could test the rxm path even without verbs.

My group at Intel doesn't currently have any windows test systems.  I'm bringing this up in a meeting tomorrow, but any community assistance is desirable.  Windows support is a low priority for my specific group.

dshinaberry:
Well, that was a rather spectacular fail of the appveyor CI tests. It looks to me like perhaps it is due to insufficient cpu cycles getting to the VM running the tests. All of them fail after one hour at differing points in the tests. I'm guessing one hour is maybe a hard limit placed on a job by appveyor. It looks like most, but not all, of the failures are due to test timeout. I did see at least one configuration that didn't have any test failures before it was terminated.

So, not sure what you want to do with this. In my running of this new runfabtests.cmd with the sockets provider on a bare metal machine takes only around 16 minutes of wall clock to run successfully to completion.

.appveyor.yml
```diff
@@ -54,9 +54,7 @@ before_test:
   - msbuild fabtests.sln
 
 test_script:
-  - set PATH=%CD%\x64\%CONFIGURATION%;%PATH%
-  - set FI_PROVIDER=sockets
-  - scripts\runfabtests.cmd
+  - scripts\runfabtests.cmd -v -v -v -p %CD%\x64\%CONFIGURATION% -t all sockets
```
shefty, line 57: We may need to keep this as quick or short tests for CI purposes.  We can also consider changing the script to limit the number of iterations to 5, even for the long tests.

shefty:
We have 12 builds in appveyor.  We probably don't need to run all fabtests on each of those builds.  Maybe we could run CI on VS 2019 v142 debug/release only, and only build test the others.

dshinaberry:
Getting better. Update to set iteration count to 100 from the default of 1000. Speed things up significantly on my bare metal. We'll see what happens with the appveyor runs.

shefty:
Some CI failure notes: Not all tests accept the -I input parameter.  The atomic tests take a really long time.  May want to skip those entirely.  The I parameter passed to the script is not being used in some of the tests.

dshinaberry:
> Not all tests accept the -I input parameter.
Yeah, I was thinking they'd be silently ignored rather than cause a failure. Looks like all the standard and multinode tests support it, but not all the functional tests. I'll take a closer look at those and tweak as needed.

> The atomic tests take a really long time. May want to skip those entirely.
I'd like to keep all of the tests if we can, just to make sure we're getting coverage on windows of everything to catch any regressions. Maybe further reducing the I option count by another order of magnitude from 100 to 10 would be enough. It looks like the passing atomic tests took about 40 min, so cutting by a factor of 10 would imply 4 min, which could work.

> The I parameter passed to the script is not being used in some of the tests.
Not sure which ones you saw. It looks like all the client/server ones used it as intended. And the script parameter overrode any built-in value because it was later on the command line.

shefty:
I thought I saw tests running with -I 5, but I can't find those now, so maybe I imagined it.

dshinaberry:
Maybe you did see -I 5 and didn't see the -I 100 later ion the command line.

Cut the -I 100 to -I 10 and took it off the functional tests since they don't all support it. We'll see what effect it has on things.

dshinaberry:
Well @shefty, it appears that all the tinkering so far has achieved success for all but the two runs of rdm_multi_client for a debug configuration. The last thing I did for this was increase the default timeout from 2 to 3 minutes. Previously I had dropped the number of clients from 10 to 5.

I could keep tinkering with fewer clients/longer timeout to see if I can find a combination that works.

Or we could consider only running tests against release configurations, which seem to run faster anyway.

Or we could take rdm_multi_client out of the list of tests to run.

Or...?

shefty:
Can we drop rdm_multi_client from the debug tests?  Is that test set to a small number of iterations?

dshinaberry:
So, it turns out that rdm_multi_client had a problem with application/library memory alloc/free balancing that I hadn't previously discovered. That was causing the client process to hang because it was trying to create a pop up dialog box.

So, with that tweak, it appears that all tests can pass successfully now.

@shefty, would you prefer that I create a separate PR for the change to rdm_multi_client or are you willing to let it slide in here?

shefty:
I don't need a separate PR, but moving that change to its own commit is preferred, so we can cherry pick it as needed.

dshinaberry:
@shefty - So, I've seen three failures in six test runs. Twice it was rdm_rma_trigger and once it was multi_ep. Incidentally, rdm_rma_trigger was commented out of the existing runfabtests.cmd with a comment "disabled because of frequent failures". multi_ep was not one of the tests been run in the existing runfabtests.cmd.

I've been unable to reproduce these failures after running each hundreds of times on my available hardware. My suspicion is that the errors are manifesting due to more limited resources in the VM that Appveyor is instantiating for the build and test.

Would you be willing to accept this PR if I disable these two tests with a comment like "disabled because of frequent failures" and we see some minimum number, I'll let you pick the number, of successful runs of the tests in this PR? That would be more test coverage on Windows than currently present although not quite to parity with Linux.

My only other thought is to try and determine what resources are being provided to the Appveyor VM and then to spin up a similarly specified VM in our IT environment to see if we can reproduce any of these failures we see on Appveyor.

shefty:
I would be perfectly happy to have those tests disabled and merge what does work.  Adding more coverage on windows is a good thing.

dshinaberry:
@shefty - Dare I suggest we might be close to getting this PR approved and merged?

After disabling the two problem tests, I've kicked this PR a total of five times to run the CI tests. All five of those were successful with no test failures.

I've separated the two test code tweaks to their own commit apart from the commit that contains the new runfabtests.cmd and appveyor.xml configuration changes.

Is there anything else you'd like tweaked or modified?

shefty:
Nope - I think we're good here.  Next stop is to replace sockets with tcp.  :)

PR #7749, opened 2022-05-09
[v1.15.x] util/av: Handle NULL address insertion

shefty:
An MPI test failed because it called fi_av_insert() with a count
of 0 and an invalid address.  Because the address was invalid, the
code assigned it an address length of 0.  This in turn resulted in
the AV setting its internal addrlen to 0, since 0 was smaller than
the current address.

To fix this issue and make the code more robust, define a new
internal AV flag.  When set, it indicates that the addrlen needs
to be determined based on the first address insertion.  After
setting the addrlen, clear the flag to prevent further updates.
As part of the change, we enhance the checks and fail addresses
that are an incorrect length.

### DIFF
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -770,6 +770,8 @@ struct util_av {
 	ofi_mutex_t		ep_list_lock;
 };
 
+#define OFI_AV_DYN_ADDRLEN (1 << 0)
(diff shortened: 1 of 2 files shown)


shefty:
bot:aws:retest

PR #7751, opened 2022-05-11
contrib/intel/jenkins: re-enable tests and change labels

zachdworkin:
Re-enable osu+verbs and psm3 multinode test.
Change eth label to cvl label in Jenkinsfile


zachdworkin:
@shefty I aborted and re-ran this PR.

I changed the eth to cvl in the replay and removed the py_scripts folder so it runs the python script changes that are in my PR.

PR #7752, opened 2022-05-11
fabtests: Enhancements to fabtests and rdm_stress specifically

shefty:
- Add new input control files for rdm_stress
- Add general fabtests timeout command line option, used for rpc timeout
- Update rxm error logging to reduce noise, use sparse logging
- Add debug assert option to convert asserts into busy loops (to attach a debugger)
- Modify rdm_stress to remove AV entries for clients that are not reachable

### DIFF
--- a/fabtests/include/shared.h
+++ b/fabtests/include/shared.h
@@ -147,11 +147,6 @@ enum op_state {
 	OP_PENDING
 };
 
-enum {
-	LONG_OPT_PIN_CORE = 1,
-	LONG_OPT_TIMEOUT,
-};
-
 struct ft_context {
 	char *buf;
 	void *desc;
@@ -300,9 +295,6 @@ extern char default_port[8];
 #define FT_RX_MR_KEY 0xFFFF
 #define FT_MSG_MR_ACCESS (FI_SEND | FI_RECV)
 #define FT_RMA_MR_ACCESS (FI_READ | FI_WRITE | FI_REMOTE_READ | FI_REMOTE_WRITE)
-
-extern int lopt_idx;
-extern struct option long_opts[];
 
 int ft_getsrcaddr(char *node, char *service, struct fi_info *hints);
 int ft_read_addr_opts(char **node, char **service, struct fi_info *hints,
@@ -567,8 +559,32 @@ const char *ft_util_name(const char *str, size_t *len);
 const char *ft_core_name(const char *str, size_t *len);
 char **ft_split_and_alloc(const char *s, const char *delim, size_t *count);
 void ft_free_string_array(char **s);
+
+
+enum {
+	LONG_OPT_PIN_CORE = 1,
+	LONG_OPT_TIMEOUT,
+	LONG_OPT_DEBUG_ASSERT,
+};
+
+extern int debug_assert;
+
+extern int lopt_idx;
+extern struct option long_opts[];
 int ft_parse_long_opts(int op, char *optarg);
 void ft_longopts_usage();
+
+#define ft_assert(expr)					\
+	do {						\
+		if (!debug_assert) {			\
+			assert(expr);			\
+		} else {				\
+			if (!(expr))			\
+				FT_WARN("assert (pid %d)", getpid()); \
+			while (!(expr))			\
+				;			\
+		}					\
+	} while (0)
 
 #define FT_PROCESS_QUEUE_ERR(readerr, rd, queue, fn, str)	\
 	do {							\
--- a/fabtests/include/shared.h
+++ b/fabtests/include/shared.h
@@ -149,6 +149,7 @@ enum op_state {
 
 enum {
 	LONG_OPT_PIN_CORE = 1,
+	LONG_OPT_TIMEOUT,
 };
 
 struct ft_context {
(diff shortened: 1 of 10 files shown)


shefty:
Added fix to rxm for #7752

chien-intel:
Repeatedly running with -C 10 produced this hang:

(1-28) rpc op write_req
(1-28) rpc op sleep
(1-28) rpc op exit
(2-33) rpc op write_req
(2-33) rpc op sleep
(2-33) rpc op exit
[ctung@d20 ~]$ fi_rdm_stress -u /home/ctung/libfabric/fabtests/test_configs/ofi_rxm/stress.json  -C 10 192.168.5.19
Starting rpc client(s)
(0-?) running
(1-?) running
(2-?) running
(3-?) running
(4-?) running
(5-?) running
(6-?) running
(7-?) running
(8-?) running
(9-?) running
(0-?) saying hello
(5-?) saying hello
(7-?) saying hello
(1-?) saying hello
(9-?) saying hello
(3-?) saying hello
(6-?) saying hello
(2-?) saying hello
(8-?) saying hello
(4-?) saying hello

chien-intel:
At the end of stress.json, write_resp is removed.  This is causing server side to timeout (takes a bit of time) but eventually it does continue and respond to connection requests.

fabtests/include/shared.h
```diff
@@ -566,9 +559,33 @@ const char *ft_util_name(const char *str, size_t *len);
...
 int ft_parse_long_opts(int op, char *optarg);
 void ft_longopts_usage();
 
+#define ft_assert(expr)					\
+	do {						\
+		if (!debug_assert) {			\
+			assert(expr);			\
+		} else {				\
+			if (!(expr))			\
+				FT_WARN("assert (pid %d)", getpid()); \
```
chien-intel, line 583: This is an indication of incorrect usage of assert.  It should of been FT_WARN instead.

prov/rxm/src/rxm_cq.c
```diff
@@ -1721,9 +1721,14 @@ void rxm_handle_comp_error(struct rxm_ep *rxm_ep)
 		return;
 	}
 
-	if (err_entry.err != FI_ECANCELED)
-		OFI_CQ_STRERROR(&rxm_prov, FI_LOG_WARN, FI_LOG_CQ,
-				rxm_ep->msg_cq, &err_entry);
+	if (err_entry.err != FI_ECANCELED) {
+		FI_WARN(&rxm_prov, FI_LOG_CQ,
```
chien-intel, line 1725: FI_WARN is one statement, no brace around it.

include/ofi_mem.h
```diff
@@ -490,7 +490,7 @@ static inline void *ofi_ibuf_alloc(struct ofi_bufpool *pool)
 	struct ofi_bufpool_region *buf_region;
 
 	assert(pool->attr.flags & OFI_BUFPOOL_INDEXED);
-	if (OFI_UNLIKELY(ofi_ibufpool_empty(pool))) {
+	if (ofi_ibufpool_empty(pool)) {
```
chien-intel, line 493: how about:
if (ofi_ibufpool_empty(pool) && (ofi_bufpool_grow(pool))
    return NULL;

chien-intel:
I changed my copy of rxm_stress.c to use rpc_timeout of 1 second and the server side held up well.

shefty:
The rdm_stress test will 'hang' only for the timeout duration.  By default, in debug mode that's about 5 minutes.  A shorter timeout makes it difficult to execute in a debugger.  One of the changes in this series adds a timeout parameter to control it from the command line.  One of the purposes of the test is to handle the case where the client exits while there are outstanding RPCs against it.  The server is expected to retry sending responses in case the client comes back within the timeout time.

fabtests/include/shared.h
shefty: Please read the commit message for the purpose of this change.  The behavior of an assert is desired.

prov/rxm/src/rxm_cq.c
shefty: Same as above.

fabtests/include/shared.h
chien-intel: Got it.  It didn't register with me the first time I read the commit message.

fabtests/test_configs/ofi_rxm/stress.json
```diff
@@ -23,33 +23,33 @@ This is a sample control file for fi_rdm_stress
...
+	{ "op" : "read_resp" },
+	{ "op" : "write_req", "size" : 86000, "offset" : 6000 },
+	{ "op" : "write_resp" },
+	{ "op" : "read_req", "size" : 1000000 },
+	{ "op" : "read_resp" },
+	{ "op" : "write_req", "size" : 56000, "offset" : 12000 },
+	{ "op" : "write_resp" },
+	{ "op" : "sleep", "ms" : 100 },
+	{ "op" : "write_req", "size" : 10000000 },
+	{ "op" : "sleep", "ms" : 1 },
```
chien-intel, line 53: Since you created the other two files, it is best to add back write_resp here so server side won't time out.
if you want to keep it as is, add a comment here so the user can know what to expect.  I tested with this and it works (even though it is no valid json):

        { "op" : "sleep", "ms" : 100 },
# Next write_req will casue a timeout on the server.  Server will continue after timeout.
        { "op" : "write_req", "size" : 10000000 },

chien-intel: stupid markdown.  the comment is prefixed with "#" character.

shefty:
Update rdm_stress to always use default timeout of 2 seconds, to avoid reports that test hangs when running debug build.

Added error message if file name is incorrect.

PR #7754, opened 2022-05-11
contrib/intel/jenkins: Interface Selection, tcp-osu-impi enabled, code cleanup

zachdworkin:
All tests that need to select an interface (every provider except shm) now explicitly select their interface.
This will allow us to know exactly what tests are doing when they run.

common.py has been refactored to follow other py files' style.

OSU tcp tests have been re-enabled for impi since they are now passing.

The general execute condition is being removed because it created a redundant check.

As usual - the ci will be aborted and replayed removing the py_scripts file path from the Jenkinsfile to ensure the changed scripts are being used.

### DIFF
--- a/contrib/intel/jenkins/common.py
+++ b/contrib/intel/jenkins/common.py
@@ -4,8 +4,7 @@ import subprocess
 import sys
 
 def get_node_name(host, interface):
-   # This is the pattern we follow in SFS team cluster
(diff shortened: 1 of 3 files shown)


zachdworkin:
@shefty once the appveyor passes you can merge this.
The Jenkins CI logs all look correct

PR #7759, opened 2022-05-13
core/indexer: Code cleanups

shefty:
Variable renames to make the code more intuitive to understand.
Added a few asserts for verifications

Should be no functionality changes


shefty:
Broke patches up into even more patches to make it easier to verify the changes.

PR #7760, opened 2022-05-13
fabtests/dgram_pingpong.c: use out-of-band sync

wzamazon:
This patch makes dgram_pingpong to always use out-of-band
sync method, because dgram endpoing is unreliable thus cannot
be used for sync.

### DIFF
--- a/fabtests/benchmarks/dgram_pingpong.c
+++ b/fabtests/benchmarks/dgram_pingpong.c
@@ -117,6 +117,12 @@ int main(int argc, char **argv)
 	if (optind < argc)
 		opts.dst_addr = argv[optind];
 
+	/*
+	 * Because dgram endpoint is not reliable, we
+	 * must use out-of-band sync
+	 */
+	opts.options |= FT_OPT_OOB_SYNC;
+
 	hints->ep_attr->type = FI_EP_DGRAM;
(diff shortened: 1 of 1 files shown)


shefty (approved):
I'm fine with the change, but it doesn't help that the pingpong messages can be lost, which will still hang the test

wzamazon:
> I'm fine with the change, but it doesn't help that the pingpong messages can be lost, which will still hang the test.

That is correct. This PR is try to mitigate the issue.

shijin-aws:
@shefty do you think we should make `dgram_pingpong` return 0 as long as the receiver can receive "some" messages? The current implementation will return `FI_ENODATA` when the receiver could not poll the completion within 5 seconds for any message. It may cause failure when one run fabtests in strict mode. But dgram is not reliable by nature and such randomness is expected?

shefty:
I think it's fine to adjust the test to return success even if we lose messages.  We probably should make the test retry sends if it doesn't get a response.  That could potentially make the test take a long time if we lost a lot of messages, so we could cap the total number of retries to ensure the test finishes within a few seconds.  For example, timeout waiting after 2 seconds and resend.  Abort the test if we timeout 5 times total.

PR #7765, opened 2022-05-14
prov/psm3: avoid bashism in configure

hzhou:
Substring substitution is bash extension. Use "remove smallest prefix
pattern" substitution instead.

Reference: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_03


acgoldma (approved):

shefty:
Can you please rebase the changes to pick up newer CI scripts?  Older scripts are the likely cause of the CI failure.

PR #7766, opened 2022-05-14
src/hmem_gdrcopy: do not cleanup on successfully initialization

wzamazon:
This patch fixed a bug in cuda_gdrcopy_hmem_init(), which
is that cuda_gdrcopy_dl_hmem_cleanup() was called even when
the initialization was successful.

### DIFF
--- a/src/hmem_cuda_gdrcopy.c
+++ b/src/hmem_cuda_gdrcopy.c
@@ -184,7 +184,7 @@ static int cuda_gdrcopy_dl_hmem_cleanup(void)
 
 int cuda_gdrcopy_hmem_init(void)
 {
-	int err, ret = 0;
+	int err;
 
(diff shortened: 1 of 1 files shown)


shefty:
Does this need to be picked into v1.15.x?

wzamazon:
> Does this need to be picked into v1.15.x?

Yes.

shefty:
Can you please pick this for v1.15.x and v1.14.x if needed?

wzamazon:
Sure. will do.

wzamazon:
> Can you please pick this for v1.15.x and v1.14.x if needed?

PRs were opened.

FWIW, this bug only affect some that is using gdrcopy with dlopen (--enable-gdrcopy-dlopen). In other cases, gdrcopy_hmem_cleanup() is a no-op.

PR #7769, opened 2022-05-16
fabtests: Add regression test for shm sighandler restoration (PR#7605)

j-xiong:
The test checks if the default signal handler is properly chained in the
handler installed by the shm provider.

### DIFF
--- a/fabtests/regression/shm_sighandler_test.c
+++ b/fabtests/regression/shm_sighandler_test.c
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2022 Intel Corporation.  All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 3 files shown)


shijin-aws:
Thank you for the PR! Do you plan to add it to `runfabtests.sh`? We can help add it (and to `runfabtests.py`) as a separate PR as well.

j-xiong:
@shijin-aws I will leave the script update to you :) Thanks!

shijin-aws (approved):

fabtests/regression/shm_sighandler_test.c
```diff
@@ -0,0 +1,70 @@
...
+			printf("pass\n");
+			exit(0);
+		}
+		printf("fail\n");
+		exit(1);	
+	} else {
+		struct fi_info *fi;
+
+		if (fi_getinfo(FI_VERSION(1,0), NULL, NULL, 0, NULL, &fi))
+			fi_freeinfo(fi);
```
shefty, line 65: Does shm replace the signal handlers when fi_getinfo is called?

shijin-aws:
@j-xiong Thanks, will do.

fabtests/regression/shm_sighandler_test.c
j-xiong: It happens in the provider initialization code (referred by the macro SHM_INI).

shefty: @aingerson - Why is smr_reg_sig_handler() called from SHM_INI?  Couldn't this be deferred until an EP is created?

j-xiong:
updated the commit message.

shefty (approved):
New test looks good, but I'd like to understand the shm behavior.

fabtests/regression/shm_sighandler_test.c
aingerson: It could be deferred to the EP. We would just need to add some code to make sure it's only called once but that's easy enough. Would you like me to do that? We'll need to change this test to match

shefty: I think it's worth changing as that would limit the impact to when only the shm provider is actually used, rather than simply being present.  The test would probably only need to add a call to ft_init_fabric().  The test changes could go in before shm was updated.  And the test could then verify that shm was present in the library as part of its check.

j-xiong:
Updated with full initialization of the shm provider.

fabtests/regression/shm_sighandler_test.c
```diff
@@ -0,0 +1,78 @@
...
+		exit(EXIT_FAILURE);
+	} else {
+		hints = fi_allocinfo();
+		if (!hints)
+			exit(EXIT_FAILURE);
+
+		hints->fabric_attr->prov_name = strdup("shm");
+		if (ft_init_fabric()) {
+			fprintf(stderr, "Failed to initialize shm provider\n");
+			fi_freeinfo(hints);
```
shefty, line 71: Use ft_freehints() here instead for consistency with other fabtests (this is a very recent update).

j-xiong: Updated.

PR #7770, opened 2022-05-17
prov/tcp: Verify ep state prior to progressing socket data

shefty:
Skip progressing data transfers unless the ep is in the
connected state.  This avoids the potential of sending data
that may be interpreted as the start of a message by the peer,
and processing received data in the middle of the data stream
as the start of a new message.

There are no identified issues in the current code requiring
this change; however, problems persist with application segfaults
when connections are abruptly terminated.


shefty:
Out of tree, I converted the if statements added by this patch into asserts, to see if they would be hit.  It turns out that the one in tcpx_progress_tx() does get hit.  That's possible if a send is posted after the ep has been disabled, for example, the peer has disconnected.  The current code looks correct, in that tcpx_progress_tx() will fail to send the message and instead generate a CQ error entry.  But with the changes here, the behavior also looks correct.  The difference is that the posted sends will be queued until the ep is closed, at which point they will be flushed.  The current behavior is slightly more desirable, as the error is reported in a more timely fashion, though a better option is to fail the send call up front.

I have not hit the asserts in the receive path, but I don't expect to.

shefty:
Added changes to rename functions in rxm.  Updated the checks in the tcp provider to remove the check from the tx path.  Updated the stress.json test file.  And added an enhancement and one fix to rdm_stress.  Fixes #7772

PR #7774, opened 2022-05-19
osd-windows/usleep: Modify usleep to sleep rather than spin

dshinaberry:
Change the implementation of the windows specific implementation of
usleep to actually sleep and guarantee and relinquishment of the
cpu rather than spin.


shefty (approved):
thank you!

PR #7776, opened 2022-05-20
prov/shm: fix hints->caps = 0 case for FI_MR_HMEM

### LINKED ISSUE #7775: fabtests/shm_sighandler_test does not have consistent behavior
I noticed that executing the shm_sighandler_test twice in a row causes inconsistent behavior.

```
[ec2-user@compute-st-c5n18xlarge-1 bin]$ ./shm_sighandler_test 
Pass: child caught SIGINT and exited as expected
[ec2-user@compute-st-c5n18xlarge-1 bin]$ ./shm_sighandler_test 
fi_getinfo(): common/shared.c:903, ret=-61 (No data available)
Failed to initialize shm provider
Fail: child killed by SIGKILL or exited with error
```

I noticed that ft_free_res() is not called, so even though it's catching SIGINT, something is causing shm to not be able to initialize again.

aingerson:
util_getinfo will return all the util caps if hints->caps is
not set which means shm will interpret this as if the user
requested all the util caps. This will cause a failure with
the MR_HMEM case if MR_HMEM is not set because shm will
think the user needs FI_HMEM which requires FI_MR_HMEM.

This adds a condition to check for no caps set in which case
it will skip the info if we got the HMEM info back but not
return ENODATA since it was not explicitly requested and might
return FI_SUCCESS for the first fi_info.

Fixes #7775

### DIFF
--- a/prov/shm/src/smr_init.c
+++ b/prov/shm/src/smr_init.c
@@ -168,6 +168,9 @@ static int smr_getinfo(uint32_t version, const char *node, const char *service,
 		if (cur->caps & FI_HMEM) {
 			if (!(mr_mode & FI_MR_HMEM)) {
 				fi_freeinfo(cur);
+				if (!(hints->caps & FI_HMEM))
+					continue;
+
 				FI_INFO(&smr_prov, FI_LOG_CORE,
 					"mr_mode does not match FI_HMEM capability.\n");
 				return -FI_ENODATA;


aingerson:
Might not be a full fix. I'll take a closer look at util_getinfo to see if there's a better way to handle this in the common code but for now, @wckzhang could you test this out to see if this is, in fact, the root cause of the issue you're seeing?

prov/shm/src/smr_init.c
```diff
@@ -168,6 +168,9 @@ static int smr_getinfo(uint32_t version, const char *node, const char *service,
 		if (cur->caps & FI_HMEM) {
 			if (!(mr_mode & FI_MR_HMEM)) {
 				fi_freeinfo(cur);
+				if (!(hints->caps & FI_HMEM))
+					continue;
+
```
shefty, line 173: I still don't think this is correct. The problem is that 1 of shm's fi_info structures may match and still need to be returned.  But this path returns a failure.  And it frees the whatever cur is, plus anything beyond cur, but if there's an fi_info before cur, that will still point to this freed structure.

As @j-xiong mentioned in the issue, you would need to pluck this entry out of the list, but still return the ones that are valid.  Or fail ever adding it to the list in the first place, which is easier.

j-xiong: Agree. This patch would cause more trouble than it fixes. Letting util_getinfo() to filter out the unwanted entries is simpler.

shefty: See #7777 for an alternate solution

PR #7780, opened 2022-05-20
prov/efa: adjust runting read protocol

wzamazon:
2 commits the adjusts the behavior of runting read protocol.


wzamazon:
bot:aws:retest

shijin-aws (approved):

PR #7787, opened 2022-05-24
Fix potential addr order change and addr array overflow in av ops.

peinanz:
Fix av_set "remove" and "diff" ops which changed the address order.
Add max_array_size in "struct util_av_set" to check addr array overflow.
Add string ending after copying addr.

### DIFF
--- a/prov/efa/src/rxr/rxr.h
+++ b/prov/efa/src/rxr/rxr.h
@@ -180,9 +180,9 @@ static inline void rxr_poison_mem_region(uint32_t *ptr, size_t size)
 
 /*
  * Flag to indicate an rx_entry has an EOR
- * in flight (the EOR has been sent, but has
- * not got send completion)
- * hence cannot be released
+ * in flight (the EOR has been sent or queued,
+ * and has not got send completion)
+ * hence the rx_entry cannot be released
  */
 #define RXR_EOR_IN_FLIGHT BIT_ULL(10)
 /*
--- a/prov/efa/src/rxr/rxr_pkt_type.h
+++ b/prov/efa/src/rxr/rxr_pkt_type.h
@@ -187,8 +187,10 @@ int rxr_pkt_init_eor(struct rxr_ep *ep,
 		     struct rxr_rx_entry *rx_entry,
 		     struct rxr_pkt_entry *pkt_entry);
 
-
-void rxr_pkt_handle_eor_sent(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry);
+static inline
+void rxr_pkt_handle_eor_sent(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
+{
+}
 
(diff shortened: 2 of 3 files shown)


prov/shm/src/smr_av.c
```diff
@@ -170,20 +170,22 @@ static int smr_av_lookup(struct fid_av *av, fi_addr_t fi_addr, void *addr,
...
 	/* the input address is a string format */
-	if (buf)
+	if (buf && *len) {
 		strncpy(buf, (const char *)addr, *len);
-
-	*len = strlen((const char *)addr) + 1;
+		*len = MIN(strlen((const char *)addr)+1, *len);
+		if (*len)
+			buf[*len - 1 ] = '\0';
+	}
```
shefty, line 188: Please move the changes to the smr provider to its own patch.  Make sure the commit title indicates the provider.  E.g. prov/shm: Add checks for string overflow

prov/util/src/util_coll.c
```diff
@@ -76,11 +76,20 @@ int ofi_av_set_union(struct fid_av_set *dst, const struct fid_av_set *src)
...
 		}
 		if (j == dst_av_set->fi_addr_count) {
-			dst_av_set->fi_addr_array[temp_count++] =
-				src_av_set->fi_addr_array[i];
+			if ((temp_count+1) >= dst_av_set->max_array_size) {
+				dst_av_set->fi_addr_count += src_av_set->fi_addr_count;
+				dst_av_set->fi_addr_array = (fi_addr_t *)realloc(
+							dst_av_set->fi_addr_array,
+							dst_av_set->fi_addr_count * sizeof(fi_addr_t));
+				if (!dst_av_set->fi_addr_array)
```
shefty, line 84: The way to handle realloc calls is to save the results to another variable.  That way if the call fails, you don't lose the existing array completely.

prov/util/src/util_coll.c
```diff
@@ -117,25 +129,26 @@ int ofi_av_set_diff(struct fid_av_set *dst, const struct fid_av_set *src)
...
-	temp = dst_av_set->fi_addr_count;
-	for (i = 0; i < src_av_set->fi_addr_count; i++) {
-		for (j = 0; j < temp; j++) {
-			if (dst_av_set->fi_addr_array[j] ==
-			    src_av_set->fi_addr_array[i]) {
-				dst_av_set->fi_addr_array[--temp] =
-					dst_av_set->fi_addr_array[j];
+	for (i = 0; i < dst_av_set->fi_addr_count; i++) {
+		for (j = 0; j < src_av_set->fi_addr_count; j++) {
+			if (dst_av_set->fi_addr_array[i] == src_av_set->fi_addr_array[j])
```
shefty, line 141: Please move fixing the set ordering to its own patch.

shefty (changes requested):

PR #7789, opened 2022-05-25
prov/util: Add max_array_size to track/check array overflow

peinanz:
Add max_array_size in "struct util_av_set" for tracking address array size, and add checkes on array overflow.

### DIFF
--- a/prov/util/src/util_coll.c
+++ b/prov/util/src/util_coll.c
@@ -60,14 +60,12 @@ int ofi_av_set_union(struct fid_av_set *dst, const struct fid_av_set *src)
 {
 	struct util_av_set *src_av_set;
 	struct util_av_set *dst_av_set;
-	size_t temp_count;
 	int i,j;
 
 	src_av_set = container_of(src, struct util_av_set, av_set_fid);
 	dst_av_set = container_of(dst, struct util_av_set, av_set_fid);
 
 	assert(src_av_set->av == dst_av_set->av);
-	temp_count = dst_av_set->fi_addr_count;
 
 	for (i = 0; i < src_av_set->fi_addr_count; i++) {
 		for (j = 0; j < dst_av_set->fi_addr_count; j++) {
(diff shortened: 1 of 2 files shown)


prov/util/src/util_coll.c
```diff
@@ -76,11 +77,22 @@ int ofi_av_set_union(struct fid_av_set *dst, const struct fid_av_set *src)
 				break;
 		}
 		if (j == dst_av_set->fi_addr_count) {
-			dst_av_set->fi_addr_array[temp_count++] =
-				src_av_set->fi_addr_array[i];
+			if ((temp_count+1) >= dst_av_set->max_array_size) {
+				dst_av_set->max_array_size += src_av_set->fi_addr_count;
```
shefty, line 81: The max_array_size should be large enough that we can just fail if we ever try to go beyond the value, versus bothering to try growing the array.

But if we do bother trying to grow the array, we can't update max_array_size until after we know that the realloc has succeeded.

It would also be better to create a function to grow the array, so we can isolate that functionality, rather than duplicating it twice (3 times if you include the initial allocation).

prov/util/src/util_coll.c
```diff
@@ -150,7 +154,14 @@ int ofi_av_set_insert(struct fid_av_set *set, fi_addr_t addr)
...
 			return -FI_EINVAL;
 	}
+	if (av_set->fi_addr_count >= av_set->max_array_size)  {
+		FI_INFO(av_set->av->prov, FI_LOG_AV,
+            "av_set_insert exceeds av addr array size\n");
+		return -FI_ENOMEM;
+    }
 	av_set->fi_addr_array[av_set->fi_addr_count++] = addr;
+	FI_DBG(av_set->av->prov, FI_LOG_AV,
+            "av_set_insert fi_addr: %" PRIu64 "\n", addr);
```
shefty, line 164: Some notes: Indentation is off in the FI_INFO statements above.  max_array_size is never initialized.  And the check should probably be (fi_addr_count >= max_array_size - 1), depending on what max_array_size is set to.

peinanz: max_array_size is set to allocated array size.  The fi_addr_count is the available # addr in the array, so when array is full, it will have "fi_addr_count == max_array_size".  Do we need to leave an ending space here?

shefty:
Picking manually with updates to fix failures and will merge via PR #7845

PR #7791, opened 2022-05-26
prov/verbs: Fix incorrect length used in memory registration

j-xiong:
The buffer length was incorrectly overwritten with the allocation length
returned from ze_hmem_get_base_addr(). This would cause registration failure
if the buffer has non-zero offset into the allocation because the address
range of registration would exceed the allocation.

Fix by ignoring the allocation length.

### DIFF
--- a/prov/verbs/src/verbs_mr.c
+++ b/prov/verbs/src/verbs_mr.c
@@ -78,7 +78,7 @@ struct ibv_mr *vrb_mr_ibv_reg_dmabuf_mr(struct ibv_pd *pd, const void *buf,
 	if (err)
 		return NULL;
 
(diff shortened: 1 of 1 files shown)


j-xiong:
LANL CI error:
```
../../prov/verbs/src/fi_verbs.c:39:10: fatal error: verbs_osd.h: No such file or directory
 #include "verbs_osd.h"
          ^~~~~~~~~~~~~
compilation terminated.
```

Don't know what went wrong with the CI setting, but the header file should be under prov/verbs/include/linux which is added to CPPFLAGS in Makefile.include.

shefty (approved):

j-xiong:
bot:lanl:retest

j-xiong:
The LANL CI error is not related. Ignoring it.

PR #7792, opened 2022-05-26
Fabtests/Multinode: Performance

alex-mckinley:
Add options to run only one pattern for the multinode tests

Show performance data for connection management

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -372,6 +372,23 @@ pipeline {
             }
           }
         }
+        stage('multinode-tcp-performance') {
+          agent {node {label 'cvl'}}
+          options { skipDefaultCheckout() }
+          steps {
+              withEnv(['PATH+EXTRA=/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin/:$PYTHONPATH'])
+              {
+                sh """
+                  env
+                  (
+                      cd /opt${env.WORKSPACE}/py_scripts/contrib/intel/jenkins/
+                      python3.7 runtests.py --prov=tcp --test=multinode
+                      echo "multinode performance completed."
+                  )
+                """
+              }
+          }
+        }
         stage('oneCCL') {
           agent {node {label 'cvl'}}
           options { skipDefaultCheckout() }
--- a/fabtests/include/shared.h
+++ b/fabtests/include/shared.h
@@ -125,6 +125,7 @@ enum {
 	FT_OPT_SRX		= 1 << 21,
 	FT_OPT_STX		= 1 << 22,
 	FT_OPT_SKIP_ADDR_EXCH	= 1 << 23,
+	FT_OPT_PERF	= 1 << 24,
 	FT_OPT_OOB_CTRL		= FT_OPT_OOB_SYNC | FT_OPT_OOB_ADDR_EXCH,
 };
 
--- a/fabtests/include/shared.h
+++ b/fabtests/include/shared.h
@@ -125,7 +125,7 @@ enum {
 	FT_OPT_SRX		= 1 << 21,
 	FT_OPT_STX		= 1 << 22,
 	FT_OPT_SKIP_ADDR_EXCH	= 1 << 23,
-	FT_OPT_PERF	= 1 << 24,
+	FT_OPT_PERF		= 1 << 24,
 	FT_OPT_OOB_CTRL		= FT_OPT_OOB_SYNC | FT_OPT_OOB_ADDR_EXCH,
 };
 
--- a/fabtests/multinode/src/core.c
+++ b/fabtests/multinode/src/core.c
@@ -45,9 +45,10 @@
 #include <rdma/fi_cm.h>
 #include <rdma/fi_trigger.h>
 
-#include <core.h>
-#include <pattern.h>
 #include <shared.h>
+#include "core.h"
+#include "pattern.h"
+#include "timing.h"
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
@@ -58,6 +59,8 @@ char *tx_barrier;
 char *rx_barrier;
 struct fid_mr *mr_barrier;
 struct fi_context2 *barrier_tx_ctx, *barrier_rx_ctx;
+
+struct multi_timer *timers;
 
 struct pattern_ops *pattern;
 struct multinode_xfer_state state;
@@ -160,8 +163,8 @@ static int multi_setup_fabric(int argc, char **argv)
 	}
 
 	for (i = 0; i < pm_job.num_ranks; i++) {
-		ret = fi_av_insert(av, (char*)pm_job.names + i * pm_job.name_len, 1,
-			   &pm_job.fi_addrs[i], 0, NULL);
+		ret = fi_av_insert(av, (char *)pm_job.names + i * pm_job.name_len, 
+				   1, &pm_job.fi_addrs[i], 0, NULL);
 		if (ret != 1) {
 			FT_ERR("unable to insert all addresses into AV table\n");
 			ret = -1;
@@ -211,7 +214,7 @@ int multi_msg_recv()
 		if (ret < 0)
 			return ret;
 
-		offset = state.recvs_posted % opts.window_size ;
+		offset = state.recvs_posted % opts.window_size;
 		assert(rx_ctx_arr[offset].state == OP_DONE);
 
 		ret = ft_post_rx_buf(ep, opts.transfer_size,
@@ -244,6 +247,9 @@ int multi_msg_send()
 
 		offset = state.sends_posted % opts.window_size;
 		assert(tx_ctx_arr[offset].state == OP_DONE);
+		if (ft_check_opts(FT_OPT_PERF))
+			multi_timer_start(&timers[timer_index(state.iter,
+							    state.cur_target)]);
 
 		dest = pm_job.fi_addrs[state.cur_target];
 		ret = ft_post_tx_buf(ep, dest, opts.transfer_size,
@@ -305,6 +311,10 @@ int multi_rma_write()
 		        pm_job.my_rank, state.cur_target,
 		        (size_t) tx_seq, pattern->name);
 
+		if (ft_check_opts(FT_OPT_PERF))
+			multi_timer_start(&timers[timer_index(state.iter,
+							    state.cur_target)]);
+
 		while (1) {
 			ret = fi_write(ep,
 				tx_buf + tx_size * state.cur_target,
@@ -364,8 +374,8 @@ int send_recv_barrier(int sync)
 
 	for (i = 0; i < pm_job.num_ranks; i++) {
 		ret = ft_post_rx_buf(ep, opts.transfer_size,
-			     &barrier_rx_ctx[i],
-			     rx_buf, mr_desc, 0);
+			     	     &barrier_rx_ctx[i],
+			     	     rx_buf, mr_desc, 0);
 		if (ret)
 			return ret;
 	}
@@ -402,11 +412,14 @@ static inline void multi_init_state()
 
 static int multi_run_test()
 {
-	int ret;
-	int iter;
+	int ret, i;
 
-	for (iter = 0; iter < opts.iterations; iter++) {
+	for (state.iter = 0; state.iter < opts.iterations; state.iter++) {
 		multi_init_state();
+		for (i = 0; i < pm_job.num_ranks && ft_check_opts(FT_OPT_PERF); i++)
+			multi_timer_init(&timers[timer_index(state.iter, i)], 
+					 pm_job.my_rank);
+
 		while (!state.all_completions_done ||
 				!state.all_recvs_posted ||
 				!state.all_sends_posted) {
@@ -423,15 +436,20 @@ static int multi_run_test()
 				return ret;
 		}
 
-		ret = send_recv_barrier(iter);
+		for (i = 0; i < pm_job.num_ranks && ft_check_opts(FT_OPT_PERF); i++)
+			multi_timer_stop(&timers[state.iter * pm_job.num_ranks + i]);
+
+		ret = send_recv_barrier(state.iter);
 		if (ret)
 			return ret;
+
 	}
 	return 0;
 }
 
 static void pm_job_free_res()
 {
+	free(timers);
 	free(pm_job.names);
 	free(pm_job.fi_addrs);
 	free(pm_job.multi_iovs);
@@ -447,7 +465,6 @@ int multinode_run_tests(int argc, char **argv)
 	int ret = FI_SUCCESS;
 	int i;
 
-
 	barrier_tx_ctx = malloc(sizeof(*barrier_tx_ctx) * pm_job.num_ranks);
 	if (!barrier_tx_ctx)
 		return -FI_ENOMEM;
@@ -460,18 +477,30 @@ int multinode_run_tests(int argc, char **argv)
 	if (ret)
 		return ret;
 
+	if (ft_check_opts(FT_OPT_PERF))
+		timers = calloc(opts.iterations * pm_job.num_ranks, sizeof(*timers));
+
 	for (i = 0; i < NUM_TESTS && !ret; i++) {
 		printf("starting %s... ", patterns[i].name);
 		pattern = &patterns[i];
+
 		ret = multi_run_test();
-		if (ret)
+		if (ret) {
 			printf("failed\n");
-		else
-			printf("passed\n");
+			goto out;
+		}
+		printf("passed\n");
+
+		if (ft_check_opts(FT_OPT_PERF)) {
+			ret = multi_timer_analyze(timers, opts.iterations * 
+							  pm_job.num_ranks);
+			if (ret)
+				goto out;
+		}
 
 		fflush(stdout);
 	}
-
+out:
 	pm_job_free_res();
 	ft_free_res();
 	return ft_exit_code(ret);
--- a/fabtests/multinode/src/core.c
+++ b/fabtests/multinode/src/core.c
@@ -480,10 +480,9 @@ int multinode_run_tests(int argc, char **argv)
 	if (ft_check_opts(FT_OPT_PERF))
 		timers = calloc(opts.iterations * pm_job.num_ranks, sizeof(*timers));
 
-	for (i = 0; i < NUM_TESTS && !ret; i++) {
-		printf("starting %s... ", patterns[i].name);
-		pattern = &patterns[i];
-
+	if (pm_job.pattern != -1) {
+		printf("starting %s... ", patterns[pm_job.pattern].name);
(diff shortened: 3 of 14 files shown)


fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,222 @@
...
+#include <string.h>
+#include <timing.h>
+#include <core.h>
+#include <limits.h>
+
+extern struct pm_job_info pm_job;
+
+int multi_timer_create(struct multi_timer **timer, int rank)
+{
+	*timer = malloc(sizeof(*timer));
```
shefty, line 41: Allocating 1 structure at a time is horribly inefficient.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,222 @@
...
+
+void multi_timer_stop(struct multi_timer *timer)
+{
+	struct timespec spec;
+
+	if (timer && timer->start != -1 && timer->end == -1) {
+		clock_gettime(CLOCK_REALTIME, &spec);
+		timer->end = spec.tv_nsec + spec.tv_sec * 1000000000;
+	}
+}
```
shefty, line 68: We have functions to get timestamps -- ft_gettime_ns() / us() / ms() -- which can be used instead, and call clock_gettime() correctly.  We shouldn't need either of the above start/stop functions.

fabtests/include/shared.h
```diff
@@ -125,6 +125,7 @@ enum {
 	FT_OPT_SRX		= 1 << 21,
 	FT_OPT_STX		= 1 << 22,
 	FT_OPT_SKIP_ADDR_EXCH	= 1 << 23,
+	FT_OPT_PERF	= 1 << 24,
```
shefty, line 128: alignment looks off here, but that could be the web view

fabtests/multinode/src/core.c
```diff
@@ -48,6 +48,7 @@
 #include <core.h>
 #include <pattern.h>
 #include <shared.h>
+#include <timing.h>
```
shefty, line 51: Local include files should be using "x.h" rather than <x.h>

fabtests/multinode/src/core.c
```diff
@@ -87,6 +90,8 @@ static int multi_setup_fabric(int argc, char **argv)
 	hints->ep_attr->type = FI_EP_RDM;
 	hints->mode = FI_CONTEXT;
 	hints->domain_attr->mr_mode = opts.mr_mode;
+	hints->tx_attr->size = pm_job.num_ranks;
+	hints->rx_attr->size = pm_job.num_ranks;
```
shefty, line 94: This change looks unrelated.

fabtests/multinode/src/core.c
```diff
@@ -247,6 +251,11 @@ int multi_msg_send()
 
 		offset = state.sends_posted % opts.window_size;
 		assert(tx_ctx_arr[offset].state == OP_DONE);
+		if (ft_check_opts(FT_OPT_PERF)) {
+			timer_index = state.iter * pm_job.num_ranks +
+				(state.sends_posted % pm_job.num_ranks);
+			multi_timer_start(timers[timer_index]);
```
shefty, line 257: I have no idea what timer_index represents or if this is correct.  I *think* timers would be better defined as some sort of multi-dimensional array, but I can't tell how the array is structured.  For example, is it this (conceptually)?
```
struct timestamp {
    uint64_t start_time;
    uint64_t stop_time;
};

struct timestamp times[RANKS][ITER];
```
in which case, this could be:
```
#define START_INDEX 0
#define STOP_INDEX 1
uint64_t times[RANKS][ITER][2];
```

shefty (commented):
Comments on 1st patch only

fabtests/multinode/src/core.c
alex-mckinley: That is basically how it is structured currently. I set it up this way to make it easier to move all the timers through a sockets call. The 2d array is much easier to read, and this is a lot more readable. 

Does the concern about sending the timers even matter?

shefty: I would wrap the timestamps in some sort of 2D array abstraction.  Something sort of like this:
```
ts_alloc(max_ranks, max_entries)
{
    ts_array = calloc(max_ranks * max_entries, sizeof(*ts_array));
    ts_array->max_ranks = max_ranks;
    ts_array->max_entries = max_entries;
}

ts_entry(ts_array, rank, index)
{
    assert(rank < ts_array->max_rank && index < ts_array->max_entries);
    return ts_array + rank * ts_array->max_entries + index;
}
```
then you could do:
```
ts_entry(ts_array, cur_rank, iter)->start = ft_gettime_ns();
...
ts_entry(ts_array, cur_rank, iter)->stop = ft_gettime_ns();
```
This makes it clearer that we're dealing with an array of timestamps per rank, and removes the questionable use of the modulo operator in finding the entry.

alex-mckinley: This makes sense, I will reimplement this way

alex-mckinley: As I'm implementing this, I'm not actually sure that this saves any sort of index magic. In the current implementation the timers are ordered the way they are used, num_ranks per iteration (at maximum, some patterns use fewer). A 2d array implementation does look a tad cleaner, but the modulo just ends up inside of the index directly.

So to start a given timer it would look like: `timer_start(timers[iter][state.sends_posted % pm_job.num_ranks])`, which is slightly cleaner given the lack of multiplication on the first index.

Moving away from all of these structs does make sense in that it saves a lot of malloc calls, but I don't know if it does the readability improvement you're looking for.

shefty: What I'm looking for is an easy and direct mapping for the array entry that a given rank uses for a specific send.  It's the modulo that's confusing things.  See the ts_entry() function I listed above.  This assumes that we can calculate the total number of timestamps that we need up front in order to do one allocation.

zachdworkin (changes requested):
runmultinode.sh needs a return value so that if it fails it does not fail silently.
Currently if it fails silently the CI will think it passed and approve the stage.

Additionally the log that gets created by runmultinode.sh needs to be cat'ed so that jenkins can view the log and put it into the output of the stage.

alex-mckinley:
@shefty Updated and ready for rereview

fabtests/multinode/src/core.c
```diff
@@ -244,6 +247,10 @@ int multi_msg_send()
 
 		offset = state.sends_posted % opts.window_size;
 		assert(tx_ctx_arr[offset].state == OP_DONE);
+		if (ft_check_opts(FT_OPT_PERF)) {
+			timer_index = (state.iter * pm_job.num_ranks) + state.cur_target;
```
shefty, line 251: The timer index calculation is repeated several times.  Please create a static inline helper function to return this.  E.g.
```
static inline int timer_index(int iter, int dest_rank)
{
    return iter * pm_job.num_ranks + dest_rank;
}
```
assuming I interpreted cur_target correctly.

fabtests/multinode/src/core.c
```diff
@@ -460,18 +479,33 @@ int multinode_run_tests(int argc, char **argv)
 	if (ret)
 		return ret;
 
+	if (ft_check_opts(FT_OPT_PERF))
+		timers = malloc(sizeof(*timers) * opts.iterations * pm_job.num_ranks);
```
shefty, line 483: Use calloc, which is safer.

fabtests/multinode/src/core.c
```diff
@@ -423,15 +438,20 @@ static int multi_run_test()
...
 		if (ret)
 			return ret;
+
 	}
 	return 0;
 }
 
 static void pm_job_free_res()
 {
+	free(timers);
```
shefty, line 454: Here you call free(timers), but below...

fabtests/multinode/src/core.c
```diff
@@ -460,18 +479,33 @@ int multinode_run_tests(int argc, char **argv)
...
+			for (j = 0; j < opts.iterations * pm_job.num_ranks; j++)
+				free(timers[j]);
+		}
 
 		fflush(stdout);
 	}
-
+out:
+	for (j = 0; j < opts.iterations * pm_job.num_ranks; j++)
+		free(timers[j]);
```
shefty: ... you call free(timers[j]).  Which is it?  May be better to have timer_alloc/timer_free calls.

fabtests/multinode/src/timing.c
shefty: Previous comment is still true. Only now I don't understand why there's a malloc here and a different malloc above.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,185 @@
...
+
+extern struct pm_job_info pm_job;
+
+int multi_timer_create(struct multi_timer **timer, int rank)
+{
+	*timer = malloc(sizeof(*timer));
+
+	(*timer)->rank = rank;
+	(*timer)->start = -1;
+	(*timer)->end = -1;
```
shefty, line 46: It's more efficient to use 0 for an invalid timer.  calloc will initialize the memory to 0, so we don't need a separate init to set it to -1..

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,185 @@
...
+	double total_timers, total_min, total_max, total_sum_time, total_duration;
+	long *min = malloc(sizeof(*min) * iterations);
+	long *max = malloc(sizeof(*max) * iterations);
+	long *first_start = malloc(sizeof(*first_start) * iterations);
+	long *last_end = malloc(sizeof(*last_end) * iterations);
+	double *sum_time = malloc(sizeof(*sum_time) * iterations);
+	long duration;
+	struct multi_timer **gather_timers;
+	double iter_timer_count;
+	gather_timers = malloc(sizeof(**gather_timers) *
```
shefty, line 76: nit: add blank line between declarations and other code

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,185 @@
...
+int multi_timer_analyze(struct multi_timer **timers, int timer_count)
+{
+	int i, j, ret = 0;
+	int iterations = timer_count / pm_job.num_ranks;
+	double total_timers, total_min, total_max, total_sum_time, total_duration;
+	long *min = malloc(sizeof(*min) * iterations);
+	long *max = malloc(sizeof(*max) * iterations);
+	long *first_start = malloc(sizeof(*first_start) * iterations);
+	long *last_end = malloc(sizeof(*last_end) * iterations);
+	double *sum_time = malloc(sizeof(*sum_time) * iterations);
```
shefty, line 72: All malloc calls would be safer as calloc, and are more natural for calloc anyway.

shefty (changes requested):

alex-mckinley:
updated with feedback

contrib/intel/jenkins/Jenkinsfile
```diff
@@ -294,6 +295,338 @@ pipeline {
                   echo "MPI-tcp-rxm-1 completed."
                 )
               """
+=======
```
zachdworkin, line 298: It doesn't look like you resolved this confict.
You need to keep the current upstream jenkinsfile and just add your multinode-test stage.
Make sure to keep the spacing correct. It is using two spaces as indentation and not tabs.

zachdworkin (changes requested):

alex-mckinley:
Fixed merge conflicts

zachdworkin (approved):
@shefty, this looks fine from the CI side

alex-mckinley:
Fixed newest set of merge conflicts in the runscripts.py

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,191 @@
...
+			total_sum_time/total_timers);
+
+err:
+	free(min);
+	free(max);
+	free(sum_time);
+	free(first_start);
+	free(last_end);
+	free(gather_timers);
+	return ret & 1;
```
shefty, line 137: This is a really weird return code.  Just return 0 on success or -error on failure.  I have no idea what the intent is here.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,191 @@
...
+			for (j = 0; j < timer_count; j++)
+				all_timer[i * timer_count + j] = recv_timer[j];
+
+			free(recv_timer);
+		}
+	} else {
+		ret = socket_send(pm_job.sock, timers, 
+						  sizeof(*recv_timer) * timer_count, 0);
+		if (ret < 0)
+			return ret;
```
shefty, line 166: If check isn't needed as we call ```return ret```  below anyway.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,191 @@
...
+	free(max);
+	free(sum_time);
+	free(first_start);
+	free(last_end);
+	free(gather_timers);
+	return ret & 1;
+}
+
+int multi_timer_gather(struct multi_timer *all_timer,
+					   struct multi_timer *timers, int timer_count)
```
shefty, line 141: alignment

fabtests/multinode/src/timing.c
alex-mckinley: So I do think I have an error here, but the reason I did this was because this ret is basically set by a sockets call and that value gets passed back to ft_exit_code, which does not like a positive value and incorrectly concludes that the tests failed. I was trying to mitigate that, but this is not the right way

alex-mckinley:
Fixed latest set of requested changes

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+	long *max = calloc(iterations, sizeof(*max));
+	long *first_start = calloc(iterations, sizeof(*first_start));
+	long *last_end = calloc(iterations, sizeof(*last_end));
+	double *sum_time = calloc(iterations, sizeof(*sum_time));
+	long duration;
+	struct multi_timer *gather_timers;
+	double iter_timer_count = 0;
+
+	gather_timers = calloc(pm_job.num_ranks * pm_job.num_ranks, 
+							sizeof(*gather_timers));
```
shefty, line 78: The indentation here is painful to my eyeballs.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+	
+	if (!(min && max && first_start && last_end && sum_time && gather_timers))
+		goto err;
```
shefty, line 81: This should set ret -FI_ENOMEM before jumping to err.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+		pm_barrier();
+	}
+
+	if (pm_job.my_rank == 0)
+		printf("%-10s %16.3lf %16.3lf %16.3lf %16.3lf\n", "Average",
+			total_min/iterations, total_max/iterations,
+			total_duration/iterations,
+			total_sum_time/total_timers);
+
+err:
```
shefty, line 130: This is a common exit path, so I'd change the label from err to out.  Also if you set ret = 0 immediately before the label, that will clearly indicate that the function was successful and allow you to remove the if-else statement below.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+			for (j = 0; j < pm_job.num_ranks * pm_job.num_ranks; j++) {
+				if (gather_timers[j].start == 0 || gather_timers[j].end == 0)
+					continue;
+
+				iter_timer_count++;
+				duration = gather_timers[j].end - gather_timers[j].start;
+				sum_time[i] += duration;
+
+				if (gather_timers[j].start < first_start[i] 
+					|| first_start[i] == 0)
```
shefty, line 101: Please re-format to move the '||' to the line above and align first_start directly under gather_timers.
```
if (gather_timers... ||
_   first_start ...)
```
ignore the '_' here -- that's a hack to show the alignment without markdown converting my spaces to a tab.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+			"Iteration", "Min Send (ns)", "Max Send (ns)", 
+			"Pattern Time(ns)", "Average Send(ns)");
+
+	for (i = 0; i < iterations; i++) {
+		ret = multi_timer_iter_gather(gather_timers, timers, i);
+		if (ret < 0)
+			goto err;
+		if (pm_job.my_rank == 0) {
+			for (j = 0; j < pm_job.num_ranks * pm_job.num_ranks; j++) {
+				if (gather_timers[j].start == 0 || gather_timers[j].end == 0)
```
shefty, line 93: Here and above, please check line lengths and limit to about 80 characters.  A couple characters over is fine if it helps with readability.

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+				sum_time[i] += duration;
+
+				if (gather_timers[j].start < first_start[i] 
+					|| first_start[i] == 0)
+					first_start[i] = gather_timers[j].start;
+				if (gather_timers[j].end > last_end[i])
+					last_end[i] = gather_timers[j].end;
+				if (duration > max[i])
+					max[i] = duration;
+				if (duration < min[i] || min [i] == 0)
```
shefty, line 107: remove space between min and [i]

fabtests/multinode/src/timing.c
```diff
@@ -0,0 +1,192 @@
...
+void multi_timer_stop(struct multi_timer *timer)
+{
+	if (timer && timer->start != 0 && timer->end == 0)
+		timer->end = ft_gettime_ns();
+}
+
+static inline void print_timer(struct multi_timer timer, char* info)
+{
+	printf("rank: %i, start: %ld, end: %ld, %s\n", 
+			timer.rank, timer.start, timer.end, info);
```
shefty, line 60: align timer.rank under "rank

shefty (commented):
Assume I'm absolutely anal about indentation, line length, and text alignment.

The problem is that ends up drawing your eyes away from focusing on the code and instead to the alignment, which slows down the actual reading of the code.

alex-mckinley:
Formatting updated

fabtests/multinode/src/harness.c
```diff
@@ -57,7 +57,23 @@ static int parse_caps(char* caps)
...
+	} else if (strcmp(pattern, "ring") == 0) {
+		return 1;
+	} else if (strcmp(pattern, "gather") == 0) {
+		return 2;
+	} else if (strcmp(pattern, "broadcast") == 0) {
+		return 3;
+	} else {
+		printf("Warn: Invalid pattern, defaulting to full_mesh\n");
+		return 0;
+	}
```
shefty, line 73: As a follow up change, please convert the pattern to an enum, so we don't need a magic decoder ring to determine that '2' is gather.  E.g.
```
enum mnode_pattern {
    PATTERN_MESH,
    PATTERN_RING,
    ...
};

enum mnode_patter pattern;
```

shefty (approved):

PR #7798, opened 2022-06-01
prov/hook/dmabuf_peer_mem: Close dmabuf fd when no longer in use

j-xiong:
zeMemGetIpcHandle() allocates a new dma-buf fd every time it is called.
It is important to close the fd when it is no longer in use, otherwise
the process may run out of file descriptors.

Add a mutex to protect the dmabuf registry access to prevent the fd from
being incorrectly closed in multi-threaded runs.


prov/hook/dmabuf_peer_mem/src/hook_dmabuf_peer_mem.c
```diff
@@ -198,13 +225,28 @@ static int hook_dmabuf_peer_mem_mr_regattr(struct fid *fid,
...
 	ret = fi_mr_regattr(dom->hdomain, attr, flags, &mymr->mr_hook.hmr);
 	if (ret) {
-		if (mymr->fd >= 0)
+		if (mymr->fd >= 0) {
+			ofi_mutex_lock(&fab->mutex);
 			dmabuf_reg_remove(fab->dmabuf_reg_fd, mymr->fd);
+			err = dmabuf_reg_query(fab->dmabuf_reg_fd, mymr->base,
+					       mymr->size, &fd);
+			if (err == -ENOENT)
+				close(mymr->fd);
```
shefty, line 247: This flow looks odd.  The function doesn't open() the fd, but needs to call close(), but only if query fails. Why doesn't dmabuf_reg_remove() close the handle?

j-xiong: dmabug_reg_remove() / and dmabuf_reg_query() are wrappers of ioctl to the KMD that maintains a resistry of active dmabuf fds. The file handle is self needs to to closed from the user space. 

When the fd was opened, is was added to the registry. The same fd could be added more than once (one per MR registration) and a refcnt is used to track that. dmabuf_reg_remove() would cause the refcnt to descrease, when when it reaches 0, the entry is removed from the kenel registry. And at this point, the dmabuf is no longer used, and the user space can close the fd.  The dmfbug_reg_query() call is to determine if the entry sill exists in the kernel registry or not.

shefty: I was kind of following what the code was doing.  It just doesn't end up intuitive.  There's no clear link between calling close() and whatever call opened the fd, which I think is ze_hmem_get_handle().  Basically, this flow looks reasonable to me:
```
ze_hmem_get_handle(...);
fd = handle;
ret = dmabuf_reg_add(...);
if (ret)
    close(fd)
```
The flow where dmabuf_reg_query() is called doesn't make as much sense.  It ends up as:
```
dmabuf_req_query(..., &fd);
ret = dmabuf_reg_add(...);
/* and we do nothing with the fd on failure */
```
I understand why; reg_add() bumps a refcnt on the fd in the case of success, but req_query() does not.  But there's no way someone reading the code can figure that out.  Maybe having a call for this segment:
```
	if (mr->fd >= 0) {
		ofi_mutex_lock(&fab->mutex);
		dmabuf_reg_remove(fab->dmabuf_reg_fd, mr->fd);
		err = dmabuf_reg_query(fab->dmabuf_reg_fd, mr->base,
				       mr->size, &fd);
		if (err == -ENOENT) {
			FI_INFO(fab->fabric_hook.hprov, FI_LOG_MR,
				"Remove entry: base 0x%"PRIx64" size %"PRIu64" fd %d\n",
				mr->base, mr->size, mr->fd);
			close(mr->fd);
		}
		ofi_mutex_unlock(&fab->mutex);
	}
```
might help, rather than duplicating it.  Then we could at least add a comment in the code about a hidden refcnt.

j-xiong: Code has been refactored based on the feedbacks.  Thanks!

shefty (approved):
Thanks - easier to follow now.

PR #7806, opened 2022-06-06
prov/efa: fix bug in error handling in efa_device_construct()

wzamazon:
In function efa_device_construct()'s error handling path, resources
(like ibv_ctx) were released, but the pointer was not reset to NULL,
which can caused double free. This patch addressed the issue.

### DIFF
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -148,12 +148,17 @@ static int efa_device_construct(struct efa_device *efa_device,
 
 err_close:
 	ibv_close_device(efa_device->ibv_ctx);
+	efa_device->ibv_ctx = NULL;
 
-	if (efa_device->rdm_info)
+	if (efa_device->rdm_info) {
 		fi_freeinfo(efa_device->rdm_info);
+		efa_device->rdm_info = NULL;
+	}
 
-	if (efa_device->dgram_info)
+	if (efa_device->dgram_info) {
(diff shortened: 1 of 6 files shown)


shijin-aws (approved):

shijin-aws:
@wzamazon How can we test this bug fix?

wzamazon:
I am writing a unit test that mock `efadv_query_attr` to return an error, which will trigger the bug.

wzamazon:
@shijin-aws I added a commit to include a unit test that mimic original bug, please review again.

prov/efa/src/efa_device.c
```diff
@@ -503,3 +508,31 @@ int efa_device_get_pci_attr(struct efa_device *device,
...
+	if (ibv_device_list == NULL) {
+		skip();
+		return;
+	}
+
+	will_return(__wrap_efadv_query_device, ibv_err);
+	efa_device_construct(&efa_device, 0, ibv_device_list[0]);
+
+	/* when error happend, efa_device.ibv_ctx should be NULL */
+	assert_null(efa_device.ibv_ctx);
```
shijin-aws, line 535: why not we check `efa_device->rdm_info` and `efa_device->dgram_info`, which are not set to NULL in the err handling path before the bug fix?

wzamazon: make sense. added in new revision.

shijin-aws (approved):

PR #7810, opened 2022-06-07
prov/efa: adopt wide completion APIs in EFA provider

wenduwan:
In this change we migrate to wide completion APIs to take advantage of additional query capabilities of the extended CQ verbs. Specifically, usage of `ibv_poll_cq verb` is deprecated in favor of `ibv_{start,next,end}_poll` verbs.


prov/efa/src/efa_cq.c
```diff
@@ -118,16 +117,39 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
 	entry[i].len = (uint64_t)wc->ibv_wc.byte_len;
 }
 
+/* Unlike ibv_poll_cq, wide completion APIs do not write to ibv_wc struct. We need to do this manually.
+ * This is inspired by rdma-core, i.e. `efa_process_cqe` and `efa_process_ex_cqe`.
+ */
+static inline void efa_unsafe_write_ibv_cq_ex_wc(struct ibv_cq_ex *cq, struct ibv_wc *wc) {
```
wenduwan, line 123: The extended CQ verbs have been available in rdma-core efa provider since since Aug. 2020. We can assume that they will work in recent versions.

shijin-aws: what is the version of that rdma-core?

wzamazon:
Some high level suggestions:

1. title of the commit need to start with "prov/efa", 
2. please sign your commit by `git commit -s`
3. add description of the commit in body of the commit message

prov/efa/src/efa_cq.c
wzamazon: Function name does not look right, should be something like `efa_cq_cw_from_ibv_cq_unsafe`.

wenduwan:
```
[ec2-user@ip-10-0-15-237 rdma-core]$ git tag --contains 40ee2e2a7df82e5ed1a58fb843474795ed52ffeb
debian/31.0-1
debian/32.0-1
debian/33.0-1
debian/33.1+git20210317-1
debian/33.1-1
debian/33.2-1
debian/36.0-1
debian/36.0-2
debian/38.0-1
debian/39.0-1
v31.0
v31.1
v31.2
v31.3
v31.4
v31.5
v31.7
v32.0
v32.1
v32.2
v32.3
v32.5
v32.6
v33.0
v33.1
v33.2
v33.3
v33.5
v33.6
v34.0
v34.1
v34.2
v34.4
v34.5
v35.0
v35.1
v35.3
v35.4
v36.0
v36.1
v36.3
v36.4
v37.0
v37.1
v37.2
v37.3
v38.0
v38.1
v38.2
v39.0
v39.1
v40.0
```

wenduwan: > Function name does not look right, should be something like `efa_cq_cw_from_ibv_cq_unsafe`.

Sure. I will change the function name.

shijin-aws: That is not a very old version. It does not hurt to check some of the API's availability in https://github.com/ofiwg/libfabric/blob/main/prov/efa/configure.m4#L44-L57

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1817,46 +1822,61 @@ static inline void rdm_ep_poll_ibv_cq(struct rxr_ep *ep,
 	efa_ep = container_of(ep->rdm_ep, struct efa_ep, util_ep.ep_fid);
 	efa_av = efa_ep->av;
 	efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+
 	for (i = 0; i < cqe_to_process; i++) {
-		ret = ibv_poll_cq(efa_cq->ibv_cq, 1, &ibv_wc);
+		if (i < 1) {
+			/* Call ibv_start_poll only once */
+			ret = ibv_start_poll(efa_cq->ibv_cq_ex, &poll_cq_attr);
```
wzamazon, line 1829: similarly, `ibv_start_poll` should be outside of loop.

For this function, I think we can get rid of `cqe_to_process`, just process until there is no CQ entries, e.g. `ibv_next_poll` return ENOENT.

So something like:

```
    err = ibv_start_poll();
    if (err)
         return;

   while(!err) {
       process_cq();

       err = ibv_next_poll();
    }

    ibv_stop_poll().
```

prov/efa/src/efa_cq.c
wenduwan: > That is not a very old version. It does not hurt to check some of the API's availability in https://github.com/ofiwg/libfabric/blob/main/prov/efa/configure.m4#L44-L57

Chatted with @shijin-aws offline. rdma-core versions older than 31.0 should not be used anymore. We don't need to check the extended CQ verbs.

prov/efa/src/efa_cq.c
```diff
@@ -146,19 +168,36 @@ ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 			continue;
 		}
 
-		ret = ibv_poll_cq(cq->ibv_cq, 1, &wc.ibv_wc);
-		if (ret != 1) {
+		if (!poll_started) {
+			poll_started = true;
+			/* Call ibv_start_poll only once */
+			ret = ibv_start_poll(cq->ibv_cq_ex, &poll_cq_attr);
```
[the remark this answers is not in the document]
wenduwan: I'm ok with `start_poll` outside the loop. But in this case it's a bit tricky.

There is a precondition before polling the cq, i.e.
```c
	for (i = 0; i < count; i++) {
		if (!slist_empty(&cq->wcq)) {
			wce = container_of(cq->wcq.head, struct efa_wce, entry);
			if (wce->wc.ibv_wc.status) {
				ret = -FI_EAVAIL;
				break;
			}
			entry = slist_remove_head(&cq->wcq);
			wce = container_of(entry, struct efa_wce, entry);
			cq->read_entry(&wce->wc, i, buf);
			ofi_buf_free(wce);
			continue;
		}
                ...
                /* call ibv_start_poll */
        }

```

`start_poll` may not be called at all, and we need additional logic to know when to call `start_poll` (and probably also need to know whether we should `end_poll` before return). In this case I think the current code is easier to understand.

prov/efa/src/efa_cq.c
```diff
@@ -282,9 +324,21 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 				  util_domain.domain_fid);
 
 	size = attr->size ? attr->size : EFA_DEF_CQ_SIZE;
-	cq->ibv_cq = ibv_create_cq(cq->domain->device->ibv_ctx, size, NULL, NULL, 0);
-	if (!cq->ibv_cq) {
-		EFA_WARN(FI_LOG_CQ, "Unable to create CQ\n");
+
+	struct ibv_cq_init_attr_ex init_attr_ex = {
```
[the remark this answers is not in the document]
wenduwan: Ack. Will address.

prov/efa/src/efa_cq.c
```diff
@@ -50,42 +49,53 @@ static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
...
+	wc->opcode = ibv_wc_read_opcode(cq);
+	wc->byte_len = ibv_wc_read_byte_len(cq);
+	wc->src_qp = ibv_wc_read_src_qp(cq);
+	wc->sl = ibv_wc_read_sl(cq);
+	wc->slid = ibv_wc_read_slid(cq);
+	wc->imm_data = ibv_wc_read_imm_data(cq);
+	wc->wr_id = cq->wr_id;
+}
+
 ssize_t efa_cq_readerr(struct fid_cq *cq_fid, struct fi_cq_err_entry *entry,
```
wenduwan, line 69: I do not have a test against this function. Ideas are welcome.

prov/efa/src/efa_cq.c
```diff
@@ -50,42 +49,53 @@ static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
 	}
 }
 
+/* @brief Unlike ibv_poll_cq, wide completion APIs do not write to ibv_wc struct. We need to do this manually.
+ * This is inspired by rdma-core, i.e. `efa_process_cqe` and `efa_process_ex_cqe`.
+ */
+static inline void efa_cq_cw_from_ibv_cq_ex_unsafe(struct ibv_cq_ex *cq, struct ibv_wc *wc) {
```
wzamazon, line 55: should be `wc`, not `cw`.

prov/efa/src/efa_cq.c
```diff
@@ -50,42 +49,53 @@ static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
 	}
 }
 
+/* @brief Unlike ibv_poll_cq, wide completion APIs do not write to ibv_wc struct. We need to do this manually.
+ * This is inspired by rdma-core, i.e. `efa_process_cqe` and `efa_process_ex_cqe`.
```
wzamazon, line 53: Can you also add description for argument and return value?

prov/efa/src/efa_cq.c
```diff
@@ -50,42 +49,53 @@ static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
...
 
 	cq = container_of(cq_fid, struct efa_cq, util_cq.cq_fid);
 
 	ofi_spin_lock(&cq->lock);
-	if (slist_empty(&cq->wcq))
-		goto err;
 
-	wce = container_of(cq->wcq.head, struct efa_wce, entry);
-	if (!wce->wc.ibv_wc.status)
+	efa_cq_cw_from_ibv_cq_ex_unsafe(cq->ibv_cq_ex, &wc->ibv_wc);
```
wzamazon, line 80: at this point, I am not sure whether we need this step. Everything in `wc->ibv_wc` was from `cq->ibv_cq_ex`, then passed to the cq_err_entry. So why not just update `cq_err_entry` with `cq->ibv_cq_ex` directly?

prov/efa/src/efa_cq.c
```diff
@@ -121,48 +131,42 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
-		if (ret != 1) {
-			if (!ret)
+	/* Call ibv_start_poll only once */
+	ret = ibv_start_poll(cq->ibv_cq_ex, &poll_cq_attr);
+	if (!ret)
+		should_end_poll = true;
+
+	do {
+		if (ret) {
+			if (ret == ENOENT) {
```
wzamazon, line 157: Can be shorten to
```
ret = (ret == ENOENT) ? -FI_EAGAIN : -ret;
```

prov/efa/src/efa_cq.c
```diff
@@ -121,48 +131,42 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
-
-		ret = ibv_poll_cq(cq->ibv_cq, 1, &wc.ibv_wc);
-		if (ret != 1) {
-			if (!ret)
+	/* Call ibv_start_poll only once */
+	ret = ibv_start_poll(cq->ibv_cq_ex, &poll_cq_attr);
+	if (!ret)
+		should_end_poll = true;
+
+	do {
```
wzamazon, line 155: How about `while (!ret && i < cout)`

prov/efa/src/efa_cq.c
```diff
@@ -175,8 +179,13 @@ ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 								  wc.ibv_wc.slid,
 								  wc.ibv_wc.src_qp);
 		}
-		cq->read_entry(&wc, i, buf);
-	}
+		cq->read_entry(&wc, i++, buf);
```
wzamazon, line 182: instead of use `i`, let us use a more meaningful variable name such as `num_cqe`.

prov/efa/src/efa_cq.c
```diff
@@ -175,8 +179,13 @@ ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
...
+
+		ret = ibv_next_poll(cq->ibv_cq_ex);
+	} while (i < count);
+
+	if (should_end_poll)
+		ibv_end_poll(cq->ibv_cq_ex);
 
 	ofi_spin_unlock(&cq->lock);
 	return i ? i : ret;
```
wzamazon, line 191: Maybe outside the scope of this PR, but I do not think this return value is right.

if ret < 0 and ret != -FI_EAGAIN,  we should returned ret here, instead of i.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1803,60 +1804,72 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
...
-		if (ret == 0)
-			return;
 
-		if (OFI_UNLIKELY(ret < 0 || ibv_wc.status)) {
-			if (ret < 0) {
-				efa_eq_write_error(&ep->util_ep, -ret, -ret);
-				return;
-			}
+	/* Call ibv_start_poll only once */
+	ret = ibv_start_poll(efa_cq->ibv_cq_ex, &poll_cq_attr);
```
wzamazon, line 1826: Replace `ret` with `err` will make code more clear.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1803,60 +1804,72 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
...
+	if (!ret)
+		should_end_poll = true;
+
+	do {
+		if (OFI_UNLIKELY(ret != 0)) {
```
wzamazon, line 1831: `!=0` is no necessary.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1803,60 +1804,72 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
...
+				ret = -ret;
+			efa_eq_write_error(&ep->util_ep, ret, ret);
+			break;
+		}
 
-			pkt_entry = (void *)(uintptr_t)ibv_wc.wr_id;
-			err = ibv_wc.status;
-			prov_errno = ibv_wc.status;
-			if (ibv_wc.opcode == IBV_WC_SEND) {
+		if (ret || efa_cq->ibv_cq_ex->status) {
```
wzamazon, line 1845: at this point `ret` must be 0, right?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1803,60 +1804,72 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
...
-
-			return;
+			break;
 		}
 
-		pkt_entry = (void *)(uintptr_t)ibv_wc.wr_id;
+		pkt_entry = (void *)(uintptr_t)efa_cq->ibv_cq_ex->wr_id;
 
-		switch (ibv_wc.opcode) {
+		switch (ibv_wc_read_opcode(efa_cq->ibv_cq_ex)) {
```
wzamazon, line 1863: `opcode` has been use multiple times, maybe store it in a variable?

wzamazon (changes requested):

prov/efa/src/efa_cq.c
wenduwan: My bad. Will address.

prov/efa/src/efa_cq.c
wenduwan: Sure. Will address.

prov/efa/src/rxr/rxr_ep.c
wenduwan: I did that and realized that it's only used once on the happy path, so I did not allocate a new variable.

prov/efa/src/rxr/rxr_ep.c
wenduwan: That's right. Will address.

prov/efa/configure.m4
```diff
@@ -52,7 +52,7 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 				[efa_happy=1],
 				[
 					efa_happy=0
-					AC_MSG_WARN([The EFA provider requires rdma-core v27 or newer.])
+					AC_MSG_WARN([The EFA provider requires rdma-core v31 or newer.])
```
shijin-aws, line 55: According our earlier discussion, I thought we should check at least one of the CQ APIs in your PR before we prompt such version requirement?

prov/efa/src/rxr/rxr_ep.c
wenduwan: I agree. IIUC `OFI_UNLIKELY` is an optimization trick, i.e. `ret != 0` is unlikely to happen, which saves instructions on the happy path.

Not sure about it's significance though. Will remove it.

prov/efa/src/rxr/rxr_ep.c
wenduwan: Sure. Will address.

prov/efa/src/efa_cq.c
wenduwan: Sure. Will address.

prov/efa/src/efa_cq.c
wenduwan: I thought about that, and we also need to check if ret is negative. The reason is that `ibv_start/next_poll` API does not guarantee that its return code should be non-negative.

prov/efa/src/efa_cq.c
wenduwan: I think `do {xxx} while` is better in this case, because the `{xxx}` block also manipulates the return code from both `ibv_start/next_poll`. We probably do not want to exit the loop if `ret != 0`.

prov/efa/src/efa_cq.c
wenduwan: I'm ok either way in this case, since `efa_cq_readerr` is on the error path. I chose to do it here to make the code somewhat easier on the eye.

Will switch to `ibv_cq_ex` queries.

prov/efa/src/efa_cq.c
wenduwan: Good question. I'm not sure. I think the only way to verify is to write a test against the error path.

prov/efa/src/rxr/rxr_ep.c
wenduwan: > Not sure about it's significance though. Will remove it.
I take this back. Without `ret != 0` we will end up with more instructions on the happy path, and the code is not easier to understand. I will keep it for now.

prov/efa/configure.m4
wenduwan: Ok sure I will add that.

prov/efa/src/rxr/rxr_ep.c
wzamazon: Can you elaborate about "more instructions"?

prov/efa/configure.m4
wenduwan: Unfortunately `libefa` does not export the wide completion APIs directly. IIUC it's nontrivial to verify the API in this way. The available functions in `libefa` are not sufficient to tell if wide completion is supported by EFA or not.
```
[ec2-user@ip-10-0-15-237 workplace]$ nm -D /usr/lib64/libefa.so
                 U calloc
                 w __cxa_finalize
0000000000000000 A EFA_1.0
0000000000000000 A EFA_1.1
0000000000004570 T efadv_create_driver_qp
0000000000004730 T efadv_create_qp_ex
0000000000005490 T efadv_query_ah
0000000000003b80 T efadv_query_device
                 U __errno_location
                 U free
                 w __gmon_start__
                 U ibv_cmd_alloc_pd
                 U ibv_cmd_create_ah
                 U ibv_cmd_create_cq_ex <--- this is declared in libibverbs, not efa
                 U ibv_cmd_create_qp_ex
                 U ibv_cmd_dealloc_pd
                 U ibv_cmd_dereg_mr
                 U ibv_cmd_destroy_ah
                 U ibv_cmd_destroy_cq
                 U ibv_cmd_destroy_qp
                 U ibv_cmd_get_context
                 U ibv_cmd_modify_qp
                 U ibv_cmd_query_device_any
                 U ibv_cmd_query_port
                 U ibv_cmd_query_qp
                 U ibv_cmd_reg_mr
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 U malloc
                 U memcpy
                 U memset
                 U mmap
                 U munmap
                 U pthread_spin_destroy
                 U pthread_spin_init
                 U pthread_spin_lock
                 U pthread_spin_unlock
                 U __snprintf_chk
                 U __stack_chk_fail
                 U sysconf
                 U _verbs_init_and_alloc_context
                 U __verbs_log
                 U verbs_register_driver_34
                 U verbs_set_ops
                 U verbs_uninit_context
```

prov/efa/src/rxr/rxr_ep.c
wenduwan: Currently, if `ret = 0`, then there is only 1 `if(OFI_UNLIKELY(ret != 0))` on the code path.
`
Suppose we remove `ret != 0`, there will be 2 `if` statements, i.e. `if (ret == ENOENT)` & `if (ret < 0)`.

prov/efa/src/efa_cq.c
wzamazon: I think we can construct a unit test through cmocka.

Some like:

```
   will_return(ibv_start_poll, err_code);

   ret = efa_cq_readfrom();
   assert (ret == FI_EAVAIL);
   ret = efa_cq_readerr(cq, entry);
   assert(entry.err == ibv_err);
```

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1868,7 +1880,12 @@ static inline void rdm_ep_poll_ibv_cq(struct rxr_ep *ep,
 				"Unhandled cq type\n");
 			assert(0 && "Unhandled cq type");
 		}
-	}
+
+		err = ibv_next_poll(efa_cq->ibv_cq_ex);
+	} while (true);
```
wzamazon, line 1885: `while(true)` is ugly. I understand the intention of using do while is to reuse the code that covert rdma-core return code to libfabric return code. Can we wrap that part in an inline function? which `efa_cq_readfrom` can also use.

wzamazon (changes requested):

prov/efa/src/efa_cq.c
wzamazon: I think the part that manipulate return code from `ibv_start/next_poll` can be encapsulated in to an inline function. That function can be used also by `rdm_ep_poll_ibv_cq`.

prov/efa/src/efa_cq.c
```diff
@@ -121,65 +134,64 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
-		}
-
-		ret = ibv_poll_cq(cq->ibv_cq, 1, &wc.ibv_wc);
-		if (ret != 1) {
-			if (!ret)
-				ret = -FI_EAGAIN;
+	/* Call ibv_start_poll only once */
+	err = ibv_start_poll(cq->ibv_cq_ex, &poll_cq_attr);
+	if (!err)
+		should_end_poll = true;
```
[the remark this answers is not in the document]
wenduwan: Sounds good. Will address.

prov/efa/src/efa_cq.c
```diff
@@ -121,65 +134,64 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
 			     cq->domain->qp_table_sz_m1]->ep->av;
 
-			src_addr[i] = efa_av_reverse_lookup_dgram(av,
+			src_addr[num_cqe] = efa_av_reverse_lookup_dgram(av,
 								  wc.ibv_wc.slid,
 								  wc.ibv_wc.src_qp);
 		}
-		cq->read_entry(&wc, i, buf);
-	}
+		cq->read_entry(&wc, num_cqe++, buf);
```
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

prov/efa/src/efa_cq.c
```diff
@@ -121,65 +134,64 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
+			err = -FI_EAVAIL;
 			break;
 		}
 
 		if (src_addr) {
 			av = cq->domain->qp_table[wc.ibv_wc.qp_num &
 			     cq->domain->qp_table_sz_m1]->ep->av;
 
-			src_addr[i] = efa_av_reverse_lookup_dgram(av,
+			src_addr[num_cqe] = efa_av_reverse_lookup_dgram(av,
```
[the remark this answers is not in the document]
wenduwan: Will address.

prov/efa/src/efa_cq.c
wenduwan: That looks good. I will try it out. But I'm not familiar with unit testing C code. I might have to do this separately from this CR for time sake. Are you ok with this?

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +340,119 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+
+	const struct CMUnitTest efa_cq_unit_tests[] = {
+		cmocka_unit_test_setup_teardown(test_efa_cq_readerr_happy, setup_readerr, teardown_readerr),
+		cmocka_unit_test_setup_teardown(test_efa_cq_readerr_sad, setup_readerr, teardown_readerr),
+	};
+
+	ret = cmocka_run_group_tests_name("efa cq unit tests", efa_cq_unit_tests, NULL, NULL);
+
+	return ret;
+}
```
wenduwan, line 456:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<testsuites>
  <testsuite name="efa cq unit tests" time="0.000" tests="2" failures="0" errors="0" skipped="0" >
    <testcase name="test_efa_cq_readerr_happy" time="0.000" >
    </testcase>
    <testcase name="test_efa_cq_readerr_sad" time="0.000" >
    </testcase>
  </testsuite>
</testsuites>
```

PR #7811, opened 2022-06-07
prov/psm3 Added missing reference to neuron_init.

mwheinz:
This fixes an undefined symbol that can result if you build
libpsm3-fi.so from inside prov/psm3.

Signed-off-by: Michael Heinz <michael.heinz@intel.com>


mwheinz:
@shefty , @j-xiong - still getting this failure. Do I need to worry about this or can it be ignored?

shefty:
Failure isn't related.  One failure looks like it's related to an old job still clinging to an address.  No idea what the other 2 failures are from.

PR #7813, opened 2022-06-08
prov/shm: Fix for SAR path

jcabr020:
Fixes incorrect use of "peer_id" to "id" when checking sar_status of local
endpoint; in smr_generic_sendmsg path.

Signed-off-by: Jorge Cabrera <jorge.cabrera@intel.com>

### DIFF
--- a/prov/shm/src/smr_msg.c
+++ b/prov/shm/src/smr_msg.c
@@ -176,7 +176,7 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 	peer_smr = smr_peer_region(ep->region, id);
 
 	pthread_spin_lock(&peer_smr->lock);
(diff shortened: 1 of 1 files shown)


aingerson (approved):

shefty:
This had the standard psm3 never passes CI failure and these 2 shm failures (different CI runs):
```
fi_rma_bw -e rdm -o write -I 5 -p "shm"
fi_rma_bw -e rdm -o read -I 5 -U -p "shm"
```

aingerson:
Not sure if the systems were updated or something recently or maybe are just tired and need to be rebooted. I'll investigate both issues but this should not be related.

aingerson:
@shefty are we backporting fixes to 1.15.x? or nah?

shefty:
Yes, please backport.

PR #7814, opened 2022-06-08
core/pollfds: Minor optimizations

shefty:
Picking a couple optimizations added as part of the tcp progress restructuring patches.


shefty:
Intel CI failure was once again psm3 sighandler test.  Added exclude for it since it never passes.

shefty:
Removed psm3 exclude change in favor of fix from #7818

PR #7815, opened 2022-06-08
[v1.15.x] core/pollfds: cherry picked fixes

shefty:


shefty:
verbs/rxd failure, re-running, but I don't think it's related

PR #7816, opened 2022-06-08
include/rdma: modify peer_srx API

aingerson:
Add the following fields to the shared receive entry:
- srx pointer to get back to the srx that owns it/is processing a message for it
- owner_context for the owner/allocator to manage resources
- user_context for the user (when the entry represents an incoming message,
not a posted receive)
- flags for a peer to know whether to write completions or not
- change iov[] to iov* - the allocator should be responsible for allocating the
appropriate iov array and can set this pointer to point to where to find the array

Separate the get_msg and get_tagged functions into the following:
- get_msg/get_tagged, queue_msg/queue_tag, free_entry
This allows the peer to allocate and set the rx_entry user_context when an unexpected
message arrives so it has the needed information when the owner calls start after
getting an appropriate posted receive.
Since the owner is always responsible for allocating the rx_entry, free_entry allows
the peer to tell the owner when it is done with the owner's rx_entry

The peer ops remain the same other than removing the srx entry from the parameters
since it is now in the entry.

The flow should be as follows:
On fi_recv:
- owner allocates rx_entry and checks unexpected queue, calling start_msg if a match
is found

On incoming message:
- peer calls get_msg which will always return an rx_entry, whether a match was found
or not - if a match was found, return FI_SUCCESS, if no match, allocate entry and
return FI_ENOENT
- if no match was found, peer allocates any necessary resources for processing the
unexpected message later and saves user_context appropriately. if a match was found,
peer can begin processing the message and complete according the the rx_entry flags

### DIFF
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -158,33 +158,37 @@ struct fid_peer_srx;
 struct fi_peer_rx_entry {
 	struct fi_peer_rx_entry *next;
 	struct fi_peer_rx_entry *prev;
+	struct fid_peer_srx *srx;
 	fi_addr_t addr;
 	size_t size;
 	uint64_t tag;
+	uint64_t flags;
 	void *context;
 	size_t count;
 	void **desc;
-	struct iovec iov[];
+	void *peer_context;
+	void *owner_context;
+	struct iovec *iov;
 };
 
 struct fi_ops_srx_owner {
 	size_t	size;
-	int	(*get_msg_entry)(struct fid_peer_srx *srx,
-			struct fi_peer_rx_entry *entry);
-	int	(*get_tag_entry)(struct fid_peer_srx *srx,
-			struct fi_peer_rx_entry *entry);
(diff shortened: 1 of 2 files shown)


shefty:
The fi_peer.3.md page needs to be updated to match.

aingerson:
Updated man page

include/rdma/fi_ext.h
```diff
@@ -158,33 +158,37 @@ struct fid_peer_srx;
...
+	struct fid_peer_srx *srx;
 	fi_addr_t addr;
 	size_t size;
 	uint64_t tag;
+	uint64_t flags;
 	void *context;
 	size_t count;
 	void **desc;
-	struct iovec iov[];
+	void *peer_context;
```
wzamazon, line 169: maybe rename `peer_context` to `peer_resource` to be clearer?

shefty: I think peer_context is clearer.  'context' is the term we use throughout the API.  How the peer uses that value is up to the peer.

Issue #7429, opened 2022-02-07
Shm provider's usage of mutex lock in smr region cause performance degradtion (especially on centos7/centos8)


wzamazon:
Commit 98782f7dd02237cedfc0a5fb0ef1e85ed9955adb made the shm provider to use a mutex lock (replacing a spin lock).

Though the intention was to improve performance, we have seen the change caused performance degradation in multiple HPC workload on several OS, in general, legacy OS (like centos7 and rhel7) are more impacted than modern OS. 

For example, on CentOS 7, we have seen the performance of OpenFOAM (which is a widely used computational fluid dynamics program), drop by 25% on a typical use case (The motorbike test case).

Given this observation, I wonder if we should reconsider this change. Or, at least make shm lock type a configurable option?

wzamazon:
@aingerson and @shefty We would really appreciate your input on this issue.

shefty:
Given how apps might spin on the shm provider, I would have expected that a spin lock would be more likely to lead to starvation.  I'm not as surprised about problems with legacy OS's, as their implementation of mutex may be less optimal.  I think we need to analyze the impact for different platforms, versus assuming that the same solution will work for all.

Also, it's not as simple as reverting the lock in the commit you referenced.  There are nested locks here, some of which are still mutex.

wzamazon:
> Also, it's not as simple as reverting the lock in the commit you referenced.

I understand that reverting will not work. Do you think making shm lock type a configurable option is possible? This way user on legacy OS can choose old behavior. 

> There are nested locks here, some of which are still mutex.

Would mix type of locks be a problem? e.g. shm usage spin lock, but other parts still use mutex lock.

> I think we need to analyze the impact for different platforms, versus assuming that the same solution will work for all.

I see. What kind of analysis do we need to do? I checked that pthread on centos7 version 2.17, and is part of glibc, so it is not easy to  just upgrade pthread.

wzamazon:
>Since a mutex can block, we can't hold a spin lock then try to acquire a mutex. 

I see. Thanks for the explanation.

> I was suggesting that we need to look at the performance on, say, Centos 7 vs Centos 8. Do they both see degradation, and if so, how much?

For this workload, e.g. openfoam motorbike test case, centos7 and centos8 show similar level of degradtion, e.g. ~25%.

Amazon linux 2 and ubuntu 18.04 also observed degradation, but the degradtion is less severe (~3%).

The following table should the numbers:

[...]

PR #7817, opened 2022-06-08
prov/shm: add shared cq, shared rx, av user id support

aingerson:
Adds shared cq, shared rx, and av user id support to shm which uses it internally with itself as a first step to use shm as a peer provider

This includes the same API change from #7816 for CI purposes but will be removed once that is merged. This PR is mainly for testing purposes but also as an initial request for comment. I am still testing and reviewing and will update it. Adding do not merge label until I've cleaned it up and we've merged #7816

### DIFF
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -512,6 +512,7 @@ struct util_cq {
 	struct dlist_entry	ep_list;
 	ofi_mutex_t		ep_list_lock;
 	struct ofi_genlock	cq_lock;
+	uint64_t		flags;
 
 	struct util_comp_cirq	*cirq;
 	fi_addr_t		*src;
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -51,6 +51,7 @@
 #include <rdma/fi_tagged.h>
 #include <rdma/fi_trigger.h>
 #include <rdma/providers/fi_prov.h>
+#include <rdma/fi_ext.h>
 
 #include <ofi.h>
 #include <ofi_enosys.h>
@@ -63,6 +64,7 @@
 #include <ofi_atomic.h>
 #include <ofi_iov.h>
 #include <ofi_mr.h>
+#include <ofi_lock.h>
 
 #ifndef _SMR_H_
 #define _SMR_H_
@@ -150,11 +152,15 @@ struct smr_sar_entry {
 };
 
 struct smr_ep;
-typedef int (*smr_rx_comp_func)(struct smr_ep *ep, void *context,
-		uint64_t flags, size_t len, void *buf, fi_addr_t addr,
-		uint64_t tag, uint64_t data);
-typedef int (*smr_tx_comp_func)(struct smr_ep *ep, void *context, uint32_t op);
 
+struct smr_cq {
+	struct util_cq util_cq;
+	struct fid_peer_cq *peer_cq;
+};
+
+typedef int (*smr_rx_comp_func)(struct smr_ep *ep, void *context,
+		uint64_t flags, size_t len, void *buf, int64_t addr,
+		uint64_t tag, uint64_t data);
 
 struct smr_match_attr {
 	int64_t		id;
@@ -267,14 +273,18 @@ struct smr_sock_info {
 struct smr_ep {
 	struct util_ep		util_ep;
 	smr_rx_comp_func	rx_comp;
-	smr_tx_comp_func	tx_comp;
 	size_t			tx_size;
 	size_t			rx_size;
 	size_t			min_multi_recv_size;
 	const char		*name;
 	uint64_t		msg_id;
 	struct smr_region	*volatile region;
-	struct smr_recv_fs	*recv_fs; /* protected by rx_cq lock */
+	//if double locking is needed, shm region lock must
+	//be acquired before any shm EP locks
+	ofi_spin_t		rx_lock;
+	ofi_spin_t		tx_lock;
+
+	struct smr_recv_fs	*recv_fs;
 	struct smr_queue	recv_queue;
 	struct smr_queue	trecv_queue;
 	struct smr_unexp_fs	*unexp_fs;
@@ -338,23 +348,15 @@ int smr_write_err_comp(struct util_cq *cq, void *context,
 		       uint64_t flags, uint64_t tag, uint64_t err);
 int smr_complete_tx(struct smr_ep *ep, void *context, uint32_t op,
 		    uint64_t flags);
-int smr_tx_comp(struct smr_ep *ep, void *context, uint32_t op);
-int smr_tx_comp_signal(struct smr_ep *ep, void *context, uint32_t op);
 int smr_complete_rx(struct smr_ep *ep, void *context, uint32_t op,
 		    uint64_t flags, size_t len, void *buf, int64_t id,
 		    uint64_t tag, uint64_t data);
 int smr_rx_comp(struct smr_ep *ep, void *context,
-		uint64_t flags, size_t len, void *buf, fi_addr_t addr,
+		uint64_t flags, size_t len, void *buf, int64_t addr,
 		uint64_t tag, uint64_t data);
-int smr_rx_src_comp(struct smr_ep *ep, void *context,
-		uint64_t flags, size_t len, void *buf, fi_addr_t addr,
-		uint64_t tag, uint64_t data);
-int smr_rx_comp_signal(struct smr_ep *ep, void *context,
-		uint64_t flags, size_t len, void *buf, fi_addr_t addr,
-		uint64_t tag, uint64_t data);
-int smr_rx_src_comp_signal(struct smr_ep *ep, void *context,
-		uint64_t flags, size_t len, void *buf, fi_addr_t addr,
-		uint64_t tag, uint64_t data);
+int smr_rx_src_comp(struct smr_ep *ep, void *context, uint64_t flags,
+		    size_t len, void *buf, int64_t id, uint64_t tag,
+		    uint64_t data);
 
 static inline uint64_t smr_rx_cq_flags(uint32_t op, uint64_t rx_flags,
 				       uint16_t op_flags)
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -154,7 +154,7 @@ typedef int (*smr_rx_comp_func)(struct smr_ep *ep, void *context, uint32_t op,
 		uint16_t flags, size_t len, void *buf, fi_addr_t addr,
 		uint64_t tag, uint64_t data, uint64_t err);
 typedef int (*smr_tx_comp_func)(struct smr_ep *ep, void *context, uint32_t op,
-		uint16_t flags, uint64_t err);
+				uint64_t err);
 
 
 struct smr_match_attr {
@@ -338,9 +338,9 @@ extern smr_proto_func smr_proto_ops[smr_src_max];
 int smr_complete_tx(struct smr_ep *ep, void *context, uint32_t op,
 		uint64_t flags, uint64_t err);
 int smr_tx_comp(struct smr_ep *ep, void *context, uint32_t op,
-		uint16_t flags, uint64_t err);
+		uint64_t err);
 int smr_tx_comp_signal(struct smr_ep *ep, void *context, uint32_t op,
-		uint16_t flags, uint64_t err);
+		       uint64_t err);
 int smr_complete_rx(struct smr_ep *ep, void *context, uint32_t op,
 		uint16_t flags, size_t len, void *buf, int64_t id,
 		uint64_t tag, uint64_t data, uint64_t err);
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -196,13 +196,13 @@ static inline uint64_t smr_get_mr_flags(void **desc)
 	return ((struct ofi_mr *) *desc)->flags;
 }
 
-struct smr_unexp_msg {
+struct smr_cmd_ctx {
 	struct dlist_entry entry;
 	struct smr_cmd cmd;
 };
 
 OFI_DECLARE_FREESTACK(struct smr_rx_entry, smr_recv_fs);
-OFI_DECLARE_FREESTACK(struct smr_unexp_msg, smr_unexp_fs);
+OFI_DECLARE_FREESTACK(struct smr_cmd_ctx, smr_cmd_ctx_fs);
 OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_pend_fs);
 OFI_DECLARE_FREESTACK(struct smr_sar_entry, smr_sar_fs);
 
@@ -287,7 +287,7 @@ struct smr_ep {
 	struct smr_recv_fs	*recv_fs;
 	struct smr_queue	recv_queue;
 	struct smr_queue	trecv_queue;
-	struct smr_unexp_fs	*unexp_fs;
+	struct smr_cmd_ctx_fs	*cmd_ctx_fs;
 	struct smr_pend_fs	*pend_fs;
 	struct smr_sar_fs	*sar_fs;
 	struct smr_queue	unexp_msg_queue;
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -115,7 +115,7 @@ struct smr_rx_entry {
 	uint64_t		ignore;
 	struct iovec		iov[SMR_IOV_LIMIT];
 	uint32_t		iov_count;
-	uint16_t		flags;
+	uint64_t		flags;
 	uint64_t		err;
 	enum fi_hmem_iface	iface;
 	uint64_t		device;
@@ -150,8 +150,8 @@ struct smr_sar_entry {
 };
 
 struct smr_ep;
-typedef int (*smr_rx_comp_func)(struct smr_ep *ep, void *context, uint32_t op,
-		uint16_t flags, size_t len, void *buf, fi_addr_t addr,
+typedef int (*smr_rx_comp_func)(struct smr_ep *ep, void *context,
+		uint64_t flags, size_t len, void *buf, fi_addr_t addr,
 		uint64_t tag, uint64_t data, uint64_t err);
 typedef int (*smr_tx_comp_func)(struct smr_ep *ep, void *context, uint32_t op,
 				uint64_t err);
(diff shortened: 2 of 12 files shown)


prov/shm/src/smr_progress.c
```diff
@@ -201,9 +201,8 @@ static void smr_progress_resp(struct smr_ep *ep)
 	int ret;
 
 	pthread_spin_lock(&ep->region->lock);
-	ofi_genlock_lock(&ep->util_ep.tx_cq->cq_lock);
-	while (!ofi_cirque_isempty(smr_resp_queue(ep->region)) &&
-	       !ofi_cirque_isfull(ep->util_ep.tx_cq->cirq)) {
+	ofi_mutex_lock(&ep->util_ep.lock);
+	while (!ofi_cirque_isempty(smr_resp_queue(ep->region))) {
```
shefty, line 205: This also loses the ofi_cirque_isfull check.  Do the changes now allow overrunning the tx cq?

prov/shm/src/smr_rma.c
```diff
@@ -135,11 +135,7 @@ ssize_t smr_generic_rma(struct smr_ep *ep, const struct iovec *iov,
 		goto unlock_region;
 	}
 
-	ofi_genlock_lock(&ep->util_ep.tx_cq->cq_lock);
-	if (ofi_cirque_isfull(ep->util_ep.tx_cq->cirq)) {
-		ret = -FI_EAGAIN;
-		goto unlock_cq;
-	}
+	ofi_mutex_lock(&ep->util_ep.lock);
```
shefty, line 138: Trying to acquiring mutex while holding spinlock

shefty (commented):
Reviewed first 2 patches so far

include/rdma/fi_ext.h
```diff
@@ -158,33 +158,37 @@ struct fid_peer_srx;
...
-	int	(*start_tag)(struct fid_peer_srx *srx,
-			struct fi_peer_rx_entry *entry);
-	int	(*discard_msg)(struct fid_peer_srx *srx,
-			struct fi_peer_rx_entry *entry);
-	int	(*discard_tag)(struct fid_peer_srx *srx,
-			struct fi_peer_rx_entry *entry);
+	int	(*start_msg)(struct fi_peer_rx_entry *entry);
+	int	(*start_tag)(struct fi_peer_rx_entry *entry);
+	int	(*discard_msg)(struct fi_peer_rx_entry *entry);
+	int	(*discard_tag)(struct fi_peer_rx_entry *entry);
```
shefty, line 191: The man page needs updating as well.

prov/shm/src/smr.h
```diff
@@ -268,32 +280,49 @@ struct smr_cq {
...
+	struct smr_queue	recv_queue;
+	struct smr_queue	trecv_queue;
+	bool			dir_recv;
+	size_t			min_multi_recv_size;
+
+	//make all these generic
+	struct smr_queue	unexp_msg_queue;
+	struct smr_queue	unexp_tagged_queue;
+	struct smr_recv_fs	*recv_fs;
+	ofi_mutex_t		lock;
```
shefty, line 295: Guessing that mutex will be an issue mixed with spinlocks.

prov/shm/src/smr_ep.c
```diff
@@ -826,6 +802,47 @@ static void smr_cleanup_epoll(struct smr_sock_info *sock_info)
...
+	}
+
+	while (!dlist_empty(&srx->unexp_msg_queue.list)) {
+		dlist_pop_front(&srx->unexp_msg_queue.list, struct smr_rx_entry,
+				rx_entry, peer_entry);
+		rx_entry->peer_entry.srx->peer_ops->discard_msg(&rx_entry->peer_entry);
+		ofi_freestack_push(srx->recv_fs, rx_entry);
+	}
+
+	while (!dlist_empty(&srx->unexp_msg_queue.list)) {
```
shefty, line 833: I think you want tagged queue here.

prov/shm/src/smr_ep.c
```diff
@@ -826,6 +802,47 @@ static void smr_cleanup_epoll(struct smr_sock_info *sock_info)
...
+static void smr_srx_close(struct smr_ep *ep)
+{
+	struct smr_srx_ctx *srx;
+	struct smr_rx_entry *rx_entry;
+
+	srx = ep->srx->fid.context;
+	if (!srx)
+		return;
+
+	while (!dlist_empty(&srx->recv_queue.list)) {
```
shefty, line 814: Do the entries on all these queues correspond to posted operations from the user?  Should this generate completions?

prov/shm/src/smr_ep.c
```diff
@@ -1284,9 +1304,243 @@ static void smr_init_ipc_socket(struct smr_ep *ep)
...
+{
+	struct smr_srx_ctx *srx = (struct smr_srx_ctx *) entry->srx->fid.context;
+	struct smr_rx_entry *smr_entry, *owner_entry;
+
+	ofi_mutex_lock(&srx->lock);
+	smr_entry = container_of(entry, struct smr_rx_entry, peer_entry);
+	if (entry->owner_context) {
+		owner_entry = container_of(entry->owner_context,
+					   struct smr_rx_entry, peer_entry);
+		ofi_atomic_dec32(&owner_entry->ref);
```
shefty, line 1463: Renaming ref to multi_recv_ref would clarify that the reference counting only applies to multi-recv buffers.

prov/shm/src/smr_ep.c
```diff
@@ -1402,19 +1663,13 @@ int smr_endpoint(struct fid_domain *domain, struct fi_info *info,
...
 	ep->sar_fs = smr_sar_fs_create(info->rx_attr->size, NULL, NULL);
-	smr_init_queue(&ep->recv_queue, smr_match_msg);
-	smr_init_queue(&ep->trecv_queue, smr_match_tagged);
-	smr_init_queue(&ep->unexp_msg_queue, smr_match_unexp_msg);
-	smr_init_queue(&ep->unexp_tagged_queue, smr_match_unexp_tagged);
+
 	dlist_init(&ep->sar_list);
 
-	ep->min_multi_recv_size = SMR_INJECT_SIZE;
+	//ep->shm_srx.peer_ops = &smr_srx_peer_ops;
```
shefty, line 1672: ?

prov/shm/src/smr_msg.c
```diff
@@ -38,46 +38,51 @@
...
-	entry->peer_id = ep->util_ep.caps & FI_DIRECTED_RECV &&
-				addr != FI_ADDR_UNSPEC ?
-				smr_addr_lookup(ep->util_ep.av, addr) : -1;
-	entry->tag = tag;
-	entry->ignore = ignore;
+	entry->peer_entry.owner_context = NULL;
 
-	entry->iface = smr_get_mr_hmem_iface(ep->util_ep.domain, desc,
-					     &entry->device);
+	ofi_atomic_initialize32(&entry->ref, 0);
```
shefty, line 84: Initialize should only be called once for the structure.  This can end up abstracted as a spinlock around an integer.

shefty (commented):
Partial review of a couple more patches.

prov/shm/src/smr.h
```diff
@@ -271,7 +278,9 @@ struct smr_ep {
 	const char		*name;
 	uint64_t		msg_id;
 	struct smr_region	*volatile region;
-	struct smr_recv_fs	*recv_fs; /* protected by rx_cq lock */
+	ofi_mutex_t		rx_lock;
```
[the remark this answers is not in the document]
aingerson: I remember we changed it back to a spin lock after I changed it to a mutex and I'm looking through the issue created to address the regression (#7429) but I don't think we figured out the source of the regression. It seems to be that a mutex should be more performant for the shm region. Did we ever figure out the explanation behind the regression?

Either way, it seems we need some way to coordinate the lock type between the owner and peer. Could we define locking through the domain? Ie owner/link requests a threading level/type and the peer has to use the same lock type? Any owner that uses shm would have to optimize its shm path to use spin locks. Or we flip it and if shm is the owner, it forces any peer to use spin locks.

prov/shm/src/smr_ep.c
aingerson: Yes. Do all posted receives that weren't matched generate completions when closing the EP?

shefty: I think only usNIC doesn't generate completions for posted receives when the EP is closed.  It's helpful, because otherwise, the user may need to track the buffers.  Not all apps want to die just because one of its peers goes away.  That's the use case we'll want to target.

prov/shm/src/smr.h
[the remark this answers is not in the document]
shefty: We never figured out why spinlocks worked better.

The locking requirement is on what the callback uses, so it needs to be specified by the peer on the owner, which makes this tough.

include/ofi_util.h
```diff
@@ -517,6 +517,7 @@ struct util_cq {
 	struct dlist_entry	ep_list;
 	ofi_mutex_t		ep_list_lock;
 	struct ofi_genlock	cq_lock;
+	uint64_t		flags;
```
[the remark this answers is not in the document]
aingerson: its "own" structure is still a util_cq. I added it to the common code to allow other providers that also use the util_cq to leverage the flag as well in future implementation (ie rxm). I can separate if you prefer though?

prov/shm/src/smr_msg.c
aingerson: Do you prefer a single initialization or to abstract it with a lock?

aingerson: Just realized we always have the srx lock when modifying this value so I just dropped the protection completely

aingerson:
@shefty I *think* I addressed all your comments. I didn't update the man pages for the API changes because that's in a separate PR (#7816)

aingerson:
Updated to include some minor cleanups

aingerson:
FYI @amirshehataornl

aingerson:
Updated with rebase after API changes were merged. PR should be good to go. Let me know if you have any more comments

prov/shm/src/smr_progress.c
```diff
@@ -643,83 +627,92 @@ static int smr_progress_inject_atomic(struct smr_cmd *cmd, struct fi_ioc *ioc,
...
+			ret = smr_complete_rx(ep, rx_entry->context, cmd->msg.hdr.op,
 				comp_flags, total_len, comp_buf, cmd->msg.hdr.id,
-				cmd->msg.hdr.tag, cmd->msg.hdr.data, entry->err);
+				cmd->msg.hdr.tag, cmd->msg.hdr.data);
+		}
 		if (ret) {
 			FI_WARN(&smr_prov, FI_LOG_EP_CTRL,
 				"unable to process rx completion\n");
 		}
+		ep->srx->owner_ops->free_entry(rx_entry);
```
amirshehataornl, line 700: should this be outside the if (!sar) statement?
if sar is not NULL, I don't see where the rx_entry will be freed.

prov/shm/src/smr_cq.c
```diff
@@ -53,19 +142,30 @@ int smr_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
...
-	(*cq_fid) = &util_cq->cq_fid;
+	if (attr->flags & FI_PEER_CQ) {
+		smr_cq->peer_cq = ((struct fi_peer_cq_context *) context)->cq;
+	} else {
+		ret = smr_init_peer_cq(smr_cq);
+		if (ret)
+			goto cleanup;
+	}
+
+	smr_cq->util_cq.cq_fid.fid.ops = &smr_cq_fi_ops;
```
amirshehataornl, line 162: should smr_cq->util_cq.cq_fid.ops be set to something other than the default util_cq_ops? It works fine now, however, as far as I recall we agreed to disable the rest of the operations on the CQ. 

As outlined in the man page: If manual progress is needed on the peer CQ, the owner should drive progress by using the fi_cq_read() function with the buf parameter set to NULL and count equal 0. The peer provider should set other functions that attempt to read the peer's CQ (i.e. fi_cq_readerr, fi_cq_sread, etc.) to return -FI_ENOSYS.

I understand why the code is setup this way, so both FI_PEER_CQ case and the standard CQ would work. However, I question if we should be disabling the rest of the callbacks in the FI_PEER_CQ case.

aingerson: Sorry for the delayed response. I was on vacation. Updated with some changes. Disabling the other CQ calls (except ofi_cq_read) makes sense to me. See if that looks better to you

prov/shm/src/smr_progress.c
aingerson: Kept this outside the if (!sar) because we can't free the entry until the peer is done and has completed that operation. But you were right, it was missing the free on the delayed sar case in progress_sar_list. Added it over there.

aingerson:
@amirshehataornl See updates.

amirshehataornl (commented):
[SRX-PR-7817-comments.pdf](https://github.com/ofiwg/libfabric/files/9116876/SRX-PR-7817-comments.pdf)
I wrote my comments in a PDF file to better organize them, since they are a bit more on the design side. I would appreciate your thoughts.

shefty:
Is this PR ready for review?

aingerson:
@shefty Updating again based on conversations with Amir. I'll let you know when it's ready.

aingerson:
@shefty @amirshehataornl Sorry for the delay in updating the PR. New changes posted including changing the expects SRX flow. Ready for second review.

man/fi_endpoint.3.md
```diff
@@ -1431,12 +1431,15 @@ be associated with a single shared transmit context.
...
 shared receive context, with the buffers usable by any endpoint bound
-to the shared receive context.  Shared receive contexts are not
-associated with completion queues or counters.  Completed receive
-operations are posted to the CQs bound to the endpoint.  An endpoint
-may only be associated with a single receive context, and all
-connectionless endpoints associated with a shared receive context must
-also share the same address vector.
+to the shared receive context.  Applications must allocate and bind
+shared transmit contexts to endpoints, but operations are posted
+directly to the endpoint. An SRX must be bound to a receive CQ in order
```
shefty, line 1436: We can't make this a *must* without breaking ABI compatibility. I would reword with *may*.  This change requires updates to the providers.

shefty (commented):
Have you analyzed the performance impact of these changes?  This appears to introduce a bit of overhead in the completion path even when peer providers aren't used.

aingerson:
Pulled API change into separate PR #7987

prov/shm/src/smr.h
```diff
@@ -263,37 +269,55 @@ struct smr_sock_info {
...
+	struct smr_queue	recv_queue;
+	struct smr_queue	trecv_queue;
+	bool			dir_recv;
+	size_t			min_multi_recv_size;
+	uint64_t		rx_op_flags;
+	uint64_t		rx_msg_flags;
+
+	struct smr_cq		*cq;
+	struct smr_queue	unexp_msg_queue;
+	struct smr_queue	unexp_tagged_queue;
```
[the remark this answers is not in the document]
aingerson: Ooh I like. I'll take a look at net.

aingerson:
Decided to do a slightly better performance examination instead of just "run 2 different sets of patches in one test a few times and average it out"...
I ran latency and bandwidth tests and took the average of 10 runs with each combination and compared 4 different combinations to see how each significant patch set affects performance. The comparisons are:
1. base (no patches)
2. cleanups only (first 4 patches)
3. cleanups+scq
4. cleanups+scq+srx

Here are the results of that comparison where 1 would be even with the base, less than 1 is underperforming, and greater than one is better. The first number is the average of all the sizes and the parentheses is the ratio range from all the sizes)
Latency:
•	With cleanup: 0.94 (0.93-1.03)
•	With cleanup+scq: 1.03 (0.89-1.13)
•	With cleanup+scq+srx: 0.98 (0.79-1.15)
 
Bandwidth:
•	With cleanup: 0.97 (0.82-1.06)
•	With cleanup+scq: 1.06 (0.92-1.22)
•	With cleanup+scq+srx: 1.18 (0.97-1.56)

So it looks like the cleanup patches struggle a bit. A lot of the range issues have to do with where the protocols switch so we might have to do some tweaking of those thresholds to make it more stable. The biggest issue I see is the full patches with small message latency which is where that 0.79 comes from. I'll look into trying to make it more stable. But for the most part, the patch set seems to clean up the contentions for the rx lock which I think is where the gains come from.

amirshehataornl (changes requested):

prov/shm/src/smr_cq.c
```diff
@@ -53,19 +153,31 @@ int smr_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
...
-	ret = ofi_cq_init(&smr_prov, domain, attr, util_cq,
+	ret = ofi_cq_init(&smr_prov, domain, attr, &smr_cq->util_cq,
 			  &ofi_cq_progress, context);
 	if (ret)
 		goto free;
 
-	(*cq_fid) = &util_cq->cq_fid;
+	if (attr->flags & FI_PEER_CQ) {
+		smr_cq->peer_cq = ((struct fi_peer_cq_context *) context)->cq;
+		smr_cq->util_cq.cq_fid.ops = &smr_peer_cq_ops;
```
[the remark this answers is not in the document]
aingerson: Can you elaborate on the concern here?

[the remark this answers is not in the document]
amirshehataornl: static struct fi_ops_cq smr_peer_cq_ops = {
	.size = sizeof(struct fi_ops_cq),
	.read = ofi_cq_read,
	.readfrom = fi_no_cq_readfrom,
	.readerr = fi_no_cq_readerr,
	.sread = fi_no_cq_sread,
	.sreadfrom = fi_no_cq_sreadfrom,
	.signal = fi_no_cq_signal,
	.strerror = fi_no_cq_strerror,
};

the above structure sets the readfrom cb to a noop. ofi_cq_read() calls cq->ops->readfrom(). This results in progress being a no-op, stalling progress. So we either not override util_cq.cq_fid.ops or we set the readfrom in the above structure. Do I have the correct understanding, or am I missing something?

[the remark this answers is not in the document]
aingerson: Totally missed that loop. Thank you for catching! Will fix

aingerson:
@shefty @amirshehataornl Updated. I think this should be pretty good to go. Let me know if you have any other comments.

shefty:
Is this PR still relevent?

amirshehataornl:
not sure what you mean by relevance. These set of patches are required to support SRX. However, I'm still integrating with it. I have a couple of comments pending. I'm just working through some of the details before I post them.

shefty:
I missed @aingerson comment from 2 days ago.  I was merely going back through the PRs, and this one has been open for over 4 months.

shijin-aws:
Hello, does this PR try to make shm communicate with itself as a peer provider, so shm is both an owner and peer? And the

aingerson:
@shijin-aws Yes that is correct

shijin-aws:
> @shijin-aws Yes that is correct

Thank you! For the performance data you shared, 
>With cleanup+scq+srx: 0.98 (0.79-1.15)

Since it is comparing shm (owner) + shm (peer) with shm (no peer/owner), does it aim to prove that the layering overhead can be eliminated by the FI_PEER API? I am trying to do a PoC for onboarding EFA to use  FI_PEER to make efa + shm have the same perf with pure shm on single node. Would like to follow this PR to get a first impression.

aingerson:
@shijin-aws Right, my data was suggesting that the impact of abstracting the owner/peer is not significant so always having shm be agnostic to whether it is the peer or owner is simpler to code instead of having a separate path for shm without a peer. I made some updates recently so I was going to rerun the performance numbers again to make sure they are still correct.
I would be interested if anything in AWS shows any regression for shm with this PR versus without. Could you possibly share some before and after numbers from your end?

shijin-aws:
I only ran `fi_rdm_tagged_pingpong -p shm` so far. Here attached my data

[perf_pr7817_ec2.txt](https://github.com/ofiwg/libfabric/files/9789548/perf_pr7817_ec2.txt)

No visible perf regression by pr-7817 at my first glance.

amirshehataornl:
@shefty, @aingerson
I opened a discussion with my comments, which are not specifically about this patch, but are related: https://github.com/ofiwg/libfabric/discussions/8118
If you could let me know what you think, that'll be great.

aingerson:
Updated to pull in #8122. I think this is ok to merge.
@amirshehataornl Any outstanding issues/comments from your end?

shijin-aws:
More data including `fi_rdm_tagged_bw`. No regression observed.
[perf_pr7817_ec2.txt](https://github.com/ofiwg/libfabric/files/9833133/perf_pr7817_ec2.txt)

amirshehataornl:
> Updated to pull in #8122. I think this is ok to merge. @amirshehataornl Any outstanding issues/comments from your end?

I'm good with this patch set.

aingerson:
@shefty And you thought the CI failures would all be gone for a day :) 
```
12:13:54  - name:   fi_rdm_rma_trigger -p "sockets"
12:13:54    timestamp: 20221021-191347+0000
12:13:54    result: Fail
12:13:54    time:   300
12:13:54    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-7817/11/reg/bin/fi_rdm_rma_trigger -p "sockets"   -s ci5-eth2
12:13:54    server_stdout: |
12:13:54    client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-7817/11/reg/bin/fi_rdm_rma_trigger -p "sockets"   -s ci6-eth2 ci5-eth2
12:13:54    client_stdout: |
12:13:54      Triggered RMA write to server
12:13:54      RMA write to server
12:13:54      Received completion events for RMA write operations
```
Seems like we've been getting a lot of tcp/sockets related failures recently. Don't think it's related to the changes.
This patch set should be ok to merge.

PR #7818, opened 2022-06-09
fabtests/sighandler_test: set caps and mode

aingerson:
The test waits for the provider to complete initialization which includes posting
a receive buffer using fi_recv. This ensures the provider anticipates the fi_recv
call and sets the mode bit to support more providers.

### DIFF
--- a/fabtests/regression/sighandler_test.c
+++ b/fabtests/regression/sighandler_test.c
@@ -78,6 +78,8 @@ int main(int argc, char **argv)
 				return EXIT_FAILURE;
 			}
 		}
+		hints->caps = FI_MSG;
+		hints->mode = FI_CONTEXT;
 		if (ft_init_fabric()) {
 			ft_freehints(hints);
 			exit(EXIT_FAILURE);


aingerson:
This should fix the recurring psm3 failure. The issue is that psm3 only set the ops for the requested capabilities, preventing the application for using disabled capabilities:
```
	if (ep_cap & FI_TAGGED)
		ep_priv->ep.tagged = &psmx3_tagged_ops;
	if (ep_cap & FI_MSG)
		ep_priv->ep.msg = &psmx3_msg_ops;
	if (ep_cap & FI_RMA)
		ep_priv->ep.rma = &psmx3_rma_ops;
	if (ep_cap & FI_ATOMICS)
		ep_priv->ep.atomic = &psmx3_atomic_ops;
```
This requests FI_MSG since it is used in the test and adds FI_CONTEXT since psm3 needs it and will return ENODATA otherwise.

j-xiong:
@aingerson Thanks. This demonstrates how a simple test can be tricky to get right.

aingerson:
@j-xiong Seriously! Who knew 40 lines of code could be so picky.

fabtests/regression/sighandler_test.c
```diff
@@ -78,6 +78,8 @@ int main(int argc, char **argv)
 				return EXIT_FAILURE;
 			}
 		}
+		hints->caps = FI_MSG;
+		hints->mode = FI_CONTEXT;
```
shefty, line 82: It's better to set these prior to parsing the command line options, in case the options want to change them.

aingerson: I went back and forth but did this because all the other tests set them after for caps/modes that are required for the test so the user can't change them and make it fail.

j-xiong:
The Intel CI failures are the usual "shm" ones that can be ignored here.

j-xiong (approved):

PR #7819, opened 2022-06-09
prov/efa: fix wrong usage of address in rxr_pkt_init_runtread_rtm()

wzamazon:
rxr_pkt_init_runtread_rtm() need to get peer from address. Currently,
it is using pkt_entry->addr, which is wrong because pkt_entry->addr
has not been initialized when the function was called.

This patch fixed the issue by using "tx_entry->addr"


shijin-aws (approved):
I should have caught this during the earlier review, but the PR was too long :)

PR #7820, opened 2022-06-09
contrib/intel/jenkins: disable ze-shm stage while we investigate failures

zachdworkin:
contrib/intel/jenkins: disable ze-shm stage while we investigate failures


zachdworkin:
@shefty I used the replay feature to make sure this runs the correct scripts with jenkins

PR #7823, opened 2022-06-10
prov/efa: adopt wide completion APIs in EFA provider

wenduwan:
Migrate efa provider to wide completion APIs. Deprecate ibv_poll_cq
usage in favor of ibv_{start,next,end}_poll verbs.

Signed-off-by: wenduwan <wenduwdan@amazon.com>

In this change we migrate to wide completion APIs to take advantage of additional query capabilities of the extended CQ verbs. Specifically, usage of ibv_poll_cq verb is deprecated in favor of ibv_{start,next,end}_poll verbs.

### DIFF
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -52,7 +52,7 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 				[efa_happy=1],
 				[
 					efa_happy=0
-					AC_MSG_WARN([The EFA provider requires rdma-core v27 or newer.])
+					AC_MSG_WARN([The EFA provider requires rdma-core v31 or newer.])
 				])
 	      ])
 
@@ -148,6 +148,7 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 						 [
 							#include <stdarg.h>
 							#include <stddef.h>
+							#include <stdint.h>
 							#include <setjmp.h>
 					 	 ])
         if [ test x"$cmocka_dir" != x""]; then
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -32,14 +32,12 @@
  */
 
 #include "config.h"
-
 #include <ofi_mem.h>
-
 #include "efa.h"
+#include <infiniband/verbs.h>
 
-static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
-{
-	switch (wc->ibv_wc.opcode) {
+static inline uint64_t efa_cq_opcode_to_fi_flags(enum ibv_wc_opcode	opcode) {
+	switch (opcode) {
 	case IBV_WC_SEND:
 		return FI_SEND | FI_MSG;
 	case IBV_WC_RECV:
@@ -50,42 +48,62 @@ static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
 	}
 }
 
+static uint64_t efa_cq_wc_to_fi_flags(struct efa_wc *wc)
+{
+	return efa_cq_opcode_to_fi_flags(wc->ibv_wc.opcode);
+}
+
+/**
+ * @brief Unlike ibv_poll_cq, wide completion APIs do not write to ibv_wc struct. We need to do this manually.
+ * This is inspired by rdma-core, i.e. `efa_process_cqe` and `efa_process_ex_cqe`.
+ * @param[in]		cq	the current extended CQ
+ * @param[in,out]	wc	WC struct to write result to
+ */
+static inline void efa_cq_wc_from_ibv_cq_ex_unsafe(struct ibv_cq_ex *cq, struct ibv_wc *wc) {
+	wc->status = cq->status;
+	wc->vendor_err = ibv_wc_read_vendor_err(cq);
+	wc->wc_flags = ibv_wc_read_wc_flags(cq);
+	wc->qp_num = ibv_wc_read_qp_num(cq);
+	wc->opcode = ibv_wc_read_opcode(cq);
+	wc->byte_len = ibv_wc_read_byte_len(cq);
+	wc->src_qp = ibv_wc_read_src_qp(cq);
+	wc->sl = ibv_wc_read_sl(cq);
+	wc->slid = ibv_wc_read_slid(cq);
+	wc->imm_data = ibv_wc_read_imm_data(cq);
+	wc->wr_id = cq->wr_id;
+}
+
+static inline uint32_t efa_cq_api_version(struct efa_cq *cq) {
+	return cq->domain->fabric->util_fabric.fabric_fid.api_version;
+}
+
 ssize_t efa_cq_readerr(struct fid_cq *cq_fid, struct fi_cq_err_entry *entry,
 		       uint64_t flags)
 {
 	struct efa_cq *cq;
-	struct efa_wce *wce;
-	struct slist_entry *slist_entry;
 	uint32_t api_version;
 
 	cq = container_of(cq_fid, struct efa_cq, util_cq.cq_fid);
 
 	ofi_spin_lock(&cq->lock);
-	if (slist_empty(&cq->wcq))
+
+	if (!cq->ibv_cq_ex->status)
 		goto err;
 
-	wce = container_of(cq->wcq.head, struct efa_wce, entry);
-	if (!wce->wc.ibv_wc.status)
-		goto err;
+	api_version = efa_cq_api_version(cq);
 
-	api_version = cq->domain->fabric->util_fabric.fabric_fid.api_version;
-
-	slist_entry = slist_remove_head(&cq->wcq);
-	ofi_spin_unlock(&cq->lock);
-
-	wce = container_of(slist_entry, struct efa_wce, entry);
-
-	entry->op_context = (void *)(uintptr_t)wce->wc.ibv_wc.wr_id;
-	entry->flags = efa_cq_wc_to_fi_flags(&wce->wc);
+	entry->op_context = (void *)(uintptr_t)cq->ibv_cq_ex->wr_id;
+	entry->flags = efa_cq_opcode_to_fi_flags(ibv_wc_read_opcode(cq->ibv_cq_ex));
 	entry->err = EIO;
-	entry->prov_errno = wce->wc.ibv_wc.status;
-	EFA_WARN(FI_LOG_CQ, "Work completion status: %s\n", ibv_wc_status_str(wce->wc.ibv_wc.status));
+	entry->prov_errno = cq->ibv_cq_ex->status;
+	EFA_WARN(FI_LOG_CQ, "Work completion status: %s\n", ibv_wc_status_str(cq->ibv_cq_ex->status));
+
+	ofi_spin_unlock(&cq->lock);
 
 	/* We currently don't have err_data to give back to the user. */
 	if (FI_VERSION_GE(api_version, FI_VERSION(1, 5)))
 		entry->err_data_size = 0;
 
-	ofi_buf_free(wce);
 	return sizeof(*entry);
 err:
 	ofi_spin_unlock(&cq->lock);
@@ -118,68 +136,79 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
 	entry[i].len = (uint64_t)wc->ibv_wc.byte_len;
 }
 
+/**
+ * @brief Convert an error code from CQ poll API, e.g. `ibv_start_poll`, `ibv_end_poll`.
+ * The returned error code must be 0 (success) or negative (error).
+ * As a special case, if input error code is ENOENT (there was no item on CQ), we should return -FI_EAGAIN.
+ * @param[in] err	Return value from `ibv_start_poll` or `ibv_end_poll`
+ * @returns	Converted error code
+ */
+static inline ssize_t efa_cq_ibv_poll_error_to_fi_error(ssize_t err) {
+	if (err == ENOENT) {
+		return -FI_EAGAIN;
+	}
+
+	if (err > 0) {
+		return -err;
+	}
+
+	return err;
+}
+
 ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 			fi_addr_t *src_addr)
 {
+	bool should_end_poll = false;
 	struct efa_cq *cq;
-	struct efa_wce *wce;
-	struct slist_entry *entry;
 	struct efa_av *av;
 	struct efa_wc wc = {0};
-	ssize_t ret = 0, i;
+	ssize_t err = 0;
+	size_t num_cqe = 0; /* Count of read entries */
+
+	/* Initialize an empty ibv_poll_cq_attr struct for ibv_start_poll.
+	 * EFA expects .comp_mask = 0, or otherwise returns EINVAL.
+	 */
+	struct ibv_poll_cq_attr poll_cq_attr = {.comp_mask = 0};
 
 	cq = container_of(cq_fid, struct efa_cq, util_cq.cq_fid);
 
 	ofi_spin_lock(&cq->lock);
 
-	for (i = 0; i < count; i++) {
-		if (!slist_empty(&cq->wcq)) {
-			wce = container_of(cq->wcq.head, struct efa_wce, entry);
-			if (wce->wc.ibv_wc.status) {
-				ret = -FI_EAVAIL;
-				break;
-			}
-			entry = slist_remove_head(&cq->wcq);
-			wce = container_of(entry, struct efa_wce, entry);
-			cq->read_entry(&wce->wc, i, buf);
-			ofi_buf_free(wce);
-			continue;
-		}
+	/* Call ibv_start_poll only once regardless of count == 0 */
+	err = ibv_start_poll(cq->ibv_cq_ex, &poll_cq_attr);
+	should_end_poll = !err;
 
-		ret = ibv_poll_cq(cq->ibv_cq, 1, &wc.ibv_wc);
-		if (ret != 1) {
-			if (!ret)
-				ret = -FI_EAGAIN;
-			break;
-		}
+	while (!err && num_cqe < count) {
+		efa_cq_wc_from_ibv_cq_ex_unsafe(cq->ibv_cq_ex, &wc.ibv_wc);
 
-		/* Insert error entry into wcq */
-		if (wc.ibv_wc.status) {
-			wce = ofi_buf_alloc(cq->wce_pool);
-			if (!wce) {
-				ofi_spin_unlock(&cq->lock);
-				return -FI_ENOMEM;
-			}
-			memset(wce, 0, sizeof(*wce));
-			memcpy(&wce->wc, &wc, sizeof(wc));
-			slist_insert_tail(&wce->entry, &cq->wcq);
-			ret = -FI_EAVAIL;
+		if (cq->ibv_cq_ex->status) {
+			err = -FI_EAVAIL;
 			break;
 		}
 
 		if (src_addr) {
 			av = cq->domain->qp_table[wc.ibv_wc.qp_num &
-			     cq->domain->qp_table_sz_m1]->ep->av;
+				cq->domain->qp_table_sz_m1]->ep->av;
 
-			src_addr[i] = efa_av_reverse_lookup_dgram(av,
-								  wc.ibv_wc.slid,
-								  wc.ibv_wc.src_qp);
+			src_addr[num_cqe] = efa_av_reverse_lookup_dgram(av,
+				wc.ibv_wc.slid,
+				wc.ibv_wc.src_qp);
 		}
-		cq->read_entry(&wc, i, buf);
+
+		cq->read_entry(&wc, num_cqe, buf);
+		num_cqe++;
+
+		err = ibv_next_poll(cq->ibv_cq_ex);
 	}
 
+	err = efa_cq_ibv_poll_error_to_fi_error(err);
+
+	if (should_end_poll)
+		ibv_end_poll(cq->ibv_cq_ex);
+
 	ofi_spin_unlock(&cq->lock);
-	return i ? i : ret;
+
+	return num_cqe ? num_cqe : err;
 }
 
 static const char *efa_cq_strerror(struct fid_cq *cq_fid,
@@ -218,25 +247,15 @@ static int efa_cq_control(fid_t fid, int command, void *arg)
 static int efa_cq_close(fid_t fid)
 {
 	struct efa_cq *cq;
-	struct efa_wce *wce;
-	struct slist_entry *entry;
 	int ret;
 
 	cq = container_of(fid, struct efa_cq, util_cq.cq_fid.fid);
-
-	ofi_spin_lock(&cq->lock);
-	while (!slist_empty(&cq->wcq)) {
-		entry = slist_remove_head(&cq->wcq);
-		wce = container_of(entry, struct efa_wce, entry);
-		ofi_buf_free(wce);
-	}
-	ofi_spin_unlock(&cq->lock);
 
 	ofi_bufpool_destroy(cq->wce_pool);
 
 	ofi_spin_destroy(&cq->lock);
 
-	ret = -ibv_destroy_cq(cq->ibv_cq);
+	ret = -ibv_destroy_cq(ibv_cq_ex_to_cq(cq->ibv_cq_ex));
 	if (ret)
 		return ret;
 
@@ -261,8 +280,18 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 		struct fid_cq **cq_fid, void *context)
 {
 	struct efa_cq *cq;
-	size_t size;
 	int ret;
+	struct ibv_cq_init_attr_ex init_attr_ex = {
+		.cqe = 0,
+		.cq_context = NULL,
+		.channel = NULL,
+		.comp_vector = 0,
+		/* EFA requires these values for wc_flags and comp_mask.
+		 * See `efa_create_cq_ex` in rdma-core.
+		 */
+		.wc_flags = IBV_WC_STANDARD_FLAGS,
+		.comp_mask = 0,
+	};
 
 	if (attr->wait_obj != FI_WAIT_NONE)
 		return -FI_ENOSYS;
@@ -281,10 +310,11 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 	cq->domain = container_of(domain_fid, struct efa_domain,
 				  util_domain.domain_fid);
 
-	size = attr->size ? attr->size : EFA_DEF_CQ_SIZE;
-	cq->ibv_cq = ibv_create_cq(cq->domain->device->ibv_ctx, size, NULL, NULL, 0);
-	if (!cq->ibv_cq) {
-		EFA_WARN(FI_LOG_CQ, "Unable to create CQ\n");
+	init_attr_ex.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE;
+
+	cq->ibv_cq_ex = ibv_create_cq_ex(cq->domain->device->ibv_ctx, &init_attr_ex);
+	if (!cq->ibv_cq_ex) {
+		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ\n");
 		ret = -FI_EINVAL;
 		goto err_free_util_cq;
 	}
@@ -318,8 +348,6 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 
 	ofi_spin_init(&cq->lock);
 
-	slist_init(&cq->wcq);
-
 	*cq_fid = &cq->util_cq.cq_fid;
 	(*cq_fid)->fid.fclass = FI_CLASS_CQ;
 	(*cq_fid)->fid.context = context;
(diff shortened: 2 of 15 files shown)


prov/efa/configure.m4
```diff
@@ -148,6 +148,7 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 						 [
 							#include <stdarg.h>
 							#include <stddef.h>
+							#include <stdint.h>
```
wenduwan, line 151: According to `cmocka.h`
```c
 * These headers or their equivalents MUST be included prior to including
 * this header file.
 * @code
 * #include <stdarg.h>
 * #include <stddef.h>
 * #include <stdint.h>
 * #include <setjmp.h>
 * @endcode
 *
 * This allows test applications to use custom definitions of C standard
 * library functions and types.
```

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +340,119 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+
+	const struct CMUnitTest efa_cq_unit_tests[] = {
+		cmocka_unit_test_setup_teardown(test_efa_cq_readerr_happy, setup_readerr, teardown_readerr),
+		cmocka_unit_test_setup_teardown(test_efa_cq_readerr_sad, setup_readerr, teardown_readerr),
+	};
+
+	ret = cmocka_run_group_tests_name("efa cq unit tests", efa_cq_unit_tests, NULL, NULL);
+
+	return ret;
+}
```
wenduwan, line 456:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<testsuites>
  <testsuite name="efa cq unit tests" time="0.000" tests="2" failures="0" errors="0" skipped="0" >
    <testcase name="test_efa_cq_readerr_happy" time="0.000" >
    </testcase>
    <testcase name="test_efa_cq_readerr_sad" time="0.000" >
    </testcase>
  </testsuite>
</testsuites>
```

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +340,119 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+	struct fi_cq_err_entry *entry;
+	struct efa_cq *efa_cq;
+};
+
+static enum ibv_wc_opcode _read_send_code(struct ibv_cq_ex *current)
+{
+	return IBV_WC_SEND;
+};
+
+static int setup_readerr(void **state)
```
wenduwan, line 366: Mock out the objects directly. This eliminates dependencies on other function/API, and more importantly does not require EFA device to run the tests.

prov/efa/src/efa_cq.c
```diff
@@ -121,65 +139,57 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
 		}
-		cq->read_entry(&wc, i, buf);
+
+		cq->read_entry(&wc, num_cqe, buf);
+		num_cqe++;
+
+		err = ibv_next_poll(cq->ibv_cq_ex);
 	}
 
+	err = (err == ENOENT) ? -FI_EAGAIN : (err > 0 ? -err : err);
```
wzamazon, line 185: This is too complicated for 1 line, can you wrap it into an inline function?

prov/efa/src/efa_cq.c
```diff
@@ -121,65 +139,57 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
+
+	if (should_end_poll)
```
wzamazon, line 187: seems we can just used `if (!err)` and do not really need `should_end_poll`.

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +340,119 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+	int ret;
+
+	cmocka_set_message_output(CM_OUTPUT_XML);
+
+	const struct CMUnitTest efa_cq_unit_tests[] = {
+		cmocka_unit_test_setup_teardown(test_efa_cq_readerr_happy, setup_readerr, teardown_readerr),
+		cmocka_unit_test_setup_teardown(test_efa_cq_readerr_sad, setup_readerr, teardown_readerr),
+	};
+
+	ret = cmocka_run_group_tests_name("efa cq unit tests", efa_cq_unit_tests, NULL, NULL);
```
wzamazon, line 453: why create a new group instead of add the test to existing group?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1868,7 +1865,17 @@ static inline void rdm_ep_poll_ibv_cq(struct rxr_ep *ep,
...
+
+		err = ibv_next_poll(efa_cq->ibv_cq_ex);
+	}
+
+	if (err && err != ENOENT) {
+		err = err > 0 ? err : -err;
+		efa_eq_write_error(&ep->util_ep, err, err);
 	}
+
+	if (should_end_poll)
```
wzamazon, line 1877: similarly, can use `if (!err)`.

prov/efa/src/efa_cq.c
wenduwan: Sure. Will address.

prov/efa/src/efa_cq.c
wenduwan: I don't see a strong connection between these tests, and they are testing distinct aspects from what I can tell.

The more important reason is that, I find the recommended approach not scalable as we add more tests, i.e.
```
1. Add unit tests to the bottom of the c source file you wish to test.
2. Add headers to efa_unit_tests.h for each new unit test.
3. Create or find an existing cmocka test group in efa_unit_tests.c
4. Add your new unit test to the cmocka test group.
```

For functions called in `efa_unit_tests.c` we will need to declare them in `efa_unit_tests.h`, which requires more code. This is also restrictive for developers, e.g. it's hard to reuse static functions in `efa_xxx.c` (declaring static functions in `efa_unit_tests.h` is not as straight forward because of dependencies).

From my perspective it's better to be frugal with what we put in `efa_unit_tests.x`, and hide detailed test fixtures in respective source files. Modifying tests is easier when the related code is localized in one or two places and shared with few other tests.

wzamazon: > I don't see a strong connection between these tests, and they are testing distinct aspects from what I can tell.

They are tests of EFA provider, therefore is in a group named "efa unit tests". It would be confusing to see two groups of tests, one is named "efa unit tests", the other is named "efa cq unit tests". 

> For functions called in efa_unit_tests.c we will need to declare them in efa_unit_tests.h, which requires more code. 

Yes, that is 1 line for each test, and one will need to declare a function so that it can be called from `efa_unit_tests.c`

IMO this PR added more code (and confusion) to avoid that declaration. The function `run_efa_cq_tests` (15 lines of code) can be avoided if we add the test directly to existing group.

wenduwan: The existing `efa unit tests` should be split into smaller groups. Even though they all 'test efa provider', groups are logically independent from each other. For example, the [cmocka_run_group_tests()](https://api.cmocka.org/group__cmocka__exec.html#ga7c62fd0acf2235ce98268c28ee262a57) function has this signature:
```c
int cmocka_run_group_tests ( 
                const struct CMUnitTest group_tests[],
		CMFixtureFunction  	group_setup,
		CMFixtureFunction  	group_teardown 
)
```
It implies that one criteria for test group is that, tests in the same group should require the same setup and teardown process. Tests in this PR do not align with the existing tests and has its own requirements (e.g. the new tests do not require EFA device).

More importantly keeping all test related code in `efa_unit_tests` is not a scalable approach and leads to unnecessary code - yes, centralized code is easier to reuse, but enforcing it means the developer must share details though that's not used elsewhere. Instead the developer should decide the right level of encapsulation, and what she wants to expose.

wzamazon: > The existing efa unit tests should be split into smaller groups.

There are only 2 test in `efa unit tests`. I think split them into smaller groups is overkill.

> More importantly keeping all test related code in efa_unit_tests is not a scalable approach and leads to unnecessary code.

I am not sure I agree. In my opinion the current PR has unnecessary code: the function `run_efa_cq_test` is not necessary.

prov/efa/src/efa_cq.c
```diff
@@ -118,68 +136,79 @@ static void efa_cq_read_data_entry(struct efa_wc *wc, int i, void *buf)
...
 }
 
+/**
+ * @brief Adjust an error code from CQ poll API, e.g. `ibv_start_poll`, `ibv_end_poll`.
+ * The returned error code must be 0 (success) or negative (error).
+ * As a special case, if input error code is ENOENT (there was no item on CQ), we should return -FI_EAGAIN.
+ * @param[in] err	Return value from `ibv_start_poll` or `ibv_end_poll`
+ * @returns Adjusted error code
+ */
+static inline ssize_t efa_cq_poll_adjust_error(ssize_t err) {
```
wzamazon, line 146: `efa_cq_poll_adjust_error` -> `efa_cq_ibv_poll_error_to_fi_error`

wzamazon (commented):
Actual implementation looks good, only has a minor comments. I would prefer it as a separate commit, and have unit test as a separate commit.

The unit test refactor commit should be merged into unit test. It does not make sense to have a follow up commit to modify a previous commit.

prov/efa/src/efa_cq.c
wenduwan: I want to clarify this comment
> The unit test refactor commit should be merged into unit test

My understanding is to squash both commits into one. Business logic change should be accompanied by unit tests.

Do you mean different?

wenduwan: Sounds good. I squashed both commits.

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +359,210 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+};
+
+static int setup_readfrom(struct efa_cq_readfrom_state *state)
+{
+	int err;
+	static struct efa_cq *efa_cq;
+
+	efa_cq = calloc(1, sizeof(struct efa_cq));
+	if (!efa_cq)
+		return -1;
```
wzamazon, line 393: `return -FI_ENOMEM`.

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +359,210 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+
+	err = efa_cq_initialize(efa_cq);
+	if (err)
+		return -1;
```
wzamazon, line 397: `return err`.

prov/efa/test/efa_unit_test_utils.h
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * @brief Initialize a rxr_ep struct
+ * @param[in] rxr_ep A pointer to `struct rxr_ep`.
+ * @param[in] efa_cq A pointer to `struct efa_cq`. This is used to inject values into rxr_ep.
+ * @param[in] efa_ep A pointer to `struct efa_ep`. This is used to inject values into rxr_ep.
+ * @returns Error code. 0 if successful, non-zero if failed.
+ */
+static int rxr_ep_initialize(struct rxr_ep *rxr_ep, struct efa_cq *efa_cq, struct efa_ep *efa_ep)
```
wzamazon, line 72: why not just call `fi_endpoint` to create an ep?

Like we did in `https://github.com/ofiwg/libfabric/blob/62521e7eef653367e3a6b88c99be7173c4e7befa/prov/efa/test/efa_unit_test_ep.c#L53`

prov/efa/test/efa_unit_test_utils.h
```diff
@@ -0,0 +1,113 @@
+#include <cmocka.h>
+#include <infiniband/verbs.h>
+#include "efa.h"
+
+/**
+ * @brief Initialize a efa_cq struct
+ * @param[in] efa_cq A pointer to `struct efa_cq`.
+ * @returns Error code. 0 if successful, non-zero if failed.
+ */
+static int efa_cq_initialize(struct efa_cq *efa_cq)
```
wzamazon, line 10: Can we use `fi_cq_open` to create a real cq?

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +359,210 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
 }
 
+
+#if EFA_UNIT_TEST
+
+#include "efa_unit_tests.h"
+#include "efa_unit_test_utils.h"
+
+struct efa_cq_readfrom_state {
+	struct efa_cq *efa_cq;
```
wzamazon, line 377: Can we put `efa_cq` to `efa_unit_test_resource`?

prov/efa/src/efa_cq.c
```diff
@@ -331,11 +359,210 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+
+	assert_non_null(s);
+
+	rc = setup_readerr(s);
+	assert_int_equal(rc, 0);
+
+	s->efa_cq->ibv_cq_ex->wr_id = 0;
+	s->efa_cq->ibv_cq_ex->status = 1;
+
+	rc = efa_cq_readerr(&s->efa_cq->util_cq.cq_fid, s->entry, 0);
```
wzamazon, line 475: Do you need `entry` in the state? would a local variable work for this case?

wzamazon (changes requested):

prov/efa/src/efa_cq.c
wenduwan: The return code from setup and teardown hooks is not all that useful.

If we really care about the error trace I think it's better to use asserts in-place, e.g. assert not null.

prov/efa/test/efa_unit_test_utils.h
wenduwan: IIUC this creates a dependency on efa device. For unit test we can deep mock the data structures directly. With this deep mock we can run the test anywhere.

prov/efa/test/efa_unit_test_utils.h
wenduwan: Same as above.

prov/efa/src/efa_cq.c
wenduwan: Same as above. Avoid hard dependency on efa device.

prov/efa/src/efa_cq.c
wenduwan: We can. But I think it's encapsulated in the setup function that's shared with other test cases.

prov/efa/test/efa_unit_test_utils.h
wzamazon: all the other tests require devices any way, and there is no way to just run this test, so I say put it in for simplicity.

prov/efa/src/efa_cq.c
wzamazon: I am not sure why it is better. To encapsulate it, you had to introduce a new struct.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -2638,3 +2645,131 @@ void rxr_ep_record_tx_op_completed(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_
...
+	teardown_efa_cq(state->efa_cq);
+	free(state->efa_cq);
+	free(state->efa_ep);
+	free(state->rxr_ep);
+}
+
+void __wrap_rxr_pkt_handle_send_completion(struct rxr_ep *ep,
+										   struct rxr_pkt_entry *pkt_entry)
+{
+	return mock();
```
wzamazon, line 2710: Do you need to return in a "void" function?

wenduwan: Nope `mock()` is sufficient.

prov/efa/src/efa_cq.c
wenduwan: A new struct is a good thing. We can set up required variables, etc. in a single call to `setup_xxx`, and reuse it for different test cases, e.g. write once use everywhere.

prov/efa/test/efa_unit_test_utils.h
wenduwan: I have a hard opinion about unit test, especially around mocking.

If possible we should make these test executable regardless of the platform, and each of our test unit should adhere to its respective test boundary. If we are testing function `void ABC(struct A, struct B, struct C)`, we should produce mocked `A, B, C` for the test, rather than introducing another dependency that is not out of the test boundary.

For example, when we test `ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count, fi_addr_t *src_addr)`, we should not introduce an external API, i.e. `fi_cq_open` which is outside our test boundary. If `fi_cq_open` is broken, that will cause `efa_cq_readfrom` tests to fail - which is against the rationale of unit testing.

prov/efa/src/efa_cq.c
wzamazon: > A new struct is a good thing.

Not necessary.

> We can set up required variables, etc. in a single call to setup_xxx

But `entry` does not need to be setup.

prov/efa/test/efa_unit_test_common.c
```diff
@@ -22,20 +23,81 @@ int efa_unit_test_resource_construct(struct efa_resource* resource)
 		return ret;
 	}
 
+	ret = fi_endpoint(resource->domain, resource->info, &resource->ep, NULL);
+	if (ret) {
+		fi_close(&resource->domain->fid);
```
wzamazon, line 28: It seems we can call `efa_unit_test_resource_destroy` on the error handling path?

wzamazon (commented):
Thank you for this significant effor! I really appreciate it.

prov/efa/test/efa_unit_test_common.c
wenduwan: Good point. Will address.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -1,4 +1,118 @@
 #include "efa_unit_tests.h"
+#include "efa_unit_test_utils.h"
+
+static void setup_dgram_ep_resources(struct efa_resource *resource)
```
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -1,4 +1,118 @@
...
+	assert_int_equal(err, 0);
+
+	/* TX and RX are symmetrical from the progress engine POV. We only test RX. */
+	fi_ep_bind(resource->ep, &resource->cq->fid, FI_RECV);
+	fi_ep_bind(resource->ep, &resource->av->fid, 0);
+
+	efa_ep = container_of(resource->ep, struct efa_ep, util_ep.ep_fid);
+	efa_ep->info->caps &= ~FI_SEND;
+
+	efa_ep->rcq->ibv_cq_ex->start_poll = &_start_poll;
```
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -1,4 +1,118 @@
...
+
+	expect_any(__wrap_ibv_create_ah, pd);
+	expect_any(__wrap_ibv_create_ah, attr);
+	will_return(__wrap_ibv_create_ah, &ibv_ah);
+
+	err = fi_enable(resource->ep);
+	if (err) {
+		/* Don't leak reource */
+		efa_unit_test_resource_destroy(resource);
+		assert_int_equal(err, 0);
```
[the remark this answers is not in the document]
wenduwan: Could you elaborate?

My intention is to destroy the resource and then raise an assertion error. This way the log will print the exact line number to help debugging.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -1,4 +1,118 @@
...
+	efa_cq->ibv_cq_ex->read_slid = &_read_slid;
+	efa_cq->ibv_cq_ex->read_imm_data = &_read_imm_data;
+	efa_cq->ibv_cq_ex->read_opcode = &_read_opcode;
+	efa_cq->read_entry = &_read_entry;
+
+	expect_any(__wrap_ibv_create_ah, pd);
+	expect_any(__wrap_ibv_create_ah, attr);
+	will_return(__wrap_ibv_create_ah, &ibv_ah);
+
+	err = fi_enable(resource->ep);
```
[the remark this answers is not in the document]
wenduwan: I think it's better to keep it separate.

`efa_unit_test_resource_construct` is a generic function. It's not tied to a particular test, and not every test needs to enable the endpoint. In order to do `fi_enable(...)` the user is required to provide additional information, i.e. configure endpoint, decide whether to mock, what to do if enable fails (some tests don't care).

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -129,14 +242,292 @@ void test_rxr_ep_dc_atomic_error_handling()
...
+	will_return_maybe(_read_opcode, IBV_WC_RECV);
+	will_return(__wrap_rxr_pkt_handle_recv_error, NULL);
+	expect_value(__wrap_rxr_pkt_handle_recv_error, ep, rxr_ep);
+	expect_value(__wrap_rxr_pkt_handle_recv_error, pkt_entry, &entry);
+	expect_any(__wrap_rxr_pkt_handle_recv_error, err);
+	expect_any(__wrap_rxr_pkt_handle_recv_error, prov_errno);
+
+	rxr_ep->util_ep.progress(&rxr_ep->util_ep);
+
+	will_return_maybe(__wrap_ibv_destroy_ah, 0);
```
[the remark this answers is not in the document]
wenduwan: Because `fi_enable(...)` indirectly called `ibv_create_ah`mock, therefore we might need to mock the destruction function during resource cleanup.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -0,0 +1,72 @@
...
+
+void __wrap_rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
+										   struct rxr_pkt_entry *pkt_entry,
+										   enum rxr_lower_ep_type lower_ep_type)
+{
+	check_expected(ep);
+	check_expected(pkt_entry);
+	check_expected(lower_ep_type);
+
+	if ((int)mock() == 42) {
```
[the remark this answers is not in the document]
wenduwan: Because of [memes](https://www.buzzfeed.com/australianredcrossbloodservice/definitive-proof-number-42-actually-is-the-answer-to-life). I have another idea in mind w.r.t. mocks in general. I put 42 as a placeholder for now. Will refactor the mocks later.

It shouldn't matter unless someone want to explicitly return `FI_ENOMSG`.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -129,14 +242,292 @@ void test_rxr_ep_dc_atomic_error_handling()
...
+
+	/* Read 1 entry from CQ and encounter a bad wc status */
+	will_return(_start_poll, 0);
+	will_return(_end_poll, NULL);
+	efa_ep->rcq->ibv_cq_ex->status = 1;
+
+	efa_ep->util_ep.progress(&efa_ep->util_ep);
+
+	/* Error entry will be written to efa_ep->rcq->util_cq.aux_queue */
+	err = !slist_empty(&efa_ep->rcq->util_cq.aux_queue);
```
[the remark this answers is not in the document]
wenduwan: As we discussed earlier, `efa_cq_readerr` is an unstable implementation detail prone to change. Therefore I chose not to call it or `efa_cq_readfrom` directly. Instead, we make this assertion to make sure the bad wc is captured.
```
...
/* Error entry will be written to efa_ep->rcq->util_cq.aux_queue */
err = !slist_empty(&efa_ep->rcq->util_cq.aux_queue);
...
assert_int_equal(err, 1);
```

If you care about `err` or `prov_errno`, we can peek into `aux_queue` instead.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -129,14 +242,292 @@ void test_rxr_ep_dc_atomic_error_handling()
...
+
+	/* Cleanup resource before assertion */
+	will_return_maybe(__wrap_ibv_destroy_ah, 0);
+	efa_unit_test_resource_destroy(&resource);
+
+	assert_int_equal(err, 1);
+}
+
+/* Verify that RDM endpoint progress works with normal send work completions */
+void test_rdm_ep_progress_send_completion_happy(void **state)
```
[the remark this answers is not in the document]
wenduwan: I'm personally ok either way.

cmocka official docs use this function signature explicitly in their user guide, e.g.
```
static void test_driver(void **state)
{
    expect_string(chef_cook, order, "hotdog");
}
```

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wzamazon: I mean `assert_int_eqaul(err,0)` will always fail in the branch

prov/efa/test/efa_unit_test_mocks.c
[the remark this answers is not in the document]
wzamazon: Maybe use a macro like EFA_UNIT_TEST_CALL_REAL, and change the value to 4242.

> someone want to explicitly return FI_ENOMSG.

This can happen in some test case.

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wzamazon: Unused variable will result in compiler warning. so we want to avoid it.

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wzamazon: Yes, `efa_cq_readerr` will change.

but `fi_cq_read` is a stable API and will not change.

shijin-aws:
could u separate your commits into the unit test and the efa provider change?

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wenduwan: Yes. That's exactly my intention. Clean up the resource and fail the test immediately.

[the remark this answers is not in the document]
wzamazon: I see. I would move this line out of the `if` branch then.

prov/efa/test/rdma_core_mocks.c
```diff
@@ -30,7 +17,11 @@ int __real_ibv_destroy_ah(struct ibv_ah *ibv_ah);
 
 int __wrap_ibv_destroy_ah(struct ibv_ah *ibv_ah)
 {
-	return (ibv_ah->context == NULL) ? 0 : __real_ibv_destroy_ah(ibv_ah);
```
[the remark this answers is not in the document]
wzamazon: Still not sure why we need to get return value for this function.

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wenduwan: Makes sense. I will use it to inspect the error entry.

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

libfabric might have disabled this warning. Functions like `ssize_t ofi_cq_readerr(struct fid_cq *cq_fid, struct fi_cq_err_entry *buf, uint64_t flags)` have unused parameters, e.g. `flags`.

prov/efa/test/efa_unit_test_mocks.c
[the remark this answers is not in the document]
wenduwan: Ok I will hard code it to 4242 for now.

prov/efa/test/efa_unit_test_ep.c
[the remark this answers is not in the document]
wenduwan: I'm ok to do that. It's just that we are doing 2 comparisons if `err == 0`.

wenduwan:
> could u separate your commits into the unit test and the efa provider change?

I can but from my experience every change should be self-contained, i.e. you make the change and you prove it works. Producing separate commit makes this opaque.

Do you have a strong opinion?

shijin-aws:
i see. i don't have a strong opinion here.

> > could u separate your commits into the unit test and the efa provider change?
> 
> I can but from my experience every change should be self-contained, i.e. you make the change and you prove it works. Producing separate commit makes this opaque.
> 
> Do you have a strong opinion?

prov/efa/test/rdma_core_mocks.c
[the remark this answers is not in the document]
wenduwan: As it is, this mock does not really give the user control over its behavior. For example, if the user wants this function to return 123, there is no obvious way to do it.

It's much symmetrical and easier to formulate these mocks as:
```
// Inside the mock
...
val = mock();

if (val == something) {
    call_real();
} else if (val ==  something_else) {
    call_another_func();
}

...

return val;
```

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -129,14 +242,292 @@ void test_rxr_ep_dc_atomic_error_handling()
...
+	setup_dgram_ep_resources(&resource);
+
+	efa_ep = container_of(resource.ep, struct efa_ep, util_ep.ep_fid);
+
+	/* Read 1 entry from CQ and encounter a bad wc status */
+	will_return(_start_poll, 0);
+	will_return(_end_poll, NULL);
+	efa_ep->rcq->ibv_cq_ex->status = 1;
+
+	efa_ep->util_ep.progress(&efa_ep->util_ep);
```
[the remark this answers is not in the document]
wenduwan: IIUC cq read is mostly used in the progress engine. So I assume testing the progress engine better simulates real-world use cases, and it covers more logic than cq read alone.

wzamazon (approved):

PR #7829, opened 2022-06-13
prov/efa:  test and fix atomic tx entry leak

wzamazon:
Added a unit test to reproduce a memory leak in DC atomic, and included a bug fix for it.


prov/efa/Makefile.include
```diff
@@ -108,6 +108,7 @@ prov_efa_test_efa_unit_test_CPPFLAGS = $(efa_CPPFLAGS)
 prov_efa_test_efa_unit_test_LDADD = $(cmocka_LIBS) $(linkback)
```
shijin-aws, line 108: nit: `aatomic` to `atomic` in commit message.

PR #7835, opened 2022-06-14
fabtests/efa: fix the command in efa_retrieve_hw_counter_value()

wzamazon:
This patch added quote ("") around the filen name in the command
used by efa_retrieve_hw_counter_value() for hw counter file.

This is to address the issue that hw counter file directory is
different between nodes.


tschuett:
What prevents you from using f-strings?

wckzhang:
Seems like fstrings can be used here, I don't care either way

tschuett:
I just want to make your life easier.

wckzhang (approved):

wzamazon:
I am more used to regular string.

PR #7836, opened 2022-06-15
fabtests/multinode: Code cleanup and switch to DELIVERY_COMPLETE barrier

shefty:
Mostly white space cleanup to multinode test.  Converted the 'barrier' implementation to use FI_DELIVERY_COMPLETE for sends, instead of unspecified completion semantics, to ensure that the sends reach the destination.

Also patch to fix a compiler warning in unexpected_msg.


shefty:
Actual patch to switch to using delivery complete was dropped, as it caused shm to fail CI.  Only merging 1 cleanup patch and 1 compile warning fix.

PR #7838, opened 2022-06-16
core: Reserve CXI provider constants to avoid conflicts

swelch:
Reserve address and protocol constants for the CXI provider
to avoid conflicts in applications that may be using a
pre-upstream release of the provider with other provider
definitions.

Signed-off-by: Steve Welch <welch@hpe.com>

### DIFF
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -215,6 +215,7 @@ enum {
 	FI_ADDR_EFA,
 	FI_ADDR_PSMX3,		/* uint64_t[4] */
 	FI_ADDR_OPX,
+	FI_ADDR_CXI,
 };
 
 #define FI_ADDR_UNSPEC		((uint64_t) -1)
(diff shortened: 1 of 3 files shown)


shefty (approved):
Looks reasonable to me.  Waiting for CI

swelch:
> Looks reasonable to me. Waiting for CI

Once  CI completes/PR is merged, I will cherry pick into v1.15.x if you are ok with that.

wzamazon:
Usually we do not back port new features.

swelch:
> Usually we do not back port new features.

Yes I am aware and why I asked. However, this is a simple reservation that will allow libfabric applications built using v1.15.x to make compile time decisions; it does not add features in itself to the libfabric library.

shefty:
Backporting the change makes it easier for an out of tree provider to work with a stable release or branch.  And the change itself doesn't really bother me.  But the v1.16 release is only a couple months away, then that will become the stable branch/release.  Is there a justification for why you want this added to v1.15.x, versus waiting for v1.16?  Is cxi being developed based on v1.15?  I don't understand what compile time check this would help with, since the provider isn't necessarily present.

I guess this becomes a question for how older releases should/can support new providers.  Obviously, we wouldn't merge the provider itself, but it can install/work as a DL provider.  Are we okay accepting the new enum values, which are pretty harmless?  I don't have a strong opinion against this myself.

PR #7840, opened 2022-06-17
fabtests/efa: add fork-related test

shijin-aws:
Run fi_rdm_tagged_bw with "-K" with environment variables
` FI_EFA_FORK_SAFE=1` and `RDMAV_FORK_SAFE=1`

### DIFF
--- a/fabtests/Makefile.am
+++ b/fabtests/Makefile.am
@@ -152,7 +152,8 @@ nobase_dist_config_DATA = \
 	pytest/efa/test_multi_recv.py \
 	pytest/efa/test_rnr.py \
 	pytest/efa/test_efa_info.py \
-	pytest/efa/test_runt.py
+	pytest/efa/test_runt.py \
+	pytest/efa/test_fork_support.py
 
 noinst_LTLIBRARIES = libfabtests.la
 
--- a/fabtests/pytest/efa/test_fork_support.py
+++ b/fabtests/pytest/efa/test_fork_support.py
@@ -1,0 +1,21 @@
+import pytest
+
+
+@pytest.mark.functional
+@pytest.mark.parametrize("environment_variable", ["FI_EFA_FORK_SAFE", "RDMAV_FORK_SAFE"])
+def test_fork_support(cmdline_args, completion_type, environment_variable):
+    from common import ClientServerTest
+    import copy
+    cmdline_args_copy = copy.copy(cmdline_args)
+
+    if cmdline_args_copy.environments:
+        cmdline_args_copy.environments += " "
+    else:
+        cmdline_args_copy.environments = ""
+
+    cmdline_args_copy.environments += "{}=1".format(environment_variable)
+    test = ClientServerTest(cmdline_args_copy, "fi_rdm_tagged_bw -K",
+                            completion_type=completion_type,
+                            datacheck_type="with_datacheck")
+    test.run()
+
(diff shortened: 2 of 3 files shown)


fabtests/pytest/efa/test_rdm.py
```diff
@@ -56,3 +56,21 @@ def test_rdm_atomic(cmdline_args, iteration_type, completion_type):
                             timeout=1800)
     test.run()
 
+
+@pytest.mark.functional
+@pytest.mark.parametrize("environment_variable", ["FI_EFA_FORK_SAFE", "RDMAV_FORK_SAFE"])
+def test_rdm_fork(cmdline_args, completion_type, environment_variable):
```
wzamazon, line 62: I do not think it belong's to `test_rdm.py`, fork support show be for any endpoint type.  maybe rename `test_fork_support` and put it in `test_fork_support.py`

fabtests/pytest/efa/test_rdm.py
```diff
@@ -56,3 +56,21 @@ def test_rdm_atomic(cmdline_args, iteration_type, completion_type):
...
+    if cmdline_args_copy.environments:
+        cmdline_args_copy.environments += " "
+    else:
+        cmdline_args_copy.environments = ""
+
+    cmdline_args_copy.environments += "{}=1".format(environment_variable)
+    test = ClientServerTest(cmdline_args_copy, "fi_rdm_tagged_bw -K",
+                            completion_type=completion_type,
+                            datacheck_type="with_datacheck")
+    test.run()
```
wzamazon, line 76: Added a new line at end?

wzamazon (commented):
Would it be possible to added negative test, that when the environment variable is not set efa will abort and print warning?

I understand that would require calling `ibv_is_fork_initialized()` and check the return against `IBV_FORK_UNNEEDED`.

shijin-aws:
> 



> Would it be possible to added negative test, that when the environment variable is not set efa will abort and print warning?
> 
> I understand that would require calling `ibv_is_fork_initialized()` and check the return against `IBV_FORK_UNNEEDED`.

I thought about that, and as you said, for the negative test we need to check kernel's fork support via rdma-core API, which is not straightforward to use in fabtests. We may add that test in the future after we figure out how to use pyverbs

wzamazon (approved):

wzamazon (commented):
Thanks! The other 2 comments still apply.

wzamazon (changes requested):
the other two comments still apply

fabtests/pytest/efa/test_rdm.py
shijin-aws: I added a new file named `test_fork.py` to have this test.

fabtests/pytest/efa/test_rdm.py
shijin-aws: Fixed

fabtests/pytest/efa/test_fork.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+
+
+@pytest.mark.functional
+@pytest.mark.parametrize("environment_variable", ["FI_EFA_FORK_SAFE", "RDMAV_FORK_SAFE"])
+def test_rdm_fork(cmdline_args, completion_type, environment_variable):
```
wzamazon, line 6: I think the function name and file name should be `test_fork_support`. 

test_fork is too ambiguous.

shijin-aws: Done

wzamazon (approved):

PR #7843, opened 2022-06-21
prov/efa: restrict rdm cq poll size

wenduwan:
The previous PR to adopt wide completion APIs caused performance regression, e.g. RMA read/write bandwidth. This is attributed to the unlimited CQ poll size. It is observed that a single progress routine polled more than 1000 work completions (the cap was 50 before the change). This causes a long pole and blocks user thread from doing other work(further details are still under investigation). This PR partially reverts the previous change and reinstates the cap for CQ poll in the progress engine.

Running the same RMA BW fabtests, it is observed that, at 1MB transfer size, without this cap the read/write bandwidth is ~30% lower than before-wide-completion; with this cap the degradation drops to ~3%.

Signed-off-by: wenduwan <wenduwan@amazon.com>

### DIFF
--- a/prov/efa/src/rxr/rxr_ep.c
+++ b/prov/efa/src/rxr/rxr_ep.c
@@ -1801,10 +1801,10 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
 /**
  * @brief poll rdma-core cq and process the cq entry
  *
- * @param[in]	ep		end point
- * @param[in]	cqe_to_process	max number of cq entry to poll and process
+ * @param[in]	ep	RDM endpoint
+ * @param[in]	cqe_to_process	Max number of cq entry to poll and process. Must be positive.
  */
-static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep)
+static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_process)
 {
 	bool should_end_poll = false;
 	struct efa_cq *efa_cq;
@@ -1816,7 +1816,10 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep)
 	struct efa_ep *efa_ep;
 	struct rxr_pkt_entry *pkt_entry;
 	ssize_t err;
+	size_t i = 0;
 	int prov_errno;
+
+	assert(cqe_to_process > 0);
 
 	efa_ep = container_of(ep->rdm_ep, struct efa_ep, util_ep.ep_fid);
 	efa_av = efa_ep->av;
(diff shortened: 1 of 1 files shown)


prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1801,10 +1801,10 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
 /**
  * @brief poll rdma-core cq and process the cq entry
  *
- * @param[in]	ep		end point
- * @param[in]	cqe_to_process	max number of cq entry to poll and process
+ * @param[in]	ep	end point
+ * @param[in]	batch_size	max number of cq entry to poll and process
```
wzamazon, line 1805: I think the old name `cqe_to_process` is better.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1815,16 +1815,19 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep)
...
 	int prov_errno;
 
 	efa_ep = container_of(ep->rdm_ep, struct efa_ep, util_ep.ep_fid);
 	efa_av = efa_ep->av;
 	efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
 
-	/* Call ibv_start_poll only once */
-	err = ibv_start_poll(efa_cq->ibv_cq_ex, &poll_cq_attr);
-	should_end_poll = !err;
+	if (OFI_LIKELY(batch_size)) {
```
wzamazon, line 1826: I think `if` branch is not necessary, we should add `assert(batch_size)` here.

prov/efa/src/rxr/rxr_ep.c
wzamazon: Maybe add a note here the value must be positive.

wzamazon (changes requested):
Looks good. Some minor comments.

wzamazon (commented):
Also, can you add commit message to the commit?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1801,10 +1801,10 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
 /**
  * @brief poll rdma-core cq and process the cq entry
  *
- * @param[in]	ep		end point
- * @param[in]	cqe_to_process	max number of cq entry to poll and process
+ * @param[in]	ep	RxR endpoint
```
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

wzamazon (commented):
commit message need to be broken into multiple lines. Each line < 80 characters.

tschuett (commented):
I briefly scanned the file. You seem to be inconsistent about pre and post increments.
Note that LLVM prefers pre increment.

wenduwan:
> I briefly scanned the file. You seem to be inconsistent about pre and post increments. Note that LLVM prefers pre increment.

That's a valid point. We probably need to change that separately since it's not unique for this file. For consistency, the majority usage follows the postfix pattern, so we will keep the change as is for now.

wzamazon (approved):

PR #7846, opened 2022-06-22
prov/net: Evolved and (theoretically) optimized tcp provider

shefty:
Introduce a new 'net' provider.  This *is* the tcp provider ported as a new provider for development purposes.  Because the changes to the provider are significant, the current tcp provider was left intact for stability.  The net provider is wire compatible with the tcp provider (the protocol engine was not changed), and the intent is that once validated, the net provider will become the tcp provider, and tcp specific optimizations will be removed from rxm.  There's no intent to have separate net and tcp provider code bases.

The net provider starts with the tcp provider as the base, with the added on progress engine changes.  The progress engine updates were previously posted as a pull request.  The goal of the progress engine changes was to optimize socket management and speed up connection setup when using RDM endpoints.  However, the rxm;tcp locking scheme made the progress engine unusable because of recursive locking issues.  The only way to resolve the locking issues was to completely rework the locks used to implement rdm endpoints over tcp.

The net provider implements rdm endpoint support directly.  Although rxm could be layered over the net provider, that support is disabled by default.  rxm;net should be wire compatible with rxm;tcp, however certain optimizations for rxm;tcp are not usable by rxm;net.  (Those optimizations are part of the reason for the recursive locking problems trying to implement the new progress engine.)  Instead, the rxm;tcp optimizations, plus others, are implemented directly in the net provider.  To support rdm in the net provider, the rxm av implementation is generalized and moved to util code.  Both rxm and net re-use that av implementation.  The rxm cm code was integrated into the net provider and adapted/simplified for the net provider's use case.

The net provider allocates one progress engine per domain for managing sockets.  For rdm endpoints, this includes listening sockets, which allows monitoring and immediately responding to new connections as part of data transfer progress.  Progress engines are associated with 2 different locks, only one of which will be active.  One lock is used when handling rdm endpoints, and the other when dealing directly with msg endpoints.  Other lower-level locks currently used by the tcp provider are removed (ep lock, close lock, fid list lock, poll set locks, etc. -- I think there were 6 locks or more acquired as part of most data transfer operations when you included the rxm layering).  For example, when using rdm endpoints, the rdm lock would be active, and all endpoint locks would be disabled.  This resolved the locking issues hit by trying to update the progress engine with the tcp provider.  The resulting code is much easier to analyze for serialization correctness and a bunch of asserts were added in the code to validate locking.

The rdm endpoint cm protocol differs from that used by rxm.  So, rxm;tcp cannot communicate with net rdm endpoints.  The rdm data transfer protocol is equivalent to the rxm;tcp passthru protocol option.  The existing tcp protocol is used, including for tag matching support.

Note that I started by calling this the tcp2 provider, made a bunch of updates, then renamed it to the net provider to make it more generic.  For example, I would eventually like to support rds or other non-tcp based protocols, if that makes sense.  However, because of the rename, it's a pain to add fixes to the files before the rename was made.  Doing so results in the rename patch hitting conflicts and the need to manually re-pick the fixes into the renamed file.


shefty:
@j-xiong - If you could add a review that would be great.  It's not as bad as it might seem if you go commit by commit.

The commit [prov/tcp2: Evolution of tcp provider](https://github.com/ofiwg/libfabric/pull/7846/commits/7512c9053e3698e1d18ff19a7401dd0708fd45cf)

is a copy-paste of the tcp provider with the progress engine patch applied.  You already reviewed those changes, so you can skip this one, which is the biggest change.

The other big commit is:

[prov/tcp2: Rename to net provider](https://github.com/ofiwg/libfabric/pull/7846/commits/401858f35c9f8f1bee4dc8fde9e320b07552cdd0)

which is just a find-replace rename, which you can also skip.

(I'm fixing the windows build now).

sydidelot:
This pull request is very exciting! Thank you!
I quickly looked at the commits and I think it's going in the right direction! I will review the code more carefully during the week-end.

j-xiong:
I went over the code roughly and saw no obvious issue. Just some nitpicks:
*  Commit 576663d97cea1da "prov/tcp2: Set pep progress engine dynamically": I didn't really understand the commit message until I looked at the code.
*  Could "rxm_av.c" and the names defined inside it benefit from some kind of renaming? It feels a little bit out of place there.

j-xiong:
@shefty For the window failure, you need to add a line to "include/windows/config.h".

PR #7847, opened 2022-06-23
prov/efa: adjust multi req data size

wzamazon:
adjust the data size of multiple req sent via one call to ibv_post_send.

### DIFF
--- a/prov/efa/src/rxr/rxr_ep.c
+++ b/prov/efa/src/rxr/rxr_ep.c
@@ -389,6 +389,7 @@ void rxr_tx_entry_init(struct rxr_ep *ep, struct rxr_tx_entry *tx_entry,
 		tx_op_flags &= ~FI_COMPLETION;
 	tx_entry->fi_flags = flags | tx_op_flags;
 	tx_entry->bytes_runt = 0;
+	tx_entry->max_req_data_size = 0;
 
 	switch (op) {
 	case ofi_op_tagged:
--- a/prov/efa/src/rxr/rxr_op_entry.c
+++ b/prov/efa/src/rxr/rxr_op_entry.c
@@ -140,3 +140,32 @@ void rxr_tx_entry_set_runt_size(struct rxr_ep *ep, struct rxr_op_entry *tx_entry
 	tx_entry->bytes_runt = MIN(rxr_env.efa_runt_size - peer->num_runt_bytes_in_flight, tx_entry->total_len);
 }
 
+/**
+ * @brief total data size that will be sent/received via the multiple REQ packets
+ *
+ * Multi-req protocols send/receive data via multiple REQ packets. Different
+ * protocols have different behavior:
+ *
+ *     Medium protocols send/receive all data via REQ packets
+ *
+ *     Runting read protocols send/receive part of the data via REQ packets.
+ *     The reminder of the data is sent via other type of packets or via RDMA operations.
+ *
+ * which is why this function is needed.
+ *
+ * param[in]		pkt_type		REQ packet type
+ * param[in]		op_entry		contains operation information
+ * return		size of total data transfered by REQ packets
+ */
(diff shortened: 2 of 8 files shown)


prov/efa/src/rxr/rxr_ep.c
```diff
@@ -389,6 +389,7 @@ void rxr_tx_entry_init(struct rxr_ep *ep, struct rxr_tx_entry *tx_entry,
 		tx_op_flags &= ~FI_COMPLETION;
```
wckzhang, line 389: this commit message has a typo: effcient -> efficient

Also I don't understand what:
"the size of these packet to be close" means, can you elaborate? Reading the code it looks like you mean something like, keep data packet sizes as similar as possible.

prov/efa/src/rxr/rxr_op_entry.c
```diff
@@ -216,3 +216,43 @@ size_t rxr_tx_entry_max_req_data_capacity(struct rxr_ep *ep, struct rxr_op_entry
 						      header_flags,
 						      tx_entry->rma_iov_count);
 }
+
+/**
+ * @breif set the max_req_data_size field of a tx_entry for multi-req
```
wckzhang, line 221: brief*

prov/efa/src/rxr/rxr_op_entry.c
```diff
@@ -216,3 +216,43 @@ size_t rxr_tx_entry_max_req_data_capacity(struct rxr_ep *ep, struct rxr_op_entry
...
+
+	assert(rxr_pkt_is_mulreq(pkt_type));
+
+	max_req_data_capacity = rxr_tx_entry_max_req_data_capacity(ep, tx_entry, pkt_type);
+	assert(max_req_data_capacity);
+
+	mulreq_total_data_size = rxr_op_entry_mulreq_total_data_size(tx_entry, pkt_type);
+	assert(mulreq_total_data_size);
+
+	num_req = (mulreq_total_data_size - 1)/max_req_data_capacity + 1;
```
wckzhang, line 253: why is this calculation different from rxr_tx_entry_num_req. Shouldn't you just use that function here?

wckzhang (changes requested):

prov/efa/src/rxr/rxr_ep.c
wzamazon: Yes. I want to make the data size as similar as possible.

wzamazon: Updated the comment and commit title.

prov/efa/src/rxr/rxr_op_entry.c
wzamazon: Thank you! Fixed.

prov/efa/src/rxr/rxr_op_entry.c
```diff
@@ -216,3 +216,43 @@ size_t rxr_tx_entry_max_req_data_capacity(struct rxr_ep *ep, struct rxr_op_entry
...
+
+	tx_entry->max_req_data_size = ofi_get_aligned_size((mulreq_total_data_size - 1)/num_req + 1, MEMORY_ALIGNMENT);
+	if (tx_entry->max_req_data_size >  max_req_data_capacity)
```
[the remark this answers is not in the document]
wzamazon: Thank you! fixed.

prov/efa/src/rxr/rxr_op_entry.c
wzamazon: No. That function uses `rxr_tx_entry->max_req_data_size`, which is set by this function.

At the very begining, we do not know how many packets we are  going to send. so we need to determine that by using data_capacity.

Once we know the number of packet, we can calculate the average packet size then round up to alignment. 

we then calculate the number of packet again, which is `rxr_op_entry_num_req` for

wckzhang (approved):

wzamazon:
Will add a unit test as a follow up PR.

PR #7848, opened 2022-06-23
LINKx and other work

amirshehataornl:
This a WIP of the work I've been doing. It contains the following set of changes

1. CQ completion callbacks
2. My implementation of the shared completion queues in shm
3. LINKx provider. Supports only tagged messages as of now
4. ROCM HSA API Support
5. IPC caching mechanism
6. IPC asynchronous writes
7. XPMEM Support
8. Support XPMEM export for specific memory regions instead of exporting entire address space.
9. ROCM HIP API Support (intended as a reference implementation at the moment)

This is intended to get feedback on the changes. A more finalized work is still coming down the pipe


rajachan:
There's a lot of stuff here. Can you split your PRs into granular ones by component / provider?

wzamazon:
@amirshehataornl

Looks like there are a lot of shm enhancement, like xpmem support and IPC handle. 

Would you please put them in separate PR?

tschuett:
There is no man page for the new provider.

amirshehataornl:
I'll split up into separate PRs and add a man page for the LINKx provider once I integrate with Intel's SRX patch.

prov/shm/src/smr_ep.c
```diff
@@ -415,7 +415,9 @@ static int smr_format_ipc(struct smr_cmd *cmd, void *ptr, size_t len,
 	cmd->msg.data.ipc_info.iface = iface;
 
 	return ofi_hmem_get_handle(cmd->msg.data.ipc_info.iface, ptr,
-				   (void **)&cmd->msg.data.ipc_info.ipc_handle);
```
shijin-aws, line 418: I think you need to fill the `ipc_info.address` and `ipc_info.base_length`

prov/shm/src/smr_progress.c
```diff
@@ -492,9 +492,9 @@ static int smr_progress_ipc(struct smr_cmd *cmd, enum fi_hmem_iface iface,
 				(void **) &cmd->msg.data.ipc_info.fd_handle,
 				&ipc_fd, ipc_device, &base);
 	} else {
-		ret = ofi_hmem_open_handle(cmd->msg.data.ipc_info.iface,
-				(void **) &cmd->msg.data.ipc_info.ipc_handle,
-				device, &base);
+		ret = ipc_cache_map_memhandle(ep->hmem_cache,
+									  &cmd->msg.data.ipc_info,
+									  &base);
```
shijin-aws, line 623: In subsequent calls the cache is queried and the mapped address is retrieved. But in L520 the handle is already closed so the address is invalid. Do I miss any change that fixes this issue?

src/hmem_cache.c
```diff
@@ -0,0 +1,284 @@
...
+						 "failed to unmap addr:%p\n", region->mapped_addr);
+			}
+
+			free(region);
+		}
+	}
+
+	ret = ofi_hmem_open_handle(key->iface, (void**) &key->ipc_handle,
+							   key->base_length, key->dev_num, mapped_addr);
+	if (OFI_UNLIKELY(ret)) {
```
shijin-aws, line 201: If the error is the address already mapped, do you need to unmap all overlapping regions and retry? Basically what ucx does here: https://github.com/openucx/ucx/blob/master/src/uct/cuda/cuda_ipc/cuda_ipc_cache.c#L278-L310

I am hitting the same error when sender gets the same ipc handle but with a different address.

prov/shm/src/smr_progress.c
shijin-aws: Thank you for your work! I am actually testing your commits in my development for using CUDA IPC in shm, I will leave more comments when I have some more issues.

include/ofi_hmem.h
```diff
@@ -253,9 +257,10 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
 			     size_t hmem_iov_count, uint64_t hmem_iov_offset,
 			     const void *src, size_t size);
 
-int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *dev_buf, void **handle);
+int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *dev_buf,
```
shijin-aws, line 331: I do not think you should modify this API. I guess you may want to deal with the length of the buffer allocation and the offset of the buffer that is related to the IPC handle. You can use the `ofi_hmem_get_base_addr` to get the base address of the allocation and then do the offset math.

include/ofi_hmem.h
```diff
@@ -253,9 +257,10 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
...
+						size_t *len, void **handle, uint64_t *offset);
 int ofi_hmem_open_handle(enum fi_hmem_iface iface, void **handle,
```
shijin-aws, line 333: similar comment, why do you need `len` to open a handle?

shijin-aws:
@amirshehataornl I opened a PR to introduce IPC caching mechanism in https://github.com/ofiwg/libfabric/pull/7893, it will be appreciated if you could review it as well. I could not request you as a reviewer for some reason.

amirshehataornl:
closing this in preparation of a more uptodate one.

PR #7852, opened 2022-06-24
[v1.15.x] cherry picked updates - init genlock, tcp signaling of cq

shefty:


shefty:
Removed cq progress pending further investigation.

shefty:
Intel CI failure is reproducible, but also occurs without these patches on the v1.15.x branch and with main.
```
fi_rdm_multi_client -C 10 -I 5 -p "verbs;ofi_rxd"
```
Test fails repeatedly on 2 of the ci nodes (21 and 22 I think), but passed on 15 & 16 when I tested there manually.  Assuming the problem is related to the nodes

shefty:
I believe the issue is with the kernel rdma drivers reporting the qpn incorrectly for additionally allocated qps.

PR #7853, opened 2022-06-24
contrib/intel/jenkins: Refactor tests.py to run mpis without shell scripts

zachdworkin:
Mpi's included - [ompi, mpich, impi]
Integrate run_impi.sh into python
Integrate run_mpich.sh into python
Integrate ompi the same way that impi, and mpich are integrated

Change imb, osu, and mpichtestsuite classes to use new mpi classes
Refactor common code into the super class
Change string notation to use newer "f string" notation which is cleaner than the old ".format()" notation


zachdworkin:
@shefty I did the replay thing to make sure it passes with my changes. @nikhilnanal also went over the code with me so this should be ready to merge.

PR #7854, opened 2022-06-25
prov/util + most: specify domain lock during initialization

shefty:
Updates the util code + most providers to specify the desired type of domain lock as part of init.  This replaces the flags parameter to init.  Update the net provider to use the progress lock in place of the cq and domain locks (latter is for memory registration).  This change results in minor changes to util/cq.


shefty:
@wzamazon @aingerson -- see provider changes (very minor)

PR #7856, opened 2022-06-27
prov/efa: recover unknown peer AH with raw address from efadv

wenduwan:
It is possible that the EFA device would return an unknown AH(-1) for a known
peer in receive op. This is a bug and could cause the received packet to be
discarded. As a mitigation, the EFA provider attempts to recover the AH if the
node supports extensible CQ. The additional efadv verbs allows the receiver
to retrieve peer raw address. Thereby the provider is able to infer the AH via
Raw Address:QPN:QKey.

Signed-off-by: wenduwan <wenduwan@amazon.com>

### DIFF
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -11,6 +11,8 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	# Determine if we can support the efa provider
 	efa_happy=0
 	efa_h_enable_poisoning=0
+	# Determine if efa device supports extensible CQ with access to additional fields from WC
+	efadv_support_extended_cq=1
 
 	AS_IF([test x"$enable_efa" != x"no"],
 	      [FI_CHECK_PACKAGE([efa_ibverbs],
@@ -107,6 +109,37 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 		])
 
 	AS_IF([test $efa_happy -eq 1 ], [$1], [$2])
+
+	AS_IF([test $efadv_support_extended_cq -eq 1],
+		[FI_CHECK_PACKAGE([efadv],
+			[infiniband/efadv.h],
+			[efa],
+			[efadv_create_cq],
+			[-libverbs],
+			[$efa_PREFIX],
+			[$efa_LIBDIR],
+			[],
+			[efadv_support_extended_cq=0])
+	      ])
+
+	AS_IF([test $efadv_support_extended_cq -eq 1],
+		[FI_CHECK_PACKAGE([efadv],
+			[infiniband/efadv.h],
+			[efa],
+			[efadv_cq_from_ibv_cq_ex],
+			[-libverbs],
+			[$efa_PREFIX],
+			[$efa_LIBDIR],
+			[],
+			[efadv_support_extended_cq=0])
+	      ])
+
+	AS_IF([test $efadv_support_extended_cq -eq 1],
+		[AC_DEFINE([HAVE_EFADV_CQ_EX], [1], [EFA device support extensible CQ])],
+		[AC_DEFINE([HAVE_EFADV_CQ_EX], [0], [EFA device does not support extensible CQ])]
+		)
+
+	AM_CONDITIONAL([HAVE_EFADV_CQ_EX], [test $efadv_support_extended_cq -eq 1])
 
 	efa_CPPFLAGS="$efa_ibverbs_CPPFLAGS $efadv_CPPFLAGS"
 	efa_LDFLAGS="$efa_ibverbs_LDFLAGS $efadv_LDFLAGS"
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2013-2015 Intel Corporation, Inc.  All rights reserved.
- * Copyright (c) 2017-2019 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright (c) 2017-2022 Amazon.com, Inc. or its affiliates. All rights reserved.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
@@ -249,13 +249,49 @@ static struct fi_ops efa_cq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
-int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
-		struct fid_cq **cq_fid, void *context)
-{
-	struct efa_cq *cq;
-	int ret;
+
+/**
+ * @brief Create and set cq->ibv_cq_ex
+ *
+ * @param[in] cq Pointer to the efa_cq. cq->ibv_cq_ex must be NULL.
+ * @param[in] attr Pointer to fi_cq_attr.
+ * @param[out] Return code = 0 if successful, or negative otherwise.
+ */
+#if HAVE_EFADV_CQ_EX
+static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
 	struct ibv_cq_init_attr_ex init_attr_ex = {
-		.cqe = 0,
+		.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE,
+		.cq_context = NULL,
+		.channel = NULL,
+		.comp_vector = 0,
+		/* EFA requires these values for wc_flags and comp_mask.
+		 * See `efa_create_cq_ex` in rdma-core.
+		 */
+		.wc_flags = IBV_WC_STANDARD_FLAGS,
+		.comp_mask = 0,
+	};
+	struct efadv_cq_init_attr efadv_cq_init_attr = {
+		.comp_mask = 0,
+		.wc_flags = EFADV_WC_EX_WITH_AH,
+	};
+
+	if (cq->ibv_cq_ex) {
+		EFA_WARN(FI_LOG_CQ, "CQ already has attached ibv_cq_ex\n");
+		return -FI_EALREADY;
+	}
+
+	/* To enable additional WC fields, create CQ with efadv_create_cq verb instead. */
+	cq->ibv_cq_ex = efadv_create_cq(cq->domain->device->ibv_ctx,
+									&init_attr_ex,
+									&efadv_cq_init_attr,
+									sizeof(efadv_cq_init_attr));
+
+	return cq->ibv_cq_ex ? 0 : -FI_ENOCQ;
+}
+#else
+static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
+	struct ibv_cq_init_attr_ex init_attr_ex = {
+		.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE,
 		.cq_context = NULL,
 		.channel = NULL,
 		.comp_vector = 0,
@@ -266,6 +302,24 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 		.comp_mask = 0,
 	};
 
+	if (cq->ibv_cq_ex) {
+		EFA_WARN(FI_LOG_CQ, "CQ already has attached ibv_cq_ex\n");
+		return -FI_EALREADY;
+	}
+
+	/* To enable additional WC fields, create CQ with efadv_create_cq verb instead. */
+	cq->ibv_cq_ex = ibv_create_cq_ex(cq->domain->device->ibv_ctx, &init_attr_ex);
+
+	return cq->ibv_cq_ex ? 0 : -FI_ENOCQ;
+}
+#endif
+
+int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
+		struct fid_cq **cq_fid, void *context)
+{
+	struct efa_cq *cq;
+	int err;
+
 	if (attr->wait_obj != FI_WAIT_NONE)
 		return -FI_ENOSYS;
 
@@ -273,9 +327,9 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 	if (!cq)
 		return -FI_ENOMEM;
 
-	ret = ofi_cq_init(&efa_prov, domain_fid, attr, &cq->util_cq,
+	err = ofi_cq_init(&efa_prov, domain_fid, attr, &cq->util_cq,
 			  &ofi_cq_progress, context);
-	if (ret) {
+	if (err) {
 		EFA_WARN(FI_LOG_CQ, "Unable to create UTIL_CQ\n");
 		goto err_free_cq;
 	}
@@ -283,18 +337,16 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 	cq->domain = container_of(domain_fid, struct efa_domain,
 				  util_domain.domain_fid);
 
-	init_attr_ex.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE;
-
-	cq->ibv_cq_ex = ibv_create_cq_ex(cq->domain->device->ibv_ctx, &init_attr_ex);
-	if (!cq->ibv_cq_ex) {
+	err = efa_cq_set_ibv_cq_ex(cq, attr);
+	if (err) {
 		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ\n");
-		ret = -FI_EINVAL;
+		err = -FI_EINVAL;
 		goto err_free_util_cq;
 	}
 
-	ret = ofi_bufpool_create(&cq->wce_pool, sizeof(struct efa_wce), 16, 0,
+	err = ofi_bufpool_create(&cq->wce_pool, sizeof(struct efa_wce), 16, 0,
 				 EFA_WCE_CNT, 0);
-	if (ret) {
+	if (err) {
 		EFA_WARN(FI_LOG_CQ, "Failed to create wce_pool\n");
 		goto err_destroy_cq;
 	}
@@ -315,7 +367,7 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 		break;
 	case FI_CQ_FORMAT_TAGGED:
 	default:
-		ret = -FI_ENOSYS;
+		err = -FI_ENOSYS;
 		goto err_destroy_pool;
 	}
 
@@ -337,5 +389,5 @@ err_free_util_cq:
 	ofi_cq_cleanup(&cq->util_cq);
 err_free_cq:
 	free(cq);
-	return ret;
+	return err;
 }
--- a/prov/efa/src/rxr/rxr_ep.c
+++ b/prov/efa/src/rxr/rxr_ep.c
@@ -1799,6 +1799,104 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
 	}
 }
 
+#if HAVE_EFADV_CQ_EX
+/**
+ * @brief Read peer raw address from EFA device and look up the peer address in AV.
+ * The caller must guarantee that the peer AH is unknown.
+ * @return Peer address, or FI_ADDR_NOTAVAIL if unavailable.
+ */
+static inline fi_addr_t rdm_ep_determine_peer_address_from_efadv(struct rxr_ep *ep,
+																 struct ibv_cq_ex *ibv_cqx)
+{
+	struct rxr_pkt_entry *pkt_entry;
+	struct efa_ep *efa_ep;
+	struct efa_ep_addr efa_ep_addr = {0};
+	fi_addr_t addr;
+	union ibv_gid gid = {0};
+	uint32_t *connid = NULL;
+
+
+	/* Attempt to read AH from EFA firmware */
+	if (efadv_wc_read_ah(efadv_cq_from_ibv_cq_ex(ibv_cqx), &gid) >= 0) {
+		/* AH should be unknown (negative) */
+		return FI_ADDR_NOTAVAIL;
+	}
+
+	pkt_entry = (void *)(uintptr_t)ibv_cqx->wr_id;
+
+	connid = rxr_pkt_connid_ptr(pkt_entry);
+	if (!connid) {
+		return FI_ADDR_NOTAVAIL;
+	}
+
+	/*
+	 * If AH is negative, use raw:qpn:connid as the key to lookup AV for peer's fi_addr
+	 */
+	memcpy(efa_ep_addr.raw, gid.raw, sizeof(efa_ep_addr.raw));
+	efa_ep_addr.qpn = ibv_wc_read_src_qp(ibv_cqx);
+	efa_ep_addr.qkey = *connid;
+	efa_ep = container_of(ep->rdm_ep, struct efa_ep, util_ep.ep_fid);
+	addr = ofi_av_lookup_fi_addr(&efa_ep->av->util_av, &efa_ep_addr);
+	if (addr != FI_ADDR_NOTAVAIL) {
+		char gid_str_cdesc[INET6_ADDRSTRLEN];
+		inet_ntop(AF_INET6, gid.raw, gid_str_cdesc, INET6_ADDRSTRLEN);
+		FI_WARN(&rxr_prov, FI_LOG_AV,
+				"Recovered peer fi_addr. [Raw]:[QPN]:[QKey] = [%s]:[%" PRIu16 "]:[%" PRIu32 "]\n",
+				gid_str_cdesc, efa_ep_addr.qpn, efa_ep_addr.qkey);
+	}
+
+	return addr;
+}
+
+/**
+ * @brief Determine peer address from ibv_cq_ex
+ * Attempt to inject or determine peer address if not available. This usually
+ * happens when the endpoint receives the first packet from a new peer.
+ * There is an edge case for EFA endpoint - the device might lose the address
+ * handle of a known peer due to a firmware bug and return FI_ADDR_NOTAVAIL.
+ * The provider needs to look up the address using Raw address:QPN:QKey.
+ * Note: This function introduces addtional overhead. It should only be called if
+ * efa_av_lookup_address_rdm fails to find the peer address.
+ * @param ep Pointer to RDM endpoint
+ * @param ibv_cqx Pointer to CQ
+ * @returns Peer address, or FI_ADDR_NOTAVAIL if unsuccessful.
+ */
+static inline fi_addr_t rdm_ep_determine_addr_from_ibv_cq_ex(struct rxr_ep *ep, struct ibv_cq_ex *ibv_cqx)
+{
+	struct rxr_pkt_entry *pkt_entry;
+	fi_addr_t addr = FI_ADDR_NOTAVAIL;
+
+	pkt_entry = (void *)(uintptr_t)ibv_cqx->wr_id;
+
+	addr = rxr_pkt_determine_addr(ep, pkt_entry);
+
+	if (addr == FI_ADDR_NOTAVAIL) {
+		addr = rdm_ep_determine_peer_address_from_efadv(ep, ibv_cqx);
+	}
+
+	return addr;
+}
+#else
+/**
+ * @brief Determine peer address from ibv_cq_ex
+ * Attempt to inject peer address if not available. This usually
+ * happens when the endpoint receives the first packet from a new peer.
+ * Note: This function introduces addtional overhead. It should only be called if
+ * efa_av_lookup_address_rdm fails to find the peer address.
+ * @param ep Pointer to RDM endpoint
+ * @param ibv_cqx Pointer to CQ
+ * @returns Peer address, or FI_ADDR_NOTAVAIL if unsuccessful.
+ */
+static inline fi_addr_t rdm_ep_determine_addr_from_ibv_cq_ex(struct rxr_ep *ep, struct ibv_cq_ex *ibv_cqx)
+{
+	struct rxr_pkt_entry *pkt_entry;
+
+	pkt_entry = (void *)(uintptr_t)ibv_cqx->wr_id;
+
+	return rxr_pkt_determine_addr(ep, pkt_entry);
+}
+#endif
+
 /**
  * @brief poll rdma-core cq and process the cq entry
  *
@@ -1857,6 +1955,11 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			break;
 		case IBV_WC_RECV:
 			pkt_entry->addr = efa_av_reverse_lookup_rdm(efa_av, ibv_wc_read_slid(efa_cq->ibv_cq_ex), ibv_wc_read_src_qp(efa_cq->ibv_cq_ex), pkt_entry);
+
+			if (pkt_entry->addr == FI_ADDR_NOTAVAIL) {
+				pkt_entry->addr = rdm_ep_determine_addr_from_ibv_cq_ex(ep, efa_cq->ibv_cq_ex);
+			}
+
 			pkt_entry->pkt_size = ibv_wc_read_byte_len(efa_cq->ibv_cq_ex);
 			assert(pkt_entry->pkt_size > 0);
 			rxr_pkt_handle_recv_completion(ep, pkt_entry, EFA_EP);
@@ -1875,6 +1978,10 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			break;
 		}
 
+		/*
+		 * ibv_next_poll MUST be call after the current WC is fully processed,
+		 * which prevents later calls on ibv_cq_ex from reading the wrong WC.
+		 */
 		err = ibv_next_poll(efa_cq->ibv_cq_ex);
 	}
 
@@ -1967,6 +2074,15 @@ static inline void rdm_ep_poll_shm_cq(struct rxr_ep *ep,
 			rxr_pkt_handle_send_completion(ep, pkt_entry);
 		} else if (cq_entry.flags & (FI_RECV | FI_REMOTE_CQ_DATA)) {
 			pkt_entry->addr = src_addr;
+
+			if (pkt_entry->addr == FI_ADDR_NOTAVAIL) {
+				/*
+				 * Attempt to inject or determine peer address if not available. This usually
+				 * happens when the endpoint receives the first packet from a new peer.
+				 */
+				pkt_entry->addr = rxr_pkt_determine_addr(ep, pkt_entry);
+			}
+
 			pkt_entry->pkt_size = cq_entry.len;
 			assert(pkt_entry->pkt_size > 0);
 			rxr_pkt_handle_recv_completion(ep, pkt_entry, SHM_EP);
--- a/prov/efa/src/rxr/rxr_pkt_cmd.c
+++ b/prov/efa/src/rxr/rxr_pkt_cmd.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2020 Amazon.com, Inc. or its affiliates.
+ * Copyright (c) 2019-2022 Amazon.com, Inc. or its affiliates.
  * All rights reserved.
  *
  * This software is available to you under a choice of one of two
@@ -1056,6 +1056,27 @@ void rxr_pkt_proc_received(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
 }
 
 /**
+ * @brief Read peer raw address from packet header and insert the peer in AV.
+ * @param ep Pointer to RDM endpoint
+ * @param pkt_entry Pointer to packet entry
+ * @returns Peer address, or FI_ADDR_NOTAVIL if the packet header does not include raw address
+ */
+fi_addr_t rxr_pkt_determine_addr(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
+{
+	struct rxr_base_hdr *base_hdr;
+
+	base_hdr = rxr_get_base_hdr(pkt_entry->pkt);
+	if (base_hdr->type >= RXR_REQ_PKT_BEGIN && rxr_pkt_req_raw_addr(pkt_entry)) {
+		void *raw_addr;
+		raw_addr = rxr_pkt_req_raw_addr(pkt_entry);
+		assert(raw_addr);
+		return rxr_pkt_insert_addr(ep, pkt_entry, raw_addr);
+	}
+
+	return FI_ADDR_NOTAVAIL;
+}
+
+/**
  * @brief handle a received packet
  *
  * @param	ep[in,out]		endpoint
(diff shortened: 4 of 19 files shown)


prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -959,103 +959,6 @@ fi_addr_t rxr_pkt_insert_addr(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry
 	return rdm_addr;
 }
 
-/**
- * @brief process a received packet
- *
- * @param[in]	ep		endpoint
- * @param[in]	pkt_entry	received packet entry
- */
-void rxr_pkt_proc_received(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
```
wenduwan, line 968: Relocate this function in order to mock it in `rxr_pkt_cmd.c`

prov/efa/src/efa_cq.c
```diff
@@ -312,7 +312,21 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 
 	init_attr_ex.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE;
 
+#ifdef SUPPORT_EX_CQ
```
wzamazon, line 315: `SUPPORT_EX_CQ` is too general. should be something like `HAVE_EFADV_CQ_EX`.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1085,34 +992,86 @@ void rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
...
+
+	/*
+	 * Attempt to inject or infer peer address if not available. This usually
+	 * happens when the endpoint receives the first packet from a new peer.
+	 * There is an edge case for EFA endpoint - the device might lose the address
+	 * handle of a known peer due to a firmware bug and return FI_ADDR_NOTAVAIL.
+	 * The provider needs to infer the address by searching the reverse AV table/map,
+	 * and re-insert the address to the AV if the peer address is found.
+	 */
+	if (pkt_entry->addr == FI_ADDR_NOTAVAIL && pkt_type >= RXR_REQ_PKT_BEGIN) {
```
wzamazon, line 1004: Do not limit the pkt_type here. the `efadv_wc_read_ah` should be available for any pkt type.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1085,34 +992,86 @@ void rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
...
+#ifdef SUPPORT_EX_CQ
+		if (lower_ep_type == EFA_EP) {
```
wzamazon, line 1006: The logic is getting super complicated here. I think it would be better to have a function for efa device and a function for shm.

prov/efa/test/efa_unit_tests.c
```diff
@@ -19,6 +19,10 @@ int main(void)
 		cmocka_unit_test(test_rdm_ep_progress_failed_poll),
 		cmocka_unit_test(test_rdm_ep_progress_bad_send_wc_status),
 		cmocka_unit_test(test_rdm_ep_progress_bad_recv_wc_status),
+#ifdef SUPPORT_EX_CQ
+		cmocka_unit_test(test_rdm_ep_progress_recover_forgotten_peer_ah),
```
wzamazon, line 23: I think in this case the testing target should not be the progress engine, but the new function that we are going to introduce that handle address.

prov/efa/src/rxr/rxr_pkt_cmd.c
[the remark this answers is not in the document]
wenduwan: In this revision I wanted to mock this function, which is not possible because it is called in the same file. Maybe with your refactor I do not need the mock anymore.

prov/efa/src/efa_cq.c
wenduwan: Sounds good. Will address.

prov/efa/src/efa_cq.c
```diff
@@ -312,7 +312,21 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+	/* To enable additional WC fields, create CQ with efadv_create_cq verb instead. */
+	struct efadv_cq_init_attr efadv_cq_init_attr = {
```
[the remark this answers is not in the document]
wenduwan: Sounds good, will address.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1085,34 +992,86 @@ void rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
 		rxr_pkt_entry_release_rx(ep, pkt_entry);
 		return;
 	}
-	
-	if (pkt_entry->addr == FI_ADDR_NOTAVAIL) {
-		if (pkt_type >= RXR_REQ_PKT_BEGIN && rxr_pkt_req_raw_addr(pkt_entry)) {
+
```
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

prov/efa/src/rxr/rxr_pkt_cmd.c
wenduwan: Do we want to skip it for shm endpoint though?

I did this check to maintain backward compatibility. The current logic only looks for raw address in packet header under this condition.

prov/efa/src/rxr/rxr_pkt_cmd.c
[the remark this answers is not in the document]
wzamazon: Thank you!

It would be ideal that we do not need to mock this function.

Mocking it will make testing it harder.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1085,34 +992,86 @@ void rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
...
+	 */
+	if (pkt_entry->addr == FI_ADDR_NOTAVAIL && pkt_type >= RXR_REQ_PKT_BEGIN) {
+#ifdef SUPPORT_EX_CQ
+		if (lower_ep_type == EFA_EP) {
+			struct efa_cq *efa_cq;
+			/* Read source GID from EFA firmware */
+			efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+
+			/* Double check the AH should be unknown (negative) in this case */
+			if (efadv_wc_read_ah(efadv_cq_from_ibv_cq_ex(efa_cq->ibv_cq_ex), &gid) < 0) {
```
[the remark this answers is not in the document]
wenduwan: This is a secondary validation for `efadv_wc_read_ah` - it should return a negative value if the AH is unknown.

Something will be wrong if AH is unknown but this function returns a non-negative code.

prov/efa/src/rxr/rxr_pkt_cmd.c
wzamazon: No, for shm. We still want to limit packet type. I think the logic is getting really complicated here, so it is better to have a function for shm and two functions for EFA (under different value of HAVE_EFADV_CQ_EX_SUPPORT).

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
 	efa_unit_test_resource_destruct(&resource);
 }
 
+#if HAVE_EFADV_CQ_EX
+static void rdm_cq_read_unknown_peer_ah(int peer_removed) {
```
wzamazon, line 291: should be `test_impl_rdm_cq_read_unknow_peer_ah`. since it implemented a test.

wzamazon (changes requested):

prov/efa/test/efa_unit_tests.c
```diff
@@ -1,22 +1,37 @@
 #include "efa_unit_tests.h"
 
+static int reset_global_mocks(void **state)
+{
+	g_ibv_send_wr_list = (struct efa_mock_ibv_send_wr_list) {
```
wzamazon, line 5: should loop through the list and release the memory.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+
+	efa_unit_test_resource_construct(&resource, FI_EP_RDM);
+
+	rxr_ep = container_of(resource.ep, struct rxr_ep, util_ep.ep_fid);
+	efa_cq = container_of(rxr_ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+
+	/* Create and register a fake peer */
+	ret = fi_getname(&resource.ep->fid, &raw_addr, &raw_addr_len);
+	assert_int_equal(ret, 0);
+	raw_addr.qpn = 0; // return value of _read_qp_num
```
wzamazon, line 315: I do not understand this comment.

prov/efa/configure.m4
```diff
@@ -108,6 +110,37 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 
 	AS_IF([test $efa_happy -eq 1 ], [$1], [$2])
 
+	AS_IF([test $efadv_support_extensible_cq -eq 1],
```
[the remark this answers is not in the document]
wenduwan: These are actually used interchangeably in rdma-core.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+	raw_addr.qkey = 0x1234;
+
+	if (!peer_removed) {
+		fi_addr_t peer_addr;
+		ret = fi_av_insert(resource.av, &raw_addr, 1, &peer_addr, 0, NULL);
```
wzamazon, line 320: The behavior does not match the description.

The description says that it test the case a peer has been removed, but `fi_av_remove` was not called.

I think we need to always call `fi_av_insert`, then call `fi_av_remove` for the removed case.

prov/efa/src/efa_cq.c
```diff
@@ -249,13 +249,42 @@ static struct fi_ops efa_cq_fi_ops = {
...
 };
 
-int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
-		struct fid_cq **cq_fid, void *context)
-{
-	struct efa_cq *cq;
-	int ret;
+
+#if HAVE_EFADV_CQ_EX
+static inline int efa_cq_attach_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
```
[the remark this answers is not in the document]
wenduwan: Sure. Will address.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+		assert_int_equal(ret, 1);
+	}
+
+	/* Setup packet entry */
+	entry = rxr_pkt_entry_alloc(rxr_ep, rxr_ep->efa_rx_pkt_pool, RXR_PKT_FROM_EFA_RX_POOL);
```
wzamazon, line 325: this part of constructing an eager rtm packet is useful, consider move it into a separate utility function.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+	efa_cq->ibv_cq_ex->read_src_qp = &efa_mock_ibv_read_src_qp_return_mock;
 
+	efadv_cq.wc_read_ah = &efa_mock_efadv_wc_read_unknown_ah;
+
+	/* Read 1 entry with unknown AH */
+	will_return(efa_mock_ibv_start_poll_return_mock, 0);
+	will_return(efa_mock_ibv_next_poll_return_mock, ENOENT);
+	will_return(efa_mock_ibv_end_poll_return_mock, NULL);
+	will_return(efa_mock_ibv_read_slid_return_mock, 0xffff);
+	will_return(efa_mock_ibv_read_byte_len_return_mock, sizeof(base_hdr) + sizeof(opt_connid_hdr));
```
wzamazon, line 361: I recommend store the size in `pkt_entry->pkt_size` during contruction of the packet. Make the packet entry more realistic.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+	will_return_maybe(efa_mock_ibv_read_opcode_return_mock, IBV_WC_RECV);
+	will_return_maybe(efa_mock_ibv_read_src_qp_return_mock, raw_addr.qpn);
+
+	/*
+	 * Make sure to call efadv efadv_wc_read_ah, but
+	 * efadv_wc_read_ah is static inline and cannot be mocked directly
+	 */
```
wzamazon, line 368: I do not understand this comment.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+	g_efa_unit_test_mocks.efadv_cq_from_ibv_cq_ex = efadv_cq_from_ibv_cq_ex_return_mock;
+	expect_value(efadv_cq_from_ibv_cq_ex_return_mock, ibvcqx, efa_cq->ibv_cq_ex);
+	will_return(efadv_cq_from_ibv_cq_ex_return_mock, &efadv_cq);
+	will_return(efa_mock_efadv_wc_read_unknown_ah, raw_addr.raw);
+
+	ret = fi_cq_read(resource.cq, &cq_entry, 1);
+	assert_int_equal(ret, -FI_EAGAIN);
+
+	ret = fi_cq_readerr(resource.cq, &cq_err_entry, 0);
```
wzamazon, line 377: This function call is unnecessary.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+	 * Make sure to call efadv efadv_wc_read_ah, but
+	 * efadv_wc_read_ah is static inline and cannot be mocked directly
+	 */
+	g_efa_unit_test_mocks.efadv_cq_from_ibv_cq_ex = efadv_cq_from_ibv_cq_ex_return_mock;
+	expect_value(efadv_cq_from_ibv_cq_ex_return_mock, ibvcqx, efa_cq->ibv_cq_ex);
+	will_return(efadv_cq_from_ibv_cq_ex_return_mock, &efadv_cq);
+	will_return(efa_mock_efadv_wc_read_unknown_ah, raw_addr.raw);
+
+	ret = fi_cq_read(resource.cq, &cq_entry, 1);
+	assert_int_equal(ret, -FI_EAGAIN);
```
wzamazon, line 375: This is not the case when peer is not removed. right? in this case we should get a completion entry.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+
+	ret = fi_cq_readerr(resource.cq, &cq_err_entry, 0);
+	assert_int_equal(ret, -FI_EAGAIN);
+
+	will_return_maybe(__wrap_ibv_destroy_ah, 0);
```
wzamazon, line 380: This is not necessary.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -130,7 +130,7 @@ int efa_mock_ibv_next_poll_use_saved_send_wr_with_mock_status(struct ibv_cq_ex *
 	return efa_mock_use_saved_send_wr(ibv_cqx, mock());
 }
 
-void efa_mock_ibv_end_poll_check_mock(struct ibv_cq_ex *ibvcqx)
+void efa_mock_ibv_end_poll_return_mock(struct ibv_cq_ex *ibvcqx)
```
wzamazon, line 133: I do not think this new name is right. This function does not return `mock`, it only call `mock`.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -155,3 +158,49 @@ int __wrap_efadv_query_device(struct ibv_context *ibv_ctx, struct efadv_device_a
...
+{
+	return mock();
+};
+
+uint32_t efa_mock_ibv_read_slid_return_mock(struct ibv_cq_ex *current)
+{
+	return mock();
+}
+
+int efa_mock_efadv_wc_read_unknown_ah(struct efadv_cq *efadv_cq, union ibv_gid *sgid)
```
wzamazon, line 178: The function name can be clear like `return_error_and_set_gid`.

prov/efa/test/efa_unit_tests.c
```diff
@@ -1,4 +1,4 @@
-#include "efa_unit_tests.h"
+prov/efa/test/efa_unit_tests.c#include "efa_unit_tests.h"
```
wzamazon, line 1: This line does not seem right.

prov/efa/test/efa_unit_tests.c
```diff
@@ -10,6 +10,9 @@ static int reset_global_mocks(void **state)
 	g_efa_unit_test_mocks = (struct efa_unit_test_mocks) {
 		.ibv_create_ah = __real_ibv_create_ah,
 		.efadv_query_device = __real_efadv_query_device,
+#if HAVE_EFADV_CQ_EX
+		.efadv_cq_from_ibv_cq_ex = __real_efadv_cq_from_ibv_cq_ex,
```
wzamazon, line 14: better always have this mock, but when `HAVE_EFADV_CQ_EX` is `0`, the function does not do anything.

prov/efa/test/efa_unit_tests.c
```diff
@@ -32,6 +35,10 @@ int main(void)
 		cmocka_unit_test_setup_teardown(test_rdm_cq_read_failed_poll, reset_global_mocks, NULL),
 		cmocka_unit_test_setup_teardown(test_rdm_cq_read_bad_send_status, reset_global_mocks, NULL),
 		cmocka_unit_test_setup_teardown(test_rdm_cq_read_bad_recv_status, reset_global_mocks, NULL),
+#if HAVE_EFADV_CQ_EX
+		cmocka_unit_test_setup_teardown(test_rdm_cq_read_recover_forgotten_peer_ah, reset_global_mocks, NULL),
```
wzamazon, line 39: better always have this test, but do not do anything if `HAVE_EFADV_CQ_EX` is 0.

wzamazon (changes requested):

wzamazon (changes requested):
I think a test case that only test `rxr_pkt_determine_addr` is needed, and might be sufficient in this case.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1055,6 +1055,109 @@ void rxr_pkt_proc_received(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
...
+	struct efa_cq *efa_cq;
+	struct efa_ep *efa_ep;
+	struct efa_ep_addr efa_ep_addr = {0};
+	int ah;
+	union ibv_gid gid = {0};
+	uint32_t *connid = NULL;
+
+	/* Attempt to read AH from EFA firmware */
+	efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+	ah = efadv_wc_read_ah(efadv_cq_from_ibv_cq_ex(efa_cq->ibv_cq_ex), &gid);
```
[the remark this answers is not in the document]
wenduwan: Thanks for pointing this out. I will rearrange the logic.

The correct sequence should be, when we process a recv WC, check `slid` and if it's unknown, try to infer it from the packet header or EFA device.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1084,34 +1187,36 @@ void rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
...
+	 * Attempt to inject or infer peer address if not available. This usually
+	 * happens when the endpoint receives the first packet from a new peer.
+	 * There is an edge case for EFA endpoint - the device might lose the address
+	 * handle of a known peer due to a firmware bug and return FI_ADDR_NOTAVAIL.
+	 * The provider needs to infer the address by searching the reverse AV table/map,
+	 * and re-insert the address to the AV if the peer address is found.
+	 */
+	if (pkt_entry->addr == FI_ADDR_NOTAVAIL) {
+#if HAVE_EFADV_CQ_EX
+		pkt_entry->addr = rxr_pkt_infer_ah(ep, pkt_entry, lower_ep_type);
```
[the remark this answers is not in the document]
wenduwan: Sounds good. Will address.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1084,34 +1187,36 @@ void rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
...
+#else
+		pkt_entry->addr = rxr_pkt_infer_ah(ep, pkt_entry);
```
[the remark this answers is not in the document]
wenduwan: Sure will address.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Sure will address.

prov/efa/test/efa_unit_test_cq.c
wenduwan: This is a left over comment. I will clean it up.

prov/efa/test/efa_unit_test_cq.c
wenduwan: That adds complexity. I will change the parameter name to `peer_in_av`.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Sure will address.

prov/efa/configure.m4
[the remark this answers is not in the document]
wzamazon: Let's use a unified term in our code base. We used "extended" in other places.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Will address.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Will remove.

prov/efa/test/efa_unit_test_cq.c
wzamazon: The test case name need to update too.

prov/efa/src/rxr/rxr_pkt_cmd.c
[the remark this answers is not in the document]
wzamazon: >The correct sequence should be, when we process a recv WC, check slid and if it's unknown, try to infer it from the packet header or EFA device.

Yes.

Also I think it is probably better to call this new function (`rxr_pkt_determine_addr`) in `rxr_ep_poll_ibv_cq` and `rxr_ep_poll_shm_cq` respectively. instead in `rxr_pkt_handle_recv`.

prov/efa/test/efa_unit_test_cq.c
wenduwan: I will take a closer look.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Will address.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Will remove.

prov/efa/test/efa_unit_test_mocks.c
wenduwan: Will address.

prov/efa/test/efa_unit_test_mocks.c
wenduwan: Will address.

prov/efa/test/efa_unit_tests.c
wenduwan: Will address.

prov/efa/test/efa_unit_tests.c
wenduwan: Will take a closer look.

prov/efa/test/efa_unit_tests.c
wenduwan: See response below.

prov/efa/test/efa_unit_tests.c
wenduwan: That is odd. The test will pass if we don't do anything, which means even if we are not handling unknown AH, the test still says it is working.

I think it should be conditional on edfadv capability.

wzamazon: The other test should say `skipped` instead of `pass`.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -155,3 +158,49 @@ int __wrap_efadv_query_device(struct ibv_context *ibv_ctx, struct efadv_device_a
...
+};
+
+struct efadv_cq *__real_efadv_cq_from_ibv_cq_ex(struct ibv_cq_ex *ibvcqx);
+
+struct efadv_cq *__wrap_efadv_cq_from_ibv_cq_ex(struct ibv_cq_ex *ibvcqx)
+{
+	return g_efa_unit_test_mocks.efadv_cq_from_ibv_cq_ex(ibvcqx);
+}
+
+struct efadv_cq *efadv_cq_from_ibv_cq_ex_return_mock(struct ibv_cq_ex *ibvcqx)
```
wzamazon, line 192: I am not sure why this mock is needed?

prov/efa/test/efa_unit_test_cq.c
wzamazon: I think the reason is that `fi_recv` was not called, so in the "recover_forgotten" case, you will get an unexpected message. so we should check the unexpected message list here.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +287,122 @@ void test_rdm_cq_read_failed_poll()
...
+	efa_cq->ibv_cq_ex->read_opcode = &efa_mock_ibv_read_opcode_return_mock;
+	efa_cq->ibv_cq_ex->read_src_qp = &efa_mock_ibv_read_src_qp_return_mock;
 
+	efadv_cq.wc_read_ah = &efa_mock_efadv_wc_read_unknown_ah;
+
+	/* Read 1 entry with unknown AH */
+	will_return(efa_mock_ibv_start_poll_return_mock, 0);
+	will_return(efa_mock_ibv_next_poll_return_mock, ENOENT);
+	will_return(efa_mock_ibv_end_poll_return_mock, NULL);
+	will_return(efa_mock_ibv_read_slid_return_mock, 0xffff);
```
wzamazon, line 360: Add a comment here that `slid` being `0xffff` means device does not know the address.

prov/efa/test/efa_unit_test_cq.c
wzamazon: parameter type should be `bool`.

prov/efa/test/efa_unit_test_mocks.c
wenduwan: We need to mock the returned `struct efadv_cq` value, and then mock `efadv_cq.wc_read_qh(...)`

wzamazon: I see. 

Can we call the real `efadv_cq_from_ibv_cq_ex` to get an `efqdv_cq`, and set the result's `wc_read_ah` to what we want?

wenduwan: That's a good idea. Then we don't need to mock this function.

prov/efa/test/efa_unit_test_common.c
```diff
@@ -153,3 +154,28 @@ void efa_unit_test_buff_destruct(struct efa_unit_test_buff *buff)
...
 	free(buff->buff);
 }
+
+/**
+ * @brief Construct RXR_EAGER_MSGRTM_PKT
+ *
+ * @param[in] pkt_entry Packet entry. Must be non-NULL.
+ * @param[in] attr Packet attributes.
+ */
+void efa_unit_test_eager_msgrtm_pkt_construct(struct rxr_pkt_entry *pkt_entry, struct efa_unit_test_eager_rtm_pkt_attr *attr)
```
wenduwan, line 164: I think we can refactor this function in the future to support other packet types.

prov/efa/src/efa_cq.c
```diff
@@ -249,13 +249,17 @@ static struct fi_ops efa_cq_fi_ops = {
...
-		struct fid_cq **cq_fid, void *context)
-{
-	struct efa_cq *cq;
-	int ret;
+
+/**
+ * @brief Create and set cq->ibv_cq_ex
+ *
+ * @param[in] cq Pointer to the efa_cq. cq->ibv_cq_ex must be NULL.
+ * @param[in] attr Pointer to fi_cq_attr.
```
wzamazon, line 257: add a not for return value.

prov/efa/src/efa_cq.c
```diff
@@ -265,6 +269,55 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
 		.comp_mask = 0,
 	};
+	struct efadv_cq_init_attr efadv_cq_init_attr = {
+		.comp_mask = 0,
+		.wc_flags = EFADV_WC_EX_WITH_AH,
+	};
+
+	if (cq->ibv_cq_ex) {
+		EFA_WARN(FI_LOG_CQ, "CQ already has attached ibv_cq_ex\n");
+		return 1;
```
wzamazon, line 279: Return a negative libfabric error code (`-FI_EALEARDY`  is good here)

prov/efa/src/efa_cq.c
```diff
@@ -265,6 +269,55 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+	}
+
+	/* To enable additional WC fields, create CQ with efadv_create_cq verb instead. */
+	cq->ibv_cq_ex = efadv_create_cq(cq->domain->device->ibv_ctx,
+									&init_attr_ex,
+									&efadv_cq_init_attr,
+									sizeof(efadv_cq_init_attr));
+
+	return !cq->ibv_cq_ex;
```
wzamazon, line 288: On failure `cq->ibv_cq-ex == NULL`, return a negative libfabric error code.

prov/efa/src/efa_cq.c
```diff
@@ -265,6 +269,55 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+		/* EFA requires these values for wc_flags and comp_mask.
+		 * See `efa_create_cq_ex` in rdma-core.
+		 */
+		.wc_flags = IBV_WC_STANDARD_FLAGS,
+		.comp_mask = 0,
+	};
+
+	if (cq->ibv_cq_ex) {
+		EFA_WARN(FI_LOG_CQ, "CQ already has attached ibv_cq_ex\n");
+		return 1;
```
wzamazon, line 306: similar comment about return value apply.

prov/efa/src/efa_cq.c
```diff
@@ -265,6 +269,55 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
+
+	return !cq->ibv_cq_ex;
+}
+#endif
+
+int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
+		struct fid_cq **cq_fid, void *context)
+{
+	struct efa_cq *cq;
+	int ret;
```
wzamazon, line 320: Use `err` here to be more clear.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1055,6 +1055,116 @@ void rxr_pkt_proc_received(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
...
+	union ibv_gid gid = {0};
+	uint32_t *connid = NULL;
+
+	efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+
+	/* Attempt to read AH from EFA firmware */
+	if (efadv_wc_read_ah(efadv_cq_from_ibv_cq_ex(efa_cq->ibv_cq_ex), &gid) >= 0) {
+		/* AH should be unknown (negative) */
+		addr = FI_ADDR_NOTAVAIL;
+		goto out;
```
wzamazon, line 1096: `return FI_ADDR_NOTAVAIL` here should be good. `goto` is used as the last resort when there are lots of resources to release.

wzamazon (changes requested):
Thank you very much for the effort!

I think this version is very nice.

I left a few comments. The biggest one is about timing of `rxr_determine_addr_from_efadv` being called.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1055,6 +1055,116 @@ void rxr_pkt_proc_received(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
...
+	struct efa_ep *efa_ep;
+	struct efa_ep_addr efa_ep_addr = {0};
+	fi_addr_t addr;
+	union ibv_gid gid = {0};
+	uint32_t *connid = NULL;
+
+	efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+
+	/* Attempt to read AH from EFA firmware */
+	if (efadv_wc_read_ah(efadv_cq_from_ibv_cq_ex(efa_cq->ibv_cq_ex), &gid) >= 0) {
```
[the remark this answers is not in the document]
wenduwan: > but there is no guarantee of it.

Could you explain with an example? Currently this code path is inline with `rdm_ep_poll_ibv_cq` and we make sure `efadv_wc_read_ah` is called before `ibv_next_poll`.

As we discussed earlier, moving this call higher up the call stack could introduce overhead, since we will essentially 1) parse packet header and 2) query efadv before looking up reverse AV.

prov/efa/test/efa_unit_tests.c
```diff
@@ -1,22 +1,37 @@
 #include "efa_unit_tests.h"
 
+static int reset_global_mocks(void **state)
```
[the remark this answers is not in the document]
wenduwan: Sure will address.

prov/efa/src/efa_cq.c
wenduwan: Will return `-FI_ENOCQ`

prov/efa/src/efa_cq.c
wenduwan: Will address.

prov/efa/src/rxr/rxr_pkt_cmd.c
wenduwan: Sure will address.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -286,4 +286,141 @@ void test_rdm_cq_read_failed_poll()
...
+
+	if (remove_peer) {
+		ret = fi_av_remove(resource.av, &peer_addr, 1, 0);
+		assert_int_equal(ret, 0);
+	}
+
+	ret = fi_cq_read(resource.cq, &cq_entry, 1);
+
+	if (remove_peer) {
+		/* Ignored WC */
```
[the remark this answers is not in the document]
wenduwan: Will rephrase.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -38,6 +38,23 @@ struct efa_mock_ibv_send_wr_list g_ibv_send_wr_list =
 	.tail = NULL,
 };
 
+/**
+ * @brief Destruct efa_mock_ibv_send_wr_list and free up memory
+ *
+ * @param[in] wr_list Pointer to efa_mock_ibv_send_wr_list
+ */
+void efa_mock_ibv_send_wr_list_destruct(struct efa_mock_ibv_send_wr_list *wr_list) {
```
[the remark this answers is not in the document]
wenduwan: Will address.

prov/efa/src/rxr/rxr_pkt_cmd.c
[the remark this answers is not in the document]
wzamazon: Yes. Today it is correct. but if someone change the call sequence in `rdm_ep_poll_ibv_cq` in the future. it became incorrect, and will the bug will be hard to debug, because `efadv_wc_read_ah` can still return correct value.

> As we discussed earlier, moving this call higher up the call stack could introduce overhead, since we will essentially 1) parse packet header and 2) query efadv before looking up reverse AV.

I think we can call this after `efa_av_lookup_address_rdm` return FI_ADDR_NOTAVAIL. Then we maintain the same logic.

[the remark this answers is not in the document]
wzamazon: > I suggest testing the function call ordering and make sure efadv_wc_read_ah is always called before ibv_next_poll.

This will not be true if we ever pull 2 CQ entries from ibv_cq.

> but I think the current code structure is cleaner.

The change to `rdm_ep_poll_ibv_cq` should be small too:

After 
```
    pkt_entry->addr = efa_av_lookup_addr_rdm(...)
```

add two lines:
```
if (pkt_entry->addr == FI_ADDR_NOTAVAIL)
    pkt_entry->addr = rxr_ep_determine_addr_from_ibv_cq_ex(rxr_ep, ibv_cq_ex);
```
> Rather than changing the critical path

Not the sure I understand, current change is also in critical path.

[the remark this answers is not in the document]
wenduwan: Addressed in new revision. Moved things around `rxr_ep.c` and `rxr_pkt_cmd.c`, such that packet address is determined before recv completion call.

wzamazon (commented):
Thank you!

Left a few comments, mainly about the comments in the code.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -1055,6 +1055,35 @@ void rxr_pkt_proc_received(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
...
+{
+	void *raw_addr;
+
+	raw_addr = rxr_pkt_req_raw_addr(pkt_entry);
+	assert(raw_addr);
+	return rxr_pkt_insert_addr(ep, pkt_entry, raw_addr);
+};
+
+
+fi_addr_t rxr_pkt_determine_addr(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
```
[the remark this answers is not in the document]
wenduwan: Will address. I will only keep `rxr_pkt_determin_addr`

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1857,6 +1940,18 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			break;
 		case IBV_WC_RECV:
 			pkt_entry->addr = efa_av_reverse_lookup_rdm(efa_av, ibv_wc_read_slid(efa_cq->ibv_cq_ex), ibv_wc_read_src_qp(efa_cq->ibv_cq_ex), pkt_entry);
+
+			if (pkt_entry->addr == FI_ADDR_NOTAVAIL) {
+				/*
+				 * Attempt to inject or determine peer address if not available. This usually
```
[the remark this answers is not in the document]
wenduwan: WIll address.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1799,6 +1799,89 @@ static inline void rxr_ep_check_peer_backoff_timer(struct rxr_ep *ep)
...
+				"Recovered peer fi_addr. [Raw]:[QPN]:[QKey] = [%s]:[%" PRIu16 "]:[%" PRIu32 "]\n",
+				gid_str_cdesc, efa_ep_addr.qpn, efa_ep_addr.qkey);
+	}
+
+	return addr;
+}
+#endif
+
+/**
+ * @brief Determine peer address from ibv_cq_ex
```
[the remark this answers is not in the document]
wenduwan: Will address.

prov/efa/src/efa_cq.c
```diff
@@ -249,13 +249,49 @@ static struct fi_ops efa_cq_fi_ops = {
...
-{
-	struct efa_cq *cq;
-	int ret;
+
+/**
+ * @brief Create and set cq->ibv_cq_ex
+ *
+ * @param[in] cq Pointer to the efa_cq. cq->ibv_cq_ex must be NULL.
+ * @param[in] attr Pointer to fi_cq_attr.
+ * @param[out] .
```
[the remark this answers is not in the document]
wenduwan: Will address.

wzamazon (approved):

wzamazon (approved):
Thank you!

PR #7857, opened 2022-06-28
prov/efa: fill tx desc before init rtm pkt.

shijin-aws:
Currently, the rxr_tx_entry_try_fill_desc is called
before `rxr_pkt_init_rtm` in rxr_pkt_init_*_medium function.
This caused a performance regression because the tx_entry->desc
was not filled before calling rxr_pkt_init_rtm, which makes
rxr_pkt_init_data_from_tx_entry do extra copy.

This patch fixes this issue by calling rxr_tx_entry_try_fill_desc
before rxr_pkt_init_rtm

### DIFF
--- a/prov/efa/src/rxr/rxr_pkt_type_req.c
+++ b/prov/efa/src/rxr/rxr_pkt_type_req.c
@@ -538,13 +538,13 @@ ssize_t rxr_pkt_init_medium_msgrtm(struct rxr_ep *ep,
 	struct rxr_medium_rtm_base_hdr *rtm_hdr;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


wzamazon (approved):
I think a unit test that call `rxr_pkt_init_medium_rtm` and verify the resulted `pkt_entry` does not use copy might be a good idea.

wzamazon:
bot:aws:retest

shijin-aws:
@wzamazon Sure, will add it within the PR.

shijin-aws:
per discussion, will open a separate PR to add unit test for this bug fix.

PR #7858, opened 2022-06-28
prov/efa: refactor unit test and cq

wzamazon:
A series of commits to refactor unit test framework.


prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -896,7 +896,7 @@ void rxr_pkt_handle_recv_error(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entr
 {
 	if (!pkt_entry->x_entry) {
 		efa_eq_write_error(&ep->util_ep, err, prov_errno);
-		rxr_pkt_entry_release_tx(ep, pkt_entry);
+		rxr_pkt_entry_release_rx(ep, pkt_entry);
```
wenduwan, line 899: Could you explain this change? Is it a bug fix?

wzamazon: Yes. this is a bug I discovered when developing this patch series.

prov/efa/test/efa_unit_test_common.c
```diff
@@ -1,10 +1,55 @@
...
+	}
+
+	hints->fabric_attr->prov_name = EFA_PROV_NAME;
+	hints->ep_attr->type = ep_type;
+
+	if (ep_type == FI_EP_DGRAM) {
+		hints->domain_attr->mr_mode |= FI_MR_LOCAL | FI_MR_ALLOCATED;
+		hints->mode |= FI_MSG_PREFIX;
+	} else {
+		hints->domain_attr->mr_mode |= FI_MR_LOCAL | FI_MR_ALLOCATED;
```
wenduwan, line 37: Should `assert_int_equal(ep_type, FI_EP_RDM)`

prov/efa/test/efa_unit_test_common.c
```diff
@@ -22,23 +67,33 @@ int efa_unit_test_resource_construct(struct efa_resource *resource)
...
+
 	ret = fi_cq_open(resource->domain, &cq_attr, &resource->cq, NULL);
 	if (ret)
 		goto err;
 
+	fi_ep_bind(resource->ep, &resource->cq->fid, FI_SEND | FI_RECV);
+
+	ret = fi_enable(resource->ep);
+	if (ret) {
+		/* Don't leak reource */
```
wenduwan, line 90: Instead, `goto err;`

prov/efa/test/efa_unit_test_common.c
```diff
@@ -1,10 +1,55 @@
...
+		hints->domain_attr->mr_mode |= FI_MR_LOCAL | FI_MR_ALLOCATED;
+		hints->mode |= FI_MSG_PREFIX;
+	} else {
+		hints->domain_attr->mr_mode |= FI_MR_LOCAL | FI_MR_ALLOCATED;
+	}
+
+	return hints;
+}
+
+int efa_unit_test_resource_construct(struct efa_resource *resource, enum fi_ep_type ep_type)
```
wenduwan, line 43: Using `(struct efa_resource *resource, enum fi_ep_type ep_type)` as input prevents the user from configuring the resources using a different hint, e.g. passing `NULL`.

For now we can ask the user to call the setup funcs separately, i.e.
```c
...
hints = efa_unit_test_alloc_hints(ep_type);
ret = efa_unit_test_resource_construct(&resource, &hints) // or the user just needs efa_unit_test_resource_construct(&resource, NULL);
...
```

prov/efa/test/efa_unit_test_common.c
```diff
@@ -77,3 +132,22 @@ void efa_unit_test_resource_destroy(struct efa_resource *resource)
...
+
+	buff->size = buff_size;
+	err = fi_mr_reg(resource->domain, buff, buff_size, FI_SEND | FI_RECV,
+			0 /*offset*/, 0 /*requested_key*/, 0 /*flags*/, &buff->mr, NULL);
+	assert_int_equal(err, 0);
+}
+
+void efa_unit_test_buff_destruct(struct efa_unit_test_buff *buff)
+{
+	fi_close(&buff->mr->fid);
```
wenduwan, line 151: Assert `fi_close(...)` returns 0. This helps detecting leak.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * @brief verify fi_cq_read() works with empty device CQ for given endpoint type
+ *
+ * When CQ is empty, fi_cq_read() should return -FI_EAGAIN.
+ *
+ * @param[in]		ep_type		endpoint type, can be FI_EP_DGRAM or FI_EP_RDM
+ */
+static
+void test_cq_read_empty_cq(enum fi_ep_type ep_type)
```
wenduwan, line 11: Since this is not an actual test, suggest removing the `test_*` prefix.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/**
+ * @brief test RDM CQ's fi_cq_read()/fi_cq_readerr() works properly when rdma-core return bad status for send.
+ *
+ * When ibv_post_send() operation failed, fi_cq_read() should return -FI_EAVAIL, which means error available.
+ * then user should call fi_cq_readerr() to get an error CQ entry that contain error code.
+ */
+static
+void test_cq_read_bad_send_status(enum fi_ep_type ep_type)
```
wenduwan, line 73: Same as above, suggest removing `test_*` prefix.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -0,0 +1,249 @@
...
+	assert_non_null(pkt_entry);
+
+	efa_cq = container_of(rxr_ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+	efa_cq->ibv_cq_ex->start_poll = &efa_mock_ibv_start_poll_return_mock;
+	efa_cq->ibv_cq_ex->end_poll = &efa_mock_ibv_end_poll_check_mock;
+	efa_cq->ibv_cq_ex->read_opcode = &efa_mock_ibv_read_opcode_return_mock;
+
+	will_return(efa_mock_ibv_start_poll_return_mock, 0);
+	will_return(efa_mock_ibv_end_poll_check_mock, NULL);
+	will_return(efa_mock_ibv_read_opcode_return_mock, IBV_WC_RECV);
```
wenduwan, line 199: This mock is unnecessary.

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -0,0 +1,249 @@
...
+	rxr_ep = container_of(resource.ep, struct rxr_ep, util_ep.ep_fid);
+
+	efa_cq = container_of(rxr_ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
+	efa_cq->ibv_cq_ex->start_poll = &efa_mock_ibv_start_poll_return_mock;
+	efa_cq->ibv_cq_ex->end_poll = &efa_mock_ibv_end_poll_check_mock;
+
+	will_return(efa_mock_ibv_start_poll_return_mock, EFAULT);
+
+	ret = fi_cq_read(resource.cq, &cq_entry, 1);
+	assert_int_equal(ret, -FI_EAGAIN);
```
wenduwan, line 239: Same as above, `FI_EAGAIN` vs `FI_EAVAIL`

prov/efa/test/efa_unit_test_device.c
```diff
@@ -0,0 +1,35 @@
...
+ */
+void test_efa_device_construct_error_handling()
+{
+	int ibv_err = 4242;
+	struct ibv_device **ibv_device_list;
+	struct efa_device efa_device = {0};
+
+	ibv_device_list = ibv_get_device_list(&g_device_cnt);
+	if (ibv_device_list == NULL) {
+		skip();
```
wenduwan, line 18: Does this test depend on EFA device? If so, I think we should assert ibv_device_list is not NULL here.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
 #include <setjmp.h>
 #include <cmocka.h>
 #include "efa.h"
+#include "efa_unit_test_mocks.h"
 
-void __real_rxr_pkt_handle_send_completion(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry);
+/* mock of rdma-core functions */
+uint8_t g_ibv_gid_mock[EFA_GID_LEN] = {0xfe, 0x80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x10, 0xcb, 0x06, 0xff, 0xfe, 0x66, 0x83, 0x2f};
```
wenduwan, line 11: Does the value of this array have any significance? A comment will be helpful.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
...
-void __real_rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
-										   struct rxr_pkt_entry *pkt_entry,
-										   enum rxr_lower_ep_type lower_ep_type);
+struct ibv_send_wr *g_ibv_send_wr_list_head;
+struct ibv_send_wr *g_ibv_send_wr_list_tail;
 
-void __wrap_rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
-										   struct rxr_pkt_entry *pkt_entry,
-										   enum rxr_lower_ep_type lower_ep_type)
+int efa_mock_ibv_post_send_save_send_wr(struct ibv_qp *qp, struct ibv_send_wr *wr,
```
wenduwan, line 30: nit This function does not call `mock()`, so consider removing 'mock' from its name. In the context of cmocka, it is tempting to do `will_return(efa_mock_ibv_post_send_save_send_wr, 0)` and later realize it does not do what the name suggests.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
...
 
-	if ((int)mock() == 4242) {
-		__real_rxr_pkt_handle_recv_completion(ep, pkt_entry, lower_ep_type);
+		memcpy(entry, wr, sizeof(struct ibv_send_wr));
+		if (!head)
+			head = entry;
+
+		if (tail)
+			tail->next = entry;
+		tail = entry;
```
wenduwan, line 50: Not understanding this while loop... did you miss `else {...}` on this line?

prov/efa/test/efa_unit_test_mocks.c
wenduwan: I don't understand this mock, could you please add a comment?

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
...
-	check_expected(pkt_entry);
-	check_expected(err);
-	check_expected(prov_errno);
+	return mock();
+}
 
-	if ((int)mock() == 4242) {
-		__real_rxr_pkt_handle_send_error(ep, pkt_entry, err, prov_errno);
+static inline
+int efa_mock_use_saved_send_wr(struct ibv_cq_ex *ibv_cqx, int status)
```
wenduwan, line 71: Same as above, consider removing 'mock' from function name.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
...
+					    uint32_t inlen)
+{
+	return mock();
 }
 
-void __real_rxr_pkt_handle_recv_completion(struct rxr_ep *ep,
-										   struct rxr_pkt_entry *pkt_entry,
-										   enum rxr_lower_ep_type lower_ep_type);
+struct ibv_send_wr *g_ibv_send_wr_list_head;
+struct ibv_send_wr *g_ibv_send_wr_list_tail;
```
wenduwan, line 28: Global vars are convenient but it could lead to undetermined state while running many tests.

When should we reset `g_ibv_send_wr_list_head/tail`? A comment would help.

prov/efa/test/efa_unit_test_mocks.c
wzamazon: I think the word `mock` is a well accepted term in area of unit testing, to indicate a function that has same signature of a real fucntion, but behave differently. so it should not be confused by the function `mock()`.

prov/efa/test/efa_unit_test_cq.c
wzamazon: I think it is still a test but an implementation, so I changed it to `test_impl`.

Any suggestion are welcome!

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -0,0 +1,249 @@
...
+
+	will_return(efa_mock_ibv_start_poll_return_mock, 0);
+	will_return(efa_mock_ibv_end_poll_check_mock, NULL);
+	will_return(efa_mock_ibv_read_opcode_return_mock, IBV_WC_RECV);
+	/* there is an assertion that called ibv_read_opcode(), which is effective in debug mode only */
+	will_return_maybe(efa_mock_ibv_read_opcode_return_mock, IBV_WC_RECV);
+	efa_cq->ibv_cq_ex->wr_id = (uintptr_t)pkt_entry;
+	efa_cq->ibv_cq_ex->status = IBV_WC_GENERAL_ERR;
+	ret = fi_cq_read(resource.cq, &cq_entry, 1);
+	assert_int_equal(ret, -FI_EAGAIN);
```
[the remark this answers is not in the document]
wzamazon: This is our current behavior. For an operation that is not tied to an TX or RX operation. we write EQ instead of CQ. So there is no CQ err available to read.

I do not think the behavior is right either. But this patch is to refactor the existing test. will need to added a separate patch to fix the behaivor and fix the test case.

I added a todo.

prov/efa/test/efa_unit_test_cq.c
wzamazon: Same reply.

prov/efa/test/efa_unit_test_common.c
wenduwan: I'm all for making user's life easier. But I prefer to do it differently. It's better to decouple resource construct from generating hint, and provide a shortcut function to the user - she can choose to use the shortcut, or do other customization.

For example, we can have 4 functions
```c
efa_unit_test_alloc_hints // same as now
efa_unit_test_resource_construct(struct efa_resource *resource, struct fi_info *hints) // Same as before, takes resource and hint
efa_unit_test_resource_enable // Binds resources and enables the endpoint
efa_unit_test_resource_setup(struct efa_resource *resource, enum fi_ep_type ep_type) // A shortcut to create hint, construct resources, bind and enable resources.
```

prov/efa/test/efa_unit_test_common.c
```diff
@@ -22,23 +67,33 @@ int efa_unit_test_resource_construct(struct efa_resource *resource)
...
 
+	fi_ep_bind(resource->ep, &resource->av->fid, 0);
+
 	ret = fi_cq_open(resource->domain, &cq_attr, &resource->cq, NULL);
 	if (ret)
 		goto err;
 
+	fi_ep_bind(resource->ep, &resource->cq->fid, FI_SEND | FI_RECV);
+
+	ret = fi_enable(resource->ep);
```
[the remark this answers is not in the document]
wenduwan: `fi_enable` could require additional mocks as we add more tests, e.g. rdma-core.

If a user simply wants to test a component that does not an enabled endpoint, she will still be required to setup the mocks, which can be repetitive.

Same as my comment above, I support the idea of hiding nuanced details for common test cases, but we should do that by composing rather than embedding smaller functions .

prov/efa/test/efa_unit_test_cq.c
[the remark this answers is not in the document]
wenduwan: Thanks for clarifying. I agree this should be addressed separately.

prov/efa/test/efa_unit_test_device.c
wenduwan: > It was trying to skip the test on a system without efa device. 

Maybe we should do this for all unit tests since currently they all require EFA device. It's odd to make this test special. Skipping it might hide real problems, i.e. EFA device is not registered properly and thus not discoverable by `ibv_get_device_list`.

prov/efa/test/efa_unit_test_av.c
```diff
@@ -0,0 +1,79 @@
...
+
+	err = efa_unit_test_resource_construct(&resource, FI_EP_RDM);
+	assert_int_equal(err, 0);
+	g_efa_unit_test_mocks.ibv_create_ah = &efa_mock_ibv_create_ah_increase_call_counter;
+
+	memcpy(raw_addr.raw, g_ibv_gid_mock, EFA_GID_LEN);
+	raw_addr.qpn = 0;
+	raw_addr.qkey = 1234;
+
+	ibv_create_ah_call_counter_before_insert = g_ibv_create_ah_call_counter;
```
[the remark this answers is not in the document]
wenduwan: > we want to make sure ibv_create_ah is not called by fi_av_insert

With cmocka, we guarantee this by not setting `will_return*` on `__wrap_ibv_create_ah`. If `ibv_create_ah` is called once we will get an error complaining about unmocked function.

prov/efa/Makefile.include
```diff
@@ -96,26 +96,20 @@ TESTS += prov/efa/test/efa_unit_test
 nodist_prov_efa_test_efa_unit_test_SOURCES = \
 	prov/efa/test/efa_unit_tests.h \
```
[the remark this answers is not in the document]
wenduwan: Sure.

prov/efa/src/efa_cq.c
```diff
@@ -110,30 +85,30 @@ ssize_t efa_cq_readerr(struct fid_cq *cq_fid, struct fi_cq_err_entry *entry,
...
 
-static void efa_cq_read_msg_entry(struct efa_wc *wc, int i, void *buf)
+static void efa_cq_read_msg_entry(struct ibv_cq_ex *ibv_cqx, int i, void *buf)
 {
 	struct fi_cq_msg_entry *entry = buf;
 
-	entry[i].op_context = (void *)(uintptr_t)wc->ibv_wc.wr_id;
-	entry[i].flags = efa_cq_wc_to_fi_flags(wc);
-	entry[i].len = (uint64_t)wc->ibv_wc.byte_len;
+	entry[i].op_context = (void *)(uintptr_t)ibv_cqx->wr_id;
```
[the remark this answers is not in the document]
wenduwan: Sounds good. I noticed that we are not using type cast consistently especially for `ibv_cqx->wr_id`; in other places we just did `(void *)ibv_cqx->wr_id`

prov/efa/test/efa_unit_test_common.c
wzamazon: > It's better to decouple resource construct from generating hint

Why is it better?

What customization is needed?

The proposed solution introduce 3 functions: construct, enable, setup. I think user will be really confused about which one to use.

prov/efa/test/efa_unit_test_cq.c
wzamazon: Why?

prov/efa/test/efa_unit_test_mocks.c
wzamazon: Removed this array, and use self address instead.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Because right after this there is `will_return_maybe(efa_mock_ibv_read_opcode_return_mock, IBV_WC_RECV);`

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
...
 {
-	check_expected(ep);
-	check_expected(pkt_entry);
-	check_expected(lower_ep_type);
+	struct ibv_send_wr *head, *tail, *entry;
+
+	head = tail = NULL;
+
+	while(wr) {
+		entry = calloc(sizeof(struct ibv_send_wr), 1);
```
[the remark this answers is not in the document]
wzamazon: No. The memory is saved in global variable `g_ibv_send_wr_list`, and used by another function `efa_mock_use_saved_send_wr`.

prov/efa/test/efa_unit_test_mocks.c
wzamazon: Added a comment.

prov/efa/test/efa_unit_test_mocks.c
wzamazon: No. This loop is adding a linked list to an existing list.

`tail` is the tail of the existing list.

if tail is not NULL, then it is pointing to current tail. we need to update it so it is pointing to the new entry.

then we will always make `tail` to point to the new entry.

prov/efa/test/efa_unit_test_mocks.c
wzamazon: same reply.

prov/efa/test/efa_unit_test_mocks.c
wzamazon: added a comment.

Also added code to verify memory is released.

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -5,68 +5,128 @@
...
+struct ibv_ah *__wrap_ibv_create_ah(struct ibv_pd *pd, struct ibv_ah_attr *attr)
+{
+	return g_efa_unit_test_mocks.ibv_create_ah(pd, attr);
+}
+
+int __wrap_efadv_query_device(struct ibv_context *ibv_ctx, struct efadv_device_attr *attr,
+			      uint32_t inlen)
+{
+	return g_efa_unit_test_mocks.efadv_query_device(ibv_ctx, attr, inlen);
 }
```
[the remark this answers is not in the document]
wzamazon: the `rxr_pkt` mocks are no longer required.

prov/efa/test/efa_unit_tests.c
```diff
@@ -5,20 +5,18 @@ int main(void)
...
-		cmocka_unit_test(test_rdm_ep_progress_send_empty_cq),
-		cmocka_unit_test(test_rdm_ep_progress_failed_poll),
-		cmocka_unit_test(test_rdm_ep_progress_bad_send_wc_status),
-		cmocka_unit_test(test_rdm_ep_progress_bad_recv_wc_status),
+		cmocka_unit_test(test_dgram_cq_read_empty_cq),
+		cmocka_unit_test(test_dgram_cq_read_bad_wc_status),
+		cmocka_unit_test(test_rdm_cq_read_empty_cq),
+		cmocka_unit_test(test_rdm_cq_read_failed_poll),
+		cmocka_unit_test(test_rdm_cq_read_bad_send_status),
+		cmocka_unit_test(test_rdm_cq_read_bad_recv_status),
```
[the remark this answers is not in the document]
wzamazon: No. As explained in the commit message. The happy path is tested extensively by fabtests.

prov/efa/test/efa_unit_test_cq.c
wzamazon: I think I need to use `will_return_always` because I want to make sure it is called at least once.

prov/efa/test/efa_unit_test_mocks.c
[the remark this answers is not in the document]
wzamazon: Notice that

 `tail->next = entry`

which append `entry` to `g_send_wr_list`. 

Then in `efa_mock_use_saved_send_wr`, entry is retrieved and freed there.

prov/efa/test/efa_unit_test_mocks.c
wenduwan: I'm still not getting it:
```c
		if (tail)
			tail->next = entry;

		tail = entry;
```
It reads to me that, if `tail` is set, we will have `tail = tail->next = entry`. Do you mean `tail->next = head` instead?

prov/efa/test/efa_unit_test_common.c
wenduwan: Better in the sense of having a single responsibility for a given function, e.g. `*_construct` only does resource creation, `*_enable` does what the name suggests - enabling the resources, etc. Provided that we document the code properly, the user should be able to understand with minimal effort.

prov/efa/test/rdma_core_mocks.h
```diff
@@ -0,0 +1,6 @@
+#ifndef EFA_UNIT_TEST_RDMA_CORE_MOCKS_H
+#define EFA_UNIT_TEST_RDMA_CORE_MOCKS_H
+
+extern int g_ibv_create_ah_call_counter;
```
wenduwan, line 4: Can we delete this file?

wenduwan:
Waiting for CI to finish. I think we can delete `prov/efa/test/rdma_core_mocks.h`. Otherwise looking good.

wzamazon:
>  I think we can delete prov/efa/test/rdma_core_mocks.h.

Thank you! Yes, it should be removed. I did that in new revision.

wenduwan (approved):

PR #7861, opened 2022-06-30
runfabtests: support an argument to specify junit report verbosity

AnasMasoud:
This change adds the --junit-logging argument which is used
to specify the verbosity of the junit report in case the latter
is enabled.

Signed-off-by: Anas Mousa <anasmous@amazon.com>


fabtests/scripts/runfabtests.py
```diff
@@ -294,6 +297,7 @@ def main():
                         help="only run tests which match the given substring expression.")
     parser.add_argument("--html", type=str, help="path to generated html report")
     parser.add_argument("--junit-xml", type=str, help="path to generated junit xml report")
+    parser.add_argument("--junit-logging", type=str, help="Write captured log messages to JUnit report: one of no|log|system-out|system-err|out-err|all")
```
wzamazon, line 300: Thank you!

The function `add_argument` has an argument `choices`, which allow you to list possible values for this argument. It seems perfect to use it here.

https://docs.python.org/3/library/argparse.html#choices

AnasMasoud: Done, thanks!

wzamazon (approved):

wzamazon:
Looks good! wait for CI to finish.

wzamazon:
@AnasMasoud

Sorry. One more thing, can you change the title to `fabtests/runfabtests.py: xxxx`. Because there are currently two scripts to run fabtests: `runfabtests.py` and `runfabtests.sh`. This only change `runfabtests.py`.

shefty:
@wzamazon - No, it does not.  The Intel CI frequently fails if a PR is updated quickly before the previous CI finishes, forcing it to abort, so a new run can start.  The failure here is with
```
fi_rdm -p "verbs;ofi_rxd"
```
not sure why

wzamazon:
I see. We will try to avoid that in the future. Ignore this error because the error is not related to the PR.

PR #7867, opened 2022-07-02
prov/efa: fix runtread bug and test

wzamazon:
This PR fixed a bug in that can cause runtread message protocol to hang, and added test case for it.

### DIFF
--- a/prov/efa/src/rxr/rxr_op_entry.c
+++ b/prov/efa/src/rxr/rxr_op_entry.c
@@ -246,7 +246,7 @@ void rxr_tx_entry_set_max_req_data_size(struct rxr_ep *ep, struct rxr_tx_entry *
 	int max_req_data_capacity;
 	int mulreq_total_data_size;
 	int num_req;
-	static int MEMORY_ALIGNMENT = 8;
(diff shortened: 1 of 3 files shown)


prov/efa/src/rxr/rxr_op_entry.c
```diff
@@ -258,7 +258,11 @@ void rxr_tx_entry_set_max_req_data_size(struct rxr_ep *ep, struct rxr_tx_entry *
 
 	num_req = (mulreq_total_data_size - 1)/max_req_data_capacity + 1;
 
-	tx_entry->max_req_data_size = ofi_get_aligned_size((mulreq_total_data_size - 1)/num_req + 1, MEMORY_ALIGNMENT);
+	if (efa_mr_is_cuda(tx_entry->desc[0])) {
```
shijin-aws, line 261: Was it a bug before as well?

wzamazon: Yes. It only impact performance though, will not cause hang.

shijin-aws (approved):

PR #7868, opened 2022-07-02
prov/efa,fabtests/pytest: fix a bug in pytest that hide failure

wzamazon:
This PR contains 2 patches:

1st one fixed a bug in pytest script that hide a testing failure.
2nd one fixed a bug in EFA provider that can cause the failure.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -136,9 +136,9 @@ ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 	bool should_end_poll = false;
 	struct efa_cq *cq;
 	struct efa_av *av;
-	struct efa_wc wc = {0};
 	ssize_t err = 0;
 	size_t num_cqe = 0; /* Count of read entries */
+	uint32_t qp_num, src_qp, slid;
 
 	/* Initialize an empty ibv_poll_cq_attr struct for ibv_start_poll.
 	 * EFA expects .comp_mask = 0, or otherwise returns EINVAL.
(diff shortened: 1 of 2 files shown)


wzamazon:
AWS CI is expected to fail, because is because I have not pushed the 2nd commit, and the 1st commit fixed the bug in testing script that will hide a failure.

wzamazon:
AWS CI failed as expected.

Now I will add the 2nd commit, which should fix the failure.

wzamazon:
2nd commit added. Ready for review.

prov/efa/src/efa_cq.c
```diff
@@ -160,12 +160,12 @@ ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 		}
 
 		if (src_addr) {
-			av = cq->domain->qp_table[wc.ibv_wc.qp_num &
-				cq->domain->qp_table_sz_m1]->ep->av;
+			qp_num = ibv_wc_read_qp_num(cq->ibv_cq_ex);
+			src_qp = ibv_wc_read_src_qp(cq->ibv_cq_ex);
+			slid = ibv_wc_read_slid(cq->ibv_cq_ex);
+			av = cq->domain->qp_table[qp_num & cq->domain->qp_table_sz_m1]->ep->av;
```
shijin-aws, line 166: so, before the fix, qp_num, src_qp, and slid was actually zero? Which test is failing due to this bug?

wzamazon: Yes. all of them were 0.

The `fi_dgram_pingpong` test was failing because of this bug.

wzamazon: But `runfabtests.py` did not report the failure due to the bug in `common.py`

shijin-aws: could you explain why it only impacts dgram?

wzamazon: only dgram endpoint call the function `efa_cq_readfrom`.

rdm endpoint call `ofi_cq_read` which call `rxr_ep_progress` then call `ibv_start_poll` and `ibv_end_poll` directly, bypassing `efa_cq_readfrom`.

BTW, we should rename the file `efa_cq.c` to `efa_dgram_cq.c` and all functions in this file `efa_dgram_xxx`.

wzamazon:
AWS CI passed as expected.

shijin-aws (approved):

PR #7869, opened 2022-07-03
prov/net: Rework active socket monitoring

shefty:
From the commit message changing pollfds:

    core/pollfds: Add ability to manually select hot/cold sockets
    
    The pollfds implementation maintains 2 arrays of fd's that it
    monitors.  One is the default fd set of all watched fd's.  The
    other is an array of active fd's.  The active fd set is disabled
    by default, but can be enabled through the FI_POLL_FAIRNESS
    variable.  An fd is moved to the active list if it had any events
    on it within the last 'fairness' number of checks.  An fd is
    removed from the active list if no events were reported.
    
    In theory, this sounds like a great idea.  In practice, it does
    not work well with some applications.
    
    The issue is that there are situations where an application is
    only trying to send or receive data from a small number of sockets,
    maybe as few as 1, to the exclusion of all other traffic.  When
    this occurs, sockets that have receive data sitting on them will
    move to the active list and continue to generate events.  But
    because the application isn't trying to receive data from the
    corresponding peers (i.e. no tagged buffers are posted using
    any source or the peer's source address), the sockets will always
    remain on the active list.
    
    The performance of poll depends on the number of sockets passed
    to it and the number of events being reported.  When there are
    many sockets generating events repeatedly, the application
    performance suffers.  This is seen running MPI with a simple
    pingpong test.  As the number of ranks increase, the pingpong
    latency increases dramatically, even though only 2 ranks are
    ever sending data to each other.  The other ranks are waiting in
    the barrier phase, have sent data to the ranks running the pingpong
    test, but the received data is not being handled.  The pending
    data will be ignored until the pingpong ranks move to the barrier
    phase.
    
    Define a mechanism that allows a provider to directly specify if
    a socket should be moved/removed from the active list.
    
This series reworks how sockets are determined to be active and moves the control from the pollfds abstraction to the user.  The net provider is updated to list sockets as active only if it recently sent or received data on that socket.

This series is work in progress, with minimal testing.  Posting the PR for feedback and to share the changes.

### DIFF
--- a/prov/net/src/xnet_progress.c
+++ b/prov/net/src/xnet_progress.c
@@ -46,6 +46,21 @@
 static ssize_t (*xnet_start_op[ofi_op_write + 1])(struct xnet_ep *ep);
 
 
+static inline void xnet_active_ep(struct xnet_ep *ep)
+{
+	ep->progress_cnt++;
+	if (ep->is_active || !ofi_poll_fairness)
+		return;
+
+	ofi_pollfds_hotfd(xnet_ep2_progress(ep)->pollfds, ep->bsock.sock);
+	if (xnet_need_rx(ep)) {
+		dlist_remove_init(&ep->need_rx_entry);
+		dlist_insert_tail(&ep->need_rx_entry,
+				  &xnet_ep2_progress(ep)->rx_poll_list);
+	}
+	ep->is_active = true;
+}
+
 static ssize_t xnet_send_msg(struct xnet_ep *ep)
 {
 	struct xnet_xfer_entry *tx_entry;
@@ -71,7 +86,7 @@ static ssize_t xnet_send_msg(struct xnet_ep *ep)
 		len = ret;
 	}
 
-	ep->progress_cnt++;
+	xnet_active_ep(ep);
 	ep->cur_tx.data_left -= len;
 	if (ep->cur_tx.data_left) {
 		ofi_consume_iov(tx_entry->iov, &tx_entry->iov_cnt, len);
@@ -94,7 +109,7 @@ static ssize_t xnet_recv_msg_data(struct xnet_ep *ep)
 	if (ret < 0)
 		return ret;
 
-	ep->progress_cnt++;
+	xnet_active_ep(ep);
 	ep->cur_rx.data_left -= ret;
 	if (!ep->cur_rx.data_left)
 		return FI_SUCCESS;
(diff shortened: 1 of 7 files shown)


shefty:
Note that most of the changes are in the net provider, so if this passes CI, the changes should be okay to merge.  The pollfds changes should only come into play if the FI_POLL_FAIRNESS environment variable is set.

src/common.c
```diff
@@ -1685,63 +1701,61 @@ static void ofi_pollfds_process_work(struct ofi_pollfds *pfds)
...
-			}
-			ret--;
-		}
+	/* If we remove an fd, we swap it with the one at the end of the
+	 * array.  Walk the array backwards to avoid skipping entries.
+	 */
+	ofi_mutex_lock(&pfds->lock);
+	for (i = pfds->hot_nfds - 1; i >= 0; i--) {
+		ctx = ofi_pollfds_get_ctx(pfds, pfds->hot_fds[i].fd);
+		if (!is_hot(ctx))
```
ooststep, line 1720: i believe this should be `if (!is_hot(ctx->context))`

prov/net/src/xnet_progress.c
```diff
@@ -46,6 +46,21 @@
 static ssize_t (*xnet_start_op[ofi_op_write + 1])(struct xnet_ep *ep);
 
 
+static inline void xnet_active_ep(struct xnet_ep *ep)
+{
+	ep->progress_cnt++;
+	if (ep->is_active || ofi_poll_fairness)
```
ooststep, line 52: I believe this should be `if (ep->is_active || ofi_poll_fairness)`, otherwise if ep is not active and ofi_poll_fairness is set, we will still never activate the ep.

shefty: I *think* you meant ```if (ep->is_active || !ofi_poll_fairness)``` ?

shefty:
Updated based on review comments.

prov/net/src/xnet_progress.c
ooststep: Oops, yes. I did.

shefty:
bot:aws:retest

shefty:
Intel CI failure was:
```
fi_dgram -p "psm3"
11:02:15    timestamp: 20220727-180203+0000
11:02:15    result: Fail
```
which I'm going to ignore as unrelated.

shefty:
bot:aws:retest

PR #7871, opened 2022-07-05
Reduce log messages

aakefbs:
I also wonder if it should enter ofi_bsock_async_done() at all, when zerocopy was never enabled.


j-xiong (approved):

aakefbs:
Out of interest, how do you enable it at run time? 
What could be done is to set either into bsock or the tcp ep a bool if it was ever enabled.

j-xiong:
It's controlled by bsock->zerocopy_size, which is a threshold for using zero-copy.

PR #7874, opened 2022-07-07
prov/efa: validate ep->util_ep before closing in efa_ep_destroy()

wzamazon:
In efa_ep_destroy(), ofi_endpoint_close() was called on
ep->util_ep unconditionally, even when ep->util_ep has not
been initialized.

This patch check whether ep->util_ep is initialized before
close it

### DIFF
--- a/prov/efa/src/efa_ep.c
+++ b/prov/efa/src/efa_ep.c
@@ -228,14 +228,20 @@ err:
 
 static void efa_ep_destroy(struct efa_ep *ep)
 {
+	int err;
+
 	if (ep->self_ah)
 		ibv_destroy_ah(ep->self_ah);
 
 	efa_ep_destroy_qp(ep->qp);
 	fi_freeinfo(ep->info);
 	free(ep->src_addr);
-	if (ofi_endpoint_close(&ep->util_ep))
-		FI_WARN(&efa_prov, FI_LOG_EP_CTRL, "Unable to close util EP\n");
+	if (ep->util_ep_initialized) {
+		err = ofi_endpoint_close(&ep->util_ep);
+		if (err)
+			FI_WARN(&efa_prov, FI_LOG_EP_CTRL, "Unable to close util EP\n");
+	}
+
 	free(ep);
 }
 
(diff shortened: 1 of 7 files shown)


wzamazon:
Added a unit test to test the fix.

prov/efa/src/efa_ep.c
```diff
@@ -228,14 +228,20 @@ static struct efa_ep *efa_ep_alloc(struct fi_info *info)
...
+
 	if (ep->self_ah)
 		ibv_destroy_ah(ep->self_ah);
 
 	efa_ep_destroy_qp(ep->qp);
 	fi_freeinfo(ep->info);
 	free(ep->src_addr);
-	if (ofi_endpoint_close(&ep->util_ep))
-		FI_WARN(&efa_prov, FI_LOG_EP_CTRL, "Unable to close util EP\n");
+	if (ep->util_ep.domain) {
```
mjkoop, line 239: It feels odd to check for the value of the domain to decide if we need to call `ofi_endpoint_close`. 
The base error that we are addressing here seems to occur since we call `efa_ep_destroy` when `ofi_endpoint_init` fails in `efa_ep_open`. It seems we should be able to track ourselves if that call was successful (and thus if we need to call the ofi-level endpoint close).

prov/efa/test/efa_unit_test_info.c
```diff
@@ -0,0 +1,42 @@
...
+
+	err = fi_fabric(info->fabric_attr, &fabric, NULL);
+	assert_int_equal(err, 0);
+
+	err = fi_domain(fabric, info, &domain, NULL);
+	assert_int_equal(err, 0);
+
+	/* because of the error in the info object, fi_endpoint() should fail with -FI_ENODATA */
+	err = fi_endpoint(domain, info, &ep, NULL);
+	assert_int_equal(err, -FI_ENODATA);
```
wenduwan, line 34: Should we make the return code `FI_EINVAL` instead?

`FI_ENODATA` is odd - why does the program complain about no data trying to open an endpoint.

prov/efa/src/efa_ep.c
wzamazon: Make sense. I will add a flag in `efa_ep` to track `ofi_endpoint_init` status.

wenduwan (approved):

wzamazon:
waiting for CI.

wzamazon:
Because this is a pure EFA change, skip intel CI.

PR #7877, opened 2022-07-07
hmem_cuda: toggle gdrcopy/cudaMemcpy according to device.

shijin-aws:
Currently, hmem_cuda_use_gdrcopy is a global variable and set once
during the hmem init, and cuda_copy_to_dev/from_dev will determine
whether to use gdrcopy according to this global variable only. When
it is using gdrcopy, the `device` union should be a device handler,
otherwise it is a device id.

But providers may want to toggle gdrcopy/cudaMemcpy
during runtime dynamically. For example, efa provider can use gdrcopy
to copy data between bounce buffer and the device buffer in inter-node
communication. When using CUDA IPC for intra-node communication, shm
provider should still use cudaMemcpy because gdrcopy does not support
device-to-device copy.

This patch allows cuda_copy_*_dev toggles between gdrcopy and cudaMemcpy
by checking whether the `device` union in the argument is a device id or
a cuda device handler.
use cudaMemcpy.


shijin-aws:
@shefty May I know what's the error in Intel CI?

j-xiong (approved):

j-xiong:
The same issue that have been showing up recently:
```
14:03:08  + cd /opt/jenkinsagent/workspace/ofi_libfabric_PR-7877/py_scripts/contrib/intel/jenkins/
14:03:08  /jenkinsagent/workspace/ofi_libfabric_PR-7877@tmp/durable-32c5368c/script.sh: line 4: cd: /opt/jenkinsagent/workspace/ofi_libfabric_PR-7877/py_scripts/contrib/intel/jenkins/: No such file or directory
```
@nikhilnanal?

zachdworkin:
@j-xiong I think the first job got "aborted" at an awkward time due to a force push that triggered the second job telling the first job to abort. Because the first job got aborted at an awkward time it somehow went into the "on success" part of the post clause which deletes the workspace directorty that the second job was actively using. So when the second job went to look for a file to use it couldn't find it resulting in the error.

A fix for this can be to remove the post workspace deletion which is only there to reduce memory consumed by jenkins before the manager thing can go through and remove unneccessary jenkins jobs that are "too old".

j-xiong:
@zachdworkin Thanks. The issue could be avoided if a unique work directory is used for each run.

j-xiong:
The failure now is:
```
15:26:43  - name:   fi_rdm -p "verbs;ofi_rxd"
15:26:43    timestamp: 20220707-222643+0000
15:26:43    result: Fail
15:26:43    time:   301
15:26:43    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-7877/3/reg/bin/fi_rdm -p "verbs;ofi_rxd"   -s ci3-ib0
15:26:43    server_stdout: |
15:26:43    client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-7877/3/reg/bin/fi_rdm -p "verbs;ofi_rxd"   -s ci4-ib0 ci3-ib0
15:26:43    client_stdout: |
```

j-xiong:
Restarted Intel CI. The failure could be temporary.

aingerson:
This issue has been popping up more and more often. I will take a look and try to resolve it.

PR #7884, opened 2022-07-14
print ofi_mr_map_remove return correctly upon failure

ryanamazon:
When ofi_mr_map_remove returns a failure return code, that code is not
the one that is included in the debug message.  This makes it so that
the returned code is logged.

Signed-off-by: Ryan Hankins <rqh@amazon.com>


wzamazon (approved):

wzamazon:
Thank you!

j-xiong:
Please add "prov/efa: " prefix to the title of commit message since it's a efa specific fix.

wzamazon (changes requested):
Please adjust the commit message as @j-xiong suggested

j-xiong (approved):

PR #7886, opened 2022-07-15
prov/efa: Avoid excessive warn logging

darrylabbate:
This mainly prevents log files from becoming massively bloated when FI_LOG_LEVEL=warn

### DIFF
--- a/prov/efa/src/rxr/rxr_pkt_cmd.c
+++ b/prov/efa/src/rxr/rxr_pkt_cmd.c
@@ -613,7 +613,7 @@ void rxr_pkt_handle_data_copied(struct rxr_ep *ep,
  *         If user wants to manager RNR by itself (FI_RM_DISABLED),
  *         an error CQ entry will be written.
  *
- *         Otherwise, the packet will be queued and resnt by progress engine.
+ *         Otherwise, the packet will be queued and resent by progress engine.
  *
  *    For other type of error, an error CQ entry is written.
  *
@@ -625,12 +625,12 @@ void rxr_pkt_handle_data_copied(struct rxr_ep *ep,
  *
  *      For other types of error, an error CQ entry is written.
  *
- * If the packet is not associated with an user operation (such packet include
+ * If the packet is not associated with a user operation (such packet include
  * HANDSHAKE):
  *
  *      If the error is RNR, the packet is queued and resent by progress engine.
  *
- *      For othre types of error, an error EQ entry is written.
+ *      For other types of error, an error EQ entry is written.
  *
  * @param[in]	ep		endpoint
  * @param[in]	pkt_entry	pkt entry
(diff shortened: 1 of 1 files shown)


prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -646,9 +646,13 @@ void rxr_pkt_handle_send_error(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entr
...
 	       pkt_entry->alloc_type == RXR_PKT_FROM_SHM_TX_POOL);
 
-	FI_WARN(&rxr_prov, FI_LOG_CQ,
-			"Packet send error: %s (%d)\n",
-			efa_strerror(prov_errno), prov_errno);
+	/* Avoid excessive warn logging */
+	if (err != FI_EAGAIN || prov_errno != FI_EFA_REMOTE_ERROR_RNR) {
+		FI_WARN(&rxr_prov, FI_LOG_CQ,
+				"Packet send error: %s (%d)\n",
+				efa_strerror(prov_errno), prov_errno);
```
shijin-aws, line 653: I think we should still log RNR in the info/debug level.

wzamazon: I am not sure.

I do not think we need to print anything here. For real error, this function will call `rxr_cq_write_rx_error` and `rxr_cq_write_tx_error`, which printed the same information. So I do not think it is useful to print here.

prov/efa/src/rxr/rxr_pkt_cmd.c
```diff
@@ -890,9 +894,13 @@ void rxr_pkt_handle_send_completion(struct rxr_ep *ep, struct rxr_pkt_entry *pkt
...
 {
-	FI_WARN(&rxr_prov, FI_LOG_CQ,
-			"Packet receive error: %s (%d)\n",
-			efa_strerror(prov_errno), prov_errno);
+	/* Avoid excessive warn logging */
+	if (err != FI_EAGAIN) {
+		FI_WARN(&rxr_prov, FI_LOG_CQ,
+				"Packet receive error: %s (%d)\n",
+				efa_strerror(prov_errno), prov_errno);
+	}
```
wzamazon, line 902: Same as about, for real errors, this function eventually call `rxr_cq_write_rx_error`or `efa_ep_write_error` so this print is not useful.

prov/efa/src/rxr/rxr_pkt_cmd.c
darrylabbate: So just remove altogether?

wzamazon: Removal would be my preference, or change to `FI_DEBUG` like @shijin-aws suggested.

shijin-aws: Then let's just change `FI_WARN` to `FI_DEBUG` for these two prints.

shijin-aws (approved):

wzamazon (approved):

PR #7890, opened 2022-07-20
fabtests/pytest: Skip cuda tests if provider does not support hmem hints

wckzhang:

### DIFF
--- a/fabtests/Makefile.am
+++ b/fabtests/Makefile.am
@@ -66,7 +66,8 @@ bin_PROGRAMS = \
 	multinode/fi_multinode	\
 	multinode/fi_multinode_coll \
 	component/sock_test \
-	regression/sighandler_test
+	regression/sighandler_test \
+	common/check_hmem
 
 dist_bin_SCRIPTS = \
 	scripts/runfabtests.sh \
@@ -452,6 +453,14 @@ regression_sighandler_test_SOURCES = \
 regression_sighandler_test_LDADD = libfabtests.la
 
 regression_sighandler_test_CFLAGS = \
+	$(AM_CFLAGS)
+
+common_check_hmem_SOURCES = \
+	common/check_hmem.c
+
+common_check_hmem_LDADD = libfabtests.la
+
+common_checK_hmem_CFLAGS = \
 	$(AM_CFLAGS)
 
 real_man_pages = \
--- a/fabtests/common/check_hmem.c
+++ b/fabtests/common/check_hmem.c
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2022, Amazon.com, Inc.  All rights reserved.
+ *
+ * This software is available to you under the BSD license
+ * below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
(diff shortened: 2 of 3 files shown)


fabtests/Makefile.am
```diff
@@ -454,6 +455,14 @@ regression_sighandler_test_LDADD = libfabtests.la
 regression_sighandler_test_CFLAGS = \
 	$(AM_CFLAGS)
 
+common_check_hmem_SOURCES = \
+	common/check_hmem.c
```
shijin-aws, line 459: can we directly use the `fi_info` bin from libfabric?

wckzhang: we could, but I wasn't sure if there was a way to guarantee it was the libfabric that fabtests was built against. There's probably a way to do so I can try looking into it

wckzhang: I'd guess we can use the configure --prefix info, I'm not sure though if we can pass the configure/prefix info to python

fabtests/common/check_hmem.c
```diff
@@ -0,0 +1,61 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * This test returns whether or not efa supports FI_HMEM
```
tschuett, line 29: This works for any provider. The name is passed as `argv[1]`.

fabtests/pytest/common.py
```diff
@@ -1,9 +1,15 @@
 import pytest
 import errno
 
-def has_cuda(ip):
+def has_cuda(cmdline_args, host_type, ip):
```
wzamazon, line 4: the function name should be changed to `has_hmem_support`.

fabtests/pytest/common.py
```diff
@@ -198,14 +204,14 @@ def prepare_base_command(self, command_type, executable,
         # in communication test, client is sender, server is receiver
         client_memory_type,server_memory_type = memory_type.split("_to_")
         if command_type == "server" and server_memory_type == "cuda":
-            if not has_cuda(self._cmdline_args.server_id):
+            if not has_cuda(self._cmdline_args, "server", self._cmdline_args.server_id):
                 pytest.skip("no cuda device")
```
wzamazon, line 208: The skip message should be updated to `no hmem support`.

fabtests/Makefile.am
shijin-aws: Yeah, I think it is not straightforward to use it. Do not bother then.

fabtests/pytest/common.py
```diff
@@ -1,9 +1,15 @@
 import pytest
 import errno
 
-def has_cuda(ip):
+def has_cuda_hmem_support(cmdline_args, host_type, ip):
```
shijin-aws, line 4: Can we split this function into two, `has_cuda` and `has_hmem_support`? In the future we will implement similar things for other hmem ifaces and they can reuse `has_hmem_support` .

wckzhang: Sure

shijin-aws (approved):

wckzhang:
Failure looks unrelated

wckzhang:
Can someone help me merge this patch?

PR #7892, opened 2022-07-22
fabtests: Fix issues in `has_hmem_support()` 

shijin-aws:
This PR contains two commits to fix issues in `has_hmem_support()` in `fabtests/pytest/common.py`:

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -6,9 +6,17 @@ def has_cuda(ip):
     proc = run(["ssh", ip, "nvidia-smi", "-L"])
     return proc.returncode == 0
 
-def has_hmem_support(cmdline_args, host_type, ip):
+def has_hmem_support(cmdline_args, ip):
     from subprocess import run
-    cmd = cmdline_args.populate_command("check_hmem " + cmdline_args.provider, host_type, 10)
+    import os
+    binpath = ""
+    if cmdline_args.binpath:
+        binpath = cmdline_args.binpath
+    cmd = "timeout " + str(cmdline_args.timeout) \
+          + " " + os.path.join(binpath, "check_hmem") \
+          + " " + "-p " + cmdline_args.provider
+    if cmdline_args.environments:
+        cmd = cmdline_args.environments + " " + cmd
     proc = run(["ssh", ip, cmd])
     return proc.returncode == 0
 
@@ -207,7 +215,7 @@ class ClientServerTest:
             if not has_cuda(self._cmdline_args.server_id):
                 pytest.skip("no cuda device")
                 return
-            if not has_hmem_support(self._cmdline_args, "server", self._cmdline_args.server_id):
+            if not has_hmem_support(self._cmdline_args, self._cmdline_args.server_id):
                 pytest.skip("no hmem support")
                 return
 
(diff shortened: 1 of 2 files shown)


fabtests/pytest/common.py
```diff
@@ -8,9 +8,12 @@ def has_cuda(ip):
 
 def has_hmem_support(cmdline_args, host_type, ip):
```
wckzhang, line 9: You can remove host_type and ip since they're no longer being used. Does UnitTest run on both clietn and server?

shijin-aws: Removed the stale arguments. Hmm it seems the unit test will only run on the first node. I need to fix this issue.

fabtests/pytest/common.py
```diff
@@ -217,7 +220,7 @@ def prepare_base_command(self, command_type, executable,
             if not has_cuda(self._cmdline_args.client_id):
                 pytest.skip("no cuda device")
                 return
-            if not has_hmem_support(self._cmdline_args, "client", self._cmdline_args.client_id):
```
wckzhang, line 220: Why run the same line twice? There's no difference between 210 and 220

shijin-aws: It's checking on client and server both.

shijin-aws: But it seems with unit test, both check are running on the same node. Need to fix.

fabtests/pytest/common.py
```diff
@@ -6,11 +6,14 @@ def has_cuda(ip):
     proc = run(["ssh", ip, "nvidia-smi", "-L"])
     return proc.returncode == 0
 
-def has_hmem_support(cmdline_args, host_type, ip):
+def has_hmem_support(cmdline_args):
     from subprocess import run
-    cmd = cmdline_args.populate_command("check_hmem " + cmdline_args.provider, host_type, 10)
-    proc = run(["ssh", ip, cmd])
-    return proc.returncode == 0
+    test = UnitTest(cmdline_args, "check_hmem")
```
wzamazon, line 11: I am not sure about using `UnitTest` for this purpose. Maybe introduce another option to `populate_command`?

shijin-aws: Since this program is very special and may not accept more opts than `-p`, I am going to construct the command manually rather than using `populate_command`.

fabtests/pytest/common.py
shijin-aws: Construct the command explicitly in newest revision.

wzamazon (changes requested):

fabtests/pytest/common.py
```diff
@@ -6,9 +6,13 @@ def has_cuda(ip):
...
     return proc.returncode == 0
 
-def has_hmem_support(cmdline_args, host_type, ip):
+def has_hmem_support(cmdline_args, ip):
     from subprocess import run
-    cmd = cmdline_args.populate_command("check_hmem " + cmdline_args.provider, host_type, 10)
+    import os
+    cmd = cmdline_args.environments \
+          + " " + "timeout " + str(cmdline_args.timeout) \
+          + " " + os.path.join(cmdline_args.binpath, "check_hmem") \
```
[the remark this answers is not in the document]
wckzhang: +1 can you test with no binpath?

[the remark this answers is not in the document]
shijin-aws: Good catch, this also applies for `environments`.

Double check that `timeout` and `provider` cannot be None here.

See newest revision. Run tests with `binpath` as None and `environments` as None and verified it works.

wzamazon (approved):

wckzhang (approved):

shijin-aws:
pending AWS CI.

PR #7893, opened 2022-07-22
core and prov/shm: Add HMEM IPC Caching Mechanism.

shijin-aws:
~~This PR is based on the commit https://github.com/ofiwg/libfabric/pull/7848/commits/d34eacd16de58e45bfb256d5ae974c1c22cb8953 in the draft PR https://github.com/ofiwg/libfabric/pull/7848 opened by @amirshehataornl~~

Opening and closing the HMEM IPC handles are expensive operations. They
    should not be performed on every data operation.
    
This patch adds a caching mechanism. On the very first IPC handle open
    call, the address is cached and the mapped memory address is stored. In
    subsequent calls the cache is queried and the mapped address is retrieved.
    
The IPC cache is implemented as an object of ofi_mr_cache, it has the following
    cache attributes:
    
 cache->add_region: open the ipc handle
 cache->delete_region: close the ipc handle
 cache->monitor: a new ipc monitor of type struct ofi_mem_monitor to check whether the
    ipc_handle is stale.

PR #7894, opened 2022-07-22
prov/efa: enable using shm via CUDA IPC.

shijin-aws:
This PR enables using shm for CUDA hmem. It cannot be merged before https://github.com/ofiwg/libfabric/pull/7893 is merged.

### DIFF
--- a/prov/efa/src/rxr/rxr_msg.c
+++ b/prov/efa/src/rxr/rxr_msg.c
@@ -230,7 +230,16 @@ ssize_t rxr_msg_post_rtm(struct rxr_ep *ep, struct rxr_op_entry *tx_entry, int u
 	assert(rtm_type >= RXR_REQ_PKT_BEGIN);
 
 	if (peer->is_local && ep->use_shm_for_tx) {
-		/* we know shm's capablity, so no need to check handshake */
+		/*
+		 * We know shm's capablity, so no need to check handshake.
+		 * AWS Neuron is currently not supported by the SHM provider.
+		 */
+
+		if (efa_mr_is_neuron(tx_entry->desc[0])) {
+			FI_WARN(&rxr_prov, FI_LOG_CQ,
+			"AWS Neuron is currently not supported by the SHM provider\n");
+			return -FI_EINVAL;
+		}
 		return rxr_pkt_post_req(ep, tx_entry, rtm_type, 0, 0);
 	}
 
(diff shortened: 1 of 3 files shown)


shijin-aws:
This PR can be verified by the single node fabtests on p4d instances. But I will add a unit test inside this PR later for the change in `prov/efa` as well.

shijin-aws:
bot:aws:retest

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -234,7 +234,16 @@ ssize_t rxr_msg_post_rtm(struct rxr_ep *ep, struct rxr_op_entry *tx_entry, int u
...
-		/* we know shm's capablity, so no need to check handshake */
+		/*
+		 * We know shm's capablity, so no need to check handshake.
+		 * AWS Neuron is currently not supported by the SHM provider.
+		 */
+
+		if (efa_mr_is_neuron(tx_entry->desc[0])) {
+			FI_WARN(&rxr_prov, FI_LOG_CQ,
+			"AWS Neuron is currently not supported by the SHM provider\n");
+			return -FI_EINVAL;
```
wckzhang, line 245: This behavior seems different than before, why not go through the original code path (ie. outside this if block), if it's neuron.

wzamazon (approved):

prov/efa/src/rxr/rxr_msg.c
shijin-aws: Not sure I understand the "original code path". So before this PR, we disable shm for all FI_HMEM in `rxr_ep_use_shm_for_tx`. But now we can only disable shm for certain FI_HMEM ifaces like neuron (and synapseai in the future). And it is expected that neuron applications or neuron-nccl will not use shm. So we should return an error in `rxr_msg_post_rtm` when posting rtm for shm. 

I don't know how we can move this code outside this if block?

wckzhang (approved):

prov/efa/src/rxr/rxr_msg.c
shijin-aws: offline talked with @wckzhang. This patch changes the code's behavior. Before this PR, shm will be turned off silently if applications requests `FI_HMEM`. After this PR, shm will not be turned off silently, but when the application tries to send rtm with shm for neuron tx buffer, the code will error and fail.

This behavior change is by design after discussing with @wzamazon, because it is expected that neuron should not use libfabric for intra-node node communication.

shijin-aws:
bot:aws:retest

shijin-aws:
https://github.com/ofiwg/libfabric/pull/7893 is merged, will rebase this PR on the latest main.

shijin-aws:
latest revision is based on the last main branch, and removes the origin commit of skipping test_runt on single node, because it is already merged in main line as https://github.com/ofiwg/libfabric/commit/80e7d6915ab7b4cf7c30f20fc2655d40e0998994

wzamazon (approved):

PR #7904, opened 2022-07-29
prov/net: Optimize receive side tag matching by src_addr

shefty:
According to our MPI architect, apps do not use ANY_SOURCE for tag matching (or at least very few do).  We update the matching in the net provider to use separate tag queues per source address in order to reduce the average time to find a matching receive buffer.  In most cases, we expect that the desired buffer will be at the head of the source tag queue.

The first patch adds an array abstraction that grows dynamically.  This is similar to the index map implementation, with a double index lookup to handle sparse array entries.  The array will store the per source queues (basically a single-linked list structure).  The second patch adds per source tag queues to the net provider.

To handle the case where the app does post ANY_SOURCE buffers, we add a sequence number to each posted receive.  Matching selects the buffer with the lowest sequence number in the case that multiple buffers match (for example, a source targeted buffer and an any source buffer).

The last patch is unrelated, but provides a fix needed to pass fi_getinfo_test.


shefty:
@ooststep - These changes pass fabtests, but I haven't tested them with MPI yet.  Hopefully I didn't break the tag matching.

We should be able to build off these changes to identify which sockets the application is expecting to receive data from (assuming RMA isn't active).  For example, if we post a receive for a specific connection, that could automatically move the connection to active if it's not.

I'd like to see if these changes improve performance at scale paired with the existing poll fairness implementation.  I'm hoping it reduces tag matching searching.

PR #7912, opened 2022-08-02
prov/net: Fixes to provider

shefty:
Set a connected endpoint to active if:

- The message queue for the endpoint is currently empty
- We post a tagged message to receive data from its peer
- The connection is not already active

The posting of the message likely indicates that data is
expected from the peer soon.

### DIFF
--- a/prov/net/src/xnet.h
+++ b/prov/net/src/xnet.h
@@ -91,15 +91,6 @@ struct xnet_ep;
 struct xnet_progress;
 struct xnet_domain;
 
-
-/* Lock ordering:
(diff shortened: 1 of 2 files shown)


shefty:
@ooststep - I have no idea if this will help or hurt performance.

wzamazon:
bot:aws:retest

shefty:
Added fix for use-after-free problem in rdm endpoint cleanup.

shefty:
swapped out 1st patch, which is a more questionable optimization, and kept only the patches that add fixes.

PR #7914, opened 2022-08-02
fabtests/efa: fix runt test

wzamazon:
Currently, test_runt fails on system with older version of EFA kernel module, which does not support some HW counters.

This patch address the issue by skipping the test under such case.

### DIFF
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -29,7 +29,12 @@ def efa_retrieve_hw_counter_value(hostname, hw_counter_name):
     return: an integer that is sum of all EFA device's counter
     """
     command = "ssh {} cat \"/sys/class/infiniband/*/ports/*/hw_counters/{}\"".format(hostname, hw_counter_name)
-    process = subprocess.run(command, shell=True, check=True, stdout=subprocess.PIPE)
+    try:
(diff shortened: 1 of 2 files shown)


fabtests/pytest/efa/efa_common.py
```diff
@@ -29,7 +29,12 @@ def efa_retrieve_hw_counter_value(hostname, hw_counter_name):
     return: an integer that is sum of all EFA device's counter
     """
     command = "ssh {} cat \"/sys/class/infiniband/*/ports/*/hw_counters/{}\"".format(hostname, hw_counter_name)
-    process = subprocess.run(command, shell=True, check=True, stdout=subprocess.PIPE)
+    try:
+        process = subprocess.run(command, shell=True, check=True, stdout=subprocess.PIPE)
+    except subprocess.CalledProcessError:
+        # this can happen when OS is using older version of EFA kernel module
```
wenduwan, line 35: IIUC we cannot reliably get in-tree EFA kmod version otherwise it should be better to skip this test based on the version number, right?

shijin-aws: I think so, in-tree kmod does not have a version.

shijin-aws (approved):

wenduwan (approved):

wzamazon:
bot:aws:retest

wckzhang (approved):

PR #7915, opened 2022-08-02
core: Fix altering of MR mode with FI_MR_UNSPEC

iziemba:
Prior to this change, if the hints had FI_MR_UNSPEC set for domain
attribute mr_mode field, fi_alter_domain_attr() would AND the provider
mr_mode bits against FI_MR_UNSPEC thus zeroing the returned mr_mode
bits.

FI_MR_UNSPEC may be used to indicate support for any registration mode.
If FI_MR_UNSPEC is set, instead of zeroing provider returned mr_mode
bits, set the mr_mode bits based on requested capabilities and default
provider mr_mode bits.

Signed-off-by: Ian Ziemba <ian.ziemba@hpe.com>

### DIFF
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -1143,7 +1143,7 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
 {
 	int hints_mr_mode;
 
-	hints_mr_mode = hints ? hints->mr_mode : 0;
+	hints_mr_mode = hints ? hints->mr_mode : FI_MR_UNSPEC;
 	if (hints_mr_mode & (FI_MR_BASIC | FI_MR_SCALABLE)) {
 		attr->mr_mode = hints_mr_mode;
 	} else if (FI_VERSION_LT(api_version, FI_VERSION(1, 5))) {
@@ -1151,12 +1151,12 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
 				FI_MR_BASIC : FI_MR_SCALABLE;
 	} else {
 		attr->mr_mode &= ~(FI_MR_BASIC | FI_MR_SCALABLE);
-
-		if (hints &&
-		    ((hints_mr_mode & attr->mr_mode) != attr->mr_mode)) {
+		if (hints_mr_mode == FI_MR_UNSPEC)
 			attr->mr_mode = ofi_cap_mr_mode(info_caps,
-						attr->mr_mode & hints_mr_mode);
-		}
+							attr->mr_mode);
+		else
+			attr->mr_mode = ofi_cap_mr_mode(info_caps,
+							attr->mr_mode & hints_mr_mode);
 	}
 
 	attr->caps = ofi_get_caps(info_caps, hints ? hints->caps : 0, attr->caps);


prov/util/src/util_attr.c
```diff
@@ -1143,20 +1143,20 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
...
 		attr->mr_mode = hints_mr_mode;
 	} else if (FI_VERSION_LT(api_version, FI_VERSION(1, 5))) {
 		attr->mr_mode = (attr->mr_mode && attr->mr_mode != FI_MR_SCALABLE) ?
 				FI_MR_BASIC : FI_MR_SCALABLE;
 	} else {
 		attr->mr_mode &= ~(FI_MR_BASIC | FI_MR_SCALABLE);
-
-		if (hints &&
-		    ((hints_mr_mode & attr->mr_mode) != attr->mr_mode)) {
+		if (hints_mr_mode == FI_MR_UNSPEC)
```
shefty, line 1154: FI_MR_UNSPEC is not a valid value for API version >= 1.5.  It's only useful to select either FI_MR_BASIC or FI_MR_SCALABLE, when mr_mode is treated as an enum, rather than a bit field.

iziemba: So if an app can handle all MR mode bits (it adapts based on the fi_info returned by the provider), how can it convey this to the provider? From the fi_mr manpage, FI_MR_UNSPEC seemed to be this way.

_Buffers used in data transfer operations may require notifying the provider of their use before a data transfer can occur. The mr_mode field indicates the type of memory registration that is required, and when registration is necessary. Applications that require the use of a specific registration mode should set the domain attribute mr_mode to the necessary value when calling fi_getinfo. The value FI_MR_UNSPEC may be used to indicate support for any registration mode._

shefty: I don't see any reference to FI_MR_UNSPEC in the man page.
```
https://ofiwg.github.io/libfabric/v1.15.1/man/fi_mr.3.html
```
An app can never handle *all* mr_mode bits.  They can only handle all that have been defined in the latest implementation.  Meaning if the app sets mr_mode=~0, for example, they will be broken as soon as a new mr_mode bit is added.  mr_mode acts the same as the standard mode bits.  A value of 0 indicates that the app does not support any bits.

The app should set ```mr_mode=FI_MR_LOCAL | FI_MR_RAW | FI_MR_VIRT_ADDR...``` to indicate that it supports all of the current bits.  This allows the app to continue working even as new bits are added.  Each mr_mode bit requires the app to perform some specific action for it to work, which is why a wild card doesn't work.

An app can use NULL hints into fi_getinfo() to obtain a list of what each provider supports and requires.

iziemba: Could you check the https://ofiwg.github.io/libfabric/v1.15.1/man/fi_domain.3.html manpage? I think there is a bit of contradication here around FI_MR_UNSPEC.

shefty: From that page:
```
FI_MR_UNSPEC
Defined for compatibility – library versions 1.4 and earlier.
```
The statement you're referring to should be updated to reference the version number as well to avoid the confusion this is causing.

iziemba: Also, above comment makes sense. I'll close out this PR.

shefty: Actually, the references to the FI_MR_xxx bits should probably be deleted entirely from the fi_domain man page, with instead a reference to the fi_mr man page, where the bits are described in more detail.

PR #7921, opened 2022-08-03
fabtests/configure: fix ze_loader and ibverbs config check for component tests

aingerson:
According to the AC_CHECK_LIB man page:
If action-if-found is not specified, the default action prepends -llibrary to LIBS

This suggests that if the default action is not provided, the library will not be added to
LIBS by default. This patch explicitly adds the ze_loader and ibverbs libraries to LIBS if
they were found to avoid linking errors.

It also changes the function that the verbs_devel check looks for since older versions might
not have the newer version which has the required function for the test.


j-xiong (approved):
Thanks!

Issue #7837, opened 2022-06-16
fabtest/fi_rma_bw: when run with -p verbs, leaks mr during run until out of memory


chien-intel:
**Describe the bug**
fi_rma_bw -p verbs leaks MRs during the test.  Use "watch -n 1 rdma resource" to monitor resource usage and see QP count go up by 1 and down by 1 but MR count does not come down and eventually program exits with this error:

transmit(): common/shared.c:1959, ret=-12 (Cannot allocate memory)

**To Reproduce**
Run fi_rma_bw with -p verbs

**Expected behavior**
MR count should go down when a QP is destroyed

**Output**
If applicable, add output to help explain your problem. (e.g. backtrace, debug logs)

**Environment:**
OS (if not Linux), provider, endpoint type, etc.

**Additional context**

chien-intel:
Start fi_rma_bw as client without server .  On every attempt to connect to the server we are consuming memory via rxm_prepost_recv and MR on ofi_bufpool_grow.

chien-intel:
Resolved with PR #7991 - commit 4184fd66d7e434bd921b9528b8aab449173e49e6

PR #7923, opened 2022-08-03
prov/rxm: stop growing ep->rx_pool on connection attempt

chien-intel:
This fixes https://github.com/ofiwg/libfabric/issues/7837

When execute fi_rma_bw -p verbs <server_ip> without
a server instance listening on <server_ip>, it will
rapidly consume memory and MRs.  The cause is on every
connection attempt, rxm_prepost_recv is allocating and
posting RX buffers without freeing the buffers back
to rx_pool after connection failure.  Add ofi_bufpool_reset()
to put all buffers in a pool back to free_list and only
call it if there is no srx_ctx and connection has not been
established.

### DIFF
--- a/prov/rxm/src/rxm_conn.c
+++ b/prov/rxm/src/rxm_conn.c
@@ -87,6 +87,10 @@ static void rxm_close_conn(struct rxm_conn *conn)
 	}
 	fi_close(&conn->msg_ep->fid);
 	rxm_flush_msg_cq(conn->ep);
+
+	if ((!conn->ep->srx_ctx) && (conn->state != RXM_CM_CONNECTED))
+		ofi_bufpool_reset(conn->ep->rx_pool);
+
 	dlist_remove_init(&conn->loopback_entry);
 	conn->msg_ep = NULL;
 
--- a/include/ofi_mem.h
+++ b/include/ofi_mem.h
@@ -370,6 +370,7 @@ ofi_bufpool_create(struct ofi_bufpool **buf_pool,
 	return ofi_bufpool_create_attr(&attr, buf_pool);
 }
 
+void ofi_bufpool_reset(struct ofi_bufpool *pool);
 void ofi_bufpool_destroy(struct ofi_bufpool *pool);
 
 int ofi_bufpool_grow(struct ofi_bufpool *pool);
--- a/prov/util/src/util_buf.c
+++ b/prov/util/src/util_buf.c
@@ -244,6 +244,42 @@ int ofi_bufpool_create_attr(struct ofi_bufpool_attr *attr,
 	return FI_SUCCESS;
 }
 
+/* Use this function with care.
+ * Reset a bufpool, put all buffers back to free_list.
+ */
+void ofi_bufpool_reset(struct ofi_bufpool *pool)
+{
+	struct ofi_bufpool_region *buf_region;
+	struct ofi_bufpool_hdr *buf_hdr;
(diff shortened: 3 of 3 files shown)


shefty (changes requested):
From the description, the correct fix belongs in the user of the pool, not by modifying the pool.  If rxm is removing entries from the pool, it is on rxm to free those entries.

Adding a problematic API (use with care) should be avoided.  "Resetting" a pool can be accomplished by destroying/creating it.  That also reset all the buffers in the pool, versus just the internal tracking that the pool does.

prov/rxm/src/rxm_conn.c
```diff
@@ -87,6 +87,10 @@ static void rxm_close_conn(struct rxm_conn *conn)
 	}
 	fi_close(&conn->msg_ep->fid);
 	rxm_flush_msg_cq(conn->ep);
+
+	if ((!conn->ep->srx_ctx) && (conn->state != RXM_CM_CONNECTED))
+		ofi_bufpool_reset(conn->ep->rx_pool);
```
shefty, line 92: Isn't the rx_pool used for all connections.  The kills the pool, even if the buffers are being used for connections that are still active.

j-xiong: Agree. The preposted recv buffers should be freed when `msg_ep` is closed. That means a proper fix would be in the base provider instead of rxm. For example, in the verbs provider, the send queue is flushed when the endpoint is closed, but the receive queue is not.

chien-intel: rxm_prepost_recv is allocating RX buffer from ep->rx_pool for msg_ep.  On a connection failure msg_ep is freed but leaving behind allocated RX buffer in ep->rx_pool .  ep->rx_pool grows from there on subsequent connection request.
Destroying ep->rx_pool was the first thing I tried.   To get that working, it would require a great deal of change and impacting other providers as well. I also looked into calling rxm_prepost_recv only once but still need to post recv to msg_ep so that didn't work out either that's why I ended up resetting the rx_pool.  Another approach is to track buffers allocated to msg_ep and only return those back to the pool but I would need to keep a list pointer somewhere, most likely in msg_ep which is a fid_ep or rxm_conn structure.  This is the least impactful patch I could come up with.  I don't like resetting the pool either but it was the easiest path.

chien-intel: I'm looking at verbs provider to see what happens on vrb_ep_close

shefty: A thought that occurred to me is that if the connection never formed, the QP is likely only in the RTR state.  I don't know if that is supposed to make a difference, or if it changes anything done in the rdmacm.

chien-intel: it shouldn't.  you can post recv as soon as QP is created and if buffers are not flushed you cannot track resources correctly.  There is something missing.  I'm trying to verify verbs provider with CVL as I know it does flush.

shefty: We need to check the call flow to see if the QP was ever transitioned into the error state.  If it was simply destroyed, it wouldn't flush.  If we called rdma_disconnect(), that will transition into the error state.  But if we were never connected at the OFI level, we're probably not calling that and going directly to rdma_destroy_ep(), which just closes the EP.

chien-intel:
This is not the correct solution to the issue

PR #7926, opened 2022-08-04
core/fabric: Make PSM2 higher default priority

timothom64:
Until the opx provider is ready, psm2 should be
the default provider selected for Omnipath fabrics.

Change ofi_ordered_provs_init() accordingly

Signed-off-by: Tim Thompson <tim.thompson@cornelisnetworks.com>

### DIFF
--- a/src/fabric.c
+++ b/src/fabric.c
@@ -411,7 +411,7 @@ static struct fi_provider *ofi_get_hook(const char *name)
 static void ofi_ordered_provs_init(void)
 {
 	char *ordered_prov_names[] = {
-		"efa", "opx", "psm2", "psm", "usnic", "gni", "bgq", "verbs",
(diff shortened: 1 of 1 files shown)


shefty (approved):
I think we're approaching needing more sophisticated ordering logic as some sort of post-discovery step.

timothom64:
> I think we're approaching needing more sophisticated ordering logic as some sort of post-discovery step.

Is this the first time two hardware providers could be used for one HPC platform? (Omnipath)

shefty:
No, we have that with verbs already.  That comment was general, not specific to your changes.

timothom64:
Can this get into 1.15.2 ?

shefty:
Absolutely, please pick for v1.15.x.

shefty:
Intel CI is down, and it looks like AWS EFA CI is also down.  PR looks fine.

PR #7928, opened 2022-08-05
configure: Check for alternative path of i915_drm.h

j-xiong:
The header 'i195_drm.h' could be under either '/usr/include/drm' or
'/usr/include/libdrm' depending on the system and the package it is
installed from.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -617,14 +617,19 @@ AS_IF([test x"$with_ze" != x"no"],
 AC_DEFINE_UNQUOTED([ENABLE_ZE_DLOPEN], [$ze_dlopen], [dlopen ZE libraries])
 
 have_drm=0
+have_libdrm=0
 AS_IF([test "$have_ze" = "1"],
-      [AC_CHECK_HEADER(drm/i915_drm.h, [have_drm=1])])
+      [AC_CHECK_HEADER(drm/i915_drm.h,
+		       [have_drm=1],
+		       [AC_CHECK_HEADER(libdrm/i915_drm.h,
+					[have_libdrm=1])])])
 
 AS_IF([test x"$with_ze" != x"no" && test -n "$with_ze" && test "$have_ze" = "0" ],
 	[AC_MSG_ERROR([ZE support requested but ZE runtime not available.])])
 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -617,14 +617,19 @@ AS_IF([test x"$with_ze" != x"no"],
 AC_DEFINE_UNQUOTED([ENABLE_ZE_DLOPEN], [$ze_dlopen], [dlopen ZE libraries])
 
 have_drm=0
+have_libdrm=1
```
aingerson, line 620: Should this be initialized to 0?

src/hmem_ze.c
```diff
@@ -309,8 +309,12 @@ ze_result_t ofi_zeDeviceGetProperties(ze_device_handle_t hDevice,
 	return (*libze_ops.zeDeviceGetProperties)(hDevice, pDeviceProperties);
 }
 
+#if HAVE_DRM || HAVE_LIBDRM
```
aingerson, line 312: Might be cleaner as 
```
#if HAVE_DRM
#include <drm/i915_drm.h>
#elif HAVE_LIBDRM
#include <libdrm/i915_drm.h>
#endif
```

aingerson (commented):
Was just about to add something similar :) Thanks!

configure.ac
j-xiong: yeah, I wasn't drinking enough coffee.

src/hmem_ze.c
j-xiong: this `#if` covers much more than the `#include` lines.

shefty:
Do we need this for v1.15.2?

aingerson (approved):
Thank you!

j-xiong:
@shefty Yes.

j-xiong:
One side effect of the new 'rebase and merge' policy is that I can no longer cherry-pick and have the correct 'cherry-picked from ...' line until the commit is merged.

wzamazon:
bot:aws:retest

j-xiong:
Intel CI error is "address already in use" for the tcp provider which is unrelated to this PR.

PR #7934, opened 2022-08-08
prov/efa: Add locks around mr map

wckzhang:
We were missing locks around the mr map operations which caused thread
unsafe behavior. Also removed a duplicate ofi_mr_map_remove call in the
error path (It's already called in the fi_close path).

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -428,8 +428,12 @@ static int efa_mr_dereg_impl(struct efa_mr *efa_mr)
 			"Unable to deregister memory registration\n");
 		ret = err;
 	}
+
+	ofi_genlock_lock(&efa_domain->util_domain.lock);
 	err = ofi_mr_map_remove(&efa_domain->util_domain.mr_map,
 				efa_mr->mr_fid.key);
+	ofi_genlock_unlock(&efa_domain->util_domain.lock);
+
 	if (err) {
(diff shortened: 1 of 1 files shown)


wzamazon (approved):
Thanks!

I think we will need to add multi-thread mr reg/dereg test to verify the fix.

ryanamazon (approved):

tschuett:
You could add a efa_ofi_mr_map_remove wrapper and give it the domain and assert that the lock is hold. This trick used in other places in libfabric.

wckzhang:
> You could add a efa_ofi_mr_map_remove wrapper and give it the domain and assert that the lock is hold. This trick used in other places in libfabric.

Let me see if I'm understanding what you're saying, something like this?

```
efa_mr_map_remove(domain)
{
    assert(domain->lock->in_use)
    ofi_mr_map_remove
}

.
.
.
lock(domain->lock)
efa_mr_map_remove(domain)
unlock(domain->lock)
.
.
.
```

tschuett:
Almost. You also have to pass in `efa_mr->mr_fid.key`. But basically yes. There is a macro somewhere for checking that a lock is hold.

tschuett:
`ofi_mutex_held`
https://github.com/ofiwg/libfabric/blob/4749c3215aae90f374094cb1430306cea19a4137/include/ofi_lock.h#L264

tschuett:
In your case it should be `ofi_genlock_hold`.

PR #7937, opened 2022-08-09
contrib/intel/jenkins: Re-enable ze-shm stage

zachdworkin:
Add the ze-shm stage back to the CI and disable the actual failing test, (xd2d: cross device to device), rather than the whole stage


zachdworkin:
@aingerson @shefty Intel CI fail is due to a zombie process from the abort and replay, and is unrelated to this change.

PR #7941, opened 2022-08-10
prov/efa: update references to efadv_wc_read_sgid

wenduwan:
`efadv_wc_read_ah` is renamed to `efadv_wc_read_sgid` in rdma-core.
The new function does not return peer AH but instead 0(success) or
negative(peer AH is known).

### DIFF
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -134,6 +134,17 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 			[efadv_support_extended_cq=0])
 	      ])
 
+	# efadv_wc_read_sgid is a static inline function, which is
+	# required to use extended CQ for recovering peer address.
+	# Extended CQ should be disabled if it is not available.
+	AS_IF([test $efadv_support_extended_cq -eq 1],
+		[AC_EGREP_HEADER(
+			[int efadv_wc_read_sgid],
+			[infiniband/efadv.h],
+			[],
+			[efadv_support_extended_cq=0])
+		])
+
 	AS_IF([test $efadv_support_extended_cq -eq 1],
 		[AC_DEFINE([HAVE_EFADV_CQ_EX], [1], [EFA device support extensible CQ])],
 		[AC_DEFINE([HAVE_EFADV_CQ_EX], [0], [EFA device does not support extensible CQ])]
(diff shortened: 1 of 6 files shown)


wzamazon (commented):
Do you need to change config script about checking for the function `efadv_wc_read_sgid`?

wenduwan:
> Do you need to change config script about checking for the function `efadv_wc_read_sgid`?

In autoconf script we are only checking the exported functions `efadv_create_cq` & `efadv_cq_from_ibv_cq_ex`, which are unchanged.

`efadv_wc_read_sgid` is a static function and we cannot check for it.

wzamazon:
> In autoconf script we are only checking the exported functions efadv_create_cq & efadv_cq_from_ibv_cq_ex, which are unchanged.

Then libfabric will not compile on a rdma-core that has `efadv_create_cq` but not have `efadv_wc_read_sgid`. Right?

> efadv_wc_read_sgid is a static function and we cannot check for it.

Is that true? I think autoconf just check whether a program compiles.

wenduwan:
> Then libfabric will not compile on a rdma-core that has efadv_create_cq but not have efadv_wc_read_sgid. Right?

Correct. We need to get this change in before that in rdma-core to avoid breaking the build.

> Is that true? I think autoconf just check whether a program compiles.

We are doing it [here](https://github.com/ofiwg/libfabric/blob/main/prov/efa/configure.m4#L113-L135). It requires the symbol to be visible.

I ingested the change in rdma-core locally and libfabric compilation failed.

shijin-aws:
bot:aws:retest

wenduwan:
bot:aws:retest

prov/efa/configure.m4
```diff
@@ -134,6 +134,15 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 			[efadv_support_extended_cq=0])
 	      ])
 
+	AS_IF([test $efadv_support_extended_cq -eq 1],
+		[AC_EGREP_HEADER(
```
wzamazon, line 138: Thank you!

Maybe add a comment here noting that the reason we want to to use extended CQ is because it provided extra functionality like `read_sgid`.  Therefore if rdma-core does not support read sgid, we do not want to use extended CQ.

prov/efa/configure.m4
```diff
@@ -134,6 +134,18 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
...
+			[int efadv_wc_read_sgid],
+			[infiniband/efadv.h],
+			[],
+			[
+				efadv_support_extended_cq=0
+				AC_MSG_RESULT([efadv_wc_read_sgid is not available from rdma-core. Do not recover peer address from extended CQ.])
```
wzamazon, line 144: `Do not recover peer address from extended CQ` is a bit confusing. Maybe just remove it.

wzamazon (approved):

PR #7951, opened 2022-08-16
fi_rdm_rpc_bw test

iziemba:
Round number two at creating an RPC benchmark. Example output.

```
RPCs Inflight: 64
RPC Warmup Iterations: 10
HMEM Type: FI_HMEM_SYSTEM
MR mode: FI_MR_ALLOCATED, FI_MR_PROV_KEY, FI_MR_ENDPOINT
Provider: cxi
name                                              bytes   iters   total       time     MB/sec    usec/xfer   Mxfers/sec
msg_req                                           64      1k      62k         0.00s     14.85       4.31       0.23
msg_req                                           256     1k      250k        0.00s     57.84       4.43       0.23
msg_req                                           1k      1k      1000k       0.00s    236.00       4.34       0.23
msg_req                                           4k      1k      3.9m        0.01s    504.19       8.12       0.12
msg_req                                           64k     1k      62m         0.03s   2542.82      25.77       0.04
msg_req                                           1m      1k      1000m       0.28s   3716.39     282.15       0.00
tag_req                                           64      1k      62k         0.00s     15.22       4.20       0.24
tag_req                                           256     1k      250k        0.00s     60.35       4.24       0.24
tag_req                                           1k      1k      1000k       0.00s    234.70       4.36       0.23
tag_req                                           4k      1k      3.9m        0.01s    523.65       7.82       0.13
tag_req                                           64k     1k      62m         0.03s   2582.90      25.37       0.04
tag_req                                           1m      1k      1000m       0.28s   3719.88     281.88       0.00
read_req                                          64      1k      62k         0.00s     13.65       4.69       0.21
read_req                                          256     1k      250k        0.00s     54.13       4.73       0.21
read_req                                          1k      1k      1000k       0.00s    216.22       4.74       0.21
read_req                                          4k      1k      3.9m        0.01s    717.84       5.71       0.18
read_req                                          64k     1k      62m         0.02s   3670.05      17.86       0.06
read_req                                          1m      1k      1000m       0.08s  13624.42      76.96       0.01
write_req                                         64      1k      62k         0.00s     14.65       4.37       0.23
write_req                                         256     1k      250k        0.00s     58.01       4.41       0.23
write_req                                         1k      1k      1000k       0.01s    203.94       5.02       0.20
write_req                                         4k      1k      3.9m        0.01s    790.89       5.18       0.19
write_req                                         64k     1k      62m         0.01s   4578.13      14.31       0.07
write_req                                         1m      1k      1000m       0.07s  14367.01      72.99       0.01
```


shefty:
First comment without reviewing the code, can you make the output more condensed, so half the width of the screen isn't taken up with blank space?  Focus the output on a more normal terminal window size.

shefty:
The rdm_stress test is trying to break the server with unexpected messages, clients that go away in the middle of operations, hitting the server hard with multiple processes, etc.  But it does provide an rpc framework.  It may be useful to separate the rcp framework from this test (I saw part of the changes here did that) and share that with a new test focused on performance.

shefty:
FWIW, rdm_stress is attempting to emulate the DAOS work model.

iziemba:
It sure is doing a good job at unexpected messaging :)

I think I can resolve the buffer use issue by implementing some sort of buffer pool which gets initialized during warmup iterations and torn down at the end of benchmark iteration.

Still thinking about unexpected messaging...

shefty:
I started a review, but I'm not done yet.  We need to look at sharing code between libfabric and fabtests.  libfabric has the abstractions that we want to re-use (buffer pool, stack, ring buffer, etc.)

iziemba:
Need to update rdm_bw.c so it can compile.

iziemba:
Comparing the RMA RPC behavior between my previous PR and this PR, I cannot achieve the same performance with this PR. I even did implement a buffer pool and that only gave another 1GB/s at 1M transfer size. I am still missing 6-7 GB/s.

What my first PR was trying create optimal situations (no unexpected message, buffer pool), and message RMA RPC performance. This does seem different from the rdm_stress use-case which is almost testing sub-optimal situations. While not ideal, maybe it makes sense to have two different RPC implementation since they address different things.

shefty:
The base rcp code should be re-usable.  The main difference should be in how the calls are made and the main structure of the test.  For example, posting the expected response message prior to sending the request - that needs a different implementation of tag_req() and tag_resp() at the client.  There should be a fair amount of reuse at the server.

shefty:
Btw, I agree that rdm_stress is not performance focused.  We need a separate test or at least set of changes for that.  But we should be able to reuse at least some of the rpc code.  Optimized 'ctrl_op' callbacks might be all that's needed.

iziemba:
Something else which is making it a bit difficult is rdm_stress not using the same control structure for the same req + resp pair.

shefty:
Does this PR replace #7922?  Or have you reverted back to 7922?  I've lost the plot

shefty:
Replaced by previous PR.

PR #7952, opened 2022-08-16
hmem_neuron.c: Use nrt_memcpy_to_device to copy from host to dev.

shijin-aws:
neuron added a memcpy function and neuron_copy_to_dev()
should use it to copy data from host to device.

### DIFF
--- a/src/hmem_neuron.c
+++ b/src/hmem_neuron.c
@@ -50,6 +50,7 @@ struct neuron_ops {
 					  const char *name, nrt_tensor_t **tensor);
 	void (*nrt_tensor_free)(nrt_tensor_t **tensor);
 	void *(*nrt_tensor_get_va)(const nrt_tensor_t *tensor);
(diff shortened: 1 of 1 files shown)


shefty:
Is this needed for v1.15.2?

bwbarrett (changes requested):
The commit message background on write combining is irrelevant.  We should have a memcpy in the NRT runtime because it's a path to optimizing memcpys from host to device memory (rather than memcpy(), which is really designed for host to host copies).  This otherwise looks good, but please update the commit message.

shijin-aws:
@bwbarrett Thanks, updated the commit message.

bwbarrett (approved):

shijin-aws:
@shefty 

> Is this needed for v1.15.2?

I think it needs to be backported to v1.15.x. Will do it today and update the EFA news

shijin-aws:
Will merge after CI passes. Will backport to v1.15.x

shijin-aws:
bot:aws:retest

PR #7954, opened 2022-08-17
[v1.15.x] src/hmem_ze: Move creation of ze command queues out of initialization…

shefty:
… and into the copy function.

In the case of ULLS (ultra low latency submission) where ze gets initialized and never used;
creating a command queue spawns a thread that repeatedly submits 0 byte copies to the command
queue so that it never gets swapped out. If ze is initialized and never used it creates a
performance impact because it is never used. Move the initialization to the copy function to
remove this overhead.

Signed-off-by: zdworkin <zachary.dworkin@intel.com>

### DIFF
--- a/src/hmem_ze.c
+++ b/src/hmem_ze.c
@@ -754,15 +754,6 @@ int ze_hmem_init(void)
 		if (ze_ret)
 			goto err;
 
-		cq_desc.ordinal = ordinals[num_devices];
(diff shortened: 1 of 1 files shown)


shefty:
Intel CI failure is the standard always failing verbs rxd problem.  AWS failure looks like some ssh connectivity issue.

shefty:
bot:aws:retest

PR #7956, opened 2022-08-24
prov/net: Updates to socket config and poll behavior

shefty:
Request socket send/receive buffering to 128k as a default.
Convert global ofi_poll_fairness variable to net provider variable.
Use a separate counter from poll fairness to move sockets off the hot set.

### DIFF
--- a/include/ofi_epoll.h
+++ b/include/ofi_epoll.h
@@ -57,8 +57,6 @@ struct ofi_epollfds_event {
 };
 #endif
 
-extern int ofi_poll_fairness;
-
 enum ofi_pollfds_ctl {
 	POLLFDS_CTL_ADD,
 	POLLFDS_CTL_DEL,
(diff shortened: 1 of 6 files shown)


prov/net/src/xnet_ep.c
```diff
@@ -113,8 +113,16 @@ static void xnet_set_no_port(SOCKET sock)
 
 int xnet_setup_socket(SOCKET sock, struct fi_info *info)
 {
-	int ret, optval = 1;
+	int ret, optval;
 
+	/* Request 128k buffer sizes, but use system defaults on failure */
+	optval = 128 * 2014;
```
aakefbs, line 119: What magic number is '2014' based on? Typo of 1024?

aakefbs (changes requested):

prov/net/src/xnet_ep.c
shefty: Yes, this should have been 128k.  This is the value used by Intel MPI 2018, so I went with it.  (2018 shows better tcp performance and scaling than later versions.)  4M per connection seems huge at scale.  With 1000 ranks, that's 4G per process, and we're currently running with about 56 ranks per node.  I can drop this patch and continue to use the system defaults.

shefty:
bot:aws:retest

PR #7958, opened 2022-08-25
fabtests/pytest/efa: skip runt_read test for single node

wzamazon:
runt_read is not intended to be used on single node, therefore
this patch skip the run_read test on it.


wenduwan (approved):

wzamazon:
bot:aws:retest

PR #7959, opened 2022-08-25
prov/efa: a unit test bug fix and handle the case when EFA DV CQ is not supported

wenduwan:
- Unit test buffer construction used the wrong buffer address, which caused `Bad address` error from `fi_mr_reg` at large buffer sizes, e.g. 8800 bytes. This is a subtle bug and we did not notice it for 4096 bytes.
- The recent change to query peer address from EFA device did not account for the case when EFA DV CQ is not supported in the kernel, which will cause `efadv_create_cq` to fail. In this change we add the handling logic to fallback to `ibv_create_cq_ex` if `efadv_create_cq` fails(for any reason). We use a bit mask to indicate whether the `efa_cq->ibv_cq_ex` is created with `efadv_create_cq` or `ibv_create_cq_ex`, and thereby decide if we should try to query peer address from EFA device.

### DIFF
--- a/prov/efa/src/efa.h
+++ b/prov/efa/src/efa.h
@@ -122,6 +122,9 @@
 /* maximum name length for shm endpoint */
 #define EFA_SHM_NAME_MAX	   (256)
 
+/* efa_cq->ibv_cq_ex is created by efadv_create_cq */
+#define EFA_CQ_SUPPORT_EFADV_CQ BIT_ULL(0)
+
 extern struct fi_provider efa_prov;
 extern struct util_prov efa_util_prov;
 
@@ -192,6 +195,7 @@ struct efa_cq {
 	efa_cq_read_entry	read_entry;
 	ofi_spin_t		lock;
 	struct ofi_bufpool	*wce_pool;
+	uint32_t	flags; /* User defined capability mask */
 
 	struct ibv_cq_ex	*ibv_cq_ex;
 };
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -31,6 +31,8 @@
  * SOFTWARE.
  */
 
+#include <errno.h>
+#include <string.h>
 #include "config.h"
 #include <ofi_mem.h>
 #include "efa.h"
@@ -248,6 +250,28 @@ static struct fi_ops efa_cq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+/**
+ * @brief Create ibv_cq_ex by calling ibv_create_cq_ex
+ *
+ * @param[in] cq Pointer to the efa_cq.
+ * @param[in] attr Pointer to fi_cq_attr.
+ * @param[out] Return pointer to ibv_cq_ex if successful, otherwise NULL
+ */
+static inline struct ibv_cq_ex *efa_cq_ibv_create_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
+	struct ibv_cq_init_attr_ex init_attr_ex = {
+		.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE,
+		.cq_context = NULL,
+		.channel = NULL,
+		.comp_vector = 0,
+		/* EFA requires these values for wc_flags and comp_mask.
+		 * See `efa_create_cq_ex` in rdma-core.
+		 */
+		.wc_flags = IBV_WC_STANDARD_FLAGS,
+		.comp_mask = 0,
+	};
+
+	return ibv_create_cq_ex(cq->domain->device->ibv_ctx, &init_attr_ex);
+}
 
 /**
  * @brief Create and set cq->ibv_cq_ex
@@ -269,6 +293,7 @@ static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *att
 		.wc_flags = IBV_WC_STANDARD_FLAGS,
 		.comp_mask = 0,
 	};
+
 	struct efadv_cq_init_attr efadv_cq_init_attr = {
 		.comp_mask = 0,
 		.wc_flags = EFADV_WC_EX_WITH_SGID,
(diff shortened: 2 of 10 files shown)


prov/efa/test/efa_unit_test_common.c
```diff
@@ -138,7 +138,7 @@ void efa_unit_test_buff_construct(struct efa_unit_test_buff *buff, struct efa_re
 	assert_non_null(buff->buff);
 
 	buff->size = buff_size;
-	err = fi_mr_reg(resource->domain, buff, buff_size, FI_SEND | FI_RECV,
+	err = fi_mr_reg(resource->domain, buff->buff, buff_size, FI_SEND | FI_RECV,
```
wzamazon, line 141: Interesting. How did it work before?

wenduwan: I'm not sure why but when we specify 4096 buff size it always works. When the size is increased to e.g. 8800, it fails ~20% of the time.

prov/efa/src/efa.h
```diff
@@ -192,10 +192,15 @@ struct efa_cq {
 	efa_cq_read_entry	read_entry;
 	ofi_spin_t		lock;
 	struct ofi_bufpool	*wce_pool;
+	uint32_t	comp_mask;
 
 	struct ibv_cq_ex	*ibv_cq_ex;
 };
 
+enum {
+	EFA_CQ_SUPPORT_EFADV_CQ = 1 << 0,
```
wzamazon, line 201: should be a macro, not enum

prov/efa/src/efa.h
```diff
@@ -192,10 +192,15 @@ struct efa_cq {
 	efa_cq_read_entry	read_entry;
 	ofi_spin_t		lock;
 	struct ofi_bufpool	*wce_pool;
+	uint32_t	comp_mask;
```
wzamazon, line 195: should be just `flags`.

prov/efa/src/efa.h
wenduwan: I'm curious about the difference? I see rdma-core prefers to use enum. Is it a convention to use macro in libfabric?

wzamazon: I think enum is for a list of constants that increases numerically 1 by 1.

But this one is being used as a bit mask. so should be a macro.

Also please use `BIT_ULL` to define it.

prov/efa/src/efa_cq.c
```diff
@@ -248,16 +250,38 @@ static struct fi_ops efa_cq_fi_ops = {
...
 #if HAVE_EFADV_CQ_EX
-static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
+/**
+ * @brief Create ibv_cq_ex by calling efadv_create_cq
+ *
+ * @param[in] cq Pointer to the efa_cq.
+ * @param[in] attr Pointer to fi_cq_attr.
+ * @param[out] Return pointer to ibv_cq_ex if successful, otherwise NULL
+ */
+static inline struct ibv_cq_ex *efa_cq_efadv_create_cq(struct efa_cq *cq, struct fi_cq_attr *attr) {
```
wzamazon, line 284: I am not sure we need this function.

Cann't we just call `efadv_create_cq` in the HAVE_EFADV_CREATE_CQ section of `efa_cq_set_ibv_cq_ex`?

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -301,11 +302,20 @@ static void test_impl_rdm_cq_read_unknow_peer_ah(bool remove_peer)
 	struct efa_unit_test_buff recv_buff;
 	int ret;
 
+	if (support_efadv_cq) {
+		g_efa_unit_test_mocks.efadv_create_cq = &efa_mock_efadv_create_cq_with_ibv_create_cq_ex;
```
wzamazon, line 306: If we want to test the case that support efadv cq, should not we use the real `efadv_create_cq` here?

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -301,11 +302,20 @@ static void test_impl_rdm_cq_read_unknow_peer_ah(bool remove_peer)
...
+		g_efa_unit_test_mocks.efadv_create_cq = &efa_mock_efadv_create_cq_set_eopnotsupp_and_return_null;
+		expect_function_call(efa_mock_efadv_create_cq_set_eopnotsupp_and_return_null);
+	}
+
 	efa_unit_test_resource_construct(&resource, FI_EP_RDM);
 
 	rxr_ep = container_of(resource.ep, struct rxr_ep, util_ep.ep_fid);
 	efa_cq = container_of(rxr_ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
 
+	EFA_WARN(FI_LOG_DEBUG, "&recv_buff: %p\n", &recv_buff);
```
wzamazon, line 318: Is this line needed?

prov/efa/src/efa.h
wenduwan: Will address. I was referring to examples like [this](https://github.com/linux-rdma/rdma-core/blob/master/providers/efa/efadv.h#L38-L42) in rdma-core which uses enum instead.

prov/efa/src/efa.h
wenduwan: Will address.

prov/efa/test/efa_unit_test_cq.c
wenduwan: Will remove. This was for debugging.

prov/efa/test/efa_unit_test_cq.c
wenduwan: We don't have to, because we also mock `efadv_cq->wc_read_sgid` on L356, so that we can use `ibv_create_cq_ex` instead of the real `efadv_create_cq`.

Using the real `efadv_create_cq` is also problematic in this case (ideally we want to detect the kernel module and set `support_efadv_cq`, but this is not easy). If we use the real `efadv_create_cq` we can get different test results depending on where we run the test:
* If the kernel supports EFA DV CQ and we set `support_efadv_cq = true`, then the test will pass
* If the kernel does **NOT** supports EFA DV CQ and we set `support_efadv_cq = true`, then the test will fail
So we want to avoid this uncertainty by mocking out `efadv_create_cq` entirely.

wenduwan:
AWS CI failed due to closed SSH connection

prov/efa/src/efa_cq.c
wenduwan: Sounds good. Will address.

prov/efa/src/efa.h
wenduwan: Addressed

prov/efa/src/efa.h
wenduwan: Addressed

prov/efa/src/efa_cq.c
wenduwan: Addressed

prov/efa/test/efa_unit_test_cq.c
wenduwan: Addressed

prov/efa/src/efa_cq.c
```diff
@@ -281,35 +306,45 @@ static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *att
...
+	if (cq->ibv_cq_ex) {
+		/* Support operations on EFA DV CQ */
+		cq->flags |= EFA_CQ_SUPPORT_EFADV_CQ;
+		return 0;
+	}
 
-	return cq->ibv_cq_ex ? 0 : -FI_ENOCQ;
+	/* This could be due to old EFA kernel module versions */
+	EFA_INFO(FI_LOG_CQ, "Unable to create EFA DV CQ: %s\n", strerror(errno));
+	cq->flags &= ~EFA_CQ_SUPPORT_EFADV_CQ;
```
wenduwan, line 321: Explicitly unset the bit

prov/efa/src/efa_cq.c
```diff
@@ -281,35 +306,45 @@ static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *att
...
+
+	if (cq->ibv_cq_ex) {
+		/* Support operations on EFA DV CQ */
+		cq->flags |= EFA_CQ_SUPPORT_EFADV_CQ;
+		return 0;
+	}
 
-	return cq->ibv_cq_ex ? 0 : -FI_ENOCQ;
+	/* This could be due to old EFA kernel module versions */
+	EFA_INFO(FI_LOG_CQ, "Unable to create EFA DV CQ: %s\n", strerror(errno));
```
wzamazon, line 320: I think `EFA_WARN` is better. It should not be something happen commonly for user.

prov/efa/test/efa_unit_test_cq.c
wzamazon: Make sense. Can you add these reasonings as comments.

wenduwan: Sure. Will address.

prov/efa/src/efa_cq.c
wenduwan: I kinda agree. I used `EFA_WARN` at the beginning but that made the unit test log noisy. Do you have a strong preference?

wzamazon: I prefer EFA_WARN. I am ok that we see some warning message when running unit test.

wenduwan: Addressed

prov/efa/test/efa_unit_test_cq.c
wenduwan: Addressed.

wzamazon: I did not see the comments?

wenduwan: Comment was in the wrong place. Updated again.

wzamazon (approved):
Thank you!

wzamazon:
bot:aws:retest

PR #7961, opened 2022-08-26
prov/verbs: Flush and save RX CQEs prior to destroying QP

chien-intel:
This patch fixes issue 7837.

Verbs provider is not polling CQ before destroying QP but CQEs
may not be available after a QP is destroyed.  Modify QP
to ERROR state to flush outstanding WRs and poll RX CQ until
RX queue depth entries are retrieved or exceed max retry.
Max retry is calculated based on RX queue depth and some HW
is populating CQEs in chunks of 64.

Current flush SQ code is left intact in favor of polling TX CQ
since it fixes a known issue.

### DIFF
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -297,32 +297,37 @@ int vrb_save_wc(struct vrb_cq *cq, struct ibv_wc *wc)
 	return FI_SUCCESS;
 }
 
-static void vrb_flush_cq(struct vrb_cq *cq)
+static void vrb_flush_cq(struct vrb_cq *cq, size_t rx_size)
 {
 	struct ibv_wc wc;
-	ssize_t ret;
+	size_t count = 0;
+	int retry = 0;
 
-	ofi_genlock_lock(&cq->util_cq.cq_lock);
-	while (1) {
-		ret = vrb_poll_cq(cq, &wc);
-		if (ret <= 0)
+	/* Some HW is populating CQEs in chunks of 64.
+	 * Set max_retry to 2 * rx_size/64 + 1
+	 */
+	int max_retry = rx_size/32 + 1;
+
+	do {
+		ofi_genlock_lock(&cq->util_cq.cq_lock);
+		while (vrb_poll_cq(cq, &wc) > 0) {
+			vrb_save_wc(cq, &wc);
+			count++;
+		}
+		ofi_genlock_unlock(&cq->util_cq.cq_lock);
+
+		if (count >= rx_size)
 			break;
-
-		vrb_save_wc(cq, &wc);
-	};
-
-	ofi_genlock_unlock(&cq->util_cq.cq_lock);
+		usleep(1);
+	} while (retry++ < max_retry);
 }
 
-void vrb_cleanup_cq(struct vrb_ep *ep)
+void vrb_flush_rx_cq(struct vrb_ep *ep)
 {
 	if (ep->util_ep.rx_cq) {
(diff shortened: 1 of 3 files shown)


prov/verbs/src/verbs_ep.c
```diff
@@ -515,14 +525,16 @@ static int vrb_ep_close(fid_t fid)
 			vrb_eq_remove_events(ep->eq, fid);
 		}
 
-		if (vrb_is_xrc_ep(ep))
+		if (vrb_is_xrc_ep(ep)) {
 			vrb_ep_xrc_close(ep);
-		else
+		}
+		else {
```
shefty, line 531: nit: We follow kernel coding guidelines for style.  Please write as:
```
} else {
```

prov/verbs/src/verbs_cq.c
```diff
@@ -297,32 +297,37 @@ int vrb_save_wc(struct vrb_cq *cq, struct ibv_wc *wc)
...
 {
 	if (ep->util_ep.rx_cq) {
 		vrb_flush_cq(container_of(ep->util_ep.rx_cq,
-					  struct vrb_cq, util_cq));
-	}
-	if (ep->util_ep.tx_cq) {
-		vrb_flush_cq(container_of(ep->util_ep.tx_cq,
-					  struct vrb_cq, util_cq));
+					  struct vrb_cq, util_cq),
+			     ep->info_attr.rx_size);
```
shefty, line 330: This assumes that there are rx_size entries posted to the QP at the time it is destroyed.  That seems unlikely, in which case we'll end up stalling endpoint destruction for several us.  It would be better to know exactly how many completions we're expecting, or to have them queued, so we can just flush them.

prov/verbs/src/verbs_cq.c
```diff
@@ -297,32 +297,37 @@ int vrb_save_wc(struct vrb_cq *cq, struct ibv_wc *wc)
...
 
-	ofi_genlock_lock(&cq->util_cq.cq_lock);
-	while (1) {
-		ret = vrb_poll_cq(cq, &wc);
-		if (ret <= 0)
-			break;
+	/* Some HW is populating CQEs in chunks of 64.
+	 * Set max_retry to 2 * rx_size/64 + 1
+	 */
+	int max_retry = rx_size/32 + 1;
```
shefty, line 309: Please declare variables at the top of all functions.  If you need a comment, do that and the assignment after the declaration.  In this case, the comment doesn't match the code exactly, which looks off on first glance.  And the reasoning for how max_retry is determined is questionable.  The code is assuming that the HW will write 64 entries to the CQ every 1 us, kind of.

That a sleep was added to the code is a strong indicator that the code is racy and the sleep simply helps us to get luckier.  (And making it far more difficult to debug later if we hit into races.)  If we need to flush all posted receive operations and the HW isn't doing that in a way that usable, then we should simply track the receives in SW and flush them manually after the CQ has been emptied (after the QP has been destroyed).  That's guaranteed to work without spin sleeping and hoping the HW writes the completions under load within our timing window.

shefty:
The receive side can follow a similar model as the send side.  See vrb_post_send() and vrb_post_recv().  In vrb_post_send(), the ctx structure is queued to the ep->sq_list.  We need a similar ep->rq_list.  After calling ibv_post_recv(), on success insert the ctx on the rq_list.  Add a vrb_flush_rq(), similar to vrb_flush_sq(), and called from the same locations.  Finally, update vrb_poll_cq() to remove the ctx from the rq_list on completion, also similar to what's done for the sq_list.

shefty:
See #7981 -- patch there isn't tested, but I believe will do what we and flush all posted receives.

PR #7964, opened 2022-08-27
prov/efa: Do not fail domain open when neuron_alloc failed.

shijin-aws:
neuron_alloc will fail if application did not call nrt_init,
which is ok if it's not running neuron workloads. libfabric
will move on and leave neuron_status->initialized as false.

Also add a unit test for this fix

### DIFF
--- a/prov/efa/test/efa_unit_test_hmem.c
+++ b/prov/efa/test/efa_unit_test_hmem.c
@@ -1,0 +1,53 @@
+#include "efa_unit_tests.h"
+
+
+#if HAVE_NEURON
+/**
+ * @brief Verify when neuron_alloc failed (return null),
+ * efa_domain_open, which call efa_hmem_support_status_update_neuron
+ * when HAVE_NEURON=1, will still return 0 but leave 
+ * hmem_support_status[FI_HMEM_NEURON].initialized and 
+ * hmem_support_status[FI_HMEM_NEURON].p2p_supported as false.
+ */
+void test_efa_hmem_support_status_update_neuron()
+{
+        int ret;
+        struct efa_resource resource = {0};
+        struct efa_domain *efa_domain;
+        uint32_t efa_device_caps_orig;
+        bool neuron_initialized_orig;
+
+        resource.hints = efa_unit_test_alloc_hints(FI_EP_RDM);
+        assert_non_null(resource.hints);
+
+        ret = fi_getinfo(FI_VERSION(1, 14), NULL, NULL, 0ULL, resource.hints, &resource.info);
+        assert_int_equal(ret, 0);
+
+        ret = fi_fabric(resource.info->fabric_attr, &resource.fabric, NULL);
+        assert_int_equal(ret, 0);
+
+        neuron_initialized_orig = hmem_ops[FI_HMEM_NEURON].initialized;
+        hmem_ops[FI_HMEM_NEURON].initialized = true;
+        efa_device_caps_orig = g_device_list[0].device_caps;
+        g_device_list[0].device_caps |= EFADV_DEVICE_ATTR_CAPS_RDMA_READ;
+        g_efa_unit_test_mocks.neuron_alloc = &efa_mock_neuron_alloc_return_null;
+
+        ret = fi_domain(resource.fabric, resource.info, &resource.domain, NULL);
+
+        /* recover the modified global variables before doing check */
+        hmem_ops[FI_HMEM_NEURON].initialized = neuron_initialized_orig;
+        g_device_list[0].device_caps = efa_device_caps_orig;
+
+        assert_int_equal(ret, 0);
+        efa_domain = container_of(resource.domain, struct efa_domain,
+				  util_domain.domain_fid.fid);
+        assert_false(efa_domain->hmem_support_status[FI_HMEM_NEURON].initialized);
+        assert_false(efa_domain->hmem_support_status[FI_HMEM_NEURON].p2p_supported);
+        efa_unit_test_resource_destruct(&resource);
+}
+#else
+void test_efa_hmem_support_status_update_neuron()
+{
+        skip();
+}
+#endif /* HAVE_NEURON */
--- a/prov/efa/Makefile.include
+++ b/prov/efa/Makefile.include
@@ -104,7 +104,8 @@ nodist_prov_efa_test_efa_unit_test_SOURCES = \
 	prov/efa/test/efa_unit_test_av.c \
 	prov/efa/test/efa_unit_test_cq.c \
 	prov/efa/test/efa_unit_test_device.c \
-	prov/efa/test/efa_unit_test_info.c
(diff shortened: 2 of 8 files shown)


prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -0,0 +1,48 @@
...
+        resource.hints = efa_unit_test_alloc_hints(FI_EP_RDM);
+        assert_non_null(resource.hints);
+
+        ret = fi_getinfo(FI_VERSION(1, 14), NULL, NULL, 0ULL, resource.hints, &resource.info);
+        assert_int_equal(ret, 0);
+
+        ret = fi_fabric(resource.info->fabric_attr, &resource.fabric, NULL);
+        assert_int_equal(ret, 0);
+
+        hmem_ops[FI_HMEM_NEURON].initialized = true;
```
wckzhang, line 27: Can you explain why you need to set this?

shijin-aws: This is to ensure `ofi_hmem_is_initialized()` https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_hmem.c#L110-L122 return true so `efa_hmem_support_status_update_neuron` can reach the `neuron_alloc` line.

prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -0,0 +1,48 @@
...
+        ret = fi_domain(resource.fabric, resource.info, &resource.domain, NULL);
+        assert_int_equal(ret, 0);
+
+        efa_domain = container_of(resource.domain, struct efa_domain,
+				  util_domain.domain_fid.fid);
+        assert_false(efa_domain->hmem_support_status[FI_HMEM_NEURON].initialized);
+        assert_false(efa_domain->hmem_support_status[FI_HMEM_NEURON].p2p_supported);
+        
+        /* need to reset this to avoid dlclose (in ofi_hmem_cleanup) being called */
+        hmem_ops[FI_HMEM_NEURON].initialized = false;
```
wzamazon, line 40: Do you need to also reset `g_device_list[0].device_caps`?

shijin-aws: Yes, added in updated revision. Thank you!

wckzhang (approved):

shijin-aws:
bot:aws:retest

prov/efa/test/efa_unit_test_hmem.c
wzamazon: I think you need to save original value of `g_device_list[0].device_caps`?

shijin-aws: I see, save and recover the original value in updated revision.

prov/efa/test/efa_unit_test_hmem.c
wzamazon: It seems this value need to be recovered after test too. Sorry I missed this earlier.

prov/efa/test/efa_unit_test_hmem.c
wenduwan: The problem with resetting value after `assert_xxx` is that it won't be called if assertion fails. We have 2 options:
1. Still reset state inside the test, but reset the state immediately after it has taken effect and before assertions
2. Reset the state outside the test, in a cleanup hook, e.g. `efa_unit_test_mocks_reset` or similar.

prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -0,0 +1,51 @@
...
+        g_device_list[0].device_caps = efa_device_caps_orig;
+        efa_unit_test_resource_destruct(&resource);
+}
+#else
+void test_efa_hmem_support_status_update_neuron()
+{
+        skip();
+}
+#endif /* HAVE_NEURON */
```
wenduwan, line 51: nit missing new line at the end of file

prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -0,0 +1,51 @@
...
+        resource.hints = efa_unit_test_alloc_hints(FI_EP_RDM);
+        assert_non_null(resource.hints);
+
+        ret = fi_getinfo(FI_VERSION(1, 14), NULL, NULL, 0ULL, resource.hints, &resource.info);
+        assert_int_equal(ret, 0);
+
+        ret = fi_fabric(resource.info->fabric_attr, &resource.fabric, NULL);
+        assert_int_equal(ret, 0);
+
+        hmem_ops[FI_HMEM_NEURON].initialized = true;
```
wenduwan, line 28: Clarify: in the current revision we are calling `fi_xxx` to initialize hints, fabric, etc., and we cannot do this instead:
```c
...
hmem_ops[FI_HMEM_NEURON].initialized = true;
efa_unit_test_resource_construct(&resource, FI_EP_RDM);
...
```
because we need this before `fi_domain` call?
```c
efa_device_caps_orig = g_device_list[0].device_caps;
g_device_list[0].device_caps |= EFADV_DEVICE_ATTR_CAPS_RDMA_READ;
```

prov/efa/test/efa_unit_test_hmem.c
shijin-aws: For number 2, how do you know the original value before the test modifying the variables?

prov/efa/test/efa_unit_test_hmem.c
shijin-aws: Correct, we need to modify these global variables after `fi_fabric` and before `fi_domain`, so we cannot call `efa_unit_test_resource_construct`

prov/efa/test/efa_unit_test_hmem.c
shijin-aws: I will do number 1 for this PR.

prov/efa/test/efa_unit_test_hmem.c
shijin-aws: It was resetting it to `false`. But I changed it to save the original value and then recover to orig value, in case the original value is `true`.

prov/efa/test/efa_unit_test_hmem.c
shijin-aws: Fixed.

prov/efa/test/efa_unit_test_hmem.c
wenduwan: For option 2, I think the easier(cleaner) way is to make sure every test starts with the same global state, including `g_device_list`. Basically we want to take out the guess work and hard code a value.

wenduwan (approved):

wzamazon (approved):

wzamazon:
bot:aws:retest

shijin-aws:
Ignoring Intel CI as the change is within efa provider only...

PR #7965, opened 2022-08-29
prov/efa: use same strategy to select RTM for neuron and cuda

wzamazon:
This patch renamed

    rxr_msg_select_rtm_for_cuda()

to

    rxr_msg_select_rtm_for_hmem()

and used it for neuron memory too.

### DIFF
--- a/prov/efa/src/rxr/rxr_msg.c
+++ b/prov/efa/src/rxr/rxr_msg.c
@@ -56,18 +56,15 @@
  */
 
 /**
- * @brief select a two-sided protocol for the send operation when send buffer is on cuda memory
+ * @brief select a two-sided protocol for the send operation when send buffer is on hmem memory
  *
  * @param[in]		rxr_ep		endpoint
  * @param[in]		tx_entry	contains information of the send operation
  * @return		the RTM packet type of the two-sided protocol.
- *                      Four types of protocol can be used: eager, medium, longread or runtread.
  *                      Each protocol has tagged/non-tagged version. Some protocols has a DC version.
- *
- *                      Note longcts protocol is not suited for cuda memory and will not be used.
  */
 static inline
-int rxr_msg_select_rtm_for_cuda(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry)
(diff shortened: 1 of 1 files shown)


prov/efa/src/rxr/rxr_msg.c
```diff
@@ -56,18 +56,15 @@
  */
 
 /**
- * @brief select a two-sided protocol for the send operation when send buffer is on cuda memory
+ * @brief select a two-sided protocol for the send operation when send buffer is on hmem memory
  *
  * @param[in]		rxr_ep		endpoint
  * @param[in]		tx_entry	contains information of the send operation
  * @return		the RTM packet type of the two-sided protocol.
- *                      Four types of protocol can be used: eager, medium, longread or runtread.
```
wckzhang, line 64: Why did you remove these comments? (also line 67

wckzhang (approved):

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -154,10 +151,9 @@ int rxr_msg_select_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry, int
 		return RXR_EAGER_MSGRTM_PKT + tagged;
 	}
 
-	if (use_p2p && efa_mr_is_cuda(tx_entry->desc[0]))
-		return rxr_msg_select_rtm_for_cuda(rxr_ep, tx_entry);
+	if (use_p2p && efa_mr_is_hmem(tx_entry->desc[0]))
```
shijin-aws, line 154: I do not think we will do runting read for habana?

shijin-aws: I think I could just insert the habana block before this `if`

wzamazon: Hababa is not in the code base yet. so I did not include it.

After habana is added, I think it is better to handle it inside the function `select_rtm_for_hmem`.

shijin-aws: Hmm I think I could just insert the habana block before this `if`.

shijin-aws: @wzamazon That sounds good. Thank you!

shijin-aws (approved):

PR #7967, opened 2022-08-31
prov/efa: Introduce rxr_env.enable_eager_message

shijin-aws:
Currently, there is no way to disable eager message
protocol when message size is < eager_rtm_max_data_size.

This pathc introduces rxr_env.enable_eager_message, which
defines whether eager message protocol is
enabled. It has a default value as 1 and can be
overriden by specifying the FI_EFA_ENABLE_EAGER_MESSAGE
environment variable.

### DIFF
--- a/prov/efa/src/rxr/rxr_env.c
+++ b/prov/efa/src/rxr/rxr_env.c
@@ -42,6 +42,7 @@ struct rxr_env rxr_env = {
 	.tx_min_credits = RXR_DEF_MIN_TX_CREDITS,
 	.tx_queue_size = 0,
 	.enable_shm_transfer = 1,
+	.enable_eager_message = 1,
 	.use_device_rdma = 0,
 	.use_zcpy_rx = 1,
 	.zcpy_rx_seed = 0,
@@ -96,6 +97,7 @@ void rxr_env_param_get(void)
 
 	fi_param_get_int(&rxr_prov, "tx_queue_size", &rxr_env.tx_queue_size);
 	fi_param_get_int(&rxr_prov, "enable_shm_transfer", &rxr_env.enable_shm_transfer);
+	fi_param_get_int(&rxr_prov, "enable_eager_message", &rxr_env.enable_eager_message);
 	fi_param_get_int(&rxr_prov, "use_device_rdma", &rxr_env.use_device_rdma);
 	fi_param_get_int(&rxr_prov, "use_zcpy_rx", &rxr_env.use_zcpy_rx);
 	fi_param_get_int(&rxr_prov, "zcpy_rx_seed", &rxr_env.zcpy_rx_seed);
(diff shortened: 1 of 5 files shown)


prov/efa/src/rxr/rxr_env.c
```diff
@@ -42,6 +42,7 @@ struct rxr_env rxr_env = {
 	.tx_min_credits = RXR_DEF_MIN_TX_CREDITS,
 	.tx_queue_size = 0,
 	.enable_shm_transfer = 1,
+	.enable_eager_message = 1,
```
wzamazon, line 45: maybe just `enable_eager` or `efa_eager_protocol`?

bwbarrett: Why have an enable at all?  Just have a max_eager_size.  0 == no eager protocol.  Much more flexible, for no real cost.

wzamazon: `max_eager_size` is not a constant though. It depends on packet size and header size.

wckzhang: @bwbarrett I think that runs into an issue with the medium message protocol. I'm assuming when you say no eager, you don't want that executing either.

prov/efa/src/rxr/rxr_msg.c
```diff
@@ -172,7 +172,7 @@ int rxr_msg_select_rtm(struct rxr_ep *rxr_ep, struct rxr_tx_entry *tx_entry, int
 
 	eager_rtm_max_data_size = rxr_tx_entry_max_req_data_capacity(rxr_ep, tx_entry, eager_rtm);
 
-	if (tx_entry->total_len <= eager_rtm_max_data_size)
+	if (rxr_env.enable_eager_message && tx_entry->total_len <= eager_rtm_max_data_size)
 		return eager_rtm;
 
 	if (tx_entry->total_len <= rxr_env.efa_max_medium_msg_size)
```
wckzhang, line 178: Maybe I'm not understanding the need here. If we don't want to send eager packets, why would we want to execute the medium message protocol (which is just sending a bunch of eager messages iirc).

prov/efa/src/rxr/rxr_env.c
bwbarrett: So make `max_eager_size` a run-time constant.  You know the packet size fairly early in initializing the EFA device.  THe difference between the max and min header size is a small fraction of expected packet sizes.  Save yourself the complexity (and computation time!) and just figure out the max header size during initialization, then use that as the eager limit.  Don't over-optimize for a couple bytes on the eager / medium threshold.

bwbarrett: For the medium / runting (note: I don't think you need a medium AND a runting protocol.  The runting protocol really is what the medium protocol always should have been (assuming the data in flight limit is sane).  But in that case, you'd set a threshold for the max runting / max medium threshold, and to avoid copies entirely, you'd set both the eager and runting / medium message to 0.

prov/efa/src/rxr/rxr_msg.c
shijin-aws: It is for experimental purpose. Users can further disable medium by setting the `max_medium_size` to 0

prov/efa/src/rxr/rxr_env.c
shijin-aws: I agree with Brian, to avoid copies entirely, you need to set both the eager and runting / medium message to 0.

wenduwan:
bot:aws:retest

PR #7968, opened 2022-09-01
prov/efa: remove redundent call to efa_mr_hmem_setup()

wzamazon:
This patch removed the call to efa_mr_hmem_setup() in
efa_mr_cache_regattr() because efa_mr_cache_regattr() already
called the function indirectly when it called ofi_mr_cache_search()

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -374,10 +374,6 @@ static int efa_mr_cache_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	efa_mr = (struct efa_mr *)entry->data;
 	efa_mr->entry = entry;
 
-	ret = efa_mr_hmem_setup(efa_mr, attr);
-	if (ret)
-		return ret;
-
 	*mr_fid = &efa_mr->mr_fid;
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -374,10 +374,6 @@ static int efa_mr_cache_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	efa_mr = (struct efa_mr *)entry->data;
 	efa_mr->entry = entry;
 
-	ret = efa_mr_hmem_setup(efa_mr, attr);
```
shijin-aws, line 377: So it's because `ofi_mr_cache_search` call `add_region`, which is `efa_mr_reg_impl`, which calls `efa_mr_hmem_setup`. Right?

shijin-aws (approved):

prov/efa/src/efa_mr.c
wzamazon: Yes.

wzamazon:
bot:aws:retest

wckzhang (approved):

PR #7969, opened 2022-09-01
prov/efa: set FI_HMEM_DEVICE_ONLY flag for hmem memory for shm

wzamazon:
This patch set the FI_HMEM_DEVICE_ONLY flag when register
a hmem region with shm provider because shm provider needs the
flag to determine IPC usage.


prov/efa/src/efa_mr.c
```diff
@@ -551,8 +552,11 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 		*/
 		original_access = mr_attr->access;
 		mr_attr->access |= FI_REMOTE_READ;
+		shm_flags = flags;
+		if (mr_attr->iface != FI_HMEM_SYSTEM)
```
shijin-aws, line 556: Could you add a comment here to call it out as well?

wzamazon: added

shijin-aws (approved):

PR #7970, opened 2022-09-01
hmem_cuda: set SYNC_MEMOPS for cuda pointer when getting ipc handle.

shijin-aws:
Set this attribute to ensure any synchronous copies
are completed prior to any other access of the memory,
region, which ensure the data consistency for CUDA IPC.

This is also what OpenMPI/smcuda does.


shijin-aws:
We have evaluated this PR does not work as expected. Closing..

PR #7978, opened 2022-09-02
prov/efa: add support for synapseai hmem iface

shijin-aws:
Use long read for msg and rma operations, regardless of what is
specified by the user for protocol switch over points.
There is no memcpy from host to device for synapseai inside
libfabric.

Test info: 
- Run hccl_demo with allreduce, broadcast, reduce
- Run fabtests between synapseai buffer on two dl1 instances.


wzamazon:
target for 1.16.0

wzamazon (approved):

PR #7979, opened 2022-09-02
prov/rxm, util, verbs, efa: add iface initialization checks to MR registration

### LINKED ISSUE #7977: prov/verbs: fi_mr_regattr() ignores `iface = FI_HMEM_CUDA` when should fail
**Describe the bug**

I was surprised (and disappointed) to find that in an install of libfabric configured without `--with-cuda` my calls to `fi_mr_regattr()` with verbs provider and `attr.iface == FI_HMEM_CUDA` succeed!

When the resulting (non-NULL) `fi_mr_desc()` result is used in a call to `fi_write()` with the source in CUDA memory, I find "regular" `memcpy()` being used (as seen in the backtrace from the SIGSEGV which results).  I would have hoped to have the memory registration fail, instead of providing a `desc` that cannot be used for the intended purpose.

**To Reproduce**

The attached reproducer opens verbs provider and _successfully_ registers CUDA memory, regardless of whether HMEM CUDA support is present or absent.

The following will compile the attachment, assuming correct values for OFI_HOME and CUDA_HOME:
```
gcc hmem_test.c -o hmem_test -I$OFI_HOME/include -I$CUDA_HOME/include -L$OFI_HOME/lib -L$CUDA_HOME/lib -lfabric -lcuda
```

An example run is shown in the **Output**, below

**Expected behavior**

In a build of libfabric verbs provider, configured without `--with-cuda`, I would expect that calls to `fi_mr_regattr()` with `attr.iface == FI_HMEM_CUDA` would fail.  In issue #7973 @shefty has **also** stated that `ENOSYS` is the expected result.

**Output**

Below is an example run with `FI_LOG_LEVEL=debug` used to confirm that `FI_HMEM_CUDA` support is *absent*, yet the test runs to completion having *not* seen a `ENOSYS` (nor `-FI_ENOSYS`) return from `fi_mr_regattr()`.  This is with `verbs;ofi_rxm` provider in libfabric 1.15.1.

```
$ env FI_LOG_LEVEL=debug ./hmem_test |& grep -e CUDA -e FATAL
libfabric:15958:1662089396::core:core:ofi_hmem_init():229<info> Hmem iface FI_HMEM_CUDA not supported
libfabric:15958:1662089396::core:core:ofi_hmem_init():229<info> Hmem iface FI_HMEM_CUDA not supported
Done: "successfully" registered CUDA memory
```

This result is "wrong" because the expected result is an error return from the call to `fi_mr_regattr()`.

**Environment:**

Linux  
`verbs;ofi_rxm` provider in libfabric versions from 1.12.1 to 1.15.1  
RDM endpoint  

**Additional context**

Note that my experience and report are specific to `verbs` provider and `FI_HMEM_CUDA` because that is what I have reliable access to.  However, I don't have any reason to think this problem is limited to a specific provider or hmem iface value.

aingerson:
Providers should fail MR registration for ifaces that are not initialized

Fixes #7977 

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -360,6 +360,12 @@ static int efa_mr_cache_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 		return -FI_EINVAL;
 	}
 
+	if (!ofi_hmem_is_initialized(attr->iface)) {
+		EFA_WARN(FI_LOG_MR,
+			 "Cannot register memory for uninitialized iface\n");
+		return -FI_ENOSYS;
+	}
+
 	domain = container_of(fid, struct efa_domain,
 			      util_domain.domain_fid.fid);
 
@@ -618,6 +624,12 @@ static int efa_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 		EFA_WARN(FI_LOG_MR, "iov count > %d not supported\n",
 			 EFA_MR_IOV_LIMIT);
 		return -FI_EINVAL;
+	}
+
+	if (!ofi_hmem_is_initialized(attr->iface)) {
+		EFA_WARN(FI_LOG_MR,
+			 "Cannot register memory for uninitialized iface\n");
+		return -FI_ENOSYS;
 	}
 
 	domain_fid = container_of(fid, struct fid_domain, fid);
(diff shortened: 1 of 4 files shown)


shefty (approved):
thanks!

wzamazon (approved):
EFA changes is good to me.

PHHargrove:
I have confirmed that verbs provider built from this branch produces the expected behavior for my reproducer in #7977

PHHargrove:
Thanks @shefty and @wzamazon 
Can I ask when this is likely to appear in a release?  Perhaps 1.16.0?

shefty:
Yes, this will get pulled into v1.16

PR #7986, opened 2022-09-06
fabtests/component/dmabuf-rdma: Add option to prepost receives

j-xiong:
Add an option to the fi-rdmabw-xe test to allow preposting receives
for the send test. Some providers may see a performance boost with
preposted receives due to the overhead associated with unexpected
messages.

Also fix the requested MR keys.


shefty:
This code obviously targets RDMA NICs, so why not just make this the default and not bother with a command line option?

j-xiong:
The option is needed to control the number of receives being pre-posted. We could use a non-zero default, though.

PR #7988, opened 2022-09-07
prov/efa: Use eager protocol for shm when msg size is 0

shijin-aws:
It is efficient to use eager for 0 bytes transfer. This patch
makes rxr_msg_select_rtm use eager protocol for shm when
msg size is 0.


wzamazon:
>We have observed fi_read via shm will cause random hang when msg size is 0. 

I do not think this issue matters.

Even without this issue, it is still more efficient to use eager for 0 bytes

shijin-aws:
@wzamazon Thank you, updated the commit message and the PR description.

wzamazon (approved):

PR #7994, opened 2022-09-13
contrib/intel/jenkins: Add functional summary stage to CI

Juee14Desai:
This commit adds a functional summary to the CI.
A new directory (log_dir) is being made to store all the logs for each stage.
These patches add a functional summary stage to the pipeline and to the post stage in the Jenkinsfile.
The pipeline stage is only viewable on passed pipelines. This is because jenkins skips future stages if earlier stages failed. So, this stage is just to provide a clear place to look on passed pipelines for the functional summary.
The functional summary will still be provided in the post stage inside of the always{} clause incase of a test failure where the summary will still provide a list of passed and failed tests during the entire run.
The stages that get summarized in these patches are:
fabtests (tcp, verbs-rxm, verbs-rxd, shm, udp, sockets, psm3) [reg, dl, dbg build modes]
fabtests_ze (shm, h2d, d2d) [reg]
shmem (uh, prk, isx) [reg]
verbs-rxm impi IMB [reg]
verbs-rxm mpich IMB [reg]
verbs-rxm impi osu [reg]
verbs-rxm mpich osu [reg]
tcp-rxm impi IMB [reg]
tcp-rxm impi osu [reg]
verbs-rxm mpich mpichtestsuite [reg]
verbs-rxm impi mpichtestsuite [reg]
tcp impi mpichtestsuite [reg]
sockets impi mpichtestsuite [reg]
oneccl [reg]
oneccl gpu [reg]
These combinations are all of the ones that currently get run in the Jenkinsfile however the summarizer has the capability to summarize additional combinations once they get re-enabled.

There are also some miscellaneous changes like:
common.run_command has been made versatile to accommodate new log filenames.
Separating the ze-shm stage into three log files the same way that shmem is for easier summarizing.
Changes to the cleanup path in the Jenkinsfile to always cleanup everything created per run.
This is because jenkins doesn't cleanup the ofi-libfabric workspace and only cleans up its checkout workspace.
This makes the space that the libfabric builds, the middlewares directory, and the logs directory take up get freed
at the end of the run.
If there was a failure general user's do not care about having the build artifacts because they
can be easily remade.
And in the case of developers re-enabling failing tests, they already build these things by hand
and use jenkins to test if its working via automation.
TLDR: It is no longer necessary to keep the build artifacts post run

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -56,11 +56,11 @@ pipeline {
                     check_target()
                 }
                 withEnv(['PATH+EXTRA=/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin/:$PYTHONPATH']) {
-                  sh """
-                    mkdir ${env.WORKSPACE}/py_scripts
-                    git clone ${env.UPSTREAM} ${env.WORKSPACE}/py_scripts
-                    ${env.SKIP_PATH}/skip.sh ${env.WORKSPACE} ${TARGET}
-                  """
+                  	sh """
+                  		mkdir ${env.WORKSPACE}/py_scripts
+                    	git clone ${env.UPSTREAM} ${env.WORKSPACE}/py_scripts
+                    	${env.SKIP_PATH}/skip.sh ${env.WORKSPACE} ${TARGET}
+                  	"""
                 }
                 script {
                     DO_RUN=skip()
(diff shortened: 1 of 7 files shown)


zachdworkin (approved):

shefty:
This PR adds 2 binary files to the git tree.

shefty:
The PR also includes a merge commit.  Please update to remove the merge commit and binaries.

zachdworkin:
@shefty does this look good now?

PR #8003, opened 2022-09-15
prov/efa: Report correct default for runt_size parameter

ryanamazon:
The runt_size parameter default changed, but the reported default did not.  Update the reported default to the actual default.


wzamazon (approved):

wzamazon:
bot:aws:retest

PR #8011, opened 2022-09-17
[v1.16.x] net: Fix for multi-recv. core: fix iface strcmp.  fabtest: Fix multi-recv test

shefty:
core: Fix NIC interface name checks to be exact match, versus prefix match
prov/net: Fix support for FI_MULTI_RECV
fabtests/multi_recv: Fix crash in cleanup and add buffer validation


shefty:
AWS failure was related to sockets provider.  Re-running Intel CI

PR #8017, opened 2022-09-19
Opx upstream September

timothom64:


timothom64:
Please add the [for-v1.16.x](https://github.com/ofiwg/libfabric/labels/for-v1.16.x) label

shefty:
Please open against v1.16.x to merge prior to rc3.  Thanks!

timothom64:
> Please open against v1.16.x to merge prior to rc3. Thanks!

Oh, I'm fine with this commit going into main and rc3 and not being patched into the previous RCs, which I think is what will happen if we just proceed as is.  

There's no need to port this to other branches/older RCs.

shefty:
I already created the v1.16.x branch.  So, if you want the changes in the v1.16.0 release, it needs to be picked into that branch.

PR #8022, opened 2022-09-21
util,verbs: FI_SOCKADDR includes support of FI_SOCKADDR_IB

sydidelot:
According to libfabric's man pages (fi_verbs.7), MSG EPs in the verbs  provider support FI_SOCKADDR, FI_SOCKADDR_IN, FI_SOCKADDR_IN6 and FI_SOCKADDR_IB.

Since commit fb758f5d7("prov/verbs: set and check address format correctly"), the verbs provider sets fi->addr_format = FI_SOCKADDR for MSG EPs.
The problem is that applications requesting FI_SOCKADDR_IB get FI_ENODATA in gi_getinfo() because ofi_valid_addr_format() only expects FI_SOCKADDR, FI_SOCKADDR_IN and FI_SOCKADDR_IN6 but not FI_SOCKADDR_IB.

### DIFF
--- a/prov/verbs/src/verbs_info.c
+++ b/prov/verbs/src/verbs_info.c
@@ -771,7 +771,7 @@ static int vrb_alloc_info(struct ibv_context *ctx, struct fi_info **info,
 		fi->caps = VERBS_MSG_CAPS;
 		*(fi->tx_attr) = verbs_tx_attr;
 		*(fi->rx_attr) = verbs_rx_attr;
-		fi->addr_format = FI_SOCKADDR;
+		fi->addr_format = FI_SOCKADDR_IB;
 		break;
 	case FI_EP_DGRAM:
 		fi->caps = VERBS_DGRAM_CAPS;


sydidelot:
@shefty could you please review this change and cherry-pick it to 1.16.x?
Also, would it be possible to add a test for `FI_SOCKADDR_IB` in CI pipeline to avoid regressions? It looks like we don't have any yet.

shefty:
I'll include fix prior to 1.16 release.  Thanks

prov/util/src/util_attr.c
```diff
@@ -47,8 +47,8 @@ int ofi_valid_addr_format(uint32_t prov_format, uint32_t user_format)
 
 	switch (prov_format) {
 	case FI_SOCKADDR:
-		/* Provider supports INET and INET6 */
-		return user_format <= FI_SOCKADDR_IN6;
+		/* Provider supports IB, INET, and INET6 */
+		return user_format <= FI_SOCKADDR_IB;
```
shefty, line 51: This change makes FI_SOCKADDR equivalent to FI_SOCKADDR_IB.  I wonder if the provider shouldn't be using FI_SOCKADDR_IB instead of FI_SOCKADDR.  My concern here is that providers that support IPv4 and IPv6 only (udp, tcp, net) will now end up accepting IB addresses.

sydidelot: You mean change this line https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L774
and have `fi->addr_format = FI_SOCKADDR_IB;`?

shefty: Yes, that is what I meant.  It looks like that would align the code with the comments in this function.  But let me look through the rest of the code to see what makes sense.  Your change makes sense from reading the API; I'm just unsure if that's what we want.  (FI_SOCKADDR appears to be used more like FI_SOCKADDR_INX or FI_SOCKADDR_IP.)

sydidelot: Understood, thanks! 
I will run some tests with your suggestion and post my results here.

shefty: I looked through the code, and if changing verbs_info.c like you mentioned above works, that's the route we should take.  There's no clean way for a provider to indicate that they only support IP addressing (needed by udp, net, tcp), so they use FI_SOCKADDR in their fi_getinfo filter.

Technically, those providers really do support a generic sockaddr structure, limited by the kernel.  But if we can filter at fi_getinfo, then that prevents an app from trying to use, say, sockaddr_ib with the tcp provider and dealing with the failure at endpoint creation.

A better longer term solution would be to define, say, FI_SOCKADDR_IP or OFI_SOCKADDR_IP and using that as a more selective filter.  The 'OFI' definition would be if we want to keep this internal.

sydidelot: I tested the change `fi->addr_format = FI_SOCKADDR_IB` in the verbs provider and it works for my use case! 
I have updated the patch accordingly. Thank you!

shefty:
added PR #8025 to pick into v1.16.x

PR #8028, opened 2022-09-26
fabtests/prov/efa: Add an opt to work with triggered RNR

miaoneng:
Add a switch so this test can work with the hooked RDMA core which returns RNR forcefully. New `-k` will set program to send only 1 packet and should expect 1 captured RNR error.

Signed-off-by: Jingyin Tang <jytang@amazon.com>

### DIFF
--- a/fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
+++ b/fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
@@ -38,6 +38,7 @@
 #include <shared.h>
 #include "efa_rnr_shared.h"
 
+static int __use_rnr_hook = 0;
 
 static int rnr_read_cq_error(void)
 {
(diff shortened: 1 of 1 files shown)


shefty:
Can you update the commit message to indicate what module the changes apply to?  e.g. fabtests/efa:

fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
```diff
@@ -162,13 +163,16 @@ int main(int argc, char **argv)
...
-	while ((op = getopt(argc, argv, ADDR_OPTS INFO_OPTS CS_OPTS)) != -1) {
+	while ((op = getopt(argc, argv, "k" ADDR_OPTS INFO_OPTS CS_OPTS)) != -1) {
 		switch (op) {
 		default:
 			ft_parse_addr_opts(op, optarg, &opts);
 			ft_parseinfo(op, optarg, hints, &opts);
 			ft_parsecsopts(op, optarg, &opts);
 			break;
+		case 'k':
+			__use_rnr_hook = 1;
```
wzamazon, line 174: I think it is probably better to reuse the `-I` option, which stands for number of iterations (or number of messages to send).

set the default to `(fi->rx_attr->size + expected_rnr_error)` at the beginning, so it can be overwritten by command line options.

PR #8036, opened 2022-09-27
prov/rxm: Add disconnect to rxm_av_remove

chien-intel:
Reduce resource consumption by disconnecting connections to peers on AV remove.  This may abruptly terminate any data transfers in progress.

To enable, set environment variable FI_OFI_RXM_AV_REMOVE_DISC to 1.

### DIFF
--- a/prov/util/src/rxm_av.c
+++ b/prov/util/src/rxm_av.c
@@ -31,7 +31,6 @@
  */
 
 #include <ofi_coll.h>
-
 #include "ofi_util.h"
 #include "uthash.h"
 
@@ -119,14 +118,19 @@ util_get_peer(struct rxm_av *av, const void *addr)
 	return peer;
 }
 
+static void _util_put_peer(struct util_peer_addr *peer)
+{
+	if (--peer->refcnt == 0)
+		rxm_free_peer(peer);
+}
+
 void util_put_peer(struct util_peer_addr *peer)
 {
 	struct rxm_av *av;
 
 	av = peer->av;
 	ofi_mutex_lock(&av->util_av.lock);
-	if (--peer->refcnt == 0)
-		rxm_free_peer(peer);
+	_util_put_peer(peer);
 	ofi_mutex_unlock(&av->util_av.lock);
 }
 
@@ -205,12 +209,16 @@ static int rxm_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
 			 size_t count, uint64_t flags)
 {
 	struct util_av_entry *av_entry;
+	struct util_peer_addr **peer;
+	struct util_ep *util_ep;
+	struct dlist_entry *item;
 	struct rxm_av *av;
 	ssize_t i;
 
-	av = container_of(av_fid, struct rxm_av, util_av.av_fid);
 	if (flags)
 		return -FI_EINVAL;
+
+	av = container_of(av_fid, struct rxm_av, util_av.av_fid);
 
 	/*
 	 * It's more efficient to remove addresses from high to low index.
@@ -226,6 +234,21 @@ static int rxm_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
 						fi_addr[i]);
 		if (!av_entry)
 			continue;
+
+		if (av->util_av.remove_handler) {
+			peer = ofi_av_addr_context(&av->util_av, fi_addr[i]);
+			(*peer)->refcnt++; /* prevent peer from being freed */
+			ofi_mutex_lock(&av->util_av.ep_list_lock);
+			dlist_foreach(&av->util_av.ep_list, item) {
+				util_ep = container_of(item, struct util_ep,
+						       av_entry);
+				ofi_mutex_unlock(&av->util_av.lock);
+				av->util_av.remove_handler(util_ep, *peer);
+				ofi_mutex_lock(&av->util_av.lock);
+			}
+			ofi_mutex_unlock(&av->util_av.ep_list_lock);
+			_util_put_peer(*peer);
+		}
 
 		if (!ofi_atomic_dec32(&av_entry->use_cnt)) {
 			rxm_put_peer_addr(av, fi_addr[i]);
@@ -362,7 +385,9 @@ static struct fi_ops_av rxm_av_ops = {
 };
 
 int rxm_util_av_open(struct fid_domain *domain_fid, struct fi_av_attr *attr,
-		     struct fid_av **fid_av, void *context, size_t conn_size)
+		     struct fid_av **fid_av, void *context, size_t conn_size,
+		     void (*remove_handler)(struct util_ep *util_ep,
+					    struct util_peer_addr *peer))
 {
 	struct util_domain *domain;
 	struct util_av_attr util_attr;
@@ -398,6 +423,7 @@ int rxm_util_av_open(struct fid_domain *domain_fid, struct fi_av_attr *attr,
 
 	av->util_av.av_fid.fid.ops = &rxm_av_fi_ops;
 	av->util_av.av_fid.ops = &rxm_av_ops;
+	av->util_av.remove_handler = remove_handler;
 	*fid_av = &av->util_av.av_fid;
 	return 0;
 
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -693,6 +693,7 @@ static inline void ofi_cntr_inc(struct util_cntr *cntr)
 
 struct util_av;
 struct util_av_set;
+struct util_peer_addr;
 
 struct util_coll_mc {
 	struct fid_mc		mc_fid;
@@ -749,6 +750,8 @@ struct util_av {
 	size_t			context_offset;
 	struct dlist_entry	ep_list;
 	ofi_mutex_t		ep_list_lock;
+	void (*remove_handler)(struct util_ep *util_ep,
+			       struct util_peer_addr *peer);
 };
 
 #define OFI_AV_DYN_ADDRLEN (1 << 0)
@@ -805,7 +808,9 @@ void util_put_peer(struct util_peer_addr *peer);
 };
 
 int rxm_util_av_open(struct fid_domain *domain_fid, struct fi_av_attr *attr,
-		     struct fid_av **fid_av, void *context, size_t conn_size);
+		     struct fid_av **fid_av, void *context, size_t conn_size,
+		     void (*remove_handler)(struct util_ep *util_ep,
+					    struct util_peer_addr *peer));
 size_t rxm_av_max_peers(struct rxm_av *av);
 void rxm_ref_peer(struct util_peer_addr *peer);
 void *rxm_av_alloc_conn(struct rxm_av *av);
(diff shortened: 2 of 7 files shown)


chien-intel:
AWS CI failure not related to this PR.

20:51:26 2022-09-27 20:51:26,848 - INFO - manual_cluster - Creating instances using ami-044642960112778d5 ....
20:51:37 2022-09-27 20:51:37,210 - INFO - manual_cluster - Error encountered when creating cluster. Message: Failed to run instances due to insufficient capacity.
20:51:37 2022-09-27 20:51:37,210 - WARNING - cluster - Creating cluster failed due to out of resources. Exception: Failed to run instances due to insufficient capacity.
20:51:37 2022-09-27 20:51:37,210 - WARNING - cluster - Will try another availability zone
20:51:37 2022-09-27 20:51:37,283 - INFO - network - Get vpc id vpc-1c582565
20:51:37 2022-09-27 20:51:37,331 - INFO - network - Get public subnet id subnet-52eb6808
20:51:37 2022-09-27 20:51:37,332 - INFO - cluster - Generating cluster multi-node-857-rhel8-efa in az: us-west-2c
20:51:37 2022-09-27 20:51:37,438 - INFO - manual_cluster - Creating instances using ami-044642960112778d5 ....
20:51:38 2022-09-27 20:51:38,853 - INFO - manual_cluster - Creating instances done, instance ids: ['i-00125cb73599886d2', 'i-0275f798f08416cd6']
20:51:38 2022-09-27 20:51:38,853 - INFO - manual_cluster - Waiting for instances to become available ....
21:01:26 2022-09-27 21:01:26,031 - INFO - manual_cluster - Error encountered when creating cluster. Message: Waiter InstanceStatusOk failed: Max attempts exceeded

prov/util/src/rxm_av.c
```diff
@@ -205,36 +204,57 @@ static int rxm_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
...
+			ofi_mutex_unlock(&av->util_av.ep_list_lock);
+			util_put_peer(*peer);
+		}
+
 		if (!ofi_atomic_dec32(&av_entry->use_cnt)) {
+			ofi_mutex_lock(&av->util_av.lock);
 			rxm_put_peer_addr(av, fi_addr[i]);
 			HASH_DELETE(hh, av->util_av.hash, av_entry);
 			ofi_ibuf_free(av_entry);
+			ofi_mutex_unlock(&av->util_av.lock);
```
shefty, line 254: The above flow is inefficient and *looks* racy, even if it's not.  The remove_handler is NULL by default.  The original locking structure ends up looking better, plus we should add a comment regarding why we're dropping the lock in the middle of this handling.  Example:
```
lock();
for (i = count - 1; ...) {
    av_entry = ofi_bufpoll_get_ibuf();
    if (!av_entry)
        continue;

    if (remove_handler) {
        /* The remove_handler may call back into the AV to remove the provider's
         * reference on the peer address.  We need to drop the lock on the AV in case
         * the handler tries to acquire it.  Before dropping the lock, we need to increment
         * the reference count on the peer, so that it's still valid to pass into the handler
         * and isn't freed by another thread after we drop the AV lock.
         */
        peer = ofi_av_addr_context();
        peer->refcnt++;
        unlock();
        remove_handler(...);
        util_put_peer(*peer);
        lock()
    }

    if (1ofi_atomic_dec3(...) {
        ....
    }
}
unlock();
```
I'm missing the ep_list_lock and other details, but hopefully that makes sense.

chien-intel: It comes down to what util_av.lock is protecting.  To put the lock around the for-loop and then unlock/relock in the middle of it is guaranteed racy, so not doing that.
If you want the lock around the for-loop, that leave one alternative, to have a version of rxm_av_free_conn that does not acquire util_av.lock (since it is already the for-loop).

shefty: The for loop is acting over parameters passed in through the app.  There's no lock needed for that.  The AV lock is protecting access to the AV pools and data structures.

chien-intel: added equivalent functions that do not acquire util_av.lock for use in the for-loop and also in the remove_handler.

prov/util/src/rxm_av.c
```diff
@@ -227,14 +242,27 @@ static int rxm_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
...
+
 		if (!ofi_atomic_dec32(&av_entry->use_cnt)) {
 			rxm_put_peer_addr(av, fi_addr[i]);
 			HASH_DELETE(hh, av->util_av.hash, av_entry);
 			ofi_ibuf_free(av_entry);
 		}
 	}
-
 	ofi_mutex_unlock(&av->util_av.lock);
+
```
shefty, line 265: The locking does not work.  This call is forming a lock order of AV lock -> AV ep_list_lock -> EP lock.  The current code has EP lock -> AV lock.  The purpose of the AV ep_list_lock is to support the lock ordering of AV ep_list_lock -> EP lock, because we can't hold the AV lock before acquiring the EP lock.

Data transfers start with an EP.  So, they acquire the EP lock first.  From there, the transfer needs to get information about the peer.  That data come from the AV, which uses the AV lock.  This creates the EP lock -> AV lock ordering.

You need to drop the AV lock prior to invoking the remove_handler().

prov/util/src/rxm_av.c
```diff
@@ -55,10 +54,16 @@ void *rxm_av_alloc_conn(struct rxm_av *av)
 	return conn_ctx;
 }
 
+void _rxm_av_free_conn(void *conn_ctx)
+{
+	/* assert(ofi_mutex_held(&av->util_av.lock)); */
```
shefty, line 59: If a lock must be held by a function, keep the assert active.  It documents the requirement and forces the issue.

prov/util/src/rxm_av.c
```diff
@@ -119,14 +124,20 @@ util_get_peer(struct rxm_av *av, const void *addr)
 	return peer;
 }
 
+void _util_put_peer(struct util_peer_addr *peer)
+{
+	/* assert(ofi_mutex_held(&av->util_av.lock)); */
```
shefty, line 129: See above comment.  But also see below, as the 2 new functions likely aren't needed.

shefty (changes requested):

chien-intel:
Now that you are done with 1.16.x release, we can resume this conversation.
I get your point about lock order.  So, it is best to reconsider the first version of the patch I posted where av lock is acquire on as needed basis and when a connection is freed, the lock order is the same as xmit functions.  Here is the relevant change in rxm_av_remove:
-       ofi_mutex_lock(&av->util_av.lock);
        for (i = count - 1; i >= 0; i--) {
                FI_INFO(av->util_av.prov, FI_LOG_AV,
                        "fi_addr %" PRIu64 "\n", fi_addr[i]);
+               ofi_mutex_lock(&av->util_av.lock);
                av_entry = ofi_bufpool_get_ibuf(av->util_av.av_entry_pool,
                                                fi_addr[i]);
+               ofi_mutex_unlock(&av->util_av.lock);
                if (!av_entry)
                        continue;

+               if (av->util_av.remove_handler) {
+                       ofi_mutex_lock(&av->util_av.lock);
+                       peer = ofi_av_addr_context(&av->util_av, fi_addr[i]);
+                       (*peer)->refcnt++; /* prevent peer from being freed */
+                       ofi_mutex_unlock(&av->util_av.lock);
+                       ofi_mutex_lock(&av->util_av.ep_list_lock);
+                       dlist_foreach(&av->util_av.ep_list, item) {
+                               util_ep = container_of(item, struct util_ep,
+                                                      av_entry);
+                               av->util_av.remove_handler(util_ep, *peer);
+                       }
+                       ofi_mutex_unlock(&av->util_av.ep_list_lock);
+                       util_put_peer(*peer);
+               }
+
                if (!ofi_atomic_dec32(&av_entry->use_cnt)) {
+                       ofi_mutex_lock(&av->util_av.lock);
                        rxm_put_peer_addr(av, fi_addr[i]);
                        HASH_DELETE(hh, av->util_av.hash, av_entry);
                        ofi_ibuf_free(av_entry);
+                       ofi_mutex_unlock(&av->util_av.lock);
                }
        }

-       ofi_mutex_unlock(&av->util_av.lock);

shefty:
See my previous comment that doesn't drop/reacquire the lock as frequently.  The 'remove_handler()' section of that code is where we would iterate over the ep_list.

chien-intel:
In the following code, the lock a is protecting the whole for loop.
lock(a)
for () {
}
unlock(a)

Your suggestions of 
lock(a)
for () {
unlock(a)
lock(a)
}
unlock(a)

The unlock/lock inside the for loop is undoing that protection.  If it is okay to do so, that means the lock around the for loop is protecting too much and semantically looks wrong.  It is bad practice to write code like this.  you may know in this context it is okay to do so but for a casual reader of the code that may not be the case.  As you have mentioned before, the for loop may be just for 1 element.  What's the difference?

shefty:
No one is going to think we're holding a lock around a stack variable doing ```i--``` because it's needed.  We want to be somewhat efficient and avoid grabbing/releasing the same lock back to back to back.

We need to hold the AV lock around the call to ofi_bufpool_get_ibuf() and ofi_av_addr_context().  Let's start with a loop that's only calling ofi_bufpool_get_ibuf()
```
for (...) {
    av_entry = ofi_bufpool_get_ibuf();
    if (!av_entry)
        continue;
}
```
Locking outside the loop; even though, the for statement itself doesn't need to be locked looks like:
```
lock();
for (...) {
    av_entry = ofi_bufpool_get_ibuf();
    if (!av_entry)
        continue;
}
unlock();
```
Any reasonable developer would accept this.  Using 'fine-grain' locking gives us:
```
for (...) {
    lock();
    av_entry = ofi_bufpool_get_ibuf();
    if (!av_entry) {
        unlock();
        continue;
    }
    unlock();
}
```
The code is less readable, locking is less efficient, and the benefit is... we don't hold a lock around the for()?  We haven't improved code maintenance.  Another option is:
```
for (...) {
    lock();
    av_entry = ofi_bufpool_get_ibuf();
    unlock();
    if (!av_entry)
        continue;
}
```
which looks better, until we expand what we're doing inside the loop.
```
for (...) {
    lock();
    av_entry = ofi_bufpool_get_ibuf();
    unlock();
    if (!av_entry)
        continue;

    if (!ofi_atomic_dec32(av_entry...)) {
        lock();
        use and free av_entry
        unlock();
    }
}
```
versus the first option:
```
lock();
for (...) {
    av_entry = ofi_bufpool_get_ibuf();
    if (!av_entry)
        continue;

    if (!ofi_atomic_dec32(av_entry...)) {
        use and free av_entry
    }
}
unlock();
```
So, yes, I do think avoiding bouncing the lock multiple times per iteration is better.  And this section of code where we bump a reference count is directly related to why we can drop the lock:
```
        /* The remove_handler may call back into the AV to remove the provider's
         * reference on the peer address.  We need to drop the lock on the AV in case
         * the handler tries to acquire it.  Before dropping the lock, we need to increment
         * the reference count on the peer, so that it's still valid to pass into the handler
         * and isn't freed by another thread after we drop the AV lock.
         */
        peer = ofi_av_addr_context();
        peer->refcnt++;
        unlock();
```
We only take this path if we have a remove_handler.  Otherwise, we continue to hold the lock we took at the start until the end of needing it.

chien-intel:
Locking scheme changed as requested.  CI errors are the same as with other PRs (net provider tests).

shefty:
Pulled changes locally and updated locking and moved env var to ofi core.  New PR #8190

PR #8039, opened 2022-09-28
prov/verbs: Add option to disable dmabuf support

j-xiong:
Currently the verbs provider uses dmabuf based memory registration whenever possible and only falls back to regular memory registration when the first approach fails. This works well most of the time. However, sometimes it may be desirable to disable the dmabuf based registration. For example, to test the functionality of the dmabuf peer-mem adapter driver.

This patch adds a runtime option to the verbs provider that allows explicit enabling/disabling of dmabuf support.

### DIFF
--- a/prov/verbs/src/fi_verbs.c
+++ b/prov/verbs/src/fi_verbs.c
@@ -705,6 +705,17 @@ static int vrb_read_params(void)
 		return -FI_EINVAL;
 	}
 
+	if (vrb_gl_data.dmabuf_support) {
+		if (vrb_get_param_bool("use_dmabuf", "Enable dmabuf based memory "
+				       "registrations, if supported. Yes by default.",
+				       (int *)&vrb_gl_data.dmabuf_support)) {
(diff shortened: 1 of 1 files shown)


prov/verbs/src/fi_verbs.c
```diff
@@ -705,6 +705,15 @@ static int vrb_read_params(void)
 		return -FI_EINVAL;
 	}
 
+	if (vrb_get_param_bool("use_dmabuf", "Enable dmabuf based memory "
+			       "registrations, if supported. Yes by default.",
+			       (int *)&vrb_gl_data.dmabuf_support)) {
+		VRB_WARN(FI_LOG_CORE, "Invalid value of use_dmabuf\n");
+		return -FI_EINVAL;
+	}
```
shefty, line 713: This could end up trying to re-enable dmabuf support after we checked that support isn't available.  Maybe logic something like this?
```
if (dmabuf_support) {
    if (get_param_bool("disable_dmabuf", ...)
        ...
}
```
I don't really care if it's use vs disable.

j-xiong: fixed.

PR #8045, opened 2022-09-30
util/pinpong: add AI_NUMERICSERV definition when it is not defined

barracuda156:
Older Darwin does not have definition for `AI_NUMERICSERV`, which is used in `util/pingpong.c` (and causes build to fail on 10.6). This PR fixes this error.

It mirrors the fix for `MoarVM` from: https://github.com/macports/macports-ports/pull/14932


shefty (approved):

barracuda156:
@shefty Thank you!

PR #8046, opened 2022-09-30
fabtests/pytest: retry when ssh connection issues are encountered

wzamazon:
When a test encountered ssh connection issue like "reset by peer", "closed by peer", or "refused", this patch will retry the test for 3 times, with each retry interval 5 seconds.

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -1,14 +1,49 @@
 import pytest
 import errno
+import os
+from tempfile import NamedTemporaryFile
+from subprocess import run
+from retrying import retry
 
+
+class SshConnectionError(Exception):
+
+    def __init__(self):
+        super().__init__(self, "Ssh connection failed")
+
+
+def is_ssh_connection_error(exception):
+    return isinstance(exception, SshConnectionError)
+
+
+def has_ssh_connection_err_msg(output):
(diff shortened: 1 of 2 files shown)


fabtests/pytest/common.py
```diff
@@ -1,14 +1,49 @@
...
+def has_ssh_connection_err_msg(output):
+    err_msgs = ["kex_exchange_identification: Connection closed by remote host",
+                "ssh_exchange_identification: read: Connection reset by peer",
+                "port 22: Connection refused"]
+
+    for msg in err_msgs:
+        if output.find(msg) != -1:
+            return True
+
+    return False
```
darrylabbate, line 28: nit: I think this can be simplified to

```python3
return output in err_msgs
```

darrylabbate (approved):

fabtests/pytest/common.py
wzamazon: That will not work because output is a large string, and the ssh error message is only part of it.

darrylabbate: I see. I was assuming `output` was the singular line/message.

shijin-aws (approved):

wzamazon:
Intel CI is not using pytest script, so the error should not be related to this PR.

PR #8051, opened 2022-10-01
fabtests/pytest: add possible ssh connection error message

wzamazon:
Add "ssh_exchange_identification: Connection closed by remote host" as an error message that is considered ssh connection issue.


darrylabbate (approved):

wzamazon:
intel CI does not use pytest, so ignore the error

PR #8055, opened 2022-10-03
gh-pages: Add yet another missing link of fi_peer man page

j-xiong:
A previous patch (commit efa7f61a51b235a5e08c32ec8d9a209e91be65c3) fixed the man pages for v1.16.0, but the link is still missing from main.


shefty:
thanks

PR #8056, opened 2022-10-03
contrib/intel: replace '-' delimiter with '_'

aingerson:
The Jenkins nodes use the naming convention "node-fabric"
Jenkins parses by the '-' and assumes the first element to be the
node name to ssh into. This delimiter conflicts with normal node
naming convention which uses dashes within the node name.
This changes the node/fabric delimiter from - to _ so it doesn't
conflict with regular node names.

Note that this will require a momentary down period between naming
the nodes and merging the PR.
After this change is merged, PRs will have to rebase on upstream
in order to pass.

Verified this works with a single pair of renamed nodes as well as with the daos testing.

Also includes 2 unrelated minor fixes.

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -450,7 +450,7 @@ pipeline {
                 env
                 (
                   echo `hostname`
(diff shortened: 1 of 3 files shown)


aingerson:
FYI All Intel CI runs will fail until this is merged

PR #8060, opened 2022-10-04
fabtests/multinode: enable multinode collective tests on verbs provider

### LINKED ISSUE #8059: Multinode tests can only be run with tcp provider
It is impossible to set ``verbs`` provider with ``-p`` argument as it is always overwritten by the ``setup_hints()`` function:
https://github.com/ofiwg/libfabric/blob/9d5395650629e9e192abb1b889b5231de71b2790/fabtests/multinode/src/core_coll.c#L528

grom72:
Fixes #8059

### DIFF
--- a/fabtests/multinode/src/core_coll.c
+++ b/fabtests/multinode/src/core_coll.c
@@ -525,7 +525,6 @@ static inline int setup_hints()
 	hints->mode = FI_CONTEXT;
 	hints->domain_attr->control_progress = FI_PROGRESS_MANUAL;
 	hints->domain_attr->data_progress = FI_PROGRESS_MANUAL;
-	hints->fabric_attr->prov_name = strdup("tcp");
 	return FI_SUCCESS;
 }
 


fabtests/multinode/src/core_coll.c
```diff
@@ -525,7 +525,8 @@ static inline int setup_hints()
 	hints->mode = FI_CONTEXT;
 	hints->domain_attr->control_progress = FI_PROGRESS_MANUAL;
 	hints->domain_attr->data_progress = FI_PROGRESS_MANUAL;
-	hints->fabric_attr->prov_name = strdup("tcp");
+	if (!hints->fabric_attr->prov_name)
```
ooststep, line 528: I wonder if we can just remove this completely and either let the `-p` option set it or let getinfo decide

shefty: We should remove this and align the behavior with all other tests.

shefty:
Please squash the 2 commits into 1 and ensure the signed-off-by line is set correctly, so the DCO check passes.

shefty:
kicked off CI

shefty:
This is a trivial change that got ignored for a bit.  Re-started CI.

shefty (approved):

shefty:
Pulled changes into PR #8217, so I could rebase to force new CI runs.

PR #8061, opened 2022-10-04
fabtests: update rdm_tagged_peek to remove reliance on unexpected messages

ooststep:
Since the net provider does not implement an unexpected queue the existing implementation of rdm_tagged_peek hangs as it attempts to read the last message first, forcing the first messages to an assumed unexpected queue in the provider.

This set of changes removes the assumption on the unexpected queue by reading messages in the correct order and ignoring FI_ENOMSG errors when we're searching for a message with FI_PEEK.

### DIFF
--- a/fabtests/functional/rdm_tagged_peek.c
+++ b/fabtests/functional/rdm_tagged_peek.c
@@ -57,7 +57,8 @@ static int wait_for_send_comp(int count)
 	return 0;
 }
 
-static int tag_queue_op(uint64_t tag, int recv, uint64_t flags)
+static int
+tag_queue_op(uint64_t tag, int recv, uint64_t flags, bool ignore_nomsg)
 {
 	int ret;
 	struct fi_cq_tagged_entry comp;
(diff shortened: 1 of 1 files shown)


j-xiong:
Does it make sense to test FI_PEEK on providers that don't support unexpected message? Or in other words, should a provider that doesn't support unexpected message claim to support FI_PEEK?

shefty:
MPI requires the use of FI_PEEK, so it's really needed by any provider wanting to support MPI.  The issue is how deep of an unexpected queue is a provider expected to support?  The current test is really implemented around the behavior and implementation of the sockets provider.

The current test does not work with the net provider, which is what we want to fix.

j-xiong:
So the net provider does support unexpected message, but the depth is only 1?  Is it because the line ` xnet_update_pollflag(ep, POLLIN, false);` after the entry is inserted to the unexpected list?

j-xiong:
Ok, I see, the unexpected list is a list of eps that have unexpected messages. Each ep can have only one unexpected message.

shefty:
@j-xiong - correct.  The net provider leaves the data on the socket.  Pulling the data off the socket ends up causing more problems.  First, there's an increase in receive side memory use to store the unexpected messages.  We've seen this grow to MBs per process.  Second, the queue it forms requires that all posted receives to walk it looking for a match.  That can add significant latency to each message.  Third, it basically forces the use of a rendezvous protocol, which slows down message bandwidth for large transfers.

So it basically reduces the unexpected list to 1, since we have to read the header to determine that there's no match.  By keeping that data on the socket, this allows tcp flow control to push back into the network.

ooststep:
rebased to pull in CI and multinode updates

PR #8071, opened 2022-10-05
contrib/intel/jenkins: Catch potential release package make issues

zachdworkin:
Add make distcheck and rpmbuild to jenkins for regular build.
We need to catch problems where a release package does not include all necessary files because of potential missing updates to makefile.includes.

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -3,7 +3,7 @@ properties([disableConcurrentBuilds(abortPrevious: true)])
 def DO_RUN=1
 def TARGET="main"
 def check_target() {
-  echo "CHANGE_TARGET = ${env.CHANGE_TARET}"
+  echo "CHANGE_TARGET = ${env.CHANGE_TARGET}"
   if (changeRequest()) {
     TARGET = env.CHANGE_TARGET
   }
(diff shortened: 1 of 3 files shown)


shefty:
Does this need to be picked for v1.16.x, or will the CI use this file once merged?

zachdworkin:
@shefty this will need to be picked. The Jenkinsfile is not like the other files in contrib/intel/jenkins that the CI would use from upstream/main.

zachdworkin:
@shefty The job 2 failure shows that this fix works for the checking of that missing verbs file in Makefile.include in prov/verbs. Job 1 failures are unrelated but you can see in the build stage that it ran what we expected it to do.

shefty:
Thanks - so this PR will fail unless #8070 is added to it?

zachdworkin:
@shefty yes.

zachdworkin:
@shefty the https://github.com/ofiwg/libfabric/pull/8070 changes also need to be merged into upstream/main for this to pass. Should I add a commit that includes the verbs file addition?

shefty:
If adding the patch from #8070 here will allow the CI to pass, then please go ahead and add that change.  I can close #8070.

zachdworkin:
@shefty added. The CI is running a replay with my changes applied.

PR #8035, opened 2022-09-27
prov/efa: Directly use ibv_cq_ex in rxr_ep 

sunkuamzn:
rxr_ep now has a new member ibv_cq_ex which is polled directly in
rxr_ep_progress_internal

Signed-off-by: Sai Sunku <sunkusa@amazon.com>

PR #8074, opened 2022-10-05
prov/efa: Open extended CQ in rxr_ep

sunkuamzn:
In a previous commit, a new member ibv_cq_ex was added to rxr_ep (https://github.com/ofiwg/libfabric/pull/8035)
But only ibv_create_cq was used to open the CQ

This commit adds support for extended CQ which is created by calling efadv_create_cq

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -36,6 +36,7 @@
 #include "config.h"
 #include <ofi_mem.h>
 #include "efa.h"
+#include "efa_cq.h"
 #include <infiniband/verbs.h>
 
 static inline uint64_t efa_cq_opcode_to_fi_flags(enum ibv_wc_opcode	opcode) {
@@ -251,102 +252,24 @@ static struct fi_ops efa_cq_fi_ops = {
 };
 
 /**
- * @brief Create ibv_cq_ex by calling ibv_create_cq_ex
- *
- * @param[in] cq Pointer to the efa_cq.
- * @param[in] attr Pointer to fi_cq_attr.
- * @param[out] Return pointer to ibv_cq_ex if successful, otherwise NULL
- */
-static inline struct ibv_cq_ex *efa_cq_ibv_create_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
-	struct ibv_cq_init_attr_ex init_attr_ex = {
-		.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE,
-		.cq_context = NULL,
-		.channel = NULL,
-		.comp_vector = 0,
-		/* EFA requires these values for wc_flags and comp_mask.
-		 * See `efa_create_cq_ex` in rdma-core.
-		 */
-		.wc_flags = IBV_WC_STANDARD_FLAGS,
-		.comp_mask = 0,
-	};
-
-	return ibv_create_cq_ex(cq->domain->device->ibv_ctx, &init_attr_ex);
-}
-
-/**
  * @brief Create and set cq->ibv_cq_ex
  *
  * @param[in] cq Pointer to the efa_cq. cq->ibv_cq_ex must be NULL.
  * @param[in] attr Pointer to fi_cq_attr.
  * @param[out] Return code = 0 if successful, or negative otherwise.
  */
-#if HAVE_EFADV_CQ_EX
-static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
-	struct ibv_cq_init_attr_ex init_attr_ex = {
-		.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE,
-		.cq_context = NULL,
-		.channel = NULL,
-		.comp_vector = 0,
-		/* EFA requires these values for wc_flags and comp_mask.
-		 * See `efa_create_cq_ex` in rdma-core.
-		 */
-		.wc_flags = IBV_WC_STANDARD_FLAGS,
-		.comp_mask = 0,
-	};
-
-	struct efadv_cq_init_attr efadv_cq_init_attr = {
-		.comp_mask = 0,
-		.wc_flags = EFADV_WC_EX_WITH_SGID,
-	};
+static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr)
+{
+	enum ibv_cq_ex_type ibv_cq_ex_type;
 
 	if (cq->ibv_cq_ex) {
 		EFA_WARN(FI_LOG_CQ, "CQ already has attached ibv_cq_ex\n");
 		return -FI_EALREADY;
 	}
 
-	/* To enable additional WC fields, create CQ with efadv_create_cq verb instead. */
-	cq->ibv_cq_ex = efadv_create_cq(cq->domain->device->ibv_ctx,
-						   &init_attr_ex,
-						   &efadv_cq_init_attr,
-						   sizeof(efadv_cq_init_attr));
-
-	if (cq->ibv_cq_ex) {
-		/* Support operations on EFA DV CQ */
-		cq->flags |= EFA_CQ_SUPPORT_EFADV_CQ;
-		return 0;
-	}
-
-	/* This could be due to old EFA kernel module versions */
-	EFA_WARN(FI_LOG_CQ, "Unable to create EFA DV CQ: %s\n", strerror(errno));
-	cq->flags &= ~EFA_CQ_SUPPORT_EFADV_CQ;
-
-	/* Fallback to ibv_create_cq_ex */
-	cq->ibv_cq_ex = efa_cq_ibv_create_cq_ex(cq, attr);
-
-	if (!cq->ibv_cq_ex) {
-		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
-		return -FI_ENOCQ;
-	}
-
-	return 0;
+	return efa_cq_ibv_cq_ex_open(attr, cq->domain->device->ibv_ctx,
+				    &cq->ibv_cq_ex, &ibv_cq_ex_type);
 }
-#else
-static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *attr) {
-	if (cq->ibv_cq_ex) {
-		EFA_WARN(FI_LOG_CQ, "CQ already has attached ibv_cq_ex\n");
-		return -FI_EALREADY;
-	}
-
-	cq->ibv_cq_ex = efa_cq_ibv_create_cq_ex(cq, attr);
-
-	if (!cq->ibv_cq_ex) {
-		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
-		return -FI_ENOCQ;
-	}
-
-	return 0;
-}
-#endif
 
 int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 		struct fid_cq **cq_fid, void *context)
--- a/prov/efa/src/efa_cq.h
+++ b/prov/efa/src/efa_cq.h
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2022 Amazon.com, Inc. or its affiliates. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "efa.h"
+
+/**
+ * @brief Create ibv_cq_ex by calling ibv_create_cq_ex
+ *
+ * @param[in] ibv_cq_init_attr_ex Pointer to ibv_cq_init_attr_ex
+ * @param[in] ibv_ctx Pointer to ibv_context
+ * @param[in,out] ibv_cq_ex Pointer to newly created ibv_cq_ex
+ * @param[in,out] ibv_cq_ex_type enum indicating if efadv_create_cq or ibv_create_cq_ex was used
+ * @return Return 0 on success, error code otherwise
+ */
+static inline int efa_cq_ibv_cq_ex_open_with_ibv_create_cq_ex(
+	struct ibv_cq_init_attr_ex *ibv_cq_init_attr_ex,
+	struct ibv_context *ibv_ctx, struct ibv_cq_ex **ibv_cq_ex,
+	enum ibv_cq_ex_type *ibv_cq_ex_type)
+{
+	*ibv_cq_ex = ibv_create_cq_ex(ibv_ctx, ibv_cq_init_attr_ex);
+
+	if (!*ibv_cq_ex) {
+		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
+		return -FI_ENOCQ;
+	}
+
+	*ibv_cq_ex_type = IBV_CQ;
+	return 0;
+}
+
+/**
+ * @brief Create ibv_cq_ex by calling efadv_create_cq or ibv_create_cq_ex
+ *
+ * @param[in] ibv_cq_init_attr_ex Pointer to ibv_cq_init_attr_ex
+ * @param[in] efadv_cq_init_attr_ex Pointer to efadv_cq_init_attr_ex
+ * @param[in] ibv_ctx Pointer to ibv_context
+ * @param[in,out] ibv_cq_ex Pointer to newly created ibv_cq_ex
+ * @param[in,out] ibv_cq_ex_type enum indicating if efadv_create_cq or ibv_create_cq_ex was used
+ * @return Return 0 on success, error code otherwise
+ */
+#if HAVE_EFADV_CQ_EX
+static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
+					struct ibv_context *ibv_ctx,
+					struct ibv_cq_ex **ibv_cq_ex,
+					enum ibv_cq_ex_type *ibv_cq_ex_type)
+{
+	struct ibv_cq_init_attr_ex init_attr_ex = {
+		.cqe = attr->size ? attr->size : EFA_DEF_CQ_SIZE,
+		.cq_context = NULL,
+		.channel = NULL,
+		.comp_vector = 0,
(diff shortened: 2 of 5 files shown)


sunkuamzn:
Builds failing because of missing `EFADV_WC_EX_WITH_SGID` which was introduced in this Aug 23 [commit](https://github.com/linux-rdma/rdma-core/commit/0784b27b947ef9dc0f0c9e55ab5de0ebd648041b). Build check using an outdated rdma-core commit?

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,91 @@
...
+
+	struct ibv_cq_ex *ibv_cq = ibv_create_cq_ex(ibv_ctx, ibv_cq_init_attr_ex);
+
+    if (ibv_cq) {
+        **ibv_cq_ex = ibv_cq;
+        return 0;
+    } else {
+        EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
+        return -FI_ENOCQ;
+    }
```
darrylabbate, line 55: Indentation looks off (spaces instead of tabs)

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,91 @@
...
+        *ibv_cq_ex = ibv_cq;
+		return 0;
+	}
+
+	/* This could be due to old EFA kernel module versions */
+    /* Fallback to ibv_create_cq_ex */
+    return ibv_cq_ex_open_with_ibv_create_cq_ex(ibv_cq_init_attr_ex, ibv_ctx, &ibv_cq_ex);
+    #endif
+
+    return ibv_cq_ex_open_with_ibv_create_cq_ex(ibv_cq_init_attr_ex, ibv_ctx, &ibv_cq_ex);
```
darrylabbate, line 90: Same here; spaces instead of tabs

prov/efa/src/efa_cq.c
```diff
@@ -304,49 +285,26 @@ static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *att
...
-						   &efadv_cq_init_attr,
-						   sizeof(efadv_cq_init_attr));
+	ret = ibv_cq_ex_open(&init_attr_ex, &efadv_cq_init_attr, cq->domain->device->ibv_ctx,
+						&ibv_cq_ex, &is_efadv_cq);
 
-	if (cq->ibv_cq_ex) {
+	if (ret)
+		return ret;
+
+	#if HAVE_EFADV_CQ_EX
```
darrylabbate, line 294: Probably want to align the preprocessor directive to the beginning of the line

prov/efa/src/efa_cq.c
```diff
@@ -304,49 +285,26 @@ static inline int efa_cq_set_ibv_cq_ex(struct efa_cq *cq, struct fi_cq_attr *att
...
-
-	if (!cq->ibv_cq_ex) {
-		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
-		return -FI_ENOCQ;
+	} else {
+		/* This could be due to old EFA kernel module versions */
+		EFA_WARN(FI_LOG_CQ, "Unable to create EFA DV CQ: %s\n", strerror(errno));
+		cq->flags &= ~EFA_CQ_SUPPORT_EFADV_CQ;
 	}
+	#endif
```
darrylabbate, line 303: Preprocessor directive alignment again

darrylabbate (changes requested):

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,91 @@
...
+ */
+static inline int ibv_cq_ex_open(struct ibv_cq_init_attr_ex *ibv_cq_init_attr_ex,
+                                struct efadv_cq_init_attr *efadv_cq_init_attr,
+                                struct ibv_context *ibv_ctx, struct ibv_cq_ex **ibv_cq_ex,
+                                bool *is_efadv_cq)
+{
+    struct ibv_cq_ex *ibv_cq;
+    *is_efadv_cq = false;
+
+    #if HAVE_EFADV_CQ_EX
```
shijin-aws, line 76: The libfabric convention is to define separate function for each preprocessor block, i.e. you define different `ibv_cq_ex_open` for `HAVE_EFADV_CQ_EX` = 0 and 1

shijin-aws (changes requested):

prov/efa/src/efa_cq.c
```diff
@@ -250,38 +251,18 @@ static struct fi_ops efa_cq_fi_ops = {
...
-}
-
 /**
  * @brief Create and set cq->ibv_cq_ex
  *
  * @param[in] cq Pointer to the efa_cq. cq->ibv_cq_ex must be NULL.
  * @param[in] attr Pointer to fi_cq_attr.
  * @param[out] Return code = 0 if successful, or negative otherwise.
  */
-#if HAVE_EFADV_CQ_EX
```
shijin-aws, line 283: Why do you move this macro check to L294? Some variables from here to L294 are only available when HAVE_EFADV_CQ_EX is defined, which explains the build error you have.

shijin-aws:
> Builds failing because of missing `EFADV_WC_EX_WITH_SGID` which was introduced in this Aug 23 [commit](https://github.com/linux-rdma/rdma-core/commit/0784b27b947ef9dc0f0c9e55ab5de0ebd648041b). Build check using an outdated rdma-core commit?

That is because you move the `if HAVE_EFADV_CQ_EX` check outside the code snippet that have the symbols that are only available when HAVE_EFADV_CQ_EX=1

prov/efa/src/efa_cq.c
sunkuamzn: I missed this. Fixed.

prov/efa/src/efa_cq.h
sunkuamzn: ACK. Fixed.

prov/efa/src/efa_cq.c
sunkuamzn: fixed

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,99 @@
...
+#if HAVE_EFADV_CQ_EX
+static inline int ibv_cq_ex_open(struct ibv_cq_init_attr_ex *ibv_cq_init_attr_ex,
+                                struct efadv_cq_init_attr *efadv_cq_init_attr,
+                                struct ibv_context *ibv_ctx, struct ibv_cq_ex **ibv_cq_ex,
+                                bool *is_efadv_cq)
+{
+    struct ibv_cq_ex *ibv_cq;
+    *is_efadv_cq = false;
+
+	ibv_cq = efadv_create_cq(ibv_ctx, ibv_cq_init_attr_ex, efadv_cq_init_attr, sizeof(*efadv_cq_init_attr));
```
shijin-aws, line 77: indent off.

sunkuamzn: yes, I'm working on fixing the indents

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -2502,14 +2501,19 @@ int rxr_ep_alloc_app_device_info(struct fi_info **app_device_info,
...
- * @param[in] cq Pointer to the efa_cq.
- * @param[in] attr Pointer to fi_cq_attr.
- * @return Return pointer to ibv_cq_ex if successful, otherwise NULL
+ * @param[in,out] ep Pointer to rxr_ep
+ * @param[in] attr Pointer to fi_cq_attr
+ * @return Return 0 on success, error code otherwise
  */
-static inline struct ibv_cq_ex *rxr_ep_create_ibv_cq_ex(struct rxr_ep *ep, struct fi_cq_attr *attr)
+#if HAVE_EFADV_CQ_EX
+static inline int rxr_ep_create_ibv_cq_ex(struct rxr_ep *ep, struct fi_cq_attr *attr)
```
shijin-aws, line 2511: These two `rxr_ep_create_ibv_cq_ex` has considerable code duplication. Consider how to reduce it.

shijin-aws (changes requested):

shijin-aws:
Left some comments. It seems to me the code can be simplified further.

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @param[in] ibv_cq_init_attr_ex Pointer to ibv_cq_init_attr_ex
+ * @param[in] efadv_cq_init_attr_ex Pointer to efadv_cq_init_attr_ex
+ * @param[in] ibv_ctx Pointer to ibv_context
+ * @param[in,out] ibv_cq_ex Pointer to newly created ibv_cq_ex
+ * @param[in,out] is_efadv_cq set to true if efadv_create_cq is used and false otherwise
+ * @return Return 0 on success, error code otherwise
+ */
+#if HAVE_EFADV_CQ_EX
+static inline int ibv_cq_ex_open(struct ibv_cq_init_attr_ex *ibv_cq_init_attr_ex,
```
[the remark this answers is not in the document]
sunkuamzn: This function is called from both `efa_cq_open` and `rxr_endpoint`, so I'm not sure if `efa_cq_create_ibv_cq_ex` makes sense..

prov/efa/src/rxr/rxr_ep.c
sunkuamzn: That's why I originally had `#if HAVE_EFADV_CQ_EX` inside the function but you said the libfabric convention is to have separate functions. The other way to reduce duplication is to move some code to `rxr_endpoint` but that would make the already long `rxr_endpoint` longer. Any ideas on how exactly to reduce the duplication?

shijin-aws: You need to find the minimum section of code that has branching for both pre-processors, make them as a function, then use the same function to call it.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -2524,8 +2528,58 @@ static inline struct ibv_cq_ex *rxr_ep_create_ibv_cq_ex(struct rxr_ep *ep, struc
...
+	return 0;
 }
+#else
+static inline int rxr_ep_create_ibv_cq_ex(struct rxr_ep *ep, struct fi_cq_attr *attr)
+{
+	int ret;
+	bool is_efadv_cq = false;
+	struct ibv_cq_ex *ibv_cq_ex;
+
+	struct ibv_cq_init_attr_ex init_attr_ex = {
```
shijin-aws, line 2558: Is this `init_attr_ex` in all cases? Can we merge it into `ibv_cq_ex_open` ?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1840,14 +1841,12 @@ static inline fi_addr_t rdm_ep_determine_peer_address_from_efadv(struct rxr_ep *
...
 	struct efa_ep *efa_ep;
-	struct efa_cq *efa_cq;
 	struct efa_ep_addr efa_ep_addr = {0};
 	fi_addr_t addr;
 	union ibv_gid gid = {0};
 	uint32_t *connid = NULL;
 
-	efa_cq = container_of(ep->rdm_cq, struct efa_cq, util_cq.cq_fid);
-	if (!(efa_cq->flags & EFA_CQ_SUPPORT_EFADV_CQ)) {
+	if (ep->ibv_cq_ex_type != EFADV_CQ) {
```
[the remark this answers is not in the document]
sunkuamzn: It looks the like the flags were set manually previously, so I don't know if there's a way to get the flags from the already created `ibv_cq_ex`

Wei suggested the enum approach and I followed it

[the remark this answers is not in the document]
wzamazon: @wenduwan  I wonder if the flag EFA_CQ_SUPPORT_EFADV_CQ is really necessary.

The only purpose of it is to prevent later code to call `efadv_wc_read_sgid` on a `ibv_cq_ex` created by using `ibv_create_cq`.

I wonder why do we need to prevent this from happening? 

I understand such a call would fail because `ibv_cq_ex` created by `ibv_create_cq` does not support `efadv_wc_read_sgid`, but we should be able to a failure.

It would be nice to be able to get rid of the flag.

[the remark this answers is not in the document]
wenduwan: > The only purpose of it is to prevent later code to call efadv_wc_read_sgid on a ibv_cq_ex created by using ibv_create_cq.
I wonder why do we need to prevent this from happening?

`EFA_CQ_SUPPORT_EFADV_CQ` is a runtime flag to indicate kernel module support for wide completion APIs. We introduced it because the installer kernel module could lag behind rdma-core APIs, such that we need to make sure both support wide completion:
1. For rdma-core APIs we just a macro, i.e. `HAVE_EFADV_CQ_EX`
1. For kernel model we cannot use a macro, but instead, we try to create edadv cq and if that fails we fallback to the normal `ibv_cq_ex`, i.e. a runtime flag `EFA_CQ_SUPPORT_EFADV_CQ`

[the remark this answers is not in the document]
sunkuamzn: Tests showed that calling `efadv_wc_read_sgid` on a `ibv_cq_ex` created by using `ibv_create_cq` leads to a segfault, so we need to keep the enum.

prov/efa/src/efa_cq.h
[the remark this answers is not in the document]
sunkuamzn: fixed!

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * @brief Create ibv_cq_ex by calling ibv_create_cq_ex
+ *
+ * @param[in] ibv_cq_init_attr_ex Pointer to ibv_cq_init_attr_ex
+ * @param[in] ibv_ctx Pointer to ibv_context
+ * @param[in,out] ibv_cq_ex Pointer to newly created ibv_cq_ex
+ * @return Return 0 on success, error code otherwise
+ */
+static inline int ibv_cq_ex_open_with_ibv_create_cq_ex(struct ibv_cq_init_attr_ex *ibv_cq_init_attr_ex,
```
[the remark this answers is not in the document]
sunkuamzn: fixed!

sunkuamzn:
Talked with @shijin-aws offline. Moved the `struct ibv_cq_init_attr_ex init_attr_ex` and `truct efadv_cq_init_attr efadv_cq_init_attr` to `efa_cq.h` and got rid of the `#if HAVE_EFADV_CQ_EX` preprocessor directives in `rxr_ep.c` and `efa_cq.c`

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * @brief Create ibv_cq_ex by calling ibv_create_cq_ex
+ *
+ * @param[in] ibv_cq_init_attr_ex Pointer to ibv_cq_init_attr_ex
+ * @param[in] ibv_ctx Pointer to ibv_context
+ * @param[in,out] ibv_cq_ex Pointer to newly created ibv_cq_ex
+ * @return Return 0 on success, error code otherwise
+ */
+static inline int efa_cq_ibv_cq_ex_open_with_ibv_create_cq_ex(
```
shijin-aws, line 43: nit: maybe use the earlier name `efa_cq_ibv_create_cq_ex`, which you removed in `efa_cq.c`.

shijin-aws:
Looks good. Some minor comments.

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,124 @@
...
+	if (*ibv_cq_ex) {
+		if (*ibv_cq_ex_type)
+			*ibv_cq_ex_type = EFADV_CQ;
+
+		return 0;
+	}
+
+	/* This could be due to old EFA kernel module versions */
+	/* Fallback to ibv_create_cq_ex */
+	return efa_cq_ibv_cq_ex_open_with_ibv_create_cq_ex(&init_attr_ex, ibv_ctx, &ibv_cq_ex);
```
wzamazon, line 102: In this case, you need to set `ibv_cq_ex_type` to `IBV_CQ`

wzamazon (changes requested):

prov/efa/src/efa_cq.h
shijin-aws: The type is `IBV_CQ` by default. So it should be OK?

wzamazon: > But it could be better to move that assignment to here.

It would be better to set type her instead of before calling the function.

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,124 @@
...
+}
+#else
+static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
```
wzamazon, line 105: also need doxygen function header for this function

prov/efa/src/efa_cq.h
sunkuamzn: fixed

prov/efa/src/efa_cq.h
shijin-aws: I do not have strong preference. Just document what the function does for the HAVE_EFADV_CQ_EX=0 case. Some duplication in narratives is fine.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -2502,29 +2501,26 @@ int rxr_ep_alloc_app_device_info(struct fi_info **app_device_info,
...
-
 	struct efa_domain *efa_domain = rxr_ep_domain(ep);
 
-	return ibv_create_cq_ex(efa_domain->device->ibv_ctx, &init_attr_ex);
+	/*
+	 * Set ep->ibv_cq_ex_type to IBV_CQ by default
+	 * If efa_cq_ibv_cq_ex_open succeeds in using efadv_create_cq,
+	 * then it will update ep->ibv_cq_ex_type to EFADV_CQ
+	 */
+	ep->ibv_cq_ex_type = IBV_CQ;
```
shijin-aws, line 2520: This assignment is already done  in `efa_cq_ibv_cq_ex_open` right?  Can you remove it here?

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,132 @@
...
+		.comp_mask = 0,
+		.wc_flags = EFADV_WC_EX_WITH_SGID,
+	};
+
+	*ibv_cq_ex = efadv_create_cq(ibv_ctx, &init_attr_ex,
+				     &efadv_cq_init_attr,
+				     sizeof(efadv_cq_init_attr));
+
+	if (*ibv_cq_ex) {
+		if (ibv_cq_ex_type)
```
shijin-aws, line 100: Is this `if` really needed?

prov/efa/src/efa_cq.h
```diff
@@ -0,0 +1,132 @@
...
+ * @return Return 0 on success, error code otherwise
+ */
+static inline int efa_cq_ibv_cq_ex_open_with_ibv_create_cq_ex(
+	struct ibv_cq_init_attr_ex *ibv_cq_init_attr_ex,
+	struct ibv_context *ibv_ctx, struct ibv_cq_ex ***ibv_cq_ex,
+	enum ibv_cq_ex_type **ibv_cq_ex_type)
+{
+	**ibv_cq_ex = ibv_create_cq_ex(ibv_ctx, ibv_cq_init_attr_ex);
+
+	if (**ibv_cq_ex) {
```
shijin-aws, line 51: I would change it to 

```
if (!**ibv_cq_ex) {
    	EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
	return -FI_ENOCQ;
}

**ibv_cq_ex_type = IBV_CQ;
return 0;
```

shijin-aws (changes requested):

prov/efa/src/efa_cq.h
shijin-aws: Also, why double ptr * here?

prov/efa/src/rxr/rxr_ep.c
sunkuamzn: If I remove this, `ep->ibv_cq_ex_type` will be uninitialized. I need a null check later on, so leaving `ep->ibv_cq_ex_type` uninitialized could cause that null check to fail

prov/efa/src/efa_cq.h
sunkuamzn: Yes, because the dgram path doesn't have this enum, so I am passing a `NULL` from there. This check is needed to avoid dereferencing the NULL pointer

prov/efa/src/efa_cq.h
sunkuamzn: double pointer because it's two function calls from `rxr_ep_create_ibv_cq_ex` or `efa_cq_set_ibv_cq_ex`

sunkuamzn: Talked offline, single pointer is enough. Fixed.

prov/efa/src/rxr/rxr_ep.c
sunkuamzn: Talked offline. No need for null check because the address is never null, fixed

prov/efa/src/efa_cq.h
sunkuamzn: Talked offline. No need for null check because the address is never null, fixed

shijin-aws (approved):

wzamazon (approved):

wzamazon:
bot:aws:retest

PR #8075, opened 2022-10-05
prov/efa: flush MR cache when fork() is called and fork support is ON

wzamazon:
When fork support is on, register memory regions was not copied to child processes's memory space.

With MR cache turned on, the registered memory regions remain registered after EFA finished send data from it.

As a result, when both fork support and MR cache are turned on, child processes cannot access some memory they need.

This patch addresses the issue by flush the MR cache when fork is called. Flushing MR cache will cause any memory region that is not actively being used (use_cnt == 0) to be de-registered, therefore they will be copied to child process's memory space

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -39,6 +39,8 @@
 #include "rxr_cntr.h"
 #include "rxr_atomic.h"
 
+struct dlist_entry g_efa_domain_list;
+
 static int efa_domain_close(fid_t fid);
 
 static struct fi_ops efa_ops_domain_fid = {
@@ -161,6 +163,7 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	if (!efa_domain)
 		return -FI_ENOMEM;
 
+	dlist_init(&efa_domain->list_entry);
 	efa_domain->fabric = container_of(fabric_fid, struct efa_fabric,
 					  util_fabric.fabric_fid);
 
@@ -263,6 +266,7 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		goto err_free;
 	}
 
+	dlist_insert_tail(&efa_domain->list_entry, &g_efa_domain_list);
 	return 0;
 
 err_free:
@@ -287,6 +291,8 @@ static int efa_domain_close(fid_t fid)
 
 	efa_domain = container_of(fid, struct efa_domain,
 				  util_domain.domain_fid.fid);
+
+	dlist_remove(&efa_domain->list_entry);
 
 	if (efa_domain->cache) {
 		ofi_mr_cache_cleanup(efa_domain->cache);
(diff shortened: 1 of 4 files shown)


prov/efa/src/efa_fork_support.c
```diff
@@ -176,6 +176,29 @@ void efa_atfork_callback()
...
+ * Going through all domains, and flush the MR caches in them
+ * until all inactive MRs are de-registered.
+ * This makes all memory regions that are not actively used
+ * in data transfer visible to the child process.
+ */
+void efa_atfork_callback_flush_mr_cache()
+{
+	struct dlist_entry *tmp;
+	struct efa_domain *efa_domain;
+	bool flush_lru = true;
```
shijin-aws, line 191: nit: I would combine this line with L197 as `while(ofi_mr_cache_flush(efa_domain->cache, true)`

prov/efa/src/efa_domain.c
```diff
@@ -263,6 +266,7 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		goto err_free;
 	}
 
+	dlist_insert_tail(&efa_domain->list_entry, &g_efa_domain_list);
```
shijin-aws, line 269: there could be multiple efa domains ?

shijin-aws: I guess it is for the convenience of getting efa domain from `g_efa_domain_list` in the fork callback.

prov/efa/src/efa_fork_support.c
wzamazon: I wanted to use variable name to remind reader what the `true` means.  

Maybe I will change to 
```
ofi_mr_cache_flush(efa_domain->cache, true /* flush_lru */)
```

prov/efa/src/efa_domain.c
wzamazon: Yes. there can be multiple efa domains. 

Application can create multiple domains. In fact, Open MPI create 2 domains: one for mtl/ofi, and other for btl/ofi.
Also aws-ofi-nccl create one domain for each NIC.

prov/efa/src/efa_fork_support.c
```diff
@@ -230,14 +253,21 @@ int efa_fork_support_enable_if_requested(struct fid_domain* domain_fid)
...
+			ret = pthread_atfork(efa_atfork_callback_flush_mr_cache, NULL, NULL);
+		}
+
 		if (ret) {
 			EFA_WARN(FI_LOG_DOMAIN,
 				 "Unable to register atfork callback: %s\n",
 				 strerror(-ret));
 			return ret;
 		}
+
```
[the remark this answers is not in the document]
wzamazon: not random, if came from that we adjusted the loop.

prov/efa/src/efa_domain.c
shijin-aws: I see, thanks!

shijin-aws (approved):

darrylabbate (approved):

Issue #7248, opened 2021-11-12
prov/rxd: segfault running fi_rdm_atomic with 128-bit integers


shefty:
The following segfault occurs with fi_rdm_atomic with a release build of libfabric:

```
#0  0x00007ffff7a2d9bd in ofi_cswap_OFI_OP_CSWAP_EQ_ofi_int128_t ()
   from /usr/local/lib/libfabric.so.1
#1  0x00007ffff7ab53e6 in rxd_do_atomic () from /usr/local/lib/libfabric.so.1
#2  0x00007ffff7ab5572 in rxd_progress_atom_op () from /usr/local/lib/libfabric.so.1
#3  0x00007ffff7ab5766 in rxd_progress_op () from /usr/local/lib/libfabric.so.1
#4  0x00007ffff7ab6536 in rxd_handle_recv_comp () from /usr/local/lib/libfabric.so.1
#5  0x00007ffff7ab8805 in rxd_ep_progress () from /usr/local/lib/libfabric.so.1
#6  0x00007ffff7a3904d in ofi_cq_progress () from /usr/local/lib/libfabric.so.1
#7  0x00007ffff7a38486 in ofi_cq_readfrom () from /usr/local/lib/libfabric.so.1
#8  0x000055555555bcae in fi_cq_read (count=1, buf=0x7fffffffdd70, 
    cq=0x555555771900) at /usr/local/include/rdma/fi_eq.h:394
#9  ft_spin_for_comp (timeout=-1, total=795, cur=0x5555557676a0 <tx_cq_cntr>, 
    cq=0x555555771900) at common/shared.c:2207
#10 ft_get_cq_comp (cq=0x555555771900, cur=0x5555557676a0 <tx_cq_cntr>, total=795, 
    timeout=-1) at common/shared.c:2299
#11 0x00005555555570e8 in execute_compare_atomic_op (op=<optimized out>)
    at functional/rdm_atomic.c:269
#12 execute_atomic_compare_op (datatype=<optimized out>, op_type=<optimized out>)
    at functional/rdm_atomic.c:285
#13 handle_atomic_compare_op (count=0x5555557750c0, op_type=<optimized out>, 
    run_all_datatypes=<optimized out>) at functional/rdm_atomic.c:285
#14 run_op () at functional/rdm_atomic.c:325
```
Test passes with debug build.

shefty:
I've narrowed a problem down to this macro:
```
[...]

PR #8067, opened 2022-10-05
v1.16.1

shefty:
New package to pull in missing verbs_osd.h file.

Signed-off-by: Sean Hefty <sean.hefty@intel.com>

PR #8076, opened 2022-10-06
prov/rxm: Disable 128-bit atomics

nmorey:
128-bit atomics fail (cause segfault) in release build of provider.  Cause is likely related to alignment.  Fail query calls for 128-bit integers to indicate support is not provided.

https://github.com/ofiwg/libfabric/issues/7248

Signed-off-by: Nicolas Morey-Chaisemartin <nmoreychaisemartin@suse.com>

### DIFF
--- a/prov/rxm/src/rxm_atomic.c
+++ b/prov/rxm/src/rxm_atomic.c
@@ -509,6 +509,12 @@ int rxm_ep_query_atomic(struct fid_domain *domain, enum fi_datatype datatype,
 		return -FI_EINVAL;
 	}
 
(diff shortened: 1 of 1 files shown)


nmorey:
As stated in #7248 last comment, this should probably be applied to the v1.16.x branch as well

shefty:
Intel CI failure is shm+ze.  ignoring.

shefty (approved):

shefty:
I've picked this change into PR #8067, which will add it to v1.16.x and the 1.16.1 release.

PR #8077, opened 2022-10-07
core/peer: New peer provider objects to support collective offload

shefty:
This is the first set of changes targeting the support of a provider that only supports collective offloads.  The peer definitions are generic, and not tied specifically to that use case, but should meet it.  The series adds the peer EQ, peer AV, a new peer data transfer flag/mode/feature, and introduces a new type of provider.

The new provider is called (for now) a function provider.  It requires special handling by the libfabric core, as it's not usable stand-alone, nor does it layer over another provider.

The peer eq and av are what one would expect based on the peer cq and srx structures.

The peer data transfer allows the function provider to send and receive data over the main provider's endpoint.  For example, the collective provider send messages between a remote collective provider.  The main provider is responsible for separating that traffic from application traffic.  (For example, by reserving and using a tag bit.)  The main provider must also route completions back to the function provider.

Before the design draft can be considered complete, a peer EP needs to be defined.  This is needed to allow the function provider to transmit peer transfers.  More thought is needed on this, but the definition will be added to this PR once done.

### DIFF
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -165,6 +165,7 @@ typedef struct fid *fid_t;
 #define FI_COMMIT_COMPLETE	(1ULL << 30)
 #define FI_MATCH_COMPLETE	(1ULL << 31)
 
+#define FI_PEER_AV		(1ULL << 39)
 #define FI_PEER_EQ		(1ULL << 40)
 #define FI_AV_USER_ID		(1ULL << 41)
 #define FI_PEER_SRX		(1ULL << 42)
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -165,6 +165,7 @@ typedef struct fid *fid_t;
 #define FI_COMMIT_COMPLETE	(1ULL << 30)
 #define FI_MATCH_COMPLETE	(1ULL << 31)
 
+#define FI_PEER_DOMAIN		(1ULL << 38)
 #define FI_PEER_AV		(1ULL << 39)
 #define FI_PEER_EQ		(1ULL << 40)
 #define FI_AV_USER_ID		(1ULL << 41)
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -165,6 +165,7 @@ typedef struct fid *fid_t;
 #define FI_COMMIT_COMPLETE	(1ULL << 30)
 #define FI_MATCH_COMPLETE	(1ULL << 31)
 
+#define FI_PEER_EQ		(1ULL << 40)
 #define FI_AV_USER_ID		(1ULL << 41)
 #define FI_PEER_SRX		(1ULL << 42)
 #define FI_PEER_CQ		(1ULL << 43)
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -165,6 +165,7 @@ typedef struct fid *fid_t;
 #define FI_COMMIT_COMPLETE	(1ULL << 30)
 #define FI_MATCH_COMPLETE	(1ULL << 31)
 
+#define FI_PEER_TRANSFER	(1ULL << 36)
 #define FI_PEER_DOMAIN		(1ULL << 38)
 #define FI_PEER_AV		(1ULL << 39)
 #define FI_PEER_EQ		(1ULL << 40)
@@ -371,6 +372,7 @@ static inline uint8_t fi_tc_dscp_get(uint32_t tclass)
 #define FI_RESTRICTED_COMP	(1ULL << 53)
 #define FI_CONTEXT2		(1ULL << 52)
 #define FI_BUFFERED_RECV	(1ULL << 51)
+/* #define FI_PEER_TRANSFER	(1ULL << 36) */
 
 struct fi_tx_attr {
 	uint64_t		caps;
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -199,6 +199,15 @@ struct fi_peer_cq_context {
 
 
 /*
+ * Peer provider domain support.
+ */
+struct fi_peer_domain_context {
+	size_t size;
+	struct fid_domain *domain;
+};
+
+
+/*
  * Peer provider EQ support.
  */
 struct fi_peer_eq_context {
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -129,7 +129,9 @@ struct fid_mem_monitor {
 };
 
 
-/* Peer provider CQ support. */
+/*
+ * Peer provider CQ support.
+ */
 struct fid_peer_cq;
 
 struct fi_ops_cq_owner {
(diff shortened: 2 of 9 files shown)


man/fi_peer.3.md
```diff
@@ -75,6 +75,47 @@ function provider may be written specifically to leverage this capability;
...
+};
+
+struct fid_peer_av {
+	struct fid fid;
+	struct fi_ops_av_owner *owner_ops;
+};
+
+struct fi_peer_av_context {
+	size_t size;
+	struct fi_peer_av *av;
```
j-xiong, line 115: should be `fid_peer_av`

include/rdma/fi_ext.h
```diff
@@ -129,6 +129,29 @@ struct fid_mem_monitor {
...
+/*
+ * Peer provider AV support.
+ */
+struct fid_peer_av;
+
+struct fi_ops_av_owner {
+	size_t	size;
+	int	(*query)(struct fid_peer_av *av, struct fi_av_attr *attr);
+	int	(*lookup)(struct fid_peer_av *av, struct fid_ep *ep,
+			  fi_addr_t *fi_addr);
```
j-xiong, line 141: Any explanation on how the lookup function works? The signature differs from the regular av lookup op.

j-xiong (commented):
Looks good in general. Not sure if peer av is needed for collective offload. We may also need to add an enumerate method to av_set (e.g. `members()`?) or define peer av_set.

man/fi_peer.3.md
```diff
@@ -184,6 +184,32 @@ on the local node.
...
+do not support peer EQs must fail the fi_eq_open() call with -FI_EINVAL
+(indicating an invalid flag).  The fid_eq referenced by struct
+fi_peer_eq_context must remain valid until the peer's EQ is closed.
+
+The data structures to support peer EQs are defined as follows:
+
+```c
+struct fi_peer_eq_context {
+    size_t size;
+    struct fid_peer_eq *eq;
```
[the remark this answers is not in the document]
shefty: This should just be fid_eq.

include/rdma/fi_ext.h
shefty: This takes a local ep as input and returns its address in the AV.  The intent is to get 'my rank'.  We can pick a different name to avoid confusion with the current lookup call.

shefty:
First, maybe we should rename function provider to offload?

The AV will be owned by the core provider.  My guess is the AV sets will be the offload provider.  To create the AV sets on the offload provider, it needs an AV, which is why I added a peer AV.  So the core can import the AV into the offload.  Part of that import is to provide the AV attributes to the offload provider, so it can get the number of entries.  That's the purpose of the query call.

I guess I need more text around how it's expected to be used.

j-xiong:
yes offload provider is more self-explaining.

I now see the purpose of peer AV when av_set is owned by the offload provider. In case two offload providers need to work together will both of them need to implement av_set or do we want a way to share that?

shefty:
I see where you're going with the shared AV set and members() function, and it makes sense.

This use of 2 offload providers that implement the same functionality is off just enough to be a problem.  There's not a reasonable way for a main provider to discover all of this dynamically.  Maybe a better answer is to have the main provider officially use 1 offload provider per feature (e.g. collective).

For the util_coll implementation, we leverage the offload provider architecture as 90% of the way to integrate it with a main provider, but for that last 10%, hook it in directly and not dynamically.  Basically, util_coll does NOT register for fi_getinfo.  The implementation is updated to use the offload peer APIs as its exposed entry points, but the main provider is updated to call some internal util_coll_init() function.

shefty:
Updated based on comments.  Added new patch for peer AV set.

j-xiong:
Looks good. Intel CI is failing with missing "verbs_osd.h" error. May need rebase to get it pass.

shefty:
I updated and expanded the peer transfer definition.  The peer transfer now acts as a 'peer ep' of sorts, providing a simpler setup between a peer and offload provider.  It's still in a draft form because as I'm writing this, I don't think we want the peer_av field there.

include/rdma/fi_ext.h
```diff
@@ -203,6 +260,20 @@ struct fi_peer_srx_context {
...
 };
 
+
+/*
+ * Peer transfers
+ */
+struct fi_peer_transfer_context {
+	size_t size;
+	struct fi_info *info;
+	struct fid_domain *domain;
```
shefty, line 270: The domain isn't needed.  We already need to have a domain opened before opening the EP.  So, adding a peer domain would cover this.  What the offload provider needs is access to memory registration calls.

include/rdma/fi_ext.h
```diff
@@ -203,6 +260,20 @@ struct fi_peer_srx_context {
...
+	struct fid_ep *ep;
+	struct fid_cq *cq;
+	struct fid_peer_av *av;
```
shefty, line 273: The AV may not be needed either.  The main provider could just bind the EP with the AV.  This would basically be an optimization.  If we go this route, the peer CQ(s) could be added too.

include/rdma/fi_ext.h
```diff
@@ -203,6 +260,20 @@ struct fi_peer_srx_context {
...
+
+/*
+ * Peer transfers
+ */
+struct fi_peer_transfer_context {
+	size_t size;
+	struct fi_info *info;
+	struct fid_domain *domain;
+	struct fid_ep *ep;
+	struct fid_cq *cq;
```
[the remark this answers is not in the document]
j-xiong: How would the callback be registered then? Should an endpoint that supports FI_PEER_TRANFER flag have a new op `set_peer_xfer_callback`? or via `fi_control`?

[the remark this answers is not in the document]
shefty: I haven't thought that far ahead, but something like how srx is setup should work.
```
struct fi_ops_transfer_peer  {
    size_t size;
    int (*done_handler)(...)
};
```

shefty:
Updated to add peer domain and simplify the peer transfer structures.  I *think* this is enough to start with an implementation, and let problems uncovered there drive further changes.

include/rdma/fabric.h
```diff
@@ -165,6 +165,7 @@ typedef struct fid *fid_t;
 #define FI_COMMIT_COMPLETE	(1ULL << 30)
 #define FI_MATCH_COMPLETE	(1ULL << 31)
 
+#define FI_PEER_TRANSFER	(1ULL << 36)
```
j-xiong, line 168: Why skip 37?

shefty: 2 reasons.  1. I've had to adjust it 3 times now.  2. Mainly to leave room for PEER_EP.

What I want to do is merge all of the FI_PEER_xxx into just FI_PEER, except for FI_PEER_TRANSFER, which needs special handling.  This use of individual FI_PEER_xxx isn't needed currently.  It's only needed if we report these as cap bits.

j-xiong:
Is it just a rebase?

shefty:
Yes - just a rebase to pull in the CI script changes.

shefty:
@j-xiong - any other comments?  Otherwise, I'd like to merge, so development can begin based on these changes, with further updates resulting from trying to use this.

j-xiong:
Looks good to me as it is.

Issue #8081, opened 2022-10-10
core: CUDA memory monitor left freed memory region in cache


wzamazon:
**Describe the bug**
CUDA memory monitor does not monitor the event of CUDA memory being released. Instead if compare the buffer id of input memory region and memory regions in the cache.

Therefore, when a registered memory region was freed (via `cudaFree`), the registration remained in the cache.

Such a dead region only got removed from the cache when application try to register a new region that overlaps with the dead region.

**To Reproduce**
1. enable MR cache for cuda memory
2. register a cuda memory by `fi_mr_reg`
3. close the registration
4. call `cudaFree` on the memory region
5. check that the region is still in the cache

**Expected behavior**
CUDA memory monitor to monitor call to `cudaFree`, and remove a freed region from cache.

shefty:
The memory monitor doesn't allocate the memory, so it can't call free.  I agree there's a problem here.  Ze has the same issue.  AFAICT, only RoCR has the support needed to release the registration by intercepting the free call.  Otherwise, the registered memory will remain in the cache until it is flushed.

wzamazon:
> The memory monitor doesn't allocate the memory, so it can't call free. I agree there's a problem here. Ze has the same issue. 

I think we need a new type of cuda memory monitor, that intercept call to cudaMalloc/cudaFree. like cuda memhooks.

shefty:
We just need to extend the existing monitor.  How do you intercept the malloc/free calls?

wzamazon:
> How do you intercept the malloc/free calls?

I think we can make symbol like `cudaFree` and `cudaMalloc` pointing to a new function.

Like what we did for the old memhooks monitor (not the new one based on patcher).

iziemba:
There was a discussion a few years about about intercepting instead of relying on buffer IDs. I think the issue with the old memhooks intercepting approach being applied to cudaFree/cudaMalloc was it could not handle applications compiled by nvcc where the cudaFree/cudaMalloc calls were static.

iziemba:
https://github.com/ofiwg/libfabric/issues/5789

wckzhang:
I think adding some context would help here. As far as I understand, with the mr_map code, the mr keys are used for lookup. This can cause an issue when looking up a staled key (new registration gets a new key), but for the wrong region.

shefty:
There isn't a problem accessing a stale region.

[...]

PR #8082, opened 2022-10-10
prov/efa: remove de-allocated CUDA memory region during registration

wzamazon:
This patch is a temporary workaround for issue:

	https://github.com/ofiwg/libfabric/issues/8081

which is current cuda memory monitor does not monitor the event of cuda memory being freed, causing freed memory regions remain in cache.

The long term fix is to implement a cuda memory monitor that monitors cuda memory free event. Before that is implemented, this patch will detect such freed cuda memory region when registering a new memory region, and remove it from cache.

This patch is to be reverted after proper cuda memory monitor was implemented and used by efa.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -33,6 +33,9 @@
 #include "config.h"
 #include <ofi_util.h>
 #include "efa.h"
+#if HAVE_CUDA
+#include <cuda.h>
+#endif
 
 static int efa_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 			  uint64_t flags, struct fid_mr **mr_fid);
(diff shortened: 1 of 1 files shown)


wzamazon:
@wckzhang Can you take a look at this PR? especially the lock applied around the call to `ofi_mr_cache_notify`.

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+
+	/* FI_ENOKEY can happen legitimately when the key was assigned a CUDA memory region, and the memory region has been freed
+	 * after the registration.
+	 *
+	 * This is because of the the combination of following 3 things:
+	 *  1) cuda memory uses memory cache.
+	 *  2) cuda memory cache's monitor is not monitoring call to cudaFree().
+	 *  3) kernel released the device part of memory registration when cudaFree() is called on a registered cuda memory.
+	 *
+	 * Therefore, when application call cudaFree() on a registered CUDA memory region, the device released the device
```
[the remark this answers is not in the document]
wzamazon: yes.

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+				 existing_mr->peer.iface,
+				 existing_mr->ibv_mr->length);
+			return -FI_ENOKEY;
+		}
+
+		assert(existing_mr->peer.iface == FI_HMEM_CUDA);
+		err = ofi_cuPointerGetAttribute(&cuda_memory_type, CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
+						(CUdeviceptr)existing_mr->ibv_mr->addr);
+		if (!err) {
+			EFA_WARN(FI_LOG_DOMAIN, "key %ld already assigned to cuda buffer: %p cuda_memor_type: %d length: %ld\n",
```
shijin-aws, line 616: Is this warning message correct? Should it be like `failed to get the cuda ptr attribute? ...`

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+		pthread_mutex_unlock(&mm_lock);
+		pthread_rwlock_unlock(&mm_list_rwlock);
+
+		/* due to MR cache's deferred de-registration, ofi_mr_cache_notify() only move the region to dead_region_list
+		 * ofi_mr_cache_flush() will actually remove the region from cache.
+		 * lru is a list of regions that are still active, so we set flush_lru to false
+		 */
+		ofi_mr_cache_flush(efa_mr->domain->cache, false /*flush_lru */);
+
+		err = ofi_mr_map_insert(&efa_mr->domain->util_domain.mr_map, mr_attr,
```
shijin-aws, line 642: do we need lock here?

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+			fi_strerror(-err),
+			efa_mr->mr_fid.key,
+			mr_attr->mr_iov->iov_base,
+			mr_attr->iface,
+			mr_attr->mr_iov->iov_len);
+		return err;
+	}
+
+	return 0;
+}
```
shijin-aws, line 683: need a unlock?

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+	 *
+	 * TODO: remove this branch once CUDA memhook monitor is implemented and used by efa.
+	 */
+	if (err == -FI_ENOKEY) {
+		existing_mr_fid = ofi_mr_map_get(&efa_mr->domain->util_domain.mr_map, efa_mr->mr_fid.key);
+		assert(existing_mr_fid);
+		existing_mr = container_of(existing_mr_fid, struct efa_mr, mr_fid);
+
+		if (existing_mr->peer.iface != FI_HMEM_CUDA) {
+			EFA_WARN(FI_LOG_DOMAIN, "key %ld already assigned to buffer: %p hmem_iface: %d length: %ld\n",
```
shijin-aws, line 604: warning message seems incorrect.

prov/efa/src/efa_mr.c
wzamazon: The message is correct. The branch is for `!err` (no error)

prov/efa/src/efa_mr.c
wzamazon: unlock of which lock?

shijin-aws: there is a `ofi_genlock_lock(&efa_mr->domain->util_domain.lock);` before the mr map insert?

prov/efa/src/efa_mr.c
shijin-aws: Nm, I get the point now.

prov/efa/src/efa_mr.c
shijin-aws: So, when ofi_cuPointerGetAttribute returns success, there must an error in the mr cache, because it failed to re-use the existing mr region. Right?

wckzhang, line 616: memor -> memory

wckzhang, line 616: Also print out that cuPointerGetAttribute failed

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+			 existing_mr->ibv_mr->length);
+
+		assert(efa_mr->domain->cache);
+		pthread_rwlock_rdlock(&mm_list_rwlock);
+		pthread_mutex_lock(&mm_lock);
+		ofi_mr_cache_notify(efa_mr->domain->cache, existing_mr->ibv_mr->addr, existing_mr->ibv_mr->length);
+		pthread_mutex_unlock(&mm_lock);
+		pthread_rwlock_unlock(&mm_list_rwlock);
+
+		/* due to MR cache's deferred de-registration, ofi_mr_cache_notify() only move the region to dead_region_list
```
wckzhang, line 636: I think you should specify why this is important. (IE, the de-register call needs to be triggered since the deregister function has the mr map remove function)

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+	return 0;
+}
+#else
+static
+int efa_mr_update_domain_mr_map(struct efa_mr *efa_mr, struct fi_mr_attr *mr_attr)
+{
+	int err;
+
+	mr_attr->requested_key = efa_mr->mr_fid.key;
+	ofi_genlock_lock(&efa_mr->domain->util_domain.lock);
```
wckzhang, line 667: You need to unlock this lock

wckzhang (changes requested):

prov/efa/src/efa_mr.c
wckzhang: I don't understand this. If cuPointerGetAttribute passed, that doesn't mean anything. That addr could have been reallocated at some point.

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,139 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+			return -FI_ENOKEY;
+		}
+
+		EFA_INFO(FI_LOG_DOMAIN, "key %ld has been assigned to cuda buffer: %p length: %ld, which has since been freed\n",
+			 existing_mr->mr_fid.key,
+			 existing_mr->ibv_mr->addr,
+			 existing_mr->ibv_mr->length);
+
+		assert(efa_mr->domain->cache);
+		pthread_rwlock_rdlock(&mm_list_rwlock);
```
wckzhang, line 630: I don't think you need these locks...IIRC these locks are to protect the monitors, not the cache

```
/* The read-write lock is an additional lock used to protect the dlist_entry
 * list of ofi_mem_monitor. Due to the necessity of releasing the mm_lock
 * while walking the dlist in ofi_monitor_notify, we need a separate lock to
 * ensure thread safety. This must be a read-write lock because
 * ofi_monitor_notify may be recursive and cannot block multiple walks from
 * occurring at the same time.
 ```
The requirement to calling ofi_mr_cache_notify is:

```
/* Caller must hold ofi_mem_monitor lock as well as unsubscribe from the region */
```

File: prov/efa/src/efa_mr.c (hunk 1 above)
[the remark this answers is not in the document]
Comment (wzamazon) [reply prov/efa/src/efa_mr.c:588]: changed to  "kernel space memory registration."

File: prov/efa/src/efa_mr.c (hunk 2 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c:616]: If cuPointerGetAttribute passed, it means there is an alive CUDA memory region that is registered with the key, which we cannot handle and has to return error.

The only situation we can handle is that there is a dead region in the cache, for which we can remove it from cache.

Comment (wzamazon) [reply prov/efa/src/efa_mr.c:616]: I did not find the term "memor".

File: prov/efa/src/efa_mr.c (hunk 8 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c:630]: but `ofi_mr_cache_notify` simply call `ofi_mr_cache_uncache_entry`, which does not have any lock in it.

File: prov/efa/src/efa_mr.c (hunk 7 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c:669]: added

File: prov/efa/src/efa_mr.c (hunk 4 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c:686]: added

File: prov/efa/src/efa_mr.c (hunk 2 above)
Comment (shijin-aws) [reply prov/efa/src/efa_mr.c:616]: >That addr could have been reallocated at some point.

In that case, shouldn't the cuda monitor have moved it to a dead region. so a `ofi_mr_cache_flush` could make it removed?

Comment (wzamazon) [reply prov/efa/src/efa_mr.c:616]: > In that case, shouldn't the cuda monitor have moved it to the dead region list. so a ofi_mr_cache_flush could make it removed?

Either way, if a pointer in mr map is valid. There is no way the function is able to handle the error, and return error is the right approach.

I think I will re-arrange the code and adjust the comment to make the point more clear.

File: prov/efa/src/efa_mr.c (hunk 8 above)
Comment (wckzhang) [reply prov/efa/src/efa_mr.c:630]: Yes, when I said you don't need these locks, I meant you don't need the mm_list_rwlock. The requirement to calling ofi_mr_cache_notify is to hold the ofi_mem_monitor_lock, which is mm_lock <-- which you do need.

I'm saying you can remove the rwlock locks.

File: prov/efa/src/efa_mr.c (hunk 2 above)
Comment (wckzhang) [reply prov/efa/src/efa_mr.c:616]: If it's valid but the key doesn't match, that means that the key is invalid. I don't see why we can't kick it out of the cache.

Comment (wckzhang) [reply prov/efa/src/efa_mr.c:616]: cuda_memor_type -> cuda_memory_type

Comment (shijin-aws) [reply prov/efa/src/efa_mr.c:616]: > If it's valid but the key doesn't match, that means that the key is invalid. I don't see why we can't kick it out of the cache.

I assume the `key` you refer to is the mr key.  But here the error is that the mr key is the same, but from a different cuda buffer.

Comment (wzamazon) [reply prov/efa/src/efa_mr.c:616]: I think this is an erronenous situation, and we should return error

Comment (shijin-aws) [reply prov/efa/src/efa_mr.c:616]: I agree we need to return error here. But it is hard for me to understand how this could happen. EFA driver returns the same key from a different cuda buffer registration?

File: prov/efa/src/efa_mr.c (hunk 8 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c]: Thank you!
I removed the `rwlock` in latest revision.

File: prov/efa/src/efa_mr.c (hunk 6 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c]: adjusted the comment in new revision

File: prov/efa/src/efa_mr.c (hunk 2 above)
Comment (wzamazon) [reply prov/efa/src/efa_mr.c]: fixed in new revision.

Comment (wzamazon) [reply prov/efa/src/efa_mr.c]: I cannot think of a situation it will happen. But if it happened, we return error.

File: prov/efa/src/efa_mr.c [hunk 9]
```diff
@@ -547,6 +550,159 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+			 existing_mr->ibv_mr->addr,
+			 existing_mr->peer.iface,
+			 existing_mr->ibv_mr->length);
+		return -FI_ENOKEY;
+	}
+
+	err = ofi_cuPointerGetAttribute(&cuda_memory_type, CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
+					(CUdeviceptr)existing_mr->ibv_mr->addr);
+	if (err != CUDA_ERROR_INVALID_VALUE) {
+		/* no way we can recover from this situation, return error code */
```
shijin-aws, line 630: Readers may not be familar what CUDA_ERROR_INVALID_VALUE mean here. I would explain a bit more that it means the cuda buffer was freed.

shijin-aws (approved):

prov/efa/src/efa_mr.c
wzamazon: Added some comments to explain the code.

shijin-aws (approved):

prov/efa/src/efa_mr.c
```diff
@@ -547,6 +550,163 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+		return -FI_ENOKEY;
+	}
+
+	/* According to CUDA document, the return code of ofi_cuPointerGetAttribute() being CUDA_ERROR_INVALID_VALUE
+	 * means the exising_mr's pointer is NOT allocated by, mapped by or registered with CUDA. In this context,
+	 * it means existing_mr's address has been de-allocated (which is the only situation we can recover from).
+	 */
+	err = ofi_cuPointerGetAttribute(&cuda_memory_type, CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
+					(CUdeviceptr)existing_mr->ibv_mr->addr);
+	if (err != CUDA_ERROR_INVALID_VALUE) {
```
wckzhang, line 633: We discussed and there needs to be an additional check here that compares buffer id as there is a valid situation where the device re-uses the virutal address for a new allocation.

wzamazon: In latest revision, I added the check for buffer ID. please take a look

shijin-aws (approved):

prov/efa/src/efa_mr.c
```diff
@@ -560,6 +563,191 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+			efa_mr->mr_fid.key,
+			mr_attr->mr_iov->iov_base,
+			mr_attr->iface,
+			mr_attr->mr_iov->iov_len);
+		return err;
+	}
+
+	return 0;
+}
+#endif
```
wckzhang, line 749: I'd like the endif and else to have #HAVE_CUDA so that it's easier to know what each block belongs to

prov/efa/src/efa_mr.c
```diff
@@ -560,6 +563,191 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
 }
 #endif /* HAVE_SYNAPSEAI */
 
+#if HAVE_CUDA
+static inline
+int efa_mr_is_cuda_memory_freed(struct efa_mr *efa_mr, int *freed)
```
wckzhang, line 568: change freed to a bool?

wckzhang (approved):
Okay for me, just address the comments I posted, thanks.

prov/efa/src/efa_mr.c
wzamazon: changed in new revision.

prov/efa/src/efa_mr.c
wzamazon: added in new revision.

PR #8085, opened 2022-10-10
prov/efa: Fix incorrect argument for rxr_pkt_post_or_queue

shijin-aws:
In rxr_pkt_post_req, the rxr_pkt_post_or_queue's inject argument was passed as flags, which is incorrect. This patch make rxr_pkt_post_or_queue called with the correct inject argument.


wzamazon (approved):

wzamazon:
Do we need to back port?

shijin-aws:
@wzamazon  Will backport to v1.16.x.

wenduwan (approved):

PR #8086, opened 2022-10-10
[v1.16.x] prov/efa: add missing efa header

aingerson:
Cherry-picked from commit 825dff06ac52825e8c8693ebfd0c07ac9ebb0b3f

### DIFF
--- a/prov/efa/Makefile.include
+++ b/prov/efa/Makefile.include
@@ -75,6 +75,7 @@ _efa_headers = \
 	prov/efa/src/efa_user_info.h \
 	prov/efa/src/efa_prov_info.h \
 	prov/efa/src/efa_fork_support.h \
+	prov/efa/src/efa_cq.h \
 	prov/efa/src/rxr/rxr.h \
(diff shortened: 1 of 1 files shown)


shefty:
@wckzhang - Does AWS provide their own libfabric package?  We *just* released v1.1.6.1 because verbs was missing a file.

The ubuntu 18.04 build builds most of the providers:
```
Built-in providers:	opx dmabuf_peer_mem hook_hmem hook_debug perf rstream shm rxd mrail rxm net tcp udp usnic efa verbs sockets psm3
```
I think we need to extend one of the ubuntu (e.g. 20.04 gcc) builds to add make distcheck and rpm_build.

wzamazon:
>Does AWS provide their own libfabric package? 

Yes.

This PR was not necessary for v1.16.x branch though. The PR that introduced `efa_cq.h` was not part of 1.16.x branch

wzamazon:
>I think we need to extend one of the ubuntu (e.g. 20.04 gcc) builds to add make distcheck and rpm_build.

I think it is a good idea to add such check.

PR #8094, opened 2022-10-11
prov/efa: Refactor the req header size functions

shijin-aws:
Currently, both `rxr_pkt_req_hdr_size` and `rxr_pkt_req_header_size` in `rxr_pkt_type_req.c`  calculate the header size of REQ packets and have considerable code duplication. This PR did some refactors in `rxr_pkt_type_req.c` to remove the code duplication and rename the functions to make them more intuitive.

### DIFF
--- a/prov/efa/src/rxr/rxr_pkt_type_req.c
+++ b/prov/efa/src/rxr/rxr_pkt_type_req.c
@@ -106,7 +106,7 @@ size_t rxr_pkt_req_data_offset(struct rxr_pkt_entry *pkt_entry)
 	size_t pkt_data_offset;
 
 	pkt_type = rxr_get_base_hdr(pkt_entry->pkt)->type;
-	pkt_data_offset = rxr_pkt_req_hdr_size(pkt_entry);
+	pkt_data_offset = rxr_pkt_req_hdr_size_from_pkt_entry(pkt_entry);
 	assert(pkt_data_offset > 0);
 
 	if (pkt_type == RXR_RUNTREAD_MSGRTM_PKT ||
@@ -210,7 +210,7 @@ void rxr_pkt_init_req_hdr(struct rxr_ep *ep,
 	}
 
 	pkt_entry->addr = tx_entry->addr;
-	assert(opt_hdr - pkt_entry->pkt == rxr_pkt_req_hdr_size(pkt_entry));
+	assert(opt_hdr - pkt_entry->pkt == rxr_pkt_req_hdr_size_from_pkt_entry(pkt_entry));
 }
 
 /**
@@ -320,34 +320,14 @@ uint32_t *rxr_pkt_req_connid_ptr(struct rxr_pkt_entry *pkt_entry)
 	return NULL;
 }
 
-size_t rxr_pkt_req_hdr_size(struct rxr_pkt_entry *pkt_entry)
+size_t rxr_pkt_req_hdr_size_from_pkt_entry(struct rxr_pkt_entry *pkt_entry)
 {
-	char *opt_hdr;
 	struct rxr_base_hdr *base_hdr;
-	struct rxr_req_opt_raw_addr_hdr *raw_addr_hdr;
+	uint32_t rma_iov_count;
 
 	base_hdr = rxr_get_base_hdr(pkt_entry->pkt);
-	opt_hdr = (char *)pkt_entry->pkt + rxr_pkt_req_base_hdr_size(pkt_entry);
-
-	/*
-	 * It is not possible to have both optional raw addr header and optional
-	 * connid header in a packet header.
-	 */
-	if (base_hdr->flags & RXR_REQ_OPT_RAW_ADDR_HDR) {
-		assert(!(base_hdr->flags & RXR_PKT_CONNID_HDR));
-		raw_addr_hdr = (struct rxr_req_opt_raw_addr_hdr *)opt_hdr;
-		opt_hdr += sizeof(struct rxr_req_opt_raw_addr_hdr) + raw_addr_hdr->addr_len;
-	}
-
-	if (base_hdr->flags & RXR_REQ_OPT_CQ_DATA_HDR)
-		opt_hdr += sizeof(struct rxr_req_opt_cq_data_hdr);
-
-	if (base_hdr->flags & RXR_PKT_CONNID_HDR) {
-		assert(!(base_hdr->flags & RXR_REQ_OPT_RAW_ADDR_HDR));
-		opt_hdr += sizeof(struct rxr_req_opt_connid_hdr);
-	}
-
-	return opt_hdr - (char *)pkt_entry->pkt;
+	rma_iov_count = rxr_pkt_hdr_rma_iov_count(pkt_entry);
+	return rxr_pkt_req_header_size(base_hdr->type, base_hdr->flags, rma_iov_count);
 }
 
 int64_t rxr_pkt_req_cq_data(struct rxr_pkt_entry *pkt_entry)
@@ -463,7 +443,7 @@ int rxr_pkt_init_rtm(struct rxr_ep *ep,
 	rtm_hdr->msg_id = tx_entry->msg_id;
 
 	data_size = MIN(tx_entry->total_len - data_offset,
-			ep->mtu_size - rxr_pkt_req_hdr_size(pkt_entry));
+			ep->mtu_size - rxr_pkt_req_hdr_size_from_pkt_entry(pkt_entry));
 	if (data_size + data_offset < tx_entry->total_len) {
 
 		if (tx_entry->max_req_data_size > 0)
@@ -473,7 +453,7 @@ int rxr_pkt_init_rtm(struct rxr_ep *ep,
 			data_size &= ~(CUDA_MEMORY_ALIGNMENT -1);
 	}
 
-	ret = rxr_pkt_init_data_from_tx_entry(ep, pkt_entry, rxr_pkt_req_hdr_size(pkt_entry),
+	ret = rxr_pkt_init_data_from_tx_entry(ep, pkt_entry, rxr_pkt_req_hdr_size_from_pkt_entry(pkt_entry),
 					      tx_entry, data_offset, data_size);
 	return ret;
 }
@@ -732,7 +712,7 @@ ssize_t rxr_pkt_init_longread_rtm(struct rxr_ep *ep,
 	rtm_hdr->send_id = tx_entry->tx_id;
 	rtm_hdr->read_iov_count = tx_entry->iov_count;
 
-	hdr_size = rxr_pkt_req_hdr_size(pkt_entry);
+	hdr_size = rxr_pkt_req_hdr_size_from_pkt_entry(pkt_entry);
 	read_iov = (struct fi_rma_iov *)((char *)pkt_entry->pkt + hdr_size);
 	err = rxr_read_init_iov(ep, tx_entry, read_iov);
 	if (OFI_UNLIKELY(err))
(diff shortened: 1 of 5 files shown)


wenduwan (commented):
I think the change is on the right track to reduce confusion on which of `rxr_pkt_req_hdr_size` and `rxr_pkt_req_header_size` to use, due to their similar names. But the resulting change makes me wonder if it is better to do this instead:
1. Rename `rxr_pkt_req_header_size` to i.e. `rxr_pkt_type_offset`. This makes the name substantially different from `rxr_pkt_req_hdr_size` to avoid confusion
2. Keep `rxr_pkt_req_hdr_rma_iov_count`, and use it in tandem with `rxr_pkt_type_offset` inside `rxr_pkt_req_hdr_size`.

This should help reduce the boilderplates i.e. hardcoding `rma_iov_count` for certain pkt types. And make it more obvious which of `rxr_pkt_req_hdr_size` and `rxr_pkt_type_offset` is appropriate.

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -213,31 +213,43 @@ void rxr_pkt_init_req_hdr(struct rxr_ep *ep,
...
 }
 
+/**
+ * @brief Get the value of rma_iov_count in req pkt header.
+ *
+ * @param[in] pkt_entry the pkt entry
+ * @return The rma_iov_count in the pkt header, if it is part of the header.
+ * Otherwise return 0.
+ */
+uint32_t rxr_pkt_req_hdr_rma_iov_count(struct rxr_pkt_entry *pkt_entry)
```
wzamazon, line 227: Do you think it would be better to use this method?

https://github.com/ofiwg/libfabric/blob/a2120d6fc633abb76f158ca4fce79a0a80a62a70/prov/efa/src/rxr/rxr_pkt_hdr.h#L48

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -213,31 +213,43 @@ void rxr_pkt_init_req_hdr(struct rxr_ep *ep,
...
-		hdr_size += rxr_get_rtr_hdr(pkt_entry->pkt)->rma_iov_count * sizeof(struct fi_rma_iov);
-	else if (base_hdr->type == RXR_WRITE_RTA_PKT ||
-		 base_hdr->type == RXR_DC_WRITE_RTA_PKT ||
-		 base_hdr->type == RXR_FETCH_RTA_PKT ||
-		 base_hdr->type == RXR_COMPARE_RTA_PKT)
-		hdr_size += rxr_get_rta_hdr(pkt_entry->pkt)->rma_iov_count * sizeof(struct fi_rma_iov);
-
-	return hdr_size;
+	return REQ_INF_LIST[base_hdr->type].base_hdr_size +
+		rxr_pkt_req_hdr_rma_iov_count(pkt_entry) * sizeof(struct fi_rma_iov);
```
wzamazon, line 256: I suggest introduce a temporary variable `rma_iov_count` and adjust the indentation.

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -104,9 +104,12 @@ size_t rxr_pkt_req_data_offset(struct rxr_pkt_entry *pkt_entry)
 {
 	int pkt_type, read_iov_count;
 	size_t pkt_data_offset;
+	struct rxr_base_hdr *base_hdr;
 
-	pkt_type = rxr_get_base_hdr(pkt_entry->pkt)->type;
-	pkt_data_offset = rxr_pkt_req_hdr_size(pkt_entry);
+	base_hdr = rxr_get_base_hdr(pkt_entry->pkt);
+	pkt_type = base_hdr->type;
+	pkt_data_offset = rxr_pkt_req_header_size(pkt_type, base_hdr->flags,
```
wzamazon, line 111: Can we use the name `rxr_pkt_req_hdr_size`?  Because we use `hdr` in almost all other places.

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -104,9 +104,12 @@ size_t rxr_pkt_req_data_offset(struct rxr_pkt_entry *pkt_entry)
...
+		rxr_pkt_req_hdr_rma_iov_count(pkt_entry));
```
wzamazon, line 112: use a variable `rma_iov_count` and adjust the indentation.

wzamazon (commented):
Maybe we can go 1 step further to have a function `rxr_pkt_hdr_size` that can handle all packet types. I think the function would be helpful.

wzamazon:
> Rename rxr_pkt_req_header_size to i.e. rxr_pkt_type_offset. 

hdr_size and offset are different, and cannot be used interchangably.

shijin-aws:
>Keep rxr_pkt_req_hdr_rma_iov_count, and use it in tandem with rxr_pkt_type_offset inside rxr_pkt_req_hdr_size.

I think it would be better to rename `rxr_pkt_req_hdr_size` to `rxr_pkt_req_header_size_from_pkt_entry`, and make it call `rxr_pkt_req_hdr_rma_iov_count` and `rxr_pkt_req_header_size` inside. So we won't introduce more changes in the consumers of the current `rxr_pkt_req_hdr_size`.

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Sure, will do.

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Could you elaborate? How can I use this method?

wzamazon:
I think we need to put all header size related functions like `hdr_size()`, `max_hdr_size()`, `_offset` into one place.

prov/efa/src/rxr/rxr_pkt_type_req.c
wzamazon: something like 

```
static const init offset_of_rma_iov_in_hdr[] = {
     [EAGER_RTW] = offset_of(rma_iov_count, struct rxr_eager_rtw_hdr)
    ....
};

int *rma_iov_pointer = (int *)(pkt_entry->pkt + offset_of_rma_iov_count_in_hdr[pkt_type];
return *rma_iov_pointer;
```

shijin-aws: Cool, will do.

shijin-aws:
> I think we need to put all header size related functions like `hdr_size()`, `max_hdr_size()`, `_offset` into one place.

Having them in a new file `rxr_pkt_hdr.c` ?

wzamazon:
@shijin-aws

I like the new proposal.

However it seems that the  function `rxr_pkt_req_hdr_rma_iov_count` will only be called by `rxr_pkt_req_hdr_from_pkt_entry`. so it seems that having this function is not really necessary.

So, overall we just need to:

1. rename current `rxr_pkt_req_header_size` to `rxr_pkt_req_hdr_size`.
2. rename current `rxr_pkt_req_hdr_size` to `rxr_pkt_req_hdr_size_from_pkt_entry`.

But my previous comments about move them to same location still stand.

shijin-aws:
>However it seems that the function rxr_pkt_req_hdr_rma_iov_count will only be called by rxr_pkt_req_hdr_from_pkt_entry. so it seems that having this function is not really necessary.

@wzamazon Not exactly true, this PR also makes `rxr_pkt_req_base_hdr_size` call `rxr_pkt_req_hdr_rma_iov_count` to remove the code duplication.

wzamazon:
@shijin-aws 

`rxr_pkt_hdr.c` make sense.

I think we should merge `rxr_pkt_req_xxxx` into `rxr_pkt_xxx` if possible to avoid confusion.

wzamazon:
> this PR also makes rxr_pkt_req_base_hdr_size call rxr_pkt_req_hdr_rma_iov_count to remove the code duplication.

I see. then add `rxr_pkt_req_hdr_rma_iov_count` probably make sense. 

I would consider make it `rxr_pkt_hdr_rma_iov_count` and return 0 for non REQ packet types.

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Think about it again, I do not think this method simplifies the function. So I would keep the current implementation.

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Fixed

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Done. Also renamed `rxr_pkt_max_header_size` and `rxr_pkt_max_req_header_size`

shijin-aws:
Offline talked with @wzamazon. The current `*req_hdr_size` is coupled with `rxr_pkt_type_req.c` , due to the const variable `REQ_INF_LIST` defined in it. So this PR will not try to move them into a separate file. Can be done in a separate refactor PR in the future.

wenduwan (approved):

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -213,31 +213,45 @@ void rxr_pkt_init_req_hdr(struct rxr_ep *ep,
...
+/**
+ * @brief Get the value of rma_iov_count in pkt header.
+ *
+ * @param[in] pkt_entry the pkt entry
+ * @return The rma_iov_count in the pkt header, if it is part of the header.
+ * Otherwise return 0.
+ */
+uint32_t rxr_pkt_hdr_rma_iov_count(struct rxr_pkt_entry *pkt_entry)
+{
+	int pkt_type = rxr_get_base_hdr(pkt_entry->pkt)->type;
```
wzamazon, line 225: empty line between variable and code.

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -213,31 +213,45 @@ void rxr_pkt_init_req_hdr(struct rxr_ep *ep,
...
+	if (pkt_type == RXR_EAGER_RTW_PKT ||
+	    pkt_type == RXR_DC_EAGER_RTW_PKT ||
+	    pkt_type == RXR_LONGCTS_RTW_PKT ||
+	    pkt_type == RXR_DC_LONGCTS_RTW_PKT ||
+	    pkt_type == RXR_LONGREAD_RTW_PKT)
+		return rxr_get_rtw_base_hdr(pkt_entry->pkt)->rma_iov_count;
+	else if (pkt_type == RXR_SHORT_RTR_PKT ||
```
wzamazon, line 232: `else` is not necessary.

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Added.

prov/efa/src/rxr/rxr_pkt_type_req.c
shijin-aws: Thank you! Fixed.

wzamazon: I still see `else`. Do you forget to push?

shijin-aws: Sorry, I misunderstood your earlier comment. I thought you only wanted to remove the last `else`. Now I remove all the `else`. Please see the latest revision.

wzamazon (approved):

PR #8097, opened 2022-10-12
prov/net(WIP!): Disable the progress thread

sydidelot:
This WIP patch is an attempt at disabling the progress for applications that want to wait directly on the fd associated with an EQ/CQ.

The patch wasn't tested and it will likely fail.


sydidelot:
@shefty I would appreciate your initial thoughts on this patch. Is the code going to the right direction? 
As soon as I get your feedback, I will clean the code and try to make the PR pass the CI pipeline.
Thank you!

shefty:
@sydidelot - I'll take a look.

@ooststep - Please take a look as well and respond if you have any similar changes that you're working on.

ooststep:
I do have a similar set of changes in progress, but am not quite as far along. I am attempting to implement @shefty's description of the changes

ooststep:
These changes are pretty much what I was going for, you were just quicker to them 😛 

#8099 modifies the exclude file for the net provider, which should result in 0 failures with current upstream.  It also renames `xnet_progress->allfds` to `xnet_progress->epoll_fd`

sydidelot:
@ooststep I was quick because Sean is a good teacher :)
I will rebase my patch after #8099 is merged.

prov/net/src/xnet_eq.c
```diff
@@ -118,6 +124,52 @@ static struct fi_ops xnet_eq_fi_ops = {
...
+static int xnet_eq_monitor_all(struct xnet_eq *eq, struct xnet_fabric *fabric)
+{
+	struct xnet_domain *domain;
+	struct dlist_entry *item;
+	int ret;
+
+	ret = xnet_eq_add_progress(eq, &fabric->progress,
+				   &fabric->util_fabric.fabric_fid);
+	if (ret)
+	    return ret;
```
shefty, line 156: Using spaces instead of tab before return

prov/net/src/xnet_eq.c
```diff
@@ -118,6 +124,52 @@ static struct fi_ops xnet_eq_fi_ops = {
...
+
+	ofi_mutex_lock(&fabric->util_fabric.lock);
+	dlist_foreach(&fabric->util_fabric.domain_list, item) {
+		domain = container_of(item, struct xnet_domain,
+				      util_domain.list_entry);
+		ret = xnet_eq_add_progress(eq, &domain->progress,
+					   &domain->util_domain.domain_fid);
+		if (ret)
+		    break;
```
shefty, line 165: Using space instead of tab before break.  If we need to skip the insert below, replace the break with a goto unlock to jump directly to the exit.  See comment below.

prov/net/src/xnet_eq.c
```diff
@@ -118,6 +124,52 @@ static struct fi_ops xnet_eq_fi_ops = {
...
+	}
+	if (ret == 0)
+	    dlist_insert_tail(&eq->wait_eq_entry, &fabric->wait_eq_list);
```
shefty, line 168: Using spaces instead of tab.

prov/net/src/xnet_eq.c
```diff
@@ -118,6 +124,52 @@ static struct fi_ops xnet_eq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+static int xnet_eq_wait_try_func(void *arg)
+{
+	(void)arg;
```
shefty, line 129: We usually use OFI_UNUSED() macro instead, as the definition changes per OS.

prov/net/src/xnet_eq.c
shefty: A failure is unlikely, but the cleanup here isn't complete.  The wait set could have been created by the user.  We may have added some progress epoll_fd's to the wait epoll fd.  We need to remove any that we added.

We need to do the same thing when destroying this EQ.  The destroy case is a little easier, as we can basically just reverse this call.  Cleanup in the error case here is a pain, but we can walk the list backwards from where an error occurred (plus release the fabric).

prov/net/src/xnet_eq.c
```diff
@@ -90,12 +90,18 @@ static int xnet_eq_close(struct fid *fid)
 	struct xnet_eq *eq;
 	int ret;
 
+	eq = container_of(fid, struct xnet_eq, util_eq.eq_fid.fid);
+
+	if (eq->util_eq.wait && eq->util_eq.wait->wait_obj == FI_WAIT_FD) {
+		ofi_mutex_lock(&eq->util_eq.fabric->lock);
+		dlist_remove(&eq->wait_eq_entry);
+		ofi_mutex_unlock(&eq->util_eq.fabric->lock);
```
shefty, line 98: See comment below about cleanup

prov/net/src/xnet_domain.c
```diff
@@ -133,6 +133,32 @@ static struct fi_ops_domain xnet_domain_ops = {
...
+	ofi_mutex_lock(&fabric->util_fabric.lock);
+	dlist_foreach(&fabric->wait_eq_list, item) {
+		eq = container_of(item, struct xnet_eq, wait_eq_entry);
+		if (add)
+		    ret = xnet_eq_add_progress(eq, &domain->progress,
+					       &domain->util_domain.domain_fid);
+		else
+		    ret = xnet_eq_del_progress(eq, &domain->progress);
+		if (ret)
+		    break;
```
shefty, line 156: spaces in place of tabs before ret and break calls

prov/net/src/xnet_cq.c
```diff
@@ -234,6 +234,21 @@ static struct fi_ops xnet_cq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+static int xnet_cq_wait_try_func(void *arg)
+{
+        (void)arg;
```
shefty, line 239: OFI_UNUSED -- see comment below

shefty (changes requested):
Yes, I think this is basically what we want.  I didn't check the CI results to see if maybe we're missing something there.  We do need to cleanup in case of errors.

prov/net/src/xnet_eq.c
sydidelot: My editor was misconfigured. Sorry for all the coding style issues.

prov/net/src/xnet_cq.c
sydidelot: Done

prov/net/src/xnet_domain.c
sydidelot: Done

prov/net/src/xnet_eq.c
sydidelot: Done

prov/net/src/xnet_eq.c
sydidelot: I reworked the error handling according to your review comment. 
Please let me know if it's OK for you now.

prov/net/src/xnet_eq.c
sydidelot: Done

ooststep:
Intel CI failure is unrelated, possibly transient ze-shm failure.  We likely won't see any net provider failures until #8108 is completed, unless we touch util code

sydidelot:
@ooststep Shall we wait for #8108 before we merge this PR? To make sure I don't introduce regressions to the `net` provider.

prov/net/src/xnet_domain.c
```diff
@@ -133,6 +133,28 @@ static struct fi_ops_domain xnet_domain_ops = {
...
+	fabric = container_of(domain->util_domain.fabric,
+			      struct xnet_fabric,
+			      util_fabric.fabric_fid);
+
+	ofi_mutex_lock(&fabric->util_fabric.lock);
+	dlist_foreach(&fabric->wait_eq_list, item) {
+		eq = container_of(item, struct xnet_eq, wait_eq_entry);
+		ret = xnet_eq_del_progress(eq, &domain->progress);
+		if (ret)
+			break;
```
shefty, line 152: There's not really any hope of recovery from an error here.  I think we add FI_WARN() to print a warning and just continue.

prov/net/src/xnet_domain.c
```diff
@@ -141,6 +163,10 @@ static int xnet_domain_close(fid_t fid)
 	domain = container_of(fid, struct xnet_domain,
 			      util_domain.domain_fid.fid);
 
+	ret = xnet_del_wait_eq_list(domain);
```
shefty, line 166: I guess it would align with the open path to wrap this call
```
if (!fabric->progress.auto_progress) {
    ret = xnet_del_wait_eq_list(...)
}
```
Though I think it's harmless to always call it.  If the progress epoll fd isn't attached to the wait epoll fd, the underlying code handles it.

shefty (commented):
2 minor quibbles.  I think this allows us to remove the progress thread.

I'm not certain about non-Linux platforms.  Even on those platforms, the net provider will request OFI_DYNPOLL_EPOLL, and the underlying abstraction will convert to poll.  We need to run these changes on another OS.

What I'm thinking is if we had a constant like 'have_epoll', that could be included in the check to start the progress threads.  Adding that would be easy.  We could use the HAVE_EPOLL define instead, though I'm hoping a compiler would be smart enough to always take the correct branch based on a const value.

prov/net/src/xnet_domain.c
sydidelot: Fixed. Thanks!

prov/net/src/xnet_domain.c
sydidelot: Fixed. Thanks!

sydidelot:
I updated the PR with the following changes:
- I realized that I forgot to update the function `xnet_cntr_open()`, which was still starting a progress thread.
- I renamed the function `xnet_cq_monitor_progress()` to `xnet_cq_add_progress()` to make it shorter and more consistent with the rest of the code.
- I added a patch that introduces the variable `static const bool ofi_have_epoll` in `include/ofi_epoll.h`. This variable is initialized to true if epoll is supported, false otherwise. Code to start the progress thread was updated accordingly. 

Unfortunately I don't have the time resources to test my PR on non-Linux platforms :-/

shefty (approved):
Can I ask you to rebase?  This will pull in updated CI scripts for the net provider, so we can better verify the changes.

shefty (commented):
I think I need to undo my approval.  The trywait function may not be correct here.  See tcpx_try_func() for reference.  If we have data queued on the socket buffers (bsock), that's the same as the socket being marked as readable.  However, if we block on the socket, the kernel buffers may be empty, and no event will be generated.

I need to look at this more.  But I think we'll need to update xnet_trywait(), since we no longer just start the progress thread.

sydidelot:
@shefty I think you are right and xnet_trywait() needs to be updated. I added a patch in the PR to implement the modifications needed for `xnet_trywait()`. Please take a look and give me your feedback https://github.com/ofiwg/libfabric/pull/8097/commits/d202f2c9eda00b15fd1609faeadf4859aa81d8e2
Also, please note that the PR was rebased to pull in the CI scripts for the net provider.

shefty:
I spent some time thinking through the trywait support, as there are significant differences between the tcp and net providers.  In both cases, any problems should be limited to the buffered sockets.

In the case where there's buffered data to send, the socket should be set to monitor for POLLOUT.  We should be okay there.

In the case where we have buffered data to receive, this occurs only in the case of handling an unexpected message.  The unexp_msg_list and/or unexp_tag_list will be non-empty in this situation.  Checking those lists should be sufficient.  But this brings up the question of whether it's okay to block on the CQ.  The app needs to post a receive to drive progress.  Failing the trywait call here only helps if the app uses that failure to post a receive, and that seems unlikely.  It seems more likely that we're either dealing with a multi-threaded app, where another thread will post the receive; or the app is broken.

I'm leaning toward not needing the trywait changes, unless I can come up with a situation where it would hang.  The tcp provider does not have the unexpected lists, so doesn't try to drive progress when posting a receive.  That's why it needs a trywait implementation.

ooststep:
I don't believe the AWS failures are related to these changes, it doesn't look like AWS tests the net provider and the failure there looks like it's in efa tests.

the Intel CI failures are below.  the `fi_recv_cancel` error is likely not an issue with your changes, I'm not sure about the peek test, yet

```
07:39:17  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8097/9/reg/bin/runfabtests.sh: line 531: 1616426 Killed                  ${SERVER_CMD} "${EXPORT_ENV} $s_cmd" &> $s_outp
07:39:17  - name:   fi_rdm_tagged_peek -p "net"
07:39:17    timestamp: 20221018-123917+0000
07:39:17    result: Fail
07:39:17    time:   1
07:39:17    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8097/9/reg/bin/fi_rdm_tagged_peek -p "net"   -s ci7-eth2
07:39:17    server_stdout: |
07:39:17    client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8097/9/reg/bin/fi_rdm_tagged_peek -p "net"   -s ci8-eth2 ci7-eth2
07:39:17    client_stdout: |
07:39:17      Searching for a bad msg
07:39:17      Searching for a bad msg with claim
07:39:17      Searching for first msg
07:39:17      Receiving first msg
07:39:17      Searching for second msg to claim
07:39:17      Receiving second msg
07:39:17      Searching for third msg to peek and discard
07:39:17      FI_PEEK | FI_DISCARD(): functional/rdm_tagged_peek.c:160, ret=-265 (Truncation error)

07:44:24  - name:   fi_recv_cancel -e rdm -V -p "net"
07:44:24    timestamp: 20221018-124418+0000
07:44:24    result: Fail
07:44:24    time:   301
07:44:24    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8097/9/reg/bin/fi_recv_cancel -e rdm -V -p "net"   -s ci7-eth2
07:44:24    server_stdout: |
07:44:24      fi_endpoint(): common/shared.c:745, ret=-98 (Address already in use)
07:44:24    client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8097/9/reg/bin/fi_recv_cancel -e rdm -V -p "net"   -s ci8-eth2 ci7-eth2
07:44:24    client_stdout: |
```

shefty:
@sydidelot - I think you may have pushed the wrong branch to this PR or dropped the changes you actually wanted to kee.

sydidelot:
@shefty I have fixed the PR. Thanks for spotting this.
According to your last comment, I removed the patch that was changing `xnet_trywait()`.

shefty (approved):
approving ... again

shefty:
thanks!

PR #8100, opened 2022-10-12
prov/efa: Fix a bug in rxr_pkt_req_max_header_size

shijin-aws:
The macro for the raw addr opt header is RXR_REQ_OPT_RAW_ADDR_HDR, not RXR_REQ_OPT_RAW_ADDR_HDR_SIZE


wzamazon (approved):

shijin-aws:
This bug was introduced since 1.14, need to backport to branches until v1.14.x

PR #8108, opened 2022-10-13
contrib/intel/jenkins: add CI support for net provider

ooststep:
The intent is to match the current TCP validations and to eventually merge the two.
It is understood that this will increase runtimes.

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -31,6 +31,7 @@ env:
     --enable-rxm
     --enable-shm
     --enable-tcp
+    --enable-net
     --enable-udp
     --enable-usnic
     --enable-verbs=$PWD/rdma-core/build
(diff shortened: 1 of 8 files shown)


ooststep:
@zachdworkin @Juee14Desai can you please take a look, not sure why it won't let me add you as reviewers.

aingerson (approved):

shefty:
Has a bunch of CI failures

shefty:
This shows up in a couple MPI failures:
```
MPIDI_OFI_mpi_init_hook(1552): OFI fi_getinfo() failed (ofi_init.c:1552:MPIDI_OFI_mpi_init_hook:No data available)
```
We'll need to run the MPI's manually to figure out what their requirements are and how net differs from rxm.

shefty:
If we have to drop MPI testing for now, so that we can start getting some CI coverage, that would be okay.  We can work to add MPI testing as a follow on.

shefty:
@ooststep - this is good to go for now, correct?

zachdworkin (approved):

PR #8120, opened 2022-10-17
[v1.15.x] prov/efa: flush MR cache when fork() is called and fork support is ON

wzamazon:
When fork support is on, register memory regions was not copied to child processes's memory space.

With MR cache turned on, the registered memory regions remain registered after EFA finished send data from it.

As a result, when both fork support and MR cache are turned on, child processes cannot access some memory they need.

This patch addresses the issue by flush the MR cache when fork is called. Flushing MR cache will cause any memory region that is not actively being used (use_cnt == 0) to be de-registered, therefore they will be copied to child process's memory space

(cherry-picked from 5242e5bf5)


shijin-aws (approved):

wzamazon:
bot:aws:retest

PR #8122, opened 2022-10-17
core/peer: Add missing FI_CLASS_PEER for AV and AV_SET

shefty:
Add FI_CLASS_PEER_AV and FI_CLASS_PEER_AV_SET needed by new peer structures.

@j-xiong - FYI

### DIFF
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -167,11 +167,8 @@ typedef struct fid *fid_t;
 
 #define FI_PEER_TRANSFER	(1ULL << 36)
 #define FI_PEER_DOMAIN		(1ULL << 38)
-#define FI_PEER_AV		(1ULL << 39)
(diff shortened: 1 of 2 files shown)


shefty:
@aingerson @amirshehataornl -- This PR includes a change to replace the FI_PEER_SRX and FI_PEER_CQ flags with a single FI_PEER flag.  It also drops the new FI_PEER_xxx flags recently added.

shefty:
Windows error:
```
name:   rdm_tagged_peek -p sockets
  timestamp: Tue 10/18/2022  2:45:42.17
  result: Fail
  time:   183
  server_cmd: C:\projects\libfabric\fabtests\x64\Release-v141\rdm_tagged_peek -p sockets  -s 127.0.0.1
  server_stdout: 

```
which is curious because that test was recently updated and should be less likely to fail now.

PR #8131, opened 2022-10-19
core: Add a config option "--with-lttng" to enable LTTNG tracing capability in EFA.

jtamzn:
This PR added a config option to enable LTTNG tracing capability for tracing EFA OPs.

Signed-off-by: Tang, Jingyin <jytang@amazon.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -695,6 +695,31 @@ AC_ARG_ENABLE([memhooks-monitor],
 AC_DEFINE_UNQUOTED(ENABLE_MEMHOOKS_MONITOR, [$enable_memhooks],
 	[Define to 1 to enable memhooks memory monitor])
 
+dnl Check for LTTNG tracking capability
+have_lttng=0
+AC_ARG_WITH([lttng],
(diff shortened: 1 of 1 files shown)


wzamazon (approved):

shefty (approved):

shijin-aws:
Do we agree that this option should be global or only a efa specific one? If it's an efa specific one, should it be in `prov/efa/configure.m4` ?

wzamazon:
> Do we agree that this option should be global or only a efa specific one? 

I think this should be global one. I do not think it will hurt any provider that does not use it.

shijin-aws (approved):

wzamazon:
@aingerson @shefty 

Can you help understand "continuous-integration/jenkins/pr-head" failure?  :)

Juee14Desai:
Hi @wzamazon, the Intel CI failure is unrelated to the change and we are trying to fix it. It can be ignored.

wzamazon:
@Juee14Desai Thank you very much! I will merge the change then.

PR #8134, opened 2022-10-19
prov/net: Fixup comment above xnet_trywait

shefty:
We no longer start the progress thread, but trywait is okay as-is. Clarify why.


shefty:
update to comment only - skipping CI

PR #8139, opened 2022-10-20
contrib/intel: disable ze shm testing

aingerson:
The node configuration is triggering too many failures. We are also going  to be moving to a different hardware for testing. Disable until that transfer is done.


shefty:
How do I insert flowers, fireworks, and a clapping emoji?

PR #8144, opened 2022-10-21
configure: Test all debug flags

raffenet:
NVHPC compilers do not support -Wno-missing-field-initializers. To avoid issues down the road, test all the debug flags before adding to CFLAGS.

See pmodels/mpich#6221 for user report.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -131,10 +131,10 @@ AC_ARG_ENABLE([debug],
 
 AS_IF([test x"$enable_debug" != x"no"],
       [dbg=1
-       # See if all the flags in $debug_c_other_flags work
+       # See if the flags in $debug_c_warn_flags and $debug_c_other_flags work
        good_flags=
        CFLAGS_save="$CFLAGS"
-       for flag in $debug_c_other_flags; do
+       for flag in $debug_c_warn_flags $debug_c_other_flags; do
            AC_MSG_CHECKING([to see if compiler supports $flag])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -43,8 +43,8 @@ AM_CONDITIONAL([LINUX], [test "x$linux" = "x1"])
 AM_CONDITIONAL([FREEBSD], [test "x$freebsd" = "x1"])
 
 base_c_warn_flags="-Wall -Wundef -Wpointer-arith"
-debug_c_warn_flags="-Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers"
-debug_c_other_flags="-fstack-protector-strong"
+debug_c_warn_flags="-Wextra -Wno-unused-parameter -Wno-sign-compare"
+debug_c_other_flags="-fstack-protector-strong -Wno-missing-field-initializers"
```
raffenet, line 47: Maybe the real solution is to test all the warn flags the same as is done for the other flags. Just let me know what is preferred and I'll update.

shefty: I'm okay with testing all the flags.  That should avoid this problem in the future.

raffenet: OK, updated to test all the debug flags.

shefty:
sign-off line is missing

raffenet:
> sign-off line is missing

just fixed.

PR #8146, opened 2022-10-21
ofi_net: add io_uring support and update bsock data structure

sydidelot:
This PR is an attempt at adding io_uring support to the bsock/byteq datastructures.
To do so, this PR adds a new interface named ofi_sockapi that aims at providing a common interface between io_uring and the traditional sockets.

### DIFF
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -315,7 +315,7 @@ static inline size_t ofi_bsock_tosend(struct ofi_bsock *bsock)
 }
 
 ssize_t ofi_bsock_flush(struct ofi_bsock *bsock);
-/* For sends started asynchronously, the return value will be -EINPROGRESS,
+/* For sends started asynchronously, the return value will be -EINPROGRESS_ASYNC,
  * and len will be set to the number of bytes that were queued.
  */
 ssize_t ofi_bsock_send(struct ofi_bsock *bsock, const void *buf, size_t *len);
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -44,6 +44,10 @@
 #include <netinet/in.h>
 #include <ifaddrs.h>
 
+#ifdef HAVE_LIBURING
+#include <liburing.h>
+#endif
+
 #include <ofi_osd.h>
 #include <ofi_list.h>
 
@@ -133,7 +137,16 @@ ssize_t ofi_discard_socket(SOCKET sock, size_t len);
 /*
  * Socket API
  */
+#ifdef HAVE_LIBURING
+typedef struct io_uring ofi_io_uring_t;
+#else
+typedef int ofi_io_uring_t;
+#endif
+
 struct ofi_sockapi {
+	ofi_io_uring_t *tx_io_uring;
+	ofi_io_uring_t *rx_io_uring;
+
 	ssize_t (*send)(struct ofi_sockapi *sockapi, SOCKET sock, const void *buf,
 			size_t len, int flags, void *ctx);
 	ssize_t (*sendv)(struct ofi_sockapi *sockapi, SOCKET sock,
@@ -180,6 +193,46 @@ ofi_sockapi_recvv_socket(struct ofi_sockapi *sockapi, SOCKET sock,
 	return ofi_recvv_socket(sock, iov, cnt, flags);
 }
 
+#ifdef HAVE_LIBURING
+ssize_t ofi_sockapi_send_uring(struct ofi_sockapi *sockapi, SOCKET sock,
+			       const void *buf, size_t len, int flags, void *ctx);
+ssize_t ofi_sockapi_sendv_uring(struct ofi_sockapi *sockapi, SOCKET sock,
+				const struct iovec *iov, size_t cnt, int flags,
+				void *ctx);
+ssize_t ofi_sockapi_recv_uring(struct ofi_sockapi *sockapi, SOCKET sock,
+			       void *buf, size_t len, int flags, void *ctx);
+ssize_t ofi_sockapi_recvv_uring(struct ofi_sockapi *sockapi, SOCKET sock,
+				struct iovec *iov, size_t cnt, int flags,
+				void *ctx);
+#else
+static inline ssize_t
+ofi_sockapi_send_uring(struct ofi_sockapi *sockapi, SOCKET sock, const void *buf,
+		       size_t len, int flags, void *ctx)
+{
+	return -FI_ENOSYS;
+}
+
+static inline ssize_t
+ofi_sockapi_sendv_uring(struct ofi_sockapi *sockapi, SOCKET sock,
+			const struct iovec *iov, size_t cnt, int flags, void *ctx)
+{
+	return -FI_ENOSYS;
+}
+
+static inline ssize_t
+ofi_sockapi_recv_uring(struct ofi_sockapi *sockapi, SOCKET sock, void *buf,
+		       size_t len, int flags, void *ctx)
+{
+	return -FI_ENOSYS;
+}
+
+static inline ssize_t
+ofi_sockapi_recvv_uring(struct ofi_sockapi *sockapi, SOCKET sock,
+			struct iovec *iov, size_t cnt, int flags, void *ctx)
+{
+	return -FI_ENOSYS;
+}
+#endif
 
 /*
  * Byte queue - streaming socket staging buffer
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -101,6 +101,11 @@ static inline uint64_t ntohll(uint64_t x) { return x; }
 #endif
 
 
+enum {
+	OFI_EINPROGRESS_ASYNC = 512,	/* Async sockets */
+	OFI_EINPROGRESS_URING = 513,	/* io_uring */
+};
+
 static inline int ofi_recvall_socket(SOCKET sock, void *buf, size_t len)
 {
 	ssize_t ret;
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -278,19 +278,12 @@ struct ofi_bsock {
 	uint32_t done_index;
 };
 
-static struct ofi_sockapi ofi_bsock_sockapi =
-{
-    .send = ofi_sockapi_send_socket,
-    .sendv = ofi_sockapi_sendv_socket,
-    .recv = ofi_sockapi_recv_socket,
-    .recvv = ofi_sockapi_recvv_socket,
-};
-
 static inline void
-ofi_bsock_init(struct ofi_bsock *bsock, ssize_t sbuf_size, ssize_t rbuf_size)
+ofi_bsock_init(struct ofi_bsock *bsock, struct ofi_sockapi *sockapi,
+	       ssize_t sbuf_size, ssize_t rbuf_size)
 {
 	bsock->sock = INVALID_SOCKET;
-	bsock->sockapi = &ofi_bsock_sockapi;
+	bsock->sockapi = sockapi;
 	ofi_byteq_init(&bsock->sq, sbuf_size);
 	ofi_byteq_init(&bsock->rq, rbuf_size);
 	bsock->zerocopy_size = SIZE_MAX;
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -126,6 +126,57 @@ static inline int ofi_sendall_socket(SOCKET sock, const void *buf, size_t len)
 ssize_t ofi_discard_socket(SOCKET sock, size_t len);
 
 /*
+ * Socket API
+ */
+struct ofi_sockapi {
+	ssize_t (*send)(struct ofi_sockapi *sockapi, SOCKET sock, const void *buf,
+			size_t len, int flags, void *ctx);
+	ssize_t (*sendv)(struct ofi_sockapi *sockapi, SOCKET sock,
+			 const struct iovec *iov, size_t cnt, int flags, void *ctx);
+	ssize_t (*recv)(struct ofi_sockapi *sockapi, SOCKET sock, void *buf,
+			size_t len, int flags, void *ctx);
+	ssize_t (*recvv)(struct ofi_sockapi *sockapi, SOCKET sock,
+			 struct iovec *iov, size_t cnt, int flags, void *ctx);
+};
+
+static inline ssize_t
+ofi_sockapi_send_socket(struct ofi_sockapi *sockapi, SOCKET sock, const void *buf,
+			size_t len, int flags, void *ctx)
+{
+	OFI_UNUSED(sockapi);
+	OFI_UNUSED(ctx);
+	return ofi_send_socket(sock, buf, len, flags);
+}
+
+static inline ssize_t
+ofi_sockapi_sendv_socket(struct ofi_sockapi *sockapi, SOCKET sock,
+			 const struct iovec *iov, size_t cnt, int flags, void *ctx)
+{
+	OFI_UNUSED(sockapi);
+	OFI_UNUSED(ctx);
+	return ofi_sendv_socket(sock, iov, cnt, flags);
+}
+
+static inline ssize_t
+ofi_sockapi_recv_socket(struct ofi_sockapi *sockapi, SOCKET sock, void *buf,
+			size_t len, int flags, void *ctx)
+{
+	OFI_UNUSED(sockapi);
+	OFI_UNUSED(ctx);
+	return ofi_recv_socket(sock, buf, len, flags);
+}
+
+static inline ssize_t
+ofi_sockapi_recvv_socket(struct ofi_sockapi *sockapi, SOCKET sock,
+			 struct iovec *iov, size_t cnt, int flags, void *ctx)
+{
+	OFI_UNUSED(sockapi);
+	OFI_UNUSED(ctx);
+	return ofi_recvv_socket(sock, iov, cnt, flags);
+}
+
+
+/*
  * Byte queue - streaming socket staging buffer
  */
 enum {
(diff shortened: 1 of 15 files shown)


sydidelot:
@shefty This PR tries to follow the design you posted in this comment: https://github.com/ofiwg/libfabric/pull/6761#issuecomment-1286436836
One problem with the PR is that I need to store the pointer to the io_uring somewhere and pass it as an argument to all the IO operations. So far, this pointer is stored in ofi_sockapi, but I'm not very happy with this design. One concern is that the bsock will need to define 2 ofi_sockapi (1 for each direction) and each of them will have it's own list of socket/io_uring ops.
I would appreciate your feedback on that.

src/common.c
```diff
@@ -1283,7 +1276,7 @@ ssize_t ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t len)
 		return bytes;
 	}
 
-	ret = ofi_recv_socket(bsock->sock, buf, len, MSG_NOSIGNAL);
+	ret = ofi_recvv_socket(bsock->sock, buf, len, MSG_NOSIGNAL);
```
shefty: This has mismatched parameters -- buf and len need to be converted to an iov.

include/ofi_net.h
```diff
@@ -158,6 +158,57 @@ ofi_recvv_socket(SOCKET sock, struct iovec *iov, size_t cnt, int flags)
...
+struct ofi_sockapi {
+	ssize_t (*send)(struct ofi_sockapi *sockapi, SOCKET sock, const void *buf,
+			size_t len, int flags, void *ctx);
+	ssize_t (*sendv)(struct ofi_sockapi *sockapi, SOCKET sock,
+			 const struct iovec *iov, size_t cnt, int flags, void *ctx);
+	ssize_t (*recv)(struct ofi_sockapi *sockapi, SOCKET sock, void *buf,
+			size_t len, int flags, void *ctx);
+	ssize_t (*recvv)(struct ofi_sockapi *sockapi, SOCKET sock,
+			 struct iovec *iov, size_t cnt, int flags, void *ctx);
+};
```
shefty, line 185: This is exactly along the lines of what I was thinking, though I was struggling to think of what to call it.

prov/net/src/xnet_ep.c
```diff
@@ -43,6 +43,13 @@ extern struct fi_ops_rma xnet_rma_ops;
 extern struct fi_ops_msg xnet_msg_ops;
 extern struct fi_ops_tagged xnet_tagged_ops;
 
+static struct ofi_sockapi xnet_sockapi =
+{
+    .send = ofi_sockapi_send_socket,
+    .sendv = ofi_sockapi_sendv_socket,
+    .recv = ofi_sockapi_recv_socket,
+    .recvv = ofi_sockapi_recvv_socket,
+};
```
shefty, line 52: nit: use tabs

prov/tcp/src/tcpx_ep.c
```diff
@@ -43,6 +43,13 @@ extern struct fi_ops_rma tcpx_rma_ops;
 extern struct fi_ops_msg tcpx_msg_ops;
 extern struct fi_ops_tagged tcpx_tagged_ops;
 
+static struct ofi_sockapi tcpx_sockapi =
+{
+    .send = ofi_sockapi_send_socket,
+    .sendv = ofi_sockapi_sendv_socket,
+    .recv = ofi_sockapi_recv_socket,
+    .recvv = ofi_sockapi_recvv_socket,
+};
```
shefty, line 52: tabs

shefty (commented):
Thanks!  If we can get this flow to work once adding in the completion handling, I think it'll end up looking fairly clean.  I see what you mean about passing through the ofi_sockapi to the byteq.  I need to think about how to handle that.

I like the idea of having separate rx and tx io_urings.  Couldn't you include 2 io_urings in the ofi_sockapi itself and have the send/recv calls simply pick the right one?

include/ofi_net.h
```diff
@@ -239,6 +379,7 @@ static inline ssize_t ofi_byteq_send(struct ofi_byteq *byteq, SOCKET sock)
  */
 struct ofi_bsock {
 	SOCKET sock;
+	struct ofi_sockapi *sockapi;
```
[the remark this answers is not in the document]
shefty: Looking at this more, I think this is the right place for sockapi, assuming we go this route.  It should be near where we declare the socket.  The problem is really in the byteq layering.  I'm going to try to rework how bsock calls it, and see if I can't move all of the socket calls up to the bsock layer.  If you want to keep what you have in this PR for now and see what the completion path looks like, we can see about restructuring the socket calls later.

src/common.c
sydidelot: Oops, thanks for catching that! This line was not supposed to be changed.

sydidelot:
Your suggestion to have 2 io_urings in the ofi_sockapi was the solution! Thanks!

include/ofi_net.h
sydidelot: Awesome!

prov/net/src/xnet_ep.c
sydidelot: Fixed.

prov/tcp/src/tcpx_ep.c
sydidelot: Fixed

shefty:
What size can the urings be?

include/ofi_net.h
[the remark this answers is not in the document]
sydidelot: We can hold of on the PR until you restructure the bsock interface, I'm fine with it! I can continue my work on the net provider based on what I have in the PR.

sydidelot:
I didn't find a limit in liburing, but there is apparently a hard limit of 32768 entries per ring in the kernel: 
https://elixir.bootlin.com/linux/latest/source/io_uring/io_uring.c#L3245

include/ofi_net.h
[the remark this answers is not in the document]
shefty: See #8150 -- the changes were simpler than I was expecting.

sydidelot:
@shefty I rebased my PR to include your patch for the byteq. I have successfully completed the implementation for tx ops, which I will push as a separate PR in a few minutes.
I looked at the code for rx ops but the changes are not trivial. We first need to modify the bsock API so that the caller knows how much data was read for the byteq (similar to what you did for the zerocopy send). Then, we will have to restructure the rx code to make it easier for the io_uring progress function to progress/complete an outstanding cur_rx entry.

shefty:
An easier change would be to replace the recv() call with io_uring, but still check for POLLIN events.  That may give us equivalent performance to what we have now.

I don't think io_uring supports MSG_WAITALL, if I'm remembering recent kernel discussions.  If it did, then we could use that for larger transfers, which would remove the need to repost buffers.

I'd still like to replace POLLIN events with pre-posting receive buffers, but if we can stage that, that would work for me.

sydidelot:
I couldn't create a new PR that depends on this current one (stacked PRs).
@shefty could you please take a look at the following PR in my forked repository? https://github.com/sydidelot/libfabric/pull/1
It implements all the code for tx ops using io_uring in the net provider (it needs to be applied on top of #8146).
I will create a new PR in libfabric upstream repo as soon as #8146 is merged.
Thank you!

sydidelot:
I don't know if MSG_WAITALL is supported. I will do a bit of research on this topic. 
I found that io_uring supports a very interesting feature called multishot recv/recvmsg: https://lore.kernel.org/lkml/165662117486.56180.16557557417345255423.b4-ty@kernel.dk/T/
The problem is that this feature is only a few months old and probably requires a very recent Linux kernel.

shefty:
To create a new PR with additional changes, you basically need to create a branch.  The new PR will include all of the changes from the other PR, but it's the best I think you can do.

include/rdma/fi_errno.h
```diff
@@ -194,6 +194,8 @@ enum {
 	FI_ENOAV	 = 267, /* Missing or unavailable address vector */
 	FI_EOVERRUN	 = 268, /* Queue has been overrun */
 	FI_ENORX	 = 269, /* Receiver not ready, no receive buffers available */
+	FI_EIOURING_PREP = 270, /* io_uring operation prepared, waiting for submission */
+	FI_EIOURING_FULL = 271, /* io_uring full */
```
shefty, line 198: I'm reviewing patch by patch.  Do these really need to be exposed to the libfabric user?

src/common.c
```diff
@@ -1154,6 +1154,9 @@ ssize_t ofi_bsock_flush(struct ofi_bsock *bsock)
 				   &bsock->sq.data[bsock->sq.head],
 				   avail, MSG_NOSIGNAL, bsock);
 	if (ret < 0) {
+		if (ret == -FI_EIOURING_PREP)
```
shefty, line 1157: Is this equivalent to EINPROGRESS?  That is, an async operation was started?

shefty (commented):
Looks good -- a couple questions on the return value

include/rdma/fi_errno.h
shefty: Is EIOURING_FULL equivalent to EBUSY or EAGAIN?

sydidelot: `FI_EIOURING_PREP` is _not_ supposed to be forwarded to the libfabric user: it can remain private to the net provider. I see that `efa` defines its own set of errno codes in `prov/efa/src/efa_errno.h`. Do you want me to do the same for io_uring in the net provider?

As for `FI_EIOURING_FULL`, we can return `-FI_EOVERFLOW` and let the provider decide what's the best strategy: retry or fail. The `net` provider will have its own control flow mechanism to ensure that it never overflows the rings, so this error code is supposed to never happen. I think we can treat it as a critical error and disable the EP if it ever happens (likely because of a bug in libfabric or io_uring/liburing).

src/common.c
sydidelot: Yes, it's strictly equivalent to `EINPROGRESS`. I have used a different error code because otherwise the caller cannot differentiate the underlying implementation (sockets vs. io_uring) and cannot decide what to do next.

shefty: EINPROGRESS currently implies that we're using async sockets.  My preference is either to use this return code in all cases and use another method to determine how to respond to the async request; OR define 2 provider specific error codes, for both io_uring and async sockets, rather than using a standard code in one case and not the other (e.g. EINPROGRESS_ASYNC vs EINPROGRESS_URING? -- not that I'm thrilled with either name).

We may be able to setup an async_handler() based on which type of async method we're using.  Using this method may allow just using EINPROGRESS and responding to that return accordingly.  But having separate return codes may make the code easier to follow, since we wouldn't need to follow function pointers.

sydidelot: I like the suggestion of using 2  error codes for both io_uring and async sockets. 
One problem is that the functions used by `sockapi` are defined in common code (`src/common.c` and `src/iouring.c`): I'm not sure how we can reference a provider specific error code in the common code.

shefty: Hmm... efa mostly starts their error codes above 4k.  They have some lower values, but the header indicates that they're trying to avoid rdma-core error codes.  I guess we could add internal error codes in ofi.h or wherever sockapi lands starting at 512.  We can always change internal error codes, but that leaves plenty of space after FI_ERRNO_MAX.

sydidelot:
I updated the PR with the new error codes. The changes include:
- Removed `FI_EIOURING_FULL` and use `FI_EOVERFLOW` instead. 
- Defined `EINPROGRESS_ASYNC` and `EINPROGRESS_URING` in `ofi_net.h` and updated the code accordingly. 
- Renamed a few sockapi functions related to io_uring: `ofi_sockapi_xx_iouring` -> `ofi_sockapi_xx_uring`

sydidelot:
I have reworked the patch that introduced `ofi_{sendv,recvv}_socket()`. I realized that the same functions already exist in `src/osx/osd.c` and `src/windows/osd.c` (static functions), which was causing a conflict with the new function I introduced.
I finally decided to expose the functions `ofi_{sendv,recvv}_socket()` in the osd interface. Please let me know if this is the right approach.

src/common.c
```diff
@@ -1154,6 +1154,9 @@ ssize_t ofi_bsock_flush(struct ofi_bsock *bsock)
 				   &bsock->sq.data[bsock->sq.head],
 				   avail, MSG_NOSIGNAL, bsock);
 	if (ret < 0) {
+		if (ret == -FI_EINPROGRESS_URING)
+		    return ret;
+
```
shefty, line 1159: Note: This isn't a problem with this change, but possibly another change that might be needed.

ofi_bsock_flush() is called from ep shutdown() to try forcing any buffered data out the socket.  I don't think there are any guarantees that that data will be sent, so it's mostly to provide a best effort on pushing out any last data.  We may need to modify the ep shutdown() call to use a standard socket send() instead of possibly going through uring.

src/common.c
```diff
@@ -1197,6 +1202,8 @@ ssize_t ofi_bsock_send(struct ofi_bsock *bsock, const void *buf, size_t *len)
 	} else {
 		ret = bsock->sockapi->send(bsock->sockapi, bsock->sock, buf, *len,
 					   MSG_NOSIGNAL, bsock);
+		if (ret == -FI_EINPROGRESS_URING)
+			return ret;
```
shefty, line 1206: If we move this check down to the ```if (ret < 0)``` check just below, it will handle this case and the zerocopy case if that's ever a possibility.

src/common.c
```diff
@@ -1240,6 +1247,8 @@ ssize_t ofi_bsock_sendv(struct ofi_bsock *bsock, const struct iovec *iov,
 	if (*len > bsock->zerocopy_size) {
 		ret = bsock->sockapi->sendv(bsock->sockapi, bsock->sock, iov, cnt,
 					    MSG_NOSIGNAL | OFI_ZEROCOPY, bsock);
+		if (ret == -FI_EINPROGRESS_URING)
+			return ret;
```
shefty, line 1251: this check...

src/common.c
```diff
@@ -1248,6 +1257,8 @@ ssize_t ofi_bsock_sendv(struct ofi_bsock *bsock, const struct iovec *iov,
 	} else {
 		ret = bsock->sockapi->sendv(bsock->sockapi, bsock->sock, iov, cnt,
 					    MSG_NOSIGNAL, bsock);
+		if (ret == -FI_EINPROGRESS_URING)
+			return ret;
```
shefty, line 1261: and this check...

src/common.c
```diff
@@ -1248,6 +1257,8 @@ ssize_t ofi_bsock_sendv(struct ofi_bsock *bsock, const struct iovec *iov,
...
 	}
 	if (ret < 0) {
```
shefty, line 1263: can drop to here as a common exit check

include/ofi_net.h
```diff
@@ -101,6 +105,11 @@ static inline uint64_t ntohll(uint64_t x) { return x; }
 #endif
 
 
+enum {
+	FI_EINPROGRESS_ASYNC = 512,	/* Async sockets */
+	FI_EINPROGRESS_URING = 513,	/* io_uring */
```
shefty, line 110: Internal definitions use the OFI prefix, so we reserve FI for the external API only.  (Yes, there are still some places where we haven't renamed things from fi to ofi, but that's the direction).

shefty (approved):
There's a couple of places where I marked that we'll need to see if further changes to the code are needed.  Other than that, I only have a couple minor comments, which could either be addressed in subsequent changes or left as-is.

This set looks good to me that I'd like to merge it and build other changes over them.  Thanks!

src/common.c
sydidelot: Done, thanks

src/common.c
sydidelot: Done, thanks

src/common.c
sydidelot: Done, thanks

src/common.c
sydidelot: Done, thanks

include/ofi_net.h
sydidelot: I changed the definitions to use the OFI prefix. Thanks!

sydidelot:
appveyor failed for 1 test but I don't see what's the actual error in the logs: https://ci.appveyor.com/project/ofiwg/libfabric/builds/45185206/job/2l79iij29et36fr8

src/common.c
sydidelot: Thanks. I will address your comment in the next PR.

src/common.c
```diff
@@ -1189,6 +1192,8 @@ ssize_t ofi_bsock_send(struct ofi_bsock *bsock, const void *buf, size_t *len)
 	if (*len > bsock->zerocopy_size) {
 		ret = bsock->sockapi->send(bsock->sockapi, bsock->sock, buf, *len,
 					   MSG_NOSIGNAL | OFI_ZEROCOPY, bsock);
+		if (ret == -FI_EINPROGRESS_URING)
+			return ret;
```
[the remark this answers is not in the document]
sydidelot: Thanks. I will disable zerocopy with uring in the next PR.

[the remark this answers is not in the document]
sydidelot: FYI, zerocopy send feature for io_uring was introduced in Linux version 6.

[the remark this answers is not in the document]
shefty: Thanks - we'll need to add some check for this.  This isn't critical, since the user can toggle these options independently and both are off by default.

shefty:
Thanks!  I chose to ignore the appveyor error, since it was only 1 build that failed out of maybe 8, and I have no idea why the build stopped.

PR #8150, opened 2022-10-21
core/net: Restructure interface between bsock and byteq

shefty:
The byte queue abstraction is managing what is basically a ring buffer.  However, that abstraction is broken by 2 calls (send and receive) where we pass in a socket.  Replace those calls and keep all socket access in the bsock abstraction.

This requires adding 2 new helper functions to add and consume data from the byteq.  This keeps the byteq abstraction generic, so that it's not tied to a socket, which will be useful in follow on patches to support io_uring.

(The byteq and ringbuf abstractions may be mergable, but that would need to be a separate change.)

### DIFF
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -166,6 +166,19 @@ static inline size_t ofi_byteq_writeable(struct ofi_byteq *byteq)
 	return byteq->size - byteq->tail;
 }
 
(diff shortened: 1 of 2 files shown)


sydidelot (approved):
Looks good to me, thanks!

PR #8151, opened 2022-10-22
prov/sockets: fix segfault for conn->av_index in sock_av_remove

wenduwan:
Check `conn != SOCK_CM_CONN_IN_PROGRESS` in addition to `conn != NULL`, because `conn` can be partially initialized in `sock_ep_get_conn`. Acessing `SOCK_CM_CONN_IN_PROGRESS->av_index` causes a segfault.


wenduwan:
bot:aws:retest

shefty:
Thanks - is there a reason you're using the socket provider instead of tcp or net?

wzamazon:
@shefty Unfortunately, we still have some customers that are using sockets. We are pushing them to migrate to tcp, but that will take some time.

wzamazon (approved):

PR #8157, opened 2022-10-24
contrib/intel/jenkins disable ompi osu tests

zachdworkin:
OMPI OSU was accidentally re-enabled in PR-8108's change of the OSUtests class's execute condition while a known hang issue still occurs. Disabling it again until we fix the problem.

The problem: ompi 4.1.0 with osu tests 5.7 fails verbs-rxm on osu_allgatherv byte transfer 128 with a hang.
I suspect it is a race condition because sometimes the test passes and sometimes not. It will require further debugging.

### DIFF
--- a/contrib/intel/jenkins/tests.py
+++ b/contrib/intel/jenkins/tests.py
@@ -586,14 +586,10 @@ class OSUtests(Test):
 
     @property
     def execute_condn(self):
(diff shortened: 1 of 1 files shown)


zachdworkin:
@shefty I replayed this and it passed the verbs-rxm stage as expected.

shefty:
thanks!

PR #8137, opened 2022-10-20
Improve rxm send/recv performance with HMEM in use

j-xiong:
Call ofi_hmem_host_register() on the buffers to be used as bounce buffer. This can greatly reduce the overhead of HMEM copy.

PR #8160, opened 2022-10-25
Revert buffer pool flag changes

j-xiong:
It turns out rxm already uses attrs.alloc_fn/free_fn to handle host registration. Out of the three commits from PR #8137, revert the two commits related to the new flag and keep only the hmem_ze change.


shefty:
Are these related to the shm failures?

aingerson:
@shefty shm failures are a race condition introduced in the recent peer implementation that was merged. I'm looking into it and will fix.

PR #8161, opened 2022-10-25
contrib/intel/jenkins: Update jenkins workspace & globalize py_scripts variable

zachdworkin:
Make jenkins checkout to a custom directory. This will fix the problem of two subjobs running together.

The problem this fixes is when Job1 is running and gets aborted by Job2's disableConcurrentBuilds option. Job2 aborts Job1 and causes Job1 to skip straight to the post cleanup stage where it deletes all workspace directories it used. The problem with this is that Job2 and Job1 shared the same workspace directory and so when Job1 deletes the current working directory of Job2, Job2 fails. Jenkins then gets stuck and cannot start a Job3 until the shared workspace directory is removed by hand.

The solution is to separate these workspaces into job_name/build_number. This way Job1 will checkout into Job/1 and therefore cleanup Job/1/* and Job2 will checkout into Job/2 and cleanup Job/2/* and the problem of having to remove directories by hand is eliminated.

There is also a patch that moves the py_scripts/contrib/intel/jenkins string into a global variable so that when developers must replay and replace all of the py_scripts/contrib/intel/jenkins with /contrib/intel/jenkins they can change it in one location instead of 25.

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -2,6 +2,7 @@
 properties([disableConcurrentBuilds(abortPrevious: true)])
 def DO_RUN=1
 def TARGET="main"
+def SCRIPT_LOCATION="py_scripts/contrib/intel/jenkins"
 def check_target() {
   echo "CHANGE_TARGET = ${env.CHANGE_TARGET}"
   if (changeRequest()) {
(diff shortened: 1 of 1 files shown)


zachdworkin:
@shefty this is currently replaying with all the jenkinsfile changes. Let it finish before merging please.

zachdworkin:
@shefty jenkins only failed the dsa stage with the known race condition. I verified its doing what I expect it to with the new directories.

PR #8162, opened 2022-10-26
core/getinfo: Reorder default fi_info list

shefty:
First patch is a general cleanup of the filtering functionality.

Second removes the restriction to allow the rxm provider to layer over the net provider. net;ofi_rxm is compatible with tcp;ofi_rxm.  This change will allow merging the net and tcp providers back together.

The last patch adds the ability to reorder the fi_info list separately from strict provider ordering.  This allows reporting the net provider's rdm ep support ahead of net;ofi_rxm.


shefty:
CI failures are known issues being investigated.

PR #8177, opened 2022-10-27
prov/shm: refactor unexpected path to fix multi_recv messaging

### LINKED ISSUE #8166: prov/shm: failing fi_multi_recv with dsa testing
Failing in CI testing:
```
name:   fi_multi_recv -e rdm -p "shm"
17:38:25    timestamp: 20221026-003825+0000
17:38:25    result: Fail
17:38:25    time:   2
17:38:25    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8162/1/reg/bin/fi_multi_recv -e rdm -p "shm"   -s jfcst-pvc3
17:38:25    server_stdout: |
17:38:25      [error] fabtests:functional/multi_recv.c:127: Received 1 completions per buffer, expected 64
17:38:25    client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8162/1/reg/bin/fi_multi_recv -e rdm -p "shm"   -s jfcst-pvc3 jfcst-pvc3
```

aingerson:
Fixes #8166


prov/shm/src/smr_msg.c
```diff
@@ -94,34 +94,80 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+			goto out;
+		}
+	}
+	dlist_insert_tail((struct dlist_entry *) (&owner_entry->peer_entry),
 			  &recv_queue->list);
 	ret = FI_SUCCESS;
 out:
 	ofi_spin_unlock(&srx->lock);
 	return ret;
+
```
shefty, line 170: random extra blank line added

prov/shm/src/smr_msg.c
```diff
@@ -94,34 +94,80 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+	while (dlist_entry) {
+		unexp = container_of(dlist_entry, struct smr_rx_entry, peer_entry);
+		smr_init_rx_entry(unexp, iov, desc, iov_count, addr, context,
+				  tag, flags & (~FI_MULTI_RECV));
+		if (owner_entry) {
+			owner_entry->multi_recv_ref++;
+			unexp->peer_entry.owner_context = owner_entry;
+		}
+
+		if (owner_entry && smr_adjust_multi_recv(srx,
```
shefty, line 136: Do you need a lock around smr_adjust_multi_recv()?

aingerson:
Updated to remove blank line and fix locking. Also refactored the adjust/iov setting code a bit to make it more readable.

prov/shm/src/smr_msg.c
```diff
@@ -94,29 +94,73 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+	while (dlist_entry) {
+		unexp = container_of(dlist_entry, struct smr_rx_entry, peer_entry);
+		smr_init_rx_entry(unexp, iov, desc, iov_count, addr, context,
+				  tag, flags & (~FI_MULTI_RECV));
+		if (owner_entry) {
+			owner_entry->multi_recv_ref++;
+			unexp->peer_entry.owner_context = owner_entry;
+		}
+
+		if (owner_entry) {
```
shefty, line 135: You have back to back if statements that check the same thing.

prov/shm/src/smr_msg.c
```diff
@@ -94,29 +94,73 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
 		uint64_t tag, uint64_t ignore, uint64_t flags,
 		struct smr_queue *recv_queue, struct smr_queue *unexp_queue)
 {
-	struct smr_rx_entry *entry;
-	ssize_t ret = -FI_EAGAIN;
-	fi_addr_t use_addr;
+	struct smr_match_attr match_attr;
+	struct smr_rx_entry *unexp, *owner_entry = NULL;
```
shefty, line 98: This function is hard to follow.  I think adding an *mrecv_entry and using that instead of owner_entry would help.  The code has a lot of checks around owner_entry being non-null, but what it is actually trying to check is if the posted receive is a multi-recv entry.  owner_entry is used for both multi-recv and normal receives, but switches its meaning 3/4th of the way through the function after the while loop.

prov/shm/src/smr_msg.c
```diff
@@ -94,29 +94,73 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+		if (owner_entry) {
+			owner_entry->multi_recv_ref++;
+			unexp->peer_entry.owner_context = owner_entry;
+		}
+
+		if (owner_entry) {
+			if (smr_adjust_multi_recv(srx, &owner_entry->peer_entry,
+						  unexp->peer_entry.size))
+				buf_done = true;
+			iov = owner_entry->iov;
```
shefty, line 139: Don't you need this before you adjust the mrecv buffer?

prov/shm/src/smr_msg.c
```diff
@@ -94,29 +94,73 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+		}
+
+		ofi_spin_unlock(&srx->lock);
+		ret = srx->peer_srx.peer_ops->start_msg(&unexp->peer_entry);
+		if (ret)
+			return ret;
+		else if (!owner_entry || buf_done)
```
shefty, line 146: nit: else isn't needed as if statement above returns

prov/shm/src/smr_msg.c
```diff
@@ -94,29 +94,73 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+			return FI_SUCCESS;
```
shefty, line 147: If done_buf is true, what happened to the original owner_entry?  Was that lost?

aingerson: The owner entry is freed in (free_entry) which is called in the start_msg processing/whenever the message is done so we don't care about freeing it here.

prov/shm/src/smr_msg.c
aingerson: the iov is saved for the next message/for posting if we can't progress any unexpected messages. the owner_entry iov is used to initialize the unexpected rx_entry. some refactoring is obviously needed so I'll smooth this out and repush

aingerson:
@shefty Updated to separate regular and mrecv paths. The unexpected message processing is different enough that it gets really hard to read if we keep them together. It also helps remove that complexity from the tagged path which does not use multi recv. Let me know what you think.

prov/shm/src/smr_msg.c
```diff
@@ -131,10 +198,14 @@ static ssize_t smr_recvmsg(struct fid_ep *ep_fid, const struct fi_msg *msg,
 
 	ep = container_of(ep_fid, struct smr_ep, util_ep.ep_fid.fid);
 
+	flags |= ep->util_ep.rx_msg_flags;
+	if (flags & FI_MULTI_RECV)
+		return smr_generic_mrecv(smr_get_smr_srx(ep),
+					 msg->msg_iov, msg->desc, msg->iov_count, 
+					 msg->addr, msg->context, flags);
```
shefty, line 205: Why not move the FI_MULTI_RECV check info smr_generic_recv, rather than duplicating it before each call to it?

prov/shm/src/smr_msg.c
```diff
@@ -94,30 +94,97 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+					       &match_attr);
+	ofi_spin_unlock(&srx->lock);
+	if (!dlist_entry) {
+		rx_entry = smr_get_recv_entry(srx, iov, desc, iov_count, addr,
+					      context, tag, ignore, flags);
+		if (!rx_entry)
+			return -FI_ENOMEM;
+
+		dlist_insert_tail((struct dlist_entry *) (&rx_entry->peer_entry),
+				  &recv_queue->list);
```
shefty, line 121: Need a lock here.  Maybe also around smr_get_recv_entry()?

prov/shm/src/smr_msg.c
```diff
@@ -94,30 +94,97 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+
+	rx_entry = container_of(dlist_entry, struct smr_rx_entry, peer_entry);
+	smr_init_rx_entry(rx_entry, iov, desc, iov_count, addr, context,
+			  tag, flags);
+
+	ret = srx->peer_srx.peer_ops->start_msg(&rx_entry->peer_entry);
+	if (ret)
+		return ret;
+
+	return FI_SUCCESS;
```
shefty, line 133: Can just return ret and skip the if statement.

prov/shm/src/smr_msg.c
```diff
@@ -94,30 +94,97 @@ static ssize_t smr_generic_recv(struct smr_srx_ctx *srx, const struct iovec *iov
...
+	assert(flags & FI_MULTI_RECV && iov_count == 1);
+	addr = srx->dir_recv ? addr : FI_ADDR_UNSPEC;
+
+	match_attr.id = addr;
+
+	ofi_spin_lock(&srx->lock);
+	mrecv_entry = smr_get_recv_entry(srx, iov, desc, iov_count, addr,
+					 context, 0, 0, flags);
+	mrecv_entry->peer_entry.size = ofi_total_iov_len(iov, iov_count);
+	if (!mrecv_entry) {
```
shefty, line 155: It doesn't do any good to check that mrecv_entry is valid after the call before has already dereferenced it.

prov/shm/src/smr_msg.c
aingerson: I had it like that before but moved it to remove the check from the tagged path which never uses multi recv and previously an app could call tagged + multi recv and it would pass but have issues later. I can just add an assert though instead.

aingerson:
Updated to move mrecv check inside generic_recv and other fixes. @shefty thanks for catching that missing lock. I've added the lock back in for that code but don't love the locking/unlocking flow. couldn't really see a better way to do it. let me know if you have a better idea. (I'm talking about lines 178-190)

shefty (approved):
Looks fine.  Locking is a little odd looking, but easy to follow, and I don't see an obvious way to improve it.

Issue #8163, opened 2022-10-26
prov/net: failing fi_rdm_tagged_peek test


shefty:
CI is failing consistently.  Best failure report may be this one:

```
name:   fi_rdm_tagged_peek -p "net"
16:50:57    timestamp: 20221021-235057+0000
16:50:57    result: Fail
16:50:57    time:   2
16:50:57    server_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8142/4/dbg/bin/fi_rdm_tagged_peek -p "net"   -s ci7-eth2
16:50:57    server_stdout: |
16:50:57    client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8142/4/dbg/bin/fi_rdm_tagged_peek -p "net"   -s ci8-eth2 ci7-eth2
16:50:57    client_stdout: |
16:50:57      FI_PEEK | FI_DISCARD(): functional/rdm_tagged_peek.c:160, ret=-265 (Truncation error)
16:50:57      fi_rdm_tagged_peek: prov/util/src/util_buf.c:256: ofi_bufpool_destroy: Assertion `(pool->attr.flags & OFI_BUFPOOL_NO_TRACK) || !ofi_atomic_get32(&buf_region->use_cnt)' failed.
```
Most reports are along the lines of this:
```
client_cmd: /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8162/1/reg/bin/fi_rdm_tagged_peek -p "net"   -s ci8-eth2 ci7-eth2
18:27:50    client_stdout: |
18:27:50      Searching for a bad msg
18:27:50      Searching for a bad msg with claim
18:27:50      Searching for first msg
18:27:50      Receiving first msg
18:27:50      Searching for second msg to claim
18:27:50      Receiving second msg
18:27:50      Searching for third msg to peek and discard
18:27:50      FI_PEEK | FI_DISCARD(): functional/rdm_tagged_peek.c:160, ret=-265 (Truncation error)
```

shefty:
Debugging, the problem occurs after processing the 3rd message (peek & discard).  It appears to be related to the freeing of the internally allocated buffer

shefty:
Capture of debug output near the crash:
```
Searching for third msg to peek and discard
xfer alloc 0x697910
malloc (0x697910) 0x688a70 1024
free (0x697910) 0x688be8 648
free(): invalid pointer
```
[...]

PR #8183, opened 2022-10-28
prov/net: Fix possible crash handling FI_DISCARD

shefty:
We need to allocate an internal buffer to receive a discarded message.  Once the message has been received, we can free the buffer.  The buffer is stored in the recv_entry iov[0] as iov_base.  The iov_base is freed later.  However, as part of the receive handling, the iov_base can be updated.  This is done to handle partial receives, where we need to wait for additional data to arrive from the peer before receiving more data.

To fix, we need to record the address of the allocation in a separate location, so we can free it later.  Use the existing user_buf pointer for this, which is used in other places to track the start of the buffer.

### DIFF
--- a/prov/net/src/xnet.h
+++ b/prov/net/src/xnet.h
@@ -553,7 +553,7 @@ xnet_free_xfer(struct xnet_progress *progress, struct xnet_xfer_entry *xfer)
 {
 	assert(xnet_progress_locked(progress));
 	if (xfer->ctrl_flags & XNET_FREE_BUF)
-		free(xfer->iov[0].iov_base);
+		free(xfer->user_buf);
 
 	ofi_buf_free(xfer);
 }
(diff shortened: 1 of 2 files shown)


shefty:
Fixes #8163.  Mac OS is forced failure by github and being addressed separately.

ooststep:
ah, much better than reserving index 1 or adding more fields.

I've run the peek test with this code over 10k times to far and no issues yet - prior code failed before 4500 on this hardware.

shefty:
Added 2 additional fixes.  @ooststep - Can you try with these changes?  I think they're likely to address the truncation errors.

ooststep:
> Added 2 additional fixes. @ooststep - Can you try with these changes? I think they're likely to address the truncation errors.

I forgot I started this test and accidentally let it run over the weekend.  by the time my slurm job died it had run over 2 million iterations. I think it's good :P

shefty:
CI failures were in shm and sockets.  Sockets has a separate PR for it.  shm may just be my tree being out of date.

PR #8184, opened 2022-10-28
fabtests/pytest: use pytest-xdist to parallelize tests

wzamazon:
This patch used the pytest-xdist package to run tests in parallel. One change made was that each xdist worker need to have its own oob port for oob address exchange.

### DIFF
--- a/fabtests/pytest/conftest.py
+++ b/fabtests/pytest/conftest.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 
 def get_option_longform(option_name, option_params):
@@ -144,9 +145,15 @@ class CmdlineArgs:
         if host_type == "host":
             return command
 
+        if self.oob_address_exchange:
+            oob_argument = "-E"
+            if "PYTEST_XDIST_WORKER" in os.environ:
+                oob_port = 9228 + int(os.environ["PYTEST_XDIST_WORKER"].replace("gw", ""))
+                oob_argument += "={}".format(oob_port)
+
         if host_type == "server":
             if self.oob_address_exchange:
-                command += " -E"
+                command += " " + oob_argument
             else:
                 command += " -s " + self.server_interface
 
@@ -157,7 +164,7 @@ class CmdlineArgs:
 
         assert host_type == "client"
         if self.oob_address_exchange:
-            command += " -E " + self.server_id
+            command += " " + oob_argument + " " + self.server_id
         else:
             command += " -s " + self.client_interface + " " + self.server_interface
 
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -5,11 +5,11 @@ def efa_run_client_server_test(cmdline_args, executable, iteration_type,
                                completion_type, memory_type, message_size,
                                warmup_iteration_type=None):
     from common import ClientServerTest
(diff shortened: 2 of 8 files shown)


fabtests/pytest/conftest.py
```diff
@@ -144,9 +145,15 @@ def _populate_normal_command(self, command, host_type):
         if host_type == "host":
             return command
 
+        if self.oob_address_exchange:
+            oob_argument = "-E"
+            if "PYTEST_XDIST_WORKER" in os.environ:
+                oob_port = 9228 + int(os.environ["PYTEST_XDIST_WORKER"].replace("gw", ""))
```
shijin-aws, line 151: what is `gw`?

shijin-aws:
I think parallelizing tests make sense for functional test, can we use it for performance test?

fabtests/pytest/conftest.py
wzamazon: Not exactly sure. pytest-xdist set the environment variable value to `gw0`, `gw1`, ...

wzamazon:
> I think parallelizing tests make sense for functional test, can we use it for performance test?

No, we cannot parallelize it for performance test.

If we want to run performance test using `runfabtests.py`, we need to set nworkers` to 1.

But, I am not sure we want to use `runfabtests.py for performance test. I think it would be better to run individual test directory for performance.

shijin-aws (approved):

fabtests/pytest/conftest.py
shijin-aws: It seems there is a `worker_id` fixture for the same usage.

wzamazon:
bot:aws:retest

fabtests/scripts/runfabtests.py
```diff
@@ -187,8 +200,10 @@ def fabtests_args_to_pytest_args(fabtests_args, shared_options):
 
     if fabtests_args.junit_xml:
         pytest_args.append("--junit-xml")
-        pytest_args.append(os.path.abspath(fabtests_args.junit_xml))
-        pytest_args.append("--self-contained-html")
```
shijin-aws, line 191: why this arg is dropped?

shijin-aws: It seems that this should be added to the `fabtests_args.html` path.

fabtests/pytest/efa/efa_common.py
```diff
@@ -5,11 +5,11 @@ def efa_run_client_server_test(cmdline_args, executable, iteration_type,
...
     from common import ClientServerTest
-    # It is observed that cuda tests requires larger time-out limit (~240 secs) to test all
-    # message sizes for libfabric's debug and mem-poisoning builds, on p4d instances.
+    # It is observed that cuda tests requires larger time-out limit to test all
+    # message sizes (especailly when running with multiple workers).
     timeout = None
-    if "cuda" in memory_type and message_size == "all":
-        timeout = 240
+    if "cuda" in memory_type:
+        timeout = max(1000, cmdline_args.timeout)
```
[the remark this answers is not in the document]
wzamazon: it is a number I tested that would work.

fabtests/scripts/runfabtests.py
wzamazon: It was not needed for xml. 

It is already in `fabtests.html` path.

shijin-aws (approved):

PR #8186, opened 2022-10-28
prov/efa: Set use_device_rdma as 1 when efa device id is 0xefa2.

shijin-aws:

### DIFF
--- a/prov/efa/src/rxr/rxr_env.c
+++ b/prov/efa/src/rxr/rxr_env.c
@@ -75,6 +75,8 @@ struct rxr_env rxr_env = {
  */
 void rxr_env_param_get(void)
 {
+	int ret;
+
 	if (getenv("FI_EFA_SHM_MAX_MEDIUM_SIZE")) {
 		fprintf(stderr,
 			"FI_EFA_SHM_MAX_MEDIUM_SIZE env variable detected! The use of this variable has been deprecated and as such execution cannot proceed.\n");
@@ -93,7 +95,13 @@ void rxr_env_param_get(void)
 
 	fi_param_get_int(&rxr_prov, "tx_queue_size", &rxr_env.tx_queue_size);
 	fi_param_get_int(&rxr_prov, "enable_shm_transfer", &rxr_env.enable_shm_transfer);
-	fi_param_get_int(&rxr_prov, "use_device_rdma", &rxr_env.use_device_rdma);
+	ret = fi_param_get_int(&rxr_prov, "use_device_rdma", &rxr_env.use_device_rdma);
+	if (ret != -FI_ENODATA && rxr_env.use_device_rdma &&
+	    !(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
+		fprintf(stderr, "FI_EFA_USE_DEVICE_RDMA=1 was set by users, "
+				"but EFA device has no rdma-read capability.\n");
+		abort();
+	}
 	fi_param_get_int(&rxr_prov, "use_zcpy_rx", &rxr_env.use_zcpy_rx);
 	fi_param_get_int(&rxr_prov, "set_cuda_sync_memops", &rxr_env.set_cuda_sync_memops);
 	fi_param_get_int(&rxr_prov, "zcpy_rx_seed", &rxr_env.zcpy_rx_seed);
@@ -139,14 +147,18 @@ void rxr_env_param_get(void)
 
 void rxr_env_define()
 {
+	char *str = "";
+
 	fi_param_define(&rxr_prov, "tx_min_credits", FI_PARAM_INT,
 			"Defines the minimum number of credits a sender requests from a receiver (Default: 32).");
 	fi_param_define(&rxr_prov, "tx_queue_size", FI_PARAM_INT,
 			"Defines the maximum number of unacknowledged sends with the NIC.");
 	fi_param_define(&rxr_prov, "enable_shm_transfer", FI_PARAM_INT,
 			"Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
+	if (!(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ))
+		str = "EFA device on your system does not support RDMA, so this variable cannot be set to 1";
 	fi_param_define(&rxr_prov, "use_device_rdma", FI_PARAM_INT,
-			"whether to use device's RDMA functionality for one-sided and two-sided transfer.");
+			"whether to use device's RDMA functionality for one-sided and two-sided transfer. (Default: %d). %s", rxr_env.use_device_rdma, str);
 	fi_param_define(&rxr_prov, "use_zcpy_rx", FI_PARAM_INT,
 			"Enables the use of application's receive buffers in place of bounce-buffers when feasible. (Default: 1)");
 	fi_param_define(&rxr_prov, "set_cuda_sync_memops", FI_PARAM_INT,
@@ -207,8 +219,26 @@ void rxr_env_define()
 			"The maximum number of bytes that will be eagerly sent by inflight messages uses runting read message protocol (Default 307200).");
 }
 
+/**
+ * @brief Initialize the variables in rxr_env.
+ * This function uses the g_device_list initialized in
+ * efa_device_list_initialize(), thus it must be called
+ * after efa_device_list_initialize().
+ */
 void rxr_env_initialize()
 {
+	uint32_t vendor_part_id = g_device_list[0].ibv_attr.vendor_part_id;
+
+	if (vendor_part_id == 0xefa0 || vendor_part_id == 0xefa1) {
+		rxr_env.use_device_rdma = 0;
+	} else {
+		rxr_env.use_device_rdma = 1;
+		if (!(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
+			FI_WARN(&rxr_prov, FI_LOG_FABRIC,
+				"EFA device with vendor_part_id 0x%x has no rdma-read capability.\n", vendor_part_id);
+			rxr_env.use_device_rdma = 0;
+		}
+	}
 	rxr_env_define();
 	rxr_env_param_get();
 }
--- a/prov/efa/src/rxr/rxr_prov.c
+++ b/prov/efa/src/rxr/rxr_prov.c
@@ -100,8 +100,6 @@ static void rxr_util_prov_finalize()
 
 EFA_INI
 {
-	rxr_env_initialize();
-
 #if HAVE_EFA_DL
 	ofi_mem_init();
 	ofi_hmem_init();
@@ -110,6 +108,8 @@ EFA_INI
 
 	if (efa_prov_initialize())
 		return NULL;
+
+	rxr_env_initialize();
 
 	if (rxr_util_prov_initialize())
 		return NULL;


prov/efa/src/efa_device.c
```diff
@@ -97,6 +97,11 @@ int efa_device_construct(struct efa_device *efa_device,
 		goto err_close;
 	}
 
+	if (efa_device->ibv_attr.vendor_part_id == 0xefa2) {
```
wzamazon, line 100: I think we should use an array for different `vendor_part_id`. e.g.

```
static int use_device_rdma[] = {
      [0] = 0,
      [1] = 0,
      [2] = 1};

rxr_env.use_device_rdma = use_device_rdma[efa_device->ibv_attr.vender_pard_id - 0xefa0);
```
Make it earier to add new device and change decision for different id.

shijin-aws: Thank you! Updated.

shijin-aws:
I think this PR needs more change.. Need to handle the case when user request FI_EFA_USE_DEVICE_RDMA for 0xefa0 and 0xefa1 case.

prov/efa/src/efa_device.c
bwbarrett: I'd skip the table. Just set it to 0 if the device id is 0xefa0 or 0xefa1. We're not going to go back on the default, so just do the future proof thing.

shijin-aws: The latest revision set `use_device_rdma` to 0 when the efa device id is < 0xefa2. There is a corner case that user could `FI_EFA_USE_DEVICE_RDMA=1` on p4d, which has `0xefa0`. So I add a check for `fi_get_param`. But I think mixing the env with the runtime default may not be the best practice. I am open to suggestions here.

prov/efa/src/efa_device.c
```diff
@@ -97,6 +97,15 @@ int efa_device_construct(struct efa_device *efa_device,
 		goto err_close;
 	}
 
+	if (-FI_ENODATA == fi_param_get_int(&rxr_prov, "use_device_rdma", &tmp_value)) {
+		if (efa_device->ibv_attr.vendor_part_id < 0xefa2) {
```
bwbarrett, line 101: Don't do this.  Explicitly compare against 0xefa0 and 0xefa1.  There's no guarantee that the device id will always start with 0xefa?.  We one day will have more than 16 EFA versions, and if for some reason, we decremented one of the first characters (like made the one after 0xefaf be 0xef00), then this would not do the right thing.

prov/efa/src/efa_device.c
```diff
@@ -97,6 +97,15 @@ int efa_device_construct(struct efa_device *efa_device,
...
+			rxr_env.use_device_rdma = 0;
+		} else {
+			assert(efa_device->efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ);
```
bwbarrett, line 104: Why is an assert right here?  Asserts aren't always compiled in and this seems less like an assumption enforced by code and more an assumption about the underlying system.  Seems better that to have always on error handling.  Also, asserts normally dump core, and this seems like a very wasteful error to dump core over.

bwbarrett (changes requested):

prov/efa/src/efa_device.c
tschuett: Could you pull out a function:
```c
bool canUseDeviceRdma(int vendor_part);
```

shijin-aws: Thank you! Fixed in the latest revision.

prov/efa/src/efa_device.c
shijin-aws: Thank you. I changed it into an error handling case in the latest revision.

prov/efa/src/efa_device.c
shijin-aws: >Could you pull out a function:
bool canUseDeviceRdma(int vendor_part);

This is only a simple `if` and I am not sure introducing an extra function can be better.

bwbarrett:
> I think this PR needs more change.. Need to handle the case when user request FI_EFA_USE_DEVICE_RDMA for 0xefa0 and 0xefa1 case.

You're right about this; you should always do what the user asks (unless the user asks for DEVICE_RDMA on a platform that doesn't have that capability).

prov/efa/src/efa_device.c
```diff
@@ -97,6 +97,20 @@ int efa_device_construct(struct efa_device *efa_device,
...
+		if (efa_device->ibv_attr.vendor_part_id == 0xefa0 ||
+		    efa_device->ibv_attr.vendor_part_id == 0xefa1) {
+			rxr_env.use_device_rdma = 0;
+		} else {
+			if (!(efa_device->efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
+				EFA_WARN(FI_LOG_DOMAIN, "EFA device has no rdma-read capability.\n");
+				err = -FI_EINVAL;
+				goto err_close;
+			}
+			rxr_env.use_device_rdma = 1;
```
wzamazon, line 110: I am not sure the action should be performed here.

IMO, this should be in `rxr_env_initialize` before `rxr_env_param_get` is called.

 There is no guarantee that this function is called before `rxr_env_initialize`.

shijin-aws: Should we just change https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rxr/rxr_env.c#L45 to 1 by default? and remove the assignment here.

wzamazon: Not sure I understand. where do you mean by "here"?

shijin-aws: I thought the `action` in your comment 

> `I am not sure the action should be performed here.` 

is the `rxr_env.use_device_rdma = 1;`  for non-0xefa0/1 device.

shijin-aws: >There is no guarantee that this function is called before rxr_env_initialize.

This function is called after `rxr_env_initialize`. What is the issue then?

shijin-aws:
> > I think this PR needs more change.. Need to handle the case when user request FI_EFA_USE_DEVICE_RDMA for 0xefa0 and 0xefa1 case.
> 
> You're right about this; you should always do what the user asks (unless the user asks for DEVICE_RDMA on a platform that doesn't have that capability).

Yes, in the latest revision I also add a check that if user asks for DEVICE_RDMA on a platform that doesn't have that capability, the function will warn and return error.

bwbarrett:
I think your latest patch still doesn't meet all the expectations of behavior customers will have.  Which are as follows:

1) If device id is `0xefa0` or `0xefa1`, default to device rdma off, otherwise default to device rdma on.
2) If the user explicitly requests to enable or disable device rdma via the environment, do that.
3) If device rdma is turned on (either by user request or by default) and the RDMA device capability is not set, then raise a useful error message (the error message may be different if it was on due to default or user enablement).

prov/efa/src/efa_device.c
wzamazon: Another advantage of moving this code to `rxr_env_initialize` is that we will be able to include the value `rxr_env.use_device_rdma` in the message for `FI_EFA_USE_DEVICE_RDMA`. This allow user to check whether RDMA read will be used by default by running `fi_info -e`.

shijin-aws:
> I think your latest patch still doesn't meet all the expectations of behavior customers will have. Which are as follows:
> 
>     1. If device id is `0xefa0` or `0xefa1`, default to device rdma off, otherwise default to device rdma on.
> 
>     2. If the user explicitly requests to enable or disable device rdma via the environment, do that.
> 
>     3. If device rdma is turned on (either by user request or by default) and the RDMA device capability is not set, then raise a useful error message (the error message may be different if it was on due to default or user enablement).

For 1, my latest patch will default to device rdma off for `0xefa0` and `0xefa1`, if user does not explicitly request.

For 2, my latest patch will NOT change the default if user explicitly request.

For 3, my latest patch will warn and fail when: a) device rdma is turned on by default for 0xefa2, but the  efa device caps does not support it; b) user requests rdma explicitly, but the returned efa device caps does not support it. For a), the current error message is `"EFA device has no rdma-read capability.\n"`. For b) the current error message is `"FI_EFA_USE_DEVICE_RDMA=1 was set, but EFA device has no rdma-read capability.\n"`. I think the error message for a) may need some improvements.

But I think it covers all the 3 situation you listed. Did I miss anything?

prov/efa/src/efa_device.c
shijin-aws: I move the code to `rxr_env_initialize` in the latest revision.

bwbarrett (approved):

prov/efa/src/rxr/rxr_env.c
```diff
@@ -93,7 +96,13 @@ void rxr_env_param_get(void)
 
 	fi_param_get_int(&rxr_prov, "tx_queue_size", &rxr_env.tx_queue_size);
 	fi_param_get_int(&rxr_prov, "enable_shm_transfer", &rxr_env.enable_shm_transfer);
-	fi_param_get_int(&rxr_prov, "use_device_rdma", &rxr_env.use_device_rdma);
+	ret = fi_param_get_int(&rxr_prov, "use_device_rdma", &rxr_env.use_device_rdma);
+	if (ret != -FI_ENODATA && rxr_env.use_device_rdma &&
+	    !(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
+		FI_WARN(&rxr_prov, FI_LOG_FABRIC, "FI_EFA_USE_DEVICE_RDMA=1 was set by users, "
```
wzamazon, line 102: I think we probably abort() here and `fprintf(stderr)` the error message, instead of return ERROR. 

Because reeturning error may cause application to use tcp silently.

shijin-aws: I thought about that. This function is actually called during libfabric registering all providers. So if users happen to have this env set on a platform that does not support rdma-read, the libfabric will abort early and user cannot use any provider. I can make the change if we are ok with that.

bwbarrett: That seems bad.  What upper layer doesn't block the tcp providers anyway?

wzamazon: Intel MPI will fallback to use tcp;ofi_rxm, if EFA provider is not available.

The abort only happens on a system with EFA device and FI_EFA_USE_DEVICE_RDMA was set explicitly. So I think it should be good.

prov/efa/src/rxr/rxr_env.c
```diff
@@ -146,7 +156,7 @@ void rxr_env_define()
 	fi_param_define(&rxr_prov, "enable_shm_transfer", FI_PARAM_INT,
 			"Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
 	fi_param_define(&rxr_prov, "use_device_rdma", FI_PARAM_INT,
-			"whether to use device's RDMA functionality for one-sided and two-sided transfer.");
+			"whether to use device's RDMA functionality for one-sided and two-sided transfer. (Default: %d)", rxr_env.use_device_rdma);
```
wzamazon, line 159: I think we can check device capability here. If the device does not support RDMDa, print something like "EFA device does not support RDMA, so you cannot set this variable to 1".

prov/efa/src/rxr/rxr_env.c
shijin-aws: Do we also abort when it's 0xefa2 but firmware does not support rdma-read ?

wzamazon: > 0xefa2 but firmware does not support rdma-read 

In that case, this patch exit early in `rxr_env_initialize()`. So we will not reach here.

That should not happen any way.

shijin-aws: I am asking whether we abort here

```
	int rxr_env_initialize()
{
	uint32_t vendor_part_id = g_device_list[0].ibv_attr.vendor_part_id;

	if (vendor_part_id == 0xefa0 || vendor_part_id == 0xefa1) {
		rxr_env.use_device_rdma = 0;
	} else {
		if (!(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
			FI_WARN(&rxr_prov, FI_LOG_FABRIC,
				"EFA device with vendor_part_id 0x%x has no rdma-read capability.\n", vendor_part_id);
			return -FI_EINVAL;
		}
		rxr_env.use_device_rdma = 1;
	}
```

wzamazon: I think we should not abort(), but we should not return error either.

I think we should print warning and proceed with `use_device_rdma` to 0.

If FI_EFA_USE_DEVICE_RDMA was set to 1,  abort()  there (in `rxr_env_param_get`).

If it is not set, we do not use read.

shijin-aws: OK, I updated the PR accordingly.

prov/efa/src/rxr/rxr_env.c
shijin-aws: Done.

wzamazon (approved):

wzamazon:
waiting for CI to finish.

shijin-aws:
bot:aws:retest

wzamazon:
bot:aws:retest

PR #8187, opened 2022-10-28
.github/workflows: Revert level-zero version workaround

aingerson:
Revert ".github/workflows: temporary workaround for failing ZE build"

This reverts commit 52f35998c2a93c37ce527641bfb43580501f8737.

The reverted commit was added as a workaround for a released version of level-zero which was buggy and failed building.

Reverting to test the latest in the distro instead of being stuck on 1.7.4


aingerson:
Not sure what version the distro is at or if the bug was fixed but testing to find out.

aingerson:
Looks like they updated the distro to 1.8.1 and it works so this can be merged

PR #8190, opened 2022-10-29
util/rxm_av: Add option to free resources on AV removal

shefty:
When sending or receiving messages, an AV may allocate resources underneath, such as establishing a connection. These connections can persist even if the peer's address is not inserted into the AV or after it's been removed. This can cause use of resources as long as the peer is running, but no longer actively communicating with another peer.

Add the ability to cleanup any underlying resources when removing an address from an AV.  This can free receive buffers, close fd's, etc. to reduce resource utilization.

This capability is disabled by default as it may abruptly terminate communication with a peer if there are data transfers in progress.  To enable, a libfabric wide environment variable must be set.

### DIFF
--- a/include/ofi.h
+++ b/include/ofi.h
@@ -328,6 +328,7 @@ uint8_t ofi_msb(uint64_t num);
 uint8_t ofi_lsb(uint64_t num);
 
 extern size_t ofi_universe_size;
+extern int ofi_av_remove_cleanup;
 
 bool ofi_send_allowed(uint64_t caps);
 bool ofi_recv_allowed(uint64_t caps);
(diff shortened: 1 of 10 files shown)


shefty:
Updates PR #8036 to fix locking and move the environment variable control to the OFI core (for later use with net provider).

chien-intel:
With so many changes from my original PR, please remove my Signed-off and authorship.  Pull in this PR when you are good with it so DAOS team can do more testing with this feature.  Thanks.

shefty:
Updated, verified using fi_av_xfer that the underlying connections are closed when av_remove is called.

shefty:
Intel CI failures related to shm.  AWS failures related to EFA.

PR #8193, opened 2022-10-31
core: API changes for peer provider and collectives

j-xiong:
* New calls to open domain & endpoint with flags
* Add flags to fi_barrier()

### DIFF
--- a/include/rdma/fi_domain.h
+++ b/include/rdma/fi_domain.h
@@ -285,6 +285,8 @@ struct fi_ops_domain {
 	int	(*query_collective)(struct fid_domain *domain,
 			enum fi_collective_op coll,
 			struct fi_collective_attr *attr, uint64_t flags);
+	int	(*endpoint2)(struct fid_domain *domain, struct fi_info *info,
+			struct fid_ep **ep, uint64_t flags, void *context);
 };
 
 /* Memory registration flags */
@@ -324,6 +326,18 @@ fi_domain(struct fid_fabric *fabric, struct fi_info *info,
 	   struct fid_domain **domain, void *context)
 {
 	return fabric->ops->domain(fabric, info, domain, context);
+}
+
+static inline int
+fi_domain2(struct fid_fabric *fabric, struct fi_info *info,
+	   struct fid_domain **domain, uint64_t flags, void *context)
+{
+	if (!flags)
+		return fi_domain(fabric, info, domain, context);
+
+	return FI_CHECK_OP(fabric->ops, struct fi_ops_fabric, domain2) ?
+		fabric->ops->domain2(fabric, info, domain, flags, context) :
+		-FI_ENOSYS;
 }
 
 static inline int
--- a/src/enosys.c
+++ b/src/enosys.c
@@ -71,6 +71,14 @@ int fi_no_domain(struct fid_fabric *fabric, struct fi_info *info,
 {
 	return -FI_ENOSYS;
 }
+int fi_no_domain2(struct fid_fabric *fabric, struct fi_info *info,
+		struct fid_domain **dom, uint64_t flags, void *context)
+{
+	if (flags)
+		return -FI_ENOSYS;
+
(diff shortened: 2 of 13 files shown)


include/rdma/fi_domain.h
```diff
@@ -326,6 +328,13 @@ fi_domain(struct fid_fabric *fabric, struct fi_info *info,
 	return fabric->ops->domain(fabric, info, domain, context);
 }
 
+static inline int
+fi_domain_ex(struct fid_fabric *fabric, struct fi_info *info,
+	     struct fid_domain **domain, uint64_t flags, void *context)
+{
+	return fabric->ops->domain_ex(fabric, info, domain, flags, context);
+}
```
shefty, line 336: I'm leaning more toward calling the new functions '2' vs 'ex'.  So, fi_domain2(), which is more forward compatible if we need yet another version.

For the static inline, we need to use FI_CHECK_OP() macro to handle DL providers and providers that are not updated to use the new APIs.  That allows us to drop changes to most providers.

prov/net/src/xnet_fabric.c
```diff
@@ -48,7 +48,7 @@ struct fi_ops_fabric xnet_fabric_ops = {
 	.eq_open = xnet_eq_create,
 	.wait_open = ofi_wait_fd_open,
 	.trywait = xnet_trywait,
-	.domain_ex = fi_no_domain__ex,
+	.domain_ex = fi_no_domain_ex,
```
shefty, line 51: change fell into wrong patch

prov/psm3/src/psmx3_fabric.c
```diff
@@ -73,7 +73,7 @@ static struct fi_ops_fabric psmx3_fabric_ops = {
 	.passive_ep = fi_no_passive_ep,
 	.eq_open = ofi_eq_create,
 	.wait_open = psmx3_wait_open,
-	.trywait = psmx3_wait_trywait
+	.trywait = psmx3_wait_trywait,
```
shefty, line 76: change fell into wrong patch

j-xiong:
Updated to use new function names, FI_CHECK_OP, and fix other issues.

src/enosys.c
```diff
@@ -71,6 +71,11 @@ int fi_no_domain(struct fid_fabric *fabric, struct fi_info *info,
 {
 	return -FI_ENOSYS;
 }
+int fi_no_domain2(struct fid_fabric *fabric, struct fi_info *info,
+		struct fid_domain **dom, uint64_t flags, void *context)
+{
+	return -FI_ENOSYS;
```
shefty, line 77: This would be different from other implementations, but I wonder if this call (and ep2) should be:
```
if (flags)
    return -FI_ENOSYS;

return fi_domain(...);
```
but with a different name?

j-xiong: Good point. This can also be handled in fi_domain2() and fi_endpoint2(). Or should it be done at both places? I think it's fine to keep the current name, even though it doesn't always return enosys.

j-xiong:
fi_barrier() is used by fabtests, so some changes there are also needed.

src/enosys.c
shefty: Now that you mention it, handling this in fi_domain2() would be helpful for migrating apps, except that I would handle it differently:
```
fi_domain2(...)
{
    if (!flags)
        return fi_domain(...);

    return FI_CHECK_OP(...) blah blah
}
```
Internally, we could have something like the call above that providers can point their implementation to (where !flags returns FI_ENOSYS).

j-xiong: updated.

j-xiong:
The single reference of fi_barrier() in fabtests has been updated.

shefty (approved):

j-xiong:
Intel CI failure is the known "fi_mutli_recv" issue with shm.  AWS CI failure is "unable to locate credentials" error for one of the builds. Both can be ignored for this PR.

PR #8194, opened 2022-10-31
util/mr_cache: Add MM valid check to cache find

iziemba:
ofi_mr_cache_find can be used to search the MR cache without creating a new entry. ofi_mr_cache_find should only return valid entries, but valid check with the corresponding memory monitor is missing. Add this check.

Signed-off-by: Ian Ziemba <ian.ziemba@hpe.com>

### DIFF
--- a/prov/util/src/util_mr_cache.c
+++ b/prov/util/src/util_mr_cache.c
@@ -380,6 +380,7 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 {
 	struct ofi_mr_info info;
 	struct ofi_mr_entry *entry;
+	struct ofi_mem_monitor *monitor;
 
 	assert(attr->iov_count == 1);
 	FI_DBG(cache->domain->prov, FI_LOG_MR, "find %p (len: %zu)\n",
@@ -395,6 +396,12 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 	}
 
 	if (!ofi_iov_within(attr->mr_iov, &entry->info.iov)) {
+		entry = NULL;
+		goto unlock;
+	}
+
+	monitor = cache->monitors[entry->info.iface];
+	if (!monitor->valid(monitor, &entry->info, entry)) {
 		entry = NULL;
 		goto unlock;
 	}


prov/util/src/util_mr_cache.c
```diff
@@ -399,6 +400,13 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 		goto unlock;
 	}
 
+	monitor = cache->monitors[entry->info.iface];
+	if (!monitor->valid(monitor, (const void *)entry->info.iov.iov_base,
+			    entry->info.iov.iov_len, &entry->hmem_info)) {
```
j-xiong, line 405: Check the definition of `valid()`. The parameters here don't match the prototype.

iziemba: Thanks. Fixed.

prov/util/src/util_mr_cache.c
```diff
@@ -399,6 +400,12 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 		goto unlock;
 	}
 
+	monitor = cache->monitors[entry->info.iface];
+	if (!monitor->valid(monitor, &entry->info, entry)) {
```
j-xiong, line 404: Need to check for NULL monitor first like how `ofi_mr_cache_search` does.

prov/util/src/util_mr_cache.c
```diff
@@ -399,6 +400,12 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
...
+		entry = NULL;
```
j-xiong, line 405: Should the invalid entry be purged here?

j-xiong:
@wckzhang You added this function a few years ago. Do you have any comments on the intended usage of this function? Since this is not being used anywhere.

shefty:
ofi_mr_cache_search() has the correct support -- calls valid(), flushes stale entries, removes overlapping entries, etc.  oi_mr_cache_find() isn't called anywhere and is incomplete.  It needs more work than simply calling valid() and should mimic search() in most areas.  The two calls should invoke some common path.

@iziemba - Do you need the find() call, or can search() work for you?  I'm wondering if we should keep find() at all.

prov/util/src/util_mr_cache.c
iziemba: I believe if this returns NULL, this means an entry was added to the cache without a memory monitor entry which seems like a fatal flaw in the MR cache. I can add an assert. Thoughts?

shefty: A 'passive' memory monitor will not verify that an entry is still valid until it is accessed.  So, the entry would have been valid when it was added, but the valid check now fails.  This is how both CUDA and Ze monitors work.  Only RoCR is active in this regard and intercepts GPU allocation/free calls to invalidate entries.

iziemba:
search() has the issue where it requires the fi_hmem_iface to be known before calling into search(). Since the CXI provider does not support FI_MR_HMEM, we have to internally query for the fi_hmem_iface each time before calling search() which is not ideal.

Since device MR caching requires unified virtual addressing, the fi_hmem_iface type only needs to be known if there is a cache miss. find() provides this functionality.

find() is missing the flush and the eviction if valid is false. This needs to be added.

Maybe we should keep and improve find() and actually change search() to insert()?

prov/util/src/util_mr_cache.c
iziemba: It should. Will add.

shefty:
The search/find names are mirroring the tsearch/tfind names.  The search and find operations *should* be basically the same, with the exception of whether the missing entry is inserted if not found.

We should have both calls invoke some common function, in this case search() should basically call find(), so we don't duplicate the checks and end up in the same situation later where only one of the calls was updated.

shefty (changes requested):
Simply updating my submitted review to request a shared path for search/find.

shefty:
Closing as stale.  Please re-open this PR (to pull up the prior discussion) once rebased and ready for submission.

PR #8198, opened 2022-10-31
prov/efa: Fix shm initialization path on error

wckzhang:

### DIFF
--- a/prov/efa/src/efa_user_info.c
+++ b/prov/efa/src/efa_user_info.c
@@ -521,27 +521,36 @@ int efa_getinfo(uint32_t version, const char *node,
 		shm_info_initialized = true;
 	}
 
-	if (hints && hints->ep_attr && hints->ep_attr->type == FI_EP_DGRAM)
-		return efa_user_info_get_dgram(version, node, service, flags, hints, info);
+	if (hints && hints->ep_attr && hints->ep_attr->type == FI_EP_DGRAM) {
+		err = efa_user_info_get_dgram(version, node, service, flags, hints, info);
+		if (err)
+			goto error;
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_user_info.c
```diff
@@ -521,27 +521,38 @@ int efa_getinfo(uint32_t version, const char *node,
 		shm_info_initialized = true;
 	}
 
-	if (hints && hints->ep_attr && hints->ep_attr->type == FI_EP_DGRAM)
-		return efa_user_info_get_dgram(version, node, service, flags, hints, info);
+	if (hints && hints->ep_attr && hints->ep_attr->type == FI_EP_DGRAM) {
+		err = efa_user_info_get_dgram(version, node, service, flags, hints, info);
+		if (err)
+			goto error;
+		else
```
wzamazon, line 528: It seems we do not need `else` and we can `return 0`.

wckzhang: That's mostly just style right? Do you want it changed or is it okay as is?

wzamazon: I will prefer the change if it is not too much trouble.

shijin-aws:
How far do we need to backport?

wzamazon:
>How far do we need to backport?

Definitely 1.16.x

prov/efa/src/efa_user_info.c
wckzhang: Done

wzamazon (approved):

wzamazon:
Can you please backport to v1.16.x?

PR #8199, opened 2022-10-31
Add Support for ROCR IPC

amirshehataornl:
Add the following support:

- Add ROCR IPC for use with SHM provider
- Support Asynchronous ROCR memory copy operations
- improve ROCR device performance by using inline copies instead of asynchronous copies for buffers less than 512KB

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -43,7 +43,10 @@
 
 extern bool ofi_hmem_disable_p2p;
 
+typedef void* ofi_hmem_async_event_t;
+
 #define MAX_IPC_HANDLE_SIZE	64
+#define MAX_NUM_ASYNC_OP	4
 
 /*
  * This structure is part of the
@@ -88,15 +91,10 @@ cudaError_t ofi_cudaFree(void *ptr);
 
 /* Libfabric support ROCr operations. */
 
-hsa_status_t ofi_hsa_memory_copy(void *dst, const void *src, size_t size);
 hsa_status_t ofi_hsa_amd_pointer_info(void *ptr, hsa_amd_pointer_info_t *info,
 				      void *(*alloc)(size_t),
 				      uint32_t *num_agents_accessible,
 				      hsa_agent_t **accessible);
-hsa_status_t ofi_hsa_init(void);
-hsa_status_t ofi_hsa_shut_down(void);
-hsa_status_t ofi_hsa_status_string(hsa_status_t status,
-				   const char **status_string);
 const char *ofi_hsa_status_to_string(hsa_status_t status);
 
 hsa_status_t ofi_hsa_amd_dereg_dealloc_cb(void *ptr,
@@ -105,17 +103,22 @@ hsa_status_t ofi_hsa_amd_reg_dealloc_cb(void *ptr,
 					hsa_amd_deallocation_callback_t cb,
 					void *user_data);
 
-hsa_status_t ofi_hsa_amd_memory_lock(void *host_ptr, size_t size,
-				     hsa_agent_t *agents, int num_agents,
-				     void **agent_ptr);
-hsa_status_t ofi_hsa_amd_memory_unlock(void *host_ptr);
-
 #endif /* HAVE_ROCR */
 
 struct ofi_hmem_ops {
 	bool initialized;
 	int (*init)(void);
 	int (*cleanup)(void);
+	int (*create_async_copy_event)(uint64_t device,
+				       ofi_hmem_async_event_t *event);
+	int (*free_async_copy_event)(uint64_t device,
+				     ofi_hmem_async_event_t event);
+	int (*async_copy_to_hmem)(uint64_t device, void *dest, const void *src,
+			    size_t size, ofi_hmem_async_event_t event);
+	int (*async_copy_from_hmem)(uint64_t device, void *dest,
+			const void *src, size_t size,
+			ofi_hmem_async_event_t event);
+	int (*async_copy_query)(ofi_hmem_async_event_t event);
 	int (*copy_to_hmem)(uint64_t device, void *dest, const void *src,
 			size_t size);
 	int (*copy_from_hmem)(uint64_t device, void *dest, const void *src,
@@ -151,6 +154,14 @@ int rocr_open_handle(void **handle, size_t size, uint64_t device,
 		     void **ipc_ptr);
 int rocr_close_handle(void *ipc_ptr);
 bool rocr_is_ipc_enabled(void);
+int rocr_create_async_copy_event(uint64_t device,
+				 ofi_hmem_async_event_t *event);
+int rocr_free_async_copy_event(uint64_t device, ofi_hmem_async_event_t event);
+int rocr_async_copy_to_dev(uint64_t device, void *dst, const void *src,
+			  size_t size, ofi_hmem_async_event_t event);
+int rocr_async_copy_from_dev(uint64_t device, void *dst, const void *src,
+			    size_t size, ofi_hmem_async_event_t event);
+int rocr_async_copy_query(ofi_hmem_async_event_t event);
 
 int cuda_copy_to_dev(uint64_t device, void *dev, const void *host, size_t size);
 int cuda_copy_from_dev(uint64_t device, void *host, const void *dev, size_t size);
@@ -241,6 +252,24 @@ static inline int ofi_memcpy(uint64_t device, void *dest, const void *src,
 	return FI_SUCCESS;
 }
 
+static inline int ofi_no_create_async_copy_event(uint64_t device,
+		ofi_hmem_async_event_t *event)
+{
+	return -FI_ENOSYS;
+}
+
+static inline int ofi_no_free_async_copy_event(uint64_t device,
+		ofi_hmem_async_event_t event)
+{
+	return -FI_ENOSYS;
+}
+
+static inline int ofi_no_async_memcpy(uint64_t device, void *dest, const void *src,
+			     size_t size, ofi_hmem_async_event_t event)
+{
+	return -FI_ENOSYS;
+}
+
 static inline int ofi_hmem_init_noop(void)
 {
 	return FI_SUCCESS;
@@ -249,6 +278,11 @@ static inline int ofi_hmem_init_noop(void)
 static inline int ofi_hmem_cleanup_noop(void)
 {
 	return FI_SUCCESS;
+}
+
+static inline int ofi_no_async_copy_query(ofi_hmem_async_event_t event)
+{
+	return -FI_ENOSYS;
 }
 
 static inline int ofi_hmem_no_get_handle(void *base_addr, size_t size,
@@ -298,6 +332,27 @@ static inline bool ofi_hmem_p2p_disabled(void)
 {
 	return ofi_hmem_disable_p2p;
 }
+
+int ofi_create_async_copy_event(enum fi_hmem_iface iface, uint64_t device,
+				ofi_hmem_async_event_t *event);
+
+int ofi_free_async_copy_event(enum fi_hmem_iface iface, uint64_t device,
+			      ofi_hmem_async_event_t event);
+
+ssize_t ofi_async_copy_from_hmem_iov(void *dest, size_t size,
+				enum fi_hmem_iface hmem_iface, uint64_t device,
+				const struct iovec *hmem_iov,
+				size_t hmem_iov_count, uint64_t hmem_iov_offset,
+				ofi_hmem_async_event_t event);
+
+ssize_t ofi_async_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
+				const struct iovec *hmem_iov,
+				size_t hmem_iov_count, uint64_t hmem_iov_offset,
+				const void *src, size_t size,
+				ofi_hmem_async_event_t event);
+
+int ofi_async_copy_query(enum fi_hmem_iface iface,
+			 ofi_hmem_async_event_t event);
 
 ssize_t ofi_copy_from_hmem_iov(void *dest, size_t size,
 			       enum fi_hmem_iface hmem_iface, uint64_t device,
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -121,13 +121,14 @@ struct ofi_hmem_ops {
 	int (*copy_from_hmem)(uint64_t device, void *dest, const void *src,
 			size_t size);
 	bool (*is_addr_valid)(const void *addr, uint64_t *device, uint64_t *flags);
-	int (*get_handle)(void *base_addr, void **handle);
-	int (*open_handle)(void **handle, uint64_t device, void **mapped_addr);
+	int (*get_handle)(void *base_addr, size_t base_length, void **handle);
+	int (*open_handle)(void **handle, size_t base_length, uint64_t device,
+			   void **mapped_addr);
 	int (*close_handle)(void *mapped_addr);
 	int (*host_register)(void *addr, size_t size);
 	int (*host_unregister)(void *addr);
 	int (*get_base_addr)(const void *addr, void **base_addr,
-			size_t *base_length);
+			     size_t *base_length);
 	bool (*is_ipc_enabled)(void);
 	int (*get_ipc_handle_size)(size_t *size);
 };
@@ -143,7 +144,13 @@ int rocr_hmem_cleanup(void);
 bool rocr_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags);
 int rocr_host_register(void *ptr, size_t size);
 int rocr_host_unregister(void *ptr);
+int rocr_get_ipc_handle_size(size_t *size);
 int rocr_get_base_addr(const void *ptr, void **base, size_t *size);
+int rocr_get_handle(void *dev_buf, size_t size, void **handle);
+int rocr_open_handle(void **handle, size_t size, uint64_t device,
+		     void **ipc_ptr);
+int rocr_close_handle(void *ipc_ptr);
+bool rocr_is_ipc_enabled(void);
 
 int cuda_copy_to_dev(uint64_t device, void *dev, const void *host, size_t size);
 int cuda_copy_from_dev(uint64_t device, void *host, const void *dev, size_t size);
@@ -154,8 +161,9 @@ int cuda_host_register(void *ptr, size_t size);
 int cuda_host_unregister(void *ptr);
 int cuda_dev_register(struct fi_mr_attr *mr_attr, uint64_t *handle);
 int cuda_dev_unregister(uint64_t handle);
-int cuda_get_handle(void *dev_buf, void **handle);
-int cuda_open_handle(void **handle, uint64_t device, void **ipc_ptr);
+int cuda_get_handle(void *dev_buf, size_t size, void **handle);
+int cuda_open_handle(void **handle, size_t size, uint64_t device,
+		     void **ipc_ptr);
 int cuda_close_handle(void *ipc_ptr);
 int cuda_get_base_addr(const void *ptr, void **base, size_t *size);
 
@@ -189,8 +197,9 @@ int ze_hmem_copy(uint64_t device, void *dst, const void *src, size_t size);
 int ze_hmem_init(void);
 int ze_hmem_cleanup(void);
 bool ze_hmem_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags);
-int ze_hmem_get_handle(void *dev_buf, void **handle);
-int ze_hmem_open_handle(void **handle, uint64_t device, void **ipc_ptr);
+int ze_hmem_get_handle(void *dev_buf, size_t size, void **handle);
+int ze_hmem_open_handle(void **handle, size_t size, uint64_t device,
+			void **ipc_ptr);
 int ze_hmem_get_shared_handle(int dev_fd, void *dev_buf, int *ze_fd,
 			      void **handle);
 int ze_hmem_open_shared_handle(int dev_fd, void **handle, int *ze_fd,
@@ -242,13 +251,14 @@ static inline int ofi_hmem_cleanup_noop(void)
 	return FI_SUCCESS;
 }
 
-static inline int ofi_hmem_no_get_handle(void *base_addr, void **handle)
+static inline int ofi_hmem_no_get_handle(void *base_addr, size_t size,
+					 void **handle)
 {
 	return -FI_ENOSYS;
 }
 
-static inline int
-ofi_hmem_no_open_handle(void **handle, uint64_t device, void **mapped_addr)
+static inline int ofi_hmem_no_open_handle(void **handle, size_t size,
+					  uint64_t device, void **mapped_addr)
 {
 	return -FI_ENOSYS;
 }
@@ -311,9 +321,10 @@ static inline int ofi_copy_from_hmem(enum fi_hmem_iface iface, uint64_t device,
 	return hmem_ops[iface].copy_from_hmem(device, dest, src, size);
 }
 
-int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *base_addr, void **handle);
+int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *base_addr,
+			size_t size, void **handle);
 int ofi_hmem_open_handle(enum fi_hmem_iface iface, void **handle,
-			 uint64_t device, void **mapped_addr);
+			 size_t size, uint64_t device, void **mapped_addr);
 int ofi_hmem_close_handle(enum fi_hmem_iface iface, void *mapped_addr);
 int ofi_hmem_get_base_addr(enum fi_hmem_iface iface, const void *addr,
 			   void **base_addr, size_t *base_length);
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -49,6 +49,11 @@ struct ofi_hmem_ops hmem_ops[] = {
 		.cleanup = ofi_hmem_cleanup_noop,
 		.copy_to_hmem = ofi_memcpy,
 		.copy_from_hmem = ofi_memcpy,
+		.create_async_copy_event = ofi_no_create_async_copy_event,
+		.free_async_copy_event = ofi_no_free_async_copy_event,
+		.async_copy_to_hmem = ofi_no_async_memcpy,
+		.async_copy_from_hmem = ofi_no_async_memcpy,
+		.async_copy_query = ofi_no_async_copy_query,
 		.get_handle = ofi_hmem_no_get_handle,
 		.open_handle = ofi_hmem_no_open_handle,
 		.close_handle = ofi_hmem_no_close_handle,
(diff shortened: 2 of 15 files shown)


src/hmem.c
```diff
@@ -208,15 +208,15 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
 				     (void *) src, size, OFI_COPY_BUF_TO_IOV);
 }
 
-int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *dev_buf, void **handle)
+int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *dev_buf, size_t size, void **handle)
```
shefty, line 298: Here and several other places, please limit line lengths to about 80 characters. A lot of the hmem related code are too long.

src/hmem_rocr.c
```diff
@@ -65,18 +65,27 @@ struct rocr_ops {
...
 };
 
 #if ENABLE_ROCR_DLOPEN
 
 #include <dlfcn.h>
 
-static void *rocr_handle;
-static struct rocr_ops rocr_ops;
+static void *hsa_handle;
+static struct hsa_ops hsa_ops;
```
shefty, line 112: Please move separate updates into separate patches to make it easier to review and cherry pick changes

src/hmem_rocr.c
```diff
@@ -183,13 +203,18 @@ static int rocr_memcpy(void *dest, const void *src, size_t size)
 	return -FI_EIO;
 }
 
-static int rocr_host_memory_ptr(void *host_ptr, void **ptr)
+static int rocr_host_memory_ptr(void *host_ptr, void **ptr,
+								hsa_agent_t *agent, size_t *size,
+								uint64_t *offset, bool *system)
```
shefty, line 305: indentation is wonky

src/hmem_rocr.c
```diff
@@ -200,12 +225,30 @@ static int rocr_host_memory_ptr(void *host_ptr, void **ptr)
...
+		if (info.type == HSA_EXT_POINTER_TYPE_UNKNOWN && system)
+			*system = true;
+		if (offset)
+			*offset = host_ptr - *ptr;
+	} else {
 		*ptr = (void *) ((uintptr_t) info.agentBaseAddress +
 				 (uintptr_t) host_ptr -
 				 (uintptr_t) info.hostBaseAddress);
+		if (system)
+			*system = true;
```
shefty, line 347: This was already set above

src/hmem_rocr.c
```diff
@@ -200,12 +225,30 @@ static int rocr_host_memory_ptr(void *host_ptr, void **ptr)
...
+
+	if (info.type != HSA_EXT_POINTER_TYPE_LOCKED) {
+		if (info.type == HSA_EXT_POINTER_TYPE_IPC ||
+			info.type == HSA_EXT_POINTER_TYPE_HSA)
+			*ptr = info.agentBaseAddress;
+		else
+			*ptr = host_ptr;
+
+		if (info.type == HSA_EXT_POINTER_TYPE_UNKNOWN && system)
+			*system = true;
```
shefty, line 339: This was already set above

include/ofi_hmem.h
```diff
@@ -121,8 +121,8 @@ struct ofi_hmem_ops {
 	int (*copy_from_hmem)(uint64_t device, void *dest, const void *src,
 			      size_t size);
 	bool (*is_addr_valid)(const void *addr, uint64_t *device, uint64_t *flags);
-	int (*get_handle)(void *dev_buf, void **handle);
-	int (*open_handle)(void **handle, uint64_t device, void **ipc_ptr);
+	int (*get_handle)(void *dev_buf, size_t size, void **handle);
+	int (*open_handle)(void **handle, size_t size, uint64_t device, void **ipc_ptr);
```
shefty, line 121: Separate these changes into its own patch.

@aingerson @j-xiong @wzamazon  - According to the ROCR documentation, hsa_amd_ipc_memory_create() (i.e. get_handle() only succeeds of the buffer is the fully allocated buffer.  dev_buf must point to the start of the buffer and len must be the allocation.  Does this same restriction hold for the other hmem interfaces?

Maybe more importantly, do we expect applications will adhere to this restriction when performing transfers?  I'm skeptical that this is the case.  And if it isn't, does the ROCR implementation need to adjust for the buffers accessed by the application, versus those which must be passed into the hsa_ calls?

shefty (commented):
Only reviewed first patch. I have open questions about the application flow through these calls.

include/ofi_hmem.h
amirshehataornl: ofi_hmem_get_base_addr() returns the length of the buffer. This is then passed to the get_handle() call. Then passed on to the other process, which uses it in the open_handle()

The ROCR version of get_base_addr() will return the entire buffer size allocated by the user. For example if you allocate a  1M buffer but the application passes a pointer within that buffer, then base_addr will return the base_addr of the original buffer + the entire length of the buffer.

this information + offsets can be used in the caching.

I'm not sure what you mean when you refer to the restrictions.

j-xiong: From onaAPI level-zero document:  `The pointer must be base pointer of the device memory allocation; i.e. the value returned from [zeMemAllocDevice]`.  CUDA is silimar `Takes a pointer to the base of an existing device memory allocation created with cudaMalloc and exports it for use in another process.`

The handle always refers to the entire allocation, so I don't think the `size` parameter is needed. If ROCR requires one, it can always be obtained via the `get_base_addr` method. There is no need to change the API here.

amirshehataornl: I did get it via get_base_addr(), that's already in format_ipc(). However, if I don't pass it then I'll need to call it twice. Not sure if this is necessary.

src/hmem_rocr.c
amirshehataornl: not sure I follow. system is initialized to false and set to true under two conditions: memory type is Unknown or locked.

shefty: I misread the top assignment

src/hmem_rocr.c
amirshehataornl: The wonkiness in the display looks to be an issue with how github displays tab. I have VI set to have tab==4 spaces. But I think the default tab display in here is tab==8spaces.
If you add "?ts=4" in the url, it displays properly.

amirshehataornl: oh. didn't know that. Will adjust.

amirshehataornl:
Rebased on the latest main

include/ofi_hmem.h
```diff
@@ -127,6 +127,7 @@ struct ofi_hmem_ops {
 			     size_t *size);
 	bool (*is_ipc_enabled)(void);
 	int (*get_ipc_handle_size)(size_t *size);
+	int (*get_async_copy_cutoff)(size_t *size);
```
aingerson, line 130: If this is an environment variable, why add it to the hmem ops to look it up? This is a lot of code added to all of the other interfaces just to add a param_get inside the rocr code.

amirshehataornl: The primary reason I did this is because I wanted to call it from SHM provider. The SHM provider makes a decision based on this value and the size of the message it's progressing to either do an asynchronous memory op or synchronous. When I was testing, I found that synchronous calls out perform asynchronous calls for message sizes below 512 KB. If the asynchronous approach is adopted by other device types, this same performance aspect might apply to them. In this case the cutoff value might differ between different devices, and it would be more generic to use the approach in this patch.

amirshehataornl:
This is an automatically generated comment.
The patch which shows the code changes this latest
update brings can be found in the following link
https://github.com/amirshehataornl/gh_diff_amirshehataornl/blob/main/8199/2022-11-17-17-23-16-919306.8199.patch

amirshehataornl:
This is an automatically generated comment.
The patch which shows the code changes this latest
update brings can be found in the following link
https://github.com/amirshehataornl/gh_diff_amirshehataornl/blob/main/8199/2022-11-17-17-37-8-189844.8199.patch

amirshehataornl:
@shefty, @j-xiong, @aingerson
This is an automatically generated comment.
The link below:
https://github.com/amirshehataornl/gh_diff_amirshehataornl/blob/main/8199/2022-11-24-16-53-24-145826.8199.patch
Shows the updates in the new revision of this PR

include/ofi_hmem.h
shefty: Until we have another implementation, it's not worth carrying the extra complexity in the hope that it may someday be useful.  It's too difficult to predict what might be needed.  The cutoff may be more closely related to the actual device, not what API it's accessed through.

include/ofi_hmem.h
```diff
@@ -105,17 +101,17 @@ hsa_status_t ofi_hsa_amd_reg_dealloc_cb(void *ptr,
...
 #endif /* HAVE_ROCR */
 
 struct ofi_hmem_ops {
 	bool initialized;
 	int (*init)(void);
 	int (*cleanup)(void);
+	int (*async_copy_to_hmem)(uint64_t device, void *dest, const void *src,
+			    size_t size, void **stream);
+	int (*async_copy_from_hmem)(uint64_t device, void *dest, const void *src,
+			      size_t size, void **stream);
```
shefty, line 113: Stream as output looks wrong.  I would have expected this to be input to indicate which stream to perform the copy, with some sort of signal or event being used to indicate the completion.  All the major GPUs support this type of operation, can we get an overview of how it would be done on each?

The application may want some control over which stream is used, versus the library picking it.

include/ofi_hmem.h
amirshehataornl: I don't call it. It was added as part of the caching patches. It's being called from smr_format_ipc(). It's called to get the base address and the base length, then the base address is being passed to ofi_hmem_get_handle() to get the ipc_handle

include/ofi_hmem.h
amirshehataornl: HSA has its own signal infrastructure which is used to indicate when the operation is complete. A stream is an abstract object which is meant to track several set of iovec operations. For example if the iovec has several entries and you do an async copy on each. This intent of this API is to allow the ROCR specific code to manage the signals and return an abstract object, stream, which is then used by the caller to query when the operations are complete. I don't see why the application would want to control that. 

The concept of stream here does not map to the hip or cuda stream. In hip, if you want to use an API like hipMemcpyAsync(), you'll need  to create the stream using hipStreamCreate() then pass that to hipMemcpyAsync(). Then you destroy it once you're done. I suppose we can follow the same flow, where we add new APIs to create and destroy streams and then pass the stream to the async_copy API.  

However, this will complicate the API for a non-existent use case. As you mentioned in a  previous comment "Until we have another implementation, it's not worth carrying the extra complexity in the hope that it may someday be useful. It's too difficult to predict what might be needed."

include/ofi_hmem.h
amirshehataornl: I removed the patch which adds this for now. I'll create a separate PR for this particular improvement.

edgargabriel:
I am not an expert on the libfabric internals, but we ran the code in this pr on some of our internal systems at AMD with Open MPI, and modulo a couple of smaller things it seems to be working. I will add a few minor comments to the code, please feel free to apply or ignore :-)

src/hmem_rocr.c
```diff
@@ -35,15 +35,43 @@
...
+	int num_signals;
+};
+
+OFI_DECLARE_FREESTACK(struct ofi_hsa_stream, rocm_ipc_stream_fs);
+OFI_DECLARE_FREESTACK(struct ofi_hsa_signal_info, rocm_ipc_signal_fs);
+
+pthread_spinlock_t fs_lock;
+struct rocm_ipc_stream_fs *ipc_stream_fs;
+struct rocm_ipc_signal_fs *ipc_signal_fs;
+
```
edgargabriel, line 64: maybe initialize the ipc_stream_fs and ipc_signal_fs pointer to NULL

src/hmem_ze.c
```diff
@@ -406,7 +406,7 @@ int ze_hmem_get_shared_handle(int dev_fd, void *dev_buf, int *ze_fd,
 	int ret;
 
 	assert(dev_fd != -1);
-	ret = ze_hmem_get_handle(dev_buf, (void **) &ze_handle);
+	ret = ze_hmem_get_handle(dev_buf, 0, (void **) &ze_handle);
```
shefty, line 409: Something is off in the abstractions.  The PR adds a new size parameter, which we ignore and pass in an invalid value here...

src/hmem_ze.c
```diff
@@ -443,7 +443,7 @@ int ze_hmem_open_shared_handle(int dev_fd, void **handle, int *ze_fd,
 	*ze_fd = open_fd.fd;
 	memset(&ze_handle, 0, sizeof(ze_handle));
 	memcpy(&ze_handle, &open_fd.fd, sizeof(open_fd.fd));
-	return ze_hmem_open_handle((void **) &ze_handle, device, ipc_ptr);
+	return ze_hmem_open_handle((void **) &ze_handle, 0, device, ipc_ptr);
```
shefty, line 446: ...and here.  We should either pass in a valid size value or call a different function that doesn't require it.

include/ofi_hmem.h
edgargabriel: I am wondering whether the name 'stream' is causing confusion, isn't this really behaving more like a request in MPI, i.e. you can use that to check for completion, with the difference being that it contains more than one element potentially.

include/ofi_hmem.h
```diff
@@ -105,17 +101,17 @@ hsa_status_t ofi_hsa_amd_reg_dealloc_cb(void *ptr,
...
 
 struct ofi_hmem_ops {
 	bool initialized;
 	int (*init)(void);
 	int (*cleanup)(void);
+	int (*async_copy_to_hmem)(uint64_t device, void *dest, const void *src,
+			    size_t size, void **stream);
+	int (*async_copy_from_hmem)(uint64_t device, void *dest,
+			const void *src, size_t size, void **stream);
+	int (*async_copy_query)(void *stream);
```
shefty, line 114: My first concern here is the name 'stream', which is confusing when considering the more generic hmem APIs.  The way I envision an async operation working is for the *caller* to pass in some sort of 'event'.  It can then wait or query on whether that event has completed.

Having the function return an event opens up issues with managing the lifetime of the object.  It's clearer if the caller allocates, uses, and frees an event, rather than having the allocation and free hidden without the user of the object knowing when it is freed.

amirshehataornl: @shefty, I'm trying to follow the change you'd like to see here. Currently the event is specific to ROCR, because the signals are ROCR specific. Are you talking about creating an event abstraction, which is populated by the the ROCR specific code, but allocated by the caller?

src/hmem_ze.c
edgargabriel: I think the line of thinking was  that in ```smr_format_ipc``` there is this sequence of operations:

```
ret = ofi_hmem_get_base_addr(cmd->msg.data.ipc_info.iface, ptr, &base,
				     &cmd->msg.data.ipc_info.base_length);
...
ret = ofi_hmem_get_handle(cmd->msg.data.ipc_info.iface, base,
				   (void **)&cmd->msg.data.ipc_info.ipc_handle);
```
and since the rocr function called internally in ofi_hmem_get_handle requires the buffer size to retrieve the ipc handle for that buffer,  the code wanted to avoid having to 'rediscover' the buffer size internally in that function again.

I agree however, that if there are other locations in the code that do not follow this pattern, the interface change might look unnecessary. In addition, I am not sure that this interface change is essential for this pr, its an optimization  that could also be discussed or introduced in a separate pr.

include/ofi_hmem.h
shefty: It looks like cuda uses a similar name, sigh...
```
cudaMemcpyAsync(dst, src, count, kind, stream)
```
kind determines the direction here, which is covered by having separate APIs in your case.  Stream isn't a request, but where the copy is performed.  From what I can see, there's no great way to determine if a specific copy has completed.  You can only query the 'stream' to determine if *all* previous operations have completed, or block on the stream until they do.

If we have a constant stream (ha) of requests flowing, it's possible that we'll never complete any of the prior ones, since the stream will never be empty.

The hsa API is more complex:
```
hsa_amd_memory_async_copy(void* dst, hsa_agent_t dst_agent, const void* src,
                              hsa_agent_t src_agent, size_t size,
                              uint32_t num_dep_signals,
                              const hsa_signal_t* dep_signals,
                              hsa_signal_t completion_signal);
```
The dep_signals parameters in this case can be used to defer the start of the copy.  I'm assuming that we don't want that and can ignore those.  The completion_signal is an event that can be updated once the copy completes.  I think this aligns with Ze events (@aingerson).

This was more aligned with what I was thinking:
User creates some event and then checks if it's been signaled.

But that flow doesn't seem to align with cuda, or even your proposal here.  Looking at the PR, it looks like you're converting the HSA model into the CUDA model.  IMO, the HSA model is better. We can check for a specific copy to have completed.  With the CUDA model, we only know if all of them have completed.  Anyway, this is what my original thoughts on an OFI abstraction for this:

```
typedef void * ofi_hmem_async_event_t;

int (*create_async_event)(device, ofi_hmem_async_event_t *event);
int (*free_async_event)(device, ofi_hmem_async_event_t event);

int (*async_copy_to_hmem)(uint64_t device, void *dest, const void *src,
	    size_t size, ofi_hmem_async_event_t event);
int (*async_copy_from_hmem)(uint64_t device, void *dest,
	const void *src, size_t size, ofi_hmem_async_event_t event);
int (*async_query_event)(ofi_hmem_async_event_t event);
```
This would allow for other async operations (if we ever need them).  It would enable checking if a single copy operation had completed, which *might* be a plus.  But also requires multiple checks to clear all async copies, which *might* be a negative.  This API should map okay to HSA and Ze, but not CUDA.  CUDA would basically convert the async_query_event() call to checking if the stream is cleared.

If this is too much overhead, we can replace async_query_event() with the more generic call like you had, async_query(), and only return if all previous async operations had completed.  If we go with the second option, I'd hide the 'stream' completely for now.  For example, we don't let the caller specify where the copy is done when calling copy_to_hmem(), so I'd align with that.  The async_query(device) would only need to know the device and would just report if everything had completed there.

wckzhang: ```From what I can see, there's no great way to determine if a specific copy has completed. You can only query the 'stream' to determine if all previous operations have completed, or block on the stream until they do.```

If I'm wildly off here, ignore my comment since I'm replying without context.

If you're referring to the cuda API, you can technically check if a specific copy has completed with the event API. When you call cuda memcpy async, you can record an event onto the same stream and then query/synchronize on that event to check for that specific copy. I think this maps to the query event behavior you are mentioning.

shefty: That sounds exactly like the behavior I'm looking for.  And if OpenMPI uses this flow, then it's likely decent.  @amirshehataornl - does the above proposal seem like it would work then?

wckzhang: To give a description on what exactly is done in Open MPI:

Create an array and initialize event objects on initialization (since events can be reused, don't have to call create_event/destroy_event each time). There are two pointers to keep track of first used and last used position on array.

During asynchronous memcpy, an event is selected from the array and the last pointer index is incremented % array size.  This event is then recorded on the same stream that the copy was queued on.

During progress engine phase, the first used (oldest) event is queried and if completion is detected, the first index will be incremented % array size and any copy completion operations will be performed.

wckzhang: Since it may be relevant to this conversation, I spent some time writing an API in Open MPI for accelerators, similar to the hmem APIs you seem to be indicating. It may be useful to reference it and can be found here: https://github.com/open-mpi/ompi/blob/main/opal/mca/accelerator/accelerator.h

It should be pretty easy to read as all the API's are documented with doxygen format.

amirshehataornl: > oesn't seem to align with cuda, or even your proposal here. Looking at the PR, it looks like you're converting the HSA model into the CUDA model. IMO, the HSA model is better. We can check for a specific copy to have completed. With the CUDA model, we only know if all of them ha

@shefty, Maybe there is a misunderstanding here. Unlike a cuda or hip stream, the way it's currently implemented is that each copy is queued asynchronously. The reason I group them together under the same "stream" is for the purpose of one operation which could have > 1 entries in the IOV.

For example if you try to do a async_copy_to(iovec) and the iovec.iov_len > 1, then each buffer in the IOV triggers an asynchronous copy. But the entire operation is not complete until all the buffers in the iov have been copied. 

So I'm not converting the HSA symantics into CUDA symantics.

If you have multiple calls to async_copy_to() then each one of these calls will result in a parallel operation, and will complete independently, once ROCR signals that the operation is complete.

From SHM, you'll have a set of queued operations waiting for completion. You traverse that list and query the operations, by calling the query API, and you complete the operations which are done.

So in essence, it basically works the same way as Open MPI.

What am I missing?

src/hmem_rocr.c
```diff
@@ -240,6 +394,160 @@ int rocr_copy_to_dev(uint64_t device, void *dest, const void *src,
...
+		}
+		ipc_signal->addr = dst;
+		agents[1] = agents[0];
+	}
+
+	ofi_hsa_signal_store_screlease(ipc_signal->sig, 1);
+
+	hsa_ret = ofi_hsa_amd_memory_async_copy(dst_hsa_ptr, agents[1],
+				src_hsa_ptr, agents[0],
+				size, 0, NULL, ipc_signal->sig);
```
shefty, line 483: The "event" associated with the async copy is ipc_signal->sig.  However, that's not what the 'stream' parameter returns.  'stream' is pointing to an array of these signals.  So, a single stream parameter actually references an arbitrary number of async copies.  This shows up as the for loop in the query call below.  This is why I'm saying that the behavior is cuda-like -- query only succeeds of all the copies have completed.

I think the general flow here is good.  But we should expose APIs for the caller to allocate and free whatever this 'event/barrier/signal/async_cntr/thingy' is, along with some typedef to enable type checking.  That's much easier for a developer to figure out how to use than having some void ** parameter act as input sometimes and output other times.

If we allow multiple async operations to be associated with this, that seems reasonable.  I don't think the changes would be too significant from what's here.

shefty:
I think it's important we get this abstraction 'right'.  I also don't know what 'right' is.  Stepping back, libfabric has async APIs, which follow this model:
```
transfer(..., void *context);
check_completion(..., void **context);
```
The caller provides a context, which is returned by cq_read, for example, when done.

Following this model would be the most expected path.  Why wouldn't this model work here?  What inefficiencies would result or how does it not align with the higher level flow?

wckzhang:
By the way, can we start adding better definitions to the hmem APIs so that it is clear what each one is supposed to do and what each parameter represents? From past experience I've had to read the implementations to figure out the intent of the APIs which is completely backward.

edgargabriel:
@shefty I have two thoughts to your comments (to which I btw agree with my limited knowledge of the libfabric APIs):

- I think @amirshehataornl solution is pretty close to what you suggest. Basically an async operation will return a 'new' stream object (i.e. an object from a pool of streams) for every operation, a stream containing either a single signal or multiple signals in the case of the iov interfaces.  It might be that it is just missing the proper abstraction for this object. 

- I am wondering whether it would help to see the code changes necessary to the shm provider to see how the async interfaces would  be incorporated there?

shefty:
The proposed 'stream' abstraction is different from the standard libfabric model as shown:

```
/* standard model */
async_op(op, IN void *context);
check_completion(OUT void **context);

/* proposal */
gpu_async(op, INOUT void **context);
check_completion(IN void *context);
```

With libfabric, the caller gives the context as *input* into the operation, and there's a 1:1 mapping between op and context.  This PR has the context of the operation given as *output*.  That context can then be reused for other operations, which then makes it act as *input*, but also results in a 1:n mapping between op and context.

The reason given for the 1:n mapping is to support iovec's.  I don't know how frequently that occurs in practice to make it worthwhile having this support as part of the abstraction.  The underlying implementation emulates this feature by tracking individual requests and not reporting a completion until all of them complete.  So, we're introducing overhead having this support, when there may be other ways of handling this.  Those mechanisms may be less efficient when dealing with an iovec, but better in the single buffer case.

As a general rule, having an API implicitly allocate and free an object that the caller is allowed to touch makes using that API safely difficult.  That's one problem I see in the current approach.  That's easily resolved by adding alloc/free calls, which the caller invokes directly.  That is what I outlined further up with the alloc/free event calls.  However, that proposal was only considering a 1:1 mapping between op and an event.

Note that I'm okay with a call like gpu_async() allocating some structure internally, which is freed when the operation completes.  But it needs to be kept internal and not given to the caller to touch.

amirshehataornl:
@shefty can you point me to an example in the code of 
```
/* standard model */
async_op(op, IN void *context);
check_completion(OUT void **context);
```

shefty:
I'm referring to the main libfabric API.  That's a conceptual example of the flow.  For example, it matches this flow:
```
fi_send(struct fid_ep *ep, const void *buf, size_t len, void *desc,
	fi_addr_t dest_addr, void *context);

fi_cq_read(struct fid_cq *cq, void *buf, size_t count)
```
and all other data transfers.  The first operation (send, read, write, tsend, recv, etc.) takes the context as input.  The second call returns that context as part of the completion.

amirshehataornl:
just to clarify, are you looking for a similar design, where the events which are completed get placed on a completion queue, and the caller is responsible for traversing this completion queue?

I had initially thought of doing it the same as the sar approach where you have a list you iterate over in a function like smr_progress_sar_list() which is called from smr_ep_progress()

The reason for this is the completion is actually done outside the context of libfabric. It's a ROCR signal which gets set when the operation completes. They don't call back into us. We have to check if it's set.

shefty:
I'm still trying to understand the trade-offs.

I'm saying that a 'CQ-like' approach aligns with the rest of the API.  It would be easy for another developer to look at APIs like those and figure out what they do and how to use them.  To me, that's a big advantage.  But the approach does have some drawbacks.  For one, the caller can't chain multiple async copies to the same event.

You also point out that the GPU APIs require us to check if an event has been set.  It looks like cuda and ze have the same constraint.

Maybe part of this is simply where the complexity resides, above or below these APIs.  But I can at least envision possible use cases for the above features.  And in that case, it leads me back to my original proposal of:
```
typedef void * ofi_hmem_async_event_t;

int (*create_async_event)(device, ofi_hmem_async_event_t *event);
int (*free_async_event)(device, ofi_hmem_async_event_t event);

int (*async_copy_to_hmem)(uint64_t device, void *dest, const void *src,
	    size_t size, ofi_hmem_async_event_t event);
int (*async_copy_from_hmem)(uint64_t device, void *dest,
	const void *src, size_t size, ofi_hmem_async_event_t event);
int (*async_query_event)(ofi_hmem_async_event_t event);
```
which is very close to what you proposed, only the event is allocated and freed explicitly by the caller, versus being hidden behind a void * context field.

aingerson:
These changes look ok to me now. Only outstanding comment I see is Sean's comment about the new base length not being passed into ze correctly. Ideally, change it to directly call zeGetIpcMemHandle, but I'm also not too worried about it because the value is ignored and also because we will be changing the shared handle implementation soon so this won't be an issue in the future. @amirshehataornl If you want to change it to make Sean happy then great. @shefty If you're fine with grumbling under your breath for a while, I think this is good to go.

shefty:
Sean (under breath): *grumble* *grumble* *grumble*

shefty:
@amirshehataornl - please see windows build failures.  Looks okay to merge once CI passes.

amirshehataornl:
@shefty is there a way to rerun the continous-integration step? Don't think it would be related to this PR.

shefty:
Restarted CI run.  There were a number of failures with the shm provider.

PR #8200, opened 2022-11-01
Add Support for XPMEM

amirshehataornl:
Add support for XPMEM the same way as other HMEM entries.
XPMEM makes use of the IPC caching to avoid repeated calls to xpmem_attach()/xpmem_detach()

### DIFF
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -58,8 +58,8 @@ struct ofi_mr_info {
 	uint64_t device;
 
 	uint64_t peer_id;
-	void     *ipc_mapped_addr;
-	uint8_t  ipc_handle[MAX_IPC_HANDLE_SIZE];
+	void     *mapped_addr;
+	uint8_t  handle[MAX_MR_HANDLE_SIZE];
 };
 
 
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -232,6 +232,7 @@ extern struct ofi_mem_monitor *rocr_monitor;
 extern struct ofi_mem_monitor *rocr_ipc_monitor;
 extern struct ofi_mem_monitor *ze_monitor;
 extern struct ofi_mem_monitor *import_monitor;
+extern struct ofi_mem_monitor *xpmem_monitor;
 
 /*
  * Used to store registered memory regions into a lookup map.  This
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -329,6 +329,7 @@ extern struct ofi_mr_cache_params	cache_params;
 
 struct ofi_mr_cache {
 	struct util_domain		*domain;
+	const struct fi_provider	*prov;
 	struct ofi_mem_monitor		*monitors[OFI_HMEM_MAX];
 	struct dlist_entry		notify_entries[OFI_HMEM_MAX];
 	size_t				entry_data_size;
--- a/include/ofi_shm_p2p.h
+++ b/include/ofi_shm_p2p.h
@@ -35,6 +35,7 @@
 #endif
 
 #include <ofi_cma.h>
+#include <ofi_xpmem.h>
 #include <ofi.h>
 #include <ofi_iov.h>
 
@@ -77,9 +78,9 @@ ofi_shm_p2p_no_copy(struct iovec *local, unsigned long local_cnt,
 static struct ofi_shm_p2p_ops p2p_ops[] = {
 	[FI_SHM_P2P_XPMEM] = {
 		.initialized = false,
-		.init = ofi_shm_p2p_no_init,
-		.cleanup = ofi_shm_p2p_no_cleanup,
-		.copy = ofi_shm_p2p_no_copy,
+		.init = xpmem_init,
+		.cleanup = xpmem_cleanup,
+		.copy = xpmem_copy,
 	},
 	[FI_SHM_P2P_CMA] = {
 		.initialized = false,
--- a/include/ofi_shm_p2p.h
+++ b/include/ofi_shm_p2p.h
@@ -1,0 +1,132 @@
+/*
+ * (C) Copyright 2023 UT-Battelle, LLC. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#if HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <ofi.h>
+#include <ofi_iov.h>
+
+#ifndef _OFI_SHM_P2P_H_
+#define _OFI_SHM_P2P_H_
+
+enum ofi_shm_p2p_type {
+	FI_SHM_P2P_XPMEM,
+	FI_SHM_P2P_CMA,
+	FI_SHM_P2P_DSA,
+};
+
+struct ofi_shm_p2p_ops {
+	bool initialized;
+	int (*init)(void);
+	int (*cleanup)(void);
+	int (*copy)(struct iovec *local, unsigned long local_cnt,
+		    struct iovec *remote, unsigned long remote_cnt,
+		    size_t total, pid_t pid, bool write, void *user_data);
+};
+
+static inline int ofi_shm_p2p_no_init(void)
+{
+	return FI_SUCCESS;
+}
+
+static inline int ofi_shm_p2p_no_cleanup(void)
+{
+	return FI_SUCCESS;
+}
+
+static inline int
+ofi_shm_p2p_no_copy(struct iovec *local, unsigned long local_cnt,
+		    struct iovec *remote, unsigned long remote_cnt,
+		    size_t total, pid_t pid, bool write, void *user_data)
+{
+	return -FI_ENOSYS;
+}
+
+static struct ofi_shm_p2p_ops p2p_ops[] = {
+	[FI_SHM_P2P_XPMEM] = {
+		.initialized = false,
+		.init = ofi_shm_p2p_no_init,
+		.cleanup = ofi_shm_p2p_no_cleanup,
+		.copy = ofi_shm_p2p_no_copy,
+	},
+	[FI_SHM_P2P_CMA] = {
+		.initialized = false,
+		.init = ofi_shm_p2p_no_init,
+		.cleanup = ofi_shm_p2p_no_cleanup,
+		.copy = ofi_shm_p2p_no_copy,
+	},
+	[FI_SHM_P2P_DSA] = {
+		.initialized = false,
+		.init = ofi_shm_p2p_no_init,
+		.cleanup = ofi_shm_p2p_no_cleanup,
+		.copy = ofi_shm_p2p_no_copy,
+	},
+};
+
+static inline void ofi_shm_p2p_init(void)
+{
+	int rc;
+	enum ofi_shm_p2p_type p2p_type;
+
+	for (p2p_type = 0; p2p_type < ARRAY_SIZE(p2p_ops); p2p_type++) {
+		rc = p2p_ops[p2p_type].init();
+		if (!rc)
+			p2p_ops[p2p_type].initialized = true;
+	}
+}
+
+static inline void ofi_shm_p2p_cleanup(void)
+{
+	int rc;
+	enum ofi_shm_p2p_type p2p_type;
+
+	for (p2p_type = 0; p2p_type < ARRAY_SIZE(p2p_ops); p2p_type++) {
+		rc = p2p_ops[p2p_type].cleanup();
+		if (!rc)
+			p2p_ops[p2p_type].initialized = false;
+	}
+}
+
+static inline int
+ofi_shm_p2p_copy(enum ofi_shm_p2p_type p2p_type, struct iovec *local,
+		 unsigned long local_cnt, struct iovec *remote,
+		 unsigned long remote_cnt, size_t total, int64_t id,
+		 bool write, void *user_data)
+{
+	return p2p_ops[p2p_type].copy(local, local_cnt, remote,
+				      remote_cnt, total, id, write, user_data);
+}
+
+#endif /* _OFI_SHM_P2P_H_ */
+
--- a/include/ofi_shm_p2p.h
+++ b/include/ofi_shm_p2p.h
@@ -34,6 +34,7 @@
 #include <config.h>
 #endif
 
+#include <ofi_cma.h>
 #include <ofi.h>
 #include <ofi_iov.h>
 
@@ -84,7 +85,7 @@ static struct ofi_shm_p2p_ops p2p_ops[] = {
 		.initialized = false,
 		.init = ofi_shm_p2p_no_init,
 		.cleanup = ofi_shm_p2p_no_cleanup,
-		.copy = ofi_shm_p2p_no_copy,
+		.copy = cma_copy,
 	},
 	[FI_SHM_P2P_DSA] = {
 		.initialized = false,
--- a/include/ofi_xpmem.h
+++ b/include/ofi_xpmem.h
@@ -1,0 +1,91 @@
+/*
+ * (C) Copyright 2023 UT-Battelle, LLC. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef OFI_XPMEM_H
+#define OFI_XPMEM_H
+
+#include <ofi_mr.h>
+#include <ofi_util.h>
+
+#if HAVE_XPMEM
+#include <xpmem.h>
+#else
+/* define XPMEM specific types to allow the code to
+ * compile on different platforms and to allow smr_region to compile
+ * without xpmem available.
+ */
+typedef int64_t xpmem_apid_t;
+typedef int64_t xpmem_segid_t;
+#endif /* HAVE_XPMEM */
+
+struct xpmem_client {
+	uint8_t cap;
+	xpmem_apid_t apid;
+	uintptr_t addr_max;
+};
+
+struct xpmem_pinfo {
+	/* XPMEM segment id for this process */
+	xpmem_segid_t seg_id;
+	/* maximum attachment address for this process. attempts to attach
+	 * past this value may fail. */
+	uintptr_t address_max;
+};
+
+struct xpmem {
+	struct xpmem_pinfo pinfo;
+	/* maximum size that will be used with a single memcpy call.
+	 * On some systems we see better peformance if we chunk the
+	 * copy into multiple memcpy calls. */
+	uint64_t memcpy_chunk_size;
+};
+
+extern struct xpmem *xpmem;
+
+int ofi_xpmem_cache_search(struct ofi_mr_cache *cache,
+			   struct iovec *iov, uint64_t peer_id,
+			   struct ofi_mr_entry **mr_entry,
+			   struct xpmem_client *xpmem);
+
+int ofi_xpmem_cache_open(struct ofi_mr_cache **cache);
+void ofi_xpmem_cache_destroy(struct ofi_mr_cache *cache);
+
+int xpmem_init(void);
+int xpmem_cleanup(void);
+int xpmem_copy(struct iovec *local, unsigned long local_cnt,
+	       struct iovec *remote, unsigned long remote_cnt,
+	       size_t total, pid_t pid, bool write, void *user_data);
+int ofi_xpmem_enable(struct xpmem_pinfo *peer,
+		     struct xpmem_client *xpmem);
+void ofi_xpmem_release(struct xpmem_client *xpmem);
+
+#endif /* OFI_XPMEM_H */
--- a/man/fi_shm.7.md
+++ b/man/fi_shm.7.md
@@ -168,6 +168,19 @@ The *shm* provider checks for the following environment variables:
   page fault is reported, so that there is valid address translation for the
   remaining addresses in the command. This minimizes DSA page faults. Default
   false
+
+*FI_SHM_USE_XPMEM*
+ : SHM can use SAR, CMA or XPMEM for host memory transfer. If
+   FI_SHM_USE_XPMEM is set to 1, the provider will select XPMEM over CMA if
+   XPMEM is available.  Otherwise, if neither CMA nor XPMEM are available
+   SHM shall default to the SAR protocol. Default 0
+
+*FI_XPMEM_MEMCPY_CHUNKSIZE*
+ :  The maximum size which will be used with a single memcpy call. XPMEM
+    copy performance improves when buffers are divided into smaller
+    chunks. This environment variable is provided to fine tune performance
+    on different systems. Default 262144
+
 # SEE ALSO
 
 [`fabric`(7)](fabric.7.html),
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -1341,8 +1341,8 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			return ret;
 
 		if (ep->util_ep.caps & FI_HMEM || smr_env.disable_cma) {
-			ep->region->cma_cap_peer = SMR_CMA_CAP_OFF;
-			ep->region->cma_cap_self = SMR_CMA_CAP_OFF;
+			ep->region->cma_cap_peer = SMR_VMA_CAP_OFF;
+			ep->region->cma_cap_self = SMR_VMA_CAP_OFF;
 			if (ep->util_ep.caps & FI_HMEM) {
 				if (ze_hmem_p2p_enabled())
 					smr_init_ipc_socket(ep);
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -42,6 +42,7 @@
 #include "smr_signal.h"
 #include "smr.h"
 #include "smr_dsa.h"
+#include "ofi_xpmem.h"
 
 extern struct fi_ops_msg smr_msg_ops, smr_no_recv_msg_ops;
 extern struct fi_ops_tagged smr_tag_ops, smr_no_recv_tag_ops;
@@ -573,7 +574,7 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 }
 
 int smr_select_proto(void **desc, size_t iov_count,
-                     bool cma_avail, uint32_t op, uint64_t total_len,
+                     bool vma_avail, uint32_t op, uint64_t total_len,
 		     uint64_t op_flags)
 {
 	struct ofi_mr *smr_desc;
@@ -599,7 +600,7 @@ int smr_select_proto(void **desc, size_t iov_count,
 	if (op == ofi_op_read_req) {
 		if (use_ipc)
 			return smr_src_ipc;
-		if (cma_avail && FI_HMEM_SYSTEM == iface)
+		if (vma_avail && FI_HMEM_SYSTEM == iface)
 			return smr_src_iov;
 		return smr_src_sar;
 	}
@@ -618,7 +619,7 @@ int smr_select_proto(void **desc, size_t iov_count,
 	if (use_ipc)
 		return smr_src_ipc;
 
-	if (total_len > SMR_INJECT_SIZE && cma_avail)
+	if (total_len > SMR_INJECT_SIZE && vma_avail)
 		return smr_src_iov;
 
 	if (op_flags & FI_DELIVERY_COMPLETE)
@@ -1387,6 +1388,13 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 		if (smr_env.use_dsa_sar)
 			smr_dsa_context_init(ep);
 
+		/* if XPMEM is on after exchanging peer info, then set the
+		 * endpoint p2p to XPMEM so it can be used on the fast
+		 * path
+		 */
+		if (ep->region->xpmem_cap_self == SMR_VMA_CAP_ON)
+			ep->p2p_type = FI_SHM_P2P_XPMEM;
+
 		break;
 	default:
 		return -FI_ENOSYS;
@@ -1506,6 +1514,9 @@ int smr_endpoint(struct fid_domain *domain, struct fi_info *info,
 	ep->util_ep.ep_fid.atomic = &smr_atomic_ops;
 
 	*ep_fid = &ep->util_ep.ep_fid;
+
+	/* default to CMA for p2p */
+	ep->p2p_type = FI_SHM_P2P_CMA;
 	return 0;
 ep:
 	ofi_endpoint_close(&ep->util_ep);
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -308,8 +308,8 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
 			    size_t iov_count, size_t *total_len,
 			    struct smr_ep *ep, int err)
 {
-	enum ofi_shm_p2p_type p2p_type = FI_SHM_P2P_CMA;
 	struct smr_region *peer_smr;
+	struct xpmem_client *xpmem;
 	struct smr_resp *resp;
 	int ret;
 
@@ -321,10 +321,12 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
 		goto out;
 	}
 
-	ret = ofi_shm_p2p_copy(p2p_type, iov, iov_count, cmd->msg.data.iov,
+	xpmem = &smr_peer_data(ep->region)[cmd->msg.hdr.id].xpmem;
+
+	ret = ofi_shm_p2p_copy(ep->p2p_type, iov, iov_count, cmd->msg.data.iov,
 			       cmd->msg.data.iov_count, cmd->msg.hdr.size,
 			       peer_smr->pid, cmd->msg.hdr.op == ofi_op_read_req,
-			       NULL);
+			       xpmem);
 	if (!ret)
 		*total_len = cmd->msg.hdr.size;
 
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -38,6 +38,7 @@
 #include "ofi_hmem.h"
 #include "ofi_atom.h"
 #include "ofi_mr.h"
+#include "ofi_shm_p2p.h"
 #include "smr.h"
 #include "smr_dsa.h"
 
@@ -307,6 +308,7 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
 			    size_t iov_count, size_t *total_len,
 			    struct smr_ep *ep, int err)
 {
+	enum ofi_shm_p2p_type p2p_type = FI_SHM_P2P_CMA;
 	struct smr_region *peer_smr;
 	struct smr_resp *resp;
 	int ret;
@@ -319,9 +321,10 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
 		goto out;
 	}
 
-	ret = smr_cma_loop(peer_smr->pid, iov, iov_count, cmd->msg.data.iov,
-			   cmd->msg.data.iov_count, 0, cmd->msg.hdr.size,
-			   cmd->msg.hdr.op == ofi_op_read_req);
+	ret = ofi_shm_p2p_copy(p2p_type, iov, iov_count, cmd->msg.data.iov,
+			       cmd->msg.data.iov_count, cmd->msg.hdr.size,
+			       peer_smr->pid, cmd->msg.hdr.op == ofi_op_read_req,
+			       NULL);
 	if (!ret)
 		*total_len = cmd->msg.hdr.size;
 
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -565,7 +565,7 @@ static struct smr_pend_entry *smr_progress_ipc(struct smr_cmd *cmd,
 	if (cmd->msg.data.ipc_info.iface == FI_HMEM_ZE)
 		ptr = (char *) base + (uintptr_t) cmd->msg.data.ipc_info.offset;
 	else
-		ptr = (char *) (uintptr_t) mr_entry->info.ipc_mapped_addr +
+		ptr = (char *) (uintptr_t) mr_entry->info.mapped_addr +
 		      (uintptr_t) cmd->msg.data.ipc_info.offset;
 
 	if (cmd->msg.data.ipc_info.iface == FI_HMEM_ROCR) {
--- a/prov/shm/src/smr_rma.c
+++ b/prov/shm/src/smr_rma.c
@@ -125,7 +125,7 @@ static ssize_t smr_generic_rma(struct smr_ep *ep, const struct iovec *iov,
 
 	cmds = 1 + !(domain->fast_rma && !(op_flags &
 		    (FI_REMOTE_CQ_DATA | FI_DELIVERY_COMPLETE)) &&
-		     rma_count == 1 && smr_cma_enabled(ep, peer_smr));
+		     rma_count == 1 && smr_vma_enabled(ep, peer_smr));
 
 	if (smr_peer_data(ep->region)[id].sar_status)
 		return -FI_EAGAIN;
@@ -167,7 +167,7 @@ static ssize_t smr_generic_rma(struct smr_ep *ep, const struct iovec *iov,
 	total_len = ofi_total_iov_len(iov, iov_count);
 	assert(!(op_flags & FI_INJECT) || total_len <= SMR_INJECT_SIZE);
 
-	proto = smr_select_proto(desc, iov_count, smr_cma_enabled(ep, peer_smr),
+	proto = smr_select_proto(desc, iov_count, smr_vma_enabled(ep, peer_smr),
 	                         op, total_len, op_flags);
 
 	ret = smr_proto_ops[proto](ep, peer_smr, id, peer_id, op, 0, data,
@@ -329,7 +329,7 @@ static ssize_t smr_generic_rma_inject(struct fid_ep *ep_fid, const void *buf,
 	peer_smr = smr_peer_region(ep->region, id);
 
 	cmds = 1 + !(domain->fast_rma && !(flags & FI_REMOTE_CQ_DATA) &&
-		     smr_cma_enabled(ep, peer_smr));
+		     smr_vma_enabled(ep, peer_smr));
 
 	if (smr_peer_data(ep->region)[id].sar_status)
 		return -FI_EAGAIN;
--- a/prov/shm/src/smr_rma.c
+++ b/prov/shm/src/smr_rma.c
@@ -36,6 +36,7 @@
 #include <sys/uio.h>
 
 #include "ofi_iov.h"
+#include "ofi_shm_p2p.h"
 #include "smr.h"
 
 static void smr_add_rma_cmd(struct smr_region *peer_smr,
@@ -60,6 +61,7 @@ static ssize_t smr_rma_fast(struct smr_region *peer_smr, const struct iovec *iov
 			uint32_t op, uint64_t op_flags)
 {
 	struct iovec cma_iovec[SMR_IOV_LIMIT], rma_iovec[SMR_IOV_LIMIT];
+	enum ofi_shm_p2p_type p2p_type = FI_SHM_P2P_CMA;
 	struct smr_cmd_entry *ce;
 	size_t total_len;
 	int ret, i;
@@ -77,8 +79,9 @@ static ssize_t smr_rma_fast(struct smr_region *peer_smr, const struct iovec *iov
 
 	total_len = ofi_total_iov_len(iov, iov_count);
 
-	ret = smr_cma_loop(peer_smr->pid, cma_iovec, iov_count, rma_iovec,
-			   rma_count, 0, total_len, op == ofi_op_write);
+	ret = ofi_shm_p2p_copy(p2p_type, rma_iovec, iov_count,
+			       rma_iovec, rma_count, total_len, peer_smr->pid,
+			       op == ofi_op_write, NULL);
 
 	if (ret) {
 		smr_cmd_queue_discard(ce, pos);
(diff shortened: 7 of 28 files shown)


src/hmem_xpmem.c
```diff
@@ -0,0 +1,237 @@
...
+int xpmem_close_handle(void *ipc_ptr)
+{
+	return xpmem_detach(ipc_ptr);
+}
+
+static inline void xpmem_memove(void *dst, void *src, size_t size)
+{
+	while (size > 0) {
+		size_t copy_size = MIN(size, xpmem->memcpy_chunk_size);
+		memcpy(dst, src, copy_size);
```
shefty, line 167: Do you need to use memmove here instead of memcpy?  memcpy doesn't handle overlapping regions, but xpmem_memove (can we add the missing 'm' too) suggests that it aligns with memmove, not memcpy

include/rdma/fi_domain.h
```diff
@@ -128,6 +128,7 @@ enum fi_hmem_iface {
 	FI_HMEM_ZE,
 	FI_HMEM_NEURON,
 	FI_HMEM_SYNAPSEAI,
+	FI_HMEM_XPMEM,
```
shefty, line 131: Hmm... does xpmem count as an hmem interface?  I thought it was a different way to setup shared memory transfers.

The approach here looks clean, and this may be an easy way to add support.

shefty (changes requested):
See the windows build failures.  You'll need to add the missing files to the windows resource files.

include/rdma/fi_domain.h
amirshehataornl: I was kinda struggling with the same question. The way I convinced myself is "FI_HMEM_SYSTEM", which is the default if you don't have HMEM. XPMEM also needs to use the the IPC Caching (which I was pondering creating a patch to just rename it to caching and remove the IPC, since it can be used for other purposes as well). Also this allows other providers to use it. For example, GNI has its own XPMEM implementation. So if GNI doesn't endup being removed in 2.0, it can theoretically use this infra. Over all this approach seems like it has precedence. It also fits nicely in the entire architecture and the changes in SHM provider to add support is fairly straight forward, same path as CMA.

shefty: @aingerson - The changes look decent.  Any concerns about this approach?  Could a similar approach be used for CMA or DSA?

wzamazon:
@amirshehataornl  This patch looks very interesting. I guess you have another patch that uses the XPMEM interface in shm provider?

amirshehataornl:
yes. I'll be updating this patch with the changes Sean requested and upload the SHM patch soon.

include/rdma/fi_domain.h
aingerson: I agree, the changes themselves look decent. The thing I struggle with if we put it in the HMEM category is that it blurs the lines for how we define HMEM. The target for defining these has been both to indicate the API to use as well as the location of the memory. We've also defined FI_MR_HMEM as needing to register any of the interfaces in this enum. So even though XPMEM matches the op structure nicely, I'm not sure it's the best grouping. At this point, now we're looking at 3 different methods just for copying FI_HMEM_SYSTEM memory (CMA, DSA, XPMEM) so I think we could consider defining a separate (internal) method of outlining these different methods together that doesn't confuse the HMEM interfaces.

shefty: This is a good point.  We're modifying the user-visible API because it's convenient for the internal implementation.  But is XPMEM something the app should be dealing with directly?

amirshehataornl: So the idea here is to have a shared memory level callback grouping similar to HMEM, and add XPMEM there?
I do think though that this shouldn't be specific to shared memory but make the infrastructure available to other providers for use. Thoughts on that?

aingerson: Yes, agreed. I think we can make it internal and not expose it to the user API but add it as an ofi_<>.h header so anyone can use it. That way gni can use them for its xpmem implementation and any other provider that wants to take advantage of xpmem or DSA can do so as well.

amirshehataornl: sounds good. I'll update the patch set

amirshehataornl:
Just an update for whoever is interested. I'm working through some issues with this patch and will update it as soon as I have a verified fix.

amirshehataornl:
@shefty, @aingerson, This is a rebase + a change to address the API concern mentioned here: https://github.com/ofiwg/libfabric/pull/8200#discussion_r1013518342

I've been thinking about the best way to satisfy the requirement of not exposing XPMEM to the application, while still using the hmem infrastructure without any code duplication.

My proposed solution is explained in one of the commit comments, listed below as well:

```
    Add a new concept: Internal HMEM types. These HMEM types are only visible
    within the libfabric library.
    
    fi_hmem_iface is part of the libfabric external API.
    
    Internal ways of transferring data should only be accessible by the
    libfabric providers, and should not be visible to the libfabric users.
    XPMEM falls into that category.
    
    However, XPMEM can be managed in the same way as the other HMEM
    interfaces. It should also be available to providers to use as they see
    fit.
    
    Instead of duplicating the HMEM infrastructure, introduce the concept of
    internal HMEM types. These are memory types which are not visible to the
    users of libfabric API, but remain visible to the providers. These types
    are coded as an enum which start after the last entry in the fi_hmem_iface
    enumeration.
    
    This approach achieves two purposes:
      1. Does not change the external API
      2. Allows for reusing the HMEM infrastructure
```
thoughts?

aingerson:
@amirshehataornl I think I would prefer if these were completed separated from the hmem types because I think it could be problematic cross versions since the internal values would change whenever new HMEM ifaces are added. This means that if a provider were to use an internal iface enum in its protocol (for example shm lets the peer know it needs to open the XPMEM memory), it would need to increment its internal protocol whenever a new HMEM iface is added.

src/hmem_xpmem.c
```diff
@@ -0,0 +1,237 @@
...
+		     size_t size)
+{
+	xpmem_memcpy(dst, (void *)src, size);
+	return 0;
+}
+
+bool xpmem_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags)
+{
+	return false;
+}
```
shefty, line 191: Always returning false seems odd.  How would the xpmem path ever be taken?

include/ofi_hmem.h
```diff
@@ -67,6 +67,10 @@ struct ipc_info {
 	uint8_t		ipc_handle[MAX_IPC_HANDLE_SIZE];
 };
 
+enum ofi_internal_hmem_iface {
+	OFI_HMEM_XPMEM = sizeof(enum fi_hmem_iface),
```
shefty, line 71: This isn't correct.  It likely returns 4 (bytes).

src/hmem_xpmem.c
amirshehataornl: To follow up on the Nov 29, 2022 meeting discussion:

I would argue that XPMEM is another variation of the HMEM_SYSTEM. 
In fact we could have something like HMEM_SYSTEM_XPMEM, HMEM_SYSTEM_CMA, etc
The HMEM_SYSTEM option behaves in the same way as XPMEM with regards to the is_addr_valid() API. In fact it doesn't define it at all, because it assumes it'll never be called. It's explicitly checked in ofi_get_hmem_iface(). It also sets a bunch of the other callbacks to the ofi_no_...() version.

It seems that the entire infrastructure was created with the assumption that not all interfaces will support all the different callbacks. So it's normal to have an interface not support some of the callbacks.

If we look at SHM usage, when it selects the iface with smr_get_mr_hmem_iface() it can return FI_HMEM_SYSTEM.
This could result in both peers using the FI_HMEM_SYSTEM for H2H runs. In that case it'll go through the HMEM API. Thinking about it some more here, it's possible to return XPMEM if XPMEM is enabled and the user indicated they want to use it.

Given the way it's currently used, it makes the argument that XPMEM option fits within this infrastructure, with the limitation of having it not exposed to the external API.

With regards to the is_addr_valid() call, I do think it's a bug to just return false. I think there is a way to verify if the pointer is in the area mapped by XPMEM (have to investigate).

The point I'm trying to make here is that the reason I'm not convinced of separating XPMEM from HMEM, is not only because of code duplication, but because I believe it's a fit within the existing infrastructure for the reasons I mentioned above

amirshehataornl: @shefty, @aingerson 
Just adding tags to make sure notifications are generated for my above comment.

aingerson:
@amirshehataornl I don't disagree that it could be smoother to integrate it, but I think your comment that xpmem is another variation of hmem_system is where my issue is. If it's another sub-type of hmem_system, then it doesn't really make sense to have it as a separate element. That's where I get hung up on.

In the end, I think the proper solution is whatever makes the most sense for the target use case (ie shm using xpmem) so that's where I'm trying to approach it from. I agree that separating the hmem_ops from host_ops make that more complicated so I don't necessarily think that's the best solution. However, I think adding xpmem as a distinct memory type doesn't make much sense either.
I'm trying to think of a good middle ground. For example, could we simply add the xpmem support to the hmem_system definition if it's available? Would that satisfy both of our issues.
Could you elaborate more on how shm would use this? I'm a little confused on how xpmem fits into the IPC ops if you don't have an op for get_handle? If there is no handle to get and send to a peer, how does xpmem fit more into the IPC protocol than the CMA protocol (ie just send an address over and the peer would call xpmem_attach/xpmem_memcpy)?

shefty:
My first concern with xpmem is whether it is safe to cache calls to xpmem_attach().  In all other situations where we cache memory addresses, we require some sort of 'memory monitor' to validate that a cached entry is still valid.  Otherwise, changes to the virtual to physical address mapping results in what appears to be memory corruption at the application level, which is extremely difficult to debug.  I tried tracing through the kernel code for xpmem_attach() (because HA! xpmem documentation? surely, you jest), and couldn't figure out if caching works.

As for adding xpmem support, I think the following sort of implementation makes the most sense:
```
struct ofi_xpmem {
    struct ofi_mr_cache cache; /* if caching is permissible */
};

int ofi_xpmem_init(struct ofi_xpmem *xpmem);
void ofi_xpmem_cleanup(struct ofi_xpmem *xpmem);
int ofi_xpmem_copy(struct ofi_xpmem *xpmem, void *src, void *dest, size_t size);
```
AFAICT, this is really all that's needed.  XPMEM is basically a process-to-process memory copy abstraction, not a device/HMEM abstraction.  The shm provider would have 3 different p2p memory copy abstractions: CMA, DSA, and XPMEM.  These all seem to follow a similar model.  Along these lines, we can create a p2p memory copy abstraction:

```
struct ofi_p2p_memcpy {
    int (*init)(struct ofi_p2pmem *p2p_memcpy);
    void (*cleanup)(struct ofi_p2pmem *p2p_memcpy);
    int (*copy)(struct ofi_p2pmem *p2p_memcpy, void *src, void *dest, size_t size);
};
```
(modeling after ofi_monitor) and update ofi_xpmem as:
```
struct ofi_xpmem {
    struct ofi_p2p_memcpy p2p_memcpy;
    struct ofi_mr_cache cache;
};
```
CMA and DSA could have similar definitions.  The shm provider could set a default p2p memcpy based on the configuration, or pick one dynamically for each peer/transfer.  E.g.
```
smr.h:

struct ofi_p2p_memcpy *def_p2p_memcpy;

smr_progress.c:

static int smr_progress_iov(...)
{
    resp = smr_get_ptr(peer_smr, ...);

    def_p2p_memcpy(...);
}
```
Note that this is incomplete.  The copy function will need to identify the peer, and it may be better to take an iov as input instead of 2 void * addresses.  Or extend with other copy routines.

This abstraction could be expanded to support implementations with asynchronous memory copies.

It may be possible to add the mmap copy routine into this... maybe.  I don't see a benefit though.

a-szegel:
What is the performance increase with large message sizes that you have seen with XPMEM?

amirshehataornl:
> What is the performance increase with large message sizes that you have seen with XPMEM?

about 1.68 times increase. On our system with 1M and 16M messages it translates into a boost of ~25GB/s

But note this is run on a cray system with cray xpmem.

I'm going to be swinging back to this patch to address Sean's comments.

shefty:
Please remove updates requested tag when ready for another round of review.

amirshehataornl:
@shefty, @aingerson Can I get an initial look at the updates for this patch series. It included p2p infrastructure we talked about and the XPMEM support which uses the p2p infra.

vidsouza:
bot:aws:retest

prov/shm/src/smr_rma.c
```diff
@@ -71,16 +72,34 @@ static ssize_t smr_rma_fast(struct smr_region *peer_smr, const struct iovec *iov
...
+				iov_count, rma_iovec, rma_count, total_len, peer_id,
+				op == ofi_op_write, xpmem);
+	} else {
+		ret = smr_cma_loop(peer_smr->pid, vma_iovec, iov_count, rma_iovec,
+				   rma_count, 0, total_len, op == ofi_op_write);
+	}
+#else
+	ret = smr_cma_loop(peer_smr->pid, vma_iovec, iov_count, rma_iovec,
 			   rma_count, 0, total_len, op == ofi_op_write);
+#endif /* HAVE_XPMEM */
```
shefty, line 102: Same comment as above.  Note that the 'else' and '#else' paths are identical.  They should follow the same path.

Larger question is why ofi_shm_p2p_copy() isn't configured correctly to just make the correct call?

prov/util/src/util_shm.c
```diff
@@ -43,6 +43,10 @@
 
 #include <ofi_shm.h>
 
+#if HAVE_XPMEM
+#include <ofi_xpmem.h>
+#endif
```
shefty, line 48: #if shouldn't be needed.  File should always exist.

prov/util/src/util_shm.c
```diff
@@ -86,9 +94,9 @@ void smr_cma_check(struct smr_region *smr, struct smr_region *peer_smr)
 	assert(remote_pid == peer_smr->pid);
 
 	if (smr == peer_smr) {
-		smr->cma_cap_self = (ret == -1) ? SMR_CMA_CAP_OFF : SMR_CMA_CAP_ON;
+		smr->cma_cap_self = (ret == -1) ? SMR_VMA_CAP_OFF : SMR_VMA_CAP_ON;
 	} else {
-		smr->cma_cap_peer = (ret == -1) ? SMR_CMA_CAP_OFF : SMR_CMA_CAP_ON;
+		smr->cma_cap_peer = (ret == -1) ? SMR_VMA_CAP_OFF : SMR_VMA_CAP_ON;
```
shefty, line 99: Randomly picking a spot for this comment, but the renames should go into a separate patch.  That would make it easier to review the xpmem changes.

prov/shm/src/smr_ep.c
```diff
@@ -1322,14 +1326,23 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
...
+			ep->region->cma_cap_self = SMR_VMA_CAP_OFF;
 			if (ep->util_ep.caps & FI_HMEM) {
 				if (ze_hmem_p2p_enabled())
 					smr_init_ipc_socket(ep);
 			}
 		}
 
+#if HAVE_XPMEM
+		if (smr_env.use_xpmem && xpmem)
+			ep->region->xpmem_cap_self = SMR_VMA_CAP_ON;
```
aingerson, line 1339: Does xpmem have any similar ptrace restrictions we need to check for before enabling?

aingerson (commented):
I agree with Sean about reducing the #if/#else code for readability and about separating out the naming. I would also say separating the XPMEM code in a separate patch (separate from plugging it into shm) would be helpful.

amirshehataornl:
@aingerson, @shefty can I get another review on these updates.

amirshehataornl:
I can't access the failure links. Can someone tell me what has gone wrong.

aingerson:
@amirshehataornl Intel CI failure is a ucx error that has been popping up. I don't think it's related. I'm going to review today btw.

include/ofi_shm.h
```diff
@@ -84,11 +84,12 @@ enum {
...
-/* CMA capability */
+/* CMA/XPMEM capability. Generic acronym used:
+ * VMA: Virtual Memory Address */
 enum {
-	SMR_CMA_CAP_NA,
-	SMR_CMA_CAP_ON,
-	SMR_CMA_CAP_OFF,
+	SMR_VMA_CAP_NA,
+	SMR_VMA_CAP_ON,
+	SMR_VMA_CAP_OFF,
```
shefty, line 92: Can someone explain why this has 3 values, versus just 2?  What's the difference between OFF and NA?

This looks like something that should be a flag, rather than an enum.

(Comment isn't directly related to this PR).

prov/shm/src/smr_ep.c
```diff
@@ -1329,6 +1333,15 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
...
+		else
+			ep->region->xpmem_cap_self = SMR_VMA_CAP_OFF;
+#else
+		ep->region->xpmem_cap_self = SMR_VMA_CAP_OFF;
+#endif
```
shefty, line 1343: The ifdef's here shouldn't be needed.

We should also be able to determine if xpmem is enabled by checking a single variable, not both smr_env.use_xpmem and xpmem.  Ideally, those two values should be aligned.  Are they not?

prov/shm/src/smr_progress.c
```diff
@@ -319,9 +321,26 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
 		goto out;
 	}
 
-	ret = smr_cma_loop(peer_smr->pid, iov, iov_count, cmd->msg.data.iov,
-			   cmd->msg.data.iov_count, 0, cmd->msg.hdr.size,
-			   cmd->msg.hdr.op == ofi_op_read_req);
+	/* TODO: Once cma is migrated to the new shm p2p infrastructure,
+	 * we should be able to call ofi_shm_p2p_copy() with the
+	 * appropriate p2p_type, instead of having a separate function
+	 * call for cma
```
shefty: Why not convert cma to use the new p2p infrastructure in a patch immediately following the patch that adds p2p?  If I'm following the this, p2p is defined, suggests that it supports cma, but doesn't actually do that.  Anyone jumping into the code at this point will be confused.

If CMA isn't being added, I'd comment out the CMA p2p_type until it is added to make it clear that it's not yet there.

prov/util/src/util_shm.c
```diff
@@ -284,6 +285,15 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
...
 	(*smr)->cma_cap_self = SMR_VMA_CAP_NA;
+#if HAVE_XPMEM
+	(*smr)->xpmem_cap_self = (xpmem) ? SMR_VMA_CAP_ON
+		: SMR_VMA_CAP_OFF;
+	if (xpmem)
+		(*smr)->xpmem_self = xpmem->pinfo;
+	else
+		memset(&(*smr)->xpmem_self, 0,
+		       sizeof((*smr)->xpmem_self));
+#endif
```
shefty, line 296: The ifdef shouldn't be needed.

prov/util/src/util_shm.c
```diff
@@ -466,6 +477,27 @@ void smr_map_to_endpoint(struct smr_region *region, int64_t id)
...
+			return;
+		}
+		local_peers[id].xpmem.cap = SMR_VMA_CAP_ON;
+		local_peers[id].xpmem.addr_max = peer_smr->xpmem_self.address_max;
+	} else {
+		local_peers[id].xpmem.cap = SMR_VMA_CAP_OFF;
+	}
+#else
+	region->xpmem_cap_self = SMR_VMA_CAP_OFF;
+#endif
```
shefty, line 498: Please move this section to another function, so that the ifdef can wrap the entire function, rather than being sprinkled with a function.

prov/util/src/util_shm.c
```diff
@@ -486,6 +518,10 @@ void smr_unmap_from_endpoint(struct smr_region *region, int64_t id)
 
 	peer_peers[peer_id].addr.id = -1;
 	peer_peers[peer_id].name_sent = 0;
+
+#if HAVE_XPMEM
+	xpmem_release(local_peers[peer_id].xpmem.apid);
+#endif
```
shefty, line 524: Please look for ways to remove ifdef's, particularly in .c files.  In this case, we can define xpmem_release() as a no-op macro in a header file if HAVE_XPMEM is not defined.

src/hmem_ipc_cache.c
```diff
@@ -38,9 +38,9 @@ static int ipc_cache_add_region(struct ofi_mr_cache *cache, struct ofi_mr_entry
 {
 	int ret;
 
-	ret = ofi_hmem_open_handle(entry->info.iface, (void **)&entry->info.ipc_handle,
+	ret = ofi_hmem_open_handle(entry->info.iface, (void **)&entry->info.handle,
 				   entry->info.iov.iov_len, entry->info.device,
-				   &entry->info.ipc_mapped_addr);
+				   &entry->info.mapped_addr);
```
shefty, line 43: nit: There are a couple of renames in this patch which would be easier to deal with as a separate patch.

include/ofi_shm_p2p.h
```diff
@@ -88,7 +92,11 @@ static struct ofi_shm_p2p_ops p2p_ops[] = {
 		.initialized = false,
 		.init = ofi_shm_p2p_no_init,
 		.cleanup = ofi_shm_p2p_no_cleanup,
+#ifdef _WIN32
 		.copy = ofi_shm_p2p_no_copy,
+#else
+		.copy = cma_copy,
+#endif
```
shefty, line 99: Let's push the _WIN32 checks down.  ofi_cma.h should be includable in any OS.  There's barely any code there.  We could even argue whether cma_copy() could go into ofi_mem.h or some other existing header file.  pid_t is already defined in windows/osd.h.  I think we just need to define the ofi_process_vm_read/writev calls.  Those are defined in the freebsd and osx osd.h files, just not windows.

shefty (commented):
Thanks!  Looking good.  I have a few comments, but nothing major.  FYI, my comments were added based on reviewing the patches one at a time, so a subsequent patch may have addressed a couple of comments.

include/ofi_xpmem.h
```diff
@@ -0,0 +1,90 @@
...
+#include <ofi_mr.h>
+#include <ofi_util.h>
+
+#if HAVE_XPMEM
+#include <xpmem.h>
+
+struct xpmem_client {
+	uint8_t cap;
+	xpmem_apid_t apid;
+	uintptr_t addr_max;
```
aingerson, line 45: if you remove the ifdef in the smr_region as suggested above, the difference in sizes for the xpmem_client for HAVE_XPMEM and !HAVE_XPMEM is going to also be problematic. will probably need to adjust the definition here.

prov/shm/src/smr_ep.c
```diff
@@ -43,6 +43,10 @@
 #include "smr.h"
 #include "smr_dsa.h"
 
+#if HAVE_XPMEM
+#include "ofi_xpmem.h"
+#endif
```
aingerson, line 48: the ofi_xpmem header already have a check within it so you shouldn't need to exclude the include here

prov/shm/src/smr_progress.c
```diff
@@ -319,9 +321,26 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
...
+	 */
+	xpmem = &smr_peer_data(ep->region)[cmd->msg.hdr.id].xpmem;
+	if (ep->region->xpmem_cap_self == SMR_VMA_CAP_ON &&
+	    xpmem->cap == SMR_VMA_CAP_ON) {
```
aingerson, line 333: why are we checking this on the receive/progress side? shouldn't the sender be checking these and only sending if xpmem is available on the peer?

prov/shm/src/smr_rma.c
```diff
@@ -69,16 +72,27 @@ static ssize_t smr_rma_fast(struct smr_region *peer_smr, const struct iovec *iov
...
 	for (i = 0; i < rma_count; i++) {
 		rma_iovec[i].iov_base = (void *) rma_iov[i].addr;
 		rma_iovec[i].iov_len = rma_iov[i].len;
 	}
 
 	total_len = ofi_total_iov_len(iov, iov_count);
 
-	ret = smr_cma_loop(peer_smr->pid, cma_iovec, iov_count, rma_iovec,
-			   rma_count, 0, total_len, op == ofi_op_write);
+	xpmem = &smr_peer_data(ep->region)[peer_id].xpmem;
```
aingerson, line 89: I believe this peer id is the one for the remote peer, not the local side. you'll need to pass in id in addition

include/ofi_mr.h
```diff
@@ -58,8 +58,8 @@ struct ofi_mr_info {
 	uint64_t device;
 
 	uint64_t peer_id;
-	void     *ipc_mapped_addr;
-	uint8_t  ipc_handle[MAX_IPC_HANDLE_SIZE];
+	void     *mapped_addr;
+	uint8_t  handle[MAX_MR_HANDLE_SIZE];
```
aingerson, line 62: these renames would be helpful in a different patch

src/xpmem.c
```diff
@@ -0,0 +1,219 @@
...
+	if (xpmem)
+		ret = xpmem_remove(xpmem->pinfo.seg_id);
+
+	return (ret) ? -FI_EINVAL : ret;
+}
+
+static inline void xpmem_memcpy(void *dst, void *src, size_t size)
+{
+	while (size > 0) {
+		size_t copy_size = MIN(size, xpmem->memcpy_chunk_size);
```
aingerson, line 145: move variable declaration out of loop

src/xpmem.c
```diff
@@ -0,0 +1,219 @@
...
+		memcpy(dst, src, copy_size);
+		dst = (void *) ((uintptr_t) dst + copy_size);
+		src = (void *) ((uintptr_t) src + copy_size);
+		size -= copy_size;
+	}
+}
+
+int xpmem_copy(struct ofi_mr_cache *cache, struct iovec *local,
```
aingerson, line 153: it feels like this cache could be internal to the xpmem code instead of having to pass it into the copy function. from what I can tell, it gets initialized and destroyed by shm but all other searches/cache delete calls are done inside xpmem

src/xpmem.c
```diff
@@ -0,0 +1,219 @@
...
+#endif
+
+#include <rdma/fi_errno.h>
+#include "ofi_mem.h"
+#include "ofi_mr.h"
+#include "ofi.h"
+#include <stdio.h>
+
+#if HAVE_XPMEM
+#include <ofi_xpmem.h>
```
aingerson, line 44: to be consistent with the other files, move the include with the other includes

aingerson (commented):
A few things here and there but looks mostly good. My biggest thing, which Sean has already mentioned, is that I would like to see all the #if HAVE_XPMEM blocks get dropped. This helps with code readability and also makes sure we don't have any discrepancies in the shm region because some fields don't exist.

prov/shm/src/smr_rma.c
amirshehataornl: I'm not sure I follow the comment. I want to get the xpmem structure for the remote peer. Can you clarify what you mean by "pass in id in addition"? I don't think I need the local id in this code path.

prov/shm/src/smr_progress.c
amirshehataornl: We'd need to do that check anyway to make the proper call to copy using XPMEM or otherwise. Am I missing something?
Basically the address space of a process is exposed via XPMEM. A pointer to the address is passed over to the receiving process. the receiving process figures that it and the sender are using XPMEM and then makes the appropriate call into p2p infra.

src/xpmem.c
amirshehataornl: When you receive a request to transfer memory using XPMEM,  you need to cache the address entry to avoid having to do xpmem_attach()/xpmem_detach() repeatedly. The cache entry needs to be keyed by peer_id. Because the virtual memory address space of different processes would overlap. if the cache is global, would you run into a problem then? Say you have multiple domains/endpoints using XPMEM?

prov/shm/src/smr_ep.c
shefty: Since xpmem is an out of tree driver, it probably ignores all security concerns.  And may be why it's an out of tree driver...

src/xpmem.c
shefty: As one option, you could include the peer id as part of the cache search/hit criteria.  Are you creating 1 cache per peer? Or just a single cache per shm domain?

aingerson: @shefty That's what's happening [right now](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_mr_cache.c#L59-L62)
There is one cache per domain which is a peer set/set of peer ids
Just spit balling but maybe we could have one internal global cache (inside of the xpmem code) and pass in the pid instead?

amirshehataornl: @aingerson, @shefty, I'm not in favor of using a global cache. Even if we use the pid, the pid will be the same globally, and you might hit duplicate cache entries if you have multiple domains talking to the same processes. I think having SHM (or other providers) initialize the cache per domain is the cleanest way to handle this issue.

prov/shm/src/smr_progress.c
amirshehataornl: just to clarify, on the sending side, there is no need to check XPMEM, because it works exactly as CMA. The local buffer address is passed to the receiver. The Receiver then checks if XPMEM is enabled locally and on the sender. If so, then it knows that the address space of the remote process is now exported via XPMEM. It can then safely attach to that memory and copy into the local buffers.

src/xpmem.c
aingerson: Would it be problematic to have it overlap between domains? The domains are probably using different memory regions but even if they are using the same memory, any duplicates should get caught by the cache and help avoid extra calls to xpmem_attach.

prov/shm/src/smr_rma.c
aingerson: There are two "ids" - "id" is used to access peers in the local shm and "peer_id" is what the remote peer uses to access the sender in the remote shm. The peer id is only really used to save into the cmd so that the receiver can easily look up who the message is coming from. Whenever you're accessing the resources locally on the sender's shm (ie ep->region) you need to use the local id - ie "id"
See how smr_generic_rma uses the id

prov/shm/src/smr_progress.c
aingerson: Gotcha. Thanks for clarifying. I thought there was something extra that needed to happen on the sender side

prov/shm/src/smr_rma.c
amirshehataornl: i see what you mean now. thanks.

src/xpmem.c
amirshehataornl: While I understand the idea behind the comment, I'm having trouble justifying the difference in cache usage between IPC and XPMEM. IPC cache exists within the domain. Why not have a global cache in that case also? Is there a functional reason IPC cache needs to be within a domain? Would it be more consistent to have both IPC and XPMEM cache be created in the same way?

aingerson: The IPC cache and XPMEM cache are super similar and can be used in the same way. The idea behind putting it in the domain is because shm is directly calling into it before and after the copy instead of passing it into the copy call. Not all interfaces support or need the cache so it's interface dependent. It's up to shm to use the cache how appropriate:
```
	if (cmd->msg.data.ipc_info.iface == FI_HMEM_ZE) {
		id = cmd->msg.hdr.id;
		ipc_device = cmd->msg.data.ipc_info.device;
		fd = ep->sock_info->peers[id].device_fds[ipc_device];
		ret = ze_hmem_open_shared_handle(fd,
				(void **) &cmd->msg.data.ipc_info.ipc_handle,
				&ipc_fd, ipc_device, &base);
	} else {
		ret = ofi_ipc_cache_search(domain->ipc_cache,
					   cmd->msg.hdr.id,
					   &cmd->msg.data.ipc_info,
					   &mr_entry);
	}
```
We could make it global to each interface which is also fine but it would also require passing in peer identifying information. If you don't want to make it global, I would suggest leaving it in the domain but calling it around the vma copy call instead of passing it in so that it matches the ipc cache usage. Would that be better for you?

include/ofi_shm_p2p.h
amirshehataornl: I defined ofi_process_vm_read/writev in indows/osd.h. Compiling it now to see if there are other changes needed.
My preference is to keep ofi_cma.h as it's more explicit and somewhat symmetrical with ofi_xpmem.h.

src/xpmem.c
amirshehataornl: As discussed I made domain optional in mr cache initialization. And moved xpmem cache to global. Now I use PID as the global identifier. Now pid is used for both CMA and XPMEM

amirshehataornl:
@aingerson @shefty I think I resolved all the comments. Can you please take another look. 
CI seems to be erroring out. Not sure if it's related to this PR or something else. Let me know if it's related to this PR. Don't have access.

prov/shm/src/smr_ep.c
```diff
@@ -1354,6 +1355,11 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 		else
 			ep->smr_progress_ipc_list = smr_progress_ipc_list_noop;
 
+		if (smr_env.use_xpmem && xpmem)
```
aingerson, line 1358: maybe I'm missing something. what is the xpmem variable here?

prov/shm/src/smr_util.h
```diff
@@ -53,7 +54,6 @@
 extern "C" {
 #endif
 
-
 #define SMR_VERSION	5
```
aingerson, line 57: this change needs to increment the version

include/ofi_xpmem.h
```diff
@@ -0,0 +1,88 @@
...
+#if HAVE_XPMEM
+#include <xpmem.h>
+#else
+#define XPMEM_RDWR -1
+#define XPMEM_PERMIT_MODE -1
+#define xpmem_get(seg_id, flags, perm_type, perm_value) -1
+#define xpmem_release(apid)
+typedef int64_t xpmem_apid_t;
+typedef int64_t xpmem_segid_t;
+#endif
```
aingerson, line 48: Can you add a comment to the end of this end if to show which if it goes with, please? And maybe also a comment to explain that this is needed for the smr_region to compile without xpmem available? I stared at this for a little too long lol

prov/shm/src/smr_util.c
```diff
@@ -466,6 +475,23 @@ void smr_map_to_endpoint(struct smr_region *region, int64_t id)
...
+			local_peers[id].xpmem.cap = SMR_VMA_CAP_OFF;
+			return;
+		}
+		local_peers[id].xpmem.cap = SMR_VMA_CAP_ON;
+		local_peers[id].xpmem.addr_max = peer_smr->xpmem_self.address_max;
+	} else {
+		local_peers[id].xpmem.cap = SMR_VMA_CAP_OFF;
+	}
+
+	return;
```
aingerson, line 494: can you hide this in an xpmem specific function that compiles to a no op for the non xpmem case?

src/xpmem.c
```diff
@@ -0,0 +1,229 @@
...
+	uint64_t offset, copy_len;
+	void *mapped_addr;
+	struct ofi_mr_entry *mr_entry;
+	long page_size = ofi_get_page_size();
+
+	assert(local_cnt == remote_cnt);
+
+	for (i = 0; i < remote_cnt; i++) {
+		iov.iov_base = (void *) ofi_get_page_start(remote[i].iov_base,
+							page_size);
```
aingerson, line 180: indentation looks off

src/xpmem.c
```diff
@@ -0,0 +1,229 @@
...
+		iov.iov_len =
+		  (uintptr_t) ofi_get_page_end(remote[i].iov_base + remote[i].iov_len,
+						   page_size) - (uintptr_t)iov.iov_base;
```
aingerson, line 183: also here

src/xpmem.c
```diff
@@ -0,0 +1,229 @@
...
+		ret = ofi_xpmem_cache_search(cache, &iov, id, &mr_entry,
+					     (struct xpmem_client *)user_data);
+		if (ret)
+			return ret;
+
+		mapped_addr = (char*) (uintptr_t)mr_entry->info.mapped_addr +
+		  offset;
+
+		copy_len = (local[i].iov_len <= iov.iov_len - offset) ?
+		  local[i].iov_len : iov.iov_len - offset;
```
aingerson, line 195: and here

src/xpmem.c
```diff
@@ -0,0 +1,229 @@
...
+			xpmem_memcpy(local[i].iov_base, mapped_addr,
+				     copy_len);
+
+		ofi_mr_cache_delete(cache, mr_entry);
+	}
+
+	return 0;
+}
+
+# else
```
aingerson, line 210: `#else` for consistency

src/xpmem_cache.c
```diff
@@ -0,0 +1,164 @@
...
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#if HAVE_CONFIG_H
+#  include <config.h>
```
aingerson, line 34: align #s for consistency with other files

aingerson (commented):
Just a few things. I'm liking the p2p abstraction! Nice!

prov/shm/src/smr_util.c
```diff
@@ -86,9 +86,9 @@ void smr_cma_check(struct smr_region *smr, struct smr_region *peer_smr)
 	assert(remote_pid == peer_smr->pid);
 
 	if (smr == peer_smr) {
-		smr->cma_cap_self = (ret == -1) ? SMR_CMA_CAP_OFF : SMR_CMA_CAP_ON;
+		smr->cma_cap_self = (ret == -1) ? SMR_VMA_CAP_OFF : SMR_VMA_CAP_ON;
```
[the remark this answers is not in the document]
amirshehataornl: We're not turning off xpmem capability here. This is just turning off cma capability. xpmem has its own set of capabilities.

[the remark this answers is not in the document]
aingerson: yeah sorry, left that before I finished the whole patch set and meant to delete! thanks!

aingerson:
also CI failures are unrelated. our clusters were rebooted over the weekend and some didn't come back properly. you can ignore them

prov/shm/src/smr_ep.c
amirshehataornl: It's a global xpmem variable which holds information about the exported memory address space of the process which is exported via XPMEM. It will be initialized if XPMEM is available. but you're right, it's redundant to explicitly check it here.

man/fi_shm.7.md
```diff
@@ -168,6 +168,13 @@ The *shm* provider checks for the following environment variables:
   page fault is reported, so that there is valid address translation for the
   remaining addresses in the command. This minimizes DSA page faults. Default
   false
+
+*FI_SHM_USE_XPMEM*
+ : SHM can use SAR, CMA or XPMEM for host memory transfer. If
+   FI_SHM_USE_XPMEM is set to 1, the provider will select XPMEM over CMA if
+   XPMEM is available.  Otherwise, if neither CMA nor XPMEM are available
+   SHM shall default to the SAR protocol. Default 0
+
```
aingerson, line 177: Since the param is defined with the core_prov instead of the shm provider would the variable be FI_USE_XPMEM? Also this is missing the second xpmem variable for chunk size

prov/shm/src/smr_domain.c
```diff
@@ -33,6 +33,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <ofi_xpmem.h>
```
aingerson, line 36: is this needed anymore?

prov/shm/src/smr_progress.c
```diff
@@ -322,6 +323,13 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
 		goto out;
 	}
 
+	xpmem = &smr_peer_data(ep->region)[cmd->msg.hdr.id].xpmem;
+	if (ep->region->xpmem_cap_self == SMR_VMA_CAP_ON &&
+	    xpmem->cap == SMR_VMA_CAP_ON) {
+		p2p_type = FI_SHM_P2P_XPMEM;
```
aingerson, line 329: can we set the p2p_type earlier when we initialize the ep so we don't have to check this every time on the fast path?

prov/shm/src/smr_progress.c
```diff
@@ -322,6 +323,13 @@ static int smr_progress_iov(struct smr_cmd *cmd, struct iovec *iov,
...
+		user_data = xpmem;
```
aingerson, line 330: probably can just pass in xpmem directly to the p2p_copy function to eliminate this local variable and the if/else

prov/shm/src/smr_util.c
```diff
@@ -284,6 +285,13 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
 
 	(*smr)->cma_cap_peer = SMR_VMA_CAP_NA;
 	(*smr)->cma_cap_self = SMR_VMA_CAP_NA;
+	(*smr)->xpmem_cap_self = (xpmem) ? SMR_VMA_CAP_ON
+		: SMR_VMA_CAP_OFF;
```
aingerson, line 289: move colon to previous line

prov/shm/src/smr_util.c
```diff
@@ -306,6 +314,7 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
 		smr_peer_addr_init(&smr_peer_data(*smr)[i].addr);
 		smr_peer_data(*smr)[i].sar_status = 0;
 		smr_peer_data(*smr)[i].name_sent = 0;
+		smr_peer_data(*smr)[i].xpmem.cap = SMR_VMA_CAP_OFF;
```
aingerson, line 317: init to NA?

src/xpmem.c
```diff
@@ -0,0 +1,256 @@
...
+	struct ofi_mr_entry *mr_entry;
+	long page_size = ofi_get_page_size();
+
+	assert(local_cnt == remote_cnt);
+
+	for (i = 0; i < remote_cnt; i++) {
+		iov.iov_base = (void *) ofi_get_page_start(remote[i].iov_base,
+							   page_size);
+		iov.iov_len =
+		  (uintptr_t) ofi_get_page_end(remote[i].iov_base + remote[i].iov_len,
```
aingerson, line 182: indentation off (extra spaces)

src/xpmem.c
```diff
@@ -0,0 +1,256 @@
...
+		offset = (uintptr_t)((uintptr_t) remote[i].iov_base -
+				     (uintptr_t) iov.iov_base);
+
+		ret = ofi_xpmem_cache_search(cache, &iov, pid, &mr_entry,
+					     (struct xpmem_client *)user_data);
+		if (ret)
+			return ret;
+
+		mapped_addr = (char*) (uintptr_t)mr_entry->info.mapped_addr +
+		  offset;
```
aingerson, line 193: same here use tabs for extra offset

aingerson (commented):
A few lingering things but otherwise looks good to me!

man/fi_shm.7.md
shefty: Why is this global to libfabric, rather than limited to the provider that implements it?

aingerson: I think it could go either way but the xpmem definition outlined here is something usable for all providers so you can condense all of the enabling into one versus having to enable it for different providers using different variables. I think you could make the same argument about other core_prov variables like FI_MR_CACHE_* or FI_HMEM_* variables for example. Is there a value to separating it out between providers? I don't personally think so.
Realistically shm is the only provider that is going to make use of it so I don't think it really matters. I don't care either way.

prov/shm/src/smr_util.c
amirshehataornl: I don't want to use NA as it has no benefity. The cap is either OFF/ON. Not sure what NA would mean.

prov/shm/src/smr_progress.c
amirshehataornl: I implemented this by adding a variable in the ep, which defaults to CMA and gets set to XPMEM if it's enabled. use_xpmem should remain a bool. It's more consistent with the rest of the other env vars: use_dsa_sar and disable cma.

prov/shm/src/smr_util.c
```diff
@@ -284,6 +286,13 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
 
 	(*smr)->cma_cap_peer = SMR_VMA_CAP_NA;
 	(*smr)->cma_cap_self = SMR_VMA_CAP_NA;
+	(*smr)->xpmem_cap_self = (xpmem) ? SMR_VMA_CAP_ON :
+		SMR_VMA_CAP_OFF;
+	if (xpmem)
+		(*smr)->xpmem_self = xpmem->pinfo;
+	else
+		memset(&(*smr)->xpmem_self, 0,
+		       sizeof((*smr)->xpmem_self));
```
aingerson, line 295: Can you refactor this to be
```
if (xpmem) {
    //set xpmem cap and info (and maybe move env var check to here too)
}
```
do we even need the else if the region initializes to all 0s?

prov/shm/src/smr_ep.c
```diff
@@ -1354,6 +1355,11 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 		else
 			ep->smr_progress_ipc_list = smr_progress_ipc_list_noop;
 
+		if (smr_env.use_xpmem)
+			ep->region->xpmem_cap_self = SMR_VMA_CAP_ON;
+		else
+			ep->region->xpmem_cap_self = SMR_VMA_CAP_OFF;
+
```
aingerson, line 1362: Can we move this setting into smr_create?
Also do we need to force disable xpmem if FI_HMEM was requested like for CMA? (see this right above in this file, line 1344)

prov/shm/src/smr_ep.c
```diff
@@ -1387,6 +1393,13 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 		if (smr_env.use_dsa_sar)
 			smr_dsa_context_init(ep);
 
+		/* if XPMEM is on after exchanging peer info, then set the
+		 * endpoint p2p to XPMEM so it can be used on the fast
+		 * path
+		 */
```
aingerson, line 1399: does exchanging peer info have anything to do with this? I think I would prefer the p2p_type to be set in a single line with the default case

prov/shm/src/smr_rma.c
```diff
@@ -55,13 +55,14 @@ static void smr_format_rma_resp(struct smr_cmd *cmd, fi_addr_t peer_id,
 	cmd->msg.hdr.size = total_len;
 }
 
-static ssize_t smr_rma_fast(struct smr_region *peer_smr, const struct iovec *iov,
-			size_t iov_count, const struct fi_rma_iov *rma_iov,
-			size_t rma_count, void **desc, int peer_id, void *context,
+static ssize_t smr_rma_fast(struct smr_ep *ep, struct smr_region *peer_smr,
+			const struct iovec *iov, size_t iov_count,
+			const struct fi_rma_iov *rma_iov, size_t rma_count,
+			void **desc, int peer_id, int peer_idx, void *context,
```
aingerson, line 61: Can you rename this field to `id` to match the previous usage of it?

src/xpmem.c
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	while (fgets(buffer, sizeof(buffer), fh)) {
+		/* each line of /proc/self/maps starts with low-high in
+		 * hexidecimal (without a 0x) */
+		low = strtoul(buffer, &tmp, 16);
+		high = strtoul(tmp + 1, NULL, 16);
+		if (address_max < high) {
+			address_max = high;
+		}
```
aingerson, line 94: drop brackets

aingerson (commented):
Just a few lingering things. Biggest thing is I think the xpmem self on/off could be streamlined a little. I'm just finding it confusing how many places we're checking xpmem/use_xpmem and all that and I think it would be better to isolate it in one place if possible.
Otherwise looks pretty good!

prov/shm/src/smr_ep.c
amirshehataornl: we do not need to disable xpmem in case of FI_HMEM. They can coexist. In fact we can memcpy between xpmem exported memory and device memory. I've tested this case.

prov/shm/src/smr_ep.c
amirshehataornl: The idea here is that it's potentially possible to turn off xpmem if any of the peers do not support xpmem. So this would be the most accurate place to set XPMEM type. The idea is all peers should support xpmem or none of them will use it.

amirshehataornl: can you please tell me what you want to see. I'm not following your comment.

amirshehataornl: we set the default to CMA and set it to XPMEM if XPMEM is on. When we exchange the peer information at that point we try to enable xpmem for that peer and if we fail we turn off xpmem locally as well. The intent here is if we can't talk to any of the peers over xpmem, then no one talks over xpmem and we fall back to cma. This behavior makes sense since XPMEM is system wide and if it's enabled on the node all peers should be able to use it and if there is a failure then no one uses it.

This is the reason I put it here.

aingerson (approved):

aingerson:
@shefty Do you want to do another review?

shefty:
I don't know.  I feel like I should, but I also want to just merge it. :)  I'll do a quick review.

prov/shm/src/smr_util.c
```diff
@@ -466,6 +477,24 @@ void smr_map_to_endpoint(struct smr_region *region, int64_t id)
...
+			region->xpmem_cap_self = SMR_VMA_CAP_OFF;
+			return;
+		}
+		local_peers[id].xpmem.cap = SMR_VMA_CAP_ON;
+		local_peers[id].xpmem.addr_max = peer_smr->xpmem_self.address_max;
+	} else {
+		local_peers[id].xpmem.cap = SMR_VMA_CAP_OFF;
+	}
+
+	return;
```
shefty, line 497: unnecessary - but let's fix this in a follow on patch

shefty (approved):

shefty:
Intel CI failures were both in the shm provider, but probably unrelated?
```
name:   fi_av_xfer -e rdm -p "shm"
11:48:33    timestamp: 20230810-184833+0000
11:48:33    result: Fail
11:48:33    time:   2
11:48:33    server_cmd:  /home/cstbuild/cloudbees_job_install//OFIWG_Libfabric/ofi_libfabric/PR-8200/28/reg/bin/fi_av_xfer -e rdm -p "shm"   -s dsa
11:48:33    server_stdout: |
11:48:33      libfabric:3460554:1691693312::shm:ep_ctrl:smr_create():227<warn> shm file in use (dsa:9228)
11:48:33      fi_enable(): common/shared.c:1337, ret=-16 (Device or resource busy)
11:48:33      AV address removal: FAIL
11:48:33    client_cmd:  /home/cstbuild/cloudbees_job_install//OFIWG_Libfabric/ofi_libfabric/PR-8200/28/reg/bin/fi_av_xfer -e rdm -p "shm"   -s dsa dsa
11:48:33    client_stdout: |
11:48:33      libfabric:3460610:1691693313::shm:ep_ctrl:smr_create():227<warn> shm file in use (dsa:36382:0)
11:48:33      fi_enable(): common/shared.c:1337, ret=-16 (Device or resource busy)
11:48:33      AV address removal: FAIL
```
There are 2 failures, both in this test.  I restarted the CI

amirshehataornl:
hmm... the difference between this revision and last which passed testing was just renaming a variable. Is this a confirmed CI issue or do I need to do more investigation?

shefty:
I think that failure is a result of some other previous job running on the CI failing.  And now our CI nodes are once again (still?) broken...  merging, because I don't believe the last CI failure was related.

shefty:
Thanks for the work here.  The p2p abstraction should be useful moving forward.

PR #8202, opened 2022-11-01
prov/shm: add support for FI_AV_USER_ID

aingerson:
If the FI_AV_USER_ID flag is set on insert, set the fi_addr to report with completions as the inputed fi_addr from the user.


aingerson:
@shefty This was supposed to be a part of the shm peer support, but I guess I accidentally got rid of it during a force push update....

aingerson:
@shefty Not really sure why Jenkins is showing as still pending. According to the logs, it finished. Failed just with the sockets rdm rma trigger failure. Should be ok. Let me know if you have any thoughts.

PR #8205, opened 2022-11-01
contrib/intel/jenkins: add Jenkinsfile.weekly

zachdworkin:
Add back jenkinsfile.weekly for running upstream/main's tip of tree. Weekly jenkinsfile must not have opt-out function as we want it to run every time the new jenkins job invokes it.


zachdworkin:
@shefty i ran this in my private job to test it for the new weekly job that Rajalaxmi asked for. It ran in my job/jenkins_ci branch.
It will not get used by the current job that happens per PR and since that job is defined to just use Jenkinsfile.

zachdworkin:
Closing, new cloudbees has a feature to enable environment variables per job which will remove the need for this file. I will collect them by hand until the migration occurs

PR #8213, opened 2022-11-04
prov/net: TX operations now support io_uring

sydidelot:
This PR adds support of io_uring to TX operations.
    
The implementation is still incomplete because of the following missing features:

1. We need a mechanism to cancel all pending io_uring operations
and wait for them to complete before we can close the ring.
Right now, pending operations would cause the code to fail
the assertion failure !ep->cur_tx.uring_busy in
xnet_ep_flush_all_queues().

2. ofi_uring_submit() might return an error or submit fewer
SQs than expected. This is handled using an assertion failure 
right now and we definitely need a better strategy there.

### DIFF
--- a/prov/net/src/xnet.h
+++ b/prov/net/src/xnet.h
@@ -253,7 +253,6 @@ void xnet_freeall_conns(struct xnet_rdm *rdm);
 struct xnet_uring {
 	struct fid fid;
 	ofi_io_uring_t ring;
-	size_t credits;
 };
 
 /* Serialization is handled at the progress instance level, using the
--- a/prov/net/src/xnet_progress.c
+++ b/prov/net/src/xnet_progress.c
@@ -808,11 +808,11 @@ static void xnet_run_ep(struct xnet_ep *ep, bool pin, bool pout, bool perr)
 static void xnet_progress_cqe(struct xnet_uring *uring,
 			      ofi_io_uring_cqe_t *cqe)
 {
-	struct ofi_bsock *bsock;
+	struct ofi_sockctx *sockctx;
 
 	assert(xnet_io_uring);
-	bsock = (struct ofi_bsock *) cqe->user_data;
-	assert(bsock);
+	sockctx = (struct ofi_sockctx *) cqe->user_data;
+	assert(sockctx);
 }
 
 static void xnet_progress_uring(struct xnet_uring *uring)
--- a/prov/net/src/xnet_progress.c
+++ b/prov/net/src/xnet_progress.c
@@ -144,10 +144,62 @@ static ssize_t xnet_recv_msg_data(struct xnet_ep *ep)
 	return -FI_EAGAIN;
 }
 
-static void xnet_progress_tx(struct xnet_ep *ep)
+static void xnet_complete_tx(struct xnet_ep *ep, ssize_t ret)
 {
 	struct xnet_xfer_entry *tx_entry;
 	struct xnet_cq *cq;
+
+	tx_entry = ep->cur_tx.entry;
+	cq = container_of(ep->util_ep.tx_cq, struct xnet_cq, util_cq);
+
+	if (ret) {
+		FI_WARN(&xnet_prov, FI_LOG_DOMAIN, "msg send failed\n");
+		xnet_cntr_incerr(ep, tx_entry);
+		xnet_cq_report_error(&cq->util_cq, tx_entry, (int) -ret);
+		xnet_free_xfer(xnet_ep2_progress(ep), tx_entry);
+	} else if (tx_entry->ctrl_flags & XNET_NEED_ACK) {
+		/* A SW ack guarantees the peer received the data, so
+		 * we can skip the async completion.
+		 */
+		slist_insert_tail(&tx_entry->entry,
+				  &ep->need_ack_queue);
+	} else if (tx_entry->ctrl_flags & XNET_NEED_RESP) {
+		// discard send but enable receive for completeion
+		assert(tx_entry->resp_entry);
+		tx_entry->resp_entry->ctrl_flags &= ~XNET_INTERNAL_XFER;
+		xnet_free_xfer(xnet_ep2_progress(ep), tx_entry);
+	} else if ((tx_entry->ctrl_flags & XNET_ASYNC) &&
+		   (ofi_val32_gt(tx_entry->async_index,
+				 ep->bsock.done_index))) {
+		slist_insert_tail(&tx_entry->entry,
+					&ep->async_queue);
+	} else {
+		ep->report_success(ep, &cq->util_cq, tx_entry);
+		xnet_free_xfer(xnet_ep2_progress(ep), tx_entry);
+	}
+
+	if (!slist_empty(&ep->priority_queue)) {
+		ep->cur_tx.entry = container_of(slist_remove_head(
+						&ep->priority_queue),
+				     struct xnet_xfer_entry, entry);
+		assert(ep->cur_tx.entry->ctrl_flags & XNET_INTERNAL_XFER);
+	} else if (!slist_empty(&ep->tx_queue)) {
+		ep->cur_tx.entry = container_of(slist_remove_head(
+						&ep->tx_queue),
+				     struct xnet_xfer_entry, entry);
+		assert(!(ep->cur_tx.entry->ctrl_flags & XNET_INTERNAL_XFER));
+	} else {
+		ep->cur_tx.entry = NULL;
+		return;
+	}
+
+	ep->cur_tx.data_left = ep->cur_tx.entry->hdr.base_hdr.size;
+	OFI_DBG_SET(ep->cur_tx.entry->hdr.base_hdr.id, ep->tx_id++);
+	ep->hdr_bswap(&ep->cur_tx.entry->hdr.base_hdr);
+}
+
+static void xnet_progress_tx(struct xnet_ep *ep)
+{
 	ssize_t ret;
 
 	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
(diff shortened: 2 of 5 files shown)


prov/net/src/xnet_progress.c
```diff
@@ -765,11 +801,35 @@ void xnet_progress_async(struct xnet_ep *ep)
...
 {
+	struct xnet_xfer_entry *tx_entry;
 	struct ofi_bsock *bsock;
+	struct xnet_ep *ep;
 
 	assert(xnet_io_uring);
 	bsock = (struct ofi_bsock *) cqe->user_data;
 	assert(bsock);
+
+	if (&uring->ring == bsock->sockapi->tx_io_uring) {
```
shefty, line 812: Is this check to distinguish sends from receives (in the future)?

prov/net/src/xnet_progress.c
```diff
@@ -94,6 +119,9 @@ static ssize_t xnet_send_msg(struct xnet_ep *ep)
 
 	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
 	assert(ep->cur_tx.entry);
+	if (xnet_io_uring && !xnet_get_uring_credit(&xnet_ep2_progress(ep)->tx_uring))
+		return -FI_EAGAIN;
```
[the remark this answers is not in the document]
sydidelot: Thanks for the review. 
We can provision the ring to be larger than the max ep's we want to support, but I have some concerns with this strategy: 
- the net provider sets `ep_cnt` to 8192, but it doesn't seem that the code actually restricts the number of endpoints.
- we will need to provision extra uring entries to post cancellation requests of pending SQEs when an endpoint is flushed (not implemented yet).

I guess the question is: do we want to flow-control io_uring or do we want to always provision the io_uring for the worst case scenario?
Personally, I like the idea where the number of uring entries is independent of the maximum number of endpoints. I would expect the system to reach its peek performance after a certain number of SQEs inflight, which obviously depends on the number of communicating EPs, but not on the max EPs allocated. However, adding a flow-control might be a problem for fairness as the last EPs in the TX/RX lists may not always be served because the first EPs in the list may have consumed all the entries already. Maybe your suggestion is the best approach after all. 

BTW, the flag [IORING_FEAT_NODROP](https://manpages.debian.org/testing/liburing-dev/io_uring_setup.2.en.html#IORING_FEAT_NODROP) could also be used to bring some kind of control flow. When this flags is set, `io_uring_submit()` would return -EBUSY if the ring is full and cannot accept new SQEs.

prov/net/src/xnet.h
```diff
@@ -156,6 +156,7 @@ struct xnet_cur_rx {
 struct xnet_cur_tx {
 	size_t			data_left;
 	struct xnet_xfer_entry	*entry;
+	bool			uring_busy;
```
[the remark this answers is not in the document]
sydidelot: It makes sense. I renamed it to `uring_sqe_inuse`.

prov/net/src/xnet_progress.c
```diff
@@ -203,11 +231,19 @@ static void xnet_progress_tx(struct xnet_ep *ep)
...
+		return;
+
 	while (ep->cur_tx.entry) {
 		ret = xnet_send_msg(ep);
 		if (OFI_SOCK_TRY_SND_RCV_AGAIN(-ret)) {
 			xnet_update_pollflag(ep, POLLOUT, true);
 			return;
+		} else if (ret == -OFI_EINPROGRESS_URING) {
+			/* io_uring will wake us when the CQE will be posted */
+			ep->cur_tx.uring_busy = true;
```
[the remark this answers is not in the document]
sydidelot: Done.

prov/net/src/xnet_progress.c
sydidelot: Right, this check is used to distinguish sends from receives. The `else` block for receives will be implemented in another PR.

prov/net/src/xnet_progress.c
```diff
@@ -203,11 +231,19 @@ static void xnet_progress_tx(struct xnet_ep *ep)
 	ssize_t ret;
 
 	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
+	if (ep->cur_tx.uring_busy)
+		return;
```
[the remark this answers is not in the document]
sydidelot: I moved this flag in the bsock API. I will update the PR after we agree on the changes related to credits/control-flow.

prov/net/src/xnet_progress.c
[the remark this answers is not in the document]
shefty: I'd prefer the explicit check, because I can envision needing to add it at some point anyway to debug some issue.  It seems that moving the credit check closer to io_uring_submit() would align with the behavior of NODROP.  I don't know what performance penalty we'd hit if NODROP is enabled and we hit that situation frequently.  The description makes it sound like NODROP might be slower than just not overrunning the cq.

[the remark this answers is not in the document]
sydidelot: @shefty There is one problem if we move the credit check closer to io_uring_submit(): it is impossible to select how many ready SQEs will be submitted: it's either all or none. If the provider has prepared too many SQEs, it has to wait for at least as many free credits as the number of SQEs prepared before it can call `io_uring_submit()`. 
I decided to move the credit check under the sockapi. Please let me know your opinion.

src/iouring.c
```diff
@@ -42,20 +42,23 @@ ssize_t ofi_sockapi_send_uring(struct ofi_sockapi *sockapi, SOCKET sock,
...
 
-	sqe = io_uring_get_sqe(sockapi->tx_io_uring);
+	sqe = io_uring_get_sqe(uring->io_uring);
 	if (!sqe)
 		return -FI_EOVERFLOW;
 
 	io_uring_prep_send(sqe, sock, buf, len, flags);
 	io_uring_sqe_set_data(sqe, ctx);
 	ctx->uring_sqe_inuse = true;
+	uring->credits--;
```
shefty, line 61: This change and the previous patch both assume that the caller into sockapi is providing synchronization.  That's true for the net provider, but in general socket calls are already considered thread safe.  I'm not sure how to do this yet, but it would be good to have the code assert() that we're synchronized through these calls.  This isn't a problem introduced by these changes, as the same requirement holds for the ofi_bsock abstraction.

Maybe we just need a debug variable, something like:
```
struct ofi_some_abstraction {
    OFI_DBG_VAR(bool, in_use)
};

void ofi_some_function(struct ofi_some_abstraction *x)
{
    assert(!x->in_use);
    OFI_DBG_SET(x->in_use, true);
    ...
    OFI_DBG_SET(x->in_use, false);
}
```
It would be a little nicer if there were an abstraction just for this purpose, or maybe we specify what lock is supposed to be held, so we can check against that specifically.

prov/net/src/xnet_progress.c
```diff
@@ -766,14 +783,39 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
...
+	assert(sockctx->uring_sqe_inuse);
+	sockctx->uring_sqe_inuse = false;
 	if (&uring->ring == progress->sockapi.tx_uring.io_uring) {
+		tx_entry = ep->cur_tx.entry;
+		assert(tx_entry);
+
 		progress->sockapi.tx_uring.credits++;
+		if (cqe->res < 0) {
+			if (!OFI_SOCK_TRY_SND_RCV_AGAIN(-cqe->res))
+				xnet_complete_tx(ep, cqe->res);
```
shefty, line 809: What happens if the result is -EAGAIN?  Does the flow simply retry sending the current message?

prov/net/src/xnet_progress.c
```diff
@@ -766,14 +783,39 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
...
+		} else {
+			ep->cur_tx.data_left -= cqe->res;
```
shefty, line 811: Maybe assert(cqe->res >= ep->cur_tx.data_left) to ensure things haven't gotten out of whack?

shefty (approved):
Thanks - I have a couple of questions, which likely have simple answers.  The series looks clean and relatively straightforward.

prov/net/src/xnet_progress.c
sydidelot: Good idea! Thanks!

prov/net/src/xnet_progress.c
sydidelot: That's correct: if the result is -EAGAIN, the function executes `xnet_progress_tx(ep)`, which would retry sending the current message.

prov/net/src/xnet_progress.c
sydidelot: Actually, it should be `assert(cqe->res <= ep->cur_tx.data_left)`

src/iouring.c
sydidelot: I understand your point. I will try to put something in place in a new PR.

shefty: I also toyed around with this idea yesterday after writing that.  If we don't want to pass the upper lock down, which I'm not sold on, I was thinking of something simple like:
```
#IF DEBUG
struct ofi_lock_check {
    int in_use;
};

static int ofi_lock_check_start(...)
{
    return check->in_use++;
}

static int ofi_lock_check_end(...)
{
    return --check->in_use;
}
#endif

struct some_example {
    OFI_DBG_VAR(struct ofi_lock_check, check)
};
void use_example()
{
    assert(!ofi_lock_check_start(&check));
    ...
    assert(!ofi_lock_check_end(&check));
}
```
It's modeled after the mutex debug checks.  At least helps document in the code that some upper locking is required.

shefty:
Can you please rebase with the top of main if you haven't?  Intel CI failures are some problem in the Intel CI scripts.

sydidelot:
Thank you. You are right and the only change from the previous submission is the assert.
Please let me know the failure and if I can help.

shefty:
The PR is good.  The problem is some syntax error in an internal script.  We're re-running everything now.

PR #8215, opened 2022-11-04
prov/shm: remove extraneous ofi_cirque_next() call

amirshehataornl:
In smr_atomic_inject() there was an extraneous ofi_cirque_next() call. This didn't cause an issue because ofi_cirque_next() would return the same entry for multiple calls as long as there was no commit function call in between.


aingerson (approved):

shefty:
CI failure is sockets provider (fixed known issue)

PR #8216, opened 2022-11-04
contrib/intel/jenkins add net and multinode performance to summary stage

zachdworkin:
Add net provider and multinode performance stage logs to summary.py
They are now included and presented in the same format as the other tests.

The entire summary.py has also been refactored to use python classes. This will help with maintainability and adding future summarizers to it.

### DIFF
--- a/contrib/intel/jenkins/summary.py
+++ b/contrib/intel/jenkins/summary.py
@@ -253,6 +253,30 @@ class ZeSummarizer(Summarizer):
             self.excluded_tests.append(self.test_name_string)
         self.result = ''
 
+class MultinodePerformanceSummarizer(Summarizer):
+    def __init__(self, log_dir, prov, build_mode):
+        super().__init__(log_dir, prov, build_mode,
+                         f'multinode_performance_{prov}_{mode}',
+                         f"multinode performance {prov} {mode}")
+
+    def check_name(self):
+        #name lines look like "starting <test_name>... <result>"
+        if 'starting' in self.line and '...' in self.line:
+            self.test_name_string = self.line.split()[1].split('.')[0]
+
+    def check_pass(self):
+        if 'pass' in self.line:
+            self.passes += 1
+
+    def check_fail(self):
+        if 'fail' in self.line:
+            self.fails += 1
+            self.failed_tests.append(self.test_name_string)
+
+        if "exiting with" in self.line:
+            self.fails += 1
+            self.failed_tests.append(self.test_name_string)
+
 class OnecclSummarizer(Summarizer):
     def __init__(self, log_dir, prov, build_mode):
         super().__init__(log_dir, prov, build_mode,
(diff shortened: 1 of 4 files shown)


zachdworkin:
@shefty let this finish running jenkins so I can double check everything

zachdworkin:
@shefty the jenkins summary looks good and all the stages passed. I would like Stephen to review the python to make sure the updates are how we want our python files to look before you merge this.

contrib/intel/jenkins/summary.py
```diff
@@ -7,389 +8,493 @@
...
-        if 'name:' in line:
-            test_name = line.split()[2:]
-            test_name_string = ' '.join(test_name)
+        if 'name:' in self.line:
+            test_name = self.line.split()[2:]
+            self.test_name_string = ' '.join(test_name)
 
-        if 'result:' in line:
-            result_line = line.split()
+    def get_result_line(self):
```
ooststep, line 154: get_result_line seems to be used to parse the line for the result, and then the caller of get_result_line is also resetting self.result.

instead, you could just return the result and not store it, then you don't have to remember to reset it.  additionally, it seems some places want to parse the full matching line, in which case you could change this function to return a tuple

```
def get_result_line(self, line) -> Tuple[str,str]:
    result = line.split("result:")
    if len(result) > 1:
        return (result[-1].lower().strip(), line)
    return (None, None)
```

then the callers would do something like
```
result, result_line = self.get_result_line(line)
```
or `result, _ = self.get_result_line(line)` if the result_line is unused

contrib/intel/jenkins/summary.py
```diff
@@ -7,389 +8,493 @@
...
+
+    def check_exclude(self):
+        self.get_result_line()
+        if self.result == 'excluded' or self.result == 'notrun':
+            self.excludes += 1
+            self.excluded_tests.append(self.test_name_string)
+
+        self.result = ''
+
+class ZeSummarizer(Summarizer):
```
ooststep, line 195: ZeSummarizer seems to closely match FabtestsSummarizer.  It could be good to try and reduce some duplication.  The same comments I made for FabtestsSummarizer apply here, too

ooststep (commented):
there are many places where fields are stored in the instance of the object when you really don't need them to live that long. I tend to prefer passing in the requirements for functions to operate on rather than storing them in the object when they don't really need to be.  It reduces coupling and makes it slightly more obvious the intent of the function

I've highlighted some of these in comments, but not all.

contrib/intel/jenkins/summary.py
```diff
@@ -7,389 +8,493 @@
...
+            if int(self.line.split()[0]) != self.fails:
+                print(f"fails {self.fails} does not match log reported fails " \
+                        f"{int(self.line.split()[0])}")
+
+    def check_fails(self):
+        if "exiting with" in self.line:
+            self.fails += 1
+            self.failed_tests.append(f"{self.prov} {self.passes + self.fails}")
+
+    def check_line(self):
```
ooststep, line 375: my comment was lost somehow, but the switch-like syntax works for callables, too (I believe it could even be a lambda if you want to define the callable in-place)

```
check = {
    "uh": self.check_uh,
    "isx": self.check_isx,
    "prk": self.check_prk
}
check[self.prov]()
```

zachdworkin:
@ooststep thanks for the review. All changes have been made. Rerunning with them.

@shefty I have verified manually that the numbers in the summary stage are correct reflections of the number of tests run.
The only one that is "doubled" is shm fabtests because it runs twice (with and without an environment variable) and appends to the same file. This will be fixed in a separate PR.
Like last time, please wait for the ci to finish running so if there are fails i can verify them.

zachdworkin:
@shefty I am pretty sure the net provider failure is known. (especially since I didnt touch that code). The summary report looks how I expect it to.

shefty:
Net provider issue should have been resolved.  If you haven't updated your tree, you can ignore the failure.

zachdworkin:
@shefty I haven't updated my tree. I last updated it 5 days ago when I opened the PR.

PR #8220, opened 2022-11-05
core/hmem_ops: Rename parameters for consistency

shefty:
At least I think this renames the parameters for consistency and to make them easier to understand.  I didn't update the underlying calls, only the hmemops.  Renames are generally

ptr -> addr
ipc_ptr -> mapped_addr
base -> base_addr
dev_buf -> base_addr
size (when referring to base_addr) -> base_length

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -56,7 +56,7 @@ extern bool ofi_hmem_disable_p2p;
  */
 struct ipc_info {
 	uint64_t	iface;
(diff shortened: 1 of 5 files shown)


shefty:
@amirshehataornl @wzamazon @iziemba - please review and comment if you see any issues with the rename.  Although we don't need to maintain API compatibility with these calls, we should start documenting some of the more important internal API calls.

aingerson (approved):

PR #8227, opened 2022-11-08
prov/net: fix error path in xnet_enable_rdm

nmorey:
If xnet_listen fails (happens 100% of the time on a system with no network interface but lo), the progress lock is not released which causes a deadlock when fi_close is called later on the endpoint.

Signed-off-by: Nicolas Morey-Chaisemartin <nmoreychaisemartin@suse.com>

### DIFF
--- a/prov/net/src/xnet_rdm.c
+++ b/prov/net/src/xnet_rdm.c
@@ -711,7 +711,7 @@ static int xnet_enable_rdm(struct xnet_rdm *rdm)
 
 	ret = xnet_listen(rdm->pep, progress);
 	if (ret)
-		return ret;
+		goto unlock;
 
 	/* TODO: Move updating the src_addr to pep_listen(). */
 	len = sizeof(rdm->addr);


nmorey:
This is also broken in 1.16.1.
Bug can be easily trigger by running openmpi 4.1.4 with libfabric 1.16.1 on a node without network interfaces with a hello world:
`[   85s] + /usr/lib64/mpi/gcc/openmpi4/bin/mpiexec --use-hwthread-cpus --mca pml_base_verbose 100 --mca btl tcp,self -n 1 python3.8 -B test/runtests.py -v --exclude=test_spawn
[   85s] [portia:03402] mca: base: components_register: registering framework pml components
[   85s] [portia:03402] mca: base: components_register: found loaded component v
[   85s] [portia:03402] mca: base: components_register: component v register function successful
[   85s] [portia:03402] mca: base: components_register: found loaded component cm
[   85s] [portia:03402] mca: base: components_register: component cm register function successful
[   85s] [portia:03402] mca: base: components_register: found loaded component monitoring
[   85s] [portia:03402] mca: base: components_register: component monitoring register function successful
[   85s] [portia:03402] mca: base: components_register: found loaded component ob1
[   85s] [portia:03402] mca: base: components_register: component ob1 register function successful
[   85s] [portia:03402] mca: base: components_register: found loaded component ucx
[   85s] [portia:03402] mca: base: components_register: component ucx register function successful
[   85s] [portia:03402] mca: base: components_open: opening pml components
[   85s] [portia:03402] mca: base: components_open: found loaded component v
[   85s] [portia:03402] mca: base: components_open: component v open function successful
[   85s] [portia:03402] mca: base: components_open: found loaded component cm
[   85s] [portia:03402] mca: base: components_open: component cm open function successful
[   85s] [portia:03402] mca: base: components_open: found loaded component monitoring
[   85s] [portia:03402] mca: base: components_open: component monitoring open function successful
[   85s] [portia:03402] mca: base: components_open: found loaded component ob1
[   85s] [portia:03402] mca: base: components_open: component ob1 open function successful
[   85s] [portia:03402] mca: base: components_open: found loaded component ucx
[   85s] [portia:03402] mca: base: components_open: component ucx open function successful
[   85s] [portia:03402] select: component v not in the include list
[   85s] [portia:03402] select: initializing pml component cm
`
And it stalls here stuck on the `ofi_genlock_loc` from `xnet_rdm_close`

The fix is still under test but the code definitly looks off to me here.

tschuett (commented):
`xnet_rdm_close` looks also fishy, but it is **close**.

tschuett:
maybe we can:
```c
assert(!ofi_genlock_held(progress->rdm_lock));
```
before and after calls to `xnet_enable_rdm`. `xnet_rdm_close` seems to use a different lock.

shefty (approved):
Thanks!  I'll also pick for v1.16.x after merging.

shefty:
@tschuett - xnet_rdm_close() does have a similar bug.  Thanks for pointing that out.  I'll create a fix for that and also backport.

nmorey:
> @tschuett - xnet_rdm_close() does have a similar bug. Thanks for pointing that out. I'll create a fix for that and also backport.

I didn't worry to much about the close part because I'm not sure what the user is expected to do when a close fail.
Apart from the deadlock issue, what will happen if the endpoint is closed again ? Won't the `fi_close(&rdm->pep->util_pep.pep_fid.fid);`fail if it went through the first time so the user is stuck in a close/fail loop anyway ?
But yes, deadlocks are bad even in those cases :)

shefty:
Correct - xnet_rdm_close() won't work if called multiple times because of a failure.  I can fixup some of this by clearing pointers once we pass those cleanup sections.  For example, add segments such as this:
```
if (rdm->pep) {
    ret = fi_close(rdm->pep);
    if (ret)
        return ret;
    rdm->pep = NULL;
}
```
But realistically, I don't know why close should ever fail in a way that the app can recover from it.  My guess is that most close failures are the result of EBUSY because a reference is held on the object.  But that's a coding issue in closing objects in the wrong order and not something that can be handled at run time.

Regardless, I'll still make the xnet_rdm_close() a little more robust.

tschuett:
For debugging, I would prefer a stack over reference counting. Then I can exactly find out in which order I have to close my objects.

PR #8229, opened 2022-11-09
prov/net: Fix loopback receive handling + other updates

shefty:
prov/net: Release lock in case of error in rdm_close
fabtests/loopback: Update to use loopback IP address, fix setting hints, and use tagged messages


shefty:
AWS errors were all in sighandler_test for different providers (2 efa, tcp, and sockets).

shefty:
bot:aws:retest

PR #8233, opened 2022-11-10
fabtests/sighandler_test: give child more time to handle signal

wzamazon:
This patch gives child more time to handle SIGINT. This allowes 3rd party library signal handler to have enough time to handle the signal.


j-xiong (approved):
Looks fine to me. Just a question, how was the 5 second value chosen?

wzamazon:
> how was the 5 second value chosen?

The Synapse AI SDK register a signal handler, and the handler takes a while for the 1st time, causing some AWS CI to fail.

I just picked a number that stopped the failure.

PR #8235, opened 2022-11-12
prov/efa, pytest/efa: fix bug and enable parallel test

wzamazon:
Fix a bug in efa provider that caused parallel test to fail, and also enable parallel test.


wzamazon:
bot:aws:retest

wenduwan (approved):

PR #8238, opened 2022-11-14
v1.17.0rc1

shefty:

Merging this change into main prior to creating v1.17.x branch.


shefty:
The fabtests makefile.am is missing an include file needed to build its release package.  I've updated the patch here with the missing file.  I'm waiting for this PR to finish CI then I'll push the updated patch directly, branch to v1.17.x, and create the rc1 package.

PR #8241, opened 2022-11-15
prov/efa: fix the calculcation of REQ header size for a packet entry

wzamazon:
Currently rxr_pkt_req_hdr_size_from_pkt_entry() uses rxr_pkt_req_hdr_size() to calculate hdr size, which assumes the the raw address header size is always RXR_REQ_OPT_RAW_ADDR_HDR_SIZE.

The assumption is wrong because older version of efa provider uses a different opt raw address header size. The correct addres lenght is stored in the "addr_len" field of the raw_addr_header.

This patch addressed the issue by making
rxr_pkt_req_hdr_size_from_pkt_entry() not to call rxr_pkt_req_hdr_size(), but calculate the size by itself.

### DIFF
--- a/prov/efa/src/rxr/rxr_pkt_type_req.c
+++ b/prov/efa/src/rxr/rxr_pkt_type_req.c
@@ -331,14 +331,49 @@ uint32_t *rxr_pkt_req_connid_ptr(struct rxr_pkt_entry *pkt_entry)
 	return NULL;
 }
 
+/**
+ * @brief calculate the exact header size for a given REQ pkt_entry
+ *
+ * The difference between this function and rxr_pkt_req_hdr_size() is
+ * the handling of the size of req opt raw address header.
+ *
+ * rxr_pkt_req_hdr_size() always use RXR_REQ_OPT_RAW_ADDR_HDR_SIZE, while
+ * this function pull raw address from pkt_entry's size.
+ *
+ * The difference is because older version of libfabric EFA provider uses
+ * a different opt header size.
+ *
+ * @param[in]	pkt_entry		packet entry
+ * @return 	header size of the REQ packet entry
+ */
 size_t rxr_pkt_req_hdr_size_from_pkt_entry(struct rxr_pkt_entry *pkt_entry)
 {
+	char *opt_hdr;
 	struct rxr_base_hdr *base_hdr;
-	uint32_t rma_iov_count;
+	struct rxr_req_opt_raw_addr_hdr *raw_addr_hdr;
 
 	base_hdr = rxr_get_base_hdr(pkt_entry->pkt);
-	rma_iov_count = rxr_pkt_hdr_rma_iov_count(pkt_entry);
-	return rxr_pkt_req_hdr_size(base_hdr->type, base_hdr->flags, rma_iov_count);
+	opt_hdr = (char *)pkt_entry->pkt + rxr_pkt_req_base_hdr_size(pkt_entry);
+
+	/*
+	 * It is not possible to have both optional raw addr header and optional
+	 * connid header in a packet header.
+	 */
+	if (base_hdr->flags & RXR_REQ_OPT_RAW_ADDR_HDR) {
+		assert(!(base_hdr->flags & RXR_PKT_CONNID_HDR));
+		raw_addr_hdr = (struct rxr_req_opt_raw_addr_hdr *)opt_hdr;
+		opt_hdr += sizeof(struct rxr_req_opt_raw_addr_hdr) + raw_addr_hdr->addr_len;
+	}
+
+	if (base_hdr->flags & RXR_REQ_OPT_CQ_DATA_HDR)
+		opt_hdr += sizeof(struct rxr_req_opt_cq_data_hdr);
+
+	if (base_hdr->flags & RXR_PKT_CONNID_HDR) {
+		assert(!(base_hdr->flags & RXR_REQ_OPT_RAW_ADDR_HDR));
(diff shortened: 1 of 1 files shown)


prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -331,14 +331,40 @@ uint32_t *rxr_pkt_req_connid_ptr(struct rxr_pkt_entry *pkt_entry)
...
 {
+	char *opt_hdr;
 	struct rxr_base_hdr *base_hdr;
-	uint32_t rma_iov_count;
+	struct rxr_req_opt_raw_addr_hdr *raw_addr_hdr;
 
 	base_hdr = rxr_get_base_hdr(pkt_entry->pkt);
-	rma_iov_count = rxr_pkt_hdr_rma_iov_count(pkt_entry);
-	return rxr_pkt_req_hdr_size(base_hdr->type, base_hdr->flags, rma_iov_count);
+	opt_hdr = (char *)pkt_entry->pkt + rxr_pkt_req_base_hdr_size(pkt_entry);
```
shijin-aws, line 347: could you quickly point out what difference this is from `rxr_pkt_req_hdr_size`? 

```
size_t rxr_pkt_req_hdr_size(struct rxr_pkt_entry *pkt_entry)
{ 
...
        /*
	 * It is not possible to have both optional raw addr header and optional
	 * connid header in a packet header.
	 */
	if (base_hdr->flags & RXR_REQ_OPT_RAW_ADDR_HDR) {
		assert(!(base_hdr->flags & RXR_PKT_CONNID_HDR));
		raw_addr_hdr = (struct rxr_req_opt_raw_addr_hdr *)opt_hdr;
		opt_hdr += sizeof(struct rxr_req_opt_raw_addr_hdr) + raw_addr_hdr->addr_len;
	}

	if (base_hdr->flags & RXR_REQ_OPT_CQ_DATA_HDR)
		opt_hdr += sizeof(struct rxr_req_opt_cq_data_hdr);

	if (base_hdr->flags & RXR_PKT_CONNID_HDR) {
		assert(!(base_hdr->flags & RXR_REQ_OPT_RAW_ADDR_HDR));
		opt_hdr += sizeof(struct rxr_req_opt_connid_hdr);
	}

	return opt_hdr - (char *)pkt_entry->pkt;
```

I cannot tell the difference

shijin-aws: Never mind, I was looking at an older branch. I see.

```
int hdr_size = REQ_INF_LIST[pkt_type].base_hdr_size;

	if (flags & RXR_REQ_OPT_RAW_ADDR_HDR) {
		/* It is impossible to have both optional connid hdr and opt_raw_addr_hdr
		 * in the header, and length of opt raw addr hdr is larger than
		 * connid hdr (which is confirmed by the following assertion).
		 */
		assert(RXR_REQ_OPT_RAW_ADDR_HDR_SIZE >= sizeof(struct rxr_req_opt_connid_hdr));
		hdr_size += RXR_REQ_OPT_RAW_ADDR_HDR_SIZE;
```

shijin-aws (approved):

prov/efa/src/rxr/rxr_pkt_type_req.c
```diff
@@ -331,14 +331,40 @@ uint32_t *rxr_pkt_req_connid_ptr(struct rxr_pkt_entry *pkt_entry)
 	return NULL;
 }
 
+/**
+ * @brief calculate the exact header size for a given REQ pkt_entry
+ *
+ * @param[in]	pkt_entry		packet entry
+ * @return 	header size of the REQ packet entry
+ */
 size_t rxr_pkt_req_hdr_size_from_pkt_entry(struct rxr_pkt_entry *pkt_entry)
```
shijin-aws, line 340: Probably better to explain why `rxr_pkt_req_hdr_size` cannot be used here.

wzamazon: Thanks! I added a comment about the difference between two functions.

wzamazon:
@aingerson 

Thank you for the explanation!

@shefty 

Is it ok that I add a call to `ft_sync` in `ft_exchange_keys` to older version of fabtests?

This change would also allow older version of fabtests to talk to newer version.

shefty:
Why do we want to support interoperability between different versions of fabtests?  We don't test for that and the overhead of trying to maintain wire protocol compatibility is unnecessary.  We don't even expect the latest version of fabtests to even run with an older version of libfabric.  That adds too much development overhead on test code where it isn't needed.

wzamazon:
@shefty 

I think your comments is about https://github.com/ofiwg/libfabric/pull/8243.

There are requests for EFA provider to maintain backward compatibility, and the provider was developed with that in mind.

e were thinking to use fabtests to test that feature. I think we will find other ways to test that.

shefty:
@wzamazon - You posted your response to #8243 here, which is why I also responded here.

A *newer* version of fabtests may not work with an *older* version of libfabric, but an *older* version of fabtests must be able to work with a *newer* version of libfabric.  The emphasis is that libfabric must ensure compatibility.

You can test, say, v1.14.0 fabtests with v1.14.0 libfabric and v1.14.0 fabtests with v1.17.0 libfabric to verify that compatibility is maintained.

However, the v1.17.0 fabtests may test functionality that wasn't available or present in an earlier version of libfabric.  That's why we don't want to enforce that fabtests be backwards compatible.

wzamazon:
@shefty 

I see. Thank you for explanation.

This make sense!

PR #8242, opened 2022-11-15
prov/net: Process response messages past an unexpected message

shefty:
If we're waiting for an ACK or an RMA read response, it may be
    blocked behind a unexpected receive message.  Unfortunately, this
    is a common MPI flow:
    
        for each rank
            RMA read()
        barrier()
    
    The barrier messages may begin to show up from peers before
    the local rank has finished its RMA read operations.  And the
    local rank won't post the receive message for the barrier until
    the RMA read operations have completed.
    
    There seems to be some sort of assumption that RMA read responses
    have a separate from from messages, which is true when using an
    RDMA device.  However, it's not true for tcp streams.  To handle
    this case, we need to look ahead in the tcp stream for the response,
    queuing the barrier message.
    
    To minimize the changes to the provider and avoid re-introducing
    unexpected message buffering at the receiver, we limit peeking
    ahead in the tcp stream to a single 0-byte message (what MPI uses
    for barrier), at least for now.

### DIFF
--- a/fabtests/functional/loopback.c
+++ b/fabtests/functional/loopback.c
@@ -88,7 +88,7 @@ int main(int argc, char **argv)
 		return EXIT_FAILURE;
 
 	opts.src_addr = "127.0.0.1";
-	hints->caps |= FI_LOCAL_COMM;
+	hints->caps = FI_LOCAL_COMM | FI_MSG | FI_TAGGED;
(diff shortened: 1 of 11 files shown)


shefty:
This requires a more comprehensive update.  That's in progress now.

PR #8243, opened 2022-11-15
common/shared.c: remove extra ft_sync() in ft_exchange_keys()

wzamazon:
Previous commit 4ea2df52e0dd2bdc3b29a627f85d7c17236838ad added a call to ft_sync() to the end of ft_exchange_keys(), which broke the backward compatiblity with previous version of fabtests.

This patch removed it.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -1595,7 +1595,7 @@ int ft_exchange_keys(struct fi_rma_iov *peer_iov)
 	if (ret)
 		return ret;
 
-	return ft_sync();
+	return 0;
 }
 
 static void ft_cleanup_mr_array(struct ft_context *ctx_arr, char **mr_bufs)


wzamazon:
@aingerson interested in your opinion about the ft_sync added.

aingerson:
@wzamazon I think the idea behind the sync was that the patch that added it (which combined and simplified the MR registration code) changed the keys exchange from a
```
if (server) {
    tx();
    rx();
} else {
    rx();
    tx();
}
```
to
```
tx();
rx();
```
In the tests, the testing begins right after exchange_keys is called, so I think the idea behind the ft_sync was to make sure we get both sides synchronized before starting any testing since the portion after the previous receive on both s sides (lines 1577-1597) includes a memcpy which could be a device copy as well. So it was to make sure to exclude that from any performance data.

shefty:
What do you mean by 'broke backward compatibility'?

We don't support different versions of fabtests talking with each other, or even the current version of fabtests working with an older version of libfabric.

wzamazon:
>We don't support different versions of fabtests talking with each other, or even the current version of fabtests working with an older version of libfabric.

Actually that is what I was trying to do, and  from what I can tell different version of fabtests can talk to each other just fine with this change.

PR #8244, opened 2022-11-16
[v1.17.x]prov/efa: fix the calculcation of REQ header size for a packet entry

wzamazon:
Currently rxr_pkt_req_hdr_size_from_pkt_entry() uses rxr_pkt_req_hdr_size() to calculate hdr size, which assumes the the raw address header size is always RXR_REQ_OPT_RAW_ADDR_HDR_SIZE.

The assumption is wrong because older version of efa provider uses a different opt raw address header size. The correct addres lenght is stored in the "addr_len" field of the raw_addr_header.

This patch addressed the issue by making
rxr_pkt_req_hdr_size_from_pkt_entry() not to call rxr_pkt_req_hdr_size(), but calculate the size by itself.

(cherry picked from commit 3f9b6cd8c21713a92678e14eb12134a46d31fc1a)

### DIFF
--- a/prov/efa/src/rxr/rxr_pkt_type_req.c
+++ b/prov/efa/src/rxr/rxr_pkt_type_req.c
@@ -331,14 +331,49 @@ uint32_t *rxr_pkt_req_connid_ptr(struct rxr_pkt_entry *pkt_entry)
 	return NULL;
 }
 
+/**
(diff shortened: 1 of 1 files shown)


wckzhang:
calculcation -> calculation

PR #8247, opened 2022-11-16
github/workflows: update tested ubuntu versions

ooststep:
ubuntu 18.04 runners are being deprecated, move to ubuntu-latest

Signed-off-by: Stephen Oost <stephen.oost@intel.com>

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -6,7 +6,6 @@ env:
     abi-dumper
     build-essential
     debhelper
-    dh-systemd
     fakeroot
     gcc
     git
@@ -18,7 +17,7 @@ env:
     ninja-build
     pandoc
     pkg-config
-    python
+    python-is-python3
     rpm
     sparse
     valgrind
@@ -43,8 +42,8 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-18.04
           - ubuntu-20.04
+          - ubuntu-22.04
         cc:
           - gcc
           - clang
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -9,7 +9,6 @@ env:
     abi-dumper
     build-essential
     debhelper
-    dh-systemd
     fakeroot
     gcc
     git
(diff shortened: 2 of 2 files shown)


.github/workflows/pr-ci.yml
```diff
@@ -43,8 +43,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-18.04
-          - ubuntu-20.04
+          - ubuntu-latest
```
shefty, line 46: It's actually useful to test on all available ubuntu options, as it can detect different issues.  What versions are now usable?

ooststep: According to [github documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), ubuntu-20.04 (ubuntu-latest) and ubuntu-22.04.  I'm not sure why latest points to the older version.

ooststep: I'll update this one to point to the explicit versions - should coverity remain on latest?

shefty:
errors on 22.04
```
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package dh-systemd
Error: Process completed with exit code 100.
```
re-running CI produced same errors.

.github/workflows/pr-ci.yml
rajachan: That doc does say `-latest` is the latest stable version, not necessarily the latest release. +1 to explicitly use the version numbers instead of latest.

Coverity is very picky, I was glad I got it working in the last go around, but just sneezing in that general direction might break it.  I do see a new release at https://scan.coverity.com/download/, so if the 2020.9 version we use doesn't work with ubuntu-20.04, you might want to try their release from this past August.

ooststep: thanks for the tips @rajachan, will take a look if we have issues.

ooststep:
> errors on 22.04
> 
> ```
> Reading package lists...
> Reading package lists...
> Building dependency tree...
> Reading state information...
> E: Unable to locate package dh-systemd
> Error: Process completed with exit code 100.
> ```
> 
> re-running CI produced same errors.

I was playing around with this some before break. it seems the packages we define may need some modifications to support the newer versions.

shefty:
Let's just see if the changes work.  We can revert if not.

PR #8248, opened 2022-11-17
ofi_hmem: new rdma flush function + prov/efa: Emulated Atomic Support for Cuda Devices

a-szegel:
Small change to ofi_hmem.h/cuda_hmem.c to add rdma flush function

Lots of changes to prov/efa to add emulated atomic support for CUDA devices.
1. Added result + fetch memory descriptors to rxr_atomic_ex
2. Switched every memcpy to be ofi_hmem memcpy
3. On the target side, added ofi_hmem memcpy's around emulated atomic impl functions, and performed them on host memory.

Testing:
```
fi_rdm_atomic -D cuda -p efa -E  # passes, and exits cleanly
fi_rdm_atomic -p efa -E                # passes, and exits cleanly
```

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -171,6 +171,7 @@ int cuda_gdrcopy_hmem_cleanup(void);
 int cuda_gdrcopy_dev_register(struct fi_mr_attr *mr_attr, uint64_t *handle);
 int cuda_gdrcopy_dev_unregister(uint64_t handle);
 int cuda_set_sync_memops(void *ptr);
+int cuda_flush_rdma_operations(void);
 
 #define ZE_MAX_DEVICES 8
 int ze_hmem_copy(uint64_t device, void *dst, const void *src, size_t size);
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -290,6 +290,18 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
 			     size_t hmem_iov_count, uint64_t hmem_iov_offset,
 			     const void *src, size_t size);
 
+static inline int ofi_copy_to_hmem(enum fi_hmem_iface iface, uint64_t device,
+				   void *dest, const void *src, size_t size)
+{
+	return hmem_ops[iface].copy_to_hmem(device, dest, src, size);
+}
+
+static inline int ofi_copy_from_hmem(enum fi_hmem_iface iface, uint64_t device,
+				     void *dest, const void *src, size_t size)
+{
+	return hmem_ops[iface].copy_from_hmem(device, dest, src, size);
+}
+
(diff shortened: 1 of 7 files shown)


src/hmem_cuda.c
```diff
@@ -760,6 +760,12 @@ bool cuda_is_gdrcopy_enabled(void)
 	return hmem_cuda_use_gdrcopy;
 }
 
+int cuda_flush_rdma_operations(void) {
+	enum cudaFlushGPUDirectRDMAWritesTarget target = cudaFlushGPUDirectRDMAWritesTargetCurrentDevice;
+	enum cudaFlushGPUDirectRDMAWritesScope scope = cudaFlushGPUDirectRDMAWritesToOwner;
+	return cudaDeviceFlushGPUDirectRDMAWrites(target, scope);
+}
```
shefty, line 767: Please fix formatting.  Opening brace for function should be on line by itself.  Limit line lengths to about 80 characters.  target and scope variables don't even look needed.

include/ofi_hmem.h
```diff
@@ -291,6 +291,11 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
 			     size_t hmem_iov_count, uint64_t hmem_iov_offset,
 			     const void *src, size_t size);
 
+int ofi_copy_to_hmem(enum fi_hmem_iface iface, uint64_t device,
+				   void *dest, const void *src, size_t size);
+int ofi_copy_from_hmem(enum fi_hmem_iface iface, uint64_t device,
+				     void *dest, const void *src, size_t size);
```
shefty, line 297: Move this set of changes into a separate patch.  Just move the entire static inline here

shefty (changes requested):
Last commit is also missing signed off by line

src/hmem_cuda.c
```diff
@@ -760,6 +760,12 @@ bool cuda_is_gdrcopy_enabled(void)
...
+
 #else
```
shijin-aws, line 770: Since you declare this new function for both HAVE_CUDA and not HAVE_CUDA, please add a stub function (return `-FI_ENOSYS`) for  `cuda_flush_rdma_operations` in the `#else` block as well .

shijin-aws (changes requested):

src/hmem_cuda.c
a-szegel: fixed

include/ofi_hmem.h
a-szegel: fixed (gave you 1 more than you asked for)

src/hmem_cuda.c
a-szegel: fixed.

a-szegel:
splitting into two pull requests.

PR #8250, opened 2022-11-17
util: Minor fixes for ofi_epoll and util_cq

sydidelot:
A small set of minor fixes for ofi_epoll and util_cq


shefty (approved):
Thanks - I actually had a note to remove hot_index and just hadn't gotten to it.

shefty:
Intel CI failures are problem in the build script, just not sure where the problems originated.

PR #8251, opened 2022-11-17
ofi_hmem: new rdma flush function + warning cleanup + change hmem_memcpy function scope

a-szegel:
Small change to ofi_hmem.h/cuda_hmem.c to add rdma flush function.  Moved ofi_copy_to_hmem() and ofi_copy_from_hmem() to public scope.  Removed a warning in hmem_cuda.c.

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -56,7 +56,7 @@ struct cuda_ops {
 	CUresult (*cuPointerGetAttribute)(void *data,
 					  CUpointer_attribute attribute,
 					  CUdeviceptr ptr);
-	CUresult (*cuPointerSetAttribute)(void *data,
+	CUresult (*cuPointerSetAttribute)(const void *data,
 					  CUpointer_attribute attribute,
 					  CUdeviceptr ptr);
 	CUresult (*cuMemGetAddressRange)( CUdeviceptr* pbase,
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -73,6 +73,11 @@ struct cuda_ops {
 	cudaError_t (*cudaIpcGetMemHandle)(cudaIpcMemHandle_t *handle,
 					   void *devptr);
 	cudaError_t (*cudaIpcCloseMemHandle)(void *devptr);
+#if CUDART_VERSION >= 11030
+	cudaError_t (*cudaDeviceFlushGPUDirectRDMAWrites)(
+				enum cudaFlushGPUDirectRDMAWritesTarget target,
+				enum cudaFlushGPUDirectRDMAWritesScope scope);
+#endif
 };
 
 static bool hmem_cuda_use_gdrcopy;
@@ -111,7 +116,11 @@ static struct cuda_ops cuda_ops = {
 	.cudaSetDevice = cudaSetDevice,
 	.cudaIpcOpenMemHandle = cudaIpcOpenMemHandle,
 	.cudaIpcGetMemHandle = cudaIpcGetMemHandle,
-	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle
+	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle,
+
+#if CUDART_VERSION >= 11030
+	.cudaDeviceFlushGPUDirectRDMAWrites = cudaDeviceFlushGPUDirectRDMAWrites,
+#endif
 };
 
 #endif /* ENABLE_CUDA_DLOPEN */
@@ -538,6 +547,16 @@ static int cuda_hmem_dl_init(void)
 		goto err_dlclose_cuda;
 	}
 
+#if CUDART_VERSION >= 11030
+	cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites = dlsym(cudart_handle,
+					    STRINGIFY(cudaDeviceFlushGPUDirectRDMAWrites));
+	if (!cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites) {
+		FI_WARN(&core_prov, FI_LOG_CORE,
+			"Failed to find cudaDeviceFlushGPUDirectRDMAWrites\n");
+		goto err_dlclose_cuda;
+	}
+#endif
+
 	return FI_SUCCESS;
 
 err_dlclose_cuda:
@@ -760,6 +779,25 @@ bool cuda_is_gdrcopy_enabled(void)
 	return hmem_cuda_use_gdrcopy;
 }
 
+#if CUDART_VERSION >= 11030
+int cuda_flush_rdma_operations(void)
+{
+	if (cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites) {
+		return cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites(
+				cudaFlushGPUDirectRDMAWritesTargetCurrentDevice,
+				cudaFlushGPUDirectRDMAWritesToOwner);
+	}
+	return -FI_ENOSYS;
+}
+#else
+int cuda_flush_rdma_operations(void)
+{
+	return -FI_ENOSYS;
+}
+#endif
+
+
+
 #else
 
 int cuda_copy_to_dev(uint64_t device, void *dev, const void *host, size_t size)
@@ -845,5 +883,10 @@ bool cuda_is_gdrcopy_enabled(void)
 int cuda_set_sync_memops(void *ptr)
 {
         return FI_SUCCESS;
+}
+
+int cuda_flush_rdma_operations(void)
+{
+	return -FI_ENOSYS;
 }
 #endif /* HAVE_CUDA */
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -171,6 +171,7 @@ int cuda_gdrcopy_hmem_cleanup(void);
 int cuda_gdrcopy_dev_register(struct fi_mr_attr *mr_attr, uint64_t *handle);
 int cuda_gdrcopy_dev_unregister(uint64_t handle);
 int cuda_set_sync_memops(void *ptr);
+int cuda_flush_rdma_operations(void);
 
 #define ZE_MAX_DEVICES 8
 int ze_hmem_copy(uint64_t device, void *dst, const void *src, size_t size);
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -290,6 +290,18 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
 			     size_t hmem_iov_count, uint64_t hmem_iov_offset,
 			     const void *src, size_t size);
 
+static inline int ofi_copy_to_hmem(enum fi_hmem_iface iface, uint64_t device,
+				   void *dest, const void *src, size_t size)
+{
+	return hmem_ops[iface].copy_to_hmem(device, dest, src, size);
+}
+
+static inline int ofi_copy_from_hmem(enum fi_hmem_iface iface, uint64_t device,
+				     void *dest, const void *src, size_t size)
+{
+	return hmem_ops[iface].copy_from_hmem(device, dest, src, size);
+}
+
 int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *base_addr, void **handle);
 int ofi_hmem_open_handle(enum fi_hmem_iface iface, void **handle,
 			 uint64_t device, void **mapped_addr);
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -131,18 +131,6 @@ struct ofi_hmem_ops hmem_ops[] = {
 	},
 };
 
-static inline int ofi_copy_to_hmem(enum fi_hmem_iface iface, uint64_t device,
-				   void *dest, const void *src, size_t size)
-{
-	return hmem_ops[iface].copy_to_hmem(device, dest, src, size);
-}
-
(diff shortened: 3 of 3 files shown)


wzamazon (approved):

a-szegel:
Following directions here to bump cuda version:

https://developer.nvidia.com/cuda-11-6-0-download-archive?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=20.04&target_type=runfile_local

a-szegel:
I need to figure out the minimum version cudaDeviceFlushGPUDirectRDMAWrites was installed in and add a #ifdef to noop if it isn't available.

src/hmem_cuda.c
```diff
@@ -760,6 +760,16 @@ bool cuda_is_gdrcopy_enabled(void)
 	return hmem_cuda_use_gdrcopy;
 }
 
+int cuda_flush_rdma_operations(void)
+{
+#if (CUDART_VERSION >= 1130)
```
darrylabbate, line 765: You need to use `11030` to specify version 11.3

darrylabbate: Also, you might be asked to use the `#if` directive to define 2 separate functions instead of inlining the preprocessor logic in this single function.

a-szegel: thanks!

darrylabbate (approved):

src/hmem_cuda.c
```diff
@@ -760,6 +760,19 @@ bool cuda_is_gdrcopy_enabled(void)
 	return hmem_cuda_use_gdrcopy;
 }
 
+#if CUDA_VERSION >= 11030
```
darrylabbate, line 763: Just noticed this: your earlier change used `CUDART_VERSION`; this one uses `CUDA_VERSION`. What's the distinction?

src/hmem_cuda.c
```diff
@@ -760,6 +760,19 @@ bool cuda_is_gdrcopy_enabled(void)
...
+int cuda_flush_rdma_operations(void)
+{
+	return cudaDeviceFlushGPUDirectRDMAWrites(cudaFlushGPUDirectRDMAWritesTargetCurrentDevice,
```
shijin-aws, line 766: Does it work with dlopen case? Is `cudaDeviceFlushGPUDirectRDMAWrites` part of the `hmem_ops[cuda]` function table?

shijin-aws: You need to add this function to 

```
static struct cuda_ops cuda_ops = {
	.cudaMemcpy = cudaMemcpy,
	.cudaDeviceSynchronize = cudaDeviceSynchronize,
	.cudaFree = cudaFree,
	.cudaMalloc = cudaMalloc,
	.cudaGetErrorName = cudaGetErrorName,
	.cudaGetErrorString = cudaGetErrorString,
	.cuGetErrorName = cuGetErrorName,
	.cuGetErrorString = cuGetErrorString,
	.cuPointerGetAttribute = cuPointerGetAttribute,
	.cuPointerSetAttribute = cuPointerSetAttribute,
	.cuMemGetAddressRange = cuMemGetAddressRange,
	.cudaHostRegister = cudaHostRegister,
	.cudaHostUnregister = cudaHostUnregister,
	.cudaGetDeviceCount = cudaGetDeviceCount,
	.cudaGetDevice = cudaGetDevice,
	.cudaSetDevice = cudaSetDevice,
	.cudaIpcOpenMemHandle = cudaIpcOpenMemHandle,
	.cudaIpcGetMemHandle = cudaIpcGetMemHandle,
	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle
};
```

And initialize it as `cudaDeviceFlushGPUDirectRDMAWrites` (non-dlopen) and via dlsym (dlopen case) . See how other functions in the table are defined.

shijin-aws (changes requested):

src/hmem_cuda.c
a-szegel: I forgot I did that... RT == Runtime API.  If a function call starts with cuda, it uses the runtime API, if it starts with cu, it used the driver API.... Should be RT.

shijin-aws:
The updated PR still misses the dlopen path, see how it is done in https://github.com/ofiwg/libfabric/blob/main/src/hmem_cuda.c#L379-L539

src/hmem_cuda.c
```diff
@@ -56,7 +56,7 @@ struct cuda_ops {
 	CUresult (*cuPointerGetAttribute)(void *data,
 					  CUpointer_attribute attribute,
 					  CUdeviceptr ptr);
-	CUresult (*cuPointerSetAttribute)(void *data,
+	CUresult (*cuPointerSetAttribute)(const void *data,
```
shefty, line 59: Leaving comment here, since I can't comment on the commit message directly.  Please use hard returns at the end of lines, about 80 characters, in the commit message.  Otherwise, the text does not show properly when viewing via git log.

src/hmem_cuda.c
```diff
@@ -760,6 +763,19 @@ bool cuda_is_gdrcopy_enabled(void)
 	return hmem_cuda_use_gdrcopy;
 }
 
+#if CUDA_VERSION >= 11030
+int cuda_flush_rdma_operations(void)
+{
+	return cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites(cudaFlushGPUDirectRDMAWritesTargetCurrentDevice,
```
shefty, line 769: CUDAIsAsBadAsWindowsForExcessivelyLongNamesThatAreAnEntireManPage.  Please move the parameter to the next line to keep the ending character closer to 80 characters.  Long lines are hard to read and frequently requires that the reader scroll left/right, which is a pain.  This comment applies to the function declaration above in cuda_ops as well.

src/hmem_cuda.c
```diff
@@ -538,6 +542,13 @@ static int cuda_hmem_dl_init(void)
 		goto err_dlclose_cuda;
 	}
 
+	cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites = dlsym(cudart_handle,
```
shijin-aws, line 545: do you need to check cuda version here? This will fail for cuda <=11.3 right?

shijin-aws (changes requested):

src/hmem_cuda.c
```diff
@@ -111,7 +114,8 @@ static struct cuda_ops cuda_ops = {
 	.cudaSetDevice = cudaSetDevice,
 	.cudaIpcOpenMemHandle = cudaIpcOpenMemHandle,
 	.cudaIpcGetMemHandle = cudaIpcGetMemHandle,
-	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle
+	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle,
+	.cudaDeviceFlushGPUDirectRDMAWrites = cudaDeviceFlushGPUDirectRDMAWrites
```
shijin-aws, line 118: cuda version check?

src/hmem_cuda.c
```diff
@@ -760,6 +771,20 @@ bool cuda_is_gdrcopy_enabled(void)
 	return hmem_cuda_use_gdrcopy;
 }
 
+#if CUDART_VERSION >= 11030
```
shijin-aws, line 774: I think you need to move this macro to the code that assigns `cuda_ops.cudaDeviceFlushGPUDirectRDMAWrites`. See my other 2 comments for dlopen/non-dlopen cases.

src/hmem_cuda.c
```diff
@@ -111,7 +114,13 @@ static struct cuda_ops cuda_ops = {
 	.cudaSetDevice = cudaSetDevice,
 	.cudaIpcOpenMemHandle = cudaIpcOpenMemHandle,
 	.cudaIpcGetMemHandle = cudaIpcGetMemHandle,
-	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle
+	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle,
+
+#if CUDART_VERSION >= 11030
+	.cudaDeviceFlushGPUDirectRDMAWrites = cudaDeviceFlushGPUDirectRDMAWrites,
+#else
+    .cudaDeviceFlushGPUDirectRDMAWrites = NULL,
```
shijin-aws, line 122: indent is off.

src/hmem_cuda.c
shijin-aws: Bump comment.

src/hmem_cuda.c
a-szegel: I did this.

src/hmem_cuda.c
a-szegel: I did it elsewhere but I can bring it here... I think thats better here.

a-szegel: Done on next push.

src/hmem_cuda.c
a-szegel: agreed.

src/hmem_cuda.c
a-szegel: fixed.

src/hmem_cuda.c
a-szegel: done

src/hmem_cuda.c
a-szegel: done

src/hmem_cuda.c
a-szegel: done

src/hmem_cuda.c
a-szegel: done

shijin-aws (approved):

src/hmem_cuda.c
```diff
@@ -111,7 +114,13 @@ static struct cuda_ops cuda_ops = {
 	.cudaSetDevice = cudaSetDevice,
 	.cudaIpcOpenMemHandle = cudaIpcOpenMemHandle,
 	.cudaIpcGetMemHandle = cudaIpcGetMemHandle,
-	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle
+	.cudaIpcCloseMemHandle = cudaIpcCloseMemHandle,
+
+#if CUDART_VERSION >= 11030
```
shijin-aws, line 119: I am not sure whether checking this hard-coded version is the best approach. A better implementation could be having a dedicated macro like `HAVE_CUDADEVICEFLUSHGPUDIRECTRDMAWRITES` and use `AC_CHECK_DECL` in `configure.ac` to assign 1. See example in https://github.com/ofiwg/libfabric/blob/main/prov/efa/configure.m4#L93-L96.

But I am not sure whether it is over-engineering. I will let @shefty comment here.

shefty: Doing an explicit check against the feature is strongly preferred.  Though since we're loading it dynamically, what have the check at all?

shijin-aws: >Though since we're loading it dynamically

That only works for `--enable-cuda-dlopen`. If user drop this flag, libfabric will compile with `libcudart.so`. If that function is not available in cuda library, we will have a compilation error for `.cudaDeviceFlushGPUDirectRDMAWrites = cudaDeviceFlushGPUDirectRDMAWrites,` , right?

a-szegel:
Failed check b/c `enum cudaFlushGPUDirectRDMAWritesScope scope)` doesn't exist in early version... switching to int.

src/hmem_cuda.c
a-szegel: No I put a guard around that...
```
#if CUDART_VERSION >= 11030
	.cudaDeviceFlushGPUDirectRDMAWrites = cudaDeviceFlushGPUDirectRDMAWrites,
#else
	.cudaDeviceFlushGPUDirectRDMAWrites = NULL,
#endif
```

shijin-aws: I am not saying the current PR is wrong. I was answering Sean's question of 

>Though since we're loading it dynamically, what have the check at all?

I do not think we are loading it dynamically in all cases. This check should be in compilation as well. Though I was suggesting a different way to check that.

a-szegel:
Switching those to ints will solve the compile problem... but it will give me this warning b/c the real function signature has enums:
```
src/hmem_cuda.c:119:40: warning: initialization from incompatible pointer type [-Wincompatible-pointer-types]
  .cudaDeviceFlushGPUDirectRDMAWrites = cudaDeviceFlushGPUDirectRDMAWrites, 
```

Can add another guard to stop this warning from happening... but its getting messy.

shefty:
You should be able to do something like this:
```
int NocudaDeviceFlushGPUDirectRDMAWrites(enum, enum) /* match the function prototype */
{
    return -ENOSYS;
}

cudainit(...)
{
    void *func;

    func = dlsym(..., "cudaDeviceFlushGPUDirectRDMAWrites")
    if (func)
        cuda.cudaDeviceFlushGPUDirectRDMAWrites = func;
    else
        cuda.cudaDeviceFlushGPUDirectRDMAWrites = 
            NocudaDeviceFlushGPUDirectRDMAWrites;
}
```

shijin-aws:
> Failed check b/c `enum cudaFlushGPUDirectRDMAWritesScope scope)` doesn't exist in early version... switching to int.

Why not adding the version check around 

```
cudaError_t (*cudaDeviceFlushGPUDirectRDMAWrites)(
				enum cudaFlushGPUDirectRDMAWritesTarget target,
				enum cudaFlushGPUDirectRDMAWritesScope scope);
};
```
as well?

You definitely do not want to add this to `cuda_ops` declaration when cuda is too old.

a-szegel:
I think I finally got it.

a-szegel:
I think what I did is good... I only add the function pointer to the struct if it exists in the cuda version, and I have a NO-OP case defined of the func when the version is low.

PR #8259, opened 2022-11-19
prov/sockets: retry setting conn->av_index

hzhou:
When we make dynamic connections, it is possible for a connect message to arrive before the receiver issue fi_av_insert. This result in conn->av_index being set to FI_ADDR_NOTAVAIL, and leading to future message mismatch.

Retry setting the av_index fixes the issue. It works because the matching will always happen after the receiver issues fi_av_insert or it has to be a wild card receive.


shefty:
Can you pick for v1.17.x and v1.16.x, so this hits the releases?

hzhou:
@shefty Are you asking me to push another two pull requests against v1.16.x and v1.17.x?

shefty:
Yes, please.  At least for v1.17.x, so we can get it into the next major release.

PR #8261, opened 2022-11-21
common: add IFF_RUNNING check to indicate iface is up and running

nikitaxgusev:
Add checking for ifa_flags for IFF_RUNNING flag to filter out the interface, which is down and not running. In WinOS, there is not definition of IFF_RUNNING, reset it by IFF_UP.

Signed-off-by: Nikita Gusev <nikita.gusev@intel.com>

### DIFF
--- a/include/windows/ifaddrs.h
+++ b/include/windows/ifaddrs.h
@@ -15,6 +15,11 @@
 
 #include <Winsock2.h>
 
+/* IFF_RUNNING is not defined, reset it by IFF_UP */
+#ifndef IFF_RUNNING
+#define IFF_RUNNING IFF_UP
+#endif
+
 /* here is minimal subset of ifaddr API required for sockets & UDP
    providers */
 struct ifaddrs {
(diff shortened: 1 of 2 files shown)


src/common.c
```diff
@@ -2077,6 +2077,7 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 	for (ifa = ifaddrs; ifa != NULL; ifa = ifa->ifa_next) {
 		if (ifa->ifa_addr == NULL ||
 			!(ifa->ifa_flags & IFF_UP) ||
+			!(ifa->ifa_flags & IFF_RUNNING) ||
```
shefty, line 2080: IFF_RUNNING isn't defined on windows, resulting in a build error there.  At least IFF_UP and IFF_LOOPBACK are defined by windows, but I haven't figured out what header file defines them (guessing winsock2.h) somewhere.  The best option here may be to add this to include/windows/ifaddrs.h:
```
#ifndef IFF_RUNNING
#define IFF_RUNNING IFF_UP
#endif
```
It's a hack, but should work and is no worse off on windows than we are now.

nikitaxgusev: @shefty done, but will wait for testing and will re-push it since extra space at the end of the comment.

include/windows/ifaddrs.h
```diff
@@ -15,6 +15,11 @@
 
 #include <Winsock2.h>
 
+/* IFF_RUNNING is not defined, reset it by IFF_UP  */
```
nikitaxgusev, line 18: extra space at the end

nikitaxgusev: @shefty done

shefty:
Can you please cherry-pick for v1.17.x, so it gets pulled into the release?

PR #8262, opened 2022-11-21
prov/net: RX operations now support io_uring

sydidelot:
This patch adds support of io_uring to RX operations.

 The implementation is still incomplete because of the
 following missing features:
 1. We need a mechanism to cancel all pending io_uring operations
    and wait for them to complete before we can close the ring.
    Pending receive operations would cause the code to fail
    an assertion failure when the endpoint is shut down.
 2. Recv byteq is disabled as we need to agree on a model. I'm not
    sure if recv byteq is useful with io_uring since we can
    pre-post receive buffers.

### DIFF
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -417,6 +417,7 @@ struct ofi_bsock {
 	size_t zerocopy_size;
 	uint32_t async_index;
 	uint32_t done_index;
+	size_t async_prefetch;
 };
 
 static inline void
@@ -430,6 +431,7 @@ ofi_bsock_init(struct ofi_bsock *bsock, struct ofi_sockapi *sockapi,
 	ofi_byteq_init(&bsock->sq, sbuf_size);
 	ofi_byteq_init(&bsock->rq, rbuf_size);
 	bsock->zerocopy_size = SIZE_MAX;
+	bsock->async_prefetch = false;
 
 	/* first async op will wrap back to 0 as the starting index */
 	bsock->async_index = UINT32_MAX;
@@ -465,6 +467,7 @@ ssize_t ofi_bsock_recvv(struct ofi_bsock *bsock, struct iovec *iov,
 			size_t cnt, size_t *len);
 uint32_t ofi_bsock_async_done(const struct fi_provider *prov,
 			      struct ofi_bsock *bsock);
+void ofi_bsock_prefetch_done(struct ofi_bsock *bsock, size_t len);
 
 
 /*
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -460,9 +460,9 @@ ssize_t ofi_bsock_flush_sync(struct ofi_bsock *bsock);
 ssize_t ofi_bsock_send(struct ofi_bsock *bsock, const void *buf, size_t *len);
 ssize_t ofi_bsock_sendv(struct ofi_bsock *bsock, const struct iovec *iov,
 			size_t cnt, size_t *len);
-ssize_t ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t len);
+ssize_t ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t *len);
 ssize_t ofi_bsock_recvv(struct ofi_bsock *bsock, struct iovec *iov,
-			size_t cnt);
+			size_t cnt, size_t *len);
 uint32_t ofi_bsock_async_done(const struct fi_provider *prov,
 			      struct ofi_bsock *bsock);
 
--- a/prov/net/src/xnet_ep.c
+++ b/prov/net/src/xnet_ep.c
@@ -324,6 +324,8 @@ static void xnet_ep_flush_all_queues(struct xnet_ep *ep)
 	struct xnet_cq *cq;
 
 	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
+	assert(!ep->bsock.tx_sockctx.uring_sqe_inuse);
+	assert(!ep->bsock.rx_sockctx.uring_sqe_inuse);
 	cq = container_of(ep->util_ep.tx_cq, struct xnet_cq, util_cq);
 	if (ep->cur_tx.entry) {
 		ep->hdr_bswap(ep, &ep->cur_tx.entry->hdr.base_hdr);
--- a/prov/net/src/xnet_progress.c
+++ b/prov/net/src/xnet_progress.c
@@ -742,6 +742,36 @@ static ssize_t xnet_op_read_rsp(struct xnet_ep *ep)
 	return xnet_recv_msg_data(ep);
 }
 
+static int xnet_progress_hdr(struct xnet_ep *ep)
+{
+	if (ep->cur_rx.hdr_done == sizeof(ep->cur_rx.hdr.base_hdr)) {
+		assert(ep->cur_rx.hdr_len == sizeof(ep->cur_rx.hdr.base_hdr));
+
+		if (ep->cur_rx.hdr.base_hdr.hdr_size > XNET_MAX_HDR) {
+			FI_WARN(&xnet_prov, FI_LOG_EP_DATA,
+				"Payload offset is too large\n");
+			return -FI_EIO;
+		}
+		ep->cur_rx.hdr_len = (size_t) ep->cur_rx.hdr.base_hdr.hdr_size;
+	}
+
+	if (ep->cur_rx.hdr_done < ep->cur_rx.hdr_len)
+		return -FI_EAGAIN;
+
+	ep->hdr_bswap(ep, &ep->cur_rx.hdr.base_hdr);
+	assert(ep->cur_rx.hdr.base_hdr.id == ep->rx_id++);
+	if (ep->cur_rx.hdr.base_hdr.op >= ARRAY_SIZE(xnet_start_op)) {
+		FI_WARN(&xnet_prov, FI_LOG_EP_DATA,
+			"Received invalid opcode\n");
+		return -FI_EIO;
+	}
+
+	ep->cur_rx.data_left = ep->cur_rx.hdr.base_hdr.size -
+			       ep->cur_rx.hdr.base_hdr.hdr_size;
+	ep->cur_rx.handler = xnet_start_op[ep->cur_rx.hdr.base_hdr.op];
+	return FI_SUCCESS;
+}
+
 static ssize_t xnet_recv_hdr(struct xnet_ep *ep)
 {
 	size_t len;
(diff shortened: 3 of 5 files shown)


sydidelot:
@shefty Here is the PR that adds support of io_uring to RX operations. 

This support is still experimental and lacks the (important) following features:
1. We need a mechanism to cancel all pending io_uring operations and wait for them to complete before we can close the ring.
    Pending receive operations would cause the code to fail an assertion failure when the endpoint is shut down. My biggest problem is that `xnet_ep_flush_all_queues()` assumes that the cancellation of in-progress TX/RX entries is asynchronous, which is no longer true with io_uring (see the man page for `io_uring_prep_cancel()`). Is there a way to defer the destruction of the endpoint until all the cancellation of in-progress TX/RX entries is completed?
2. Recv byteq is disabled as we need to agree on a model. I'm not sure if recv byteq is useful with io_uring since we can
    pre-post receive buffers. 

We can use this PR as a placeholder for our discussions. Thank you!

shefty:
Can you please rebase?  There were critical changes to the net provider to allow receiving out of order messages, relative to the sender.  Those changes may impact the changes you made to progress.

sydidelot:
Thanks for the update! Let me know when you get a chance to review the io_uring changes. I will put this work on hold until I hear from you.

shefty:
See PR #8353 (which is still undergoing testing).  I added comments in my changes where I believe we'll need updates based on how io_uring Rx is implemented.  It ended up not being as bad as I anticipated, assuming that I also anticipated how io_uring will fit into those changes correctly.

My target for io_uring integration is still by the end of the year.

prov/net/src/xnet_progress.c
```diff
@@ -811,6 +706,53 @@ static ssize_t xnet_recv_hdr(struct xnet_ep *ep)
 	return ep->cur_rx.handler(ep);
 }
 
+static void xnet_complete_rx(struct xnet_ep *ep, ssize_t ret)
+{
+	struct xnet_xfer_entry *rx_entry;
+	struct util_cq *cq;
+
+	rx_entry = ep->cur_rx.entry;
+	if (!rx_entry) {
```
shefty, line 728: When would this if statement be true?

prov/net/src/xnet_ep.c
```diff
@@ -666,6 +666,9 @@ int xnet_endpoint(struct fid_domain *domain, struct fi_info *info,
 	if (ret)
 		goto err1;
 
+	if (xnet_io_uring)
+		xnet_prefetch_rbuf_size = 0;
```
shefty, line 672: This function (ep creation) doesn't look like the correct place to make this change to a global variable

src/common.c
```diff
@@ -1336,7 +1336,11 @@ ssize_t ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t len)
 		return bytes + ret;
 
 out:
-	assert(ret != -OFI_EINPROGRESS_URING);
+	if (ret == -OFI_EINPROGRESS_URING) {
+		assert(!bytes);
+		return ret;
+	}
```
shefty, line 1342: This comment is for the commit message, not this section of code.  The bsock prefetching receive data is helpful when dealing with a large number of small messages.

prov/net/src/xnet_progress.c
```diff
@@ -778,8 +778,14 @@ void xnet_progress_rx(struct xnet_ep *ep)
 			ret = ep->cur_rx.handler(ep);
 		}
 
-		if (OFI_SOCK_TRY_SND_RCV_AGAIN(-ret))
+		if (OFI_SOCK_TRY_SND_RCV_AGAIN(-ret)) {
+			xnet_update_pollflag(ep, POLLIN, true);
+			return;
+		} else if (ret == -OFI_EINPROGRESS_URING) {
+			assert(xnet_io_uring);
+			xnet_update_pollflag(ep, POLLIN, false);
```
shefty, line 786: Where did the xnet_update_pollflag() calls above come from?  Or why are these needed here now and not before?

shefty:
Regarding the use of receive side buffering, consider this scenario:

1. We receive the base header
2. We receive the extended header
3. We receive the message data into the destination buffer

If we remove all receive buffering, then each of those calls results in some sort of user to kernel communication.  Without io_uring, those are kernel transitions.  With io_uring, we'll take a thread context switch (from the application's thread to some kernel thread processing the io_uring).  For small messages, or even messages that frequently use the extended headers (e.g. RMA operations or CQ data), I think we'll be worse off without the extra buffering.  Multiple small messages back-to-back could see a significant impact.

My guess is that whenever we're at state 1 and need to read a new base header, we'll be better off trying to fill the bsock receive buffer with whatever data is available.  That makes it likely all extended header data is readily available, along with all data from short message.

sydidelot:
Happy New Year @shefty! Sorry for the late reply, I was quite busy with some other topics! 

I understand your concerns about the removal of receive buffering with io_uring. What you suggest is a hybrid approach where we would buffer data using traditional recv() calls to speed up small messages and io_uring for large data transfers. 

One benefit of io_uring is to progress multiple connections by polling one single ring and therefore, limit the number of system calls. You suggestion would require to iterate through all the sockets that need rx progression (not to mention repeated calls to epoll_ctl() to update the poll flags), which is at the opposite of io_uring philosophy. 

I am pretty sure this problem can be addressed with pre-posted receive buffers (`IORING_FEAT_FAST_POLL`). I believe this feature can be used to buffer receive data, similarly to the buffered bsock data-structure. I would rather prefer to keep a full io_uring implementation than mixing io_uring with synchronous/traditional socket calls.

shefty:
I wasn't necessarily suggesting that we need a hybrid model, just that we may need to receive into small buffer, but one that's larger than the exact amount of data we're looking for.  This is only needed when we're trying to receive the start of a new message.  Most or my use cases involve tagged messages, so until I read the header, I don't know what application buffer needs to receive the data.

So, while the code is only wanting to receive 16-bytes in order to read the header, it would be better if the receive posted to the io_uring were larger.  That does mean that receive data may end up pending in a user buffer.  The current progress loops are structured to handle this.

The bsock code checks the size of the requested receive.  If it's smaller than half the bsock buffer size (9k by default), it swaps in the bsock buffer in place of the requested buffer.  Maybe a similar strategy can be used.  See ofi_bsock_recv().

sydidelot:
Initially, I thought about modifying `ofi_bsock_recvv()/ofi_bsock_recv()` similarly to the changes you've made to `ofi_bsock_send()/ofi_bsock_sendv` for asynchronous send: add a new input/output argument `size_t *len` where the function would return the number of bytes that were consumed from the _prefetch_ buffer (if the return value is 0 or `OFI_EINPROGRESS_URING`).
This way we wouldn't need to disable all receive buffering and we could keep the code generic whether io_uring is used or not.
The main problem is that we need to update all the callers of `ofi_bsock_recvv()/ofi_bsock_recv()`. What do you think?

shefty:
I'm okay if we need to update callers of ofi_bsock_recv().  But could you explain the flow more?

sydidelot:
@shefty I finally decided to implement the changes I had in mind in regards to `ofi_bsock_recv()`. The changes are available in the first 3 patches in the PR.

shefty:
I understand what you're trying to do with ofi_bsock_recv() now, and the concept makes sense.  What I don't like is that the value of len is modified somewhat randomly from the viewpoint of the caller and really only has meaning in the case where ret == -OFI_EINPROGRESS_URING.  I would carry the concept further (or is it farther?).  For example, have ofi_bsock_recv() return 0 on success, negative value on error, similar to nearly all other libfabric calls.  Always have it return the number of bytes received and available for processing.

I *think* ofi_bsock_send() always sets *len to the number of bytes sent (or queued).  The main difference relative to the above definition is that it may also return *len if *len > 0.  We could probably change that to just return 0 on success and always use the returned *len value.

include/ofi_net.h
```diff
@@ -430,6 +431,7 @@ ofi_bsock_init(struct ofi_bsock *bsock, struct ofi_sockapi *sockapi,
 	ofi_byteq_init(&bsock->sq, sbuf_size);
 	ofi_byteq_init(&bsock->rq, rbuf_size);
 	bsock->zerocopy_size = SIZE_MAX;
+	bsock->async_prefetch = false;
```
shefty, line 434: Either make async_prefetch a boolean value or size_t, mixing is confusing.

src/common.c
```diff
@@ -1403,6 +1407,16 @@ ssize_t ofi_bsock_recvv(struct ofi_bsock *bsock, struct iovec *iov, size_t cnt,
 	return ret ? -ofi_sockerr(): -FI_ENOTCONN;
 }
 
+void ofi_bsock_prefetch_done(struct ofi_bsock *bsock, size_t len)
```
shefty, line 1410: I think declaring this, but not using it results in a build warning if using git bisect

prov/net/src/xnet_progress.c
shefty: Wouldn't that case result in xnet_recv_hdr() returning -FI_EAGAIN?  See comment below.

prov/net/src/xnet_progress.c
```diff
@@ -914,10 +852,12 @@ void xnet_progress_rx(struct xnet_ep *ep)
 			ret = ep->cur_rx.handler(ep);
 		}
 
-	} while (!ret && ofi_bsock_readable(&ep->bsock));
+		if (OFI_SOCK_TRY_SND_RCV_AGAIN(-ret))
```
shefty: Should this check be expanded to include OFI_EINPROGRESS_URING?  It seems that OFI_EINPROGRESS_URING is equivalent to EAGAIN.

prov/net/src/xnet_progress.c
```diff
@@ -865,12 +865,22 @@ void xnet_progress_rx(struct xnet_ep *ep)
 			ret = ep->cur_rx.handler(ep);
 		}
 
-		if (OFI_SOCK_TRY_SND_RCV_AGAIN(-ret))
-			return;
+		if (OFI_SOCK_TRY_SND_RCV_AGAIN(-ret) ||
+		    ret == -OFI_EINPROGRESS_URING)
```
shefty, line 869: Oh, hey, you added that check in this patch.  Can you fixup (or remove) the braces following the if statement.

sydidelot:
@shefty I have addresses all your review comments in regards to the bsock interface. I have created a new PR that focuses on the bsock changes only: https://github.com/ofiwg/libfabric/pull/8394
I preferred not to mix up the bsock changes with the io_uring changes to the net provider. I will resume on #8262 (and address your other review comments) after #8394 is merged.

shefty:
I think this PR is no longer needed.

sydidelot:
@shefty I will need to reopen this PR (or create a new one) because #8394 was only focusing on bsock changes and we still need a couple of changes to the RX path of the net provider.

shefty:
Sorry about that.  I assumed all the patches had been moved there.  Either way works (re-open or new PR) once updated.

PR #8267, opened 2022-11-23
[v1.17.x] common: add IFF_RUNNING check to indicate iface is up and running

nikitaxgusev:
Add checking for ifa_flags for IFF_RUNNING flag to filter out the interface, which is down and not running. In WinOS, there is not definition of IFF_RUNNING, reset it by IFF_UP.

Signed-off-by: Nikita Gusev <nikita.gusev@intel.com>

### DIFF
--- a/include/windows/ifaddrs.h
+++ b/include/windows/ifaddrs.h
@@ -15,6 +15,11 @@
 
 #include <Winsock2.h>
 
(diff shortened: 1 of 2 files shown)


shefty:
Intel CI failure are related to shm tests with DSA.  Some sort of permission issue.

PR #8273, opened 2022-11-28
fabtests/pytest/efa: extend cuda client server test timeout

wenduwan:
Double cuda client-server test timeout to reduce noise.

### DIFF
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -10,7 +10,7 @@ def efa_run_client_server_test(cmdline_args, executable, iteration_type,
     # message sizes (especailly when running with multiple workers).
     timeout = None
     if "cuda" in memory_type:
-        timeout = max(1000, cmdline_args.timeout)
+        timeout = max(2000, cmdline_args.timeout)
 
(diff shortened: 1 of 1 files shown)


fabtests/pytest/efa/efa_common.py
```diff
@@ -10,7 +10,7 @@ def efa_run_client_server_test(cmdline_args, executable, iteration_type,
     # message sizes (especailly when running with multiple workers).
     timeout = None
     if "cuda" in memory_type:
-        timeout = max(1000, cmdline_args.timeout)
+        timeout = max(2000, cmdline_args.timeout)
```
wzamazon, line 13: I am not sure we want to extend this timeout (1000 is already quite large). I think we want to extend general timeout

wenduwan: Currently we are seeing cuda test timeout after 1000 seconds, which is long. Therefore I want to localize the change to cuda only to avoid extending timeout for other tests.

Could you clarify what `general timeout` refers to?

wenduwan: Synced with wzamazon@ offline. The cuda test timeout is likely caused by a bug. Increasing timeout is not the right solution. Closing PR.

PR #8275, opened 2022-11-29
contrib/intel/jenkins: add verbose summary to post stage

zachdworkin:
Add a printout of almost every test that passed to the verbose summary in post stage.
Tests that don't have unique names aren't printed.
Sometimes having a list of every test that gets run is better than just having the total number of tests run.


zachdworkin:
@shefty This did what I expected it to after the replays. It is good to merge now.

PR #8280, opened 2022-11-29
fabtests/component: Support FI_CONTEXT mode in fi-rdmabw-xe test

j-xiong:
This allows the test to work with more providers (e.g. psm3).

Also remove FI_HMEM from the hints since it is not a hard requirement for the test.

### DIFF
--- a/fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
+++ b/fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
@@ -75,9 +75,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdint.h>
+#include <stdbool.h>
 #include <unistd.h>
 #include <string.h>
 #include <errno.h>
+#include <assert.h>
 #include <arpa/inet.h>
 #include <rdma/fabric.h>
 #include <rdma/fi_domain.h>
@@ -90,6 +92,7 @@
 #include <level_zero/ze_api.h>
 #include "util.h"
 #include "xe.h"
+#include "ofi_ctx_pool.h"
 
 #define MAX_SIZE	(4*1024*1024)
 #define MIN_PROXY_BLOCK	(131072)
@@ -141,6 +144,8 @@ struct nic {
 
 static struct nic 		nics[MAX_NICS];
 static int			num_nics;
+
+struct context_pool		*context_pool;
 
 struct buf {
 	struct xe_buf		xe_buf;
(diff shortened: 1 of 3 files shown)


aingerson:
Why is FI_HMEM not a requirement for the test if you're passing device memory to a provider?

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
```diff
@@ -548,6 +551,65 @@ static void finalize_ofi(void)
...
+	assert(context_pool_initialized);
+
+	if (!ctxt)
+		return;
+
+	entry = container_of(ctxt, struct context_list, context);
+	entry->next = NULL;
+	free_tail->next = entry;
+	free_tail = entry;
+}
```
shefty, line 611: Can we move this to a common area?

shefty (commented):
Similar question on removing FI_HMEM.  That seems like it's inherent to the test.

j-xiong:
The memory type is a command line option, which can be device memory or host memory. I think we can have the cap set | unset depending on the memory type.

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
j-xiong: Like where? The component tests are mostly self-contained. Separating this part into its own file is fine, but I am reluctant to put this into `common/share.c`.

aingerson:
@j-xiong That makes sense, but I think if you're removing FI_HMEM from the general caps, then you should include it in the device run. Just removing it completely would break the device case

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
j-xiong: Thanks. Will do.

j-xiong:
@aingerson Thanks. That's what I was planning to do.

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
```diff
@@ -314,7 +319,10 @@ static int init_nic(int nic, char *domain_name, char *server_name, int port,
 	hints->ep_attr->rx_ctx_cnt = 1;
 	if (prov_name)
 		hints->fabric_attr->prov_name = strdup(prov_name);
-	hints->caps = FI_MSG | FI_RMA | FI_HMEM;
+	hints->caps = FI_MSG | FI_RMA;
+	if (buf_location != MALLOC)
+		hints->caps = FI_HMEM;
```
aingerson, line 324:
```
hints->caps |= FI_HMEM;
```

j-xiong: Oops! fixed now.

shefty (approved):

j-xiong:
No need to wait for CI to finish since this part of code is not exercised by CI. The AWS failure is unrelated.

PR #8281, opened 2022-11-30
prov/shm: Enable CMA Protocol with FI_HMEM and additional refactorings

zachdworkin:
The goal of these patches is to enable CMA (Cross Memory Attach) protocol with H->H transfers when FI_HMEM is enabled. The previous way of doing this was defaulting to SAR (Segmentation and Reassembly) protocol because if FI_HMEM is enabled the sender cannot be sure that the destination buffer is not a device buffer. If it was a device buffer it would error out because that memory access operation is not allowed.

Now H->H becomes possible again and a new protocol for doing H->D and D->H transfers exists. In order to create this new protocol: sar buffers, statusses, and progress_entry needed to be repurposed. They have now become ifbs (in flight buffers:  which can be used for both SAR and storing ipc information), new statusses, and a merged sar_entry and tx_entry. The sar_entry and tx_entry merged into a progress_entry because they are both used for in pending operations and it simplifies the use for this new protocol, SAR, and the DSA (Data Streaming Accelerator) paths. [The peer_id has been renamed to send_id to make it clear that it is only applicable for the sender]

The use of the IFB is necessary because if the sender sends its H buffer to a destination that is a D buffer, the reciever can utilize the IFB to save ipc information to give back to the sender so that the sender can resend the message with device operations.

The select protocol function has also been simplified because of these changes and the addition of an ipc_enabled flag.

[H=Host, D=Device, CMA=Cross Memory Attach, SAR=Segmentation and Reassembly, DSA=Data Streaming Accelerator]

### DIFF
--- a/include/ofi_shm.h
+++ b/include/ofi_shm.h
@@ -246,7 +246,6 @@ struct smr_region {
 				    Might not always be paired consistently with
 				    cmd alloc/free depending on protocol
 				    (Ex. unexpected messages, RMA requests) */
-	size_t		sar_cnt;
 
 	/* offsets from start of smr_region */
 	size_t		cmd_queue_offset;
--- a/include/ofi_shm.h
+++ b/include/ofi_shm.h
@@ -183,7 +183,7 @@ struct smr_addr {
 
 struct smr_peer_data {
 	struct smr_addr		addr;
-	uint32_t		sar_status;
+	uint32_t		status;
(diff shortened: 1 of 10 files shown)


shefty:
PR is several months old, needs updates and rebased.  Closing this PR.  Once updated and ready, please open a new PR.

PR #8286, opened 2022-11-30
prov/rxm: Always use rendezvous protocol for ZE device memory send

j-xiong:
Bounce buffer copy overhead is high for ZE device memory. The rendezvous protocol takes advantage of GPU RDMA and performs better even for small messages.

### DIFF
--- a/prov/rxm/src/rxm_msg.c
+++ b/prov/rxm/src/rxm_msg.c
@@ -802,6 +802,10 @@ rxm_send_common(struct rxm_ep *rxm_ep, struct rxm_conn *rxm_conn,
 		(data_len > rxm_ep->rxm_info->tx_attr->inject_size)) ||
 	       (data_len <= rxm_ep->rxm_info->tx_attr->inject_size));
 
+	iface = rxm_mr_desc_to_hmem_iface_dev(desc, count, &device);
+	if (iface == FI_HMEM_ZE)
+		goto rndv_send;
+
 	if (data_len <= rxm_ep->eager_limit) {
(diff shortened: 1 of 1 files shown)


prov/rxm/src/rxm_msg.c
```diff
@@ -802,17 +802,17 @@ rxm_send_common(struct rxm_ep *rxm_ep, struct rxm_conn *rxm_conn,
...
 
-	if (data_len <= rxm_ep->eager_limit) {
+	iface = rxm_mr_desc_to_hmem_iface_dev(desc, count, &device);
+
+	if (data_len <= rxm_ep->eager_limit && iface != FI_HMEM_ZE) {
 		ret = rxm_send_eager(rxm_ep, rxm_conn, iov, desc, count,
 				     context, data, flags, tag, op,
 				     data_len, total_len);
-	} else if (data_len <= rxm_ep->sar_limit) {
+	} else if (data_len <= rxm_ep->sar_limit && iface != FI_HMEM_ZE) {
```
shefty, line 811: Can we restructure the checks to avoid comparing iface against ZE multiple times?  I don't know if we want to move the rndv check to the top or use a goto to jump down.  Eager might be most common for MPI, but rndv may be for oneCCL.

j-xiong: I vote for `goto`. That introduces the least changes to the original flow.

shefty (approved):

shefty:
I will pick for 1.17

PR #8288, opened 2022-12-01
Make fabtests capable of selecting an EFA device.

vidsouza:
The --info-index option is used to specify the index of the desired EFA device present in the cluster.
A test is also introduced in this commit that tests this functionality. The test runs a fabtest using each EFA device present in the cluster and checks for the transmitted bytes before and after the execution of the fabtest.


fabtests/common/shared.c
```diff
@@ -918,12 +918,35 @@ int ft_getinfo(struct fi_info *hints, struct fi_info **info)
...
 
+	/* The following print describes which EFA device is being used 
+	 * and its attributes. 
+	 * Commented intentionally. Can be used when necessary. */
+	// fprintf(stderr, "selected provider %s dev name: %s bdf: %x:%x.%x\n",
+	//        (*info)->domain_attr->name,
+	//        (*info)->nic->device_attr->name,
+	//        (*info)->nic->bus_attr->attr.pci.bus_id,
+	//        (*info)->nic->bus_attr->attr.pci.device_id,
+	//        (*info)->nic->bus_attr->attr.pci.function_id);
```
shefty, line 948: Do not add dead code

fabtests/common/hmem_neuron.c
```diff
@@ -224,6 +224,8 @@ int ft_neuron_alloc(uint64_t device, void **buf, size_t size)
 	*buf = region->ptr;
 	dlist_insert_tail(&region->entry, &neuron_alloc_list);
 
+	fprintf(stderr, "allocated nrt region ptr: %p size: %lu logical_nc_id: %lu\n", region->ptr, size, device);
```
shefty, line 227: Keep line lengths to about 80 characters.

fabtests/common/shared.c
```diff
@@ -899,7 +899,7 @@ int ft_getinfo(struct fi_info *hints, struct fi_info **info)
 {
 	char *node, *service;
 	uint64_t flags = 0;
-	int ret;
+	int ret, efa_dev_idx;
```
shefty, line 902: This is common code and should not have a provider specific indexing scheme.  Please find an API way of selecting a device

shefty (changes requested):
This needs to be done through a generic API mechanism that's maintainable.  Also remove the Makefile.in from the patch submission.

PR #8302, opened 2022-12-02
prov/peer: Introducing peer "provider" with peer_cq to begin with

grom72:
include/ofi_peer.h file along with prov/peer directory defines space where peer-related common code is placed.

peer_cq is the first structure that has been defined. It is based on the coll_cq structure.  The coll provider uses peer_cq instead of removed coll_cq.

See [man/fi_peer.3.md](man/fi_peer.3.md)

Signed-off-by: Tomasz Gromadzki <tomasz.gromadzki@intel.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/ofiwg/libfabric/8302)


prov/peer/src/peer.h
```diff
@@ -0,0 +1,51 @@
...
+#endif
+
+#include <ofi.h>
+#include <ofi_util.h>
+#include <ofi_peer.h>
+
+struct peer_cq {
+	struct util_cq util_cq;
+	struct fid_peer_cq *peer_cq;
+};
```
shefty, line 48: We can just move this to ofi_peer.h  I don't see an advantage to splitting the definitions between 2 header files.

prov/peer/src/peer_cq.c
```diff
@@ -66,21 +66,27 @@ static struct fi_ops_cq coll_cq_ops = {
...
-	struct coll_cq *cq;
-	struct fi_peer_cq_context *peer_context = context;
+	struct peer_cq *cq;
 	int ret;
 
+	const struct util_domain *util_domain;
+	const struct fi_provider* provider;
+
+	util_domain = container_of(domain, struct util_domain, domain_fid.fid);
+	provider = util_domain->fabric->prov;
```
shefty, line 79: I would update the call to take the provider pointer as an input parameter, so we don't need to dereference the domain pointer.  That matches how other calls handle this situation (e.g. see ofi_cq_init() below).

prov/peer/src/peer_cq.c
```diff
@@ -89,17 +95,36 @@ int coll_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
...
+			size_t len, void *buf, uint64_t data, uint64_t tag,
+			fi_addr_t src)
+{
+	struct peer_cq *cq;
+
+	cq = container_of(cq_fid, struct peer_cq, util_cq.cq_fid);
+
+	return cq->peer_cq->owner_ops->write(cq->peer_cq, context,
+					  flags, len, buf, data, tag, src);
+}
```
shefty, line 130: This function could move into the header as a static inline.

shefty (commented):
Minor nits -- thanks, this looks good.

prov/peer/src/peer_cq.c
grom72: Yes, as soon as we decide to move `peer_cq` there as well and I'm still not convinced it's necessary

prov/peer/src/peer.h
grom72: My intention is to make ofi_peer.h as simple as possible and avoid, if possible, making the `peer_cq` structure public.

aingerson:
I have a similar patch that adds peer_cq support to the util_cq so that any provider that uses the util_cq can be used as a peer. This patch seems very similar. I'm fine with either solution, just bringing it up in case anyone had any thoughts on preferences?

aingerson:
(I also a similar patch that adds a util_srx definition which can either be kept as util code or go into ofi_peer.h)

shefty:
ofi_peer.h is an internal header, so I'm not worried about anything there being exposed publicly.  There are more uses of util_cq, so adding peer support there makes sense.  But that's for an actual CQ being usable by a peer.

I thought the intent behind peer_cq is that the structure is only usable as a peer.  That is, there would never be an actual queue of completions behind the structure.  If that's the case, then peer_cq really shouldn't be using the util_cq structure.

This discussion goes back to the original submission of the off_coll provider.  That used the util_xx structures as bases; even though, it doesn't needs most of the functionality that the util_xx structures provide.  I think it was simply to leverage the bind call.  We should revisit that.

prov/peer/src/peer_cq.c
grom72: Done

grom72:
@shefty do you mean something as below?
```
struct peer_cq {
	struct fid	fid;
	struct fid_peer_cq *peer_cq;
};
```
But can we name it still `..._cq`?

include/ofi_peer.h
```diff
@@ -0,0 +1,47 @@
...
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _OFI_PEER_H_
+#define _OFI_PEER_H_
+
+#include <rdma/fi_collective.h>
```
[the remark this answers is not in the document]
grom72: Done

shefty:
I mean:
```
struct ofi_peer_cq {
    struct fid_cq cq_fid;
    struct fid_peer_cq *cq;
};
```
This is for a CQ that is only ever used with a peer.  Before making that change, we need to analyze the impact of not having the util_cq structure present.  The above can be made to work, but it's likely incomplete.

There are certain util_cq fields that we probably don't need: wait, cirq, cq_lock, aux_queue, internal_wait, wakeup, src; and others that I'm unsure about: progress, read_entry, ep_list, flags.

My concern is that the util_xx classes have gotten huge because if one provider needs something, the util_xx gets modified, and now all providers end up carrying the extra code.  It makes it difficult to know if certain fields even apply to a provider.

It's definitely going to be confusing to have util_cq contain a peer_cq field, while also having a peer_cq structure contain a util_cq field (which then contains a peer_cq field).

grom72:
So far, ``peer_cq`` is used only to access the peer's CQ.
``ep.progress()``function which normally is reached via ``util_cq.read()  is called directly from `rxm_ep.progress()`` in case of ``coll_ep.progress()`` and ``offload_ep.progress()``. 
I think we can start with the following:
```
struct ofi_peer_cq {
    struct fid_cq cq_fid;
    struct fid_peer_cq *cq;
};
```
if we agree to continue whit approach, that peer ep's ``progress()`` calls other ep's ``progress()``.
Later, we can upgrade it to ``util_cq`` based version if we find some issues.

shefty:
I got notification of another comment, but can't find it here.  Maybe you edited it?

Anyway, ofi_ep_bind_cq() is a relative short, simple function.  Carrying all of util_cq just for that one call doesn't seem worth it.

We have FI_CLASS_CQ and FI_CLASS_PEER_CQ, so we could distinguish between different parent structures if we wanted ofi_ep_bind_cq() to handle both cq formats.

If we keep with a single structure, can we at least add comments that identify a bunch of fields in util_cq as not applying to FI_CLASS_PEER_CQ.  I'd say introduce a union, but that would produce rippling changes throughout the code base.

grom72:
Yes, I've changed it after a more detailed source code analysis.
There is almost no common source code for util_cq and peer_cq.
peer_cq must support only `write()` and `writeerr()` functions.
And the `ep` must be aware of how to `bind` peer_cq.
Now we have a very complicated `rxm_cq`:
```
struct rxm_cq {
	struct util_cq util_cq;
	struct fid_peer_cq peer_cq;
	struct fid_cq *util_coll_cq;
	struct fid_cq *offload_coll_cq;
};
```
with `util_coll_cq` and `offload_coll_cq` used only for creating, binding, and closing.
The rest of the operations are done via `peer_cq` embedded in a very complex way into `util_coll_cq` and `offload_coll_cq`.
We call:
`ofi_peer_cq_write(&ep->util_ep.tx_cq->cq_fid, coll_op->context, FI_COLLECTIVE, 0, 0, 0, 0, 0);`
which immediately converts `cq_fid` to `struct peer_cq` to get access to `fid_peer_cq` from `rxm_cq`.

What happens if we use `struct fid_peer_cq` directly in ep_bind?
Do we have any value in treating `peer_cq` like a regular 'cq'?
If any `ep` want to support `peer_cq` it must provide its own bind() implementation that handles `FI_CLASS_PEER_CQ` in a specific way, and the rest can be done via `ofi_ep_fid_bind()`

ldorau (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ofiwg/libfabric/8302)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @j-xiong and @shefty)

prov/rxm/src/rxm_ep.c
```diff
@@ -1623,7 +1623,7 @@ static int rxm_ep_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
 
 	case FI_CLASS_CQ:
 		rxm_cq = container_of(bfid, struct rxm_cq, util_cq.cq_fid.fid);
-		if (rxm_ep->util_coll_ep && rxm_cq->util_coll_cq) {
+		if (rxm_ep->util_coll_ep) {
```
j-xiong, line 1626: need to add FI_CLASS_PEER_CQ to the `case` clause.

j-xiong (commented):
This title has wrong provider name: `prov/proxy: Minimalist version of peer_cq`.

grom72:
> This title has wrong provider name: `prov/proxy: Minimalist version of peer_cq`.

Done

prov/rxm/src/rxm_ep.c
grom72: Do you expect that somebody externally can bind peer_cq to rxm provider?
Or do we just want to reject such a case as rxm_ep does not support peer_cq so far?
```
	case FI_CLASS_PEER_CQ:
		return -FI_ENOSYS;
		break;
```

j-xiong: My bad. This is not needed here. I was cross reading this and the util_coll additions.

shefty:
When updating patches, please update the original patch directly, rather than adding on changes in a separate patch.  For example, we only want to merge 1 patch with the correct set of changes, rather than 2 patches, with the second modifying the first.  This is needed in order to cherry pick fixes into stable branches and easily deal with conflicts.

shefty:
I have added the changes from this branch into my personal tree and pushed those into github.  I'll make the changes available to the next developer who picks up this work.  Thanks!

PR #8303, opened 2022-12-02
fabtests/multinode check event's context when read FI_JOIN_COMPLETE

grom72:
fi_join() operation may produce several completion events if more than one ep is called. This can happen especially when offload and util providers are joining at the same time. In such a case, we should wait for final completion that comes from the top collective provider.

Signed-off-by: Tomasz Gromadzki <tomasz.gromadzki@intel.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/ofiwg/libfabric/8303)

### DIFF
--- a/fabtests/multinode/src/core_coll.c
+++ b/fabtests/multinode/src/core_coll.c
@@ -76,29 +76,33 @@ static bool is_my_rank_participating()
 	return true;
 }
 
-static int wait_for_event(uint32_t event)
+static int wait_for_event(uint32_t event, const void *context)
 {
 	uint32_t ev;
 	int err;
 	struct fi_cq_err_entry comp = { 0 };
+	struct fi_eq_entry entry;
 
 	do {
-		err = fi_eq_read(eq, &ev, NULL, 0, 0);
+		err = fi_eq_read(eq, &ev, &entry, sizeof(entry), 0);
 		if (err >= 0) {
-			FT_DEBUG("found eq entry %d\n", event);
+			FT_DEBUG("found eq entry %d\n", ev);
 			if (ev == event) {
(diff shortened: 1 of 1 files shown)


shefty:
From the view of the application, it calls fi_join() once and should get back 1 event.  Any internal joins that happen must be hidden from the app.

grom72:
@shefty I have changed the logic.
Now we are waiting only for one completion event with the given context.

ldorau (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ofiwg/libfabric/8303)* status: 0 of 1 files reviewed, all discussions resolved

j-xiong:
Those events w/ the wrong context should not exist in the queue.

We probably should change peer_eq from using the default fi_eq_write() to using a write() method in `struct fi_ops_peer_eq`. That would provide a way for the eq owner to process the event before writing it to the actual event queue.

shefty:
@j-xiong - My first thought is that the FI_PEER flag can be passed to fi_eq_write() to convey the same data.  But maybe that's causing some of the confusion here.

We can extend the peer eq ops with a write call and remove the fid_eq, which would align the peer eq more with the peer cq.  That could also make it clearer that the only operation the peer is allowed to invoke is write, and that they lack access to the other eq functions.

ldorau (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ofiwg/libfabric/8303)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @grom72)

shefty (approved):
re-started CI

shefty:
Thanks!

PR #8305, opened 2022-12-02
contrib/intel/jenkins: Add local file logging and release job version control to summary

zachdworkin:
Additions to Summary.py:
    Add a Logger class to print output to stdout and a specific local file
    Add a Release class to handle printing of Makefile.am and configure.ac diffs, as well as the release number
    Simplify the main class by moving the big logic block into a function
    Small variable updates/localizations

Additions to Jenkinsfile:
    Add logic to determine if a job is suspected to be a release
    This will only push if all stages succeeded
    Add version control pushing of local log files that contain:
        release number,
        Makefile.am diff (if there is one),
        configure.ac diff (if there is one)
        Non verbose summary
    Add release arguments to build.py and summary.py calls

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -683,8 +683,24 @@ pipeline {
           sh """
             env
             (
-              [[ $RELEASE -eq 1 ]] && rel_arg="--release" || rel_arg=""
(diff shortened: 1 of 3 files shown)


ooststep (changes requested):
one minor issue, but should be addressed

ooststep (approved):

PR #8309, opened 2022-12-03
prov/efa: Make fabtests capable of selecting an EFA device

vidsouza:
Fabtests take -d <domain name> argument and execute the tests on the selected EFA domain. 
The implementation of how this argument must act is implemented incorrectly in the existing code. 
This change fixes this bug and runs the fabtests using the specified domain of the EFA device.
A test is also added to test this functionality.


wzamazon (changes requested):
The change to EFA provider is a bug fix and should be in a separate commit

prov/efa/src/efa_prov_info.c
```diff
@@ -722,3 +722,21 @@ int efa_prov_info_compare_src_addr(const char *node, uint64_t flags, const struc
...
+ * @param      info[in]        info object
+ * @param      hints[in]       hints from user's call to fi_getinfo()
+ *
+ * return      1 - If the names are different
+ *             0 - No difference, names match.
+ */
+int efa_prov_info_compare_domain_name(const struct fi_info *hints,
+                                      const struct fi_info *info)
+{
+       if (hints && hints->domain_attr->name)
```
wzamazon, line 738: hints->domain_attr can be NULL too.

fabtests/pytest/efa/test_efa_device_selection.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if cmdline_args.server_id == cmdline_args.client_id:
+        pytest.skip("EFA device selection test requires 2 nodes")
+        return
+
+    timeout = 60
+    process_timed_out = False
+
+    # This command returns a list of EFA domain names and its related info
+    command = "ssh {} fi_info -p efa".format(cmdline_args.server_id)
```
wzamazon, line 19: It is better to extract this part into a separate function `get_efa_domain_names`.

since the function uses `ssh`,  better add retry on ssh connection failure.

prov/efa/src/efa_prov_info.c
```diff
@@ -722,3 +722,21 @@ int efa_prov_info_compare_src_addr(const char *node, uint64_t flags, const struc
 	return 0;
 }
 
+/*
+ * @brief If domain name of the EFA device is specified (using -d <domain_name>),
```
wzamazon, line 726: It is not necessary to say `-d`, which is a fabtests behavior, and not related to the behavior.

Should be something like 
```
check whether hints's domain name setting matches prov_info
```

prov/efa/src/efa_prov_info.c
```diff
@@ -722,3 +722,21 @@ int efa_prov_info_compare_src_addr(const char *node, uint64_t flags, const struc
...
+ *        compare and match the desired EFA device for execution
+ *
+ * @param      info[in]        info object
+ * @param      hints[in]       hints from user's call to fi_getinfo()
+ *
+ * return      1 - If the names are different
```
wzamazon, line 732: When name does not match, we should return `-FI_ENODATA`

wzamazon (approved):

PR #8313, opened 2022-12-06
src/hmem, prov/shm: register shm with hmem ifaces, avoid duplicate hmem initialization

aingerson:


prov/util/src/util_shm.c
```diff
@@ -419,6 +425,11 @@ int smr_map_to_region(const struct fi_provider *prov, struct smr_peer *peer_buf)
 	peer = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
 	peer_buf->region = peer;
 
+	ret = ofi_hmem_host_register(peer, peer->total_size);
```
j-xiong, line 428: Is it possible to only do this when HMEM is enabled?

prov/util/src/util_shm.c
```diff
@@ -316,6 +321,7 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
 void smr_free(struct smr_region *smr)
 {
 	shm_unlink(smr_name(smr));
+	(void) ofi_hmem_host_unregister(smr);
```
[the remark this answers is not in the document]
shefty: I don't think OFI_UNUSED() works for this case.  On windows that converts to unreferenced_parameter.  Casting to (void) should be portable.

prov/util/src/util_shm.c
aingerson: It would require some extra parameters passed around but I can add that. Good call

aingerson: @j-xiong Had to add some flags to pass this around. Not the prettiest thing but it's there.

prov/util/src/util_shm.c
```diff
@@ -525,8 +543,11 @@ void smr_map_del(struct smr_map *map, int64_t id)
 	pthread_mutex_unlock(&ep_list_lock);
 
 	ofi_spin_lock(&map->lock);
-	if (!entry)
+	if (!entry) {
 		munmap(map->peers[id].region, map->peers[id].region->total_size);
+		if (map->caps & FI_HMEM)
+			(void) ofi_hmem_host_unregister(map->peers[id].region);
+	}
```
j-xiong, line 550: host unregister should be called before munmap.

aingerson: Updated

j-xiong (approved):

prov/util/src/util_shm.c
```diff
@@ -267,7 +267,7 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
 
 	(*smr)->map = map;
 	(*smr)->version = SMR_VERSION;
-	(*smr)->flags = SMR_FLAG_ATOMIC | SMR_FLAG_DEBUG;
+	(*smr)->flags = SMR_FLAG_ATOMIC | SMR_FLAG_DEBUG | attr->flags;
```
shefty, line 277: It is odd to pass in flags via the attributes, but also have the code assume certain flags should be set.

While looking into this, it's confusing that the ATOMIC and DEBUG flags are sometimes only defined.  That's normal for an ifdef type of check, but for a flag it's unusual.  It means that code such as:
```
if ((flags & DEBUG) == DEBUG)
```
returns true, even though no flag is actually set.

prov/util/src/util_shm.c
```diff
@@ -315,6 +322,8 @@ int smr_create(const struct fi_provider *prov, struct smr_map *map,
 
 void smr_free(struct smr_region *smr)
 {
+	if (smr->flags & SMR_FLAG_HMEM_ENABLED)
```
shefty, line 325: Sometimes the check is against a flag; sometimes it's against caps.  It would be simpler to always use the same check, rather than requiring both options to be in sync and mean the same thing.

prov/util/src/util_shm.c
aingerson: @shefty Agreed that is weird and shouldn't be used like that. I can rethink this.

Not sure I understand what you mean by the flags are sometimes only defined. From what I can tell, they are always defined, just always correlate to how libfabric is configured/built.
```
#ifdef HAVE_ATOMICS
#define SMR_FLAG_ATOMIC	(1 << 0)
#else
#define SMR_FLAG_ATOMIC	(0 << 0)
#endif

#if ENABLE_DEBUG
#define SMR_FLAG_DEBUG	(1 << 1)
#else
#define SMR_FLAG_DEBUG	(0 << 1)
#endif
```

shefty: SMR_FLAG_ATOMIC and SMR_FLAG_DEBUG can both be defined as 0.  0 isn't really a flag value.

aingerson:
@shefty Updated.

aingerson:
failure is unrelated rxm av xfer test failure
Any thoughts on the updates?

shefty (approved):
thanks!

shefty:
bot:aws:retest

aingerson:
bot:aws:retest

wzamazon:
@aingerson  The AWS CI has been failing consistently for this PR. CUDA related tests hangs. I think the failures are real. I will try to take a look

aingerson:
@wzamazon Thanks. I figured they could be real hence the retesting. I'm assuming an issue with cudaHostRegister. It looks like that call is only valid for devices that set cudaDevAttrHostRegisterSupported which it looks like we don't check. Could it be that? I can get my hands on a CUDA system and do some debugging if nothing jumps out to you.

aingerson:
@wzamazon I tested these patches on a CUDA system and it worked for me. Could you provide more information about the test/failure/system that it's using?

wzamazon:
@aingerson  I am on vacation, will report back on Friday.

aingerson:
@wzamazon No rush. I leave for vacation Friday :) So we can touch base after the holidays

aingerson:
@wzamazon Any information you can provide on the CUDA failure?

wzamazon:
bot:aws:retest

wzamazon:
just started a manual run. will try to get back to you soon.

wzamazon:
@aingerson I figured out the reason for the failure.

The issue was that shm called `ofi_hmem_host_register`, which call `host_register` for all hmem interfaces.

Our CI has synapse ai  library installed, so FI_HMEM_SYNAPSEAI was considered initialized. but `synpaseai_host_register` was called, which returned `-FI_ENOSYS`, and caused the failure.

I opened PR https://github.com/ofiwg/libfabric/pull/8408 to make `synapseai_host_register/unregister` to return 0, which should fix your issue.

Long term, I think we should set `hmem_ops[FI_HMEM_SYNAPSEAI].initialized` to true only when there are synapse hardware. that will came later.

aingerson:
@wzamazon Thank you!

wzamazon:
Rebase to main should make aws ci pass

shefty:
congrats!  PR is now complete.  Have a free banana, compliments of Great Place to Work (tm).

PR #8329, opened 2022-12-07
prov/efa: Call efa_mr_dereg_impl() unconditionally

darrylabbate:
This essentially reverts fcc1139, which was necessary to prevent segfaults when calling `ibv_dereg_mr()` with a `NULL` pointer. Since that commit, the call to `ibv_dereg_mr()` was replaced with a call to `efa_mr_dereg_impl()`, which does not necessarily need `efa_mr->ibv_mr`.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -303,8 +303,6 @@ void efa_mr_cache_entry_dereg(struct ofi_mr_cache *cache,
 	struct efa_mr *efa_mr = (struct efa_mr *)entry->data;
 	int ret;
 
-	if (!efa_mr->ibv_mr)
-		return;
 	ret = efa_mr_dereg_impl(efa_mr);
 	if (ret)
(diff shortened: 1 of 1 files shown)


wzamazon (approved):

darrylabbate:
On second thought, I should add a check for `efa_mr->ibv_mr` in `efa_mr_dereg_impl()` as part of this PR ([link](https://github.com/ofiwg/libfabric/blob/297c9fdc8c1cc5ea472c572c22ce6483af933e70/prov/efa/src/efa_mr.c#L444-L449)):

```c
	if (efa_mr->ibv_mr) {
		err = -ibv_dereg_mr(efa_mr->ibv_mr);
		if (err) {
			EFA_WARN(FI_LOG_MR,
				"Unable to deregister memory registration\n");
			ret = err;
		}
	}
```

@wzamazon do you agree?

wzamazon:
Yes, we should

wzamazon (approved):

PR #8333, opened 2022-12-08
contrib/intel/jenkins: Make git clone specifiy which branch

zachdworkin:
In the event where git cloning upstream/main provides newer python files than the Jenkinsfile is expecting. There will be failures when the new python expected the old Jenkinsfile to call something that it didn't. Changing to checkout the same branches python files to avoid this mismatch.


nikhilnanal:
these changes also be merged in 1.17.x in order for changes to take effect?

zachdworkin:
@nikhilnanal yes they are. I ran jenkins twice with this one. One with default --branch ${TARGET} and one with --branch v1.17.x to test it.

PR #8334, opened 2022-12-09
prov/efa: Clean up efa_hmem_info initialization

darrylabbate:
This commit:
- Centralizes the protocol variable initialization for the various HMEM interfaces
- Defines `EFA_DEF_XXX` macros for default protocol message sizes, to abstract the hardcoded default values from the core logic
- Changes the "update" nomenclature to "init", which better indicates the intention/behavior of the functions
- Initializes the protocol variables for `FI_HMEM_CUDA` when p2p is unavailable (uses `FI_HMEM_SYSTEM` defaults)
- Adds a missing check for `FI_EFA_MIN_READ_WRITE_SIZE` when initializing the `FI_HMEM_SYNAPSEAI` interface info

### DIFF
--- a/prov/efa/src/efa.h
+++ b/prov/efa/src/efa.h
@@ -122,6 +122,11 @@
 /* maximum name length for shm endpoint */
 #define EFA_SHM_NAME_MAX	   (256)
 
+#define EFA_DEFAULT_RUNT_SIZE (307200)
+#define EFA_DEFAULT_INTER_MAX_MEDIUM_MESSAGE_SIZE (65536)
+#define EFA_DEFAULT_INTER_MIN_READ_MESSAGE_SIZE (1048576)
+#define EFA_DEFAULT_INTER_MIN_READ_WRITE_SIZE (65536)
+
 extern struct fi_provider efa_prov;
 extern struct util_prov efa_util_prov;
 
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -44,43 +44,69 @@ static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_
 
 	return mtu_size - rxr_pkt_max_hdr_size();
 }
+#else
+static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_domain) {
+	return 0;
+}
 #endif
 
-/**
- * @brief Update the System hmem_info struct
- *
- * @param system_info[out]	System hmem_info struct
- * @return 0 on success
- */
-static int efa_hmem_info_update_system(struct efa_hmem_info *system_info)
-{
-	system_info->initialized = true;
-	system_info->p2p_supported = true;
-
-	/* We have not yet tested runting with system memory */
-	system_info->runt_size = 0;
-	fi_param_get_size_t(&rxr_prov, "runt_size", &system_info->runt_size);
-
-	system_info->max_medium_msg_size = 65536;
-	fi_param_get_size_t(&rxr_prov, "inter_max_medium_message_size", &system_info->max_medium_msg_size);
-
-	system_info->min_read_msg_size = 1048576;
-	fi_param_get_size_t(&rxr_prov, "inter_min_read_message_size", &system_info->min_read_msg_size);
-
-	system_info->min_read_write_size = 65536;
-	fi_param_get_size_t(&rxr_prov, "inter_min_read_write_size", &system_info->min_read_write_size);
-
+static int efa_hmem_info_init_protocol_thresholds(struct efa_hmem_info *info, enum fi_hmem_iface iface, struct efa_domain *efa_domain) {
+	switch (iface) {
+	case FI_HMEM_SYSTEM:
+		/* We have not yet tested runting with system memory */
+		info->runt_size = 0;
+		info->max_medium_msg_size = EFA_DEFAULT_INTER_MAX_MEDIUM_MESSAGE_SIZE;
+		info->min_read_msg_size = EFA_DEFAULT_INTER_MIN_READ_MESSAGE_SIZE;
+		info->min_read_write_size = EFA_DEFAULT_INTER_MIN_READ_WRITE_SIZE;
+		fi_param_get_size_t(&rxr_prov, "runt_size", &info->runt_size);
+		fi_param_get_size_t(&rxr_prov, "inter_max_medium_message_size", &info->max_medium_msg_size);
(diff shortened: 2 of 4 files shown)


prov/efa/src/efa_hmem.c
```diff
@@ -311,28 +312,28 @@ int efa_hmem_info_update_all(struct efa_domain *efa_domain)
...
+	err = efa_hmem_info_init_neuron(&hmem_info[FI_HMEM_NEURON], efa_domain);
 	if (err) {
 		ret = err;
 		EFA_WARN(FI_LOG_DOMAIN, "Failed to populate the Neuron hmem_info struct! err: %d\n",
 			 err);
 	}
 
-	err = efa_hmem_info_update_synapseai(&hmem_info[FI_HMEM_SYNAPSEAI]);
+	err = efa_hmem_info_init_synapseai(&hmem_info[FI_HMEM_SYNAPSEAI], efa_domain);
 	if (err) {
```
darrylabbate, line 337: Note: this is currently unreachable since `efa_hmem_info_init_synapseai()` only returns `0`.

prov/efa/src/efa_hmem.c
```diff
@@ -267,28 +267,29 @@ static int efa_hmem_info_update_synapseai(struct efa_hmem_info *synapseai_info)
...
 	synapseai_info->p2p_supported = true;
+	efa_hmem_info_init_msg_sizes(synapseai_info, FI_HMEM_SYNAPSEAI, efa_domain);
 
 	/*  Only the long read protocol is supported */
-	if (-FI_ENODATA != fi_param_get(&rxr_prov, "inter_max_medium_message_size", &tmp_value) ||
-		-FI_ENODATA != fi_param_get(&rxr_prov, "inter_min_read_message_size", &tmp_value) ||
-		-FI_ENODATA != fi_param_get(&rxr_prov, "runt_size", &tmp_value)) {
+	if (-FI_ENODATA != fi_param_get_size_t(&rxr_prov, "inter_max_medium_message_size", &tmp_value) ||
+		-FI_ENODATA != fi_param_get_size_t(&rxr_prov, "inter_min_read_message_size", &tmp_value) ||
+		-FI_ENODATA != fi_param_get_size_t(&rxr_prov, "inter_min_read_write_size", &tmp_value) ||
```
darrylabbate, line 275: This check was previously missing (`FI_EFA_INTER_MIN_READ_WRITE_SIZE`)

prov/efa/src/efa.h
```diff
@@ -122,6 +122,11 @@
 /* maximum name length for shm endpoint */
 #define EFA_SHM_NAME_MAX	   (256)
 
+#define EFA_DEF_RUNT_SIZE (307200)
+#define EFA_DEF_INTER_MAX_MEDIUM_MESSAGE_SIZE (65536)
+#define EFA_DEF_INTER_MIN_READ_MESSAGE_SIZE (1048576)
+#define EFA_DEF_INTER_MIN_READ_WRITE_SIZE (65536)
+
```
lrbison, line 129: should these include DEFAULT in the name, since it looks like they could be overwritten based on fi_param_get ?

lrbison (commented):
looks like a good restructure

prov/efa/src/efa.h
darrylabbate: I'm following the convention set elsewhere in `efa.h` (ex: `EFA_DEF_POOL_ALIGNMENT`, `EFA_DEF_CQ_SIZE`). I infer `DEF` to mean "default" here, but it could simply be "definition."

darrylabbate:
bot:aws:retest

prov/efa/src/efa_hmem.c
```diff
@@ -44,43 +44,69 @@ static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_
 
 	return mtu_size - rxr_pkt_max_hdr_size();
 }
+#else
+static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_domain) {
+	return 0;
+}
 #endif
 
+static int efa_hmem_info_init_msg_sizes(struct efa_hmem_info *info, enum fi_hmem_iface iface, struct efa_domain *efa_domain) {
```
shijin-aws, line 53: This function name is not intuitive.. it is actually setting the protocol threshold. I would prefer `efa_hmem_info_init_protocol_thresholds`

prov/efa/src/efa_hmem.c
```diff
@@ -91,8 +117,7 @@ static int efa_hmem_info_update_cuda(struct efa_hmem_info *cuda_info, struct efa
 	int ret;
 
 	if (!ofi_hmem_is_initialized(FI_HMEM_CUDA)) {
-		EFA_INFO(FI_LOG_DOMAIN,
-		         "FI_HMEM_CUDA is not initialized\n");
+		EFA_INFO(FI_LOG_DOMAIN, "FI_HMEM_CUDA is not initialized\n");
```
shijin-aws, line 120: check the length of the line. I remember it was split to 2 lines to not break the 80 characters limit. Similar comment for other changes.

prov/efa/src/efa_hmem.c
darrylabbate: Agree, thanks

prov/efa/src/efa_hmem.c
darrylabbate: This particular line is 79 columns wide (with tabs == 8 spaces). There are many other untouched lines in this file well over 80 columns wide.

```bash
$ sed 's/\t/        /g' prov/efa/src/efa_hmem.c | egrep '.{80,}' | wc -l
      49
```

shijin-aws (approved):

prov/efa/src/efa.h
wzamazon: I think change to `DEFAULT` is better.  we should change the old ones.

darrylabbate: I can easily `s/EFA_DEF_/EFA_DEFAULT_/g`, but it would touch multiple files (only for `EFA_DEF_CQ_SIZE` since `EFA_DEF_POOL_ALIGNMENT` is actually unused). Do you want that as part of this PR?

wzamazon: No. Do not change the old ones in this PR. Just start using `DEFAULT` for new ones in this PR.

darrylabbate: Ack, see latest: 52ee1f426a27db5a4a3276b748151bfa7ddaf048..1d3f6842f5912baa4e926e6e08b914af2d671621

wzamazon (approved):

PR #8337, opened 2022-12-09
prov/efa: Remove HMEM check from rxr_ep_ctrl()

darrylabbate:
This check had essentially no effect


sunkuamzn (approved):

wzamazon (changes requested):
Can you add some detail in the commit message regarding why the check has no effect?

wzamazon (approved):
It would be better to add that

```
the function check whether efa_domain->util_domain.info_domain_caps has FI_HMEM", but FI_HMEM is not a domain capability according to document to fi_domain
```

darrylabbate:
@wzamazon See latest commit message. Good to merge?

PR #8343, opened 2022-12-12
source/ofi_hmem: Make Copy Utils Public + Warning Cleanup + New HMEM Utility Method

a-szegel:
Added const to data parameter of cuPointerSetAttribute function pointer to
    remove compiler warning.

    Move implementation of ofi_copy_to_hmem() and ofi_copy_from_hmem() from
    source to header.

Add New HMEM Utility Method.

### DIFF
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -131,18 +131,6 @@ struct ofi_hmem_ops hmem_ops[] = {
 	},
 };
 
-static inline int ofi_copy_to_hmem(enum fi_hmem_iface iface, uint64_t device,
-				   void *dest, const void *src, size_t size)
-{
-	return hmem_ops[iface].copy_to_hmem(device, dest, src, size);
-}
-
-static inline int ofi_copy_from_hmem(enum fi_hmem_iface iface, uint64_t device,
-				     void *dest, const void *src, size_t size)
-{
-	return hmem_ops[iface].copy_from_hmem(device, dest, src, size);
-}
-
 static ssize_t ofi_copy_hmem_iov_buf(enum fi_hmem_iface hmem_iface, uint64_t device,
 				     const struct iovec *hmem_iov,
 				     size_t hmem_iov_count,
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -290,6 +290,18 @@ ssize_t ofi_copy_to_hmem_iov(enum fi_hmem_iface hmem_iface, uint64_t device,
 			     size_t hmem_iov_count, uint64_t hmem_iov_offset,
 			     const void *src, size_t size);
 
+static inline int ofi_copy_to_hmem(enum fi_hmem_iface iface, uint64_t device,
+				   void *dest, const void *src, size_t size)
+{
+	return hmem_ops[iface].copy_to_hmem(device, dest, src, size);
+}
+
+static inline int ofi_copy_from_hmem(enum fi_hmem_iface iface, uint64_t device,
+				     void *dest, const void *src, size_t size)
+{
+	return hmem_ops[iface].copy_from_hmem(device, dest, src, size);
+}
+
 int ofi_hmem_get_handle(enum fi_hmem_iface iface, void *base_addr, void **handle);
 int ofi_hmem_open_handle(enum fi_hmem_iface iface, void **handle,
 			 uint64_t device, void **mapped_addr);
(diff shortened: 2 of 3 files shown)


darrylabbate (approved):

wzamazon (approved):

a-szegel:
This PR is a pre-cursor to https://github.com/a-szegel/libfabric/pull/2.

include/rdma/fi_domain.h
```diff
@@ -128,6 +128,7 @@ enum fi_hmem_iface {
 	FI_HMEM_ZE,
 	FI_HMEM_NEURON,
 	FI_HMEM_SYNAPSEAI,
+	FI_NUM_HMEM_TYPES, /* Keep this flag last to calculate the length of hmem_ops*/
```
darrylabbate, line 131: There's an `OFI_HMEM_MAX` in [include/ofi_mr.h](https://github.com/ofiwg/libfabric/blob/55e77cda043849f47dcb9780e947b4b28fd5d4de/include/ofi_mr.h#L308), though it's hardcoded.

Your addition is better but I'd prefer the name `FI_HMEM_MAX`

lrbison, line 131: @darrylabbate uses `(sizeof efa_hmem_ifaces)` which I think is a little cleaner

src/hmem.c
```diff
@@ -359,3 +359,13 @@ size_t ofi_hmem_get_ipc_handle_size(enum fi_hmem_iface iface)
...
 }
+
+int ofi_num_hmem_interfaces_initialized(void) {
+	int i, hmem_init_count = 0;
+	for (i = 1; i < FI_NUM_HMEM_TYPES; i++) {
+		if (hmem_ops[i].initialized == true) {
+			hmem_init_count++;
+		}
+	}
+	return hmem_init_count;
```
lrbison, line 370: This would be another place to use the iterator defined in this PR:
https://github.com/ofiwg/libfabric/pull/8351/files#diff-c8ef9d9845e5822a14e1c219002ba1c60a2b44c4ab90a321cf34321882b41eb6

include/rdma/fi_domain.h
darrylabbate: That only works since `efa_hmem_ifaces` is an array of `fi_hmem_iface` (which I created for my PR). C doesn't offer any way to easily retrieve the number of values defined in some enum.

shefty: Please do not modify the application headers.  In particular, you cannot safely add anything like FI_HMEM_MAX,  It does not work.  Once an enum value is defined, regardless of what you call it, *it cannot change* without the potential for breaking an application.

src/hmem.c
darrylabbate: Also, assuming we're happy with this function, it should be renamed to indicate it's counting the number of **non-standard** `FI_HMEM` interfaces, since the loop starts at 1.

include/rdma/fi_domain.h
a-szegel: Thats a good point.  I couldn't do sizeof(hmem_ops) b/c I initially had this loop in a different file.  I think I can do this now and it will work.

darrylabbate: what you would want is this:

```c
#define FI_HMEM_IFACE_FOREACH(var) \
	for ((var) = 0; (var) < ((sizeof hmem_ops) / (sizeof (struct ofi_hmem_ops))); ++(var))

#define FF_HMEM_IFACE_FOREACH_NON_SYSTEM(var) \
	for ((var) = 1; (var) < ((sizeof hmem_ops) / (sizeof (struct ofi_hmem_ops))); ++(var))
```

src/hmem.c
a-szegel: fixed.

include/rdma/fi_domain.h
a-szegel: The macro would need to be defined in the .h, and it wouldn't work because `hmem_ops` is defined as `struct ofi_hmem_ops hmemops [] = {x, y, z}` in the .c, which leads to a compile issue of not being able to known the size of the array at compile time.

a-szegel: I mean FI_HMEM_MAX to be used as an internal sentinel (not used by applications)... in which case I don't think we would break them, but the current impl of not touching the .h is much cleaner.

lrbison (approved):

wzamazon (approved):

shefty (approved):

PR #8350, opened 2022-12-13
include/ofi_mr.h: remove general MR modes from OFI_MR_MODE_RMA_TARGET

wzamazon:
ofi_cap_mr_mode() will remove OFI_MR_MODE_RMA_TARGET from mr_mode, if application did not ask for RMA target related capability.

However, OFI_MR_MODE_RMA_TARGET contains FI_MR_PROV_KEY and FI_MR_VIRT_ADDR, which are relevent whenever application registers memory, not just RMA.

For example, an application might want to use FI_MR_LOCAL mr mode to manage memory registration by itself. For that, the application need to know whether the provider support FI_MR_PROV_KEY, and act accordingly.

However, currently the FI_MR_PROV_KEY mode will be removed by ofi_cap_mr_mode() because OFI_MR_MODE_RMA_TARGET contains FI_MR_PROV_KEY.

To address the issue, this patch removed FI_MR_PROV_KEY and FI_MR_VIRT_ADDR from OFI_MR_MODE_RMA_TARGET

### DIFF
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -85,8 +85,7 @@ check_local_mr:
 	return (info->mode & FI_LOCAL_MR) ? 1 : 0;
 }
 
-#define OFI_MR_MODE_RMA_TARGET (FI_MR_RAW | FI_MR_VIRT_ADDR |			\
-				 FI_MR_PROV_KEY | FI_MR_RMA_EVENT)
+#define OFI_MR_MODE_RMA_TARGET (FI_MR_RAW | FI_MR_RMA_EVENT)
 
 /* If the app sets FI_MR_LOCAL, we ignore FI_LOCAL_MR.  So, if the
  * app doesn't set FI_MR_LOCAL, we need to check for FI_LOCAL_MR.


shefty:
I don't think the description is capturing the problem.

FI_MR_PROV_KEY - This controls whether the application or the provider sets the return value of a memory registration key.  A key is only needed and used if the memory will be the target of an RMA operation.  For memory that is only used locally, there isn't a key associated with the region.  The user passes in a memory descriptor instead, which is technically independent from a key.

FI_MR_VIRT_ADDR - This too only has meaning in the context when the region is the target of an RMA operation.  It indicates if the initiator references an offset into the region based on the value 0 or the starting virtual address of the memory region.

The existing code and behavior looks correct to me.  Note that FI_MR_RAW is similar here, in that it only applies to regions that are the target of an RMA operation, which is why it is included in this set.

wzamazon:
> For memory that is only used locally, there isn't a key associated with the region. The user passes in a memory descriptor instead, which is technically independent from a key.

PR #8351, opened 2022-12-14
prov/efa: Add `p2p_required_by_impl` to `efa_hmem_info` struct, implement generic p2p opt initialization/validation

darrylabbate:
This adds a `p2p_required_by_impl` field to the `efa_hmem_info` struct as a way to track the minimum p2p requirements for each supported `FI_HMEM` interface. Since the p2p requirements of each interface is more-or-less unwritten/institutional knowledge, it should be defined in a single place and queried as needed, instead of hardcoding in various places throughout the codebase or deducing from a combination of other parameters.

The `p2p_required_by_impl` field also offers some dynamic configurability if we choose to utilize it. I.e. if we want to set `p2p_required_by_impl` based on some runtime condition.

As an initial use-case, this PR also adds a function for checking the validity of an endpoint p2p opt against a given HMEM interface in a more generic way (i.e. without the need to hardcode). The benefit is if/when the p2p requirements change for any given HMEM interface, the p2p opt validation will work as intended.

### DIFF
--- a/prov/efa/src/efa_ep.c
+++ b/prov/efa/src/efa_ep.c
@@ -627,7 +627,7 @@ int efa_ep_open(struct fid_domain *domain_fid, struct fi_info *user_info,
 	struct efa_domain *domain;
 	const struct fi_info *prov_info;
 	struct efa_ep *ep;
-	int ret;
+	int ret, i;
 
 	domain = container_of(domain_fid, struct efa_domain,
 			      util_domain.domain_fid);
@@ -699,22 +699,22 @@ int efa_ep_open(struct fid_domain *domain_fid, struct fi_info *user_info,
 		memcpy(ep->src_addr, user_info->src_addr, user_info->src_addrlen);
 	}
 
-	if (ep->domain->hmem_info[FI_HMEM_CUDA].initialized) {
-		/*
-		 * Set the default to required. NCCL plugin requires p2p, but
-		 * does not call setopt for this option in older NCCL plugin
-		 * versions.
-		 */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-	} else if (ep->domain->hmem_info[FI_HMEM_NEURON].initialized) {
-		/* Neuron requires p2p and supports no other modes. */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-	} else if (ep->domain->hmem_info[FI_HMEM_SYNAPSEAI].initialized) {
-		/* SynapseAI requires p2p and supports no other modes. */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-	} else {
-		/* no hmem devices, disable p2p */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_DISABLED;
+	/* Set p2p opt to disabled by default */
+	ep->hmem_p2p_opt = FI_HMEM_P2P_DISABLED;
+
+	/*
+	 * TODO this assumes only one non-stantard interface is initialized at a
+	 * time. Refactor to handle multiple initialized interfaces to impose
+	 * tighter requirements for the default p2p opt
+	 */
+	EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(i) {
+		if (ep->domain->hmem_info[efa_hmem_ifaces[i]].initialized &&
+			ep->domain->hmem_info[efa_hmem_ifaces[i]].p2p_supported_by_device) {
+			ep->hmem_p2p_opt = ep->domain->hmem_info[efa_hmem_ifaces[i]].p2p_required_by_impl
+				? FI_HMEM_P2P_REQUIRED
+				: FI_HMEM_P2P_PREFERRED;
+			break;
+		}
 	}
 
 	*ep_fid = &ep->util_ep.ep_fid;
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -94,7 +94,8 @@ static int efa_hmem_info_init_protocol_thresholds(struct efa_hmem_info *info, en
 static int efa_hmem_info_init_system(struct efa_hmem_info *system_info, struct efa_domain *efa_domain)
 {
 	system_info->initialized = true;
-	system_info->p2p_required_by_impl = true;
+	system_info->p2p_disabled_by_user = false;
+	system_info->p2p_required_by_impl = false;
 	system_info->p2p_supported_by_device = true;
 	efa_hmem_info_init_protocol_thresholds(system_info, FI_HMEM_SYSTEM, efa_domain);
 	return 0;
@@ -135,11 +136,11 @@ static int efa_hmem_info_init_cuda(struct efa_hmem_info *cuda_info, struct efa_d
 		return -FI_ENOMEM;
 	}
 
-	/*
-	 * TODO set to false when non-p2p support is added, or set conditionally
-	 * based on some environment variable
-	 */
+	cuda_info->p2p_disabled_by_user = false;
+
+	/* TODO set to false when non-p2p support is added */
 	cuda_info->p2p_required_by_impl = true;
+
 	ibv_mr = ibv_reg_mr(g_device_list[0].ibv_pd, ptr, len, ibv_access);
 	if (!ibv_mr) {
 		cuda_info->p2p_supported_by_device = false;
@@ -215,6 +216,7 @@ static int efa_hmem_info_init_neuron(struct efa_hmem_info *neuron_info, struct e
 	}
 
 	neuron_info->initialized = true;
+	neuron_info->p2p_disabled_by_user = false;
 	/* Neuron currently requires P2P */
 	neuron_info->p2p_required_by_impl = true;
 
@@ -274,6 +276,7 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
 	}
 
 	synapseai_info->initialized = true;
+	synapseai_info->p2p_disabled_by_user = false;
 	/* SynapseAI currently requires P2P */
 	synapseai_info->p2p_required_by_impl = true;
 	synapseai_info->p2p_supported_by_device = true;
@@ -296,6 +299,56 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
 
 #endif
 	return 0;
+}
+
+/**
+ * @brief   Validate an FI_OPT_FI_HMEM_P2P (FI_OPT_ENDPOINT) option for a
+ *          specified HMEM interface. If p2p_opt is FI_HMEM_P2P_DISABLED, we
+ *          update hmem_info[iface]->p2p_disabled_by_user to be true only if the
+ *          interface doesn't require p2p.
+ *
+ * @param   domain  The efa_domain struct which contains an efa_hmem_info array
+ * @param   iface   The fi_hmem_iface enum of the HMEM interface to validate
+ * @param   p2p_opt The P2P option to validate
+ *
+ * @return  FI_SUCCESS if the P2P option is valid for the given interface
+ *         -FI_OPNOTSUPP if the P2P option is invalid
+ *         -FI_ENODATA if the given HMEM interface was not initialized
+ *         -FI_EINVAL if p2p_opt is not a valid FI_OPT_FI_HMEM_P2P option
+ */
+int efa_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem_iface iface, int p2p_opt)
+{
+	struct efa_hmem_info *info = &efa_domain->hmem_info[iface];
+
+	if (OFI_UNLIKELY(!info->initialized))
+		return -FI_ENODATA;
+	
+	switch (p2p_opt) {
+
+	/* REQUIRED is valid even if an interface doesn't require P2P */
+	case FI_HMEM_P2P_REQUIRED:
+		return info->p2p_supported_by_device ? FI_SUCCESS : -FI_EOPNOTSUPP;
+
+	/*
+	 * Invalidate ENABLED or PREFERRED if an iface requires P2P, or if an iface
+	 * doesn't support P2P at all
+	 */
+	case FI_HMEM_P2P_PREFERRED:
+	case FI_HMEM_P2P_ENABLED:
+		return info->p2p_supported_by_device && !info->p2p_required_by_impl
+			? FI_SUCCESS
+			: -FI_EOPNOTSUPP;
+
+	/* DISABLED is valid unless an interface requires P2P */
+	case FI_HMEM_P2P_DISABLED:
+		if (info->p2p_required_by_impl) {
+			return -FI_EOPNOTSUPP;
+		} else {
+			info->p2p_disabled_by_user = true;
+			return FI_SUCCESS;
+		}
+	}
+	return -FI_EINVAL;
 }
 
 /**
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -94,7 +94,8 @@ static int efa_hmem_info_init_protocol_thresholds(struct efa_hmem_info *info, en
 static int efa_hmem_info_init_system(struct efa_hmem_info *system_info, struct efa_domain *efa_domain)
 {
 	system_info->initialized = true;
-	system_info->p2p_supported = true;
+	system_info->p2p_required_by_impl = true;
+	system_info->p2p_supported_by_device = true;
 	efa_hmem_info_init_protocol_thresholds(system_info, FI_HMEM_SYSTEM, efa_domain);
 	return 0;
 }
@@ -134,9 +135,14 @@ static int efa_hmem_info_init_cuda(struct efa_hmem_info *cuda_info, struct efa_d
 		return -FI_ENOMEM;
 	}
 
+	/*
+	 * TODO set to false when non-p2p support is added, or set conditionally
+	 * based on some environment variable
+	 */
+	cuda_info->p2p_required_by_impl = true;
 	ibv_mr = ibv_reg_mr(g_device_list[0].ibv_pd, ptr, len, ibv_access);
 	if (!ibv_mr) {
-		cuda_info->p2p_supported = false;
+		cuda_info->p2p_supported_by_device = false;
 		/* Use FI_HMEM_SYSTEM message sizes when p2p is unavailable */
 		efa_hmem_info_init_protocol_thresholds(cuda_info, FI_HMEM_SYSTEM, efa_domain);
 		EFA_WARN(FI_LOG_DOMAIN,
@@ -154,7 +160,7 @@ static int efa_hmem_info_init_cuda(struct efa_hmem_info *cuda_info, struct efa_d
 		return ret;
 	}
 
-	cuda_info->p2p_supported = true;
+	cuda_info->p2p_supported_by_device = true;
 	efa_hmem_info_init_protocol_thresholds(cuda_info, FI_HMEM_CUDA, efa_domain);
 	if (-FI_ENODATA != fi_param_get(&rxr_prov, "inter_max_medium_message_size", &tmp_value)) {
 		EFA_WARN(FI_LOG_DOMAIN,
@@ -209,10 +215,12 @@ static int efa_hmem_info_init_neuron(struct efa_hmem_info *neuron_info, struct e
 	}
 
 	neuron_info->initialized = true;
+	/* Neuron currently requires P2P */
+	neuron_info->p2p_required_by_impl = true;
 
 	ibv_mr = ibv_reg_mr(g_device_list[0].ibv_pd, ptr, len, ibv_access);
 	if (!ibv_mr) {
-		neuron_info->p2p_supported = false;
+		neuron_info->p2p_supported_by_device = false;
 		/* We do not expect to support Neuron on non p2p systems */
 		EFA_WARN(FI_LOG_DOMAIN,
 		         "Failed to register Neuron buffer with the EFA device, "
@@ -230,7 +238,7 @@ static int efa_hmem_info_init_neuron(struct efa_hmem_info *neuron_info, struct e
 		return ret;
 	}
 
-	neuron_info->p2p_supported = true;
+	neuron_info->p2p_supported_by_device = true;
 	efa_hmem_info_init_protocol_thresholds(neuron_info, FI_HMEM_NEURON, efa_domain);
 	if (-FI_ENODATA != fi_param_get(&rxr_prov, "inter_max_medium_message_size", &tmp_value)) {
 		EFA_WARN(FI_LOG_DOMAIN,
@@ -266,7 +274,9 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
 	}
 
 	synapseai_info->initialized = true;
-	synapseai_info->p2p_supported = true;
+	/* SynapseAI currently requires P2P */
+	synapseai_info->p2p_required_by_impl = true;
+	synapseai_info->p2p_supported_by_device = true;
 	efa_hmem_info_init_protocol_thresholds(synapseai_info, FI_HMEM_SYNAPSEAI, efa_domain);
 
 	/*  Only the long read protocol is supported */
--- a/prov/efa/src/efa_hmem.h
+++ b/prov/efa/src/efa_hmem.h
@@ -34,8 +34,23 @@
 #ifndef EFA_HMEM_H
 #define EFA_HMEM_H
 
+#define EFA_HMEM_IFACE_FOREACH(var) \
+	for ((var) = 0; (var) < ((sizeof efa_hmem_ifaces) / (sizeof (enum fi_hmem_iface))); ++(var))
+
+#define EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(var) \
+	for ((var) = 1; (var) < ((sizeof efa_hmem_ifaces) / (sizeof (enum fi_hmem_iface))); ++(var))
+
+/* Order matters */
+static const enum fi_hmem_iface efa_hmem_ifaces[] = {
+	FI_HMEM_SYSTEM,	/* Must be first here */
+	FI_HMEM_CUDA,
+	FI_HMEM_NEURON,
+	FI_HMEM_SYNAPSEAI
+};
+
 struct efa_hmem_info {
 	bool initialized; 	/* do we support it at all */
+	bool p2p_disabled_by_user;	/* Did the user disable p2p via FI_OPT_FI_HMEM_P2P? */
 	bool p2p_required_by_impl;	/* Is p2p required for this interface? */
 	bool p2p_supported_by_device;	/* do we support p2p with this device */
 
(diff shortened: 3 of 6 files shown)


prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1120,44 +1120,28 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
...
+		FI_HMEM_NEURON,
+		FI_HMEM_SYNAPSEAI
+	};
+	
 	/*
-	 * TODO: support the other options. We can only support ENABLED
-	 * and PREFERRED when p2p is available. DISABLED is not
-	 * supported yet.
+	 * Check the opt's validity against the first initialized FI_HMEM interface
+	 * encountered from the list above
```
lrbison, line 1134: Checking against the first is somewhat strange.  Should it be all or any?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1120,44 +1120,28 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
...
-				return -FI_EOPNOTSUPP;
-			efa_ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-		} else {
-			return -FI_EOPNOTSUPP;
+	for (i = 0; i < ((sizeof ifaces) / (sizeof (enum fi_hmem_iface))); ++i) {
+		ret = efa_hmem_validate_p2p_opt(efa_ep->domain, ifaces[i], opt);
+		if (ret != FI_ENODATA) {
+			if (!ret)
+				efa_ep->hmem_p2p_opt = opt;
+			return -ret;
```
lrbison, line 1141: So we return the negative of the either FI_SUCCESS  or FI_EOPNOTSUPP?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1120,44 +1120,28 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
...
-		break;
-	case FI_HMEM_P2P_DISABLED:
-		return -FI_EOPNOTSUPP;
-		break;
-	default:
-		return -FI_EINVAL;
 	}
-
-	return 0;
+	return -FI_EINVAL;
```
lrbison, line 1144: If no hmem interface is enabled, we return -FI_EINVAL, but perhaps that would be considered a success as well?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1120,44 +1120,28 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
  */
 static int efa_set_fi_hmem_p2p_opt(struct efa_ep *efa_ep, int opt)
 {
```
lrbison, line 1122: I think this function need to document what it is doing, and what the possible return values are.

prov/efa/src/efa_hmem.c
```diff
@@ -298,6 +298,49 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
...
+	case FI_HMEM_P2P_PREFERRED:
+	case FI_HMEM_P2P_ENABLED:
+		return info.p2p_supported && !info.p2p_required
+			? FI_SUCCESS
+			: FI_EOPNOTSUPP;
+
+	/* DISABLED is valid unless an interface requires P2P */
+	case FI_HMEM_P2P_DISABLED:
+		return !info.p2p_required ? FI_SUCCESS : FI_EOPNOTSUPP;
+	}
```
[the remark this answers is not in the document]
wzamazon: I think the flag name `p2p_required` is a bit confusing.

The flag introduced here meant to say whether p2p is required by the implementation (which user cannot change).

The confusing part is that there is setopt flag p2p_required, which means that even if provider can use non-p2p to support the memory type, the user does not want provider to do that.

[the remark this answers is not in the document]
darrylabbate: @lrbison `FI_HMEM_P2P_DISABLED` should be supported when non-p2p `FI_HMEM_CUDA` modes are supported; the switch case here will be necessary then.

@wzamazon I can see how the nomenclature of `p2p_required` might be confusing since it might indicate a user-defined restriction for libfabric. Perhaps something like `p2p_only`?

prov/efa/src/rxr/rxr_ep.c
darrylabbate: Compare against the previous logic; this is essentially what was happening before. We would check if `FI_HMEM_CUDA` was initialized (and p2p supported) and set the p2p opt if valid. Otherwise, we would check if `FI_HMEM_NEURON` was intialized and assert the option was `FI_HMEM_P2P_REQUIRED`.

Checking against all is a no-go, since that would restrict valid options to the most restrictive option (`FI_HMEM_P2P_DISABLED`). Checking against any might be a better idea, but what I've implemented is essentially a transposed generic version of what we had before.

prov/efa/src/rxr/rxr_ep.c
darrylabbate: Is the [existing comment](https://github.com/ofiwg/libfabric/blob/55e77cda043849f47dcb9780e947b4b28fd5d4de/prov/efa/src/rxr/rxr_ep.c#L1114-L1120) not sufficient?

```c
/*
 * Validate p2p opt passed by the user and set the endpoint option if it is
 * valid. If the option chosen is invalid or not supported, return an error.
 *
 * @param[in]	efa_ep	efa endpoint
 * @return 	0 on success, negative errno on error
 */
```

prov/efa/src/rxr/rxr_ep.c
darrylabbate: Shouldn't a `FI_OPT_FI_HMEM_P2P` option technically be invalid if there are no available `FI_HMEM` interfaces?

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1120,44 +1120,28 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
...
 
-	hmem_info = efa_ep->domain->hmem_info;
-
-	switch (opt) {
+	/* FI_HMEM interfaces to check against */
+	static enum fi_hmem_iface ifaces[] = {
+		FI_HMEM_CUDA,
+		FI_HMEM_NEURON,
+		FI_HMEM_SYNAPSEAI
+	};
```
darrylabbate, line 1130: We did not validate against `FI_HMEM_SYNAPSEAI` previously. I assume we should do so.

Also not sure about `FI_HMEM_SYSTEM`.

prov/efa/src/rxr/rxr_ep.c
darrylabbate: See another example in our provider [here](https://github.com/ofiwg/libfabric/blob/55e77cda043849f47dcb9780e947b4b28fd5d4de/prov/efa/src/efa_ep.c#L702-L718):

```c
	if (ep->domain->hmem_info[FI_HMEM_CUDA].initialized) {
		/*
		 * Set the default to required. NCCL plugin requires p2p, but
		 * does not call setopt for this option in older NCCL plugin
		 * versions.
		 */
		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
	} else if (ep->domain->hmem_info[FI_HMEM_NEURON].initialized) {
		/* Neuron requires p2p and supports no other modes. */
		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
	} else if (ep->domain->hmem_info[FI_HMEM_SYNAPSEAI].initialized) {
		/* SynapseAI requires p2p and supports no other modes. */
		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
	} else {
		/* no hmem devices, disable p2p */
		ep->hmem_p2p_opt = FI_HMEM_P2P_DISABLED;
	}
```

As you can see, we just iterate through an arbitrary set of `FI_HMEM` interfaces and short-circuit upon the first initialized interface.

prov/efa/src/efa_hmem.h
```diff
@@ -34,8 +34,23 @@
 #ifndef EFA_HMEM_H
 #define EFA_HMEM_H
 
+#define EFA_HMEM_IFACE_FOREACH(var) \
+	for ((var) = 0; (var) < ((sizeof efa_hmem_ifaces) / (sizeof (enum fi_hmem_iface))); ++(var))
+
+#define EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(var) \
+	for ((var) = 1; (var) < ((sizeof efa_hmem_ifaces) / (sizeof (enum fi_hmem_iface))); ++(var))
```
darrylabbate, line 41: I will use these macros elsewhere in upcoming PRs as well.

prov/efa/src/efa_hmem.c
```diff
@@ -94,7 +94,9 @@ static int efa_hmem_info_init_protocol_thresholds(struct efa_hmem_info *info, en
 static int efa_hmem_info_init_system(struct efa_hmem_info *system_info, struct efa_domain *efa_domain)
 {
 	system_info->initialized = true;
+	system_info->p2p_required = false;
```
darrylabbate, line 97: Changed this to `false` in the latest commit based on the fact `FI_HMEM_P2P_DISABLED` was the default `ep->hmem_p2p_opt` value when no other `FI_HMEM` interface was initialized.

lrbison (approved):
Thanks for the updates, looks good

prov/efa/src/efa_hmem.h
```diff
@@ -36,6 +36,7 @@
 
 struct efa_hmem_info {
 	bool initialized; 	/* do we support it at all */
+	bool p2p_required;	/* Does the iface require p2p? */
```
wzamazon, line 53: The name will confused with FI_HMEM_P2P_REQUIRED,  which is a user option.

This flag should should be `p2p_required_by_implementation`

prov/efa/src/efa_hmem.c
```diff
@@ -298,6 +298,49 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
...
+ *         specified HMEM interface
+ *
+ * @param  domain  The efa_domain struct which contains an efa_hmem_info array
+ * @param  iface   The fi_hmem_iface enum of the HMEM interface to validate
+ * @param  p2p_opt The P2P option to validate
+ *
+ * @return FI_SUCCESS if the P2P option is valid for the given interface
+ *         FI_OPNOTSUPP if the P2P option is invalid
+ *         FI_ENODATA if the given HMEM interface was not initialized
+ *         FI_EINVAL if p2p_opt is not a valid FI_OPT_FI_HMEM_P2P option
```
wzamazon, line 316: return value should be negative for an error

prov/efa/src/efa_hmem.c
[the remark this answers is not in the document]
wzamazon: I think a different name is needed. But neither `p2p_only` or `p2p_mandatory` sounds right. should be something more explicit like `p2p_required_by_implementation`.

prov/efa/src/efa_ep.c
```diff
@@ -699,24 +699,16 @@ int efa_ep_open(struct fid_domain *domain_fid, struct fi_info *user_info,
...
-		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-	} else if (ep->domain->hmem_info[FI_HMEM_SYNAPSEAI].initialized) {
-		/* SynapseAI requires p2p and supports no other modes. */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-	} else {
-		/* no hmem devices, disable p2p */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_DISABLED;
+	EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(i) {
+		if (ep->domain->hmem_info[efa_hmem_ifaces[i]].initialized) {
+			ep->hmem_p2p_opt = ep->domain->hmem_info[efa_hmem_ifaces[i]].default_p2p_opt;
```
wzamazon, line 704: I am not sure about this. What if there are 2 interfaces initialized? Today initialized just means libfabric is compiled with the support, and library has been loaded.

wzamazon: It is possible for an instance to have both Neuron library and CUDA library installed. With this, neuron setting end up always overwrite CUDA setting.

wzamazon (changes requested):

prov/efa/src/efa_ep.c
darrylabbate: > I am not sure about this. What if there are 2 interfaces initialized? Today initialized just means libfabric is compiled with the support, and library has been loaded.

You and I spoke about this yesterday. I thought we agreed this was all we could really do given the way the `FI_HMEM` interfaces are modeled.

Also, this is no different than the previous logic. The previous logig was simply an unrolled version of this for loop.

> With this, neuron setting end up always overwrite CUDA setting.

You have it backwards. `FI_HMEM_CUDA` would be checked first (see enum definition) and break out of the loop before `FI_HMEM_NEURON` is checked at all.

prov/efa/src/efa_hmem.h
darrylabbate: Ack, will change

prov/efa/src/efa_hmem.c
darrylabbate: Ack

prov/efa/src/rxr/rxr_ep.c
lrbison: Thank you for expanding the comment here to specify which error messages mean what.

prov/efa/src/rxr/rxr_ep.c
lrbison: I see.  Related to do we assume we have multiple initialized issue.

prov/efa/src/efa_ep.c
darrylabbate: > Also, this is no different than the previous logic. The previous logic was simply an unrolled version of this for loop.

To demonstrate my point in pseudocode:

```c
    // Old
    if CUDA.initialized:
        p2p_opt = CUDA.default_p2p_opt
    else if NEURON.initialized:
        p2p_opt = NEURON.default_p2p_opt
    else if SYNAPSEAI.initialized:
        p2p_opt = SYNAPSEAI.default_p2p_opt
    else:
        p2p_opt = <default_p2p_opt>
```
is the same as
```c
    // New
    for iface in [CUDA, NEURON, SYNAPSEAI]:
        if iface.initialized:
            p2p_opt = iface.default_p2p_opt
            goto p2p_opt_set
    p2p_opt = <default_p2p_opt>

p2p_opt_set:
    ...
```

So while I would agree the default p2p option logic is flawed, it's simply a generic translation of what we currently have. So we can brainstorm how to approach this issue, but I don't think this should be a blocker for the PR as a whole since it virtually changes nothing wrt current behavior.

wzamazon (commented):
Overall I think this PR is in the right direction.

A few things need to change:

`limit_to_p2p` is even worse than `p2p_required` . I would use `p2p_required_by_impl`.

In the same spirit, `p2p_supported` should be renamed to `p2p_supported_by_device`. 

Regarding `default_hmem_opt`, we do not need to keep record of all the user options.

Among the 4 hmem user options, we only need to keep record of user want to disable. hence we need other flag `p2p_disabled_by_user`.

This is because our default behavior is user p2p when it is supported.

User setting  FI_HMEM_OPT to ENABLED and PREFERRED should be no-op for us, because they match the default behavior.

User setting FI_HMEM_OPT to REQUIRED will either fail with -FI_EOPNOTSUPP, when p2p is not supported by device, or no-op because default behavior is to use p2p when supported.

User set p2p to FI_HMEM_OPT_ENABLED or PREFERED should be a no-op for us, because that is our default behavior.

Therefore, the only user option we need to keep track on  is whether user want to disable p2p, so I would add `p2p_disabled_by_user`.

In summary, we add two flags `p2p_required_by_impl`, `p2p_disabled_by_user`. rename `p2p_supported` to `p2p_support_by_device`.

We should also remove `hmem_p2p_opt` from `efa_ep`, and use the information we saved in hmem_info[iface] instead.

prov/efa/src/efa_ep.c
wzamazon: We should refactor how `efa_ep->hmem_opt` is evaluated, but we do not need `hmem_info->default_hmem_opt`.

it can be derived from existing information in `hmem_info`

```
int default_hmem_opt()
{
ninited = 0;
EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(i) {
  if (ep->domain->hmem_info[efa_hmem_ifaces[i]].initialized)  {
      if (ep->domain->hmem_info[efa_hmem_ifaces[i]].p2p_required_by_impl)
              return p2p_required;

    ninited +=1;
  }
  if (ninited == 0)
       return p2p_disabled;

  return p2p_prefered
}

int efa_ep_open()
{
....
efa_ep->hmem_opt = default_hmem_opt();
}
```

darrylabbate:
Summarizing the current status based on offline conversations:

- `efa_ep->hmem_p2p_opt` can't be removed at the moment. While we can easily deduce the p2p operations based on other parameters and our own provider's preferences, we still need to return the correct `FI_OPT_FI_HMEM_P2P` option when the user calls `fi_getopt()`; we can't disambiguate `FI_HMEM_P2P_PREFERRED` and `FI_HMEM_P2P_ENABLED`.
- I've removed the `default_p2p_opt` field from the `efa_hmem_info` struct, instead deducing the default p2p option in `efa_ep_open()`.
- Struct fields renamed per suggestions from @wzamazon

prov/efa/src/efa_hmem.c
```diff
@@ -135,11 +136,12 @@ static int efa_hmem_info_init_cuda(struct efa_hmem_info *cuda_info, struct efa_d
 		return -FI_ENOMEM;
 	}
 
-	/*
-	 * TODO set to false when non-p2p support is added, or set conditionally
-	 * based on some environment variable
-	 */
+	/* TODO query FI_HMEM_CUDA_ENABLE_XFER and set conditionally */
```
wzamazon, line 139: Why would it depend on FI_HMEM_CUDA_ENABLE_XFER? Maybe you meant `FI_HMEM_P2P_DISABLED`?

prov/efa/src/efa_hmem.c
```diff
@@ -298,6 +302,54 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
...
+ * @param   domain  The efa_domain struct which contains an efa_hmem_info array
+ * @param   iface   The fi_hmem_iface enum of the HMEM interface to validate
+ * @param   p2p_opt The P2P option to validate
+ *
+ * @return  FI_SUCCESS if the P2P option is valid for the given interface
+ *         -FI_OPNOTSUPP if the P2P option is invalid
+ *         -FI_ENODATA if the given HMEM interface was not initialized
+ *         -FI_EINVAL if p2p_opt is not a valid FI_OPT_FI_HMEM_P2P option
+ */
+int efa_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem_iface iface, int p2p_opt)
```
wzamazon, line 318: should be `set_p2p_opt` because it does not only validate but also set `disabled_by_user`.

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1065,44 +1065,21 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
  */
 static int efa_set_fi_hmem_p2p_opt(struct efa_ep *efa_ep, int opt)
 {
-	struct efa_hmem_info *hmem_info;
+	int i, ret;
```
wzamazon, line 1068: `ret` -> `err`

prov/efa/src/rxr/rxr_ep.c
```diff
@@ -1065,44 +1065,21 @@ static ssize_t rxr_ep_cancel(fid_t fid_ep, void *context)
...
-			 * Neuron requires p2p support and supports no
-			 * other modes.
-			 */
-			if (opt != FI_HMEM_P2P_REQUIRED)
-				return -FI_EOPNOTSUPP;
-			efa_ep->hmem_p2p_opt = FI_HMEM_P2P_REQUIRED;
-		} else {
-			return -FI_EOPNOTSUPP;
+	EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(i) {
+		ret = efa_hmem_validate_p2p_opt(efa_ep->domain, efa_hmem_ifaces[i], opt);
```
wzamazon, line 1075: the logic is a bit hard to understand. Also `efa_ep->hmem_p2p_opt` should be updated after we go through all interfaces.

How about

```
EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(i) {
err = efa_hmem_validate_p2p_opt(efa_ep->domain, efa_hmem_ifaces[i], opt);
if (err == -FI_ENODATA)
    continue

if (err)
   return err;
}

efa_ep->hmem_p2p_opt  = opt;
```

prov/efa/src/efa_ep.c
```diff
@@ -699,22 +699,17 @@ int efa_ep_open(struct fid_domain *domain_fid, struct fi_info *user_info,
...
-	} else {
-		/* no hmem devices, disable p2p */
-		ep->hmem_p2p_opt = FI_HMEM_P2P_DISABLED;
+	/* Set p2p opt to disabled by default */
+	ep->hmem_p2p_opt = FI_HMEM_P2P_DISABLED;
+
+	EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(i) {
+		if (ep->domain->hmem_info[efa_hmem_ifaces[i]].initialized &&
+			ep->domain->hmem_info[efa_hmem_ifaces[i]].p2p_supported_by_device) {
+			ep->hmem_p2p_opt = ep->domain->hmem_info[efa_hmem_ifaces[i]].p2p_required_by_impl
```
wzamazon, line 708: Not sure. If there are multiple interfaces, any one's  p2p_require_by_impl is true should make hmem_p2p_opt to P2P_REQUIRED.
Better wrap this up in a utility function

wzamazon (changes requested):
The commit message of 1st commit is not entirely actuate. It not only introduced `p2p_requred_by_impl`, but also renamed `p2p_supprted` to `p2p_supported_by_device`.

prov/efa/src/efa_ep.c
darrylabbate: > If there are multiple interfaces, any one's p2p_require_by_impl is true should make hmem_p2p_opt to P2P_REQUIRED. Better wrap this up in a utility function

That directly contradicts your concern from yesterday:

> It is possible for an instance to have both Neuron library and CUDA library installed. With this, neuron setting end up always overwrite CUDA setting.

The logic in this PR matches our current behavior. What you're suggesting would impose a tighter restriction on the p2p opt initialization. Maybe that's fine, but I want to make sure this is what we actually want.

darrylabbate:
> The commit message of 1st commit is not entirely actuate. It not only introduced `p2p_requred_by_impl`, but also renamed `p2p_supprted` to `p2p_supported_by_device`.

It's there in the first commit message: https://github.com/ofiwg/libfabric/pull/8351/commits/90747e67bfda1dfd8810e12e3f510794a0f61cea

```
commit 90747e67bfda1dfd8810e12e3f510794a0f61cea
Author: Darryl Abbate <drl@amazon.com>
Date:   Tue Dec 13 22:17:48 2022 -0800

    prov/efa: Add p2p_required_by_impl to efa_hmem_info

    This adds a p2p_required_by_impl field to the efa_hmem_info struct as a
    way to track the minimum p2p requirements for each supported FI_HMEM
    interface. Since the p2p requirements of each interface is more-or-less
    unwritten/institutional knowledge, it should be defined in a single
    place and queried as needed, instead of hardcoding in various places
    throughout the codebase or deducing from a combination of other
    parameters.

    This also renames p2p_supported to p2p_supported_by_device to better
    indicate the intention of the struct field.

```

prov/efa/src/efa_ep.c
wzamazon: >Maybe that's fine, but I want to make sure this is what we actually want.

I agree to not change behavior in this PR.

Can you add a TODO note here about this? I think we need the tighter restrictions eventually.

prov/efa/src/rxr/rxr_ep.c
darrylabbate: > after we go through all interfaces

This would diverge from our current behavior. We currently iterate through until we encounter the first initialized interface and break. Validating the p2p option against *every* initialized interface would be new logic. Is this what we're doing going forward?

wzamazon: Yes. I think so. Current behavior is written with single interface in mind. Multiple interface logic should like this

prov/efa/src/efa_hmem.c
darrylabbate: I'll remove the comment

prov/efa/src/efa_hmem.c
darrylabbate: `p2p_opt` here refers to the `fi_setopt()`/`FI_OPT_FI_HMEM_P2P` options, which is what's being validated in this function. This function is appropriately named. Including `set` in the function name implies setting of the user opt. The caller doesn't need to be made aware the function modifies internal `efa_hmem_info` state as a result of `FI_HMEM_P2P_DISABLED`.

wzamazon: >The caller doesn't need to be made aware the function modifies internal efa_hmem_info state as a result of FI_HMEM_P2P_DISABLED.

I am not sure. the caller of `fi_setopt` does not need to know `efa_hmem_info` has been updated. But this function is called by an internal function.

I understand the behavior might not be easy to be reflected in function name. Can you added the behavior (updating disabled_by_user) to the doxgen header of this function?

prov/efa/src/rxr/rxr_ep.c
darrylabbate: `s/ret/err/g` makes sense when the value of `efa_hmem_validate_p2p_opt()` is only returned in a non-successful case, but as it stands, we also return the value from `efa_hmem_validate_p2p_opt()` if it's `FI_SUCCESS`

wzamazon: FI_SUCESS is 0, which is no error

darrylabbate: That's my point. Having `return err` when `err` can be `FI_SUCCESS` isn't quite accurate.

wzamazon: It increase readability.  with `err = xxx`.  `if (!err)` means "if there is no error". Meanwhile `if (!ret)` is hard to understand.

This is widely used convention throughout the code base.

darrylabbate:
Latest changes:
- Added `TODO` comments where the `FI_HMEM` interfaces are being iterated over to note we want to impose tighter restrictions in the future
- Added a note in the doxygen comment for `efa_hmem_validate_p2p_opt()` that it sets `p2p_disabled_by_user` to `true` when `p2p_opt` is `FI_HMEM_P2P_DISABLED` (if the interface doesn't require p2p)

Non-changes:
- The `s/p2p_supported/p2p_supported_by_device/g` change was already documented in the first commit message
- Validating/initializing the p2p options against *every* interface is out of scope for this PR. The current implemented logic is transposed from the existing logic.

wzamazon (approved):
It has been a long journey.

Thank you for your patience!

PR #8354, opened 2022-12-15
core/hmem_cuda: adjust FI_HMEM_CUDA_ENABLE_XFER behavior

wzamazon:
The environment variable FI_HMEM_CUDA_ENABLE_XFER control whether CUDA API can be used to copy data. Currently, its help string states that it default value is true, which means by default a provider is safe to make CUDA calls.

However, by default it is not always safe for a provider to make CUDA calls, and some providers does not make CUDA calls by default. Such provider still need to make CUDA calls when user explictly set this environment variable to true.

This patch used an enum (instead of a boolean) to store the value of the user setting. The enum has three possible values:

    CUDA_XFER_DISABLED means the environment variable is set to 0
    CUDA_XFER_ENABLED means the environment variable is set to 1
    CUDA_XFER_UNSPECIED means the environment variable is not set

It then exposed the value of this setting through a public function, such that provider can access.

It also updated the help string to:
"""
By default, each provider decide whether it will make CUDA calls. """
which better describes current situation.

It is worth noting that this patch did not change the current behavior of any provider because:

Prior to this patch, the value of this environment variable is not accessible to a provider.

Prior to this patch, if the envrionment variable was set to 0, cuda_ops.cudaMemcpy was set to a no-op and cuda_ipc_enabled was set to false. This patch maintained this behavior.

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -158,6 +158,18 @@ int cuda_get_handle(void *dev_buf, void **handle);
 int cuda_open_handle(void **handle, uint64_t device, void **ipc_ptr);
 int cuda_close_handle(void *ipc_ptr);
 int cuda_get_base_addr(const void *ptr, void **base, size_t *size);
+
+/*
+ * This enum lists all the possible situations about how
+ * the environment variable FI_HMEM_CUDA_ENABLE_XFER was set.
+ */
+enum cuda_xfer_setting {
+	CUDA_XFER_UNSPECIFIED, /* FI_HMEM_CUDA_ENABLE_XFER was not set */
+	CUDA_XFER_ENABLED, /* FI_HMEM_CUDA_ENABLE_XFER set to 1/true/yes */
+	CUDA_XFER_DISABLED /* FI_HMEM_CUDA_ENABLE_XFER set to 0/false/no */
+};
+
+enum cuda_xfer_setting cuda_get_xfer_setting(void);
 bool cuda_is_ipc_enabled(void);
 int cuda_get_ipc_handle_size(size_t *size);
 bool cuda_is_gdrcopy_enabled(void);
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -76,6 +76,7 @@ struct cuda_ops {
 };
 
 static bool hmem_cuda_use_gdrcopy;
+static enum cuda_xfer_setting hmem_cuda_xfer_setting;
 static bool cuda_ipc_enabled;
 static int cuda_device_count;
 
@@ -588,9 +589,8 @@ static int cuda_hmem_verify_devices(void)
 
 int cuda_hmem_init(void)
 {
-	int ret;
+	int ret, param_value;
 	int gdrcopy_ret;
-	bool cuda_enable_xfer;
 
 	fi_param_define(NULL, "hmem_cuda_use_gdrcopy", FI_PARAM_BOOL,
 			"Use gdrcopy to copy data to/from CUDA GPU memory. "
@@ -600,7 +600,8 @@ int cuda_hmem_init(void)
 			"Enable use of CUDA APIs for copying data to/from CUDA "
 			"GPU memory. This should be disabled if CUDA "
 			"operations on the default stream would result in a "
-			"deadlock in the application. (default: true)");
+			"deadlock in the application. (default: each provider "
+			"decide whether to use CUDA API).");
 
 	ret = cuda_hmem_dl_init();
 	if (ret != FI_SUCCESS)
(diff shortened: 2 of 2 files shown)


include/ofi_hmem.h
```diff
@@ -158,6 +158,19 @@ int cuda_get_handle(void *dev_buf, void **handle);
 int cuda_open_handle(void **handle, uint64_t device, void **ipc_ptr);
 int cuda_close_handle(void *ipc_ptr);
 int cuda_get_base_addr(const void *ptr, void **base, size_t *size);
+
+/*
+ * This enum lists all the poissble situations about how
+ * the environment variable FI_HMEM_CUDA_ENABLE_XFER was set.
+ */
+enum cuda_xfer_setting
+{
```
shefty, line 167: nit: Move { to enum line

include/ofi_hmem.h
```diff
@@ -158,6 +158,19 @@ int cuda_get_handle(void *dev_buf, void **handle);
...
+	CUDA_XFER_DISABLED = 0, /* FI_HMEM_CUDA_ENABLE_XFER set to 0/false/no */
+	CUDA_XFER_ENABLED = 1, /* FI_HMEM_CUDA_ENABLE_XFER set to 1/true/yes */
+	CUDA_XFER_UNSPECIFIED = 2, /* FI_HMEM_CUDA_ENABLE_XFER was not set */
+};
```
shefty, line 171: It's better for an enum if unspecified = 0, which will be the default value for any static variable, and aligns with other libfabric values where 0 indicates not set.

include/ofi_hmem.h
```diff
@@ -158,6 +158,19 @@ int cuda_get_handle(void *dev_buf, void **handle);
 int cuda_open_handle(void **handle, uint64_t device, void **ipc_ptr);
 int cuda_close_handle(void *ipc_ptr);
 int cuda_get_base_addr(const void *ptr, void **base, size_t *size);
+
+/*
+ * This enum lists all the poissble situations about how
```
darrylabbate, line 163: Typo: `poissble`

src/hmem_cuda.c
```diff
@@ -610,10 +611,10 @@ int cuda_hmem_init(void)
 	if (ret != FI_SUCCESS)
 		goto dl_cleanup;
 
-	ret = 1;
+	param_value = 1;
 	fi_param_get_bool(NULL, "hmem_cuda_use_gdrcopy",
-			  &ret);
-	hmem_cuda_use_gdrcopy = (ret != 0);
+			  &param_value);
```
shefty, line 616: You can check the result from fi_param_get to know if the variable was set by the app (ret == 0), and update the value accordingly.  E.g.
```
ret = fi_param_get(..)
if (!ret)
    hmem_cuda_use_gdr_copy = param_value ? XFER_ENABLED : XFER_DISABLED
```

src/hmem_cuda.c
```diff
@@ -744,6 +751,11 @@ int cuda_host_unregister(void *ptr)
 	return -FI_EIO;
 }
 
+enum cuda_xfer_setting cuda_get_xfer_setting()
+{
+	return hmem_cuda_xfer_setting;
+}
```
shefty, line 757: Why add a wrapper function, versus just defining hmem_cuda_xfer as an extern and reading that directly?

src/hmem_cuda.c
```diff
@@ -744,6 +751,11 @@ int cuda_host_unregister(void *ptr)
 	return -FI_EIO;
 }
 
+enum cuda_xfer_setting cuda_get_xfer_setting()
```
darrylabbate, line 754: nit: Add `void` inside the parameter list of the function signature

src/hmem_cuda.c
wzamazon: I was following the convention that other similar variables used, like `cuda_ipc_enabled` was returned via function `cuda_is_ipc_enabled()`. I do not have preference. Do you perfer extern variable over function?

include/ofi_hmem.h
wzamazon: fixed

include/ofi_hmem.h
wzamazon: fixed in new revision

src/hmem_cuda.c
wzamazon: This gdrcopy change is out of scope for this PR. I removed this change in new revision.

include/ofi_hmem.h
wzamazon: Thank you! fixed in new revision

src/hmem_cuda.c
wzamazon: fixed in new revision

darrylabbate (approved):

src/hmem_cuda.c
```diff
@@ -625,17 +626,23 @@ int cuda_hmem_init(void)
...
-	ret = 1;
-	fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &ret);
-	cuda_enable_xfer = (ret != 0);
+	param_value = 1;
+	ret = fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &param_value);
+	if (ret == -FI_ENODATA) {
+		hmem_cuda_xfer_setting = CUDA_XFER_UNSPECIFIED;
+	} else {
+		hmem_cuda_xfer_setting = (param_value > 0) ? CUDA_XFER_ENABLED : CUDA_XFER_DISABLED;
+	}
```
shefty, line 635: It would be safer to write this as:
```
if (!ret) {
    hmem_cuda_xfer_setting = ...
}
```
fi_param_get can return other failure codes (EINVAL, ENOENT) -- neither of those should happen.

shefty (approved):
Approving, assuming the check from fi_param_get is updated.  Thanks!

src/hmem_cuda.c
wzamazon: Updated to 

```
         if (ret) {
		hmem_cuda_xfer_setting = CUDA_XFER_UNSPECIFIED;
	} else {
		hmem_cuda_xfer_setting = (param_value > 0) ? CUDA_XFER_ENABLED : CUDA_XFER_DISABLED;
	}
```

as suggested.

shefty (approved):

PR #8357, opened 2022-12-16
prov/efa: use cuda_xfer_setting to decide whether to use gdrcopy

wzamazon:
When cudaMemcpy is allowed, we want to use it and do not want to use gdrcopy. Gdrcopy is used when the memory is registered with gdrcopy.

Whether cudaMemcpy is allowed can now be determined by calling cuda_get_xfer_setting(). This patch therefore do not register memory with gdrcopy when cudaMemcpy is allowed.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -180,6 +180,24 @@ static struct fi_ops efa_mr_cache_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+/**
+ * @brief determine whether gdrcopy should used for a MR
+ *
+ * gdrcopy should be used for CUDA memory when cudaMemcpy
+ * is not allowed, and gdrcopy is available. cudaMemcpy
+ * has bettern host to device performance.
+ *
+ * @param[in] efa_mr EFA memory registration
+ *
+ * @return a boolean value
+ * @related efa_mr
+ */
+static inline
+bool efa_mr_use_gdrcopy(struct efa_mr *efa_mr)
+{
+	return efa_mr->peer.iface == FI_HMEM_CUDA && cuda_get_xfer_setting()!=CUDA_XFER_ENABLED && cuda_is_gdrcopy_enabled();
+}
+
 /*
  * @brief Validate HMEM attributes and populate efa_mr struct
  *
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -239,12 +239,18 @@ static int efa_mr_hmem_setup(struct efa_mr *efa_mr,
...
-		if (err) {
-			EFA_WARN(FI_LOG_MR,
-				 "Unable to register handle for GPU memory. err: %d buf: %p len: %zu\n",
-				 err, attr->mr_iov->iov_base, attr->mr_iov->iov_len);
-			return err;
+
+		/* When cudaMemcpy is allowed (cuda_xfer_setting==CUDA_XFER_ENABLED), use cudaMemcpy (which does not need registration)
+		 * Otherwise, use gdrcopy and register the memory with gdrcopy.
+		 */
+		if (cuda_get_xfer_setting()!=CUDA_XFER_ENABLED) {
```
lrbison, line 246: Although registration for cudamemcpy itself is not required, I think we may depend on the registration time to allow libfabric to set CU_POINTER_ATTRIBUTE_SYNC_MEMOPS for the allocation so that we can ensure our memcpy's are synchronous.

a-szegel, line 246: What happens when someone sets `CUDA_XFER_ENABLED == TRUE` and `HMEM_CUDA_USE_GDR_COPY == TRUE`

a-szegel: What happens is that we register as a cuda buffer, but the util code ries to go down the GDR copy path and we hit an assert.

wzamazon: Not sure I understand.

The call to sync_memops happens inside efa provider:
https://github.com/ofiwg/libfabric/blob/1a25af05d48ed49cf3c3075c0df99f4a09b0068c/prov/efa/src/efa_mr.c#L235

wzamazon: >  What happens when someone sets CUDA_XFER_ENABLED == TRUE and HMEM_CUDA_USE_GDR_COPY == TRUE

We use cudaMemcpy in that case.

> What happens is that we register as a cuda buffer, but the util code ries to go down the GDR copy path and we hit an assert.

Not sure. where is the assert?

a-szegel: sorry.. im wrong

darrylabbate, line 246: This implies the user should not set `FI_HMEM_CUDA_ENABLE_XFER=1` if they want to use GDRCopy, but I was under the impression `FI_HMEM_CUDA_ENABLE_XFER=1` would still allow GDRCopy if it's available. Does this mean only `CUDA_XFER_UNSPECIFIED` would be the only way to use GDRCopy?

lrbison: ok yes, there it is 👍

wzamazon: Yes. cudaMemcpy should be used when it is allowed to be used. Gdrcopy is an alternative.

a-szegel (approved):

darrylabbate (commented):
I would suggest some kind of utility function/macro to abstract this condition so it can be easily reused elsewhere. I'll probably make use of it. Ex:

```c
#define EFA_MR_USE_GDRCOPY() \
    (cuda_get_xfer_setting() != CUDA_XFER_ENABLED && cuda_is_gdrcopy_enabled())
```

wzamazon:
@darrylabbate 

>I would suggest some kind of utility function/macro to abstract this condition so it can be easily reused elsewhere. I'll probably make use of it.

In new revision, I  added an inline function `efa_mr_use_gdrcopy`.

darrylabbate:
I think this looks good.

Your changes make me wonder if `use_gdrcopy` should be a property of `struct efa_mr`. It's unnecessary as long as `hmem_cuda_xfer_setting` and `hmem_cuda_use_gdrcopy` are only set in `cuda_hmem_init()` (otherwise `efa_mr_use_gdrcopy()` could return different values between reg and dereg calls), but it's something worth considering.

wzamazon:
>Your changes make me wonder if use_gdrcopy should be a property of struct efa_mr

This might be something worth adding later.

darrylabbate (approved):

wzamazon:
The " continuous-integration/appveyor/pr" failure is a real issue but it was introduced by a different PR.

PR #8358, opened 2022-12-16
src/hmem_cuda: implement cuda_get_xfer_setting for non cuda build

wzamazon:
Previous commit 148dfd7d7e3a introduced cuda_get_xfer_setting() but did not implement it for non-CUDA build. This patch fixed it.

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -839,6 +839,11 @@ int cuda_get_base_addr(const void *ptr, void **base, size_t *size)
 	return -FI_ENOSYS;
 }
 
+enum cuda_xfer_setting cuda_get_xfer_setting(void)
+{
+	return CUDA_XFER_DISABLED;
+}
+
 bool cuda_is_ipc_enabled(void)
 {
 	return false;


lrbison (approved):

src/hmem_cuda.c
```diff
@@ -839,6 +839,11 @@ int cuda_get_base_addr(const void *ptr, void **base, size_t *size)
 	return -FI_ENOSYS;
 }
 
+enum cuda_xfer_setting cuda_get_xfer_setting(void)
+{
+	return CUDA_XFER_UNSPECIFIED;
```
a-szegel, line 844: For a non cuda build, we should return not allowed.

a-szegel (changes requested):

src/hmem_cuda.c
a-szegel: CUDA_XFER_DISABLED

wzamazon: Done.

shefty (approved):

PR #8359, opened 2022-12-17
pytest/efa: handle device differences in test_device_selection

wzamazon:
The efa_test_device_selection assume both nodes have same device name/numbers, which is not always the case.

This patch addressed the issue.

### DIFF
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -113,3 +113,12 @@ def get_efa_domain_names(server_id):
 
     return efa_domain_names
 
+def get_efa_device_names(server_id):
+    domain_names = get_efa_domain_names(server_id)
+    device_names = set()
+    for domain_name in domain_names:
+        itemlist = domain_name.split("-")
+        assert len(itemlist) == 2
+        assert itemlist[1] in ["rdm", "dgrm"]
+        device_names.add(itemlist[0])
+    return list(device_names)
--- a/fabtests/pytest/efa/test_efa_device_selection.py
+++ b/fabtests/pytest/efa/test_efa_device_selection.py
@@ -1,39 +1,52 @@
+import copy
 import pytest
-
+from efa.efa_common import efa_retrieve_hw_counter_value, get_efa_device_names
+from common import ClientServerTest
 
 # This test must be run in serial mode because it checks the hw counter
 @pytest.mark.serial
 @pytest.mark.functional
 def test_efa_device_selection(cmdline_args):
-    from efa.efa_common import efa_retrieve_hw_counter_value, get_efa_domain_names
-    from common import ClientServerTest
 
     if cmdline_args.server_id == cmdline_args.client_id:
         pytest.skip("EFA device selection test requires 2 nodes")
         return
 
-    efa_domain_names = get_efa_domain_names(cmdline_args.server_id)
+    server_device_names = get_efa_device_names(cmdline_args.server_id)
+    client_device_names = get_efa_device_names(cmdline_args.client_id)
 
-    for efa_domain_name in efa_domain_names:
-        if '-rdm' in efa_domain_name:
-            assert not('-dgrm') in efa_domain_name
-            fabtest_opts = "fi_rdm_pingpong"
-        elif '-dgrm' in efa_domain_name:
-            assert not('-rdm') in efa_domain_name
-            fabtest_opts = "fi_dgram_pingpong -k"
+    server_num_devices = len(server_device_names)
+    client_num_devices = len(server_device_names)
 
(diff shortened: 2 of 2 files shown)


fabtests/pytest/efa/efa_common.py
```diff
@@ -113,3 +113,12 @@ def get_efa_domain_names(server_id):
 
     return efa_domain_names
 
+def get_efa_device_names(server_id):
+    domain_names = get_efa_domain_names(server_id)
+    device_names = set()
+    for domain_name in domain_names:
+        itemlist = domain_name.split("-")
+        assert len(itemlist)==2
```
sunkuamzn, line 121: nit: space

lrbison (approved):
Looks good.

fabtests/pytest/efa/test_efa_device_selection.py
```diff
@@ -1,39 +1,54 @@
...
 
-    efa_domain_names = get_efa_domain_names(cmdline_args.server_id)
+    server_device_names = get_efa_device_names(cmdline_args.server_id)
+    client_device_names = get_efa_device_names(cmdline_args.client_id)
+
+    server_num_devices = len(server_device_names)
+    client_num_devices = len(server_device_names)
+
+    for i in range(max(server_num_devices, client_num_devices)):
+
```
sunkuamzn, line 22: nit: blank line

fabtests/pytest/efa/test_efa_device_selection.py
```diff
@@ -1,39 +1,54 @@
...
+        server_device_idx = i % server_num_devices
+        client_device_idx = i % client_num_devices
+
+        server_device_name = server_device_names[server_device_idx]
+        client_device_name = client_device_names[client_device_idx]
+
+        for suffix in ["rdm", "dgrm"]:
+
```
sunkuamzn, line 30: nit: blank line

fabtests/pytest/efa/test_efa_device_selection.py
```diff
@@ -1,39 +1,54 @@
...
+            else:
+                command = "fi_dgram_pingpong -k" # efa provider requires prefix mode for dgram provider, hence "-k"
 
-        efa_device_name = efa_domain_name.split('-')[0]
+            server_domain_name = server_device_name + "-" + suffix
+            client_domain_name = client_device_name + "-" + suffix
 
-        server_tx_bytes_before_test = efa_retrieve_hw_counter_value(cmdline_args.server_id, "tx_bytes", efa_device_name)
-        client_tx_bytes_before_test = efa_retrieve_hw_counter_value(cmdline_args.client_id, "tx_bytes", efa_device_name)
+            cmdline_args_copy = copy.copy(cmdline_args)
```
sunkuamzn, line 42: `copy.copy` does a shallow copy. If you want to keep the original `cmdline_args` unchanged, you need to use `copy.deepcopy`

sunkuamzn: If `additional_server_arguments` and `additional_client_arguments` are not set in the original `cmdline_args`, then a shallow copy is good enough

lrbison: I think actually no copy is required at all (as long as ClientServerTest doesn't modify it) because each loop we just assign a new value to additional_*_arguments.

fabtests/pytest/efa/efa_common.py
a-szegel: we should be running flake8 formatter on these files too.

fabtests/pytest/efa/test_efa_device_selection.py
wzamazon: I think copy is needed because cmdline_args is a fixture created once during initialization.

I think we can change the scope of `cmdline_args`. 

@sunkuamzn  I tried to use deepcopy, but it failed with an error message about not able to deepcopy a pattern object?
Since copy is fine here, I will keep this

fabtests/pytest/efa/efa_common.py
wzamazon: added

fabtests/pytest/efa/test_efa_device_selection.py
wzamazon: removed

fabtests/pytest/efa/test_efa_device_selection.py
wzamazon: removed

fabtests/pytest/efa/test_efa_device_selection.py
lrbison: 👍

PR #8360, opened 2022-12-17
prov/efa: fix setopt hmem p2p opt and refactor

wzamazon:
fix a bug in fi_setopt of FI_HMEM_P2P_OPT, and a refactor

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -303,15 +303,14 @@ static int efa_hmem_info_init_synapseai(struct efa_hmem_info *synapseai_info, st
 
 /**
  * @brief   Validate an FI_OPT_FI_HMEM_P2P (FI_OPT_ENDPOINT) option for a
- *          specified HMEM interface. If p2p_opt is FI_HMEM_P2P_DISABLED, we
- *          update hmem_info[iface]->p2p_disabled_by_user to be true only if the
- *          interface doesn't require p2p.
+ *          specified HMEM interface.
+ *          Also update hmem_info[iface]->p2p_disabled_by_user accordingly.
  *
  * @param   domain  The efa_domain struct which contains an efa_hmem_info array
  * @param   iface   The fi_hmem_iface enum of the HMEM interface to validate
  * @param   p2p_opt The P2P option to validate
  *
- * @return  FI_SUCCESS if the P2P option is valid for the given interface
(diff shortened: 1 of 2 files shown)


prov/efa/src/efa_hmem.c
```diff
@@ -322,32 +321,39 @@ int efa_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem_iface
...
+	 * nor do they prohibit that p2p is reqruied by implementation. Therefore
+	 * they are always supported.
 	 */
-	case FI_HMEM_P2P_PREFERRED:
-	case FI_HMEM_P2P_ENABLED:
-		return info->p2p_supported_by_device && !info->p2p_required_by_impl
-			? FI_SUCCESS
-			: -FI_EOPNOTSUPP;
+	if (p2p_opt == FI_HMEM_P2P_PREFERRED || p2p_opt == FI_HMEM_P2P_ENABLED) {
+		info->p2p_disabled_by_user = false;
```
darrylabbate, line 336: Aside from removing the conditions `info->p2p_supported_by_device && !info->p2p_required_by_impl` to comply with the Libfabric API spec, this line (and 344) are the only real notable changes. This change is necessary since it updates `info->p2p_disabled_by_user` regardless of `p2p_opt`.

Everything else is just style preference. Naturally I think the `switch`...`case` is cleaner/easier to understand.

```c
switch (p2p_opt) {
case FI_HMEM_P2P_REQUIRED:
    if (!info->p2p_supported_by_device)
        return -FI_EOPNOTSUPP;
    /* Fall-through */
case FI_HMEM_P2P_PREFERRED:
case FI_HMEM_P2P_ENABLED:
    return 0;
case FI_HMEM_P2P_DISABLED:
    if (info->p2p_required_by_impl)
        return -FI_EOPNOTSUPP;
    info->p2p_disabled_by_user = true;
    return 0;
}
return -FI_EINVAL;
```

darrylabbate (approved):

prov/efa/src/efa_hmem.c
wzamazon: Thanks! Switched back to use switch case. I did not use fall through though.

darrylabbate (approved):
lgtm, thanks!

wzamazon:
pure efa change. ignore intel CI

PR #8361, opened 2022-12-17
src/ofi_cuda.h: remove this unused file

wzamazon:
The file src/of_cuda.h is not used any where. It should not be used either because it calls cudaMemcpy() directly.

Therefore this patch removed it.


shefty:
Interesting, this file isn't even part of the makefile or release package.  Thanks!

PR #8365, opened 2022-12-19
src/hmem_cuda: Allow CUDA IPC when P2P disabled

darrylabbate:
This effectively defers any limitation(s) on CUDA IPC to the initialization of `cuda_ipc_enabled`, which in turn depends on the status of `FI_HMEM_CUDA_ENABLE_XFER`

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -758,7 +758,7 @@ enum cuda_xfer_setting cuda_get_xfer_setting(void)
 
 bool cuda_is_ipc_enabled(void)
 {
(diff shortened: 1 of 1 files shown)


wzamazon (approved):
I think we should mention that p2p is for network to access hmem memory, so disable p2p should not disable IPC.

shefty (approved):
I agree with @wzamazon that the commit message could be clearer that p2p (used for internode transfers) is independent from IPC transfers (used for intranode)

wzamazon:
bot:aws:retest

PR #8367, opened 2022-12-20
ft_sync exchanges 0-byte messages

lrbison:
This change prevents calling ft_sync() from modifying tx_buf and rx_buf,
as it did previously when syncing with a 1-byte message.  This is necessary
to prepare for a PR which can verify rdm_atomic operations.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -2659,11 +2659,11 @@ int ft_sync()
 
 	if (opts.dst_addr) {
 		if (!(opts.options & FT_OPT_OOB_SYNC)) {
-			ret = ft_tx(ep, remote_fi_addr, 1, &tx_ctx);
+			ret = ft_tx(ep, remote_fi_addr, 0, &tx_ctx);
 			if (ret)
 				return ret;
 
-			ret = ft_rx(ep, 1);
+			ret = ft_rx(ep, 0);
 		} else {
 			ret = ft_sock_send(oob_sock, &buf, 1);
 			if (ret)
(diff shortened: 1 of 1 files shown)


fabtests/common/shared.c
```diff
@@ -2659,33 +2659,33 @@ int ft_sync()
...
 		if (!(opts.options & FT_OPT_OOB_SYNC)) {
-			ret = ft_tx(ep, remote_fi_addr, 1, &tx_ctx);
+			ret = ft_tx(ep, remote_fi_addr, 0, &tx_ctx);
 			if (ret)
 				return ret;
 
 			ret = ft_rx(ep, 1);
 		} else {
-			ret = ft_sock_send(oob_sock, &buf, 1);
+			ret = ft_sock_send(oob_sock, &buf, 0);
```
shefty, line 2668: Can you update the commit message to say why this change is made?  Note the receiving 0-bytes makes it impossible to know if recv returned 0 because that's how many bytes were requested or the connection was torn down.  I also don't know what the behavior is if recv is called with 0-bytes and there's no data to be read because the send hasn't happened.  Did you confirm this sort of sync still works?

lrbison: Sure, I can update the commit message with something short.  More detail for you here:

I'm working on making a version of fi_rdm_atomic that validates the atomic operation results in the expected answer so that we can gain confidence in some of the recent HMEM atomics we've been working on in prov/efa.  I realized the current ft_sync implementation was incompatible because the tx and rx modify the (global) buffers invalidating the check.


I don't actually need to change the `ft_sock_send` lines, because those operate on that local `buf` instead of the rx_buf/tx_buf.  I'll put those back to 1 byte to avoid the recv confusion.  After testing the out-of-band sync I think you are right that recv isn't really doing anything.

shefty:
Can you please rebase to ensure that this is aligned with the latest main?

wzamazon:
Can you add `fabtests/common/shared.c` to the title of the commit?

lrbison:
Putting this on the shelf considering the problem with SHM.

PR #8368, opened 2022-12-21
prov/shm: Fix HMEM Support for Atomics

a-szegel:
Fixed the current implementation or HMEM + SHM.

### DIFF
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -70,7 +70,7 @@ smr_try_progress_from_sar(struct smr_ep *ep, struct smr_region *smr,
 {
 	if (*bytes_done < cmd->msg.hdr.size) {
 		if (smr_env.use_dsa_sar && iface == FI_HMEM_SYSTEM) {
-			(void) smr_dsa_copy_from_sar(ep, sar_pool, resp, cmd, 
+			(void) smr_dsa_copy_from_sar(ep, sar_pool, resp, cmd,
 					iov, iov_count, bytes_done, entry_ptr);
 			return;
 		} else {
@@ -569,29 +569,59 @@ out:
 }
 
 static void smr_do_atomic(void *src, void *dst, void *cmp, enum fi_datatype datatype,
-			  enum fi_op op, size_t cnt, uint16_t flags)
+                          enum fi_op op, size_t cnt, uint16_t flags,
+                          enum fi_hmem_iface iface, uint64_t device)
 {
 	char tmp_result[SMR_INJECT_SIZE];
+	char tmp_dst[SMR_INJECT_SIZE];
 
 	if (ofi_atomic_isswap_op(op)) {
-		ofi_atomic_swap_handler(op, datatype, dst, src, cmp,
-					tmp_result, cnt);
+		if (iface == FI_HMEM_SYSTEM) {
+			ofi_atomic_swap_handler(op, datatype, dst, src, cmp, tmp_result, cnt);
+		} else {
+			ofi_copy_from_hmem(iface, device, tmp_dst, dst,
+			                   cnt * ofi_datatype_size(datatype));
+			ofi_atomic_swap_handler(op, datatype, tmp_dst, src, cmp, tmp_result,
+			                        cnt);
+			ofi_copy_to_hmem(iface, device, dst, tmp_dst,
+			                 cnt * ofi_datatype_size(datatype));
+		}
 	} else if (flags & SMR_RMA_REQ && ofi_atomic_isreadwrite_op(op)) {
-		ofi_atomic_readwrite_handler(op, datatype, dst, src,
-					     tmp_result, cnt);
+		if (iface == FI_HMEM_SYSTEM) {
+			ofi_atomic_readwrite_handler(op, datatype, dst, src, tmp_result, cnt);
+		} else {
+			ofi_copy_from_hmem(iface, device, tmp_dst, dst,
+			                   cnt * ofi_datatype_size(datatype));
+			ofi_atomic_readwrite_handler(op, datatype, tmp_dst, src, tmp_result,
+			                             cnt);
+			ofi_copy_to_hmem(iface, device, dst, tmp_dst,
+			                 cnt * ofi_datatype_size(datatype));
+		}
 	} else if (ofi_atomic_iswrite_op(op)) {
-		ofi_atomic_write_handler(op, datatype, dst, src, cnt);
+		if (iface == FI_HMEM_SYSTEM) {
+			ofi_atomic_write_handler(op, datatype, dst, src, cnt);
+		} else {
+			ofi_copy_from_hmem(iface, device, tmp_dst, dst,
+			                   cnt * ofi_datatype_size(datatype));
+			ofi_atomic_write_handler(op, datatype, tmp_dst, src,
+			                         cnt);
+			ofi_copy_to_hmem(iface, device, dst, tmp_dst,
+			                 cnt * ofi_datatype_size(datatype));
+		}
 	} else {
 		FI_WARN(&smr_prov, FI_LOG_EP_DATA,
 			"invalid atomic operation\n");
 	}
 
-	if (flags & SMR_RMA_REQ)
-		memcpy(src, op == FI_ATOMIC_READ ? dst : tmp_result,
-		       cnt * ofi_datatype_size(datatype));
+	if (flags & SMR_RMA_REQ) {
+		ofi_copy_from_hmem(iface, device, src,
+		                   op == FI_ATOMIC_READ ? dst : tmp_result,
+		                   cnt * ofi_datatype_size(datatype));
+	}
 }
 
-static int smr_progress_inline_atomic(struct smr_cmd *cmd, struct fi_ioc *ioc,
+static int smr_progress_inline_atomic(enum fi_hmem_iface iface, uint64_t device,
+			       struct smr_cmd *cmd, struct fi_ioc *ioc,
 			       size_t ioc_count, size_t *len)
 {
 	int i;
@@ -602,7 +632,7 @@ static int smr_progress_inline_atomic(struct smr_cmd *cmd, struct fi_ioc *ioc,
 	for (i = *len = 0; i < ioc_count && *len < cmd->msg.hdr.size; i++) {
 		smr_do_atomic(&src[*len], ioc[i].addr, NULL,
 			      cmd->msg.hdr.datatype, cmd->msg.hdr.atomic_op,
-			      ioc[i].count, cmd->msg.hdr.op_flags);
+			      ioc[i].count, cmd->msg.hdr.op_flags, iface, device);
 		*len += ioc[i].count * ofi_datatype_size(cmd->msg.hdr.datatype);
 	}
 
@@ -614,9 +644,9 @@ static int smr_progress_inline_atomic(struct smr_cmd *cmd, struct fi_ioc *ioc,
 	return FI_SUCCESS;
 }
 
-static int smr_progress_inject_atomic(struct smr_cmd *cmd, struct fi_ioc *ioc,
-			       size_t ioc_count, size_t *len,
-			       struct smr_ep *ep, int err)
+static int smr_progress_inject_atomic(enum fi_hmem_iface iface, uint64_t device,
+			       struct smr_cmd *cmd, struct fi_ioc *ioc, size_t ioc_count,
+			       size_t *len, struct smr_ep *ep, int err)
 {
 	struct smr_inject_buf *tx_buf;
 	size_t inj_offset;
@@ -642,7 +672,7 @@ static int smr_progress_inject_atomic(struct smr_cmd *cmd, struct fi_ioc *ioc,
 	for (i = *len = 0; i < ioc_count && *len < cmd->msg.hdr.size; i++) {
 		smr_do_atomic(&src[*len], ioc[i].addr, comp ? &comp[*len] : NULL,
 			      cmd->msg.hdr.datatype, cmd->msg.hdr.atomic_op,
-			      ioc[i].count, cmd->msg.hdr.op_flags);
+			      ioc[i].count, cmd->msg.hdr.op_flags, iface, device);
 		*len += ioc[i].count * ofi_datatype_size(cmd->msg.hdr.datatype);
 	}
 
@@ -959,6 +989,9 @@ static int smr_progress_cmd_rma(struct smr_ep *ep, struct smr_cmd *cmd)
 
 static int smr_progress_cmd_atomic(struct smr_ep *ep, struct smr_cmd *cmd)
 {
+	void *mr;
+	enum fi_hmem_iface iface;
+	uint64_t device;
 	struct smr_region *peer_smr;
 	struct smr_domain *domain;
 	struct smr_cmd *rma_cmd;
@@ -986,6 +1019,10 @@ static int smr_progress_cmd_atomic(struct smr_ep *ep, struct smr_cmd *cmd)
 		if (ret)
 			break;
 
+		mr = ofi_mr_map_get(&ep->util_ep.domain->mr_map,
+		                    rma_cmd->rma.rma_iov[ioc_count].key);
+		iface = smr_get_mr_hmem_iface(ep->util_ep.domain, &mr, &device);
+
 		ioc[ioc_count].addr = (void *) rma_cmd->rma.rma_ioc[ioc_count].addr;
 		ioc[ioc_count].count = rma_cmd->rma.rma_ioc[ioc_count].count;
 	}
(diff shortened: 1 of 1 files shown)


wzamazon:
Can you move the EFA change to a different PR? keep this for shm provider only.

prov/shm/src/smr_progress.c
```diff
@@ -569,29 +569,47 @@ static int smr_progress_ipc(struct smr_cmd *cmd, enum fi_hmem_iface iface,
 }
 
 static void smr_do_atomic(void *src, void *dst, void *cmp, enum fi_datatype datatype,
-			  enum fi_op op, size_t cnt, uint16_t flags)
+			  enum fi_op op, size_t cnt, uint16_t flags, enum fi_hmem_iface iface, uint64_t device)
```
shefty, line 572: Please keep line lengths to about 80 characters for better readability. Comment applies throughout patch.

prov/shm/src/smr_progress.c
```diff
@@ -569,29 +569,47 @@ static int smr_progress_ipc(struct smr_cmd *cmd, enum fi_hmem_iface iface,
...
 {
 	char tmp_result[SMR_INJECT_SIZE];
 
 	if (ofi_atomic_isswap_op(op)) {
-		ofi_atomic_swap_handler(op, datatype, dst, src, cmp,
-					tmp_result, cnt);
+		if (iface == FI_HMEM_SYSTEM) {
+			ofi_atomic_swap_handler(op, datatype, dst, src, cmp, tmp_result, cnt);
+		} else {
+			char tmp_dst[SMR_INJECT_SIZE];
```
shefty, line 580: Declare variables at the top of the function.  That, too, helps with readability.

prov/shm/src/smr_progress.c
```diff
@@ -569,29 +569,47 @@ static int smr_progress_ipc(struct smr_cmd *cmd, enum fi_hmem_iface iface,
...
+			ofi_atomic_swap_handler(op, datatype, tmp_dst, src, cmp, tmp_result, cnt);
```
shefty, line 581: How does this work?  tmp_dst in all of these cases is uninitialized.

a-szegel:
There SHM has some bad performance issues with fi_rdm_atomic test.

HMEM Atomics over EFA (bottom of fi_rdm_atomic EFA on two nodes):
```
FI_INT8_FI_MSWAP_compare_lat                      1       1k      1000        0.11s      0.01     113.30       0.01
FI_UINT8_FI_MSWAP_compare_lat                     1       1k      1000        0.11s      0.01     113.17       0.01
FI_INT16_FI_MSWAP_compare_lat                     2       1k      1.9k        0.11s      0.02     113.19       0.01
FI_UINT16_FI_MSWAP_compare_lat                    2       1k      1.9k        0.11s      0.02     113.45       0.01
FI_INT32_FI_MSWAP_compare_lat                     4       1k      3.9k        0.11s      0.04     113.43       0.01
FI_UINT32_FI_MSWAP_compare_lat                    4       1k      3.9k        0.11s      0.04     113.44       0.01
FI_INT64_FI_MSWAP_compare_lat                     8       1k      7.8k        0.11s      0.07     113.42       0.01
FI_UINT64_FI_MSWAP_compare_lat                    8       1k      7.8k        0.11s      0.07     113.45       0.01
```

HMEM Atomics over SHM (bottom of fi_rdm_atomic EFA -> SHM):
```
FI_INT8_FI_MSWAP_compare_lat                      1       1k      1000        0.61s      0.00     612.11       0.00
FI_UINT8_FI_MSWAP_compare_lat                     1       1k      1000        0.61s      0.00     609.69       0.00
FI_INT16_FI_MSWAP_compare_lat                     2       1k      1.9k        0.61s      0.00     614.68       0.00
FI_UINT16_FI_MSWAP_compare_lat                    2       1k      1.9k        0.61s      0.00     612.67       0.00
FI_INT32_FI_MSWAP_compare_lat                     4       1k      3.9k        0.61s      0.01     614.47       0.00
FI_UINT32_FI_MSWAP_compare_lat                    4       1k      3.9k        0.61s      0.01     610.72       0.00
FI_INT64_FI_MSWAP_compare_lat                     8       1k      7.8k        0.61s      0.01     614.38       0.00
FI_UINT64_FI_MSWAP_compare_lat                    8       1k      7.8k        0.61s      0.01     613.91       0.00
```

a-szegel:
EFA on two instances with FI_HMEM_SYSTEM:
```
FI_INT8_FI_MSWAP_compare_lat                      1       1k      1000        0.04s      0.03      39.47       0.03
FI_UINT8_FI_MSWAP_compare_lat                     1       1k      1000        0.04s      0.03      39.41       0.03
FI_INT16_FI_MSWAP_compare_lat                     2       1k      1.9k        0.04s      0.05      39.50       0.03
FI_UINT16_FI_MSWAP_compare_lat                    2       1k      1.9k        0.04s      0.05      39.62       0.03
FI_INT32_FI_MSWAP_compare_lat                     4       1k      3.9k        0.04s      0.10      39.55       0.03
FI_UINT32_FI_MSWAP_compare_lat                    4       1k      3.9k        0.04s      0.10      39.44       0.03
FI_INT64_FI_MSWAP_compare_lat                     8       1k      7.8k        0.04s      0.20      39.39       0.03
FI_UINT64_FI_MSWAP_compare_lat                    8       1k      7.8k        0.04s      0.20      39.41       0.03
```

EFA + SHM FI_HMEM_SYSTEM:
```
FI_INT8_FI_MSWAP_compare_lat                      1       1k      1000        0.01s      0.17       6.05       0.17
FI_UINT8_FI_MSWAP_compare_lat                     1       1k      1000        0.01s      0.17       6.03       0.17
FI_INT16_FI_MSWAP_compare_lat                     2       1k      1.9k        0.01s      0.34       5.92       0.17
FI_UINT16_FI_MSWAP_compare_lat                    2       1k      1.9k        0.01s      0.34       5.95       0.17
FI_INT32_FI_MSWAP_compare_lat                     4       1k      3.9k        0.01s      0.65       6.12       0.16
FI_UINT32_FI_MSWAP_compare_lat                    4       1k      3.9k        0.01s      0.67       5.96       0.17
FI_INT64_FI_MSWAP_compare_lat                     8       1k      7.8k        0.01s      1.33       6.03       0.17
FI_UINT64_FI_MSWAP_compare_lat                    8       1k      7.8k        0.01s      1.34       5.95       0.17
```

Pretty sure this has to do with cuda locks... (although two different devices (-i DEV_NUM) didn't fix it either)

prov/shm/src/smr_progress.c
a-szegel: Good Catch... passes fi_rdm_atomic b/c it currently does no data validation.  Luke has a branch that is adding data validation to the test.  I will re-comment on this once I have successfully ran shm with his verification.

prov/shm/src/smr_progress.c
a-szegel: done

prov/shm/src/smr_progress.c
a-szegel: done

prov/shm/src/smr_progress.c
shefty: Try using ubertest.  That should include data verification checks

a-szegel:
fi_rdm_atomic FI_HMEM_CUDA EFA Single Instance SHM turned off:
```
FI_INT8_FI_MSWAP_compare_lat                      1       1k      1000        0.43s      0.00     427.41       0.00
FI_UINT8_FI_MSWAP_compare_lat                     1       1k      1000        0.43s      0.00     427.41       0.00
FI_INT16_FI_MSWAP_compare_lat                     2       1k      1.9k        0.43s      0.00     427.40       0.00
FI_UINT16_FI_MSWAP_compare_lat                    2       1k      1.9k        0.43s      0.00     427.33       0.00
FI_INT32_FI_MSWAP_compare_lat                     4       1k      3.9k        0.43s      0.01     427.14       0.00
FI_UINT32_FI_MSWAP_compare_lat                    4       1k      3.9k        0.43s      0.01     427.30       0.00
FI_INT64_FI_MSWAP_compare_lat                     8       1k      7.8k        0.43s      0.02     427.28       0.00
FI_UINT64_FI_MSWAP_compare_lat                    8       1k      7.8k        0.43s      0.02     427.14       0.00
```

fi_rdm_atomic FI_HMEM_SYSTEM EFA Single Instance SHM turned off:
```
FI_INT8_FI_MSWAP_compare_lat                      1       1k      1000        0.00s      0.31       3.24       0.31
FI_UINT8_FI_MSWAP_compare_lat                     1       1k      1000        0.00s      0.31       3.23       0.31
FI_INT16_FI_MSWAP_compare_lat                     2       1k      1.9k        0.00s      0.62       3.24       0.31
FI_UINT16_FI_MSWAP_compare_lat                    2       1k      1.9k        0.00s      0.62       3.25       0.31
FI_INT32_FI_MSWAP_compare_lat                     4       1k      3.9k        0.00s      1.23       3.26       0.31
FI_UINT32_FI_MSWAP_compare_lat                    4       1k      3.9k        0.00s      1.24       3.23       0.31
FI_INT64_FI_MSWAP_compare_lat                     8       1k      7.8k        0.00s      2.44       3.29       0.30
FI_UINT64_FI_MSWAP_compare_lat                    8       1k      7.8k        0.00s      2.45       3.26       0.31
```

Something about SHM is killing performance.

prov/shm/src/smr_progress.c
```diff
@@ -569,29 +569,59 @@ static int smr_progress_ipc(struct smr_cmd *cmd, enum fi_hmem_iface iface,
 }
 
 static void smr_do_atomic(void *src, void *dst, void *cmp, enum fi_datatype datatype,
-			  enum fi_op op, size_t cnt, uint16_t flags)
+                          enum fi_op op, size_t cnt, uint16_t flags, enum fi_hmem_iface
+                          iface, uint64_t device)
```
darrylabbate, line 573: nit: The type signature reads a little weird here with the line break between `enum fi_hmem_iface` and `iface`

a-szegel: done

prov/shm/src/smr_progress.c
a-szegel: @shefty I commented out the entire `smr_do_atomic()` method and ran fi_ubertest for SHM, and all 774 test passed.  I ran like this:
```
# client
./fi_ubertest -u fabtests/test_configs/shm/all.test 127.0.0.1
# server
./fi_ubertest
```

This means that fi_ubertest must not be doing any data validation checks.

aingerson: @a-szegel The all.test config file does not include the verfication tests - this is done in FT_TEST_UNIT (as opposed to latency/bandwidth). Try running with verify.test instead

aingerson (approved):
@a-szegel @wzamazon Changes themselves look fine to me. I noticed the performance issue last week and have been taking a look. I think it is a regression introduced by recent commits. I'm looking into it.

prov/shm/src/smr_progress.c
a-szegel: I ran with @lrbison's validation tests and it "seems" to be working.  Luke's new tests still have some intermittent failure problems (which seem to be fabtest related).  They validate data on each iteration, and 1 iteration fails per 1000 iterations on most tests (consistently).

a-szegel:
one last bit: These changes have the assumption that every IOV entry will correspond to the same type of HMEM.  I think the code is cleaner this way, but I can change `smr_progress_inject_atomic()` and `smr_progress_inline_atomic()` to take the endpoint, and the 2nd command message if desired.

aingerson:
@a-szegel All of the libfabric hmem iov code has the same assumption. I'm not sure it's specified in the man pages (it probably should be...) but I think that is a fine assumption to make across the board.

prov/shm/src/smr_progress.c
a-szegel: @aingerson Ran verify.test for fi_uber shm and 100% of tests passed.  I didn't see any FI_HMEM in the output, so I am guessing I need to modify the config in someway to get FI_HMEM to run.

aingerson: @a-szegel Thanks for testing! HMEM support is enabled with the following arguments
-H -> HMEM aware but use host memory
-D <iface> -> use iface for device allocation
-i <idx> -> which device to use for device allocation

For example to run CUDA device->host test, run the server and client as so:
fi_ubertest -D cuda
fi_ubertest -H -u <config>

a-szegel: It seems like the fabtests have a bug.  `ft_fill_buf()` is being called before `tx_msg_buf` is being initialized, so we are hitting an assert before we even get into the testing.

a-szegel:
@shefty @aingerson

Since fi_ubertest seems to be broken for hmem, is it ok if I take the atomic data validation from https://github.com/ofiwg/libfabric/pull/8378 for this PR?

prov/shm/src/smr_progress.c
```diff
@@ -569,29 +569,59 @@ static int smr_progress_ipc(struct smr_cmd *cmd, enum fi_hmem_iface iface,
...
-					tmp_result, cnt);
+		if (iface == FI_HMEM_SYSTEM) {
+			ofi_atomic_swap_handler(op, datatype, dst, src, cmp, tmp_result, cnt);
+		} else {
+			ofi_copy_from_hmem(iface, device, tmp_dst, dst,
+			                   cnt * ofi_datatype_size(datatype));
+			ofi_atomic_swap_handler(op, datatype, tmp_dst, src, cmp, tmp_result,
+			                        cnt);
+			ofi_copy_to_hmem(iface, device, dst, tmp_dst,
+			                 cnt * ofi_datatype_size(datatype));
```
shefty, line 587: We'll need to decide what it means to be atomic here.  These flows (including below) are only atomic if nothing else is touching the same regions, since we're manually doing read-modify-write.  In these flows, the modify portion doesn't need to use CPU atomics, though it's harmless to keep them.  That begs the question of whether we need CPU atomics in the system memory case.  If we need CPU atomics for system memory, then the non-system memory case would be broken.

a-szegel: On the [man page:](https://ofiwg.github.io/libfabric/v1.17.0/man/fi_atomic.3.html)

> The correctness of atomic operations on a target memory region is guaranteed only when performed by a single actor for a given window of time. An actor is defined as a single libfabric domain (identified by the domain name, and not an open instance of that domain), a coherent CPU complex, or other device (e.g. GPU) capable of performing atomic operations on the target memory. The results of atomic operations performed by multiple actors simultaneously are undefined. For example, issuing CPU based atomic operations to a target region concurrently being updated by NIC based atomics may leave the region’s data in an unknown state. The results of a first actor’s atomic operations must be visible to a second actor prior to the second actor issuing its own atomics.


It would appear that we do not need CPU atomics for system memory.

shefty: I'm assuming the shm provider is using locks to serialize a single process with multiple threads.  However, could 2 different processes access the same GPU memory?  The current atomic implementation to system memory works in this case, as the CPU atomics themselves are used.  In the case of GPU memory, the proposed change may break that.

Example: It looks like shm uses the domain name 'shm' for all domains.  That makes sense.  Supposed an app opens 2 domains, each with access to the same GPU memory region.  If 2 peers perform atomic updates to the GPU memory, but through different domains at the target, can we still be considered atomic?

I think the answer is no, and part of the reason is the use of memcpy.  We could read out the GPU region while another process is writing to the same region.  Even if the memcpy's were atomic, I don't think that's sufficient, since it's possible for both to read out the old value, but only one of the operations would end up being stored in the result.  (Imagine if both wanted to add 1 to the stored value.  The result should be value + 2, but we could end up with value +1.)

I don't know what the answer is here, but it seems like there needs to be some sort of global synchronization between processes when accessing non-system memory.  CPU atomics give us that sort of synchronization.  For GPUs, either we need our own, or we need to schedule the GPU to execute the atomic.

a-szegel: What is the reasoning behind defining an actor a libfabric domain identified by name, not instance of domain?

shefty: It is needed to ensure atomicity from the viewpoint of the remote peer.  Imagine if a domain is a NIC.  All atomic operations that target that NIC are guaranteed to be atomic, regardless of which endpoint is used to access the memory or open software instance.  In this case, the domain name is the name of the NIC.

a-szegel: So you are saying that we need an interprocess lock per domain (NIC), per memory region to make our emulated implementation correct?

a-szegel: or at the very minimum a 1 atomic per domain lock (if we don't care about performance).

shefty: Basically, yes, though I'm not sure about per memory region, at least not a libfabric MR.  We somehow need to serialize accesses to the same GPU memory location if we're going to copy the data of to the host, modify it, then move it back to the GPU.

Is there a way to invoke atomic operations on the accelerator itself?  That's what we do with system memory -- call CPU atomics.  This is likely a lot more work, but it's probably what we really want.

aingerson:
@a-szegel You're right, ubertest device validation was broken with the introduction of the host bounce buffer. I've fixed that with a patch. There is no support for hmem atomic validation because the fill/check functions in the ubertest code will have to make sure to use a host bounce buffer for those calculations. I started doing it but didn't get far. I'm on vacation until 1/4 so I can continue when I get back but feel free to take it and run with it while I'm gone.
https://github.com/aingerson/libfabric/tree/fabtests

PR #8373, opened 2022-12-21
prov/efa: allocate ibv_wr for RxR provider in rxr_pkt_entry

sunkuamzn:
This commit is part 2 of the endpoint separation
The changes in this commit are temporary and will removed when the endpoints are fully separated

### DIFF
--- a/prov/efa/src/efa_msg.c
+++ b/prov/efa/src/efa_msg.c
@@ -78,26 +78,28 @@ static inline void dump_msg(const struct fi_msg *msg, const char *context)
 }
 #endif /* EFA_MSG_DUMP */
 
-static void free_send_wr_list(struct ibv_send_wr *head)
+static void free_send_wr_list(struct ibv_send_wr *head, bool free)
 {
 	struct ibv_send_wr *wr = head;
 	struct ibv_send_wr *tmp;
 
 	while (wr) {
 		tmp = wr->next;
-		ofi_buf_free(container_of(wr, struct efa_send_wr, wr));
+		if (free)
+			ofi_buf_free(container_of(wr, struct efa_send_wr, wr));
 		wr = tmp;
 	}
 }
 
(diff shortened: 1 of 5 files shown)


prov/efa/src/efa_msg.c
```diff
@@ -145,7 +147,11 @@ static ssize_t efa_post_recv(struct efa_ep *ep, const struct fi_msg *msg, uint64
 	ssize_t err, post_recv_err;
 	size_t i;
 
-	ewr = ofi_buf_alloc(ep->recv_wr_pool);
+	if (!msg->data) {
```
darrylabbate, line 150: `msg->data` can still be `NULL`/`0` even with your changes?

sunkuamzn: It can be NULL/0 for the dgram provider

sunkuamzn: I think we should prevent someone from setting `msg->data` when they directly use the DGRAM provider

Not sure show though. Adding `assert(msg->data == 0)` [here](https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_msg.c#L408) would affect both RDM and DGRAM.

wzamazon:
This is not the right approach, we should looking into how to call `ibv_post_send` directly inside `rxr_pkt_entry_send`

wzamazon (changes requested):
try to call `ibv_post_send/recv` diectly from RxR, bypassing efa_post_send

sunkuamzn:
Calling `ibv_post_send`/`ibv_post_recv` directly is part 3. This PR is part 2. Should we combine part 2 with part 3? https://quip-amazon.com/x51HABgruwes/EFA-and-RXR-squash-endpoint-separation#temp:C:DHF763468a0d2384618951123c41

PR #8374, opened 2022-12-21
Fabtests/multinode: Fix hanging ze + verbs multinode tests

zachdworkin:
ze + zerbs was causing hangs due to the wrong processing order of messages.
Enabling FI_DIRECTED_RECV and properly initializing/using barrier buffers fixes this issue.


fabtests/multinode/src/core.c
```diff
@@ -55,9 +55,13 @@
 #include <assert.h>
 #include <hmem.h>
 
+#define BARRIER_SIZE	16
+#define BARRIER_KEY	0x1DEA
```
shefty, line 59: Barrier doesn't need a message size.  Even MPI uses a 0-byte barrier.  We should just use a specific key for barrier messages to match them correctly.  I'd rename KEY to TAG and use this as a tag message.  A MR key isn't useful

shefty (changes requested):

fabtests/multinode/src/core.c
zachdworkin: updated

fabtests/multinode/src/core.c
```diff
@@ -135,6 +140,17 @@ static int multi_setup_fabric(int argc, char **argv)
 	if (ret)
 		return ret;
 
+	if (fi->domain_attr->mr_mode & FI_MR_LOCAL) {
+		ret = fi_mr_reg(domain, rx_barrier,
+				FT_MAX_CTRL_MSG * pm_job.num_ranks,
+				FT_MSG_MR_ACCESS, 0, BARRIER_TAG, 0,
+				&mr_barrier, NULL);
```
shefty: We don't need a separate barrier buffer.  We only need to send a 0-byte message that uses the new tag.  The send will match with the receive.  Also, the memory registration key field is ignored unless the buffer is used for remote RMA operations.  But moving to a 0-byte barrier removes the need for this change anyway.

You probably don't even need directed receive as long as barrier messages use a separate tag, but I'm okay with that change.

fabtests/multinode/src/core.c
```diff
@@ -221,10 +237,11 @@ int multi_msg_recv()
 		offset = state.recvs_posted % opts.window_size;
 		assert(rx_ctx_arr[offset].state == OP_DONE);
 
+		remote_fi_addr = pm_job.fi_addrs[state.cur_source];
 		ret = ft_post_rx_buf(ep, opts.transfer_size,
 				     &rx_ctx_arr[offset].context,
 				     rx_ctx_arr[offset].buf,
-				     rx_ctx_arr[offset].desc, 0);
+				     rx_ctx_arr[offset].desc, 1);
```
shefty, line 242: This change shouldn't be needed.  A tag of 0 is fine.

fabtests/multinode/src/core.c
```diff
@@ -372,17 +389,19 @@ int send_recv_barrier(int sync)
...
 
 	for (i = 0; i < pm_job.num_ranks; i++) {
-		ret = ft_post_rx_buf(ep, opts.transfer_size,
-			     	     &barrier_rx_ctx[i],
-			     	     rx_buf, mr_desc, 0);
+		remote_fi_addr = pm_job.fi_addrs[i];
+		ret = ft_post_rx_buf(ep, FT_MAX_CTRL_MSG,
+				     &barrier_rx_ctx[i],
+				     rx_barrier + (FT_MAX_CTRL_MSG * i),
+				     barrier_desc, BARRIER_TAG);
```
shefty, line 394: Using a 0-byte message, we don't need the barrier context or buffers

zachdworkin: updated

fabtests/multinode/src/core.c
zachdworkin: This has to be 1 because of https://github.com/ofiwg/libfabric/blob/main/fabtests/common/shared.c#L2233

shefty: thanks!

fabtests/multinode/src/core.c
shefty: Comment still applies.  We can use a 0-byte send for barrier.  There's nothing that we to register, since there's no actual buffer that we need to exchange or post.

fabtests/functional/multi_ep.c
```diff
@@ -107,7 +107,8 @@ static int alloc_multi_ep_res()
 	}
 
 	ret = ft_reg_mr(fi, data_bufs, num_eps * 2 * opts.transfer_size,
-			ft_info_to_mr_access(fi), FT_MR_KEY + 1, &data_mr, &data_desc);
+			ft_info_to_mr_access(fi), FT_MR_KEY + 1, opts.iface, opts.device,
```
shefty, line 110: line length

fabtests/multinode/src/core.c
shefty: We still don't need to post a buffer -- that can be NULL, size 0, and no memory descriptor needed.

shefty (changes requested):

shefty:
Thanks!

PR #8378, opened 2022-12-22
fabtests/functional: Add data_validation option to fi_rdm_atomic

lrbison:
This change adds a mock for atomic operations, initializes values before executing each atomic, and checks the final values for consistency with the mocked operation.  This checking requires additional synchronization which affects performance, so it is off by default.

The data_validation is enabled using the '-v' flag.  When the flag is provided, a summary is printed and return value is EXIT_FAILURE if any validations fail.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -3760,6 +3760,11 @@ void ft_longopts_usage()
 		"spin until a debugger can be attached.");
 }
 
+void ft_verify_data_usage()
+{
+	FT_PRINT_OPTS_USAGE("-v", "enables data_integrity checks (implies -U)");
+}
+
 int debug_assert;
 
 int lopt_idx = 0;
@@ -3783,4 +3788,31 @@ int ft_parse_long_opts(int op, char *optarg)
 	default:
 		return EXIT_FAILURE;
 	}
+}
+
+int ft_sync_for_validation()
+{
+	char buf;
+	int ret;
+	void *mr_desc = NULL;
+
+	FT_POST(fi_send, ft_progress, txcq, tx_seq,
+			&tx_cq_cntr, "transmit (ft_sync_no_buffer)", ep, &buf, 1,
+			mr_desc, remote_fi_addr, NULL);
+
+	ret = ft_get_tx_comp(tx_seq);
+	if (ret) {
+		FT_ERR("failed to get tx completion during ft_sync_no_buffer!\n");
+		return ret;
+	}
+	ret = ft_get_rx_comp(rx_seq);
+	if (ret) {
+		FT_ERR("failed to get rx completion during ft_sync_no_buffer!\n");
+		return ret;
+	}
+
+	FT_POST(fi_recv, ft_progress, rxcq, rx_seq,
+			&rx_cq_cntr, "receive (ft_sync_no_buffer)", ep, &buf, 1,
+			mr_desc, remote_fi_addr, NULL);
+	return ret;
 }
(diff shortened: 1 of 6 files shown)


lrbison:
This should run on any supported HMEM and provider, as long as FI_DELIVERY_COMPLETE can be enabled.

I'm consistently seeing errors but only on the first (or sometimes last) iteration of a test after changing data types.  The rest pass.  This suggests I might have synchronization logic problems that I haven't yet figured out.

My output looks like this:
```
fi_rdm_atomic -p efa -E -v -I 10
```
```
FAILED: FI_ATOMIC_WRITE on FI_UINT8 had 1 of 10 tests fail data validation.
                First failure at trial 4, last failure at trial 4.
FAILED: FI_ATOMIC_WRITE on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_ATOMIC_READ on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_BXOR on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_LXOR on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 8, last failure at trial 8.
FAILED: FI_BAND on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_BOR on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_PROD on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_SUM on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_MAX on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
FAILED: FI_MIN on FI_INT8 had 1 of 10 tests fail data validation.
                First failure at trial 1, last failure at trial 1.
```

Note that all errors are on FI_INT8, but that's just because FI_INT8 is the first datatype tested.  All the rest seem to pass, unless you change the datatype to test only FI_INT32, in which case you get similar 1/N failures.

shefty:
Please just update ubertest to support HMEM, rather than duplicating the atomic code.  rdm_atomic is limited in its testing.  In the worst case, it should re-use the ofi_atomic.* implementation under ubertest (just move the files).

lrbison:
@shefty  Yes, I think we are realizing we don't have very much institutional knowledge about ubertest, and we probably need to fix that.  My understanding was that ubertest as a whole might not support HMEM testing and validation?  Perhaps I'm wrong. These validations do support HMEM.

Looks like I could have saved myself some time had I found the ofi_atomic.c beneath ubertests.  Thanks -- I'll take a look at those.

Would suggest pursuing ubertest to test HMEM atomics or do you think adding ofi_atomic.c's implementation of atomics to fi_rdm_atomic is a valid path to get HMEM testing?

fabtests/common/shared.c
```diff
@@ -3784,3 +3789,30 @@ int ft_parse_long_opts(int op, char *optarg)
 		return EXIT_FAILURE;
 	}
 }
+
+int ft_sync_for_validation()
```
lrbison, line 3793: I have discovered that this function is causing stack corruption.  I'll need to fix it if we decide to continue working this PR.

shefty:
I think adding validation to rdm_atomic is a valid path forward.  We just need to reuse the atomic code.  That's actually a copy of the libfabric atomic code.  Someday we should move to using a single version of the code.  I think that means moving the shared code between libfabric and fabtests under the fabtests directory.  That's not ideal from the viewpoint of libfabric, but I don't see a better option.  Anyway, for now, just moving ofi_atomic.* should work.

We will eventually need to update ubertest to support HMEM.  I'm guessing that's more involved.

shefty:
PR is outdated and needs to be rebased, plus updates are needed to reuse the atomic implementation.  Please either re-open this PR or open a new PR once updated and ready for further review.

PR #8381, opened 2022-12-23
man/atomic: Clarify Atomic Correctness Definition in Man Pages

a-szegel:


shefty:
thanks!

PR #8384, opened 2022-12-26
prov/efa: simplify implementation of DC longcts protocol

wzamazon:
Prior to this patch, the TX entry of a DC longcts
protocol was released when both all DATA packets' send completion have arrived and RECEIPT packet was released.

This is not ideal, because we want to write TX completion as soon as RECEIPT was received.

The current implementation is because of two things:

First, there is no information in the packet
entry can be used to determine whether a DATA packet is used by a DC longcts protocol or a normal longcts protocol,

Second, if a DATA packet is used by a normal longcts protocol, the op_entry's byte_acked must be updated.

Beacuse of 1, we have to update tx_entry->byte_acked when we receive the send completion of a DATA packet, which means the tx_entry must not be released when we received the send completion of the DATA packet.

As a result, we cannot release TX entry when we received the RECEIPT packet.

This patch addressed the issue by introdce a flag

     RXR_PKT_ENTRY_DC_LONGCTS_DATA

and set it in the pkt_entry->flags for DATA packet used by DC longcts protocol.

rxr_pkt_hand_data_send_completion() can check this flag to determine whether a DATA packet is used by a DC longcts protocol, and do not do anything for such a DATA packet.

The function rxr_pkt_handle_receipt_recv() was simplied to just write TX completion.

The two tx_entry flags RXR_LONGCTS_PROTOCOL and RXR_RECEIPT_RECEIVED was removed.

The two functions

    rxr_pkt_handle_dc_longcts_rtw_send_completion()
    rxr_pkt_handle_dc_longcts_rtm_send_completion()

was removed.

### DIFF
--- a/prov/efa/src/rxr/rxr_pkt_entry.h
+++ b/prov/efa/src/rxr/rxr_pkt_entry.h
@@ -40,6 +40,7 @@
 #define RXR_PKT_ENTRY_IN_USE		BIT_ULL(0) /**< this packet entry is being used */
 #define RXR_PKT_ENTRY_RNR_RETRANSMIT	BIT_ULL(1) /**< this packet entry encountered RNR and is being retransmitted*/
 #define RXR_PKT_ENTRY_LOCAL_READ	BIT_ULL(2) /**< this packet entry is used as context of a local read operation */
+#define RXR_PKT_ENTRY_DC_LONGCTS_DATA	BIT_ULL(3) /**< this DATA packet entry is used by a delivery complete LONGCTS send/write protocol*/
 
 /**
  * @enum for packet entry allocation type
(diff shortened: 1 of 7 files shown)


shijin-aws:
LGTM, some typos in the commit message:

>This patch addressed the issue by introdce a flag

`introdce` -> `introducing`

> The function rxr_pkt_handle_receipt_recv() was simplied to just write TX completion.

`simplied` -> `simplified`

prov/efa/src/rxr/rxr_pkt_entry.h
```diff
@@ -172,7 +173,8 @@ struct rxr_pkt_entry {
 	 * @brief flags indicating the status of the packet entry
 	 * 
 	 * @details
-	 * Possisle flags include  #RXR_PKT_ENTRY_IN_USE #RXR_PKT_ENTRY_RNR_RETRANSMIT and #RXR_PKT_ENTRY_LOCAL_READ
+	 * Possisle flags include  #RXR_PKT_ENTRY_IN_USE #RXR_PKT_ENTRY_RNR_RETRANSMIT,
```
lrbison, line 176: nit: could you correct this (pre-existing) typo: "Possisle" -> "Possible"

lrbison: Another nit: The commit message mentions two functions "was removed", should be "were removed".

LGTM

wzamazon: Thank you! fixed

wzamazon:
@shijin-aws  fixed in new revision

shijin-aws (approved):

lrbison (approved):
TY, LGTM

PR #8391, opened 2023-01-04
prov/efa: Remove fi_sendmsg/recvmsg but still use efa_ep's qp and av

sunkuamzn:
For SHM EP, use fi_sendv and fi_recvv
For EFA EP
* Allocate ibv_send_wr and ibv_recv_wr in rxr_pkt_entry
* Still use efa_ep->qp, efa_ep->av and efa_ep->xmit_more_wr*
* Directly call efa_post_flush or ibv_post_recv in rxr_pkt_entry.c

### DIFF
--- a/prov/efa/src/efa.h
+++ b/prov/efa/src/efa.h
@@ -231,16 +231,6 @@ struct efa_ep {
 	bool			util_ep_initialized;
 };
 
-struct efa_send_wr {
-	struct ibv_send_wr wr;
-	struct ibv_sge sge[];
-};
-
-struct efa_recv_wr {
-	struct ibv_recv_wr wr;
-	struct ibv_sge sge[];
-};
-
 struct efa_av {
 	struct fid_av		*shm_rdm_av;
 	fi_addr_t		shm_rdm_addr_map[EFA_SHM_MAX_AV_COUNT];
@@ -333,7 +323,7 @@ int efa_prov_initialize(void);
 
 void efa_prov_finalize(void);
 
-ssize_t efa_post_flush(struct efa_ep *ep, struct ibv_send_wr **bad_wr);
+ssize_t efa_post_flush(struct efa_ep *ep, struct ibv_send_wr **bad_wr, bool free);
 
 ssize_t efa_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count, fi_addr_t *src_addr);
 
--- a/prov/efa/src/efa_msg.c
+++ b/prov/efa/src/efa_msg.c
@@ -319,7 +319,7 @@ static void efa_post_send_sgl(struct efa_ep *ep, const struct fi_msg *msg,
 	}
 }
 
-ssize_t efa_post_flush(struct efa_ep *ep, struct ibv_send_wr **bad_wr)
+ssize_t efa_post_flush(struct efa_ep *ep, struct ibv_send_wr **bad_wr, bool free)
 {
 	ssize_t ret;
 
@@ -332,7 +332,10 @@ ssize_t efa_post_flush(struct efa_ep *ep, struct ibv_send_wr **bad_wr)
 #endif
 
 	ret = ibv_post_send(ep->qp->ibv_qp, ep->xmit_more_wr_head.next, bad_wr);
-	free_send_wr_list(ep->xmit_more_wr_head.next);
+	if (free)
+		free_send_wr_list(ep->xmit_more_wr_head.next);
+	else
+		ep->xmit_more_wr_head.next = NULL;
 	ep->xmit_more_wr_tail = &ep->xmit_more_wr_head;
 	return ret;
 }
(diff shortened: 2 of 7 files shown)


prov/efa/src/efa.h
```diff
@@ -231,16 +231,6 @@ struct efa_ep {
 	bool			util_ep_initialized;
 };
 
-struct efa_send_wr {
```
wzamazon, line 234: Why move the definitinion? it is still used by dgram endpoint, and should remain here.

prov/efa/src/rxr/rxr_cq.c
```diff
@@ -310,6 +310,9 @@ void rxr_cq_queue_rnr_pkt(struct rxr_ep *ep,
 #endif
 	dlist_insert_tail(&pkt_entry->entry, list);
 
+	// Remove other packets that were sent with the packet that encountered RNR
```
wzamazon, line 313: Should be `/* ... */`

The comment is a little confusing. maybe explain what `pkt_entry->send_wr.wr.next` is

prov/efa/src/efa_msg.c
```diff
@@ -286,7 +286,6 @@ static void efa_post_send_sgl(struct efa_ep *ep, const struct fi_msg *msg,
 	struct efa_mr *efa_mr;
 	struct ibv_send_wr *wr = &ewr->wr;
 	struct ibv_sge *sge;
-	size_t sgl_idx = 0;
```
wzamazon, line 289: This change is good. But probably should go into a separate commit

prov/efa/src/efa_msg.c
```diff
@@ -391,7 +392,7 @@ static ssize_t efa_post_send(struct efa_ep *ep, const struct fi_msg *msg, uint64
 	if (flags & FI_MORE)
 		return 0;
 
-	ret = efa_post_flush(ep, &bad_wr);
+	ret = efa_post_flush(ep, &bad_wr, true);
```
wzamazon, line 395: add `true /* free */` to indicate is meaning

prov/efa/src/rxr/rxr_pkt_entry.h
```diff
@@ -114,6 +114,17 @@ struct rxr_pkt_sendv {
  * is not registered (when it is unexpected or out-of-order). A new packet entry will be cloned
  * using endpoint's read_copy_pkt_pool, whose memory was registered.
  */
+
+struct efa_send_wr {
```
wzamazon, line 118: As mentioned before, this should remain in `efa.h` because it used by both rdm and dgram

wzamazon (changes requested):
With the change to `efa_send_wr` and `efa_recv_wr` you need to update  how `efa_ep->send_wr_bufpool` and `efa_ep->recv_wr_pool` is created. Specifically the entry size should be just `sizeof(efa_send_wr)` and `sizeof(efa_recv_wr)`

prov/efa/src/rxr/rxr_pkt_entry.h
sunkuamzn: The import hierarchy right now is that `efa.h` imports `rxr.h` and `rxr.h` does not import `efa.h`. So if we keep this in `efa.h`, we have to add `#import "efa.h"` to `rxr_pkt_entry.h` which leads to circular dependency issues and compilation failure

I can fix the import hierarchy in a later commit when we rename `efa` to `efa_dgram`, `rxr` to `efa_rdm` and move common code to common files.

sunkuamzn:
> With the change to `efa_send_wr` and `efa_recv_wr` you need to update how `efa_ep->send_wr_bufpool` and `efa_ep->recv_wr_pool` is created. Specifically the entry size should be just `sizeof(efa_send_wr)` and `sizeof(efa_recv_wr)`

But the DGRAM provider needs memory for the SGEs if a higher applications calls the DGRAM provider directly

prov/efa/src/efa_msg.c
sunkuamzn: ACK

prov/efa/src/efa_msg.c
sunkuamzn: done

prov/efa/src/rxr/rxr_pkt_entry.h
wzamazon: In that case, please add a todo comment here.

prov/efa/src/rxr/rxr_pkt_entry.h
```diff
@@ -114,6 +114,17 @@ struct rxr_pkt_sendv {
...
+	struct ibv_send_wr wr;
+	struct ibv_sge sge[2]; // EFA device supports a maximum of 2 iov/SGE
```
wzamazon, line 120: "//" should be "/**< ... */" for doxygen comment

sunkuamzn: done

prov/efa/src/rxr/rxr_pkt_entry.h
sunkuamzn: done

wzamazon (approved):

PR #8394, opened 2023-01-05
bsock: A couple of changes to the interface and related to async operations

sydidelot:
This PR implements a couple of changes to the bsock API in order to support async operations. 
These changes were initially implemented as part of #8262 but I think they deserve their own PR.

The most significant change is the conversion of the return type for the bsock interface. 
The functions now return a int (instead of a size_t) where a value of 0 means success and a value < 0 corresponds to a failure.
The number of bytes received and available for processing is returned in the input/output argument *len.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1300,22 +1300,23 @@ ssize_t ofi_bsock_sendv(struct ofi_bsock *bsock, const struct iovec *iov,
 	return ret;
 }
 
-ssize_t ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t len)
+int ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t *len)
 {
 	size_t bytes, avail;
 	ssize_t ret;
 
-	bytes = ofi_byteq_read(&bsock->rq, buf, len);
+	bytes = ofi_byteq_read(&bsock->rq, buf, *len);
 	if (bytes) {
-		if (bytes == len)
-			return len;
+		if (bytes == *len) {
+			return 0;
+		}
 
 		buf = (char *) buf + bytes;
-		len -= bytes;
+		*len -= bytes;
 	}
 
 	assert(!ofi_bsock_readable(bsock));
-	if (len < (bsock->rq.size >> 1)) {
+	if (*len < (bsock->rq.size >> 1)) {
 		avail = ofi_byteq_writeable(&bsock->rq);
 		assert(avail);
 		ret = bsock->sockapi->recv(bsock->sockapi, bsock->sock,
@@ -1326,43 +1327,50 @@ ssize_t ofi_bsock_recv(struct ofi_bsock *bsock, void *buf, size_t len)
 
 		ofi_byteq_add(&bsock->rq, (size_t) ret);
 		assert(ofi_bsock_readable(bsock));
-		bytes += ofi_byteq_read(&bsock->rq, buf, len);
-		return bytes;
+		bytes += ofi_byteq_read(&bsock->rq, buf, *len);
+		*len = bytes;
+		return 0;
 	}
 
-	ret = bsock->sockapi->recv(bsock->sockapi, bsock->sock, buf, len,
+	ret = bsock->sockapi->recv(bsock->sockapi, bsock->sock, buf, *len,
 				   MSG_NOSIGNAL, &bsock->rx_sockctx);
-	if (ret > 0)
-		return bytes + ret;
+	if (ret > 0) {
+		*len = bytes + ret;
+		return 0;
+	}
 
 out:
 	assert(ret != -OFI_EINPROGRESS_URING);
-	if (bytes)
-		return bytes;
+	*len = bytes;
+	if (*len)
+		return 0;
 	return ret ? -ofi_sockerr(): -FI_ENOTCONN;
 }
 
-ssize_t ofi_bsock_recvv(struct ofi_bsock *bsock, struct iovec *iov, size_t cnt)
+int ofi_bsock_recvv(struct ofi_bsock *bsock, struct iovec *iov, size_t cnt,
+		    size_t *len)
 {
-	size_t len, bytes, avail;
+	size_t bytes, avail;
 	ssize_t ret;
 
-	if (cnt == 1)
-		return ofi_bsock_recv(bsock, iov[0].iov_base, iov[0].iov_len);
+	if (cnt == 1) {
+		*len = iov[0].iov_len;
+		return ofi_bsock_recv(bsock, iov[0].iov_base, len);
+	}
 
-	len = ofi_total_iov_len(iov, cnt);
+	*len = ofi_total_iov_len(iov, cnt);
 	if (ofi_byteq_readable(&bsock->rq)) {
 		bytes = ofi_byteq_readv(&bsock->rq, iov, cnt, 0);
-		if (bytes == len)
-			return len;
+		if (bytes == *len)
+			return 0;
 
-		len -= bytes;
+		*len -= bytes;
 	} else {
 		bytes = 0;
 	}
 
 	assert(!ofi_bsock_readable(bsock));
-	if (len < (bsock->rq.size >> 1)) {
+	if (*len < (bsock->rq.size >> 1)) {
 		avail = ofi_byteq_writeable(&bsock->rq);
 		assert(avail);
 		ret = bsock->sockapi->recv(bsock->sockapi, bsock->sock,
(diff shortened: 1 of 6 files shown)


src/common.c
```diff
@@ -1196,22 +1196,27 @@ int ofi_bsock_flush_sync(struct ofi_bsock *bsock)
...
 		}
 
-		ret = ofi_bsock_flush(bsock);
-		if (ret)
-			return ret;
+		err = ofi_bsock_flush(bsock);
+		if (err) {
+			*len = 0;
+			return err;
+		}
```
shefty, line 1219: In the commit message, it mentions 'bytes received' rather than 'bytes sent'.  But from the code, I can't tell what len is supposed to be set to.  In the above flows, len is set to 0, but some data was actually queued on the bsock buffer.  Below, len is set to the number of bytes sent on the socket.

shefty (commented):
Thanks -- looks good overall, one concern on the value of len returned from bsock_send

src/common.c
sydidelot: I see what you mean. IMO, *len should always contain the number of bytes consumed by the bsock interface for the given call, whether the data is buffered in the bsock buffer or sent on the socket. 
In the case where `ofi_bsock_flush()` fails with an error different that -FI_EAGAIN, I would be tempted to rollback the changes to the bsock buffer (i.e., `byteq->tail -= *len`) and set `*len = 0`. If `ofi_bsock_flush()` returns `0` or -FI_EAGAIN, we cab consider that the function ofi_bsock_send() was successful () and return 0.

For example:
```
avail = ofi_bsock_tosend(bsock);                                    
if (avail) {                                                        
        if (*len < ofi_byteq_writeable(&bsock->sq)) {               
                ofi_byteq_write(&bsock->sq, buf, *len);             
                err = ofi_bsock_flush(bsock);                       
                if (err && err != -FI_EAGAIN) {                     
                        /* Add a call to rollback the changes done by ofi_byteq_write(): byteq->tail -= *len */
                        *len = 0;                                   
                        return err;                                 
                }                                                   
                return 0;                                           
        }                                                           
                                                                    
        err = ofi_bsock_flush(bsock);                               
        if (err) {                                                  
                *len = 0;                                           
                return err;                                         
        }                                                           
}                                                                   
```

But it seems to me that the problem existed because my change, no? Say if `ofi_bsock_flush()` fails with  an error different than -FI_EAGAIN. In this case the data is queued to the bsock buffer, but the caller doesn't know about it.

shefty: If flush fails with anything other the EAGAIN, the socket is in an error state, and there's no telling what may have made it to the other side.  I would not try to rollback or anything like that.  The transfer is doomed anyway.  I wouldn't even bother adjusting len based on that, though it should be harmless to set it to 0.

shefty (approved):
Thanks - the changes look good.  Waiting for CI to verify nothing was missed.

src/common.c
sydidelot: I updated the code with the following changes to ofi_bsock_send/ofi_bsock_sendv: 
- I clarified the usage of "*len" in the commit message.
- I fixed the error handling of ofi_bsock_flush. Here is how the code looks now:
```
if (*len < ofi_byteq_writeable(&bsock->sq)) {
        ofi_byteq_writev(&bsock->sq, iov, cnt);
        err = ofi_bsock_flush(bsock);
        return !err || err == -FI_EAGAIN ? 0 : err;
}
```

shefty:
An fi_multi_ep test on sockets failed on windows.  I haven't seen this error before, and it only failed on 1 of the runs.  I don't think any of these changes are related, but re-running.

sydidelot:
@shefty Thanks for the review! I have rebased the PR: there is no code change compared to previous submission.

sydidelot:
It failed again, but that could be a different issue. 
It doesn't seem that the `sockets` provider uses the bsock interface, I don't see how my changes could break that provider.

shefty:
Windows passed CI.  But, the tcp provider failed across all AWS configurations.  tcp also failed on Intel CI across several fabtests.  Here are a couple:

```
name:   fi_msg_pingpong -p "tcp"
01:19:07    timestamp: 20230106-091901+0000
01:19:07    result: Fail
01:19:07    time:   301
01:19:07    server_cmd:  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8394/4/reg/bin/fi_msg_pingpong -p "tcp"   -s ci5-eth2
01:19:07    server_stdout: |
01:19:07    client_cmd:  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8394/4/reg/bin/fi_msg_pingpong -p "tcp"   -s ci6-eth2 ci5-eth2
01:19:07    client_stdout: |
01:24:13  - name:   fi_msg_pingpong -v -p "tcp"
01:24:13    timestamp: 20230106-092402+0000
01:24:13    result: Fail
01:24:13    time:   301
...
name:   fi_msg_bw -p "tcp"
01:24:13    timestamp: 20230106-092404+0000
01:24:13    result: Fail
01:24:13    time:   1
01:24:13    server_cmd:  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8394/4/reg/bin/fi_msg_bw -p "tcp"   -s ci5-eth2
01:24:13    server_stdout: |
01:24:13      [error] fabtests:common/shared.c:2635: cq_readerr 265 (Truncation error), provider errno: 11 (Resource temporarily unavailable)
01:24:13    client_cmd:  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8394/4/reg/bin/fi_msg_bw -p "tcp"   -s ci6-eth2 ci5-eth2
01:24:13    client_stdout: |
01:24:13      [error] fabtests:common/shared.c:2635: cq_readerr 125 (Operation canceled), provider errno: 2 (No such file or directory)
```

I'm going to guess the return code changes to bsock send/recv may be off in the tcp provider somewhere.

sydidelot:
Ok, thanks. I will take a look later today. 
Is the net provider also tested by CI with the same fabtest config as the tcp provider?

shefty:
AWS does not test the net provider.  Intel CI does run fabtests (the tcp provider has more testing).  However, in this case, the Intel CI aborted after the tcp failures, so the net provider tests weren't run.

sydidelot:
I couldn't reproduce the issues with `runfabtests.sh tcp -vv` (I ran it multiple times):
```
# ------------------------------------------------------------------------------
# Total Pass                                                64
# Total Notrun/Excluded                                     21
# Total Fail                                                 2
# Percentage of Pass                                        96
# ------------------------------------------------------------------------------
```
2 tests always fail because of `Function not implemented`, which also happens without my changes.
I also reviewed the PR very carefully and everything looks good. I'm tempted to split the PR into multiple smaller PRs to bisect the commit that introduces the issue.

shefty:
I didn't notice any problems in the changes either, which makes me think there's somewhere that needs some change, which doesn't have it.  If you want to update the PR with 1 patch at a time until the CI breaks, that works for me.

sydidelot:
I updated the PR with only the first 2 first patches of the series:
- ofi_net: Cleanup of ofi_bsock_flush*() functions
- ofi_net: Adjust ofi_bsock_recv and ofi_bsock_recvv for async

shefty:
CI failures with just the changes for the 1st 2 patches are similar.  I'll try to investigate more.

src/common.c
```diff
@@ -1374,23 +1378,27 @@ ssize_t ofi_bsock_recvv(struct ofi_bsock *bsock, struct iovec *iov, size_t cnt)
...
-		return ret;
+	if (ret > 0) {
+		*len = ret;
+		return 0;
+	}
 out:
 	assert(ret != -OFI_EINPROGRESS_URING);
-	if (bytes)
-		return bytes;
+	*len = bytes;
```
shefty, line 1401: This change here and the same in the above function are the problem.  bytes can be > 0, so *len is set correctly, but now we fall through to the return call below, which can return a negative errno instead of 0.  We can take this path if we read data from the bsock buffer (bytes > 0), but still have more data to read.  That can call recv() at line 1372 to refill the bsock buffer, which can return EAGAIN.

sydidelot: Thank you very much for your help! I fixed the bug and resubmitted the first 2 patches. If they pass, I will resubmit the whole patch series.

sydidelot: I did some tests over the week-end and my conclusions are as follows:
- AppVeyor is always failing the jobs `Debug-EFA`, which is unrelated to my changes. Actually, all recent PRs fail the same tests.
- I tried a first version where I added the following return statement at the end of the function, right after the `out:` label (in both `ofi_bsock_recv` functions):
```
if (ret == -FI_EAGAIN)
    return 0;
```
With above changes, Intel and AWS both fail. 
- I updated the PR with the following change, which is more in line with the code before my changes:
```
if (*len)
    return 0;
```
AWS seems happy now, but Intel CI keeps failing. I leave the PR in this state in case you want to look at the failure in Intel CI.

shefty:
I don't think the Intel CI failures are related.  All DAOS tests failed with the tcp and verbs providers.  There's nothing in the output regarding the failures, so I'm going to guess it's a problem with the nodes.

shefty:
Confirmed that DAOS testing on the Intel CI is down and the result of the failure.

sydidelot:
Thanks for the feedback! I updated the PR with all the commits that you have originally reviewed.

shefty:
Thank you!  Cheers to having io_uring support now!  Are there any next steps?

PR #8406, opened 2023-01-10
prov/efa: Introduce efa_base_ep, change directory structure, clean up efa_ep and fix a bug in debug path

sunkuamzn:
This PR is for several changes

### Introduce `struct efa_base_ep`
This struct contains members common to both DGRAM and RDM endpoints and will be the first member in both endpoints. The RDM endpoint is still defined in `rxr.h`. The DGRAM endpoint is now defined in a new file called `efa_dgram.h`

### Change directory structure
The intended directory structure is that files related to DGRAM provider only should go in `prov/efa/src/dgram`. Files related to RDM provider only should go in `prov/efa/src/rdm`. Only files shared by both providers should remain in `prov/efa/src`.

This PR doesn't fully accomplish this change.
* I need to go over the files in `prov/efa/src` and separate out the code that is only used RDM/DGRAM and move that into their respective folders.
* `rxr_ep` still has `rdm_ep` as a member. So we still need to import `dgram/efa_dgram.h` in a few places in RDM code. The `rdm_ep` member and those imports will be removed in future commits.

### Remove `efa_ep->hmem_p2p_opt` ~~and `efa_ep->util_ep_initialized`~~
`hmem_p2p_opt` is only used in RDM code ~~and `util_ep_initialized` doesn't seem to be necessary~~

And a fix for a bug introduced in https://github.com/ofiwg/libfabric/pull/8391


a-szegel:
Can we change the name rxr to be rdm?  RXR seems legacy now.

sunkuamzn:
> Can we change the name rxr to be rdm? RXR seems legacy now. This can happen in a future patch.

Yes, we will do that in future commits. Don't want to put everything in one commit or PR.

prov/efa/src/dgram/efa_ep.c
```diff
@@ -237,11 +237,9 @@ static void efa_ep_destroy(struct efa_ep *ep)
 	efa_ep_destroy_qp(ep->base_ep.qp);
 	fi_freeinfo(ep->base_ep.info);
 	free(ep->base_ep.src_addr);
-	if (ep->util_ep_initialized) {
-		err = ofi_endpoint_close(&ep->base_ep.util_ep);
-		if (err)
-			FI_WARN(&efa_prov, FI_LOG_EP_CTRL, "Unable to close util EP\n");
-	}
+	err = ofi_endpoint_close(&ep->base_ep.util_ep);
```
wzamazon, line 240: You cannot do this.

calling `ofi_endpoint_close` on an util_ep that has not been initialized will result in segfault

sunkuamzn: Should we add a check on rxr_ep also then? https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rxr/rxr_ep.c#L795

wzamazon: Yes.
or can we make `rxr_ep` use the `util_ep` inside `efa_base_ep`?

sunkuamzn:
bot:aws:retest

prov/efa/src/dgram/efa_ep.c
sunkuamzn: `rxr_ep` is using the `util_ep` inside `efa_base_ep`. I will move `util_ep_initialized` to `efa_base_ep` and add the check in both endpoints

sunkuamzn: temporarily fixed by adding `util_ep_initialized` check to `rxr_ep`. Will eventually move this to `efa_base_ep_construct`

sunkuamzn:
Windows build must be failing because I didn't update `libfabric.vcxproj`. Will fix in next push.

wzamazon (approved):
Thank you! LGTM.

I think it would be better to add more information in the commit message of the commit that introduced `efa_base_ep`. In the commit, "rxr_ep" has `base_ep` in it, but `rxr_ep` did not actually use the `qp` inside `base_ep`. It is still using `qp` inside `rdm_ep`.

Next PR will get rid of `rdm_ep` inside `rxr_ep` and make it use `base_ep.qp`

wzamazon:
Looks like there are still conflict with `rxr_cq.c`

prov/efa/src/efa_base_ep.h
```diff
@@ -0,0 +1,85 @@
...
+#include "ofi_util.h"
+#include "ofi_file.h"
+
+struct efa_base_ep {
+	struct util_ep		util_ep;
+	struct efa_domain	*domain;
+	struct efa_qp		*qp;
+	struct efa_av		*av;
+	struct fi_info		*info;
+	size_t				rnr_retry;
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/efa/src/dgram/efa_dgram.h
```diff
@@ -0,0 +1,52 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "efa_base_ep.h"
+
+#ifndef EFA_DGRAM_H
+#define EFA_DGRAM_H
+
+struct efa_ep {
```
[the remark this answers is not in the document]
sunkuamzn: Yes, planning to do all the renaming in future commits

sunkuamzn:
> I think it would be better to add more information in the commit message of the commit that introduced efa_base_ep. 
Added

> Looks like there are still conflict with rxr_cq.c
Conflict was because of Darryl's commit which was merged this morning. But the rebase was clean, so it should be fine.

wzamazon:
Thanks! waiting for CI to finish

PR #8407, opened 2023-01-11
prov/efa: Fail unit test early if resource struct fails

sunkuamzn:
No point in continuing the test or checking the return value in the test body

### DIFF
--- a/prov/efa/test/efa_unit_test_common.c
+++ b/prov/efa/test/efa_unit_test_common.c
@@ -38,7 +38,7 @@ struct fi_info *efa_unit_test_alloc_hints(enum fi_ep_type ep_type)
 	return hints;
 }
 
-int efa_unit_test_resource_construct(struct efa_resource *resource, enum fi_ep_type ep_type)
+void efa_unit_test_resource_construct(struct efa_resource *resource, enum fi_ep_type ep_type)
 {
 	int ret = 0;
 	struct fi_av_attr av_attr = {0};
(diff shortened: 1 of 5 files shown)


prov/efa/test/efa_unit_test_common.c
```diff
@@ -87,11 +87,13 @@ int efa_unit_test_resource_construct(struct efa_resource *resource, enum fi_ep_t
 	if (ret)
 		goto err;
 
-	assert_int_equal(ret, 0);
```
sunkuamzn, line 90: No need to assert return value here because of the `goto` statements above

sunkuamzn:
In this PR, I only modified the cases where the return value from `efa_unit_test_resource_construct` was being checked. There are other cases where the return value was simply ignored. https://github.com/ofiwg/libfabric/search?q=efa_unit_test_resource_construct

It's in those cases that the early failure is helpful

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -23,8 +23,7 @@ void test_rxr_ep_pkt_pool_flags(struct efa_resource **state)
 	int ret;
```
wenduwan, line 23: Now this is used variable.

sunkuamzn: fixed

wenduwan (approved):
Thank you

lrbison (approved):

wzamazon (approved):

PR #8408, opened 2023-01-11
hmem_synpaseai: return 0 for host_register and host_deregister

wzamazon:
host pointer register and deregister are no-ops for synpase ai. return FI_SUCCESS for them

Signed-off-by: EC2 Default User <ec2-user@ip-172-31-58-230.us-west-2.compute.internal>

### DIFF
--- a/src/hmem_synapseai.c
+++ b/src/hmem_synapseai.c
@@ -122,12 +122,12 @@ int synapseai_close_handle(void *ipc_ptr)
 
 int synapseai_host_register(void *ptr, size_t size)
 {
-	return -FI_ENOSYS;
+	return FI_SUCCESS;
 }
 
 int synapseai_host_unregister(void *ptr)
 {
-	return -FI_ENOSYS;
+	return FI_SUCCESS;
 }
 
 int synapseai_get_base_addr(const void *ptr, void **base, size_t *size)


src/hmem_synapseai.c
```diff
@@ -206,12 +206,12 @@ int synapseai_close_handle(void *ipc_ptr)
...
 {
-	return -FI_ENOSYS;
+	return FI_SUCCESS;
 }
 
 int synapseai_host_unregister(void *ptr)
 {
-	return -FI_ENOSYS;
+	return FI_SUCCESS;
 }
```
lrbison, line 215: These are in the `else` of ``#if HAVE_SYNAPSEAI`, should they return -FI_ENOSYS to be consistent with an unsupported HMEM?

wzamazon: Yes. I think you are right. Updated the PR

lrbison (approved):
LGTM

lrbison (changes requested):
sign-off user name again :-(

lrbison (approved):
Sorry -- I see the commit is good, just the description was still bad

Approved.

shijin-aws (approved):
LGTM, what test will fail before this change?

wzamazon:
> LGTM, what test will fail before this change?

It impacted this PR https://github.com/ofiwg/libfabric/pull/8313

wzamazon:
@aingerson 

can you help with the "continuous-integration/jenkins/pr-head" failure?

aingerson:
@wzamazon Looks like a node issue. You can ignore

PR #8410, opened 2023-01-12
ofi_net: Push ofi_sockerr() down to the socket sockapi functions

sydidelot:
ofi_sockerr() is only relevant for sockets operations: calling it for uring sockapi functions would return an incoherent code.

Problem is that the uring sockapi functions may return error codes like -FI_EAGAIN (no credit available, SQE is in use) or -FI_EOVERFLOW (ring overflowed). If those error codes are returned, the bsock API would try to convert them with ofi_sockerr(), which is wrong.

The patch pushes the calls to ofi_sockerr() down to the socket sockapi functions.


shefty (approved):
Thanks!  Better structure.  Waiting on CI

PR #8413, opened 2023-01-13
fabtests/pytest/conftest.py: changes to running remote commands

wenduwan:
Neuron runtime runs `lspci` in shell. This is a problem in the current implementation, which runs the remote command in non-interactive mode, e.g. `ssh 1.2.3.4 lspci` produces `sh: lspci: command not found`, because the non-interactive shell has a different environment. The solution is to run the command in interactive mode, i.e. `ssh 1.2.3.4 bash --login -c lspci`.
  However, please note the side effect that we are now locked into `bash`.

### DIFF
--- a/fabtests/pytest/conftest.py
+++ b/fabtests/pytest/conftest.py
@@ -1,6 +1,7 @@
 import builtins
 import os
 import re
+import shlex
 
 import yaml
 
@@ -83,6 +84,8 @@ class CmdlineArgs:
         '''
             populate base command with informations in command line: provider, environments, etc
         '''
+        assert host_type in ("host", "server", "client")
+
         command = base_command
         # use binpath if specified
         if not (self.binpath is None):
@@ -90,8 +93,6 @@ class CmdlineArgs:
 
         if timeout is None:
             timeout = self.timeout
-
-        command = "timeout " + str(timeout) + " " + command
 
         # set environment variables if specified
         if not (self.environments is None):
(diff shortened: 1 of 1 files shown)


wzamazon (commented):
Why not just pass LD_LIBRARY_PATH through `-E` option?

wenduwan:
> Why not just pass LD_LIBRARY_PATH through `-E` option?

It doesn't actually work - dlopen happens before OOB exchange.
```
ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.53.150 FI_EFA_USE_DEVICE_RDMA=1 timeout 360 /home/ec2-user/libfabric/fabtests/install/bin/fi_rdm_pingpong -I 5 -v -S r:0,4,64 -D neuron -p efa -E LD_LIBRARY_PATH=/opt/aws/neuron/lib:$LD_LIBRARY_PATH
[error] fabtests:common/hmem_neuron.c:85: Failed to dlopen libnrt.so.1
```

wzamazon:
I meant to use the `-E` option of `runfabtests.py`. e.g.

`runfabtests.py -E "LD_LIBRARY_PATH=/opt/aws/neuron/lib FI_EFA_USE_DEVICE_RDMA=1"`

wenduwan:
> I meant to use the `-E` option of `runfabtests.py`. e.g.
> 
> `runfabtests.py -E "LD_LIBRARY_PATH=/opt/aws/neuron/lib FI_EFA_USE_DEVICE_RDMA=1"`

I see. That's better. Thanks!

fabtests/pytest/conftest.py
```diff
@@ -101,7 +101,7 @@ def populate_command(self, base_command, host_type, timeout=None):
             command = self._populate_ubertest_command(command, host_type)
         elif command.find("fi_multinode") != -1:
             command = self._populate_multinode_command(command, host_type)
-        else:
+        elif command.find("fi_") != -1:
```
wzamazon, line 105: Not sure this is necessary. Does it cause any harm?

wzamazon: There are some tests that does not start with `fi_`

wenduwan: I see. Thanks for pointing that out.

It's not a bug. I wanted to reuse `populate_command` to execute `check_hmem` which does not support the same fabtest parameters.

I guess that's fine. I don't have to use this. Will update the change.

wzamazon (approved):

PR #8417, opened 2023-01-14
prov/tcp: Merge net provider changes back into tcp provider

shefty:
This PR removes the tcp provider, then renames the net provider back as the tcp provider.  This merges the changes to the tcp provider which were made in the net provider fork back into the tcp provider.

The PR is being opened to check CI.

Since the updated tcp provider supports RDM endpoints directly, apps (including CI) that request *only* the tcp provider will no longer pick-up ofi_rxm automatically.  ofi_rxm;tcp still works and is wire compatible with previous libfabric releases, but ofi_rxm must be explicitly requested as a default.

@ooststep - These changes are derived from your branch, but updated to ensure git bisect works.

@sydidelot - FYI.  git should be able to handle the changes that you're working on, since the files are simply moved.  We did not rename internal functions or structures of the net provider (for example, changing the prefix from net back to tcp), and likely won't.


shefty:
@sydidelot - thanks - I rebased this PR after merging the fixes that you posted.  Do you use msg or rdm endpoints in your use case?

sydidelot:
@shefty it's `FI_EP_MSG`:

shefty:
CI failures: spawn2 hang (last patch was intended to address) + ubertest failure

```
Starting test 188 / 384:
14:54:33      [tcp, bandwidth, senddata--, FI_EP_RDM, FI_AV_UNSPEC, eq_wait_none, cq_wait_none, cntr_wait_none, comp_cntr -- tx (bind-FI_SELECTIVE_COMPLETION, op-NONE), rx: (bind-FI_SELECTIVE_COMPLETION, op-NONE), [FI_CQ_FORMAT_DATA],  FI_PROGRESS_MANUAL, FI_THREAD_SAFE (only hints), [], [], [FI_MSG, FI[error] fabtests:common/shared.c:2652: cq_readerr 125 (Operation canceled), provider errno: 107 (Transport endpoint is not connected)
14:54:33      ft_run_unit(): ubertest/test_ctrl.c:1095, ret=-125 (Operation canceled)
14:54:33      [error] fabtests:common/shared.c:2652: cq_readerr 104 (Connection reset by peer), provider errno: 104 (Connection reset by peer)
```
Previous failure with ubertest was a different test (285?).  This one generated more error information, though likely that other side crashed.  Haven't been able to reproduce this failure manually yet after a significant number of attempts.

shefty:
Updated - ubertest failures were caused by memory leaks in fabtests, which resulted in the kernel killing the process.  Unclear why the new provider triggers this more readily.  Will separate out fabtests changes into separate PR.

Spawn tests now appear to be working, and now the next test hangs, whatever that is:
```
17:45:08  bash -c source /home/cstbuild/intel/oneapi/mpi/latest/env/vars.sh -i_mpi_ofi_internal=0; export FI_PROVIDER=tcp; export I_MPI_FABRICS=ofi; export LD_LIBRARY_PATH=/home/cstbuild/intel/oneapi/mpi/latest/lib:$LD_LIBRARY_PATH; export LD_LIBRARY_PATH=/home/cstbuild/intel/oneapi/mpi/latest/lib/release:$LD_LIBRARY_PATH; export LD_LIBRARY_PATH=/home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8417/24/reg/lib/:$LD_LIBRARY_PATH; export PATH=/home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8417/24/reg/bin:$PATH; /home/cstbuild/intel/oneapi/mpi/latest/bin/mpiexec -n 1 -ppn 1 -hosts ci21-ib0,ci22-ib0 ./spawn1
17:45:08  None
17:45:08   No Errors
17:45:10  filename: 
17:45:10  bash -c source /home/cstbuild/intel/oneapi/mpi/latest/env/vars.sh -i_mpi_ofi_internal=0; export FI_PROVIDER=tcp; export I_MPI_FABRICS=ofi; export LD_LIBRARY_PATH=/home/cstbuild/intel/oneapi/mpi/latest/lib:$LD_LIBRARY_PATH; export LD_LIBRARY_PATH=/home/cstbuild/intel/oneapi/mpi/latest/lib/release:$LD_LIBRARY_PATH; export LD_LIBRARY_PATH=/home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8417/24/reg/lib/:$LD_LIBRARY_PATH; export PATH=/home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8417/24/reg/bin:$PATH; /home/cstbuild/intel/oneapi/mpi/latest/bin/mpiexec -n 1 -ppn 1 -hosts ci21-ib0,ci22-ib0 ./spawn2
17:45:10  None
17:45:10   No Errors
20:58:38  Sending interrupt signal to process
20:58:43  Sending interrupt signal to process
```
I interpret this as spawn2 passing, but the next one hanging?  No useful output beyond this that identifies the test that hung.

shefty:
@ooststep - The last patch disables mpich spawn tests and onecclgpu tests.  This allows CI to pass, so we can continue forward with the merge.  We'll need to debug and re-enable once fixed.

shefty:
@sydidelot - There have been a number of additional changes to this PR since you last tested.  The changes are focused on rdm endpoints, mostly around tagged message support, but obviously have the possibility of impacting msg endpoints, particularly when shared receive contexts are used.

sydidelot:
@shefty Thanks for the information! I will run some smoke tests with your PR next week. For your information, the code we are developing at DDN no longer uses shared receive queues.
By the way, I've seen several patches in your PR where a transfer can be completed after an EP is destroyed. Will it address the problem of canceling outstanding io_uring requests?

shefty:
I'm don't think it helps io_uring cancelation.

The issue is the sender transfers a message, waits for an ack, then exits, which closes the socket.  The ack is sent when the message is received, but MPI hasn't posted a buffer for it.  The provider detects the closed socket prior to the buffer being posted, triggering cleanup, including freeing all queued messages.

The endpoint and socket are still being destroyed when it is closed by the peer.  The changes keep already received messages around, with the assumption that the app is eventually going to request that data.  Any partially received messages still on the socket are still discarded.  Only acked messages end up being saved.

ooststep:
> I interpret this as spawn2 passing, but the next one hanging? No useful output beyond this that identifies the test that hung.

I think this is actually spawn2 hanging.  It appears rank 0 is not getting to the finalize stage.  The test is 'successful' (all ranks seem to send/recv expected messages), and is reported as such, but something is keeping the process from finalizing, leading to the hang.

shefty:
@ooststep - mpich tests passed.  oneCCL still hangs on the first test.  :(  The fix to my fix wasn't the fix there.

shefty:
@ooststep - build 28 passed CI.  Woohoo.  Needed to rebase to pull in changes to jenkins in order to merge though.

sydidelot:
@shefty I finally had a chance to test the latest `main` branch with your changes in a small setup: everything works well! One thing I noticed is a lock when the CQ is polled (libfabric is configured with `FI_THREAD_COMPLETION` and both `control_progress` and `data_progress` as set to data_progress).
This lock is initialized in `xnet_init_locks()`:
```
/* Because we may need to start the progress thread to support blocking CQ      
 * or EQ calls, we always need to enable an active lock, independent from       
 * the threading model requested by the app.                                    
 */                                                                             
static int xnet_init_locks(struct xnet_progress *progress, struct fi_info *info)
{                                                                               
    enum ofi_lock_type base_type, rdm_type;                                     
    int ret;                                                                    
                                                                                
    if (info && info->ep_attr && info->ep_attr->type == FI_EP_RDM) {            
        base_type = OFI_LOCK_NONE;                                              
        rdm_type = OFI_LOCK_MUTEX;                                              
        progress->active_lock = &progress->rdm_lock;                            
    } else {                                                                    
        base_type = OFI_LOCK_MUTEX;                                             
        rdm_type = OFI_LOCK_NONE;                                               
        progress->active_lock = &progress->lock;                                
    }                                                                           
                                                                                
    ret = ofi_genlock_init(&progress->lock, base_type);                         
```

The comment above the function states that the lock is needed if we start the progress thread. However, progress thread is not started if `attr->wait_obj == FI_WAIT_FD` and epoll is supported. I think we could disable the locking in this case.

On a related topic, I did some experiments with io_uring and it actually seems to work :) There are still a few improvements I would like to work on, especially regarding the polling and connection establishment. I will update #6761 with more details. 

Thank you!

shefty:
@sydidelot  - thanks!  Yes, that comment is stale now.  Subsequent changes removed some of those requirements.  I'll open an issue for this.

PR #8419, opened 2023-01-14
fabtests/pytest/common.py: retry client process separately in client server tests

wenduwan:
In client-server tests, SSH connection error will cause both the client and server processes to be terminated and retried. This is inefficient and problematic.
When the server startup is slow, e.g. > 1 second, OOB address exchange will fail due to SSH connection. In this case, restarting both the server and client does not solve the problem. Instead, we only need to retry the client process.

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -1,6 +1,7 @@
 import copy
 import errno
 import os
+import subprocess
 from subprocess import Popen, TimeoutExpired, run
 from tempfile import NamedTemporaryFile
 from time import sleep
@@ -8,7 +9,8 @@ from time import sleep
 import pytest
 from retrying import retry
 
-
+SERVER_RESTART_DELAY_MS = 10_1000
+CLIENT_RETRY_INTERVAL_MS = 1_000
 class SshConnectionError(Exception):
 
     def __init__(self):
@@ -128,6 +130,7 @@ def check_returncode_list(returncode_list, strict):
 
     if result == SKIP:
         pytest.skip(reason)
+
 
 class UnitTest:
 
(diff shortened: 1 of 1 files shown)


fabtests/pytest/common.py
```diff
@@ -287,65 +290,89 @@ def prepare_base_command(self, command_type, executable,
...
-        client_process = Popen(self._client_command + " > " + client_outfile + " 2>&1", shell=True)
 
-        server_timed_out = False
+        client_returncode = -1
         try:
-            server_process.wait(timeout=self._timeout)
-        except TimeoutExpired:
+            # Start client
+            # Retry on SSH connection error until server timeout
+            client_returncode = retry(
```
wzamazon, line 343: It seems you can put the `retry` decorator on `_run_client_command`, and call `self._run_client_command` directly here.

wenduwan: Yes -  the reason I put it here is because of the use of `self._timeout` variable. Instance variables are not accessible in the decorator.

wzamazon:
```
stop_max_delay=self._timeout * 1000,
```
Is this value necessary? what does it do?

wenduwan: In the new approach the client behavior is:
1. Check if the server is running. Terminate if not.
2. Try the client command.
3. In case of SSH error, start over again.
 
This all happens before we timeout the server, so we need to limit the max amount of time of retrying the client. `self._timeout` suits that purpose.

wzamazon: What does `stop_max_delay` do?

Also, it seems the unit is in millisecond, that is why you multiply timeout by 1000?

wenduwan: `stop_max_delay=T` stops the retry from `T` after the first attempt.

Yes - 1000 is the conversion from sec to ms.

wzamazon (approved):
LGTM

fabtests/pytest/common.py
wzamazon: Thanks! I think a comment about converting second to ms would be helpful here.

wenduwan: Thanks! Fixed in new push.

PR #8420, opened 2023-01-16
ofi_net: Fix wrong error handling for socket recv operations

sydidelot:
Patch 9360a1d70("ofi_net: Push ofi_sockerr() down to the socket sockapi functions") introduced a regression where the sockapi interface returns 0 instead of -FI_SHUTDOWN if the socket has shut down.


sydidelot:
@shefty sorry for the stupid bug! I found it while testing your PR #8417.

shefty:
Thanks - coverity actually reported on this, but I hadn't investigated it yet.

PR #8422, opened 2023-01-17
prov/net: A couple of hotfixes found while testing the provider

sydidelot:
The PR fixes 2 issues I have found while testing the provider:
- 1 memory leak when the EP is closed
- 1 annoying warn message that is printed for every EP that is closed.


shefty:
thanks!

PR #8428, opened 2023-01-20
fabtests/pytest/common.py: skip neuron to neuron tests on single node

wenduwan:
Because both client and server utilize all Neuron cores, they cannot request -D neuron on the same node. Thus skip those tests.

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -292,6 +292,9 @@ class ClientServerTest:
                 pytest.skip("no cuda device")
             elif host_memory_type == "neuron" and not has_neuron(host_ip):
                 pytest.skip("no neuron device")
+            elif (client_memory_type == server_memory_type == "neuron") and (
+                    self._cmdline_args.server_id == self._cmdline_args.client_id):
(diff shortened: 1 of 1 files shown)


fabtests/pytest/common.py
```diff
@@ -292,6 +292,9 @@ def prepare_base_command(self, command_type, executable,
                 pytest.skip("no cuda device")
             elif host_memory_type == "neuron" and not has_neuron(host_ip):
                 pytest.skip("no neuron device")
+            elif (client_memory_type == server_memory_type == "neuron") and (
+                    self._cmdline_args.server_id == self._cmdline_args.client_id):
+                pytest.skip("Neuron to Neuron tests require 2 nodes")
```
wzamazon, line 297: only "neuron to neuron" has this limitation?

wzamazon (approved):

fabtests/pytest/common.py
wenduwan: Right. The failure mode is that we cannot run 2 processes with `-D neuron` on the same node. The 1st process will occupy all Neuron cores, and the 2nd process will fail because there is no Neuron cores available.

PR #8429, opened 2023-01-21
include,src/hmem: Define missing neuron hmem ops function pointers

wenduwan:
Neuron hmem iface is missing ops function pointers, which caused segfault on trn1 hosts.
The function pointers are used in `ofi_hmem_host_register` by SHM provider etc.

This problem was not revealed in AWS CI because it does not test on trn1 hosts, thus the
Neuron code path was not activated.

### DIFF
--- a/src/hmem_neuron.c
+++ b/src/hmem_neuron.c
@@ -122,6 +122,16 @@ int neuron_copy_from_dev(uint64_t device, void *host, const void *dev, size_t si
 	return -FI_ENOSYS;
 }
 
+int neuron_host_register(void *ptr, size_t size)
+{
+	return FI_SUCCESS;
+}
+
+int neuron_host_unregister(void *ptr)
+{
+	return FI_SUCCESS;
+}
+
 int neuron_hmem_init(void)
 {
 	int ret;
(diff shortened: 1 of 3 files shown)


wzamazon (approved):

src/hmem_neuron.c
```diff
@@ -122,6 +122,41 @@ int neuron_copy_from_dev(uint64_t device, void *host, const void *dev, size_t si
 	return -FI_ENOSYS;
 }
 
+int neuron_get_handle(void *dev_buf, void **handle)
+{
+	return -FI_ENOSYS;
+}
```
shefty, line 128: ofi_hmem.h has ENOSYS implementations that can be used instead.  E.g. ofi_hmem_no_get_handle

wzamazon: The thing is that this function need to return 0 when there are neuron has been intialized. because shm calls it for all initialized hmem interfaces.

shefty: I'm referring to all of the functions here that return ENOSYS.  We have existing ENOSYS calls.

The functions should not return success unless it actually succeeds.  This sounds like a bug in the shm implementation

wenduwan: In the implementation
```c
```

wzamazon: Sorry. Just realized that shm called `host_register` which should be a no-op for neuron if neuron is initialized.

shm did not call `get_handle`. 

@wenduwan I think we should continue use `ofi_hmem_no_get_handle` here.

wenduwan: I see. Thanks I will open a separate PR to substitute those no op functions in both neuron and synapseai.

shefty (approved):

wzamazon (approved):

PR #8433, opened 2023-01-24
prov/hook/trace: Add trace log for CQ operation APIs.

peinanz:

### DIFF
--- a/prov/hook/trace/src/hook_trace.c
+++ b/prov/hook/trace/src/hook_trace.c
@@ -29,17 +29,15 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-
 #include "ofi_hook.h"
 #include "ofi_prov.h"
 #include "ofi_iov.h"
-
 
 #define TRACE_EP(ret, ep, buf, len, addr, tagkey, ignoredata, context) \
 	if (!(ret)) { \
 		FI_TRACE((ep)->domain->fabric->hprov, FI_LOG_EP_DATA, \
 			"buf %p len %zu src %zu tag/key 0x%zx " \
-			"ignore/data 0x%zx ctx %p", \
+			"ignore/data 0x%zx ctx %p\n", \
 			buf, len, addr, (uint64_t) tagkey, \
 			(uint64_t) ignoredata, context); \
 	}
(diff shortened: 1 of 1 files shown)


prov/hook/trace/src/hook_trace.c
```diff
@@ -48,6 +53,98 @@
...
+				}
+			}
+			index = strlen(cq_str);
+			snprintf(&cq_str[index], len-index, "}");
+			break;
+		default: // should not go here
+			return NULL;
+	}
+	return cq_str;
+}
```
shefty, line 146: We can reduce the overhead here (avoiding the extra string manipulation calls) by restructuring to:
```
#define TRACE_CQ(mycq, ctx, flags, len, buf, data, tag) \
    FI_TRACE(...blah blah...)

static void trace_cq_context(mycq, buf, count)
{
    struct fi_cq_entry *entry;
    size_t i;

    for (entry = buf, i = 0; i < count; i++) {
        TRACE_CQ(mycq, entry->context, entry->flags, 0, NULL, 0, 0)
        entry = entry + 1;
    }
}

static void trace_cq_msg(mycq, buf, count)
{
    struct fi_cq_msg_entry *entry;
    size_t i;

    for (entry = buf, i = 0; i < count; i++) {
        TRACE_CQ(mycq, entry->context, entry->flags, entry->len, NULL, 0, 0)
        entry = entry + 1;
    }
}

/* repeat for other formats */

static ssize_t trace_cq_read(...)
{
    ret = fi_cq_read(...);
    if (ret > 0)
        mycq->trace(mycq, buf, count);
}
```

We can set mycq->trace during cq creation to reference the correct function.  Although, since we're currently not intercepting cq create from this provider directly, you could also declare trace() as an array of function pointers and access that instead.  E.g.
```
static void (*trace)[FI_CQ_FORMAT_TAGGED] = {trace_cq_context, ...};

static ssize_t trace_cq_read(...)
{
    ...
    trace[mycq->format](mycq, buf, count);
}
```
You could add the flags checks directly to TRACE_CQ macro, or handle those in a common routine that then calls TRACE_CQ.

peinanz: If we call mycq->trace,  ==>  TRACE_CQ ==> FI_TRACE(), the <function: line of code> will point to the trace function for each format, rather than the original cq operations.  This may not be useful for debugging.

shefty: I'm also trying to limit the string manipulation (snprintf) calls.  We could replace FI_TRACE with direct calls to the logging functions.  FI_TRACE is a wrapper to FI_LOG, which is a macro that calls fi_log_enabled() and fi_log().  We could pass __func__ and __LINE__ into our calls and pass them through.  If fi_log_enabled() returns false, there's nothing we actually need to do anyway.

peinanz: When using fi_log, each call will print the <func:line>.  For an cq operation with more than 1 cq entries,  we may want to concatenate all entries into one printout in the log,  instead of one entry per printout with <func:line>.  Otherwise, it may be hard to tell whether the cq operation return multiple entries or multiple cq operations return a single entry. Will this an issue for debugging? To avoid string manipulation, we can add one fi_log_raw(level, format, ...), if fi_log_enabled(),  it can simply do as "fprintf" on the raw data, not including prefixes.

shefty: I would print 1 completion per line.  It will be easier to read, and the number of completions that an app can retrieve can be arbitrarily large.  We can have a separate print to display the number of completions found, though that's not a priority.

Don't use fprintf directly, as apps can override the fi_log functionality.

prov/hook/trace/src/hook_trace.c
```diff
@@ -48,6 +46,99 @@
 	TRACE_EP(ret, ep, count ? iov[0].iov_base : NULL, \
 		 ofi_total_iov_len(iov, count), addr, tagkey, ignoredata, context)
 
+typedef void (*trace_cq_entry_fn)(const struct fi_provider *prov,
+				const char *func, int line,
+				int count, const void *buf, uint64_t data);
+
+static void
+dummy(const struct fi_provider *prov, const char *func, int line,
+		int count, const void *buf, uint64_t data) { }
```
shefty, line 55: Please move braces to their own lines.  I'd rename to trace_cq_unknown() or noop.
Align second function line one space to the right of the '(', so under the const.  This comment applies to the other functions in this file.

prov/hook/trace/src/hook_trace.c
```diff
@@ -48,6 +46,99 @@
...
+
+static void
+trace_cq_context_entry(const struct fi_provider *prov, const char *func, int line,
+				int count, const void *buf, uint64_t data)
+{
+	struct fi_cq_entry *entry = (struct fi_cq_entry *)buf;
+	for (int i=0; i<count; i++) {
```
shefty, line 62: nit: Please add blank line after variable declaration and rest of code.  I *think* we can drop the case if entry is declared as const.  This comment applies to all functions.

second nit: Please add spaces before & after the math operations in the for loop.  e.g. ```for (i = 0; i < count; i++)```

prov/hook/trace/src/hook_trace.c
```diff
@@ -48,6 +46,99 @@
...
+		dummy,
+		trace_cq_context_entry,
+		trace_cq_msg_entry,
+		trace_cq_data_entry,
+		trace_cq_tagged_entry
+};
+
+static inline void
+trace_cq(struct hook_cq *cq, const char *func, int line,
+				int count, const void *buf, uint64_t data)
```
shefty, line 137: indentation of second line is off, please align under 'struct'

prov/hook/trace/src/hook_trace.c
```diff
@@ -48,6 +46,99 @@
...
+		}
+	}
+}
+
+static trace_cq_entry_fn trace_cq_entry[] = {
+		dummy,
+		trace_cq_context_entry,
+		trace_cq_msg_entry,
+		trace_cq_data_entry,
+		trace_cq_tagged_entry
```
shefty, line 132: indentation is off -- only indent 1 tab

prov/hook/trace/src/hook_trace.c
```diff
@@ -48,6 +46,99 @@
...
+};
+
+static inline void
+trace_cq(struct hook_cq *cq, const char *func, int line,
+				int count, const void *buf, uint64_t data)
+{
+	if ((count>0) && fi_log_enabled(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ))
+		trace_cq_entry[cq->format](cq->domain->fabric->hprov, func, line, count, buf, data);
```
shefty, line 140: line length should be limited to about 80 character

prov/hook/trace/src/hook_trace.c
```diff
@@ -702,6 +793,7 @@ trace_cq_readfrom_op(struct fid_cq *cq, void *buf, size_t count, fi_addr_t *src_
 	ssize_t ret;
 
 	ret = fi_cq_readfrom(mycq->hcq, buf, count, src_addr);
+	trace_cq(mycq, __func__, __LINE__, ret, buf, (src_addr)?(*src_addr):0);
```
shefty, line 796: Add spacing with src_addr: ```src_addr ? *src_addr : 0```

prov/hook/trace/src/hook_trace.c
```diff
@@ -724,6 +817,7 @@ trace_cq_sreadfrom_op(struct fid_cq *cq, void *buf, size_t count,
 	ssize_t ret;
 
 	ret = fi_cq_sreadfrom(mycq->hcq, buf, count, src_addr, cond, timeout);
+	trace_cq(mycq, __func__, __LINE__, ret, buf, (src_addr)?(*src_addr):0);
```
shefty, line 820: spacing with src_addr

shefty (changes requested):
See comments -- thanks, I think this change will be easier to maintain and parse the output.

PR #8438, opened 2023-01-25
prov/net,uring: cancel outstanding SQEs + various bugfixes 

sydidelot:
The PR covers the following changes:
- it implements the cancelation of outstanding SQEs when the EP is flushed.
- it fixes a bunch of bugs I found while testing the io_uring support.

With this PR, fabtests reports the same percentage of passed tests whether io_uring is enabled or not (command line `./runfabtests.sh net`).

There is still one test that randomly fails with io_uring, I'm not sure why:
```
$ FI_NET_IO_URING=1 fi_cq_data -e rdm -p "net"   -s 127.0.0.1
Waiting for CQ data from client
remote_cq_data: success
fi_cq_data: prov/net/src/xnet_progress.c:1490: xnet_close_progress: Assertion `slist_empty(&progress->event_list)' failed.
Aborted (core dumped)
```
It looks like the process receives the FI_SHUTDOWN event in the EQ but the application doesn't consume it.

### DIFF
--- a/prov/net/src/xnet_progress.c
+++ b/prov/net/src/xnet_progress.c
@@ -923,12 +923,12 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
 
 	assert(sockctx->uring_sqe_inuse);
 	sockctx->uring_sqe_inuse = false;
+	uring->sockapi->credits++;
 
 	if (sockctx->type == OFI_SOCKCTX_TX) {
 		tx_entry = ep->cur_tx.entry;
 		assert(tx_entry);
 
-		progress->sockapi.tx_uring.credits++;
 		if (cqe->res < 0) {
 			if (!OFI_SOCK_TRY_SND_RCV_AGAIN(-cqe->res))
 				xnet_complete_tx(ep, cqe->res);
@@ -945,7 +945,6 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
 	} else {
 		assert(sockctx->type == OFI_SOCKCTX_RX);
 
-		progress->sockapi.rx_uring.credits++;
 		if (bsock->async_prefetch) {
 			if (cqe->res > 0)
 				ofi_bsock_prefetch_done(bsock, cqe->res);
--- a/prov/net/src/xnet_progress.c
+++ b/prov/net/src/xnet_progress.c
@@ -1348,6 +1348,8 @@ static int xnet_init_uring(struct xnet_uring *uring, size_t entries,
 
 	uring->fid.fclass = XNET_CLASS_URING;
 	uring->sockapi = sockapi;
+	uring->sockapi->io_uring = &uring->ring;
+	uring->sockapi->credits = ofi_uring_sq_space_left(&uring->ring);
 
 	ret = ofi_dynpoll_add(dynpoll,
 			      ofi_uring_get_fd(&uring->ring),
@@ -1410,6 +1412,8 @@ int xnet_init_progress(struct xnet_progress *progress, struct fi_info *info)
 		goto err4;
 
 	if (xnet_io_uring) {
+		progress->sockapi = xnet_sockapi_uring;
+
 		ret = xnet_init_uring(&progress->tx_uring,
 				      info ? info->tx_attr->size :
 					     xnet_default_tx_size,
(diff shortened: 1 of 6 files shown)


shefty:
@sydidelot - rdm ep's manage the underlying connections internally.  So any FI_SHUTDOWN event is consumed by the rdm ep, which will use that to close the msg ep.

prov/net/src/xnet_progress.c
```diff
@@ -950,17 +950,22 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
 			else {
 				bsock->async_prefetch = false;
 				xnet_ep_disable(ep, 0, NULL, 0);
+				return;
 			}
 		} else if (cqe->res <= 0 && !OFI_SOCK_TRY_SND_RCV_AGAIN(-cqe->res)) {
 			if (ep->cur_rx.entry)
 				xnet_complete_rx(ep, cqe->res);
-			else
+			else {
```
shefty, line 957: nit: The coding standard is to add braces to both the if and else statements if either needs them.

Maybe we should use a goto disable in this call to jump to a common point.

prov/net/src/xnet_progress.c
```diff
@@ -1345,6 +1346,8 @@ static int xnet_init_uring(struct xnet_uring *uring, size_t entries,
 		return ret;
 
 	uring->fid.fclass = XNET_CLASS_URING;
+	uring->sockapi = sockapi;
+	uring->sockapi->credits = ofi_uring_sq_space_left(&uring->ring);
```
shefty, line 1378: This change is in the wrong patch, or should be in a new patch, or needs some description in the commit message regarding why it's needed.

prov/net/src/xnet_progress.c
```diff
@@ -1004,6 +1006,33 @@ static void xnet_progress_uring(struct xnet_progress *progress,
...
+	while (canceled_ctx->uring_sqe_inuse) {
+		if (!submitted) {
+			ret = ofi_sockctx_uring_cancel(uring->sockapi,
+						       canceled_ctx,
+						       ctx);
+			if (ret == -OFI_EINPROGRESS_URING) {
+				ofi_uring_submit(&uring->ring);
+				submitted = true;
+			}
+			else if (ret != -FI_EAGAIN)
```
shefty, line 1027: nit: move else to same line as } and add braces to else-if statement

shefty (commented):
minor quibbles -- looks good -- thanks!

prov/net/src/xnet_progress.c
sydidelot: Thanks. I updated the code to use a goto, as you suggested.

prov/net/src/xnet_progress.c
sydidelot: Thanks! I isolated this change in a new patch.

prov/net/src/xnet_progress.c
sydidelot: Fixed. Thanks!

sydidelot:
@shefty Thanks for the review! I have updated the PR. 
In regards to the assertion failure in fi_cq_data, I think it's a side-effect of asynchronous operations with io_uring. When the connection is closed, the provider needs to poll the io_uring to find out that the connection is effectively shut down and generate the FI_SHUTDOWN event. The problem is that the generation of the FI_SHUTDOWN event may happen after the endpoint gets closed.
I added a patch to the PR where `xnet_close_progress()` destroys all events listed in `progress->event_list`. Let me know if you have a better suggestion.

prov/net/src/xnet_progress.c
```diff
@@ -1479,10 +1479,17 @@ int xnet_init_progress(struct xnet_progress *progress, struct fi_info *info)
...
+
 	assert(dlist_empty(&progress->unexp_msg_list));
 	assert(dlist_empty(&progress->unexp_tag_list));
 	assert(dlist_empty(&progress->saved_tag_list));
-	assert(slist_empty(&progress->event_list));
+	while (!slist_empty(&progress->event_list)) {
+		item = slist_remove_head(&progress->event_list);
+		event = container_of(item, struct xnet_event, list_entry);
+		free(event);
+	}
```
shefty, line 1492: It might work to clean up the list for this test, but I'm concerned about the problem of having events that reference destroyed endpoints.  That could show up for other apps that have other active endpoints and don't close the progress engine.  They would read these events, try to process them, and likely access freed memory.

Maybe we need some sort of cleanup in the rdm ep close path for this instead.

shefty (commented):
I only see a problem with the last patch.  Given the commit message, it sounds like we can remove that and merge the other patches, since io_uring must be explicitly enabled.  We'll likely need to identify the reason the events aren't being cleaned up to see if the impact is only io_uring, or more general, with io_uring exposing some race easier.

prov/net/src/xnet_progress.c
sydidelot: Thanks! I have removed the last commit, as you suggested. 
I will address that particular issue in a new PR.

shefty:
The CI failures are related to GPU memory allocation, I think.  The GPU nodes might be down.  Your last update prior to popping of the topmost patch passed CI.

shefty (approved):

PR #8439, opened 2023-01-27
prov/trace: Add support for fi_cq_readerr

peinanz:


shefty:
Please pop the fi_cq_readerr patch off, rebase to main, and push back on.  The PR currently has a merge commit, which needs to be removed.

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,34 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
 	}
 }
 
+static inline void
+trace_cq_err(struct hook_cq *cq, const char *func, int line,
+             int count, struct fi_cq_err_entry *entry,  uint64_t flags)
```
shefty, line 155: please rename count to ret, since that's what's being passed

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,34 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
...
+{
+	if ((count > 0) &&
```
shefty, line 157: This if forces the entire function to be indented.  You could move the ```if (ret > 0)``` portion of the check to the caller (or keep it here)

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,34 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
...
+	    fi_log_enabled(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ)) {
```
shefty, line 158: For fi_log_enable(), if it is false, just return from the function.  That way the rest of the function doesn't have an extra indentation.

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,34 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
...
+		if (entry->flags & FI_RECV) {
+			fi_log(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ,
+			       func, line,
+			       "ctx %p flags 0x%lx, len %zu buf %p data %lu tag %lu "
+			       "olen %zu err_code %d prov_err [%s]\n",
+			       entry->op_context, entry->flags, entry->len, entry->buf,
+			       entry->data, entry->tag, entry->olen,  entry->err,
+			       fi_cq_strerror(cq->hcq, entry->prov_errno, entry->err_data,
+			                      NULL, 0));
```
shefty, line 167: It's okay to pass NULL into fi_cq_strerror, but the result means that we're not thread safe.  We'd be better off having a local ```char buf[80]``` variable, passing that to fi_cq_strerror(), and printing the results.  The buffer length is abitrary, but 80 should be sufficient.

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,34 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
...
+static inline void
+trace_cq_err(struct hook_cq *cq, const char *func, int line,
+             int count, struct fi_cq_err_entry *entry,  uint64_t flags)
+{
+	if ((count > 0) &&
+	    fi_log_enabled(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ)) {
+		if (entry->flags & FI_RECV) {
+			fi_log(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ,
+			       func, line,
+			       "ctx %p flags 0x%lx, len %zu buf %p data %lu tag %lu "
```
shefty, line 162: The tag is better printed in hex here and below.

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,36 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
...
 
+static inline void
+trace_cq_err(struct hook_cq *cq, const char *func, int line,
+             struct fi_cq_err_entry *entry,  uint64_t flags)
+{
+	char err_buf[80];
+
+	if (!fi_log_enabled(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ)) {
+		return;
+	}
```
shefty, line 161: nit: braces aren't needed.  (I wouldn't have bothered with this comment, only a separate update is needed.)

prov/hook/trace/src/hook_trace.c
```diff
@@ -150,6 +150,36 @@ trace_cq(struct hook_cq *cq, const char *func, int line,
...
+	if (entry->flags & FI_RECV) {
+		fi_log(cq->domain->fabric->hprov, FI_LOG_TRACE, FI_LOG_CQ,
+		       func, line,
+		       "ctx %p flags 0x%lx, len %zu buf %p data %lu tag %lu "
+		       "olen %zu err_code %d prov_err [%s]\n",
+		       entry->op_context, entry->flags, entry->len, entry->buf,
+		       entry->data, entry->tag, entry->olen,  entry->err,
+		       fi_cq_strerror(cq->hcq, entry->prov_errno, entry->err_data,
+		                      err_buf, 80));
```
shefty, line 170: You want to call fi_cq_strerror before and pass err_buf into the fi_log message.

PR #8440, opened 2023-01-27
prov/net: Fix memory leaks with RDM

sydidelot:
Running fabtests with ASAN has revealed 2 memory leaks with RDM. When a new endpoint (active or passive) is created, the fi_info object is duplicated and can be deleted by the caller.


shefty:
@wzamazon - upstream efa provider seems broken.  I've seen several CI errors on that provider.

shefty:
thanks!

PR #8443, opened 2023-01-30
prov/efa: remove one unnecessary copy and added a unit test

wzamazon:
fixed a bug that caused an unnecessary copy
added a unit test to verify it

### DIFF
--- a/prov/efa/src/rdm/rxr_pkt_type_base.c
+++ b/prov/efa/src/rdm/rxr_pkt_type_base.c
@@ -108,7 +108,8 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
 	int tx_iov_index;
 	char *data;
 	size_t tx_iov_offset, copied;
-	struct efa_mr *desc;
+	bool iov_accessible_by_device;
+	struct efa_mr *iov_mr;
 	int ret;
 
 	assert(pkt_data_offset > 0);
@@ -122,24 +123,37 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
 
 	rxr_locate_iov_pos(op_entry->iov, op_entry->iov_count, tx_data_offset,
 			   &tx_iov_index, &tx_iov_offset);
-	desc = op_entry->desc[0];
+	assert(tx_iov_index < op_entry->iov_count);
+	iov_mr = op_entry->desc[tx_iov_index];
 	assert(tx_iov_index < op_entry->iov_count);
 	assert(tx_iov_offset < op_entry->iov[tx_iov_index].iov_len);
 
-	ret = rxr_ep_use_p2p(ep, desc);
-	if (ret == 0)
-		goto copy;
+	iov_accessible_by_device = false;
+	if (pkt_entry->mr) {
+		/* When using EFA device, EFA device can access memory that
+		 * that has been registered, and p2p is allowed to be used.
+		 */
+		if (iov_mr) {
+			ret = rxr_ep_use_p2p(ep, iov_mr);
+			if (ret < 0)
+				return ret;
+			iov_accessible_by_device = ret;
+		} else {
+			iov_accessible_by_device = false;
+		}
+	} else {
+		/* When using shm, shm can access host memory without registration */
+		iov_accessible_by_device = !iov_mr || iov_mr->peer.iface == FI_HMEM_SYSTEM;
+	}
 
 	/*
 	 * Copy can be avoid if the following 2 conditions are true:
-	 * 1. user provided memory descriptor, or message is sent via shm provider
-	 *    (which does not require a memory descriptor)
+	 * 1. EFA/shm can directly access the memory
 	 * 2. data to be send is in 1 iov, because device only support 2 iov, and we use
 	 *    1st iov for header.
 	 */
-	if ((!pkt_entry->mr || op_entry->desc[tx_iov_index]) &&
+	if (iov_accessible_by_device &&
 	    (tx_iov_offset + data_size <= op_entry->iov[tx_iov_index].iov_len)) {
-
 		assert(ep->core_iov_limit >= 2);
 		pkt_entry->send.iov[0].iov_base = pkt_entry->wiredata;
 		pkt_entry->send.iov[0].iov_len = pkt_data_offset;
@@ -153,12 +167,11 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
 		return 0;
 	}
 
-copy:
 	data = pkt_entry->wiredata + pkt_data_offset;
 	copied = ofi_copy_from_hmem_iov(data,
 					data_size,
-					desc ? desc->peer.iface : FI_HMEM_SYSTEM,
-					desc ? desc->peer.device.reserved : 0,
+					iov_mr ? iov_mr->peer.iface : FI_HMEM_SYSTEM,
+					iov_mr ? iov_mr->peer.device.reserved : 0,
 					op_entry->iov,
 					op_entry->iov_count,
 					tx_data_offset);
--- a/prov/efa/test/efa_unit_test_ep.c
+++ b/prov/efa/test/efa_unit_test_ep.c
@@ -15,7 +15,7 @@ static void check_ep_pkt_pool_flags(struct efa_resource *resource, int expected_
 
 /**
  * @brief Test the pkt pool flags in rxr_ep_init()
- * 
+ *
  * @param[in]	state		struct efa_resource that is managed by the framework
  */
 void test_rxr_ep_pkt_pool_flags(struct efa_resource **state)
@@ -37,7 +37,7 @@ void test_rxr_ep_pkt_pool_flags(struct efa_resource **state)
 /**
  * @brief When the buf pool is created with OFI_BUFPOOL_NONSHARED,
  * test if the allocated memory is page aligned.
- * 
+ *
  * @param[in]	state		struct efa_resource that is managed by the framework
  */
 void test_rxr_ep_pkt_pool_page_alignment(struct efa_resource **state)
@@ -70,7 +70,7 @@ void test_rxr_ep_pkt_pool_page_alignment(struct efa_resource **state)
 /**
  * @brief when delivery complete atomic was used and handshake packet has not been received
  * verify there is no tx entry leak
- * 
+ *
  * @param[in]	state		struct efa_resource that is managed by the framework
  */
 void test_rxr_ep_dc_atomic_error_handling(struct efa_resource **state)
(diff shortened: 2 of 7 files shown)


prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -122,24 +123,35 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
 
 	rxr_locate_iov_pos(op_entry->iov, op_entry->iov_count, tx_data_offset,
 			   &tx_iov_index, &tx_iov_offset);
-	desc = op_entry->desc[0];
+	iov_mr = op_entry->desc[tx_iov_index];
```
wenduwan, line 126: Shouldn't this line be after L127?

Correct me if I'm wrong - first we should `assert(tx_iov_index < op_entry->iov_count);` before `op_entry->desc[tx_iov_index];`? Is it possible to segfault because of out of bound index?

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -108,7 +108,8 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
 	int tx_iov_index;
 	char *data;
 	size_t tx_iov_offset, copied;
-	struct efa_mr *desc;
+	bool iov_accessible_by_device;
```
wenduwan, line 111: nit: I think it's better to initialize it to `false` here in case it falls through and remain undefined.

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -122,24 +123,35 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
...
-	if (ret == 0)
-		goto copy;
+	if (pkt_entry->mr) {
+		/* When using EFA device, EFA device can access memory that
+		 * that has been registered, and p2p is allowed to be used.
+		 */
+		if (iov_mr) {
+			ret = rxr_ep_use_p2p(ep, iov_mr);
+			if (ret < 0)
+				return ret;
```
wenduwan, line 137: I think it's better(easier to debug) if we do `assert(ret >= 0)` instead of return.

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -122,24 +123,35 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
...
+			iov_accessible_by_device = ret;
+		} else {
+			iov_accessible_by_device = false;
+		}
+	} else {
+		/* When using shm, shm can access host memory without registration */
+		iov_accessible_by_device = !iov_mr || iov_mr->peer.iface == FI_HMEM_SYSTEM;
```
wenduwan, line 144: Could you explain why `iov_accessible_by_device` should  be `true` if `iov_mr == NULL`?

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -127,3 +127,40 @@ void test_rxr_ep_dc_atomic_error_handling(struct efa_resource **state)
 	/* make sure there is no leaking of tx_entry */
 	assert_true(dlist_empty(&rxr_ep->tx_entry_list));
 }
+
+/**
+ * @brief verify that when shm was used to send small message (<4k), no copy should be perform.
```
wenduwan, line 132: typo `Verify that when shm is used to send small messages (<4k), no copy should be performed.`

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -255,3 +266,11 @@ void *efa_mock_neuron_alloc_return_null(void **handle, size_t size)
...
 }
 #endif
+
+ssize_t __wrap_ofi_copy_from_hmem_iov(void *dest, size_t size,
+				      enum fi_hmem_iface hmem_iface, uint64_t device,
+				      const struct iovec *hmem_iov,
+				      size_t hmem_iov_count, uint64_t hmem_iov_offset)
+{
+	return g_efa_unit_test_mocks.ofi_copy_from_hmem_iov(dest, size, hmem_iface, device, hmem_iov, hmem_iov_count, hmem_iov_offset);
+}
```
wenduwan, line 276: Missing new line

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -127,3 +127,40 @@ void test_rxr_ep_dc_atomic_error_handling(struct efa_resource **state)
...
+ *
+ * @param[in]	state		struct efa_resource that is managed by the framework
+ */
+#define BUFF_LEN 8
```
wenduwan, line 136: Need to undefine afterwards

wenduwan: .. or maybe just use a normal variable.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -127,3 +127,40 @@ void test_rxr_ep_dc_atomic_error_handling(struct efa_resource **state)
...
+
+	/* create a fake peer */
+	err = fi_getname(&resource->ep->fid, &raw_addr, &raw_addr_len);
+	assert_int_equal(err, 0);
+	raw_addr.qpn = 1;
+	raw_addr.qkey = 0x1234;
+	num_addr = fi_av_insert(resource->av, &raw_addr, 1, &peer_addr, 0, NULL);
+	assert_int_equal(num_addr, 1);
+
+	g_ofi_copy_from_hmem_iov_call_counter = 0;
```
wenduwan, line 157: Move this to mock setup/teardown hook.

wenduwan: Also instead of adding a global counter, we can use cmocka mock utilities, i.e. 'will_return_count(efa_mock_ofi_copy_from_hmem_iov_inc_counter, NULL, 0)' - if it was called, cmock will throw an error
https://api.cmocka.org/group__cmocka__mock.html

wenduwan: where 
```c
ssize_t efa_mock_ofi_copy_from_hmem_iov_inc_counter(void *dest, size_t size,
						    enum fi_hmem_iface hmem_iface, uint64_t device,
						    const struct iovec *hmem_iov,
						    size_t hmem_iov_count, uint64_t hmem_iov_offset)
{
	mock();
        return 0;
}
```

wzamazon: I tried that, but `will_return_count` does not support 0.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: assertion is for something cannot happen. It can happen when that `ret < 0`. so we need to return and print an error message

prov/efa/test/efa_unit_test_ep.c
wenduwan: I see that's a shame.

Then why not even go simpler:

```c
ssize_t efa_mock_ofi_copy_from_hmem_iov_should_not_be_called (void *dest, size_t size,
						    enum fi_hmem_iface hmem_iface, uint64_t device,
						    const struct iovec *hmem_iov,
						    size_t hmem_iov_count, uint64_t hmem_iov_offset)
{
        /* This function should not be called */
        assert(1 == 0);
	return mock();
}
```

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: As comment said, shm can direct access memory without a  descriptor.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: added assertion in new revision.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: set to `false` in new revision.

prov/efa/test/efa_unit_test_ep.c
wzamazon: changed to an integer in new revision.

prov/efa/test/efa_unit_test_ep.c
wzamazon: > Then why not even go simpler:

I tried this too.

This will cause an exception to be throw, while `fi_send` is still holding a lock. 

Then the endpoint cannot be teared down.

wenduwan: This is a bit strange... if the test is written correctly this function should not have been invoked at all.

wenduwan (approved):

prov/efa/test/efa_unit_test_ep.c
wenduwan: > Move this to mock setup/teardown hook.

On this note - it is better than resetting the value manually in the test. The hook is guaranteed to run.

In general, if we use global variables and choose to set/reset its value, we need to make sure the logic always work - if the test fails for some reason and the value is not reset properly, it could cause side effects.

Now you mention that this variable is only used in this 1 test, it is probably ok. But imagine if it is used in multiple tests and it's not reset properly, we could get wrong test results due to the side effect.

wenduwan (approved):
Thanks! I will try this change on Neuron. Not sure if it is related or now.

prov/efa/src/rdm/rxr_pkt_type_base.c
shijin-aws: In what case shm path needs a copy? non-p2p FI_HMEM?

prov/efa/test/efa_unit_test_mocks.c
wzamazon: added

prov/efa/test/efa_unit_test_ep.c
wzamazon: The test is correct. The libfabric code has a bug which caused the copy function to be called.

wzamazon:
@shijin-aws  Somehow I did not see your comments. Can you comment again?

darrylabbate:
> @shijin-aws Somehow I did not see your comments. Can you comment again?

He commented here: https://github.com/ofiwg/libfabric/pull/8443#discussion_r1091160821

> It seems in old code, we skip copy anyway for shm. In this change it seems not. So in what case shm path needs a copy?

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: shm copy need a copy if buffer is on hmem. which means user provided a desc, and iface is not FI_HMEM_SYSTEM

shijin-aws: IIRC cuda to cuda via shm + ipc does not require a copy for tx, is this still needed?

shijin-aws:
@wzamazon  For some reason the conversation I commented in was resolved, I unresolved it so you should see it now.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: With current protocol selection, when this function is called, the memory is also on host memory.

Still, logically we should perform a copy for shm + HMEM, so I made the change.

wzamazon:
@shijin-aws see my recent reply

prov/efa/src/rdm/rxr_pkt_type_base.c
shijin-aws: I see, this function is not actually called in long read protocol. Currently for shm + cuda we are using long read only, so this copy actually will not happen. If we are going to use eager for shm + cuda for small msg, this will be called. Make sense.

shijin-aws (approved):

prov/efa/src/rdm/rxr_pkt_type_base.c
shijin-aws: `rxr_pkt_init_data_from_op_entry` is called by `rxr_pkt_init_rtm`. But `rxr_pkt_init_rtm` is only called by protocols that have data in the rtm packet. I think the name `rxr_pkt_init_rtm` is a bit misleading though because not all rtm really use it. But it can be a refactor in the future.

PR #8448, opened 2023-02-01
prov/util,shm: add FI_PEER support to util_cq and update shm to use util FI_PEER support

aingerson:
Moves FI_PEER support to util_cq so that any provider using the util_cq to read and write entries can leverage support to be either an owner or peer
Modifies providers using util_cq to streamline CQ functionality and isolate FI_SOURCE/signal optimizations to fewer places

### DIFF
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -514,14 +514,17 @@ struct util_cq {
 	struct ofi_genlock	cq_lock;
 	uint64_t		flags;
 
-	struct util_comp_cirq	*cirq;
-	fi_addr_t		*src;
-
-	struct slist		aux_queue;
-	fi_cq_read_func		read_entry;
 	int			internal_wait;
 	ofi_atomic32_t		wakeup;
 	ofi_cq_progress_func	progress;
+
+	struct fid_peer_cq	*peer_cq;
+
+	/* Only valid if not FI_PEER */
+	struct util_comp_cirq	*cirq;
+	fi_addr_t		*src;
+	struct slist		aux_queue;
+	fi_cq_read_func		read_entry;
 };
 
 int ofi_cq_init(const struct fi_provider *prov, struct fid_domain *domain,
@@ -611,14 +614,33 @@ ofi_cq_write_src(struct util_cq *cq, void *context, uint64_t flags, size_t len,
 	return ret;
 }
 
-int ofi_cq_insert_error(struct util_cq *cq,
-			const struct fi_cq_err_entry *err_entry);
 int ofi_cq_write_error(struct util_cq *cq,
 		       const struct fi_cq_err_entry *err_entry);
 int ofi_cq_write_error_peek(struct util_cq *cq, uint64_t tag, void *context);
 int ofi_cq_write_error_trunc(struct util_cq *cq, void *context, uint64_t flags,
 			     size_t len, void *buf, uint64_t data, uint64_t tag,
 			     size_t olen);
+
+static inline int
+ofi_peer_cq_write(struct util_cq *cq, void *context, uint64_t flags, size_t len,
+		  void *buf, uint64_t data, uint64_t tag, uint64_t src)
+{
+	return cq->peer_cq->owner_ops->write(cq->peer_cq, context, flags, len,
+					     buf, data, tag, src);
+}
+
+static inline int ofi_peer_cq_write_error(struct util_cq *cq,
+		const struct fi_cq_err_entry *err_entry)
+{
+	return cq->peer_cq->owner_ops->writeerr(cq->peer_cq, err_entry);
+}
+
+int ofi_peer_cq_write_error_peek(struct util_cq *cq, uint64_t tag,
+				 void *context);
+
+int ofi_peer_cq_write_error_trunc(struct util_cq *cq, void *context,
+				  uint64_t flags, size_t len, void *buf,
+				  uint64_t data, uint64_t tag, size_t olen);
 
 static inline int ofi_need_completion(uint64_t cq_flags, uint64_t op_flags)
 {
--- a/prov/util/src/util_cq.c
+++ b/prov/util/src/util_cq.c
@@ -42,7 +42,7 @@
 /* While the CQ is full, we continue to add new entries to the auxiliary
  * queue.
  */
-static void ofi_cq_insert_aux(struct util_cq *cq,
+static void util_cq_insert_aux(struct util_cq *cq,
 			      struct util_cq_aux_entry *entry)
 {
 	assert(ofi_genlock_held(&cq->cq_lock));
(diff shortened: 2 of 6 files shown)


include/ofi_util.h
```diff
@@ -504,6 +504,13 @@ OFI_DECLARE_CIRQUE(struct fi_cq_tagged_entry, util_comp_cirq);
 
 typedef void (*ofi_cq_progress_func)(struct util_cq *cq);
 
+struct util_cq_ctx {
+	struct util_comp_cirq	*cirq;
+	fi_addr_t		*src;
+	struct slist		aux_queue;
+	fi_cq_read_func		read_entry;
+};
```
shefty, line 512: The cirq and src are already pointers that we don't need to allocate.  Having this structure reduces the size of the util_cq by 3 pointers total and introduces a significant amount of changes.  We don't expect apps to allocate more than a few, probably only 1, CQ.

This patch contains a plethora of updates to the providers.  Is there no way to break those out?

aingerson: I'll rework it to keep the resources in the util_cq and conditionally allocate it. That should reduce the changes for the providers but it will still touch all the providers to update the cq write calls. I'll see what I can do. Stand by.

include/ofi_util.h
```diff
@@ -546,79 +551,52 @@ int ofi_cq_signal(struct fid_cq *cq_fid);
...
-			  size_t len, void *buf, uint64_t data, uint64_t tag,
-			  fi_addr_t src);
-
-
-static inline void
-ofi_cq_write_entry(struct util_cq *cq, void *context, uint64_t flags,
-		   size_t len, void *buf, uint64_t data, uint64_t tag)
+static inline int
+ofi_cq_write(struct util_cq *cq, void *context, uint64_t flags, size_t len,
+	     void *buf, uint64_t data, uint64_t tag, uint64_t src)
```
shefty, line 556: If you make this change up front, it may simplify the change set.  The util_cq has a read_entry function.  Adding a write_entry equivalent, based on FI_SOURCE, should be straightforward. That's essentially what peer_cq->owner_ops->write() is doing once the peer cq support ends up being added.

aingerson:
@shefty Updated to take a different approach. Removed all changes from other providers for more of an 'opt-in' method.

prov/util/src/util_cq.c
```diff
@@ -570,66 +510,242 @@ void ofi_cq_progress(struct util_cq *cq)
...
+	.ops_open = fi_no_ops_open,
+};
+
+static int util_init_peer_cq(struct util_cq *cq, struct fi_cq_attr *attr)
+{
+	cq->peer_cq = calloc(1, sizeof(*cq->peer_cq));
+	if (!cq->peer_cq)
+		return -FI_ENOMEM;
+
+
```
shefty, line 627: extra blank line.  I don't feel like I'm doing my job if I don't find *something* wrong.

prov/util/src/util_cq.c
```diff
@@ -570,66 +510,242 @@ void ofi_cq_progress(struct util_cq *cq)
...
+			return ret;
+		break;
+	case FI_WAIT_SET:
+		wait = attr->wait_set;
+		break;
+	default:
+		assert(0);
+		return ret;
+	}
+	if (wait) {
```
shefty, line 731: You can move the extra blank line above here.

See the comment in the original code that was removed about making sure the CQ is ready prior to adding it to the wait set.  As soon as the CQ is attached to the wait set, it can be called to drive progress by another thread.

prov/util/src/util_cq.c
```diff
@@ -570,66 +510,242 @@ void ofi_cq_progress(struct util_cq *cq)
...
+	case FI_WAIT_POLLFD:
+	case FI_WAIT_MUTEX_COND:
+	case FI_WAIT_YIELD:
+		memset(&wait_attr, 0, sizeof wait_attr);
+		wait_attr.wait_obj = attr->wait_obj;
+		cq->internal_wait = 1;
+		ret = fi_wait_open(&cq->domain->fabric->fabric_fid,
+				   &wait_attr, &wait);
+		if (ret)
+			return ret;
```
shefty, line 722: I'm not sure the cleanup in the error paths is correct.  We called ofi_genlock_init() above, but are exiting on error without calling ofi_genlock_destroy().  The prior code jumped to cleanup

shefty (commented):
Easier set of changes to examine.  Thanks

prov/util/src/util_cq.c
aingerson: I saw that comment and removed it because I thought it didn't make sense what "fully operational" meant since the code created the queue and source array after it was added to the wait set. Does this mean the old code was wrong? Do you want the wait set to be the last thing it does?

shefty: Yes, adding to the wait set should be the last thing.  Doing that technically exposing the CQ to the user, regardless of what state it's in.

aingerson:
@shefty Updated to fix the cleanup and init paths. You were right; the error paths were all off. Sorry about that.
Also added the shortcut truncation and peek error paths.

prov/util/src/util_cq.c
```diff
@@ -570,68 +549,265 @@ void ofi_cq_progress(struct util_cq *cq)
...
+			}
 			goto cleanup;
 		}
 	}
 	return 0;
-
+destroy2:
+	ofi_genlock_destroy(&cq->cq_lock);
+destroy1:
+	ofi_mutex_destroy(&cq->ep_list_lock);
```
shefty, line 810: The cleanup order with destroy1 and destroy2 look correct.  But the jump to cleanup when an error occurs even further into the function (that is, we skip the destroy cleanups), looks wrong.  I guess I need to go see what ofi_cq_cleanup() actually does relative to the above changes.

aingerson: I think the skipping the destroys is fine but I think the fallthroughs from the destroys are not right. Gah I'll tweak it again. Sorry

aingerson: Updated again to fix cleanup path. I think it should be good this time 🤞

shefty (approved):

shefty:
Intel CI failure is a new sighting with the net provider.  Captured in an issue.

aingerson:
Latest changes should fix #8453 8453

aingerson:
@shefty Latest works with the fi_cq_test but failed ubertest with some disconnect issues which have happened in other PRs so I think they are unrelated. Do you know what is up with those failures?

shefty:
Stephen and I are debugging net provider failures with ubertest.  So far it looks like it's the result of memory leaks and the kernel is killing off the process.  We have a valgrind log that we're analyzing, and the problem looks mostly related to fabtests leaking hmem bounce buffers.  I'm working on a fix for some of that now.

aingerson:
@shefty Ok to merge?

PR #8451, opened 2023-02-01
prov/efa: disable shm when HMEM is requrired and cuda is not allowed

wzamazon:
shm provider must make cuda calls to transfer cuda memory.

Therefore this patch disable shm usage when application require FI_HMEM capability but cuda is not available.

### DIFF
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -2071,6 +2071,18 @@ bool rxr_ep_use_shm_for_tx(struct fi_info *info)
 	    && !(info->caps & FI_LOCAL_COMM))
 		return 0;
 
+	/*
+	 * shm provider must make cuda calls to transfer cuda memory.
+	 * if cuda call is not allowed, we cannot use shm for transfer.
+	 *
+	 * Note that the other two hmem interfaces supported by EFA,
+	 * AWS Neuron and Habana Synapse, have no SHM provider
+	 * support anyways, so disabling SHM will not impact them.
+	 */
+	if (info && (info->caps & FI_HMEM) &&
+	    cuda_get_xfer_setting() == CUDA_XFER_DISABLED)
+		return 0;
+
 	return rxr_env.enable_shm_transfer;
 }
 
--- a/prov/efa/src/rdm/rxr_pkt_type_base.c
+++ b/prov/efa/src/rdm/rxr_pkt_type_base.c
@@ -377,31 +377,51 @@ int rxr_pkt_copy_data_to_cuda(struct rxr_ep *ep,
 	static const int max_blocking_copy_rx_entry_num = 4;
 	struct rxr_op_entry *rx_entry;
 	struct efa_mr *desc;
-	bool p2p_available, blocking_copy_available;
+	bool p2p_available, local_read_available, gdrcopy_available, cuda_memcpy_available;
 	int ret, err;
 
 	rx_entry = pkt_entry->x_entry;
 	desc = rx_entry->desc[0];
 	assert(efa_mr_is_cuda(desc));
 
-	blocking_copy_available = cuda_is_gdrcopy_enabled() || (cuda_get_xfer_setting() == CUDA_XFER_ENABLED);
+	local_read_available = false;
+	gdrcopy_available = false;
+	cuda_memcpy_available = false;
+
+	if (cuda_get_xfer_setting() == CUDA_XFER_ENABLED) {
+		cuda_memcpy_available = true;
+	} else {
+		/**
+		 * When CUDA API is available, EFA provider will not use gdrcopy even if gdrcopy is enabled.
+		 * see #efa_mr_use_gdrcopy
+		 */
+		gdrcopy_available = cuda_is_gdrcopy_enabled();
+	}
 
 	ret = rxr_ep_use_p2p(ep, desc);
 	if (ret < 0)
 		return ret;
 
 	p2p_available = ret;
+	local_read_available = p2p_available && efa_domain_support_rdma_read(rxr_ep_domain(ep));
 
-	if (!blocking_copy_available && !p2p_available) {
-		EFA_WARN(FI_LOG_CQ, "None of the copy methods: p2p, gdrcopy or cudaMemcpy is available,"
+	if (!local_read_available && !gdrcopy_available && !cuda_memcpy_available) {
+		EFA_WARN(FI_LOG_CQ, "None of the copy methods: localread, gdrcopy or cudaMemcpy is available,"
 			"thus libfabric is not able to copy received data to Nvidia GPU");
 		return -FI_EINVAL;
 	}
 
-	if (blocking_copy_available && !p2p_available)
+	if (!local_read_available) {
+		assert(cuda_memcpy_available || gdrcopy_available);
 		return rxr_pkt_queued_copy_data_to_hmem(ep, pkt_entry, data, data_size, data_offset);
+	}
 
-	if (!blocking_copy_available && p2p_available) {
+	assert(local_read_available);
+
+	if (!gdrcopy_available) {
+		/* prefer local read over cudaMemcpy (when it is available)
+		 * because local read copy is faster
+		 */
 		err = rxr_read_post_local_read_or_queue(ep, rx_entry, data_offset,
 							pkt_entry, data, data_size);
 		if (err)
@@ -409,7 +429,10 @@ int rxr_pkt_copy_data_to_cuda(struct rxr_ep *ep,
 		return err;
 	}
 
-	/* when both p2p and blocking copy are available, we use a mixed approach */
+	assert(gdrcopy_available && local_read_available);
+
+	/* when both local read and gdrcopy are available, we use a mixed approach */
+
 	if (rx_entry->cuda_copy_method != RXR_CUDA_COPY_LOCALREAD) {
 		assert(rx_entry->bytes_copied + data_size <= rx_entry->total_len);
 


prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2197,6 +2197,17 @@ bool rxr_ep_use_shm_for_tx(struct fi_info *info)
...
 
+	/*
+	 * shm provider must make cuda calls to transfer cuda memory.
+	 * if cuda call is not allowed, we cannot use shm for transfer.
+	 * Note that, the other 2 hmem interfaces supported by EFA:
+	 * AWS Neuron and Habans synapse, will not pass intra-node traffic
+	 * to EFA (and shm provider does not support them either),
+	 * therefore disable shm will not impact them.
+	 */
+	if (info && (info->caps & FI_HMEM) && cuda_get_xfer_setting()!= CUDA_XFER_ENABLED)
```
wenduwan, line 2208: nit missing space `cuda_get_xfer_setting()!<space>= CUDA_XFER_ENABLED`

wenduwan (approved):

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2197,6 +2197,17 @@ bool rxr_ep_use_shm_for_tx(struct fi_info *info)
 	    && !(info->caps & FI_LOCAL_COMM))
 		return 0;
 
+	/*
+	 * shm provider must make cuda calls to transfer cuda memory.
+	 * if cuda call is not allowed, we cannot use shm for transfer.
+	 * Note that, the other 2 hmem interfaces supported by EFA:
+	 * AWS Neuron and Habans synapse, will not pass intra-node traffic
+	 * to EFA (and shm provider does not support them either),
```
lrbison, line 2205: I think this comment is misleading.

I think the change is good for CUDA, but perhaps we should limit it to only when CUDA is selected, and leave a comment about maybe disabling it for other HMEM types as well.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2197,6 +2197,17 @@ bool rxr_ep_use_shm_for_tx(struct fi_info *info)
 	    && !(info->caps & FI_LOCAL_COMM))
 		return 0;
 
+	/*
+	 * shm provider must make cuda calls to transfer cuda memory.
+	 * if cuda call is not allowed, we cannot use shm for transfer.
+	 * Note that, the other 2 hmem interfaces supported by EFA:
+	 * AWS Neuron and Habans synapse, will not pass intra-node traffic
```
darrylabbate, line 2204: typo: `Habans`

prov/efa/src/rdm/rxr_ep.c
wzamazon: > I think the change is good for CUDA, but perhaps we should limit it to only when CUDA is selected

In the info object, there is no information about which type of interface the application is going to use.

prov/efa/src/rdm/rxr_ep.c
wzamazon: fixed in new revision

prov/efa/src/rdm/rxr_ep.c
wzamazon: fixed in new revision

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -377,39 +377,59 @@ int rxr_pkt_copy_data_to_cuda(struct rxr_ep *ep,
...
-	blocking_copy_available = cuda_is_gdrcopy_enabled() || (cuda_get_xfer_setting() == CUDA_XFER_ENABLED);
+	local_read_available = false;
+	gdrcopy_available = false;
+	cuda_memcpy_available = false;
+
+	if (cuda_get_xfer_setting() == CUDA_XFER_ENABLED) {
+		cuda_memcpy_available = true;
+	} else {
+		/* gdrcopy is not enabled when CUDA API is allowed to be used */
+		gdrcopy_available = cuda_is_gdrcopy_enabled();
```
wenduwan, line 395: This `if... else...` block is dubious to me.

Intuitively `gdrcopy_available` should always be `cuda_is_gdrcopy_enabled()`. The `if... else...` implies a mutual exclusion, i.e. `cuda_memcpy_available` and `gdrcopy_available` cannot both be true.

Symmetry would be better IMO, i.e. extract a new function `cuda_is_memcpy_enabled()`, then replace
```c
        if (cuda_get_xfer_setting() == CUDA_XFER_ENABLED) {
		cuda_memcpy_available = true;
	} else {
		/* gdrcopy is not enabled when CUDA API is allowed to be used */
		gdrcopy_available = cuda_is_gdrcopy_enabled();
        {
```
with 
```c
        cuda_memcpy_available = cuda_is_memcpy_enabled();
        gdrcopy_available = cuda_is_gdrcopy_enabled();
```

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -377,39 +377,59 @@ int rxr_pkt_copy_data_to_cuda(struct rxr_ep *ep,
...
-		EFA_WARN(FI_LOG_CQ, "None of the copy methods: p2p, gdrcopy or cudaMemcpy is available,"
+	if (!local_read_available && !gdrcopy_available && !cuda_memcpy_available) {
+		EFA_WARN(FI_LOG_CQ, "None of the copy methods: localread, gdrcopy or cudaMemcpy is available,"
 			"thus libfabric is not able to copy received data to Nvidia GPU");
 		return -FI_EINVAL;
 	}
 
-	if (blocking_copy_available && !p2p_available)
+	if (!local_read_available) {
+		assert(cuda_memcpy_available || gdrcopy_available);
```
wenduwan, line 412: Unnecessary assert because L405 already covers the case - similarly L416, L429 can be removed.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: Your understanding is correct. `cuda_memcpy_available` and `gdrcopy_available` are mutually exclusive. As the comment stated "gdrcopy is not enabled when CUDA API is allowed to be used".

Maybe I should rephrase to that "When CUDA API is available, efa provider will not use gdrcopy even if it is enabled". (Would that be move clear to you)?

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: This reminds the both reader/writer of current status. I found them to be very helpful when reading code. I would prefer to keep them

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -377,39 +377,59 @@ int rxr_pkt_copy_data_to_cuda(struct rxr_ep *ep,
...
 		return rxr_pkt_queued_copy_data_to_hmem(ep, pkt_entry, data, data_size, data_offset);
+	}
 
-	if (!blocking_copy_available && p2p_available) {
+	assert(local_read_available);
+
+	if (!gdrcopy_available) {
+		/* prefer local read over cudaMemcpy (when it is availbale)
```
lrbison, line 419: nit: `available`

lrbison: These decisions about performance... do they apply to all size messages?

prov/efa/src/rdm/rxr_pkt_type_base.c
wenduwan: Yes I think the new comment is more clear and describes the actual intention.

wzamazon: Fixed in new revision.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2071,6 +2071,18 @@ bool rxr_ep_use_shm_for_tx(struct fi_info *info)
 	    && !(info->caps & FI_LOCAL_COMM))
 		return 0;
 
+	/*
+	 * shm provider must make cuda calls to transfer cuda memory.
+	 * if cuda call is not allowed, we cannot use shm for transfer.
+	 * Note that, the other 2 hmem interfaces supported by EFA:
+	 * AWS Neuron and Habana synapse, will not pass intra-node traffic
+	 * to EFA (and shm provider does not support them either),
+	 * therefore disable shm will not impact them.
```
lrbison, line 2080: how about:

```
Note that the other two hmem interfaces supported by EFA,
AWS Neuron and Habana Synapse, have no SHM provider
support anyways, so disabling SHM will not impact them.
```

prov/efa/src/rdm/rxr_pkt_type_base.c
lrbison: I think the difference is "available" vs "preferred".  I think GDR and CudaMemCopy can both be available, but the logic here is trying to decide which is preferred.

wzamazon: >  I think GDR and CudaMemCopy can both be available

No they cannot, because gdrcopy require register memory with the gdrdrv kernel module, and we do not register memory with gdrcopy when cuda API is available to be used. See. https://github.com/ofiwg/libfabric/blob/76c449a93e482e8c2bd2ab7a85c29c6067bc9d75/prov/efa/src/efa_mr.c#L213 and https://github.com/ofiwg/libfabric/blob/76c449a93e482e8c2bd2ab7a85c29c6067bc9d75/prov/efa/src/efa_mr.c#L520

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: The maximum data size a packet can hold is ~8k, this statement is true for such message size range.

prov/efa/src/rdm/rxr_ep.c
wzamazon: Thanks! used in new revision.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: In new revision, I added a link to the function `#efa_mr_use_gdrcopy` for reader's reference.

lrbison (approved):

wenduwan (approved):

wzamazon:
bot:aws:retest

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2071,6 +2071,18 @@ bool rxr_ep_use_shm_for_tx(struct fi_info *info)
...
+	/*
+	 * shm provider must make cuda calls to transfer cuda memory.
+	 * if cuda call is not allowed, we cannot use shm for transfer.
+	 *
+	 * Note that the other two hmem interfaces supported by EFA,
+	 * AWS Neuron and Habana Synapse, have no SHM provider
+	 * support anyways, so disabling SHM will not impact them.
+	 */
+	if (info && (info->caps & FI_HMEM) &&
+	    cuda_get_xfer_setting() == CUDA_XFER_DISABLED)
```
lrbison, line 2083: I thought that looked fishy... good catch!

lrbison (approved):

PR #8456, opened 2023-02-04
prov/util, shm: add util_srx implementation and move shm to use the generic srx code

aingerson:


include/ofi_util.h
```diff
@@ -1156,7 +1157,54 @@ enum {
...
+	uint64_t		rx_op_flags;
+	uint64_t		rx_msg_flags;
+	size_t			iov_limit;
+
+	struct util_cq		*cq;
+	struct util_queue	unexp_msg_queue;
+	struct util_queue	unexp_tagged_queue;
+	struct ofi_bufpool	*rx_pool;
+	ofi_spin_t		lock;
+};
```
shefty, line 1186: Since this is meant to be generic, here's a comparison with tcp/net:
```
struct xnet_srx {
	struct fid_ep		rx_fid;
	struct xnet_domain	*domain;
	struct slist		rx_queue;
	struct slist		tag_queue;
	struct ofi_dyn_arr	src_tag_queues;
	struct ofi_dyn_arr	saved_msgs;

	struct xnet_xfer_entry	*(*match_tag_rx)(struct xnet_srx *srx,
						 struct xnet_ep *ep,
						 uint64_t tag);

	uint64_t		tag_seq_no;
	struct ofi_bufpool	*buf_pool;
	uint64_t		op_flags;
	size_t			min_multi_recv_size;

	/* Internal use when srx is part of rdm endpoint */
	struct xnet_rdm		*rdm;
	struct xnet_cq		*cq;
	struct util_cntr	*cntr;
};
```
It's worth noting that tcp needs to support srx with msg ep's.  The tcp msg queue (rx_queue) is based on slist not dlist.  dlist insertion/removals are expensive vs slist.  (dlist is useful when you need to remove an item from the middle of a list *and* you're not traversing the list at the time.)  For tagged msgs, it maintains index-based (src addr) queues (src_tagged_queues, and saved_msgs).  There isn't an rx_pool associated with the srx.

No additional locking is used by srx, as a progress level lock is held instead.  That lock is a mutex (genlock actually, as it can be a no-op), since the calls made while holding it can block.  Even the spinlock above is held during malloc calls, which can also block.  We need to figure out how locking works across providers.  In any case, if we tried to use this with tcp, an srx level lock doesn't quite work.  Internally, the correct lock is already held.  Only an external caller would need to acquire the right lock.

The unexpected message queue ends up being an issue when messages are received prior to the peer having been inserted into the local AV.  I don't know how this code handles that situation.

aingerson: The idea of this implementation (for now) is to have a simple, non-optimized srx. My goal is to change the dlist to an slist and add the src address unexpected messaging optimization that tcp/net has. But for now, this implementation matches what shm uses which is why I've only updated shm to use this implementation. But the goal is improve this generic implementation to support other providers without a loss of optimizations/support.

I think we definitely need to figure out locking across providers. There will have to be some sort of resource control, obviously. Perhaps we could use the new atomic queue for a lockless approach? The limitation there would be that the pool would not be dynamic, but if we stick to a specific rx_size then probably would be ok? (I haven't thought that out at all, just spit balling.)

As for dealing with unexpected messages before the address is inserted into the AV, this implementation does not take that into account at all so I'll definitely need to add that. How does tcp fix that?

shefty: TCP is selective about which messages it wants to save.  If the receiving endpoint doesn't have the peer in the local AV, it leaves the message unprocessed until later.  Basically, the srx can't save just any unexpected message that an endpoint wants to toss at it.  There's no source address available, so it doesn't know where in the src_tag_queues array to go.  Storing it somewhere else means that once the peer is inserted into the AV, any messages that have been received by that peer already need their source address updated.  (I think this problem shows up with the mpich spawn tests.)  We may need to require this behavior for users of peer srx queues.

Based on tcp, for a util_srx, the lock and rx_pool aren't general requirements.  The buf_pool in the xnet_srx is a lie.  It's not used anywhere.

The other wonderful discovery I recently made is that MPI will sometimes post a receive for a message *after* the peer process has exited.  shm probably doesn't care, because it doesn't support msg ep's, but verbs would.  In the case of tcp and verbs, the local msg ep would be disconnected and likely cleaned up prior to the receive being claimed by the app.  This is why you see both cq and cntr fields in the tcp srx struct.  I need to be able to complete a saved receive message after the receiving endpoint is no more.

shefty: I'd like to compare the tcp, verbs, rxm, and efa (if there's one) implementations, plus thinking about the locking more.  One of the other implementations or some combination may be a better starting point.

With FI_THREAD_DOMAIN, we probably don't need any locking at all.  The peer providers will also be sharing a domain.  For anything else, we'll need serialization, but need to agree on what that will look like.

aingerson: Are the peer providers always supposed to be sharing a domain? That's not how I'm doing it....

shefty: They don't have to be, but if they are, they have more information available, like the threading model used by the app.  Specifically, pairing shm with another provider, peer domain sharing makes sense.  Conceptually, that's what the app would view.

Can this help with locking?  I don't know yet.

prov/shm/src/smr_ep.c
```diff
@@ -1649,10 +1255,11 @@ int smr_srx_context(struct fid_domain *domain, struct fi_rx_attr *attr,
 
 	if (attr->op_flags & FI_PEER) {
 		smr_domain->srx = ((struct fi_peer_srx_context *) (context))->srx;
-		smr_domain->srx->peer_ops = &smr_srx_peer_ops;
+		//smr_domain->srx->peer_ops = &smr_srx_peer_ops;
```
shijin-aws, line 1258: Is it commented intentionally?

aingerson: Oh whoops. Should have just been deleted. Thanks for catching!

shijin-aws: May I know why it is deleted? Is it moved somewhere else?

aingerson: Oh, yes. The detection of an imported srx is now happening when the EP gets enabled [here](https://github.com/aingerson/libfabric/blob/peer/prov/shm/src/smr_ep.c#L1322)

This is just a draft proposal so a lot of it will probably change when I update :)

prov/shm/src/smr_ep.c
```diff
@@ -1697,16 +1305,25 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			domain = container_of(ep->util_ep.domain,
 					      struct smr_domain,
 					      util_domain.domain_fid);
-			ret = smr_ep_srx_context(domain, ep->rx_size,
```
shijin-aws, line 1700: would u mind explaining what's the use case for `ep->srx` is NULL and not? I still could not understand it after reading example SRX setup in the man page.

aingerson: Yeah the man pages are pretty confusing so I don't blame you. This is my flow that I've implemented in rxm for a shm offload:
1. rxm allocates it's own srx internally using the new util_ep_srx_context (util_srx)
2. rxm opens all of the shm resources (fabric/domain, ep, etc)
3. rxm "imports" its srx into shm with fi_srx_context (passing in FI_PEER and the peer context)
     - Since the fi_srx_context is associated with a domain, not an EP, we have to call fi_ep_bind to bind the srx (imported from rxm) to that specific EP. That's where ep->srx gets set. This indicates that we are using a pre-set up srx (whether that was created in fi_srx_context or imported in fi_srx_context. If there is no srx associated with the EP, we have to create a new one (util_ep_srx_context). If there is already an srx associated with it, we can use the owner ops from the imported one, but we still have to allocate a new one to save our peer ops in order to pass it back to the owner

So after the whole set up, rxm and shm will each have their own srx objects
rxm - srx resources (queue, pools, etc); owner_ops = rxm_srx_ops; peer_ops = rxm_start/discard ops;
shm - no srx resources; owner_ops = rxm_srx_ops (imported); peer_ops = shm_start/discard ops

So when shm calls owner_ops.get_msg(), it passes in the shm_srx so that rxm knows what peer ops to use to start that message once it gets a match
When rxm calls owner_ops.get_msg(), it passes in the rxm_srx to use its own peer_ops

Let me know if still still doesn't make sense.

shijin-aws: > Since the fi_srx_context is associated with a domain, not an EP, we have to call fi_ep_bind to bind the srx (imported from rxm) to that specific EP. That's where ep->srx gets set. This indicates that we are using a pre-set up srx (whether that was created in fi_srx_context or imported in fi_srx_context. If there is no srx associated with the EP, we have to create a new one (util_ep_srx_context). If there is already an srx associated with it, we can use the owner ops from the imported one, but we still have to allocate a new one to save our peer ops in order to pass it back to the owner

So rxm needs to 
1. call `util_srx_context(rxm_domain->util_domain, ... &rxm_ep->srx )` to allocate a new srx, (modify the owner_ops if necessary?)
2. create shm resources via `fi_domain`, `fi_endpoint` ....
3. construct a `fi_peer_srx_context` object `context`, set `context.srx = container_of(rxm_ep->srx, struct fid_peer_srx, ep_fid)`. construct an `attr |= FI_PEER`
4. call `fi_srx_context(&shm_domain, attr, ..., context)` to export the srx to shm domain
5. call `fi_ep_bind(shm_ep, srx, ..)` to bind the srx to shm ep

Inside shm provider, 
1. `smr_ep_bind` will set `smr_ep->srx` as the srx imported by rxm via the fi_ep_bind call
2. `smr_ep_ctrl` will see `smr_ep->srx` already associated, so it will calloc a new srx, get the owner_ops from the imported one, then re-assign the peer_ops as shm's peer_ops.

If this is correct, I am a bit confused on how step 4 is implemented in shm. I can see `smr_srx_context` only set `smr_domain->srx = ((struct fi_peer_srx_context *) (context))->srx;`  when  the attr has `FI_PEER` flag. Then how `smr_domain->srx` is used? It seems like shm only interacts with the peer_srx via ep but not domain.

> So after the whole set up, rxm and shm will each have their own srx objects
rxm - srx resources (queue, pools, etc); owner_ops = rxm_srx_ops; peer_ops = rxm_start/discard ops;
shm - no srx resources; owner_ops = rxm_srx_ops (imported); peer_ops = shm_start/discard ops
So when shm calls owner_ops.get_msg(), it passes in the shm_srx so that rxm knows what peer ops to use to start that message once it gets a match
When rxm calls owner_ops.get_msg(), it passes in the rxm_srx to use its own peer_ops

If rxm's srx has peer_ops as rxm_start/discard ops, how it can make shm start progressing the msg when rxm finds a match in the unexp queue, and the msg was queued by shm? I was referring to these steps in the man page

```
....
5. The application calls fi_recv() / fi_trecv() on owner, posting the
   matching receive buffer.
6. The owner matches the receive with the queued message on the peer.
7. The owner removes the queued request, fills in the rest of the known fields
   and calls the peer->start_msg() / start_tag() function.
```
Shouldn't rxm call shm_start_msg() in this situation? Do I understand the flow incorrectly?

aingerson: In regards to the shm context, the srx is directly set/imported from an owner when FI_PEER is set. If it is not set, we create a new one in shm:
```
int smr_srx_context(struct fid_domain *domain, struct fi_rx_attr *attr,
		    struct fid_ep **rx_ep, void *context)
{
	struct smr_domain *smr_domain;

	smr_domain = container_of(domain, struct smr_domain, util_domain.domain_fid);

	if (attr->op_flags & FI_PEER) {
		smr_domain->srx = ((struct fi_peer_srx_context *) (context))->srx;
		return FI_SUCCESS;
	}
	return util_ep_srx_context(&smr_domain->util_domain, attr->size,
				   SMR_IOV_LIMIT, SMR_INJECT_SIZE, rx_ep);
}
```
The second case is to support the existing fi_srx_context() API where an application may want to share an fi_srx_context between EPs under the same domain. I haven't tested that path, so there definitely could be issues in using that. I'm not sure. But in the context of importing it from the peer, it is still under the domain as a resource but can only be used directly by any EPs under that domain. So the domain doesn't necessarily become relevant in the operations, but rather is the owner of the resource itself. 
So there are two cases that we are interested in for this support - either an fi_srx_context was imported from an owner with FI_PEER to use shm as an intranode offload provider or there is no peer in which case shm creates its own srx when it does not detect an existing srx (this happens on enable). In the other, untested case where the application is sharing a shm srx to use among shm EPs, the individual EP would not create a second one on enable, but rather set the owner ops to use.


For the second question about the flow, and which set of start/discard ops to use, this is the reason why we need two srxs, one for each peer. In the flow, when a peer calls get_msg(), it passes in the srx to use for processing the message. So shm would pass in its srx (which has the shm_start_ops) and rxm would pass in its srx (which has the rxm_start_ops). The owner sets the srx to use for processing the message later.

When rxm saves an unexpected message for later, it allocates a new rx_entry and sets the peer_entry.srx to the inputted srx (if shm called it, it would be the shm srx with the shm start ops and if rxm called it, it would be the rxm start ops):
```
	match_attr.addr = addr;
	dlist_entry = dlist_find_first_match(&srx_ctx->recv_queue.list,
					     srx_ctx->recv_queue.match_func,
					     &match_attr);
	if (!dlist_entry) {
		util_entry = util_alloc_rx_entry(srx_ctx);
		if (!util_entry) {
			ret = -FI_ENOMEM;
		} else {
			util_entry->peer_entry.owner_context = NULL;
			util_entry->peer_entry.addr = addr;
			util_entry->peer_entry.size = size;
			util_entry->peer_entry.srx = srx;
			*rx_entry = &util_entry->peer_entry;
			ret = -FI_ENOENT;
		}
		goto out;
	}
```
When a new receive gets posted, it searches the unexpected queue to see if there is a waiting unexpected message that came in that would match that receive and calls the start_msg for the peer srx that added that unexpected message:
```
	ofi_spin_lock(&srx->lock);
	dlist_entry = dlist_remove_first_match(&unexp_queue->list,
					       unexp_queue->match_func,
					       &match_attr);
	if (!dlist_entry) {
		rx_entry = util_get_recv_entry(srx, iov, desc, iov_count, addr,
					       context, tag, ignore, flags);
		if (!rx_entry)
			ret = -FI_ENOMEM;
		else
			dlist_insert_tail((struct dlist_entry *)
					  (&rx_entry->peer_entry),
					  &recv_queue->list);
		ofi_spin_unlock(&srx->lock);
		return ret;
	}
	ofi_spin_unlock(&srx->lock);

	rx_entry = container_of(dlist_entry, struct util_rx_entry, peer_entry);
	util_init_rx_entry(srx, rx_entry, iov, desc, iov_count, addr, context,
			   tag, flags);

	return srx->peer_srx.peer_ops->start_msg(&rx_entry->peer_entry);
```

shijin-aws: Thank you very much for the explanation!

>When a new receive gets posted, it searches the unexpected queue to see if there is a waiting unexpected message that came in that would match that receive and calls the start_msg for the peer srx that added that unexpected message:

Yes, I get all of what you are talking for the expected flow. When peers call the `get_msg` op, it will set the srx to be used in `rx_entry->peer_entry.srx`, and when owner processes the matched rx entry from the unexp queue, it should find which srx to use from the rx entry.  

But looking at the code in `util_srx_generic_recv`,  the `srx` in `srx->peer_srx.peer_ops->start_msg(&rx_entry->peer_entry);`  is derived from the local ep, 

```
srx = container_of(ep_fid, struct util_srx_ctx, peer_srx.ep_fid);
```
 I was expecting it should be derived from the peer_srx in the matched `rx_entry`? The `rx_entry->peer_entry` passed to the op does not really make it toggle which srx to use.

include/ofi_util.h
shijin-aws: EFA creates unexp queue per ep and per peer both. The former case is for unspecified addr receive and the other one is for directed receive. 

I would like to see the util recv queue did the same thing, but it may need some more work. Currently my plan is to use EFA's own queue for SRX because the API is not specifying the queue implementation. Finally we could help making this code in util to remove the code duplication.

shefty: Does EFA have a base SRX implementation somewhere?

shijin-aws: The SRX is not implemented in EFA provider yet. We are working on it, hopefully will be available within a month or so.

shefty: I could find 2 relevant implementations for srx in tree, shm and tcp, and they basically have different implementations.  I'm not convinced that a util_srx will be all that usable.  I don't see that it hurts anything to have it, but it may not be ideal.

This really targets receiving on an rdm ep.  For example, the verbs srx targets msg eps's, and is even more different than shm or tcp because that's mapping to hardware.  It doesn't have tag matching or need to support complex protocols.

We may need to clarify that this is only usable with a single AV or single rdm ep + the peer.

prov/shm/src/smr_ep.c
```diff
@@ -1697,16 +1305,25 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
...
-			ret = smr_srx_bind(&ep->srx->fid,
+
+			util_get_peer_srx(ep->srx)->peer_ops = &smr_srx_peer_ops;
+			ret = util_srx_bind(&ep->srx->fid,
 					   &ep->util_ep.rx_cq->cq_fid.fid,
 					   FI_RECV);
 			if (ret)
 				return ret;
 		} else {
+			srx = calloc(1, sizeof(*srx));
```
shijin-aws, line 1321: I am not sure how this could work as peer provider get nothing but the owner_ops from the owner provider. If owner provider does not use `util_ep_srx_context` to initiate fid_peer_srx (as part of util_srx_ctx), how could peer provider access the resources (like queues, cq, etc.) in the owner provider when calling `get_msg/tag`? I see this change still make peer provider get the imported peer_srx's `ep_fid.fid.context`, which will be the `srx_ctx` in the owner provider. But I do not see the man page forces owner provider to setup `fid_peer_srx` resource in this way?

aingerson: @shijin-aws I'm not sure I understand your question. I have not updated the PR to include the fix which srx it uses when processing the unexpected message as I'm working on other updates to include with the next push as well.
What is the concern with initialization? What it does here is either allocate a util srx if it is not already bound to something (and it will create its own srx that it will use directly), or it creates a peer srx (without the resources since it will use the owner's) and imports the owner ops while setting its own peer ops. This is the srx that will be passed to the owner when getting a buffer/queuing a message.
Can you elaborate more on the concern?

shijin-aws: > or it creates a peer srx (without the resources since it will use the owner's)

Yes, this is my question, how does peer provider can access owner's resouces (in util_srx_ctx) from this fresh new `fid_peer_srx`? I guess you make it work by doing 

```
srx->ep_fid.fid.context =
				smr_get_peer_srx(ep)->ep_fid.fid.context;
```
So this new srx still has the context that was pointed to owner's `util_srx_ctx`.  Do I understand it correctly? My question is what if owner does not create its resource via `util_ep_srx_context` ?

aingerson: Oh I see your question. 
Yes, we get the context from the ep->srx which is a generic fid_peer_srx type (imported directly from the owner), so even if the owner is not using the util resources, it shouldn't matter. We will always just use the peer_srx->owner_ops.
The import of the srx is defines as follows by the man page:
`Peer SRXs are configured by the owner calling the peer’s fi_srx_context() call with the FI_PEER flag set. The context parameter passed to fi_srx_context() must be a struct fi_peer_srx_context..`
And the fi_peer_srx_context is defines as:
```
struct fid_peer_srx {
    struct fid_ep ep_fid;
    struct fi_ops_srx_owner *owner_ops;
    struct fi_ops_srx_peer *peer_ops;
};

struct fi_peer_srx_context {
    size_t size;
    struct fid_peer_srx *srx;
};
```
So when the peer imports fi_peer_srx_context, we save the entire fid_peer_srx *srx which is how we get to the owner ops

shijin-aws: I know I have bothered you on this many times so thank you for your patience in advance :)

The proposed workflow (I will just use shm as the peer)
1. Owner provider sets up it's own `struct fid_peer_srx` resource (in whatever way) as `owner_peer_srx`;
2. Owner provider sets up shm provider resources (info, domain, ep, etc.)
3. Owner provider creates `struct fi_peer_srx_context srx_context` and assign `srx_context.srx = owner_peer_srx`
4. Owner provider has an `attr.op_flags |= FI_PEER` and  call `fi_srx_context(shm_domain, attr, &dummy_srx, srx_context)`. In shm provider, `smr_srx_context` will set `smr_domain->srx = ((struct fi_peer_srx_context *) (context))->srx;` when there is a `FI_PEER` bit in `attr.op_flags`, and return success without creating a new `fid_peer_srx` object.
5. (This one is not listed in the example in man page but I think we need to do it according to code read and our earlier conversation) owner provider calls `fi_ep_bind(shm_ep_fid, owner_peer_srx.ep_fid, 0)`. In `smr_ep_bind`, it will finally set `ep->srx = owner_peer_srx`
6. Owner provider call fi_enable(shm_ep). `smr_ep_ctrl` will check if `ep->srx` is already set. And because of step 5 this is set. Then in this PR `smr_ep_ctrl` does 
```
srx = calloc(1, sizeof(*srx));
srx->peer_ops = &smr_srx_peer_ops;
			srx->owner_ops = smr_get_peer_srx(ep)->owner_ops;
			srx->ep_fid.fid.context =
				smr_get_peer_srx(ep)->ep_fid.fid.context;
			ep->srx = &srx->ep_fid;
```
which means it create a fresh new `fid_peer_srx`,  inherit  the owner ops from the owner provider as `srx->owner_ops = smr_get_peer_srx(ep)->owner_ops`, and the `ep_fid.fid.context` from the owner provider.

Now the issue is the owner_ops like `get_msg` need to access owner provider resources, e.g. packet pool, recv queue, in some way. And this "some way" really depends on how this `owner_peer_srx` was created in step 1. For example, in efa provider, I would like to make owner_peer_srx as a field in `efa_ep` struct, which has all the resources I need, so I will call some function like `efa_create_peer_srx(efa_ep, &efa_ep->peer_srx)` to create this `peer_srx` as part of an `efa_ep` object;  Then to implement efa provider's `get_msg` I will need to get `efa_ep` from the input `srx` variable. An easy and natural way to do that is to do this inside the `get_msg` call:
`efa_ep = container_of(srx, struct efa_ep, efa_ep.peer_srx)`.

But with your PR, this workflow will not work any more as the final `ep->srx` in shm provider will not be part of `efa_ep` because it is created by `calloc`. Then when shm_provider call `ep->srx->peer_ops->owner_ops->get_msg` it will likely to segfault because  `efa_ep = container_of(srx, struct efa_ep, efa_ep.peer_srx)` will not point to a valid `efa_ep` pointer.

In my mind this issue has 2 potential solutions:

Solution 1: efa provider needs to do `peer_srx.ep_fid.fid.context = efa_ep`, and retrieve `efa_ep` back in `get_msg` as `efa_ep = srx.ep_fid.fid.context` because shm provider's srx copied this information. 

Solution 2: Do not create a fresh new srx via calloc, but just make `ep->srx->peer_ops = &smr_srx_peer_ops;`, which means you just replace the peer_ops in the imported srx.

One final question is that I still do not see how `smr_domain->srx` was finally used by shm provider in the workflow I proposed above. If it is not used, what is the influence of Step 3 and 4 in my workflow?

I hope the explanation to my issue makes sense and It will be greatly appreciated if you could give me some feedback on them, thanks!

aingerson: @shijin-aws No worries, I'm happy to explain my vision as much as you need!
The util srx implementation sets the peer_srx context to be the util_srx
`srx->peer_srx.ep_fid.fid.context = srx;`
srx here is of type struct util_srx_ctx * which is what owns all of the resources (rx pool, queues, etc)
So when we set the duplicate srx (for the peer/shm), we set its context to be the same as the imported context:
`srx->ep_fid.fid.context = smr_get_peer_srx(ep)->ep_fid.fid.context;`
When we pass in the duplicate srx, it will have the same context and owner ops as the owner so we will call the owner's get_msg which will point to the util get_msg code which will cast the context of the srx into a util_srx (since it set it, it knows that it is a util_srx type)
`srx_ctx = srx->ep_fid.fid.context;`
If a provider is not using the util implementation, that is fine; it will just have to set the context to a pointer that will help it get to its resources. I would assume that this would be separate from the EP since the srx is not owned by a single EP. If you need a specific piece of information during get/queue that is currently in the EP, then I would think it would need to be moved to the srx set of resources instead.
Also I want to note that the srx fid is of type ep_fid because it can be passed in to a fi_recv call so I'm not sure when you say "get to the efa_ep", I'm not sure if you mean the efa_ep for an EP or the srx ep_fid.
I hope that makes sense?
In regards to your question about the domain - it's not necessarily that it is needed/used, it's just that when sharing the srx, the resource does not belong to a single EP so we associate it with a domain instead. As far as whether that is a separate domain per peer or a shared domain is up to interpretation (see the first conversation in this PR which discusses the shared domain question).

shijin-aws: Thanks a lot! I think your comment answers all of my questions

> If a provider is not using the util implementation, that is fine; it will just have to set the context to a pointer that will help it get to its resources. I would assume that this would be separate from the EP since the srx is not owned by a single EP. If you need a specific piece of information during get/queue that is currently in the EP, then I would think it would need to be moved to the srx set of resources instead.

Yes, I think shm choose to move these resources to `util_srx_ctx` because it wants to support the public `fi_srx_context` API which makes the srx can be shared among multiple EP. But if the owner provider is not going support `fi_srx_context` that allow user to access SRX and bind it to multiple EPs, I think it should be fine to make srx access resources in owner's ep directly.

> Also I want to note that the srx fid is of type ep_fid because it can be passed in to a fi_recv call so I'm not sure when you say "get to the efa_ep", I'm not sure if you mean the efa_ep for an EP or the srx ep_fid.

I was just making an example of how the `get_msg` ops can get to the owner resources from the input `fid_peer_srx` variable. It does not relate to the ep_fid in fi_recv. Now it's clear to me that I need to access the resource via `ep_fid.fid.context`. So this is not an issue.

> In regards to your question about the domain - it's not necessarily that it is needed/used, it's just that when sharing the srx, the resource does not belong to a single EP so we associate it with a domain instead. As far as whether that is a separate domain per peer or a shared domain is up to interpretation (see the first conversation in this PR which discusses the shared domain question).

I see. Just want to double check: If I only want to make `shm_ep->srx` get the owner_ops and context from owner's peer_srx, the `fi_ep_bind(shm_ep, ..., owner_peer_srx)` should be enough, right?

aingerson: It might be enough for your specific case but I think it makes the peer vs generic srx case foggy. The combination of calling fi_srx_context with the FI_PEER flag + fi_ep_bind makes it clear that the expectation is to go through the peer API. If we're just calling fi_ep_bind, it might not be explicit enough to fit the definition. I think that's why we defined the peer flow to go through both calls.

shijin-aws: I see, I will follow this flow in my implementation, thanks!

PR #8457, opened 2023-02-04
prov/efa: move op_entry flags from rxr.h to rxr_op_entry.h

wzamazon:
This patch move flags that are only applied to rxr_op_entry from rxr.h to rxr_op_entry.h.

It also rename some flags and added OP/TX/RX_ENTRY to indicate their usage.


wenduwan (approved):
Thanks

lrbison (approved):
lgtm

PR #8476, opened 2023-02-09
prov/hook/trace: Change tag in trace log to hex format.

peinanz:


shefty:
thanks!

PR #8481, opened 2023-02-10
ucx provider code refactoring

j-xiong:
Address the two remaining review comments for the initial push:
(1) use standard macro for checking lookback devices.
(2) use ofi_dyn_arr for variable length tables used in address vector.

### DIFF
--- a/prov/ucx/src/ucx_av.c
+++ b/prov/ucx/src/ucx_av.c
@@ -111,8 +111,6 @@ static int ucx_av_insert(struct fid_av *fi_av, const void *addr, size_t count,
 {
 	struct ucx_av *av;
 	struct ucx_ep *ep;
-	struct ucx_avblock *avb_eps;
-	struct ucx_avblock *avb_addrs;
 	size_t i;
 	ucs_status_t status = UCS_OK;
 	int added = 0;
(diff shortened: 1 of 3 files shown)


prov/ucx/src/ucx_fabric.c
```diff
@@ -97,30 +97,24 @@ static char* ucx_local_host_resolve()
...
-			host[0] = '\0';
-			continue;
-		}
-
 		/* Skip loopback devices */
-		if (strncmp(host, UCX_IGNORED_LO_ADDR,
-			    strlen(UCX_IGNORED_LO_ADDR)) == 0) {
-			host[0] = '\0';
+		sin = (struct sockaddr_in *) ifa->ifa_addr;
+		if (sin->sin_addr.s_addr == htonl(INADDR_LOOPBACK))
```
shefty, line 107: Use (ifa->ifa_flags & IFF_LOOPBACK) above instead?

prov/ucx/src/ucx_av.c
```diff
@@ -269,6 +245,11 @@ static struct fi_ops_av ucx_av_ops = {
 	.lookup = ucx_av_lookup,
 };
 
+static void ucx_init_array(struct ofi_dyn_arr *arr, void *item)
+{
+	memset(item, 0, arr->item_size);
+}
```
shefty, line 251: The array uses calloc to initialize entries to 0.  We can document that behavior.

j-xiong:
Updated to address the comments.

shefty (approved):

j-xiong:
No need to wait for ci since the provider is not being tested.

PR #8483, opened 2023-02-10
prov/efa: add some docstrings, rename non-inclusive variables

shijin-aws:


prov/efa/src/rdm/rxr_msg.c
```diff
@@ -699,6 +699,16 @@ struct rxr_op_entry *rxr_msg_alloc_unexp_rx_entry_for_tagrtm(struct rxr_ep *ep,
 	return rx_entry;
 }
 
+/**
+ * @brief Split a posted rx entry to be consumed by multi-recv buffers.
+ * After the split, the iov of the posted entry will be adjusted to
+ * reflect the consumption.
+ * @param[in] ep endpoint
+ * @param[in/out] posted_entry the rx entry posted by user for multi-recv
```
wenduwan, line 707: Does your ide highlight this syntax?
On my side
```
@param[in,out]
```
is highlighted, but
```
@param[in/out]
```
is not.

wenduwan (approved):
Thank you!

prov/efa/src/rdm/rxr_msg.c
shijin-aws: Yes, it should be a comma. Updated, thanks.

prov/efa/src/rdm/rxr_op_entry.h
```diff
@@ -185,7 +185,11 @@ struct rxr_op_entry {
 	 */
 	struct dlist_entry multi_recv_consumers;
 	struct dlist_entry multi_recv_entry;
-	struct rxr_op_entry *master_entry;
+	/*
+	 * The main entry for the multi-recv buffer, which could be split into
+	 * multiple consumer entries.
+	 */
+	struct rxr_op_entry *main_entry;
```
darrylabbate, line 192: I think `producer` would be clearer than `main` if the companion term is `consumer`. Just a thought.

shijin-aws: See this man page: https://ofiwg.github.io/libfabric/main/man/fi_endpoint.3.html
The main entry is a  
>single buffer that will receive multiple incoming messages. Received messages will be packed into the receive buffer until the buffer has been consumed.

I guess the `consumer` was inherited from the `consumed` in the man page. But naming the main entry as `producer` is not quite appropriate to me either.

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -699,6 +699,16 @@ struct rxr_op_entry *rxr_msg_alloc_unexp_rx_entry_for_tagrtm(struct rxr_ep *ep,
 	return rx_entry;
 }
 
+/**
+ * @brief Split a posted rx entry to be consumed by multi-recv buffers.
```
wzamazon, line 703: It is not clear what

`to be consumed by multi-recv buffers`

means. Maybe rephrase a little.

Essentially, this function make consumer_entry, to use buffer in `posted_entry`. Then update posted_entry.

shijin-aws:
This PR is too old. Will open a new one when it's ready.

PR #8484, opened 2023-02-11
prov/hook/trace: Add trace log for CM operation APIs.

peinanz:
Add trace log for cm operations.
Remove postfix "_op" in trace cq ops, to be consistent with other trace ops.

### DIFF
--- a/prov/hook/trace/src/hook_trace.c
+++ b/prov/hook/trace/src/hook_trace.c
@@ -37,6 +37,12 @@
 #define IOV_LEN(iov, count)	    ofi_total_iov_len(iov, count)
 #define MSG_DATA(data, flags)   (flags & FI_REMOTE_CQ_DATA ? data : 0)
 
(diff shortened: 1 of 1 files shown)


prov/hook/trace/src/hook_trace.c
```diff
@@ -830,6 +836,110 @@ static struct fi_ops_tagged trace_tagged_ops = {
 	.injectdata = trace_tinjectdata,
 };
 
+static int trace_setname(fid_t fid, void *addr, size_t addrlen)
+{
+	int ret = 0;
+	struct hook_ep *myep = NULL;
+
+	if (fid->fclass == FI_CLASS_EP) {
```
j-xiong, line 844: Could FI_CLASS_PEP also be handled here? same for getname.

peinanz: Yes, I can enable trace cm for FI_CLASS_PEP.  The current functions are set up for FI_CLASS_EP only.

Issue #8472, opened 2023-02-08
Does the len field in cq_data_entry apply to write with remote cq data?


wzamazon:
**Is your feature request related to a problem? Please describe.**

currently, libfabric document states that the "len" field in "fi_cq_data_entry" apply only to fi_recv and fi_trecv.

https://ofiwg.github.io/libfabric/v1.17.0/man/fi_cq.3.html

However, fi_write with FI_REMOTE_CQ_DATA will generate a cq_entry on the responder side, and this field will be very useful under such case. With out this, there might need to be additional message  to exchange the write length.

As far as I know, at least EFA provider and SHM provider will set this field as the actual write length in the fi_cq_data_entry generated on the responder. 

**Describe the solution you'd like**
Make sure all providers write this field properly on responder side, for write with FI_REMOTE_CQ_DATA, then update the document to reflect that his field apply to write with FI_REMOTE_CQ_DATA on the responder side.

j-xiong:
Remote CQ data can also happen with send. Associating len with the CQ data won't be consistent between write and send. In addition, the size of CQ data is fixed for a provider and thus doesn't need an extra setting for that.

wzamazon:
Sorry, I did not mean the length of the CQ data (which is fixed). I meant the message size.

For example, when user call `fi_write` to write a 1 MB message

If the user also specify FI_REMOTE_CQ_DATA, the responder of the `fi_write` will get a CQ entry. I think it would be useful for the `len` field in the CQ entry to be `1048576`.  

Currently, some provider already do this. (like EFA and SHM).

wzamazon:
The reason the field may be useful for application is that application often provide a large buffer for other to write to.

For example, application know that the maximum write size is 4MB, then it can prepare a 4MB buffer for others to write to.

But the writer does not always write 4MB, it may only write 1MB. so this field can be used for the write responder to know how many bytes has been written to it.

j-xiong:
Ah, that makes sense. I agree this is useful.

shefty:
Verbs cannot provide this.  CQ data is really orthogonal here.  More generically, the request is to generate a completion at the target of an RMA operation with information about the operation that just occurred.  This could be a read or write.  The current API requires the sender to convey this information in a separate message.

[...]

PR #8488, opened 2023-02-13
man/fi_cq: Extend len field to write with remote cq data on responder

wzamazon:
Extend the len field of the CQ entry to cover write with remote CQ data.


wzamazon:
addresses https://github.com/ofiwg/libfabric/issues/8472

man/fi_cq.3.md
```diff
@@ -477,12 +477,15 @@ of these fields are the same for all CQ entry structure formats.
...
 
 *len*
-: This len field only applies to completed receive operations (e.g. fi_recv,
-  fi_trecv, etc.).  It indicates the size of received _message_ data --
-  i.e. how many data bytes were placed into the associated receive buffer by
-  a corresponding fi_send/fi_tsend/et al call.  If an endpoint has
+: This len field applies to completed receive operations (e.g. fi_recv,
+  fi_trecv, etc.) and the completed write with remote cq data on the
+  responder side (e.g. fi_write, with FI_REMOTE_CQ_DATA flag).
+  It indicates the size of transfered _message_ data --
```
darrylabbate, line 483:
```suggestion
  It indicates the size of transferred _message_ data --
```

darrylabbate (changes requested):

man/fi_cq.3.md
wzamazon: fixed

darrylabbate (approved):

man/fi_cq.3.md
```diff
@@ -477,12 +477,15 @@ of these fields are the same for all CQ entry structure formats.
...
+: This len field applies to completed receive operations (e.g. fi_recv,
+  fi_trecv, etc.) and the completed write with remote cq data on the
+  responder side (e.g. fi_write, with FI_REMOTE_CQ_DATA flag).
+  It indicates the size of transferred _message_ data --
+  i.e. how many data bytes were placed into the associated receive/target buffer by
+  a corresponding fi_send/fi_tsend/fi_write et al call.  If an endpoint has
   been configured with the FI_MSG_PREFIX mode, the len also reflects the size
-  of the prefix buffer.
+  of the prefix buffer. For fi_write with remote cq data, the len does
+  not include CQ data size.
```
j-xiong, line 488: Regular receive may contain CQ data, which is also not included in `len`.  I suggest removing the `For fi_write with remote cq data, ` part.

wzamazon: Thank you! Removed in new revision

j-xiong (approved):

man/fi_cq.3.md
```diff
@@ -477,10 +477,12 @@ of these fields are the same for all CQ entry structure formats.
...
-: This len field only applies to completed receive operations (e.g. fi_recv,
-  fi_trecv, etc.).  It indicates the size of received _message_ data --
-  i.e. how many data bytes were placed into the associated receive buffer by
-  a corresponding fi_send/fi_tsend/et al call.  If an endpoint has
+: This len field applies to completed receive operations (e.g. fi_recv,
+  fi_trecv, etc.) and the completed write with remote cq data on the
+  responder side (e.g. fi_write, with FI_REMOTE_CQ_DATA flag).
+  It indicates the size of transferred _message_ data --
+  i.e. how many data bytes were placed into the associated receive/target buffer by
+  a corresponding fi_send/fi_tsend/fi_write et al call.  If an endpoint has
```
shefty, line 485: fi_write -> fi_writedata

shijin-aws:
> @shefty
> 
> Can you take a look?

I do not quite follow, isn't Sean asking for a change of `fi_write` to `fi_writedata` in your patch?

wzamazon:
>I do not quite follow, isn't Sean asking for a change of fi_write to fi_writedata in your patch?


Sorry, I missed that comment

man/fi_cq.3.md
```diff
@@ -477,10 +477,12 @@ of these fields are the same for all CQ entry structure formats.
   set for all relevant completions.
 
 *len*
-: This len field only applies to completed receive operations (e.g. fi_recv,
-  fi_trecv, etc.).  It indicates the size of received _message_ data --
-  i.e. how many data bytes were placed into the associated receive buffer by
-  a corresponding fi_send/fi_tsend/et al call.  If an endpoint has
+: This len field applies to completed receive operations (e.g. fi_recv,
+  fi_trecv, etc.) and the completed write with remote cq data on the
+  responder side (e.g. fi_write, with FI_REMOTE_CQ_DATA flag).
```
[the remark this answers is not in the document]
wzamazon: Fixed in new revision.

man/fi_cq.3.md
wzamazon: Fixed in new revision

shijin-aws:
Sean also asked this

> Do we have a fabtest that verifies this behavior? I don't believe all providers adhere to this definition.

Can we add that?

fabtests/common/shared.c
```diff
@@ -3372,6 +3397,18 @@ uint64_t ft_init_cq_data(struct fi_info *info)
 	}
 }
 
+int ft_check_cq_data()
+{
+	if (rx_cq_entry_has_data) {
+		if (rx_cq_data != ft_init_cq_data(fi)) {
```
shefty, line 3403: A call like 'init' should only be called once.  Calling this every time there's a check against the data looks wrong.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -161,6 +161,10 @@ static int bw_rx_comp()
 				   opts.transfer_size);
 		if (ret)
 			return ret;
+
+		ret = ft_check_cq_data();
+		if (ret)
+			return ret;
```
shefty, line 167: The benchmarks are performance focused tests (arguably).  Adding checks to the test is not desirable.

fabtests/common/shared.c
```diff
@@ -2397,6 +2414,14 @@ static int ft_fdwait_for_comp(struct fid_cq *cq, uint64_t *cur,
...
 				return -FI_EOTHER;
 			(*cur)++;
+
+			if (comp.flags & FI_REMOTE_CQ_DATA) {
+				rx_cq_entry_has_data = true;
+				rx_cq_data = comp.data;
+			} else {
+				rx_cq_entry_has_data = false;
+			}
+
```
shefty, line 2424: We have a fabtest that verifies cq_data.  Sure, it's lame, but is this much of an improvement?  Second, the above structure is repeated 3 times.  Why not just check the cq data directly in all places, similar to how the tag check is done?  Saving the information to check later is going to make it hard to figure out what's happening and make changes more difficult.

Finally, the update to the man page was to define 'len' as valid for RMA writes with cq data.  The request was for something in fabtests to verify that len is actually set, so we can identify what providers need to be updated to support the new definition.

wzamazon: > We have a fabtest that verifies cq_data. Sure, it's lame, but is this much of an improvement? Second, the above structure is repeated 3 times. Why not just check the cq data directly in all places, similar to how the tag check is done? Saving the information to check later is going to make it hard to figure out what's happening and make changes more difficult.

I just noticed that there is `fi_cq_data` test, that test senddata, I am working on adding `writedata` support to it.

> Finally, the update to the man page was to define 'len' as valid for RMA writes with cq data. The request was for something in fabtests to verify that len is actually set, so we can identify what providers need to be updated to support the new definition.

So we can merge the 1st commit as is, and I will open a separate PR to add the test.

Does the plan sounds right?

shefty: ubertest already supports writedata, so I don't know that we even need a new test for cq data.

The problem with the man page update is that it is defining new API requirements, and I don't think that any of the providers actually support that requirement.  We need to coordinate updating all of the providers with the man page.

What we don't have is a fabtest that verifies the len field in the cq_entry is set correctly for an RMA write with cq data.  We're not trying to validate the cq data, but the len.  And I expect that if we add that test, CI will fail because the providers aren't taking that action.

wzamazon: I can add verfication of `len` to `fi_cq_data`, and open PR.

The PR will trigger AWS CI, which will test shm, tcp, sockets and EFA.

shefty: You can also update ubertest.  There's a ft_verify_comp() call for this purpose.

wzamazon:
@shefty

Can we merge this PR too?

PR #8490, opened 2023-02-13
prov/efa: LTTng tracepoint touch-ups + expanded documentation

darrylabbate:
This PR:
- Touches up the pre-defined tracepoints
- Changes the `*_tracing()` macros to `*_tracepoint()`, matching `lttng_ust_tracepoint()`
- Utilizes the `CLASS()` / `INSTANCE()` macros to alleviate a small amount of boilerplate for tracepoints with common fields
- Includes the lttng/tracef.h and lttng/tracelog.h headers (with corresponding `*_trace(f|log)` macros) for developers to easily plug into the source code for quick debugging with LTTng
- Add docs/lttng.md, which expands on the documentation provided by docs/efa_tracing.txt (which has now been removed)


prov/efa/src/dgram/efa_dgram_msg.c
```diff
@@ -326,7 +326,7 @@ ssize_t efa_dgram_post_flush(struct efa_dgram_ep *ep, struct ibv_send_wr **bad_w
 #if HAVE_LTTNG
 	struct ibv_send_wr *head = ep->base_ep.xmit_more_wr_head.next;
 	while (head) {
-		efa_tracing(post_send, (void *) head->wr_id);
+		efa_tracepoint_wr_id_post_flush((void *) head->wr_id);
```
wzamazon, line 329: `post_flush` -> `post_send`

prov/efa/src/efa_tp_def.h
```diff
@@ -21,30 +23,20 @@
...
-	),
-	LTTNG_UST_TP_FIELDS(
-		X_PKT_FIELDS  
-	)
-)
-LTTNG_UST_TRACEPOINT_LOGLEVEL(EFA_TP_PROV, post_send, LTTNG_UST_TRACEPOINT_LOGLEVEL_INFO)
-
-LTTNG_UST_TRACEPOINT_EVENT(
-	EFA_TP_PROV,
+LTTNG_UST_TRACEPOINT_EVENT_CLASS(EFA_TP_PROV, post_wr_id,
```
wzamazon, line 26: post_send and post_recv are two different events. 

Both events have wr_id as an argument.

wzamazon (changes requested):

prov/efa/src/dgram/efa_dgram_msg.c
darrylabbate: I changed it to `post_flush` since those were the only functions this tracepoint was used, but I'm indifferent so I'll change it back.

prov/efa/src/efa_tp_def.h
darrylabbate: Is your comment re: the tracepoints being otherwise identical? If so, the purpose would be to isolate each event if you wish. E.g.

```
lttng enable-event --userspace efa:post_recv
```

which would only execute the `efa_tracepoint_wr_id_post_recv()` traces.

Keep in mind these tracepoints existed before this PR.

wzamazon: I was not familiar with `LTTNG_UST_TRACEPOINT_EVENT_CLASS` and `LTTNG_UST_TRACEPOINT_EVENT_INSTANCE`. 

I thought you changed the tracepoint's name from `post_send` and `post_recv` to `post_wr_id`.

While in fact, `post_wr_id` is the template you defined.

I think the name is a little confusing.

prov/efa/src/efa_tp.h
```diff
@@ -1,51 +1,51 @@
...
-	case post_recv:
-		lttng_ust_tracepoint(EFA_TP_PROV, post_recv,
-				     (size_t)wr_id,
-				     (size_t)op_entry,
-				     (size_t)op_entry->cq_entry.op_context);
-		break;
-	default:
-		assert(0);
-	}
+	efa_tracepoint(post_flush, (size_t) wr_id, (size_t) op_entry, (size_t) op_entry->cq_entry.op_context);
```
wzamazon, line 31: should be `post_send`. Right?

darrylabbate: Thanks, I missed that

prov/efa/src/efa_tp_def.h
darrylabbate: Correct. On a side note, I spent some time trying to create some macros to clean it up a little bit - e.g. `EFA_TRACEPOINT_EVENT()` but something with the overall setup causes some preprocessor expansion issues with the nested macros' commas.

Do you have a preferred alternative to `post_wr_id` for the tracepoint class?

wzamazon (approved):

prov/efa/src/efa_tp_def.h
wzamazon: I do not have better idea. I think it is good for now.

PR #8491, opened 2023-02-14
prov/hook/trace: Add trace log for CM operation APIs.

peinanz:
Add trace log on cm operations for ep/pep.
Remove postfix "_op" in trace cq ops, to be consistent with other trace ops.

### DIFF
--- a/prov/hook/trace/src/hook_trace.c
+++ b/prov/hook/trace/src/hook_trace.c
@@ -37,6 +37,23 @@
 #define IOV_LEN(iov, count)	    ofi_total_iov_len(iov, count)
 #define MSG_DATA(data, flags)   (flags & FI_REMOTE_CQ_DATA ? data : 0)
 
+#define TRACE_CM(ret, fabric, ep, flags)  \
+	if (!(ret)) { \
+		if ((flags)) \
+			FI_TRACE((fabric)->hprov, FI_LOG_EP_CTRL, \
+			         "ep/pep %p flags 0x%lx\n", \
+			         (void *)(ep), (uint64_t)(flags)); \
+		else \
+			FI_TRACE((fabric)->hprov, FI_LOG_EP_CTRL, \
+			         "ep/pep %p\n", (void *)(ep)); \
+	}
+
+#define TRACE_CM_ADDR(ret, fabric, addr)  \
+	if (!(ret)) { \
+		ofi_straddr_log(fabric->hprov, FI_LOG_TRACE, FI_LOG_EP_CTRL, \
+		                "addr", addr);	\
+	}
+
 #define TRACE_EP_MSG(ret, ep, buf, len, addr, data, flags, context) \
 	if (!(ret)) { \
 		FI_TRACE((ep)->domain->fabric->hprov, FI_LOG_EP_DATA, \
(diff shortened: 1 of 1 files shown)


prov/hook/trace/src/hook_trace.c
```diff
@@ -831,6 +837,101 @@ static struct fi_ops_tagged trace_tagged_ops = {
...
+	TRACE_CM(ret, myep->domain->fabric, (addr) ? (*(uint64_t *)addr) : 0);
+
+	return ret;
+}
+
+static int trace_listen(struct fid_pep *pep)
+{
+	struct hook_pep *mypep = container_of(pep, struct hook_pep, pep);
+
+    return fi_listen(mypep->hpep);
```
j-xiong, line 889: Any reason these calls (listen/accept/reject/shutdown/join) are not traced? A single `TRACE_CM` might not fit all of them but you can always call `FI_TRACE` directly.

peinanz: Is there any useful information we want to log for listen/accept/reject/shutdown/join ops?  These ops only have pointers to ep/pep.  Will such information be useful for debugging?

j-xiong: Indeed, the details are carried in `param`, but those are provider specific so not much can be done other than printing the address and length, which may not be very useful. However, merely info of these functions being called can sometimes be useful.

prov/hook/trace/src/hook_trace.c
```diff
@@ -831,6 +837,101 @@ static struct fi_ops_tagged trace_tagged_ops = {
 };
 
 
+static int trace_setname(fid_t fid, void *addr, size_t addrlen)
+{
+	int ret = 0;
+	struct hook_fabric *myfabric = hook_to_fabric(fid);
+
+	ret = fi_setname(hook_to_hfid(fid), addr, addrlen);
+	TRACE_CM(ret, myfabric, (addr) ? (*(uint64_t *)addr) : 0);
```
j-xiong, line 846: What if addrlen > sizeof(uint64_t)?

peinanz: Good catch, will fix the log of addr in all relevant cm ops. I thought the addr parameter is in the same type as in ep ops.

shefty: Use ofi_straddr_log() to print the address in a readable format.  It only works with sockaddr, I think, but that may be the best we can do for now.

prov/hook/trace/src/hook_trace.c
shefty: indentation above is off.  I agree with @j-xiong , just knowing the calls were made can be useful.  Printing the ep pointers helps.

prov/hook/trace/src/hook_trace.c
```diff
@@ -910,16 +1022,15 @@ static int trace_cq_signal_op(struct fid_cq *cq)
...
-	.readerr = trace_cq_readerr_op,
-	.sread = trace_cq_sread_op,
-	.sreadfrom = trace_cq_sreadfrom_op,
-	.signal = trace_cq_signal_op,
+	.read = trace_cq_read,
+	.readfrom = trace_cq_readfrom,
+	.readerr = trace_cq_readerr,
+	.sread = trace_cq_sread,
+	.sreadfrom = trace_cq_sreadfrom,
+	.signal = trace_cq_signal,
```
shefty, line 1030: The CQ changes should be in a separate patch.

shefty (changes requested):

PR #8492, opened 2023-02-14
prov/efa: re-arrange packet pool

wzamazon:
A series of changes to re-arrange packet pool, aimed at reducing memory usage and improve performance

### DIFF
--- a/prov/efa/src/rdm/rxr_pkt_entry.c
+++ b/prov/efa/src/rdm/rxr_pkt_entry.c
@@ -64,23 +64,20 @@ struct rxr_pkt_entry *rxr_pkt_entry_alloc(struct rxr_ep *ep, struct rxr_pkt_pool
 	struct rxr_pkt_entry *pkt_entry;
 	void *mr = NULL;
 
-	pkt_entry = ofi_buf_alloc_ex(pkt_pool->localinfo_pool, &mr);
+	pkt_entry = ofi_buf_alloc_ex(pkt_pool->entry_pool, &mr);
 	if (!pkt_entry)
 		return NULL;
 
 #ifdef ENABLE_EFA_POISONING
-	memset(pkt_entry, 0, sizeof(*pkt_entry));
+	rxr_poison_mem_region(pkt_entry, sizeof(struct rxr_pkt_entry) + ep->mtu_size);
 #endif
 
-	pkt_entry->wiredata = ofi_buf_alloc_ex(pkt_pool->wiredata_pool, &mr);
-	if (!pkt_entry->wiredata) {
-		ofi_buf_free(pkt_entry);
-		return NULL;
+	pkt_entry->send = NULL;
+	if (pkt_pool->sendv_pool) {
+		pkt_entry->send = ofi_buf_alloc(pkt_pool->sendv_pool);
+		assert(pkt_entry->send);
+		pkt_entry->send->iov_count = 0; /* rxr_pkt_init methods expect iov_count = 0 */
 	}
-
-#ifdef ENABLE_EFA_POISONING
-	memset(pkt_entry->wiredata, 0, ep->mtu_size);
-#endif
 
 	dlist_init(&pkt_entry->entry);
 
@@ -96,8 +93,6 @@ struct rxr_pkt_entry *rxr_pkt_entry_alloc(struct rxr_ep *ep, struct rxr_pkt_pool
 	pkt_entry->flags = RXR_PKT_ENTRY_IN_USE;
 	pkt_entry->next = NULL;
 	pkt_entry->x_entry = NULL;
-	pkt_entry->send.iov_count = 0; /* rxr_pkt_init methods expect iov_count = 0 */
-
 	pkt_entry->send_wr.wr.next = NULL;
 	pkt_entry->send_wr.wr.send_flags = 0;
 
@@ -111,9 +106,15 @@ struct rxr_pkt_entry *rxr_pkt_entry_alloc(struct rxr_ep *ep, struct rxr_pkt_pool
  *
  * @related rxr_pkt_entry
  */
-void rxr_pkt_entry_release(struct rxr_pkt_entry *pkt_entry)
+void rxr_pkt_entry_release(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry)
 {
-	ofi_buf_free(pkt_entry->wiredata);
+	if (pkt_entry->send)
+		ofi_buf_free(pkt_entry->send);
+
+#ifdef ENABLE_EFA_POISONING
+	rxr_poison_mem_region(pkt_entry, sizeof(struct rxr_pkt_entry) + ep->mtu_size);
+#endif
+	pkt_entry->flags = 0;
 	ofi_buf_free(pkt_entry);
 }
 
@@ -150,11 +151,8 @@ void rxr_pkt_entry_release_tx(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry
 		       "reset backoff timer for peer: %" PRIu64 "\n",
 		       pkt_entry->addr);
 	}
-#ifdef ENABLE_EFA_POISONING
-	rxr_poison_pkt_entry(pkt_entry, ep->mtu_size);
-#endif
-	pkt_entry->flags = 0;
-	rxr_pkt_entry_release(pkt_entry);
+
+	rxr_pkt_entry_release(ep, pkt_entry);
 }
 
 /**
@@ -193,12 +191,7 @@ void rxr_pkt_entry_release_rx(struct rxr_ep *ep,
 #if ENABLE_DEBUG
 	dlist_remove(&pkt_entry->dbg_entry);
 #endif
-#ifdef ENABLE_EFA_POISONING
-	/* the same pool size is used for all types of rx pkt_entries */
-	rxr_poison_pkt_entry(pkt_entry, ep->mtu_size);
-#endif
-	pkt_entry->flags = 0;
-	rxr_pkt_entry_release(pkt_entry);
+	rxr_pkt_entry_release(ep, pkt_entry);
 }
 
 void rxr_pkt_entry_copy(struct rxr_ep *ep,
@@ -276,11 +269,7 @@ void rxr_pkt_entry_release_cloned(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_e
 	while (pkt_entry) {
 		assert(pkt_entry->alloc_type == RXR_PKT_FROM_OOO_POOL ||
 		       pkt_entry->alloc_type == RXR_PKT_FROM_UNEXP_POOL);
-#ifdef ENABLE_EFA_POISONING
-	rxr_poison_pkt_entry(pkt_entry, ep->mtu_size);
-#endif
-		pkt_entry->flags = 0;
-		rxr_pkt_entry_release(pkt_entry);
+		rxr_pkt_entry_release(ep, pkt_entry);
 		next = pkt_entry->next;
 		pkt_entry = next;
 	}
@@ -367,7 +356,7 @@ ssize_t rxr_pkt_entry_send(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry,
 {
 	assert(pkt_entry->pkt_size);
 	struct efa_rdm_peer *peer;
-	struct rxr_pkt_sendv *send = &pkt_entry->send;
+	struct rxr_pkt_sendv *send = pkt_entry->send;
 	struct ibv_send_wr *bad_wr, *send_wr = &pkt_entry->send_wr.wr;
 	struct ibv_sge *sge;
 	int ret, total_len;
@@ -385,6 +374,7 @@ ssize_t rxr_pkt_entry_send(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry,
 	conn = efa_av_addr_to_conn(ep->base_ep.av, pkt_entry->addr);
 	assert(conn && conn->ep_addr);
 
+	assert(send);
 	if (send->iov_count == 0) {
 		send->iov_count = 1;
 		send->iov[0].iov_base = pkt_entry->wiredata;
--- a/prov/efa/src/rdm/rxr_pkt_entry.c
+++ b/prov/efa/src/rdm/rxr_pkt_entry.c
@@ -46,6 +46,7 @@
 #include "rxr_rma.h"
 #include "rxr_op_entry.h"
 #include "rxr_pkt_cmd.h"
+#include "rxr_pkt_pool.h"
 
 /**
  * @brief allocate a packet entry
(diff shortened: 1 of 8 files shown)


prov/efa/src/rdm/rxr_pkt_pool.c
```diff
@@ -175,6 +198,118 @@ int rxr_pkt_pool_grow(struct rxr_pkt_pool *rxr_pkt_pool)
...
+ * @returns
+ * On success, return 0.
+ * On failure, return negative libfabric error code.
+ */
+static inline
+int rxr_pkt_pool_bulk_post_recv_efa(struct rxr_pkt_pool *pkt_pool, int nrecv)
+{
+	int irecv, err;
+	struct rxr_pkt_entry *pkt_entry;
+	struct ibv_recv_wr *head_recv_wr, *prev_recv_wr, *curt_recv_wr, *bad_wr;
```
sunkuamzn, line 249: I think `curt_recv_wr` should be `cur_recv_wr`

sunkuamzn:
This PR also removes `FI_MORE` along the receive path. Only send path left now.

prov/efa/src/rdm/rxr_pkt_pool.c
```diff
@@ -79,34 +95,36 @@ int rxr_pkt_pool_create(struct rxr_ep *ep, size_t size, size_t chunk_cnt,
...
+
+	ret = ofi_bufpool_create_attr(&wiredata_attr, &pool->entry_pool);
 	if (ret) {
-		ofi_bufpool_destroy(pool->localinfo_pool);
 		free(pool);
 		return ret;
 	}
 
+	if (with_sendv_pool) {
+		int sendv_pool_alignment = 8;
```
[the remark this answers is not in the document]
wzamazon: I was experimenting different alignment. changed back to RXR_BUFPOOL_ALIGNMENT in new revision

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -64,20 +64,20 @@ struct rxr_pkt_entry *rxr_pkt_entry_alloc(struct rxr_ep *ep, struct rxr_pkt_pool
...
 
-	pkt_entry = ofi_buf_alloc_ex(pkt_pool->localinfo_pool, &mr);
+	pkt_entry = ofi_buf_alloc_ex(pkt_pool->entry_pool, &mr);
 	if (!pkt_entry)
 		return NULL;
 
+	pkt_entry->send = NULL;
+	if (pkt_pool->sendv_pool) {
+		pkt_entry->send = ofi_buf_alloc(pkt_pool->sendv_pool);
+		assert(pkt_entry->send);
```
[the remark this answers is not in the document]
wzamazon: Because we know that sendv has the same size of entry_pool. So the allocation should not fail. therefore assertion is proper here.

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -74,13 +75,9 @@ struct rxr_pkt_entry *rxr_pkt_entry_alloc(struct rxr_ep *ep, struct rxr_pkt_pool
...
-#ifdef ENABLE_EFA_POISONING
-	memset(pkt_entry, 0, sizeof(*pkt_entry));
-#endif
-
-#ifdef ENABLE_EFA_POISONING
-	memset(pkt_entry->wiredata, 0, ep->mtu_size);
-#endif
+	pkt_entry->send_wr = NULL;
+	if (pkt_pool->efa_send_wr_pool)
+		pkt_entry->send_wr = &pkt_pool->efa_send_wr_pool[ofi_buf_index(pkt_entry)];
```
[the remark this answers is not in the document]
wzamazon: Yes. `ofi_buf_index` has lower cost then `ofi_buf_alloc`.

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -74,13 +75,9 @@ struct rxr_pkt_entry *rxr_pkt_entry_alloc(struct rxr_ep *ep, struct rxr_pkt_pool
 		assert(pkt_entry->send);
 	}
 
-#ifdef ENABLE_EFA_POISONING
```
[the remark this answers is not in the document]
wzamazon: The intention of memory poisoning is that when a new buffer is allocated, the buffer is poisoned, and if the code did not properly initialize some field in the buffer. The code will fail. Set the whole region to 0 defeats the purpose of memory poison.

prov/efa/src/rdm/rxr_pkt_entry.h
```diff
@@ -219,10 +218,8 @@ struct rxr_pkt_entry {
 	struct rxr_pkt_sendv *send;
 
 	/** @brief Work request struct used by rdma-core */
-	union {
-		struct efa_send_wr send_wr;
-		struct efa_recv_wr recv_wr;
-	};
+	struct efa_send_wr *send_wr;
+	struct efa_recv_wr recv_wr;
```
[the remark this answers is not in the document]
wzamazon: With this change, rxr_pkt_entry is not exactly 64 bytes. Some follow up commits will make it 64 bytes again, and at that time we can add the static assert back again.

prov/efa/src/rdm/rxr_pkt_pool.c
```diff
@@ -151,8 +180,13 @@ int rxr_pkt_pool_grow(struct rxr_pkt_pool *rxr_pkt_pool)
...
 void rxr_pkt_pool_destroy(struct rxr_pkt_pool *pkt_pool)
 {
-	ofi_bufpool_destroy(pkt_pool->entry_pool);
+	if (pkt_pool->entry_pool)
+		ofi_bufpool_destroy(pkt_pool->entry_pool);
+
 	if (pkt_pool->sendv_pool)
 		ofi_bufpool_destroy(pkt_pool->sendv_pool);
+
+	free(pkt_pool->efa_send_wr_pool);
```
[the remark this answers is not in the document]
wzamazon: free can be applied to NULL. 

The following is from manpage of `free`:

```
 The free() function frees the memory space pointed to by ptr, which must have been returned by a previous call to malloc(), calloc() or realloc().  Otherwise, or if free(ptr) has already been called before,
       undefined behavior occurs.  If ptr is NULL, no operation is performed.
```

prov/efa/src/rdm/rxr_pkt_pool.c
wzamazon: fixed in new revision.

prov/efa/src/rdm/rxr_pkt_entry.c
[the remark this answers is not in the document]
sunkuamzn: Makes sense

prov/efa/src/rdm/rxr_pkt_entry.c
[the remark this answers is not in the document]
sunkuamzn: In that case, we should set the buffer to the poison value `0xDEADBEEF` instead of 0, right? If we remove the `memset` completely, then the buffer is uninitialized..

prov/efa/src/rdm/rxr_pkt_pool.c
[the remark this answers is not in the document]
sunkuamzn: My concern was calling `free` when `send_wr_pool` wasn't allocated. You have a check for the `recv_wr_pool` right below. `rxr_pkt_pool` is allocated with `calloc`, so checking for NULL is same as checking for allocation.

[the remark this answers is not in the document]
wzamazon: > My concern was calling free when send_wr_pool wasn't allocated.

When `send_wr_pool` is not allocated. It is NULL, and free(NULL) is ok.

> You have a check for the recv_wr_pool right below. 

I see. That is unnecessary, and should be removed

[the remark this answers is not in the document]
wzamazon: Remove the check on `if (pool->recv_wr_pool)`

prov/efa/src/rdm/rxr_pkt_entry.c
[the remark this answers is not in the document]
wzamazon: The buffer was poisoned when they are released in `rxr_pkt_entry_release`.

But I agree poison during allocation make more sense.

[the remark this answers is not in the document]
wzamazon: added a separate commit for that.

prov/efa/src/rdm/rxr_pkt_entry.h
```diff
@@ -216,27 +215,21 @@ struct rxr_pkt_entry {
...
 	/** @brief Work request struct used by rdma-core */
-	union {
-		struct efa_send_wr send_wr;
-		struct efa_recv_wr recv_wr;
-	};
+	struct efa_send_wr *send_wr;
 
 	/** @brief buffer that contains data that is going over wire */
-	char *wiredata;
+	char wiredata[0];
```
sunkuamzn, line 224: Still missing a comment or static assert saying that `sizeof(rxr_pkt_entry)` should be 64 bytes

Also need to add a pad for debug mode so that `sizeof(rxr_pkt_entry)` is 128 bytes

prov/efa/src/rdm/rxr_pkt_pool.c
```diff
@@ -115,24 +184,144 @@ int rxr_pkt_pool_grow(struct rxr_pkt_pool *rxr_pkt_pool)
...
+	for (irecv = 0; irecv < nrecv; ++irecv) {
+		pkt_entry = rxr_pkt_entry_alloc(pkt_pool->ep, pkt_pool, RXR_PKT_FROM_EFA_RX_POOL);
+		if (pkt_entry == NULL) {
+			EFA_WARN(FI_LOG_CQ, "allocating EFA rx pakcet entry failed!");
+			return -FI_ENOBUFS;
+		}
+
+		pkt_entry->x_entry = NULL;
+
+		cur_recv_wr = &pkt_pool->efa_recv_wr_pool[irecv].wr;
```
sunkuamzn, line 263: Actually, shouldn't this be `cur_recv_wr = &pkt_pool->efa_recv_wr_pool[ofi_buf_index(pkt_entry)].wr`?

My understanding is as follows: we bulk post recv packets when `rxr_ep_progress_internal` is called for the first time. Then, every time we release a RX packet, we increment `ep->efa_rx_pkts_to_post`. In the next call to `rxr_ep_progress_internal`, we post `ep->efa_rx_pkts_to_post` packets and set `ep->efa_rx_pkts_to_post = 0`.

So when we call `rxr_pkt_entry_alloc` on line 255 here, the returned memory can be from anywhere in the RX packet pool. It's not guaranteed that the first call returns a pointer to the first memory chunk.

prov/efa/src/rdm/rxr_pkt_pool.c
```diff
@@ -115,24 +184,144 @@ int rxr_pkt_pool_grow(struct rxr_pkt_pool *rxr_pkt_pool)
...
+	{
+		struct ibv_recv_wr *work = head_recv_wr;
+		while (work) {
+			efa_tracepoint_wr_id_post_recv((void *) work->wr_id);
+			work = work->next;
+		}
+	}
+#endif
+
+	err = ibv_post_recv(pkt_pool->ep->base_ep.qp->ibv_qp, &pkt_pool->efa_recv_wr_pool[0].wr, &bad_wr);
```
sunkuamzn, line 291: Why is `&pkt_pool->efa_recv_wr_pool[0].wr` acceptable? It used to be `head_recv_wr`

sunkuamzn (changes requested):

prov/efa/src/rdm/rxr_pkt_pool.c
wzamazon: `head_recv_wr` is a pointer pointing to `&pkt_pool->efa_recv_wr_pool[0].wr`

prov/efa/src/rdm/rxr_pkt_pool.c
wzamazon: the `recv_wr`  memory does not have to be adjacent to the actual packet pool.

`recv_wr` is used to store information of the buffer to be posted, then passed to `ibv_post_recv`. It is not necessary that `recv_wr` be part of the `rxr_pkt_entry`

prov/efa/src/rdm/rxr_pkt_entry.h
wzamazon: With this change `sizeof(rxr_pkt_entry)` is not 64 bytes. (it is 80)

There will follow up commits to make `sizeof(rxr_pkt_entry)` . Once that commits is in, I will add the `static_assert`.

I will add some todo item.

sunkuamzn: OK. We can't merge until you change this to 64 bytes, right? If it's currently not 64 bytes, how are you able to tell that the performance is good?

wzamazon: This PR already improve performance, and is also a sizeable change. so I think we should merge it. Meanwhile, I am active working on bring done the header size.

prov/efa/src/rdm/rxr_pkt_pool.c
wzamazon: You are right. I initially thought the memory of `recv_wr` can be immediately reused after call to `ibv_post_recv`, because this is how efa provider in rdma-core is implemented.

However, after checking the man page of rdma-core, it clearly states that buffers of `recv_wr` can not be reused util we get the completion.

https://github.com/linux-rdma/rdma-core/blob/6644617f085050cff4d583546007ffe96a1a1697/libibverbs/man/ibv_post_recv.3#L52

wzamazon: So we will need to allocate recv-wr memory and release the memory when we get completion

sunkuamzn: Thanks. I'm still not sure why this wasn't caught by CI.

rdma-core is just copying the buffer address we send in ibv_recv_wr ([link](https://github.com/linux-rdma/rdma-core/blob/c18417c56773fd311794ed44b7a2a31cd44c893c/providers/efa/verbs.c#L2226)). In line 277, you are setting the buffer address to a memory region that's potentially still in use.. https://github.com/ofiwg/libfabric/blob/6a2eb1713bf12a890b48543bf93447056e7c79cb/prov/efa/src/rdm/rxr_pkt_pool.c#L279

wzamazon: `wiredata` is part of `rxr_pkt_entry` and is not released until `rxr_pkt_entry_release` is called.

`wiredata` is memory is not part of `recv_wr`.

`sg_list` is part of `recv_wr`, it is used in  [link](https://github.com/linux-rdma/rdma-core/blob/c18417c56773fd311794ed44b7a2a31cd44c893c/providers/efa/verbs.c#L2226) on the fly.

When `ibv_post_recv` is returned, `recv_wr` is not used (for rdma-core's EFA provider)

wzamazon: @sunkuamzn  I was not able to implement that change without impacting performance. So I dropped the 3 patches that implement `efa_recv_wr_pool`. I will do that in follow up patches.

wzamazon: @sunkuamzn  I added a commit to ensure `rxr_pkt_entry` is multiple of 64.

sunkuamzn (approved):
It would be great if we could document the results of your performance tests for future reference

wzamazon:
>  It would be great if we could document the results of your performance tests for future reference

The performance change would show up in our internal dashboard (hopefully)

PR #8494, opened 2023-02-14
prov/efa: Introduce RDMA Write Feature to EFA

lrbison:
Update documentation to describe RDMA-Write based data transfer to the EFA provider.  This is implemented as an Extra Feature as described in the EFA Protocol version 4.

Update the EFA handshake protocol to exchange the RDMA-Write feature flag.

Add a function to detect if RDMA-Write is supported by the domain endpoint.  Currently, it always returns false.


prov/efa/src/efa_domain.h
```diff
@@ -125,6 +125,20 @@ bool efa_domain_support_rdma_read(struct efa_domain *domain)
...
+ *
+ * @param[in]	domain	struct efa_domain
+ *
+ * @return: true if rdma write is supported. false otherwise.
+ */
+static inline
+bool efa_domain_support_rdma_write(struct efa_domain *domain)
+{
+	/* Not yet supported in rdma-core. */
+	return 0;
```
wzamazon, line 139: `return false`.

prov/efa/docs/efa_rdm_protocol_v4.md
```diff
@@ -1445,6 +1449,22 @@ The binary format of a RUNTREAD_RTM packet's mandatory header is listed in table
...
+EFA kernel module and rdma-core) to be met before an endpoint can use the RDMA
+write capability, therefore an endpoint cannot assume the other party supports
+RDMA write.
+
+#### Direct write subprotocol
+
+The direct write subprotocol is the simplest write subprotocol in protocol v4.
+It does not involve a REQ packet.  The workflow is just for the writer to keep
+using RDMA write on the responder. For this protocol, it is not necessary for
+the responder to keep the progress engine running.
```
wzamazon, line 1466: > "It is not necessary for the responder to keep the progress engine running."

This is not true for write. `fi_write`   has a special case that requester can pass FI_REMOTE_CQ_DATA flag, which will cause the responder to get a CQ entry for the write. To support this  feature, the responder need to keep the progress engine running

wzamazon (changes requested):

prov/efa/docs/efa_rdm_protocol_v4.md
lrbison: Is the updated comment accurate?

wzamazon (approved):

PR #8500, opened 2023-02-14
Man fixes for opx provider for 1.17.x

charlesshereda:
Man page fixes relevant to the OPX provider, to be pulled into 1.17.x.

### DIFF
--- a/man/fi_opx.7.md
+++ b/man/fi_opx.7.md
@@ -11,13 +11,12 @@ fi_opx \- The Omni-Path Express Fabric Provider
 
 # OVERVIEW
 
-The OPX provider is a native implementation of the libfabric interfaces
-that makes direct use of Omni-Path fabrics as well as libfabric
-acceleration features.
-The purpose of this provider is to show the scalability and
-performance of libfabric, providing an "extreme scale" development
-environment for applications and middleware using the libfabric API, and
-to support a functional and performant version of MPI on Omni-Path fabrics.
+The *opx* provider is a native libfabric provider suitable for 
+use with Omni-Path fabrics.  OPX features great scalability and
+performance when running libfabric-enabled message layers.  
+OPX requires 3 additonal external development libraries to build: 
+libuuid, libnuma, and the Linux kernel headers.
+
 
 # SUPPORTED FEATURES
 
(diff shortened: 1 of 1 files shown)


man/fi_opx.7.md
```diff
@@ -51,10 +51,11 @@ Additional features
   and *counters*.
 
 Progress 
-: Only *FI_PROGRESS_MANUAL* is supported.
+: *FI_PROGRESS_MANUAL* and *FI_PROGRESS_AUTO* are supported, for best performance, use 
+  *FI_PROGRESS_MANUAL* when possible. *FI_PROGRESS_AUTO* will spawn 1 thread per CQ.
 
 Address vector 
-: Only the *FI_AV_MAP* address vector format is supported.
+: *FI_AV_MAP* and *FI_AV_TABLE* are both supported. *FI_AV_UNSPEC* is default.
```
j-xiong, line 57: The FI_AV_UNSPEC part is unnecessary. Alternatively, something like this is more useful: `if *FI_AV_UNSPEC* is specified, *FI_AV_xxx* is used by default`.

man/man7/fi_opx.7
```diff
@@ -50,11 +50,14 @@ Additional features
 Supported additional features include \f[I]FABRIC_DIRECT\f[R],
 \f[I]scalable endpoints\f[R], and \f[I]counters\f[R].
 .TP
-Progress
-Only \f[I]FI_PROGRESS_MANUAL\f[R] is supported.
+.B Progress
```
j-xiong, line 53: the change to `fi_opx.7` is unnecessary. The actual man page is going to be automatically generated from the corresponding '.md' file.

charlesshereda:
rerun CI

man/fi_opx.7.md
charlesshereda: We'll make this change.

man/man7/fi_opx.7
charlesshereda: Is there something in the *.7 that you see that is out of sync with the *.md file? Because it's our understanding that sometimes the *.7 isn't regenerated when it should be, so it's better to be safe and push modifications to it too.

charlesshereda: We talked about it internally and agree that we should be requesting the regen rather than trying to push this file. I updated the relevant commit and will also do it for the corresponding PR against main.

man/fi_opx.7.md
charlesshereda: The new wording just says that MAP is default without mentioning FI_AV_UNSPEC.

j-xiong (approved):

PR #8502, opened 2023-02-14
Opx upstream for 1.17.x

charlesshereda:
OPX provider fixes and enhancements to be pulled into 1.17.x release candidate.


charlesshereda:
rerun CI

j-xiong:
Intel CI error:
```
runtests.py: error: argument --prov: invalid choice: 'net' (choose from 'verbs', 'tcp', 'udp', 'sockets', 'shm', 'psm3')
```

Looks like the script from `main` is used here @shefty

shefty:
@charlesshereda Intel CI can be ignored for this PR, but I want to see what update to the scripts is needed.  @j-xiong - Can you clarify what you see as the issue?  Are we mixing CI scripts between v1.17.x and main?

j-xiong:
@sheft. Right. Based on the log, the CI is using `runtests.py` from main, which has already removed the net provider from the list of valid providers. The same script from v1.17.x still allows the net provider.

PR #8503, opened 2023-02-15
prov/sm2: Create sm2 provider for development

a-szegel:
Create an experimental sm2 provider to hold the development for the nemesis shared memory algorithm. sm2 is based on the sm provider, but names have been changed and all the protocols besides inject have been removed.  This will allow us to quickly develop and test the new queue mechanism, and we will restore functionality once we are confident it is working.

This will allow us to develop in small, easy to review PR's instead of large code dumps.


a-szegel:
once this is in, I will add this to AWS CI

include/ofi_shm2.h
```diff
@@ -0,0 +1,351 @@
...
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _OFI_SHM_H_
+#define _OFI_SHM_H_
```
j-xiong, line 34: `_OFI_SHM2_H_`?

prov/shm2/configure.m4
```diff
@@ -0,0 +1,91 @@
...
+	      AS_IF([test $dsa_happy -eq 1],
+		    [AC_CHECK_HEADER(linux/idxd.h, [], [dsa_happy=0])])
+
+	      AS_IF([test "x$with_dsa" != "xno" && test -n "$with_dsa" && test $dsa_happy -eq 0 ],
+		    [AC_MSG_ERROR([shm DSA support requested but DSA not available.])])
+
+	      AS_IF([test $dsa_happy -eq 1 && test "x$with_dsa" != "xyes"],
+	            [shm2_CPPFLAGS="$shm2_CPPFLAGS $dsa_CPPFLAGS $numa_LIBS"
+		     shm2_LDFLAGS="$shm2_LDFLAGS $dsa_LDFLAGS $numa_LIBS"])
+	      shm_LIBS="$shm2_LIBS $dsa_LIBS $numa_LIBS"
```
j-xiong, line 79: `shm2_LIBS`

prov/util/src/util_shm2.c
```diff
@@ -0,0 +1,550 @@
+/*
+ * Copyright (c) 2016-2021 Intel Corporation. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
```
wzamazon, line 4: I am not sure that shm and shm2 need different util. Can they share the same util? If not, can we move these to inside shm2?

a-szegel: `smr2_calculate_size_offsets` will change b/c there response queue
`void smr2_cleanup(void)` will change b/c there won't be mutex locks
`struct smr2_region {` region structure definition will likely change 
`int smr2_create()` will change

The new util code is required AFAIK, but will look very different from SHM.

wzamazon: Sounds like these functions is tied to how shm2 provider is implemented, so they should be inside shm2 provider. Right?

aingerson: We originally separated the shm region and map functions into the util code because we thought there would be other providers that might want to use them to implement an internal shm implementation directly. Since no one is using them and we are targeting the peer provider approach for shm offload, I think we can use this opportunity to move them under the shm provider directly and, once we merge the shm2 provider back with the shm provider, remove the util_shm.c code, replacing it with the one inside the provider itself.

a-szegel: done

include/ofi_shm2.h
a-szegel: done

prov/shm2/configure.m4
a-szegel: done

wzamazon (approved):

j-xiong (approved):

shefty:
The provider and file names can be whatever, but if the plan is to merge this back into the shm provider, it would be best if the prefix were the same length (e.g. sm2).  That way, a simple find-replace doesn't throw off the indentation and alignment.

a-szegel:
> The provider and file names can be whatever, but if the plan is to merge this back into the shm provider, it would be best if the prefix were the same length (e.g. sm2). That way, a simple find-replace doesn't throw off the indentation and alignment.

@shefty Switched from shm2 to sm2

j-xiong:
@a-szegel The upper case `SMR2`'s are still there.

a-szegel:
@j-xiong good catch, thank you... fixed :)

shefty:
Be aware that the shm provider is not static, and there will be parallel development on it.

shefty:
What is the plan for merging any changes back?  Our current priority is AI, not HPC, so the GPU paths are the critical performance paths.

a-szegel:
@shefty 

> Our current priority is AI

Ack, All GPU paths will be restored before merging.  This is simplified on purpose to allow rapid development. After we get the basic Nemesis implmentation working, we will add back the optimizations. 

> Be aware that the shm provider is not static, and there will be parallel development on it.

Ack.  We will monitor for relevant PR's and cherry pick them and merge them across while SHM2 is under development.  Would be required if we used a branch too.

> What is the plan for merging any changes back? 

The plan is to get SHM2 fully operational, and supported within EFA via env switch (something like FI_EFA_USE_SM2=false).  This will allow us to invite some customers to use it for beta testing.  Once we have proven that it is stable, we plan to change the default of FI_EFA_USE_SM2 to default to true.  After we are confident that it is working, and we are happy with its performance, I would like to do a replace `mv sm2 shm`, and a rename `git grep -lz prov | xargs -0 sed -i'' -e 's/SM2/SHM/g'`.

shefty:
Is there any sane way of doing these changes within the existing provider, or does the layout of the shared region need to change?

aingerson:
@shefty The shm region format needs to be redone hence the separate development. Thoughts on doing this in a separate branch instead of a duplicate provider? That would allow us to cherry-pick commits more easily into the new development.

shefty:
I think it needs to be a separate provider to align with EFA's plan, and I agree that that approach makes sense.  My biggest concern is that we don't lose changes to shm when picking those into sm2.  I had it easier with tcp and net, since I basically just stopped working on tcp.  But I don't expect that to be the case here.  So, I think the goal is "let's do this fast". :)

include/ofi_sm2.h
```diff
@@ -0,0 +1,351 @@
+/*
```
aingerson, line 1: If we're moving the util_shm code inside the provider then we can move the ofi_smr.h code into smr.h (or a separate header). It was only in the common includes because it was in the util folder

j-xiong: While we're here maybe also rename the file `util_sm2.c` to `sm2_util.c` for a consistent naming scheme.

aingerson:
While I think the separate providers would satisfy your use case of wanting the customers to test out the implementation, I think this is also fairly easily done with the branching as well. You would just need to provide a second build for anyone to point to if they want to test out the new implementation. I think both are a similar amount of work in terms of having someone test it. I think the branch helps keep conflicts isolated to a more sequential/per change basis and I worry that even though the goal is to make the second provider's lifetime as short as possible, that there is going to be a lot of conflicts at the end in order to move the new features back to the updated provider. For example, there are 3 PRs open right now that make pretty significant changes to the provider. It would be easier to make sure those changes are picked up one by one as opposed to all at the end.

include/ofi_sm2.h
aingerson: ^also that!

shefty:
A branch means having 2 versions of libfabric (but same version number), each with the same reported provider, but completely different implementations.  That's very hard for an end-user to figure out what they're using, and doesn't work at all for apps that statically link against the library.  The problem I see is that this fork is already dropping a lot of the provider, so cherry-picking changes between them seems impossible.

j-xiong (approved):

a-szegel:
@aingerson @shefty 

Any other actions required for this PR?

a-szegel:
last push was include/ofi_prov.h change that I realized I forgot.

include/ofi_sm2.h
a-szegel: done

include/ofi_prov.h
```diff
@@ -226,6 +226,17 @@ SHM_INI ;
 #  define SHM_INIT NULL
 #endif
 
+#if (HAVE_SM2) && (HAVE_SM2_DL)
+#  define SHM_INI FI_EXT_INI
+#  define SHM_INIT NULL
```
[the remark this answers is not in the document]
a-szegel: done

shefty (approved):
I think we just more forward and need to check on merge back that updates don't get lost.

aingerson:
@wzamazon Yes, that's fine

PR #8516, opened 2023-02-16
[v1.17.x] update Intel CI script

shefty:
updates to shm, util, common


aingerson:
The first commit (9f7bf96) was a mistake that was introduced with a new change and I fixed in a later commit (96e4bd0). I already cherry picked the bug fix so this is adding the bug back in. In other words remove that commit :)

shefty:
Added fix for Intel CI script.

shefty:
Removed everything but the CI fix... which is apparently needed as merged prior to testing the other patches.

PR #8522, opened 2023-02-16
prov/opx: Add build check for HFI1_CAP_TID_RDMA

shefty:


shefty:
@charlesshereda

charlesshereda:
Sorry for the slow response, I'm testing it but it's not working correctly for me in the case where HFI1_CAP_TID_RDMA is correctly declared. Need a little more time to figure it out.

shefty:
It doesn't look like AC_CHECK_DECL works with a define, plus I didn't add the include file to check.  (Adding the include file doesn't work.)

charlesshereda:
Ok, I was just in the process of adding the include file, so never mind to that, but I'll see if I can figure it out.

prov/opx/configure.m4
```diff
@@ -121,6 +121,9 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 		    [],
 		    [opx_happy=0])
 
+		AS_IF([test $opx_happy -eq 1],[
+			AC_CHECK_DECL([HFI1_CAP_TID_RDMA], [], [opx_happy=0])
```
charlesshereda, line 125:
```suggestion
						AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
				[[#include <rdma/hfi/hfi1_user.h>]],
				[[
					#if HFI1_USER_SWMAJOR != 6
					#error "incorrect version of hfi1_user.h"
					#endif
				]])],
				[AC_MSG_NOTICE([hfi1_user.h version 6.x found... yes])],
				[
				AC_MSG_NOTICE([hfi1_user.h version 6.x found... no])
				opx_happy=0
				])
```

charlesshereda: Can you try with that AC_COMPILE_IFELSE instead of the AC_CHECK_DECL and see if it skips the OPX build for your failing system? It depends on the HFI1_USER_SWMAJOR in that file.

If not, we'll have to check for the #define of HFI1_CAP_TID_RDMA inside there instead, but it should still work.

charlesshereda (changes requested):

PR #8523, opened 2023-02-16
[v1.17.x] prov/efa: fix double free in rxr endpoint init

wenduwan:
Manually apply a patch to prevent rdm_info from being freed twice. Cherry-picking the fix on main branch results in conflict, and thus we do a simpler fix instead.

### DIFF
--- a/prov/efa/src/rxr/rxr_ep.c
+++ b/prov/efa/src/rxr/rxr_ep.c
@@ -2465,9 +2465,9 @@ static int rxr_ep_setup_device_cq(struct rxr_ep *rxr_ep)
 int rxr_endpoint(struct fid_domain *domain, struct fi_info *info,
 		 struct fid_ep **ep, void *context)
 {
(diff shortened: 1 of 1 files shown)


wzamazon (approved):

lrbison (approved):
lgtm

qkoziol (changes requested):
Although this is technically correct today, it would be better if the rdm_info variable was initialized to NULL when it was declared.  That way if the code is restructured and the goto operations are changed, the code will be safer.   (Probably should set rxp_ep to NULL also, and protect freeing it also)

wenduwan:
> 

Thank you. The suggestion makes sense to me. Updated commit accordingly.

qkoziol (approved):
Thanks for updating!

lrbison (approved):

wzamazon (approved):

PR #8525, opened 2023-02-16
prov/efa: query 128-bytes aligned writing capacity

jtamzn:
Some EFA device may support in-order write for each 128-bytes aligned piece over RDMA write. This PR allows applications to enable this feature via `fi_setopt`.

### DIFF
--- a/README
+++ b/README
@@ -1,4 +1,4 @@
-Version @VERSION@
+ Version @VERSION@
 Released on @DATE@
 
 Introduction


prov/efa/configure.m4
```diff
@@ -13,6 +13,8 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	efa_h_enable_poisoning=0
 	# Determine if efa device supports extensible CQ with access to additional fields from WC
 	efadv_support_extended_cq=1
+	# Determine if efa device support in-order write at 128 bytes aligned boundary.
+	efa_support_ll128=1
```
lrbison, line 17: small nit: `ll123` is visually confusing.  Can we use `LL128` ?

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1040,6 +1040,16 @@ static int rxr_ep_setopt(fid_t fid, int level, int optname,
 		if (ret)
 			return ret;
 		break;
+	case FI_OPT_EFA_SUPPORT_LL128:
+#if HAVE_EFA_LL128_SUPPORT
```
lrbison, line 1044: My vote would be to move this CPP logic inside of ibv_query_qp_data_in_order

include/rdma/fi_ext.h
```diff
@@ -70,6 +70,7 @@ extern "C" {
 /* negative options are provider specific */
 enum {
        FI_OPT_EFA_RNR_RETRY = -FI_PROV_SPECIFIC_EFA,
+	   FI_OPT_EFA_SUPPORT_LL128
```
lrbison, line 73: Indention.  Make sure tab size=8

lrbison: Does changing this enumeration mean we need to bump some libfabric version number somewhere?

qkoziol: > Does changing this enumeration mean we need to bump some libfabric version number somewhere?

This should be a backwardly compatible change (i.e. old code won't ask for this feature), so it should be safe to not bump the version #.  (Unless libfabric has a much stronger policy around interface changes like this)

qkoziol (approved):
lgtm, as long as no version # bump is required

prov/efa/configure.m4
wzamazon: ll128 is a term used by NCCL. We should use the term from rdma-core,

wzamazon (changes requested):

wzamazon:
I think it should be verbatim, e.g. DATA_IN_ORDER_ALIGNED_128_BYTES

wzamazon:
write is confusing because this property support both send and write.

include/rdma/fi_ext.h
wzamazon: As we discussed previously, we need 2 options, one for send, another for RMA. Also do not use LL128, something like `EFA_SEND_DATA_IN_ORDER_ALIGNED_128_BYTE` and `EFA_RMA_DATA_IN_ORDER_ALIGNED_128_BYTE`

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1040,6 +1040,19 @@ static int rxr_ep_setopt(fid_t fid, int level, int optname,
 		if (ret)
 			return ret;
 		break;
+	case FI_OPT_EFA_SEND_IN_ORDER_128_BYTES:
+		/* For future use only, currently we do not support in-order 
+		transfer in send/recv */
+		return -FI_EOPNOTSUPP;
+	case FI_OPT_EFA_RMA_IN_ORDER_128_BYTES:
```
wzamazon, line 1047: I did not notice `ibv_query_qp_data_in_order` work on op. so this should be `FI_OPT_EFA_WRITE_IN_ORDER_128_BYTES

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1040,6 +1040,19 @@ static int rxr_ep_setopt(fid_t fid, int level, int optname,
...
+#if HAVE_EFA_DATA_IN_ORDER_ALIGNED_128_BYTES
+		ret = ibv_query_qp_data_in_order(rxr_ep->base_ep.qp, 
+						IBV_WR_RDMA_WRITE,
+						IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES);
+		if (ret)
```
wzamazon, line 1052: better wrap `#if` logic in a separate function `efa_base_ep_support_write_in_order_128_bytes()` and return `true` or `false`.

wzamazon (changes requested):

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1040,6 +1040,19 @@ static int rxr_ep_setopt(fid_t fid, int level, int optname,
 		if (ret)
 			return ret;
 		break;
+	case FI_OPT_EFA_SEND_IN_ORDER_128_BYTES:
```
wzamazon, line 1043: better make it `SENDRECV_IN_ORDER_ALIGNED_128_BYTES`

prov/efa/src/rdm/rxr_ep.c
wzamazon: Also added `ALIGNED`

prov/efa/src/efa_base_ep.c
```diff
@@ -341,3 +341,21 @@ int efa_base_ep_getname(fid_t fid, void *addr, size_t *addrlen)
 	memcpy(addr, &base_ep->src_addr, EFA_EP_ADDR_LEN);
 	return 0;
 }
+
+bool efa_base_ep_support_op_in_order_aligned_128_bytes(struct efa_base_ep *base_ep, enum ibv_wr_opcode op) {
+#if HAVE_EFA_DATA_IN_ORDER_ALIGNED_128_BYTES
+	int ret;
+	switch(op) {
+	case IBV_WR_SEND:
+		break;
```
wzamazon, line 350: `return false` here is easier to understand.

prov/efa/src/efa_base_ep.c
```diff
@@ -341,3 +341,21 @@ int efa_base_ep_getname(fid_t fid, void *addr, size_t *addrlen)
...
+	case IBV_WR_RDMA_WRITE:
+		ret = ibv_query_qp_data_in_order(base_ep->qp,
```
wzamazon, line 352: It seems we can simply `return ibv_query_qp_data_in_order`

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1040,6 +1040,16 @@ static int rxr_ep_setopt(fid_t fid, int level, int optname,
 		if (ret)
 			return ret;
 		break;
+	case FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES:
+		ret = efa_base_ep_support_op_in_order_aligned_128_bytes(&rxr_ep->base_ep, IBV_WR_SEND);
```
wzamazon, line 1044: This is not right. The fact that base_ep support send 128 bytes does not guarantee rdm endpoint support 128 bytes in order send. Because rdm endpoint use device's send, read, localread, and gdrcopy to implement send/receive. To support 128 in_order sendrecv, all of the components must support. Let's simply return FI_EOPNOTSUPP here.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1040,6 +1040,16 @@ static int rxr_ep_setopt(fid_t fid, int level, int optname,
...
+		if (ret)
+			break;
+		return -FI_EOPNOTSUPP;
+	case FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES:
+		ret = efa_base_ep_support_op_in_order_aligned_128_bytes(&rxr_ep->base_ep, IBV_WR_RDMA_WRITE);
```
wzamazon, line 1049: it is easier to understand that

```
if (!support_op_in_order_aligned_128_bytes(&rxr_ep->base_ep, IBV_WR_RDMA_WRITE)
    return -FI_EOPNOTSUPP;

File: prov/efa/src/efa_base_ep.c [hunk 11]
```diff
@@ -341,3 +341,21 @@ int efa_base_ep_getname(fid_t fid, void *addr, size_t *addrlen)
 	memcpy(addr, &base_ep->src_addr, EFA_EP_ADDR_LEN);
 	return 0;
 }
+
+bool efa_base_ep_support_op_in_order_aligned_128_bytes(struct efa_base_ep *base_ep, enum ibv_wr_opcode op) {
```
wzamazon, line 345: `{` should start at the beginning of the next line

prov/efa/src/efa_base_ep.c
```diff
@@ -341,3 +341,21 @@ int efa_base_ep_getname(fid_t fid, void *addr, size_t *addrlen)
 	memcpy(addr, &base_ep->src_addr, EFA_EP_ADDR_LEN);
 	return 0;
 }
+
```
wzamazon, line 344: add a doxygen style function header

wzamazon (changes requested):

prov/efa/src/rdm/rxr_ep.c
wzamazon: This should be obvious from the function name.

prov/efa/src/efa_base_ep.c
wzamazon: > yep, we can, but would like to keep a pattern for future.

Future code can also use `return`.

wzamazon (approved):

bwbarrett (changes requested):
There really needs to be a section in the documentation for EFA-specific options.  And these two options really need clear definitions.

wzamazon:
This PR would depend on https://github.com/linux-rdma/rdma-core/pull/1312

PR #8527, opened 2023-02-16
prov/efa: Check FI_HMEM_CUDA_ENABLE_XFER per domain

vidsouza:
Prior to this change FI_HMEM_CUDA_ENABLE_XFER was checked only once when the provider is initialized. Due to this different plugins/libraries setting up the environment for various applications cannot modify the value of this environment variable.

This patch provides this possibility by checking the value of FI_HMEM_CUDA_ENABLE_XFER once per domain.


vidsouza:
This PR is related to [prov/efa: Check FI_EFA_USE_DEVICE_RDMA per domain](url) PR

PR #8528, opened 2023-02-16
prov/efa: Implement RDMA write for EFA

lrbison:
Implement rxr_op_entry_post_remote_write, which can be used by rxr_rma_writemsg to post writes using rdma-core.

A single FI_WRITE may generate multiple RDMA writes, and completions are tracked in rxr_pkt_handle_write_completion. When the entire write is complete, rxr_pkt_handle_rma_completion is called.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_peer.h
+++ b/prov/efa/src/rdm/efa_rdm_peer.h
@@ -92,6 +92,12 @@ struct efa_rdm_peer {
 	int64_t num_read_msg_in_flight;
 };
 
+/**
+ * @brief check for peer's RDMA_READ support, assuming HANDSHAKE has already occurred
+ *
+ * @param[in] peer	A peer which we have already received a HANDSHAKE from
+ * @return bool		The peer's RDMA_READ support
+ */
 static inline
 bool efa_rdm_peer_support_rdma_read(struct efa_rdm_peer *peer)
 {
@@ -101,6 +107,23 @@ bool efa_rdm_peer_support_rdma_read(struct efa_rdm_peer *peer)
 	 */
 	return (peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED) &&
 	       (peer->extra_info[0] & RXR_EXTRA_FEATURE_RDMA_READ);
+}
+
+/**
+ * @brief check for peer's RDMA_WRITE support, assuming HANDSHAKE has already occurred
+ *
+ * @param[in] peer	A peer which we have already received a HANDSHAKE from
+ * @return bool		The peer's RDMA_WRITE support
+ */
+static inline
+bool efa_rdm_peer_support_rdma_write(struct efa_rdm_peer *peer)
+{
+	/* RDMA WRITE is an extra feature defined in version 4 (the base version).
+	 * Because it is an extra feature, an EP will assume the peer does not support
+	 * it before a handshake packet was received.
+	 */
+	return (peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED) &&
+	       (peer->extra_info[0] & RXR_EXTRA_FEATURE_RDMA_WRITE);
 }
 
 static inline
@@ -116,11 +139,36 @@ bool efa_rdm_peer_support_delivery_complete(struct efa_rdm_peer *peer)
 	       (peer->extra_info[0] & RXR_EXTRA_FEATURE_DELIVERY_COMPLETE);
 }
 
+/**
+ * @brief determine if both peers support RDMA read
+ *
+ * This function can only return true if a handshake packet has already been
+ * exchanged, and the peer set the RXR_EXTRA_FEATURE_RDMA_READ flag.
+ * @params[in]		ep		Endpoint for communication with peer
+ * @params[in]		peer		An EFA peer
+ * @return		boolean		both self and peer support RDMA read
+ */
 static inline
 bool efa_both_support_rdma_read(struct rxr_ep *ep, struct efa_rdm_peer *peer)
 {
 	return efa_domain_support_rdma_read(rxr_ep_domain(ep)) &&
 	       (peer->is_self || efa_rdm_peer_support_rdma_read(peer));
+}
+
+/**
+ * @brief determine if both peers support RDMA write
+ *
+ * This function can only return true if a handshake packet has already been
+ * exchanged, and the peer set the RXR_EXTRA_FEATURE_RDMA_WRITE flag.
+ * @params[in]		ep		Endpoint for communication with peer
+ * @params[in]		peer		An EFA peer
+ * @return		boolean		both self and peer support RDMA write
+ */
+static inline
+bool efa_both_support_rdma_write(struct rxr_ep *ep, struct efa_rdm_peer *peer)
+{
+	return efa_domain_support_rdma_write(rxr_ep_domain(ep)) &&
+	       (peer->is_self || efa_rdm_peer_support_rdma_write(peer));
 }
 
 /**
--- a/prov/efa/src/rdm/rxr_env.h
+++ b/prov/efa/src/rdm/rxr_env.h
@@ -69,6 +69,7 @@ struct rxr_env {
 	size_t shm_cq_read_size;
 	size_t efa_max_gdrcopy_msg_size;
 	size_t efa_read_segment_size;
+	size_t efa_write_segment_size;
 	/* If first attempt to send a packet failed,
 	 * this value controls how many times firmware
 	 * retries the send before it report an RNR error
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -51,6 +51,8 @@
 #include "rxr_tp.h"
 #include "rxr_cntr.h"
 
+void recv_rdma_with_imm_completion(struct rxr_ep *ep, int32_t imm_data, uint64_t flags);
+
 struct efa_ep_addr *rxr_ep_raw_addr(struct rxr_ep *ep)
 {
 	return &ep->base_ep.src_addr;
@@ -703,13 +705,13 @@ static int rxr_ep_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
  * For a given peer, trigger a handshake packet and determine if both peers
  * support rdma read.
  *
- * @param[in]	ep	rxr_ep
- * @param[in]	addr	remote address
- * @param[in]	peer	remote peer
- * @return 	1 if supported, 0 if not, negative errno on error
+ * @param[in,out]	ep	rxr_ep
+ * @param[in]		addr	remote address
+ * @param[in,out]	peer	remote peer
+ * @return 		1 if supported, 0 if not, negative errno on error
  */
-int rxr_ep_determine_rdma_support(struct rxr_ep *ep, fi_addr_t addr,
-				  struct efa_rdm_peer *peer)
+int rxr_ep_determine_rdma_read_support(struct rxr_ep *ep, fi_addr_t addr,
+				       struct efa_rdm_peer *peer)
 {
 	int ret;
 
@@ -723,6 +725,42 @@ int rxr_ep_determine_rdma_support(struct rxr_ep *ep, fi_addr_t addr,
 	}
 
 	if (!efa_both_support_rdma_read(ep, peer))
+		return 0;
+
+	return 1;
+}
+
+/**
+ * @brief determine if both peers support rdma write.
+ *
+ * If no prior communication with the given peer, and we support RDMA WRITE
+ * locally, then trigger a handshake packet and determine if both peers
+ * support rdma write.
+ *
+ * @param[in,out]	ep	rxr_ep
+ * @param[in]		addr	remote address
+ * @param[in,out]	peer	remote peer
+ * @return 		1 if supported, 0 if not, negative errno on error
+ */
+int rxr_ep_determine_rdma_write_support(struct rxr_ep *ep, fi_addr_t addr,
+					struct efa_rdm_peer *peer)
+{
+	int ret;
+
+	/* no need to trigger handshake if we don't support RDMA WRITE locally */
+	if (!efa_domain_support_rdma_write(rxr_ep_domain(ep)))
+		return false;
+
+	if (!peer->is_local) {
+		ret = rxr_pkt_trigger_handshake(ep, addr, peer);
+		if (OFI_UNLIKELY(ret))
+			return ret;
+
+		if (!(peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED))
+			return -FI_EAGAIN;
+	}
+
+	if (!efa_both_support_rdma_write(ep, peer))
 		return 0;
 
 	return 1;
@@ -1639,16 +1677,16 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			break;
 		}
 
-		pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
-
 		switch (ibv_wc_read_opcode(ep->ibv_cq_ex)) {
 		case IBV_WC_SEND:
 #if ENABLE_DEBUG
 			ep->send_comps++;
 #endif
+			pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
 			rxr_pkt_handle_send_completion(ep, pkt_entry);
 			break;
 		case IBV_WC_RECV:
+			pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
 			pkt_entry->addr = efa_av_reverse_lookup_rdm(efa_av, ibv_wc_read_slid(ep->ibv_cq_ex),
 								ibv_wc_read_src_qp(ep->ibv_cq_ex), pkt_entry);
 
@@ -1662,6 +1700,15 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 #if ENABLE_DEBUG
 			ep->recv_comps++;
 #endif
+			break;
+		case IBV_WC_RDMA_WRITE:
+			pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
+			rxr_pkt_handle_rma_completion(ep, pkt_entry);
+			break;
+		case IBV_WC_RECV_RDMA_WITH_IMM:
+			recv_rdma_with_imm_completion(ep,
+				ibv_wc_read_imm_data(ep->ibv_cq_ex),
+				FI_REMOTE_CQ_DATA | FI_RMA | FI_REMOTE_WRITE );
 			break;
 		default:
 			EFA_WARN(FI_LOG_EP_CTRL,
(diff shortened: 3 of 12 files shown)


bwbarrett:
Did I miss the code to support `fi_writedata()`?

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -322,6 +326,15 @@ ssize_t rxr_rma_post_write(struct rxr_ep *ep, struct rxr_op_entry *tx_entry)
 	if (peer->is_local && ep->use_shm_for_tx)
 		return rxr_rma_post_shm_write(ep, tx_entry);
 
+	if (!(tx_entry->fi_flags & FI_INJECT) && rxr_ep_determine_rdma_write_support(ep, tx_entry->addr, peer)) {
+		/* RDMA_WRITE does not support FI_INJECT, because device may
+		 * need to re-send data and FI_INJECT allows user to re-use
+		 * these buffers immediately
+		 */
+		/* TODO: do we ever use a SEND instead of a write based on total_len? */
+		rxr_op_entry_post_remote_write(tx_entry);
```
wzamazon, line 335: return here.

lrbison:
> Did I miss the code to support `fi_writedata()`?

No, thank you, that's a bug as this PR stands right now.  There is another change I need to do to support writedata, so I think I'll make sure we continue our existing send-based implementation for writedata in this PR, and then have another PR that enables writedata via an `RDMA_WRITE_WITH_IMM`

bwbarrett:
> > Did I miss the code to support `fi_writedata()`?
> 
> No, thank you, that's a bug as this PR stands right now. There is another change I need to do to support writedata, so I think I'll make sure we continue our existing send-based implementation for writedata in this PR, and then have another PR that enables writedata via an `RDMA_WRITE_WITH_IMM`

I'm fine with splitting `fi_write` and `fi_writedata` implementations, but the teams that will use write are planning on using `fi_wrtiedata`.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -1357,6 +1387,130 @@ int rxr_op_entry_post_remote_read(struct rxr_op_entry *op_entry)
...
+				 * which means memory registration failed.
+				 * return -FI_EAGAIN here will cause user to run progress
+				 * engine, which will cause some memory registration
+				 * in MR cache to be released.
+				 */
+				return -FI_EAGAIN;
+			}
+		}
+
+		if (use_shm)
```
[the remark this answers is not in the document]
lrbison: done

prov/efa/src/rdm/rxr_rma.c
lrbison: done, reworked some of this logic too to fix FI_REMOTE_CQ_DATA case too

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -307,7 +311,46 @@ ssize_t rxr_rma_read(struct fid_ep *ep, void *buf, size_t len, void *desc,
 	return rxr_rma_readv(ep, &iov, &desc, 1, src_addr, addr, key, context);
 }
 
-/* rma_write functions */
+/**
+ * @brief Decide if we should issue this WRITE using rdma-core, or via emulation.
+ *
+ * @param ep		The endpoint.
+ * @param tx_entry	The op_entry context for this write.
+ * @param peer		The peer we will be writing to.
```
darrylabbate, line 319: nit: can we add the `[in]`/`[out]` annotations to the params here (and elsewhere)?

prov/efa/src/rdm/rxr_env.h
```diff
@@ -70,6 +70,7 @@ struct rxr_env {
 	size_t shm_cq_read_size;
 	size_t efa_max_gdrcopy_msg_size;
 	size_t efa_read_segment_size;
+	size_t efa_write_segment_size;
```
lrbison, line 73: should these be renamed to `efa_rdma_segment_size` ?

prov/efa/src/rdm/rxr_op_entry.h
```diff
@@ -199,6 +199,11 @@ struct rxr_op_entry {
 	uint64_t bytes_read_submitted;
 	uint64_t bytes_read_total_len;
 	uint64_t bytes_read_offset;
+
+	/* counters for rma writes */
+	uint64_t bytes_write_completed;
+	uint64_t bytes_write_submitted;
+	uint64_t bytes_write_total_len;
```
lrbison, line 206: should these be renamed to `bytes_rdma_*` ?

prov/efa/src/rdm/rxr_pkt_cmd.c
```diff
@@ -798,6 +798,7 @@ void rxr_pkt_handle_send_completion(struct rxr_ep *ep, struct rxr_pkt_entry *pkt
 		rxr_pkt_handle_eor_send_completion(ep, pkt_entry);
 		break;
 	case RXR_RMA_CONTEXT_PKT:
+		// how do we get an RMA Context Packet in a Send completion?
```
lrbison, line 801: I added this comment as a reminder to ask others, how do we end up at this case statement?  I want to make sure we handle it correctly in `rxr_pkt_handle_rma_completion` which I made some changes too.

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -122,6 +122,8 @@ size_t rxr_rma_post_shm_write(struct rxr_ep *rxr_ep, struct rxr_op_entry *tx_ent
 		return -FI_EAGAIN;
 
 	rxr_pkt_init_write_context(tx_entry, pkt_entry);
+	rxr_op_entry_prepare_to_post_write(tx_entry);
+	pkt_entry->pkt_size = tx_entry->bytes_write_total_len;
```
lrbison, line 126: I need to change this pkt_size to `rma_context_pkt->seg_size`

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -509,6 +509,89 @@ int rxr_pkt_entry_read(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry,
...
+			assert( len == tx_entry->bytes_write_total_len );
+			ibv_wr_rdma_write_imm(qp->ibv_qp_ex, remote_key,
+				remote_buf, tx_entry->cq_entry.data);
+		} else {
+			ibv_wr_rdma_write(qp->ibv_qp_ex, remote_key, remote_buf);
+		}
+
+		sge.addr = (uint64_t)local_buf;
+		sge.length = len;
+		sge.lkey = ((struct efa_mr *)desc)->ibv_mr->lkey;
```
darrylabbate, line 569: I think we should do some kind of check on `ibv_mr` for `NULL` here. Not every `efa_mr` will have an initialized `ibv_mr`. Even if such `efa_mr` objects wouldn't currently go through this code path, we could save ourselves some headache if that changes in the future.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1656,16 +1687,16 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			break;
 		}
 
-		pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
```
wzamazon, line 1659: why remove this. For RECV_RDMA_WITH_IMM, this will just be NULL. Right?

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2484,6 +2523,51 @@ void rxr_ep_handle_misc_shm_completion(struct rxr_ep *ep,
...
 
+/**
+ * @brief handle RX completion due to FI_REMOTE_CQ_DATA via RECV_RDMA_WITH_IMM
+ *
+ * This function handles hardware-assisted RDMA writes with immediate data at
+ * remote endpoint.  These do not have a packet context, nor do they have a
+ * connid available.
+ *
+ * @param[in]		ep		endpoint
+ * @param[in]		cq_entry	CQ entry from shm provider
```
wzamazon, line 2534: There is no `cq_entry` and `src_addr` in this function

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2484,6 +2523,51 @@ void rxr_ep_handle_misc_shm_completion(struct rxr_ep *ep,
...
+ * @param[in]		src_addr	source address
+ */
+void rxr_ep_handle_remote_cq_completion(struct rxr_ep *ep,
```
wzamazon, line 2537: shouldn't the function name be `recv_rdma_with_imm_completion`?

prov/efa/src/rdm/rxr_pkt_cmd.c
wzamazon, line 801: RDMA-core did not return correct op code for read.

prov/efa/src/rdm/rxr_pkt_entry.c
lrbison: Yeah, this code path will always require it because of the RDMA lkey, but I've put in a simple assert at the top of this block to perhaps make it more clear:
`assert(((struct efa_mr *)desc)->ibv_mr);`

prov/efa/src/rdm/rxr_ep.c
lrbison: yes, that's a better name

prov/efa/src/rdm/rxr_ep.c
lrbison: Thank you

prov/efa/src/rdm/rxr_ep.c
lrbison: I moved it below to the relevant case statements so that we wouldn't accidentally try to de-reference it in future revision.

prov/efa/src/rdm/rxr_pkt_cmd.c
lrbison: Seems like if that's the only reason, we should let it fall through to the default.  I've left it in for now.

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -103,6 +103,17 @@ bool efa_rdm_peer_support_rdma_read(struct efa_rdm_peer *peer)
 	       (peer->extra_info[0] & RXR_EXTRA_FEATURE_RDMA_READ);
 }
 
+static inline
```
vidsouza, line 106: Doxygen comments missing

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -708,7 +710,7 @@ static int rxr_ep_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
  * @param[in]	peer	remote peer
  * @return 	1 if supported, 0 if not, negative errno on error
  */
-int rxr_ep_determine_rdma_support(struct rxr_ep *ep, fi_addr_t addr,
+int rxr_ep_determine_rdma_read_support(struct rxr_ep *ep, fi_addr_t addr,
 				  struct efa_rdm_peer *peer)
```
vidsouza, line 714: Indentation off

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -728,6 +730,35 @@ int rxr_ep_determine_rdma_support(struct rxr_ep *ep, fi_addr_t addr,
 	return 1;
 }
 
+/*
+ * For a given peer, trigger a handshake packet and determine if both peers
```
vidsouza, line 734: `@brief` missing

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -728,6 +730,35 @@ int rxr_ep_determine_rdma_support(struct rxr_ep *ep, fi_addr_t addr,
...
+ * support rdma write.
+ *
+ * @param[in]	ep	rxr_ep
+ * @param[in]	addr	remote address
+ * @param[in]	peer	remote peer
+ * @return 	1 if supported, 0 if not, negative errno on error
+ */
+int rxr_ep_determine_rdma_write_support(struct rxr_ep *ep, fi_addr_t addr,
+				  struct efa_rdm_peer *peer)
```
vidsouza, line 743: Indentation off

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -303,7 +303,9 @@ int rxr_ep_post_user_recv_buf(struct rxr_ep *ep, struct rxr_op_entry *rx_entry,
 
 struct efa_rdm_peer;
 
-int rxr_ep_determine_rdma_support(struct rxr_ep *ep, fi_addr_t addr,
+int rxr_ep_determine_rdma_read_support(struct rxr_ep *ep, fi_addr_t addr,
+				  struct efa_rdm_peer *peer);
```
vidsouza, line 307: Indentations off in 307 and 309

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -509,6 +509,91 @@ int rxr_pkt_entry_read(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry,
...
+ * @param[in]		local_buf 	local buffer, where data will be copied from.
+ * @param[in]		len		write size.
+ * @param[in]		desc		memory descriptor of local buffer.
+ * @param[in]		remote_buff	remote buffer, where data will be written to.
+ * @param[in]		remote_key	memory key of remote buffer.
+ * @return	On success, return 0
+ * 		On failure, return a negative error code.
+ */
+int rxr_pkt_entry_write(struct rxr_ep *ep, struct rxr_pkt_entry *pkt_entry,
+		       void *local_buf, size_t len, void *desc,
```
vidsouza, line 528: Indentations off

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -322,6 +370,10 @@ ssize_t rxr_rma_post_write(struct rxr_ep *ep, struct rxr_op_entry *tx_entry)
 	if (peer->is_local && ep->use_shm_for_tx)
 		return rxr_rma_post_shm_write(ep, tx_entry);
 
+	if (rxr_rma_should_write_using_rdma(ep, tx_entry, peer)) {
+		return rxr_op_entry_post_remote_write(tx_entry);
```
vidsouza, line 374: Although brackets aren't wrong to keep, to maintain consistency in the code you might want to remove them.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1680,6 +1711,14 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
...
 #endif
 			break;
+		case IBV_WC_RDMA_WRITE:
+			pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
+			rxr_pkt_handle_write_completion(ep, pkt_entry);
+			break;
+		case IBV_WC_RECV_RDMA_WITH_IMM:
+			recv_rdma_with_imm_completion(ep,
+				ibv_wc_read_imm_data(ep->ibv_cq_ex),
+				FI_REMOTE_CQ_DATA | FI_RMA | FI_REMOTE_WRITE );
```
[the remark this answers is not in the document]
lrbison: oh, thank you!

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -92,6 +92,12 @@ struct efa_rdm_peer {
 	int64_t num_read_msg_in_flight;
 };
 
+/**
+ * @brief check for peer's RDMA_READ support, assuming HANDSHAKE has already occurred
+ *
+ * @param peer[in]	A peer which we have already received a HANDSHAKE from
```
wenduwan, line 98: Does your IDE highlight this syntax? Mine only works with
```
@param[in] peer
```

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -116,13 +139,38 @@ bool efa_rdm_peer_support_delivery_complete(struct efa_rdm_peer *peer)
...
+ * @brief determine if both peers support RDMA write
+ *
+ * This function can only return true if a handshake packet has already been
+ * exchanged, and the peer set the RXR_EXTRA_FEATURE_RDMA_WRITE flag.
+ * @params[in]		ep		Endpoint for communication with peer
+ * @params[in]		peer		An EFA peer
+ * @return		boolean		both self and peer support RDMA write
+ */
+static inline
+bool efa_both_support_rdma_write(struct rxr_ep *ep, struct efa_rdm_peer *peer)
```
shijin-aws, line 168: I feel like `**_rdma_read**` and `**_rdma_write**` functions are mostly identical. Is it possible to make `read` and `write` as op code that can be toggled, and use a single set of rdma_op functions to cover the both cases?

prov/efa/src/rdm/efa_rdm_peer.h
lrbison: Fixed.  Thank you.

prov/efa/src/rdm/efa_rdm_peer.h
lrbison: How about something like `efa_both_support_handshake_flag( ep, peer, RXR_EXTRA_FEATURE_RDMA_WRITE)`

And that function can just check `ep->extra_info` for self, (optionally) trigger a handshake, and check peer's extra info.

Downside is now we have a longer function with a very verbose (but precise!) RXR_EXTRA_FEATURE_* but the upside is we don't have to duplicate the logic for all of these.

shijin-aws: That makes sense. Thanks.

shijin-aws: I think this macro trick look better.

lrbison: I think this is a good change, but after looking at it, I'd prefer to do it in a separate PR, as it touches other code paths and I don't want to tangle them up in the RDMA_WRITE support.

shijin-aws: I am ok with leaving it as a separate refactor.

lrbison:
My last push fixed a logic error I had which incorrectly took this RDMA Write path for single-iov writes.

With that fixed, CI passes.

Ready to merge?

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -412,6 +464,7 @@ ssize_t rxr_rma_writemsg(struct fid_ep *ep,
 		err = -FI_EAGAIN;
 		goto out;
 	}
+	rxr_op_entry_prepare_to_post_write(tx_entry);
```
wzamazon, line 467: I think this is used only in true write.

lrbison: It is needed in both rxr_rma_post_shm_write (where it was already being called... ) and in `rxr_op_entry_post_remote_write`.

I removed the call from `rxr_rma_writemsg` and included it in `rxr_op_entry_post_remote_write`.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -1357,6 +1387,117 @@ int rxr_op_entry_post_remote_read(struct rxr_op_entry *op_entry)
...
+	int err;
+	int iov_idx = 0, rma_iov_idx = 0;
+	size_t iov_offset = 0, rma_iov_offset = 0;
+	size_t write_once_len, max_write_once_len;
+	struct rxr_ep *ep;
+	struct rxr_pkt_entry *pkt_entry;
+
+	assert(op_entry->iov_count > 0);
+	assert(op_entry->rma_iov_count > 0);
+	rxr_op_entry_prepare_to_post_write(op_entry);
```
wzamazon, line 1414: `rxr_op_entry_prepare_to_post_write` cannot be called here.
this function can be called repeated when write encounter -FI_EAGAIN.

`rxr_op_entry_prepare_to_post_write` will reset the counter `bytes_write_submitted` and `bytes_write_completed`.

lrbison: As discussed offline, we will make `rxr_rma_post_shm_write` and `rxr_op_entry_post_remote_write` both as being capable of being called again after returning -FI_AGAIN, but both of these functions are currently not called from a routine like `rxr_op_entry_post_remote_read_or_queue`, and so any failures filter all the way back to the user of libfabric.

See my latest push

prov/efa/src/rdm/rxr_pkt_cmd.c
```diff
@@ -883,6 +883,25 @@ void rxr_pkt_handle_send_completion(struct rxr_ep *ep, struct rxr_pkt_entry *pkt
...
+	default:
+		EFA_WARN(FI_LOG_CQ,
+			"invalid pkt type %d in rxr_pkt_handle_write_completion\n",
+			rxr_get_base_hdr(pkt_entry->wiredata)->type);
+		assert(0 && "invalid control pkt type");
+		efa_eq_write_error(&ep->base_ep.util_ep, FI_EIO, FI_EFA_ERR_INVALID_PKT_TYPE);
+		return;
+	}
+
+	rxr_ep_record_tx_op_completed(ep, pkt_entry);
```
wzamazon, line 901: These 2 lines are never be reached, should be removed

wzamazon: In fact, we do not need this function. The caller can simply call `rxr_pkt_handle_rma_completion`.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1663,6 +1701,15 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			ep->recv_comps++;
 #endif
 			break;
+		case IBV_WC_RDMA_WRITE:
+			pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
+			rxr_pkt_handle_write_completion(ep, pkt_entry);
```
wzamazon, line 1706: It seems we can call `rxr_pkt_handle_rma_completion` here.

prov/efa/src/rdm/rxr_pkt_cmd.c
lrbison: Good point.  Thank you, removed.

prov/efa/src/rdm/rxr_ep.c
lrbison: done

wzamazon (approved):

PR #8524, opened 2023-02-16
prov/efa: Check FI_EFA_USE_DEVICE_RDMA per domain

vidsouza:
Prior to this change FI_EFA_USE_DEVICE_RDMA was checked only once when the provider is initialized. Due to this different plugins/libraries setting up the environment for various applications cannot modify the value of this environment variable.

This patch provides this possibility by checking the value of FI_EFA_USE_DEVICE_RDMA once per domain.

PR #8529, opened 2023-02-16
src/hmem_cuda: Check FI_HMEM_CUDA_ENABLE_XFER in cuda_get_xfer_setting()

vidsouza:
Prior to this patch the value of FI_HMEM_CUDA_ENABLE_XFER was checked only once when hmem cuda was initialized in cuda_hmem_init().

This patch always checks the value of the environment variable FI_HMEM_CUDA_ENABLE_XFER whenever cuda_get_xfer_setting() is called.

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -76,12 +76,8 @@ struct cuda_ops {
 };
 
 static bool hmem_cuda_use_gdrcopy;
-static enum cuda_xfer_setting hmem_cuda_xfer_setting;
-static bool cuda_ipc_enabled;
+static bool cuda_hmem_devices_verified = false;
 static int cuda_device_count;
-
-static cudaError_t cuda_disabled_cudaMemcpy(void *dst, const void *src,
-					    size_t size, enum cudaMemcpyKind kind);
 
 #if ENABLE_CUDA_DLOPEN
 
@@ -366,15 +362,6 @@ int cuda_get_base_addr(const void *ptr, void **base, size_t *size)
 	return -FI_EIO;
 }
 
-static cudaError_t cuda_disabled_cudaMemcpy(void *dst, const void *src,
-					    size_t size, enum cudaMemcpyKind kind)
-{
-	FI_WARN(&core_prov, FI_LOG_CORE,
-		"cudaMemcpy was called but FI_HMEM_CUDA_ENABLE_XFER = 0, "
-		"no copy will occur to prevent deadlock.");
-
-	return cudaErrorInvalidValue;
-}
 
 static int cuda_hmem_dl_init(void)
 {
@@ -590,7 +577,7 @@ static int cuda_hmem_verify_devices(void)
 
 int cuda_hmem_init(void)
 {
-	int ret, param_value;
+	int ret;
 	int gdrcopy_ret;
 
 	fi_param_define(NULL, "hmem_cuda_use_gdrcopy", FI_PARAM_BOOL,
@@ -627,24 +614,7 @@ int cuda_hmem_init(void)
 		}
 	}
 
-	param_value = 1;
-	ret = fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &param_value);
-	if (ret) {
-		hmem_cuda_xfer_setting = CUDA_XFER_UNSPECIFIED;
-	} else {
-		hmem_cuda_xfer_setting = (param_value > 0) ? CUDA_XFER_ENABLED : CUDA_XFER_DISABLED;
-	}
-
-	FI_INFO(&core_prov, FI_LOG_CORE, "cuda_xfer_setting: %d\n", hmem_cuda_xfer_setting);
-
-	if (hmem_cuda_xfer_setting == CUDA_XFER_DISABLED)
-		cuda_ops.cudaMemcpy = cuda_disabled_cudaMemcpy;
-
-	/*
-	 * CUDA IPC is only enabled if cudaMemcpy can be used.
-	 */
-	cuda_ipc_enabled = (hmem_cuda_xfer_setting != CUDA_XFER_DISABLED);
-
+	cuda_hmem_devices_verified = true;
 	return FI_SUCCESS;
 
 dl_cleanup:
@@ -754,12 +724,27 @@ int cuda_host_unregister(void *ptr)
 
 enum cuda_xfer_setting cuda_get_xfer_setting(void)
 {
+	enum cuda_xfer_setting hmem_cuda_xfer_setting;
+	int ret, param_value = 1;
+
+	/* Fetch the value of the environment variable FI_HMEM_CUDA_ENABLE_XFER */
+	ret = fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &param_value);
+
+	if (ret) {
+		hmem_cuda_xfer_setting = CUDA_XFER_UNSPECIFIED;
+	} else {
+		hmem_cuda_xfer_setting = (param_value > 0) ? CUDA_XFER_ENABLED : CUDA_XFER_DISABLED;
+	}
+
+	FI_INFO(&core_prov, FI_LOG_CORE, "cuda_xfer_setting: %d\n", hmem_cuda_xfer_setting);
+
 	return hmem_cuda_xfer_setting;
 }
 
 bool cuda_is_ipc_enabled(void)
 {
-	return cuda_ipc_enabled;
+	return (cuda_get_xfer_setting() != CUDA_XFER_DISABLED) && 
+		cuda_hmem_devices_verified;
 }
 
 int cuda_get_ipc_handle_size(size_t *size)


vidsouza:
This PR and [prov/efa: Check FI_EFA_USE_DEVICE_RDMA per domain](https://github.com/ofiwg/libfabric/pull/8524) have come common files. Rebasing either of the PRs is necessary before merging.

src/hmem_cuda.c
```diff
@@ -754,11 +726,33 @@ int cuda_host_unregister(void *ptr)
 
 enum cuda_xfer_setting cuda_get_xfer_setting(void)
 {
+	enum cuda_xfer_setting hmem_cuda_xfer_setting;
+	int ret, param_value = 1;
+
+	ret = fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &param_value);
+	if (ret == -EINVAL){
+		fprintf(stderr, "FI_HMEM_CUDA_ENABLE_XFER was set to an invalid value by the user."
+			" FI_HMEM_CUDA_ENABLE_XFER is boolean and can be set to only 0 or 1. \n");
```
darrylabbate, line 735: We can probably afford to be a little less verbose here. Pedantically, a boolean parameter can be something other than 0/1 - e.g. `yes`/`no`, `on`/`off`, etc.

prov/efa/src/efa_domain.c
```diff
@@ -197,6 +197,8 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		return -FI_EINVAL;
 	}
 
+	// Check the value of environment variable FI_HMEM_CUDA_ENABLE_XFER
```
wzamazon, line 200: Use `/* */` for C

src/hmem_cuda.c
```diff
@@ -366,15 +364,6 @@ int cuda_get_base_addr(const void *ptr, void **base, size_t *size)
 	return -FI_EIO;
 }
 
-static cudaError_t cuda_disabled_cudaMemcpy(void *dst, const void *src,
```
wzamazon, line 369: This change to core code, need to be in a separate PR

wzamazon (changes requested):

prov/efa/src/efa_domain.c
qkoziol: // is valid for C comments since the C99 standard.   Does libfabric require only using C90 standard features?

wzamazon: Since other comments are using `/* */`, it is better to use `/* */` for consistency.

src/hmem_cuda.c
vidsouza: @wzamazon do you want the change in lines 640 and 641 to be in line 746 in this PR and remove that code in a different PR? Or remove line 640 in this PR and remove the function definition in another PR?

wzamazon: The change made to `src/hmem_cuda.c` should be in a separate PR.

src/hmem_cuda.c
```diff
@@ -754,11 +726,35 @@ int cuda_host_unregister(void *ptr)
...
+
+	FI_INFO(&core_prov, FI_LOG_CORE, "cuda_xfer_setting: %d\n", hmem_cuda_xfer_setting);
+
 	return hmem_cuda_xfer_setting;
 }
 
 bool cuda_is_ipc_enabled(void)
 {
+	/* CUDA IPC is only enabled if cudaMemcpy can be used. */
+	cuda_ipc_enabled = ((cuda_get_xfer_setting() != CUDA_XFER_DISABLED) &&
```
wzamazon, line 755: I think this would impact shm provider performance. because shm provider currently call `cuda_is_ipc_enabled()` for each sendmsg.

wzamazon: In fact, there is no way for shm to support CUDA memory if `cuda_get_xfer_setting` is CUDA_XFER_DISABLED. 

So I suggest that we remove the check of `cuda_get_xfer_setting()` in `cuda_is_ipc_enabled()`. Meanwhile, shm provider directly  call `cuda_get_xfer_setting` to determine whether it can support FI_HMEM during fi_getinfo().

@aingerson Would that make sense to you?

src/hmem_cuda.c
```diff
@@ -42,6 +42,7 @@
 
 #include <cuda.h>
 #include <cuda_runtime.h>
+#include <stdio.h>
```
wzamazon, line 45: I do not think this header is needed.

src/hmem_cuda.c
```diff
@@ -754,11 +726,34 @@ int cuda_host_unregister(void *ptr)
...
+
+	FI_INFO(&core_prov, FI_LOG_CORE, "cuda_xfer_setting: %d\n", hmem_cuda_xfer_setting);
+
 	return hmem_cuda_xfer_setting;
 }
 
 bool cuda_is_ipc_enabled(void)
 {
+	/* CUDA IPC is only enabled if cuda_hmem_init() is successful */
+	cuda_ipc_enabled = cuda_hmem_devices_verified == true;
```
wzamazon, line 755: I do not think we need the global variable `cuda_ipc_enabled` any more.

wzamazon (changes requested):
The commit message need some reword. 

There is no domain in this commit.

This  patch makes cuda_get_xfer_setting() to  always check FI_HMEM_CUDA_ENABLE_XFER

src/hmem_cuda.c
```diff
@@ -754,12 +725,33 @@ int cuda_host_unregister(void *ptr)
...
+	FI_INFO(&core_prov, FI_LOG_CORE, "cuda_xfer_setting: %d\n", hmem_cuda_xfer_setting);
+
 	return hmem_cuda_xfer_setting;
 }
 
 bool cuda_is_ipc_enabled(void)
 {
-	return cuda_ipc_enabled;
+	/* CUDA IPC is only enabled if cuda_hmem_init() is successful */
+	return cuda_hmem_devices_verified;
```
shefty, line 754: This change looks unrelated.

src/hmem_cuda.c
shijin-aws: is it enough? I do not think ipc_enabled can be true is `hmem_cuda_enable_xfer` is false?

src/hmem_cuda.c
shijin-aws, line 754: Is it enough? ipc_enable should not be true when xfer is disabled. Are we changing this behavior?

```
/*
	 * CUDA IPC is only enabled if cudaMemcpy can be used.
	 */
	cuda_ipc_enabled = (hmem_cuda_xfer_setting != CUDA_XFER_DISABLED);
```

src/hmem_cuda.c
```diff
@@ -754,12 +725,33 @@ int cuda_host_unregister(void *ptr)
...
+	enum cuda_xfer_setting hmem_cuda_xfer_setting;
+	int ret, param_value = 1;
+
+	/* Fetch the value of the environment variable FI_HMEM_CUDA_ENABLE_XFER */
+	ret = fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &param_value);
+	if (ret == -EINVAL){
+		fprintf(stderr, "FI_HMEM_CUDA_ENABLE_XFER was set to an invalid value by the user."
+			" FI_HMEM_CUDA_ENABLE_XFER is boolean and can be set to only 0/false/no/off or"
+			" 1/true/yes/on.\n");
+		abort();
```
shefty, line 737: Aborting the application is too harsh of a response to an invalid environment variable.  Just use the default.

src/hmem_cuda.c
```diff
@@ -76,13 +77,9 @@ struct cuda_ops {
 };
 
 static bool hmem_cuda_use_gdrcopy;
-static enum cuda_xfer_setting hmem_cuda_xfer_setting;
-static bool cuda_ipc_enabled;
+static bool cuda_hmem_devices_verified = false;
```
shefty, line 80: This looks unrelated.

shefty (changes requested):

src/hmem_cuda.c
vidsouza: I think you are looking at the older version of the commit, is it? It is changed in the recent version.

shijin-aws: But `cuda_hmem_devices_verified` is true as long as cuda_hmem_init succeed, right? It does not depend on xfer settings.

vidsouza: @shijin-aws, in one of the comments (The comment is open above in this PR), @wzamazon suggested not to do that check as that would deteriorate the performance of shm provider.

src/hmem_cuda.c
shijin-aws: We are setting CUDA_XFER_DISABLED in the nccl path right? In that path shm is not used. Why cannot we keep this check?

src/hmem_cuda.c
shijin-aws: I see, I will follow in that thread instead.

src/hmem_cuda.c
shijin-aws: >Meanwhile, shm provider directly call cuda_get_xfer_setting to determine whether it can support FI_HMEM during fi_getinfo().

Is it a propose to change shm provider?

src/hmem_cuda.c
wzamazon: Yes. this function (`cuda_is_ipc_enabled`) is called every time shm provider send a message, with this change, `cuda_get_xfer_setting()` became expensive because it will call `getenv` and parse the result.

So if we call `cuda_get_xfer_setting()` in `cuda_is_ipc_enabled()`. SHM performance will be impacted.

src/hmem_cuda.c
wzamazon: > Is it a proposal to change shm provider?

SHM provider cannot suport CUDA memory if it cannot make cuda calls. so it is better for it to check that during `fi_getinfo`.

shijin-aws: AFAIK, shm cannot use gdrcopy in ipc protocol because it does not support device-to-device transfer. But shm can use gdrcopy for D-H H-D copies in other protocols. I do not think we should make shm not selected in this case.

wzamazon: > But shm can use gdrcopy for D-H H-D copies in other protocols. I do not think we should make shm not selected in this case.

I do not think shm provider use gdrcopy currently. To use gdrcopy, it must call `cuda_gdrcopy_register`, which shm does not.

src/hmem_cuda.c
vidsouza: `cuda_is_ipc_enabled` was previously dependent on `hmem_cuda_xfer_setting` which was set in `cuda_hmem_init`. With these changes `hmem_cuda_xfer_setting` was moved out of `cuda_hmem_init` Therefore the dependency of `cuda_is_ipc_enabled` on `hmem_cuda_xfer_setting` is included in this PR and that dependency is removed. 

Related to https://github.com/ofiwg/libfabric/pull/8529/files#r1114774730

src/hmem_cuda.c
shijin-aws: Based on offline discussion, I am not against removing this check as there should be no impact on shm provider's current behavior.

shijin-aws (approved):

wzamazon (approved):

src/hmem_cuda.c
shefty: If cuda_hmem_init() fails, why would we ever call this function?  It looks like this should be an assert that it's true.

In any case, this is changing the check for *how* ipc_enabled is determined.  *That* change looks like it belongs as a separate patch.  If we wanted to keep the same behavior, the code should look more like this:
```
/* CUDA IPC is only enabled if cudaMemcpy can be used. */
return cuda_get_xfer_setting() !=CUDA_XFER_DISABLE;
```
Either the previous code was wrong or the update is wrong, and it would be nice to have some comment in a commit message as to why the change is being made.

shefty:
Based on PR #8562, can someone explain *why* this change is being made?  Why are we wanting to read an environment variable every time a function is called, versus once?

src/hmem_cuda.c
vidsouza: I have created #8562 as a pre-requisite to this PR. I will then change the logic of `cuda_is_ipc_enabled` to `cuda_get_xfer_setting() !=CUDA_XFER_DISABLE && cuda_hmem_devices_verified`.

Previously in a local change I had implemented the check logic you suggested above and that broke the `ofi_monitors_add_cache` function

wzamazon:
> Based on PR https://github.com/ofiwg/libfabric/pull/8562, can someone explain why this change is being made? Why are we wanting to read an environment variable every time a function is called, versus once?

It is that some middleware want to set the environment variable, so the customer does not need to set it. But because libfabric only read the environment variable during initialization, it will be too late for middleware to set it.

shefty:
Can we capture that in the commit message?

j-xiong:
IMHO, a better approach would be having the providers to defer the reading of environment variables to when the first time fi_getinfo() is called.

wzamazon:
> IMHO, a better approach would be having the providers to defer the reading of environment variables to when the first time fi_getinfo() is called.

Yes, for provider specific environment variable provider can do that. This particular envvar FI_HMEM_CUDA_ENABLE_XFER is accessed by src/hmem_cuda.c, which do not know when `fi_getinfo` is called.

wzamazon (approved):

shefty:
ofi_hmem_init(), which is what calls cuda_hmem_init(), is called from fi_ini().  That's called from fi_getinfo(), plus a couple of other places (e.g. fi_fabric() and fi_getparams()).  fi_fabric is unlikely, but fi_getparams() is called during provider initialization.

j-xiong:
@wzamazon You are right. However, static variables can still be used to avoid reading the environment variable each time the status is checked.  This way PR #8562 is not necessary.

shefty:
Can we agree that libraries trying to define environment variables is goofy?  That will always be racy.  Environmental variables themselves are a problem because middleware could request different values.  We've seen this in practice.

Anyway, I agree with @j-xiong, PR #8562 isn't desirable.  We're still trying to read an environment variable, not a per smr domain setting.

I think what you're really wanting is a cuda_init_xfer_setting() call that can be invoked separately from cuda_hmem_init().  That allows for a window where middleware _could_ set the xfer_setting value after cuda_hmem_init() has been called.  Is that window guaranteed to work?  I'm skeptical, which could indicate that the proposed solution is racy.  If MPI could call fi_getinfo() before the window closes, then I don't see why it couldn't also open a shm domain before it closes.

bwbarrett:
I generally agree with everything Sean said.  I think the primary problem is that application users really shouldn't be setting this environment variable.  Expecting them to know answers for these type of flags for the various libraries they're using is a bad customer experience.  I don't have strong feelings about the mechanism, but I do have strong feelings that NCCL users shouldn't need to know they need to set the flag in a particular way.

qkoziol:
> > Based on PR #8562, can someone explain why this change is being made? Why are we wanting to read an environment variable every time a function is called, versus once?
> 
> It is that some middleware want to set the environment variable, so the customer does not need to set it. But because libfabric only read the environment variable during initialization, it will be too late for middleware to set it.

I would *never* do this in HDF5 (for example) - which middleware are you thinking of?   (I would set values in an MPI info object, and expect MPI to pass them down to the underlying network layer)

bwbarrett:
> > > Based on PR #8562, can someone explain why this change is being made? Why are we wanting to read an environment variable every time a function is called, versus once?
> > 
> > 
> > It is that some middleware want to set the environment variable, so the customer does not need to set it. But because libfabric only read the environment variable during initialization, it will be too late for middleware to set it.
> 
> I would _never_ do this in HDF5 (for example) - which middleware are you thinking of? (I would set values in an MPI info object, and expect MPI to pass them down to the underlying network layer)

No middleware *wants* to set an environment variable.  But MPI or NCCL or similar are the right place to know the correct value for this stetting, and currently the only way to configure this value is via an environment variable.  Give NCCL or MPI a programmatic way to configure their resources, and that is better for everyone.  But if an environment variable is the only mechanism we have, that's what we're going to use, because asking our users to get it right is not reasonable.

shefty:
We agree. Environment variables are terrible, but frequently the best option available.  This PR isn't adding a new variable, it's just changing one that's already there.  The library uses the best default option that it can.  My concern about this particular change is whether it actually solves anything.

qkoziol:
> We agree. Environment variables are terrible, but frequently the best option available. This PR isn't adding a new variable, it's just changing one that's already there. The library uses the best default option that it can. My concern about this particular change is whether it actually solves anything.

Got it - wasn't certain if this was new or not.   Anything we can do to avoid environment variable in the future is a good thing, too.

shefty:
This PR did not introduce something like cuda_init_xfer_setting().  It changed the behavior of an existing call, which negatively impacted performance and resulted in additional changes being needed.

Maybe we can start with why hmem_cuda_xfer_setting has a nonsensical value like CUDA_XFER_UNSPECIFIED to begin with.  How does that differ from ENABLED or DISABLED?  Why isn't ENABLED or DISABLED the default based on either the build or some runtime check?

aingerson:
I'm also not clear on what the use case here is or why this environment variable is here in the first place. It looks like it defines whether IPC is enabled or not but we already have an environment variable for this across all ifaces (FI_HMEM_DISABLE_P2P). 

The cuda_hmem_devices_verified value also seems silly since it's pretty much just storing whether the interface initialized successfully.

Why would an application/middleware want to toggle the xfer setting? If a system has CUDA devices and CUDA memory that it is passing into OFI and cudaMemcpy is available, then wouldn't they want to use it? Why would they turn it off? Is the target of what you're trying to change is whether IPC is used? Or whether to use gdrcopy? 

More information on the impact of the xfer toggle and need for it would be helpful.

wzamazon:
> Why would an application/middleware want to toggle the xfer setting? If a system has CUDA devices and CUDA memory that it is passing into OFI and cudaMemcpy is available, then wouldn't they want to use it? Why would they turn it off?  Is the target of what you're trying to change is whether IPC is used? Or whether to use gdrcopy?

Basically, NCCL does not allow its plugin (hence libfabric) to make CUDA calls (even when that is available), which is because NCCL launch CUDA kernel on receiver buffers,

NCCL expect libfabric to either use gdrcopy or use device's RDMA capablity to copy data to receiver buffer.

Meanwhile, NCCL handles intra-node communication by itself, therefore NCCL will not use libfabric's SHM provider.

MPI on the other hand allow libfabric to make CUDA calls.

aingerson:
Thanks for the background.
So you're saying MPI has to turn cudaMemcpy off in order to run with NCCL? How does it know it's being used with NCCL?

shefty:
It sounds like NCCL won't ever invoke this call path anyway, so is there an actual problem?

Is an app using both NCCL and MPI?  Does NCCL being present in the process shut everybody down, or can MPI use CUDA?

wzamazon:
> So you're saying MPI has to turn cudaMemcpy off in order to run with NCCL? How does it know it's being used with NCCL?

The idea is that NCCL and MPI call `setenv` to this envvar as they need, then create domain/endpoint, and the created domain/endpoint will remember the setting, so their behavior will not change if the envvar was set by other middleware later.
tion, the communication was done by NCCL. 

> Is an app using both NCCL and MPI?

Yes, some application use MPI to launch NCCL jobs. Although MPI was not used for actually data communication, it will call `fi_getinfo` which will call `cuda_hmem_init`.

shefty:
I think we need a setopt call for this, instead of coordinated environment variables.

aingerson:
Looks like we already have a similar option. Can it be used/modified/extended to solve this?
```FI_OPT_FI_HMEM_P2P - int
Defines how the provider should handle peer to peer FI_HMEM transfers for this endpoint. By default, the provider will chose whether to use peer to peer support based on the type of transfer (FI_HMEM_P2P_ENABLED). Valid values defined in fi_endpoint.h are:
FI_HMEM_P2P_ENABLED: Peer to peer support may be used by the provider to handle FI_HMEM transfers, and which transfers are initiated using peer to peer is subject to the provider implementation.
FI_HMEM_P2P_REQUIRED: Peer to peer support must be used for transfers, transfers that cannot be performed using p2p will be reported as failing.
FI_HMEM_P2P_PREFERRED: Peer to peer support should be used by the provider for all transfers if available, but the provider may choose to copy the data to initiate the transfer if peer to peer support is unavailable.
FI_HMEM_P2P_DISABLED: Peer to peer support should not be used.
fi_setopt() will return -FI_EOPNOTSUPP if the mode requested cannot be supported by the provider.
The FI_HMEM_DISABLE_P2P environment variable discussed in [fi_mr(3)](https://ofiwg.github.io/libfabric/main/man/fi_mr.3.html) takes precedence over this setopt option.```

**Comment by wzamazon** [comment]
> I think we need a setopt call for this, instead of coordinated environment variables.

Unfortunately, setopt is not suitable for this information (whether cuda call is allowed), because the information is used by `fi_getinfo` to determine whether FI_HMEM can be supported.

If CUDA call is not allowed, some provider cannot support CUDA memory and should return -FI_ENODATA. (It can be argued that shm provider should do that.)

fi_setopt is called on an endpoint, which is created after successful `fi_getinfo` call.

**Comment by shefty** [comment]
Can #8572 be used as a solution here?  I really don't want to get into trying to handle dualing middleware changing global variables.

**Comment by wzamazon** [comment]
> Can https://github.com/ofiwg/libfabric/pull/8572 be used as a solution here? 

I am not sure how to use that PR for this environment variable. Can you give some hint?

>  I really don't want to get into trying to handle dualing middleware changing global variables.

Me either.

Ideally this information should be passed to provider through call to `fi_getinfo()`. 

Can we add a flag that can be passed as `mr_mode`?

**Comment by shefty** [comment]
That PR allows putting environment variables into a config file, so it can be set up front.

I'm still not understanding what the desired behavior for this is supposed to be when mixing NCCL + MPI.  But it still feels like the solution should be handled as some sort of fi_setopt().  This can be done prior to calling fi_getinfo() by using fi_open() and determining what the correct object is to open.

For example, "params" to access the environment variable parameters, or "core" to open the libfabric core, or "hmem" for generic hmem settings, or "cuda" for cuda specific settings.  I'm not sure which makes the most sense.  "hmem" might work, and could allow re-use of the FI_OPT_FI_HMEM_P2P as @aingerson suggests.

The latter option would result in the value being set for all users, since libfabric wouldn't have context to separate users.

**Comment by wzamazon** [comment]
> But it still feels like the solution should be handled as some sort of fi_setopt(). This can be done prior to calling fi_getinfo() by using fi_open() and determining what the correct object is to open.

I do not understand this. `fi_setopt` can only be applied on an endpoint. Is that not the case?

**Comment by shefty** [comment]
I was being more conceptual.  It could be fi_control(), though it depends on what sort of fid fi_open() returns and the ops that we define for it.

**Comment by wzamazon** [comment]
> I was being more conceptual. It could be fi_control(), though it depends on what sort of fid fi_open() returns and the ops that we define for it.

I see, but I do not think it would help. If it is called before `fi_getinfo` on a global object like `params` or `hmem`, then it is essentiallly setting a global variable (and is not different from `setenv`).

You asked before  that 
"I'm still not understanding what the desired behavior for this is supposed to be when mixing NCCL + MPI."

From my understanding,

If libfabric keep using `setenv("FI_HMEM_CUDA_ENABLE_XFER"), NCCL/MPI want to:

MPI/NCCL call `setenv` before `fi_getinfo()`,
`fi_getinfo()` will check the env to determine whether it can support the hints,
MPI/NCCL then call `fi_domain/fi_endpoint` to create domain and endpoint, the created domain/endpoint will remember the option.

If another middleware call setenv("FI_HMEM_CUDA_ENABLE_XFER") to a different value, the domain/endpoint created previously did not change behavior.

**Comment by shefty** [comment]
The problem is that environment variables *are* global.  They're intended to reflect the environment, not a time constrained setting.  Trying to use them to control the behavior of some future function call is racy.

It seems silly for a provider to return ENODATA from one fi_getinfo() call, but actual data to another called milliseconds later with the same input values, which is what you're asking for.

I'm on @qkoziol side here.  This entire approach seems fundamentally wrong.

If the apps are only using MPI to start some job, why does it need HMEM support anyway?  Just disable it.  And does any of this matter if fi_getinfo() is just going to fail when called by the second middleware (e.g. NCCL) anyway? What good does it do to return success from fi_getinfo() to MPI, but then fail the call to NCCL?  None of this is making sense to me.

Add to all of this, that hmem_cuda_enable_xfer variable seems misnamed.  It enables/disables cudamemcpy and should be renamed to reflect that.

**Comment by wzamazon** [comment]
@shefty

I see your point.

Let's say we desert the environmental variable approach entirely.

What about we introduce two new MR mode flag, and middleware set them in `hint->domain_attr->mr_mode`, and pass  it to `fi_getinfo`?

**Comment by wzamazon** [comment]
@shefty

I think we agreed not to use environment variable for this purpose.

Do you think my suggestion about passing this info (whether libfabric can use CUDA API on the memory regions passed by application) via a `mr_mode` flag to `fi_getinfo` make sense?

**Comment by shefty** [comment]
No, I don't think it makes sense to pass this as an mr_mode bit or general mode bit.  Those are intended as application requirements upward, not app restrictions to pass down.  I still don't even understand the problem that this is trying to solve.

Why do we want one call to fi_getinfo() (from MPI) to succeed, but a second call to fi_getinfo() (from NCCL) to fail?  What is the app supposed to do with that?  Why not fail both?

**Comment by wzamazon** [comment]
>  I still don't even understand the problem that this is trying to solve.

In my mind, the problem:

When a provider implement support of CUDA memory, the support can be implemented by using CUDA API calls, or it can be implemented by using GPUDirect RDMA.

Some middleware (Like NCCL) cares about how the CUDA support is implemented. If a provider support CUDA memory by making CUDA call, NCCL want them NOT to claim support CUDA memory. For this purpose, NCCL will call `fi_getinfo` twice, first time with FI_HMEM cap (with the understanding that provider will not use CUDA API to support FI_HMEM). If the 1st call failed, it will call `fi_getinfo` 2nd time with out FI_HMEM and will not pass CUDA memory to libfabric. 

Some middleware (Like MPI) does not care about how CUDA support is implemented.

Currently in the libfabric API, there is no way for an user to specify this requirement (The envar FI_HMEM_CUDA_ENABLE_XFER was introduced for this,  and I think we agreed this is not a good)

I think from NCCL point of view,  it need a way to achieve:

1. when calling `fi_getinfo`,  if a provider has to make CUDA call to support CUDA memory, then the provider should not cliam support of FI_HMEM.
2. the returned `fi_info` object has this information. The endpoint created using this info object will not make CUDA calls during transfer.

**Comment by shefty** [comment]
The environment variable can work for this.  But it's not something that we can expect to toggle.  It's the expectation that toggling the value would work that's the problem IMO.

The alternative to trying to pass this information down is for the provider to pass up details on its operation, and let the app decide how to act based on that.  We don't necessarily need to fail the first fi_getinfo call.

**Comment by wzamazon** [comment]
> The environment variable can work for this. But it's not something that we can expect to toggle. It's the expectation that toggling the value would work that's the problem IMO.

Yes.

> The alternative to trying to pass this information down is for the provider to pass up details on its operation, and let the app decide how to act based on that. We don't necessarily need to fail the first fi_getinfo call.

Yes. that would work too. For example:

App call `fi_getinfo` and request `FI_HMEM` capability. If a provider must make CUDA calls to support CUDA memory, it set a flag in `mr_mode`, and return success.

App check `mr_mode` upon success return to `fi_getinfo`,  and decide what to do.

For NCCL, it can simply not to pass CUDA pointer (no need to call `fi_getinfo`for the 2nd time).

Does it sound about right?

**Comment by shefty** [comment]
This isn't how mr_mode works.  That's related to specific steps an application must take relative to memory registration, not internal implementation details for dealing with HMEM.

We need a more generic solution if a simple environment variable isn't usable.  Maybe the addition of fi_hmem_attr to describe hmem support and restrictions in more detail.

**Comment by wzamazon** [comment]
> We need a more generic solution if a simple environment variable isn't usable. Maybe the addition of fi_hmem_attr to describe hmem support and restrictions in more detail.

I totally agree. I see you opened https://github.com/ofiwg/libfabric/issues/8599, I left comments there.

For this particular issue though, I just realized we can use `fi_getopt/fi_setopt` after all, after you point out that the 1st `fi_getinfo` call can succeed.

My idea is:

1. We introduce an opt flag like FI_HMEM_OPT_CUDA_API_CALL, value can be ENABLED/DISABLED. Libfabric API version need to be bump for this new flag.

2. provider always claim support of FI_HMEM.

3. NCCL plugin call `fi_getinfo` with correct libfabric API version. provider claim support of FI_HMEM

4. NCCL plugin create endpoint from it. then call `fi_setopt(FI_HMEM_OPT_CUDA_MAKE_API_CALL, DISABLED)` to disable CUDA api call.

5. If provider cannot support CUDA without making CUDA API call, it return -FI_EOPNOTSUPP.

The key point is that Provider can determine its behavior based on `version` passed by `getinfo`. If app user older version, provider maintain current behavior.

**Comment by shefty** [comment]
If I understand the NCCL requirement correctly, having it call:
```
fi_setop(ep, FI_OPT_FI_HMEM_P2P, FI_HMEM_P2P_REQUIRED, ...)
```
would indicate that only PCI peer-to-peer transfers are allowed.  Since, EFA cannot support this option, it would fail the call (returning -FI_EOPNOTSUPP).  NCCL can check the return value and adjust its behavior accordingly, so that it never passes device memory through the API.

AFAICT, this aligns with what's already defined by the API and the needs of NCCL.  It also seems to be part of the reason why this option was added in the first place:

https://github.com/ofiwg/libfabric/commit/d906745c30d94ddb999dba9b5204acadfc7862fd

wzamazon:
I think you are right. Let me close this PR.

PR #8531, opened 2023-02-17
man/fi_peer: Fix the input argument in fi_ops_srx_peer

shijin-aws:
The input argument should be struct fi_peer_rx_entry *entry.


lrbison (approved):

shefty:
thanks - skipping CI since this updates man page only

PR #8537, opened 2023-02-17
prov/efa: Check FI_HMEM_CUDA_ENABLE_XFER per domain

vidsouza:
This patch provides EFA provider the possibility to check the value of FI_HMEM_CUDA_ENABLE_XFER per
domain.


wzamazon:
There is another occurance of `cuda_get_xfer_setting()`,


https://github.com/ofiwg/libfabric/blob/2d9ba8f1de87ba467ae5b440f3303b5a66891bf1/prov/efa/src/rdm/rxr_pkt_type_base.c#L391

which we need to change to `domain->cuda_xfer_setting`.

You can get domain of ep by calling `rxr_ep_domain()`

wzamazon (changes requested):
Another occurance of `cuda_get_xfer_setting()` need to be converted to `domain->cuda_xfer_setting`

vidsouza:
Rebased the commit.

PR #8539, opened 2023-02-17
[v1.17.x] Exclude OPX from build if older hfi1_user.h version

charlesshereda:
OPX will not compile if the system's hfi1_user.h is older than version 6; for now, exclude OPX from build if that header's macro HFI1_USER_SWMAJOR != 6.


charlesshereda:
Closing this in lieu of Sean's commit #8541 which should also be pulled into v1.17.x.

PR #8540, opened 2023-02-17
Exclude OPX from build if older hfi1_user.h version

charlesshereda:
OPX will not compile if the system's hfi1_user.h is older than version 6; for now, exclude OPX from build if that header's macro HFI1_USER_SWMAJOR != 6.

### DIFF
--- a/prov/opx/configure.m4
+++ b/prov/opx/configure.m4
@@ -120,13 +120,26 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 		    [],
 		    [],
 		    [opx_happy=0])
-
-               AC_CHECK_DECL([HAVE_ATOMICS],
-                             [],
-                             [cc_version=`$CC --version | head -n1`
-                              AC_MSG_WARN(["$cc_version" doesn't support native atomics.  Disabling OPX provider.])
-                              opx_happy=0])
-	])
+		AS_IF([test $opx_happy -eq 1],[
+			AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
+				[[#include <rdma/hfi/hfi1_user.h>]],
+				[[
+					#if HFI1_USER_SWMAJOR != 6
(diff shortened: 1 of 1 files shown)


prov/opx/configure.m4
```diff
@@ -120,12 +120,25 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
...
+				[
+				AC_MSG_NOTICE([hfi1_user.h version 6.x found... no])
+				opx_happy=0
+				])
+		])
+		AC_CHECK_DECL([HAVE_ATOMICS],
+                     [],
+                     [cc_version=`$CC --version | head -n1`
+                      AC_MSG_WARN(["$cc_version" doesn't support native atomics.  Disabling OPX provider.])
+                      opx_happy=0])
```
j-xiong, line 141: Are these lines changed accidentally?  They don't seem to be necessary or related.

charlesshereda: I just changed the indentation.

j-xiong: After the change the indentation is still not consistent anyway :) (check L110 vs L119 vs L124 vs L138 for example)

charlesshereda: The old version used spaces where it shouldn't have and I left some there that I shouldn't have. I just force pushed an updated version with the proper alignment and tabs. If you really want, I can revert to the original version that doesn't line up properly and uses spaces...

shefty:
See #8541 for alternative patch.

j-xiong:
Just checked, `HFI1_CAP_TID_RDMA` was introduced in Kernel 5.0. It's missing from all Kernel 4.x headers. The version number `HFI1_USER_SWMAJOR` and `HF1_USER_SWMINOR` remain unchanged (6 and 3, respectively). That's why this PR doesn't work while #8541 does.

charlesshereda:
Sounds good, closing this one, I'll test #8451 now.

PR #8541, opened 2023-02-17
prov/opx: Exclude opx from build if missing needed defines

shefty:
Provider needs HFI1_CAP_TID_RDMA defined for debug output. Disable from build if not present in hfi1_user.h file.


charlesshereda (approved):
Looks good, thank you.

PR #8542, opened 2023-02-18
fabtests/msg_inject: Rename test to inject_test

### LINKED ISSUE #8515: fabtests: fi_msg_inject is a rdm test by default
**Describe the bug**
In general, tests starting with `fi_msg_` test *FI_EP_MSG*, and tests starting with `fi_rdm_` test *FI_EP_RDM*.  The `fi_msg_inject` is different from others in that it tests *FI_EP_RDM* by default and can test other EP types with the `-e` option. This causes confusion and inconvenience when trying to interpret the test results and/or to exclude test based on test name patterns. 

This test used to be called `fi_inj_complete` which didn't have the problem. Alternatively, simply calling it `fi_inject` would also be less confusing.

shefty:
'msg' tests usually refer to tests based on msg endpoints, however this test uses rdm endpoints by default.  Rename to better reflect the test.

Fixes #8515


j-xiong (approved):

shefty:
CI failures for udp, ofi_rxd, and shm -- adding inject_test to exclude file.  Prior test was excluded based on name starting with fi_msg

PR #8543, opened 2023-02-19
prov/efa: bugfix uninitialized use_device_rdma value

wenduwan:
This is a bugfix for uninitialized use_device_rdma value. fi_param_get_* does not change the variable value if the environment variable is not set. Hence the unitialized int caused undefined behavior.

### DIFF
--- a/prov/efa/src/rdm/rxr_env.c
+++ b/prov/efa/src/rdm/rxr_env.c
@@ -270,6 +270,14 @@ void rxr_env_initialize()
 int rxr_env_get_use_device_rdma()
 {
 	int ret, use_device_rdma;
+	uint32_t vendor_part_id = g_device_list[0].ibv_attr.vendor_part_id;
+
+	if (vendor_part_id == 0xefa0 || vendor_part_id == 0xefa1) {
+		/* Device RDMA is disabled on 1st/2nd gen EFA by default */
+		use_device_rdma = 0;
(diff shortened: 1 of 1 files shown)


bwbarrett (changes requested):
Did you verify that this does not change the default behavior on C7gn?

wenduwan:
>  Did you verify that this does not change the default behavior on C7gn?

Good catch thanks! Will update the logic in a moment.

prov/efa/src/rdm/rxr_env.c
```diff
@@ -269,7 +269,8 @@ void rxr_env_initialize()
  */
 int rxr_env_get_use_device_rdma()
 {
-	int ret, use_device_rdma;
+	/* Disable device RDMA by default */
+	int ret, use_device_rdma = 0;
```
vidsouza, line 273: Based on my discussion with @wzamazon, `use_device_rdma` should not be set to default value of 0 always. The value of this variable depends on conditions defined in `rxr_env_define_use_device_rdma()`

wzamazon:
> Good catch thanks! Will update the logic in a moment.

The default value of should be determined from the same logic in `rxr_env_define_use_device_rdm()`.

wenduwan:
Synced with  @vidsouza. He will refactor the code and fix the issue.

PR #8545, opened 2023-02-19
prov/tcp,io_uring: Move connect() and CM messages to io_uring

sydidelot:
This PR revisits the existing code for io_uring and starts porting a few CM-related operations associated to the EP.

### DIFF
--- a/prov/tcp/src/xnet_cm.c
+++ b/prov/tcp/src/xnet_cm.c
@@ -131,26 +131,11 @@ int xnet_send_cm_msg(struct xnet_ep *ep)
 	return FI_SUCCESS;
 }
 
-void xnet_req_done(struct xnet_ep *ep)
+static int xnet_req_done_internal(struct xnet_ep *ep)
 {
 	struct xnet_cm_entry cm_entry;
 	uint16_t len;
-	ssize_t ret;
-
-	FI_DBG(&xnet_prov, FI_LOG_EP_CTRL, "connect request done\n");
-	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
-
-	ret = xnet_recv_cm_msg(ep->bsock.sock, ep->cm_msg, ofi_ctrl_connresp);
-	if (ret) {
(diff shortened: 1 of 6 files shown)


prov/tcp/src/xnet_cm.c
```diff
@@ -51,18 +51,21 @@ struct xnet_cm_entry {
...
-		goto err;
+	if (!skip_hdr) {
+		len = sizeof(msg->hdr);
+		ret = ofi_recv_socket(sock, &msg->hdr, len, 0);
+		if ((size_t) ret != len) {
+			FI_WARN(&xnet_prov, FI_LOG_EP_CTRL,
+				"Failed to read cm header\n");
+			ret = ofi_sockerr() ? -ofi_sockerr() : -FI_EIO;
+			goto err;
+		}
```
shefty, line 68: I think it would it be better to separate the call into 2 functions, with the first receiving the header.  The io_uring path does this.  This function would then become more of a 'handle_cm_msg' than recv call then.

shefty (commented):
Change look good -- see comment on recv_cm_msg change.

prov/tcp/src/xnet_cm.c
sydidelot: Thanks. The new revision implements your suggestion.

prov/tcp/src/xnet_ep.c
```diff
@@ -219,7 +219,9 @@ static int xnet_ep_connect(struct fid_ep *ep_fid, const void *addr,
 	}
 
 	ep->state = XNET_CONNECTING;
-	ret = connect(ep->bsock.sock, (struct sockaddr *) addr,
+	ep->addr = malloc(ofi_sizeofaddr(addr));
+	memcpy(ep->addr, addr, ofi_sizeofaddr(addr));
```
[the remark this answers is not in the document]
sydidelot: Done.

shefty:
Thanks!

PR #8546, opened 2023-02-20
prov/shm: Fix a bug in smr_av_insert

shijin-aws:
When FI_AV_USER_ID flag is set, shm should record the fi_addr passed in as the user id in the map before re-assigning it as the real shm fi_addr.

### DIFF
--- a/prov/shm/src/smr_av.c
+++ b/prov/shm/src/smr_av.c
@@ -91,26 +91,29 @@ static int smr_av_insert(struct fid_av *av_fid, const void *addr, size_t count,
 		}
 
 		FI_INFO(&smr_prov, FI_LOG_AV, "fi_addr: %" PRIu64 "\n", util_addr);
-		if (fi_addr)
-			fi_addr[i] = util_addr;
 
 		if (ret) {
+			if (fi_addr)
+				fi_addr[i] = util_addr;
 			if (util_av->eq)
 				ofi_av_write_event(util_av, i, -ret, context);
 			if (shm_id >= 0)
 				smr_map_del(smr_av->smr_map, shm_id);
 			continue;
-		} else {
-			assert(shm_id >= 0 && shm_id < SMR_MAX_PEERS);
-			if (flags & FI_AV_USER_ID) {
-				assert(fi_addr);
-				smr_av->smr_map->peers[shm_id].fiaddr = fi_addr[i];
-			} else {
-				smr_av->smr_map->peers[shm_id].fiaddr = util_addr;
-			}
-			succ_count++;
-			smr_av->used++;
 		}
+
+		assert(shm_id >= 0 && shm_id < SMR_MAX_PEERS);
+		if (flags & FI_AV_USER_ID) {
+			assert(fi_addr);
+			smr_av->smr_map->peers[shm_id].fiaddr = fi_addr[i];
+		} else {
+			smr_av->smr_map->peers[shm_id].fiaddr = util_addr;
+		}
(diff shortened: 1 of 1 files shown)


shijin-aws:
bot:aws:retest

aingerson:
The shm insert call should still be returning the shm address to use, even when USER_ID is used. How else will the owner tell shm what address to use?
The USER_ID flag is just used to specify what address to report back in a completion, not what address to use for the send call. That should still be assigned by shm

shijin-aws:
>The shm insert call should still be returning the shm address to use, even when USER_ID is used. How else will the owner tell shm what address to use?
The USER_ID flag is just used to specify what address to report back in a completion, not what address to use for the send call. That should still be assigned by shm

I see. But reading the code https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_av.c#L94-L95, it seems the `fi_addr` is always overridden to util_addr before being recorded by shm? Which makes it not used? Reading the man page of `fi_av`:

>User-specified fi_addr_t values are provided as part of address insertion (e.g. fi_av_insert) through the fi_addr parameter. The fi_addr parameter acts as input/output in this case. When the FI_AV_USER_ID flag is passed to any of the insert calls, the caller must specify an fi_addr_t identifier value to associate with each address. The provider will record that identifier and use it where required as part of any completion event. Note that the output from the AV insertion call is unchanged. The provider will return an fi_addr_t value that maps to each address, and that value must be used for all data transfer operations.

aingerson:
Oh ok then the bug fix I think you're talking about it moving the fi_addr set to after parsing the user_id flag or really moving the user_id check to before setting the fi_addr.
But you're totally right, this looks incorrect. Thanks for catching!

shijin-aws:
>Oh ok then the bug fix I think you're talking about it moving the fi_addr set to after parsing the user_id flag or really moving the user_id check to before setting the fi_addr.
But you're totally right, this looks incorrect. Thanks for catching!

Yes, my current patch is wrong because I missed the point that the returned `fi_addr` should always be the shm assigned one. I am proposing to move the [`fi_addr[i] = util_addr`](https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_av.c#L94-L95) after the https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_av.c#L105-L110 block.

Does that sound reasonable to you?

prov/shm/src/smr_av.c
```diff
@@ -112,6 +110,9 @@ static int smr_av_insert(struct fid_av *av_fid, const void *addr, size_t count,
 			smr_av->used++;
 		}
 
+		if (fi_addr)
+			fi_addr[i] = util_addr;
```
shefty, line 114: There's a continue case above that skips this assignment.  nit: There's an else statement that follows the continue which can be removed.

shijin-aws: > There's a continue case above that skips this assignment.

Yes, I observed that too. That means we were assigning this addr always, no matter the assertion succeed or not. Was that legitimate?

aingerson: Yes, the util insert call will return FI_ADDR_NOTAVAIL if it was unassigned which is the behavior we want I think

shijin-aws: I see, I will get that assignment back in this case. Thanks!

aingerson (approved):

shefty (approved):
thanks

PR #8548, opened 2023-02-21
prov/efa: grow read_entry_pool when progress engine is called 1st time

wzamazon:
With recent change that op_entry directly post remote read, read_entry is only allocated for local read. Therefore, the 1st read_entry can be allocated at different time of an application.

The 1st read_entry allocation is expensive because it will cause 1 chunk of memory to be allocated. When it happen at different time, the application overall performance can be impacted.

This patch address the issue by allowing the 1st chunk of the memory of read_entry_pool when progress engine is called the 1st time (inside rxr_grow_rx_pkt_pools).

### DIFF
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -1260,18 +1260,22 @@ struct fi_ops_cm rxr_ep_cm = {
 };
 
 /*
- * @brief explicitly allocate a chunk of memory for 5 packet pools on RX side:
+ * @brief explicitly allocate a chunk of memory for 6 pools on RX side:
  *     efa's receive packet pool (efa_rx_pkt_pool)
  *     shm's receive packet pool (shm_rx_pkt_pool)
  *     unexpected packet pool (rx_unexp_pkt_pool),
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1260,18 +1260,22 @@ struct fi_ops_cm rxr_ep_cm = {
...
  *     unexpected packet pool (rx_unexp_pkt_pool),
  *     out-of-order packet pool (rx_ooo_pkt_pool), and
  *     local read-copy packet pool (rx_readcopy_pkt_pool).
+ *     read entry pool (read_entry_pool)
+ *
+ * This function is called when the progress engine is called for
+ * the 1st time on this endpoint.
  *
  * @param ep[in]	endpoint
  * @return		On success, return 0
```
lrbison, line 1275: Not a change you made but:
 - I think this should be `@param[in,out] ep`  rather than `@param ep[in]`

wzamazon: Fixed

lrbison (approved):
thank you

PR #8550, opened 2023-02-21
prov/rxm, rxd, shm: fix coverity warnings

aingerson:


shefty (approved):
nit: 1st patch uses the word 'patch' instead of 'path' in commit message

PR #8553, opened 2023-02-22
prov/efa and fabtests/efa: Use shm provider's FI_AV_USER_ID support and add test 

shijin-aws:
This PR makes efa provider use shm provider's FI_AV_USER_ID support so the efa/shm address conversion is not needed in the cq read.

Also add efa specific fabtests item `test_efa_shm_addr` that validate the fi_addr_t returned by efa provider's `fi_cq_readfrom` is correct in the scenario where a non-shm peer is inserted first, and a shm peer is inserted later so
the efa fi_addr and shm fi_addr could differ.

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -300,20 +300,30 @@ class ClientServerTest:
 
         return command
 
-    def _run_client_command(self, server_process):
+    def _run_client_command(self, server_process, client_command, client_output_file=None,
+                            run_client_asynchronously=False):
 
         if server_process.poll():
+            if client_output_file:
+                with open(client_output_file, "w") as f:
+                    f.write("")
             raise RuntimeError("Server has terminated")
 
         print("")
-        print("client_command: " + self._client_command)
+        print("client_command: " + client_command)
 
-        process = Popen(self._client_command, stdout=subprocess.PIPE,
+        process = Popen(client_command, stdout=subprocess.PIPE,
                         stderr=subprocess.STDOUT, shell=True, universal_newlines=True)
+        if run_client_asynchronously:
+            return process
+
         client_timed_out = False
         output = ""
         try:
             output, _ = process.communicate(timeout=self._timeout)
+            if client_output_file:
+                with open(client_output_file, "w") as f:
+                    f.write(output)
         except TimeoutExpired:
             client_timed_out = True
             process.terminate()
@@ -329,7 +339,7 @@ class ClientServerTest:
         if client_timed_out:
             raise RuntimeError("Client timed out")
 
-        return process.returncode
+        return process
 
     @retry(retry_on_exception=is_ssh_connection_error, stop_max_attempt_number=3, wait_fixed=SERVER_RESTART_DELAY_MS)
     def run(self):
@@ -354,7 +364,7 @@ class ClientServerTest:
                 retry_on_exception=is_ssh_connection_error,
                 stop_max_delay=self._timeout * 1000,  # Convert to milliseconds
                 wait_fixed=CLIENT_RETRY_INTERVAL_MS,
-            )(self._run_client_command)(server_process)
+            )(self._run_client_command)(server_process, self._client_command).returncode
         except Exception as e:
             print("Client error: {}".format(e))
             # Clean up server if client is terminated unexpectedly
@@ -384,34 +394,57 @@ class ClientServerTest:
                               self._cmdline_args.strict_fabtests_mode)
 
 
-class MultinodeTest:
+class MultinodeTest(ClientServerTest):
 
-    def __init__(self, cmdline_args, base_command, numproc):
+    def __init__(self, cmdline_args, server_base_command,
+                 client_base_command, client_hostname_list,
+                 run_client_asynchronously=True):
         self._cmdline_args = cmdline_args
-        self._base_command = base_command
-        self._numproc = numproc
+        self.numclient = len(client_hostname_list)
         self._timeout = self._cmdline_args.timeout
 
-        multinode_command = self._base_command + " -n {}".format(self._numproc)
-        self._server_command = cmdline_args.populate_command(multinode_command, "server", self._timeout)
-        self._client_command = cmdline_args.populate_command(multinode_command, "client", self._timeout)
+        self._server_base_command = cmdline_args.populate_command(server_base_command, "server", self._timeout)
+        self._client_base_command_list = []
+        for client_hostname in client_hostname_list:
+            cmdline_args_copy = copy.copy(cmdline_args)
+            cmdline_args_copy.client_id = client_hostname
+            self._client_base_command_list.append(cmdline_args_copy.populate_command(client_base_command, "client", self._timeout))
+        self._run_client_asynchronously = run_client_asynchronously
 
+    @retry(retry_on_exception=is_ssh_connection_error, stop_max_attempt_number=3, wait_fixed=SERVER_RESTART_DELAY_MS)
     def run(self):
-        if self._cmdline_args.is_test_excluded(self._base_command):
+        if self._cmdline_args.is_test_excluded(self._server_base_command):
+            pytest.skip("excluded")
+
+        # _client_base_command_list is populated from the same client_base_command,
+        # so it should be safe to only check the first one.
+        if self._cmdline_args.is_test_excluded(self._client_base_command_list[0]):
             pytest.skip("excluded")
 
         server_outfile = NamedTemporaryFile(prefix="fabtests_server.out.").name
 
         # start running
-        server_process = Popen(self._server_command + "> " + server_outfile + " 2>&1", shell=True)
+        server_process = Popen(self._server_base_command + "> " + server_outfile + " 2>&1", shell=True)
         sleep(1)
 
-        numclient = self._numproc - 1
-        client_process_list = [None] * numclient
-        client_outfile_list = [None] * numclient
-        for i in range(numclient):
+        client_process_list = [None] * self.numclient
+        client_outfile_list = [None] * self.numclient
+        for i in range(self.numclient):
             client_outfile_list[i] = NamedTemporaryFile(prefix="fabtests_client_{}.out.".format(i)).name
-            client_process_list[i] = Popen(self._client_command + "> " + client_outfile_list[i] + " 2>&1", shell=True)
+            try:
+                # Start client
+                # Retry on SSH connection error until server timeout
+                client_process_list[i] = retry(
+                    retry_on_exception=is_ssh_connection_error,
+                    stop_max_delay=self._timeout * 1000,  # Convert to milliseconds
+                    wait_fixed=CLIENT_RETRY_INTERVAL_MS,
+                )(self._run_client_command)(server_process, self._client_base_command_list[i],
+                                            client_outfile_list[i],
+                                            self._run_client_asynchronously)
+            except Exception as e:
+                print("Client error: {}".format(e))
+                # Clean up server if client is terminated unexpectedly
+                server_process.terminate()
 
         server_timed_out = False
         try:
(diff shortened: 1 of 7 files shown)


shijin-aws:
The AWS CI is expected to fail before https://github.com/ofiwg/libfabric/pull/8546 is merged

So this PR cannot be merged before the shm provider fix is merged.

prov/efa/src/efa_av.c
```diff
@@ -317,7 +317,17 @@ int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn)
...
+		/*
+		 * The shm provider supports FI_AV_USER_ID flag. This flag
+		 * associates a user-assigned identifier with each av entry that is
+		 * returned with any completion entry in place of the AV's address.
+		 * EFA provider could use efa's fi_addr as the user id (passed as
+		 * as an input to the *fi_addr argument in fi_av_insert call)
+		 * With this change, there is no need to translate shm address back 
+		 * to efa address when reading shm cq entry.
+		 */
+		peer->shm_fiaddr = conn->fi_addr;
```
wzamazon, line 329: should we remove `shhm_fiaddr` from `efa_rdm_peer`?

shijin-aws: No, we still need it. The `peer->shm_fiaddr` returned by `fi_av_insert` is still the addr assigned by shm provider, which could be different from efa_fi_addr. efa code should still use `peer->shm_fiaddr` for shm sendmsg. 

The user id will only show in the cq_entry->addr. I will add comments to explain this.

fabtests/prov/efa/src/efa_src_addr_test.c
```diff
@@ -0,0 +1,267 @@
+/*
+ * Copyright (c) 2021, Amazon.com, Inc.  All rights reserved.
```
wenduwan, line 2: The current guideline is to be dateless, i.e.
```
Copyright Amazon.com Inc. or its affiliates
```

prov/efa/src/efa_av.c
```diff
@@ -317,7 +317,17 @@ int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn)
...
+		ret = fi_av_insert(av->shm_rdm_av, smr_name, 1, &peer->shm_fiaddr, FI_AV_USER_ID, NULL);
```
wenduwan, line 330: Clarify - according to this statement
> EFA provider could use **efa's fi_addr** as the user id (passed as as an input to the *fi_addr argument in fi_av_insert call)

It sounds like we should use `&peer->efa_fiaddr` instead of `&peer->shm_fiaddr`. Could you explain?

fabtests/pytest/efa/efa_common.py
```diff
@@ -122,3 +128,79 @@ def get_efa_device_names(server_id):
         assert itemlist[1] in ["rdm", "dgrm"]
         device_names.add(itemlist[0])
     return list(device_names)
+
+
+class EFAMultinodeClientServerTest:
+    """
+    This class establishes tests involving a single server process
+    and multiple client processes which can be initiated on multiple
+    nodes (configured via client_host_name_list).
```
wenduwan, line 137: -> `client_hostname_list`

fabtests/pytest/efa/efa_common.py
```diff
@@ -122,3 +128,79 @@ def get_efa_device_names(server_id):
...
+    """
+
+    def __init__(self, cmdline_args, base_command, client_hostname_list):
+        self._cmdline_args = cmdline_args
+        self._base_command = base_command
+        self._timeout = self._cmdline_args.timeout
+        self.client_hostname_list = client_hostname_list
+        self.numclient = len(client_hostname_list)
+        self._server_command = cmdline_args.populate_command(self._base_command + " -C {}".format(self.numclient), "server", self._timeout)
+        assert(isinstance(client_hostname_list, list) and len(client_hostname_list) >= 1)
```
wenduwan, line 147: Move L147 before L141. Input validation should happen earlier. Otherwise the program could panic on L145 if `client_hostname_list=None`.

prov/efa/src/efa_av.c
shijin-aws: The `peer->shm_fiaddr` here is both an input and an output. When FI_AV_USER_ID flag is used, shm will record the passed in `peer->shm_fiaddr` (which was assigned as the conn->fi_addr) as user id. Then shm will overwrite `peer->shm_fiaddr` to the real shm addr. We cannot use `&peer->efa_fiaddr` in `fi_av_insert` call because it records the efa fi_addr and we do not want it to be overridden.

fabtests/pytest/efa/efa_common.py
```diff
@@ -122,3 +128,79 @@ def get_efa_device_names(server_id):
...
+
+    def run(self):
+        if self._cmdline_args.is_test_excluded(self._base_command):
+            pytest.skip("excluded")
+
+        server_outfile = NamedTemporaryFile(prefix="fabtests_server.out.").name
+
+        # start running
+        server_process = Popen(self._server_command + "> " + server_outfile + " 2>&1", shell=True)
+        sleep(1)
```
wenduwan, line 163: For robustness, the server startup could be slow and we should allow the client to retry if it fails to connect to the server similar to [this](https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/common.py#L303)

Maybe we should consider inheriting the current ClientServerTest, e.g. `class MultiClientServerTest(ClientServerTest)`. WDYT?

fabtests/pytest/efa/efa_common.py
```diff
@@ -1,6 +1,12 @@
 import subprocess
-from common import SshConnectionError, is_ssh_connection_error, has_ssh_connection_err_msg
+from common import SshConnectionError, is_ssh_connection_error, has_ssh_connection_err_msg, check_returncode_list
 from retrying import retry
+import copy
+from subprocess import Popen, TimeoutExpired, run
+from tempfile import NamedTemporaryFile
+from time import sleep
+import os
```
wenduwan, line 8: Suggest `isort efa_comment.py`

shijin-aws:
>For robustness, the server startup could be slow and we should allow the client to retry if it fails to connect to the server similar to [this](https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/common.py#L303)
Maybe we should consider inheriting the current ClientServerTest, e.g. class MultiClientServerTest(ClientServerTest). WDYT?

That is a good idea, will look into it.

fabtests/pytest/efa/efa_common.py
shijin-aws: Done. See latest revision.

prov/efa/src/efa_av.c
shijin-aws: Added comment in latest revision.

prov/efa/src/efa_av.c
shijin-aws: Added comment in latest revision.

wenduwan (approved):
Thank you!

fabtests/prov/efa/src/efa_src_addr_test.c
```diff
@@ -0,0 +1,267 @@
+/*
```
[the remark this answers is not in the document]
shijin-aws: See the documentation in the code:

```
The main difference between this program and fi_rdm_multi_client is as follows:
 * 1. This program makes the server reset the oob when a client leaves, which means
 * the server will wait for the subsequent clients to connect rather than re-using
 * the oob channel created earlier. This allows a persistent server process on one 
 * node, and multiple client process created in sequence on different nodes.
 * 2. This program will check the remote source address returned by fi_cq_readfrom
 * call on the receiver side.
```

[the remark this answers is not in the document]
shijin-aws: Maybe I should consider making `fi_rdm_multi_client` more flexible so I could just reuse it.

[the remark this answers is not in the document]
shijin-aws: Reuse `fi_rdm` in latest revision

fabtests/pytest/efa/efa_common.py
```diff
@@ -122,3 +134,91 @@ def get_efa_device_names(server_id):
         assert itemlist[1] in ["rdm", "dgrm"]
         device_names.add(itemlist[0])
     return list(device_names)
+
+
+class EFAMultinodeClientServerTest(ClientServerTest):
```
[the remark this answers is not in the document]
shijin-aws: Done, see latest revision.

fabtests/pytest/common.py
```diff
@@ -300,20 +300,30 @@ def prepare_base_command(self, command_type, executable,
 
         return command
 
-    def _run_client_command(self, server_process):
+    def _run_client_command(self, server_process, client_command, client_output_file=None,
```
wenduwan, line 303: `client_output_file` seems unused in this change?

wenduwan: NVM saw the use case.

fabtests/pytest/common.py
```diff
@@ -300,20 +300,30 @@ def prepare_base_command(self, command_type, executable,
...
+                            run_client_asynchronously=False):
 
         if server_process.poll():
+            if client_output_file:
+                with open(client_output_file, "w") as f:
+                    f.write("")
```
wenduwan, line 309: Similar logic can be simplified/more robust if we change function signature to
```python
from typing import IO, Optional # Probably not supported in 3.6 - just to show type.
...
def _run_client_command(self, ..., client_output_file: Optional[IO],..):
    ...
    process = Popen(..., stdout=client_output_file, stderr=subprocess.STDOUT, ....)
    ...
```
In this case the caller needs to open and close the file handles, e.g.
```python
        for i in range(self.numclient):
            client_outfile_handle_list[i] = NamedTemporaryFile(prefix="fabtests_client_{}.out.".format(i))
            ...
            try:
                # Start client
                # Retry on SSH connection error until server timeout
                client_process_list[i] = retry(
                    retry_on_exception=is_ssh_connection_error,
                    stop_max_delay=self._timeout * 1000,  # Convert to milliseconds
                    wait_fixed=CLIENT_RETRY_INTERVAL_MS,
                )(self._run_client_command)(server_process, client_command,
                                            client_outfile_handle_list[i],
                                            self._run_client_asynchronously)
    ....
    for f in client_outfile_handle_list:
        f.close()
```

fabtests/pytest/common.py
```diff
@@ -384,34 +394,58 @@ def run(self):
...
-        numclient = self._numproc - 1
-        client_process_list = [None] * numclient
-        client_outfile_list = [None] * numclient
-        for i in range(numclient):
+        client_process_list = [None] * self.numclient
+        client_outfile_list = [None] * self.numclient
+        for i in range(self.numclient):
             client_outfile_list[i] = NamedTemporaryFile(prefix="fabtests_client_{}.out.".format(i)).name
-            client_process_list[i] = Popen(self._client_command + "> " + client_outfile_list[i] + " 2>&1", shell=True)
+            client_command = self._client_base_command_list[i]  + " 2>&1"
```
wenduwan, line 434: `2>&1` is redundant because of `stderr=STDOUT`

fabtests/pytest/common.py
shijin-aws: Since  this requires certain python version, can we make it a separate change in the future?

fabtests/pytest/common.py
shijin-aws: Thank you! Done

fabtests/pytest/common.py
wenduwan: Just to be clear, we don't need to import typing which requires a higher version - it was provided FYI. We can still implement the logic though.

shijin-aws: I am still not sure what's the benefits of using this handle?

wenduwan: Passing the file handle to the client process means output is captured in the file automatically.

Therefore, we do not need to handle edge cases when the process is interrupted(or not) and write output manually to the output file, e.g.
```python
if client_output_file:
    # Currently this has to be added manually case by case.
    # Instead, pass the output file to subprocess and let it write to the file directly.
    with open(...):
        ...
```

shijin-aws: `client_outfile_handle_list[i] = NamedTemporaryFile(prefix="fabtests_client_{}.out.".format(i)` this is filename, not handle?

shijin-aws: you mean `open(NamedTemporaryFile(prefix="fabtests_client_{}.out.".format(i))`

shijin-aws: One blocker is if I use the client_outfile_handle as stdout, there will be no output in `process.communicate`, and the following code that use `output` will fail as it is None.

wenduwan: > One blocker is if I use the client_outfile_handle as stdout, there will be no output in `process.communicate`, and the following code that use `output` will fail as it is None.

I see. In that case it is troublesome to use the file handle. I'm ok with passing file name then.

wenduwan (approved):

shijin-aws:
@wzamazon do you have more comments? Thx.

wzamazon (approved):

PR #8559, opened 2023-02-22
prov/efa: pin gdr buffer when gdrcopy is available

wenduwan:
gdrcopy is found to provide faster memory copy for small data sizes. This change allows gdr memory pinning when gdrcopy is enabled, regardless of FI_HMEM_CUDA_ENABLE_XFER flag.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -195,24 +195,6 @@ static struct fi_ops efa_mr_cache_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
-/**
- * @brief determine whether gdrcopy should used for a MR
- *
- * gdrcopy should be used for CUDA memory when cudaMemcpy
- * is not allowed, and gdrcopy is available. cudaMemcpy
- * has bettern host to device performance.
- *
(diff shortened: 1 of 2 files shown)


wenduwan:
> When FI_HMEM_CUDA_ENABLE_XFER is set, currently efa provider will only use cudaMemcpy. What is the meaning of this PR then?

Prior to this change, we assume that cudaMemcpy always provides better performance than gdrcopy, and is *safer* than the latter. Now based on our data points we believe gdrcopy is safe to use on its own, and we  also observe advantage of it especially for copying smaller message sizes.

Unfortunately, when we implemented FI_HMEM_CUDA_ENABLE_XFER, we decided to disable gdrcopy even when it was available. This means the buffer memory is never pinned with gdr and we do not have the gdr memory handle. Therefore we unable to take advantage of gdrcopy to copy small messages.

After this change, we always register the memory with gdr (even if `FI_HMEM_CUDA_ENABLE_XFER=1`)  and store the memory handle in `efa_mr.peer.device.cuda` - this allows us to use gdrcopy for small data sizes.

shijin-aws (approved):

PR #8562, opened 2023-02-23
prov/shm: Store IPC availability in smr_domain

vidsouza:
This patch serves as a prerequisite to an upcoming patch. The logic of the function cuda_is_ipc_enabled() will be changed to fetch the value of an environment variable. With that change the calls to ofi_hmem_is_ipc_enabled() in smr_generic_rma() and smr_generic_sendmsg() will be costly.

This patch eliminates this problem by storing the values IPC availbility in smr_ep structure.

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -190,6 +190,7 @@ struct smr_domain {
 	/* cache for use with hmem ipc */
 	struct ofi_mr_cache	*ipc_cache;
 	struct fid_peer_srx	*srx;
+	bool			hmem_is_ipc_enabled[OFI_HMEM_MAX];
 };
 
 #define SMR_PREFIX	"fi_shm://"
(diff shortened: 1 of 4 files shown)


shefty:
See #8529.

wzamazon (approved):

wzamazon:
@aingerson  can you take a look?

shefty:
#8529 changes the function to read the value dynamically, then this PR changes the provider so that it saves the value from the first time it's read.  I don't understand the reason behind either of these changes.

aingerson:
I was about to say the same thing as Sean. I'm confused about why this is needed

wzamazon:
It is that some middleware want to set the environment variable, so the enduser does not need to set it. But libfabric read the environment variable during library initialization, at which point the middleware may has not been initialized yet.

vidsouza:
To elucidate, with the change @shefty suggested in #8529, `cuda_is_ipc_enabled` will call `cuda_get_xfer_setting()` which fetches the value of environment variable with the change in #8529. In SHM provider, everytime `smr_generic_rma()` or `smr_generic_sendmsg` is called, a call to `cuda_is_ipc_enabled` will be made and with changes in #8529, these calls will degrade the performance (Since this will now make fetching the environment variable calls). With this PR, such calls will not be made.

shijin-aws (approved):

PR #8566, opened 2023-02-24
configure: Fix the handling of some '--enable-<feature>' options

### LINKED ISSUE #8222: configure: --disable-{cuda,rocr}-dlopen *enable* the respective dlopen behavior
Using (at least) versions 1.12.1, 1.15.1 or 1.16.1 of libfabric, I find that `--disable-rocr-dlopen` and `--disable-cuda-dlopen` actually cause `config.h` to define `ENABLE_ROCR_DLOPEN` or `ENABLE_CUDA_DLOPEN` to `1` instead of `0`.

```shell
$ [...]/libfabric-1.16.1/configure --with-rocr=/opt/rocm-5.1.0 --disable-rocr-dlopen >/dev/null
$ grep ROCR_DLOPEN config.h
#define ENABLE_ROCR_DLOPEN 1
```

The default (passing neither `--enable-*-dlopen` nor `--disable-*-dlopen) works fine.  So this is not a serious problem.

j-xiong:
With the `--enable-<feature>` option, it is possible to disable the feature by passing `--enable-<feature>=no` or `--disable-<feature>`. The handling of some of such options (e.g. `--enable-cuda-dlopen`) doesn't take that into account and may incorrectly enable the feature when shouldn't.

Fix by checking the value when the option is given before enabling the feature.

Also fix a few places where a prefix is added to only one side of string comparison.

Fix #8222

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -567,9 +567,9 @@ AC_ARG_ENABLE([cuda-dlopen],
         [Enable dlopen of CUDA libraries @<:@default=no@:>@])
     ],
     [
-        AS_IF([test x"$with_dlopen" = "no"],
+        AS_IF([test "$with_dlopen" = "no"],
               [AC_MSG_ERROR([dlopen not found.  libfabric requires libdl.])])
-	cuda_dlopen=1
+	AS_IF([test "$enable_cuda_dlopen" != "no"], [cuda_dlopen=1])
     ])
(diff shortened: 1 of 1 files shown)


shefty (approved):
Looks like copy-paste error throughout configure.ac.

Issue #8505, opened 2023-02-15
prov/tcp: Need improved multi-threaded support driving progress with EQ


shefty:
From @sydidelot 

One thing I noticed is a lock when the CQ is polled (libfabric is configured with FI_THREAD_COMPLETION and both control_progress and data_progress as set to data_progress).
This lock is initialized in xnet_init_locks():

```
/* Because we may need to start the progress thread to support blocking CQ      
 * or EQ calls, we always need to enable an active lock, independent from       
 * the threading model requested by the app.                                    
 */                                                                             
static int xnet_init_locks(struct xnet_progress *progress, struct fi_info *info)
{                                                                               
    enum ofi_lock_type base_type, rdm_type;                                     
    int ret;                                                                    
                                                                                
    if (info && info->ep_attr && info->ep_attr->type == FI_EP_RDM) {            
        base_type = OFI_LOCK_NONE;                                              
        rdm_type = OFI_LOCK_MUTEX;                                              
        progress->active_lock = &progress->rdm_lock;                            
    } else {                                                                    
        base_type = OFI_LOCK_MUTEX;                                             
        rdm_type = OFI_LOCK_NONE;                                               
        progress->active_lock = &progress->lock;                                
    }                                                                           
                                                                                
    ret = ofi_genlock_init(&progress->lock, base_type);                         
```
The comment above the function states that the lock is needed if we start the progress thread. However, progress thread is not started if attr->wait_obj == FI_WAIT_FD and epoll is supported. I think we could disable the locking in this case.

---
I believe the above comment is stale, resulting from earlier changes in the code before improved wait support was added.  We need to revisit that section of the code and see if locks are needed at all.

sydidelot:
Thanks for the issue!

sydidelot:
@shefty I ran some more tests with the new TCP provider and I clearly see a performance regression compared to the former TCP provider because of this lock. First of all, let me give you some details on how we use libfabric at DDN: our application creates multiple threads, and all of them run on separate cores. Each thread runs with its own libfabric domain with the following domain's attributes:
```
control_progress = FI_PROGRESS_MANUAL;
data_progress = FI_PROGRESS_MANUAL;
threading = FI_THREAD_COMPLETION:
```

In addition, each thread has its own collection of endpoints and CQs that are associated with the domain it owns. As a consequence,  domains, EPs and CQs are always accessed by the same thread.

[...]

PR #8570, opened 2023-02-26
prov/tcp: Move from fabric progress to EP progress

sydidelot:
Context: With the new TCP provider, polling an EQ results in progressing all the domains that belong to the same
fabric + the fabric itself. As a consequence, polling some EQs from the same fabric but from different threads would result
in a high contention on the fabric and the domain lock.

The PR moves the progress from the fabric to the EP. The intention behind this change is to improve multi-threaded applications that are driving progress with the EQ (see issue #8505 for more details).
With this PR, progressing an EQ results in progressing the domains that have EPs active for the given EQ only. As
a consequence, lock contention is reduced, and the performance of multi-threaded applications that drive progress of the
EQs is improved.

### DIFF
--- a/prov/tcp/src/xnet_domain.c
+++ b/prov/tcp/src/xnet_domain.c
@@ -133,7 +133,7 @@ static struct fi_ops_domain xnet_domain_ops = {
 	.query_collective = fi_no_query_collective,
 };
 
-static void xnet_del_wait_eq_list(struct xnet_domain *domain)
+static void xnet_domain_remove_eq_list(struct xnet_domain *domain)
 {
 	struct xnet_fabric *fabric;
 	struct xnet_eq *eq;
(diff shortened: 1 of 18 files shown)


shefty (commented):
I've started reviewing this, but have had limited time.  FWIW, I started with an implementation, but didn't quite finished.  I stopped once I noticed this PR.  You can see my start of changes at:

https://github.com/shefty/libfabric/commit/cf5a93da5d2aa9d3013fd7f9a2bfb2c3b8789eeb

One difference is that I added a new fid_list_search call to avoid updating all callers of fid_list_insert.  That's in the same branch, but the patch before.  I didn't get to the add/remove of the domain to the eq when binding an ep.  That's where I expected to start/stop progress as well.

The couple comments I did make here were based on noting differences between our implementations, and don't actually mean that what's here needs updating or is off.  I plan on getting back to this this afternoon (for me).

prov/tcp/src/xnet_domain.c
```diff
@@ -258,10 +257,10 @@ int xnet_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	if (ret)
 		goto close;
 
-	if (fabric->progress.auto_progress) {
+	if (domain->progress.auto_progress) {
 		ret = xnet_start_progress(&domain->progress);
 	} else {
-		ret = xnet_add_wait_eq_list(domain);
+		ret = xnet_domain_add_eq_list(domain);
```
shefty, line 263: Isn't this adding the domain to every EQ?  I think we only need to add from xnet_ep_bind() like you have below.

sydidelot:
@shefty I looked at your private branch and the changes look good to me! 
For the missing call to `xnet_start_all()`, I think we can add it when the EQ is created or when a domain is added to the EQ (where you have the TODO in your code). The latter solution seems better because we won't be creating a progress thread if no domain is added to the EQ. 
We should proceed with your implementation, which looks better in terms of coding and documentation. Could you please create a PR after you fix the missing call to `xnet_start_all()`? I will abandon my PR afterward.

shefty:
@sydidelot - Okay, I'll add the missing calls to my PR.  I'll delay creating a progress thread until it is needed, for both the EQ and domain.

sydidelot:
@shefty Thank you! Let me know when you have a PR ready. I'll be happy to review and test it!

shefty:
@sydidelot, see PR #8617.  Hopefully I didn't mess anything up.

sydidelot:
Thank @shefty! I'm abandoning this PR.

Issue #8308, opened 2022-12-03
core: Allow external control over provider ordering


shefty:
Provider ordering is currently hard-coded.  Allow the use of an external file to specify ordering criteria.  This feature was requested by end-users in order to set the defaults independent from using an environment variable.

tschuett:
Would it be feasible to pass an array of strings resp. provider names to `fi_getinfo` to express my preference of priority?

shefty:
For this request, that's not usable.  The intent with this request is to allow the system manager, which understands the network configuration, control over the ordering, so that applications do not need to change or modify their environments.

For what you're suggesting, I think it could be achieved by the app calling fi_getinfo multiple times, each time requesting a different provider, and stopping once they find one they want.

ooststep:
would a `.libfabricrc` in `/etc` or `%PROGRAMDATA%/libfabric` be feasible?  Is Windows a priority for this feature?

shefty:
Windows isn't a priority, but if it involves parsing a file, the OS shouldn't matter.

I guess the file could contain an ordered list of filters.  The output from fi_getinfo could pass through the filter, moving all filtered options to the front of the list.  For example, a simple filter could be:

```
{
    prov_name: net
},
{
    prov_name: tcp
}
```
The fi_getinfo() output would move all net provider options first, followed by tcp provider.  All others would follow those.

We could add more filtering and control options.  Example:
```
{
    prov_name: net,
    ep_type: [FI_EP_MSG, FI_EP_RDM],
    filter: first,
},
{
    prov_name: tcp,
    ep_type: FI_EP_RDM,
    filter: last,
},
{
    prov_name: udp,
    filter: remove
}
```

ooststep:
> Windows isn't a priority, but if it involves parsing a file, the OS shouldn't matter.
>
it doesn't matter for parsing the file, only locating it - since the requirement is essentially no run-time user interaction, libfabric has to know where to look.

> I guess the file could contain an ordered list of filters. The output from fi_getinfo could pass through the filter, moving all filtered options to the front of the list. For example, a simple filter could be:
>
We should probably distinguish util providers from core providers, too.  for example, in your first example, `net;rxm` would move below `tcp`.

shefty:
I'd make the filtering explicit, so, yeah, the second filter in my second example wouldn't actually match anything.

tschuett:
clang-format solves a similar challenge with priorities:
```
IncludeCategories:
[...]

PR #8572, opened 2023-02-27
core: Add basic system-config support

ooststep:
Add support for reading environment variables from a config file on the system.  The config file supports a single definition per line, defined as KEY=VALUE any whitespaces (other than '\n') will be preserved.

When loading environment variables, check this file for definitions as a fallback

Runtime definitions will overwrite system definitions

This targets an initial solution for #8308

### DIFF
--- a/src/var.c
+++ b/src/var.c
@@ -330,6 +330,15 @@ find_conf_entry(const char *param_name)
 	return NULL;
 }
 
+void ofi_dump_sysconfig(void)
+{
+	struct ofi_conf_entry *conf;
+	dlist_foreach_container(&conf_list, struct ofi_conf_entry, conf, entry) {
+		FI_INFO(&core_prov, FI_LOG_CORE,
+                        "Read config variable: %s=%s\n",conf->name, conf->value);
+	}
+}
+
 __attribute__((visibility ("default"),EXTERNALLY_VISIBLE))
 int DEFAULT_SYMVER_PRE(fi_param_get)(struct fi_provider *provider,
 		const char *param_name, void *value)
--- a/src/var.c
+++ b/src/var.c
@@ -42,6 +42,13 @@
 #include "ofi.h"
 #include "ofi_list.h"
 
+#ifdef SYSCONFDIR
+#define DEFAULT_CONF_FILE_PATH SYSCONFDIR "/libfabric.conf"
+#else
+#define DEFAULT_CONF_FILE_PATH "libfabric.conf"
+#endif
+
+#define MAX_CONF_LINE_LENGTH 2048
 
 extern void fi_ini(void);
 
@@ -54,8 +61,15 @@ struct fi_param_entry {
 	struct dlist_entry entry;
 };
 
+struct ofi_conf_entry {
+	char *name;
+	char *value;
+	struct dlist_entry entry;
+};
+
 /* TODO: Add locking around param_list when adding dynamic removal */
 static DEFINE_LIST(param_list);
+static DEFINE_LIST(conf_list);
 
 
 static struct fi_param_entry *
@@ -143,6 +157,13 @@ static void fi_free_param(struct fi_param_entry *param)
 	free(param->help_string);
 	free(param->env_var_name);
 	free(param);
+}
+
+static void free_conf(struct ofi_conf_entry *conf)
+{
+	free(conf->name);
+	free(conf->value);
+	free(conf);
 }
 
 void fi_param_undefine(const struct fi_provider *provider)
@@ -250,11 +271,70 @@ static int fi_parse_bool(const char *str_value)
 	return -1;
 }
 
+static void load_conf(void)
+{
+	char *tmp;
+	struct ofi_conf_entry *conf;
+	size_t line_len = 0;
+	char line[MAX_CONF_LINE_LENGTH];
+
+	dlist_init(&conf_list);
+
+	FILE *fptr = fopen(DEFAULT_CONF_FILE_PATH, "r");
+	if (!fptr)
+		return;
+
+	while (fgets(line, MAX_CONF_LINE_LENGTH, fptr) != NULL) {
+		tmp = strchr(line, '=');
+		if (!tmp)
+			continue;
+
+		line_len = strlen(line);
+		tmp[0] = '\0';
+		if (line[line_len-1] == '\n')
+			line[line_len-1] = '\0';
+
+		conf = calloc(1, sizeof(*conf));
+		if (!conf)
+			break;
+
+		conf->name = strdup(line);
+		if (!conf->name) {
+			free_conf(conf);
+			break;
+		}
+
+		conf->value = strdup(&tmp[1]);
+		if (!conf->value) {
+			free_conf(conf);
+			break;
+		}
+
+		dlist_insert_tail(&conf->entry, &conf_list);
+	}
+
+	fclose(fptr);
+}
+
+static struct ofi_conf_entry *
+find_conf_entry(const char *param_name)
+{
+	struct ofi_conf_entry *conf;
+
+	dlist_foreach_container(&conf_list, struct ofi_conf_entry, conf, entry) {
+		if (strcmp(conf->name, param_name) == 0)
+			return conf;
+	}
+
+	return NULL;
+}
+
 __attribute__((visibility ("default"),EXTERNALLY_VISIBLE))
 int DEFAULT_SYMVER_PRE(fi_param_get)(struct fi_provider *provider,
 		const char *param_name, void *value)
 {
 	struct fi_param_entry *param;
+	struct ofi_conf_entry *conf;
 	char *str_value;
 	int parsed_boolean;
 	int ret = FI_SUCCESS;
@@ -275,10 +355,16 @@ int DEFAULT_SYMVER_PRE(fi_param_get)(struct fi_provider *provider,
 
 	str_value = getenv(param->env_var_name);
 	if (!str_value) {
-		FI_INFO(provider, FI_LOG_CORE,
-			"variable %s=<not set>\n", param_name);
-		ret = -FI_ENODATA;
-		goto out;
+		FI_DBG(provider, FI_LOG_CORE,
+			"variable %s=<not set>, checking system\n", param_name);
+		conf = find_conf_entry(param->env_var_name);
+		if (!conf) {
+			FI_INFO(provider, FI_LOG_CORE,
+				"variable %s=<not set>\n", param_name);
+			ret = -FI_ENODATA;
+			goto out;
+		}
+		str_value = conf->value;
 	}
 
 	switch (param->type) {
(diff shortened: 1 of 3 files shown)


src/var.c
```diff
@@ -250,11 +267,76 @@ static int fi_parse_bool(const char *str_value)
...
+
+	// rather than reading the file every time a variable is queried
+	// cache the file during param initialization and check the cache
+	// when values are queried
+
+	dlist_init(&conf_list);
+
+	FILE *fptr = fopen(DEFAULT_CONF_FILE_PATH, "r");
+	if (!fptr) {
+		printf("couldn't open %s\n", DEFAULT_CONF_FILE_PATH);
```
ooststep, line 283: need to remove this printf

src/var.c
```diff
@@ -43,6 +43,9 @@
 #include "ofi_list.h"
 
 
+#define MAX_CONFIG_LINE_LENGTH  2048
+#define DEFAULT_CONF_FILE_PATH SYSCONFDIR "/ofi.conf"
```
ooststep, line 47: rename to libfabric.conf

ooststep (commented):
adding some notes to take care of before merge - avoiding cycling CI before further review

ooststep:
oops, need to use some OS abstraction to open/read the file.

src/var.c
```diff
@@ -250,11 +267,76 @@ static int fi_parse_bool(const char *str_value)
...
+	fclose(fptr);
+}
+
+static struct fi_conf_entry *
+find_conf_entry(const struct fi_provider *provider, const char *param_name)
+{
+	struct fi_conf_entry *conf;
+	struct dlist_entry *entry;
+
+	for (entry = conf_list.next; entry != &conf_list; entry = entry->next) {
```
j-xiong, line 322: use `dlist_foreach` or `dlist_foreach_container`.

src/var.c
```diff
@@ -250,11 +267,76 @@ static int fi_parse_bool(const char *str_value)
...
 }
 
+static void load_conf() {
+	char *line;
+	char *tmp;
+	struct fi_conf_entry *conf;
+
+	// rather than reading the file every time a variable is queried
+	// cache the file during param initialization and check the cache
+	// when values are queried
```
j-xiong, line 277: This comment is unnecessary.

ooststep:
> oops, need to use some OS abstraction to open/read the file.

I misunderstood the CI error at first. I think how I access the file should work, but the SYSCONFDIR preprocessor define seems undefined for Windows build.

src/var.c
```diff
@@ -54,8 +61,15 @@ struct fi_param_entry {
 	struct dlist_entry entry;
 };
 
+struct fi_conf_entry {
```
shefty, line 64: use ofi prefix, no need to make this worse

src/var.c
```diff
@@ -250,11 +271,67 @@ static int fi_parse_bool(const char *str_value)
...
+	FILE *fptr = fopen(DEFAULT_CONF_FILE_PATH, "r");
+	if (!fptr)
+		return;
+
+	line = calloc(1, MAX_CONFIG_LINE_LENGTH);
+	if (!line)
+		goto out;
+
+	while (fgets(line, MAX_CONFIG_LINE_LENGTH, fptr) != NULL) {
+		tmp = strtok(strtok(line, "\n"), "=");
```
shefty, line 290: Best to use strtok_r() in the off chance some other thread somewhere is invoking strtok.  Though, you could probably just use strchr() instead of strtok(), since we're searching for '=' and not overly concerned about mis-formatted data.

src/var.c
```diff
@@ -250,11 +271,67 @@ static int fi_parse_bool(const char *str_value)
...
+
+	FILE *fptr = fopen(DEFAULT_CONF_FILE_PATH, "r");
+	if (!fptr)
+		return;
+
+	line = calloc(1, MAX_CONFIG_LINE_LENGTH);
+	if (!line)
+		goto out;
+
+	while (fgets(line, MAX_CONFIG_LINE_LENGTH, fptr) != NULL) {
```
shefty, line 289: You can use getline() and avoid the hard-coded max line length define and separate line calloc.  getline() returns the length of the string, which makes it easy to check for and replace '\n' with '\0'.  The latter is needed to handle config files that don't end in a new line, which would be easy for a user to not add.

src/var.c
```diff
@@ -250,11 +271,67 @@ static int fi_parse_bool(const char *str_value)
...
+		tmp = strtok(strtok(line, "\n"), "=");
+		if (tmp) {
```
shefty, line 291: Use
```
if (!tmp)
    continue;
```
to avoid extra nesting

src/var.c
ooststep: done, I was just being consistent with what was there

src/var.c
ooststep: good call on strchr. seems a bit cleaner and safer

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
 	return -1;
 }
 
+static void load_conf() {
+	char *line = NULL;
+	char *tmp;
+	struct ofi_conf_entry *conf;
+	size_t length;
```
j-xiong, line 276: length needs to be initialized to 0.

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
...
+	if (!fptr)
+		return;
+
+	while ((read = getline(&line, &length, fptr)) != -1) {
+		conf = calloc(1, sizeof(*conf));
+		if (!conf)
+			break;
+
+		tmp = strchr(line, '=');
+		tmp[0] = 0;
```
j-xiong, line 291: need to handle the case that `=` is not found.

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
...
+		conf->value = strdup(&tmp[1]);
+		dlist_insert_tail(&conf->entry, &conf_list);
+	}
+
+	free(line);
+	fclose(fptr);
+}
+
+static struct ofi_conf_entry *
+find_conf_entry(const struct fi_provider *provider, const char *param_name)
```
j-xiong, line 304: `provider` is only used for debugging output. suggest to remove it from the parameter and move the debug output to the caller.

src/var.c
```diff
@@ -54,8 +59,15 @@ struct fi_param_entry {
 	struct dlist_entry entry;
 };
 
+struct ofi_conf_entry {
+	char *var_name;
```
j-xiong, line 63: nit-picking: just use `name`.

j-xiong (approved):

shefty:
Please update the commit message title to something like "core: Add support for system-config files".  I'm mostly focused on the 'core' part, so I can identify what component the patch applies to when running git log --oneline 6 months from now.

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
 	return -1;
 }
 
+static void load_conf() {
+	char *line = NULL;
+	char *tmp;
+	struct ofi_conf_entry *conf;
+	size_t length = 0;
+	ssize_t read;
```
shefty, line 277: I'd pick a different name than one that collides with a standard function name.  Maybe bytes?

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
...
+
+	FILE *fptr = fopen(DEFAULT_CONF_FILE_PATH, "r");
+	if (!fptr)
+		return;
+
+	while ((read = getline(&line, &length, fptr)) != -1) {
+		tmp = strchr(line, '=');
+		if (!tmp)
+			continue;
+		tmp[0] = 0;
```
shefty, line 289: Please use the character '\0' instead of an integer value here and below.

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
...
+		if (line[read-1] == '\n')
+			line[read-1] = 0;
+
+		conf = calloc(1, sizeof(*conf));
+		if (!conf)
+			break;
+		conf->name = strdup(line);
+		conf->value = strdup(&tmp[1]);
```
shefty, line 297: Need to check for strdup() failure

src/var.c
```diff
@@ -250,11 +269,58 @@ static int fi_parse_bool(const char *str_value)
 	return -1;
 }
 
+static void load_conf() {
```
shefty, line 272: Formatting:
```
static void load_conf(void)
{
```

shefty (commented):
I'm thinking we may want to add an option to the config file to force it to override the system values in case of a conflict.

src/var.c
ooststep: Getline is not part of c library, it's a POSIX function. it's unavailable on windows.  Should I have a special case for windows or revert back to fgets?

shefty: sigh... of course it isn't.  I'd revert back to fgets.  You could define getline for windows in the osd files as an option, but I'm not sure that's worth it.

ooststep:
> I'm thinking we may want to add an option to the config file to force it to override the system values in case of a conflict.

what system values?  environment defined values will override config defined values

shefty:
I'm suggesting a separate change that allows the config file to override environment variables

src/var.c
```diff
@@ -353,18 +354,37 @@ int DEFAULT_SYMVER_PRE(fi_param_get)(struct fi_provider *provider,
...
+			if (!conf) {
+				FI_INFO(provider, FI_LOG_CORE,
+					"variable %s=<not set>\n", param_name);
+				ret = -FI_ENODATA;
+				goto out;
+			}
+			str_value = conf->value;
 		}
-		str_value = conf->value;
 	}
```
j-xiong, line 388: To reduce code duplication, you can read both conf and env first, and choose the right one based on preference. Something like:
```
if (ofi_prefer_sysconfig)
    str_value = conf->value ? conf->value : env_value;
else
    str_value = env_value ? env_value : conf->value;

if (!str_value)
    FI_DBG(.....);
```

ooststep: I did think about this and can move to this if preferred. my concern was adding extra calls to query the environment or looping through the config cache.  I suppose the expectation is the cache should be pretty small so querying the environment would be the bigger concern, but we're doing that anyway today so maybe my concern was misplaced...

src/var.c
```diff
@@ -250,11 +271,70 @@ static int fi_parse_bool(const char *str_value)
...
+			break;
+		}
+
+		conf->value = strdup(&tmp[1]);
+		if (!conf->value) {
+			free_conf(conf);
+			break;
+		}
+
+		dlist_insert_tail(&conf->entry, &conf_list);
```
shefty, line 314: It would be nice to add FI_INFO() here using the core to print all read name-value pairs.  This can be added as a separate change.

shefty (approved):

src/var.c
ooststep: the logging interface requires an instance of a provider, so we can either dump this late (maybe during getinfo?) or we need to add a logging function with that requirement removed.

src/var.c
ooststep: updated to remove duplication!

src/var.c
ooststep: ah, tmyk... 

Other issue is that this takes place prior to fi_log_init, so the logging system hasn't been initialized. The logging system relies on these variables, so it seems a catch 22

shefty (approved):

PR #8576, opened 2023-02-28
infra: Add atomic indirection

amirshehataornl:
Add atomic indirection to handle cases where an older compiler is being used which doesn't support atomic functions, or have other versions of the atomic functions. The three variants which were added are:

atomic_compare_exchange_weak()
atomic_store_explicit()
atomic_load_explicit()

These are needed for the atomic queue implementation

### DIFF
--- a/include/ofi_atom.h
+++ b/include/ofi_atom.h
@@ -166,6 +166,25 @@ typedef atomic_long	ofi_atomic_int64_t;
 					int##radix##_t desired)						\
 	{												\
 		return ofi_atomic_cas_bool_strong##radix(atomic, expected, desired);			\
+	}												\
+	static inline											\
+	bool ofi_atomic_compare_exchange_weak##radix(ofi_atomic##radix##_t *atomic, 			\
+					int##radix##_t *expected,					\
+					int##radix##_t desired)						\
+	{												\
+		return atomic_compare_exchange_weak(&atomic->val, expected, desired);			\
+	}												\
+	static inline											\
+	void ofi_atomic_store_explicit##radix(ofi_atomic##radix##_t *atomic,				\
+					      int##radix##_t value, int memmodel)			\
+	{												\
+		atomic_store_explicit(&atomic->val, value, memmodel);					\
+	}												\
+	static inline											\
+	int##radix##_t ofi_atomic_load_explicit##radix(ofi_atomic##radix##_t *atomic,			\
+						       int memmodel)					\
+	{												\
+		return atomic_load_explicit(&atomic->val, memmodel);					\
 	}
 
 #elif defined HAVE_BUILTIN_ATOMICS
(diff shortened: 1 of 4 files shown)


include/ofi_atom.h
```diff
@@ -249,6 +275,26 @@ typedef atomic_long	ofi_atomic_int64_t;
 					     int##radix##_t desired)					\
 	{												\
 		return ofi_atomic_cas_bool##radix(atomic, expected, desired);				\
+	}												\
+	static inline											\
+	bool ofi_atomic_compare_exchange_weak##radix(ofi_atomic##radix##_t *atomic,			\
+					int##radix##_t *expected,					\
+					int##radix##_t desired)						\
+	{												\
+		return __atomic_compare_exchange_n(ofi_atomic_ptr(atomic), expected, desired,		\
```
j-xiong, line 284: This (and the other two) may not be available on all OSes. Please go back a few blocks to see how `atomic_add_and_fetch` is handled (with definitions in `osd.h`).

amirshehataornl:
@shefty, We probably should build libfabric on all supported platforms/OSes, including Centos7. Also do you guys maintain a list of official platforms/OSes supported?

include/ofi_atom.h
amirshehataornl: @j-xiong, I'm unsure what the windows equivalent of these functions should be.

What I have is below. I think the compare exchange and store should work. But I don't know what the load should be? I can't find an interlockedAcquire() function. Any ideas?

```
1034 #define ofi_atomic_compare_exchange_weak(radix, ptr, expected, desired, \
1035 »·······»·······»·······»·······»······· succ_memmodel, fail_memmodel) \
1036 »·······InterlockedCompareExchange((ofi_atomic_int_##radix##_t volatile *)ptr, desired, expected)
1037 #define ofi_atomic_store_explicit(radix, ptr, value, memmodel) \
1038 »·······InterlockedExchange((ofi_atomic_int_##radix##_t volatile *)ptr, value)
1039 #define ofi_atomic_load_explicit(radix, ptr, memmodel) \
1040 »·······__atomic_load_n(ptr, memmodel)
```

wzamazon: IIUC, this code is only used by shm. Right? Can we disable shm compilation on windows? I do not think shm works on windows any way.

aingerson: shm shouldn't be building on windows. It will only build if CMA is found which is only on linux
https://github.com/ofiwg/libfabric/blob/main/prov/shm/configure.m4#L89

shefty: This file and the functions in it are built on windows and used there.

I'm not sure what the windows equivalent to load is, but guessing you could use Add(0) or just return the value.

shefty:
Each provider typically has their own list of supported platforms.  For the general code, we try to at least build on as many platforms as reasonable.  The documentation states this:
```
Libfabric targets support for the Linux, Free BSD, Windows, and OS X. A reasonable effort is made to support all major, modern Linux distributions; however, validation is limited to the most recent 2-3 releases of Red Hat Enterprise Linux (RHEL)and SUSE Linux Enterprise Server (SLES). Support for a particular operating system version or distribution is vendor specific. The exceptions are the tcp and udp based socket providers are available on all platforms.
```
But I wrote that, so your mileage may vary.  I, personally, run on ubuntu (which carried over from kernel work), but most of our systems are redhat or suse based.

And I agree that we should at least try to build on as many platforms as possible.  I don't know what github actions makes available.

amirshehataornl:
@shefty @aingerson can someone let me know what's failing in the continuous-integration?

aingerson:
@amirshehataornl everything passed except our DAOS testing but I'm not sure why. Rerunning to see if it's consistent.

wzamazon:
All tests passed. Can we merge?

PR #8577, opened 2023-02-28
prov/efa: select eager rtm protocol for small intra-node messages via shm

wenduwan:
This PR aims to improve single-node latency for small cuda message sizes <4k by taking advantage of gdrcopy. It contains 2 related changes:
1. Switch from longread to eager rtm protocol for intro-node messages < 4k via shm.
1. Rename the `iov_accessible_by_device` to explicitly state our intention of copying hmem data to bounce buffer before sending it to device(in this context shm).

The result is to allow prov/efa to copy data in/out of cuda buffer with gdrcopy when it is available, which has lower latency for small data sizes.


lrbison (approved):
thank you

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -111,8 +111,7 @@ int rxr_msg_select_rtm(struct rxr_ep *rxr_ep, struct rxr_op_entry *tx_entry, int
 		 * Have the remote side issue a read to copy the data instead
 		 * to work around this issue.
 		 */
-		if (tx_entry->total_len > rxr_env.shm_max_medium_size ||
-		   (tx_entry->total_len > 0 && efa_mr_is_hmem(tx_entry->desc[0])))
+		if (tx_entry->total_len >= rxr_env.shm_max_medium_size)
```
wzamazon, line 114: `shm_max_medium_size` is 4k by default. I thought the optimum size for cuda memory is 2k. Right?

wenduwan: That's a good point. I should also mention that based on my experiment, eager rtm + gdrcopy is faster than cudaMemcpy up to 3k.

For that we can have another threshold, e.g. in `efa_hmem_info` instead of `rxr_env.` I preferred not to do that initially for simplicity.

Do you think the additional threshold makes more sense? We can set it to 2k as you suggested as well.

wzamazon: I think we should use a threshold that is optimum

wenduwan: My understanding of gdrcopy is that, its performance is also affected by the CPU speed. This could make it hard to find a general optimal threshold for all platforms.

We can set the threshold to 3k based on the data on p4d for now then.

wenduwan:
AWS CI is failing on ubuntu 18.04

prov/efa/src/efa_hmem.c
```diff
@@ -67,6 +67,7 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 	case FI_HMEM_SYSTEM:
 		/* We have not yet tested runting with system memory */
 		info->runt_size = 0;
+		info->max_gdrcopy_from_device_size = EFA_DEFAULT_INTRA_MAX_GDRCOPY_FROM_DEV_SIZE;
```
shijin-aws, line 70: The name is not quite right. Actually we not only copy from device, we also gdrcopy from host to device in the recv side.

wenduwan: Yes this is on purpose. This value should only be used for device to host copy. gdrcopy claims asymmetrical performance between D2H and H2D. Here we are concerned with D2H.

shijin-aws: But finally you still use gdrcopy on receive side. You silently ignore its impact.. And your change on the protocol selection seems not explicitly force gdrcopy to be used, i.e. if gdrcopy is not available, it will still use cudamemcpy for msg smaller than the threshold. Is that true?

wenduwan: The choice between gdrcopy and cudamemcpy always happens within prov/efa with this change.

>  if gdrcopy is not available, it will still use cudamemcpy for msg smaller than the threshold. Is that true
This is a good point. Let me think about it.

wenduwan: @shijin-aws  You are right. The previous logic will cause regression if gdrcopy is not available - we would do 2 cudaMemcpy, which would double the latency.

I have updated the change to separate the threshold selection logic for FI_HMEM_SYSTEM and FI_HMEM_CUDA - and check if gdrcopy is enabled or not.

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -96,26 +96,27 @@ int rxr_msg_select_rtm(struct rxr_ep *rxr_ep, struct rxr_op_entry *tx_entry, int
...
+		 * This applies to data on both system and cuda buffers.
+		 * The threshold is determined from osu_latency result on p4d.24xlarge.
 		 */
-		if (tx_entry->total_len > rxr_env.shm_max_medium_size ||
-		   (tx_entry->total_len > 0 && efa_mr_is_hmem(tx_entry->desc[0])))
-			return RXR_LONGREAD_MSGRTM_PKT + tagged;
+		if (iface == FI_HMEM_SYSTEM && tx_entry->total_len < rxr_env.shm_max_medium_size) {
+			return RXR_EAGER_MSGRTM_PKT + tagged;
+		} else if (iface == FI_HMEM_CUDA && cuda_is_gdrcopy_enabled() &&
+					tx_entry->total_len <= hmem_info[FI_HMEM_SYSTEM].max_gdrcopy_from_device_size) {
```
shijin-aws, line 115: indent is off.

prov/efa/src/efa_hmem.c
```diff
@@ -78,6 +79,7 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 	case FI_HMEM_CUDA:
 	case FI_HMEM_NEURON:
 		info->runt_size = EFA_DEFAULT_RUNT_SIZE;
+		info->max_gdrcopy_from_device_size = 0;
```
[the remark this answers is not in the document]
wenduwan: > I do not think we have to maintain two thresholds

Are you referring to `max_gdrcopy_from_device_size` vs `shm_max_medium_size`, or something else?

Personally I think the current name makes more sense in:
```c
iface == FI_HMEM_CUDA &&
cuda_is_gdrcopy_enabled() &&
tx_entry->total_len <= hmem_info[FI_HMEM_SYSTEM].max_gdrcopy_from_device_size
```
which is more explicit about cuda and gdr.

A more generic name `max_copy_from_device_size`  obfuscates the intention. Meanwhile, so far I don't know the use case for a max limit for cudaMemcpy - it is the defacto method for copying large cuda messages AFAIK.

Please correct me if I was mistaken.

prov/efa/src/rdm/rxr_msg.c
wenduwan: Thanks. Fixed.

prov/efa/src/efa_hmem.c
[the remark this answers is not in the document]
shijin-aws: I was referring to `max_gdrcopy_from_device_size` and `max_cudamemcopy_from_device_size` (if we care). I do not think it hurt to use a more generic name as long as we make sure to use the correct copy method in the protocol selection. `ofi_hmem_memcpy` as a wrapper does not distinguish cudamemcpy and gdrcopy in its name. So using a generic name for `copy_from_device` makes sense to me.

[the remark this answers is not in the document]
shijin-aws: The macro `EFA_DEFAULT_INTRA_MAX_GDRCOPY_FROM_DEV_SIZE` does not need a rename because it is the size for gdrcopy. The field in `info` should be generic. All these discrepancy are due to the shenanigans of Nvidia. We do not have to make our code comply with that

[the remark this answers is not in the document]
wenduwan: > ofi_hmem_memcpy as a wrapper does not distinguish cudamemcpy and gdrcopy in its name

I see. It makes sense from ofi_hmem perspective. My perspective is to help the developer understand this threshold. IMO `max_gdrcopy_from_device_size` does better to remind the reader that this value is only relevant for gdrcopy, not cudaMemcpy. But I'm open to rename it once we have a use case for cudaMemcpy, and want to reuse this variable.

My intention for `info->max_gdrcopy_from_device_size = 0;` is to make it clear that gdrcopy does not support copying from cuda to this device(cuda or neuron) - does that make sense?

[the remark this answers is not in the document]
shijin-aws: Now it makes sense to me. Thanks.

shijin-aws (approved):

prov/efa/src/efa_hmem.c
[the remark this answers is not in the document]
wenduwan: Thank you. Added a comment to clarify the intention.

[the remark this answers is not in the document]
wenduwan: @shijin-aws I clarified with @wzamazon about the `efa_hmem_info` variable - the iface refers the user buffer type specifically. What we really want is a threshold for eager protocol based on the user buffer type.

Please see update.

wzamazon (approved):
LGTM

PR #8578, opened 2023-02-28
prov/efa: fix declaration of rxr_rma_should_write_using_rdma()

wzamazon:
"inline" should be before the return type. This patch fixed it.


wenduwan (approved):
Thank you

shijin-aws (approved):

lrbison (approved):
thank you

PR #8579, opened 2023-02-28
prov/shm: Update SHM to use ROCR

amirshehataornl:
Update the SHM provider to use the ROCR HMEM asynchronous memory operations.

   . Add an IPC entry freestack
   . When progressing an IPC operation, check if the device is ROCR and trigger
     an asynchronous operation.
   . When an asynchronous operation is queued, create an IPC entry and
     add it to the queue of pending operations.
   . During the top level progress loop check the queue of pending asynchronous
     operations and query the state of each one. Generate a complete event
     for finished operations. Since completions happen outside the context
     of libfabric we can't rely on the ep->region->signal flag to be set.
     Always check the pending queue. This shouldn't introduce much of an
     overhead if the queue is empty

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -135,7 +135,7 @@ struct smr_tx_entry {
 	int			fd;
 };
 
-struct smr_sar_entry {
+struct smr_pend_entry {
 	struct dlist_entry	entry;
 	struct smr_cmd		cmd;
 	struct fi_peer_rx_entry	*rx_entry;
@@ -145,6 +145,8 @@ struct smr_sar_entry {
 	size_t			iov_count;
 	struct ofi_mr		*mr[SMR_IOV_LIMIT];
 	bool			in_use;
+	struct ofi_mr_entry	*ipc_entry;
+	ofi_hmem_async_event_t	async_event;
 };
 
 struct smr_match_attr {
@@ -172,8 +174,8 @@ struct smr_cmd_ctx {
 
 OFI_DECLARE_FREESTACK(struct smr_rx_entry, smr_recv_fs);
 OFI_DECLARE_FREESTACK(struct smr_cmd_ctx, smr_cmd_ctx_fs);
-OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_pend_fs);
-OFI_DECLARE_FREESTACK(struct smr_sar_entry, smr_sar_fs);
+OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_tx_fs);
+OFI_DECLARE_FREESTACK(struct smr_pend_entry, smr_pend_fs);
 
 struct smr_queue {
 	struct dlist_entry list;
@@ -271,10 +273,10 @@ struct smr_ep {
 
 	struct fid_ep		*srx;
 	struct smr_cmd_ctx_fs	*cmd_ctx_fs;
+	struct smr_tx_fs	*tx_fs;
 	struct smr_pend_fs	*pend_fs;
-	struct smr_sar_fs	*sar_fs;
-
 	struct dlist_entry	sar_list;
+	struct dlist_entry	ipc_cpy_pend_list;
 
 	int			ep_idx;
 	struct smr_sock_info	*sock_info;
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -237,7 +237,7 @@ static void smr_progress_resp(struct smr_ep *ep)
 				"unable to process tx completion\n");
 			break;
 		}
-		ofi_freestack_push(ep->pend_fs, pending);
+		ofi_freestack_push(ep->tx_fs, pending);
 		ofi_cirque_discard(smr_resp_queue(ep->region));
 	}
 	ofi_spin_unlock(&ep->tx_lock);
@@ -420,13 +420,13 @@ static int smr_progress_mmap(struct smr_cmd *cmd, struct ofi_mr **mr,
 	return ret;
 }
 
-static struct smr_sar_entry *smr_progress_sar(struct smr_cmd *cmd,
+static struct smr_pend_entry *smr_progress_sar(struct smr_cmd *cmd,
 			struct fi_peer_rx_entry *rx_entry, struct ofi_mr **mr,
 			struct iovec *iov, size_t iov_count,
 			size_t *total_len, struct smr_ep *ep)
 {
 	struct smr_region *peer_smr;
-	struct smr_sar_entry *sar_entry;
+	struct smr_pend_entry *sar_entry;
 	struct smr_resp *resp;
 	struct iovec sar_iov[SMR_IOV_LIMIT];
 	int next = 0;
@@ -438,7 +438,7 @@ static struct smr_sar_entry *smr_progress_sar(struct smr_cmd *cmd,
 	(void) ofi_truncate_iov(sar_iov, &iov_count, cmd->msg.hdr.size);
 
 	ofi_ep_lock_acquire(&ep->util_ep);
-	sar_entry = ofi_freestack_pop(ep->sar_fs);
+	sar_entry = ofi_freestack_pop(ep->pend_fs);
 	sar_entry->in_use = true;
 	dlist_insert_tail(&sar_entry->entry, &ep->sar_list);
 	ofi_ep_lock_release(&ep->util_ep);
@@ -457,7 +457,7 @@ static struct smr_sar_entry *smr_progress_sar(struct smr_cmd *cmd,
 
 	if (*total_len == cmd->msg.hdr.size) {
 		dlist_remove(&sar_entry->entry);
-		ofi_freestack_push(ep->sar_fs, sar_entry);
+		ofi_freestack_push(ep->pend_fs, sar_entry);
 		ofi_ep_lock_release(&ep->util_ep);
 		return NULL;
 	}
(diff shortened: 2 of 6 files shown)


prov/shm/src/smr.h
```diff
@@ -173,7 +175,8 @@ struct smr_cmd_ctx {
 OFI_DECLARE_FREESTACK(struct smr_rx_entry, smr_recv_fs);
 OFI_DECLARE_FREESTACK(struct smr_cmd_ctx, smr_cmd_ctx_fs);
 OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_pend_fs);
-OFI_DECLARE_FREESTACK(struct smr_sar_entry, smr_sar_fs);
+OFI_DECLARE_FREESTACK(struct smr_pend_entry, smr_sar_fs);
+OFI_DECLARE_FREESTACK(struct smr_pend_entry, smr_ipc_fs);
```
aingerson, line 179: Since these are the same type and we should only ever need rx_size number of any of them, let's move this into one freestack (ie smr_pend_fs) and insert them into the same list (ie ep->rx_pend_list) that we can process differently based on the op. I think that will save memory and simplify the progressing code paths

prov/shm/src/smr_progress.c
```diff
@@ -477,9 +477,69 @@ static struct smr_sar_entry *smr_progress_sar(struct smr_cmd *cmd,
...
+		   struct ofi_mr_entry *mr_entry, struct smr_cmd *cmd, int *rc)
+{
+	struct smr_pend_entry *ipc_entry;
+	enum fi_hmem_iface iface = cmd->msg.data.ipc_info.iface;
+	uint64_t device = cmd->msg.data.ipc_info.device;
+	int ret;
+
+	ipc_entry = ofi_freestack_pop(ep->ipc_pend_fs);
+	if (!ipc_entry)
+		assert(0);
```
aingerson, line 492: Let's handle this better with a warn and return

prov/shm/src/smr_progress.c
```diff
@@ -519,6 +579,30 @@ static int smr_progress_ipc(struct smr_cmd *cmd, struct ofi_mr **mr,
...
+		if (!rx_entry) {
+			rx = calloc(sizeof(*rx), 1);
+			if (!rx) {
+				ret = -FI_ENOMEM;
+				goto out;
+			}
+
+			rx->iov = iov;
+			rx->count = iov_count;
+		}
```
aingerson, line 596: Why are we allocating an rx_entry every time? Just to save the iov/count? If we enter this if, that probably means we are doing an RMA which gets the iov from the cmd data instead of the rx_entry. The sar_entry struct has an iov/count for that reason. I don't think it is necessary to allocate a new one.

prov/shm/src/smr_progress.c
```diff
@@ -519,6 +579,30 @@ static int smr_progress_ipc(struct smr_cmd *cmd, struct ofi_mr **mr,
 		ptr = (char *) (uintptr_t) mr_entry->info.ipc_mapped_addr +
 		      (uintptr_t) cmd->msg.data.ipc_info.offset;
 
+	if (cmd->msg.data.ipc_info.iface == FI_HMEM_ROCR) {
+		struct smr_pend_entry *ipc_entry;
```
aingerson, line 583: Pull out to the top even though it is only used for the if

prov/shm/src/smr_progress.c
```diff
@@ -1062,18 +1149,65 @@ static void smr_progress_cmd(struct smr_ep *ep)
 	}
 }
 
+static void smr_progress_ipc_list(struct smr_ep *ep)
```
aingerson, line 1152: Merge with sar list progressing per first comment

prov/shm/src/smr_progress.c
```diff
@@ -1062,18 +1149,65 @@ static void smr_progress_cmd(struct smr_ep *ep)
...
+{
+	struct smr_pend_entry *ipc_entry;
+	struct smr_domain *domain;
+	struct dlist_entry *tmp;
+	int ret;
+
+	domain = container_of(ep->util_ep.domain, struct smr_domain,
+			      util_domain);
+
+	pthread_spin_lock(&ep->region->lock);
```
aingerson, line 1162: Do we need to lock the whole region for this? Can we just use the util_ep lock like the sar processing does?

prov/shm/src/smr_progress.c
```diff
@@ -1062,18 +1149,65 @@ static void smr_progress_cmd(struct smr_ep *ep)
...
+	/* after the synchronize all operations should be complete */
+	dlist_foreach_container_safe(&ep->ipc_cpy_pend_list,
+				     struct smr_pend_entry,
+				     ipc_entry, entry, tmp) {
+		enum fi_hmem_iface iface =
+			ipc_entry->cmd.msg.data.ipc_info.iface;
+		uint64_t device = ipc_entry->cmd.msg.data.ipc_info.device;
+
+		if (ofi_async_copy_query(iface, ipc_entry->async_event)
+				 != FI_SUCCESS)
```
aingerson, line 1173: Nit - move operand (!=) to first line before new line. Or I think you should just be able to check
`if (!ofi_async_copy_query(iface, ipc_entry->async_event))`

prov/util/src/rocr_mem_monitor.c
```diff
@@ -37,6 +37,7 @@
 
 #include "ofi_tree.h"
 #include "ofi_iov.h"
+#include "ofi_hmem.h"
```
aingerson, line 40: This looks unrelated. Split into separate patch

aingerson (commented):
Overall concept looks ok but would prefer some restructuring of the resources and progress

prov/shm/src/smr.h
amirshehataornl: The issue I see with this is that iterating through the ipc pend entries is fundamentally different from the sar entries. The completion of the async operations are external to the libfabric so it can't be gated by the signal. More over, I think it's cleaner to have separate lists for outstanding sar operations and ipc operations. They can use the same stack, but I believe the code is more straight forward if we queue them on separate lists. The processing functions of the operations is also different. I don't believe there is much of a memory overhead if we maintain the separate lists.

I'm okay with the rest of the comments. Will update.

prov/shm/src/smr_progress.c
amirshehataornl: as discussed in the earlier comment, it makes more sense to keep this separate.

amirshehataornl:
@aingerson can I get another review on the PR please. And wondering what the failure in the continuous integration is? The DAOS test again?
Also when a PR has an "updates requested" tag, is there something I'm suppose to do? remove the tag? Doesn't look like I'm able to make any changes of that sort.

prov/shm/src/smr.h
```diff
@@ -173,7 +175,7 @@ struct smr_cmd_ctx {
 OFI_DECLARE_FREESTACK(struct smr_rx_entry, smr_recv_fs);
 OFI_DECLARE_FREESTACK(struct smr_cmd_ctx, smr_cmd_ctx_fs);
 OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_pend_fs);
-OFI_DECLARE_FREESTACK(struct smr_sar_entry, smr_sar_fs);
```
aingerson, line 176: Not sure what opr is? Can we call this rx_pend_fs or something to indicate that it is to track receive side pending operations? And change the tx one to tx_pend_fs?

prov/shm/src/smr.h
```diff
@@ -145,6 +145,8 @@ struct smr_sar_entry {
 	size_t			iov_count;
 	struct ofi_mr		*mr[SMR_IOV_LIMIT];
 	bool			in_use;
+	struct ofi_mr_entry	*mr_entry;
```
aingerson, line 148: Since this is only used for the IPC path to get the handle from the IPC cache, let's rename to make that clear so it's not confusing with the above ofi_mr data.

prov/shm/src/smr_progress.c
```diff
@@ -519,6 +584,17 @@ static int smr_progress_ipc(struct smr_cmd *cmd, struct ofi_mr **mr,
 		ptr = (char *) (uintptr_t) mr_entry->info.ipc_mapped_addr +
 		      (uintptr_t) cmd->msg.data.ipc_info.offset;
 
+	if (cmd->msg.data.ipc_info.iface == FI_HMEM_ROCR) {
+		*total_len = 0;
+		ret = 0;
+		ipc_entry = smr_ipc_async_copy(ep,
+				(char*)ptr+cmd->msg.data.ipc_info.offset,
```
aingerson, line 591: Spaces around the +

prov/shm/src/smr_progress.c
```diff
@@ -1062,18 +1141,79 @@ static void smr_progress_cmd(struct smr_ep *ep)
...
+
+	ofi_ep_lock_acquire(&ep->util_ep);
+
+	/* after the synchronize all operations should be complete */
+	dlist_foreach_container_safe(&ep->ipc_cpy_pend_list,
+				     struct smr_pend_entry,
+				     ipc_entry, entry, tmp) {
+		enum fi_hmem_iface iface =
+			ipc_entry->cmd.msg.data.ipc_info.iface;
+		uint64_t device = ipc_entry->cmd.msg.data.ipc_info.device;
```
aingerson, line 1164: Move the definitions to the beginning of the function

aingerson:
@amirshehataornl CI failure is a compile error in the shm dsa code (see smr_dsa.c) - it accesses the sar entry directly so needs to be updated to use the pend_entry
prov/shm/src/smr_dsa.c:565:11: error: dereferencing pointer to incomplete type ‘struct smr_sar_entry’
sar_entry->bytes_done += dsa_cmd_context->bytes_in_progress;

prov/shm/src/smr.h
amirshehataornl: I renamed it to smr_tx_fs and smr_rx_fs respectively to remain consistent with the smr_ prepended to the other names. I also named the actual fields in the  smr_ep to tx_fs and rx_fs respectively

aingerson (approved):

prov/shm/src/smr.h
```diff
@@ -172,8 +174,8 @@ struct smr_cmd_ctx {
 
 OFI_DECLARE_FREESTACK(struct smr_rx_entry, smr_recv_fs);
 OFI_DECLARE_FREESTACK(struct smr_cmd_ctx, smr_cmd_ctx_fs);
-OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_pend_fs);
-OFI_DECLARE_FREESTACK(struct smr_sar_entry, smr_sar_fs);
+OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_tx_fs);
+OFI_DECLARE_FREESTACK(struct smr_pend_entry, smr_rx_fs);
```
shefty, line 178: It is confusing to see both a recv_fs and rx_fs.  It's doubly confusing to see that rx_entry is associated with the recv_fs, but tx_entry with tx_fs.

shefty:
Once you're both happy with the changes, I'm fine.  Just the naming is very confusing to me.

prov/shm/src/smr.h
aingerson: that's a very good point. can we do a bit more refactoring and take something like what zach did to merge the tx and pend entry into one progress (or pend) entry?
https://github.com/ofiwg/libfabric/pull/8281/commits/e1720109269a23bc980dd5114dba4adeed83b914
Then we can keep rx_entry -> rx_fs
and then have one type of pend_fs but two different instances of them - rx_pend_fs and tx_pend_fs?

amirshehataornl: Do you mean you want to do this refactoring in this patch? Wouldn't it make sense to add the refactoring in a separate PR as this is not related to the target of this PR?

aingerson: I was suggesting adding a patch within this series, between the first and the async one, to merge them but I'm ok with doing that as a follow up if you prefer

shefty: For this patch, maybe just rename rx_fs to pend_fs?

amirshehataornl: If you guys are okay with it, I'd like to rename the rx_fs->pend_fs as Sean suggested for this PR, then I can touch base with Alexia to more fully understand the refactoring she's suggesting. I don't mind creating a separate PR for that.

aingerson: That's fine with me. Thanks!

amirshehataornl:
bot:aws:retest

amirshehataornl:
@aingerson, Does the PR look good now?

PR #8582, opened 2023-03-01
prove/sm2: pt2pt Nemesis queue implementation

a-szegel:
I found some blocking issues when attempting to put in the full nemesis implementation.  I was trying to reuse all of the SHM provider's initialization code, and realized that wouldn't be possible. The main issue is that each process needs to have every other processes SHM region mapped, and we only map pt2pt as needed.  A different dev is doing the work to fix this problem in parallel, which will let me keep evaluating the Nemesis algorithm.


a-szegel:
Converting this to a draft b/c it only works pt2pt... but feedback is welcome!

shefty:
If all you're concerned with is small transfers, only test using small transfers and don't merge anything until a full set of test passes.  You're not trying to remove 'complexity'.  You're removing significant features, like transfers over 4k.

For the code: Use tabs, not spaces, for indentation.  Limit line length to about 80 characters.  Braces for functions go on a line by themselves.

There are so many changes being made here, that this isn't even really the same provider.  And the changes are sweeping, not small, self-contained patches.

Also, you can't add a copyright for the year 2017 or anything earlier, unless that was actually missed by some earlier change.  I don't think anyone actually needs to include a year anyway...

a-szegel:
Hey @shefty and @aingerson,

Thanks for feedback!  

I understand (and agree) with not wanting to merge this until it is working for the entire suite of tests.  Right now, SM2 in the main branch is scoped down to be 4k without DSA, accelerator, CMA, and inline  (inject only, and advertised honestly through capabilities).  The removal in this PR is a combination of things that I missed on the initial commit, and things that aren't being used anymore.

My initial thought was that a pt2pt nemesis solution would be more valuable than what we have in their currently (but I have been convinced otherwise).

 I care about much more than small message transfers, and I will put all the major capabilities back (I promise, although I might need some help testing them b/c idk if I have access to a DSA). My reason for removing all of the capabilities we care about deeply is speed. I wanted to prove to myself that it was as simple as I thought it would be to get inject working (and I was wrong :)).  Luke's working on the fix to the initialization code (basically everyone needs everyone else's memory mapped during initialization b/c the fifo receive queue (linked list) spans across N processes shared memory segments).

1. I updated copyrights (will push in a sec)
2. Removed tabs (vs code defaults to whatever is in the file, but new files get spaces)
3. I did the changes incrementally (saved in a branch), but re-based to something that I thought would be more readable... I can work on doing a better job on the history fixup

```
0d9f8a44 Remove sm_region lock
cb88616e Remove Signal optimization
7dbbc6eb Development commit
f940d044 Remove more RMA stuff
b54a9c82 Remove CMA vars from sm2_region
27d49c76 Remove cmdcount from sm2_region
1c190a25 Remove sar stuff
2773594b Delete response queue
5890f132 Cleaning up sm2_region
c6aefa3a Compiles, and makes it to infinite loop on recv, and into shm_send_name on send
5acdf572 Rename cmd_queue -> recv_queue and inject_pool -> free_stack
d8f90808 More minor cleanup deletions
7b3cff91 Change shm2_cmd to sm2_free_queue_entry
35436ab8 Rename cmd to fqe
1dcb58fc Cleanup sm2_common.h (rename stuff, and delete unused stuff to make it make sense for nemesis protocol
f93d89d7 Lots more cleanup... still compiles, starting to get ready to get full transfer going.
62c45140 Development commit
d94bb7a6 Code compiles but doesn't run b/c hack is only halfway done
bdf0cacb Got Nemesis Working with Hack + Breakpoint in init
00fb93be Got Fabtests Working while returning buffers
33be105b reverting fabtest changes
```

a-szegel:
The good news is that the pt2pt only implementation passes pt2pt fabtests with data integrity checks, and the performance of it doesn't look so bad.  I am expecting a few changes to come in to make the performance worse (like properly returning FQE's, Memory Barriers for ARM, and potentially a hashmap lookup on the critical path)... but right now its pretty.

c5n osu_latency SM2 (this patch)
```
/home/ec2-user/Developer/ompi/install/bin/mpirun --wdir . -n 2 --hostfile /home/ec2-user/Developer/ompi/install/hostfile --map-by ppr:2:node --bind-to core  --mca mtl_ofi_provider_include "sm2" --mca mtl ofi  -x PATH /home/ec2-user/Developer/ompi/install/omb/libexec/osu-micro-benchmarks/mpi/pt2pt/osu_latency
# OSU MPI Latency Test v5.8
# Size          Latency (us)
0                       0.23
1                       0.26
2                       0.26
4                       0.26
8                       0.26
16                      0.30
32                      0.30
64                      0.32
128                     0.38
256                     0.42
512                     0.60
1024                    0.68
2048                    0.84
4096                    1.61
8192                    1.77
16384                   2.31
32768                   3.13
65536                   4.91
131072                  8.35
262144                 14.86
524288                 31.89
1048576                78.52
2097152               189.22
4194304               441.06
```

c5n OMPI (with fastboxes) osu_latency
```
/home/ec2-user/Developer/ompi/install/bin/mpirun --wdir . -n 2 --hostfile /home/ec2-user/Developer/ompi/install/hostfile --map-by ppr:2:node --bind-to core --mca mtl ^ofi --mca btl ^ofi  -x PATH /home/ec2-user/Developer/ompi/install/omb/libexec/osu-micro-benchmarks/mpi/pt2pt/osu_latency
# OSU MPI Latency Test v5.8
# Size          Latency (us)
0                       0.28
1                       0.30
2                       0.30
4                       0.29
8                       0.29
16                      0.36
32                      0.33
64                      0.35
128                     0.44
256                     0.44
512                     0.58
1024                    0.61
2048                    0.80
4096                    1.55
8192                    1.74
16384                   2.35
32768                   3.16
65536                   5.05
131072                  8.34
262144                 15.25
524288                 32.29
1048576                77.15
2097152               189.58
4194304               440.55
```

SHM (most recent main branch including Amir's updates):
```
/home/ec2-user/Developer/ompi/install/bin/mpirun --wdir . -n 2 --hostfile /home/ec2-user/Developer/ompi/install/hostfile --map-by ppr:2:node --bind-to core  --mca mtl_ofi_provider_include "shm" --mca mtl ofi  -x PATH /home/ec2-user/Developer/ompi/install/omb/libexec/osu-micro-benchmarks/mpi/pt2pt/osu_latency
# OSU MPI Latency Test v5.8
# Size          Latency (us)
0                       0.34
1                       0.36
2                       0.36
4                       0.36
8                       0.36
16                      0.40
32                      0.40
64                      0.41
128                     0.47
256                     0.51
512                     0.67
1024                    0.74
2048                    0.86
4096                    1.69
8192                    1.89
16384                   2.40
32768                   3.25
65536                   4.99
131072                  8.42
262144                 14.83
524288                 32.15
1048576                76.12
2097152               190.33
4194304               435.64
```

I know we don't really care about pt2pt... but it is nice to know that we are on track.

aingerson:
@a-szegel Thanks for the data. Those OMPI numbers look significantly worse than the numbers provided on the proposal. Is there a significant difference between the c5n system and the hpc6a systems from earlier runs that explains that difference? Or is that a run to run difference? Trying to understand the baseline we should be aiming for.

a-szegel:
Yes big difference, hpc6a is expected to be a bit faster:

hpc6a (2x48 cores @  3.6GHz, AMD Milan,  386GiB memory) 
c5n (2x18 cores @ 3.0 GHz Intel Xeon Platinum processors (Skylake 8124) 

I am happy that we are on the same level as OMPI.

PR #8583, opened 2023-03-01
prov/efa: Fix a bug in rxr_pkt_handle_rma_read_completion

shijin-aws:
rxr_pkt_handle_rma_read_completion call rxr_tracepoint with read_entry->context as an argument, while read_entry is a null ptr at this time. This patch fixes this issue by changing read_entry->context to rx_entry, which is what it was referring to.


wenduwan (approved):
Thank you!

On a separate note, I think it also helps to change our code style in this case:
Instead of declaring `struct rxr_op_entry *rx_entry;` on L395, we can move it to the `else` branch on L444. Therefore code on the other branch won't have access to `rx_entry`, and we get a compiler error.

lrbison (approved):

wzamazon (approved):

PR #8585, opened 2023-03-01
prov/efa: handle the IBV_WC_RDMA_READ opcode.

wzamazon:
Currently, rdma-core efa provider return IBV_WC_SEND as opcode of a read operation. This issue will be addressed in the future version of rdma-core.

This patch handle the IBV_WC_RDMA_READ opcode to prepare for the future rdma-core change.

### DIFF
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -1715,6 +1715,7 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			ep->recv_comps++;
 #endif
 			break;
+		case IBV_WC_RDMA_READ:
 		case IBV_WC_RDMA_WRITE:
 			pkt_entry = (void *)(uintptr_t)ep->ibv_cq_ex->wr_id;
 			rxr_pkt_handle_rma_completion(ep, pkt_entry);


wenduwan (approved):

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1715,6 +1715,7 @@ static inline void rdm_ep_poll_ibv_cq_ex(struct rxr_ep *ep, size_t cqe_to_proces
 			ep->recv_comps++;
 #endif
 			break;
+		case IBV_WC_RDMA_READ:
```
shijin-aws, line 1718: not related to this PR, when will we remove the assert in L1688?

shijin-aws (approved):

prov/efa/src/rdm/rxr_ep.c
wzamazon: I do not think we will remove the assert.

wzamazon: assert(0) is something that should never happen. It helps debugging.

j-xiong: I think the asset at L1688 is no longer valid. It would cause assertion failure (instead of reporting error) if a RDMA op fails.

wzamazon: Sorry. I misread the code. I was talking about L1730.

L1688 is definitely wrong

wzamazon: That will be handled in a separate PR

PR #8588, opened 2023-03-01
prov/efa: Add rdma_core support for RDMA write

lrbison:
During configure, check efadv.h to see if rdma-core is new enough to include the RDMA write capability flag.

If the flag is available, use it to check the device for RDMA write capability during runtime.

### DIFF
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -75,14 +75,14 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	AS_IF([test x"$enable_efa" != x"no"],
 	      [AC_CHECK_MEMBER(struct efadv_device_attr.max_rdma_size,
 			      [AC_DEFINE([HAVE_RDMA_SIZE], [1], [efadv_device_attr has max_rdma_size])],
-			      [],
+			      [AC_DEFINE([HAVE_RDMA_SIZE], [0], [efadv_device_attr does not have max_rdma_size])],
 			      [[#include <infiniband/efadv.h>]])
 	      ])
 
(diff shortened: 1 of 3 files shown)


prov/efa/configure.m4
```diff
@@ -87,6 +87,14 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	      ])
 	CPPFLAGS=$save_CPPFLAGS
 
+	AS_IF([test x"$enable_efa" != x"no"],
+	      [AC_CHECK_DECL(EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE,
+			    [AC_DEFINE([HAVE_CAPS_RDMA_WRITE], [1], [HAVE_CAPS_RDMA_WRITE is defined])],
```
wzamazon, line 92: The common practice is to always define the Macro, but set the value to 0 if the flag is not available.

Then use `if` instead of `ifdef` in the code

lrbison: This is not the case for the 2 preceding defines: `HAVE_CAPS_RNR_RETRY` and `HAVE_RDMA_SIZE`

Should I favor (local) consistency or `if`?

wzamazon: > Should I favor (local) consistency or if?

This practise came from a bug that a macro defined as 0 was treated as a yes.

lrbison: I see.  That's a good reason.

I changed my commit, and added a new one to switch the other to defines to be used as you say.

wzamazon (approved):

PR #8592, opened 2023-03-01
prov/hook/dmabuf_peer_mem: Handle IPC handle caching in L0

j-xiong:
In older oneAPI L0 library IPC handles are not cached so that the returned dmabuf fds need to be explicitly closed to avoid running out of file descriptors.

In newer oneAPI L0 libraries IPC handles are cached so that it becomes not only unnecessary but also errorous to explicitly close the fds.

Detect the IPC handle caching status at runtime so that the dmabuf fds can be correctly handled.

### DIFF
--- a/prov/hook/dmabuf_peer_mem/src/hook_dmabuf_peer_mem.c
+++ b/prov/hook/dmabuf_peer_mem/src/hook_dmabuf_peer_mem.c
@@ -99,6 +99,61 @@ static int dmabuf_reg_query(int reg_fd, uint64_t addr, uint64_t size, int *fd)
 }
 
 /*
+ * IPC handles are not cached in old oneAPI L0. Each time a call is made to
+ * get the IPC handle, a new dmabuf fd is created. In this case, the fd needs
+ * to be explicitly closed when no longer in use to avoid running out of file
+ * descriptors.
+ *
+ * In newer L0 library, IPC handles are cached. The same dmabuf fd is returned
+ * for multiple calls to get IPC handles as long as the buffer is the same. As
+ * a result, the fd SHOULD NOT be closed explicitly otherwise later use of the
+ * fd will fail.
+ *
+ * By default assume the IPC handle is cached. Not only this is more up-to-date,
+ * but also the side effect is less severe if the assumption turns out wrong.
+ */
+static bool ze_ipc_handle_is_cached = true;
+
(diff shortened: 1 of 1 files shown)


prov/hook/dmabuf_peer_mem/src/hook_dmabuf_peer_mem.c
```diff
@@ -98,6 +98,61 @@ static int dmabuf_reg_query(int reg_fd, uint64_t addr, uint64_t size, int *fd)
...
+
+	err = ze_hmem_get_handle(buf, len, &handle);
+	if (err)
+		goto end;
+
+	fd2 = (int)(uintptr_t)handle;
+	ze_ipc_handle_is_cached = (fd == fd2);
+
+	if (!ze_ipc_handle_is_cached)
+		close(fd2);
```
shefty, line 142: Can the above check be done once and the result saved, so that we avoid calling get_handle() twice every time?

j-xiong: It's done only once. Note how the static flag `first` is set and checked.

shefty: oh, yeah, so it is.  :)

prov/hook/dmabuf_peer_mem/src/hook_dmabuf_peer_mem.c
```diff
@@ -98,6 +98,61 @@ static int dmabuf_reg_query(int reg_fd, uint64_t addr, uint64_t size, int *fd)
...
+ *
+ * In newer L0 library, IPC handles are cached. The same dmabuf fd is returned
+ * for multiple calls to get IPC handles as long as the buffer is the same. As
+ * a result, the fd SHOULD NOT be closed explicitly otherwise later use of the
+ * fd will fail.
+ *
+ * By default assume the IPC handle is cached. Not only this is more up-to-date,
+ * but also the side effect is less severe if the assumption turns out wrong.
+ */
+static int ze_ipc_handle_is_cached = 1;
```
shefty, line 115: nit: this should be a bool

j-xiong: fixed.

PR #8605, opened 2023-03-02
prov/efa: Fix the error handling of gdrcopy_dev_register.

shijin-aws:
cuda_gdrcopy_dev_register could fail due to some hardware limitation, e.g. memory limit. In this case, if user enable cuda API, efa should fallback to cudaMemcpy.
Per this, efa provider should only unregister cuda gdr handle when it's indeed registered, i.e. the device.cuda is a gdr handle rather than a device id

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -246,6 +246,7 @@ static int efa_mr_hmem_setup(struct efa_mr *efa_mr,
 
 	/* efa_mr->peer.device is an union. Setting reserved to 0 cleared everything in it (cuda, neuron, synapseai etc) */
 	efa_mr->peer.device.reserved = 0;
+	efa_mr->peer.use_gdrcopy = false;
 	if (efa_mr->peer.iface == FI_HMEM_CUDA) {
 		if (rxr_env.set_cuda_sync_memops) {
 			err = cuda_set_sync_memops(attr->mr_iov->iov_base);
@@ -261,7 +262,12 @@ static int efa_mr_hmem_setup(struct efa_mr *efa_mr,
 				EFA_WARN(FI_LOG_MR,
 					 "Unable to register handle for GPU memory. err: %d buf: %p len: %zu\n",
 					 err, attr->mr_iov->iov_base, attr->mr_iov->iov_len);
-				return err;
+				/* When gdrcopy pin buf failed, fallback to cudaMemcpy when user enables cuda xfer */
+				if (efa_mr->domain->cuda_xfer_setting != CUDA_XFER_ENABLED)
+					return err;
+				efa_mr->peer.device.cuda = attr->device.cuda;
+			} else {
+				efa_mr->peer.use_gdrcopy = true;
 			}
 		} else {
 			efa_mr->peer.device.cuda = attr->device.cuda;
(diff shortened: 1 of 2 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -261,7 +261,12 @@ static int efa_mr_hmem_setup(struct efa_mr *efa_mr,
 				EFA_WARN(FI_LOG_MR,
 					 "Unable to register handle for GPU memory. err: %d buf: %p len: %zu\n",
 					 err, attr->mr_iov->iov_base, attr->mr_iov->iov_len);
-				return err;
+				/* When gdrcopy pin buf failed, fallback to cudaMemcpy when user enables cuda xfer */
+				if (efa_mr->domain->cuda_xfer_setting != CUDA_XFER_ENABLED)
+					return err;
+				efa_mr->peer.device.cuda = attr->device.cuda;
```
wenduwan, line 267: Missing `efa_mr->peer.use_gdrcopy = false` - could cause uninitialized variable issue.

prov/efa/src/efa_mr.h
```diff
@@ -49,6 +49,7 @@ struct efa_mr_peer {
 		int             neuron;
 		int             synapseai;
 	} device;
+	bool use_gdrcopy;
```
wenduwan, line 52: I have a hunch that we do not need this flag - simply set `reserved = 0` should be enough. WDYT?

prov/efa/src/efa_mr.c
shijin-aws: I thought about that. I could add it explicitly, but I assume this should be handled by the very beginning like `efa_mr = {0}`

prov/efa/src/efa_mr.h
shijin-aws: That will disregard user device, though it does not actually matter.

wenduwan (approved):

prov/efa/src/efa_mr.h
shijin-aws: Use `ofi_cudaIsDeviceId` to check whether it's a device id or handle in the new revision.

prov/efa/src/efa_mr.c
shijin-aws: This is unnecessary in new revision as we will use `device.reserved` which is reset to 0 already.

include/ofi_hmem.h
```diff
@@ -227,6 +227,8 @@ bool synapseai_is_addr_valid(const void *addr, uint64_t *device,
 int synapseai_host_register(void *ptr, size_t size);
 int synapseai_host_unregister(void *ptr);
 
+bool ofi_cudaIsDeviceId(uint64_t device);
```
shefty, line 230: Exposed libfabric functions use all lower-case.  Commit message should indicate why it's being exposed.

prov/efa/src/efa_mr.c
```diff
@@ -498,7 +501,7 @@ static int efa_mr_dereg_impl(struct efa_mr *efa_mr)
 		efa_mr->shm_mr = NULL;
 	}
 
-	if (efa_mr->peer.iface == FI_HMEM_CUDA && cuda_is_gdrcopy_enabled()) {
+	if (efa_mr->peer.iface == FI_HMEM_CUDA && !ofi_cudaIsDeviceId(efa_mr->peer.device.cuda)) {
```
shefty, line 504: This sort of check looks like an incorrect use of the APIs.  Why would the device value be invalid?  If it isn't being set, how do you know what's there isn't accidentally valid?

shijin-aws: EFA will set `device.cuda` as gdr handle when gdrcopy is available. We used to use this to check whether `device` is a handle or a device id. But your point is totally right that this logic is problematic. We could introduce explicit check whether gdrcopy is being used so this `ofi_cudaIsDeviceId` call can be avoided.

prov/efa/src/efa_mr.c
shijin-aws: The latest revision set efa_mr->peer.use_gdrcopy = false at the beginning.

include/ofi_hmem.h
shijin-aws: Dropped this change in latest revision

wenduwan (approved):
Thank you!

shefty (approved):

PR #8612, opened 2023-03-03
prov/efa: enrich error messages with debug and connection info

wenduwan:
Add additional debug info for common errors.
Add unit test cases to cover error string format.


prov/efa/src/efa_strerror.c
```diff
@@ -101,6 +102,22 @@ static const char *efa_errno_str(enum efa_errno err)
 		: errno_str[err - FI_EFA_ERRNO_OFFSET];
 }
 
+static const char *efa_debug_str(enum efa_errno err)
+{
+	switch (err) {
+	case FI_EFA_REMOTE_ERROR_BAD_DEST_QPN:
+		return "\nThis error can be caused by a crashed peer. "
+			   "Please verify the receiver status";
```
lrbison, line 110: At this point do we have any information as to which peer we are trying to address?  I think that would be very helpful to the user.

prov/efa/src/efa_strerror.c
```diff
@@ -101,6 +102,22 @@ static const char *efa_errno_str(enum efa_errno err)
 		: errno_str[err - FI_EFA_ERRNO_OFFSET];
 }
 
+static const char *efa_debug_str(enum efa_errno err)
+{
+	switch (err) {
```
lrbison, line 107: I forget, would it be appropriate to handle RNR here?

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -156,38 +158,83 @@ void test_cq_read_bad_send_status(struct efa_resource *resource, enum fi_ep_type
...
- * 
+ *
  * @param[in]	state		struct efa_resource that is managed by the framework
  */
-void test_ibv_cq_ex_read_bad_send_status(struct efa_resource **state)
+void test_ibv_cq_ex_read_bad_send_status_unresponsive_receiver(struct efa_resource **state)
 {
 	struct efa_resource *resource = *state;
-	test_cq_read_bad_send_status(resource, FI_EP_RDM);
+	test_cq_read_bad_send_status(resource, FI_EP_RDM, FI_EFA_LOCAL_ERROR_UNRESP_REMOTE,
```
lrbison, line 179: Tests which strcmp error messages seem a little brittle, but I guess there aren't a whole lot of alternatives to a test for errors.

prov/efa/src/efa_strerror.c
wenduwan: That's the plan. We want to log the peer instance id (or some identifier) to help debug. That's the next step.

prov/efa/src/efa_strerror.c
wenduwan: Sorry I must be missing the background. What is the context here, i.e. handle RNR in what sense?

prov/efa/test/efa_unit_test_cq.c
wenduwan: Yeah this test is to effectively cover the switch branches.

prov/efa/src/efa_strerror.c
wenduwan: Also for that I need to evaluate what the best place is - this function as-is is not capable to accept additional metadata from the caller, i.e. static/stateless.
We might do it at the place of erroring.

prov/efa/src/efa_strerror.c
```diff
@@ -101,6 +102,22 @@ static const char *efa_errno_str(enum efa_errno err)
...
+	case FI_EFA_REMOTE_ERROR_BAD_DEST_QPN:
+		return "\nThis error can be caused by a crashed peer. "
+			   "Please verify the receiver status";
+	case FI_EFA_LOCAL_ERROR_UNRESP_REMOTE:
```
sunkuamzn, line 111: Does `FI_EFA_LOCAL_ERROR_UNRESP_REMOTE` necessarily mean problems with security group rules? Could it not be other network failures?

wenduwan: Misconfigured ingress/egress rule is a common/basic cause. There could be other reasons - the hint states the obvious to remind the user to avoid those traps.

sunkuamzn: In that case, the error message should specify that problems with SG rules is a common but just one of the possible root causes. As it is, the message is misleading.

wenduwan: Thank you will address!

Do you have a handy rephrase in mind?

prov/efa/src/efa_strerror.c
```diff
@@ -101,6 +102,21 @@ static const char *efa_errno_str(enum efa_errno err)
...
+		return "\nThis error can be caused by incrorrect inbound/outbound "
+			   "rules in the security group. EFA requires \"All traffic\" "
+			   "type allowlisting within the security group";
```
darrylabbate, line 114: Nit: omit preceding newlines from the string literals here. Allow the callers to determine whether or not the extra newline is necessary in their resulting formatted strings.
```suggestion
	case FI_EFA_REMOTE_ERROR_BAD_DEST_QPN:
		return "This error can be caused by a crashed peer. "
			   "Please verify the receiver status";
	case FI_EFA_LOCAL_ERROR_UNRESP_REMOTE:
		return "This error can be caused by incrorrect inbound/outbound "
			   "rules in the security group. EFA requires \"All traffic\" "
			   "type allowlisting within the security group";
```

prov/efa/src/efa_strerror.c
```diff
@@ -112,7 +128,16 @@ static const char *efa_errno_str(enum efa_errno err)
  */
 const char *efa_strerror(enum efa_errno err)
 {
-	return err >= FI_EFA_ERRNO_OFFSET
-		? efa_errno_str(err)
-		: efa_io_comp_status_str(err);
+	const char *base_err = err >= FI_EFA_ERRNO_OFFSET
+							   ? efa_errno_str(err)
+							   : efa_io_comp_status_str(err);
```
darrylabbate, line 133: Mega indent
```suggestion
	const char *base_err = err >= FI_EFA_ERRNO_OFFSET
		? efa_errno_str(err)
		: efa_io_comp_status_str(err);
```

prov/efa/src/efa_strerror.c
```diff
@@ -112,7 +128,16 @@ static const char *efa_errno_str(enum efa_errno err)
...
+	const char *debug_hint = efa_debug_str(err);
+
+	if (debug_hint) {
+		char *err_str = malloc(strlen(base_err) + strlen(debug_hint) + 1);
+		strcpy(stpcpy(err_str, base_err), debug_hint);
```
darrylabbate, line 138: Avoiding the non-standard `stpcpy()` is simple enough:
```suggestion
		strcpy(err_str, base_err);
		strcat(err_str, debug_str);
```

There's also a million other more principled ways to do this, but since this is an error path we probably shouldn't care too much.

prov/efa/src/efa_strerror.c
```diff
@@ -112,7 +128,16 @@ static const char *efa_errno_str(enum efa_errno err)
...
+		return err_str;
```
darrylabbate, line 139: Might want to cast the result to match the type signature
```suggestion
		return (const char *) err_str;
```

prov/efa/src/efa_strerror.c
sunkuamzn: Your new version looks good to me!

prov/efa/src/efa_strerror.c
```diff
@@ -101,6 +102,21 @@ static const char *efa_errno_str(enum efa_errno err)
...
+static const char *efa_debug_str(enum efa_errno err)
+{
+	switch (err) {
+	case FI_EFA_REMOTE_ERROR_BAD_DEST_QPN:
+		return "This error can be caused by a crashed peer. "
+			   "Please verify the receiver status";
+	case FI_EFA_LOCAL_ERROR_UNRESP_REMOTE:
+		return "This error can be caused by incrorrect inbound/outbound "
+			   "rules in the security group. EFA requires \"All traffic\" "
+			   "type allowlisting within the security group";
```
wzamazon, line 114: This error can also be caused by crashed receiver

prov/efa/src/efa_strerror.c
```diff
@@ -101,6 +102,21 @@ static const char *efa_errno_str(enum efa_errno err)
 		: errno_str[err - FI_EFA_ERRNO_OFFSET];
 }
 
+static const char *efa_debug_str(enum efa_errno err)
+{
+	switch (err) {
+	case FI_EFA_REMOTE_ERROR_BAD_DEST_QPN:
+		return "This error can be caused by a crashed peer. "
+			   "Please verify the receiver status";
```
wzamazon, line 110: Can we print peer address here?

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -156,38 +158,81 @@ void test_cq_read_bad_send_status(struct efa_resource *resource, enum fi_ep_type
 	strerror = fi_cq_strerror(resource->cq, cq_err_entry.prov_errno, NULL, &err_buf, 0);
 	assert_int_equal(ret, 1);
 	assert_int_equal(cq_err_entry.err, FI_EIO);
-	assert_int_equal(cq_err_entry.prov_errno, FI_EFA_LOCAL_ERROR_UNRESP_REMOTE);
```
wzamazon, line 159: I would keep this assert

prov/efa/src/efa_strerror.c
darrylabbate: I think this function/file should be limited to returning `const` string fragments which callers can stitch together to build useful error messages. The peer address is more of a dynamic/runtime thing and should be a separate utility.

What we should do is defer all error reporting to central utilities which take the current state as input and log error messages (kind of like our `_write_error()` functions). That would give us more flexibility and allow us to easily configure different combinations of various information, rather than shoehorning as much information as possible in the error strings themselves.

wenduwan:
I will update the PR after this change is finalized https://github.com/ofiwg/libfabric/pull/8657

prov/efa/src/efa_strerror.c
```diff
@@ -63,6 +69,24 @@ static const char *efa_io_comp_status_str(enum efa_errno status)
...
+static inline char *efa_host_id_str(uint64_t host_id)
+{
+	if (!host_id) {
+		return NULL;
+	}
+
+	char *host_id_str = calloc(HOST_ID_LENGTH + 1, sizeof(host_id_str));
+	if (!host_id_str) {
+		return NULL;
+	}
```
lrbison, line 81: If caller should free this memory, it should be clearly stated in the doxygen block for this function.

I would prefer if this function wrote to a global variable (like fi_tostr) or to a caller-provided string, rather than allocate new memory.

prov/efa/src/efa_strerror.c
```diff
@@ -101,18 +125,100 @@ static const char *efa_errno_str(enum efa_errno err)
...
 
+static const char *efa_debug_str(enum efa_errno err)
+{
+	switch (err) {
+	case FI_EFA_REMOTE_ERROR_BAD_DEST_QPN:
+		return "This error can be caused by a crashed peer. "
+			   "Please verify the receiver status";
+	case FI_EFA_LOCAL_ERROR_UNRESP_REMOTE:
+		return "This error can be caused by a crashed peer or "
+			   "incrorrect inbound/outbound rules in the security group - "
```
lrbison, line 136: *incorrect

prov/efa/src/efa_strerror.c
```diff
@@ -101,18 +125,100 @@ static const char *efa_errno_str(enum efa_errno err)
...
+			   "EFA requires \"All traffic\" type allowlisting. "
+			   "Please also verify the receiver status.";
```
lrbison, line 138: maybe a little more explicit than "peer status":
"Please also verify the peer is still healthy and has not terminated unexpectedly"

prov/efa/src/efa_strerror.c
```diff
@@ -101,18 +125,100 @@ static const char *efa_errno_str(enum efa_errno err)
...
+		return NULL;
+	}
+
+	static const char NA_STR[] = "N/A";
+	static const char format[] = "Local: %s LHost id: %s\tPeer: %s RHost id: %s";
+
+	size_t max_length = sizeof(format) - 4 +
+						MAX(HOST_ID_LENGTH + 1, sizeof(NA_STR)) * 2 +
+						MAX(sizeof(err_data->ep_raw_addr_str), sizeof(NA_STR)) +
+						MAX(sizeof(err_data->peer_raw_addr_str), sizeof(NA_STR));
```
lrbison, line 156: I think these can just be OFI_ADDRSTRLEN, since we know NA_STR < OFI_ADDRSTRLEN, and sizeof(...) is just going to return OFI_ADDRSTRLEN.

prov/efa/src/efa_strerror.c
```diff
@@ -101,18 +125,100 @@ static const char *efa_errno_str(enum efa_errno err)
...
+		return "This error can be caused by a crashed peer or "
+			   "incrorrect inbound/outbound rules in the security group - "
+			   "EFA requires \"All traffic\" type allowlisting. "
+			   "Please also verify the receiver status.";
+	default:
+		return NULL;
+	}
+}
+
+const char *efa_err_conn_str(const struct efa_error_data *err_data)
```
lrbison, line 144: Following Darryl's reasoning, the memory management required by this function would be simpler if this function simply logged all this information rather than returning buffers that the caller now needs to free.

Also: missing doxygen

wzamazon (commented):
I think there are too many allocations. Can you allocate 1 string and use it as `err_data`?

prov/efa/src/efa_strerror.c
```diff
@@ -101,18 +128,92 @@ static const char *efa_errno_str(enum efa_errno err)
...
+ * @param[in]	err_data	efa_error_data pointer that contains the connection information
+ * @param[out]	err_str		The destination string
+ * @return		A status code. 0 if successful, otherwise a negative integer.
+*/
+int write_efa_err_conn_str(const struct efa_error_data *err_data, char *err_str)
+{
+	if (!err_data || !err_str)
+		return -FI_EINVAL;
+
+	int ret = snprintf(err_str, CONN_INFO_MAX_LENGTH, CONN_INFO_FORMAT,
```
wzamazon, line 159: if you use `asprintf`, then you do not need to calculate length.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -615,18 +613,30 @@ void rxr_rx_entry_handle_error(struct rxr_op_entry *rx_entry, int err, int prov_
...
+	err_entry.err_data = (void *)&err_data;
+	err_entry.err_data_size = sizeof(err_data);
 
 	rxr_msg_multi_recv_free_posted_entry(ep, rx_entry);
 
-	buflen = sizeof(ep_addr_str);
-	rxr_ep_raw_addr_str(ep, ep_addr_str, &buflen);
-	buflen = sizeof(peer_addr_str);
-	rxr_ep_get_peer_raw_addr_str(ep, rx_entry->addr, peer_addr_str, &buflen);
+	err_data = calloc(1, sizeof(struct efa_error_data));
```
wzamazon, line 621: Why not just construct the string here and use it as `err_data`?

prov/efa/src/efa_strerror.c
```diff
@@ -101,18 +128,92 @@ static const char *efa_errno_str(enum efa_errno err)
...
+							   : efa_io_comp_status_str(err);
+	const char *debug_hint = efa_debug_str(err);
+	size_t err_str_len = strlen(base_err);
+
+	/* Separate string with a newline */
+	err_str_len += debug_hint ? strlen(debug_hint) + 1 : 0;
+	err_str_len += err_data ? CONN_INFO_MAX_LENGTH + 1 : 0;
+	err_str_len += 1; /* \0 */
+
+	char *err_str = malloc(err_str_len);
```
wzamazon, line 199: Each call to this function will malloc a piece of memory, which is totally unnecessary.

wzamazon (changes requested):
Overall, I found this patch to be unnecessarily complicated. I was envisioning something much simpler.

e.g. 1 new function that will construct the `cq_entry.err_data` as a string. something like

```
int efa_err_data_alloc(rxr_ep *ep, fi_addr_t *addr, int err, int prov_errno, void ** err_data, size_t *err_data_len)
{
      char *err_msg;

      err_msg_len  = asprintf(&err_msg"%s My EFA address is %s My host id is i-%017lx. Peer EFA address is %s. Peer host id is i-%-017x\n", efa_strerror(prov_errno), rxr_ep_raw_addr_str(xxx), rxr_ep_peer_raw_addr_str(xxx), ep->host_id, peer->hostid);

    strstr(err_msg, "i-0000000000", "not available");
    *err_data = err_msg;
    *err_data_len = err_msg_len;
}
````


I do not see an advantage to introduce `struct efa_error_data`.

**Comment by wenduwan** [comment]
> Overall, I found this patch to be unnecessarily complicated. I was envisioning something much simpler.
> 
> e.g. 1 new function that will construct the `cq_entry.err_data` as a string. something like
> 
> ```
> int efa_err_data_alloc(rxr_ep *ep, fi_addr_t *addr, int err, int prov_errno, void ** err_data, size_t *err_data_len)
> {
>       char *err_msg;
> 
>       err_msg_len  = asprintf(&err_msg"%s My EFA address is %s My host id is i-%017lx. Peer EFA address is %s. Peer host id is i-%-017x\n", efa_strerror(prov_errno), rxr_ep_raw_addr_str(xxx), rxr_ep_peer_raw_addr_str(xxx), ep->host_id, peer->hostid);
> 
>     strstr(err_msg, "i-0000000000", "not available");
>     *err_data = err_msg;
>     *err_data_len = err_msg_len;
> }
> ```
> 
> I do not see an advantage to introduce `struct efa_error_data`.

Thanks! I think this makes sense. Will incorporate in the update.

File: prov/efa/src/rdm/efa_rdm_error.h [hunk 20]
```diff
@@ -0,0 +1,105 @@
...
+    struct efa_rdm_peer *peer = rxr_ep_get_peer(ep, addr);
+
+    len = sizeof(ep_addr_str);
+    rxr_ep_raw_addr_str(ep, ep_addr_str, &len);
+    len = sizeof(peer_addr_str);
+    rxr_ep_get_peer_raw_addr_str(ep, addr, peer_addr_str, &len);
+
+    if (prov_errno == FI_EFA_REMOTE_ERROR_RNR) {
+        /* RNR error is retried. No need to write error data */
+        return -FI_EAGAIN;
```
lrbison, line 75: Seems like FI_EAGAIN should be returned from send, but to me this return code indicates user should call efa_rdm_error_write_msg again.  Maybe here we want to set ep->err_msg[0] = '\0'; and return success?

prov/efa/src/rdm/efa_rdm_error.h
```diff
@@ -0,0 +1,105 @@
...
+
+    int ret = snprintf(ep->err_msg, len, ENRICHED_ERR_FORMAT, base_msg, ep_addr_str, local_host_id_str, peer_addr_str, peer_host_id_str);
+
+    if (ret < 0 || ret >= len) {
+        return -FI_EINVAL;
+    }
+
+    if (strlen(ep->err_msg) >= RXR_ERROR_MSG_BUFFER_LENGTH) {
+        return -FI_ENOBUFS;
+    }
```
lrbison, line 98: Should we set buf to ep->err_msg or maybe to a static "Problem creating enriched error string" or something when returning early?

looking at the strlen(ep->err_msg) check made me realize if that's the relevant buffer size, it should probably be the argument to snprintf on line 90.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -615,18 +612,15 @@ void rxr_rx_entry_handle_error(struct rxr_op_entry *rx_entry, int err, int prov_
 	err_entry.buf = rx_entry->cq_entry.buf;
 	err_entry.data = rx_entry->cq_entry.data;
 	err_entry.tag = rx_entry->cq_entry.tag;
+	if (OFI_UNLIKELY(efa_rdm_error_write_msg(ep, rx_entry->addr, err, prov_errno,
+											 &err_entry.err_data, &err_entry.err_data_size) < 0)) {
+		err_entry.err_data_size = 0;
+	}
```
lrbison, line 618: nit: indent

functional: I think efa_rdm_error_write_msg may leave err_entry.err_data un-initialized on error.  I can't tell from github if err_entry.err_data is set to NULL ahead of time.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -716,18 +704,14 @@ void rxr_tx_entry_handle_error(struct rxr_op_entry *tx_entry, int err, int prov_
 	err_entry.buf = tx_entry->cq_entry.buf;
 	err_entry.data = tx_entry->cq_entry.data;
 	err_entry.tag = tx_entry->cq_entry.tag;
-	if (FI_VERSION_GE(api_version, FI_VERSION(1, 5)))
+	if (OFI_UNLIKELY(efa_rdm_error_write_msg(ep, tx_entry->addr, err, prov_errno,
+											 &err_entry.err_data, &err_entry.err_data_size) < 0)) {
+		EFA_WARN(FI_LOG_DEBUG, "efa_rdm_error_write_msg failed\n");
```
lrbison, line 709: same comment as above: indent and err_data

lrbison (commented):
This looks pretty good, only minor feedback

prov/efa/src/efa_strerror.c
```diff
@@ -51,11 +51,19 @@ static const char *efa_io_comp_status_str(enum efa_errno status)
 		[FI_EFA_LOCAL_ERROR_BAD_LENGTH]        = "Message too long",
 		[FI_EFA_REMOTE_ERROR_BAD_ADDRESS]      = "Invalid address",
 		[FI_EFA_REMOTE_ERROR_ABORT]            = "Receiver connection aborted",
-		[FI_EFA_REMOTE_ERROR_BAD_DEST_QPN]     = "Invalid receiver queue pair number (QPN)",
+		[FI_EFA_REMOTE_ERROR_BAD_DEST_QPN]     = "Invalid receiver queue pair number (QPN). "
+		                                         "This error can be caused by a crashed peer. "
```
wzamazon, line 55: I think we need stronger word. Instead  of "can be", I think we use "is typically"

prov/efa/src/efa_strerror.c
```diff
@@ -51,11 +51,19 @@ static const char *efa_io_comp_status_str(enum efa_errno status)
...
+		                                         "Please verify the peer application has not "
+		                                         "terminated unexpectedly.",
 		[FI_EFA_REMOTE_ERROR_RNR]              = "Receiver not ready",
 		[FI_EFA_REMOTE_ERROR_BAD_LENGTH]       = "Receiver scatter-gather list (SGL) too short",
 		[FI_EFA_REMOTE_ERROR_BAD_STATUS]       = "Unexpected status received from remote",
-		[FI_EFA_LOCAL_ERROR_UNRESP_REMOTE]     = "Unresponsive receiver",
+		[FI_EFA_LOCAL_ERROR_UNRESP_REMOTE]     = "Unresponsive receiver. "
+		                                         "This error can be caused by a peer hardware failure or "
```
wzamazon, line 62: "can be" -> "is typically caused either peer hardware failure or incorrect inbound/outbound rule"

prov/efa/src/rdm/efa_rdm_error.h
```diff
@@ -0,0 +1,97 @@
...
+
+    if (!ep->host_id || HOST_ID_STR_LENGTH != snprintf(local_host_id_str, HOST_ID_STR_LENGTH + 1, "i-%017lx", ep->host_id)) {
+        strcpy(local_host_id_str, "N/A");
+    }
+
+    if (!ep->host_id || HOST_ID_STR_LENGTH != snprintf(peer_host_id_str, HOST_ID_STR_LENGTH + 1, "i-%017lx", peer->host_id)) {
+        strcpy(peer_host_id_str, "N/A");
+    }
+
+    int ret = snprintf(ep->err_msg, RXR_ERROR_MSG_BUFFER_LENGTH + 1, "%s My EFA addr: %s My host id: %s Peer EFA addr: %s Peer host id: %s",
```
wzamazon, line 80: No `+1` here.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -615,18 +612,15 @@ void rxr_rx_entry_handle_error(struct rxr_op_entry *rx_entry, int err, int prov_
 	err_entry.buf = rx_entry->cq_entry.buf;
 	err_entry.data = rx_entry->cq_entry.data;
 	err_entry.tag = rx_entry->cq_entry.tag;
+	if (OFI_UNLIKELY(efa_rdm_error_write_msg(ep, rx_entry->addr, err, prov_errno,
+	                                         &err_entry.err_data, &err_entry.err_data_size) < 0)) {
```
wzamazon, line 616: no need `< 0`.

prov/efa/src/rdm/rxr_op_entry.c
```diff
@@ -716,18 +704,13 @@ void rxr_tx_entry_handle_error(struct rxr_op_entry *tx_entry, int err, int prov_
...
-	rxr_ep_raw_addr_str(ep, ep_addr_str, &buflen);
-	buflen = sizeof(peer_addr_str);
-	rxr_ep_get_peer_raw_addr_str(ep, tx_entry->addr, peer_addr_str, &buflen);
-
-	EFA_WARN(FI_LOG_CQ,
-		"err: %d, prov_err: %s (%d) our address: %s, peer address %s\n",
-		err_entry.err, efa_strerror(err_entry.prov_errno),
-		err_entry.prov_errno, ep_addr_str, peer_addr_str);
+	EFA_WARN(FI_LOG_CQ, "err: %d, message: %s (%d)\n", err_entry.err,
+			 efa_strerror(err_entry.prov_errno, err_entry.err_data), err_entry.prov_errno);
```
wzamazon, line 713: indent?

wzamazon (commented):
A few small comments

wzamazon (approved):

wzamazon:
@wenduwan  Please backport

PR #8614, opened 2023-03-05
prov/tcp: accept() now supports io_uring

sydidelot:
The PR enables io_uring for the code path executed when an incoming request is accepted.


shefty:
thanks!  changes look good!

PR #8617, opened 2023-03-07
prov/tcp: Replace fabric->progress with eq->progress

shefty:


shefty:
Updated with a slight optimization when adding a domain to an eq.

prov/tcp/src/xnet_eq.c
```diff
@@ -158,57 +128,64 @@ static int xnet_eq_wait_try_func(void *arg)
...
+{
+	int ret;
 
-clean:
-	/* Traverse the list backwards from where the error occurred */
-	dlist_foreach_reverse(error_item, item) {
-		domain = container_of(item, struct xnet_domain,
-				      util_domain.list_entry);
-		xnet_eq_del_progress(eq, &domain->progress);
+	assert(!xnet_progress_locked(&domain->progress));
```
sydidelot, line 168: I tested your PR and it immediately triggers this assertion failure. This is expected.
Here, you want to verify that the current thread doesn't hold the progress lock in order to avoid a deadlock later in `xnet_start_progress()`. However, `xnet_progress_locked()` checks if the lock is acquired by anyone, not only the current thread. The lock might be acquired by another thread when the assertion failure is executed. 

After I remove this line (and the one line 134), everything works fine for my use case.

sydidelot (changes requested):
@shefty Thanks for the patch! I found an issue with the assertion failure `assert(!xnet_progress_locked(&domain->progress));`. I think it should be removed. 
FYI, I will be running some performance benchmarks with the new TCP provider later this week. I will keep you posted!

prov/tcp/src/xnet_eq.c
shefty: Excellent point, and yes, you understood why I added this check.  I'll update with the check removed.  I don't know if there's a way to check against the current thread only.

prov/tcp/src/xnet_eq.c
```diff
@@ -158,57 +128,62 @@ static int xnet_eq_wait_try_func(void *arg)
...
-int xnet_eq_add_progress(struct xnet_eq *eq, struct xnet_progress *progress,
-			 void *context)
+/* Safe to call even if domain was never added */
+static void xnet_eq_del_domain(struct xnet_eq *eq, struct xnet_domain *domain)
 {
-	return ofi_wait_add_fd(eq->util_eq.wait,
-			       ofi_dynpoll_get_fd(&progress->epoll_fd),
-			       POLLIN, xnet_eq_wait_try_func, NULL, context);
-}
+	assert(!xnet_progress_locked(&domain->progress));
```
sydidelot, line 134: Can you remove this one as well?

shefty: Updated -- I checked and didn't see anywhere else in the provider where I have a similar check for !locked.

sydidelot: Thank you!

sydidelot (approved):

shefty:
@sydidelot -- thanks for your help identifying the problem and validating this!

Issue #8587, opened 2023-03-01
fabtest/rdm_stress: count option should be reworked in the context of how the test operates as an RPC type of test


shefty:
Running rdm_stress with the tcp provider with stress.json produces this:

```
Data check error (8!=) at byte 1000 for buffer size 1000000
Data check error (8!=) at byte 1000 for buffer size 1000000
Data check error (8!=) at byte 1000 for buffer size 1000000
(11-13) rpc op tag_req iteration 0
(11-13) rpc op tag_resp iteration 0
Data check error (8!=) at byte 1000 for buffer size 1000000
(5-14) we're friends now
(5-14) rpc op msg_req iteration 0
(5-14) rpc op msg_req iteration 1
(5-14) rpc op msg_resp iteration 0
(14-10) rpc op msg_req iteration 0
(14-10) rpc op msg_resp iteration 0
(8-12) rpc op tag_req iteration 0
(8-12) rpc op tag_resp iteration 0
Data check error (8!=) at byte 1000 for buffer size 1000000
```
The errors at byte 1000 coincide with the length of the first transfer:
```
	{ "op" : "msg_req", "size" : 1000, "count" : 2 },
	{ "op" : "msg_resp" },
	{ "op" : "tag_req", "size" : 2000 },
	{ "op" : "tag_resp" },
	{ "op" : "msg_req", "size" : 1000000 },
```
Updating the .json file to remove the count parameter allows the test to pass.  I suspect this is because the test is sending 2 msg requests, since count is 2, but only requesting 1 response.  This throws off the rest of the test for the next msg transfer.

The count control may need to be reworked in the context of how the test operates as an RPC type of test.  For example, we may want something more like this in the config file:
```
	{ { "op" : "msg_req", "size" : 1000 },
	  { "op" : "msg_resp" }, "count" : 2},
	{ "op" : "tag_req", "size" : 2000 },
	{ "op" : "tag_resp" },
	{ "op" : "msg_req", "size" : 1000000 },
```

PR #8619, opened 2023-03-07
fabtests/rdm_stress: Remove count from json test configs

shefty:
The count parameter does not work generically when paired with message requests.  For each request, we need a corresponding response for the test to make progress.  However, we can't assume that the server will queue thousands of requests before we start looking for a response.  Flow control can back up the sending of requests.

Remove the count from the config files.


shefty:
@wckzhang - The code still handles the count parameter, but I removed it from the config files.  The test will fail with the count parameter included, since the config file will send multiple requests, but only ask for one response.  The count needs to be handled differently for RPCs, such as placing the count around both the request and response.  See #8587

PR #8630, opened 2023-03-09
core/hmem: Allow specifying Ze devices as device_handles

shefty:
Level zero has the concept of drivers and devices, with devices belonging to driver objects.  As a result, two devices can have the same index as long as they are from different drivers. To disambiguate what device might be referred to, extend the API to allow specifying the ze device using a ze_device_handle_t value.  (This is a pointer to a structure).

We determine if the device is given as an index or a pointer based on the range of values.

### DIFF
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -40,6 +40,9 @@ fi_mr_refresh
 fi_mr_enable
 : Enables a memory region for use.
 
+fi_hmem_ze_device
+: Returns an hmem device identifier for a level zero driver and device.
+
 # SYNOPSIS
 
 ```c
@@ -76,6 +79,8 @@ int fi_mr_refresh(struct fid_mr *mr, const struct iovec *iov,
     size_t count, uint64_t flags);
 
 int fi_mr_enable(struct fid_mr *mr);
+
+int fi_hmem_ze_device(int driver_index, int device_index);
 ```
 
 # ARGUMENTS
@@ -502,20 +507,22 @@ into calls as function parameters.
 
 ```c
 struct fi_mr_attr {
-	const struct iovec *mr_iov;
-	size_t             iov_count;
-	uint64_t           access;
-	uint64_t           offset;
-	uint64_t           requested_key;
-	void               *context;
-	size_t             auth_key_size;
-	uint8_t            *auth_key;
-	enum fi_hmem_iface iface;
-	union {
-		uint64_t         reserved;
-		int              cuda;
-		int		 ze
-	} device;
+    const struct iovec *mr_iov;
+    size_t             iov_count;
+    uint64_t           access;
+    uint64_t           offset;
+    uint64_t           requested_key;
+    void               *context;
+    size_t             auth_key_size;
+    uint8_t            *auth_key;
+    enum fi_hmem_iface iface;
+    union {
+        uint64_t       reserved;
+        int            cuda;
+        int            ze
+        int            neuron;
+        int            synapseai;
+    } device;
 };
 ```
 ## mr_iov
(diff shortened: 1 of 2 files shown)


man/fi_mr.3.md
```diff
@@ -650,7 +653,14 @@ This field is ignore unless the iface field is valid.
 : For FI_HMEM_CUDA, this is equivalent to CUdevice (int).
 
 *ze*
-: For FI_HMEM_ZE, this is equivalent to the ze_device_handle_t index (int).
+: For FI_HMEM_ZE, this is equivalent to the index of the device in the
+  ze_device_handle_t array.  A device index is only usable if the ze
+  driver is known, such as only one being present or usable by the app.
+
+*ze_device_handle*
+: For FI_HMEM_ZE, this is the ze_device_handle_t, cast to void *.  The
```
shijin-aws, line 661: I guess we should have done similar thing for gdr_handle in FI_HMEM_CUDA. Currently the handle and cuda device id shares the same `cuda` field in the API, and we need to have a way to toggle different memcpy ops for these two modes.

j-xiong:
I don't think the `ze` field has ever been used.

man/fi_mr.3.md
shefty: We may want to fix that now as part of this, in case we decide to go a different path.  For example, I considered adding a new FI_HMEM_ZE_HANDLE (or similar) value instead, to make it clear which value was being set.  I picked keeping with FI_HMEM_ZE and checking the range, but we can discuss that option, especially if we might need something like FI_HMEM_GDR_HANDLE.

shijin-aws: A separate hmem iface looks good to me, especially we already makes `hmem_gdrcopy.c` a separate file. And it will be easier for provider code to toggle. What do you think @wzamazon @wenduwan

shefty: This field is a device handle of some sort.  I would need to understand what would be passed in here by an app.

wenduwan: I'm leaning toward separate `FI_HMEM_*` to avoid the ambiguity of many device handles to one hmem iface mapping. It is simpler to have only 1 type of device handle for each iface. It also scales if somehow a device supports more than 2 memcpy modes, e.g. imagine cuda one day supports 1) cudaMemcpy 2) gdrcopy and 3) cudaFastCopy, whose device handles might overlap.

However, this is not perfect either, because sometimes we need optional methods to do memcpy. In the case of cuda hmem, we might choose to use gdrcopy for cuda <> host copy, but we have to use cudaMemcpy for cuda <> cuda copy. Suppose `FI_HMEM_CUDA` is separate from `FI_HMEM_GDR`, then we need a way to register both for the same MR. This is not straight forward.

For now, if we can always infer the handle by its range(aka no overlap), then we can probably keep the current impl using the `reserved` magic handle.

aingerson:
@shefty The device is passed into the registration call and is stored in the ofi_mr. That device is passed into the hmem copy calls so we will have to update the ze copy calls to use it as the handle and not index ([here](https://github.com/ofiwg/libfabric/blob/main/src/hmem_ze.c#L898)) as well as fabtests to register with the handle instead of the index which is passed in through the command line ([here](https://github.com/ofiwg/libfabric/blob/main/fabtests/common/shared.c#L393))

aingerson:
The thing we will have to figure out is a good way to go between device handle and picking the right Command Queue to use. That's why we used the index in the first place - because we have a lot of resources associated with each device so an index makes them easier to access.

man/fi_mr.3.md
shefty: This is used to identify the device.  It is not intended to convey information about memory copy modes or what calls are being used underneath.  We're only trying to identify the device, and in the case of Ze, a simple index is not guaranteed to work.

shefty:
@aingerson we could continue to use an index if we can define a fixed mapping that an app can use to convert a device to an index.

Instead of a handle, we could define something like:
```
struct ze_device {
    int driver_index;
    int device_index;
};
```
and update the union to include this instead.

shefty:
Alternatively, we could provide a macro for setting the device instead.  E.g.
```
fi_ze_device(driver_id, device_id) driver_id << 16 | device_id
```

aingerson:
@shefty I think updating the union to include the driver and device is more intuitive and more user friendly

shefty:
Updated.  I went with adding a new call, ```fi_hmem_ze_device(driver_index, device_index)``` that can be used to set the fi_mr_attr::device.ze field.  This is ABI/API compatible.  Introducing a new structure into the union requires adding a FI_HMEM_ZE_OOPS iface type to know if the driver index has been provided.

aingerson:
@shefty This looks good to me. Once this is merged, I can update the ze calls and fabtests to use it properly.

PR #8635, opened 2023-03-10
man: Remove the man page for the net provider

sydidelot:
The net provider was merged in the tcp provider and no longer exists. We can remove the associated man page.


shefty:
Can you also remove the generated page in man/man7/fi_net.7?

shefty:
No - I'll update that man page.

PR #8641, opened 2023-03-10
prov/efa: Implement start_msg and start_tag in peer API

sunkuamzn:
This commit implements the peer ops of the peer API and uses them internally for the EFA provider

The discard_msg and discard_tag ops are not implemented because they are not called by external applications and are not used by the EFA provider

### DIFF
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -168,6 +168,14 @@ struct rxr_op_entry *rxr_ep_alloc_rx_entry(struct rxr_ep *ep, fi_addr_t addr, ui
 	rx_entry->efa_outstanding_tx_ops = 0;
 	rx_entry->shm_outstanding_tx_ops = 0;
 	rx_entry->op = op;
+
+	rx_entry->peer_rx_entry.addr = addr;
+	/* This field points to the fid_peer_srx struct that's part of the peer API
+	*  We always set it to the EFA provider's SRX here. For SHM messages, we will set
+	*  this to SHM provider's SRX in the get_msg/get_tag function call
+	*/
+	rx_entry->peer_rx_entry.srx = &ep->peer_srx;
+
 	dlist_init(&rx_entry->entry);
 	switch (op) {
 	case ofi_op_tagged:
--- a/prov/efa/src/rdm/rxr_msg.c
+++ b/prov/efa/src/rdm/rxr_msg.c
@@ -536,7 +536,6 @@ int rxr_msg_match_peer_unexp_by_tag(struct dlist_entry *item, const void *arg)
 			     match_info->tag);
 }
 
-static
 int rxr_msg_handle_unexp_match(struct rxr_ep *ep,
 			       struct rxr_op_entry *rx_entry,
 			       uint64_t tag, uint64_t ignore,
@@ -544,6 +543,7 @@ int rxr_msg_handle_unexp_match(struct rxr_ep *ep,
 			       uint32_t op, uint64_t flags)
 {
 	struct rxr_pkt_entry *pkt_entry;
+	struct fid_peer_srx *srx;
 	uint64_t data_len;
 
 	rx_entry->fi_flags = flags;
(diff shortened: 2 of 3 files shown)


shijin-aws:
Sorry, I was suggesting making `srx` as an additional argument of `rxr_ep_alloc_rx_entry`. But I find it is not necessary as I can always override `rx_entry->peer_rx_entry.srx` in `get_msg` call directly. So you can simply assign `rx_entry->peer_rx_entry.srx = rxr_ep` in this function without more modification.

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -635,6 +645,9 @@ struct rxr_op_entry *rxr_msg_alloc_rx_entry(struct rxr_ep *ep,
 		memset(&rx_entry->desc[0], 0, sizeof(rx_entry->desc));
 
 	rx_entry->cq_entry.op_context = msg->context;
+
+	// TODO - set fi_peer_rx_entry fields required by SHM provider when allocating
```
shijin-aws, line 649: Yes, I plan to have it in my PR that implements get_*, we should refer to how shm uses the fi_peer_rx_entry: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_progress.c#L648-L721

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -580,7 +580,17 @@ int rxr_msg_handle_unexp_match(struct rxr_ep *ep,
 		rx_entry->ignore = ~0;
 	}
 
-	return rxr_pkt_proc_matched_rtm(ep, rx_entry, pkt_entry);
+	srx = rx_entry->peer_rx_entry.srx;
+
+	// TODO - populate rx_entry->fi_peer_rx_entry with info required by the SHM provider
```
shijin-aws, line 585: I think shm provider does similar thing, but it uses peer_context: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_progress.c#L769-L784. So this shouldn't be an issue. EFA provider should not care how shm use peer_rx_entry

shijin-aws:
Overall LGTM, thank you!

prov/efa/src/rdm/rxr_msg.c
shijin-aws: The `owner_context` seems to be filled as the owner entry of multi-recv buffer: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_ep.c#L1411-L1426. Though there is no definition of this field in the man page. Maybe open an issue to ask Intel for clarification?

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -168,6 +168,7 @@ struct rxr_op_entry *rxr_ep_alloc_rx_entry(struct rxr_ep *ep, fi_addr_t addr, ui
 	rx_entry->efa_outstanding_tx_ops = 0;
 	rx_entry->shm_outstanding_tx_ops = 0;
 	rx_entry->op = op;
+	rx_entry->peer_rx_entry.srx = peer_srx;
```
shijin-aws, line 171: we should also assign `rx_entry->peer_rx_entry.addr = addr` here.

shijin-aws, line 171: Also just do `rx_entry->peer_rx_entry.srx = ep->srx` here as suggested, and remove the additional input

prov/efa/src/rdm/rxr_msg.c
sunkuamzn: I started a discussion https://github.com/ofiwg/libfabric/discussions/8646 and emailed Alexia

shijin-aws: Cool, it seems using `owner_context` is good here. Thanks!

prov/efa/src/rdm/rxr_msg.c
shijin-aws: My latest finding is we really do not need to set `fi_peer_rx_entry` fields in this commonly used function as owner provider (EFA) will not use them for now. I can set them in `get_msg/tag` after getting an rx entry.

prov/efa/src/rdm/rxr_msg.c
shijin-aws: Back to your `TODO:`, now I think you should populate the info in `rx_entry` to `rx_entry->peer_rx_entry`.  I think they should be implemented somewhere here: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/rxr_msg.c#L860-L871 before calling `rxr_msg_handle_unexp_match`

prov/efa/src/rdm/rxr_ep.c
sunkuamzn: fixed

sunkuamzn: done

prov/efa/src/rdm/rxr_msg.c
sunkuamzn: > you should populate the info in `rx_entry` to `rx_entry->peer_rx_entry` in unexp path

Only SHM provider needs this info. EFA provider will just use `rx_entry`. So I didn't implement that now and added the `TODO` instead. I can add the info when implementing the SHM provider?

shijin-aws: I see, sure. It can be in a separate commit when integrating with shm.

prov/efa/src/rdm/rxr_msg.c
shijin-aws: >Only SHM provider needs this info. EFA provider will just use rx_entry. So I didn't implement that now and added the TODO instead. I can add the info when implementing the SHM provider?

After thinking about this comment. I think I will hold this change in my owner_ops PR as well before integrating with shm. Let's keep the TODO here then.

shijin-aws (approved):

Issue #6745, opened 2021-05-07
prov/tcp RFC: CQ spin polling without poll() overhead


sydidelot:
Hi,

I am currently running some performance measurements and profiling with the TCP provider (latest master).

The TCP provider relies on epoll()/poll() to monitor the socket events and progress pending socket operations.
It also needs to update the POLLOUT event in the poll set if outgoing buffers are waiting for the given socket. 
This scenario also happens when the user application wants to spin for completion (CQ configured with `FI_WAIT_NONE`).

I understand that poll is the best solution to minimize the CPU utilization while waiting for socket events.
However, some users (I am part of theem :)) may want to minimize the impact of poll() in the fast path. 

I did run some latency benchmarks with fabtest (command line is `./fi_msg_pingpong -S all -c spin`) where I compare libfabric *with* and *without* calling poll(). Here is the corresponding patch if you want to give it a try:
```
diff --git a/prov/tcp/src/tcpx_cq.c b/prov/tcp/src/tcpx_cq.c
index 0828b6194..94ea25555 100644
--- a/prov/tcp/src/tcpx_cq.c
+++ b/prov/tcp/src/tcpx_cq.c
@@ -37,6 +37,7 @@
 
 #define TCPX_DEF_CQ_SIZE (1024)
 
+#define POLL_ENABLED
 
 void tcpx_cq_progress(struct util_cq *cq)
 {
@@ -63,14 +64,22 @@ void tcpx_cq_progress(struct util_cq *cq)
                 * of length 0, there's no additional data to read, so failing
                 * to progress can result in application hangs.
                 */
+#ifndef POLL_ENABLED
+               tcpx_progress_tx(ep);
+               tcpx_progress_rx(ep);
+#else
                if (ofi_bsock_readable(&ep->bsock) ||
                    (ep->cur_rx.handler && !ep->cur_rx.entry))
                        tcpx_progress_rx(ep);
+#endif
 
+#ifdef POLL_ENABLED
                (void) tcpx_update_epoll(ep);
+#endif
                fastlock_release(&ep->lock);
        }
 
+#ifdef POLL_ENABLED
        nfds = (wait_fd->util_wait.wait_obj == FI_WAIT_FD) ?
               ofi_epoll_wait(wait_fd->epoll_fd, wait_contexts,
                              MAX_POLL_EVENTS, 0) :
@@ -96,6 +105,7 @@ void tcpx_cq_progress(struct util_cq *cq)
[...]

PR #8645, opened 2023-03-13
prov/tcp: Complete the io_uring implementation (poll_add + polling without poll/epoll overhead)

sydidelot:
This PR is hopefully the last set of major io_uring changes for the TCP provider.
It moves all the poll/epoll operations to io_uring, which allows for progressing outstanding io_uring operations without the extra overhead of epoll()/poll().
As a side-effect it also fixes the issue #6745, which was the real trigger for the entire io_uring work. The circle is now complete :)

### DIFF
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -533,6 +533,7 @@ struct ofi_bsock {
 	struct ofi_sockapi *sockapi;
 	struct ofi_sockctx tx_sockctx;
 	struct ofi_sockctx rx_sockctx;
+	struct ofi_sockctx pollin_sockctx;
 	struct ofi_byteq sq;
 	struct ofi_byteq rq;
 	size_t zerocopy_size;
@@ -549,6 +550,7 @@ ofi_bsock_init(struct ofi_bsock *bsock, struct ofi_sockapi *sockapi,
 	bsock->sockapi = sockapi;
 	ofi_sockctx_init(&bsock->tx_sockctx, context);
 	ofi_sockctx_init(&bsock->rx_sockctx, context);
+	ofi_sockctx_init(&bsock->pollin_sockctx, context);
 	ofi_byteq_init(&bsock->sq, sbuf_size);
 	ofi_byteq_init(&bsock->rq, rbuf_size);
 	bsock->zerocopy_size = SIZE_MAX;
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -521,7 +521,6 @@ struct ofi_bsock {
 	struct ofi_sockapi *sockapi;
 	struct ofi_sockctx tx_sockctx;
 	struct ofi_sockctx rx_sockctx;
-	struct ofi_sockctx cancel_sockctx;
 	struct ofi_byteq sq;
 	struct ofi_byteq rq;
 	size_t zerocopy_size;
@@ -538,7 +537,6 @@ ofi_bsock_init(struct ofi_bsock *bsock, struct ofi_sockapi *sockapi,
 	bsock->sockapi = sockapi;
 	ofi_sockctx_init(&bsock->tx_sockctx, context);
 	ofi_sockctx_init(&bsock->rx_sockctx, context);
-	ofi_sockctx_init(&bsock->cancel_sockctx, context);
 	ofi_byteq_init(&bsock->sq, sbuf_size);
 	ofi_byteq_init(&bsock->rq, rbuf_size);
 	bsock->zerocopy_size = SIZE_MAX;
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -318,6 +318,10 @@ int ofi_sockctx_uring_cancel(struct ofi_sockapi_uring *uring,
 			     struct ofi_sockctx *canceled_ctx,
 			     struct ofi_sockctx *ctx);
 
+int ofi_sockctx_uring_poll_add(struct ofi_sockapi_uring *uring,
+			       int fd, short poll_mask, bool multishot,
+			       struct ofi_sockctx *ctx);
+
 int ofi_uring_init(ofi_io_uring_t *io_uring, size_t entries);
 int ofi_uring_destroy(ofi_io_uring_t *io_uring);
 
@@ -409,6 +413,14 @@ static inline int
 ofi_sockctx_uring_cancel(struct ofi_sockapi_uring *uring,
 			 struct ofi_sockctx *canceled_ctx,
 			 struct ofi_sockctx *ctx)
+{
+	return -FI_ENOSYS;
+}
+
+static inline int
+ofi_sockctx_uring_poll_add(struct ofi_sockapi_uring *uring,
+			   int fd, short poll_mask, bool multishot,
+			   struct ofi_sockctx *ctx)
 {
 	return -FI_ENOSYS;
 }
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -362,6 +362,8 @@ static inline void ofi_uring_cq_advance(ofi_io_uring_t *io_uring, unsigned int c
 	io_uring_cq_advance(io_uring, count);
 }
 #else
+#define IORING_CQE_F_MORE	(1U << 1)
+
 static inline int
 ofi_sockapi_connect_uring(struct ofi_sockapi *sockapi, SOCKET sock,
 			  const struct sockaddr *addr, socklen_t addrlen,
--- a/prov/tcp/src/xnet_ep.c
+++ b/prov/tcp/src/xnet_ep.c
@@ -192,6 +192,12 @@ int xnet_setup_socket(SOCKET sock, struct fi_info *info)
 
 static int xnet_monitor_ep(struct xnet_progress *progress, struct xnet_ep *ep)
 {
+	if (xnet_io_uring) {
+		assert(!(ep->pollflags & POLLOUT));
+		return xnet_uring_pollin_add(progress, ep->bsock.sock,
+					     false, &ep->bsock.pollin_sockctx);
+	}
+
 	return xnet_monitor_sock(progress, ep->bsock.sock, ep->pollflags,
 				 &ep->util_ep.ep_fid.fid);
 }
@@ -398,7 +404,8 @@ void xnet_ep_disable(struct xnet_ep *ep, int cm_err, void* err_data,
 	};
 
 	dlist_remove_init(&ep->unexp_entry);
-	xnet_halt_sock(xnet_ep2_progress(ep), ep->bsock.sock);
+	if (!xnet_io_uring)
+		xnet_halt_sock(xnet_ep2_progress(ep), ep->bsock.sock);
 
 	ret = ofi_shutdown(ep->bsock.sock, SHUT_RDWR);
 	if (ret && ofi_sockerr() != ENOTCONN)
@@ -549,7 +556,8 @@ static int xnet_ep_close(struct fid *fid)
 	progress = xnet_ep2_progress(ep);
 	ofi_genlock_lock(&progress->lock);
 	dlist_remove_init(&ep->unexp_entry);
-	xnet_halt_sock(progress, ep->bsock.sock);
+	if (!xnet_io_uring)
+		xnet_halt_sock(progress, ep->bsock.sock);
 	xnet_ep_flush_all_queues(ep);
 	ofi_genlock_unlock(&progress->lock);
 
--- a/prov/tcp/src/xnet_ep.c
+++ b/prov/tcp/src/xnet_ep.c
@@ -403,6 +403,7 @@ void xnet_ep_disable(struct xnet_ep *ep, int cm_err, void* err_data,
 		return;
 	};
 
+	ep->state = XNET_DISCONNECTED;
 	dlist_remove_init(&ep->unexp_entry);
 	if (!xnet_io_uring)
 		xnet_halt_sock(xnet_ep2_progress(ep), ep->bsock.sock);
@@ -430,7 +431,6 @@ void xnet_ep_disable(struct xnet_ep *ep, int cm_err, void* err_data,
 		(void) xnet_eq_write(ep->util_ep.eq, FI_SHUTDOWN,
 				     &cm_entry, sizeof(cm_entry), 0);
 	}
-	ep->state = XNET_DISCONNECTED;
 }
 
 static int xnet_ep_shutdown(struct fid_ep *ep_fid, uint64_t flags)
@@ -555,9 +555,11 @@ static int xnet_ep_close(struct fid *fid)
 
 	progress = xnet_ep2_progress(ep);
 	ofi_genlock_lock(&progress->lock);
+	ep->state = XNET_DISCONNECTED;
 	dlist_remove_init(&ep->unexp_entry);
 	if (!xnet_io_uring)
 		xnet_halt_sock(progress, ep->bsock.sock);
+	ofi_close_socket(ep->bsock.sock);
 	xnet_ep_flush_all_queues(ep);
 	ofi_genlock_unlock(&progress->lock);
 
@@ -568,7 +570,6 @@ static int xnet_ep_close(struct fid *fid)
 
 	free(ep->cm_msg);
 	free(ep->addr);
-	ofi_close_socket(ep->bsock.sock);
 
 	ofi_endpoint_close(&ep->util_ep);
 	free(ep);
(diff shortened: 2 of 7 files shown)


prov/tcp/src/xnet_progress.c
```diff
@@ -1130,22 +1130,35 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
...
 		xnet_uring_run_conn(conn, cqe->res);
+		break;
+	default:
+		assert(fid->fclass == FI_CLASS_PEP);
+	case FI_CLASS_PEP:
+		pep = container_of(fid, struct xnet_pep, util_pep.pep_fid.fid);
+		if (sockctx == &pep->pollin_sockctx) {
+		    if (cqe->res >= 0)
+				xnet_accept_sock(pep);
+		} else
```
shefty, line 1160: nit: need braces around else

shefty (commented):
2 nits, but I need to review this a second time to make sure I understand what's happening.  Do we reserve space in the iouring for the management requests (cancel, pollin, pollout)?

prov/tcp/src/xnet_progress.c
sydidelot: > Do we reserve space in the iouring for the management requests (cancel, pollin, pollout)?
No we don't reserve space for management requests. That's definitely a problem with the current design if the ring runs out of credits. I'm not sure exactly how to gracefully address this problem.

shefty: Can we reduce credits by 4 (or whatever) and track those as mgmt_credits?  Though I'm not sure we even need to track that, versus just tracking that we only have 1 mgmt request per type outstanding at a time.

include/ofi_net.h
```diff
@@ -522,6 +536,8 @@ struct ofi_bsock {
 	struct ofi_sockctx tx_sockctx;
 	struct ofi_sockctx rx_sockctx;
 	struct ofi_sockctx cancel_sockctx;
+	struct ofi_sockctx pollin_sockctx;
+	struct ofi_sockctx pollout_sockctx;
```
shefty, line 540: We're up to 5 sockctx structs per socket now.  Does it make sense to allocate a pool of sockctx with the domain and grow that as needed?  Quickly searching the code, it looks like that may be a significant change, given how frequently uring_sqe_inuse is accessed.  I'm guessing we'd still need to keep that level of state, but maybe a single set of flags could track that instead?

prov/tcp/src/xnet_progress.c
```diff
@@ -1061,22 +1130,35 @@ static void xnet_progress_cqe(struct xnet_progress *progress,
...
+	case FI_CLASS_CONNREQ:
 		conn = container_of(fid, struct xnet_conn_handle, fid);
 		xnet_uring_run_conn(conn, cqe->res);
+		break;
+	default:
+		assert(fid->fclass == FI_CLASS_PEP);
+	case FI_CLASS_PEP:
+		pep = container_of(fid, struct xnet_pep, util_pep.pep_fid.fid);
+		if (sockctx == &pep->pollin_sockctx) {
+		    if (cqe->res >= 0)
```
shefty, line 1158: indentation is off

prov/tcp/src/xnet_pep.c
```diff
@@ -241,8 +252,14 @@ int xnet_listen(struct xnet_pep *pep, struct xnet_progress *progress)
 	}
 
 	ofi_genlock_lock(&progress->lock);
-	ret = xnet_monitor_sock(progress, pep->sock, POLLIN,
-				&pep->util_pep.pep_fid.fid);
+	if (xnet_io_uring) {
+		ret = xnet_uring_poll_add(progress, pep->sock, POLLIN,
+					  true, &pep->pollin_sockctx, NULL);
```
shefty, line 257: If I understand correctly, multishot means that this pollin_sockctx will remain inuse even after an event has been written the cqe.  When is that cleared?

prov/tcp/src/xnet_progress.c
sydidelot: The problem is that the ring is shared between all endpoints from the same domain. If we run out of `mgmt_credits` and an endpoint needs to submit a new `mgmt` SQE (pollin, pollout or cancel), it will need to wait for a new credit to become available. Before this PR, the code was relying on epoll to retry an endpoint when the ring runs out of credits. But since we are moving epoll operations to the ring, this solution no longer works. 

I see several options here:
- Provision enough entries in the ring in order to make sure it never overflows.
- Leave the responsibility to the application to guarantee that no overflow would occur on the ring. If an overflow occurs, endpoints will be automatically shut down and disabled. This option is a similar to the Completion Queues with Infiniband where an overflow would transition all associated endpoints to an error state.
- Each domain has a pool of rings, which can grow on-demand.

shefty: I don't see how we can make the app responsible for this.  With CQs, the app at least knows that there's a completion per request, so it has a way to track it.  But pollin/pollout are internal and/or related to the remote side.  It seems like the current solution would be to always allocate a giant ring and hope it's big enough.  I know there have been recent discussions on the kernel mailing list about having the ability to resize the rings, specifically because network apps can't guess the needed size up front.  Hopefully we can figure out some way to keep things running smoothly in cases where the ring does get full (and assuming that progress is being made on completing some number of queued requests).

shefty: Avoiding overrunning io_uring is somewhat equivalent to avoiding overrunning the verbs CQ.  So, maybe this isn't that big of a deal, at least I think it's the same verbs would have.

sydidelot: Hi @shefty, I’ve been thinking a bit more about that io_uring overrunning issue the last couple of days. Here are my conclusions:

When the provider issues an operation but the ring is running out of credits, we could submit all pending SQEs, reap the ready CQEs and save them into a bufpool until one credit gets released. This solution is however not perfect because there is absolutely no guarantee that a single CQE will be released if the ring is full. For example, a posted receive buffer may be released only if a packet is received, the remote process is dead or the corresponding SQE is canceled.

The credit system we have implemented in the TCP provider guarantees that the SQ never overruns. With io_uring, the SQE is immediately released after the kernel has consumed the request. Moreover, with relatively recent Linux kernels, io_uring guarantees that a single CQ almost never drops events in case of overflow (unless the kernel runs out of memory). If the overflow condition is reached, submitting new SQEs would fail with -EBUSY.

I’ve reviewed a couple of applications supporting io_uring and some of them (e.g., tokio-rs https://github.com/tokio-rs/tokio-uring) seem to completely rely on the capacity of io_uring to overflow. It also seems to be the solution recommended by Jens (see https://github.com/axboe/liburing/issues/396)
If get_sqe returns NULL, then the SQ is overflowed and ready SQEs need to be submitted. SQEs that are consumed by the kernel are immediately released and become available to the application. If submitting SQEs fails with -EBUSY, then the CQ is overflowed and the application needs to poll the CQ and reap CQEs. Overflowing the queues may significantly impact the overflow performance, but at least the application should continue running if the number of ring entries is underestimated by the upper layer.

I’ve implemented the above solution in libfabric (i.e., removed the ring credits from the TCP provider and let the ring overflow), but the code doesn’t work well. It seems that the version of the Linux kernel I’m using for development doesn’t retry overflowed CQEs and, as a result, CQEs get processed out of order. This problem is described here: https://lore.kernel.org/io-uring/20200707132420.2007-1-xiaoguang.wang@linux.alibaba.com/

For the moment, it’s probably safer to keep our own ring credits in the TCP provider and reap SQEs if all the credits are exhausted. Reaped SQEs would be copied into a bufpool entry and added into a list that would be processed when polling the ring, similarly to the verbs provider when the CQ is flushed. What do you think?

include/ofi_net.h
sydidelot: I need to think a bit more about that... The `ofi_sockctx` structure is composed of a boolean + a pointer to the `ofi_bsock` structure. When a CQE is polled, the completion function needs to resolve what `ofi_bsock` structure has triggered the completion, and the type of the completion event. The problem is that io_uring allows for a 8-byte context only :)

shefty: Can the ofi_sockctx carry this information?  It makes sense for ofi_bsock to have the tx/rx contexts.  But if the others are used sparingly, a pool might make sense.

prov/tcp/src/xnet_progress.c
shefty: It should work okay in practice to track credits ourselves.  It's easy to fail transmit requests with EAGAIN if there aren't credits available.  Receives are another matter, as that could lead to deadlock on large systems.  As long as the number of credits is > 2 x # peers, we should be okay.  We just need a large enough queue to support the maximum number of peers that we expect to support.  I would expect that to be in the low thousands.

include/ofi_net.h
sydidelot: > Can the ofi_sockctx carry this information?

What information you mean? `ofi_sockctx` contains all the information to figure out what operation (RX, TX, cancel, etc...) completed and for which FID. 

>  It makes sense for ofi_bsock to have the tx/rx contexts. But if the others are used sparingly, a pool might make sense.

Mixing tx/rx sockctxs from the ofi_bsock and sockctx from a dynamic pool for the others would significantly complicate the sockapi interface. I also tried to move all our sockctx into a bufpool but that also requires many changes to the sockapi interface.

Considering that canceling a request is a blocking operation, `cancel_sockctx` does not need to stored in the sockapi. It can just be allocated on the caller's stack. As for `pollin_sockctx`, I'm not sure it's effectively required: if we can keep a receive buffer always posted, we should never need to wait for a POLLIN event. It should be the case already, but I haven't be able to confirm it's always true.

shefty: I don't want to complicate the implementation for a minor optimization, like saving a few bytes of data per ep. If 2 of the sockctx aren't needed and can be removed at some point, then we're okay.

sydidelot:
@shefty I updated the PR with a the following changes: 
- I removed the cancel sockctx from EP. Since cancellation is always blocking, the cancel sockctx can be allocated on the caller's stack.
- I reworked the code to remove the pollout sockctx from the EP. Because of the asynchronous nature of io_uring, we don't expect the provider to register POLLOUT events unless the ring is running out of credits.

I have a draft patch in my private branch to mitigate the issues when rings run out of credits. I'll push the code in a separate PR after this one is merged.

Please review when you have time and let me know what you think of the implementation. Thanks!

shefty (commented):
Posting review comments, which are minor.  But I'm still analyzing the patch "prov/tcp: Add return error to xnet_update_pollflag".

prov/tcp/src/xnet_progress.c
```diff
@@ -241,27 +241,29 @@ void xnet_recv_saved(struct xnet_xfer_entry *saved_entry,
...
 
 		ep->pollflags &= ~pollflag;
 	}
 
-	ofi_dynpoll_mod(&progress->epoll_fd, ep->bsock.sock,
-			ep->pollflags, &ep->util_ep.ep_fid.fid);
+	ret = ofi_dynpoll_mod(&progress->epoll_fd, ep->bsock.sock,
+			      ep->pollflags, &ep->util_ep.ep_fid.fid);
 	xnet_signal_progress(progress);
+	return ret;
```
shefty, line 305: In the case of epoll, if we need to set a flag and fail, we have the potential to hang the app by missing events.  If we're clearing the flag and can't, I think that's recoverable.  We may generate events that we don't want, which can be inefficient, but it's not fatal.  Plus, we'll retry clearing the flag later (until we eventually succeed).

Whether we want to ignore errors clearing the flag is debatable.  The larger question is the behavior when we can't ignore the errors, and disabling the ep seems reasonable.

prov/tcp/src/xnet_progress.c
```diff
@@ -386,12 +388,14 @@ static void xnet_progress_tx(struct xnet_ep *ep)
...
+			if (!ret)
+				return;
 		} else if (ret == -OFI_EINPROGRESS_URING) {
-			xnet_update_pollflag(ep, POLLOUT, false);
-			return;
-		}
+			ret = xnet_update_pollflag(ep, POLLOUT, false);
+			if (!ret)
+				return;
+		 }
```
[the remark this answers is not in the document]
shefty: My concern with the above change is that the handling of xnet_update_pollflag() isn't obvious in our intent.  For example, xnet_send_msg() could have been successful, with data sent on the wire.  However, if xnet_update_pollflag() fails, we fall through and fail the send (xnet_complete_tx() below).  AFAICT, the ep state remains connected.  But xnet_send_msg() could have written the header into the tcp stream.  If we drop the message, the stream is now corrupt, as the rest of the message won't be sent.

It may be better on xnet_update_pollflag() failure to jump to where xnet_ep_disable() is called.  That should fail the queued send as canceled.

prov/tcp/src/xnet_progress.c
```diff
@@ -551,7 +557,9 @@ int xnet_start_recv(struct xnet_ep *ep, struct xnet_xfer_entry *rx_entry)
 	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
 	if (!dlist_empty(&ep->unexp_entry)) {
 		dlist_remove_init(&ep->unexp_entry);
-		xnet_update_pollflag(ep, POLLIN, true);
+		ret = xnet_update_pollflag(ep, POLLIN, true);
+		if (ret)
+			goto poll_err;
```
shefty, line 602: This flow looks okay to me (for now).  I think the error propagates back and disables the endpoint.  But, I have a request from middleware for truncation errors to report an error completion without disabling the ep.  That can be handled later, and likely requires more work anyway (like flushing the truncated data from the tcp stream).

shefty (commented):
See second review regarding error handling in the case of update_pollflag().  I think the only request is to change the error handling in xnet_send_msg() to force the ep into an error state.

prov/tcp/src/xnet_progress.c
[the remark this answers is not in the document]
sydidelot: Thanks for the explanation. I have updated the code as you suggested and the function now disables the EP if `xnet_update_pollflag()` returns an error.

prov/tcp/src/xnet_ep.c
```diff
@@ -415,6 +415,7 @@ void xnet_ep_disable(struct xnet_ep *ep, int cm_err, void* err_data,
 	if (ret && ofi_sockerr() != ENOTCONN)
 		FI_WARN(&xnet_prov, FI_LOG_EP_DATA, "shutdown failed\n");
 
+	ep->state = XNET_DISCONNECTED;
```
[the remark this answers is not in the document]
sydidelot: That's a good suggestion. I moved the line right after checking the ep state as you suggested.

prov/tcp/src/xnet_ep.c
```diff
@@ -562,6 +562,8 @@ static int xnet_ep_close(struct fid *fid)
 	dlist_remove_init(&ep->unexp_entry);
 	if (!xnet_io_uring)
 		xnet_halt_sock(progress, ep->bsock.sock);
+	ofi_close_socket(ep->bsock.sock);
+	ep->state = XNET_DISCONNECTED;
```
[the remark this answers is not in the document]
sydidelot: I moved the line right after taking the lock as you suggested.

prov/tcp/src/xnet_progress.c
```diff
@@ -260,9 +276,32 @@ static int xnet_update_pollflag(struct xnet_ep *ep, short pollflag, bool set)
...
+				return -FI_ENOSYS;
+
+			/* A TX SQE is in use and will wake us up */
+			ep->pollflags &= ~POLLOUT;
+		}
+
+		if ((ep->pollflags & POLLIN) &&
+		    ep->bsock.rx_sockctx.uring_sqe_inuse) {
+			/* A RX SQE is in use and will wake us up */
+			ep->pollflags &= ~POLLIN;
```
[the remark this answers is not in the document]
sydidelot: You are right, we can return 0 here. I also added an assert to make sure POLLIN and POLLOUT flags are both unset in ep->pollflags before leaving the function.

prov/tcp/src/xnet_progress.c
sydidelot: Ack.

sydidelot:
@shefty Thanks for the review! I have addressed all your review comments and CI seems happy :tada:

shefty:
Thanks!

PR #8647, opened 2023-03-13
fabtests/pytest/efa: execute lsmod in interactive shell

wenduwan:
This is a follow up bugfix to PR8628. On RHEL7 lsmod is not on the default ssh PATH. For portability, execute the command in interactive shell.

Need to backport to v1.17.x.

### DIFF
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -58,7 +58,7 @@ def has_gdrcopy(hostname):
     hostname: a host
     return: a boolean
     """
-    command = "ssh {} lsmod | grep gdrdrv".format(hostname)
+    command = "ssh {} /bin/bash --login -c lsmod | grep gdrdrv".format(hostname)
     process = subprocess.run(command, shell=True, check=False, stdout=subprocess.PIPE)
     return process.returncode == 0
 


lrbison (approved):

lrbison:
How strange. where is it located in that case?

wenduwan:
> How strange. where is it located in that case?

It is at `/usr/sbin/lsmod` which is on the PATH of interactive shell sessions. But `ssh host lsmod` does not follow that PATH.

We were snagged on this previously for `lspci` as well. https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/conftest.py#L110

wzamazon:
> It is at /usr/sbin/lsmod which is on the PATH of interactive shell sessions. But ssh host lsmod does not follow that PATH.

Why not just use `/usr/sbin/lsmod`?

wenduwan:
> > It is at /usr/sbin/lsmod which is on the PATH of interactive shell sessions. But ssh host lsmod does not follow that PATH.
> 
> Why not just use `/usr/sbin/lsmod`?

/usr/sbin/lsmod does not exist for e.g. ubuntu1804

wzamazon (approved):

wzamazon:
Ignore intel CI result because intel CI does not use pytest scripts

PR #8648, opened 2023-03-13
man/fi_peer: fix and clarify context descriptions

aingerson:
user_context was split into two context - owner_context and peer_context for the owner and peer to use, respectively. This updates references to user_context to the applicable new context and adds some more clarification to the usage of the rx_entry fields, especially the contexts.


shijin-aws (approved):

shijin-aws:
Makes sense to me, thank you!

aingerson:
Updated just to fix numbering issue in example flow

sunkuamzn:
Thanks!

PR #8649, opened 2023-03-13
prov/efa: remove redundant progress after triggering handshake

wenduwan:
It is unnecessary to call progress engine immediately after triggering a handshake. rxr_msg_generic_send does this when an error is returned.

### DIFF
--- a/prov/efa/src/rdm/rxr_msg.c
+++ b/prov/efa/src/rdm/rxr_msg.c
@@ -191,17 +191,13 @@ ssize_t rxr_msg_post_rtm(struct rxr_ep *ep, struct rxr_op_entry *tx_entry, int u
 	}
 
 	/*
-	 * rtm_type requrirs an extra feature, which peer might not support.
+	 * rtm_type requires an extra feature, which peer might not support.
 	 *
 	 * Check handshake packet from peer to verify support status.
 	 */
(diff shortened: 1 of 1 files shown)


lrbison:
It does?  Doesn't it take the rxr_pkt_post_one path? I don't see retries there, and there is an [early exit ](https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/rxr_pkt_cmd.c#L482) from trigger handshake, although I'm not certain progress helps.

wenduwan:
> It does? Doesn't it take the rxr_pkt_post_one path? I don't see retries there, and there is an [early exit ](https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/rxr_pkt_cmd.c#L482) from trigger handshake, although I'm not certain progress helps.

The caller progresses [here](https://github.com/ofiwg/libfabric/blob/95f9a16afdbe2a78d8043f9aa19afe682439d431/prov/efa/src/rdm/rxr_msg.c#L267-L268), because we return EAGAIN here.

wzamazon:
Is there a performance benefit of this change?

wenduwan:
> Is there a performance benefit of this change?

I did not benchmark it. I don't imagine a noticeable improvement either since handshake only occurs once per peer pair.

PR #8650, opened 2023-03-13
core/tostr: Add helper functions ofi_tostr_size() and ofi_tostr_count().

peinanz:
ofi_tostr_size() converts size to str with 1024 base G/M/K. ofi_tostr_count() convers count to str with 1000 base B/M/K.

### DIFF
--- a/include/windows/osd.h
+++ b/include/windows/osd.h
@@ -869,7 +869,34 @@ static inline char * strndup(char const *src, size_t n)
 	return dst;
 }
 
-char *strcasestr(const char *haystack, const char *needle);
+static inline char *strcasestr(const char *haystack, const char *needle)
+{
+	char *uneedle, *uhaystack, *pos = NULL;
+	int i;
+
+	uneedle = malloc(strlen(needle) + 1);
+	uhaystack = malloc(strlen(haystack) + 1);
+	if (!uneedle || !uhaystack) {
+		free(uneedle);
+		free(uhaystack);
+		return pos;
+	}
+
+	for (i = 0; i < strlen(needle); i++)
+		uneedle[i] = (char) toupper(needle[i]);
+	uneedle[i] = '\0';
+	for (i = 0; i < strlen(haystack); i++)
+		uhaystack[i] = (char) toupper(haystack[i]);
+	uhaystack[i] = '\0';
+	pos = strstr(uhaystack, uneedle);
+	if (pos)
+		pos = (char *) ((uintptr_t) haystack + (uintptr_t) pos -
+						(uintptr_t) uhaystack);
+	free(uneedle);
+	free(uhaystack);
+
+	return pos;
+}
 
 #ifndef _SC_PAGESIZE
 #define _SC_PAGESIZE	0
--- a/src/shared/ofi_str.c
+++ b/src/shared/ofi_str.c
@@ -33,11 +33,14 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <assert.h>
 #include <ctype.h>
 
 #include <rdma/fi_errno.h>
+
+#include "ofi.h"
 
 /*
  * TODO remove this and include ofi_osd.h which would require merging osd.h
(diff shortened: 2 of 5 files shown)


src/shared/ofi_str.c
```diff
@@ -199,3 +200,50 @@ void ofi_free_string_array(char **s)
...
 	free(s);
 }
+
+char *ofi_tostr_size(char *str, size_t len, uint64_t size)
+{
+	uint64_t base = 0;
+	uint64_t fraction = 0;
+	char mag;
+
+	memset(str, '\0', len);
```
shefty, line 210: We don't need to memset the entire character array, which may be large.  Just ensure that the array is null terminated before exiting.

src/shared/ofi_str.c
```diff
@@ -199,3 +200,50 @@ void ofi_free_string_array(char **s)
...
+		snprintf(str, len, "%lu.%lu%c", size / base, fraction, mag);
+	else
+		snprintf(str, len, "%lu%c", size / base, mag);
+
+	return str;
+}
+
+char *ofi_tostr_count(char *str, size_t len, uint64_t count)
+{
+	memset(str, '\0', len);
```
shefty, line 238: Same as above.

src/shared/ofi_str.c
```diff
@@ -199,3 +200,50 @@ void ofi_free_string_array(char **s)
...
+		mag = '\0';
+	}
+
+	if (size / base < 10)
+		fraction = (size % base) * 10 / base;
+
+	if (fraction)
+		snprintf(str, len, "%lu.%lu%c", size / base, fraction, mag);
+	else
+		snprintf(str, len, "%lu%c", size / base, mag);
```
shefty, line 231: We may want to match the behavior of the other ofi_tostr... functions that are in fi_tostr.c.  Those call ofi_strncatf() instead of snprintf(), and concatenate the new string to the end of the current one, rather than replacing the existing string.

That may match the expected use case for these functions as well.

src/shared/ofi_str.c
```diff
@@ -33,6 +33,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <stdint.h>
```
shefty, line 36: Need to include "ofi.h" for ofi_strncatf().  Cause of build error on OS X

shefty (approved):

shefty:
Windows build failed:
```
"C:\projects\libfabric\libfabric.sln" (default target) (1) ->
"C:\projects\libfabric\libfabric.vcxproj" (default target) (2) ->
(ClCompile target) -> 
  c:\projects\libfabric\src\shared\ofi_str.c(53): error C2084: function 'char *strsep(char **,const char *)' already has a body [C:\projects\libfabric\libfabric.vcxproj]
```
This may be a problem in the existing code, and not directly introduced here, except by the header file inclusion.

shefty:
I *think* we need to remove the strsep() call in ofi_str.c that's surrounded by #if defined _WIN32.  strsep() is already defined in include/windows/osd.h, so this is a duplicate definition.  Also, the strcasestr() definition in ofi_str.c is mis-located.  Ideally, that should move into src/windows/osd.c

src/shared/ofi_str.c
```diff
@@ -42,59 +42,8 @@
...
-	if (pos)
-		pos = (char *) ((uintptr_t) haystack + (uintptr_t) pos -
-				(uintptr_t) uhaystack);
-out:
-	free(uneedle);
-	free(uhaystack);
-	return pos;
-}
-
+#include <windows/osd.h>
```
shefty, line 46: We should not need this include here.  The osd.h file should have been included earlier separately.

include/windows/osd.h
```diff
@@ -682,6 +682,8 @@ static inline char* strsep(char **stringp, const char *delim)
 	return ptr;
 }
 
+char *strcasestr(const char *haystack, const char *needle);
```
shefty, line 685: This should already exist on line 872 (please verify) of this file and can be removed.

src/windows/osd.c
```diff
@@ -647,3 +647,32 @@ struct ofi_pollfds_ctx *ofi_pollfds_alloc_ctx(struct ofi_pollfds *pfds, int fd)
...
+
+    pos = strstr(uhaystack, uneedle);
+    if (pos)
+        pos = (char *) ((uintptr_t) haystack + (uintptr_t) pos -
+                (uintptr_t) uhaystack);
+out:
+    free(uneedle);
+    free(uhaystack);
+    return pos;
+}
```
shefty, line 677: Your editor isn't set correctly.  These should use tabs, not spaces, for indentation.  Tabs should show as 8 characters to ensure alignment.

src/shared/ofi_str.c
```diff
@@ -42,61 +42,6 @@
...
-	if(!p)
-		*stringp = NULL;
-	else
-	{
-		*p = 0;
-		*stringp = p + 1;
-	}
-
-	return ptr;
-}
```
shefty, line 65: This function was deleted, not moved, in this commit.

j-xiong (approved):

shefty:
Woo-hoo windows build by trial and error has been successful!  Thanks!

PR #8654, opened 2023-03-14
prov/efa: Implement owner_ops in peer API.

shijin-aws:
Implement all the functions in efa_rdm_srx_owner_ops. Also refactor the code in rxr_pkt_proc_msgrtm() and rxr_pkt_proc_tagrtm() so they can be reused by efa_rdm_srx_owner_ops.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_srx.c
+++ b/prov/efa/src/rdm/efa_rdm_srx.c
@@ -34,6 +34,64 @@
 #include "efa.h"
 #include "efa_rdm_srx.h"
 #include "rxr_msg.h"
+#include "rxr_pkt_type_req.h"
+
+/**
+ * @brief Construct a packet entry that will be used as input of
+ * rxr_pkt_get_msg(tag)rtm_rx_entry in efa_rdm_srx_get_msg(tag).
+ *
+ * @param[in] ep rxr_ep
+ * @param[in] addr the fi_addr_t of the pkt entry
+ * @param[in] size the data size of the pkt entry
+ * @param[in] tag the tag of the pkt entry, ignored unless the op is ofi_op_tagged
+ * @param[in] op the ofi_op code, allowed values: ofi_op_msg and ofi_op_tagged
+ * @param[in,out] pkt_entry the pkt_entry to be constructed.
+ */
+static
+void efa_rdm_srx_construct_pkt_entry(struct rxr_ep *ep,
+				     fi_addr_t addr,
+				     size_t size,
+				     uint64_t tag,
+				     int op,
+				     struct rxr_pkt_entry *pkt_entry)
+{
+	struct rxr_longread_rtm_base_hdr *rtm_hdr;
+	int pkt_type;
+	assert(op == ofi_op_msg || op == ofi_op_tagged);
+
+	/*
+	 * Use longread msg/tag rtm pkt type because it has the
+	 * msg length in the pkt hdr that is used to derive the
+	 * pkt size for the multi-recv path.
+	 * And we cannot make the data size part of pkt_entry->pkt_size
+	 * because we are not allocating memory for it.
+	 */
+	if (op == ofi_op_msg)
+		pkt_type = RXR_LONGREAD_MSGRTM_PKT;
+	else
+		pkt_type = RXR_LONGREAD_TAGRTM_PKT;
+
+	rtm_hdr = (struct rxr_longread_rtm_base_hdr *)pkt_entry->wiredata;
+	rtm_hdr->hdr.type = pkt_type;
+	rtm_hdr->hdr.version = RXR_PROTOCOL_VERSION;
+	rtm_hdr->hdr.flags |= RXR_REQ_MSG;
+	rtm_hdr->msg_length = size;
+
+	if (op == ofi_op_tagged) {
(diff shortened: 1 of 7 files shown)


lrbison (approved):
LGTM

shijin-aws:
Rebased on last main head.

prov/efa/src/rdm/efa_rdm_srx.c
```diff
@@ -42,12 +87,46 @@
...
+
+	rxr_ep = (struct rxr_ep *) srx->ep_fid.fid.context;
+	/*
+	 * TODO:
+	 * In theory we should not need to allocate a pkt_entry to get a rx entry,
+	 * but the current EFA code needs an pkt_entry as input. A major refactor
+	 * is needed to split the context (addr, size etc.) and data from pkt entry
+	 * and make rxr_*_get_*_rx_entry needs context only.
+	 */
+	rx_pkt_entry = rxr_pkt_entry_alloc(rxr_ep, rxr_ep->efa_rx_pkt_pool, RXR_PKT_FROM_EFA_RX_POOL);
```
sunkuamzn, line 109: Should we be concerned about memory exhaustion? `get_msg` will get called for every message received by either the SHM or EFA endpoints, so we are effectively using two `pkt_entry`s for every message. I know we release this `pkt_entry` at the end of this function.

prov/efa/src/rdm/efa_rdm_srx.c
```diff
@@ -42,12 +87,46 @@
...
+	if (OFI_UNLIKELY(!rx_pkt_entry)) {
+		EFA_WARN(FI_LOG_EP_CTRL,
+			"Unable to allocate rx_pkt_entry\n");
+		return -FI_ENOMEM;
+	}
+	efa_rdm_srx_construct_pkt_entry(rxr_ep, addr, size, 0, ofi_op_msg, rx_pkt_entry);
+
+	rx_entry = rxr_pkt_get_msgrtm_rx_entry(rxr_ep, &rx_pkt_entry);
```
sunkuamzn, line 117: I was confused why it's OK to always call `rxr_pkt_get_msgrtm_rx_entry` even though it has `msgrtm` in its name.

But `rxr_pkt_get_msgrtm_rx_entry` is the only function where we search the `rx_list` in `prov/efa`. For other types of messages, we either don't need rx_entry (e.g. handshake packets) or we already have it (e.g. data packets).

prov/efa/src/rdm/efa_rdm_srx.c
shijin-aws: We can have a separate pkt pool for it, like shm rx pkt pool. But we may need to rename it as peer srx pkt pool or something similar. I was hesitant to do it within this PR as I hope to make this pkt pool change in the PR that integrates with shm.

shijin-aws: > get_msg will get called for every message received by either the SHM or EFA endpoints, so we are effectively using two pkt_entrys for every message. so we are effectively using two pkt_entrys for every message.

This is not true, `get_msg` will only be called by shm provider.

shijin-aws: I think I can make it allocated from shm_rx_pool in this PR, but add a TODO to change it. Does that sound OK to you?

sunkuamzn: > This is not true, `get_msg` will only be called by shm provider.

You're right. `start_msg` is called by both EFA and SHM providers but `get_msg` is only called by SHM

> I think I can make it allocated from shm_rx_pool in this PR, but add a TODO to change it. Does that sound OK to you?

I don't have enough experience to say for sure. On instances with high core count, shm pool will be used heavily. Could we make the pools larger?

shijin-aws: This PR will not change any behavior for the current code. In the integration, we will need to make `get_msg` allocate pkt entry from a dedicated pool. 

> On instances with high core count, shm pool will be used heavily. Could we make the pools larger?
I do not think we will use more memory after the integration. This pkt pool is basically a replacement of the current `shm_rx_pool`, and we will drop shm_tx_pool and shm_rx_pool. The memory usage is actually smaller.

sunkuamzn: My concern wasn't the total memory usage but the usage within each pool.

All `efa_*_pool` and `shm_*_pool` are fixed size and cannot grow. I'll let you make the best decision.

https://github.com/ofiwg/libfabric/blob/8bb6f708e175ed9a8845142b715340ccdab75f79/prov/efa/src/rdm/rxr_ep.c#L1137-L1245

shijin-aws: Ah I get your point. Currently efa will allocate a pkt from shm tx pkt pool for every `fi_sendmsg` it posted to shm, and it will return an `EAGAIN` and retry later if the tx pool is drained. I guess we may need to introduce a retry here as well. Let me think about it.

sunkuamzn (approved):

shijin-aws:
AWS CI failure is real, will fix

shijin-aws:
Close this PR, will open a new one with full integration

PR #8655, opened 2023-03-14
prov/shm: srx fixes

aingerson:


prov/shm/src/smr_msg.c
```diff
@@ -130,7 +130,8 @@ static ssize_t smr_generic_mrecv(struct smr_srx_ctx *srx,
 			buf_done = true;
 
 		ofi_spin_unlock(&srx->lock);
-		ret = srx->peer_srx.peer_ops->start_msg(&rx_entry->peer_entry);
+		ret = rx_entry->peer_entry.srx->peer_ops->start_msg(
```
shijin-aws, line 133: Thanks for the fix, this is exactly what we are doing in the efa code.

PR #8657, opened 2023-03-15
prov/efa: exchange rxr endpoint host id information during handshake

wenduwan:
This change introduces a new optional handshake packet field hid(host id) enabled by the HID_HDR flag. The hid value is parsed from a host id file configured in rxr_env.

Notably, hid is optional and provided at best effort, which means:
- It is set to 0 by default.
- If the endpoint fails to parse hid, it will skip the error.
- hid is included in handshake if and only if it is non-zero.

### DIFF
--- a/prov/efa/docs/efa_rdm_protocol_v4.md
+++ b/prov/efa/docs/efa_rdm_protocol_v4.md
@@ -383,6 +383,7 @@ Table: 2.2 binary format of the HANDSHAKE packet
 | `extra_info`  | `8 * (nextra_p3 - 3)` | integer array | `uint64_t[]` |
 | `connid`  | 4 | integer | sender connection ID, optional, present when the CONNID_HDR flag is on `flags` |
 | `padding` | 4 | integer | padding for `connid`, optional, present when the CONNID_HDR flag is on `flags` |
+| `host_id` | 8 | integer | sender host id, optional, present when the HANDSHAKE_HOST_ID_HDR flag is on `flags` (table 2.3) |
 
 The first 4 bytes (3 fields: `type`, `version`, `flags`) is the EFA RDM base header (section 1.3).
 
@@ -429,6 +430,15 @@ to 8 bytes boundary.
 These two fields were introduced with the extra request "connid in header". They are optional,
 therefore an implemenation is not required to set them. (section 4.4 for more details) If an implementation
 does set the connid, the implementation needs to toggle on the CONNID_HDR flag in `flags` (table 1.4).
+
+`connid` and `padding` fields are followed by an optional `host_id` (8 bytes) which is the sender's host id.
+If `connid` and `padding` are not present, `host_id` will follow `extra_info`.
+
+Table: 2.3 A list of handshake packet flags
+
+| Bit Id | Value | Name | Meaning |
+|---|---|---|---|
+|  0     | 0x1    | `HANDSHAKE_HOST_ID_HDR` | This packet has the optional sender host id header |
 
 ### 2.2 Handshake subprotocol and raw address exchange
 
--- a/prov/efa/src/rdm/rdm_proto_v4.h
+++ b/prov/efa/src/rdm/rdm_proto_v4.h
@@ -364,6 +364,9 @@ struct rxr_handshake_hdr {
 	uint64_t extra_info[0];
 };
 
+/* indicate this package has the sender host id */
+#define RXR_HANDSHAKE_HOST_ID_HDR	BIT_ULL(0)
+
 #if defined(static_assert) && defined(__x86_64__)
 static_assert(sizeof(struct rxr_handshake_hdr) == 8, "rxr_handshake_hdr check");
 #endif
@@ -373,8 +376,13 @@ struct rxr_handshake_opt_connid_hdr {
 	uint32_t padding; /* padding to 8 bytes boundary */
 };
 
+struct rxr_handshake_opt_host_id_hdr {
+	uint64_t host_id;
+};
+
 #if defined(static_assert) && defined(__x86_64__)
 static_assert(sizeof(struct rxr_handshake_opt_connid_hdr) == 8, "rxr_handshake_opt_connid_hdr check");
+static_assert(sizeof(struct rxr_handshake_opt_host_id_hdr) == 8, "rxr_handshake_opt_host_id_hdr check");
 #endif
 
 /* @brief header format of RECEIPT packet */
--- a/prov/efa/src/rdm/rxr_env.c
+++ b/prov/efa/src/rdm/rxr_env.c
@@ -68,8 +68,8 @@ struct rxr_env rxr_env = {
 	.efa_read_segment_size = 1073741824,
 	.efa_write_segment_size = 1073741824, /* need to confirm this constant. */
 	.rnr_retry = 3, /* Setting this value to EFA_RNR_INFINITE_RETRY makes the firmware retry indefinitey */
+	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
 };
-
 
 /**
  * @brief Get default value of using device's RDMA
--- a/prov/efa/src/rdm/rxr_env.h
+++ b/prov/efa/src/rdm/rxr_env.h
@@ -81,6 +81,23 @@ struct rxr_env {
 	 *      EFA_RNR_INFINITY_RETRY (retry infinitely)
 	 */
 	int rnr_retry;
+	/**
+	 * The absolute path to a file that contains an EC2 instance id-like string.
+	 * If host_id_file is provided, the program will attempt to read the
+	 * 16 hex characters starting at the 4th character and convert it to an 8-byte
+	 * integer as the host id.
+	 *
+	 * 	Accept example:
+	 *		i-0f7c826e5f3bd8685	->	0xf7c826e5f3bd8685 is a valid hex string
+	 *
+	 * 	Illegal examples,
+	 *		quickfoxbrownfence	->	Not a hex string
+	 *		789e1cab			->	String is too short
+	 *
+	 * Host id parsing is best-effort. If the file does not exist, or the file
+	 * is malformatted, the program should proceed with a default host id, e.g. 0.
+	 */
+	char *host_id_file;
 };
 
 extern struct rxr_env rxr_env;
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -304,7 +304,7 @@ int rxr_ep_post_internal_rx_pkt(struct rxr_ep *ep, uint64_t flags, enum rxr_lowe
 
 	rx_pkt_entry->x_entry = NULL;
 
-	msg_iov.iov_base = (void *)rxr_pkt_start(rx_pkt_entry);
+	msg_iov.iov_base = (void *)(rx_pkt_entry->wiredata);
 	msg_iov.iov_len = ep->mtu_size;
 
 	switch (lower_ep_type) {
@@ -2272,6 +2272,11 @@ int rxr_endpoint(struct fid_domain *domain, struct fi_info *info,
 	if (!rxr_ep->user_info) {
 		ret = -FI_ENOMEM;
 		goto err_free_ep;
+	}
+
+	rxr_ep->host_id = rxr_get_host_id(rxr_env.host_id_file);
+	if (rxr_ep->host_id) {
+		EFA_INFO(FI_LOG_EP_CTRL, "rxr_ep->host_id: i-%017lx\n", rxr_ep->host_id);
 	}
 
 	rxr_ep->rx_size = info->rx_attr->size;
--- a/prov/efa/src/rdm/rxr_ep.h
+++ b/prov/efa/src/rdm/rxr_ep.h
@@ -34,7 +34,10 @@
 #ifndef _EFA_RDM_EP_H
 #define _EFA_RDM_EP_H
 
+#include <ctype.h>
 #include "efa_tp.h"
+
+#define HOST_ID_LENGTH 16
 
 enum ibv_cq_ex_type {
 	IBV_CQ,
@@ -64,6 +67,11 @@ struct rxr_queued_copy {
 
 struct rxr_ep {
 	struct efa_base_ep base_ep;
+
+	/**
+	 * Default to 0
+	 */
+	uint64_t host_id;
 
 	/* per-version extra feature/request flag */
 	uint64_t extra_info[RXR_MAX_NUM_EXINFO];
@@ -280,6 +288,48 @@ static inline size_t rxr_get_tx_pool_chunk_cnt(struct rxr_ep *ep)
 static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
 {
 	return ep->msg_order & FI_ORDER_SAS;
+}
+
+static inline uint64_t rxr_get_host_id(char *host_id_file)
+{
+	FILE *fp = NULL;
+	char host_id_str[HOST_ID_LENGTH + 1];
+	char *end_ptr = NULL;
+	const size_t HOST_ID_PREFIX_LENGTH = 3; /* Ignore the i-0 sequence */
+	size_t length = 0;
+	uint64_t host_id = 0;
+
+	if (!host_id_file) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Host id file is not specified\n");
+		goto out;
+	}
+
+	fp = fopen(host_id_file, "r");
+	if (!fp) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Cannot open host id file: %s\n", host_id_file);
+		goto out;
+	}
+
+	fseek(fp, HOST_ID_PREFIX_LENGTH, SEEK_SET);
+
+	length = fread(host_id_str, 1, HOST_ID_LENGTH, fp);
+	if (length != HOST_ID_LENGTH) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Failed to read host id. Read length: %lu Expect length: %d\n", length, HOST_ID_LENGTH);
+		goto out;
+	}
+	host_id_str[HOST_ID_LENGTH] = '\0';
+
+	host_id = (uint64_t)strtoul(host_id_str, &end_ptr, 16);
+	if (*end_ptr != '\0') {
+		EFA_WARN(FI_LOG_EP_CTRL, "Host id is not a valid hex string: %s\n", host_id_str);
+		return 0;
+	}
+
+out:
+	if (fp) {
+		fclose(fp);
+	}
+	return host_id;
 }
 
 static inline int rxr_ep_use_zcpy_rx(struct rxr_ep *ep, struct fi_info *info)
--- a/prov/efa/src/rdm/rxr_pkt_entry.c
+++ b/prov/efa/src/rdm/rxr_pkt_entry.c
@@ -654,8 +654,7 @@ ssize_t rxr_pkt_entry_inject(struct rxr_ep *ep,
 	assert(peer);
 
 	assert(ep->use_shm_for_tx && peer->is_local);
-	ret = fi_inject(ep->shm_ep, rxr_pkt_start(pkt_entry), pkt_entry->pkt_size,
-			 peer->shm_fiaddr);
+	ret = fi_inject(ep->shm_ep, pkt_entry->wiredata, pkt_entry->pkt_size, peer->shm_fiaddr);
 
 	if (OFI_UNLIKELY(ret))
 		return ret;
--- a/prov/efa/src/rdm/rxr_pkt_type.h
+++ b/prov/efa/src/rdm/rxr_pkt_type.h
@@ -67,6 +67,28 @@ struct rxr_handshake_opt_connid_hdr *rxr_get_handshake_opt_connid_hdr(void *pkt)
 	return (struct rxr_handshake_opt_connid_hdr *)((char *)pkt + base_hdr_size);
 }
 
+static inline
+struct rxr_handshake_opt_host_id_hdr *rxr_get_handshake_opt_host_id_hdr(void *pkt)
+{
+	struct rxr_handshake_hdr *handshake_hdr;
+	size_t offset = 0;
+
+	handshake_hdr = (struct rxr_handshake_hdr *)pkt;
+	assert(handshake_hdr->type == RXR_HANDSHAKE_PKT);
+
+	offset += sizeof(struct rxr_handshake_hdr) +
+					(handshake_hdr->nextra_p3 - 3) * sizeof(uint64_t);
+
+	assert(handshake_hdr->flags & RXR_HANDSHAKE_HOST_ID_HDR);
+
+	if (handshake_hdr->flags & RXR_PKT_CONNID_HDR) {
+		/* HOST_ID_HDR is always immediately after CONNID_HDR(if present) */
+		offset += sizeof(struct rxr_handshake_opt_connid_hdr);
+	}
+
+	return (struct rxr_handshake_opt_host_id_hdr *)((char *)pkt + offset);
+}
+
 ssize_t rxr_pkt_init_handshake(struct rxr_ep *ep,
 			       struct rxr_pkt_entry *pkt_entry,
 			       fi_addr_t addr);
--- a/prov/efa/src/rdm/rxr_pkt_type_misc.c
+++ b/prov/efa/src/rdm/rxr_pkt_type_misc.c
@@ -37,6 +37,7 @@
 #include "rxr_cntr.h"
 #include "rxr_pkt_cmd.h"
 #include "rxr_pkt_type_base.h"
+#include "rxr_pkt_type_misc.h"
 #include "rxr_read.h"
 
 #include "rxr_tp.h"
@@ -50,6 +51,7 @@
  */
 
 /* HANDSHAKE packet related functions */
+
 ssize_t rxr_pkt_init_handshake(struct rxr_ep *ep,
 			       struct rxr_pkt_entry *pkt_entry,
 			       fi_addr_t addr)
@@ -57,6 +59,7 @@ ssize_t rxr_pkt_init_handshake(struct rxr_ep *ep,
 	int nex;
 	struct rxr_handshake_hdr *handshake_hdr;
 	struct rxr_handshake_opt_connid_hdr *connid_hdr;
+	struct rxr_handshake_opt_host_id_hdr *host_id_hdr;
 
 	handshake_hdr = (struct rxr_handshake_hdr *)pkt_entry->wiredata;
 	handshake_hdr->type = RXR_HANDSHAKE_PKT;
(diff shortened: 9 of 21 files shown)


prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +286,78 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
...
+
+	host_id_str = calloc(HOST_ID_LENGTH, sizeof(*host_id_str));
+	parsed_host_id_str = calloc(HOST_ID_LENGTH, sizeof(*parsed_host_id_str));
+	if (!host_id_str || !parsed_host_id_str) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Failed to allocate memory\n");
+		goto out;
+	}
+
+	read_length = fread(host_id_str, sizeof(*host_id_str), HOST_ID_LENGTH, fp);
+	if (read_length != HOST_ID_LENGTH) {
```
lrbison, line 335: This seems like a lot of logic to ensure there are a constant number of hex characters in the file.

Why not just use a 16-byte array, read 16 bytes, and fail if you read less or they are invalid characters?  If you really want to confirm the file is not too long, read one more character and assert it's either non-hex or EOF.

wenduwan: Yes that's my purpose. We want to read the last 16 characters - for that reason we need to find out the end of the first line and walk back 16 chars, then read from there.

> if they are invalid characters
Do you mean we test if the characters are 0-9/a-f? I can do that as well, but I'm not sure if that makes the logic simpler.

prov/efa/src/rdm/rxr_pkt_type_misc.c
```diff
@@ -76,11 +95,21 @@ ssize_t rxr_pkt_init_handshake(struct rxr_ep *ep,
...
+	/**
+	 * Include the optional hid(host id) if it is non-zero
+	 */
+	if (ep->host_id) {
+		hid_hdr = (struct rxr_handshake_opt_hid_hdr *)rxr_pkt_end(pkt_entry);
+		hid_hdr->hid = ep->host_id;
+		handshake_hdr->flags |= RXR_PKT_HID_HDR;
+		pkt_entry->pkt_size += sizeof(struct rxr_handshake_opt_hid_hdr);
+	}
+
```
lrbison, line 112: I think all this logic would be simplified if we simply reported the host_id all the time.  Other side can interpret 0 as "not provided"

wenduwan:
bot:aws:retest

prov/efa/src/rdm/rxr_pkt_type_misc.c
wenduwan: The reason for this check is for API compatibility in case the peer is on an older version - we cannot assume that host id is available.

prov/efa/src/rdm/rxr_ep.h
wenduwan: @lrbison I attempted a simpler version, could you see if it's better or could be made better?

prov/efa/docs/efa_rdm_protocol_v4.md
```diff
@@ -430,6 +431,15 @@ These two fields were introduced with the extra request "connid in header". They
...
 does set the connid, the implementation needs to toggle on the CONNID_HDR flag in `flags` (table 1.4).
 
+`connid` and `padding` fields are followed by an optional `host_id` (8 bytes) which is the sender's host id.
+If `connid` and `padding` are not present, `host_id` will follow `extra_info`.
+
+Table: 2.3 A list of handshake packet flags
+
+| Bit Id | Value | Name | Meaning |
+|---|---|---|---|
+|  0     | 0x1    | `HOST_ID_HDR` | This packet has the optional sender host id header |
```
wzamazon, line 441: This flag is specific to handshake packet, therefore its name should be `HANDSHAKE_HOST_ID_HDR`.

prov/efa/src/rdm/rxr_pkt_type_misc.c
```diff
@@ -76,11 +95,21 @@ ssize_t rxr_pkt_init_handshake(struct rxr_ep *ep,
 	 * Always include connid at the end of a handshake packet.
 	 * If peer cannot make use of connid, the connid will be ignored.
 	 */
-	connid_hdr = (struct rxr_handshake_opt_connid_hdr *)(pkt_entry->wiredata + pkt_entry->pkt_size);
+	connid_hdr = (struct rxr_handshake_opt_connid_hdr *)rxr_pkt_end(pkt_entry);
```
wzamazon, line 98: I do not think use `rxr_pkt_end` is useful. it is better to use `pkt_entry->wiredata + pkt_entry->pkt_size)` because `pkt_entry->pkt_size` keep changing.

prov/efa/src/rdm/rxr_pkt_type_misc.c
```diff
@@ -184,12 +216,16 @@ void rxr_pkt_handle_handshake_recv(struct rxr_ep *ep,
...
 	peer->nextra_p3 = handshake_pkt->nextra_p3;
 	memcpy(peer->extra_info, handshake_pkt->extra_info,
-	       (handshake_pkt->nextra_p3 - 3) * sizeof(uint64_t));
+		   (handshake_pkt->nextra_p3 - 3) * sizeof(uint64_t));
 	peer->flags |= EFA_RDM_PEER_HANDSHAKE_RECEIVED;
-	EFA_DBG(FI_LOG_CQ,
-	       "HANDSHAKE received from %" PRIu64 "\n", pkt_entry->addr);
-	rxr_pkt_entry_release_rx(ep, pkt_entry);
 
+	host_id = rxr_pkt_handshake_host_id_ptr(pkt_entry);
```
wzamazon, line 222: should be `host_id_ptr`

prov/efa/test/efa_unit_test_common.c
```diff
@@ -180,3 +182,61 @@ void efa_unit_test_eager_msgrtm_pkt_construct(struct rxr_pkt_entry *pkt_entry, s
...
+
+/**
+ * @brief Construct RXR_HANDSHAKE_PKT
+ *	The function will include the optional connid/host id headers if and only if
+ *	attr->connid/host id are non-zero.
+ *
+ * @param[in,out] pkt_entry Packet entry. Must be non-NULL.
+ * @param[in] attr Packet attributes.
+ */
+void efa_unit_test_handshake_pkt_construct(struct rxr_pkt_entry *pkt_entry, struct efa_unit_test_handshake_pkt_attr *attr)
```
wzamazon, line 194: Why do we need this function?

Cannot we just use `rxr_pkt_init_handshake`?

prov/efa/src/rdm/rxr_ep.h
wenduwan: @lrbison Thanks for your suggestion. I adopted your solution with minor tweaks. We still need the alloc.

prov/efa/docs/efa_rdm_protocol_v4.md
wenduwan: Thank you. Addressed.

prov/efa/src/rdm/rdm_proto_v4.h
```diff
@@ -364,6 +364,9 @@ struct rxr_handshake_hdr {
 	uint64_t extra_info[0];
 };
 
+/* indicate this package has the sender host id */
+#define RXR_PKT_HOST_ID_HDR			BIT_ULL(0)
```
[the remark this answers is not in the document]
wenduwan: Thank you. Addressed.

prov/efa/src/rdm/rdm_proto_v4.h
```diff
@@ -373,8 +376,13 @@ struct rxr_handshake_opt_connid_hdr {
 	uint32_t padding; /* padding to 8 bytes boundary */
 };
 
+struct rxr_handshake_opt_host_id_hdr {
+	uint64_t host_id; /* Aligns on 8-byte boundary */
```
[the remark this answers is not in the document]
wenduwan: Thank you. Addressed.

prov/efa/src/rdm/rxr_env.c
```diff
@@ -68,9 +68,9 @@ struct rxr_env rxr_env = {
 	.efa_read_segment_size = 1073741824,
 	.efa_write_segment_size = 1073741824, /* need to confirm this constant. */
 	.rnr_retry = 3, /* Setting this value to EFA_RNR_INFINITE_RETRY makes the firmware retry indefinitey */
+	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
```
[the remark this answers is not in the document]
wenduwan: I think it makes sense to store the file path in rxr_env - I'm not making it configurable by the user though.

This makes testing much easier - in unit tests we are able to update rxr_env and point the path to a temporary file.

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +286,72 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
...
+		goto out;
+	}
+
+	fp = fopen(host_id_file, "r");
+	if (!fp) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Cannot open host id file: %s\n", host_id_file);
+		goto out;
+	}
+
+	for (char c = fgetc(fp); c != '\n' && c != EOF; c = fgetc(fp)) {
```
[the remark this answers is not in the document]
wenduwan: I chatted with @lrbison about the general implementation. This particular length calc has been deprecated.

Overall I don't find `fscanf` suitable here, because we want to validate the string is a hex and has expected length. `fread` nicely returns the actual string length while `fscanf` does not.

prov/efa/src/rdm/rxr_pkt_type_misc.c
wenduwan: Actually I find this pretty handy. There is already a `rxr_pkt_start` function and I was able to use `rxr_pkt_end` in a couple places to make the code more concise.

prov/efa/test/efa_unit_test_common.c
wenduwan: We cannot use `rxr_pkt_init_handshake` - that is part of the logic I am testing against so we cannot directly use it.

The current testing flow starts from the receiver of the handshake packet:
1. Local endpoint receives a handshake packet (constructed from this function) from a remote peer.
2. The endpoint processes the packet, updates the peer host id and posts back a handshake to the peer, which calls `rxr_pkt_init_handshake`

The test asserts on 1) the peer host id is updated according to the received handshake packet, and 2) the local endpoint correctly sends back a handshake packet to the peer that optionally includes local host id.

prov/efa/src/rdm/rxr_env.h
```diff
@@ -81,6 +81,23 @@ struct rxr_env {
...
 	 */
 	int rnr_retry;
+	/**
+	 * The absolute path to a file that contains an EC2 instance id-like string.
+	 * If host_id_file is provided, the program will attempt to read the
+	 * 16 hex characters starting at the 4th character and convert it to an 8-byte
+	 * integer as the host id.
+	 *
+	 * 	Accept example:
+	 *		i-0f7c826e5f3bd8685	->	0xf7c826e5f3bd8685 is a valid hex string
```
wckzhang, line 91: Is the 3rd character always 0?

wenduwan: That's what I observed so far.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2195,6 +2195,11 @@ int rxr_endpoint(struct fid_domain *domain, struct fi_info *info,
 		rxr_ep->use_shm_for_tx = false;
 	}
 
+	rxr_ep->host_id = rxr_get_host_id(rxr_env.host_id_file);
+	if (rxr_ep->host_id) {
+		EFA_INFO(FI_LOG_EP_CTRL, "rxr_ep->host_id: i-0%lx\n", rxr_ep->host_id);
```
wzamazon, line 2200: format -> `i-%08lx` would be better

prov/efa/src/rdm/rxr_ep.h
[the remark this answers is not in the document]
wzamazon: Why isn't `fscanf` suited here?

prov/efa/src/rdm/rxr_pkt_type_misc.c
wzamazon: Overall I think the name `rxr_pkt_start/end` is too ambiguous to be useful, because each packet entry has two parts: local info and wire data. The two parts are in two different buffers. So name `rxr_pkt_start/end` does not tell which part it refer to. So if we really need a utility function, the name need to be `rxr_pkt_entry_wiredata_star/end`.

The usage of the function here is even more confusing because usually `xxx_end` was applied on a constant object. but here  the size keep increasing.

prov/efa/src/rdm/rxr_ep.h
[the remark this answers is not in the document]
wenduwan: With `fscanf`, I believe the intention is to read the hex into an integer directly, e.g.
```c
uint64_t id;

ret = fscanf(fp, "i-0%x", &id);
```

Consider the case
```
i-01234 // invalid input
```
vs
```
i-00000000000001234 // valid input
```

With `fscanf` we cannot tell apart those 2. On the other hand `fread` returns the number of bytes read from the file and we know the first input is invalid.

prov/efa/src/rdm/rxr_ep.c
wenduwan: I think you meant `i-0%8lx`. Addressed.

prov/efa/src/rdm/rxr_pkt_type_misc.c
wenduwan: Thanks for clarifying. I do see convenience of these utilities. Renamed the function as suggested.

prov/efa/src/rdm/rxr_ep.c
wzamazon: No. 0 after `%` means fill empty with `0` instead of space.

and sorry it should be `%016lx`

prov/efa/src/rdm/rxr_pkt_type_misc.h
```diff
@@ -0,0 +1,41 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _RXR_PKT_TYPE_MISC_H
+#define _RXR_PKT_TYPE_MISC_H
+
+#include "rxr.h"
+
+uint64_t *rxr_pkt_handshake_host_id_ptr(struct rxr_pkt_entry *pkt_entry);
```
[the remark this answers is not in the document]
wzamazon: Can you find another way to test?

I would prefer define this function as `static inline`, which has performance benefit

prov/efa/src/rdm/rxr_ep.c
wenduwan: Instance id is formatted this way `i-0[16 char hex string]` - total length 19 characters. I think the correct format is `i-0%016lx`

prov/efa/src/rdm/rxr_pkt_type_misc.h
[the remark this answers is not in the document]
wenduwan: Thanks for suggestion. Addressed.

prov/efa/src/rdm/rxr_ep.c
wenduwan: Addressed. `%017lx`

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -295,7 +295,7 @@ int rxr_ep_post_internal_rx_pkt(struct rxr_ep *ep, uint64_t flags, enum rxr_lowe
 
 	rx_pkt_entry->x_entry = NULL;
 
-	msg_iov.iov_base = (void *)rxr_pkt_start(rx_pkt_entry);
+	msg_iov.iov_base = (void *)rxr_pkt_entry_wiredata_start(rx_pkt_entry);
```
wzamazon, line 298: Why not just `rx_pkt_entry->wiredata`?

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +285,57 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
...
+
+	fp = fopen(host_id_file, "r");
+	if (!fp) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Cannot open host id file: %s\n", host_id_file);
+		goto out;
+	}
+
+	fseek(fp, HOST_ID_PREFIX_LENGTH, SEEK_SET);
+
+	host_id_str = calloc(HOST_ID_LENGTH, sizeof(*host_id_str));
```
wzamazon, line 310: No need to `calloc`. You can define a `char host_id_str[HOST_ID_LENGTH];`

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +285,57 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
 	return ep->msg_order & FI_ORDER_SAS;
 }
 
+static inline uint64_t rxr_get_host_id(char *host_id_file)
+{
+	FILE *fp = NULL;
+	char *host_id_str = NULL;
+	const size_t HOST_ID_LENGTH = 16;
```
wzamazon, line 292: should be 17 to include ending `\0`.

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +285,57 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
...
+		goto out;
+	}
+
+	length = fread(host_id_str, 1, HOST_ID_LENGTH, fp);
+	if (length != HOST_ID_LENGTH) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Failed to read host id. Read length: %lu Expect length: %lu\n", length, HOST_ID_LENGTH);
+		goto out;
+	}
+
+	for (char *s = host_id_str; *s; ++s) {
```
wzamazon, line 322: No need to do validation by yourself. `strtoul` can do it for you. Just pass a pointer to `strtoul` instead of NULL. and make sure the pointer is pointing to the end of the string

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +285,57 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
...
+
+	fseek(fp, HOST_ID_PREFIX_LENGTH, SEEK_SET);
+
+	host_id_str = calloc(HOST_ID_LENGTH, sizeof(*host_id_str));
+	if (!host_id_str) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Cannot allocate memory\n");
+		goto out;
+	}
+
+	length = fread(host_id_str, 1, HOST_ID_LENGTH, fp);
```
wzamazon, line 316: `fgets` is better.

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -654,7 +654,7 @@ ssize_t rxr_pkt_entry_inject(struct rxr_ep *ep,
 	assert(peer);
 
 	assert(ep->use_shm_for_tx && peer->is_local);
-	ret = fi_inject(ep->shm_ep, rxr_pkt_start(pkt_entry), pkt_entry->pkt_size,
+	ret = fi_inject(ep->shm_ep, rxr_pkt_entry_wiredata_start(pkt_entry), pkt_entry->pkt_size,
```
wzamazon, line 657: just use `pkt_entry->wiredata`

prov/efa/test/efa_unit_test_common.c
```diff
@@ -180,3 +182,61 @@ void efa_unit_test_eager_msgrtm_pkt_construct(struct rxr_pkt_entry *pkt_entry, s
...
+	calloc((uintptr_t)handshake_hdr->extra_info, nex * sizeof(uint64_t));
+	pkt_entry->pkt_size = sizeof(struct rxr_handshake_hdr) + nex * sizeof(uint64_t);
+	memcpy(pkt_entry->wiredata, handshake_hdr, sizeof(struct rxr_handshake_hdr));
+	assert_int_equal(rxr_get_base_hdr(pkt_entry->wiredata)->type, RXR_HANDSHAKE_PKT);
+
+	if (attr->connid) {
+		struct rxr_handshake_opt_connid_hdr opt_connid_hdr = {0};
+		opt_connid_hdr.connid = attr->connid;
+		handshake_hdr->flags |= RXR_PKT_CONNID_HDR;
+		memcpy(rxr_pkt_entry_wiredata_end(pkt_entry), &opt_connid_hdr, sizeof(struct rxr_handshake_opt_connid_hdr));
```
wzamazon, line 214: I do not think use `rxr_pkt_entry_wiredata_end` is good idea. It indirectly refer to `pkt_entry->pkt_size`, which was updated on the fly.

prov/efa/test/efa_unit_test_common.c
wenduwan: Could you elaborate you concerns?

I actually think this is where unit test shines. With mocks we can verify fine-grained aspects of code behavior. Especially for host id - which is a private endpoint attribute underneath the `fi_*` APIs - we can test its value at different points of endpoint lifecycle.

I'm not sure how we would test the same aspects in integration test - do we need to inspect the endpoint memory structure, or filter EFA packet to find the host id? The only quick way I can think of is through logging - but that is not a good approach. Please let me know.

prov/efa/src/rdm/rxr_ep.c
wenduwan: Thank you. Given your feedback, I have removed `rxr_pkt_start` and `rxr_pkt_end` entirely.

prov/efa/src/rdm/rxr_ep.h
wenduwan: Thanks. Addressed.

prov/efa/src/rdm/rxr_ep.h
wenduwan: Actually 16 also works. In this case null char is dropped silently at the end. Also it does not really matter for this function whether the string is null terminated, since we specified read length below.

prov/efa/src/rdm/rxr_ep.h
wenduwan: Thanks! Addressed.

prov/efa/src/rdm/rxr_ep.h
wenduwan: I double checked the man pages for fread vs fgets but I still think fread is better in this case, since it allows specifying length to read and returns the actual read length. This is very handy for checking the string format.

prov/efa/src/rdm/rxr_ep.h
wzamazon: but `strtoul` require the input to be `\0` terminated string.

prov/efa/test/efa_unit_test_common.c
wenduwan: Addressed.

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +285,48 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
 	return ep->msg_order & FI_ORDER_SAS;
 }
 
+static inline uint64_t rxr_get_host_id(char *host_id_file)
+{
+	FILE *fp = NULL;
+	const size_t HOST_ID_LENGTH = 16;
+	char host_id_str[HOST_ID_LENGTH];
```
lrbison, line 292: Need a +1 for trailing \0 so strtoul doesn't overrun

prov/efa/src/rdm/rxr_ep.h
wzamazon: `fgets` handles string better. For example, `fgets` will add `\0` to the end.

wenduwan: Thanks I manually added `\0` to the end.

prov/efa/src/rdm/rxr_ep.h
wenduwan: Thanks for being insistent. I just realized that this is a bug and undefined behavior. So undefined that all my unit test cases passed.

prov/efa/src/rdm/rxr_ep.h
wenduwan: Thanks. I realized that this is a bug. Fixed now.

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -279,6 +285,49 @@ static inline int rxr_need_sas_ordering(struct rxr_ep *ep)
...
+	fp = fopen(host_id_file, "r");
+	if (!fp) {
+		EFA_WARN(FI_LOG_EP_CTRL, "Cannot open host id file: %s\n", host_id_file);
+		goto out;
+	}
+
+	fseek(fp, HOST_ID_PREFIX_LENGTH, SEEK_SET);
+
+	length = fread(host_id_str, 1, HOST_ID_LENGTH, fp);
+	host_id_str[HOST_ID_LENGTH + 1] = '\0';
```
wzamazon, line 312: should be `host_id_str[HOST_ID_LENGTH] = '\0'`

wzamazon: Also, I prefer move this line to after L316

wenduwan: Thank you! Fixed.

prov/efa/test/efa_unit_test_common.c
wzamazon: I see. thanks!

wzamazon (approved):

prov/efa/src/rdm/rxr_pkt_type.h
```diff
@@ -67,6 +67,28 @@ struct rxr_handshake_opt_connid_hdr *rxr_get_handshake_opt_connid_hdr(void *pkt)
...
+	handshake_hdr = (struct rxr_handshake_hdr *)pkt;
+	assert(handshake_hdr->type == RXR_HANDSHAKE_PKT);
+
+	offset += sizeof(struct rxr_handshake_hdr) +
+					(handshake_hdr->nextra_p3 - 3) * sizeof(uint64_t);
+
+	assert(handshake_hdr->flags & RXR_HANDSHAKE_HOST_ID_HDR);
+
+	if (handshake_hdr->flags & RXR_PKT_CONNID_HDR) {
+		/* HOST_ID_HDR is always immediately after CONNID_HDR(if present) */
```
shijin-aws, line 85: This logic confuses me, we won't attach host id to pkt hdr without connid hdr. Is that for filtering old version of libfabric?

prov/efa/src/rdm/rxr_pkt_type_misc.c
```diff
@@ -184,12 +200,16 @@ void rxr_pkt_handle_handshake_recv(struct rxr_ep *ep,
 	 */
 	peer->nextra_p3 = handshake_pkt->nextra_p3;
 	memcpy(peer->extra_info, handshake_pkt->extra_info,
-	       (handshake_pkt->nextra_p3 - 3) * sizeof(uint64_t));
+		   (handshake_pkt->nextra_p3 - 3) * sizeof(uint64_t));
```
shijin-aws, line 203: why changing the alignment here?

prov/efa/src/rdm/rxr_pkt_type.h
wenduwan: Yes
> The reason for this check is for API compatibility in case the peer is on an older version - we cannot assume that host id is available.

prov/efa/src/rdm/rxr_pkt_type_misc.c
wenduwan: I reformatted this function. I appears git shows a different tab size.

prov/efa/src/rdm/rxr_pkt_type.h
shijin-aws: OK, makes sense.

shijin-aws (approved):

wzamazon:
@wenduwan  Can you address the CI failure? it is compilation on windows.

wenduwan:
> @wenduwan Can you address the CI failure? it is compilation on windows.

Thanks for reminder! Fixed.

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -34,8 +34,11 @@
 #ifndef _EFA_RDM_EP_H
 #define _EFA_RDM_EP_H
 
+#include <ctype.h>
 #include "efa_tp.h"
 
+#define HOST_ID_LENGTH 16
```
wenduwan, line 40: MSVC requires constant length array. Introduced this macro.

wzamazon (approved):

PR #8660, opened 2023-03-15
prov/tcp: Suspicion of bug where RX entry is freed twice

sydidelot:
When xnet_op_read_rsp() is executed, ep->cur_rx.entry is set to the head element of the list 'rma_read_queue'. The RX entry is effectively removed from the list only after the operation completes. This pattern becomes problematic if the EP is disabled before the RX completes (i.e., EP disabled because of a failing TX entry or an explicit shutdown). In this case, xnet_ep_flush_all_queues would complete the same RX entry twice: once when rma_read_queue is flushed (the RX entry is still listed here) and another time in the block for the condition "if (ep->cur_rx.entry)".

The patch removes the RX entry from rma_read_queue before it gets assigned to ep->cur_rx.entry. It guarantees that the RX entry will be completed only once if the EP is disabled.

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -752,14 +752,13 @@ static int xnet_op_write(struct xnet_ep *ep)
 static int xnet_op_read_rsp(struct xnet_ep *ep)
 {
 	struct xnet_xfer_entry *rx_entry;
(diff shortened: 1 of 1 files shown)


shefty (approved):
thanks - change removes a check in that fast path too

PR #8558, opened 2023-02-22
prov/efa: Add EFA domain specific unit tests

vidsouza:
Add unit tests to check the behaviour of the environment variables FI_EFA_USE_DEVICE_RDMA and FI_HMEM_CUDA_ENABLE_XFER when EFA domain is being initialized.

PR #8661, opened 2023-03-15
prov/efa: set use_device_rdma with fi_setopt

lrbison:
move the use_device_rdma option from the domain to the endpoint, and use fi_setopt to control it rather than an environment variable.

This change does not modify the behavior for applications that use API version <1.18.  For these API versions we query the environment value during domain creation, which is (hopefully) the same value we queried during getinfo.

For applications that use API version >= 1.18, we still allow the user to disable RDMA via the environment (FI_EFA_USE_DEVICE_RDMA=0/1).  if fi_setopt() tries to enable RDMA but user has disabled it, then fi_setopt will return `-FI_EINVAL`.  Likewise if the application tries to disable it when the environment has it disabled.

Motivation: https://github.com/ofiwg/libfabric/pull/8529

### DIFF
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -70,9 +70,10 @@ extern "C" {
 /* negative options are provider specific */
 enum {
        FI_OPT_EFA_RNR_RETRY = -FI_PROV_SPECIFIC_EFA,
-       FI_OPT_EFA_EMULATED_READ,       /* bool */
-       FI_OPT_EFA_EMULATED_WRITE,      /* bool */
-       FI_OPT_EFA_EMULATED_ATOMICS,    /* bool */
+       FI_OPT_EFA_EMULATED_READ,	/* bool */
+       FI_OPT_EFA_EMULATED_WRITE,	/* bool */
+       FI_OPT_EFA_EMULATED_ATOMICS,	/* bool */
+       FI_OPT_EFA_USE_DEVICE_RDMA,	/* bool */
 };
 
 struct fi_fid_export {
--- a/man/fi_efa.7.md
+++ b/man/fi_efa.7.md
@@ -103,6 +103,22 @@ provider for AWS Neuron or Habana SynapseAI.
   emulating Read, Write, and Atomic operations (return value is true), or if
   these operations are assisted by hardware support (return value is false).
 
+*FI_OPT_EFA_USE_DEVICE_RDMA - bool*
+: Only available if the application selects a libfabric API version >= 1.18.
+  This option allows an application to change libfabric's behavior
+  with respect to RDMA transfers.  Note that there is also an environment
+  variable FI_EFA_USE_DEVICE_RDMA which the user may set as well.  If the
+  environment variable and the argument provided with this variable are in
+  conflict, then fi_setopt will return -FI_EINVAL, and the environment variable
+  will be respected.  If the hardware does not support RDMA and the argument
+  is true, then fi_setopt will return -FI_EOPNOTSUPP.  If the application uses
+  API version < 1.18, the argument is ignored and fi_setopt returns
+  -FI_ENOPROTOOPT.
+  The default behavior for RDMA transfers depends on API version.  For
+  API >= 1.18 RDMA is enabled by default on any hardware which supports it.
+  For API<1.18, RDMA is enabled by default only on certain newer hardware
+  revisions.
+
 # RUNTIME PARAMETERS
 
 *FI_EFA_TX_SIZE*
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -210,9 +210,6 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		return -FI_EINVAL;
 	}
 
-	/* Check the value of environment variable FI_EFA_USE_DEVICE_RDMA */
-	efa_domain->use_device_rdma = rxr_env_get_use_device_rdma();
-	
 	efa_domain->mr_local = ofi_mr_local(info);
 	if (EFA_EP_TYPE_IS_DGRAM(info) && !efa_domain->mr_local) {
 		EFA_WARN(FI_LOG_EP_DATA, "dgram require FI_MR_LOCAL, but application does not support it\n");
--- a/prov/efa/src/efa_user_info.c
+++ b/prov/efa/src/efa_user_info.c
@@ -278,45 +278,84 @@ err_free:
  */
 bool efa_user_info_should_support_hmem(int version)
 {
+	bool any_hmem, rdma_allowed;
+	char *extra_info = "";
+	int hmem_type;
 
 	/* Note that the default behavior of EFA provider is different between
-	 * libfabric API version.
+	 * libfabric API version when CUDA is used as HMEM system.
 	 *
 	 * For libfabric API version 1.17 and earlier, EFA provider does not
-	 * support FI_HMEM unless GPUDirect RDMA is available.
+	 * support FI_HMEM on CUDA unless GPUDirect RDMA is available.
 	 *
 	 * For libfabric API version 1.18 and later, EFA provider will claim support
-	 * of FI_HMEM as long as CUDA library is initialized and a CUDA device is
+	 * of FI_HMEM on CUDA as long as CUDA library is initialized and a CUDA device is
 	 * available. On an system without GPUDirect RDMA, the support of CUDA memory
 	 * is implemented by calling CUDA library. If user does not want EFA provider
 	 * to use CUDA library, the user can call use fi_setopt to set
 	 * FI_OPT_CUDA_API_PERMITTED to false.
 	 * On an system without GPUDirect RDMA, such a call would fail.
+	 *
+	 * For NEURON and SYNAPSEAI HMEM types, use_device_rdma is required no
+	 * matter the API version, as is P2P support.
 	 */
-	if (FI_VERSION_GE(version, FI_VERSION(1, 18)) && hmem_ops[FI_HMEM_CUDA].initialized) {
+	if (hmem_ops[FI_HMEM_CUDA].initialized && FI_VERSION_GE(version, FI_VERSION(1, 18))) {
 		EFA_INFO(FI_LOG_CORE,
-			"User is using API version >= 1.18. CUDA library and device are availble, claim support of FI_HMEM.");
-		return true;
+			"User is using API version >= 1.18. CUDA library and "
+			"device are available, claim support of FI_HMEM.");
+			/* For this API we can support HMEM regardless of
+			   use_device_rdma and P2P support, because we can use
+			   CUDA api calls.*/
+			return true;
 	}
 
+	if (hmem_ops[FI_HMEM_CUDA].initialized) {
+		extra_info = "For CUDA and libfabric API version <1.18 ";
+	}
+
+	any_hmem = false;
+	EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(hmem_type) {
+		/* Note that .initialized doesn't necessarily indicate there are
+		   hardware devices available, only that the libraries are
+		   available. */
+		if (hmem_ops[hmem_type].initialized) {
+			any_hmem = true;
+		}
+	}
+
+	if (!any_hmem) {
+		EFA_WARN(FI_LOG_CORE,
+			"FI_HMEM cannot be supported because no compatible "
+			"libraries were found.");
+		return false;
+	}
+
+	/* All HMEM providers require P2P support. */
 	if (ofi_hmem_p2p_disabled()) {
 		EFA_WARN(FI_LOG_CORE,
-			"FI_HMEM capability currently requires peer to peer "
-			"support, which is disabled because FI_HMEM_P2P_DISABLED was set to 1/on/true.");
+			"%sFI_HMEM capability requires peer to peer "
+			"support, which is disabled because "
+			"FI_HMEM_P2P_DISABLED was set to 1/on/true.",
+			extra_info);
 		return false;
 	}
 
+	/* all devices require use_device_rdma. */
 	if (!efa_device_support_rdma_read()) {
 		EFA_WARN(FI_LOG_CORE,
-			"FI_HMEM capability requires RDMA, which this device does not support.");
+		"%sEFA cannot support FI_HMEM because the EFA device "
+		"does not have RDMA support.\n",
+		extra_info);
 		return false;
 	}
 
-	if (!rxr_env_get_use_device_rdma()) {
+	rdma_allowed = efa_rdm_get_use_device_rdma(version);
+	/* not allowed to use rdma, but device supports it... */
+	if (!rdma_allowed) {
 		EFA_WARN(FI_LOG_CORE,
-			"FI_HMEM capability requires RDMA, which is turned off. "
-			"You can turn it on by setting environment variable "
-			"FI_EFA_USE_DEVICE_RDMA to 1.\n");
+		"%sEFA cannot support FI_HMEM because the environment "
+		"variable FI_EFA_USE_DEVICE_RDMA is 0.\n",
+		extra_info);
 		return false;
 	}
 
@@ -327,6 +366,9 @@ bool efa_user_info_should_support_hmem(int version)
 
 bool efa_user_info_should_support_hmem(int version)
 {
+	EFA_WARN(FI_LOG_CORE,
+		"EFA cannot support FI_HMEM because it was not compiled "
+		"with any supporting FI_HMEM capabilities\n");
 	return false;
 }
 
--- a/prov/efa/src/rdm/efa_rdm_peer.h
+++ b/prov/efa/src/rdm/efa_rdm_peer.h
@@ -152,7 +152,7 @@ bool efa_rdm_peer_support_delivery_complete(struct efa_rdm_peer *peer)
 static inline
 bool efa_both_support_rdma_read(struct rxr_ep *ep, struct efa_rdm_peer *peer)
 {
-	return efa_domain_support_rdma_read(rxr_ep_domain(ep)) &&
+	return efa_rdm_ep_support_rdma_read(ep) &&
 	       (peer->is_self || efa_rdm_peer_support_rdma_read(peer));
 }
 
@@ -168,7 +168,7 @@ bool efa_both_support_rdma_read(struct rxr_ep *ep, struct efa_rdm_peer *peer)
 static inline
 bool efa_both_support_rdma_write(struct rxr_ep *ep, struct efa_rdm_peer *peer)
 {
-	return efa_domain_support_rdma_write(rxr_ep_domain(ep)) &&
+	return efa_rdm_ep_support_rdma_write(ep) &&
 	       (peer->is_self || efa_rdm_peer_support_rdma_write(peer));
 }
 
--- a/prov/efa/src/rdm/efa_rdm_util.c
+++ b/prov/efa/src/rdm/efa_rdm_util.c
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) Amazon.com, Inc. or its affiliates.
+ * All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "efa_rdm_util.h"
+
+/**
+ * @brief Fetch the initial value for use_device_rdma
+ *
+ * This function fetches the value of FI_EFA_USE_DEVICE_RDMA from the
+ * user's environment.  If not set, it uses the API version and the EFA
+ * hardware's version and capabilities to decide when to enable or
+ * disable use_device_rdma during endpoint initialization.
+ *
+ * This value can be modified per-endpoint by using fi_setopt if the
+ * application uses API>=1.18.
+ *
+ * The application may abort() during this method for two reasons:
+ *  - If the Environment variable is set with non-boolean-like value
+ *  - If the Environment variable requests RDMA but no hardware support
+ *    is available.
+ *
+ * @return	bool: use_device_rdma default or environment setting.
+ */
+bool efa_rdm_get_use_device_rdma(uint32_t fabric_api_version)
+{
+	int ret;
+	int param_val;
+	bool hw_support;
+	bool default_val;
+	uint32_t vendor_part_id;
+
+	vendor_part_id = g_device_list[0].ibv_attr.vendor_part_id;
+	hw_support = efa_device_support_rdma_read();
+
+	if (FI_VERSION_GE(fabric_api_version, FI_VERSION(1,18))) {
+		default_val = hw_support;
+	} else {
+		if (vendor_part_id == 0xefa0 || vendor_part_id == 0xefa1) {
+			default_val = false;
+		} else {
+			default_val = true;
+		}
+
+		if (default_val && !hw_support) {
+			fprintf(stderr, "EFA device with vendor id %x unexpectedly has "
+				"no RDMA support. Application will abort().\n",
+				vendor_part_id);
+			abort();
+		}
+	}
+	param_val = default_val;
+
+	/* Fetch the value of environment variable set by the user if any. */
+	ret = fi_param_get_bool(&efa_prov, "use_device_rdma", &param_val);
+	if (ret == -EINVAL) {
+		fprintf(stderr, "FI_EFA_USE_DEVICE_RDMA was set to an invalid value by the user."
+			" FI_EFA_USE_DEVICE_RDMA is boolean and can be set to only 0/false/no/off or"
+			" 1/true/yes/on.  Application will abort().\n");
+		abort();
+	}
+	if (ret < 0) return default_val;
+
+	/* When the user requests use device RDMA but the device does not
+	   support RDMA, exit the run. */
+	if (param_val && !hw_support) {
+		fprintf(stderr, "FI_EFA_USE_DEVICE_RDMA=1 was set by user, but "
+			"EFA device has no rdma-read capability.  Application "
+			"will abort().\n");
+		abort();
+	}
+
+	return param_val;
+}
--- a/prov/efa/src/rdm/efa_rdm_util.h
+++ b/prov/efa/src/rdm/efa_rdm_util.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) Amazon.com, Inc. or its affiliates.
+ * All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _EFA_RDM_UTIL_H
+#define _EFA_RDM_UTIL_H
+
+#include "efa.h"
+
+bool efa_rdm_get_use_device_rdma(uint32_t fabric_api_version);
+
+#endif /* _EFA_RDM_UTIL_H */
--- a/prov/efa/src/rdm/rxr_env.c
+++ b/prov/efa/src/rdm/rxr_env.c
@@ -72,43 +72,6 @@ struct rxr_env rxr_env = {
 };
 
 /**
- * @brief Get default value of using device's RDMA
- *
- * This function gets the default value of using device's RDMA
- * capability. This function uses the g_device_list initialized
- * in efa_device_list_initialize(), thus it must be called
- * after efa_device_list_initialize().
- *
- * @return 	1 - Use EFA device's RDMA capability by default
- * 		0 - Don't use EFA device's RDMA capability by default
- */
-static int rxr_env_get_default_use_device_rdma()
-{
-	int use_device_rdma;
-	uint32_t vendor_part_id = g_device_list[0].ibv_attr.vendor_part_id;
-
-	/* This is the default scenario. This could be overwritten
-	 * later by the user when fi_param_get_int() is called.
-	 * By default we would not want to use device RDMA in EFA
-	 * gen-0 and gen-1 because it wasn't shipped with RDMA
-	 * capability. Starting from EFA gen-2, if the device
-	 * supports RDMA, use it. */
-	if (vendor_part_id == 0xefa0 || vendor_part_id == 0xefa1) {
-		use_device_rdma = 0;
-	} else {
-		use_device_rdma = 1;
-		if (!(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
-			EFA_WARN(FI_LOG_FABRIC,
-				"EFA device with vendor_part_id 0x%x has no rdma-read capability.\n", vendor_part_id);
-			use_device_rdma = 0;
-		}
-	}
-
-	return use_device_rdma;
-}
-
-
-/**
  * @brief Define FI_EFA_USE_DEVICE_RDMA as a configuration parameter
  *
  * This function fetches the default value of using EFA device's
@@ -117,18 +80,18 @@ static int rxr_env_get_default_use_device_rdma()
 void rxr_env_define_use_device_rdma()
 {
 	char *str = "";
-	int use_device_rdma;
-
-	/* Get the default value of using device's RDMA capability */
-	use_device_rdma = rxr_env_get_default_use_device_rdma();
 
 	/* Specify the help info about the usage of RDMA in the device. */
-	if (!(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ))
-		str = "EFA device on your system does not support RDMA, so this variable cannot be set to 1";
-	fi_param_define(&efa_prov, "use_device_rdma", FI_PARAM_BOOL,
-			"Specifies whether to use device's RDMA functionality for one-sided and two-sided transfer. (Default: %d). %s", use_device_rdma, str);
-}
+	if (!efa_device_support_rdma_read()) {
+		str = "  EFA device on your system does not support RDMA,"
+			" so this variable cannot be set to 1.";
+	}
 
+	fi_param_define(&efa_prov, "use_device_rdma", FI_PARAM_BOOL,
+			"Specifies whether to use device's RDMA functionality"
+			" for one-sided and two-sided transfers.%s",
+			str);
+}
 
 /* @brief Read and store the FI_EFA_* environment variables.
  */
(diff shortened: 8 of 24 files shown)


lrbison:
Sorry @shijin-aws, fixed!

lrbison:
I will work to add unit tests to this PR.  I still welcome any feedback in the meantime.

bwbarrett:
I understand not using setopt() before 1.18, but I'm not sure I understand ignoring the env variable.  The env variable is a safety valve (similar to the other cross-over point settings, and it seems like you're closing that safety valve.  I don't have strong feelings on env var vs. programmatic API priority when there are conflicts, but it seems weird to ignore the env var going forward.

lrbison:
> I understand not using setopt() before 1.18, but I'm not sure I understand ignoring the env variable. The env variable is a safety valve (similar to the other cross-over point settings, and it seems like you're closing that safety valve. I don't have strong feelings on env var vs. programmatic API priority when there are conflicts, but it seems weird to ignore the env var going forward.

@bwbarrett 
Yes.  This was something we meant to discuss this week but didn't have time yesterday.  My reasoning is that part of the motivation for this change was to avoid segfaults when one thread is calling setenv() while libfabric is calling getenv().  I cannot prevent other threads from calling setenv() (although, now they won't need to) so if I call getenv() I'm risking the same segfault.  On the other hand it leaves the potential for conflicting configuration information to be silently ignored, and as you point out, removes a human-user knob from their toolbox.

My understanding was that the FI_EFA_USE_DEVICE_RDMA variable was primarily used by applications wishing to control libfabric behavior rather than users themselves, but it sounds like you think there are valid use-cases for users as well?

If we believe the risk of segfault will be removed because the appliation will no longer be calling setenv(), then I can happily check for conflicting

vidsouza:
@wzam, if Luke is working on writing a unit test for this, should we delete #8558 ?

lrbison:
@vidsouza Thank you for the pointer.  I should probably start with that PR and make some updates to handle the new API version cases.

bwbarrett:
Libfabric already calls getenv() so many times, that the important change is to avoid setenv(), not to avoid getenv().  Today, plenty of users set the env var (because we tell them to in our public docs).  My point is that, similar to the message size cross-over points, having an env var to override what the app wants or the default as a safety valve isn't a bad thing.

wzamazon:
> @wzam, if Luke is working on writing a unit test for this, should we close https://github.com/ofiwg/libfabric/pull/8558 ?
Yes

lrbison:
@bwbarrett @wzamazon 

In order to provide a useful tuning knob to the user, we must allow conflicting values for the environment variable and the setopt value without an abort().  I started thinking we would just have the environment variable always override the setopt, but realized this might be strange for when the application is trying to disable RDMA and the user is requesting it on.

As a result, I think the guiding principle should be conflicting settings should only turn off RDMA, with of course ample logging.

Do you agree?  Would it ever make sense to let the user force RDMA on even if application requests it off?

bwbarrett:
Unlike enabling, I'm not sure that there's any reason that a code would want to disable rdma with setopt.  But even if they did, I think it's more important to have a consistent behavior (not just across this option, but all the EFA provider options if not all the Libfabric options).  Don't over think this; the goal should be consistency, not handling every unlikely edge case.

lrbison:
I've updated the PR description comment to reflect the new behavior, namely:

> For applications that use API version >= 1.18, we still allow the user to disable RDMA via the environment (FI_EFA_USE_DEVICE_RDMA=0). if fi_setopt() tries to enable RDMA but user has disabled it, then fi_setopt will return -FI_EINVAL

lrbison:
@bwbarrett environment setting will always be respected now.  This is consistent with FI_HMEM_P2P_DISABLED which can be controlled either by environment variable or by fi_setopt on the endpoint.

prov/efa/src/efa_domain.c
```diff
@@ -210,9 +211,14 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		return -FI_EINVAL;
 	}
 
-	/* Check the value of environment variable FI_EFA_USE_DEVICE_RDMA */
-	efa_domain->use_device_rdma = rxr_env_get_use_device_rdma();
```
wzamazon, line 214: I think we should move`use_device_rdma` to inside `rxr_ep`.

prov/efa/src/efa_user_info.c
```diff
@@ -302,28 +304,42 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
 			"FI_HMEM capability currently requires peer to peer "
-			"support, which is disabled because FI_HMEM_P2P_DISABLED was set to 1/on/true.");
+			"support, which is disabled because "
+			"FI_HMEM_P2P_DISABLED was set to 1/on/true.");
 		return false;
 	}
 
-	if (!efa_device_support_rdma_read()) {
+	if (!FI_VERSION_GE(version, FI_VERSION(1, 18))) {
+		extra_version_info = " the application is using libfabric "
```
wzamazon, line 321: it seems we just need `extra_info`, and we can make it point to the two different message

prov/efa/src/efa_user_info.c
```diff
@@ -302,28 +304,42 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+			"API <1.18, so CUDA cannot be used.";
+	}
```
wzamazon, line 323: It seems that the structure can be 

```
if (!FI_VERSION_GE(version, FI_VERSION(1, 18))) {
   extra_info = "xxx"
} else {
     assert(!hmem_ops[FI_HMEM_CUDA].initialized);
   extra_info = "xxx"
}
```

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -151,7 +151,7 @@ bool efa_rdm_peer_support_delivery_complete(struct efa_rdm_peer *peer)
 static inline
 bool efa_both_support_rdma_read(struct rxr_ep *ep, struct efa_rdm_peer *peer)
 {
-	return efa_domain_support_rdma_read(rxr_ep_domain(ep)) &&
+	return efa_endpoint_support_rdma_read(ep) &&
```
wzamazon, line 154: the function name should be `efa_rdm_ep_support_rdma_read`

prov/efa/src/rdm/rxr_env.c
```diff
@@ -281,39 +248,40 @@ void rxr_env_initialize()
...
 		abort();
 	}
 
-	/* When the default value was to not use it, but the user sets
-	 * the value to use device RDMA in a device that does not support
-	 * RDMA, exit the run */
-	if (ret != -FI_ENODATA && use_device_rdma &&
+	/* When the user requests use device RDMA but the device does not
+	   support RDMA, exit the run. */
+	if (ret == 1 &&
```
wzamazon, line 280: This does not look right. I think you want to check `param_val`. 

`ret` is the return value of the function

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2283,6 +2355,17 @@ int rxr_endpoint(struct fid_domain *domain, struct fi_info *info,
 	rxr_ep->efa_max_outstanding_rx_ops = efa_domain->device->rdm_info->rx_attr->size;
 	rxr_ep->efa_device_iov_limit = efa_domain->device->rdm_info->tx_attr->iov_limit;
 
+	api_version = efa_domain->util_domain.fabric->fabric_fid.api_version;
+	if (FI_VERSION_LT(api_version, FI_VERSION(1,18))) {
+		/* for API<1.18 we get use_device_rdma from the environment as
+		   it was queried during domain open. */
+		rxr_ep->use_device_rdma = efa_domain->use_device_rdma_api_before_118;
```
wzamazon, line 2362: It seems that we need not only the value of `use_device_rdma` and whether user set it explicitly. Can we check whether user set the envar explicitly on the fly? Then we do not need to save that in `rxr_ep`.

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -451,3 +460,45 @@ ssize_t efa_rdm_ep_post_flush(struct rxr_ep *ep, struct ibv_send_wr **bad_wr)
...
+bool efa_endpoint_support_rdma_write(struct rxr_ep *ep)
+{
+	struct efa_domain *domain;
+
+	if (!ep->use_device_rdma)
+		return 0;
+
+#if HAVE_CAPS_RDMA_WRITE
+	domain = rxr_ep_domain(ep);
+	return domain->device->device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE;
```
wzamazon, line 499: I would get rid of `domain`.

bwbarrett:
Also, no objection to the current code, but there's no longer a requirement to check the DEVICE_RDMA env var per-endpoint.  The NCCL plugin will set it using setopt() instead.  There's no harm in repeatedly checking, but it's also not a functional requirement from us.  Obviously, setopt working per endpoint is still a requirement.

prov/efa/src/rdm/rxr_ep.c
lrbison: small behavioral change for API 1.17 after this update: the environment variable would be read later and could change between enabling several endpoints.  I assume this is not a concern, and so we could remove the fetch during domain open.

wzamazon: Note that behavior of fetch envar during domain open was introduced very recently, and was not part of any release.

prov/efa/src/efa_user_info.c
wzamazon: > I was trying to leave room for the possibility that both of those checks would fail, in which case I would like to capture that in the log.

I think it is possible for both to fail. But if you want to capture both in log,  you will need to put an "and" between the two sentence. I am not sure it worth the added complexity

prov/efa/src/efa_domain.c
lrbison: It is.  I just cached it to strictly replicate pre-1.18 API behavior.  My latest push removes that cache and the value lives only in the rxr_ep.

prov/efa/src/efa_user_info.c
lrbison: Done.  I also re-worked the return logic a little.  Please take a close look here.

prov/efa/src/efa_user_info.c
lrbison: re-worked.  Please take a look and see what you think.

prov/efa/src/rdm/efa_rdm_peer.h
lrbison: The functions are now named efa_rdm_ep_support_rdma_read and *_rdma_write.

prov/efa/src/rdm/rxr_env.c
lrbison: Yes, thank you!  Fixed.

prov/efa/src/rdm/rxr_ep.h
lrbison: Good idea.  Done.

lrbison:
> Also, no objection to the current code, but there's no longer a requirement to check the DEVICE_RDMA env var per-endpoint. The NCCL plugin will set it using setopt() instead. There's no harm in repeatedly checking, but it's also not a functional requirement from us. Obviously, setopt working per endpoint is still a requirement.

hm.  I see your point.  My latest push removes the env var check on domain open, so we need to check the environment variable at least once per endpoint open() to set the default, but we continue to re-fetch each time setopt() is called.  While we could try to store the environment variable, I feel like the logic to ensure we don't let the application's setopt conflict with the environment doesn't simplify.

I think I'm happy with the logic as is unless you object.

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
+#include "efa_unit_tests.h"
```
wenduwan, line 1: Did you intend to put the tests in `efa_unit_test_info.c` instead?

`*_use_device_rdma.c` looks out of place(too specific).

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
...
+	assert_int_equal(fi_close(&ep->fid), 0);
+	assert_int_equal(fi_close(&domain->fid), 0);
+	assert_int_equal(fi_close(&fabric->fid), 0);
+	fi_freeinfo(info);
+	unsetenv("FI_EFA_USE_DEVICE_RDMA");
+
+	return;
+}
+
+const int x = -1; /* indicates "don't set" the variable below */
```
wenduwan, line 79: I guess this var is not necessary - the fact that we need to explain it made me think so.

prov/efa/src/efa_user_info.c
```diff
@@ -302,38 +303,58 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+		/* support FI_HMEM because we can use RDMA. */
+		return true;
 	}
+	/* Otherwise, we cannot support FI_HMEM.  Log some relevant reason. */
 
-	if (!rxr_env_get_use_device_rdma()) {
+	if (!hmem_ops[FI_HMEM_CUDA].initialized) {
+		extra_info = "no CUDA library or device was found.";
+	}
+	else {
```
wzamazon, line 328: `} else {'

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
...
+
+// settings agree: on
```
wenduwan, line 81: nit `/* */` is preferred currently

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
+#include "efa_unit_tests.h"
+
+
```
wenduwan, line 3: Also wonder about the code style. The max line length looks short - is that on purpose? Looks out of place.

prov/efa/src/efa_user_info.c
```diff
@@ -302,38 +303,58 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+	if (false == rxr_get_default_use_device_rdma()) {
 		EFA_WARN(FI_LOG_CORE,
-			"FI_HMEM capability requires RDMA, which is turned off. "
-			"You can turn it on by setting environment variable "
-			"FI_EFA_USE_DEVICE_RDMA to 1.\n");
-		return false;
+			"EFA cannot support FI_HMEM because the EFA device "
+			"does not have RDMA support and %s\n",
+			extra_info);
+	} else if (0 == rxr_env_get_use_device_rdma()) {
```
wzamazon, line 339: `if (!rxr_env_get_use_device_rdma())`

prov/efa/src/efa_user_info.c
```diff
@@ -302,38 +303,58 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+		EFA_WARN(FI_LOG_CORE,
+			"EFA cannot support FI_HMEM because the environment "
+			"variable FI_EFA_USE_DEVICE_RDMA is 0 and %s\n"
+			"You can enable FI_HMEM by setting "
```
wzamazon, line 343: I think we need to remove "You can xxx".

prov/efa/src/rdm/rxr_env.c
```diff
@@ -120,13 +83,17 @@ void rxr_env_define_use_device_rdma()
...
+	use_device_rdma = rxr_get_default_use_device_rdma();
 
 	/* Specify the help info about the usage of RDMA in the device. */
 	if (!(g_device_list[0].efa_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ))
-		str = "EFA device on your system does not support RDMA, so this variable cannot be set to 1";
+		str = "EFA device on your system does not support RDMA,"
+			" so this variable cannot be set to 1";
 	fi_param_define(&efa_prov, "use_device_rdma", FI_PARAM_BOOL,
-			"Specifies whether to use device's RDMA functionality for one-sided and two-sided transfer. (Default: %d). %s", use_device_rdma, str);
+			"Ignored if user selects libfabric API version >=1.18."
```
wzamazon, line 93: I do not think we ignore the environment variable.

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
...
+	struct fid_fabric *fabric = NULL;
+	struct fid_domain *domain = NULL;
+	struct fid_ep *ep = NULL;
+	struct rxr_ep *rxr_ep;
+	bool rdma_capable_hw;
+	char env_str[8];
+
+	if (env_val >= 0) {
+		snprintf(env_str,7,"%d",env_val);
+		setenv("FI_EFA_USE_DEVICE_RDMA",env_str, 1);
```
wenduwan, line 21: nit `",env` missing space

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: I could pass (-1,1,1) to functions below, but I found it visibly easier to parse as (x,1,1).  x is code for ignore.  Should I update the comment perhaps?

prov/efa/src/rdm/rxr_env.c
```diff
@@ -281,39 +248,42 @@ void rxr_env_initialize()
...
+ * Environments with non-boolean (1, true, yes, on) values for the variable
+ * will print to stderr and abort the program.
+ *
  * @return		0 - If FI_EFA_USE_DEVICE_RDMA is set to 0/false/no/off
- * 			1 - If FI_EFA_USE_DEVICE_RDMA is set to 1/true/yes/on
- * 			If FI_EFA_USE_DEVICE_RDMA is not set, then the value
- * 			returned from rxr_env_get_default_use_device_rdma()
- * 			will be the return value of this function.
+ *			1 - If FI_EFA_USE_DEVICE_RDMA is set to 1/true/yes/on
+ *			-FI_ENODATA - if FI_EFA_USE_DEVICE_RDMA is not set.
```
wzamazon, line 261: I think use return value both for result and return code is problematic.

Do we really need a function to both return the value of the envar and the value of the envar?

Can we just have a function `rxr_env_has_use_device_rdma()`? and use the other two functions to

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
...
+	} else {
+		unsetenv("FI_EFA_USE_DEVICE_RDMA");
+	}
+
+	hints = efa_unit_test_alloc_hints(FI_EP_RDM);
```
wenduwan, line 26: Any reason we cannot use `efa_unit_test_resource_construct`? I

prov/efa/src/efa_user_info.c
```diff
@@ -302,38 +303,58 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+	if (!hmem_ops[FI_HMEM_CUDA].initialized) {
+		extra_info = "no CUDA library or device was found.";
+	}
+	else {
+		assert(!FI_VERSION_GE(version, FI_VERSION(1, 18)));
+		extra_info = "the application is using libfabric "
+			"API <1.18, so CUDA cannot be used.";
+	}
+
+	if (false == rxr_get_default_use_device_rdma()) {
```
wzamazon, line 334: Do we really need `rxr_get_default_use_device_rdma` here? 

It seems, we just want to check whether device support rdma,  for that we can use `efa_device_support_rdma()`. Right?

prov/efa/test/efa_unit_test_use_device_rdma.c
wenduwan: I see... maybe instead of `x`, make it `#define NOT_SET -1`?

prov/efa/src/rdm/rxr_env.h
```diff
@@ -83,10 +85,56 @@ struct rxr_env {
...
+			use_device_rdma = 0;
+		}
+	}
+
+	return use_device_rdma;
+}
 
 int rxr_env_get_use_device_rdma();
 
+static inline bool rxr_env_or_default_use_device_rdma() {
```
wzamazon, line 126: I think we need `get` in the name

prov/efa/test/efa_unit_test_use_device_rdma.c
```diff
@@ -0,0 +1,118 @@
...
+
+	rxr_ep = container_of(ep, struct rxr_ep,
+			base_ep.util_ep.ep_fid.fid);
+	assert_int_equal( rxr_ep->use_device_rdma, expected_val );
+
+	assert_int_equal(fi_close(&ep->fid), 0);
+	assert_int_equal(fi_close(&domain->fid), 0);
+	assert_int_equal(fi_close(&fabric->fid), 0);
+	fi_freeinfo(info);
+	unsetenv("FI_EFA_USE_DEVICE_RDMA");
```
wenduwan, line 74: Hmmmm setting and resetting environment is tricky. If the test fails above this line then we end up with a dirty environment.

If we have to set/unsetenv, I think the correct way is to use the setup/teardown hook provided by cmocka.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: Line length is 80.  Was there something else about the style?

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1098,6 +1101,65 @@ static int rxr_ep_set_cuda_api_permitted(struct rxr_ep *ep, bool cuda_api_permit
...
+ * @return		0 on success
+ *
+ * @related rxr_ep
+ */
+static int rxr_ep_set_use_device_rdma(struct rxr_ep *ep, bool use_device_rdma)
+{
+	uint32_t api_version =
+		 rxr_ep_domain(ep)->util_domain.fabric->fabric_fid.api_version;
+
+	int env_setting = rxr_env_get_use_device_rdma();
```
wzamazon, line 1117: Following previous comments. It seems we can use the combination of `rxr_get_env_default_use_device_rdma()` and `rxr_env_has_use_device_rdma()`

prov/efa/src/efa_user_info.c
lrbison: Frankly efa_device_support_rdma_read and rxr_get_default_use_device_rdma  are essentially redundant, but they use a different approach: rxr_get_default checks the vendor id and then double-checks the device caps, while efa_device_support_rdma_read just checks device caps only.  Personally I think the device-capabilities is the right way.

Should we remove get_default_use_device_rdma ?

prov/efa/src/efa_user_info.c
lrbison: no.  The env_get will return -FI_ENODATA when the environment variable is not set.  I only want to print the message if the user has really set the variable.

prov/efa/src/efa_user_info.c
lrbison: because you think it's redundant or misleading?

wzamazon: I think it is misleading because at this point there is a possiblity that CUDA device or CUDA library is not available. Therefore simply set the envar does not guarantee user will get FI_HMEM support.

prov/efa/src/rdm/rxr_env.c
lrbison: I think our logic could be:

`if env_has_device_rdma() && env_get_rdma()`

or it could be

`if env_get_rdma() == 1`

I like the second one.

prov/efa/src/rdm/rxr_ep.c
lrbison: yes.  We could use a single function which either returns -1, 0, or 1, or we could use two functions.  You prefer two functions?

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: You are probably right.  Likely related, these tests can't come after the neuron tests, probably for a similar reason.

I think for these tests maybe we can just add the unsetenv() to teardown, because we want to setenv() to a variety of values

prov/efa/src/rdm/rxr_ep.c
wzamazon: Yes. I prefer two functions, both of them the return a boolean.

I do not like a function's return can be both a return code and a value,

prov/efa/src/efa_user_info.c
wzamazon: > Should we remove get_default_use_device_rdma ?

Yes. I think so.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: hm, we do need to do getinfo with a couple different versions.  maybe we add an argument to resource_construct?

prov/efa/src/rdm/rxr_env.h
lrbison: the name should match the file too right?  so: `rxr_env_get_default_or_env_use_device_rdma()` ?

wzamazon: It seems we can merge this one with `rxr_env_get_use_device_rdma()` and have a standalone `rxr_env_has_use_device_rdma()`?

prov/efa/test/efa_unit_test_use_device_rdma.c
wenduwan: > maybe we add an argument to resource_construct?

I agree - or provide/overload another entry point and keep the current callers unchanged.

lrbison: oh shoot.  We also skip() the test early on instances without hardware RDMA.  Not sure how to do that cleanly.  Any other ideas how to check for hardware support?

wenduwan: If that is known at compile time we use a macro and skipped versions of the tests https://github.com/ofiwg/libfabric/blob/main/prov/efa/test/efa_unit_test_cq.c#L283

If it is inferred at runtime then we use the skip function from cmocka https://github.com/ofiwg/libfabric/blob/main/prov/efa/test/efa_unit_test_device.c#L14

lrbison: I am using the skip() function, but halfway through the resource creation.  If I let the endpoint get created with the wrong environment value the code will abort() rather than skip().

I don't see an easy way to re-use that function.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: moved into test_info.c.  Thanks.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: I've named it VALUE_NOT_SET.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: fixed.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: fixed.

prov/efa/test/efa_unit_test_use_device_rdma.c
lrbison: moved the unset to the teardown hook.

Because setup is test-dependent, I'm still setting the value there.

prov/efa/src/efa_user_info.c
lrbison: fixed.

prov/efa/src/efa_user_info.c
lrbison: `get_default_use_device_rdma`  is gone.  `rxr_env_get_use_device_rdma` returns the environment value or the default, and a  new function (`rxr_env_is_set_use_device_rdma`) can be called to query if the environment variable is set.

prov/efa/src/efa_user_info.c
lrbison: I think this is more clear now.

prov/efa/src/efa_user_info.c
lrbison: Removed, although at this point we know that CUDA device and library are available.

prov/efa/src/rdm/rxr_env.c
lrbison: No, not anymore.  Thank you, fixed.

prov/efa/src/rdm/rxr_env.c
lrbison: Using two boolean functions now.

prov/efa/src/rdm/rxr_ep.c
lrbison: Done.

lrbison:
Rebased to fix merge conflict.

prov/efa/src/rdm/rxr_env.h
lrbison: Good idea.  Done.

lrbison:
The AppVeyor faillure seems real but unrelated:
```
- name:   rdm_tagged_peek -p sockets
  timestamp: Sun 03/19/2023 15:24:52.14
  result: Fail
  time:   183
  server_cmd: C:\projects\libfabric\fabtests\x64\Release-v141\rdm_tagged_peek -p sockets  -s 127.0.0.1
  server_stdout:
  client_cmd: C:\projects\libfabric\fabtests\x64\Release-v141\rdm_tagged_peek -p sockets  -s 127.0.0.1 127.0.0.1
  client_stdout
  ```

File: prov/efa/src/rdm/rxr_env.c [hunk 23]
```diff
@@ -281,39 +247,55 @@ void rxr_env_initialize()
...
- * 			returned from rxr_env_get_default_use_device_rdma()
- * 			will be the return value of this function.
+ *			1 - If FI_EFA_USE_DEVICE_RDMA is set to 1/true/yes/on
  */
-int rxr_env_get_use_device_rdma()
+bool rxr_env_get_use_device_rdma()
 {
-	int ret, use_device_rdma;
+	int ret;
+	int param_val;
```
wzamazon, line 265: Don't we need a default value for `param_val`? which is used when FI_EFA_USE_DEVICE_RDMA is not set.

prov/efa/test/efa_unit_test_info.c
```diff
@@ -99,6 +99,64 @@ void test_info_open_ep_with_api_1_1_info()
...
+	 * two options: FI_MR_BASIC or FI_MR_SCALABLE, (EFA does not support FI_MR_SCALABLE).
+	 *
+	 * Additional information about memory registration is specified as bits in
+	 * "mode". For example, the requirement of local memory registration
+	 * is specified as FI_LOCAL_MR.
+	 */
+	hints->mode = FI_LOCAL_MR;
+	hints->domain_attr->mr_mode = FI_MR_BASIC;
+
+	err = fi_getinfo(FI_VERSION(1, 1), NULL, NULL, 0ULL, hints, &info);
```
wzamazon: The function name says use 1.18 info, but the code is using 1.1.

wzamazon: This test was not added to the list of tests.

wzamazon (changes requested):

prov/efa/src/efa_user_info.c
```diff
@@ -302,38 +304,57 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+	/* support FI_HMEM because we can use RDMA. */
+	if (rdma_allowed) return true;
+
+	/* Otherwise, we cannot support FI_HMEM.  Log some relevant reason. */
+
+	if (!hmem_ops[FI_HMEM_CUDA].initialized) {
+		extra_info = "no CUDA library or device was found.";
+	} else {
+		assert(!FI_VERSION_GE(version, FI_VERSION(1, 18)));
+		extra_info = "the application is using libfabric "
```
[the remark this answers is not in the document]
lrbison: I've put all that logic into an if-block which requires FI_HMEM_CUDA to be initialized.

This made me realize the return value is a little strange in a few cases (and has been before my changes):
 - If no HMEM system is initialized, this function still returns true.
 - If multiple HMEM systems are present, we might return true if one of them can be supported even if the other can't (ie, our short-circuit with API version and CUDA enabled ignores any other HMEM systems and their requirements).

prov/efa/test/efa_unit_test_info.c
```diff
@@ -99,6 +99,64 @@ void test_info_open_ep_with_api_1_1_info()
...
+	assert_non_null(hints->domain_attr);
+
+	hints->fabric_attr = calloc(sizeof(struct fi_fabric_attr), 1);
+	assert_non_null(hints->fabric_attr);
+
+	hints->ep_attr = calloc(sizeof(struct fi_ep_attr), 1);
+	assert_non_null(hints->ep_attr);
+
+	hints->fabric_attr->prov_name = "efa";
+	hints->ep_attr->type = FI_EP_RDM;
```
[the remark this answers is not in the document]
lrbison: Somehow it looks like you reviewed an old version of this file.  This function has been significantly reworked and renamed.

prov/efa/src/efa_user_info.c
```diff
@@ -287,43 +293,76 @@ bool efa_user_info_should_support_hmem(int version, const struct fi_info *hints)
...
+			/* For this API we can support HMEM regardless of
+			   use_device_rdma and P2P support, because we can use
+			   CUDA api calls.*/
+			return true;
+	} else if (hmem_ops[FI_HMEM_CUDA].initialized) {
+		extra_info = "For CUDA and libfabric API version <1.18 ";
+	}
+	rdma_allowed = rxr_env_get_use_device_rdma();
+
+	EFA_HMEM_IFACE_FOREACH_NON_SYSTEM(hmem_type) {
```
wzamazon, line 325: Can you put `any_hmem = false` here?

prov/efa/src/rdm/rxr_env.c
```diff
@@ -281,39 +247,56 @@ void rxr_env_initialize()
...
 	/* Fetch the value of environment variable set by the user if any. */
-	ret = fi_param_get_bool(&efa_prov, "use_device_rdma", &use_device_rdma);
+	ret = fi_param_get_bool(&efa_prov, "use_device_rdma", &param_val);
 	if (ret == -EINVAL){
 		fprintf(stderr, "FI_EFA_USE_DEVICE_RDMA was set to an invalid value by the user."
 			" FI_EFA_USE_DEVICE_RDMA is boolean and can be set to only 0/false/no/off or"
 			" 1/true/yes/on.\n");
 		abort();
 	}
+	if (ret == -FI_ENODATA) return hw_support;
```
wzamazon, line 279: This is not correct. For device ID that is 0xefa0 and 0xefa1, we do not enable RDMA by default even if they support RDMA, this is to maintain consistent behavior.

wzamazon: One thing need to be considered is the API version.

When user use 1.18 API and cuda is initialized, we current support FI_HMEM. In this case, we need to enable rdma if device support it.

prov/efa/src/rdm/rxr_env.c
```diff
@@ -281,39 +247,56 @@ void rxr_env_initialize()
...
+	return param_val;
+}
+
+/**
+ * @brief Return true if the environment variable FI_EFA_USE_DEVICE_RDMA is present
+ *
+ * @return true  - FI_EFA_USE_DEVICE_RDMA is defined in the environment
+ * @return false - otherwise.
+ */
+bool rxr_env_is_set_use_device_rdma() {
```
wzamazon, line 298: I would prefer `has` over `is_set`. But no strong opinion.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1098,6 +1101,75 @@ static int rxr_ep_set_cuda_api_permitted(struct rxr_ep *ep, bool cuda_api_permit
...
+		/* let the application developer know something is wrong */
+		EFA_WARN( FI_LOG_EP_CTRL,
+			"Applications using libfabric API version <1.18 are not "
+			"allowed to call fi_setopt with FI_OPT_EFA_USE_DEVICE_RDMA.  "
+			"Please select a newer libfabric API version in "
+			"fi_getinfo during startup to use this option.\n");
+		return -FI_ENOPROTOOPT;
+	}
+
+	if (env_set && use_device_rdma && !env_value) {
```
wzamazon, line 1139: can we just check `use_device_rdma != env_value`?

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1098,6 +1101,75 @@ static int rxr_ep_set_cuda_api_permitted(struct rxr_ep *ep, bool cuda_api_permit
...
+		return -FI_EINVAL;
+	}
+	if (use_device_rdma && !efa_device_support_rdma_read()) {
+		/* conflict: application on, hardware off. */
+		/* hardware always wins ;-) */
+		ep->use_device_rdma = false;
+		EFA_WARN(FI_LOG_EP_CTRL,
+		"Application used setopt to request use_device_rdma, "
+		"but EFA device does not support it\n");
+		return -FI_EINVAL;
```
wzamazon, line 1167: I think it should be `-FI_EOPNOTSUPP`

prov/efa/test/efa_unit_test_info.c
```diff
@@ -99,6 +99,64 @@ void test_info_open_ep_with_api_1_1_info()
 	assert_int_equal(err, 0);
 }
 
+/**
+ * @brief test use_device_rdma behavior before and after API 1.18
+ */
+void test_info_open_ep_with_api_1_18_info()
```
wzamazon: There are changes  to unit test that went into 1st commit.

Can you put all unit test changes in a separate commit?

wzamazon (changes requested):
Looks much better.

I think one thing missing is that default value handling.

default value are different for different vendor ID.

and also different for version of libfabric API

include/rdma/fi_ext.h
```diff
@@ -70,9 +70,10 @@ extern "C" {
 /* negative options are provider specific */
 enum {
        FI_OPT_EFA_RNR_RETRY = -FI_PROV_SPECIFIC_EFA,
-       FI_OPT_EFA_EMULATED_READ,       /* bool */
-       FI_OPT_EFA_EMULATED_WRITE,      /* bool */
-       FI_OPT_EFA_EMULATED_ATOMICS,    /* bool */
+       FI_OPT_EFA_EMULATED_READ,	/* bool */
+       FI_OPT_EFA_EMULATED_WRITE,	/* bool */
+       FI_OPT_EFA_EMULATED_ATOMICS,	/* bool */
+       FI_OPT_EFA_USE_DEVICE_RDMA,	/* bool */
```
wzamazon, line 76: need update document `fi_efa`

lrbison: documentation updated.

prov/efa/src/rdm/rxr_ep.c
lrbison: No.  For instance, the application can set use_device_rdma off, but the default environment variable could have been on, but not explicitly set.  We want to ensure we only enter the `if` when the user has explicitly set the variable to be in conflict with the provided argument.

prov/efa/src/rdm/rxr_env.c
lrbison: changed to has.

prov/efa/src/rdm/rxr_env.c
lrbison: done.

prov/efa/src/efa_user_info.c
lrbison: Done.

wenduwan (approved):
Reviewed test code. lgtm. Approve.

prov/efa/src/efa_user_info.c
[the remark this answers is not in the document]
lrbison: Reworked this function to fix the above issues.

prov/efa/src/efa_user_info.c
```diff
@@ -278,45 +278,82 @@ int efa_user_info_get_dgram(uint32_t version, const char *node, const char *serv
...
 		EFA_INFO(FI_LOG_CORE,
-			"User is using API version >= 1.18. CUDA library and device are availble, claim support of FI_HMEM.");
-		return true;
+			"User is using API version >= 1.18. CUDA library and "
+			"device are available, claim support of FI_HMEM.");
+			/* For this API we can support HMEM regardless of
+			   use_device_rdma and P2P support, because we can use
+			   CUDA api calls.*/
+			return true;
+	} else if (hmem_ops[FI_HMEM_CUDA].initialized) {
```
wzamazon, line 310: Nit. No need for `else` here because we already returned in previous branch

prov/efa/src/rdm/efa_rdm_util.c
```diff
@@ -0,0 +1,103 @@
...
+ * SOFTWARE.
+ */
+
+#include "efa_rdm_util.h"
+
+/**
+ * @brief Fetch the value of the environment variable FI_EFA_USE_DEVICE_RDMA
+ *
+ * This function fetches the value of FI_EFA_USE_DEVICE_RDMA from the
+ * user's environment.
```
wzamazon, line 40: This function check more than environment variable. It also check device capability and vendor_part_id, etc.

prov/efa/src/rdm/efa_rdm_util.c
```diff
@@ -0,0 +1,103 @@
...
+	   support RDMA, exit the run. */
+	if (param_val && !hw_support) {
+		fprintf(stderr, "FI_EFA_USE_DEVICE_RDMA=1 was set by user, but "
+			"EFA device has no rdma-read capability.  Application "
+			"will abort().\n");
+		abort();
+	}
+
+	return param_val;
+}
```
wzamazon, line 103: No end of line?

prov/efa/src/rdm/efa_rdm_util.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#ifndef _EFA_RDM_UTIL_H
+#define _EFA_RDM_UTIL_H
+
+#include "efa.h"
+
+bool efa_rdm_get_use_device_rdma(uint32_t fabric_api_version);
+
+#endif /* _EFA_RDM_UTIL_H */
```
wzamazon, line 41: No end of line?

prov/efa/src/rdm/rxr_env.c
```diff
@@ -120,13 +83,16 @@ void rxr_env_define_use_device_rdma()
 	int use_device_rdma;
 
 	/* Get the default value of using device's RDMA capability */
-	use_device_rdma = rxr_env_get_default_use_device_rdma();
+	use_device_rdma = efa_device_support_rdma_read();
```
wzamazon, line 86: This also depend on vendor part ID

wzamazon (commented):
Most comments are cosmetic.

only important one is the one in defining the environment variable

prov/efa/src/rdm/rxr_env.c
lrbison: addressed

prov/efa/src/rdm/efa_rdm_util.h
lrbison: fixed

prov/efa/src/rdm/efa_rdm_util.c
lrbison: re-wrote the documentation

prov/efa/src/efa_user_info.c
lrbison: removed

man/fi_efa.7.md
```diff
@@ -103,6 +103,22 @@ provider for AWS Neuron or Habana SynapseAI.
   emulating Read, Write, and Atomic operations (return value is true), or if
   these operations are assisted by hardware support (return value is false).
 
+*FI_OPT_EFA_USE_DEVICE_RDMA - bool*
+: Only available if the application selects a libfabric API version >= 1.18.
```
wzamazon, line 107: Need to specify that it apply to setopt.

wzamazon (approved):
one minor comments

man/fi_efa.7.md
lrbison: it applies to both setopt and getopt.  Should I indicate that explicitly?

wzamazon: I see. The commit title says "set use_device_rdma with fi_setopt". I think you just need to apply commit title.

lrbison: commit updated.

shijin-aws (approved):

PR #8663, opened 2023-03-16
core/util_cq: fix the behavior of cq_read for FI_PEER

shijin-aws:
For peer cq, fi_cq_read is not expected to access cirq and reading cqes. It should only progress the cq.

### DIFF
--- a/prov/util/src/util_cq.c
+++ b/prov/util/src/util_cq.c
@@ -628,14 +628,21 @@ static struct fi_ops_cq_owner util_peer_cq_src_owner_ops = {
 	.writeerr = &util_peer_cq_writeerr,
 };
 
-static ssize_t util_cq_read(struct fid_cq *cq_fid, void *buf, size_t count)
+/* For peer cq, just do progress */
+static ssize_t ofi_peer_cq_read(struct fid_cq *cq_fid, void *buf, size_t count)
 {
-       return ofi_cq_readfrom(cq_fid, buf, count, NULL);
+	struct util_cq *cq;
+
+	cq = container_of(cq_fid, struct util_cq, cq_fid);
+
+	cq->progress(cq);
+
+	return 0;
 }
 
 static struct fi_ops_cq util_peer_cq_ops = {
        .size = sizeof(struct fi_ops_cq),
-       .read = util_cq_read,
+       .read = ofi_peer_cq_read,
        .readfrom = fi_no_cq_readfrom,
        .readerr = fi_no_cq_readerr,
        .sread = fi_no_cq_sread,


shijin-aws:
@shefty @aingerson I hit a segfault when I call `fi_cq_read(shm_cq, NULL, 0)` after I configured shm_cq as a peer cq. This patch fixes my patch and I think it should be right according to the man page and this comment in `struct util_cq`:

```
/* Only valid if not FI_PEER */
	struct util_comp_cirq	*cirq;
```

Thank you for your feedback!

prov/util/src/util_cq.c
```diff
@@ -258,6 +258,10 @@ ssize_t ofi_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 	cq = container_of(cq_fid, struct util_cq, cq_fid);
 
 	cq->progress(cq);
+
+	if (cq->flags & FI_PEER)
+		return 0;
```
shefty, line 263: Is there not an ofi_cq_peer_read() call that simply calls progress that can be used?  If not, having one would be preferred to inserting this check in the fast path of all cq read calls.

aingerson: Agreed with Sean. I would prefer not to have this added check in the regular path. The definition of the peer implementation is that you can only drive progress on it by calling fi_cq_read with NULL and 0 to it never tries to read the CQ. We should set the FI_PEER cq_read function pointer so it only drives progress.

shijin-aws: I see, that makes sense to me, currently the util_peer_cq_ops has cq_read as util_cq_read, which is ofi_cq_readfrom. I think I just need to introduce a ofi_cq_peer_read() that only progress the cq. 

```
static ssize_t util_cq_read(struct fid_cq *cq_fid, void *buf, size_t count)
{
       return ofi_cq_readfrom(cq_fid, buf, count, NULL);
}

static struct fi_ops_cq util_peer_cq_ops = {
       .size = sizeof(struct fi_ops_cq),
       .read = util_cq_read,
       .readfrom = fi_no_cq_readfrom,
       .readerr = fi_no_cq_readerr,
       .sread = fi_no_cq_sread,
       .sreadfrom = fi_no_cq_sreadfrom,
       .signal = fi_no_cq_signal,
       .strerror = fi_no_cq_strerror,
};
```

Is that right?

aingerson: @shijin-aws Yes, I think that's better. Thanks for catching!

shijin-aws: Thanks. `util_cq_read` becomes unused after I introduce `ofi_peer_cq_read` so I removed it. Question: Is `ofi_*` function aimed to be non-static?

aingerson (approved):

prov/util/src/util_cq.c
shefty: I think this is good, and I wouldn't worry about ofi vs util as the prefix.  I'd rather see ofi myself.

aingerson: I believe ofi_* functions are meant to be called from providers and util_* functions called internally within the util code. That's at least what I've been doing by @shefty can confirm or contradict

shijin-aws: Thank you both. Rephrased my commit message to reflect the updated change.

shefty: I believe that's the case in general.  I.e. util_ are static.  But I don't think that distinction is particularly useful, so I wouldn't worry about it.

PR #8668, opened 2023-03-17
prov/shm: fix the srx import procedure.

shijin-aws:
In smr_srx_context, shm should not modify the peer_ops of the imported srx. It should allocate a new srx and set it there.

This bug is fixed in https://github.com/ofiwg/libfabric/pull/8456 but that PR was closed. Open this PR for now

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -1005,45 +1005,6 @@ static int smr_ep_bind_cntr(struct smr_ep *ep, struct util_cntr *cntr, uint64_t 
 	return FI_SUCCESS;
 }
 
-static int smr_ep_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
-{
-	struct smr_ep *ep;
-	struct util_av *av;
-	int ret = 0;
-
-	ep = container_of(ep_fid, struct smr_ep, util_ep.ep_fid.fid);
-	switch (bfid->fclass) {
-	case FI_CLASS_AV:
-		av = container_of(bfid, struct util_av, av_fid.fid);
-		ret = ofi_ep_bind_av(&ep->util_ep, av);
-		if (ret) {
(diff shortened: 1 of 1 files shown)


prov/shm/src/smr_ep.c
```diff
@@ -1696,6 +1696,12 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			if (ret)
 				return ret;
 		} else {
+			srx = calloc(1, sizeof(*srx));
+			srx->peer_ops = &smr_srx_peer_ops;
+			srx->owner_ops = smr_get_peer_srx(ep)->owner_ops;
+			srx->ep_fid.fid.context =
+				smr_get_peer_srx(ep)->ep_fid.fid.context;
+			ep->srx = &srx->ep_fid;
```
aingerson, line 1704: I know this is my bug and this fix came from my recent srx PR that I closed and I think it's a solution but I'm wondering if it would be better to move this to where we bind the srx to the ep [here](https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_ep.c#L1036)? I think it's weird that we set the ep->srx and then override it. Would probably be better to just set it accordingly in the first place? What do you think?

shijin-aws: That sounds better to me, will update, thank you!

aingerson: Thanks!

shijin-aws: Updated, I need to move smr_ep_bind after the definition of smr_srx_peer_ops.

shijin-aws:
bot:aws:retest

shijin-aws:
@aingerson may I know what's the error in  continuous-integration/jenkins/pr-head ?

shijin-aws:
bot:aws:retest

PR #8670, opened 2023-03-17
prov/shm: refactor smr structures

amirshehataornl:
Refactor the smr_tx_entry and smr_pend_entry structures. Both these structures serve very similar purposes and contain many of the same attributes. Instead of maintaining two different definitions combine into one definition and unionize the different fields.

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -121,32 +121,33 @@ struct smr_rx_entry {
 };
 
 struct smr_tx_entry {
-	struct smr_cmd	cmd;
-	int64_t		peer_id;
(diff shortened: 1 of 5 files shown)


amirshehataornl:
@aingerson, is that what you were more or less thinking about?

aingerson:
@amirshehataornl Yeah same. I think the main motivation for this in the first place was the DSA path which is weird and benefits from them being the same structure. Otherwise, I think it's more or less of a toss up.

aingerson:
@amirshehataornl Going to abandon this refactor if that's ok with you.

PR #8673, opened 2023-03-19
prov/efa: fix getinfo return code when hmem support is requested

wzamazon:
Fix a bug in `efa_user_info_alter_rxr` regarding hmem support, and added a test case for it.

### DIFF
--- a/prov/efa/test/efa_unit_test_info.c
+++ b/prov/efa/test/efa_unit_test_info.c
@@ -1,3 +1,4 @@
+#include <stdlib.h>
 #include "efa_unit_tests.h"
 
 /**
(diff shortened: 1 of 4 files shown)


prov/efa/test/efa_unit_test_info.c
```diff
@@ -151,3 +152,75 @@ void test_info_check_shm_info()
...
+ *        using different version of libfabric API, the EFA provider
+ *        behavior as expected on a system has CUDA
+ */
+void test_info_check_hmem_cuda_support_on_api_version()
+{
+	struct fi_info *hints, *info = NULL;
+	int err;
+
+	if (!hmem_ops[FI_HMEM_CUDA].initialized)
+		return;
```
wenduwan, line 167: Instead of return(which means the test succeeded without testing anything), we should skip.

prov/efa/test/efa_unit_test_info.c
```diff
@@ -151,3 +152,75 @@ void test_info_check_shm_info()
...
+
+	if (efa_device_support_rdma_read()) {
+		/* on a system that support GPUDirect RDMA read,
+		 * HMEM cuda is available when GPUDirect RDMA is available,
+		 * and environment variable FI_EFA_USE_DEVICE_RDMA set to 1/on/true
+		 */
+		setenv("FI_EFA_USE_DEVICE_RDMA", "1", true /* overwrite */);
+		err = fi_getinfo(FI_VERSION(1, 14), NULL, NULL, 0, hints, &info);
+		assert_int_equal(err, 0);
+		fi_freeinfo(info);
```
wenduwan, line 194: Any reason not to separate this into another test, i.e. `test_xxx_on_api_lt_1_18` and `test_xxx_on_api_ge_1_18`

prov/efa/test/efa_unit_test_info.c
```diff
@@ -151,3 +152,75 @@ void test_info_check_shm_info()
...
+
+	hints->caps |= FI_HMEM;
+	hints->domain_attr->mr_mode |= FI_MR_HMEM;
+
+	/* When a user used >= 1.18 API, CUDA is always supported.
+	 * This is because in version 1.18, libfabric EFA provider implemented
+	 * universal CUDA support through CUDA library. However, this feature
+	 * can cause some middleware to deadlock, thus it is only available
+	 * when a user is using 1.18 API.
+	 */
```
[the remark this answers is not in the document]
wzamazon: > CUDA support was introduced to EFA before 1.18 IIUC.

Yes, but it requires instance to support GPUDirect RDMA

prov/efa/test/efa_unit_test_info.c
wzamazon: I will do that.

prov/efa/test/efa_unit_test_info.c
```diff
@@ -151,3 +152,75 @@ void test_info_check_shm_info()
...
+ *        it is not requested
+ */
+void test_info_check_no_hmem_support_when_not_requested()
+{
+	struct fi_info *hints, *info = NULL;
+	int err;
+
+	hints = efa_unit_test_alloc_hints(FI_EP_RDM);
+
+	err = fi_getinfo(FI_VERSION(1,6), NULL, NULL, 0, hints, &info);
```
wenduwan, line 221: A comment or better test name can be helpful - why test 1.6 version?

prov/efa/test/efa_unit_test_info.c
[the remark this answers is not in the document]
wzamazon: Updated commits

prov/efa/test/efa_unit_test_info.c
wzamazon: used skip in new version

prov/efa/test/efa_unit_test_info.c
wzamazon: No particular reason. any version >= 1.5 should work

prov/efa/test/efa_unit_test_info.c
```diff
@@ -151,3 +152,90 @@ void test_info_check_shm_info()
...
+		fi_freeinfo(info);
+	} else {
+		assert_int_equal(err, -FI_ENODATA);
+	}
+
+	setenv("FI_EFA_USE_DEVICE_RDMA", "0", true /* overwrite */);
+	err = fi_getinfo(FI_VERSION(1, 14), NULL, NULL, 0, hints, &info);
+	assert_int_equal(err, -FI_ENODATA);
+
+	unsetenv("FI_EFA_USE_DEVICE_RDMA");
```
wenduwan, line 192: I had a similar comment in Luke's PR. unsetting env should happen in the teardown hook. I see Luke already added that.

prov/efa/test/efa_unit_test_info.c
```diff
@@ -151,3 +152,90 @@ void test_info_check_shm_info()
...
+	 * on a system that support GPUDirect RDMA read,
+	 * HMEM cuda is available when GPUDirect RDMA is available,
+	 * and environment variable FI_EFA_USE_DEVICE_RDMA set to 1/on/true
+	 * otherwise it is not available.
+	 */
+	setenv("FI_EFA_USE_DEVICE_RDMA", "1", true /* overwrite */);
+	err = fi_getinfo(FI_VERSION(1, 6), NULL, NULL, 0, hints, &info);
+	if (efa_device_support_rdma_read()) {
+		assert_int_equal(err, 0);
+		fi_freeinfo(info);
```
wenduwan, line 183: Being pedantic - `fi_freeinfo` should be called before assertion. Same in `test_info_check_hmem_cuda_support_on_api_ge_1_18`

wzamazon: I do not think so. If err is not 0, we should not freeinfo.

prov/efa/test/efa_unit_test_info.c
wzamazon: Thanks! will do

prov/efa/test/efa_unit_test_info.c
wenduwan: I see. You are right. I was mistaken.

PR #8676, opened 2023-03-20
prov/efa: fix wrong mocked return of ibv_wc_read_slid() in unit test

wzamazon:
In test_rxr_ep_handshake_exchange_host_id(), the mocked return of ibv_wr_read_slid() is always 0, which
is wrong, because it should return peer's ahn.

This patch fixed it.


wenduwan (approved):
Thank you!

PR #8678, opened 2023-03-20
fabtests: call cudaSetDevice() for selected device

wzamazon:
If user selected device by using "-i" option and
selected CUDA memory, this patch call cudaSetDevice() to use that device.

### DIFF
--- a/fabtests/common/hmem_cuda.c
+++ b/fabtests/common/hmem_cuda.c
@@ -50,6 +50,7 @@ struct cuda_ops {
 	cudaError_t (*cudaMemset)(void *ptr, int value, size_t count);
 	const char *(*cudaGetErrorName)(cudaError_t error);
 	const char *(*cudaGetErrorString)(cudaError_t error);
+	cudaError_t (*cudaSetDevice)(int device);
 	CUresult (*cuPointerSetAttribute)(void *data,
 					  CUpointer_attribute attribute,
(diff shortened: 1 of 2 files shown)


fabtests/common/shared.c
```diff
@@ -782,7 +782,21 @@ int ft_init(void)
...
 	}
 
-	return ft_hmem_init(opts.iface);
+	ret = ft_hmem_init(opts.iface);
+	if (ret != FI_SUCCESS) {
+		FT_ERR("ft_hmem_init: %d", ret);
+		return ret;
+	}
+
+	if (opts.iface == FI_HMEM_CUDA) {
```
shijin-aws, line 791: shouldn't we do this in `fabtests/common/hmem_cuda.c`?

shijin-aws: neuron does this in `ft_neuron_alloc`, we should do it similarly in `ft_cuda_alloc`.

shefty: I think it makes sense in ft_cuda_init().  Likely don't need to expose ft_cuda_set_device().

wzamazon: I moved it to inside `ft_cuda_init()` in new revision, also removed `ft_cuda_set_device()`

shijin-aws (approved):

j-xiong (approved):

PR #8679, opened 2023-03-20
prov/efa: refactor handshake test to prevent hang

wenduwan:
Currently test_rxr_ep_handshake_exchange_host_id makes early assertions
that can potentially cause leaked CQE. This prevents RDM endpiont from
closing properly(a hang). This change ensures the CQ be purged before
making assertions.

The test is also simplified to eliminate the redundant fi_recv and
associated rx buffer. Handshake happens behind the scene in the progress
engine. The application does not participate in request initiation.

### DIFF
--- a/prov/efa/test/efa_unit_test_ep.c
+++ b/prov/efa/test/efa_unit_test_ep.c
@@ -85,20 +85,22 @@ void test_rxr_ep_ignore_non_hex_host_id(struct efa_resource **state)
 void test_rxr_ep_handshake_exchange_host_id(struct efa_resource **state, uint64_t local_host_id, uint64_t peer_host_id, bool include_connid)
 {
 	fi_addr_t peer_addr = 0;
-	int ret;
+	int cq_read_recv_ret, cq_read_send_ret;
 	struct efa_ep_addr raw_addr = {0};
 	size_t raw_addr_len = sizeof(raw_addr);
 	struct efa_rdm_peer *peer;
 	struct efa_resource *resource = *state;
-	struct efa_unit_test_buff recv_buff;
 	struct efa_unit_test_handshake_pkt_attr pkt_attr = {0};
 	struct fi_cq_data_entry cq_entry;
 	struct ibv_qp *ibv_qp;
 	struct rxr_ep *rxr_ep;
 	struct rxr_pkt_entry *pkt_entry;
+	uint64_t actual_peer_host_id = UINT64_MAX;
 
 	g_efa_unit_test_mocks.local_host_id = local_host_id;
 	g_efa_unit_test_mocks.peer_host_id = peer_host_id;
+
+	assert_false(actual_peer_host_id == g_efa_unit_test_mocks.peer_host_id);
 
 	efa_unit_test_resource_construct(resource, FI_EP_RDM);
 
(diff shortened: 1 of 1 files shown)


prov/efa/test/efa_unit_test_ep.c
```diff
@@ -168,18 +163,10 @@ void test_rxr_ep_handshake_exchange_host_id(struct efa_resource **state, uint64_
 	will_return(efa_mock_ibv_read_vendor_err_return_mock, FI_EFA_ERR_OTHER);
```
wckzhang, line 163: Just an observation, the "Setup CQ" code seems like we could probably export it into a function. I assume we do this in multiple places

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -106,17 +106,12 @@ void test_rxr_ep_handshake_exchange_host_id(struct efa_resource **state, uint64_
 	rxr_ep->host_id = g_efa_unit_test_mocks.local_host_id;
 	rxr_ep->use_shm_for_tx = false;
 
-	/* Construct a minimal recv buffer */
-	efa_unit_test_buff_construct(&recv_buff, resource, rxr_ep->min_multi_recv_size);
```
wckzhang, line 110: Why isn't this necessary anymore?

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -168,18 +163,10 @@ void test_rxr_ep_handshake_exchange_host_id(struct efa_resource **state, uint64_
 	will_return(efa_mock_ibv_read_vendor_err_return_mock, FI_EFA_ERR_OTHER);
 	will_return(efa_mock_ibv_start_poll_return_mock, IBV_WC_SUCCESS);
 
-	/* Post receive buffer */
-	ret = fi_recv(resource->ep, recv_buff.buff, recv_buff.size, fi_mr_desc(recv_buff.mr), peer_addr, NULL /* context */);
```
wckzhang, line 172: Why did you remove this section?

prov/efa/test/efa_unit_test_ep.c
wenduwan: Handshake in the context of EFA protocol happens in the background without the application knowing. It actually does not require `fi_recv/send` - instead it happens purely in the progress engine using prov/efa bounce buffer.

prov/efa/test/efa_unit_test_ep.c
wenduwan: Yes. We do similar mocking in a few places. My assessment on this particular case is that it's probably not easy to generalize and extract to a common function, unless we make that function accept a long list of function pointers(mocks). The other challenge is to mock ordered function calls.

prov/efa/test/efa_unit_test_ep.c
wenduwan: > Handshake in the context of EFA protocol happens in the background without the application knowing. It actually does not require fi_recv/send - instead it happens purely in the progress engine using prov/efa bounce buffer.

prov/efa/test/efa_unit_test_ep.c
wckzhang: Ah got it, yeah EFA automatically posts recv bufs, but if you want to recv it into a user buffer you need to call fi_recv. Can you put the reasoning into the commit description?

wenduwan: Good idea. Thanks! Reworded the commit.

wckzhang (approved):

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -85,17 +85,17 @@ void test_rxr_ep_ignore_non_hex_host_id(struct efa_resource **state)
...
 	size_t raw_addr_len = sizeof(raw_addr);
 	struct efa_rdm_peer *peer;
 	struct efa_resource *resource = *state;
-	struct efa_unit_test_buff recv_buff;
 	struct efa_unit_test_handshake_pkt_attr pkt_attr = {0};
 	struct fi_cq_data_entry cq_entry;
 	struct ibv_qp *ibv_qp;
 	struct rxr_ep *rxr_ep;
 	struct rxr_pkt_entry *pkt_entry;
+	uint64_t actual_peer_host_id = peer_host_id + 1;
```
wzamazon, line 98: Why does the " + 1"?

wenduwan: > Why does the " + 1"?

This ensures `actual_peer_host_id != peer_host_id` at the beginning of the test.

wzamazon: I think use `(uint64_t)-1` is better

wenduwan: Cool. Incorporated in update. Thanks!

PR #8682, opened 2023-03-21
prov/efa: introduce setopt flags for 128 bytes in order transmission

wzamazon:
This PR introduced setopt flags to enable the in order transmission of aligned 128 bytes.

For this change, this PR included a commit that moved the enablement of base_ep from `rxr_ep_ctrl` to `rxr_endpoint`, because the call to `fi_setopt` need to query ibv_qp, and ibv_qp is created during enablement of base_ep.

### DIFF
--- a/man/fi_efa.7.md
+++ b/man/fi_efa.7.md
@@ -119,6 +119,18 @@ provider for AWS Neuron or Habana SynapseAI.
   For API<1.18, RDMA is enabled by default only on certain newer hardware
   revisions.
 
+*FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES - bool*
+: It is used to force the endpoint to use in-order send/recv operation for each 128 bytes
+  aligned block. Enabling the option will guarantee data inside each 128 bytes
+  aligned block being sent and received in order. If endpoint is not able to
+  support this feature, it will return -FI_EOPNOTSUPP for the call to fi_setopt().
+
+*FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES - bool*
+: It is used to set the endpoint to use in-order RDMA write operation for each 128 bytes
+  aligned block. Enabling the option will guarantee data inside each 128 bytes
+  aligned block being written in order. If endpoint is not able to support
+  this feature, it will return -FI_EOPNOTSUPP for the call to fi_setopt().
+
 # RUNTIME PARAMETERS
 
 *FI_EFA_TX_SIZE*
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -13,6 +13,8 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	efa_h_enable_poisoning=0
 	# Determine if efa device supports extensible CQ with access to additional fields from WC
 	efadv_support_extended_cq=1
+	# Determine if efa device support in-order write at 128 bytes aligned boundary.
+	efa_support_data_in_order_aligned_128_byte=1
 
 	AS_IF([test x"$enable_efa" != x"no"],
 	      [FI_CHECK_PACKAGE([efa_ibverbs],
@@ -169,6 +171,21 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 		)
 
 	AM_CONDITIONAL([HAVE_EFADV_CQ_EX], [test $efadv_support_extended_cq -eq 1])
+
+	# IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES is the flag we need test,
+	# but it is flag. If it is not there, we should disable it.
+	AS_IF([test $efa_support_data_in_order_aligned_128_byte -eq 1],
+		[AC_EGREP_HEADER(
+			[IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES],
+			[infiniband/verbs.h],
+			[],
+			[efa_support_data_in_order_aligned_128_byte=0])
+		])
+
+	AS_IF([test $efa_support_data_in_order_aligned_128_byte -eq 1],
+		[AC_DEFINE([HAVE_EFA_DATA_IN_ORDER_ALIGNED_128_BYTES], [1], [EFA device supports 128 bytes in-order in writing.])],
+		[AC_DEFINE([HAVE_EFA_DATA_IN_ORDER_ALIGNED_128_BYTES], [0], [EFA device does not support 128 bytes in-order in writing.])]
+		)
 
 	efa_CPPFLAGS="$efa_ibverbs_CPPFLAGS $efadv_CPPFLAGS"
 	efa_LDFLAGS="$efa_ibverbs_LDFLAGS $efadv_LDFLAGS"
--- a/prov/efa/src/efa_base_ep.c
+++ b/prov/efa/src/efa_base_ep.c
@@ -173,15 +173,12 @@ static int efa_base_ep_modify_qp_rst2rts(struct efa_base_ep *base_ep,
 					   IBV_QP_STATE | IBV_QP_SQ_PSN);
 }
 
-static int efa_base_ep_create_qp_ex(struct efa_base_ep *base_ep,
-			     struct ibv_qp_init_attr_ex *init_attr_ex)
+int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
+			  struct ibv_qp_init_attr_ex *init_attr_ex)
 {
-	struct efa_domain *domain;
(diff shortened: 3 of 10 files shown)


prov/efa/src/efa_base_ep.c
```diff
@@ -342,3 +342,23 @@ int efa_base_ep_getname(fid_t fid, void *addr, size_t *addrlen)
...
+bool efa_base_ep_support_op_in_order_aligned_128_bytes(struct efa_base_ep *base_ep, enum ibv_wr_opcode op)
+{
+#if HAVE_EFA_DATA_IN_ORDER_ALIGNED_128_BYTES
+	return ibv_query_qp_data_in_order(base_ep->qp,
+					 op,
+					 IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES);
+#else
+	return false;
+#endif
+}
```
wenduwan, line 364: I think the convention in libfabric is to have 2 functions based on the macro.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1117,6 +1120,47 @@ static int rxr_ep_set_cuda_api_permitted(struct rxr_ep *ep, bool cuda_api_permit
...
+ * @brief set sendrecv_in_order_aligned_128_bytes flag in rxr_ep
+ * Supporting send/receive aligned 128 bytes buffer in order is more complicated than
+ * supporting RDMA write. There is no plan to support it in the near future.
+ * Therefore this function always returns -FI_EOPNOTSUPP if the user tries to enable it.
+ *
+ * @param[in,out]	ep					endpoint
+ * @param[in]		sendrecv_in_order_aligned_128_bytes	whether to enable in_order send/recv
+ *								for each 128 bytes aligned buffer
+ * @return		0 on success, -FI_EOPNOTSUPP if the option cannot be supported
+ */
```
wenduwan, line 1133: This is not intuitive, even with extensive explanation.

Suggest being expressive with the function name, e.g.
```
rxr_ep_validate_sendrecv_in_order_aligned_128_bytes_unset(...)
```

man/fi_efa.7.md
```diff
@@ -103,6 +103,20 @@ provider for AWS Neuron or Habana SynapseAI.
   emulating Read, Write, and Atomic operations (return value is true), or if
   these operations are assisted by hardware support (return value is false).
 
+*FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES - bool*
+: This option only apply to the fi_setopt() call.
```
lrbison, line 107: only applies

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -1117,6 +1120,47 @@ static int rxr_ep_set_cuda_api_permitted(struct rxr_ep *ep, bool cuda_api_permit
...
+ *								for each 128 bytes aligned buffer.
+ * @return		0 on success, -FI_EOPNOTSUPP if the option cannot be supported.
+ */
+static
+int rxr_ep_set_write_in_order_aligned_128_bytes(struct rxr_ep *ep,
+						bool write_in_order_aligned_128_bytes)
+{
+	if (write_in_order_aligned_128_bytes &&
+	    !efa_base_ep_support_op_in_order_aligned_128_bytes(&ep->base_ep, IBV_WR_RDMA_WRITE))
+		return -FI_EOPNOTSUPP;
```
wenduwan, line 1158: Is this correct? If the user sets`write_in_order_aligned_128_bytes=0`, should we ever return `-FI_EOPNOTSUPP`?

man/fi_efa.7.md
```diff
@@ -103,6 +103,20 @@ provider for AWS Neuron or Habana SynapseAI.
...
+  It is used to force the endpoint to use in-order send/recv operation for each 128 bytes
+  aligned block. Enabling the option will guarantee data inside each 128 bytes
+  aligned block being sent and received in order. If endpoint is not able to
```
lrbison, line 110: Questions:
 - Does setting this option apply to local receives or remote deliveries?  is the flag required on both ends?
 - is the alignment also 128 bytes?

man/fi_efa.7.md
```diff
@@ -103,6 +103,20 @@ provider for AWS Neuron or Habana SynapseAI.
...
+  support this feature, it will return -FI_EOPNOTSUPP for the call to fi_setopt().
+
+*FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES - bool*
+: This option only apply to the fi_setopt() call.
```
lrbison, line 114: only applies.
Why can't the application query the result with getopt?

prov/efa/configure.m4
```diff
@@ -170,6 +172,21 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 
 	AM_CONDITIONAL([HAVE_EFADV_CQ_EX], [test $efadv_support_extended_cq -eq 1])
 
+	# IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES is the flag we need test,
+	# but it is flag. If it is not there, we should disable it.
+	AS_IF([test $efa_support_data_in_order_aligned_128_byte -eq 1],
```
lrbison, line 177: I'm not great with .m4 files, but it looks to me like the test here should be on a different variable name?  It looks to me like you will  never bother to grep the header because on entry the variable will be 0.

prov/efa/src/efa_base_ep.c
lrbison: I like it split like this for small functions.

Although it would be nice to be consistent with efa_device_support_rdma_write below

prov/efa/src/rdm/rxr_ep.c
lrbison: The function makes sense to me, but I'm confused as to why we would want to introduce such an option with no plans for support.  I suppose because we want to distinguish it from the _RDMA variant, but RDMA variant's name seems sufficiently specific to avoid confusion.

Do we really need the sendrecv variant?

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -2397,6 +2455,12 @@ int rxr_endpoint(struct fid_domain *domain, struct fi_info *info,
 	}
 
 	rxr_ep->cuda_api_permitted = (FI_VERSION_GE(info->fabric_attr->api_version, FI_VERSION(1, 18)));
+	rxr_ep->sendrecv_in_order_aligned_128_bytes = false;
+	rxr_ep->write_in_order_aligned_128_bytes = false;
```
lrbison, line 2459: Are these values uninitialized if API version is <1.18, or did I miss that?

prov/efa/src/rdm/rxr_ep.h
```diff
@@ -245,6 +245,8 @@ struct rxr_ep {
 	struct fid_peer_srx peer_srx; /* support sharing receive context with peer providers */
 	bool cuda_api_permitted; /**< whether end point is permitted to call CUDA API */
 	struct fi_info *user_info; /**< fi_info passed by user when calling fi_endpoint */
+	bool sendrecv_in_order_aligned_128_bytes; /**< whether to support in order send/recv of each aligned 128 bytes memory region */
```
lrbison, line 248: what is the `/**<` syntax?  Is that doxygen?

prov/efa/src/rdm/rxr_ep.c
wzamazon: > Do we really need the sendrecv variant?

In the future, I think we may implement this feature for sendrecv, so it is better to introduce it now.

prov/efa/src/rdm/rxr_ep.h
wzamazon: Yes. that is doxygen.

prov/efa/src/rdm/rxr_ep.c
wzamazon: There are always initialized here.

prov/efa/configure.m4
wzamazon: The variable `efa_support_data_in_order_aligned_128_byte` was initialized to 1 earlier.

man/fi_efa.7.md
wzamazon: implemented getopt

prov/efa/src/efa_base_ep.c
wzamazon: Had separate functions.

man/fi_efa.7.md
wzamazon: Implemented getopt, so removing this

wenduwan (approved):

PR #8683, opened 2023-03-21
prov/efa: fix string initialization bug in unit test

wenduwan:
Fix a bug in unit test resulting in a string not ending properly with the null terminator.

Also add a null check before deleting the tempfile.

### DIFF
--- a/prov/efa/test/efa_unit_test_ep.c
+++ b/prov/efa/test/efa_unit_test_ep.c
@@ -12,19 +12,26 @@
  */
 void test_rxr_ep_host_id(struct efa_resource **state, bool file_exists, char *raw_id, uint64_t expect_id)
 {
+	int fd;
+	ssize_t written_len;
+	char host_id_file[] = "XXXXXXXXXX";
 	struct efa_resource *resource = *state;
 	struct rxr_ep *rxr_ep;
-	size_t file_name_length = 10;
-	char host_id_file[file_name_length];
-	FILE *f_handle;
 
 	rxr_env.host_id_file = NULL;
 
 	if (file_exists) {
-		new_temp_file(host_id_file, file_name_length);
-		f_handle = fopen(host_id_file, "w");
-		fprintf(f_handle, raw_id);
-		fclose(f_handle);
+		fd = mkstemp(host_id_file);
+		if (fd < 0) {
+			fail();
+		}
+
+		written_len = write(fd, raw_id, strlen(raw_id));
+		if (written_len != strlen(raw_id)) {
+			close(fd);
+			fail();
+		}
+
 		rxr_env.host_id_file = host_id_file;
 	}
 
@@ -33,7 +40,10 @@ void test_rxr_ep_host_id(struct efa_resource **state, bool file_exists, char *ra
 	rxr_ep = container_of(resource->ep, struct rxr_ep, base_ep.util_ep.ep_fid);
 
 	/* Remove the temporary file */
-	unlink(host_id_file);
+	if (rxr_env.host_id_file) {
+		unlink(rxr_env.host_id_file);
+		close(fd);
+	}
 
 	assert_int_equal(rxr_ep->host_id, expect_id);
 }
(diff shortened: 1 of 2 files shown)


wckzhang (approved):

sunkuamzn (approved):

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -12,19 +12,38 @@
...
+		host_id_file = malloc(host_id_file_length + 1);
+		assert_non_null(host_id_file);
+
+		for (int i = 0; i < host_id_file_length; ++i) {
+			host_id_file[i] = 'X';
+		}
+
+		host_id_file[host_id_file_length] = '\0';
+
+		fd = mkstemp(host_id_file);
```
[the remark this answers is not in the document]
wenduwan: Could you elaborate? `mkstemp` could fail, do you suggest we do not need to check the failure? Also `close(fd)` makes the file inaccessible so it should only be called after the test if the file was actually created successfully.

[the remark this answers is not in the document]
lrbison: I was suggesting the highlighted section of code could be replaced with the first 3 lines of my snippet.  No need to catch an error for malloc (I guess just the one error).  Whatever you are doing with fd afterwards would remain the same, but if you can't mkstemp, then obviously the test will fail.

As far as closing the file, I think the logic you have is fine, just trying to point out you wouldn't need a free() as well as a close().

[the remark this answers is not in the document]
wenduwan: I don't know how it could fail but `snprintf` could return a negative error code. Are you suggesting we ignore the failure and proceed to `mkstemp` anyways, assuming that the failure is not significant as long as `mkstemp` works?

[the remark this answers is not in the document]
lrbison: If you prefer to keep the loop or do a memset or whatever, that's fine.  Mostly I prefer to avoid the unnecessary alloc

lrbison (approved):

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -12,19 +12,38 @@
...
+		if (fd < 0) {
+			free(host_id_file);
+			fail();
+		}
+
+		written_len = write(fd, raw_id, strlen(raw_id));
```
lrbison, line 40: now that fd is not a FILE*, I think I would do a fprintf to the fd.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -12,19 +12,38 @@
...
-	FILE *f_handle;
 
 	rxr_env.host_id_file = NULL;
 
 	if (file_exists) {
-		new_temp_file(host_id_file, file_name_length);
-		f_handle = fopen(host_id_file, "w");
-		fprintf(f_handle, raw_id);
-		fclose(f_handle);
+		host_id_file = malloc(host_id_file_length + 1);
```
wzamazon, line 25: Why not just use a string  like

```
char host_id_file[] = "XXXXXXXXX"
```
?

wenduwan: > Why not just use a string like

Thanks! I tried `char *host_id_file = "xxxxx"` and it does not work for `mkstemp`, but `host_id_file[]` does!

Will update.

wzamazon (approved):

prov/efa/test/efa_unit_test_ep.c
lrbison: @wenduwan @wzamazon 

> char host_id_file[] = "XXXXXXXXX"

My understanding is that mkstemp would modify such a string in-place, and that `host_id_file[]` and `char *host_id_file` are both equivalent in that they simply point to a portion of the binary which has the string "XXXXXXXXXX" sitting.  I would then expect that the first call to mkstemp would succeed in modifying the string to "XXXX7d8c7s" but then subsequent calls would fail because the (persistent) string doesn't end in 6 "X" characters.

Maybe `char host_id_file[]` is actually setting some local memory each time??  I'm surprised this PR works as is.

wenduwan: So far the tests have passed(with multiple calls to mkstemp) so I think this issue did not happen.

My understanding is `char host_id_file[]`(not static) is reallocated in each test call, so it is always initialized to `XXXXXXXXXX`

PR #8701, opened 2023-03-22
doc/fi_getinfo: add FI_ENOSYS to return value

wzamazon:
Current implementation of fi_getinfo() will return -FI_ENOSYS if the "version" argument of fi_getinfo() is too high to be supported. This patch added that return value to document.


shefty (approved):

wzamazon:
This is a pure document change. I will ignore CI.

shefty:
Can you pick for v1.18?

PR #8705, opened 2023-03-22
prov/tcp: Fix a use-after if ep closed after reject

sydidelot:
The following sequence causes a use-after-free that can be
detected by ASAN:
- Incoming connection request received on PEP
- Create the endpoint with fi_endpoint(info)
- Reject the connection request with fi_reject(pep, info->handle);
- Close the endpoint with fi_close(&ep->fid)

The use-after free occurs in tcpx_ep_close() when the following
condition is evaluated:
if (ep->fid && ep->fid->fclass == TCPX_CLASS_CM)

The pointer stored in ep->fid references a memory that is freed.
When tcpx_endpoint() is executed the EP saves a reference
to the handle in ep->handle. Later, this handle is freed
by fi_reject() and ep->handle becomes a dangling pointer as
a consequence.
Finally, tcpx_ep_close() accesses ep->fid, which triggers
the use-after-free (note that ep->fid and ep->handle are part
of the same union).

The fix implemented in the patch moves the destruction of
ep->handle from fi_reject() to fi_close().

### DIFF
--- a/prov/tcp/src/tcpx_ep.c
+++ b/prov/tcp/src/tcpx_ep.c
@@ -627,8 +627,10 @@ static int tcpx_ep_close(struct fid *fid)
 					 &ep->util_ep.ep_fid.fid);
 	}
 
-	if (ep->fid && ep->fid->fclass == TCPX_CLASS_CM)
+	if (ep->state != TCPX_RCVD_REQ && ep->fid) {
+	        assert(ep->fid->fclass == TCPX_CLASS_CM);
 		tcpx_free_cm_ctx(ep->cm_ctx);
+	}
 
 	ofi_close_socket(ep->bsock.sock);
 	if (ep->util_ep.eq)


sydidelot:
@shefty The bug was found in the tcp provider but it also exists in the net provider.
I will write a patch for the net provider as soon as I get your approval for the fix.
Thank you!

prov/tcp/src/tcpx_ep.c
```diff
@@ -217,6 +217,7 @@ tcpx_ep_accept(struct fid_ep *ep_fid, const void *param, size_t paramlen)
 	    !handle || (handle->fid.fclass != FI_CLASS_CONNREQ))
 		return -FI_EINVAL;
 
+	free(handle);
```
shefty, line 220: The thinking here is if fi_accept() fails, the application will call fi_reject().  So the handle is only freed on success.

shefty (changes requested):
We had problems with the handle destruction before, which is why it is only freed from either accept or reject.  The app is required to call at least one.  There's a problem if the app creates an ep, hits some failure or checks, closes the ep, *then* call fi_reject().  If we free the handle on close, fi_reject() now references invalid memory.

shefty:
I don't see that net (new tcp) has this issue.  There's not a similar check in xnet_ep_close().  Can we check the ep state in close to determine if we're in the situation where the ep was created then closed, prior to the app calling accept?

prov/tcp/src/tcpx_ep.c
```diff
@@ -1011,12 +1015,7 @@ static int tcpx_pep_reject(struct fid_pep *pep, fid_t fid_handle,
 
 	ofi_shutdown(handle->sock, SHUT_RDWR);
 	ret = ofi_close_socket(handle->sock);
-	if (ret)
-		return ret;
-
-free:
-	free(handle);
```
shefty, line 1018: The user can call reject without allocating an endpoint.  In this case, we'll leak the handle allocation.

prov/tcp/src/tcpx_ep.c
```diff
@@ -627,8 +627,12 @@ static int tcpx_ep_close(struct fid *fid)
 					 &ep->util_ep.ep_fid.fid);
 	}
 
-	if (ep->fid && ep->fid->fclass == TCPX_CLASS_CM)
+	if (ep->state == TCPX_RCVD_REQ && ep->fid) {
+		assert(ep->fid->fclass == FI_CLASS_CONNREQ);
+		free(ep->fid);
+	} else if (ep->fid && ep->fid->fclass == TCPX_CLASS_CM) {
 		tcpx_free_cm_ctx(ep->cm_ctx);
+	}
```
shefty, line 635: I think we'll need this to be:
```
if (ep->state != TCPX_RCVD_REQ && ep->fid) {
    assert(ep->fid->fclass == FI_CLASS_CM);
    tcpx_free_cm_ctx(ep->cm_ctx);
}
```
see comment below.

Though, we're really trying to free ep->cm_ctx.  Could we just check that directly instead?

sydidelot: Your suggestion makes sense. Let me update the code and re-upload.

sydidelot: PR was updated with your suggested and commit message was updated accordingly. Thanks again!

shefty: Are you able to confirm this fixes the ASAN report?

shefty (approved):
thanks - it looks correct :)

prov/tcp/src/tcpx_ep.c
sydidelot: I haven't tried to reproduce the bug for now. The bug occurs in a state that is difficult to reproduce. 
I think we can easily trigger the bug and verify the patch with some new tests in fabtests. 
I will look at it later today.

prov/tcp/src/tcpx_ep.c
```diff
@@ -627,8 +627,10 @@ static int tcpx_ep_close(struct fid *fid)
 					 &ep->util_ep.ep_fid.fid);
 	}
 
-	if (ep->fid && ep->fid->fclass == TCPX_CLASS_CM)
+	if (ep->state != TCPX_RCVD_REQ && ep->fid) {
+	        assert(ep->fid->fclass == FI_CLASS_CM);
```
shefty, line 631: oops - should be TCPX_CLASS_CM - not FI

sydidelot: Updated. Thanks!

prov/tcp/src/tcpx_ep.c
sydidelot: FYI, I managed to reproduce the bug with some manual fault injection in my code.
And I confirm the PR fixes the user-after-free detected by ASAN :+1: .

shefty:
thanks for the fix and verifying it

PR #8706, opened 2023-03-22
[v1.18.x] man/getinfo: include FI_ENOSYS in return values

wzamazon:
Current implementation of fi_getinfo() will return -FI_ENOSYS if the "version" argument of fi_getinfo() is too high to be supported. This patch added that return value to document.

(cherry picked from commit a923f34e9975e0b31b7896777e344baff3669963)


shefty:
thanks

PR #8708, opened 2023-03-23
prov/efa: Use shm as a peer provider

shijin-aws:
This PR contains a series of commits that allow efa use shm as a peer provider to offload the intra-node communication.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -51,15 +51,26 @@ const char *efa_rdm_cq_strerror(struct fid_cq *cq_fid, int prov_errno,
 static
 int rxr_cq_close(struct fid *fid)
 {
-	int ret;
-	efa_rdm_cq *cq;
+	int ret, retv;
+	struct efa_rdm_cq *cq;
 
-	cq = container_of(fid, efa_rdm_cq, cq_fid.fid);
-	ret = ofi_cq_cleanup(cq);
+	retv = 0;
+
+	cq = container_of(fid, struct efa_rdm_cq, util_cq.cq_fid.fid);
+
+	if (cq->shm_cq) {
+		ret = fi_close(&cq->shm_cq->fid);
+		if (ret) {
+			EFA_WARN(FI_LOG_CQ, "Unable to close shm cq: %s\n", fi_strerror(-ret));
+			retv = ret;
+		}
+	}
+
+	ret = ofi_cq_cleanup(&cq->util_cq);
 	if (ret)
 		return ret;
 	free(cq);
-	return 0;
+	return retv;
 }
 
 static struct fi_ops efa_rdm_cq_fi_ops = {
@@ -70,10 +81,28 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count, fi_addr_t *src_addr)
+{
+	struct efa_rdm_cq *cq;
+	ssize_t ret;
+
+	cq = container_of(cq_fid, struct efa_rdm_cq, util_cq.cq_fid.fid);
+
+	if (cq->shm_cq)
+		fi_cq_read(cq->shm_cq, NULL, 0);
+
+	ret = ofi_cq_read_entries(&cq->util_cq, buf, count, src_addr);
+
+	if (ret > 0)
+		return ret;
+
+	return ofi_cq_readfrom(&cq->util_cq.cq_fid, buf, count, src_addr);
+}
+
 static struct fi_ops_cq efa_rdm_cq_ops = {
 	.size = sizeof(struct fi_ops_cq),
 	.read = ofi_cq_read,
-	.readfrom = ofi_cq_readfrom,
+	.readfrom = efa_rdm_cq_readfrom,
 	.readerr = ofi_cq_readerr,
 	.sread = fi_no_cq_sread,
 	.sreadfrom = fi_no_cq_sreadfrom,
@@ -98,8 +127,10 @@ int efa_rdm_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 		    struct fid_cq **cq_fid, void *context)
 {
 	int ret;
-	efa_rdm_cq *cq;
+	struct efa_rdm_cq *cq;
 	struct efa_domain *efa_domain;
+	struct fi_cq_attr shm_cq_attr = {0};
+	struct fi_peer_cq_context peer_cq_context = {0};
 
 	if (attr->wait_obj != FI_WAIT_NONE)
 		return -FI_ENOSYS;
@@ -113,15 +144,31 @@ int efa_rdm_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 	/* Override user cq size if it's less than recommended cq size */
 	attr->size = MAX(efa_domain->rdm_cq_size, attr->size);
 
-	ret = ofi_cq_init(&efa_prov, domain, attr, cq,
+	ret = ofi_cq_init(&efa_prov, domain, attr, &cq->util_cq,
 			  &ofi_cq_progress, context);
 
 	if (ret)
 		goto free;
 
-	*cq_fid = &cq->cq_fid;
+	*cq_fid = &cq->util_cq.cq_fid;
 	(*cq_fid)->fid.ops = &efa_rdm_cq_fi_ops;
 	(*cq_fid)->ops = &efa_rdm_cq_ops;
+
+	/* open shm cq as peer cq */
+	if (efa_domain->shm_domain) {
+		memcpy(&shm_cq_attr, attr, sizeof(*attr));
+		/* Bind ep with shm provider's cq */
+		shm_cq_attr.flags |= FI_PEER;
+		peer_cq_context.size = sizeof(peer_cq_context);
+		peer_cq_context.cq = cq->util_cq.peer_cq;
+		ret = fi_cq_open(efa_domain->shm_domain, &shm_cq_attr,
+				 &cq->shm_cq, &peer_cq_context);
+		if (ret) {
+			EFA_WARN(FI_LOG_CQ, "Unable to open shm cq: %s\n", fi_strerror(-ret));
+			goto free;
+		}
+	}
+
 	return 0;
 free:
 	free(cq);
--- a/prov/efa/src/rdm/rxr_ep.c
+++ b/prov/efa/src/rdm/rxr_ep.c
@@ -468,7 +468,9 @@ static void rxr_ep_free_res(struct rxr_ep *rxr_ep)
 		EFA_WARN(FI_LOG_EP_CTRL,
 			"Closing ep with unmatched unexpected rx_entry: %p pkt_entry %p\n",
 			rx_entry, rx_entry->unexp_pkt);
-		rxr_pkt_entry_release_rx(rxr_ep, rx_entry->unexp_pkt);
+		/* rx entry for peer srx does not allocate unexp_pkt */
+		if (!(rx_entry->rxr_flags & RXR_RX_ENTRY_FOR_PEER_SRX))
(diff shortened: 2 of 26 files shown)


sunkuamzn:
CI failures are all `osu_latency_mt`. Should be fixed when @darrylabbate merges his changes.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -843,12 +800,13 @@ void rxr_ep_set_use_shm_for_tx(struct rxr_ep *ep)
...
 	 *
-	 * Using the shm provider comes with some overheads, particularly in the
-	 * progress engine when polling an empty completion queue, so avoid
+	 * Using the shm provider comes with some overheads, so avoid
 	 * initializing the provider if the app provides a hint that it does not
 	 * require node-local communication. We can still loopback over the EFA
 	 * device in cases where the app violates the hint and continues
 	 * communicating with node-local peers.
+	 *
+	 * aws-ofi-nccl relies on this feature.
```
wenduwan, line 809: Is this comment from another commit?

shijin-aws: I could happen to add this comment when I rebase. Let me look at it.

prov/efa/src/rdm/rxr_msg.h
```diff
@@ -31,6 +31,44 @@
...
+/**
+ * @brief Populate the fields in fi_peer_rx_entry object
+ * with the equivalences in rxr_op_entry
+ *
+ * @param[in,out] peer_rx_entry the fi_peer_rx_entry object to be updated
+ * @param[in] rx_entry the rxr_op_entry
+ * @param[in] op the ofi op code
+ */
+static inline
+void rxr_msg_update_peer_rx_entry(struct fi_peer_rx_entry *peer_rx_entry,
```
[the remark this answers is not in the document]
shijin-aws: Yes, your concern is real. We should not make `shm` explicit here. But there is a blocker as we have to convert the efa_mr desc to shm_mr desc as efa_mr has its own format and could not be understood by shm. Currently the convert function explicitly uses `efa_mr->shm_mr` to make the transition. If we want to get rid of shm completely here we would have to propose a shared MR API

shijin-aws:
>  Add Peer Provider API to EFA
    Switch EFA to use SHM as Peer Provider
    Remove old SHM code

For commit 1, there is already some commits merged like peer_ops, for commit 2, I could squash the current middle ones. I still prefer to make `fix rx flags` commit separate because it is a bug in current efa provider.

shijin-aws:
>CI failures are all osu_latency_mt. Should be fixed when @darrylabbate merges his changes.

I would have to look at it. It seems consistently hit for my PR.

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -233,6 +207,55 @@ ssize_t rxr_msg_generic_send(struct fid_ep *ep, const struct fi_msg *msg,
...
+		shm_msg = (struct fi_msg *) msg;
+		shm_msg->addr = peer->shm_fiaddr;
+		if (msg->desc) {
+			memcpy(shm_desc, msg->desc, msg->iov_count * sizeof(void *));
+			rxr_convert_desc_for_shm(msg->iov_count, shm_desc);
+			shm_msg->desc = shm_desc;
+		} else {
+			shm_msg->desc = NULL;
+		}
+		if (op == ofi_op_msg) {
```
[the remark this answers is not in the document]
shijin-aws: This could reduce code duplication, but I think it could cause more instructions executed. But if you make shm call in a higher level, all the changes @wenduwan  made in EFA for gdrcopy hacks would be copied and pasted everywhere.

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -233,6 +207,55 @@ ssize_t rxr_msg_generic_send(struct fid_ep *ep, const struct fi_msg *msg,
 		goto out;
 	}
 
+	if (peer->is_local && rxr_ep->use_shm_for_tx) {
+		/*
+		 * AWS Neuron and SynapseAI are currently not supported by the SHM provider.
```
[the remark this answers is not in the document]
bwbarrett: I'm not sure this comment is really right, even for CUDA.  If you can't use CUDA, you don't have a sane way of reading data from accelerator memory that doesn't use the network (PCIe reads over GDRCopy or similar are not sane).  I think there's a nature of ugly here, when you have a NCCL-like app that breaks CUDA expectations.

prov/efa/src/rdm/rxr_msg.h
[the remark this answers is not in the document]
shijin-aws: Never mind, I think I have a solution to fix your concern.

prov/efa/src/rdm/rxr_msg.c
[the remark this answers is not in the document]
shijin-aws: I see, a check is still needed whether shm can be used. I will keep it but refactor it to be more clean.

prov/efa/src/rdm/rxr_ep.c
sunkuamzn: Yes, I added this line. The comment is about why we don't enable SHM by default. The reason provided is that the additional CQ polling adds additional overhead when SHM is enabled. With peer API, that's no longer a valid reason. However, we still cannot enable SHM by default because OFI NCCL plugin does not want SHM.

prov/efa/src/rdm/rxr_msg.c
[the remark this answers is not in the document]
shijin-aws: I was not realizing that a recent merged PR handles this check already. See https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/rxr_ep.c#L868-L881. So this extra `if` was not actually required.

prov/efa/src/rdm/rxr_msg.c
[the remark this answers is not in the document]
shijin-aws: Fixed it in the newest revision.

prov/efa/src/rdm/rxr_ep.c
```diff
@@ -427,6 +383,30 @@ struct rxr_op_entry *rxr_ep_alloc_tx_entry(struct rxr_ep *rxr_ep,
...
+ * EFA provider is of struct efa_mr *, which shm provider cannot
+ * understand. This function convert EFA descriptors to descriptors
+ * shm can use.
+ *
+ * @param numdesc[in]       number of descriptors in the array
+ * @param efa_desc[in]      efa descriptors
+ * @param shm_desc[out]     shm descriptors
+ *                          is shm descriptor.
+ */
+void rxr_get_desc_for_shm(int numdesc, void **efa_desc, void **shm_desc)
```
wenduwan, line 399: Largely duplicates `rxr_convert_desc_for_shm` - can we merge them and only keep 1?

prov/efa/src/rdm/rxr_msg.c
```diff
@@ -283,6 +252,34 @@ static
...
+	struct efa_rdm_peer *peer;
+	struct rxr_ep *rxr_ep;
+	void *shm_desc[RXR_IOV_LIMIT] = {NULL};
+	struct fi_msg *shm_msg;
+	void **efa_desc = NULL;
+	fi_addr_t efa_addr;
+	int ret;
+
+	rxr_ep = container_of(ep, struct rxr_ep, base_ep.util_ep.ep_fid.fid);
+	peer = rxr_ep_get_peer(rxr_ep, msg->addr);
```
wenduwan, line 264: Come to think about it, now that we have peer api, we can do a separate refactoring to encapsulate the peer ops in rdm peer.

The code as-is is rather verbose and repetitive. Looking at the change stat, when the logic was in generic send, we were removing 100+ LOC overall. After we scattered the logic in each send api we are adding 100+ LOC.

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -219,6 +172,25 @@ ssize_t rxr_rma_readmsg(struct fid_ep *ep, const struct fi_msg_rma *msg, uint64_
 		err = -FI_EAGAIN;
 		goto out;
 	}
+
+	if (peer->is_local && rxr_ep->use_shm_for_tx) {
+		tmp_addr = msg->addr;
+		tmp_desc = msg->desc;
+		msg_clone = (struct fi_msg_rma *)msg;
```
wenduwan, line 179: Is there a way in peer api to translate MR desc between peers, instead of 'cloning' and resetting msg?

wenduwan:
Change is clear and straightforward. Left minor comments. Please unblock AWS CI.

Ready to approve.

prov/efa/src/rdm/rxr_rma.c
shijin-aws: No such API exists, we can propose one, but in my experiment this desc convert does not impact performance

prov/efa/src/rdm/rxr_ep.c
shijin-aws: That should be the case in my comment https://github.com/ofiwg/libfabric/pull/8708#discussion_r1149684783, which I think could involve more function call stacks.

shijin-aws:
windows test failure is also fixed.

prov/efa/src/rdm/rxr_ep.c
wenduwan: After this change the only remaining usage of `rxr_convert_desc_for_shm` is [here](https://github.com/ofiwg/libfabric/blob/2b33134370e1634c1612104865aa429eaadab102/prov/efa/src/rdm/rxr_atomic.c#L61)

I suggest we get rid of that and only keep `rxr_get_desc_for_shm`

shijin-aws: That can be doable. Will update.

shijin-aws: Latest revision uses `rxr_get_desc_for_shm` everywhere and gets rid of `rxr_convert_desc_for_shm`

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -70,10 +81,28 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
...
+	ssize_t ret;
+
+	cq = container_of(cq_fid, struct efa_rdm_cq, util_cq.cq_fid.fid);
+
+	if (cq->shm_cq)
+		fi_cq_read(cq->shm_cq, NULL, 0);
+
+	ret = ofi_cq_read_entries(&cq->util_cq, buf, count, src_addr);
+
+	if (ret > 0)
```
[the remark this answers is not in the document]
shijin-aws: I can run WRF on single node to verify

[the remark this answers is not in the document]
shijin-aws: I ran both WRF and a collective benchmark (OMB allreduce) on 4 c5n, 144 ranks.

WRF:
- Return early when cqe is written by shm: 321.98 sec wall time
- Always progress shm + efa before returning: 321.65 sec wall time.
This difference is within the fluctuation range, i.e. there is just no difference.

I also ran OMB allreduce in the same number of ranks

- Return early when cqe is written by shm
```
# OSU MPI Allreduce Latency Test v5.6.2
# Size       Avg Latency(us)
4                     139.38
8                     141.35
16                    141.53
32                    141.39
64                    142.30
128                   141.81
256                   143.74
512                   145.61
1024                  148.95
2048                  159.29
4096                  175.47
8192                  214.72
16384                 273.68
32768                 356.58
65536                 328.73
131072               1516.84
262144                725.51
524288               1254.14
1048576              2442.72
```

- always progress shm + efa before returning
```
# OSU MPI Allreduce Latency Test v5.6.2
# Size       Avg Latency(us)
4                     156.22
8                     142.97
16                    142.86
32                    141.57
64                    150.73
128                   142.74
256                   143.77
512                   145.50
1024                  150.37
2048                  158.36
4096                  179.06
8192                  215.56
16384                 270.32
32768                 355.54
65536                 328.54
131072               1518.15
262144                723.39
524288               1251.70
1048576              2488.29
```
In this benchmark, `Return early when cqe is written by shm` is either better or equal than the latter one.

wzamazon (approved):

wenduwan (approved):

PR #8709, opened 2023-03-23
prov/efa,hmem_neunron: bug fixes

wzamazon:
fix bug in hmem_neuron and prov/efa in handling neuron


src/hmem_neuron.c
```diff
@@ -143,11 +150,24 @@ int neuron_host_unregister(void *ptr)
...
+	uint32_t total_nc_count;
 
 	ret = neuron_dl_init();
 	if (ret)
 		return ret;
 
+	/* Note that nrt_get_total_nc_count() is one of the few neuron functions
+	 * that can be called before nrt_init()
+	 */
+	if (neuron_ops.nrt_get_total_nc_count(&total_nc_count) != NRT_SUCCESS) {
```
shijin-aws, line 162: Is it something neuron team suggested, or we are following what cuda does in init?

shijin-aws (approved):

src/hmem_neuron.c
wzamazon: We are following what cuda does in init.

PR #8718, opened 2023-03-27
core/util_cq: split the cq progress and reading entries in ofi_cq_readfrom

shijin-aws:
Currently, ofi_cq_readfrom always progress the cq before reading the cq entries.
    When an owner provider is sharing its cq to a peer provider, it may first progress
    the peer provider cq via `fi_cq_read(peer_cq, NULL, 0)`. If a cqe is already written
    after the progress, the owner provider can just return it without progressing its owner
    cq again. So it is ideal to introduce a method to purely read the cq entries without
    progressing the cq. This patch is intended to make that goal. Based on this patch, the
    owner provider can return immediately after fi_cq_read(peer_cq, NULL, 0) is called and
    ofi_cq_read_entries returns a postive integer.

### DIFF
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -554,6 +554,63 @@ int ofi_cq_write_overflow(struct util_cq *cq, void *context, uint64_t flags,
 			  size_t len, void *buf, uint64_t data, uint64_t tag,
 			  fi_addr_t src);
 
+static inline
+ssize_t ofi_cq_read_entries(struct util_cq *cq, void *buf, size_t count,
+			fi_addr_t *src_addr)
+{
+	struct fi_cq_tagged_entry *entry;
+	struct util_cq_aux_entry *aux_entry;
+	ssize_t i;
+
+	ofi_genlock_lock(&cq->cq_lock);
+	if (ofi_cirque_isempty(cq->cirq)) {
+		i = -FI_EAGAIN;
+		goto out;
+	}
+
+	if (count > ofi_cirque_usedcnt(cq->cirq))
+		count = ofi_cirque_usedcnt(cq->cirq);
+
+	for (i = 0; i < (ssize_t) count; i++) {
+		entry = ofi_cirque_head(cq->cirq);
+		if (!(entry->flags & UTIL_FLAG_AUX)) {
+			if (src_addr && cq->src)
+				src_addr[i] = cq->src[ofi_cirque_rindex(cq->cirq)];
+			cq->read_entry(&buf, entry);
+			ofi_cirque_discard(cq->cirq);
+		} else {
+			assert(!slist_empty(&cq->aux_queue));
+			aux_entry = container_of(cq->aux_queue.head,
+						 struct util_cq_aux_entry,
+						 list_entry);
+			assert(aux_entry->cq_slot == entry);
+			if (aux_entry->comp.err) {
+				if (!i)
+					i = -FI_EAVAIL;
+				break;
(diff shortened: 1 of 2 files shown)


include/ofi_util.h
```diff
@@ -554,6 +554,63 @@ int ofi_cq_write_overflow(struct util_cq *cq, void *context, uint64_t flags,
 			  size_t len, void *buf, uint64_t data, uint64_t tag,
 			  fi_addr_t src);
 
+static inline
+ssize_t ofi_cq_read_cq(struct util_cq *cq, void *buf, size_t count,
+			fi_addr_t *src_addr)
```
wzamazon, line 559: it seems the function mainly flush the cirque, can we name it `ofi_cq_flush_cirque` to be more specific?

a-szegel: If this is going to be implemented by SM2, there is no circular queue... I think cq is overloaded for completion queue

prov/util/src/util_cq.c
```diff
@@ -250,62 +250,13 @@ static void util_cq_read_tagged(void **dst, void *src)
...
-							list_entry);
-				if (aux_entry->cq_slot != ofi_cirque_head(cq->cirq))
-					ofi_cirque_discard(cq->cirq);
-			}
-		}
-	}
-out:
-	ofi_genlock_unlock(&cq->cq_lock);
-	return i;
+	return ofi_cq_read_cq(cq, buf, count, src_addr);
```
shefty, line 259: Why is this change being made?  What is the motivation?

shijin-aws: In our peer provider implementation, we want to make efa's `fi_cq_readfrom` exit early after we do `fi_cq_read(shm_cq, NULL, 0)` and the cqe is written, so we do not have to progress efa cq, which is more expensive than progressing shm cq. This will reduce some overhead. So after introducing this split, we can do the following in efa's cq_readfrom:

```
static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count, fi_addr_t *src_addr)
{
	struct efa_rdm_cq *cq;
	ssize_t ret;

	cq = container_of(cq_fid, struct efa_rdm_cq, util_cq.cq_fid.fid);

	if (cq->shm_cq)
		fi_cq_read(cq->shm_cq, NULL, 0);

	ret = ofi_cq_flush_cirque(&cq->util_cq, buf, count, src_addr);

	if (ret > 0)
		return ret;

	return ofi_cq_readfrom(&cq->util_cq.cq_fid, buf, count, src_addr);
}
```

include/ofi_util.h
shijin-aws: Done.

shefty: I disagree here.  In other places in the code, 'flush' means we're discarding the entries.  That's not the case here.  This is actually reading from the cq.  The behavior is closer to ofi_cq_read_entries(), if we match up with the naming for ofi_cq_write_entry.

prov/util/src/util_cq.c
shijin-aws: Thank you, added!

include/ofi_util.h
shijin-aws: Thanks, updated the name.

shefty (approved):
thanks!

wzamazon (approved):

shijin-aws:
is this PR ok for merge?

PR #8719, opened 2023-03-27
prov/psm3: update provider to sync with IEFS 11.4.1.1.2

acgoldma:
Updates:
- Fixes for sockets HAL
- Improved Debug prints

PR #8720, opened 2023-03-27
v1.18.0rc2

shefty:
My best guess is that I skipped using '--ignore-case' with sort when updating the author list for rc1.  That or the sort command behaved differently somehow on the other system I used for rc1.


shefty:
Rebased to pick up commit from #8719

shefty:
It looks like all CI passed that can run.

PR #8723, opened 2023-03-28
prov/efa: increase default shm_av_size to 256

wzamazon:
Increase default shm_av_size to 256 to match SMR_MAX_PEERS.


prov/efa/src/rdm/rxr_env.c
```diff
@@ -45,7 +45,7 @@ struct rxr_env rxr_env = {
 	.use_zcpy_rx = 1,
 	.set_cuda_sync_memops = 1,
 	.zcpy_rx_seed = 0,
-	.shm_av_size = 128,
+	.shm_av_size = 256,
```
wenduwan, line 48: Clarify - If the user set `shm_av_size` to 256 on a small instance, will that cause any issue, i.e. oversubscription? Do we need to defend against that?

shijin-aws:
what is the motivation for this PR?

prov/efa/src/rdm/rxr_env.c
wzamazon: >  Clarify - If the user set shm_av_size to 256 on a small instance, will that cause any issue, i.e. oversubscription? Do we need to defend against that?

It should out. This handle the case that larger number of processes are used. It does not cause more processes to be used.

wenduwan (approved):

shijin-aws:
> what is the motivation for this PR?

Nvm, I see this `Increase default shm_av_size to 256 to match SMR_MAX_PEERS.`

wckzhang (approved):

PR #8725, opened 2023-03-29
[v1.18.x] prov/tcp: Reduce stack size for epoll event array

shefty:
Allocate the epoll event array dynamically with the progress engine to reduce the stack size (2 x 8 bytes x 1024 entries). Further reduce the memory by setting max events down to 128 from 1024, since epoll implements fairness on event reporting. (This was verified using a modified version of sock_test.)

### DIFF
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -69,7 +69,7 @@
 
 #define XNET_RDM_VERSION	0
 #define XNET_DEF_INJECT		128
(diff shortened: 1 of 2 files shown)


shefty:
@sydidelot - Note that this change will impact iouring.  Ideally, we can move the iouring events from the stack to the progress structure.

sydidelot:
@shefty Thanks for the heads up. You are right and the epoll event array should be allocated in the progress structure. :+1:

PR #8728, opened 2023-03-29
prov/efa: adjust efa_base_ep_support_op_in_order_aligned_128_bytes()

wzamazon:
check the presence of the flag using "&", instead of comparing with the flag.


wzamazon:
@mrgolin

wenduwan (approved):

lrbison (approved):
Looks good.

in general I prefer `(cap & BITS) == BITS`, just in case BITS is more than a single bit.

PR #8730, opened 2023-03-30
prov/hook/trace: Add trace for resource creation operations.

peinanz:
Add trace log on mr reg operations and trace on cq, endpoint, scalable_ep, av,
 and cntr open operation.

### DIFF
--- a/prov/hook/trace/src/hook_trace.c
+++ b/prov/hook/trace/src/hook_trace.c
@@ -1008,7 +1008,7 @@ static int trace_pep_init(struct fid *fid)
 	return 0;
 }
 
-static ssize_t trace_cq_read_op(struct fid_cq *cq, void *buf, size_t count)
+static ssize_t trace_cq_read(struct fid_cq *cq, void *buf, size_t count)
 {
 	struct hook_cq *mycq = container_of(cq, struct hook_cq, cq);
 	ssize_t ret;
@@ -1019,7 +1019,7 @@ static ssize_t trace_cq_read_op(struct fid_cq *cq, void *buf, size_t count)
 }
 
 static ssize_t
-trace_cq_readerr_op(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t flags)
+trace_cq_readerr(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t flags)
 {
 	struct hook_cq *mycq = container_of(cq, struct hook_cq, cq);
 	ssize_t ret;
@@ -1031,7 +1031,7 @@ trace_cq_readerr_op(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t fla
 }
 
 static ssize_t
-trace_cq_readfrom_op(struct fid_cq *cq, void *buf, size_t count, fi_addr_t *src_addr)
+trace_cq_readfrom(struct fid_cq *cq, void *buf, size_t count, fi_addr_t *src_addr)
 {
 	struct hook_cq *mycq = container_of(cq, struct hook_cq, cq);
 	ssize_t ret;
@@ -1042,7 +1042,7 @@ trace_cq_readfrom_op(struct fid_cq *cq, void *buf, size_t count, fi_addr_t *src_
 }
 
 static ssize_t
-trace_cq_sread_op(struct fid_cq *cq, void *buf, size_t count,
+trace_cq_sread(struct fid_cq *cq, void *buf, size_t count,
 	      const void *cond, int timeout)
 {
 	struct hook_cq *mycq = container_of(cq, struct hook_cq, cq);
@@ -1054,7 +1054,7 @@ trace_cq_sread_op(struct fid_cq *cq, void *buf, size_t count,
 }
 
 static ssize_t
-trace_cq_sreadfrom_op(struct fid_cq *cq, void *buf, size_t count,
+trace_cq_sreadfrom(struct fid_cq *cq, void *buf, size_t count,
 		  fi_addr_t *src_addr, const void *cond, int timeout)
 {
 	struct hook_cq *mycq = container_of(cq, struct hook_cq, cq);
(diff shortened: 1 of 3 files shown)


shefty (changes requested):
We should be able to use fi_tostr() to convert all attributes to strings through here.  In some cases, we may need to extend fi_tostr() to support new structures (e.g. cntr_attr, mr_attr, etc.)

prov/hook/trace/src/hook_trace.c
```diff
@@ -80,6 +82,39 @@
...
+
+#define TRACE_RX_ATTR(hprov, rx_attr)	\
+	FI_TRACE(hprov, FI_LOG_DOMAIN,                                       \
+	        "\trx_attr [caps 0x%lx mode 0x%lx op_flags 0x%lx "           \
+	        "msg_order 0x%lx comp_order 0x%lx total_buffered_recv %zu "  \
+	        "size %zu iov_limit %zu]\n",                                 \
+	        rx_attr->caps, rx_attr->mode, rx_attr->op_flags,             \
+	        rx_attr->msg_order, rx_attr->comp_order,                     \
+	        rx_attr->total_buffered_recv, rx_attr->size,                 \
+	        rx_attr->iov_limit);
```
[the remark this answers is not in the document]
peinanz: The current ofi_tostr_xx_attr() and many other ofi_tostr()_yyy() functions are statically implemented in fi_tostr.c. 
Do we want to open up these functions?  These functions do not have any checks on the input parameters, if we want to open as non-static functions, we may need to add checks on input parameters.  These functions' return type are void, how to indicate failure in return?

prov/hook/trace/src/hook_trace.c
```diff
@@ -1064,11 +1097,339 @@ struct fi_ops_cq trace_cq_ops = {
...
+		if (attr) {
+			FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN,
+			         "buf %p, len %lu access 0x%lx "
+			         "offset 0x%lx key 0x%lx ctx %p key_size 0x%lx "
+			         "iface %d flags 0x%lx mr %p\n",
+			         IOV_BASE(attr->mr_iov, attr->iov_count),
+			         IOV_LEN(attr->mr_iov, attr->iov_count),
+			         attr->access, attr->offset, attr->requested_key,
+			         attr->context, attr->auth_key_size, attr->iface,
+			         flags, *mr);
```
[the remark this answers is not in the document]
shefty: No, you'll likely need to add a couple new types.  Just add those as separate patches.

src/fi_tostr.c
```diff
@@ -883,6 +1043,18 @@ char *DEFAULT_SYMVER_PRE(fi_tostr_r)(char *buf, size_t len,
...
+		break;
+	case FI_TYPE_CQ_ATTR:
+		ofi_tostr_cq_attr(buf, len, data, "");
+		break;
+	case FI_TYPE_MR_ATTR:
+		ofi_tostr_mr_attr(buf, len, data, "");
+		break;
+	case FI_TYPE_CNTR_ATTR:
+		ofi_tostr_cntr_attr(buf, len, data, "");
+		break;
```
shefty, line 1057: We shouldn't need a prefix parameter for the new functions.  We have a prefix for the other attributes because they can be printed separately or as part of printing fi_info.

prov/hook/trace/src/hook_trace.c
```diff
@@ -82,6 +82,25 @@
 			(uint64_t)tag, (uint64_t)ignore, context); \
 	}
 
+#define TRACE_MR_ATTR(ret, dom, len, flags, attr)    \
+	if (!(ret)) { \
+		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN, \
+		         "len %lu, flags 0x%lx\n%s",  (len), (flags), \
+		          fi_tostr((void *)attr, FI_TYPE_MR_ATTR)); \
```
shefty, line 89: fi_tostr() is not threadsafe.  It would be better to use fi_tostr_r() and pass in a buffer.

prov/hook/trace/src/hook_trace.c
```diff
@@ -82,6 +82,25 @@
...
+	}
+
+#define TRACE_ENDPOINT(dom, name, ep, context, info)  \
+	do {	 \
+		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN,        \
+	         "%s %p context %p\n", name, ep, context);     \
```
shefty, line 95: nit: 1 character indentation on lines after FI_TRACE make this hard to read.

prov/hook/trace/src/hook_trace.c
```diff
@@ -1154,10 +1173,10 @@ trace_av_open(struct fid_domain *domain, struct fi_av_attr *attr,
...
 	if (!ret) {
 		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN,
-		         "av %p, type %d rx_bits 0x%x count %zu ep_per_node %zu "
-		         "name %p flags 0x%lx ctx %p\n",
-		         &myav->av, attr->type, attr->rx_ctx_bits, attr->count,
-		         attr->ep_per_node, attr->name, attr->flags, context);
+		         "av %p, context %p\n"
+		         "%s",
+		         &myav->av, context,
+		         fi_tostr(attr, FI_TYPE_AV_ATTR));
```
shefty, line 1179: Swap the order of the patches to add the new fi_tostr() support first, then only use those calls.  This avoids adding a change in the same series that a subsequent patch updates.

shefty (commented):
getting close -- thanks!

prov/hook/trace/src/hook_trace.c
```diff
@@ -1064,11 +1085,329 @@ struct fi_ops_cq trace_cq_ops = {
...
+	myav->av.fid.fclass = FI_CLASS_AV;
+	myav->av.fid.context = context;
+	myav->av.fid.ops = &hook_fid_ops;
+	myav->av.ops = &trace_av_ops;
+
+	ret = fi_av_open(dom->hdomain, attr, &myav->hav, &myav->av.fid);
+	if (!ret) {
+		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN,
+		        "av %p, context %p\n"
+		        "%s",
```
shefty, line 1180: nit: I'd toss the %s into the previous line to avoid having a line dedicated to just that format string and tighten the code.  Same comment applies to other functions below.

shefty (approved):

PR #8732, opened 2023-03-30
prov/tcp: Move io_uring cqe from the stack to progress struct

shefty:
Reduce stack use by allocating the io_uring events with the progress structure, rather than on the stack.

note: untested (system developed on doesn't even have io_uring, so compile is questionable atm).

### DIFF
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -322,6 +322,8 @@ struct xnet_progress {
 
 	struct xnet_uring	tx_uring;
 	struct xnet_uring	rx_uring;
+	ofi_io_uring_cqe_t	**cqes;
+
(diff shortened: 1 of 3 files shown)


sydidelot (approved):
Sorry for the late review @shefty. The last week was very intense for me :)
Thanks for the patch, it looks good to me!

sydidelot:
BTW, I haven't tested the code. Feel free to merge it: I'll fix the bugs (if any) next time I'll test libfabric with io_uring.

shefty:
@sydidelot -- thanks.  It turns out that we don't have systems with io_uring enabled.  We're working on updating our OS versions to ones that have it, so we can start testing it and analyzing the performance.

PR #8733, opened 2023-03-30
prov/ucx: Publish FI_LOCAL_COMM and FI_REMOTE_COMM capabilities

j-xiong:
These bits were missing from fi_info->caps and domain_attr->caps.`


shefty:
We should expand the getinfo_test to check that at least one of these bits are set.

j-xiong:
See #8738

PR #8734, opened 2023-03-30
core/tostr: Moved ofi_strncatf() and ofi_nic_tostr() to ofi_str.c.

peinanz:
Moved functions ofi_strncatf(), ofi_nic_tostr() and other static ofi_tostr_xxx functions from common.c to ofi_str.c. The declaration of these functions are moved from ofi.h to ofi_str.h.

### DIFF
--- a/src/fi_tostr.c
+++ b/src/fi_tostr.c
@@ -44,6 +44,7 @@
 #include <arpa/inet.h>
 
 #include "ofi.h"
+#include "shared/ofi_str.h"
 #include <rdma/fabric.h>
 #include <rdma/fi_domain.h>
 #include <rdma/fi_endpoint.h>
--- a/include/ofi.h
+++ b/include/ofi.h
@@ -257,7 +257,6 @@ int ofi_apply_filter(struct ofi_filter *filter, const char *name);
 
 int ofi_nic_close(struct fid *fid);
 struct fid_nic *ofi_nic_dup(const struct fid_nic *nic);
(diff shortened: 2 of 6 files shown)


src/fi_tostr.c
```diff
@@ -44,6 +44,7 @@
 #include <arpa/inet.h>
 
 #include "ofi.h"
+#include "shared/ofi_str.h"
```
shefty, line 47: This isn't a result of your changes and a rhetorical question, but why is this header under a shared subdirectory?

shefty (approved):
Changes look reasonable.  Thanks!

src/fi_tostr.c
peinanz: You pointed out before that this shared/xxx is shared between libfabric and fi_info, fi_pingpong.  Is this still a valid case?  Otherwise, I can remove the subdirectory of shared.

shefty: I don't think we need include/shared if you want to remove it.  AFAICT, only libfabric uses that file.  It's also used by providers, but the providers also just make use of files directly under include.  So, there's no benefit there.

The src/shared file used to be valid, as that was shared between libfabric and providers (built as DL).  But that's also no longer the case, as the providers now build with files directly under src.

If you can remove the include/shared and src/shared subdirectories, I think that will simplify things for now.  Maybe we can do a more in depth source restructuring later of files under src and prov/util.

shefty:
merging - any movement of the files to new locations can come separately.  Thanks!

PR #8735, opened 2023-03-30
src/hmem_rocr fixes in the rocr ipc execution path

edgargabriel:
This is a collection of three smaller fixes:
 - remove some warnings at compilation due to const'ness of an argument, and typecasts for pointer arithmetic
 - replace the pthread_spin_lock functionality in the code by the corresponding ofi functions
 - fix in the offset calculation of the rocr ipc path

Co-authored-by: Edgar Gabriel <edgar.gabriel@amd.com>
Co-authored-by: Amir Shehata <shehataa@ornl.gov>


shefty:
Please correct signed-off-by lines.  If this is needed for v1.18.x, please cherry-pick after updating sign-off-by

PR #8738, opened 2023-03-31
fabtests: Add FI_LOCAL_COMM/FI_REMOTE_COMM presence check to fi_getin…

j-xiong:
…fo_test

Providers are expected to report either FI_LOCAL_COMM or FI_REMOTE_COMM support, or both. Not every provider is following this guide line. This patch adds a new test case to fi_getinfo_test to help discovering non-conforming provders.

### DIFF
--- a/fabtests/unit/getinfo_test.c
+++ b/fabtests/unit/getinfo_test.c
@@ -729,6 +729,31 @@ static int test_caps_regression(char *node, char *service, uint64_t flags,
 	return 0;
 }
 
+/* Check if at least one of FI_LOCAL_COMM and FI_REMOTE_COMM is set */
+static int test_comm_caps(char *node, char *service, uint64_t flags,
+		struct fi_info *hints, struct fi_info **info)
+{
+	uint64_t test_caps[2] = { FI_LOCAL_COMM, FI_REMOTE_COMM };
+	int ret, i;
+
+	for (i = 0; i < 2; i++) {
+		hints->caps &= ~(FI_LOCAL_COMM | FI_REMOTE_COMM);
+		hints->domain_attr->caps &= ~(FI_LOCAL_COMM | FI_REMOTE_COMM);
+		hints->caps |= test_caps[i];
+		hints->domain_attr->caps |= test_caps[i];
+		ret = fi_getinfo(FT_FIVERSION, node, service, flags, hints, info);
+		if (ret == 0) {
+			if (((*info)->caps & test_caps[i]) == test_caps[i] &&
+			    ((*info)->domain_attr->caps & test_caps[i]) == test_caps[i])
+				return 0;
+
+			fi_freeinfo(*info);
+		}
+	}
+
+	FT_DEBUG("Neither FI_LOCAL_COMM nor FI_REMOTE_COMM is set\n");
+	return -FI_EIO;
+}
 
 /*
  * getinfo test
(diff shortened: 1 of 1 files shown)


j-xiong:
My understanding is that domain caps should also appear in provider caps although it's unclear in the man pages.  A few providers (efa, shm) only set these in domain caps. Open MPI ofi mtl checks these in provider caps and has to add an exception for efa so that the provider can be used (https://github.com/open-mpi/ompi/blob/main/ompi/mca/mtl/ofi/mtl_ofi_component.c#L701-L707).

shefty:
FI_REMOTE_COMM and FI_LOCAL_COMM are both defined as fi_info::caps and domain_attr::caps.  At least one should be set for both, and the two fields should agree.

fabtests/unit/getinfo_test.c
```diff
@@ -729,6 +729,37 @@ static int test_caps_regression(char *node, char *service, uint64_t flags,
...
+	}
+
+	for (i = 0; i < 2; i++) {
+		test_hints->caps &= ~(FI_LOCAL_COMM | FI_REMOTE_COMM);
+		test_hints->domain_attr->caps &= ~(FI_LOCAL_COMM | FI_REMOTE_COMM);
+		test_hints->caps |= test_caps[i];
+		test_hints->domain_attr->caps |= test_caps[i];
+		ret = fi_getinfo(FT_FIVERSION, node, service, flags, test_hints, info);
+		if (ret == 0)
+			goto out;
```
shefty, line 753: We can strengthen this check a little with something like this:
```
((info->caps & test_caps[i]) == test_caps[i]) &&
(((info->caps & test_caps[i]) == ((info->domain_attr->caps & test_caps[i])))
```

j-xiong: Done.

shefty:
Will rebasing this PR allow CI to pass?

j-xiong:
Rebased.

fabtests/unit/getinfo_test.c
```diff
@@ -729,6 +729,43 @@ static int test_caps_regression(char *node, char *service, uint64_t flags,
...
+{
+	struct fi_info *test_hints = NULL;
+	uint64_t test_caps[2] = { FI_LOCAL_COMM, FI_REMOTE_COMM };
+	int ret, i;
+
+	if (base_hints) {
+		test_hints = fi_dupinfo(base_hints);
+		if (!test_hints)
+			return -FI_ENOMEM;
+	}
```
aingerson, line 744: I think every iteration of the tests does this already and passes in the duplicated hints. Since the hints are a const going into fi_getinfo you should just be able to modify the hints for both cases.

fabtests/unit/getinfo_test.c
```diff
@@ -729,6 +729,43 @@ static int test_caps_regression(char *node, char *service, uint64_t flags,
...
+		test_hints->caps |= test_caps[i];
+		test_hints->domain_attr->caps |= test_caps[i];
+		ret = fi_getinfo(FT_FIVERSION, node, service, flags, test_hints, info);
+		if (ret == 0) {
+			if (((*info)->caps & test_caps[i]) == test_caps[i] &&
+			    ((*info)->domain_attr->caps & test_caps[i]) == test_caps[i]) {
+				goto out;
+			} else {
+				fi_freeinfo(*info);
+			}
```
aingerson, line 758: Drop else case and outdent. Optional: drop backets on if

fabtests/unit/getinfo_test.c
```diff
@@ -729,6 +729,43 @@ static int test_caps_regression(char *node, char *service, uint64_t flags,
...
+		}
+	}
+
+	FT_DEBUG("Neither FI_LOCAL_COMM nor FI_REMOTE_COMM is set\n");
+	ret = -FI_EIO;
```
aingerson, line 763: The ENODATA case will also go here. I think that error needs to be handled differently.

fabtests/unit/getinfo_test.c
j-xiong: you are right. Thanks.

fabtests/unit/getinfo_test.c
j-xiong: That's by design. ENODATA here means none of the flags is set and that is the error we want to report.

aingerson: Oh I must have misread it. I thought you were continuing to the second test if the first one passed. Looks good. Thanks!

Issue #8745, opened 2023-03-31
Proposal: SHM cuda-to-cuda performance improvement


wenduwan:
**Is your feature request related to a problem? Please describe.**
This is a feature request from AWS EFA team to improve the cuda-to-cuda performance in SHM.

**Describe the solution you'd like**
Currently SHM implements cuda-to-cuda copies with `cudaMemcpy`whenever it's available, i.e. ipc -
```c
static struct smr_pend_entry *smr_progress_ipc(...)
{
        ...
	if (cmd->msg.hdr.op == ofi_op_read_req) {
		hmem_copy_ret = ofi_copy_from_hmem_iov(ptr, cmd->msg.hdr.size,
					cmd->msg.data.ipc_info.iface,
					cmd->msg.data.ipc_info.device, iov,
					iov_count, 0);
	} else {
		hmem_copy_ret = ofi_copy_to_hmem_iov(cmd->msg.data.ipc_info.iface,
					cmd->msg.data.ipc_info.device, iov,
					iov_count, 0, ptr, cmd->msg.hdr.size);
	}
        ...
}
```
Based on EFA team's experiment, `cudaMemcpy` is slower than gdrcopy(if available) for message sizes up to 3kb(based on p4d.24xlarge platform). But due to the current implementation in SHM, we cannot register both the tx and rx MR using gdrcopy memory handle, because it does not support device-to-device copies.

Currently in EFA we work around this issue by gdrcopying application memory from cuda to host before dispatching it to SHM. But this has become problematic since we are migrating to the Peer API - it is designed to allow EFA to call SHM `fi_send*` using the application MR directly.

Therefore we are planning to enhance SHM(and later SM2), around the handling of `FI_HMEM_CUDA` - the idea is to take advantage of gdrcopy over `cudaMemcpy` for smaller messages. This will roughly incur changes in:
- Additional logic to determine if `struct ofi_mr->device` is a gdr memory handle
- De-select ipc when gdr memory handle is present and message is smaller than X bytes(going to start with 3kb, see table below), e.g.
```c
static ssize_t smr_generic_sendmsg(...)
{       ...
	/* Do not inline/inject if IPC is available so device to device
	 * transfer may occur if possible. - unless gdrcopy should be used */
	if (iov_count == 1 && desc && desc[0]) {
                if (iface == FI_HMEM_CUDA && is_gdr_mem_handle(device) && total_len < some_size) {
                       use_ipc = false;
                } else {
[...]

PR #8746, opened 2023-03-31
core/hmem_cuda: expose cuda_is_device_id API

wenduwan:
cuda currently supports 2 memcpy ops, i.e. cudaMemcpy(always available) and gdrcopy. struct ofi_mr->device is the only attribute to determine if gdrcopy can be used.

This patch renames ofi_cudaIsDeviceId and exposes it to provider code, thus enables provider to switch between the memcpy ops and achieve performance gains.

Related: https://github.com/ofiwg/libfabric/issues/8745

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -221,7 +221,7 @@ static cudaError_t ofi_cudaGetDeviceCount(int *count)
 	return cuda_ops.cudaGetDeviceCount(count);
 }
 
-static bool ofi_cudaIsDeviceId(uint64_t device) {
+bool cuda_is_device_id(uint64_t device) {
 	return device < cuda_device_count;
 }
 
@@ -237,7 +237,7 @@ cudaError_t ofi_cudaFree(void *ptr)
 
 int cuda_copy_to_dev(uint64_t device, void *dst, const void *src, size_t size)
 {
-	if (hmem_cuda_use_gdrcopy && !ofi_cudaIsDeviceId(device)) {
+	if (hmem_cuda_use_gdrcopy && !cuda_is_device_id(device)) {
 		cuda_gdrcopy_to_dev(device, dst, src, size);
 		return FI_SUCCESS;
 	}
@@ -258,7 +258,7 @@ int cuda_copy_to_dev(uint64_t device, void *dst, const void *src, size_t size)
 
 int cuda_copy_from_dev(uint64_t device, void *dst, const void *src, size_t size)
 {
-	if (hmem_cuda_use_gdrcopy && !ofi_cudaIsDeviceId(device)) {
+	if (hmem_cuda_use_gdrcopy && !cuda_is_device_id(device)) {
 		cuda_gdrcopy_from_dev(device, dst, src, size);
 		return FI_SUCCESS;
 	}
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -176,6 +176,7 @@ enum cuda_xfer_setting cuda_get_xfer_setting(void);
 bool cuda_is_ipc_enabled(void);
 int cuda_get_ipc_handle_size(size_t *size);
 bool cuda_is_gdrcopy_enabled(void);
+bool cuda_is_device_id(uint64_t device);
 
 void cuda_gdrcopy_to_dev(uint64_t handle, void *dev,
 			 const void *host, size_t size);


src/hmem_cuda.c
```diff
@@ -221,7 +221,7 @@ static cudaError_t ofi_cudaGetDeviceCount(int *count)
 	return cuda_ops.cudaGetDeviceCount(count);
 }
 
-static bool ofi_cudaIsDeviceId(uint64_t device) {
+bool cuda_is_device_id(uint64_t device) {
```
shefty, line 224: The change is fine, but I don't understand the logic behind having this check.  What is 'device' if it's not actually a device?  At best, this looks like a sanity check, not something to base the software flow on.

wenduwan: I agree with the last sentence - this function exists because we don't have a fast cuda api to validate if the given value is a gdr memory handle. Basically what I really wanted is `cuda_is_gdr_handle_valid(...)`.

That said I don't see a better place to store this value - it only makes sense for cuda. In an early thread I also wondered if it is better to make gdr an iface by itself, i.e. `FI_HMEM_GDR`, in addition to cuda. That's a bigger change though.

shefty: Let's try to fix this, rather than abusing a variable to carry something value that's not its intended use.  What is 'device' if not an index to a device?  Where did this value come from?

gdrcopy isn't an interface like CUDA or RoCR; just a copy library.  FI_HMEM_GDR doesn't make sense.  gdrcopy aligns better with other copy routines, like using IPC, DSA, CMA, and XPMEM.  It's worth looking at the comments in PR #8200 for ways to handle this.  @amirshehataornl was working on integrating xpmem support, and gdrcopy may fit into that same model.

wzamazon: > I agree with the last sentence - this function exists because we don't have a fast cuda api to validate if the given value is a gdr memory handle. Basically what I really wanted is cuda_is_gdr_handle_valid(...).

Can't we store that information (whether gdrcopy is used) inside `ofi_mr`?

tschuett: The namespace FI_HMEM_ is reserved for offload devices. You could introduce a new namespace for copy-like tools, e.g, FI_COPY_

wenduwan: > Can't we store that information (whether gdrcopy is used) inside ofi_mr?

@wzamazon This is a good question. Yes we can set a flag to indicate the presence of gdr memory handle. But I'm not sure if this alone is the desired solution.

I agree with @shefty that using `device` for a non-device id is not ideal. In reality, for cuda hmem we probably need to store the gdr handle in an additional field in `ofi_mr` along side `device`, e.g. `void* context`. It is up to each hmem iface to decide how to interpret this context. Concretely:
```c
struct ofi_mr {
	struct fid_mr mr_fid;
	struct util_domain *domain;
	uint64_t key;
	uint64_t flags;
	enum fi_hmem_iface iface;
	uint64_t device;
        void* context; // <- new field
};
```

In the context of peer provider workflow, where the application memory is registered with the owner and passed to the peer, this allows the owner provider(e.g. EFA) to convey the presence of e.g. gdr handle, to the peer(e.g. SHM). 

To use the handle we also need to modify the hmem API, e.g.
```c
ssize_t ofi_copy_from_hmem_iov(void *dest, size_t size,
			       enum fi_hmem_iface hmem_iface,
                               uint64_t device, void* context,
			       const struct iovec *hmem_iov,
			       size_t hmem_iov_count,
			       uint64_t hmem_iov_offset)
```

@shefty Thanks for the pointer. After reading Amir's work I'm not convinced that it is applicable to gdrcopy. IIUC XPMEM provides universal ipc capabilities, but gdrcopy is rather a situational optimization - and more importantly it does not support ipc(no cuda -> cuda). To use gdrcopy the provider needs to provision&manage host memory buffers.

Another observation is that, `device` is not equally used amongst hmem ifaces. ze appears to be the most reliance iface on device id, probably due to its API design. In comparison the other ifaces have limited use of the device id. But I do see value of storing the device id value for future proof.

@tschuett Could you elaborate on the suggestion? It sounds interesting but my understanding is for any offload device, it should also provide eligible memcpy methods, so FI_HMEM and FI_COPY are essentially coupled. CMIIW.

tschuett: FI_HMEM_ROCR will work fine with FI_COPY_XPMEM. But FI_HMEM_XPMEM and FI_HMEM_GDRCOPY sound strange.

shefty: FI_HMEM is used to indicate an external set of APIs that an application is using: CUDA, RoCR, Level-0.  There is no xpmem API to allocate buffers.  gdrcopy is part of CUDA.

AFAICT, CUDA was designed around a single GPU per system.  It's had to hack in support for multiple GPUs in a backward compatible way, such as relying on setting some global device value before each call.

Why can't ofi_copy_from_hmem_iov() use the fastest copy routine?  If we're having to pass in a variable to that call that is directly telling it that: a) we have CUDA buffers and b) to use gdrcopy(), then the call isn't a useful an abstraction.  The caller should just call gdrcopy().

What do we need to create a call like cuda_is_gdr_handle_valid()?  What parameters are needed for that call?  And where should it be called?

@wenduwan - xpmem supports IPC copies (and will never be an upstream solution), but DSA does not.  DSA is currently an async copy engine.  In that regard, gdrcopy can be viewed similarly (whether copies acts sync or async).  CMA is also used as a simple copy routine.

wenduwan: > but DSA does not. DSA is currently an async copy engine. In that regard, gdrcopy can be viewed similarly (whether copies acts sync or async).

This makes sense. I didn't know dsa but looking at its interface, it does resemble gdrcopy. We could implement similar routines (essentially exporting some functions in `src/hmem_cuda_gdrcopy.c`) and let the provider call them rather than e.g. `ofi_copy_from_hmem_iov`. But still, we need to solve the problem of passing around the gdr handle, with `ofi_mr` being the obvious candidate vehicle for that.

> Why can't ofi_copy_from_hmem_iov() use the fastest copy routine? 
What do we need to create a call like cuda_is_gdr_handle_valid()? What parameters are needed for that call? And where should it be called?

In the case of gdrcopy , the question is rather how does it know whether the copy is legal. You correctly pointed out that we are missing the `cuda_is_gdr_handle_valid` piece, but it is not necessary if we simply use a flag(which replaces the device id check in this PR), i.e.
```
1. Owner provider pins the buffer, receives the handle
2. Owner passes the handle and a flag to the peer, indicating this is a gdr handle
3. Peer sees the flag and chooses to use gdrcopy during hmem memcpy(when gdrcopy is faster)
```

On the other hand, to implement `cuda_is_gdr_handle_valid` we probably need something akin to the ipc cache. The [official api ](https://github.com/NVIDIA/gdrcopy/blob/master/include/gdrapi.h)does not appear to provide this cap - I need to quantify the time cost of using the cache though.

So, I'm seeing a rough shape, i.e.

We still need to extend `ofi_mr` for the (owner) provider to register MR with gdr handle
```c
struct ofi_mr {
	struct fid_mr mr_fid;
	struct util_domain *domain;
	uint64_t key;
	uint64_t flags; // Set gdr flag
	enum fi_hmem_iface iface;
	uint64_t device;
        void* context; // <- store gdr handle
};
```

Then later the (peer) provider prioritizes gdrcopy over the default hmem copy. Basically we export the gdrcopy ops similar to DSA, e.g.
```
gdr_copy_from_dev
gdr_copy_to_dev
```
so that we can do something similar to
```c
if (ofi_mr->flags & gdr_flag && other_conditions) {
        ret = gdr_copy_from_dev(...)
} else {
        ofi_copy_from_hmem_iov(...)
}
```

This is manageable since only the EFA provider uses gdrcopy right now.

shefty: ofi_mr_info contains an ipc_handle[] field.  Can a new hmem_handle be added there (possibly as a union)?

shijin-aws: True, but it seems this object is currently only used for searching entries in mr cache, how can we get it from a general mr desc?

shefty: I believe xpmem has a similar requirement for mapping an address to a handle and storing it in a cache.  This still seems similar to the discussion in #8200.

shijin-aws: I read that PR, it is currently making xpmem one of the internal hmem iface, are we converging to do it for gdrcopy similarly ?

shefty: Read my comment from Dec. 7th, not the PR changes itself.

shijin-aws: I see, post the link of this comment https://github.com/ofiwg/libfabric/pull/8200#issuecomment-1341810471 here for convenience.

shefty: To be clear, I'm okay accepting these changes.  They at least don't make things worse than they already are.  But I would like to work toward a more reasonable solution.

shijin-aws: I prefer we go with the direction of https://github.com/ofiwg/libfabric/pull/8200#issuecomment-1341810471 for long term benefits.

wzamazon: @shefty  @shijin-aws 

From what I can tell, the issue is:

To use gdrcopy, the memory must be registered with gdrcopy by calling `gdrcopy_pin_buffer`, for which gdrcopy return a handle. The process is expensive.

In the context of using peer API, if owner has already registered the memory with gdrcopy, the owner want to pass the handle to peer, so peer does not need to pin the memory again.

However, current `fi_mr_regattr` API does not have a good way for owner to pass the handle to peer. 

I think two possible options are:

1. extend the `fi_mr_regattr` API to allow caller to pass gdrcopy handle when register CUDA memory (this potentially allow app that using gdrcopy pass existing handle).
2. implement a gdrcopy handle cache.

shefty: ofi_mr isn't shared between providers or exposed above the API.  I'm also unsure of the software flow you're envisioning.  A MR cache should only be needed at the 'topmost' provider.  The cached entry will control the lifetime of registrations done by other providers.

Currently, memory registration does not have a peer-model concept.  We have peer domains, which allows registering memory with a peer, but that targets FI_PEER_TRANSFER.  It is not sharing memory registration.  Maybe peer MRs are needed, but that's unclear to me.

I'm wondering if struct fi_mr_attr needs to convey this information.

wzamazon:
> I'm wondering if struct fi_mr_attr needs to convey this information.

I think it would be ideal that we can extend `fi_mr_attr` so it can be used to pass a gdrcopy registration handle.

This would allow two scenaro:

an owner provider that has register a buffer with gdrcopy to pass the handle to its peer provider.

An application that has registered a buffer with gdrcopy to pass the handle to libfabric

shefty:
Option 1 is to use the MR flags and pair it with a new field at the end of the struct (e.g. ```void *hmem_data``` or ```void *hmem_handle```).

Option 2 is to add another ```enum fi_hmem_something``` + ```union { } something``` pair.

With the potential to expose this directly to end-users, we should pick easy to understand without constraining ourselves too much.  It would help to identify other possible uses of any proposed extension.

src/hmem_cuda.c
shefty: The app calls register, does some transfers, then close on the MR.  The purpose of the cache is to keep the MR data around after it's been closed, in case it's accessed again.  In the case of EFA + SHM, EFA would maintain the cache.  As a result, SHM would not see that its (the shm) MR is closed until EFA flushes its (the efa) MR from the cache.  This concept would extend to multiple coordinating peers, or, for example, a link provider.

wenduwan: I see. In the context of Owner + Peer, the owner owns the cache, and decides when to destroy it.

In this case, is there a convenient mechanism for the Owner to shared the cache with the Peer? That is the missing piece in my thought experiment. Maybe `fi_import/export`?

shefty: There shouldn't be a need for the owner to share the cache.  The peer will get a call, fi_mr_regattr(), create some structure to track the mr, and keep that around until owner decides to call fi_close on it.  Everything the peer needs should be provided through the fi_mr_regattr() call.

wenduwan:
> With the potential to expose this directly to end-users, we should pick easy to understand without constraining ourselves too much. It would help to identify other possible uses of any proposed extension.

Understood. Another reason I looked into this option is the choice of `device` id in the current API. As far as CUDA is concerned this field is not useful in memcpy - which is basically why I wanted to reuse it for gdr handle.

I suspect, generally speaking, the name `hmem_handle` is more useful/interesting for memcpies. Anecdote only though.

src/hmem_cuda.c
wenduwan: hmmmmm. If the MR info can be passed to fi_mr_regattr and retrieved on the Peer side, I guess we wouldn't need the cache - instead we directly pass the gdr handle.

Currently gdr handle lifetime is also simple - there isn't mid-air invalidation, so we should be covered there.

wenduwan:
@shefty I thought more about the options. Extending `struct fi_mr_attr` sounds most reasonable to me, i.e.
```
struct fi_mr_attr {
	const struct iovec	*mr_iov;
	size_t			iov_count;
	uint64_t		access;
	uint64_t		offset;
	uint64_t		requested_key;
	void			*context;
	size_t			auth_key_size;
	uint8_t			*auth_key;
	enum fi_hmem_iface	iface;
	union {
		uint64_t	reserved;
		int		cuda;
		int		ze;
		int		neuron;
		int		synapseai;
	} device;
        void                    *hmem_data;
};
```

> It would help to identify other possible uses of any proposed extension.

Regarding this I agree with @wzamazon, i.e.
```
This would allow two scenaro:

an owner provider that has register a buffer with gdrcopy to pass the handle to its peer provider.

An application that has registered a buffer with gdrcopy to pass the handle to libfabric
```

Do you have any reservation or need more time to think about it? If we are aligned on this approach I can start the work. Let me know.

wenduwan:
Opened https://github.com/ofiwg/libfabric/pull/8787 to materialize the proposal.

wenduwan:
Closing this PR in favor of https://github.com/ofiwg/libfabric/pull/8787

PR #8748, opened 2023-04-03
prov/shm: init map fi addrs to FI_ADDR_NOTAVAIL

aingerson:
The CQ should set FI_ADDR_NOTAVAIL, not FI_ADDR_UNSPEC if the address has not been inserted into the AV so the default value should be ADDR_NOTAVAIL.


aingerson:
@shefty Did not realize they were the same value! Updated the commit message.

PR #8749, opened 2023-04-03
src/hmem_ze: fix ZE is valid check

aingerson:
When seeing if an address is valid within ZE, non ZE allocated memory will still return success but will set the memory type as unknown. Check this and return false if unknown

### DIFF
--- a/src/hmem_ze.c
+++ b/src/hmem_ze.c
@@ -970,7 +970,7 @@ bool ze_hmem_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags)
 
 	ze_ret = ofi_zeMemGetAllocProperties(context, addr, &mem_props,
 					     &device_ptr);
-	if (ze_ret)
+	if (ze_ret || mem_props.type == ZE_MEMORY_TYPE_UNKNOWN)
 		return false;
 
 	if (flags)


aingerson:
@amirshehataornl This is the error we saw - it is ZE-specific so it shouldn't affect you

shefty:
That's non-intuitive behavior.  Maybe you can submit a patch to improve the Level-0 API?

j-xiong:
The behavior is not totally without merit :) The function successfully identified that the memory is not ZE-allocated and set the memory type accordingly. To this extent, the call can be considered successful.

shefty:
And did this call to zeMemGetAllocProperties() actually return the allocation properties?  I'm going to say no, it "successfully" did not.  :P  #whathappenswhenhardwareengineerswritesoftware

j-xiong:
They do document this as a successful scenario (https://spec.oneapi.io/level-zero/latest/core/api.html#zememgetallocproperties):
```
The application may query attributes of a memory allocation unrelated to the context. 
When this occurs, the returned allocation type will be ZE_MEMORY_TYPE_UNKNOWN, 
and the returned identifier and associated device is unspecified.
```

shefty:
They can document that, still doesn't make it intuitive behavior.  return "We successfully don't know anything";

shefty:
Do we need this in v1.18.x?

aingerson:
@shefty This is only visible on the hmem hook path which isn't tested or used anywhere so not really but I'll cherry pick it anyway.

PR #8751, opened 2023-04-04
prov/shm: store incoming message size in tagged rx_entry.

shijin-aws:
Currently, when getting an unexpected tagged msg, shm does not fill the size of incoming message to the rx_entry before queueing it. However, middlewares like MPI will call fi_trecvmsg with FI_PEEK flag in MPI_Probe() to peek if there is an incoming message, and application will call MPI_Get_count to derive the size of the incoming message from the output of MPI_Probe(). In this case, shm needs to store the incoming message size in tagged rx_entry before queueing it, so owner provider can get the size and write it in the CQE for FI_PEEK.

### DIFF
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -878,6 +878,7 @@ static int smr_progress_cmd_msg(struct smr_ep *ep, struct smr_cmd *cmd)
 			if (ret)
 				return ret;
 
+			rx_entry->size = cmd->msg.hdr.size;
 			ret = peer_srx->owner_ops->queue_tag(rx_entry);
 			goto out;
 		}


wzamazon (approved):

aingerson:
The definition in the man page is that the rx_entry fields are only editable by the owner, not the peer:
"All fields are only modifiable by the owner, except for the peer_context which is provided for the peer to use to save peer-specific information for unexpected message processing."
So I think I more appropriate fix would be to add a size parameter in the get_tag path so the owner can initialize the field on the unexpected path. That would align with what the get_msg path does (at least in shm) where it initializes the size field when allocating the unexpected rx_entry.

shijin-aws:
>  So I think I more appropriate fix would be to add a size parameter in the get_tag path so the owner can initialize the field on the unexpected path. That would align with what the get_msg path does (at least in shm) where it initializes the size field when allocating the unexpected rx_entry.

@aingerson I thought about that, adding `size` to `get_tag` is doable, but that requires the API change so I did not go that direction firstly.

@shefty  what do you think?

shijin-aws:
@aingerson A side node: I did not see how shm handles FI_PEEK in trecvmsg, am I missing something or shm just does not support it yet?

shijin-aws:
@shefty Thanks, I opened a new PR https://github.com/ofiwg/libfabric/pull/8753 to change this API.

shijin-aws:
Close this PR.

PR #8752, opened 2023-04-04
prov/efa: update rx_entry.total_len in srx queue_tag

shijin-aws:
Currently, when peer provider getting an unexpected tagged msg, efa incorrectly set the total_len in the unexp rx entry as 0 because get_tag API does not have size as the input. However, middlewares like MPI will call fi_trecvmsg with FI_PEEK flag in MPI_Probe() to peek if there is an incoming message, and application will call MPI_Get_count to derive the size of the incoming message from the output of MPI_Probe(). In this case, efa needs to store the incoming message size in tagged rx_entry before queueing it, so efa can get the size and write it in the CQE after finding the rx_entry in the unexp queue.

The value of peer_rx_entry.size will be updated by peer provider before calling queue_tag.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_srx.c
+++ b/prov/efa/src/rdm/efa_rdm_srx.c
@@ -228,6 +228,12 @@ static int efa_rdm_srx_queue_tag(struct fi_peer_rx_entry *peer_rx_entry)
 	ep = rx_entry->ep;
 
 	ofi_mutex_lock(&ep->base_ep.util_ep.lock);
+	/*
+	 * rx_entry->peer_rx_entry.size is updated by peer provider before calling queue_tag.
+	 * we can remove this line after replacing total_len by peer_rx_entry.size in all
+	 * the efa provider code.
+	 */
+	rx_entry->total_len = rx_entry->peer_rx_entry.size;
 	rxr_msg_queue_unexp_rx_entry_for_tagrtm(ep, rx_entry);
 	ofi_mutex_unlock(&ep->base_ep.util_ep.lock);
 	return FI_SUCCESS;


prov/efa/src/rdm/efa_rdm_srx.c
```diff
@@ -228,6 +228,7 @@ static int efa_rdm_srx_queue_tag(struct fi_peer_rx_entry *peer_rx_entry)
 	ep = rx_entry->ep;
 
 	ofi_mutex_lock(&ep->base_ep.util_ep.lock);
+	rx_entry->total_len = rx_entry->peer_rx_entry.size;
```
wenduwan, line 231: Is this unique to tagged message? Do we need to, or does it hurt to do the same on L208 for untagged message?

wzamazon:
commit title is too long

prov/efa/src/rdm/efa_rdm_srx.c
shijin-aws: >Is this unique to tagged message?

Yes, this is unique to tagged message because of 2 reasons.

1. FI_PEEK only applies to tagged message
2. For the non-tagged case, the size of rx_entry is already updated when peer provider calls `get_msg`, as the msg size is the input of this API call.

shijin-aws: Actually, this change can be avoided if we can combine the duplicate fields in rx_entry and rx_entry.peer_rx_entry. I will add a TODO here.

wzamazon (approved):

shijin-aws:
Please do not merge this PR before we addressing the discussion in https://github.com/ofiwg/libfabric/pull/8751

shijin-aws:
The community suggests a different solution that adding `size` into the `get_tag` API. I opened a new PR for that: https://github.com/ofiwg/libfabric/pull/8753

PR #8759, opened 2023-04-05
fabtests/functional: av_xfer test remove av before sync

wenduwan:
This test occasionally fails under load for EFA provider due to its
    handshake protocol. The recv completion triggers a handshake exchange
    that could outlast ft_sync. This causes the sender(client) to receive 2
    handshake acknowledgements from the same receiver(server). This is
    an unexpected behavior.
    
    To prevent such issues, the test should remove the peer address from the
    av before ft_sync. For EFA provider, this allows the endpoint to ignore
    the completions from its peer, thus avoiding the unexpected behavior.

### DIFF
--- a/fabtests/functional/av_xfer.c
+++ b/fabtests/functional/av_xfer.c
@@ -109,6 +109,13 @@ static int av_removal_test(void)
 		}
 	}
 
+	/* Purge the av before sync to avoid unexpected completions */
+	ret = fi_av_remove(av, &remote_fi_addr, 1, 0);
+	if (ret) {
+		FT_PRINTERR("fi_av_remove", ret);
+		goto out;
+	}
+
 	(void) ft_sync();
 out:
 	fprintf(stdout, "%s\n", ret ? "FAIL" : "PASS");
@@ -175,6 +182,13 @@ static int av_reinsert_test(void)
 			FT_PRINTERR("ft_rx", -ret);
 			goto out;
 		}
+	}
+
+	/* Purge the av before sync to avoid unexpected completions */
+	ret = fi_av_remove(av, &remote_fi_addr, 1, 0);
+	if (ret) {
+		FT_PRINTERR("fi_av_remove", ret);
+		goto out;
 	}
 
 	(void) ft_sync();


shefty:
This test:

1. Inserts an address into the AV
2. Uses the address to send/recv data, verifying that it works.
3. Removes the address from the AV
4. Syncs that this worked on both sides (line 34) to avoid possible races
5. Repeats steps 1-3

The purpose is checking that an address can be inserted then removed then inserted again.  It helps test providers that implement connections underneath the AV that they can handle this flow.  I do expect the test to work with all providers, including EFA.

wenduwan:
Thanks @shefty  for clarifying - I updated the patch to remove the peer address before sync. Based on my understanding this should not change the test invariant.

Please let me know otherwise.

fabtests/functional/av_xfer.c
```diff
@@ -109,6 +109,13 @@ static int av_removal_test(void)
 		}
 	}
 
+	/* Purge the av before sync to avoid unexpected completions */
+	ret = fi_av_remove(av, &remote_fi_addr, 1, 0);
+	if (ret) {
+		FT_PRINTERR("fi_av_remove", ret);
+		goto out;
+	}
```
shefty, line 117: I think the change to the test here is to work around a bug in the efa provider, and not needed for correctness.  Immediately after calling ft_sync(), we destroy the AV.  Other tests do not remove addresses from the AV before destroying it.

It sounds like the test is exposing the problem.  Changing the test to no longer expose the problem somewhat defeats the purpose of the test.

fabtests/functional/av_xfer.c
```diff
@@ -177,6 +184,13 @@ static int av_reinsert_test(void)
...
```
shefty, line 192: Same situation as above, we destroy the AV after the sync.

fabtests/functional/av_xfer.c
wenduwan: > Immediately after calling ft_sync(), we destroy the AV.

I'm a little confused about this. Are you referring to somewhere outside of this test?

> It sounds like the test is exposing the problem. Changing the test to no longer expose the problem somewhat defeats the purpose of the test.

Thinking about the issue, I have a hunch that it is not specific to EFA under heavy load.

Re-inserting the same `remote_fi_addr` **shortly** after removing it from av is a somewhat undefined behavior. Just like in our case, for a given provider, it is possible that the endpoint receives a stale completion from the same address, and it is not clear to me whether this completion should be matched with a new peer with the same address. For EFA this is currently a forbidden/unexpected behavior, but other providers might have different guarantees. But do you think such behavior should be asserted in this test?

shefty: I don't see undefined behavior here.  The address was removed, then re-inserted.  Completions match with posted operations, not AV entries.

Imagine this app behavior.  Clients insert a server's address into an AV to send their own address to the server.  The server collects all client addresses, adds its own to the list, and sends that data back to the clients.  Now everyone has each others' addresses.  To align with each other, existing entries in the AV are removed, and the new list of addresses are inserted.  The removal is needed to ensure that the same address isn't inserted twice.

libfabric allows *receiving* messages from peers that are not inserted into the local AV.  This is needed to avoid requiring a second communication API to exchange addresses out of band.  If a provider can't support that feature, that's fine and why fabtests also supports OOB address exchange.

I know with tcp and rxm, AV removal doesn't impact lower-level connections by default.  Those continue to exist.

wenduwan: FYI. EFA provider does extra work in this process. When Sender and Receiver communicate for the first time, the Receiver will process the rx completion and send back a handshake request to the Sender, to exchange information such as Sender's supported capabilities, etc. This handshake is handled by the Sender in the next progress. In this test this process happens twice. Normally, the Sender receives the handshake in each ft_sync call, but in rare occasions, it does not receive the handshake in the 1st ft_sync and 2 handshakes in the 2nd ft_sync. EFA does not expect 2 handshakes from the same peer, thus the issue.


> The address was removed, then re-inserted.

Now I see your point, and it comes down to the assumption that `ft_rx` does nothing more than receive - which EFA does not satisfy. There is more going on between `removed` and `re-inserted`.

Let me think more about the solution.

shefty: If it helps, tcp and rxm also do more work the first time 2 ep's communicate.  However, that information is maintained separately from the addresses in the AV table.  In our case, the AV table provides an index into a separate table managing connection state.  If no connection exists (which is the common case), that index will be invalid.  However, if there's an existing connection to the peer when its address has been inserted, we'll find it through a lookup and set the index accordingly.

Removing an address doesn't impact the existing connection.  Though we have an option that can be set to force cleanup of any connection when an address is removed.  This is useful for client-server apps that use rdm ep's, where the clients can go away and never return.

wenduwan: @shefty thanks for the tips! After discussion with my team I will close this PR and leave the test unchanged.

We will relax the check in EFA provider instead.

Appreciate the discourse!

PR #8760, opened 2023-04-05
prov/shm: Fix ROCR data coherency

amirshehataornl:
Set the response status to SMR_STATUS_BUSY after launching an async copy operation. Set the response status to SMR_STATUS_SUCCESS after the async operation completes.

This will prevent the tx complete on the transmit side until the data from the sending buffer is no longer in use.

This avoids a scenario where the application can overwrite the buffer before it has been copied over, causing the receiving side to get unexpected data.

### DIFF
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -56,6 +56,8 @@ struct ofi_mr_info {
 	struct iovec iov;
 	enum fi_hmem_iface iface;
 	uint64_t device;
+
+	uint64_t peer_id;
 	void     *ipc_mapped_addr;
 	uint8_t  ipc_handle[MAX_IPC_HANDLE_SIZE];
 };
@@ -351,8 +353,9 @@ void ofi_mr_cache_notify(struct ofi_mr_cache *cache, const void *addr, size_t le
 int ofi_ipc_cache_open(struct ofi_mr_cache **cache,
 			struct util_domain *domain);
 void ofi_ipc_cache_destroy(struct ofi_mr_cache *cache);
-int  ofi_ipc_cache_search(struct ofi_mr_cache *cache, struct ipc_info *ipc_info,
-			   struct ofi_mr_entry **mr_entry);
+int  ofi_ipc_cache_search(struct ofi_mr_cache *cache, uint64_t peer_id,
+			  struct ipc_info *ipc_info,
+			  struct ofi_mr_entry **mr_entry);
 
 static inline bool ofi_mr_cache_full(struct ofi_mr_cache *cache)
 {
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -572,8 +572,9 @@ static struct smr_pend_entry *smr_progress_ipc(struct smr_cmd *cmd,
 				&ipc_fd, ipc_device, &base);
 	} else {
 		ret = ofi_ipc_cache_search(domain->ipc_cache,
-				           &cmd->msg.data.ipc_info,
-				           &mr_entry);
+					   cmd->msg.hdr.id,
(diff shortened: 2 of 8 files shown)


prov/shm/src/smr_progress.c
```diff
@@ -587,10 +587,12 @@ static struct smr_pend_entry *smr_progress_ipc(struct smr_cmd *cmd,
 	if (cmd->msg.data.ipc_info.iface == FI_HMEM_ROCR) {
 		*total_len = 0;
 		ret = 0;
+		resp->status = SMR_STATUS_BUSY;
 		ipc_entry = smr_ipc_async_copy(ep,
 				(char*)ptr,
 				rx_entry, iov, iov_count, mr_entry, cmd, &ret);
-		resp->status = ret;
+		if (!ipc_entry)
+			resp->status = ret;
```
shefty, line 595: The changes are pointing out that smr_ipc_async_copy() is odd, in that the return code is a parameter, rather than, well, a return code.  Can we change that function to return the return code and move ipc_entry as an output parameter?  That's more intuitive and makes the assignment of resp->status = ret more obvious.  Because on first reading, all that looked like it did was set resp->status = 0, since ret = 0 immediately above the call.

aingerson (approved):
Shm change looks good to me. Glad you were able to find it!

src/hmem_rocr.c
```diff
@@ -856,6 +889,13 @@ int rocr_hmem_init(void)
 	if (ret)
 		goto fail;
 
+	memset(&rocr_agents, 0, sizeof(rocr_agents));
+
+	hsa_ret = ofi_hsa_iterate_agents(rocr_hsa_agent_callback, NULL);
+	if (hsa_ret != HSA_STATUS_SUCCESS &&
+	    hsa_ret != HSA_STATUS_INFO_BREAK)
+	    goto fail;
+
```
edgargabriel, line 898: I think the patch is missing a piece of code in rocr_hmem_dl_init(), something like

   hsa_ops.hsa_iterate_agents = dlsym(hsa_handle, "hsa_iterate_agents");
        if (!hsa_ops.hsa_iterate_agents) {
                FI_WARN(&core_prov, FI_LOG_CORE,
                        "Failed to find hsa_iterate_agents\n");
                goto err;
        }

zachdworkin:
Please rebase to pick up our newest CI changes

amirshehataornl:
Let's hold off on landing this patch. I have a few tests I'm doing first to ensure things work as expected.

include/ofi_mr.h
```diff
@@ -56,6 +56,7 @@ struct ofi_mr_info {
 	struct iovec iov;
 	enum fi_hmem_iface iface;
 	uint64_t device;
+	uint64_t id;
```
shefty, line 59: Please add a blank line above id and rename id to peer_id.  (The blank line is to help identify the field as an ipc related value.)  Please rename other places where you added id to peer_id, because without the commit message, what id actually is isn't obvious.

shefty (changes requested):
minor change requested

include/ofi_mr.h
```diff
@@ -56,7 +56,8 @@ struct ofi_mr_info {
 	struct iovec iov;
 	enum fi_hmem_iface iface;
 	uint64_t device;
-	uint64_t id;
+
+	uint64_t peer_id;
```
shefty, line 60: these changes landed in the wrong commit

edgargabriel (commented):
I ran some tests on internal resources using this branch, and it looks good from our point of view.

shefty:
Intel CI and appveyor failures are both with sockets provider.

PR #8761, opened 2023-04-06
prov/verbs: code restructuring

shefty:
These changes are intended as a starting point in optimizing the verbs provider, and the eventually addition of rdm ep support directly into the provider.  The changes are based on the optimizations integrated into the tcp provider.  As a starting point, the code separates the 1:1 relationship between a CQ exposed to the application and the HW CQ allocated through verbs.  The verbs CQ is instead moved to the domain and hidden from the app.  The reasoning is that in order to drive progress over an endpoint, both the send and receive queues must be checked.  For example, the receive queue is checked to handle acks, flow control messages, etc.  An app may create a send-only endpoints, but internally, some operations will still need receive handling.

The resulting code ends up quite a bit simpler and should simplify further as optimizations are moved up the call path.

These changes are incomplete.  (I'm sure the locking is messed up.)  I'm posting early for feedback, as I'm unsure if I want to break the 1:1 relationship between a libfabric CQ and HW CQ.  An alternative is to maintain that mapping, but implement the other optimizations involved in moving progress up to the domain.

One downside to the changes is if the app uses a shared Rx and has multiple endpoints bound to *separate* CQs, then there's no easy way to figure out which libfabric CQ to write a completion to.  A lookup using the qpn in the verbs completion is needed to identify the libfabric ep, which then points to the cq.  I'm skeptical any apps are actually doing this, since it's technically silly.  It's worth noting that XRC QPs already have this restriction.

Another downside is that this essentially sets the threading model to FI_THREAD_DOMAIN.  Apps that use msg ep's with separate cq's allocated to different threads could see an impact to performance.  This downside is likely even if a 1:1 mapping is maintained, but the progress model is reworked.

### DIFF
--- a/libfabric.vcxproj
+++ b/libfabric.vcxproj
@@ -989,7 +989,7 @@
   </ItemGroup>
   <ItemGroup Condition="'$(Configuration)|$(Platform)'!='Debug-Efa-v142|x64' AND '$(Configuration)|$(Platform)'!='Release-Efa-v142|x64'">
     <ClInclude Include="prov\verbs\include\fi_verbs.h" />
-    <ClInclude Include="prov\verbs\include\ofi_verbs_priv.h" />
+    <ClInclude Include="prov\verbs\include\ofi_verbs_compat.h" />
     <ClInclude Include="prov\verbs\include\windows\verbs_nd.h" />
     <ClInclude Include="prov\verbs\include\windows\verbs_osd.h" />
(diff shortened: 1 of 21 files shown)


shefty:
removed CQ update changes - only kept code restructures, which are possible to merge.  I'll separate out the CQ updates into a series of changes, so that some can be merged earlier.

shefty:
closing to fix windows build, will re-open once verified

PR #8764, opened 2023-04-06
prov/shm: increase default rx size and print warning

wzamazon:
Two commits:

1. increase default RX size
2. print warning when `smr_alloc_cmd_ctx` failed.

### DIFF
--- a/prov/shm/src/smr_attr.c
+++ b/prov/shm/src/smr_attr.c
@@ -58,7 +58,7 @@ struct fi_rx_attr smr_rx_attr = {
 	.op_flags = SMR_RX_OP_FLAGS,
 	.comp_order = FI_ORDER_STRICT,
 	.msg_order = SMR_RMA_ORDER | FI_ORDER_SAS,
-	.size = 1024,
+	.size = 16384,
 	.iov_limit = SMR_IOV_LIMIT
 };
 
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -875,8 +875,12 @@ static int smr_progress_cmd_msg(struct smr_ep *ep, struct smr_cmd *cmd)
 		ofi_ep_lock_release(&ep->util_ep);
 		if (ret == -FI_ENOENT) {
 			ret = smr_alloc_cmd_ctx(ep, rx_entry, cmd);
-			if (ret)
+			if (ret) {
+				FI_WARN(&smr_prov, FI_LOG_EP_CTRL,
+					"smr_alloc_cmd_ctx failed! ret: %d\n",
+					ret);
 				return ret;
+			}
 
 			ret = peer_srx->owner_ops->queue_tag(rx_entry);
 			goto out;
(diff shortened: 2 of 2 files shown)


prov/shm/src/smr_attr.c
```diff
@@ -58,7 +58,7 @@ struct fi_rx_attr smr_rx_attr = {
 	.op_flags = SMR_RX_OP_FLAGS,
 	.comp_order = FI_ORDER_STRICT,
 	.msg_order = SMR_RMA_ORDER | FI_ORDER_SAS,
-	.size = 1024,
+	.size = 16384,
```
shijin-aws, line 61: do we know how 16384 is sufficiently enough? and how much more memory this could cost?

prov/shm/src/smr_progress.c
```diff
@@ -875,8 +875,12 @@ static int smr_progress_cmd_msg(struct smr_ep *ep, struct smr_cmd *cmd)
 		ofi_ep_lock_release(&ep->util_ep);
 		if (ret == -FI_ENOENT) {
 			ret = smr_alloc_cmd_ctx(ep, rx_entry, cmd);
-			if (ret)
+			if (ret) {
```
shijin-aws, line 878: there is one more places that this could fail, the `queue_msg` path

prov/shm/src/smr_attr.c
wzamazon: > do we know how 16384 is sufficiently enough? 

This would depend on application. This value can be overridden by an env var FI_SHM_RX_SIZE.

>  and how much more memory this could cost?

This value is number of `smr_cmd_ctx`. Each `smr_cmd_ctx` is 280 bytes. So set this value to 16384 increase memory usage by ~4.5 MB per endpoint.

prov/shm/src/smr_progress.c
wzamazon: added

shijin-aws (approved):

shijin-aws:
I am wondering whether there is a better way to handle this rx resource exhaustion issue. Making application set `FI_SHM_RX_SIZE` makes me unsettled.

aingerson:
The rx_size determines the number of cmd_ctxs as well as the number of commands in the shm AND the number of inject buffers in the shared memory. This is a HUGE addition of memory here. Is the concern the number of unexpected messages (cmd_ctxs)? If so, we should made the cmd_ctxs use the buf pool code which dynamically increases as needed.
What exactly is the concern/failure/use case that this addresses?

wzamazon:
> Is the concern the number of unexpected messages (cmd_ctxs)? 

Yes.  

> What exactly is the concern/failure/use case that this addresses?

The concern as you mentioned is that when processing unexpected messages, `smr_alloc_cmd_ctx` would fail, and the message did not get queued. causing application to hang.

> If so, we should made the cmd_ctxs use the buf pool code which dynamically increases as needed.

That sounds like a better solution. Do you have time to implement it?

wzamazon:
Closed in favor of use buf pool for cmd_ctx

PR #8766, opened 2023-04-06
prov/shm: switch cmd_ctx pool from freestack to bufpool

aingerson:
The command context can run out of memory if too many incoming unexpected messages are posted. Switching to a bufpool allows shm to dynamically allocate more memory as needed to accommodate higher unexpected traffic.

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -944,7 +944,8 @@ static int smr_ep_close(struct fid *fid)
 	if (ep->util_ep.ep_fid.msg != &smr_no_recv_msg_ops)
 		smr_srx_close(&ep->srx->fid);
 
-	smr_cmd_ctx_fs_free(ep->cmd_ctx_fs);
+	if (ep->cmd_ctx_pool)
+		ofi_bufpool_destroy(ep->cmd_ctx_pool);
 	smr_tx_fs_free(ep->tx_fs);
 	smr_pend_fs_free(ep->pend_fs);
 	ofi_spin_destroy(&ep->tx_lock);
@@ -1517,7 +1518,7 @@ static int smr_discard(struct fi_peer_rx_entry *rx_entry)
 {
 	struct smr_cmd_ctx *cmd_ctx = rx_entry->peer_context;
 
-	ofi_freestack_push(cmd_ctx->ep->cmd_ctx_fs, cmd_ctx);
+	ofi_buf_free(cmd_ctx);
 	return FI_SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


aingerson:
@wzamazon @shijin-aws  Test this out and let me know if this does or does not solve the issue you're seeing!

prov/shm/src/smr_ep.c
```diff
@@ -1800,7 +1801,14 @@ int smr_endpoint(struct fid_domain *domain, struct fi_info *info,
...
 
-	ep->cmd_ctx_fs = smr_cmd_ctx_fs_create(info->rx_attr->size, NULL, NULL);
+	ret = ofi_bufpool_create(&ep->cmd_ctx_pool, sizeof(struct smr_cmd_ctx),
+				 16, 0, info->rx_attr->size,
+				 OFI_BUFPOOL_NO_TRACK);
+	if (ret) {
+		FI_WARN(&smr_prov, FI_LOG_EP_CTRL,
+			"Unable to create cmd ctx pool\n");
+		return ret;
+	}
```
wzamazon, line 1811: I suggest call `ofi_bufpool_grow` here, so every endpoint start with 1 chunk of `cmd_ctx`.

Otherwise, different endpoint allocate memory at different time, which can cause overall delay.

prov/shm/src/smr_progress.c
```diff
@@ -849,10 +849,7 @@ static int smr_alloc_cmd_ctx(struct smr_ep *ep,
 {
 	struct smr_cmd_ctx *cmd_ctx;
 
-	if (ofi_freestack_isempty(ep->cmd_ctx_fs))
-		return -FI_EAGAIN;
-
-	cmd_ctx = ofi_freestack_pop(ep->cmd_ctx_fs);
+	cmd_ctx = ofi_buf_alloc(ep->cmd_ctx_pool);
```
wzamazon, line 852: Do you need to check against `cmd_ctx` against `NULL`?

prov/shm/src/smr_ep.c
aingerson: Hmm, I thought alloc called grow on initialization but I guess I'm wrong. I will add. Thanks!

shijin-aws: It is actually lazy initialization.

aingerson:
Updated

prov/shm/src/smr_progress.c
```diff
@@ -849,10 +849,9 @@ static int smr_alloc_cmd_ctx(struct smr_ep *ep,
...
+	if (!cmd_ctx)
```
wzamazon, line 853: Can you add a `FI_WARN` here too?

wzamazon (approved):

shijin-aws (approved):

wzamazon:
> Test this out and let me know if this does or does not solve the issue you're seeing!

@aingerson  Test is good from our side

shijin-aws:
>@aingerson Test is good from our side
@wzamazon No performance change? I was curious whether buf pool is less or more performant than free stack

wzamazon:
>No performance impact? I was curious whether buf pool is less or more performant than free stack

No significant difference in performance from the test I ran.

shijin-aws:
>No significant difference in performance from the test I ran.

If I understand correctly, this patch still keeps the bufpool of initial size 1024, and it will be finally grown dynamically later when it is exhausted, right?

wzamazon:
> If I understand correctly, this patch still keeps the bufpool of initial size 1024, and it will be finally grown dynamically later when it is exhausted, right?

Yes. That is my understanding.

aingerson:
Correct. I guess the question is how many times are you seeing it having to grow? ie what is the maximum number of unexpected messages you're seeing at any one time?

shijin-aws:
I am curious to see how many times `ofi_bufpool_grow` is called after the ep init.

wzamazon:
> Correct. I guess the question is how many times are you seeing it having to grow? ie what is the maximum number of unexpected messages you're seeing at any one time?

I did not check. I know it is less than 16384 because that number let my application pass.

wzamazon:
@aingerson

Can you help check the "continuous-integration/jenkins/pr-head" CI error?

aingerson:
@wzamazon Our cluster went down this week and we've been trying to get it back up but have had some lingering issues. The build passed but not all tests were run including shm. I retriggered it.

shijin-aws:
test passed, can we merge now?

PR #8767, opened 2023-04-06
prov/shm & prov/sm2: Add FI_LOCAL_COMM to SHM's rx/tx caps

a-szegel:
FI_LOCAL_COMM
At a conceptual level, this field indicates that the underlying device supports loopback communication. More specifically, this field indicates that an endpoint may communicate with other endpoints that are allocated from the same underlying named domain. If this field is not set, an application may need to use an alternate domain or mechanism (e.g. shared memory) to communicate with peers that execute on the same node.

Seems like SHM/SM2 should advertise FI_LOCAL_COMM from that definition.


shijin-aws (approved):

a-szegel:
The only reason I think it would make sense to to support this is if we fail the EP sending to itself.... which is weird and possibly a user error?

shijin-aws (commented):
This is actually SM2 change

a-szegel:
Whoops let me fix that

prov/shm/src/smr_attr.c
```diff
@@ -32,10 +32,10 @@
 
 #include "smr.h"
 
-#define SMR_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED | OFI_TX_RMA_CAPS | FI_ATOMICS)
+#define SMR_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED | OFI_TX_RMA_CAPS | FI_ATOMICS | FI_LOCAL_COMM)
 #define SMR_RX_CAPS (FI_SOURCE | FI_RMA_EVENT | OFI_RX_MSG_CAPS | FI_TAGGED | \
 		     OFI_RX_RMA_CAPS | FI_ATOMICS | FI_DIRECTED_RECV | \
-		     FI_MULTI_RECV)
+		     FI_MULTI_RECV | FI_LOCAL_COMM)
```
shefty, line 38: FI_LOCAL_COMM does not apply to either tx or rx caps.  It only applies to fi_info::caps and fi_domain_attr::caps

shefty:
Btw, the man pages explicitly list what caps bits are applicable for fi_info::caps, fi_domain_attr::caps, fi_tx_attr::caps, fi_rx_attr::caps

a-szegel:
Will put up another PR to move these lower.

PR #8770, opened 2023-04-07
fabtests: Make rdm_tagged_peek more general

shijin-aws:
The current rdm_tagged_peek test uses sread which is not
    available for some provider like efa. This patch
    makes this test more general to support all available
    comp_method, including sread and the default non-wait read.
    It also makes the do_send post tx via ft_post_tx_buf so it
    can have retry/progressing procedure to handle the EAGAIN
    case


shijin-aws:
I tested it with tcp with and without `-c sread`. Should work for both cases.

aingerson:
@shijin-aws Haven't looked through the change but I think conceptually that's fine. Just wanted to mention that we added a wait object for providers that don't support waiting but still want to support the sread API through an internal spin. It's called FI_WAIT_YIELD and it is what shm uses by default in order to support fi_cq_sread. See the fi_cq man page as well as the util_wait.c code in case you want to implement it for efa to avoid that limitation.

fabtests/functional/rdm_tagged_peek.c
```diff
@@ -282,15 +263,15 @@ static int do_sends(void)
 	(void) ft_fill_buf(tx_buf, tx_size);
 
 	for (i = 1; i <= SEND_CNT; i++) {
-		ret = fi_tsend(ep, tx_buf, tx_size, mr_desc,
-				remote_fi_addr, BASE_TAG + i,
-				&tx_ctx_arr[i].context);
+		ret = ft_post_tx_buf(ep, remote_fi_addr, tx_size, NO_CQ_DATA, &tx_ctx_arr[i].context, tx_buf, mr_desc, BASE_TAG + i);
```
shefty, line 266: ugh... please don't make me scroll left and right

fabtests/functional/rdm_tagged_peek.c
```diff
@@ -44,26 +44,8 @@
...
-		ret = fi_cq_sread(txcq, &comp, 1, NULL, -1);
-		if (ret != 1) {
-			FT_PRINTERR("fi_cq_sread", ret);
-			return ret;
-		}
-		completions++;
-	} while (completions < count);
-
-	return 0;
-}
```
shefty, line 62: Why did you remove this function, rather than just changing it, so it doesn't call fi_cq_sread?

fabtests/functional/rdm_tagged_peek.c
```diff
@@ -89,12 +71,11 @@ static int trecv_op(uint64_t tag, uint64_t flags, bool ignore_nomsg)
...
 			return ret;
 		}
-
-		ret = fi_cq_sread(rxcq, &comp, 1, NULL, -1);
-		if (ret == 1) {
-			ret = 0;
+		ft_tag = tag;
+		rx_cq_cntr = 0;
+		ret = ft_read_cq(rxcq, &rx_cq_cntr, 1, timeout);
+		if (ret == FI_SUCCESS)
```
shefty, line 77: Can we change wait_for_send_comp() to wait_for_comp(), pass in the cq to check, and call it from here?

fabtests/functional/rdm_tagged_peek.c
```diff
@@ -310,13 +291,11 @@ static int run(void)
 		/* sync with sender before ft_finalize, since we sent
 		 * and received messages outside of the sequence numbers
 		 * maintained by common code */
-		ret = fi_tsend(ep, tx_buf, 1, mr_desc,
-				remote_fi_addr, 0xabc,
-				&tx_ctx_arr[0].context);
+		ret = ft_post_tx_buf(ep, remote_fi_addr, 1, NO_CQ_DATA, &tx_ctx_arr[0].context, tx_buf, mr_desc, 0xabc);
```
shefty, line 294: scrolling...

fabtests/functional/rdm_tagged_peek.c
shijin-aws: I thought it does the same thing with `ft_get_tx_comp` for sread. Why can't we replace it with the function in the common code?

fabtests/functional/rdm_tagged_peek.c
shijin-aws: So you want a single `wait_for_comp()` that can handle both send/recv completion in this test, right?

shefty: The change below to call ft_get_tx_comp() directly instead of wait_for_send_comp() makes sense.  Maybe we also need ft_get_rx_comp() as a peer?  It's the direct setting of ft_tag and rx_cq_cntr that look odd here.  (And something that we someday need to address to make these tests easier to maintain.

fabtests/functional/rdm_tagged_peek.c
```diff
@@ -330,7 +309,19 @@ static int run(void)
...
+	 * ft_finalize() will post a send and its tag (the tx_seq here)
+	 * should match the tag used by the last recv posted by ft_rx, which
+	 * should be rx_seq - 1 (the rx_cq_cntr at that time).
+	 * And the rx_cq_cntr should also be set to rx_seq - 1 so ft_get_rx_comp()
+	 * will poll 1 cqe.
+	 */
+	tx_seq = rx_seq - 1 ;
+	tx_cq_cntr = tx_seq;
+	rx_cq_cntr = rx_seq - 1;
+	ft_tag = 0;
```
shefty, line 324: Can we somehow not need the above code?  This is a difficulty for maintainability.  If there's no easy way around it for now, we can come back someday to try to address.

fabtests/functional/rdm_tagged_peek.c
shijin-aws: I thought about using `ft_get_rx_comp`, one thing a bit tricky is `trecv_op` could retry the trecv and cq poll when `ignore_nomsg` is true, which will makes the rx_seq/rx_cq_cntr messed up. About the ft_tag set,  AFAICT it is set as 0 by default in `shared.c` and overridden by test code during the test. The common code in `shared.c` will use it  when it is non-zero otherwise it will use tx/rx_cq_cntr as the tag. I am not sure whether it is the expected workflow for using `ft_tag` but it really seems foggy to me. Should we make tag an argument of `ft_post_tx/rx_buf` instead. `ft_tx`/`ft_rx` can still use cq cntr as tags, but it leaves callers flexibility to customize the tag.

shefty: As for a single wait_for_comp() -- don't add that yet (and maybe not)

shijin-aws: Also `ft_get_rx_comp` will always log error if there is any. But the test wants to ignore enomesg error if requested. That is why I break `ft_get_*_comp` to two parts, 1 for `ft_read_cq` and the other for `fi_cq_readerr`. If I use `ft_get_rx_comp` in recv path I will lose this flexibility, do you have better ideas here?

shefty: The global tag is likely a mistake.  I haven't looked into details as to how to start restructuring fabtests, but I know the current approach makes it difficult to isolate test changes.  If you think it helps adding tag as a parameter to some of the lower-level calls and using those directly, I'd say go for it.  We can review those changes until we're happy with them, then look at adopting them to other tests.

shefty: I don't have better ideas.  I don't think the problems are in your changes or this test, but the common code.  And there's no easy way to handle that without a deep dive into serious restructuring of the code.  I think we accept the above changes for now and figure out how to handle a broader cleanup later.

shijin-aws: > If you think it helps adding tag as a parameter to some of the lower-level calls and using those directly

Thanks. I think I will try in this direction to get rid of `ft_tag` and the shenanigans with the tx/rx_seq, cntr at the cleanup stage. I agree that the current implementation is hard for maintenance.

fabtests/functional/rdm_tagged_peek.c
shijin-aws: Removed this hack in latest revision.

fabtests/functional/rdm_tagged_peek.c
shijin-aws: Fixed.

fabtests/functional/rdm_tagged_peek.c
shijin-aws: @shefty  I think I had a simpler solution in the latest revision so I can get rid all the hack on `ft_tag`, `tx/rx_seq` and `tx/rx_cq_cntr`. Please see the latest revision.

fabtests/functional/rdm_tagged_peek.c
shijin-aws: Fixed.

shefty (commented):
Looks cleaner -- thanks.  Waiting on CI.

shijin-aws:
@shefty what is the error in continuous-integration/jenkins/pr-head?

shefty:
abort of a bunch of tests -- we lost a network switch and are dealing with system network shuffling (aka system unavailable)

shefty (approved):
ignore intel CI (for now)

PR #8772, opened 2023-04-07
include/ofi_iov.h: locate iov_index and iov_offset for an offset

wzamazon:
This patch introduced two functions:

    ofi_iov_locate() and
    ofi_rma_iov_locate()

to find iov_index and iov_offset for an offset


shefty:
These look good to me.  Do you want to optimize for the case that iov_count = 1?

wzamazon:
> Do you want to optimize for the case that iov_count = 1?

Added in new revision.

shefty (approved):

wenduwan (approved):

shefty:
Intel CI failure is because we have some nodes unavailable.  It can be ignored

PR #8773, opened 2023-04-07
src/hmem_cuda: Restrict CUDA IPC based on peer accessibility of devices + refactor

darrylabbate:
This PR:
- Implements logic to check the P2P capabilities of the detected CUDA devices, and conditionally enable CUDA IPC availability
    - If any device cannot access some other device, consider peer access to be unsupported on the host
        - "peer access" in this context refers to the peer-to-peer access capabilities between CUDA devices on the same node
    - If peer access is unsupported, prevent CUDA IPC usage
        - This mitigates an issue where  `cudaIpcOpenMemHandle()` can fail on hosts with multiple CUDA devices without P2P access - e.g. Amazon EC2 instance types g4dn.12xlarge or g5.48xlarge
        - This logic is admittedly not perfect, but is an improvement over simply checking for `cudaMemcpy()` availability
    - This change still allows CUDA IPC usage when the host has a single device, regardless of its peer capabilities
        - I also tested `CUDA_VISIBLE_DEVICES=1` on a g5.48xlarge EC2 instance (8 CUDA devices w/o peer access support) to confirm this works. As a future improvement, we may also be able to allow the user to limit the number of GPUs discoverable by Libfabric via `cudaSetValidDevices()`
- Refactors the file-scoped variables (e.g. `cuda_ipc_enabled`) into a single `cuda_attr` struct
    - Variables were inconsistently named and becoming increasingly unruly for every new dimension added to the CUDA HMEM interface (`cuda_xfer_setting`, GDR Copy usage, etc)
- Adds some preprocessor hullabaloo to reduce boilerplate
    - Notably the long `dlsym()` sequence in `cuda_hmem_dl_init()` for the `cuda_ops` function pointers

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -62,6 +62,8 @@ CUresult ofi_cuGetErrorName(CUresult error, const char** pStr);
 CUresult ofi_cuGetErrorString(CUresult error, const char** pStr);
 CUresult ofi_cuPointerGetAttribute(void *data, CUpointer_attribute attribute,
 				   CUdeviceptr ptr);
+CUresult ofi_cuDeviceCanAccessPeer(int *canAccessPeer, CUdevice srcDevice,
+				   CUdevice dstDevice);
 cudaError_t ofi_cudaHostRegister(void *ptr, size_t size, unsigned int flags);
 cudaError_t ofi_cudaHostUnregister(void *ptr);
 cudaError_t ofi_cudaMalloc(void **ptr, size_t size);
--- a/prov/efa/src/efa_user_info.c
+++ b/prov/efa/src/efa_user_info.c
@@ -302,7 +302,7 @@ bool efa_user_info_should_support_hmem(int version)
 	if (hmem_ops[FI_HMEM_CUDA].initialized && FI_VERSION_GE(version, FI_VERSION(1, 18))) {
 		EFA_INFO(FI_LOG_CORE,
 			"User is using API version >= 1.18. CUDA library and "
-			"device are available, claim support of FI_HMEM.");
+			"devices are available, claim support of FI_HMEM.\n");
 			/* For this API we can support HMEM regardless of
 			   use_device_rdma and P2P support, because we can use
 			   CUDA api calls.*/
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -474,6 +474,10 @@ static int cuda_hmem_verify_devices(void)
 		return -FI_EIO;
 	}
 
+	FI_INFO(&core_prov, FI_LOG_CORE,
+		"Number of CUDA devices detected: %d\n",
+		cuda_attr.device_count);
+
 	if (cuda_attr.device_count <= 0)
 		return -FI_ENOSYS;
 
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -56,6 +56,7 @@
 	_(cuGetErrorString)		\
 	_(cuPointerGetAttribute)	\
 	_(cuPointerSetAttribute)	\
+	_(cuDeviceCanAccessPeer)	\
 	_(cuMemGetAddressRange)
 
 #define CUDA_RUNTIME_FUNCS_DEF(_)	\
@@ -76,18 +77,20 @@
 
 static struct {
 	int   device_count;
+	bool  p2p_access_supported;
 	bool  use_gdrcopy;
 	bool  use_ipc;
 	void *driver_handle;
 	void *runtime_handle;
 	enum cuda_xfer_setting xfer_setting;
 } cuda_attr = {
-	.device_count   = -1,
-	.use_gdrcopy    = false,
-	.use_ipc        = false,
-	.driver_handle  = NULL,
-	.runtime_handle = NULL,
-	.xfer_setting   = CUDA_XFER_UNSPECIFIED
+	.device_count         = -1,
+	.p2p_access_supported = false,
+	.use_gdrcopy          = false,
+	.use_ipc              = false,
+	.driver_handle        = NULL,
+	.runtime_handle       = NULL,
+	.xfer_setting         = CUDA_XFER_UNSPECIFIED
 };
 
 static struct {
@@ -108,6 +111,8 @@ static struct {
 					  CUdeviceptr ptr);
 	CUresult (*cuMemGetAddressRange)( CUdeviceptr* pbase,
 					  size_t* psize, CUdeviceptr dptr);
+	CUresult (*cuDeviceCanAccessPeer)(int *canAccessPeer,
+					  CUdevice srcDevice, CUdevice dstDevice);
 	cudaError_t (*cudaHostRegister)(void *ptr, size_t size,
 					unsigned int flags);
 	cudaError_t (*cudaHostUnregister)(void *ptr);
@@ -220,6 +225,12 @@ int cuda_set_sync_memops(void *ptr)
 CUresult ofi_cuMemGetAddressRange(CUdeviceptr* pbase, size_t* psize, CUdeviceptr dptr)
 {
 	return cuda_ops.cuMemGetAddressRange(pbase, psize, dptr);
+}
+
+CUresult ofi_cuDeviceCanAccessPeer(int *canAccessPeer, CUdevice srcDevice,
+				   CUdevice dstDevice)
+{
+	return cuda_ops.cuDeviceCanAccessPeer(canAccessPeer, srcDevice, dstDevice);
 }
 
 cudaError_t ofi_cudaHostRegister(void *ptr, size_t size, unsigned int flags)
@@ -484,6 +495,54 @@ static int cuda_hmem_verify_devices(void)
 	return FI_SUCCESS;
 }
 
+/**
+ * @brief   Determine overall peer access support for the CUDA HMEM interface
+ *
+ * This checks each CUDA device visible to Libfabric for peer accessibility to
+ * the next visible device. #cuda_attr.p2p_access_supported is set to true only
+ * if every query between two devices indicates peer access support. Otherwise
+ * the function returns early, leaving #cuda_attr.p2p_access_supported
+ * unmodified.
+ *
+ * @return  FI_SUCCESS if peer access check(s) are successful
+ *         -FI_EIO upon CUDA API error
+ */
+static int cuda_hmem_detect_p2p_access_support(void)
+{
+	CUresult cuda_ret;
+	CUdevice dev, peer;
+	int can_access_peer = 1;
+
+	if (cuda_attr.device_count <= 1)
+		return FI_SUCCESS;
+
+	/*
+	 * CUDA API always enumerates available devices contiguously starting
+	 * from index 0.
+	 */
+	for (dev = 0; dev < cuda_attr.device_count - 1; ++dev) {
+		peer = dev + 1;
+		cuda_ret = ofi_cuDeviceCanAccessPeer(&can_access_peer, dev, peer);
+		if (CUDA_SUCCESS != cuda_ret) {
+			FI_WARN(&core_prov, FI_LOG_CORE,
+				"Failed to detect support for peer-to-peer "
+				"access between CUDA devices via "
+				"cuDeviceCanAccessPeer(): %s:%s\n",
+				ofi_cudaGetErrorName(cuda_ret),
+				ofi_cudaGetErrorString(cuda_ret));
+			return -FI_EIO;
+		}
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Peer access from CUDA device %d -> CUDA device %d : %s\n",
+			dev, peer, can_access_peer ? "Yes" : "No");
+		if (!can_access_peer)
+			return FI_SUCCESS;
+	}
+
+	cuda_attr.p2p_access_supported = true;
+	return FI_SUCCESS;
+}
+
 int cuda_hmem_init(void)
 {
 	int ret, param_value;
@@ -505,6 +564,10 @@ int cuda_hmem_init(void)
 		return ret;
 
 	ret = cuda_hmem_verify_devices();
+	if (ret != FI_SUCCESS)
+		goto dl_cleanup;
+
+	ret = cuda_hmem_detect_p2p_access_support();
 	if (ret != FI_SUCCESS)
 		goto dl_cleanup;
 
@@ -536,9 +599,15 @@ int cuda_hmem_init(void)
 		cuda_ops.cudaMemcpy = cuda_disabled_cudaMemcpy;
 
 	/*
-	 * CUDA IPC is only enabled if cudaMemcpy can be used.
+	 * CUDA IPC can be safely utilized if:
+	 * - All devices on the bus have peer access to each other
+	 *     - This includes configurations with only a single CUDA device,
+	 *       regardless of the device's p2p capabilities
+	 * - cudaMemcpy() is available
 	 */
-	cuda_attr.use_ipc = (cuda_attr.xfer_setting != CUDA_XFER_DISABLED);
+	cuda_attr.use_ipc =
+		(cuda_attr.p2p_access_supported || cuda_attr.device_count == 1) &&
+		(cuda_attr.xfer_setting != CUDA_XFER_DISABLED);
 
 	return FI_SUCCESS;
 
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -1,6 +1,6 @@
 /*
  * (C) Copyright 2020 Hewlett Packard Enterprise Development LP
- * (C) Copyright 2021-2022 Amazon.com, Inc. or its affiliates.
+ * (C) Copyright Amazon.com, Inc. or its affiliates.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
@@ -43,7 +43,54 @@
 #include <cuda.h>
 #include <cuda_runtime.h>
 
-struct cuda_ops {
+#if ENABLE_CUDA_DLOPEN
+#include <dlfcn.h>
+#endif
+
+/*
+ * Convenience higher-order macros for enumerating CUDA driver/runtime API
+ * function names
+ */
+#define CUDA_DRIVER_FUNCS_DEF(_)	\
+	_(cuGetErrorName)		\
+	_(cuGetErrorString)		\
+	_(cuPointerGetAttribute)	\
+	_(cuPointerSetAttribute)	\
+	_(cuMemGetAddressRange)
+
+#define CUDA_RUNTIME_FUNCS_DEF(_)	\
+	_(cudaMemcpy)			\
+	_(cudaDeviceSynchronize)	\
+	_(cudaFree)			\
+	_(cudaMalloc)			\
+	_(cudaGetErrorName)		\
+	_(cudaGetErrorString)		\
+	_(cudaHostRegister)		\
+	_(cudaHostUnregister)		\
+	_(cudaGetDeviceCount)		\
+	_(cudaGetDevice)		\
+	_(cudaSetDevice)		\
+	_(cudaIpcOpenMemHandle)		\
+	_(cudaIpcGetMemHandle)		\
+	_(cudaIpcCloseMemHandle)
+
+static struct {
+	int   device_count;
+	bool  use_gdrcopy;
+	bool  use_ipc;
+	void *driver_handle;
+	void *runtime_handle;
+	enum cuda_xfer_setting xfer_setting;
+} cuda_attr = {
+	.device_count   = -1,
+	.use_gdrcopy    = false,
+	.use_ipc        = false,
+	.driver_handle  = NULL,
+	.runtime_handle = NULL,
+	.xfer_setting   = CUDA_XFER_UNSPECIFIED
+};
+
+static struct {
 	cudaError_t (*cudaMemcpy)(void *dst, const void *src, size_t size,
 				  enum cudaMemcpyKind kind);
 	cudaError_t (*cudaDeviceSynchronize)(void);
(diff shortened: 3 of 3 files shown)


src/hmem_cuda.c
```diff
@@ -1,6 +1,6 @@
 /*
  * (C) Copyright 2020 Hewlett Packard Enterprise Development LP
- * (C) Copyright 2021-2022 Amazon.com, Inc. or its affiliates.
```
lrbison, line 3: we should just remove the date according to current guidance.

src/hmem_cuda.c
```diff
@@ -376,176 +388,51 @@ static cudaError_t cuda_disabled_cudaMemcpy(void *dst, const void *src,
...
-
-	cuda_ops.cudaIpcCloseMemHandle = dlsym(cudart_handle,
-					    STRINGIFY(cudaIpcCloseMemHandle));
-	if (!cuda_ops.cudaIpcCloseMemHandle) {
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find cudaIpcCloseMemHandle\n");
-		goto err_dlclose_cuda;
-	}
+	CUDA_DRIVER_FUNCS_DEF(CUDA_DRIVER_FUNCS_DLOPEN)
+	CUDA_RUNTIME_FUNCS_DEF(CUDA_RUNTIME_FUNCS_DLOPEN)
```
lrbison, line 440: beautiful!

src/hmem_cuda.c
```diff
@@ -76,14 +77,17 @@
...
 	bool  use_gdrcopy;
 	bool  use_ipc;
 	void *driver_handle;
 	void *runtime_handle;
 	enum cuda_xfer_setting xfer_setting;
 } cuda_attr = {
-	.driver_handle  = NULL,
-	.runtime_handle = NULL,
+	/* Assume CUDA device(s) are incapable of p2p access by default */
+	.p2p_access_supported = false,
```
lrbison, line 88: why not also initialize device_count and the other bools?  do we use p2p_access_supported differently somehow?

lrbison (commented):
A few small comments added

src/hmem_cuda.c
darrylabbate: I agree, though I'm curious if @shefty prefers the dates in the copyright notices. Otherwise we can start removing them as we update files.

src/hmem_cuda.c
darrylabbate: No real reason - just wanted to initialize `p2p_access_supported` with a default in this case. I opted to leave everything else untouched other than the driver/runtime handles. But I'm fine with initializing all the struct members.

src/hmem_cuda.c
shefty: No - the dates aren't needed.  FWIW, Intel now has the same guidance.

include/ofi_hmem.h
```diff
@@ -62,6 +62,7 @@ CUresult ofi_cuGetErrorName(CUresult error, const char** pStr);
 CUresult ofi_cuGetErrorString(CUresult error, const char** pStr);
 CUresult ofi_cuPointerGetAttribute(void *data, CUpointer_attribute attribute,
 				   CUdeviceptr ptr);
+CUresult ofi_cuDeviceGetP2PAttribute(int* value, CUdevice_P2PAttribute attrib, CUdevice srcDevice, CUdevice dstDevice);
```
shefty, line 65: please don't make me scroll left and right to read code

src/hmem_cuda.c
shefty: I think cuda_attr as defined will be initialized to all 0's anyway.

src/hmem_cuda.c
darrylabbate: Removed year range in latest push

src/hmem_cuda.c
darrylabbate: Initialized remaining struct members in latest update

include/ofi_hmem.h
darrylabbate: Oversight on my part; fixed

src/hmem_cuda.c
```diff
@@ -73,50 +125,19 @@ struct cuda_ops {
...
-static bool cuda_ipc_enabled;
-static int cuda_device_count;
+} cuda_ops
+#if !ENABLE_CUDA_DLOPEN
+#define CUDA_OPS_INIT(sym) .sym = sym,
+= {
+	CUDA_DRIVER_FUNCS_DEF(CUDA_OPS_INIT)
+	CUDA_RUNTIME_FUNCS_DEF(CUDA_OPS_INIT)
+}
+#endif
```
wenduwan, line 135: Oooooof. Stumped on this for a second.

For readability I think it's better to split declaration and initialization, e.g.
```
static struct {
	cudaError_t (*cudaMemcpy)(void *dst, const void *src, size_t size,
	cudaError_t (*cudaMemcpy)(void *dst, const void *src, size_t size,
				  enum cudaMemcpyKind kind);
				  enum cudaMemcpyKind kind);
	cudaError_t (*cudaDeviceSynchronize)(void);
	cudaError_t (*cudaDeviceSynchronize)(void);
@@ -61,6 +111,8 @@ struct cuda_ops {
					  CUdeviceptr ptr);
					  CUdeviceptr ptr);
	CUresult (*cuMemGetAddressRange)( CUdeviceptr* pbase,
	CUresult (*cuMemGetAddressRange)( CUdeviceptr* pbase,
					  size_t* psize, CUdeviceptr dptr);
					  size_t* psize, CUdeviceptr dptr);
	CUresult (*cuDeviceGetP2PAttribute)(int* value, CUdevice_P2PAttribute
			attrib, CUdevice srcDevice, CUdevice dstDevice);
	cudaError_t (*cudaHostRegister)(void *ptr, size_t size,
	cudaError_t (*cudaHostRegister)(void *ptr, size_t size,
					unsigned int flags);
					unsigned int flags);
	cudaError_t (*cudaHostUnregister)(void *ptr);
	cudaError_t (*cudaHostUnregister)(void *ptr);
@@ -73,50 +125,19 @@ struct cuda_ops {
	cudaError_t (*cudaIpcGetMemHandle)(cudaIpcMemHandle_t *handle,
	cudaError_t (*cudaIpcGetMemHandle)(cudaIpcMemHandle_t *handle,
					   void *devptr);
					   void *devptr);
	cudaError_t (*cudaIpcCloseMemHandle)(void *devptr);
	cudaError_t (*cudaIpcCloseMemHandle)(void *devptr);
};
} cuda_ops;

#if !ENABLE_CUDA_DLOPEN
cuda_ops = ...
#endif
```

src/hmem_cuda.c
```diff
@@ -582,12 +484,55 @@ static int cuda_hmem_verify_devices(void)
...
+static int cuda_hmem_detect_p2p_access_support(void)
+{
+	CUresult cuda_ret;
+	CUdevice dev, peer;
+	int can_access_peer = 1;
+
+	if (cuda_attr.device_count <= 1)
+		return FI_SUCCESS;
+
+	for (dev = 0; dev < cuda_attr.device_count - 1; ++dev) {
```
wenduwan, line 508: Thought experiment. This loop exhausts all pairs and only set `p2p_access_supported=true` if all of them supported p2p - is this an invariant from Nvidia? Any chance for partial p2p support, and what do you think should be the correct behavior for libfabric?

darrylabbate: This is kind of why I consider this solution imperfect. However, supposedly NVIDIA [doesn't support P2P communications between "dissimilar GPUs"](https://forums.developer.nvidia.com/t/problem-with-ipc/62039/2), so unless a user's topology is composed of multiple identical CUDA devices where only some GPUs have P2P enabled, we can't really do much. Even with a "weird" topology like that, a user still may be able to hide dissimilar GPUs from Libfabric via `CUDA_VISIBLE_DEVICES`. In which case, the p2p check should pass if all remaining devices have peer access support.

Overall I'm happy with this solution since we were initially concerned we may have to disable CUDA IPC even for single-device configurations.

Keep in mind, the current logic causes actual failures in some use cases on known GPU configurations (g4dn, g5 EC2 instances). So we're prioritizing correctness/functionality over optimizations for a niche/theoretical use case.

src/hmem_cuda.c
darrylabbate: That wouldn't work as a drop-in replacement without the struct being defined/`typedef`'d before declaring `cuda_ops`. I intentionally converted this into an untyped struct literal since `cuda_ops` is a singleton object.

src/hmem_cuda.c
```diff
@@ -376,176 +392,51 @@ static cudaError_t cuda_disabled_cudaMemcpy(void *dst, const void *src,
...
+ * Trailing semicolon is necessary for use with higher-order macro
+ */
+#define CUDA_FUNCS_DLOPEN(type, sym) \
+	do { \
+		cuda_ops.sym = dlsym(cuda_attr.type##_handle, #sym); \
+		if (!cuda_ops.sym) { \
+			FI_WARN(&core_prov, FI_LOG_CORE, "Failed to find " #sym "\n"); \
+			goto err_dlclose_cuda_driver; \
+		} \
+	} while (0);
```
aingerson, line 417: Not sure what the style guide has to say about this (@shefty can probably answer) but usually we line up the \s

src/hmem_cuda.c
```diff
@@ -630,20 +521,20 @@ int cuda_hmem_init(void)
 	param_value = 1;
 	ret = fi_param_get_bool(NULL, "hmem_cuda_enable_xfer", &param_value);
 	if (ret) {
-		hmem_cuda_xfer_setting = CUDA_XFER_UNSPECIFIED;
+		cuda_attr.xfer_setting = CUDA_XFER_UNSPECIFIED;
 	} else {
-		hmem_cuda_xfer_setting = (param_value > 0) ? CUDA_XFER_ENABLED : CUDA_XFER_DISABLED;
+		cuda_attr.xfer_setting = (param_value > 0) ? CUDA_XFER_ENABLED : CUDA_XFER_DISABLED;
```
aingerson, line 584: while you're here, can you fix the if/else to remove the {}

include/ofi_hmem.h
```diff
@@ -62,6 +62,8 @@ CUresult ofi_cuGetErrorName(CUresult error, const char** pStr);
 CUresult ofi_cuGetErrorString(CUresult error, const char** pStr);
 CUresult ofi_cuPointerGetAttribute(void *data, CUpointer_attribute attribute,
 				   CUdeviceptr ptr);
+CUresult ofi_cuDeviceGetP2PAttribute(int* value, CUdevice_P2PAttribute attrib,
```
aingerson, line 65: For consistency with the rest of the code - `int *value`

aingerson (commented):
Just a couple of style comments that are not a big deal but logic looks good to me!

src/hmem_cuda.c
shefty: I'm not sure the guide says anything about this.  We usually do line up the '\' as it makes it a little easier to identify the define block, but I'm okay either way.

darrylabbate: I prefer the `\` to be aligned, but with tabs set to anything other than 8 spaces (I prefer 4), it looks pretty bad due to the varying levels of indentation in this particular macro.

shefty: The coding style is tabs are tabs, which consist of 8 spaces.  We only need to worry about alignment around that constraint, which is also needed for indentation alignment anyway.

include/ofi_hmem.h
darrylabbate: I'll change this one, but not all the type signatures here "bind" `*`/`**` to the variable (probably just pasted from the NVIDIA docs like I did here). I'll go ahead and change those as well.

src/hmem_cuda.c
darrylabbate: Ack. This can actually be simplified a bit since I initialize this to `CUDA_XFER_UNSPECIFIED` anyway

wzamazon: Actually we should deprecate xfer_setting because we introduced FI_OPT_CUDA_API_PERMITTED. 

Can we just remove it (maybe a separate patch)?

src/hmem_cuda.c
```diff
@@ -482,6 +493,46 @@ static int cuda_hmem_verify_devices(void)
 	return FI_SUCCESS;
 }
 
+/*
+ * Loop over pairs of available devices to deduce the overall P2P capabilities
+ */
+static int cuda_hmem_detect_p2p_access_support(void)
```
shijin-aws, line 499: I am bit concerned this loop will considerably increase the initialization time cost.

darrylabbate: Tested this with 2 MPI processes on a single node with a g4dn.12xlarge (8 GPUs) and altered the code to let the full loop execute, and every log statement after calling `cuDeviceGetP2PAttribute()` (112 total calls) had the exact same timestamp - indicating little initialization overhead.

Can't say how it scales in practice beyond 8 GPUs per node, but this looks perfectly fine to me.

shijin-aws: I am curious to see how it impacts the performance for more ranks per node. But I do not think having more ranks per node could change the story much since this is only called once during initialization.

shijin-aws: But can we just simply check 1 pair? This loop looks a bit over-engineering to me.

darrylabbate: > But can we just simply check 1 pair? This loop looks a bit over-engineering to me.

I'd be more comfortable checking peer access between GPU 0 and the rest of the CUDA devices, which reduces the time complexity to $O(n)$. Checking a single pair doesn't feel bulletproof enough for me.

shijin-aws: O(n) is better than O(n^2) to me

darrylabbate: Actually on second thought, this should be just as effective as the current $O(n^2)$ search:

```c
for (dev = 0, peer = 1; peer < cuda_attr.device_count; dev += 2, peer += 2) {
    ...
}
```

This way each GPU is checked only once

darrylabbate:
Important detail I overlooked in my original commits: `cuDeviceGetP2PAttribute(..., CU_DEVICE_P2P_ATTRIBUTE_ACCESS_SUPPORTED ...)` is not equivalent to `cuDeviceCanAccessPeer()`. The former will only check if two devices can communicate via peer access in the currently enabled contexts (i.e. peer access must be explicitly enabled first), whereas the latter actually checks if the two devices are capable of peer access with each other. This is important since we utilize lazy enabled peer access (`cudaIpcMemLazyEnablePeerAccess`).

include/ofi_hmem.h
darrylabbate: I opted not to change all the other type signatures since I like to preserve the line-by-line Git blame as much as possible.

src/hmem_cuda.c
```diff
@@ -582,12 +485,49 @@ static int cuda_hmem_verify_devices(void)
...
+				ofi_cudaGetErrorName(cuda_ret),
+				ofi_cudaGetErrorString(cuda_ret));
+			return -FI_EIO;
+		}
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Peer access from CUDA device %d -> CUDA device %d : %s\n",
+			dev, peer, can_access_peer ? "Yes" : "No");
+		if (!can_access_peer)
+			return FI_SUCCESS;
+	}
```
darrylabbate, line 525: Went with a simpler non-exhaustive linear algo for now to appease some initialization overhead concerns, but I'm not entirely sure this is the right approach. NVIDIA's [`deviceQuery`](https://github.com/zchee/cuda-sample/blob/master/1_Utilities/deviceQuery/deviceQuery.cpp) reports the peer access support for each permutation, which somewhat indicates it may be possible for peer access to be uni-directional between 2 devices? I've found no evidence online that this is the case, though there's also no evidence that any meaningful amount of overhead from the exhaustive search is incurred.

darrylabbate: > it may be possible for peer access to be uni-directional between 2 devices?

Follow-up: peer access itself (enabled via `cudaDeviceEnablePeerAccess()` or otherwise) is unidirectional, but I don't think there's any reason something like `cudaDeviceCanAccessPeer()` would evaluate differently based on the order of `{src,dst}Device`.

src/hmem_cuda.c
shijin-aws: your new revision is actually `++dev, ++peer`?

src/hmem_cuda.c
```diff
@@ -482,6 +494,40 @@ static int cuda_hmem_verify_devices(void)
...
+				"cuDeviceCanAccessPeer(): %s:%s\n",
+				ofi_cudaGetErrorName(cuda_ret),
+				ofi_cudaGetErrorString(cuda_ret));
+			return -FI_EIO;
+		}
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Peer access from CUDA device %d -> CUDA device %d : %s\n",
+			dev, peer, can_access_peer ? "Yes" : "No");
+		if (!can_access_peer)
+			return FI_SUCCESS;
```
shijin-aws, line 524: So if can_access_peer is false, we still return FI_SUCCESS? what is the expected return code for this function. Can you clarify it in function's docstring?

darrylabbate: Ack, added a Doxygen comment for this function

src/hmem_cuda.c
shijin-aws: Fair enough.

aingerson: I'm not familiar with the IPC limitations of CUDA devices but is it possible that 0->1 is ok, 1->2 is ok, but 0->2 is not?

aingerson (commented):
Overall looks good to me, a couple of random things and a clarifying question so I can understand the CUDA application more. Thanks!

src/hmem_cuda.c
darrylabbate: > I'm not familiar with the IPC limitations of CUDA devices but is it possible that 0->1 is ok, 1->2 is ok, but 0->2 is not?

I don't believe so. In this scenario, if 0 cannot access 2, that means 0 cannot access 1 and/or 1 cannot access 2. This would indicate some kind of mixed topology/configuration. It would be recommended to restrict the devices visible to Libfabric via `CUDA_VISIBLE_DEVICES` to only "similar" devices.

On that note, this makes me realize we definitely shouldn't employ the [(0,1),(2,3),...] check since that wouldn't catch potential "dissimilar" devices e.g. (1,2). So I think the simpler linear check is perfectly suitable here.

shijin-aws (approved):

src/hmem_cuda.c
```diff
@@ -396,13 +407,14 @@ static cudaError_t cuda_disabled_cudaMemcpy(void *dst, const void *src,
...
-			goto err_dlclose_cuda_driver; \
-		} \
+#define CUDA_FUNCS_DLOPEN(type, sym)					\
+	do {								\
+		cuda_ops.sym = dlsym(cuda_attr.type##_handle, #sym);	\
+		if (!cuda_ops.sym) {                                    \
+			FI_WARN(&core_prov, FI_LOG_CORE,		\
+				"Failed to find " #sym "\n");		\
+			goto err_dlclose_cuda_driver;			\
+		}							\
```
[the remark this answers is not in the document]
darrylabbate: Good catch - I realigned `\` on the wrong commit

src/hmem_cuda.c
```diff
@@ -473,6 +473,9 @@ static int cuda_hmem_verify_devices(void)
 		return -FI_EIO;
 	}
 
+	FI_INFO(&core_prov, FI_LOG_CORE, "Number of CUDA devices detected: %d\n",
+			cuda_attr.device_count);
```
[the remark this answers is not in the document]
darrylabbate: Fixed, thanks

prov/efa/src/efa_user_info.c
```diff
@@ -302,7 +302,7 @@ bool efa_user_info_should_support_hmem(int version)
 	if (hmem_ops[FI_HMEM_CUDA].initialized && FI_VERSION_GE(version, FI_VERSION(1, 18))) {
 		EFA_INFO(FI_LOG_CORE,
 			"User is using API version >= 1.18. CUDA library and "
-			"device are available, claim support of FI_HMEM.");
+			"device are available, claim support of FI_HMEM.\n");
```
lrbison, line 305: minor, but since you have a whole commit for this change: s/device are/devices are/

include/ofi_hmem.h
```diff
@@ -62,6 +62,8 @@ CUresult ofi_cuGetErrorName(CUresult error, const char** pStr);
 CUresult ofi_cuGetErrorString(CUresult error, const char** pStr);
 CUresult ofi_cuPointerGetAttribute(void *data, CUpointer_attribute attribute,
 				   CUdeviceptr ptr);
+CUresult ofi_cuDeviceCanAccessPeer(int *canAccessPeer, CUdevice srcDevice,
+		CUdevice dstDevice);
```
lrbison, line 66: alignment

src/hmem_cuda.c
```diff
@@ -108,6 +111,8 @@ static struct {
 					  CUdeviceptr ptr);
 	CUresult (*cuMemGetAddressRange)( CUdeviceptr* pbase,
 					  size_t* psize, CUdeviceptr dptr);
+	CUresult (*cuDeviceCanAccessPeer)(int *canAccessPeer,
+			CUdevice srcDevice, CUdevice dstDevice);
```
lrbison, line 115: alignment

src/hmem_cuda.c
```diff
@@ -222,6 +227,12 @@ CUresult ofi_cuMemGetAddressRange(CUdeviceptr* pbase, size_t* psize, CUdeviceptr
 	return cuda_ops.cuMemGetAddressRange(pbase, psize, dptr);
 }
 
+CUresult ofi_cuDeviceCanAccessPeer(int *canAccessPeer, CUdevice srcDevice,
+		CUdevice dstDevice)
```
lrbison, line 231: alignment

src/hmem_cuda.c
lrbison: I think I would have pinned dev=0 and perform (0-1, 0-2, 0-3 ...), seems clearer to me.  My 2c.

src/hmem_cuda.c
lrbison: Could the doc string mention that p2p_access_supported  will be set on FI_SUCCESS?

lrbison (approved):

prov/efa/src/efa_user_info.c
darrylabbate: ack

src/hmem_cuda.c
darrylabbate: > Could the doc string mention that p2p_access_supported will be set on FI_SUCCESS?

That's only true in the case where the `for` loop finishes (i.e. p2p access available across all visible devices), but I can document this.

shijin-aws (approved):

shefty (approved):
I'm fine with the changes.  It's an improvement from where we are.

shefty:
Intel CI failure is unrelated to this PR.  failure: fi_rdm_tagged_peek -p "sockets"

lrbison (approved):

PR #8779, opened 2023-04-10
prov/util: update log level for invalid AV type message

sunkuamzn:
Invalid AV type should be a warning

### DIFF
--- a/prov/util/src/util_av.c
+++ b/prov/util/src/util_av.c
@@ -502,12 +502,18 @@ static int util_verify_av_attr(struct util_domain *domain,
 	case FI_AV_TABLE:
 		if ((domain->av_type != FI_AV_UNSPEC) &&
 		    (attr->type != domain->av_type)) {
-			FI_INFO(domain->prov, FI_LOG_AV, "Invalid AV type\n");
-		   	return -FI_EINVAL;
+				char buf1[20], buf2[20];
+				FI_WARN(domain->prov, FI_LOG_AV,
+							"Invalid AV type. "
+							"domain->av_type: %s "
+							"attr->type: %s\n",
+						fi_tostr_r(&buf1, 20, &domain->av_type, FI_TYPE_AV_TYPE),
+						fi_tostr_r(&buf2, 20, &attr->type, FI_TYPE_AV_TYPE));
+				return -FI_EINVAL;
 		}
 		break;
 	default:
-		FI_WARN(domain->prov, FI_LOG_AV, "invalid av type\n");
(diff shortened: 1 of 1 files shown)


prov/util/src/util_av.c
```diff
@@ -502,12 +502,12 @@ static int util_verify_av_attr(struct util_domain *domain,
 	case FI_AV_TABLE:
 		if ((domain->av_type != FI_AV_UNSPEC) &&
 		    (attr->type != domain->av_type)) {
-			FI_INFO(domain->prov, FI_LOG_AV, "Invalid AV type\n");
+			FI_WARN(domain->prov, FI_LOG_AV, "Invalid AV type\n");
```
wzamazon, line 505: Since you are here. I suggest to print `domain->av_type` and `attr->type`

shefty:
patch needs sign-off-by line

shefty:
@sunkuamzn - changes are okay, but I can't merge this unless there's a proper signed-off-by line in the commit message.

prov/util/src/util_av.c
sunkuamzn: done

sunkuamzn:
@shefty sorry for the delay. fixed.

prov/util/src/util_av.c
```diff
@@ -502,12 +502,14 @@ static int util_verify_av_attr(struct util_domain *domain,
 	case FI_AV_TABLE:
 		if ((domain->av_type != FI_AV_UNSPEC) &&
 		    (attr->type != domain->av_type)) {
-			FI_INFO(domain->prov, FI_LOG_AV, "Invalid AV type\n");
+			FI_WARN(domain->prov, FI_LOG_AV, "Invalid AV type. domain->av_type: %s attr->type: %s\n",
```
shefty, line 505: line length is too long.  Please limit to about 80 characters.

prov/util/src/util_av.c
```diff
@@ -502,12 +502,14 @@ static int util_verify_av_attr(struct util_domain *domain,
...
+					fi_tostr(&domain->av_type, FI_TYPE_AV_TYPE),
+					fi_tostr(&attr->type, FI_TYPE_AV_TYPE));
```
shefty, line 507: fi_tostr() uses an internal buffer that gets overwritten on subsequent calls.  I'm not sure how this ends up being handled, but using fi_tostr_r() would be safer.

shefty:
I pulled these changes in manually with minor updates and pushed to PR #8966.  I updated the patch to fix the indentation and use sizeof(str) instead of hard-coding the size.  Thanks!

PR #8781, opened 2023-04-11
core,prov/sm2: Initial Implementation of Nemesis Style SHM

a-szegel:
SM2 is an experimental provider which implements a lockless FIFO queue, which follows the Nemesis algorithm and is implemented similarly to OMPI.  The goal of the SM2 provider is to help identify the source of the extra latency in libfabric SHM compared to more optimal implementations.  In order to make it easier to identify the latency gaps, I decided it would be best to strip the provider of everything but the inject protocol.  This allows us to move quickly with patches to the SM2 code, and focus our effort on the critical path (including the layers above us). Once the root cause of the performance issue is identified, we will ensure that the entire community is able to benefit from this work.

 At first, it was envisioned that we would be able to easily swap out the queue from SHM with a Nemesis based queue.  Complexities in the implementation of the Nemesis algorithm have caused SHM and SM2 to diverge.  SHM and SM2 are almost fundamentally different providers.  The major differences between SM2 and SHM include but are not limited to:
 
* Single Shared Memory File for all processes within a users account
  * New SHM File Structure (File Header, "Shared AV", and region definition)
  * The single shared memory file causes SM2 to be stateful between runs which leads to complex initialization and cleanup edge cases
* New data transport algorithm
  * New lockless FIFO queue
  * Sender now owns the buffers that they write to their peers (natural flow control without starvation)
  * New memory barriers for correctness which are required on modern processors
* No more response protocol b/c there is a natural response path path when we return buffers

Some things that are similar between SHM/SM2:
* Completions
* Libfabric constructs (fabric, domain, msg, ep)

**This PR is a step forward in helping to identify the latency gap, but is by no means the end of the road for this effort.** AWS needs a stable place where we can work from so that we can put this code in our CI, and have multiple developers working together to solve these issues (without continuously changing history).  Once we have identified the root cause in the performance gap, we will work with the community to make sure it gets fixed for everyone.


wzamazon:
Ubuntu 20.04 and 22.04 compilation failure need attention.

wzamazon (approved):

prov/sm2/src/sm2_ep.c
```diff
@@ -91,379 +89,215 @@ int sm2_getname(fid_t fid, void *addr, size_t *addrlen)
 	*addrlen = strlen(ep->name) + 1;
 
 	if (!ret)
-		((char *) addr)[*addrlen - 1] = '\0';
+		((char *)addr)[*addrlen - 1] = '\0';
+
+	*addrlen = SM2_NAME_MAX;
```
a-szegel, line 94: This is worthy of attention.  This fixed a startup race condition in OMPI + SM2.  OMPI takes the addrlen returned by fi_getname and creates a char[num_ranks][addrlen], and memcpys in addrlen into the char* (assuming the string fits, and is null terminated).  SM2 coordination was using strlen() to determine where one string ended, and the next string began.  By setting the size to a constant, both work well.  This approach seems to also align with the docs (but there might also be a bug in OMPI):


fi_getname / fi_getpeer
The fi_getname and fi_getpeer calls may be used to retrieve the local or peer endpoint address, respectively. On input, the addrlen parameter should indicate the size of the addr buffer. If the actual address is larger than what can fit into the buffer, it will be truncated and -FI_ETOOSMALL will be returned. **On output, addrlen is set to the size of the buffer needed to store the address, which may be larger than the input value.**

fi_getname is not guaranteed to return a valid source address until after the specified endpoint has been enabled or has had an address assigned. An endpoint may be enabled explicitly through fi_enable, or implicitly, such as through fi_connect or fi_listen. An address may be assigned using fi_setname. fi_getpeer is not guaranteed to return a valid peer address until an endpoint has been completely connected – an FI_CONNECTED event has been generated.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+		for (jentry = 0; jentry < header->ep_enumerations_max; jentry++) {
+			entries[jentry].pid = 0;
+			entries[jentry].ep_name[0] = '\0';
+			entries[jentry].startup_ready = 0;
+		}
+
+		/* Shrink File */
+		sm2_mmap_shrink_to_size(map, header->ep_regions_offset);
+	}
+}
```
lrbison, line 675: shoot, looks like clang-format doesn't have an easy way to add lines at the end of files.  At least I couldn't find one when I looked.

lrbison (approved):

a-szegel:
Now SM2 correctly support FI_ADDR_NOTAVAIL

aingerson:
I'm still in the middle of reviewing this but figured I'd mention this now in case it's a difficult fix - for the clang format, our width is 80 characters, not 90.
And I know this would be a lot to fix but I think it would make more sense to either do the clang application last OR do the removal of code first. The sm2 format patch applies the format to a lot of code that just ends up getting removed.

a-szegel:
The reason I do the clang-format first is that I can continously rebase my history without having a conflict every single time.  If I want to reapply clang-format to every single patch, I would manually do the following
```
git rebase -i clang-format-patch
# select every commit for editing
# Update clang format file
For Each commit
    # Copy off commit message
    git reset HEAD~1
    git add .
    git clang-format
    git add .
    git commit -s "Original Commit Message"
```

If the clang-format patch was at the end, every time you ask me to rebase the history I would have to deal with conflicts. That resolution might be as simple as deleting and re-applying the clang-format patch (but it still makes things annoying that they don't rebase cleanly). 

I would prefer to keep the clang-format fix first if that is ok with you too.  I have no problems updating the clang-formating across every commit, but I would like to do it a minimal amount of times, so if we could gather all of our clang-format comments together, I will go ahead and do that!

Thanks :)

a-szegel:
On second thought... I think I have committed us to doing clang-format this way.  If I move the formatting patch to the end, I suspect it will make all my intermediary patches look messy.  The second reason why I chose to do it at the beginning is so you could have clean formatting when you do a commit by commit review instead of relying on all the formatting being fixed at the end.

shefty:
The general rule for submission of a patch series is that one of the patches later in the series shouldn't change the same code that an earlier patch changed.  If the change is needed, then the change should be in the earlier patch!  There are some exceptions to this, but this is what keeps the changes bisectable.

a-szegel:
From a functionality perspective, this change set is still bisectable.  The exception I am seeking is a big formatting change upfront for readability, and clean patches throughout the entire change set.  Doing the formatting change first actually makes the git history more useful b/c when the developer does a git blame, they will see the commit ID that changes the functionality and not the formatting.

Also everything compiles along the way: A, AB, ABC, ABCD ..... (but ordering is required).

shefty:
Your first formatting patch has line lengths that are too long.  Adjust to 80 characters.

a-szegel:
Yes, I will do that.  I want to bulk fix the formatting changes so I am waiting for Alexia/You to come back to me with all the edits you want before I fix them.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+
+	for (jentry = 0; jentry < header->ep_enumerations_max; jentry++) {
+		if (0 == strncmp(name, entries[jentry].ep_name, SM2_NAME_MAX)) {
+			FI_WARN(&sm2_prov, FI_LOG_AV, "Found existing %s in slot %d\n",
+				name, jentry);
+			return jentry;
+		}
+		if (entries[jentry].ep_name[0] != '\0') {
+			FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+				"Searching for %s. Not yet found in spot AV[%d]=%s, PID= "
```
aingerson, line 517: This mapping has nothing to do with the AV. Same below

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+{
+
+	struct sm2_coord_file_header *header = (void *)map->base;
+	struct sm2_ep_allocation_entry *entries;
+	int jentry;
+
+	entries = sm2_mmap_entries(map);
+
+	for (jentry = 0; jentry < header->ep_enumerations_max; jentry++) {
+		if (0 == strncmp(name, entries[jentry].ep_name, SM2_NAME_MAX)) {
```
aingerson, line 510: Doing a O(n) walk through to see if the name matches is not very efficient. I recommend replacing this lookup with a rbmap like the current smr_map uses.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+ *
+ * @param[in] name
+ * @param[inout] map
+ * @param[out] av_key
+ */
+ssize_t sm2_coordinator_free_entry(struct sm2_mmap *map, int av_key)
+{
+	struct sm2_ep_allocation_entry *entries;
+	int pid = getpid();
+	(void)pid;
```
aingerson, line 542: does this do anything?

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+}
+
+/**
+ * @brief clear the pid for this entry.  must already hold lock.
+ *
+ * @param[in] name
+ * @param[inout] map
+ * @param[out] av_key
+ */
+ssize_t sm2_coordinator_free_entry(struct sm2_mmap *map, int av_key)
```
aingerson, line 538: Rename parameter to shm_id or something similar. This isn't related to the AV but rather an internal way for shms to look each other up.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+			"We should (%d) hold the lock, but pid_lock_hint is %d!", pid,
+			hint_pid);
+	}
+	pthread_mutex_unlock(&header->write_lock);
+
+	/* use CAS to clear pid because it's possible another process has
+	   already acquired the lock.  CAS will only clear it if we hold it. */
+	ofi_atomic_cas_bool32(&header->pid_lock_hint, pid, 0);
+	return 0;
+}
```
aingerson, line 575: I would be interested in knowing the overhead of having all of this hint code inside the locking/unlocking path. I see that it is needed for clearing up the mapping if the peer has died while holding the lock but I think this has the potential to introduce more overhead than it's worth. We have signal handlers to clean up a peer if it does die. I'm wondering if we can use something there to handle the case that this protects against?

aingerson (changes requested):
I'm only done reviewing the first 6 patches but wanted to submit them anyway so you have time to address them. There are A LOT of changes that need to happen in the coordinator change. The files themselves are super messy and the formatting is all off (I don't think the clang file was run on it). I've left a lot of comments on various things to change through as well as most of the logic but I'll do another pass once the patch is in better shape. I will continue to review the reset of the patches as well.

Also, I'd like to see some performance numbers and improvement for these patches. If these changes aren't helping a lot of the performance issues we are seeing, then I don't think there is much value in merging them. We need to know we are on the right path. Otherwise, we are replacing one problematic shm implementation with another one.

prov/sm2/src/sm2.h
```diff
@@ -218,7 +219,7 @@ struct sm2_domain {
 	struct fid_peer_srx *srx;
 };
 
-#define SM2_PREFIX "fi_shm://"
+#define SM2_PREFIX "fi_sm2://"
```
aingerson, line 170: Not related to nemesis I don't think. Move to a separate patch that addresses all lingering shms that got missed during the duplication

prov/sm2/src/sm2.h
```diff
@@ -267,6 +268,8 @@ struct sm2_srx_ctx {
 	struct sm2_queue unexp_msg_queue;
 	struct sm2_queue unexp_tagged_queue;
 	struct sm2_recv_fs *recv_fs;
+
+	// TODO Determine if this spin lock is needed.
```
aingerson: I understand the addition of comments but this one isn't related to nemesis and adds unnecessary noise to this large patch.

prov/sm2/src/sm2.h
```diff
@@ -279,13 +282,11 @@ struct sm2_ep {
 	size_t rx_size;
 	const char *name;
 	uint64_t msg_id;
-	struct sm2_region *volatile region;
-	// if double locking is needed, shm region lock must
-	// be acquired before any shm EP locks
+	int self_fiaddr;
+	struct sm2_mmap *mmap_regions;
```
aingerson, line 189: It is easy to get from the ep to the map since it is part of the sm2_av (util_ep->av). No need to double this pointer.

prov/sm2/src/sm2.h
```diff
@@ -279,13 +282,11 @@ struct sm2_ep {
 	size_t rx_size;
 	const char *name;
 	uint64_t msg_id;
-	struct sm2_region *volatile region;
-	// if double locking is needed, shm region lock must
-	// be acquired before any shm EP locks
+	int self_fiaddr;
```
aingerson, line 188: The global ids are not the same as the fiaddrs. Rename to something that indicates this is part of the internal naming index, not the fiaddr. In shm we use "shm_id" which you can use as well unless you have something else in mind

prov/sm2/src/sm2.h
```diff
@@ -296,6 +297,11 @@ struct sm2_ep {
 	void *dsa_context;
 };
 
+static inline struct sm2_region *sm2_smr_region(struct sm2_ep *ep, int id)
+{
+	return sm2_mmap_ep_region(ep->mmap_regions, id);
+}
+
```
aingerson: An inline function that calls one thing that is almost identical?

prov/sm2/src/sm2.h
```diff
@@ -328,29 +334,15 @@ int sm2_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
...
-			  int64_t id, struct sm2_resp *resp);
-void sm2_generic_format(struct sm2_cmd *cmd, int64_t peer_id, uint32_t op, uint64_t tag,
-			uint64_t data, uint64_t op_flags);
-size_t sm2_copy_to_sar(struct smr_freestack *sar_pool, struct sm2_resp *resp,
-		       struct sm2_cmd *cmd, enum fi_hmem_iface, uint64_t device,
-		       const struct iovec *iov, size_t count, size_t *bytes_done,
-		       int *next);
-size_t sm2_copy_from_sar(struct smr_freestack *sar_pool, struct sm2_resp *resp,
-			 struct sm2_cmd *cmd, enum fi_hmem_iface iface, uint64_t device,
-			 const struct iovec *iov, size_t count, size_t *bytes_done,
```
aingerson: There's a lot in this patch (like the above) that should be part of the removal patch, not this one.

prov/sm2/src/sm2.h
```diff
@@ -328,29 +334,15 @@ int sm2_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
...
-			 int *next);
-
-int sm2_select_proto(bool use_ipc, bool cma_avail, enum fi_hmem_iface iface, uint32_t op,
-		     uint64_t total_len, uint64_t op_flags);
+void sm2_generic_format(struct sm2_free_queue_entry *cmd, int64_t peer_id, uint32_t op,
+			uint64_t tag, uint64_t data, uint64_t op_flags);
+int sm2_select_proto();
```
aingerson, line 231: I'm guessing this does nothing now - you can just remove it completely now that we're not assuming parallel development with shm

prov/sm2/src/sm2_ep.c
```diff
@@ -274,86 +246,59 @@ static void sm2_init_queue(struct sm2_queue *queue, dlist_func_t *match_func)
...
+	assert(total_len <= 4096);
 
-	sm2_generic_format(cmd, peer_id, op, tag, data, op_flags);
-	sm2_format_inject(cmd, iface, device, iov, iov_count, peer_smr, tx_buf);
+	self_region = sm2_smr_region(ep, ep->self_fiaddr);
 
-	ofi_cirque_commit(sm2_cmd_queue(peer_smr));
-	peer_smr->cmd_cnt--;
+	if (smr_freestack_isempty(sm2_free_stack(self_region))) {
+		sm2_progress_recv(ep);
```
aingerson, line 212: Is there a significant performance improvement because of this change? If so, can you show it, please?

prov/sm2/src/sm2_attr.c
```diff
@@ -35,8 +36,8 @@
 #define SM2_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED)
 #define SM2_RX_CAPS                                                                      \
 (FI_SOURCE | OFI_RX_MSG_CAPS | FI_TAGGED | FI_DIRECTED_RECV | FI_MULTI_RECV)
-#define SM2_HMEM_TX_CAPS ((SM2_TX_CAPS))
-#define SM2_HMEM_RX_CAPS ((SM2_RX_CAPS))
+#define SM2_HMEM_TX_CAPS (0)
+#define SM2_HMEM_RX_CAPS (0)
```
aingerson, line 40: This shouldn't be in this patch and if you're going to set it to 0, why not just remove it?

prov/sm2/src/sm2_attr.c
```diff
@@ -126,8 +127,8 @@ struct fi_domain_attr sm2_hmem_domain_attr = {
 struct fi_fabric_attr sm2_fabric_attr = {.name = "sm2",
 					 .prov_version = OFI_VERSION_DEF_PROV};
 
-struct fi_info sm2_hmem_info = {.caps =
-				    SM2_HMEM_TX_CAPS | SM2_HMEM_RX_CAPS | FI_MULTI_RECV,
+struct fi_info sm2_hmem_info = {.caps = SM2_HMEM_TX_CAPS | SM2_HMEM_RX_CAPS |
```
aingerson, line 130: Clang did not play well with this....

prov/sm2/src/sm2_attr.c
```diff
@@ -136,7 +137,7 @@ struct fi_info sm2_hmem_info = {.caps =
 				.fabric_attr = &sm2_fabric_attr};
 
 struct fi_info sm2_info = {
-    .caps = SM2_TX_CAPS | SM2_RX_CAPS | FI_MULTI_RECV,
+    .caps = SM2_TX_CAPS | SM2_RX_CAPS | FI_MULTI_RECV | FI_LOCAL_COMM,
```
aingerson, line 140: Separate fix

prov/sm2/src/sm2_av.c
```diff
@@ -45,155 +46,104 @@ static int sm2_av_close(struct fid *fid)
...
-			if (!ret) {
-				ofi_mutex_lock(&util_av->lock);
-				ret = ofi_av_insert_addr(util_av, &shm_id, &util_addr);
-				ofi_mutex_unlock(&util_av->lock);
-			}
+	/* acquire the lock */
+	sm2_coordinator_lock(&sm2_av->sm2_mmap);
+
+	for (i = 0; i < count; i++, addr = (char *)addr + SM2_NAME_MAX) {
+		FI_WARN(&sm2_prov, FI_LOG_AV,
```
aingerson, line 76: This shouldn't not be a warning (same below). If you want to log it because it is helpful you can use FI_DBG for debug logs but this doesn't seem terribly useful to me. We already have an fi insert log with FI_INFO

prov/sm2/src/sm2_av.c
```diff
@@ -45,155 +46,104 @@ static int sm2_av_close(struct fid *fid)
...
+			(char *)addr, getpid(), count);
+		ret = sm2_coordinator_allocate_entry(addr, &sm2_av->sm2_mmap, &util_addr,
+						     false);
+		FI_WARN(&sm2_prov, FI_LOG_AV,
+			"fi_av_insert(): finished sm2_coordinator_allocate_entry() "
+			"resulting AV Found = %d \n",
+			util_addr);
+		if (ret && util_av->eq) {
+			ofi_av_write_event(util_av, i, -ret, context);
 		}
```
aingerson, line 88: Still some lingering format issues to figure out with clang (no {} around single line ifs)

prov/sm2/src/sm2_av.c
```diff
@@ -45,155 +46,104 @@ static int sm2_av_close(struct fid *fid)
...
-			if (shm_id >= 0)
-				sm2_map_del(sm2_av->sm2_map, shm_id);
+		if (ret)
 			continue;
+
+		if (flags & FI_AV_USER_ID) {
+			sm2_av->sm2_aux[util_addr].cqfid = fi_addr[i];
+			FI_INFO(&sm2_prov, FI_LOG_AV,
+				"fi_addr: %d, USER_ID: %" PRIu64 "\n", util_addr,
+				fi_addr[i]);
```
aingerson, line 96: Drop this and just add FI_AV_USER_ID support to the util_av

prov/sm2/src/sm2_av.c
```diff
@@ -45,155 +46,104 @@ static int sm2_av_close(struct fid *fid)
...
-			else
-				sm2_ep->region->max_sar_buf_per_peer = SM2_BUF_BATCH_MAX;
-		}
-		sm2_av->used--;
-	}
-
-	ofi_mutex_unlock(&util_av->lock);
-	return ret;
+	/* No-OP */
+	return 0;
```
aingerson, line 122: This should not be a no-op. We need to remove it from the util_av (ie the private_aux for what you're doing right now).

prov/sm2/src/sm2_av.c
```diff
@@ -45,155 +46,104 @@ static int sm2_av_close(struct fid *fid)
...
 	util_av = container_of(av, struct util_av, av_fid);
 	sm2_av = container_of(util_av, struct sm2_av, util_av);
 
-	id = sm2_addr_lookup(util_av, fi_addr);
-	name = sm2_av->sm2_map->peers[id].peer.name;
+	header = (void *)sm2_av->sm2_mmap.base;
+	entries = (void *)(sm2_av->sm2_mmap.base + header->ep_enumerations_offset);
 
-	strncpy((char *)addr, name, *addrlen);
+	strncpy(addr, entries[fi_addr].ep_name, *addrlen - 1);
```
aingerson, line 143: This looks wrong. You're looking up the peer using the fiaddr which is not the same as the global shm mapping index. You'll need to look up the shm_id using the util_av (like the old code did) and then use that to access the mmap and get the name.

prov/sm2/src/sm2_av.c
```diff
@@ -245,15 +207,20 @@ int sm2_av_open(struct fid_domain *domain, struct fi_av_attr *attr, struct fid_a
 	}
 
 	if (attr->type == FI_AV_UNSPEC)
-		attr->type = FI_AV_TABLE;
+		attr->type = FI_AV_MAP;
+
+	if (attr->type != FI_AV_MAP) {
+		FI_WARN(&sm2_prov, FI_LOG_AV, "Only FI_AV_MAP Supported\n");
+		return -FI_EINVAL;
+	}
```
aingerson, line 215: No applications really support FI_AV_MAP and it's really restrictive for the application. FI_AV_TABLE is much more common and usable with applications and if you use the util_av to store the fi_addr you will easily support it with no performance loss.

prov/sm2/src/sm2_common.h
```diff
@@ -53,17 +55,29 @@ extern "C" {
...
 #define SM2_FLAG_ATOMIC (1 << 0)
 #define SM2_FLAG_DEBUG (1 << 1)
 #define SM2_FLAG_IPC_SOCK (1 << 2)
 #define SM2_FLAG_HMEM_ENABLED (1 << 3)
+#define SM2_CMD_SIZE		256	/* align with 64-byte cache line */
+// reserves 0-255 for defined ops and room for new ops
+// 256 and beyond reserved for ctrl ops
+#define SM2_REMOTE_CQ_DATA (1 << 0)
+#define SM2_TX_COMPLETION (1 << 2)
+#define SM2_RX_COMPLETION (1 << 3)
```
aingerson: A lot of movement of code going on in this patch. It would be much better for this to solely contain the algorithmic changes and for any moving of things to happen before this.

prov/sm2/src/sm2_common.h
```diff
@@ -86,45 +100,75 @@ enum {
...
+/*
+ * Unique smr_op_hdr for smr message protocol:
+ * 	addr - local shm_id of peer sending msg (for shm lookup)
+ * 	op - type of op (ex. ofi_op_msg, defined in ofi_proto.h)
+ * 	op_src - msg src (ex. smr_src_inline, defined above)
+ * 	op_flags - operation flags (ex. SMR_REMOTE_CQ_DATA, defined above)
+ * 	src_data - src of additional op data (inject offset / resp offset)
+ * 	data - remote CQ data
+ */
+struct sm2_msg_hdr {
```
aingerson: What's the difference between the sm2_protocol_hdr and the sm2_msg_hdr? Do you even use this? Is this something that was moved only to be removed in the next patch? And why isn't the formatting matching the clang format specification?

prov/sm2/src/sm2_ep.c
aingerson: I agree with Luke. This seems like an Open MPI issue. We found the same issue in OneCCL and had to fix it in there.

prov/sm2/src/sm2.h
a-szegel: The reason these were removed in the "Change to Nemesis patch" is they no longer compile after the change to Nemesis.  I think moving the removal patch to before the Nemesis patch would help.  I can't remember why I put it after, but I was pretty sure I did it for a reason... time to re-discover it :).

prov/sm2/src/sm2.h
a-szegel: I kept it in there b/c I have more patches lined up to add additional features.  I can remove it for now.

prov/sm2/src/sm2_ep.c
a-szegel: The sender sends its own FQE's to the receiver.  The receiver processes the message and sends them back to the senders recv queue (dual purpose: receiving messages, and returning spent FQE's).  If the Sender only calls Send, and never calls progress, we will run out of FQE's to send.  This change ensures that if we are out of FQE's, we progress our receive to try and get more.  We then check again to see if we were able to get more, and if we weren't, we return -FI_EAGAIN, b/c the sender needs to wait until the other receivers progress and return its FQE's in order to send more (flow control).

a-szegel: TLDR: This is a correctness change, not a performance change.

prov/sm2/src/sm2_attr.c
a-szegel: Yea, I noticed that too... may need to investigate setting to figure out how to deal with structures better.

prov/sm2/configure.m4
```diff
@@ -81,11 +83,16 @@ AC_DEFUN([FI_SM2_CONFIGURE],[
 	      AC_DEFINE_UNQUOTED([SM2_HAVE_DSA],[$dsa_happy],
 				 [Whether DSA support is available])
 
+		  AC_CHECK_DECL([HAVE_ATOMICS], [atomics_happy=1], [])
+		  AS_IF([test $atomics_happy -eq 0],
+		  		[AC_CHECK_DECL([HAVE_BUILTIN_MM_ATOMICS], [atomics_happy=1], [])], [])
```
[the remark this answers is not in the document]
a-szegel: clang is only doing .c/.h files

prov/sm2/src/sm2_common.h
```diff
@@ -339,6 +339,17 @@ static inline void sm2_signal(struct sm2_region *smr)
...
+{
+	/* STUB Will implement on later commit when sm2_region changes */
+	return NULL;
+}
+static inline struct smr_freestack *sm2_free_stack(struct sm2_region *smr)
+{
+	/* STUB, Will implement on later commit when sm2_region changes */
+	return NULL;
+}
+
```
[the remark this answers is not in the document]
a-szegel: I can't b/c the sm2_coordination depends on them, and they depend on the updated sm2_region structure.  This was how I was able to split that dependency to make cleaner patches.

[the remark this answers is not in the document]
a-szegel: It is actually the only thing the SM2 coordination depends on... so I figured it would be better to stub, than to merge the two patches.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	if (self) {
+		entries[jentry].startup_ready = 0;
+		atomic_wmb();
+		entries[jentry].pid = pid;
+	}
+
+	if (!self && entries[jentry].pid == 0) {
+		entries[jentry].startup_ready = 0;
+		atomic_wmb();
+		entries[jentry].pid = -pid;
```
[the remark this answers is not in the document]
a-szegel: The negative PID means that a processes reserved a spot in `struct sm2_ep_allocation_entry`  for a process with an ADDR that isn't up yet.  This happens because on fi_av_insert(), we need to immediately return the SHM_ID (the requirement may lossen if we use util_av).  

Example:
Luke comes alive (takes spot 1)
Luke fi_av_inserts() Alexia's address, pre-allocates a spot for her, and returns spot 1
seth comes allve (takes spot 2)
Seth fi_av_inserts() Alexia's address, finds Alexia's spot and returns spot 1
Alexia comes alive and realizes that spot 1 has been, pre-allocated for her, fills it out, and marks her region initialized, and ready to receive messages

prov/sm2/src/sm2_coordination.c
a-szegel: This is startup, and not used on the critical path.

a-szegel: On the critical path, we have:
```
sm2_mmap_remap() // on very rare occasions after late arrivers join
```

prov/sm2/src/sm2_coordination.c
a-szegel: this is also startup only, and the lock is only used for peers that wish to modify the SHM file header (i.e.. insert themselves into header)... not saying it shouldn't be fixed, but it is lower priority.  I think we care more about startup correctness than performance.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+ * @param[in] fd
+ * @param[out] map
+ */
+void *sm2_mmap_map(int fd, struct sm2_mmap *map)
+{
+	struct stat st;
+	int err;
+	char *stage;
+
+	stage = "fstat of sm2_mmaps file";
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	struct stat st;
+	int err;
+	char *stage;
+
+	/* return quickly if no need to check the file. */
+	assert(at_least > 0);
+	if (map->size >= at_least)
+		return map->base;
+
+	stage = "fstat of sm2_mmaps file";
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	if (OFI_UNLIKELY(err))
+		goto out;
+
+	if (st.st_size < at_least) {
+		/* we need to grow the file. */
+		stage = "ftruncate of sm2_mmaps file";
+		err = ftruncate(map->fd, at_least);
+		if (OFI_UNLIKELY(err))
+			goto out;
+	}
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+		stage = "funmap of sm2_mmaps file";
+		err = munmap(map->base, map->size);
+		if (OFI_UNLIKELY(err))
+			goto out;
+
+		stage = "fmap of sm2_mmaps file";
+		map->base =
+		    mmap(0, at_least, PROT_READ | PROT_WRITE, MAP_SHARED, map->fd, 0);
+		if (OFI_UNLIKELY(map->base == MAP_FAILED)) {
+			perror("failed to remap when increasing the map size in "
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+		}
+		map->size = at_least;
+	}
+
+out:
+	if (OFI_UNLIKELY(err)) {
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+			"Failed syscall during sm2_mmap_remap()\n");
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "%s : %s\n", stage, strerror(errno));
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "at_least: %" PRIu64 "\n", at_least);
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+				struct sm2_region *peer_region =
+				    sm2_mmap_ep_region(map, jentry);
+
+				if (entries[jentry].startup_ready &&
+				    smr_freestack_isfull(sm2_free_stack(peer_region))) {
+					/* we found a slot with a dead PID and the
+					 * freestack is full */
+					entries[jentry].pid = 0;
+					goto found;
+				};
```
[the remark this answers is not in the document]
a-szegel: removed

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	}
+
+	fprintf(stderr,
+		"No available entries were found in the coordination file, all %d were "
+		"used\n",
+		header->ep_enumerations_max);
+	return -FI_EAVAIL;
+
+found:
+
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
a-szegel: Done. It was stopping a compile warning when compiling without asserts, unused variable.  Got ride of variable.

prov/sm2/src/sm2_attr.c
```diff
@@ -32,129 +32,116 @@
 
 #include "sm2.h"
 
-#define SM2_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED )
-#define SM2_RX_CAPS (FI_SOURCE | OFI_RX_MSG_CAPS | FI_TAGGED | \
-		     FI_DIRECTED_RECV | FI_MULTI_RECV)
+#define SM2_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED)
```
aingerson, line 36: Clang still off

prov/sm2/src/sm2_util.c
```diff
@@ -53,9 +52,8 @@ void sm2_cleanup(void)
 	struct dlist_entry *tmp;
 
 	pthread_mutex_lock(&sm2_ep_list_lock);
-	dlist_foreach_container_safe(&sm2_ep_name_list, struct sm2_ep_name,
-				     ep_name, entry, tmp)
-		free(ep_name);
+	dlist_foreach_container_safe(&sm2_ep_name_list, struct sm2_ep_name, ep_name,
+				     entry, tmp) free(ep_name);
```
aingerson, line 58: Formatting wrong

prov/sm2/src/sm2.h
```diff
@@ -208,36 +207,36 @@ struct sm2_queue {
...
+#define SM2_ZE_SOCK_PATH "/dev/shm/ze_"
 
-#define SM2_RMA_ORDER (OFI_ORDER_RAR_SET | OFI_ORDER_RAW_SET | FI_ORDER_RAS |	\
-		       OFI_ORDER_WAR_SET | OFI_ORDER_WAW_SET | FI_ORDER_WAS |	\
-		       FI_ORDER_SAR | FI_ORDER_SAW)
-#define sm2_fast_rma_enabled(mode, order) ((mode & FI_MR_VIRT_ADDR) && \
-			!(order & SM2_RMA_ORDER))
+#define SM2_RMA_ORDER                                                                    \
+(OFI_ORDER_RAR_SET | OFI_ORDER_RAW_SET | FI_ORDER_RAS | OFI_ORDER_WAR_SET |              \
+ OFI_ORDER_WAW_SET | FI_ORDER_WAS | FI_ORDER_SAR | FI_ORDER_SAW)
```
aingerson: Don't love how it does these. Would be interested in modifying but I'm not passionate about it.

aingerson (commented):
Just adding some comments on the clang formatting to document issues that I see that we can resolve as those get figured out. I can play around with it and see if I can figure out why the struct definitions are all off and all the other things.

prov/sm2/src/sm2_attr.c
aingerson: (this was supposed to be for the line under this one...)

shijin-aws:
@aingerson are you going to make the util SRX patches as a separate PR? It will be great we can make that through main branch first

aingerson:
@shijin-aws Yes, that's my plan. I deprioritized those to prioritize getting the sm2 changes through but am going to pick them back up and open a PR.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,831 @@
...
+}
+
+static void util_init_rx_entry(struct util_rx_entry *entry,
+			       const struct iovec *iov, void **desc,
+			       size_t count, fi_addr_t addr, void *context,
+			       uint64_t tag, uint64_t flags)
+{
+	int i;
+
+	for (i = 0; i < count; i++) {
```
qkoziol, line 63: Suggest refactoring this loop into a memcpy() on the iov[] and a single 'if' with a memcpy() or memset().  Should speed things up.  :-)

prov/sm2/src/sm2_atom.h
```diff
@@ -0,0 +1,115 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
```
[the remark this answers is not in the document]
a-szegel: This is the header from OMPI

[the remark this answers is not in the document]
a-szegel: removed

prov/sm2/src/sm2_atom.h
```diff
@@ -0,0 +1,115 @@
...
+ * Copyright (c) 2018      Los Alamos National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c) 2018      Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
+ * Copyright (c) 2019-2021 Google, LLC. All rights reserved.
+ * Copyright (c) 2019      Triad National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c) 2023      Amazon.com, Inc. or its affiliates.
+ *                         All Rights reserved.
+ * $COPYRIGHT$
```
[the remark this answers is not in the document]
a-szegel: Copied entire copyright from LICENSE into here

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	pid = getpid();
+	sprintf(template, "%s/fi_sm2_pid%d_XXXXXX", SM2_COORDINATION_DIR, pid);
+
+	char lock_status;
+
+	/* Assume we are the first.
+	   Open a tmpfile file in the shm directory */
+	fd = mkostemp(template, O_RDWR);
+	assert(fd > 0);
+	ftruncate(fd, sizeof(struct sm2_coord_file_header));
```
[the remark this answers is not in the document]
a-szegel: Need to fix returns still.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	/* mmap the file */
+	header = sm2_mmap_map(fd, &map_ours);
+
+	pthread_mutexattr_init(&att);
+	// pthread_mutexattr_setrobust(&att, PTHREAD_MUTEX_ROBUST);
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+		usleep(10000);
+	} while (tries-- > 0);
+
+	switch (lock_status) {
+	case 'F':
+		/* failed to acquire. now what? */
+		fprintf(
+		    stderr,
+		    "Failed to acquire the lock to the coordination file. Aborting\n");
+		abort();
```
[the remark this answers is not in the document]
a-szegel: Need to fix use of aborts.  Alignment should be fixed with new clang-format file.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	int pid = getpid();
+	int peer_pid;
+
+#ifndef NDEBUG
+	{
+		int pid = getpid();
+		int pid_hint = ofi_atomic_get32(&header->pid_lock_hint);
+		assert(pid_hint == pid);
+	}
+#endif
```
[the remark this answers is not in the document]
a-szegel: removed

prov/sm2/src/sm2_attr.c
a-szegel: We fixed clang-format file, and added , to the end of the structures

prov/sm2/src/sm2_av.c
a-szegel: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

clang-format 15 (which we use) has a parameter named InsertBraces, which will add braces for all 1 line if's, while's, for's etc... Unfortunately, it doesn't remove braces because most people like the braces (explicit is safer when you add more lines to the if, and easier to read).

I think clang-tidy might be able to do this but I am not sure.

Anyways... if we really care about this, it will have to be done manually from what I can tell.

.clang-format
```diff
@@ -0,0 +1,179 @@
...
+BreakBeforeConceptDeclarations: true
+BreakBeforeBraces: Stroustrup
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     90
```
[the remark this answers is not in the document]
a-szegel: done, merged in your clang-format fixes.

prov/sm2/src/sm2_attr.c
```diff
@@ -32,129 +32,116 @@
...
-	.rma_iov_limit = SM2_IOV_LIMIT
-};
+struct fi_tx_attr sm2_tx_attr = {.caps = SM2_TX_CAPS,
+				 .op_flags = SM2_TX_OP_FLAGS,
+				 .comp_order = FI_ORDER_NONE,
+				 .msg_order = FI_ORDER_SAS,
+				 .inject_size = SM2_INJECT_SIZE,
+				 .size = 1024,
+				 .iov_limit = SM2_IOV_LIMIT,
+				 .rma_iov_limit = SM2_IOV_LIMIT};
```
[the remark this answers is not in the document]
a-szegel: Fixed with the patch you provided by adding , to the end

prov/sm2/src/sm2_av.c
```diff
@@ -208,25 +209,25 @@ static const char *sm2_av_straddr(struct fid_av *av, const void *addr,
 }
 
 static struct fi_ops sm2_av_fi_ops = {
-	.size = sizeof(struct fi_ops),
-	.close = sm2_av_close,
-	.bind = ofi_av_bind,
-	.control = fi_no_control,
-	.ops_open = fi_no_ops_open,
+    .size = sizeof(struct fi_ops),
```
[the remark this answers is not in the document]
a-szegel: fixed, same as above

prov/sm2/src/sm2_av.c
```diff
@@ -70,20 +70,19 @@ static int sm2_av_insert(struct fid_av *av_fid, const void *addr, size_t count,
...
 			if (!ret) {
 				ofi_mutex_lock(&util_av->lock);
-				ret = ofi_av_insert_addr(util_av, &shm_id,
-							 &util_addr);
+				ret = ofi_av_insert_addr(util_av, &shm_id, &util_addr);
 				ofi_mutex_unlock(&util_av->lock);
 			}
-		} else {
+		}
+		else {
```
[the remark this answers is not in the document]
a-szegel: This looks fixed by new clang-format changes.

prov/sm2/src/sm2_av.c
```diff
@@ -271,9 +272,10 @@ int sm2_av_open(struct fid_domain *domain, struct fi_av_attr *attr,
 	(*av)->fid.ops = &sm2_av_fi_ops;
 	(*av)->ops = &sm2_av_ops;
 
-	ret = sm2_map_create(&sm2_prov, SM2_MAX_PEERS,
-			     util_domain->info_domain_caps & FI_HMEM ?
-			     SM2_FLAG_HMEM_ENABLED : 0, &sm2_av->sm2_map);
+	ret = sm2_map_create(
+	    &sm2_prov, SM2_MAX_PEERS,
+	    util_domain->info_domain_caps & FI_HMEM ? SM2_FLAG_HMEM_ENABLED : 0,
+	    &sm2_av->sm2_map);
```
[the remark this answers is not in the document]
a-szegel: this gets deleted b/c there is no map anymore.... looks like similar things like this are fixed throughout the patch

prov/sm2/src/sm2_attr.c
a-szegel: Is this good enough or should we do more investigation on clang-format style?

```
#define SM2_RX_CAPS                                                   \
	(FI_SOURCE | OFI_RX_MSG_CAPS | FI_TAGGED | FI_DIRECTED_RECV | \
	 FI_MULTI_RECV)
```

prov/sm2/src/sm2_common.h
```diff
@@ -35,48 +35,47 @@
...
-#define SM2_REMOTE_CQ_DATA	(1 << 0)
-#define SM2_RMA_REQ		(1 << 1)
-#define SM2_TX_COMPLETION	(1 << 2)
-#define SM2_RX_COMPLETION	(1 << 3)
-#define SM2_MULTI_RECV		(1 << 4)
+#define SM2_REMOTE_CQ_DATA (1 << 0)
+#define SM2_RMA_REQ (1 << 1)
+#define SM2_TX_COMPLETION (1 << 2)
+#define SM2_RX_COMPLETION (1 << 3)
+#define SM2_MULTI_RECV (1 << 4)
```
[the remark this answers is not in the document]
a-szegel: new clang-format file you provided aligns these

prov/sm2/src/sm2_util.c
a-szegel: We probably need to specify our macros similarly to the linux kernel clang-format file https://github.com/torvalds/linux/blob/master/.clang-format#L71

prov/sm2/src/sm2.h
a-szegel: Looks like you did manage to fix these in clang-format file.

prov/util/src/util_srx.c
a-szegel: @aingerson

shefty:
Why don't you just remove this provider completely, get a basic one that's functional that shows some benefit and add it back as a new provider?  That would be easier to deal with.  Any changes to common code can be submitted separately for review.

a-szegel:
So if we add a new provider, would that be 1 commit? for Inject?

a-szegel:
Also, why is this failing Intel CI?

a-szegel:
I don't know if @aingerson will agree with that approach.  Starting from scratch means that we do not care about the diff from SHM to SM2. Yes, the diff will be cleaner b/c everything will be green, but I am wondering if we are losing valuable history if we take that approach.

shefty:
@a-szegel - Yes.  As long as the new provider doesn't change any of the core files (other than the build and initialization code), then those changes can come in as a single commit.  The provider just needs to pass fabtests, so if it only supports transfers less than 4k (or some other very small value), that's okay.

This seems to be the approach being taken with sm2, only in a much less indirect way.

prov/sm2/src/sm2_coordination.h
```diff
@@ -0,0 +1,128 @@
...
+#define SM2_COORDINATOR_MAX_TRIES 1000
+#define SM2_COORDINATOR_MAX_UNIVERSE_SIZE 4096
+
+struct sm2_mmap {
+	char *base;
+	size_t size;
+	int fd;
+};
+
+struct sm2_private_aux {
```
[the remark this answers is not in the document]
lrbison: I am working to address several of these comments.  I've read through them, but I think it's worth I clarify something here to make sure I'm not missing anything.

SM2 is an FI_AV_TABLE, with one exception.  fi_addr_t's are assigned sequentially but their scope is not a single AV, but rather all fabrics using sm2 as the same user on the same system.  So instead of an AV_TABLE, we claim AV_MAP.

So back to your your comment "The fiaddrs are going to change depending on the EP".  This is not true in sm2.  The "Coordinator" is in fact coordinating on AV insert to generate these sequential, globally unique IDs.  This is to help out the Nemesis algorithm later, because Nemesis has to be able to address memory of peers that may never added to our AV.  

In our testing we've found that fi_av_map is indeed supported by open mpi, but might have a performance penalty.  I didn't expect the performance penalty so I didn't attempt to make a translation layer to fi_av_table.  I do think we have some pretty straight-forward approaches to address this, probably using the "auxiliary" structure, but you are suggesting we remove the sm2_aux struct in favor of util_av.

I am trying to understand how we could best implement this with util av and it's not making sense to me.   I'll admit some of that is because I'm not familiar with the util_av and I'm having trouble finding documentation on it (since it's internal).

Would you mind describing how you would use the util_av to store and retrieve the global id?

Thanks

prov/sm2/src/sm2_common.h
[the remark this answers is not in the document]
a-szegel: done in single commit

prov/sm2/src/sm2_coordination.c
a-szegel: I guess we will have to add manual new lines for now... or maybe we shouldn't care b/c editors should take care of this.

vscode settings.json:

```
{
    "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true
}
```

Note: You will probably get feedback to separate these formatting changes into their own commits.

prov/sm2/src/sm2_attr.c
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2015-2018 Intel Corporation. All rights reserved.
+ * Copyright (c) 2023 Amazon.com, Inc. or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_ep.c
a-szegel: @lrbison is going to own the OMPI bugfix

prov/sm2/configure.m4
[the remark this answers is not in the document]
a-szegel: fixed formatting of configure/make files

prov/sm2/src/sm2_util.c
a-szegel: done

prov/sm2/src/sm2_attr.c
a-szegel: I am going to assume this is good enough, if you still don't like it... call it out

prov/sm2/src/sm2_common.h
a-szegel: Formatting is fixed. 

sm2_msg_header was essentially renamed sm2_protocol_header.  At one point, I thought I was going to have multiple data members relating to the Nemesis algorithm, and I wanted to group them all together.  It seems like I only need one, the next pointer.  I was planning on having:

```
sm2_msg_header {
   struct sm2_nemesis_hdr
   struct sm2_protocol_hdr
   char data[4096]
};
```

I don't think naming between sm2_protocol_hdr and sm2_msg_hdr should block the PR.

prov/sm2/src/sm2_common.h
a-szegel: fixed.

prov/sm2/src/sm2_attr.c
a-szegel: not anymore that we are one commit

prov/sm2/src/sm2_coordination.h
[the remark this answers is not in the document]
aingerson: Yeah totally. First of all, you're totally right in the sense that the above comment (my original one about the EP fiaddrs being different) is incorrect when you use FI_AV_MAP :) I think I reviewed that before I saw you changed the av type.
We can choose to change it to AV_MAP and what you're saying above would be correct though I still don't think you would need the aux struct since the fiaddr would just be equivalent to the global index or some pointer to the info. There's no reason to maintain a backwards lookup since the fiaddr is the internal pointer or index. That's the whole point of FI_AV_MAP is that the provider doesn't have to maintain an index lookup but the application has to instead. This is really useful for more complicated addressing but since the global_id is already indexed, I don't think it's necessary in this use case and would just add overhead and potential limitations to the application. FI_AV_TABLE allows the application to directly use fi_addrs as index lookups. When you make the application use a map, it requires them to do some translation between the fi_addr and their addressing. Therefore FI_AV_TABLE is recommended and I think the implementation you have here is really close to supporting it. I think you'll need the aux struct in the av (though I would remove the struct in general and just make it a simple array allocated in av_init) as you're using it. This will let you go from global_id back to the fiaddr in order to report rx completions. On the send side, you can use the util_av to go from fiaddr to the global_id. Basically you're just using 2 arrays fiaddr[global_id] and global_id[fiaddr]. The first of those arrays is the one currently in the util_av and the second one is what you have with the aux struct. Alternatively, you can get rid of the util_av and just use those 2 arrays if that makes more sense to you.

Basically, what I'm saying is, this implementation is really close to FI_AV_TABLE so unless there's a huge reason to move to FI_AV_MAP, I would suggest leaving it at FI_AV_TABLE so we don't impose that limitation and overhead to the app.

prov/sm2/src/sm2_av.c
a-szegel: Cleaned up FI_WARN's throughout the provider

prov/sm2/src/sm2_attr.c
a-szegel: It will be added in soon and this is just a place holder.

prov/sm2/src/sm2.h
a-szegel: removed

prov/sm2/src/sm2.h
a-szegel: removed

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+#define ZOMBIE_ALLOCATION_NAME "ZOMBIE"
+
+static void sm2_coordinator_attempt_shm_file_shrink(struct sm2_mmap *map);
+
+/**
+ * @brief take an open file, and mmap its contents
+ *
+ * @param[in] fd
+ * @param[out] map
+ */
```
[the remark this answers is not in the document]
a-szegel: I really do not want to do this.  The function description you pointed out is not very useful, but I think the majority of comments (both function description and in-code comments) are incredibly useful for understanding.  I understand that there are different perspectives on the value of comments.  Comments are often the first part of code to rot.  But I think these are super useful for what is going on (not all, but most).  I would like to be able to keep the useful comments in the code, and I will happily remove the ones that aren't useful.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	    header->ep_enumerations_offset +
+	    (header->ep_enumerations_max * sizeof(struct sm2_ep_allocation_entry));
+
+	header->ep_regions_offset = NEXT_MULTIPLE_OF(header->ep_regions_offset, 4096);
+
+	/* allocate enough space in the file for all our allocations, but no
+	   data exchange regions yet. */
+	header = sm2_mmap_remap(&map_ours, header->ep_regions_offset);
+	entries = sm2_mmap_entries(&map_ours);
+	for (int jentry = 0; jentry < header->ep_enumerations_max; jentry++) {
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	pthread_mutex_lock(&header->write_lock);
+
+	ofi_atomic_initialize32(&header->pid_lock_hint, pid);
+	header->file_version = 1;
+	header->ep_region_size = 16777216; // TODO: base on sm2_calculate_size
+	header->ep_enumerations_max = SM2_COORDINATOR_MAX_UNIVERSE_SIZE;
+
+	header->ep_enumerations_offset = sizeof(struct sm2_coord_file_header);
+	header->ep_enumerations_offset =
+	    NEXT_MULTIPLE_OF(header->ep_enumerations_offset, 4096);
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+			}
+			sm2_mmap_unmap_and_close(map_shared);
+		}
+		/* we could not acquire the lock, sleep and try again. */
+		// TODO Consider reducing the amount of time we sleep here
+		usleep(10000);
+	} while (tries-- > 0);
+
+	switch (lock_status) {
+	case 'F':
```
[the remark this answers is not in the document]
a-szegel: Done, using ENUM

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+ *			When False, we set pid = -getpid(), allowing owner to claim later
+ * @return 0 on success, some negative FI_ERROR on error
+ */
+ssize_t sm2_coordinator_allocate_entry(const char *name, struct sm2_mmap *map,
+				       int *av_key, bool self)
+{
+
+	struct sm2_coord_file_header *header = (void *)map->base;
+	struct sm2_ep_allocation_entry *entries;
+	int jentry;
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+			// and there is no current way to check this... need to keep this
+			// entry in the file until we clean up
+			continue;
+		}
+		else {
+			if (!pid_lives(peer_pid)) {
+				sm2_coordinator_extend_for_entry(map, jentry);
+				entries = sm2_mmap_entries(map);
+				header = (void *)map->base;
+				struct sm2_region *peer_region =
```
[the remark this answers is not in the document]
a-szegel: done as far as I can tell

prov/sm2/src/sm2_common.h
a-szegel: reverted back to msg_hdr

prov/sm2/src/sm2.h
a-szegel: renamed to gid

prov/sm2/src/sm2.h
a-szegel: removed

prov/sm2/src/sm2.h
a-szegel: 1 patch

prov/sm2/src/sm2.h
```diff
@@ -119,7 +120,7 @@ struct sm2_rx_entry {
 };
 
 struct sm2_tx_entry {
-	struct sm2_cmd cmd;
+	struct sm2_free_queue_entry cmd;
```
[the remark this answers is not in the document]
a-szegel: deleted sm2_tx_entry b/c it isn't needed.

Each endpoint has a free queue (your naming and Nemesis's naming align here), and that queue is initialized with a finite number of free queue entries.  The free queue entry holds the sm2 assigned header and the user's message in SharedMemory, and can be placed on a peers recv queue.

Command (cmd) doesn't add any value b/c we are not commanding anything.  sm2_cmd is nice because it has 6 letters instead of 20.  I would happily change `sm2_free_queue_entry` to `sm2_fqe` for conciseness.  I think the word we are really looking for is a wrapper around a header + message.  In this sense, I think fqe does a better job than command, b/c it is clear that the free queue entry lives on the free queue.

Also, in SM2, the sender owns the FQ, FQE whereas in SHM, the receiver owns the CMD + CMD QUEUE.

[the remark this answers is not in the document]
a-szegel: switching to `sm2_fqe`

prov/sm2/src/sm2_coordination.c
a-szegel: re-named av_key to gid

prov/sm2/src/sm2_coordination.c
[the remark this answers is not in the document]
shefty: Comments are detrimental to maintenance, outside of published APIs.  It's another area of the code that MUST be changed when the source is changed -- a change that is frequently skipped -- resulting in comments that are not aligned with the code, and is most often used to avoid designing a function call that is obvious in its use, or which tries to do way more than it should.  Make the functions names clear as to what they do, with arguments that are obvious.  Use const to distinguish between input and output pointers.

Comments should be for things that aren't obvious by reading the code, like lock ordering.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#define NEXT_MULTIPLE_OF(x, mod) x % mod ? ((x / mod) + 1) * mod : x
+#define ZOMBIE_ALLOCATION_NAME "ZOMBIE"
+
+static void sm2_coordinator_attempt_shm_file_shrink(struct sm2_mmap *map);
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.h
```diff
@@ -0,0 +1,128 @@
...
+#include <ofi_rbuf.h>
+#include <ofi_tree.h>
+
+#include <rdma/providers/fi_prov.h>
+
+#define SM2_DIR "/dev/shm/"
+#define SM2_NAME_MAX 256
+#define SM2_PATH_MAX (SM2_NAME_MAX + sizeof(SM2_DIR))
+#define SM2_COORDINATION_FILE "/dev/shm/fi_sm2_mmaps"
+#define SM2_COORDINATION_DIR "/dev/shm"
```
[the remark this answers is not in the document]
a-szegel: Dropped what isn't used and kept what is used.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+			goto out;
+	}
+	else if (st.st_size >= map->size) {
+		/* file has been extended by another process since
+		   last we checked.  Re-map the entire file. */
+		at_least = st.st_size;
+	}
+	else {
+		/* file has shrunk since we checked. */
+		// TODO we should find a way to tell other processes to re-map and
```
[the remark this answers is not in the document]
a-szegel: This scenario should never happen and is here for completeness.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+		map->base =
+		    mmap(0, at_least, PROT_READ | PROT_WRITE, MAP_SHARED, map->fd, 0);
+		if (OFI_UNLIKELY(map->base == MAP_FAILED)) {
+			perror("failed to remap when increasing the map size in "
+			       "sm2_mmap_remap");
+			FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+				"Failed syscall during sm2_mmap_map(). st.st_size=%ld\n "
+				"at_least=%ld",
+				st.st_size, at_least);
+			abort();
```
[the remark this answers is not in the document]
a-szegel: This is an unrecoverable error and requires an abort.  sm2_mmap_remap() is used in the critical path, so we cannot return NULL, or we would have to check for NULL, or core.  Abort is the right behavior if we fail to re-map.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	// pthread_mutexattr_setrobust(&att, PTHREAD_MUTEX_ROBUST);
+	pthread_mutexattr_setpshared(&att, PTHREAD_PROCESS_SHARED);
+	pthread_mutex_init(&header->write_lock, &att);
+	/* since this is currently our own private mutex,
+	   this lock is uncontested */
+	pthread_mutex_lock(&header->write_lock);
+
+	ofi_atomic_initialize32(&header->pid_lock_hint, pid);
+	header->file_version = 1;
+	header->ep_region_size = 16777216; // TODO: base on sm2_calculate_size
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.h
```diff
@@ -0,0 +1,128 @@
...
+#define SM2_COORDINATOR_MAX_TRIES 1000
+#define SM2_COORDINATOR_MAX_UNIVERSE_SIZE 4096
```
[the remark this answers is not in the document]
a-szegel: kept the file name but removed it via function/defines

prov/sm2/src/sm2_coordination.c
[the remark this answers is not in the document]
a-szegel: can't fix aborts... left comment open on that above.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+			return -FI_EADDRINUSE;
+		}
+	}
+
+	/* fine, we could not find the entry, so now look for an empty slot */
+	for (jentry = 0; jentry < header->ep_enumerations_max; jentry++) {
+		peer_pid = entries[jentry].pid;
+		if (peer_pid == 0)
+			goto found;
+		else if (peer_pid < 0) {
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+			// A third peer might have entered this address into their AV,
+			// and there is no current way to check this... need to keep this
+			// entry in the file until we clean up
+			continue;
+		}
+		else {
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.h
```diff
@@ -0,0 +1,128 @@
...
+	pthread_mutex_t write_lock;
+	ofi_atomic32_t pid_lock_hint;
+	int64_t ep_region_size;
+	int ep_enumerations_max;
+
+	ptrdiff_t ep_enumerations_offset; /* struct sm2_ep_allocation_entry */
+	ptrdiff_t ep_regions_offset;	  /* struct ep_region */
+};
+
+ssize_t sm2_mmap_unmap_and_close(struct sm2_mmap *map);
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "sm2_coordination.h"
```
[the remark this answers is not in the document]
a-szegel: I don't think this makes sense.  The coordinatior does more than just mmap.  It is in charge of assigning global ID's in a shared file.  I removed the coordinatior from the function names as a compromise.

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+	/* mmap the file */
+	header = sm2_mmap_map(fd, &map_ours);
+
+	pthread_mutexattr_init(&att);
+	// pthread_mutexattr_setrobust(&att, PTHREAD_MUTEX_ROBUST);
+	pthread_mutexattr_setpshared(&att, PTHREAD_PROCESS_SHARED);
+	pthread_mutex_init(&header->write_lock, &att);
+	/* since this is currently our own private mutex,
+	   this lock is uncontested */
+	pthread_mutex_lock(&header->write_lock);
```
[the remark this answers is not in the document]
a-szegel: The idea is that we create our new Shared Memory file and we immediately grab the lock.  There will be no contention on this lock (until the file gets hard linked) so it doesn't matter when it is taken.

prov/sm2/src/sm2_coordination.h
```diff
@@ -0,0 +1,128 @@
...
+#define SM2_COORDINATOR_MAX_UNIVERSE_SIZE 4096
+
+struct sm2_mmap {
+	char *base;
+	size_t size;
+	int fd;
+};
+
+struct sm2_private_aux {
+	fi_addr_t cqfid; // fi_addr to report during in cqe's if av_user_data is on
```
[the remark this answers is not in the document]
a-szegel: changing to cq_id

prov/sm2/src/sm2_av.c
a-szegel: done

prov/sm2/src/sm2_av.c
a-szegel: done

prov/sm2/src/sm2_av.c
a-szegel: done

prov/sm2/src/sm2_coordination.c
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.h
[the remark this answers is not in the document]
a-szegel: Switched to use util_av + an array of FI_ADDRs (removed struct sm2_private_aux)

prov/sm2/src/sm2.h
```diff
@@ -86,8 +87,8 @@ int sm2_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric, void *co
 
 struct sm2_av {
 	struct util_av util_av;
-	struct sm2_map *sm2_map;
-	size_t used;
+	struct sm2_mmap sm2_mmap;
+	struct sm2_private_aux *sm2_aux;
```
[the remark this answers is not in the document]
a-szegel: removed

a-szegel:
Closing this because I don't think there is anything of value still on here... it will forever be in our history for us to jump back to :)

prov/sm2/src/sm2_av.c
a-szegel: fixed

prov/sm2/src/sm2_attr.c
a-szegel: fixed

prov/sm2/src/sm2.h
a-szegel: fixed

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+
+	/* Assume we are the first.
+	   Open a tmpfile file in the shm directory */
+	fd = mkostemp(template, O_RDWR);
+	assert(fd > 0);
+	ftruncate(fd, sizeof(struct sm2_coord_file_header));
+	/* mmap the file */
+	header = sm2_mmap_map(fd, &map_ours);
+
+	pthread_mutexattr_init(&att);
```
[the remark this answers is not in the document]
a-szegel: comment captured in new PR

prov/sm2/src/sm2_coordination.c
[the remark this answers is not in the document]
a-szegel: captured in new PR

prov/sm2/src/sm2_coordination.c
```diff
@@ -0,0 +1,675 @@
...
+		sm2_coordinator_attempt_shm_file_shrink(map_shared);
+		sm2_mmap_unmap_and_close(&map_ours);
+		break;
+	case 'O':
+		/* We are using the memory we initialized.  Duplicate it and
+		   leave the map open. */
+		memcpy(map_shared, &map_ours, sizeof(struct sm2_mmap));
+		break;
+	default:
+		/* catch cosmic ray bit-flips */
```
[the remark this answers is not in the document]
a-szegel: done

prov/sm2/src/sm2_coordination.h
```diff
@@ -0,0 +1,128 @@
...
+	int pid; // This is for verify peer, make sure entry didn't get replaced with
+		 // someone new
+};
+
+struct sm2_ep_allocation_entry {
```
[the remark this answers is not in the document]
a-szegel: transferred to new PR

prov/sm2/src/sm2_coordination.c
a-szegel: Will discuss entries table in shared mem.

a-szegel:
officially answered all comments here... now this is no longer usefule.

PR #8783, opened 2023-04-11
prov/efa: RMA operations will ensure FI_RMA cap

lrbison:
Before performing any RMA operation, ensure that the application has set the FI_RMA capability in fi_info during domain initialization.


prov/efa/src/rdm/rxr_rma.c
```diff
@@ -155,6 +155,10 @@ ssize_t rxr_rma_readmsg(struct fid_ep *ep, const struct fi_msg_rma *msg, uint64_
 	       flags);
 
 	rxr_ep = container_of(ep, struct rxr_ep, base_ep.util_ep.ep_fid.fid);
+	if ( OFI_UNLIKELY(0 == rxr_ep_domain(rxr_ep)->info->caps & FI_RMA) ) {
```
wenduwan, line 158: Since it is a common util, should we extract it to static inline?

wenduwan (approved):
Thank you

prov/efa/src/rdm/rxr_rma.c
```diff
@@ -155,6 +155,10 @@ ssize_t rxr_rma_readmsg(struct fid_ep *ep, const struct fi_msg_rma *msg, uint64_
 	       flags);
 
 	rxr_ep = container_of(ep, struct rxr_ep, base_ep.util_ep.ep_fid.fid);
+	if ( OFI_UNLIKELY(0 == (rxr_ep_domain(rxr_ep)->info->caps & FI_RMA)) ) {
+		return -FI_EOPNOTSUPP;
```
wzamazon, line 159: Should we add `FI_WARN_ONCE` here?

wzamazon (commented):
Since you are here, can you add similar check in `rxr_atomic.c`

prov/efa/src/rdm/rxr_rma.c
wzamazon:
```
rxr_ep_domain(rxr_ep)->info
```
is not the right info object to use. You need to use `rxr_ep->user_info`.

wzamazon: Probably add a unit test to verify the behavior

prov/efa/src/rdm/rxr_rma.c
lrbison: Done: `rxr_ep_cap_check_rma` and `rxr_ep_cap_check_atomic`

lrbison:
@wzamazon FI_ATOMIC capability is checked, and warnings added.

Thinking about unit testing, the RMA operations need a remote to issue RMA to, and exchanging keys with MR, etc.  That feels like a lot of mocks or maybe I'm missing something.

For the atomics, we could simply check fi_atomicvalid without having a remote.

What do you think about deferring unit tests to some other time?

wzamazon:
> Thinking about unit testing, the RMA operations need a remote to issue RMA to, and exchanging keys with MR, etc. That feels like a lot of mocks or maybe I'm missing something.

An endpoint can do rma to itself. See example here: https://github.com/ofiwg/libfabric/blob/bf1828bed67f32a646c4eb1c2c06437da1538b8d/prov/efa/test/efa_unit_test_ep.c#L407

prov/efa/src/rdm/rxr_rma.c
lrbison: unit test added.

lrbison:
Added unit tests, rebased these before the commits which add the checks.  Sure enough, they initially fail, and at the end of this PR, they pass.

Both Atomics and RMA are checked.

wzamazon (approved):

wzamazon:
Thank you!

wzamazon:
socket provider failure is not related to this PR

PR #8789, opened 2023-04-11
include/ofi_atom: fix no atomics support

aingerson:
When build with --enable-atomics=no, OFI defines its own set of emulated atomics. Some functions were missing from that definition:
	ofi_atomic_compare_exchange_weak
	ofi_atomic_store_explicit
	ofi_atomic_load_explicit

The memory order definitions need to be included in the no atomics as well. Move the ifdef to include them.


shefty:
Changes passed CI for v1.18.x testing.  Failing here due to node availability.  Ignoring and merging anyway.

PR #8792, opened 2023-04-11
CI: Run CI on every patch in Change Set

a-szegel:
The libfabric community wants to make sure that every patch in a change set builds.  We should enforce this behavior.

I haven't tested this b/c I don't know how, but I got the idea from here: https://stackoverflow.com/questions/64708371/how-to-run-github-workflow-on-every-commit-of-a-push

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -122,12 +122,18 @@ jobs:
            brew install automake
            brew install libtool
       - uses: actions/checkout@v2
+        with:
+          # checkout full tree
+          fetch-depth: 0
       - name: Build Check
         run: |
-          ./autogen.sh
-          ./configure --prefix=$PWD/install
-          make -j 2; make install
-          $PWD/install/bin/fi_info -l
+          for commit in $(git rev-list origin/main..${{ github.sha}}); do
+            git checkout $commit
+            ./autogen.sh
+            ./configure --prefix=$PWD/install
+            make clean; make -j 2; make install
+            $PWD/install/bin/fi_info -l
+          done
       - name: Upload build logs
         if: failure()
         uses: actions/upload-artifact@v2


darrylabbate:
You can probably test this by committing it to `main` in your own fork and opening a dummy PR with multiple commits

darrylabbate (commented):
I would suggest only running the build checks on every commit (sanity check) while running the full CI only against the tip of the PR branch. Just my 2c

a-szegel:
most recent test: https://github.com/a-szegel/libfabric/pull/12 ... still running

a-szegel:
This works now as expected

darrylabbate:
@shefty Just curious, where are we in terms of average monthly GitHub Actions resource utilization? I believe it's measured in total minutes per month

shefty:
@darrylabbate - I have to request a report, which can take a couple hours.  I've done that, but it'll be a while before we have results.  But the project is listed as limited to 2000 minutes per month.  (I can't tell if that is enforced.)  The actions take about 20 minutes on each PR.  So multiply that by each time the PR is forced update.

This PR isn't solving an actual problem and can increase the action time significantly.

shefty:
Btw, I just got the github report, and it's blank (titles only).  I'm going to assume this is because we don't have an account and use the free option.

a-szegel:
Finally got it working... switched it to just use macos build to run every commit

shefty:
The macos build is limited to 5 parallel sessions, isn't a prioritized development target, and can't build most of the providers.  We're not solving a problem with this, only greatly increasing CI time, which already takes too long. There is no good justification to increase CI resources and power to build a commit that has maybe a 0.01% chance of ever being built.  And this still doesn't guarantee that the build can't break, since parallel PRs can merge into unbuildable code.

a-szegel:
I picked MacOS b/c it didn't have a matrix... but in doing that removed a bunch of the usefulness of this.  I am ok dropping this.

PR #8796, opened 2023-04-12
prov/efa: make pkt_entry->x_entry to be rxr_op_entry

wzamazon:
Prior to this patch pkt_entry->x_entry's type is "void *", after we eliminate rxr_read_entry, this pointer can only be pointer to  rxr_op_entry. This patch makes this explicit.

### DIFF
--- a/prov/efa/src/rdm/rxr_pkt_entry.h
+++ b/prov/efa/src/rdm/rxr_pkt_entry.h
@@ -158,8 +158,8 @@ struct rxr_pkt_entry {
 	struct dlist_entry dbg_entry;
 	uint8_t pad[48];
 #endif
-	/** @brief pointer to #rxr_op_entry or #rxr_read_entry */
-	void *x_entry;
+	/** @brief pointer to #rxr_op_entry */
+	struct rxr_op_entry *x_entry;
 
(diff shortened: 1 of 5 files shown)


wenduwan (approved):

prov/efa/src/rdm/rxr_pkt_entry.h
```diff
@@ -158,8 +158,8 @@ struct rxr_pkt_entry {
 	struct dlist_entry dbg_entry;
 	uint8_t pad[48];
 #endif
-	/** @brief pointer to #rxr_op_entry or #rxr_read_entry */
-	void *x_entry;
+	/** @brief pointer to #rxr_op_entry */
+	struct rxr_op_entry *x_entry;
```
lrbison, line 162: since this is now always an op_entry, will there be a follow-up to rename x_entry to just `entry` or `ope`?

lrbison (approved):

prov/efa/src/rdm/rxr_pkt_entry.h
wzamazon: Yes.

Issue #8784, opened 2023-04-11
fabtests: fi_rma_bw using WRITE doesn't perform verification


lrbison:
On main branch, while testing a development driver, we find that `fi_rma_bw -e rdm -o writedata -j 0 -p efa -v -E` is finding data validation errors that should also occur in ` fi_rma_bw -e rdm -o write -j 0 -p efa -v -E`, yet the write test passes without complaint.

Need to investigate and fix the missing validation for the write case.

shefty:
It's possible the test itself is wrong.  A bw tests repeatedly sends writes back to back.  If the target is supposed to verify the data, each write needs to go to a separate target buffer.  I would first check that the test does this before assuming there is a problem in the provider.

lrbison:
So the test does indeed overwrite the same same memory buffer over and over, and as you say Wei there is some difficulty in knowing when is a valid time to test for valid data arriving on the remote side.

The window size is a convenient break.  Right now for a window size of W, fabtest submits W writes, then gets all W completions.  I decided that rather than putting each write into a separate buffer or separate section of the buffer, I that I would just verify data during this completion phase.

For hardware accelerated RDMA this seems to work, after I add sufficient synchronization, but when I was testing emulated RDMA on reads, I find that the buffer is corrupt.  I think this might be because ft_sync is using the same buffers that RDMA is using, and in fact ft_sync sends 1-byte messages in these buffers, but it also takes the time to set them up with valid data.

I'm hopeful I can figure out a well-rounded approach tomorrow.

aingerson:
@lrbison In case it is helpful - the common code already allocates a buffer big enough to hold window_size number of transfers (8M * 64 * 2 - one for tx and one for rx) so I think to make it use different parts of the buffer would be pretty simple. You can just change the remote that you pass into the rma call to use an offset.

lrbison:
I have taken the advice of both @wzamazon and @aingerson in a PR I submitted which should address this issue: https://github.com/ofiwg/libfabric/pull/8798

PR #8798, opened 2023-04-12
Enable data integrity checks for fi_rma_bw

lrbison:
A series of commits to enable `-o write` and `-o read` to perform data validation.

There is a notable performance impact when validation is enabled.

Addresses https://github.com/ofiwg/libfabric/issues/8784

### DIFF
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -166,6 +166,25 @@ static int bw_rx_comp()
 	return ft_tx(ep, remote_fi_addr, 4, &tx_ctx);
 }
 
+static int rma_bw_rx_comp()
+{
+	int ret;
+
+	/* rx_seq is always one ahead */
+	ret = ft_get_rx_comp(rx_seq - 1);
+	if (ret)
+		return ret;
+
+	if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
+		ret = ft_check_buf((char *) rx_buf,
+				   opts.transfer_size * opts.window_size);
+		if (ret)
+			return ret;
+	}
+
+	return ft_tx(ep, remote_fi_addr, 4, &tx_ctx);
+}
+
 int bandwidth(void)
 {
 	int ret, i, j, inject_size;
@@ -256,7 +275,7 @@ static int bw_rma_comp(enum ft_rma_opcodes rma_op)
 		if (opts.dst_addr) {
 			ret = bw_tx_comp();
 		} else {
-			ret = bw_rx_comp();
+			ret = rma_bw_rx_comp();
 		}
 	} else {
 		ret = ft_get_tx_comp(tx_seq);
@@ -270,6 +289,7 @@ static int bw_rma_comp(enum ft_rma_opcodes rma_op)
 int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 {
 	int ret, i, j, inject_size;
+	size_t offset;
 
 	inject_size = inject_size_set ?
 			hints->tx_attr->inject_size: fi->tx_attr->inject_size;
@@ -281,17 +301,28 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 	if (ret)
 		return ret;
 
+	offset = 0;
+
 	for (i = j = 0; i < opts.iterations + opts.warmup_iterations; i++) {
 		if (i == opts.warmup_iterations)
 			ft_start();
+		if (j == 0 ) {
+			offset = 0;
 
+			if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
+				ret = ft_fill_buf((char *) tx_buf + offset,
+					opts.transfer_size*opts.window_size);
+				if (ret)
+					return ret;
+			}
+		}
 		switch (rma_op) {
 		case FT_RMA_WRITE:
 			if (opts.transfer_size < inject_size) {
-				ret = ft_post_rma_inject(FT_RMA_WRITE, tx_buf,
+				ret = ft_post_rma_inject(FT_RMA_WRITE, tx_buf + offset,
 						opts.transfer_size, remote);
 			} else {
-				ret = ft_post_rma(FT_RMA_WRITE, tx_buf,
+				ret = ft_post_rma(FT_RMA_WRITE, tx_buf + offset,
 						opts.transfer_size, remote,
 						&tx_ctx_arr[j].context);
 			}
@@ -308,28 +339,21 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 					rx_seq++;
 
 			} else {
-				if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
-					ret = ft_fill_buf((char *) tx_buf + ft_tx_prefix_size(),
-							  opts.transfer_size);
-					if (ret)
-						return ret;
-				}
-
 				if (opts.transfer_size < inject_size) {
 					ret = ft_post_rma_inject(FT_RMA_WRITEDATA,
-							tx_buf,
+							tx_buf + offset,
 							opts.transfer_size,
 							remote);
 				} else {
 					ret = ft_post_rma(FT_RMA_WRITEDATA,
-							tx_buf,
+							tx_buf + offset,
 							opts.transfer_size,
 							remote,	&tx_ctx_arr[j].context);
 				}
 			}
 			break;
 		case FT_RMA_READ:
-			ret = ft_post_rma(FT_RMA_READ, rx_buf, opts.transfer_size,
+			ret = ft_post_rma(FT_RMA_READ, rx_buf + offset, opts.transfer_size,
 					remote,	&tx_ctx_arr[j].context);
 			break;
 		default:
@@ -345,6 +369,7 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 				return ret;
 			j = 0;
 		}
+		offset += opts.transfer_size;
 	}
 	ret = bw_rma_comp(rma_op);
 	if (ret)
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -288,11 +288,12 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 		switch (rma_op) {
 		case FT_RMA_WRITE:
 			if (opts.transfer_size < inject_size) {
-				ret = ft_post_rma_inject(FT_RMA_WRITE, ep,
+				ret = ft_post_rma_inject(FT_RMA_WRITE, tx_buf,
 						opts.transfer_size, remote);
 			} else {
-				ret = ft_post_rma(rma_op, ep, opts.transfer_size,
-						remote,	&tx_ctx_arr[j].context);
+				ret = ft_post_rma(FT_RMA_WRITE, tx_buf,
+						opts.transfer_size, remote,
+						&tx_ctx_arr[j].context);
 			}
 			break;
 		case FT_RMA_WRITEDATA:
@@ -316,19 +317,19 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 
 				if (opts.transfer_size < inject_size) {
 					ret = ft_post_rma_inject(FT_RMA_WRITEDATA,
-							ep,
+							tx_buf,
 							opts.transfer_size,
 							remote);
 				} else {
 					ret = ft_post_rma(FT_RMA_WRITEDATA,
-							ep,
+							tx_buf,
 							opts.transfer_size,
 							remote,	&tx_ctx_arr[j].context);
 				}
 			}
 			break;
 		case FT_RMA_READ:
-			ret = ft_post_rma(FT_RMA_READ, ep, opts.transfer_size,
+			ret = ft_post_rma(FT_RMA_READ, rx_buf, opts.transfer_size,
 					remote,	&tx_ctx_arr[j].context);
 			break;
 		default:
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -39,13 +39,19 @@
 #include "shared.h"
 #include "benchmark_shared.h"
 
-/*
- * when the -j option is set, user supplied inject_size must be honored,
+/* when the -j option is set, user supplied inject_size must be honored,
  * even if the provider may return a larger value. This flag is used to
  * distinguish between the '-j 0' option and no '-j' option at all. For
  * both cases hints->tx_attr->inject_size is 0.
  */
 static int inject_size_set;
+
+/* When performing RMA with validation, READ needs to ensure it deconflicts
+ * it's memory access with the messages sent by ft_sync().  Do this by
+ * offsetting all RMA operations away from the beginning of the buffer and
+ * leave ft_sync to operate in that region.
+ */
+static int offset_rma_start = 0;
 
 void ft_parse_benchmark_opts(int op, char *optarg)
 {
@@ -144,7 +150,7 @@ static int bw_tx_comp()
 	ret = ft_get_tx_comp(tx_seq);
 	if (ret)
 		return ret;
-	return ft_rx(ep, 4);
+	return ft_rx(ep, FT_RMA_SYNC_MSG_BYTES);
 }
 
 static int bw_rx_comp()
@@ -163,7 +169,7 @@ static int bw_rx_comp()
 			return ret;
 	}
 
-	return ft_tx(ep, remote_fi_addr, 4, &tx_ctx);
+	return ft_tx(ep, remote_fi_addr, FT_RMA_SYNC_MSG_BYTES, &tx_ctx);
 }
 
 static int rma_bw_rx_comp()
@@ -175,14 +181,7 @@ static int rma_bw_rx_comp()
 	if (ret)
 		return ret;
 
-	if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
-		ret = ft_check_buf((char *) rx_buf,
-				   opts.transfer_size * opts.window_size);
-		if (ret)
-			return ret;
-	}
-
-	return ft_tx(ep, remote_fi_addr, 4, &tx_ctx);
+	return ft_tx(ep, remote_fi_addr, FT_RMA_SYNC_MSG_BYTES, &tx_ctx);
 }
 
 int bandwidth(void)
@@ -267,23 +266,31 @@ int bandwidth(void)
 	return 0;
 }
 
-static int bw_rma_comp(enum ft_rma_opcodes rma_op)
+static int bw_rma_comp(enum ft_rma_opcodes rma_op, int num_completions)
 {
 	int ret;
 
 	if (rma_op == FT_RMA_WRITEDATA) {
-		if (opts.dst_addr) {
-			ret = bw_tx_comp();
-		} else {
-			ret = rma_bw_rx_comp();
-		}
+		/* for writedata, only the client sends,
+		 * and only the server verifies. */
+		if (opts.dst_addr)
+			return bw_tx_comp();
+		ret = rma_bw_rx_comp();
 	} else {
 		ret = ft_get_tx_comp(tx_seq);
 	}
+
 	if (ret)
 		return ret;
 
-	return 0;
+	if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
+		if (rma_op == FT_RMA_WRITE)
+			ft_sync();
+		ret = ft_check_buf(rx_buf + offset_rma_start,
+			           opts.transfer_size * num_completions);
+	}
+
+	return ret;
 }
 
 int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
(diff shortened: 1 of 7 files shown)


fabtests/include/shared.h
```diff
@@ -526,11 +526,11 @@ ssize_t ft_rx(struct fid_ep *ep, size_t size);
 ssize_t ft_tx(struct fid_ep *ep, fi_addr_t fi_addr, size_t size, void *ctx);
 ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size);
 ssize_t ft_post_rma(enum ft_rma_opcodes op, struct fid_ep *ep, size_t size,
-		struct fi_rma_iov *remote, void *context);
+		struct fi_rma_iov *remote, void *context, size_t offset);
```
shefty, line 529: It would make more sense to just pass in the buffer directly, rather than using a global and giving an offset.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -47,6 +47,14 @@
  */
 static int inject_size_set;
 
+/*
+ * When performing RMA with validation, READ needs to ensure it deconflicts
+   it's memory access with the messages sent by ft_sync().  Do this by
+   offsetting all RMA operations away from the beginning of the buffer and
+   leave ft_sync to operate in that region.
+ */
```
shefty, line 55: nit: multi-line comments are usually written as:
```
/* Start writing here, and then
 * continue and
 * continue and
 * continue until done.
 */
```
comment applies to other newly added comments.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -47,6 +47,14 @@
...
+static int offset_rma_start=0;
```
shefty, line 56: nit: add space before/after =

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -267,28 +275,46 @@ int bandwidth(void)
...
 		}
-	} else {
+	} else if (rma_op == FT_RMA_WRITE) {
+		ret = ft_get_tx_comp(tx_seq);
+
+		if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
+			/* ensure remote has gotten their completions. */
+			ft_sync();
+			ret = ft_check_buf((char *) rx_buf + offset_rma_start,
+				   opts.transfer_size*valid_windows);
```
shefty, line 297: add space before/after *

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -267,28 +275,46 @@ int bandwidth(void)
...
+			if (ret) return ret;
```
shefty, line 298: Place return on its own line, though it doesn't look needed.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -267,28 +275,46 @@ int bandwidth(void)
...
+	} else { /* FT_RMA_READ */
 		ret = ft_get_tx_comp(tx_seq);
-	}
-	if (ret)
-		return ret;
 
+		if (ft_check_opts(FT_OPT_VERIFY_DATA)) {
+			ret = ft_check_buf((char *) tx_buf + offset_rma_start,
+				   opts.transfer_size*valid_windows);
+			if (ret) return ret;
```
shefty, line 306: return on own line, though it doesn't look needed

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -267,28 +275,46 @@ int bandwidth(void)
...
+		}
+
+	}
+	if (ret) return ret;
```
shefty, line 310: Drop if and just return ret;

fabtests/common/shared.c
```diff
@@ -538,6 +538,10 @@ int ft_alloc_msgs(void)
 			   MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size;
 	}
 
+	/* Allow enough space for RMA to operate in a distinct memory
+	   region that ft_sync() won't touch */
+	buf_size += 4 + MAX( ft_tx_prefix_size(), ft_rx_prefix_size() );
```
shefty, line 543: 4 is a magic number.  Can we use sizeof(whatever 4 is supposed to represent)?

fabtests/benchmarks/benchmark_shared.c
wckzhang: and remove return 0;

fabtests/include/shared.h
lrbison: hm, but the ft_post_rma function applies the same offset to both the local and remote addr, so I'd need to calculate that offset from the global buffer anyways.  After all these patches, post_rma always uses tx_buf as the pointer, so I guess that's not too bad, but as of this commit the post_rma function itself decided if it was going to use tx_buf or rx_buf, and so offset seemed more natural.

You would prefer I calculate remote addr based on the pointer-difference from tx_buf?

fabtests/benchmarks/benchmark_shared.c
lrbison: Thank you, I will fix the formatting.

I don't follow how it's not needed.  I'm trying to ensure ft_check_buf didn't find any errors, and it returns -FI_EIO on failure.

fabtests/include/shared.h
shefty: I'd like to start updating these internal calls to be more maintainable and obvious in how they are used. And since you're needing to change this call, it's a good time to update it.  A function signature of:
```
ft_post_rma(op, ep, buf, size, remote, context);
```
is more intuitive to use.  An offset parameter that applies to both the remote parameter and a global source buffer which may be the tx or may be the rx is confusing.

It's questionable if we should even keep this generic call, versus having separate calls, such as ft_post_write, ft_post_writedata, ft_post_read, and ft_post_write_inject.

lrbison: @shefty Please let me know what you think of the refactoring of ft_post_rma's arguments.  It keeps the arguments to post_rma simple, but it makes some assumptions, namely that every rma operation will occur within tx_buf/rx_buf, and the remote side should be in the opposite buffer (rx->tx, tx->rx).

These assumptions are guarded by an assert, and have been true previously, because the function chose tx_buf or rx_buf based on the RMA operation.

fabtests/benchmarks/benchmark_shared.c
lrbison: Fixed.

fabtests/common/shared.c
lrbison: I introduced #define FT_RMA_SYNC_MSG_BYTES in shared.h.

There is some funny business.  ft_sync() itself uses only 1 byte, and again I'm hesitant to change that, but ft_rma_comp() routines use 4 bytes, I think arbitrarily.  I've replaced the RMA completion functions that do 4-byte sends so they now use the new constant.

This patch uses both ft_sync and rma_comp, so the requirement is that this constant be >= the maximum of all those message sizes.

fabtests/common/shared.c
```diff
@@ -2060,25 +2060,60 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
...
 {
+	uint64_t addr = remote->addr + ft_rma_addr_from_buf( buf );
+
 	switch (op) {
 	case FT_RMA_WRITE:
 		FT_POST(fi_write, ft_progress, txcq, tx_seq, &tx_cq_cntr,
-			"fi_write", ep, tx_buf, size, mr_desc,
-			remote_fi_addr, remote->addr, remote->key, context);
+			"fi_write", ep, buf, size, mr_desc,
+			remote_fi_addr, addr, remote->key, context);
```
shefty, line 2115: I would replace the proposed ft_rma_addr_from_buf() with separate calls.  Example:
```
static size_t ft_write_offset(char *buf)
{
    assert(buf >= tx_buf && buf < (tx_buf + buf_size / 2));
    return buf - tx_buf;
}

/* Reads copy data from the remote tx_buf area into the local rx_buf area */
static size_t ft_read_offset(char *buf)
{
    assert(buf >= rx_buf && buf < (rx_buf + buf_size / 2));
    /* tx_buf area is the latter half of the remote region */
    return buf - rx_buf + buf_size / 2;
}
```

fabtests/common/shared.c
```diff
@@ -2060,25 +2060,60 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
...
 		break;
 	case FT_RMA_WRITEDATA:
 		FT_POST(fi_writedata, ft_progress, txcq, tx_seq, &tx_cq_cntr,
-			"fi_writedata", ep, tx_buf, size, mr_desc,
-			remote_cq_data, remote_fi_addr,	remote->addr,
+			"fi_writedata", ep, buf, size, mr_desc,
+			remote_cq_data, remote_fi_addr,	addr,
```
shefty, line 2120: See above.  Then here we can do: ```remote->addr + ft_write_offset(buf)```

fabtests/common/shared.c
```diff
@@ -2060,25 +2060,60 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
...
 			remote->key, context);
 		break;
 	case FT_RMA_READ:
 		FT_POST(fi_read, ft_progress, txcq, tx_seq, &tx_cq_cntr,
-			"fi_read", ep, rx_buf, size, mr_desc,
-			remote_fi_addr, remote->addr,remote->key, context);
+			"fi_read", ep, buf, size, mr_desc,
+			remote_fi_addr, addr, remote->key, context);
```
shefty, line 2126: similarly, ```remote->addr + ft_read_offset(buf)```

fabtests/common/shared.c
```diff
@@ -2060,25 +2060,60 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
...
+		offset = buf - rx_buf;
+		/* remote buf is within tx_buf */
+		return offset + buf_bytes;
+	}
+}
+
+ssize_t ft_post_rma(enum ft_rma_opcodes op, char *buf, size_t size,
 		struct fi_rma_iov *remote, void *context)
 {
+	uint64_t addr = remote->addr + ft_rma_addr_from_buf( buf );
```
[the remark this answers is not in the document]
lrbison: Thank you, corrected.

fabtests/common/shared.c
lrbison: Done.

lrbison:
Some real failures in AWS CI.  I will address today.

lrbison:
In the latest CI failure, I find that for EFA's emulated writes, I required another synchronization to ensure the write was delivered not just to the endpoint, but also to the application memory:

```
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -295,6 +295,9 @@ static int bw_rma_comp(enum ft_rma_opcodes rma_op, int valid_windows)
                        ret = rma_bw_rx_comp(valid_windows);
                }
        } else {
+               if (rma_op == FT_RMA_WRITE && ft_check_opts(FT_OPT_VERIFY_DATA))
+                       ft_sync();
+
                ret = ft_get_tx_comp(tx_seq);
                if (ret)
                        return ret;
```

I'm open to other approaches to ensure this.

shefty:
The extra sync is likely just adding enough time that the race condition isn't as likely to be hit.  With RMA writes, you need FI_DELIVERY_COMPLETE for a local completion to indicate that the peer now has access to the data.  There are 2 options here:

Either all RMA writes must use FI_DELIVERY_COMPLETE (in the case data validation is desired), or...

After the RMA writes, post a transfer with FI_DELIVERY_COMPLETE | FI_FENCE.  The transfer can be a write or send, but it needs to be both fenced and have delivery complete.  The fence requires that all previously posted transmits meet the same completion semantic as delivery complete.

wzamazon:
>The extra sync is likely just adding enough time that the race condition isn't as likely to be hit. With RMA writes, you need FI_DELIVERY_COMPLETE for a local completion to indicate that the peer now has access to the data. There are 2 options here:

I think the extra sync is needed because we have a pre-posted receive on the server side. so when server call `ft_sync` and get receive completion, it did not actually get the message from client, but a  previous message.

aingerson:
@wzamazon A lot of tests (especially bw style tests) purposefully consume the preposted buffer to make sure all the rest of the completions align with the actual testing. For example, see [bw.c](https://github.com/ofiwg/libfabric/blob/main/fabtests/functional/bw.c#L164-L174)

lrbison:
After talking to Wei I think the problem is the pre-posted rx buffer allows one side to proceed without confidence that the other has actually posted the write.  It seems that I could call `ft_sync(); ft_sync();` to ensure that both-sides have actually completed tx before proceeding past the sync.  I thought perhaps also I could more explicitly do this with the following:

```
                ret = ft_get_tx_comp(tx_seq);
		if (ret)
			return ret;

		if (rma_op == FT_RMA_WRITE && ft_check_opts(FT_OPT_VERIFY_DATA)) {
			if (!opts.dst_addr) {
				ret = bw_tx_comp();
				ret |= rma_bw_rx_comp();
				ret |= bw_tx_comp();
			} else {
				ret = rma_bw_rx_comp();
				ret |= bw_tx_comp();
				ret |= rma_bw_rx_comp();
			}
		}
```
Note that bw_rx_comp gets completion and then posts a tx(), and v.v. for bw_tx_comp.

but again I find that with repeated testing, eventually I get a validation failure, typically for one of the last writes.

@aingerson Thank your for the tip.  I gave a quick try to that but fabtests hang.  I might need to look into why if I don't find something else that makes sense.

@shefty I agree the sync before the completions doesn't make much sense.  We do turn on DELIVERY_COMPLETE semantics once the user requests validation.

I'm still scratching my head how both the `sync(); sync();` solution as well as the updated PR I posted still have occasional data integrity failures....

lrbison:
Another option is to execute an atomic or a write to indicate completion, but I'd like to understand why my current approach is invalid (and make sure I'm not actually finding a more serious bug in prov/efa)

shefty:
Ignore the test for a moment.  The question is how does the target of an RMA write operation know when the write data is available?  There are a few options for this.

1. If the initiator requests FI_DELIVERY_COMPLETE on the write.   Then once the sender gets a completion for the write, the data is available at the target.  If the sender sends a message to the target, once the target receives the message, the data is available.
2. The initiator posts the write, then posts a message, where the message uses FI_DELIVERY_COMPLETE **and** FI_FENCE.  Upon receiving the message, the target knows that the write data is visible.
3. The write generates a completion at the target, such as using writedata to write CQ data into a completion.  Once the target reads the completion, the write data is available.

shefty:
In the case of option 2 above, if FI_FENCE isn't specified, then the endpoint must have send-after-write (SAW) message and data ordering for the write data to be visible at the target at the time it processes a completion for the received message.

wzamazon:
I would prefer option 1. AFAIK,  this is how applications (like Open MPI's  OSC/RDMA) uses libfabric's `fi_write` API.

shefty:
IMO, option 1 is easy.  If data validation is requested, adjust the write completion semantics accordingly.

lrbison:
@shefty Yes, I've already made validation use [FI_DELIVERY_COMPLETE here](https://github.com/lrbison/libfabric/blob/rma_bw_verify/fabtests/benchmarks/rma_bw.c#L132), and yet I still get validation failures.

I think @aingerson and @wzamazon are right about the pre-posted buffer causing me problems, so I took Alexia's approach with [this commit](https://github.com/lrbison/libfabric/commit/76c3211dd69191ff76489db3b4f1fdbd76f6ca71), but I still get data validation errors.

My best solution so far is to do `ft_sync(); ft_syc()` everywhere, and then I'm not sure if it's because I'm hiding the errors behind timing delays or not.

I will pursue another task for a while and hope some good idea trickles back into my head.  I don't want to double-call sync.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -248,28 +266,57 @@ int bandwidth(void)
...
+/**
+ * @brief Get completions of RMA operations, and verify_data if requested
+ *
+ * Completions are received for the rma_op, according to the sequence counter.
+ * Verifications assume <valid_windows> many operations of opts.transfer_size
+ * were completed starting at buf + offset_rma_start.
+ *
+ * @return 0 on success.
+*/
+static int bw_rma_comp(enum ft_rma_opcodes rma_op, int valid_windows)
```
shefty, line 278: This call is the right place to handle data verification.  For RMA read operations, once all reads complete, we can just check that the data is correct.  For RMA writes, I see that you set FI_DELIVERY_COMPLETE below.  So, once the writes complete locally, the data should be visible at the peer.  We just need to send a message to the peer at this point for notification.  That is, the flow should be something like this:
```
complete all previously posted writes (opts.window_size completions)
send/recv sync message between peers (ft_sync)
verify write data
```

lrbison:
Fixed the HMEM-related CI errors.

I have been filling all data for the whole window at once, but for HMEM, the host-memory buffer used to stage data to/from HMEM was only big enough for a single message, not all messages for the window.

I've chosen to grow the tx_msg_buf to be the same size as tx_buf, rather than doing many ft_fill_buf() calls, and updated the commit message for `fabtests: bw_rma operates on distinct buffer offsets` to mention that.

lrbison:
I've fixed the synchronization errors.

The pre-posted rx_buf was a red herring.  The real cause of the problem was that we should not allow `ft_post_rma_inject` when we need FI_DELIVERY_COMPLETE semantics for data validation.

Now when data validation is requested, we silently change inject_size to 0 during the test.  Something similar already happens for FT_OPT_ENABLE_HMEM, so I don't think we need to print anything for the user.

With that change in place, I tested just regular-old ft_sync and everything works fine, so that's what went into the PR.

Thanks for all the eyes on this one!  Re-posting it as ready for review, I expect CI to pass (🤞).

Luke

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -294,26 +313,32 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 	inject_size = inject_size_set ?
 			hints->tx_attr->inject_size: fi->tx_attr->inject_size;
 
-	if (opts.options & FT_OPT_ENABLE_HMEM)
+	if (ft_check_opts(FT_OPT_ENABLE_HMEM) || ft_check_opts(FT_OPT_VERIFY_DATA))
```
wzamazon, line 316: Probably need to explain why inject_size was set to 0 when verify data is requested.

wzamazon: as comment

lrbison: done.

lrbison:
> I expect CI to pass.

I probably cursed myself, and was made to pay for it by casting a (-1) into a type `size_t`, which typically doesn't go well when filling a 0-sized buffer.

Updated and will wait for CI.

wenduwan:
bot:aws:retest

lrbison:
only remaining failure is appveyor.  it's the 
```- name:   rdm_tagged_peek -p tcp
  timestamp: Mon 04/24/2023 23:08:26.15
  result: Fail
  time:   183
  server_cmd: C:\projects\libfabric\fabtests\x64\Release-v142\rdm_tagged_peek -p tcp  -s 127.0.0.1
  server_stdout:
[error] fabtests:common\shared.c:2318: Tag mismatch!. Expected: 2748, actual: 1
fi_cq_read/fi_cq_sread(): functional\rdm_tagged_peek.c:107, ret=-256 (Unspecified error)
Receive sync(): functional\rdm_tagged_peek.c:334, ret=-256 (Unspecified error)
```
Ignored on main by 4aa3fa933fb92c520dda3a0dd34e756e602df89a anyways.

fabtests/common/shared.c
```diff
@@ -2060,25 +2060,52 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
 	return ret;
 }
 
+/**
+ * @brief for write operations, get remote addr offset
+ * @param[in] buf	local pointer within the tx_buf region.
+ * @return an offset to the corresponding region in the remote's rx_buf
+*/
+static size_t ft_write_offset(char *buf)
```
shefty, line 2078: Drop the comment -- it's just extra maintenance cost.  Declare buf as const char to indicate that it's input.  Rename to ft_remote_write_offset() to clarify what the function returns.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -267,23 +266,43 @@ int bandwidth(void)
...
+		if (rma_op == FT_RMA_WRITE && ft_check_opts(FT_OPT_VERIFY_DATA)) {
+			ft_sync();
+		}
 	}
-	if (ret)
+	if (ret || !ft_check_opts(FT_OPT_VERIFY_DATA))
 		return ret;
 
-	return 0;
+	ret = ft_check_buf(rx_buf + offset_rma_start,
```
shefty, line 303: The structure of the above code is confusing.  Having a check for error and returning early is acceptable.  e.g.
```
if (ret)
    return ret;
```
Note that we won't need this check if the code above is restructured.  The call to ft_check_buf() is dependent on ft_check_opts(VERIFY).  That's easier to understand when placed together.  E.g.
```
if (ft_check_opts(VERIFY))
    ret = ft_check_buf(...)

return ret;
```

fabtests/common/shared.c
```diff
@@ -2060,25 +2060,52 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
...
+	/* rx_buf area is the first half of the remote region */
+	return buf - tx_buf;
+}
+
+/**
+ * @brief for read operations, get remote addr offset
+ * @param[in] buf	local pointer within the rx_buf region.
+ * @return an offset to the corresponding region in the remote's tx_buf
+*/
+static size_t ft_read_offset(char *buf)
```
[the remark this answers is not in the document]
lrbison: done

fabtests/common/shared.c
lrbison: done.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -267,23 +266,43 @@ int bandwidth(void)
...
+/**
+ * @brief Get completions of RMA operations, and verify_data if requested
+ *
+ * Completions are received for the rma_op, according to the sequence counter.
+ * Verifications assume <valid_windows> many operations of opts.transfer_size
+ * were completed starting at buf + offset_rma_start.
+ *
+ * @return 0 on success.
+*/
+static int bw_rma_comp(enum ft_rma_opcodes rma_op, int valid_windows)
```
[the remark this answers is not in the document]
lrbison: done

fabtests/benchmarks/benchmark_shared.c
lrbison: Restructured.  Hopefully it is more clear now.

lrbison:
Updated to address comments, and rebased onto today's main

shefty:
thanks!

PR #8802, opened 2023-04-13
core/tostr: Remove the extra dir "shared/" from "include/" and "src/" .

peinanz:
To simplify the codes layout, rename include/shared/ofi_str.h to include/ofi_str.h and src/shared/ofi_str.c to src/ofi_str.c, so to remove the extra directory "shared" under "include" and "src". Update corresponding files which include "ofi_str.h".


shefty (approved):

shefty:
All ci failures are unrelated.

PR #8803, opened 2023-04-14
prov/efa: check return of ofi_iov_locate

shijin-aws:
Currently, rxr_pkt_init_data_from_op_entry calls ofi_iov_locate without checking the return code, which causes compilation error as tx_iov_index and tx_iov_offset could be uninitialized in the error path. This patch fixes this issue.


wzamazon (approved):

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -121,8 +121,13 @@ int rxr_pkt_init_data_from_op_entry(struct rxr_ep *ep,
 		return 0;
 	}
 
-	ofi_iov_locate(op_entry->iov, op_entry->iov_count, tx_data_offset,
-		       &tx_iov_index, &tx_iov_offset);
+	ret = ofi_iov_locate(op_entry->iov, op_entry->iov_count, tx_data_offset,
+			     &tx_iov_index, &tx_iov_offset);
+	if (ret) {
```
wzamazon, line 126: add `OFI_UNLIKELY`?

shijin-aws: Sure, there are some other places in efa call this function too, I can add it for them as well.

shijin-aws: Added

lrbison (approved):

PR #8805, opened 2023-04-14
prov/efa: skip() tests that require HAVE_EFADV_CQ_EX

lrbison:
Otherwise unittests can't compile.


wzamazon:
Pure EFA change, so ignored Intel CI

PR #8806, opened 2023-04-14
Makefile.am: workaround for am__is_gnu_make

lrbison:
In automake <1.15, am__is_gnu_make fails to execute when we have too many makefile fragments.  This has caused make check to fail for some time on such systems.

This fix replaces am__is_gnu_make with an alternative.

See: http://debbugs.gnu.org/cgi/bugreport.cgi?bug=18744
See: https://tracker.ceph.com/issues/14723


shefty (approved):
approving, but I try not to understand make as much as feasibly possible.  :)  (I did check the referenced bug reports and work-around posts)

lrbison:
> but I try not to understand make

I'm with you.  Doubly-so for automake and autoconf.

---

I updated this PR only because I noticed I left Makefile.am without a trailing newline.

lrbison:
AppVeyor error seems unrelated:

```
 name:   rdm_tagged_peek -p sockets
  timestamp: Sat 04/15/2023  2:46:39.18
  result: Fail
  time:   183
  server_cmd: C:\projects\libfabric\fabtests\x64\Release-Efa-v142\rdm_tagged_peek -p sockets  -s 127.0.0.1
  server_stdout:
[error] fabtests:common\shared.c:2310: Tag mismatch!. Expected: 2748, actual: 1
fi_cq_read/fi_cq_sread(): functional\rdm_tagged_peek.c:107, ret=-256 (Unspecified error)
Receive sync(): functional\rdm_tagged_peek.c:334, ret=-256 (Unspecified error)
Sending 10 tagged messages
Waiting for messages to complete
```

PR #8807, opened 2023-04-14
include/ofi_mem: smr_freestack_init enhancements

a-szegel:
Correctly initialize fs->free in smr_freestack_init(). 
Add new macro smr_freestack_isfull() to determine if freestack is full.

This is required for SM2 b/c it re-uses the SHM file, so we require that we initialize to 0 every time.

### DIFF
--- a/include/ofi_mem.h
+++ b/include/ofi_mem.h
@@ -226,6 +226,7 @@ struct smr_freestack {
 };
 
 #define smr_freestack_isempty(fs)	((fs)->top == SMR_FREESTACK_EMPTY)
+#define smr_freestack_isfull(fs)	((fs)->free == (fs)->size)
 
 static inline void* smr_freestack_get_entry_from_index(struct smr_freestack *fs,
 		int16_t index)
@@ -271,6 +272,7 @@ static inline void smr_freestack_init(struct smr_freestack *fs, size_t elem_coun
 	ssize_t i, next_aligned_addr;
 	assert(elem_count == roundup_power_of_two(elem_count));
 	fs->size = elem_count;
+	fs->free = 0;
 	fs->object_size = fs_object_size;
 	fs->top = SMR_FREESTACK_EMPTY;
 	fs->entry_base_offset =


aingerson (approved):

include/ofi_mem.h
```diff
@@ -226,6 +226,7 @@ struct smr_freestack {
 };
 
 #define smr_freestack_isempty(fs)	((fs)->top == SMR_FREESTACK_EMPTY)
+#define smr_freestack_isfull(fs)	((fs)->free == (fs)->size)
```
shefty, line 229: Can you add to the commit message why this is being added?

a-szegel: done

shefty: I was asking about the addition of the isfull function.  I understand the init change.  :)

a-szegel: fixed properly

shefty: I'm trying to understand the intent with isfull.  Are you saying that if process A exits, it can break the correct behavior of process B trying to communicate with process C?

a-szegel: If process A exits **WHILE** its free queue entries are still in processes B's recev queue, **THEN** process A Shared Memory Region can never be re-used, and it will turn into a zombie.

The way we clean up zombie processes is when every other PID in the file is DEAD, we shrink the file down to contain just the header (on exit... so we don't leave big files in peoples systems, and startup... startup is required in case everyone dies badly).

shefty: I don't know what a 'free queue entry' is.  If this is the naming Alexia is commenting on, then I agree it needs to change.  "Free" queue entries are, well, free, meaning not in use...

What process placed the entries in process' B's receive queue?  Is that done *only* by process A?

aingerson: @a-szegel While we're here, also a question about cleanup - what happens if A sent all of its entries to B and then B exits without processing them? Does A get starved? How do the entries go back to A?

a-szegel: yes, A will not be able to send anymore.

a-szegel: 1. All processes own their own "entries" for sending
2. Any process can place their entries in any other process's recv queue
 a. The recv queue is a linked list through different processes's Shared Memory Regions.
 b. All of the SMR's happen to exist in the same file (we are stateful), so we have to be careful about when we can reuse a region.

a-szegel: SM2's Nemesis concept can still work with 1 SMR file per processes if we can have globally unique ID's and some way to look up late arrivers

a-szegel:
@shefty @aingerson Can we please merge this so I can rebase it out of my other PR?

shefty:
I'm trying to understand the isfull change, because the reason for needing it looks like a highly suspect design choice.

shefty:
Merging.  The change by themselves look fine.  There can be possible use for a call like isfull(), but the argument for it sounds a lot like justification to leak memory in racy ways.

PR #8810, opened 2023-04-14
windows/appveyor: Remove 'efa' config

shefty:
The EFA config is the same as the normal debug/release configs. It's not actually EFA related, since the standard builds will also build the EFA provider.  (Actual testing of the EFA provider must be done on AWS).

This reduces CI runtime by half on windows.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,9 +6,7 @@ build:
 
 configuration:
   - Debug-v142
-  - Debug-Efa-v142
   - Release-v142
-  - Release-Efa-v142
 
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

shefty:
The appveyor changes are resulting in more frequent fails in the cm_data test.  The test is failing to start-up properly.  I suspect the timing windows are tighter now that the tests run back to back.

I'm working toward moving the CI testing from sockets to tcp provider and will adjust the test coverage as part of that change.

PR #8811, opened 2023-04-14
core/hmem: Deprecate `FI_HMEM_CUDA_ENABLE_XFER`

darrylabbate:
This removes the initialization logic related to `FI_HMEM_CUDA_ENABLE_XFER` for the `FI_HMEM` CUDA interface in favor of the endpoint option `FI_OPT_CUDA_API_PERMITTED`.

This also removes some facilities which are no longer needed, such as the `cuda_xfer_setting` enum and the dummy `cuda_disabled_cudaMemCpy()` function.

The environment variable itself doesn't seem to have been explicitly defined in any of the section 7 man pages, so no explicit deprecation notices or warning messages were added.

### DIFF
--- a/man/fabric.7.md
+++ b/man/fabric.7.md
@@ -300,23 +300,18 @@ portability across providers.
   on write restrictions.
 
 ## CUDA deadlock
-In some cases, calls to `cudaMemcpy` within libfabric may result in a
-deadlock. This typically occurs when a CUDA kernel blocks until a
-`cudaMemcpy` on the host completes.  To avoid this deadlock,
-`cudaMemcpy` may be disabled by setting
-`FI_HMEM_CUDA_ENABLE_XFER=0`. If this environment variable is set and
-there is a call to `cudaMemcpy` with libfabric, a warning will be
-emitted and no copy will occur. Note that not all providers support
-this option.
+In some cases, calls to `cudaMemcpy()` within libfabric may result in a deadlock.
+This typically occurs when a CUDA kernel blocks until a `cudaMemcpy` on the host
+completes.  Applications which can cause such behavior can restrict Libfabric's
+ability to invoke CUDA API operations with the endpoint option
+`FI_OPT_CUDA_API_PERMITTED`. See [`fi_endpoint`(3)](fi_endpoint.3.html) for more
+details.
 
 Another mechanism which can be used to avoid deadlock is Nvidia's
-gdrcopy. Using gdrcopy requires an external library and kernel module
-available at https://github.com/NVIDIA/gdrcopy. Libfabric must be
-configured with gdrcopy support using the `--with-gdrcopy` option, and
-be run with `FI_HMEM_CUDA_USE_GDRCOPY=1`. This may be used in
-conjunction with the above option to provide a method for copying
-to/from CUDA device memory when `cudaMemcpy` cannot be used. Again,
-this may not be supported by all providers.
+GDRCopy. Using GDRCopy requires an external library and kernel module available
+at https://github.com/NVIDIA/gdrcopy. Libfabric must be configured with GDRCopy
+support using the `--with-gdrcopy` option, and be run with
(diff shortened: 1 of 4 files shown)


man/fabric.7.md
```diff
@@ -300,23 +300,13 @@ portability across providers.
   on write restrictions.
 
 ## CUDA deadlock
-In some cases, calls to `cudaMemcpy` within libfabric may result in a
-deadlock. This typically occurs when a CUDA kernel blocks until a
-`cudaMemcpy` on the host completes.  To avoid this deadlock,
-`cudaMemcpy` may be disabled by setting
-`FI_HMEM_CUDA_ENABLE_XFER=0`. If this environment variable is set and
```
darrylabbate, line 307: This was the only reference I could find to `FI_HMEM_CUDA_ENABLE_XFER` in the man pages, but it doesn't feel appropriate to add a deprecation notice in this section.

wzamazon (approved):
M

man/fabric.7.md
```diff
@@ -300,23 +300,13 @@ portability across providers.
...
-conjunction with the above option to provide a method for copying
-to/from CUDA device memory when `cudaMemcpy` cannot be used. Again,
-this may not be supported by all providers.
+In some cases, calls to `cudaMemcpy` within libfabric may result in a deadlock.
+This typically occurs when a CUDA kernel blocks until a `cudaMemcpy` on the host
+completes.  One mechanism which can be used to avoid deadlock is Nvidia's
+GDRCopy. Using GDRCopy requires an external library and kernel module available
+at https://github.com/NVIDIA/gdrcopy. Libfabric must be configured with GDRCopy
+support using the `--with-gdrcopy` option, and be run with
+`FI_HMEM_CUDA_USE_GDRCOPY=1`. This may not be supported by all providers.
```
wzamazon, line 309: I think we should mention that if application have such behavior that may cause deadlock, it may use the FI_OPT_CUDA_API_PERMITTED to inform libfabric of such restriction.

src/hmem_cuda.c
```diff
@@ -606,8 +573,7 @@ int cuda_hmem_init(void)
 	 * - cudaMemcpy() is available
 	 */
 	cuda_attr.use_ipc =
-		(cuda_attr.p2p_access_supported || cuda_attr.device_count == 1) &&
-		(cuda_attr.xfer_setting != CUDA_XFER_DISABLED);
+		cuda_attr.p2p_access_supported || cuda_attr.device_count == 1;
```
shijin-aws, line 576: is `p2p_access_supported` already true when `cuda_attr.device_count == 1` ?

darrylabbate: No, `p2p_access_supported` is initialized to `false` and the P2P detection function returns early if the number of CUDA devices detected is $\leqslant 1$

shijin-aws (approved):

wzamazon:
bot:aws:retest

wzamazon:
@shefty 

Does Intel CI failure need to be looked into?

darrylabbate:
bot:aws:retest

man/fabric.7.md
shefty: Does configure enable gdrcopy, if present, without requiring a separate configure option?

shefty (approved):

man/fabric.7.md
wzamazon: > Does configure enable gdrcopy, if present, without requiring a separate configure option?


Yes.

wzamazon (approved):

man/fabric.7.md
shefty: Since these lines are changing, I would capture something to that effect.  E.g. ```...using the --with-gdrcopy option (by default, configure will automatically attempt to detect if gdrcopy is present and usable) ...```

wzamazon: Sorry I merged after your approval. 

@darrylabbate  Can you update the doc as @shefty  suggested?

PR #8817, opened 2023-04-17
prov/tcp: Hold domain lock while deregistering memory

sydidelot:
The domain lock must be held while deregistering a memory region, otherwise the underlying MR map might be corrupted if it is accessed by concurrent threads.


shefty (approved):
thanks!  Can you cherry-pick this for v1.18.x?

(intel CI failure is unrelated)

sydidelot:
Thanks for the quick review! The patch for 1.18.x can be found here: #8818

PR #8825, opened 2023-04-18
prov/efa: remove rxr_rm_tx/rx_cq_check()

wzamazon:
rxr_rm_cq_check() check whether util_cq is overflow, and set the RXR_RM_TX/RX_CQ_FULL flag.

If the flags are full, fi_send/read/write/atomic will return -FI_EAGAIN.

The code as is does not work, because when -FI_EAGAIN is returned, progress engine is not called so if the RXR_RM_TX/RX_CQ_FULL flag is ever set, an application will hang.

Meanwhile, the check itself is not necesary because util_cq can handle the situation that CQ is full
by writing overflow CQ entry.

Therefore, this patch remove the code from critical code path.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -738,8 +738,6 @@ void efa_rdm_rxe_report_completion(struct efa_rdm_ope *rxe)
 					       rxe->total_len -
 					       rxe->cq_entry.len);
 
-		rxr_rm_rx_cq_check(ep, rx_cq);
-
 		if (OFI_UNLIKELY(ret)) {
 			EFA_WARN(FI_LOG_CQ,
 				"Unable to write recv error cq: %s\n",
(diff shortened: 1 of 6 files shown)


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -738,8 +738,6 @@ void efa_rdm_rxe_report_completion(struct efa_rdm_ope *rxe)
 					       rxe->total_len -
 					       rxe->cq_entry.len);
 
-		rxr_rm_rx_cq_check(ep, rx_cq);
```
wenduwan, line 741: Did this actually cause a hang in application?

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_ope.c
shijin-aws: I did not see it cause a hang actually, potentially because application polls tx cq efficiently so that it is never full.

shijin-aws (approved):

shijin-aws:
bot:aws:retest

wzamazon:
bot:aws:retest

wzamazon:
bot:aws:retest

wzamazon:
AWS CI failure is from sockets provider, and is not related to this PR

PR #8826, opened 2023-04-19
prov/util: store hmem_data in ofi_mr_map

wenduwan:
Store hmem_data in ofi_mr_map such that it can be retrieved later

### DIFF
--- a/prov/util/src/util_mr_map.c
+++ b/prov/util/src/util_mr_map.c
@@ -257,9 +257,13 @@ void ofi_mr_update_attr(uint32_t user_version, uint64_t caps,
 	if (caps & FI_HMEM) {
 		cur_abi_attr->iface = user_attr->iface;
 		cur_abi_attr->device = user_attr->device;
+		cur_abi_attr->hmem_data = FI_VERSION_GE(user_version, FI_VERSION(1, 19))
+		                          ? user_attr->hmem_data
+		                          : NULL;
 	} else {
 		cur_abi_attr->iface = FI_HMEM_SYSTEM;
 		cur_abi_attr->device.reserved = 0;
+		cur_abi_attr->hmem_data = NULL;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


wenduwan:
Build failure is Azure mirror issue.

prov/util/src/util_mr_map.c
```diff
@@ -257,9 +257,11 @@ void ofi_mr_update_attr(uint32_t user_version, uint64_t caps,
 	if (caps & FI_HMEM) {
 		cur_abi_attr->iface = user_attr->iface;
 		cur_abi_attr->device = user_attr->device;
+		cur_abi_attr->hmem_data = user_attr->hmem_data;
```
wzamazon, line 260: I think you need to check version here. Only when user_version >= 1.19, can you access `user_attr->hmem_data`.

wenduwan: It that necessary(or does it work that way)? `hmem_data` is defined in `fi_domain.h` and we shouldn't expect a version mismatch between these files right?

shefty: You have to deal with compiled apps which may be against the older header.

wenduwan: I see. That makes sense. Good catch!

prov/util/src/util_mr_map.c
```diff
@@ -257,9 +257,13 @@ void ofi_mr_update_attr(uint32_t user_version, uint64_t caps,
 	if (caps & FI_HMEM) {
 		cur_abi_attr->iface = user_attr->iface;
 		cur_abi_attr->device = user_attr->device;
+		cur_abi_attr->hmem_data = FI_VERSION_GE(user_version, FI_VERSION(1, 19))
+		                          ? user_attr->hmem_data
+		                          : NULL;
```
wenduwan, line 262: I guess this indent is correct.

shefty (approved):

aingerson (approved):

prov/util/src/util_mr_map.c
wenduwan: @wzamazon Do you have other comments?

wzamazon (approved):

PR #8827, opened 2023-04-20
prov/efa: fix coverity identified bugs

wzamazon:
Fix various bugs identified by coverity

### DIFF
--- a/prov/efa/src/rdm/rxr_msg.h
+++ b/prov/efa/src/rdm/rxr_msg.h
@@ -55,7 +55,7 @@ void rxr_msg_update_peer_rxe(struct fi_peer_rx_entry *peer_rxe,
 	 * We will do the second way, see efa_rdm_srx_free_entry().
 	 */
 	peer_rxe->flags = (rxe->fi_flags & ~FI_MULTI_RECV);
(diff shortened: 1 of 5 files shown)


prov/efa/src/rdm/rxr_msg.h
```diff
@@ -55,7 +55,7 @@ void rxr_msg_update_peer_rxe(struct fi_peer_rx_entry *peer_rxe,
 	 * We will do the second way, see efa_rdm_srx_free_entry().
 	 */
 	peer_rxe->flags = (rxe->fi_flags & ~FI_MULTI_RECV);
-	if (rxe->desc && rxe->desc[0] && ((struct efa_mr *)rxe->desc[0])->shm_mr) {
```
wenduwan, line 58: Did coverity complain about`rxe->desc && rxe->desc[0]`

wzamazon: Yes `rxe->desc` is an array, so it will never be NULL. comparing it with NULL does not make sense.

wenduwan (approved):

prov/efa/src/rdm/rxr_msg.h
wenduwan: Never mind. I thought `rxe->desc` is equivalent to `rxe->desc[0]`

PR #8828, opened 2023-04-20
core/hmem: introduce gdrcopy from/to iov functions

wenduwan:
Introduce equivalent utilities to ofi_copy_from/to_hmem_iov

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -172,6 +172,12 @@ void cuda_gdrcopy_to_dev(uint64_t handle, void *dev,
 			 const void *host, size_t size);
 void cuda_gdrcopy_from_dev(uint64_t handle, void *host,
 			   const void *dev, size_t size);
+ssize_t ofi_gdrcopy_to_cuda_iov(uint64_t handle, const struct iovec *iov,
+                                size_t iov_count, uint64_t iov_offset,
+                                const void *host, size_t len);
+ssize_t ofi_gdrcopy_from_cuda_iov(uint64_t handle, void *host,
+                                  const struct iovec *iov, size_t iov_count,
+                                  uint64_t iov_offset, size_t len);
 int cuda_gdrcopy_hmem_init(void);
 int cuda_gdrcopy_hmem_cleanup(void);
 int cuda_gdrcopy_dev_register(struct fi_mr_attr *mr_attr, uint64_t *handle);
--- a/src/hmem_cuda_gdrcopy.c
+++ b/src/hmem_cuda_gdrcopy.c
@@ -37,6 +37,7 @@
 
 #include "ofi_hmem.h"
 #include "ofi.h"
+#include "ofi_iov.h"
 
 #if HAVE_GDRCOPY
 
@@ -292,6 +293,44 @@ void cuda_gdrcopy_from_dev(uint64_t handle, void *hostptr,
 			  GDRCOPY_FROM_DEVICE);
 }
 
+static ssize_t cuda_gdrcopy_iov_buf(uint64_t handle,
+                                    const struct iovec *iov, size_t iov_count,
+                                    uint64_t iov_offset, void *buf,
+                                    size_t size, enum gdrcopy_dir dir)
+{
+	uint64_t done = 0, len;
+	char *dev_buf;
+	size_t i;
+
+	for (i = 0; i < iov_count && size; i++) {
+		len = ofi_iov_bytes_to_copy(&iov[i], &size, &iov_offset, &dev_buf);
+
+		if (!len)
+			continue;
+
+		cuda_gdrcopy_impl(handle, dev_buf, (void *)((char *)buf + done), len, dir);
+
+		done += len;
+	}
+	return done;
+}
+
+ssize_t ofi_gdrcopy_to_cuda_iov(uint64_t handle, const struct iovec *iov,
+                                size_t iov_count, uint64_t iov_offset,
+                                const void *host, size_t len)
+{
+	return cuda_gdrcopy_iov_buf(handle, iov, iov_count, iov_offset,
+	                           (void *) host, len, GDRCOPY_TO_DEVICE);
+}
+
+ssize_t ofi_gdrcopy_from_cuda_iov(uint64_t handle, void *host,
+                                  const struct iovec *iov, size_t iov_count,
+                                  uint64_t iov_offset, size_t len)
+{
+	return cuda_gdrcopy_iov_buf(handle, iov, iov_count, iov_offset,
+	                            host, len, GDRCOPY_FROM_DEVICE);
+}
+
 int cuda_gdrcopy_dev_register(struct fi_mr_attr *mr_attr, uint64_t *handle)
 {
 	int err;
@@ -408,6 +447,20 @@ int cuda_gdrcopy_dev_register(struct fi_mr_attr *mr_attr, uint64_t *handle)
 int cuda_gdrcopy_dev_unregister(uint64_t handle)
 {
 	return FI_SUCCESS;
+}
+
+ssize_t ofi_gdrcopy_to_cuda_iov(uint64_t handle, const struct iovec *iov,
+                                size_t iov_count, uint64_t iov_offset,
+                                const void *host, size_t len)
+{
+	return -FI_ENOSYS;
+}
+
+ssize_t ofi_gdrcopy_from_cuda_iov(uint64_t handle, void *host,
+                                  const struct iovec *iov, size_t iov_count,
+                                  uint64_t iov_offset, size_t len)
+{
+	return -FI_ENOSYS;
 }
 
 #endif /* HAVE_GDRCOPY */


src/hmem_cuda_gdrcopy.c
```diff
@@ -292,6 +293,45 @@ void cuda_gdrcopy_from_dev(uint64_t handle, void *hostptr,
 			  GDRCOPY_FROM_DEVICE);
 }
 
+static ssize_t cuda_gdrcopy_iov_buf(uint64_t handle,
```
wzamazon, line 296: I am not sure about this. Can you extend `ofi_copy_to/from_hmem_iov` to cover gdrcopy?

wenduwan: That is possible but I want to get consensus from the community.

To extend `ofi_copy_to/from_hmem_iov` for gdrcopy, we need to:
1. Define the behavior in the cuda->cuda scenario - the caller could pass cuda buffers for both src and dest, but that is not allowed for gdrcopy. This is currently a segfault, and it is not trivial to make it graceful - we need a mechanism to detect that both src and dest are cuda buffers.
1. Pass `hmem_data` alongside `device` to e.g. `ofi_copy_to_hmem_iov`. That's very doable and I'm ok with that. But it requires changes in multiple place. Just want to make sure the community is equally happy before I do it.

So, my current plan is to lift gdrcopy from `ofi_copy_to/from_hmem_iov` path, meaning:
1. `ofi_copy_to/from_hmem_iov` will only use `cudaMemcpy`, which supports cuda->cuda.
2. The caller can opt to call `cuda_gdrcopy_to/from` only when they know what they are doing. This essentially makes gdrcopy an optional copy utility.

include/ofi_hmem.h
```diff
@@ -172,6 +172,12 @@ void cuda_gdrcopy_to_dev(uint64_t handle, void *dev,
 			 const void *host, size_t size);
 void cuda_gdrcopy_from_dev(uint64_t handle, void *host,
 			   const void *dev, size_t size);
+ssize_t cuda_gdrcopy_to_dev_iov(uint64_t handle, const struct iovec *iov,
+                                size_t iov_count, uint64_t iov_offset,
+                                const void *hostptr, size_t len);
```
shefty, line 177: nit: Other functions nearby just use 'host' as the parameter name.  'ptr' is implied by it being a void *.  Similar comment applies below.

src/hmem_cuda_gdrcopy.c
shefty: ofi_copy_to/from_hmem_iov() only copy between device and host memory.  Device-to-device or host-to-host copies are not supported.  The 'buf' parameter into those calls is always a host buffer.

Passing in hmem_data is more complicated.

Regardless, I think the intent is for the caller to request gdrcopy, which is better handled explicitly, versus calling an abstraction that hides its use.  That is, either hide gdrcopy and the decision to use it lower in the stack, or expose it and let the decision to use it be controlled higher up.  The current approach seems to follow the latter, so, I agree with @wenduwan that it makes sense to keep this path separate.

I don't like that we duplicate the implementation of ofi_copy_hmem_iov_buf() below, but it's honestly only about 3 lines of code that end up duplicated.

src/hmem_cuda_gdrcopy.c
```diff
@@ -292,6 +293,45 @@ void cuda_gdrcopy_from_dev(uint64_t handle, void *hostptr,
...
+                                    const struct iovec *iov, size_t iov_count,
+                                    uint64_t iov_offset, void *buf,
+                                    size_t size, bool to_device)
```
shefty, line 299: ofi_copy_hmem_iov_buf() uses ```int dir``` instead of bool to_device.  We can align with that and have the callers pass in GDRCOPY_TO/FROM_DEVICE directly, which ends up making the code easier to read  (Seeing GDRCOPY_TO_DEVICE is more descriptive than seeing 'true')

include/ofi_hmem.h
wenduwan: Good point. Addressed.

src/hmem_cuda_gdrcopy.c
wenduwan: Good idea. Addressed.

src/hmem_cuda_gdrcopy.c
```diff
@@ -292,6 +293,44 @@ void cuda_gdrcopy_from_dev(uint64_t handle, void *hostptr,
...
+			continue;
+
+		cuda_gdrcopy_impl(handle, dev_buf, (void *)((char *)buf + done), len, dir);
+
+		done += len;
+	}
+	return done;
+}
+
+ssize_t cuda_gdrcopy_to_dev_iov(uint64_t handle, const struct iovec *iov,
```
wzamazon, line 318: if we are exposing these functions to providers, I would like the adjust name to `ofi_gdrcopy_to/from_cuda_iov`

src/hmem_cuda_gdrcopy.c
wenduwan: > ofi_copy_to/from_hmem_iov() only copy between device and host memory. Device-to-device or host-to-host copies are not supported. The 'buf' parameter into those calls is always a host buffer.

Oh that was opposite to my observation. I see SHM is using those routines on ipc path: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_progress.c#L598-L607

But if we make the intention explicit, i.e. only D2H/H2D are supported, then I see gdrcopy can be nicely hidden inside it.

But still we need to pass around `hmem_data`. Trying to pick @shefty 's brain on:
- Do we want to enforce `ofi_copy_to/from_hmem_iov() only copy between device and host memory`?
- Mind talking more about `Passing in hmem_data is more complicated`? Coding wise it sounds simple, but verbose. What other complexity do you have in mind for this change?

src/hmem_cuda_gdrcopy.c
wenduwan: Good idea. Addressed.

shefty (approved):
These look fine to me

src/hmem_cuda_gdrcopy.c
```diff
@@ -410,4 +449,18 @@ int cuda_gdrcopy_dev_unregister(uint64_t handle)
 	return FI_SUCCESS;
 }
 
+ssize_t ofi_gdrcopy_to_cuda_iov(uint64_t handle, const struct iovec *iov,
+                                size_t iov_count, uint64_t iov_offset,
+								const void *hostptr, size_t len)
```
wckzhang, line 454: Formatting is off, also name has random changed from const void *host to const void *hostptr, can we keep it consistent?

src/hmem_cuda_gdrcopy.c
```diff
@@ -410,4 +449,18 @@ int cuda_gdrcopy_dev_unregister(uint64_t handle)
...
+{
+	return -FI_EOPNOTSUPP;
+}
+
+ssize_t ofi_gdrcopy_from_cuda_iov(uint64_t handle, void *hostptr,
```
wckzhang, line 459: Same as previous comment

src/hmem_cuda_gdrcopy.c
```diff
@@ -410,4 +449,18 @@ int cuda_gdrcopy_dev_unregister(uint64_t handle)
 	return FI_SUCCESS;
 }
 
+ssize_t ofi_gdrcopy_to_cuda_iov(uint64_t handle, const struct iovec *iov,
+                                size_t iov_count, uint64_t iov_offset,
+								const void *hostptr, size_t len)
+{
+	return -FI_EOPNOTSUPP;
```
wzamazon, line 456: I think we usually return `-FI_ENOSYS` for this

src/hmem_cuda_gdrcopy.c
```diff
@@ -292,6 +293,44 @@ void cuda_gdrcopy_from_dev(uint64_t handle, void *hostptr,
...
+			continue;
+
+		cuda_gdrcopy_impl(handle, dev_buf, (void *)((char *)buf + done), len, dir);
+
+		done += len;
+	}
+	return done;
+}
+
+ssize_t ofi_gdrcopy_to_cuda_iov(uint64_t handle, const struct iovec *iov,
```
wckzhang, line 318: The naming doesn't seem to align with how the rest of the functions are named, perhaps it should be called
```ofi_cuda_gdrcopy_to_iov```? This is a minor NIT, feel free to ignore it if you'd like.

**Review by wckzhang — CHANGES_REQUESTED** [review]

**Review by wzamazon — APPROVED** [review]
Minor comment

File: src/hmem_cuda_gdrcopy.c (hunk 7 above)
Comment (wenduwan) [reply src/hmem_cuda_gdrcopy.c]: Right. Addressed.

File: src/hmem_cuda_gdrcopy.c (hunk 5 above)
Comment (wenduwan) [reply src/hmem_cuda_gdrcopy.c]: Thanks. Addressed.

File: src/hmem_cuda_gdrcopy.c (hunk 8 above)
Comment (wenduwan) [reply src/hmem_cuda_gdrcopy.c:318]: The way I look at it is this symmetry
```
ofi_copy_from_hmem_iov
ofi_gdrcopy_from_cuda_iov
```

On the other hand `ofi_cuda_gdrcopy_to_iov` is vague on whether the iov is on host or cuda.

wckzhang (approved):

wenduwan:
@aingerson Do you see any issue/alternative solutions here?

shefty:
Appveyor failures are all some sort of timing related test start-up issue

aingerson (approved):
Looks good to me!

wenduwan:
AWS CI failed due to SSH connection issue not related to this PR.

wenduwan:
bot:aws:retest

wenduwan:
@shefty @wzamazon AWS CI passed. Good to merge?

PR #8829, opened 2023-04-20
core/ofi.h: Add helper functions to check alignment

shijin-aws:
Introduce two new helper functions:
    - ofi_is_size_multiple: check whether a size is a multiple
    of given integer.
    - ofi_is_addr_aligned: check whether an address is aligned
    to given alignment.

### DIFF
--- a/include/ofi.h
+++ b/include/ofi.h
@@ -326,6 +326,16 @@ static inline void *ofi_get_page_end(const void *addr, size_t page_size)
 			+ page_size, page_size) - 1);
 }
 
+static inline bool ofi_is_size_multiple(size_t size, size_t multiple)
+{
+	return ((size % multiple) == 0);
+}
+
+static inline bool ofi_is_addr_aligned(void *addr, size_t alignment)
+{
+	return ((((uintptr_t) addr) % alignment) == 0);
+}
+
 static inline size_t
 ofi_get_page_bytes(const void *addr, size_t len, size_t page_size)
 {


lrbison (approved):

include/ofi.h
```diff
@@ -326,6 +326,16 @@ static inline void *ofi_get_page_end(const void *addr, size_t page_size)
 			+ page_size, page_size) - 1);
 }
 
+static inline bool ofi_is_size_aligned(size_t size, size_t alignment)
+{
+	return ((size % alignment) == 0);
```
wzamazon, line 331: given alignment is always power of 2, maybe (size & (alignment -1) ==0) would be more efficient?

include/ofi.h
```diff
@@ -326,6 +326,16 @@ static inline void *ofi_get_page_end(const void *addr, size_t page_size)
...
+}
```
shefty, line 332: You're using the word 'aligned' to mean 'multiple'.  In that regard, the function name is confusing.  ofi_is_size_multiple() is clearer to the intent, or at least based on the operation that is being performed.

include/ofi.h
shijin-aws: I see all the size alignment checks in this file are using `%`. I wonder if there is a reason. @shefty

include/ofi.h
```diff
@@ -326,6 +326,16 @@ static inline void *ofi_get_page_end(const void *addr, size_t page_size)
...
+
+static inline bool ofi_is_addr_aligned(void *addr, size_t alignment)
+{
+	return (addr == ofi_get_page_start(addr, alignment));
+}
```
shefty, line 337: Because here you're using the word 'aligned' to mean something else.  The implementation is confusing me as to the intent.  ofi_get_page_start() takes a page_size as its second parameter.  'Alignment' suggests a byte alignment value, not a page size.  Couldn't this just be:
```
(((uintptr_t) addr) % alignment) == 0)
```

shefty (changes requested):

include/ofi.h
shijin-aws: Thx, updated.

shefty (approved):
Thanks - unless I'm messing up the math, this looks right (with more obvious naming).

include/ofi.h
shefty: Do we know that alignment and multiple values are always a value of 2?  We can enforce that through assert(), but it's currently not a requirement.

As long as the input values are either constants or use sizeof(), a smart compiler should be able to identify this and generate an optimal check.  Though, the use of sizeof() suggests that a power of 2 isn't likely.

PR #8832, opened 2023-04-21
ci/appveyor: Switch CI from sockets to tcp provider

shefty:
The tcp provider is actively supported.  Call the tcp provider. Update the runfabtests.cmd script to invoke only tcp supported tests.  (The cmd script does not process the exclude file.)


wenduwan (approved):
Thank you!

shijin-aws (approved):

PR #8833, opened 2023-04-21
include/ofi_atom.h: Set ofi_atom atomic->is_initalized in release builds

a-szegel:
prov/sm2 keeps a stateful SHM file which has persistent atomics int32's in it.  When we run a release build, and then we run a debug build, we fail the assert(atomic->is_initialized).


a-szegel:
There are 3 options as I see them:
1. Hack AWS CI to delete SHM file after every run
2. This PR 
3. Make SM2 not use ofi_atom atomics in the persistent SHM file

a-szegel:
removing atomic from shared memory header.

PR #8835, opened 2023-04-21
fabtests: new options for the fi-rdmabw-xe tests

j-xiong:


fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
```diff
@@ -1137,11 +1136,24 @@ int main(int argc, char *argv[])
...
+			if (batch <= 0) {
+				fprintf(stderr,
+					"Batch too small, adjusted to 1\n");
+				batch = 1;
+			} else if (batch > TX_DEPTH) {
+				fprintf(stderr,
+					"Batch too large, adjusted to %d\n",
+					TX_DEPTH);
+				batch = TX_DEPTH;
+			}
```
shefty, line 1174: Does this option serve any significant purpose for testing?

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
```diff
@@ -1205,7 +1224,13 @@ int main(int argc, char *argv[])
...
 			break;
 		case 'S':
-			msg_size = atoi(optarg);
+			msg_size = parse_size(optarg);
+			max_size = msg_size;
+			proxy_block = msg_size;
+			break;
+		case 'M':
+			max_size = parse_size(optarg);
+			proxy_block = max_size;
```
shefty, line 1233: I think all RDMA protocols have a 2GB message limit

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
j-xiong: Main purpose is to see how completions affect performance.

fabtests/component/dmabuf-rdma/fi-rdmabw-xe.c
j-xiong: I don't think we need to enforce that limit here. Just let the run fail if the specified size is too large.

shefty: The memory registration is limited as well, however.

j-xiong: It should also fail by itself. 

However, in a few places int the test, `int` is used for the size. That actually can overflow and cut off the test at 2GB. Let me make the change to see how the test behavior for really large sizes.

shefty: It's not obvious that changing the batch size can result in memory registration overflow.  I'd add a check with an error message to make the problem obvious, versus the user getting an error later for what looks like an unrelated problem.

j-xiong: I ran some tests. It appears that registration itself is fine, but the RDMA fails after 1GB.  I tried `-b 64 -M 1G -n 128` and it worked fine which means the entire 64G memory is registered & accessed correctly if the transfer size doesn't exceed 1GB. However, the run with `-b 1 -M 1073741825 -n 1` failed immediately with I/O error.

j-xiong: We could check if `(max_size * batch)` overflow `size_t` though.

shefty (approved):

j-xiong:
Intel CI failure is unrelated (random fi_rdm_tagged_peek fail with sockets provider).

PR #8849, opened 2023-04-24
prov/shm: add iface parameter to smr_select_proto

wenduwan:
Pass through hmem iface to proto selection logic.

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -634,13 +634,13 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 	return 0;
 }
 
-int smr_select_proto(bool use_ipc, bool cma_avail,
-		     uint32_t op, uint64_t total_len, uint64_t op_flags)
+int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
+                     uint32_t op, uint64_t total_len, uint64_t op_flags)
 {
 	if (op == ofi_op_read_req) {
 		if (use_ipc)
 			return smr_src_ipc;
-		if (cma_avail && FI_HMEM_SYSTEM)
+		if (cma_avail && FI_HMEM_SYSTEM == iface)
 			return smr_src_iov;
 		return smr_src_sar;
 	}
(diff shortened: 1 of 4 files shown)


prov/shm/src/smr_ep.c
```diff
@@ -634,13 +634,13 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
...
 
-int smr_select_proto(bool use_ipc, bool cma_avail,
-		     uint32_t op, uint64_t total_len, uint64_t op_flags)
+int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
+                     uint32_t op, uint64_t total_len, uint64_t op_flags)
 {
 	if (op == ofi_op_read_req) {
 		if (use_ipc)
 			return smr_src_ipc;
-		if (cma_avail && FI_HMEM_SYSTEM)
```
wenduwan, line 643: I guess `cma_avail && FI_HMEM_SYSTEM` does not make much sense. Are we trying to test `cma_avail && FI_HMEM_SYSTEM == iface` instead?

aingerson, line 643: Woah, that's a big miss. Can you add something in the commit message about how this fixes a bug where CMA would never be selected for a read?

aingerson (approved):
Code change looks good. Would add something to the commit message about fixing/reenabling CMA for reads. Can you backport this as well please? Thank you for catching!

prov/shm/src/smr_ep.c
aingerson: Yeah definitely a bug that this fixes. Thank you!

wenduwan: Thanks for confirming!

wenduwan: Thanks. Updated commit message. I will backport the change to 1.18.x branch.

Do we need it for 1.17.x as well?

aingerson: Well the only real use case of shm is efa so I'll let you decide how far back you want to go. The issue technically goes all the way back to 1.15.x

wenduwan: Cool then we only need to backport to 1.18.x

wzamazon (approved):

wenduwan:
bot:aws:retest

wzamazon:
Please back port

PR #8851, opened 2023-04-25
[v1.17.x] contrib/intel/jenkins: Update with files from main

j-xiong:
Old files don't work properly with current CI infrastructure. The difference between the branch and main is too big to cherry-pick.

The 'net' provider is no long tested since it has been renamed to 'tcp' in main.


j-xiong:
Change to Jenkinsfile won't take effect until merged. @shefty Any way to test this before merging? The updated files are exactly the same as main.

shefty:
@zachdworkin can force test the changes through some manual intervention

zachdworkin:
@shefty jenkins is testing them now. The jenkinsfile is replaced with the one in this PR and it won't run the upstream scripts in the contrib/intel/jenkins folder but will instead run this PR's scripts.

shefty:
Passed Intel CI

PR #8854, opened 2023-04-25
mr/cache: Fix fi_mr_info initialization

amirshehataornl:
In the ofi_mr_rbt_overlap() function an iovec pointer is passed in as the key. However, it's then passed to util_mr_find_overlap() and assigned to an fi_mr_info pointer. This assumes that only the iovec pointer is accessed. With the addition of the peer_id check that is no longer true.

To avoid this situation allocate an fi_mr_info block on the stack, initialize the block to 0, copy in the iovec info, and pass it to the call to ofi_rbmap_search()

### DIFF
--- a/prov/util/src/util_mr_cache.c
+++ b/prov/util/src/util_mr_cache.c
@@ -168,8 +168,11 @@ static struct ofi_mr_entry *ofi_mr_rbt_overlap(struct ofi_rbmap *tree,
 					       const struct iovec *key)
 {
 	struct ofi_rbnode *node;
+	struct ofi_mr_info info = {0};
 
-	node = ofi_rbmap_search(tree, (void *) key,
+	info.iov = *key;
+
+	node = ofi_rbmap_search(tree, (void *) &info,
 				util_mr_find_overlap);
 	if (!node)
 		return NULL;
@@ -398,6 +401,7 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 	pthread_mutex_lock(&mm_lock);
 	cache->search_cnt++;
 
+	info.peer_id = 0;
 	info.iov = *attr->mr_iov;
 	entry = ofi_mr_rbt_find(&cache->tree, &info);
 	if (!entry) {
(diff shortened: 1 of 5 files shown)


prov/util/src/util_mr_cache.c
```diff
@@ -168,8 +168,12 @@ static struct ofi_mr_entry *ofi_mr_rbt_overlap(struct ofi_rbmap *tree,
 					       const struct iovec *key)
 {
 	struct ofi_rbnode *node;
+	struct ofi_mr_info info;
```
wzamazon, line 171: I think `struct ofi_mr_info info = {0};` is probably better.

prov/util/src/util_mr_cache.c
```diff
@@ -391,6 +395,8 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 	struct ofi_mr_info info;
```
wzamazon, line 395: similarly, use `struct ofi_mr_info info = {0}` instead of `memset`

wzamazon (approved):

wzamazon:
@jtamzn  Can we verify the latest revision still address the issue?

prov/util/src/util_mr_cache.c
```diff
@@ -388,7 +391,7 @@ int ofi_mr_cache_search(struct ofi_mr_cache *cache, const struct ofi_mr_info *in
 struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
 				       const struct fi_mr_attr *attr)
 {
-	struct ofi_mr_info info;
+	struct ofi_mr_info info = {0};
```
shefty, line 394: ofi_mr_info is over 100 bytes in size, and cache lookups are in the fast path.  I get it's safer to always clear the entire structure, but I'd prefer that we initialize only the fields that are needed.  We can include setting iface for future protection, since ipc and device are only defined if iface != system.

shijin-aws (approved):

shefty (approved):

shijin-aws:
AWS CI failure is due to ssh connection issue, can be ignored.

wzamazon:
bot:aws:retest

shijin-aws:
> Tested on c5n with 4 nodes, working normally.

thanks, waiting on CI

PR #8855, opened 2023-04-25
common: fix the cast of epoll_t to int when HAVE_EPOLL is not defined

### LINKED ISSUE #8847: src/common.c: incompatible pointer to integer conversion (L1866)
Hi all,

I am unable to install the lib using clang 16.0.1 from sources, and it fails with the following error:
```
src/common.c:1866:9: error: incompatible pointer to integer conversion returning 'ofi_epoll_t' (aka 'struct ofi_pollfds *') from a function with result type 'int' [-Wint-conversion]
        return dynpoll->ep;
               ^~~~~~~~~~~
1 warning and 1 error generated.
```
when looking at the code, it corresponds to this chunk:
```c
static int
ofi_dynpoll_get_fd_epoll(struct ofi_dynpoll *dynpoll)
{
	assert(dynpoll->type == OFI_DYNPOLL_EPOLL);
	return dynpoll->ep;
}
```

EDIT: `gcc-12` works, it seems to be a clang-specific issue

I am happy to submit a PR but I don't know what is the intended solution here.
Thanks for your help and your time.

thomasgillis:
fixes #8847, obtained when compiling with clang-16 on a system not supporting epoll (no `HAVE_EPOLL`)

the logic behind `get_fd_epoll` is not clear to me when the system doesn't have `epoll`.
I guess it then becomes an invalid call (see the commit), but happy to change it to something else

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1863,7 +1863,11 @@ static int
 ofi_dynpoll_get_fd_epoll(struct ofi_dynpoll *dynpoll)
 {
 	assert(dynpoll->type == OFI_DYNPOLL_EPOLL);
+#ifdef HAVE_EPOLL
 	return dynpoll->ep;
+#else
+	return INVALID_SOCKET; /* unsupported if not HAVE_EPOLL */
+#endif
 }
 
 static int


shefty:
The updates to the functions look reasonable; however, I'm looking deeper into where those functions are call.  Because at first glance, something doesn't look correct, yet the code is working, and I need to figure out what I'm missing.

Please update your signed-off-by line to match the patch author listing.  The 'DCO' check needs to pass.

thomasgillis:
@shefty great thanks :-)
We have spent some time yesterday trying to figure it out with @raffenet but couldn't really understand the whole picture.
So very happy to read the summary of your investigation.

PS: just fixed the DCO thingy, sorry for that

shefty:
I can get the build warning with ``` configure --disable-epoll```.  I'm trying to figure out how the code works on windows, or without having actual epoll support.

shefty:
Tests with epoll disabled fail if the app requests blocking eq or cq read support.  That's what I would have expected.  A deeper fix is needed in the provider to handle this.

shefty:
Please see #8869.  I took a slightly different approach with the build warning fix, and the tcp provider needed updating.  I'm unsure if the tcp provider fix is complete, but it is at least passing fabtests with epoll disabled through configure.

thomasgillis:
Great, thx @shefty

PR #8857, opened 2023-04-25
[v1.16.x] contrib/intel/jenkins: Update with files from main

j-xiong:
Old files don't work properly with current CI infrastructure. The difference between the branch and main is too big to cherry-pick.

The v1.16.x branch doesn't support dsa. The jenkins scripts have some dsa related code but they are inactive.

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -2,12 +2,42 @@
 properties([disableConcurrentBuilds(abortPrevious: true)])
 def DO_RUN=1
 def TARGET="main"
-
+def SCRIPT_LOCATION="py_scripts/contrib/intel/jenkins"
+def RELEASE=0
 def check_target() {
   echo "CHANGE_TARGET = ${env.CHANGE_TARGET}"
   if (changeRequest()) {
     TARGET = env.CHANGE_TARGET
   }
+
+  if (TARGET) {
+    return TARGET
+  }
+
+  return "main"
+}
+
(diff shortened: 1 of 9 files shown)


j-xiong:
@shefty This is similar to #8851 
@zachdworkin Could you also force it to test with Jenkinsfile from the PR?

zachdworkin:
@j-xiong @shefty Yes, its going now.

aingerson:
@shefty It looks like the runmultinode.sh script is missing from Makefile.am

shefty:
The simplest option looks like removing references to multinode from the updated config files.

aingerson:
@shefty I think the simplest option would be to try to add the script to the makefile....

shefty:
Here's the commit that added multinode.sh to the Makefile.am.

https://github.com/ofiwg/libfabric/commit/8d3150b2ffb3274d4c5e30b3cc15e75b5cffe316

We would likely need to bring in the changes to multinode.sh as well.  Hopefully pulling in the lastest file would work.

zachdworkin:
@j-xiong do you need me to do the jenkins "magic" again?

j-xiong:
@zachdworkin Yes. Thanks!

j-xiong:
@zachdworkin Looks like the new "magic" runs didn't include the newly added commit (12ee75c).

j-xiong:
Tests passed. No need to wait for the ubuntu 18.04 builds (will backport more fixes to remove old ubuntu versions).

PR #8858, opened 2023-04-25
prov/verbs: Recover RXM connection from verbs QP in error state

chien-intel:
Verbs QP can transition to error state for number of reasons such as retransmission time out or protection error, rendering RXM connection unusable until QP is back in RTS state.  Since not all RDMA adapters can transition a QP from error state back to RTS, the solution is to destroy the QP and create a new one on subsequent use.

On a completion with error status, check current QP state (expensive) against error state. Write a FI_SHUTDOWN event to event queue for RXM to process and close the connection.  When a new request to use the connection, a new QP will be created.

### DIFF
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -253,6 +253,8 @@ int vrb_poll_cq(struct vrb_cq *cq, struct ibv_wc *wc)
 
 		ctx = (struct vrb_context *) (uintptr_t) wc->wr_id;
 		wc->wr_id = (uintptr_t) ctx->user_ctx;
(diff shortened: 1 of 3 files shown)


chien-intel:
v2 renamed vrb_check_qp_in_err to vrb_shutdown_qp_in_err.

shefty (approved):

shefty:
AWS CI failure should not be related to verbs only changes.  @wzamazon can you confirm please?  (I can't see the CI results directly now.)

shijin-aws:
@shefty it's some ssh connection issue in our test. Safe to ignore

PR #8859, opened 2023-04-26
fabtests/rdm_tagged_peek: fix race condition synchronization

aingerson:
It was possible for the client to complete the finalize send before the server gets the last 0xabc completion, causing a tag mismatch when processing the incorrect receive.

This removes the extra synchronization sends and adds an OOB sync to make sure all of the test sends have finished before quitting the test.

### DIFF
--- a/fabtests/functional/rdm_tagged_peek.c
+++ b/fabtests/functional/rdm_tagged_peek.c
@@ -309,34 +309,13 @@ static int run(void)
 		ret = do_recvs();
 		if (ret)
 			return ret;
-
-		/* sync with sender before ft_finalize, since we sent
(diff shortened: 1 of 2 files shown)


shijin-aws (approved):
Thanks

shijin-aws:
I think we can revert this then https://github.com/ofiwg/libfabric/commit/4aa3fa933fb92c520dda3a0dd34e756e602df89a. If you want to make it in a separate PR, I can do it.

aingerson:
@shijin-aws Oh good point. I'll add that here so we can test it in this PR.

shefty (approved):
CI failures were in rxd and DAOS (I think a node issue).  Re-started CI anyway.

PR #8864, opened 2023-04-26
prov/efa: Correct CPPFLAGS during configure

lrbison:
configure.m4 restored CPPFLAGS before it finished testing various feature support in rdma-core.  This meant feature testing occurred with system default includes rather than the user-specified include paths, resulting in incorrect configure results.

### DIFF
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -94,7 +94,6 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 			    [AC_DEFINE([HAVE_CAPS_RDMA_WRITE], [0], [EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE is not defined])],
 			    [[#include <infiniband/efadv.h>]])
 	      ])
-	CPPFLAGS=$save_CPPFLAGS
 
 	dnl Check for ibv_is_fork_initialized() in libibverbs
 	have_ibv_is_fork_initialized=0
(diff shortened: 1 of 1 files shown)


wzamazon (approved):

shijin-aws:
nit: `prov/efa/configure.m4` -> `prov/efa` in title for consistency

prov/efa/configure.m4
```diff
@@ -239,6 +238,8 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 
 	AM_CONDITIONAL([ENABLE_EFA_UNIT_TEST], [ test x"$enable_efa_unit_test" != x"" && test x"$enable_efa_unit_test" != x"no" ])
 	AM_CONDITIONAL([HAVE_NEURON], [ test "$have_neuron" = "1" ])
+
+	CPPFLAGS=$save_CPPFLAGS
```
bwbarrett, line 242: GitHub won't let me make a comment on line 189, but I think this undoing of CPPFLAGS should be near the storing of efa_CPPFLAGS (on line 189 right now).  Note that that block is actually probably in the wrong place, since line 219 modifies efa_LIBS.  So you should probably just pull lines 189-196 down here with the CPPFLAGS rollback.  I know it's not directly related to this patch, but might as well fix all the FLAGS handling in one commit?

lrbison: I noticed later that I think the early call of `AC_SUBST(cmocka_rpath)` is probably why fi_info is broken when we build with efa unit tests, but I hadn't noticed the efa_LIBS change.  Thank you.

lrbison:
looks like we need to have efa_LIBS and flags set before cmocka unit test block, this is because we append `-static` to the end of the efa_LIBs, and that has to remain on the end of the string.  I missed that last night and so normal builds are fine, but unit-test builds all broke.

Re-pushed this morning to keep the SUBST at the bottom, and the efa_LIBS and company just before unit-test block.

shijin-aws (approved):

bwbarrett (approved):

wzamazon:
Backport?

lrbison:
pending CI completion (failed on sunday due to a dependency change in pip for our testing framework)

https://github.com/ofiwg/libfabric/pull/8880

PR #8866, opened 2023-04-26
prov/efa: register shm MR with hmem_data

wenduwan:
When EFA provider registers MR with SHM endpoint, pass in the associated hmem_data via fi_mr_attr.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -214,6 +214,8 @@ static int efa_mr_hmem_setup(struct efa_mr *efa_mr,
 {
 	int err;
 
+	efa_mr->peer.flags = flags;
+
 	if (attr->iface == FI_HMEM_SYSTEM) {
 		efa_mr->peer.iface = FI_HMEM_SYSTEM;
 		return FI_SUCCESS;
@@ -922,14 +924,20 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 		*/
 		original_access = mr_attr->access;
 		mr_attr->access |= FI_REMOTE_READ;
+		/* Inherit peer.flags with addtional feature bits such as gdrcopy handle switch */
 		shm_flags = efa_mr->peer.flags;
 		if (mr_attr->iface != FI_HMEM_SYSTEM) {
 			/* shm provider need the flag to turn on IPC support */
 			shm_flags |= FI_HMEM_DEVICE_ONLY;
 		}
 
-		ret = fi_mr_regattr(efa_mr->domain->shm_domain, attr,
+		if (FI_VERSION_GE(efa_mr->domain->util_domain.fabric->fabric_fid.api_version, FI_VERSION(1, 19))) {
+			mr_attr->hmem_data = efa_mr->peer.hmem_data;
+		}
+
+		ret = fi_mr_regattr(efa_mr->domain->shm_domain, mr_attr,
 				    shm_flags, &efa_mr->shm_mr);
+
 		mr_attr->access = original_access;
 		if (ret) {
 			EFA_WARN(FI_LOG_MR,


prov/efa/src/efa_mr.c
```diff
@@ -856,14 +856,16 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 	efa_mr->mr_fid.mem_desc = NULL;
 	efa_mr->mr_fid.key = FI_KEY_NOTAVAIL;
 
-	ret = efa_mr_hmem_setup(efa_mr, mr_attr, flags);
+	memcpy(&mr_attr, attr, sizeof(struct fi_mr_attr));
+
```
shijin-aws, line 860: Can we avoid the memcpy by recovering the attr after registering shm mr?

shijin-aws: AFAICT, if we change *attr to attr in `static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)` , we can modify attr inside the function, and it can be recovered after the function call.

wenduwan: The reason I changed `mr_attr` from a pointer to a struct is to enforce `efa_mr_reg_impl(..., const void *attr)`, such that `attr` cannot be mutated. In such way we do not need to worry about restoring the value through the test.

wrt `memcpy`, I did not observe any performance difference but, yes, it will add to the overall cost. I'm ok with `memcpy` in this case since MR reg is mostly a one-time procedure. WDYT?

shijin-aws: >in this case since MR reg is mostly a one-time procedure

this may not be true, Ompi MTL/ofi reg MR every time before it do send, we have mr cache to reduce the cost, but efa_mr_reg_impl can still be called multiple times for buffers that do not overlaps.

wenduwan: I see. Then it makes sense to get rid of memcpy. Thanks!

prov/efa/src/efa_mr.c
```diff
@@ -845,6 +845,7 @@ static uint64_t efa_mr_cuda_non_p2p_keygen(void) {
 static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 {
 	uint64_t core_access, original_access;
+	void *hmem_data = NULL;
```
shijin-aws, line 848: i still think we can simply this by passing `attr` instead of a pointer. This function is an internal function called by `efa_mr_cache_entry_reg` and `efa_mr_regattr`, so we should be good to do it.

wenduwan: I'm not sure about that. Passing the struct will cause an implicit copy right?

shijin-aws: are other variables in `smr_env` configurable?

prov/efa/src/efa_mr.c
```diff
@@ -928,8 +929,17 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 			shm_flags |= FI_HMEM_DEVICE_ONLY;
 		}
 
-		ret = fi_mr_regattr(efa_mr->domain->shm_domain, attr,
+		if (FI_VERSION_GE(efa_mr->domain->util_domain.fabric->fabric_fid.api_version, FI_VERSION(1, 19))) {
+			hmem_data = mr_attr->hmem_data;
+			mr_attr->hmem_data = efa_mr->peer.hmem_data;
```
wzamazon, line 934: I do not understand why we will pass hmem_data from user. Don't you want to pass gdrcopy handle?

wenduwan: > I think we need to add the OFI_GDRCOPY_HANDLE flag

Yes. It is done on L926 `shm_flags = efa_mr->peer.flags`

wenduwan: On a second look. This is a good catch. We also need to respect the user flag. Will update.

wzamazon: I see. A comment would help here.

wenduwan: Thank you. Addressed.

wenduwan:
intel CI failure should be unrelated.

prov/efa/src/efa_mr.c
shijin-aws: Sorry, i did not follow, where is the `OFI_GDRCOPY_HANDLE` being set?

shijin-aws: i see, makes sense

shijin-aws (approved):

wenduwan:
@shefty Could you please let us know if the CI failure is related? Thanks!

shefty:
CI restarted, but we can't test EFA, so unrelated.  There were several failures, which may have been related to the CI aborting because the PR was updated a second time a couple minutes after being updated.

shijin-aws:
>@shefty Could you please let us know if the CI failure is related? Thanks!

We can safely ignore Intel CI failure for pure efa provider change.

prov/efa/src/efa_mr.c
```diff
@@ -922,14 +925,24 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
...
+		}
+
+		ret = fi_mr_regattr(efa_mr->domain->shm_domain, mr_attr,
 				    shm_flags, &efa_mr->shm_mr);
+
+		if (FI_VERSION_GE(efa_mr->domain->util_domain.fabric->fabric_fid.api_version, FI_VERSION(1, 19)))
```
wzamazon, line 943: I am not sure this branch and L936 are needed. Currently `hmem_data` is reserved for libfabric internal usage, we do no expect user to pass it in.

wenduwan: Cool. We don't have to deal with hmem_data for now. Removed in update.

wzamazon (approved):

wenduwan:
@wzamazon CI completed. Ready to merge.

PR #8867, opened 2023-04-27
prov/shm: no-op logic to toggle data transfer protocol for gdrcopy

wenduwan:
This patch introduces the no-op logic to switch data transfer protocol based on the availability of gdrcopy, and data size.

For small data sizes, gdrcopy offers better performance than cudaMemcpy. Based on experiments, we have selected 3kb as the threshold.

Note that, gdrcopy is claimed to offer asymmetric H2D/D2H performance, and for larger messages it could be slower than cudaMemcpy. However, here we make a simplification based on our experiments, and use only one threshold for both H2D/D2H copies.

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -635,8 +635,13 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 }
 
 int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
-                     uint32_t op, uint64_t total_len, uint64_t op_flags)
+                     bool gdrcopy_avail, uint32_t op, uint64_t total_len,
+                     uint64_t op_flags)
 {
+	/* TODO: Move gdrcopy check out of non-cuda fast paths */
+	if (gdrcopy_avail && total_len <= smr_env.max_gdrcopy_size)
+		return total_len <= SMR_MSG_DATA_LEN ? smr_src_inline : smr_src_inject;
+
 	if (op == ofi_op_read_req) {
 		if (use_ipc)
 			return smr_src_ipc;
--- a/prov/shm/src/smr_init.c
+++ b/prov/shm/src/smr_init.c
@@ -44,6 +44,7 @@ struct smr_env smr_env = {
 	.sar_threshold = SIZE_MAX,
 	.disable_cma = false,
 	.use_dsa_sar = false,
+	.max_gdrcopy_size = 3072,
 };
 
 static void smr_init_env(void)
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -73,6 +73,7 @@ struct smr_env {
 	size_t sar_threshold;
 	int disable_cma;
 	int use_dsa_sar;
+	size_t max_gdrcopy_size;
 };
 
 extern struct smr_env smr_env;
@@ -330,8 +331,9 @@ size_t smr_copy_from_sar(struct smr_freestack *sar_pool, struct smr_resp *resp,
 			 struct smr_cmd *cmd, struct ofi_mr **mr,
 			 const struct iovec *iov, size_t count,
 			 size_t *bytes_done, int *next);
-int smr_select_proto(enum fi_hmem_iface, bool use_ipc, bool cma_avail, uint32_t op,
-                     uint64_t total_len, uint64_t op_flags);
+int smr_select_proto(enum fi_hmem_iface, bool use_ipc, bool cma_avail,
+                     bool gdrcopy_avail, uint32_t op, uint64_t total_len,
+                     uint64_t op_flags);
 typedef ssize_t (*smr_proto_func)(struct smr_ep *ep, struct smr_region *peer_smr,
 		int64_t id, int64_t peer_id, uint32_t op, uint64_t tag,
 		uint64_t data, uint64_t op_flags, struct ofi_mr **desc,
--- a/prov/shm/src/smr_msg.c
+++ b/prov/shm/src/smr_msg.c
@@ -280,7 +280,7 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 	int64_t id, peer_id;
 	ssize_t ret = 0;
 	size_t total_len;
-	bool use_ipc = false;
+	bool use_ipc = false, gdrcopy_available = false;
 	int proto;
 	struct smr_cmd_entry *ce;
 	int64_t pos;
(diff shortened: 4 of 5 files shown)


shefty (approved):
I'm assuming there's a separate patch where gdrcopy_avail isn't always set to false.

wenduwan:
> I'm assuming there's a separate patch where gdrcopy_avail isn't always set to false.

Yes. I will post follow-up PRs tomorrow.

prov/shm/src/smr_init.c
```diff
@@ -44,6 +44,7 @@ struct smr_env smr_env = {
 	.sar_threshold = SIZE_MAX,
 	.disable_cma = false,
 	.use_dsa_sar = false,
+	.max_gdrcopy_size = 3072,
```
shijin-aws, line 47: Can it be overridden by  `fi_param_get`  ?

wenduwan: Currently not. The configuration is not exposed  yet.

shijin-aws: Yeah, I am asking if there is a follow-up PR on it.

wenduwan: I see. I don't plan to make it configurable at this point. My experiments are based on p4d only and I don't see supporting data to change it at this point.

prov/shm/src/smr_ep.c
```diff
@@ -635,8 +635,16 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 }
 
 int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
-                     uint32_t op, uint64_t total_len, uint64_t op_flags)
+                     bool gdrcopy_avail, uint32_t op, uint64_t total_len,
+                     uint64_t op_flags)
 {
+	if (gdrcopy_avail) {
```
shijin-aws, line 641: I think this can be reformatted as

```
if (gdrcopy_avail and total_len <= smr_env.max_gdrcopy_size)
    return total_len <= SMR_MSG_DATA_LEN? smr_src_inline : smr_src_inject;
``` 

so gdrcopy related variable can be put in the same line.

wenduwan: Cool thanks! Will update.

prov/shm/src/smr_init.c
shijin-aws: are other variables in smr_env configurable?

wenduwan: Yes. I don't think that's necessary though.

prov/shm/src/smr_ep.c
```diff
@@ -635,8 +635,12 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 }
 
 int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
-                     uint32_t op, uint64_t total_len, uint64_t op_flags)
+                     bool gdrcopy_avail, uint32_t op, uint64_t total_len,
+                     uint64_t op_flags)
 {
+	if (gdrcopy_avail && total_len <= smr_env.max_gdrcopy_size)
```
aingerson, line 641: Not sure what the end vision is for this but can we just initialize max_grdcopy_size to either 3072 or 0 whether gdrcopy is available or not? Is gdrcopy_avail a global setting or per transfer? 
Eventually I think we should think about reordering this proto selection code to optimize for the fast path

wenduwan: My current solution is to have a static threshold 3072, plus a flag `OFI_HMEM_DATA_GDRCOPY_HANDLE` that's passed from EFA to `fi_mr_regattr`.

In e.g. `smr_generic_sendmsg` we will inspect `smr_desc = (struct ofi_mr *) *desc;` and set `gdrcopy_avail` if the flag is set.

So overall the idea is to separate 1) whether gdrcopy handle is available(based on flag) from 2) when gdrcopy is preferred over cudaMemcpy(based on smr env).

> Eventually I think we should think about reordering this proto selection code to optimize for the fast path

Yes. I agree. With this change I do see that we add an additional check `gdrcopy_avail` to all paths. Let me think about this for a min.

wenduwan: @aingerson It took me more than 1min. Sadly it eludes me how to do it cleanly... I messed around the code and ended up with plugging in `gdrcopy_avail` left and right.

That said, I tried to understand the impact of this change but so far it appears benign 🫤

|  pt2pt/osu_latency  |                   |                   |
|---------------------|-------------------|-------------------|
|                Size |           Before  |            After  |
|                   0 |             0.95  |             0.93  |
|                   1 |             0.97  |             0.93  |
|                   2 |             0.96  |             0.91  |
|                   4 |             0.96  |             0.92  |
|                   8 |             0.95  |             0.92  |
|                  16 |             0.95  |             0.92  |
|                  32 |             0.96  |             0.94  |
|                  64 |             0.97  |             0.93  |
|                 128 |             1.02  |             0.97  |
|                 256 |             1.16  |             1.13  |
|                 512 |             1.21  |             1.17  |
|                1024 |             1.30  |             1.27  |
|                2048 |             1.47  |             1.44  |
|                2560 |             1.57  |             1.53  |
|                3072 |             1.65  |             1.62  |
|                3584 |             1.70  |             1.69  |
|                4096 |             1.85  |             1.82  |
|                8192 |             2.73  |             2.68  |
|               16384 |             4.02  |             4.00  |
|               32768 |             6.05  |             6.01  |
|               65536 |             9.94  |             9.96  |
|              131072 |            18.48  |            18.47  |
|              262144 |            35.04  |            34.80  |
|              524288 |            80.30  |            79.97  |
|             1048576 |           171.43  |           172.10  |
|             2097152 |           338.99  |           339.70  |

prov/shm/src/smr_init.c
shijin-aws: I think we should make it configurable too. This allows benefits for future experiments without the need of rebuilding the code. Like all the variables in rxr_env.

prov/shm/src/smr_ep.c
aingerson: No worries. Thanks for taking a look! We can just put it on a future todo to see if we can optimize this.

aingerson (approved):

shijin-aws:
> I think we should make it configurable too. This allows benefits for future experiments without the need of rebuilding the code. Like all the variables in rxr_env.

Bump this comment

prov/shm/src/smr_init.c
wenduwan: Actually we do have variables in rxr_env that are not configurable from CLI.

Introducing the environment variable can be left to the future IMO. It is easy to add new variables than removing old ones.

prov/shm/src/smr_ep.c
wenduwan: Good idea. Added a TODO comment.

shijin-aws (approved):

wenduwan:
AWS CI failure is unrelated to this change.

wenduwan:
@aingerson This should be ready for merge.

PR #8869, opened 2023-04-27
prov/tcp: Handle emulated epoll support

shefty:
We can't get an epoll fd if we don't have real support for epoll.  Update the epoll abstraction and tcp provider to handle the emulated epoll case.

### DIFF
--- a/include/ofi_epoll.h
+++ b/include/ofi_epoll.h
@@ -182,6 +182,11 @@ static inline void ofi_epoll_close(int ep)
 	close(ep);
 }
 
+static inline int ofi_epoll_fd(int ep)
+{
+	return ep;
+}
+
 #else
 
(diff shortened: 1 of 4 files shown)


thomasgillis:
this PR addresses #8847 
thanks @shefty

shefty:
bot:aws:retest

shefty:
Can someone from AWS report what the CI failure is?

shefty:
bot:aws:retest

raffenet:
@shefty this should be backported to v1.18.x. There is a small conflict when cherry-picking that I'm not confident how to fix, so your help would be appreciated.

shefty:
@raffenet - I'll cherry pick.  It's not needed prior to v1.18, as that was using the older tcp implementation.

PR #8873, opened 2023-04-27
prov/sm2: Use FI_NAME_MAX for names

a-szegel:
Remove `#define SM2_NAME_MAX 256` and use FI_NAME_MAX for names.  SM2 doesn't care how long names are... and they should be less than FI_NAME_MAX.


lrbison (approved):

shefty (approved):

a-szegel:
verified no performance hit

aingerson (commented):
Same here. Thanks!

PR #8874, opened 2023-04-27
prov/shm: Use FI_NAME_MAX for name length

a-szegel:
SMR_NAME_LENGTH is 256 b/c the max file size is 256.  We should be using FI_NAME_MAX b/c a user should be able to use any endpoint with their app.  If a user is specifying endpoint names that are longer than FI_NAME_MAX (64)... then they will not be able to take advantage of using other providers (the entire point of having a framework :)).

Also... the user doesn't have a good way of knowing that SHM has a SMR_NAME_LENGTH of 256... the only way they can find out is by trying to use fi_setname and providing a buffer that is too big.

### DIFF
--- a/include/ofi_shm.h
+++ b/include/ofi_shm.h
@@ -164,12 +164,11 @@ struct smr_cmd {
 #define SMR_SAR_SIZE		32768
 
 #define SMR_DIR "/dev/shm/"
-#define SMR_NAME_MAX	256
-#define SMR_PATH_MAX	(SMR_NAME_MAX + sizeof(SMR_DIR))
+#define SMR_PATH_MAX	(FI_NAME_MAX + sizeof(SMR_DIR))
 #define SMR_SOCK_NAME_MAX sizeof(((struct sockaddr_un *)0)->sun_path)
 
 struct smr_addr {
-	char		name[SMR_NAME_MAX];
+	char		name[FI_NAME_MAX];
 	int64_t		id;
 };
 
@@ -187,7 +186,7 @@ extern pthread_mutex_t sock_list_lock;
 struct smr_region;
 
 struct smr_ep_name {
-	char name[SMR_NAME_MAX];
+	char name[FI_NAME_MAX];
 	struct smr_region *region;
 	struct dlist_entry entry;
 };
--- a/include/ofi_shm.h
+++ b/include/ofi_shm.h
@@ -164,7 +164,6 @@ struct smr_cmd {
 #define SMR_SAR_SIZE		32768
 
 #define SMR_DIR "/dev/shm/"
-#define SMR_PATH_MAX	(FI_NAME_MAX + sizeof(SMR_DIR))
 #define SMR_SOCK_NAME_MAX sizeof(((struct sockaddr_un *)0)->sun_path)
 
 struct smr_addr {
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -298,7 +298,7 @@ static inline struct fid_peer_srx *smr_get_peer_srx(struct smr_ep *ep)
 static inline int smr_mmap_name(char *shm_name, const char *ep_name,
 				uint64_t msg_id)
 {
-	return snprintf(shm_name, SMR_NAME_MAX - 1, "%s_%ld",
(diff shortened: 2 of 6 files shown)


a-szegel:
The FI_NAME_MAX has a performance hit in osu_alltoallw.... I have no idea why that is happening.

lrbison:
The name length was a major driver for the SM2 allocation table size.  I wonder if the performance change would go away if you pad out the header to be the same size as it was before (I'm not sure why that would matter though....)

a-szegel:
Yes... this does indicate that we have some padding issues in SHM (this PR) / SM2.

a-szegel:
I think this PR should get merged, and another PR should be made to fix padding issues.

a-szegel:
To elaborate more on why this is required... I want to change the OMPI behavior of creating the contiguous addr array to be created with: `char* addr = malloc(FI_NAME_MAX * num_eps);`... right now the behavior is bugged (new PR coming for that and I will tag you guys in it).

a-szegel:
FYI:
https://github.com/open-mpi/ompi/pull/11629
https://github.com/open-mpi/ompi/pull/11630

shefty:
FI_NAME_MAX is not documented as meaning anything.  The define itself never should have been in the public header file.  It's used as a max node name, service name, and file name.

a-szegel:
So the problem is... how should OMPI know the max name size for every endpoint it needs to call fi_av_insert() on?

shefty:
There is no defined maximum size that all providers need to adhere to.  The app can arbitrarily decide to use its own max and fail to work with providers that have addresses larger than that value.

sockaddr_storage is defined as 128 bytes.  I would expect most addresses to fit into that size.  However, if a provider is using multi-rail, plus supporting shared memory, the size of its address could end up being significant, depending on the implementation.

The ucx provider has its max name size set to 1024.

a-szegel:
I still think we have a problem which is that there is no good way for the provider to communicate their max name length to the app.

wzamazon:
>I still think we have a problem which is that there is no good way for the provider to communicate their max name length to the app.

There is. `fi_getname` will set `addrlen` to correct length regardless whether given buffer is bigger enough.
so app can get the correct name line through `fi_getname`

aingerson:
@wzamazon I think Seth is talking about an absolute max across the entire provider. In shm there is no consistent address length like in other providers. Every endpoint can have a different name length.
There needs to be a standard for the app to figure out how much memory it needs to allocate per EP name. I see a few different solutions:
1) Force shm to adhere to the consistent addrlen for all EPs - ie even if the address is just one character and the rest of the address is NULL characters, the whole address is still 64 bytes
2) Expose the address length in some sort of magical variable that the app can check (what Seth is proposing)
3) Make the app figure it out by using a random large value and/or iterating over EPs
@shefty Do you have any opinions?

shefty:
There is a standard for apps to determine the amount of memory needed for each ep name - fi_getname().  What is being requested is that all providers enforce some sort of arbitrary, hard-coded maximum address length because apps are too lazy to handle this dynamically.  Why doesn't the app enforce its own arbitrary, hard-coded maximum size and reject any provider that wants to use a larger size?

aingerson:
@shefty Man, car troubles got you SASSY today. You can just say you prefer option 3 :)

a-szegel:
@shefty @aingerson So we don't want to merge this PR until we figure out the regression correct?... But I should still leave it open, b/c it is something that we do want to do.

aingerson:
@a-szegel Sean is saying that it is up to the app to figure out how much memory it needs and adjust accordingly. I'm fine with using 64 instead of 256 but if the reason is because you want to use FI_NAME_MAX in OMPI then that's not the correct solution. Sean is saying FI_NAME_MAX should never have been put in the exposed headers in the first place and that the app needs to call getname on each endpoint to figure out how long each one is.

a-szegel:
FYI... this regression goes away when we use my freestack alignment patch: https://github.com/ofiwg/libfabric/pull/8895

a-szegel:
I think we should:
1. Merge the alignment patch
2. Merge this
3. fix the OMPI bug patch to calculate the total namelength as Sean suggested

shefty:
Note that this change has the potential to break apps that set their own name, if that name is longer than 64 bytes.

PR #8875, opened 2023-04-28
[v1.17.x] prov/verbs: Recover RXM connection from verbs QP in error state

chien-intel:
Verbs QP can transition to error state for number of reasons such as retransmission time out or protection error, rendering RXM connection unusable until QP is back in RTS state.  Since not all RDMA adapters can transition a QP from error state back to RTS, the solution is to destroy the QP and create a new one on subsequent use.

On a completion with error status, check current QP state (expensive) against error state. Write a FI_SHUTDOWN event to event queue for RXM to process and close the connection.  When a new request to use the connection, a new QP will be created.

### DIFF
--- a/prov/verbs/include/fi_verbs.h
+++ b/prov/verbs/include/fi_verbs.h
@@ -926,6 +926,7 @@ do {								\
 	( wr->opcode == IBV_WR_SEND || wr->opcode == IBV_WR_SEND_WITH_IMM	\
 	|| wr->opcode == IBV_WR_RDMA_WRITE_WITH_IMM )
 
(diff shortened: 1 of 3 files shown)


chien-intel:
Appveyor failure is not related to this patch.  Don't know about Amazon CI failure.

  server_cmd: C:\projects\libfabric\fabtests\x64\Debug-v142\scalable_ep -p **sockets**  -s 127.0.0.1
  server_stdout:
wait for recv completion for ctx: 0
failed compare expected 0x12345670, read 0x12345671

shefty:
sockets provider does not use rxm, so unrelated.

chien-intel:
bot:aws:retest

PR #8878, opened 2023-04-28
prov/sm2: Add HMEM Support to SM2 Inject Protocol

a-szegel:
EFA is not picking up SM2 b/c it does not support FI_HMEM.  OMPI attempts to create a libfabric provider with HMEM support, and if it fails, it tries to create one without HMEM support.  EFA is choosing to not hard-fail if SHM provider cannot be picked up, so when we specify SM2 for EFA, we run without SHM.

This patch adds support for HMEM, and allows EFA to pick up SM2.

### DIFF
--- a/fabtests/pytest/sm2/conftest.py
+++ b/fabtests/pytest/sm2/conftest.py
@@ -1,7 +1,12 @@
 import pytest
 
-# TODO Restore GPU Memory Types
-@pytest.fixture(scope="module", params=["host_to_host"])
+@pytest.fixture(scope="module", params=["host_to_host",
+                                        pytest.param("host_to_cuda", marks=pytest.mark.cuda_memory),
+                                        pytest.param("cuda_to_host", marks=pytest.mark.cuda_memory),
+                                        pytest.param("cuda_to_cuda", marks=pytest.mark.cuda_memory),
+                                        pytest.param("neuron_to_neuron", marks=pytest.mark.neuron_memory),
+                                        pytest.param("neuron_to_host", marks=pytest.mark.neuron_memory),
+                                        pytest.param("host_to_neuron", marks=pytest.mark.neuron_memory)])
 def memory_type(request):
     return request.param
 
--- a/prov/sm2/src/sm2_msg.c
+++ b/prov/sm2/src/sm2_msg.c
@@ -295,6 +295,7 @@ static ssize_t sm2_generic_sendmsg(struct sm2_ep *ep, const struct iovec *iov,
 	sm2_gid_t peer_gid;
 	ssize_t ret = 0;
 	size_t total_len;
+	struct ofi_mr **mr = (struct ofi_mr **) desc;
 
 	assert(iov_count <= SM2_IOV_LIMIT);
 
@@ -310,8 +311,8 @@ static ssize_t sm2_generic_sendmsg(struct sm2_ep *ep, const struct iovec *iov,
 	assert(!(op_flags & FI_INJECT) || total_len <= SM2_INJECT_SIZE);
 
 	ret = sm2_proto_ops[sm2_proto_inject](ep, peer_smr, peer_gid, op, tag,
-					      data, op_flags, FI_HMEM_SYSTEM, 0,
-					      iov, iov_count, total_len);
+					      data, op_flags, mr, iov,
+					      iov_count, total_len);
 	if (ret)
 		goto unlock_cq;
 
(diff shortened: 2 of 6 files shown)


a-szegel:
@aingerson @shefty Can this get a review, I would like to merge this simple patch today.

fabtests/pytest/sm2/conftest.py
```diff
@@ -1,11 +1,16 @@
...
 
-# TODO Restore GPU Memory Types
-@pytest.fixture(scope="module", params=["host_to_host"])
+@pytest.fixture(scope="module", params=["host_to_host",
+                                        pytest.param("host_to_cuda", marks=pytest.mark.cuda_memory),
+                                        pytest.param("cuda_to_host", marks=pytest.mark.cuda_memory),
+                                        pytest.param("cuda_to_cuda", marks=pytest.mark.cuda_memory),
+                                        pytest.param("neuron_to_neuron", marks=pytest.mark.neuron_memory),
+                                        pytest.param("neuron_to_host", marks=pytest.mark.neuron_memory),
+                                        pytest.param("host_to_neuron", marks=pytest.mark.neuron_memory)])
```
aingerson, line 9: All test stuff should be in a separate patch

prov/sm2/src/sm2_msg.c
```diff
@@ -309,9 +309,9 @@ static ssize_t sm2_generic_sendmsg(struct sm2_ep *ep, const struct iovec *iov,
 	total_len = ofi_total_iov_len(iov, iov_count);
 	assert(!(op_flags & FI_INJECT) || total_len <= SM2_INJECT_SIZE);
 
-	ret = sm2_proto_ops[sm2_proto_inject](ep, peer_smr, peer_gid, op, tag,
-					      data, op_flags, FI_HMEM_SYSTEM, 0,
-					      iov, iov_count, total_len);
+	ret = sm2_proto_ops[sm2_proto_inject](
+		ep, peer_smr, peer_gid, op, tag, data, op_flags,
+		(struct ofi_mr **) desc, iov, iov_count, total_len);
```
aingerson, line 314: still confused on why clang feels the need to update this entire function call when just replacing `FI_HMEM_SYSTEM, 0` with `(struct ofi_mr **) desc` would do. looks like it does it properly on the call below

aingerson (commented):
One request to separate patches and some clang annoyances. Changes themselves look fine.
Can you add more in the commit message please? Even if you just copy what you did in the PR comment? That PR comment doesn't stay around; all we get is the commit message in the history of the code so it's nice to have context in there. In general, a commit message should never be empty unless it's truly the most simple patch in the world like "missing cast" or "fix line lengths" or "fix typo"

prov/sm2/src/sm2_msg.c
a-szegel: I think it is the cast that it doesn't like

fabtests/pytest/sm2/conftest.py
a-szegel: done

prov/sm2/src/sm2_msg.c
a-szegel: fixed

aingerson (approved):
Thanks!

fabtests/pytest/sm2/conftest.py
```diff
@@ -1,11 +1,16 @@
 import pytest
 
-# TODO Restore GPU Memory Types
-@pytest.fixture(scope="module", params=["host_to_host"])
+@pytest.fixture(scope="module", params=["host_to_host",
+                                        pytest.param("host_to_cuda", marks=pytest.mark.cuda_memory),
+                                        pytest.param("cuda_to_host", marks=pytest.mark.cuda_memory),
+                                        pytest.param("cuda_to_cuda", marks=pytest.mark.cuda_memory),
+                                        pytest.param("neuron_to_neuron", marks=pytest.mark.neuron_memory),
```
shijin-aws, line 7: i don't think you can run sm2 with neuron traffic. libfabric does not support copy from neuron buffer to host buffer, and there is no IPC support for it.

a-szegel: I copied the same fabtests that are in SHM.

shijin-aws: ok, that's fine then.

shijin-aws (approved):

PR #8880, opened 2023-05-01
[v1.18.x] prov/efa: Correct CPPFLAGS during configure

lrbison:
configure.m4 restored CPPFLAGS before it finished testing various feature support in rdma-core.  This meant feature testing occurred with system default includes rather than the user-specified include paths, resulting in incorrect configure results.

(cherry picked from commit 302bc43d1cb4c3a8a19115a770232deca77e8735)


lrbison:
bot:aws:retest

shijin-aws (approved):

PR #8881, opened 2023-05-02
prov/shm: introduce gdrcopy awareness to hmem copy

wenduwan:
From v1.19, OFI_HMEM_DATA_GDRCOPY_HANDLE flag will signal the presence of gdrcopy handle in ofi_mr.hmem_data. SHM provider can take advantage of gdrcopy to achieve lower memcpy latencies from/to cuda devices.

This patch introduces the logic to select gdrcopy on hmem copy paths, with the exception of cuda IPC which is not supported by gdrcopy.

For OMB single-node cuda->cuda latency we are seeing improvement(via EFA owner provider, which is responsible for gdr pinning).
|  pt2pt/osu_latency  |        |       |
|---------------------|--------|-------|
|         Size        | Before | After |
|                   0 |   0.95 |  0.93 |
|                   1 |  15.41 |  2.78 |
|                   2 |  15.18 |  4.36 |
|                   4 |  15.04 |  4.45 |
|                   8 |  15.11 |  4.38 |
|                  16 |  15.13 |  3.00 |
|                  32 |  15.29 |  2.99 |
|                  64 |  14.85 |  3.07 |
|                 128 |  14.85 |  4.66 |
|                 256 |  14.62 |  3.34 |
|                 512 |  14.82 |  4.96 |
|                1024 |  15.13 |  6.38 |
|                2048 |  15.68 | 10.99 |
|                2560 |  15.57 | 12.77 |
|                3072 |  15.57 | 15.79 |
|                3584 |  15.65 | 15.80 |
|                4096 |  15.89 | 15.95 |
|                8192 |  15.58 | 15.59 |
|               16384 |  15.31 | 15.24 |
|               32768 |  15.11 | 15.12 |
|               65536 |  15.24 | 15.27 |
|              131072 |  15.52 | 15.60 |
|              262144 |  16.01 | 16.07 |
|              524288 |  17.24 | 19.95 |
|             1048576 |  19.28 | 19.29 |
|             2097152 |  23.23 | 23.20 |
|             4194304 |  30.93 | 30.90 |

### DIFF
--- a/prov/shm/src/smr_msg.c
+++ b/prov/shm/src/smr_msg.c
@@ -315,7 +315,14 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 		use_ipc = ofi_hmem_is_ipc_enabled(iface) &&
 				smr_desc->flags & FI_HMEM_DEVICE_ONLY &&
 				!(op_flags & FI_INJECT);
+
+		if (iface == FI_HMEM_CUDA &&
+		    (smr_desc->flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+			assert(smr_desc->hmem_data);
+			gdrcopy_available = true;
+		}
 	}
+
 	proto = smr_select_proto(iface, use_ipc, smr_cma_enabled(ep, peer_smr),
 	                         gdrcopy_available, op, total_len, op_flags);
 
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -273,7 +273,8 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
 		size_t size, int dir)
 {
 	uint64_t done = 0, len;
-	uint64_t hmem_iface, hmem_device;
+	uint64_t hmem_iface, hmem_device, hmem_flags;
+	void *hmem_data;
 	char *hmem_buf;
 	size_t i;
 	int ret;
@@ -285,12 +286,29 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
 
 		if (mr && mr[i]) {
 			hmem_iface = mr[i]->iface;
+			hmem_flags = mr[i]->flags;
 			hmem_device = mr[i]->device;
+			hmem_data = mr[i]->hmem_data;
 		} else {
 			hmem_iface = FI_HMEM_SYSTEM;
+			hmem_flags = 0;
 			hmem_device = 0;
+			hmem_data = NULL;
 		}
-		if (dir == OFI_COPY_BUF_TO_IOV)
+
+		if (hmem_iface == FI_HMEM_CUDA && (hmem_flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+			/**
+			 * TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy
+			 * Note: buf must be on the host since gdrcopy does not support D2D copy
+			 */
+			if (dir == OFI_COPY_BUF_TO_IOV)
+				cuda_gdrcopy_to_dev((uint64_t) hmem_data, hmem_buf,
+				                    (char *) buf + done, len);
+			else
+				cuda_gdrcopy_from_dev((uint64_t) hmem_data, (char *) buf + done,
+				                      hmem_buf, len);
+			ret = FI_SUCCESS;
+		} else if (dir == OFI_COPY_BUF_TO_IOV)
 			ret = ofi_copy_to_hmem(hmem_iface, hmem_device, hmem_buf,
 						(char *)buf + done, len);
 		else
--- a/prov/shm/src/smr_rma.c
+++ b/prov/shm/src/smr_rma.c
@@ -182,6 +182,12 @@ static ssize_t smr_generic_rma(struct smr_ep *ep, const struct iovec *iov,
 		use_ipc = ofi_hmem_is_ipc_enabled(iface) &&
 				smr_desc->flags & FI_HMEM_DEVICE_ONLY &&
 				!(op_flags & FI_INJECT);
+
+		if (iface == FI_HMEM_CUDA &&
+		    (smr_desc->flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+			assert(smr_desc->hmem_data);
+			gdrcopy_available = true;
+		}
 	}
 	proto = smr_select_proto(iface, use_ipc, smr_cma_enabled(ep, peer_smr),
 	                         gdrcopy_available, op, total_len, op_flags);


prov/shm/src/smr_hmem.h
```diff
@@ -0,0 +1,74 @@
...
+#include "smr.h"
+
+static inline ssize_t smr_copy_from_mr_iov(void *dest, size_t size,
+                                           struct ofi_mr **mr,
+                                           const struct iovec *iov,
+                                           size_t iov_count,
+                                           uint64_t iov_offset)
+{
+    if (FI_VERSION_GE(smr_prov.fi_version, FI_VERSION(1,19)) &&
+        1 == iov_count && mr && mr[0] && FI_HMEM_CUDA == mr[0]->iface &&
```
wzamazon, line 48: Why `1 == iov_count`?

shijin-aws: i think it is following the ompi style.

shijin-aws: I remember there was such condition in shm for ipc, I guess here is that we may not want to support mixed iface type.

wenduwan: Yes. SHM currently does not support mixed HMEM iface. Here is an example: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_msg.c#L312

wenduwan:
bot:aws:retest

prov/shm/src/smr_hmem.h
aingerson: shm does not support mixed iface for the IPC path but does support mixed ifaces in the inline/inject paths which is what this is optimizing. I added the ofi_copy_from_mr_iov functions for that reason.
I'm thinking it could be better to have the gdr copy size limit be global and move this to the mr_iov copy path since the hmem data is part of the mr info. thoughts?

wenduwan: > shm does not support mixed iface for the IPC path but does support mixed ifaces in the inline/inject paths which is what this is optimizing.

Actually we are optimizing the current cuda ipc path, which only supports 1 hmem iface as mentioned. But what you are suggesting here also makes sense:

> it could be better to have the gdr copy size limit be global ...

It could be simpler to make the size limit global (because this limit is only related to the hardware and cuda driver, not efa/shm/etc provider). 

>  ... and move this to the mr_iov copy path

I considered this option but it sorta conflicted with our intention of separating gdrcopy from generic hmem copy.  Our idea is to make the caller explicitly choose the copy method. Therefore hiding gdrcopy inside mr iov copy path sounds contradictory to this end.

You pointed out a valid issue - the current impl will not use gdrcopy for inline/inject paths **when** there are mixed ifaces. This is not an issue here since we are optimizing the cuda ipc path that has only 1 iface, i.e.
```
	/* Do not inline/inject if IPC is available so device to device
	 * transfer may occur if possible. */
	if (iov_count == 1 && desc && desc[0]) {
		smr_desc = (struct ofi_mr *) *desc;
		iface = smr_desc->iface;
		use_ipc = ofi_hmem_is_ipc_enabled(iface) &&
				smr_desc->flags & FI_HMEM_DEVICE_ONLY &&
				!(op_flags & FI_INJECT);

		if (FI_VERSION_GE(smr_prov.fi_version, FI_VERSION(1,19)) &&
		    FI_HMEM_CUDA == iface &&
		    (OFI_HMEM_DATA_GDRCOPY_HANDLE & smr_desc->flags)) {
			assert(smr_desc->hmem_data);
			gdrcopy_available = true;
		}
	}
```

An alternative to changing the common mr iov path is to change e.g. `smr_copy_from_mr_iov` and iterate over the ifaces, then we can choose the proper memcopy method case by case. The downside is code duplication with mr iov copy, plus we don't have an immediate use for that.

aingerson: I'm a little confused by this. You're saying you're trying to target the IPC path for the gdrcopy optimization but the previous PR that added the gdrcopy size limit was directing it to the inline and inject path, not IPC. And in this patch you're updating all the copy functions (including the inline and inject paths) to use gdrcopy if it's available. What is the ultimate goal for these optimizations?

wenduwan: @aingerson Yes it sounds a little confusing. What I meant is that, before this change, cuda IPC would have been selected for cuda->cuda data transfers via cudaMemcpy. After this change, inline/inject would be selected in place of cuda IPC, and gdrcopy is used instead. So we are actually optimizing the **current** cuda IPC path(by not doing cuda ipc).

As a result, in such cases we know there is only 1 iov with `FI_HMEM_CUDA`, so we don't have to worry about mixed ifaces.

Does this help clear things up?

aingerson: Ah ok thank you that makes more sense :)
When you say "we know there is only 1 iov with `FI_HMEM_CUDA`" are you saying that's the case you're trying to target for the gdrcopy optimization? If there are multiple gdr vapable iovs in the array, is it less optimal to do iov_count number of gdr copies? I guess I'm trying to understand why we are only targeting the single iov rather than iterating over the array.

wenduwan: > why we are only targeting the single iov rather than iterating over the array.

This is right on - yes this PR only addresses/improves the single-iov case(which has been my focus so far), while not touching the mixed iov path. In other words, this is the minimal impl to make it "work".

I also recognize the opportunity to make it work for mixed iovs. We could do what you suggested, change common mr iov copy code, but I prefer to do it in `smr_copy_from_mr_iov`, so that each provider, e.g. shm, can explicitly request gdrcopy.

> is it less optimal to do iov_count number of gdr copies

It should be equally performant. We will be iterating over the iov and call gdrcopy individually. Reference: https://github.com/ofiwg/libfabric/pull/8836

If it sounds good to you, I will go ahead and change `smr_copy_from_mr_iov` to support mixed iov, which also shortens the `if(...)` conditions.

aingerson: @wenduwan Thanks! Yeah I think we're on the same page. My concern is mostly about the number of checks it inserts into the normal path. Something generic (but also still in shm to make the gdr decision as you suggest) is best. I'll trust whatever you come up with!
Also I think we're ok not checking the fi version here. The ofi_mr is not coming from the user but the provider/utility code which will allocate the full MR and would just set the fields to 0. 
Also the OFI convention for compares is var == val for readability ie `iov_count == 1 && mr == NULL`

wenduwan: @aingerson I stole code from [here](https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_hmem.h#L80) and [here](https://github.com/ofiwg/libfabric/blob/main/src/hmem.c#L271) in the update. This causes some duplication (passable IMO) so let me know if you think it's excessive.

aingerson: @wenduwan I see what you mean about the duplicate code... I'm leaning towards saying to just do it in the mr iov path with maybe a compile check around the gdr check? shm is the only provider using the mr iov copy functions and since the gdrcopy hmem data is in the ofi mr, I think it makes sense to say if you're going to use those functions that you would be opting in to the gdr copy if the gdr copy handle was registered it.
if there comes another provider/use case that needs to use it that explicitly needs to avoid the gdr code on a system with it available when it was registered with the gdr handle in the hmem data... then I think we can cross that bridge when we come to it.

wenduwan: @aingerson Sneaky! Yeah if SHM is the only user then we can definitely do that.

But I just realized that there is another caveat. For gdrcopy to work, the caller should make sure the src and dest cannot both be on cuda. Based on my reading SHM does not have this issue(need to run more tests), so I will leave a note in code.

aingerson: Oh if that's the case, then we can never use gdr copy for the inline/inject paths since those always copy to the shm region which will never be cuda. If we have to make sure they are both cuda then we can only call it on the progress ipc path if the cmd->iface and rx_entry->iface are both cuda. I think it would be best to keep the ofi_mr copy functions as is and just add a case in the progress ipc path to call the gdrcopy function

wenduwan: @aingerson lol I think you read this in reverse?

>  For gdrcopy to work, the caller should make sure the src and dest **cannot** both be on cuda

aingerson: @wenduwan oh lol yup. my bad!

prov/shm/src/smr_msg.c
```diff
@@ -315,7 +315,15 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 		use_ipc = ofi_hmem_is_ipc_enabled(iface) &&
 				smr_desc->flags & FI_HMEM_DEVICE_ONLY &&
 				!(op_flags & FI_INJECT);
+
+		if (FI_VERSION_GE(smr_prov.fi_version, FI_VERSION(1,19)) &&
```
aingerson, line 319: We should be able to drop the version check here. The ofi_mr is cast from the internal shm pointer so should reflect the current version as long as it's initialized which is done [here](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_mr_map.c#L260)

src/hmem.c
```diff
@@ -273,7 +273,8 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
 		size_t size, int dir)
 {
 	uint64_t done = 0, len;
-	uint64_t hmem_iface, hmem_device;
+	uint64_t hmem_iface = FI_HMEM_SYSTEM, hmem_device = 0, hmem_flags = 0;
+	void *hmem_data = NULL;
```
aingerson, line 277: you can drop the initialization here since you initialize it on the loop

src/hmem.c
```diff
@@ -285,12 +286,28 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
...
+			hmem_flags = 0;
 			hmem_device = 0;
+			hmem_data = NULL;
 		}
-		if (dir == OFI_COPY_BUF_TO_IOV)
+
+		if (hmem_iface == FI_HMEM_CUDA && (hmem_flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+			/**
+			 * TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy
+			 * Note: buf must be on the host since gdrcopy does not suport D2D copy
```
aingerson, line 302: *support

src/hmem.c
```diff
@@ -285,12 +286,28 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
...
+			 */
+			dir == OFI_COPY_BUF_TO_IOV
+			? cuda_gdrcopy_to_dev((uint64_t) hmem_data, hmem_buf,
+			                      (char *)buf + done, len)
+			: cuda_gdrcopy_from_dev((uint64_t) hmem_data,
+			                        (char *)buf + done, hmem_buf, len);
```
aingerson, line 308: put the ? and : on the top line and indent if possible depending on line length ie
```
			dir == OFI_COPY_BUF_TO_IOV ?
				cuda_gdrcopy_to_dev((uint64_t) hmem_data,
					hmem_buf, (char *)buf + done, len) :
				cuda_gdrcopy_from_dev((uint64_t) hmem_data,
			                (char *)buf + done, hmem_buf, len);
```

or alternatively do an if/else to make it more readable

aingerson (commented):
Just a few comments. I think this approach seems the best to me. Thank you!

src/hmem.c
wenduwan: Yeah. That was my PTSD because of coverity.

I will remove it for prov/shm for coding consistency.

prov/shm/src/smr_msg.c
wenduwan: Addressed

src/hmem.c
wenduwan: Addressed

src/hmem.c
wenduwan: Addressed

wenduwan:
@aingerson Please see update. Thanks for simplifying the end result!

wenduwan:
@shijin-aws @shefty Any additional comments?

src/hmem.c
```diff
@@ -285,12 +286,29 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
...
+			hmem_data = NULL;
 		}
-		if (dir == OFI_COPY_BUF_TO_IOV)
+
+		if (hmem_iface == FI_HMEM_CUDA && (hmem_flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+			/**
+			 * TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy
+			 * Note: buf must be on the host since gdrcopy does not support D2D copy
+			 */
+			if (dir == OFI_COPY_BUF_TO_IOV)
```
shijin-aws, line 304: hmmm. so there is no way to make cuda_gdrcopy* called by ofi_hmem_copy* ?

shijin-aws (approved):
LGTM, a monor comment

src/hmem.c
shijin-aws: Got it , makes sense. I see it is wrapped inside ofi_mr_iov_copy, which should be fine

shijin-aws:
what is the error in Intel CI?

aingerson:
@shijin-aws Unrelated CI error. You can ignore.

PR #8885, opened 2023-05-02
[v1.17.x] prov/hook,verbs,rxm: Fix credit size parameter for flow ctrl

raffenet:
ofi_ops_flow_ctrl defines callbacks for use between rxm and verbs.  Those callbacks include u64 parameters for credit values.  However, the callers use size_t instead.  This results in build warnings.


shefty (approved):
thanks for pulling this in

PR #8879, opened 2023-04-29
prov/sm2: Add FI_DELIVERY_COMPLETE

a-szegel:
~~This adds a 20ns regression to fabtests.  I am not sure if I did this right, specifically around:~~

edit: No regression... this is ready.

PR #8887, opened 2023-05-02
prov/sm2: Add CMA capability to SM2

a-szegel:
AWS needs SM2 to have the capability to send any message size in order to perform benchmarking on real world applications. 

The FI_DELEIVERY_COMPLETE commits will be rebased away when they are merged in with https://github.com/ofiwg/libfabric/pull/8879/files#diff-f2de75d3514282e839536e27cf5c4055b0114c7e0ff2bf7e81f9fe69d635b93d

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -87,6 +87,7 @@ extern pthread_mutex_t sm2_ep_list_lock;
 
 enum {
 	sm2_proto_inject,
+	sm2_proto_cma,
 	sm2_proto_return,
 	sm2_proto_max,
 };
@@ -104,7 +105,7 @@ enum {
  * 		   NOTE: Only grabbing the bottom 32 bits
  * 	proto - sm2 operation
  * 	sender_gid - id of msg sender
- * 	user_data - the message
+ * 	user_data - the message, for sm2_proto_inject
  */
 struct sm2_xfer_hdr {
 	volatile long int next;
@@ -122,6 +123,11 @@ struct sm2_xfer_entry {
 	struct sm2_xfer_hdr hdr;
 	uint8_t user_data[SM2_INJECT_SIZE];
 } __attribute__((packed));
+
+struct sm2_cma_data {
+	size_t iov_count;
+	struct iovec iov[SM2_IOV_LIMIT];
+};
 
 struct sm2_ep_name {
 	char name[FI_NAME_MAX];
--- a/prov/sm2/src/sm2_ep.c
+++ b/prov/sm2/src/sm2_ep.c
@@ -296,6 +296,47 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
 	return FI_SUCCESS;
 }
 
+static void sm2_format_cma(struct sm2_xfer_entry *xfer_entry,
+			   const struct iovec *iov, size_t count)
+{
+	struct sm2_cma_data *cma_data =
+		(struct sm2_cma_data *) xfer_entry->user_data;
+
+	/* All CMA messages must be delivery complete to stop completion from
+	 * being written instantly */
+	xfer_entry->hdr.op_flags |= FI_DELIVERY_COMPLETE;
+	xfer_entry->hdr.proto = sm2_proto_cma;
+	xfer_entry->hdr.size = ofi_total_iov_len(iov, count);
+	cma_data->iov_count = count;
+	memcpy(cma_data->iov, iov, sizeof(*iov) * count);
+}
+
+static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
+			  sm2_gid_t peer_gid, uint32_t op, uint64_t tag,
+			  uint64_t data, uint64_t op_flags, struct ofi_mr **mr,
+			  const struct iovec *iov, size_t iov_count,
+			  size_t total_len, void *context)
+{
+	struct sm2_xfer_entry *xfer_entry;
+	struct sm2_av *av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_region *self_region = sm2_mmap_ep_region(&av->mmap, ep->gid);
+
+	if (smr_freestack_isempty(sm2_freestack(self_region))) {
+		sm2_progress_recv(ep);
+		if (smr_freestack_isempty(sm2_freestack(self_region)))
+			return -FI_EAGAIN;
+	}
+
+	xfer_entry = smr_freestack_pop(sm2_freestack(self_region));
+	sm2_generic_format(xfer_entry, ep->gid, op, tag, data, op_flags,
+			   context);
+	sm2_format_cma(xfer_entry, iov, iov_count);
+	sm2_fifo_write(ep, peer_gid, xfer_entry);
+
+	return FI_SUCCESS;
+}
+
 int sm2_srx_bind(struct fid *fid, struct fid *bfid, uint64_t flags)
 {
 	struct sm2_srx_ctx *srx;
@@ -970,4 +1011,5 @@ ep:
 
 sm2_proto_func sm2_proto_ops[sm2_proto_max] = {
 	[sm2_proto_inject] = &sm2_do_inject,
+	[sm2_proto_cma] = &sm2_do_cma,
 };
--- a/prov/sm2/src/sm2_msg.c
+++ b/prov/sm2/src/sm2_msg.c
@@ -38,6 +38,14 @@
 #include "ofi_iov.h"
 #include "sm2.h"
 
+static inline int sm2_select_proto(uint64_t total_len)
+{
+	if (total_len <= SM2_INJECT_SIZE)
+		return sm2_proto_inject;
+
+	return sm2_proto_cma;
+}
+
 struct sm2_rx_entry *sm2_alloc_rx_entry(struct sm2_srx_ctx *srx)
 {
 	if (ofi_freestack_isempty(srx->recv_fs)) {
@@ -296,6 +304,7 @@ static ssize_t sm2_generic_sendmsg(struct sm2_ep *ep, const struct iovec *iov,
 	ssize_t ret = 0;
 	size_t total_len;
 	struct ofi_mr **mr = (struct ofi_mr **) desc;
+	int proto;
 
 	assert(iov_count <= SM2_IOV_LIMIT);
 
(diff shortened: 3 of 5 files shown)


prov/sm2/src/sm2.h
```diff
@@ -118,6 +119,11 @@ struct sm2_xfer_entry {
 	uint8_t user_data[SM2_INJECT_SIZE];
 } __attribute__((aligned(16)));
 
+struct sm2_cma_data {
+	size_t iov_count;
+	struct iovec iov[11];
```
aingerson, line 124: why 11...?

prov/sm2/src/sm2_ep.c
```diff
@@ -297,6 +297,49 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
 	return FI_SUCCESS;
 }
 
+static void sm2_format_cma(struct sm2_xfer_entry *xfer_entry,
+			   enum fi_hmem_iface iface, uint64_t device,
+			   const struct iovec *iov, size_t count)
+{
+	struct sm2_cma_data *cma_data =
+		(struct sm2_cma_data *) xfer_entry->user_data;
```
aingerson, line 305: semantically I think it makes more sense to turn this into a union now instead of casting it

prov/sm2/src/sm2_ep.c
```diff
@@ -297,6 +297,49 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+	xfer_entry->proto = sm2_proto_cma;
+	xfer_entry->size = ofi_total_iov_len(iov, count);
+	cma_data->iov_count = count;
+	memcpy(cma_data->iov, iov, sizeof(*iov) * count);
+}
+
+static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
+			  sm2_gid_t peer_gid, uint32_t op, uint64_t tag,
+			  uint64_t data, uint64_t op_flags,
+			  enum fi_hmem_iface iface, uint64_t device,
```
aingerson, line 319: same here, no need for iface/device

prov/sm2/src/sm2_ep.c
```diff
@@ -297,6 +297,49 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
 	return FI_SUCCESS;
 }
 
+static void sm2_format_cma(struct sm2_xfer_entry *xfer_entry,
+			   enum fi_hmem_iface iface, uint64_t device,
```
aingerson, line 301: you shouldn't need iface/device on the cma path

prov/sm2/src/sm2_msg.c
```diff
@@ -38,6 +38,15 @@
 #include "ofi_iov.h"
 #include "sm2.h"
 
+static inline int sm2_select_proto(uint64_t total_len)
+{
+	if (total_len <= SM2_INJECT_SIZE) {
```
aingerson, line 43: brackets

prov/sm2/src/sm2_progress.c
```diff
@@ -42,6 +42,58 @@
 #include "sm2.h"
 #include "sm2_fifo.h"
 
+static inline int sm2_cma_loop(pid_t pid, struct iovec *local,
+			       unsigned long local_cnt, struct iovec *remote,
+			       unsigned long remote_cnt, unsigned long flags,
```
aingerson, line 47: I know you're just copying code I wrote but let's drop the flags argument which is always 0.

prov/sm2/src/sm2_progress.c
```diff
@@ -42,6 +42,58 @@
...
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_ep_allocation_entry *entries =
+		sm2_mmap_entries(&sm2_av->mmap);
+	int ret;
+
+	/* TODO Need to update last argument for RMA support (as well as generic
+	 * format) */
+	ret = sm2_cma_loop(entries[xfer_entry->sender_gid].pid, iov, iov_count,
+			   cma_data->iov, cma_data->iov_count, 0,
+			   xfer_entry->size, 0);
```
aingerson, line 90: let's pass in `false` here to match the function semantically

prov/sm2/src/sm2_msg.c
```diff
@@ -38,6 +38,15 @@
...
+		return sm2_proto_inject;
+	}
+
+	return sm2_proto_cma;
```
aingerson, line 47: are you planning on having some sort of similar cma check for sm2/will that come later?

a-szegel: Yes... I want to have a 1 sided CMA check.

a-szegel:
Also this PR + HMEM PR advertise that SM2 has SIZE_MAX functionality for all messages, where in reality it will only support HMEM up to 4k.

prov/sm2/src/sm2.h
a-szegel: oh I forgot that was there lol (need to look at diff closer before pushing)... I was debugging an assert in OMPI and ofi_shm.h does this:
```
		struct iovec	iov[(SMR_MSG_DATA_LEN - sizeof(size_t)) /
				    sizeof(struct iovec)];
```
Which reduces down to 11... turned out my assert was b/c I was throwing two completions for the same message b/c I forgot to reject the transmit complete completion.

Will change back to: `SM2_IOV_LIMIT`

prov/sm2/src/sm2_ep.c
a-szegel: rebased and this went away.

prov/sm2/src/sm2_ep.c
a-szegel: done

prov/sm2/src/sm2_msg.c
a-szegel: done

prov/sm2/src/sm2_progress.c
a-szegel: done

prov/sm2/src/sm2_progress.c
a-szegel: done

prov/sm2/src/sm2_ep.c
a-szegel: done.

a-szegel:
I think this won't pass EFA CI until we have IPC protocol.

prov/sm2/src/sm2_msg.c
shijin-aws: Nit: it could be reformatted to

```
return total_len <= SM2_INJECT_SIZE ? sm2_proto_inject : sm2_proto_cma; 
```

I guess we are not doing that now as there are more protocols to come in?

shijin-aws (approved):
I haven't read the sm2_cma_loop closely. But I trust it works if it copies from shm provider.

prov/sm2/src/sm2_ep.c
```diff
@@ -289,12 +289,52 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
 }
 
+static void sm2_format_cma(struct sm2_xfer_entry *xfer_entry,
+			   const struct iovec *iov, size_t count)
+{
+	struct sm2_cma_data *cma_data = &xfer_entry->cma_data;
+
+	/* All CMA messages must be delivery complete to stop completion from
+	 * being written instantly */
+	xfer_entry->op_flags |= FI_DELIVERY_COMPLETE;
```
lrbison, line 305: By `|=` delivery completion flag, it seems to be possible to have multiple, conflicting completion semantic flags set.  

Should we consider clearing the other completion flags?

prov/sm2/src/sm2_msg.c
```diff
@@ -310,13 +319,14 @@ static ssize_t sm2_generic_sendmsg(struct sm2_ep *ep, const struct iovec *iov,
...
-					      iov_count, total_len, context);
+	proto = sm2_select_proto(total_len);
+	ret = sm2_proto_ops[proto](ep, peer_smr, peer_gid, op, tag, data,
+				   op_flags, mr, iov, iov_count, total_len,
+				   context);
 	if (ret)
 		goto unlock_cq;
 
-	if (!(op_flags & FI_DELIVERY_COMPLETE)) {
+	if (!(op_flags & FI_DELIVERY_COMPLETE) && proto != sm2_proto_cma) {
```
lrbison, line 329: Seems like by setting the op_flags |= FI_DELIVERY_COMPLETE earlier, you don't need the proto check.

prov/sm2/src/sm2_ep.c
a-szegel: I don't think it matters if you give a higher completion status... i.e.. the user asked for transmit complete, and we return delivery complete.

prov/sm2/src/sm2_msg.c
a-szegel: I do need the proto check b/c that op_flags is the op_flags passed in, and we modify the op_flags in the shared memory region xfer_entry.

prov/sm2/src/sm2_msg.c
a-szegel: more protocols to come.

a-szegel:
We need to add IPC (Device to Device) and SAR (host -> device / device to host), to this PR in order to advertise SIZE_MAX for all HMEM + Hostmem.

shijin-aws:
> device to host

device to host will use IPC. The protocol selection is determined on sender side.

prov/sm2/src/sm2_ep.c
a-szegel: Now that I have made user_data[] a flexible array, I need to undo this change b/c you can't have flexible arrays in a union:

```
In file included from prov/sm2/src/sm2_cntr.c:34:
prov/sm2/src/sm2.h:136:25: error: flexible array member in union
  136 |                 uint8_t user_data[];
```

prov/sm2/src/sm2.h
```diff
@@ -108,7 +109,9 @@ enum {
  * 		   NOTE: Only grabbing the bottom 32 bits
  * 	proto - sm2 operation
  * 	sender_gid - id of msg sender
- * 	user_data - the message
+ * 	union:
```
sunkuamzn, line 112: There's no union. Instead, the CMA protocol just casts `user_data` into `struct sm2_cma_data`.

a-szegel:
Broke out SM2 ep attr structure into host, and hmem... where HMEM is still limited by inject size.... We are now honest with our capabilities to the user (and should pass CI).

prov/sm2/src/sm2_ep.c
```diff
@@ -296,6 +296,47 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+			  const struct iovec *iov, size_t iov_count,
+			  size_t total_len, void *context)
+{
+	struct sm2_xfer_entry *xfer_entry;
+	struct sm2_av *av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_region *self_region = sm2_mmap_ep_region(&av->mmap, ep->gid);
+
+	if (smr_freestack_isempty(sm2_freestack(self_region))) {
+		sm2_progress_recv(ep);
```
aingerson, line 326: I remember you had this in your inject protocol as well. Remind me, is it for functional or performance reasons? I don't love the idea of the receive call driving receive progress. This is why we return EAGAIN and have the app do it.

aingerson (commented):
A couple of things but neither is a blocker so you can consider this a light approval :)

prov/sm2/src/sm2_ep.c
a-szegel: Functional... this is saying:
1. Check if you have available free stack entries
2. if no freestack entries are available, call recv (b/c someone might have returned one to you)
3. re-check if you have a freestack entry
4. If you do, great... send... if you don't return EAGAIN

This is required b/c some sender-only apps might never call progress.... and without this, they would never get their XFER entries back

a-szegel:
closing to https://github.com/ofiwg/libfabric/pull/8919

prov/sm2/src/sm2_ep.c
aingerson: If a send/recv call returns EAGAIN, the app needs to call progress. It is part of the API definition.

PR #8889, opened 2023-05-02
prov/efa: update error message for unresponsive receive

wzamazon:
Add a common cause for this error: instance not in the same subnet. Also adjust the language.

### DIFF
--- a/prov/efa/src/efa_strerror.c
+++ b/prov/efa/src/efa_strerror.c
@@ -59,11 +59,11 @@ static const char *efa_io_comp_status_str(enum efa_errno status)
 		[FI_EFA_REMOTE_ERROR_BAD_LENGTH]       = "Receiver scatter-gather list (SGL) too short",
 		[FI_EFA_REMOTE_ERROR_BAD_STATUS]       = "Unexpected status received from remote",
 		[FI_EFA_LOCAL_ERROR_UNRESP_REMOTE]     = "Unresponsive receiver. "
-		                                         "This error is typically caused by a peer hardware failure or "
-		                                         "incorrect inbound/outbound rules in the security group - "
-		                                         "EFA requires \"All traffic\" type allowlisting. "
-		                                         "Please also verify the peer application has not "
-		                                         "terminated unexpectedly.",
+		                                         "This error is typically caused by one of the following reasons: \n"
+							 "1) a peer hardware failure. Please verify peer application has not terminate unexpectedly.\n"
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_strerror.c
```diff
@@ -59,11 +59,11 @@ static const char *efa_io_comp_status_str(enum efa_errno status)
...
-		                                         "This error is typically caused by a peer hardware failure or "
-		                                         "incorrect inbound/outbound rules in the security group - "
-		                                         "EFA requires \"All traffic\" type allowlisting. "
-		                                         "Please also verify the peer application has not "
-		                                         "terminated unexpectedly.",
+		                                         "This error is typically caused by one of the following reasons: \n"
+							 "1) a peer hardware failure. Please verify peer application has not terminate unexpectedly.\n"
+		                                         "2) incorrect inbound/outbound rules in the security group. Please verify \"All traffic\" "
+							 "are allowed in both inbound and outbound rule for the same security group.\n"
+							 "3) instances are not in same subnet.",
```
wenduwan, line 66: - `terminate` -> `terminated`
- `same` -> `the same`

Likely you also need to change the error message in https://github.com/ofiwg/libfabric/blob/main/prov/efa/test/efa_unit_test_cq.c

bwbarrett: I get that this PR is just to change the error message, but we should be able to do better here.  The second two issues (SGs and subnet) are only ever going to happen on the first packet.  They aren't causes for failures after the handshake.  So we should be able to split those two cases out from the first.

Second, I don't think you want to suggest that it's a peer hardware failure, as much as the peer process is no longer there.  There are plenty of reasons (bugs in software, memory exhaustion, etc.) that processes die that aren't hardware failures.

lrbison: I feel like testing for specific error messages makes the unit tests more brittle than necessary.  I hope we don't decide to do such a test for all error messages.

Maybe we should also replace "typically" with "often" to soften the statement a little.

1) I think we can address Brian's concern by simply deleting the world "hardware" in the first reason.

2) "are allowed in both inbound" -> "is allowed for both inbound"

I think Wenduo's "the same" applies to number 3. (I was puzzling over how it fit in 2 [my mind's regex is in greedy mode I guess])

wzamazon:
Need more work to print different error message for before and after we got handshake from a peer

PR #8890, opened 2023-05-03
prov/tcp: separate the max inject and recv buf size

ooststep:
this change allows for larger buffers on the receive side without forcing larger copies on the send side

### DIFF
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -69,6 +69,7 @@
 
 #define XNET_RDM_VERSION	0
 #define XNET_DEF_INJECT		128
+#define XNET_DEF_BUF_SIZE	16384
 #define XNET_MAX_EVENTS		128
 #define XNET_MIN_MULTI_RECV	16384
 #define XNET_PORT_MAX_RANGE	(USHRT_MAX)
@@ -92,7 +93,7 @@ extern int xnet_disable_autoprog;
 extern int xnet_io_uring;
 extern int xnet_max_saved;
 extern size_t xnet_max_inject;
-
+extern size_t xnet_buf_size;
 struct xnet_xfer_entry;
 struct xnet_ep;
 struct xnet_rdm;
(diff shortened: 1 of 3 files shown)


prov/tcp/src/xnet_init.c
```diff
@@ -132,6 +131,12 @@ static void xnet_init_env(void)
 			"applications to prevent hangs. (default: %d)",
 			xnet_max_saved);
 	fi_param_get_int(&xnet_prov, "max_saved", &xnet_max_saved);
+
+	fi_param_define(&xnet_prov, "max_rx_size", FI_PARAM_SIZE_T,
+			"maximum size for message buffers at the receiver "
+			"(default: %zu)", xnet_max_rx_size);
+	fi_param_get_size_t(&xnet_prov, "max_rx_size", &xnet_max_rx_size);
```
shefty, line 138: We need to check that the size >= max_inject and increase if not.  I'd leave the name for the environment variable as-is (max_rx_size), as it's how a user would view it.

prov/tcp/src/xnet.h
```diff
@@ -69,6 +69,7 @@
 
 #define XNET_RDM_VERSION	0
 #define XNET_DEF_INJECT		128
+#define XNET_DEF_MAX_RX		16384
```
shefty, line 72: We'll need to figure out the correct size here.  If the receiver is slow posting buffers, a larger size allows buffering the message, but results in more data being copied twice.  I'd rename, see comment below.

prov/tcp/src/xnet.h
```diff
@@ -92,6 +93,7 @@ extern int xnet_disable_autoprog;
 extern int xnet_io_uring;
 extern int xnet_max_saved;
 extern size_t xnet_max_inject;
+extern size_t xnet_max_rx_size;
```
shefty, line 96: I'd rename to xnet_buf_size.  This impacts the buffer size for both the send and receive side.  And it technically is the buffer size, not a max.

shefty (approved):

PR #8895, opened 2023-05-04
include/ofi_mem.h: Fix Freestack Alignment

a-szegel:
Switching freestack alignment to be at 4096 page boundaries has a latency benefit for both SHM/SM2.

For SHM: .33us -> .32us
For SM2: .2us -> .19us

### DIFF
--- a/include/ofi_mem.h
+++ b/include/ofi_mem.h
@@ -213,7 +213,7 @@ void dummy ## name (void) /* work-around global ; scope */
 /*
  * Buffer pool (free stack) template for shared memory regions
  */
-#define SMR_ALIGN_BOUNDARY	64
+#define SMR_ALIGN_BOUNDARY	4096
 #define SMR_FREESTACK_EMPTY	(-1)
 
 struct smr_freestack {


sunkuamzn:
Is 4096 the best alignment for all OSs and architectures? Why?

a-szegel:
This change seems to improve pt2pt latency, but slightly negatively impacts osu_alltoallw latency by about 1-2%.

a-szegel:
I think the SMR freestack still has alignment issues... I think this is not a complete patch yet.

a-szegel:
I am confident in this change.  I have verified that there is no regression in collectives, and I get the same latency speedup.  I have verified that the freestack entries are all aligned on the 4k boundary.

shefty:

shefty:
I don't think I would call either of these 'fixes'.  They simply change the alignment for a very small latency improvement at best.  Is the latency improvement repeatable over multiple test iterations?

a-szegel:
Yes it is consistent.

a-szegel:
@shefty Updated commit message to get rid of the word fix, do you have any other comments?

prov/sm2/src/sm2_coordination.h
```diff
@@ -48,7 +48,7 @@
 
 #include <rdma/providers/fi_prov.h>
 
-#define SM2_INJECT_SIZE	      4096
+#define SM2_INJECT_SIZE	      4048
```
shijin-aws, line 51: Can we calculate this size from SMR_ALIGN_BOUNDARY, like https://github.com/ofiwg/libfabric/blob/main/include/ofi_shm.h#L123

a-szegel: ### Option 1:
The `struct sm2_xfer_entry` is header + data.. this would require separating out the header which is a few hundered lines of code... which seems like overkill for a 1 line fix.  This is something that I want to do eventually... but I think it can wait.

### Option 2:
Change these to use SM2_INJECT_SIZE
```
└─╼ grep -rI "sizeof(struct sm2_xfer_entry)" prov/sm2
prov/sm2/src/sm2_util.c:	total_size += freestack_size(sizeof(struct sm2_xfer_entry),
prov/sm2/src/sm2_util.c:			   sizeof(struct sm2_xfer_entry));
```
and update:
```
struct sm2_xfer_entry {
	volatile long int next;
	uint64_t size;
	uint64_t cq_data;
	uint64_t tag;
	uint32_t op;
	uint32_t op_flags;
	uint32_t proto;
	sm2_gid_t sender_gid;
	uint8_t user_data[];
} __attribute__((aligned(16)));
```
### Option 3:
A good comment, and fix later, when we have a functional reason to change the header... I prefer option 3.

a-szegel: I just tried to do option 1, and it doesn't work b/c we get padding:
```
#define SM2_INJECT_USER_DATA_LENGTH \
	(SM2_INJECT_SIZE - sizeof(struct sm2_inject_header))
...
struct sm2_inject_header {
	volatile long int next;
	uint64_t size;
	uint64_t cq_data;
	uint64_t tag;
	uint64_t context;
	uint32_t op;
	uint32_t op_flags;
	uint32_t proto;
	sm2_gid_t sender_gid;
} __attribute__((aligned(16)));

struct sm2_xfer_entry {
	struct sm2_inject_header hdr;
	uint8_t user_data[SM2_INJECT_USER_DATA_LENGTH];
} __attribute__((aligned(16)));
```

```
	printf("sm2_freestack_size: %d\n", SM2_INJECT_USER_DATA_LENGTH);
```

 sm2_freestack_size: 4032

a-szegel: No but I don't think we need that b/c our memory performance is good, its the number of instructions on the critical path that is the problem.

aingerson: I'm confused. You're saying we're good with the memory utilization so we don't need it but you're also trying we need to align the freestack? Sounds like there are memory performance issues then?

lrbison: @a-szegel I'm not sure you want alignment.  I think you want to define that struct using `__attribute__((packed))` and then you can use the straight-forward computation of SM2_INJECT_SIZE as @shijin-aws suggests, right?

@a-szegel Can you expand on "memory performance is good"?  I think what you mean is that you are looking at two metrics: perhaps cache miss rate or page-faults, and instructions/cycle, and comparing SM2 to SHM as a baseline.  Is that correct?

a-szegel: @lrbison 
1. Packing doesn't work as you expect, the inject size calculation is: SM2_INJECT_USER_DATA_LENGTH: 3992
2. From some benchmarking I did... it seemed like libfabric could gain a lot more performance by being algorithmically more efficient, rather than trying to be more memory efficient when compared to OMPI.

@aingerson 
1. This fix aims to improve page access via alignment.  You are suggesting that I improve cache hits via locality which is different than this fix... and I will look at in the future, but isn't in the same spirit as this fix.

a-szegel:
I simplified this PR to just be the ofi_hmem 4k alignment.  It is good for SHM/SM2 to start their freestack on a new page (huge pages don't matter), instead of a 64byte alignment.

lrbison (approved):
Looks good to me.

I see also that prov/psm3/inc has a copy of SMR_ALIGN_BOUNDARY (as well as several other functions).  Do they know we are diverging?

a-szegel:
The psm3 provider choose to not use the public shared implementation of libfabric, so it is their responsibility to monitor for divergences (and take what they care about).

prov/sm2/src/sm2_coordination.h
aingerson: @a-szegel I was suggesting the improvement as both - the smr freestack code has all of the payload memory back to back chunks so that it could easily be aligned without the need for headers and such. This payload itself is still not aligned because of the header, and is going to span 2 pages because of the header. Separating it into inline + inject allows the inline protocol to only grab the single page and the inject protocol to grab 2 where one is entirely the payload

a-szegel: On my other branch: https://github.com/ofiwg/libfabric/pull/8913, I lowered the inject size to 256 and 128 and didn't see an improvement in pt2pt latency... will do more tests with that.

a-szegel: after doing more testing... it seems like the locality does provide a slight improvement on 96 rank alltoallw for small message sizes but doesn't impact latency... I can put out a separate fix

aingerson (approved):

a-szegel:
EFA CI failure is unrelated.

wzamazon:
@shefty @aingerson  Can you share the error from "continuous-integration/jenkins/pr-head"?

aingerson:
@wzamazon Unrelated

a-szegel:
Thanks for merge!

PR #8898, opened 2023-05-05
[v1.18.x] prov/efa: Bugfix for initializing SHM offload

a-szegel:
Do not set efa_env.enable_shm_transfer=0, when we are not able to initialize SHM b/c of its capabilities.  This is b/c efa_env is a global variable. When using OMPI, SM2 fails to be launched for the BTL, which will make the next EP instance of EFA fail for MTL as well.

I verified this doesn't mess up other places where efa_env.enable_shm_transfer is used.  The only other questionable place is in efa_prov_info, where we resize the EFA inject size to the SHM inject size before we are sure that we will be able to use a SHM offload.


wenduwan (approved):

shijin-aws:
Please update the commit message to say it's modified based on b4647eeb0e1cbf039cb4297eb61051fc5ed01b31

a-szegel:
@shijin-aws Done.

shijin-aws (approved):

PR #8899, opened 2023-05-05
fabtests: enable a list of arbitrary message sizes

lrbison:
In addition to the range feature (r:start,step,stop) the -S option now supports an explicit list of message sizes (l:1,1,2,3,5,8,13) of any length.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -157,7 +157,7 @@ unsigned int test_cnt = (sizeof def_test_sizes / sizeof def_test_sizes[0]);
 
 struct test_size_param *test_size = def_test_sizes;
 /* range of messages is dynamically allocated */
-struct test_size_param *range_test_size;
+struct test_size_param *user_test_sizes;
 
 static const char integ_alphabet[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 static const int integ_alphabet_length = (sizeof(integ_alphabet)/sizeof(*integ_alphabet)) - 1;
@@ -1721,7 +1721,7 @@ void ft_free_res(void)
 	rx_ctx_arr = NULL;
 
 	ft_close_fids();
-	free(range_test_size);
+	free(user_test_sizes);
 	if (buf) {
 		ret = ft_hmem_free(opts.iface, buf);
 		if (ret)
@@ -3204,16 +3204,16 @@ void ft_parse_opts_range(char* optarg)
 	}
 	assert(end >= start && inc > 0);
 	test_cnt = (end - start) / inc + 1;
-	range_test_size = calloc(test_cnt, sizeof(*range_test_size));
-	if (!range_test_size) {
+	user_test_sizes = calloc(test_cnt, sizeof(*user_test_sizes));
+	if (!user_test_sizes) {
 		perror("calloc");
 		exit(EXIT_FAILURE);
 	}
 	for (i = 0; i < test_cnt && i < end; i++) {
-		range_test_size[i].size = start + (i * inc);
-		range_test_size[i].enable_flags = 0;
+		user_test_sizes[i].size = start + (i * inc);
+		user_test_sizes[i].enable_flags = 0;
 	}
-	test_size = range_test_size;
+	test_size = user_test_sizes;
 }
 
 void ft_parsecsopts(int op, char *optarg, struct ft_opts *opts)
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -3058,7 +3058,8 @@ void ft_csusage(char *name, char *desc)
 	FT_PRINT_OPTS_USAGE("-Q", "bind EQ to domain (vs. endpoint)");
 	FT_PRINT_OPTS_USAGE("-w <number>", "number of warmup iterations");
 	FT_PRINT_OPTS_USAGE("-S <size>", "specific transfer size or "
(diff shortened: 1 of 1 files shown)


wzamazon (commented):
commit message does not seem right. 
```
supports an explicit list of message sizes (r:1,1,2,3,5,8,13) of any
```
I think you meant `l:1,1,2,3

fabtests/common/shared.c
```diff
@@ -3216,6 +3217,38 @@ void ft_parse_opts_range(char* optarg)
...
+	token = strtok(optarg, ",");
+	i = 0;
+	while (token != NULL) {
+		range_test_size[i].enable_flags = 0;
+		ret = sscanf(token, "%ld", &range_test_size[i].size);
+		if (ret != 1) {
+			perror("sscanf cannot parse integer in list");
+			exit(EXIT_FAILURE);
+		}
+		printf("size[%d] = %ld\n",i,range_test_size[i].size);
```
lrbison, line 3245: oops

fabtests/common/shared.c
```diff
@@ -3216,6 +3217,37 @@ void ft_parse_opts_range(char* optarg)
...
+{
+	int i, ret;
+	char *token;
+
+	optarg += 2; // remove 'l:'
+	test_cnt = 1;
+	for (i=0; optarg[i] != '\0'; i++) {
+		test_cnt += optarg[i] == ',';
+	}
+	range_test_size = calloc(test_cnt, sizeof(*range_test_size));
```
wenduwan, line 3230: My guess is we need another variable `list_test_size`

fabtests/common/shared.c
```diff
@@ -3216,6 +3217,37 @@ void ft_parse_opts_range(char* optarg)
...
 
+void ft_parse_opts_list(char* optarg)
+{
+	int i, ret;
+	char *token;
+
+	optarg += 2; // remove 'l:'
+	test_cnt = 1;
+	for (i=0; optarg[i] != '\0'; i++) {
+		test_cnt += optarg[i] == ',';
```
wenduwan, line 3228: It would be annoying for someone with a sticky keyboard, e.g. `1,,2,,3,`

fabtests/common/shared.c
```diff
@@ -3216,6 +3217,37 @@ void ft_parse_opts_range(char* optarg)
...
+	if (!range_test_size) {
+		perror("calloc");
+		exit(EXIT_FAILURE);
+	}
+
+	token = strtok(optarg, ",");
+	i = 0;
+	while (token != NULL) {
+		range_test_size[i].enable_flags = 0;
+		ret = sscanf(token, "%ld", &range_test_size[i].size);
```
wenduwan, line 3240: `%ld` -> `%lu`?

wenduwan (approved):

fabtests/common/shared.c
lrbison: maybe.  I kept formatting and size consistent with the range parsing

fabtests/common/shared.c
lrbison: It will raise error during parsing and print `sscanf cannot parse integer in list`

fabtests/common/shared.c
lrbison: it's called `test_cnt`

shefty: He's referring to range_test_size -> list_test_size.  I think we need to rename range_test_size to user_test_sizes.

fabtests/common/shared.c
```diff
@@ -3216,6 +3217,37 @@ void ft_parse_opts_range(char* optarg)
...
 }
 
+void ft_parse_opts_list(char* optarg)
+{
+	int i, ret;
+	char *token;
+
+	optarg += 2; // remove 'l:'
+	test_cnt = 1;
+	for (i=0; optarg[i] != '\0'; i++) {
```
shefty, line 3227: nit: please add spaces around = in ```i = 0```

fabtests/common/shared.c
shefty: Will it?  strtok() will skip over multiple tokens, and we don't verify that test_cnt integers were read.  I'm not actually worried about this.  Worse case, we have some 0 size values in the array, if it doesn't error out.

fabtests/common/shared.c
shefty: unsigned would be safer

fabtests/common/shared.c
```diff
@@ -3216,6 +3217,37 @@ void ft_parse_opts_range(char* optarg)
...
+	token = strtok(optarg, ",");
+	i = 0;
+	while (token != NULL) {
+		range_test_size[i].enable_flags = 0;
+		ret = sscanf(token, "%ld", &range_test_size[i].size);
+		if (ret != 1) {
+			perror("sscanf cannot parse integer in list");
+			exit(EXIT_FAILURE);
+		}
+		i++;
```
shefty, line 3245: If we use test_cnt instead of i in this second loop, extra ```,``` should be silently discarded.

fabtests/common/shared.c
lrbison: Ah, I didn't realize that was strtok's behavior, but of course you are correct.

Your solution with test_cnt as loop variable is a good one, I did that, and tested to confirm.

I also noticed sscanf doesn't typically set errno, so perror may not be appropriate.  Testing a list with some garbage values gave me:
`sscanf cannot parse integer in list: Success` so I changed the perror to a fprintf(stderr,...).

shefty (approved):

PR #8900, opened 2023-05-05
fabtests/pytests: skip neuron memory type for shm and sm2

shijin-aws:
intra node communication involving neuron iface won't work for shm and sm2 because the neuron to host memcpy is currently not supported in libfabric. This patch skips neuron for shm and sm2.


shijin-aws:
Reference: https://github.com/ofiwg/libfabric/blob/main/src/hmem_neuron.c#L133-L138

wenduwan (approved):

PR #8901, opened 2023-05-06
prov/verbs(ND): Release lock in network direct error paths

shefty:
Flagged by coverity.  Return in case of errors without releasing lock.


shefty:
bot:aws:retest

PR #8902, opened 2023-05-07
prov/efa: fix the gdrcopy_handle procedure

shijin-aws:
This PR contains 2 commits to fix the procedure that efa pass gdrcopy handle to shm.

1. Use shm with 1.19 API, since hmem_data is introduced for libfabric >= 1.19
2. Remove the version check in efa_mr.c, since efa is using hmem_data internally so it should not care what api version that application is using

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -929,9 +929,7 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 			shm_flags |= FI_HMEM_DEVICE_ONLY;
 		}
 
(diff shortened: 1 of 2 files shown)


shijin-aws:
This PR needs some major change after discussing with the team. We need to remove the condition of FI_VERSION check for the hmem_data and gdrcopy handle flags within efa_mr.c, use 1.19 API version when calling shm's get_info.

All of the changes mentioned above should be based on a separate PR that update libfabric version to 1.19.

wenduwan (approved):

wenduwan:
nit PR description should be updated.

shijin-aws:
> .

Thx, updated.

wzamazon (approved):

PR #8903, opened 2023-05-08
prov/efa: serialize get_msg/tag and queue_msg/tag

wzamazon:
Per fi_peer document, get_msg/tag and query_msg/tag should be serialized. Otherwise, there may be deadlock, when another thread call fi_recv/fi_trecv between get_msg/tag and queue_msg/tag.

This patch serialize the two calls by put them under same lock.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_srx.c
+++ b/prov/efa/src/rdm/efa_rdm_srx.c
@@ -126,16 +126,25 @@ static int efa_rdm_srx_get_msg(struct fid_peer_srx *srx, fi_addr_t addr,
 	rxe = rxr_pkt_get_msgrtm_rxe(efa_rdm_ep, &pkt_entry);
 	if (OFI_UNLIKELY(!rxe)) {
 		efa_base_ep_write_eq_error(&efa_rdm_ep->base_ep, FI_ENOBUFS, FI_EFA_ERR_RXE_POOL_EXHAUSTED);
-		ret = -FI_ENOBUFS;
-		goto out;
+		ofi_mutex_unlock(&efa_rdm_ep->base_ep.util_ep.lock);
+		return -FI_ENOBUFS;
 	}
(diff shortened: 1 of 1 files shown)


sunkuamzn (approved):
Did you observe the deadlock when running a multithreaded application?

prov/efa/src/rdm/efa_rdm_srx.c
```diff
@@ -126,16 +126,25 @@ static int efa_rdm_srx_get_msg(struct fid_peer_srx *srx, fi_addr_t addr,
...
+	/* Notice that ep lock is not released when no rxe was found.
+	 * This is because when no rxe was found, peer must allocate
+	 * its own rxe then call efa_rdm_srx_queue_msg() to put
+	 * the rxe in EFA's unexpected msg list.
+	 * efa_rdm_srx_get_msg() and efa_rdm_srx_queue_msg() must be
+	 * serialized, therefore the ep lock cannot be released.
+	 * efa_rdm_srx_queue_msg() will release the lock.
+	 */
+	if (ret == FI_SUCCESS)
+		ofi_mutex_unlock(&efa_rdm_ep->base_ep.util_ep.lock);
```
wenduwan, line 146: Is it a good approach to use the lock in such a way that:
1. We acquire the lock in Function A, and somehow do not release it under certain scenarios. Then,
2. We expect that the lock will be released in Function B.

Shall we consider using a dedicated atomic ref counter instead - I have no proof but `base_ep.util_ep.lock` is used extensively, which makes me wonder if this will cause deadlocks somewhere else.

wzamazon:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_srx.c
wzamazon: This is how fi_peer document say how peer provider should implement. If we follow the document, we should not deadlock.

PR #8906, opened 2023-05-08
include/fi_peer: add peer srx progress functionality

aingerson:
Add peer srx functions for queueing and driving progress entries. This specifically allows the peer srx API to handle unexpected messages from unknown peers. The idea behind this API addition is that the peer can define a progress call for the owner to call on fi_recv in order to update an entry's fields.

I went back and forth about a few things which I'm open to feedback on:
1) Whether or not to define separate queue/queue_progress functions. The patch has separate functions for it to make it more generic but if the only aim for this is to handle the unspec address situation, we can just have one and make the owner check the address and pick the queue depending on the address.
2) Similar to the above, what the progress calls should return/how the owner knows when to consider it a "normal" queue entry. Right now I have it returning FI_SUCCESS if the entry is good to go. We can extend that to define -FI_EAGAIN as a need to drive progress again and error codes for other things - removing the entry, etc.
3) Whether or not to consider ordering on unexpected messages posted through the above sequence. In the follow up util srx code, I have sequence numbers attached to queued entries (whether known or unknown peers). When an entry from the progress queue is updated and moves to the source unexpected list, it walks through the list to put it in the right order. This catches the case of unexpected unspec addr, av insert, unexpected unspec addr, post recv (ie the first message should be matched first but the second one got inserted into the per peer list first). Not sure if this is important or not to catch. It's not a huge overhead for this specific edge case but something I would want to document in the man pages just in case.


man/fi_peer.3.md
```diff
@@ -449,6 +449,17 @@ peer's discard function if the message is to be discarded)
 (TBD: The peer may need to update the src addr if the remote endpoint is
```
shijin-aws: Do we still need this TBD after the change?

shijin-aws: Nvm, see it removed in the second commit.

man/fi_peer.3.md
```diff
@@ -525,20 +545,25 @@ application has posted the matching receive buffer.
...
+   FI_DIRECTED_RECV and can only match to receives posted with FI_ADDR_UNSPEC.
 3. The owner fails to find a matching receive buffer.
 4. The owner allocates a rx_entry with any known fields and returns -FI_ENOENT.
 5. The peer allocates any resources needed to handle the asynchronous processing
-   and sets peer_context accordingly.
-6. The peer allocates any needed resources for processing the unexpected
-   message and sets the peer_context accordingly, calling the owner's queue
-   function when ready to queue the unexpected message from the peer.
+   and sets peer_context accordingly, calling the owner's queue
+   function if the address is known or queue_progress if the address is
```
shijin-aws, line 555: the step number should be shifted.

man/fi_peer.3.md
```diff
@@ -514,20 +545,25 @@ application has posted the matching receive buffer.
...
+   FI_ADDR_UNSPEC when ready to queue the unexpected message from the peer.
 7. The application calls fi_recv() / fi_trecv() on owner, posting the
    matching receive buffer.
-8. The owner matches the receive with the queued message on the peer.
-9. The owner removes the queued request, fills in the rest of the known fields
+8. The owner iterates through all queued progress entries and calls the peer's
```
shijin-aws, line 559: I'd like to see how the `progress()` is implemented. Is there a link to how it is handled in tcp provider? I can search it too.

shijin-aws: Nvm, find it in https://github.com/ofiwg/libfabric/pull/8907/commits/914c979b0a5af1b3260ef9e18697553b4ec85f0b#diff-8b61228cbd6da6eedda142bb8d2e0318c31db45b90cea54e091af319fa4d0f5e for util srx

shijin-aws:
>but if the only aim for this is to handle the unspec address situation, we can just have one and make the owner check the address and pick the queue depending on the address.

So, in that case, there is no `queue_msg_progress`, peer only call `queue_msg` and owner selects different queues based on different address? I am slightly inclined to keep them separate to keep `queue_msg` simple and clear. We probably can rename `queue_msg` to `queue_msg_unexpected`? Since in this case it is for unexpected queue only?

shijin-aws:
>Right now I have it returning FI_SUCCESS if the entry is good to go. We can extend that to define -FI_EAGAIN as a need to drive progress again and error codes for other things - removing the entry, etc.

If that `progress` returns no good, I think it means the peer is not inserted to the av, which is unexpected and we should warn and fail in this case.

man/fi_peer.3.md
```diff
@@ -446,15 +451,22 @@ call to the owner's queue_msg/tag. The get and queue messages should be serializ
...
 matched to a receive buffer, or discard it if needed.
 
+## fi_ops_srx_owner::queue_msg_progress() / queue_tag_progress();
+
+Queues an unexpected message with a progress function call. This progress
+function gets called by the owner on every post receive call to allow the peer
+to update any fields, specifically the address field. This allows the peer to
+queue a message from an unknown peer and update it later once the fi_addr is
+known. Note that messages in the progress queue can still be matched to receives
+posted with FI_ADDR_UNSPEC.
```
shefty, line 468: I need more time to think about this.  But calling the progress function every progress call is inefficient.  Unless an unmatched directed receive has been posted, there's nothing to do.  And even in that case, unless some address has been inserted into the AV, no progress would be made.

include/rdma/providers/fi_peer.h
```diff
@@ -164,6 +164,8 @@ struct fi_ops_srx_owner {
 			size_t size, uint64_t tag, struct fi_peer_rx_entry **entry);
 	int	(*queue_msg)(struct fi_peer_rx_entry *entry);
 	int	(*queue_tag)(struct fi_peer_rx_entry *entry);
+	int	(*queue_msg_progress)(struct fi_peer_rx_entry *entry);
+	int	(*queue_tag_progress)(struct fi_peer_rx_entry *entry);
```
shefty, line 168: The only know issue with the current APIs is that the address may not be known until after the message is queued.  That leads me to consider a callback such as ```(*modify_addr)(struct fi_peer_rx_entry *entry, fi_addr_t addr)```.

For the peer to make such a call, it would need to track fi_peer_rx_entry when the address isn't known.  I don't like the idea that both the peer and owner would need to maintain references to fi_peer_rx_entry.  Passing the structure back and forth seems far more desirable for managing the lifetime.  Soooo....

I'm wondering if we shouldn't reverse the flow here.  As proposed, the owner will iterate over all queued messages that don't have an address, so that the peer can update the entry.  We could instead place the burden on the peer to notify the owner that an address has changed.  This could look similar to a list iterator operation.  For example,

```
owner::int (*update_addr)(struct fid_peer_srx *srx, fi_addr_t (*get_addr)(struct fi_peer_rx_entry *entry));
```
1. The peer calls update_addr() (Possibly in response to an AV insertion.  We could instead require the callback be deferred until normal CQ progress is called for easier synchronization.)
2. The owner iterates over the rx_entry list for that peer, where the src_addr is set to NOTAVAIL.
3. The owner calls the get_addr() method for each rx_entry
4. The peer either returns an updated fi_addr_t value or NOTAVAIL if it's still unknown.

If the peer isn't configured with FI_DIRECTED_RECV, then the address can be set to anything.  But the flow of calling get_msg() + queue_msg() remains the same.  Otherwise, when FI_DIRECTED_RECV is not set, which call flow to use isn't obvious.

shefty: This is the current behavior for tcp:

If a message arrives that is unmatched and the peer address is unknown, the message is left on the socket and socket events are disabled.  The unexpected queue is size 1 until it knows the peer's address.  When a new address is inserted into the AV, a 'peer' data structure is updated with the address, but there's no message progress made.  Basically, there's nothing that can be done with this socket until a receive is posted.

When a receive is posted, it starts progress.  If the receive is for ADDR_UNSPEC, all sockets with a message queued on it are checked for a possible match.  If the receive targets a specific src, the same thing happens.  However, at this point, we know that the src address has been inserted into the AV, and the peer structure would have been updated with an address.  So, the src address check is part of the matching.

In both cases, the queued message is received into the posted buffer.  Once done, the process repeats.  If a message arrives that is unmatched and the peer address is unknown, the message is left on the socket... blah blah.

Sooo... I think tcp can be updated to handle an update_addr() flow.  Instead iterating over the list of sockets with unmatched messages when a receive is posted, AV insert iterates over the list, finds which socket has been updated (if any), and updates the src_addr.  Note that a call to update_addr() can never result in matching a posted receive buffer and would be invoked outside of any progress loop.

aingerson: As long as you're sure tcp can be updated to move that to the AV instead of the recv path, that's fine with me. I just wasn't sure.
So peer always calls get/queue but has to keep track of all unspec entries on its own in order to call update_addr later?
owner can check the unspec addr list for an unspec addr recv post
on peer av insert, peer goes through the list of rx entries for the updated peer and calls owner_srx->update_addr on each one?
on start_msg, if the rx_entry/unexpected message is unspec addr, the peer is also responsible for removing the entry from its queued update list so it doesn't call update addr on it again?

the part about the peer being responsible for duplicate tracking it the part I want to verify. This is going to add more complexity on the peer's side

shefty: I'm fairly sure tcp can move to this model.

The peer does *not* need to track entries.  Example code:

```
/* peer */
fi_addr_t peer_get_addr(struct fi_peer_rx_entry *entry)
{
    dest_context = entry->peer_context;
    return dest_ip_array[dest_context->dest_ip_index].fi_addr
}

peer_av_insert(..dest_ip..)
{
    av[next].dest_ip = dest_ip
    x = find index of dest_ip in dest_ip_array[]
    if (x == -1) {
        dest_ip_array[last++].fi_addr = next++;
        return;
    }

    assert(dest_ip_array[x].fi_addr == FI_ADDR_UNSPEC)
    dest_ip_array[x].fi_addr = next++;
    srx->owner->update_addr(srx, peer_get_addr)
}
```

```
/* owner */
owner_update_addr(..., get_addr())
{
    for each entry in unknown addr list {
        fi_addr =get_addr(entry)
        if (fi_addr != FI_ADDR_UNSPEC) {
            remove entry
            insert entry on src_addr_list
        }
    }
}
```

1. address is inserted into peer AV
2. peer always calls srx->update_addr()
3. owner iterates over all saved entries for that peer where the address is UNSPEC.
3a. owner calls peer->get_addr()
3b. peer returns the new address for the entry or UNSPEC.
3c. owner updates the saved entry

The peer needs some way to know which source a message arrived on, so that once the source is inserted into the AV, it can update it correctly.

Note: get_addr() could be set earlier, but I defined the API as shown to make the operation clearer, which apparently didn't work.

aingerson: Ah, that makes more sense. I thought you were just thinking of a single update_addr call that the peer would call into the owner for a single entry. This makes more sense with the get_addr call as well. Thanks!

aingerson:
Updated to reflect update_addr/get_addr model

aingerson:
Updated with foreach_unspec function that accepts a function pointer

man/fi_peer.3.md
```diff
@@ -387,6 +387,9 @@ struct fi_ops_srx_owner {
                    size_t size, uint64_t tag, struct fi_peer_rx_entry **entry);
     int (*queue_msg)(struct fi_peer_rx_entry *entry);
     int (*queue_tag)(struct fi_peer_rx_entry *entry);
+    void	(*foreach_unspec_addr)(struct fid_peer_srx *srx,
+                  fi_addr_t (*get_addr)(struct fi_peer_rx_entry *));
```
shefty, line 391: nit: alignment is off from other lines

man/fi_peer.3.md
```diff
@@ -525,23 +533,32 @@ application has posted the matching receive buffer.
...
+   the owner cannot match a directed receive with an unexpected message whose
+   address is unknown.
+8. The owner removes the queued request, fills in the rest of the known fields
    and calls the peer->start_msg() / start_tag() function.
-10. When the peer finishes processing the message and completes it on its own
+9. When the peer finishes processing the message and completes it on its own
    CQ, the peer will call free_entry to free the entry with the owner.
 ```
 
+Whenever a peer's addressing is updated, it needs to call the owner's
```
shefty, line 555: May be worth clarifying this, e.g. ``` address is updated (e.g. via fi_av_insert()), it needs```

shefty (approved):
minor nits on man page wording

aingerson:
Updated

shijin-aws (approved):
LGTM, thank you!

PR #8907, opened 2023-05-08
prov/util, shm: add util srx implementation, replace shm srx with util_srx

aingerson:
First 2 commits are from other PR (#8906) in order to build and test but will be removed once it is merged.

### DIFF
--- a/fabtests/test_configs/shm/shm.exclude
+++ b/fabtests/test_configs/shm/shm.exclude
@@ -13,7 +13,6 @@ poll
 
 # Exclude tests with unsupported capabilities
 -k
-rdm_tagged_peek
 cm_data
 trigger
 shared_ctx
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -302,9 +302,7 @@ struct util_ep {
 	uint64_t		caps;
 	uint64_t		flags;
 	ofi_ep_progress_func	progress;
-	ofi_mutex_t		lock;
-	ofi_mutex_lock_t	lock_acquire;
-	ofi_mutex_unlock_t	lock_release;
+	struct ofi_genlock	lock;
 
 	struct ofi_bitmask	*coll_cid_mask;
 	struct slist		coll_ready_queue;
@@ -326,22 +324,6 @@ ofi_ep_fid_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
 {
 	return ofi_ep_bind(container_of(ep_fid, struct util_ep, ep_fid.fid),
 			   bfid, flags);
-}
-
-static inline void ofi_ep_lock_acquire(struct util_ep *ep)
-{
-	ep->lock_acquire(&ep->lock);
-}
-
-static inline void ofi_ep_lock_release(struct util_ep *ep)
-{
-	ep->lock_release(&ep->lock);
-}
-
-static inline bool ofi_ep_lock_held(struct util_ep *ep)
-{
-	return (ep->lock_acquire == ofi_mutex_lock_noop) ||
-		ofi_mutex_held(&ep->lock);
 }
 
 static inline void ofi_ep_tx_cntr_inc(struct util_ep *ep)
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -199,6 +199,7 @@ struct util_domain {
 	struct dlist_entry	list_entry;
 	struct util_fabric	*fabric;
 	struct util_eq		*eq;
+	struct fid_peer_srx	*srx;
 
 	struct ofi_genlock	lock;
 	ofi_atomic32_t		ref;
@@ -1215,7 +1216,82 @@ enum {
 	OFI_OPT_TCP_FI_ADDR = -FI_PROV_SPECIFIC_TCP
 };
 
+struct util_rx_entry {
+	struct fi_peer_rx_entry	peer_entry;
+	uint64_t		seq_no;
+	uint64_t		ignore;
+	int			multi_recv_ref;
+	/* extra memory allocated at the end of each entry to hold iovecs and
+	 * MR descriptors. The amount of memory is determined by the provider's
+	 * iov limit.
+	 * struct iovec iov[]
+	 * void *desc[]
+	 */
+};
 
+struct util_srx_ctx;
+
+typedef void(*ofi_update_func_t)(struct util_srx_ctx *srx,
+				 struct util_rx_entry *rx_entry);
+
+struct util_unexp_peer {
+	struct dlist_entry	entry;
+	struct slist		msg_queue;
+	struct slist		tag_queue;
+	int			cnt;
+};
+
+struct util_srx_ctx {
+	struct fid_peer_srx	peer_srx;
+	bool			dir_recv;
+	size_t			min_multi_recv_size;
+	uint64_t		rx_op_flags;
+	uint64_t		rx_msg_flags;
+	size_t			iov_limit;
+	ofi_update_func_t	update_func;
+
+	struct util_cq		*cq;
+
+	uint64_t		rx_seq_no;
+	struct slist		msg_queue;
+	struct slist		tag_queue;
+	struct ofi_dyn_arr	src_recv_queues;
+	struct ofi_dyn_arr	src_trecv_queues;
+
+	struct dlist_entry	unspec_unexp_msg_queue;
+	struct dlist_entry	unspec_unexp_tag_queue;
+
+	struct dlist_entry	unexp_peers;
+	struct ofi_dyn_arr	src_unexp_peers;
+
+	struct ofi_bufpool	*rx_pool;
+	struct ofi_genlock	*lock;
+};
+
+struct util_match_attr {
+	fi_addr_t	addr;
+	uint64_t	tag;
+	uint64_t	ignore;
+};
+
+static inline struct fid_peer_srx *util_get_peer_srx(struct fid_ep *rx_ep)
+{
+	return (struct fid_peer_srx *) rx_ep->fid.context;
+}
+
+int util_ep_srx_context(struct util_domain *domain, size_t rx_size,
+			size_t iov_limit, size_t default_min_mr,
+			ofi_update_func_t update_func,
+			struct ofi_genlock *lock, struct fid_ep **rx_ep);
+int util_srx_close(struct fid *fid);
+int util_srx_bind(struct fid *fid, struct fid *bfid, uint64_t flags);
+ssize_t util_srx_generic_recv(struct fid_ep *ep_fid, const struct iovec *iov,
+			      void **desc, size_t iov_count, fi_addr_t addr,
+			      void *context, uint64_t flags);
+ssize_t util_srx_generic_trecv(struct fid_ep *ep_fid, const struct iovec *iov,
+			       void **desc, size_t iov_count, fi_addr_t addr,
+			       void *context, uint64_t tag, uint64_t ignore,
+			       uint64_t flags);
 #ifdef __cplusplus
 }
 #endif
--- a/man/fi_peer.3.md
+++ b/man/fi_peer.3.md
@@ -415,6 +415,20 @@ struct fi_peer_srx_context {
 The ownership of structure field values and callback functions is similar
 to those defined for peer CQs, relative to owner versus peer ops.
 
+The owner is responsible for acquiring any necessary locks before anything that
+could result in peer callbacks.
+The following functions are progress level functions:
+get_msg(), get_tag(), queue_msg(), queue_tag(), free_entry(), start_msg(),
+start_tag(), discard_msg(), discard_tag(). If needed, it is the owner's
+responsibility to acquire the appropriate lock prior to calling into a peer's
+fi_cq_read(), or similar, function that drives progress.
+
+The following functions are domain level functions:
+foreach_unspec_addr(). This function is used outside of message progress flow
+(i.e. during fi_av_insert()). The owner of the srx is responsible for acquiring
+the same lock, if needed.
+
+
 ## fi_peer_rx_entry
 
 fi_peer_rx_entry defines a common receive entry for use between the owner and
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -110,17 +110,6 @@ int smr_query_atomic(struct fid_domain *domain, enum fi_datatype datatype,
 
 #define SMR_IOV_LIMIT		4
 
-struct smr_rx_entry {
-	struct fi_peer_rx_entry	peer_entry;
-	struct iovec		iov[SMR_IOV_LIMIT];
-	void			*desc[SMR_IOV_LIMIT];
-	int64_t			peer_id;
-	uint64_t		ignore;
-	int			multi_recv_ref;
-	uint64_t		err;
-	struct ofi_mr		*mr[SMR_IOV_LIMIT];
-};
-
 struct smr_tx_entry {
 	struct smr_cmd	cmd;
 	int64_t		peer_id;
@@ -150,37 +139,14 @@ struct smr_pend_entry {
 	ofi_hmem_async_event_t	async_event;
 };
 
-struct smr_match_attr {
-	fi_addr_t	id;
-	uint64_t	tag;
-	uint64_t	ignore;
-};
-
-static inline int smr_match_id(fi_addr_t addr, fi_addr_t match_addr)
-{
-	return (addr == FI_ADDR_UNSPEC) || (match_addr == FI_ADDR_UNSPEC) ||
-		(addr == match_addr);
-}
-
-static inline int smr_match_tag(uint64_t tag, uint64_t ignore, uint64_t match_tag)
-{
-	return ((tag | ignore) == (match_tag | ignore));
-}
-
 struct smr_cmd_ctx {
 	struct dlist_entry entry;
 	struct smr_ep *ep;
 	struct smr_cmd cmd;
 };
 
-OFI_DECLARE_FREESTACK(struct smr_rx_entry, smr_recv_fs);
 OFI_DECLARE_FREESTACK(struct smr_tx_entry, smr_tx_fs);
 OFI_DECLARE_FREESTACK(struct smr_pend_entry, smr_pend_fs);
-
-struct smr_queue {
-	struct dlist_entry list;
-	dlist_func_t *match_func;
-};
 
 struct smr_fabric {
 	struct util_fabric	util_fabric;
@@ -242,24 +208,6 @@ struct smr_sock_info {
 	struct smr_cmap_entry	peers[SMR_MAX_PEERS];
 };
 
-struct smr_srx_ctx {
-	struct fid_peer_srx	peer_srx;
-	struct smr_queue	recv_queue;
-	struct smr_queue	trecv_queue;
-	bool			dir_recv;
-	size_t			min_multi_recv_size;
-	uint64_t		rx_op_flags;
-	uint64_t		rx_msg_flags;
-
-	struct util_cq		*cq;
-	struct smr_queue	unexp_msg_queue;
-	struct smr_queue	unexp_tagged_queue;
-	struct smr_recv_fs	*recv_fs;
-	ofi_spin_t		lock;
-};
-
-struct smr_rx_entry *smr_alloc_rx_entry(struct smr_srx_ctx *srx);
-
 struct smr_ep {
 	struct util_ep		util_ep;
 	size_t			tx_size;
@@ -282,11 +230,6 @@ struct smr_ep {
 	struct smr_sock_info	*sock_info;
 	void			*dsa_context;
 };
-
-static inline struct smr_srx_ctx *smr_get_smr_srx(struct smr_ep *ep)
-{
-	return (struct smr_srx_ctx *) ep->srx->fid.context;
-}
 
 static inline struct fid_peer_srx *smr_get_peer_srx(struct smr_ep *ep)
 {
@@ -357,16 +300,6 @@ static inline uint64_t smr_rx_cq_flags(uint32_t op, uint64_t rx_flags,
 		rx_flags |= FI_REMOTE_CQ_DATA;
 	return rx_flags;
 }
-
-
-bool smr_adjust_multi_recv(struct smr_srx_ctx *srx,
-			   struct fi_peer_rx_entry *rx_entry, size_t len);
-void smr_init_rx_entry(struct smr_rx_entry *entry, const struct iovec *iov,
-		       void **desc, size_t count, fi_addr_t addr,
-		       void *context, uint64_t tag, uint64_t flags);
-struct smr_rx_entry *smr_get_recv_entry(struct smr_srx_ctx *srx,
-		const struct iovec *iov, void **desc, size_t count, fi_addr_t addr,
-		void *context, uint64_t tag, uint64_t ignore, uint64_t flags);
 
 void smr_ep_progress(struct util_ep *util_ep);
 
--- a/prov/shm/src/smr_av.c
+++ b/prov/shm/src/smr_av.c
@@ -50,6 +50,15 @@ static int smr_av_close(struct fid *fid)
 	return 0;
 }
 
+
+static fi_addr_t smr_get_addr(struct fi_peer_rx_entry *rx_entry)
+{
+	struct smr_cmd_ctx *cmd_ctx = rx_entry->peer_context;
+
+	return cmd_ctx->ep->region->map->peers[cmd_ctx->cmd.msg.hdr.id].fiaddr;
+}
+
+
 /*
  * Input address: smr name (string)
  * output address: index (fi_addr_t), the output from util_av
@@ -61,6 +70,7 @@ static int smr_av_insert(struct fid_av *av_fid, const void *addr, size_t count,
 	struct util_ep *util_ep;
 	struct smr_av *smr_av;
 	struct smr_ep *smr_ep;
+	struct fid_peer_srx *srx;
 	struct dlist_entry *av_entry;
 	fi_addr_t util_addr;
 	int64_t shm_id = -1;
@@ -123,6 +133,8 @@ static int smr_av_insert(struct fid_av *av_fid, const void *addr, size_t count,
 			smr_map_to_endpoint(smr_ep->region, shm_id);
 			smr_ep->region->max_sar_buf_per_peer =
 				SMR_MAX_PEERS / smr_av->smr_map->num_peers;
+			srx = smr_get_peer_srx(smr_ep);
+			srx->owner_ops->foreach_unspec_addr(srx, &smr_get_addr);
 		}
 	}
 
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -43,8 +43,8 @@
 #include "smr.h"
 #include "smr_dsa.h"
 
-extern struct fi_ops_msg smr_msg_ops, smr_no_recv_msg_ops, smr_srx_msg_ops;
-extern struct fi_ops_tagged smr_tag_ops, smr_no_recv_tag_ops, smr_srx_tag_ops;
+extern struct fi_ops_msg smr_msg_ops, smr_no_recv_msg_ops;
+extern struct fi_ops_tagged smr_tag_ops, smr_no_recv_tag_ops;
 extern struct fi_ops_rma smr_rma_ops;
 extern struct fi_ops_atomic smr_atomic_ops;
 DEFINE_LIST(sock_name_list);
@@ -112,32 +112,29 @@ static struct fi_ops_cm smr_cm_ops = {
 	.shutdown = fi_no_shutdown,
 };
 
-int smr_getopt(fid_t fid, int level, int optname,
-	       void *optval, size_t *optlen)
+int smr_ep_getopt(fid_t fid, int level, int optname, void *optval,
+		  size_t *optlen)
 {
 	struct smr_ep *smr_ep =
 		container_of(fid, struct smr_ep, util_ep.ep_fid);
 
-	if ((level != FI_OPT_ENDPOINT) || (optname != FI_OPT_MIN_MULTI_RECV))
-		return -FI_ENOPROTOOPT;
-
-	*(size_t *)optval = smr_get_smr_srx(smr_ep)->min_multi_recv_size;
-	*optlen = sizeof(size_t);
-
-	return FI_SUCCESS;
+	return smr_ep->srx->ops->getopt(&smr_ep->srx->fid, level, optname,
+					optval, optlen);
 }
 
-int smr_setopt(fid_t fid, int level, int optname,
-	       const void *optval, size_t optlen)
+int smr_ep_setopt(fid_t fid, int level, int optname, const void *optval,
+		  size_t optlen)
 {
 	struct smr_ep *smr_ep =
 		container_of(fid, struct smr_ep, util_ep.ep_fid);
+	struct util_srx_ctx *srx;
 
 	if (level != FI_OPT_ENDPOINT)
 		return -FI_ENOPROTOOPT;
 
 	if (optname == FI_OPT_MIN_MULTI_RECV) {
-		smr_get_smr_srx(smr_ep)->min_multi_recv_size = *(size_t *)optval;
+		srx = util_get_peer_srx(smr_ep->srx)->ep_fid.fid.context;
+		srx->min_multi_recv_size = *(size_t *)optval;
 		return FI_SUCCESS;
 	}
 
@@ -156,61 +153,19 @@ int smr_setopt(fid_t fid, int level, int optname,
 	return -FI_ENOPROTOOPT;
 }
 
-static int smr_match_recv_ctx(struct dlist_entry *item, const void *args)
-{
-	struct smr_rx_entry *pending_recv;
-
-	pending_recv = container_of(item, struct smr_rx_entry, peer_entry);
-	return pending_recv->peer_entry.context == args;
-}
-
-static int smr_ep_cancel_recv(struct smr_ep *ep, struct smr_queue *queue,
-			      void *context, uint32_t op)
-{
-	struct smr_srx_ctx *srx = smr_get_smr_srx(ep);
-	struct smr_rx_entry *recv_entry;
-	struct dlist_entry *entry;
-	int ret = 0;
-
-	ofi_spin_lock(&srx->lock);
-	entry = dlist_remove_first_match(&queue->list, smr_match_recv_ctx,
-					 context);
-	if (entry) {
-		recv_entry = container_of(entry, struct smr_rx_entry, peer_entry);
-		ret = smr_write_err_comp(ep->util_ep.rx_cq,
-			recv_entry->peer_entry.context,
-			smr_rx_cq_flags(op, recv_entry->peer_entry.flags, 0),
-			recv_entry->peer_entry.tag, FI_ECANCELED);
-		ofi_freestack_push(srx->recv_fs, recv_entry);
-		ret = ret ? ret : 1;
-	}
-
-	ofi_spin_unlock(&srx->lock);
-	return ret;
-}
-
 static ssize_t smr_ep_cancel(fid_t ep_fid, void *context)
 {
 	struct smr_ep *ep;
-	int ret;
 
 	ep = container_of(ep_fid, struct smr_ep, util_ep.ep_fid);
-
-	ret = smr_ep_cancel_recv(ep, &smr_get_smr_srx(ep)->trecv_queue, context,
-				 ofi_op_tagged);
-	if (ret)
-		return (ret < 0) ? ret : 0;
-
-	ret = smr_ep_cancel_recv(ep, &smr_get_smr_srx(ep)->recv_queue, context,
-				 ofi_op_msg);
-	return (ret < 0) ? ret : 0;
+	return ep->srx->ops->cancel(&ep->srx->fid, context);
 }
 
 static struct fi_ops_ep smr_ep_ops = {
 	.size = sizeof(struct fi_ops_ep),
 	.cancel = smr_ep_cancel,
-	.getopt = smr_getopt,
-	.setopt = smr_setopt,
+	.getopt = smr_ep_getopt,
+	.setopt = smr_ep_setopt,
 	.tx_ctx = fi_no_tx_ctx,
 	.rx_ctx = fi_no_rx_ctx,
 	.rx_size_left = fi_no_rx_size_left,
@@ -274,33 +229,6 @@ int64_t smr_verify_peer(struct smr_ep *ep, fi_addr_t fi_addr)
 	smr_send_name(ep, id);
 
 	return -1;
-}
-
-static int smr_match_msg(struct dlist_entry *item, const void *args)
-{
-	struct smr_match_attr *attr = (struct smr_match_attr *)args;
-	struct smr_rx_entry *recv_entry;
-
-	recv_entry = container_of(item, struct smr_rx_entry, peer_entry);
-	return smr_match_id(recv_entry->peer_entry.addr, attr->id);
-}
-
-static int smr_match_tagged(struct dlist_entry *item, const void *args)
-{
-	struct smr_match_attr *attr = (struct smr_match_attr *)args;
-	struct smr_rx_entry *recv_entry;
-
-	recv_entry = container_of(item, struct smr_rx_entry, peer_entry);
-	return smr_match_id(recv_entry->peer_entry.addr, attr->id) &&
-	       smr_match_tag(recv_entry->peer_entry.tag, recv_entry->ignore,
-			     attr->tag);
-}
-
-static void smr_init_queue(struct smr_queue *queue,
-			   dlist_func_t *match_func)
-{
-	dlist_init(&queue->list);
-	queue->match_func = match_func;
 }
 
 void smr_format_pend_resp(struct smr_tx_entry *pend, struct smr_cmd *cmd,
@@ -863,79 +791,6 @@ static void smr_cleanup_epoll(struct smr_sock_info *sock_info)
 	ofi_epoll_close(sock_info->epollfd);
 }
 
-int smr_srx_bind(struct fid *fid, struct fid *bfid, uint64_t flags)
-{
-	struct smr_srx_ctx *srx;
-
-	if (flags != FI_RECV || bfid->fclass != FI_CLASS_CQ)
-		return -FI_EINVAL;
-
-	srx = container_of(fid, struct smr_srx_ctx, peer_srx.ep_fid.fid);
-	srx->cq = container_of(bfid, struct util_cq, cq_fid.fid);
-	ofi_atomic_inc32(&srx->cq->ref);
-	return FI_SUCCESS;
-}
-
-static void smr_close_recv_queue(struct smr_srx_ctx *srx,
-				 struct smr_queue *recv_queue)
-{
-	struct fi_cq_err_entry err_entry;
-	struct smr_rx_entry *rx_entry;
-	int ret;
-
-	while (!dlist_empty(&recv_queue->list)) {
-		dlist_pop_front(&recv_queue->list, struct smr_rx_entry,
-				rx_entry, peer_entry);
-
-		memset(&err_entry, 0, sizeof err_entry);
-		err_entry.op_context = rx_entry->peer_entry.context;
-		err_entry.flags = rx_entry->peer_entry.flags;
-		err_entry.tag = rx_entry->peer_entry.tag;
-		err_entry.err = FI_ECANCELED;
-		err_entry.prov_errno = -FI_ECANCELED;
-		ret = srx->cq->peer_cq->owner_ops->writeerr(srx->cq->peer_cq, &err_entry);
-		if (ret)
-			FI_WARN(&smr_prov, FI_LOG_EP_CTRL,
-				"Error writing recv entry error to rx cq\n");
-
-		ofi_freestack_push(srx->recv_fs, rx_entry);
-	}
-}
-
-static int smr_srx_close(struct fid *fid)
-{
-	struct smr_srx_ctx *srx;
-	struct smr_rx_entry *rx_entry;
-
-	srx = container_of(fid, struct smr_srx_ctx, peer_srx.ep_fid.fid);
-	if (!srx)
-		return -FI_EINVAL;
-
-	smr_close_recv_queue(srx, &srx->recv_queue);
-	smr_close_recv_queue(srx, &srx->trecv_queue);
-
-	while (!dlist_empty(&srx->unexp_msg_queue.list)) {
-		dlist_pop_front(&srx->unexp_msg_queue.list, struct smr_rx_entry,
-				rx_entry, peer_entry);
-		rx_entry->peer_entry.srx->peer_ops->discard_msg(
-							&rx_entry->peer_entry);
-	}
-
-	while (!dlist_empty(&srx->unexp_tagged_queue.list)) {
-		dlist_pop_front(&srx->unexp_tagged_queue.list,
-				struct smr_rx_entry, rx_entry, peer_entry);
-		rx_entry->peer_entry.srx->peer_ops->discard_tag(
-							&rx_entry->peer_entry);
-	}
-
-	ofi_atomic_dec32(&srx->cq->ref);
-	smr_recv_fs_free(srx->recv_fs);
-	ofi_spin_destroy(&srx->lock);
-	free(srx);
-
-	return FI_SUCCESS;
-}
-
 static int smr_ep_close(struct fid *fid)
 {
 	struct smr_ep *ep;
@@ -954,13 +809,13 @@ static int smr_ep_close(struct fid *fid)
 		free(ep->sock_info);
 	}
 
+	if (ep->srx && ep->util_ep.ep_fid.msg != &smr_no_recv_msg_ops)
+		(void) util_srx_close(&ep->srx->fid);
+
 	ofi_endpoint_close(&ep->util_ep);
 
 	if (ep->region)
 		smr_free(ep->region);
-
-	if (ep->util_ep.ep_fid.msg != &smr_no_recv_msg_ops)
-		smr_srx_close(&ep->srx->fid);
 
 	if (ep->cmd_ctx_pool)
 		ofi_bufpool_destroy(ep->cmd_ctx_pool);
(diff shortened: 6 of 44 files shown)


a-szegel:
This changes adds 20ns of latency to 0 byte sm2 pt2pt latency via fabtests.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1121 @@
...
+	slist_init((struct slist *) item);
+}
+
+static void util_srx_init_dlist(struct ofi_dyn_arr *arr, void *item)
+{
+	dlist_init((struct dlist_entry *) item);
+}
+
+int util_ep_srx_context(struct util_domain *domain, size_t rx_size,
+			size_t iov_limit, size_t default_min_mr,
```
shijin-aws, line 1048: nit: I realize `default_min_mr` is actually for multi recv, but the name makes me think of it's memory registration

shijin-aws:
Generally LGTM, I am rebasing my EFA change on this PR. Will chime in when I have update.

shijin-aws:
Can we add FI_PEEK/FI_CLAIM support to util srx? I know shm does not support it yet. But if we are going to make it common we may need to add them. I can work on it if you would like to have them in a separate PR.

shijin-aws:
I currently had a blocker to use `util_srx_generic_recv` in efa provider. I opened https://github.com/ofiwg/libfabric/issues/8936 for discussion.

prov/shm/src/smr.h
```diff
@@ -150,21 +139,22 @@ struct smr_pend_entry {
...
+	return ((struct ofi_mr *) *desc)->iface;
 }
 
-static inline int smr_match_tag(uint64_t tag, uint64_t ignore, uint64_t match_tag)
+static inline uint64_t smr_get_mr_flags(void **desc)
 {
-	return ((tag | ignore) == (match_tag | ignore));
+	assert(desc && *desc);
+	return ((struct ofi_mr *) *desc)->flags;
 }
```
wenduwan, line 158: Where are they used?

prov/shm/src/smr_ep.c
```diff
@@ -1337,266 +1193,34 @@ static void smr_init_ipc_socket(struct smr_ep *ep)
...
+int smr_progress_addr(struct fi_peer_rx_entry *rx_entry)
 {
 	struct smr_cmd_ctx *cmd_ctx = rx_entry->peer_context;
 
-	ofi_buf_free(cmd_ctx);
-	return FI_SUCCESS;
+	rx_entry->addr =
+		cmd_ctx->ep->region->map->peers[cmd_ctx->cmd.msg.hdr.id].fiaddr;
+
+	return rx_entry->addr == FI_ADDR_UNSPEC ? -FI_EAGAIN: FI_SUCCESS;
```
wenduwan, line 1211: nit missing space around :

prov/shm/src/smr_ep.c
```diff
@@ -1337,266 +1193,34 @@ static void smr_init_ipc_socket(struct smr_ep *ep)
...
-	.get_tag = smr_get_tag,
-	.queue_msg = smr_queue_msg,
-	.queue_tag = smr_queue_tag,
-	.free_entry = smr_free_entry,
-};
-
-static int smr_discard(struct fi_peer_rx_entry *rx_entry)
+int smr_progress_addr(struct fi_peer_rx_entry *rx_entry)
 {
 	struct smr_cmd_ctx *cmd_ctx = rx_entry->peer_context;
```
wenduwan, line 1206: nit allocating cmd_ctx seems a little redundant

prov/shm/src/smr_ep.c
```diff
@@ -1698,16 +1324,25 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			domain = container_of(ep->util_ep.domain,
 					      struct smr_domain,
 					      util_domain.domain_fid);
-			ret = smr_ep_srx_context(domain, ep->rx_size,
-						 &ep->srx);
+			ret = util_ep_srx_context(&domain->util_domain,
+					ep->rx_size, SMR_IOV_LIMIT,
+					SMR_INJECT_SIZE, &ep->srx);
```
wenduwan, line 1329: oops alignment. same on L1334-L1336, and many other places...

prov/sm2/src/sm2.h
```diff
@@ -152,16 +152,22 @@ int sm2_eq_open(struct fid_fabric *fabric, struct fi_eq_attr *attr,
...
 }
 
-static inline int sm2_match_tag(uint64_t tag, uint64_t ignore,
-				uint64_t match_tag)
+static inline uint64_t sm2_get_mr_flags(void **desc)
 {
-	return ((tag | ignore) == (match_tag | ignore));
+	assert(desc && *desc);
+	return ((struct ofi_mr *) *desc)->flags;
 }
```
wenduwan, line 171: Same question - where are they used?

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1121 @@
...
+	struct util_rx_entry *util_entry, *owner_entry;
+
+	srx = (struct util_srx_ctx *) entry->srx->ep_fid.fid.context;
+
+	srx->lock_acquire(&srx->lock);
+	util_entry = container_of(entry, struct util_rx_entry, peer_entry);
+	if (entry->owner_context) {
+		owner_entry = (struct util_rx_entry *) entry->owner_context;
+		if (!--owner_entry->multi_recv_ref &&
+		    owner_entry->peer_entry.size < srx->min_multi_recv_size) {
```
wenduwan, line 415: Wow this is testing my knowledge of associative rules.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1121 @@
...
+}
+
+static int util_match_tag(struct fid_peer_srx *srx, fi_addr_t addr,
+			  size_t size, uint64_t tag,
+			  struct fi_peer_rx_entry **rx_entry)
+{
+	struct util_srx_ctx *srx_ctx;
+	struct util_rx_entry *util_entry;
+	struct slist_entry *item, *prev;
+	int ret = FI_SUCCESS;
```
wenduwan, line 250: I guess this variable is redundant.

wenduwan (commented):
Overall looking good, based on my limited understanding of srx.

Many formatting itches but that's nit.

Mostly concerned with
1. Perf impact according to @a-szegel 
2. MR descriptor translation between owner and peer(I assume it will be addressed separately)

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1121 @@
...
+
+	dlist_init(&srx->all_unexp_msg);
+	dlist_init(&srx->all_unexp_tag);
+
+	dlist_init(&srx->progress_msg_queue);
+	dlist_init(&srx->progress_tag_queue);
+	srx->rx_seq_no = 0;
+
+	pool_attr.size = sizeof(struct util_rx_entry) +
+		(sizeof(struct iovec) + sizeof(void *)) * iov_limit;
```
shijin-aws, line 1089: Nit: I would add comment here to explain the math of `(sizeof(struct iovec) + sizeof(void *)) * iov_limit`. I understand it is needed to count the size of `iov` and `desc`.

 In the future it may be better to make util_rx_entry have a `data` field so providers can embed additional context inside it, like the util_av_entry today. But it does not have to be part of this PR.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1121 @@
...
+static int util_get_msg(struct fid_peer_srx *srx, fi_addr_t addr,
+		        size_t size, struct fi_peer_rx_entry **rx_entry)
+{
+	struct util_srx_ctx *srx_ctx;
+	struct util_rx_entry *util_entry, *any_entry;
+	struct slist *queue;
+	int ret;
+
+	srx_ctx = srx->ep_fid.fid.context;
+	srx_ctx->lock_acquire(&srx_ctx->lock);
```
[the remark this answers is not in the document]
shijin-aws: And I think there is a pitfall here... @wzamazon has pointed out in https://github.com/ofiwg/libfabric/commit/ab529c244a8ca74c0327e88c3baf1226e051f345 that `get_msg` and `queue_msg` must be serialized... which means you cannot release the lock after `get_msg` and before `queue_msg`, in case `generic_recv` will chime in and cause deadlock.

Example:
thread 1 call get_msg, did not find rx_entry in rx_list
thread 2 call fi_recv, put rx_entry to rx_list
thread 1 call queue_msg, put a different rx_entry to unexpected_rx_list

When thread 2 post the recv, it cannot find the unexp rx entry, because it was not queued yet. Then no thread can pick up that unexp rx entry

aingerson:
Thanks for the review, Wen! I am currently working on reworking the changes to include the change to update_addr (see other PR for that discussion), add the MR update function, and add peek/claim/discard support. I will try to include you comments in the revision but assume that it will change significantly :)

prov/util/src/util_srx.c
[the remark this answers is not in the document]
aingerson: @wenduwan The main locking change is [this](https://github.com/aingerson/libfabric/blob/peer/prov/util/src/util_srx.c#L1060-L1066) which turns the locking into a no-op. 

@shijin-aws Going to have to think about that example some more. I don't like the idea of not releasing the lock and relying on the peer to call queue properly. If it errors out during context allocation, we would need a different call just for the peer to release the lock without queueing. We will need some protection against that but I don't think playing with the locking is the safest choice. Let me weigh the options here a bit... Thanks for bringing this up.

prov/shm/src/smr.h
aingerson: I think this is a weird git-ism. This patch didn't touch this function, just deleted the matching functions around it and git got confused. But you're right this isn't needed anymore. It was left unused when we moved to passing the ofi_mr around instead of the desc to support the multiple iface iovs. It was used in what now looks like this:
```
		use_ipc = ofi_hmem_is_ipc_enabled(iface) &&
				smr_desc->flags & FI_HMEM_DEVICE_ONLY &&
				!(op_flags & FI_INJECT);
```
I will add a patch to remove it. Thanks!

aingerson: Ha nevermind. That was a me-ism from not resolving a conflict correctly after rebasing 😅

prov/shm/src/smr_ep.c
aingerson: What do you mean?

prov/shm/src/smr_ep.c
aingerson: I usually don't align with the parantheses with functions that are either long or indented a lot to reduce line count

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1121 @@
...
+			     context))
+		return 0;
+
+	if (ofi_array_iter(&srx->src_trecv_queues, context, util_cancel_src))
+		return 0;
+
+	if (ofi_array_iter(&srx->src_recv_queues, context, util_cancel_src))
+		return 0;
+
+	return 0;
```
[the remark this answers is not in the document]
aingerson: Yeah the API is confusing to me for the cancel call - return 0 just means it was successfully submitted (whether it was found or not). The array iter return means it was found in that list so we can stop searching the rest of the lists. I'm updating it to FI_SUCCESS though instead of 0

prov/shm/src/smr_ep.c
wenduwan: I meant this is a one-line function
```
rx_entry->addr =
              rx_entry->peer_context->ep->region->map->peers[cmd_ctx->cmd.msg.hdr.id].fiaddr;
```

Also just noticed the signature
```
static void smr_get_addr(...)
```
IMO either it should return the address, or we call it 
```
static void smr_update_addr(...)
```

aingerson: Ah I understand what you mean. It's just a pointer, not allocation, and I thought setting it helped with readability since the line is already quite a doozy... and if I used it inline I would need a cast so the whole line would look like this:
```
rx_entry->addr = rx_entry->peer_context->ep->region->map->peers[((struct smr_cmd_ctx *) rx_entry->peer_context)->cmd.msg.hdr.id].fiaddr;
```

I get your issue with the update/get_addr naming. I feel pretty neutral about the names but I'll let Sean chime in about preferences.

@shefty, do you have any naming preferences for the API or return/setting?
Options could be to rename the owner op to update_addrs() to make it clearer that we are updating a list and/or set the peer op to update_addr() as well to indicate it will be changing the value.
Or get_addr could return the addr and the owner could set it to the return?

wenduwan: > quite a doozy

Indeed. Didn't realize the array index was also that long. Then let's keep it as is.

include/ofi_util.h
```diff
@@ -1215,7 +1216,73 @@ enum {
...
+	struct fi_peer_rx_entry	peer_entry;//entry used for per-peer/progress
+	struct dlist_entry	entry;//entry used for all list
+	uint64_t		seq_no;
+	uint64_t		ignore;
+	int			multi_recv_ref;
+};
+
+struct util_srx_ctx;
+
+typedef void(*ofi_update_func_t)(struct util_srx_ctx *srx,
```
shijin-aws, line 1229: Thanks for implementing that!

aingerson:
Updated with foreach looping and passing in function pointer. Included in the revision are the MR update specification as well as claim/discard/peek support. Only outstanding issue is the get/queue/receive issue but that is not an issue with the util_srx implementation, rather an issue with the API itself. Still thinking about it....
Should be ready for another round of testing/review!

man/fi_peer.3.md
```diff
@@ -387,6 +387,9 @@ struct fi_ops_srx_owner {
                    size_t size, uint64_t tag, struct fi_peer_rx_entry **entry);
     int (*queue_msg)(struct fi_peer_rx_entry *entry);
     int (*queue_tag)(struct fi_peer_rx_entry *entry);
+	 void	(*foreach_unspec_addr)(struct fid_peer_srx *srx,
```
shijin-aws, line 390: indent is off

aingerson: Thansk! Fixed it in PR #8906 which has the API change commits. Will rebase here once that is merged.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1155 @@
...
+			if (!util_entry) {
+				FI_WARN(&core_prov, FI_LOG_EP_CTRL,
+					"cannot allocate multi receive "
+					"buffer\n");
+				return -FI_ENOMEM;
+			}
+		} else {
+			(void) slist_remove_head(&srx_ctx->msg_queue);
+		}
+		srx_ctx->update_func(srx_ctx, util_entry);
```
shijin-aws, line 186: Do you need to make this call after L188? provider (efa) may use `util_entry->peer_entry.srx`  inside `update_func`

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1155 @@
...
+			FI_WARN(&core_prov, FI_LOG_EP_CTRL,
+				"cannot allocate multi receive buffer\n");
+			ret = -FI_ENOMEM;
+			goto out;
+		}
+	} else {
+		(void) slist_remove_head(queue);
+	}
+
+	srx_ctx->update_func(srx_ctx, util_entry);
```
shijin-aws, line 236: Do you need to make this call after `util_entry->peer_entry.srx = srx;`? provider (efa) may use util_entry->peer_entry.srx inside update_func

aingerson: oh sure, good call. Thanks!

shijin-aws: Similarly for other places.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1155 @@
...
+
+	util_entry = util_alloc_rx_entry(srx);
+	if (!util_entry)
+		return NULL;
+
+	util_entry->peer_entry.owner_context = NULL;
+	util_entry->peer_entry.size = size;
+	util_entry->peer_entry.addr = addr;
+	util_entry->peer_entry.tag = tag;
+
```
shijin-aws, line 109: Should we also set `util_entry->peer_entry.desc = NULL` here? Not a bug in this PR, but there was some bug in my development which was partly due to using an unexp entry with garbage (non-null) desc.

aingerson: We can if you like but I do think it's a waste since the unexpected message isn't attached to a receive yet and shouldn't have any desc attached. The peer should never access the desc. It is initialized correctly initialized when it is actually matched to a posted receive, before the update mr function. If it is fixing an issue in efa I think that means efa is incorrectly accessing the desc when it shouldn't

shijin-aws: Makes sense. No bother then !

prov/util/src/util_srx.c
aingerson: @shijin-aws Thanks! I've fixed it locally. Was just waiting for the API PR to get merged before repushing to limit CI testing.

aingerson:
Updated with rebase + srx setting fix. Also added some locking updates.
@wenduwan @shijin-aws Sean and I spoke about the locking and have decided to rework the locking definition with owner and peers. Since the owner can acquire the appropriate lock when driving progress for the peer, we don't need any locking within the owner ops. This is because all of the owner ops are called on the peer cq_read path. So before calling into the peer's fi_cq_read to drive progress, the owner should acquire any locking necessary (the srx->lock still exists for this purpose). I left the srx locking in the receive path since it's not directly on the cq_read path and since we need to be locked for the start/discard paths. I added a patch to elaborate on this in the man pages
So the serialization issue you described shouldn't be an issue with this definition.
Let me know if this doesn't make sense.
@a-szegel Could you run this updated code to check for regressions, please?

shijin-aws:
>the owner should acquire any locking necessary (the srx->lock still exists for this purpose)

So owner has to acquire/release the lock at the beginning and ending of the ep_progress? I need to think about it more but it sounds reasonable to me. Currently efa already uses `util_ep.lock` to lock it, I see shm uses that as well... Need they also acquire srx->lock?

aingerson:
@shijin-aws Right, the owner has to acquire a specific lock that is tied to the certain function groups before calling fi_cq_read on the peer to drive progress. I left the srx->lock as an option for the owner to use it if it doesn't have a good other option but it could be whatever the owner wants, just consistent. You _might_ be able to use the ep lock but note that the srx is defined per domain so you could only use the ep lock if you force a 1:1 ratio of domain->EP. I guess that can be provider specific, but I would suggest using a domain level lock.
I think I'm going to redo the generic recv functions to let the caller do the locking instead of the util_srx (this code is assuming the user will be using the srx->lock) so ignore those. Basically the owner needs to do acquire the same lock before calling generic_recv as it does before calling fi_cq_read(shm)

prov/shm/src/smr_progress.c
```diff
@@ -890,11 +896,16 @@ static int smr_progress_cmd_msg(struct smr_ep *ep, struct smr_cmd *cmd)
 		ofi_ep_lock_release(&ep->util_ep);
 		if (ret == -FI_ENOENT) {
 			ret = smr_alloc_cmd_ctx(ep, rx_entry, cmd);
-			if (ret)
+			if (ret) {
```
a-szegel, line 899: We probably need these changes in SM2 as well

a-szegel (changes requested):

prov/shm/src/smr_progress.c
a-szegel: Here you go: https://github.com/aingerson/libfabric/pull/484!

shijin-aws:
>You might be able to use the ep lock but note that the srx is defined per domain so you could only use the ep lock if you force a 1:1 ratio of domain->EP. I guess that can be provider specific, but I would suggest using a domain level lock.

How does shm provider handle this currently? Is it using a domain level lock somewhere?

fabtests/test_configs/shm/shm.exclude
```diff
@@ -13,7 +13,6 @@ poll
 
 # Exclude tests with unsupported capabilities
 -k
-rdm_tagged_peek
 cm_data
```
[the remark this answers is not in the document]
shijin-aws: Yes, can you or alexia just have 1 more commit to add the test to fabtests/pytest/shm/test_rdm.py

Need to copy this https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/efa/test_rdm.py#L62-L69 to https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/shm/test_rdm.py

shijin-aws:
>Since the owner can acquire the appropriate lock when driving progress for the peer

I think efa currently does not acquire that lock for peer (shm) correctly. It just call `fi_cq_read(shm_cq, NULL, 0)` inside efa's `fi_cq_read`. We need to fix when rebasing on your latest PR.

shijin-aws:
> Since the owner can acquire the appropriate lock when driving progress for the peer

I am not sure this is optimal for performance. You are going to lock for a big chunk of function that mostly won't compete with owner unless the shared receive queues.

shefty:
Locking is dependent on the threading model specified by the application.  For single-threaded apps (MPI), no locking is ideal.  For multi-threaded apps, FI_THREAD_DOMAIN indicates that the app will not share domain level resources among threads.  So, a domain level lock, which can be disabled, is also ideal.  For other threading models, more fined grained locking might be possible, but I'm skeptical.

In any case, there is no race in calling get_msg() then queue_msg() that the peer provider cannot handle.  The needed serialization between get_msg() and queue_msg() is restricted to messages _which come from the same source address_.  If messages come from different source addresses and a posted receive buffer can match either, it doesn't matter which message the buffer is assigned.  If a second message arrives between the window of the first message calling get_msg() and queue_msg(), the second message can be treated as arriving at the same time as the first, making the matching arbitrary.

shijin-aws:
@aingerson 
> I guess that can be provider specific, but I would suggest using a domain level lock.

Maybe a dumb question. I understand srx is domain level resource. But how could I retrieve `srx` from `fi_cq_read` ? I guess you have to make `srx` part of provider's domain struct. That means the srx should have been created during domain open? This is kind of confusing to me as shm seems to create the srx during ep enable.

wzamazon:
> In any case, there is no race in calling get_msg() then queue_msg() that the peer provider cannot handle. The needed serialization between get_msg() and queue_msg() is restricted to messages which come from the same source address. If messages come from different source addresses and a posted receive buffer can match either, it doesn't matter which message the buffer is assigned. If a second message arrives between the window of the first message calling get_msg() and queue_msg(), the second message can be treated as arriving at the same time as the first, making the matching arbitrary.

The race condition mentioned happen between get_msg(), which read owner's receive list, and `fi_recv`, which write to owner's rx list.

The following an example:

There are two threads, thread A is running the progress engine, thread B is calling `fi_recv`. The following is the work flow:

1. thread A got an incoming message, thread A call `get_msg` which check the owner's receive list, and found no match.
2. thread B call `fi_recv`, which first check owner's unexpected message list, found no match, then add the receive info to owner's receive list
3. thread A call `queue_msg`, put the incoming message to unexpected message list.

At this point there is a deadlock because the fi_recv should have been matched with the message in the unexpected message list.

shefty:
queue_msg() needs to check if the message is matched as part of queuing it.  Or the owner provider can hold a lock between get_msg() and queue_msg(), which is also acquired as part of fi_recv().

shijin-aws:
@wzamazon Now the proposed solution is to make efa ep progress, shm ep progress, and fi_recv all under the same lock. So the get_msg/queue_msg will be serialized by nature because they are inside a ep_progress call, either shm or efa. 

And it is suggested that such lock should be domain level

shijin-aws:
> Or the owner provider can hold a lock between get_msg() and queue_msg()

I think this makes more sense to me. It's provider's responsibility to hold a lock when calling get_msg and queue_msg. And we can still use srx->lock here. IMO this is better having another lock around the whole cq read.

shijin-aws:
AWS CI failure (some of them)

- fi_bw data validation failure:
```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.32.215 'timeout 360 /bin/bash --login -c '"'"'/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr8907/install/fabtests/bin/fi_bw -e rdm -v -T 1 -p shm -s 172.31.32.215'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.32.215 'timeout 360 /bin/bash --login -c '"'"'/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr8907/install/fabtests/bin/fi_bw -e rdm -v -T 1 -p shm -s 172.31.32.215 172.31.32.215'"'"''
client_stdout:

client returncode: None
Client error: Client timed out
server_stdout:
Data check error (0!=f) at byte 0 for buffer size 64
```

- fi_rdm_atomic failure
```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.32.215 'timeout 360 /bin/bash --login -c '"'"'/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/bin/fi_rdm_atomic -I 5 -p shm -s 172.31.32.215'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.32.215 'timeout 360 /bin/bash --login -c '"'"'/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/bin/fi_rdm_atomic -I 5 -p shm -s 172.31.32.215 172.31.32.215'"'"''
client_stdout:
fi_rdm_atomic: ./include/ofi_mem.h:437: ofi_buf_free: Assertion `ofi_atomic_dec32(&ofi_buf_region(buf)->use_cnt) >= 0' failed.
backtrace (up to 30)
```

prov/shm/src/smr_progress.c
```diff
@@ -890,11 +896,16 @@ static int smr_progress_cmd_msg(struct smr_ep *ep, struct smr_cmd *cmd)
...
+				peer_srx->owner_ops->free_entry(rx_entry);
 				return ret;
+			}
 
 			ret = peer_srx->owner_ops->queue_msg(rx_entry);
-			goto out;
```
shijin-aws, line 897: This one shouldn't be removed :) You will make start_common called for ENOENT.

shijin-aws:
@aingerson I went ahead to make two shm changes in my PR https://github.com/ofiwg/libfabric/pull/8957 that is based on this PR. See https://github.com/ofiwg/libfabric/pull/8957/commits/c1bdf3d1ecd394f235de3ff690dc5f6c328f5622 and https://github.com/ofiwg/libfabric/pull/8957/commits/7d1ec6759d34413cc792ed91e281d29444dde5d8. The first one is to add the missing goto. The second one is to adding srx->lock in shm's progress_cmd_msg. I would appreciate your feedback for the second patch. I will drop/modify them after we getting consensus and this PR updated.

aingerson:
@shijin-aws Thanks for finding the goto issue! I will pull it into this PR.
As far as the locking goes, I don't think that will work as it assumed shm is the owner of the srx. Only the owner of the srx has access to the lock. I don't think the locking is an issue here. If shm is the owner, then it creates an srx within the EP so acquiring the ep lock is enough (which we already do). If you do want to change it, I would change it to the domain lock but I think that's overkill since there isn't a use case for it.

shijin-aws:
But the problem is when shm is used as a peer. The ep lock is different for efa and shm. You finally have to use a common (srx) lock. I don't like the idea of having a domain level lock around cq read that drives peer progress, because the SRX is only for msg/recv path. The cq read can be used for send/rma etc, we are finally introducing more locks in this direction

shijin-aws:
>If queue_msg always follows get_msg, would it be possible to only keep get_msg?

peer will do something before queue the message. That is main reason why we need to two APIs. But if we can extend the get_msg API to cover the procedure in the middle. We may merge them into one.

shefty:
The expected case is that get_msg() returns the buffer that the provider needs.  In that case, queue_msg() is never invoked.  It's only called when get_msg() can't find a matching buffer (unexpected message case).  The calls are setup for the optimal case.

aingerson:
> But the problem is when shm is used as a peer. The ep lock is different for owner and shm. You finally have to use a common (srx) lock. I don't like the idea of having a domain level lock around cq read that drives peer progress, because the SRX is only for msg/recv path. The cq read can be used for send/rma etc, we are finally introducing more locks in this direction

Most applications (like MPI) will set FI_THREAD_DOMAIN which means no threads will cross the domain boundary so this is a non issue. Even if we lock the the entire cq read call, it is not an issue because everything within the domain is serialized.

The issue you're talking about is when an application sets FI_THREAD_SAFE meaning they might have different threads calling into the same domain's resources. The CQ is a domain resource so it is not unreasonable that we would have to lock an entire CQ call, even though it is the peer's. I don't think the trade offs of constantly locking/unlocking the srx for other threads to grab entries is worth it. We would have to keep locking/unlocking over and over again all to maybe let another thread grab a receive buffer.

shijin-aws:
>Most applications (like MPI) will set FI_THREAD_DOMAIN which means no threads will cross the domain boundary so this is a non issue. Even if we lock the the entire cq read call, it is not an issue because everything within the domain is serialized.
The issue you're talking about is when an application sets FI_THREAD_SAFE meaning they might have different threads calling into the same domain's resources.

Open MPI will set FI_THREAD_SAFE for multi-threading case https://github.com/open-mpi/ompi/blob/main/ompi/mca/mtl/ofi/mtl_ofi_component.c#L645-L648.

The issue I am currently facing is running osu_latency_mt, which is exactly hitting this case.

shijin-aws:
>The CQ is a domain resource so it is not unreasonable that we would have to lock an entire CQ call, even though it is the peer's. I don't think the trade offs of constantly locking/unlocking the srx for other threads to grab entries is worth it. We would have to keep locking/unlocking over and over again all to maybe let another thread grab a receive buffer.

I see, I see srx->lock is a no-op for FI_THREAD_DOMAIN. Now I am less concerned to have a domain level lock around cq read. I will try it.

shefty:
If you lock at the domain level, that can support FI_THREAD_SAFE.  It's just that when FI_THREAD_DOMAIN is specified by the app, that lock can become a no-op.  If you instead lock individual objects at a lower-level, trying to disable those locks becomes nearly impossible.

shijin-aws:
One more question @aingerson . Since srx is a domain level resource, should we creates it during domain initialization rather than in ep stage? It looks currently shm only creates srx under smr_ep, which makes me confused for a while

aingerson:
@shijin-aws Yeah it's a bit confusing. The srx is created under the domain but its fid is a fid_ep in order for it to be passed into the recv API calls.
```
int fi_srx_context(struct fid_domain *domain,
    struct fi_rx_attr *attr, struct fid_ep **rx_ep,
    void *context);
```
So for shm, the peer srx implemention goes through the above call and so is a domain resource. The case where it is created under the EP is the case where no srx was opened so we just open a non-shared receive context. We happen to go through the util_srx implementation to make it easier to plug into the peer srx API but it is not actually a "shared" context so belongs to the EP only.

shijin-aws:
@aingerson but if application does not call `fi_srx_context`, how should owner provider create srx that is shared among providers? Just call `util_ep_srx_context` during domain open?

aingerson:
If a provider wants to share the srx between peers, it must open an srx with srx_open and bind it to the appropriate EPs. If srx_open/bind was not called on an EP, then the EP cannot use a shared context and needs to manage receives itself

prov/shm/src/smr_msg.c
```diff
@@ -228,46 +73,8 @@ static ssize_t smr_recv(struct fid_ep *ep_fid, void *buf, size_t len,
 	iov.iov_base = buf;
 	iov.iov_len = len;
 
-	return smr_generic_recv(smr_get_smr_srx(ep), &iov, &desc, 1, src_addr,
```
shijin-aws, line 231: Since you removed srx->lock in util_srx_generic_recv. Do you need a ep lock here to match the ep lock in progress engine?

shijin-aws:
bot:aws:retest

shijin-aws (approved):
LGTM. I need to rebase to see and test to see if I miss any rabbit holes.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1105 @@
...
+	.size = sizeof(struct fi_ops_msg),
+	.recv = util_srx_recv,
+	.recvv = util_srx_recvv,
+	.recvmsg = util_srx_recvmsg,
+	.send = fi_no_msg_send,
+	.sendv = fi_no_msg_sendv,
+	.sendmsg = fi_no_msg_sendmsg,
+	.inject = fi_no_msg_inject,
+	.senddata = fi_no_msg_senddata,
+	.injectdata = fi_no_msg_injectdata,
```
shefty, line 741: I don't understand how any of these can be exposed to a user above libfabric.  There's no locking provided anywhere

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1105 @@
...
+	.size = sizeof(struct fi_ops_tagged),
+	.recv = util_srx_trecv,
+	.recvv = util_srx_trecvv,
+	.recvmsg = util_srx_trecvmsg,
+	.send = fi_no_tagged_send,
+	.sendv = fi_no_tagged_sendv,
+	.sendmsg = fi_no_tagged_sendmsg,
+	.inject = fi_no_tagged_inject,
+	.senddata = fi_no_tagged_senddata,
+	.injectdata = fi_no_tagged_injectdata,
```
shefty, line 795: same as above

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1105 @@
...
+	return (int) util_cancel_recv(srx, queue, flags, context);
+}
+
+static ssize_t util_srx_cancel(fid_t ep_fid, void *context)
+{
+	struct util_srx_ctx *srx;
+
+	srx = container_of(ep_fid, struct util_srx_ctx, peer_srx.ep_fid);
+
+	srx->lock_acquire(&srx->lock);
```
shefty, line 949: Why is this the only place where a lock is used?

shefty (commented):
Submitting what little review I've done.  I don't understand the locking, or lack thereof, in the srx code.

wenduwan (approved):

prov/util/src/util_srx.c
shijin-aws: A relevant question: This is being called inside `smr_ep_cancel`, do we still need a ep lock in `smr_ep_cancel` if after we already acquire srx->lock? 

Similar question for generic_recv and ep_progress function. Currently efa acquire both srx lock and ep lock. I know they are different level of locking, so I tend to keep them both. 

@wenduwan  FYI for your comment in efa PR https://github.com/ofiwg/libfabric/pull/8957#discussion_r1210522298

aingerson:
Updated with new unexpected messaging method to reduce dlist usage while allowing for adequate fairness.
Also included appropriate locking.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1105 @@
...
+	slist_init((struct slist *) item);
+}
+
+static void util_srx_init_dlist(struct ofi_dyn_arr *arr, void *item)
+{
+	dlist_init((struct dlist_entry *) item);
+}
+
+int util_ep_srx_context(struct util_domain *domain, size_t rx_size,
+			size_t iov_limit, size_t default_min_multi_recv,
```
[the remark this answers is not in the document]
aingerson: Didn't add any other attributes since I think everything is in the domain already. Not checking MR mode but it is available if needed.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1155 @@
...
+static void util_init_rx_entry(struct util_rx_entry *entry,
+			       const struct iovec *iov, void **desc,
+			       size_t count, fi_addr_t addr, void *context,
+			       uint64_t tag, uint64_t flags)
+{
+	memcpy(entry->peer_entry.iov, iov, sizeof(*iov) * count);
+	if (desc)
+		memcpy(entry->peer_entry.desc, desc, sizeof(*desc) * count);
+	else
+		memset(entry->peer_entry.desc, 0, sizeof(*desc) * count);
```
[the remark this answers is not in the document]
aingerson: I kept the allocation because it could be that the owner provider does not need memory registration but its peer does so this keeps the memory there in case it uses a peer that does need it and needs to update the descriptors for the peer.

shijin-aws:
>Updated with new unexpected messaging method to reduce dlist usage while allowing for adequate fairness.

For your earlier revision, I was seeing worse (2x slower) performance after efa using util srx, on large scale (4608) MPI jobs. I need to do more profiling to figure out where the regression come from. Let me rebase on your current revision and try again

aingerson:
@shijin-aws Oh wow yeah please let me know how this new version looks and I can take a look if it's problematic. Thank you!

prov/shm/src/smr_ep.c
```diff
@@ -156,61 +153,19 @@ int smr_setopt(fid_t fid, int level, int optname,
...
-
-	ret = smr_ep_cancel_recv(ep, &smr_get_smr_srx(ep)->trecv_queue, context,
-				 ofi_op_tagged);
-	if (ret)
-		return (ret < 0) ? ret : 0;
-
-	ret = smr_ep_cancel_recv(ep, &smr_get_smr_srx(ep)->recv_queue, context,
-				 ofi_op_msg);
-	return (ret < 0) ? ret : 0;
+	return ep->srx->ops->cancel(&ep->srx->fid, context);
```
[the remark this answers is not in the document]
shijin-aws: similar for sm2

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1180 @@
...
+	srx = container_of(ep_fid, struct util_srx_ctx, peer_srx.ep_fid);
+
+	if (flags & FI_MULTI_RECV)
+		return util_generic_mrecv(srx, iov, desc, iov_count, addr,
+					  context, flags);
+
+	assert(iov_count <= srx->iov_limit);
+	addr = srx->dir_recv ? addr : FI_ADDR_UNSPEC;
+
+	ofi_genlock_lock(&srx->lock);
```
shijin-aws, line 721: just find you add this lock back in util recvs... I need to remove them in efa's recv accordingly

aingerson: Yeah sorry should have mentioned it. Now all the owner srx functions have the lock. I think we will have to restrict users of the util srx to use the srx lock for the peer locking since they are so reliant on each other.

shijin-aws: >Now all the owner srx functions have the lock.

I think you still remove them in owner/peer ops like `get_msg`, which is good as they should be inside the progress engine (cq read) and it's owner's responsibility to grab that lock before driving peers' ep progress.

shijin-aws:
bot:aws:retest

prov/util/src/util_srx.c
shijin-aws: I notice one caveat of having `srx->lock` inside util generic recv:

If the owner provider:
1. acquire a ep lock in its generic recv before it calls util recv
2. acquire a srx lock in its cq_read before calling it progress engine, which could acquire ep lock

There could be a deadlock in this situation, as thread A (generic recv) tries to acquire a srx lock held by thread B (cq read), while thread B tries to acquire a ep lock held by A.

I would have to either remove the ep lock in owner provider's generic recv, or move the srx lock from util recv to the owner's generic recv before acquiring ep lock.

shijin-aws: ... Or we could move the srx lock inside ep lock in provider's ep_progress.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1180 @@
...
+					 flags & (FI_CLAIM | FI_DISCARD));
+	if (!rx_entry) {
+		FI_DBG(&core_prov, FI_LOG_EP_CTRL, "Message not found\n");
+		return ofi_cq_write_error_peek(srx->cq, tag, context);
+	}
+	FI_DBG(&core_prov, FI_LOG_EP_CTRL, "Message found\n");
+
+	if (flags & (FI_CLAIM | FI_DISCARD)) {
+		if (flags & FI_DISCARD) {
+			ret = rx_entry->peer_entry.srx->peer_ops->discard_tag(
```
shijin-aws, line 617: So we do not release rx entry in the discard op, but release it outside explicitly?

shijin-aws: I see, the man page has stated that discard call should only release resources associated with the fi_peer_rx_entry.

prov/util/src/util_srx.c
aingerson: @shijin-aws I think we need to add some lock definition to the srx lock and require it be the top most lock. shm (or the owner) might need the ep lock but that's really only needed because of the resource management in start_msg. So I think we can remove the shm acquire before the generic recv call and move that into the start_msg call so we always acquire in the same order (srx, ep). Does that seem reasonable to you?
I was chatting to Sean and he also suggested to passing in a genlock into srx for it to use as the srx lock which would allow the util srx code to use a lock that the owner already has when calling peer progress. Would that be of interest for efa?

shijin-aws: >I think we need to add some lock definition to the srx lock and require it be the top most lock ... So I think we can remove the shm acquire before the generic recv call and move that into the start_msg call so we always acquire in the same order (srx, ep). Does that seem reasonable to you?

Yes, I agree. I think we should make it clear that srx->lock should be the top most lock. In that case shm or efa's generic_recv does not require more lock before acquiring the srx lock.

> I was chatting to Sean and he also suggested to passing in a genlock into srx for it to use as the srx lock which would allow the util srx code to use a lock that the owner already has when calling peer progress. Would that be of interest for efa?

That may give us more flexibility, yes. But I am ok with the current implementation as long as we agree on the sequence of locking. I am happy to delegate the locking to util srx so efa does not need to have more code.

aingerson: @shijin-aws I've updated it to switch the locking so that shm acquires the ep lock in start_msg. Let me know if this still looks problematic.

prov/sm2/src/sm2_ep.c
```diff
@@ -519,197 +384,13 @@ static int sm2_ep_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
...
-	.get_msg = sm2_get_msg,
-	.get_tag = sm2_get_tag,
-	.queue_msg = sm2_queue_msg,
-	.queue_tag = sm2_queue_tag,
-	.free_entry = sm2_free_entry,
-};
-
 static int sm2_discard(struct fi_peer_rx_entry *rx_entry)
 {
 	struct sm2_xfer_ctx *xfer_ctx = rx_entry->peer_context;
```
aingerson, line 389: @a-szegel I noticed a bug here and started to fix it but then realized I have to do protocol things to return the command to the sender. I'm going to let you fix that. Interesting that no testing caught this though. Might point to something missing in cleanup. Something to double check.

prov/util/src/util_srx.c
shijin-aws: The change looks reasonable to me. I applied the same change in efa's start and discard to make them consistent. I ran the multi thread tests (osu_latency_mt) on both single node and multi node. Both look good to me.

aingerson: Excellent, thanks! Do you see any performance difference with these changes?

shijin-aws:
>@shijin-aws Oh wow yeah please let me know how this new version looks and I can take a look if it's problematic. Thank you!

An update for this: the application performance has large variance between runs. And using micro-benchmarks I was not able to observe visible difference brought by the util srx change. I am still gathering more data but I don't think it's a blocker for this PR to be merged. I had a brief review for the implementation of the src_unexp_queue and I didn't see how it can cause worse performance.

prov/util/src/util_srx.c
shijin-aws: osu_latency_mt does not show much difference.

Most of the application I ran do not request FI_THREAD_SAFE, so they shouldn't be impacted either.

shijin-aws:
AWS CI failure is for SM2 provider. One of them 

```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.28.49 'timeout 360 /bin/bash --login -c '"'"'/home/centos/PortaFiducia/build/libraries/libfabric/pr8907/install/fabtests/bin/fi_bw -e rdm -v -T 1 -p sm2 -s 172.31.28.49'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.28.49 'timeout 360 /bin/bash --login -c '"'"'/home/centos/PortaFiducia/build/libraries/libfabric/pr8907/install/fabtests/bin/fi_bw -e rdm -v -T 1 -p sm2 -s 172.31.28.49 172.31.28.49'"'"''
client_stdout:

client returncode: 255
server_stdout:

server returncode: 124
```

prov/shm/src/smr_av.c
```diff
@@ -123,6 +133,8 @@ static int smr_av_insert(struct fid_av *av_fid, const void *addr, size_t count,
 			smr_map_to_endpoint(smr_ep->region, shm_id);
 			smr_ep->region->max_sar_buf_per_peer =
 				SMR_MAX_PEERS / smr_av->smr_map->num_peers;
+			srx = smr_get_peer_srx(smr_ep);
+			srx->owner_ops->foreach_unspec_addr(srx, &smr_get_addr);
```
shijin-aws, line 137: This change will cause a segfault in efa, because efa hasn't implement such op

shijin-aws: It's a little tricky here. We would have to implement such op in efa, efa in main branch hasn't onboarded util srx. We may have to make it as a no-op for this PR. The peer API ops in efa provider are defined in https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_srx.c. I can open a PR to add this as a no-op if you prefer that.

shijin-aws: I opened a PR https://github.com/ofiwg/libfabric/pull/8998 to add this op as no-op. You need to rebase on this PR after its merged.

aingerson:
@shijin-aws Any other changes needed while I rebase? I've fixed a bug in sm2 which might take care of the hang

shijin-aws:
Is Intel CI failure a real issue? I have seen it failed for many recent PRs.

aingerson:
@shijin-aws Not related to this change. It's a tcp error that I've seen happen more recently.
@shefty Do you have any idea?
```
08:50:38  - name:   fi_rdm_tagged_bw -v -U -p "tcp"
08:50:38    timestamp: 20230608-155037+0000
08:50:38    result: Fail
08:50:38    time:   1
08:50:38    server_cmd:  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8907/17/dbg/bin/fi_rdm_tagged_bw -v -U -p "tcp"   -s ci7-eth2
08:50:38    server_stdout: |
08:50:38      [error] fabtests:common/shared.c:2687: cq_readerr 265 (Truncation error), provider errno: 11 (Resource temporarily unavailable)
08:50:38    client_cmd:  /home/cstbuild/ofi-Install/libfabric/ofi_libfabric/PR-8907/17/dbg/bin/fi_rdm_tagged_bw -v -U -p "tcp"   -s ci8-eth2 ci7-eth2
08:50:38    client_stdout: |
08:50:38      [error] fabtests:common/shared.c:2687: cq_readerr 125 (Operation canceled), provider errno: 115 (Operation now in progress)

**Comment by shefty** [comment]
I'll see if I can reproduce that error.  There were changes related to IMPI support around this.  It could be that updates to the test got dropped.

**Comment by aingerson** [comment]
@shijin-aws I think so from my end. @shefty Do you want to give this another review?

**Comment by aingerson** [comment]
@shijin-aws @shefty Updated again to address Sean's comments. I ended up adding the lock assignment that Sean recommended before because it simplified and reduced the lock calls but also because it was necessary in order to add the is lock held asserts he asked for (in the shm owner case we were relying on the ep lock to control the locking but we couldn't assert it was held from the util srx code).
This addition required an update to the util_ep code to move it from the mutex specific lock (with function pointers to the acquire/release functions) to the genlock (as a mutex/no op). So now this change touches all the providers that were using the ep lock/unlock function pointers. We should make sure to double all of them.
You'll have to update the efa+srx usage to pass in the lock to use.

**Comment by shijin-aws** [comment]
>This addition required an update to the util_ep code to move it from the mutex specific lock (with function pointers to the acquire/release functions) to the genlock (as a mutex/no op)

Is it still ok for efa to use `ofi_ep_acquire_lock`?

File: include/ofi_util.h [hunk 31]
```diff
@@ -327,22 +326,6 @@ ofi_ep_fid_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
 			   bfid, flags);
 }
 
-static inline void ofi_ep_lock_acquire(struct util_ep *ep)
-{
-	ep->lock_acquire(&ep->lock);
```
shijin-aws, line 332: can't we call ofi_genlock here?

aingerson: We can. I just thought it was silly for an inline function to just be one line that calls another inline function.

aingerson:
> > This addition required an update to the util_ep code to move it from the mutex specific lock (with function pointers to the acquire/release functions) to the genlock (as a mutex/no op)
> 
> Is it still ok for efa to use `ofi_ep_acquire_lock`? I have made efa to use that in all places. See #8987

@shijin-aws I replaced all calls of ofi_ep_acquire/release_lock to the genlock calls which do the exact same thing. So it can still call the ep lock, just using a different function

shijin-aws:
>@shijin-aws I replaced all calls of ofi_ep_acquire/release_lock to the genlock calls which do the exact same thing. So it can still call the ep lock, just using a different function

If so, will you include the function name change within your PR? I couldn't open a separate PR inside efa to do that without basing on your change, that won't pass the CI.

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1187 @@
...
+	if (ret) {
+		free(srx);
+		return ret;
+	}
+
+	srx->min_multi_recv_size = default_min_multi_recv;
+	srx->iov_limit = iov_limit;
+	srx->dir_recv = domain->info_domain_caps & FI_DIRECTED_RECV;
+	srx->update_func = update_func;
+	srx->lock = lock;
```
shijin-aws, line 1171: Can we have a default lock here? So provider doesn't have to create itself?

prov/shm/src/smr_ep.c
```diff
@@ -1621,7 +1239,9 @@ int smr_srx_context(struct fid_domain *domain, struct fi_rx_attr *attr,
 		smr_domain->srx = ((struct fi_peer_srx_context *) (context))->srx;
 		return FI_SUCCESS;
 	}
-	return smr_ep_srx_context(smr_domain, attr->size, rx_ep);
+	return util_ep_srx_context(&smr_domain->util_domain, attr->size,
+			SMR_IOV_LIMIT, SMR_INJECT_SIZE, &smr_update,
+			&smr_domain->util_domain.lock, rx_ep);
```
shijin-aws, line 1244: Currently efa provider could not use util domain lock as the srx lock. Efa uses util domain lock in mr operations, which are invoked inside the progress engine.

shijin-aws: I may need to audit whether efa really needs that domain lock when the functions are called inside the progress engine.

shijin-aws: Also, it seems shm call ofi_domain_init with lock type as mutex lock... Do you need to add a check for domain_attr->threading there?

shijin-aws:
I have rebased efa srx work on this. Everything looks good.

shefty (approved):
Need to wait on Intel CI to re-run.  The CI controller is being updated early this week.

prov/util/src/util_srx.c
aingerson: I originally had that but it made cleanup messy (since we can only destroy the lock if we created it). Would you like me to add that in for this PR?

prov/util/src/util_srx.c
```diff
@@ -0,0 +1,1187 @@
...
+
+static int util_get_msg(struct fid_peer_srx *srx, fi_addr_t addr,
+		        size_t size, struct fi_peer_rx_entry **rx_entry)
+{
+	struct util_srx_ctx *srx_ctx;
+	struct util_rx_entry *util_entry, *any_entry;
+	struct slist *queue;
+
+	srx_ctx = srx->ep_fid.fid.context;
+	assert(ofi_genlock_held(srx_ctx->lock));
```
shijin-aws, line 202: Would you mind pointing to me where shm grab such lock in ep progress?

aingerson: Well I think I need to disable the path you are talking about. That is the path for when the application calls fi_srx_context without FI_PEER (so sharing an srx between shm EPs). I implemented that for completion but I don't think that is supported in shm without some extra domain locking. The "normal" path for shm is where each EP owns the srx so we use the EP lock [here](https://github.com/aingerson/libfabric/blob/peer/prov/shm/src/smr_ep.c#L1338)

shijin-aws: I see, so it's actually a ep lock. Could be a workaround for efa as well (I mean make efa and shm share the same ep lock). I know it's not recommended though.

aingerson: When shm is working as a peer, it uses the ep lock to manage its internal ep resources but the srx lock (owned by the peer) is in charge of the srx resource locking (which is the one efa needs to grab before calling progress for shm).

prov/shm/src/smr_ep.c
aingerson: I don't think we need to check the threading level on the domain lock. The FI_THREAD_DOMAIN specification that we optimize for in the EP specifies that:
`A domain serialization model requires applications to serialize access to all objects belonging to a domain.`
So that means that anything within the domain is serialized, but not necessarily the domain itself.

shijin-aws:
AWS CI failed for shm provder in the uber test

```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.27.231 'timeout 1000 /bin/bash --login -c '"'"'/home/ubuntu/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/bin/fi_ubertest -x'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.27.231 'timeout 1000 /bin/bash --login -c '"'"'/home/ubuntu/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/bin/fi_ubertest -u /home/ubuntu/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/share/fabtests/test_configs/shm/quick.test 172.31.27.231'"'"''
client_stdout:

client returncode: None
Client error: Client timed out
server_stdout:
fi_ubertest: prov/util/src/util_srx.c:375: util_free_entry: Assertion `ofi_genlock_held(srx->lock)' failed.
timeout: the monitored command dumped core

server returncode: 255
```

Note this case explicitly turned cma off in the OS.

prov/shm/src/smr_ep.c
```diff
@@ -954,14 +809,14 @@ static int smr_ep_close(struct fid *fid)
 		free(ep->sock_info);
 	}
 
+	if (ep->util_ep.ep_fid.msg != &smr_no_recv_msg_ops)
```
shijin-aws, line 812: Consider shm is used as both owner and peer. It seems srx_ctx is only initiated when ep is enabled. But here `ep->util_ep.ep_fid.msg` is set in fi_endpoint. Will it cause a segfault if application do `fi_endpoint` and then `fi_close(ep_fid)`? I catch a similar issue in one efa unit test that has such corner case.

wenduwan:
@aingerson AWS CI failure looks real
```
:33:52 
14:33:52 server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.27.231 'timeout 1000 /bin/bash --login -c '"'"'/home/ubuntu/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/bin/fi_ubertest -x'"'"''
14:33:52 
14:33:52 client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.27.231 'timeout 1000 /bin/bash --login -c '"'"'/home/ubuntu/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/bin/fi_ubertest -u /home/ubuntu/PortaFiducia/build/libraries/libfabric/pr8907-debug/install/fabtests/share/fabtests/test_configs/shm/quick.test 172.31.27.231'"'"''
14:33:52 client_stdout:
14:33:52 
14:33:52 client returncode: None
14:33:52 Client error: Client timed out
14:33:52 server_stdout:
14:33:52 fi_ubertest: prov/util/src/util_srx.c:375: util_free_entry: Assertion `ofi_genlock_held(srx->lock)' failed.
14:33:52 timeout: the monitored command dumped core
14:33:52 
```

aingerson:
@shijin-aws Thanks! I will take a look

shijin-aws:
Bump this comment

> Consider when shm is used as both owner and peer, it seems srx_ctx is only initiated when ep is enabled. But here ep->util_ep.ep_fid.msg is set in fi_endpoint. Will it cause a segfault if application do fi_endpoint and then fi_close(ep_fid)? I catch a similar issue in one efa unit test that has such corner case.

aingerson:
@shijin-aws Sorry I saw that and then thought to myself, "That is a good point. I should fix that!" and then got distracted by the actual failure. I will fix it! Thanks!

shijin-aws (approved):
Thank you!

shijin-aws:
@aingerson shall we merge ?

aingerson:
@shijin-aws I think so. @shefty Ok to merge?

shefty:
If everyone is happy, sure.

shijin-aws:
OK, @a-szegel @wenduwan  thumb up?

wenduwan (approved):

a-szegel (approved):

PR #8908, opened 2023-05-09
fabtests: Add rdm_unknown_peer test

shijin-aws:
This test is used to check whether libfabric
provider can handle unknown peers in the following scenario:

Client:
- send message to server x2

Server:
- post with rx any source (during ft_enable_ep_recv)
- remove the client from av to make it unknown
- get the rx comp for the first rx, which should succeed anyway because it was posted for any source
- insert client back into AV
- post rx with fi addr 0 (client fiaddr)
- get the rx comp for the second rx.

### DIFF
--- a/fabtests/Makefile.am
+++ b/fabtests/Makefile.am
@@ -46,6 +46,7 @@ bin_PROGRAMS = \
 	functional/fi_bw \
 	functional/fi_rdm_multi_client \
 	functional/fi_loopback \
+	functional/fi_rdm_unknown_peer \
 	benchmarks/fi_msg_pingpong \
 	benchmarks/fi_msg_bw \
 	benchmarks/fi_rma_bw \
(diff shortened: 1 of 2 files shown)


aingerson:
@shijin-aws I think a better stand alone test would be to validate the directed recv path (like I mentioned over email). During my testing, I extended the unexpected message test to do this which could be a better place for it. See [here](https://github.com/ofiwg/libfabric/commit/dd94d1508aeb5805ee484f1c9c1b1d43770d0b6f)
@shefty What are your thoughts on adding a test vs extending the test? I know tcp currently does not support the unexpected test but we could reduce the concurrent messages to run over tcp?

aingerson:
@shefty I'm seeing tcp hang on the unexpected test as is but it passes with -M 1 (only one unexpected message at a time, default is 5). Is this what you're saying should pass or do you expect it to pass for more?

shefty:
By default, tcp only supports 4 unexpected messages, each of a max size of 128 bytes.  Both those values are configurable.

shijin-aws:
@aingerson I agree that extending the unexpected message test could be better. Since you already have that done in your branch. I will close this PR and use your extended test. Currently we run this test for efa with `fi_unexpected_msg -e rdm -I 10`

aingerson:
Oh, I knew about the limit of 4 but not of the size limit. -M 4 -S 128 passes with tcp, including with the unexpected addr extension I added. Would you like to have that addition to the test and enable tcp testing?

shefty:
@aingerson - sure - that sounds good

PR #8909, opened 2023-05-09
prov/verbs: Remove saved_wc_list and use CQ directly

shefty:
The provider reads completions from the HW CQ, processes them, and if needed reports them up to the user in the appropriate format.  Through these steps, it doesn't actually write entries into the libfabric CQ.  However, there are times when it reads from the HW CQ as part of internal progress, but can't report the completions to the user yet.  Rather than writing an entry into the libfabric CQ, it insteads stores the HW completion on a separate list.

The entire process is confusing, results in extra memory, leaves the CQ allocated memory unused, and increases code by not properly leveraging the common CQ functions.

Remove the saved_wc_list and related memory pools.  Instead, after reading from the HW CQ, simply write a completion into the libfabric CQ if one is needed.

### DIFF
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -53,112 +53,85 @@ enum ibv_wc_opcode vrb_wr2wc_opcode(enum ibv_wr_opcode wr)
 	return (wr < ARRAY_SIZE(wc)) ? wc[wr] : IBV_WC_SEND;
 }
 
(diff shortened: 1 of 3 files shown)


shefty:
Failure is in the tcp provider (fi_rdm_tagged_bw -v -U -p "tcp").  This is a recent issue that's frequently popping up, but unrelated to this PR.  We're looking into the tcp failure separately to see if it's a problem in libfabric or the test.

PR #8913, opened 2023-05-09
prov/sm2: sm2 breakout inject header, and align freestack to 4k boundary

a-szegel:
This change breaks out the SM2 header into multiple structures, and makes sure the sm2 freestack entries stay aligned.

I will rebase out the ofi_mem change when that PR gets merged.


lrbison (commented):
As far as the implementation goes, this looks good.

Does it actually improve performance?

a-szegel:
https://github.com/ofiwg/libfabric/pull/8895
```
        "program": "/home/ec2-user/libfabric/fabtests/install/bin/fi_rdm_tagged_pingpong",
        "args": ["--pin-core", "0", "-S", "0", "-s", "127.0.0.1", "-p", "sm2"],

0       10k     0           0.00s      0.00       0.20       4.99
```

This PR:
```
0       10k     0           0.00s      0.00       0.19       5.36
```

Before the deleivery_complete PR was merged, we were at 5.44 Mxfers/sec... and that slowed us down a bit... I didn't realize that b/c I have functional changes in branches, and performance changes in other branches... and I failed to cross test all of them.

lrbison:
LGTM.  Small comment, you change `blah_blah_SIZE` to `blah_blah_LENGTH` when you introduce USER_DATA_LENGTH.  Is _SIZE or _LENGTH generally preferred?

lrbison (approved):

prov/sm2/src/sm2.h
```diff
@@ -105,7 +105,7 @@ enum {
  * 	sender_gid - id of msg sender
  * 	user_data - the message
  */
-struct sm2_xfer_entry {
+struct sm2_xfer_header {
```
aingerson, line 110: I would prefer hdr to match shm/the field name and for shorter names in general but I don't care that match so if that's my only comment, it's ok
also the commit message is a bit off - it says the new struct is called "sm2_inject_header"

prov/sm2/src/sm2_progress.c
```diff
@@ -91,18 +92,18 @@ static int sm2_start_common(struct sm2_ep *ep,
...
 		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "Error processing op\n");
 		ret = sm2_write_err_comp(ep->util_ep.rx_cq, rx_entry->context,
 					 comp_flags, rx_entry->tag, err);
 	} else {
-		ret = sm2_complete_rx(ep, rx_entry->context, xfer_entry->op,
-				      comp_flags, total_len, comp_buf,
-				      xfer_entry->sender_gid, xfer_entry->tag,
-				      xfer_entry->cq_data);
+		ret = sm2_complete_rx(
+			ep, rx_entry->context, xfer_entry->hdr.op, comp_flags,
```
aingerson, line 104: ugh why is the formatting like this?

prov/sm2/src/sm2.h
```diff
@@ -76,6 +76,8 @@
 #define SM2_PREFIX_NS		"fi_ns://"
 #define SM2_VERSION		1
 #define SM2_IOV_LIMIT		4
+#define SM2_INJECT_USER_DATA_SIZE \
+	(SM2_INJECT_SIZE - sizeof(struct sm2_xfer_header))
```
aingerson, line 80: I think this makes the difference between SM2_INJECT_SIZE and SM2_INJECT_USER_DATA_SIZE confusing. I would leave this one as SM2_INJECT_SIZE (which I read as the size of message which you can inject) and the other one to something like SM2_XFER_ENTRY_SIZE

prov/sm2/src/sm2_progress.c
a-szegel: We can play with clang format more in the future... nothing we can do about this on this commit.

prov/sm2/src/sm2_coordination.h
```diff
@@ -48,7 +48,7 @@
 
 #include <rdma/providers/fi_prov.h>
 
-#define SM2_INJECT_SIZE	      4096
+#define SM2_INJECT_XFER_SIZE  4096
```
aingerson, line 51: Again- still confusing. I would rename to SM2_XFER_ENTRY_SIZE. Otherwise that looks like it's the actual transfer size

aingerson (commented):
Just one last thing. otherwise looks good

lrbison:
Could you update the wording in you last commit message? I don't think it's accurate any more.

lrbison:
Thank you!

lrbison (approved):

aingerson (approved):

PR #8915, opened 2023-05-10
prov/rxm: revert util peer CQ support

aingerson:
rxm_cq_owner_writeerr() expects the err_entry has a valid op_context pointing to ‘struct rxm_coll_buf’

Updated to revert peer CQ support

### DIFF
--- a/prov/rxm/src/rxm_cq.c
+++ b/prov/rxm/src/rxm_cq.c
@@ -141,14 +141,14 @@ static void rxm_cq_write_error_trunc(struct rxm_rx_buf *rx_buf, size_t done_len)
 	FI_WARN(&rxm_prov, FI_LOG_CQ, "Message truncated: "
 		"recv buf length: %zu message length: %" PRIu64 "\n",
 		done_len, rx_buf->pkt.hdr.size);
-	ret = ofi_peer_cq_write_error_trunc(rx_buf->ep->util_ep.rx_cq,
-				rx_buf->recv_entry->context,
-				rx_buf->recv_entry->comp_flags |
-				rx_buf->pkt.hdr.flags,
-				rx_buf->pkt.hdr.size,
-				rx_buf->recv_entry->rxm_iov.iov[0].iov_base,
-				rx_buf->pkt.hdr.data, rx_buf->pkt.hdr.tag,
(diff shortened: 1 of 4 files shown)


aingerson:
@j-xiong Went with a slightly different fix to initialize all the fields to NULL. The ep is only accessed if the flag is set so it shouldn't matter.

prov/rxm/src/rxm_cq.c
```diff
@@ -1680,8 +1680,10 @@ void rxm_cq_write_error(struct util_cq *cq, struct util_cntr *cntr,
 void rxm_cq_write_error_all(struct rxm_ep *rxm_ep, int err)
 {
 	struct fi_cq_err_entry err_entry = {0};
+	struct rxm_coll_buf req = {0};
 	ssize_t ret = 0;
 
+	err_entry.op_context = &req;
```
shefty, line 1686: Something seems off in the call path if the code always expects op_context to reference some collective buffer structure.

j-xiong: The rxm peer cq implementation was initially for collective offload only. See comments at L1883~L1889. If here we expect broader usage, some refactoring will be needed.

aingerson: Yeah I though so too. If rxm needs to write an error for peer then it feels like that should be done separately on that path and not in the peer cq path? I'm not really sure what the target for this write_error_all call is. Would it ever have a collective buffer?

j-xiong: The changes were introduced by commit  93c1a52d2d6b11c82a78035265c00ed3366cbd07. Did we miss something there or do we forget some tricks that made it work?

aingerson: @shefty @j-xiong Updated to just revert the commit since it's not necessary. The revert should fix it since the error path since it goes through the normal util cq insert path. We can revisit the peer CQ later when/if we add peer CQ support outside of peer collectives.

shefty: Thanks - can you also pick for v1.18.x?

shefty (approved):

PR #8918, opened 2023-05-10
prov/sm2: Fix SM2 fabtest issues

lrbison:
This PR fixes two bugs which prevented fabtests for SM2 from being run in CI.

The primary issue is the return value of verify peer.

The failed startup is a symptom of that, but causes a crash rather than a clean exit.

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -253,7 +253,7 @@ int sm2_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 int sm2_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
 		  struct fid_cntr **cntr_fid, void *context);
 
-sm2_gid_t sm2_verify_peer(struct sm2_ep *ep, fi_addr_t fi_addr, sm2_gid_t *gid);
+int sm2_verify_peer(struct sm2_ep *ep, fi_addr_t fi_addr, sm2_gid_t *gid);
 
 typedef ssize_t (*sm2_proto_func)(struct sm2_ep *ep,
 				  struct sm2_region *peer_smr,
--- a/prov/sm2/src/sm2_av.c
+++ b/prov/sm2/src/sm2_av.c
@@ -249,13 +249,13 @@ int sm2_av_open(struct fid_domain *domain, struct fi_av_attr *attr,
 	if (ret)
 		goto out;
 
-	*av = &sm2_av->util_av.av_fid;
-	(*av)->fid.ops = &sm2_av_fi_ops;
-	(*av)->ops = &sm2_av_ops;
-
 	ret = sm2_file_open_or_create(&sm2_av->mmap);
 	if (ret)
 		goto out;
+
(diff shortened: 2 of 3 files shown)


wzamazon (approved):

aingerson (approved):
Nice catches!

a-szegel (approved):
Thanks Luke, great catches.

prov/sm2/src/sm2.h
```diff
@@ -253,7 +253,7 @@ int sm2_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 int sm2_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
 		  struct fid_cntr **cntr_fid, void *context);
 
-sm2_gid_t sm2_verify_peer(struct sm2_ep *ep, fi_addr_t fi_addr, sm2_gid_t *gid);
+int sm2_verify_peer(struct sm2_ep *ep, fi_addr_t fi_addr, sm2_gid_t *gid);
```
a-szegel, line 256: We should make this a `ssize_t` b/c it gets everywhere it gets called puts it into a ssize_t, and the implicit case from a unsigned int to signed long got us:

```
ssize_t ret = 0;
ret = sm2_verify_peer(ep, addr, &peer_gid);
```

The libfabric API decided to make every return value be a ssize_t, so we should use that convention in SM2 everywhere.

lrbison: It's hard to know what an error should be because errors are just `#defined` without specific types, but [fi_strerror](https://github.com/ofiwg/libfabric/blob/main/include/rdma/fi_errno.h#L200) takes an int, so I think if there is a standard type, it would be `int`

a-szegel: It is not an int.
```
struct fi_ops_msg {
	size_t	size;
	ssize_t (*recv)(struct fid_ep *ep, void *buf, size_t len, void *desc,
			fi_addr_t src_addr, void *context);
	ssize_t (*recvv)(struct fid_ep *ep, const struct iovec *iov, void **desc,
			size_t count, fi_addr_t src_addr, void *context);
	ssize_t (*recvmsg)(struct fid_ep *ep, const struct fi_msg *msg,
			uint64_t flags);
	ssize_t (*send)(struct fid_ep *ep, const void *buf, size_t len, void *desc,
			fi_addr_t dest_addr, void *context);
	ssize_t (*sendv)(struct fid_ep *ep, const struct iovec *iov, void **desc,
			size_t count, fi_addr_t dest_addr, void *context);
	ssize_t (*sendmsg)(struct fid_ep *ep, const struct fi_msg *msg,
			uint64_t flags);
	ssize_t	(*inject)(struct fid_ep *ep, const void *buf, size_t len,
			fi_addr_t dest_addr);
	ssize_t (*senddata)(struct fid_ep *ep, const void *buf, size_t len, void *desc,
			uint64_t data, fi_addr_t dest_addr, void *context);
	ssize_t	(*injectdata)(struct fid_ep *ep, const void *buf, size_t len,
			uint64_t data, fi_addr_t dest_addr);
};
```

PR #8624, opened 2023-03-08
core: introduce FI_OPT_CUDA_API_PERMITTED option

wzamazon:
This option allow user to control an endpoint's behavior in making calls to CUDA API.

Issue #8610, opened 2023-03-03
Definition peer to peer support


wzamazon:
This came from a discussion https://github.com/ofiwg/libfabric/pull/8529

Background is that application like NCCL need a way to specify libfabric endpoint cannot make calls to CUDA API to support CUDA memory. 

@shefty suggested to use the FI_OPT_HMEM_P2P_REQUIRED, which currently states as the following:

```
FI_HMEM_P2P_REQUIRED: Peer to peer support must be used for transfers, transfers that cannot be performed using p2p will be reported as failing.
```
From https://ofiwg.github.io/libfabric/main/man/fi_endpoint.3.html

However, to use this option for the purpose I described, we need a definition of "peer to peer support", which is lacking in the `fi_endpoint` document. So I opened this issue to ask whether libfabric community to agree on a definition for "peer to peer" support.

One thing I want to mention is that NCCL does allow libfabric to use GDRcopy, see this [comment](https://github.com/ofiwg/libfabric/issues/6124#issuecomment-665113829) from @jdinan. EFA provider does use GDRcopy when used by NCCL, and found it to be efficient for small messages.

I understand that other providers, like RxM, also want to use GDRcopy to support NCCL.

Therefore, I think it would be ideal if we can define "peer to peer support" in a way that mechanisms like GDRcopy is counted as "peer to peer" support.

shefty:
Peer to peer is meant to describe PCI peer to peer transfers, or device to device transfers that do not require bouncing data through host buffers.  This could also apply to other device buses, not just PCI.

wzamazon:
I see.

I think for the case of NCCL, `HMEM_P2P_REQUIRED` is too strong. Basically, it need a way to know whether the provider is capable of P2P, not necessarily that all transfer must be through peer 2 peer.

I am reading  the man page for `FI_HMEM_P2P_ENABLED`. It did not specify what provider should do if it does not support Peer 2 peer.

Would it be reasonable for a provider to return -FI_EOPNOSUPP, if user set FI_HMEM_P2P_ENABLED and the provider is incapable of peer 2 peer support?

shefty:
Maybe the question is whether HMEM_P2P_REQUIRED is useful?  Or is it only useful if it also allows gdrcopy?

Does gdrcopy behave the same as if p2p were used?

wzamazon:
> Maybe the question is whether HMEM_P2P_REQUIRED is useful? Or is it only useful if it also allows gdrcopy?

[...]

PR #8919, opened 2023-05-11
prov/sm2: Add CMA and CUDA IPC capabilities to sm2

sunkuamzn:
AWS needs SM2 to have the capability to send any message size in order to perform benchmarking on real world applications.

The CUDA IPC commit only enables device-device and device-host transfers using CUDA IPC.

~~Host-device transfers need a new protocol which will come later.~~ Added host to CUDA device commit

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -92,6 +92,7 @@ extern pthread_mutex_t sm2_ep_list_lock;
 enum {
 	sm2_proto_inject,
 	sm2_proto_return,
+	sm2_proto_cma,
 	sm2_proto_max,
 };
 
@@ -108,7 +109,7 @@ enum {
  * 		   NOTE: Only grabbing the bottom 32 bits
  * 	proto - sm2 operation
  * 	sender_gid - id of msg sender
- * 	user_data - the message
+ * 	user_data - the message, for sm2_proto_inject
  */
 struct sm2_xfer_hdr {
 	volatile long int next;
@@ -149,6 +150,11 @@ struct sm2_atomic_data {
 struct sm2_atomic_entry {
 	struct sm2_atomic_hdr atomic_hdr;
 	struct sm2_atomic_data atomic_data;
+};
+
+struct sm2_cma_data {
+	size_t iov_count;
+	struct iovec iov[SM2_IOV_LIMIT];
 };
 
 struct sm2_ep_name {
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -109,6 +109,7 @@ enum {
  * 	op_flags - flags associated with op,
  * 		   NOTE: Only grabbing the bottom 32 bits
  * 	proto - sm2 operation
+ * 	proto_flags - Flags used by the sm2 operation
  * 	sender_gid - id of msg sender
  * 	user_data - the message, for sm2_proto_inject
  */
@@ -120,7 +121,8 @@ struct sm2_xfer_hdr {
 	uint64_t context;
 	uint32_t op;
 	uint32_t op_flags;
-	uint32_t proto;
+	uint16_t proto;
+	uint16_t proto_flags;
 	sm2_gid_t sender_gid;
 };
 
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -93,6 +93,7 @@ enum {
 	sm2_proto_inject,
 	sm2_proto_return,
 	sm2_proto_cma,
+	sm2_proto_ipc,
 	sm2_proto_max,
 };
 
@@ -213,6 +214,7 @@ struct sm2_xfer_ctx {
 
 struct sm2_domain {
 	struct util_domain util_domain;
+	struct ofi_mr_cache *ipc_cache;
 	struct fid_peer_srx *srx;
 };
 
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -82,6 +82,9 @@
 #define SM2_ATOMIC_INJECT_SIZE	    (SM2_INJECT_SIZE - sizeof(struct sm2_atomic_hdr))
 #define SM2_ATOMIC_COMP_INJECT_SIZE (SM2_ATOMIC_INJECT_SIZE / 2)
 
+/* Protocol flags for all protocols */
+#define FI_SM2_UNEXP	   (1 << 0)
+
 extern struct fi_provider sm2_prov;
 extern struct fi_info sm2_info;
 extern struct util_prov sm2_util_prov;
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -85,6 +85,10 @@
 /* Protocol flags for all protocols */
 #define FI_SM2_UNEXP	   (1 << 0)
 
+/* Protocol flags for SM2 CMA protocol */
+#define FI_SM2_CMA_HOST_TO_DEV	   (1 << 1)
+#define FI_SM2_CMA_HOST_TO_DEV_ACK (1 << 2)
+
 extern struct fi_provider sm2_prov;
 extern struct fi_info sm2_info;
 extern struct util_prov sm2_util_prov;
@@ -161,6 +165,10 @@ struct sm2_atomic_entry {
 struct sm2_cma_data {
 	size_t iov_count;
 	struct iovec iov[SM2_IOV_LIMIT];
+
+	/* Used for IPC host to device protocol */
+	struct ipc_info ipc_info;
+	struct fi_peer_rx_entry *rx_entry;
 };
 
 struct sm2_ep_name {
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -244,7 +244,7 @@ ssize_t sm2_verify_peer(struct sm2_ep *ep, fi_addr_t fi_addr, sm2_gid_t *gid);
 typedef ssize_t (*sm2_proto_func)(struct sm2_ep *ep,
 				  struct sm2_region *peer_smr,
 				  sm2_gid_t peer_gid, uint32_t op, uint64_t tag,
-				  uint64_t data, uint64_t op_flags,
+				  uint64_t data, uint64_t *op_flags,
 				  struct ofi_mr **mr, const struct iovec *iov,
 				  size_t iov_count, size_t total_len,
 				  void *context);
--- a/prov/sm2/src/sm2_attr.c
+++ b/prov/sm2/src/sm2_attr.c
@@ -90,6 +90,19 @@ struct fi_ep_attr sm2_ep_attr = {
 	.type = FI_EP_RDM,
 	.protocol = FI_PROTO_SHM,
 	.protocol_version = 1,
+	.max_msg_size = SIZE_MAX,
+	.max_order_raw_size = SIZE_MAX,
+	.max_order_waw_size = SIZE_MAX,
+	.max_order_war_size = SIZE_MAX,
+	.mem_tag_format = FI_TAG_GENERIC,
+	.tx_ctx_cnt = 1,
+	.rx_ctx_cnt = 1,
+};
+
+struct fi_ep_attr sm2_hmem_ep_attr = {
+	.type = FI_EP_RDM,
+	.protocol = FI_PROTO_SHM,
+	.protocol_version = 1,
 	.max_msg_size = SM2_INJECT_SIZE,
 	.max_order_raw_size = SM2_INJECT_SIZE,
 	.max_order_waw_size = SM2_INJECT_SIZE,
@@ -150,7 +163,7 @@ struct fi_info sm2_hmem_info = {
 	.addr_format = FI_ADDR_STR,
 	.tx_attr = &sm2_hmem_tx_attr,
 	.rx_attr = &sm2_hmem_rx_attr,
-	.ep_attr = &sm2_ep_attr,
+	.ep_attr = &sm2_hmem_ep_attr,
 	.domain_attr = &sm2_hmem_domain_attr,
 	.fabric_attr = &sm2_fabric_attr,
 };
--- a/prov/sm2/src/sm2_attr.c
+++ b/prov/sm2/src/sm2_attr.c
@@ -99,19 +99,6 @@ struct fi_ep_attr sm2_ep_attr = {
 	.rx_ctx_cnt = 1,
 };
 
-struct fi_ep_attr sm2_hmem_ep_attr = {
-	.type = FI_EP_RDM,
-	.protocol = FI_PROTO_SHM,
-	.protocol_version = 1,
-	.max_msg_size = SM2_INJECT_SIZE,
-	.max_order_raw_size = SM2_INJECT_SIZE,
-	.max_order_waw_size = SM2_INJECT_SIZE,
-	.max_order_war_size = SM2_INJECT_SIZE,
-	.mem_tag_format = FI_TAG_GENERIC,
-	.tx_ctx_cnt = 1,
-	.rx_ctx_cnt = 1,
-};
-
 struct fi_domain_attr sm2_domain_attr = {
 	.name = "sm2",
 	.threading = FI_THREAD_SAFE,
@@ -163,7 +150,7 @@ struct fi_info sm2_hmem_info = {
 	.addr_format = FI_ADDR_STR,
 	.tx_attr = &sm2_hmem_tx_attr,
 	.rx_attr = &sm2_hmem_rx_attr,
-	.ep_attr = &sm2_hmem_ep_attr,
+	.ep_attr = &sm2_ep_attr,
 	.domain_attr = &sm2_hmem_domain_attr,
 	.fabric_attr = &sm2_fabric_attr,
 };
--- a/prov/sm2/src/sm2_domain.c
+++ b/prov/sm2/src/sm2_domain.c
@@ -58,6 +58,9 @@ static int sm2_domain_close(fid_t fid)
 	domain = container_of(fid, struct sm2_domain,
 			      util_domain.domain_fid.fid);
 
+	if (domain->ipc_cache)
+		ofi_ipc_cache_destroy(domain->ipc_cache);
+
 	ret = ofi_domain_close(&domain->util_domain);
 	if (ret)
 		return ret;
@@ -103,6 +106,15 @@ int sm2_domain_open(struct fid_fabric *fabric, struct fi_info *info,
 	}
 
 	sm2_domain->util_domain.threading = FI_THREAD_SAFE;
+
+	ret = ofi_ipc_cache_open(&sm2_domain->ipc_cache,
+				 &sm2_domain->util_domain);
+	if (ret) {
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+			"Unable to open IPC cache\n");
+		free(sm2_domain);
+		return ret;
+	}
 
 	*domain = &sm2_domain->util_domain.domain_fid;
 	(*domain)->fid.ops = &sm2_domain_fi_ops;
--- a/prov/sm2/src/sm2_ep.c
+++ b/prov/sm2/src/sm2_ep.c
@@ -193,7 +193,7 @@ static void sm2_format_inject(struct sm2_xfer_entry *xfer_entry,
 
 static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
 			     sm2_gid_t peer_gid, uint32_t op, uint64_t tag,
-			     uint64_t data, uint64_t op_flags,
+			     uint64_t data, uint64_t *op_flags,
 			     struct ofi_mr **mr, const struct iovec *iov,
 			     size_t iov_count, size_t total_len, void *context)
 {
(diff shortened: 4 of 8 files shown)


wzamazon:
Need rebase to fix CI failure.

sunkuamzn:
@tschuett I cannot find that flag https://github.com/search?q=repo%3Aofiwg%2Flibfabric%20FI_CUDA_API_PERMITTED&type=code

Can you elaborate?

tschuett:
See this discussion:
https://github.com/ofiwg/libfabric/issues/8610
and PR: https://github.com/ofiwg/libfabric/pull/8624

prov/sm2/src/sm2_ep.c
```diff
@@ -337,6 +337,68 @@ static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+			  size_t len, enum fi_hmem_iface iface, uint64_t device)
+{
+	void *base;
+	int ret;
+	struct ipc_info *ipc_info = (struct ipc_info *) xfer_entry->user_data;
+
+	xfer_entry->hdr.proto = sm2_proto_ipc;
+	xfer_entry->hdr.size = len;
+	ipc_info->iface = iface;
+	ipc_info->device = device;
```
a-szegel, line 350: I think you need to add delivery complete here.

prov/sm2/src/sm2_msg.c
```diff
@@ -319,7 +311,8 @@ static ssize_t sm2_generic_sendmsg(struct sm2_ep *ep, const struct iovec *iov,
 	total_len = ofi_total_iov_len(iov, iov_count);
 	assert(!(op_flags & FI_INJECT) || total_len <= SM2_INJECT_SIZE);
 
-	proto = sm2_select_proto(total_len);
+	proto = sm2_select_proto(desc, iov_count, op_flags, total_len);
+
 	ret = sm2_proto_ops[proto](ep, peer_smr, peer_gid, op, tag, data,
 				   op_flags, mr, iov, iov_count, total_len,
 				   context);
```
a-szegel, line 318: Line 322... this is as close as it will let me make a comment on...  We need to add sm2_proto_ipc here... Another better alternative solution is to pass op_flags as a pointer, so we don't need to keep extending it... if you choose to do it this way, please make the initial change for the other protocols in a separate commit.
```
if (!(op_flags & FI_DELIVERY_COMPLETE) && proto != sm2_proto_cma) {
```

a-szegel (changes requested):

prov/sm2/src/sm2_ep.c
```diff
@@ -722,6 +825,39 @@ static void sm2_free_entry(struct fi_peer_rx_entry *entry)
...
+int sm2_select_proto(void **desc, size_t iov_count, uint64_t op_flags,
+		     uint64_t total_len)
+{
+	struct ofi_mr *sm2_desc;
+	enum fi_hmem_iface iface;
+
+	sm2_desc = (struct ofi_mr *) *desc;
+	iface = sm2_desc->iface;
+
+	/* TODO: Move gdrcopy check out of non-cuda fast paths */
```
wenduwan, line 837: Reminder, we fixed a bug on read path. We cannot use inject in that case.

https://github.com/ofiwg/libfabric/pull/8925

prov/sm2/src/sm2_attr.c
```diff
@@ -85,6 +85,19 @@ struct fi_rx_attr sm2_hmem_rx_attr = {
...
+	.max_msg_size = SIZE_MAX,
+	.max_order_raw_size = SIZE_MAX,
+	.max_order_waw_size = SIZE_MAX,
+	.max_order_war_size = SIZE_MAX,
+	.mem_tag_format = FI_TAG_GENERIC,
+	.tx_ctx_cnt = 1,
+	.rx_ctx_cnt = 1,
+};
+
+struct fi_ep_attr sm2_hmem_ep_attr = {
```
a-szegel, line 100: You will need to undo this change. in your PR... how did you test this?

vidsouza:
bot:aws:retest

prov/sm2/src/sm2_domain.c
```diff
@@ -98,12 +101,21 @@ int sm2_domain_open(struct fid_fabric *fabric, struct fi_info *info,
 	ret = ofi_domain_init(fabric, info, &sm2_domain->util_domain, context,
 			      OFI_LOCK_SPINLOCK);
 	if (ret) {
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+			"Unable to open IPC cache\n");
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/sm2/src/sm2_ep.c
```diff
@@ -763,6 +825,39 @@ static void sm2_free_entry(struct fi_peer_rx_entry *entry)
...
+	if (iface == FI_HMEM_CUDA &&
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/sm2/src/sm2_ep.c
```diff
@@ -763,6 +825,39 @@ static void sm2_free_entry(struct fi_peer_rx_entry *entry)
 	ofi_spin_unlock(&srx->lock);
 }
 
+int sm2_select_proto(void **desc, size_t iov_count, uint64_t op_flags,
```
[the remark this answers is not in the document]
sunkuamzn: I added this according to SHM before your CMA change. I fixed it now.

prov/sm2/src/sm2_ep.c
```diff
@@ -337,6 +337,68 @@ static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+			  uint64_t data, uint64_t op_flags, struct ofi_mr **mr,
+			  const struct iovec *iov, size_t iov_count,
+			  size_t total_len, void *context)
+{
+	struct sm2_xfer_entry *xfer_entry;
+	struct sm2_region *self_region;
+	struct sm2_av *av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_mmap *map = &av->mmap;
+	int ret;
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/sm2/src/sm2_ep.c
```diff
@@ -337,6 +337,68 @@ static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
 	return FI_SUCCESS;
 }
 
+static int sm2_format_ipc(struct sm2_xfer_entry *xfer_entry, void *ptr,
+			  size_t len, enum fi_hmem_iface iface, uint64_t device)
+{
+	void *base;
+	int ret;
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/sm2/src/sm2_ep.c
```diff
@@ -337,6 +337,68 @@ static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
 	return FI_SUCCESS;
 }
 
+static int sm2_format_ipc(struct sm2_xfer_entry *xfer_entry, void *ptr,
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/sm2/src/sm2_ep.c
```diff
@@ -337,6 +337,68 @@ static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+	}
+
+	xfer_entry = smr_freestack_pop(sm2_freestack(self_region));
+
+	sm2_generic_format(xfer_entry, ep->gid, op, tag, data, op_flags,
+			   context);
+	ret = sm2_format_ipc(xfer_entry, iov[0].iov_base, total_len,
+			     mr[0]->iface, mr[0]->device);
+	if (ret)
+		return ret;
```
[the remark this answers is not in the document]
sunkuamzn: fixed

prov/sm2/src/sm2_progress.c
```diff
@@ -97,7 +97,7 @@ static int sm2_progress_cma(struct sm2_xfer_entry *xfer_entry,
 static int sm2_progress_inject(struct sm2_xfer_entry *xfer_entry,
 			       struct ofi_mr **mr, struct iovec *iov,
 			       size_t iov_count, size_t *total_len,
-			       struct sm2_ep *ep, int err)
+			       struct sm2_ep *ep, int *err)
```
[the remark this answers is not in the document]
sunkuamzn: i removed it

prov/sm2/src/sm2_ep.c
sunkuamzn: fixed

prov/sm2/src/sm2_msg.c
sunkuamzn: went with the pointer approach and put that in a separate commit

prov/sm2/src/sm2_ep.c
sunkuamzn: we don't have RMA yet

prov/sm2/src/sm2_attr.c
sunkuamzn: keeping it because we don't have large message device-host

prov/sm2/src/sm2_ep.c
```diff
@@ -313,7 +312,7 @@ static void sm2_format_cma(struct sm2_xfer_entry *xfer_entry,
 
 static ssize_t sm2_do_cma(struct sm2_ep *ep, struct sm2_region *peer_smr,
 			  sm2_gid_t peer_gid, uint32_t op, uint64_t tag,
-			  uint64_t data, uint64_t op_flags, struct ofi_mr **mr,
```
a-szegel: So I know I asked you to do this... but can you actually re-base this change in front of the cma change, and then squash the cma specific part into the cma commit?  This will provide a cleaner development history for a reader.

sunkuamzn: done

a-szegel (approved):
LGTM... tyvm :)

prov/sm2/src/sm2_ep.c
```diff
@@ -296,6 +309,115 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+	struct sm2_xfer_entry *xfer_entry;
+	struct sm2_av *av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_region *self_region = sm2_mmap_ep_region(&av->mmap, ep->gid);
+
+	if (smr_freestack_isempty(sm2_freestack(self_region))) {
+		sm2_progress_recv(ep);
+		if (smr_freestack_isempty(sm2_freestack(self_region)))
+			return -FI_EAGAIN;
+	}
```
lrbison, line 341: Don't forget to use the new [sm2_pop_xfer_entry](https://github.com/ofiwg/libfabric/pull/8941) function

sunkuamzn: done

sunkuamzn:
Some comments on the device to host commit are on https://github.com/ofiwg/libfabric/pull/8956

a-szegel:
Looks like this change causes a 1-2us second regression in osu_alltoallw (minor), but osu_latency is fine.

prov/sm2/src/sm2.h
```diff
@@ -176,11 +176,11 @@ static inline int sm2_match_tag(uint64_t tag, uint64_t ignore,
 static inline void sm2_generic_format(struct sm2_xfer_entry *xfer_entry,
```
a-szegel, line 195: this is wild but the osu_alltoallw degredation seems to come from this patch.

a-szegel (changes requested):
This feels very "hacked in".  We add ~4 new if statements to the critical path of the host memory case.  This is done b/c we are trying to re-use the `sm2_progress_recv` -> `sm2_progress_recv_msg` -> `sm2_start_common` abstraction that we inherited from SHM.  I don't really think this holds for us anymore, and I would like to start thinking about a new abstraction.

My major concerns are:
1. I don't want to impact the performance of the non-cuda path (which honestly doesn't look that bad, but it isn't non-zero)
2. I want the HMEM stuff to be better isolated behind protocols, then big if statements in common receive path utilities
3. I want the code to look a little cleaner.

a-szegel:
Can you run this with validation for a lot of iterations in an until_fail script?

sunkuamzn:
bot:aws:retest

sunkuamzn:
bot:aws:retest

sunkuamzn:
bot:aws:retest

sunkuamzn:
bot:aws:retest

prov/sm2/src/sm2.h
a-szegel: I would ignore this for now.

prov/sm2/src/sm2_progress.c
```diff
@@ -303,6 +470,33 @@ void sm2_progress_recv(struct sm2_ep *ep)
 				sm2_freestack(sm2_mmap_ep_region(map, ep->gid)),
 				xfer_entry);
 			continue;
+		} else if (xfer_entry->hdr.proto ==
```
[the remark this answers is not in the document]
a-szegel: after following back... this is low priority

sunkuamzn:
bot:aws:retest

prov/sm2/src/sm2.h
lrbison: I don't understand why the `sm2_generic_format` function should modify op_flags, and it doesn't seem to here or in your final commit.

sunkuamzn:
@aingerson are the Intel CI failures real?

aingerson:
@sunkuamzn The CI failure was the tcp tagged bw failure that has been popping up recently. Sean is looking into it. By the way, our new CI will display which step actually failed so you can see that it was tcp (the other two are the parent steps of the test, I will see if we can just print the one).

a-szegel (changes requested):
final comments, then I am good

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,178 @@
...
+	struct sm2_ipc_dev_to_host_data *dev_to_host_data;
+	struct ipc_info *ipc_info;
+	uint64_t device;
+	void *device_ptr = rx_entry->iov[0].iov_base;
+	void *base;
+	int ret;
+
+	if (cuda_is_addr_valid(device_ptr, &device, NULL)) {
+		/* The destination address is a valid CUDA address */
+		xfer_entry->hdr.proto = sm2_proto_ipc_dev_to_host;
```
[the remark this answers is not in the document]
sunkuamzn: there's actually not much here. I could move the assert statement inside `sm2_progress_ipc_host_to_dev` but I would have to pass in `cma_data` to that function..

a-szegel (approved):

prov/sm2/src/sm2.h
lrbison: That is in `sm2_do_cma`.  In `sm2_generic_format` the op_flag of the caller should not be modified, but passing a pointer to the int suggests it can be.

I will propose a separate follow-up later, I don't think this small nit needs to hold up this merge.

prov/sm2/src/sm2_progress.c
```diff
@@ -42,6 +42,193 @@
...
+		if (write)
+			ret = ofi_process_vm_writev(
+				pid, rx_entry->iov, rx_entry->count,
+				cma_data->iov, cma_data->iov_count, 0);
+		else
+			ret = ofi_process_vm_readv(
+				pid, rx_entry->iov, rx_entry->count,
+				cma_data->iov, cma_data->iov_count, 0);
+		if (ret < 0) {
+			if (errno == 14) {
```
lrbison, line 176: I think this should test `errno == EFAULT` rather than 14.

lrbison (approved):

prov/sm2/src/sm2_progress.c
sunkuamzn: Fixed

prov/sm2/src/sm2.h
sunkuamzn: Fixed

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
+	struct sm2_ipc_host_to_dev_data *host_to_dev_data;
+	struct ipc_info *ipc_info;
+	uint64_t device;
+	void *device_ptr = rx_entry->iov[0].iov_base;
+	struct ofi_mr **mr = (struct ofi_mr **) rx_entry->desc;
+	enum fi_hmem_iface iface = mr[0]->iface;
+	void *base;
+	int ret;
+
+	if (hmem_ops[iface].is_addr_valid(device_ptr, &device, NULL)) {
```
aingerson, line 60: Please add a similar ofi_is_addr_valid(iface...) call in ofi_hmem.h so we are consistent with how we call into the hmem ops array).

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
+		/* The destination address is a valid device address */
```
aingerson, line 61: Also unnecessary comment

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
-			ret = ofi_process_vm_readv(pid, local, local_cnt,
-						   remote, remote_cnt, 0);
+			ret = ofi_process_vm_readv(
+				pid, rx_entry->iov, rx_entry->count,
+				cma_data->iov, cma_data->iov_count, 0);
 		if (ret < 0) {
-			FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "CMA error %d\n",
-				errno);
-			return -FI_EIO;
+			if (errno == EFAULT) {
```
aingerson, line 176: Seems dramatic to try CMA and fail. We know the iface for the destination at this point so we don't have to fail CMA to fallback. We can just check rx_entry->desc (ofI_mr *) and fallback before calling the cma_loop function

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
+					return -FI_EIO;
+				}
+
+				err = sm2_progress_ipc_host_to_dev(
+					ep, xfer_entry, rx_entry,
+					cma_data->iov[0].iov_base);
+				if (err == FI_SUCCESS)
+					*ipc_host_to_dev = true;
+				return err;
+			} else {
```
aingerson, line 198: drop else, always returns in if

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
 #include "sm2.h"
 #include "sm2_fifo.h"
 
-static inline int sm2_cma_loop(pid_t pid, struct iovec *local,
-			       unsigned long local_cnt, struct iovec *remote,
-			       unsigned long remote_cnt, size_t total,
-			       bool write)
+static int sm2_progress_ipc_host_to_dev(struct sm2_ep *ep,
```
aingerson, line 45: This name seems misleading. This sounds like you are processing a xfer_entry that had proto = ipc_host_to_dev which is not what you're doing here. This is where you send back the entry to fallback to the IPC call.

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
+
+		ipc_info->base_addr = (uintptr_t) base;
+		ipc_info->offset = (uintptr_t) device_ptr -
+				   (uintptr_t) ipc_info->base_addr;
+
+		/* Send xfer_entry with IPC handle back to sender */
+		sm2_fifo_write(ep, sender_gid, xfer_entry);
+
+		return FI_SUCCESS;
+	} else {
```
aingerson, line 105: drop else

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
+		assert(1 == 0);
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "CMA error %d\n", errno);
+		return -FI_EIO;
+	}
+}
+
+static int sm2_ipc_host_to_dev_hmem_memcpy(struct sm2_ep *ep,
```
aingerson, line 112: Can we reuse/consolidate this with the normal ipc path? Same with the get base addr/open path. Also some of these names are getting long. Consolidating would help simplify them/make them more generic.

prov/sm2/src/sm2_progress.c
```diff
@@ -42,52 +42,187 @@
...
+		(struct sm2_cma_data *) xfer_entry->user_data;
+
+	pid_t pid = entries[xfer_entry->hdr.sender_gid].pid;
 
 	while (1) {
 		if (write)
-			ret = ofi_process_vm_writev(pid, local, local_cnt,
-						    remote, remote_cnt, 0);
+			ret = ofi_process_vm_writev(
+				pid, rx_entry->iov, rx_entry->count,
```
aingerson, line 169: how does the fallback IPC protocol deal with multiple target buffers?

prov/sm2/src/sm2_progress.c
```diff
@@ -211,18 +346,23 @@ static int sm2_start_common(struct sm2_ep *ep,
...
+				xfer_entry->hdr.cq_data);
 	}
 
 	if (ret) {
 		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
 			"Unable to process rx completion\n");
 	}
 
-	if (return_xfer_entry) {
+	if (return_xfer_entry && !ipc_host_to_dev) {
```
aingerson, line 365: drop brackets

prov/sm2/src/sm2_progress.c
```diff
@@ -267,12 +407,42 @@ static int sm2_alloc_xfer_entry_ctx(struct sm2_ep *ep,
...
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_mmap *map = &av->mmap;
 	struct fid_peer_srx *peer_srx = sm2_get_peer_srx(ep);
 	struct fi_peer_rx_entry *rx_entry;
 	struct sm2_av *sm2_av;
 	fi_addr_t addr;
 	int ret;
 
+	/* sm2_proto_ipc_host_to_dev protocol means that the sender received an
+	 * IPC handle from the receiver. So the sender needs to do a CUDA memcpy
```
aingerson, line 420: Personally, I would drop the comment entirely since that's the definition of this protocol. If an explanation needs to go anywhere, it should just be where the protocol is defined/in the header.

prov/sm2/src/sm2_progress.c
```diff
@@ -267,12 +407,42 @@ static int sm2_alloc_xfer_entry_ctx(struct sm2_ep *ep,
...
+			FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+				"Device memcpy in host to device IPC protocol "
+				"failed\n");
+			smr_freestack_push(
+				sm2_freestack(sm2_mmap_ep_region(map, ep->gid)),
+				xfer_entry);
+			goto out;
+		}
+
+		/* Generate send completion on sender side */
```
aingerson, line 434: drop comment

prov/sm2/src/sm2_progress.c
```diff
@@ -267,12 +407,42 @@ static int sm2_alloc_xfer_entry_ctx(struct sm2_ep *ep,
...
+		ret = sm2_complete_tx(ep, (void *) xfer_entry->hdr.context,
+				      xfer_entry->hdr.op,
+				      xfer_entry->hdr.op_flags);
+
+		/* Send a message to receiver with protocol
```
aingerson, line 439: Again, this is definition of the protocol itself, so I would drop the comment.

aingerson (commented):
I would add more in each commit message (generally speaking, I would say no commit should be blank unless it's absolutely self-explanatory)
In general, my biggest suggestion would be to rework the op_flags (setting them and unsetting them in order to trigger/avoid completions in specific places) and also streamlining the overlap between the normal ipc path and fallback ipc, especially in terms of how we pass variables around or how we interpret the user_data. I think ideally we reduce the number of ways we can interpret the user_data so we're not casting it left and right. The whole xfer_entry is 4k and we don't need any extra for the payload for cma/ipc so I don't think there's much harm in including all the variables we might need for cma+ipc in case of the fallback so we don't have to move the cma iov ptr to the host pointer and other weird adjustments. I think it could be simplified by putting it all into one, sending the CMA, if the target is on the device, format the ipc section and send it back to the sender with almost exactly what they sent, just with the ipc info also. I think that would make it a lot easier to follow and read.

prov/sm2/src/sm2.h
```diff
@@ -192,7 +192,7 @@ static inline void sm2_generic_format(struct sm2_xfer_entry *xfer_entry,
 {
 	xfer_entry->hdr.op = op;
 	/* We only care about lower 32 bits */
-	xfer_entry->hdr.op_flags = (uint32_t) op_flags;
+	xfer_entry->hdr.op_flags = op_flags;
```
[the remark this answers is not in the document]
sunkuamzn: this was a mistake. fixed

prov/sm2/src/sm2_ep.c
```diff
@@ -213,6 +213,42 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+			  size_t total_len, void *context)
+{
+	ssize_t ret;
+	struct sm2_xfer_entry *xfer_entry;
+
+	ret = sm2_pop_xfer_entry(ep, &xfer_entry);
+	if (ret)
+		return ret;
+
+	*op_flags |= FI_DELIVERY_COMPLETE;
```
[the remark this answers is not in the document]
sunkuamzn: That's a great idea. I can make a separate PR for that after this is merged.

prov/sm2/src/sm2.h
```diff
@@ -130,6 +135,13 @@ struct sm2_cma_data {
 	struct iovec iov[SM2_IOV_LIMIT];
 };
 
+struct sm2_ipc_host_to_dev_data {
+	struct ipc_info ipc_info;
+	void *host_ptr;
+	void *rx_context;
+	uint64_t rx_flags;
```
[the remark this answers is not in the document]
sunkuamzn: i'm using just rx_entry, like you suggested in a different comment. rx_flags are needed when generating the receive completion.

prov/sm2/src/sm2_progress.c
sunkuamzn: fixed

prov/sm2/src/sm2_progress.c
sunkuamzn: thanks for catching. i've added this.

prov/sm2/src/sm2.h
```diff
@@ -90,6 +90,11 @@ enum {
 	sm2_proto_return,
 	sm2_proto_cma,
 	sm2_proto_ipc,
+	/* Used by receiver to send the IPC handle to the sender */
+	sm2_proto_ipc_host_to_dev,
+	/* Used by sender to notify the receiver that the device memcpy is
+	   complete */
+	sm2_proto_ipc_host_to_dev_ack,
```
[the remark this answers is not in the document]
sunkuamzn: I've addressed this by getting rid of the two new protocols and instead using flags in the header. I split the 32 bit proto into 16 bit proto and 16 bit proto_flags. So now instead of setting a different protocol every time, I just change the flags.

This does make the recv path a bit messy but we are planning a refactor where we switch on protocol before switching on op and that would make this a lot cleaner.

prov/sm2/src/sm2_progress.c
sunkuamzn: not using this function anymore, as we discussed

sunkuamzn:
bot:aws:retest

sunkuamzn:
Continued in https://github.com/ofiwg/libfabric/pull/9103

PR #8921, opened 2023-05-11
prov/sm2: Switch sm2_verify_peer to return ssize_t

a-szegel:
This removes the implicit cast from int to ssize_t in the critical sm2 path saving an instruction.


aingerson (approved):
I'm fine with the change, but I doubt it will save you an instruction.
It would also have been nice to have that in the other PR which already changed the return from a gid to an int...

a-szegel:
The negative sign goes from the 2^31'st bit to the 2^63'ed bit in the implicit conversion.

PR #8925, opened 2023-05-12
prov/shm: fix cuda rma path

wenduwan:
This patch consists of 2 changes:
1. Do not use inline/inject for read operations. The previous gdrcopy optimization mistakenly activated this path.
2. Explicitly assert inject protocol and read op are mutually exclusive.

We need to devise another approach to optimize cuda read operation other than the current inline/inject proto.

Currently evaluating the change.

### DIFF
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -275,6 +275,8 @@ static int smr_progress_inject(struct smr_cmd *cmd, struct ofi_mr **mr,
 	size_t inj_offset;
 	ssize_t hmem_copy_ret;
 
+	assert(cmd->msg.hdr.op != ofi_op_read_req);
+
 	inj_offset = (size_t) cmd->msg.hdr.src_data;
 	tx_buf = smr_get_ptr(ep->region, inj_offset);
 
@@ -283,15 +285,9 @@ static int smr_progress_inject(struct smr_cmd *cmd, struct ofi_mr **mr,
 		return err;
 	}
 
-	if (cmd->msg.hdr.op == ofi_op_read_req) {
-		hmem_copy_ret = ofi_copy_from_mr_iov(tx_buf->data,
-					cmd->msg.hdr.size, mr, iov,
-					iov_count, 0);
-	} else {
-		hmem_copy_ret = ofi_copy_to_mr_iov(mr, iov, iov_count, 0,
-					tx_buf->data, cmd->msg.hdr.size);
-		smr_release_txbuf(ep->region, tx_buf);
-	}
+	hmem_copy_ret = ofi_copy_to_mr_iov(mr, iov, iov_count, 0, tx_buf->data,
+	                                   cmd->msg.hdr.size);
+	smr_release_txbuf(ep->region, tx_buf);
 
 	if (hmem_copy_ret < 0) {
 		FI_WARN(&smr_prov, FI_LOG_EP_CTRL,
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -638,10 +638,6 @@ int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
                      bool gdrcopy_avail, uint32_t op, uint64_t total_len,
                      uint64_t op_flags)
 {
-	/* TODO: Move gdrcopy check out of non-cuda fast paths */
-	if (gdrcopy_avail && total_len <= smr_env.max_gdrcopy_size)
-		return total_len <= SMR_MSG_DATA_LEN ? smr_src_inline : smr_src_inject;
-
 	if (op == ofi_op_read_req) {
 		if (use_ipc)
 			return smr_src_ipc;
@@ -649,6 +645,10 @@ int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
 			return smr_src_iov;
 		return smr_src_sar;
 	}
+
+	/* TODO: Move gdrcopy check out of non-cuda fast paths */
+	if (gdrcopy_avail && total_len <= smr_env.max_gdrcopy_size)
+		return total_len <= SMR_MSG_DATA_LEN ? smr_src_inline : smr_src_inject;
 
 	if (op_flags & FI_INJECT) {
 		if (op_flags & FI_DELIVERY_COMPLETE)


wzamazon:
Looks good to me. Why is it still in draft?

prov/shm/src/smr_progress.c
```diff
@@ -283,15 +285,9 @@ static int smr_progress_inject(struct smr_cmd *cmd, struct ofi_mr **mr,
 		return err;
 	}
 
-	if (cmd->msg.hdr.op == ofi_op_read_req) {
```
shijin-aws, line 286: I see this line is there 3 years ago, which means inject protocol may support read some time. Though from the select_proto function it seems inject was not used for read op. @aingerson do you have ideas here?

shijin-aws:
> Looks good to me. Why is it still in draft?

I think we need to understand why inject protocol does not work for read first.

prov/shm/src/smr_progress.c
aingerson: Reads used to go through inject but once SAR was added it became redundant to keep the inject read support since the SAR buffer acted as a larger inject buffer. This is likely an artifact from that support that was not removed

shijin-aws: We are seeing segfault when using inject protocol + gdrcopy in rma read for shm. @wenduwan may have more details but I am not sure whether it's because inject protocol does not work for read or some other bugs when introducing gdrcopy to shm.

aingerson: I'm guessing it has to do with the inject+read path. When RMA reads went through inject, we had to add a response entry to track when the operation was done by the receiver in order to copy the data back. When RMA read through inject was removed, that accompanying response entry was removed because it would go through SAR. So I'm guessing your segfault is something to do with that. The inject protocol does not have delivery complete support (ie something that needs notification after the sender has processed the request). Any RMA reads have to go through SAR which does have that support. I think the issue is that the gdr copy check is before the read check and should probably be after (and probably inside the RMA read check as well).

aingerson (approved):
@wenduwan Thanks for finding and fixing!

prov/shm/src/smr_progress.c
shijin-aws: I see, the stacktrace:

```
#0  0x00007efc3b84c9b0 in memcpy_uncached_load_sse41 () from /lib64/libgdrapi.so
#1  0x00007efc3b84b98a in gdr_copy_from_mapping () from /lib64/libgdrapi.so
#2  0x00007efc456194a3 in cuda_gdrcopy_impl (handle=45034880, devptr=0x7efbbc002804, hostptr=0xeb5930c9404f, len=2048, dir=GDRCOPY_FROM_DEVICE) at src/hmem_cuda_gdrcopy.c:275
#3  0x00007efc45619515 in cuda_gdrcopy_from_dev (handle=45034880, hostptr=0xeb5930c9404f, devptr=0x7efbbc002804, len=2048) at src/hmem_cuda_gdrcopy.c:292
#4  0x00007efc45615c82 in ofi_copy_mr_iov (mr=0x7ffd98248060, iov=0x7ffd98248090, iov_count=1, offset=0, buf=0xeb5930c9404f, size=2048, dir=0) at src/hmem.c:308
#5  0x00007efc45615d8b in ofi_copy_from_mr_iov (dest=0xeb5930c9404f, size=4096, mr=0x7ffd98248060, iov=0x7ffd98248090, iov_count=1, iov_offset=0) at src/hmem.c:330
#6  0x00007efc4572a989 in smr_format_inject (cmd=0x7efc30c46340, mr=0x7ffd98248060, iov=0x7ffd98248090, count=1, smr=0x7efc30c00000, tx_buf=0xeb5930c9404f) at prov/shm/src/smr_ep.c:354
#7  0x00007efc4572b8ec in smr_do_inject (ep=0x28aa970, peer_smr=0x7efc30c00000, id=0, peer_id=0, op=2, tag=0, data=0, op_flags=16777216, desc=0x7ffd98248060, iov=0x7ffd98248090, iov_count=1,
    total_len=2048, context=0x2af31c0, cmd=0x7efc30c46340) at prov/shm/src/smr_ep.c:704
#8  0x00007efc457250c5 in smr_generic_rma (ep=0x28aa970, iov=0x7ffd98248090, iov_count=1, rma_iov=0x7ffd982480a0, rma_count=1, desc=0x7ffd98248060, addr=0, context=0x2af31c0, op=2, data=0,
    op_flags=16777216) at prov/shm/src/smr_rma.c:195
#9  0x00007efc457252af in smr_read (ep_fid=0x28aa970, buf=0x7efbbc002804, len=2048, desc=0x2af2ec0, src_addr=0, addr=139635963471880, key=8388871, context=0x2af31c0) at prov/shm/src/smr_rma.c:237
#10 0x00007efc456ba51e in fi_read (ep=0x28aa970, buf=0x7efbbc002804, len=2048, desc=0x2af2ec0, src_addr=0, addr=139635963471880, key=8388871, context=0x2af31c0) at ./include/rdma/fi_rma.h:101
```

aingerson: @shijin-aws That isn't what I would expect... Off the top of my head I can't think of any reason why gdr would struggle on the copy to the inject buffer. Unless gdr copy doesn't like that nothing was put in that buffer in the first place? If it's a read, the contents of the source buffer haven't been touched or initialized yet.

shijin-aws: This only happens beyond certain message size. It was running ok until 2048 bytes

```
fi_rma_bw -o read -p efa -S all -m -E --pin-core 1-2 -D cuda queue-p4d24xlarge-st-p4d24xlarge-1 :
- { xfer_size: 1, iterations: 20000, total: 20000, time: 0.050067, MB/sec: 0.399465, usec/xfer: 2.503350, Mxfers/sec: 0.399465 }
...
- { xfer_size: 128, iterations: 20000, total: 2560000, time: 0.209822, MB/sec: 12.200818, usec/xfer: 10.491100, Mxfers/sec: 0.095319 }
- { xfer_size: 192, iterations: 20000, total: 3840000, time: 0.207718, MB/sec: 18.486602, usec/xfer: 10.385900, Mxfers/sec: 0.096284 }
- { xfer_size: 256, iterations: 20000, total: 5120000, time: 0.399262, MB/sec: 12.823660, usec/xfer: 19.963100, Mxfers/sec: 0.050092 }
- { xfer_size: 384, iterations: 20000, total: 7680000, time: 0.435381, MB/sec: 17.639722, usec/xfer: 21.769051, Mxfers/sec: 0.045937 }
- { xfer_size: 512, iterations: 20000, total: 10240000, time: 0.439891, MB/sec: 23.278494, usec/xfer: 21.994551, Mxfers/sec: 0.045466 }
- { xfer_size: 768, iterations: 20000, total: 15360000, time: 0.513002, MB/sec: 29.941404, usec/xfer: 25.650101, Mxfers/sec: 0.038986 }
- { xfer_size: 1024, iterations: 20000, total: 20480000, time: 0.663388, MB/sec: 30.871828, usec/xfer: 33.169399, Mxfers/sec: 0.030148 }
- { xfer_size: 1536, iterations: 20000, total: 30720000, time: 0.838611, MB/sec: 36.632002, usec/xfer: 41.930550, Mxfers/sec: 0.023849 }
Segmentation fault (core dumped)
```

I think @wenduwan found something related to MR range && the alignment. I will let him comment here when he is online.

wenduwan:
> Looks good to me. Why is it still in draft?

Was waiting for CI. Just marked it ready.

prov/shm/src/smr_progress.c
wenduwan: Thanks both for your insights. To add to Shi's description, the segfault was due to `txbuf` not being released when gdrcopy was used on read + inject path - note the `smr_release_txbuf(ep->region, tx_buf);` call was missing on read path. I added prints that showed that, when the segfault happened the `txbuf` address would jump to a large region(uninitialized memory). More prints shows that the inject pool was exhausted - the freestack ends up with [negative free entries](https://github.com/ofiwg/libfabric/blob/f524c9aa4335f9736a17d0dc91cce2abed4119f4/include/ofi_mem.h#L306-L307).

To be clear this is entirely my bug. I didn't test this path for the previous PR(and it eluded our CI as well). Now I realize that gdrcopy does not make sense for RMA read + inline/inject in the current impl. We need to do something special, e.g. inside `smr_progress_ipc` use a bounce buffer to copy out the responder's cuda memory, and then copy it into the requester's cuda memory. We will need to think about it, since SM2 will need a similar change.

wenduwan:
@aingerson Oh that was totally on me. Glad this makes sense to you!

prov/shm/src/smr_progress.c
wenduwan: > related to MR range && the alignment

This was revealed in data validation. I need to test theory, but the idea is, gdr memory pinning could(likely) pins a larger region than requested due to page alignment. Currently we do not account for this difference when calling `fi_mr_regattr`. As a result, gdrcopy could be using memory regions outside of the registered region. IMO this is a legal use case so, after testing this out, we may want to change `mr_attr->mr_iov` after gdr pinning and use the page aligned MR in `fi_mr_regattr`.

Again this is still a hypothesis. Will open a separate PR.

PR #8927, opened 2023-05-12
prov/sm2: Respect sm2 file version

lrbison:
Return -FI_EAVAIL if trying to use an existing SM2 file for coordination but the existing file uses a different version than the current library.

---

From now on, all changes the the coordination file size or structure should be accompanied by a version bump!


prov/sm2/src/sm2_coordination.c
```diff
@@ -62,6 +62,18 @@ static inline bool pid_lives(int pid)
 	return err == 0;
 }
 
+static inline int sm2_mmap_check_version(struct sm2_coord_file_header* tmp_header) {
+	if (tmp_header->file_version == SM2_VERSION)
+		return 0;
```
lrbison, line 67: And before you ask @a-szegel, I did actually go to re-write this function to return a ssize_t as you'd likely prefer, but I find that most of our error paths within `sm2_file_open_or_create` come from system calls which return `int`'s, so it seemed cleaner to be consistent within the function and keep `err` as an `int`, along with all the functions which might set `err`.

aingerson (approved):

lrbison:
Push to appease clang-format.

I really don't like 
```
-static inline int sm2_mmap_check_version(struct sm2_coord_file_header* tmp_header) {
+static inline int
+sm2_mmap_check_version(struct sm2_coord_file_header *tmp_header)
+{
```

Because I often use the `int foo()` as a visual indication while grepping that I've found the function definition 😞

lrbison:
@aingerson is the intel CI a real problem?

aingerson:
@lrbison Per usual, no. Random udp error.

sunkuamzn (approved):

PR #8928, opened 2023-05-12
prov/sm2: Align sm2 region start to page size

lrbison:
Remove the alignment of the list of allocation entries.  These are not page-aligned entries anyways, and are not in the critical path.

Correct the alignment of the regions which hold the xfer entries.  The alignment was previously based on universe size, but I believe that was the result of an erroneous refactor.

---

Note that this is not actually a format change to the sm2 coordination file.  The processes which creates the file gets to pick these pointer offsets and all other processes read and use them.

Also interesting to note that SM2_MAX_UNIVERSE_SIZE is 2048, so likely half a page.
on a machine with 4k pages:
ep_allocation_offset was 2048, now it is 72 bytes.
regions_offset was 149504 bytes (36.5 pages), now it is 151552 bytes (37 pages).

### DIFF
--- a/prov/sm2/src/sm2_coordination.c
+++ b/prov/sm2/src/sm2_coordination.c
@@ -183,6 +183,14 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 	struct sm2_ep_allocation_entry *entries;
 	int fd, common_fd, err, tries, item;
 	bool have_file_lock = false;
+	long int page_size;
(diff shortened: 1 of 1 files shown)


aingerson (approved):

lrbison:
push to appease the clang-format

lrbison:
bot:aws:retest

a-szegel (approved):

PR #8934, opened 2023-05-15
prov/efa: refactor send of multiple packets of same packet type

wzamazon:
For certain packet types such as MEDIUM_RTM, RUNTREAD_RTM and DATA, we send out multiple packets with one call to `ibv_post_send`. Current code use the FI_MORE flag to control the behavior, which require the ep to maintain an internal list of packets that are pending send. In current setting, 1 call to `rxr_pkt_entry_send` does not necessary lead to 1 call to `ibv_post_send`.

This PR has multiple patches that eventually that eliminate the internal list, thus one call to `rxr_pkt_entry_send` will result in 1 call to `ibv_post_send`.

### DIFF
--- a/prov/efa/src/rdm/rxr_pkt_cmd.c
+++ b/prov/efa/src/rdm/rxr_pkt_cmd.c
@@ -49,114 +49,127 @@
  *          dump (for debug only)
  */
 
-/*
- *   rxr_pkt_init_ctrl() uses init functions declared in rxr_pkt_type.h
+/**
+ * @brief fill the "wiredata" field of a pkt_entry by packet type
+ * 
+ * @param[out]		pkt_entry	packet entry whose "wiredata" is to be filled
+ * @param[in]		ope		operation entry
+ * @param[in]		pkt_type	packet type
+ * 
+ * This function call the init functions of each packet type defined in
+ * rxr_pkt_type.h
+ * 
+ * @return 	0 on success
+ * 		negative libfabric error code on error
+ * 
+ * @related rxr_pkt_entry
  */
 static
-int rxr_pkt_init_ctrl(struct efa_rdm_ep *efa_rdm_ep, int entry_type, struct efa_rdm_ope *x_entry,
-		      int ctrl_type, struct rxr_pkt_entry *pkt_entry)
+int rxr_pkt_entry_fill_wiredata(struct rxr_pkt_entry *pkt_entry,
+				int pkt_type,
+				struct efa_rdm_ope *ope)
 {
 	int ret = 0;
 
-	switch (ctrl_type) {
+	switch (pkt_type) {
 	case RXR_READRSP_PKT:
-		ret = rxr_pkt_init_readrsp(efa_rdm_ep, x_entry, pkt_entry);
+		ret = rxr_pkt_init_readrsp(pkt_entry, ope);
 		break;
 	case RXR_CTS_PKT:
-		ret = rxr_pkt_init_cts(efa_rdm_ep, x_entry, pkt_entry);
+		ret = rxr_pkt_init_cts(pkt_entry, ope);
 		break;
 	case RXR_EOR_PKT:
-		ret = rxr_pkt_init_eor(efa_rdm_ep, x_entry, pkt_entry);
+		ret = rxr_pkt_init_eor(pkt_entry, ope);
 		break;
 	case RXR_ATOMRSP_PKT:
-		ret = rxr_pkt_init_atomrsp(efa_rdm_ep, x_entry, pkt_entry);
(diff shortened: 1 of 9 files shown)


wenduwan:
> This PR has multiple patches that eventually that eliminate the internal list, thus one call to rxr_pkt_entry_send will result in 1 call to ibv_post_send.

Is there any limitation to `ibv_post_send` that would require us to redrive the send? What is the benefit of fewer `ibv_post_send` calls?

wzamazon:
>Is there any limitation to ibv_post_send that would require us to redrive the send? 

Not sure I understand the question.

> What is the benefit of fewer ibv_post_send calls?

1 call to `ibv_post_send` with multiple buffers would allow the hardware to optimize send more.

wenduwan:
>    Is there any limitation to ibv_post_send that would require us to redrive the send?

I'm curious if `ibv_post_send` is able to handle arbitrarily long packet lists.

wzamazon:
> I'm curious if ibv_post_send is able to handle arbitrarily long packet lists.

No. it can send no more than `device->efa_attr.max_sq_wr`.

the packet entry pool's size is configured as this size, so as along as we can allocate a packet entry, we should be able to call `ibv_post_send`

prov/efa/src/rdm/rxr_pkt_cmd.c
```diff
@@ -308,7 +308,7 @@ ssize_t rxr_pkt_post_one(struct efa_rdm_ep *efa_rdm_ep, struct efa_rdm_ope *ope,
 	/* If the send succeeded, the function rxr_pkt_entry_send will increase the
 	 * counter in efa_rdm_ep that tracks number of outstanding TX ops.
 	 */
-	err = rxr_pkt_entry_send(efa_rdm_ep, pkt_entry, flags);
+	err = rxr_pkt_entry_send(efa_rdm_ep, &pkt_entry, 1, flags);
```
shijin-aws, line 311: I feel like we should rename this function to rxr_pkt_entry_sendv, to match the libfabric and MPI naming convention

shijin-aws:
Is there a future PR to call `rxr_pkt_entry_send` with multiple pkts?

prov/efa/src/rdm/rxr_pkt_cmd.c
wzamazon: sure. sounds good.

wzamazon:
> Is there a future PR to call rxr_pkt_entry_send with multiple pkts?

Yes. I am working on it.

wzamazon:
The next PR will get rid of the usage of FI_MORE internally.

prov/efa/src/rdm/rxr_pkt_cmd.c
wzamazon: addressed in new revision

shijin-aws (approved):

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -364,82 +364,94 @@ void rxr_pkt_entry_append(struct rxr_pkt_entry *dst,
  * @brief Populate pkt_entry->ibv_send_wr with the information stored in pkt_entry,
```
wenduwan, line 364: > Populate pkt_entry->ibv_send_wr with the information stored in pkt_entry

This is no longer accurate.

prov/efa/src/rdm/rxr_pkt_entry.c
```diff
@@ -364,82 +364,94 @@ void rxr_pkt_entry_append(struct rxr_pkt_entry *dst,
...
+		send = pkt_entry->send;
+
+		/* EFA device supports a maximum of 2 iov/SGE
+		 */
+		assert(send->iov_count <= 2);
+
+		peer = efa_rdm_ep_get_peer(ep, pkt_entry->addr);
+		assert(peer);
+		if (peer->flags & EFA_RDM_PEER_IN_BACKOFF)
+			return -FI_EAGAIN;
```
wenduwan, line 402: Returning error in this loop is problematic.

It reads to me that, it is possible the function leaves `ep->base_ep.xmit_more_wr_tail` in an unknown state if it returns an error, e.g. we appended 10 send_wr to it but then returns an error to the caller.

Am I missing anything? I'm trying to understand the atomicity of this function.

wzamazon: You are right this should have been moved to the before the loop.

wzamazon: addressed in new revision.

wenduwan:
bot:aws:retest

PR #8945, opened 2023-05-18
prov/psm3: filter out FI_TRIGGER when registering rx_flags and tx_flags

### LINKED ISSUE #8944: prov/psm3: `FI_TRIGGER` is used when it shouldn't
**Describe the bug**

I use an *trigger-able* endpoint and then use `inject_writedata` with it.
According to the API documentation,
> An endpoint must be created with the FI_TRIGGER capability in order for triggered operations to be specified. A triggered operation is requested by specifying the FI_TRIGGER flag as part of the operation. Such an endpoint is referred to as a trigger-able endpoint.

However, `psm3` passes the ep flags (including the`FI_TRIGGER` one) to the `write_generic` and therefore generates a triggered operation.
The latter segfaults when looking for the trigger context.

**Expected behavior**
I assume that the `tx_flags` and `rx_flags` should not contain `FI_TRIGGER`

**Output**

```
[1] =================================================================
[1] ==2426286==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x7f7ccce866a9 bp 0x7ffd17fdcad0 sp 0x7ffd17fdca90 T0)
[1] ==2426286==The signal is caused by a READ memory access.
[1] ==2426286==Hint: address points to the zero page.
[1]     #0 0x7f7ccce866a9 in psmx3_trigger_queue_write prov/psm3/src/psmx3_trigger.h:809
[1]     #1 0x7f7ccce89942 in psmx3_write_generic prov/psm3/src/psmx3_rma.c:951
[1]     #2 0x7f7ccce8b421 in psmx3_inject_writedata prov/psm3/src/psmx3_rma.c:1437
[1]     #3 0x40a49b in fi_inject_writedata /gpfs/home/acad/ucl-tfl/tgillis/lib-OFI-1.18.0-dbg/include/rdma/fi_rma.h:159
```

thomasgillis:
hopefully fixes #8944 

I am not entirely sure the patch is complete.
Also could you confirm this is the right place to filter out the tags

### DIFF
--- a/prov/psm3/src/psmx3_ep.c
+++ b/prov/psm3/src/psmx3_ep.c
@@ -566,10 +566,11 @@ int psmx3_ep_open_internal(struct psmx3_fid_domain *domain_priv,
 	psmx3_domain_acquire(domain_priv);
 
 	if (info) {
+		// FI_TRIGGER does not belong to tx/rx_flags
 		if (info->tx_attr)
-			ep_priv->tx_flags = info->tx_attr->op_flags;
+			ep_priv->tx_flags = info->tx_attr->op_flags & ~(FI_TRIGGER);
 		if (info->rx_attr)
-			ep_priv->rx_flags = info->rx_attr->op_flags;
+			ep_priv->rx_flags = info->rx_attr->op_flags & ~(FI_TRIGGER);
 	}
 
 	psmx3_ep_optimize_ops(ep_priv);


shefty:

ToddRimmer:
We have potential concerns regarding this patch.  Trying to track down more information on FI_TRIGGER and what fi_info fields are allowed to have that flag.  Analysis, then review, then testing here may take some time.

acgoldma:
It appears `FI_TRIGGER` is a valid capability for rx/tx according to [fi_endpoint.3.md](https://github.com/ofiwg/libfabric/blob/main/man/fi_endpoint.3.md)
See subsections on [TRANSMIT CONTEXT ATTRIBUTES](https://github.com/ofiwg/libfabric/blob/main/man/fi_endpoint.3.md#transmit-context-attributes) and [RECEIVE CONTEXT ATTRIBUTES](https://github.com/ofiwg/libfabric/blob/main/man/fi_endpoint.3.md#receive-context-attributes)

> The following capabilities apply to the transmit attributes: FI_MSG, FI_RMA, FI_TAGGED, FI_ATOMIC, FI_READ, FI_WRITE, FI_SEND, FI_HMEM, `FI_TRIGGER`, FI_FENCE, FI_MULTICAST, FI_RMA_PMEM, FI_NAMED_RX_CTX, FI_COLLECTIVE, and FI_XPU.

There is a similar list for rx that also includes `FI_TRIGGER`.

Your patch seems to cause psm3 to disable FI_TRIGGER when opening a port, but still advertise that it supports the capability.

Can you supply a reproducer of the issue you are seeing?

thomasgillis:
you are right, it's a valid parameter, the fix is wrong.

The problem is detailed in #8944: when using `fi_write` (no `FI_TRIGGER` flag on the operation itself), it is my impression that `fi_write_generic` [will direct towards `psmx3_trigger_queue_write`](https://github.com/ofiwg/libfabric/blob/36353c3a2b445fe79ebb47ba1f5cac07a986dc26/prov/psm3/src/psmx3_rma.c#LL950C1-L950C1) which then [assumes that the context available to the provider is trigger/threshold context](https://github.com/ofiwg/libfabric/blob/36353c3a2b445fe79ebb47ba1f5cac07a986dc26/prov/psm3/src/psmx3_trigger.h#L809).

From the sentence in the documentation:
> triggered operations require that applications use struct fi_triggered_context as their per operation context parameter
 I conclude that that assumption is not correct (but might be wrong on this)

thomasgillis:
closing this as `struct fi_context` and `struct fi_triggered_context` have the same size and `psm3` uses `FI_CONTEXT` mode.
The original issue is therefore a non-issue

PR #8946, opened 2023-05-18
fabtests/common: fix wrong size used to allocate tx_msg_buf     

wzamazon:
tx_msg_buf was used to stage outgoing message, so its size must be larger than the maximum outgoing message size.
    
Prior to this change, it is allocated by ft_alloc_tx_host_buf(), with the size being MAX(tx_size, FT_MAX_CTRL_MSG_SIZE) * opt.window_size.
    
When FT_OPT_ALLOC_MULT_MR is requested, this value is too small because at that time "tx_mr_size" is max message size.
    
Otherwise, this value is too large, because "tx_size" is enough (no need to mulitply by opts.window_size).
    
This patch removed the function ft_alloc_tx_host_buf(), and pass correct size to ft_hmem_alloc_host() directly.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -494,11 +494,6 @@ static void ft_set_tx_rx_sizes(size_t *set_tx, size_t *set_rx)
 	*set_tx += ft_tx_prefix_size();
 }
 
-int ft_alloc_host_tx_buf(size_t size)
-{
-	return ft_hmem_alloc_host(opts.iface, &tx_msg_buf, size);
-}
-
 void ft_free_host_tx_buf(void)
 {
 	int ret;
@@ -519,6 +514,7 @@ int ft_alloc_msgs(void)
 	int ret;
 	int rma_resv_bytes;
 	long alignment = 1;
+	size_t max_msg_size;
 
 	if (buf)
 		return 0;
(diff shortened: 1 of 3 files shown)


fabtests/common/shared.c
```diff
@@ -563,7 +563,8 @@ int ft_alloc_msgs(void)
 		if (ret)
 			return ret;
 
-		ret = ft_alloc_host_tx_buf(MAX(tx_size, FT_MAX_CTRL_MSG) * opts.window_size);
+		size_t host_tx_size =  (opts.options & FT_OPT_ALLOC_MULT_MR) ? tx_mr_size : tx_size;
```
shijin-aws, line 566: can we move the declaration to the beginning of the function.

shijin-aws (approved):
nit: buff->buf in the commit message

fabtests/common/shared.c
```diff
@@ -563,7 +563,8 @@ int ft_alloc_msgs(void)
...
+		ret = ft_alloc_host_tx_buf(MAX(host_tx_size, FT_MAX_CTRL_MSG) * opts.window_size);
```
shefty, line 567: Please declare variables at the top of functions, so that they can always be quickly located.  Check line lengths and limit to about 80 characters.

As for the changes, this function is becoming a mess with repeated checks.  Just in these 2 lines we have 2 branches, both of which are somewhat duplicated by earlier checks (lines 529-540) and later checks (lines 576-581).  We need to start separating some of the functionality out and make things more consistent and easier to understand.

For example, why isn't buf_size used here?  Isn't that the buffer size to allocate?  Why the host buffer is a different size isn't clear at all.

The ft_alloc_host_tx_buf() call should go away completely.  It's a 1-line function called from 2 places that only obscures what is changing.

Can we get a function that returns the size of the buffer(s) to allocate, size of that reserved for tx, size reserved for rx, etc.?

fabtests/common/shared.c
wzamazon: Done.

fabtests/common/shared.c
wzamazon: > Please declare variables at the top of functions, so that they can always be quickly located. Check line lengths and limit to about 80 characters.

Fixed.

> For example, why isn't buf_size used here? Isn't that the buffer size to allocate? Why the host buffer is a different size isn't clear at all.

Explained in comment.

>Can we get a function that returns the size of the buffer(s) to allocate, size of that reserved for tx, size reserved for rx, etc.?

I am not sure I understand the ask. Do you want a function that set all the size, like tx_size, rx_size, tx_mr_size, rx_mr_size, buf_size, etc?

shefty (approved):

wzamazon:
The aws CI failure is real. It is because rma_bw test need a larger tx_msg_buf. need to update commit message

fabtests/common/shared.c
```diff
@@ -563,7 +559,22 @@ int ft_alloc_msgs(void)
 		if (ret)
 			return ret;
 
-		ret = ft_alloc_host_tx_buf(MAX(tx_size, FT_MAX_CTRL_MSG) * opts.window_size);
+		max_msg_size = (opts.options & FT_OPT_ALLOC_MULT_MR)
+				? tx_mr_size : tx_size;
```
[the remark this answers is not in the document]
wzamazon: No.

This buffer `tx_msg_buf` is used for data setup (in `ft_fill_buf`) and data verification (in `ft_check_buf`).

control messages are not verified, so does not use this buffer.

PR #8949, opened 2023-05-18
prov/sm2 & fabtests: Add FI_ATOMIC support to SM2, and add atomic fabtests

a-szegel:
testing


prov/sm2/src/sm2.h
```diff
@@ -78,6 +79,11 @@
 #define SM2_IOV_LIMIT		4
 #define SM2_INJECT_SIZE		(SM2_XFER_ENTRY_SIZE - sizeof(struct sm2_xfer_hdr))
 
+#define SM2_ATOMIC_INJECT_SIZE \
+	(SM2_INJECT_SIZE - sizeof(struct sm2_atomic_secondary_hdr))
+#define SM2_ATOMIC_COMP_INJECT_SIZE (SM2_ATOMIC_INJECT_SIZE / 2)
```
lrbison, line 84: I was doing something similar for RMA.  Both RMA and Atomic have similar wording in the docs here:

> The requested message size that can be used with fi_inject_atomic is limited by inject_size.

I think that means that fi_inject_atomic, fi_inject_write, and fi_inject all must have the same limit, so we may need to report the minimum of these.

a-szegel: I thought user got size through these functions:
```
static int sm2_atomic_fetch_valid(struct fid_ep *ep, enum fi_datatype datatype,
				  enum fi_op op, size_t *count);

...

	.writevalid = sm2_atomic_valid,
	.readwritevalid = sm2_atomic_fetch_valid,
	.compwritevalid = sm2_atomic_comp_valid,
```

aingerson: Users should definitely query the intended operation/datatype to figure out how many items they can call it on but that doesn't indicate anything about inject. The inject atomic/RMA whether you should pick the smallest value is kind of a grey zone... The way we get around it in shm is that inject size is the full payload buffer so doesn't depend on the headers and since an atomic inject is never a comp, we can use the full inject size so they can be the same. If you're planning on moving to a model similar to that (where the headers are separate than the inject buffer itself) then I think it is a nonissue. What is the plan with the buffer contents?

a-szegel: I don't want to add anything else to the primary header, we should aim for the primary header to be as small as possible. I might even want to take context and user data out of the primary header and make them secondary header attachments if we find that it is faster. I don't mind casting the user data to be the start of the secondary header object. Still undecided on a small message size optimization (separate freestack vs header only freestack).

aingerson: hmm, then I think you're going to want to reduce the inject size to the minimum like Luke says, especially if you're thinking of including other, non-atomic things and including the header here. I think putting the second header in front of the inject buffer is not intuitive, weird, and not worth it. If you're really seeing a performance improvement with that vs leaving the stuff in the header and doing the full 4k for the inject then we can talk...

a-szegel:
So it seems like this is blocked on RMA change

prov/sm2/src/sm2.h
a-szegel: Updated Inject size

prov/sm2/src/sm2_progress.c
```diff
@@ -201,11 +202,143 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+			atomic_entry->atomic_hdr.rma_ioc[ioc_count].count *
+				ofi_datatype_size(
+					atomic_entry->atomic_hdr.datatype),
+			(uintptr_t *) &(
+				atomic_entry->atomic_hdr.rma_ioc[ioc_count]
+					.addr),
+			atomic_entry->atomic_hdr.rma_ioc[ioc_count].key,
+			ofi_rx_mr_reg_flags(
+				xfer_entry->hdr.op,
+				atomic_entry->atomic_hdr.atomic_op));
```
lrbison, line 292: Two nits:
1. this formatting is terrible.  Can we have `fi_ioc *ioc = &(atomic_entry->atomic_hdr.rma_ioc[ioc_count]).  Then all the names are much shorter and will fit in one line.
2. Why not do this in sm2_progress_inject_atomic right before we do the operation?  is it because it's atomic and it all has to be done or not done (I don't think so)?
3. `ioc_count` seems misleading, maybe ioc_index or just `i`?

prov/sm2/src/sm2_progress.c
```diff
@@ -215,6 +348,17 @@ void sm2_progress_recv(struct sm2_ep *ep)
 			break;
 
 		if (xfer_entry->hdr.proto == sm2_proto_return) {
+			if (xfer_entry->hdr.op_flags & SM2_RMA_REQ) {
+				atomic_entry = (struct sm2_atomic_entry *)
+						       xfer_entry->user_data;
+				// TODO Is this copy correct?
+				ofi_copy_to_iov(
```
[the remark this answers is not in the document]
a-szegel: removed todo

prov/sm2/src/sm2.h
```diff
@@ -123,6 +129,36 @@ struct sm2_xfer_entry {
...
+struct sm2_atomic_secondary_hdr {
+	// What atomic operation is being done
+	uint8_t datatype;
+	uint8_t atomic_op;
+
+	// Where the data is going on the receiver side
+	int rma_ioc_count;
+	struct fi_rma_ioc rma_ioc[SM2_IOV_LIMIT];
+
+	// This is needed for writing the result, all data will be kept in the
```
aingerson, line 141: Match the description of field format that you use in the xfer_hdr.

prov/sm2/src/sm2.h
```diff
@@ -123,6 +129,36 @@ struct sm2_xfer_entry {
 	uint8_t user_data[SM2_INJECT_SIZE];
 } __attribute__((packed));
 
+struct sm2_atomic_secondary_hdr {
```
aingerson, line 132: Is there an sm2_atomic_primary_hdr? If not, I would just rename to sm2_atomic_hdr

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,548 @@
...
+		return ret;
+
+	xfer_entry->hdr.proto = sm2_proto_inject;
+
+	sm2_generic_format(xfer_entry, ep->gid, op, 0, 0, op_flags, context);
+	sm2_format_atomic_op(xfer_entry, datatype, atomic_op);
+	sm2_format_atomic_rma_ioc(xfer_entry, rma_ioc, rma_ioc_count);
+	sm2_format_atomic_result(xfer_entry, result_iov, result_count);
+	sm2_format_atomic_data(xfer_entry, iov, iov_count, compare_iov,
+			       compare_count);
```
aingerson, line 136: Two things:
- it seems silly to have all the atomic format functions be separate since they all set like one or two things each and do the same casting of the xfer_entry->user_data. Could we move it to one function? Or just do it all here without the need for functions?
- I think it might make more sense to define the user data as a union so we don't have to keep casting it as the atomic entry.

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,548 @@
...
+	assert(iov_count <= SM2_IOV_LIMIT);
+	assert(result_count <= SM2_IOV_LIMIT);
+	assert(compare_count <= SM2_IOV_LIMIT);
+	assert(rma_ioc_count <= SM2_IOV_LIMIT);
+
+	ret = sm2_verify_peer(ep, addr, &peer_gid);
+	if (ret < 0)
+		return ret;
+
+	/* TODO locking?... */
```
aingerson, line 166: Probably should add this...
The ep->tx_lock controls access to all transmit resources, ie xfer entry fs. Should be holding that here.

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,548 @@
...
+		ofi_ioc_to_iov(compare_ioc, compare_iov, compare_count,
+			       ofi_datatype_size(datatype));
+		total_len *= 2;
+		/* fall through */
+	case ofi_op_atomic_fetch:
+		assert(result_ioc);
+		ofi_ioc_to_iov(result_ioc, result_iov, result_count,
+			       ofi_datatype_size(datatype));
+		op_flags |= SM2_RMA_REQ;
+		op_flags |= FI_DELIVERY_COMPLETE;
```
aingerson, line 183: I can't remember if we've already talked about this but I don't think it makes sense to use the op_flags to hold both sm2 flags and fi flags. In shm we have them as separate. I know you've simplified it to remove the need for the sm2 specific flags which I think is a good direction (I never liked the translation between fi and shm flags) but now you're bringing it back in and using them at the same time. Maybe use FI_REMOTE_READ instead?

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,548 @@
...
+	total_len = count * ofi_datatype_size(datatype);
+	assert(total_len <= SM2_ATOMIC_INJECT_SIZE);
+
+	iov.iov_base = (void *) buf;
+	iov.iov_len = total_len;
+
+	rma_ioc.addr = addr;
+	rma_ioc.count = count;
+	rma_ioc.key = key;
+
```
aingerson, line 305: Probably need locking here too (and looks like your missing a comparable lock in the generic inject path too).

prov/sm2/src/sm2.h
a-szegel: deleted comments

prov/sm2/src/sm2_atomic.c
a-szegel: our locking is a little messed up, and I will fix it in another patch.  We only need locks around CQ, freestack, and the fifo recv queue.  Since 99% of our applications aren't multi-threaded, it makes sense to have 1-2 lock per endpoint.  The receive modifies the freestack which is guarded by the tx_lock... (which is bad).

prov/sm2/src/sm2_atomic.c
a-szegel: > it seems silly to have all the atomic format functions be separate since they all set like one or two things each and do the same casting of the xfer_entry->user_data. Could we move it to one function? Or just do it all here without the need for functions?

Not in love with this but done

> I think it might make more sense to define the user data as a union so we don't have to keep casting it as the atomic entry

I prefer the casts, and we have 3 PR's coming in that are all using casts... I think unions clutter up the structures.

prov/sm2/src/sm2_atomic.c
a-szegel: Done, this might have implications to @lrbison rma PR.

prov/sm2/src/sm2_atomic.c
a-szegel: will add generic inject on different PR.

prov/sm2/src/sm2_atomic.c
lrbison: I agree with Alexia here.  I tried to keep my SM2-specific flags for RMA/SAR separate so I don't think this will impact me.

I think it may be worth while to have 'protocol_flags' in the xfer_entry header, where we can put SM2 specific things.

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,534 @@
...
+	sm2_gid_t peer_gid;
+	ssize_t ret = 0;
+	size_t total_len;
+
+	ret = sm2_verify_peer(ep, dest_addr, &peer_gid);
+	if (ret < 0)
+		return ret;
+
+	total_len = count * ofi_datatype_size(datatype);
+	assert(total_len <= SM2_ATOMIC_INJECT_SIZE);
```
lrbison, line 280: This assert is the only place I could find which ensures user does not request atomics on a buffer which is too large.

I think we need to do a check on size and return EINVAL or similar if the user attempts to send too many atomics at once.  This assert will do nothing in non-debug code.

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,534 @@
...
+	struct iovec compare_iov[SM2_IOV_LIMIT];
+	struct iovec result_iov[SM2_IOV_LIMIT];
+	sm2_gid_t peer_gid;
+	ssize_t ret = 0;
+	size_t total_len;
+
+	assert(iov_count <= SM2_IOV_LIMIT);
+	assert(result_count <= SM2_IOV_LIMIT);
+	assert(compare_count <= SM2_IOV_LIMIT);
+	assert(rma_ioc_count <= SM2_IOV_LIMIT);
```
lrbison, line 140: Similar comment as below about assert vs check.  I think these should be checks which return an error rather than an assert.

lrbison (commented):
Looks good.  The asserts caught my eye.

prov/sm2/src/sm2_atomic.c
aingerson: I think all of the asserts are ok as asserts. The ofi_copy_to/from functions take in the maximum amount of data to copy (inject size) so we shouldn't overwrite the data. Other providers treat errors like this similarly (for example in efa [here](https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_msg.c#LL202)). I don't think it's worth trying to cleanly catch any API misuse errors the application has since it hurts the fast path.

prov/sm2/src/sm2_atomic.c
```diff
@@ -0,0 +1,534 @@
...
+
+	ofi_spin_lock(&ep->tx_lock);
+
+	ret = sm2_do_atomic_inject(ep, peer_gid, ofi_op_atomic, 0, datatype,
+				   atomic_op, &iov, 1, &rma_ioc, 1, NULL, 0,
+				   NULL, 0, total_len, 0);
+
+	if (!ret) {
+		ofi_ep_tx_cntr_inc_func(&ep->util_ep, ofi_op_atomic);
+	}
```
aingerson, line 297: Brackets :(

prov/sm2/src/sm2_progress.c
```diff
@@ -201,11 +202,139 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+		ioc_ptr = &(atomic_entry->atomic_hdr.rma_ioc[i]);
+		ret = ofi_mr_verify(
+			&domain->util_domain.mr_map,
+			ioc_ptr->count *
+				ofi_datatype_size(
+					atomic_entry->atomic_hdr.datatype),
+			(uintptr_t *) &(ioc_ptr->addr), ioc_ptr->key,
+			ofi_rx_mr_reg_flags(
+				xfer_entry->hdr.op,
+				atomic_entry->atomic_hdr.atomic_op));
```
aingerson, line 291: This function got all sorts of wonky. If it's not too much trouble with clang, can you fix how it handles functions called inside parameters? Otherwise, don't really care, just will grumble about it for now.

aingerson (commented):
Just one small formatting change and question/request for future. Otherwise looks good!

prov/sm2/src/sm2_atomic.c
lrbison: Ah, I missed the maximum amount to copy.  Mostly I wanted to make sure we didn't overrun some buffer because application was negligent.  OK, looks good to me.

prov/sm2/src/sm2_progress.c
a-szegel: Lets talk more about what looking better entails... I think the issue is with the multiplication / having our line length be so short.

sunkuamzn: How about something like

```
	size_t datatype_sz;
	uint64_t mr_flags;

	for (i = 0; i < ioc_count; i++) {
		ioc_ptr = &(atomic_entry->atomic_hdr.rma_ioc[i]);

		datatype_sz =
			ofi_datatype_size(atomic_entry->atomic_hdr.datatype);
		mr_flags = ofi_rx_mr_reg_flags(
			xfer_entry->hdr.op, atomic_entry->atomic_hdr.atomic_op);

		ret = ofi_mr_verify(&domain->util_domain.mr_map,
				    ioc_ptr->count * datatype_sz,
				    (uintptr_t *) &(ioc_ptr->addr),
				    ioc_ptr->key, mr_flags);
		if (ret)
			break;

		ioc[i].addr = (void *) ioc_ptr->addr;
		ioc[i].count = ioc_ptr->count;
	}
```

sunkuamzn (approved):

prov/sm2/src/sm2_attr.c
```diff
@@ -49,7 +51,7 @@ struct fi_tx_attr sm2_tx_attr = {
 	.op_flags = SM2_TX_OP_FLAGS,
 	.comp_order = FI_ORDER_NONE,
 	.msg_order = FI_ORDER_SAS,
-	.inject_size = SM2_INJECT_SIZE,
+	.inject_size = SM2_ATOMIC_INJECT_SIZE,
```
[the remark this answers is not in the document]
a-szegel: SM2_ATOMIC_INJECT_SIZE is defined as SM2_INJECT_SIZE - sm2_atomic_header (so SM2_ATOMIC_INJECT_SIZE will always be smaller than SM2_INJECT_SIZE).  I was making an assumption that this PR would go in first, and then Luke would add the min in his PR.

PR #8950, opened 2023-05-19
prov/psm3: add missing `FI_REMOTE_CQ_DATA` for `fi_inject_writedata`

thomasgillis:
in the provider `psm3`, the function `fi_inject_writedata` does not forward the flag `FI_REMOTE_CQ_DATA` to the `write_generic` function, leading to no cq data being actually generated.

Assuming that the function `fi_writedata` is correct, I presume that adding the flag is the right fix

### DIFF
--- a/prov/psm3/src/psmx3_rma.c
+++ b/prov/psm3/src/psmx3_rma.c
@@ -1435,7 +1435,7 @@ STATIC ssize_t psmx3_inject_writedata(struct fid_ep *ep, const void *buf, size_t
 	ep_priv = container_of(ep, struct psmx3_fid_ep, ep);
 
 	return psmx3_write_generic(ep, buf, len, NULL, dest_addr, addr, key, NULL,
-				   ep_priv->tx_flags | FI_INJECT | PSMX3_NO_COMPLETION,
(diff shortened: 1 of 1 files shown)


shefty:

acgoldma:
> @acgoldma - can someone from the psm3 team look at the reported issues and proposed fixes?

Sorry for the late reply, we were in the middle of a release. We will be reviewing these patches now.

ToddRimmer:
On the surface this patch looks appropriate, we'll be setting up some validation tests to ensure the changes work and have no surprises or side effects.

thomasgillis:
Thank you @ToddRimmer

acgoldma:
I was able to produce a failure(hang) while unit testing of this function using `fi_ubertest`. I was then able to verify this patch fixes the hang. More in depth studying of the failure appears that we are waiting for the cq data and not getting it (which would be expected with this bug). 

We are going to run some internal regression testing with this patch and will let you know. 

@shefty From code inspection, this issue appears to also be in `psm` and `psm2` providers.

acgoldma (approved):
Sorry about the delay, this is good to merge.

PR #8952, opened 2023-05-19
prov/tcp: Handle prefetched data after reporting ETRUNC error

shefty:
When progressing receive data, we need to process all data that's been read into the socket prefetch buffer.  Prefetched data won't generate POLLIN events on the socket, since it's already been read from the kernel.

In the case of handling a truncated message, we write an error to the CQ, update the current rx msg to read off the rest of the message data, then return EAGAIN back to the progress loop.  This results in the progress loop exiting.  However, if there's data in the prefetch buffer, we won't return to claim it until the socket signals the POLLIN event.  In some MPICH tests, that doesn't occur, resulting in the test hanging.

Instead of returning to the progress loop, continue trying to read the message data and only return once the message completes or there's no data left to read.

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -385,7 +385,7 @@ static int xnet_handle_truncate(struct xnet_ep *ep)
 		return ret;
 	}
 
-	return -FI_EAGAIN;
+	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


shefty:
@ooststep - you can try this patch for the truncate test failure.  It ran fabtests, but I didn't check it against my truncate tests.

PR #8954, opened 2023-05-19
prov/shm: add memory barrier before update resp->status for SAR

wzamazon:
This patch added an write memory barrier to function
smr_copy_to_sar() and smr_copy_from_sar before updating
resp->status.

This is because resp->status is used by other functions
as indicator for completion of copy. Without the memory barrier,
the copy might not complete when resp-status is updated, which
will lead to data corruption.

### DIFF
--- a/include/ofi_mb.h
+++ b/include/ofi_mb.h
@@ -1,0 +1,204 @@
+/*
+ * Copyright (c) 2018      Los Alamos National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c) 2018      Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
+ * Copyright (c) 2019-2021 Google, LLC. All rights reserved.
+ * Copyright (c) 2019      Triad National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c)           Amazon.com, Inc. or its affiliates.
+ *                         All Rights reserved.
+ * $COPYRIGHT$
+ * Most files in this release are marked with the copyrights of the
+ * organizations who have edited them.  The copyrights below are in no
+ * particular order and generally reflect members of the Open MPI core
+ * team who have contributed code to this release.  The copyrights for
+ * code used under license from other parties are included in the
+ * corresponding files.
+ *
+ * Copyright (c) 2004-2012 The Trustees of Indiana University and Indiana
+ *                         University Research and Technology
+ *                        Corporation.  All rights reserved.
+ * Copyright (c) 2004-2021 The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ * Copyright (c) 2004-2018 High Performance Computing Center Stuttgart,
+ *                         University of Stuttgart.  All rights reserved.
+ * Copyright (c) 2004-2008 The Regents of the University of California.
+ *                         All rights reserved.
+ * Copyright (c) 2006-2018 Los Alamos National Security, LLC.  All rights
+ *                         reserved.
+ * Copyright (c) 2006-2021 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2006-2010 Voltaire, Inc. All rights reserved.
+ * Copyright (c) 2006-2021 Sandia National Laboratories. All rights reserved.
+ * Copyright (c) 2006-2010 Sun Microsystems, Inc.  All rights reserved.
+ *                         Use is subject to license terms.
+ * Copyright (c) 2006-2021 The University of Houston. All rights reserved.
+ * Copyright (c) 2006-2009 Myricom, Inc.  All rights reserved.
+ * Copyright (c) 2007-2017 UT-Battelle, LLC. All rights reserved.
+ * Copyright (c) 2007-2021 IBM Corporation.  All rights reserved.
+ * Copyright (c) 1998-2005 Forschungszentrum Juelich, Juelich Supercomputing
+ *                         Centre, Federal Republic of Germany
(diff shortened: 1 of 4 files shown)


lrbison:
Your reasoning makes sense.  Without the memory barrier the compiler could re-order the assignment which would be bad.

If we need this here in copy_to_sar, we might need something similar in `smr_copy_from_sar` when we set status to SMR_STATUS_SAR_FREE

shijin-aws:
Why it is special to SAR? There are memcpy in other protocols as well?

wzamazon:
> Why it is special to SAR? There are memcpy in other protocols as well?

If I understand protocol correctly, only SAR can have two copy happen at the same time.

wzamazon:
@shijin-aws 

SAR is special because this commit

```
commit d8ffb8991ed12391807aa153fe16aec44c931658
Author: aingerson <alexia.ingerson@intel.com>
Date:   Mon Apr 6 13:51:28 2020 -0700

    prov/shm: minimize locking in progress_resp_entry
    
    The peer smr region only needs to be locked when freeing
    the region's resources. If it's locked, the peer cannot
    make any progress on any transfers either.
    
    Moving this to the end minimizes locked code and allows
    for SAR transfers to make use of their 2 buffers
    simultaneously.
    
    A few extra checks are needed to make sure we are not
    processing a transfer twice if it was not able to
    complete last time due to the try_acquire.
    
    Signed-off-by: aingerson <alexia.ingerson@intel.com>
```
removed some locks around copy_from_sar, other protocol was not impacted by this change

aingerson:
@wzamazon I don't think that's true anymore. When we added DSA support, the SAR protocol was changed from using 2 buffers simultaneously to sending multiple buffers once
Are you able to reproduce this issue with a certain setup?

wzamazon:
>I don't think that's true anymore. When we added DSA support, the SAR protocol was changed from using 2 buffers simultaneously to sending multiple buffers once

Sorry, I do not understand. What is not true any more?

> Are you able to reproduce this issue with a certain setup?

Yes. On ubuntu + ARM cpu + non-debug build, the `fi_rma_bw -o read` test consistently fail with data verification turned on.

wzamazon:
@aingerson

> When we added DSA support, the SAR protocol was changed from using 2 buffers simultaneously to sending multiple buffers once

I am not sure multiple buffer matters for this bug. The problem is `resp->status` was updated to `SAR_READY` when the data was not actually copied to sar bounce buffer

aingerson:
@wzamazon Gotcha. That makes sense then. This looks good to me but I think we should move the smr_atom/sm2_atom to the common code instead of copying it. I know Seth was hesitant because he wasn't completely confident with the implementation but I would rather move it at this point if shm is using it as well.

wzamazon:
@aingerson

>  Gotcha. That makes sense then. This looks good to me but I think we should move the smr_atom/sm2_atom to the common code instead of copying it.

See latest revision. I move the function to `include` (via a separate commit)

aingerson (dismissed):
@a-szegel Note this adds the functions to the common code. Consider moving the rest over/removing from sm2.

include/ofi_mb.h
```diff
@@ -0,0 +1,204 @@
...
+
+#elif defined(HAVE_BUILTIN_MM_ATOMICS)
+
+static inline void ofi_wmb(void)
+{
+	__atomic_thread_fence(__ATOMIC_RELEASE);
+}
+
+#else
+#error "No atomics support found for SHM provider."
```
aingerson, line 203: Actually just noticed this shm-specific error code here and no fallback. That's going to fail non atomics builds. Please add a fallback or else remove the error else but add a requirement of HAVE_ATOMICS on shm

wzamazon: > That's going to fail non atomics builds. Please add a fallback or else remove the error else but add a requirement of HAVE_ATOMICS on shm

Fallback would be hard to implement, because it would require us to write assembly code for every CPU architechture.

I have updated the error message, also updated shm's configure.m4 to check HAVE_ATOMICS

wzamazon: @aingerson  I think I addressed your comments, please take a look. :)

include/ofi_mb.h
```diff
@@ -0,0 +1,204 @@
...
+{
+	atomic_thread_fence(memory_order_release);
+}
+
+#elif defined(HAVE_BUILTIN_MM_ATOMICS)
+
+static inline void ofi_wmb(void)
+{
+	__atomic_thread_fence(__ATOMIC_RELEASE);
+}
```
shefty, line 200: How are 1-line wrapper functions around intrinsic calls derived from the MPICH source?

shefty (changes requested):
I second @lrbison's comment about SMR_STATUS_SAR_FREE.  Why isn't a barrier needed before setting that?

wzamazon:
> I second @lrbison's comment about SMR_STATUS_SAR_FREE. Why isn't a barrier needed before setting that?

I added of `ofi_wmb()` before updating `resp->status` to `SMR_STATUS_SAR_FREE`, and updated commit message.

include/ofi_mb.h
wzamazon: The original header file has all kinds of operations: rmb(), mb(), compare_swap etc. I removed most of them, and kept only wmb()

shefty (approved):
Looks okay to me, but I'm not the expert on this code.

wzamazon:
@aingerson

If you do not object, I plan to merge this PR later today.

aingerson (approved):
Thanks for updating with the config change!

PR #8955, opened 2023-05-19
prov/sm2: Bugfixes for sm2 progress

a-szegel:
This patch fixes three subtle bugs:

1. We were failing to return xfer_entry if we could not write a completion
2. Remove FI_DELIVERY_COMPLETE flag (sender won't write completion) if we fail to process message
3. Remove FI_DELIVERY_COMPLETE flag (sender won't write completion) if we fail to process completion

### DIFF
--- a/prov/sm2/src/sm2_progress.c
+++ b/prov/sm2/src/sm2_progress.c
@@ -97,6 +97,7 @@ static int sm2_start_common(struct sm2_ep *ep,
 
 	if (err) {
 		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "Error processing op\n");
+		xfer_entry->hdr.op_flags &= ~FI_DELIVERY_COMPLETE;
 		ret = sm2_write_err_comp(ep->util_ep.rx_cq, rx_entry->context,
 					 comp_flags, rx_entry->tag, err);
 	} else {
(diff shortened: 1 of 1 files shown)


prov/sm2/src/sm2_progress.c
```diff
@@ -105,10 +106,14 @@ static int sm2_start_common(struct sm2_ep *ep,
 			total_len, comp_buf, xfer_entry->hdr.sender_gid,
 			xfer_entry->hdr.tag, xfer_entry->hdr.cq_data);
 	}
+
 	if (ret) {
+		xfer_entry->hdr.op_flags &= ~FI_DELIVERY_COMPLETE;
```
aingerson, line 111: I don't think I understand this one. If the sender requested delivery complete, then don't you need that to trigger the tx completion, even on error? How do you send back the completion status? Ie in this case, the receive failed so the sender should have an error entry written with a reason of why it failed.

prov/sm2/src/sm2_progress.c
```diff
@@ -97,6 +97,7 @@ static int sm2_start_common(struct sm2_ep *ep,
 
 	if (err) {
 		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "Error processing op\n");
+		xfer_entry->hdr.op_flags &= ~FI_DELIVERY_COMPLETE;
```
aingerson, line 100: I think it makes more sense to add this in the write_err_comp call in case it happens in the future. We should get rid of all the completion flags. You should probably add it in the complete_tx/rx paths so we don't report it over there either.

prov/sm2/src/sm2_progress.c
a-szegel: We don't communicate it back right now

aingerson: I think that's the bigger bug here. Adding this would just be replacing one bug (communicating success on the send when it was actually an error) with another (never writing a send completion on error). I would even prefer for the send to complete in success in this case since the message did arrive to the receiver.

PR #8960, opened 2023-05-22
prov/sm2: Get rid of xfer_ctx_pool in unexpected recv path

a-szegel:
Remove a legacy SHM construct, xfer_ctx_pool. For the unexpected path, SHM needs to copy the message out of the circular buffer and into a temporary context until a receive is posted.  SM2 doesn't need to do this behavior.  SM2 pops the message from the fifo queue, and then can hold onto the peer's xfer_entry until the receiving endpoint posts a buffer that matches it. This change reduces the amount of code on SM2's critical path, and speeds it up a little bit.

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -95,6 +95,7 @@ enum {
  * 	next - fifo linked list next ptr
  * 		This is volatile for a reason, many things touch this
  * 		and we do not want compiler optimization here
+ * 	ep - A pointer to receiver's ep (used on unexp msg path)
  * 	size - Holds total size of message
  * 	cq_data - user defined CQ data
  * 	tag - used for tagged messages
@@ -107,7 +108,10 @@ enum {
  * 	user_data - the message
  */
 struct sm2_xfer_hdr {
-	volatile long int next;
+	union {
+		volatile long int next;
+		struct sm2_ep *ep;
+	};
 	uint64_t size;
 	uint64_t cq_data;
 	uint64_t tag;
@@ -245,7 +249,6 @@ struct sm2_ep {
 	sm2_gid_t gid;
 	ofi_spin_t tx_lock;
 	struct fid_ep *srx;
-	struct ofi_bufpool *xfer_ctx_pool;
 	int ep_idx;
 };
 
--- a/prov/sm2/src/sm2_progress.c
+++ b/prov/sm2/src/sm2_progress.c
@@ -70,8 +70,7 @@ static int sm2_progress_inject(struct sm2_xfer_entry *xfer_entry,
 
 static int sm2_start_common(struct sm2_ep *ep,
 			    struct sm2_xfer_entry *xfer_entry,
-			    struct fi_peer_rx_entry *rx_entry,
-			    bool return_xfer_entry)
+			    struct fi_peer_rx_entry *rx_entry)
 {
 	size_t total_len = 0;
 	uint64_t comp_flags;
@@ -108,11 +107,9 @@ static int sm2_start_common(struct sm2_ep *ep,
 	if (ret) {
 		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
 			"Unable to process rx completion\n");
-	} else if (return_xfer_entry) {
-		/* Return Free Queue Entries here */
-		sm2_fifo_write_back(ep, xfer_entry);
 	}
 
+	sm2_fifo_write_back(ep, xfer_entry);
 	sm2_get_peer_srx(ep)->owner_ops->free_entry(rx_entry);
 
 	return 0;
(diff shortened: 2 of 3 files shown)


prov/sm2/src/sm2_progress.c
```diff
@@ -168,34 +139,28 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
 			peer_srx, addr, xfer_entry->hdr.size,
 			xfer_entry->hdr.tag, &rx_entry);
 		if (ret == -FI_ENOENT) {
-			ret = sm2_alloc_xfer_entry_ctx(ep, rx_entry,
-						       xfer_entry);
-			sm2_fifo_write_back(ep, xfer_entry);
-			if (ret)
-				return ret;
-
+			xfer_entry->hdr.next = (long int) ep;
```
aingerson, line 142: set xfer_entry->hdr.ep

prov/sm2/src/sm2_progress.c
```diff
@@ -168,34 +139,28 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
 		ret = peer_srx->owner_ops->get_msg(
 			peer_srx, addr, xfer_entry->hdr.size, &rx_entry);
 		if (ret == -FI_ENOENT) {
-			ret = sm2_alloc_xfer_entry_ctx(ep, rx_entry,
-						       xfer_entry);
-			sm2_fifo_write_back(ep, xfer_entry);
-			if (ret)
-				return ret;
-
+			xfer_entry->hdr.next = (long int) ep;
```
aingerson, line 151: same here

aingerson (commented):
Just one thing. Otherwise looks good.

a-szegel:
Intel CI failure not related?

aingerson:
@a-szegel Nope, random sockets/rxd errors. Build passed.

sunkuamzn (approved):

prov/sm2/src/sm2.h
```diff
@@ -107,7 +108,10 @@ enum {
  * 	user_data - the message
  */
 struct sm2_xfer_hdr {
-	volatile long int next;
+	union {
+		volatile long int next;
+		struct sm2_ep *ep;
+	};
```
lrbison, line 114: Hm.  The fact that we have to hijack a field in the xfer_hdr for use in unexpected messages strikes me as strange.  This particular field will almost certainly make us segfault if we get it wrong.

I wish we had a few extra bytes in the fi_peer_rx_entry to use for this.  If not, I wonder if it would be cleaner to dedicate a field in xfer_hdr to this for clarity.

In any case, the thing that strikes me as odd is that most of the xfer_hdr is for moving data from peer to peer, and this is really specific to the unexpected message path.

aingerson: I'm ok with adding an extra field if it doesn't impact performance - I'll defer to Seth for that decision.
It is specific to the unexpected message path but the next is specific to having a message actively in the queue. It's just a different path - ie the "next" field is only used when it is in the queue, the "ep" field is only used when it is not in the queue. The field itself just holds a piece of information for who is tracking/using the entry.

a-szegel: No more fields in the header lol!

a-szegel: If we did want to make this little change into a big change... we should add a `void* ep` (or a `struct util_ep *ep`) here:
```
struct fi_ops_srx_owner {
	size_t	size;
	int	(*get_msg)(struct fid_peer_srx *srx, fi_addr_t addr,
			size_t size, struct fi_peer_rx_entry **entry);
	int	(*get_tag)(struct fid_peer_srx *srx, fi_addr_t addr,
	int	(*queue_msg)(struct fi_peer_rx_entry *entry);
	int	(*queue_tag)(struct fi_peer_rx_entry *entry);
	void	(*foreach_unspec_addr)(struct fid_peer_srx *srx,
			fi_addr_t (*get_addr)(struct fi_peer_rx_entry *));

	void	(*free_entry)(struct fi_peer_rx_entry *entry);
};

struct fi_ops_srx_peer {
	size_t	size;
	int	(*start_msg)(struct fi_peer_rx_entry *entry);
	int	(*start_tag)(struct fi_peer_rx_entry *entry);
	int	(*discard_msg)(struct fi_peer_rx_entry *entry);
	int	(*discard_tag)(struct fi_peer_rx_entry *entry);
};
```

aingerson: the problem is that we already have a void * in the entry to specify peer-specific information...

a-szegel: This diff hurts our 96rank osu_alltoallw from 154us -> 162us... we either need to accept this change, or modify the peer API.
```
└─╼ git diff
diff --git a/prov/sm2/src/sm2.h b/prov/sm2/src/sm2.h
index 8e136517..6905ab05 100644
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -108,14 +108,12 @@ enum {
  *     user_data - the message
  */
 struct sm2_xfer_hdr {
-       union {
-               volatile long int next;
-               struct sm2_ep *ep;
-       };
+       volatile long int next;
        uint64_t size;
        uint64_t cq_data;
        uint64_t tag;
        uint64_t context;
+       struct sm2_ep *ep;
        uint32_t op;
        uint32_t op_flags;
        uint32_t proto;
```

lrbison: I can accept the PR as is, I'm fine with the union for next and *ep.

That said, I'm scratching my head trying to understand how it could possibly affect performance.  My mental model is not tracking.

lrbison (approved):

PR #8961, opened 2023-05-22
configure: Increase compiler optimization to -O3

a-szegel:
Increase the level of compiler optimization from -O2 to -O3.


a-szegel:
Motivation:

I noticed that this commit https://github.com/ofiwg/libfabric/commit/a13a35d330c7e3d43fda77152eabfb4be93faf01 put in a slight regression raising the SM2 pt2pt latency from 190ns to 200ns.  I tried removing static, and the inline keywords but I was still seeing the latency degradation.  I set CFLAGS=-O3 and the latency degradation went away.  The improvement is subtle... but it is there, and I think we should take everything that we can get.

a-szegel:
OMPI also uses -O3.

PR #8969, opened 2023-05-24
prov/sm2: Fix startup resource leaks

lrbison:
Fix CID 384050 by allowing common_fd=0 and catching mmap-related errors.

Fix similar leak when mkostemp file creation or map fails.

Change size of mkostemp file's template, as it's not related to FI_NAME_MAX.

check return values of several system calls which were previously ignored.

### DIFF
--- a/prov/sm2/src/sm2_coordination.c
+++ b/prov/sm2/src/sm2_coordination.c
@@ -268,8 +268,12 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 		}
 
 		common_fd = open(SM2_COORDINATION_FILE, O_RDWR);
-		if (common_fd > 0) {
+		if (common_fd >= 0) {
 			tmp_header = sm2_mmap_map(common_fd, map_shared);
+			if (!tmp_header) {
+				close(common_fd);
+				continue;
+			}
 			err = sm2_mmap_check_version(tmp_header);
 			if (err)
 				break;
--- a/prov/sm2/src/sm2_coordination.c
+++ b/prov/sm2/src/sm2_coordination.c
@@ -192,7 +192,9 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 {
 	pthread_mutexattr_t att;
 	struct sm2_mmap map_ours;
-	char template[FI_NAME_MAX];
+	static const int template_len = sizeof(SM2_COORDINATION_DIR) +
+					sizeof("/fi_sm2_pid1234567_XXXXXX") + 1;
+	char template[template_len];
 	struct sm2_coord_file_header *header, *tmp_header;
 	struct sm2_ep_allocation_entry *entries;
 	int fd, common_fd, err, tries, item;
@@ -205,8 +207,7 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 			"Unable to determine page size %ld\n", page_size);
 		return -FI_EINVAL;
 	}
-
-	sprintf(template, "%s/fi_sm2_pid%d_XXXXXX", SM2_COORDINATION_DIR,
+	sprintf(template, "%s/fi_sm2_pid%7d_XXXXXX", SM2_COORDINATION_DIR,
 		getpid());
 
 	/* Assume we are the first process here.
--- a/prov/sm2/src/sm2_coordination.c
+++ b/prov/sm2/src/sm2_coordination.c
@@ -213,16 +213,23 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 	 * Try to open a tmpfile file in the shm directory
 	 */
 	fd = mkostemp(template, O_RDWR);
-	assert(fd > 0);
+	if (fd < 0)
+		goto early_exit;
 	err = ftruncate(fd, sizeof(*header));
 	if (err)
-		return -FI_ENOMEM;
-
+		goto early_exit;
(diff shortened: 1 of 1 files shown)


prov/sm2/src/sm2_coordination.c
```diff
@@ -205,20 +205,20 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
...
-
-	sprintf(template, "%s/fi_sm2_pid%d_XXXXXX", SM2_COORDINATION_DIR,
+	sprintf(template, "%s/fi_sm2_pid%7d_XXXXXX", SM2_COORDINATION_DIR,
 		getpid());
 
 	/* Assume we are the first process here.
 	 * Try to open a tmpfile file in the shm directory
 	 */
 	fd = mkostemp(template, O_RDWR);
-	assert(fd > 0);
```
wenduwan, line 216: Shouldn't we check fd immediately upon return?

prov/sm2/src/sm2_coordination.c
```diff
@@ -192,7 +192,7 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 {
 	pthread_mutexattr_t att;
 	struct sm2_mmap map_ours;
-	char template[FI_NAME_MAX];
+	char template[sizeof(SM2_COORDINATION_DIR) + 32];
```
a-szegel, line 195: Not in love with the magic number, but its not the end of the world for me.

prov/sm2/src/sm2_coordination.c
```diff
@@ -268,8 +268,13 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 		}
 
 		common_fd = open(SM2_COORDINATION_FILE, O_RDWR);
-		if (common_fd > 0) {
+		if (common_fd >= 0) {
```
a-szegel, line 271: This looks like a separate bugfix... plz put in different commit.

prov/sm2/src/sm2_coordination.c
lrbison: it's really: `sizeof(SM2_COORDINATION_DIR) + sizeof("/fi_sm2_pid1234567_XXXXXX") + 1` Which IIRC was something like 27.

prov/sm2/src/sm2_coordination.c
lrbison: we could, or we could let mmap will fail on an invalid fd, and return then.

prov/sm2/src/sm2_coordination.c
lrbison: addressed.

prov/sm2/src/sm2_coordination.c
lrbison: addressed.

prov/sm2/src/sm2_coordination.c
lrbison: 3 commits now.

wenduwan (approved):

a-szegel (approved):

prov/sm2/src/sm2_coordination.c
```diff
@@ -268,8 +268,13 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 		}
 
 		common_fd = open(SM2_COORDINATION_FILE, O_RDWR);
-		if (common_fd > 0) {
```
shijin-aws, line 271: it can be 0? I thought fd should be positive

prov/sm2/src/sm2_coordination.c
```diff
@@ -237,8 +244,15 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 	 * data exchange regions yet.
 	 */
 	err = sm2_mmap_remap(&map_ours, header->ep_regions_offset);
-	if (err)
-		return err;
+
+	if (err) {
+	early_exit:
```
aingerson, line 250: Don't love the look of this error handling. Can you modify to have a normal goto err and have the err case at the bottom with a goto err here instead? Looks kind of weird with this goto in the middle of the function and inside the if. Personal prefernce

prov/sm2/src/sm2_coordination.c
lrbison: lol, so did I, which is why I wrote it that way in the first place.
> The return value of open() is a file descriptor, a small,
       nonnegative integer that is an index to an entry in the process's
       table of open file descriptors.

prov/sm2/src/sm2_coordination.c
lrbison: I moved it to the bottom.  You are right, looks better.

prov/sm2/src/sm2_coordination.c
```diff
@@ -268,8 +268,13 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
 		}
 
 		common_fd = open(SM2_COORDINATION_FILE, O_RDWR);
-		if (common_fd > 0) {
+		if (common_fd >= 0) {
 			tmp_header = sm2_mmap_map(common_fd, map_shared);
+			if (!tmp_header) {
+				/* failed to mmap. */
```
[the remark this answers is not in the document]
lrbison: removed.

aingerson (approved):

shijin-aws (approved):

PR #8971, opened 2023-05-25
[v1.18.x] prov/shm: add memory barrier before update resp->status for SAR

wzamazon:
backport bugfix


wzamazon:
@aingerson Can you take a look?

PR #8972, opened 2023-05-25
prov/psm3: Convert error message to debug output

raffenet:
In pmodels/mpich#6518, a user reported seeing a "No HALs registered" message when using MPICH+libfabric. The message appears to come from the failed initialization of the psm3 provider. Convert it to debug output to avoid spamming applications.

### DIFF
--- a/prov/psm3/psm3/psm2_hal.c
+++ b/prov/psm3/psm3/psm2_hal.c
@@ -572,7 +572,7 @@ static struct _psmi_hal_instance *psm3_hal_get_pi_inst(void)
 		return psm3_hal_current_hal_instance;
 
 	if (!psm3_num_hal) {
-		_HFI_ERROR("No HALs registered\n");
+		_HFI_DBG("No HALs registered\n");
 		return NULL;
 	}
 


shefty:

raffenet:
> The error indicates that PSM3 was not built properly or that constructors are not being run when the provider is loaded. This should remain an error and the issue with the build or library load in the test environment should be investigated. Cannot reproduce this issue, need more details on how you got into this situation. Note this is not a function of NIC discovery, something more fundamental is wrong in the given system.

What details do you need/want to investigate? `config.log`? Capture verbose output from `MPI_Init`?

ToddRimmer:
To be clear, this message means constructors such as this one in hal_verbs.c/verbs_hal.c are not getting run at library load time.
static void __attribute__ ((constructor)) __psmi_hal_verbs_constructor(void)
{
        psm3_hal_register_instance((psmi_hal_instance_t*)&psm3_verbs_hi);
}

Normally PSM3_VERBS or PSM3_SOCKETS are defined and cause the above function or _psmi_hal_sockets_constructor getting called to populate the HAL table at library load time.  By default both are defined, but maybe your build is undefining both?

The message means neither constructor got called, a situation which means PSM3 as built is not usable at all.  As such we consider this a serious error worthy of an error message as there may be other serious but subtle side effects going on in the rest of PSM3 or the rest of lbifabric if other constructors are also not being run.

raffenet:
Would static linking libfabric into my library cause the constructors not to run?

raffenet:
Here are steps to reproduce:
```
./autogen.sh
./configure --prefix=$PWD/i --disable-shared
make -j; make install
FI_PROVIDER=psm3 i/bin/fi_info
```
```
compute-386-08% FI_PROVIDER=psm3 i/bin/fi_info
compute-386-08:pid2800529.fi_info: No HALs registered
fi_getinfo: -61 (No data available)
```

raffenet:
If this should remain an error message then I'll close this and just carry the patch in the MPICH submodule for now. I'll open an issue with the reproducer steps.

ToddRimmer:
We'd prefer to resolve the issue instead of have a patch in MPICH.  What is the distro.  What is the compiler rev.  What is the config.log.  static link should still allow constructors to run, so there is something subtle about your build environment or system.

raffenet:
> We'd prefer to resolve the issue instead of have a patch in MPICH. What is the distro. What is the compiler rev. What is the config.log. static link should still allow constructors to run, so there is something subtle about your build environment or system.

The issue is not limited to the compute nodes here are ANL. These steps work in a fresh Ubuntu 22.04 Docker container, as well.
```
./autogen.sh
./configure --prefix=$PWD/i --disable-shared
make -j; make install
FI_PROVIDER=psm3 i/bin/fi_info
```
```
root@09ac6e8daff5:~/libfabric# FI_PROVIDER=psm3 ./i/bin/fi_info
09ac6e8daff5:pid63998.fi_info: No HALs registered
fi_getinfo: -61 (No data available)
root@09ac6e8daff5:~/libfabric# cat /etc/issue
Ubuntu 22.04.1 LTS \n \l

root@09ac6e8daff5:~/libfabric# uname -a
Linux 09ac6e8daff5 5.10.104-linuxkit #1 SMP PREEMPT Thu Mar 17 17:05:54 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
root@09ac6e8daff5:~/libfabric# gcc --version
gcc (Ubuntu 11.3.0-1ubuntu1~22.04.1) 11.3.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

root@09ac6e8daff5:~/libfabric# ld --version
GNU ld (GNU Binutils for Ubuntu) 2.38
Copyright (C) 2022 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) a later version.
This program has absolutely no warranty.
root@09ac6e8daff5:~/libfabric#
```

StackOverflow suggests that because the constructor functions are not referenced anywhere, the linker is discarding them. https://stackoverflow.com/questions/1202494/why-doesnt-attribute-constructor-work-in-a-static-library

raffenet:
[config.log](https://github.com/ofiwg/libfabric/files/11619402/config.log)

acgoldma:
looks like the `__psmi_hal_verbs_constructor` is grepable in `libfabric.a`, but not in `fi_info`.

It looks like constructors are not always included when linking a static lib to an executable. 

I was able to fix this for fi_info by adding this:
```diff
 util_fi_info_SOURCES = \
        util/info.c
 util_fi_info_LDADD = $(linkback)
+util_fi_info_LDFLAGS = -Wl,--whole-archive,src/.libs/libfabric.a,--no-whole-archive

 util_fi_strerror_SOURCES = \
        util/strerror.c
```
This will link the whole archive instead of just the sections `fi_info` calls.

Not sure why some constructors worked and others did not. 
Specifically all `psm3` constructors, excpet the ones to register the HALs, worked fine.

Structurally, `__psmi_gethostname_lock_constructor()` works the same way as the hals, but seems to work fine.

As a work around for now, using the above patch should work.

PR #8973, opened 2023-05-26
prov/shm: fix bug in configure.m4 in atomics_happy assignment happy

wzamazon:
This patch added the missing assignment of atomics_happy when c11 atomics is not available.


wzamazon:
Without this patch, shm will be disable on older version of compiler that does not support c11 (such as gcc 4.8.5 on centos7)

aingerson (approved):

shijin-aws:
>Without this patch, shm will be disabled on older version of compiler that does not support c11 (such as gcc 4.8.5 on centos7)

We are still testing centos7? You mean shm was disabled on this OS?

wzamazon:
>We are still testing centos7? You mean shm was disabled on this OS?

Yes. which is by the recent commit that introduced memory barrier. No official release was impacted.

PR #8974, opened 2023-05-26
prov/tcp: Handle truncation of active message

shefty:
Fix handling of the case where the user posts a buffer that's too small to receive a message while the message is being actively received into a saved buffer.

xnet_recv_saved() handles the case where the user posts a receive buffer for a message that's currently being saved into a provider buffer.  This is for unexpected message handling where the receive was posted before we could fully read the data from the tcp stream.

The code checks for message truncation and reports an error to the user.  However, it doesn't reset the processing, so that the excess message data is read from the socket.  This results in the excess data being treated as a new message now that truncation errors are no longer fatal to the connection.

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -153,6 +153,36 @@ free_xfer:
 	return NULL;
 }
 
+static int xnet_handle_truncate(struct xnet_ep *ep)
+{
+	struct xnet_xfer_entry *rx_entry;
+	int ret;
+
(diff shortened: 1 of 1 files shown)


shefty:
@ooststep - I'm still trying to figure out a way to verify the updated code path, but this is an attempt to handle the msgtrunc1 failure case, assuming my analysis of the problem is correct.  I'm working to hack fabtests, combined with using a debugger, to force hitting the racy path.

ooststep (approved):
Changes look good.

I was able to reproduce this about 33% of attempts without the fix and am unable to reproduce with this fix.

shefty:
Thanks -- I never got to the point where I was able to verify this path

PR #8981, opened 2023-06-01
prov/tcp: Fix spinning on fi_trywait()

shefty:
When a completion is written to the CQ, the signaling fd is also written to.  This wakes up any thread blocked on the cq.  However, that signal is not cleared if the app is using fi_trywait() followed by blocking on the cq's completion fd.  This can result in the application thread spinning on fi_trywait() + fi_cq_read() loop.

Before returning from fi_trywait(), clear the signaling fd.  Also fail the trywait call if there are any pending completions to read.

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -1429,6 +1429,17 @@ void xnet_progress_all(struct xnet_eq *eq)
 	xnet_progress(&eq->progress, false);
 }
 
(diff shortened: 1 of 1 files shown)


shefty:
Fixes #8947

shefty:
bot:aws:retest

shijin-aws:
bot:aws:retest

shefty:
bot:aws:retest

PR #8983, opened 2023-06-01
common/bsock: Fix error checking for -EAGAIN

shefty:
In ofi_bsock_send(), we need to check if (-ret) == EAGAIN. The negation is missing, as a result, we won't use the buffered space for queuing messages that can't be written directly to the socket.

This problem was reported by coverity as unreachable code.


shefty:
bot:aws:retest

Issue #8976, opened 2023-05-30
How does shm handle inject pool exhaustion?


shijin-aws:
I was debugging a segfault when running [LAMMPS](https://www.lammps.org/) on 48 aws hpc6a instances: 96 ranks per node, 4608 ranks totally, with libfabric main branch. The backtrace shows the error happens when shm tries to copy data from user buffer to bounce buffer (tx_buf->data, allocated in inject buffer pool) in peer smr.

```
(gdb) bt
#0  0x00007f324119c053 in __memcpy_ssse3 () from /lib64/libc.so.6
#1  0x00007f32342394c4 in ofi_memcpy (device=0,
    dest=0xff0b59ebcfd27fd1, src=0x3778440, size=840)
    at ./include/ofi_hmem.h:233
#2  0x00007f323423966f in ofi_copy_from_hmem (iface=FI_HMEM_SYSTEM,
    device=0, dest=0xff0b59ebcfd27fd1, src=0x3778440, size=840)
    at ./include/ofi_hmem.h:358
#3  0x00007f3234239cee in ofi_copy_mr_iov (mr=0x7ffefb0677d0,
    iov=0x7ffefb067800, iov_count=1, offset=0,
    buf=0xff0b59ebcfd27fd1, size=3256, dir=0) at src/hmem.c:315
#4  0x00007f3234239d8b in ofi_copy_from_mr_iov (
    dest=0xff0b59ebcfd27fd1, size=4096, mr=0x7ffefb0677d0,
    iov=0x7ffefb067800, iov_count=1, iov_offset=0) at src/hmem.c:330
#5  0x00007f323434ee0b in smr_format_inject (cmd=0x7f31da70a060,
    mr=0x7ffefb0677d0, iov=0x7ffefb067800, count=1,
    smr=0x7f31da6fc000, tx_buf=0xff0b59ebcfd27fd1)
    at prov/shm/src/smr_ep.c:355
#6  0x00007f323434fd78 in smr_do_inject (ep=0x27bf090,
    peer_smr=0x7f31da6fc000, id=77, peer_id=13, op=1,
    tag=4294967279, data=3661, op_flags=16908288,
    desc=0x7ffefb0677d0, iov=0x7ffefb067800, iov_count=1,
    total_len=840, context=0x3eade68, cmd=0x7f31da70a060)
    at prov/shm/src/smr_ep.c:710
#7  0x00007f323434759d in smr_generic_sendmsg (ep=0x27bf090,
[...]

PR #8984, opened 2023-06-01
prov/shm: Handle Freestack Exhaustion, and ensure smr_signal() on EAGAIN

lrbison:
A series of commits to address https://github.com/ofiwg/libfabric/issues/8976

Ultimately the real cause of the hang may be that there was a path that could use command entries or freestack entries without signalling the peer.  These changes try to address those paths, but also signals the peer any time the peer seems to be out of resources, which should ensure the peer attempts progress.

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -418,12 +418,14 @@ static inline int smr_cma_loop(pid_t pid, struct iovec *local,
 static inline struct smr_inject_buf *
 smr_get_txbuf(struct smr_region *smr)
 {
-	struct smr_inject_buf * txbuf;
+	struct smr_inject_buf *txbuf;
 
 	pthread_spin_lock(&smr->lock);
-	txbuf = smr_freestack_pop(smr_inject_pool(smr));
+	if (!smr_freestack_isempty(smr_inject_pool(smr)))
+		txbuf = smr_freestack_pop(smr_inject_pool(smr));
+	else
+		txbuf = NULL;
 	pthread_spin_unlock(&smr->lock);
-
 	return txbuf;
 }
 
--- a/prov/shm/src/smr_atomic.c
+++ b/prov/shm/src/smr_atomic.c
@@ -125,6 +125,8 @@ static ssize_t smr_do_atomic_inject(struct smr_ep *ep, struct smr_region *peer_s
 	struct smr_resp *resp;
 
 	tx_buf = smr_get_txbuf(peer_smr);
+	if (!tx_buf)
+		return -FI_EAGAIN;
 
 	smr_generic_format(cmd, peer_id, op, 0, 0, op_flags);
 	smr_generic_atomic_format(cmd, datatype, atomic_op);
--- a/prov/shm/src/smr_atomic.c
+++ b/prov/shm/src/smr_atomic.c
@@ -199,8 +199,10 @@ static ssize_t smr_generic_atomic(struct smr_ep *ep,
 		return -FI_EAGAIN;
 
 	ret = smr_cmd_queue_next(smr_cmd_queue(peer_smr), &ce, &pos);
-	if (ret == -FI_ENOENT)
+	if (ret == -FI_ENOENT) {
+		smr_signal(peer_smr);
 		return -FI_EAGAIN;
+	}
 
 	ofi_spin_lock(&ep->tx_lock);
 	total_len = ofi_datatype_size(datatype) * ofi_total_ioc_cnt(ioc, count);
@@ -245,7 +247,7 @@ static ssize_t smr_generic_atomic(struct smr_ep *ep,
 				smr_flags, &ce->cmd);
 		if (ret) {
 			smr_cmd_queue_discard(ce, pos);
-			goto unlock_cq;
+			goto signal;
 		}
 	}
 
@@ -259,8 +261,8 @@ static ssize_t smr_generic_atomic(struct smr_ep *ep,
 
 	smr_format_rma_ioc(&ce->rma_cmd, rma_ioc, rma_count);
 	smr_cmd_queue_commit(ce, pos);
+signal:
 	smr_signal(peer_smr);
-unlock_cq:
 	ofi_spin_unlock(&ep->tx_lock);
 	return ret;
 }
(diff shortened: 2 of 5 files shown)


prov/shm/src/smr_atomic.c
```diff
@@ -259,8 +261,8 @@ static ssize_t smr_generic_atomic(struct smr_ep *ep,
 
 	smr_format_rma_ioc(&ce->rma_cmd, rma_ioc, rma_count);
 	smr_cmd_queue_commit(ce, pos);
-	smr_signal(peer_smr);
 unlock_cq:
```
shijin-aws, line 264: I would rename this as `signal`

prov/shm/src/smr_msg.c
```diff
@@ -426,6 +426,10 @@ static ssize_t smr_generic_inject(struct fid_ep *ep_fid, const void *buf,
 	proto = len <= SMR_MSG_DATA_LEN ? smr_src_inline : smr_src_inject;
 	ret = smr_proto_ops[proto](ep, peer_smr, id, peer_id, op, tag, data,
 			op_flags, NULL, &msg_iov, 1, len, NULL, &ce->cmd);
+	if (ret) {
+		smr_cmd_queue_discard(ce, pos);
+		return -FI_EAGAIN;
+	}
 	smr_cmd_queue_commit(ce, pos);
 
 	assert(!ret);
```
aingerson: remove this assert

prov/shm/src/smr_msg.c
```diff
@@ -420,23 +422,23 @@ static ssize_t smr_generic_inject(struct fid_ep *ep_fid, const void *buf,
...
+		ret = -FI_EAGAIN;
+		goto signal;
+	}
 
 	proto = len <= SMR_MSG_DATA_LEN ? smr_src_inline : smr_src_inject;
 	ret = smr_proto_ops[proto](ep, peer_smr, id, peer_id, op, tag, data,
 			op_flags, NULL, &msg_iov, 1, len, NULL, &ce->cmd);
 	if (ret) {
 		smr_cmd_queue_discard(ce, pos);
-		return -FI_EAGAIN;
```
shijin-aws: do you need to `ret = -FI_EAGAIN` here?

prov/shm/src/smr_progress.c
```diff
@@ -902,7 +902,6 @@ static int smr_progress_cmd_msg(struct smr_ep *ep, struct smr_cmd *cmd)
 		return ret;
 	}
 	ret = smr_start_common(ep, cmd, rx_entry);
-
```
aingerson, line 905: remove this change please 😬

prov/shm/src/smr.h
```diff
@@ -418,15 +418,13 @@ static inline int smr_cma_loop(pid_t pid, struct iovec *local,
...
-	if (smr_freestack_isempty(smr_inject_pool(smr)))
-		goto out;
-
-	txbuf = smr_freestack_pop(smr_inject_pool(smr));
-
-out:
+	if (OFI_LIKELY(!smr_freestack_isempty(smr_inject_pool(smr))))
+		txbuf = smr_freestack_pop(smr_inject_pool(smr));
+	else
+		txbuf = NULL;
```
aingerson, line 427: I'm fine with an if/else here to assign tx_buf but using OFI_LIKELY/UNLIKELY is discouraged. Please drop

aingerson (commented):
Could you also add a bit to the commit message of the first patch please? (I know I'm being picky about my own patch but I didn't proofread it lol sorry)

prov/shm/src/smr_msg.c
lrbison: I don't think so.  If the protocol function returns FI_EAGAIN, then we pass that up the stack, but maybe the protocol function has a different error.  In that case a signal isn't necessary (but doesn't hurt) and we don't hide it behind EAGAIN.

prov/shm/src/smr.h
lrbison: Oh, I didn't realize LIKELY/UNLIKELY was discouraged.  Thanks.  I hope you don't mind, I merged this into your first commit :-)

prov/shm/src/smr_progress.c
lrbison: done!

prov/shm/src/smr_msg.c
lrbison: gone.

prov/shm/src/smr.h
aingerson: Perfect! Thanks!
Yeah in general these compiler optimizations can act a bit weird and can have the opposite effects if you don't REALLY know your code.
See [here](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-g_t_005f_005fbuiltin_005fexpect-4159) for more information, specifically the line:
You may use __builtin_expect to provide the compiler with branch prediction information. In general, you should prefer to use actual profile feedback for this (-fprofile-arcs), as programmers are notoriously bad at predicting how their programs actually perform.

aingerson (approved):

shijin-aws (approved):

prov/shm/src/smr.h
shefty: Exactly!  :)  Even in this case, we don't know if the inject pool will be empty or not when this is called, especially under load.  And if we get this wrong under load, we just made the situation worse.

prov/shm/src/smr_msg.c
```diff
@@ -426,9 +426,12 @@ static ssize_t smr_generic_inject(struct fid_ep *ep_fid, const void *buf,
 	proto = len <= SMR_MSG_DATA_LEN ? smr_src_inline : smr_src_inject;
 	ret = smr_proto_ops[proto](ep, peer_smr, id, peer_id, op, tag, data,
 			op_flags, NULL, &msg_iov, 1, len, NULL, &ce->cmd);
+	if (ret) {
+		smr_cmd_queue_discard(ce, pos);
+		return -FI_EAGAIN;
```
shefty: Why wouldn't we pass through the value of 'ret' here, rather than converting it to EAGAIN?

lrbison: fixed.

wzamazon:
bot:aws:retest

wzamazon:
bot:aws:retest

wzamazon:
bot:aws:retest

PR #8985, opened 2023-06-02
fabtests/pytest: efa test to flood peer during startup

lrbison:
Add a test to flood a peer during startup to ensure good error handling in FI_EAGAIN paths.

---

This test tries to simulate a similar situation as in https://github.com/ofiwg/libfabric/issues/8976, but it's not a perfect reproducer.  In particular main passes this test, but main still had issues on the workloads we were investigating.  Despite that, this is a helpful test, and it does currently error on 1.18.x branch.

The primary reason this test is different than the workload is that the sleep in this test occurs before any exchange of messages.  In the real application we saw an initial exchange, then a flood.

### DIFF
--- a/fabtests/Makefile.am
+++ b/fabtests/Makefile.am
@@ -161,6 +161,7 @@ nobase_dist_config_DATA = \
 	pytest/default/test_sighandler.py \
 	pytest/efa/conftest.py \
 	pytest/efa/efa_common.py \
+	pytest/efa/test_flood_peer.py \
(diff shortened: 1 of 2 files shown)


shijin-aws (approved):

wzamazon (approved):

shijin-aws:
When you backport, you can backport this PR with https://github.com/ofiwg/libfabric/pull/8984 in a single PR. so it can get tested in CI

wzamazon:
bot:aws:retest

PR #8987, opened 2023-06-02
prov/efa: optimize ep lock usage.

shijin-aws:
Currently, efa provider always do ofi_mutex_lock/ofi_mutex_unlock to acquire and release ep lock. However, this is not optimal because such lock is unnecessary unless for multi-thread. This patch makes efa provider use ofi_ep_lock_acquire and ofi_ep_lock_release to acquire/release ep lock, which will be no-ops in single thread situation.


shijin-aws:
> Did you notice any performance improvements?

I am still evaluating.

lrbison (approved):

shijin-aws:
I haven't observed any visible performance difference in all tests I have run so far, but I believe this is the right thing to do to take advantage the threads hint in libfabric.

wzamazon (approved):

PR #8988, opened 2023-06-02
prov/shm: fix DSA enable

aingerson:
When multiple interface iov support was added, DSA support was skipped because of an incorrect MR check. This adds an extra check to make sure all iovs are host buffers in order to enable the DSA path correctly.

### DIFF
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -283,6 +283,17 @@ struct ofi_mr {
 	void *hmem_data;
 };
 
+static inline bool ofi_mr_all_host(struct ofi_mr **mr, size_t count)
+{
+	int i;
+
+	for (i = 0; i < count; i++) {
+		if (mr[i] && mr[i]->iface != FI_HMEM_SYSTEM)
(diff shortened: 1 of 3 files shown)


aingerson:
@jcabr020

include/ofi_mr.h
```diff
@@ -283,6 +283,17 @@ struct ofi_mr {
 	void *hmem_data;
 };
 
+static inline bool ofi_mr_all_host(struct ofi_mr **mr, size_t count)
+{
+	int i;
+
+	for (i = 0; i < count; i++) {
+		if (mr[i] && mr[i] != FI_HMEM_SYSTEM)
```
shefty, line 291: mr[i].iface ?

Does this build cleanly?

aingerson: whoops. it does build cleanly but yup that's a mistake! will fix. Sorry.

aingerson:
@shefty Fixed and everything passing now. Should be good.

PR #8989, opened 2023-06-02
contrib/intel/jenkins: Migrate from Jenkins to Cloudbees Jenkins

zachdworkin:
This set of changes touches all of the code in this folder.
Most Python classes got updated to no longer rely on old bash scripts
The jenkins_config file has become a cloudbees_config with new contents
The Jenkinsfile has been completely overhauled to enable slurm support on the main cluster. This support also contains generified functions that should make it easy to move DAOS and DSA nodes to use the slurm solution.
There are quality of life changes made to the Jenkinsfile to help with upgrading python versions and middleware versions.
DSA is getting re-enabled since the node is back online. It will be used in cloudbees and not the current controller to make the migration smoother
Oneccl testing has changed to use the oneccl team's test suite since some tests got deprecated in their release package


contrib/intel/jenkins/Jenkinsfile
```diff
@@ -202,296 +260,163 @@ pipeline {
...
-                  cd ${env.WORKSPACE}/${SCRIPT_LOCATION}/
-                  python3.7 runtests.py --prov=udp --test=fabtests
-                  python3.7 runtests.py --prov=udp --test=fabtests --ofi_build_mode=dbg
-                  python3.7 runtests.py --prov=udp --test=fabtests --ofi_build_mode=dl
-                  echo "udp completed."
-                )
-              """
+            script {
+              dir ("${env.WORKSPACE}/${SCRIPT_LOCATION}/") {
+                run_fabtest("udp", "cvl", "2", "udp")
```
nikhilnanal, line 334: may be it makes better sense to have a site config variable for hardware type?

PR #8990, opened 2023-06-02
fabtests: make tx_buf and rx_buf aligned to 64 bytes by default

wzamazon:
Prior to this change, tx_buf and rx_buf is not aligned by default. Moreover, tx_buf is usually on the 2nd half of the whole buf. Therefore, any change to buf_size would change the alignment of tx_buf, which leads to the performance change.

For example, recently fabtests added the data verificiaton of rma_bw, for which 4 bytes was added to buf size, which shifted tx_buf's alignment and changed performance.

To address this issue, this patch make tx_buf and rx_buf to align on 64 bytes boundary by default.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -92,7 +92,8 @@ char *buf = NULL, *tx_buf, *rx_buf;
 void *tx_msg_buf = NULL;
 
 char **tx_mr_bufs = NULL, **rx_mr_bufs = NULL;
-size_t buf_size, tx_size, rx_size, tx_mr_size, rx_mr_size;
+size_t buf_size, tx_buf_size, rx_buf_size;
+size_t tx_size, rx_size, tx_mr_size, rx_mr_size;
 int rx_fd = -1, tx_fd = -1;
 char default_port[8] = "9228";
 static char default_oob_port[8] = "3000";
@@ -513,7 +514,7 @@ int ft_alloc_msgs(void)
 {
 	int ret;
 	int rma_resv_bytes;
-	long alignment = 1;
+	long alignment = 64;
 	size_t max_msg_size;
 
 	if (buf)
@@ -526,13 +527,14 @@ int ft_alloc_msgs(void)
 		ft_set_tx_rx_sizes(&tx_mr_size, &rx_mr_size);
 		rx_size = FT_MAX_CTRL_MSG + ft_rx_prefix_size();
 		tx_size = FT_MAX_CTRL_MSG + ft_tx_prefix_size();
-		buf_size = rx_size + tx_size;
+		rx_buf_size = rx_size;
+		tx_buf_size = tx_size;
 	} else {
 		ft_set_tx_rx_sizes(&tx_size, &rx_size);
 		tx_mr_size = 0;
 		rx_mr_size = 0;
-		buf_size = MAX(tx_size, FT_MAX_CTRL_MSG) * opts.window_size +
-			   MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size;
+		rx_buf_size = MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size;
+		tx_buf_size = MAX(tx_size, FT_MAX_CTRL_MSG) * opts.window_size;
 	}
 
 	/* Allow enough space for RMA to operate in a distinct memory
@@ -540,14 +542,20 @@ int ft_alloc_msgs(void)
 	 */
 	rma_resv_bytes = FT_RMA_SYNC_MSG_BYTES +
 			 MAX( ft_tx_prefix_size(), ft_rx_prefix_size() );
-	buf_size += rma_resv_bytes*2;
+	tx_buf_size += rma_resv_bytes;
+	rx_buf_size += rma_resv_bytes;
 
 	if (opts.options & FT_OPT_ALIGN && !(opts.options & FT_OPT_USE_DEVICE)) {
 		alignment = sysconf(_SC_PAGESIZE);
 		if (alignment < 0)
 			return -errno;
-		buf_size += alignment;
+	}
 
+	rx_buf_size = ft_get_aligned_size(rx_buf_size, alignment);
+	tx_buf_size = ft_get_aligned_size(tx_buf_size, alignment);
+
+	buf_size = rx_buf_size + tx_buf_size;
+	if (opts.options & FT_OPT_ALIGN && !(opts.options & FT_OPT_USE_DEVICE)) {
 		ret = posix_memalign((void **) &buf, (size_t) alignment,
 				buf_size);
 		if (ret) {
@@ -555,7 +563,12 @@ int ft_alloc_msgs(void)
 			return ret;
 		}
 	} else {
-		ret = ft_hmem_alloc(opts.iface, opts.device, (void **) &buf, buf_size);
+		/* allocate extra "alignment" bytes, to handle the case
+		 * "buf" returned by ft_hmem_alloc() is not aligned.
+		 */
+		buf_size += alignment;
+		ret = ft_hmem_alloc(opts.iface, opts.device, (void **) &buf,
+				    buf_size);
 		if (ret)
 			return ret;
 
@@ -578,23 +591,19 @@ int ft_alloc_msgs(void)
 		if (ret)
 			return ret;
 	}
+
 	ret = ft_hmem_memset(opts.iface, opts.device, (void *) buf, 0, buf_size);
 	if (ret)
 		return ret;
-	rx_buf = buf;
 
-	if (opts.options & FT_OPT_ALLOC_MULT_MR)
-		tx_buf = (char *) buf + MAX(rx_size, FT_MAX_CTRL_MSG);
-	else
-		tx_buf = (char *) buf +
-			 MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size +
-			 rma_resv_bytes;
-
+	rx_buf = (char *)ft_get_aligned_addr(buf, alignment);
+	tx_buf = rx_buf + rx_buf_size;
 	remote_cq_data = ft_init_cq_data(fi);
 
 	mr = &no_mr;
 	if (!ft_mr_alloc_func && !ft_check_opts(FT_OPT_SKIP_REG_MR)) {
-		ret = ft_reg_mr(fi, buf, buf_size, ft_info_to_mr_access(fi),
+		ret = ft_reg_mr(fi, rx_buf, rx_buf_size + tx_buf_size,
+				ft_info_to_mr_access(fi),
 				FT_MR_KEY, opts.iface, opts.device, &mr,
 				&mr_desc);
 		if (ret)
@@ -2083,16 +2092,18 @@ ssize_t ft_inject(struct fid_ep *ep, fi_addr_t fi_addr, size_t size)
 
 static size_t ft_remote_write_offset(const char *buf)
 {
-	assert(buf >= tx_buf && buf < (tx_buf + buf_size / 2));
-	/* rx_buf area is the first half of the remote region */
+	assert(buf >= tx_buf && buf < (tx_buf + tx_buf_size));
+	/* rx_buf area is at the beginning of the remote region */
 	return buf - tx_buf;
 }
 
 static size_t ft_remote_read_offset(const char *buf)
 {
-	assert(buf >= rx_buf && buf < (rx_buf + buf_size / 2));
-	/* tx_buf area is the latter half of the remote region */
-	return buf - rx_buf + buf_size / 2;
+	assert(buf >= rx_buf && buf < (rx_buf + rx_buf_size));
+	/* We want to read from remote peer's tx_buf area,
+	 * which immediately follow the rx_buf, hence add rx_buf_size
+	 */
+	return buf - rx_buf + rx_buf_size;
 }
 
 ssize_t ft_post_rma(enum ft_rma_opcodes op, char *buf, size_t size,
--- a/fabtests/include/shared.h
+++ b/fabtests/include/shared.h
@@ -42,6 +42,7 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <getopt.h>
+#include <assert.h>
 
 #include <rdma/fabric.h>
 #include <rdma/fi_rma.h>
@@ -507,6 +508,21 @@ static inline bool ft_check_prefix_forced(struct fi_info *info,
 
 	/* Continue if forced prefix wasn't requested. */
 	return true;
+}
+
+static inline
+size_t ft_get_aligned_size(size_t size, size_t alignment)
+{
+	return ((size % alignment) == 0) ?
+		size : ((size / alignment) + 1) * alignment;
+}
+
+static inline
+void *ft_get_aligned_addr(void *ptr, size_t alignment)
+{
+	/* alignment must be power of 2, hence the assertion */
+	assert((alignment & (alignment - 1)) == 0);
+	return (void *)(((uintptr_t)ptr + alignment - 1) & ~((uintptr_t)alignment - 1));
 }
 
 int ft_read_cq(struct fid_cq *cq, uint64_t *cur, uint64_t total,


fabtests/common/shared.c
```diff
@@ -526,36 +532,38 @@ int ft_alloc_msgs(void)
...
 		tx_size = FT_MAX_CTRL_MSG + ft_tx_prefix_size();
-		buf_size = rx_size + tx_size;
+		half_buf_size = MAX(tx_size, rx_size);
 	} else {
 		ft_set_tx_rx_sizes(&tx_size, &rx_size);
 		tx_mr_size = 0;
 		rx_mr_size = 0;
-		buf_size = MAX(tx_size, FT_MAX_CTRL_MSG) * opts.window_size +
-			   MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size;
+		half_buf_size = MAX(MAX(rx_size, FT_MAX_CTRL_MSG), tx_size) * opts.window_size;
```
shefty, line 540: nit: check line length

fabtests/common/shared.c
```diff
@@ -526,36 +532,38 @@ int ft_alloc_msgs(void)
...
 
 	if (opts.options & FT_OPT_ALIGN && !(opts.options & FT_OPT_USE_DEVICE)) {
 		alignment = sysconf(_SC_PAGESIZE);
 		if (alignment < 0)
 			return -errno;
-		buf_size += alignment;
-
+	}
+	half_buf_size = ft_get_aligned_size(half_buf_size, alignment);
+	buf_size = 2*half_buf_size;
```
shefty, line 556: nit: add spaces before/after '*'

fabtests/common/shared.c
```diff
@@ -578,18 +586,13 @@ int ft_alloc_msgs(void)
...
-	rx_buf = buf;
-
-	if (opts.options & FT_OPT_ALLOC_MULT_MR)
-		tx_buf = (char *) buf + MAX(rx_size, FT_MAX_CTRL_MSG);
-	else
-		tx_buf = (char *) buf +
-			 MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size +
-			 rma_resv_bytes;
 
+	rx_buf = (char *)(((uintptr_t)buf + alignment - 1) / alignment * alignment);
```
shefty, line 594: This doesn't look right.  Dividing by alignment, then multiplying by alignment is basically multiplying by 1.  It's also non-trivial why the rx_buf has some weird adjustment, but tx_buf below does not.  Maybe tx_buf should have been rx_buf + half_buf_size?

I think we need some macro here to help with readability as to the intent.

shefty: I think you want something like this:
```
rx_buf = ((uintptr_t) buf + alignment - 1) & ~((uintptr_t) (alignment - 1));
```
The right hand side is an aligned buffer given an input pointer and could be a macro.  I think this is close to the C++ align call, but don't know if that's available for us anywhere.  I didn't see a similar macro in the libfabric source.

fabtests/common/shared.c
wzamazon: fixed

fabtests/common/shared.c
wzamazon: added

fabtests/common/shared.c
wzamazon: Cool, I added a function `ft_get_aligned_addr()` which wrap up your code.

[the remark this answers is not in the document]
wzamazon: You are right, that was the intention. I added a comment for it.

Though I am not sure this is no-op for host memory, buffer returned by `malloc` is not necessarily aligned.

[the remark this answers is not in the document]
sunkuamzn: Host memory is using `posix_memalign`, not `malloc` though

https://github.com/ofiwg/libfabric/blob/5038634893d243ffec86406c8782966c0fdd0ea8/fabtests/common/shared.c#L557-L559

sunkuamzn:
@wzamazon you did not push your new commit

wzamazon:
> @wzamazon you did not push your new commit

You are right. Sorry!

fabtests/common/shared.c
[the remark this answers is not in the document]
wzamazon: >Host memory is using posix_memalign, not malloc though

only when user specified `-l` in command line.

fabtests/common/shared.c
```diff
@@ -578,18 +597,14 @@ int ft_alloc_msgs(void)
...
-
-	if (opts.options & FT_OPT_ALLOC_MULT_MR)
-		tx_buf = (char *) buf + MAX(rx_size, FT_MAX_CTRL_MSG);
-	else
-		tx_buf = (char *) buf +
-			 MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size +
-			 rma_resv_bytes;
 
+	rx_buf = (char *)ft_get_aligned_addr(buf, alignment);
+	assert((rx_buf - buf) >= 0 && (rx_buf - buf) < alignment);
```
sunkuamzn, line 606: I think it's better to assert `alignment > 0` in `ft_get_aligned_addr` than to do it here

fabtests/common/shared.c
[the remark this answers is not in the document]
sunkuamzn: I didn't know that thanks!

But sounds like we're making the `-l` option obsolete if we are going to always align.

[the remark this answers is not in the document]
sunkuamzn: Oh `-l` aligns to page size but your change always aligns to 64 byte boundary. Never mind.
https://github.com/ofiwg/libfabric/blob/5038634893d243ffec86406c8782966c0fdd0ea8/fabtests/common/shared.c#L551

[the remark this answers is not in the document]
wzamazon: Yes.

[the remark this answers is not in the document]
sunkuamzn: My original point still stands though. We are always using `posix_memalign` for host memory. So your calculation should be a no-op no matter the `-l` option

fabtests/common/shared.c
```diff
@@ -526,36 +539,42 @@ int ft_alloc_msgs(void)
...
 			FT_PRINTERR("posix_memalign", ret);
 			return ret;
 		}
 	} else {
-		ret = ft_hmem_alloc(opts.iface, opts.device, (void **) &buf, buf_size);
+		/* allocate extra "alignment" bytes, to handle the case
+		 * "buf" is not aligned.
+		 */
+		ret = ft_hmem_alloc(opts.iface, opts.device, (void **) &buf,
+				    buf_size + alignment);
```
shefty, line 577: It's confusing that buf_size isn't actually the size of the buffer.

fabtests/common/shared.c
shefty: Agree that asserts should be moved into the macro

fabtests/common/shared.c
```diff
@@ -578,18 +597,14 @@ int ft_alloc_msgs(void)
...
-	if (opts.options & FT_OPT_ALLOC_MULT_MR)
-		tx_buf = (char *) buf + MAX(rx_size, FT_MAX_CTRL_MSG);
-	else
-		tx_buf = (char *) buf +
-			 MAX(rx_size, FT_MAX_CTRL_MSG) * opts.window_size +
-			 rma_resv_bytes;
 
+	rx_buf = (char *)ft_get_aligned_addr(buf, alignment);
+	assert((rx_buf - buf) >= 0 && (rx_buf - buf) < alignment);
+	tx_buf = buf + half_buf_size;
```
shefty, line 607: Similar comment as before.  Why is rx_buf aligned, but tx_buf is not?

fabtests/common/shared.c
wzamazon: The assertion here is to make sure the returned buffer is equal to or larger than input. Because the function name `ft_get_aligned_addr` does not tell caller which way it goes

fabtests/common/shared.c
wzamazon: because `half_buf_size` is returned from `get_aligned_size`. So, if `rx_buf` is aligned, `tx_buf` must be aligned.

fabtests/common/shared.c
wzamazon: The comment was trying to explain the reason, does that help?

fabtests/common/shared.c
shefty: I think you want
```tx_buf = rx_buf + half_buf_size```
not *buf* + half_size.  buf isn't guaranteed to be aligned.

File: fabtests/common/shared.c [hunk 7]
```diff
@@ -504,6 +504,19 @@ void ft_free_host_tx_buf(void)
...
+{
+	return ((size % alignment) == 0) ?
+		size : ((size / alignment) + 1) * alignment;
+}
+
+void *ft_get_aligned_addr(void *ptr, size_t alignment)
+{
+	/* alignment must be power of 2, hence the assertion */
+	assert((alignment & (alignment - 1)) == 0);
+	return (void *)(((uintptr_t)buf + alignment - 1) & ~((uintptr_t)alignment - 1));
```
sunkuamzn, line 517: We want to align `ptr` not `buf`

So `(void *)(((uintptr_t)ptr + alignment - 1) & ~((uintptr_t)alignment - 1));`

fabtests/common/shared.c
shefty: I'd go further with this and match std::align (somewhat).  E.g.
```
ft_get_aligned_addr(void *buf, size_t buf_size, size_t alignment, size_t *new_size)
```
This helps clarify that we're aligning into the buffer and the returned buffer has fewer bytes available.

fabtests/common/shared.c
wzamazon: Yes. You are right. Fixed in new revision.

fabtests/common/shared.c
wzamazon: Yes. you are right. Fixed in new revision.

fabtests/common/shared.c
wzamazon: I do not follow. What do we do with `new_size` returned from this function? add an asserion that `new_size <= size`?

shefty: There are a couple options.

The new_size is the size of what we can access starting at rx_buf.  Technically, our memory registration (if needed) starts at rx_buf and extends to new_size.  Or, if we want to be more precise, we expect rx_buf to be usable to size half_buf_size (I think?).  Then tx_buf would start at rx_buf + half_buf_size, and the size of the tx_buf would be new_size - half_buf_size.  The memory registration for rx_buf and tx_buf could be separated.

What I would assert is that new_size >= the size that we need.  I'm not sure what values give us that.  (new_size >= tx_size + rx_size) ?  (new_size >= MAX(tx_size, rx_size)) ?

Or maybe (rx_size < new_size - half_buf_size) && (tx_size <= new_size - half_buf_size) ?

Note that ft_get_aligned_addr() could fail (return null) if the buffer size isn't large enough to meet the alignment requirement.  For example, buf_size = 128, but alignment = 4k, could fail.

fabtests/common/shared.c
shefty: I get the comment and why the adjustment is made.  It's that the variable named 'buf_size' isn't actually the size of the buffer.  I'd modify to:
```
buf_size += alignment;
ret = ft_hmem_alloc(..., buf_size);
```
So it's clear that buf_size is always the size of the buffer.

wzamazon: I see. will do that

wzamazon:
bot:aws:retest

shijin-aws:
bot:aws:retest

fabtests/common/shared.c
wzamazon: @shefty I changed to `buf_size += alignment` as you suggested.

Along with this change, I had to desert the variable `half_buf_size` because with the added `alignemnt`, `half_buf_size` is not 1/2 of `buf_size`. I replaced it with `tx_buf_size` and `rx_buf_size`.

fabtests/common/shared.c
```diff
@@ -526,36 +540,48 @@ int ft_alloc_msgs(void)
...
 		if (ret) {
 			FT_PRINTERR("posix_memalign", ret);
 			return ret;
 		}
 	} else {
-		ret = ft_hmem_alloc(opts.iface, opts.device, (void **) &buf, buf_size);
+		/* allocate extra "alignment" bytes, to handle the case
+		 * "buf" returned by ft_hmem_alloc() is not aligned.
+		 */
+		buf_size += alignment;
```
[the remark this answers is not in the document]
wzamazon: I think it is still needed because the address returned by `ft_hmem_alloc` is not guaranteed to be aligned.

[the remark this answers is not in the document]
sunkuamzn: Yes, you're right. It's still needed.

sunkuamzn (approved):

fabtests/common/shared.c
```diff
@@ -504,6 +505,19 @@ void ft_free_host_tx_buf(void)
...
+	return ((size % alignment) == 0) ?
+		size : ((size / alignment) + 1) * alignment;
+}
+
+void *ft_get_aligned_addr(void *ptr, size_t alignment)
+{
+	/* alignment must be power of 2, hence the assertion */
+	assert((alignment & (alignment - 1)) == 0);
+	return (void *)(((uintptr_t)ptr + alignment - 1) & ~((uintptr_t)alignment - 1));
+}
```
shefty, line 519: The above functions can be static, but I'd actually move them into the header file as static inline, to save the trouble of doing that further down the road.

shefty (approved):
Minor comment.  Changes look good and are okay to merge.

wzamazon:
The continuous-integration/appveyor/pr  error need to be looked into.

shefty:
Appveyor failure persists even after re-running.  I don't see what's off by looking at the changes.  We're overlooking something.

wzamazon:
I am spinning up a windows instance to try to reproduce.

wzamazon:
>Appveyor failure persists even after re-running. I don't see what's off by looking at the changes. We're overlooking something.

I accdientally remove the `opts.options & FT_OPT_ALIGN` condition, so all  allocation of host buf was through `posix_memalign`, but `buf` was still freed using regular `free`, which cause the issue.

Added back the condition in new revision.

sunkuamzn:
> I accdientally remove the `opts.options & FT_OPT_ALIGN` condition, so all allocation of host buf was through `posix_memalign`, but `buf` was still freed using regular `free`, which cause the issue.

I thought that was intentional because previously `alignment` was 1 (not a power of 2) but now it's 64 (which is power of 2). Can you explain the issue a bit more? According to the man page for `posix_memalign`

> Memory that is allocated via posix_memalign() can be used as an argument in subsequent calls to realloc(3), reallocf(3), and free(3).

sunkuamzn:
The AWS CI failures appear to be real

wzamazon:
Now sockets provider failed the `fi_rdm_rma_event` test. :)

wzamazon:
> Memory that is allocated via posix_memalign() can be used as an argument in subsequent calls to realloc(3), reallocf(3), and free(3).

It looks like windows compiler does not follow this rule. You have to

shefty:
@sunkuamzn - The failures were on windows, where posix_memalign does not exist.

The changes are exposing a bug in fabtests.

libfabric has the correct abstractions for aligned malloc/free.  See ofi_memalign() and ofi_freealign().  Fabtests implements posix_memlign() on windows (calls _aligned_malloc), but there's no corresponding 'aligned free' abstraction for fabtests to call.  Instead fabtests assumes that free() is suitable for all platforms, which isn't the case.

wzamazon:
> ibfabric has the correct abstractions for aligned malloc/free. See ofi_memalign() and ofi_freealign(). Fabtests implements posix_memlign() on windows (calls _aligned_malloc), but there's no corresponding 'aligned free' abstraction for fabtests to call. Instead fabtests assumes that free() is suitable for all platforms, which isn't the case.

I see. I guess it is better handled in a separate patch.

Issue #7827, opened 2022-06-11
fi_info -d/--domain flag not filtering


sdvormwa:
**Describe the bug**
Starting with v1.10.0, the -d/--domain flag to the fi_info utility no longer filters output by the domain.

**To Reproduce**
Download release tarball.  Build with default options.  Invoke fi_info with the -d flag.  Observe non-matching entries.

**Expected behavior**
Only entries matching the passed domain should be returned.

**Output**
$ libfabric/1.15.1/bin/fi_info -d lo | grep domain | sort -u
    domain: br-cc55c03d48df
    domain: docker0
    domain: eth0
    domain: lo
    domain: shm
$ libfabric/1.9.1/bin/fi_info -d lo | grep domain | sort -u
    domain: lo
$ libfabric/1.10.0/bin/fi_info -d lo | grep domain | sort -u
    domain: br-cc55c03d48df
    domain: docker0
    domain: eth0
    domain: lo
    domain: shm

**Environment:**
Linux.  Does not appear to be provider specific.

PR #8991, opened 2023-06-02
prov/tcp: Filter loopback device if iface is specified

### LINKED ISSUE #7706: prov/tcp: Setting FI_TCP_IFACE does not filter loopback device
```
FI_TCP_IFACE=eno1 fi_info -p tcp
provider: tcp;ofi_rxm
    fabric: 10.23.219.0/25
    domain: eno1
    version: 115.0
    type: FI_EP_RDM
    protocol: FI_PROTO_RXM
provider: tcp;ofi_rxm
    fabric: fe80::/64
    domain: eno1
    version: 115.0
    type: FI_EP_RDM
    protocol: FI_PROTO_RXM
provider: tcp;ofi_rxm
    fabric: 127.0.0.1/32
    domain: lo
    version: 115.0
    type: FI_EP_RDM
    protocol: FI_PROTO_RXM
```
The loopback device should have been filtered.  Problem is in ofi_get_list_of_addr(), which always tacks on the loopback devices at the end of the list.

shefty:
Fixes #7706

If the user specified a specific interface, the loopback device should be filtered out (unless the loopback device was the selected device).

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -2155,7 +2155,9 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 
 insert_lo:
 	/* Always add loopback address at the end */
-	ofi_insert_loopback_addr(prov, addr_list);
+	if (!iface || !strncmp(iface, "lo", strlen(iface) + 1) ||
+	    !strncmp(iface, "loopback", strlen(iface) + 1))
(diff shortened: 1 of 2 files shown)


shefty:
Added filtering for domain and fabric names.

Fixes #7827

PR #8997, opened 2023-06-07
contrib/intel/jenkins: Migrate from Jenkins to Cloudbees Jenkins

zachdworkin:
This set of changes touches all of the code in this folder.
Most Python classes got updated to no longer rely on old bash scripts
The jenkins_config file has become a cloudbees_config with new contents
The Jenkinsfile has been completely overhauled to enable slurm support on the main cluster. This support also contains generified functions that should make it easy to move DAOS and DSA nodes to use the slurm solution.
There are quality of life changes made to the Jenkinsfile to help with upgrading python versions and middleware versions.
DSA is getting re-enabled since the node is back online. It will be used in cloudbees and not the current controller to make the migration smoother
Oneccl testing has changed to use the oneccl team's test suite since some tests got deprecated in their release package


shefty:
Do not merge until Monday June 12

aingerson (approved):

PR #9001, opened 2023-06-08
fabtests/efa: test_flood_peer.py increase timeout

lrbison:
Increase timeout of new test_flood_peer.py test to 300 seconds to avoid any question if timeout is due to a hard hang or a slow run.


darrylabbate (approved):

lrbison:
After merged, we need to backport this to 1.18.x

shijin-aws (approved):

shijin-aws:
Need to backport

shijin-aws:
This script is only used by AWS CI. Ignoring Intel CI failure

PR #9002, opened 2023-06-08
prov/psm: Remove psm as support was EOL several years ago

acgoldma:


include/rdma/fabric.h
```diff
@@ -205,7 +205,6 @@ enum {
 	FI_SOCKADDR_IN,		/* struct sockaddr_in */
 	FI_SOCKADDR_IN6,	/* struct sockaddr_in6 */
 	FI_SOCKADDR_IB,		/* struct sockaddr_ib */
-	FI_ADDR_PSMX,		/* uint64_t */
```
shefty, line 208: This needs to remain.  Removing it breaks binary compatibility.  Just add similar comment as you used below.

shefty (changes requested):
Cannot remove defines from public header file

include/rdma/fabric.h
acgoldma: will do.

PR #9004, opened 2023-06-08
fabtests/component: Rename executables to prefix with xe_ or fi_xe_

zachdworkin:
Adding component tests to the gitignore by renaming them to have xe_ or fi_xe_ prefix to follow previous convention.


j-xiong (approved):
Name changes are fine for me. Note that the names of the executables also appear in the comments at the beginning of the source code. You can change them or leave them as they are. Just want to let you know.

zachdworkin:
@j-xiong updated the executable names in the comments

fabtests/component/dmabuf-rdma/memcopy-xe.c
```diff
@@ -34,69 +34,69 @@
  *
  *	Copy using memcpy() between buffers allocated with malloc():
  *
- *	    ./memcopy-xe -c memcpy M M
+ *	    ./memcopy_xe -c memcpy M M
```
j-xiong, line 37: should be xe_memcopy

zachdworkin: fixed, thanks!

j-xiong:
The commit message is missing some spaces and needs to be broken down to shorter lines.

PR #9025, opened 2023-06-12
contrib/intel/jenkins: Migrate DAOS tests to cloudbees

zachdworkin:
Re-enable daos tests with cloudbees solution.
Update daos cart tests function in tests.py with cloudbees terms and remove an unused function.


nikhilnanal:
looks good!

PR #9026, opened 2023-06-13
[v1.18.x] fabtests: backport cuda parallel test change

wenduwan:
Need to backport another dependent change for parallel cuda tests.


shijin-aws (approved):

wzamazon:
Intel CI does not use pytest script, so ignore the failure of Intel CI

PR #9031, opened 2023-06-13
man/fi_msg: require progress driving for -FI_EAGAIN

aingerson:
The man page currently specifies that driving progress when an operation returns -FI_EAGAIN is "strongly recommended" but should be defined as required for manual progress for applications to be fully compliant.


aingerson:
@shijin-aws

PR #9039, opened 2023-06-14
prov/sm2: Add CMA Capability to SM2

a-szegel:
* Re-organizes how we handle op_flags
* Adds the CMA Protocol

Great work @sunkuamzn on correctly root causing and fixing AWS CI issues.

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -176,11 +176,11 @@ static inline int sm2_match_tag(uint64_t tag, uint64_t ignore,
 static inline void sm2_generic_format(struct sm2_xfer_entry *xfer_entry,
 				      sm2_gid_t self_gid, uint32_t op,
 				      uint64_t tag, uint64_t cq_data,
-				      uint64_t op_flags, void *context)
+				      uint64_t *op_flags, void *context)
 {
 	xfer_entry->hdr.op = op;
 	/* We only care about lower 32 bits */
-	xfer_entry->hdr.op_flags = (uint32_t) op_flags;
+	xfer_entry->hdr.op_flags = (uint32_t) *op_flags;
 	xfer_entry->hdr.tag = tag;
 	xfer_entry->hdr.sender_gid = self_gid;
 	xfer_entry->hdr.cq_data = cq_data;
@@ -280,7 +280,7 @@ ssize_t sm2_verify_peer(struct sm2_ep *ep, fi_addr_t fi_addr, sm2_gid_t *gid);
 typedef ssize_t (*sm2_proto_func)(struct sm2_ep *ep,
 				  struct sm2_region *peer_smr,
 				  sm2_gid_t peer_gid, uint32_t op, uint64_t tag,
-				  uint64_t data, uint64_t op_flags,
+				  uint64_t data, uint64_t *op_flags,
 				  struct ofi_mr **mr, const struct iovec *iov,
 				  size_t iov_count, size_t total_len,
 				  void *context);
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -88,6 +88,7 @@ extern pthread_mutex_t sm2_ep_list_lock;
 enum {
 	sm2_proto_inject,
 	sm2_proto_return,
+	sm2_proto_cma,
 	sm2_proto_max,
 };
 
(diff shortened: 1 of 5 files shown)


prov/sm2/src/sm2.h
```diff
@@ -105,7 +106,7 @@ enum {
  * 		   NOTE: Only grabbing the bottom 32 bits
  * 	proto - sm2 operation
  * 	sender_gid - id of msg sender
- * 	user_data - the message
```
lrbison, line 108: Maybe more generically: the message or protocol-specific headers

prov/sm2/src/sm2_ep.c
```diff
@@ -273,6 +273,42 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
 }
 
+static void sm2_format_cma(struct sm2_xfer_entry *xfer_entry,
+			   const struct iovec *iov, size_t count)
+{
+	struct sm2_cma_data *cma_data =
+		(struct sm2_cma_data *) xfer_entry->user_data;
+
+	/* All CMA messages must be delivery complete to stop completion from
+	 * being written instantly */
```
lrbison, line 283: We discussed this a little bit today.  My opinion on this matter is that setting op_flags to be FI_DELIVERY_COMPLETE is different than saying we can't generate completion until we get the xfer_entry back.  I think we can know this simply based on the protocol.

You aren't setting op_flags here, but this comment is misleading.

prov/sm2/src/sm2_ep.c
```diff
@@ -273,6 +273,42 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+	xfer_entry->hdr.proto = sm2_proto_cma;
+	xfer_entry->hdr.size = ofi_total_iov_len(iov, count);
```
lrbison, line 285: More general than this PR: maybe we should rename hdr.size to hdr.data_byes.  To me currently it reads like either the header or the xfer entry size is this iov length.

prov/sm2/src/sm2_ep.c
```diff
@@ -273,6 +273,42 @@ static ssize_t sm2_do_inject(struct sm2_ep *ep, struct sm2_region *peer_smr,
...
+			  size_t total_len, void *context)
+{
+	ssize_t ret;
+	struct sm2_xfer_entry *xfer_entry;
+
+	ret = sm2_pop_xfer_entry(ep, &xfer_entry);
+	if (ret)
+		return ret;
+
+	*op_flags |= FI_DELIVERY_COMPLETE;
```
lrbison, line 303: See above comment about DELIVERY_COMPLETE

prov/sm2/src/sm2_ep.c
sunkuamzn: If we don't set FI_DELIVERY_COMPLETE, then we will have to check for the protocol while writing completion entries. I end up having to do that anyway for the CUDA host-device protocol but Seth was concerned about additional if statements in the critical path. Personally, I'm fine either way.

sunkuamzn (approved):

sunkuamzn:
AWS CI failures are CUDA related and legitimate. I'm checking now.

a-szegel:
So the AWS CI failures are from a bug in fabtests... HMEM CUDA message sizes are only supported for inject, but we are trying to send for any message size.

prov/sm2/src/sm2.h
```diff
@@ -176,11 +176,11 @@ static inline int sm2_match_tag(uint64_t tag, uint64_t ignore,
...
-				      uint64_t op_flags, void *context)
+				      uint64_t *op_flags, void *context)
 {
 	xfer_entry->hdr.op = op;
 	/* We only care about lower 32 bits */
-	xfer_entry->hdr.op_flags = (uint32_t) op_flags;
+	xfer_entry->hdr.op_flags = (uint32_t) *op_flags;
 	xfer_entry->hdr.tag = tag;
 	xfer_entry->hdr.sender_gid = self_gid;
 	xfer_entry->hdr.cq_data = cq_data;
```
lrbison, line 192: I don't understand this commit's purpose.  It says change flags to pointer to allow modification, but it doesn't modify them.

Also, why would generic format do such a modification?  It seems like if we allow this, it should be protocol-specific to decide to do so, not the job of generic_format.

Can we remove this commit?

a-szegel:
I was wrong about the fabtest causing the bug... the bug is delivery complete + HMEM for inject with these CMA changes.

a-szegel:
narrowing it down more... it is actually CMA Host to Device that is failing (must have forgot about this case on my vacation.... closing this PR.

sunkuamzn:
huh? if you set HMEM and non HMEM caps separately, host to device shouldn't use CMA and CI should pass? In any case, the HMEM PR is ready

PR #9040, opened 2023-06-14
Add RMA to SM2

lrbison:


prov/sm2/src/sm2.h
```diff
@@ -143,6 +160,20 @@ static inline struct smr_freestack *sm2_freestack(struct sm2_region *smr)
...
+					enum fi_hmem_iface *iface,
+					uint64_t *device)
+{
+	const struct ofi_mr *mr = desc;
+	if (false && mr) {
+		*iface = mr->iface;
+		*device = mr->device;
+	} else {
+		*iface = FI_HMEM_SYSTEM;
+		*device = 0;
```
lrbison, line 173: Fix me

prov/sm2/src/sm2_attr.c
```diff
@@ -104,7 +108,7 @@ struct fi_domain_attr sm2_domain_attr = {
 	.data_progress = FI_PROGRESS_MANUAL,
 	.resource_mgmt = FI_RM_ENABLED,
 	.av_type = FI_AV_UNSPEC,
-	.mr_mode = FI_MR_BASIC | FI_MR_SCALABLE,
+	.mr_mode = OFI_MR_BASIC_MAP | FI_MR_LOCAL | FI_MR_BASIC,
```
lrbison, line 111: Requesting feedback here to make sure these flags are right

prov/sm2/src/sm2_progress.c
```diff
@@ -165,11 +167,257 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+	cmd_rma = (struct sm2_cmd_rma_msg *) xfer_entry->user_data;
+
+	ofi_consume_rma_iov(cmd_rma->rma_iov, &rma_count,
+			    cmd_rma->request_offset);
+	iov = cmd_rma->rma_iov;
+	payload_used = 0;
+	while (rma_count) {
+		// mr = (struct sm2_mr*)
+		// ofi_mr_map_get(&ep->base_ep.util_ep.domain->mr_map,
+		// (rta_hdr->rma_iov + i)->key);
```
lrbison, line 201: Need to get HMEM working

prov/sm2/src/sm2.h
```diff
@@ -356,4 +356,14 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
...
 }
 
+static void inline sm2_freestack_push(struct sm2_ep *ep,
+				      struct sm2_xfer_entry *xfer_entry)
+{
+	struct sm2_av *av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_mmap *map = &av->mmap;
+	struct sm2_region *region = sm2_mmap_ep_region(map, ep->gid);
+	smr_freestack_push(sm2_freestack(region), xfer_entry);
```
shijin-aws, line 426: Is `smr_freestack_push` in `ofi*.h`? The naming looks for shm only

prov/sm2/src/sm2_attr.c
shijin-aws: It seems the current mode is copied from shm, anything wrong with it? May need to consult @aingerson

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,499 @@
+/*
+ * Copyright (c) 2013-2021 Intel Corporation. All rights reserved
```
shijin-aws, line 2: Is it because we are copying it from shm?

shijin-aws (commented):
I reviewed with my limited knowledge. Didn't dive deep into the protocols

prov/sm2/src/sm2.h
lrbison: yes, it is in ofi_mem.h.  The naming is... unfortunate to me.

prov/sm2/src/sm2_rma.c
lrbison: Yes.  Originally this was smr_rma.c.  Much of it has changed, but not all of it.

prov/sm2/src/sm2_attr.c
lrbison: I switched them back and it sill works.  I don't remember why I changed this in the first place 🤷 .  Removing this change.

prov/sm2/src/sm2.h
```diff
@@ -335,14 +335,24 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
...
 	}
-
 	*xfer_entry = smr_freestack_pop(sm2_freestack(self_region));
+#if ENABLE_DEBUG
+	{
+		assert(sm2_freestack(self_region)->free >= 0);
+		assert((char *) *xfer_entry >=
+		       (char *) sm2_freestack(self_region));
+		assert((char *) *xfer_entry <
+		       (char *) sm2_mmap_ep_region(map, ep->gid + 1));
```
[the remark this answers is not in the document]
shijin-aws: So, `sm2_mmap_ep_region(map, ep->gid + 1)` is the beginning of the region for the next ep (if there is)? Does that mean the freestack is the only part of each region. This check does not look intuitive to me.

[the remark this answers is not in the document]
lrbison: > So, `sm2_mmap_ep_region(map, ep->gid + 1)` is the beginning of the region for the next ep (if there is)?

Correct.  As I said above this is a crude check.  It's not a tight bound on the freestack but should be enough to find runaway problems.

I think it might actually be the last section of the region, but I'm not 100% sure.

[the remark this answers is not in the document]
lrbison: Also, if you prefer I can just remove these asserts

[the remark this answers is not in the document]
shijin-aws: No it doesn't hurt to check more things in DEBUG build, it would be good if you can add a comment here to explain

lrbison:
I've decided I want to change some names, also I find that we hang for 0-sized RMA.  Let me fix that.

prov/sm2/src/sm2.h
a-szegel: I have a backlog item in my notes to get us off of the SMR freestack... but yes, it should be renamed too.

a-szegel (changes requested):
first round of comments... didn't get to the main part yet.

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,509 @@
...
+	sm2_fill_rma_ctx(ep, msg, op, op_flags, peer_gid, rma_ctx);
+
+	if (op == ofi_op_read_req && rma_ctx->bytes_total == 0) {
+		uint64_t comp_flags = ofi_tx_cq_flags(ofi_op_read_req);
+		ret = sm2_complete_tx(ep, (void *) msg->context,
+				      ofi_op_read_req,
+				      comp_flags | FI_COMPLETION);
+		ret = FI_SUCCESS;
+		goto out;
+	}
```
a-szegel, line 235: This 0-byte read looks weird (comment would be nice), also it looks like it leaks a ctx object.

a-szegel:
You need another commit where you enable the RMA fabtests (see Atomic PR for inspiration).

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,509 @@
...
+		bytes_to_send =
+			MIN(msg->msg_iov[0].iov_len, msg->rma_iov[0].len);
+		bytes_to_send =
+			MIN(bytes_to_send, SM2_RMA_INJECT_SIZE - payload_used);
+		if (bytes_to_send == 0)
+			break;
+
+		src = (uint8_t *) msg->msg_iov[0].iov_base;
+		sm2_get_iface_device((void **) msg->desc, &iface, &device);
+		cmd_rma->rma_iov[jv].addr = msg->rma_iov[0].addr;
```
a-szegel, line 172: if we are always going to hard code rma_iov[0] to 0, should it even be an array?

The answer is that I think we need to send the full rma_iov so the receiving side knows where to put the data.

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,509 @@
...
+		payload_used += bytes_to_send;
+		ctx->bytes_sent += bytes_to_send;
+	}
+
+	if (ctx->bytes_sent == ctx->bytes_total) {
+		cmd_rma->sar_hdr.proto_flags |= FI_SM2_SAR_LAST_MESSAGE_FLAG;
+	}
+	ctx->msgs_in_flight++;
+
+	return FI_SUCCESS;
```
a-szegel, line 196: for 0 byte operations, you are sending 1 message which I am guessing is for you to post a completion on the remote side. Do we want to do this?

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,509 @@
...
+				ret = FI_SUCCESS;
+			goto out;
+		}
+		ret = sm2_rma_cmd_fill_sar_xfer(xfer_entry, rma_ctx);
+		if (!ret)
+			sm2_fifo_write(ep, peer_gid, xfer_entry);
+		else {
+			sm2_rma_handle_local_error(ep, xfer_entry, rma_ctx,
+						   ret);
+			break;
```
a-szegel, line 254: It looks possible for you to have an error completion, and a transmit complete for messages (0 byte or messages that fail on last send).

a-szegel (changes requested):
still not done but here are more comments.

prov/sm2/src/sm2.h
```diff
@@ -88,6 +94,7 @@ extern pthread_mutex_t sm2_ep_list_lock;
 enum {
 	sm2_proto_inject,
 	sm2_proto_return,
+	sm2_proto_sar,
```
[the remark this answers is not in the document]
lrbison: I expect it to be shared with message-based SAR, so I'm trying to leave it generic for now.

prov/sm2/src/sm2.h
```diff
@@ -129,6 +136,28 @@ struct sm2_ep_name {
 	struct dlist_entry entry;
 };
 
+struct sm2_cmd_sar_hdr {
+	union {
+		uint32_t proto_flags;
+		size_t reserved;
+	};
+	void *context;
```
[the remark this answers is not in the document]
lrbison: The generic header has the context that the user wants to see in completion.  This context is the state that the SAR protocol needs.  perhaps it could be called "sar_context" or "protocol_context"

prov/sm2/src/sm2.h
```diff
@@ -261,14 +261,27 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
...
 
 	if (smr_freestack_isempty(sm2_freestack(self_region))) {
 		sm2_progress_recv(ep);
+		self_region = sm2_mmap_ep_region(map, ep->gid);
 		if (smr_freestack_isempty(sm2_freestack(self_region)))
 			return -FI_EAGAIN;
 	}
-
 	*xfer_entry = smr_freestack_pop(sm2_freestack(self_region));
+#if ENABLE_DEBUG
```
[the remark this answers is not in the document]
lrbison: ah, an older version had a variable and filled it prior to the asserts.  I can remove the block now.

prov/sm2/src/sm2_attr.c
```diff
@@ -33,17 +33,22 @@
 
 #include "sm2.h"
 
-#define SM2_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED)
-#define SM2_RX_CAPS                                                   \
-	(FI_SOURCE | OFI_RX_MSG_CAPS | FI_TAGGED | FI_DIRECTED_RECV | \
-	 FI_MULTI_RECV)
+#define SM2_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED | OFI_TX_RMA_CAPS)
+#define SM2_RX_CAPS                                                  \
```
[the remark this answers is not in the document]
lrbison: Added to RX capabilities.

prov/sm2/src/sm2_attr.c
```diff
@@ -154,7 +159,8 @@ struct fi_info sm2_hmem_info = {
 };
 
 struct fi_info sm2_info = {
-	.caps = SM2_TX_CAPS | SM2_RX_CAPS | FI_MULTI_RECV | FI_LOCAL_COMM,
+	.caps = SM2_TX_CAPS | SM2_RX_CAPS | FI_MULTI_RECV | FI_LOCAL_COMM |
+		SM2_RMA_CAPS,
```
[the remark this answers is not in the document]
lrbison: done

prov/sm2/src/sm2_attr.c
```diff
@@ -33,17 +33,22 @@
...
+	 FI_DIRECTED_RECV | FI_MULTI_RECV)
+
 #define SM2_HMEM_TX_CAPS (SM2_TX_CAPS | FI_HMEM)
 #define SM2_HMEM_RX_CAPS (SM2_RX_CAPS | FI_HMEM)
 #define SM2_TX_OP_FLAGS                                              \
 	(FI_COMPLETION | FI_INJECT_COMPLETE | FI_TRANSMIT_COMPLETE | \
 	 FI_DELIVERY_COMPLETE)
 #define SM2_RX_OP_FLAGS (FI_COMPLETION | FI_MULTI_RECV)
 
+#define SM2_RMA_CAPS                                                      \
```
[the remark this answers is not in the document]
lrbison: done

prov/sm2/src/sm2_ep.c
```diff
@@ -225,7 +227,8 @@ static void cleanup_shm_resources(struct sm2_ep *ep)
 	/* Return all free queue entries in queue without processing them */
 return_incoming:
 	while (NULL != (xfer_entry = sm2_fifo_read(ep))) {
-		if (xfer_entry->hdr.proto == sm2_proto_return) {
+		if (xfer_entry->hdr.proto == sm2_proto_return ||
```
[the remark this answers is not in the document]
lrbison: Yes.  Perhaps we should rename it from "sender_gid" to "owner_gid" as well.

prov/sm2/src/sm2_ep.c
```diff
@@ -564,13 +570,20 @@ int sm2_endpoint(struct fid_domain *domain, struct fi_info *info,
 	if (ret || ofi_bufpool_grow(ep->xfer_ctx_pool)) {
 		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
 			"Unable to create xfer_entry ctx pool\n");
+	}
+	ret = ofi_bufpool_create(&ep->sar_ctx_pool, sizeof(struct sm2_sar_ctx),
+				 16, 0, info->rx_attr->size,
+				 OFI_BUFPOOL_NO_TRACK);
```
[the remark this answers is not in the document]
lrbison: Why?  What fixed size would you prefer?

lrbison (commented):
Thanks for the feedback Seth.  I've addressed your changes so far in my local copy.  I'll save the CI server a few cycles and wait to push until you've finished reviewing.

prov/sm2/src/sm2.h
```diff
@@ -261,14 +261,27 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
 		container_of(ep->util_ep.av, struct sm2_av, util_av);
 	struct sm2_mmap *map = &av->mmap;
 	struct sm2_region *self_region = sm2_mmap_ep_region(map, ep->gid);
+	*xfer_entry = NULL;
```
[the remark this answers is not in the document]
lrbison: done

prov/sm2/src/sm2.h
```diff
@@ -78,6 +78,12 @@
 #define SM2_IOV_LIMIT		4
 #define SM2_INJECT_SIZE		(SM2_XFER_ENTRY_SIZE - sizeof(struct sm2_xfer_hdr))
 
+#define FI_SM2_SAR_FIRST_MESSAGE_FLAG (1 << 0)
```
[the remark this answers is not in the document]
lrbison: removed

prov/sm2/src/sm2_rma.c
lrbison: I have to remove it now that we claim RMA_EVENT capability.

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,509 @@
...
+	}
+
+	while (rma_ctx->msgs_in_flight < 16 &&
+	       rma_ctx->bytes_sent != rma_ctx->bytes_total) {
+		ret = sm2_pop_xfer_entry(ep, &xfer_entry);
+		if (ret) {
+			if (rma_ctx->msgs_in_flight == 0) {
+				sm2_free_rma_ctx(rma_ctx);
+				ret = -FI_EAGAIN;
+			} else
```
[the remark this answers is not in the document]
lrbison: Agree, just trying to follow the rules but I was informed of this caveat last week!

prov/sm2/src/sm2_rma.c
```diff
@@ -0,0 +1,509 @@
...
+		ctx->msgs_in_flight++;
+
+		return FI_SUCCESS;
+	}
+
+	for (jv = 0; jv < SM2_IOV_LIMIT; jv++) {
+		if (ctx->bytes_sent == ctx->bytes_total)
+			break;
+		bytes_to_send =
+			MIN(msg->msg_iov[0].iov_len, msg->rma_iov[0].len);
```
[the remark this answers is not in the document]
lrbison: I'll rework this.  I think you are right that it's not quite correct.  In cases where there are non-overlapping remote and local iovs we may have empty space left in our packets.

The zero-indexing is intentional though, because we are calling `ofi_consume_*` which will eat some of the iov and leave `msg_iov[0]` as the next data to send.

prov/sm2/src/sm2_rma.c
lrbison: hinted above.  rma_iov[0] is always the right address for the remote to write into for this particular packet because we keep consuming the iov list which might fill the whole array.  Because we fill this rma list each time, the remote needs no context.

The alternative I think is to send the original, full RMA iov, then the remote needs to know which offset this packet corresponds to, does a "consume" of that many bytes, and uses the remaining values in the iov array to perform the copy of however many bytes are in the packet.  This also requires a packet size I suppose, since the RMA iov currently provides that for me.

prov/sm2/src/sm2_rma.c
lrbison: yes, we must send a message in order for remote to post completion.  For 0-byte reads I avoided it, but for writedata we couldn't.  Now with RMA_EVENT capability, we must always send the message for 0-sized operations.  In fact, I had to re-order the loop to `do {...} while()` so that we send even for zero-sized operations.

prov/sm2/src/sm2_progress.c
```diff
@@ -202,6 +203,278 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+		uint64_t comp_flags = ofi_tx_cq_flags(xfer_entry->hdr.op);
+		ret = sm2_complete_tx(ep, (void *) xfer_entry->hdr.context,
+				      xfer_entry->hdr.op,
+				      comp_flags | FI_COMPLETION);
+		sm2_free_rma_ctx(ctx);
+		sm2_freestack_push(ep, xfer_entry);
+		return FI_SUCCESS;
+	}
+
+	if (ctx->bytes_requested == ctx->bytes_total) {
```
a-szegel, line 332: When do we get into this case... are we leaking a ctx?

prov/sm2/src/sm2_progress.c
```diff
@@ -202,6 +203,278 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+
+static int sm2_progress_rma_write_origin(struct sm2_ep *ep,
+					 struct sm2_xfer_entry *xfer_entry)
+{
+	struct sm2_cmd_sar_rma_msg *cmd_rma = (void *) xfer_entry->user_data;
+	struct sm2_sar_ctx *ctx = (void *) cmd_rma->sar_hdr.context;
+	bool tx_complete, rx_complete;
+	int ret;
+
+	if (!ctx) {
```
a-szegel, line 354: It looks to me like rma_inject useses generic_rma which creates a context for every message... so this branch is dead code which will never be hit.

prov/sm2/src/sm2_progress.c
```diff
@@ -41,6 +41,7 @@
 #include "ofi_mr.h"
 #include "sm2.h"
 #include "sm2_fifo.h"
+#include "sm2_rma.h"
```
a-szegel, line 44: I think we should put these functions in sm2_rma.c ... what do you think?

prov/sm2/src/sm2_rma.h
```diff
@@ -0,0 +1,52 @@
...
+	*rma_ctx = ctx;
+	return FI_SUCCESS;
+}
+
+static inline void sm2_free_rma_ctx(struct sm2_sar_ctx *rma_ctx)
+{
+	ofi_buf_free(rma_ctx);
+}
+
+void sm2_fill_rma_ctx(struct sm2_ep *ep, const struct fi_msg_rma *msg,
```
a-szegel, line 23: nit: i sort of hate declarations between definitions, can we have all the declarations above or below the definitions?

a-szegel (changes requested):
Finished my review, very well done overall.  Thank you Luke!

prov/sm2/src/sm2_ep.c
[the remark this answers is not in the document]
a-szegel: `SM2_NUM_XFER_ENTRY_PER_PEER` which I believe will one day become tx_size

prov/sm2/src/sm2_progress.c
lrbison: No, this is similar to rx_complete/tx_complete but for read.  Essentially we don't need to send back the packet requesting more data, because we already have sent enough requests, we just need to return this entry to freestack and wait for the last request to return to us.

prov/sm2/src/sm2_progress.c
```diff
@@ -202,6 +203,278 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+		if (ret) {
+			cmd_rma->sar_hdr.proto_flags |= FI_SM2_SAR_ERROR_FLAG;
+			break;
+		}
+		ret = ofi_copy_to_hmem(mr->iface, mr->device,
+				       (void *) hmem_addr, src, iov->len);
+		src += iov->len;
+		assert((long) (src - (uint8_t *) &cmd_rma->user_data) <=
+		       SM2_RMA_INJECT_SIZE);
+		if (ret) {
```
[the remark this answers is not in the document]
lrbison: Addressed.  Sequence for this case is:
1. Remote (the ep in this progress function) emits error to CQ
2. Origin emits error to CQ
3. Origin cleans up context.

As long as remote has no context and only 1 message is in flight at a time, this is sufficient.

If we want more than 1 message in flight at a time, we need some state at remote to prevent a subsequent message from  succeeding and accidentally submitting 2 CQs.

prov/sm2/src/sm2_progress.c
```diff
@@ -202,6 +203,278 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+			ep, (void *) xfer_entry->hdr.context,
+			xfer_entry->hdr.op, comp_flags,
+			total_bytes_for_completion, buf_for_completion,
+			xfer_entry->hdr.sender_gid, 0, xfer_entry->hdr.cq_data);
+		if (ret) {
+			cmd_rma->sar_hdr.proto_flags |= FI_SM2_SAR_ERROR_FLAG;
+		}
+	}
+
+	if (cmd_rma->sar_hdr.proto_flags & FI_SM2_SAR_ERROR_FLAG)
```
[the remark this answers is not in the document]
lrbison: fixed using goto out rather than break in the loop, and out skips the path that might generate successful completion.

prov/sm2/src/sm2_progress.c
```diff
@@ -202,6 +203,278 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
 	return ret < 0 ? ret : 0;
 }
 
+static inline void sm2_rma_write_back(struct sm2_ep *ep,
```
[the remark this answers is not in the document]
lrbison: moved to sm2_fifo.

prov/sm2/src/sm2_progress.c
```diff
@@ -202,6 +203,278 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+	ctx->status_flags |=
+		cmd_rma->sar_hdr.proto_flags & FI_SM2_SAR_ERROR_FLAG;
+	if (ctx->status_flags & FI_SM2_SAR_ERROR_FLAG) {
+		sm2_rma_handle_remote_error(ep, xfer_entry, ctx);
+		return FI_SUCCESS;
+	}
+
+	while (iov_count) {
+		bytes_to_copy =
+			MIN(iov[0].iov_len, SM2_RMA_INJECT_SIZE - bytes_used);
```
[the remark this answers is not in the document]
lrbison: consume does the work for us.

prov/sm2/src/sm2_progress.c
lrbison: Thank you, old code :-)

prov/sm2/src/sm2_progress.c
lrbison: I wasn't sure which functions you meant here.  You mean alloc and free?  We still need free in progress as well.

lrbison:
I think I've addressed most if not all comments.  I did add ubertests, but there are some tests that it complains about ENODATA.  We'll see what CI says about that.  They did help my find one additional problem: I had the wrong level of pointer indirection for `desc` which made multiple iov RMA operations not work.

One additional problem I discovered is that when I added FI_RMA_EVENT capabilities, I have all sorts of problems with the fi_rma_bw fabtest.  SHM seems to have this capability and passes the test fine, so there must be something wrong in my implementation.  It seemed that the completions I was generating for writes on remote was being interpreted as message completions and so it would not sync correctly.  Oddly, I'd even get verification errors on the message ft_sync() itself was sending (probably because the message never arrived but the CQ counter incremented because of RMA).

For now I've removed FI_RMA_EVENT capabilities and this solves the immediate problem.

prov/sm2/src/sm2_attr.c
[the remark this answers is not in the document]
lrbison: Removed it again.  There is some completion issue during testing.

prov/sm2/src/sm2_attr.c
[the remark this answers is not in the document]
lrbison: Oops, this snuck back in during modifications.  The define is present but the value is unused.

prov/sm2/src/sm2_progress.c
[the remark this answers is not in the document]
lrbison: latest update does this now.

prov/sm2/src/sm2_rma.h
lrbison: Resolved when I moved some functions into sm2_fifo.h

vidsouza:
bot:aws:retest

a-szegel:
bot:aws:retest

a-szegel:
closing to https://github.com/ofiwg/libfabric/pull/9089

prov/sm2/src/sm2_ep.c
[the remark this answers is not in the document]
lrbison: we can address the `s/sender_gid/owner_gid` this in the future

prov/sm2/src/sm2_rma.c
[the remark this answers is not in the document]
lrbison: resolved.

a-szegel (approved):
Looks good!  Thank you :)

prov/sm2/src/sm2.h
```diff
@@ -289,6 +289,9 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
 		return -FI_EAGAIN;
 
 	*xfer_entry = smr_freestack_pop(sm2_freestack(ep->self_region));
+
+	/* ensure we haven't over-exhausted the freestack: */
+	assert(sm2_freestack(ep->self_region)->free >= 0);
```
aingerson, line 366: I think this assert would make more sense before you pop
`assert(sm2_freestack(ep->region)->free);`

prov/sm2/src/sm2_fifo.h
```diff
@@ -241,4 +241,30 @@ static inline void sm2_fifo_write_back(struct sm2_ep *ep,
...
 }
 
+static inline const char *sm2_print_ptr(struct sm2_ep *ep, void *ptr)
+{
+	static char outstr[80];
+	struct sm2_av *av;
+	struct sm2_mmap *map;
+	av = container_of(ep->util_ep.av, struct sm2_av, util_av);
+	map = &av->mmap;
+	struct sm2_coord_file_header *header = (void *) map->base;
```
aingerson, line 260: Fix variable declaration block - declare (inline initialization is ok), space, rest of code.
Also I think Seth added a helper function to simplify the map pointer get

prov/sm2/src/sm2_fifo.h
```diff
@@ -241,4 +241,30 @@ static inline void sm2_fifo_write_back(struct sm2_ep *ep,
...
+	if (map->base > (char *) ptr)
+		sprintf(outstr, "%p (not a mapped pointer, address before map)",
+			ptr);
+	else if (map->base + map->size < (char *) ptr)
+		sprintf(outstr, "%p (not a mapped pointer, address after map)",
+			ptr);
+	else {
```
aingerson, line 267: Fix bracket use

prov/sm2/src/sm2_fifo.h
```diff
@@ -241,4 +241,30 @@ static inline void sm2_fifo_write_back(struct sm2_ep *ep,
...
+		int64_t relptr = sm2_absptr_to_relptr(ptr, map);
+		int jregion = (relptr - header->ep_regions_offset) /
+			      header->ep_region_size;
+		int64_t region_off =
+			(int64_t) ptr - (int64_t) sm2_peer_region(ep, jregion);
+		sprintf(outstr, "%p (%ld, %ld bytes into region %d)", ptr,
+			relptr, region_off, jregion);
+	}
+	return outstr;
```
aingerson, line 276: Returned local variable isn't accessible after the function exits. See what the fi_tostr function does to get around it.

prov/sm2/src/sm2_fifo.h
```diff
@@ -241,4 +241,30 @@ static inline void sm2_fifo_write_back(struct sm2_ep *ep,
 	sm2_fifo_write(ep, xfer_entry->hdr.sender_gid, xfer_entry);
 }
 
+static inline const char *sm2_print_ptr(struct sm2_ep *ep, void *ptr)
```
aingerson, line 253: Please add more of a description in the commit explaining what exactly this does (who owns the ptr, what look up is it doing, etc)

prov/sm2/src/sm2.h
```diff
@@ -299,4 +299,10 @@ int sm2_query_atomic(struct fid_domain *domain, enum fi_datatype datatype,
 		     enum fi_op op, struct fi_atomic_attr *attr,
 		     uint64_t flags);
 
+inline static void sm2_freestack_push(struct sm2_ep *ep,
+				      struct sm2_xfer_entry *xfer_entry)
+{
+	smr_freestack_push(sm2_freestack(ep->self_region), xfer_entry);
```
aingerson, line 377: This seems like a rather silly function to have... it's pretty clear what it's doing and is only called twice. what is the purpose of this?

prov/sm2/src/sm2_rma.h
```diff
@@ -0,0 +1,34 @@
+#include "sm2.h"
```
aingerson, line 1: Missing copyright.
Also I don't think this separate file is really necessary? All of these can just be static inside rma.c/progress.c

prov/sm2/src/sm2.h
```diff
@@ -82,6 +82,15 @@
 #define SM2_ATOMIC_INJECT_SIZE	    (SM2_INJECT_SIZE - sizeof(struct sm2_atomic_hdr))
 #define SM2_ATOMIC_COMP_INJECT_SIZE (SM2_ATOMIC_INJECT_SIZE / 2)
 
+/* protocol flags */
+#define SM2_SAR_LAST_MESSAGE_FLAG (1 << 1)
+#define SM2_SAR_RETURN		  (1 << 2)
+#define SM2_SAR_ERROR_FLAG	  (1 << 3)
+#define SM2_SAR_STATUS_COMPLETED  (1 << 4)
+
```
aingerson, line 90: This patch is adding RMA but it looks like it's adding SAR stuff as well? Please separate

aingerson (commented):
I'm going to submit my initial review early (still going to finish, but it needs some larger adjustments so I'll let you get started on them).
The RMA patch looks like it's adding RMA as well as SAR so that needs to be 2 separate commits.
Also please get in the habit of always including information in the commit message. Some are ok empty (small change, one line, etc). But for the most part, more information is better. The RMA patch especially (and subsequent SAR patch as well) needs more information explaining the flow of the protocols. It makes it much easier for reviewing and for future developers to look at commits and understand what they did/are doing.

prov/sm2/src/sm2.h
lrbison: well, when I wrote it there were locks involved and ep->self_region were not present.  It was originally much longer to account for that.  Also although it's only used twice during this commit, it will get used a total of 8 times in later commits.

Finally, when scrolling through code with a bunch of `sm2_*` functions, I find it useful to have push be an `sm2_*` function.

prov/sm2/src/sm2.h
lrbison: We intend on introducing SAR for messages and for RMA.  Rather than have one massive PR I've started with implementing RMA using SAR, and intend to do messages with SAR next.  One of them has to come first.  I cannot have a commit introducing SAR without having some mechanism which uses it.

prov/sm2/src/sm2_rma.h
lrbison: ack on copyright thanks.

I actually would rather have all of these be in a sm2_sar.c/h file and have progress jump into various protocols, but I don't want to refactor current organization right now.

Would you prefer this file be named sm2_sar.h?

sunkuamzn (commented):
Just a couple of minor comments. Thanks Luke!

prov/sm2/src/sm2_attr.c
```diff
@@ -86,14 +89,15 @@ struct fi_rx_attr sm2_hmem_rx_attr = {
 	.iov_limit = SM2_IOV_LIMIT,
 };
 
+#define SMALLEST_INJECT MIN(SM2_RMA_INJECT_SIZE, SM2_ATOMIC_INJECT_SIZE)
```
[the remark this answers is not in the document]
lrbison: This define is really a line-length convenience.  It's only scope is in this .c file.  I can add SM2_ at the front.

prov/sm2/src/sm2_progress.c
```diff
@@ -330,6 +458,138 @@ static int sm2_progress_atomic(struct sm2_ep *ep,
...
+		assert((long) (src - (uint8_t *) &cmd_rma->user_data) <=
+		       SM2_RMA_INJECT_SIZE);
+	}
+
+	/* report completion if this is the last packet. */
+	if (xfer_entry->hdr.proto_flags & SM2_SAR_LAST_MESSAGE_FLAG) {
+		comp_flags = ofi_rx_cq_flags(xfer_entry->hdr.op);
+		comp_flags |= FI_REMOTE_CQ_DATA & xfer_entry->hdr.op_flags;
+		/* TODO: when sm2 supports FI_RMA_EVENT: */
+		/* comp_flags |= FI_COMPLETION & xfer_entry->hdr.op_flags; */
```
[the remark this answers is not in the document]
lrbison: I suppose I could remove both comments and let whoever implements sm2's FI_RMA_EVENT re-discover this, or I could leave these breadcrumbs.

I guess I'll remove.

prov/sm2/src/sm2_fifo.h
lrbison: The variable is static, so it should be allocated in the data segment.  I suppose we are uselessly pre-allocating it even though we don't expect it to be used (this is primarily for debug support convenience).  I'll make it a NULL pointer and allocate it if needed.

prov/sm2/src/sm2.h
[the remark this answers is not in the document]
lrbison: done

prov/sm2/src/sm2_ep.c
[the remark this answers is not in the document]
lrbison: done

prov/sm2/src/sm2.h
lrbison: I removed some other checks this commit originally had.  I think now this commit is not so useful.  I'll remove it.

prov/sm2/src/sm2_rma.h
aingerson: I'm ok with having a separate header but it really needs to be isolated to what it does (whether SAR or RMA). It looks like this does both which is why it's confusing.

prov/sm2/src/sm2_fifo.h
aingerson: That's fine but please put it in both places

prov/sm2/src/sm2.h
aingerson: I think it's very confusing to have it come in as one. SAR and RMA are two completely different things. SAR is a protocol. RMA is API functionality. SAR will be needed for messages and RMA (the flow should be totally the same with both, just the way you resolve the destination address is different) so doesn't rely on RMA. RMA should also work the same as the other protocols as well so you should be able to use RMA with inline/inject so it shouldn't rely on SAR. SAR and RMA are completely separate. I think avoiding massive PRs is great and I think that's what this is if you combine them. 
I think you need to either add SAR for regular messages first and then add RMA for all protocols or add RMA for the existing protocols and then add SAR for both. Either order is fine with me, but merging them into one is extremely confusing and hard to read and review.

prov/sm2/src/sm2_fifo.h
lrbison: on second thought that might make valgrind mad.  Let's just leave it as is.

prov/sm2/src/sm2.h
lrbison: > SAR is a protocol. RMA is API functionality.

100% Agree and I understand what you are saying.  However I am not a prescient developer.  I have a concept for a SAR protocol in my head, and I have a concept of how that protocol fits into implementing RMA and messaging functionalities.

It seemed like RMA capability was a good place to start because other developers were providing large message capabilities, and RMA was the sore-thumb of missing functionality.

I did my best to clearly delineate SAR that I expected to re-use in the future for messaging by including SAR in the name, however as I made progress my mental model of which functions would be re-used and how was adjusted, and I'm sure they'll continue to adjust when messages are implemented.  Please don't mistake SAR being used for many RMA operations in this commit to mean that I have somehow confused the two.

I do expect that the Messaging-using-SAR commit will probably further solidify and clarify where the boundaries fall, but given the pace of development, the number of revisions each PR gets, and the amount of parallel work which would be hurt by merge conflicts, I think the choice to implement RMA with SAR and then implement Messages with SAR is the best way forward.  These both need to happen because the SAR-protocol path provides large message when CMA is not enabled and this PR should provide all-size support for RMA, even if in the future we have a faster option.

> you should be able to use RMA with inline/inject so it shouldn't rely on SAR.
The SAR I've written gracefully degrades to something akin to inject, but I'm using the SAR header, so it remains "SAR".

prov/sm2/src/sm2_attr.c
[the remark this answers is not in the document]
lrbison: done

prov/sm2/src/sm2_fifo.h
lrbison: done

sunkuamzn (approved):

prov/sm2/src/sm2_fifo.h
lrbison: fixed

prov/sm2/src/sm2_rma.h
lrbison: Removed the rma.h header and renamed several of those functions to be sar specific rather than rma specific.  You can find them in sm2.h now.

prov/sm2/src/sm2.h
aingerson: Sorry, I wasn't suggesting you were confusing the two. I was saying that putting them together in the same patch is confusing to me as a reviewer because it's hard to tell if a change has to do with SAR or RMA specifically. 
I understand that simultaneous development causes headaches but I think it's for that reason to split it out into easier to handle chunks. Taking the easy way out of writing a lot of code and then just merging it to avoid conflicts so people can rebase I think is problematic for the project as a whole. If simultaneous development within your team is causing issues because you are relying on all the code to go into upstream in order for others to pick it up, I would suggest using someone's branch as a development playground instead (ie have everyone merge to Seth's AWS_sm2 branch and everyone can rebase off that) and then squash commits accordingly. There's no reason to rush things into upstream that aren't fully thought out.
If the SAR you have written gracefully degrades to inject, I would think that it would be easy to plug the RMA you've written into the existing inject protocol. That change, I think, would be simple and could be merged right away. My concern with them together is that SAR is a much more complicated beast since it requires multiple round trips and that requires more thorough review. I would rather remove the RMA specific parts of the code so that we can better view the SAR protocol to understand how it moves data around, whether it's a send or a read/write.

lrbison: I'll do my best to split RMA into a commit with inject only followed by SAR.

lrbison:
Closing this as OBE.

PR #9045, opened 2023-06-15
pytest/efa: unset strict mode for dgram pingpong test

wenduwan:
Increase timeout from 5s(default) to 30s


a-szegel (approved):

shijin-aws (changes requested):
dgram pingpong is not reliable. We shouldn't fail if we got a ENODATA in cq read. Increasing the timeout is not the right fix.

wenduwan:
I think @shijin-aws is correct. Please see update.

shijin-aws (approved):
thx

wenduwan:
Strictly EFA change. Ready to merge @shijin-aws

PR #9049, opened 2023-06-15
[Intel/CI]: enable fabtests for ucx 

nikhilnanal:


contrib/intel/jenkins/Jenkinsfile
```diff
@@ -359,6 +359,29 @@ pipeline {
...
+                           """${command}${build_item} \
+                           --ofi_build_mode=$mode""")
+                echo "Building Fabtests $mode"
+                build_item="fabtests"
+                run_python(PYTHON_VERSION,
+                           """${command}${build_item} \
+                           --ofi_build_mode=$mode""")
+              }
+              dir ("${env.WORKSPACE}/${SCRIPT_LOCATION}/") {
+                run_fabtests("ucx", "totodile", "2", "ucx")
```
zachdworkin, line 380: can we also use squirtle here or no?

contrib/intel/jenkins/build.py
```diff
@@ -119,11 +121,13 @@ def log_dir(install_path, release=False):
     parser.add_argument('--release', help="This job is likely testing a "\
                         "release and will be checked into a git tree.",
                         action='store_true')
+    parser.add_argument('--ucx', help="build with ucx", default=False, action='store_true')
```
zachdworkin, line 124: line length

zachdworkin (changes requested):
mostly looks good, can you also add the ucx provider to the summary.py file? I think its not getting picked up for the summarizer.

fabtests/test_configs/ucx/ucx.exclude
```diff
@@ -14,10 +11,36 @@ av_test
 ^fi_dgram
 -e msg
 -e dgram
-
+#
```
aingerson, line 14: Extra # here

fabtests/test_configs/ucx/ucx.exclude
```diff
@@ -14,10 +11,36 @@ av_test
...
 getinfo
 rdm g00n13s
```
aingerson, line 16: Since the negative test is already excluded here, we shouldn't need the -N argument in the CI run

aingerson (commented):
The commits feel out of order. The second one which adds a special ucx build is after the one that adds the testing and modifies things the first one changed. 
I also don't really understand the need for the second commit in general (the separate build)

aingerson:
@nikhilnanal I know this just got merged, but could you explain the need for the separate ucx build? I didn't understand why it was added or how it works.

shefty:
Isn't that just checking if ucx libraries are present and enabling the provider in the build if so?

aingerson:
It didn't look like that to me? It seemed like it was iterating through all the nodes and installing libfabric again on all the nodes with ucx? We already have provider lists for every cluster that specifies which providers to specifically enable on the config line:
```
    if (cluster == 'daos'):
        prov_list = common.daos_prov_list
    elif (cluster == 'dsa'):
        prov_list = common.dsa_prov_list
    else:
        prov_list = common.default_prov_list

    for prov in prov_list:
       config_cmd.append(f'--enable-{prov}={enable_prov_val}')
```

nikhilnanal:
> @nikhilnanal I know this just got merged, but could you explain the need for the separate ucx build? I didn't understand why it was added or how it works.

CVLs and mlx conflict on mofed(contains libuc*) installation. test binaries complain of missing libuc* dependencies on cvls and nodes which dont have mofed installed. To make that work on cvl we have a build without ucx enabled which does not  need those dependencies.

nikhilnanal:
> It didn't look like that to me? It seemed like it was iterating through all the nodes and installing libfabric again on all the nodes with ucx? We already have provider lists for every cluster that specifies which providers to specifically enable on the config line:
> 
> ```
>     if (cluster == 'daos'):
>         prov_list = common.daos_prov_list
>     elif (cluster == 'dsa'):
>         prov_list = common.dsa_prov_list
>     else:
>         prov_list = common.default_prov_list
> 
>     for prov in prov_list:
>        config_cmd.append(f'--enable-{prov}={enable_prov_val}')
> ```

it doesnt iterate through all the nodes. It  builds a separate libfabric with enable-ucx.  The Jenkins ucx nodes uses this build while other nodes use the the non ucx build.

Issue #9035, opened 2023-06-14
ofi_consume_iov: inconsistent final consume


lrbison:
In [iov.c: ofi_consume_iov_desc](https://github.com/ofiwg/libfabric/blob/main/src/iov.c#L63)

**Describe the bug**
The final consume of an iov has different behavior depending on if `*iov_count` is 1.
 - Typical case:
   - `*iov_count` is 1, the loop and memmove are skipped, and `*iov_count` remains 1, and that `iov_len` is now 0.
 - Alternate Case:
   - `*iov_count` is >1, but will be entirely consumed.  In this path the loop is executed, but the memove is a no-op because `i=*iov_count`.  `*iov_count` will be 0, but `iov[0].iov_len` can be non-zero.
   - Calling consume again ignores 0-size and modifies `iov_len` to be 0.

**Expected behavior**
Although in both cases consume_iov indicates fully-consumed iov with either `*iov_count = 0` or `iov_len = 0`, the fact that calling code must check both conditions is error-prone.  This function itself ignores `iov_count=0` case, suggesting that perhaps `iov_count=0` is not even a valid configuration..

What is the preferred behavior here?  Should I make this routine handle 0-sized iovs?  Should I ensure that if an iov is fully consumed that both "empty" conditions are set?  I suspect the iov_len=0 is the condition more commonly checked, or is the function OK as is and I should be careful to check both empty conditions?

shefty:
Reading the code, I agree it can be improved.  I see that the tcp provider has both checks:
```
ofi_consume_iov(rx_entry->iov, &rx_entry->iov_cnt, len);
if (!rx_entry->iov_cnt || !rx_entry->iov[0].iov_len) {
```

I think ofi_consume_iov() makes these assertions:
```
assert(to_consume && *iov_count);
assert(to_consume <= ofi_total_iov_len(iov, *iov_count));
```

To make the code consistent, the iov_count = 1 case needs a check similar to the if statement inside the for-loop, which determines whether we fully consume iov[0] or not.

PR #9050, opened 2023-06-16
src/iov.c: ofi_consume_iov consistency

lrbison:
Previously either iov_count would be 0, or iov_count=1 and iov[0].iov_len=0.  Now both iov_count=0 and iov[0].iov_len0.

Same for ofi_consume_rma_iov.

---

See https://github.com/ofiwg/libfabric/issues/9035

### DIFF
--- a/src/iov.c
+++ b/src/iov.c
@@ -63,21 +63,36 @@ uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, uint64_t io
 void ofi_consume_iov_desc(struct iovec *iov, void **desc,
 			  size_t *iov_count, size_t to_consume)
 {
-	size_t i;
+	struct iovec *cur_iov;
+	void **cur_desc;
 
-	if (*iov_count == 1)
+	assert(to_consume && *iov_count);
+	assert(to_consume <= ofi_total_iov_len(iov, *iov_count));
+
+	if (*iov_count == 1) {
+		if (iov[0].iov_len == to_consume) {
+			*iov_count = 0;
+			return;
+		}
 		goto out;
-
-	for (i = 0; i < *iov_count; i++) {
-		if (to_consume < iov[i].iov_len)
-			break;
-		to_consume -= iov[i].iov_len;
 	}
-	memmove(iov, &iov[i], sizeof(*iov) * (*iov_count - i));
-	if (desc)
-		memmove(desc, &desc[i],
-			sizeof(*desc) * (*iov_count - i));
-	*iov_count -= i;
+
+	cur_iov = iov;
+	cur_desc = desc;
+	while (*iov_count && to_consume >= cur_iov->iov_len) {
+		to_consume -= cur_iov->iov_len;
+		cur_iov++;
+		cur_desc++;
+		*iov_count -= 1;
+	}
+	if (*iov_count == 0)
+		return;
+
+	if (iov != cur_iov) {
+		memmove(iov, cur_iov, sizeof(*iov) * (*iov_count));
+		if (desc)
+			memmove(desc, cur_desc, sizeof(*desc) * (*iov_count));
+	}
 out:
 	iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
 	iov[0].iov_len -= to_consume;
@@ -91,19 +106,31 @@ void ofi_consume_iov(struct iovec *iov, size_t *iov_count, size_t to_consume)
 void ofi_consume_rma_iov(struct fi_rma_iov *rma_iov, size_t *rma_iov_count,
 			 size_t to_consume)
 {
-	size_t i;
+	struct fi_rma_iov *cur_iov;
 
-	if (*rma_iov_count == 1)
-		goto out;
+	assert(to_consume && *rma_iov_count);
+	assert(to_consume <= ofi_total_rma_iov_len(rma_iov, *rma_iov_count));
 
-	for (i = 0; i < *rma_iov_count; i++) {
-		if (to_consume < rma_iov[i].len)
-			break;
-		to_consume -= rma_iov[i].len;
+	if (*rma_iov_count == 1) {
+		 if (rma_iov[0].len == to_consume) {
+			*rma_iov_count = 0;
+			return;
+		 }
+		 goto out;
 	}
-	memmove(rma_iov, &rma_iov[i],
-		sizeof(*rma_iov) * (*rma_iov_count - i));
-	*rma_iov_count -= i;
+
+	cur_iov = rma_iov;
+	while (*rma_iov_count && to_consume >= cur_iov->len) {
+		to_consume -= cur_iov->len;
+		cur_iov++;
+		*rma_iov_count -= 1;
+	}
+	if (*rma_iov_count == 0)
+		return;
+
+	if (rma_iov != cur_iov) {
+		memmove(rma_iov, cur_iov, sizeof(*rma_iov) * (*rma_iov_count));
+	}
 out:
 	rma_iov[0].addr += to_consume;
 	rma_iov[0].len -= to_consume;


src/iov.c
```diff
@@ -79,8 +79,13 @@ void ofi_consume_iov_desc(struct iovec *iov, void **desc,
...
 out:
-	iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
-	iov[0].iov_len -= to_consume;
+	if (to_consume < iov[0].iov_len) {
+		iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
+		iov[0].iov_len -= to_consume;
+	} else if (*iov_count) {
+		*iov_count--;
+		iov[0].iov_len = 0;
+	}
```
wenduwan, line 88: What if `to_consume == 0 && *iov_count == 1`, does this branch still apply?

Same in ofi_consume_rma_iov

src/iov.c
```diff
@@ -79,8 +79,13 @@ void ofi_consume_iov_desc(struct iovec *iov, void **desc,
 			sizeof(*desc) * (*iov_count - i));
 	*iov_count -= i;
 out:
-	iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
-	iov[0].iov_len -= to_consume;
+	if (to_consume < iov[0].iov_len) {
+		iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
+		iov[0].iov_len -= to_consume;
+	} else if (*iov_count) {
```
shefty, line 85: This check implies that iov_count can be 0. But we've already assumed that it is 1 by dereferencing iov[0] above.

I don't think we need to handle iov_count = 0.  Just assert that it's >= 1.

We also don't need to set iov[0].iov_len =0.  If iov_count = 0 on return, then the user shouldn't access iov[0].

src/iov.c
lrbison: In the issue I filed Sean seems to think this function should assume:
```
assert(to_consume && *iov_count);
assert(to_consume <= ofi_total_iov_len(iov, *iov_count));
```
I think I've modified to handle the *iov_count=0 case, so now I think we can assume:
```
assert(to_consume);
assert(to_consume <= ofi_total_iov_len(iov, *iov_count));
```
But reading this I think there are two cases to consider: more than 1 iov, with the first one being zero and the rest non-zero, and 1 remaining iov with 0.

For the first case, the loop will actually consume the zero-sized iov, even if we are consuming 0 bytes.  so I think that's fine.
For the second case, iov_count-- will consume the last iov.

lrbison: See conversation below with Sean.  We are asserting to_consume != 0 now.

src/iov.c
lrbison: Done.  I also explicitly added the asserts.

wenduwan (approved):

src/iov.c
```diff
@@ -79,8 +82,11 @@ void ofi_consume_iov_desc(struct iovec *iov, void **desc,
 			sizeof(*desc) * (*iov_count - i));
 	*iov_count -= i;
 out:
-	iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
-	iov[0].iov_len -= to_consume;
+	if (to_consume < iov[0].iov_len) {
```
shefty, line 85: nit: if the if statement needs braces, then add them to the else statement.

Btw, the code is optimized for the case ofi_total_iov_len() > to_consume.  It works if it's equal, but ends up calling memmove(..., 0), iov_base + 0, and iov_len - 0.

At least with the tcp provider, this function is only called if iov_len > to_consume.  Otherwise, it's skipped.  I'm not sure about EFA or other use cases, such that we may want to add an optimization for it (at the expense of adding another branch to the current optimized case.

lrbison: Addressed the nit and added protection for memmove.

src/iov.c
```diff
@@ -73,14 +76,20 @@ void ofi_consume_iov_desc(struct iovec *iov, void **desc,
...
+	}
 	*iov_count -= i;
 out:
-	iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
-	iov[0].iov_len -= to_consume;
+	if (to_consume < iov[0].iov_len) {
+		iov[0].iov_base = (uint8_t *)iov[0].iov_base + to_consume;
+		iov[0].iov_len -= to_consume;
+	} else {
+		*iov_count -= 1;
```
shefty, line 91: It took me waaaay too much time looking at this function to see if it was returning what we actually want.  The case where we fully consume the iov with an iov_count > 1 is the confusing part.  That flows through the if statement after out: and adjusts iov_base and iov_len by adding/subtracting 0.

In any case, I think we need to make the function more intuitive to understand.  Separating iov_count == 1 into its own section can help.  I also want to avoid repeating the same check twice - once in the for-loop and again after.  This is what I've come up with so far:

```
if (*iov_count == 1) {
    if (to_consume < iov[0].iov_len)
        ofi_consume_sge(iov, to_consume);
    else
        *iov_count = 0;
    return;
}

for (i = 0; i < *iov_count; i++) {
    if (to_consume < iov[0].iov_len) {
        ofi_consume_sge(iov, to_consume)
        break;
    }
    to_consume -= iov[i].iov_len;
}
if (i < *iov_count) {
    memmove(...)
}
iov_count -= i;
```
This removes the -1 adjustment, which now I'm questioning.  Whatever i is after the for loop is the number of iov's that were fully consumed.

shefty: ofi_consume_sge() would be the iov_base + to_consume / iov_len -= to_consume portion of the original code.  I also noticed that the if statement in my above for loop uses iov[0] instead of iov[i]..

shefty: And another note, in the code in my comment, we can move the memmove() into the for-loop in the if statement.

lrbison: Apologies Sean, I will be unable to work on this for the next 3 weeks while I'm OOO.

lrbison: Between my first rev and this one you drafted we've pretty substantially re-written the function, so I took a stab at just writing it from scratch.  I liked what I came up with more than your proposed solution.  It has the added benefit that calling consume with iov_count=0 doesn't segfault, never performs a memmove of 0 bytes, and I find it quite readable.  Let me know if it's readable to you as well, or if you can think of another corner case I missed.

shefty (commented):
I don't see errors in the new structure, but it no longer optimizes for the common case.  With count = 1, the old code had 1 branch and 2 assignments.  The new code ends up with 3 branches plus 4 assignments.

lrbison:
I couldn't work out an readable case that was optimized for common case without just explicitly short-circuiting the common case.

Hopefully this is sufficient.  consuming part of a single iov is assumed to be common case: 1 branch & 1 assignment.  Second is 2 branch & 1 assignment for consuming the whole iov.  Common case has the overhead of the first 2 branches and then does the loop as before.

src/iov.c
```diff
@@ -63,24 +63,39 @@ uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, uint64_t io
...
+	assert(to_consume && *iov_count);
+	assert(to_consume <= ofi_total_iov_len(iov, *iov_count));
+
+	if (*iov_count == 1 && iov[0].iov_len > to_consume) {
+		iov[0].iov_len -= to_consume;
+		return;
+	} else if (*iov_count == 1) {
+		*iov_count = 0;
+		return;
+	}
```
shefty, line 78: I'd adjust the above to:
```
if (*iov_count == 1) {
    if (iov[0].iov_len == to_consume) {
        *iov_count = 0;
        return;
    }
    goto out;
}
```
See comment below.

src/iov.c
```diff
@@ -63,24 +63,39 @@ uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, uint64_t io
...
+		cur_desc++;
+		*iov_count -= 1;
+	}
+	if (*iov_count == 0)
+		return;
 
-	if (*iov_count == 1)
-		goto out;
+	cur_iov->iov_len -= to_consume;
+	cur_iov->iov_base = (uint8_t *)iov[0].iov_base + to_consume;
```
shefty, line 92: FWIW, this code is updating the iov that will become iov[0].  It's just being done prior to the memmove below.  If this code were moved to after the if statement below, cur_iov could be replaced with iov[0], which is what the original code had.  The iov_count == 1 check above could then use a goto to fall to it, also similar to the original code.  The missing piece from the original code was setting iov_count = 0 when fully consuming, which is now handled above.

shefty (commented):
Who knew such a simple function could require so much brainpower.

src/iov.c
shefty: An if statement with an && condition technically ends up as multiple branches when converted to assembly.  My intent with the above was to avoid duplicating the same check twice (i.e. iov_count == 1)

lrbison: Ah, right, because of guarantee for short-circuit evaluation.  (Not true for fortran!)

shefty:
bot:aws:retest

shefty:
Intel CI failures -- tcp provider (which is fixed upstream, so this is likely resolved by rebasing), and ucx provider which went bonkers.  Changes look good to me.  Waiting on AWS CI

shefty (approved):

lrbison:
Thanks, I was about to re-trigger aws.  Last time several of the one-node intelmpi tests failed with either segfaults or timeouts.  Lets see if that happens again.

shijin-aws:
@lrbison  you need to rebase on latest libfabric main, the one-sided intel mpi segfault is fixed by https://github.com/ofiwg/libfabric/commit/22b253ec95c5ddbec0e05dfe5cbe877fee7d8a07 so I enabled such test.

lrbison:
Thank you @shijin-aws.  Rebased!

PR #9057, opened 2023-06-19
prov/efa: Only select readbase_rtm when both sides support rdma-read

shijin-aws:
Currently, efa_rdm_msg_select_rtm return readbase_rtm if local ep support rdma-read. This is wrong because the read protocols require both local and peer support rdma-read. This patch fixes this issue.


shijin-aws:
Need to add a fabtests, will update.

wzamazon (approved):

wenduwan (approved):

PR #9060, opened 2023-06-20
[v1.18.x] prov/efa: Only select readbase_rtm when both sides support rdma-read

shijin-aws:
Currently, rxr_msg_select_rtm return readbase_rtm if local ep support rdma-read. This is wrong because the read protocols require both local and peer support rdma-read. This patch fixes this issue.

(cherry picked from commit 085c21c)


wzamazon (approved):

wenduwan (approved):

lrbison (approved):

shijin-aws:
Pure efa provider change, ignoring Intel CI failure

PR #9064, opened 2023-06-21
prov/sm2: Fix locking

a-szegel:
Add tx_lock to generic inject path.
Add tx_lock (which currently controls freestack access) to buffer return.

### DIFF
--- a/prov/sm2/src/sm2_msg.c
+++ b/prov/sm2/src/sm2_msg.c
@@ -182,6 +182,7 @@ static ssize_t sm2_generic_inject(struct fid_ep *ep_fid, const void *buf,
 
 	peer_smr = sm2_peer_region(ep, peer_gid);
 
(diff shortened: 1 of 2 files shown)


aingerson (approved):

prov/sm2/src/sm2_msg.c
```diff
@@ -182,13 +182,15 @@ static ssize_t sm2_generic_inject(struct fid_ep *ep_fid, const void *buf,
 
 	peer_smr = sm2_peer_region(ep, peer_gid);
 
+	ofi_spin_lock(&ep->tx_lock);
```
shijin-aws, line 185: should we use ofi gen lock? which will be no-op unless it requires thread safe.

shijin-aws: offline synced, will be a separate PR.

PR #9065, opened 2023-06-21
fabtests: Fix array OOB during fabtest list parsing

lrbison:
An incorrect index was used to attempt to set test_sizes.enable_flags. For small list sizes (<3 chars) malloc would become corrupted and future calls would fail unexpectedly.

Setting this variable to 0 was uncessary as the calloc() cleared the memory already.


shijin-aws (approved):

lrbison:
The erroneous function doesn't exist in v1.18.x, no backport required.

shefty (approved):
thanks!

PR #9067, opened 2023-06-22
prov/efa: set and use pke->payload and pke->payload_size

wzamazon:
This patch added following changes to properly set payload and payload_size of a packet:

1. Upon receiving a packet, set its payload and payload_size if the packet contains data.

2. When clone a packet, set cloned packet's payload and payload_size if the packet contains data.

With the above changes, a packet's payload and payload_size is always set properly.  This patch then started using payload and payload_size:

1. it used payload_size to replace fucntion calls to rxr_pkt_data_size() and rxr_pkt_req_data_size(), and eliminated these two functions.

2. it used payload in rxr_pkt_copy_data_to_ope() and renamed the function efa_rdm_pke_copy_payload()

It then eliminated two functions that calculate payload size on the fly: rxr_pkt_data_size() rxr_pkt_req_data_size()

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke_utils.h
+++ b/prov/efa/src/rdm/efa_rdm_pke_utils.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Amazon.com, Inc. or its affiliates.
+ * Copyright (c) Amazon.com, Inc. or its affiliates.
  * All rights reserved.
  *
  * This software is available to you under a choice of one of two
@@ -31,20 +31,26 @@
  * SOFTWARE.
  */
 
-#ifndef _RXR_PKT_HDR_H
-#define _RXR_PKT_HDR_H
+#ifndef _EFA_RDM_PKE_UTILS_H
+#define _EFA_RDM_PKE_UTILS_H
 
+#include "efa_rdm_pke.h"
 #include "efa_rdm_protocol.h"
+#include "efa_rdm_pkt_type.h"
 
 /**
- * @brief return the seg_offset field in a packet entry
+ * @brief return the segment offset of user data in packet entry
+ *
+ * segment_offset is the user data's offset in repect of user's
+ * buffer.
  *
  * @param[in]	pkt_entry	packet entry
  * @return	the value of seg_offset in the packet
  */
 static inline
-size_t rxr_pkt_hdr_seg_offset(char *pkt)
+size_t efa_rdm_pke_get_segment_offset(struct efa_rdm_pke *pke)
 {
+	int pkt_type, hdr_offset;
 	static const int offset_of_seg_offset_in_header[] = {
 		[RXR_DATA_PKT] = offsetof(struct rxr_data_hdr, seg_offset),
 		[RXR_MEDIUM_MSGRTM_PKT] = offsetof(struct rxr_medium_rtm_base_hdr, seg_offset),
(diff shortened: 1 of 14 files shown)


prov/efa/src/rdm/efa_rdm_pkt_type.h
```diff
@@ -205,4 +205,34 @@ bool efa_rdm_pkt_type_contains_data(int pkt_type)
...
+		       pkt_type == RXR_DC_EAGER_MSGRTM_PKT ||
+		       pkt_type == RXR_DC_EAGER_TAGRTM_PKT ||
+		       pkt_type == RXR_DC_MEDIUM_MSGRTM_PKT ||
+		       pkt_type == RXR_DC_MEDIUM_TAGRTM_PKT ||
+		       pkt_type == RXR_DC_LONGCTS_MSGRTM_PKT ||
+		       pkt_type == RXR_DC_LONGCTS_TAGRTM_PKT ||
+		       pkt_type == RXR_DC_EAGER_RTW_PKT ||
+		       pkt_type == RXR_DC_LONGCTS_RTW_PKT ||
+		       pkt_type == RXR_RUNTREAD_MSGRTM_PKT ||
+		       pkt_type == RXR_RUNTREAD_TAGRTM_PKT);
```
wenduwan, line 231: Why do we need this here? `if (pkt_type >= RXR_REQ_PKT_BEGIN)` should cover it.

The assertion is more like a unit test IMO.

wzamazon: I think this is what assertion is for, which is to fail if something should be happen actually happened.

prov/efa/src/rdm/rxr_pkt_type_base.c
```diff
@@ -524,27 +462,26 @@ int rxr_pkt_copy_data_to_cuda(struct efa_rdm_ep *ep,
...
+ *                      	buffer.
  * @return		On success, return 0
  * 			On failure, return libfabric error code
  */
-ssize_t rxr_pkt_copy_data_to_ope(struct efa_rdm_ep *ep,
-				      struct efa_rdm_ope *ope,
-				      size_t data_offset,
-				      struct efa_rdm_pke *pkt_entry,
-				      char *data, size_t data_size)
+ssize_t efa_rdm_pke_copy_payload(struct efa_rdm_pke *pke,
```
shijin-aws, line 472: can we name it explicitly as `copy_payload_to_ope`

prov/efa/src/rdm/efa_rdm_pkt_type.h
wenduwan: This assertion should be true in any case, not limited to this function branch.

My general comment is, invariants should be tested in dedicated unit tests rather than assertions in business logic.

wzamazon: I am not sure. This is not a test.

prov/efa/src/rdm/rxr_pkt_type_base.c
wzamazon: I can. but techniquely, we are not copy data to ope, but copy data to user buffer. ope has the information of the user buffer,

wzamazon: I think I will name it `copy_payload_to_ope`, and rename the other funciton "init_payload_from_ope"

wzamazon: I have an commit that merge this file into efa_rdm_pke_utils.c, can I do the renaming in that commit?

shijin-aws: Sure

prov/efa/src/rdm/efa_rdm_pke_utils.h
```diff
@@ -55,11 +61,21 @@ size_t rxr_pkt_hdr_seg_offset(char *pkt)
 		[RXR_RUNTREAD_TAGRTM_PKT] = offsetof(struct rxr_runtread_rtm_base_hdr, seg_offset),
 	};
 
-	int pkt_type = rxr_get_base_hdr(pkt)->type;
-	int hdr_offset = offset_of_seg_offset_in_header[pkt_type];
+	pkt_type = rxr_get_base_hdr(pke->wiredata)->type;
+	if (efa_rdm_pkt_type_is_eager(pkt_type) ||
```
shijin-aws, line 65: Trying to think of a better way to check this to make it cleaner... Does it mean tthe pkt contains data and it's req? Does `...contain_data(pkt_type) && ...is_req(pkt_type)` work?

shijin-aws (approved):
only a minor comment

prov/efa/src/rdm/efa_rdm_pke_utils.h
wzamazon: It is a packet that has data but do not have `seg_offset`. which is `contains_data() && (!is_mulreq) && (!is_data))` I think it is not better than what we have now

shijin-aws: then we should add a comment or make this check an inline function to say it explicitly

wzamazon: good idea. let me do that

vidsouza:
bot:aws:retest

PR #9070, opened 2023-06-22
prov/efa: Switch to use SM2 as default provider

a-szegel:


wzamazon (approved):

a-szegel:
I don't think this can go in until Sai's PR goes in at the earliest... but I just wanted to put it up for when we need it

wzamazon:
>I don't think this can go in until Sai's PR goes in at the earliest... but I just wanted to put it up for when we need it

I see. I converted it to draft, so we do not merge it by accident.

a-szegel:
Alexia told me intel canceled their CI to conserve resource... intel failures unrelated.

PR #9073, opened 2023-06-24
prov/efa: migrate all NON req packets 

wzamazon:
Move all the packet types that are not REQ packets from rxr to efa_rdm


wzamazon:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -595,3 +596,80 @@ ssize_t efa_rdm_ep_trigger_handshake(struct efa_rdm_ep *ep,
 
 	return 0;
 }
+
```
shijin-aws, line 599: why put this function in ep_utils.c rather than efa_rdm_pke_nonreq.c?

shijin-aws (approved):

prov/efa/src/rdm/efa_rdm_ep_utils.c
wzamazon: This function operates on an endpoint, therefore need to be in ep,

PR #9078, opened 2023-06-27
prov/efa: Add missing locking around cntr progress

shijin-aws:
ofi_cntr_progress will call ep->progress for every ep
    it was bound to. There was a ep->lock in efa_rdm_ep_progress,
    but it was recently changed to srx_ctx->lock
    and moved from efa_rdm_ep_progress() to its upper layer callers.
    This patch adds this locking around ofi_cntr_progress.

Also added a test that poll completions with cntr instead of queue, to cover this code path.

### DIFF
--- a/prov/efa/src/efa_cntr.c
+++ b/prov/efa/src/efa_cntr.c
@@ -74,6 +74,16 @@ static int efa_cntr_wait(struct fid_cntr *cntr_fid, uint64_t threshold, int time
 	return ret;
 }
 
+static void efa_cntr_progress(struct util_cntr *cntr)
+{
+	struct util_srx_ctx *srx_ctx;
+	srx_ctx = cntr->domain->srx->ep_fid.fid.context;
+
+	ofi_genlock_lock(srx_ctx->lock);
+	ofi_cntr_progress(cntr);
+	ofi_genlock_unlock(srx_ctx->lock);
+}
+
 int efa_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
 		  struct fid_cntr **cntr_fid, void *context)
 {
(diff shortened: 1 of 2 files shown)


shefty:
Please pick for v1.18.x

shijin-aws:
>Please pick for v1.18.x

The bug was not introduced to 1.18, so no need for backport

prov/efa/src/efa_cntr.c
```diff
@@ -85,7 +95,7 @@ int efa_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
 		return -FI_ENOMEM;
 
 	ret = ofi_cntr_init(&efa_prov, domain, attr, cntr,
-			    &ofi_cntr_progress, context);
```
wzamazon, line 88: Was ofi_cntr_progress used else where?

wzamazon (approved):

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

prov/efa/src/efa_cntr.c
shijin-aws: It's the only place in efa provider.

wzamazon: It is out of scope of this PR, but I wonder if other provider use this function and need to make similar change.
I

shijin-aws: This locking is required after efa onboards the util SRX framework, which requires the owner provider acquire a domain level lock (the srx_ctx->lock here) outside the progress engine. Currently efa is the only owner provider that onboards the util SRX, so other providers shouldn't need this change

PR #9079, opened 2023-06-27
prov/efa: fix bugs in efa_rdm_ope_prepare_to_post_send() and add unit tests

wzamazon:


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -538,15 +538,19 @@ ssize_t efa_rdm_ope_prepare_to_post_send(struct efa_rdm_ope *ope,
...
 			pkt_entry_data_size_vec[i] = single_pkt_entry_data_size;
 
-		remainder = total_pkt_entry_data_size - (*pkt_entry_cnt - 1) * single_pkt_entry_data_size;
-		pkt_entry_data_size_vec[*pkt_entry_cnt - 1] = remainder;
-		return 0;
+		remainder = total_pkt_entry_data_size - (*pkt_entry_cnt) * single_pkt_entry_data_size;
+		if (single_pkt_entry_data_size + remainder <= single_pkt_entry_max_data_size) {
+			pkt_entry_data_size_vec[*pkt_entry_cnt - 1] += remainder;
+		} else {
+			pkt_entry_data_size_vec[*pkt_entry_cnt] = remainder;
```
shijin-aws, line 548: So before the change, remainder was sent in a separate pkt, could it explain the performance regression we were seeing?

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -520,11 +520,10 @@ ssize_t efa_rdm_ope_prepare_to_post_send(struct efa_rdm_ope *ope,
 		single_pkt_entry_max_data_size = efa_rdm_txe_max_req_data_capacity(ep, ope, pkt_type);
 		assert(single_pkt_entry_max_data_size);
 
-		if (efa_mr_is_cuda(ope->desc[0])) {
-			if (ep->sendrecv_in_order_aligned_128_bytes)
-				memory_alignment = EFA_RDM_IN_ORDER_ALIGNMENT;
-			else
-				memory_alignment = EFA_RDM_CUDA_MEMORY_ALIGNMENT;
+		if (ep->sendrecv_in_order_aligned_128_bytes) {
+			memory_alignment = EFA_RDM_IN_ORDER_ALIGNMENT;
```
shijin-aws, line 524: do we need to backport this?

prov/efa/src/rdm/efa_rdm_ope.c
wzamazon: > So before the change, remainder was sent in a separate pkt,

Yes.

> could it explain the performance regression we were seeing?

I think so.

prov/efa/src/rdm/efa_rdm_ope.c
wzamazon: More like re-implement the patch on v1.18.x because the code has changed a lot.

shijin-aws (approved):

PR #9081, opened 2023-06-27
prov/efa: Remove rx_iov_limit and tx_iov_limit environment variables

sunkuamzn:
If these environment variables are set to values larger than what the shared memory provider can support, the expected behavior is unclear. So remove them.


prov/efa/src/efa_env.c
```diff
@@ -149,8 +149,6 @@ void efa_env_param_get(void)
 			    &efa_mr_max_cached_size);
 	fi_param_get_size_t(&efa_prov, "tx_size", &efa_env.tx_size);
 	fi_param_get_size_t(&efa_prov, "rx_size", &efa_env.rx_size);
-	fi_param_get_size_t(&efa_prov, "tx_iov_limit", &efa_env.tx_iov_limit);
-	fi_param_get_size_t(&efa_prov, "rx_iov_limit", &efa_env.rx_iov_limit);
```
wzamazon, line 153: I think you need to something like we did for other deprecated envar, abort if some user try to set it. (like like 110). Also check whether these evnvars are listed in our manpage.

sunkuamzn: fixed

wzamazon (approved):

PR #9084, opened 2023-06-27
prov/efa: Fix a bug in desc convert and adding tests

shijin-aws:
In efa_rdm_msg_sendv, when desc is NULL, we shouldn't try to get desc for shm.

Also add a unit test to scan all fi_send* and fi_tsend APIs for such issue.

### DIFF
--- a/prov/efa/test/efa_unit_test_send.c
+++ b/prov/efa/test/efa_unit_test_send.c
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) Amazon.com, Inc. or its affiliates. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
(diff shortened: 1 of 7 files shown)


prov/efa/test/efa_unit_test_send.c
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+#include "efa_unit_tests.h"
+#include "ofi_util.h"
+#include "efa_rdm_ep.h"
+
+void test_efa_rdm_msg_send_to_local_peer_with_null_desc(struct efa_resource **state)
+{
+        struct efa_resource *resource = *state;
+        char buf[10];
```
sunkuamzn, line 40: nit: can replace all `10` with a variable

prov/efa/test/efa_unit_test_send.c
```diff
@@ -0,0 +1,106 @@
...
+	ret = fi_av_insert(resource->av, &raw_addr, 1, &addr, 0 /* flags */, NULL /* context */);
+	assert_int_equal(ret, 1);
+
+        for (i = 0; i < 10; i++)
+                buf[i] = 'a' + i;
+
+        iov.iov_base = buf;
+        iov.iov_len = 10;
+
+        msg.msg_iov = &iov;
```
sunkuamzn, line 66: nit: can move the msg and tmsg construction into a small function and reuse it. Function can be in `efa_unit_test_common.c`

sunkuamzn (approved):

wzamazon (approved):

wenduwan (approved):

prov/efa/test/efa_unit_test_send.c
shijin-aws: I defined a macro for it.

prov/efa/test/efa_unit_test_send.c
shijin-aws: Done

PR #9090, opened 2023-06-28
UCX CI Summary/improvements

nikhilnanal:
CI code changes to incorporate fabtests summary for ucx provider
and improvements in build step for ucx.

### DIFF
--- a/contrib/intel/jenkins/build.py
+++ b/contrib/intel/jenkins/build.py
@@ -12,7 +12,7 @@ import common
 import re
 import shutil
 
-def build_libfabric(libfab_install_path, mode, cluster=None):
+def build_libfabric(libfab_install_path, mode, cluster=None, ucx=None):
 
     if (os.path.exists(libfab_install_path) != True):
         os.makedirs(libfab_install_path)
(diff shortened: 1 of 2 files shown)


contrib/intel/jenkins/build.py
```diff
@@ -31,11 +31,12 @@ def build_libfabric(libfab_install_path, mode, cluster=None):
     else:
         prov_list = common.default_prov_list
 
-    if ('ucx' in libfab_install_path):
-        config_cmd.append('--enable-ucx=yes')
-
     for prov in prov_list:
-       config_cmd.append(f'--enable-{prov}={enable_prov_val}')
+       if (prov == 'ucx'):
```
zachdworkin, line 35: why is this not an 'and'?

zachdworkin (approved):
1 nitpick but overall looks good

contrib/intel/jenkins/build.py
nikhilnanal: > why is this not an 'and'?
it is to avoid the case when ucx flag is false but 
for loop iterates on prov_list which will lands in the else case 
for ucx and still enable it.

PR #9092, opened 2023-06-29
prov/tcp: Restrict which EPs can be opened per domain

shefty:
The tcp provider cannot mix rdm and msg endpoints on the same domain because of locking restrictions and optimizations targeting rdm endpoint support.  Fail calls that try to open and endpoint type not supported by the domain.


shefty:
bot:aws:retest

shefty:
bot:aws:retest

PR #9095, opened 2023-06-29
prov/sm2: fix sm2 locking

a-szegel:


a-szegel:
This PR still leaves 1 issue with locking, but will address that in my next PR.  The remaining issue is sm2 needs to pre-allocate the entire SM2 file.  That change was different enough that I wanted it to get its own PR.

sunkuamzn:
IIUC the reason we need locks is because the lockless queue that we're using is designed for a multiple-writer-single-reader use. So why not just put the locks in the `sm2_fifo_read` and `sm2_fifo_write` functions so that we minimize the code that's under locks?

prov/sm2/src/sm2.h
```diff
@@ -291,9 +291,7 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
 	struct sm2_region *self_region = sm2_mmap_ep_region(map, ep->gid);
 
 	if (smr_freestack_isempty(sm2_freestack(self_region))) {
-		sm2_progress_recv(ep);
-		if (smr_freestack_isempty(sm2_freestack(self_region)))
-			return -FI_EAGAIN;
+		return -FI_EAGAIN;
```
aingerson, line 293: drop brackets now

aingerson (commented):
Just one small thing. Changes look good though

aingerson:
@sunkuamzn The whole progress is locked already with the ep lock (this is to make sure that incoming commands are matched to receives in the correct order. If you wanted to reduce the locked code, you can but it's going to have to include the read + get_msg function which in my opinion is going to add code complexity and a lot of lock/unlock calls (which are probably going to be no-ops anyways) in order to support the multi-threaded case. For now at least, I think this is the easier option but we can look into trade offs for doing finer locking within that path later.

prov/sm2/src/sm2_ep.c
```diff
@@ -257,6 +259,8 @@ static void cleanup_shm_resources(struct sm2_ep *ep)
 			"until SHM file is cleared\n",
 			ep->gid, ep->gid);
 	}
+
+	ofi_genlock_unlock(&ep->util_ep.lock);
```
aingerson, line 263: You sleep in this function. If you lock and sleep that is bad. I would either remove the retry/sleep or unlock, relock. (I still don't like that sleep.... if we remove it that would make me happier)

a-szegel: I can remove the lock b/c we should only be calling this function after we are destroying the endpoint.

aingerson (approved):

PR #9096, opened 2023-06-29
prov/sm2: Allocate and map maximum size shm file at init

a-szegel:
This change causes the entire shared memory file to be allocated and mapped into each processes address space on startup. This change fixes a multi-threaded race condition where one thread calls remap in sm2_av_insert() while another thread is in the middle of a sm2_fifo_read(), or sm2_fifo_write(). We need increased memory usage in order to provide correctness to the user.

### DIFF
--- a/prov/sm2/src/sm2_coordination.c
+++ b/prov/sm2/src/sm2_coordination.c
@@ -49,8 +49,6 @@
 
 static void sm2_file_attempt_shrink(struct sm2_mmap *map);
 int sm2_entry_lookup(const char *name, struct sm2_mmap *map);
-static void sm2_file_extend_for_region(struct sm2_mmap *map,
-				       int last_valid_entry);
 
 /*
  * Sends signal 0 to the pid, if the call succeeds, it means the pid exists.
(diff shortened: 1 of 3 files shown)


prov/sm2/src/sm2_coordination.c
```diff
@@ -312,6 +312,16 @@ ssize_t sm2_file_open_or_create(struct sm2_mmap *map_shared)
...
 	}
 
+	/* Remap the file to be big enough to hold every region upfront.
+	 * This is required to get rid of race conditions that may have
+	 * previously occured when we tried to remap in sm2_av_insert(),
+	 * sm2_fifo_send() or sm2_fifo_recv().
+	 */
+	header = (struct sm2_coord_file_header *) map_shared->base;
+	max_file_size = header->ep_regions_offset +
+			header->ep_region_size * SM2_MAX_UNIVERSE_SIZE;
```
sunkuamzn, line 322: Can we also reduce `SM2_MAX_UNIVERSE_SIZE`? With 2048, we would need about 8GB of memory for every run (2048 * 1024 * 4096) which seems a lot to me.

aingerson: I agree with @sunkuamzn. Now that we allocate the whole, we've got to play it safer with how much memory we anticipate.

a-szegel: done

sunkuamzn (approved):

aingerson (approved):

PR #9098, opened 2023-06-30
contrib/intel/jenkins: use python to invoke launch.py

zachdworkin:
Use python to invoke launch.py for daos instead of ./ since ./will use the latest version.


zachdworkin:
ci failed because of tcp race condition.

PR #9101, opened 2023-07-01
prov/sm2: Default to ep->domain->threading

a-szegel:
SM2 is currently hard coding FI_THREAD_SAFE, but the correct behavior for the provider is to assume the domain thread safety-ness.

### DIFF
--- a/prov/shm/src/smr_domain.c
+++ b/prov/shm/src/smr_domain.c
@@ -104,7 +104,6 @@ int smr_domain_open(struct fid_fabric *fabric, struct fi_info *info,
 		return ret;
 	}
 
-	smr_domain->util_domain.threading = FI_THREAD_SAFE;
 	smr_fabric = container_of(fabric, struct smr_fabric, util_fabric.fabric_fid);
 	ofi_mutex_lock(&smr_fabric->util_fabric.lock);
 	smr_domain->fast_rma = smr_fast_rma_enabled(info->domain_attr->mr_mode,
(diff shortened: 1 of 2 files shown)


a-szegel:
@shefty Wondering why https://github.com/ofiwg/libfabric/commit/7e584f6170b2b1d142ece805aee45d95f9e9abf5 is required.

a-szegel:
the SM2 provider (via Nemesis) is ok with multiple processes touching the file... I can remove the SHM commit

a-szegel:
I thought all the SHM/SM2 ofi_genlock's were only being used to protect the provider from thread saftey issues (in send, progress, and cq)... I wasn't aware that an ofi_genlock in SHM was being used as an interprocess lock.

aingerson:
@shefty Are you saying access across processes and access across threads are the same in the eyes of the API? I would argue the multiple process access is a shm/sm2 specific issue that needs to be controlled within the provider, not tied to something through the API, like the threading level.

shefty:
I'm saying that this is a shared memory implementation issue.  There's no changes to the API 7e584f6170b2b1d142ece805aee45d95f9e9abf5.  That sets the value of an internal structure, which has the impact of always creating locks.

aingerson:
Yeah, I know it's not a change to the API for the app, but I think we don't want this to be turned on. We already handle multi process access in the region itself so we don't need extra locking/access control for the intra process resources like the ep, av, etc, unless the application requests threading support specifically.

aingerson (approved):

a-szegel:
Alexia talked to Sean offline and got his approval.

PR #9107, opened 2023-07-03
fabtests: Add HMEM support to fi_unexpected_msg test

sunkuamzn:
Allows for testing unexpected messages to and from device memory

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -401,6 +401,14 @@ void ft_fill_mr_attr(struct iovec *iov, int iov_count, uint64_t access,
 	}
 }
 
+bool ft_need_mr_reg(struct fi_info *fi)
+{
+	return (fi->caps & (FI_RMA | FI_ATOMIC)) ||
+	       (fi->domain_attr->mr_mode & FI_MR_LOCAL) ||
+	       ((fi->domain_attr->mr_mode & FI_MR_HMEM) &&
+		(opts.options & FT_OPT_USE_DEVICE));
+}
+
 int ft_reg_mr(struct fi_info *fi, void *buf, size_t size, uint64_t access,
 	      uint64_t key, enum fi_hmem_iface iface, uint64_t device,
 	      struct fid_mr **mr, void **desc)
@@ -410,11 +418,7 @@ int ft_reg_mr(struct fi_info *fi, void *buf, size_t size, uint64_t access,
 	int ret;
 	uint64_t flags;
 
-	if (((!(fi->domain_attr->mr_mode & FI_MR_LOCAL) &&
-	      !(opts.options & FT_OPT_USE_DEVICE)) ||
-	     (!(fi->domain_attr->mr_mode & FI_MR_HMEM) &&
-	      opts.options & FT_OPT_USE_DEVICE)) &&
-	    !(fi->caps & (FI_RMA | FI_ATOMIC)))
+	if (!ft_need_mr_reg(fi))
 		return 0;
 
 	iov.iov_base = buf;
--- a/fabtests/functional/unexpected_msg.c
+++ b/fabtests/functional/unexpected_msg.c
@@ -47,6 +47,7 @@
 #include <rdma/fi_cm.h>
 
 #include "shared.h"
+#include "hmem.h"
 
 
 static size_t concurrent_msgs = 4;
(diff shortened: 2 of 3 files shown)


fabtests/functional/unexpected_msg.c
```diff
@@ -71,9 +79,25 @@ static int alloc_bufs(void)
 	rx_buf = buf;
 	tx_buf = (char *) buf + rx_size * concurrent_msgs;
 
-	if (fi->domain_attr->mr_mode & FI_MR_LOCAL) {
-		ret = fi_mr_reg(domain, buf, buf_size, FI_SEND | FI_RECV,
-				 0, FT_MR_KEY, 0, &mr, NULL);
+	no_mr_reg = (((!(fi->domain_attr->mr_mode & FI_MR_LOCAL) &&
```
sunkuamzn, line 82: I copied this logic from https://github.com/ofiwg/libfabric/blob/e0dc241d428dd898f0af7a49fd1287a10175f82e/fabtests/common/shared.c#L413-L418

wzamazon:
Do you need to update the pytest script to add a test case for this?

sunkuamzn:
> Do you need to update the pytest script to add a test case for this?

Yes making a new PR for that

fabtests/functional/unexpected_msg.c
```diff
@@ -71,9 +79,25 @@ static int alloc_bufs(void)
...
+		       !(opts.options & FT_OPT_USE_DEVICE)) ||
+		      (!(fi->domain_attr->mr_mode & FI_MR_HMEM) &&
+		       opts.options & FT_OPT_USE_DEVICE)) &&
+		     !(fi->caps & (FI_RMA | FI_ATOMIC)));
```
shefty, line 86: Since this is copied, let's move it into a function.

fabtests/functional/unexpected_msg.c
```diff
@@ -71,9 +79,25 @@ static int alloc_bufs(void)
...
+
+	if (!no_mr_reg) {
+		iov.iov_base = buf;
+		iov.iov_len = buf_size;
+
+		access = ft_info_to_mr_access(fi);
+
```
shefty, line 93: nit: remove 2 blank lines above and remove 'access' variable.  Just place ft_info_to_mr_access()  directly below.  (The changes in this section has a few too many blank lines.)

fabtests/functional/unexpected_msg.c
```diff
@@ -71,9 +79,25 @@ static int alloc_bufs(void)
...
+		ft_fill_mr_attr(&iov, 1, access, FT_MR_KEY, opts.iface,
+				opts.device, &mr_attr);
+
+		flags = (opts.iface) ? FI_HMEM_DEVICE_ONLY : 0;
+
+		ret = fi_mr_regattr(domain, &mr_attr, flags, &mr);
+
```
shefty, line 100: nit: remove blank line

fabtests/functional/unexpected_msg.c
```diff
@@ -57,12 +58,19 @@ static bool send_data = false;
 static int alloc_bufs(void)
 {
 	int ret;
+	uint64_t access, flags;
+	struct iovec iov = {0};
```
shefty, line 62: nit: no need to initialize to 0

fabtests/functional/unexpected_msg.c
```diff
@@ -57,12 +58,19 @@ static bool send_data = false;
...
+	struct fi_mr_attr mr_attr = {0};
+	bool no_mr_reg = false;
```
shefty, line 64: nit: no need to initialize

sunkuamzn:
Fixed, thank you @shefty

fabtests/common/shared.c
```diff
@@ -401,6 +401,15 @@ void ft_fill_mr_attr(struct iovec *iov, int iov_count, uint64_t access,
 	}
 }
 
+bool ft_need_mr_reg(struct fi_info *fi)
+{
+	return (((fi->domain_attr->mr_mode & FI_MR_LOCAL) ||
+		  (opts.options & FT_OPT_USE_DEVICE)) &&
+		 ((fi->domain_attr->mr_mode & FI_MR_HMEM) ||
+		  !(opts.options & FT_OPT_USE_DEVICE))) ||
+		(fi->caps & (FI_RMA | FI_ATOMIC));
```
shefty, line 410: I know this is the same logic as before (just negated), but is this correct?

The last part -- using RMA -- makes sense.  The other checks look backwards.

For example, combining FI_MR_LOCAL with using device memory _looks_ wrong. FI_MR_LOCAL is independent of device buffers.  The next part of the check is also confusing.  FI_MR_HMEM being OR'ed against NOT using device buffers is unexpected.  FI_MR_HMEM only applies to device buffers.  The following check seem more natural to me and are easier to understand:

(fi->domain_attr->mr_mode & FI_MR_LOCAL) ||
(fi->caps & (FI_RMA | FI_ATOMIC) ||
((fi->domain_attr->mr_mode & FI_MR_HMEM) &&
 (opts.options & FT_OPT_USE_DEVICE))

Sooo... *if* the above checks *don't* work, why not?

sunkuamzn: I *think* the intended logic of the old code was something like this: we need memory registration when (1) RMA or Atomic capabilities are present (2) FT_OPT_USE_DEVICE and FI_MR_HMEM are set (3) FT_OPT_USE_DEVICE is not set and FI_MR_LOCAL is set. Since most of our providers support RMA and/or Atomic, we probably never ran the other paths.

I've updated it with the logic you propose.

shefty:
Intel CI failure is unrelated.  Given that other CI passed, I'm going to assume the logic changes are correct.  Thanks!

PR #9108, opened 2023-07-03
prov/efa: fix rxr in comments and documents

wzamazon:
Replace the term RXR will proper terms in comments and documents


wenduwan (approved):
Thank you!

PR #9112, opened 2023-07-05
prov/efa: add use_huge_page to efa_env

wzamazon:
Also add unit tests

### DIFF
--- a/prov/efa/src/efa_env.c
+++ b/prov/efa/src/efa_env.c
@@ -68,6 +68,7 @@ struct efa_env efa_env = {
 	.rnr_retry = 3, /* Setting this value to EFA_RNR_INFINITE_RETRY makes the firmware retry indefinitey */
 	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
 	.use_sm2 = false,
+	.huge_page_setting = EFA_ENV_HUGE_PAGE_UNSPEC,
 };
 
 /**
@@ -164,6 +165,12 @@ void efa_env_param_get(void)
 	fi_param_get_size_t(&efa_prov, "inter_max_gdrcopy_message_size",
 			    &efa_env.efa_max_gdrcopy_msg_size);
 	fi_param_get_bool(&efa_prov, "use_sm2", &efa_env.use_sm2);
+
+	int use_huge_page;
+	if (fi_param_get_bool(&efa_prov, "use_huge_page", &use_huge_page) ==0) {
+		efa_env.huge_page_setting = use_huge_page ? EFA_ENV_HUGE_PAGE_ENABLED : EFA_ENV_HUGE_PAGE_DISABLED;
+	}
+
 	efa_fork_support_request_initialize();
 }
 
@@ -230,6 +237,11 @@ void efa_env_define()
 			"The maximum number of bytes that will be eagerly sent by inflight messages uses runting read message protocol (Default 307200).");
 	fi_param_define(&efa_prov, "use_sm2", FI_PARAM_BOOL,
 			"Use the experimental shared memory provider SM2 for intra node communication.");
+	fi_param_define(&efa_prov, "use_huge_page", FI_PARAM_BOOL,
+			"Whether EFA provider can use huge page memory for internal buffer. "
+			"Using huge page memory has a small performance advantage, but can "
+			"cause system to run out of huge page memory. By default, EFA provider "
+			"will use huge page unless FI_EFA_FORK_SAFE is set to 1/on/true.");
 }
 
 
--- a/prov/efa/src/efa_fork_support.c
+++ b/prov/efa/src/efa_fork_support.c
@@ -57,8 +57,37 @@ void efa_fork_support_request_initialize()
 	 * libibverbs' fork support are set. These variables are
 	 * defined by ibv_fork_init(3).
 	 */
-	if (fork_support_requested || getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE"))
+	if (fork_support_requested || getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE")) {
 		g_efa_fork_status = EFA_FORK_SUPPORT_ON;
+		/* libibverbs's fork support does not work for huge page.
+		 * libibverbs does have huge page fork support, which can be activated by
+		 * setting environment RDMAV_HUGEPAGE_SAFE to 1. However, huge page fork
+		 * support is extremely expensive, enabling it would defeat the purpose
+		 * of using huge page. Therefore, we simply disable huge page usage
+		 * when libfabric's fork support is enabled.
+		 */
+		if (efa_env.huge_page_setting == EFA_ENV_HUGE_PAGE_ENABLED) {
+			fprintf(stderr,
+				"The envrionment variable FI_EFA_USE_HUGE_PAGE has been\n"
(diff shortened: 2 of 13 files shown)


prov/efa/src/efa_fork_support.c
```diff
@@ -57,8 +57,17 @@ void efa_fork_support_request_initialize()
...
+	if (fork_support_requested || getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE")) {
 		g_efa_fork_status = EFA_FORK_SUPPORT_ON;
+		/* libibvers's fork support does not work for huge page.
+		 * libibvers does have huge page fork support, which can be activated by
+		 * setting environment RDMAV_HUGEPAGE_SAFE to 1. However, huge page fork
+		 * support is extremely expensive, enabling it would defeat the purpose
+		 * of using huge page. Therefore, we simply disable huge page usage
+		 * when libfabric's fork support is enabled.
+		 */
+		efa_env.use_huge_page = false;
```
wenduwan, line 69: I'm not sure about this coupling. The logic would be simpler if we always respect the user environment variable even if that could cause issues, e.g. user wants to use huge page and also asks for fork support - we just print a warning.

Accordingly we can advise users to disable hugepage explicitly.

wzamazon: I think the point to not overwrite user setting is fair.

I will add additional variable to record that  user set `FI_EFA_HUGE_PAGE`, but we will have to abort() here, because of the risk of silent data corruption.

shijin-aws: We should add a check here, if FI_EFA_USE_HUGE_PAGE is explicitly set, we should warn

wenduwan: > we will have to abort() here, because of the risk of silent data corruption.

Whether data corruption occurs also depends on how the application manages memory, so technically it could/may be ok to proceed.

But I'm ok with abort given the very bad worst case. It also makes testing easier, i.e. catch exception.

wzamazon: >Whether data corruption occurs also depends on how the application manages memory, so technically it could/may be ok to proceed.

the corruption if happens will be silent, so user might not know it happened.

wzamazon: added the abort in new revision, also added a test to verify

prov/efa/src/efa_fork_support.c
```diff
@@ -57,8 +57,33 @@ void efa_fork_support_request_initialize()
 	 * libibverbs' fork support are set. These variables are
 	 * defined by ibv_fork_init(3).
 	 */
-	if (fork_support_requested || getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE"))
+	if (fork_support_requested || getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE")) {
 		g_efa_fork_status = EFA_FORK_SUPPORT_ON;
+		/* libibvers's fork support does not work for huge page.
+		 * libibvers does have huge page fork support, which can be activated by
```
shijin-aws, line 63: nit: `libibverbs`

wzamazon: Thank you! fixed

shijin-aws (approved):

shijin-aws:
So, for newer kernel >=5.15, users still need to set `FI_EFA_FORK_SAFE=1` to use fork support safely?

wzamazon:
>So, for newer kernel >=5.15, users still need to set FI_EFA_FORK_SAFE=1 to use fork support safely?

No. On kernel 5.15, they should never set FI_EFA_FORK_SAFE.

They might want to set FI_EFA_USE_HUGE_PAGE to 0 to disable huge page usage.

shijin-aws:
> They might want to set FI_EFA_USE_HUGE_PAGE to 0 to disable huge page usage.

Hmmm, do we need to warn it if they are not set ?

shijin-aws:
oops, clicked close by accident, re-opened. Sorry

wzamazon:
>Hmmm, do we need to warn it if they are not set ?

I think this is better to be added to the public doc. Most of application will not be impacted.

For the few applications that use fork() extensively, we educate them through doc.

prov/efa/test/efa_unit_test_mocks.h
```diff
@@ -132,4 +134,10 @@ void *__real_neuron_alloc(void **handle, size_t size);
 void *efa_mock_neuron_alloc_return_null(void **handle, size_t size);
 #endif
 
+enum ibv_fork_status __real_ibv_is_fork_initialized(void);
+
+enum ibv_fork_status efa_mock_ibv_is_fork_initialize_return_unneeded(void);
+
+enum ibv_fork_status efa_mock_ibv_is_fork_initialize_return_disabled(void);
```
wenduwan, line 141: typo, should be `initialized`

prov/efa/test/efa_unit_test_fork_support.c
```diff
@@ -0,0 +1,33 @@
+#include "efa_unit_tests.h"
+
+/**
+ * @brief verify efa_fork_support_request_initialize() set value of g_efa_fork_status and efa_env.use_huge_page correctly
+ */
+void test_efa_fork_support_request_initialize_when_ibv_fork_support_is_needed(void **state)
+{
+	setenv("FI_EFA_FORK_SAFE", "1", true);
+	g_efa_unit_test_mocks.ibv_is_fork_initialized = &efa_mock_ibv_is_fork_initialize_return_disabled;
```
wenduwan, line 9: Curious why we didn't use a single mock function and utilize `will_return` to change the return value?

prov/efa/src/efa_env.c
```diff
@@ -230,6 +237,8 @@ void efa_env_define()
 			"The maximum number of bytes that will be eagerly sent by inflight messages uses runting read message protocol (Default 307200).");
 	fi_param_define(&efa_prov, "use_sm2", FI_PARAM_BOOL,
 			"Use the experimental shared memory provider SM2 for intra node communication.");
+	fi_param_define(&efa_prov, "use_huge_page", FI_PARAM_BOOL,
+			"Whether EFA provider can use huge page memory for internal buffer. Using huge page memory has a small performance advantage, but can cause system to run out of huge page memory. By default: EFA provider will use huge page unless FI_EFA_FORK_SAFE is set to 1/on/true.");
```
wenduwan, line 241: Max line length.

wzamazon: Fixed

prov/efa/test/efa_unit_test_fork_support.c
wzamazon: Should have used `will_return` and `mock`. Latest version did that

prov/efa/test/efa_unit_test_mocks.h
wzamazon: fixed

prov/efa/src/efa_env.c
wenduwan: Oh did you push this commit? It's still showing a single long line.

wenduwan (approved):
One pending minor comment. Otherwise looking good. Thank you!

prov/efa/src/efa_env.c
wzamazon: Ah. I update the message in `efa_fork_support_initialize()`. Did not update here.

New revision updated here.

zachdworkin:
@wzamazon please rebase to pick up recent intel ci changes

PR #9121, opened 2023-07-07
prov/sm2: Introduce proto_flags

sunkuamzn:
CMA host to device, SAR and RMA protocols in SM2 all need additional flags to handle different messages internal to the protocol.

One approach is to use a different protocol for each message type but this is cumbersome. Instead, we can set flags in the xfer entry that indicate the message type.

To avoid alignment issues and keep the header size the same, we use 16 bits for the protocol and 16 bits for protocol flags.

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -107,6 +107,7 @@ enum {
  * 	op_flags - flags associated with op,
  * 		   NOTE: Only grabbing the bottom 32 bits
  * 	proto - sm2 operation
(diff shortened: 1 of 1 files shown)


sunkuamzn:
@aingerson @a-szegel we need proto_flags irrespective of whether host to device uses them because (1) RMA uses of flags (2) we want to set a "don't generate send/receive completion" flag to stop abusing `FI_DELIVERY_COMPLETE`

aingerson (approved):

PR #9122, opened 2023-07-07
prov/efa: Change FI_EFA_USE_SM2 to FI_EFA_INTRANODE_PROVIDER

a-szegel:
Change the env variable that controls EFA's intranode provider to be more generic.  Switch variable from a bool to a string.

### DIFF
--- a/prov/efa/src/efa_env.c
+++ b/prov/efa/src/efa_env.c
@@ -67,8 +67,8 @@ struct efa_env efa_env = {
 	.efa_write_segment_size = 1073741824, /* need to confirm this constant. */
 	.rnr_retry = 3, /* Setting this value to EFA_RNR_INFINITE_RETRY makes the firmware retry indefinitey */
 	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
-	.use_sm2 = false,
 	.huge_page_setting = EFA_ENV_HUGE_PAGE_UNSPEC,
+	.intranode_provider = "shm",
 };
 
 /**
@@ -127,7 +127,6 @@ void efa_env_param_get(void)
 	}
 
 	fi_param_get_int(&efa_prov, "tx_queue_size", &efa_env.tx_queue_size);
-	fi_param_get_int(&efa_prov, "enable_shm_transfer", &efa_env.enable_shm_transfer);
 	fi_param_get_int(&efa_prov, "use_zcpy_rx", &efa_env.use_zcpy_rx);
 	fi_param_get_int(&efa_prov, "set_cuda_sync_memops", &efa_env.set_cuda_sync_memops);
 	fi_param_get_int(&efa_prov, "zcpy_rx_seed", &efa_env.zcpy_rx_seed);
@@ -164,11 +163,28 @@ void efa_env_param_get(void)
 			    &efa_env.efa_read_segment_size);
 	fi_param_get_size_t(&efa_prov, "inter_max_gdrcopy_message_size",
 			    &efa_env.efa_max_gdrcopy_msg_size);
-	fi_param_get_bool(&efa_prov, "use_sm2", &efa_env.use_sm2);
 
 	int use_huge_page;
 	if (fi_param_get_bool(&efa_prov, "use_huge_page", &use_huge_page) ==0) {
 		efa_env.huge_page_setting = use_huge_page ? EFA_ENV_HUGE_PAGE_ENABLED : EFA_ENV_HUGE_PAGE_DISABLED;
+	}
+
+	fi_param_get_int(&efa_prov, "enable_shm_transfer", &efa_env.enable_shm_transfer);
+	if (efa_env.enable_shm_transfer == 0) {
+		efa_env.intranode_provider = "efa";
+		EFA_WARN(FI_LOG_CORE, "FI_EFA_ENABLE_SHM_TRANSFER is deprecated (and will be removed in a future release), "
+		         "use FI_EFA_INTRANODE_PROVIDER=efa to specify the EFA provider for intra-node communication.\n");
+	}
+
+	/* Setting FI_EFA_INTRANODE_PROVIDER will override FI_EFA_ENABLE_SHM_TRANSFER=0 */
+	fi_param_get_str(&efa_prov, "intranode_provider", &efa_env.intranode_provider);
+	if (strcmp(efa_env.intranode_provider, "efa") &&
+	    strcmp(efa_env.intranode_provider, "shm") &&
+	    strcmp(efa_env.intranode_provider, "sm2")) {
+		EFA_WARN(FI_LOG_CORE, "FI_EFA_INTRANODE_PROVIDER=%s, EFA supports 'shm', 'sm2' and 'efa'"
+			 " for intra-node communication. Unsupported provider name. Aborting...\n",
+			 efa_env.intranode_provider);
+		abort();
 	}
 
 	efa_fork_support_request_initialize();
@@ -181,8 +197,9 @@ void efa_env_define()
 			"Defines the minimum number of credits a sender requests from a receiver (Default: 32).");
 	fi_param_define(&efa_prov, "tx_queue_size", FI_PARAM_INT,
 			"Defines the maximum number of unacknowledged sends with the NIC.");
+	/* TODO Remove enable_shm_transfer on future release */
 	fi_param_define(&efa_prov, "enable_shm_transfer", FI_PARAM_INT,
-			"Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
+			"(Deprecated, use FI_EFA_INTRANODE_PROVIDER=efa to turn off SHM. Will remove in future release.) Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
 	fi_param_define(&efa_prov, "use_zcpy_rx", FI_PARAM_INT,
 			"Enables the use of application's receive buffers in place of bounce-buffers when feasible. (Default: 1)");
 	fi_param_define(&efa_prov, "set_cuda_sync_memops", FI_PARAM_INT,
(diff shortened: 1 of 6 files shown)


prov/efa/src/efa_env.c
```diff
@@ -67,7 +67,7 @@ struct efa_env efa_env = {
 	.efa_write_segment_size = 1073741824, /* need to confirm this constant. */
 	.rnr_retry = 3, /* Setting this value to EFA_RNR_INFINITE_RETRY makes the firmware retry indefinitey */
 	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
-	.use_sm2 = false,
+	.intranode_provider = "shm",
```
wenduwan, line 70: This change might make `FI_EFA_ENABLE_SHM_TRANSFER` ambiguous...

What about `FI_EFA_SHM_PROVIDER`?

a-szegel: It is the selection for a peer provider and it doesn't have to be SHM... anything that implements the peer API (i.e.. tcp).

wenduwan: Not sure about having 'peer' in the name either - peer api is an implementation detail to use another provider. 

My point is, if `FI_EFA_INTRANODE_PROVIDER=sm2`, then what should `FI_EFA_ENABLE_SHM_TRANSFER` control - does it control sm2?

wzamazon: It seems we can deprecate the `FI_EFA_ENANLE_SHM_TRANSFER`, and replace it with setting `FI_EFA_INTRANODE_PROVIDER` to `efa`

wzamazon: e.g.

FI_EFA_INTRANODE_PROVIDER = efa means use EFA device for intranode communication

shijin-aws: That seems reasonable to me. For the deprecation, can we also make `FI_EFA_ENANLE_SHM_TRANSFER=0` equivalent to `FI_EFA_INTRANODE_PROVIDER=efa` in the code? We need to update the doc, but I don't want to cause unexpected behaviors for customers who have been using this env

prov/efa/src/efa_env.c
```diff
@@ -228,8 +228,8 @@ void efa_env_define()
 			"Enables fork support and disables internal usage of huge pages. Has no effect on kernels which set copy-on-fork for registered pages, generally 5.13 and later. (Default: false)");
 	fi_param_define(&efa_prov, "runt_size", FI_PARAM_INT,
 			"The maximum number of bytes that will be eagerly sent by inflight messages uses runting read message protocol (Default 307200).");
-	fi_param_define(&efa_prov, "use_sm2", FI_PARAM_BOOL,
-			"Use the experimental shared memory provider SM2 for intra node communication.");
+	fi_param_define(&efa_prov, "intranode_provider", FI_PARAM_STRING,
+			"The name of the peer provider that EFA should offload intra-node communications to (Default shm).");
```
shijin-aws, line 232: you would have to add check here for allowed values. Currently only shm and sm2 are allowed.

prov/efa/src/efa_env.c
sunkuamzn: Doesn't the ofi-nccl plugin use FI_EFA_ENABLE_SHM=0? We need to get them to update.

zachdworkin:
@a-szegel please rebase to pick up recent intel ci changes

prov/efa/src/efa_env.c
```diff
@@ -228,8 +229,8 @@ void efa_env_define()
...
```
sunkuamzn, line 233: I agree with @wenduwan that we shouldn't use the word "peer"

>The name of the provider that EFA should offload intra-node communications to (Default shm).

sounds great to me.

prov/efa/src/efa_shm.c
```diff
@@ -98,11 +98,13 @@ void efa_shm_info_create(const struct fi_info *app_info, struct fi_info **shm_in
 	int ret;
 	struct fi_info *shm_hints;
 
-	char *shm_provider;
-	if (efa_env.use_sm2) {
-		shm_provider = "sm2";
-	} else {
-		shm_provider = "shm";
+	if (strcmp(efa_env.intranode_provider, "shm") &&
```
sunkuamzn, line 101: Check is case sensitive. Which might be OK but it contradicts your PR description.

a-szegel: I didn't have any checks before... those were recently added.

prov/efa/src/efa_env.c
a-szegel: done

prov/efa/src/efa_env.c
a-szegel: done

shijin-aws: > Doesn't the ofi-nccl plugin use FI_EFA_ENABLE_SHM=0? We need to get them to update.

It's in the doc, but without this env it should also work.

prov/efa/src/efa_env.c
```diff
@@ -163,7 +162,14 @@ void efa_env_param_get(void)
 			    &efa_env.efa_read_segment_size);
 	fi_param_get_size_t(&efa_prov, "inter_max_gdrcopy_message_size",
 			    &efa_env.efa_max_gdrcopy_msg_size);
-	fi_param_get_bool(&efa_prov, "use_sm2", &efa_env.use_sm2);
+	fi_param_get_str(&efa_prov, "intranode_provider", &efa_env.intranode_provider);
+	fi_param_get_int(&efa_prov, "enable_shm_transfer", &efa_env.enable_shm_transfer);
```
sunkuamzn, line 166: If both `intranode_provider` and `enable_shm_transfer` are set, then `enable_shm_transfer` will override `intranode_provider` just because it's parsed later. I think that's OK until we deprecate `enable_shm_transfer`.

prov/efa/src/efa_env.c
```diff
@@ -163,7 +162,14 @@ void efa_env_param_get(void)
...
+	if (efa_env.enable_shm_transfer == 0) {
+		efa_env.intranode_provider = "efa";
+		EFA_WARN(FI_LOG_CORE, "FI_EFA_ENABLE_SHM_PROVIDER is deprecated (and will be removed in a future release), "
```
shijin-aws, line 169: `FI_EFA_ENABLE_SHM_PROVIDER` -> `FI_EFA_ENABLE_SHM_TRANSFER`

sunkuamzn (approved):

prov/efa/src/efa_shm.c
shijin-aws, line 101: Shouldn't we move this check to efa_env.c?  If efa_env.intranode_provider is none of `efa`, `shm`, `sm2` we should raise an error.

prov/efa/src/efa_env.c
shijin-aws: I think fi_param_get_str(&efa_prov, "intranode_provider", &efa_env.intranode_provider); whould be moved to after L172, 

Also, if `efa_env.intranode_provider` is none of `efa`, `shm`, sm2`, we should fail

prov/efa/src/efa_env.c
a-szegel: We are deprecating enable_shm_transfer now it will still override the behavior until we remove it from the code, and then it will do nothing.

prov/efa/src/efa_shm.c
a-szegel: done.

prov/efa/src/efa_env.c
a-szegel: BY fail... we shouldn't abort, we should warn and run with EFA... we decided before that we want this behavior.

prov/efa/src/efa_env.c
a-szegel: This change will make NCCL initialization faster b/c if we turn off SHM, we no longer initialize it now.

prov/efa/src/efa_env.c
```diff
@@ -163,7 +162,25 @@ void efa_env_param_get(void)
...
+		EFA_WARN(FI_LOG_CORE, "FI_EFA_ENABLE_SHM_TRANSFER is deprecated (and will be removed in a future release), "
+		         "use FI_EFA_INTRANODE_PROVIDER=efa to specify the EFA provider for intra-node communication.\n");
+	}
+
+	/* Setting FI_EFA_INTRANODE_PROVIDER will override FI_EFA_ENABLE_SHM_TRANSFER=0 */
+	fi_param_get_str(&efa_prov, "intranode_provider", &efa_env.intranode_provider);
+	if (strcmp(efa_env.intranode_provider, "efa") &&
+	    strcmp(efa_env.intranode_provider, "shm") &&
+	    strcmp(efa_env.intranode_provider, "sm2")) {
+		EFA_WARN(FI_LOG_CORE, "FI_EFA_INTRANODE_PROVIDER=%s, EFA supports 'shm' and 'sm2'"
```
shijin-aws, line 178: nit: 'efa' is also supported

shijin-aws (approved):
looks cleaner, a minor comment

prov/efa/src/efa_env.c
a-szegel: done

shijin-aws:
@wzamazon @wenduwan more comments?

prov/efa/src/efa_env.c
wzamazon: NCCL disable shm by passing FI_REMOTE_COMM but not not FI_LOCAL_COMM. It did not use the env.

prov/efa/src/efa_env.c
```diff
@@ -174,8 +191,9 @@ void efa_env_define()
 			"Defines the minimum number of credits a sender requests from a receiver (Default: 32).");
 	fi_param_define(&efa_prov, "tx_queue_size", FI_PARAM_INT,
 			"Defines the maximum number of unacknowledged sends with the NIC.");
+	/* TODO Remove enable_shm_transfer on future release */
 	fi_param_define(&efa_prov, "enable_shm_transfer", FI_PARAM_INT,
-			"Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
+			"(Deprecated, use FI_EFA_INTRANODE_PROVIDER=efa to turn off SHM. Will remove in future release.) Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
```
wzamazon, line 196: Nit: "will remove" -> "will be removed"

wzamazon (approved):
LGTM, 1 minor comment

wzamazon:
seems there is some conflict need to be resolved.

PR #9124, opened 2023-07-07
fabtests/common: separate RX and RMA counters

aingerson:
this fixes running fi_rma_bw -t counter which previously allowed one side to exit while the other was still in the middle of the test

**I did not update the deferred work test because it was convoluted and I think already broken and nobody supports deferred work except the sockets provider... but I can update/rewrite the test it if someone feels strongly about it

### DIFF
--- a/fabtests/functional/rdm_rma_trigger.c
+++ b/fabtests/functional/rdm_rma_trigger.c
@@ -62,7 +62,7 @@ static int rma_write_trigger(void *src, size_t size,
 static int run_test(void)
 {
 	int ret = 0;
-	uint64_t start_tx, start_rx;
+	uint64_t start_tx;
 
 	ret = ft_init_fabric();
 	if (ret)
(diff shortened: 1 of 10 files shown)


aingerson:
@shijin-aws Made some more updates. Let me know if this fixes your issue!

shijin-aws:
@aingerson Thanks, I can confirm this patch fixed the fi_rma_bw -t counter error I hit.

shijin-aws:
On a side note, when I run fi_rdm_atomic with `-t counter` I hit a hang as well. I suspect it may have the same issue as this PR is fixing. I will dig it more.

aingerson:
@shijin-aws Thanks again for catching! Updated to fix tx counter with RMA which was also incorrect.

fabtests/functional/rdm_rma_trigger.c
```diff
@@ -109,7 +108,7 @@ static int run_test(void)
 	} else {
 		/* The value of the rx counter should have increased by 2
 		 * for both operations (write and triggered) */
-		ret = fi_cntr_wait(rxcntr, start_rx + 2, -1);
+		ret = fi_cntr_wait(rma_cntr, 2, -1);
```
shefty, line 111: please update comment

aingerson:
Updated to fix comment and CI failures for the rma event test incorrectly reading the rma counter instead of the tx counter

PR #9126, opened 2023-07-07
fabtests/efa: Add new test cases to unexpected message test

sunkuamzn:
Added a dependency on the message_size and memory_type fixtures
    
We use different protocols for different message sizes, so it's good to test the unexpected path for all message sizes
    
Also good to test unexpected path with HMEM buffers

### DIFF
--- a/fabtests/pytest/efa/test_unexpected_msg.py
+++ b/fabtests/pytest/efa/test_unexpected_msg.py
@@ -1,7 +1,19 @@
 import pytest
+from efa.efa_common import efa_run_client_server_test
 
+@pytest.mark.parametrize("iteration_type",
+                         [pytest.param("short", marks=pytest.mark.short),
+                          pytest.param("standard", marks=pytest.mark.standard)])
 @pytest.mark.functional
-def test_unexpected_msg(cmdline_args):
-    from common import ClientServerTest
-    test = ClientServerTest(cmdline_args, "fi_unexpected_msg -e rdm -I 10")
-    test.run()
+def test_unexpected_msg(cmdline_args, iteration_type, completion_semantic, memory_type, message_size):
+    if completion_semantic == "delivery_complete":
+        pytest.skip("Unexpected message test does not support delivery_complete completion semantic")
+
+    # SHM provider currently supports 1 in flight message for host<->device transfers
+    if memory_type in ["cuda_to_host", "host_to_cuda"]:
+        inflight_msgs = 1
+    else:
+        inflight_msgs = 4
+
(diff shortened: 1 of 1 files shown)


fabtests/pytest/efa/test_unexpected_msg.py
```diff
@@ -1,7 +1,13 @@
...
 @pytest.mark.functional
-def test_unexpected_msg(cmdline_args):
-    from common import ClientServerTest
-    test = ClientServerTest(cmdline_args, "fi_unexpected_msg -e rdm -I 10")
-    test.run()
+def test_unexpected_msg(cmdline_args, completion_semantic, memory_type, message_size):
+    if completion_semantic == "delivery_complete":
+        pytest.skip("Unexpected message test does not support delivery_complete completion semantic")
+
+    efa_run_client_server_test(cmdline_args, "fi_unexpected_msg", iteration_type,
```
shijin-aws, line 12: why turn off data check?

sunkuamzn: Test failed with unrecognized option when I passed in `-v`, so I thought the test doesn't support data verification. But it actually does. I'm having other issues with the HMEM buffer though. So making this PR a draft for now.

sunkuamzn: Data check is back

fabtests/pytest/efa/test_unexpected_msg.py
```diff
@@ -1,7 +1,19 @@
...
-    test.run()
+def test_unexpected_msg(cmdline_args, iteration_type, completion_semantic, memory_type, message_size):
+    if completion_semantic == "delivery_complete":
+        pytest.skip("Unexpected message test does not support delivery_complete completion semantic")
+
+    # SHM provider currently supports 1 in flight message for host<->device transfers
+    if memory_type in ["cuda_to_host", "host_to_cuda"]:
+        inflight_msgs = 1
+    else:
+        inflight_msgs = 4
```
shijin-aws, line 16: won't this fail for platform without CMA? Or you'd like to skip the test in the caller of fabtests

sunkuamzn: You're right. Can't think of a good way to skip this just for CMA disabled. I think it's best to wait until we fix SHM to merge this.

shijin-aws:
offline talked with @sunkuamzn, I will take over this PR and get a newer version after https://github.com/ofiwg/libfabric/pull/9187 is merged

PR #9130, opened 2023-07-10
prov/sm2: Cleanup warnings on startup

a-szegel:
Modify SM2 to only produce 1 warning per endpoint on successful startup (that SM2 is being used + the endpoints allocation).

### DIFF
--- a/prov/sm2/src/sm2_init.c
+++ b/prov/sm2/src/sm2_init.c
@@ -67,7 +67,7 @@ int sm2_create(const struct fi_provider *prov, const struct sm2_attr *attr,
 
 	sm2_calculate_size_offsets(&recv_queue_offset, &freestack_offset);
 
-	FI_WARN(prov, FI_LOG_EP_CTRL, "Claiming an entry for (%s)\n",
+	FI_INFO(prov, FI_LOG_EP_CTRL, "Claiming an entry for (%s)\n",
 		attr->name);
 	sm2_file_lock(sm2_mmap);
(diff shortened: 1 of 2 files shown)


sunkuamzn (approved):

prov/sm2/src/sm2_init.c
```diff
@@ -124,6 +124,9 @@ int sm2_create(const struct fi_provider *prov, const struct sm2_attr *attr,
 	/* Need to unlock coordinator so that others can add themselves to
 	 * header */
 	sm2_file_unlock(sm2_mmap);
+
+	FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
```
aingerson, line 128: It feels like this should also be an info log, not really  a warning.

aingerson (commented):
Small comment, otherwise looks fine. Up to you if you want to address it or not.

prov/sm2/src/sm2_init.c
a-szegel: I like it as a warn so you can see that you are using SM2 from FI_WARN.... EFA has a very similar warning.  Via FI_WARN, with EFA + SM2, you can see that both are being used without going to INFO.

aingerson: ok

a-szegel: But I do hear you that we should treat FI_WARN like an error log to make it as little noisy as possible.

a-szegel: Going to keep this is warn for now and maybe move to info in the future.

PR #9131, opened 2023-07-10
prov/mrail: Turn Noisy WARN to INFO

a-szegel:
Currently, "OFI_MRAIL_ADDR_STRC env variable not set!" gets logged at warning on every run, even if we aren't using the mrail provider. Change this to be an info to clean up the warning log.


wzamazon (approved):

shefty (changes requested):
please remove level stack variable, which is no longer used

aingerson (approved):

a-szegel:
@shefty Are you good with this?

a-szegel:
bot:aws:retest

PR #9132, opened 2023-07-11
fabtests: Allow to run with non IP providers

traxys:
Some providers do not work with IP addresses at all, as such fabtests are not really compatible. This adds two changes to make it possible to use such providers easily:
- Supporting fi_addr_str as an address format, as it enables any kind of address format
- Setting the out-of-band address differently than the node address (as a fi_addr_str address may not be an IP)


sunkuamzn:
bot:aws:retest

shefty:
The OOB address exchange should support non-IP based addressing.  We just require that the system support tcp/ip sockets, but that's not a requirement on the provider.

traxys:
Well the fabtests do require using a socket for OOB, and I think allowing to specify an IP network is the easiest (I feel like most networks that are not IP based have some kind of bridge to expose IP to high level applications, albeit maybe slower than the native path, but performance is not a problem here).

shefty (approved):
Please update the fabtests man page with the -O option details.

traxys:
I added the option to the man page & re-generated the nroff page (using the same pandoc version as the previous person to minimize breakage)

fabtests/man/man7/fabtests.7
```diff
@@ -1,6 +1,6 @@
 .\" Automatically generated by Pandoc 1.19.2.4
 .\"
-.TH "fabtests" "7" "2020\-12\-01" "Libfabric Programmer\[aq]s Manual" "\@VERSION\@"
+.TH "fabtests" "7" "2023\-07\-24" "Libfabric Programmer\[aq]s Manual" "#VERSION#"
```
shefty, line 3: There's a missing github action somewhere that should be updating this file automatically, similar to the libfabric man pages.

fabtests/functional/av_xfer.c
```diff
@@ -237,6 +237,7 @@ int main(int argc, char **argv)
 		      FI_TAGGED : FI_MSG;
 	hints->mode = FI_CONTEXT;
 	hints->domain_attr->mr_mode = opts.mr_mode;
+	hints->addr_format = opts.address_format;
```
shefty, line 240: The above 2 lines are either in every fabtest, or nearly every test.  We need a more maintainable way of handling this.  But I'm okay dealing with that later.

shefty (approved):

shefty:
ucx failure:
```
02:01:31  - name:   fi_rdm_multi_client -C 10 -I 5 -p "ucx"
02:01:31    timestamp: 20230724-085943+0000
02:01:31    result: Fail
02:01:31    time:   1
02:01:31    server_cmd:  /home/cstbuild/cloudbees_job_install/OFIWG_Libfabric/ofi_libfabric/PR-9132/3/reg/ucx/bin/fi_rdm_multi_client -C 10 -I 5 -p "ucx"   -E
02:01:31    server_stdout: |
02:01:31    client_cmd:  /home/cstbuild/cloudbees_job_install/OFIWG_Libfabric/ofi_libfabric/PR-9132/3/reg/ucx/bin/fi_rdm_multi_client -C 10 -I 5 -p "ucx"   -E cb23-ib0
02:01:31    client_stdout: |
02:01:31      Starting client: 0
02:01:31      connect: Invalid argument
02:01:31      ft_init_oob(): functional/rdm_multi_client.c:124, ret=1 (Operation not permitted)
02:01:31      run_client(): functional/rdm_multi_client.c:228, ret=1 (Operation not permitted)
```
Restarted.  I don't know if the failures are related or not.  There are a few more failures after this one, but I think are related to this test failing.

traxys:
It seems strange, if the `-O` option is not passed nothing should really change (I think). I'm not really sure how we could get to the `connect` without an error, but fail to `connect` if the remote is there

fabtests/common/shared.c
```diff
@@ -3197,6 +3205,9 @@ void ft_parse_addr_opts(int op, char *optarg, struct ft_opts *opts)
 	case 'C':
 		opts->options |= FT_OPT_SERVER_PERSIST;
 		opts->num_connections = atoi(optarg);
+	case 'O':
```
j-xiong, line 3208: Need a `break;` before this line.

j-xiong (commented):
This is the cause of the ucx test failure.

fabtests/common/shared.c
traxys: sorry....

shefty:
TCP failure:
```
fi_rdm_tagged_bw -v -U -p "tcp"
12:45:18    timestamp: 20230725-194243+0000
12:45:18    result: Fail
12:45:18    time:   2
```
which has been resolved, so the branch has likely not been recently rebased.

PR #9135, opened 2023-07-12
prov/sm2: Introduce new proto_flags

sunkuamzn:
Using `SM2_RETURN` flag instead of `sm2_proto_return` preserves the original protocol in the xfer_entry.
Meaning we can control when to generate completions without explicitly modifying `FI_DELIVERY_COMPLETE`

`SM2_UNEXP` flag specifies that the xfer entry was immediately returned because a corresponding buffer was not posted by the receiver. So no send completion is generated.

The `SM2_UNEXP` flag also replaces the `return_xfer_entry` bool in `sm2_start_common`

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -91,9 +91,10 @@ extern pthread_mutex_t sm2_ep_list_lock;
 
 enum {
 	sm2_proto_inject,
-	sm2_proto_return,
 	sm2_proto_max,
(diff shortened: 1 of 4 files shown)


sunkuamzn:
This PR will not pass CI because it is missing logic for dealing with unexpected messages when delivery complete is requested. Closing in favor of https://github.com/ofiwg/libfabric/pull/9139

PR #9136, opened 2023-07-12
prov/tcp: fix missing iov truncation on saved message path

### LINKED ISSUE #9099: prov/tcp: Frequent CI failures with fi_rdm_tagged_bw
Running test with -v -U options, frequently generates an error along the lines of "truncation error".  It's not clear if this is a problem with the provider or the test itself lacking some sort of synchronization.  Attempts to reproduce manually have failed so far.

aingerson:
Commit 489d7804cf78fe2cd6fe119aff595df5b8b8659c added truncation handling in the saved message path but removed the iov truncation of the rx entry iov. This caused erroneous truncation errors because too much data would be read off the socket, causing the next send to report a truncation.

Fixes #9099

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -272,6 +272,9 @@ void xnet_recv_saved(struct xnet_xfer_entry *saved_entry,
 		done_len = msg_len - ep->cur_rx.data_left;
 		assert(msg_len && ep->cur_rx.data_left);
 
+		(void) ofi_truncate_iov(&saved_entry->iov[0],
+				        &saved_entry->iov_cnt, msg_len);
+
 		copy_len = ofi_copy_iov_buf(saved_entry->iov,
 					    saved_entry->iov_cnt, 0, msg_data,
 					    done_len, OFI_COPY_BUF_TO_IOV);


shefty:
Thanks for the fix.  I need to look at the underlying issue in more detail to determine the scope of the problem and how much damage it causes MPI.

shefty:
It took me way too long to figure out why I was even truncating the iov.  The reason is that the iov is passed into the socket recv() call.   So without truncation, yes, too much data can be read from the socket.  The impact here isn't that the next message is truncated, it's that the tcp stream is now 'corrupt' relative to the receiver.  It's reading random data from somewhere in the stream and interpreting it as a message.  Can you please update the commit message with something along these lines?

The scope is limited to the following condition: The user must post a receive buffer for a messages that is actively being saved off the tcp stream, and the posted buffer must be larger than the message being received.  This might be a fairly narrow window, or even an unlikely one, but I don't really know.  So far, MPI testing has been passing.  Based on traced I've looked at, MPI usually posts receives that match the size of the send.

shefty (approved):

shefty:
Please pick for v1.18.x

aingerson:
@shefty Better?

PR #9137, opened 2023-07-12
fabtests/benchmarks/bw: fix bandwidth test buffer utilization

aingerson:


j-xiong (commented):
The naming of functions is getting confusing. We probably should mirror the tx functions and have a new ft_post_inject_buf() function which can be called directly or by the original ft_post_inject() function.

aingerson:
@j-xiong Yeah it's super confusing I agree. It's weird comparing the normal send path with the inject path too though because the comparable ft_tx to the ft_inject call also waits for a completion which we have no equivalent to the inject version so really it would make more sense to rename ft_inject to ft_post_inject and the new call to ft_post_inject_buf. Is that what you're suggesting?

j-xiong:
I am suggesting keep the original fi_inject(), rename your modified fi_post_inject() to fi_post_inject_buf(), and add a new fi_post_inject() that calls fi_post_inject_buf() with tx_buf and tx_seq.

aingerson:
@j-xiong Cool, just wanted to make sure I understood. Updated!

j-xiong (approved):
Looks good.

PR #9138, opened 2023-07-13
fabtests: Add barrier before starting test

sunkuamzn:
The messages from the address exchange can overlap with the test messages and cause spurious failures. Add a barrier to prevent that.

### DIFF
--- a/fabtests/functional/unexpected_msg.c
+++ b/fabtests/functional/unexpected_msg.c
@@ -319,6 +319,7 @@ static int run_test(void)
 			return ret;
 	}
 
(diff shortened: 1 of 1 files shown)


aingerson:
Trying to understand the failures that get thrown. The end of the address exchange has a pseudo-barrier by doing a send/receive with the inserted address. Can you explain the failure case you're trying to avoid with this change?

sunkuamzn:
I think it was actually a bug in SM2. I don't see that failure anymore, so it's likely not necessary to have this barrier. Closing.

PR #9139, opened 2023-07-13
prov/sm2: Implement CMA protocol and HMEM

sunkuamzn:
This PR implements (again) the CMA protocol for arbitrary sized host-host message transfers. It also implements the IPC protocol for arbitrary sized device->host/device transfers and a IPC fallback protocol for host->device transfers.

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -99,6 +99,12 @@ enum {
 
 #define SM2_RMA_REQ (1 << 1)
 
+/* The SM2_UNEXP flag is used for unexpected receives. On the receiver, it
+indicates that the xfer_entry that the receiver is processing is actually in the
+xfer_ctx_pool and not in the receiver's freestack.
+*/
+#define SM2_UNEXP (1 << 2)
+
 /*
  * 	next - fifo linked list next ptr
  * 		This is volatile for a reason, many things touch this
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -105,6 +105,12 @@ xfer_ctx_pool and not in the receiver's freestack.
 */
 #define SM2_UNEXP (1 << 2)
 
+/* SM2_GENERATE_COMPLETION is used in protocols that require delivery complete
+ * semantics (CMA and IPC). It is set when the receiver has finished processing
+ * the request and wants to tell the sender that the sender should generate a
+ * send completion. */
+#define SM2_GENERATE_COMPLETION (1 << 3)
+
 /*
  * 	next - fifo linked list next ptr
  * 		This is volatile for a reason, many things touch this
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -91,11 +91,12 @@ extern pthread_mutex_t sm2_ep_list_lock;
 
 enum {
 	sm2_proto_inject,
-	sm2_proto_return,
 	sm2_proto_max,
 };
 
 /* Protocol flags */
+#define SM2_RETURN (1 << 0)
+
 #define SM2_RMA_REQ (1 << 1)
 
 /*
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -95,6 +95,9 @@ enum {
 	sm2_proto_max,
 };
 
+/* Protocol flags */
+#define SM2_RMA_REQ (1 << 1)
+
 /*
  * 	next - fifo linked list next ptr
  * 		This is volatile for a reason, many things touch this
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -124,6 +124,10 @@ completions instead of waiting for the application to post the receive buffer.
  * send completion. */
 #define SM2_GENERATE_COMPLETION (1 << 3)
 
+/* Protocol flags for SM2 CMA protocol */
+#define SM2_CMA_HOST_TO_DEV	(1 << 4)
+#define SM2_CMA_HOST_TO_DEV_ACK (1 << 5)
+
 /*
  * 	next - fifo linked list next ptr
  * 		This is volatile for a reason, many things touch this
@@ -186,6 +190,10 @@ struct sm2_atomic_entry {
 struct sm2_cma_data {
 	size_t iov_count;
 	struct iovec iov[SM2_IOV_LIMIT];
+
+	/* Used for IPC host to device protocol */
+	struct ipc_info ipc_info;
+	struct fi_peer_rx_entry *rx_entry;
 };
 
 struct sm2_ep_name {
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -92,6 +92,7 @@ extern pthread_mutex_t sm2_ep_list_lock;
 enum {
 	sm2_proto_inject,
 	sm2_proto_cma,
+	sm2_proto_ipc,
 	sm2_proto_max,
 };
 
@@ -244,6 +245,7 @@ struct sm2_xfer_ctx {
 
 struct sm2_domain {
 	struct util_domain util_domain;
+	struct ofi_mr_cache *ipc_cache;
 	struct fid_peer_srx *srx;
 };
 
@@ -330,6 +332,7 @@ static inline bool sm2_proto_imm_send_comp(uint16_t proto)
 {
 	switch (proto) {
 	case sm2_proto_cma:
+	case sm2_proto_ipc:
 		return false;
 	default:
 		return true;
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -91,6 +91,7 @@ extern pthread_mutex_t sm2_ep_list_lock;
 
 enum {
 	sm2_proto_inject,
+	sm2_proto_cma,
 	sm2_proto_max,
 };
 
@@ -102,6 +103,17 @@ enum {
 /* The SM2_UNEXP flag is used for unexpected receives. On the receiver, it
 indicates that the xfer_entry that the receiver is processing is actually in the
 xfer_ctx_pool and not in the receiver's freestack.
+
+For protocols that require delivery complete semantics (CMA and IPC), the
+receiver pops an xfer_entry from its own freestack, sets SM2_UNEXP flag and uses
+the new xfer_entry to process the receive. Once the receive is processed, the
+receiver sets the SM2_RETURN and SM2_GENERATE_COMPLETION and sends it to the
+sender. The sender will generate the completion entry when it receives such an
+xfer_entry.
+
+TODO: Copy the data to a temporary buffer so that the sender is agnostic to
+unexpected receives. That also allows the sender to immediately generate
+completions instead of waiting for the application to post the receive buffer.
 */
 #define SM2_UNEXP (1 << 2)
 
@@ -125,7 +137,8 @@ xfer_ctx_pool and not in the receiver's freestack.
  * 	proto - sm2 operation
  * 	proto_flags - Flags used by the sm2 protocol
  * 	sender_gid - id of msg sender
- * 	user_data - the message
+ * 	user_data - Protocol dependent data. For inject, it's the message.
+ * 				For CMA protocol, it's struct sm2_cma_data.
  */
 struct sm2_xfer_hdr {
 	volatile long int next;
@@ -167,6 +180,11 @@ struct sm2_atomic_data {
 struct sm2_atomic_entry {
 	struct sm2_atomic_hdr atomic_hdr;
 	struct sm2_atomic_data atomic_data;
+};
+
+struct sm2_cma_data {
+	size_t iov_count;
+	struct iovec iov[SM2_IOV_LIMIT];
 };
 
 struct sm2_ep_name {
@@ -306,6 +324,16 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
 
 	*xfer_entry = smr_freestack_pop(sm2_freestack(ep->self_region));
 	return FI_SUCCESS;
+}
+
+static inline bool sm2_proto_imm_send_comp(uint16_t proto)
+{
+	switch (proto) {
+	case sm2_proto_cma:
+		return false;
+	default:
+		return true;
+	}
 }
 
 int sm2_query_atomic(struct fid_domain *domain, enum fi_datatype datatype,
--- a/prov/sm2/src/sm2_atomic.c
+++ b/prov/sm2/src/sm2_atomic.c
@@ -109,9 +109,9 @@ sm2_do_atomic_inject(struct sm2_ep *ep, int64_t peer_gid, uint32_t op,
 		return ret;
 
 	xfer_entry->hdr.proto = sm2_proto_inject;
-	xfer_entry->hdr.proto_flags |= proto_flags;
 
 	sm2_generic_format(xfer_entry, ep->gid, op, 0, 0, op_flags, context);
+	xfer_entry->hdr.proto_flags |= proto_flags;
 	sm2_atomic_format(xfer_entry, datatype, atomic_op, rma_ioc,
 			  rma_ioc_count, result_iov, result_count, iov,
 			  iov_count, compare_iov, compare_count);
--- a/prov/sm2/src/sm2_atomic.c
+++ b/prov/sm2/src/sm2_atomic.c
@@ -99,7 +99,7 @@ sm2_do_atomic_inject(struct sm2_ep *ep, int64_t peer_gid, uint32_t op,
 		     const struct fi_rma_ioc *rma_ioc, size_t rma_ioc_count,
 		     const struct iovec *compare_iov, size_t compare_count,
 		     const struct iovec *result_iov, size_t result_count,
-		     size_t total_len, void *context)
+		     size_t total_len, void *context, uint16_t proto_flags)
 {
 	struct sm2_xfer_entry *xfer_entry;
 	size_t ret;
@@ -109,6 +109,7 @@ sm2_do_atomic_inject(struct sm2_ep *ep, int64_t peer_gid, uint32_t op,
 		return ret;
 
 	xfer_entry->hdr.proto = sm2_proto_inject;
+	xfer_entry->hdr.proto_flags |= proto_flags;
 
 	sm2_generic_format(xfer_entry, ep->gid, op, 0, 0, op_flags, context);
 	sm2_atomic_format(xfer_entry, datatype, atomic_op, rma_ioc,
@@ -132,6 +133,7 @@ static inline ssize_t sm2_generic_atomic(
 	struct iovec result_iov[SM2_IOV_LIMIT];
 	sm2_gid_t peer_gid;
 	ssize_t ret = 0;
+	uint16_t proto_flags = 0;
 	size_t total_len;
 
 	assert(iov_count <= SM2_IOV_LIMIT);
@@ -157,7 +159,7 @@ static inline ssize_t sm2_generic_atomic(
 		assert(result_ioc);
 		ofi_ioc_to_iov(result_ioc, result_iov, result_count,
 			       ofi_datatype_size(datatype));
-		op_flags |= FI_REMOTE_READ;
+		proto_flags |= SM2_RMA_REQ;
 		op_flags |= FI_DELIVERY_COMPLETE;
 		/* fall through */
 	case ofi_op_atomic:
@@ -179,12 +181,12 @@ static inline ssize_t sm2_generic_atomic(
 	ret = sm2_do_atomic_inject(
 		ep, peer_gid, op, op_flags, datatype, atomic_op, iov, iov_count,
 		rma_ioc, rma_ioc_count, compare_iov, compare_count, result_iov,
-		result_count, total_len, context);
+		result_count, total_len, context, proto_flags);
 
 	if (ret)
 		goto out;
 
-	if (!(op_flags & FI_REMOTE_READ) &&
+	if (!(proto_flags & SM2_RMA_REQ) &&
 	    !(op_flags & FI_DELIVERY_COMPLETE)) {
 		ret = sm2_complete_tx(ep, context, op, op_flags);
 		if (ret)
@@ -289,7 +291,7 @@ static ssize_t sm2_atomic_inject(struct fid_ep *ep_fid, const void *buf,
 
 	ret = sm2_do_atomic_inject(ep, peer_gid, ofi_op_atomic, 0, datatype,
 				   atomic_op, &iov, 1, &rma_ioc, 1, NULL, 0,
-				   NULL, 0, total_len, 0);
+				   NULL, 0, total_len, 0, 0);
 
 	if (!ret)
 		ofi_ep_peer_tx_cntr_inc(&ep->util_ep, ofi_op_atomic);
--- a/prov/sm2/src/sm2_atomic.c
+++ b/prov/sm2/src/sm2_atomic.c
@@ -160,7 +160,6 @@ static inline ssize_t sm2_generic_atomic(
 		ofi_ioc_to_iov(result_ioc, result_iov, result_count,
 			       ofi_datatype_size(datatype));
 		proto_flags |= SM2_RMA_REQ;
-		op_flags |= FI_DELIVERY_COMPLETE;
 		/* fall through */
 	case ofi_op_atomic:
 		if (atomic_op != FI_ATOMIC_READ) {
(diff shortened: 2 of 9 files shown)


prov/sm2/src/sm2.h
```diff
@@ -304,6 +316,16 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
 	return FI_SUCCESS;
 }
 
+static inline bool sm2_proto_imm_send_comp(uint16_t proto)
+{
+	switch (proto) {
```
a-szegel, line 331: I think this could be written better as:
```
if (inject) {
   return true
}
return false
```

My gut instinct is telling me switch statements have more overhead than if, but I would need to verify with godbolt (might be wrong).

prov/sm2/src/sm2_ep.c
```diff
@@ -228,6 +261,7 @@ static void cleanup_shm_resources(struct sm2_ep *ep)
 		} else {
 			/* TODO Tell other side that we haven't processed their
 			 * message, just returned xfer_entry */
+			xfer_entry->hdr.proto_flags &= ~SM2_GENERATE_COMPLETION;
```
a-szegel, line 322: I think this should be in a different commit, and that we should add an error flag to generate error completion.

prov/sm2/src/sm2_progress.c
```diff
@@ -351,8 +438,9 @@ void sm2_progress_recv(struct sm2_ep *ep)
 					0, atomic_entry->atomic_data.data,
 					xfer_entry->hdr.size);
 			}
-			if (xfer_entry->hdr.op_flags & FI_DELIVERY_COMPLETE &&
-			    !(xfer_entry->hdr.proto_flags & SM2_UNEXP)) {
+
+			if (xfer_entry->hdr.proto_flags &
```
a-szegel, line 725: I would really like to hide the return handling logic in a static inline function for the following reasons:
1. it looks horrible with the 80 char line length limit
2. I think it makes the sm2 progress loop go from readable to very unclean.

prov/sm2/src/sm2_msg.c
```diff
@@ -38,8 +38,34 @@
...
+static inline int sm2_select_proto(void **desc, size_t iov_count,
+				   uint64_t op_flags, uint64_t total_len)
 {
+	struct ofi_mr *sm2_desc;
+	enum fi_hmem_iface iface;
+
+	if (desc && *desc) {
+		sm2_desc = (struct ofi_mr *) *desc;
+		iface = sm2_desc->iface;
+		if (iface == FI_HMEM_CUDA) {
```
a-szegel, line 50: I think we should add an else with a warn printout "Non CUDA devices not supported"

prov/sm2/src/sm2_progress.c
```diff
@@ -293,10 +438,93 @@ static int sm2_progress_recv_msg(struct sm2_ep *ep,
...
+	uint64_t comp_flags;
 	int ret;
 
+	/* TODO - Switch on protocol before switching on op to avoid messy
+	 * checks like this */
+
+	/* We don't want to look for rx entries for xfer entries with
+	 * SM2_CMA_HOST_TO_DEV and SM2_CMA_HOST_TO_DEV_ACK flags*/
+	if (xfer_entry->hdr.proto == sm2_proto_cma) {
+		if (xfer_entry->hdr.proto_flags & SM2_CMA_HOST_TO_DEV) {
```
a-szegel, line 453: I would like it if lines 453 -> 522 lived in a static inline function like sm2_host_to_dev_progress (feel free to come up with better name).  This is already messy... no need to make the common path unreadable.

prov/sm2/src/sm2_progress.c
```diff
@@ -192,8 +329,8 @@ static int sm2_start_common(struct sm2_ep *ep,
 			rx_entry->iov, rx_entry->count, &total_len, ep);
 		break;
 	case sm2_proto_cma:
-		err = sm2_progress_cma(xfer_entry, rx_entry->iov,
-				       rx_entry->count, &total_len, ep, 0);
+		err = sm2_progress_cma(ep, xfer_entry, rx_entry, &total_len, 0,
+				       &ipc_host_to_dev);
```
a-szegel, line 333: I would like to add one if statement here that handles all the device to host stuff for this function, and returns early... so we do not have to mess up the rest of the common path with "!if device to host"

a-szegel (changes requested):
Looks better... some minor changes please:

prov/sm2/src/sm2.h
a-szegel: if statement: https://godbolt.org/z/j8aW6novo

prov/sm2/src/sm2.h
```diff
@@ -96,6 +96,17 @@ enum {
 
 #define SM2_RETURN (1 << 0)
 
+/* The SM2_UNEXP flag is used for unexpected receives. On the receiver, it
+indicates that the xfer_entry that the receiver is processing is actually in the
+xfer_ctx_pool and not in the receiver's freestack.
+
+For protocols that require delivery complete semantics (CMA and IPC), the
+receiver pops an xfer_entry from its own freestack, sets SM2_UNEXP flag together
+with SM2_RETURN and SM2_GENERATE_COMPLETION and sends it to the sender. The
```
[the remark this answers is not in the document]
sunkuamzn: done

prov/sm2/src/sm2_ep.c
sunkuamzn: done

prov/sm2/src/sm2_progress.c
sunkuamzn: done

prov/sm2/src/sm2_msg.c
sunkuamzn: Host memory can also have memory descriptors. So we would need to add another check for FI_HMEM_SYSTEM and that doesn't seem worth it to me.

prov/sm2/src/sm2_progress.c
sunkuamzn: Done

prov/sm2/src/sm2_progress.c
sunkuamzn: Done

sunkuamzn:
The first commit is in https://github.com/ofiwg/libfabric/pull/9165 and will be rebased away when that PR is merged

a-szegel (approved):

sunkuamzn:
bot:aws:retest

prov/sm2/src/sm2_ep.c
```diff
@@ -228,6 +228,7 @@ static void cleanup_shm_resources(struct sm2_ep *ep)
...
```
aingerson, line 322: Maybe these patches are out of order and this will become clear in the next commits but in the previous commit you added the flag but never use it and in this one you just unset it?

prov/sm2/src/sm2_atomic.c
```diff
@@ -183,15 +182,24 @@ static inline ssize_t sm2_generic_atomic(
...
-		ret = sm2_complete_tx(ep, context, op, op_flags);
-		if (ret)
-			FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
-				"unable to process tx completion\n");
+	/* Generate immediate send completion for ofi_op_atomic unless
+	 * FI_DELIVERY_COMPLETE are set */
+	switch (op) {
+	case ofi_op_atomic:
+		if (!(op_flags & FI_DELIVERY_COMPLETE)) {
+			ret = sm2_complete_tx(ep, context, op, op_flags);
```
aingerson, line 195: I think we need a different check here. We can't complete yet for read requests

prov/sm2/src/sm2_progress.c
```diff
@@ -329,6 +329,21 @@ static int sm2_progress_atomic(struct sm2_ep *ep,
...
 out:
+
+	/* Send completion for ofi_op_atomic is generated immediately on the
+	 * sender (unless FI_DELIVERY_COMPLETE flag is set). Other ops require
+	 * delivery complete semantics, so set the SM2_GENERATE_COMPLETION flag
+	 */
+	switch (xfer_entry->hdr.op) {
+	case ofi_op_atomic:
+		if (xfer_entry->hdr.op_flags & FI_DELIVERY_COMPLETE)
+			xfer_entry->hdr.proto_flags |= SM2_GENERATE_COMPLETION;
```
aingerson, line 716: I'm a little confused about the usage of these two flags. The commit message says that one flag is basically replacing the other but you're using them together? I think the use cases for the op_flags vs proto_flags needs to be outlined a little better to make it clearer what each is used for

prov/sm2/src/sm2.h
```diff
@@ -102,6 +103,11 @@ enum {
 /* The SM2_UNEXP flag is used for unexpected receives. On the receiver, it
 indicates that the xfer_entry that the receiver is processing is actually in the
 xfer_ctx_pool and not in the receiver's freestack.
+
+For protocols that require delivery complete semantics (CMA and IPC), the
+receiver pops an xfer_entry from its own freestack, sets SM2_UNEXP flag together
+with SM2_RETURN and SM2_GENERATE_COMPLETION and sends it to the sender. The
```
aingerson, line 110: This is a little confusing. Are you saying that's for the unexpected path? Setting UNEXP with GENERATE_COMPLETION seems contradictory

prov/sm2/src/sm2.h
```diff
@@ -125,7 +131,7 @@ xfer_ctx_pool and not in the receiver's freestack.
  * 	proto - sm2 operation
  * 	proto_flags - Flags used by the sm2 protocol
  * 	sender_gid - id of msg sender
- * 	user_data - the message
+ * 	user_data - the message, for sm2_proto_inject
```
aingerson, line 139: add description for cma

prov/sm2/src/sm2.h
```diff
@@ -308,6 +319,16 @@ static inline size_t sm2_pop_xfer_entry(struct sm2_ep *ep,
 	return FI_SUCCESS;
 }
 
+static inline bool sm2_proto_imm_send_comp(uint16_t proto)
```
aingerson, line 332: I haven't seen where this is used yet but it seems like a rather silly function and implementation for
`proto == sm2_proto_cma`
This also seems confusing given the earlier implementation of the generate completion flag/delivery complete flag. not all non cma will be able to generate an immediate send completion, right?
maybe it becomes more important in other patches or you're anticipating other protocols but for now I would drop this function and add a simpler local check

prov/sm2/src/sm2_progress.c
```diff
@@ -42,6 +42,58 @@
...
+
+static int sm2_progress_cma(struct sm2_xfer_entry *xfer_entry,
+			    struct iovec *iov, size_t iov_count,
+			    size_t *total_len, struct sm2_ep *ep, int err)
+{
+	struct sm2_cma_data *cma_data =
+		(struct sm2_cma_data *) xfer_entry->user_data;
+	struct sm2_av *sm2_av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_ep_allocation_entry *entries =
```
aingerson: use new `sm2_mmap_entries` function

prov/sm2/src/sm2_progress.c
```diff
@@ -110,8 +166,35 @@ static int sm2_start_common(struct sm2_ep *ep,
...
+		 * generate a send completion */
+		ret = sm2_pop_xfer_entry(ep, &new_xfer_entry);
+		if (ret) {
+			FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+				"Unable to send xfer_entry back to "
+				"sender. Sender has requested "
+				"FI_DELIVERY_COMPLETE but will not be "
+				"able to generate a send "
+				"completion!\n");
+			return ret;
```
aingerson, line 391: These seems like a dangerous way to handle this issue... just error out the sender because the app is sending too many things of its own? If you want to keep that fix for the later patch that fixes the unexpected messaging, that's ok, but maybe include that in your comment regarding the unexpected messaging handling future plans.

prov/sm2/src/sm2_progress.c
```diff
@@ -120,6 +120,56 @@ static int sm2_progress_inject(struct sm2_xfer_entry *xfer_entry,
...
+
+	if (hmem_copy_ret < 0) {
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL,
+			"IPC recv failed with code %d\n",
+			(int) (-hmem_copy_ret));
+		err = hmem_copy_ret;
+	} else if (hmem_copy_ret != xfer_entry->hdr.size) {
+		FI_WARN(&sm2_prov, FI_LOG_EP_CTRL, "IPC recv truncated\n");
+		err = -FI_ETRUNC;
+	} else
```
aingerson: brackets here

prov/sm2/src/sm2_progress.c
```diff
@@ -120,6 +120,56 @@ static int sm2_progress_inject(struct sm2_xfer_entry *xfer_entry,
...
+		err = ret;
+
+	*total_len = hmem_copy_ret;
+
+	if (err)
+		return err;
+	else
+		return FI_SUCCESS;
```
aingerson: `return err`

prov/sm2/src/sm2_progress.c
```diff
@@ -42,20 +42,136 @@
...
+
+	dest = (char *) (uintptr_t) mr_entry->info.mapped_addr +
+	       (uintptr_t) ipc_info->offset;
+
+	ret = ofi_copy_to_hmem(ipc_info->iface, ipc_info->device, dest,
+			       cma_data->iov[0].iov_base, xfer_entry->hdr.size);
+
+	ofi_mr_cache_delete(domain->ipc_cache, mr_entry);
+
+	return ret ? ret : FI_SUCCESS;
```
aingerson, line 148: `return ret`

prov/sm2/src/sm2_progress.c
```diff
@@ -42,20 +42,136 @@
...
+
+static inline int sm2_cma_loop(struct sm2_ep *ep,
+			       struct sm2_xfer_entry *xfer_entry,
+			       struct fi_peer_rx_entry *rx_entry, size_t total,
 			       bool write)
 {
 	ssize_t ret;
+	struct sm2_av *sm2_av =
+		container_of(ep->util_ep.av, struct sm2_av, util_av);
+	struct sm2_ep_allocation_entry *entries =
```
aingerson, line 159: use `sm2_mmap_entries` call

prov/sm2/src/sm2_atomic.c
```diff
@@ -109,9 +109,9 @@ sm2_do_atomic_inject(struct sm2_ep *ep, int64_t peer_gid, uint32_t op,
 		return ret;
 
 	xfer_entry->hdr.proto = sm2_proto_inject;
-	xfer_entry->hdr.proto_flags |= proto_flags;
 
 	sm2_generic_format(xfer_entry, ep->gid, op, 0, 0, op_flags, context);
+	xfer_entry->hdr.proto_flags |= proto_flags;
```
aingerson, line 114: seems like this should just be fixed in the first patch

aingerson (commented):
I didn't do a very in-depth review since sm2 development is being discontinued for the moment but enough to catch some potential issues. I think the generate completion vs unexp vs delivery complete flag usage is extremely confusing and is prone to a lot of issues. I'm ok with just adding comments to outline the future direction for if/when development continues because I think fixing it would be an invasive procedure and I know you're just wanting to get this in to preserve what you have worked on instead of losing it completely.

prov/sm2/src/sm2.h
```diff
@@ -95,6 +95,9 @@ enum {
 	sm2_proto_max,
 };
 
+/* Protocol flags */
+#define SM2_ATOMIC_RMA_REQ (1 << 1)
```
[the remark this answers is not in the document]
sunkuamzn: Done

prov/sm2/src/sm2.h
```diff
@@ -99,6 +99,12 @@ enum {
 
 #define SM2_ATOMIC_RMA_REQ (1 << 1)
 
+/* The SM2_UNEXP flag is used for unexpected receives. On the receiver, it
+indicates that the xfer_entry that the receiver is processing is actually in the
+xfer_ctx_pool and not in the receiver's freestack.
+*/
```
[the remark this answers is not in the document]
sunkuamzn: Not sure what you mean here. The sender is as agnostic as possible to whether the receive is unexpected. If the sender has requested delivery complete, then the sender can't generate completion until the application has posted the receive and the receiver has copied the data into the buffer. So the sender needs to be aware of the unexpected receive in case of delivery complete semantics.

The unexpected flow is different for different protocols, so I didn't add all of it in this commit. I added comments for the specific protocols in those commits. See e.g. https://github.com/ofiwg/libfabric/blob/8e8a706a4e7d978b4a3e4a3b704b2659a5475f63/prov/sm2/src/sm2.h#L107-L112

prov/sm2/src/sm2_progress.c
```diff
@@ -355,7 +355,8 @@ void sm2_progress_recv(struct sm2_ep *ep)
 					0, atomic_entry->atomic_data.data,
 					xfer_entry->hdr.size);
 			}
-			if (xfer_entry->hdr.op_flags & FI_DELIVERY_COMPLETE) {
+			if (xfer_entry->hdr.op_flags & FI_DELIVERY_COMPLETE &&
+			    !(xfer_entry->hdr.proto_flags & SM2_UNEXP)) {
```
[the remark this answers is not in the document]
sunkuamzn: This was a bug in SM2. I fixed it in the next commit. This and previous commits are just adding/removing flags.

prov/sm2/src/sm2_ep.c
sunkuamzn: Squashed this commit into the previous one

prov/sm2/src/sm2_atomic.c
sunkuamzn: Do you mean RMA read? RMA is not in SM2

prov/sm2/src/sm2.h
sunkuamzn: Updated comment

prov/sm2/src/sm2.h
sunkuamzn: Done

prov/sm2/src/sm2.h
sunkuamzn: It's also used for IPC. I think it's worth having because it's used for at least two protocols. https://github.com/ofiwg/libfabric/blob/e996e60f5d1cfb9253dcb6486ae0a84db14f2500/prov/sm2/src/sm2.h#L327-L336

Unexpected messages in other protocols can't generate completion if delivery complete is requested, yes. I can't think of any other exceptions.

prov/sm2/src/sm2_progress.c
sunkuamzn: Done

prov/sm2/src/sm2_progress.c
sunkuamzn: Makes sense. Add a TODO to add a retry.

prov/sm2/src/sm2_atomic.c
sunkuamzn: The first patch introduced a new proto flag but this patch fixed a bug in how atomics are using proto flags. I think they are different enough to be in separate commits. I can squash them if you want.

prov/sm2/src/sm2_progress.c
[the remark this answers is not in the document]
sunkuamzn: The correct logic
https://github.com/ofiwg/libfabric/blob/8e8a706a4e7d978b4a3e4a3b704b2659a5475f63/prov/sm2/src/sm2_progress.c#L459-L473

prov/sm2/src/sm2.h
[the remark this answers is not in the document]
aingerson: Delivery complete just means the data is somewhere in the receiver (whether the target buffer or a bounce buffer). For the case of an unexpected message, this means we can copy the payload into an internal buffer so when the receiver actually posts the buffer, the data is actually coming from an internal location rather than the command. This is how shm does it right now for inline, inject, and sar. The sender is completely unaware whether the data is in the posted receive or not.

prov/sm2/src/sm2_atomic.c
aingerson: No, I meant an atomic with FI_ATOMIC_READ but I see this is only for the base atomics, and the atomic read is a fetch atomic so nevermind!
Looking at this again though, I think a single if statement here would work better instead of a switch.

prov/sm2/src/sm2.h
aingerson: I'm ok with merging it as is just to get it in but if we do resume development of sm2 I think we need to rework the completion flag usage. Sometimes the completion is triggered using the delivery complete flag and sometimes it's the generate completion flag and it's set later in the command and then we have this function that is confusing. I think we can make the decision up front with a single flag set and always check that flag.
For example:
```
#define SMR_COMPLETE_LATER

send_atomic()
{
if (!(atomic base && !(op_flags & FI_DELIVERY_COMPLETE))
    xfer_entry->proto_flags |= SMR_COMPLETE_LATER

if (!proto_flags & SMR_COMPLETE_LATER)
    complete_tx() 
}

progress_return()
{
//do return things

if (proto_flags & SMR_COMPLETE_LATER)
    complete_tx()
```

prov/sm2/src/sm2_progress.c
```diff
@@ -122,6 +122,54 @@ static int sm2_progress_inject(struct sm2_xfer_entry *xfer_entry,
...
+	struct ofi_mr_entry *mr_entry;
+	struct ipc_info *ipc_info = (struct ipc_info *) xfer_entry->user_data;
+
+	domain = container_of(ep->util_ep.domain, struct sm2_domain,
+			      util_domain);
+
+	ret = ofi_ipc_cache_search(domain->ipc_cache,
+				   xfer_entry->hdr.sender_gid, ipc_info,
+				   &mr_entry);
+
```
aingerson, line 277: Extra space here if you feel like removing it. If you're not going to update the patch set, don't worry about it.

prov/sm2/src/sm2_progress.c
```diff
@@ -138,6 +186,11 @@ static int sm2_start_common(struct sm2_ep *ep,
 			xfer_entry, (struct ofi_mr **) rx_entry->desc,
 			rx_entry->iov, rx_entry->count, &total_len, ep, 0);
 		break;
+	case sm2_proto_ipc:
+		err = sm2_progress_ipc(
+			xfer_entry, (struct ofi_mr **) rx_entry->desc,
```
aingerson, line 326: Format looks funky

aingerson (commented):
Just did a quick pass so we can get this in. A few lingering things that are not big issues or just need a comment in there for a future TODO or something.

prov/sm2/src/sm2.h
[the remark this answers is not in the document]
sunkuamzn: > Delivery complete just means the data is somewhere in the receiver (whether the target buffer or a bounce buffer).

That was not my understanding of delivery complete. While working on these changes, both @a-szegel  and I thought delivery complete meant that the data has to be in the receiver's application buffer. So this set of commits uses an unnecessarily higher standard and has quite a bit of logic to make that happen..

Are you OK with me just adding a comment about it for now?

prov/sm2/src/sm2_atomic.c
sunkuamzn: I can turn it into an `if`

prov/sm2/src/sm2_progress.c
sunkuamzn: We are all at the mercy of clang-format gods

prov/sm2/src/sm2.h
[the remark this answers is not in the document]
aingerson: Yeah that's fine to just have a comment for now. Thanks!

prov/sm2/src/sm2.h
sunkuamzn: Agree. The atomics work was done independently, so we didn't think about using a common flag..

prov/sm2/src/sm2_progress.c
sunkuamzn: The formatting actually got fixed in a later commit

prov/sm2/src/sm2.h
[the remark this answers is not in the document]
j-xiong: @sunkuamzn You might have mixed delivery complete with match complete.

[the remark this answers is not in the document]
sunkuamzn: Yes I think we both made that mistake https://ofiwg.github.io/libfabric/main/man/fi_cq.3.html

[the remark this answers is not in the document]
a-szegel: But the receiver's libfabric had to touch the message... it can't be sitting the the receivers fifo queue.  The receiver needs to do some processing, and return the FQE.

j-xiong:
@aingerson Is it ready to be included in 1.20?

aingerson (approved):
@j-xiong Good go!

PR #9117, opened 2023-07-06
prov/util, shm, sm2: Add FI_PEER support to util counter and make shm/sm2 use it.

shijin-aws:
This PR contains a series of commits to add FI_PEER support to util counter and make shm/sm2 use it.

Still working on adding fabtests in fabtests/pytest for the cntr for shm and sm2, will push more commits later.

PR #9140, opened 2023-07-13
prov/util, efa: use shm cntr as a peer cntr

shijin-aws:
This PR makes efa provider open shm's cntr as a peer cntr, so the cntr completion written by shm provider can be populated to efa's cntr.

Also extended the efa rdm fabtests to cover counter completion type. Before this PR, this test will fail on single node.

### DIFF
--- a/fabtests/pytest/efa/test_rdm.py
+++ b/fabtests/pytest/efa/test_rdm.py
@@ -7,9 +7,9 @@ import pytest
 @pytest.mark.parametrize("iteration_type",
                          [pytest.param("short", marks=pytest.mark.short),
                           pytest.param("standard", marks=pytest.mark.standard)])
-def test_rdm_pingpong(cmdline_args, iteration_type, completion_semantic, memory_type):
+def test_rdm_pingpong(cmdline_args, iteration_type, completion_semantic, memory_type, completion_type):
     efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong", iteration_type,
-                               completion_semantic, memory_type, "all")
+                               completion_semantic, memory_type, "all", completion_type=completion_type)
 
 @pytest.mark.functional
 def test_rdm_pingpong_range(cmdline_args, completion_semantic, memory_type, message_size):
@@ -19,9 +19,9 @@ def test_rdm_pingpong_range(cmdline_args, completion_semantic, memory_type, mess
 @pytest.mark.parametrize("iteration_type",
                          [pytest.param("short", marks=pytest.mark.short),
                           pytest.param("standard", marks=pytest.mark.standard)])
-def test_rdm_tagged_pingpong(cmdline_args, iteration_type, completion_semantic, memory_type):
+def test_rdm_tagged_pingpong(cmdline_args, iteration_type, completion_semantic, memory_type, completion_type):
     efa_run_client_server_test(cmdline_args, "fi_rdm_tagged_pingpong", iteration_type,
-                               completion_semantic, memory_type, "all")
+                               completion_semantic, memory_type, "all", completion_type=completion_type)
 
 @pytest.mark.functional
 def test_rdm_tagged_pingpong_range(cmdline_args, completion_semantic, memory_type, message_size):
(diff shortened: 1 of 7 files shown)


prov/util/src/util_cntr.c
```diff
@@ -204,11 +204,55 @@ static struct fi_ops_cntr util_cntr_no_wait_ops = {
...
+	.size = sizeof(struct fi_ops_cntr_owner),
+	.inc = &util_peer_cntr_inc,
+	.incerr = &util_peer_cntr_incerr,
+};
+
+static uint64_t ofi_peer_cntr_read(struct fid_cntr *cntr_fid)
+{
+	struct util_cntr *cntr = container_of(cntr_fid, struct util_cntr, cntr_fid);
+
+	assert(cntr->cntr_fid.fid.fclass == FI_CLASS_CNTR);
```
wenduwan, line 231: Forgot to remove?

prov/util/src/util_cntr.c
```diff
@@ -204,11 +204,55 @@ static struct fi_ops_cntr util_cntr_no_wait_ops = {
...
+	util_cntr->cntr_fid.ops->adderr(&util_cntr->cntr_fid, 1);
+}
+
+static struct fi_ops_cntr_owner util_peer_cntr_owner_ops = {
+	.size = sizeof(struct fi_ops_cntr_owner),
+	.inc = &util_peer_cntr_inc,
+	.incerr = &util_peer_cntr_incerr,
+};
+
+static uint64_t ofi_peer_cntr_read(struct fid_cntr *cntr_fid)
```
wenduwan, line 227: Not sure about the return type - I think it should be either size_t or int.

prov/efa/src/efa_cntr.c
```diff
@@ -77,30 +77,80 @@ static int efa_cntr_wait(struct fid_cntr *cntr_fid, uint64_t threshold, int time
...
 
+static int efa_cntr_close(struct fid *fid)
+{
+	struct efa_cntr *cntr;
+	int ret, retv;
+
+	retv = 0;
+	cntr = container_of(fid, struct efa_cntr, util_cntr.cntr_fid.fid);
+
+	if (cntr->shm_cntr) {
```
wenduwan, line 100: Should we do NULL cntr check here?

fabtests/pytest/efa/test_rdm.py
```diff
@@ -7,9 +7,9 @@
 @pytest.mark.parametrize("iteration_type",
                          [pytest.param("short", marks=pytest.mark.short),
                           pytest.param("standard", marks=pytest.mark.standard)])
-def test_rdm_pingpong(cmdline_args, iteration_type, completion_semantic, memory_type):
+def test_rdm_pingpong(cmdline_args, iteration_type, completion_semantic, memory_type, completion_type):
     efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong", iteration_type,
-                               completion_semantic, memory_type, "all")
+                               completion_semantic, memory_type, "all", completion_type=completion_type)
```
wenduwan, line 12: Rebase?

wenduwan (approved):
Mostly looking good. Left comments. Please see if they should be addressed. Approve in principle.

src/enosys.c
```diff
@@ -504,6 +504,18 @@ int fi_no_cntr_wait(struct fid_cntr *cntr, uint64_t threshold, int timeout)
 {
 	return -FI_ENOSYS;
 }
+uint64_t fi_no_cntr_readerr(struct fid_cntr *cntr)
```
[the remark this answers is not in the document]
shijin-aws: This no op function should have the same return with its corresponding op, `fi_cntr_readerr` will return `uint64_t`, so we have to use `uint64_t` here as well. Otherwise the function pointer will be a mismatched type.

prov/util/src/util_cntr.c
shijin-aws: See this: Only the last 2 commits are related, the first 3 commits are from https://github.com/ofiwg/libfabric/pull/9117 and will be dropped after it's merged

prov/util/src/util_cntr.c
shijin-aws: It's the change in https://github.com/ofiwg/libfabric/pull/9117. I put it here just for making the PR work.

For the return type, it needs to be consistent with fi_cntr_read, which returns uint64_t

prov/efa/src/efa_cntr.c
shijin-aws: it's a container of, it's unlikely that cntr can be NULL here.

fabtests/pytest/efa/test_rdm.py
shijin-aws: not sure I follow, why rebase?

wenduwan: nvm I thought this was from a previous commit.

shijin-aws:
AWS CI failure is real, looking now

wenduwan:
bot:aws:retest

shijin-aws:
@shefty could you review the first commit https://github.com/ofiwg/libfabric/pull/9140/commits/f8037f8422ad24bc27b7b1dddad48ff14cf5db6a that make the cntr ops public. I would like to do that so efa could customize its own cntr ops function table.

shefty (approved):
1st commit looks good.  thanks!

PR #9142, opened 2023-07-14
fabtests/unexpected_msg: Changes to unexpected_msg test

sunkuamzn:
unexpected_msg test uses the `-v` option but is not a benchmark, so move the `-v` option from benchmark to shared options

Also allocate `tx_msg_buf` which is required for verification when using device buffers

### DIFF
--- a/fabtests/functional/unexpected_msg.c
+++ b/fabtests/functional/unexpected_msg.c
@@ -70,6 +70,13 @@ static int alloc_bufs(void)
 	if (ret)
 		return ret;
 
+	if (opts.iface != FI_HMEM_SYSTEM) {
+		ret = ft_hmem_alloc_host(opts.iface, &tx_msg_buf,
+					 tx_size * opts.window_size);
+		if (ret)
+			return ret;
+	}
+
 	tx_ctx_arr = calloc(concurrent_msgs, sizeof(*tx_ctx_arr));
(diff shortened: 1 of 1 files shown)


sunkuamzn:
bot:aws:retest

lrbison:
Changes look good to me.  I'd like to see the motivation for moving the -v option in the commit message itself rather than just the PR.

CI error is in `rhel8.test_suites.libfabric.test_fabtests.test_fabtests_functional[single_node-debug].test_rdm_tagged_peek` which times out.

sunkuamzn:
> Changes look good to me. I'd like to see the motivation for moving the -v option in the commit message itself rather than just the PR.
Fixed

aingerson (commented):
The second patch looks fine (but could you update the commit title to say "fabtests/unexpected_msg" to indicate that it is only relevant to that test?

The first patch I think I would drop because not all tests that use the cs opts support data verification so I think it could lead into users thinking their test is doing data validation when it isn't. The tests that do add data verification (but aren't a benchmark) do this by manually parsing the -v argument within their arg parsing (see bw.c for example).  I would instead change the first patch to just manually check it for just the unexpected message test.

aingerson (approved):
Thanks!

shefty (approved):

shefty:
Intel CI failure is known issue (now resolved) in tcp provider.

lrbison (approved):

PR #9151, opened 2023-07-18
prov/efa: use an internal fi_mr_attr for memory registration

wenduwan:
EFA provider should be allowed to customize fi_mr_attr struct for internal memory registration, without affecting application fi_mr_attr with EFA.

This also helps avoiding ABI compatibility issues with new fi_mr_attr fields, if the application is compiled with an older libfabric.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -39,7 +39,7 @@
 
 static int efa_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 			  uint64_t flags, struct fid_mr **mr_fid);
-static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr);
+static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *attr);
 static int efa_mr_dereg_impl(struct efa_mr *efa_mr);
 
 
@@ -317,7 +317,7 @@ int efa_mr_cache_entry_reg(struct ofi_mr_cache *cache,
 	else if (attr.iface == FI_HMEM_SYNAPSEAI)
 		attr.device.synapseai = entry->info.device;
 
-	ret = efa_mr_reg_impl(efa_mr, 0, (void *)&attr);
+	ret = efa_mr_reg_impl(efa_mr, 0, (const void *)&attr);
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -810,14 +811,19 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 	efa_mr->mr_fid.mem_desc = NULL;
 	efa_mr->mr_fid.key = FI_KEY_NOTAVAIL;
 
-	ret = efa_mr_hmem_setup(efa_mr, mr_attr, flags);
+	memcpy(&mr_attr, attr, sizeof(struct fi_mr_attr));
```
shijin-aws, line 814: I think you should use `sizeof(*attr)`, `sizeof(struct fi_mr_attr)` is using the new header, which could be larger than the actual size of `attr` and cause segfault.

wenduwan: I don't think `sizeof(*attr)` works because `attr` is `void *`.

I also considered this problem. memcpy does not define the behavior in this case. To be extra safe we can add FI_VERSION check and modify the size.

shijin-aws: but how did you get the size for an older libfabric version in the new libfabric code?

wenduwan: Please see update. We can exclude hmem_data from the size.

shijin-aws: This is messy, I am not sure whether it is desirable. @wzamazon WDYT?

wzamazon: I do not quite understand the problem it try to solve. So it appears that user requested 1.19 API but did not pass hmem_data?

shijin-aws: It's application use old libfabric headers to compile, but run with latest libfabric main branch

shijin-aws: application doesn't pass hmem_data

prov/efa/src/efa_mr.c
```diff
@@ -874,29 +887,29 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, void *attr)
 		/* We need to add FI_REMOTE_READ to allow for Read implemented
 		* message protocols.
 		*/
-		original_access = mr_attr->access;
-		mr_attr->access |= FI_REMOTE_READ;
```
wzamazon, line 878: I think it would be better to create a new mr_attribute in this branch, like `shm_mr_attr`, and fill `shm_mr_attr.hmem_data` directly with `efa_mr.gdrcopy_handler`.

wenduwan: Cool thanks. Turns out we can use `ofi_mr_update_attr` to fill `mr_attr`, and it handles ABI version checks.

shijin-aws (approved):
thanks

wenduwan:
Reworded commit

wenduwan:
Intel CI failure should be unrelated

PR #9152, opened 2023-07-18
prov/shm: Make SHM use clang-format

a-szegel:
After the clang-format test with sm2 was largely successful, expand clang-format to the shm provider.


shefty:
Widespread formatting changes to an existing provider are not encouraged.  It makes it near impossible to cherry pick fixes from main to the stable branches.

darrylabbate:
I strongly suggest modifying the config file to enforce a minimal set of style restrictions instead; such as hard tabs for indentation. Otherwise I'd prefer to see some examples of clang-format actually improving readability, because this diff is doing it no favors. Not to mention the damage done to the Git blame view and cherry-pick functionality.

a-szegel:
For git blame (I mention the solution in the commit message):

1. [github picks up the .git-blame-ignore-revs file and hides it from git blame (on github)](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/#fun-fact%3A-github-picks-up-%60.git-blame-ignore-revs%60)
2. [Developers can also easily fix this locally for their environment](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/#add-ignored-revisions-to-your-local-git-config)

For cherry picking... yes this will make it harder.

a-szegel:
This clang-format config was hand crafted to align with the formatting standards of the libfabric project.  This PR shows how bad humans are at being consistent with their formatting, even when a lot of effort has gone into making it so that is consistent.  Why not let a computer do this, and enforce it for us?

a-szegel:
As for cherry picking, Alexia is currently working on a branch of SHM that makes SHM very similar to SM2. We performance tested the beta version of that change and it has fantastic improvements, so we want to proceed with it. After that change goes in, we will not cherry-pick going forward anymore b/c the new SHM will be so different than the old SHM, that it won't be worth it.

It makes sense to push this change in now if we prioritize our git blame history... b/c we have a functional change coming down the pipeline that changes almost everything... we can largely get rid of the formatting fixes now (and hidden in git blame).

It makes sense to do this change after the SHM re-write if we want to prioritize the ease of our cherry-picking from now, until that change goes into effect (when cherry picking won't be useful anymore).

a-szegel:
I think the most valuable part of clang-format is the developer time saved (both implementer and reviewer) from not having to deal with formatting comments.  It makes on-boarding new developers not used to our standards 100x easier.

aingerson:
To be clear, I don't agree with this change. The part I was agreeing with is that if there are any widespread formatting/stylistic things we want to fix that we can do that as part of the shm rearchitecting since, after that, cherry-picking will be much harder anyway because the provider will essentially be a whole new provider.
There are some things I really like about clang; for example, it saves reviewers time and reduces noise when it comes to catching things like spaces after ifs/fors, bracket usage, stray spaces, horrible line length violations. But I think the way the clang formatting is imposed right now is not helpful. As Sean said, these are guidelines, not laws, and we shouldn't bow down to whatever clang says, because the whole point is to try to make it more readable for humans, so if we want it one way and not the other and clang disagrees, I think that's very annoying. There have been too many instances to count in sm2 where we hesitantly merged code in that looked really weird just because clang was complaining otherwise.
I think there is a world where we can use clang to help with readability and developer onboarding but I think it needs to be rethought out so it's on more of a minimally suggestive basis.

PR #9156, opened 2023-07-19
prov/verbs: Add additional checks to vrb_shutdown_qp_in_err

chien-intel:
Flushed CQEs would come ether after the CQE for the op that caused the QP tranition or in QP destrouction.  There is no need to check with flushed CQEs.

Add missing return code check for ibv_query_qp.


shefty:
bot:aws:retest

shefty (approved):
thanks

PR #9162, opened 2023-07-20
include/osd.h: add defintion of memory ordering for no atomics

aingerson:
If no atomics are available, ofi will emulate atomics using locks. In this case, no memory ordering is required.
If the atomics ordering primitives are not available, the build will fail. This adds a definition for the memory ordering to allow it to build


shefty:
If this is needed for v1.18.x, please pick

PR #9165, opened 2023-07-21
prov/sm2: Replace FI_REMOTE_READ op flag with  SM2_RMA_REQ proto flag

sunkuamzn:
Now that SM2 has proto flags, we can use a proto flag instead of overloading the op flag. The other op flag FI_DELIVERY_COMPLETE will be removed in a future commit

### DIFF
--- a/prov/sm2/src/sm2.h
+++ b/prov/sm2/src/sm2.h
@@ -95,6 +95,9 @@ enum {
 	sm2_proto_max,
 };
 
+/* Protocol flags */
+#define SM2_RMA_REQ (1 << 1)
+
 /*
  * 	next - fifo linked list next ptr
  * 		This is volatile for a reason, many things touch this
--- a/prov/sm2/src/sm2_atomic.c
+++ b/prov/sm2/src/sm2_atomic.c
@@ -99,7 +99,7 @@ sm2_do_atomic_inject(struct sm2_ep *ep, int64_t peer_gid, uint32_t op,
 		     const struct fi_rma_ioc *rma_ioc, size_t rma_ioc_count,
 		     const struct iovec *compare_iov, size_t compare_count,
 		     const struct iovec *result_iov, size_t result_count,
-		     size_t total_len, void *context)
+		     size_t total_len, void *context, uint16_t proto_flags)
 {
 	struct sm2_xfer_entry *xfer_entry;
 	size_t ret;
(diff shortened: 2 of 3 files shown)


a-szegel (approved):

prov/sm2/src/sm2.h
```diff
@@ -95,6 +95,9 @@ enum {
 	sm2_proto_max,
 };
 
+/* Protocol flags */
+#define SM2_ATOMIC_RMA_REQ (1 << 1)
```
aingerson, line 99: I would suggest making this as generic as possible to make it more usable if needed in the future so I would change this to SM2_RMA_REQ in case RMA could use it

prov/sm2/src/sm2_atomic.c
```diff
@@ -109,6 +109,7 @@ sm2_do_atomic_inject(struct sm2_ep *ep, int64_t peer_gid, uint32_t op,
 		return ret;
 
 	xfer_entry->hdr.proto = sm2_proto_inject;
+	xfer_entry->hdr.proto_flags |= proto_flags;
```
aingerson, line 112: the field needs to be reset after the pop - don't OR in, just set

aingerson (commented):
Small things, otherwise looks good!

prov/sm2/src/sm2.h
sunkuamzn: @lrbison are you using such a flag in RMA? Are you OK with calling it `SM2_RMA_REQ`?

prov/sm2/src/sm2_atomic.c
sunkuamzn: We set proto_flags to 0 in generic format, so OR is OK. But it was in the wrong place. Fixed now. https://github.com/ofiwg/libfabric/blob/main/prov/sm2/src/sm2.h#L198

It would've been caught by FT_UNIT_TEST but there are other failures that @a-szegel will work on

prov/sm2/src/sm2.h
```diff
@@ -95,6 +95,9 @@ enum {
 	sm2_proto_max,
 };
 
+/* Protocol flags */
+#define SM2_RMA_REQ (1 << 1)
```
aingerson, line 99: Sorry just realized this one - let's have the first flag start at 1<<0

aingerson (approved):
One thing that I missed on the first one, sorry! Giving my approval anyway though because it's small and not a huge deal and can be fixed later

PR #9166, opened 2023-07-21
prov/efa: consume iov only if offset is non-zero

wenduwan:
ofi_consume_iov_desc asserts consume length is non-zero. This change guards it with a length check.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -1629,15 +1629,18 @@ int efa_rdm_rxe_post_local_read_or_queue(struct efa_rdm_ope *rxe,
 	/* setup iov */
 	assert(pkt_entry->ope == rxe);
 	assert(rxe->desc && efa_mr_is_hmem(rxe->desc[0]));
+	assert(rxe->iov_count);
 	iov_count = rxe->iov_count;
 	memcpy(iov, rxe->iov, rxe->iov_count * sizeof(struct iovec));
 	memcpy(desc, rxe->desc, rxe->iov_count * sizeof(void *));
-	ofi_consume_iov_desc(iov, desc, &iov_count, rx_data_offset);
-	if (iov_count == 0) {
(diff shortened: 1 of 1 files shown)


shijin-aws:
I was wondering if this bug passed the CI, i.e. do we need some fabtests to catch it?

sunkuamzn:
I think a library function like `ofi_consume_iov_desc` should allow a zero byte consume call. I know @lrbison spent a lot of time on that function in the past month. Is there a reason why zero byte consume is not allowed?

lrbison:
Sean preferred to require iov_count>0 on entrance to the function [here](https://github.com/ofiwg/libfabric/issues/9035#issuecomment-1591670563)

wenduwan:
> I was wondering why this bug passed the CI, i.e. do we need some fabtests to catch it?

It was revealed in fabtests on p4d

shijin-aws:
>It was revealed in fabtests on p4d

Yeah, then we should be able to hit this on non-p4d with a simpler reproducer. Can we do that?

lrbison:
> to_consume>0
But now that I look at https://github.com/ofiwg/libfabric/pull/9050 again, I think the function is still correct when to_consume = 0.  @shefty do you have any objection to removing that assert from ofi_consume_iov_desc?

wenduwan:
> hit this on non-p4d with a simpler reproducer

@shijin-aws It looks to me that this code path is only activated on p4d, which requires p2p and EFA RDMA read.

https://github.com/wenduwan/libfabric/blob/1fd36f13f4581a6fbeca83e5d55c58d99a75f96f/prov/efa/src/rdm/efa_rdm_pke_utils.c#L308

Do we have a fabtest that simulates that?

shefty:
If the new function works when to_consume = 0, I have no issue removing the assert.  And I agree, the function looks like it handles to_consume = 0.

wenduwan:
@lrbison Please remove the assertion then. That simplifies many things. Thank you!

PR #9167, opened 2023-07-21
src/iov.c: ofi_consume_iov allows 0-byte consume

lrbison:
Removed assert(to_consume) in ofi_consume_iov and ofi_consume_rma_iov. This check is not required as the functions can correctly consume 0 bytes from an iov.

--

A regression was noticed by EFA nightly CI after merging #9050.  Rather than prevent zero-sized consumes, it was decided that it should be legal to consume 0 bytes (as it was before #9050).  See https://github.com/ofiwg/libfabric/pull/9166 for additional context.

### DIFF
--- a/src/iov.c
+++ b/src/iov.c
@@ -66,7 +66,7 @@ void ofi_consume_iov_desc(struct iovec *iov, void **desc,
 	struct iovec *cur_iov;
 	void **cur_desc;
 
-	assert(to_consume && *iov_count);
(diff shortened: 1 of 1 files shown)


sunkuamzn (approved):

wenduwan (approved):

shefty (approved):

wenduwan:
@lrbison Ok to merge?

PR #9170, opened 2023-07-25
Use nvmlDeviceGetCount_v2() first for CUDA check

qkoziol:
Check for CUDA devices with nvmlDeviceGetCount_v2() first, to avoid more expensive call to cudaGetDeviceCount() when possible.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -595,7 +595,7 @@ AS_IF([test x"$with_cuda" != x"no"],
 				 [cuda_runtime.h],
 				 [cudart],
 				 [cudaMemcpy],
-				 [-lcuda],
+				 [-lcuda -lnvidia-ml],
 				 [$with_cuda],
 				 [],
 				 [have_cuda=1])
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -42,14 +42,15 @@
 
 #include <cuda.h>
 #include <cuda_runtime.h>
+#include <nvml.h>
 
 #if ENABLE_CUDA_DLOPEN
 #include <dlfcn.h>
 #endif
 
 /*
- * Convenience higher-order macros for enumerating CUDA driver/runtime API
- * function names
+ * Convenience higher-order macros for enumerating CUDA driver/runtime and
+ * NVML API function names
  */
 #define CUDA_DRIVER_FUNCS_DEF(_)	\
 	_(cuGetErrorName)		\
@@ -75,6 +76,11 @@
 	_(cudaIpcGetMemHandle)		\
 	_(cudaIpcCloseMemHandle)
 
+#define NVML_FUNCS_DEF(_)	        \
+	_(nvmlInit_v2)                  \
+	_(nvmlDeviceGetCount_v2)        \
+	_(nvmlShutdown)
+
 static struct {
 	int   device_count;
 	bool  p2p_access_supported;
@@ -82,13 +88,15 @@ static struct {
 	bool  use_ipc;
 	void *driver_handle;
 	void *runtime_handle;
+	void *nvml_handle;
 } cuda_attr = {
 	.device_count         = -1,
 	.p2p_access_supported = false,
 	.use_gdrcopy          = false,
 	.use_ipc              = false,
 	.driver_handle        = NULL,
-	.runtime_handle       = NULL
+	.runtime_handle       = NULL,
+	.nvml_handle          = NULL
 };
 
 static struct {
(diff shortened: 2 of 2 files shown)


src/hmem_cuda.c
```diff
@@ -430,38 +454,60 @@ static int cuda_hmem_dl_init(void)
...
+	nvml_ret = ofi_nvmlDeviceGetCount_v2(&nvml_device_count);
+	if (NVML_SUCCESS == nvml_ret) {
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Number of NVIDIA devices detected: %u\n",
+			nvml_device_count);
 
-	case cudaErrorNoDevice:
-		return -FI_ENOSYS;
+		if (nvml_device_count > 0) {
+			cudaError_t cuda_ret;
```
shefty, line 480: Declare variables at the top of the function.  It makes them easier to find.

configure.ac
```diff
@@ -595,7 +595,7 @@ AS_IF([test x"$with_cuda" != x"no"],
 				 [cuda_runtime.h],
 				 [cudart],
 				 [cudaMemcpy],
-				 [-lcuda],
+				 [-lcuda -lnvidia-ml],
```
shefty, line 598: What is the impact of requiring this new library?  Is it always available?

There's a missing set of checks somewhere, since we now need nvml.h present.

qkoziol: Yes, the nvml header is installed with CUDA, since 2012.  I don't think we need another check for it.

src/hmem_cuda.c
```diff
@@ -411,11 +425,21 @@ static int cuda_hmem_dl_init(void)
 		goto err_dlclose_cuda_runtime;
 	}
 
+	cuda_attr.nvml_handle = dlopen("libnvidia-ml.so", RTLD_NOW);
+	if (!cuda_attr.nvml_handle) {
+		FI_WARN(&core_prov, FI_LOG_CORE,
+			"Failed to dlopen libnvidia-ml.so\n");
+		goto err_dlclose_cuda_driver;
+	}
+
```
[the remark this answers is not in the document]
qkoziol: Yes, this should be installed with the CUDA libraries, and it's a failure if the install is wrong.

src/hmem_cuda.c
```diff
@@ -430,38 +454,60 @@ static int cuda_hmem_dl_init(void)
...
+	 */
 
-	/* Verify CUDA compute-capable devices are present on the host. */
-	cuda_ret = ofi_cudaGetDeviceCount(&cuda_attr.device_count);
-	switch (cuda_ret) {
-	case cudaSuccess:
-		break;
+	/* Verify NVIDIA devices are present on the host. */
+	nvml_ret = ofi_nvmlDeviceGetCount_v2(&nvml_device_count);
+	if (NVML_SUCCESS == nvml_ret) {
```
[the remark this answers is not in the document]
qkoziol: OK, will update

src/hmem_cuda.c
qkoziol: OK, will update

src/hmem_cuda.c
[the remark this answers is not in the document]
shijin-aws: ok, thanks.

shijin-aws:
Please look at the AWS CI failure

src/hmem_cuda.c
```diff
@@ -463,51 +463,49 @@ static void cuda_hmem_dl_cleanup(void)
...
+
+                FI_INFO(&core_prov, FI_LOG_CORE,
+                        "Number of CUDA devices detected: %d\n",
+                        cuda_attr.device_count);
+        } else {
+                cuda_attr.device_count = 0;
+        }
+
+        if (cuda_attr.device_count <= 0)
+                return -FI_ENOSYS;
```
shefty, line 508: Don't do updates in a separate patch.  Just update the original patch directly.  Otherwise, there are potentially 2 patches that need to be picked into the stable branches.

For the original patch, please format the commit messages, so that the 1st line is only about 72 characters long.  It needs to be easily viewable using git log --oneline for fast identification and cherry picking of patches.  Add more details into the main message.  This seems like a very minor optimization for a function that's called once.  Some details on why we're adding a second get device count call into this path should be captured in the commit message.

qkoziol: > Don't do updates in a separate patch. Just update the original patch directly. Otherwise, there are potentially 2 patches that need to be picked into the stable branches.
> 

Why not squash during the merge from my branch into main?

qkoziol: >  This seems like a very minor optimization for a function that's called once. Some details on why we're adding a second get device count call into this path should be captured in the commit message.

nvmlDeviceGetCount is much lighter-weight than cudaGetDeviceCount and allows an early-out from the routine, improving performance for a simple simple IMB Allreduce test on AWS instances without GPUs by 40-300%.

qkoziol:
bot:aws:retest

src/hmem_cuda.c
shijin-aws: >Why not squash during the merge from my branch into main?

I don't quite follow. It is a convention in PRs that updates on a commit should be combined with that original commits via `git commit --amend`. There is no auto squash for github merge AFAIK

qkoziol: > > Why not squash during the merge from my branch into main?
> 
> I don't quite follow. It is a convention in PRs that updates on a commit should be combined with that original commits via `git commit --amend`. There is no such kind of "auto squash" for github merge AFAIK

All of the projects i've contributed to have done this in different ways.  ¯\_(ツ)_/¯

qkoziol:
I'll close this PR and address all the formatting and commit-related issues by submitting another PR with a single commit covering the combined set of changes.

shefty:
Please don't close the PR.  That loses the comments.  Just update the original patch and force push.

qkoziol:
Squashed, updated the commit message, and force-pushed

shefty (approved):

shijin-aws (approved):

shijin-aws:
We are evaluating the updated version before merging it.

qkoziol:
Performance is still good, merging now.

qkoziol:
@shefty - Should this be backported to any release branches?

shefty:
I'll let AWS decide that.  IMO, it doesn't seem critical enough to backport.  I doubt many apps would notice this outside of some benchmarks.

PR #9171, opened 2023-07-25
contrib/intel/jenkins: Fix DAOS logging and summary

zachdworkin:
DAOS was using the output of the wrong command for its log. This PR fixes that issue.
Any invocation of runtests.py can now provide its own logfile instead of using the one generated with slurm or the jenkinsfile.
It also addresses the failure summary case for daos by cleaning up the multiple prints of the same test and providing a better log of what failed.
The summary gets better spacing in addition to the failure case for daos.
The release case gets updated to match the true/false boolean variable update that all other variables in the Jenkinsfile follow.

### DIFF
--- a/contrib/intel/jenkins/runtests.py
+++ b/contrib/intel/jenkins/runtests.py
@@ -35,6 +35,8 @@ parser.add_argument('--user_env', help="Run with additional environment " \
                     "variables", nargs='*', action=ParseDict, default={})
 parser.add_argument('--mpi', help="Select mpi to use for middlewares",
                     choices=['impi', 'mpich', 'ompi'], default='impi')
+parser.add_argument('--log_file', help="Full path to log file",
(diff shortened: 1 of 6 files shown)


contrib/intel/jenkins/runtests.py
```diff
@@ -35,13 +35,16 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+parser.add_argument('--log', help="Full path to log file",
+                    default=os.environ['DEFAULT_LOG_LOCATION'], type=str)
 
 args = parser.parse_args()
 args_core = args.prov
 
 args_util = args.util
 args_device = args.device
 user_env = args.user_env
+log = args.log
```
nikhilnanal, line 48: nit: why not just call it log file. :)

zachdworkin: Done! There was no reason I called it log vs log_file

PR #9172, opened 2023-07-25
prov/sm2: Reduce unexpected message copy to message size

a-szegel:
Previously, SM2 was copying the entire xfer_entry into the unexpected message queue. We can reduce the amount of bytes copied to the header + the message size. This will give us a minor performance improvement.

### DIFF
--- a/prov/sm2/src/sm2_progress.c
+++ b/prov/sm2/src/sm2_progress.c
@@ -144,7 +144,8 @@ static int sm2_alloc_xfer_entry_ctx(struct sm2_ep *ep,
 		return -FI_ENOMEM;
 	}
 
-	memcpy(&xfer_ctx->xfer_entry, xfer_entry, sizeof(*xfer_entry));
+	memcpy(&xfer_ctx->xfer_entry, xfer_entry,
+	       sizeof(xfer_entry->hdr) + xfer_entry->hdr.size);
 	xfer_ctx->ep = ep;
 
 	rx_entry->peer_context = xfer_ctx;


prov/sm2/src/sm2_progress.c
```diff
@@ -144,7 +144,8 @@ static int sm2_alloc_xfer_entry_ctx(struct sm2_ep *ep,
 		return -FI_ENOMEM;
 	}
 
-	memcpy(&xfer_ctx->xfer_entry, xfer_entry, sizeof(*xfer_entry));
+	memcpy(&xfer_ctx->xfer_entry, xfer_entry,
+	       sizeof(struct sm2_xfer_hdr) + xfer_entry->hdr.size);
```
aingerson, line 148: Change the sizeof to get the size of the field to stay consistent with the rest of the code
`sizeof(xfer_entry->hdr)`

aingerson (approved):
Conditional approval based on one coding style change

aingerson (approved):

sunkuamzn:
This change only works for inject protocol. It will break for CMA and IPC protocols because we set the `hdr.size` there to be the size of the message, which can be a lot larger than the xfer_entry size.

I could add checks for those protocols to copy only the header.

a-szegel:
I think that size needs to be the size of the inject packet that you are using, and in the CMA message body... you need a cma transfer size.

a-szegel:
AWS is stopping development on SM2 provider.

PR #9173, opened 2023-07-27
fabtest: Fix reading addressing options

shefty:
When running the following client/server apps:

fi_rdm_tagged_peek -p tcp -s 192.168.1.1
fi_rdm_tagged_peek -p tcp 192.168.1.1

The client app will segfault due to trying to insert a NULL address into the AV.  The problem can be traced back to ft_read_addr_opts() and the following if statement:

if (opts->dst_addr && (opts->src_addr || !opts->oob_port))

Whatever the intent of this check is isn't clear.  However, the result is that because the client has only specified a server address (opts->dst_addr is set, but opts->src_addr is NULL), it gets skipped.  The code falls through to the else clause, winds it way through libfabric calls, and eventually ends up with fi_getinfo() returning fi_info::dst_addr = NULL.  This address is then inserted into the AV, which leads to the crash.

Fix by removing the odd extra requirement that we either need to specify a src address with the dst address, or the oob_port must be set.  A src address can either be set or not; the code handles both cases.  And the use of oob sockets for setup is independent of whether we're acting as a client or server for the test.


shefty:
I suspect this will break CI somewhere because of the checks that we removed.  The question will be to figure out what those tests are doing and how the checks should be structured.

shefty:
@shijin-aws - Can you provide feedback on the AWS failures?

shijin-aws:
@shefty I think this PR breaks the oob addr exchange path somehow, all fabtests with efa provider failed. 

```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.38.241 'timeout 360 /bin/bash --login -c '"'"'FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr9173/install/fabtests/bin/fi_av_xfer -e rdm -p efa -E=9228'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.38.241 'timeout 360 /bin/bash --login -c '"'"'FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr9173/install/fabtests/bin/fi_av_xfer -e rdm -p efa -E=9228 172.31.38.241'"'"''
client_stdout:
fi_getinfo(): common/shared.c:987, ret=-61 (No data available)
fi_getinfo(): common/shared.c:987, ret=-61 (No data available)
AV address removal: FAIL
AV re-insertion address: FAIL

client returncode: 61
server_stdout:
AV address removal: FAIL

server returncode: 104
```

EFA can only work with OOB address exchange.

shefty:
Thanks.  It looks like the requirement is when the -b or -E options are given, but -O is not, then the src/dst addresses map to the OOB socket address.  In this case, no src/dst address is passed into fi_getinfo().  Does this align with EFA's requirements?

It's definitely confusing what the src/dst addresses actually mean.  We should re-examine the command options.  Maybe -b or -E should require -O, for example.

shijin-aws:
> In this case, no src/dst address is passed into fi_getinfo(). Does this align with EFA's requirements?

Yes, I think so.

shijin-aws:
CI is still running but the new revision fixes all the fabtests failure earlier.

PR #9174, opened 2023-07-27
contrib/intel/jenkins: Add file importing to Summarizer & Cleanup Jenkinsfile

zachdworkin:
Import files from sub-clusters (non main pipeline clusters) to be run through the summarizer once rather than have to invoke the summarizer on each cluster.
Add DSA and GPU-V3 to the summary

Cleanup the Jenkinsfile:
Make re-used variables global pipeline environment variables
Add methods for withEnv statements to remove large triple quote shell blocks.
Replace non script{} blocks with script{} blocks for better maintainability 
Reduce new line use in post stage by consolidating one line bracket statements into one line
Reduce most lines to less than 80 characters to follow libfabric coding style

### DIFF
--- a/contrib/intel/jenkins/Jenkinsfile
+++ b/contrib/intel/jenkins/Jenkinsfile
@@ -89,6 +89,33 @@ def gather_logs(cluster, key, dest, source) {
   }
 }
 
+def summarize(item, verbose=false, release=false) {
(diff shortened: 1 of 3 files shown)


shefty:
Are the changes needed in the stable branches to avoid CI breaking there?

zachdworkin:
@shefty Possibly. These changes are summary related so "theoretically" the jenkinsfile should just run the upstream summary file and not print out values for the new things added. I was going to cherry pick them if it causes failures.

Issue #9123, opened 2023-07-07
[1.18.1] prov/verbs: Double FI_SHUTDOWN event received (regression from 1.18.0)


sydidelot:
With 1.18.0 and below, there is an implicit assumption that the verbs provider writes the `FI_SHUTDOWN` event only once when an endpoint shuts down.
With 1.18.1, this assumption is no longer true and the application may receive 2 `FI_SHUTDOWN` events for the same endpoint.
This new behavior was introduced by the following commit: https://github.com/ofiwg/libfabric/commit/2abe57cd893d70a52137758c2c5b3c7fbf0be2c1

With the above commit, the first `FI_SHUTDOWN` event is written by `vrb_shutdown_qp_in_err()` and the second originates from `vrb_eq_cm_process_event()` when the CM event `RDMA_CM_EVENT_DISCONNECTED` is received.

One problem I see in the code is that `vrb_shutdown_qp_in_err()` moves the QP to an error state, but `ep->id` remains valid. As a consequence, the next call to `rdma_disconnect()` is allowed to generate the CM event `RDMA_CM_EVENT_DISCONNECTED`.

sydidelot:
FYI, the order is not always `vrb_shutdown_qp_in_err()` then `RDMA_CM_EVENT_DISCONNECTED`. Sometimes the CM event happens before.

chien-intel:
do you have a specific test case that's demonstrating this behavior?

shefty:
@chien-intel , @sydidelot works with a storage app (DDN).  He's likely seeing the issue there.

The vrb_ep likely needs its own state that we can track, so we can discard rdma_cm events if we're in the wrong state.  I have plans to add an vrb_ep state, but I'm not there yet.  The code currently uses the ibv_qp.state, which can be modified outside of libfabric (via libibverbs).

sydidelot:
@chien-intel @shefty Sorry for the missing context: I'm seeing the issue with the storage app we are developing at DDN, which is closed source unfortunately. The app uses MSG endpoint (no RDM) and the double FI_SHUTDOWN event triggers on normal endpoint shutdown on both client and server sides.
I don't think the pattern is specific to our code and it can probably be reproduced on a synthetic application. FYI, I will be on PTO for the next 3 weeks and I won't be able to progress on this issue. For the time being, I reverted the patch in my local branch.
I agree with @shefty and I seems to me that the verbs provider needs its own state.

chien-intel:
[...]

PR #9175, opened 2023-07-27
prov/verbs: Track ep state to prevent duplicate shutdown events

shefty:
Track the connection state of msg endpoints.  This is needed to safely transition into a disconnect state, while ensuring that we only report a shutdown event to the user once.

There are 2 separate ways that we can determine that an endpoint has been disconnected.  One is by reading a disconnect event from the rdma cm.  Another is by reading an error completion from the CQ associated with the underlying QP.  Because we're not guaranteed to get a disconnect event (the remote peer system may have crashed without generating an RDMA disconnect traffic), if we read an error completion from the CQ, we need to notify the user that the local EP is no longer usable.  We do this by generating a local shutdown event.

However, we _could_ read a disconnect event from the EQ.  We have no way of knowing beforehand, or which order the disconnect notification could occur (from a CQ or EQ event).  The state checks allow us to transition the EP once and generate a single local shutdown event.

### DIFF
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -612,12 +612,12 @@ int vrb_init_progress(struct vrb_progress *progress, struct ibv_context *verbs)
 {
 	int ret;
 
-	ret = ofi_genlock_init(&progress->lock, OFI_LOCK_MUTEX);
+	ret = ofi_genlock_init(&progress->ep_lock, OFI_LOCK_MUTEX);
 	if (ret)
 		return ret;
 
 	/* Active lock will be needed when adding rdm ep support */
-	progress->active_lock = &progress->lock;
+	progress->active_lock = &progress->ep_lock;
 
 	ret = ofi_bufpool_create(&progress->ctx_pool, sizeof(struct fi_context),
 				 16, 0, 1024, OFI_BUFPOOL_NO_TRACK);
@@ -627,12 +627,12 @@ int vrb_init_progress(struct vrb_progress *progress, struct ibv_context *verbs)
 	return 0;
 
 err1:
-	ofi_genlock_destroy(&progress->lock);
+	ofi_genlock_destroy(&progress->ep_lock);
 	return ret;
 }
 
 void vrb_close_progress(struct vrb_progress *progress)
 {
 	ofi_bufpool_destroy(progress->ctx_pool);
-	ofi_genlock_destroy(&progress->lock);
+	ofi_genlock_destroy(&progress->ep_lock);
 }
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2015 Intel Corporation, Inc.  All rights reserved.
+ * Copyright (c) Intel Corporation, Inc.  All rights reserved.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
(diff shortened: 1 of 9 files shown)


shefty:
This targets a fix for #9123, but the changes are untested at the moment.  (Verbs wasn't enabled on the cluster I was developing on.)  My larger concern is that there is no locking through many of these calls.  That may not be an issue for some of the CM paths.  The app can call multiple CM calls on the same endpoint in parallel.  But the EQ handling does need some sort of coordination with reading the CQ.

The end goal is to use similar locking as implemented by the tcp provider (progress level locks at the domain level), but that's not yet in place.  Posting for early review/feedback.

prov/verbs/src/verbs_cq.c
```diff
@@ -227,7 +227,7 @@ int vrb_poll_cq(struct vrb_cq *cq, struct ibv_wc *wc)
 		ctx = (struct vrb_context *) (uintptr_t) wc->wr_id;
 		wc->wr_id = (uintptr_t) ctx->user_ctx;
 		if (wc->status != IBV_WC_SUCCESS && wc->status != IBV_WC_WR_FLUSH_ERR)
-			vrb_shutdown_qp_in_err(ctx->ep);
+			vrb_shutdown_ep(ctx->ep);
```
chien-intel, line 230: I don't know what non-IBV_WC_SUCCESS would result in QP transitioning to IBV_QPS_ERR by the adapter.  That's the reason behind ibv_query_qp to make sure we are only closing on QP in IBV_QPS_ERR.  This would potentially close QP that's still in RTS with a non-fatal status.  Also, don't think status and behavior would be the same between HW vendors thus checking current qp state is the only safe way to do this.

shefty:
Update on locking: Until we can integrate rdm support directly into the provider and rework locking, we're in a racy situation.  The problem is that an endpoint can be closed (closing the rdma_cm_id) at the same time we try to read and process an event from the rdma_cm.  The current code protects this using eq->lock.  We hold the eq->lock when reading and processing rdma_cm events.  We also acquire the eq->lock when closing the ep and destroying the rdma_cm_id.

However, we hold the progress->ep_lock when reading and processing CQ entries.  If we see a failed transfer, we transition the EP into a disconnected state and write an entry to the EQ.  This requires a lock ordering of: progress->ep_lock -> eq->lock.  So, I can clearly not choose the wine in front of you.

But, the progress->ep_lock is needed to protect the state of the EP.  And the EP state can be driven by rdma_cm events.  This requires a lock ordering of: eq->lock -> progress->ep_lock.  So, I can clearly not choose the wine in front or me.

The current code has implemented the first lock ordering option: progress->ep_lock -> eq->lock,

shefty:
@zachdworkin - Can you please check the failures for this PR?  The failure is in trying to clone the repo, so maybe there's an issue with the proxy or something...

zachdworkin:
@shefty Fixed

shefty:
@swelch - These changes update the locking used by XRC.  XRC was using the eq->lock to protect some state information.  To avoid nested locking between the ep and eq, I added a new eq lock, an event_lock.  It's not a great name, but hopefully it's a temporary solution.  I added state checking to the ep, but did not try to integrate the xrc state into the new state.  It was easier not to touch it.

prov/verbs/src/verbs_ep.c
```diff
@@ -114,26 +114,18 @@ ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
...
-	if (!ep || !ep->ibv_qp || ep->ibv_qp->state == IBV_QPS_UNKNOWN || !ep->eq)
-		return;
-
-	memset(&attr, 0, sizeof(attr));
-	memset(&init_attr, 0, sizeof(init_attr));
-	if (ibv_query_qp(ep->ibv_qp, &attr, IBV_QP_STATE, &init_attr))
-		return;
-	if (attr.cur_qp_state != IBV_QPS_ERR)
+	ofi_genlock_held(vrb_ep2_progress(ep)->active_lock);
+	if (!ep || ep->state != VRB_CONNECTED || !ep->eq)
```
chien-intel, line 122: ep is already used in the line above, no need to check it here.

shefty: good catch... coverity would have complained about this.  I'll update

chien-intel: when I get a chance I will run io500 with PR

chien-intel: No performance regression with IO500/DAOS

PR #9181, opened 2023-08-02
prov/shm: Remove mmap protocol

a-szegel:
The mmap protocol is not used or tested, and is just dead code taking up space in our repository. Remove FI_SHM_SAR_THRESHOLD which controls switching between mmap and sar protocols. We used to default to SAR if CMA wasn't available until message size SIZE_MAX, with the env var to modify our switching point.  This patch should not change any behavior in the SHM provider.


aingerson:
@a-szegel Yes, the code is rather useless but it's also rather useless to remove it if we're just going to redo the entire provider. Removing it will require bumping the provider version and I don't think it's worth it personally. Let's just wait until the new provider integration

PR #9182, opened 2023-08-02
prov/shm: limit CQ polling to 8

a-szegel:
Limit the number of messages progressed to 8. This prevents SHM from starving the ep of sends. This change improves collective performance for SHM.


a-szegel:
After more performance testing... I am not convinced we actually want this for SHM.  CQ polling at 8 hurts osu_allreduce.  Moving it up to 32 makes osu_allreduce better.

PR #9183, opened 2023-08-02
contrib/intel/jenkins: Add tracer check to DSA

zachdworkin:
Add FI_LOG_LEVEL=warn to DSA stage so that it prints out information to determine if DSA is actually being used.
The summary stage will then parse this information to determine if DSA was correctly used in the tests and will fail if it did not.


aingerson:
Code looks good to me but I'd like to test this out with triggering two differently failure cases to double check:
Where DSA does not get enabled/initialized properly - this can be replicated by not setting the FI_SHM_USE_DSA_SAR variable
Where DSA is enabled but the send path is not taken - this can be replicated by always returning false for [this function](https://github.com/ofiwg/libfabric/blob/main/include/ofi_mr.h#L286)

aingerson (approved):

PR #9184, opened 2023-08-02
prov/shm: move all ofi_shm/util_shm code to prov/shm

a-szegel:
No one else uses the shm specific code which is specific to the provider and should be in the shm provider folder.


a-szegel:
This is Alexia's patch that I cherry picked into main... I approve this.

aingerson:
Is the git rm of the old files missing? I don't see it. (That was my bad sorry). I think this should just show up as a rename, not an addition

a-szegel:
I also did some greps and saw a few things were missed.

aingerson (approved):
That looks correct now! Thanks for fixing my mistake!

shefty (approved):

a-szegel:
AWS CI error not related.

Issue #9141, opened 2023-07-13
prov/shm: smr_progress_ipc_list() causes 200ns performance regression in 256 rank osu_gather benchmark


a-szegel:
**Describe the bug**
This commit causes a 200ns performance regression in osu_gather benchmark.
```
commit 6c86459680e00af11a98d43d60547eb2f239b7df (HEAD)
Author: Amir Shehata <shehataa@ornl.gov>
Date:   Thu Nov 24 16:59:33 2022 -0800

    prov/shm: Update SHM to use ROCR
    
    Update the SHM provider to use the ROCR HMEM asynchronous memory
    operations.
    
       . Unify the ipc and sar freestack, since they use the same structure.
       . When progressing an IPC operation, check if the device is ROCR and trigger
         an asynchronous operation.
       . When an asynchronous operation is queued, create an IPC entry and
         add it to the queue of pending operations.
       . During the top level progress loop check the queue of pending asynchronous
         operations and query the state of each one. Generate a complete event
         for finished operations. Since completions happen outside the context
         of libfabric we can't rely on the ep->region->signal flag to be set.
         Always check the pending queue. This shouldn't introduce much of an
         overhead if the queue is empty
       . use the ep->util_ep lock to protect the free stack and the ipc list
         of pending operations.
    
    Signed-off-by: Amir Shehata <shehataa@ornl.gov>
```
I isolated the cause of the regression to be this: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_progress.c#L1287

This regression is mostly fixed if we move `smr_progress_ipc_list(ep);` into 
```
	if (ofi_atomic_cas_bool32(&ep->region->signal, 1, 0)) {
		if (smr_env.use_dsa_sar)
			smr_dsa_progress(ep);
		smr_progress_resp(ep);
		smr_progress_cmd(ep);
		smr_progress_sar_list(ep);
		smr_progress_ipc_list(ep);
	}
```

There is still an extra 50ns overhead when I compare to commenting `smr_progress_ipc_list()` out completely (which I understand isn't ok b/c it is required for correctness with hmem ipc capable devices).

Is it possible to make polling the ipc_list lighter (and other lists)?
Is it possible to move polling the IPC list inside the signal if statement and maintain correctness?
Is it possible to get rid of this polling at runtime if we do not have any GPU's on the instance?

**To Reproduce**
Run osu_gather on 8 c6gn.16xlarge with 32 processes per node (256 rank total).

[...]

PR #9185, opened 2023-08-02
prov/shm: Only poll IPC list when ROCR IPC is enabled

a-szegel:
This optimization disables the polling of the async ipc list when there are no users of it.

https://github.com/ofiwg/libfabric/issues/9141

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -1329,6 +1329,11 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			}
 		}
 
+		if (ofi_hmem_any_ipc_enabled())
+			ep->smr_progress_ipc_list = smr_progress_ipc_list;
+		else
+			ep->smr_progress_ipc_list = smr_progress_ipc_list_noop;
+
 		if (!ep->srx) {
 			domain = container_of(ep->util_ep.domain,
 					      struct smr_domain,
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -1127,7 +1127,7 @@ static void smr_progress_cmd(struct smr_ep *ep)
 	ofi_genlock_unlock(&ep->util_ep.lock);
 }
 
-static void smr_progress_ipc_list(struct smr_ep *ep)
+void smr_progress_ipc_list(struct smr_ep *ep)
 {
 	struct smr_pend_entry *ipc_entry;
 	struct smr_region *peer_smr;
@@ -1273,5 +1273,5 @@ void smr_ep_progress(struct util_ep *util_ep)
 
 	/* always drive forward the ipc list since the completion is
 	 * independent of any action by the provider */
-	smr_progress_ipc_list(ep);
+	ep->smr_progress_ipc_list(ep);
 }
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -168,6 +168,16 @@ struct ofi_hmem_ops hmem_ops[] = {
 	},
 };
 
+bool ofi_hmem_any_ipc_enabled(void)
+{
+	int iface;
+
+	for (iface = 0; iface < ARRAY_SIZE(hmem_ops); iface++)
+		if (ofi_hmem_is_initialized(iface) && ofi_hmem_is_ipc_enabled(iface))
+			return true;
+	return false;
+}
+
 int ofi_create_async_copy_event(enum fi_hmem_iface iface, uint64_t device,
 				ofi_hmem_async_event_t *event)
 {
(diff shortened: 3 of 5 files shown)


prov/shm/src/smr_ep.c
```diff
@@ -1329,6 +1329,8 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			}
 		}
 
+		ep->smr_async_device_copy = rocr_is_ipc_enabled();
```
sunkuamzn, line 1332: `rocr_is_ipc_enabled` is internal to the HMEM API. Might be better to check `hmem_ops[FI_HMEM_ROCR].initialized`

prov/shm/src/smr_progress.c
```diff
@@ -1273,5 +1273,6 @@ void smr_ep_progress(struct util_ep *util_ep)
 
 	/* always drive forward the ipc list since the completion is
 	 * independent of any action by the provider */
-	smr_progress_ipc_list(ep);
+	if (ep->smr_async_device_copy)
+		smr_progress_ipc_list(ep);
```
shefty, line 1277: Rather than adding a check here, we can setup a function pointer that either calls a no-op or smr_progress_ipc_list().  No-op functions are typically faster than adding an if statement and doesn't add a check when we do need to progress the ipc list.

prov/shm/src/smr_ep.c
shefty: The call should be ofi_hmem_is_ipc_enabled().  This may need to be called in a loop for all HMEM, and if any of them return true, then we need to call progress_ipc_list() below.

aingerson: ROCR is currently the only iface that uses the async protocol so I think they are trying to optimize this out for CUDA which doesn't need it.
@sunkuamzn @a-szegel Are there plans or desires to extend the CUDA protocol to use it asynchronously? If so, I agree with Sean, make it general to all interfaces. If the plan is to always use CUDA synchronously and you're looking to optimize that case specifically, then I think it's ok to call it directly. If the aim here is only to optimize the non HMEM case, then I would still make it general.

aingerson (commented):
Agreed with Sean about the function pointer

prov/shm/src/smr_ep.c
sunkuamzn: We don't have plans to implement async CUDA copies in the near future. We need to do some benchmarking to see if it's even worth doing.

a-szegel: kept at ROCR IPC and added a comment.

prov/shm/src/smr_ep.c
```diff
@@ -1329,6 +1329,13 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 			}
 		}
 
+		/* TODO: When we add support ipc support to other HMEM providers,
+		 * include them in this check */
+		if (rocr_is_ipc_enabled())
```
shefty, line 1334: I'd still prefer a generic check here.  We're overriding smr_progress_ipc_list(), which is a generic function not specific to rocr.  I'd rather see something like:

```
if (ofi_hmem_any_ipc_enabled())
    blah blah
```
And have that call make the specific checks that are needed.  And the abstracted call we should use is ofi_hmem_is_ipc_enabled(), not a direct call to rocr.
```
bool ofi_hmem_any_ipc_enabled(void)
{
    int iface;

    for (iface = 0; iface < ARRAY_SIZE(hmem_ops); iface++)
        if (ofi_hmem_is_ipc_enabled(iface))
            return true;
    return false;
}
```
and then adding support for other async calls doesn't require any changes through this path.

src/hmem.c
```diff
@@ -168,6 +168,16 @@ struct ofi_hmem_ops hmem_ops[] = {
 	},
 };
 
+bool ofi_hmem_any_ipc_enabled(void)
+{
+	int iface;
+
+	for (iface = 0; iface < ARRAY_SIZE(hmem_ops); iface++)
+		if (ofi_hmem_is_ipc_enabled(iface))
```
shefty, line 176: We probably need to add ```ofi_hmem_is_initialized(iface && ``` to this check.  I missed this in my prior comment.

aingerson (approved):

src/hmem.c
a-szegel: Hmmm... I guess my local non-gpu dev instance wasn't going in there b/c I didn't compile in support for those things, but we can't assume package managers will be the same.  Good Catch.

shefty (approved):

a-szegel:
bot:aws:retest

sunkuamzn:
A bit late to this PR but this implementation won't get us any improvement on GPU instances because we will have CUDA enabled. We should instead add something like `ofi_hmem_async_ipc_enabled` which return true only for ROCR

PR #9186, opened 2023-08-03
prov/shm: improve DSA library linking

aingerson:
Currently, if DSA libraries are found on the build node, the libfabric build will always get linked to the accfg-config library. If the compute nodes do not have this library, runs will fail even if DSA is not being used.

This changes the DSA config checks to only check the headers and then adds DL loading/functions to the DSA code to use the library dynamically so that it can be used on a cluster with different configurations.

Also fixes a compiler warning to remove an unused variable in the DSA code

### DIFF
--- a/prov/shm/src/smr_dsa.c
+++ b/prov/shm/src/smr_dsa.c
@@ -536,14 +536,12 @@ dsa_process_partially_completed_desc(struct smr_dsa_context *dsa_context,
 static void dsa_update_tx_entry(struct smr_region *smr,
 				struct dsa_cmd_context *dsa_cmd_context)
 {
-	struct smr_region *peer_smr;
 	struct smr_resp *resp;
 	struct smr_cmd *cmd;
 	struct smr_tx_entry *tx_entry = dsa_cmd_context->entry_ptr;
 
 	tx_entry->bytes_done += dsa_cmd_context->bytes_in_progress;
 	cmd = &tx_entry->cmd;
-	peer_smr = smr_peer_region(smr, tx_entry->peer_id);
 	resp = smr_get_ptr(smr, cmd->msg.hdr.src_data);
 
 	assert(resp->status == SMR_STATUS_BUSY);
--- a/prov/shm/src/smr_dsa.c
+++ b/prov/shm/src/smr_dsa.c
@@ -38,6 +38,7 @@
 
 #if SHM_HAVE_DSA
 
+#include <dlfcn.h>
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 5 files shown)


prov/shm/src/smr_dsa.c
```diff
@@ -97,6 +98,42 @@ struct smr_dsa_context {
 	unsigned long page_fault_stats[2];
 };
 
+bool dsa_init = false;
```
shefty, line 101: static?

prov/shm/src/smr_dsa.c
```diff
@@ -642,6 +679,154 @@ static bool dsa_check_cmd_status(struct smr_dsa_context *dsa_context,
 }
 
 /* SMR functions */
+
+void smr_dsa_init(void)
+{
+	if (dsa_init)
```
shefty, line 685: How does this get called more than once?

aingerson: It shouldn't but I added it just in case/if the code were to be removed from shm and used more generally like the HMEM stuff but I can remove it. It isn't necessary

prov/shm/src/smr_dsa.c
```diff
@@ -644,6 +677,150 @@ static bool dsa_check_cmd_status(struct smr_dsa_context *dsa_context,
...
+
+	dsa_ops.accfg_device_get_next = dlsym(libdsa_handle,
+					      "accfg_device_get_next");
+	if (!dsa_ops.accfg_device_get_next) {
+		FI_WARN(&core_prov, FI_LOG_CORE,
+			"Failed to find accfg_device_get_next\n");
+		goto err_dlclose;
+	}
+
+	return;
```
shefty, line 818: close libdsa_handle somewhere?

aingerson: ah yes probably good

shefty:
bot:aws:retest

prov/shm/src/smr_dsa.c
```diff
@@ -97,6 +98,40 @@ struct smr_dsa_context {
...
+#define device_foreach(ctx, device) \
+	for (device = dsa_ops.accfg_device_get_first(ctx); \
+	     device != NULL; \
+	     device = dsa_ops.accfg_device_get_next(device))
+
+
+#define wq_foreach(device, wq) \
+	for (wq = dsa_ops.accfg_wq_get_first(device); \
+	     wq != NULL; \
+	     wq = dsa_ops.accfg_wq_get_next(wq))
```
shefty, line 131: Consider adding 'dsa' somewhere in the above macro names to avoid confusion and namespace collision. E.g. dsa_foreach_device / dsa_foreach_wq

shefty (commented):
minor request for a name change

prov/shm/src/smr_dsa.c
aingerson: good call

PR #9188, opened 2023-08-03
[v1.18.x] prov/efa: remove rxr_rm_tx/rx_cq_check()

sunkuamzn:
rxr_rm_cq_check() check whether util_cq is overflow, and set the RXR_RM_TX/RX_CQ_FULL flag.

If the flags are full, fi_send/read/write/atomic will return -FI_EAGAIN.

The code as is does not work, because when -FI_EAGAIN is returned, progress engine is not called so if the RXR_RM_TX/RX_CQ_FULL flag is ever set, an application will hang.

Meanwhile, the check itself is not necesary because util_cq can handle the situation that CQ is full
by writing overflow CQ entry.

Therefore, this patch remove the code from critical code path.

Reimplementation of commit 1de4f5a9032cd3b2df0cca128735b95dd4a1b5d2

### DIFF
--- a/prov/efa/src/rdm/rxr_atomic.c
+++ b/prov/efa/src/rdm/rxr_atomic.c
@@ -139,11 +139,6 @@ ssize_t rxr_atomic_generic_efa(struct rxr_ep *rxr_ep,
 
 	ofi_mutex_lock(&rxr_ep->base_ep.util_ep.lock);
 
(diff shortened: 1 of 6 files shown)


shijin-aws (approved):

lrbison (approved):

sunkuamzn:
bot:aws:retest

PR #9189, opened 2023-08-03
contrib/intel/jenkins: Add automatic emailing support to summary

zachdworkin:
With these changes Cloudbees will send an email at the end of an automatic weekly run to the mail recipients detailed in the controller. This can be turned on or off depending on the weekly environment variable or by removing/adding the new flag --send_mail to the summarizer.


nikhilnanal:
looks good to me

PR #9193, opened 2023-08-04
prov/shm: Remove unnecessary SAR check from generic_send

a-szegel:
Remove SAR status check from smr_generic_sendmsg() b/c the check is only required for the SAR protocol which is yet to be selected.  The check already happens in the SAR protocol in smr_format_sar().

### DIFF
--- a/prov/shm/src/smr_msg.c
+++ b/prov/shm/src/smr_msg.c
@@ -99,13 +99,6 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 	peer_id = smr_peer_data(ep->region)[id].addr.id;
 	peer_smr = smr_peer_region(ep->region, id);
 
-	if (smr_peer_data(ep->region)[id].sar_status)
-		return -FI_EAGAIN;
-
-	ret = smr_cmd_queue_next(smr_cmd_queue(peer_smr), &ce, &pos);
-	if (ret == -FI_ENOENT)
-		return -FI_EAGAIN;
-
 	ofi_genlock_lock(&ep->util_ep.lock);
 
 	total_len = ofi_total_iov_len(iov, iov_count);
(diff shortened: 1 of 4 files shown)


shefty (approved):
Change looks fine to me and the commit message makes sense, but I need to defer to others on correctness.

aingerson (commented):
This check happens in other places as well (inject and rma) so we should update those as well.
I would also be curious if this affects the SAR performance. I think ideally the send flow would be restructured so we get the command entry right before doing the protocol and keep the sar status here before we get the command. If there is a large SAR blocking more SAR sends, we're going to waste time getting a command entry just to have to release it again, both of which require atomic coordination.
I think ideally it would look like this:
- select proto
- if (proto == sar && sar_status) return eagain
- get command
- do proto

and remove the sar status from the do sar function. I'm not really sure why it's over there with a different lock around it...? I mean I probably wrote that but I have no idea what I was doing... I think the sar status can be protected with the same lock as the send resources

shijin-aws:
If we don't return early, will it hurt performance?

a-szegel:
I want to get rid of that if statement check from the inject path, so I think we should move popping command queue within the do_* protocols to avoid this (like SM2).

aingerson:
That's how it used to be (getting and committing the command inside the protocol functions) but Amir's refactoring changed that because the rma command is now inside the command so in the rma case you need to format the rma iovs before it is committed.

a-szegel:
The two stage send (claim + commit) is less than ideal.  We could save this if statement on the non-SAR path by claiming the entry in do_sar, and still committing it in generic_send / generic_rma.  I don't like this option very much b/c it makes the code more difficult to understand for a very minor optimization.  I will keep thinking about this.

a-szegel:
Are these real failures?

a-szegel:
bot:aws:retest

prov/shm/src/smr_msg.c
```diff
@@ -114,6 +107,13 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 	proto = smr_select_proto(desc, iov_count, smr_cma_enabled(ep, peer_smr),
 	                         op, total_len, op_flags);
 
+	if (proto == smr_src_sar && smr_peer_data(ep->region)[id].sar_status)
+		return -FI_EAGAIN;
```
aingerson, line 111: goto unlock instead of return here and below

prov/shm/src/smr_rma.c
```diff
@@ -154,19 +151,22 @@ static ssize_t smr_generic_rma(struct smr_ep *ep, const struct iovec *iov,
...
 	}
 
+	total_len = ofi_total_iov_len(iov, iov_count);
+	assert(!(op_flags & FI_INJECT) || total_len <= SMR_INJECT_SIZE);
+
+	proto = smr_select_proto(desc, iov_count, smr_cma_enabled(ep, peer_smr),
+	                         op, total_len, op_flags);
+
+	if (proto == smr_src_sar && smr_peer_data(ep->region)[id].sar_status)
+		return -FI_EAGAIN;
```
aingerson, line 161: same here

aingerson (commented):
Fix locking, otherwise good

a-szegel:
Should be ready once it passes CI.

PR #9194, opened 2023-08-04
prov/shm: Move selection logic to smr_select_proto()

a-szegel:
The same protocol selection logic was copy/pasted into both smr_generic_rma() and smr_generic_send(). Remove copy/pasted logic and put it in the common function where it belongs.

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -275,9 +275,8 @@ size_t smr_copy_from_sar(struct smr_freestack *sar_pool, struct smr_resp *resp,
 			 struct smr_cmd *cmd, struct ofi_mr **mr,
 			 const struct iovec *iov, size_t count,
 			 size_t *bytes_done, int *next);
-int smr_select_proto(enum fi_hmem_iface, bool use_ipc, bool cma_avail,
-                     bool gdrcopy_avail, uint32_t op, uint64_t total_len,
-                     uint64_t op_flags);
+int smr_select_proto(void **desc, size_t iov_count, bool cma_avail,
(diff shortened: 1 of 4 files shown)


shijin-aws (approved):

prov/shm/src/smr.h
```diff
@@ -275,9 +275,9 @@ size_t smr_copy_from_sar(struct smr_freestack *sar_pool, struct smr_resp *resp,
 			 struct smr_cmd *cmd, struct ofi_mr **mr,
 			 const struct iovec *iov, size_t count,
 			 size_t *bytes_done, int *next);
-int smr_select_proto(enum fi_hmem_iface, bool use_ipc, bool cma_avail,
-                     bool gdrcopy_avail, uint32_t op, uint64_t total_len,
-                     uint64_t op_flags);
+int smr_select_proto(enum fi_hmem_iface iface, void **desc, size_t iov_count,
+                     bool use_ipc, bool cma_avail, uint32_t op, uint64_t total_len,
```
aingerson, line 279: you shouldn't need to pass in iface or use_ipc anymore. this is all calculated locally now

a-szegel: I think it would be better on the performance use parameters instead of re-calculating right?

aingerson: I'm not sure I'm following here. You're passing in use_ipc which now is no longer calculated in the calling function so there should be no use for it?

a-szegel: scratch what I said... you are right.

aingerson (approved):
Thanks!

a-szegel:
bot:aws:retest

PR #9198, opened 2023-08-05
prov/shm: Replace tx_lock with ep_lock

a-szegel:
The tx_lock allows a sender thread to send at the same time a receiver thread can receive on an endpoint. This change gets rid of that functionality and only allows 1 thread in an endpoint send or receive at a time. When we move all SHM resources to the sender, we suspect that the sender/receiver will be to intertwined to be able to progress independently.

This change also boosts performance by switching to a ofi_genlock instead of an ofi_spinlock.

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -514,13 +514,10 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 	if (peer_smr->max_sar_buf_per_peer == 0)
 		return -FI_EAGAIN;
 
-	ofi_genlock_lock(&ep->util_ep.lock);
 	if (smr_peer_data(ep->region)[id].sar_status) {
-		ofi_genlock_unlock(&ep->util_ep.lock);
 		return -FI_EAGAIN;
 	}
 	smr_peer_data(smr)[id].sar_status = SMR_STATUS_SAR_FULL;
-	ofi_genlock_unlock(&ep->util_ep.lock);
 
 	sar_needed = (total_len + SMR_SAR_SIZE - 1) / SMR_SAR_SIZE;
 	cmd->msg.data.buf_batch_size = MIN(SMR_BUF_BATCH_MAX,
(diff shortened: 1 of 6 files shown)


a-szegel:
bot:aws:retest

shefty (approved):
Changes are okay with me, but please confirm with @aingerson as I don't have a clue about the locking in shm. :)

prov/shm/src/smr_ep.c
```diff
@@ -599,7 +599,7 @@ int smr_select_proto(enum fi_hmem_iface iface, bool use_ipc, bool cma_avail,
 		return smr_src_ipc;
 
 	if (total_len > SMR_INJECT_SIZE && cma_avail)
-		return smr_src_iov;
+		return smr_src_sar;
```
aingerson, line 602: ?

prov/shm/src/smr_msg.c
```diff
@@ -109,7 +109,7 @@ static ssize_t smr_generic_sendmsg(struct smr_ep *ep, const struct iovec *iov,
 	if (ret == -FI_ENOENT)
 		return -FI_EAGAIN;
 
-	ofi_spin_lock(&ep->tx_lock);
+	ofi_genlock_lock(&ep->util_ep.lock);
```
aingerson, line 112: I believe do_sar acquires the ep lock so you'll have to remove that to prevent from deadlocking

a-szegel: this should have been caught in AWS CI... hmmm

prov/shm/src/smr_ep.c
a-szegel: oh no... I made this change to step through SAR, and must not have noticed it got added to this PR (I thought I reset before)... need to check my diffs more carefully

prov/shm/src/smr_msg.c
a-szegel: Was able to reproduce deadlock when I added `smr_domain->util_domain.threading = FI_THREAD_SAFE;` back in to smr_domain... it wasn't deadlocking b/c we weren't configuring threading in our tests.

prov/shm/src/smr_ep.c
```diff
@@ -514,13 +514,11 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 	if (peer_smr->max_sar_buf_per_peer == 0)
 		return -FI_EAGAIN;
 
-	ofi_genlock_lock(&ep->util_ep.lock);
 	if (smr_peer_data(ep->region)[id].sar_status) {
 		ofi_genlock_unlock(&ep->util_ep.lock);
```
aingerson, line 518: one more

a-szegel: done

aingerson (approved):

PR #9200, opened 2023-08-07
prov/efa: Use correct tx/rx op_flags for shm

shijin-aws:
Currently, the tx/rx op flags are hard-coded as FI_COMPLETION for shm, which is wrong because currently efa is using shm as a peer provider and should pass application requested op flags to shm as well.

Also add unit test for this change


shijin-aws:
need to backport to v1.19.x as well

wenduwan (approved):

a-szegel (approved):

shijin-aws:
AWS CI is real... looking now.

shijin-aws:
The AWS CI failure is a hang in `fi_rma_bw -S r:0;4;64 -p efa -v -E`, which ends up being a SHM bug that couldn't handle 0 byte transfer in SAR protocol. I opened https://github.com/ofiwg/libfabric/pull/9203 to fix it and will rebase this PR after it's merged

PR #9203, opened 2023-08-08
prov/shm: Handle 0 byte transfer in SAR protocol

shijin-aws:
Currently, SAR protocol doesn't handle 0 byte transfer because it fails to update the resp->status correctly in this corner case. This patch fixes this issue.

Also added test to verify the fix.

The issue can be reproduced by running 

```
fi_rma_bw -p shm -I 5 -v -S 0 -e rdm -s localhost
```

`-v` is needed to make fi_rma_bw run with FI_DELIVERY_COMPLETE, which will force SHM to use SAR protocol.

### DIFF
--- a/fabtests/test_configs/shm/shm.exclude
+++ b/fabtests/test_configs/shm/shm.exclude
@@ -7,7 +7,6 @@ inject_test
 -e dgram
 
 # Exclude tests that use sread/polling
--S
 rdm_cntr_pingpong
 poll
 
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -549,6 +549,10 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 	pending->bytes_done = 0;
 	pending->next = 0;
 
+	/* Nothing to copy for 0 byte transfer */
+	if (!cmd->msg.hdr.size)
+		return 0;
+
 	if (cmd->msg.hdr.op != ofi_op_read_req) {
 		if (smr_env.use_dsa_sar && ofi_mr_all_host(mr, count)) {
 			ret = smr_dsa_copy_to_sar(ep, smr_sar_pool(peer_smr),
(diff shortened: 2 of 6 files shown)


fabtests/test_configs/shm/shm.exclude
```diff
@@ -7,7 +7,6 @@ inject_test
 -e dgram
 
 # Exclude tests that use sread/polling
--S
```
shijin-aws, line 10: I am not sure if any test is using `-S` for sread. This may need change if that is true

aingerson (approved):
Thanks!

shijin-aws:
bot:aws:retest

prov/shm/src/smr_ep.c
```diff
@@ -549,6 +549,10 @@ static int smr_format_sar(struct smr_ep *ep, struct smr_cmd *cmd,
 	pending->bytes_done = 0;
 	pending->next = 0;
 
+	/* Nothing to copy for 0 byte transfer */
+	if (!cmd->msg.hdr.size)
+		return 0;
```
shefty, line 554: Can you update the commit message to indicate *why* SAR needs to handle a 0-byte transfer?  Because on the surface it doesn't make sense.

shijin-aws: Yeah, it is actually caught by a corner case when I run `fi_rma_bw -p shm -I 5 -v -S 0 -e rdm -s localhost`. In this test, data validation is enabled, and it forces SHM to run with FI_DELIVERY_COMPLETE (DC), and turn off inject. SHM will use SAR always for DC, so if you run fi_send/recv with SHM + DC + 0 byte, this bug can be triggered. I don't know if any real application is doing that though

shefty: I'd like that information in the commit message, so that's it's captured in the git logs, not the PR text.

shijin-aws: Done.

shijin-aws: @shefty do you have more comments?

shijin-aws:
Is Intel CI failure related?

aingerson:
@shijin-aws We had some systems come up incorrectly after the reboot, you can ignore these failures. shm tests passed

shefty (approved):

PR #9206, opened 2023-08-09
Try libnvidia-ml.so.1 if .so symlink missing, and skip NVML when not available

qkoziol:
Local testing on Debian has shown a case where the libnvidia-ml-so symlink to libnvidia-ml.so.1 doesn't exist (but libnvidia-ml.so.1 exists).  So, try to dlopen the .so.1 library if the generic .so dlopen fails.  In either case, if libnvidia-ml doesn't exist, but the CUDA shared libraries do, skip trying to use the NVML routines for getting the device count and just proceed to using the CUDA routine.

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -442,18 +442,25 @@ static int cuda_hmem_dl_init(void)
 	cuda_attr.nvml_handle = dlopen("libnvidia-ml.so", RTLD_NOW);
 	if (!cuda_attr.nvml_handle) {
 		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to dlopen libnvidia-ml.so\n");
-		goto err_dlclose_cuda_driver;
+			"Failed to dlopen libnvidia-ml.so.  Trying libnvidia-ml.so.1\n");
+		cuda_attr.nvml_handle = dlopen("libnvidia-ml.so", RTLD_NOW);
+		if (!cuda_attr.nvml_handle) {
+			FI_WARN(&core_prov, FI_LOG_CORE,
+			"Failed to dlopen libnvidia-ml.so.1 also, bypassing nvml calls\n");
+		}
 	}
 
 	CUDA_DRIVER_FUNCS_DEF(CUDA_DRIVER_FUNCS_DLOPEN)
 	CUDA_RUNTIME_FUNCS_DEF(CUDA_RUNTIME_FUNCS_DLOPEN)
-	NVML_FUNCS_DEF(NVML_LIB_FUNCS_DLOPEN)
+	if (cuda_attr.nvml_handle) {
+		NVML_FUNCS_DEF(NVML_LIB_FUNCS_DLOPEN)
+	}
 
 	return FI_SUCCESS;
 
 err_dlclose_nvml_lib:
-	dlclose(cuda_attr.nvml_handle);
+	if (cuda_attr.nvml_handle)
+		dlclose(cuda_attr.nvml_handle);
 err_dlclose_cuda_driver:
 	dlclose(cuda_attr.driver_handle);
 err_dlclose_cuda_runtime:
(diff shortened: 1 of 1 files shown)


src/hmem_cuda.c
```diff
@@ -442,18 +442,25 @@ static int cuda_hmem_dl_init(void)
...
+	if (cuda_attr.nvml_handle) {
+		NVML_FUNCS_DEF(NVML_LIB_FUNCS_DLOPEN)
+	}
 
 	return FI_SUCCESS;
 
 err_dlclose_nvml_lib:
-	dlclose(cuda_attr.nvml_handle);
+	if (cuda_attr.nvml_handle)
+		dlclose(cuda_attr.nvml_handle);
```
shefty, line 463: This entire label section does not appear to be accessible.

src/hmem_cuda.c
```diff
@@ -442,18 +442,25 @@ static int cuda_hmem_dl_init(void)
...
 err_dlclose_cuda_driver:
 	dlclose(cuda_attr.driver_handle);
```
shefty, line 465: Neither is this label section.  The goto for it was removed.

src/hmem_cuda.c
```diff
@@ -484,29 +492,46 @@ static int cuda_hmem_verify_devices(void)
...
+#endif
 
-	/* Verify NVIDIA devices are present on the host. */
-	nvml_ret = ofi_nvmlDeviceGetCount_v2(&nvml_device_count);
-	if (nvml_ret != NVML_SUCCESS) {
-	        ofi_nvmlShutdown();
-                return -FI_ENOSYS;
-        }
+	/* Check for NVIDIA devices, if NVML library is dlopen-ed*/
+	if (nvml_device_count == 0) {
```
shefty, line 506: This check should be ```if(cuda_attr.nvml_handle)```, which directly captures what's needed prior to executing the selected code.  Don't use a proxy variable.

src/hmem_cuda.c
```diff
@@ -484,29 +492,46 @@ static int cuda_hmem_verify_devices(void)
...
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Number of NVIDIA devices detected: %u\n",
+			nvml_device_count);
+	} else {
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Skipping check for NVIDIA devices with NVML routines\n");
+	}
 
         /* If NVIDIA devices are present, now perform more expensive check
          * for actual GPUs.
```
shefty, line 537: The check below here would be something like ```if (!nvml_handle || nvml_device_count)```.  Though we probably don't even need checks here.  If an nvml_device wasn't found, we should have already returned from the call.

src/hmem_cuda.c
qkoziol: The label is referenced within the CUDA_FUNCS_DLOPEN macro, within the *_FUNCS_DEF macros.

qkoziol:
I agree with your suggestions for the checks in cuda_hmem_verify_devices(), I'll update the PR

src/hmem_cuda.c
shefty: got it -- thanks

src/hmem_cuda.c
qkoziol: We need the check, for the case when the NVML library failed to dlopen() and for when it was opened and found devices that we are confirming with cudaGetDeviceCount.

src/hmem_cuda.c
```diff
@@ -484,34 +492,41 @@ static int cuda_hmem_verify_devices(void)
...
+		nvml_ret = ofi_nvmlShutdown();
+		if (nvml_ret != NVML_SUCCESS)
+			return -FI_ENOSYS;
 
-        FI_INFO(&core_prov, FI_LOG_CORE,
-                "Number of NVIDIA devices detected: %u\n",
-                nvml_device_count);
+		FI_INFO(&core_prov, FI_LOG_CORE,
+			"Number of NVIDIA devices detected: %u\n",
+			nvml_device_count);
```
shefty, line 520: Couldn't we check nvml_device_count here and return if == 0?  We wouldn't need to initialize nvm_device_count above, and this is the only path where it might change.  Then the if statement below could be removed.

qkoziol: We still need to set cuda_attr.device_count to 0 and return an error.  It seems better to have a single code path to setting the cuda_attr.device_count value and returning the error.  Otherwise, the code is duplicated and (IMHO) harder to understand.

sunkuamzn (approved):

PR #9232, opened 2023-08-20
prov/verbs: Fix memory leak when creating EQ with unsupported wait object

k0zmo:
Calling `fi_eq_open` with `wait_obj` equal to `FI_WAIT_SET`, `FI_WAIT_MUTEX_COND` or `FI_WAIT_YIELD` for verbs provider causes a memory leak as we don't release all the objects created prior to validating value of `wait_obj`. Instead of just changing a jump to `err3` this PR moves argument validation to be done sooner (similar to other providers).


shefty (approved):
Intel CI will fail currently, but problem is unrelated.  I will ignore the CI since this passed previously, and only the commit message was updated.  Thanks!

I'll also cherry-pick for v1.19.x and v1.18.x after merging.

PR #9236, opened 2023-08-21
include/fi_peer, efa, shm, sm2: update peer srx API to allow for peer modification of the rx_entry

### LINKED ISSUE #9227: prov/efa: cq data not available with FI_PEEK
**Describe the bug**
In MPICH, we transmit the source rank of pt2pt messages using CQ data when available. In testing MPICH+efa, we found that an `MPI_Probe` operation does not return the correct source rank in the MPI_Status object. The cause seems to be that the CQ data is not available for FI_PEEK operations processed in `util_srx_peek`.

**To Reproduce**
1. configure and build MPICH 4.1.2 with libfabric main branch or v1.19.0rc1. EDIT: adding additional configure flags to speed up build time:
```
./configure --with-device=ch4:ofi --with-libfabric=</path/to/install> --with-datatype-engine=dataloop --disable-cxx --disable-fortran
```
2. Build and run this MPI reproducer using MPICH+efa. This can be done on a single node with the environment variable `MPIR_CVAR_NOLOCAL=1`.
```
mpicc foo.c
MPIR_CVAR_NOLOCAL=1 mpiexec -n 2 ./a.out
```

```c
#include <mpi.h>
#include <assert.h>
#include <stdlib.h>

int main(void) {
  MPI_Init(NULL, NULL);

  int rank;
  MPI_Comm_rank(MPI_COMM_WORLD, &rank);

  if (rank == 1) {
    MPI_Send(NULL, 0, MPI_DATATYPE_NULL, 0, 0, MPI_COMM_WORLD);
  }

  if (rank == 0) {
    MPI_Status status;
    MPI_Probe(MPI_ANY_SOURCE, 0, MPI_COMM_WORLD, &status);
    assert(status.MPI_SOURCE == 1);
    MPI_Recv(NULL, 0, MPI_DATATYPE_NULL, MPI_ANY_SOURCE, 0, MPI_COMM_WORLD, &status);
    assert(status.MPI_SOURCE == 1);
  }

  MPI_Finalize();
  return 0;
}
```

**Expected behavior**
The above reproducer should complete normally without triggering either assertion.

**Output**
With `FI_LOG_LEVEL=debug`, we see this output directly before the assert failure:
```
libfabric:52078:1692287803::efa:ep_data:efa_rdm_msg_generic_recv():926<debug> efa_rdm_msg_generic_recv: iov_len: 0 tag: 0 ignore: 60000000000000 op: 1 flags: 1080000
libfabric:52078:1692287803::core:ep_ctrl:util_srx_peek():620<debug> Message not found
libfabric:52079:1692287803::efa:ep_data:efa_rdm_msg_generic_recv():926<debug> efa_rdm_msg_generic_recv: iov_len: 0 tag: 0 ignore: 6000007fffffff op: 1 flags: 1080000
libfabric:52079:1692287803::core:ep_ctrl:util_srx_peek():620<debug> Message not found
libfabric:52078:1692287803::efa:ep_data:efa_rdm_msg_generic_recv():926<debug> efa_rdm_msg_generic_recv: iov_len: 0 tag: 0 ignore: 60000000000000 op: 1 flags: 1080000
libfabric:52078:1692287803::core:ep_ctrl:util_srx_peek():623<debug> Message found
a.out: foo.c:18: main: Assertion `status.MPI_SOURCE == 1' failed.
```

**Environment:**
provider: efa, endpoint type: rdm

**Additional context**
cc: @shijin-aws @aingerson

aingerson:
Fixes #9227

### DIFF
--- a/include/rdma/providers/fi_peer.h
+++ b/include/rdma/providers/fi_peer.h
@@ -167,6 +167,7 @@ struct fi_peer_rx_entry {
 	fi_addr_t addr;
 	size_t size;
 	uint64_t tag;
+	uint64_t cq_data;
 	uint64_t flags;
 	void *context;
 	size_t count;
@@ -181,7 +182,7 @@ struct fi_ops_srx_owner {
 	int	(*get_msg)(struct fid_peer_srx *srx, fi_addr_t addr,
 			size_t size, struct fi_peer_rx_entry **entry);
 	int	(*get_tag)(struct fid_peer_srx *srx, fi_addr_t addr,
-			size_t size, uint64_t tag, struct fi_peer_rx_entry **entry);
+			uint64_t tag, struct fi_peer_rx_entry **entry);
 	int	(*queue_msg)(struct fi_peer_rx_entry *entry);
 	int	(*queue_tag)(struct fi_peer_rx_entry *entry);
 	void	(*foreach_unspec_addr)(struct fid_peer_srx *srx,
--- a/man/fi_peer.3.md
+++ b/man/fi_peer.3.md
@@ -488,7 +488,7 @@ peer can initialize any extra available data before queuing the message with
 the owner. The peer_context and owner_context fields are only modifiable by the
 peer and owner, respectively, to store extra provider-specific information.
 
-## fi_ops_srx_owner::get_msg_entry() / get_tag_entry()
+## fi_ops_srx_owner::get_msg() / get_tag()
 
 These calls are invoked by the peer provider to obtain the receive buffer(s)
 where an incoming message should be placed.  The peer provider will pass in
@@ -533,7 +533,7 @@ entry.
 
 ## fi_ops_srx_peer::start_msg() / start_tag()
 
-These calls indicate that an asynchronous get_msg_entry() or get_tag_entry()
+These calls indicate that an asynchronous get_msg() or get_tag()
 has completed and a buffer is now available to receive the message.  Control
 of the fi_peer_rx_entry is returned to the peer provider and has been
 initialized for receiving the incoming message.
--- a/man/fi_peer.3.md
+++ b/man/fi_peer.3.md
@@ -418,6 +418,7 @@ struct fi_peer_rx_entry {
     fi_addr_t addr;
     size_t size;
     uint64_t tag;
+    uint64_t cq_data;
     uint64_t flags;
     void *context;
     size_t count;
@@ -482,10 +483,10 @@ the same lock, if needed.
 fi_peer_rx_entry defines a common receive entry for use between the owner and
 peer. The entry is allocated and set by the owner and passed between owner and
 peer to communicate details of the application-posted receive entry. All fields
-are only modifiable by the owner, except for the peer_context which is provided
-for the peer to use to save peer-specific information for unexpected message
-processing. Similarly, the owner_context can be used by the owner_context as
-needed for storing extra owner-specific information.
+are initialized by the owner, except in the unexpected message case where the
+peer can initialize any extra available data before queuing the message with
+the owner. The peer_context and owner_context fields are only modifiable by the
+peer and owner, respectively, to store extra provider-specific information.
 
 ## fi_ops_srx_owner::get_msg_entry() / get_tag_entry()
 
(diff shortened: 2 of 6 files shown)


aingerson:
@shijin-aws @shefty

shijin-aws:
Did I miss anything? I didn't see you modify util_srx_peek to write cq data to cq.

aingerson:
oh duh. no I didn't add that 🤦‍♀️

shijin-aws:
@aingerson Thanks, I will test this PR with the reproducer in https://github.com/ofiwg/libfabric/issues/9227. I think we also need to make rdm_tagged_peek has an option to check cq data. I can work on that in a separate PR.

shijin-aws:
I can confirm this PR fixes https://github.com/ofiwg/libfabric/issues/9227

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -854,13 +854,25 @@ struct efa_rdm_ope *efa_rdm_msg_alloc_rxe_for_tagrtm(struct efa_rdm_ep *ep,
...
 	int pkt_type;
+	struct efa_rdm_base_hdr *base_hdr;
+	uint64_t cq_data, cq_flags;
 
 	peer_srx = util_get_peer_srx(ep->peer_srx_ep);
 	data_size = efa_rdm_pke_get_rtm_msg_length(*pkt_entry_ptr);
 
+	base_hdr = efa_rdm_pke_get_base_hdr(*pkt_entry_ptr);
+	if (base_hdr->flags & EFA_RDM_REQ_OPT_CQ_DATA_HDR) {
+		cq_flags = FI_REMOTE_CQ_DATA;
```
shijin-aws, line 865: Is `FI_REMOTE_CQ_DATA` enough here? Do we need to |= `ofi_rx_cq_flags(ofi_op_tagged)` ?

aingerson: So far, the peer srx AP flags are only defined for FI_REMOTE_CQ_DATA so I think for now, yes. The peer utility saves the FI_TAGGED | FI_RECV flags in since it knows it is a tagged receive. We should think about more concretely defining the ownership of the flags but that would require extending the flags to the get_msg API as well

shijin-aws: >The peer utility saves the FI_TAGGED | FI_RECV flags in since it knows it is a tagged receive.

Which peer utility saves those flags?

aingerson: sorry I meant the util srx here in these changes does that

shijin-aws: I missed this change 

```
util_entry = util_init_unexp(srx_ctx, addr, size, tag, cq_data,
				     FI_TAGGED | FI_RECV | flags);
```
Should be good, thanks!

shijin-aws (approved):

include/rdma/providers/fi_peer.h
```diff
@@ -181,7 +182,8 @@ struct fi_ops_srx_owner {
 	int	(*get_msg)(struct fid_peer_srx *srx, fi_addr_t addr,
 			size_t size, struct fi_peer_rx_entry **entry);
 	int	(*get_tag)(struct fid_peer_srx *srx, fi_addr_t addr,
-			size_t size, uint64_t tag, struct fi_peer_rx_entry **entry);
+			size_t size, uint64_t tag, uint64_t cq_data,
+			uint64_t flags, struct fi_peer_rx_entry **entry);
```
shefty, line 186: cq_data and flags are completion data.  It seems odd to pass that into a call that is trying to retrieve the buffer.

I wonder if util_init_unexp() is being called from the correct side.  There's a mismatch here setting some of the fi_peer_rx_entry fields based on whether we matched a buffer or not that seems confusing.  The size, cq_data, and flags fields are ignored unless we don't find a match?

aingerson: yes I agree, it seems much. Right now the definition of the peer entry is that only the owner is allowed to touch the fields (hence this change). But the other alternative is to modify the definition so that the peer is allowed to initialize all the variables only in the unexpected case. I think I like that better. Thoughts?

shefty: Because we have a separate queue call, it should be okay for the peer to modify the structure prior to queuing it.

aingerson:
Alright, updated with a different modification of the peer API where we only pass in the necessary fields to match and then allow the peer to modify the fields with extra information

shefty (approved):
Thanks - need AWS to recheck the fix and EFA changes.

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -812,6 +813,14 @@ struct efa_rdm_ope *efa_rdm_msg_alloc_rxe_for_msgrtm(struct efa_rdm_ep *ep,
 			return NULL;
 		}
 		(*pkt_entry_ptr)->ope = rxe;
+
+		peer_rxe->size = data_size;
```
shijin-aws, line 817: This is non-tagged msg path. You should implement the change in efa_rdm_msg_alloc_rxe_for_tagrtm

aingerson: @shijin-aws Sorry about that! Updated.

shijin-aws (approved):
This revision passed the test for the original issue as well.

shijin-aws:
bot:aws:retest

shijin-aws:
ok to merge

man/fi_peer.3.md
```diff
@@ -482,32 +483,32 @@ the same lock, if needed.
 fi_peer_rx_entry defines a common receive entry for use between the owner and
 peer. The entry is allocated and set by the owner and passed between owner and
 peer to communicate details of the application-posted receive entry. All fields
-are only modifiable by the owner, except for the peer_context which is provided
-for the peer to use to save peer-specific information for unexpected message
-processing. Similarly, the owner_context can be used by the owner_context as
-needed for storing extra owner-specific information.
+are initialized by the owner, except in the unexpected message case where the
+peer can initialize any extra available data before queuing the message with
+the owner. The peer_context and owner_context fields are ony modifiable by the
```
raffenet, line 488:
```suggestion
the owner. The peer_context and owner_context fields are only modifiable by the
```

aingerson: Thanks! Updated

PR #9243, opened 2023-08-23
[v1.19.x] include/fi_peer: add cq_data to rx_entry, allow peer to modify on unexp

aingerson:
Switch the peer srx API to allow modification of the rx_entry on the unexpected path before queuing it with the owner. The peer is expected to fill in any available fields that may be returned with a subsequent FI_PEEK operation which reports back the size, CQ data, and flags of the operation.

This also updates the efa, shm, and sm2 providers to match the new definition.


shijin-aws:
bot:aws:retest

aingerson:
Intel CI failure is an intermittent ucx failure and is unrelated. I think this is ready to merge whenever.

PR #9247, opened 2023-08-23
prov/ucx: Initialize ep_flush to 1

zachdworkin:
Fixes intermittent segmentation fault with ucx on cleanup.

UCX requires ft_finalize because it passes in the ft_transmit_complete flag which forces the queue to be flushed. If the queue isn't flushed, it results in a segmentation fault because ucx marks it as still in use.

### DIFF
--- a/prov/ucx/src/ucx_init.c
+++ b/prov/ucx/src/ucx_init.c
@@ -40,7 +40,7 @@ struct ucx_global_descriptor ucx_descriptor = {
 	.use_ns = 0,
 	.ns_port = FI_UCX_DEFAULT_NS_PORT,
 	.localhost = NULL,
-	.ep_flush = 0,
+	.ep_flush = 1,
 	.check_req_leak = 0,
 };
 
@@ -291,7 +291,7 @@ static int ucx_getinfo(uint32_t version, const char *node,
 
 	status = fi_param_get(&ucx_prov, "ep_flush", &ucx_descriptor.ep_flush);
 	if (status != FI_SUCCESS)
-		ucx_descriptor.ep_flush = 0;
+		ucx_descriptor.ep_flush = 1;
 
 	status = fi_param_get(&ucx_prov, "check_req_leak", &ucx_descriptor.check_req_leak);
 	if (status != FI_SUCCESS)


j-xiong:
Thanks. Is it now working with ipv6 enabled?

zachdworkin:
@j-xiong it is working with ipv6

aingerson:
@j-xiong Zach and I were working on debugging this and I think this might help eliminate the issue but I'm not entirely sure why.
I was reading about reasons for segfaulting when destroying the AH and it says this:

It is up to the user waiting for all outstanding Send Requests, that were posted with this AH, to be completed or flushed before he destroys the AH. Not doing so, may lead to Work Completion errors or segmentation fault.

You have two options:
Wait until the outstanding Send Requests will be completed
Flush the Send Queue that those Send Requested were posted to


So I was thinking that is what's happening - there are outstanding requests that is preventing the AH from getting cleaned up, but looking at the test code, we don't sync until we've verified everything has been sent and received, so I think it does satisfy option 1. The code Zach pushed enforces it to satisfy option 2 as well so that's why I think it will help but I don't fully understand why the first isn't enough in this case. Do you have any idea?

j-xiong:
@aingerson I don't have a better theory for this. It seems that completion at the OFI level may still leave something behind for the ucp worker to clean up.

j-xiong (approved):

fabtests/functional/rdm_tagged_peek.c
```diff
@@ -278,15 +278,24 @@ static int do_recvs(void)
...
 	(void) ft_fill_buf(tx_buf, tx_size);
 
 	for (i = 1; i <= SEND_CNT; i++) {
 		do {
-			ret = fi_tsend(ep, tx_buf, tx_size, mr_desc,
-					remote_fi_addr, BASE_TAG + i,
-					&tx_ctx_arr[i].context);
+			tmsg.tag = BASE_TAG + i;
+			tmsg.context = &tx_ctx_arr[i].context;
+			ret = fi_tsendmsg(ep, &tmsg, FI_TRANSMIT_COMPLETE);
```
shefty, line 298: Changing a test that doesn't appear to be broken, so that a provider can be made to pass the test kind of defeats part of the purpose for having the tests...

This flag results in the ucx provider calling ucp_worker_flush().  That's also called when closing the ep.

shefty: The test causes a segfault.  That doesn't sound like a false positive, even if the crash isn't related to the PR.  The test can be disabled for ucx, if needed.  If the problem is outside of the provider, then disabling the test is reasonable.

aingerson: Yes, I agree that the test should be disabled if it's outside of the provider. I'm also suggesting a modification of the provider (see above) that might allow us to keep testing this for the provider. Do you or @j-xiong have thoughts about the above variable?

shefty: I don't know what ucp_worker_flush() does, but always calling it from ep close sounds reasonable.  (Remove the if statement).  The changes to the test result in that function being called, which suggests that there's something relative to ucp_worker_flush() being called that's needed.

aingerson:
Great. @zachdworkin Can you update this PR to initialize that variable to 1? @shefty Does that sound ok to you? That way we can turn it off easily if there is some reason that was added for some application somewhere

shefty:
Works for me.  We'll see if it works for the CI.

j-xiong (approved):

PR #9254, opened 2023-08-24
prov/efa: Check opt_len in efa_rdm_ep_getopt

shijin-aws:
Currently, efa_rdm_ep_getopt treats opt_len as an output only. But it should treat opt_len as an input/output and return -FI_ETOOSMALL when the inputted opt_len is not large enough. This patch fixes this issue.

Also added a unit test test_efa_rdm_ep_getopt to validate this fix.


prov/efa/test/efa_unit_test_ep.c
```diff
@@ -520,3 +520,43 @@ void test_efa_rdm_ep_atomic_without_caps(struct efa_resource **state)
...
+		FI_OPT_FI_HMEM_P2P,
+		FI_OPT_EFA_EMULATED_READ,
+		FI_OPT_EFA_EMULATED_WRITE,
+		FI_OPT_EFA_EMULATED_ATOMICS,
+		FI_OPT_EFA_USE_DEVICE_RDMA,
+		FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES,
+		FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES
+	};
+
+    efa_unit_test_resource_construct(resource, FI_EP_RDM);
```
wenduwan, line 547: Indent.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -520,3 +520,43 @@ void test_efa_rdm_ep_atomic_without_caps(struct efa_resource **state)
...
+		assert_int_equal(ret, -FI_ETOOSMALL);
+	}
+
+	/* use a large enough opt_len */
+	opt_len = 16;
+	for (i = 0; i < 9; i++) {
+		ret = fi_getopt(&resource->ep->fid, FI_OPT_ENDPOINT, opt_names[i], &opt_val, &opt_len);
+		assert_int_equal(ret, FI_SUCCESS);
+	}
+}
```
wenduwan, line 562: New line

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1358,38 +1358,56 @@ static int efa_rdm_ep_getopt(fid_t fid, int level, int optname, void *optval,
 
 	switch (optname) {
 	case FI_OPT_MIN_MULTI_RECV:
+		if (*optlen < sizeof(size_t))
+			return -FI_ETOOSMALL;
```
wenduwan, line 1362: This looks repetitive. How about adding another stack variable `size_t min_len;` and then
```
case ABC:
    min_len = sizeof(type);
    do_something_else;
    break
...

/* After switch cases */
if (*optlen < min_len)
    return -FI_ETOOSMALL;

...
```

shijin-aws: we should return early before do_something_else because the application may only allocate that size of memory for the opt_val. In this case we cannot defer the check after min_len is assigned.

shijin-aws: In theory you can still get min_size by another set of `switch case` at the beginning before the block of do something. But that is even worse in duplication

wenduwan: > application may only allocate that size of memory for the opt_val

So what happens if we write more data larger than the allocated size? I think it won't crash, does it?

prov/efa/test/efa_unit_test_ep.c
shijin-aws: fixed.

prov/efa/test/efa_unit_test_ep.c
shijin-aws: fixed

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: it will cause a crash if `opt_val` has 3 bytes allocated but it call fi_getopt for a opt_name of `int` (4 bytes) size and opt_len as 3. Here we should return early when checking opt_len.

bwbarrett: even if it's not an allocated buffer.  Let's take the case of an option that writes a int64_t sized value, but we call it this way:

```
int32_t optval
int32_t really_important_variable;
size_t optlen = sizeof(optval);
fi_getopt(...., &optval, &optlen);
```
If you write the int64_t sized data before you checked the input optlen, you probably just overwrote `really_important_variable`, and that's probably bad.

wenduwan: Cool that makes sense.

In that regard we can also be more defensive by only writing `MIN(*optlen, sizeof(actual_type))` to `optval`

wenduwan (approved):

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -520,3 +520,43 @@ void test_efa_rdm_ep_atomic_without_caps(struct efa_resource **state)
...
+		ret = fi_getopt(&resource->ep->fid, FI_OPT_ENDPOINT, opt_names[i], &opt_val, &opt_len);
+		assert_int_equal(ret, -FI_ETOOSMALL);
+	}
+
+	/* use a large enough opt_len */
+	opt_len = 16;
+	for (i = 0; i < 9; i++) {
+		ret = fi_getopt(&resource->ep->fid, FI_OPT_ENDPOINT, opt_names[i], &opt_val, &opt_len);
+		assert_int_equal(ret, FI_SUCCESS);
+	}
```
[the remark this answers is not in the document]
darrylabbate: A couple other changes:

- Calculate and store the array size instead of hardcoding the upper bound (`9`).
    ```c
	size_t num_opts = (sizeof opt_names) / (sizeof opt_names[0]);
    ```
- For both loops (especially the "successful" one), `opt_len` needs to be reset before each iteration. The only reason you didn't see a failure when testing the "successful" loop is because the sizes of each opt are already in descending order; `opt_len` was always sufficient in each successive call to `fi_getopt()`. But naturally this will become an issue if the order is ever changed.
    ```c
	for (i = 0; i < num_opts; i++) {
		/* use a too small opt_len */
		opt_len = 0;
		ret = fi_getopt(&resource->ep->fid, FI_OPT_ENDPOINT, opt_names[i], &opt_val, &opt_len);
		assert_int_equal(ret, -FI_ETOOSMALL);
	}

	for (i = 0; i < num_opts; i++) {
		/* use a large enough opt_len */
		opt_len = 16;
		ret = fi_getopt(&resource->ep->fid, FI_OPT_ENDPOINT, opt_names[i], &opt_val, &opt_len);
		assert_int_equal(ret, FI_SUCCESS);
	}
    ```

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1358,38 +1358,56 @@ static int efa_rdm_ep_getopt(fid_t fid, int level, int optname, void *optval,
 
 	switch (optname) {
 	case FI_OPT_MIN_MULTI_RECV:
+		if (*optlen < sizeof(size_t))
```
darrylabbate, line 1361: nit: I like `OFI_UNLIKELY()` for error checks.

darrylabbate (changes requested):

prov/efa/src/rdm/efa_rdm_ep_fiops.c
darrylabbate: > This looks repetitive.

Very solvable problem :)

```c
#define EFA_RDM_EP_GETOPT_CASE(opt, T, expr)	\
	case opt:				\
		if (*optlen < sizeof (T))	\
			return -FI_ETOOSMALL;	\
		*(T *) optval = (expr); 	\
		*optlen = sizeof (T);		\
		break;

// ...

	switch (optname) {
	EFA_RDM_EP_GETOPT_CASE(FI_OPT_MIN_MULTI_RECV, size_t, efa_rdm_ep->min_multi_recv_size)
	EFA_RDM_EP_GETOPT_CASE(FI_OPT_EFA_RNR_RETRY,  size_t, efa_rdm_ep->base_ep.rnr_retry)
	EFA_RDM_EP_GETOPT_CASE(FI_OPT_FI_HMEM_P2P,    int,    efa_rdm_ep->hmem_p2p_opt)
	// ...
	}
```

shijin-aws:
@darrylabbate I merged before you posted comments. I will could open a follow up PR to address your comment though

PR #9256, opened 2023-08-25
prov/efa: Fix and improve test_efa_rdm_ep_getopt

shijin-aws:
Currently, in test_efa_rdm_ep_getopt, the opt_len isn't reset after the fi_getopt call, which is a bug because the opt_len could be set by fi_getopt as an output value and violates the goal of the test.

Use num_opt_names to calculate the length of opt_names array instead of hardcoding it as 9.

Break the test_efa_rdm_ep_getopt into two tests:
1. test_efa_rdm_ep_getopt_undersized_optlen
2. test_efa_rdm_ep_getopt_oversized_optlen

Makes it easier to see what failed at a glance when running unit tests.


prov/efa/test/efa_unit_test_ep.c
```diff
@@ -543,20 +542,25 @@ void test_efa_rdm_ep_getopt(struct efa_resource **state)
...
+	size_t	num_opt_names = sizeof(opt_names) / sizeof(int);
 
 	efa_unit_test_resource_construct(resource, FI_EP_RDM);
 
-	/* use a too small opt_len */
-	opt_len = 0;
-	for (i = 0; i < 9; i++) {
-		ret = fi_getopt(&resource->ep->fid, FI_OPT_ENDPOINT, opt_names[i], &opt_val, &opt_len);
-		assert_int_equal(ret, -FI_ETOOSMALL);
+	for (i = 0; i < num_opt_names; i++) {
```
darrylabbate, line 549: Nit: `i < num_opt_names` is a comparison between a signed and unsigned integer type. Not necessarily an issue in this scenario (both values will be >= 0 for each iteration), but it's worth pointing out. Changing `i` to `size_t` is the easy fix here.

darrylabbate (approved):
Just one comment. lgtm otherwise, thanks

Ideally, the next logical step would be to actually assert that `optlen` is the correct value for each endpoint option when expected to succeed. Doesn't need to be part of this PR, though.

prov/efa/test/efa_unit_test_ep.c
shijin-aws: Fixed

shijin-aws:
>Ideally, the next logical step would be to actually assert that optlen is the correct value for each endpoint option when expected to succeed. Doesn't need to be part of this PR, though

Can be improved later

PR #9257, opened 2023-08-25
contrib/intel/jenkins: Quality of life updates to build and mpi-verbs

zachdworkin:
Imrove readability and understandability of how build.py works. It is now invoked with a full path and uses the dir command to make sure it starts in the same location every time. UCX run stage was also updated to mirror this.

MPI verbs-rxm was split into IMB and OSU stages to improve overall runtime of the pipeline.


nikhilnanal:
looks good to me!

PR #9260, opened 2023-08-28
configure.ac: Add configure flags for more sanitizers

darrylabbate:
This adds convenience flags for enabling [Leak](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dleak), [Thread](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dthread) and [UndefinedBehavior](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dundefined) sanitizers; on top of the already available [AddressSanitizer](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003daddress) flag.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,15 +155,23 @@ AS_IF([test x"$enable_debug" != x"no"],
 AC_DEFINE_UNQUOTED([ENABLE_DEBUG],[$dbg],
                    [defined to 1 if libfabric was configured with --enable-debug, 0 otherwise])
 
-AC_ARG_ENABLE([asan],
-	      [AS_HELP_STRING([--enable-asan],
-			      [Enable address sanitizer @<:@default=no@:>@])
-	      ],
-	      [],
-	      [enable_asan=no])
+AC_DEFUN([FI_ARG_ENABLE_SANITIZER],[
+        AC_ARG_ENABLE([$1],
+                      [AS_HELP_STRING([--enable-$1],
+                                      [Enable $3Sanitizer @<:@default=no@:>@])
+                      ],
+                      [],
+                      [enable_$1=no])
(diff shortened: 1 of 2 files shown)


shefty (approved):
Looks reasonable to me.  I'd like to have the function code indented, similar to how AC_DEFUN is written in other configure.m4 files.

darrylabbate:
> I'd like to have the function code indented, similar to how AC_DEFUN is written in other configure.m4 files.

I originally had everything indented, but the build checks were failing with `syntax error: unexpected end of file`. I'm definitely no m4 expert, but I believe the `AS_IF()` macro being indented was the problem (causing the whitespace to be retained).

```diff
-if test x"$enable_asan" != x"no"
+          if test x"$enable_asan" != x"no"
```

I'm sure there's a way to indent the `AS_IF()` without retaining the leading whitespace. I can take another look.

shefty:
FWIW, here's from prov/tcp/configure.m4
```
AC_DEFUN([FI_TCP_CONFIGURE],[
       # Determine if we can support the tcp provider
       xnet_h_happy=0
       AS_IF([test x"$enable_tcp" != x"no"], [xnet_h_happy=1])
       AS_IF([test $xnet_h_happy -eq 1], [$1], [$2])
])
```
The other configure.m4 files have something similar. The above file is pretty simple.  usnic and verbs are both more complex.

darrylabbate:
Indented the code inside `AC_DEFUN()` more similarly to the prov/tcp/configure.m4 example and the build checks seem to be happy now.

shefty (approved):
thanks!

PR #9262, opened 2023-08-28
prov/hook/trace: fixed trace log format on some attributes.

peinanz:
Add newline to start trace log for fi_ep_attr, fi_rx_attr and fi_tx_attr. This makes the logs in a consistent format for the attributes.

### DIFF
--- a/prov/hook/trace/src/hook_trace.c
+++ b/prov/hook/trace/src/hook_trace.c
@@ -95,11 +95,11 @@
 	do {	 \
 		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN,   \
 		    "%s %p context %p\n", name, ep, context);  \
-		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN, "%s",  \
+		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN, "\n%s",  \
 		    fi_tostr_r(buf, len, info->ep_attr, FI_TYPE_EP_ATTR));  \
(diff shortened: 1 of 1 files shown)


prov/hook/trace/src/hook_trace.c
```diff
@@ -95,11 +95,11 @@
 	do {	 \
 		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN,   \
 		    "%s %p context %p\n", name, ep, context);  \
-		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN, "%s",  \
+		FI_TRACE(dom->fabric->hprov, FI_LOG_DOMAIN, "\n%s",  \
```
shefty, line 98: Is there a reason why the newline is at the start of the output, rather than the end?

peinanz: The newline is added at the end of log prefix (pkg, pid, time, provname, ...), before the real attr.
The current log is:
libfabric:2634136:1692911343::tcp:domain:trace_endpoint():1256<trace> fi_ep_attr:
type: FI_EP_MSG
    protocol: FI_PROTO_SOCK_TCP
    protocol_version: 0
    ..
After the change, the log is:
libfabric:2634136:1692911343::tcp:domain:trace_endpoint():1256<trace>
fi_ep_attr:
type: FI_EP_MSG
    protocol: FI_PROTO_SOCK_TCP
    protocol_version: 0
    ...

This will be consistent with other attributes, e.g, fi_cq_attr. 
This will simplify paring the trace logs for test coverage.

shefty: Ah - makes sense.  Thanks!

PR #9263, opened 2023-08-28
contrib/intel/jenkins: Fix summary.py file not found error

zachdworkin:
Summary.py will fail with no file found if the opt-out stage forced the pipeline to not run. Opening the file in the -x option will fix this and create the file if it doesn't exist.

### DIFF
--- a/contrib/intel/jenkins/summary.py
+++ b/contrib/intel/jenkins/summary.py
@@ -931,6 +931,8 @@ if __name__ == "__main__":
 
     mpi_list = ['impi', 'mpich', 'ompi']
     log_dir = f'{cloudbees_config.install_dir}/{jobname}/{buildno}/log_dir'
+    if (not os.path.exists(log_dir)):
+        os.makedirs(log_dir)
 
     job_name = os.environ['JOB_NAME'].replace('/', '_')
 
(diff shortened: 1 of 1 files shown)


contrib/intel/jenkins/summary.py
```diff
@@ -943,8 +943,15 @@ def summarize_items(summary_item, logger, log_dir, mode):
 
 
     full_file_name = f'{log_dir}/{output_name}'
+    if (os.exists(full_file_name)):
+        open_opts = 'a'
+    else:
+        open_opts = 'w'
```
j-xiong, line 949: `open_opts` doesn't seem to be used.

zachdworkin: Thank you! I wrote out the inline thing before putting it into the with statement and forgot to remove it.

contrib/intel/jenkins/summary.py
```diff
@@ -941,8 +943,10 @@ def summarize_items(summary_item, logger, log_dir, mode):
     else:
         output_name = f'summary_{job_name}.log'
 
-
     full_file_name = f'{log_dir}/{output_name}'
+    if (not os.path.exists(full_file_name)):
+        with open(full_file_name, 'x') as output_file:
+            print(f"Writing summary to {full_file_name}")
```
j-xiong, line 949: Python open() with mode 'a' is supposed to create the file if it doesn't exist so this part is redundant.

zachdworkin: fixed, thanks!

j-xiong (approved):

PR #9271, opened 2023-08-30
prov/psm3: fix remaining size comparison for `MULTI_RECV`

thomasgillis:
This PR addresses an issue I have encountered with `FI_MULTI_RECV`: I never get the expected `FI_MULTI_RECV` flag in the cq when the buffer is up.
The issue seems to come from the missing equality sign (and the default of `min_buf_size = 0`):
```c
if (multi_recv_req->offset + PSMX3_STATUS_RCVLEN(req) + multi_recv_req->min_buf_size > multi_recv_req->len)
				flags |= FI_MULTI_RECV;	/* buffer used up */
```
where `>` should be `>=`.
The reason is that the default size is `0` and that condition should be triggered if we have reached the end of the buffer, so that the user can repost it (if I read the API correctly).

However by doing so, I also need to change the other comparison of the other remaining sizes.
The one following the initial change (L741) makes sense but I am unsure about the third one (L789).

### DIFF
--- a/prov/psm3/src/psmx3_cq.c
+++ b/prov/psm3/src/psmx3_cq.c
@@ -623,7 +623,7 @@ psmx3_mq_status_copy(struct psm2_mq_req_user *req, void *status_array, int entry
 			if (PSMX3_HAS_IMM(PSMX3_GET_FLAGS(PSMX3_STATUS_TAG(req))))
 				flags |= FI_REMOTE_CQ_DATA;
 			if (multi_recv_req->offset + PSMX3_STATUS_RCVLEN(req) +
-				multi_recv_req->min_buf_size > multi_recv_req->len)
+				multi_recv_req->min_buf_size >= multi_recv_req->len)
 				flags |= FI_MULTI_RECV;	/* buffer used up */
 			err = psmx3_cq_rx_complete(
 					status_data->poll_cq, ep->recv_cq, ep->av,
@@ -638,7 +638,7 @@ psmx3_mq_status_copy(struct psm2_mq_req_user *req, void *status_array, int entry
 		/* repost multi-recv buffer */
 		multi_recv_req->offset += PSMX3_STATUS_RCVLEN(req);
 		len_remaining = multi_recv_req->len - multi_recv_req->offset;
-		if (len_remaining >= multi_recv_req->min_buf_size) {
+		if (len_remaining > multi_recv_req->min_buf_size) {
 			if (len_remaining > PSMX3_MAX_MSG_SIZE)
 				len_remaining = PSMX3_MAX_MSG_SIZE;
 			err = psm3_mq_irecv2(ep->rx->psm2_mq,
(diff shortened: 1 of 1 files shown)


thomasgillis:
I don't have a code to share yet in order to reproduce, but it happens when you do the following
- send a multiple data of size `x`
- have a buffer that is a multiple of the data size `Cx`
- set the `FI_OPT_MIN_MULTI_RECV` to `0` or `x`
I will try to add a test to fabtest to trigger it but in the meantime if you can adapt one of yours that might be faster :-)

thomasgillis:
@acgoldma do you have any update on this one? :-)

thomasgillis:
I could get some errors in the fabtest:
```bash
[u100155@mel2011 bin]$ ./fi_multi_recv -S 8 & ./fi_multi_recv -S 8 localhost
[2] 24090
8       1k      7.8k        0.00s     11.20       0.71       1.40
8       1k      7.8k        0.00s      9.39       0.85       1.17
[2]+  Done                    ./fi_multi_recv -S 8
[u100155@mel2011 bin]$ export FI_PROVIDER=psm3
[u100155@mel2011 bin]$ ./fi_multi_recv -S 8 & ./fi_multi_recv -S 8 localhost
[2] 24094
[error] fabtests:functional/multi_recv.c:126: Received 127 completions per buffer, expected 64
8       1k      7.8k        0.00s      2.20       3.63       0.28
[2]+  Exit 5                  ./fi_multi_recv -S 8
```

thomasgillis:
I could get a failure with the `fabtest` but it's not solved by the patch:
```bash
export FI_PROVIDER=psm3
./fi_multi_recv -S 8 & ./fi_multi_recv -S 8 localhost
[2] 32577
[error] fabtests:functional/multi_recv.c:126: Received 127 completions per buffer, expected 64
8       1k      7.8k        0.00s      2.23       3.58       0.28
```

j-xiong:
@thomasgillis The patch works for `min_buf_size = 0` case but is actually incorrect when `min_buf_size` is not 0. when the remaining buffer is at least `min_buf_size`, the buffer should be enough to hold another message and FI_MULTI_RECV flag should not be set in the completion.  A fix like the following works for both cases:
```
buf_size_left = multi_recv_req->len - multi_recv_req->offset - PSMX3_STATUS_RCVLEN(req);
if (buf_size_left < multi_recv_req->min_buf_size  || buf_size_left == 0)
				flags |= FI_MULTI_RECV;	/* buffer used up */
```
However, this fix won't get fi_multi_recv test to pass. The reason is that the test posts two multi-recv buffers and counts the completions when a buffer is used up. For the psm3 provider, the consumption of the two buffers is interleaved instead of consecutive. So the counting is incorrect (`127, 1, 127, 1, ...` instead of `64, 64, 64, 64, ...`). More changes are needed to get this test pass.

j-xiong:
@thomasgillis Also setting `min_buf_size` to 0 is not the intended usage model for multi-recv. `min_buf_size` should be set to a value large enough to hold any imcoming message. For example, if posting a multi-recv buffer for 50 messages of size `x`, the total buffer size is `50x` and the minimal buf size should be `x`.

thomasgillis:
@j-xiong thanks for your help and review.
To my understanding, the value of `FI_OPT_MIN_MULTI_RECV` is the available space in the buffer when the buffer is released. So if the space is `<= min_buf_size` I would expect to get the cq entry. Therefore if I want to use the buffer entirely I would put 0.
From your explanation, I understand that the actual condition is `<` but then the phrasing of the API seems wrong to me:

> FI_OPT_MIN_MULTI_RECV - size_t
    Defines the minimum receive buffer space available when the receive buffer is released by the provider (see FI_MULTI_RECV). Modifying this value is only guaranteed to set the minimum buffer space needed on receives posted after the value has been changed. It is recommended that applications that want to override the default MIN_MULTI_RECV value set this option before enabling the corresponding endpoint.

j-xiong:
@thomasgillis Yes that page is not phrased accurately. I will update that. Thanks!

j-xiong:
Fixed in #9878

thomasgillis:
Thanks for the fix @j-xiong

PR #9280, opened 2023-09-04
fabtests/pytest/efa: add cuda memory marker

shijin-aws:
Add cuda memory marker to test_transfer_with_read_protocol_cuda so they can collected with other cuda tests in the same group.


shijin-aws:
bot:aws:retest

PR #9285, opened 2023-09-06
fabtests/component/dmabuf: Handle partial read scenario for fi_xe_rdmabw test

Juee14Desai:
For cross node tests, under certain conditions, the read() call may return with bytes read less than size, 
causing exchange_info() to return failure.
This fix would ensure that all the expected data are read.
This bug was found while running 2 node fi_xe_rdmabw test on CI.
Changes have been tested and verified both manually and on the CI.

### DIFF
--- a/fabtests/component/dmabuf-rdma/util.c
+++ b/fabtests/component/dmabuf-rdma/util.c
@@ -133,7 +133,7 @@ int exchange_info(int sockfd, size_t size, void *me, void *peer)
 		return -1;
 	}
 
-	if (read(sockfd, peer, size) != size) {
+	if (recv(sockfd, peer, size, MSG_WAITALL) != size) {
 		fprintf(stderr, "Failed to read peer info\n");
 		return -1;
 	}


fabtests/component/dmabuf-rdma/util.c
```diff
@@ -126,14 +126,32 @@ void sync_tcp(int sockfd)
 	(void)! read(sockfd, &dummy2, sizeof dummy2);
 }
 
+static int readFully(int fd, void *obuf, int len)
```
j-xiong, line 129: Please use naming convention consistent with the rest of the file.  For example, 'readFully' --> 'read_fully'. Same for the variable names.

fabtests/component/dmabuf-rdma/util.c
```diff
@@ -126,14 +126,32 @@ void sync_tcp(int sockfd)
...
+{
+	int bytesLeft = len;
+	char *buf = (char *) obuf;
+	int bytesRead = 0;
+
+	while (bytesLeft > 0 && (bytesRead = read(fd, (void *) buf,
```
j-xiong, line 135: Break the line after `&&` instead of `buf, `.

fabtests/component/dmabuf-rdma/util.c
```diff
@@ -126,14 +126,32 @@ void sync_tcp(int sockfd)
...
+
 int exchange_info(int sockfd, size_t size, void *me, void *peer)
 {
 	if (write(sockfd, me, size) != size) {
 		fprintf(stderr, "Failed to send local info\n");
 		return -1;
 	}
 
-	if (read(sockfd, peer, size) != size) {
+	if (readFully(sockfd, peer, size) != size) {
```
shefty, line 154: If this is a nonblocking socket exchanging data of a fixed size, use recvmsg(..., MSG_WAITALL).  Same applies to the write() call, which can write fewer bytes than size.

j-xiong: It is a blocking socket, but sometimes the read can still return before reading all the data.

shefty: Sorry, if this is nonblocking, write is fine.  You can also just use recv() instead of recvmsg().

j-xiong: Agree that a similar call for write is also preferred, even though the error hasn't been encountered so far.

shefty: And in my comments above, I meant *blocking* not nonblocking... sheesh, I need sleep.  It's the MSG_WAITALL flag that's needed for the recv() call.

j-xiong: @Juee14Desai Please try the recvmsg() approach first to see if it fixes the issue.

Juee14Desai: I made the recv() changes. CI is passing in my private build with this change.

shefty (approved):

j-xiong:
Nit-pick: the commit message `the fix would retry until ...` can be changed to `the fix would ensure ...` to reflect the most recent update.

Juee14Desai:
> Nit-pick: the commit message `the fix would retry until ...` can be changed to `the fix would ensure ...` to reflect the most recent update.

Done. CI has passed. If all looks good, this can be merged.

j-xiong:
@Juee14Desai I mean the message in the commit (which can be viewed by `git log`), not the description of the PR.

Juee14Desai:
> @Juee14Desai I mean the message in the commit (which can be viewed by `git log`), not the description of the PR.

Done. The checks and CI have triggered again though.

j-xiong (approved):

shefty:
skipping re-run of CI-- thanks!

PR #9288, opened 2023-09-07
[v1.18.x] contrib/intel/jenkins: Add two node dmabuf tests to CI

Juee14Desai:
This commit adds two node dmabuf tests to CI.


zachdworkin:
It looks like the log for this is messed up. Its using SHMEM log instead of DMABUF log. Can you look into why?

zachdworkin (approved):

PR #9294, opened 2023-09-07
[v1.18.x] fabtests/component/dmabuf: Handle partial read scenario for fi_xe_rdmabw test

Juee14Desai:
For cross node tests, under certain conditions, the read() call may return with bytes read less than size,
causing exchange_info() to return failure.
This fix would ensure that all the expected data are read.
This bug was found while running 2 node fi_xe_rdmabw test on CI.
Changes have been tested and verified both manually and on the CI.

This commit needs to be cherry-picked for v1.18.x for CI to work on this branch.
(cherry picked from commit 9e4fdbf)

### DIFF
--- a/fabtests/component/dmabuf-rdma/util.c
+++ b/fabtests/component/dmabuf-rdma/util.c
@@ -133,7 +133,7 @@ int exchange_info(int sockfd, size_t size, void *me, void *peer)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


j-xiong (approved):

j-xiong:
skipping appveyor ci which is not affected by this PR

PR #9298, opened 2023-09-07
[v1.19.x] contrib/intel/jenkins: Pick change from mpi tcp-rxm testing to just mpi-tcp

zachdworkin:


zachdworkin:
@nikhilnanal it looks like the CI runs didn't use the jenkinsfile in this PR. Can you use the replay function to make sure it runs with the changed jenkinsfile?

PR #9301, opened 2023-09-07
prov/efa: fix efa device name matching

wenduwan:
This patch fixes a bug in EFA device name matching logic, which only accounts for the first prefix match. As a result, efa_10-rdm domain name is matched with efa_1 device instead of efa_10.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -88,18 +88,28 @@ static struct fi_ops_domain efa_ops_domain_rdm = {
  *
  * @param efa_domain[in,out]	efa domain to be set.
  * @param domain_name		domain name
+ * @param ep_type		endpoint type
  * @return 0 if efa_domain->device and efa_domain->ibv_pd has been set successfully
  *         negative error code if err is encountered
  */
-static int efa_domain_init_device_and_pd(struct efa_domain *efa_domain, const char *domain_name)
+static int efa_domain_init_device_and_pd(struct efa_domain *efa_domain,
+                                         const char *domain_name,
+                                         enum fi_ep_type ep_type)
 {
 	int i;
+	char *device_name = NULL;
+	const char *domain_name_suffix = efa_domain_name_suffix(ep_type);
 
 	if (!domain_name)
 		return -FI_EINVAL;
 
 	for (i = 0; i < g_device_cnt; i++) {
-		if (strstr(domain_name, g_device_list[i].ibv_ctx->device->name) == domain_name) {
+		device_name = g_device_list[i].ibv_ctx->device->name;
+		if (strstr(domain_name, device_name) == domain_name &&
+		    strlen(domain_name) - strlen(device_name) ==
+		            strlen(domain_name_suffix) &&
+		    strcmp((const char *) (domain_name + strlen(device_name)),
+		           domain_name_suffix) == 0) {
 			efa_domain->device = &g_device_list[i];
 			break;
 		}
@@ -108,6 +118,7 @@ static int efa_domain_init_device_and_pd(struct efa_domain *efa_domain, const ch
 	if (i == g_device_cnt)
 		return -FI_ENODEV;
 
+	EFA_INFO(FI_LOG_DOMAIN, "Domain %s selected device %s\n", domain_name, device_name);
 	efa_domain->ibv_pd = efa_domain->device->ibv_pd;
 	return 0;
 }
@@ -223,7 +234,7 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		goto err_free;
 	}
 
-	err = efa_domain_init_device_and_pd(efa_domain, info->domain_attr->name);
+	err = efa_domain_init_device_and_pd(efa_domain, info->domain_attr->name, info->ep_attr->type);
 	if (err) {
 		ret = err;
 		goto err_free;


prov/efa/src/efa_domain.c
```diff
@@ -88,18 +88,25 @@ static struct fi_ops_domain efa_ops_domain_rdm = {
...
+	const char* domain_name_suffix = efa_domain_name_suffix(ep_type);
 
 	if (!domain_name)
 		return -FI_EINVAL;
 
 	for (i = 0; i < g_device_cnt; i++) {
-		if (strstr(domain_name, g_device_list[i].ibv_ctx->device->name) == domain_name) {
+		device_name = g_device_list[i].ibv_ctx->device->name;
+		if (strstr(domain_name, device_name) == domain_name &&
+		    strlen(domain_name) - strlen(device_name) == strlen(domain_name_suffix)) {
```
shijin-aws, line 109: can we do an exact match (using strcmp) between domain_name and device_name + domain_name_suffix?

wenduwan: I considered that option, but my understanding is that approach requires an extra memory allocation to construct device_name + domain_name_suffix.

From a correctness perspective I think the current approach is also reasonable, because we know domain_name = device_name + domain_name_suffix, and the `if` test verifies 2 things:
1. The domain name is prefixed by device name
2. The domain name length is the same as device name + suffix

So the only corner case would be `efa_10-rdm` is matched against `efa_1` when requested ep_type is DGRAM, i.e. DGRAM and RDM are used together, which should not happen at all.

I'm open to do the full string match if you think the tradeoff is worthwhile.

darrylabbate: The simpler solution is to `memcmp()` up to `domain_name_suffix[0]` (i.e. $n_{bytes} =$ `(ptrdiff_t) memchr(domain_name, *domain_name_suffix) - domain_name`. No extra null-terminated string required.

prov/efa/src/efa_domain.c
```diff
@@ -88,18 +88,26 @@ static struct fi_ops_domain efa_ops_domain_rdm = {
...
+	const char *domain_name_suffix = efa_domain_name_suffix(ep_type);
 
 	if (!domain_name)
 		return -FI_EINVAL;
 
 	for (i = 0; i < g_device_cnt; i++) {
-		if (strstr(domain_name, g_device_list[i].ibv_ctx->device->name) == domain_name) {
+		device_name = g_device_list[i].ibv_ctx->device->name;
+		if (strstr(domain_name, device_name) == domain_name &&
+		    strlen(domain_name) - strlen(device_name) ==
```
wzamazon, line 109: I think you can do 
```
		if (strstr(domain_name, device_name) == domain_name &&
		    domain_name[strlen[device_name]] == '-')
```

e.g. check

shijin-aws: This works but I feel it's still an ad-hoc check compared to an exact match.

wenduwan:
@darrylabbate @shijin-aws @wzamazon I added another condition to make it exact match without additional memcpy. Please review again.

prov/efa/src/efa_domain.c
```diff
@@ -88,18 +88,28 @@ static struct fi_ops_domain efa_ops_domain_rdm = {
...
+		            strlen(domain_name_suffix) &&
+		    strcmp((const char *) (domain_name + strlen(device_name)),
+		           domain_name_suffix) == 0) {
```
darrylabbate, line 112: The conditions inside the actual check here are a little wild. This is my suggestion:

Loop setup:

```c
// Get pointer to actual domain name suffix
const char *domain_name_suffix = memchr(domain_name, efa_domain_name_suffix(ep_type)[0], strlen(domain_name));
ptrdiff_t n_bytes = (ptrdiff_t) (domain_name_suffix - domain_name);
```

Actual check inside loop:

```c
// Short-circuits, avoids out-of-bounds access if device_name is too short
if (strlen(device_name) == n_bytes && memcmp(domain_name, device_name, n_bytes) == 0) {
  // ...
}
```

prov/efa/src/efa_domain.c
```diff
@@ -108,6 +118,7 @@ static int efa_domain_init_device_and_pd(struct efa_domain *efa_domain, const ch
 	if (i == g_device_cnt)
 		return -FI_ENODEV;
 
+	EFA_INFO(FI_LOG_DOMAIN, "Domain  %s selected device %s\n", domain_name, device_name);
```
darrylabbate, line 121: Superfluous whitespace?

```suggestion
	EFA_INFO(FI_LOG_DOMAIN, "Domain %s selected device %s\n", domain_name, device_name);
```

prov/efa/src/efa_domain.c
wenduwan: Overall I think this approach adds more complexity without real benefits. Especially `memchr` appears unnecessary since we want to do full string match.

The if block looks larger than it is because of the line length restrictions. It looks reasonable if we increase the line length:
```c
if (strstr(domain_name, device_name) == domain_name &&
    strlen(domain_name) - strlen(device_name) == strlen(domain_name_suffix) &&
    strcmp((const char *) (domain_name + strlen(device_name)), domain_name_suffix) == 0)
```

prov/efa/src/efa_domain.c
wenduwan: Thank you! Fixed.

prov/efa/src/efa_domain.c
shijin-aws: I am ok with the additional `strcmp((const char *) (domain_name + strlen(device_name)), domain_name_suffix) == 0`. This should make sure domain_name is a concatenation of device_name and domain_name_suffix

shijin-aws (approved):

shijin-aws:
need a backport

wenduwan:
@shijin-aws Ready to merge?

shijin-aws:
@darrylabbate do you have more comments?

darrylabbate:
If it fixes the device selection logic, it's fine with me

PR #9318, opened 2023-09-12
prov/efa: Add missing locks in efa_cntr_wait.

shijin-aws:
Currently, efa_cntr_wait call cntr->progress, which finally call efa_ep_progress_internal(). However,
efa_ep_progress_internal() must be called inside the srx->lock. This patch fixes this issue.

### DIFF
--- a/prov/efa/src/efa_cntr.c
+++ b/prov/efa/src/efa_cntr.c
@@ -44,6 +44,12 @@ static int efa_cntr_wait(struct fid_cntr *cntr_fid, uint64_t threshold, int time
 	int numtry = 5;
 	int tryid = 0;
 	int waitim = 1;
+	struct util_srx_ctx *srx_ctx;
+
+	srx_ctx = efa_cntr_get_srx_ctx(cntr_fid);
+
+	if (srx_ctx)
+		ofi_genlock_lock(srx_ctx->lock);
 
 	cntr = container_of(cntr_fid, struct util_cntr, cntr_fid);
 	assert(cntr->wait);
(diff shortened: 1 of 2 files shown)


prov/efa/src/efa_cntr.c
```diff
@@ -44,6 +44,13 @@ static int efa_cntr_wait(struct fid_cntr *cntr_fid, uint64_t threshold, int time
 	int numtry = 5;
 	int tryid = 0;
 	int waitim = 1;
+	struct util_srx_ctx *srx_ctx;
+	struct efa_cntr *efa_cntr;
+
+	efa_cntr = container_of(cntr_fid, struct efa_cntr, util_cntr.cntr_fid);
+	srx_ctx = efa_cntr->util_cntr.domain->srx->ep_fid.fid.context;
```
wenduwan, line 51: Could this be NULL?

shijin-aws: Good call, srx is only available for RDM ep type. I will add a check here (and all cntr ops)

shijin-aws: It seems we never support counter for DGRAM ep type. But I will still add such check for correctness.

shijin-aws: But I will still add such check for correctness

shijin-aws:
It seems before the RXR squash, we used to put cntr code only in `rxr/` directory. I am thinking whether it makes sense to keep this convention to rename `efa_cntr.c` to `efa_rdm_cntr.c` since it's not used by DGRAM code... @wzamazon WDYT?

shijin-aws:
Nvm, since we never forbid users to fi_cntr_open for a dgram EP type, I would just keep it general as is and add the check so the srx won't be accessed for DGRAM path

wenduwan:
bot:aws:retest

wenduwan (approved):

Issue #9009, opened 2023-06-09
libfabric-2.0: AV changes


shefty:
- Only support FI_AV_TABLE
- Remove support for async insertion

shefty:
verbs ud and ucx only support AV map

PR #9319, opened 2023-09-12
core: Support multiple auth keys per EP

iziemba:
EPs bound to an AV configure with FI_AUTH_KEY can support 1+ auth keys per EPs.

All eligible auth keys must be pre-inserted into the AV via fi_av_insert_auth_key(). Acceptable flags are the following:
- FI_TRANSMIT: Restrict the auth key to outbound data transfers. This includes send message, RMA, and atomic operations.
- FI_RECV: Restrict the auth key to inbound data transfers. This includes received messages and target MRs of RMA and atomic operations.

The FI_AUTH_KEY_MATCH_ALL fi_av_attr flag can be used to pre-configure the AV for all possible auth keys with a FI_TRANSMIT | FI_RECV configuration.

Once the AV is bound to an EP and the EP is successfully enabled, the EP will be configured to support all auth keys in the AV at that point in time. Later fi_av_insert_auth_key() will not propagate to already enabled EPs.

For AVs configured with FI_AUTH_KEY,
fi_av_insert_auth_key_{addr, svc, sym} must be used to generate fi_addr_t's for a specific <EP addr, auth key> tuple.
fi_av_insert_{addr, svc, sym} will not be supported. The resulting fi_addr_t's can be used for outgoing data transfers. If the endpoint is configured with FI_DIRECTED_RECV, the resulting fi_addr_t's can be used to restrict a receive buffer to a specific <EP addr, auth key> tuple. FI_ADDR_UNSEPC can be used to match any <EP addr, auth key> combination. An insert can be done with FI_AUTH_KEY_MATCH_ALL to generate an fi_addr_t to match on all EP addr and a specific auth key.

All inserted auth_keys must have buffer size equal to fi_info::domain_attr::auth_key_size reported by the provider. Failure to ensure this may lead to memory corruption.

For FI_EADDRNOTAVAIL CQ errors, the FI_SOURCE_ERR flag in fi_cq_err_entry::flags can be used to distinguish if fi_cq_err_entry::src_err_data or fi_cq_err_entry::err_data is valid. If set, fi_cq_err_entry::src_err_data is valid.

For EPs configured with FI_SOURCE_ERR and bound to an AV with FI_AUTH_KEY, all FI_EADDRNOTAVAIL CQ errors need to be generated with the FI_SOURCE_ERR flag set and fi_cq_err_entry::src_err_data filled in accordingly. fi_cq_err_source_err_data::addr and
fi_cq_err_source_err_data::auth_key will be set to valid pointers which can be used for fi_av_insert_auth_key_addr().

### DIFF
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -190,7 +190,7 @@ typedef struct fid *fid_t;
 /* Tagged messages, buffered receives, CQ flags */
 #define FI_CLAIM		(1ULL << 59)
 #define FI_DISCARD		(1ULL << 58)
-
+#define FI_AUTH_KEY		(1ULL << 57)
 
 struct fi_ioc {
 	void			*addr;
@@ -227,6 +227,7 @@ enum {
 #define FI_ADDR_NOTAVAIL	((uint64_t) -1)
 #define FI_KEY_NOTAVAIL		((uint64_t) -1)
 #define FI_SHARED_CONTEXT	SIZE_MAX
+#define FI_AV_AUTH_KEY		SIZE_MAX
 typedef uint64_t		fi_addr_t;
 
 enum fi_av_type {
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -449,6 +449,7 @@ struct fi_domain_attr {
 	size_t			max_err_data;
 	size_t			mr_cnt;
 	uint32_t		tclass;
+	size_t			max_ep_auth_key;
 };
 
 struct fi_fabric_attr {
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -392,6 +392,7 @@ struct fi_tx_attr {
 	size_t			iov_limit;
 	size_t			rma_iov_limit;
 	uint32_t		tclass;
+	uint64_t		optional_caps;
 };
 
 struct fi_rx_attr {
@@ -403,6 +404,7 @@ struct fi_rx_attr {
 	size_t			total_buffered_recv;
 	size_t			size;
 	size_t			iov_limit;
+	uint64_t		optional_caps;
 };
 
 struct fi_ep_attr {
@@ -450,6 +452,7 @@ struct fi_domain_attr {
 	size_t			mr_cnt;
 	uint32_t		tclass;
 	size_t			max_ep_auth_key;
+	uint64_t		optional_caps;
 };
 
 struct fi_fabric_attr {
@@ -476,6 +479,7 @@ struct fi_info {
 	struct fi_domain_attr	*domain_attr;
 	struct fi_fabric_attr	*fabric_attr;
 	struct fid_nic		*nic;
+	uint64_t		optional_caps;
 };
 
 struct fi_device_attr {
--- a/include/rdma/fi_domain.h
+++ b/include/rdma/fi_domain.h
@@ -102,6 +102,11 @@ struct fi_ops_av {
 			char *buf, size_t *len);
 	int	(*av_set)(struct fid_av *av, struct fi_av_set_attr *attr,
 			struct fid_av_set **av_set, void *context);
+	int	(*insert_auth_key)(struct fid_av *av, const void *auth_key,
+				   size_t auth_key_size, fi_addr_t *fi_addr,
+				   uint64_t flags);
+	int	(*lookup_auth_key)(struct fid_av *av, fi_addr_t fi_addr,
+				   void *auth_key, size_t *auth_key_size);
 };
 
 struct fid_av {
@@ -139,6 +144,11 @@ struct fi_mr_dmabuf {
 	int		fd;
 	uint64_t	offset;
 	size_t		len;
+};
+
+struct fi_mr_auth_key {
+	struct fid_av		*av;
+	fi_addr_t		src_addr;
 };
 
 struct fi_mr_attr {
@@ -530,6 +540,24 @@ static inline const char *
 fi_av_straddr(struct fid_av *av, const void *addr, char *buf, size_t *len)
 {
 	return av->ops->straddr(av, addr, buf, len);
+}
+
+static inline int
+fi_av_insert_auth_key(struct fid_av *av, const void *auth_key,
+		      size_t auth_key_size, fi_addr_t *fi_addr, uint64_t flags)
+{
+	return FI_CHECK_OP(av->ops, struct fi_ops_av, insert_auth_key) ?
+		av->ops->insert_auth_key(av, auth_key, auth_key_size, fi_addr,
+					 flags) : -FI_ENOSYS;
+}
+
+static inline int
+fi_av_lookup_auth_key(struct fid_av *av, fi_addr_t addr, void *auth_key,
+		      size_t *auth_key_size)
+{
+	return FI_CHECK_OP(av->ops, struct fi_ops_av, lookup_auth_key) ?
+		av->ops->lookup_auth_key(av, addr, auth_key, auth_key_size) :
+		-FI_ENOSYS;
 }
 
 static inline fi_addr_t
--- a/include/rdma/fi_domain.h
+++ b/include/rdma/fi_domain.h
@@ -107,6 +107,8 @@ struct fi_ops_av {
 				   uint64_t flags);
 	int	(*lookup_auth_key)(struct fid_av *av, fi_addr_t fi_addr,
 				   void *auth_key, size_t *auth_key_size);
+	int	(*set_user_id)(struct fid_av *av, fi_addr_t fi_addr,
+			       fi_addr_t user_id);
 };
 
 struct fid_av {
@@ -558,6 +560,13 @@ fi_av_lookup_auth_key(struct fid_av *av, fi_addr_t addr, void *auth_key,
 	return FI_CHECK_OP(av->ops, struct fi_ops_av, lookup_auth_key) ?
 		av->ops->lookup_auth_key(av, addr, auth_key, auth_key_size) :
 		-FI_ENOSYS;
+}
+
+static inline int
+fi_av_set_user_id(struct fid_av *av, fi_addr_t fi_addr, fi_addr_t user_id)
+{
+	return FI_CHECK_OP(av->ops, struct fi_ops_av, set_user_id) ?
+		av->ops->set_user_id(av, fi_addr, user_id) : -FI_ENOSYS;
 }
 
 static inline fi_addr_t
--- a/include/rdma/fi_eq.h
+++ b/include/rdma/fi_eq.h
@@ -239,6 +239,7 @@ struct fi_cq_err_entry {
 	/* err_data is available until the next time the CQ is read */
 	void			*err_data;
 	size_t			err_data_size;
+	fi_addr_t		src_addr;
 };
 
 enum fi_cq_wait_cond {
@@ -403,6 +404,9 @@ fi_cq_readfrom(struct fid_cq *cq, void *buf, size_t count, fi_addr_t *src_addr)
 static inline ssize_t
 fi_cq_readerr(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t flags)
 {
+	/* For compatibility with older providers. */
+	if (buf)
+		buf->src_addr = FI_ADDR_NOTAVAIL;
 	return cq->ops->readerr(cq, buf, flags);
 }
 
--- a/man/fi_av.3.md
+++ b/man/fi_av.3.md
@@ -70,6 +70,9 @@ int fi_av_insert_auth_key(struct fid_av *av, const void *auth_key,
 
 int fi_av_lookup_auth_key(struct fid_av *av, fi_addr_t addr,
       void *auth_key, size_t *auth_key_size);
+
+int fi_av_set_user_id(struct fid_av *av, fi_addr_t fi_addr,
+      fi_addr_t user_id);
 ```
 
 # ARGUMENTS
@@ -116,6 +119,10 @@ int fi_av_lookup_auth_key(struct fid_av *av, fi_addr_t addr,
 *auth_key_size*
 : On input, specifies size of auth_key buffer.  On output, stores number
   of bytes written to auth_key buffer.
+
+*user_id*
+: For address vectors configured with FI_AV_USER_ID, this defines the
+  the user ID to be associated with a specific fi_addr_t.
 
 # DESCRIPTION
 
@@ -285,6 +292,14 @@ struct fi_av_attr {
   targets distributed applications on large fabrics and allows for
   highly-optimized storage of remote endpoint addressing.
 
+- *FI_AV_USER_ID*
+: Indicates that the user will be associating user-defined IDs with
+  address vector. Specifying this flag during address vector creation
+  allows providers to optimize the address vector accordingly.
+
+  If users require FI_AV_AUTH_KEY and FI_AV_USER_ID support, using the
+  FI_AV_USER_ID flag per insert is not supported.
+
 ## fi_close
 
 The fi_close call is used to release all resources associated with an
@@ -363,6 +378,11 @@ used as the target for local data transfer operations.
 If the endpoint supports `FI_DIRECTED_RECV`, these fi_addr_t's can be used to
 restrict receive buffers to a specific endpoint address and authorization key.
 
+For address vectors configured with FI_AV_USER_ID, all subsequent target events
+corresponding to the address being inserted will return FI_ADDR_NOTAVAIL until
+the user defines a user ID for this fi_addr_t. This is done by using
+fi_av_set_user_id.
+
 *flags*
 : The following flag may be passed to AV insertion calls: fi_av_insert,
   fi_av_insertsvc, or fi_av_insertsym.
@@ -387,10 +407,16 @@ restrict receive buffers to a specific endpoint address and authorization key.
   will be updated to 0.  Failures will contain a fabric errno code.
 
 - *FI_AV_USER_ID*
-: This flag associates a user-assigned identifier with each AV entry
-  that is returned with any completion entry in place of the AV's address.
+: For address vectors configured without the FI_AV_USER_ID specified,
+  this flag associates a user-assigned identifier with each AV entry that
+  is returned with any completion entry in place of the AV's address.
+  Requesting this flag during insert may result in unexpected runtime
+  failures should a provider not support FI_AV_USER_ID with insert.
 
   This flag is invalid with FI_AV_AUTH_KEY.
+
+  With libfabric 1.20, users are encouraged to specify the FI_AV_USER_ID
+  when opening an address vector and use fi_av_set_user_id.
 
 - *FI_AUTH_KEY*
 : Denotes that the address being inserted should be associated with the
@@ -555,6 +581,12 @@ is larger than what can fit into the buffer, it will be truncated.  On
 output, auth_key_size is set to the size of the buffer needed to store the
 authorization key, which may be larger than the input value.
 
+## fi_av_set_user_id
+
+If the address vector has been opened with FI_AV_USER_ID, this function
+defines the user ID for a specific fi_addr_t. By default, all fi_addr_t's
+will be assigned the user ID FI_ADDR_NOTAVAIL.
+
 # NOTES
 
 An AV should only store a single instance of an address.
@@ -594,15 +626,19 @@ providers that do not support FI_AV_USER_ID, users may be able to trade
 off lookup processing with protocol overhead, by carrying source
 identification within a message header.
 
-User-specified fi_addr_t values are provided as part of address insertion
-(e.g. fi_av_insert) through the fi_addr parameter.  The fi_addr parameter
-acts as input/output in this case.  When the FI_AV_USER_ID flag is passed
-to any of the insert calls, the caller must specify an fi_addr_t identifier
-value to associate with each address.  The provider will record that
-identifier and use it where required as part of any completion event.  Note
-that the output from the AV insertion call is unchanged.  The provider will
-return an fi_addr_t value that maps to each address, and that value must be
-used for all data transfer operations.
+For address vectors opened without FI_AV_USER_ID, user-specified fi_addr_t
+values are provided as part of address insertion (e.g. fi_av_insert)
+through the fi_addr parameter. The fi_addr parameter acts as input/output
+in this case.  When the FI_AV_USER_ID flag is passed to any of the insert
+calls, the caller must specify an fi_addr_t identifier value to associate
+with each address.  The provider will record that identifier and use it
+where required as part of any completion event.  Note that the output from
+the AV insertion call is unchanged.  The provider will return an fi_addr_t
+value that maps to each address, and that value must be used for all data
+transfer operations.
+
+For address vectors opened with FI_AV_USER_ID, fi_av_set_user_id is used
+to defined the user-specified fi_addr_t.
 
 # RETURN VALUES
 
--- a/man/fi_av.3.md
+++ b/man/fi_av.3.md
@@ -24,6 +24,12 @@ fi_av_lookup
 fi_av_straddr
 : Convert an address into a printable string.
 
+fi_av_insert_auth_key
+: Insert an authorization key into the address vector.
+
+fi_av_lookup_auth_key
+: Retrieve an authorization key stored in the address vector.
+
 # SYNOPSIS
 
 ```c
@@ -58,6 +64,12 @@ fi_addr_t fi_rx_addr(fi_addr_t fi_addr, int rx_index,
 
 const char * fi_av_straddr(struct fid_av *av, const void *addr,
       char *buf, size_t *len);
+
+int fi_av_insert_auth_key(struct fid_av *av, const void *auth_key,
+      size_t auth_key_size, fi_addr_t *fi_addr, uint64_t flags);
+
+int fi_av_lookup_auth_key(struct fid_av *av, fi_addr_t addr,
+      void *auth_key, size_t *auth_key_size);
 ```
 
 # ARGUMENTS
@@ -96,6 +108,14 @@ const char * fi_av_straddr(struct fid_av *av, const void *addr,
 
 *flags*
 : Additional flags to apply to the operation.
+
+*auth_key*
+: Buffer containing authorization key to be inserted into the address
+  vector.
+
+*auth_key_size*
+: On input, specifies size of auth_key buffer.  On output, stores number
+  of bytes written to auth_key buffer.
 
 # DESCRIPTION
 
@@ -328,6 +348,21 @@ that calls to fi_av_insert following a call to fi_av_remove always reference a
 valid buffer in the fi_addr parameter.  Otherwise it may be difficult to
 determine what the next assigned index will be.
 
+If the address vector is configured with authorization keys, the fi_addr
+parameter may be used as input to define the authorization keys associated
+with the endpoint addresses being inserted. This is done by setting the fi_addr to an
+authorization key fi_addr_t generated from generated from `fi_av_insert_auth_key`
+and setting the FI_AUTH_KEY flag. If the FI_AUTH_KEY flag is not set, addresses
+being inserted will not be associated with any authorization keys. Whether or not
+an address can be disassociated with an authorization key is provider
+specific. If a provider cannot support this disassociation, an error will be returned.
+Upon successful insert with FI_AUTH_KEY flag, the returned fi_addr_t's will map to
+endpoint address against the specified authorization keys. These fi_addr_t's can be
+used as the target for local data transfer operations.
+
+If the endpoint supports `FI_DIRECTED_RECV`, these fi_addr_t's can be used to
+restrict receive buffers to a specific endpoint address and authorization key.
+
 *flags*
 : The following flag may be passed to AV insertion calls: fi_av_insert,
   fi_av_insertsvc, or fi_av_insertsym.
@@ -354,6 +389,13 @@ determine what the next assigned index will be.
 - *FI_AV_USER_ID*
 : This flag associates a user-assigned identifier with each AV entry
   that is returned with any completion entry in place of the AV's address.
+
+  This flag is invalid with FI_AV_AUTH_KEY.
+
+- *FI_AUTH_KEY*
+: Denotes that the address being inserted should be associated with the
+  passed in authorization key fi_addr_t.
+
   See the user ID section below.
 
 ## fi_av_insertsvc
@@ -414,7 +456,17 @@ accessed.  Inserted addresses are not required to be removed.
 fi_av_close will automatically cleanup any resources associated with
 addresses remaining in the AV when it is invoked.
 
-Flags are reserved for future use and must be 0.
+If the address being removed came from `fi_av_insert_auth_key`, the address
+will only be removed if all endpoints, which have been enabled against the
+corresponding authorization key, have been closed. If all endpoints are not
+closed, -FI_EBUSY will be returned. In addition, the FI_AUTH_KEY flag must be
+set when removing an authorization key fi_addr_t.
+
+*flags*
+: The following flags may be used when opening an AV.
+
+- *FI_AUTH_KEY*
+: Denotes that the fi_addr_t being removed is an authorization key fi_addr_t.
 
 ## fi_av_lookup
 
@@ -449,6 +501,59 @@ size of the buffer referenced by buf.  On output, addrlen is set to the
 size of the buffer needed to store the address.  This size may be
 larger than the input len.  If the provided buffer is too small, the
 results will be truncated.  fi_av_straddr returns a pointer to buf.
+
+## fi_av_insert_auth_key
+
+This function associates authorization keys with an address vector. This
+requires the domain to be opened with `FI_AV_AUTH_KEY`. `FI_AV_AUTH_KEY`
+enables endpoints and memory regions to be associated with authorization
+keys from the address vector. This behavior enables a single endpoint
+or memory region to be associated with multiple authorization keys.
+
+When endpoints or memory regions are enabled, they are configured with
+address vector authorization keys at that point in time. Later authorization
+key insertions will not propagate to already enabled endpoints and memory
+regions.
+
+The `auth_key` and `auth_key_size` parameters are used to input the
+authorization key into the address vector. The structure of the
+authorization key is provider specific.
+
+The output of `fi_av_insert_auth_key` is an authorization key fi_addr_t
+handle representing all endpoint addresses against this specific
+authorization key. For all operations, including address vector, memory
+registration, and data transfers, which may accept an authorization key
+fi_addr_t as input, the FI_AUTH_KEY flag must be specified. Otherwise,
+the fi_addr_t will be treated as an fi_addr_t returned from the `fi_av_insert`
+and related functions.
+
+For endpoints enabled with FI_DIRECTED_RECV, authorization key fi_addr_t's
+can be used to restrict incoming messages to only endpoint addresses
+within the authorization key. This will require passing in the FI_AUTH_KEY
+flag to `fi_recvmsg` and `fi_trecvmsg`.
+
+For domains enabled with FI_DIRECTED_RECV, authorization key fi_addr_t's can
+be used to restrict memory region access to only endpoint addresses within the
+authorization key. This will require passing in the FI_AUTH_KEY flag to
+`fi_mr_regattr`.
+
+These authorization key fi_addr_t's can later be used an input for
+endpoint address insertion functions to generate an fi_addr_t for a
+specific endpoint address and authorization key. This will require passing in
+the FI_AUTH_KEY flag to `fi_av_insert` and related functions.
+
+Flags are reserved for future use and must be 0.
+
+## fi_av_lookup_auth_key
+
+This functions returns the authorization key associated with a fi_addr_t.
+Acceptable fi_addr_t's input are the output of `fi_av_insert_auth_key` and
+AV address insertion functions. The returned authorization key is in a
+provider specific format. On input, the auth_key_size parameter should
+indicate the size of the auth_key buffer. If the actual authorization key
+is larger than what can fit into the buffer, it will be truncated.  On
+output, auth_key_size is set to the size of the buffer needed to store the
+authorization key, which may be larger than the input value.
 
 # NOTES
 
@@ -501,26 +606,24 @@ used for all data transfer operations.
 
 # RETURN VALUES
 
-Insertion calls for an AV opened for synchronous operation will return
-the number of addresses that were successfully inserted.  In the case of
-failure, the return value will be less than the number of addresses that
-was specified.
+Insertion calls, excluding `fi_av_insert_auth_key`, for an AV opened for
+synchronous operation will return the number of addresses that were
+successfully inserted.  In the case of failure, the return value will be
+less than the number of addresses that was specified.
 
-Insertion calls for an AV opened for asynchronous operation (with FI_EVENT
-flag specified) will return 0 if the operation was successfully initiated.
-In the case of failure, a negative fabric errno will be returned.  Providers
-are allowed to abort insertion operations in the case of an error. Addresses
-that are not inserted because they were aborted will fail with an error code
-of FI_ECANCELED.
+Insertion calls, excluding `fi_av_insert_auth_key`, for an AV opened for
+asynchronous operation (with FI_EVENT flag specified) will return FI_SUCCESS
+if the operation was successfully initiated. In the case of failure, a
+negative fabric errno will be returned.  Providers are allowed to abort
+insertion operations in the case of an error. Addresses that are not inserted
+because they were aborted will fail with an error code of FI_ECANCELED.
 
 In both the synchronous and asynchronous modes of operation, the fi_addr
 buffer associated with a failed or aborted insertion will be set to
 FI_ADDR_NOTAVAIL.
 
-All other calls return 0 on success, or a negative value corresponding to
-fabric errno on error.
-Fabric errno values are defined in
-`rdma/fi_errno.h`.
+All other calls return FI_SUCCESS on success, or a negative value corresponding
+to fabric errno on error. Fabric errno values are defined in `rdma/fi_errno.h`.
 
 # SEE ALSO
 
--- a/man/fi_cq.3.md
+++ b/man/fi_cq.3.md
@@ -414,6 +414,7 @@ struct fi_cq_err_entry {
 	int      prov_errno;  /* provider error code */
 	void    *err_data;    /*  error data */
 	size_t   err_data_size; /* size of err_data */
+	fi_addr_t src_addr; /* error source address */
 };
 ```
 
@@ -547,6 +548,10 @@ of these fields are the same for all CQ entry structure formats.
   subsequent read call against the CQ.  Applications must serialize access
   to the CQ when processing errors to ensure that the buffer referenced by
   err_data does not change.
+
+*src_addr*
+: Used to return source addressed related information for error events. How
+  this field is used is error event specific.
 
 # COMPLETION FLAGS
 
--- a/man/fi_cq.3.md
+++ b/man/fi_cq.3.md
@@ -448,6 +448,11 @@ Notable completion error codes are given below.
   passed directly into an fi_av_insert call to add the source address
   to the address vector.
 
+  For API versions 1.20 and later, if the EP is configured with
+  FI_AV_AUTH_KEY, src_addr will be set to the fi_addr_t authorization key
+  handle corresponding to the incoming data transfer. Otherwise, the
+  value will be set to FI_ADDR_NOTAVAIL.
+
 ## fi_cq_signal
 
 The fi_cq_signal call will unblock any thread waiting in fi_cq_sread
--- a/man/fi_domain.3.md
+++ b/man/fi_domain.3.md
@@ -710,6 +710,13 @@ The following are support primary capabilities:
 : When the domain is configured with FI_DIRECTED_RECV and FI_AV_AUTH_KEY,
   memory regions can be limited to specific authorization keys.
 
+*FI_AV_USER_ID*
+: Indicates that the domain supports the ability to open address vectors
+  with the FI_AV_USER_ID flag. If this domain capability is not set,
+  address vectors cannot be opened with FI_AV_USER_ID. Note that
+  FI_AV_USER_ID can still be support without this domain capability set.
+  See [`fi_av`(3)](fi_av.3.html).
+
 The following are supported secondary capabilities:
 
 *FI_LOCAL_COMM*
--- a/man/fi_domain.3.md
+++ b/man/fi_domain.3.md
@@ -214,6 +214,7 @@ struct fi_domain_attr {
 	size_t                max_err_data;
 	size_t                mr_cnt;
 	uint32_t              tclass;
+	size_t                max_ep_auth_key;
 };
 ```
 
@@ -771,6 +772,11 @@ provider to optimize any memory registration cache or lookup tables.
 This specifies the default traffic class that will be associated any endpoints
 created within the domain.  See [`fi_endpoint`(3)](fi_endpoint.3.html)
 for additional information.
+
+## Max Authorization Keys per Endpoint (max_ep_auth_key)
+
+: The maximum number of authorization keys which can be supported per connectionless
+  endpoint.
 
 # RETURN VALUE
 
--- a/man/fi_domain.3.md
+++ b/man/fi_domain.3.md
@@ -705,6 +705,13 @@ that a single memory registration operation may reference.
 Domain level capabilities.  Domain capabilities indicate domain
 level features that are supported by the provider.
 
+The following are support primary capabilities:
+*FI_DIRECTED_RECV*
+: When the domain is configured with FI_DIRECTED_RECV and FI_AV_AUTH_KEY,
+  memory regions can be limited to specific authorization keys.
+
+The following are supported secondary capabilities:
+
 *FI_LOCAL_COMM*
 : At a conceptual level, this field indicates that the underlying device
   supports loopback communication.  More specifically, this field
@@ -726,8 +733,7 @@ level features that are supported by the provider.
   feature.
 
 See [`fi_getinfo`(3)](fi_getinfo.3.html) for a discussion on primary versus
-secondary capabilities.  All domain capabilities are considered secondary
-capabilities.
+secondary capabilities.
 
 ## mode
 
@@ -744,6 +750,9 @@ The default authorization key to associate with endpoint and memory
 registrations created within the domain. This field is ignored unless the
 fabric is opened with API version 1.5 or greater.
 
+If domain auth_key_size is set to the value FI_AV_AUTH_KEY, auth_key must be
+NULL.
+
 ## Default authorization key length (auth_key_size)
 
 The length in bytes of the default authorization key for the domain. If set to 0,
@@ -751,6 +760,11 @@ then no authorization key will be associated with endpoints and memory
 registrations created within the domain unless specified in the endpoint or
 memory registration attributes. This field is ignored unless the fabric is
 opened with API version 1.5 or greater.
+
+If the size is set to the value FI_AV_AUTH_KEY, all endpoints and memory
+regions will be configured to use authorization keys associated with the AV.
+Providers which support authorization keys and connectionless endpoint must
+support this option.
 
 ## Max Error Data Size (max_err_data)
 : The maximum amount of error data, in bytes, that may be returned as part of
@@ -777,7 +791,7 @@ for additional information.
 ## Max Authorization Keys per Endpoint (max_ep_auth_key)
 
 : The maximum number of authorization keys which can be supported per connectionless
-  endpoint.
+  endpoint. If this value is non-zero, providers must support FI_AV_AUTH_KEY.
 
 ## Optional Capabilities (optional_caps)
 
--- a/man/fi_domain.3.md
+++ b/man/fi_domain.3.md
@@ -215,6 +215,7 @@ struct fi_domain_attr {
 	size_t                mr_cnt;
 	uint32_t              tclass;
 	size_t                max_ep_auth_key;
+	uint64_t              optional_caps;
 };
 ```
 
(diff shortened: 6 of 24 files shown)


jswaro:
What would be useful to this request is man page changes describing each API, and what it intends to accomplish from the application writer's and provider's perspectives.

include/rdma/fi_domain.h
```diff
@@ -102,6 +104,23 @@ struct fi_ops_av {
 			char *buf, size_t *len);
 	int	(*av_set)(struct fid_av *av, struct fi_av_set_attr *attr,
 			struct fid_av_set **av_set, void *context);
+	int	(*insert_auth_key)(struct fid_av *av, const void *auth_key,
+				   uint64_t flags, void *context);
+	int	(*lookup_auth_key)(struct fid_av *av, fi_addr_t addr,
+				   void *auth_key, size_t *auth_key_len);
```
shefty, line 110: If we go with the above changes, then lookup would take the key_index as input, not the fi_addr_t.  Requiring an fi_addr_t value here means that we can't actually lookup an auth_key until after it has been associated with some address, which seems restrictive.  This suggests an API:
```
(*lookup_auth_key)(struct fid_av *av, int key_index, void *auth_key, size_t auth_key_size);
```

shefty: Adding to the previous comment, having the ability to lookup the auth key given an fi_addr_t still seems useful.  But if we change lookup as suggested, then we either need another lookup call or a 2-step process.  I'm suggesting the latter:
```
(*lookup_key_index)(struct fid_av *av, fi_addr_t, int *key_index);
```
A lookup from an fi_addr_t would involve calling both lookup functions.

include/rdma/fi_domain.h
```diff
@@ -102,6 +104,23 @@ struct fi_ops_av {
...
+	int	(*insert_auth_key_addr)(struct fid_av *av, const void *addr,
+					size_t count, const void *auth_key,
+					fi_addr_t *fi_addr, uint64_t flags,
+					void *context);
```
shefty, line 114: With the above proposals, the insert calls would take ```int key_index``` as input instead ```const void *auth_key```.

include/rdma/fi_domain.h
```diff
@@ -102,6 +104,23 @@ struct fi_ops_av {
 			char *buf, size_t *len);
 	int	(*av_set)(struct fid_av *av, struct fi_av_set_attr *attr,
 			struct fid_av_set **av_set, void *context);
+	int	(*insert_auth_key)(struct fid_av *av, const void *auth_key,
+				   uint64_t flags, void *context);
```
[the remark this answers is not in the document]
shefty: Btw, I'm thinking whether we want to return an int to identify an auth_key, a new datatype (fi_auth_t), or reuse fi_addr_t.

The benefits of using fi_addr_t are:

- The caller automatically gets back an any-source address, which is restricted to this auth key.
- We can have a single lookup_auth_key() function (which would use your definition, not my proposed updated one)
- We could re-use the existing AV insert/insertsvc/insertsym calls by having fi_addr_t be an input/output parameter.
- We could also re-use the existing AV lookup/remove calls for handling auth_key's as well as addresses.

Basically, the API update could be limited to adding this call:
```
(*insert_auth_key)(struct fid_av *av, const void *auth_key, size_t auth_key_size,
    fi_addr_t *auth_addr, uint64_t flags);
```
And, honestly, not even that much is needed.  We can use the current insert call with a flag to indicate an auth_key is being inserted.  So, there's some appeal here, but also concern that it might be trying to be too clever.

Note that we can still have new static inline wrappers to make using the API easier.

shefty:
Thanks - seeing the API makes it easier to discuss and identify potential changes.

include/rdma/fi_domain.h
```diff
@@ -102,6 +104,23 @@ struct fi_ops_av {
...
+					void *context);
+	int	(*insert_auth_key_svc)(struct fid_av *av, const char *node,
+				       const char *service,
+				       const void *auth_key, fi_addr_t *fi_addr,
+				       uint64_t flags, void *context);
+	int	(*insert_auth_key_sym)(struct fid_av *av, const char *node,
+				       size_t nodecnt, const char *service,
+				       size_t svccnt, const void *auth_key,
+				       fi_addr_t *fi_addr, uint64_t flags,
+				       void *context);
```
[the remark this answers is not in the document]
iziemba: Either way works. Let me know what you think is best.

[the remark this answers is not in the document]
shefty: I think this comes down to how many new function calls we add, and if we want to assign an auth key its own fi_addr_t value.

include/rdma/fi_domain.h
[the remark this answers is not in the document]
shefty: My second comment above was the last one I wrote on the review.  The comments below reference my top comment.  It may help to read the second comment last, since it's presenting a possible alternative that popped into my head after writing the others.

include/rdma/fi_eq.h
```diff
@@ -237,7 +242,10 @@ struct fi_cq_err_entry {
 	int			err;
 	int			prov_errno;
 	/* err_data is available until the next time the CQ is read */
-	void			*err_data;
+	union {
+		void				 *err_data;
+		struct fi_cq_err_source_err_data *src_err_data;
+	};
```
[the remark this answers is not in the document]
iziemba: Yes. I need to fix this. I'll align better to current definition of err_data and err_data_size.

The following is what I was trying to accomplish. To support FI_SOURCE_ERR with these auth key AV extensions, both addr and auth_key need to be returned. Since the majority of the CQ error fields may be valid with FI_SOURCE_ERR, the user needs to provide an structure (e.g. fi_cq_err_source_err_data) where the provider can fill in the addr and auth_key fields.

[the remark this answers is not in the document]
iziemba: Here is what I am thinking.

```
struct fi_cq_err_entry {
	void			*op_context;
	uint64_t		flags;
	size_t			len;
	void			*buf;
	uint64_t		data;
	uint64_t		tag;
	size_t			olen;
	int			err;
	int			prov_errno;
	/* err_data is available until the next time the CQ is read */
	void			*err_data;
	size_t			err_data_size;
	void			*auth_key;
	size_t			auth_key_size;
};

For EPs configured with FI_SOURCE_ERR and bound to an AV with
FI_AUTH_KEY, when FI_EADDRNOTAVAIL CQ errors occur, users need to
provide a buffer via the fi_cq_err_entry::auth_key and
fi_cq_err_entry::auth_key_size fields to retrieve the auth_key for the
event. fi_cq_err_entry::err_data and fi_cq_err_entry:err_data_size will
be used, as normal, to return addr. The returned addr and auth_key can
be used for fi_av_insert_auth_key_addr().
```

include/rdma/fi_domain.h
iziemba: I think being able to lookup an auth_key from a fi_addr_t is required for DAOS to use this. CQ event with FI_SOURCE will return fi_addr_t encoding the EP addr and auth_key. DAOS servers need to serialize the the EP addr and auth_key when forwarding the RPC to other servers.

include/rdma/fi_domain.h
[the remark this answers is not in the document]
iziemba: >The caller automatically gets back an any-source address, which is restricted to this auth key.

This would remove the need to have an FI_AUTH_KEY_MATCH_ALL flag passed into fi_av_insert() to return a wildcarded EP addr fi_addr_t.

Let me update this proposal taking the fi_addr_t approach instead of the key index approach.

include/rdma/fi_domain.h
shefty: The point I was making is that with the proposal, there's isn't a way to lookup the auth_key *without* it being attached to an address.  The app has to keep the full auth_key around for later use.

iziemba:
Version 2 of the API is ready.

```
    core: Support multiple auth keys per EP

    FI_MULTI_AUTH_KEY is a new, domain-level, secondary capability flag used
    to denote if a provider supports multiple authorization keys per
    connectionless EP (flag does not apply to connected EPs). The number of
    authorization keys which can be supported will be reported via
    fi_ep_attr::auth_key_cnt.

    Applications which do not set FI_MULTI_AUTH_KEY will not see a change to
    their current usage of authorization keys.

    Applications which set FI_MULTI_AUTH_KEY must associate authorization
    keys with the AV being bound to the EP.
    fi_domain_attr::auth_key and fi_ep_attr::auth_key will be ignored.

    All eligible auth keys must be pre-inserted into the AV via
    fi_av_insert_auth_key(). Acceptable flags are the following:
    - FI_TRANSMIT: Restrict the auth key to outbound data transfers. This
      includes send message, RMA, and atomic operations.
    - FI_RECV: Restrict the auth key to inbound data transfers. This
      includes received messages and target MRs of RMA and atomic operations.

    fi_av_insert_auth_key() output is an fi_addr_t handle specific to this
    authorization key. If the EP is configured with FI_DIRECTED_RECV, this
    fi_addr_t can be used to match all EP addrs associated with this
    authorization key. Calling fi_av_remove() with this fi_addr_t will
    delete the authorization key. -FI_EBUSY will be returned from
    fi_av_remove() should this key still be used by en EP. In other words,
    all EPs using this authorization key need to be closed for
    fi_av_remove() to succeed.

    The FI_AUTH_KEY_MATCH_ALL fi_av_attr flag can be used to pre-configure
    the AV for all possible auth keys with a FI_TRANSMIT | FI_RECV
    configuration.

    Once the AV is bound to an EP and the EP is successfully enabled, the
    EP will be configured to support all auth keys in the AV at that point
    in time.

    If the domain is configured with FI_MULTI_AUTH_KEY, users must provide
    an authorization key fi_addr_t with fi_av_insert_{addr, svc, sym}. This
    is done by using the fi_addr as input. For fi_av_insert_{addr, sym},
    since fi_addr may be an array, only index 0 will be looked at for the
    authorization key fi_addr_t. That is only a single authorization key
    fi_addr_t will be supported for these functions. The output of
    fi_av_insert_{addr, svc, sym} is an fi_addr_t mapping to a specific
    <EP addr, auth_key> tuple.

    For FI_EADDRNOTAVAIL CQ errors, users must specify
    fi_cq_err_entry::auth_key and fi_cq_err_entry::auth_key_size to retrieve
    the authorization key. fi_cq_err_entry::auth_key and
    fi_cq_err_entry::auth_key_size behavior similar to
    fi_cq_err_entry::err_data and fi_cq_err_entry::err_data_size. That is
    users need to provide the buffer for providers to memcpy the
    authorization key into. fi_cq_err_entry::auth_key, combined with the
    existing behavior of fi_cq_err_entry::err_data, enables users to
    generate a fi_addr_t mapping to the specific <EP addr, auth_key> tuple
    which triggered the FI_EADDRNOTAVAIL event.
```

include/rdma/fi_domain.h
```diff
@@ -52,6 +52,7 @@ extern "C" {
 #define FI_SYMMETRIC		(1ULL << 59)
 #define FI_SYNC_ERR		(1ULL << 58)
 #define FI_UNIVERSE		(1ULL << 57)
+#define FI_AUTH_KEY_MATCH_ALL	(1ULL << 56)
```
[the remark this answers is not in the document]
iziemba: That's fair. Let me run this pass the DAOS/Mercury folks since this was a potential way to easy their integration with this API.  But, removing this I think opens up another optimization for FI_SOURCE_ERR and the fi_cq_err_entry (I'll start new comment for this).

include/rdma/fi_eq.h
```diff
@@ -239,6 +239,8 @@ struct fi_cq_err_entry {
 	/* err_data is available until the next time the CQ is read */
 	void			*err_data;
 	size_t			err_data_size;
+	void			*auth_key;
+	size_t			auth_key_size;
```
iziemba, line 243: If FI_AUTH_KEY_MATCH_ALL is removed (thus ensuring a fi_addr_t exists for each auth key), I am not sure if we need to have `fi_cq_err_entry::auth_key` and `fi_cq_err_entry::auth_key_size` as currently defined. Seems like we can just defined `fi_addr_t fi_cq_err_entry::auth_key` and return the `fi_addr_t` from insert. Thoughts?

shefty: I think you're right, and you're also exposing a gap in the current API.  We can't report the src_addr with any error completion, which may be useful.  For example, if we report a truncated receive completion at the target, we can't report the source of the message.

So... I like the change you're proposing, but I'd rename the 'auth_key' field to 'src_addr' to make it more generic.  The app would need to use the error code to know to interpret the src_addr as an auth_key only address.

include/rdma/fi_domain.h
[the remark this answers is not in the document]
iziemba: Synced with DAOS/Mercury team. Seems to be OK to remove magic insert flag.

include/rdma/fabric.h
```diff
@@ -418,6 +419,7 @@ struct fi_ep_attr {
 	size_t			rx_ctx_cnt;
 	size_t			auth_key_size;
 	uint8_t			*auth_key;
+	size_t			auth_key_cnt;
```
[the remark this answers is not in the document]
iziemba: I think we want to ensure that the existing API (auth per per domain/ep) works and require users to signal to the provider they are going to AV auth keys. Defining `ep_attr.auth_key_size = FI_AV_AUTH_KEY` seems like a good way to achieve this.

If a provider reports `domain_attr::max_ep_auth_key_cnt > 1`, I think the app can safely assume AV auth key is supported. If a provider reports `domain_attr::max_ep_auth_key_cnt == 1`, do we need a way to signal if FI_AV_AUTH_KEY is supported or not? Seems like if we do not not report this, apps could hit unexpected runtime failures.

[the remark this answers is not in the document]
shefty: I was thinking that we would require providers that support auth_keys to support the AV option.  If max_ep_auth_key_cnt = 1, AV auth_keys are slightly more restrictive than assigning an auth_key to an EP.

include/rdma/fi_eq.h
iziemba: Sounds good. Do you want me to open a new issue for truncated recvs not reporting fi_addr_t?

shefty: You can if you want; it seems like a gap that should be addressed, which is easier to handle with your proposal.

iziemba, line 243: Libfabric 2.0 thought. To maintain ABI compat, each provider needs to check API version to select the correct version of fi_cq_err_entry. I am finding myself updating multiple providers with this check. While not difficult, seems like if fi_cq_readerr() was an external symbol, a compat symbol could be defined to do this translation in a single place. Thoughts?

shefty: See ofi_cq_readerr() in util_cq.c.  Yes, if we change a data structure, we're forced to check the API version that the app has opened the fabric with to know the size of the structure that was passed in.

Since fi_cq_readerr() is not an external symbol, but a static inline, all pre-compiled apps would be using the cq ops function pointer and the original sized structure.  Adding a new external symbol doesn't help.

You may be able to avoid changing providers that use the util_cq, except for those that want to use the new functionality.  For example, you can add a new ofi_cq_write_error_trunc2() call that adds the src_addr.  ofi_cq_write_error_trunc() would call v2.  Existing providers are unchanged.  But ones that want to report the src_addr can use v2.  Obviously, the readerr() call still needs to handle things on the application side.

iziemba:
Version 3 of the API is ready. Since it seems like we are converging on something, I have added in markdown documentation along with ABI compat.

shefty:
Please separate the ABI breaking changes (modification to fi_domain_attr) from the other API change (fi_cq_err_entry) into their own commits.

include/rdma/fi_domain.h
```diff
@@ -102,6 +102,11 @@ struct fi_ops_av {
 			char *buf, size_t *len);
 	int	(*av_set)(struct fid_av *av, struct fi_av_set_attr *attr,
 			struct fid_av_set **av_set, void *context);
+	int	(*insert_auth_key)(struct fid_av *av, const void *auth_key,
+				   size_t auth_key_size, fi_addr_t *fi_addr,
+				   uint64_t flags);
+	int	(*lookup_auth_key)(struct fid_av *av, fi_addr_t fi_addr,
+				   void *auth_key, size_t *auth_key_size);
```
shefty, line 109: Please move these changes into yet a 3rd commit.

include/rdma/fi_eq.h
```diff
@@ -239,6 +239,7 @@ struct fi_cq_err_entry {
 	/* err_data is available until the next time the CQ is read */
 	void			*err_data;
 	size_t			err_data_size;
+	fi_addr_t		src_addr;
```
shefty, line 242: There's not a great place for this comment, so putting it here.  As an option, we can modify fi_cq_readerr() to:
```
static inline ssize_t
fi_cq_readerr(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t flags)
{
    /* For compatibility with older providers */
    buf->src_addr = FI_ADDR_NOTAVAIL;
    return cq->ops->readerr(cq, buf, flags);
}
```
If we go through this path, the app has recompiled with the updated structure.  This ensures that the src_addr has been set to a valid value for providers that aren't updated to set it directly.

This change may only be worth doing if it allows us to avoid updating all the providers.

man/fi_av.3.md
```diff
@@ -328,6 +348,16 @@ that calls to fi_av_insert following a call to fi_av_remove always reference a
...
 
+If the address vector is configured with authorization keys, the fi_addr
+parameter cannot be NULL and is used to define the authorization key associated
+with the address being inserted. Acceptable inputs are the fi_addr_t's generated
+from `fi_av_insert_auth_key`. Only `fi_addr[0]` can be used to pass in an
+authorization key fi_addr_t. Return returned fi_addr_t's will map to endpoint
+address against the specified authorization key. These fi_addr_t's can be
+used as the target for local data transfer operations. If the endpoint
+supports `FI_DIRECTED_RECV`, these fi_addr_t's can be used to restrict
+recieve buffers to a specific endpoint address and authorization key.
```
shefty, line 359: Hmm... "if the address vector is configured with authorization keys"...

That brings up that I think we'll want a way for the app to clearly indicate that it wants to use auth_keys, probably as part of AV creation.  Based on what's been discussed so far, setting the fi_domain_attr::auth_key_size == FI_AV_AUTH_KEY would work.  (As defined, we're determining this dynamically based on whether the app calls fi_av_insert_auth_key prior to fi_av_insert).

Adding that requirement basically removes the need to set the ep attribute

man/fi_av.3.md
```diff
@@ -450,6 +485,51 @@ size of the buffer needed to store the address.  This size may be
 larger than the input len.  If the provided buffer is too small, the
 results will be truncated.  fi_av_straddr returns a pointer to buf.
 
+## fi_av_insert_auth_key
+
+This function associates authorization keys with an address vector. This
+behavior, combined with endpoints opened with `FI_AV_AUTH_KEY``, enables an
```
shefty, line 491: See above comment on domain vs endpoint control of FI_AV_AUTH_KEY.

man/fi_cq.3.md
```diff
@@ -447,6 +448,11 @@ Notable completion error codes are given below.
   passed directly into an fi_av_insert call to add the source address
   to the address vector.
 
+  For API versions 1.20 and later, if the EP is configured with
+  FI_AV_AUTH_KEY, src_addr will be set to the fi_addr_t authorization key
+  handle corresponding to the incoming data transfer. Otherwise, the
+  value will be set to FI_ADDR_UNSPEC.
```
shefty, line 454: FI_ADDR_UNSPEC should be FI_ADDR_NOTAVAIL.  They're the same value, so it doesn't matter to code, but not available is more descriptive and aligns with the other CQ calls.

man/fi_domain.3.md
```diff
@@ -772,6 +773,12 @@ This specifies the default traffic class that will be associated any endpoints
 created within the domain.  See [`fi_endpoint`(3)](fi_endpoint.3.html)
 for additional information.
 
+## Max Authorization Keys per Endpoint (max_ep_auth_key_cnt)
+
+: The maximum number of authorization keys which can be supported per endpoint. If
+connectionless endpoints are supported and this value is non-zero, providers must
+support FI_AV_AUTH_KEY.
```
shefty, line 780: I think we can state that this value only applies to connectionless endpoints.

nit: indentation needed for lines after ':'

shefty (commented):
Some comments to fill your time. :)  Thanks!

man/fi_av.3.md
iziemba: Originally, I thought about fi_domain_attr::auth_key_size == FI_AV_AUTH_KEY. The challenge I ran into was how to support fi_mr_attr::auth_key and fi_mr_attr::auth_key_size. Specifically, the following behavior.

_Indicates the key to associate with this memory registration. Authorization keys are used to limit communication between endpoints. Only peer endpoints that are programmed to use the same authorization key may access the memory region. The domain authorization key will be used if the auth_key_size provided is 0. This field is ignored unless the fabric is opened with API version 1.5 or greater._

Thoughts on how to handle fi_mr_attr::auth_key_size?

shefty:
Targeting v1.20 looks doable.

man/fi_av.3.md
```diff
@@ -328,6 +348,16 @@ that calls to fi_av_insert following a call to fi_av_remove always reference a
 valid buffer in the fi_addr parameter.  Otherwise it may be difficult to
 determine what the next assigned index will be.
 
+If the address vector is configured with authorization keys, the fi_addr
```
iziemba, line 351: I think this behavior conflicts with FI_AV_USER_ID. We are planning to support FI_AV_USER_ID with FI_AV_AUTH_KEY.

iziemba: To handle this issue, maybe users need to provide the below context via the insert context arg.
```
struct fi_av_auth_key_context {
    fi_addr_t auth_key;
    void *context;
};
```
Thoughts?

man/fi_av.3.md
```diff
@@ -450,6 +485,51 @@ size of the buffer needed to store the address.  This size may be
...
+*flags*
+: The following flag may be passed to `fi_av_insert_auth_key`:
+
+- *FI_TRANSMIT*
+: Restrict the authorization key to outbound data transfers. This includes
+  send message, RMA, and atomic operations.
+
+- *FI_RECV*
+: Restrict the authorization key to inbound data transfers. This includes
+  received messages and target MRs of RMA and atomic operations.
```
shefty, line 520: If the access flags aren't immediately usable, I would leave flags as reserved for now.  The issues is there's no easy way to extend flags in the future to have an auth_key for message sends that's separate from that used for RMA writes, or an auth_key for message receives from remote RMA.

Is that separation useful?  I don't know.  Honestly it seems doubtful.  But if we can avoid limiting the use of flags until we have a use case, it would be better.

shefty (commented):
See comments.  I need to think about the comments posted to the last patch.

man/fi_av.3.md
shefty: Ugh...  Separately, FI_AV_USER_ID and FI_AV_AUTH_KEY seem to work well with the existing API.

Zeroth Issue: You're about to experience my thought process, starting with me inserting this issue above the 5 that I wrote below first.

First issue: FI_AV_USER_ID is not identified as either a primary or secondary capability.  Primary means the app has to request it.  Secondary means the provider can offer it even if not requested.  This particular bit hits me as a secondary capability.  That leads to:

Second issue: On AV creation, the provider doesn't know if it needs to support FI_AV_USER_ID.  The API kind of makes a mess of this.  FI_AV_USER_ID is specified during insert.  Meaning some addresses may have a user ID, while others won't.  That's not a big deal, since the provider could internally set the 'user ID' to the actual fi_addr_t value.  However, it would be easier if the provider knew that the AV would store user IDs.  Goto second issue.

Third issue: We could make FI_AV_USER_ID a primary capability.  That would let the provider know if user IDs were going to be used.  No, that's not the issue.  The issue is that the shm provider doesn't even set the capability bit.  It simply accepts the flag.  That aligns closer with a use case of FI_AV_USER_ID being a secondary capability.  I don't know how cxi treats this flag.

Fourth issue: Given the above, what I want is: 1. Define FI_AV_USER_ID as a secondary capability.  2. Specify FI_AV_USER_ID as a flag on AV creation.  3. Keeping the flag on insert would be kept for compatibility.  The issue is that this changes the API, which we aren't quite compliant with anyway.

Fifth issue: Even with the above changes, the original problem being posed is still unresolved.  All I did was raise 5 more issues, though counting this issue as a new issue is questionable.

I didn't count the Zeroth issue in the Fifth issue because I added the Zeroth issue between writing the fifth issue and this statement.

Where was I?  Oh, yes, getting paid by the word.

I can think of some options.  I don't like asking the user to cast fi_addr_t * to some other structure.  So, I'm going to rule that possibility out.  We could have the user pass in an array of structs through the context parameter.  That's usable if we're okay not supporting asynchronous insertions.  I propose that feature for removal in 2.0 (#9009).  Second option is to have the user pass in twice the number of fi_addr_t values as the count.  Half would be the auth_key, and half the user ID.  Whether we split the array or alternate values doesn't matter much.  Finally, we could add a new call to set the user ID and do that separately.

man/fi_av.3.md
shefty: I think I would go with:

fi_mr_attr::auth_key_size = sizeof(fi_addr_t)
fi_mr_attr::auth_key = fi_addr_t

Since the domain is set to FI_AV_AUTH_KEY, we can expect the fi_mr_attr fields to align.

iziemba: Since a domain could have multiple AVs, do we need a way to bind the MR to the AV?

man/fi_av.3.md
iziemba: >That aligns closer with a use case of FI_AV_USER_ID being a secondary capability. I don't know how cxi treats this flag.

We are still working on adding support for FI_AV_USER_ID. I think we would treat this as a secondary cap flag. If with this cap flag set at the AV level, do we want users to still have the flexibility about which inserts FI_AV_USER_ID will apply to?

If it helps simplify the problem, could we limit each insert call to only be associated with a single auth_key?

prov/bgq/src/fi_bgq_cq.c
```diff
@@ -186,7 +189,7 @@ fi_bgq_cq_readerr(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t flags
 		if (NULL == bgq_cq->err_head)
 			bgq_cq->err_tail = NULL;
 
-		*buf = ext->err_entry;
+		memcpy(buf, &ext->err_entry, size);
```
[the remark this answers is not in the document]
iziemba: `ext->err_entry` is of type `struct fi_cq_err_entry`. I'll change this to `ext->err_entry`.

man/fi_av.3.md
shefty: Defining FI_AV_USER_ID as a secondary capability doesn't work seamlessly.  That bit changes the behavior of the API, which the user has to agree with.  From the viewpoint of the user, I would consider introducing another flavor of capability bits -- optional bits.  These require the app to request, but may be cleared by the provider.  The main reason for adding this new flavor is so that app can avoid calling fi_getinfo multiple times toggling bits on/off in order to determine what features it can use.

The other proposed changes in issue 4 seem okay.  I'd rather configure the AV as either supporting user IDs or not.  Handling this per insertion call doesn't seem useful and just adds complexity.

As for the auth key, limiting insertion to a single key doesn't help.  Even if inserting a single address, the user needs to provide 2 fi_addr_t values.  I'm leaning toward adding a call to set/change the user ID for an inserted address.  There's a race between calling insert and setting the user ID where a completion could be reported.  But I think that race exists anyway, where a completion can be generated before the address has been inserted.

prov/psm2/src/psmx2_cq.c
```diff
@@ -1693,8 +1693,13 @@ STATIC ssize_t psmx2_cq_readerr(struct fid_cq *cq, struct fi_cq_err_entry *buf,
...
 			      fabric_fid.api_version;
-		size = FI_VERSION_GE(api_version, FI_VERSION(1, 5)) ?
-			sizeof(*buf) : sizeof(struct fi_cq_err_entry_1_0);
+
+		if (FI_VERSION_GE(api_version, FI_VERSION(1, 20)))
+			size = sizeof(*buf);
+		else if (FI_VERSION_GE(api_version, FI_VERSION(1, 5)))
+			size = sizeof(struct fi_cq_err_entry_1_1);
+		else
+			size = sizeof(struct fi_cq_err_entry_1_0);
```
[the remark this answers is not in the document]
iziemba: Yes. From what I have seen, providers typically embedded 'struct fi_cq_err_entry` in their structure. I'll make a util function for this.

src/fabric.c
```diff
@@ -987,7 +987,7 @@ void DEFAULT_SYMVER_PRE(fi_freeinfo)(struct fi_info *info)
 		free(info);
 	}
 }
-DEFAULT_SYMVER(fi_freeinfo_, fi_freeinfo, FABRIC_1.3);
+DEFAULT_SYMVER(fi_freeinfo_, fi_freeinfo, FABRIC_1.7);
```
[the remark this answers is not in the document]
iziemba: Will fix.

man/fi_av.3.md
iziemba: I'll update the PR with `struct fi_mr_auth_key`

man/fi_av.3.md
iziemba: >We could have the user pass in an array of structs through the context parameter. That's usable if we're okay not supporting asynchronous insertions.

As long as there is a `void * context` in the struct the user passes in, how does this limit async insertions?

man/fi_av.3.md
iziemba: Sounds good. So the default behavior would be to enable tx/rx on the authorization key.

man/fi_av.3.md
shefty: It's more that the API becomes difficult to describe and use.  The context param would need to point to something like:
```
struct blah blah {
    void *async_context;
    fi_addr_t *auth_key;  // or user ID
};
```
The size of auth_key array would come from the count param.  The auth_key could instead be auth_key[0], but that makes extending blah blah more difficult.

iziemba:
Another round is update. Only comment not address is how to handle FI_AV_USER_ID.

man/fi_mr.3.md
```diff
@@ -681,6 +698,22 @@ FI_HMEM_ZE interfaces.  The driver and device index values represent
...
+FI_AV_AUTH_KEY, av is used to define the fid_av which contains the authorization
+keys to be associated with the memory region. If the domain is also opened with
+FI_MR_ENDPOINT, the specified AV must be the same AV bound to the endpoint.
+
+By default, the memory region will be associated with all authorization keys
+in the AV.
+
+## key
+
+If the domain was opened with FI_DIRECTED_RECV, key can be used to limit the
```
iziemba, line 713: I tried to re-used the FI_DIRECTED_RECV for RMA/AMOs. For clarity, should a new macro be defined for directed one-sided operations?

man/fi_av.3.md
```diff
@@ -328,6 +348,16 @@ that calls to fi_av_insert following a call to fi_av_remove always reference a
...
 determine what the next assigned index will be.
 
+If the address vector is configured with authorization keys, the fi_addr
+parameter cannot be NULL and is used to define the authorization key associated
+with the address being inserted. Acceptable inputs are the fi_addr_t's generated
+from `fi_av_insert_auth_key`. Only `fi_addr[0]` can be used to pass in an
+authorization key fi_addr_t. Return returned fi_addr_t's will map to endpoint
+address against the specified authorization key. These fi_addr_t's can be
+used as the target for local data transfer operations. If the endpoint
+supports `FI_DIRECTED_RECV`, these fi_addr_t's can be used to restrict
```
iziemba, line 358: While I like the idea of reusing FI_DIRECTED_RECV to restrict recv buffers to a specific auth_key, I wonder if we need to restrict FI_DIRECTED_RECV to only be applicable if max_ep_auth_key_cnt <= 1 and define a new cap for this capability. Maybe FI_DIRECTED_AUTH_KEY_RECV?

man/fi_mr.3.md
shefty: We need a wider opinion on this.  At first glance, this seems reasonable to me.

shefty (commented):
I think this is in decent shape.  Thanks!

iziemba:
Another issue which needs to be addressed is management of fi_addr_t's from fi_av_insert_auth_key() and fi_av_remove(). Specifically, the following behavior needs to be supported for DAOS/Mercury.

1. I/O forwarder receives client RPC on specific fi_addr_t.
2. I/O forwarder calls fi_av_lookup_auth_key() to extract binary data for this authorization key.
3. I/O forwarder serializes binary authorization key data and forwards RPC to server.
4. Server deserializes binary authorization key and needs to lookup the local fi_addr_t corresponding to this authorization key. Thus, fi_av_insert_auth_key() is called.
5. Server generates a specific endpoint + auth_key fi_addr_t via fi_av_insert() to be used for RDMA.
6. Server issues RDMA.
7. Upon completion of the RDMA, what management actions should the server do on these fi_addr_t's? For example, should the server call fi_av_remove() on both the inserted fi_addr_t's?

shefty:
I think that's up to DAOS, but I would envision that they would want to keep the auth_keys in the AV (or even pre-insert all keys) and only add/remove the client address.

man/fi_av.3.md
iziemba: Since the ABI is changing in this PR, I'll take a shot at adding in optional caps to better support FI_AV_USER_ID.

> I'm leaning toward adding a call to set/change the user ID for an inserted address.

If I followed correctly, this would result in an insert being 2 functions calls. Would this result in users have to introduce locks to achieve the expected behavior?

iziemba:
> I think that's up to DAOS, but I would envision that they would want to keep the auth_keys in the AV (or even pre-insert all keys) and only add/remove the client address.

I have reached out with them on how they want to manage the fi_addr_t's.

Maybe the below could be done?
- Assuming providers refcount each fi_addr_t, during server init, all auth keys would be pre-inserted. Thus, refcounts == 1.
- Since servers will be receiving binary auth_key data from I/O forwarder, they have to call fi_av_insert_auth_key() to get the fi_addr_t handle. This would increase refcount == 2.
- On completion of RDMA ops, fi_av_remove() is called on fi_addr_T. Refcount == 1 now.
- The initial refcount cannot be removed until all EPs are closed using this auth_key.

shefty:
We don't enforce refcounts for other items inserted into the AV.  What we're really asking the provider to support is a search() function.  DAOS could perform that task itself, but I'm not against having the call.  If we're only looking up auth_keys, a linear search would probably be simplest.  If the number of auth_keys falls into the dozens range, then maintaining a sort of the keys is doable.  If we want to search for addresses, that's more involved.  I'd have some sort of API restriction on that capability, such as some way for the app to convey that it's needed.

man/fi_av.3.md
shefty: Yes for 2 calls.  E.g. insert() + set_user_id().

I don't know that a lock is necessary.  Prior to the address being inserted, a message could be received from that peer.  The completion would indicate FI_ADDR_NOTAVAIL.  With 2 calls, the message could arrive between calling insert() and set_user_id().  However, from the viewpoint of the app, that's the same occurring before insert() being called.  That is, the app can't distinguish between the two situations.  So, until set_user_id() is set, the provider should use FI_ADDR_NOTAVAIL.

2 calls does allow for the app to change the user ID later, if desired, and incur the raciness of doing so.

iziemba: >2 calls does allow for the app to change the user ID later, if desired, and incur the raciness of doing so.

This is the raciness I was thinking about.

shefty: I don't see there's any real race with the flow of insert + set_id.  Any race already exists prior to calling insert.  There's a race with calling set_id + set_id.  That's on the application to handle.  But I also don't expect apps to do that anyway.

man/fi_av.3.md
iziemba: I think I am going to define FI_DIRECTED_AUTH_KEY_RECV for now as a discussion point.

include/rdma/fabric.h
```diff
@@ -392,6 +392,7 @@ struct fi_tx_attr {
 	size_t			iov_limit;
 	size_t			rma_iov_limit;
 	uint32_t		tclass;
+	uint64_t		optional_caps;
```
iziemba, line 397: To support FI_AV_USER_ID, optional_caps was added. This was selected over using the existing caps field due to easier adoption by providers. Specifically, providers should not have to be updated to conform to the defined behavior of optional_caps.

include/rdma/fabric.h
```diff
@@ -168,6 +168,7 @@ typedef struct fid *fid_t;
 #define FI_PEER_TRANSFER	(1ULL << 36)
 #define FI_MR_DMABUF		(1ULL << 40)
 #define FI_AV_USER_ID		(1ULL << 41)
+#define FI_DIRECTED_AUTH_KEY	(1ULL << 42)
```
iziemba, line 171: One of the problems this capability is trying to solve is if FI_DIRECTEC_RECV can be support with multiple authorization keys.

iziemba:
> We don't enforce refcounts for other items inserted into the AV. What we're really asking the provider to support is a search() function. DAOS could perform that task itself, but I'm not against having the call. If we're only looking up auth_keys, a linear search would probably be simplest. If the number of auth_keys falls into the dozens range, then maintaining a sort of the keys is doable. If we want to search for addresses, that's more involved. I'd have some sort of API restriction on that capability, such as some way for the app to convey that it's needed.

@soumagne forwarded me so mercury code, and they have a cache of EP addr to fi_addr_t's. This is used to prevent redundant fi_av_insert() calls. Seems like this can be extend to cache output from fi_av_insert_auth_key.

iziemba:
PR updated again. Major changes:

- Define optional_caps field
- Define FI_DIRECTED_AUTH_KEY
- Define fi_av_set_user_id()

include/rdma/fabric.h
shefty: I was envisioning using the standard caps field, but categorizing some of the bits as optional caps, similar to what's done for primary/secondary/modifier caps.  Based on this, FI_AV_USER_ID would be defined as a cap bit, but as optional.

This is going a different route, where the app could conceivably list any cap bit as optional.  This is more useful but also more involved.  In this case, the caps bits stay as mandatory if set.  This option could basically remove the need to define any cap bit as 'secondary', since the app can indicate support by setting it as an optional_bit.

We should gather feedback on these options.  The latter (your change) may be appealing to apps, and I don't think ends up that difficult to support.

man/fi_domain.3.md
```diff
@@ -778,6 +779,14 @@ for additional information.
 : The maximum number of authorization keys which can be supported per connectionless
   endpoint.
 
+## Optional Capabilities (optional_caps)
+
+Domain level optional capabilities. The following are the supported optional
+capabilities: None
```
shefty: See comment on previous patch.  This path basically enables all cap bits as optional.

include/rdma/fabric.h
shefty: Can't that be determined by checking the auth_key_cnt (or max or whatever we called it) and whether FI_DIRECTED_RECV is enabled?  We require support for any-source.  Couldn't we extend that requirement to auth_keys?

man/fi_av.3.md
```diff
@@ -117,6 +119,10 @@ int fi_av_lookup_auth_key(struct fid_av *av, fi_addr_t addr,
 : On input, specifies size of auth_key buffer.  On output, stores number
   of bytes written to auth_key buffer.
 
+*user_id*
+: For address vectors configured with FI_AV_USER_ID, this defines the
+  the user ID to be associated with future inserts. NULL is valid.
```
shefty, line 124: This is assigning a generic user ID across all insertions.  I was thinking of something like this:
```
fi_set_user_id(*av, fi_addr_t addr, fi_addr_t user_id)
```

include/rdma/fabric.h
iziemba: I think it can be determined based on max_ep_auth_key. The issue here is how can the provider advertise two different capability sets based on max_ep_auth_key? Maybe reporting multiple fi_info's is the solution here.

fi_info 1:
- max_ep_auth_key == 1
- caps |= FI_DIRECTED_RECV

fi_info 2:
- max_ep_auth_key > 2
- caps &= ~FI_DIRECTED_RECV

This would put the burden on the app to sort through if multiple fi_infos are reported for the same NIC. Thoughts?

man/fi_av.3.md
iziemba: Ah. So the order of operations would be:

1. fi_av_insert() -> output is fi_addr_t addr
2. fi_av_set_user_id() -> Update addr from step 1 with user defined ID

shefty: Correct.

include/rdma/fabric.h
shefty: I'm not overly concerned about needing to update the providers.  For the providers that use the utility code for getinfo processing, that code would handle the check(s).  I'm more concerned that if we do break the API that we do that in the best way possible.

Updating the standard caps doesn't break the API.  Plus.  But this series already breaks the API, so we're needing to deal with that anyway.  Providers that don't support FI_AV_USER_ID would need to filter that bit off.  So, I agree that I think this ends up changing all providers to do something with that bit.  Negative.

Adding new optional_caps fields gives us more flexibility and makes things easier on the apps.  Plus.  Most providers shouldn't need updating, since they would end up with optional_caps set to 0.  Plus.  But until a provider does update, it won't include optional_caps as being supported.  Neutral.

My current preference is:

- Define FI_AV_USER_ID as a primary caps bit.
- Add optional_caps as new fields, as you've done.
- Allow requesting any cap bit as optional.
- Apps should only set a given bit in either the caps or optional_caps fields.  If the same bit is set in both, it is treated as mandatory (caps bit takes preference).
- Provider support for caps bits is unchanged.

Handling for optional_caps would be done as:
```
if (user->caps & ~(prov->caps))
    return -FI_ENODATA; // User requires a cap we don't support.

// Get list of optional caps that provider can support
opt_caps = user->optional_caps & prov->caps;

// output info
info->caps = user->caps & prov->caps;
info->caps |= opt_caps;
```

man/fi_av.3.md
shefty: That race already exists without a new call.  A CQ event could be generated prior to fi_av_insert() being called.  In that case the fi_addr_t will be FI_ADDR_NOTAVAIL.  In the new model, the CQ event could be generated prior to calling fi_av_set_user_id().  The fi_addr_t value would also be FI_ADDR_NOTAVAIL.  It's the same conceptual race, and there's no reasonable way for the app to determine if the CQ event was generated prior to calling fi_av_insert() or immediately after calling fi_av_insert() but before fi_av_set_user_id().

iziemba: >In the new model, the CQ event could be generated prior to calling fi_av_set_user_id(). The fi_addr_t value would also be FI_ADDR_NOTAVAIL. 

Ah yes. I missed this part. If FI_ADDR_NOTAVAIL is returned, all is good (same race as today).

include/rdma/fabric.h
iziemba: I'll take the approach of having to fi_info's based on max_ep_auth_key.

iziemba:
Latest round of comments are addressed. I think this is in a good place. Assuming there are no major changes needed, seems like a good next step may be able to socialize this version at OFIWG?

Also, I would like feedback around how many fi_info's per NIC should be reported.

shefty:
I'll look this over and respond.  I thought we were already in a pretty decent place already.  :)  So, I'm sure this is good to discuss next Tuesday.

As for how many fi_info's to report, that's really up to the provider and how the app has set the hints.  I believe tcp, shm, and verbs all report multiple fi_info's per NIC in the absence of hints, with different attributes.  There can obviously be diffferent endpoint types, but there can also be adjustments to report using shared receive queues and without, RC QPs and XRC QPs, 0-based and virtual address based memory regions, collective calls supported and not, etc.

shefty:
The changes to the fi_addr_t in this PR will need to be coordinated with the concepts proposed in #9387.

What I think this will require is defining the format of the fi_addr_t.  Because fi_addr_t is a typedef, we have the potential to change the typedef, thought we'd need to analyze the impact to apps if we wanted to make that change.

iziemba:
> The changes to the fi_addr_t in this PR will need to be coordinated with the concepts proposed in #9387.
> 
> What I think this will require is defining the format of the fi_addr_t. Because fi_addr_t is a typedef, we have the potential to change the typedef, thought we'd need to analyze the impact to apps if we wanted to make that change.

Looking at https://github.com/ofiwg/libfabric/pull/9387/commits/16a6ed69b9c47f5ec68f0fb185e1b7166ab3c4e5 and how it relates to auth_keys, I think it really depends on the FI_AV_AUTH_KEY implementation.

For example, the same dynamic array could be used to implement both fi_av_insert_auth_key() and fi_av_insert(). Assuming FI_AV_TABLE is used, the returned fi_addr_t would just be an index. If this is the case, then no bits need to be reserved in the fi_addr_t fo FI_AV_AUTH_KEY. Thoughts?

shefty:
I would like to maintain the existing semantics that allow mapping a rank number directly to an fi_addr_t value.  Ideally, the first fi_av_insert() call returns 0.  It's not horrible for the app to add + auth_key_cnt to the rank number to derive the fi_addr_t value, but may be error prone if updating existing code.

We would need 1 bit in the fi_addr_t to distinguish between an address that's an auth key value alone versus a peer endpoint address (which may be associated with an auth key). That would allow rank == fi_addr_t (because the said bit would be 0).  We just need a location for that bit that doesn't conflict with scalable rx support.

shefty:
The only change in PR #9384 that impacts the ABI that's relative to this PR is the addition of a u32 field to fi_domain_attr.  I don't have confidence in my change being correct or complete.  So, I wouldn't worry about syncing with that PR on the ABI changes.  If the advanced tagged API can move far enough along by the 1.20 release, the new field(s) can be added (maybe as reserved) to avoid needing to add more compat structs.

shefty:
An alternative to reserving a bit in the fi_addr_t for the auth_key only addresses is to require the app to pass in a flag (e.g. FI_AUTH_KEY) to functions that need a way to identify the fi_addr_t type.  This is what's been defined for multicast.  A multicast group is also identified by an fi_addr_t value and requires passing in the FI_MULTICAST flag into sendmsg/recvmsg calls in order to distinguish between a unicast and multicast address.  This seems like a better option, as that means the fi_addr_t value associated with an auth_key is determined by the provider.  Only unicast fi_addr_t values will be defined by the API (as an index).

iziemba:
> An alternative to reserving a bit in the fi_addr_t for the auth_key only addresses is to require the app to pass in a flag (e.g. FI_AUTH_KEY) to functions that need a way to identify the fi_addr_t type. This is what's been defined for multicast. A multicast group is also identified by an fi_addr_t value and requires passing in the FI_MULTICAST flag into sendmsg/recvmsg calls in order to distinguish between a unicast and multicast address. This seems like a better option, as that means the fi_addr_t value associated with an auth_key is determined by the provider. Only unicast fi_addr_t values will be defined by the API (as an index).

Just to make sure I understand this correctly, fi_addr_t's returned from fi_av_auth_key_insert() and fi_av_insert() could be the same indexes. If users what to use the auth key only fi_addr_t for FI_DIRECTED_RECV, this also have to pass in a FI_AUTH_KEY flag to fi_trecvmsg(), correct?

shefty:
Correct -- that provides the mechanism to distinguish what the fi_addr_t value references.

shefty:
I see these main options: We can reference auth_keys as <av, auth_key_index> (~FI_AV_TABLE) or require the app to store a u64 with each key (~FI_AV_MAP).  The first option has fi_addr_t as an index.  The latter it's basically a pointer.

If the number of potential keys an app will need could be more than a handful (which I suspect is the case), then based on past experience with the AV, I would choose <av, auth_key_index>.  Auth keys would be referenced by index, the same as ep addresses.

The impact to this PR would be to define an FI_AUTH_KEY flag.  When inserting an ep address into the AV, I would require passing in FI_AUTH_KEY to fi_av_insert() to associate the ep address with an auth_key.  This aligns the behavior of fi_av_insert(... FI_AUTH_KEY) with the behavior of fi_av_insert(... FI_AV_USER_ID).  Those flags are exclusive and require passing in an fi_addr_t value as part of the insert call.  (No change to what you're doing in this PR, other than requiring that the flag be set.)  If the flag isn't set, the ep address isn't associated with any auth key.

iziemba:
>The impact to this PR would be to define an FI_AUTH_KEY flag. When inserting an ep address into the AV, I would require passing in FI_AUTH_KEY to fi_av_insert() to associate the ep address with an auth_key. This aligns the behavior of fi_av_insert(... FI_AUTH_KEY) with the behavior of fi_av_insert(... FI_AV_USER_ID). Those flags are exclusive and require passing in an fi_addr_t value as part of the insert call. (No change to what you're doing in this PR, other than requiring that the flag be set.) If the flag isn't set, the ep address isn't associated with any auth key.

Makes sense to me.

Some small details which I think need to be worked through. Specifically, if my AV has 2+ auth_keys inserted, what auth_keys are the fi_addr_t's associated with if FI_AUTH_KEY is not passed into fi_av_insert()? I think the returned fi_addr_t would have to be associated with all AV auth_keys.

Then, how can this auth_key be used for outgoing RDMA operations? Are the providers required to issue  transactions to all auth_keys associated with the fi_addr_t and generate a single completion?

shefty:
I would not associate any auth_key with an ep address if one isn't provided.  Maybe that ep is accessible without an auth_key.  If an auth_key should have been provided, but wasn't, then the call can return EINVAL.  Currently, we're forcing using auth_keys or not, so this would allow for a mix.

iziemba:
FI_AUTH_KEY flag added.

man/fi_domain.3.md
```diff
@@ -744,6 +750,9 @@ The default authorization key to associate with endpoint and memory
 registrations created within the domain. This field is ignored unless the
 fabric is opened with API version 1.5 or greater.
 
+If domain auth_key_size is set to the value FI_AV_AUTH_KEY, domain auth_key
+is ignored.
```
iziemba, line 761: I am going to change this to domain auth_key must be NULL.

man/fi_endpoint.3.md
```diff
@@ -923,6 +925,8 @@ cross traffic.  The domain authorization key will be used if auth_key_size
 is set to 0.  This field is ignored unless the fabric is opened with API
 version 1.5 or greater.
 
+If the domain is opened with FI_AV_AUTH_KEY, auth_key is ignored.
```
iziemba, line 928: I think I want to change to this to if ep_attr::auth_key must be NULL.

man/fi_endpoint.3.md
```diff
@@ -911,6 +911,8 @@ The length of the authorization key in bytes.  This field will be 0 if
 authorization keys are not available or used.  This field is ignored
 unless the fabric is opened with API version 1.5 or greater.
 
+If the domain is opened with FI_AV_AUTH_KEY, auth_key_size is ignored.
```
iziemba, line 914: I think I want to change to this to if ep_attr::auth_key_size must be zero.

shefty: agreed on the null/0 requirements

include/rdma/fi_eq.h
```diff
@@ -403,6 +404,8 @@ fi_cq_readfrom(struct fid_cq *cq, void *buf, size_t count, fi_addr_t *src_addr)
 static inline ssize_t
 fi_cq_readerr(struct fid_cq *cq, struct fi_cq_err_entry *buf, uint64_t flags)
 {
+	/* For compatibility with older providers. */
+	buf->src_addr = FI_ADDR_NOTAVAIL;
```
iziemba, line 408: This is tripping a CXI provider test failure where buf is NULL. I am going to update to:
```
if (buf)
    buf->src_addr = FI_ADDR_NOTAVAIL
```

iziemba:
As HPE works through implementing this API, I would expect other minor changes, like setting domain_attr::auth_key to NULL, to happen.

man/fi_domain.3.md
```diff
@@ -772,6 +773,11 @@ This specifies the default traffic class that will be associated any endpoints
 created within the domain.  See [`fi_endpoint`(3)](fi_endpoint.3.html)
 for additional information.
 
+## Max Authorization Keys per Endpoint (max_ep_auth_key)
+
+: The maximum number of authorization keys which can be supported per connectionless
+  endpoint.
```
iziemba: What is the correct behavior here if users specify zero? Following the precedence of similar fields, seems like if this value is zero, providers can return their default value?

shefty: I think this depends on what a provider should do with this value if passed into an open domain/ep call.  Most apps will pass this value directly through.

man/fi_av.3.md
```diff
@@ -450,6 +502,59 @@ size of the buffer needed to store the address.  This size may be
 larger than the input len.  If the provided buffer is too small, the
 results will be truncated.  fi_av_straddr returns a pointer to buf.
 
+## fi_av_insert_auth_key
```
iziemba, line 531: The return code of this function needs to be addressed. Since the API only supports one auth_key insertion per call, there is no need to return a count upon success. Thus, FI_SUCCESS will be returned on success.

iziemba:
I _think_ the fi_cq_err_entry, max_ep_auth_key, and optional_caps commits are ready to be merged. Thus, I think I am going to open up new PRs with intent of lands for each of these functional changes.

man/fi_av.3.md
```diff
@@ -387,11 +407,17 @@ restrict receive buffers to a specific endpoint address and authorization key.
   will be updated to 0.  Failures will contain a fabric errno code.
 
 - *FI_AV_USER_ID*
-: This flag associates a user-assigned identifier with each AV entry
-  that is returned with any completion entry in place of the AV's address.
+: For address vectors configured without the FI_AV_USER_ID specified,
+  this flag associates a user-assigned identifier with each AV entry that
+  is returned with any completion entry in place of the AV's address.
+  Requesting this flag during insert may result in unexpected runtime
+  failures should a provider not support FI_AV_USER_ID with insert.
```
iziemba, line 414: Need to update documentation status that FI_AV_USER_ID will not be supported as an insert flag if the AV is opened with FI_AV_USER_ID.

PR #9323, opened 2023-09-12
prov/shm: Reduce atomic locking with ofi_mr_map_verify

a-szegel:
ofi_mr_verify() grabs and releases the domain lock every iteration through the rma iov vector loop.  Grab the domain lock at the begining, and hold it until we are done verifying all the RMA IOV's.  It does not make sense to grab/free a lock in a loop.


shefty (approved):
LGTM

shefty:
FYI - Alexia is out this week.

a-szegel:
Ill let her merge it.

PR #9324, opened 2023-09-12
prov/shm: respect the mr mode from the hints

shijin-aws:
Currently, when user requests FI_MR_VIRT_ADDR
in hints->domain_attr->mr_mode, shm actually
doesn't respect it unless fast_rma is true.
This patch makes shm use the mr_mode of the hints
by default.

### DIFF
--- a/prov/shm/src/smr_init.c
+++ b/prov/shm/src/smr_init.c
@@ -164,6 +164,7 @@ static int smr_getinfo(uint32_t version, const char *node, const char *service,
 				smr_resolve_addr(NULL, NULL, (char **) &cur->src_addr,
 						 &cur->src_addrlen);
 		}
+		cur->domain_attr->mr_mode = mr_mode;
 		if (fast_rma) {
 			cur->domain_attr->mr_mode |= FI_MR_VIRT_ADDR;
 			cur->tx_attr->msg_order = FI_ORDER_SAS;


shijin-aws:
An alternative fix is to define FI_MR_VIRT_ADDR in smr_domain_attr by default.

I am not sure which way is a better fix. I am open to suggestions here @shefty @aingerson

shefty:
This isn't how mr_mode bits are used.  mr_mode bits are not requests by the application.  They are agreements that the application can fulfill if the provider needs it to, the same as fi_info::mode bits.

The shm provider doesn't need VA addressing, which is a better overall option, since the app doesn't need to exchange VA prior to initiating an RMA operation.  So, it clears that mr_mode bit.

shijin-aws:
But if I don't make this change, shm won't insert the virtual address into the mr map, and I would get an error when the receiver verifying the mr map

```
libfabric:3676:1694560694::shm:mr:ofi_mr_map_verify():142<warn> target region (0xff8d82e90084 - 0xff8d82e92084) out of registered range (0x7fc6c1748040 - 0x7fc6c18480c0)
```

shefty:
shm is using 0-based offsets, not VA based offsets. The caller should have checked the mr_mode bit, seen that FI_MR_VIRT_ADDR was cleared and adjusted accordingly.

shijin-aws:
I see, I have seen we did the same adjust in atomic but not for rma, will fix it. Thanks!

PR #9327, opened 2023-09-13
prov/util: Avoid attempting to destroy an uninitialized cache entry pool

k0zmo:
On Windows, where `HAVE_MEMHOOKS_MONITOR` is undefined `ofi_monitors_add_cache` fails with `FI_ENOSYS`. Consequently, it doesn't invoke `ofi_bufpool_create` function, which results in the `cache->entry_pool` being set to `NULL`. Subsequently, when `ofi_bufpool_destroy` is called with a `NULL` pointer it dereferences a null pointer.

### DIFF
--- a/prov/util/src/util_mr_cache.c
+++ b/prov/util/src/util_mr_cache.c
@@ -461,8 +461,8 @@ buf_free:
 
 void ofi_mr_cache_cleanup(struct ofi_mr_cache *cache)
 {
-	/* If we don't have a domain, initialization failed */
-	if (!cache->domain)
+	/* If we don't have a prov, initialization failed */
+	if (!cache->prov)
 		return;
 
 	FI_INFO(cache->prov, FI_LOG_MR, "MR cache stats: "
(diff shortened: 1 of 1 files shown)


prov/util/src/util_mr_cache.c
```diff
@@ -478,7 +478,8 @@ void ofi_mr_cache_cleanup(struct ofi_mr_cache *cache)
 	ofi_rbmap_cleanup(&cache->tree);
 	if (cache->domain)
 		ofi_atomic_dec32(&cache->domain->ref);
-	ofi_bufpool_destroy(cache->entry_pool);
+	if (cache->entry_pool)
+		ofi_bufpool_destroy(cache->entry_pool);
```
shefty, line 482: Is the code calling ofi_mr_cache_cleanup() after ofi_mr_cache_init() failed?  If so, that's not supported by the code and would need some other check to handle that case completely.

k0zmo: There's a regression between v1.18.1 and main. Previously when `ofi_mr_cache_init()` failed, its `domain` field was set to NULL so that `ofi_mr_cache_cleanup()` could return-fast based on nullness of `cache->domain`:

5eaaf0677174788394ecbef22f84015eaebd9965 commit changed that and now `cache->domain` is not `NULL`ed in case of an error during `ofi_mr_cache_init()`.

The new code is wrong - it skips the `ofi_mr_cache_cleanup()` if the `cache->domain` is `NULL` which now doesn't mean that initialization failed, only that domain wasn't provided to `ofi_mr_cache_init()`. Because of that we lost the ability to distinguish between initialized MR cache with its domain set to NULL and MR cache that failed an initialization.

shefty:
The first commit needs to be removed or merged into the second.  The final update looks correct (changing the check for prov == NULL).  Thanks!

wenduwan:
bot:aws:retest

PR #9332, opened 2023-09-14
prov/shm: only increment tx cntr when inject rma succeeded.

shijin-aws:
Currently, smr_generic_rma_inject still increments tx cntr when smr_rma_fast returns a non-zero, which is wrong.

This patch fixes this bug.

### DIFF
--- a/prov/shm/src/smr_rma.c
+++ b/prov/shm/src/smr_rma.c
@@ -364,7 +364,8 @@ static ssize_t smr_generic_rma_inject(struct fid_ep *ep_fid, const void *buf,
 	smr_cmd_queue_commit(ce, pos);
 
 out:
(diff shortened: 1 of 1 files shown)


shijin-aws:
This bug caused a crash when running Intel MPI All_put_all with MPICH, as shm generates more tx completion than MPICH expects.

shefty (approved):

a-szegel (approved):

shijin-aws:
Is Intel CI failure related?

shefty:
TCP failure is definitely not related, and I think the DMABUF-test is tied into the same problem that tcp is seeing

shijin-aws:
Thanks, I will merge then

PR #9333, opened 2023-09-14
[v1.19.x] prov/shm: only increment tx cntr when inject rma succeeded.

shijin-aws:
Currently, smr_generic_rma_inject still increments tx cntr when smr_rma_fast returns a non-zero, which is wrong.

This patch fixes this bug.

(cherry picked from commit a80b72472cbc4ef9677453727979db4ee196498b)


shefty (approved):

shijin-aws:
@shefty I would treat the TCP failure in Intel CI non-related in the same way as the main branch..

PR #9335, opened 2023-09-14
prov/util: Switch util_cq, util_av, and util_cntr ep_list_lock to be ofi_genlock

a-szegel:
This change will get rid of unnecessary locking in the common utility code.

### DIFF
--- a/prov/util/src/util_av.c
+++ b/prov/util/src/util_av.c
@@ -403,7 +403,7 @@ int ofi_av_close_lightweight(struct util_av *av)
 	if (av->eq)
 		ofi_atomic_dec32(&av->eq->ref);
 
-	ofi_mutex_destroy(&av->ep_list_lock);
+	ofi_genlock_destroy(&av->ep_list_lock);
 
 	ofi_atomic_dec32(&av->domain->ref);
 	ofi_mutex_destroy(&av->lock);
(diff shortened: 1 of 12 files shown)


prov/util/src/util_av.c
```diff
@@ -552,7 +552,9 @@ int ofi_av_init_lightweight(struct util_domain *domain, const struct fi_av_attr
 	 */
 	av->context = context;
 	av->domain = domain;
-	ofi_mutex_init(&av->ep_list_lock);
+	ofi_genlock_init(&av->ep_list_lock,
+			 domain->threading != FI_THREAD_SAFE ?
+			       OFI_LOCK_NOOP : OFI_LOCK_MUTEX);
```
shefty, line 557: We can only disable locks for FI_THREAD_DOMAIN.

prov/util/src/util_cntr.c
```diff
@@ -390,7 +390,9 @@ int ofi_cntr_init(const struct fi_provider *prov, struct fid_domain *domain,
 			return ret;
 	}
 
-	ofi_mutex_init(&cntr->ep_list_lock);
+	ofi_genlock_init(&cntr->ep_list_lock,
+			 cntr->domain->threading != FI_THREAD_SAFE ?
+			 OFI_LOCK_NOOP : OFI_LOCK_MUTEX);
```
shefty, line 395: Can only disable for FI_THREAD_DOMAIN or FI_THREAD_COMPLETION

prov/util/src/util_av.c
a-szegel: done

prov/util/src/util_cntr.c
a-szegel: done

shefty:
Thanks - I doubled checked the threading models with the locks, and I believe they are set correctly.

shefty:
We need to revisit and possibly revert or modify this change.  The man pages state that control paths are always thread safe, and the ep_list_lock is associated with binding an EP with the various objects.  For apps that use msg ep's, if there's a separate threading to handle connection events, the changes to remove the locks may break them.

We can keep the switch to genlock, but may need to keep that lock enabled, regardless of the threading level selected by the app for the data path.

PR #9338, opened 2023-09-15
prov/efa: adjust posted receive size to pkt_size

wenduwan:
This change adjusts the posted buffer size to ibv_post_recv. Currently ep->mtu_size is used regardless of the actual pkt_size. In the case of zero-copy receive, the user provided buffer could be smaller than mtu_size - the actual useable size is stored in pkt_size.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke.c
+++ b/prov/efa/src/rdm/efa_rdm_pke.c
@@ -85,6 +85,14 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 	 */
 	pkt_entry->ep = ep;
 	pkt_entry->mr = mr;
+	/**
+	 * Initialize pkt_entry->pkt_size to the allocated buf size of the
+	 * bufpool. This is the data size posted to rdma-core, and MUST NOT
+	 * exceed the memory registration size. Therefore pkt_entry->pkt_size
+	 * should be adjusted according to the actual data size.
+	 */
+	pkt_entry->pkt_size = pkt_pool->attr.size - sizeof(struct efa_rdm_pke);
+	assert(pkt_entry->pkt_size == ep->mtu_size);
 	pkt_entry->alloc_type = alloc_type;
 	pkt_entry->flags = EFA_RDM_PKE_IN_USE;
 	pkt_entry->next = NULL;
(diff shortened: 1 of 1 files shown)


wenduwan:
Evaluating impact. I'm also thinking about a fabtest to cover the case(if there is not one already).

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -85,6 +85,11 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 	 */
 	pkt_entry->ep = ep;
 	pkt_entry->mr = mr;
+	/* pkt_pool should have been created by efa_rdm_ep_create_pke_pool
+	 * with size sizeof(struct efa_rdm_pke) + ep->mtu_size
+	 */
+	pkt_entry->pkt_size = pkt_pool->attr.size - sizeof(struct efa_rdm_pke);
```
shijin-aws, line 91: i don't think make it has pkt_size as mtu_size is good. It reflect the actual message size and is assigned when initiating the pkt.

wenduwan: This only affects RX. We post the entire buffer size to ibv_post_recv.

For TX the size is being adjusted to the message size.

shijin-aws: Then we should make this assignment inside `efa_rdm_ep_bulk_post_internal_rx_pkts`, around

```
for (i = 0; i < ep->efa_rx_pkts_to_post; ++i) {
		pke_vec[i] = efa_rdm_pke_alloc(ep, ep->efa_rx_pkt_pool,
					       EFA_RDM_PKE_FROM_EFA_RX_POOL);
		assert(pke_vec[i]);
	}
```

wenduwan: That should work but I see advantages to initialize members when creating the struct. Overall it should reduce code verbosity, and it does not prevent use cases which can still overwrite pkt_size to their needs.

shijin-aws: But pkt_size is not the only entry point for caller to get the pkt size. For some pkt type like long read rtm, we cannot use it to get the data size.... Assigning it inside alloc_pke seems to not hurt on this, but gives reader an impression that this field is general for all pkt type. But it doesn't.

wenduwan: I see. In that regard I think the right approach is to add the desired pkt size to the function call, i.e.
```
efa_rdm_pke_alloc(struct efa_rdm_ep *ep, struct ofi_bufpool *pkt_pool, size_t pkt_size, enum efa_rdm_pke_alloc_type alloc_type);
``` 

WDYT?

shijin-aws: That involves more unnecessary changes considering we currently have a convention to assign it inside `efa_rdm_pke_init_**`.

I am ok with you adding a comment here to say that this default value will be overridden by the initiator of the pkt inside `efa_rdm_pke_init_**`

wenduwan: Cool that sounds reasonable.

Also I don't think we currently have a convention wrt pkt_size, or other members that are not initialized explicitly. If we want to establish a convention we should probably consolidate usages and make initialization explicit.

shijin-aws (approved):

PR #9339, opened 2023-09-15
fabtests/scripts: use yaml safe_load

wenduwan:
Follow security best practice.

### DIFF
--- a/fabtests/scripts/parseyaml.py
+++ b/fabtests/scripts/parseyaml.py
@@ -82,7 +82,7 @@ def main(argv=None):
 	else:
 		fd = open(args[0], 'r')
 
-	yi = yaml.load_all(fd.read())
+	yi = yaml.safe_load_all(fd.read())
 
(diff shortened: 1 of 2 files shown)


wenduwan:
@shefty Amazon scans our source code and warned about this. Please take a look.

shefty (approved):

darrylabbate:
What about these instances?

https://github.com/ofiwg/libfabric/blob/a80b72472cbc4ef9677453727979db4ee196498b/fabtests/scripts/rft_yaml_to_junit_xml#L38

https://github.com/ofiwg/libfabric/blob/8b559ad9411e1e0fee8811e5f9f3d953f342a0d6/fabtests/scripts/parseyaml.py#L85

wenduwan:
@darrylabbate Thank you for pointing it out.

@shefty I'm not familiar with Ruby so I did not include it in this change.

PR #9341, opened 2023-09-15
contrib/intel/jenkins: Add summary fast forward

zachdworkin:
Fast forwarding the summary to the start of test output. This will eliminate false positives where a user could put a keyword in their title to get the summary to think its something to pay attention to when it actually is just the title.


Juee14Desai (approved):
Thank you for working on this. Looks good to me!

PR #9342, opened 2023-09-15
man/fi_tagged: Remove the peek for data ability

shijin-aws:
Currently, no providers support returning message data for peek operations. It also has a contradiction on the usage of cq_entry's buf field with fi_cq man page. The latter one says the buf field only applies when the receive buffer was posted with the FI_MULTI_RECV flag.

Remove this paragraph to avoid confusion.


shefty (approved):
Obivously don't need CI to run, but leaving PR open for a bit for others to view/comment on.  Thanks!

PR #9343, opened 2023-09-16
core/verbs: Add native API support for DMA-buf regions

shefty:
This expands the memory registration API to support passing in a DMA-buf region.  DMA buf regions are specified using a struct fi_mr_dmabuf.  The struct fi_mr_attr is updated to accept passing in the dmabuf structure as input, with a flag used to indicate that the registration is for a DMA-buf region.

The verbs provider is updated to support DMA-buf registration.

Changes are compile tested only.  There's no test yet available to verify functionality.

### DIFF
--- a/fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c
+++ b/fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c
@@ -71,7 +71,7 @@ static struct fid_domain	*domain;
 static struct fid_ep		*ep;
 static struct fid_av		*av;
 static struct fid_cq		*cq;
-static struct fid_mr		*mr;
+static struct fid_mr		*mr, *dmabuf_mr;
 
 static void	*buf;
 static size_t	buf_size = 65536;
@@ -171,10 +171,43 @@ err_out:
 	return;
 }
 
+void reg_dmabuf_mr(void)
+{
+	struct fi_mr_dmabuf dmabuf = {
+		.fd = xe_get_buf_fd(buf),
+		.offset = 0,
+		.len = buf_size,
+	};
+	struct fi_mr_attr mr_attr = {
+		.dmabuf = &dmabuf,
+		.access = FI_REMOTE_READ | FI_REMOTE_WRITE,
+		.requested_key = 2,
+	};
+
+	CHECK_ERROR(fi_mr_regattr(domain, &mr_attr, FI_MR_DMABUF, &dmabuf_mr));
+
+	if (fi->domain_attr->mr_mode & FI_MR_ENDPOINT) {
+		CHECK_ERROR(fi_mr_bind(dmabuf_mr, &ep->fid, 0));
+		CHECK_ERROR(fi_mr_enable(dmabuf_mr));
+	}
+
+	printf("mr %p, buf %p, rkey 0x%lx, len %zd\n",
+		dmabuf_mr, buf, fi_mr_key(dmabuf_mr), buf_size);
+
+err_out:
+	return;
+}
+
 void dereg_mr(void)
 {
 	if (mr)
-		fi_close((fid_t)mr);
+		fi_close(&mr->fid);
+}
+
+void dereg_dmabuf_mr(void)
+{
+	if (dmabuf_mr)
+		fi_close(&dmabuf_mr->fid);
 }
 
 static void finalize_ofi(void)
@@ -262,8 +295,10 @@ int main(int argc, char *argv[])
 	init_buf();
 	init_ofi();
 	reg_mr();
+	reg_dmabuf_mr();
 
 	dereg_mr();
+	dereg_dmabuf_mr();
 	finalize_ofi();
 	free_buf();
 
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -235,6 +235,12 @@ The following apply to memory registration.
   of this bit typically implies that peers must exchange addressing data
   prior to initiating any RMA or atomic operation.
 
+  For memory regions that are registered using FI_MR_DMABUF, the starting
+  'virtual address' of the DMA-buf region is equal to the offset field
+  of struct fi_mr_dmabuf that was specified through the registration call.
+  That is, a DMA-buf region starts at 'virtual address' 0, with offset
+  being used as the starting address of the registration.
+
 *FI_MR_ALLOCATED*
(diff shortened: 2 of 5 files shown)


prov/verbs/src/verbs_mr.c
```diff
@@ -377,6 +409,13 @@ static int vrb_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 			   domain->util_domain.info_domain_caps, attr,
 			   &cur_abi_attr);
 
+	if (flags & FI_MR_DMABUF) {
+		return vrb_reg_dmabuf(domain, attr->dmabuf->fd,
+				      attr->dmabuf->offset, attr->dmabuf->len,
+				      vrb_mr_ofi2ibv_access(attr->access, domain),
+				      mr);
```
j-xiong, line 416: This bypasses the reg cache.

shefty: That was intentional because I don't know how we deal with a dma-buf in the cache or if we need to.

shefty:
@j-xiong - It looks like we might be able to adapt fi_mr_reg_xe and fi_rdmabw-xe to verify the new registration call.

prov/verbs/src/verbs_mr.c
j-xiong: Good for now. In the long run, we may still want the registration be cached, probably by using a separate mr cache that use <fd, offset> instead of addr as the key.

j-xiong:
Yes, the existing component tests can be extended to have an option to use the new API.

man/fi_mr.3.md
```diff
@@ -740,6 +775,11 @@ The follow flag may be specified to any memory registration call.
   API AllocHost call (as opposed to using malloc-like host allocation,
   unified/shared memory allocation, or AllocDevice).
 
+*FI_MR_DMABUF*
```
j-xiong, line 778: Should this also be a capability bit so that the application can check if this supported by the provider?

prov/verbs/src/verbs_mr.c
shefty: The problem is that I don't know how you verify that the entry is still valid or eject it from the cache.

man/fi_mr.3.md
shefty: Maybe, and I had the same thought.  Can the current FI_HMEM work, which would require providers support FI_HMEM to also support dmabuf registration?  I'd like to avoid capability bit overload when possible.  But I'm not sure if FI_HMEM is sufficient.

I was thinking that a provider could call mmap() to map the region into the VA space, then treat the dmabuf the same as a VA registered region.  The provider may need to discover what hmem API was used to allocate the region.  Is there an issue if the provider were forced down this path?  IOW, dmabuf support becomes an optimization that the provider could leverage, otherwise it needs to fallback to whatever its standard HMEM support is.

shefty:
Updated: I changed the value of FI_MR_DMABUF, so that it *could* be defined as a capability bit in the future, if needed.  However, I left it defined only as a memory registration flag.  I added a requirement that the flag requires FI_HMEM support.

j-xiong (approved):
Good enough for starting to write tests.

shefty:
@j-xiong - I added an update to fi-mr-reg-xe.  It's not tested (I couldn't even build that on my system, so I need to move my patches.)  The Intel CI passed, but I'm not sure that test is built or run.  Only verbs supports the new API directly, but it looks like rxm should pass the call through okay.

j-xiong:
The test passed:
```
$ ./fi_xe_mr_reg -m device -S 1048576
Using first driver: 0x2fc8950 (total >= 1)
using GPU device 0: 0x2d89fd0 (total 2)
Using OFI device: verbs;ofi_rxm (mlx5_0)
mr 0x30c4dc0, buf 0xff007fffffef0000, rkey 0x175916, len 1048576
mr 0x30c4dc0, buf 0xff007fffffef0000, rkey 0x175916, len 1048576
```

fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c
```diff
@@ -171,10 +171,43 @@ void reg_mr(void)
...
+
+	CHECK_ERROR(fi_mr_regattr(domain, &mr_attr, FI_MR_DMABUF, &dmabuf_mr));
+
+	if (fi->domain_attr->mr_mode & FI_MR_ENDPOINT) {
+		CHECK_ERROR(fi_mr_bind(dmabuf_mr, &ep->fid, 0));
+		CHECK_ERROR(fi_mr_enable(dmabuf_mr));
+	}
+
+	printf("mr %p, buf %p, rkey 0x%lx, len %zd\n",
+		mr, buf, fi_mr_key(mr), buf_size);
```
j-xiong, line 195: should print out `dmabuf_mr` instead of `mr`, same for the mr key.

shefty: Thanks - updated.  And thanks for testing. :)

j-xiong:
The rkey doen't look right:
```
 ./fi_xe_mr_reg -m device -S 1048576
Using first driver: 0x2bda950 (total >= 1)
using GPU device 0: 0x299bfd0 (total 2)
Using OFI device: verbs;ofi_rxm (mlx5_0)
mr 0x2d034e0, buf 0xff007fffffef0000, rkey 0x1738f7, len 1048576
mr 0x2d03690, buf 0xff007fffffef0000, rkey 0x0, len 1048576
```

shefty:
I see the issue.  The libfabric fid_mr wasn't initialized after the register call returned.

shefty:
I need to either go through vrb_mr_reg_common() or duplicate the functionality after ibv_reg_mr().

shefty:
Restructured the verbs changes to go through vrb_mr_reg_common(), so that the memory region is properly initialized.

j-xiong:
Works fine now:
``` 
$ ./fi_xe_mr_reg -m device -S 1048576
Using first driver: 0x27c9950 (total >= 1)
using GPU device 0: 0x258afd0 (total 2)
Using OFI device: verbs;ofi_rxm (mlx5_0)
mr 0x28df930, buf 0xff007fffffef0000, rkey 0x1726e5, len 1048576
mr 0x28f2450, buf 0xff007fffffef0000, rkey 0x17f6b5, len 1048576
```

shefty:
bot:aws:retest

PR #9345, opened 2023-09-18
prov/efa: Disallow RDMA read between incompatible EFA devices

darrylabbate:
This simply adds the "device version" (`ibv_device_attr.vendor_part_id`) to the handshake subprotocol; storing it in the `efa_rdm_peer` struct upon recv. This is then used during RDM protocol selection; ensuring the devices can interoperate before selecting RDMA read.

Currently, the only known incompatibility is between the oldest device versions (`0xEFA0`) and newer ones (`0xEFA1`, `0xEFA2`).

This was tested by running `fi_pingpong` with various EC2 instance types and checking the `rdma_read_bytes` hardware counters before/after.

I'll write unit tests in the meantime, but this can be merged if the feature is considered urgent enough.

Side note: the replacement of license text with SPDX identifiers in file headers was [blessed by Sean](https://libfabric.slack.com/archives/C05K5E1RE7L/p1694819124677119?thread_ts=1694810617.796649&cid=C05K5E1RE7L) and aligns with the latest internal Amazon guidance.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_peer.h
+++ b/prov/efa/src/rdm/efa_rdm_peer.h
@@ -129,6 +129,41 @@ bool efa_both_support_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer
 }
 
 /**
+ * @brief Determine interoperability for RDMA read between peers
+ *
+ * RDMA read is currently not interoperable between all EFA platforms; older
+ * platforms cannot fully interoperate with newer platforms. The platform
+ * version can be inferred from ibv_device_attr.vendor_part_id (see
+ * ibv_query_device(3)).
+ *
+ * In this context, version 0xEFA0 is considered "older", while versions 0xEFA1+
+ * are considered "newer"
+ *
+ * @todo Either:
+ * - Refactor this logic if the interoperability is ever reported
+ *   directly via firmware or rdma-core
+ * - Remove altogether if RDMA read becomes interoperable between all EFA
+ *   platforms
+ *
+ * @param[in]	ep	Endpoint for communication with peer
+ * @param[in]	peer	An EFA peer
+ *
+ * @return true if peers can interoperate via RDMA read; false otherwise
+ */
+static inline
+bool efa_rdm_interop_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer)
+{
+	bool rdma_read_support = efa_both_support_rdma_read(ep, peer);
+	uint32_t ep_dev_ver = efa_rdm_ep_domain(ep)->device->ibv_attr.vendor_part_id,
+		 peer_dev_ver = peer->device_version;
+
+	if (ep_dev_ver == 0xEFA0) {
+		return rdma_read_support && peer_dev_ver == 0xEFA0;
+	}
+	return rdma_read_support && peer_dev_ver != 0xEFA0;
+}
+
+/**
  * @brief determine if both peers support RDMA write
  *
  * This function can only return true if a handshake packet has already been
--- a/prov/efa/src/rdm/efa_rdm_peer.h
+++ b/prov/efa/src/rdm/efa_rdm_peer.h
@@ -1,35 +1,6 @@
-/*
- * Copyright (c) 2019-2023 Amazon.com, Inc. or its affiliates.
- * All rights reserved.
- *
- * This software is available to you under a choice of one of two
- * licenses.  You may choose to be licensed under the terms of the GNU
- * General Public License (GPL) Version 2, available from the file
- * COPYING in the main directory of this source tree, or the
- * BSD license below:
- *
- *     Redistribution and use in source and binary forms, with or
- *     without modification, are permitted provided that the following
- *     conditions are met:
- *
- *      - Redistributions of source code must retain the above
- *        copyright notice, this list of conditions and the following
- *        disclaimer.
- *
- *      - Redistributions in binary form must reproduce the above
- *        copyright notice, this list of conditions and the following
- *        disclaimer in the documentation and/or other materials
- *        provided with the distribution.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
- * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
- * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
- * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
- * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
- */
+/* Copyright Amazon.com, Inc. or its affiliates. All rights reserved. */
+/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
+
 #ifndef EFA_RDM_PEER_H
 #define EFA_RDM_PEER_H
 
@@ -47,7 +18,7 @@ OFI_DECL_RECVWIN_BUF(struct efa_rdm_pke*, efa_rdm_robuf, uint32_t);
 #define EFA_RDM_PEER_IN_BACKOFF BIT_ULL(3) /**< peer is in backoff mode due to RNR (Endpoint should not send packet to this peer) */
 /**
  * @details
- * FI_EAGAIN error was encountered when sending handsahke to this peer,
+ * FI_EAGAIN error was encountered when sending handshake to this peer,
  * the peer was put in efa_rdm_ep->handshake_queued_peer_list.
  * Progress engine will retry sending handshake.
  */
@@ -56,6 +27,7 @@ OFI_DECL_RECVWIN_BUF(struct efa_rdm_pke*, efa_rdm_robuf, uint32_t);
 struct efa_rdm_peer {
 	bool is_self;			/**< flag indicating whether the peer is the endpoint itself */
 	bool is_local;			/**< flag indicating wehther the peer is local (on the same instance) */
+	uint32_t device_version;	/**< EFA device version */
 	fi_addr_t efa_fiaddr;		/**< libfabric addr from efa provider's perspective */
 	fi_addr_t shm_fiaddr;		/**< libfabric addr from shm provider's perspective */
 	uint64_t host_id; 		/* Optional peer host id. Default 0 */
--- a/prov/efa/src/rdm/efa_rdm_pke_nonreq.h
+++ b/prov/efa/src/rdm/efa_rdm_pke_nonreq.h
@@ -1,38 +1,8 @@
-/*
- * Copyright (c) Amazon.com, Inc. or its affiliates.
- * All rights reserved.
- *
- * This software is available to you under a choice of one of two
- * licenses.  You may choose to be licensed under the terms of the GNU
- * General Public License (GPL) Version 2, available from the file
- * COPYING in the main directory of this source tree, or the
- * BSD license below:
- *
- *     Redistribution and use in source and binary forms, with or
- *     without modification, are permitted provided that the following
- *     conditions are met:
- *
- *      - Redistributions of source code must retain the above
- *        copyright notice, this list of conditions and the following
- *        disclaimer.
- *
- *      - Redistributions in binary form must reproduce the above
- *        copyright notice, this list of conditions and the following
- *        disclaimer in the documentation and/or other materials
- *        provided with the distribution.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
- * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
- * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
- * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
- * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
- */
+/* Copyright Amazon.com, Inc. or its affiliates. All rights reserved. */
+/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
 
-#ifndef _EFA_RDM_PKE_MISC_H
-#define _EFA_RDM_PKE_MISC_H
+#ifndef _EFA_RDM_PKE_NONREQ_H
+#define _EFA_RDM_PKE_NONREQ_H
 
 #include "efa_rdm_ope.h"
 #include "efa_rdm_protocol.h"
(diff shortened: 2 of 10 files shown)


darrylabbate:
AWS CI failure looks like it's due to a change that was reverted earlier this morning. Will retry a little later.

wenduwan:
bot:aws:retest

darrylabbate:
Looking into timed out EFA unit tests

shijin-aws:
currently, EFA unit test may timeout if there is a failure in any test items. There was a lock that couldn't be released when the test crashed in the middle.

shijin-aws:
unit test build failed

```
9:09:38 2023-09-19 02:00:57,099 - ERROR - utils - Command make -j failed with error:
19:09:38 In file included from prov/efa/test/efa_unit_tests.h:10:0,
19:09:38                  from prov/efa/test/efa_unit_test_common.c:1:
19:09:38 prov/efa/test/efa_unit_test_common.c: In function ‘efa_unit_test_handshake_pkt_construct’:
19:09:38 prov/efa/test/efa_unit_test_common.c:258:20: error: invalid type argument of unary ‘*’ (have ‘uint32_t {aka unsigned int}’)
19:09:38    assert_int_equal(*efa_rdm_pke_get_handshake_opt_device_version(pkt_entry), attr->device_version);
19:09:38                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
19:09:38 /home/ec2-user/PortaFiducia/libraries/cmocka/cmocka_install/include/cmocka.h:116:28: note: in definition of macro ‘cast_to_largest_integral_type’
19:09:38      ((LargestIntegralType)(value))
19:09:38                             ^~~~~
19:09:38 prov/efa/test/efa_unit_test_common.c:258:3: note: in expansion of macro ‘assert_int_equal’
19:09:38    assert_int_equal(*efa_rdm_pke_get_handshake_opt_device_version(pkt_entry), attr->device_version);
19:09:38    ^~~~~~~~~~~~~~~~
19:09:38 make[1]: *** [prov/efa/test/prov_efa_test_efa_unit_test-efa_unit_test_common.o] Error 1
19:09:38 make[1]: *** Waiting for unfinished jobs....
19:09:38 make: *** [all] Error 2
19:09:38 
```

darrylabbate:
That wasn't even the commit I wanted to push, lol. I'll convert this to a draft PR until I get these passing locally.

prov/efa/src/rdm/efa_rdm_protocol.h
```diff
@@ -372,8 +342,10 @@ static_assert(sizeof(struct efa_rdm_handshake_hdr) == 8, "efa_rdm_handshake_hdr
 #endif
 
 struct efa_rdm_handshake_opt_connid_hdr {
-	uint32_t connid;
-	uint32_t padding; /* padding to 8 bytes boundary */
+	union {
+		uint32_t connid;
+		uint64_t padding;
```
shijin-aws, line 348: Is this change necessary? Your updated protocol doc still makes padding a 4 byte

prov/efa/src/rdm/efa_rdm_pke_nonreq.h
```diff
@@ -97,6 +67,30 @@ uint64_t *efa_rdm_pke_get_handshake_opt_host_id_ptr(struct efa_rdm_pke *pke)
...
+	struct efa_rdm_handshake_hdr *handshake_hdr;
+	struct efa_rdm_handshake_opt_device_version_hdr *device_version_hdr;
+	size_t offset;
+
+	handshake_hdr = efa_rdm_pke_get_handshake_hdr(pke);
+	assert(handshake_hdr->type == EFA_RDM_HANDSHAKE_PKT);
+	assert(handshake_hdr->flags & EFA_RDM_HANDSHAKE_DEVICE_VERSION_HDR);
+
+	offset = sizeof (struct efa_rdm_handshake_hdr)
+		+ ((handshake_hdr->nextra_p3 - 3) * sizeof handshake_hdr->extra_info[0]);
```
shijin-aws, line 82: silly c question:  should it be `sizeof(handshake_hdr->extra_info[0])` ?

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -128,6 +128,47 @@ bool efa_both_support_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer
...
+ */
+static inline
+bool efa_rdm_interop_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer)
+{
+	bool rdma_read_support = efa_both_support_rdma_read(ep, peer);
+	uint32_t ep_dev_ver = efa_rdm_ep_domain(ep)->device->ibv_attr.vendor_part_id,
+		 peer_dev_ver = peer->device_version;
+
+	switch (ep_dev_ver) {
+	case 0xEFA0:
```
shijin-aws, line 161: should we use lower case here to keep consistent with other places in the code for the vendor id.

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -128,6 +128,47 @@ bool efa_both_support_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer
...
+		return rdma_read_support && peer_dev_ver == 0xEFA0;
+	case 0xEFA1:
+	case 0xEFA2:
```
shijin-aws, line 164: We should only care 0xefa0 and NOT 0xefa0 cases here. No need to restrict the version id within the [0xefa0, 0xefa2] range. See Brian's comment in https://github.com/ofiwg/libfabric/pull/8186#discussion_r1008588620

shijin-aws:
Overall LGTM, main comment is for the interop rdma read check.

prov/efa/src/rdm/efa_rdm_pke_nonreq.h
darrylabbate: `sizeof` is a unary operator where the operand can be an expression or a parenthesized data type name. I prefer to stylize `sizeof` expressions to match their semantics, rather than treating `sizeof` as if it's a function/macro.

prov/efa/src/rdm/efa_rdm_peer.h
darrylabbate: I opted for uppercase since that's what `ibv_devinfo` outputs for the `hw_ver` field (`vendor_part_id` is actually decimal for some reason). I don't feel strongly either way. Obviously it makes no difference functionally in this context.

prov/efa/src/rdm/efa_rdm_protocol.h
darrylabbate: Right, it's a minor stylistic change in this case. I feel it's a more accurate representation of a primitive value aligned to an 8-byte boundary. Whereas the original syntax might be misconstrued as if there needs to be 4 bytes of padding/reserved space, regardless of the size of its preceding field(s).

prov/efa/src/rdm/efa_rdm_peer.h
darrylabbate: Lol, I chose to explicitly compare against `0xEFA[0-2]` due to his other comment further down: https://github.com/ofiwg/libfabric/pull/8186#discussion_r1009783855

I don't feel strongly either way.

shijin-aws: I think they don't conflict. Brian wants to have a explicit check for 0xefa0 and 0xefa1 because we don't want to enable rdma-read for them by default. He doesn't care the others because we are enabling them by default.

Here the case is similar. We just don't allow rdma-read traffic between AL7 and non-AL7.

darrylabbate:
`test_efa_rdm_ep_handshake_receive_and_send_valid_host_ids_with[out]_connid()` are the failing unit tests. I'm gonna need to deep dive into the code, because from a cursory glance I'm not sure exactly what's supposed to be happening.

E.g. Complete misuse of `calloc()`. What is this even for? It just attempts (fails according to ASan) to allocate a massive chunk of memory. https://github.com/ofiwg/libfabric/blob/99f87709c4e50f53a2de5a2a3c6ec79aa3abedde/prov/efa/test/efa_unit_test_common.c#L230

shijin-aws:
yeah that calloc doesn't make any sense to me. @wenduwan ?

wenduwan:
I remember trying to replicate [this line](https://github.com/ofiwg/libfabric/blob/99f87709c4e50f53a2de5a2a3c6ec79aa3abedde/prov/efa/src/rdm/efa_rdm_pke_nonreq.c#L74C55-L74C55)

The [handshake protocol](https://github.com/ofiwg/libfabric/blob/main/prov/efa/docs/efa_rdm_protocol_v4.md) specifies the `extra_info` fields as an array of `8 * (nextra_p3 - 3)` bytes, which is what the calloc is doing.

shijin-aws:
but what is this calloc for? You didn't assign its return to any pointer

wenduwan:
Yeah maybe should use memset instead.

darrylabbate:
> The [handshake protocol](https://github.com/ofiwg/libfabric/blob/main/prov/efa/docs/efa_rdm_protocol_v4.md) specifies the `extra_info` fields as an array of `8 * (nextra_p3 - 3)` bytes, which is what the calloc is doing.

I would encourage you to read the [documentation for `calloc()`](https://en.cppreference.com/w/c/memory/calloc) again. That call is attempting to allocate more memory than the machine has (in most cases), since `(uintptr_t) handshake_hdr->extra_info` will be some memory address and `nex * sizeof (uint64_t)` is going to evaluate to `8` at the minimum. It's going to fail. Plus the returned pointer is not being stored anywhere, so it's a completely pointless no-op even if the `calloc()` succeeded.

My point here is that I don't think this set of unit tests was sufficiently reviewed - and that the tests have been ineffective. From a cursory glance, there are other simple mistakes like this (logical instead of bitwise AND): https://github.com/ofiwg/libfabric/blob/99f87709c4e50f53a2de5a2a3c6ec79aa3abedde/prov/efa/test/efa_unit_test_ep.c#L130

Given the dubious implementation details, I recommend these tests get `skip()`'d until they can be audited for correctness.

wenduwan:
Oh is the wrong calloc call the root cause for the failure?

wenduwan:
> Given the dubious implementation details, I recommend these tests get skip()'d until they can be audited for correctness.

Please do that for this change if possible. This is the only place we test the impl details for handshake. Do you suggest testing it in another way?

darrylabbate:
> Oh is the wrong calloc call the root cause for the failure?

Not necessarily (I'm not sure how it even works currently); I'm suspicious of the implementation of these particular unit tests as a whole if this slipped through the cracks. I still need to dig into it.

> Do you suggest testing it in another way?

Not necessarily, but auditing/refactoring this adds more time than what would be expected for the feature itself. Obviously it needs to be done, but it depends on the urgency of this feature. There's justification for skipping the tests temporarily to get the feature merged since I don't believe the tests are currently effective.

For now, I'm adding a commit to `skip()` them for evaluation purposes while I dig into them a bit.

wenduwan:
Yeah I think it overall makes sense and saves time to fix the found problems and extend the test to cover the change in this PR.

From my experience it helps to add prints along the call chain and figure out where the hang happens. The complexity of this test is to simulate the handshake exchange and understand the interaction with the mocked remote peer.

darrylabbate:
Last time I looked at it, the hang happens during the `fi_cq_read()` call here https://github.com/ofiwg/libfabric/blob/99f87709c4e50f53a2de5a2a3c6ec79aa3abedde/prov/efa/test/efa_unit_test_ep.c#L193

Running `make check` when configured with ASan reveals the `calloc()` goof mentioned earlier.

Removing that line (on this branch) altogether spits out the following trace:

```
$ ASAN_OPTIONS=detect_leaks=0 prov/efa/test/efa_unit_test
ASAN:DEADLYSIGNAL
=================================================================
==36610==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000002 (pc 0x7f198cf48a32 bp 0x7ffc892618e0 sp 0x7ffc89261880 T0)
==36610==The signal is caused by a READ memory access.
==36610==Hint: address points to the zero page.
    #0 0x7f198cf48a31  (/lib64/libefa.so.1+0x1a31)
    #1 0x50d4f7 in ibv_wr_set_sge_list /usr/include/infiniband/verbs.h:1419
    #2 0x50ff34 in efa_rdm_pke_sendv prov/efa/src/rdm/efa_rdm_pke.c:434
    #3 0x507ab9 in efa_rdm_ep_post_handshake prov/efa/src/rdm/efa_rdm_ep_utils.c:619
    #4 0x507bc0 in efa_rdm_ep_post_handshake_or_queue prov/efa/src/rdm/efa_rdm_ep_utils.c:657
    #5 0x520ef0 in efa_rdm_pke_handle_recv_completion prov/efa/src/rdm/efa_rdm_pke_cmd.c:950
    #6 0x50ad9c in efa_rdm_ep_poll_ibv_cq prov/efa/src/rdm/efa_rdm_ep_progress.c:502
    #7 0x50b739 in efa_rdm_ep_progress_internal prov/efa/src/rdm/efa_rdm_ep_progress.c:615
    #8 0x50c6c2 in efa_rdm_ep_progress prov/efa/src/rdm/efa_rdm_ep_progress.c:810
    #9 0x6746c8 in ofi_cq_progress prov/util/src/util_cq.c:498
    #10 0x673275 in ofi_cq_readfrom prov/util/src/util_cq.c:257
    #11 0x6fa84d in efa_rdm_cq_readfrom prov/efa/src/rdm/efa_rdm_cq.c:105
    #12 0x66f2b7 in fi_cq_readfrom include/rdma/fi_eq.h:400
    #13 0x6732bc in ofi_cq_read prov/util/src/util_cq.c:264
    #14 0x409dae in fi_cq_read include/rdma/fi_eq.h:394
    #15 0x40b7e3 in test_efa_rdm_ep_handshake_exchange_host_id prov/efa/test/efa_unit_test_ep.c:181
    #16 0x40bad5 in test_efa_rdm_ep_handshake_receive_and_send_valid_host_ids_with_connid prov/efa/test/efa_unit_test_ep.c:210
    #17 0x7f198cb23c0e in cmocka_run_one_test_or_fixture /home/ec2-user/cmocka/src/cmocka.c:3314
    #18 0x7f198cb23f07 in cmocka_run_one_tests /home/ec2-user/cmocka/src/cmocka.c:3430
    #19 0x7f198cb24488 in _cmocka_run_group_tests /home/ec2-user/cmocka/src/cmocka.c:3558
    #20 0x4064cf in main prov/efa/test/efa_unit_tests.c:138
    #21 0x7f198bf5e139 in __libc_start_main (/lib64/libc.so.6+0x21139)
    #22 0x405d19 in _start (/home/ec2-user/libfabric/prov/efa/test/efa_unit_test+0x405d19)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/lib64/libefa.so.1+0x1a31)
==36610==ABORTING
```

Will look further into this. Looks as if some argument to `ibv_wr_set_sge_list()` that expects an address is being passed in as `2`. Could be that `qp->wr_set_sge_list` isn't being initialized (mocked?). @wenduwan points out `ibv_wr_set_sge_list()` should be mocked if the code path is valid.

darrylabbate:
> @wenduwan points out `ibv_wr_set_sge_list()` should be mocked if the code path is valid.

This ended up being the issue. There was already a mock for it; just needed to overwrite the QP function pointer in the utility function for handshake unit tests.

For cases where `local_host_id` was `0` and the tests passed, the handshake ended up being 32 bytes, which hit a different code path - and the called functions were properly mocked.

darrylabbate:
Pushing latest just for evaluation - unit tests no longer skipped. Working on unit tests for the RDMA read interoperability.

darrylabbate:
On second thought, it's best to protect this feature via a simple correctness test in the AWS CI. There's nothing to really unit test here without testing implementation details. The unit test code I modified should confirm the device version is being exchanged during handshake.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -127,7 +127,7 @@ void test_efa_rdm_ep_handshake_exchange_host_id(struct efa_resource **state, uin
 	assert_non_null(peer);
 	/* Peer host id is uninitialized before handshake */
 	assert_int_equal(peer->host_id, 0);
-	assert_false(peer->flags && EFA_RDM_PEER_HANDSHAKE_SENT);
+	assert_false(peer->flags & EFA_RDM_PEER_HANDSHAKE_SENT);
```
shijin-aws, line 130: how about adding a `!!` here? it's now a uint64 I guess.

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -128,6 +128,41 @@ bool efa_both_support_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer
...
+bool efa_rdm_interop_rdma_read(struct efa_rdm_ep *ep, struct efa_rdm_peer *peer)
+{
+	bool rdma_read_support = efa_both_support_rdma_read(ep, peer);
+	uint32_t ep_dev_ver = efa_rdm_ep_domain(ep)->device->ibv_attr.vendor_part_id,
+		 peer_dev_ver = peer->device_version;
+
+	if (ep_dev_ver == 0xEFA0) {
+		return rdma_read_support && peer_dev_ver == 0xEFA0;
+	}
+	return rdma_read_support && peer_dev_ver != 0xEFA0;
```
shijin-aws, line 163: this is good, thank you!

shijin-aws (approved):

prov/efa/test/efa_unit_test_ep.c
darrylabbate: Yeah so there's definitely some weirdness with pedantically asserting `false` for a masked `uint32_t`. It's probably ok in this scenario since the expression can only evaluate to `0` or `1` since `EFA_RDM_PEER_HANDSHAKE_SENT` is `1`. I'll just change this to an `assert_int_not_equal()` call.

Side note: CMocka's suite of `assert_*` macros is actually lacking a little. There are macros for comparing `int` and `double`, but that's really it for primitive values. Unity has [corresponding macros](https://github.com/ThrowTheSwitch/Unity/blob/master/docs/UnityAssertionsReference.md#signed-and-unsigned-integers-of-all-sizes) for just about everything in `stdint.h`

shijin-aws:
please backport

PR #9348, opened 2023-09-18
fabtests/rdm_tagged_*: Add FI_DIRECTED_RECV to requested capabilities

shefty:
This change better aligns with common use of FI_TAGGED.

Bonus patch - update help text to the tcp provider max_saved_size


shefty:
@j-xiong - Does the ucx provider not support FI_DIRECTED_RECV? The Ci output indicates no.  I thought it was required by mpi.

j-xiong:
The ucx provider doesn't support FI_DIRECTED_RECV. Intel MPI can work with that. Upstream MPICH can work with that by setting MPIR_CVAR_CH4_OFI_ENABLE_DATA=0. Open MPI can be modified to work with that by adding an option to turn off FI_DIRECT_RECV requirement.

shefty:
It looks like the only way for ucx to support this is to steal tag bits...  This change will need to be a command line option.

PR #9349, opened 2023-09-19
prov/xnet,common: Disable the EP if an error is detected for zero-copy

### LINKED ISSUE #7887: prov/tcp: zerocopy being disabled under load
**Describe the bug**
While running an aggressive RDMA test using TCP provider, I started seeing these pair of warnings from libfabric:

tcp  ep_data   warn ofi_bsock_async_done:1270 Error reading MSG_ERRQUEUE (Resource temporarily unavailable)
tcp  ep_data   warn ofi_bsock_async_done:1296 disabling zerocopy

Looking at the code, it seems we are trying to read the MSG_ERRQUEUE via a readmsg() call and are getting EAGAIN back.  I'm not sure if this should be expected under heavy load or not.  Triggers us to disable zerocopy.

Now, I'm not sure if it is related, but shortly after that, I see this assert from the TCP provider:

red: prov/tcp/src/tcpx_progress.c:674: tcpx_recv_hdr: Assertion `ep->cur_rx.hdr.base_hdr.id == ep->rx_id++' failed.

In another case where I saw those warnings, it was followed by this assert:

red: prov/tcp/src/tcpx_progress.c:415: tcpx_handle_ack: Assertion `!slist_empty(&ep->need_ack_queue)' failed.

I'm starting to get the feeling that dynamically disabling zerocopy might open us up to asserts.

**To Reproduce**
My setup is the following:
- libfabric version 1.14
- 30 ARM-based node cluster
- Each node has two endpoints to every other node
- Dedicated polling threads are used.  Each polling thread "owns" a set of endpoints and only it accesses them.  I've reproduced this for both COMPLETION mode and THREAD_SAFE mode.
- An RDMA test is performed which loads up all endpoints with 32 1M RDMA transfers each direction
- Zerocopy seems to be on by default now (I've see the patches that introduced this)
- The NICs are Mellanox CX5 in RoCE mode.  We are using them with the TCP provider to test it.
- This issue only happens when pushing RDMA to the limit.  Does not happen with aggressive messaging.

**Expected behavior**
Zerocopy does not get disabled.  Also, we don't assert!!

**Output**
Output above.

**Environment:**
Ubuntu Linux 20.04.4 LTS

**Additional context**

sydidelot:
This patch disables the endpoint if zero-copy reports an error. The goal is to prevent the stream to be corrupted if zero-copy gets disabled after the transfer has started.

Fixes #7887

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1402,8 +1402,8 @@ void ofi_bsock_prefetch_done(struct ofi_bsock *bsock, size_t len)
 }
 
 #ifdef MSG_ZEROCOPY
-uint32_t ofi_bsock_async_done(const struct fi_provider *prov,
-			      struct ofi_bsock *bsock)
+int ofi_bsock_async_done(const struct fi_provider *prov,
+			 struct ofi_bsock *bsock)
 {
 	struct msghdr msg = {};
 	struct sock_extended_err *serr;
@@ -1418,7 +1418,7 @@ uint32_t ofi_bsock_async_done(const struct fi_provider *prov,
 	if (ret < 0) {
 		FI_WARN(prov, FI_LOG_EP_DATA,
 			"Error reading MSG_ERRQUEUE (%s)\n", strerror(errno));
-		goto disable;
+		return -errno;
 	}
 
 	assert(!(msg.msg_flags & MSG_CTRUNC));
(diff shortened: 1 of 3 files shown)


src/common.c
```diff
@@ -1427,31 +1427,30 @@ uint32_t ofi_bsock_async_done(const struct fi_provider *prov,
...
-		goto disable;
+		return -EINVAL;
 	}
 
 	serr = (void *) CMSG_DATA(cmsg);
 	if ((serr->ee_origin != SO_EE_ORIGIN_ZEROCOPY) || serr->ee_errno) {
 		FI_WARN(prov, FI_LOG_EP_DATA,
 			"Unexpected sock err origin or errno\n");
-		goto disable;
+		return -EINVAL;
```
shefty, line 1437: nit: Can you convert this and the above to -FI_EINVAL, just for consistency.

shefty (approved):
one minor change request.  Looks good -- thanks!

src/common.c
sydidelot: Thanks for the fast review! Sure, I'll fix the error codes.

sydidelot: Done

PR #9355, opened 2023-09-20
fabtests/shared.c: Check if fi_info is returned correctly in case of FI_CONNREQ

Juee14Desai:
This commit adds validation that the fi_info returned with an FI_CONNREQ
event is filled out properly with valid fields. The fi_info should be
similar to the fi_info that is used to open passive endpoint.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -1083,6 +1083,20 @@ int ft_complete_connect(struct fid_ep *ep, struct fid_eq *eq)
 	return 0;
 }
 
+int ft_verify_info(struct fi_info *fi_pep, struct fi_info *info)
+{
+	if (!info || !info->fabric_attr || !info->domain_attr ||
+	    !info->ep_attr || !info->tx_attr || !info->rx_attr)
+		return -FI_EINVAL;
+
+	if (!info->fabric_attr->prov_name ||
+	    !info->fabric_attr->name || !info->domain_attr->name ||
+	    info->fabric_attr->api_version != fi_pep->fabric_attr->api_version)
+		return -FI_EINVAL;
+
+	return 0;
+}
+
 int ft_retrieve_conn_req(struct fid_eq *eq, struct fi_info **fi)
 {
 	struct fi_eq_cm_entry entry;
@@ -1100,6 +1114,11 @@ int ft_retrieve_conn_req(struct fid_eq *eq, struct fi_info **fi)
 	if (event != FI_CONNREQ) {
 		fprintf(stderr, "Unexpected CM event %d\n", event);
 		ret = -FI_EOTHER;
+		return ret;
+	}
+
+	if ((ret = ft_verify_info(fi_pep, entry.info))) {
+		printf("ret: %d\n", ret);
 		return ret;
 	}
 
--- a/prov/sockets/src/sock_ep.c
+++ b/prov/sockets/src/sock_ep.c
@@ -1417,7 +1417,8 @@ static void sock_set_fabric_attr(void *src_addr, const struct fi_fabric_attr *hi
 	if (!attr->name)
 		attr->name = strdup(sock_fab_name);
 
-	attr->prov_name = NULL;
+	attr->prov_name = strdup(hint_attr->prov_name);
+	attr->api_version = hint_attr->api_version;
 }
 
 static void sock_set_domain_attr(uint32_t api_version, void *src_addr,
@@ -1498,11 +1499,17 @@ struct fi_info *sock_fi_info(uint32_t version, enum fi_ep_type ep_type,
 {
 	struct fi_info *info;
 
-	info = fi_allocinfo();
+	info = fi_dupinfo(hints);
 	if (!info)
 		return NULL;
+	free(info->src_addr);
+	free(info->dest_addr);
+	info->src_addr = NULL;
+	info->dest_addr = NULL;
+	info->src_addrlen = 0;
+	info->dest_addrlen = 0;
 
-	info->src_addr = calloc(1, sizeof(union ofi_sock_ip));
+	info->src_addr = calloc(1, ofi_sizeofip(src_addr));
 	if (!info->src_addr)
 		goto err;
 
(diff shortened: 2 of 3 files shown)


fabtests/common/shared.c
```diff
@@ -1083,6 +1083,24 @@ int ft_complete_connect(struct fid_ep *ep, struct fid_eq *eq)
...
+int ft_verify_info(struct fi_info *fi_pep, struct fi_info *info)
+{
+	if (!info || !info->fabric_attr || !info->domain_attr ||
+	    !info->ep_attr || !info->tx_attr || !info->rx_attr)
+		return -FI_EINVAL;
+
+	if (!info->fabric_attr->prov_name ||
+	    !info->src_addr || !info->dest_addr ||
+	    info->fabric_attr->api_version != fi_pep->fabric_attr->api_version ||
+	    info->src_addr != fi_pep->src_addr ||
```
shefty, line 1095: This is comparing pointers of src_addr, which won't be true.  Remove this part of the check

fabtests/common/shared.c
```diff
@@ -1083,6 +1083,24 @@ int ft_complete_connect(struct fid_ep *ep, struct fid_eq *eq)
...
+	    info->dest_addr != fi_pep->dest_addr ||
```
shefty, line 1096: Similar to above.  It's comparing pointers, which won't be true.  And, the fi_pep->dest_addr will be null.  Remove

fabtests/common/shared.c
```diff
@@ -1083,6 +1083,24 @@ int ft_complete_connect(struct fid_ep *ep, struct fid_eq *eq)
...
+	    info->fabric_attr->name != fi_pep->fabric_attr->name ||
```
shefty, line 1097: Names are strings.  Again, the pointers won't match.  Even the names may not match.  Check that info->fabric_attr->name is non-null

fabtests/common/shared.c
```diff
@@ -1083,6 +1083,24 @@ int ft_complete_connect(struct fid_ep *ep, struct fid_eq *eq)
...
+	    info->domain_attr->name != fi_pep->domain_attr->name)
```
shefty, line 1098: Same -- remove.  I'm not sure the domains names are guaranteed to match.  Simply check that info->domain_attr->name is non-null.

shefty (changes requested):

fabtests/common/shared.c
```diff
@@ -1103,6 +1117,10 @@ int ft_retrieve_conn_req(struct fid_eq *eq, struct fi_info **fi)
 		return ret;
 	}
 
+	if (ft_verify_info(fi_pep, entry.info)) {
+		return -FI_EOTHER;
```
shefty, line 1121: nit: Since ft_verify_info returns an error code, it would be better to capture that return code and pass it through, rather than converting it to FI_EOTHER.

Juee14Desai: Would this be better:

if (entry.info != NULL){
        ret = ft_verify_info(fi_pep, entry.info);
        return ret;
}

Juee14Desai:
Sockets msg ep tests are failing. Would this validation not work for sockets provider?

fabtests/common/shared.c
shefty: entry.info needs to be non-null here.  It's better to set ret = ft_verify_info(...) and return ret.

shefty:
Hmm... You may need to break apart the if statements and print an error to identify what portion of the check failed.  The checks should work for any provider.  If sockets is failing, can you identify why?  Maybe the provider isn't correct.

Juee14Desai:
> Hmm... You may need to break apart the if statements and print an error to identify what portion of the check failed. The checks should work for any provider. If sockets is failing, can you identify why? Maybe the provider isn't correct.

For sockets it looks like the prov_name is NULL and the api versions of info->fabric_attr (0) and fi_pep->fabric_attr (65554) do not match.

shefty:
That sounds like bugs in the provider.  We may need to add a fix to it as part of this PR for the tests to pass.

shefty:
It does look like the sockets provider is not setting those fields correctly.  prov_name is left as NULL.  And I think abi_version isn't set either.  The good news is that this means the test is working. :)

Juee14Desai:
AWS CI failed.

shijin-aws:
It's a bug in our CI code. We will rerun the test after it's fixed.

shijin-aws:
bot:aws:retest

wenduwan:
bot:aws:retest

Juee14Desai:
Hi @shefty, Please let me know if there are any more comments.

prov/sockets/src/sock_ep.c
```diff
@@ -1417,7 +1417,8 @@ static void sock_set_fabric_attr(void *src_addr, const struct fi_fabric_attr *hi
 	if (!attr->name)
 		attr->name = strdup(sock_fab_name);
 
-	attr->prov_name = NULL;
+	attr->prov_name = hint_attr->prov_name;
```
shefty, line 1420: This needs to be strdup.  Otherwise it's pointing the prov_name to the same string that hint_attr points to.  If hint_attr are freed, attr->prov_name will reference invalid memory.

prov/sockets/src/sock_ep.c
```diff
@@ -1498,11 +1499,15 @@ struct fi_info *sock_fi_info(uint32_t version, enum fi_ep_type ep_type,
 {
 	struct fi_info *info;
 
-	info = fi_allocinfo();
+	info = fi_dupinfo(hints);
 	if (!info)
 		return NULL;
+	free(info->src_addr);
+	free(info->dest_addr);
```
shefty, line 1506: To be safe, please init info->src_addr and info->dest_addr to NULL after calling free.  src_addr won't matter, since we set it below.  But dest_addr might.

src/fabric.c
```diff
@@ -977,7 +977,9 @@ void DEFAULT_SYMVER_PRE(fi_freeinfo)(struct fi_info *info)
 		}
 		if (info->fabric_attr) {
 			free(info->fabric_attr->name);
-			free(info->fabric_attr->prov_name);
+			if (info->fabric_attr->prov_name == "sockets" ||
+                            info->fabric_attr->prov_name == NULL)
+				free(info->fabric_attr->prov_name);
```
shefty, line 982: See the comment on the previous patch about using strdup.  That's likely leading to this crash.

We shouldn't have provider specific checks in this common path.  This likely fixes the crash by instead leaking memory.

shefty (changes requested):

Juee14Desai:
Intel CI error is unrelated to this change.

shefty:
thanks!

PR #9360, opened 2023-09-21
ofi_list: Add dlist_size and slist_size

sunkuamzn:
Useful for printing the size of the list during debugging

### DIFF
--- a/include/ofi_list.h
+++ b/include/ofi_list.h
@@ -57,15 +57,23 @@ enum ofi_list_end {
 struct dlist_entry {
 	struct dlist_entry	*next;
 	struct dlist_entry	*prev;
+	OFI_DBG_VAR(ssize_t, size)
+	OFI_DBG_VAR(struct dlist_entry *, list)
 };
 
+#if ENABLE_DEBUG
+#define DLIST_INIT(addr) { addr, addr, 0, addr }
+#else
 #define DLIST_INIT(addr) { addr, addr }
+#endif
 #define DEFINE_LIST(name) struct dlist_entry name = DLIST_INIT(&name)
 
 static inline void dlist_init(struct dlist_entry *head)
 {
 	head->next = head;
 	head->prev = head;
+	OFI_DBG_SET(head->size, 0);
+	OFI_DBG_SET(head->list, head);
 }
 
 static inline int dlist_empty(struct dlist_entry *head)
@@ -80,6 +88,8 @@ dlist_insert_after(struct dlist_entry *item, struct dlist_entry *head)
 	item->prev = head;
 	head->next->prev = item;
 	head->next = item;
+	OFI_DBG_SET(item->list, head->list);
+	OFI_DBG_ADD(item->list->size, 1);
 }
 
 static inline void
@@ -95,6 +105,8 @@ static inline void dlist_remove(struct dlist_entry *item)
 {
 	item->prev->next = item->next;
 	item->next->prev = item->prev;
+	OFI_DBG_ADD(item->list->size, -1);
+	OFI_DBG_SET(item->list, NULL);
 }
 
 static inline void dlist_remove_init(struct dlist_entry *item)
@@ -155,6 +167,18 @@ static inline void dlist_remove_init(struct dlist_entry *item)
 	     (tmp) = (container)->member.prev)
 
 typedef int dlist_func_t(struct dlist_entry *item, const void *arg);
+
+#if ENABLE_DEBUG
+static inline ssize_t dlist_size(struct dlist_entry *item)
+{
+	return item->list->size;
+}
+#else
+static inline ssize_t dlist_size(struct dlist_entry *item)
+{
+	return -1;
+}
+#endif
 
 static inline struct dlist_entry *
 dlist_find_first_match(struct dlist_entry *head, dlist_func_t *match,
@@ -218,6 +242,15 @@ static inline void dlist_splice_head(struct dlist_entry *head,
 	/* put first element of 'to_splice' as first element of 'head' */
 	head->next = to_splice->next;
 	head->next->prev = head;
+
+	OFI_DBG_ADD(head->size, to_splice->size);
+#if ENABLE_DEBUG
+	struct dlist_entry *item;
+	dlist_foreach(head, item) {
+		item->list = head->list;
+		item->size = head->size;
+	}
+#endif
 
 	/* set list to empty */
 	dlist_init(to_splice);
--- a/include/ofi_list.h
+++ b/include/ofi_list.h
@@ -393,11 +393,13 @@ struct slist_entry {
 struct slist {
 	struct slist_entry	*head;
 	struct slist_entry	*tail;
+	OFI_DBG_VAR(ssize_t, size)
 };
 
 static inline void slist_init(struct slist *list)
 {
 	list->head = list->tail = NULL;
+	OFI_DBG_SET(list->size, 0);
 }
 
 static inline int slist_empty(struct slist *list)
@@ -415,6 +417,7 @@ static inline void slist_insert_head(struct slist_entry *item, struct slist *lis
 	}
 
 	list->head = item;
+	OFI_DBG_ADD(list->size, 1);
 }
 
 static inline void slist_insert_tail(struct slist_entry *item, struct slist *list)
@@ -426,6 +429,7 @@ static inline void slist_insert_tail(struct slist_entry *item, struct slist *lis
 
 	item->next = NULL;
 	list->tail = item;
+	OFI_DBG_ADD(list->size, 1);
 }
 
 static inline struct slist_entry *slist_remove_head(struct slist *list)
@@ -435,8 +439,10 @@ static inline struct slist_entry *slist_remove_head(struct slist *list)
 	item = list->head;
 	if (list->head == list->tail)
 		slist_init(list);
-	else
+	else {
 		list->head = item->next;
+		OFI_DBG_ADD(list->size, -1);
+	}
 #if ENABLE_DEBUG
 	if (item) {
 		item->next = NULL;
@@ -461,6 +467,18 @@ static inline struct slist_entry *slist_remove_head(struct slist *list)
 	} while (0)
 
 typedef int slist_func_t(struct slist_entry *item, const void *arg);
+
+#if ENABLE_DEBUG
+static inline ssize_t slist_size(struct slist *list)
+{
+	return list->size;
+}
+#else
+static inline ssize_t slist_size(struct slist *list)
+{
+	return -1;
+}
+#endif
 
 static inline struct slist_entry *
 slist_find_first_match(const struct slist *list, slist_func_t *match,
@@ -505,6 +523,7 @@ static inline void slist_remove(struct slist *list,
 
 	if (!item->next)
 		list->tail = prev;
+	OFI_DBG_ADD(list->size, -1);
 }
 
 static inline struct slist_entry *
@@ -558,8 +577,9 @@ slist_splice_head(struct slist *dst, struct slist *src)
 	src->tail->next = dst->head;
 	dst->head = src->head;
 
-	slist_init(src);
+	OFI_DBG_ADD(dst->size, src->size);
 
+	slist_init(src);
 	return dst;
 }
 
@@ -587,8 +607,9 @@ slist_splice_tail(struct slist *dst, struct slist *src)
 	dst->tail->next = src->head;
 	dst->tail = src->tail;
 
-	slist_init(src);
+	OFI_DBG_ADD(dst->size, src->size);
 
+	slist_init(src);
 	return dst;
 }
 


include/ofi_list.h
```diff
@@ -457,6 +468,17 @@ static inline struct slist_entry *slist_remove_head(struct slist *list)
...
+static inline size_t slist_size(struct slist *list)
+{
+	size_t size = 0;
+	struct slist_entry *cur, *prev;
+
+	slist_foreach(list, cur, prev) {
+		size++;
+	}
+	return size;
+}
```
shefty, line 480: We probably want both of these to be defined only for debug builds.  The implementation is slow, and we want to make sure someone doesn't rely on these function in production code, assuming that they were implemented efficiently.

shefty: On further thought, instead of making these calls debug only, I'd add a debug size variable that's incremented/decremented on list add/remove.  That way, the overhead of calling size() is minimized during debug builds, which will allow it to be checked more frequently.  See OFI_DGB_VAR/ADD in ofi_osd.h.

sunkuamzn: Where should this size variable go? It could go on the head of the list or on every entry. But updating would be messy in both cases with entry removals and splicing.

shefty: Uhm, yeah, this would be easy for slist.  Not so much for dlist, depending on how dlist is used.

Looking at dlist, we could definitely use some debug level tracking to help catch bugs.  This may be more work that you're wanting to do now.  I wonder if we could start with something like this:
```
struct dlist_entry {
        struct dlist_entry      *next;
        struct dlist_entry      *prev;
+
+       struct dlist_entry      *list;
 };
 
 #define DLIST_INIT(addr) { addr, addr }
@@ -66,6 +68,8 @@ static inline void dlist_init(struct dlist_entry *head)
 {
        head->next = head;
        head->prev = head;
+
+       head->list = head;
 }
 
 static inline int dlist_empty(struct dlist_entry *head)
@@ -76,6 +80,10 @@ static inline int dlist_empty(struct dlist_entry *head)
 static inline void
 dlist_insert_after(struct dlist_entry *item, struct dlist_entry *head)
 {
+       assert(!item->list || item->list == item);
+       assert(head->list);
+       item->list = head->list;
+
        item->next = head->next;
        item->prev = head;
        head->next->prev = item;
@@ -93,6 +101,9 @@ dlist_insert_before(struct dlist_entry *item, struct dlist_entry *he
 
 static inline void dlist_remove(struct dlist_entry *item)
 {
+       assert(item->list);
+       item->list = NULL;
+
        item->prev->next = item->next;
        item->next->prev = item->prev;
 }
```
The list accesses should be debug only, and I only spent about 5 minutes adding this.  I could easily have missed something, plus I didn't try to update all the calls.  Anyway, the goal is to have each entry reference the head of the list that the entry is in.  If we can add this tracking, then adding a size to the head of the list is doable.  A spliced list would sum the 2 sizes together.

sunkuamzn: @shefty can you take a look at this version?

sunkuamzn:
AWS CI failures are static assert failures. I will fix them.

include/ofi_list.h
```diff
@@ -555,6 +568,7 @@ slist_splice_head(struct slist *dst, struct slist *src)
 
 	slist_init(src);
 
+	OFI_DBG_ADD(dst->size, src->size);
```
shefty, line 597: This needs to go before the slist_init(), which resets src->size to 0.

include/ofi_list.h
```diff
@@ -388,11 +388,13 @@ struct slist_entry {
 struct slist {
 	struct slist_entry	*head;
 	struct slist_entry	*tail;
+	OFI_DBG_VAR(size_t, size);
```
shefty, line 417: Because of how this macro works, we can't add the ';' at the end of the line.  Some compilers don't like that.

include/ofi_list.h
```diff
@@ -57,15 +57,23 @@ enum ofi_list_end {
 struct dlist_entry {
 	struct dlist_entry	*next;
 	struct dlist_entry	*prev;
+	OFI_DBG_VAR(size_t, size);
+	OFI_DBG_VAR(struct dlist_entry *, list);
```
shefty, line 61: Remove ';' from end of the OFI_DBG_VAR() macro lines

include/ofi_list.h
```diff
@@ -95,6 +105,7 @@ static inline void dlist_remove(struct dlist_entry *item)
 {
 	item->prev->next = item->next;
 	item->next->prev = item->prev;
+	OFI_DBG_ADD(item->list->size, -1);
```
shefty, line 108: Please add ```OFI_DBG_SET(item->list, NULL)```

include/ofi_list.h
```diff
@@ -151,6 +162,13 @@ static inline void dlist_remove_init(struct dlist_entry *item)
 
 typedef int dlist_func_t(struct dlist_entry *item, const void *arg);
 
+#if ENABLE_DEBUG
+static inline size_t dlist_size(struct dlist_entry *item)
+{
+	return item->list->size;
+}
+#endif
```
shefty, line 170: Same as for slist, please add a non-debug version to return either 0 or -1

shefty (changes requested):
Thanks!  I think this will be very useful for debugging.  See comments for changes

sunkuamzn:
I won't be able to work on this for the next few weeks

include/ofi_list.h
```diff
@@ -435,8 +439,10 @@ static inline struct slist_entry *slist_remove_head(struct slist *list)
 	item = list->head;
 	if (list->head == list->tail)
 		slist_init(list);
-	else
+	else {
```
shefty, line 475: nit: Please add braces to the if side if needed on the else side.

shefty (commented):
thanks!  no idea about the CI failures, but the changes look good to me.

j-xiong:
Close for now due to lack of activity to address the remaining review comments.

include/ofi_list.h
```diff
@@ -219,6 +243,15 @@ static inline void dlist_splice_head(struct dlist_entry *head,
 	head->next = to_splice->next;
 	head->next->prev = head;
 
+	OFI_DBG_ADD(head->size, to_splice->size);
+#if ENABLE_DEBUG
+	struct dlist_entry *item;
+	dlist_foreach(head, item) {
+		item->list = head->list;
+		item->size = head->size;
```
shefty: Setting size for an item doesn't seem needed and results in an inconsistent value of size for the elements on the list: some will be 0, others will have a value.

sunkuamzn: Thank you, fixed!

sunkuamzn:
@aingerson could you take a look at the change to `struct fi_peer_rx_entry`?

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -638,6 +638,9 @@ int efa_rdm_pke_write(struct efa_rdm_pke *pkt_entry)
 				   conn->ep_addr->qpn, conn->ep_addr->qkey);
 	}
 
+#if ENABLE_DEBUG
+		dlist_insert_tail(&pkt_entry->dbg_entry, &ep->tx_pkt_list);
```
aingerson: indentation is off. this also feels like it needs to be in a separate patch

aingerson (commented):
Please update the second patch to include reference to adding the dlist list variable as well and a description of what it does.
peer changes look ok to me, thanks!

include/rdma/providers/fi_peer.h
```diff
@@ -161,8 +162,7 @@ struct fid_peer_srx;
 
 /* Castable to dlist_entry */
 struct fi_peer_rx_entry {
-	struct fi_peer_rx_entry *next;
-	struct fi_peer_rx_entry *prev;
+	struct dlist_entry dlist_entry;
```
shefty: This header file is exposed externally.  It cannot include internal headers, such as ofi_list.h.  That's why the code intentionally avoided dlist and had explicit pointers.

prov/efa/src/rdm/efa_rdm_pke.c
shefty: I agree.  Changes outside of changing dlist should be in separate commits.

include/rdma/providers/fi_peer.h
aingerson: Oh duh. Thanks for catching that, Sean!

sunkuamzn: Hmm how can I implement this change then? Without this change, `struct fi_peer_rx_entry` does not include the new members of `struct dlist_entry` and that causes memory corruption when running with debug enabled

Can I import `include/ofi_osd.h` and add the new fields like so?

```suggestion
struct fi_peer_rx_entry {
	struct fi_peer_rx_entry *next;
	struct fi_peer_rx_entry *prev;
	OFI_DBG_VAR(ssize_t, size)
	OFI_DBG_VAR(struct fi_peer_rx_entry *, list)
```

sunkuamzn: Or should I instead do something like

```
struct fi_peer_rx_entry {
	struct fi_peer_rx_entry *next;
	struct fi_peer_rx_entry *prev;
#ifndef NDEBUG
	ssize_t size;
	struct fi_peer_rx_entry * list;
#endif
```

shefty: The structure is protected by the external facing ABI.  An external provider may be built against these headers.  That provider may or may not be built in debug mode, and libfabric or the peer provider may be built differently.

Since you're changing dlist_entry, the comment about the structure being castable to dlist_entry no longer applies and should be removed.

Where is struct fi_peer_rx_entry allocated?  The only place I can quickly find is lnx.h (struct lnx_rx_entry).  Is that it?

sunkuamzn: It's used in the peer SRX mechanism. It's a member of `util_rx_entry`
https://github.com/ofiwg/libfabric/blob/5933367e8e9fc3568d9b86868b2e93084c979c02/include/ofi_util.h#L1289
https://github.com/ofiwg/libfabric/blob/5933367e8e9fc3568d9b86868b2e93084c979c02/prov/util/src/util_srx.c#L41

aingerson: Honestly, I don't even think these fields are even needed in the peer_rx_entry. The owner implementation of it will always have a wrapper struct around it that can have the entry depending on the implementation. I don't see a reason to require that for use in the peer. I think the idea was to use that entry to track unexpected or in progress messages but in all cases, the peer is going to have an extra struct to track that can have the entry. There's no need to share that between the owner and peer.

shefty: We currently require peer providers to be built against the same version of libfabric.  I'd be okay with adding the requirement that the providers must also both be either debug or release builds.  If we go that route, then your second option of ifndef NDEBUG may work.  At least I think that's portable...

shefty: @aingerson - If they can be removed, that's even better.

aingerson: Yeah I feel confident about it. I can prepare a patch

sunkuamzn: Thanks @aingerson!

aingerson: Ok I think I got all the references to it. See PR #11207  If I missed one, things will crash horribly because it will try to use the srx pointer as a dlist entry so it'll be easy to catch anything I missed :)

shefty: Sooo... when looking at PR #11207, the dlist_entry is used both with a dlist and slist.  The casting of it to slist likely breaks the debuggability trying to be introduced here.  I couldn't tell if the dlist_entry was always used one way or the other, or if the use changed based on application use (e.g. any address versus matched).

aingerson: That's a good point... the same dlist_entry field can be either one. But I think it should be ok since the debugging fields are only valid when it's in a list and it should only be in one list (slist or dlist) at once. dlist and slist fields are meaningless when outside of a list

aingerson: Looking at the dlist code though, I actually don't see an easy way to standardize these debug values. The slist is probably ok since you have a designated list struct and an entry struct but since the dlist entries can be both heads and items, I think you have a problem. The implementation is based on these generic insert before or after calls where you pretend on entry is the head when it actually is just a random entry in the list. Does the size indicate how many entries go after it? Do you iterate through the list on every insert call and update the whole list to increment one? It gets messy really quickly.
For example, if I have a list that looks like this 1->2->3->4 and then I call dlist_insert_after(5,2), the list should end up looking like this: 1->2->5->3->4. But how do you go to entry->1 and say 1->size++? Right now it looks like you just increment the new item's size by 1 but that doesn't make any sense either. Is 5's size 5 because it's in a list of 5? Or is it 3 because there are 3 after it? I'm not sure this is a consistent solution

aingerson: Update - reread and you do a size++ of the previous list which I think makes more sense. ie in my example 5->list = 2->list, 5->list++ so I think that does work after all

sunkuamzn: @aingerson For the dlist, I added a `list` pointer which points to the first entry that the dlist was initialized with. Whenever a new entry is added, I update the `list` pointer of the new entry to the `list` pointer from its neighbor. The `size` is only stored on the `list` entry. So if I implemented this correctly, `entry->list->size` would always be equal to the total list length.

While typing this out, I just realized I need to handle the case when the `list` entry itself is removed from the dlist. In that case, I would need to pick another `list` entry, iterate over the entire list and update the `list` pointer for all of the remaining entries in the list.

shefty: I considered this scenario, but I'm skeptical that it occurs.  I would assert(entry != entry->list || entry->list->size == 1).  That is, either there's always a specific list head, or the item is in a list by itself.  I wouldn't do anything fancy unless we're hitting that assert.

include/ofi_list.h
```diff
@@ -219,6 +248,14 @@ static inline void dlist_splice_head(struct dlist_entry *head,
 	head->next = to_splice->next;
 	head->next->prev = head;
 
+	OFI_DBG_ADD(head->size, to_splice->size);
+#if ENABLE_DEBUG
+	struct dlist_entry *item;
+	dlist_foreach(head, item) {
+		item->list = head;
+	}
```
shefty, line 293: To be safe in a generic use case splicing into the middle of a list, you want ```head->list->size``` and ```head->list``` instead of ```head->size``` and ```head```, respectively.

prov/efa/src/rdm/efa_rdm_pke.c
sunkuamzn: Fixed

include/rdma/providers/fi_peer.h
sunkuamzn: I added `assert(entry != entry->list || entry->list->size == 1)`

include/ofi_list.h
sunkuamzn: Fixed, thank you!

sunkuamzn:
@shefty for `dlist_entry_in_list`, is it better to include the slow iterative implementation for the non-debug case or remove it so that providers that use `dlist_entry_in_list` without debug will fail? I included the slow implementation now

shefty:
I think keeping the dlist_entry_in_list() as a search function is fine.  The function itself is simple enough, and maybe someone will have a legitimate use case.

sunkuamzn:
bot:aws:retest

sunkuamzn:
bot:aws:retest

aingerson (approved):

sunkuamzn:
bot:aws:retest

include/ofi_list.h
```diff
@@ -80,6 +88,10 @@ dlist_insert_after(struct dlist_entry *item, struct dlist_entry *head)
 	item->prev = head;
 	head->next->prev = item;
 	head->next = item;
+
+	OFI_DBG_SET(item->list, head->list);
+	assert(item->list);
+	OFI_DBG_ADD(item->list->size, 0);
```
shefty: add 1

include/ofi_list.h
```diff
@@ -95,6 +107,17 @@ static inline void dlist_remove(struct dlist_entry *item)
...
 	item->prev->next = item->next;
 	item->next->prev = item->prev;
+
+#if ENABLE_DEBUG
+	/* TODO: Fix usage by some providers that remove dlist entries that have
+	 * not been inserted in to any list */
+	if (item->list) {
+		/* Make sure the list is empty when removing the head */
+		assert(item != item->list || item->list->size == 0);
+		OFI_DBG_ADD(item->list->size, -1);
```
shefty: It looks odd seeing a check that size can be 0, then decremented to a negative value.  Maybe we only get into this flow calling remove_init(), and this doesn't matter...

If dlist_remove() is called on the head of a list that's empty, the result of the call today is a no-op.  I like setting list = NULL in this flow.  I'm uncertain about setting size = -1.  The majority of dlist_entry's that are not in a list will likely have size = 0.

include/ofi_list.h
sunkuamzn: Fixed

include/ofi_list.h
sunkuamzn: Fixed

PR #9361, opened 2023-09-22
prov/efa: Do not try to register memory for RDMA write

sunkuamzn:
Application is responsible for registering memory for one sided operations, so Libfabric is not supposed to handle the memory registration

EFA provider tries to do memory registration in the read path because EFA provider can use the long read protocol for sending messages and those buffers need not be registered by the application

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -1458,7 +1458,6 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
 
 	assert(ope->iov_count > 0);
 	assert(ope->rma_iov_count > 0);
-	efa_rdm_ope_try_fill_desc(ope, 0, FI_WRITE);
 	ep = ope->ep;
 	if (ope->bytes_write_total_len == 0) {
 		/* According to libfabric document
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1513,15 +1512,7 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
...
-		if (!ope->desc[iov_idx]) {
-			/* efa_rdm_ope_try_fill_desc() did not fill the desc,
-			 * which means memory registration failed.
-			 * return -FI_EAGAIN here will cause user to run progress
-			 * engine, which will cause some memory registration
-			 * in MR cache to be released.
-			 */
-			return -FI_EAGAIN;
-		}
+		assert(ope->desc[iov_idx]);
```
lrbison, line 1515: Hm, I think if application calls write without a desc that's illegal, but I'm not sure it's an assert, should probably be a FI_EINVAL?

shijin-aws: Another concern is for inject_write, which doesn't have a desc passed by application. We currently use emulated write for that but I am concerned that if we in the future choose to use rdma write for inject we will miss this pitfall.

shijin-aws: +1 f or Luke, here we should return FI_EINVAL as it's not a valid user case. assert should be used to validate something we have guaranteed in our own code.

sunkuamzn:
Canceling this PR because Libfabric does not require registration of the source buffer

PR #9362, opened 2023-09-22
prov/efa: Use long CTS protocol if long read protocol fails with ENOMEM

sunkuamzn:
Long read protocol could fail with ENOMEM if the EFA provider is unable to register the buffer with the NIC. In that case, we should fall back to Long CTS instead of returning ENOMEM to app


prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -172,7 +172,8 @@ ssize_t efa_rdm_msg_post_rtm(struct efa_rdm_ep *ep, struct efa_rdm_ope *txe, int
 
 	if (rtm_type < EFA_RDM_EXTRA_REQ_PKT_BEGIN) {
 		/* rtm requires only baseline feature, which peer should always support. */
-		return efa_rdm_ope_post_send(txe, rtm_type);
```
sunkuamzn, line 175: I could return the result of`1efa_rdm_ope_post_send` here because the error handling is only for long read protocl which is not selected here

PR #9363, opened 2023-09-22
prov/efa: Fix coverity warning in efa_mr_reg_impl

sunkuamzn:
If an applications calls fi_mr_reg with an incomplete attr, a segfault could occur

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -900,8 +900,8 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *at
 				ret,
 				fi_strerror(-ret),
 				efa_mr->mr_fid.key,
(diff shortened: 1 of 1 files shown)


shijin-aws:
I don't think it's a legal `mr_attr` with `mr_iov` not assigned. But I think we can check it before print

```
EFA_WARN(FI_LOG_MR,
				"Unable to register shm MR. errno: %d err_msg: (%s) key: %ld buf: %p len: %zu\n",
				ret,
				fi_strerror(-ret),
				efa_mr->mr_fid.key,
				mr_attr.mr_iov ? mr_attr.mr_iov->iov_base : NULL,
				mr_attr.mr_iov ? mr_attr.mr_iov->iov_len: NULL);
```

shijin-aws (approved):

shijin-aws:
Would be nice to have it in 1.19.x too. Can you backport

PR #9367, opened 2023-09-25
prov/efa: Do not create SHM endpoint when SHM is disabled

a-szegel:
When the user sets FI_EFA_ENABLE_SHM_TRANSFER=0, do not create a SHM endpoint.  The current behavior is that we just don't send via SHM when one sets FI_EFA_ENABLE_SHM_TRANSFER=0, but we still create and poll the peer endpoint.  New behavior is to not create SHM endpoint when peer is disabled.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -140,7 +140,11 @@ static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *in
 {
 	int err;
 
-	efa_shm_info_create(info, &efa_domain->shm_info);
+	efa_domain->shm_info = NULL;
+	if (efa_env.enable_shm_transfer)
+		efa_shm_info_create(info, &efa_domain->shm_info);
+	else
+		EFA_INFO(FI_LOG_CORE, "EFA will not use SHM for intranode communication because FI_EFA_ENABLE_SHM_TRANSFER=0\n");
 
 	if (efa_domain->shm_info) {
 		err = fi_fabric(efa_domain->shm_info->fabric_attr,
--- a/prov/efa/src/efa_env.c
+++ b/prov/efa/src/efa_env.c
@@ -182,7 +182,7 @@ void efa_env_define()
 	fi_param_define(&efa_prov, "tx_queue_size", FI_PARAM_INT,
 			"Defines the maximum number of unacknowledged sends with the NIC.");
 	fi_param_define(&efa_prov, "enable_shm_transfer", FI_PARAM_INT,
-			"Enable using SHM provider to perform TX operations between processes on the same system. (Default: 1)");
+			"Enable using SHM provider to perform TX/RX operations between processes on the same system. (Default: 1)");
 	fi_param_define(&efa_prov, "use_zcpy_rx", FI_PARAM_INT,
 			"Enables the use of application's receive buffers in place of bounce-buffers when feasible. (Default: 1)");
 	fi_param_define(&efa_prov, "set_cuda_sync_memops", FI_PARAM_INT,


prov/efa/src/efa_domain.c
```diff
@@ -140,7 +140,9 @@ static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *in
 {
 	int err;
 
-	efa_shm_info_create(info, &efa_domain->shm_info);
+	efa_domain->shm_info = NULL;
+	if (efa_env.enable_shm_transfer)
+		efa_shm_info_create(info, &efa_domain->shm_info);
```
darrylabbate, line 145: Consider some logging, e.g.
```suggestion
	if (efa_env.enable_shm_transfer)
		efa_shm_info_create(info, &efa_domain->shm_info);
	else
		EFA_INFO(FI_LOG_CORE, "Skipping fi_info allocation for SHM.\n");
```

wzamazon:
What if two endpoints have different environemnt variable setting?

a-szegel:
I think if a user configures FI_EFA_ENABLE_SHM=0, they should not receive any messages from a SHM endpoint.  I think it is a bug in EFA that a user can turn off SHM, and still get messages from SHM.  Keeping SHM alive when it was turned off by the user hurts performance of Apps that do not use EFA + SHM (additional allocation of resources + more polling).

prov/efa/src/efa_domain.c
a-szegel: done.

wzamazon:
>I think if a user configures FI_EFA_ENABLE_SHM=0, they should not receive any messages from a SHM endpoint. I think it is a bug in EFA that a user can turn off SHM, and still get messages from SHM. 

I think the most a user can do is to instruct EFA not to send data using shm. A user cannot control how other side of the communication want to use shm or not.

a-szegel:
I disagree.  If a user turns off SHM, I think they turn off both sending + receiving from SHM for that endpoint.  Are there any users you know about which require receiving but no sending from SHM?

a-szegel:
Would a documentation update to the parameter make you feel better about this change?

wzamazon:
>I disagree. If a user turns off SHM, I think they turn off both sending + receiving from SHM for that endpoint.

My point is a user cannot turn off receiving using SHM, because it has not control of sender's behavior.

Thinking in a server/client scenario, if server has `FI_EFA_ENABLE_SHM_TRANSER` set to 0, and client has `FI_EFA_ENABLE_SHM_TRANSER` set to 1.

> Are there any users you know about which require receiving but no sending from SHM?

I do not know any real world application would hit this issue. Fabtests will have problem. 

> Would a documentation update to the parameter make you feel better about this change?

I guess we can explicit state that both ends that use EFA must have same `FI_EFA_ENABLE_SHM_TRANSFER` setting.

One the other hand, how much do we gain by making this change (in terms of memory usage and performance)?

bwbarrett:
> > I disagree. If a user turns off SHM, I think they turn off both sending + receiving from SHM for that endpoint.
> 
> My point is a user cannot turn off receiving using SHM, because it has not control of sender's behavior.
> 
> Thinking in a server/client scenario, if server has `FI_EFA_ENABLE_SHM_TRANSER` set to 0, and client has `FI_EFA_ENABLE_SHM_TRANSER` set to 1.

I disagree.  If I tell Libfabric not to use the EFA provider by specifying `FI_PROVIDER=^efa`, I don't expect the provider to be loaded and running in case someone tries to talk to me over EFA.  I expect EFA to not work, and if I later try to communicate with that process over EFA, I expect it to fail.

If I care about being able to talk to another peer that does have SHM enabled, then I should enable SHM.  But if I want to disable SHM, then I want to disable SHM and I'm going to be upset if I later find that SHM is still running.

a-szegel:
This will also save users who do not require EFA + SHM 16MB of RAM per rank.

a-szegel:
Updated env var documentation to make it explicit.

a-szegel:
@wzamazon   Is there anything that you see from https://github.com/ofiwg/libfabric/commit/2beef0e897197e1c02371071db2e34c2451cf6ec which needs to be updated as part of this change?

wzamazon:
> Is there anything that you see from https://github.com/ofiwg/libfabric/commit/2beef0e897197e1c02371071db2e34c2451cf6ec which needs to be updated as part of this change? I think the peer provider change cleaned up a bunch of that stuff.

I do not think any more change are needed.

prov/efa/src/efa_domain.c
```diff
@@ -140,7 +140,11 @@ static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *in
...
+	else
+		EFA_INFO(FI_LOG_CORE, "EFA will be used for intranode communication because FI_EFA_ENABLE_SHM_TRANSFER=0\n");
```
wzamazon, line 147: will or "will NOT" ?

shijin-aws (approved):

PR #9368, opened 2023-09-25
core: Introduce new profiling interface for low-level statistics

peinanz:


include/rdma/fi_profile.h
```diff
@@ -0,0 +1,163 @@
...
+	size_t size;
+	const char *name;
+	const char *desc;
+};
+
+struct fi_profile_event_desc {
+	uint32_t id;
+	const char *name;
+	const char *desc;
+};
```
shefty, line 69: Let's merge the above structs into just fi_profile_desc.  As mentioned separately, this gives us some versioning protection on the data that's passed into a callback.  We can add new values to fi_datatype to define specific structures for libfabric defined data as needed, as well as something like FI_PROV_TYPE to indicate that it's a provider specific structure.  New types would carve out a portion of the enum space, similar to what's been done for collectives.

This also makes it easier to have a value be both a variable and an event, such that the app could monitor changes to the variable.

man/fi_profile.3.md
```diff
@@ -0,0 +1,222 @@
...
+*data_size*
+: The amount of data that the provider is provided in the "param" 
+when the callback is called.
+
+*data*
+: An uint64_t value of a variable.
+
+# DESCRIPTION
+
+Fabric profiling interface allows users to obtain some of configuration or runtime properties from the providers in the form of "variables". Providers define variables and make them readable to users through the profiling interface. Providers define events and use them to report an operation occured or a state changed where one or more variables may change their values. Users can register callback functions to one or more events, and receive notifications from providers once an event occurs.
```
shefty, line 106: Please add line breaks directly to the file at about 80 characters.  Otherwise some editors (vi) make this very difficult to read and edit.

man/fi_profile.3.md
```diff
@@ -0,0 +1,222 @@
...
+};
+{% endhighlight %}
+
+## id
+
+Each variable has a unique identifier (id). Providers provide this id and users use it to get variable value.
+
+## type
+
+The data type of the variable. Most variables are in basic type, i.e, FI_INT, FI_UINT32, etc. FI_VOID is used for a struct data type.
```
shefty, line 134: See comment on first patch.  It'll be easier on the user to explicitly indicate the structure, when needed.

man/fi_profile.3.md
```diff
@@ -0,0 +1,222 @@
...
+
+Each variable has a unique identifier (id). Providers provide this id and users use it to get variable value.
+
+## type
+
+The data type of the variable. Most variables are in basic type, i.e, FI_INT, FI_UINT32, etc. FI_VOID is used for a struct data type. 
+
+## size
+
+The size (in byte) of the variable data. It can be used to allocate space for the variable data, and is only useful for a struct data type.
```
shefty, line 138: size is useful for all data types, not just struct.  E.g. floating point variable data type sizes aren't well-defined.

man/fi_profile.3.md
```diff
@@ -0,0 +1,222 @@
...
+Reference to the name of a variable. It is either defined or allocated by providers. Variables with the same name should have an identical variable id. 
+
+## desc
+Reference to the description of a variable. It is either defined or allocated by providers.
+
+# EVENTS
+
+The profiling event is defined by the providers. It is used to notify users that an operation has occured or a state has been moved, which may change the values of some variables. Event is a way for providers to inform changes on the variables.  Like the variables, some events are pre-defined and shared by many providers, while others are provider-specific events and defined by each provider. Users can register a callback function on an event to get notified when the event occurs.
+
+UNEXP_MSG_RECD and UNEXP_MSG_MATCHED are common events which imply there are changes on variable UNEXP_MSG_CNT.
```
shefty, line 151: Here and elsewhere use the full matching name of any definitions.  Ie. FI_UNEXP_MSG_RECVD (with the FI prefix).  It's tedious, but makes it clear to a developer what needs to be in the code.

man/fi_profile.3.md
```diff
@@ -0,0 +1,222 @@
...
+Callback functions are used to receive notifications from a provider when an event occurs. A callback function is defined by the user and called by a provider as an additional operation, normally before or after data transfer. This will increase the overhead of the operation.  Callback functions should avoid having blocking calls, or taking non-trival time. And callback functions should limit interactions with a provider to reading the variables only.
+
+The signature of the callabck function is:
+
+{% highlight c %}
+int (*callback)(struct fid_profile *prof_fid, uint32_t event,
+                void *param, size_t size, void *context)
+{% endhighlight %}
+
+The *param* is provided by the user when registering the callabck, and the *size* is data contained in the *param*.
```
shefty, line 186: Param is passed by the provider, not the user.

man/fi_profile.3.md
```diff
@@ -0,0 +1,222 @@
...
+
+## fi_profile_reg_callback
+
+fi_profile_reg_callback registers a callback function to the event. The event should be in the event list obtained using the fi_profile_query_events, otherwise, an error code -FI_EINVAL will be returned. 
+
+Callback function can be registered with a parameter and its size. This allows the provider to add data into the memory space pointed by the parameter when calling the callback.
+
+## fi_profile_read_u64
+
+fi_profile_read_u64 call returns an uint64_t value of the specified variable. The variable should be in the variable list obtained using fi_profile_query_vars, otherwise, an error code -FI_EINVAL will be returned. Upon returning successfully, users can use the type field in fi_profile_var_desc to cast data to an appropriate type.
```
shefty, line 208: This should only be called for variables that are of type FI_UINT64.  The goal of having the wrapper function is that the app calls the correct wrapper based on the datatype.

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+*callback*
+: Callback function registered for an event.
+
+*param*
+: The parameter of the callback function.
+
+*size*
+: The amount of the data the provider is provided in the "param" 
+when the callback is called.
```
shefty, line 97: nit: indent "when" to align with "The", or just simplify the description.
```
: The amount of data referenced by the param parameter.
```

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+*data*
+: An uint64_t value of a variable.
```
shefty, line 100: nit: Remove the "uint64_t" portion.  E.g. "The current value of the variable".  (Since not all variables will end up being 64-bits.)

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+# DESCRIPTION
+
+Fabric profiling interface allows users to obtain some of configuration or
```
shefty, line 104: wording: "... users to obtain configuration or..."  - remove "some of"

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+runtime properties from the providers in the form of "variables". Providers 
+define variables and make them readable to users through the profiling 
+interface. Providers define events and use them to report an operation
```
shefty, line 107: wording: "... use them to report when an operation has occurred ..."

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+runtime properties from the providers in the form of "variables". Providers 
+define variables and make them readable to users through the profiling 
+interface. Providers define events and use them to report an operation 
+occured or a state changed where one or more variables may change their 
+values. Users can register callback functions to one or more events, and 
+receive notifications from providers once an event occurs.
+
+In order to support the profiling interface, an endpoing or a domain in a 
+provider needs to maintain a fid_profile object, and implement the profiling 
+operations there. To starts a profiling,  an open call must be made on the
```
shefty, line 114: wording: "starts -> start".  remove extra space before "an"

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+provider needs to maintain a fid_profile object, and implement the profiling 
+operations there. To starts a profiling,  an open call must be made on the 
+target endpoint or domain object to obtain the fid_profile object.  If the 
+target object does not support profiling, the open call will fail with error 
+code -FI_ENOSYS. Once a fid_profile object is obtained successfully, it can be 
+used to query available variables or events, and read variables' values.
+
+# VARIABLES
+
+The profiling variables are defined in providers, and become accessable
```
shefty, line 122: wording/spelling: "are defined by providers and are accessible through the profiling interface.  Variables are specified using struct fi_profile_desc described below.  A variable can be a configuration..."

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+target object does not support profiling, the open call will fail with error 
+code -FI_ENOSYS. Once a fid_profile object is obtained successfully, it can be 
+used to query available variables or events, and read variables' values.
+
+# VARIABLES
+
+The profiling variables are defined in providers, and become accessable 
+(readable) externally through the profiling interface.  A variable can be 
+a configuration setting, a runtime state or a counter in an endpoint or a 
+domain in a provider. Some variables are common and available in most
```
shefty, line 125: wording/punctuation: "...a runtime state, or a counter associated with an endpoint or domain."

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+# VARIABLES
+
+The profiling variables are defined in providers, and become accessable 
+(readable) externally through the profiling interface.  A variable can be 
+a configuration setting, a runtime state or a counter in an endpoint or a 
+domain in a provider. Some variables are common and available in most 
+providers, while others are provider-specific, and are only available in 
+one provider. Users can query available variables and their values using 
+the profiling interface.
```
shefty, line 128: add: "Provider specific variables are outside the scope of this man page.  libfabric defined variables and their datatypes are described below."

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+FI_VAR_UNEXP_MSG_CNT is a common variable and it counts the number of  
+unexpected message recevied and unmatched. The variable has a unint64_t 
+data value.
```
shefty, line 132: Reformat:
```
## FI_VAR_UNEXP_MSG_CNT (uint64_t)

This variable returns the current number of unexpected messages that have been received
by the provider which are waiting for matching receive buffers.
```
(This gives each variable its own clear section.

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+A variable is specified using struct fi_profile_desc and is described below.
```
shefty, line 134: remove (or rather move upward -- see above)

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+# EVENTS
+
+The profiling event is defined by the providers. It is used to notify users 
+that an operation has occured or a state has been moved, which may change
```
shefty, line 139: wording:
```
Profiling events notify users that an operation has occurred or the state of an object has
changed.  Events typically indicate that the values of one or more variables have been
modified.  Similar to variables, events may be pre-defined by libfabric or be provider
specific.  Users can register a callback to be notified when an event occurs. 
libfabric defined events and the data that corresponds with the event, if any, are described below.

Applications may read variable values from the event callback, but may not register or unregister
for events, query for profiling variable or events, or call the profile lock and unlock functions.
```

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+The profiling event is defined by the providers. It is used to notify users 
+that an operation has occured or a state has been moved, which may change 
+the values of some variables. Event is a way for providers to inform changes 
+on the variables. Similar to the variables, some events are pre-defined and 
+shared by many providers, while others are provider-specific events and 
+defined by each provider. Users can register a callback function on an event 
+to get notified when the event occurs.
+
+FI_EVENT_UNEXP_MSG_RECVD and FI_EVENT_UNEXP_MSG_MATCHED are common events 
+which imply there are changes on variable FI_VAR_UNEXP_MSG_CNT.
```
shefty, line 147: Rework similar to above to give these their own sections:
```
## FI_EVENT_UNEXP_MSG_RECVD (param = ?)

This event indicates that an unexpected message was received and has been
queued by the provider.

## FI_EVENT_UNEXP_MSG_MATCHED (param = ?)

This event indicates that an unexpected messages that was previously queued
by the provider has been matched.
 ```
We need to decide what information is passed to both of the above events.  This might
be a completion structure (e.g. fi_cq_tagged_entry) that contains the src_addr + tag + size, etc.

File: man/fi_profile.3.md [hunk 20]
```diff
@@ -0,0 +1,260 @@
...
+
+Same as variables, events are specified by the fi_profile_desc struct.
```
shefty, line 149: Start a new section here.  E.g.
```
# PROFILE DESCRIPTION

Variables and events are described using struct fi_profile_desc:
```

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+## size
+
+The size (in byte) of the data. "type" and "size" can be used to support 
+any provider-specific data.
+
+## name
+
+Reference to the name of a variable or an event. It is either defined or 
+allocated by providers. Variables or events with the same name should have
```
shefty, line 179: Wording: "Reference to the name of a variable or an event.  Storage is allocated and managed by the provider."

Similar wording for desc below.

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+This will increase the overhead of the operation.  Callback functions should 
+avoid having blocking calls, or taking non-trival time. And callback functions 
+should limit interactions with a provider to reading the variables only.
+
+The signature of the callabck function is:
+
+{% highlight c %}
+int (*callback)(struct fid_profile *prof_fid, uint32_t event,
+                void *param, size_t size, void *context)
+{% endhighlight %}
```
shefty, line 200: The above callback description should go under the fi_profile_reg_callback function below.

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+
+## fi_profile_open
```
shefty, line 203: Set to '#' to set function call as top-level item.  Apply this to all functions defined below.  E.g.
```
# fi_profile_open
open blah blah....

# fi_profile_close
close blah blah
```

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+of entries will be set in the buffer, and "-FI_ETOOSMALL" will be returned. 
+If the buffer is NULL, upon returning, the "count" indicates the total number 
+of entries available in the target endpoint or domain. User can allocate 
+space and call the function again.
+
+## fi_profile_reg_callback
+
+fi_profile_reg_callback registers a callback function to the event. The event 
+should be in the event list obtained using the fi_profile_query_events, 
+otherwise, an error code -FI_EINVAL will be returned.
```
shefty, line 233: Move callback description under here as a sub-bullet.

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+fi_profile_read_u64 is a wrapper function for FI_UINT64 type of varialbles.
+It returns an uint64_t value of the specified variable.  The variable should 
+be in the variable list obtained using fi_profile_query_vars, otherwise, 
+an error code -FI_EINVAL will be returned. 
+
+## fi_profile_lock / fi_profile_unlock
+
+fi_profile_lock and fi_profile_unlock calls are used to block/unblock 
+the concurrenct accessing of the profiling resources when needed.
```
shefty, line 245: Add:  "These calls are useful when reading multiple variables at the same time to ensure that the variable snapshot is consistent."

man/fi_profile.3.md
```diff
@@ -0,0 +1,260 @@
...
+
+Users should call fi_profile_close to release all resources allocated for 
+profiling in the provider.
+
+# RETURN VALUES
+
+Returns 0 on success.  On error, a negative value corresponding to fabric 
+errno is returned. If the provider doee not support profiling interface, 
+an -FI_ENOSYS will be returned. For fi_profile_query_vars and 
+fi_profile_query_events, a positvie return value indicates the total number
```
shefty, line 257: spelling: "positive".  The query calls should return an integer count (number of entries in the array), rather than the number of bytes.

shefty (commented):
I think this is good.  Comments are around fixing up wording and structure in the man page.

man/fi_profile.3.md
```diff
@@ -0,0 +1,281 @@
...
+
+## FI_EVENT_UNEXP_MSG_MATCHED
+
+This event indicates that an unexpected messages that was previously queued
+by the provider has been matched. 
+
+These two events can notify a user without any data and rely on the user 
+to read variables (e.g, FI_VAR_UNEXP_MSG_CNT) when needed.  It also
+can notify user with a fi_cq_tagged_entry data ("param") which contains 
+<src_addr, tag, size> to indicate how the event occured.
```
shefty, line 166: Basically repeat that "The callback will be passed a struct fi_unexp_msg_event (defined above) through the callback parameter."

shefty (commented):
Thanks - ready to merge.  I have a minor comment regarding the event param.  We can adjust what the param actually is when implementing the support.

man/fi_profile.3.md
```diff
@@ -0,0 +1,281 @@
...
+fi_profile_reg_callback registers a callback function to the event. The event 
+should be in the event list obtained using the fi_profile_query_events, 
+otherwise, an error code -FI_EINVAL will be returned. 
+
+## Callback
+
+It is used to receive notifications from a provider when an event occurs. 
+It is defined by the user and called by a provider as an additional operation,
+normally before or after data transfer.  This will increase the overhead of 
+the operation.  Callback functions should avoid having blocking calls, or
```
swelch, line 244: A nit, but I think it would be preferable to go with the semantic that "Callback functions must avoid blocking calls...". Time spent in this callback can be preventing hardware queues from progressing.

man/fi_profile.3.md
```diff
@@ -0,0 +1,281 @@
...
+Applications may read variable values from the event callback, but may not 
+register or unregister for events, query for profiling variable or events, 
+or call the profile lock and unlock functions
+
+libfarbic pre-defined events include:
+
+## FI_EVENT_UNEXP_MSG_RECVD 
+
+This event indicates that an unexpected message was received and has been
+queued by the provider.
```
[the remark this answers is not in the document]
peinanz: I was initially thinking of making the event simple with no param, completely depends on the user to pull the data. 

If defining this event data struct,  we need to add it into fi_profile.h. 
Can we expect for the same event, any provider will notify with the same type of "param"? If this is the case, we can define an event as <event, data_type, data_size>,  different events can be extended from a base "data_type" for easy expansion.

man/fi_profile.3.md
```diff
@@ -0,0 +1,281 @@
...
+## FI_EVENT_UNEXP_MSG_MATCHED
+
+This event indicates that an unexpected messages that was previously queued
+by the provider has been matched. 
+
+These two events can notify a user without any data and rely on the user 
+to read variables (e.g, FI_VAR_UNEXP_MSG_CNT) when needed.  It also
+can notify user with a fi_cq_tagged_entry data ("param") which contains 
+<src_addr, tag, size> to indicate how the event occured.
+
```
swelch, line 167: Is it a hole, that you can get the current number of unexpected messages, but not get information on any previously received unexpected messages (at time of opening the profiling) until those messages are matched with a posted receive? Is the intent for profiling be started before the EP is enabled?

peinanz: With the current pre-defined variable (unexpected message count) and events, we cannot get info on previously received unexpected messages.  This can be added as a provider-specific variable, say "unexpected_info_queue",  a provider needs to maintain this "queue",  and the user can use the profiling interface to query this variable value, to get the info in the queue.   We are looking for input/feedback to add more meaningful variables and events.  Profiling is enabled on EP after PE is created.

swelch: The CXI provider currently has a provider specific call that allows the number of unexpected messages to be queried and for retrieving the existing entries. They are returned in an array of CQ entries and an optional array of addresses (so no new cq format). The unexpected list itself is maintained by hardware. Ideally as you indicate, it would be great if this "pull" model could fit into the profile API. This work was done to provide CXI prototype support for unexpected messages in the MPI Message Queue Dumping Interface.

man/fi_profile.3.md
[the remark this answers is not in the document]
shefty: Any structure that we define here will need to be added to the header file.  And we need to define that the same parameter be used by all providers for common events.  Otherwise, there's nothing an app can realistically do with any of this.  It's the same as reading a variable.

man/fi_profile.3.md
shefty: @swelch - See my comment above.  We looked at using one of the existing cq formats, but those lack the ability to return a src_addr, which may be useful.  So I think we will end up with a new structure.  I think we work to get the pull model you're describing to work, which will help prove that the API is adaptable.  The pull model should be the same as reading a variable -- just a more complex one -- that can return an array of structs (like the one defined above).

include/rdma/fi_profile.h
```diff
@@ -0,0 +1,156 @@
...
+struct fid_profile;
+
+struct fi_profile_ops {
+	size_t size;
+	void (*reset)(struct fid_profile *prof_fid,  uint64_t flags);
+	int (*query_vars)(struct fid_profile *prof_fid,
+	                     struct fi_profile_desc *varlist, int *count);
+	int (*query_events)(struct fid_profile *prof_fid,
+	                     struct fi_profile_desc *eventlist, int *count);
+	int (*read_var)(struct fid_profile *prof_fid, uint32_t var_id, void *data);
```
shefty, line 74: Based on @swelch 's comments about pulling the unexpected list, we may want to change this function to return ssize_t instead of an int, plus add a size_t *size parameter.  We need to discuss how we want a pull model to work when the data being returned may be variable sized.  There are other options here, such as adding iterator calls.

swelch:
Dropping the current cxi function to pull the unexpected message list from the provider.
```
        /* Get unexpected message information.
         *
         * Obtain a list of unexpected messages associated with the endpoint.
         * The list is returned as an array of CQ tagged entries. The following
         * is how the fields in fi_cq_tagged_entry are used.
         *
         * op_context: NULL since this message has not matched a posted receive
         *      flags: A combination of FI_MSG, FI_TAGGED, FI_RECV,
         *      and/or FI_REMOTE_CQ_DATA
         *      len: Unexpected message request length
         *      data: Completion queue data (only valid if FI_REMOTE_CQ_DATA
         *      is set)
         *      tag: Unexpected message tag (only valid if FI_TAGGED is set)
         *
         * @ep: Endpoint FID to have unexpected messages returned to user.
         * @entry: Tagged entry array to be filled in by the provider. If the
         * entry is NULL, only ux_count will be set.
         * @count: Number of entries in entry and src_addr array. If count is
         * zero,then only the ux_count will be set on return.
         * @src_addr: Source address array to be filled in by the provider. If
         * the entry is NULL, only ux_count will be set.
         * @ux_count: Output variable used to return the number of unexpected
         * messages queued on the given endpoint.
         *
         * Return: On success, number of entries copied into the users entry
         * and src_addr arrays. On error, -FI_ERRNO.
         */
        size_t (*ep_get_unexp_msgs)(struct fid_ep *fid_ep,
                                    struct fi_cq_tagged_entry *entry,
                                    size_t count, fi_addr_t *src_addr,
                                    size_t *ux_count);

```

include/rdma/fi_profile.h
peinanz: Will change op for pull data:
ssize_t (*read_var)(struct fid_profile *prof_fid, uint32_t var_id, void *data, size_t *count);
        on input, *count is the max entries in the data.  On output, *count is the total entries currently 
         available for the var_id.  The return is the actual count of entries in the data buffer if successed.

callback for pushing data will be:
define a new struct, 
struct event_data {
        void *data;
        size_t count;    // # of entries in the data
};
int (*callback)(struct fid_profile *prof_fid, uint32_t event, uint64_t flags, struct event_data *data,
                       size_t count, void *context);
         flags can set control bit to limit pulling data in callback; 
         data will be the array of event_data struct; and 
         the count will be the # of event data in the data array. 

In this way, we should be able to support the "ex_get_unexpected_msg()" call below in either pull or push mode.
This will allow one call to pull or push data as one snapshot.

shefty: For read_var and the callback, I would use size instead of count.  Since data is a void *, I would allow the size to be interpreted as bytes, which provides more flexibility.  We don't know what future events may need to report.

We have the option to interpret size as a count for a given var_id or event, but I suspect that may end up confusing.

Please shift flags in the callback to the second to last parameter, just before void * context.  That aligns best with other APIs.  We don't need to define any flags for now, just having the parameter gives us flexibility in the future.

For the callback, it's easier if the provider can just return an array of whatever data they need.  The struct event_data that's defined above makes things difficult on the provider.  It forces them to allocate 2 arrays and link them together.

So...
```
ssize_t (*read_var)(prof_fid, u32 var_id, void *data, size_t *size);

int (*callback)(prof_fid, u32 event, void *data, size_t size, u64 flags, void *context);
```
To cover the pull model, we'll need a new variable, say FI_VAR_UNEXP_QUEUE.  That can use Ian's updated struct fi_cq_err_entry for output.  And that leads us to... somewhat of a mess.

enum fi_datatype works great for simple data types.  It's not so good for structs.  But enum fi_type is good for structs.  It even makes sense to add support for FI_TYPE_CQ_ERR_ENTRY.

What if we include both enum fi_type and enum fi_datatype in fi_profile_desc (maybe as a union)?  Then we just need a way to indicate which one applies, such as adding a u64 flags to the fi_profile_desc.

peinanz: Will add fi_type as union with fi_datatype.  Since we need to add a flags in fi_profile_desc, we can use a bit in there to indicate if the variable can be read inside a callback. We can remove the flags in the callback.

include/rdma/fi_profile.h
```diff
@@ -0,0 +1,156 @@
...
+	int (*reg_callback)(struct fid_profile *prof_fid, uint32_t event,
+	            int (*callback)(struct fid_profile *prof_fid, uint32_t event,
+	                    void *param, size_t size, void *context), 
+	            void *context);
+	void (*lock)(struct fid_profile *prof_fid);
+	void (*unlock)(struct fid_profile *prof_fid);
```
shefty, line 80: I was thinking about this last night, and I would rename these functions to start_reads / end_reads, plus add a flags field.  This more clearly indicates what the app is trying to do, without implying any specific implementation on the part of the provider.

peinanz: Ok, will change lock/unlock to start_reads/end_reads.

include/rdma/fi_profile.h
```diff
@@ -0,0 +1,167 @@
...
+		enum fi_datatype basic;
+		enum fi_type comp;
+	} type;
+	uint64_t flags;
+	size_t size;
+	const char *name;
+	const char *desc;
+};
+
+#define FI_VAR_TYPE_COMP  0x1   //  flags bit0,  0: basic, 1: comp
```
shefty, line 70: This isn't easily extendable or intuitive.  Please switch to something more like this:
```
enum fi_profile_type {
    fi_primitive_type,
    fi_defined_type,
};

struct fi_profile_desc _
    ...
    enum fi_profile_type datatype;
    union {
        enum fi_datatype primitive;
        enum fi_type defined;
    } datatype;
    ...
```

include/rdma/fi_profile.h
```diff
@@ -0,0 +1,167 @@
...
+	void (*reset)(struct fid_profile *prof_fid,  uint64_t flags);
+	ssize_t (*query_vars)(struct fid_profile *prof_fid,
+	                      struct fi_profile_desc *varlist, size_t *count);
+	ssize_t (*query_events)(struct fid_profile *prof_fid,
+	                       struct fi_profile_desc *eventlist, size_t *count);
+	ssize_t (*read_var)(struct fid_profile *prof_fid, uint32_t var_id,
+	                    void *data, size_t *size);
+	int (*reg_callback)(struct fid_profile *prof_fid, uint32_t event,
+	        int (*callback)(struct fid_profile *prof_fid, uint32_t event, 
+	                void *param, size_t size, uint32_t type, void *context),
```
shefty, line 85: I'm unsure about the type parameter here.  Now that type could be fi_datatype or fi_type, it's unclear.  I think the user will just need to know/get this from the fi_profile_desc that's associated with the event id.

Hmm... maybe instead of passing the u32 event to the callback, we should just pass struct fi_profile_desc *?  That includes the event, plus type information, name, etc., which avoids the caller needing to look it up via the context parameter.

shefty (commented):
very close -- change to add user-defined types made me rethink the callback

include/rdma/fi_profile.h
peinanz: We can use fi_profile_desc instead event_Id in the callback,  here the size and datatype in the fi_profile_desc will be for the data (param) of callback for that event.  We can pass fi_profile_desc for variable in read_var(),  this can reduce the looking up as well.

shefty: The provider owns the internal allocation of fi_profile_desc and the event id.  It's trivial for the provider to lookup the fi_profile_desc given an id.  I would leave read_var() as-is.

Having the provider hand fi_profile_desc to the user via the callback makes sense.  But the reverse isn't true (user passing an fi_profile_desc to the provider).

include/rdma/fabric.h
```diff
@@ -765,6 +766,7 @@ enum fi_type {
 	FI_TYPE_CQ_ATTR,
 	FI_TYPE_MR_ATTR,
 	FI_TYPE_CNTR_ATTR,
+	FI_TYPE_CQ_ERR_ENTRY,
```
shefty, line 769: We need a separate patch that adds support for this in fi_tostr()

shefty (approved):
minor comment for an additional update, but this looks ready to me.  @swelch - does the API match with your use case?

PR #9370, opened 2023-09-26
prov/shm: Add memory barrier before updating resp for atomic

shijin-aws:
In smr_progress_cmd_atomic, resp is updated after calling smr_progress_inline/inject_atomic, which involves a memcpy in smr_do_atmoic for compare and fetch cases. Currently, there is no memory barrier between the memcpy and the resp update, this can make resp updated before the memcpy is done on ARM, which has a weak memory model. This patch adds a memory barrier before updating the resp to make the operations serialized.


a-szegel (approved):

aingerson (approved):

shijin-aws:
I believe intel CI failure is not related. @aingerson

aingerson:
@shijin-aws Correct. shm tests all passed. Should be good to merge pending your CI. Thank you!

PR #9372, opened 2023-09-26
[v1.19.x] prov/util: backport genlock conversion to utilities

a-szegel:


shijin-aws (approved):

shefty (changes requested):
This is not a bug fix.  It's an optimization, which we need to revisit.  The man pages state that control APIs are always thread safe.  This includes the bind() functions, where the ep_list_lock is used.  I don't think we can disable that lock, and we need to re-examine the changes to main that did that.

PR #9376, opened 2023-09-27
fabtests/efa: set cuda read test rma_op name

wenduwan:
The cuda read test is using the default write rma_op, but we want to test the read path. Hence we should set the -o option explicitly.

### DIFF
--- a/fabtests/pytest/conftest.py
+++ b/fabtests/pytest/conftest.py
@@ -80,6 +80,18 @@ class CmdlineArgs:
         else:
             self.environments = environ[:]
 
+    def append_server_arguments(self, *args):
+        if self.additional_server_arguments is None:
+            self.additional_server_arguments = ""
+        
+        self.additional_server_arguments += " " + " ".join(args)
+
+    def append_client_arguments(self, *args):
+        if self.additional_client_arguments is None:
+            self.additional_client_arguments = ""
+        
+        self.additional_client_arguments += " " + " ".join(args)
+
     def populate_command(self, base_command, host_type, timeout=None, additional_environment=None):
         '''
             populate base command with informations in command line: provider, environments, etc
--- a/fabtests/pytest/efa/test_efa_protocol_selection.py
+++ b/fabtests/pytest/efa/test_efa_protocol_selection.py
@@ -34,6 +34,10 @@ def test_transfer_with_read_protocol_cuda(cmdline_args, fabtest_name, cntrl_env_
     cmdline_args_copy.append_environ(f"{cntrl_env_var}=1000")
     cmdline_args_copy.append_environ("FI_EFA_RUNT_SIZE=0")
 
+    if fabtest_name == "fi_rma_bw":
+            cmdline_args_copy.append_server_arguments("-o read")
+            cmdline_args_copy.append_client_arguments("-o read")
+
     # wrs stands for work requests
     server_read_wrs_before_test = efa_retrieve_hw_counter_value(cmdline_args.server_id, "rdma_read_wrs")
     if server_read_wrs_before_test is None:


wenduwan:
bot:aws:retest

shijin-aws:
I am pretty sure the goal of this test was to verifying the protocol selection in operations that have non-read protocols (like eager). @a-szegel can correct me if wrong.

However, fi_rma_bw -o read will only use rdma read when it is available. This change will make this test useless..

before p5, fi_rma_bw -o write will use long read protocol for message size > the min read msg size when rdma-read is available. On p5, it will use rdma write directly.

I think to make the test still meaningful, we would need to switch the test from fi_rma_bw to something still have non-read protocols, e.g. fi_rdm_tagged_bw or fi_rdm_pingpong

wenduwan:
It seems there are multiple concerns about the expected behavior around `-o` value.

I think overall what we want to test is:
- If `-o read`, then read should be used if the device supports RDMA read. This PR does that.
- If `-o write` then write should be used if the device supports RDMA write; if the device only supports read then it should fall back to the above. Current we don't have test coverage for that.

This test requires knowledge of EFA device capability. I think we need a C program script that calls rdma core write to determine that.

shijin-aws:
Why not change fi_rma_bw to a send benchmark? Then it will work for both p4 and p5

shijin-aws:
the test is created to verify read protocol is being used, why do you want to use it for validating write?

I am proposing using a send benchmark because it will use read protocols on compatible platforms

wenduwan:
I'm confused. In this PR  added `-o read` to force the test to use READ instead of the default WRITE. Isn't this what you are also suggesting?

shijin-aws:
Using `-o read` will test fi_read, in which efa only use rdma-read. That is not the original purpose of this test - It tries to examine whether toggling `FI_EFA_MIN_READ_SIZE` can make fi_send or fi_write use read-based protocols as expected. Using `-o read` breaks this goal because efa has no other protocols to choose.

wenduwan:
Ah I think you meant `FI_EFA_INTER_MIN_READ_MESSAGE_SIZE` variable. 
Note that this test is parametrized over
```python
@pytest.mark.parametrize("fabtest_name,cntrl_env_var", [("fi_rdm_tagged_bw", "FI_EFA_INTER_MIN_READ_MESSAGE_SIZE"), ("fi_rma_bw", "FI_EFA_INTER_MIN_READ_WRITE_SIZE")])
```

This change only affects `FI_EFA_INTER_MIN_READ_WRITE_SIZE` which only affects write.
> The mimimum message size for inter EFA write to use read write protocol.
If firmware support RDMA read, and FI_EFA_USE_DEVICE_RDMA is 1, write
requests whose size is larger than this value will use the read write
protocol (Default 65536).

But reading how it is being used on main branch, I think the document is inaccurate:
The switch is only used on [this line](https://github.com/ofiwg/libfabric/blob/7cdce3767db2aca7147e8c9f607b24fa6f979f53/prov/efa/src/rdm/efa_rdm_rma.c#L396). But before that, we will [directly post rdma write](https://github.com/ofiwg/libfabric/blob/7cdce3767db2aca7147e8c9f607b24fa6f979f53/prov/efa/src/rdm/efa_rdm_rma.c#L358) if it's supported regardless of the message size.

We should either update the doc or change the code(my preference).

shijin-aws:
Ah, you made a good point, that environment variable kinds of conflicted with the rdma-write protocol. I think rdma write should still have the highest priority when it is available. 

So... we should either modify the doc to indicate that this env only works when disabling rdma write, or have a way to make `FI_EFA_INTER_MIN_READ_WRITE_SIZE` unreasonably high by default when rdma-write is available. And some refactor of the protocol selection is needed in efa_rdm_rma_writemsg function

shijin-aws:
But .... take a step back. Is this env really being used?

```
(venv) (venv) [ec2-user@ip-172-31-18-57 efa]$ git grep inter_min_read_write_size
src/efa_env.c:  fi_param_define(&efa_prov, "inter_min_read_write_size", FI_PARAM_INT,
src/efa_hmem.c:         fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
src/efa_hmem.c:         fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
src/efa_hmem.c:         fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
src/efa_hmem.c:         -FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &tmp_value) ||
```

PR #9377, opened 2023-09-27
prov/util: Switch ep_list_lock to always be mutex

a-szegel:
Switch ep_list_lock to always be a mutex lock until we can agree on a better long term decision. The issue is that some endpoints want the ability to not require FI_THREAD_SAFE at the domain level, but they want to be able to control the endpoint in a seperate thread.  The ep_list_lock is used for both the data path, and the control path.  In order to adhere to the current api, and not break providers that follow the current api, we need to revert this lock to a mutex until we can get a optimal solution for everyone.

### DIFF
--- a/prov/util/src/util_av.c
+++ b/prov/util/src/util_av.c
@@ -552,9 +552,7 @@ int ofi_av_init_lightweight(struct util_domain *domain, const struct fi_av_attr 
 	 */
 	av->context = context;
 	av->domain = domain;
-	ofi_genlock_init(&av->ep_list_lock,
-			 domain->threading == FI_THREAD_DOMAIN ?
-			       OFI_LOCK_NOOP : OFI_LOCK_MUTEX);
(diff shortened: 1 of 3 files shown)


a-szegel:
I do not think this is the right long term solution, but it gives us time to come up with a better option (without fully reverting).

shefty:
I agree that we don't want to revert completely.  We will also need to update the other places where this change was made.  Counter and EQ?  (I don't remember).

FWIW, I think apps using msg ep's may be more impacted by the threading options than rdm ep's.  That's where my concern is coming from.

a-szegel:
Done.  I don't want to trade use cases.  I want to get the optimal approach for both use cases.

shefty (approved):
thanks!

shefty:
I agree.  However I think doing so may require changing the behavior outlined in the man pages.  That's doable for version 2, but may be difficult in v1.

PR #9378, opened 2023-09-27
prov/verbs: Add support for IBV_ACCESS_RELAXED_ORDERING

sydidelot:
IBV_ACCESS_RELAXED_ORDERING allows the system to reorder
Send/Write/Atomic operations to improve performance.

The patch enables IBV_ACCESS_RELAXED_ORDERING if the application
has requested no ordering in TX/RX attributes.

### DIFF
--- a/prov/verbs/src/verbs_info.c
+++ b/prov/verbs/src/verbs_info.c
@@ -113,6 +113,15 @@ const struct fi_rx_attr verbs_rx_attr = {
 	.total_buffered_recv	= 0,
 };
 
+const struct fi_rx_attr verbs_ro_rx_attr = {
+	.caps			= VERBS_MSG_RX_CAPS,
+	.mode			= VERBS_RX_MODE,
+	.op_flags		= FI_COMPLETION,
+	.msg_order		= 0,
+	.comp_order		= FI_ORDER_STRICT,
+	.total_buffered_recv	= 0,
+};
+
 const struct fi_rx_attr verbs_dgram_rx_attr = {
 	.caps			= VERBS_DGRAM_RX_CAPS,
 	.mode			= VERBS_DGRAM_RX_MODE | VERBS_RX_MODE,
@@ -132,6 +141,16 @@ const struct fi_tx_attr verbs_tx_attr = {
 	.rma_iov_limit		= 1,
 };
 
+const struct fi_tx_attr verbs_ro_tx_attr = {
+	.caps			= VERBS_MSG_TX_CAPS,
+	.mode			= 0,
+	.op_flags		= VERBS_TX_OP_FLAGS,
+	.msg_order		= 0,
+	.comp_order		= FI_ORDER_STRICT,
+	.inject_size		= 0,
+	.rma_iov_limit		= 1,
+};
+
 const struct fi_tx_attr verbs_dgram_tx_attr = {
 	.caps			= VERBS_DGRAM_TX_CAPS,
 	.mode			= 0,
@@ -146,18 +165,28 @@ const struct verbs_ep_domain verbs_msg_domain = {
 	.suffix			= "",
 	.type			= FI_EP_MSG,
 	.protocol		= FI_PROTO_UNSPEC,
+	.relaxed_ordering	= false,
+};
+
+const struct verbs_ep_domain verbs_msg_ro_domain = {
+	.suffix			= "-ro",
+	.type			= FI_EP_MSG,
+	.protocol		= FI_PROTO_UNSPEC,
+	.relaxed_ordering	= true,
 };
 
 const struct verbs_ep_domain verbs_msg_xrc_domain = {
 	.suffix			= "-xrc",
 	.type			= FI_EP_MSG,
 	.protocol		= FI_PROTO_RDMA_CM_IB_XRC,
+	.relaxed_ordering	= false,
 };
 
 const struct verbs_ep_domain verbs_dgram_domain = {
 	.suffix			= "-dgram",
 	.type			= FI_EP_DGRAM,
 	.protocol		= FI_PROTO_UNSPEC,
+	.relaxed_ordering	= false,
 };
 
 /* The list (not thread safe) is populated once when the provider is initialized */
@@ -770,8 +799,13 @@ static int vrb_alloc_info(struct ibv_context *ctx, struct fi_info **info,
 	switch (ep_dom->type) {
 	case FI_EP_MSG:
 		fi->caps = VERBS_MSG_CAPS;
-		*(fi->tx_attr) = verbs_tx_attr;
-		*(fi->rx_attr) = verbs_rx_attr;
+		if (ep_dom->relaxed_ordering) {
+			*(fi->tx_attr) = verbs_ro_tx_attr;
+			*(fi->rx_attr) = verbs_ro_rx_attr;
+		} else {
+			*(fi->tx_attr) = verbs_tx_attr;
+			*(fi->rx_attr) = verbs_rx_attr;
+		}
 		fi->addr_format = FI_SOCKADDR_IB;
 		break;
 	case FI_EP_DGRAM:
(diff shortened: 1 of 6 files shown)


man/fi_mr.3.md
```diff
@@ -310,6 +310,10 @@ The following apply to memory registration.
   requires registering regions passed to collective calls using the
   FI_COLLECTIVE flag.
 
+*FI_MR_RELAXED_ORDERING*
+: This mode bit indicates that the provider is allowed to reorder accesses
+  to the MR to improve performance.
```
shefty, line 315: How is this visible to the app?  I checked the libibverbs and other found documentation, and none of it adds any clarity.

For the other FI_MR_xxx bits, when those are set, there are specific steps that an application must take to support the provider implementation.  What specific step(s) must an app take if this bit is set?

For example, libfabric doesn't define data ordering within a message and has separate message ordering semantics.  Why couldn't the verbs provider just enable this if available?

sydidelot: I have found a couple of documents on the Internet where enabling RELAXED_ORDERING was suggested to improve performance, especially on AMD systems: 
- https://www.amd.com/content/dam/amd/en/documents/epyc-business-docs/performance-briefs/epyc-7002-series-processors-broadcom.pdf
- https://hpcadvisorycouncil.atlassian.net/wiki/spaces/HPCWORKS/pages/1280442391/AMD+2nd+Gen+EPYC+CPU+Tuning+Guide+for+InfiniBand+HPC?preview=%2F1280442391%2F1280409615%2Fimage-20200206-071308.png

We could enable it by default as you suggested, but I'm concerned about breaking applications that require a strict ordering of RDMA operations. 

As a reference, relaxed ordering in UCX is enabled by default if an AMD processor is detected. This feature can be disabled by the environment variable `UCX_IB_PCI_RELAXED_ORDERING=no`. Maybe we should take the same approach?

shefty: Unless there's something specific that an app needs to do in regard to setting this, I would keep it internal.  I would assume we could use the current ordering semantics to know if the app requires RDMA ordering (e.g. FI_ORDER_RMA_WAW)

include/ofi_util.h
```diff
@@ -209,6 +209,8 @@ struct util_domain {
 	uint64_t		info_domain_caps;
 	uint64_t		info_domain_mode;
 	int			mr_mode;
+	uint64_t		tx_msg_order;
+	uint64_t		rx_msg_order;
```
shefty, line 213: This ordering can change per endpoint.  We'll need to figure out how to deal with this, or force all endpoints to use the same ordering.

It's highly likely all endpoints will have the same ordering.  But that's not guaranteed, especially since this is updating the util code.

prov/verbs/src/verbs_mr.c
```diff
@@ -201,6 +201,11 @@ vrb_mr_ofi2ibv_access(uint64_t ofi_access, struct vrb_domain *domain)
 			      IBV_ACCESS_REMOTE_WRITE |
 			      IBV_ACCESS_REMOTE_ATOMIC;
 
+	/* No strict ordering requested */
+	if (domain->util_domain.tx_msg_order == FI_ORDER_NONE &&
+	    domain->util_domain.rx_msg_order == FI_ORDER_NONE)
```
shefty, line 206: FI_ORDER_NONE is 0, which isn't particularly useful as a flag definition.  Can you change these checks to !tx_msg_order && !rx_msg_order?  (I remove FI_ORDER_NONE from the header files in v2 because it's a flag of 0.)

Of course, this change depends on how we handle the comments from the first patch.

prov/util/src/util_domain.c
```diff
@@ -112,6 +112,8 @@ util_domain_init(struct util_domain *domain, const struct fi_info *info,
 	domain->info_domain_caps = info->caps | info->domain_attr->caps;
 	domain->info_domain_mode = info->mode | info->domain_attr->mode;
 	domain->mr_mode = info->domain_attr->mr_mode;
+	domain->tx_msg_order = info->tx_attr->msg_order;
+	domain->rx_msg_order = info->rx_attr->msg_order;
```
shefty, line 116: I was wondering if we could OR in the msg_order values every time we allocate an endpoint.  That at least updates the tx/rx_msg_order fields.  However, it's possible for the app to register memory prior to creating any endpoints.  In that case, we would incorrectly use relaxed ordering.

So, I'm wondering if we have to have a domain that only supports relaxed ordering.  If the app calls fi_info with no ordering set, we can return that domain as one of the options.  Endpoints allocated on that domain are only unordered.

This brings up that I think most apps need ordering for sends.  AFAICT, the relax ordering here mostly impacts RMA ordering: WAW, RAR, etc.  We may only need to check that those bits are not set.

sydidelot:
@shefty I have updated the PR based on your suggestion to create one domain for relaxed ordering.
The PR only supports msg endpoints for now, but I believe the feature could easily be extended for XRC and DGRAM domains.
I verified the code with a modified version of `fi_pingpong` that sets `tx_attr->msg_order = rx_attr->msg_order = 0` before calling `fi_getinfo()`.

prov/verbs/src/verbs_info.c
```diff
@@ -113,6 +113,15 @@ const struct fi_rx_attr verbs_rx_attr = {
 	.total_buffered_recv	= 0,
 };
 
+const struct fi_rx_attr verbs_ro_rx_attr = {
+	.caps			= VERBS_MSG_RX_CAPS,
+	.mode			= VERBS_RX_MODE,
+	.op_flags		= FI_COMPLETION,
+	.msg_order		= 0,
+	.comp_order		= FI_ORDER_STRICT | FI_ORDER_DATA,
```
shefty, line 121: I think we need to drop FI_ORDER_DATA, since the data within a message may be placed into memory OOO

shefty:
CI failure is related.  Changes break verbs;ofi_rxd.  Most or all fabtests fail, but here's a simple one:
```
fi_getinfo_test -s cb14-ib0 cb14-ib0 -p "verbs;ofi_rxd"
13:41:03    timestamp: 20231010-203537+0000
13:41:03    result: Fail
```
Quickly scanning the changes didn't point anything to me.  AFAICT, dgram endpoints shouldn't have been impacted by this change.

iziemba:
Question: Is the correct behavior to zero the message order flags? Or, is the correct behavior to set max_order_waw_size to zero?

shefty:
IMO, the feature isn't well enough documented at the verbs layer to figure out what exactly it does.  The PR comment indicates that sends could be received out of order.  Is that true, or would sends still match receive buffers in order?  Hmm... the latter seems reasonable, since matching would be done by the NIC.  That implies that WAW ordering, and RAR I guess, is not guaranteed.

thomasgillis:
Following a recent issue I have with PCIe ordering (https://github.com/ofiwg/libfabric/issues/9621), I am worried that the possible reordering of the PCIe writes at the target will break the completion ordering of `RDMA_WRITE_WITH_IMM_DATA`.
I couldn't find many details on `IBV_ACCESS_RELAXED_ORDERING`, is there any guarantee that it will not break that?

shefty:
I haven't found any information on what IBV_ACCESS_RELAXED_ORDERING actually does.  If completions can be written before the message data is received, then it's use seems suspect.  I will try to find out more details.

shefty:
Based on initial feedback, IBV_ACCESS_RELAXED_ORDERING should not impact completions.  After reading a completion, all data should be present.  I'm still investigating what it does beyond that and if it impacts message ordering.

shefty:
See https://github.com/linux-rdma/rdma-core/pull/1413 for a proposed update to better document the change in verbs API behavior.  My understanding is that @iziemba 's earlier proposal to disable waw ordering is correct.

sydidelot:
@shefty Thanks for the feedback and for the clarification of `IBV_ACCESS_RELAXED_ORDERING`.
Should we enable relaxed ordering if the user provides `hints->ep_attr->max_order_waw_size == 0` as suggested by @iziemba earlier in the discussion?

shefty:
I honestly still don't know what that flag does.  It sets the relaxed ordering bit for PCI transfers, but it's unclear when that bit is set. 
 I submitted text for it to libibverbs, which was merged, but I don't know if that text was actually correct.  WAW ordering may actually be maintained, so the situation may be better than it seems.  I'm still trying to determine this.

The flag applies at the receiving side.  *If* WAW ordering is impacted, I would also check/update the msg_order flags.

PR #9379, opened 2023-09-27
prov/efa: RDMA write with immediate data completion bugfix

wenduwan:
This change contains 2 fixes:
1. Payload length should be written to completion.
2. Immediate data and payload length are 32-bit integers from rdma-core and should be converted to 64-bit as per libfabric API.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_progress.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_progress.c
@@ -290,11 +290,13 @@ void efa_rdm_ep_check_peer_backoff_timer(struct efa_rdm_ep *ep)
  * connid available.
  *
  * @param[in,out]	ep		endpoint
- * @param[in]		int32_t		Data provided in the IMMEDIATE value.
+ * @param[in]		imm_data	Data provided in the IMMEDIATE value.
+ * @param[in]		len		Payload	length
  * @param[in]		flags		flags (such as FI_REMOTE_CQ_DATA)
  */
 void efa_rdm_ep_proc_ibv_recv_rdma_with_imm_completion(struct efa_rdm_ep *ep,
-						       int32_t imm_data,
+						       uint32_t imm_data,
+						       uint32_t len,
 						       uint64_t flags,
 						       struct efa_rdm_pke *pkt_entry)
 {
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_ep_progress.c
```diff
@@ -306,14 +309,18 @@ void efa_rdm_ep_proc_ibv_recv_rdma_with_imm_completion(struct efa_rdm_ep *ep,
 	target_cq = ep->base_ep.util_ep.rx_cq;
 	efa_av = ep->base_ep.av;
 
+	/* Convert 32-bit integer to 64-bit by clearing out MSB's */
+	imm_data &= mask;
+	len &= mask;
+
```
lrbison, line 315: these lines do nothing.  These are already 32-bit values.  You are sign-extending them, then masking, then storing back the original values.

lrbison:
Hm, My understanding was that EFA doesn't implement write with payload.  I thought IMMEDIATE didn't count towards payload.

Also, I think you haven't changed any code related to 64-bit APIs.  I recall specifically changing the immediate size that EFA provider supported when implementing the RDMA write to 32-bit.  You are probably correct that the API should stay 64-bit, but I think all the public functions already use 64-bit values.

prov/efa/src/rdm/efa_rdm_ep_progress.c
wenduwan: Yes. I realized that this is undefined behavior. Please see update.

wenduwan:
> EFA doesn't implement write with payload. I thought IMMEDIATE didn't count towards payload.

I verified that the actual write payload length is available(and correct) from `ibv_wc_read_byte_len`. I need to confirm with the rdma-core team.

prov/efa/src/rdm/efa_rdm_ep_progress.c
lrbison: This is still not necessary, C will sign-extend these values when they are stored as arguments into the function.  The only uncertainty is if the leading bits are 0s or 1s.  Your mask here forces them to be 0s, which may not be desired if imm_data is actually interpreted as a negative (signed) value.

wenduwan: This was revealed by fabtests `test_cq_data`.

The leading bits were not cleared automatically
```
error, Expected data:0x89abcdef, Received data:0xffffffff89abcdef
```

wenduwan: The issue is data type mismatch between rdma-core and libfabric

rdma-core returns `uint32_t` but libfabric wants `uint64_t`.

prov/efa/src/rdm/efa_rdm_ep_progress.c
```diff
@@ -306,14 +310,18 @@ void efa_rdm_ep_proc_ibv_recv_rdma_with_imm_completion(struct efa_rdm_ep *ep,
 	target_cq = ep->base_ep.util_ep.rx_cq;
 	efa_av = ep->base_ep.av;
 
+	/* Convert 32-bit integer to 64-bit by clearing out MSB's */
```
shijin-aws, line 313: why not just cast?

wenduwan:
@lrbison @shijin-aws We figured that the actual problem was data was passed as signed int. Changing the function signature to unsigned int is sufficient.

Please see update.

shijin-aws (approved):

shijin-aws:
@lrbison do you have more comments?

lrbison:
Thank you, looks good to me.

lrbison (approved):

PR #9380, opened 2023-09-27
prov/verbs_nd: Fetch rejected connection data

k0zmo:
Currently, the verbs provider on Windows always generates an error event for rejected connections without associated error data.


shefty (approved):

shefty:
Intel CI failure is unrelated.  Restarted appveyor, which hit some internal server error.

PR #9381, opened 2023-09-27
prov/util: Add missing calls to (de)initialize monitor's mutexes

k0zmo:
This is especially needed on Windows where `PTHREAD_MUTEX_INITIALIZER` is not enough and explicit initialization via call to `pthread_mutex_init`/`InitializeCriticalSection` is required before using a mutex.

### DIFF
--- a/prov/util/src/util_mem_monitor.c
+++ b/prov/util/src/util_mem_monitor.c
@@ -167,7 +167,9 @@ void ofi_monitor_cleanup(struct ofi_mem_monitor *monitor)
  */
 void ofi_monitors_init(void)
 {
+	pthread_mutex_init(&mm_lock, NULL);
 	pthread_mutex_init(&mm_state_lock, NULL);
+	pthread_rwlock_init(&mm_list_rwlock, NULL);
 
 	uffd_monitor->init(uffd_monitor);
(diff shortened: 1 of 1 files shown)


shefty (approved):
Does this code work on Windows?  Or are we just going through some common function that ends up doing nothing, but still needs the lock?

shefty:
Intel CI failure is unrelated

k0zmo:
Currently, the provided VS solution builds libfabric without HMEM CUDA support or memhooks monitor (both `default_monitor` and `default_cuda_monitor` are NULL). Because of this, `ofi_mr_cache_init` fails, and no further calls to `ofi_mr_cache_*` are performed.

If you add CUDA support (manually define `HAVE_CUDA`, add `$(CUDA_PATH)\include` to the include directory, and link to `$(CUDA_PATH)\lib\x64\{cuda,cudart,nvml}.lib`), then `ofi_mr_cache_init` succeeds, and attempting to lock `mm_lock` causes a segmentation fault.

After these changes, there are no more segmentation faults in that area.

shefty:
Thanks!  That makes sense.  I wasn't considering the hmem case.

PR #9384, opened 2023-09-29
libfabric version2: Initial set of proposed changes

shefty:
For discussion and review only.  CANNOT BE MERGED!!!


shefty:
Update: adjusting patches to start working through CI issues (many of which are related to building on other platforms and missing updates to the CI scripts)

shefty:
@zachdworkin - earlier Intel CI failed on socket provider testing, which makes sense, since I removed that provider.  Is there some way I can disable sockets from being run in the contrib/intel scripts, or would this require updates to the CI system?  (I'm NOT asking to disable anything in the CI system now -- that wouldn't be until we're closer to accepting these sort of changes).

zachdworkin:
> @zachdworkin - earlier Intel CI failed on socket provider testing, which makes sense, since I removed that provider. Is there some way I can disable sockets from being run in the contrib/intel scripts, or would this require updates to the CI system? (I'm NOT asking to disable anything in the CI system now -- that wouldn't be until we're closer to accepting these sort of changes).

The only way I know that you can do this is with the replay function and either comment out that stage or just delete it from your replay. Otherwise I will need to remove it from the pipeline with a different PR

shefty:
mpich references FI_NAME_MAX, so removing that from the public header breaks that build.  We either need to add it back in and document what it actually means, leave it purely for compatibility, or update mpich to use an internal definition.

man/fi_msg.3.md
```diff
@@ -222,7 +222,7 @@ fi_sendmsg.
 
 *FI_CLAIM*
 : Applies to posted receive operations for endpoints configured
-  for FI_BUFFERED_RECV or FI_VARIABLE_MSG.  This flag is used to
+  for FI_BUFFERED_RECV.  This flag is used to
```
mblocksome, line 225: In addition to the `FI_VARIABLE_MSG` flag, the `FI_BUFFERED_RECV` flag was _also_ removed in this pull request. Therefore, shouldn't the `FI_CLAIM` and `FI_DISCARD` flags _also_ be removed?

man/fi_msg.3.md
```diff
@@ -234,7 +234,7 @@ fi_sendmsg.
 
 *FI_DISCARD*
 : Applies to posted receive operations for endpoints configured
-  for FI_BUFFERED_RECV or FI_VARIABLE_MSG.  This flag is used to
+  for FI_BUFFERED_RECV.  This flag is used to
```
mblocksome, line 237: In addition to the `FI_VARIABLE_MSG` flag, the `FI_BUFFERED_RECV` flag was _also_ removed in this pull request. Therefore, shouldn't the `FI_CLAIM` and `FI_DISCARD` flags _also_ be removed?

man/fi_tagged.3.md
```diff
@@ -319,11 +319,6 @@ The following flags may be used with fi_trecvmsg.
   fi_context structure used for an FI_PEEK + FI_CLAIM operation must be used
```
mblocksome, line 319: remove text about `fi_recv_context` since buffered receives have been removed?

mblocksome (changes requested):

include/rdma/fabric.h
```diff
@@ -333,6 +333,13 @@ enum {
 	FI_PROTO_SM2,
 };
 
+enum {
+	FI_TAG_BITS,
+	FI_TAG_HPC,
+	FI_TAG_AI,
```
shefty, line 339: The above should revert back to FI_TAG_MPI and FI_TAG_NCCL, respectively, as the tag formats specifically target the tag definitions for that middleware.

j-xiong: FI_TAG_CCL to be more implementation neutral.

man/fi_msg.3.md
j-xiong: FI_CLAIM and FI_DISCARD can also be used with FI_PEEK.

shefty: @mblocksome and @j-xiong are both correct.  However, I agree with Mike.  This change is to the fi_msg man page, and FI_PEEK does not apply there.  FI_CLAIM and FI_DISCARD can be removed as flags for the fi_msg calls.  They remain for fi_tagged.

j-xiong:
Replaced by #10009.

PR #9385, opened 2023-09-29
fabtests/cq_data: relax CQ data validation to cq_data_size

wenduwan:
The provider only supports cq_data_size bytes of CQ data, thus the test can only reliably verify the last cq_data_size*8 bits.

### DIFF
--- a/fabtests/functional/cq_data.c
+++ b/fabtests/functional/cq_data.c
@@ -43,6 +43,10 @@ static int run_test()
 	size_t size = 1000;
 	struct fi_cq_data_entry comp = {0};
 	struct fi_rma_iov remote;
+	uint64_t mask = UINT64_MAX;
+
+	if (fi->domain_attr->cq_data_size < sizeof(uint64_t))
+		mask ^= mask << (fi->domain_attr->cq_data_size * 8);
 
 	if (opts.cqdata_op == FT_CQDATA_WRITEDATA) {
 		ret = ft_exchange_keys(&remote);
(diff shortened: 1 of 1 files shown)


fabtests/functional/cq_data.c
```diff
@@ -43,6 +43,7 @@ static int run_test()
 	size_t size = 1000;
 	struct fi_cq_data_entry comp = {0};
 	struct fi_rma_iov remote;
+	uint64_t mask = (0x1ULL << (fi->domain_attr->cq_data_size * 8)) - 1;
```
shefty, line 46: If cq_data_size is 8, this left shifts 1 by 64, which is undefined.

shefty (changes requested):

fabtests/functional/cq_data.c
darrylabbate: For your consideration:

```c
size_t cq_data_bits = fi->domain_attr->cq_data_size * 8;
uint64_t mask = (((uint64_t) (cq_data_bits < 64)) << (cq_data_bits & 63)) - 1;
```

wenduwan: Thanks both. Please see the update.

fabtests/functional/cq_data.c
```diff
@@ -43,6 +43,10 @@ static int run_test()
 	size_t size = 1000;
 	struct fi_cq_data_entry comp = {0};
 	struct fi_rma_iov remote;
+	uint64_t mask = -1;
```
shefty, line 46: Please use UINT64_MAX instead of assigning a signed value to an unsigned variable.

wenduwan: Done. Thank you!

shefty:
thanks!

PR #9387, opened 2023-10-01
core: Optimize tag APIs for CCL and MPI

shefty:
Introduces 2 new features to align the API with the *needs* of MPI and NCCL, allowing optimizing for both use cases.

Add new peer group definition.  This allows the app to specify what group of peers a given data transfer belongs to.  This provides a direct mapping for MPI and NCCL communicators.  Peer groups give the provider semantic knowledge about the communication, which cannot be extracted from generic tag bits.

Extend the mem_tag_format definition to allow optimized use of the FI_TAGGED APIs.  The new formats technically break the API, so are marked for libfabric 2.0, but in practice should have no impact on existing applications.  The proposed formats enable a provider to optimize their implementations specifically for either MPI and/or NCCL, or continue to use the current more generic tagged APIs.

### DIFF
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -348,6 +348,13 @@ enum {
 };
 
 enum {
+	FI_TAG_BITS,
+	FI_TAG_MPI,
+	FI_TAG_CCL,
+	FI_TAG_MAX_FORMAT = (1ULL << 16),
+}
+
+enum {
 	FI_TC_UNSPEC = 0,
 	FI_TC_DSCP = 0x100,
 	FI_TC_LABEL = 0x200,
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -449,6 +449,7 @@ struct fi_domain_attr {
 	size_t			max_err_data;
 	size_t			mr_cnt;
 	uint32_t		tclass;
+	uint32_t		max_group_id;
 };
 
 struct fi_fabric_attr {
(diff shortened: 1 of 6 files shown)


man/fi_av.3.md
```diff
@@ -499,6 +501,33 @@ that the output from the AV insertion call is unchanged.  The provider will
 return an fi_addr_t value that maps to each address, and that value must be
 used for all data transfer operations.
 
+# PEER GROUPS
+
+Peer groups provide a direct mapping to HPC and AI communicator constructs.
```
shijin-aws, line 506: Does peer group have some overlap with av set, which is also used to identify the peers in a collective communications... I am wondering whether it could finally replace av set.

shefty: The concepts are similar, but I don't see them merging.  AV sets have a heavy weight setup costs.  They are there as part of setting up collectives, so there's some assumption that creating an AV set involves communication with some fabric management agent to allocate and reserved network resources.  Peer groups currently don't have any setup defined.  Apps simply set some bits in the fi_addr_t which conceptually map to a peer group.  Because these are just bits in the address, they really could be used for anything that the app wants.  The requirement on the provider is that receive buffers must have similar bits set.

If we add a peer group creation call, that moves them closer to an AV set.  But it depends if the create call is synchronous or asynchronous.  It also depends if the peers who are part of the peer group must be identified.  That's true for AV sets, but not peer groups.  If the provider is implementing collective APIs, it needs that information.

include/rdma/fabric.h
```diff
@@ -449,6 +456,7 @@ struct fi_domain_attr {
 	size_t			max_err_data;
 	size_t			mr_cnt;
 	uint32_t		tclass;
+	uint32_t		max_group_id;
```
shefty, line 459: This change breaks ABI compatibility and requires an updated abi_compat struct.  We can limit the break by coordinating these changes with #9319, which also updates the structure, in order to avoid breaking the ABI twice.

shefty:
Commits were added to PR #9384 for review there.  I have a separate set of slides to discuss these specific changes, which are still a work in progress.

PR #9390, opened 2023-10-03
prov/xnet: Reset ep->rx_avail to 0 after RX queue is flushed

sydidelot:
We are hitting an issue where the following assertion fails in
xnet_queue_recv(): assert(!dlist_empty(&ep->unexp_entry));

In GDB, the endpoint state is XNET_DISCONNECTED, but ep->cur_rx
seems to indicate that the endpoint has received a header:
```
(gdb) print ep->cur_rx
[...]
  hdr_len = 24,
  hdr_done = 24,
  data_left = 528,
  entry = 0x0,
  handler = 0xfffff26bba00 <xnet_op_msg>,
  claim_ctx = 0x0
```

ep->cur_rx is supposed to be reset by xnet_reset_rx() when the
endpoint is marked as disconnected in xnet_ep_{disable,close}().
The GDB output above clearly shows the evidence that xnet_progress_rx()
was called after the endpoint was disconnected, which is not supposed
to happen.

The only function that may call xnet_progress_rx() after the
endpoint is disconnected is xnet_queue_recv(), when ep->rx_avail
is > 0.

The problem is that ep->rx_avail is never reset to 0 when the RX
queue is flushed, so the provider wrongly assumes that RX buffers
are available.

### DIFF
--- a/prov/tcp/src/xnet_ep.c
+++ b/prov/tcp/src/xnet_ep.c
@@ -418,6 +418,7 @@ static void xnet_ep_flush_all_queues(struct xnet_ep *ep)
 	}
 	xnet_reset_rx(ep);
 	xnet_flush_xfer_queue(progress, &ep->rx_queue, NULL);
+	ep->rx_avail = 0;
 	ofi_bsock_discard(&ep->bsock);
 }
 
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -1105,6 +1105,7 @@ void xnet_progress_rx(struct xnet_ep *ep)
 
 	assert(xnet_progress_locked(xnet_ep2_progress(ep)));
 	do {
+		assert(ep->state == XNET_CONNECTED);
 		if (ep->cur_rx.hdr_done < ep->cur_rx.hdr_len) {
 			ret = xnet_recv_hdr(ep);
 		} else {


sydidelot:
@shefty I'm planning to cherry-pick the patch to 1.8x and 1.9x after this PR merges.

shefty:
These changes look correct to me, make sense, and I think we want.  But there's still a gap somewhere.  xnet_queue_recv() only calls xnet_progress_rx() if there's also an unexpected message pending.  Your debug output shows that's the case.

I see how xnet_queue_recv() posts receives to the ep without your patch.  rx_avail is acting somewhat as a proxy for the ep state.  But assuming xnet_ep_disable() was called, the cur_rx was cleared.  So somewhere else tried to progress the rx, picked up the posted buffer, and found some data in the process.

Where did this other progress call come from?  Where did the data come from?  I'm going to assume the data was queued on the socket when it was disconnected.  If that was the case, then the progress call picking up the newly posted rx and marking it as unexpected makes sense.  But that still leaves from where progress was invoked?

(It's entirely possible that by not posting the rx buffer, the progress call doesn't go anywhere, so things just work.  But I wonder if wherever that call is invoked from should have checked the ep state.)

sydidelot:
Thanks for the review, Sean. I understand your concern. Unfortunately, the bug has only been seen once and I'm not able to reproduce it.

To limit the scope of the research, the bug happened without io-uring and without Shared-Receive Queues. The remaining places where `xnet_progress_rx()` is used are: 
- `xnet_uring_run_ep()`, `xnet_run_ep()` and `xnet_progress_unexp()`, which are all executed only if `ep->state == XNET_CONNECTED`.
- `xnet_queue_recv()` where this is no check for ep state.

I also suspected a bug in `xnet_progress_rx()` where the EP is disabled by a function from the `do {} while()` block and we don't exit the loop but I could not find a path where this could happen. Maybe we should move the `assert(ep->state == XNET_CONNECTED)` right at the beginning of the loop, so we can catch this scenario. 

As I said at the beginning of my comment, the bug cannot be reproduced and my only debug data is a core dump. Our best chance to understand the issue is to add the assertion failure in `xnet_progress_rx()` and hope it reproduces :crossed_fingers:

shefty:
That's a good idea.  Can you move the assert into the loop?  We might be able to hit that even with your other change added.

And, yes, it's helpful to know that iouring and shared receives weren't enabled.  I'll keep looking at the code to see if I can identify how a receive can be processed after the ep has been disabled.

sydidelot:
I moved the assert() into the loop :+1:

sydidelot:
@shefty The new assertion failure triggered on a test:
```
#10 0x0000fffff15e3fd0 in __assert_fail_base (fmt=0xfffff16fd3c8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0xfffff5820860 "ep->state == XNET_CONNECTED", 
    file=file@entry=0xfffff581fda0 "prov/tcp/src/xnet_progress.c", line=line@entry=985, function=function@entry=0xfffff5821d80 <__PRETTY_FUNCTION__.30> "xnet_progress_rx") at ./assert/assert.c:92
#11 0x0000fffff15e4040 in __GI___assert_fail (assertion=0xfffff5820860 "ep->state == XNET_CONNECTED", file=0xfffff581fda0 "prov/tcp/src/xnet_progress.c", line=985, 
    function=0xfffff5821d80 <__PRETTY_FUNCTION__.30> "xnet_progress_rx") at ./assert/assert.c:101
#12 0x0000fffff49875fc in xnet_progress_rx (ep=0xffffe5387200) at prov/tcp/src/xnet_progress.c:985
#13 0x0000fffff498943c in xnet_run_ep (ep=0xffffe5387200, pin=true, pout=false, perr=true) at prov/tcp/src/xnet_progress.c:1307
#14 0x0000fffff49898cc in xnet_handle_events (progress=0xffffb7e25b80, events=0xffffb7e25f60, nfds=4, clear_signal=false) at prov/tcp/src/xnet_progress.c:1352
#15 0x0000fffff4989e98 in xnet_run_progress (progress=0xffffb7e25b80, clear_signal=false) at prov/tcp/src/xnet_progress.c:1409
#16 0x0000fffff4978864 in xnet_cq_progress (util_cq=0xffffe57055c0) at prov/tcp/src/xnet_cq.c:84
```

ep state is set to `XNET_DISCONNECTED` when `xnet_progress_rx()` is executed because `xnet_run_ep()` is called with both `pin == true` and `per == true` (see GDB frame 13 in above):
```
static void xnet_run_ep(struct xnet_ep *ep, bool pin, bool pout, bool perr)
{                                                                          
    switch (ep->state) {                                                   
    case XNET_CONNECTED:                                                   
        if (perr)                                                          
            xnet_progress_async(ep);                                       
        if (pin)                                                           
            xnet_progress_rx(ep);                                          
        if (pout)                                                          
            xnet_progress_tx(ep);                                          
        break;                                                             
```

As a consequence, `xnet_progress_async()` is executed (which disables the endpoint) and `xnet_progress_rx()` gets immediately executed after that. Also note that the application does not use zero-copy.
I guess we need to check the endpoint state before each of the 3 functions `xnet_progress_async()`,` xnet_progress_rx()` and `xnet_progress_tx()`.

On a side-note, the assertion failure when zero-copy is enabled (https://github.com/ofiwg/libfabric/issues/7887) is still impacting us from time to time and I wonder if it wouldn't be a side-effect of `xnet_progress_rx()` being called when the EP is disconnected.

shefty:
@sydidelot - can you copy your information above into an issue?  I'm glad the asserts made it easy to find where things were going off.  I'll look at the checks tomorrow, but I agree that we'll need to check in function.  Checking within each call may be safer, but I want to see how that might change xnet_run_ep() and other progress entry points.

PR #9396, opened 2023-10-03
MR Cache Updates

iziemba:
MR cache updates to make ofi_mr_cache_find() more usable. The benefit of this is providers can search the MR cache without knowing the HMEM iface.

Moving MR cache max count and size to live in the MR cache struct lead to small message rate bump.


prov/util/src/util_mr_cache.c
```diff
@@ -393,12 +393,21 @@ struct ofi_mr_entry *ofi_mr_cache_find(struct ofi_mr_cache *cache,
...
 	FI_DBG(cache->prov, FI_LOG_MR, "find %p (len: %zu)\n",
 	       attr->mr_iov->iov_base, attr->mr_iov->iov_len);
 
 	pthread_mutex_lock(&mm_lock);
+	flush_lru = ofi_mr_cache_full(cache);
+	if (flush_lru || !dlist_empty(&cache->dead_region_list)) {
+		pthread_mutex_unlock(&mm_lock);
+		ofi_mr_cache_flush(cache, flush_lru);
+		pthread_mutex_lock(&mm_lock);
+	}
```
shefty, line 409: I'm not following the reasoning for this change.  find() doesn't insert new entries into the cache, so why the need to flush?  It would have the impact of slowing down the find().  That may be okay or desirable, but I'm not sure.

shefty (commented):
no objection, just a question

prov/util/src/util_mr_cache.c
iziemba: That's a fair point. Maybe this change to the following to having some cleanup of evicted entries?

```
if (!dlist_empty(&cache->dead_region_list)) {
    ofi_mr_cache_flush(cache, flse);
    pthread_mutex_lock(&mm_lock);
}
```

shefty: I'm okay with flushing dead entries, but can you move this part of the change to its own commit?  If it shows some problem with performance, it'll be easier to revert if needed.

iziemba: Can do.

shefty (approved):
thanks!

shijin-aws:
how is `ofi_mr_cache_find()` triggered today? I couldn't see it is called by other functions inside libfabric

PR #9397, opened 2023-10-03
core: Implement memhooks atfork child handler

iziemba:
If the parent process forks, memhooks could be installed in the child process. If the fork occurred with the MR cache locked and the child memhooks intercepts fire, the child process would deadlock.

Add a pthread_atfork() handler to clear memhooks in the child process.

Note: RDMA is NOT supported in the child process.

### DIFF
--- a/include/windows/osd.h
+++ b/include/windows/osd.h
@@ -684,6 +684,12 @@ static inline char* strsep(char **stringp, const char *delim)
 
 #define __attribute__(x)
 
+static inline int pthread_atfork(void (*prepare)(void), void (*parent)(void),
+				 void (*child)(void))
+{
+	return FI_ENOSYS;
+}
+
 static inline int ofi_mmap_anon_pages(void **memptr, size_t size, int flags)
 {
 	return -FI_ENOSYS;
--- a/src/fabric.c
+++ b/src/fabric.c
@@ -50,6 +50,7 @@
 #include "ofi_prov.h"
 #include "ofi_perf.h"
 #include "ofi_hmem.h"
+#include "ofi_mr.h"
 #include <ofi_shm_p2p.h>
 #include <rdma/fi_ext.h>
 
@@ -916,6 +917,8 @@ void fi_ini(void)
 	ofi_register_provider(HOOK_NOOP_INIT, NULL);
 
 	ofi_register_provider(COLL_INIT, NULL);
+
+	pthread_atfork(NULL, NULL, ofi_memhooks_atfork_handler);
 
 	ofi_init = 1;
 
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -417,4 +417,6 @@ int ofi_mr_cache_reg(struct ofi_mr_cache *cache, const struct fi_mr_attr *attr,
 void ofi_mr_cache_delete(struct ofi_mr_cache *cache, struct ofi_mr_entry *entry);
 
 
+void ofi_memhooks_atfork_handler(void);
+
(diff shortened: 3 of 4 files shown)


src/fabric.c
```diff
@@ -917,6 +923,11 @@ void fi_ini(void)
 
 	ofi_register_provider(COLL_INIT, NULL);
 
+	ret = pthread_atfork(NULL, NULL, ofi_child_atfork_handler);
```
shefty, line 926: This breaks windows build.  Why the indirection, versus calling ofi_memhooks_atfork_handler() directly?

src/fabric.c
```diff
@@ -917,6 +923,11 @@ void fi_ini(void)
...
+	if (ret)
+		FI_WARN(&core_prov, FI_LOG_CORE, "pthread_atfork failed: %d\n",
+			ret);
```
shefty, line 929: Please add braces after if (ret) or merge ret); with line above

shefty (changes requested):
need to fix build

src/fabric.c
iziemba: Will look into build issue. I'll remove the indirection.

src/linux/osd.c
```diff
@@ -71,6 +72,16 @@ ssize_t ofi_get_hugepage_size(void)
 	return val * 1024;
 }
 
+void ofi_fork_handlers(void)
+{
+	int ret;
+
+	ret = pthread_atfork(NULL, NULL, ofi_memhooks_atfork_handler);
```
shefty, line 79: In similar circumstances, we would define pthread_atfork for the various OS's and just have them as no-ops.

iziemba: Will do.

src/fabric.c
```diff
@@ -917,6 +919,10 @@ void fi_ini(void)
 
 	ofi_register_provider(COLL_INIT, NULL);
 
+	ret = ofi_pthread_atfork(NULL, NULL, ofi_memhooks_atfork_handler);
+	if (ret)
+		FI_WARN(&core_prov, FI_LOG_CORE, "ofi_pthread_atfork failed: %d\n", ret);
```
shefty, line 924: Please remove the warning here, otherwise, we'll see it for non-Linux OS's.

include/windows/osd.h
```diff
@@ -958,6 +958,13 @@ static inline ssize_t ofi_get_hugepage_size(void)
 	return -FI_ENOSYS;
 }
 
+static inline int
+ofi_pthread_atfork(void (*prepare)(void), void (*parent)(void),
+		   void (*child)(void))
+{
+	return -FI_ENOSYS;
+}
```
shefty, line 966: If you look at the other pthread calls in this file, we use the pthread name directly, without a rename.  E.g. 'pthread_atfork', not 'ofi_pthread_atfork'.  I would align with that for consistency.  (Basically, we emulate pthread API calls, rather than abstract them.)

On error, return positive ENOSYS, which aligns with the pthread_atfork() call.

shefty (changes requested):

include/windows/osd.h
iziemba: I think windows should only need pthread_atfork() defined.

shefty (approved):
If this builds, the changes look good.  :)  Thanks

iziemba:
I cannot view those build failures. Are they false alarms?

shefty:
CI failure is unrelated -- some udp multinode test failure

PR #9398, opened 2023-10-03
Support ofi_hmem_get_base_addr() with FI_HMEM_SYSTEM

iziemba:
Implementation uses the same approach libibverbs does to expand memory region: https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/memory.c

### DIFF
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -420,9 +420,9 @@ int ofi_hmem_close_handle(enum fi_hmem_iface iface, void *mapped_addr)
 }
 
 int ofi_hmem_get_base_addr(enum fi_hmem_iface iface, const void *addr,
-			   void **base_addr, size_t *base_length)
+			   size_t len, void **base_addr, size_t *base_length)
 {
-	return hmem_ops[iface].get_base_addr(addr, base_addr, base_length);
+	return hmem_ops[iface].get_base_addr(addr, len, base_addr, base_length);
 }
 
 bool ofi_hmem_is_initialized(enum fi_hmem_iface iface)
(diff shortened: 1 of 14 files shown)


src/hmem.c
```diff
@@ -1,4 +1,6 @@
 /*
+ * Copyright (c) 2004, 2005 Topspin Communications.  All rights reserved.
+ * Copyright (c) 2006 Cisco Systems, Inc.  All rights reserved.
```
iziemba, line 3: Since this code comes from https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/memory.c, I think the copyright needs to be updated. Thoughts?

j-xiong:
The first and second commits should be merged, otherwise the first commit would break the build.

j-xiong:
The prototype for `cuda_get_base_addr` is not updated in `ofi_hmem.h`.

src/hmem.c
shefty: Can you locate the patch that added this there and identify the author?

j-xiong: I think this is it: https://github.com/linux-rdma/rdma-core/commit/422642f1cf021b9de1f445f4392c99d3c2fe6624

iziemba:
Looking into windows build issues.

src/hmem.c
shefty: I think yes.  The author of the changes listed ibm as an email.  Topspin didn't exist by the time they were added.

j-xiong:
A note on the Intel CI failures (jenkins and cloudbee): the `mpitestsuite` stage has one timeout failure `pt2pt/isendrecv 10` but the stage is marked green. The error is captured in the `summary` stage though. @zachdworkin could you take a look?

j-xiong (approved):

PR #9399, opened 2023-10-03
Misc util updates

iziemba:
Each commit is an independent change. Let me know if this should be broken out into a separate PR.

### DIFF
--- a/prov/util/src/util_main.c
+++ b/prov/util/src/util_main.c
@@ -165,6 +165,36 @@ static int util_find_domain(struct dlist_entry *item, const void *arg)
 		 ((info->domain_attr->mr_mode & domain->mr_mode) == domain->mr_mode);
 }
 
+static int ofi_dup_auth_keys(const struct fi_info *hints, struct fi_info *info)
+{
+	if (!hints)
+		return FI_SUCCESS;
+
+	if (hints->domain_attr && hints->domain_attr->auth_key) {
+		info->domain_attr->auth_key =
+			mem_dup(hints->domain_attr->auth_key,
+				hints->domain_attr->auth_key_size);
+		if (!info->domain_attr->auth_key)
+			return -FI_ENOMEM;
+
+		info->domain_attr->auth_key_size =
+			hints->domain_attr->auth_key_size;
+	}
+
+	if (hints->ep_attr && hints->ep_attr->auth_key) {
+		info->ep_attr->auth_key =
+			mem_dup(hints->ep_attr->auth_key,
+				hints->ep_attr->auth_key_size);
+		if (!info->ep_attr->auth_key)
+			return -FI_ENOMEM;
+
+		info->ep_attr->auth_key_size =
+			hints->ep_attr->auth_key_size;
+	}
+
+	return FI_SUCCESS;
+}
+
 /*
  * Produces 1 fi_info output for each fi_info entry in the provider's base
  * list (stored with util_prov), subject to the base fi_info meeting the
@@ -287,6 +317,10 @@ int util_getinfo(const struct util_prov *util_prov, uint32_t version,
 					"cannot resolve source address\n");
 			}
 		}
+
+		ret = ofi_dup_auth_keys(hints, *info);
+		if (ret)
+			goto err;
 	}
 
 	*info = saved_info;
--- a/include/ofi_list.h
+++ b/include/ofi_list.h
@@ -103,6 +103,11 @@ static inline void dlist_remove_init(struct dlist_entry *item)
 	dlist_init(item);
 }
 
+#define dlist_first_entry_or_null(head, type, member) ({	\
+	struct dlist_entry *pos = (head)->next;				\
+	pos != (head) ? container_of((pos), type, member) : NULL;	\
+})
+
 #define dlist_pop_front(head, type, container, member)			\
 	do {								\
 		container = container_of((head)->next, type, member);	\
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -654,6 +654,23 @@ int ofi_check_domain_attr(const struct fi_provider *prov, uint32_t api_version,
 		return -FI_ENODATA;
 	}
 
+	if (user_attr->auth_key_size &&
+	    (user_attr->auth_key_size != prov_attr->auth_key_size)) {
+		FI_INFO(prov, FI_LOG_CORE, "Unsupported authentication size\n");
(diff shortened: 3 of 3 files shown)


include/ofi_mem.h
```diff
@@ -382,6 +382,7 @@ struct ofi_bufpool_hdr {
 	} entry;
 	struct ofi_bufpool_region	*region;
 	size_t 				index;
+	bool				allocated;
```
shefty, line 385: I wouldn't necessarily say that bounds checking is missing, since that implies that it's needed.  IMO, it's nice to have.

It looks like you want to update the indexed buffer pool to allow invalid indices and have it verify if the index is valid.  Can this be a debug only assert?

In other places where I had a similar requirement, I added separate calls -- one which did a direct lookup and another which checked the index.  E.g. ofi_idx_at() vs. ofi_idx_lookup().  This is slightly harder, since I don't see another way to determine if the buffer has been removed from the pool or not, versus just verifying the index.

Main issue is that these pools are used in the fast path, and this is adding debug overhead, unless you're considering some other use case.

prov/util/src/util_main.c
```diff
@@ -287,6 +287,30 @@ int util_getinfo(const struct util_prov *util_prov, uint32_t version,
...
+			    hints->ep_attr->auth_key) {
+				(*info)->ep_attr->auth_key =
+					mem_dup(hints->ep_attr->auth_key,
+						hints->ep_attr->auth_key_size);
+				if (!(*info)->ep_attr->auth_key) {
+					ret = -FI_ENOMEM;
+					goto err;
+				}
+			}
+		}
```
shefty, line 313: This function is getting long.  Can we move the new code into its own function to start breaking it down?

include/ofi_mem.h
iziemba: The CXI provider AV implementation, along with other provider AV implementations, use an ibuf pool for fi_addr_t -> provider specific address translation. Typically, the user provided fi_addr_t is handed directly to the AV ofi_bufpool_get_ibuf() (e.g. ofi_av_remove_addr()).

The CXI provider was doing the same thing. We had unit tests failing where a fi_av_lookup() was able to be successfully issued against an fi_addr_t which was just removed. The fix we had was to add checks to bufpool to know what is valid.

prov/util/src/util_main.c
iziemba: Can do

include/ofi_mem.h
shefty: This sounds like a use-after-free bug in the app using an fi_addr_t value that it had just removed.  The same sort of check could be applied to the data added/removed from the pool.  Is there a way to add this check only for the indexed version of the pool and not the generic version?  I can see validating an index in the free build, but otherwise, I think the checks are better as debug checks.

iziemba: Will change to ibuf and debug only.

iziemba:
Are the CI failures valid?

include/ofi_mem.h
```diff
@@ -474,11 +478,21 @@ static inline size_t ofi_buf_index(void *buf)
...
 
-	buf = pool->region_table[(size_t)(index / pool->attr.chunk_cnt)]->
-		mem_region + (index % pool->attr.chunk_cnt) * pool->entry_size;
+	buf = pool->region_table[region_index]->mem_region +
+		(index % pool->attr.chunk_cnt) * pool->entry_size;
+
+#ifndef NDEBUG
+	if (!ofi_buf_hdr(buf)->allocated)
+		return NULL;
+#endif
```
shefty, line 494: Why not just assert on the above 2 checks if they're going to be debug only code?

shefty (commented):
CI failures look real, but it's unclear how they relate to this PR.  Restarting them

include/ofi_mem.h
iziemba: Works for me.

a-szegel:
Seems to have broken lots of tests on the debug build for EFA:

```
--------------------------------- Captured Log ---------------------------------

--------------------------------- Captured Out ---------------------------------

server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.37.11 'timeout 360 /bin/bash --login -c '"'"'FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr9399-debug/install/fabtests/bin/fi_rdm_pingpong -t counter -v -S all -p efa -E=9231'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.47.18 'timeout 360 /bin/bash --login -c '"'"'FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr9399-debug/install/fabtests/bin/fi_rdm_pingpong -t counter -v -S all -p efa -E=9231 172.31.37.11'"'"''
client_stdout:
fi_rdm_pingpong: ./include/ofi_mem.h:488: ofi_bufpool_get_ibuf: Assertion `ofi_buf_hdr(buf)->allocated' failed.
timeout: the monitored command dumped core

client returncode: 255
server_stdout:

server returncode: 124
```

iziemba:
I think "util/buf: Add bounds checks to ibuf get" is exposing real issues with providers. For example, running `./fi_rdm_pingpong -t counter -v -S 32768 -E=9231 -p "tcp;ofi_rxm" 127.0.0.1` also tripped this assertion. Looking at RXM, it is allocating its tx_pool as not index, but then uses ofi_bufpool_get_ibuf() to index references.

I am going to move this commit to a new PR.

iziemba:
> I think "util/buf: Add bounds checks to ibuf get" is exposing real issues with providers. For example, running `./fi_rdm_pingpong -t counter -v -S 32768 -E=9231 -p "tcp;ofi_rxm" 127.0.0.1` also tripped this assertion. Looking at RXM, it is allocating its tx_pool as not index, but then uses ofi_bufpool_get_ibuf() to index references.
> 
> I am going to move this commit to a new PR.

Maybe this behavior is OK. I have update https://github.com/ofiwg/libfabric/pull/9425/commits/7e154eb5de5a64099491820f5e8d68bfdf4eba13 to set the valid bit for both ofi_buf_alloc() and ofi_ibuf_alloc() and things work.

iziemba:
I don't know which commit is triggering the CI failures. I am going to close this PR and open up a new PR for each commit.

PR #9414, opened 2023-10-09
prov/verbs: Bug fix for matching domain name with device name

j-xiong:
The domain name may contain a `-<xxx>` suffix, the original code used `strncmp(domain_name, dev_name, strlen(dev_name))` to check if the names match. However, this scheme does not work when the device name is a substring of the domain name (e.g. `mlx5_11-rdm` vs `mlx5_1`) and would incorrectly return match for such case.

### DIFF
--- a/prov/verbs/src/verbs_info.c
+++ b/prov/verbs/src/verbs_info.c
@@ -1220,6 +1220,24 @@ vrb_info_add_dev_addr(struct fi_info **info, struct verbs_dev_info *dev)
 	return 0;
 }
 
+/* domain name may have a "-<xxx>" suffix */
+static inline int vrb_cmp_domain_and_dev_name(const char *domain_name,
+					      const char *dev_name)
+{
+	size_t cmp_len;
+	char *s = strrchr(domain_name, '-');
+
+	if (s)
+		cmp_len = s - domain_name;
+	else
+		cmp_len = strlen(domain_name);
+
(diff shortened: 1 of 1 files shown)


prov/verbs/src/verbs_info.c
```diff
@@ -1220,6 +1220,24 @@ vrb_info_add_dev_addr(struct fi_info **info, struct verbs_dev_info *dev)
 	return 0;
 }
 
+/* domain name may have a "-<xxx>" suffix */
+static inline int vrb_cmp_domain_and_dev_name(const char *domain_name,
+					      const char *dev_name)
+{
+	size_t cmp_len;
+	char *s = strchr(domain_name, '-');
```
shefty, line 1228: Is there any chance '-' could appear in the device name?  If so, strrchr() may work.

shefty (approved):

prov/verbs/src/verbs_info.c
shefty: Unless we could have 2 suffixes...

j-xiong: Theoretically yes, but practically no.  I can change it to `strrchr()` just in case.

j-xiong: The suffix is controlled by the provider. We only have one.

PR #9415, opened 2023-10-09
prov/shm: Only copy header + msg on unexpected path

a-szegel:
Currently, we always copy the entire command message of 256 bytes on the unexpected path. For inline, we only need the header (sizeof(struct smr_msg_hdr): 64) + size.  For inject, we only need the header.  SAR, IOV, and IPC use the data field in the message, so we continue to copy the entire command message for them (for large message sizes, optimization is less important).

### DIFF
--- a/prov/shm/src/smr_progress.c
+++ b/prov/shm/src/smr_progress.c
@@ -908,7 +908,6 @@ static int smr_alloc_cmd_ctx(struct smr_ep *ep,
 			"Error allocating cmd ctx\n");
 		return -FI_ENOMEM;
 	}
-	memcpy(&cmd_ctx->cmd, cmd, sizeof(*cmd));
 	cmd_ctx->ep = ep;
 
 	rx_entry->size = cmd->msg.hdr.size;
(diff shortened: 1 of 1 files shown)


a-szegel:
This will get rid of a 256 byte copy when we send unexpected 0 byte sync packets.

a-szegel:
via testing by hand, this doesn't seem to have noticeable impacts to osu_alltoallw benchmark... but I still think it is something we should do.

shefty (approved):
Makes sense to me

shefty:
@aingerson - any concerns with this?

prov/shm/src/smr_progress.c
```diff
@@ -917,7 +916,10 @@ static int smr_alloc_cmd_ctx(struct smr_ep *ep,
 		rx_entry->cq_data = cmd->msg.hdr.data;
 	}
 
-	if (cmd->msg.hdr.op_src == smr_src_inject) {
+	if (cmd->msg.hdr.op_src == smr_src_inline) {
+		memcpy(&cmd_ctx->cmd, cmd, sizeof(struct smr_msg_hdr) + cmd->msg.hdr.size);
```
aingerson, line 920: change this and below to `sizeof(cmd->msg.hdr)`

aingerson (commented):
One small comment, otherwise looks good. Thanks!

prov/shm/src/smr_progress.c
a-szegel: done

aingerson (approved):
Thanks!

shefty:
bot:aws:retest

PR #9420, opened 2023-10-09
prov/tcp: Set FI_MULTI_RECV for last completed RX slice

chien-intel:
When FI_MULTI_RECV is enabled for an RX buffer, the last completed slice of the RX buffer should have FI_MULTI_RECV set in the corresponding CQE.

Currently TCP provider assumes the last slice of the RX buffer will get completed last and sets FI_MULTI_RECV flag when the last slice is used.  Completions can get out of order with multiple connections and varied payload sizes using the same RX buffer.  Last slice will not always complete last.

To make sure the last completed RX buffer slice has FI_MULTI_RECV set in the CQE, use a reference counter to track outstanding slices.  Count is initialized to 1 when the RX buffer is first posted and incrememted on every slice allocation.  Upon completion, count is decremented.  When it reaches 0, FI_MULTI_RECV is set for that CQE.

### DIFF
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -399,6 +399,10 @@ static inline void xnet_signal_progress(struct xnet_progress *progress)
 #define XNET_NEED_CTS		BIT(11)
 #define XNET_MULTI_RECV		FI_MULTI_RECV /* BIT(16) */
 
+struct xnet_mrecv {
+	size_t			ref_cnt;
+};
+
 struct xnet_xfer_entry {
 	struct slist_entry	entry;
 	void			*user_buf;
@@ -409,9 +413,10 @@ struct xnet_xfer_entry {
 	struct util_cntr	*cntr;
 	uint64_t		tag_seq_no;
 	uint64_t		tag;
-	struct {
-		uint64_t	ignore;
-		size_t		rts_iov_cnt;
+	union {
+		uint64_t		ignore;
+		size_t			rts_iov_cnt;
+		struct xnet_mrecv	*mrecv;
 	};
 	fi_addr_t		src_addr;
 	uint64_t		cq_flags;
--- a/prov/tcp/src/xnet_cq.c
+++ b/prov/tcp/src/xnet_cq.c
@@ -151,6 +151,14 @@ void xnet_report_success(struct xnet_xfer_entry *xfer_entry)
 	flags = xfer_entry->cq_flags & ~FI_COMPLETION;
 	if (flags & FI_RECV) {
 		len = xnet_msg_len(&xfer_entry->hdr);
+		if (xfer_entry->ctrl_flags & XNET_MULTI_RECV &&
+		    xfer_entry->mrecv) {
+			xfer_entry->mrecv->ref_cnt--;
+			if (!xfer_entry->mrecv->ref_cnt) {
+				flags |= FI_MULTI_RECV;
+				free(xfer_entry->mrecv);
+			}
+		}
 		xnet_get_cq_info(xfer_entry, &flags, &data, &tag);
 	} else if (flags & FI_REMOTE_CQ_DATA) {
 		assert(flags & FI_REMOTE_WRITE);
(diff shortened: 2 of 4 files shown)


prov/tcp/src/xnet_srx.c
```diff
@@ -48,17 +48,28 @@ xnet_match_tag(struct xnet_srx *srx, struct xnet_ep *ep, uint64_t tag);
...
 	struct xnet_xfer_entry *xfer;
+	struct xnet_mrecv *mrecv = NULL;
 	struct xnet_progress *progress;
 
+	if (multi_recv) {
+		mrecv = calloc(1, sizeof(struct xnet_mrecv));
+		if (!mrecv)
+			return NULL;
+		mrecv->ref_cnt = 1;
+	}
```
shefty, line 62: Could we do this allocation the first time we call xnet_alter_mrecv()?  That would leave the recv path unchanged and not impact performance when mrecv is not being used.  Most posted buffers will not have this flag set.

shefty (commented):
The behavior looks correct to me.  I would like to move the new functionality around a little bit to limit the extra overhead of managing the new struct.

prov/tcp/src/xnet.h
```diff
@@ -611,6 +616,7 @@ xnet_alloc_xfer(struct xnet_progress *progress)
 	xfer->ctrl_flags = 0;
 	xfer->context = NULL;
 	xfer->user_buf = NULL;
+	xfer->mrecv = NULL;
```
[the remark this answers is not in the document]
chien-intel: moved this to xnet_alloc_srx_xfer

shefty:
CI failed running mpi over tcp.  Restarted it, but the failure may be related to the changes.  I didn't notice anything off on the updated changes on a first review, so I may be overlooking an issue.  The changes looked good to me.

prov/tcp/src/xnet_cq.c
```diff
@@ -148,9 +148,17 @@ void xnet_report_success(struct xnet_xfer_entry *xfer_entry)
 		return;
 	}
 
+
 	flags = xfer_entry->cq_flags & ~FI_COMPLETION;
 	if (flags & FI_RECV) {
 		len = xnet_msg_len(&xfer_entry->hdr);
+		if (xfer_entry->mrecv) {
```
shefty, line 155: Wait -- this check doesn't work since moving mrecv into the union.  The checks like this may need to be updated to something like : xfer_entry->ctrl_flags & FI_MULTI_RECV

shefty: meant XNET_MULTI_RECV for the ctrl_flag... it's the same value as FI_MULTI_RECV, but I'd use the internal definition instead

chien-intel: yeah, it was in the back of my mind.  v1 guarantees mrecv but not v2.  There are 2 possible cases where mrecv is null, one, obviously is the failed allocation (unlikely) and the second is for the first receive to consume the RX buffer beyond the minimum.  I'm guessing it is the second case that's causing the failure.  I will make a change to accommodate these.

chien-intel:
Looks better MPI_TCP passed

shefty (approved):
Thanks -- I missed catching the earlier problem when I asked you to move mrecv into a union, which broke the code you had.  Code looks reasonable, but I did list a couple cleanups that I think are possible.

prov/tcp/src/xnet_cq.c
```diff
@@ -151,6 +151,17 @@ void xnet_report_success(struct xnet_xfer_entry *xfer_entry)
...
+		if (xfer_entry->ctrl_flags & XNET_MULTI_RECV) {
+			if (xfer_entry->mrecv) {
+				xfer_entry->mrecv->ref_cnt--;
+				if (!xfer_entry->mrecv->ref_cnt) {
+					flags |= FI_MULTI_RECV;
+					free(xfer_entry->mrecv);
+				}
+			} else {
+				flags |= FI_MULTI_RECV;
+			}
```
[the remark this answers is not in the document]
chien-intel: yup.  I will remove this and same for report_error.

shefty (approved):
looks good -- thanks!  Just waiting for CI to catch up, but is passing our tests

PR #9421, opened 2023-10-09
Support optimized memcpy for device memory

iziemba:
Various device memory types support additional device memory registration routines to enable optimized memcpy access. Typically, the optimization is load/store access over PCIe instead of using device DMA engines.

This PR defines an HMEM abstraction to enable all providers access to optimized memcpy. FI_HMEM_SYSTEM, FI_HMEM_ROCR, and FI_HMEM_CUDA are supported.


src/hmem.c
```diff
@@ -97,6 +97,25 @@ static int ofi_hmem_no_dev_reg_copy_from_hmem(uint64_t handle, void *dest,
 	return -FI_ENOSYS;
 }
 
+static int ofi_hmem_system_dev_register(const void *addr, size_t size,
+					uint64_t *handle)
+{
+	*handle = (uint64_t)addr;
```
shefty, line 103: cast should be ```(uintptr_t) addr```

shefty:
@shijin-aws - Would efa be able to adopt these changes easily?  I'm not asking that you do it, just verify that the proposal aligns with the current use case.

shefty:
I'm okay with these changes.  EFA uses gdrcopy, so I'd like to verify with AWS that the changes could work for them.  I'm not asking that efa start using the API, just confirm that it makes sense and would work.  Even then it's not a huge deal, since these are internal APIs which could change.

iziemba:
I would have look into if ROCR supports device-to-device memcpy.

This API proposal is only intended to work between device and host memory. I think applications need to ensure they are passing in valid addresses to this API (just like normal memcpy). Debug checks could be implemented to ensure the addresses are valid if needed.

shefty (approved):

Issue #7844, opened 2022-06-21
prov/tcp: tcpx_progress.c:677: tcpx_recv_hdr: Assertion `ep->cur_rx.hdr.base_hdr.id == ep->rx_id++' failed.


aakefbs:
This came up with libfabric-1.14 as soon as clients mounted IME servers, with libfabric-1.15 it seems to come up more rarely, maybe with clients that have not been updated to 1.15 yet, but I'm not sure yet. 

tcpx_recv_hdr prov/tcp/src/tcpx_progress.c:678:29 (is_stmt)
tcpx_progress_rx prov/tcp/src/tcpx_progress.c:701:10 (is_stmt)
tcpx_progress prov/tcp/src/tcpx_cq.c:108:16 (is_stmt)
tcpx_cq_progress prov/tcp/src/tcpx_cq.c:118:2 (is_stmt)
ofi_cq_readfrom prov/util/src/util_cq.c:233:3 (is_stmt)
ofi_cq_read prov/util/src/util_cq.c:286:1 (is_stmt)

shefty:
This is a debug field in the tcp provider protocol headers to validate that the tcp stream is as expected.  It cannot work if you mix debug with non-debug versions or different versions of libfabric (at least with a version that doesn't have this debug field).  The field re-uses a reserved field.

aakefbs:
I had already noticed that it is a debug field only and I'm already in the process to disable debug on our servers.

But then, the remote system should not touch this field, but just forward it then, whatever is in it? I.e. remote configuration should not matter for local config?

shefty:
This field is a message sequence number.  The receiver checks that the value in the received header is the value that it expects.  So it needs to be set by the sender.  It's simply used as a validation that what the receiver thinks is the header of a new message is likely to be the header of a new message.

aakefbs:
Hmm ok, I see. I didn't check the code and was assuming the debug side would just read its own value created at tx time (this is what we have in IME).
I was going to suggest to use UINT64_MAX as id value when debug is not enabled, but given it is just an 8 bit variable that would't work. What about letting remote side set a tcpx_base_hdr::flags telling if debug is enabled?

aakefbs:
[...]

PR #9424, opened 2023-10-10
prov/xnet: Return an error if invalid sequence number received

sydidelot:
If compiled with --enable-debug, the provider would fail an assertion if the sequence number received in the header doesn't match the one expected.

Crashing on an assertion failure is fine for programming errors, but the assertion failure described above relies on external events: it may for example trigger because of a buffers corrupted by the TCP stack or the physical layer.

Instead of an assertion failure, the function returns an error code, which will cause the EP to be disabled as a consequence.


prov/tcp/src/xnet_progress.c
```diff
@@ -1011,7 +1011,15 @@ static int xnet_progress_hdr(struct xnet_ep *ep)
...
 	ep->hdr_bswap(ep, &ep->cur_rx.hdr.base_hdr);
-	assert(ep->cur_rx.hdr.base_hdr.id == ep->rx_id++);
+
+#ifndef NDEBUG
+	if (ep->cur_rx.hdr.base_hdr.id != ep->rx_id++) {
+		FI_WARN(&xnet_prov, FI_LOG_EP_DATA,
+			"Received invalid hdr sequence number\n");
+		return -FI_EIO;
+	}
+#endif
```
shefty, line 1021: Do we want to make the id an always on check?

sydidelot: The problem with making the id an always on check is that it would change the wire protocol: sender is required to set the id, which never happens if debug is disabled.

bsbernd:
Btw, relates to https://github.com/ofiwg/libfabric/issues/7844

PR #9429, opened 2023-10-11
util: Add EP and domain auth key checking

iziemba:
Verify that user provided auth key hints are sane.

### DIFF
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -654,6 +654,13 @@ int ofi_check_domain_attr(const struct fi_provider *prov, uint32_t api_version,
 		return -FI_ENODATA;
 	}
 
+	if (user_attr->auth_key && user_attr->auth_key_size &&
+	    (user_attr->auth_key_size != prov_attr->auth_key_size)) {
+		FI_INFO(prov, FI_LOG_CORE, "Unsupported authentication size\n");
+		OFI_INFO_CHECK_SIZE(prov, prov_attr, user_attr, auth_key_size);
+		return -FI_ENODATA;
+	}
+
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


shefty:
CI failure:
```
18:36:43  - name:   fi_multi_ep -e rdm -v -p "psm3"
18:36:43    timestamp: 20231011-013047+0000
18:36:43    result: Fail
...
18:36:43      fi_getinfo(): functional/multi_ep.c:278, ret=-61 (No data available)
```
I don't know why, might be a problem in fabtests multi_ep

shefty:
fabtests looks okay -- problem looks like it's in the psm3 provider or the definition of auth_key_size needs clarity.

I see psm3 sets the auth_key_size to a non-zero value, but auth_key = null as a default attribute. The man pages suggest that auth_key_size should be set to 0 on return from fi_getinfo() if auth_keys are not used.

shefty:
We may be able to fix this by modifying fi_alter_domain_attr / fi_alter_ep_attr to set the auth_key_size = 0 if auth_key=null.  The provider is setting the auth_key_size to its max value, which allows the checks against hints to work.  But since auth keys aren't being used, the returned fi_info structures should have those cleared.  I'm not sure if that's sufficient, but it may be.  psm3 keys off the auth_key being null when creating resources.

j-xiong:
The psm2 provider is the same.

The original understanding was that the auth_key_size returned from fi_getinfo() indicates the size of the auth_key expected by the provider. The actual auth_key is later set by the application when creating domain or endpoint.

prov/util/src/util_attr.c
```diff
@@ -654,6 +654,23 @@ int ofi_check_domain_attr(const struct fi_provider *prov, uint32_t api_version,
 		return -FI_ENODATA;
 	}
 
+	if (user_attr->auth_key_size &&
+	    (user_attr->auth_key_size != prov_attr->auth_key_size)) {
```
shefty, line 658: Based on @j-xiong's comment, lets change this check to
```
if (user_attr->auth_key && user_attr->auth_key_size &&
   (user_attr->auth_key_size != prov_attr->auth_key_size))
```
The below checks may not be needed then

PR #9440, opened 2023-10-13
Support for max_ep_auth_key

iziemba:
Since both of this changes require ABI updates, I decided to include them in the same PR. If it is easier to have these changes in separate PRs, I can separate it out.

Update: the optional_caps changes have been removed.

### DIFF
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -1187,6 +1187,8 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
 		attr->data_progress = hints->data_progress;
 	if (hints->av_type)
 		attr->av_type = hints->av_type;
+	if (hints->max_ep_auth_key)
+		attr->max_ep_auth_key = hints->max_ep_auth_key;
 }
 
 static void fi_alter_ep_attr(struct fi_ep_attr *attr,
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -661,6 +661,13 @@ int ofi_check_domain_attr(const struct fi_provider *prov, uint32_t api_version,
 		return -FI_ENODATA;
 	}
 
+	if (FI_VERSION_GE(api_version, FI_VERSION(1, 20)) &&
+	    user_attr->max_ep_auth_key > prov_attr->max_ep_auth_key) {
+		OFI_INFO_CHECK_SIZE(prov, prov_attr, user_attr,
+				    max_ep_auth_key);
+		return -FI_ENODATA;
+	}
+
 	return 0;
 }
 
--- a/src/abi_1_0.c
+++ b/src/abi_1_0.c
@@ -212,6 +212,79 @@ fi_info_1_3 -> fi_info
 			memcpy(dst, src, sizeof(*src));	\
 	} while (0);
 
+#define ofi_free_attr(attr)						\
+	do {								\
+		if (attr) {						\
+			free(attr);					\
+			attr = NULL;					\
+		}							\
+	} while (0);
+
+/* Macro used to duplicate versions of old fi_infos and convert it to the latest
+ * definition. Any new fields in the latest definition will be zeroed.
+ */
+#define ofi_dup_info(base, info)					\
+	do {								\
+		ofi_dup_attr(base, info);				\
+		if (!base)						\
+			break;						\
+									\
+		if (info->tx_attr) {					\
+			ofi_dup_attr(base->tx_attr, info->tx_attr);	\
+			if (!base->tx_attr)				\
+				goto ofi_dup_info_err_free_base;	\
+		}							\
+									\
+		if (info->rx_attr) {					\
+			ofi_dup_attr(base->rx_attr, info->rx_attr);	\
+			if (!base->rx_attr)				\
+				goto ofi_dup_info_err_free_tx_attr;	\
+		}							\
+									\
+		if (info->ep_attr) {					\
+			ofi_dup_attr(base->ep_attr, info->ep_attr);	\
+			if (!base->ep_attr)				\
+				goto ofi_dup_info_err_free_rx_attr;	\
+		}							\
+									\
+		if (info->domain_attr) {				\
+			ofi_dup_attr(base->domain_attr,			\
+				     info->domain_attr);		\
+			if (!base->domain_attr)				\
+				goto ofi_dup_info_err_free_ep_attr;	\
+		}							\
+									\
+		if (info->fabric_attr) {				\
+			ofi_dup_attr(base->fabric_attr,			\
+				     info->fabric_attr);		\
+			if (!base->fabric_attr)				\
+				goto ofi_dup_info_err_free_domain_attr;	\
+		}							\
+									\
+		/* Success. */						\
+		break;							\
+									\
+ofi_dup_info_err_free_domain_attr:					\
+		ofi_free_attr(base->domain_attr);			\
+ofi_dup_info_err_free_ep_attr:						\
+		ofi_free_attr(base->ep_attr);				\
+ofi_dup_info_err_free_rx_attr:						\
+		ofi_free_attr(base->rx_attr);				\
+ofi_dup_info_err_free_tx_attr:						\
+		ofi_free_attr(base->tx_attr);				\
+ofi_dup_info_err_free_base:						\
+		ofi_free_attr(base);					\
+	} while (0);
+
+#define ofi_free_info(base)						\
+	do {								\
+		ofi_free_attr(base->fabric_attr);			\
+		ofi_free_attr(base->domain_attr);			\
+		ofi_free_attr(base->ep_attr);				\
+		ofi_free_attr(base->rx_attr);				\
+		ofi_free_attr(base->tx_attr);				\
+		ofi_free_attr(base);					\
+	} while (0);
 
 /*
  * ABI 1.0
@@ -366,13 +439,13 @@ struct fi_info_1_1 *fi_dupinfo_1_1(const struct fi_info_1_1 *info)
 	if (!info)
 		return (struct fi_info_1_1 *) ofi_allocinfo_internal();
 
-	ofi_dup_attr(base, info);
+	ofi_dup_info(base, info);
 	if (base == NULL)
 		return NULL;
 
 	dup = fi_dupinfo(base);
 
-	free(base);
+	ofi_free_info(base);
 	return (struct fi_info_1_1 *) dup;
 }
 COMPAT_SYMVER(fi_dupinfo_1_1, fi_dupinfo, FABRIC_1.1);
@@ -418,13 +491,13 @@ struct fi_info_1_2 *fi_dupinfo_1_2(const struct fi_info_1_2 *info)
 	if (!info)
 		return (struct fi_info_1_2 *) ofi_allocinfo_internal();
 
-	ofi_dup_attr(base, info);
+	ofi_dup_info(base, info);
 	if (base == NULL)
 		return NULL;
 
 	dup = fi_dupinfo(base);
 
-	free(base);
+	ofi_free_info(base);
 	return (struct fi_info_1_2 *) dup;
 }
 COMPAT_SYMVER(fi_dupinfo_1_2, fi_dupinfo, FABRIC_1.2);
--- a/src/abi_1_0.c
+++ b/src/abi_1_0.c
@@ -196,14 +196,59 @@ struct fi_info_1_2 {
         struct fid_nic_1_2        *nic;
 };
 
-/*
+struct fi_domain_attr_1_3 {
+	struct fid_domain	*domain;
+	char			*name;
+	enum fi_threading	threading;
+	enum fi_progress	control_progress;
+	enum fi_progress	data_progress;
+	enum fi_resource_mgmt	resource_mgmt;
+	enum fi_av_type		av_type;
+	int			mr_mode;
+	size_t			mr_key_size;
+	size_t			cq_data_size;
+	size_t			cq_cnt;
+	size_t			ep_cnt;
+	size_t			tx_ctx_cnt;
+	size_t			rx_ctx_cnt;
+	size_t			max_ep_tx_ctx;
+	size_t			max_ep_rx_ctx;
+	size_t			max_ep_stx_ctx;
+	size_t			max_ep_srx_ctx;
+	size_t			cntr_cnt;
+	size_t			mr_iov_limit;
+	uint64_t		caps;
+	uint64_t		mode;
+	uint8_t			*auth_key;
+	size_t 			auth_key_size;
+	size_t			max_err_data;
+	size_t			mr_cnt;
+	uint32_t		tclass;
+};
+
 #define fi_tx_attr_1_3 fi_tx_attr
 #define fi_rx_attr_1_3 fi_rx_attr_1_2
 #define fi_ep_attr_1_3 fi_ep_attr_1_2
-#define fi_domain_attr_1_3 fi_domain_attr
 #define fi_fabric_attr_1_3 fi_fabric_attr_1_2
-fi_info_1_3 -> fi_info
-*/
+#define fid_nic_1_3 fid_nic_1_2
+
+struct fi_info_1_3 {
+        struct fi_info            *next;
+        uint64_t                  caps;
+        uint64_t                  mode;
+        uint32_t                  addr_format;
+        size_t                    src_addrlen;
+        size_t                    dest_addrlen;
+        void                      *src_addr;
+        void                      *dest_addr;
+        fid_t                     handle;
+        struct fi_tx_attr_1_3     *tx_attr;
+        struct fi_rx_attr_1_3     *rx_attr;
+        struct fi_ep_attr_1_3     *ep_attr;
+        struct fi_domain_attr_1_3 *domain_attr;
+        struct fi_fabric_attr_1_3 *fabric_attr;
+        struct fid_nic_1_3        *nic;
+};
 
 #define ofi_dup_attr(dst, src)				\
 	do {						\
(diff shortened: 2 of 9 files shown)


shefty (changes requested):
I'm unsure of the correctness of the optional_caps.  They might be correct, but it's not obvious from the code.  I believe we need to update fi_getinfo_test to verify that the optional_caps are being set/cleared as expected.

prov/util/src/util_attr.c
```diff
@@ -1175,7 +1176,13 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
...
 	}
 
-	attr->caps = ofi_get_caps(info_caps, hints ? hints->caps : 0, attr->caps);
+	if (hints) {
+		caps = hints->caps;
+		if (FI_VERSION_GE(api_version, FI_VERSION(1, 20)))
+			caps |= hints->optional_caps;
+	}
+
+	attr->caps = ofi_get_caps(info_caps, caps, attr->caps);
```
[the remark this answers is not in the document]
iziemba: Will fix and add tests.

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,259 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+	fi_freeinfo(*info);
+	*info = NULL;
+
+	if (!hints->domain_attr->name) {
+		ret = -FI_ENOMEM;
+		goto out;
+	}
+
+	/* Verify that all optional caps are ignore since requested API version
+	 * is less than 1.20.
```
shefty, line 1187: This isn't quite true.  The provider may only ignore primary caps.  It can still enable secondary caps, even if it never checks the optional_caps field.

This brings up a point that the optional_caps field really only applies to primary caps.

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,259 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+out:
+	return ret;
+}
+
+static int test_ignore_optional_cap(char *node, char *service, uint64_t flags,
+				    struct fi_info *hints,
+				    struct fi_info **info)
+{
+	uint64_t required_caps = FI_MSG | FI_TAGGED | FI_MULTI_RECV |
+		FI_LOCAL_COMM | FI_REMOTE_COMM;
```
shefty, line 1163: FI_MULTI_RECV, FI_LOCAL_COMM, FI_REMOTE_COMM are secondary caps and can always be set, even if not requested.  When setting the optional_caps, we'll want to specify primary caps only.  Maybe use FI_COLLECTIVE and FI_MULTICAST there for bits that are unlikely to be supported by the provider, and FI_RMA as a bit that is likely.

iziemba: Will fix.

fabtests/unit/getinfo_test.c
iziemba: I'll update this test to pass in only primary caps.

From API point-of-view, I think we still want to allow for secondary caps to be passed in via optional_caps. For example, if a provider does not support FI_SOURCE_ERR and the app supports either with or without FI_SOURCE_ERR, having the app pass this in via optional_caps would avoid multiple calls to fi_getinfo().

shefty: It's fine to allow the app to pass in secondary caps, but it doesn't make a difference from the app's viewpoint.  Secondary caps an always be returned even if not requested.  They're kind of always optional anyway.

j-xiong: The difference is that if app asks for secondary caps via the regular caps and the provider doesn't support it fi_getinfo() would fail.  A provider may choose not to report a secondary caps unless the application asks for it due to the overhead associated with supporting it. By passing via optional caps both scenarios are handled.

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,249 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+	caps = (*info)->caps;
+	domain_caps = (*info)->domain_attr->caps;
+	tx_caps = (*info)->tx_attr->caps;
+	rx_caps = (*info)->rx_attr->caps;
+
+	fi_freeinfo(*info);
+	*info = NULL;
+
+	/* Feed in all unsupported capabilities as optional caps. These should
+	 * not be set in the returned fi_infos.
```
shefty, line 1019: This isn't quite true.  The provider could return multiple fi_info structures with different capabilities.  That's true today.  So, some caps which aren't set in the topmost fi_info may be set in later fi_info's.  You could probably just say "Many of the unsupported caps are highly likely to be unset in returned fi_infos".  I'm not sure there are any guarantees here without checking the full list of fi_info's, which doesn't seem worthwhile.  I think the structure of this test is reasonable, but the check for invalid optional_caps may need to be:

```
/* At least 1 optional cap should have been rejected */
if (!(~info->caps & hints->optional_caps))
    return -FI_EINVAL;
```
```

File: fabtests/unit/getinfo_test.c (hunk 2 above)
Comment (shefty) [reply fabtests/unit/getinfo_test.c:1180]: That difference should be made clear to the app then.  If requested as optional, it will be returned if supported, even if there is a performance penalty for supporting it.

Comment (j-xiong) [reply fabtests/unit/getinfo_test.c:1180]: Right, if requested as optional, it will be returned if supported and still succeed if not supported. However, if requested as normal caps, it will fail if not supported.

Comment (shefty) [reply fabtests/unit/getinfo_test.c:1180]: I was referring to the cases of not being requested versus being requested as optional.

If the app is concerned about performance setting the secondary capability bit as an optional_cap is a worse choice than leaving it unset.  That result is not obvious and likely counterintuitive to a developer.  Since the bit is marked as optional, the app can obviously work around not having that capability.

Maybe we need a crisper definition of how a provider interprets an optional_cap.  Do we say that the provider MUST support the cap if it can, or that it MAY support the cap if it can?  And, if the later, what conditions do we allow the provider to say no?

It seems weird that a secondary capability bit suddenly becomes a mandatory requirement of the provider if set as an optional capability (assuming the provider supports the feature).

Comment (j-xiong) [reply fabtests/unit/getinfo_test.c:1180]: I think it is still up to the provider to decide whether it should report the support of an optional caps or not. It is also allowed that the provider may behave differently (or identically) for the case of a secondary cap not being requested vs requested as optional.

Comment (iziemba) [reply fabtests/unit/getinfo_test.c:1180]: One of the main use-cases for optional_caps would be apps can discover if a provider supports a capability without having to go through multiple iterations of passing in different or NULL hints to fi_getinfo(). For example, since FI_SOURCE is typically not reported by default from providers (even if supported), this could be a good way for users to discover if providers support FI_SOURCE. It would still be up to users if they want to use FI_SOURCE when allocating endpoints. Assuming since they have asked for it, they understand the potential performance implications.

>I think it is still up to the provider to decide whether it should report the support of an optional caps or not.

Allowing providers to selectively support optional_caps may result in users reverting to multiple iterations of hints and fi_getinfo() to discover features.

File: fabtests/unit/getinfo_test.c (hunk 4 above)
Comment (iziemba) [reply fabtests/unit/getinfo_test.c:1019]: Good point. I don't think it is worthwhile to iterate through the entire list. I'll update to the requested change.

File: fabtests/unit/getinfo_test.c (hunk 2 above)
Comment (shefty) [reply fabtests/unit/getinfo_test.c:1186]: I like Ian's suggestion that apps can then set the caps that they want based on what's returned as an optional_caps.  That would still need to be called out, since optional_caps would still be ignored on object allocation.

Comment (j-xiong) [reply fabtests/unit/getinfo_test.c:1186]: Looks good to me.

Comment (iziemba) [reply fabtests/unit/getinfo_test.c:1186]: This works for me.

**Comment by iziemba** [comment]
- Update the documentation based on the discussion
- The common util code has been updated to support optional_caps hints
- RxM has been updated to support optional_caps hints
- getinfo test has been updated to use RxM and optional_caps

File: fabtests/unit/getinfo_test.c [hunk 5]
```diff
@@ -964,6 +964,428 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+{
+	uint64_t required_caps = FI_MSG | FI_TAGGED;
+	int ret;
+
+	/* Verify that the provider meets the required capabilities of this
+	 * test.
+	 */
+	hints->caps = required_caps;
+	ret = fi_getinfo(FI_VERSION(1, 19), node, service, flags, hints, info);
+	if (ret != FI_SUCCESS) {
```
iziemba, line 1340: I am going to update these tests to use more of the fabtest test status and error reporting infrastructure

iziemba:
Can the Intel CI failures output be attached to this ticket?

Looking at the windows builds, I am not sure why debug build is failing while the release build is a success.

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,509 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+		sprintf(err_buf, "No providers found");
+		return -FI_ENOSYS;
+	}
+
+	hints->domain_attr->name = strdup((*info)->domain_attr->name);
```
iziemba, line 1416: This strdup is causing windows debug job to fail. Allocation is coming from one heap and being freed on another. Since it is not strdup is not strictly needed for this test, I'll remove it.

iziemba, line 1416: These strdup's are causing windows builds to crash. I am going to remove them.

prov/util/src/util_attr.c
```diff
@@ -1285,14 +1451,30 @@ static uint64_t ofi_get_info_caps(const struct fi_info *prov_info,
  * the hints have been validated and the starting fi_info is properly
  * configured by the provider.
  */
-void ofi_alter_info(struct fi_info *info, const struct fi_info *hints,
-		    uint32_t api_version)
+void ofi_alter_info_opt_caps(struct fi_info *info, const struct fi_info *hints,
```
iziemba, line 1454: Need to update util_getinfo() to call ofi_alter_info_opt_caps()

j-xiong:
Intel CI error: SHMEM test failed w/o any detailed console output. The only message got was "script exited with error code 143".

prov/rxm/src/rxm_attr.c
```diff
@@ -260,8 +260,13 @@ static struct fi_info rxm_verbs_info = {
...
 
+/* These caps are emulated and/or may degrade performance. */
+#define VALID_OPTIONAL_CAPS_ONLY (FI_DIRECTED_RECV | FI_SOURCE | FI_ATOMIC)
+
 struct util_prov rxm_util_prov = {
 	.prov = &rxm_prov,
 	.info = &rxm_verbs_info,
 	.flags = 0,
+	.opt_caps_valid_caps = ~VALID_OPTIONAL_CAPS_ONLY,
+	.opt_caps_valid_opt_caps = VALID_OPTIONAL_CAPS_ONLY,
```
shefty, line 271: These 2 fields are the same, except for flipping the bits.  Better to keep only opt_caps_valid_opt_caps, but under a more reasonable name, since that directly lists what caps are optional.

include/ofi_util.h
```diff
@@ -162,6 +162,21 @@ struct util_prov {
...
+	 */
+	uint64_t			opt_caps_valid_caps;
+
+	/* Bit mask of optional capabilities which can be returned in
+	 * optional_cap fields. Recommendation is to set this to caps which
+	 * providers have to emulate and/or cannot support optimally. Setting to
+	 * zero will result in no optional capabilities being returned in
+	 * optional caps.
+	 */
+	uint64_t			opt_caps_valid_opt_caps;
```
shefty, line 179: Providers can specify optional_caps in their base fi_info's.  I think we should make use of those structs, rather than assuming the optional caps apply provider wide.  They may apply per fi_info.

We could have the provider set optional_caps be either the recommended or not recommended caps and update the util code accordingly.  (I would go with not recommended).  The info->caps would continue to be the full set of caps, same as today.

This path likely reduces the changes needed to the util code.  If optional_caps = 0, the current behavior is used, so we don't have to update all providers.  If optional_caps is set, then the util code would adjust the returned info->caps based on hints->caps and hints->optional_caps, removing those flags set in the provider info->optional_caps and not in hints->caps.

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,477 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+		FT_UNIT_STRERR(err_buf,
+			       "FI_ATOMIC incorrectly returned in caps",
+			       ret);
+		return ret;
+	}
+
+	fi_freeinfo(*info);
+	*info = NULL;
+
+	/* Verify that using domain_attr::optional_caps, tx_attr::optional_caps,
```
shefty, line 1320: I'm not following the purpose of the testing in this function from here down.

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,477 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+		sprintf(err_buf, "No providers found");
+		return -FI_ENOSYS;
+	}
+
+	/* This test is only setup to support selected providers. */
+	if (!strstr(hints->fabric_attr->prov_name, "ofi_rxm")) {
+		sprintf(err_buf, "%s not support for this test",
+			hints->fabric_attr->prov_name);
+		return -FI_ENOSYS;
+	}
```
shefty, line 1139: I think we can remove this and test any provider

fabtests/unit/getinfo_test.c
```diff
@@ -964,6 +964,477 @@ getinfo_test(caps, 4, "Test for capability bit regression",
...
+	if (ret) {
+		FT_UNIT_STRERR(err_buf, "Optional caps incorrectly returned",
+			       ret);
+		return ret;
+	}
+
+	fi_freeinfo(*info);
+	*info = NULL;
+
+	/* Verify that using domain_attr::optional_caps, tx_attr::optional_caps,
```
shefty, line 1185: It's unclear what the test is checking in this function from here down.

src/abi_1_0.c
```diff
@@ -451,3 +518,55 @@ int fi_getinfo_1_2(uint32_t version, const char *node, const char *service,
...
+	struct fi_info *dup, *base;
+
+	if (!info)
+		return (struct fi_info_1_3 *) ofi_allocinfo_internal();
+
+	ofi_dup_attr(base, info);
+	if (base == NULL)
+		return NULL;
+
+	dup = fi_dupinfo(base);
```
j-xiong, line 544: This only works for the case that new fields are added into `struct fi_info`. However, for new fields in `tx_attr`, `rx_attr`, `domain_attr` etc, the resulting `dup` would have those new fields filled with unknown values.

The CI failure with SHMEM is caused by this. It failed at checking `domain_attr` for rxm due to a randomly large number was seen by the provider as `max_ep_auth_key`. The SHMEM test was pre-built against an older version of libfabric.

prov/rxm/src/rxm_attr.c
iziemba: These two fields serve different purposes:
1. The first defines the optional_caps which can be returned in caps fi_info output
2. The second defines the optional_caps which can be returned in optional_caps fi_info output

By setting both to zero, a provider will not support optional caps. This results in all providers uses the core infrastructure not having optional_caps supported out-of-the-box. Support has to be explicitly added in.

I could change this to the following:
bool optional_caps_enabled: If set, the provider supports optional caps
uint64_t optional_caps_only: If optional_caps is enabled, bit mask of capabilities which can only be returned in optional_caps fi_info output. All other provider capabilities would be eligible to be returned in caps fi_info output.

Thoughts?

src/abi_1_0.c
iziemba: Thanks for finding this. Will fix!

include/ofi_util.h
iziemba: I like the idea of using providers fi_info::optional_caps field. The issue is I need to know when optional_caps are supported and which optional_caps should be returned through fi_info::caps or fi_info::optional_caps. Thoughts on the below behavior?

```
/* Enable/disable optional caps. This would be global per provider.
bool optional_caps_enabled;

/* Per provider fi_info::optional_caps value. If optional_caps are supported, this
 * defines the optional capabilities which can only be returned in fi_info::optional_caps.
 * All other optional capabilities are eligible to be returned in fi_info::caps.
 */
bool prov_optional_caps;

/* All the capabilities supported by the provider. */
uint64_t prov_caps;

/* User input from fi_info hints. */
uint64_t hints_optional_caps;
uint64_t hints_caps;

/* fi_info::caps and fi_info::optional_caps returned to users. */
uint64_t info_caps = 0;
uint64_t info_optional_caps = 0;

if (optional_caps_enabled) {
	uint64_t requested_caps = hints_caps | hints_optional_caps;
	uint64_t supported_caps = requested_caps & prov_caps;
	
	info_caps = supported_caps & ~prov_optional_caps;
	info_optional_caps = supported_caps & prov_optional_caps;
} else {
	info_caps = hints_caps;
}

return info_caps, info_optional_caps
```

fabtests/unit/getinfo_test.c
iziemba: This test is trying to verify the hints::{tx,rx}_attr::optional_caps against a super-set of hints::caps.

1. With all optional_caps == 0, verify that only the user request hints::{tx,rx}_attr::caps, which are a subset of hints::caps, is returned in the fi_info from fi_getinfo(). This is the default behavior of fi_getinfo() and really doesn't test optional capabilities.
2. Adding in hints::{tx,rx}_attr::optional_caps == FI_MSG to the previous test, the returned fi_info::{tx,rx}_attr::caps should now contain FI_MSG.

fabtests/unit/getinfo_test.c
iziemba: This test is trying to verify the hints::{tx,rx}_attr::optional_caps against a super-set of hints::caps.

1. With all optional_caps == 0, verify that only the user request hints::{tx,rx}_attr::caps, which are a subset of hints::caps, is returned in the fi_info from fi_getinfo(). This is the default behavior of fi_getinfo() and really doesn't test optional capabilities.
1. Adding in hints::{tx,rx}_attr::optional_caps == FI_ATOMIC to the previous test, the returned fi_info::{tx,rx}_attr::optional_caps (not caps) should now contain FI_ATOMIC. For RxM, since FI_ATOMIC may negatively impact performance, when requested as a optional_cap, it should only be returned in optional_cap output.

fabtests/unit/getinfo_test.c
iziemba: I think that is safe to remove for this test. The subsequent test, test_valid_suboptimal_optional_cap, is more RxM specific since it expects specific handling of FI_ATOMIC optional_cap.

src/abi_1_0.c
j-xiong: fi_dupinfo_1_1() and fi_dupinfo_1_2() need to be updated similarly.

iziemba: Hmm I think the are more problems that just this with the ABI compat. For example, the FABRIC_1.2 ABI compat uses a 1.1 domain attr definition which is wrong.

iziemba: Actually, may be OK. I may have to redo my compat support.

iziemba: Ok. I uploaded a patch to fix this.

iziemba:
I would like to take a step back and re-examine optional caps. The way the behavior is currently defined is leading to this complicated util behavior.

The objective of optional_caps is for users to avoid multiple calls to fi_getinfo() to discover non-essential provider caps. The current proposal essentially ORs in the optional_caps to outputted fi_info::caps. What if we do not do this OR, and require the user to OR in the caps. So the outputted fi_info::optional_caps would be essentially equal to hints::optional_caps & provider_caps.

j-xiong:
@iziemba I am in favor of your updated proposal. Making the optional caps separated from regular caps should simplify the logic.

iziemba:
On second thought, unless we force all providers to support optional caps, I do not know how useful the feature is in preventing applications for using a single fi_getinfo() call to discover provider caps. In addition, I am not convinced it is needed to support FI_AV_USER_ID as defined in https://github.com/ofiwg/libfabric/pull/9319. Thus, I think I am going to remove it from this PR and we can revisit in the future (if needed).

j-xiong (approved):
Looks good.

PR #9441, opened 2023-10-13
core, fabtests, prov/verbs: Add base_addr to fi_mr_dmabuf

sunkuamzn:
base_addr field can be used to pass in the virtual address of the memory region to be registered

### DIFF
--- a/fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c
+++ b/fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c
@@ -177,6 +177,7 @@ void reg_dmabuf_mr(void)
 		.fd = xe_get_buf_fd(buf),
 		.offset = 0,
 		.len = buf_size,
+		.base_addr = NULL,
 	};
 	struct fi_mr_attr mr_attr = {
 		.dmabuf = &dmabuf,
--- a/include/rdma/fi_domain.h
+++ b/include/rdma/fi_domain.h
@@ -139,6 +139,7 @@ struct fi_mr_dmabuf {
 	int		fd;
 	uint64_t	offset;
 	size_t		len;
+	void 		*base_addr;
 };
 
 struct fi_mr_attr {
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -543,6 +543,10 @@ by iov_count.
 
 ## dmabuf
 
+DMA-buf registrations are used to share device memory between a given
+device and the fabric NIC and does not require that the device memory
+be mmap'ed into the virtual address space of the calling process.
+
 This structure references a DMA-buf backed device memory region.  This field
 is only usable if the application has successfully requested support for
 FI_HMEM and the FI_MR_DMABUF flag is passed into the memory registration
@@ -554,17 +558,20 @@ struct fi_mr_dmabuf {
 	int      fd;
 	uint64_t offset;
 	size_t   len;
+	void     *base_addr;
 };
 ```
 The fd is the file descriptor associated with the DMA-buf region.  The
 offset is the offset into the region where the memory registration should
 begin.  And len is the size of the region to register, starting at the
-offset.  The selection of dmabuf over the mr_iov field is controlled by
-specifying the FI_MR_DMABUF flag.
+offset. The base_addr is the page-aligned starting virtual address of
+the memory region allocated by the DMA-buf. If a base virtual address
+is not available (because, for example, the calling process has not
+mapped the memory region into its address space), base_addr can be
+set to NULL.
 
-DMA-buf registrations are used to share device memory between a given
-device and the fabric NIC and does not require that the device memory
-be mmap'ed into the virtual address space of the calling process.
+The selection of dmabuf over the mr_iov field is controlled by
+specifying the FI_MR_DMABUF flag.
 
 ## iov_count
 
--- a/prov/verbs/src/verbs_mr.c
+++ b/prov/verbs/src/verbs_mr.c
@@ -115,8 +115,8 @@ failover:
 #endif
 
 static int
-vrb_mr_reg_common(struct vrb_mem_desc *md, int vrb_access, const void *buf,
-		  size_t len, void *context, enum fi_hmem_iface iface,
+vrb_mr_reg_common(struct vrb_mem_desc *md, int vrb_access, const void *base_addr,
+		  const void *buf, size_t len, void *context, enum fi_hmem_iface iface,
 		  uint64_t device, uint64_t flags)
 {
 	if (!ofi_hmem_is_initialized(iface)) {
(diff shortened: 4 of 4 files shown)


man/fi_mr.3.md
```diff
@@ -554,18 +558,22 @@ struct fi_mr_dmabuf {
...
 The fd is the file descriptor associated with the DMA-buf region.  The
 offset is the offset into the region where the memory registration should
 begin.  And len is the size of the region to register, starting at the
-offset.  The selection of dmabuf over the mr_iov field is controlled by
+offset. The base_addr is the base virtual address of the memory region to
+be registered. If a base virtual address is not available (because, for
+example, the calling process has not mapped the memory region into its
+address space), base_addr can be set to NULL. If both base_addr and
+offset are provided, then the page offset of both the base_addr and
+offset parameters must be identical.
```
j-xiong, line 572: base address is supposed to be the page aligned starting address of the allocation.  offset is the difference between the virtual address and the base address.
```
virtual address = base address + offset
```

man/fi_mr.3.md
```diff
@@ -554,18 +558,22 @@ struct fi_mr_dmabuf {
...
 	uint64_t offset;
 	size_t   len;
+	void     *base_addr;
 };
 ```
 The fd is the file descriptor associated with the DMA-buf region.  The
 offset is the offset into the region where the memory registration should
 begin.  And len is the size of the region to register, starting at the
-offset.  The selection of dmabuf over the mr_iov field is controlled by
+offset. The base_addr is the base virtual address of the memory region to
```
shijin-aws, line 567: I would modify this to `base_addr is the base address of the dmabuf region that is mapped to the virtual memory space.` IIUC the base_addr + offset is the base virtual address for the registered memory.

shijin-aws (changes requested):

prov/verbs/src/verbs_mr.c
```diff
@@ -252,9 +252,10 @@ vrb_reg_dmabuf(struct vrb_domain *domain, const struct fi_mr_attr *attr,
 	 * to monitor or verify if the region is invalidated.
 	 */
 	ret = vrb_mr_nocache_reg(domain, NULL, attr->dmabuf->len, attr->access,
-				 attr->dmabuf->offset, attr->requested_key,
-				 flags, mr, attr->context, attr->iface,
-				 (uint64_t) attr->dmabuf->fd);
+				 (uint64_t) attr->dmabuf->base_addr +
+					 attr->dmabuf->offset,
```
j-xiong, line 256: This won't work. In the end this will reach ibv_reg_dmabuf_mr() as offset. We need to pass both offset and virtual address down.

include/rdma/fi_domain.h
```diff
@@ -139,6 +139,7 @@ struct fi_mr_dmabuf {
 	int		fd;
 	uint64_t	offset;
 	size_t		len;
+	void 		*base_addr;
```
shijin-aws, line 142: better to use `uint64_t`, otherwise you will have a warning

```
prov/efa/src/efa_mr.c:565:31: warning: pointer of type 'void *' used in arithmetic [-Wpointer-arith]
    mr_attr->dmabuf->base_addr + mr_attr->dmabuf->offset,
                               ^
prov/efa/src/efa_mr.c:565:4: warning: passing argument 4 of 'ibv_reg_dmabuf_mr' makes integer from pointer without a cast [-Wint-conversion]
    mr_attr->dmabuf->base_addr + mr_attr->dmabuf->offset,
    ^~~~~~~
```

shefty: This is a user facing API structure.  void * aligns with the value that he user has.

Internally, pointers should be converted to integers by casting to (uintptr_t)

man/fi_mr.3.md
sunkuamzn: I see, thanks

prov/verbs/src/verbs_mr.c
shijin-aws: you cannot pass iova as NULL when offset is not 0. They must have the same page offset

sunkuamzn: Thanks. Updated PR.

prov/verbs/src/verbs_mr.c
```diff
@@ -357,11 +357,11 @@ vrb_mr_reg_iface(struct fid *fid, const void *buf, size_t len, uint64_t access,
 			      util_domain.domain_fid.fid);
 
 	if (domain->cache.monitors[iface])
-		return vrb_mr_cache_reg(domain, buf, len, access, offset,
+		return vrb_mr_cache_reg(domain, buf, len, access, NULL, offset,
```
sunkuamzn, line 360: The verbs MR cache path doesn't support DMA buf right now. I did not add that support.

shijin-aws:
A side note: I didn't see how we are updating MR map for dmabuf MR - the current logic relies on `attr->mr_iov[0].iov_base` in `ofi_mr_map_verify`

prov/verbs/src/verbs_mr.c
shefty: We shouldn't try to cache dmabuf registrations.  The cache is mainly for internal registrations.

prov/verbs/src/verbs_mr.c
```diff
@@ -138,7 +138,7 @@ vrb_mr_reg_common(struct vrb_mem_desc *md, int vrb_access, const void *buf,
 #if VERBS_HAVE_DMABUF_MR
 	if (flags & FI_MR_DMABUF)
 		md->mr = ibv_reg_dmabuf_mr(md->domain->pd, (uintptr_t) buf,
-					   len, (uintptr_t) buf,
+					   len, (uintptr_t) base_addr,
```
j-xiong, line 141: Should use 'base_addr + buf (which is offset here)' as iova.

sunkuamzn: Fixed

fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c
```diff
@@ -177,6 +177,7 @@ void reg_dmabuf_mr(void)
 		.fd = xe_get_buf_fd(buf),
 		.offset = 0,
 		.len = buf_size,
+		.base_addr = NULL;
```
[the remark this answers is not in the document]
j-xiong: That's the reason of Intel CI failure.

prov/verbs/src/verbs_mr.c
sunkuamzn: Why is that?

shijin-aws: Open MPI 5.0 is doing fi_mr_regattr for each fi_tsenddata https://github.com/open-mpi/ompi/blob/fee1d707432d4586cf5d1df8f3e3fd2aeff52469/ompi/mca/mtl/ofi/mtl_ofi.h#L340. It relies on the MR cache to reduce the costs of MR registration for both host and hmem memory.

sunkuamzn:
bot:aws:retest

shijin-aws:
@shefty @j-xiong ok for merge?

j-xiong (approved):

shijin-aws (approved):

PR #9449, opened 2023-10-18
prov/efa: do not insert shm av inside efa progress engine

shijin-aws:
Currently, efa progress engine will call efa_av_insert_one() when it gets a message from unknown peer through efa device. which means peer is not local or shm is disabled for transmission. We shouldn't insert shm av anyway in this case.
Also, calling fi_av_insert internally inside progress engine is violating Libfabric standard for FI_AV_TABLE.

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -284,10 +284,11 @@ void efa_conn_release(struct efa_av *av, struct efa_conn *conn);
  *
  * @param[in]	av	address vector
  * @param[in]	conn	efa_conn object
+ * @param[in]	insert_shm_av	whether insert address to shm av
  * @return	On success return 0, otherwise return a negative error code
  */
 static
-int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn)
+int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn, bool insert_shm_av)
 {
 	int err, ret;
 	char smr_name[EFA_SHM_NAME_MAX];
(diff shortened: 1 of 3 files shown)


prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -739,8 +739,15 @@ fi_addr_t efa_rdm_pke_insert_addr(struct efa_rdm_pke *pkt_entry, void *raw_addr)
...
+	/*
+	 * The message is from a peer through efa device, which means peer is not local
+	 * or shm is disabled for transmission.
+	 * We shouldn't insert shm av anyway in this case.
+	 * Also, calling fi_av_insert internally inside progress engine is violating
+	 * Libfabric standard for FI_AV_TABLE.
+	 */
 	ret = efa_av_insert_one(ep->base_ep.av, (struct efa_ep_addr *)raw_addr,
-	                        &rdm_addr, 0, NULL);
+	                        &rdm_addr, 0, NULL, false);
```
wzamazon, line 750: Was this open called by progress engine?

prov/efa/src/efa_av.c
```diff
@@ -680,7 +691,7 @@ int efa_av_insert(struct fid_av *av_fid, const void *addr,
 	for (i = 0; i < count; i++) {
 		addr_i = (struct efa_ep_addr *) ((uint8_t *)addr + i * EFA_EP_ADDR_LEN);
 
-		ret = efa_av_insert_one(av, addr_i, &fi_addr_res, flags, context);
+		ret = efa_av_insert_one(av, addr_i, &fi_addr_res, flags, context, !!(av->shm_rdm_av));
```
j-xiong, line 694: `av->shm_rdm_av` is checked inside `efa_conn_rdm_init`, passing `true` here should be enough.

prov/efa/src/rdm/efa_rdm_pke_cmd.c
shijin-aws: `efa_rdm_pke_insert_addr` is only called inside progress engine

wzamazon (approved):

prov/efa/src/efa_av.c
shijin-aws: Yes, you are right, updated. Thx

wenduwan (approved):
Approving as a temporary fix.

We will discuss about the long-term solution around av insert and srx locking.

PR #9452, opened 2023-10-19
prov/verbs: Add synapseai dmabuf mr support

nbeer-habana:
The following changes were made to enable Gaudi-direct support over libfabric verbs provider, to allow peer-to-peer communication using dmabuf mr. They are applicble to device memory with iface value of FI_HMEM_SYNAPSEAI:
* Add function for dmabuf registration with synapseai.
* Use rndv protocol for rxm msg send to avoid copy between host and device memory.


prov/verbs/src/verbs_mr.c
```diff
@@ -57,6 +57,64 @@ static struct fi_ops vrb_mr_fi_ops = {
...
+		return NULL;
+	}
+	if (failover_policy == TRY) {
+		failover_policy = ALWAYS;
+		FI_INFO(&vrb_prov, FI_LOG_MR,
+			"Failover on: ibv_reg_dmabuf_mr() ==> ibv_reg_mr()\n");
+	}
+	return mr;
+}
+#endif
```
j-xiong: This function is mostly the same as vrb_reg_ze_dmabuf().  The duplication can be eliminated by:
(1) add an `iface` argument to the call;
(2) add `get_dmabuf_fd` method to `ofi_hmem_ops` and define `ofi_hmem_get_dmabuf_fd()`;
(2) define `get_base_addr` method for synapseai;
(3) use `ofi_hmem_get_dmabuf_fd()` and `ofi_hmem_get_base_addr()` instead of iface specific calls;
(4) change the function name to something like `vrb_reg_dmabuf`.

j-xiong (commented):
Also please add `prov/verbs: ` prefix to the title of the second commit since that's the scope of the changes.

prov/verbs/src/verbs_mr.c
j-xiong: See #9466 for step (2) mentioned in previous comment.

j-xiong: @nbeer-habana I noticed that in your patch, the `offset` passed to `ibv_reg_dmabuf_mr` is always 0. Does Synapseai mandate that the `buf` here must be the starting address of the dmabuf object? If so, is it checked inside the `hcclLookupDMABuff` call? If not, is there a way to get the base address of the dmabuf object so that the offset can be calculated?

nbeer-habana: @j-xiong Yes, SynapseAI assumes that `buf` is the starting address of the dmabuf object. We don't see a use-case where we would need to register part of a dmabuf object.

j-xiong:
@nbeer-habana Please rebase. The majority of the needed changes are already covered by #9466. Your new patch probably only needs two lines.

PR #9454, opened 2023-10-19
rxm: use `gdrcopy` for injection

thomasgillis:
This PR adds the usage of gdrcopy when injection is used of send/recv and rma

- [x] grdcopy register is done as part of `fi_mr_regattr` and cleanup when closing the MR
- [x] used for eager send/recv
- [x] used when faking injection for RMA. When the base provider does the injection (e.g. `verbs`), nothing is done in `rxm`
- [x] ~used for SAR~ > I cannot get good performance with grdcopy and SAR

note: largely based on (https://github.com/ofiwg/libfabric/pull/8836)

comparison with and without gdrcopy (`FI_HMEM_CUDA_USE_GDRCOPY=0/1`):

Screenshot 2023-10-19 at 9 59 49 PM

### DIFF
--- a/prov/rxm/src/rxm_domain.c
+++ b/prov/rxm/src/rxm_domain.c
@@ -447,6 +447,11 @@ static int rxm_mr_close(fid_t fid)
 	if (rxm_mr->domain->util_domain.info_domain_caps & FI_ATOMIC)
 		rxm_mr_remove_map_entry(rxm_mr);
 
+	if (rxm_mr->hmem_handle) {
+		ofi_hmem_dev_unregister(rxm_mr->iface,
+					(uint64_t) rxm_mr->hmem_handle);
+	}
+
 	ret = fi_close(&rxm_mr->msg_mr->fid);
 	if (ret)
 		FI_WARN(&rxm_prov, FI_LOG_DOMAIN, "Unable to close MSG MR\n");
@@ -544,6 +549,8 @@ static void rxm_mr_init(struct rxm_mr *rxm_mr, struct rxm_domain *domain,
 	rxm_mr->mr_fid.mem_desc = rxm_mr;
 	rxm_mr->mr_fid.key = fi_mr_key(rxm_mr->msg_mr);
 	rxm_mr->domain = domain;
+	rxm_mr->hmem_flags = 0x0;
+	rxm_mr->hmem_handle = NULL;
 	ofi_atomic_inc32(&domain->util_domain.ref);
 }
 
@@ -553,7 +560,7 @@ static int rxm_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	struct rxm_domain *rxm_domain;
 	struct fi_mr_attr msg_attr = *attr;
 	struct rxm_mr *rxm_mr;
-	int ret;
+	int ret,gdrerr;
 
 	rxm_domain = container_of(fid, struct rxm_domain,
 				  util_domain.domain_fid.fid);
@@ -588,6 +595,16 @@ static int rxm_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	rxm_mr->iface = msg_attr.iface;
 	rxm_mr->device = msg_attr.device.reserved;
 	*mr = &rxm_mr->mr_fid;
+
+	gdrerr = ofi_hmem_dev_register(rxm_mr->iface, attr->mr_iov->iov_base,
+				       attr->mr_iov->iov_len,
+				       (uint64_t *) &rxm_mr->hmem_handle);
+	if (gdrerr) {
+		rxm_mr->hmem_flags = 0x0;
+		rxm_mr->hmem_handle = NULL;
+	} else {
+		rxm_mr->hmem_flags = OFI_HMEM_DATA_DEV_REG_HANDLE;
+	}
 
 	if (rxm_domain->util_domain.info_domain_caps & FI_ATOMIC) {
 		ret = rxm_mr_add_map_entry(&rxm_domain->util_domain,
--- a/prov/rxm/src/rxm_hmem.c
+++ b/prov/rxm/src/rxm_hmem.c
@@ -1,0 +1,81 @@
+#include "rxm.h"
+
+/**
+ * @brief copy list of IOVs from/to the host to/from an hmem device
+ *
+ * @param buf       host buffer address
+ * @param buf_size  size of the host buffer
+ * @param hmem_iov  list of IOVs on the device
+ * @param iov_count number of IOVs in the list
+ * @param iov_offset offset in bytes in the IOVs struct (cummulative over IOVs)
+ *
+ * note: inspired by efa_copy_from/to_hmem_iov
+ */
+ssize_t rxm_copy_hmem_iov(void **desc, char *buf, size_t buf_size,
+			  const struct iovec *hmem_iov, int iov_count,
+			  size_t iov_offset, int dir)
+{
+	int i, ret = -1;
+	ssize_t data_size = 0;
+
+	for (i = 0; i < iov_count; i++) {
+		char *hmem_buf;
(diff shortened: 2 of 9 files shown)


prov/rxm/src/rxm_domain.c
```diff
@@ -447,8 +447,13 @@ static int rxm_mr_close(fid_t fid)
 	if (rxm_mr->domain->util_domain.info_domain_caps & FI_ATOMIC)
 		rxm_mr_remove_map_entry(rxm_mr);
 
-	ret = fi_close(&rxm_mr->msg_mr->fid);
-	if (ret)
+    // release the gdrcopy if present
+    if (rxm_mr->hmem_handle) {
```
wenduwan, line 451: Being pedantic. This branch should also check iface == CUDA

Also alignment is off.

prov/rxm/src/rxm_domain.c
```diff
@@ -587,9 +594,20 @@ static int rxm_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
 	rxm_mr->device = msg_attr.device.reserved;
-	*mr = &rxm_mr->mr_fid;
-
-	if (rxm_domain->util_domain.info_domain_caps & FI_ATOMIC) {
+    *mr = &rxm_mr->mr_fid;
+
+    // get the GDR copy registration if possible, inspired from efa_mr_hmem_setup
+    if (flags & FI_HMEM_DEVICE_ONLY && cuda_is_gdrcopy_enabled()) {
+        int gdrerr =
+            cuda_gdrcopy_dev_register((struct fi_mr_attr *)attr, (uint64_t *)&rxm_mr->hmem_handle);
```
wenduwan, line 602: Actually I think both EFA and RXM should consider adopting the new API in https://github.com/ofiwg/libfabric/commit/fed171974610ec010c2433afbb54ba6e713f9663

shijin-aws: Yes, I am actually opening a PR https://github.com/ofiwg/libfabric/pull/9456/files to update this function's signature. It's weird to use mr_attr as input

shefty: Agree - let's wait for #9456 or rebase off those changes.

j-xiong:
The new source file needs also be added to the VC project in order to pass windows build.

prov/rxm/src/rxm_hmem.c
```diff
@@ -0,0 +1,81 @@
...
+				iface, (uint64_t) hmem_handle, host_buf,
+				dev_buf, size);
+		} else {
+			ofi_hmem_dev_reg_copy_to_hmem(iface,
+						      (uint64_t) hmem_handle,
+						      dev_buf, host_buf, size);
+		}
+		return FI_SUCCESS;
+	}
+	int ret;
```
j-xiong, line 73: Please move the variable declaration to the beginning of the function.

prov/rxm/src/rxm_domain.c
```diff
@@ -589,6 +597,20 @@ static int rxm_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	rxm_mr->device = msg_attr.device.reserved;
 	*mr = &rxm_mr->mr_fid;
 
+	// get the GDR copy registration if possible, inspired from
+	// efa_mr_hmem_setup
+	if (cuda_is_gdrcopy_enabled()) {
```
j-xiong, line 602: Controlling the behavior of non-cuda iface with a cuda only method does not look right. Can we have an iface-aware method here?

thomasgillis: that's an issue I have quickly discussed last week with @shijin-aws, I can try to add a cuda agnostic function but I would expect i to be another PR.
FWIW, the issue is shared with EFA (https://github.com/ofiwg/libfabric/blob/14594abdcf3b664eece27e4978efd1ae82497a94/prov/efa/src/efa_mr.c#L260).

shijin-aws: I merged https://github.com/ofiwg/libfabric/pull/9456. 

I think we should introduce a new ops named like `is_dev_reg_enabled`

prov/rxm/src/rxm_domain.c
```diff
@@ -589,6 +597,20 @@ static int rxm_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+		int gdrerr = ofi_hmem_dev_register(
+			rxm_mr->iface, attr->mr_iov->iov_base,
```
shijin-aws, line 604: I think for @j-xiong's comment ^^, shouldn't only do this whole thing in a if block of `if (rxm_mr->iface == FI_HMEM_CUDA)`? Then it's less concerning to me that you replace `rxm_mr->iface` to `FI_HMEM_CUDA` within such block

prov/rxm/src/rxm_domain.c
shijin-aws: To be fair, EFA's code ^^ is executed within a dedicated function for cuda iface (for `iface == FI_HMEM_CUDA`) so there is no risk here.

thomasgillis: I have started here: https://github.com/ofiwg/libfabric/pull/9469
but I am not familiar enough with the different backend (rocm etc) to finish it

shijin-aws: I think it is safe to leave it as false for the iface that you are not familiar. We are just introducing a new op and it's not necessary to make all providers to onboard it.

prov/rxm/src/rxm_domain.c
j-xiong: We have a good chance to enable other ifaces here. Limiting it to cuda only is less desirable.

j-xiong:
Mostly fine. Just some nit-picks.

j-xiong (approved):

j-xiong:
@shijin-aws Is the AWS CI failure related?

shijin-aws:
There is a failure in Windows build test

```
:19:33      5>C:\9454_1698595386\libfabric\prov\efa\src\efa_prov.c(89,1): warning C4113: 'void (__cdecl *)()' differs in parameter lists from 'void (__cdecl *)(void)' [C:\9454_1698595386\libfabric\libfabric.vcxproj]
09:19:33      5>C:\9454_1698595386\libfabric\prov\efa\src\efa_prov.c(168,1): warning C4431: missing type specifier - int assumed. Note: C no longer supports default-int [C:\9454_1698595386\libfabric\libfabric.vcxproj]
09:19:33      5>C:\9454_1698595386\libfabric\prov\efa\src\efa_prov.c(168,1): error C2054: expected '(' to follow 'EFA_INI' [C:\9454_1698595386\libfabric\libfabric.vcxproj]
09:19:33          efa_env.c
09:19:33      5>C:\9454_1698595386\libfabric\include\windows\co
```

But I have no idea how this PR can break it.

shijin-aws:
bot:aws:retest

j-xiong:
@shijin-aws I see what went wrong. PR #9472 was pulled with latest rebase. The file `include/windows/config.h` was updated by that PR, including a new definition of `#define HAVE_EFA 0`. That probably breaks the `/DHAVE_EFA` option passed to the windows build (defined in `libfabric.vcxproj`). 

One way to solve this is to wrap this specific definition in config.h with `#ifndef HAVE_EFA`.

shijin-aws:
I see, https://github.com/ofiwg/libfabric/pull/9472 already breaks AWS CI

shijin-aws:
I need to open a separate PR to fix the windows build. then this PR needs a rebase again

thomasgillis:
Sounds good, ping me when you have the PR ready :)

j-xiong:
@thomasgillis Please rebase to allow AWS CI to pass.

PR #9457, opened 2023-10-19
util: Add NIC FID functions to ofi.h

iziemba:
This enables providers to implement their own fid_nic while uses common functions.


j-xiong:
Why not just use fi_close() and fi_control()?

iziemba:
The CXI provider implements its own fid_nic and overrides the default fi_control and fi_close with its own implementation. The following is the fi_close example.

```
static int cxip_nic_close(struct fid *fid)
{
	struct fid_nic *nic = (struct fid_nic *) fid;

	free(nic->prov_attr);
	return ofi_nic_close(fid);
}
```

j-xiong (approved):

shefty (approved):

j-xiong:
CI failures are unrelated.

PR #9460, opened 2023-10-20
prov/tcp: Correct setting of completion flags 

nikhilnanal:
This PR makes 3 changes:
1. setting of FI_COMPLETION flags for data transfer interfaces(sends and rmas) by combining it with default  settings for the msg flags and op flags
2. Inheriting tx rx flags set during msg endpoint creation from the corresponding rdm endpoint when connection is being created.  For example, if the application requested selective completion then the corresponding msg endpoint is created with the same settings. 
3. re-enables mpichsuite rma tests into Intel CI for tcp provider.

### DIFF
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -583,18 +583,27 @@ xnet_set_commit_flags(struct xnet_xfer_entry *xfer, uint64_t flags)
 }
 
 static inline uint64_t
-xnet_tx_completion_flag(struct xnet_ep *ep, uint64_t op_flags)
+xnet_tx_completion_get_msgflags(struct xnet_ep *ep, uint64_t flags)
 {
-	/* Generate a completion if op flags indicate or we generate
-	 * completions by default
+	/* Generate a completion if msg flags indicate or app
+	 * requests
 	 */
-	return (ep->util_ep.tx_op_flags | op_flags) & FI_COMPLETION;
+	return (ep->util_ep.tx_msg_flags | flags) & FI_COMPLETION;
 }
 
 static inline uint64_t
 xnet_rx_completion_flag(struct xnet_ep *ep)
 {
 	return ep->util_ep.rx_op_flags & FI_COMPLETION;
+}
+
+static inline uint64_t
+xnet_tx_completion_get_opflags(struct xnet_ep *ep)
+{
+	/* Generate a completion if op flags indicate or we generate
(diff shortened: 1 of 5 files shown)


shefty (changes requested):
The patches aren't clean when viewed individually.  The first patch removes several blank lines and adds printf calls.  The second path adds several blank lines, and removes the printfs.  The blank lines and printfs shouldn't be part of the first patch, so it stands alone better for cherry picking.  The second patch then wouldn't need to cleanup after the first patch.

nikhilnanal:
> The patches aren't clean when viewed individually. The first patch removes several blank lines and adds printf calls. The second path adds several blank lines, and removes the printfs. The blank lines and printfs shouldn't be part of the first patch, so it stands alone better for cherry picking. The second patch then wouldn't need to cleanup after the first patch.

nikhilnanal:
I have corrected the commits and then clicked on close by mistake, reopened again. Passes on the Jenkins CI.

prov/tcp/src/xnet.h
```diff
@@ -583,12 +583,12 @@ xnet_set_commit_flags(struct xnet_xfer_entry *xfer, uint64_t flags)
...
-xnet_tx_completion_flag(struct xnet_ep *ep, uint64_t op_flags)
+xnet_tx_completion_get_msgflags(struct xnet_ep *ep, uint64_t flags)
 {
-	/* Generate a completion if op flags indicate or we generate
-	 * completions by default
+	/* Generate a completion if msg flags indicate or app
+	 * requests
 	 */
-	return (ep->util_ep.tx_op_flags | op_flags) & FI_COMPLETION;
+	return (ep->util_ep.tx_msg_flags | flags) & FI_COMPLETION;
```
j-xiong, line 591: `tx_msg_flags` and `tx_op_flags` are the same w.r.t. `FI_COMPLETION`. There is no behavior difference before and after this patch.  The `*_msg_flags` fields are used as a way to pass optional `FI_COMPLETION` to other util calls (such as in `util_srx_recvmsg()`). The `*_op_flags` should be sufficient to cover all the scenarios that the default flags are checked.

j-xiong: @shefty You are right. When enabling completion by default while using selective completions, the msg flags won't have FI_COMPLETION on.

j-xiong (approved):

PR #9463, opened 2023-10-21
prov/efa: close shm resource when it is disabled in ep 

shijin-aws:
When shm is disabled during ep enable, we should shut down shm resources as they will never be used. This will reduce unnecessary cost when poking shm ep/cq/av in the later operations.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -139,9 +139,25 @@ static int efa_domain_init_qp_table(struct efa_domain *efa_domain)
 static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *info)
 {
 	int err;
+	bool enable_shm = efa_env.enable_shm_transfer;
+
+	/* App provided hints supercede environmental variables.
+	 *
+	 * Using the shm provider comes with some overheads, so avoid
+	 * initializing the provider if the app provides a hint that it does not
+	 * require node-local communication. We can still loopback over the EFA
+	 * device in cases where the app violates the hint and continues
+	 * communicating with node-local peers.
+	 *
+	 */
+	if ((info->caps & FI_REMOTE_COMM)
+	    /* but not local communication */
+	    && !(info->caps & FI_LOCAL_COMM)) {
+		enable_shm = false;
+	}
 
 	efa_domain->shm_info = NULL;
-	if (efa_env.enable_shm_transfer)
+	if (enable_shm)
 		efa_shm_info_create(info, &efa_domain->shm_info);
 	else
 		EFA_INFO(FI_LOG_CORE, "EFA will not use SHM for intranode communication because FI_EFA_ENABLE_SHM_TRANSFER=0\n");
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -94,13 +94,19 @@ static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t coun
 
 	ofi_genlock_lock(srx_ctx->lock);
 
-	if (cq->shm_cq)
+	if (cq->shm_cq) {
 		fi_cq_read(cq->shm_cq, NULL, 0);
 
-	ret = ofi_cq_read_entries(&cq->util_cq, buf, count, src_addr);
+		/* 
+		 * fi_cq_read(cq->shm_cq, NULL, 0) will progress shm ep and write
+		 * completion to efa. Use ofi_cq_read_entries to get the number of
+		 * shm completions without progressing efa ep again.
+		 */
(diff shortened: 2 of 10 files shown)


prov/efa/src/efa_domain.c
```diff
@@ -139,9 +139,25 @@ static int efa_domain_init_qp_table(struct efa_domain *efa_domain)
...
+	/* App provided hints supercede environmental variables.
+	 *
+	 * Using the shm provider comes with some overheads, so avoid
+	 * initializing the provider if the app provides a hint that it does not
+	 * require node-local communication. We can still loopback over the EFA
+	 * device in cases where the app violates the hint and continues
+	 * communicating with node-local peers.
+	 *
+	 */
+	if ((info->caps & FI_REMOTE_COMM)
```
wzamazon, line 153: Is this part still needed? IMO, this is a mis-interpretation of FI_LOCAL_COMM capability.

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -94,13 +94,18 @@ static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t coun
 
 	ofi_genlock_lock(srx_ctx->lock);
 
-	if (cq->shm_cq)
+	if (cq->shm_cq) {
 		fi_cq_read(cq->shm_cq, NULL, 0);
 
-	ret = ofi_cq_read_entries(&cq->util_cq, buf, count, src_addr);
+		/* 
+		 * This call aims to check whether there are completions written
```
wzamazon, line 101: This comment is not very clear.  I think it would be better to explain why would we want to call `ofi_cq_read_entries` after calling `fi_cq_read` on shm_cq

wzamazon (commented):
Overall looks good to me.

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: Updated comment

shijin-aws:
> Is this part still needed? IMO, this is a mis-interpretation of FI_LOCAL_COMM capability.

For some reason github doesn't display this comment to me. I think you are right. I deleted this part as they are not needed

shijin-aws:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -935,12 +991,11 @@ void efa_rdm_ep_set_use_shm_for_tx(struct efa_rdm_ep *ep)
 	if ((ep->user_info->caps & FI_HMEM)
 	    && hmem_ops[FI_HMEM_CUDA].initialized
 	    && !ep->cuda_api_permitted) {
-		ep->use_shm_for_tx = false;
-		return;
+		use_shm = false;
 	}
```
a-szegel, line 995: If Neuron or a Habana Synapse device is initialized, we should set use_shm to false.

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -935,12 +991,11 @@ void efa_rdm_ep_set_use_shm_for_tx(struct efa_rdm_ep *ep)
...
 
-	ep->use_shm_for_tx = efa_env.enable_shm_transfer;
```
a-szegel, line 942: If Neuron or Habana Synapse device is successfully initialized, we should set use_shm to false here, and tear it down. We are currently seeing Neuron performance issues as well... and this seems like the perfect place to fix it.

a-szegel (changes requested):

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -94,13 +94,19 @@ static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t coun
 
 	ofi_genlock_lock(srx_ctx->lock);
 
-	if (cq->shm_cq)
+	if (cq->shm_cq) {
 		fi_cq_read(cq->shm_cq, NULL, 0);
 
-	ret = ofi_cq_read_entries(&cq->util_cq, buf, count, src_addr);
+		/*
```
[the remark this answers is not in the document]
shijin-aws: It is not a bug, but may cause some inefficiency. I didn't see it cause a perf regression though

prov/efa/src/rdm/efa_rdm_ep_fiops.c
a-szegel: It will still work functionally with EFA, but it will be slow.  I am seeing a ~3% application level regression from keeping SHM on.  There is no "ofi nccl plugin" for trainium, so the only way to turn it off is the env variable.

a-szegel (approved):

wzamazon (approved):

PR #9465, opened 2023-10-23
prov/efa: allocate pke_vec, recv_wr_vec, sge_vec from heap

shijin-aws:
Currently, efa_rdm_pke_recvv and efa_rdm_ep_bulk_post_internal_rx_pkts allocate pke_vec, recv_wr_vec, sge_vec from stack inside the function. This patch moves these allocation to heap during ep's creation.

### DIFF
--- a/prov/efa/src/dgram/efa_dgram_ep.h
+++ b/prov/efa/src/dgram/efa_dgram_ep.h
@@ -58,19 +58,6 @@ struct efa_send_wr {
 	struct ibv_sge sge[2];
 };
 
-struct efa_recv_wr {
-	/** @brief Work request struct used by rdma-core */
-	struct ibv_recv_wr wr;
-
-	/** @brief Scatter gather element array
-	 *
-	 * @details
-	 * EFA device supports a maximum of 2 iov/SGE
-	 * For receive, we only use 1 SGE
-	 */
-	struct ibv_sge sge[1];
-};
-
 
 int efa_dgram_ep_open(struct fid_domain *domain_fid, struct fi_info *info,
 		      struct fid_ep **ep_fid, void *context);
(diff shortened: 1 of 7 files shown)


sunkuamzn (approved):
LGTM. Thanks!

wenduwan:
My understanding is that heap allocation incurs higher overhead and requires more management. Did you observe performance gains after the change?

wzamazon (approved):
one possible change is to use `efa_recv_wr` instead of `ibv_recv_wr`. `efa_recv_wr` has an `ibv_sge` in it, therefore you just need 1 pool.

Also, the comment about 'the stack memory is allocated every time a function is called' is not accurate to my best knowledge. These memories was static with the function's memory space.

wenduwan:
> These memories was static with the function's memory space.

I'm not sure about this, but that is a good point.

I think instead of malloc at initialization, we can try 2 simpler options:
1. Declare the function stack variables as static, e.g.
```
static struct ibv_recv_wr recv_wr_vec[EFA_RDM_EP_MAX_WR_PER_IBV_POST_RECV], *bad_wr;
```
2. If we move the variables to endpoint, we can still declare them statically, e.g.
```
        struct ibv_recv_wr recv_wr_vec[EFA_RDM_EP_MAX_WR_PER_IBV_POST_RECV];
	struct ibv_sge sge_vec[EFA_RDM_EP_MAX_WR_PER_IBV_POST_RECV];
```

wenduwan:
Cool. I'm ok either way.

I'm not concerned with stack vs heap - before this change we were using stack and got enough memory. I do like the fact that we don't have to deal with alloc and free with stack variable though.

shijin-aws:
>one possible change is to use efa_recv_wr instead of ibv_recv_wr. efa_recv_wr has an ibv_sge in it, therefore you just need 1 pool.

Thank you! Updated

shijin-aws:
>Also, the comment about 'the stack memory is allocated every time a function is called' is not accurate to my best knowledge. These memories was static with the function's memory space.

I removed this statement in the commit message

wzamazon (approved):

sunkuamzn (approved):

PR #9466, opened 2023-10-23
hmem: Add new op to hmem_ops for getting dmabuf fd

j-xiong:
Now that we have multiple ifaces that support dmabuf, it makes sense to have a common method to retrieve the fd associated with the dmabuf object.

### DIFF
--- a/src/hmem_ze.c
+++ b/src/hmem_ze.c
@@ -1130,12 +1130,19 @@ int ze_hmem_host_unregister(void *ptr)
 	return FI_SUCCESS;
 }
 
-static int ze_get_dma_buf_fd(const void *addr)
+int ze_hmem_get_dmabuf_fd(void *addr, uint64_t size, int *fd, uint64_t *offset)
 {
+	int ret;
 	ze_result_t ze_ret;
 	ze_memory_allocation_properties_t mem_props = {};
 	ze_device_handle_t device_ptr;
 	ze_external_memory_export_fd_t export_fd = {};
+	void *base_addr;
+	uint64_t total_size;
+
+	ret = ze_hmem_get_base_addr(addr, size, &base_addr, &total_size);
+	if (ret)
+		return ret;
 
 	export_fd.stype = ZE_STRUCTURE_TYPE_EXTERNAL_MEMORY_EXPORT_FD;
 	export_fd.flags = ZE_EXTERNAL_MEMORY_TYPE_FLAG_DMA_BUF;
@@ -1151,7 +1158,9 @@ static int ze_get_dma_buf_fd(const void *addr)
 		return -FI_EINVAL;
 	}
 
-	return export_fd.fd;
+	*fd = export_fd.fd;
+	*offset = (uint64_t)(uintptr_t)addr - (uint64_t)(uintptr_t)base_addr;
+	return 0;
 }
 
 struct ze_dev_reg_handle {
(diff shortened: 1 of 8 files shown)


shijin-aws:
Thanks for the PR! I am working on something very close to this. In my work I have added `offset` as output

```
int cuda_get_dmabuf_fd_offset(uint64_t device, void *buf, size_t len,
			       int *dmabuf_fd, uint64_t *dmabuf_offset);
```

j-xiong:
@shijin-aws The correct cuda call to get the dmabuf fd is:
```
CUresult cuMemGetHandleForAddressRange ( void* handle, CUdeviceptr dptr, size_t size, CUmemRangeHandleType handleType, unsigned long long flags )
```
with handleType set to `CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD`. Neither device nor offset is needed here.

shijin-aws:
`offset` is an output, it returns the `offset` of the `va` in the dmabuf region. You think it's better for the caller to calculate itself?

For cuda dmabuf, it is recommended by nvidia to check whether dmabuf is supported for a given device, before making this `cuMemGetHandleForAddressRange` call.
```
cuDeviceGetAttribute(&is_supported,
				CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED, dev);
```

Again, it may be the caller's responsibility to make this call before calling the ops. I ever thought if it is safe to only check the device attribute once during hmem init and make it global (like gdrcopy). But I am not sure yet.

j-xiong:
Yes, checking the dmabuf support when cuda iface is being initialized makes more sense. The support depends on how the drivers are built, not something would change at runtime.

We have another call ofi_hmem_get_base_addr(). Offset is calculated as `address - base_address`. It does not belongs to here.

shijin-aws:
>We have another call ofi_hmem_get_base_addr(). Offset is calculated as address - base_address. It does not belongs to here.

Does `ofi_hmem_get_base_addr` necessarily gives you the page-aligned base address?

shijin-aws:
I searched a bit, it seems cudamalloc is only 256 bytes aligned

edgargabriel:
not sure whether its relevant, but maybe useful as an additional data point in the discussion: the interface to create a dmabuf handle on the ROCm runtime layer is

```
hsa_status_t hsa_amd_portable_export_dmabuf (const void *ptr, size_t size, int *fd, uint64_t *offset);
```

iziemba:
@edgargabriel : For ROCR, if hsa_amd_portable_export_dmabuf is called, does this need to be paired with hsa_amd_portable_close_dmabuf?

iziemba:
@edgargabriel : Thanks! Will hsa_amd_portable_export_dmabuf () result in a dma buf FD being opened such that the close is required? Or, does ROCM/ROCR already have a handle open and the same handle is returned via hsa_amd_portable_export_dmabuf ?

edgargabriel:
@iziemba I am not sure about the internal details, we could clarify that with the runtime/kernel team. However, I know that in the UCX code path where we added dmabuf support earlier this year, the close() is being executed on the dmabuf handle

chuckfossen:
> > We have another call ofi_hmem_get_base_addr(). Offset is calculated as address - base_address. It does not belongs to here.
> 
> Does `ofi_hmem_get_base_addr` necessarily give you the page-aligned base address? AFAICT it only returns the base address of cuda malloc. Is it host page aligned for sure?

The base address is not host page aligned.

j-xiong:
The issue here is that the base address is the allocation base, not the base for the dmabuf object which needs to be page aligned. Now we have both cuda and rocm that expect explicit offset output, I agree it makes sense to add that to this call. Will update the PR.

j-xiong:
PR has been updated with the new `offset` output added. The `efa` and `verbs` providers are also updated accordingly. 

One note on `neuron` and `synapseai`: the current usage model assumes that the address passed in is always the starting address of the dma-buf object. Comments have been added to make this clear. This limitation can be removed if a method to get the base address becomes available.

shijin-aws (approved):

src/hmem_ze.c
```diff
@@ -1130,12 +1130,19 @@ int ze_hmem_host_unregister(void *ptr)
...
 {
+	int ret;
 	ze_result_t ze_ret;
 	ze_memory_allocation_properties_t mem_props = {};
 	ze_device_handle_t device_ptr;
 	ze_external_memory_export_fd_t export_fd = {};
+	void *base_addr;
+	uint64_t total_size;
+
+	ret = ze_hmem_get_base_addr(addr, size, &base_addr, &total_size);
```
shijin-aws, line 1143: Does this necessarily be page aligned base addr?

j-xiong: Yes, that's true for Ze.

shijin-aws: Gotcha, I just learned it's the same for neuron. We may need to add such get_base_addr op for neuron in the future to make it consistent

aingerson (approved):

j-xiong:
Rebased to pick-up util cq fixes so that mpitests rma tests can pass over tcp.

iziemba (approved):

j-xiong:
Intel CI failure is with a random tcp mpich test. Unrelated to this PR.

PR #9468, opened 2023-10-23
prov/util: Fix revert PR #9377 to revert correct lock

a-szegel:
I mistakenly reverted the wrong lock on #9377, and caused a performance regression.  This change makes the ep->lock be domain dependent, and properly reverts util_cntr ep_list_lock to always be mutex.  This restores most of the performance regression.

### DIFF
--- a/prov/util/src/util_ep.c
+++ b/prov/util/src/util_ep.c
@@ -247,8 +247,9 @@ int ofi_endpoint_init(struct fid_domain *domain, const struct util_prov *util_pr
 	if (util_domain->eq)
 		ofi_ep_bind_eq(ep, util_domain->eq);
 
-	/* TODO Figure out how to optimize this lock for rdm and msg endpoints */
-	ret = ofi_genlock_init(&ep->lock, OFI_LOCK_MUTEX);
+	ret = ofi_genlock_init(&ep->lock,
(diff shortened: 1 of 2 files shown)


aingerson (commented):
Changes are good, but can you put into two separate patches to describe each one? I think in the util ep one you can also remove the comment you added that says to optimize it. Also typo in the commit message if you want to fix while you're at it (revers->reverts)

prov/util/src/util_ep.c
```diff
@@ -248,7 +248,9 @@ int ofi_endpoint_init(struct fid_domain *domain, const struct util_prov *util_pr
 		ofi_ep_bind_eq(ep, util_domain->eq);
 
 	/* TODO Figure out how to optimize this lock for rdm and msg endpoints */
-	ret = ofi_genlock_init(&ep->lock, OFI_LOCK_MUTEX);
+	ret = ofi_genlock_init(&ep->lock,
```
aingerson, line 251: Remove above comment since it is still optimized

aingerson (commented):
Remove comment otherwise looks good, thanks!

prov/util/src/util_ep.c
a-szegel: done

aingerson (approved):

PR #9469, opened 2023-10-23
hmem: remove cuda-specific functions when using `ofi_hmem_dev_reg_*`

thomasgillis:
remove cuda-specific naming and functions when using `ofi_hmem_dev_reg_*` function family
- [x] removes the need to check for `cuda_is_gdrcopy_enabled()`: `ofi_hmem_dev_register` fails if GDR copy is not available
- [x] update the bit mask to remove reference to cuda
- [x] ~update the EFA provider accordingly (@wenduwan do you want to do it here or in another PR?)~

### DIFF
--- a/include/ofi_mr.h
+++ b/include/ofi_mr.h
@@ -270,10 +270,10 @@ int ofi_mr_map_verify(struct ofi_mr_map *map, uintptr_t *io_addr,
 
 /* libfabric internal flags starting from 60 to 63 */
 /**
- * OFI_HMEM_DATA_GDRCOPY_HANDLE indicates that hmem_data points to
- * a gdrcopy_handle data structure
+ * OFI_HMEM_DATA_DEV_REG_HANDLE indicates that hmem_data points to
+ * a dev_reg data structure, e.g. gdrcopy handle in cuda
  */
-#define OFI_HMEM_DATA_GDRCOPY_HANDLE	(1ULL << 60)
+#define OFI_HMEM_DATA_DEV_REG_HANDLE	(1ULL << 60)
 
 struct ofi_mr {
 	struct fid_mr mr_fid;
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -410,17 +410,15 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
 			hmem_data = NULL;
 		}
 
-		if (hmem_iface == FI_HMEM_CUDA && (hmem_flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
-			/**
-			 * TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy
-			 * Note: buf must be on the host since gdrcopy does not support D2D copy
-			 */
+		if (hmem_flags & OFI_HMEM_DATA_DEV_REG_HANDLE) {
 			if (dir == OFI_COPY_BUF_TO_IOV)
-				cuda_gdrcopy_to_dev((uint64_t) hmem_data, hmem_buf,
-				                    (char *) buf + done, len);
+				ofi_hmem_dev_reg_copy_to_hmem(
+					hmem_iface, (uint64_t) hmem_data,
+					hmem_buf, (char *) buf + done, len);
 			else
-				cuda_gdrcopy_from_dev((uint64_t) hmem_data, (char *) buf + done,
-				                      hmem_buf, len);
+				ofi_hmem_dev_reg_copy_from_hmem(
+					hmem_iface, (uint64_t) hmem_data,
+					(char *) buf + done, hmem_buf, len);
 			ret = FI_SUCCESS;
 		} else if (dir == OFI_COPY_BUF_TO_IOV)
 			ret = ofi_copy_to_hmem(hmem_iface, hmem_device, hmem_buf,
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -314,7 +314,9 @@ int cuda_copy_from_dev(uint64_t device, void *dst, const void *src, size_t size)
 
 int cuda_dev_register(const void *addr, size_t size, uint64_t *handle)
 {
-	return cuda_gdrcopy_dev_register(addr, size, handle);
+	if (cuda_is_gdrcopy_enabled())
+		return cuda_gdrcopy_dev_register(addr, size, handle);
+	return -FI_ENOSYS;
 }
 
 int cuda_dev_unregister(uint64_t handle)
--- a/prov/efa/src/efa_hmem.h
+++ b/prov/efa/src/efa_hmem.h
@@ -90,7 +90,7 @@ static inline int efa_copy_from_hmem(void *desc, void *buff, const void *src, si
 		hmem_data = ((struct efa_mr *)desc)->peer.hmem_data;
 	}
 
-	if (FI_HMEM_CUDA == iface && (flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+	if (FI_HMEM_CUDA == iface && (flags & OFI_HMEM_DATA_DEV_REG_HANDLE)) {
 		assert(hmem_data);
 		/* TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy */
 		cuda_gdrcopy_from_dev((uint64_t)hmem_data, buff, src, size);
@@ -122,7 +122,7 @@ static inline int efa_copy_to_hmem(void *desc, void *dest, const void *buff, siz
 		hmem_data = ((struct efa_mr *)desc)->peer.hmem_data;
 	}
 
-	if (FI_HMEM_CUDA == iface && (flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+	if (FI_HMEM_CUDA == iface && (flags & OFI_HMEM_DATA_DEV_REG_HANDLE)) {
 		assert(hmem_data);
 		/* TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy */
 		cuda_gdrcopy_to_dev((uint64_t)hmem_data, dest, buff, size);
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -244,7 +244,7 @@ static int efa_mr_hmem_setup(struct efa_mr *efa_mr,
 
 	/* efa_mr->peer.device is an union. Setting reserved to 0 cleared everything in it (cuda, neuron, synapseai etc) */
 	efa_mr->peer.device.reserved = 0;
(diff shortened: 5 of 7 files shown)


shijin-aws:
AFAIK neuron doesn't support this yet. It is safe to make it as ENOSYS

thomasgillis:
what should I use for `rocm` and `ze`?

shijin-aws:
> Adds a new operation to `dev_reg` path:
> 
>     * [x]  system
> 
>     * [x]  cuda
> 
>     * [ ]  rocr
> 
>     * [ ]  ze
> 
>     * [ ]  others

AFAIK neuron doesn't support this yet. It is safe to make it as false. Likely the same for synapseai

shijin-aws:
>what should I use for rocm and ze?

I would check with @j-xiong  for ze, and @edgargabriel  for rocm

j-xiong:
I think it's fine to set ze and rocm to the default "no" variation. Both ze and rocm has the dev_reg function added, but we may want some control over whether or not to enable gdrcopy. We can later enbale ze and rocm seperately.

thomasgillis:
@edgargabriel the current version of `rocr_dev_register` simply gets the device and host pointer using
https://github.com/ofiwg/libfabric/blob/14594abdcf3b664eece27e4978efd1ae82497a94/src/hmem_rocr.c#L995-L996

is there any good side effect in enabling it by default?

thomasgillis:
Ok sounds good, will wait for his input then.
Should I introduce an environment variable? Or a default to always `true`?

include/ofi_hmem.h
```diff
@@ -120,6 +120,7 @@ struct ofi_hmem_ops {
 			     size_t *base_length);
 	bool (*is_ipc_enabled)(void);
 	int (*get_ipc_handle_size)(size_t *size);
+	bool (*is_dev_reg_enabled)(void);
```
iziemba, line 123: Hi Thomas,

Could you explain the use-case for this and why the current dev_reg/unreg APIs will not work?

The use-case that dev_reg/unreg enables is for providers to cache a dev_reg handle with their memory registration. During memory registration, if dev_reg fails, the expectation is the provider implements a fallback. I would expect the fallback to be using the non-dev_reg memcpy APIs which could hurt performance.

thomasgillis: sure, so all this came from the need of adding `cuda gdrcopy` to `rxm` (as done already in `efa`) when doing injection/inlining for send/recv. With cuda, `gdrcopy` is used through the `dev_reg` features and is conditionally registered at mr time (with cuda_is_gdrcopy_enabled):

https://github.com/ofiwg/libfabric/blob/c37c44e5b485b5185de96c53453a7249ca3fd672/prov/rxm/src/rxm_domain.c#L602-L613

https://github.com/ofiwg/libfabric/blob/9f604d658277c3a4851506327cb7a75572622767/prov/efa/src/efa_mr.c#L253-L256

one of the comment in [this PR](https://github.com/ofiwg/libfabric/pull/9454#discussion_r1369355045) highlighted that it doesn't make sense to keep the conditional in cuda as the `dev_reg` feature should be used by all the iface (rocr, cuda, ze, etc), hence the present draft PR.
It is still unclear whether or not it should be enabled by default for rocr iface. From your answer I extrapolate that it should be done at every `mr`?

iziemba: I guess I question if this is_dev_reg_enabled needed. I think the code in https://github.com/ofiwg/libfabric/blob/c37c44e5b485b5185de96c53453a7249ca3fd672/prov/rxm/src/rxm_domain.c#L602-L613 could be implemented as follows.

```
 	int ret = ofi_hmem_dev_register(rxm_mr->iface, attr->mr_iov->iov_base, attr->mr_iov->iov_len, (uint64_t *) &rxm_mr->hmem_handle);
	if (ret) {
		rxm_mr->hmem_flags = 0x0; 
 		rxm_mr->hmem_handle = NULL; 
	} else if (rxm_mr->iface == FI_HMEM_CUDA) {
		rxm_mr->hmem_flags = OFI_HMEM_DATA_GDRCOPY_HANDLE;
	}
```

thomasgillis: I don't have the history, but from the code one of the reason is that`ofi_hmem_dev_register` does not return an error if `FI_HMEM_CUDA_USE_GDRCOPY=0`.
@wenduwan @shijin-aws do you have more input?

Also, why only setting `OFI_HMEM_DATA_GDRCOPY_HANDLE` for cuda?
From Edgar's answer it would be useful for AMD as well.

iziemba: Making OFI_HMEM_DATA_GDRCOPY_HANDLE more generic for all HMEM seems to be the correct change.

wenduwan: @thomasgillis  Thanks for the heads up. I agree with @iziemba wrt if `ofi_hmem_dev_register` fails then the caller should have a fallback.

EFA only attempts `ofi_hmem_dev_register` if gdrcopy is enabled and available https://github.com/ofiwg/libfabric/blob/9f604d658277c3a4851506327cb7a75572622767/prov/efa/src/efa_mr.c#L253

Are you suggesting that we change `ofi_hmem_dev_register` so that it returns error if gdrcopy is disabled? I think that is also fine and backward compatible for EFA.

thomasgillis: I think making it fail would be more iface agnostic and help to comply to the if(error) don't do it strategy. I will try to make the changes asap then

thomasgillis: @wenduwan I have done the changes for the common part but only replaced the naming in EFA.
Is this the right way to do things to not break anything in EFA?

shijin-aws: Please make the efa change in the same commit where you renamed the macro. We shouldn't have a commit that cannot be compiled

include/ofi_mr.h
```diff
@@ -270,10 +270,10 @@ int ofi_mr_map_verify(struct ofi_mr_map *map, uintptr_t *io_addr,
 
 /* libfabric internal flags starting from 60 to 63 */
 /**
- * OFI_HMEM_DATA_GDRCOPY_HANDLE indicates that hmem_data points to
- * a gdrcopy_handle data structure
+ * OFI_HMEM_DATA_DEV_REG_HANDLE indicates that hmem_data points to
+ * a dev reg data structure
```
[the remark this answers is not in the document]
thomasgillis: is `OFI_HMEM_DATA_DEV_REG_HANDLE` the right name for this flag?

src/hmem_cuda.c
```diff
@@ -314,7 +314,10 @@ int cuda_copy_from_dev(uint64_t device, void *dst, const void *src, size_t size)
 
 int cuda_dev_register(const void *addr, size_t size, uint64_t *handle)
 {
-	return cuda_gdrcopy_dev_register(addr, size, handle);
+	if (cuda_gdr_copy_enable()) {
+		return cuda_gdrcopy_dev_register(addr, size, handle);
+	}
+	return FI_EIO;
```
iziemba, line 320: This change be its own commit. Also, -FI_EIO should be returned.

include/ofi_mr.h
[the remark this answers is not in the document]
wenduwan: It sounds reasonable to me.

wenduwan:
bot:aws:retest

src/hmem.c
```diff
@@ -404,17 +404,19 @@ static ssize_t ofi_copy_mr_iov(struct ofi_mr **mr, const struct iovec *iov,
 			hmem_data = NULL;
 		}
 
-		if (hmem_iface == FI_HMEM_CUDA && (hmem_flags & OFI_HMEM_DATA_GDRCOPY_HANDLE)) {
+		if (hmem_flags & OFI_HMEM_DATA_DEV_REG_HANDLE) {
 			/**
-			 * TODO: Fine tune the max data size to switch from gdrcopy to cudaMemcpy
-			 * Note: buf must be on the host since gdrcopy does not support D2D copy
+			 * TODO: Fine tune the max data size to switch from
+			 * def_reg copy to normal copy
```
wenduwan, line 410: I wasn't clear in my previous comment. I think we should remove this comment in src/hmem.c

This is the common code path and not concerned with tuning(which is up to the provider).

wenduwan:
bot:aws:retest

wenduwan (approved):
Thank you!

shijin-aws (approved):

shijin-aws:
need @aingerson's review for shm change

src/hmem_cuda.c
```diff
@@ -314,7 +314,10 @@ int cuda_copy_from_dev(uint64_t device, void *dst, const void *src, size_t size)
 
 int cuda_dev_register(const void *addr, size_t size, uint64_t *handle)
 {
-	return cuda_gdrcopy_dev_register(addr, size, handle);
+	if (cuda_is_gdrcopy_enabled()) {
```
aingerson, line 317: drop brackets here

src/hmem_cuda.c
```diff
@@ -314,7 +314,10 @@ int cuda_copy_from_dev(uint64_t device, void *dst, const void *src, size_t size)
...
+		return cuda_gdrcopy_dev_register(addr, size, handle);
+	}
+	return -FI_EIO;
```
aingerson, line 320: FI_ENOSYS makes more sense to me

aingerson (commented):
A couple of nitpicky things. shm change looks fine. I would prefer the second patch to have a little bit in the commit message to provide context behind the change.
Thanks!

src/hmem_cuda.c
thomasgillis: Hopefully the new commits address your change request. let me know if something else is needed at this point

shijin-aws:
bot:aws:retest

shijin-aws:

j-xiong:
@shijin-aws There was a CI script change in one of the PRs merged yesterday which turned on some tests. Successful run of those tests depends on other fixes in that PR.  You could do a rebase to pick up those changes, or we can ignore the Intel CI failure.

thomasgillis:
I can rebase if needed, let me know :-)

j-xiong:
@thomasgillis Then rebase :)

aingerson (approved):
Thanks!

thomasgillis:
@shijin-aws could you restart the aws ci? (hopefully for the last time)
@aingerson should i worry about the CI failure in `ze-shm-v3`?

shijin-aws:
bot:aws:retest

shijin-aws:
@j-xiong can we add @thomasgillis to the organization? Then I don't need to manually trigger AWS CI

aingerson:
@thomasgillis Node issue. All the other shm tests passed so it should be fine.

j-xiong:
@shijin-aws Just invited @thomasgillis to the organization.

iziemba (approved):

PR #9471, opened 2023-10-24
prov/efa: Improve `efa_rdm_pke` definition

darrylabbate:
This removes the hardcoded padding bytes from the struct definition, which were ostensibly used as a mechanism to enforce alignment of `wiredata` to the 128 byte boundary.

While the `static_assert`s somewhat protect against changes made to the members which the struct is composed of, the manual calculation is a maintenance headache which is prone to developer error, as well as implementation-defined quirks.

More importantly, this change more clearly indicates the intent to align to the 128-byte boundary.

This commit also makes the struct size uniform when debug mode is enabled. This wasn't (currently) necessary, as the 32 bytes of padding were already sufficient for the 16-byte `dlist_entry` struct.

This is the C11-compliant version, utilizing `_Alignas` (stdalign.h)

For sake of comparison, I also pushed a (less-preferred) C99 version: https://github.com/darrylabbate/libfabric/commit/b479b5cd76d7395d9ade57bbdc501d8e0c3f79a4

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke.h
+++ b/prov/efa/src/rdm/efa_rdm_pke.h
@@ -1,49 +1,21 @@
-/*
- * Copyright (c) Amazon.com, Inc. or its affiliates.
- * All rights reserved.
- *
- * This software is available to you under a choice of one of two
- * licenses.  You may choose to be licensed under the terms of the GNU
- * General Public License (GPL) Version 2, available from the file
- * COPYING in the main directory of this source tree, or the
- * BSD license below:
- *
- *     Redistribution and use in source and binary forms, with or
- *     without modification, are permitted provided that the following
- *     conditions are met:
- *
- *      - Redistributions of source code must retain the above
- *        copyright notice, this list of conditions and the following
- *        disclaimer.
- *
- *      - Redistributions in binary form must reproduce the above
- *        copyright notice, this list of conditions and the following
- *        disclaimer in the documentation and/or other materials
- *        provided with the distribution.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
- * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
- * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
- * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
- * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
- */
+/* Copyright Amazon.com, Inc. or its affiliates. All rights reserved. */
+/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
 
 #ifndef _EFA_RDM_PKE_H
 #define _EFA_RDM_PKE_H
 
+#include <stdalign.h>
 #include <uthash.h>
 #include <ofi_mem.h>
 #include <ofi_list.h>
 
 #define EFA_RDM_PKE_IN_USE		BIT_ULL(0) /**< this packet entry is being used */
 #define EFA_RDM_PKE_RNR_RETRANSMIT	BIT_ULL(1) /**< this packet entry encountered RNR and is being retransmitted*/
-#define EFA_RDM_PKE_LOCAL_READ	BIT_ULL(2) /**< this packet entry is used as context of a local read operation */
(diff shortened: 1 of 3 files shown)


prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -222,15 +191,12 @@ struct efa_rdm_pke {
...
 };
 
+
 #if defined(static_assert) && defined(__x86_64__)
-#if ENABLE_DEBUG
-static_assert(sizeof(struct efa_rdm_pke) == 256, "efa_rdm_pke check");
-#else
-static_assert(sizeof(struct efa_rdm_pke) == 128, "efa_rdm_pke check");
-#endif
+static_assert(sizeof (struct efa_rdm_pke) % EFA_RDM_PKE_ALIGNMENT == 0, "efa_rdm_pke alignment check");
```
shijin-aws, line 199: Do we need to make EFA_RDM_PKE_ALIGNMENT 256 when ENABLE_DEBUG=1?

darrylabbate: No. This change makes the size of `struct efa_rdm_pke` 128 bytes regardless of whether debug mode is enabled or not. The extra `struct dlist_entry` when debug mode is enabled is currently smaller than the unused space added by `_Alignas(128)`.

In the event the preceding struct members exceed 128 bytes, `wiredata` would be at the 256 byte offset, since `_Alignas(128)` specifies it be aligned to the nearest 128 byte boundary ([Try it online!](https://tio.run/##jY9dC4IwFIav5684FIEDEwoKywj6HRUy5lYD28LNG6PfbmciiH1fDZ7nnPc949MT500zVpoXVS5gY12uTHzeBgPECnXSLzQX8oUp7TzDBy5M6ZDCLSDWlRV3II3JZvPEE1LhRJI5yNroi9Buj@qYosp2HjEbIqDAz6yEnDm29/KeBoO4@WL5OW71TxwhAbmWuC7DkVW1MBKG565hUlcRGCmtcGg7HPuQ1h30KIJuNRzu0n7vyUTtEZSm39rxd@/aEf9ux6EP7Wj69nvTPAA)). So `EFA_RDM_PKE_ALIGNMENT` should remain as the intended alignment for `wiredata`

darrylabbate:
AWS CI failing due to Windows build test, which uses `msbuild.exe` to compile as suggested by [aws/efawin](https://github.com/aws/efawin). Likely due to the C11 feature. Curious if it can be configured to play nicely with C11.

prov/efa/src/rdm/efa_rdm_pke.h
sunkuamzn: Then why was there a `uint8_t pad[112]` in the `ENABLE_DEBUG` condition? It seems like we were trying to align to 256 when ENABLE_DEBUG=1

sunkuamzn: But performance doesn't matter in debug mode, so I think it's OK to align to 128 all the time

sunkuamzn (approved):

darrylabbate:
> AWS CI failing due to Windows build test, which uses `msbuild.exe` to compile as suggested by [aws/efawin](https://github.com/aws/efawin). Likely due to the C11 feature. Curious if it can be configured to play nicely with C11.

Reached out to @gurugubs for some guidance here. We can either override the `VisualStudioVersion` by specifying `-property:VisualStudioVersion=...` for `msbuild.exe`, or redefine it in the `.sln` files themselves. I'm not a Windows/VS expert though, so I may be wrong.

darrylabbate:
Added a (test) commit to bump `VisualStudioVersion` to 17 just to see what happens.

darrylabbate:
Per @gurugubs, the VS-related files are fine to edit manually (and have been in the past). Specifying `/std:c11` in `AdditionalOptions` seems to do the trick wrt C11 features, along with a new-enough VS version. Currently testing and trying out various changes to get a sufficient set of minimal changes.

darrylabbate:
bot:aws:retest

darrylabbate:
Capturing an internal Slack conversation: GCC 4.7 is the earliest version with `_Alignas` support. The few older OSs I tested all had GCC 4.8+ installed by default. From a cursory glance, [this list] also seems to indicate 4.8+ as the minimum default version for a respective `gcc` package.

darrylabbate:
Rebased commits

darrylabbate:
Rebasing again

sunkuamzn (approved):

prov/efa/src/rdm/efa_rdm_pke.h
shijin-aws: >Then why was there a uint8_t pad[112] in the ENABLE_DEBUG condition? It seems like we were trying to align to 256 when ENABLE_DEBUG=1

It was aimed to make both debug and non-debug builds to be 128 aligned. There are more members introduced in debug build which makes efa_rdm_pke exceed 128 bytes, so we seek for the next 128 aligned integer (256)

shijin-aws: We are safe as long as the total size of efa_rdm_pke can be multiples of 128 bytes for both builds

shijin-aws (approved):

shijin-aws:
Do you think this need to be part of 1.20 release? @darrylabbate

Issue #9296, opened 2023-09-07
Add FI_PROVIDER_FILTER and FI_HMEM_MONITOR_FILTER environment variables


jswaro:
*Is your feature request related to a problem? Please describe.*
In Libfabric deployments built with dlopen support, it is the case that Libfabric will probe and attempt to initialize each provider and HMEM monitor that has support built into Libfabric.
In order to support the widest variety of customers, vendors may compile all HMEM solutions into Libfabric and use dlopen to relax restrictions on deployment instead of directly compiling the library into Libfabric. In this case, if a customer only has a single library for a single vendor (CUDA, as an example), Libfabric will successfully initialize the CUDA HMEM monitor, but probe and fail on the other monitors.

Additionally, if vendors do not compile other providers into the Libfabric deployment, then Libfabric will probe the file system to discover any possible DL provider solutions that exist as part of provider discovery. 

In large scale deployments, the cost of traversing all possible library locations and failing at each one can be a substantial cost in terms of time and network operations for monitors that are known by the customer to not exist. At a specific scale, the amount of calls to the network file system can exceed the capacity of the metadata servers, causing an exponential increase to start-up time relative to the size of the job. 

*Describe the solution you’d like*
I think it would make sense to have a filter exposed via a environment variable to allow users to define which HMEM monitors and providers can be discovered.

Example 1:
`FI_PROVIDER_FILTER=verbs FI_HMEM_MONITOR_FILTER=cuda,rocm ./my_application`
 
The above invocation would limit Libfabric to initialize only the cuda and rocm monitor implementations, and only attempt to discover the verbs core provider. Any other implementation would not be initialized. This would result in fewer dlopen calls, and reduce the overall amount of time spent in start-up and provider discovery.

Example 2:
./my_appication
In this case, the environment variable isn’t set. This means that all providers and monitors will be probed, and initialized if present.

*Describe alternatives you’ve considered*

Alternative 1: source rebuild by customer/user
In this case, it would be possible to provide the user with libfabric and the options that the vendor would like them to use when compiling libfabric. Customers/users could then selectively enable the hmem providers without dlopen support, and omit all other implementations.
This provides customers with the ability to compile to the target environment, but results in a new requirement on customer/user deployments, and doesn’t solve the problem for heterogenous environments where some vendor GPUs are present on a portion of the hosts in a system, and other vendor GPUs are present on other hosts. In this case, probe/init failures are unavoidable.

Additionally, this does not address the DL provider probe cost. 

[...]

PR #9472, opened 2023-10-25
core: Add restricted dlopen flag to configure options

jswaro:
This commit introduces a new flag, --enable-restricted-dl. By default, restricted_dl is false, which continues the existing behavior of fi_getinfo. The current behavior of fi_getinfo causes the library to probe all paths in LD_LIBRARY_PATH for all possible provider implementations known by ofi_ordered_provs_init, or in cases where FI_PROVIDER_PATH is defined, a search along a colon-delimited set of paths for any library named lib*-fi.so. A side-effect of the non-FI_PROVIDER_PATH option is that a vast majority of the providers in the ofi_ordered_provs_init list, ordered_prov_names, will not exist on most systems. Usually a small subset of the providers exist, and the rest will not be present. A significant number of metadata fetches to the network filesystem will fail to find the desired library via dlopen incurring significant penalties during job launch at scale. As an application scales to larger rank and node counts, the impact by this behavior is expected to grow linearly, and then exponentially as the network file system metadata servers are overwhelmed.

This impact of dlopen misses can be largely mitigated with the use of FI_PROVIDER_PATH, but that eliminates the ability to naturally load DL libraries through dlopen's search capability unless the admin or workload manager specify possible paths for each known set of DL provider implementations.

This commit seeks to find a solution in the middle -- dlopen capability and natural discovery of DL providers without searching for every possible provider and no prerequisite knowledge of the available DL providers in the workload manager or module environments.

When enabled, ofi_ordered_provs_init will seed the initial prov_head with providers that satisfy the following conditions:
- The provider has been built into the library (detected at compile-time without --enable-only, or --enable-<provider>[=yes])
- The provider has been built standalone as a DL provider (--enable-<provider>=dl)
- The provider exists in the 'ordered_load_list' in ofi_ordered_provs_init.

Providers that do not meet the above conditions will not be loaded by libfabric if --enable-restricted-dl is provided to configure.

A similar feature will be required for HMEM provider solutions. Vendors may need to compile support for GPU libraries, but customers of the vendor may not use one or more of the supported GPU libraries.

### DIFF
--- a/src/fabric.c
+++ b/src/fabric.c
@@ -437,6 +437,17 @@ static struct fi_provider *ofi_get_hook(const char *name)
 	return provider;
 }
 
+struct load_entry {
+	const char* provider_name;
+	int should_load;
+};
+
+#define DEFINE_LOAD_ENTRY(NAME, SHOULD_LOAD) \
+    { \
+        .provider_name = NAME, \
+        .should_load = SHOULD_LOAD \
+    }
+
 /* This is the default order that providers will be accessed when available.
  * This, in turn, sets the default ordering of fi_info's reported to the user.
  * However, ofi_reorder_info() may re-arrange the list based on hard-coded
@@ -444,35 +455,61 @@ static struct fi_provider *ofi_get_hook(const char *name)
  */
 static void ofi_ordered_provs_init(void)
 {
-	char *ordered_prov_names[] = {
-		"efa", "psm2", "opx", "usnic", "gni", "bgq", "verbs",
-		"netdir", "psm3", "ucx", "ofi_rxm", "ofi_rxd", "shm",
-
+	struct load_entry ordered_load_list[] = {
+		DEFINE_LOAD_ENTRY("efa", (HAVE_EFA | HAVE_EFA_DL)),
+		DEFINE_LOAD_ENTRY("psm2", (HAVE_PSM2 | HAVE_PSM2_DL)),
+		DEFINE_LOAD_ENTRY("opx", (HAVE_OPX | HAVE_OPX_DL)),
+		DEFINE_LOAD_ENTRY("usnic", (HAVE_USNIC | HAVE_USNIC_DL)),
+		DEFINE_LOAD_ENTRY("gni", (HAVE_GNI | HAVE_GNI_DL)),
+		DEFINE_LOAD_ENTRY("bgq", (HAVE_BGQ | HAVE_BGQ_DL)),
+		DEFINE_LOAD_ENTRY("verbs", (HAVE_VERBS | HAVE_VERBS_DL)),
+		DEFINE_LOAD_ENTRY("netdir", 0),
+		DEFINE_LOAD_ENTRY("psm3", (HAVE_PSM3 | HAVE_PSM3_DL)),
+		DEFINE_LOAD_ENTRY("ucx", (HAVE_UCX | HAVE_UCX_DL)),
+		DEFINE_LOAD_ENTRY("ofi_rxm", (HAVE_RXM | HAVE_RXM_DL)),
+		DEFINE_LOAD_ENTRY("ofi_rxd", (HAVE_RXD | HAVE_RXD_DL)),
+		DEFINE_LOAD_ENTRY("shm", (HAVE_SHM | HAVE_SHM_DL)),
 		/* Initialize the socket based providers last of the
 		 * standard providers.  This will result in them being
 		 * the least preferred providers.
 		 */
 
 		/* Before you add ANYTHING here, read the comment above!!! */
-		"udp", "tcp", "sockets", "net", /* NOTHING GOES HERE! */
+		DEFINE_LOAD_ENTRY("udp", (HAVE_UDP | HAVE_UDP_DL)),
+		DEFINE_LOAD_ENTRY("tcp", (HAVE_TCP | HAVE_TCP_DL)),
+		DEFINE_LOAD_ENTRY("sockets", (HAVE_SOCKETS | HAVE_SOCKETS_DL)),
+		DEFINE_LOAD_ENTRY("net", 0), /* NOTHING GOES HERE! */
 		/* Seriously, read it! */
 
 		/* These are hooking providers only.  Their order
 		 * doesn't matter
 		 */
-		"ofi_hook_perf", "ofi_hook_trace", "ofi_hook_profile", "ofi_hook_debug",
-		"ofi_hook_noop", "ofi_hook_hmem", "ofi_hook_dmabuf_peer_mem",
+		DEFINE_LOAD_ENTRY("ofi_hook_perf", (HAVE_PERF | HAVE_PERF_DL)),
+		DEFINE_LOAD_ENTRY("ofi_hook_trace", (HAVE_TRACE | HAVE_TRACE_DL)),
+		DEFINE_LOAD_ENTRY("ofi_hook_profile", (HAVE_PROFILE_DL)),
+		DEFINE_LOAD_ENTRY("ofi_hook_debug", (HAVE_HOOK_DEBUG_DL)),
+		DEFINE_LOAD_ENTRY("ofi_hook_noop", 0),
+		DEFINE_LOAD_ENTRY("ofi_hook_hmem", (HAVE_HOOK_HMEM_DL)),
+		DEFINE_LOAD_ENTRY("ofi_hook_dmabuf_peer_mem",
+			(HAVE_DMABUF_PEER_MEM | HAVE_DMABUF_PEER_MEM_DL)),
 
 		/* So do the offload providers. */
-		"off_coll",
+		DEFINE_LOAD_ENTRY("off_coll", 0),
 	};
 	struct ofi_prov *prov;
 	int num_provs, i;
 
-	num_provs = sizeof(ordered_prov_names) / sizeof(ordered_prov_names[0]);
+	num_provs = sizeof(ordered_load_list) / sizeof(ordered_load_list[0]);
 
 	for (i = 0; i < num_provs; i++) {
-		prov = ofi_alloc_prov(ordered_prov_names[i]);
+		/* If restricted DL is enabled, then only attempt to load
+		 * providers which were present at compile-time that were
+		 * enabled as 'yes' or 'dl'
+		 */
+		if (HAVE_RESTRICTED_DL && !ordered_load_list[i].should_load)
+			continue;
+
+		prov = ofi_alloc_prov(ordered_load_list[i].provider_name);
 		if (prov)
 			ofi_insert_prov(prov);
 	}
--- a/src/hmem.c
+++ b/src/hmem.c
@@ -528,12 +528,70 @@ bool ofi_hmem_is_initialized(enum fi_hmem_iface iface)
 	return hmem_ops[iface].initialized;
 }
 
+void ofi_hmem_set_iface_filter(const char* iface_filter_str, bool* filter)
+{
+	int iface;
+	char* entry = NULL;
+	const char* token = ";";
+	const char* iface_labels[ARRAY_SIZE(hmem_ops)] = {
+		"system",
+		"cuda",
+		"rocr",
+		"ze",
+		"neuron",
+		"synapseai"
+	};
+
+	memset(filter, false, sizeof(bool) * ARRAY_SIZE(hmem_ops));
+
+	/* always enable system hmem interface */
+	filter[FI_HMEM_SYSTEM] = true;
+
+	entry = strtok(iface_filter_str, token);
+	while (entry != NULL) {
+		for (iface = 0; iface < ARRAY_SIZE(hmem_ops); iface++) {
+			if (!strcasecmp(iface_labels[iface], entry)) {
+				filter[iface] = true;
+				break;
+			}
+		}
+
+		if (iface >= ARRAY_SIZE(hmem_ops)) {
+			FI_WARN(&core_prov, FI_LOG_CORE,
+					"unknown HMEM interface specified in FI_HMEM, entry=\"%s\"\n",
+					entry);
+		}
+
+		entry = strtok(NULL, token);
+	}
+}
+
 void ofi_hmem_init(void)
 {
 	int iface, ret;
 	int disable_p2p = 0;
+	char* hmem_filter = NULL;
+	bool filter_hmem_ifaces = false;
+	bool iface_filter_array[ARRAY_SIZE(hmem_ops)];
+
+	fi_param_define(NULL, "hmem", FI_PARAM_STRING,
+			"List of hmem interfaces to attempt to initialize (default: all available interfaces)");
+	fi_param_get_str(NULL, "hmem", &hmem_filter);
+
+	if (hmem_filter) {
+		if (strlen(hmem_filter) != 0) {
+			ofi_hmem_set_iface_filter(hmem_filter, &iface_filter_array);
+			filter_hmem_ifaces = true;
+		} else {
+			FI_WARN(&core_prov, FI_LOG_CORE,
+					"zero-length FI_HMEM provided, enabling all interfaces\n");
+		}
+	}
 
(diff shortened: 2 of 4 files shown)


src/fabric.c
```diff
@@ -437,42 +437,79 @@ static struct fi_provider *ofi_get_hook(const char *name)
...
 	for (i = 0; i < num_provs; i++) {
-		prov = ofi_alloc_prov(ordered_prov_names[i]);
+		/* If restricted DL is enabled, then only attempt to load
+		 * providers which were present at compile-time that were
+		 * enabled as 'yes' or 'dl'
+		 */
+		if (HAVE_RESTRICTED_DL && !ordered_load_list[i].should_load)
+			continue;
+
+		prov = ofi_alloc_prov(ordered_load_list[i].provider_name);
```
j-xiong, line 512: This won't be able to limit which providers to load.

As I commented on #9296, this only allocate a provider structure so that they appear in the predefined order. Providers not allocated here are still going to be searched and opened later.  

See `ofi_init_dir`, `ofi_reg_dl_prov` and `ofi_register_provider` for more details.

jswaro: I've reviewed those functions before, but I don't see what you are trying to point out. Please help me to see what you are trying to point out. 

> See ofi_init_dir, ofi_reg_dl_prov and ofi_register_provider for more details.

`ofi_ini_dir` is only used by libfabric when `FI_PROVIDER_PATH` is provided. It happens here: https://github.com/ofiwg/libfabric/blob/main/src/fabric.c#L760. `dirs` is only set if provdir is set, which is directly associated with `FI_PROVIDER_PATH` on line 744. If `FI_PROVIDER_PATH` isn't defined, then libfabric will search the pre-populated `prov_head`. The `prov_head` is populated as part of `fi_ini` through `ofi_ordered_provs_init`. 

There are two entry points for `prov_head`/`ofi_insert_prov`: `ofi_register_provider`, and `ofi_ordered_provs_init`. For any other provider to be added to the list, it would need to go through the `ofi_register_provider` path -> which has entry points from `ofi_reg_dl_prov`, and `fi_ini`. 

When libfabric traverses into `ofi_register_provider` from `ofi_reg_dl_prov`, it only happens from a limited set of contexts: 
- when a provider is being searched via `FI_PROVIDER_PATH` (`ofi_ini_dir`-> `ofi_reg_dl_prov` -> `ofi_register_provider` -> `ofi_insert_prov`)
- when a provider already exists within the `prov_head` list, and DL providers in the process of being discovered by libfabric `ofi_find_prov_libs`. At line 700, the contents of `prov_head` is what was populated by `ofi_ordered_provs_init` (`ofi_find_prov_libs` -> `ofi_reg_dl_prov` -> `ofi_register_provider` -> `ofi_insert_prov`)

Otherwise, when `ofi_register_provider` is called from `fi_ini`, it is done within the context of basic initialization after `ofi_load_dl_prov` is called on line 888.  Since `ofi_load_dl_prov` is called before the [static initializers of the providers](https://github.com/ofiwg/libfabric/blob/main/src/fabric.c#L890-L919), none of the static initializers have the opportunity to affect the `prov_head` list to determine what providers could be discovered as DL loadable libraries. Even if the static initializers were moved ahead if the call to `ofi_load_dl_prov`, the first parameter of `ofi_register_provider` is the `<prov>_INIT` define, which would be NULL if support for the library wasn't present at the time libfabric was compiled. To me, this means that unless the provider is in `ofi_orders_provs_init`, it won't have DL provider support as a separate module shipped by the vendor independent of the distribution provider (libfabric from SuSE with libsm2-fi.so from the vendor -- as an example, if SuSE didn't enable sm2 in libfabric via --enable-sm2=yes|dl). As each of the static initializers are called, `fi_register_provider(GNI_INIT, NULL)` as an example, it will initialize the core provider that was built in, but will not attempt to load any DL providers because there isn't a call at this point which feeds back into the DL provider loads. Libfabric will add the pre-built/in-library version of the provide via the call to `ofi_insert_prov` and `ofi_init_prov` on lines 559-560, but it stops there. There are no additional calls to `ofi_reg_dl_prov` after the static initializers are called. Providers that do not exist in `prov_head` prior to the last call of `ofi_reg_dl_prov` are not discoverable. 

So at lines [744-747](https://github.com/ofiwg/libfabric/blob/main/src/fabric.c#L745-L747), if `FI_PROVIDER_PATH` isn't provided, then it will use what information was populated already to find dl libraries. `dirs` will be populated with the OFI `PROVDLDIR` default, and that will be the only path searched by `ofi_ini_dir`. 

Otherwise, if a value is provided to `FI_PROVIDER_PATH`, libfabric will delimit the string and traverse it by order given in the string, with special treatment for paths prefixed with `@`. 

The real problem that is being addressed here is the behavior for https://github.com/ofiwg/libfabric/blob/main/src/fabric.c#L746, as it searches the many different possible providers that could exist that do not. Every entry on `prov_head` will result in a call to `dlopen`. The only way that it does not happen is if `FI_PROVIDER_PATH` is set. 

-----------------------

This is observable from a fairly simple example. 

```
./autogen.sh 
./configure --enable-only --enable-tcp
./util/fi_info -p udp 
```
 
 No providers will be shown because only the TCP provider has been compiled in. However, every provider within the `ordered_prov_names` list will be called via `dlopen` on the expected DL provider name `lib<prov>-fi.so`. This results in N(number of providers known) * M (number of directories in LD_LIBRARY_PATH) * P (number of ranks per node) * Q (number of nodes allocated to the job) failed queries in the worst case scenario. This is a realistic problem for deployments which only use a very small subset of libfabric providers, and only gets worse at scale. 
 
 If you add debug code at line 645 to change the FI_DBG to a printf (required because it doesn't seem like the logging subsystem is fully set up by this time in the code), then you will see all of the spurious calls to dlopen, and if they don't exist on the system -- then every probe made by libdl will fail. The list of directories is longer than indicated by LD_LIBRARY_PATH because libdl does search extra directories for architecture specific variants of libraries. @timattox caught that.

j-xiong: What I tried to point out was that the method wouldn't prevent a provider from being loaded even if it is not in the list. However, you clarification indicates that this is not the purpose.

> The real problem that is being addressed here is the behavior for https://github.com/ofiwg/libfabric/blob/main/src/fabric.c#L746, as it searches the many different possible providers that could exist that do not. Every entry on prov_head will result in a call to dlopen.

If that's all what you want to address, then I think it's fine.

Nitpick:

> ofi_ini_dir is only used by libfabric when FI_PROVIDER_PATH is provided

When FI_PROVIDER_PATH is not set, this function is still called with a compile time set path. I see you also mentioned that later.

jswaro:
@j-xiong : I've added a second commit which rounds out the ask from #9296. I think from our discussion in the OFIWG call from 2-3 weeks ago, there is a missing analogous selection mechanism like FI_PROVIDER for the HMEM interfaces. In order to drive that discussion, I've added that comment here.

src/hmem.c
```diff
@@ -528,12 +528,74 @@ bool ofi_hmem_is_initialized(enum fi_hmem_iface iface)
...
+		llen = strlen(entry);
+
+		/* search the list of known hmem interfaces, switch to true if found */
+		for (iface = 0; iface < ARRAY_SIZE(hmem_ops); iface++) {
+			rlen = strlen(iface_labels[iface]);
+
+			/* if the length is the same, and entry matches the iface
+			 * string (case-insensitive), then enable the iface in the filter
+			 */
+			if (llen == rlen && !strncasecmp(iface_labels[iface], entry, llen)) {
```
j-xiong, line 565: Could just use `strcasecmp()` to simplify the logic.

src/hmem.c
```diff
@@ -528,12 +528,74 @@ bool ofi_hmem_is_initialized(enum fi_hmem_iface iface)
...
+	const char* iface_labels[ARRAY_SIZE(hmem_ops)] = {
+		"system",   // FI_HMEM_SYSTEM
+		"cuda",     // FI_HMEM_CUDA
+		"rocr",     // FI_HMEM_ROCR
+		"ze",       // FI_HMEM_ZE
+		"neuron",   // FI_HMEM_NEURON
+		"synapseai" // FI_HMEM_SYNAPSEAI
+	};
+
+	/* initialize array */
```
j-xiong, line 547: All the comments in the commit are unnecessary.

j-xiong (commented):
In addition to the comments below, I would use a more descriptive commit title such as: `core: Add environment variable for selective HMEM initialization`.

src/hmem.c
jswaro: That's fine. I lean toward more comments than less. I'll remove them.

src/hmem.c
jswaro: I'll make the switch. I tend to aim toward defensive strategies when dealing with user input.

jswaro:
@j-xiong : I've updated the PR with the feedback. I've left in one comment which should explain the lone statement, `filter[FI_HMEM_SYSTEM] = true`.

src/hmem.c
j-xiong: `strncasecmp` is not more defensive than `strcasecmp` here since (1) the prior call to `strlen` would have encountered the same risk, and (2) the string `entry` is known to have limited length.

src/hmem.c
```diff
@@ -528,12 +528,59 @@ bool ofi_hmem_is_initialized(enum fi_hmem_iface iface)
...
+	bool iface_filter_array[ARRAY_SIZE(hmem_ops)];
+
+	fi_param_define(NULL, "hmem", FI_PARAM_STRING,
+			"List of hmem interfaces to attempt to initialize (default: all available interfaces)");
+	fi_param_get_str(NULL, "hmem", &hmem_filter);
+
+	if (hmem_filter && strlen(hmem_filter) != 0) {
+		ofi_hmem_set_iface_filter(hmem_filter, &iface_filter_array);
+		filter_hmem_ifaces = true;
+	}
```
j-xiong, line 578: A question for discussion: does it make sense to treat empty setting of this variable (i.e. `FI_HMEM=`) as disabling all the HMEM interface?

src/fabric.c
```diff
@@ -437,42 +437,79 @@ static struct fi_provider *ofi_get_hook(const char *name)
...
 static void ofi_ordered_provs_init(void)
 {
-	char *ordered_prov_names[] = {
-		"efa", "psm2", "opx", "usnic", "gni", "bgq", "verbs",
-		"netdir", "psm3", "ucx", "ofi_rxm", "ofi_rxd", "shm",
-
+	struct load_entry ordered_load_list[] = {
+		DEFINE_LOAD_ENTRY("efa", (HAVE_EFA | HAVE_EFA_DL)),
+		DEFINE_LOAD_ENTRY("psm2", (HAVE_PSM2 | HAVE_PSM2_DL)),
+		DEFINE_LOAD_ENTRY("opx", (HAVE_OPX | HAVE_OPX_DL)),
```
j-xiong, line 461: Window build is broken becasue many of these `HAVE_*` definitions are missing. Need to update `include/windows/config.h`.

wenduwan:
bot:aws:retest

wenduwan:
bot:aws:retest

wenduwan:
bot:aws:retest

src/hmem.c
jswaro: I've been thinking about it as well. FI_HMEM set to the empty string could be used to say 'do not initialize anything' or just enable FI_HMEM_SYSTEM.

I think an empty setting should be treated as FI_HMEM_SYSTEM. I cannot conceive of a use case where FI_HMEM_SYSTEM shouldn't be available.

src/fabric.c
jswaro: Which build is the windows build, and if it isn't public -- what are the errors?

iziemba: https://ci.appveyor.com/project/ofiwg/libfabric/builds/48374739 is windows build

src/hmem.c
iziemba: It seems like FI_HMEM="" should be treated as a user error and a warn message logged. Since FI_HMEM="" is ambiguous, rather than make the assumption that the user wants FI_HMEM_SYSTEM, I think we should init all HMEM ifaces. Yes, this will result in more dlopens. But, the user needs to use this env properly if they want to avoid this.

src/fabric.c
jswaro: Thanks @iziemba

src/hmem.c
jswaro: FI_HMEM="" is currently set up to enable all provider. However, after implementing it like that, I saw it as a case for users to say -- do not init anything. 

The only question in my mind is whether we should make it very clear -- and init nothing, init everything, or throw an error. 

I would lean toward throwing an error perhaps because if a user writes a script to launch their application, and makes a mistake -- then they could fill out FI_HMEM with an empty string. It could be useful to them to log that as an unexpected input.

j-xiong: I think giving a message with FI_WARN() should be sufficient.

jswaro: Agreed. That is what I meant, but I can see where what I said might have been confusing.

jswaro:
```
C02ZH00YLVCJ:libfabric jswaro$ FI_LOG_LEVEL=warn FI_HMEM="" ./util/fi_info -p udp
libfabric:69111:1698431409::core:core:ofi_hmem_init():587<warn> zero-length FI_HMEM provided, enabling all interfaces
fi_getinfo: -96 (No message available on STREAM)
C02ZH00YLVCJ:libfabric jswaro$ FI_LOG_LEVEL=warn FI_HMEM="something" ./util/fi_info -p udp
libfabric:69125:1698431414::core:core:ofi_hmem_set_iface_filter():562<warn> unknown HMEM interface specified in FI_HMEM, entry="something"
fi_getinfo: -96 (No message available on STREAM)
C02ZH00YLVCJ:libfabric jswaro$ FI_LOG_LEVEL=warn FI_HMEM="cuda" ./util/fi_info -p udp
fi_getinfo: -96 (No message available on STREAM)
C02ZH00YLVCJ:libfabric jswaro$  ./util/fi_info -p udp
fi_getinfo: -96 (No message available on STREAM)
```

Examples of each case above.

jswaro:
```
C02ZH00YLVCJ:libfabric jswaro$ FI_LOG_LEVEL=warn FI_HMEM="cuda;rocr;something" ./util/fi_info -p udp
libfabric:69209:1698431519::core:core:ofi_hmem_set_iface_filter():562<warn> unknown HMEM interface specified in FI_HMEM, entry="something"
fi_getinfo: -96 (No message available on STREAM)
```

iziemba (approved):

j-xiong (approved):

j-xiong:

jswaro:
@j-xiong : I thought about this over the weekend, and there may be a more elegant solution than what was proposed. 

Instead of conditionally initializing or searching for the providers, an alternative could be to set prov_dir if the user hasn't set it. This would then bypass the function call that we are attempting to avoid anyway with a very minor change. 


```
        fi_param_define(NULL, "provider_path", FI_PARAM_STRING,
                        "Search for providers in specific path.  Path is "
                        "specified similar to dir1:dir2:dir3.  If the path "
                        "starts with @, loaded providers are given preference "
                        "based on discovery order, rather than version. "
                        "(default: " PROVDLDIR ")");

        fi_param_get_str(NULL, "provider_path", &provdir);
 
 #if HAVE_RESTRICTED_DL
        if (!provdir || !strlen(provdir)) {
            provdir = PROVDLDIR;
        }
 #endif
        
        if (!provdir || !strlen(provdir)) {
         ....
```

It would remove all of that init code and allow users to continue to use FI_PROVIDER_PATH as it works today. However, if they don't use FI_PROVIDER_PATH, then it would be set to PROVDLDIR, which skips the traversals of LD_LIBRARY_PATH via dlopen against every possible known provider.

j-xiong:
@jswaro That sounds good to me. Do you want to revert the merged commits and submit a new one?

jswaro:
No, it's fine. I'll make a separate PR for that.

PR #9481, opened 2023-10-27
hmem/cuda: Fallback to CUDA calls when NVML call fails.

shijin-aws:
The NVML calls to verify nvidia devices are nice-to-have features that allow a cheaper check and earlier exit when cuda devices are not present. When NVML call fails, we shouldn't fail the whole cuda hmem init but should fall back to the normal cuda api calls.

It is observed that users may use an NVML libraries that only contain stub functions for build purpose.

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -493,24 +493,33 @@ static int cuda_hmem_verify_devices(void)
 	nvmlReturn_t nvml_ret;
         cudaError_t cuda_ret;
 	unsigned int nvml_device_count = 0;
+	bool nvml_succeeded = false;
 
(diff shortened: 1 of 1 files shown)


j-xiong (approved):

wenduwan (approved):

shijin-aws:
Need to investigate the AWS CI failure

shijin-aws:
bot:aws:retest

j-xiong:
@shijin-aws Any update for this?

shijin-aws:
@j-xiong we are now inclined to close this PR as it will cause CUDA calls made on non-GPU platforms, which we aimed to avoid.

PR #9483, opened 2023-10-30
prov/verbs: Add synapseai dmabuf mr support

nbeer-habana:
The following changes were made to enable Gaudi-direct support over libfabric verbs provider, to allow peer-to-peer communication using dmabuf mr. They are applicble to device memory with iface value of FI_HMEM_SYNAPSEAI:

* Use dmabuf hmem registration function with synapseai.
* Use rndv protocol for rxm msg send to avoid copy between host and device memory.

### DIFF
--- a/prov/verbs/src/verbs_mr.c
+++ b/prov/verbs/src/verbs_mr.c
@@ -142,9 +142,10 @@ vrb_mr_reg_common(struct vrb_mem_desc *md, int vrb_access, const void *base_addr
 		md->mr = ibv_reg_dmabuf_mr(md->domain->pd, (uintptr_t) buf,
 					   len, (uintptr_t) base_addr + (uintptr_t) buf,
 					   (int) device, vrb_access);
-	else if (iface == FI_HMEM_ZE && vrb_gl_data.dmabuf_support)
+	else if ((vrb_gl_data.dmabuf_support && iface == FI_HMEM_ZE) ||
+		 (vrb_gl_data.dmabuf_support && iface == FI_HMEM_SYNAPSEAI))
 		md->mr = vrb_reg_hmem_dmabuf(iface, md->domain->pd, buf, len,
-					     vrb_access);
+						vrb_access);
 	else
 #endif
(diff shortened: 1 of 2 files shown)


ytava (commented):
Ignore, just check

prov/verbs/src/verbs_mr.c
```diff
@@ -73,8 +73,8 @@ static struct ibv_mr *vrb_reg_hmem_dmabuf(enum fi_hmem_iface iface,
 		[FI_HMEM_CUDA] = TRY,
 		[FI_HMEM_ROCR] = TRY,
 		[FI_HMEM_ZE] = TRY,
+		[FI_HMEM_SYNAPSEAI] = TRY,
```
j-xiong, line 76: Does synapseai have peer_mem driver to support the retry?

nbeer-habana: You are right, we don't want to fallback to this retry mechanism. I fixed it.

prov/verbs/src/verbs_mr.c
```diff
@@ -142,9 +142,10 @@ vrb_mr_reg_common(struct vrb_mem_desc *md, int vrb_access, const void *base_addr
...
 					   len, (uintptr_t) base_addr + (uintptr_t) buf,
 					   (int) device, vrb_access);
-	else if (iface == FI_HMEM_ZE && vrb_gl_data.dmabuf_support)
-		md->mr = vrb_reg_hmem_dmabuf(iface, md->domain->pd, buf, len,
-					     vrb_access);
+	else if (vrb_gl_data.dmabuf_support)
+		if (iface == FI_HMEM_ZE || iface == FI_HMEM_SYNAPSEAI)
+			md->mr = vrb_reg_hmem_dmabuf(iface, md->domain->pd, buf, len,
+							vrb_access);
 	else
```
j-xiong, line 149: The `else` would match the wrong `if`. Merge the two `if`s with `&&`.

j-xiong (approved):

PR #9486, opened 2023-10-30
core: simplify restricted-dl feature

### LINKED ISSUE #9296: Add FI_PROVIDER_FILTER and FI_HMEM_MONITOR_FILTER environment variables

jswaro:
The goal of the feature can be achieved more simply by setting FI_PROVIDER_PATH for the user to PROVDLDIR if the variable hasn't been set by the user. This bypasses the most expensive dlopen operations without forcing users to set FI_PROVIDER_PATH to achieve the greatest benefit from the new compile-time flag.

With this change, users can load any provider found in PROVDLDIR by default, which is more flexible than the prior solution which only allowed any provider which was present at compile-time. Additionally, since this change completely avoids the call to ofi_find_prov_libs which incurred the overwhelming majority of the dlopen penalties.

closes #9296

### DIFF
--- a/src/fabric.c
+++ b/src/fabric.c
@@ -437,17 +437,6 @@ static struct fi_provider *ofi_get_hook(const char *name)
 	return provider;
 }
 
-struct load_entry {
-	const char* provider_name;
-	int should_load;
-};
-
-#define DEFINE_LOAD_ENTRY(NAME, SHOULD_LOAD) \
-    { \
-        .provider_name = NAME, \
-        .should_load = SHOULD_LOAD \
-    }
-
 /* This is the default order that providers will be accessed when available.
  * This, in turn, sets the default ordering of fi_info's reported to the user.
  * However, ofi_reorder_info() may re-arrange the list based on hard-coded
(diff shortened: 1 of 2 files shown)


jswaro:
@j-xiong : PR is up. Thanks for reviewing the suggestion.

j-xiong:
@jswaro The description of the new configure option also needs to be changed accordingly. The behavior with the restricted dlopen enabled is more like `only look for dl providers under default location if FI_PROVIDER_PATH is not set`.

jswaro:
Right. Thanks for the catch.

jswaro:
Fixed

src/fabric.c
```diff
@@ -437,79 +437,42 @@ static struct fi_provider *ofi_get_hook(const char *name)
...
-		DEFINE_LOAD_ENTRY("opx", (HAVE_OPX | HAVE_OPX_DL)),
-		DEFINE_LOAD_ENTRY("usnic", (HAVE_USNIC | HAVE_USNIC_DL)),
-		DEFINE_LOAD_ENTRY("gni", (HAVE_GNI | HAVE_GNI_DL)),
-		DEFINE_LOAD_ENTRY("bgq", (HAVE_BGQ | HAVE_BGQ_DL)),
-		DEFINE_LOAD_ENTRY("verbs", (HAVE_VERBS | HAVE_VERBS_DL)),
-		DEFINE_LOAD_ENTRY("netdir", 0),
-		DEFINE_LOAD_ENTRY("psm3", (HAVE_PSM3 | HAVE_PSM3_DL)),
-		DEFINE_LOAD_ENTRY("ucx", (HAVE_UCX | HAVE_UCX_DL)),
-		DEFINE_LOAD_ENTRY("ofi_rxm", (HAVE_RXM | HAVE_RXM_DL)),
-		DEFINE_LOAD_ENTRY("ofi_rxd", (HAVE_RXD | HAVE_RXD_DL)),
```
iziemba, line 470: Do all the windows marcos need to be removed?

j-xiong: It's fine to leave them there.

j-xiong:
@jswaro Please rebase to allow AWS CI to pass.

PR #9489, opened 2023-10-30
include/ofi_atomic_queue: properly align atomic values

aingerson:
The atomic queue implementation uses atomic operations to allow threads/processes to claim and commit entries into the queue. Because of the use of atomics, padding is essential to allow atomic values to have a dedicated cache line (as atomics will lock the cache line over the course of the operation).
The atomic queue already had padding between the atomics to make sure they were not on the same cache line as each other but this didn't take into account the other fields that might be accessed by other processes as well such as the size and size mask. This could also potentially cause issues with any fields immediately preceeding the queue as well as with fields within the entrytype.
This patch requires the AQ base address to be cache line aligned and pads the atomic values properly so that they are guaranteed to be on their own cache line.

This patch also changes the shm AQ alignment to 64 to adhere to this new requirement.

### DIFF
--- a/include/ofi_atomic_queue.h
+++ b/include/ofi_atomic_queue.h
@@ -94,19 +94,28 @@ extern "C" {
 
 #define OFI_CACHE_LINE_SIZE (64)
 
+/*
+ * Base address of atomic queue must be cache line aligned to maximize atomic
+ * value perforamnce benefits
+ */
 #define OFI_DECLARE_ATOMIC_Q(entrytype, name)			\
 struct name ## _entry {						\
 	ofi_atomic64_t	seq;					\
 	bool		noop;					\
 	entrytype	buf;					\
-};								\
+} __attribute__((__aligned__(64)));				\
+								\
 struct name {							\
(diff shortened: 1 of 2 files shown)


j-xiong:
Is it a typo in the first line of commit message?  `allow .... from claiming and committing ...`

aingerson:
@j-xiong Yeah I guess that could be phrased better... I'll fix it. Thanks!

j-xiong:
@aingerson Please rebase to allow AWS CI to pass.

a-szegel (approved):

a-szegel:
I tested this and it seems to have great osu_alltoallw collective performance improvements!

aingerson:
@a-szegel Whenever you get a chance, could you share the AWS failure, please?

amirshehataornl (approved):

amirshehataornl (approved):

j-xiong:
bot:aws:retest

a-szegel:
It looks like previous failure was a random TCP failure, I expect this build to pass.

PR #9490, opened 2023-10-30
uitl/profile: Fix a bug in ofi_prof_add_event()

j-xiong:


j-xiong:
@peinanz Could you verify?

peinanz:
> @peinanz Could you verify?
Yes, it is a bug.

PR #9491, opened 2023-10-30
contrib/intel/jenkins: Switch dsa node to pikachu partition

zachdworkin:
The Mudkip partition is being moved to a different cluster. The new Pikachu node will be used instead of the Mudkip ones for DSA tests.


j-xiong:
No need to wait for CI. Jenkinsfile changes won't take effect until the PR is merged.

PR #9432, opened 2023-10-12
prov/efa: Use long CTS protocol if long read protocol fails because of memory registration limits

sunkuamzn:
The long read protocol can fail because of MR registration limits on the hardware. This PR has changes to switch to long CTS protocol when that happens.

PR #9493, opened 2023-10-31
prov/efa: Use long CTS protocol if runting read protocol fails because of memory registration limits

sunkuamzn:
The runting read protocol can fail because of MR registration limits on the hardware. This PR has changes to switch to long CTS protocol when that happens.

The first four commits are in https://github.com/ofiwg/libfabric/pull/9432


prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -131,11 +131,17 @@ int efa_rdm_pke_fill_data(struct efa_rdm_pke *pkt_entry,
 		ret = efa_rdm_pke_init_medium_tagrtm(pkt_entry, ope, data_offset, data_size);
 		break;
 	case EFA_RDM_LONGCTS_MSGRTM_PKT:
-		assert(data_offset == 0 && data_size == -1);
+		/* The data_offset will be non-zero when the long CTS RTM packet
+		 * is sent to continue a runting read transfer after the
+		 * receiver has run out of memory registrations */
+		assert((data_offset == 0 || ope->internal_flags & EFA_RDM_OPE_READ_NACK) && data_size == -1);
```
shijin-aws, line 137: So for runting read, receiver already gets some data before try to register the memory. If we fallback to long cts, we restart from the data_offset. Is that correct?

prov/efa/src/rdm/efa_rdm_pke_rtm.c
```diff
@@ -137,7 +137,14 @@ ssize_t efa_rdm_pke_init_rtm_with_payload(struct efa_rdm_pke *pkt_entry,
 	if (txe->internal_flags & EFA_RDM_OPE_READ_NACK)
 		rtm_hdr->flags |= EFA_RDM_REQ_READ_NACK;
 
-	if (data_size == -1) {
+	/* If this RTM packet is sent after the runting read protocol has failed
+	because of a MR registration limit on the receiver, we don't want to
+	send any data with the RTM packet. This is because the runting read RTM
+	packets have already delivered some of the data and the long CTS RTM
+	packet does not have a seg_offset field */
+	if (txe->internal_flags & EFA_RDM_OPE_READ_NACK) {
```
shijin-aws, line 145: so it is because long cts rtm currently doesn't support sending from a non-zero offset. data pkts supports offset so it is not impacted. Do I understand correctly?

prov/efa/src/rdm/efa_rdm_pke_cmd.c
sunkuamzn: Yes, that's correct

prov/efa/src/rdm/efa_rdm_pke_rtm.c
sunkuamzn: Yes, exactly

shijin-aws (approved):

shijin-aws:
@wzamazon could you have a look at it?

PR #9501, opened 2023-10-31
github/actions: migrate from "hub" to "gh"

j-xiong:
GitHub has removed "hub" from their runners since Oct.2, 2023. The recommendation is to use "gh" instead. One benefits is that "gh" has a command for merging the PR, with option to rebase. This allows simpler action script and makes it possible to make "rebase" as the only merge strategy allowed.


j-xiong:
Action scripts can only be tested on the main branch. No need to wait for CI. Let's merge and run the actions manually.

PR #9512, opened 2023-11-03
prov/efa: Enable inject rdma write

a-szegel:


a-szegel:
This file `prov/efa/src/rdm/efa_rdm_pkt_type.c` doesn't have a new line at the end... and it looks ugly in the diff so tried to add one to see if it made it look prettier, but it didn't, so then removed it.  Also first commit message to be more clear.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1473,15 +1472,17 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
 		if (OFI_UNLIKELY(!pkt_entry))
 			return -FI_EAGAIN;
 
+		/* Use bounce buffer (registered) for buffer/desc, user input doesn't matter */
```
sunkuamzn, line 1476: Nit:
> Use bounce buffer (registered) for buffer/desc, the buffer we pass to rdma-core doesn't matter

sounds better

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1458,7 +1458,6 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
 
 	assert(ope->iov_count > 0);
 	assert(ope->rma_iov_count > 0);
-	efa_rdm_ope_try_fill_desc(ope, 0, FI_WRITE);
```
sunkuamzn, line 1461: What's the reasoning for avoiding an MR reg call for zero byte write?

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1528,6 +1531,16 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
...
+			copied = ofi_copy_from_hmem_iov(pkt_entry->wiredata + sizeof(struct efa_rdm_rma_context_pkt),
+				ope->total_len, FI_HMEM_SYSTEM, 0, ope->iov, ope->iov_count, 0);
+			assert(copied == ope->total_len);
+			ope->desc[0] = fi_mr_desc(pkt_entry->mr);
+			ope->iov[0].iov_base = pkt_entry->wiredata + sizeof(struct efa_rdm_rma_context_pkt);
+		}
+
 		write_once_len = MIN(ope->iov[iov_idx].iov_len - iov_offset,
 				    ope->rma_iov[rma_iov_idx].len - rma_iov_offset);
 		write_once_len = MIN(write_once_len, max_write_once_len);
```
sunkuamzn, line 1546: You copy the data from user buffer to the packet entry but the `efa_rdm_pke_init_write_context` call a few lines below is still using the user buffer

sunkuamzn (changes requested):

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: `efa_rdm_ope_try_fill_desc()` calls `fi_mr_regv()` on the buffer.  This line assumes that the user will pass a real registered buffer to do a 0 byte write.  If you are doing a 0 byte write, then your buffer could be NULL, and your desc could be NULL b/c EFA shouldn't actually do anything with them.  The `fi_mr_regv()` will segfault on NULL desc.  Also segfaults for FI_INJECT which always gives a NULL desc.

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: I hack this by:
```
			ope->desc[0] = fi_mr_desc(pkt_entry->mr);
			ope->iov[0].iov_base = pkt_entry->wiredata + sizeof(struct efa_rdm_rma_context_pkt);
```

and that works b/c:
```
		efa_rdm_pke_init_write_context(
			pkt_entry, ope,
			(char *) ope->iov[iov_idx].iov_base + iov_offset,
			write_once_len, ope->desc[iov_idx],
			ope->rma_iov[rma_iov_idx].addr + rma_iov_offset,
			ope->rma_iov[rma_iov_idx].key);
```
passes in `ope->iov[iov_idx].iov_base + iov_offset` and `ope->desc[iov_idx]`, and it will only ever go around this loop once.

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: didn't quite use your words exactly, but made this more clear.

prov/efa/src/rdm/efa_rdm_ope.c
sunkuamzn: The txe has a separate `cq_entry` field that has the buffer and length data. That information is copied to txe when the txe is created. So changing the `ope->iov` doesn't affect the completion and this OK to do https://github.com/ofiwg/libfabric/blob/d43c492191a4f8a90b4531fd25deba536a87ecfd/prov/efa/src/rdm/efa_rdm_ope.c#L85-L86

sunkuamzn (approved):

prov/efa/src/rdm/efa_rdm_ope.c
shijin-aws: did you observe any performance impact for this commit?

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1531,6 +1534,16 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
 		if (OFI_UNLIKELY(!pkt_entry))
 			return -FI_EAGAIN;
 
+		if (ope->fi_flags & FI_INJECT) {
+			assert(ope->iov_count == 1);
+			assert(ope->total_len <= ep->inject_size);
+			copied = ofi_copy_from_hmem_iov(pkt_entry->wiredata + sizeof(struct efa_rdm_rma_context_pkt),
```
shijin-aws, line 1540: length is too long

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1531,6 +1534,16 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
...
+				ope->total_len, FI_HMEM_SYSTEM, 0, ope->iov, ope->iov_count, 0);
```
shijin-aws, line 1541: This doesn't look correct, `FI_HMEM_SYSTEM` needs to be replaced by the real hmem iface.

shijin-aws: Sorry I am wrong here, it should be system as there is no MR passed in for inject

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1531,6 +1534,16 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
...
+			assert(copied == ope->total_len);
```
shijin-aws, line 1542: This shouldn't be an assert. It should be warning and return error

a-szegel: It is an assert here: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke_utils.c#L123-L139

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: EFA doesn't enforce line length.

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: Looks like a speedup usec/xfer 1.30 (before), .98 (after) for
```
            "program": "/home/ec2-user/libfabric/fabtests/install/bin/fi_rma_bw",
            "args": ["-e", "rdm", "-o", "write", "-S", "10", "-p", "efa", "-E=9235"],
```

prov/efa/src/rdm/efa_rdm_ope.c
sunkuamzn: https://github.com/ofiwg/libfabric/blob/387a7338492e0b875a1f1b668b818f55b1babcda/prov/efa/src/rdm/efa_rdm_pke_utils.c#L123-L139

PR #9523, opened 2023-11-06
hmem/synapseai: Refine the error handling and warning

shijin-aws:
Currently, synapseai_init return FI_ENODATA when there is no synapseai device on the platform. This will cause warnings are printed on any non-synapseai platform. This patch improves this by making synapseai_init return ENOSYS when all the API calls succeeded but zero devices are detected.

### DIFF
--- a/src/hmem_synapseai.c
+++ b/src/hmem_synapseai.c
@@ -108,20 +108,32 @@ int synapseai_init(void)
 
 	err = synapseai_dl_init();
 	if (err)
-		return -FI_ENODATA;
+		return err;
 
 	status = synapseai_ops.synInitialize();
-	if (status != synSuccess)
-		return -FI_ENODATA;
+	if (status != synSuccess) {
+		FI_WARN(&core_prov, FI_LOG_CORE,
+			"synInitialize failed: %d\n", status);
+		return -FI_EIO;
+	}
 
 	status = synapseai_ops.synDeviceGetCount(&device_count);
-	if (status != synSuccess || device_count == 0)
-		/*
-		 * TODO We should call destroy here to free resources allocated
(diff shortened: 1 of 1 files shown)


src/hmem_synapseai.c
```diff
@@ -108,20 +108,31 @@ int synapseai_init(void)
...
 	status = synapseai_ops.synDeviceGetCount(&device_count);
-	if (status != synSuccess || device_count == 0)
-		/*
-		 * TODO We should call destroy here to free resources allocated
-		 * in initialize, but the destroy call hangs on instances without
-		 * a habana device
-		 */
-		return -FI_ENODATA;
+
+	/*
```
a-szegel, line 122: We want to call synDestroy if `status != synSuccess` or `status != synSuccess`.  By separating these if statements, you make it more likely for us to miss this in the future.  I understand you did it b/c you want to return different error codes for each case, plz add a TODO comment for each block.

shijin-aws: We should destroy unconditionally after synGetDeviceCount. That is the point. I can update the text to make it more clear

j-xiong:
> when there is synapseai device on the platform

Is a `no` missing here?

shijin-aws:
> > when there is synapseai device on the platform
> 
> Is a `no` missing here?

Yes, there is a typo, will update the commit

src/hmem_synapseai.c
j-xiong: Something like this should work: `Starting from here we should call destroy before returning error to free ....`

shijin-aws: Updated

j-xiong (approved):

PR #9525, opened 2023-11-07
prov/efa: Introduce efa specific domain operations

shijin-aws:
Make efa onboard the fi_open_ops API, which allows user to access efa specific domain ops. The usage of this API is documented in fi_efa.7.md.

Also added a unit test.

### DIFF
--- a/man/fi_efa.7.md
+++ b/man/fi_efa.7.md
@@ -134,6 +134,69 @@ provider for AWS Neuron or Habana SynapseAI.
   delivered to the target buffer only once. If endpoint is not able to support
   this feature, it will return -FI_EOPNOTSUPP for the call to fi_setopt().
 
+# PROVIDER SPECIFIC DOMAIN OPS
+The efa provider exports extensions for operations
+that are not provided by the standard libfabric interface. These extensions
+are available via the "`fi_ext_efa.h`" header file.
+
+## Domain Operation Extension
+
+Domain operation extension is obtained by calling `fi_open_ops`
+(see [`fi_domain(3)`](fi_domain.3.html))
+```c
+int fi_open_ops(struct fid *domain, const char *name, uint64_t flags,
+    void **ops, void *context);
+```
+and requesting `FI_EFA_DOMAIN_OPS` in `name`. `fi_open_ops` returns `ops` as
+the pointer to the function table `fi_efa_ops_domain` defined as follows:
+
+```c
+struct fi_efa_ops_domain {
+	int (*query_mr)(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr);
+};
+```
+
+It contains the following operations
+
+### query_mr
+This op query an existing memory registration as input, and outputs the efa
+specific mr attribute which is defined as follows
+
+```c
+struct fi_efa_mr_attr {
+    uint16_t ic_id_validity;
+    uint16_t recv_ic_id;
+    uint16_t rdma_read_ic_id;
+    uint16_t rdma_recv_ic_id;
+};
+```
+
+*ic_id_validity*
+:	Validity mask of interconnect id fields. Currently the following bits are supported in the mask:
+
+	FI_EFA_MR_ATTR_RECV_IC_ID:
+		recv_ic_id has a valid value.
+
+	FI_EFA_MR_ATTR_RDMA_READ_IC_ID:
+		rdma_read_ic_id has a valid value.
+
+	FI_EFA_MR_ATTR_RDMA_RECV_IC_ID:
+		rdma_recv_ic_id has a valid value.
+
+*recv_ic_id*
+:	Physical interconnect used by the device to reach the MR for receive operation. It is only valid when `ic_id_validity` has the `FI_EFA_MR_ATTR_RECV_IC_ID` bit.
+
+*rdma_read_ic_id*
+:	Physical interconnect used by the device to reach the MR for RDMA read operation. It is only valid when `ic_id_validity` has the `FI_EFA_MR_ATTR_RDMA_READ_IC_ID` bit.
+
+*rdma_recv_ic_id*
+:	Physical interconnect used by the device to reach the MR for RDMA write receive. It is only valid when `ic_id_validity` has the `FI_EFA_MR_ATTR_RDMA_RECV_IC_ID` bit.
+
+#### Return value
+**query_mr()** returns 0 on success, or the value of errno on failure
+(which indicates the failure reason).
+
+
 # RUNTIME PARAMETERS
 
 *FI_EFA_TX_SIZE*
--- a/prov/efa/src/fi_ext_efa.h
+++ b/prov/efa/src/fi_ext_efa.h
@@ -1,0 +1,26 @@
+/* Copyright Amazon.com, Inc. or its affiliates. All rights reserved. */
+/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
+
+#ifndef _FI_EXT_EFA_H_
+#define _FI_EXT_EFA_H_
+
+#define FI_EFA_DOMAIN_OPS "efa domain ops"
+
+struct fi_efa_mr_attr {
+    uint16_t ic_id_validity;
+    uint16_t recv_ic_id;
+    uint16_t rdma_read_ic_id;
+    uint16_t rdma_recv_ic_id;
+};
+
+enum {
+    FI_EFA_MR_ATTR_RECV_IC_ID = 1 << 0,
+    FI_EFA_MR_ATTR_RDMA_READ_IC_ID = 1 << 1,
+    FI_EFA_MR_ATTR_RDMA_RECV_IC_ID = 1 << 2,
+};
+
+struct fi_efa_ops_domain {
+	int (*query_mr)(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr);
+};
+
+#endif /* _FI_EXT_EFA_H_ */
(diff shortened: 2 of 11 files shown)


prov/efa/src/fi_ext_efa.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef _FI_EXT_EFA_H_
+#define _FI_EXT_EFA_H_
+
+#define FI_EFA_DOMAIN_OPS_1 "efa domain ops 1"
+
+struct fi_efa_mr_attr {
+    /* The pci bus id of the rx performed into the mr */
+    uint32_t pci_bus_id;
+};
```
wenduwan, line 41: We can reuse https://github.com/ofiwg/libfabric/blob/f5decb0558865275d2acd16cebc53056409698c0/include/rdma/fabric.h#L497

shijin-aws:
This PR needs more evaluation and we are not targeting 1.20.x for it

man/fi_efa.7.md
```diff
@@ -134,6 +134,43 @@ provider for AWS Neuron or Habana SynapseAI.
...
+### query_mr
+This op query an existing memory registration (mr), and returns the efa
+specific mr attribute which is defined as follows
+
+```c
+struct fi_efa_mr_attr {
+    uint32_t pci_bus_id;
+};
+```
+. The `pci_bus_id` represents the pci bus id of the rx (receive, remote-write etc.)
```
bwbarrett, line 170: This needs more description.  I'm not actually sure what this means.  I'm also not sure how to think about it without also knowing where CQs get emitted.

shijin-aws: @mrgolin mentioned CQ is emmited at 0 by default.. If the user doesn't know this info, I could change this to a boolean that indicates whether rx and cqe arrives in the same pcie port

man/fi_efa.7.md
```diff
@@ -134,6 +134,77 @@ provider for AWS Neuron or Habana SynapseAI.
   delivered to the target buffer only once. If endpoint is not able to support
   this feature, it will return -FI_EOPNOTSUPP for the call to fi_setopt().
 
+# PROVIDER SPECIFIC DOMAIN OPS
+The efa provider exports extensions for additional operations
```
rajachan, line 138: Nit: _additional operations_ -> _operations_

man/fi_efa.7.md
```diff
@@ -134,6 +134,77 @@ provider for AWS Neuron or Habana SynapseAI.
...
+struct fi_efa_ops_domain {
+	int (*query_mr)(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr);
+};
+```
+
+It contains the following operations
+
+### query_mr
+This op query an existing memory registration (mr) as input, and outputs the efa
+specific mr attribute which is defined as follows
```
rajachan, line 163: Nit: _mr_ -> _MR_

prov/efa/test/efa_unit_test_domain.c
```diff
@@ -0,0 +1,150 @@
...
+    skip();
+}
+
+/* run mr query for both recv and rdma read's pci bus id */
+void test_efa_domain_open_ops_mr_query_recv_and_rdma_read(struct efa_resource **state)
+{
+    skip();
+}
+
+#endif /* HAVE_EFADV_QUERY_MR */
```
rajachan, line 150: Is this the general pattern we follow with our unit tests? We are mocking out the implementation of our dependencies so we can unit test a particular interface. Why skip these tests if configure does not find the capability in rdma-core?

shijin-aws: @rajachan To mock the rdma-core verbs we need the symbols of `efadv_query_mr` `efadv_mr_attr` and `EFADV**`, which aren't available when they are not found in rdma-core headers? How can we not skip them in this case?

shijin-aws: >Is this the general pattern we follow with our unit tests?

Yes, we have done the same pattern to unit test something that is only introduced in newer rdma-core

shijin-aws: > In other words, I would expect to have a test here and see an expected error ENOSYS, rather than skipping.

I get this point, will add it

rajachan: That's the whole point of a mock. What behavior do you expect apps to see when they call into the open_ops API with the EFA provider built against a version of rdma-core that does not have that definition? ENOSYS? Then we need a test to check that behavior.

shijin-aws: Updated

prov/efa/src/fi_ext_efa.h
```diff
@@ -0,0 +1,55 @@
...
+    uint16_t pci_bus_id_validity;
+    uint16_t recv_pci_bus_id;
+    uint16_t rdma_read_pci_bus_id;
+    uint16_t rdma_recv_pci_bus_id;
+};
+
+enum {
+    FI_EFA_MR_ATTR_VALIDITY_RECV_PCI_BUS_ID = 1 << 0,
+    FI_EFA_MR_ATTR_VALIDITY_RDMA_READ_PCI_BUS_ID = 1 << 1,
+    FI_EFA_MR_ATTR_VALIDITY_RDMA_RECV_PCI_BUS_ID = 1 << 2,
```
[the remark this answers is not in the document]
shijin-aws: dropped

man/fi_efa.7.md
shijin-aws: updated

man/fi_efa.7.md
shijin-aws: updated

prov/efa/src/fi_ext_efa.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
```
darrylabbate, line 31: We can use the short SPDX headers in place of the license text

```c
/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
```

man/fi_efa.7.md
```diff
@@ -134,6 +134,77 @@ provider for AWS Neuron or Habana SynapseAI.
...
+    uint16_t rdma_recv_pci_bus_id;
+};
+```
+
+*pci_bus_id_validity*
+:	Validity mask of PCI bus id fields. Currently the following bits are supported in the mask:
+
+```c
+enum {
+    FI_EFA_MR_ATTR_RECV_PCI_BUS_ID = 1 << 0,
```
bwbarrett, line 179: Why document the actual values?  Just document that there are said constants.

bwbarrett (commented):
I didn't review the code for correctness, so I'm not going to approve the request, but the API looks like what we expect.  Thanks!

man/fi_efa.7.md
```diff
@@ -134,6 +134,77 @@ provider for AWS Neuron or Habana SynapseAI.
...
 
+# PROVIDER SPECIFIC DOMAIN OPS
+The efa provider exports extensions for operations
+that are not provided by the standard libfabric interface. These extensions
+are available via the "`fi_ext_efa.h`" header file.
+
+## Domain Operation Extension
+
+Domain operation extension is obtained by calling `fi_open_ops`
+(see [`fi_domain`(3)](fi_domain.3.html))
```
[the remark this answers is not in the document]
shijin-aws: updated

man/fi_efa.7.md
```diff
@@ -134,6 +134,77 @@ provider for AWS Neuron or Habana SynapseAI.
...
+```c
+struct fi_efa_ops_domain {
+	int (*query_mr)(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr);
+};
+```
+
+It contains the following operations
+
+### query_mr
+This op query an existing memory registration (MR) as input, and outputs the efa
```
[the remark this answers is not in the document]
shijin-aws: I will just drop it

man/fi_efa.7.md
shijin-aws: Thanks, updated.

rajachan (approved):
Thanks for adding the fallback test. Unrelated to this change, but please track a work item internally to have a similar expect-error test for other fallback paths instead of skip()

shijin-aws:
Updated the PR to accommodate the rdma-core API renaming https://github.com/linux-rdma/rdma-core/pull/1412 (merged)

prov/psm3/psm3/hal_sockets/sockets_hal.c
```diff
@@ -235,10 +235,10 @@ static int psm3_hfp_sockets_get_default_pkey(void)
 	return 0;	/* use slot 0 as default */
 }
 
-static int psm3_hfp_sockets_get_unit_pci_bus(int unit, uint32_t *domain,
+static int psm3_hfp_sockets_get_unit_ic(int unit, uint32_t *domain,
```
mrgolin, line 238: Looks like you accidently renamed another provider symbols.

PR #9534, opened 2023-11-09
prov/efa: Do not abort on all deprecated env vars

a-szegel:
Log an info statement and continue on deprecated env variables that are low impact to the user.

### DIFF
--- a/prov/efa/src/efa_env.c
+++ b/prov/efa/src/efa_env.c
@@ -103,15 +103,24 @@ void efa_env_param_get(void)
 	 * Therefore, its cap must be < INT_MAX/2 too.
 	 */
 	size_t max_rnr_backoff_wait_time_cap = INT_MAX/2 - 1;
+	char *abort_deprecated_env_vars[] = {"FI_EFA_MTU_SIZE", "FI_EFA_TX_IOV_LIMIT", "FI_EFA_RX_IOV_LIMIT"};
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_env.c
```diff
@@ -103,15 +103,24 @@ void efa_env_param_get(void)
...
+		if (getenv(abort_deprecated_env_vars[i])) {
+			fprintf(stderr, "%s env variable detected! The use of this variable has been deprecated "
+					"and as such execution cannot proceed.\n", abort_deprecated_env_vars[i]);
+			abort();
+		};
+	}
+
+	for (i = 0; i < sizeof(info_deprecated_env_vars) / sizeof(info_deprecated_env_vars[0]); i++) {
+		if (getenv(info_deprecated_env_vars[i])) {
+			EFA_INFO(FI_LOG_CORE, "%s env variable detected! The use of this variable "
```
sunkuamzn, line 120: I think this should be a warning

wenduwan (approved):
Good stuff. Thank you!

sunkuamzn (approved):

prov/efa/src/efa_env.c
shijin-aws: Brian would like to even not warn when FI_EFA_SET_CUDA_SYNC_MEMOPS=0, which will the make the logic more complicated

shijin-aws: This revision doesn't do the exactly same thing, but at least won't puke customers who were told to set it to 0 to mitigate the earlier libfabric

shijin-aws (approved):

PR #9543, opened 2023-11-13
prov/verbs: Non blocking EP creation (asynchronous route resolution using event channel)

### LINKED ISSUE #9494: prov/verbs: EP creation blocks during RDMA address resolution
When a new endpoint is created with `fi_endpoint()`, the verbs provider blocks the caller while the destination address is being resolved (i.e., `rdma_resolve_addr()`). If the destination address is not reachable, the caller has to wait 2 seconds for the timeout provided to `rdma_resolve_addr()` to expire. This is a killer for scalability.

Here is the implementation for `vrb_create_ep()`: https://github.com/ofiwg/libfabric/blob/c5cace8361197ad7d9e449eea87f8dd8c93199f9/prov/verbs/src/verbs_init.c#L338
```
	/* TODO convert this call to non-blocking (use event channel) as well:
	 * This may likely be needed for better scaling when running large
	 * MPI jobs.
	 * Making this non-blocking would mean we can't create QP at EP enable
	 * time. We need to wait for RDMA_CM_EVENT_ADDR_RESOLVED event before
	 * creating the QP using rdma_create_qp. It would also require a SW
	 * receive queue to store recvs posted by app after enabling the EP.
	 */
	if (rdma_resolve_addr(*id, rai->ai_src_addr, rai->ai_dst_addr,
			      VERBS_RESOLVE_TIMEOUT)) {
		ret = -errno;
		VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_resolve_addr");
		ofi_straddr_log(&vrb_prov, FI_LOG_WARN, FI_LOG_EP_CTRL,
				"src addr", rai->ai_src_addr);
		ofi_straddr_log(&vrb_prov, FI_LOG_WARN, FI_LOG_EP_CTRL,
				"dst addr", rai->ai_dst_addr);
		goto err2;
	}
```

The solution is described in the TODO comment above: `rdma_resolve_addr()` should use an event channel, so it doesn't block the caller.

**Describe alternatives you've considered**
AFAIK there is no alternative. 

**Additional context**
More context can be provided if needed

sydidelot:
The PR reworks the verbs provider to avoid blocking the caller in `rdma_resolve_route()`. 
The solution I implemented is based on the following TODO from `vrb_create_ep()`:
```
       /* TODO convert this call to non-blocking (use event channel) as well:
        * This may likely be needed for better scaling when running large
        * MPI jobs.
        * Making this non-blocking would mean we can't create QP at EP enable
        * time. We need to wait for RDMA_CM_EVENT_ADDR_RESOLVED event before
        * creating the QP using rdma_create_qp. It would also require a SW
        * receive queue to store recvs posted by app after enabling the EP.
        */
```

Fixes #9494

### DIFF
--- a/prov/verbs/src/verbs_cm.c
+++ b/prov/verbs/src/verbs_cm.c
@@ -193,6 +193,17 @@ vrb_msg_ep_connect(struct fid_ep *ep_fid, const void *addr,
 	if (ep->srx)
 		ep->conn_param.srq = 1;
 
+	if (addr) {
+		free(ep->info_attr.dest_addr);
+		ep->info_attr.dest_addr = mem_dup(addr, ofi_sizeofaddr(addr));
+		if (!ep->info_attr.dest_addr) {
+			free(ep->cm_priv_data);
+			ep->cm_priv_data = NULL;
+			return -FI_ENOMEM;
+		}
+		ep->info_attr.dest_addrlen = ofi_sizeofaddr(addr);
+	}
+
 	ofi_genlock_lock(&vrb_ep2_progress(ep)->ep_lock);
 	assert(ep->state == VRB_IDLE);
 	ep->state = VRB_RESOLVE_ROUTE;
--- a/prov/verbs/src/verbs_cm.c
+++ b/prov/verbs/src/verbs_cm.c
@@ -133,7 +133,7 @@ vrb_ep_prepare_rdma_cm_param(struct rdma_conn_param *conn_param,
 	conn_param->rnr_retry_count = 7;
 }
 
-static void
+void
 vrb_msg_ep_prepare_rdma_cm_hdr(void *priv_data,
 				const struct rdma_cm_id *id)
 {
@@ -159,8 +159,7 @@ vrb_msg_ep_connect(struct fid_ep *ep_fid, const void *addr,
 	struct vrb_ep *ep = container_of(ep_fid, struct vrb_ep, util_ep.ep_fid);
 	size_t priv_data_len;
 	struct vrb_cm_data_hdr *cm_hdr;
-	off_t rdma_cm_hdr_len = 0;
-	int ret;
+	int ret = 0;
 
 	if (OFI_UNLIKELY(paramlen > VERBS_CM_DATA_SIZE))
 		return -FI_EINVAL;
@@ -173,18 +172,12 @@ vrb_msg_ep_connect(struct fid_ep *ep_fid, const void *addr,
 		}
 	}
 
-	if (ep->id->route.addr.src_addr.sa_family == AF_IB)
-		rdma_cm_hdr_len = sizeof(struct vrb_rdma_cm_hdr);
-
-	priv_data_len = sizeof(*cm_hdr) + paramlen + rdma_cm_hdr_len;
+	priv_data_len = sizeof(*cm_hdr) + paramlen + sizeof(struct vrb_rdma_cm_hdr);
 	ep->cm_priv_data = malloc(priv_data_len);
 	if (!ep->cm_priv_data)
 		return -FI_ENOMEM;
 
-	if (rdma_cm_hdr_len)
-		vrb_msg_ep_prepare_rdma_cm_hdr(ep->cm_priv_data, ep->id);
-
-	cm_hdr = (void *)((char *)ep->cm_priv_data + rdma_cm_hdr_len);
+	cm_hdr = (void *)((char *)ep->cm_priv_data + sizeof(struct vrb_rdma_cm_hdr));
 	vrb_msg_ep_prepare_cm_data(param, paramlen, cm_hdr);
 	vrb_ep_prepare_rdma_cm_param(&ep->conn_param, ep->cm_priv_data,
 					priv_data_len);
--- a/prov/verbs/src/verbs_cm.c
+++ b/prov/verbs/src/verbs_cm.c
@@ -199,11 +199,15 @@ vrb_msg_ep_connect(struct fid_ep *ep_fid, const void *addr,
 
 	ofi_genlock_lock(&vrb_ep2_progress(ep)->ep_lock);
 	assert(ep->state == VRB_IDLE);
-	ep->state = VRB_RESOLVE_ROUTE;
-	ret = rdma_resolve_route(ep->id, VERBS_RESOLVE_TIMEOUT);
-	if (ret) {
+	ep->state = VRB_RESOLVE_ADDR;
+	if (rdma_resolve_addr(ep->id, ep->info_attr.src_addr,
+			      ep->info_attr.dest_addr, VERBS_RESOLVE_TIMEOUT)) {
 		ret = -errno;
-		VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_resolve_route");
+		VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_resolve_addr");
+		ofi_straddr_log(&vrb_prov, FI_LOG_WARN, FI_LOG_EP_CTRL,
+				"src addr", ep->info_attr.src_addr);
+		ofi_straddr_log(&vrb_prov, FI_LOG_WARN, FI_LOG_EP_CTRL,
+				"dst addr", ep->info_attr.dest_addr);
 		free(ep->cm_priv_data);
 		ep->cm_priv_data = NULL;
 		ep->state = VRB_IDLE;
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -624,8 +624,17 @@ int vrb_init_progress(struct vrb_progress *progress, struct fi_info *info)
 	if (ret)
 		goto err1;
 
+	ret = ofi_bufpool_create(&progress->recv_wr_pool,
+				 sizeof(struct vrb_recv_wr) +
+				 info->rx_attr->iov_limit * sizeof(struct ibv_sge),
+				 16, 0, 1024, OFI_BUFPOOL_NO_TRACK);
+	if (ret)
+		goto err2;
+
 	return 0;
 
+err2:
+	ofi_bufpool_destroy(progress->ctx_pool);
 err1:
 	ofi_genlock_destroy(&progress->ep_lock);
 	return ret;
@@ -633,6 +642,7 @@ err1:
 
 void vrb_close_progress(struct vrb_progress *progress)
 {
+	ofi_bufpool_destroy(progress->recv_wr_pool);
 	ofi_bufpool_destroy(progress->ctx_pool);
 	ofi_genlock_destroy(&progress->ep_lock);
 }
--- a/prov/verbs/src/verbs_cq.c
+++ b/prov/verbs/src/verbs_cq.c
@@ -608,7 +608,7 @@ err1:
 }
 
 
-int vrb_init_progress(struct vrb_progress *progress, struct ibv_context *verbs)
+int vrb_init_progress(struct vrb_progress *progress, struct fi_info *info)
 {
 	int ret;
 
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -1033,15 +1033,19 @@ static int vrb_ep_enable(struct fid_ep *ep_fid)
 			return -FI_EINVAL;
 		}
 
-		ret = rdma_create_qp(ep->id, domain->pd, &attr);
-		if (ret) {
-			VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_create_qp");
-			return -errno;
-		}
+		/* Server-side QP creation, after RDMA_CM_EVENT_CONNECT_REQUEST
+		 * is recevied */
+		if (ep->id->verbs && ep->ibv_qp == NULL) {
+			ret = rdma_create_qp(ep->id, domain->pd, &attr);
+			if (ret) {
+				VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_create_qp");
+				return -errno;
+			}
 
-		/* Allow shared XRC INI QP not controlled by RDMA CM
-		 * to share same post functions as RC QP. */
-		ep->ibv_qp = ep->id->qp;
+			/* Allow shared XRC INI QP not controlled by RDMA CM
+			 * to share same post functions as RC QP. */
+			ep->ibv_qp = ep->id->qp;
+		}
 		break;
 	case FI_EP_DGRAM:
 		assert(domain);
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -57,19 +57,18 @@ void vrb_add_credits(struct fid_ep *ep_fid, uint64_t credits)
 	ofi_genlock_unlock(&vrb_ep2_progress(ep)->ep_lock);
 }
 
-ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
+int vrb_post_recv_internal(struct vrb_ep *ep, struct ibv_recv_wr *wr)
 {
 	struct vrb_context *ctx;
 	struct ibv_recv_wr *bad_wr;
 	uint64_t credits_to_give;
 	int ret, err;
 
-	ofi_genlock_lock(&vrb_ep2_progress(ep)->ep_lock);
+	assert(ofi_genlock_held(&vrb_ep2_progress(ep)->ep_lock));
+
 	ctx = vrb_alloc_ctx(vrb_ep2_progress(ep));
-	if (!ctx) {
-		ret = -FI_EAGAIN;
-		goto unlock;
-	}
+	if (!ctx)
+		return -FI_EAGAIN;
 
 	ctx->ep = ep;
 	ctx->user_ctx = (void *) (uintptr_t) wr->wr_id;
@@ -80,8 +79,7 @@ ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
 	wr->wr_id = (uintptr_t) ctx->user_ctx;
 	if (ret) {
 		vrb_free_ctx(vrb_ep2_progress(ep), ctx);
-		ret = -FI_EAGAIN;
-		goto unlock;
+		return -FI_EAGAIN;
 	}
 
 	slist_insert_tail(&ctx->entry, &ep->rq_list);
@@ -109,7 +107,45 @@ ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
 			ep->rq_credits_avail += credits_to_give;
 	}
 
-unlock:
+	return ret;
+}
+
+static int vrb_prepost_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
+{
+	struct vrb_recv_wr *save_wr;
+	size_t i;
+
+	assert(ofi_genlock_held(&vrb_ep2_progress(ep)->ep_lock));
+
+	if (wr->next)
+		return -FI_EINVAL;
+
+	save_wr = vrb_alloc_recv_wr(vrb_ep2_progress(ep));
+	if (!save_wr)
+		return -FI_ENOMEM;
+
+	save_wr->wr.wr_id = wr->wr_id;
+	save_wr->wr.next = NULL;
+	save_wr->wr.num_sge = wr->num_sge;
+	for (i = 0; i < wr->num_sge; i++)
+		save_wr->sge[i] = wr->sg_list[i];
+	save_wr->wr.sg_list = save_wr->sge;
+	slist_insert_tail(&save_wr->entry, &ep->prepost_wr_list);
+	return 0;
+}
+
+ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
+{
+	int ret;
+
+	if (wr->num_sge > ep->info_attr.rx_iov_limit)
+		return -FI_EINVAL;
+
+	ofi_genlock_lock(&vrb_ep2_progress(ep)->ep_lock);
+	if (!ep->ibv_qp)
+		ret = vrb_prepost_recv(ep, wr);
+	else
+		ret = vrb_post_recv_internal(ep, wr);
 	ofi_genlock_unlock(&vrb_ep2_progress(ep)->ep_lock);
 	return ret;
 }
@@ -437,6 +473,7 @@ vrb_alloc_init_ep(struct fi_info *info, struct vrb_domain *domain,
 
 	slist_init(&ep->sq_list);
 	slist_init(&ep->rq_list);
+	slist_init(&ep->prepost_wr_list);
 	ep->util_ep.ep_fid.msg = calloc(1, sizeof(*ep->util_ep.ep_fid.msg));
 	if (!ep->util_ep.ep_fid.msg)
 		goto err3;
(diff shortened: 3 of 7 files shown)


sydidelot:
@j-xiong, @chien-intel it seems that you are the most active developers for the verbs provider :)
May I request your feedback on this patch? Thank you!

prov/verbs/src/verbs_cm.c
```diff
@@ -190,11 +190,15 @@ vrb_msg_ep_connect(struct fid_ep *ep_fid, const void *addr,
 
 	ofi_genlock_lock(&vrb_ep2_progress(ep)->ep_lock);
 	assert(ep->state == VRB_IDLE);
-	ep->state = VRB_RESOLVE_ROUTE;
-	ret = rdma_resolve_route(ep->id, VERBS_RESOLVE_TIMEOUT);
-	if (ret) {
+	ep->state = VRB_RESOLVE_ADDR;
+	if (rdma_resolve_addr(ep->id, ep->info_attr.src_addr,
+			      ep->info_attr.dest_addr, VERBS_RESOLVE_TIMEOUT)) {
```
j-xiong, line 195: What's the pros and cons calling `rdma_resovle_addr` here vs when the endpoint is enabled? The EQ should have already been bound to the endpoint when the endpoint is enabled.

prov/verbs/src/verbs_ep.c
```diff
@@ -109,7 +108,44 @@ ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
...
+{
+	struct vrb_recv_wr *save_wr;
+	size_t i;
+
+	assert(ofi_genlock_held(&vrb_ep2_progress(ep)->ep_lock));
+
+	if (wr->num_sge > vrb_gl_data.def_tx_iov_limit)
+		return -FI_EINVAL;
+	if (wr->next)
+		return -FI_EINVAL;
```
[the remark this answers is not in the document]
sydidelot: The verbs provider does not use chained wrs. If multiple recv buffers are posted at the same time, they would all be part of the same `ibv_recv_wr::sg_list` in the same wr. 
Adding support for chained wrs will add complexity to the code and won't be used by the verbs provider anyway. Do we really want this change?

prov/verbs/src/verbs_cm.c
sydidelot: I believe that calling `rdma_resolve_addr()` when the endpoint is enabled would add more complexity to the code: `fi_connect()` would have to wait for the event `RDMA_CM_EVENT_ADDR_RESOLVED` before it can resume the connection establishment process (rdma_resolve_route() + rdma_connect()).
Having all the CM logic inside the EQ makes the code cleaner IMHO.

prov/verbs/src/verbs_init.c
```diff
@@ -295,17 +295,20 @@ int vrb_get_rai_id(const char *node, const char *service, uint64_t flags,
...
-				(*rai)->ai_dst_addr, VERBS_RESOLVE_TIMEOUT);
-	if (ret) {
-		VRB_WARN_ERRNO(FI_LOG_FABRIC, "rdma_resolve_addr");
-		ofi_straddr_log(&vrb_prov, FI_LOG_INFO, FI_LOG_FABRIC,
-				"src addr", (*rai)->ai_src_addr);
-		ofi_straddr_log(&vrb_prov, FI_LOG_INFO, FI_LOG_FABRIC,
-				"dst addr", (*rai)->ai_dst_addr);
-		ret = -errno;
-		goto err2;
+	if (node) {
```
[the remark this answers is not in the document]
sydidelot: I tested your suggestion but it doesn't work as expected: the problem is that `(*rai)->ai_dst_addr` is set even if no hints are provided. I'd assume it's the behavior of `rdma_getaddrinfo()` to set `ai_dst_addr` when the `node` is given.
If `dest_addr` is supplied, I would expect the user application to give NULL for the `node` argument.

prov/verbs/src/verbs_cq.c
```diff
@@ -624,15 +624,25 @@ int vrb_init_progress(struct vrb_progress *progress, struct ibv_context *verbs)
 	if (ret)
 		goto err1;
 
+	ret = ofi_bufpool_create(&progress->recv_wr_pool,
+				 sizeof(struct vrb_recv_wr) +
+				 vrb_gl_data.def_rx_iov_limit * sizeof(struct ibv_sge),
+				 16, 0, 1024, OFI_BUFPOOL_NO_TRACK);
+	if (ret)
+		goto err2;
```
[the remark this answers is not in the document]
sydidelot: `err2` destroys the pool `progress->ctx_pool` created line 622, not the one line 627 creates.

sydidelot:
I added an extra patch to the series. I believe there are some core paths where connection establishment may fail and the provider doesn't report the failure to the EQ.

prov/verbs/src/verbs_cq.c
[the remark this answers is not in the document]
j-xiong: ahh, my mistake.

prov/verbs/src/verbs_ep.c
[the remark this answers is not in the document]
j-xiong: That's fair. No need to change.

sydidelot:
@shijin-aws Sorry to ping you directly. Can you please take a look at the AWS CI results and tell me if the error relates to my changes? 
It's a pity that all CI pipelines are now private an there is no output when a test fail... UCX is so much better in this aspect.

prov/verbs/src/verbs_ep.c
```diff
@@ -109,7 +108,44 @@ ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
...
+
+static int vrb_prepost_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
+{
+	struct vrb_recv_wr *save_wr;
+	size_t i;
+
+	assert(ofi_genlock_held(&vrb_ep2_progress(ep)->ep_lock));
+
+	if (wr->num_sge > vrb_gl_data.def_tx_iov_limit)
+		return -FI_EINVAL;
```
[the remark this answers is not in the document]
chien-intel: Don't know what the future holds.  Placing the check in vrb_post_recv would be most generic but with current code base the best place is vrb_msg_ep_recvmsg.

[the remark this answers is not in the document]
sydidelot: `vrb_post_recv()` is also called from `vrb_dgram_ep_recvmsg()`.
I think it seems better to have the check directly in `vrb_post_recv()` as it's more generic (as you said).

prov/verbs/src/verbs_ep.c
```diff
@@ -966,15 +1034,17 @@ static int vrb_ep_enable(struct fid_ep *ep_fid)
 			return -FI_EINVAL;
 		}
 
-		ret = rdma_create_qp(ep->id, domain->pd, &attr);
-		if (ret) {
-			VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_create_qp");
-			return -errno;
-		}
+		if (ep->state == VRB_REQ_RCVD) {
```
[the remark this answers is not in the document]
sydidelot: `vrb_eq_addr_resolved_event()` only creates the QP for the connection initiator. The target side needs to create the QP when the EP is enabled, after `RDMA_CM_EVENT_CONNECT_REQUEST` is received (EP state is VRB_REQ_RCVD in this case). 
To be honest, I don't like this check neither. Do you have a better suggestion? Is there a way to determine what side enables the EP, for example based on the content of ep->id?

prov/verbs/src/verbs_ep.c
```diff
@@ -581,6 +646,7 @@ static int vrb_ep_close(fid_t fid)
 		vrb_cleanup_cq(ep);
 		vrb_flush_sq(ep);
 		vrb_flush_rq(ep);
+		vrb_flush_prepost_wr(ep);
```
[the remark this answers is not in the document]
sydidelot: Your suggestion makes sense: we should always process WR in order. 
I have inverted the 2 lines `vrb_flush_prepost_wr` and `vrb_flush_rq`

prov/verbs/src/verbs_ep.c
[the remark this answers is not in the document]
sydidelot: I changed the check and the code now uses `ep->id->verbs` to figure out what side (`server` or `client`) executes the function. I looked at rdma-core and it seems that `ep->id->verbs` is set only when the addr or the route are resolved:
https://github.com/linux-rdma/rdma-core/blob/0cf342c301dfef1ba31c7ab763682ed0c59e7763/librdmacm/cma.c#L618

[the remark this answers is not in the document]
chien-intel: id->verbs field is ibv_context and its creation/assignment is not related to addr/route resolution.

Please consider creating QP in RDMA_CM_EVENT_CONNECT_REQUEST handler see https://github.com/linux-rdma/rdma-core/blob/master/librdmacm/man/rdma_cm.7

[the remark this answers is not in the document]
sydidelot: We cannot move QP creation in `RDMA_CM_EVENT_CONNECT_REQUEST` handler because `rdma_create_qp()` requires to set the QP init attributes (`ibv_qp_init_attr`), and these attributes are known only when the EP is enabled. The EP doesn't even exist when the event `RDMA_CM_EVENT_CONNECT_REQUEST` is received.

> id->verbs field is ibv_context and its creation/assignment is not related to addr/route resolution.

I agree with your statement, but in practice and according to the code, id->verbs set when the addr/route is resolved. 

An alternative solution to find out if the route was resolved would be to check if `rdma_cm_id::route::num_paths` is > 0 or `rdma_cm_id::route::path_rec` is != NULL.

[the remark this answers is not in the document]
chien-intel: I'm still running tests, trying out large MPI jobs.  It is taking a bit of time.

prov/verbs/src/verbs_init.c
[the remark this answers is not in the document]
shefty: Does this check work ```node || hints->dest_addr```?  This matches with @j-xiong comment.

I believe we would have already exited the call if FI_SOURCE was set.

prov/verbs/src/verbs_cm.c
shefty: Until rdma_resolve_addr() completes, isn't necessarily bound to a verbs device.  That's likely the con -- we don't have a device, don't have a qp, can't post receives...

prov/verbs/src/verbs_cm.c
```diff
@@ -193,6 +193,11 @@ vrb_msg_ep_connect(struct fid_ep *ep_fid, const void *addr,
 	if (ep->srx)
 		ep->conn_param.srq = 1;
 
+	if (addr) {
+		free(ep->info_attr.dest_addr);
+		ep->info_attr.dest_addr = mem_dup(addr, ofi_sizeofaddr(addr));
```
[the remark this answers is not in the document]
sydidelot: Thanks for the review and for catching this bug!

sydidelot:
@chien-intel @shefty @j-xiong While testing the changes, I found an issue when multiple verbs adapters are being used. The issue was found with RoCE but I believe Infiniband adapters are also impacted.
*Note*: I am aware of the issues when running multiple local interfaces in the same network subnet. This a different issue here.

When `fi_getinfo()` is executed with `hints->src_addr` but without `node` or `hints->dest_addr`,  the function may return domains that do not match the source address given in `hints->src_addr`. The bug doesn't reproduce without my changes.

I think the bug is in `vrb_set_info_addrs()`, which gets executed with the following call stack:
```
vrb_getinfo()
 vrb_get_match_infos()
  vrb_handle_sock_addr() /* id->verbs is NULL because `rdma_resolve_addr()` was not called */
   vrb_fill_addr() /* The condition rai->ai_src_addr && !ofi_is_loopback_addr(rai->ai_src_addr is true */
    vrb_set_info_addrs() 
 ```

The problem is that `vrb_set_info_addrs()` updates all the `info` structures with the same `rdma_addrinfo()` info (see function `vrb_rai_to_fi()`).
However, there is no guarantee that all the `info` structures listed there have the same source address as `rai->ai_src_addr`.
I believe we need to check if `iter->src_addr` matches `rai->ai_src_addr` in `vrb_set_info_addrs()` and remove the info from the list if so. 

The PR doesn't address this issue. I decided to work around the bug in the application: I skip all the info returned by `fi_getinfo()` that does not match the source address provided in the hints.

**Comment by chien-intel** [comment]
> @chien-intel @shefty @j-xiong While testing the changes, I found an issue when multiple verbs adapters are being used. The issue was found with RoCE but I believe Infiniband adapters are also impacted. _Note_: I am aware of the issues when running multiple local interfaces in the same network subnet. This a different issue here.

I did notice some issues when I included nodes with multiple infiniband adapters on same logical subnet.  I will create an issue for this when I have more info with easy reproducer.

**Comment by sydidelot** [comment]
@chien-intel it's likely the same issue as the one I described above.
Do you see warning messages similar to this one in your logs when you open an endpoint (active side)?
```verbs   domain   warn vrb_open_ep:1220 Invalid info->domain_attr->name: rocep65s0f0 and rocep193s0f0```

**Comment by sydidelot** [comment]
@chien-intel Any update for this PR?

File: prov/verbs/src/verbs_eq.c [hunk 9]
```diff
@@ -854,6 +854,68 @@ vrb_eq_xrc_disconnect_event(struct vrb_eq *eq,
...
+
+	if (ep->util_ep.type == FI_EP_MSG) {
+		vrb_msg_ep_get_qp_attr(ep, &attr);
+
+		if (rdma_create_qp(ep->id, vrb_ep2_domain(ep)->pd, &attr)) {
+			ep->state = VRB_DISCONNECTED;
+			ret = -errno;
+			VRB_WARN(FI_LOG_EP_CTRL,
+				 "rdma_create_qp failed: %d\n", -ret);
+			return ret;
```
chien-intel, line 879: what should we do with preposted receives in case of failure?

prov/verbs/src/verbs_ep.c
```diff
@@ -1034,15 +1034,19 @@ static int vrb_ep_enable(struct fid_ep *ep_fid)
...
 		}
 
-		ret = rdma_create_qp(ep->id, domain->pd, &attr);
-		if (ret) {
-			VRB_WARN_ERRNO(FI_LOG_EP_CTRL, "rdma_create_qp");
-			return -errno;
-		}
+		/* Server-side QP creation, after RDMA_CM_EVENT_CONNECT_REQUEST
+		 * is recevied */
+		if (ep->id->verbs) {
```
chien-intel, line 1039: I'm still having a hard time with this one.  Using internal variable from another library for this purpose is just a bad idea.  Unfortunately, I do not have an alternative for you.

sydidelot: The verbs provider already has a couple of existing conditions based on `id->verbs`. For example:
https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L1316
https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L1789

There is even this comment in the code:
```
	/* Handle the case when rdma_cm doesn't fill src address even
	 * though it fills the destination address (presence of id->verbs
	 * corresponds to a valid dest addr) */
```
See https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L1319C3-L1321C41

I'm not adding something new to the code.

chien-intel: I don't have that kind of insight into libibverbs nor librdmacm and I certainly would not code that way.  Just me.

shefty: id->verbs is externally visible.  It has to be that way to coordinate the use of librdmacm with libibverbs.  struct rdma_cm_id has both visible and hidden fields.  The hidden fields are not accessible to a user.

j-xiong: `struct rdma_cm_id` is part of the public API of RDMA CM. Referencing its member is fine when needed.

chien-intel: It is one thing to use the value, it is another to infer behavior.  Can we be certain when id->verbs is set, it is always after RDMA_CM_EVENT_CONNECT_REQUEST is received on the server side?

shefty: If we rephrase the question to: Is id->verbs always set for an RDMA_CM_EVENT_CONNECT_REQUEST?  The answer is yes.  Connect requests are received over a specific device.

Chien's question is subtly different, however, and I think the answer to his question is *no*, which might be exposing an issue.

An rdma_cm_id is associated with a verbs device in 3 cases.  On the client side, this is the purpose of rdma_resolve_addr().  Once that operation completes, we have a device.  On the server side, all connection requests are associated with a device.  The 3rd case, however, is through rdma_bind_addr().

The restructured connect() path calls enable() first, then rdma_resolve_addr().  If we're taking the client code path, then id->verbs is *usually* not set.  However, if we went through rdma_bind_addr() prior to calling connect(), it may be.  In that case, I think we would end up trying to create 2 QPs.

sydidelot: Sorry for the very, very late reply :)

I updated the code to check that ep->ibv_qp is NULL before creating the QP. The change is available here:
https://github.com/ofiwg/libfabric/pull/9543/files#diff-0c780e52c4b4880ab4a4012f359d1ae818bbdddcd650d9192f250f622e7bf927R1039

FYI, our product has been running with this PR for the last 2 months with no issue: plain Verbs (no RXM) and MSG endpoints. The only real issue we had was with the following patch that introduces EP states to the Verbs provider: https://github.com/ofiwg/libfabric/pull/9175
The above PR was causing assertion failures during connection shutdown because the EP state was incorrect. We simply reverted this patch and everything works now.

Anyway, I'm still convinced this PR is important for upstream as some runtime models require all the calls to be non-blocking (I believe this is the case for DAOS too because of the SPDK model). Please review it when you can.

prov/verbs/src/verbs_ofi.h
```diff
@@ -933,6 +934,12 @@ do {								\
 	( wr->opcode == IBV_WR_SEND || wr->opcode == IBV_WR_SEND_WITH_IMM	\
 	|| wr->opcode == IBV_WR_RDMA_WRITE_WITH_IMM )
 
+struct vrb_recv_wr {
+	struct slist_entry	entry;
```
shefty, line 943: nit: ibv_recv_wr has a next pointer which can be used to create a single-linked list.  Using that requires duplicating part of the slist implementation, so I'm not sure yet in the review if it's worth pursuing this change, as this is all non-critical path.

prov/verbs/src/verbs_ep.c
```diff
@@ -109,7 +108,45 @@ ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
...
+	slist_insert_tail(&save_wr->entry, &ep->prepost_wr_list);
+	return 0;
+}
+
+ssize_t vrb_post_recv(struct vrb_ep *ep, struct ibv_recv_wr *wr)
+{
+	int ret;
+
+	if (wr->num_sge > vrb_gl_data.def_rx_iov_limit)
+		return -FI_EINVAL;
```
shefty, line 143: I'd say just assert here, but can we get the limit that was associated with the ep?

prov/verbs/src/verbs_eq.c
```diff
@@ -884,6 +884,13 @@ vrb_eq_cm_process_event(struct vrb_eq *eq,
 		ofi_genlock_lock(&vrb_ep2_progress(ep)->ep_lock);
 		assert(ep->state == VRB_RESOLVE_ROUTE);
 		ep->state = VRB_CONNECTING;
+
+		vrb_msg_ep_prepare_rdma_cm_hdr(ep->cm_priv_data, ep->id);
```
shefty, line 963: Can this be part of an else statement to the if check below?

prov/verbs/src/verbs_eq.c
```diff
@@ -861,6 +861,68 @@ vrb_eq_xrc_disconnect_event(struct vrb_eq *eq,
...
+	}
+
+	assert(ep->ibv_qp);
+	while (!slist_empty(&ep->prepost_wr_list)) {
+		entry = slist_remove_head(&ep->prepost_wr_list);
+		wr = container_of(entry, struct vrb_recv_wr, entry);
+
+		ret = vrb_post_recv_internal(ep, &wr->wr);
+		if (ret && ep->util_ep.rx_cq) {
+			cq = container_of(ep->util_ep.rx_cq, struct vrb_cq, util_cq);
```
shefty, line 901: I think on an error we can just return here.  Whatever path the failure above at line 885 takes should also flush any remaining preposted receives.

prov/verbs/src/verbs_eq.c
```diff
@@ -861,6 +861,68 @@ vrb_eq_xrc_disconnect_event(struct vrb_eq *eq,
 	}
 }
 
+static inline int
```
shefty, line 864: nit: I'd drop the inline and let the compiler decide as this is a long function (and the compiler will decide anyway since we're in a .c file...)

shefty (commented):
Looks good -- I didn't see any actual problems with the changes.

prov/verbs/src/verbs_ep.c
sydidelot: I would prefer not assert here: `wr->num_sge` is set by the application when the receive buffer is posted: I think it's better to return `-FI_EINVAL` and let the application deal with this error. 

I slightly modified the code based on your suggestion to use the limit that is associated with the ep and I added an assertion failure to ensure that the EP limit is effectively lower than `vrb_gl_data.def_rx_iov_limit`
```
assert(ep->info_attr.rx_iov_limit <= vrb_gl_data.def_rx_iov_limit); 
if (wr->num_sge > ep->info_attr.rx_iov_limit)                       
    return -FI_EINVAL;                                              
```

prov/verbs/src/verbs_ofi.h
sydidelot: I need some time to think about it. I'm not sure the change is worth it.

prov/verbs/src/verbs_ep.c
shefty: I'm not sure the assert is always true.  The requested limit is subject to the device limits, not necessarily the default limit.

shefty (approved):
I think the assert that's marked needs to be removed, but otherwise, this looks good.  Thanks!

prov/verbs/src/verbs_ep.c
sydidelot: The problem is that the pool of WRs belongs to the progress object (shared between EPs of the same domain) and the max number of SGEs is set to `vrb_gl_data.def_rx_iov_limit`. Also, note that a similar assert() was added by the PR at line 68 (same file).
One possible fix would be to detect the max rx_iov_limit set by the HCA and use this value when the pool of WRs is created.

sydidelot (commented):
@shefty Thanks for the review! I did a few modifications related to the pool of WRs:
- The pool is now created with the max number of SGEs defined by `domain->info->rx_attr->iov_limit` instead of the default value set by the provider.
- I removed the 2 asserts that checked that the number of SGEs is lower than `vrb_gl_data.def_rx_iov_limit`. The code assumes that `ep->info_attr.rx_size` is always lower or equal to `domain->info->rx_attr->iov_limit`.

prov/verbs/src/verbs_ep.c
shefty: Hmm... I believe ibv_post_recv() checks num_sge.  So, I think we're okay moving this check into vrb_prepost_recv(), which is where we really need it.  I agree if we can use the device limit when initializing the pool, that should cover (unlikely) edge cases.  The check in vrb_prepost_recv() just need to verify that we don't overrun the local array.  When the receives are actually posted to the device, vrb_post_recv_internal() -> ibv_post_recv() will check that the QP can handle it.  A failure there is a user error, which at least will be handled (by tearing down the QP).  Does this work for you?

sydidelot:
Can someone please look at the CI logs and tell me if the errors are related to the PR?

j-xiong:
The CI job was cancelled in the middle probably due to infrastructure failure.

sydidelot:
> The CI job was cancelled in the middle probably due to infrastructure failure.
@j-xiong I see that you have re-triggered the job. Thanks!

zachdworkin:
It hit the same node failure. I am taking that node offline and will restart your job again. Sorry its taking so long

sydidelot:
> It hit the same node failure. I am taking that node offline and will restart your job again. Sorry its taking so long

No worry. I know from experience how difficult it is to keep a CI pipeline running :)

zachdworkin:
@sydidelot thanks for your understanding. We had some updates to our controller's server that look like they may have waterfalled some issues to our server that we are fixing.

sydidelot:
@zachdworkin I rebased the PR but it still fails. Please let me know when the CI infrastructure is fixed.

zachdworkin:
@sydidelot Will do. Keep an eye on #9898 which should contain a fix for the ze-shm-v3 stage that failed in your PR.

j-xiong:
The CI error is a known issue that has been fixed by #9898. I am going to ignore that and merge this PR.

PR #9548, opened 2023-11-13
github/actions: Add fabtests man page to the trigger condition

j-xiong:
Changes to fabtests man pages should also trigger the man page converter and man page updater.


j-xiong:
No need to wait for CI.

PR #9551, opened 2023-11-14
v1.20.0rc2

j-xiong:
Don't merge from the web GUI. Will be pushed directly to preserve the hash value (together with a tag).


j-xiong:
@ofiwg/aws-efa-team I populated the change list with results from `git log` and made some minor adjustment such as consolidating some bug fixes into a single entry. Please check the `EFA` section of `NEWS.md` to see if you want it be written differently.

shijin-aws:
Some of the changes are already backported in v1.19.x. But it may not hurt to include them in the NEWS.md?

shijin-aws (approved):

j-xiong:
@shijin-aws v1.19.0 is the latest release available, we are including all updates since then. When we release 1.19.1 in the future, we will need to make a decision on how to handle the duplications.

Issue #9535, opened 2023-11-09
core/memhooks, efa (?): segfault when starting memhooks monitor on RHEL9 aarch64 OS


shijin-aws:
**Describe the bug**
I got a segfault when running `fi_mr_test` with efa provider on the latest RHEL9 aarch64 OS, the backtrace shows the segfault happens when the memhooks monitor starts and tries to patch the `shmdt` function

```
#0  0x0000ffffb9e7a6f0 in __memcpy_simd () from /lib64/libc.so.6
#1  0x0000ffffba079fe0 in ofi_write_patch (patch_data=0xffffba212ce8 <intercepts+680> "\017", address=0xffffb9f3cff0 <shmdt>, data_size=20)
    at prov/util/src/util_mem_hooks.c:226
#2  0x0000ffffba07a0e8 in ofi_apply_patch (intercept=0xffffba212cc0 <intercepts+640>) at prov/util/src/util_mem_hooks.c:247
#3  0x0000ffffba07a344 in ofi_patch_function (intercept=0xffffba212cc0 <intercepts+640>) at prov/util/src/util_mem_hooks.c:353
#4  0x0000ffffba07a648 in ofi_intercept_symbol (intercept=0xffffba212cc0 <intercepts+640>) at prov/util/src/util_mem_hooks.c:411
#5  0x0000ffffba07abb0 in ofi_memhooks_start (monitor=0xffffba2129d8 <memhooks>) at prov/util/src/util_mem_hooks.c:604
#6  0x0000ffffba0bf740 in efa_mr_cache_open (cache=0x12146978, domain=0x12146840) at prov/efa/src/efa_mr.c:121
#7  0x0000ffffba0bee7c in efa_domain_open (fabric_fid=0x12146530, info=0x12142e30, domain_fid=0x430a68 <domain>, context=0x0)
    at prov/efa/src/efa_domain.c:288
#8  0x0000000000403330 in fi_domain (fabric=0x12146530, info=0x12142e30, domain=0x430a68 <domain>, context=0x0)
    at /home/ec2-user/PortaFiducia/build/libraries/libfabric/v1.19.x-debug/install/libfabric/include/rdma/fi_domain.h:334
#9  0x0000000000405ce0 in ft_open_domain_res () at common/shared.c:642
#10 0x00000000004060bc in ft_open_fabric_res () at common/shared.c:690
[...]

PR #9552, opened 2023-11-14
util/memhooks: Fix a bug when calculating mprotect region

shijin-aws:
In ofi_write_patch, the bound should be the page end for  address + data_size. And since it's the end of the page, when calculating the length, it needs to add 1 to make the length a multiple of page size.

Fix https://github.com/ofiwg/libfabric/issues/9535

### DIFF
--- a/prov/util/src/util_mem_hooks.c
+++ b/prov/util/src/util_mem_hooks.c
@@ -220,8 +220,8 @@ static inline int ofi_write_patch(unsigned char *patch_data, void *address,
 	}
 
 	base = ofi_get_page_start(address, page_size);
-	bound = ofi_get_page_end(address, page_size);
-	length = (uintptr_t) bound - (uintptr_t) base;
+	bound = ofi_get_page_end( (void *) ((uintptr_t) address + data_size - 1), page_size);
+	length = (uintptr_t) bound - (uintptr_t) base + 1;
 
 	if (mprotect(base, length, PROT_EXEC|PROT_READ|PROT_WRITE)) {
 		FI_WARN(&core_prov, FI_LOG_MR,


shijin-aws:
Not sure how far I should backport it.. This bug has been there for 3 years

wenduwan (approved):

wzamazon:
What is the impact of this bug?

shijin-aws:
> What is the impact of this bug?

This bug happens when Libfabric's memhooks monitor starts, it can happen when meeting the following criteria:

1. Application doesn't set FI_MR_LOCAL in mr_mode so efa provider will start memory monitor.
2. It's running with Open MPI < 4.1.2. For Open MPI 4.1.2, it exports Open MPI's patch monitor and Libfabric will use it. I checked the Ompi code and it doesn't have such bug. I am not sure if Intel MPI has internal memory monitor

I am not why we haven't seen it earlier. Currently we only found crash for fi_mr_test on RHEL9 aarch64 https://github.com/ofiwg/libfabric/issues/9535

prov/util/src/util_mem_hooks.c
```diff
@@ -220,8 +220,8 @@ static inline int ofi_write_patch(unsigned char *patch_data, void *address,
 	}
 
 	base = ofi_get_page_start(address, page_size);
-	bound = ofi_get_page_end(address, page_size);
-	length = (uintptr_t) bound - (uintptr_t) base;
+	bound = ofi_get_page_end( (void *) ((uintptr_t) address + data_size), page_size);
```
j-xiong, line 223: should be `addr + data_size -1`.

shijin-aws: Ah, right, thanks

j-xiong:
The patch data size is relatively small (<= 20 bytes). The bug only shows up when the patched address is at the very end of a page. That's probably why the bug hasn't been exposed until now.

j-xiong:
For backporting, we should at least go back to 1.17, and can go all the way to 1.15. Below that seems to be unnecessary.

j-xiong (approved):

Issue #9336, opened 2023-09-15
prov/verbs: Few fabtests failing after setting FI_OFI_RXM_USE_SRX=true for verbs;ofi_rxm


Juee14Desai:
**Description**
Running the following tests after setting FI_OFI_RXM_USE_SRX=true results in assertion failure for verbs;ofi_rxm:
fi_cq_data -e rdm -o writedata -p "verbs;ofi_rxm"
fi_rma_bw -e rdm -o writedata -p "verbs;ofi_rxm"
fi_rma_bw -e rdm -o writedata -u -p "verbs;ofi_rxm"
fi_multinode_coll -n 3 -p "verbs;ofi_rxm"

**To Reproduce**

set FI_OFI_RXM_USE_SRX=true environment variable and run following tests:
server: fi_rma_bw -e rdm -o writedata -p "verbs;ofi_rxm"   -s <server_ip>
client: fi_rma_bw -e rdm -o writedata -p "verbs;ofi_rxm"   -s <client_ip> <server_ip>

**Output**
Backtrace from gdb:
(gdb) bt
#0  0x00007ffff56fbcbb in raise () from /lib64/libc.so.6
#1  0x00007ffff56fd355 in abort () from /lib64/libc.so.6
#2  0x00007ffff56f3cba in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007ffff56f3d42 in __assert_fail () from /lib64/libc.so.6
#4  0x00007ffff77e3e20 in vrb_alloc_ctx (progress=0x4341c0)
    at prov/verbs/src/verbs_ofi.h:983
#5  0x00007ffff77e9760 in vrb_post_srq (srx=0xf4b810, 
    wr=0x7fffffffd180) at prov/verbs/src/verbs_ep.c:1536
#6  0x00007ffff77e9a82 in vrb_srx_recv (ep_fid=0xf4b810, 
    buf=0x7ffff0d80258, len=16448, desc=0x444610, 
    src_addr=18446744073709551615, context=0x7ffff0d80170)
    at prov/verbs/src/verbs_ep.c:1582
#7  0x00007ffff780fb92 in fi_recv (ep=0xf4b810, buf=0x7ffff0d80258, 
    len=16448, desc=0x444610, src_addr=18446744073709551615, 
    context=0x7ffff0d80170) at ./include/rdma/fi_endpoint.h:297
#8  0x00007ffff7819363 in rxm_post_recv (rx_buf=0x7ffff0d80170)
[...]

PR #9572, opened 2023-11-15
prov/verbs: Add missing lock to protect SRX

sydidelot:
When compiled with --enable-debug, the test fi_shared_ctx fails on the following assertion failure:
`fi_shared_ctx: prov/verbs/src/verbs_ofi.h:993: vrb_alloc_ctx: Assertion 'ofi_genlock_held(progress->active_lock)' failed.`

And here is the associated stack:
```
6  0x00007ffff7a39e96 in __GI___assert_fail (assertion=0x7ffff7f2ec28 "ofi_genlock_held(progress->active_lock)", file=0x7ffff7f2ec08 "prov/verbs/src/verbs_ofi.h", line=993, function=0x7ffff7f2f338 <__PRETTY_FUNCTION__.45> "vrb_alloc_ctx") at ./assert/assert.c:101 
7  0x00007ffff7e06b9b in vrb_alloc_ctx (progress=0x5555555ca1a0) at prov/verbs/src/verbs_ofi.h:993 
8  0x00007ffff7e0b9fe in vrb_post_srq (srx=0x5555555ca7c0, wr=0x7fffffffe120) at prov/verbs/src/verbs_ep.c:1603 
9  0x00007ffff7e0bd2d in vrb_srx_recv (ep_fid=0x5555555ca7c0, buf=0x0, len=1024, desc=0x0, src_addr=18446744073709551615, context=0x55555556b740 <rx_ctx>) at prov/verbs/src/verbs_ep.c:1649 
10 0x000055555555d720 in fi_recv (context=0x55555556b740 <rx_ctx>, src_addr=<optimized out>, desc=0x0, len=1024, buf=0x0, ep=0x5555555ca7c0) at /home/sdidelot/libfabric/include/rdma/fi_endpoint.h:297 
11 ft_post_rx_buf (ep=0x5555555ca7c0, size=1024, ctx=0x55555556b740 <rx_ctx>, op_buf=0x0, op_mr_desc=0x0, op_tag=0) at common/shared.c:2392 
12 0x000055555555d937 in ft_post_rx (ep=<optimized out>, size=<optimized out>, ctx=<optimized out>) at common/shared.c:2400 
13 0x00005555555576cf in server_connect () at functional/shared_ctx.c:502 
14 run () at functional/shared_ctx.c:547
15 main (argc=<optimized out>, argv=<optimized out>) at functional/shared_ctx.c:629
```

The problem is that `vrb_post_srq()` doesn't acquire `progress->active_lock` when `vrb_alloc_ctx(`) is called, which may result in a race condition if multiple threads concurrently access the same SRX queue.

### DIFF
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -1533,9 +1533,12 @@ ssize_t vrb_post_srq(struct vrb_srx *srx, struct ibv_recv_wr *wr)
 	struct ibv_recv_wr *bad_wr;
 	int ret;
 
+	ofi_genlock_lock(vrb_srx2_progress(srx)->active_lock);
 	ctx = vrb_alloc_ctx(vrb_srx2_progress(srx));
-	if (!ctx)
-		return -FI_EAGAIN;
+	if (!ctx) {
+		ret = -FI_EAGAIN;
+		goto unlock;
+	}
 
 	ctx->srx = srx;
 	ctx->user_ctx = (void *) (uintptr_t) wr->wr_id;
@@ -1549,6 +1552,9 @@ ssize_t vrb_post_srq(struct vrb_srx *srx, struct ibv_recv_wr *wr)
 		vrb_free_ctx(vrb_srx2_progress(srx), ctx);
 		ret = FI_EAGAIN;
 	}
+
+unlock:
+	ofi_genlock_unlock(vrb_srx2_progress(srx)->active_lock);
 	return ret;
 }
 


sydidelot:
@j-xiong I suggest we switch on debug mode (`./configure --enable-debug`) in the CI pipeline. What do you think?

sydidelot:
Ah, it seems that my patch fixes #9336 by mistake :)

j-xiong:
@sydidelot For CI build check, I think we want both debug and non-debug mode. It's not the bottleneck of CI pipelines anyway.

j-xiong (approved):

j-xiong:
Please backport.

j-xiong:
@sydidelot Current CI pipeline do test regular, debug, and dl builds. The issue was not caught before because the `fi_share_ctx` test is effectively excluded for the verbs provider. The intention was to enable shared rx ctx test only, but the filter setting filtered out more than what was intended to.

PR #9574, opened 2023-11-15
prov/verbs_nd: Check error code from GetPrivateData

k0zmo:
On `ECONNREFUSED` Mellanox drivers return `0xc000023a` while still writing to `len` a non-zero value. This commit addresses the issue by checking both the error code and `len` for non-zero values.


j-xiong:
Please add a 'Signed-off-by:' line to the commit message.

j-xiong (approved):

PR #9567, opened 2023-11-15
[v1.15.x] Cherry-picked fixes to allow CI build check to pass

j-xiong:
Fix MacOS build issues.

PR #9575, opened 2023-11-15
[v1.15.x] contrib/intel: synchronize with main

aingerson:
New CI uses slurm instead of Jenkins node allocation. Update branch to new scripts but remove invalid tests (dmabuf tests and DSA testing which came after)


j-xiong:
Ignore the macos build failure. It won't succeed until #9567 is merged.

j-xiong:
Looks good. No need to wait for AWS CI which is unrelated to this PR.

PR #9576, opened 2023-11-16
[v1.16.x] contrib/intel: synchronize with main

aingerson:
New CI uses slurm instead of Jenkins node allocation. Update branch to new scripts but remove invalid tests (ucx provider and DSA testing which came after).

Also changes name of dmabuf test which was changed after 1.16 from fi-rdmabw-xe


j-xiong:
Intel CI passed. That's all matter for this PR.

PR #9578, opened 2023-11-16
github/actions: Add "-Werror" to CI build check runs

j-xiong:
Instead of turning on "-Werror" permanently, which has been proven to be troublesome, add the flag to the CI build check pipeline. This way we are still able to identify and fix warnings as early as possible while avoid causing failures to other CI jobs.


j-xiong:
AWS CI failure is unrelated. Ignoring.

PR #9587, opened 2023-11-17
man: Add ucx provider to the fi_provider man page

j-xiong:


j-xiong:
No need to wait for CI for man page update.

PR #9616, opened 2023-11-28
prov/efa: Update lttng directions, specify glibc version requirement

a-szegel:


darrylabbate:
You can just add glibc >= 2.28 to the Prerequisites section of the doc (`glib2-devel` on AL2). The outdated glibc on AL2 is a common issue.

darrylabbate:
Out of scope, but I should probably just add my build script for LTTng 2.13+ and Babeltrace 2 as a "quick start." LTTng >= 2.13 can be easily installed via the system package manager for most relevant distros, but there are exceptions like AL2.

prov/efa/docs/lttng.md
```diff
@@ -17,6 +17,8 @@ If these are missing, they should be fine to install via `yum` or other equivale
 * [libuuid](https://sourceforge.net/projects/libuuid/)
 * [popt](https://directory.fsf.org/wiki/Popt)
 * [libxml2](http://www.xmlsoft.org/)
+* [glib2-devel] (glibc 2.28 is required for babeltrace2)
+* [elfutils-devel] (required for babeltrace2)
```
darrylabbate, line 21: - Provide a link or remove the link formatting.
- Use the generic names for the packages. `glib2-devel` and `elfutils-devel` are specific to `yum`.
- Babeltrace2's dependency is pedantically GLib, not glibc ([ref](https://github.com/efficios/babeltrace/blob/stable-2.0/README.adoc#build-time-requirements))
- I would also recommend moving these to the section on Babeltrace2, since it's not a hard requirement for LTTng-UST (i.e. you can analyze LTTng traces using other tools, like Trace Compass)

```suggestion
* [GLib](https://developer.gnome.org/glib/) >= 2.28
* [elfutils](https://sourceware.org/elfutils/) >= 0.154
```

darrylabbate:
> These seem to be non-AL2 specific.

The dependencies themselves are not specific to AL2. My point was that the package manager will resolve them for you - AL2 being kind of an outlier since LTTng 2.13 isn't available via `yum`. It's only tangentially relevant to this PR.

darrylabbate (approved):

PR #9617, opened 2023-11-29
man: Clarify the usage of FI_RMOTE_CQ_DATA flag

j-xiong:
The man pages of `fi_msg`, `fi_tagged`, and `fi_rma` gave the false impression that the behavior of `fi_*data` calls could be affected by whether this flag is set. In reality, the flag is not a valid choice for endpoint `op_flags` and the `fi_*data` calls always act as if the flag is set.

Update the man pages to eliminate the confusion.

Reference: https://github.com/ofiwg/libfabric/discussions/9527

### DIFF
--- a/man/fi_msg.3.md
+++ b/man/fi_msg.3.md
@@ -219,10 +219,10 @@ The following list of flags are usable with fi_recvmsg and/or
 fi_sendmsg.
 
 *FI_REMOTE_CQ_DATA*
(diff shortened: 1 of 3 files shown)


shijin-aws (approved):

j-xiong:
Ignoring the AWS CI failure which is unrelated.

PR #9618, opened 2023-11-30
hmem/rocr: Add ROCr dmabuf support under verbs

AtlantaPepsi:
- Added `get_dmabuf_fd` for ROCr under HMEM interface, it acquires dmabuf handle through HSA runtime and is now invoked by `verbs` providers.
- 3 conditions for support of this functionality: ROCr support, Linux kernel support, and a user set environmental variable `FI_HMEM_ROCR_USE_DMABUF` (please let us know if naming is ok) as a switch that will be checked during runtime
- currently for verbs, if dmabuf handle cannot be successfully acquired for any reason, it will opt to not register any memory region and [just crash](https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_mr.c#L85). We are not sure if this is intended or a bug, but tentatively we fixed this and make it invoke failover routine in case of failure.

### DIFF
--- a/src/hmem_rocr.c
+++ b/src/hmem_rocr.c
@@ -40,6 +40,9 @@
 
 #if HAVE_ROCR
 
+#include <stdio.h>
+#include <sys/utsname.h>
+
 #define HSA_MAX_SIGNALS 512
 #define HSA_MAX_STREAMS (HSA_MAX_SIGNALS / MAX_NUM_ASYNC_OP)
 #define D2H_THRESHOLD 16384
@@ -122,6 +125,10 @@ struct hsa_ops {
 	hsa_status_t (*hsa_iterate_agents)(
 		hsa_status_t (*callback)(hsa_agent_t agent, void *data),
 		void *data);
+	hsa_status_t (*hsa_system_get_info)(hsa_system_info_t attribute, 
+					    void* value);
+	hsa_status_t (*hsa_amd_portable_export_dmabuf)(const void* ptr, size_t size, 
+						       int* dmabuf, uint64_t* offset);
 };
 
 #if ENABLE_ROCR_DLOPEN
@@ -172,6 +179,8 @@ static struct hsa_ops hsa_ops = {
 	.hsa_signal_create = hsa_signal_create,
 	.hsa_signal_destroy = hsa_signal_destroy,
 	.hsa_iterate_agents = hsa_iterate_agents,
+	.hsa_amd_portable_export_dmabuf = hsa_amd_portable_export_dmabuf,
+	.hsa_system_get_info = hsa_system_get_info,
 };
 
 #endif /* ENABLE_ROCR_DLOPEN */
@@ -881,6 +890,9 @@ int rocr_hmem_init(void)
 		"Threshold for switching to hsa memcpy for device-to-host"
 		"  copies. (Default 16384");
 
+	fi_param_define(NULL, "hmem_rocr_use_dmabuf", FI_PARAM_INT,
+			"Use dma-buf for sharing buffer with hardware. (default:0)");
+
 	ret = rocr_hmem_dl_init();
 	if (ret != FI_SUCCESS)
 		return ret;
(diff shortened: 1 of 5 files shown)


AtlantaPepsi:
@edgargabriel @nusislam for additional review

src/hmem_rocr.c
```diff
@@ -1054,6 +1063,100 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
...
+	hsa_status_t hsa_ret;
+	bool dmabuf_support = false, dmabuf_kernel = false;
+
+	char *s;
+	if (!(s = getenv("FI_HMEM_ROCR_USE_DMABUF")) || !atoi(s)) goto out;
+
+	// HSA_AMD_SYSTEM_INFO_DMABUF_SUPPORTED = 0x204, we use the number instead of var name
+	// for backward compatibility reasons.
+	hsa_ret = hsa_ops.hsa_system_get_info((hsa_system_info_t) 0x204, &dmabuf_support);
+
```
j-xiong, line 1077: nitpick: here we usually don't add the empty line between the call and the return value checking.

src/hmem_rocr.c
```diff
@@ -1054,6 +1063,100 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
...
+	fclose(fp);
+#endif
+out:
+	return dmabuf_support && dmabuf_kernel;
+}
+
+int rocr_hmem_get_dmabuf_fd(const void *addr, uint64_t size, int *dmabuf_fd,
+			    uint64_t *offset)
+{
+	static bool supported = false, checked = false;
```
j-xiong, line 1133: add an empty line here.

src/hmem_rocr.c
```diff
@@ -1054,6 +1063,100 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
...
+	if (!checked)
+	{
```
j-xiong, line 1135: move `{` to the end of the previous line.

src/hmem_rocr.c
```diff
@@ -1054,6 +1063,100 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
...
+		return -FI_EOPNOTSUPP;
+
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
+	hsa_status_t hsa_ret;
+
+	// maybe need base addr calculation here? empirically the hsa call here does it internally
+	hsa_ret = hsa_ops.hsa_amd_portable_export_dmabuf(addr, size, dmabuf_fd, offset);
+
+	if (hsa_ret != HSA_STATUS_SUCCESS)
+	{
```
j-xiong, line 1149: same here.

src/hmem_rocr.c
```diff
@@ -1054,6 +1063,100 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
 	return FI_SUCCESS;
 }
 
+bool rocr_is_dmabuf_supported(void)
```
j-xiong, line 1066: Is this function intended to be used anywhere else? If not, maybe define it as static. If yes, consider moving the `checked` logic from the caller side to here to avoid duplicated checking.

src/hmem_rocr.c
AtlantaPepsi: I'm assuming the return value checking clause after this should also have { at same line?

src/hmem_rocr.c
AtlantaPepsi: that makes sense and I made it static, the `hmem_cuda` counterpart also have this function but non-static, I wonder if it had a specific reason there?

shijin-aws:
bot:aws:retest

src/hmem_rocr.c
```diff
@@ -1054,6 +1066,99 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
...
+
+	if (!checked) {
+		supported = rocr_is_dmabuf_supported();
+		checked = true;
+	}
+	if (!supported)
+		return -FI_EOPNOTSUPP;
+
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
+	hsa_status_t hsa_ret;
```
j-xiong, line 1146: You forgot this one.

j-xiong (commented):
Also please squash the third commit into the first one and update the PR by "git push -f". We don't do auto-squash here and don't expect one commit fixing another commit in the same PR.

src/hmem_rocr.c
AtlantaPepsi: @j-xiong would you please check again?

j-xiong:
Windows build error:
```
  C:\projects\libfabric\src\hmem_rocr.c(42,1): fatal error C1083: Cannot open include file: 'sys/utsname.h': No such file or directory [C:\projects\libfabric\libfabric.vcxproj]
```

j-xiong:
I think this `#include` line can be moved below `#if HAVE_ROCR`.

shijin-aws:
bot:aws:retest

AtlantaPepsi:
bot:aws:retest

> I think this `#include` line can be moved below `#if HAVE_ROCR`.

@j-xiong how about now?

j-xiong:
@AtlantaPepsi Let's wait for the CI test result. FYI, the `appveyor` CI is the windows build test.

AtlantaPepsi:
@j-xiong I think all checks have been passed

j-xiong (approved):

PR #9622, opened 2023-12-04
prov/verbs: Address cm_id resource leak in rdma_reject path

k0zmo:
Add missing calls to `rdma_destroy_id` after a call to `rdma_reject` to prevent `cm_id` resource leak.

The resource leak can be observed by running `rdma res` and checking the value of `cm_id`. Each time a passive endpoint rejects an incoming connection, `cm_id` is increased. This behavior is not observed when a connection is accepted and then closed abruptly or gracefully by a call to `fi_shutdown`.

I'm not sure about the correctness of `rdma_destroy_id` in XRC path in verbs_eq.c

### DIFF
--- a/prov/verbs/src/verbs_cm.c
+++ b/prov/verbs/src/verbs_cm.c
@@ -296,6 +296,9 @@ vrb_msg_xrc_ep_reject(struct vrb_connreq *connreq,
 			       connreq->xrc.conn_tag, connreq->xrc.port, 0, 0);
 	ret = rdma_reject(connreq->id, cm_data,
 			  (uint8_t) paramlen) ? -errno : 0;
+	if (rdma_destroy_id(connreq->id))
+		VRB_WARN_ERR(FI_LOG_EP_CTRL, "rdma_destroy_id", -errno);
+	connreq->id = NULL;
 	free(cm_data);
 	return ret;
 }
@@ -324,6 +327,9 @@ vrb_msg_ep_reject(struct fid_pep *pep, fid_t handle,
 	} else if (connreq->id) {
 		ret = rdma_reject(connreq->id, cm_hdr,
 			(uint8_t)(sizeof(*cm_hdr) + paramlen)) ? -errno : 0;
+		if (rdma_destroy_id(connreq->id))
+			VRB_WARN_ERR(FI_LOG_EP_CTRL, "rdma_destroy_id", -errno);
+		connreq->id = NULL;
 	} else {
 		ret = -FI_EBUSY;
 	}
--- a/prov/verbs/src/verbs_eq.c
+++ b/prov/verbs/src/verbs_eq.c
@@ -540,6 +540,13 @@ vrb_eq_xrc_connreq_event(struct vrb_eq *eq, struct fi_eq_cm_entry *entry,
 send_reject:
 	if (rdma_reject(connreq->id, *priv_data, *priv_datalen))
 		VRB_WARN(FI_LOG_EP_CTRL, "rdma_reject %d\n", -errno);
+	if (rdma_destroy_id(connreq->id))
+		VRB_WARN(FI_LOG_EP_CTRL, "rdma_destroy_id %d\n", -errno);
+
+	ep->base_ep.info_attr.handle = NULL;
+	ep->tgt_id = NULL;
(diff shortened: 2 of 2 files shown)


prov/verbs/src/verbs_cm.c
```diff
@@ -332,6 +332,10 @@ vrb_msg_ep_reject(struct fid_pep *pep, fid_t handle,
 	if (ret)
 		VRB_WARN_ERR(FI_LOG_EP_CTRL, "rdma_reject", ret);
 
+	ret = rdma_destroy_id(connreq->id) ? -errno : 0;
```
chien-intel, line 335: Move this code to after rdma_reject call above, inside if (connreq->id) check.  Also assign null to connreq->id after destroy.

prov/verbs/src/verbs_eq.c
```diff
@@ -540,6 +540,8 @@ vrb_eq_xrc_connreq_event(struct vrb_eq *eq, struct fi_eq_cm_entry *entry,
 send_reject:
 	if (rdma_reject(connreq->id, *priv_data, *priv_datalen))
 		VRB_WARN(FI_LOG_EP_CTRL, "rdma_reject %d\n", -errno);
+	if (rdma_destroy_id(connreq->id))
```
chien-intel, line 543: Need to reset connreq->id.  I don't know XRC code but it seems we need to reset ep->tgt_id and maybe other info as well on reject.  i will leave this for others to comment.

prov/verbs/src/verbs_cm.c
k0zmo: Moving it under the `else if (connreq->id)` would mean we won't call `rdma_destroy_id` on `if (connreq->is_xrc)` path which also calls `rdma_reject` and then frees the `connreq` structure leaving `id` unreleased. Is that what you had in mind:

```c
	ofi_mutex_lock(&_pep->eq->event_lock);
	if (connreq->is_xrc) {
		ret = vrb_msg_xrc_ep_reject(connreq, cm_hdr,
				(uint8_t)(sizeof(*cm_hdr) + paramlen));
	} else if (connreq->id) {
		ret = rdma_reject(connreq->id, cm_hdr,
			(uint8_t)(sizeof(*cm_hdr) + paramlen)) ? -errno : 0;
		ret = rdma_destroy_id(connreq->id) ? -errno : 0;
		connreq->id = NULL;
	} else {
		ret = -FI_EBUSY;
	}
	ofi_mutex_unlock(&_pep->eq->event_lock);

	if (ret)
		VRB_WARN_ERR(FI_LOG_EP_CTRL, "rdma_reject", ret);

	free(connreq);
	return ret;
```

prov/verbs/src/verbs_cm.c
```diff
@@ -296,6 +296,10 @@ vrb_msg_xrc_ep_reject(struct vrb_connreq *connreq,
 			       connreq->xrc.conn_tag, connreq->xrc.port, 0, 0);
 	ret = rdma_reject(connreq->id, cm_data,
 			  (uint8_t) paramlen) ? -errno : 0;
+	if (rdma_destroy_id(connreq->id)) {
+		VRB_WARN_ERR(FI_LOG_EP_CTRL, "rdma_destroy_id", -errno);
+	}
```
j-xiong, line 301: We don't use braces for a single line after `if`. Same for the other one.

prov/verbs/src/verbs_eq.c
j-xiong: Yes, also need to reset `ep->tgt_id` since it is referencing a destroyed id, and `ep->base_ep.info_attr.handle`.  Not 100% sure, but `ep->recip_req_received` may need to be reset as well.

prov/verbs/src/verbs_cm.c
k0zmo: Thanks for suggestions. I can see that `connreq->id` can only be NULL since de5b92d00eec91338d451e0a55818b3b0a2abaa5 for XRC endpoint hence we elide the check for NULL.

j-xiong (approved):

j-xiong:
@chien-intel Do you have more comments on this one?

PR #9626, opened 2023-12-06
prov/efa: Make runt_size aligned

shijin-aws:
Currently, txe->runt_size can be non-multiple of the memory alignment, which caused the following issues:

1. It caused the data size to be copied on the receiver side not a multiple of memory alignment. This not only caused non-performant data copy (gdrcopy or local read), but also breaks the LL128 protocol for send/recv, which requires the data size to be copied must be a multiple of 128 (the memory alignment in this case).

2. It caused the single_pkt_entry_data_size variable in efa_rdm_ope_prepare_to_post_send() to be 0 after doing the alignment trim.

This patch makes the runt size always aligned before we decide whether to use runting read protocol. If the aligned runt size is 0, we won't do runting read.

Also added a series of unit test to validate this change


prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -673,3 +673,16 @@ void efa_rdm_ep_post_handshake_or_queue(struct efa_rdm_ep *ep, struct efa_rdm_pe
 
 	peer->flags |= EFA_RDM_PEER_HANDSHAKE_SENT;
 }
+
+size_t efa_rdm_ep_get_memory_alignment(struct efa_rdm_ep *ep, enum fi_hmem_iface iface)
+{
+	size_t memory_alignment = 8;
```
sunkuamzn, line 679: Can we make the 8 also a macro `EFA_RDM_DEFAULT_MEMORY_ALIGNMENT` similar to the other ones?

prov/efa/test/efa_unit_test_runt.c
```diff
@@ -0,0 +1,277 @@
...
+	memset(&mock_txe, 0, sizeof(mock_txe));
+	mock_txe.total_len = total_len;
+	mock_txe.addr = addr;
+	mock_txe.iov_count = 1;
+	mock_txe.iov[0].iov_base = NULL;
+	mock_txe.iov[0].iov_len = 9000;
+	mock_txe.desc[0] = &mock_mr;
+
+	runt_size = efa_rdm_peer_get_runt_size(peer, efa_rdm_ep, &mock_txe);
+	printf("runt_size: %lu, expected size: %lu\n", runt_size, expected_runt_size);
```
wenduwan, line 55: Forgot to delete

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -535,8 +529,18 @@ ssize_t efa_rdm_ope_prepare_to_post_send(struct efa_rdm_ope *ope,
 		 */
 		single_pkt_entry_data_size = (total_pkt_entry_data_size - 1) / *pkt_entry_cnt + 1;
 
-		/* each packet must be aligned */
+		/*
+		 * Each packet must be aligned.
+		 * single_pkt_entry_data_size & ~(memory_alignment - 1) is 0 when total_pkt_entry_data_size
```
wenduwan, line 534: > single_pkt_entry_data_size & ~(memory_alignment - 1) is 0 when total_pkt_entry_data_size is smaller than memory_alignment

I think this is a typo `single_pkt_entry_data_size & ~(memory_alignment - 1) is 0 when single_pkt_entry_data_size is smaller than memory_alignment`

Then this comment does not help... I think the assertion on L542 should be sufficient.

prov/efa/test/efa_unit_test_runt.c
```diff
@@ -0,0 +1,277 @@
...
+	peer->num_runt_bytes_in_flight = peer_num_runt_bytes_in_flight;
+
+	mock_mr.peer.iface = iface;
+
+	memset(&mock_txe, 0, sizeof(mock_txe));
+	mock_txe.total_len = total_len;
+	mock_txe.addr = addr;
+	mock_txe.iov_count = 1;
+	mock_txe.iov[0].iov_base = NULL;
+	mock_txe.iov[0].iov_len = 9000;
```
wenduwan, line 51: `mock_txe.iov[0].iov_len = total_len` or simply skip initializing `mock_txe.iov`

Same in `test_efa_rdm_peer_select_readbase_rtm_impl`

prov/efa/test/efa_unit_test_runt.c
```diff
@@ -0,0 +1,277 @@
...
+	size_t expected_runt_size;
+	size_t peer_num_runt_bytes_in_flight;
+	size_t total_runt_size;
+
+	efa_unit_test_resource_construct(resource, FI_EP_RDM);
+
+	msg_length = 12000;
+	peer_num_runt_bytes_in_flight = 10000;
+	total_runt_size = 16384;
+	/* 16384 - 10000 is smaller than 12000 (total_len), runt size must be (16384 - 10000) // 64 * 64 = 11968 */
```
wenduwan, line 126: `(16384 - 10000) // 64 * 64 = 6336`

wenduwan (approved):

prov/efa/test/efa_unit_test_runt.c
shijin-aws: Shoot...

prov/efa/src/rdm/efa_rdm_ep_utils.c
shijin-aws: Done.

prov/efa/test/efa_unit_test_runt.c
shijin-aws: Good catch!

prov/efa/src/rdm/efa_rdm_ope.c
shijin-aws: Thanks, removed the unnecessary comment

prov/efa/test/efa_unit_test_runt.c
shijin-aws: Done

prov/efa/test/efa_unit_test_runt.c
shijin-aws: Thank you! Fixed.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -365,11 +363,10 @@ void efa_rdm_txe_set_runt_size(struct efa_rdm_ep *ep, struct efa_rdm_ope *txe)
...
 	peer = efa_rdm_ep_get_peer(ep, txe->addr);
 
-	iface = txe->desc[0] ? ((struct efa_mr*) txe->desc[0])->peer.iface : FI_HMEM_SYSTEM;
-	hmem_info = &efa_rdm_ep_domain(ep)->hmem_info[iface];
-
 	assert(peer);
-	txe->bytes_runt = MIN(hmem_info->runt_size - peer->num_runt_bytes_in_flight, txe->total_len);
+	txe->bytes_runt = efa_rdm_peer_get_runt_size(peer, ep, txe);
+
+	assert(txe->bytes_runt);
```
sunkuamzn, line 369: You have an assert on `txe->bytes_runt` but the value could be zero

prov/efa/test/efa_unit_test_runt.c
```diff
@@ -0,0 +1,270 @@
...
+	size_t expected_runt_size;
+	size_t peer_num_runt_bytes_in_flight;
+	size_t total_runt_size;
+
+	efa_unit_test_resource_construct(resource, FI_EP_RDM);
+
+	msg_length = 12000;
+	peer_num_runt_bytes_in_flight = 1000;
+	total_runt_size = 1004;
+	/* 1048 - 1000 is smaller than host memory alignment, runt size must be 0 */
```
sunkuamzn, line 140: Comment says 1048 but runt size is 1004

prov/efa/src/rdm/efa_rdm_ope.c
sunkuamzn: Oh never mind. If `efa_rdm_peer_get_runt_size` returns 0, then we go to long read. There cannot be a race condition because of the srx lock.

sunkuamzn (approved):

prov/efa/test/efa_unit_test_runt.c
shijin-aws: Thank you! Fixed

PR #9638, opened 2023-12-12
contrib/intel/jenkins: Update jenkins architecture for future changes

zachdworkin:
Separate software builds by hardware.
This will allow easier debugging in the case of test failures as well as make it easier to distinguish which packages are in use on a given node.
This separates each job by HW type and then builds on that specific hardware. The directory structure is now structured by workspace with everything in sub directories inside it instead of installing to a new location not controlled by cloudbees.
These changes will make it easier to upgrade our CI in the future by enabling movement of an entire HW type to the new software solution instead of having to wait for everything to be ready.


nikhilnanal:
This PR looks good to merge. The failing test idup_nb on a weekly is unrelated to this change and will be excluded for now in a separate PR.

PR #9639, opened 2023-12-12
Opx upstream December 2023

charlesshereda:
Commits include:

- Make update HDRQ register frequency configurable at build time
- Memory leak fixes
- General bug fixes
- CN5000 prepatory work and code changes
- Work to enable GPUs and associated bugfixes
- Fix max SHM connections to allow up to 16 HFI
- TID fixes
- SDMA replay fixes
- Reduce size of opx_shm_tx
- Fix multi-receive to work with contiguous rzv
- Fix intranode check to handle multi-rail
- Limit multipacket eager implementation to tagged sends
- Debug counter improvements


charlesshereda:
Change to draft until we resolve errors.

prov/opx/include/rdma/opx/fi_opx_endpoint.h
```diff
@@ -1409,6 +1512,17 @@ void complete_receive_operation_internal (struct fid_ep *ep,
 			if (OFI_LIKELY(niov == 1)) {
 				assert(payload != NULL);
 
+				uint64_t rbuf_device;
+				enum fi_hmem_iface rbuf_iface;
```
j-xiong, line 1643: I understand the existing code in the opx provider has already broken the rule, but the convention is not to interleave declarations with statements. I will leave your team to decide if a refactoring is needed in the future.

prov/opx/include/rdma/opx/fi_opx_endpoint.h
```diff
@@ -1419,35 +1533,42 @@ void complete_receive_operation_internal (struct fid_ep *ep,
...
+							 is_intranode,
+							 reliability,		/* compile-time constant expression */
 							 u32_ext_rx);
 
 				/*
 				 * copy the immediate payload data
 				 */
+				if (is_hmem) {
+					rbuf = hmem_buf;
+				}
```
j-xiong, line 1706: unnecessary braces.

prov/opx/include/rdma/opx/fi_opx_endpoint.h
```diff
@@ -1848,59 +2007,57 @@ void fi_opx_ep_rx_process_header_rzv_data(struct fi_opx_ep * opx_ep,
...
-						       tid_length);
-				/* free the rendezvous completion structure */
-				OPX_BUF_FREE(rzv_comp);
-			}
+				"tid vaddr>buf [%p - %p] tid len %lu/%#lX\n",
+				(void *)tid_vaddr,
+				(void *)(tid_vaddr + tid_length),
+				tid_length, tid_length);
+			opx_deregister_for_rzv(opx_ep, tid_vaddr,
+						tid_length);
```
j-xiong: No need to break the line.

prov/opx/src/fi_opx_ep.c
```diff
@@ -1433,11 +1434,16 @@ static int fi_opx_open_command_queues(struct fi_opx_ep *opx_ep)
 			FI_WARN(fi_opx_global.prov, FI_LOG_EP_DATA,
 					"Too many tx contexts\n");
 			errno = FI_ENOENT;
-			goto err;
+			goto unlock;
 		}
 	}
 
+	/* Unlock */
```
j-xiong, line 1456: comments like this are unnecessary.

prov/opx/include/rdma/opx/fi_opx_endpoint.h
```diff
@@ -37,29 +37,46 @@
...
+#include "opx_shm.h"
+#include "fi_opx_tid.h"
 
 void fi_opx_cq_debug(struct fid_cq *cq, char *func, const int line);
 
 #define FI_OPX_KIND_TAG	(0)
 #define FI_OPX_KIND_MSG	(1)
 
+#define OPX_INTRANODE_TRUE		(1)
+#define OPX_INTRANODE_FALSE		(0)
```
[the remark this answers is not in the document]
belynam: The rationale behind these is that we have many functions that take several of these values as parameters, often one after another. Simply hardcoding true/false or 0/1 makes it difficult to easily discern which value belongs to which parameter, especially in situations where someone is not looking at the code with an IDE (e.g. PR reviews). It also makes it easier to grep for places we may be hardcoding true/false for a given parameter.

prov/opx/include/rdma/opx/fi_opx_endpoint.h
```diff
@@ -1278,26 +1367,29 @@ void complete_receive_operation_internal (struct fid_ep *ep,
...
-						"offset=%0X, payload_length=%ld, buffer altered at %p, "
-						"current value is %0hhX\n",
-						getpid(), __FILE__, __func__, __LINE__,
-						hdr->mp_eager_nth.mp_egr_uid, recv_buf,
-						hdr->mp_eager_nth.payload_offset, send_len,
-						buf_check, *buf_check);
-					abort();
+			if (!is_hmem) {
+				uint8_t *payload_end = ((uint8_t *)recv_buf) + send_len;
+				for (uint8_t *buf_check = (uint8_t *)recv_buf + hdr->mp_eager_nth.payload_offset;
```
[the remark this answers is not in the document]
belynam: For this comment and the next one where you said:

> the convention is not to interleave declarations with statements.

Is there a particular guideline about this? I didn't see anything in the Linux Kernel coding guidelines about it (perhaps I missed it)? I know that much older versions of the C standard did not support this, but I would think that the best practice of limiting variable scope to where it is needed would be preferable?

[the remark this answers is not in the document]
j-xiong: Yeah it's not in the current Linux kernel guide: https://www.kernel.org/doc/Documentation/process/coding-style.rst, but we have been sticking to this in the past. The opx provider was not strictly reviewed previously, partially due the amount of changes presented for each PR. 

Mixing declaration and statement makes code harder to read.  The scope is only partially limited since the other end is open. To really serve the purpose of limiting scope, use `{` and `}` instead.

I will update our coding style page to have this one explicitly listed.

charlesshereda:
> I roughly went through the changes and made some nitpicking comments along the way, mostly about coding style. The comments are only made on some examples, not for all the occurrences. The comments are for reference only. I am not holding you off on them.

@j-xiong  Ok - we had an internal discussion about this stuff, we're going to make a linter part of our process, but when we do that it will mean major changes and another upstream. So sounds like you're willing to merge this as-is for now? That would be our preference if so.

Also, FWIW, we're reluctant to remove braces from single line `if..else` statements because of the potential for bugs to be introduced with that style, even though we recognize that lack-of-braces style is consistent with the Linux kernel style.

j-xiong:
@charlesshereda Yes, I can merge it as it is.

charlesshereda:
Thank you.

j-xiong (approved):

PR #9640, opened 2023-12-14
prov/efa: Don't do handshake for local read

shijin-aws:
Currently, we always trigger a handshake to check the interoperatability on two sides for rdma read capabilities, before determining the protocol to use. However, if the read is from the self peer, we don't need such handshake and just need to check the rdma capability for local ep.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_rma.c
+++ b/prov/efa/src/rdm/efa_rdm_rma.c
@@ -173,8 +173,10 @@ ssize_t efa_rdm_rma_readmsg(struct fid_ep *ep, const struct fi_msg_rma *msg, uin
 
 	/*
 	 * A handshake is required to choose the correct protocol (whether to use device read).
+	 * For local read (read from self ep), such handshake is not needed because we only
+	 * need to check the local ep's capabilities.
 	 */
-	if (!(peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED)) {
+	if (!(peer->is_self) && !(peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED)) {
 		err = efa_rdm_ep_trigger_handshake(efa_rdm_ep, txe->addr);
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_rma.c
```diff
@@ -173,8 +173,10 @@ ssize_t efa_rdm_rma_readmsg(struct fid_ep *ep, const struct fi_msg_rma *msg, uin
 
 	/*
 	 * A handshake is required to choose the correct protocol (whether to use device read).
+	 * For local read (read from self ep), such handshake is not needed because we only
+	 * need to check the local ep's capabilities.
 	 */
-	if (!(peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED)) {
+	if (!(peer->is_self) && !(peer->flags & EFA_RDM_PEER_HANDSHAKE_RECEIVED)) {
 		err = efa_rdm_ep_trigger_handshake(efa_rdm_ep, txe->addr);
```
darrylabbate, line 180: Do you think `efa_rdm_ep_trigger_handshake()` itself should prevent handshakes when `peer->is_self` is true?

darrylabbate (approved):

prov/efa/src/rdm/efa_rdm_rma.c
shijin-aws: Don't think so. this function just shouldn't be called when peer is self. We should make sure `peer->is_self` is false inside the function, that can be an `assert()`

shijin-aws:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_rma.c
shijin-aws: The special thing for this read and write op is we enforce a handshake and return EAGAIN before the handshake is made, which can cause extra delay. In that regard preventing self peer to call this function and return EAGAIN is essential

PR #9644, opened 2023-12-15
Add missing timeout in util_wait_yield_run()

franzpoeschel:
`util_wait_yield_run()` in `prov/util/src/util_wait.c` currently ignores the timeout parameter, leading to hanging applications on our end. This PR copies the timeout logic from `util_wait_fd_run()` in the same file.


j-xiong (approved):
Thanks for your contribution!

PR #9650, opened 2023-12-18
prov/rxm: Fix data error with FI_OFI_RXM_USE_RNDV_WRITE=1

j-xiong:
Data verification would fail with rxm/verbs when FI_OFI_RXM_USE_RNDV_WRITE is set to 1. It appears that the content of the receive buffer remains unchanged after the rendezvous send is completed.

This only happens when the core provider requires FI_MR_LOCAL mode. The root cause is that MR descriptors passed through the receive call are not translated into rkeys when generating the rendezvous response. As a result, the sender always writes 0 byte.


j-xiong:
bot:aws:retest

PR #9652, opened 2023-12-18
prov/efa: Fix the ibv cq error handling.

shijin-aws:
Currently, efa_rdm_ep_poll_ibv_cq couldn't
    handle error for IBV_WC_RECV_RDMA_WITH_IMM
    and IBV_WC_RDMA_READ. This patch fixes it.
    
It also removed the failed_send/write_comps
    in the debug build, because these symbols
    are never used.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_progress.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_progress.c
@@ -450,7 +450,8 @@ static inline void efa_rdm_ep_poll_ibv_cq(struct efa_rdm_ep *ep, size_t cqe_to_p
 	struct ibv_poll_cq_attr poll_cq_attr = {.comp_mask = 0};
 	struct efa_av *efa_av;
 	struct efa_rdm_pke *pkt_entry;
-	ssize_t err, opcode;
+	ssize_t err;
+	int opcode;
 	size_t i = 0;
 	int prov_errno;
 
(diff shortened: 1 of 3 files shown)


prov/efa/src/rdm/efa_rdm_ep_progress.c
```diff
@@ -465,25 +466,39 @@ static inline void efa_rdm_ep_poll_ibv_cq(struct efa_rdm_ep *ep, size_t cqe_to_p
...
 				efa_rdm_pke_handle_tx_error(pkt_entry, FI_EIO, prov_errno);
-			} else {
-				assert(opcode == IBV_WC_RECV);
+				break;
+			case IBV_WC_RDMA_READ:
+#if ENABLE_DEBUG
+				ep->failed_read_comps++;
+#endif
+				efa_rdm_pke_handle_tx_error(pkt_entry, FI_EIO, prov_errno);
+				break;
```
wenduwan, line 490: nit we can also do fallthrough e.g.
```c
case IBV_WC_SEND:
#if ENABLE_DEBUG
    ep->failed_send_comps++;
#endif
case IBV_WC_RDMA_WRITE:
#if ENABLE_DEBUG
    ep->failed_write_comps++;
#endif
case IBV_WC_RDMA_READ:
#if ENABLE_DEBUG
    ep->failed_read_comps++;
#endif
    efa_rdm_pke_handle_tx_error(pkt_entry, FI_EIO, prov_errno);
    break;
```

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_ep_progress.c
shijin-aws: doesn't that will make `failed_write_comps` incremented for send as well?

wenduwan: I see. The counter indeed introduces extra coding. We can find a more concise approach but I don't have a strong preference.

shijin-aws: yes, ideally we should declare `failed_comps` as an array, and have a map between the ibv wc code to the ofi op code.

shijin-aws: offline talked with @wenduwan, new revision removed the unused failed_send/write_comps to reduce the code duplication

wenduwan (approved):

PR #9654, opened 2023-12-19
prov/efa: Get rid of noop instruction from empty #define

a-szegel:
I was playing around on godbolt and saw that while (0) {} generates a noop instruction, but (void) 0; does not.  Get rid of noop instructions from critical path without hurting the readability of the code.
Screenshot 2023-12-18 at 5 49 51 PM
Screenshot 2023-12-18 at 5 52 43 PM

### DIFF
--- a/prov/efa/src/efa_tp.h
+++ b/prov/efa/src/efa_tp.h
@@ -42,9 +42,9 @@ static inline void efa_tracepoint_wr_id_post_recv(const void *wr_id)
 
 #else
 
-#define efa_tracepoint(...)	while (0) {}
-#define efa_tracef(...)	while (0) {}
-#define efa_tracelog(...)	while (0) {}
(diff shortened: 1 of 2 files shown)


darrylabbate:
1. `do {} while (0)` (as some of the macros use) [doesn't produce a `nop`](https://godbolt.org/z/r4j4vbqoG). I'd prefer this over `(void) 0`.
2. Optimizations aren't enabled, but even if the compiler is producing a `nop`, it's likely for optimization/scheduling/alignment purposes, i.e. minimum sizes for fetch/decode chunks. IOW, I'd be curious if `nop`s are being emitted unconditionally or only in cases where the compiler has a good reason to do so.
3. You should remove the trailing semicolons from the macro definitions.

darrylabbate (approved):

shijin-aws (approved):

PR #9655, opened 2023-12-19
prov/efa: remove unneeded memset

a-szegel:
The memset assigns the same length as the variable we are setting.... no chance anything bad will happen.

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -138,7 +138,6 @@ fi_addr_t efa_av_reverse_lookup_rdm(struct efa_av *av, uint16_t ahn, uint16_t qp
 	struct efa_prv_reverse_av_key prv_key;
 	uint32_t *connid;
 
-	memset(&cur_key, 0, sizeof(cur_key));
 	cur_key.ahn = ahn;
 	cur_key.qpn = qpn;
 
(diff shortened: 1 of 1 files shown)


wzamazon (approved):
The reason memset can be removed is because there is no gap in `struct efa_prv_reverse_av_key`

shijin-aws:
> The reason memset can be removed is because there is no gap in `struct efa_prv_reverse_av_key`

@wzamazon what gap do you mean?

shijin-aws (approved):
~~Will it be safer to `struct efa_cur_reverse_av_key cur_key = {0};` at the beginning?~~

Ignore this, it doesn't matter since it only has ahn and qpn and we already assign them explicitly

shijin-aws:
I thought the memset isn't needed because `efa_cur_reverse_av_key` only has ahn and qpn and we already assign them explicitly

darrylabbate:
The padding would only be relevant if the object is ever cast or [punned](https://en.wikipedia.org/wiki/Type_punning) to some other struct. Otherwise, you don't (and should rarely) need to be acutely aware of the member padding/alignment.

It should be noted that compilers are pretty good at detecting superfluous `memset(&foo, 0, ...)`, so I wouldn't be surprised if this were optimized out anyway.

darrylabbate (approved):

PR #9658, opened 2023-12-19
prov/efa: reduce rma write path duplicate checks

a-szegel:


prov/efa/src/rdm/efa_rdm_rma.c
```diff
@@ -600,6 +600,10 @@ ssize_t efa_rdm_rma_writedata(struct fid_ep *ep, const void *buf, size_t len,
 	if (err)
 		return err;
 
+	err = efa_rdm_attempt_to_sync_memops(efa_rdm_ep, (void *)buf, desc);
```
shijin-aws, line 624: Hmmm this needs a backport. I am not sure how we can fix the earlier nccl 2.19 setmemops issue earlier without having this. fi_writedata is primarily called on p5. Probably we are lucky because plugin always call tsenddata first before calling this

shijin-aws: OK I take it back, it is called in writemsg anyway

prov/efa/src/rdm/efa_rdm_rma.c
```diff
@@ -429,27 +429,54 @@ ssize_t efa_rdm_rma_post_write(struct efa_rdm_ep *ep, struct efa_rdm_ope *txe)
...
+	ssize_t err;
+	struct efa_rdm_peer *peer;
+	struct efa_rdm_ep *efa_rdm_ep;
+	fi_addr_t tmp_addr;
+	struct fi_msg_rma *msg_clone;
+	void *shm_desc[EFA_RDM_IOV_LIMIT];
+	void **tmp_desc;
+
+	efa_rdm_ep = container_of(ep, struct efa_rdm_ep, base_ep.util_ep.ep_fid.fid);
+	efa_perfset_start(efa_rdm_ep, perf_efa_tx);
```
shijin-aws, line 478: I think this should be inside the generic one, same for the end

prov/efa/src/rdm/efa_rdm_rma.c
```diff
@@ -628,7 +650,10 @@ ssize_t efa_rdm_rma_writedata(struct fid_ep *ep, const void *buf, size_t len,
 	msg.rma_iov_count = 1;
 	msg.data = data;
 
-	return efa_rdm_rma_writemsg(ep, &msg, FI_REMOTE_CQ_DATA);
+	err = efa_rdm_generic_writemsg(efa_rdm_ep, &msg, FI_REMOTE_CQ_DATA);
```
shijin-aws, line 653: If you put the efa_perf_start to the generic one, you can get rid of the goto. Same comments for other places

a-szegel: I want the efa_perf_start to start as soon as possible so we can capture as much of our cost as possible.

shijin-aws: This finally increase the instruction count for the shm path. Before it was just a return, now it is a `err = ..` and `return err`

darrylabbate: I'd be more concerned with how this kind of refactor affects the likelihood of tail call optimization/elimination for this function. @a-szegel do these changes improve performance or is it just speculative?

a-szegel: I am removing duplicate calls to:
1. check rma capabilities
2. sync memops
3. getting the peer
4. seeing if we can use SHM

This should remove the amount of instructions in the critical path for EFA, which should help make us faster (less code is better than more code). 

I also fix our lttng code to make us able to capture the SHM cost, but this in turn gives us a goto out statement.  I can move the lttng code into the generic function as Shi suggests, but then we won't capture the SHM code.  I can run fi_rma_write tests with SHM before and after this change to confirm if I see a performance decrement via SHM.

a-szegel: I do see a small perf hit for SHM... will move the lttng code inside the generic function.

a-szegel: restored performance by removing goto + lttng macro.

prov/efa/src/rdm/efa_rdm_rma.c
a-szegel: done

prov/efa/src/rdm/efa_rdm_rma.c
a-szegel: no aciton.

darrylabbate:
> less code is better than more code

All I will say is there is more to optimizing programs than simply aiming for smaller machine code; the opposite generally holds true with optimized programs typically being much larger. Your (previous) changes removed some function calls from their tail positions, which _could_ preclude the compiler's ability to optimize away some call frame management overhead.

As always, trust measurements. Good luck :)

a-szegel:
on two p5's with these changes with 3 runs:
```
            "program": "/home/ec2-user/libfabric/fabtests/install/bin/fi_rma_pingpong",
            "args": ["--pin-core", "1", "-i", "0", "-o", "writedata", "-p", "efa", "-D", "cuda", "-w", "1000", "-I", "500000", "-S", "10000", "-E", "172.31.35.142"],

9.7k    500k    9.3g       31.50s    317.44      31.50       0.03
9.7k    500k    9.3g       31.54s    317.05      31.54       0.03
9.7k    500k    9.3g       31.70s    315.50      31.70       0.03
```

on the same two p5's without these changes with 3 runs:

```
            "program": "/home/ec2-user/libfabric/fabtests/install/bin/fi_rma_pingpong",
            "args": ["--pin-core", "1", "-i", "0", "-o", "writedata", "-p", "efa", "-D", "cuda", "-w", "1000", "-I", "500000", "-S", "10000", "-E", "172.31.35.142"],

9.7k    500k    9.3g       31.74s    315.09      31.74       0.03
9.7k    500k    9.3g       31.58s    316.69      31.58       0.03
9.7k    500k    9.3g       31.73s    315.20      31.73       0.03
```

a wash b/c the network dominates.

a-szegel:
I wrote some custom timing code to measure (https://github.com/a-szegel/libfabric/tree/seth-timing-of-libfabric)

Before this PR
```
Server fi_writedata libfabric to rdma_core average ns: 81.022575015003, pstd: 11.60196863061804, min: 69, max: 400, median: 80, num_samples: 1999600, unique: 69, num_outliers removed (5x mean): 400
Client fi_writedata libfabric to rdma_core average ns: 79.29778275820858, pstd: 10.276816074024078, min: 69, max: 391, median: 80, num_samples: 1999827, unique: 69, num_outliers removed (5x mean): 173
```
This PR:
```
Server fi_writedata libfabric to rdma_core average ns: 79.09410696164451, pstd: 11.778737158116673, min: 60, max: 391, median: 80, num_samples: 1999427, unique: 66, num_outliers removed (5x mean): 573
Client fi_writedata libfabric to rdma_core average ns: 76.47041087203533, pstd: 10.333130404791646, min: 60, max: 380, median: 71, num_samples: 1999552, unique: 69, num_outliers removed (5x mean): 448
```

It looks like we save a few nanoseconds.

prov/efa/src/rdm/efa_rdm_rma.c
```diff
@@ -460,16 +488,11 @@ ssize_t efa_rdm_rma_writemsg(struct fid_ep *ep,
 
 	assert(msg->iov_count <= efa_rdm_ep->tx_iov_limit);
 
-	efa_perfset_start(efa_rdm_ep, perf_efa_tx);
-	ofi_genlock_lock(srx_ctx->lock);
-
 	peer = efa_rdm_ep_get_peer(efa_rdm_ep, msg->addr);
```
shijin-aws, line 491: this should be in the generic_writemsg

a-szegel: good catch, done

prov/efa/src/rdm/efa_rdm_rma.c
```diff
@@ -430,27 +430,59 @@ ssize_t efa_rdm_rma_post_write(struct efa_rdm_ep *ep, struct efa_rdm_ope *txe)
...
+	struct efa_rdm_peer *peer;
+	struct efa_rdm_ep *efa_rdm_ep;
+	fi_addr_t tmp_addr;
+	struct fi_msg_rma *msg_clone;
+	void *shm_desc[EFA_RDM_IOV_LIMIT];
+	void **tmp_desc;
+	int err;
+
+	efa_rdm_ep = container_of(ep, struct efa_rdm_ep, base_ep.util_ep.ep_fid.fid);
 	err = efa_rdm_ep_cap_check_rma(efa_rdm_ep);
```
shijin-aws, line 486: Sigh, I missed this earlier, this should be in the generic function

shijin-aws: Sorry, I was wrong. Ignore this

shijin-aws (approved):

a-szegel:
bot:aws:retest

PR #9659, opened 2023-12-19
prov/efa: Pass in peer pointer to txe

a-szegel:
The EFA provider needs to find the peer pointer early so we can decide if we should use SHM to send the message, or if the message should go through EFA.  Instead of losing the pointer, and re-finding it when we create the txe, pass it through parameters to the txe so we don't have to duplicate the effort of finding the peer.


shijin-aws (approved):
Thanks

PR #9661, opened 2023-12-20
prov/efa: use peer pointer from txe in read, write and send

a-szegel:
The peer pointer is calculated and stored in the txe.  Use this pointer instead of constantly re-calling efa_rdm_ep_get_peer() to get this pointer for performance reasons.


shijin-aws (approved):

shijin-aws:
Can you take a look the unit test failure in AWS CI

a-szegel:
fixed... was a unit test failure b/c we weren't fully mocking the txe.

PR #7267, opened 2021-11-20
core (+ most providers): Replace spin locks with mutex

shefty:
PR is updated to manually replace most uses of spin lock with mutex.  Fastlock abstraction is replaced with ofi_spin_t and ofi_mutex_t abstractions instead.  The typedef allows us to keep the debug validation checks that we currently have.

The current code uses spin locks throughout.  However, as pointed out in issue #7053, the result can be a lack of fairness resulting in very low performance.  Additionally, if the thread holding the lock is swapped out, then another thread that tries to acquire the lock will spin on the CPU for their entire time quantum, resulting in 100% CPU utilization for that core without any work being done.  Because the locks are held during progress, it's is common for the calling thread to make kernel calls (e.g. tcp provider), register memory (efa/verbs), or allocate memory.  All of these can take significant time and result in the thread sleeping.  Unlike the kernel, there's no way for a user space thread to avoid being swapped out while holding the lock.

Modern mutex implementations already spin for a brief period of time before putting the thread to sleep.

Opening PR with changes and will discuss in the Nov. 30th OFIWG.

Note that many HPC apps are single threaded, which the code already optimizes for by using no-op locking routines.  These changes are not expected to impact those applications negatively.

Issue #7053, opened 2021-09-12
Add configure knob to use mutex in place of spinlock


songweijia:
Our libfabric application uses a dedicated thread to poll the completion queue for low latency. We are using the verbs provider. The `fi_cq_read()` implementation for verbs provider will [lock the completion_queue](https://github.com/ofiwg/libfabric/blob/10941c1f241dae15fdaab0755b6ef3f3a5e65355/prov/verbs/src/verbs_cq.c#L322) before checking its contents, and [unlock it](https://github.com/ofiwg/libfabric/blob/10941c1f241dae15fdaab0755b6ef3f3a5e65355/prov/verbs/src/verbs_cq.c#L359) after finished. However, its `fi_sendmsg()` implementation, which calls `vrb_post_send()` did something similar [here](https://github.com/ofiwg/libfabric/blob/10941c1f241dae15fdaab0755b6ef3f3a5e65355/prov/verbs/src/verbs_ep.c#L119) and [here](https://github.com/ofiwg/libfabric/blob/10941c1f241dae15fdaab0755b6ef3f3a5e65355/prov/verbs/src/verbs_ep.c#L156).

Ideally, the pthread spinlock, which the 'fastlock' uses, should be efficient enough so that the sender can always grab it to send messages. However, pthread's spinlock is neither fair nor FIFO, sometimes resulting in a terrible situation where the polling thread acquires the spinlock again and again while the sender thread waits for it for a long time (milliseconds) to proceed. This situation is very common when the number of the sender threads is small. When I disabled spinlock in `config.h` by hard-wiring `PT_LOCK_SPIN` to 0. The end-to-end throughput of our system jumps from ~700MB/s to 5GB/s.

I argue that libfabric could provide a configuration knob to disable using spin lock for the completion queue.

shefty:
The locking is provider specific, but libfabric defines threading models, which are exposed to the application via the domain attributes.  If the correct threading model is selected, locking can be converted into no-ops.  It is not safe to simply remove necessary locking.  There are states being managed that low-level verbs does not handle through its locking.

See fi_domain_attr::threading.

If locking cannot be removed because the threading model does not allow it, an alternative lock implementation would be an acceptable solution.

songweijia:
[...]

PR #9663, opened 2023-12-21
prov/util: Switch cq->ep_list_lock from mutex to spinlock

a-szegel:
This PR switches the cq's ep_list_lock from a mutex lock to a spinlock. It was shown via EFA progress benchmarks that this change cuts ~200ns from our fi_cq_read calls.

before this PR:
```
Server fi_senddata fruitful cq progress p1 average ns: 1041.468177699095, pstd: 126.21655460969343, min: 29, max: 5220, median: 1000, num_samples: 1998457, unique: 428, num_outliers removed (5x mean): 1543
```

after this PR:
```
Server fi_senddata fruitful cq progress p1 average ns: 829.3492229892696, pstd: 105.98675207435754, min: 29, max: 3140, median: 800, num_samples: 1998428, unique: 364, num_outliers removed (5x mean): 1572
```

### DIFF
--- a/prov/util/src/util_cq.c
+++ b/prov/util/src/util_cq.c
@@ -737,7 +737,7 @@ int ofi_cq_init(const struct fi_provider *prov, struct fid_domain *domain,
 		goto destroy1;
 
 	/* TODO Figure out how to optimize this lock for rdm and msg endpoints */
-	ret = ofi_genlock_init(&cq->ep_list_lock, OFI_LOCK_MUTEX);
+	ret = ofi_genlock_init(&cq->ep_list_lock, OFI_LOCK_SPINLOCK);
 	if (ret)
 		return ret;
 


a-szegel:
Sean Hefty changed this from a spinlock to a mutex in this commit: https://github.com/ofiwg/libfabric/commit/209c9a2f13fc0e267165624e338208e9f8b6d627.

aingerson:
@a-szegel The cq->ep_list_lock protects the entire loop for iterating for all eps and driving progress. That's a lot of locked code that can include kernel calls, etc as Sean mentions in the commit message that turned it into a mutex. It might result in a 200 ns improvement for efa (in one use case or some, not sure what you ran) but I don't think it's wise to use a spin lock for all providers' progress calls.

shijin-aws:
Can we make the lock type configurable?

a-szegel:
PR that set it to mutex: https://github.com/ofiwg/libfabric/pull/7267
Original issue: https://github.com/ofiwg/libfabric/issues/7053

aingerson:
After reading the issue @a-szegel linked to, I think you'll have to be really careful with using spin locks in threaded applications. Do you have any concerns about this? Does replacing with a spin lock help in every application, including threaded ones?
@shijin-aws If configuring that lock to be a spin lock is something you really want for efa then I don't have any huge objections. It's not super invasive for other providers.
@j-xiong Do you have any thoughts on this? (he is out of office for this and the next week so he will be slow in responding)

shijin-aws:
@aingerson we are focusing on NCCL which use FI_THREAD_SAFE threading model. 

@a-szegel can you evaluate the perf improvement by running nccl-tests?

skypexu:
userland spinlock is not good for multiple-threads application.

PR #9666, opened 2023-12-22
prov/efa: Distinguish unresponsive receiver errors

darrylabbate:
> [!NOTE]
> Still need to add/reconfigure unit tests. This is mainly up for evaluation purposes for now.

This makes a best-effort attempt to deduce the connection health of an RDM endpoint when RDMA Core reports an unresponsive receiver error. Obviously this can still produce false negatives - unresponsive receiver can still be reported before a successful handshake even if the EC2 cluster is properly configured, but I feel this is still a step in the right direction.

This PR includes a number of other refactors, notably an `efa_show_help()` function as a dedicated method for printing long-form messages to the user. As noted in the Doxygen `@todo` comment, this should probably eventually mimic Open MPI's help functionality, where longer help messages are read from text files.

Comments welcome on the verbiage of the long-form help messages; particularly for the EC2-related help. I feel it's ultimately best to point users to the public docs, but embedding a URL is probably less future-proof than outlining the exact requirements.


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -711,7 +712,7 @@ void efa_rdm_txe_handle_error(struct efa_rdm_ope *txe, int err, int prov_errno)
 			? (const char *) err_entry.err_data
 			: efa_strerror(err_entry.prov_errno),
 		err_entry.prov_errno);
-
```
shijin-aws, line 736: nit: add the line back

shijin-aws (approved):
Awesome PR!

prov/efa/src/efa_strerror.c
```diff
@@ -1,134 +1,91 @@
...
+		break;
+	case EFA_IO_COMP_STATUS_LOCAL_ERROR_UNRESP_REMOTE:
+		help = "This error is detected locally; "
+		"The connection status is unknown or was never established via "
+		"handshake. This typically indicates one or more misconfigured "
+		"EC2 instances; most often due to incorrect inbound/outbound "
+		"security group rules. Refer to the public AWS documentation "
+		"for EFA for up-to-date configuration requirements. "
+		"This error can also be encountered when a peer process is no "
+		"longer present.";
```
wenduwan, line 79: Could you also add a note about subnet constraint? An internal customer hit this error last week trying to communicate across subnets.

prov/efa/src/rdm/efa_rdm_ep.h
```diff
@@ -1,35 +1,5 @@
...
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
- * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
- * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
- * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
- * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
- */
+/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
```
a-szegel, line 1: Can we get one commit that goes across the provider and updates all the headers?

prov/efa/src/efa_strerror.c
```diff
@@ -1,134 +1,91 @@
...
+		break;
+	case FI_EFA_ERR_ESTABLISHED_RECV_UNRESP:
+		help = "This error is detected locally. "
+		"The connection was previously established via handshake, "
+		"which indicates the error is likely due to the peer process no "
+		"longer being present.";
+		break;
+	default:
```
shijin-aws, line 87: Can you add some another item for `FI_EFA_REMOTE_ERROR_BAD_ADDRESS`? It means `/* Destination ENI is down or does not run EFA */` but we currently only say `invalid address`, which is not very intuitive.

shijin-aws:
@darrylabbate can we remove "draft" now? It seems CI is good (there is a socket provider failure which is irrelevant)

shijin-aws:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_ep.h
darrylabbate: Sean Hefty suggested I do the whole codebase when I asked a while back, but writing a script that handled all the different edge/corner cases wrt copyright statements is kind of a chore. Plus I'd rather not muddy up each file's history with (purely) administrative commits.

prov/efa/src/efa_strerror.c
darrylabbate: Would you rather change the base error string to `Destination ENI is down or does not run EFA`? There's no reason it needs to stay as `Invalid address`.

shijin-aws: That works too

shijin-aws (approved):

shijin-aws:
@wenduwan @a-szegel can you review again

wenduwan (approved):

Issue #9665, opened 2023-12-22
Change Control interfaces Thread Safeness for RDM EP's


a-szegel:
The [libfabric domain man pages](https://ofiwg.github.io/libfabric/main/man/fi_domain.3.html) currently make the following statement:
>  Control interfaces are always considered thread safe, and may be accessed by multiple threads.

I would like to change this statement to be:

>  Control interfaces are always considered thread safe for MSG EP's, and may be accessed by multiple threads.  For RDM EP's, control interfaces follow the threading model of the domain.

This would allow us to have the `cq->ep_list_lock`, `av->ep_list_lock`, and `cntr->ep_list_lock` default to no-op when the user requests FI_THREAD_DOMAIN.  

I want this b/c when I commented out `cq->ep_list_lock` [here](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_cq.c#L522-L529), and ran allreduce NCCL tests on a p5, I saw the avg bus bandwidth increase by 1.3%.

I understand this changes the API, and has the potential to break applications.

shijin-aws:
I prefer we define a new threading level as FI_THREAD_CTRL to refine the access to control interface. @j-xiong do you have any feedback on this?

j-xiong:
No, we don't want new threading levels. Instead, part of the 2.0 API effort is to reduce the number of threading levels.

I agree this is unrelated to EP type. It's more about if the application needs the control path to be thread safe. A specific application being able to run correctly w/o the locking doesn't mean other applications can do the same. The default behavior should always take correctness as higher priority over performance. 

I would prefer a provider specific option for optimizing control path locks.

j-xiong:
For example, via the fi_set_val() API.

a-szegel:
From my perspective, FI_THREAD_DOMAIN should mean that only one thread per domain is allowed to make **ALL** libfabric API calls.  Adding a thread safe control API adds unnecessary complexity in order to optimize the performance of connection based msg endpoints.  [This is related to EP type, b/c this decision was made to optimize a specific EP type.](https://github.com/ofiwg/libfabric/pull/9335#issuecomment-1733944490)

I would prefer to change the API to say:

"FI_THREAD_DOMAIN means 1 thread per domain at a time", and have apps that were using the previous threading model be forced to claim FI_THREAD_SAFE.

[...]

PR #9667, opened 2023-12-22
prov/efa: Remove FI_THREAD_SAFE control interface from EFA (stopping the EFA provider from following the Libfabric API)

a-szegel:
This patch improves the performance of the EFA provider by removing locking in the CQ progress path.  It also causes the EFA provider to not comply with the Libfabric API promise of offering thread safe control interfaces.  When a user specifies FI_THREAD_DOMAIN, we will make the control interfaces thread safe to the level of FI_THREAD_DOMAIN. It is common for message endpoints to have a separate thread for just control calls, and those endpoints still want to claim FI_THREAD_DOMAIN, instead of FI_THREAD_SAFE for performance reasons. The EFA endpoint is an RDM endpoint, so the use case of a thread for the control API is not an expected use case.  I created an issue for adjusting the libfabric API https://github.com/ofiwg/libfabric/issues/9665 to make RDM points not maintain the FI_THREAD_SAFE control interface promise. Those changes might take a long time (if they ever) get passed the community. This change gets the EFA Provider the performance boost immediately.
    
### DIFF
--- a/man/fi_efa.7.md
+++ b/man/fi_efa.7.md
@@ -89,6 +89,11 @@ requires peer to peer transaction support between the EFA and the FI_HMEM
 device. Therefore, the FI_HMEM_P2P_DISABLED option is not supported by the EFA
 provider for AWS Neuron or Habana SynapseAI.
 
+*FI_THREAD_DOMAIN*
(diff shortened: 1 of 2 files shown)


a-szegel:
Got some feedback that I should try to focus on the long term solution.

PR #9671, opened 2023-12-27
fabtests/pytest: add a new ssh connection error pattern

wenduwan:
Same as title


wenduwan:
Intel CI does not use pytest so ignoring the failure.

a-szegel (approved):

sunkuamzn (approved):

PR #9672, opened 2023-12-27
prov/efa: Make msg path slightly more efficient

a-szegel:
This commit makes a few changes to make the msg path slightly more efficient.

1. Stop initializing msg to 0  in recv path b/c efa_rdm_msg_construct fully init's struct fi_msg
2. Make efa_rdm_msg_recv() and efa_rdm_msg_recvv() directly call efa_rdm_msg_generic_recv()
3. Make efa_rdm_msg_generic_recv() and efa_rdm_msg_generic_recv() an inline function
4. Pass in efa_rdm_ep in send/recv path to make it so we do not need duplicate calls to container_of()


wenduwan (approved):

a-szegel:
Data shows this saves us 1ns.

wenduwan (approved):

PR #9680, opened 2023-12-28
prov/efa: Reduce # of error cases in happy path

a-szegel:
This patch uses nested if cases to reduce the number of error checks in the fast happy path from 3 to 1 inside
efa_rdm_pke_handle_rtm_rta_recv(). It also cleans up some extra spacing in comments.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke_rtm.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_rtm.c
@@ -57,7 +57,7 @@
  * @details
  * A RTM packet is sent/received for an user's message, this function
  * return the total length of that message
- * 
+ *
  * @param[in]	pkt_entry	RTM packet entry
  *
  * @returns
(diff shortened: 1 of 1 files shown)


wenduwan (approved):

prov/efa/src/rdm/efa_rdm_pke_rtm.c
```diff
@@ -492,31 +492,30 @@ void efa_rdm_pke_handle_rtm_rta_recv(struct efa_rdm_pke *pkt_entry)
...
-		/* Packet with same msg_id has been processed before */
-		EFA_WARN(FI_LOG_EP_CTRL,
-			"Invalid msg_id: %" PRIu32
-			" robuf->exp_msg_id: %" PRIu32 "\n",
-		       msg_id, peer->robuf.exp_msg_id);
-		efa_base_ep_write_eq_error(&ep->base_ep, FI_EIO, FI_EFA_ERR_PKT_ALREADY_PROCESSED);
-		efa_rdm_pke_release_rx(pkt_entry);
-		return;
-	}
+	} else if (OFI_UNLIKELY(ret < 0)) {
```
shijin-aws, line 501: no need to use `else if`, just `if`

PR #9681, opened 2023-12-28
fabtests: Add comments to match initial send and receive

sunkuamzn:
Fabtests posts a receive in ft_enable_ep_receive that's called from ft_init_fabric. The corresponding send for this receive is posted in ft_finialize_ep. Add comments in both functions to explain the matching.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -1430,7 +1430,8 @@ int ft_enable_ep_recv(void)
 
 	if (!ft_check_opts(FT_OPT_SKIP_MSG_ALLOC) &&
 	    (fi->caps & (FI_MSG | FI_TAGGED))) {
-		/* Initial receive will get remote address for unconnected EPs */
+		/* Initial receive will get remote address for unconnected EPs*/
+		/* The corresponding send for this receive is posted in ft_finalize_ep */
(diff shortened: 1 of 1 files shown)


a-szegel (approved):

fabtests/common/shared.c
```diff
@@ -1430,7 +1430,8 @@ int ft_enable_ep_recv(void)
 
 	if (!ft_check_opts(FT_OPT_SKIP_MSG_ALLOC) &&
 	    (fi->caps & (FI_MSG | FI_TAGGED))) {
-		/* Initial receive will get remote address for unconnected EPs */
+		/* Initial receive will get remote address for unconnected EPs*/
```
shijin-aws, line 1433: I don't think the first receive always get remote address. The address can be exchanged with OOB. The first rx just matches whatever the first tx

j-xiong: Agree. This just preposts a receive. It doesn't necessarily match the send posted in ft_finalize_ep. IMO, these comments are inaccurate and unnecessary.

sunkuamzn: @shijin-aws @j-xiong do you want me to just delete the comment on line 1433? Or do you also not like the new comments I added?

sunkuamzn: Closed PR

shijin-aws: I would just close it, yes

PR #9696, opened 2024-01-04
fabtests/pytest: add another ssh connection error pattern

sunkuamzn:


shijin-aws (approved):

fabtests/pytest/common.py
```diff
@@ -27,6 +27,7 @@ def is_ssh_connection_error(exception):
 
 def has_ssh_connection_err_msg(output):
     err_msgs = ["Connection closed by remote host",
+                r"Connection closed by .* port 22",
```
shijin-aws, line 30: I am guessing we can just use `Connection closed`

sunkuamzn: Thanks. Fixed.

PR #9706, opened 2024-01-10
man: Regulate the fi_setopt call sequence.

shijin-aws:
fi_setopt adjust the protocol and implementation
details of an endpoint, so it must be called before the endpoint is enabled via fi_enable.


j-xiong (approved):

aingerson (approved):
Thank you!

PR #9710, opened 2024-01-10
intel/ci: exclude threads/comm/idup_nb test mpichsuite tests for tcp.

nikhilnanal:
The idup_nb test is showing occasional failures for tcp provider. It is excluded until resolution.


zachdworkin (approved):

zachdworkin:
Change is to Intel CI not AWS CI so AWS CI completion is not needed.

PR #9712, opened 2024-01-10
[v1.19.x] contrib/intel/jenkins: Pick CI architecture changes

zachdworkin:
Pick CI architecture changes from PR-9638

### DIFF
--- a/contrib/intel/jenkins/tests.py
+++ b/contrib/intel/jenkins/tests.py
@@ -15,11 +15,13 @@ import time
 # it is building for. for e.g. in our case jobname = 'ofi_libfabric/master'
 class Test:
 
-    def __init__ (self, jobname, buildno, testname, core_prov, fabric,
-                  hosts, ofi_build_mode, user_env, log_file, mpitype=None, util_prov=None):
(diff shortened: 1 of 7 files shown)


contrib/intel/jenkins/tests.py
```diff
@@ -879,7 +887,7 @@ def __init__(self, jobname, buildno, testname, hw, core_prov, fabric,
                                 'sycl_allreduce_inplace_usm_test',
                                 'sycl_allreduce_test',
                                 'sycl_allreduce_usm_test',
-                                'sycl_alltoall_test',
+                                'sycl_allto ',
```
j-xiong, line 890: Is it an accidental change?

zachdworkin: yes

zachdworkin:
@j-xiong I just added a patch that applies the architecture changes and also leaves the ze-shm stage with using the wrapper script. Pulling in all those extra patches was causing too many other failures. I will remove the "do not merge" and "evaluating" labels when the CI passes. When this passes I will do the same to the v1.18.x cherrypicks as well.

PR #9714, opened 2024-01-10
hmem/ze: Increase ZE_MAX_DEVICES to 32

j-xiong:
We have already seen setups with 16 ZE devices. Increase the maximum to 32 to leave some room for future growth.

Ultimately we will need to remove this hard limit and allocated the resources dynamicaly. That can be done later.

Also change the behavior for the case when the limit is exceeded.


j-xiong:
@aingerson

aingerson (approved):

PR #9724, opened 2024-01-12
src/hmem_ze, prov/shm: dynamically allocate device resource, create mapping between physical devices and user space devices

aingerson:
-dynamically allocate device info and resource to remove hard coded maximum
-create mapping between physical devices and user space devices to fix shm IPC fd mapping
-fix copy engine environment variable

### DIFF
--- a/src/hmem_ze.c
+++ b/src/hmem_ze.c
@@ -773,7 +773,7 @@ int ze_hmem_init(void)
 	fi_param_define(NULL, "hmem_ze_copy_engine", FI_PARAM_STRING,
                         "Specify GPU engine used for copy operation: <group>, "
 			"<group>.<index> (default: 1st copy-only engine)");
-        fi_param_get_str(NULL, "log_level", &enginestr);
+        fi_param_get_str(NULL, "hmem_ze_copy_engine", &enginestr);
 	if (enginestr)
 		sscanf(enginestr, "%d.%d", &ordinal, &index);
 
--- a/src/hmem_ze.c
+++ b/src/hmem_ze.c
@@ -46,18 +46,22 @@
 #include <sys/ioctl.h>
 #include <linux/dma-buf.h>
 
+struct ofi_ze_dev_info {
+	ze_command_queue_handle_t cmd_queue;
+	int ordinal;
+	int index;
+	ze_device_uuid_t uuid;
+	struct ofi_bufpool *cl_pool;
+};
+
 static ze_driver_handle_t driver;
 static ze_context_handle_t context;
-static ze_device_handle_t devices[ZE_MAX_DEVICES];
-static ze_command_queue_handle_t cmd_queue[ZE_MAX_DEVICES];
+static ze_device_handle_t *devices = NULL;
+static struct ofi_ze_dev_info *dev_info = NULL;
 static int num_devices = 0;
-static int ordinals[ZE_MAX_DEVICES];
-static int indices[ZE_MAX_DEVICES];
-static int dev_fds[ZE_MAX_DEVICES];
-static ze_device_uuid_t dev_uuids[ZE_MAX_DEVICES];
+static int *dev_fds = NULL;
 static bool p2p_enabled = false;
 static bool host_reg_enabled = true;
-static struct ofi_bufpool *cl_pool[ZE_MAX_DEVICES];
 ofi_spin_t cl_lock;
 
 static ze_command_queue_desc_t cq_desc = {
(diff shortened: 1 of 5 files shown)


shijin-aws:
Not related to the PR, but I am thinking whether we can make shm max peers configurable and smr_region dynamically allocated.. Currently this happens in ep enable so I think we should be able to make it configured via fi_setopt..

src/hmem_ze.c
```diff
@@ -48,6 +48,7 @@
 
 struct ofi_ze_dev_info {
 	ze_command_queue_handle_t cmd_queue;
+	int device;
```
j-xiong, line 51: Could use a different name to avoid confusion. For example, pci_device.

src/hmem_ze.c
```diff
@@ -443,11 +446,12 @@ int ze_hmem_get_shared_handle(int dev_fd, void *dev_buf, int *ze_fd,
...
 }
 
-int ze_hmem_open_shared_handle(int dev_fd, void **handle, int *ze_fd,
-			       uint64_t device, void **ipc_ptr)
+int ze_hmem_open_shared_handle(uint64_t device, int *peer_fds, void **handle,
+			       int *ze_fd, void **ipc_ptr)
 {
 	struct drm_prime_handle open_fd = {0, 0, 0};
 	ze_ipc_mem_handle_t ze_handle;
+	int dev_fd = peer_fds[dev_info[device].device];
```
j-xiong, line 454: Here is an example of how the two `device` terms can get confusing.

src/hmem_ze.c
```diff
@@ -849,6 +854,10 @@ int ze_hmem_init(void)
 
 		dev_info[num_devices].uuid = dev_prop.uuid;
 
+		if (!dev_prop.subdeviceId)
+			pci_dev++;
+		dev_info[num_devices].device = pci_dev;
```
j-xiong, line 859: This may not work as expected when ZE_AFFINITY_MASK is set.

j-xiong: can find the correct pci device by comparing uuid.

aingerson: I'm not really sure how to find it with the uuid. That's a 16 byte uint8 array that seems to be unique for each tile. For example on my system:
```
{id = "\206\200\326\v/\000\000\000:\000\000\000\000\000\000\001"}
{id = "\206\200\326\v/\000\000\000:\000\000\000\000\000\000\002"}
{id = "\206\200\326\v/\000\000\000\232\000\000\000\000\000\000\001"}
{id = "\206\200\326\v/\000\000\000\232\000\000\000\000\000\000\002"}
```

aingerson:
@shijin-aws Totally! It's on my list of things to do and I'm planning on working it into my new shm changes :)

src/hmem_ze.c
j-xiong: The bytes around the 9th byte are the pci address of the device. Not sure where is the start and end.
Note that ":" is 0x3a and "\232" is 0x9a. The pci addresses are `0000:3a:00.0` and `0000:9a:00.0`.

aingerson: Right I see that. But how do I map ":" and "\232" to 0 and 1? Is there a way to lookup all ZE PCI devices (regardless of the affinity mask)?

aingerson: @j-xiong See update which matches it based on the PCI bus ID. This isn't the prettiest thing ever but it does the job...

src/hmem_ze.c
```diff
@@ -403,26 +407,36 @@ static int ze_hmem_init_fds(void)
...
+		if (dev_fds[num_pci_devices] == -1)
 			goto err;
-		i++;
+		uuid_str = strtok(ent->d_name, ":");
+		uuid_str = strtok(NULL, ":");
+		uuid = (uint8_t) strtol(uuid_str, NULL, 16);
+
+		for (i = 0; i < num_devices; i++) {
+			if (dev_info[i].uuid.id[8] == uuid)
+				dev_info[i].pci_device = num_pci_devices;
```
j-xiong, line 419: For the full pci address `domain:bus:dev.func`, the `domain` part needs to be considered as well. For example, here is a list of pci addresses on an actual machine with 6 PVCs: `0000:18:00.0`, `0000:42:00.0`, `0000:6c:00.0`, `0001:18:00.0`, `0001:42:00.0`, `0001:6c:00.0`.

aingerson: @j-xiong Updated. Do the dev and func need to be considered as well?

j-xiong: So far dev and func are always 0 and can be ignored.

j-xiong (approved):

PR #9725, opened 2024-01-12
Add FI_PROGRESS_CONTROL_UNIFIED

a-szegel:


man/fi_domain.3.md
```diff
@@ -261,6 +261,11 @@ interfaces enables a provider to eliminate lower-level locks.
 : A domain serialization model requires applications to serialize
   access to all objects belonging to a domain.
 
+*FI_THREAD_DOMAIN_RELAXED*
+: A domain serialization model requires applications to serialize
+  access to all objects belonging to a domain. Control interfaces
+  are not thread safe.
```
j-xiong, line 267: May also add to the start of L241:  `Unless otherwise noted, ` or `Except for FI_THREAD_DOMAIN_RELAXED`

fabtests/benchmarks/dgram_pingpong.c
```diff
@@ -129,7 +129,7 @@ int main(int argc, char **argv)
 	hints->caps = FI_MSG;
 	hints->mode |= FI_CONTEXT;
 	hints->domain_attr->mr_mode = opts.mr_mode;
-	hints->domain_attr->threading = FI_THREAD_DOMAIN;
+	hints->domain_attr->threading = FI_THREAD_DOMAIN_RELAXED;
```
j-xiong, line 132: All providers don't support this threading model. Drop the changes or make the new threading model an option.

prov/util/src/util_av.c
```diff
@@ -553,7 +553,9 @@ int ofi_av_init_lightweight(struct util_domain *domain, const struct fi_av_attr
 	av->context = context;
 	av->domain = domain;
 
-	ret = ofi_genlock_init(&av->ep_list_lock, OFI_LOCK_MUTEX);
+	ret = ofi_genlock_init(&av->ep_list_lock,
+			       av->domain->threading != FI_THREAD_DOMAIN_RELAXED ?
+			       OFI_LOCK_MUTEX : OFI_LOCK_NOOP);
```
j-xiong, line 558: Add indentation to this line. Same to other occurrences.

prov/efa/src/efa_domain.c
```diff
@@ -213,8 +214,9 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	efa_domain->fabric = container_of(fabric_fid, struct efa_fabric,
 					  util_fabric.fabric_fid);
 
-	err = ofi_domain_init(fabric_fid, info, &efa_domain->util_domain,
-			      context, OFI_LOCK_MUTEX);
+	domain_lock_type = info->domain_attr->threading != FI_THREAD_DOMAIN_RELAXED ?
+			   OFI_LOCK_MUTEX : OFI_LOCK_NOOP;
```
j-xiong, line 218: Add indentation here.

fabtests/benchmarks/dgram_pingpong.c
a-szegel: Why don't all providers support the new threading model?

j-xiong: Some providers (e.g. opx) doesn't recognize this new value and will give runtime error. Some other providers accept this value but won't perform intended lock optimization. Until all providers have been changed to take act properly on this threading model, let's keep this as opt-in.

shijin-aws: Agree with j-xiong. This should be an opt-in model

shijin-aws: @j-xiong a related topic: I find fabtests always use `FI_TC_LOW_LATENCY` in tclass but it's currently ignored by all providers. Should we remove it in fabtests as the default tclass?

a-szegel: I searched for FI_THREAD_DOMAIN in the code, and it doesn't seem like providers do specific actions for it.

j-xiong: @shijin-aws Yes, I think that makes sense.

prov/efa/src/efa_prov_info.c
```diff
@@ -102,7 +102,7 @@ int efa_prov_info_set_fabric_attr(struct fi_info *prov_info, struct efa_device *
  */
 const struct fi_domain_attr efa_domain_attr = {
 	.caps			= EFA_DOMAIN_CAPS,
-	.threading		= FI_THREAD_DOMAIN,
+	.threading		= FI_THREAD_DOMAIN_RELAXED,
```
shijin-aws, line 105: We shouldn't change it here. Two reasons

1. For rdm ep type, it won't make any impact. efa finally changes it to FI_THREAD_SAFE as the default provider threading model: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_prov_info.c#L583-L587
2. Every provider must support FI_THREAD_SAFE. If we make it default to use FI_THREAD_DOMAIN, and application requests FI_THREAD_SAFE, this call will fail https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_attr.c#L563-L567

a-szegel:
We are still talking about the RCU change internally.

man/fi_efa.7.md
```diff
@@ -66,11 +66,6 @@ The following features are supported:
   this bug would break those applications. This will be fixed in a future version
   of the EFA provider by adding proper support for *FI_PROGRESS_AUTO*.
 
-*Threading*
-: The RDM endpoint supports *FI_THREAD_SAFE*, the DGRAM endpoint supports
```
shijin-aws, line 70: We shouldn't change this? RDM endpoints should always support FI_THREAD_SAFE. I don't know if there is any application is using DGRAM endpoint + FI_THREAD_DOMAIN, but it's better to keep the default threading level unchanged to avoid breaking any. As we talked about, FI_THREAD_DOMAIN_RELAXED should be an opt-in

shijin-aws (approved):

j-xiong (approved):

a-szegel:
Closing this for now, and I am going to try and put up a custom RCU solution.  Will re-open if that fails.

a-szegel:
@shefty What did you want this called instead of `FI_THREAD_DOMAIN_RELAXED`?

shefty:
The man page changes suggest something which means "control threading is unsafe" or "data and control threads are unified".  FI_THREAD_DOMAIN_UNSAFE or FI_THREAD_UNIFIED_DOMAIN ?  I'm leaning towards the unified meaning.  But the threading model still needs to somehow also be aligned with the progress model, which still has data and control separate.  The threading field is not necessarily sufficient to eliminate locking, depending on the progress models which are selected. That's why I think both need to be addressed, even if they use different solutions.

a-szegel:
I like UNSAFE b/c unified doesn't capture no thread safe control interface... going to update with my understanding of the requested changes to see if I understood correctly.

bwbarrett:
How is it actually unsafe?  I think that's terribly misleading.

shefty:
The 'unsafe' is that the proposed option removes thread safety from the control APIs. I prefer 'unified', but don't like it.  I just haven't come up with a term that I like better.

j-xiong:
Among the three, I still prefer FI_THREAD_DOMAIN_RELEAXED. It basically means FI_THREAD_DOMAIN, with the requirement "control be thread-safe" relaxed. The other two, while capturing some information about the control threading model (unsafe, or unified with data threading model), tend to be more misleading.

Maybe we should reconsider the approach of adding a separate "control_threading" variable. This changes the ABI, but is more understandable.

bwbarrett:
Sean, I was mostly poking Seth there, but the point is that if you name something `FI_THREAD_DOMAIN_UNSAFE`, someone without all this history is going to read it as an unsafe option.  Which it really isn't.  It's saying that a subset of operations aren't safe in the way that they are in `FI_THREAD_DOMAIN`, which arguably was a weird definition of thread safety.  I tend to agree with @j-xiong on RELAXED, but also think the big thing is just to fix this properly in 2.0 when ABI can be mucked with a bit more.

a-szegel:
pushed before I refreshed and saw all the new discussion...

a-szegel:
After reading the discussion, I think the best option is to stick with FI_THREAD_DOMAIN_RELAXED, and take the additional changes I made to the if condition of removing the lock to only remove the locks if the progress is manual.  Does everyone agree?

shefty:
A separate variable isn't what's desired.  What the provider wants to know is that the app is using a single thread to access all objects of a specific container space (in this case the domain).  It doesn't matter what the API is that is called.  That is, no distinction is made between 'control' APIs and 'data' APIs.  It was trying to define such a distinction that led to this mess.

The word we want should mean "no distinction is made between control and data APIs".  That's the viewpoint that the user has.  (Single threaded is conveyed by the threading variable, and domain indicates the container space).

And I still think this is leaving a gap with the control_progress and data_progress variables.

shefty:
As a completely different approach, leave enum fi_threading unchanged.  Instead, add a new enum fi_progress value that applies only to the control_progress field (by definition).  The new value would indicate that control progress is handled the same as data progress.  In the case of FI_PROGRESS_MANUAL, the same thread(s) are used, which aligns the control threading model with the selected threading model.

a-szegel:
@j-xiong Are you ok with FI_PROGRESS_CONTROL_UNIFIED?  I pushed the new changes.  AWS and Sean are both good with it.

a-szegel:
@shefty @shijin-aws @j-xiong please re-review when you have time.

man/fi_domain.3.md
```diff
@@ -375,6 +378,15 @@ are defined.
   manual progress may still need application assistance to process
   received operations.
 
+*FI_PROGRESS_CONTROL_UNIFIED*
+: This progress model indicates that the user will synchronize progressing the
+  data and control operations themselves (i.e.. this makes control interface NOT
+  thread safe). It is only valid for control progress (not data progress). If
+  the control progress is set to FI_PROGRESS_CONTROL_UNIFIED, the data progress
+  must be set to FI_PROGRESS_MANUAL. Setting control=FI_PROGRESS_CONTROL_UNIFIED,
```
shefty, line 386: I don't think this restriction is needed.  Progress can still be auto, and the provider can decide how to deal with that.

prov/util/src/util_attr.c
```diff
@@ -454,8 +454,10 @@ static int fi_progress_level(enum fi_progress progress_model)
 		return 1;
 	case FI_PROGRESS_MANUAL:
 		return 2;
-	case FI_PROGRESS_UNSPEC:
+	case FI_PROGRESS_CONTROL_UNIFIED:
```
shefty, line 457: Did you verify that users of this function still work properly after this change, without needing changes at the calling location?

shefty (commented):
I'm happy with this.  Couple of minor comments.

man/fi_domain.3.md
a-szegel: So you are saying both `data_progress == AUTO && control_progress == UNIFIED` should somehow work?  I think a better way of approaching that situation is `data_progress == auto && control_progress == auto` and letting the provider figure out how to deal with that.

I don't see how it is possible for  `data_progress == AUTO && control_progress == UNIFIED` to be a valid configuration.

prov/util/src/util_attr.c
a-szegel: I see the issue. Thanks... pushing fix.

shijin-aws (approved):

man/fi_domain.3.md
shefty: control = UNIFIED still means that calls to the control and data APIs are serialized by the app.  Setting both to AUTO doesn't convey that.  Do I expect an app to use that combo?  No.  But that's also true for control=manual & data=auto, which is still an allowed API option.

If a provider offloads everything, control=UNIFIED and data=AUTO may be a valid pairing for it to return.  That's probably what the UDP provider can support and still be lockless.

prov/util/src/util_attr.c
a-szegel: All the users of `fi_progress_level()` in main today are inside util_attr.c:
```
(env) ┌─[ec2-user@ip-172-31-22-242]─[1774:23h59]─[~/libfabric]
└─╼ grep -rI fi_progress_level
prov/util/src/util_attr.c:static int fi_progress_level(enum fi_progress progress_model)
prov/util/src/util_attr.c:	if (fi_progress_level(user_attr->control_progress) <
prov/util/src/util_attr.c:	    fi_progress_level(prov_attr->control_progress)) {
prov/util/src/util_attr.c:	if (fi_progress_level(user_attr->data_progress) <
prov/util/src/util_attr.c:	    fi_progress_level(prov_attr->data_progress)) {
```

The four usages are here:
```
	if (fi_progress_level(user_attr->control_progress) <
	    fi_progress_level(prov_attr->control_progress)) {
		FI_INFO(prov, FI_LOG_CORE, "Invalid control progress model\n");
		return -FI_ENODATA;
	}

	if (fi_progress_level(user_attr->data_progress) <
	    fi_progress_level(prov_attr->data_progress)) {
		FI_INFO(prov, FI_LOG_CORE, "Invalid data progress model\n");
		return -FI_ENODATA;
	}
```

We should block FI_PROGRESS_CONTROL_UNIFIED in data progress check here.

It works correctly b/c providers are either specifying FI_PROGRESS_AUTO, (1) or FI_PROGRESS_MANUAL (2) as the level of supported automation.  B/c FI_PROGRESS_CONTROL_UNIFIED returns 3, all providers will accept the threading level without need for modification.

man/fi_domain.3.md
```diff
@@ -375,6 +378,15 @@ are defined.
   manual progress may still need application assistance to process
   received operations.
 
+*FI_PROGRESS_CONTROL_UNIFIED*
+: This progress model indicates that the user will synchronize progressing the
+  data and control operations themselves (i.e.. this makes control interface NOT
+  thread safe). It is only valid for control progress (not data progress). If
```
j-xiong, line 384: I would rephrase this to `this allows control interface to be NOT thread safe`.

j-xiong (commented):
I don't like name very much, but can't come up with anything better.

shefty:
@j-xiong I use a thesaurus WAY more with coding than I ever thought I would.  Unified - bonded, connected, paired, linked, aligned, identical, combined... or we go the OpenMPI route and call it something like Spock.

man/fi_domain.3.md
a-szegel: reworded

man/fi_domain.3.md
```diff
@@ -238,10 +238,13 @@ The name of the access domain.
...
-accessed by multiple threads.  Applications which can guarantee
-serialization in their access of provider allocated resources and
-interfaces enables a provider to eliminate lower-level locks.
+Control interfaces are always considered thread safe unless the
+control progress model is FI_PROGRESS_CONTROL_UNIFIED. A thread safe
+control interface allows multiple threads to progress the control
+interface, and (depending on threading model selected) one or more
+threads to progress the data interfaces at the same time. Applications
+which can guarantee serialization in their access of provider allocated
+resources and interfaces enables a provider to eliminate lower-level locks.
```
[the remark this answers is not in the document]
a-szegel: fixed

man/fi_domain.3.md
a-szegel: Removed the sentence, and the if check in util_domain.c

prov/util/src/util_av.c
```diff
@@ -553,7 +553,11 @@ int ofi_av_init_lightweight(struct util_domain *domain, const struct fi_av_attr
 	av->context = context;
 	av->domain = domain;
 
-	ret = ofi_genlock_init(&av->ep_list_lock, OFI_LOCK_MUTEX);
+	ret = ofi_genlock_init(&av->ep_list_lock,
+			       (av->domain->threading == FI_THREAD_DOMAIN ||
+			       av->domain->threading == FI_THREAD_COMPLETION) &&
+			       av->domain->control_progress == FI_PROGRESS_CONTROL_UNIFIED ?
+					OFI_LOCK_NOOP : OFI_LOCK_MUTEX);
```
shefty, line 560: Please separate out this set of checks (repeated 3 times) into a function somewhere.  E.g. ofi_lockless_progress()?  I'm unsure of the name.

Otherwise, I think this works, is still simple to use, and has minimal impact on providers.

a-szegel: done

a-szegel:
Shi just told me another idea for EFA provider to get rid of the lock without using this change.  If we choose to go down this path, then we would not need this.  I am hesitant to add a new feature into the API, and then immediately not use it.  Marked this PR as evaluating and do not merge again.

a-szegel:
We will use this even with @shijin-aws's EFA progress engine future changes.

a-szegel:
bot:aws:retest

a-szegel:
libfabric CI was failing sockets provider, re-pulled main... if it fails again, I will root cause tomorrow.

a-szegel:
Seems like fi_msg_pingpong is hanging:
```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.37.187 'timeout 360 /bin/bash --login -c '"'"'/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr9842-debug/install/fabtests/bin/fi_msg_pingpong -I 5 -v -p sockets -s 172.31.37.187'"'"''
```

a-szegel:
Seems sockets provider is failing rdm_pingpong tests as well with manual progress, and control progress set to unified.

a-szegel:
If I change rdm_pingpong.c in main to have:
```
	hints->domain_attr->data_progress = FI_PROGRESS_MANUAL;
```
The sockets provider will hang in fi_rdm_pingpong.

fabtests/ubertest/config.c
```diff
@@ -369,6 +369,7 @@ static int ft_parse_num(char *str, int len, struct key_t *key, void *buf)
 		TEST_ENUM_SET_N_RETURN(str, len, FI_PROGRESS_MANUAL, int, buf);
 		TEST_ENUM_SET_N_RETURN(str, len, FI_PROGRESS_AUTO, int, buf);
 		TEST_ENUM_SET_N_RETURN(str, len, FI_PROGRESS_UNSPEC, int, buf);
+		TEST_ENUM_SET_N_RETURN(str, len, FI_PROGRESS_CONTROL_UNIFIED, int, buf);
```
aingerson, line 372: I think this should get removed so it can get isolated to the API change. These enums in ubertest are ones that the application can set in order to test it. If there's no support in ubertest to verify it works then we probably shouldn't add it to the enums yet.

include/ofi_util.h
```diff
@@ -1370,6 +1370,14 @@ static inline void ofi_cq_err_memcpy(uint32_t api_version,
...
 }
 
+static inline
+bool ofi_lockless_progress(enum fi_threading threading, enum fi_progress control)
+{
+	return (threading == FI_THREAD_DOMAIN || threading == FI_THREAD_COMPLETION) &&
+		control == FI_PROGRESS_CONTROL_UNIFIED ?
+			true : false;
+}
+
```
aingerson, line 1380: The double negative in the name here is confusing. You also use it in the exact same way every time (ie find the bool and then pick noop vs mutex. I think it would make more sense to combine and simplify:
```
static enum ofi_lock_type ofi_progress_lock_type(enum fi_threading threading, enum fi_progress control)
{
	return (threading == FI_THREAD_DOMAIN || threading == FI_THREAD_COMPLETION) &&
		control == FI_PROGRESS_CONTROL_UNIFIED ? OFI_LOCK_NOOP : OFI_LOCK_MUTEX;
}
```

fabtests/benchmarks/rdm_cntr_pingpong.c
```diff
@@ -104,6 +104,8 @@ int main(int argc, char **argv)
 	hints->caps = FI_MSG;
 	hints->domain_attr->mr_mode = opts.mr_mode;
 	hints->domain_attr->threading = FI_THREAD_DOMAIN;
+	hints->domain_attr->data_progress = FI_PROGRESS_MANUAL;
```
aingerson, line 107: Do we want to force all these tests to manual progress? What if we want to test auto progress? I think it makes more sense to leave data progress alone.

shijin-aws: +1, I talked with Seth and we agree this should be an opt-in feature and specified as a new argument

a-szegel: made it as an opt in feature.

fabtests/ubertest/config.c
a-szegel: done

include/ofi_util.h
a-szegel: done

fabtests/common/shared.c
```diff
@@ -4165,9 +4169,25 @@ struct option long_opts[] = {
...
+int ft_parse_progress_model_string(char* progress_str)
+{
+	int ret = -1;
+
+	if (!strncasecmp("FI_PROGRESS_MANUAL", progress_str, 19))
+		ret = FI_PROGRESS_MANUAL;
+	else if (!strncasecmp("FI_PROGRESS_AUTO", progress_str, 17))
+		ret = FI_PROGRESS_AUTO;
+	else if (!strncasecmp("FI_PROGRESS_CONTROL_UNIFIED", progress_str, 28))
+		ret = FI_PROGRESS_CONTROL_UNIFIED;
```
aingerson, line 4186: Are you including the NULL terminator in these? Is that correct/more correct? Asking because the rest of fabtests just does the length of the characters. Wondering if all of those are wrong

fabtests/common/shared.c
```diff
@@ -4165,9 +4169,25 @@ struct option long_opts[] = {
 	{"pin-core", required_argument, NULL, LONG_OPT_PIN_CORE},
 	{"timeout", required_argument, NULL, LONG_OPT_TIMEOUT},
 	{"debug-assert", no_argument, &debug_assert, LONG_OPT_DEBUG_ASSERT},
+	{"data-progress-model", required_argument, NULL, LONG_OPT_DATA_PROGRESS},
+	{"control-progress-model", required_argument, NULL, LONG_OPT_CONTROL_PROGRESS},
```
aingerson, line 4173: These feel like a mouthful. Not sure if there's a much better alternative, maybe shorten to --data_progress=manual / auto / unified?

man/fi_domain.3.md
```diff
@@ -375,6 +378,14 @@ are defined.
   manual progress may still need application assistance to process
   received operations.
 
+*FI_PROGRESS_CONTROL_UNIFIED*
+: This progress model indicates that the user will synchronize progressing the
+  data and control operations themselves (i.e.. this allows the control interface
```
aingerson, line 383: I'm going to be annoying and comment about this extra period. Sorry.

fabtests/common/shared.c
a-szegel: I am not including the NULL terminator.  Since I give the max length to compare (specified length, or NULL terminator of progress_str, whichever comes first), I don't think I need to include the NULL terminator.  I am also not doing anything with the progress_str after I parse it besides setting ret = ENUM, if the entire specified string matches.  In this case, I think we are fine.

man/fi_domain.3.md
a-szegel: done

fabtests/common/shared.c
a-szegel: also we are using [getopt_long](https://linux.die.net/man/3/getopt_long) which should always give us null terminated strings in optarg

a-szegel:
@aingerson @shijin-aws @shefty @j-xiong I think this is ready for merge.

fabtests/common/shared.c
aingerson: @a-szegel Sorry I should have clarified my comment more - in this code you are adding an extra char to compare, implying you should compare the NULL terminator but I agree, I don't think you need it - ie FI_PROGRESS_MANUAL is 18 characters but you're comparing 19, FI_PROGRESS_AUTO is 16 but you're comparing 17, etc

shefty: It's usually better to include the null terminator to match the string exactly.  Consider if the user passed in "FI_PROGRESS_MANUAL-v" and left the space out before the -v argument.  Without the null terminator check, this string passes and the -v gets lost without the user knowing that their mistake was ignored.

j-xiong: Any reason not to use `strcasecmp()` here? That should behave the same as `strncasecmp()` with the null included.

aingerson: I was more just asking about which convention we should adopt throughout fabtests. Every other use of it in fabtests uses the strncasecmp without the NULL character. I like Jianxin's suggestion of switching all of them to use strcasecmp which seems to handle it explicitly. @a-szegel Can you change this to strcasecmp? I can open a PR to standardize the rest of fabtests.

a-szegel: I was just doing `echo FI_PROGRESS_CONTROL_UNIFIED | wc` to count for me, and I guess that got the extra character

a-szegel: fixed.

aingerson (approved):
Looks good to me! @j-xiong I'll let you merge/approve when you feel it's good to go.

PR #9740, opened 2024-01-18
prov/util: Add name field to memory monitors:

muttormark:
Log when monitor states are changed.
    Log name of default monitor.

### DIFF
--- a/prov/util/src/util_mem_monitor.c
+++ b/prov/util/src/util_mem_monitor.c
@@ -56,6 +56,7 @@ static struct ofi_uffd uffd = {
 	.monitor.cleanup = ofi_monitor_cleanup,
 	.monitor.start = ofi_uffd_start,
 	.monitor.stop = ofi_uffd_stop,
+	.monitor.name = "uffd",
 };
 struct ofi_mem_monitor *uffd_monitor = &uffd.monitor;
 
@@ -63,6 +64,9 @@ struct ofi_mem_monitor *default_monitor;
 struct ofi_mem_monitor *default_cuda_monitor;
 struct ofi_mem_monitor *default_rocr_monitor;
 struct ofi_mem_monitor *default_ze_monitor;
+
+struct ofi_mem_monitor **monitor_list;
+size_t monitor_list_size;
 
 static size_t ofi_default_cache_size(void)
 {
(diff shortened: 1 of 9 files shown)


prov/util/src/util_mem_monitor.c
```diff
@@ -277,12 +297,21 @@ void ofi_monitors_init(void)
...
-	ze_monitor->cleanup(ze_monitor);
-	import_monitor->cleanup(import_monitor);
+	struct ofi_mem_monitor *monitors[] = {
+		uffd_monitor,
+		memhooks_monitor,
+		cuda_monitor,
+		rocr_monitor,
+		ze_monitor,
+		import_monitor,
+	};
```
j-xiong, line 307: I know it's an existing issue: the monitors being cleaned up doesn't match those are initialized.  How about defining the monitors array as file scope static and use the same array for both initialization and cleanup?

muttormark: I have implemented the later, ie, save the list at init time.  This change also calls cleanup on all the monitors (some of which did not have cleanup() processing previously).

prov/util/src/util_mem_monitor.c
```diff
@@ -171,15 +213,14 @@ void ofi_monitors_init(void)
...
-	cuda_ipc_monitor->init(cuda_ipc_monitor);
-	rocr_monitor->init(rocr_monitor);
-	rocr_ipc_monitor->init(rocr_ipc_monitor);
-	xpmem_monitor->init(xpmem_monitor);
-	ze_monitor->init(ze_monitor);
-	import_monitor->init(import_monitor);
+	initialize_monitor_list();
+
+	for (size_t i = 0; i < monitor_list_size; i++) {
+		FI_DBG(&core_prov, FI_LOG_MR,
```
shefty, line 219: This could be FI_INFO instead of debug, and maybe assert that name is non-null?  Cleanup could also be FI_INFO

muttormark: I implemented the changes suggested by @shefty

prov/util/src/util_mem_monitor.c
j-xiong: @muttormark I mean the array you defined here, why can't it be moved outside the functions, and adding a "static" keyword to limit the scope? This way the same array is shared by init and cleanup.

j-xiong (approved):

PR #9741, opened 2024-01-19
util: Fix coverity issue about missing lock

j-xiong:
All write accesses to `wait->signal` are protected by the lock. The read access should be protected, too.

### DIFF
--- a/prov/util/src/util_wait.c
+++ b/prov/util/src/util_wait.c
@@ -599,7 +599,16 @@ static int util_wait_yield_run(struct fid_wait *wait_fid, int timeout)
 	wait = container_of(wait_fid, struct util_wait_yield, util_wait.wait_fid);
 	endtime = ofi_timeout_time(timeout);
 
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

prov/util/src/util_wait.c
```diff
@@ -599,7 +599,16 @@ static int util_wait_yield_run(struct fid_wait *wait_fid, int timeout)
 	wait = container_of(wait_fid, struct util_wait_yield, util_wait.wait_fid);
 	endtime = ofi_timeout_time(timeout);
 
-	while (!wait->signal) {
+	while (1) {
+		int signaled;
+
+		ofi_mutex_lock(&wait->signal_lock);
```
shijin-aws, line 605: why not move this outside the while loop and use `while(signaled`?

shijin-aws: Nvm, didn't realize the signaled can change

PR #9742, opened 2024-01-19
prov/shm: adjust the order of smr_region fields.

shijin-aws:
xpmem is a opt-in feature, and xpmem_cap_self, xpmem_self, and xpmem_peer are not used in fast path. The previous commit e0906f6f inserts xpmem related fields in the middle of smr_region struct, which makes the size of smr_region struct increase from 104 (2 cache lines) to 144 ( 3 cache lines). The insertion of xpmem_cap_self, xpmem_self, and xpmem_peer make the fields (offsets) used in the fast path shifted to the next cache line and causes more cache misses.

This patch moves the xpmem fields to the end of the struct to fix this issue.

### DIFF
--- a/prov/shm/src/smr_util.h
+++ b/prov/shm/src/smr_util.h
@@ -226,9 +226,6 @@ struct smr_region {
 	uint8_t		cma_cap_peer;
 	uint8_t		cma_cap_self;
 	uint32_t	max_sar_buf_per_peer;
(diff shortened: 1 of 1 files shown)


shijin-aws:
Our benchmark shows this patch fixes a 2% regression for shm's SAR protocol (16k bytes) on AWS machine

j-xiong (approved):

PR #9745, opened 2024-01-19
contrib/intel/jenkins: Fix minor CI bugs

zachdworkin:
Daos cleanup was missing on abort.
verbs-rxm-mpich wasn't being run because it was looking in the wrong location.
ze_v3_shm filenames changed and weren't being picked up by the summarizer.
All three of these are fixed in this PR.


zachdworkin:
@j-xiong is this good to merge?

j-xiong (approved):

PR #9750, opened 2024-01-20
include, man, prov/efa: Introduce and implement FI_OPT_SHARED_MEMORY_PERMITTED 

sunkuamzn:
FI_OPT_SHARED_MEMORY_PERMITTED is a new endpoint option that controls the use of shared memory for intra-node communication


sunkuamzn:
@j-xiong can you please check the Intel CI failures?

j-xiong:
issue with one test node. restarted the test.

j-xiong (approved):

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1175,6 +1175,35 @@ static int efa_rdm_ep_set_cuda_api_permitted(struct efa_rdm_ep *ep, bool cuda_ap
...
+ *			-FI_EINVAL if shm is requested but the FI_EFA_ENABLE_SHM_TRANSFER environment variable is set to false
+ * @related efa_rdm_ep
+ */
+static int efa_rdm_ep_set_shared_memory_permitted(struct efa_rdm_ep *ep, bool shm_permitted)
+{
+	if (!shm_permitted) {
+		EFA_WARN(FI_LOG_EP_CTRL,
+			 "FI_OPT_SHARED_MEMORY_PERMITTED set to false. "
+			 "Tearing down shm provider resources.");
+		efa_rdm_ep_close_shm_resources(ep);
```
shijin-aws, line 1192: We have another place that close shm resource during ep enable: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L979. I think it's better to keep the logic consistent: tear down shm inside setopt, or both clean up during ep enable. I think the first one is better (as you do now). Can you update the setopt for cuda_api_permitted to do the same?

shijin-aws (approved):
Overall LGTM, thank you!

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: Then we can remove the `efa_rdm_ep_update_shm` function

man/fi_efa.7.md
```diff
@@ -205,6 +205,12 @@ These OFI runtime parameters apply only to the RDM endpoint.
   [`ptrace protection`](https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening#ptrace_Protection)
   is turned on. You can turn it off to enable shm transfer.
 
+  FI_EFA_ENABLE_SHM_TRANSFER is related to the FI_OPT_SHARED_MEMORY_PERMITTED endpoint option.
+  If FI_EFA_ENABLE_SHM_TRANSFER is set to true, the FI_OPT_SHARED_MEMORY_PERMITTED endpoint
+  option overrides FI_EFA_ENABLE_SHM_TRANSFER. If FI_EFA_ENABLE_SHM_TRANSFER is set to false,
+  but the FI_OPT_SHARED_MEMORY_PERMITTED is set to true, the FI_OPT_SHARED_MEMORY_PERMITTED
```
a-szegel, line 211: Why do we want to fail instead of always trusting the env var?

shijin-aws: We shouldn't ignore the conflicting configuration. We have the same handling for use_device_rdma

prov/efa/src/rdm/efa_rdm_ep_fiops.c
sunkuamzn: I will move the teardown in cuda_api_permitted to setopt in a separate commit

man/fi_efa.7.md
sunkuamzn: Because if the env var is set to false, we don't create SHM resources in the `fi_domain` call. I don't think we should go back and re-create those resources.

I can update the man page to clarify that `FI_EFA_ENABLE_SHM_TRANSFER` is parsed in the `fi_domain` call

prov/efa/src/rdm/efa_rdm_ep_fiops.c
sunkuamzn: I see. Then I will move all cleanup to ep enable.

shijin-aws (approved):

shijin-aws:
@a-szegel do you have more comments?

a-szegel (approved):

PR #9751, opened 2024-01-20
rdma/fi_eq: Support cntr byte counting

chadkoster-hpe:
Define new enum value to support byte counting with counters.
Document new define.

### DIFF
--- a/man/fi_cntr.3.md
+++ b/man/fi_cntr.3.md
@@ -121,7 +121,16 @@ struct fi_cntr_attr {
   and/or receives -- which are counted may be restricted using control
   flags when binding the counter and the endpoint.  Counters increment
   on all successful completions, separately from whether the operation
-  generates an entry in an event queue.
+  generates an entry in an event queue.  This includes truncated messages.
+
+- *FI_CNTR_EVENTS_BYTES*
+: The counter increments by the manipulated length for every successful 
+  completion that occurs on an associated bound endpoint.  The type of 
+  completions -- sends and/or receives -- which are counted may be 
+  restricted using control flags when binding the counter and the 
+  endpoint.  Counters increment on all successful completions, 
+  separately from whether the operation generates an entry in an 
+  event queue.  This includes truncated messages.
 
 *wait_obj*
 : Counters may be associated with a specific wait object.  Wait
--- a/man/fi_cntr.3.md
+++ b/man/fi_cntr.3.md
@@ -121,16 +121,18 @@ struct fi_cntr_attr {
   and/or receives -- which are counted may be restricted using control
   flags when binding the counter and the endpoint.  Counters increment
   on all successful completions, separately from whether the operation
-  generates an entry in an event queue.  This includes truncated messages.
+  generates an entry in an event queue.
 
 - *FI_CNTR_EVENTS_BYTES*
-: The counter increments by the manipulated length for every successful 
+: The counter increments by the the transer size for every successful 
   completion that occurs on an associated bound endpoint.  The type of 
   completions -- sends and/or receives -- which are counted may be 
   restricted using control flags when binding the counter and the 
   endpoint.  Counters increment on all successful completions, 
   separately from whether the operation generates an entry in an 
-  event queue.  This includes truncated messages.
+  event queue. On error, the tranfer size is not applied to the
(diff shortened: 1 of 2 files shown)


j-xiong:
DCO error: 
```
Commit sha: 6a6d1bf, Author: Chad Koster, Committer: chadkoster-hpe; 
Expected "Chad Koster chad.koster@hpe.com", but got "Ian Ziemba Ian.Ziemba@hpe.com".
```

man/fi_cntr.3.md
```diff
@@ -121,7 +121,16 @@ struct fi_cntr_attr {
   and/or receives -- which are counted may be restricted using control
   flags when binding the counter and the endpoint.  Counters increment
   on all successful completions, separately from whether the operation
-  generates an entry in an event queue.
+  generates an entry in an event queue.  This includes truncated messages.
```
iziemba, line 124: I have concerns that the "this includes truncated messages" for FI_CNTR_EVENTS_COMP breaks the existing API. Specifically, looking at the fi_domain manpage, there is this section.

```
Rx Buffer Overrun
This refers to buffers posted to receive incoming tagged or untagged messages, with the behavior defined from the viewpoint of the sender. The behavior for handling received messages that are larger than the buffers provided by the application is provider specific. Providers may either truncate the message and report a successful completion, or fail the operation. For datagram endpoints, failed sends will result in the message being dropped. For reliable endpoints, send operations may complete successfully, yet be truncated at the receive side. This can occur when the target side buffers received data until an application buffer is made available. The completion status may also be dependent upon the completion model selected byt the application (e.g. FI_DELIVERY_COMPLETE versus FI_TRANSMIT_COMPLETE).
```

While a truncation at the sender may be considered a success, truncation at the receive is typically reported via `struct fi_cq_err_entry`. Does this mean truncation at receive is an error event? @j-xiong thoughts?

j-xiong: Agree, this clause should be removed.

man/fi_cntr.3.md
```diff
@@ -121,7 +121,16 @@ struct fi_cntr_attr {
...
+
+- *FI_CNTR_EVENTS_BYTES*
+: The counter increments by the manipulated length for every successful 
+  completion that occurs on an associated bound endpoint.  The type of 
+  completions -- sends and/or receives -- which are counted may be 
+  restricted using control flags when binding the counter and the 
+  endpoint.  Counters increment on all successful completions, 
+  separately from whether the operation generates an entry in an 
+  event queue.  This includes truncated messages.
```
shefty, line 133: It's unclear to me what the counter is actually counting.  I'm also unsure what is meant by 'manipulated length'.

The above definition references 'successful completions'.  Does that mean that transfers which do not generate a completion are not counted?  What about CQ data carried in a message?  Do those bytes get counted?  For atomics that return a result to the sender, is the result data also included at the initiator, and if so, how?  Is an atomic result counted a send data or receive data?  It may help to clarify exactly what gets counted for every type of data transfer which could be impacted by this feature.

There are a couple of ways an implementation can count the number of bytes of a data transfer.  One, it could increment the count as each packet is processed.  That would provide an accurate number of bytes which have been currently updated.  In the case of a failure, the count would still have been updated.  However, that behavior is not aligned with the above definition, which disables such an implementation.

If only fully successful transfers are updated, then the byte count must be tracked separately and updated only after the last packet has been processed.

I would not reference the term 'successful completion'.  The count is relative to either the initiator or target processing of the transfer.  The operation itself could fail.  For example, if the ack to the sender is lost, the receiver will still have the data and updated count, but the sender's operation may fail.

swelch:
@chadkoster-hpe - The commit that corrects the typo should just be squashed into the commit that introduced the text, and you need to correctly sign-off on your commits "git commit -s". 

You also need update src/fi_tostr.c function "ofi_tostr_cntr_events() to handle the new enum value.

chadkoster-hpe:
This pull request has become too far out of date with the main branch, creating a new pull request

shefty:
@chadkoster-hpe - It's better if you rebase the branch and force push to update a PR.  That way it maintains the comments.  For this particular PR, it doesn't really matter as it's mostly updating the man page.

PR #9762, opened 2024-01-24
fabtests, contrib/intel/jenkins: Add IO_URING support to CI

zachdworkin:
Add io_uring.exclude file to pass to runfabtests.sh
The goal of this file add is to allow us to enable IO_URING tests in the CI without having massive failures. This is because the code path for io_uring is not 100% complete but there are still many passing tests that we would like to run the CI on.

The CI changes required for this are adding a new build step to the CI to build and run on a new pair of nodes that have IO_URING support. We need to add the build step, run step, summary setep, and add support for specific hardware selection in the CI.


contrib/intel/jenkins/runtests.py
```diff
@@ -18,7 +18,7 @@ def __call__(self, parser, namespace, values, option_string=None):
 parser = argparse.ArgumentParser()
 parser.add_argument('--build_hw', help="HW type for build",
                     choices=['water', 'grass', 'fire', 'electric', 'daos',\
-                                'gpu', 'ucx'])
+                                'gpu', 'ucx', 'ivysaur'])
```
shefty, line 21: The CI naming scheme brings me much joy.

ooststep (approved):

PR #9771, opened 2024-01-27
prov/efa: Fix a bug in efa_rdm_ope_post_send

shijin-aws:
efa_rdm_pke_alloc can return NULL when tx pkt pool is exhausted. In this case we should release allocated pkt entries and return EAGAIN.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -1703,8 +1703,8 @@ ssize_t efa_rdm_ope_post_send(struct efa_rdm_ope *ope, int pkt_type)
 	struct efa_rdm_pke *pkt_entry_vec[EFA_RDM_EP_MAX_WR_PER_IBV_POST_SEND];
 	ssize_t err;
 	size_t segment_offset;
-	int pkt_entry_cnt, pkt_entry_data_size_vec[EFA_RDM_EP_MAX_WR_PER_IBV_POST_SEND];
-	int i, j;
+	int pkt_entry_cnt, pkt_entry_cnt_allocated = 0, pkt_entry_data_size_vec[EFA_RDM_EP_MAX_WR_PER_IBV_POST_SEND];
+	int i;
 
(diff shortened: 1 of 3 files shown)


shijin-aws:
will add another commit to add a fabtest

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1716,39 +1716,45 @@ ssize_t efa_rdm_ope_post_send(struct efa_rdm_ope *ope, int pkt_type)
 	segment_offset = efa_rdm_pkt_type_contains_data(pkt_type) ? ope->bytes_sent : -1;
 	for (i = 0; i < pkt_entry_cnt; ++i) {
 		pkt_entry_vec[i] = efa_rdm_pke_alloc(ep, ep->efa_tx_pkt_pool, EFA_RDM_PKE_FROM_EFA_TX_POOL);
-		assert(pkt_entry_vec[i]);
+
+		if (OFI_UNLIKELY(!pkt_entry_vec[i])) {
+			err = -FI_EAGAIN;
```
sunkuamzn, line 1721: If we return EAGAIN, it's possible for the app to hang if the tx size is smaller than the number of packets needed to send the message. There's no guarantee that the packet entries will be freed the next time we try to send. And there will be no warning which will make it hard to debug.

I think we should print a warning, return ENOMEM and let the app crash. The customer can then set a larger tx size for the app to run properly.

shijin-aws: There are two possible cases: 1. tx is too small and smaller than pkt_entry_cnt 2. tx queue is big enough but now exhausted. I think we should print warning and crash for 1 but not for 2. There are similar EAGAIN in the code for the pkt entry allocation failure.

shijin-aws: To make this bug fix specific for case 2, I will modify the test to test with default tx size but larger window (like 5000)

sunkuamzn: Discussed offline. EAGAIN should be OK.

sunkuamzn (approved):

shijin-aws:
pure efa provider change, ignoring intel ci failure

PR #9773, opened 2024-01-29
core: Print provider path on fabric creation

chien-intel:
Print provider path on fi_fabric call to help with provider debug.  For users of FI_PROVIDER_PATH with multiple instances of the same provider, this will help to determine the exact provider code being executed.

### DIFF
--- a/src/fabric.c
+++ b/src/fabric.c
@@ -1471,6 +1471,9 @@ int DEFAULT_SYMVER_PRE(fi_fabric)(struct fi_fabric_attr *attr,
 {
 	struct ofi_prov *prov;
 	const char *top_name;
+#ifdef HAVE_LIBDL
+	Dl_info dl_info;
+#endif
 	int ret;
 
 	if (!attr || !attr->prov_name || !attr->name)
@@ -1499,6 +1502,16 @@ int DEFAULT_SYMVER_PRE(fi_fabric)(struct fi_fabric_attr *attr,
 			attr->name);
 
 		ofi_hook_install(*fabric, fabric, prov->provider);
+
+#ifdef HAVE_LIBDL
+		if (dladdr(prov->provider->fabric, &dl_info))
+			FI_INFO(&core_prov, FI_LOG_CORE,
+				"Using %s provider %u.%u, path:%s\n",
+				prov->prov_name,
+				FI_MAJOR(prov->provider->fi_version),
+				FI_MINOR(prov->provider->fi_version),
+				dl_info.dli_fname);
+#endif
 	}
 
 	return ret;


src/fabric.c
```diff
@@ -1499,6 +1502,11 @@ int DEFAULT_SYMVER_PRE(fi_fabric)(struct fi_fabric_attr *attr,
 			attr->name);
```
shefty, line 1502: I would replace this print with a new function that includes the additional information.  E.g. "Opened fabric: blah from provider: blahblah (ver: 1.15) path: blah/blah".  That would isolate the ifdef to the new function.  I included the provider version as that indirectly indicates which provider was picked up.  But I'd also include the path, to make it easier on the user and avoid needing to cross reference this print with an earlier print.  For providers that were built in, I would indicate that in the print, so that it's conveyed directly, rather than the reader needing to know that the absence of a print means build-in,

zachdworkin:
@chien-intel DAOS CI Failed. I am investigating to see why. We can re-run your PR after the changes Sean suggested are made or manually if they don't need to be made.

src/fabric.c
chien-intel: I don't want to combine the two prints and put it under #ifdef HAVE_LIBDL because you will lose the print without DL.  I've added version number as it may be useful.  If the provider is built-in, path to libfabric.so will get printed.

shefty: Having ifdef's within a function make the maintenance of the function more difficult.  That's why all or nearly all ifdefs we have surround entire functions.  I was suggesting something like:
```
#ifdef HAVE_LIBDL
void ofi_print_fabric_prov(...)
{
    FI_INFO(...)
}
#else
void ofi_print_fabric_prov(...)
{
    FI_INFO(..."internal"...)
}
#endif

Then just call ofi_print_fabric_prov().
```

chien-intel: You are bringing back nightmares with irdma_kcompat.h, please refrain from discussing maintainability with #ifdef.  :-)  With that, this one is more inline with the one in ofi_cleanup_prov than the big wrapper around ofi_reg_dl_prov, ofi_ini_dir, etc where those were not called and only applicable with libdl.  To create a function out of a print that's only needed with libdl is overkill.

j-xiong:
bot:aws:retest

src/fabric.c
j-xiong: Just to clarify, `HAVE_LIBDL` is almost always true unless the use of DL provider is explicitly disabled (`--with-dlopen=no`). Both built-in providers and DL providers are reported under this code path. The latest patch revision looks fine to me.

j-xiong (approved):

PR #9778, opened 2024-01-31
Revert "prov/efa: Disable RUNTING for Neuron"

a-szegel:
This reverts commit 56b22865ffc39015cf6cbf54f2888a0e0855f683.


liralon:
Please add to the commit message the context and rational of why this revert should be done.

a-szegel:
I don't think this is the fix we want.

PR #9783, opened 2024-02-01
prov/efa: Add a missing rx pkt counter

shijin-aws:
Currently efa rdm ep has 2 rx pkt counters:
efa_rx_pkt_posted, efa_rx_pkt_to_post. The
progress engine uses efa_rx_pkt_posted==0
and efa_rx_pkt_to_post==0 to identify whether
the endpoint is progressed the first time.

However, these 2 counters do not cover the activity of all rx pkts. Some rx pkts can be held within
a round of progress engine due to queued hmem
copy or being posted for local read. And when the
rx pkt pool is small (customized by env or hints), the held rx pkts can cause both efa_rx_pkt_posted
and efa_rx_pkt_to_post as 0 in the middle of the
progress engine, which will trigger another
pkt pool grow and fail.

This patch adds an additional counter: efa_rx_pkts_held in efa_rdm_ep to cover this part of rx pkts. And added a stronger assertion that

efa_rx_pkts_posted + efa_rx_pkts_to_post + efa_rx_pkts_held must be equal to the rx pool size during the beginning of progress engine call. This assertion will make us catch any rx packet leak in the future.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_progress.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_progress.c
@@ -180,8 +180,9 @@ void efa_rdm_ep_progress_post_internal_rx_pkts(struct efa_rdm_ep *ep)
 			ep->efa_rx_pkts_to_post = 0;
 		}
 	} else {
-		if (ep->efa_rx_pkts_posted == 0 && ep->efa_rx_pkts_to_post == 0) {
-			/* Both efa_rx_pkts_posted and efa_rx_pkts_to_post equal to 0 means
+		if (ep->efa_rx_pkts_posted == 0 && ep->efa_rx_pkts_to_post == 0 && ep->efa_rx_pkts_held == 0) {
+			/* All of efa_rx_pkts_posted, efa_rx_pkts_to_post and
+			 * efa_rx_pkts_held equal to 0 means
 			 * this is the first call of the progress engine on this endpoint.
 			 *
 			 * In this case, we explictly allocate the 1st chunk of memory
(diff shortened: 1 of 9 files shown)


prov/efa/src/rdm/efa_rdm_ep_progress.c
```diff
@@ -180,8 +180,9 @@ void efa_rdm_ep_progress_post_internal_rx_pkts(struct efa_rdm_ep *ep)
 			ep->efa_rx_pkts_to_post = 0;
 		}
 	} else {
-		if (ep->efa_rx_pkts_posted == 0 && ep->efa_rx_pkts_to_post == 0) {
-			/* Both efa_rx_pkts_posted and efa_rx_pkts_to_post equal to 0 means
+		if (ep->efa_rx_pkts_posted == 0 && ep->efa_rx_pkts_to_post == 0 && ep->efa_rx_pkts_held == 0) {
```
wzamazon, line 183: should this change be in this commit?

shijin-aws: sure, I can make it a separate commit. There is another change that decrements the efa_rx_pksts_posted counter in the error path which should be a separate commit too.

a-szegel (approved):
LGTM, fix Wei's comment!

prov/efa/src/rdm/efa_rdm_ep_progress.c
shijin-aws: Updated

wzamazon (approved):

PR #9786, opened 2024-02-05
adding riscv support to opx conf

ct-clmsn:
adding riscv support to the ops provider.

### DIFF
--- a/prov/opx/configure.m4
+++ b/prov/opx/configure.m4
@@ -49,7 +49,7 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 	dnl and is not supported for non-x86 processors.
 	AS_IF([test "x$macos" = "x1"],[opx_happy=0],
 		[test "x$freebsd" = "x1"],[opx_happy=0],
-		[test x$host_cpu != xx86_64],[opx_happy=0],
+		[test x$host_cpu != xx86_64 && test x$host_cpu != xriscv && test x$host_cpu != xriscv64],[opx_happy=0],
 		[test x"$enable_opx" != x"no"],[
 
 		AC_MSG_CHECKING([for opx provider])
--- a/prov/opx/include/rdma/opx/fi_opx_compiler.h
+++ b/prov/opx/include/rdma/opx/fi_opx_compiler.h
@@ -59,14 +59,33 @@
 
 static inline void fi_opx_compiler_msync_writes()
 {
-	asm volatile ("sfence" :  :  : "memory");
+#if defined(__riscv)
+
+#if defined(__riscv_xlen) && (__riscv_xlen == 64)
+ 	asm volatile ("fence ow,ow" :  :  : "memory");
+#else
+#error "Unsupported CPU type"
+#endif
+
+#else
+ 	asm volatile ("sfence" :  :  : "memory");
+#endif
 }
 
 static inline void fi_opx_compiler_msync_reads()
 {
-	asm volatile ("lfence" :  :  : "memory");
-}
+#if defined(__riscv)
 
+#if defined(__riscv_xlen) && (__riscv_xlen == 64)
+	asm volatile ("fence ir,ir" :  :  : "memory");
+#else
+#error "Unsupported CPU type"
+#endif
+
+#else
+ 	asm volatile ("lfence" :  :  : "memory");
+#endif
+}
 
 #define fi_opx_compiler_barrier()		__asm__ __volatile__ ( "" ::: "memory" )
 
@@ -87,9 +106,19 @@ void fi_opx_compiler_store_u64(volatile uint64_t * const variable, const uint64_
 static inline
 void fi_opx_compiler_inc_u64(volatile uint64_t * const variable)
 {
+#if defined(__riscv)
+
+#if defined(__riscv_xlen) && (__riscv_xlen == 64)
+        (*variable) += 1;
+#else
+#error "Unsupported CPU type"
+#endif
+
+#else
 	__asm__ __volatile__ ("lock ; incq %0"
 				: "=m" (*variable)
 				: "m" (*variable));
+#endif
 	return;
 }
 
@@ -97,18 +126,41 @@ static inline
 uint64_t fi_opx_compiler_fetch_and_inc_u64(volatile uint64_t * const variable)
 {
 	uint64_t value = 1;
+#if defined(__riscv)
+
+#if defined(__riscv_xlen) && (__riscv_xlen == 64)
+	const uint64_t tmp = (*variable);
+	(*variable) = value;
+	value = tmp;
+	(*variable) = (*variable) + value;
+#else
+#error "Unsupported CPU type"
+#endif
+
+#else
 	__asm__ __volatile__ ("lock ; xadd %0,%1"
 				: "=r" (value), "=m" (*variable)
 				: "0" (value), "m" (*variable));
+#endif
 	return value;
 }
 
 static inline
 void fi_opx_compiler_dec_u64(volatile uint64_t * const variable)
 {
+#if defined(__riscv)
+
+#if defined(__riscv_xlen) && (__riscv_xlen == 64)
+	(*variable) -= 1;
+#else
+#error "Unsupported CPU type"
+#endif
+
+#else
 	__asm__ __volatile__ ("lock ; decq %0"
 				: "=m" (*variable)
 				: "m" (*variable));
+#endif
 	return;
 }
 
--- a/prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
+++ b/prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
@@ -43,7 +43,12 @@
 #include "rdma/fabric.h"	/* only for 'fi_addr_t' ... which is a typedef to uint64_t */
 #include "rdma/opx/fi_opx_addr.h"
 
-
+#if defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)
+#ifndef PAGE_SIZE
+/* 4K pages default */
+#define PAGE_SIZE 4096
+#endif
+#endif
 
 #define FI_OPX_ADDR_SEP_RX_MAX			(4)
 #define FI_OPX_HFI1_PACKET_MTU			(8192)
--- a/prov/opx/include/rdma/opx/fi_opx_timer.h
+++ b/prov/opx/include/rdma/opx/fi_opx_timer.h
@@ -62,6 +62,14 @@ static inline uint64_t fi_opx_timer_get_cycles()
 	cycles = ((uint64_t)a) | (((uint64_t)d) << 32);
 	return cycles;
 }
+#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)
+__attribute__((always_inline))
+static inline uint64_t fi_opx_timer_get_cycles()
+{
+        uint64_t dst = 0;
+        asm volatile ("rdcycle %0" : "=r" (dst) );
+        return dst;
+}
 #else
 #error "Cycle timer not defined for this platform"
 #endif


prov/opx/configure.m4
```diff
@@ -50,6 +50,7 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 	AS_IF([test "x$macos" = "x1"],[opx_happy=0],
 		[test "x$freebsd" = "x1"],[opx_happy=0],
 		[test x$host_cpu != xx86_64],[opx_happy=0],
+		[test x$host_cpu != xriscv64],[opx_happy=0],
```
j-xiong, line 53: This check is already implied by the previous line.

j-xiong: Or do you mean `=` instead of `!=` ?

ct-clmsn: @j-xiong thank you for double checking, looks like there's an issue on my end updating the autoconf scripts, cached versions are being run. Will fix this issue soon.

ct-clmsn:
@j-xiong alright, fixed several things. Turns out I had to manually delete the configure script. Went through a build and found out several macros needed default values in the opx header file. Also found some missing implementations for riscv64 atomics. Managed to get all of that completed. Hope it looks good to ya'll.

prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
```diff
@@ -43,7 +43,10 @@
 #include "rdma/fabric.h"	/* only for 'fi_addr_t' ... which is a typedef to uint64_t */
 #include "rdma/opx/fi_opx_addr.h"
 
-
+#ifndef PAGE_SIZE
+/* 4K pages default */
+#define PAGE_SIZE 4000
+#endif
```
belynam, line 49: Should this be 4096 instead of 4000? This may break some things in OPX that assume PAGE_SIZE is a power of 2, such as alignment checking/adjustment.

Please also surround this with an `#if defined(__riscv)`

prov/opx/include/rdma/opx/fi_opx.h
```diff
@@ -50,6 +50,18 @@
...
+#define OPX_AV FI_AV_UNSPEC
+#endif
+
+#if !defined(OPX_THREAD)
+#define OPX_THREAD FI_THREAD_ENDPOINT
+#endif
+
+#if !defined(OPX_MR)
+#define OPX_MR FI_MR_BASIC
+#endif
```
belynam, line 63: Why do these need to be added here? They should already be defined from prov/opx/configure.m4. If they're not, then configure.m4 should be fixed to do so, rather than adding them here.

belynam (changes requested):

prov/opx/include/rdma/opx/fi_opx.h
ct-clmsn: @belynam the build process assumes several of these macro definitions are created, it turns out they are not created. The documentation for OPX provides those as default value. If ya'll would prefer them placed in configure.m4 I'll go ahead and do that.

prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
ct-clmsn: Ah right, thanks. Will make the proper correction.

prov/opx/include/rdma/opx/fi_opx.h
ct-clmsn: correction, these preprocessor variables supposed to be defined in the configure.ac script using `AC_DEFINE_UNQUOTED` for opx. During compilation of the ops provider, the compiler claims these variables have not defined. Is this a configuration variable caching issue or is there a specific fix for this situation you all would prefer to see?

belynam: These 3 values (OPX_AV, OPX_THREAD, OPX_MR) are defined in our configure.m4 already. For example, OPX_AV is defined [here](https://github.com/ofiwg/libfabric/blob/2f605226ce5c41106b72a664f711c91dcb7d6ca0/prov/opx/configure.m4#L74).

If they are not getting defined, that would suggest that portion of the configure script is not being executed, even though it should be.

prov/opx/configure.m4
```diff
@@ -49,7 +49,9 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 	dnl and is not supported for non-x86 processors.
 	AS_IF([test "x$macos" = "x1"],[opx_happy=0],
 		[test "x$freebsd" = "x1"],[opx_happy=0],
-		[test x$host_cpu != xx86_64],[opx_happy=0],
+		[test x$host_cpu = xx86_64],[opx_happy=1],
+		[test x$host_cpu = xriscv],[opx_happy=1],
+		[test x$host_cpu = xriscv64],[opx_happy=1],
```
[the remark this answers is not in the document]
ct-clmsn: thank you for the suggestion, this fixed the issue

ct-clmsn:
@j-xiong : saw you note about the squash operation, made an unsuccessful attempt to squash the commits. my apologies for not being able to make that work. I'll walk through the commit history and attempt to fix the commit messages with the appropriate prefix.

ct-clmsn:
@j-xiong made updates to the commit message history.

j-xiong:
@ct-clmsn I think all these 5 commits can be squashed into one since the later ones are mostly fixing errors of earlier ones. To do this, use the following steps:

(1) run `git rebase -i HEAD^^^^^^`, mark the last four commits as 'fixup' and save the changes;
(2) run `git commit --amend` to update the commit message of the squashed commit;
(3) run `git push -f` to update the pull request.

ct-clmsn:
@j-xiong almost done - will post when the work is complete.

ct-clmsn:
@j-xiong all set, thanks for the assistance!

prov/opx/configure.m4
```diff
@@ -181,6 +181,7 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 		])
 	])
 
+	AM_CONDITIONAL([HAVE_CRITERION], [test "x$have_criterion" = "xtrue"])
```
j-xiong, line 184: Is it ever used in the opx provider?

prov/opx/include/rdma/opx/fi_opx_compiler.h
```diff
@@ -59,12 +59,20 @@
 
 static inline void fi_opx_compiler_msync_writes()
 {
-	asm volatile ("sfence" :  :  : "memory");
+#if defined(__x86_64__) || defined(__i386__)
+ 	asm volatile ("sfence" :  :  : "memory");
+ #elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)
+ 	asm volatile ("fence ow,ow" :  :  : "memory");
+ #endif
```
j-xiong, line 66: extra space before `#`

prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
```diff
@@ -43,7 +43,11 @@
 #include "rdma/fabric.h"	/* only for 'fi_addr_t' ... which is a typedef to uint64_t */
 #include "rdma/opx/fi_opx_addr.h"
 
-
+#if defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)
+#ifndef PAGE_SIZE
+/* 4K pages default */
```
j-xiong, line 48: This comment is unnecessary.

prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
```diff
@@ -43,7 +43,11 @@
 #include "rdma/fabric.h"	/* only for 'fi_addr_t' ... which is a typedef to uint64_t */
 #include "rdma/opx/fi_opx_addr.h"
 
-
+#if defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)
```
j-xiong, line 46: The matching `#endif` is missing.

j-xiong:
BTW, the commit title can be changed to "prov/opx: Add RISC-V support".

prov/opx/include/rdma/opx/fi_opx_compiler.h
```diff
@@ -59,12 +59,20 @@
...
 }
 
 static inline void fi_opx_compiler_msync_reads()
 {
-	asm volatile ("lfence" :  :  : "memory");
+#if defined(__x86_64__) || defined(__i386__)
+ 	asm volatile ("lfence" :  :  : "memory");
+#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)
+	asm volatile ("fence ir,ir" :  :  : "memory");
+#endif
```
belynam, line 75: Same comment here as for msync_writes(). Please make x86 the default case.

prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
```diff
@@ -43,7 +43,11 @@
...
+#ifndef PAGE_SIZE
+/* 4K pages default */
+#define PAGE_SIZE 4096
```
belynam, line 49: In light of fixing up the config.m4, is defining PAGE_SIZE still necessary for riscv?

prov/opx/configure.m4
ct-clmsn: configure fails b/c `HAVE_CRITERION` is not defined in the opx provider. I wasn't able to find a fix for it other than to place that line of code in there.

j-xiong: Just in case, the steps to do the rebase (from your topic branch):
```
git pull --rebase upstream main
git push -f
```

j-xiong: That's assuming you have configured a remote `upstream`. If not, do the following in advance:
```
git add remote upstream https://github.com/ofiwg/libfabric
```

prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h
ct-clmsn: @belynam I'm not able to get that working on the autoconf side of things. Is this something ya'll could fix on your side after a PR merge?

belynam: We're picking up `PAGE_SIZE` as defined in **/usr/include/sys/user.h** via `#include <sys/user.h>` in fi_opx_hfi1_packet.h. Perhaps the system you're building on doesn't have it defined in there?

In any case, I think it's fine to leave this as is if needed for riscv. I don't consider this a blocker to merging.

ct-clmsn: Ah, that explains it. I'm on a riscv64 ubuntu distribution and `/usr/include/sys` does not exist.

ct-clmsn: @belynam I could modify the source to use `getpagesize` not sure what the pros/cons are to that approach.

belynam: > I could modify the source to use getpagesize not sure what the pros/cons are to that approach.

I think for now, just leave it as you have it in this PR (defining PAGE_SIZE as 4K for riscv). It looks to me like we shouldn't be getting PAGE_SIZE from /usr/include/sys/user.h. We should be using `sysconf(_SC_PAGESIZE)` instead. But we'll take care of fixing that, as there are bound to be performance impacts.

ct-clmsn: thanks! PR has been updated with the latest changes!

prov/opx/configure.m4
j-xiong: This line should have been removed.

ct-clmsn: @j-xiong all set; thank you for the assistance!

j-xiong (approved):

belynam (approved):

PR #9787, opened 2024-02-05
Intel/CI: Update dmabuf stage to support subdevice testing

Juee14Desai:
This commit updates the dmabuf stage to support tile-to-tile 
testing based on hierarchy update to newer agama versions.

### DIFF
--- a/contrib/intel/jenkins/tests.py
+++ b/contrib/intel/jenkins/tests.py
@@ -1062,7 +1062,7 @@ class DMABUFTest(Test):
                                             else 0
 
         if util_prov:
-            self.prov = f'{self.core_prov}\;{self.util_prov}'
+            self.prov = f"{self.core_prov}\;{self.util_prov}"
         else:
             self.prov = self.core_prov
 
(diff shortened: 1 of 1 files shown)


contrib/intel/jenkins/tests.py
```diff
@@ -1062,7 +1062,7 @@ def __init__(self, jobname, buildno, testname, hw, core_prov, fabric,
                                             else 0
 
         if util_prov:
-            self.prov = f'{self.core_prov}\;{self.util_prov}'
+            self.prov = f'{self.core_prov};{self.util_prov}'
```
zachdworkin, line 1065: Didn't we decide we need the '\' here? I thought it was making the command line think this was two different commands separated by the ';'?

Juee14Desai: Yes, I'll fix that but I am more inclined towards using self.core_prov since we are only using verbs.

zachdworkin: I think its better to have it this way even if we are only testing verbs. That way if we add more things in the future we won't have to figure out the same bug again.

contrib/intel/jenkins/tests.py
```diff
@@ -1073,27 +1073,40 @@ def __init__(self, jobname, buildno, testname, hw, core_prov, fabric,
...
+                            'H' : ['-d 0', '-d 1']
+                        },
+                'D2H'   : {
+                            '-d 0' : 'H',
+                            '-d 1' : 'H'
+                        },
+                'D2D'   : {
+                            '-d 0' : ['-d 1', '-d 2', '-d 3'],
+                            '-d 1' : ['-d 2', '-d 3']
+                        }
```
j-xiong, line 1090: The control logic later would be much simplified if the test configuration are defined as command line options directly and the lists are expanded. e.g.:

```
      'H2D'  : {
              '-m malloc' : '-m device -d 0'
              '-m malloc' : '-m device -d 1'
       },
```

Juee14Desai: I figured that adding -m device and -m malloc to each config would be redundant and hence just specified H for malloc  and -d <device_no> for device. This way any combination could be added later on in the config.

j-xiong: That results in the triple loop control structure that is much harder to understand vs the single loop if the redundancy is tolerated in the configuration. IMHO, the configuration with the `-m` option added is also easier to read.

Juee14Desai:
Intel CI failure is due to an unrelated summary stage failure, dmabuf stage has passed and summary shows correct results.

zachdworkin (approved):

PR #9790, opened 2024-02-06
prov/cxi: honor `FI_CXI_DISABLE_NON_INJECT_MSG_IDC` with rma

thomasgillis:
Hi all, 

I would like the `cxi` provider to honor the use variable `FI_CXI_DISABLE_NON_INJECT_MSG_IDC=1` in both the `msg` and `rma` paths.
When disabling the use of GDR copy on CUDA devices, here is the latency I get when disabling `NON_INJECT_MSG_IDC` (before the PR):
- `fi_tsend`: `9.779660us` -> `3.314845us`
- `fi_write`: `10.573375us` -> `10.326375us`

### DIFF
--- a/prov/cxi/src/cxip_rma.c
+++ b/prov/cxi/src/cxip_rma.c
@@ -503,6 +503,10 @@ static bool cxip_rma_is_idc(struct cxip_txc *txc, uint64_t key, size_t len,
 {
 	size_t max_idc_size = unr ? CXIP_INJECT_SIZE : C_MAX_IDC_PAYLOAD_RES;
 
+	/* DISABLE_NON_INJECT_MSG_IDC disables the IDC
+	 */
+	if (cxip_env.disable_non_inject_msg_idc)
+		return false;
 	/* IDC commands are not supported for unoptimized MR since the IDC
 	 * small message format does not support remote offset which is needed
 	 * for RMA commands.


prov/cxi/src/cxip_rma.c
```diff
@@ -503,6 +503,10 @@ static bool cxip_rma_is_idc(struct cxip_txc *txc, uint64_t key, size_t len,
 {
 	size_t max_idc_size = unr ? CXIP_INJECT_SIZE : C_MAX_IDC_PAYLOAD_RES;
 
+    /* DISABLE_NON_INJECT_MSG_IDC disables the IDC
+     */
+    if(cxip_env.disable_non_inject_msg_idc)
```
jswaro, line 508: There should be a space between if and (

prov/cxi/configure.m4
```diff
@@ -46,11 +46,18 @@ AC_DEFUN([FI_CXI_CONFIGURE],[
 		[
 			AC_CHECK_HEADER(cxi_prov_hw.h,
 				[],
-				[cxi_happy=0])
+				[cxi_happy=0],
+                [AC_INCLUDES_DEFAULT
```
jswaro, line 50: We shouldn't take these m4 changes. 

The problem that these changes are working around are only present until the next release (2.2.0), and so I'd like to prevent a regression by introducing these and having someone remove the headers again by accident. It is a useful check that these failed since it indicated a lower level problem. 

That said, if this is really critical, then we could discuss it further.

thomasgillis: I can remove those :-) added them accidentally

jswaro:
Looks fine to me. I'll ask @swelch  to double check it

swelch (approved):
It may be useful to keep disablement of IDC separate for msg and RMA, but if we determine that we can create a new PR.

PR #9810, opened 2024-02-14
fabtests/cq_test: Fix a bug in opened cq cnt

shijin-aws:
opened should only be incremented when
create_cq returns 0. The current implementation
will make it set as the actual opened count
plus 1 when some error happens. This will
cause a double free error when fi_close
the last cq which is not opened.

### DIFF
--- a/fabtests/unit/cq_test.c
+++ b/fabtests/unit/cq_test.c
@@ -79,13 +79,15 @@ static int cq_open_close_simultaneous(void)
 		return -FI_ENOMEM;
 
 	ret = 0;
-	for (opened = 0; opened < count && !ret; opened++) {
+	for (opened = 0; opened < count; opened++) {
 		ret = create_cq(&cq_array[opened], 0, 0, FI_CQ_FORMAT_UNSPEC,
 				FI_WAIT_UNSPEC);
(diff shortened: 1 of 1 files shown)


fabtests/unit/cq_test.c
```diff
@@ -79,13 +79,17 @@ static int cq_open_close_simultaneous(void)
...
 		ret = create_cq(&cq_array[opened], 0, 0, FI_CQ_FORMAT_UNSPEC,
 				FI_WAIT_UNSPEC);
-		if (ret) {
+		if (ret)
 			ret = create_cq(&cq_array[opened], 0, 0,
 					FI_CQ_FORMAT_UNSPEC, FI_WAIT_NONE);
-		}
+
+		if (ret)
+			break;
```
j-xiong, line 90: Adding these two lines to the original loop should achieve the same result. Can remove the redundant ` && !ret` condition, though.

shijin-aws: Done

shijin-aws: Updated

j-xiong (approved):

PR #9811, opened 2024-02-14
prov/efa: Create 1:1 relationship between libfabric CQs and IBV CQs 

shijin-aws:
Today, each endpoint has its own completion queue with the EFA provider, even if all endpoints bind the same CQ.
This caused applications to poll many more hardware completion queues than expected, at significantly reduced performance.
This patch fixes this issue by moving the IBV cq creation to fi_cq_open,
and bind it with ep explicitly via fi_ep_bind.

As a consequence of this change, the QP creation is moved to fi_ep_enable
because the ibv cq must be bound with qp during creation, which is guaranteed
during the ep enable.


shijin-aws:
Still working on adding new tests. Open the PR for CI to run

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -246,48 +250,6 @@ void test_efa_rdm_ep_handshake_receive_without_peer_host_id_and_do_not_send_loca
 	test_efa_rdm_ep_handshake_exchange_host_id(state, 0x0, 0x0, true);
 }
 
-/**
- * @brief Test efa_rdm_ep_open() handles rdma-core CQ creation failure gracefully
- *
- * @param[in]	state		struct efa_resource that is managed by the framework
- */
-void test_efa_rdm_ep_cq_create_error_handling(struct efa_resource **state)
```
shijin-aws, line 254: moved this test to efa_unit_test_cq.c as we are creating ibv cq in fi_cq_open now.

shefty:
FWIW, we tried to have a 1:1 relationship between ibv_cq and libfabric CQs with the verbs provider. That didn't end up being efficient because there are transmit operations which couldn't be reported to the user as done until a software ack was received.  The receive required polling on the receive CQ, even though, from the viewpoint of the app, it only wanted to poll on the send CQ for the completion.  So, both CQs ended up needing to be polled regardless of which libfabric CQ the user wanted to read.  Handling blocking semantics (either cq sread or polling on the fd) was likewise troublesome.  In the end, I believe we moved the ibv_cq to the libfabric domain, and all CQs simply referenced it.  It might progress data on endpoints that an app doesn't currently care about, but that's usually not a real concern.

I don't know if EFA would have the same restrictions.  Verbs is more troublesome since it maps RC QPs to RDM endpoints and has significant SW protocol.

shijin-aws:
> I don't know if EFA would have the same restrictions. Verbs is more troublesome since it maps RC QPs to RDM endpoints and has significant SW protocol.

Yep yep, we had the same restriction that some protocols requires to poll both tx and rx ibv cq (if separate) for a transmit completion.

 I introduced a `ibv_cq_poll_list` in efa's cq struct, which contains the tx and rx ibv cq if application is using separate tx/rx cq. The update of this list is done during `fi_enable` and `fi_ep_close`. It will do some cross referencing to collect all the relevant tx/rx cqs from endpoints that the Libfabric CQ is bound to.

shijin-aws:
>I don't know if EFA would have the same restrictions. Verbs is more troublesome since it maps RC QPs to RDM endpoints and has significant SW protocol.

That may be an alternative too. Worth evaluating.

shijin-aws:
>The receive required polling on the receive CQ, even though, from the viewpoint of the app, it only wanted to poll on the send CQ for the completion. So, both CQs ended up needing to be polled regardless of which libfabric CQ the user wanted to read. 

In my observation, both NCCL ofi plugin and Open MPI use a shared cq for send and recv. MPICH (likely Intel MPI) is using separate send / recv queues. Same for fabtests. Need to evaluate the performance impact when application has two cqs bind to single ep.

shefty:
The proposal for libfabric v2 is that an EP may only be bound to a single CQ.  This may require an update to apps, but once the apps update, they can use libfabric v1 or v2.

prov/efa/src/efa_cq.h
```diff
@@ -3,6 +3,82 @@
...
+enum ibv_cq_ex_type {
+	IBV_CQ,
+	EFADV_CQ
+};
+
+struct efa_ibv_cq {
+	struct ibv_cq_ex *ibv_cq_ex;
+	enum ibv_cq_ex_type ibv_cq_ex_type;
+	/* dlist entry inserted to cq/cntr's poll_list */
+	struct dlist_entry entry;
```
wenduwan, line 15: Why do we need to make efa_ibv_cq a linked list?

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -55,6 +71,330 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
...
+ *
+ * @param[in]		imm_data	Data provided in the IMMEDIATE value.
+ * @param[in]		len		Payload	length
+ * @param[in]		flags		flags (such as FI_REMOTE_CQ_DATA)
+ */
+static
+void efa_rdm_cq_proc_ibv_recv_rdma_with_imm_completion(
+						       uint64_t flags,
+						       struct efa_rdm_pke *pkt_entry,
+						       struct ibv_cq_ex *ibv_cq_ex)
```
wenduwan, line 89: Doc and function do not match

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -55,6 +71,330 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
...
+	uint32_t imm_data = ibv_wc_read_imm_data(ibv_cq_ex);
+	uint32_t len = ibv_wc_read_byte_len(ibv_cq_ex);
+
+	assert(ep);
+
+	target_cq = ep->base_ep.util_ep.rx_cq;
+	efa_av = ep->base_ep.av;
+
+	if (ep->base_ep.util_ep.caps & FI_SOURCE) {
+		src_addr = efa_av_reverse_lookup_rdm(efa_av,
```
wenduwan, line 105: Not introduced in this PR, but do we need to handle FI_ADDR_NOTAVAIL?

prov/efa/src/rdm/efa_rdm_cq.h
```diff
@@ -19,4 +22,5 @@ struct efa_rdm_cq {
 int efa_rdm_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 		    struct fid_cq **cq_fid, void *context);
 
+
```
wenduwan, line 25: nit The empty line should be after L26

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -27,30 +53,30 @@ static
...
-	attr_ex.send_cq = ibv_cq_ex_to_cq(ep->ibv_cq_ex);
-
-	attr_ex.cap.max_recv_wr = ep->base_ep.domain->device->rdm_info->rx_attr->size;
-	attr_ex.cap.max_recv_sge = ep->base_ep.domain->device->rdm_info->rx_attr->iov_limit;
-	attr_ex.recv_cq = ibv_cq_ex_to_cq(ep->ibv_cq_ex);
+	if (!ep->tx_cq && !ep->rx_cq) {
+		EFA_WARN(FI_LOG_EP_CTRL,
+			"Endpoint is not bound to a send or receive completion queue\n");
+		return -FI_ENOCQ;
+	}
```
wenduwan, line 62: Do we need this check after all? L64-L73 looks sufficient to me.

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -27,30 +53,30 @@ static
...
+		EFA_WARN(FI_LOG_EP_CTRL,
+			"Endpoint is not bound to a receive completion queue when it has receive capabilities enabled. (FI_RECV)\n");
+		return -FI_ENOCQ;
 	}
-	attr_ex.pd = efa_rdm_ep_domain(ep)->ibv_pd;
 
-	attr_ex.qp_context = ep;
-	attr_ex.sq_sig_all = 1;
+	tx_cq = ep->tx_cq ? ep->tx_cq->ibv_cq.ibv_cq_ex : ep->rx_cq->ibv_cq.ibv_cq_ex;
+	rx_cq = ep->rx_cq ? ep->rx_cq->ibv_cq.ibv_cq_ex : ep->tx_cq->ibv_cq.ibv_cq_ex;
```
wenduwan, line 77: Do we always need non-NULL tx/rx CQ? This seems restrictive...

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -595,6 +607,12 @@ static int efa_rdm_ep_bind(struct fid *ep_fid, struct fid *bfid, uint64_t flags)
 		if (ret)
 			return ret;
 
+		if (flags & FI_TRANSMIT)
```
wenduwan, line 610: Should this be FI_TRANSMIT or FI_SEND? It's conflicting with L66

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -773,12 +791,66 @@ void efa_rdm_ep_wait_send(struct efa_rdm_ep *efa_rdm_ep)
...
+	rx_cq = ep->rx_cq;
+
+	/* Remove the cross referencing of the CQs.
+	 * It must happen after ofi_endpoint_close
+	 * so we have cq's reference counters updated.
+	 */
+	if (tx_cq && !ofi_atomic_get32(&tx_cq->util_cq.ref)) {
+		efa_ibv_cq_poll_list_remove(&tx_cq->ibv_cq_poll_list, &tx_cq->util_cq.ep_list_lock, &tx_cq->ibv_cq);
+		if (rx_cq)
+			efa_ibv_cq_poll_list_remove(&rx_cq->ibv_cq_poll_list, &rx_cq->util_cq.ep_list_lock, &tx_cq->ibv_cq);
```
wenduwan, line 844: This looks odd - why is it needed?

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -773,12 +791,66 @@ void efa_rdm_ep_wait_send(struct efa_rdm_ep *efa_rdm_ep)
...
+{
+	int i;
+	struct efa_cntr *efa_cntr;
+	struct util_cntr *util_cntr;
+	struct efa_rdm_cq *tx_cq, *rx_cq;
+
+	tx_cq = ep->tx_cq;
+	rx_cq = ep->rx_cq;
+
+	for (i=0; i<CNTR_CNT; i++) {
```
wenduwan, line 816: nit Space around `=` and `<`
Same below

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -982,6 +1056,71 @@ void efa_rdm_ep_update_shm(struct efa_rdm_ep *ep)
...
+
+	if (tx_cq) {
+		ret = efa_ibv_cq_poll_list_insert(&tx_cq->ibv_cq_poll_list, &tx_cq->util_cq.ep_list_lock, &tx_cq->ibv_cq);
+		if (ret)
+			return ret;
+
+		if (rx_cq) {
+			ret = efa_ibv_cq_poll_list_insert(&tx_cq->ibv_cq_poll_list, &tx_cq->util_cq.ep_list_lock, &rx_cq->ibv_cq);
+			if (ret)
+				return ret;
```
wenduwan, line 1106: Same as above. This looks odd...

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1384,16 +1554,26 @@ static int efa_rdm_ep_setopt(fid_t fid, int level, int optname,
 	case FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES:
 		if (optlen != sizeof(bool))
 			return -FI_EINVAL;
-		ret = efa_rdm_ep_set_sendrecv_in_order_aligned_128_bytes(efa_rdm_ep, *(bool *)optval);
-		if (ret)
-			return ret;
+		/*
```
wenduwan, line 1557: nit Alignment

prov/efa/test/efa_unit_test_common.c
```diff
@@ -136,7 +138,24 @@ void efa_unit_test_resource_construct_ep_not_enabled(struct efa_resource *resour
...
 	efa_unit_test_resource_construct_with_hints(resource, ep_type,
-						    resource->hints, false);
+						    resource->hints, false, true);
+	return;
+
+err:
+	efa_unit_test_resource_destruct(resource);
+
+	/* Fail test early if the resource struct fails to initialize */
+	assert_int_equal(1, 0);
```
wenduwan, line 148: cmocka provides `fail();`

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: Same, you need to a cross referencing. Application needs to poll both tx and rx ibv cq even if it only read OFI tx cq. This complexity will be removed in OFI 2.0, which proposes to only allow bind 1 cq per endpoint for both tx and rx operations

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: fixed.

prov/efa/test/efa_unit_test_common.c
shijin-aws: Thanks, updated.

prov/efa/src/efa_cq.h
shijin-aws: Good catch, the `entry` should be removed. It was added  when I experimented another approach to implement that ibv_cq_poll_list but I failed to clean up it.

prov/efa/src/efa_cq.h
```diff
@@ -3,6 +3,82 @@
...
+int efa_ibv_cq_poll_list_match(struct dlist_entry *entry, const void *cq)
+{
+	struct efa_ibv_cq_poll_list_entry *item;
+	item = container_of(entry, struct efa_ibv_cq_poll_list_entry, entry);
+	return (item->cq == cq);
+}
+
+/* Serialization must be provided by the caller. */
+static inline
+int efa_ibv_cq_poll_list_search(struct dlist_entry *poll_list, struct efa_ibv_cq *cq)
```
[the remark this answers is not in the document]
shijin-aws: Updated

prov/efa/src/efa_cntr.c
```diff
@@ -159,11 +182,17 @@ int efa_cntr_open(struct fid_domain *domain, struct fi_cntr_attr *attr,
...
 
-	ret = ofi_cntr_init(&efa_prov, domain, attr, &cntr->util_cntr,
-			    &ofi_cntr_progress, context);
+	if (efa_domain->info->ep_attr->type == FI_EP_RDM)
+		ret = ofi_cntr_init(&efa_prov, domain, attr, &cntr->util_cntr,
+				    &efa_rdm_cntr_progress, context);
+	else
+		ret = ofi_cntr_init(&efa_prov, domain, attr, &cntr->util_cntr,
+				    &ofi_cntr_progress, context);
+
```
darrylabbate, line 195: nit for DRY/maintainability (plus there's an existing `typedef`):

```suggestion
	ofi_cntr_progress_func cntr_progress_func = efa_domain->info->ep_attr->type == FI_EP_RDM
		? efa_rdm_cntr_progress
		: ofi_cntr_progress;

	ret = ofi_cntr_init(&efa_prov, domain, attr, &cntr->util_cntr,
			    &cntr_progress_func, context);

```

prov/efa/src/efa_cq.h
```diff
@@ -3,6 +3,77 @@
...
+	struct efa_ibv_cq_poll_list_entry *item;
+
+	ofi_genlock_lock(lock);
+	entry = dlist_find_first_match(poll_list, efa_ibv_cq_poll_list_match, cq);
+	if (entry) {
+		ret = -FI_EALREADY;
+		goto out;
+	}
+
+	item = calloc(1, sizeof(*item));
```
darrylabbate, line 46: Is `calloc()` necessary over regular `malloc()`?

prov/efa/src/efa_base_ep.c
```diff
@@ -149,34 +157,52 @@ static int efa_base_ep_modify_qp_rst2rts(struct efa_base_ep *base_ep,
...
-int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
-			  struct ibv_qp_init_attr_ex *init_attr_ex)
+/**
+ * @brief Create a efa_qp
+ *
+ * @param qp double pointer of struct efa_qp
+ * @param init_attr_ex ibv_qp_init_attr_ex
+ * @return int 0 on success, negative integer on failure
+ */
+int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex)
```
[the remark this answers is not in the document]
shijin-aws: I followed the pattern in other places that an input pointer being allocated inside the function should be passed as `**`. What is your preferred approach here? Making qp a returned pointer?

prov/efa/src/efa_base_ep.c
```diff
@@ -149,34 +157,52 @@ static int efa_base_ep_modify_qp_rst2rts(struct efa_base_ep *base_ep,
...
+	return FI_SUCCESS;
+}
+
+int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
+			  struct ibv_qp_init_attr_ex *init_attr_ex)
+{
+	struct efa_qp *qp;
+	int ret;
+
+	ret = efa_qp_create(&qp, init_attr_ex);
```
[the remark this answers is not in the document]
shijin-aws: I think it's better to use `&base_ep->qp` directly here. But we need to make sure `*qp` is reset to NULL in the error path in `efa_qp_create`. I updated that function as well for that sake.

prov/efa/src/efa_cntr.c
shijin-aws: Thanks! updated.

prov/efa/src/efa_cq.h
shijin-aws: I don't think it's necessary, I followed other similar functions that allocate the list entry.

prov/efa/src/efa_cq.h
```diff
@@ -3,6 +3,77 @@
...
+int efa_ibv_cq_poll_list_match(struct dlist_entry *entry, const void *cq)
+{
+	struct efa_ibv_cq_poll_list_entry *item;
+	item = container_of(entry, struct efa_ibv_cq_poll_list_entry, entry);
+	return (item->cq == cq);
+}
+
+
+static inline
+int efa_ibv_cq_poll_list_insert(struct dlist_entry *poll_list, struct ofi_genlock *lock, struct efa_ibv_cq *cq)
```
sunkuamzn, line 33: Could move this logic to `ofi_list.h` as `dlist_insert_if_not_in_list`. The name of the function is `efa_ibv_cq_poll_list_insert` but the actual logic is inserting if the item is not in the list.

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -767,18 +796,76 @@ static inline
...
+		if (rx_cq)
+			efa_rdm_cq_poll_ibv_cq(-1, &rx_cq->ibv_cq);
 		efa_rdm_ep_progress_internal(efa_rdm_ep);
 	}
 
 	ofi_genlock_unlock(srx_ctx->lock);
 }
 
+static inline
+void efa_rdm_ep_remove_cntr_ibv_cq_poll_list(struct efa_rdm_ep *ep)
```
sunkuamzn, line 821: Do we have tests for the counter logic?

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -27,30 +66,34 @@ static
...
-	attr_ex.cap.max_send_wr = ep->base_ep.domain->device->rdm_info->tx_attr->size;
-	attr_ex.cap.max_send_sge = ep->base_ep.domain->device->rdm_info->tx_attr->iov_limit;
-	attr_ex.send_cq = ibv_cq_ex_to_cq(ep->ibv_cq_ex);
+	tx_rdm_cq = efa_rdm_ep_get_tx_rdm_cq(ep);
+	rx_rdm_cq = efa_rdm_ep_get_rx_rdm_cq(ep);
 
-	attr_ex.cap.max_recv_wr = ep->base_ep.domain->device->rdm_info->rx_attr->size;
-	attr_ex.cap.max_recv_sge = ep->base_ep.domain->device->rdm_info->rx_attr->iov_limit;
-	attr_ex.recv_cq = ibv_cq_ex_to_cq(ep->ibv_cq_ex);
+	if (!tx_rdm_cq && !rx_rdm_cq) {
```
sunkuamzn, line 75: do we need this check? Is it possible to create an endpoint not bound to either tx or rx cq and without send or receive caps?

prov/efa/src/efa_cq.h
```diff
@@ -3,6 +3,77 @@
...
+	IBV_CQ,
+	EFADV_CQ
+};
+
+struct efa_ibv_cq {
+	struct ibv_cq_ex *ibv_cq_ex;
+	enum ibv_cq_ex_type ibv_cq_ex_type;
+};
+
+struct efa_ibv_cq_poll_list_entry {
```
sunkuamzn, line 16: what's the advantage of this separate entry struct instead of adding `dlist_entry` to `efa_ibv_cq`?

sunkuamzn (commented):
A few comments from a quick review

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -15,6 +14,46 @@
...
+	attr_ex->pd = efa_rdm_ep_domain(ep)->ibv_pd;
+	attr_ex->qp_context = ep;
+	attr_ex->sq_sig_all = 1;
+
+	attr_ex->send_cq = ibv_cq_ex_to_cq(tx_cq);
+	attr_ex->recv_cq = ibv_cq_ex_to_cq(rx_cq);
+}
+
+static inline
+struct efa_rdm_cq *efa_rdm_ep_get_tx_rdm_cq(struct efa_rdm_ep *ep)
```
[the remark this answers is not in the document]
shijin-aws: I have `rdm` explicitly to distinguish with ibv cq

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: You cannot create a QP without any CQ. It's not required before because we create IBV CQ as part of EP, but now since we need to bind the CQ to ep explicitly we need this check

prov/efa/src/efa_cq.h
shijin-aws: Because the same efa_ibv_cq can be added to different lists (tx cq's poll list and rx cq's poll list). I find if I add a dlist_entry in efa_ibv_cq and make it be inserted directly, it will mess up and cause data corruption

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1384,16 +1564,26 @@ static int efa_rdm_ep_setopt(fid_t fid, int level, int optname,
...
 			return -FI_EINVAL;
-		ret = efa_rdm_ep_set_sendrecv_in_order_aligned_128_bytes(efa_rdm_ep, *(bool *)optval);
-		if (ret)
-			return ret;
+		/*
+		 * RDMA read is used to copy data from host bounce buffer to the
+		 * application buffer on device
+		 */
+		if (*(bool *)optval) {
+			ret = efa_rdm_ep_check_qp_in_order_aligned_128_bytes(efa_rdm_ep, IBV_WR_RDMA_READ);
```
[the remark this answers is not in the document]
shijin-aws: yes, because the delivery of data on receiver side is via rdma-read

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -100,6 +440,30 @@ static struct fi_ops_cq efa_rdm_cq_ops = {
...
+	struct util_ep *ep;
+	struct fid_list_entry *fid_entry;
+	struct dlist_entry *item;
+	struct efa_rdm_cq *efa_rdm_cq;
+	struct efa_ibv_cq_poll_list_entry *poll_list_entry;
+
+	ofi_genlock_lock(&cq->ep_list_lock);
+	efa_rdm_cq = container_of(cq, struct efa_rdm_cq, util_cq);
+
+	dlist_foreach(&efa_rdm_cq->ibv_cq_poll_list, item) {
```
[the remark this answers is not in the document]
shijin-aws: Yeah I know... But I haven't got a better solution right now as I commented in another place.. I would make it a future work and track it somewhere

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: OK, will add

prov/efa/src/efa_cq.h
shijin-aws: We need to handle different type of data structure, I would make this logic a generalized macro and make both fid_list and efa_ibv_cq_poll_list onboard it.

prov/efa/src/rdm/efa_rdm_cq.c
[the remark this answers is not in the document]
shijin-aws: I talked with @a-szegel , I think reusing his rcu list implementation will help here: https://github.com/a-szegel/libfabric/blob/modified-rcu-cq-ep-list-lock-attempt-3/include/ofi_rcu.h

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_cq.c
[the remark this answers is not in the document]
shijin-aws: Thinking about it again, we should be able to keep the dlist implementation, we just need to allocate the dlist entry in a continuous heap. We can do that via ofi_bufpool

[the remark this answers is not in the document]
shijin-aws: Thinking about it again, this poll list will only have 1 entry for MPI and ofi nccl plugin today. And it is proposed in ofi 2.0 that each ep can only be bound to 1 CQ (for both send and recv), so I would like to have a trade-off here to make it simple as is and get rid of it in ofi 2.0. But I will document it somewhere

prov/efa/src/efa_cq.h
shijin-aws: Would make this as a separate commit

prov/efa/src/efa_cq.h
sunkuamzn: OK instead of adding multiple dlist_entry to efa_ibv_cq, you create a separate struct here. Makes sense.

sunkuamzn (approved):

darrylabbate (approved):

PR #9812, opened 2024-02-14
prov/efa: Handle rx pkts error without ope

shijin-aws:
When getting the error with IBV_WC_RECV op
code, it is possible that packet entry is not associated with an ope yet, if the packet is posted by
efa_rdm_ep_bulk_post_internal_rx_pkts().

For such error, we can only warn and write eq
error as it is an EFA internal error and not
associated with application's rx operation.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke_cmd.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_cmd.c
@@ -673,6 +673,26 @@ void efa_rdm_pke_handle_rx_error(struct efa_rdm_pke *pkt_entry, int err, int pro
 	EFA_DBG(FI_LOG_CQ, "Packet receive error: %s (%d)\n",
 	        efa_strerror(prov_errno), prov_errno);
 
+	/*
+	 * pkes posted by efa_rdm_ep_bulk_post_internal_rx_pkts
+	 * are not associated with ope before being progressed
+	 */
+	if (!pkt_entry->ope) {
+		char ep_addr_str[OFI_ADDRSTRLEN];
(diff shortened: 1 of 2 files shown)


prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -673,6 +673,22 @@ void efa_rdm_pke_handle_rx_error(struct efa_rdm_pke *pkt_entry, int err, int pro
 	EFA_DBG(FI_LOG_CQ, "Packet receive error: %s (%d)\n",
 	        efa_strerror(prov_errno), prov_errno);
 
+	if (!pkt_entry->ope) {
+		char ep_addr_str[OFI_ADDRSTRLEN];
```
wenduwan, line 677: Suggest adding a comment as reminder, e.g. the error might come from write with immediate data.

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_pke_cmd.c
shijin-aws: it can be from both RECV and RECV_WITH_IMM. Don't think we need to call immediate data specially. But I will add comments to describe it

a-szegel (approved):

prov/efa/src/rdm/efa_rdm_pke_cmd.c
shijin-aws: Updated

wenduwan (approved):

PR #9817, opened 2024-02-15
prov/efa: Remove unnecessary check in rdma write.

shijin-aws:
efa_both_support_rdma_write already covers the
check for self peer. No need to check it explicitly again. Also remove the assertion because we don't need a handshake if the peer is self.


a-szegel (approved):
LGTM!

sunkuamzn (approved):

PR #9818, opened 2024-02-15
prov/shm: Fix Various Lock Related Coverity Issues

zachdworkin:


prov/shm/src/smr_signal.h
```diff
@@ -54,6 +55,7 @@ static void smr_handle_signal(int signum, siginfo_t *info, void *ucontext)
 				sock_name, entry) {
 		unlink(sock_name->name);
 	}
+	pthread_mutex_unlock(&ep_list_lock);
```
chien-intel, line 58: You need to move this up and for sock_name_list use sock_list_lock

zachdworkin: I closed the ep_list_lock around the ep list and used the sock_list_lock around the sock_name_list. Is this correct?

prov/shm/src/smr_signal.h
```diff
@@ -46,14 +46,19 @@ static void smr_handle_signal(int signum, siginfo_t *info, void *ucontext)
 	struct smr_sock_name *sock_name;
 	int ret;
 
+	pthread_mutex_lock(&ep_list_lock);
```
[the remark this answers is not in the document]
shijin-aws: I see, I got it mixed with the ep_list_lock in cq

aingerson (approved):

prov/shm/src/smr_util.c
```diff
@@ -393,9 +393,6 @@ int smr_map_to_region(const struct fi_provider *prov, struct smr_map *map,
 	const char *name = smr_no_prefix(peer_buf->peer.name);
 	char tmp[SMR_PATH_MAX];
 
-	if (peer_buf->region)
-		return FI_SUCCESS;
-
```
chien-intel, line 398: How often do we return FI_SUCCESS from here?  May not be a good idea to remove this check if we hit this all the time.

prov/shm/src/smr_util.c
```diff
@@ -573,15 +573,15 @@ void smr_map_del(struct smr_map *map, int64_t id)
 {
 	struct dlist_entry *entry;
 
-	if (id >= SMR_MAX_PEERS || id < 0 || map->peers[id].peer.id < 0)
-		return;
-
```
chien-intel, line 574: Same comment here, how often do we return from here?  Do not remove if we return from here often.

prov/shm/src/smr_util.c
aingerson: This return should get hit pretty rarely and only on start up. The only case I can think of is if 2 EPs are sharing the same AV and they both get an incoming connection request from the same peer and the threads both try to simultaneously add the peer. It's probably needed but just got moved below the ep name list check to isolate the locking. Ideally it would be above to eliminate the need to go through the EP list but in reality the EP list won't be very long and it would be very rare to get this and it's only on start up so I think it's a reasonable trade off.

prov/shm/src/smr_util.c
aingerson: This one I don't think is ever hit. We should never be deleting an id that isn't valid. But again, it's moved to below so it's still there, just after the ep list check for locking simplicity.

j-xiong:
bot:aws:retest

PR #9819, opened 2024-02-15
prov/util: Add Mem Monitor State Lock to state init

zachdworkin:
Fix this coverity issue.
When initializing state two threads can access the field at the same timie and update it concurrently. Adding this lock prevents this case.

### DIFF
--- a/prov/util/src/util_mem_monitor.c
+++ b/prov/util/src/util_mem_monitor.c
@@ -163,7 +163,9 @@ out:
 void ofi_monitor_init(struct ofi_mem_monitor *monitor)
 {
 	dlist_init(&monitor->list);
+	pthread_mutex_lock(&mm_state_lock);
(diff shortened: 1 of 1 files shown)


aingerson (approved):
Seems fine to suppress the coverity warning but it feels dumb to have to lock the state on initialization...  @j-xiong Do you have any thoughts?

j-xiong:
I would dismiss this Coverity warning as "intentional" with a comment like "This is part of initialization code that won't be executed currently with other threads accessing the monitor state."

zachdworkin:
@j-xiong should I close this PR since we will mark coverity on its website?

j-xiong:
@zachdworkin Yes.

zachdworkin:
This change is not necessary. We will mark coverity as resolved on its website to avoid this unnecessary change.

PR #9824, opened 2024-02-17
prov/efa: Decouple efa av entry and efa_rdm_peer.

shijin-aws:
Test CI


shijin-aws:
I will close this PR as it's not needed in the near future.

PR #9825, opened 2024-02-18
src/var:Avoiding duplicate parameter insertion

liangchaoxi:
Preventing duplicate parameter insertion

### DIFF
--- a/src/var.c
+++ b/src/var.c
@@ -191,6 +191,7 @@ int DEFAULT_SYMVER_PRE(fi_param_define)(const struct fi_provider *provider,
 {
 	int i, ret;
 	struct fi_param_entry *v;
+	struct dlist_entry *entry;
 	char *tmp_str;
 	va_list vargs;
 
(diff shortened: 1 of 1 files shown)


src/var.c
```diff
@@ -216,6 +217,14 @@ int DEFAULT_SYMVER_PRE(fi_param_define)(const struct fi_provider *provider,
 	v->name = strdup(param_name);
 	v->type = type;
 
+	entry = dlist_find_first_match(&param_list,param_match_name, v);
```
j-xiong, line 220: where is `param_match_name` defined?

src/var.c
```diff
@@ -216,6 +217,14 @@ int DEFAULT_SYMVER_PRE(fi_param_define)(const struct fi_provider *provider,
...
+
+        if (entry) {
+          FI_INFO(provider, FI_LOG_CORE,
+                "%s variable alredy exist\n",param_name);
+          fi_free_param(v);
+          return FI_SUCCESS;
+        }
```
j-xiong, line 227: Please use tabs instead of spaces for indentation.

src/var.c
```diff
@@ -216,6 +217,14 @@ int DEFAULT_SYMVER_PRE(fi_param_define)(const struct fi_provider *provider,
 	v->name = strdup(param_name);
 	v->type = type;
 
+	entry = dlist_find_first_match(&param_list,param_match_name, v);
+
+        if (entry) {
+          FI_INFO(provider, FI_LOG_CORE,
+                "%s variable alredy exist\n",param_name);
+          fi_free_param(v);
+          return FI_SUCCESS;
```
shefty, line 226: I would consider this an error on the part of the provider.  If we want to allow duplicate insertions, in order to return success, the rest of the fields should match., but at least the type.

PR #9826, opened 2024-02-20
Fabtests: Resolve various coverity issues

Juee14Desai:
Multiple coverity issues are addressed in this commit.

Most of them are return value check type and some are uninitialized variable type of issues.

Total 16 issues are addressed.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -2913,7 +2913,7 @@ void eq_readerr(struct fid_eq *eq, const char *eq_str)
 
 int ft_sync()
 {
-	char buf;
+	char buf = 'a';
 	int ret;
 
 	if (opts.dst_addr) {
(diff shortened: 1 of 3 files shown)


fabtests/component/sock_test.c
```diff
@@ -470,7 +470,9 @@ static int run_server(void)
 	if (ret)
 		return ret;
 
-	send(fds[0], &c, 1, 0);
+	ret = send(fds[0], &c, 1, 0);
+	if (ret < 0)
+		return -EINVAL;
```
j-xiong, line 475: Maybe use `return -errno;` instead?

fabtests/common/shared.c
```diff
@@ -2913,7 +2913,7 @@ void eq_readerr(struct fid_eq *eq, const char *eq_str)
 
 int ft_sync()
 {
-	char buf;
+	char buf[8];
```
j-xiong, line 2916: This doesn't initialize buf.  Since only one byte is used, something like `char buf = 'a';` should work.

j-xiong (commented):
I would suggest consolidating these into a few commits: 
(1) suppress complaint about missing return value checking by adding `(void)` cast
(2) fix missing return value checking by handling the return value (including using FT_CLOSE_FID)
(3) fix complaint about uninitialized variables

j-xiong:
bot:aws:retest

j-xiong (approved):

PR #9827, opened 2024-02-20
fabtests/common,multi_ep: Allow testing multi ep with shared/non-shared cqs and avs

shijin-aws:
Currently, multi_ep can only test with shared av but non-shared tx/rx cq among multiple eps. This PR contains a set of commits to allow testing multi ep with shared and non-shared cqs and avs, by toggling the `--shared-cq` and `--shared-av` options.


fabtests/functional/multi_ep.c
```diff
@@ -125,7 +125,7 @@ static int ep_post_rx(int idx)
 		ret = fi_recv(eps[idx], recv_bufs[idx], opts.transfer_size,
 			      data_desc, FI_ADDR_UNSPEC, &recv_ctx[idx]);
 		if (ret == -FI_EAGAIN)
-			(void) fi_cq_read(rxcqs[idx], NULL, 0);
+			(void) fi_cq_read(rxcqs[0], NULL, 0);
```
j-xiong: why?

fabtests/functional/multi_ep.c
```diff
@@ -207,19 +228,25 @@ static int do_transfers(void)
...
 		return ret;
 	}
 
-	ret = ft_alloc_ep_res(fi, &txcqs[idx], &rxcqs[idx], NULL, NULL, NULL, &av);
+	ret = ft_alloc_ep_res(fi, &txcqs[idx], &rxcqs[idx], NULL, NULL, NULL, &avs[idx]);
 	if (ret)
 		return ret;
 
-	ret = ft_enable_ep(eps[idx], eq, av, txcqs[idx], rxcqs[idx],
+	ret = ft_enable_ep(eps[idx], eq, avs[av_bind_idx], txcqs[cq_bind_idx], rxcqs[idx],
```
j-xiong, line 249: should use `cq_bind_idx` for `rxcqs` as well.

j-xiong (commented):
Found one more:

fabtests/functional/multi_ep.c
shijin-aws: Oops... Fixed!

j-xiong (approved):

fabtests/functional/multi_ep.c
```diff
@@ -390,12 +437,24 @@ int main(int argc, char **argv)
...
+		case LONG_OPT_SHARE_CQ:
+			share_cq = true;
+			break;
 		case '?':
 		case 'h':
 			ft_usage(argv[0], "Multi endpoint test");
 			FT_PRINT_OPTS_USAGE("-c <int>",
 				"number of endpoints to create and test (def 3)");
 			FT_PRINT_OPTS_USAGE("-v", "Enable data verification");
+			FT_PRINT_OPTS_USAGE("--share-cq",
```
aingerson, line 452: There is another test that toggles shared resources and an existing option for sharing the CQ. They all use "shared" instead of "share" - could you change these to "shared" to align with those please?

aingerson (commented):
Just one small suggestion. Fun addition to the test! Thanks!

fabtests/functional/multi_ep.c
shijin-aws: Updated

fabtests/functional/multi_ep.c
```diff
@@ -207,19 +228,25 @@ static int do_transfers(void)
...
+		av_bind_idx = 0;
 
 	ret = fi_endpoint(domain, fi, &eps[idx], NULL);
 	if (ret) {
 		FT_PRINTERR("fi_endpoint", ret);
 		return ret;
 	}
 
-	ret = ft_alloc_ep_res(fi, &txcqs[idx], &rxcqs[idx], NULL, NULL, NULL, &av);
+	ret = ft_alloc_ep_res(fi, &txcqs[idx], &rxcqs[idx], NULL, NULL, NULL, &avs[idx]);
```
aingerson, line 245: Just to clarify intent here - we still create all CQs but then only bind to the first one if the user wants to share it?

aingerson (approved):
Just a quick clarification on intent but otherwise looks good!

fabtests/functional/multi_ep.c
shijin-aws: Yes, that allows for minimal code change. I can also do the way that only create 1 cq/av when shared, if you prefer that.

aingerson: No need, just wanted to clarify that that was intended. Thanks!

PR #9829, opened 2024-02-21
prov/util: Set import monitor state to idle upon close

belynam:
When the import monitor is closed, its state is not updated. Then later, during provider cleanup when ofi_monitors_del_cache() is called, and the last cache is removed from the import monitor's cache list, the monitor's state is seen as running, and an attempt is made to stop it. Since the close already set impmon.impfid to NULL, this results in a failed assert/segfault.

This change sets the import monitor's state to idle when closed. In ofi_monitors_del_cache(), only update a monitor when the monitor's state needs updating.


j-xiong (approved):

shijin-aws:
bot:aws:retest

j-xiong:
bot:aws:retest

PR #9830, opened 2024-02-22
Fix ft_osd includes to point to fabtests/include

dariuszsciebura:
E.g. #include <unix/osd.h> points to libfabric include, while the demanded include file is located in fabtests/include.

Signed-off-by: Dariusz Ściebura <dariusz.sciebura@gmail.com>


dariuszsciebura:
wrong commit msg

PR #9833, opened 2024-02-22
core: Add new CXI endpoint protocol FI_PROTO_CXI_RNR

swelch:
Reserve new Receiver-Not-Ready endpoint protocol for CXI provider.

### DIFF
--- a/man/fi_endpoint.3.md
+++ b/man/fi_endpoint.3.md
@@ -753,6 +753,15 @@ protocol value set to one.
 : Protocol for intra-node communication using shared memory segments
   used by the sm2 provider
 
+*FI_PROTO_CXI*
+: Reliable-datagram protocol optimized for HPC applications
+  used by cxi provider.
+
+*FI_PROTO_CXI_RNR*
+: A version of the FI_PROTO_CXI protocol that implements an RNR
+  protocol which can be used when messaging is primarily expected
+  and FI_ORDER_SAS ordering is not required.
+
 *FI_PROTO_UNSPEC*
 : The protocol is not specified.  This is usually provided as input,
   with other attributes of the socket or the provider selecting the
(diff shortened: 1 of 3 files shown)


man/fi_endpoint.3.md
```diff
@@ -753,6 +753,15 @@ protocol value set to one.
...
   used by the sm2 provider
 
+*FI_PROTO_CXI*
+: Reliable-datagram protocol optimized for HPC applications
+  used by cxi provider.
+
+*FI_PROTO_CXI_RNR*
+: A version of the FI_PROTO_CXI protocol that implements an RNR
+  protocol which can be used when messaging is primarily expected
+  and FI_ORDER_SAS ordering is not required.
```
shefty, line 763: Is this a different protocol or a slight modification to the CXI protocol based on the endpoint attributes?  You call out SAS ordering, so I'm wondering if that is sufficient for the endpoint to determine if RNR messages can be sent.  What actually happens if one side chooses CXI but the other CXI_RNR?  Can they still talk?

swelch: This is a significantly different protocol with different hardware match bits formats and messaging protocols that are supported in addition to the ordering changes. While the FI_PROTO_CXI supports FI_ORDER_SAS it does not require it; it offers many other features that this new protocol does not (rendezvous messaging for instance).

If the two sides select different protocols, they cannot physically communicate. The protocols map to a different receive portals table entries, so they are physically disconnected. If a process selected this protocol and targeted an EP that did not use the protocol, the sender would get an error completion without interaction from software at the target.

shefty (approved):

swelch:
@j-xiong @jswaro  I don't see how the CI failures could be related to this PR, are you OK for a merge?

j-xiong:
The ci failure is unrelated.

PR #9834, opened 2024-02-22
prov/shm: Fix new coverity issue introduced by 7f23127

zachdworkin:
Must assert that the id is within the allowable range before using it to check a peer's name

### DIFF
--- a/prov/shm/src/smr_util.c
+++ b/prov/shm/src/smr_util.c
@@ -572,13 +572,15 @@ void smr_map_del(struct smr_map *map, int64_t id)
 {
 	struct dlist_entry *entry;
 
+	assert(id >= 0 && id < SMR_MAX_PEERS);
+
 	pthread_mutex_lock(&ep_list_lock);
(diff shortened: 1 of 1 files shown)


aingerson (commented):
Can you add something in the commit message about how no code paths leading to here should be deleting an out of bounds index? Thanks!

j-xiong:
Is it for https://scan4.scan.coverity.com/#/project-view/61947/10344?selectedIssue=414886? I have dismissed it as "false positive" because if id < 0 the '||' operator would shortcut the if statement and the array access would not happen.

zachdworkin:
@j-xiong what do you want done with this PR then?

j-xiong:
I would just close it.

aingerson:
@j-xiong It isn't a false positive. The out of bounds array access happens in the first if before the check

j-xiong:
Hmm, I missed this one: `smr_no_prefix(map->peers[id].peer.name)`.

PR #9836, opened 2024-02-22
Cornelis Networks OPX provider upstream

charlesshereda:
Commits include:

-Cuda and GPU-specific improvements
-Excluding empty immediate date 
-Improving HFI (NIC) selection logic
-Small refactoring and bug fixes
-Memory leak fixes.

### DIFF
--- a/prov/opx/include/rdma/opx/fi_opx_hfi1_sdma.h
+++ b/prov/opx/include/rdma/opx/fi_opx_hfi1_sdma.h
@@ -59,15 +59,6 @@ static const uint16_t OPX_SDMA_REQ_SET_MEMINFO[2] = {0,
 static const size_t OPX_SDMA_REQ_HDR_SIZE[2] = {FI_OPX_HFI1_SDMA_HDR_SIZE,
 						FI_OPX_HFI1_SDMA_HDR_SIZE + OPX_SDMA_MEMINFO_SIZE};
 
-static const unsigned OPX_SDMA_OFI_TO_KERN_MEM_TYPE[4] = {
-					#ifdef OPX_HMEM
-							  HFI1_MEMINFO_TYPE_SYSTEM,
-							  HFI1_MEMINFO_TYPE_NVIDIA,
-							  2, /* HFI1_MEMINFO_TYPE_AMD */
-							  1 /* HFI1_MEMINFO_TYPE_DMABUF */
-					#endif
-							  };
-
 struct fi_opx_hfi1_sdma_header_vec {
 	union {
 		struct {
@@ -515,7 +506,7 @@ void fi_opx_hfi1_sdma_set_meminfo(struct sdma_req_info *req_info,
 		// setting meminfo, and it will be the fist one:
 		// index 0 (the first payload IOV, or iov[1]).
 		const unsigned meminfo_idx = 0;
-		const unsigned type = OPX_SDMA_OFI_TO_KERN_MEM_TYPE[iface];
+		const unsigned type = OPX_HMEM_KERN_MEM_TYPE[iface];
 		struct sdma_req_meminfo *meminfo = (struct sdma_req_meminfo *) (req_info + 1);
 		meminfo->types = 0;
 		HFI1_MEMINFO_TYPE_ENTRY_SET(meminfo->types, meminfo_idx, type);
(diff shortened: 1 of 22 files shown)


prov/opx/include/rdma/opx/fi_opx_hmem.h
```diff
@@ -83,16 +66,44 @@ enum fi_hmem_iface fi_opx_hmem_get_iface(const void *ptr,
...
+
+			if (mem_type == CU_MEMORYTYPE_DEVICE && !is_managed) {
+				*device = device_ordinal;
+				return FI_HMEM_CUDA;
+			}
+		} else if (cuda_rc != CUDA_ERROR_INVALID_CONTEXT) {
+			FI_DBG_TRACE(fi_opx_global.prov, FI_LOG_EP_DATA,
+				"Bad return code %hu from cuPointerGetAttributes()",
+				cuda_rc);
+		}
```
j-xiong, line 99: Can this change be applied to `cuda_is_addr_valid` (in `src/hmem_cuda.c`) instead?

prov/opx/include/rdma/opx/fi_opx_hfi1_sdma.h
```diff
@@ -212,14 +212,14 @@ void fi_opx_hfi1_dput_sdma_init(struct fi_opx_ep *opx_ep,
...
-				      (params->opcode == FI_OPX_HFI_DPUT_OPCODE_GET) ||
-				      (params->opcode == FI_OPX_HFI_DPUT_OPCODE_PUT) ||
-				      (params->opcode == FI_OPX_HFI_DPUT_OPCODE_ATOMIC_FETCH) ||
-				      (params->opcode == FI_OPX_HFI_DPUT_OPCODE_ATOMIC_COMPARE_FETCH);
+	params->sdma_no_bounce_buf = (length >= opx_ep->tx->sdma_bounce_buf_threshold) ||
+                            	(is_hmem) ||
+                            	(params->opcode == FI_OPX_HFI_DPUT_OPCODE_GET) ||
+                            	(params->opcode == FI_OPX_HFI_DPUT_OPCODE_PUT) ||
+                            	(params->opcode == FI_OPX_HFI_DPUT_OPCODE_ATOMIC_FETCH) ||
+                            	(params->opcode == FI_OPX_HFI_DPUT_OPCODE_ATOMIC_COMPARE_FETCH);
```
j-xiong, line 211: Looks line some tabs are accidentally converted to spaces here. Please use tabs as long as possible.

prov/opx/src/fi_opx_reliability.c
```diff
@@ -2690,19 +2690,20 @@ void fi_opx_reliability_client_fini (struct fi_opx_reliability_client_state * st
...
 	if (state->rx_flow_rbtree) {
 		rbtDelete(state->rx_flow_rbtree);
+		state->rx_flow_rbtree = NULL;
 	}
 
 	if(state->tx_flow_rbtree) {
 		rbtDelete(state->tx_flow_rbtree);
-	}*/
+		state->tx_flow_rbtree = NULL;
+	}
```
j-xiong, line 2706: Are these changes resolving memory leaks? Looks more like resolving double free.

prov/opx/src/fi_opx_hfi1.c
```diff
@@ -315,13 +318,21 @@ struct fi_opx_hfi1_context *fi_opx_hfi1_context_open(struct fid_ep *ep, uuid_t u
...
 	int hfi_candidates_count = 0;
 	int hfi_candidate_index = -1;
 	struct _hfi_ctrl *ctrl = NULL;
 	bool use_default_logic = true;
+	int dirfd = -1;
 
 	struct fi_opx_hfi1_context_internal *internal =
 		calloc(1, sizeof(struct fi_opx_hfi1_context_internal));
+	if (!internal)
+	{
```
j-xiong, line 336: Put the `{` at the end of the previous line.

prov/opx/include/rdma/opx/fi_opx_hmem.h
charlesshereda: @belynam will you respond?

belynam: > Can this change be applied to `cuda_is_addr_valid` (in `src/hmem_cuda.c`) instead?

I can port the change to `cuda_is_addr_valid` to get multiple attributes at once instead of making separate calls to get each attribute one at a time. However, `cuda_is_addr_valid` does not currently check for managed/unified memory, and if we wanted it to, it's unclear to me how best to add that check and communicate back to the user the managed status without changing the existing API (perhaps not turning on the `FI_HMEM_DEVICE_ONLY` flag?).

Even if `cuda_is_addr_valid` provided all that we need, I'm not sure we'd be able to make use of it without incurring a performance hit. With the current  `ofi_get_hmem_iface()`, we'd be stacking at least two more function calls to get to `cuda_is_addr_valid()`, and also looping through the 5 other possible device types before we get to Nvidia/Cuda on the last iteration of the for-loop.

I can take a crack at it and open a separate PR to update `cuda_is_add_valid()`, and then assuming that is approved/merged, we can update OPX code to try it. But for this PR, I think we'd like to keep the code here as is.

belynam: @j-xiong See #9844

prov/opx/src/fi_opx_reliability.c
belynam: > Are these changes resolving memory leaks? Looks more like resolving double free.

Both. Note the removal of the trailing '*/' on line 2705. This code to delete these rb trees was uncommented. Adding the assignments to NULL is also to prevent double frees.

PR #9840, opened 2024-02-25
prov/hook: Proposal for updating the hook provider

amirshehataornl:
As far as I could tell the hook provider dumps all the trace to the console, which is a lot to look at. This is a proposal to store the tracing information internally and then dump it when shutting down.

This patch makes a change to how the send/recv operation tracing is done.

This example code stores a record in a database. The record has a key consistent of addr/data len/operation. This way we can keep track of the number of operations of a specific len per address. I then dump it to a csv file, which is a lot easier to parse. Typically we don't need to see each operation on the console. A summary statistics of the job would be more useful (IMHO)

This is just a proposal to get some feedback.

### DIFF
--- a/include/ofi_hook.h
+++ b/include/ofi_hook.h
@@ -45,6 +45,7 @@
 #include <rdma/fi_rma.h>
 #include <rdma/fi_tagged.h>
 
+#include <ofi_tree.h>
 #include <ofi.h>
 #include <rdma/providers/fi_prov.h>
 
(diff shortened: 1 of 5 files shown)


amirshehataornl:
That's true. Maybe we should give the user the option through an environment variable. I guess the scenario I'm trying to address is the ability to collect statistics from a run, rather than tracing the send/recv calls. They are two different workflows, but can be handled within the same logic. What do you think?

j-xiong:
The traces are generated by the "trace" hooking provider. There is another hooking provider called "profile" which produces statistics output. We should look at both for potential improvements targeted for different needs.

amirshehataornl:
hmmm... I mixed up the profile and perf in my head and thought it only collects performance only. Profile does what I was thinking... sorry for the churn.

amirshehataornl:
you can go ahead and close this pull request.

PR #9842, opened 2024-02-29
prov/efa, fabtests/efa: Do not progress cq inside transmission call when hitting EAGAIN

shijin-aws:
It's application's responsibility to poll cq
    when hitting EAGAIN, efa provider shouldn't have
    to progress the ep again inside the transmission
    call.

Also fix the EAGAIN handling in rnr_cq_read_error test

### DIFF
--- a/fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
+++ b/fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
@@ -60,7 +60,12 @@ static int rnr_read_cq_error(void)
 	for (i = 0; i < total_send; i++) {
 		do {
 			ret = fi_send(ep, tx_buf, 32, mr_desc, remote_fi_addr, &tx_ctx);
-			if (ret < 0 && ret != -FI_EAGAIN) {
+			if (ret == -FI_EAGAIN) {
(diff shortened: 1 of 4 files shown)


fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
```diff
@@ -60,7 +60,12 @@ static int rnr_read_cq_error(void)
 	for (i = 0; i < total_send; i++) {
 		do {
 			ret = fi_send(ep, tx_buf, 32, mr_desc, remote_fi_addr, &tx_ctx);
-			if (ret < 0 && ret != -FI_EAGAIN) {
+			if (ret == -FI_EAGAIN) {
+				fi_cq_read(txcq, NULL, 0);
```
wenduwan, line 64: Why is this cq read necessary?

wenduwan (approved):

fabtests/prov/efa/src/rdm_rnr_read_cq_error.c
shijin-aws: because this test posts rx + rnr_count number of tx, which is way bigger than the default tx queue (4k), tx pkt pool is exhausted and return EAGAIN

PR #9843, opened 2024-02-29
Fabtests Coverity Issues

alex-mckinley:
fix coverity # 414977 and # 414791

### DIFF
--- a/fabtests/unit/getinfo_test.c
+++ b/fabtests/unit/getinfo_test.c
@@ -695,11 +695,16 @@ static int test_caps_regression(char *node, char *service, uint64_t flags,
 	struct fi_info *fi;
 	int ret;
 
+	if (!hints) {
+		printf("invalid test case: hints may not be null");
+		return -FI_EINVAL;
+	}
+
 	ret = fi_getinfo(FT_FIVERSION, node, service, flags, NULL, info);
 	if (ret)
 		return ret;
 
-	if (!hints || !hints->fabric_attr || !hints->fabric_attr->prov_name) {
+	if (!hints->fabric_attr || !hints->fabric_attr->prov_name) {
 		fi = *info;
 	} else {
 		for (fi = *info; fi; fi = fi->next) {


fabtests/prov/efa/src/rdm_rnr_queue_resend.c
```diff
@@ -110,10 +112,15 @@ static int alloc_atomic_res(struct fi_info *fi, void **result, void **compare,
 			0, 0, mr_compare, NULL);
 	if (ret) {
 		FT_PRINTERR("fi_mr_reg", ret);
-		return ret;
+		goto err3;
```
chien-intel, line 115: Not blocking this PR but not sure why Coverity didn't flag this.
Need to unregister mr_result if register mr_compare fails.
Just good practice for a function to clean up after itself on failure.

aingerson: @chien-intel I generally agree but I think fabtests has a convention to never clean up fids/OFI resources in the moment and always clean them up when the parent return happens, closing all the fids at once. I think this can follow the same pattern and let the cleanup handle the fid closing if needed.
That said, I think this test doesn't actually call cleanup properly :) It should be calling free_res on error and add the atomic close call in there as well. ie
this
```
	ret = rnr_queue_resend_test(req_pkt, atomic_op);
	if (ret) {
		FT_PRINTERR("rnr_queue_resend_test", -ret);
		return ret;
	}

	ret = ft_close_oob();
	if (ret) {
		FT_PRINTERR("ft_close_oob", -ret);
		return ret;
	}
	ft_free_res();

	return 0;
```
should be something like
```

static int run(int req_pkt, enum fi_op atomic_op)
{
	int ret;

	ret = ft_efa_rnr_init_fabric();
	if (ret) {
		FT_PRINTERR("ft_efa_rnr_init_fabric", -ret);
		goto out;
	}

	ret = rnr_queue_resend_test(req_pkt, atomic_op);
	if (ret) {
		FT_PRINTERR("rnr_queue_resend_test", -ret);
		goto out;
	}

	ret = ft_close_oob();
	if (ret) {
		FT_PRINTERR("ft_close_oob", -ret);
		return ret;
	}

out:
	free_atomic_res();
	ft_free_res();

	return 0;
}
```
But the atomic resources go out of scope returning from the resend test. Really, I think this test needs some refactoring to handle cleanup better. @shijin-aws @a-szegel Can anyone from AWS handle this cleanup so we don't mess up your test too much?

chien-intel: > @chien-intel I generally agree but I think fabtests has a convention to never clean up fids/OFI resources in the moment and always clean them up when the parent return happens, closing all the fids at once.

It is best I stop commenting here. :-)

shijin-aws: I agree with @aingerson, I will create a separate PR to refactor and handle this.

shijin-aws: Opened https://github.com/ofiwg/libfabric/pull/9848, you can drop this change in this PR

alex-mckinley: change dropped

j-xiong (approved):

PR #9844, opened 2024-02-29
hmem/cuda: Get multiple attributes at once in cuda_is_addr_valid

belynam:
The cuda_is_addr_valid() function makes 2 separate calls to cuPointerGetAttribute() to get the memory type and device. If the caller also wants to know if the cuda buffer is "managed" (i.e. unified memory), a 3rd subsequent call to cuPointerGetAttribute() is necessary (outside of cuda_is_addr_valid()) to retrieve that.

This change gets all three attributes in a single call to cuPointerGetAttributes(). If the address in question points to cuda memory, and that memory is managed/unified, the resulting flags will not have FI_HMEM_DEVICE_ONLY set.

---
Making this change per discussion [here](https://github.com/ofiwg/libfabric/pull/9836#discussion_r1506097433).

Communicating managed/unified status by not setting FI_HMEM_DEVICE_ONLY based on this [discussion](https://github.com/ofiwg/libfabric/discussions/6540).

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -62,6 +62,9 @@ CUresult ofi_cuGetErrorName(CUresult error, const char** pStr);
 CUresult ofi_cuGetErrorString(CUresult error, const char** pStr);
 CUresult ofi_cuPointerGetAttribute(void *data, CUpointer_attribute attribute,
 				   CUdeviceptr ptr);
+CUresult ofi_cuPointerGetAttributes(unsigned int num_attributes,
+				    CUpointer_attribute *attributes,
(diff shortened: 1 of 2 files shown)


j-xiong (approved):
Looks good. Thanks!

belynam:
@sunkuamzn @a-szegel @wzamazon I see you all made changes that involve `FI_HMEM_DEVICE_ONLY`. This PR will change behavior of `cuda_is_addr_valid()` to *not* set `FI_HMEM_DEVICE_ONLY` in `flags` in the case of managed/unified memory. Please review to ensure that is okay. Thank you in advance!

PR #9848, opened 2024-03-01
fabtests/efa: fix coverity issues in rnr_queue_resend test

shijin-aws:
Currently, atomic resources can be not cleaned up when alloc_atomic_res failed. This patch fixes it by always cleaning up all atomic resources once at the parent test function returns.


aingerson (approved):
Thanks for the cleanup!

PR #9849, opened 2024-03-01
fabtests/efa: Fix coverity issue in rnr_read_cq_err

shijin-aws:
Add a (void) before fi_cq_read(txcq, NULL, 0)
as in this case we don't care the return of fi_cq_read.


wenduwan (approved):

shijin-aws:
Fix # 414986

PR #9850, opened 2024-03-01
Src: Fix Coverity Log Issues

zachdworkin:
Add locking to potentially unsafe log_fid.ops functions

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -254,17 +254,17 @@ static int ofi_log_ready(const struct fi_provider *prov,
 			 enum fi_log_level level, enum fi_log_subsys subsys,
 			 uint64_t flags, uint64_t *showtime)
 {
-    uint64_t cur;
+	uint64_t cur;
 
-    if (log_fid.ops->enabled(prov, level, subsys, flags)) {
-	    cur = ofi_gettime_ms();
-	    if (cur >= *showtime) {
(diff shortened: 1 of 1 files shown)


shijin-aws:
@zachdworkin efa fix is already in https://github.com/ofiwg/libfabric/pull/9849

zachdworkin:
@shijin-aws Thanks, I missed that PR. I removed the patch that had that change in it.

src/log.c
```diff
@@ -323,11 +323,15 @@ int DEFAULT_SYMVER_PRE(fi_log_enabled)(const struct fi_provider *prov,
...
 	uint64_t flags = 0;
+	int ret = 0;
 
 	if (ofi_prov_ctx(prov)->disable_logging)
 		flags |= FI_LOG_PROV_FILTERED;
 
-	return log_fid.ops->enabled(prov, level, subsys, flags);
+	pthread_mutex_lock(&common_locks.ini_lock);
+	ret = log_fid.ops->enabled(prov, level, subsys, flags);
+	pthread_mutex_unlock(&common_locks.ini_lock);
```
shefty, line 333: This introduces significant overhead to logging, even when it's disabled.  We don't want any of these locks in the logging calls

zachdworkin: Thanks! I was noticing that so I added the evaluating label. Do you have a different fix suggestion?

PR #9851, opened 2024-03-01
contrib/intel/jenkins: Add cuda fabtests to CI

zachdworkin:
Add cuda + fabtests tests to CI. These will be the same tests that are run through shm + ze.


Juee14Desai (approved):
Looks good. Good job!

PR #9855, opened 2024-03-04
configure: Do not check for xpmem if disabled

raffenet:
If --disable-xpmem is passed to configure, skip checking. Fixes a build issue when xpmem is located in the default system directories.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -426,14 +426,16 @@ AC_ARG_ENABLE([xpmem],
 			      PATH: enable xpmem and use xpmem installed under PATH)])],
 	      )
 
-FI_CHECK_PACKAGE([xpmem],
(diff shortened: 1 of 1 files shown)


raffenet:
I also think this option would be better as `--with-xpmem/--without-xpmem`. Thoughts?

j-xiong:
I think `--enable-xpmem/--disable-xpmem` is fine, as (1) it's already there; (2) we also allow directory in the `--enable` options if needed, see `config/fi_provider.m4` for example.

raffenet:
OK. Is there any other comment on this PR? It would be great if this could backport to 1.20.x and be included in the next release.

j-xiong (approved):

PR #9856, opened 2024-03-04
prov/shm: revert the smr_region fields adjustment

shijin-aws:
The earlier change caused regression for inline message size and didn't bump the SMR protocol version. Revert it before we have a better fix on the alignment and cahce coherency.

This reverts commit b069cb494dc411f1cba9b8f66292c142862da254.


shijin-aws:
@aingerson I would like to revert this change before we come up with a better fix (I know you have a branch) as the earlier commit caused regression for inline msg size

aingerson (approved):

PR #9863, opened 2024-03-07
configure: relax xpmem libdir prefix

wenduwan:
For some installations the library files are under lib/ instead of lib64/. This patch relaxes the libdir prefix to accommodate both scenarios.


j-xiong (approved):

j-xiong:
Already part of #9855

PR #9864, opened 2024-03-07
fabtests: improve ft_sendmsg, use FI_DELIVERY_COMPLETE for ft_sync

shijin-aws:
This PR contains two changes
1. Improve ft_sendmsg: Use FT_POST to handle EAGAIN, make inject_test.c and ft_finalize_ep use the new ft_sendmsg function to remove code duplication.

2. Make ft_sync do tx with FI_DELIVERY_COMPLETE: Use FI_DELIVERY_COMPLETE to have a stronger sync up between the sender and receiver, which can enforce the handshake made in the protocols of RDM endpoints.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -2826,11 +2826,9 @@ int ft_recvmsg(struct fid_ep *ep, fi_addr_t fi_addr,
 		tagged_msg.tag = ft_tag ? ft_tag : tx_seq;
 		tagged_msg.ignore = 0;
 
-		ret = fi_trecvmsg(ep, &tagged_msg, flags);
-		if (ret) {
-			FT_PRINTERR("fi_trecvmsg", ret);
-			return ret;
-		}
+		FT_POST(fi_trecvmsg, ft_progress, rxcq, rx_seq,
+			&rx_cq_cntr, "trecvmsg", ep, &tagged_msg,
+			flags);
 	} else {
 		msg.msg_iov = &msg_iov;
 		msg.desc = &mr_desc;
@@ -2839,11 +2837,9 @@ int ft_recvmsg(struct fid_ep *ep, fi_addr_t fi_addr,
 		msg.data = NO_CQ_DATA;
 		msg.context = ctx;
 
-		ret = fi_recvmsg(ep, &msg, flags);
-		if (ret) {
-			FT_PRINTERR("fi_recvmsg", ret);
-			return ret;
-		}
+		FT_POST(fi_recvmsg, ft_progress, rxcq, rx_seq,
+			&rx_cq_cntr, "recvmsg", ep, &msg,
+			flags);
 	}
 
 	return 0;
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -2768,13 +2768,12 @@ int ft_get_tx_comp(uint64_t total)
 int ft_sendmsg(struct fid_ep *ep, fi_addr_t fi_addr,
 		size_t size, void *ctx, int flags)
 {
-	int ret;
 	struct fi_msg msg;
 	struct fi_msg_tagged tagged_msg;
 	struct iovec msg_iov;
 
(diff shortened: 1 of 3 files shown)


fabtests/common/shared.c
```diff
@@ -2786,11 +2785,9 @@ int ft_sendmsg(struct fid_ep *ep, fi_addr_t fi_addr,
...
 		tagged_msg.ignore = 0;
 
-		ret = fi_tsendmsg(ep, &tagged_msg, flags);
-		if (ret) {
-			FT_PRINTERR("fi_tsendmsg", ret);
-			return ret;
-		}
+		FT_POST(fi_tsendmsg, ft_progress, txcq, tx_seq,
+			&tx_cq_cntr, "tsendmsg", ep, &tagged_msg,
+			flags);
```
j-xiong, line 2790: The old code didn't increase tx_seq. Is that considered a bug being fixed by this patch? In addition, should similar changes be made to `ft_recvmsg`?

fabtests/common/shared.c
```diff
@@ -2915,7 +2915,11 @@ int ft_sync()
 
 	if (opts.dst_addr) {
 		if (!(opts.options & FT_OPT_OOB_SYNC)) {
-			ret = ft_tx(ep, remote_fi_addr, 1, &tx_ctx);
+			ret = ft_sendmsg(ep, remote_fi_addr, 1, &tx_ctx, FI_DELIVERY_COMPLETE);
```
j-xiong, line 2918: Do all providers support delivery completion w/o first asking for it?

shijin-aws: The info will use the op_flags of hints (via fi_alter_tx_attr), so we don't know whether provider supports DC without application requesting it. I see many tests in fabtests use FI_TRANSMIT_COMPLETE and FI_INJECT_COMPLETE flags in fi_sendmsg without specifying them in the hints, i.e. provider may ignore them if it doesn't support those completion semantics. Does DC have to be special in this case? I would expect that if a provider doesn't support DC it may just ignore this flag

shijin-aws:
AWS CI  failed for shm provider for all the fabtests that involve data check `-v`. 

```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.5.132 'timeout 540 /bin/bash --login -c '"'"'/home/ubuntu/PortaFiducia/build/libraries/libfabric/pr9864/install/fabtests/bin/fi_rma_bw -e rdm -o write -I 5 -v -S r:0,1024,1048576 -D cuda -i 0 -p shm -E=9233'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.5.132 'timeout 540 /bin/bash --login -c '"'"'/home/ubuntu/PortaFiducia/build/libraries/libfabric/pr9864/install/fabtests/bin/fi_rma_bw -e rdm -o write -I 5 -v -S r:0,1024,1048576 -D cuda -i 0 -p shm -E=9233 172.31.5.132'"'"''
client_stdout:

client returncode: None
Client error: Client timed out
server_stdout:
Data check error (0!=#x00) at byte 0 for buffer size 1
0       5       0           0.00s      0.00      18.20       0.05
Data check error (0!=#x00) at byte 0 for buffer size 1
1k      5       5k          0.00s      2.64     387.20       0.00
Data check error (0!=#x00) at byte 0 for buffer size 1
Data check error (1!=0) at byte 1 for buffer size 30720
```

Looking now

fabtests/common/shared.c
shijin-aws: > The old code didn't increase tx_seq. Is that considered a bug being fixed by this patch?

I think so. Currently this function is only used by inject.c test, I didn't see it caused an issue though.


> should similar changes be made to ft_recvmsg?

Yes we should, but it seems this function is obsolete.

shijin-aws:
Fixed CI failure in newer revision

PR #9870, opened 2024-03-08
contrib/intel/jenkins: Add back missing DELETE_LOCATION

zachdworkin:
Delete location accidentally got removed and needs to be added back for proper cleanup on abort case


zachdworkin:
We don't need AWS CI to finish for Intel CI changes

PR #9872, opened 2024-03-09
core: Add runtime parameter for choosing a specific DL provider

j-xiong:
Previously, if multiple instances of a provider with the same name are found, either the one with the highest version or the one registered first is chosen, depending on if and how FI_PROVIDER_PATH is set. This works fine for most cases. However, sometimes it is preferrable to be able to choose a specific provider instance, regardless of the version and the registration order the provider discovery mechanism would have resulted in.

Here a new variable FI_PROVIDER_PRELOAD is introduced. It is specified as a colon separated list, with each item being the full path to a DL provider. Upon successfully registration, such provider takes precedence over other providers with the same name, including the built-in one.

### DIFF
--- a/include/windows/osd.h
+++ b/include/windows/osd.h
@@ -281,6 +281,9 @@ do						\
 #define be64toh ntohll
 #define strncasecmp _strnicmp
 
+#define access(path, mode) _access(path, mode)
+#define F_OK 0
+
 typedef int pid_t;
 #define getpid (int)GetCurrentProcessId
 
--- a/src/fabric.c
+++ b/src/fabric.c
@@ -65,11 +65,13 @@ struct ofi_prov {
 	struct fi_provider	*provider;
 	void			*dlhandle;
 	bool			hidden;
+	bool			preload;
 };
 
 enum ofi_prov_order {
 	OFI_PROV_ORDER_VERSION,
 	OFI_PROV_ORDER_REGISTER,
+	OFI_PROV_ORDER_PRELOAD,
 };
 
 struct ofi_info_match {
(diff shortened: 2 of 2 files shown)


shefty:
We really don't need MORE environment variables to select a provider.  How do you even document how the variables interact?  Just extend the definition of the current variables.

shefty:
I also don't see how the functionality differs much from "FI_PROVIDER_PATH=@..."

j-xiong:
`FI_PROVIDER_PATH=@...` applies to all providers. This variable selects individual provider(s). For example, `FI_PROVIDER_PRELOAD=/home/me/install/libfabric-1.18.3/lib/libfabric/libtcp-fi.so` would put that provider instance before any other tcp providers while allow other providers still select the one with highest version.

Potentially we could expand `FI_PROVIDER_PATH` to cover this use case. However, that may bring more confusion to the user than a separate variable. We have `LD_LIBRARY_PATH` vs `LD_PRELOAD` as an analogy.

j-xiong:
I have an alternative design that extends the definition of `FI_PROVIDER_PATH` instead of adding the new variable. Currently `FI_PROVIDER_PATH` is a list of directories where provider libraries can be found. The optional `@` at the beginning indicates ordering by registration instead of version. The proposed extension is to allow any list entry to be the full path to a provider library. Such an entry indicates that this instance of provider should be the first choice among all providers with the same name. To simplify the parsing logic, we could require such path start with `+`. Here is an example:
```
FI_PROVIDER_PATH=+/home/me/install/libfabric-1.18.3/lib/libfabirc/libtcp-fi.so:/usr/lib64/libfabric
``` 
@shefty What do you think about this approach?

shefty:
I think the alternative makes sense.  Would the '+' also accept a directory, as shown?

j-xiong:
@shefty No, the '+' only accepts a full path to library. See #9882 for details.

PR #9873, opened 2024-03-11
prov/rxm: Add option to auto detect hmem iface of user buffers

j-xiong:
A new option FI_OFI_RXM_DETECT_HMEM_IFACE is added. With the option turned on, RxM tries to detect the HMEM iface of user buffers if no mr_desc is supplied. This allows successful copy between user buffers located in device memory and internal bounce buffers, as well as registering such user buffer in RxM Rendezvous protocol.

Note that this doesn't allow such buffers to be used in the RxM direct send mode if the core provider requires FI_MR_HMEM because no memory registration is introduced by the option.

### DIFF
--- a/man/fi_rxm.7.md
+++ b/man/fi_rxm.7.md
@@ -175,6 +175,12 @@ with (default: 256).
   consecutively read across progress calls without checking to see if the
   CM progress interval has been reached (default: 128)
 
+*FI_OFI_RXM_DETECT_HMEM_IFACE*
+: Set this to 1 to allow automatic detection of HMEM iface of user buffers
+  when such information is not supplied. This feature allows such buffers be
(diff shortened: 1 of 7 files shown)


aingerson:
@j-xiong Isn't this pretty much what the hmem hook does?

j-xiong:
@aingerson It's a rxm specific light weight implementation targeted for middleware such as Intel MPI. It only tries to detect the hmem iface (instead of registering the memory) so that it doesn't work with the rxm direct send mode. rxm now forces rndv for FI_HMEM_ZE and FI_HMEM_SYNAPSEAI and that's all we are targeting now.

j-xiong:
Updated the man page and commit message for a more precise description.

aingerson (approved):

PR #9887, opened 2024-03-14
man: Add description of FI_PROVIDER_PATH to fabric.7.md

j-xiong:
This is a follow up to #9882

### DIFF
--- a/man/fabric.7.md
+++ b/man/fabric.7.md
@@ -251,7 +251,9 @@ Users can enable or disable available providers through build configuration
 options.  See 'configure --help' for details.  In general, a specific provider
 can be controlled using the configure option '--enable-<provider_name>'.  For
 example, '--enable-udp' (or '--enable-udp=yes') will add the udp provider to the
-build.  To disable the provider, '--enable-udp=no' can be used.
+build.  To disable the provider, '--enable-udp=no' can be used.  To build the
+provider as a stand-alone dynamically loadable library (i.e. DL provider),
+'--enable-udp=dl' can be used.
 
 Providers can also be enable or disabled at run time using the FI_PROVIDER
 environment variable.  The FI_PROVIDER variable is set to a comma separated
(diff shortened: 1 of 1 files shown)


man/fabric.7.md
```diff
@@ -251,15 +251,58 @@ Users can enable or disable available providers through build configuration
...
+	`FI_PROVIDER="udp,tcp"`
+
+By default, DL providers can be found under the `libfabric` subdirectory
+of the directory where `libfabric.so` is installed. This location is
+determined when libfabric is built and installed. Libfabric also tries to
+find DL providers following the naming convention `lib<prov_name>-fi.so`
+using the search path of 'ld.so'.
+
+The FI_PROVIDER_PATH variable can be used to change the location to search
+for DL providers and how to resolve conflict if multiple providers with the
```
shefty, line 273: conflict -> conflicts

man/fabric.7.md
```diff
@@ -251,15 +251,58 @@ Users can enable or disable available providers through build configuration
...
+If FI_PROVIDER_PATH is set, but no directory is supplied, the default
+provider directory is used. Some examples:
+
+	FI_PROVIDER_PATH=
+	FI_PROVIDER_PATH=@
+	FI_PROVIDER_PATH=+/opt/libfabric/libtcp-fi.so
+	FI_PROVIDER_PATH=@+/opt/libfabric/libtcp-fi.so
+
+When FI_PROVIDER_PATH is set, even if it is empty, no search is performed
+using the search path of 'ld.so'.
```
shefty, line 305: This statement and the one above the examples seem to contradict each other.  What is the behavior if FI_PROVIDER_PATH is set, but empty? What's the difference between empty versus = "@"?

j-xiong: I could have worded it better. Basically, there are two things that are "default":
(1) the default provider directory under <install-dir>/lib/libfabric;
(2) the search path of ld.so

When FI_PROVIDER_PATH is not set at all, both (1) and (2) are searched, but the searching (2) are only for those well-formed filennames.

When FI_PROVIDER_PATH is set, but doesn't have directory (i.e.,is empty, or only have @, or only have items starts with +), (1) is searched but not (2).

Both 'FI_PROVIDER_PATH=' and 'FI_PROVIDER_PATH=@' would cause (1) be searched, the difference is the ordering.

j-xiong:
@shefty I reworked some of the paragraphs trying to make them clearer.

shefty (approved):

j-xiong:
CI failure is unrelated.

PR #9888, opened 2024-03-14
fabtests: Utilize `junitparser` module directly

darrylabbate:
This resolves issues when `junitparser` isn't available in `PATH` (or whatever `os.system` searches)


j-xiong (approved):

j-xiong:
The CI failure with ucx is unrelated.

PR #9889, opened 2024-03-14
fabtests: Fix coverity issues 414806, 414801, 414809

zachdworkin:
414806: Fix resouce leak
414801, 414809: Replace rand() use with custom get_random_data() that does the same thing but uses /dev/urandom to get the data.


j-xiong (commented):
In addition, I would group the three random functions and the global variable together.

j-xiong (approved):

j-xiong:
Ignoring the known CI issue that has been fixed in main.

PR #9895, opened 2024-03-15
Fabtests/rdm_stress: Fix coverity issue 414806

Juee14Desai:
Coverity is flagging an issue for not closing file control ctrl_f in case of error getting file attributes. 
The fix is to close ctrl_f in this path before returning.


Juee14Desai:
PR #9889 already open to address this issue.

PR #9898, opened 2024-03-17
prov/shm: Fix a few issues related to IPC with HMEM ZE

j-xiong:
Recent CI infrastructure change enabled the IPC for HMEM ZE. Failures were observed in the "ze-shm-v3" stage. This PR contains two fixes:

* don't close dma-buf fd when req is complete
* allow send that uses IPC to complete when the message is discarded at the receive side


prov/shm/src/smr_progress.c
```diff
@@ -99,8 +99,16 @@ static int smr_progress_resp_entry(struct smr_ep *ep, struct smr_resp *resp,
 		break;
 	case smr_src_ipc:
 		assert(pending->mr[0]);
+		/*
+		 * Don't close the fd here since it's cached by level-zero.
+		 * Otherwise later reference to the buffer via the fd got
+		 * from level-zero would fail. The fd should only be closed
+		 * when the buffer is freed.
+		 */
+		/*
```
aingerson, line 108: I think we can just drop this then

prov/shm/src/smr_ep.c
```diff
@@ -1259,6 +1259,12 @@ static void smr_init_ipc_socket(struct smr_ep *ep)
 static int smr_discard(struct fi_peer_rx_entry *rx_entry)
 {
 	struct smr_cmd_ctx *cmd_ctx = rx_entry->peer_context;
+	struct smr_region *peer_smr;
+	struct smr_resp *resp;
+
+	peer_smr = smr_peer_region(cmd_ctx->ep->region, cmd_ctx->cmd.msg.hdr.id);
+	resp = smr_get_ptr(peer_smr, cmd_ctx->cmd.msg.hdr.src_data);
+	resp->status = SMR_STATUS_SUCCESS;
```
aingerson, line 1269: Good catch on this one! The response is not always valid though. In the inline and path it's not pointing to anything and the inject path it's pointing to an inject buffer which could result in a corruption somewhere. This needs to be
```
if (cmd_ctx->cmd_msg.hdr.src_data >= smr_src_iov) {
    //update resp
}
```
or something along those lines

aingerson (commented):
A couple of comments. Thanks for finding and fixing!

zachdworkin (approved):

PR #9903, opened 2024-03-18
contrib/intel/jenkins: Disable ZE IPC path and use SAR instead

zachdworkin:
The performance is really bad and takes very long. Until we debug it the ZE IPC path it will be disabled and will use SAR instead.


j-xiong:
@zachdworkin The 'oneCCL-GPU-v3' stage is also slow with ZE IPC.

j-xiong:
Wait until the 'v1.21.x' branch is created.

j-xiong:
The v1.21.x branch has been created.

j-xiong:
Doesn't seem to make any difference.

aingerson:
@j-xiong Based on the performance numbers in the log, it definitely did. It just doesn't look like it because it spent a long time waiting for nodes from other PRs that are still having the issue.

PR #9904, opened 2024-03-18
Intel/CI: Add OneCCL CPU tests with DSA + shm to the CI

Juee14Desai:
- Add OneCCL DSA stage.
- There was an issue with libfabric being built with ze and oneccl built without.
This led to failure and CCL transport being switched to MPI because oneccl looked for the library.
To avoid this, an else case is added with --with-ze=no.
- Update oneccl summarizer to support DSA + shm summary.
- Add the logic to check whether DSA was actually used or not.

### DIFF
--- a/contrib/intel/jenkins/summary.py
+++ b/contrib/intel/jenkins/summary.py
@@ -395,12 +395,7 @@ class OnecclSummarizer(Summarizer):
         self.file_path = os.path.join(self.log_dir, self.file_name)
         self.exists = os.path.exists(self.file_path)
         self.name = 'no_test'
-
-    def read_file(self):
-        with open(self.file_path, 'r') as log_file:
-            self.fast_forward(log_file)
-            for line in log_file:
-                self.check_line(line)
+        self.trace = False
 
     def check_name(self, line):
         #OneCCL GPU tests:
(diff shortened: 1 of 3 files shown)


contrib/intel/jenkins/summary.py
```diff
@@ -430,6 +431,23 @@ def check_fail(self, line):
             self.fails += 1
             self.failed_tests.append(self.name)
 
+    def check_trace(self, line):
```
zachdworkin, line 434: I don't think this is getting called in the OneCCL Summarizer. You probably need to overload the summarize() and check_line() functions from the superclass just like in the Fabtests Summarizer. This will also move your final check in this function to the summarize() function like in fabtests so the print only happens once.

zachdworkin (changes requested):

j-xiong:
Wait until the 'v1.21.x' branch is created.

j-xiong:
The v1.21.x branch has been created.

zachdworkin (approved):
looks good, we still need to wait for cloudbees to pass before merging

zachdworkin (approved):

zachdworkin:
AWS CI does not need to finish since these changes are only to Intel CI

PR #9908, opened 2024-03-19
prov/shm: manually align 8 byte fields in memory region

aingerson:
Move all of the 8 byte fields together to eliminate extra padding. Previously, the last 8 byte field was after a 4 byte field, forcing the compiler to add extra padding before the 4 byte field and after the third 8 byte field. This increased the size of the memory region and pushed some fields to a new cache line. Consolidating them eliminates those extra bytes and allows for better caching performance.
This also adds a reserved value to force alignment of the next 32 bit field though the compiler should automatically add that memory anyway.

Altering the order of the fields in the shared memory region requires a version increment.

### DIFF
--- a/prov/shm/src/smr_util.h
+++ b/prov/shm/src/smr_util.h
@@ -54,7 +54,7 @@
 extern "C" {
 #endif
 
-#define SMR_VERSION	6
+#define SMR_VERSION	7
 
 #define SMR_FLAG_ATOMIC	(1 << 0)
 #define SMR_FLAG_DEBUG	(1 << 1)
@@ -225,8 +225,10 @@ struct smr_region {
 	int		pid;
 	uint8_t		cma_cap_peer;
 	uint8_t		cma_cap_self;
-	uint32_t	max_sar_buf_per_peer;
 	uint8_t		xpmem_cap_self;
+	uint8_t		resv2;
+
+	uint32_t	max_sar_buf_per_peer;
 	struct xpmem_pinfo xpmem_self;
 	struct xpmem_pinfo xpmem_peer;
 	void		*base_addr;


shijin-aws (approved):

zachdworkin (approved):

prov/shm/src/smr_util.h
```diff
@@ -225,8 +225,10 @@ struct smr_region {
 	int		pid;
 	uint8_t		cma_cap_peer;
 	uint8_t		cma_cap_self;
-	uint32_t	max_sar_buf_per_peer;
 	uint8_t		xpmem_cap_self;
+	uint8_t		resv2;
+
+	uint32_t	max_sar_buf_per_peer;
```
darrylabbate, line 231: Why not use `_Alignas()`?

```suggestion
	_Alignas(8) uint32_t	max_sar_buf_per_peer;
```

aingerson: Mostly because I didn't know about it :)
Can you elaborate on what that would do?
I personally like the explicit value because it's easier to see that extra byte but I'm open to suggestions if anyone feels strongly about it.

j-xiong:
@aingerson Your feature branch needs a rebase to pick up the shm fixes. It's 27 commits behind main.

prov/shm/src/smr_util.h
shijin-aws: @darrylabbate are you still pushing it?

darrylabbate: I guess it depends on the intent. If you want to reserve the byte `resv2` for future use, the explicit field makes more sense. Otherwise, [`_Alignas()`](https://en.cppreference.com/w/c/language/_Alignas) will set the alignment requirement of `max_sar_buf_per_peer` however you wish, without the hardcoded `uint8_t` fields littering the struct definition.

This is mainly useful since you don't need to worry about how various systems/architectures handle data type alignment (C standard makes few guarantees), and future devs don't need to guess what the random `uint8_t`s are for.

darrylabbate: > @darrylabbate are you still pushing it?

No this is not a blocker

PR #9909, opened 2024-03-19
prov/efa: Propagate error codes from core Libfabric functionality untouched

darrylabbate:
This is simply a best-effort attempt at removing some of the overridden error codes that are returned from core Libfabric functions. There are also a few random commits for minor code touchups.

Special attention appreciated for signedness - it's an easy thing to goof up.

### DIFF
--- a/prov/efa/src/dgram/efa_dgram_ep.c
+++ b/prov/efa/src/dgram/efa_dgram_ep.c
@@ -303,7 +303,7 @@ static void efa_dgram_ep_progress_internal(struct efa_dgram_ep *ep, struct efa_d
 	if (OFI_UNLIKELY(ret < 0)) {
 		if (OFI_UNLIKELY(ret != -FI_EAVAIL)) {
 			EFA_WARN(FI_LOG_CQ, "no error available errno: %ld\n", ret);
-			efa_base_ep_write_eq_error(&ep->base_ep, FI_EIO, FI_EFA_ERR_DGRAM_CQ_READ);
+			efa_base_ep_write_eq_error(&ep->base_ep, -ret, FI_EFA_ERR_DGRAM_CQ_READ);
 			return;
 		}
 
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -811,27 +811,24 @@ static void efa_av_close_reverse_av(struct efa_av *av)
 static int efa_av_close(struct fid *fid)
 {
 	struct efa_av *av;
-	int ret = 0;
 	int err = 0;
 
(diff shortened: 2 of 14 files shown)


shijin-aws (approved):

shijin-aws:
The propagation of rdma-core error code to cq error will be in a follow-up PR, right?

shijin-aws:
AWS CI failure looks real

darrylabbate:
> The propagation of rdma-core error code to cq error will be in a follow-up PR, right?

This PR tries to fix cases where we blatantly throw away core Libfabric error codes in favor of `FI_EIO` or similar. We already propagate CQ error codes we receive from RDMA Core mostly untouched via `fi_eq_err_entry.prov_errno` (`EFA_IO_COMP_STATUS_*`). It'd feel redundant to map RDMA Core error codes to core Libfabric error codes for `fi_eq_err_entry.err` as well.

shijin-aws:
But prov_errno is provider specific, they want some refined err_entry.err which is universal for all providers

darrylabbate:
> AWS CI failure looks real

Failure looks to be due to this change propagating `-FI_ENOCQ` via `ibv_create_cq_ex()` failures in `efa_cq_ibv_cq_ex_open_with_ibv_create_cq_ex()`. The `test_cq` fabtests will only tolerate `-FI_EINVAL`. Either the `-FI_ENOCQ` needs to be changed or the previous hardcoded `-FI_EINVAL` needs to be reinstated.

shijin-aws:
We may need to be careful with the propagation because some API call has expected return codes for error. For example, fi_cq_read can only accept -FI_EAVAIL and non-negative integers as return values.

darrylabbate:
> But prov_errno is provider specific, they want some refined err_entry.err which is universal for all providers

Yeah we should make this a separate change.

darrylabbate:
> We may need to be careful with the propagation because some API call has expected return codes for error. For example, fi_cq_read can only accept -FI_EAVAIL and non-negative integers as return values.

For the failure I referenced above, I believe it only applies to `fi_cq_open()`. The man seems to be pretty permissive in this regard:

> Returns 0 on success. On error, returns a negative fabric errno.

But the fabtest isn't too happy with `-FI_ENOCQ`. Could be that `FI_ENOCQ` was added later.

shijin-aws:
FI_ENOCQ is usually used to report the error that an endpoint was not binding a CQ before being enabled. I think fi_cq_open shouldn't return that error code. We should fix that in efa provider

shijin-aws:
FWIW: https://github.com/ofiwg/libfabric/pull/9914

darrylabbate:
Build checks failing due to Ubuntu PPA hullabaloo (#9915)

PR #9913, opened 2024-03-20
prov/efa: Make the inflight read msg per domain

shijin-aws:
Make the inflight read msg counter per domain rather than per peer.

This counter is used to prevent using runting read when EFA is busy with a read, since in that case runting read would be less performant than a read.

Since any ongoing read, regardless of peer, makes EFA busy the counter should be domain scoped and not peer scoped.

cherry-picked from @ygoldamzn 's PR in https://github.com/aws/libfabric/pull/13

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke_rtm.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_rtm.c
@@ -1172,11 +1172,7 @@ ssize_t efa_rdm_pke_init_longread_tagrtm(struct efa_rdm_pke *pkt_entry,
  */
 void efa_rdm_pke_handle_longread_rtm_sent(struct efa_rdm_pke *pkt_entry)
 {
-	struct efa_rdm_peer *peer;
-
-	peer = efa_rdm_ep_get_peer(pkt_entry->ep, pkt_entry->addr);
-	assert(peer);
-	peer->num_read_msg_in_flight += 1;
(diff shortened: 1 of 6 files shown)


prov/efa/src/rdm/efa_rdm_pke_rtm.c
```diff
@@ -1172,11 +1172,11 @@ ssize_t efa_rdm_pke_init_longread_tagrtm(struct efa_rdm_pke *pkt_entry,
...
 void efa_rdm_pke_handle_longread_rtm_sent(struct efa_rdm_pke *pkt_entry)
 {
-	struct efa_rdm_peer *peer;
+	struct efa_domain *domain;
 
-	peer = efa_rdm_ep_get_peer(pkt_entry->ep, pkt_entry->addr);
-	assert(peer);
-	peer->num_read_msg_in_flight += 1;
+	domain = efa_rdm_ep_domain(pkt_entry->ep);
+	assert(domain);
```
sunkuamzn, line 1178: nit: I think the `assert(domain)` are not necessary because we cannot have endpoint without a domain

sunkuamzn (approved):

shijin-aws:
@ygoldamzn

prov/efa/src/rdm/efa_rdm_pke_rtm.c
shijin-aws: Done

PR #9915, opened 2024-03-20
.github/workflows: Use latest Ubuntu

darrylabbate:
I couldn't find any context why Ubuntu 20.04 would be explicitly required for our Actions workflows. Feel free to discard otherwise.

This simplifies the testing matrix and allows us to test with the minimum required LTTng version without needing to deal with the flaky PPA repo for Ubuntu 20.04.

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -23,8 +23,6 @@ env:
     sparse
     valgrind
     wget
-  APT_REPOS: >-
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

j-xiong (approved):

ooststep (approved):
thanks!

shijin-aws (approved):

PR #9927, opened 2024-03-21
prov/sockets: Fix coveritylocking issues

zachdworkin:


shefty:
It's more useful if the commit header describe the problem/fix rather than a reference to a coverity report

prov/sockets/src/sock_wait.c
```diff
@@ -137,8 +137,11 @@ static int sock_wait_wait(struct fid_wait *wait_fid, int timeout)
 			cq = container_of(list_item->fid,
 					  struct sock_cq, cq_fid);
 			sock_cq_progress(cq);
-			if (ofi_rbused(&cq->cqerr_rb))
+			pthread_mutex_lock(&cq->lock);
+			if (ofi_rbused(&cq->cqerr_rb)) {
+				pthread_mutex_unlock(&cq->lock);
 				return 1;
+			}
```
sydidelot, line 144: There is a `pthread_mutex_lock(&cq->lock);` missing here, before the `break`.

sydidelot (commented):
I think the PR has a bug, please check my comment.

prov/sockets/src/sock_wait.c
zachdworkin: fixed in latest push, thanks!

j-xiong:
bot:aws:retest

j-xiong (approved):

PR #9928, opened 2024-03-21
prov/coll: Fix Coverity issue about race condition accessing av->av_set

j-xiong:
CID 407572: Data race condition (MISSING_LOCK)

Accessing av_set->av->av_set without holding lock util_av.lock. Elsewhere, util_av.av_set is written to with util_av.lock held 1 out of 1 times (prov/util/src/util_av.c: ofi_av_close()).


shefty:
nit: It's more useful if the commit header described the problem/fix, rather than referring the reader to go lookup a coverity report

j-xiong:
bot:aws:retest

shijin-aws:
earlier AWS CI issue should be fixed.

PR #9942, opened 2024-03-26
prov/efa: Remove unconditional assignment of err_entry

shijin-aws:
Currently, efa_rdm_cq_poll_ibv_cq always assigns err_entry as {0} at the beginning while err_entry is only used in the error path. This unconditional assignment (88 bytes) caused additonal overhead in the fast path. This patch fixes it by moving the assignment of err_entry to the error path only


darrylabbate (approved):

shijin-aws:
AWS CI failure is unrelated

PR #9967, opened 2024-04-02
prov/efa: Make num_runt_bytes_in_flight per domain

shijin-aws:
This counter is used to prevent using runting read when the outstanding send bytes is reaching a limit that makes EFA NIC busy. In that case runting read would be less performant than a read.


shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

shijin-aws:
Will revisit later

PR #9973, opened 2024-04-02
prov/efa: Fix error handling in efa_rdm_cq_poll_ibv_cq

shijin-aws:
When ibv_start_poll or ibv_next_poll return error, it may read a cqe from a destroyed EP (QP). In this case we shouldn't call efa_rdm_cq_get_prov_errno, which may use a freed EP from the pkt_entry (wr_id) and cause segmentation fault.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -381,7 +381,7 @@ void efa_rdm_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 
 	if (err && err != ENOENT) {
 		err = err > 0 ? err : -err;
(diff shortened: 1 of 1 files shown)


darrylabbate:
Is it possible to capture this in a condition inside `efa_rdm_cq_get_prov_errno()` instead?

shijin-aws:
>Is it possible to capture this in a condition inside efa_rdm_cq_get_prov_errno() instead?

Unless we want to make the return code (say `err`) of `ibv_start_poll` as an input for `efa_rdm_cq_get_prov_errno()`. When `err` is `EINVAL`, that likely means the polled pkt entry is from a destroyed QP.

There is no good way to tell that from pkt_entry (wr_id) itself.

So I prefer to just call `ibv_wc_read_vendor_err` directly to avoid these troubles

shijin-aws:
bot:aws:retest

darrylabbate (approved):

PR #9982, opened 2024-04-05
contrib/aws: Add Windows Test

a-szegel:
Adds windows test to new AWS CI

### DIFF
--- a/contrib/aws/common.groovy
+++ b/contrib/aws/common.groovy
@@ -97,7 +97,12 @@ def get_single_node_windows_test_stage(stage_name) {
     return {
         stage("${stage_name}") {
             def ret = sh (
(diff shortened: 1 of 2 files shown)


contrib/aws/common.groovy
```diff
@@ -97,7 +97,12 @@ def get_single_node_windows_test_stage(stage_name) {
     return {
         stage("${stage_name}") {
             def ret = sh (
-                            script: ". venv/bin/activate; cd PortaFiducia/scripts; env AWS_DEFAULT_REGION=us-west-2 ./test_orchestrator_windows.py public",
+                            script: """
+                                . venv/bin/activate;
+                                cd PortaFiducia/scripts;
+                                export PULL_REQUEST_ID=${env.CHANGE_ID};
```
a-szegel, line 103: The reason we need both of these is b/c `test_orchestrator_windows.py` currently uses the env var as a default value (which means it must exist).

a-szegel:
The windows test passed the pipeline.

darrylabbate (approved):

PR #9988, opened 2024-04-09
prov/efa: Set msg_order based on hints

shijin-aws:
Efa device doesn't have ordering (EFA_MSG_ORDER == FI_ORDER_NONE)
    If apps request an ordering that is relaxed than what provider supports, we should respect that.
    This is specially true for FI_ORDER_NONE:
    No ordering is specified. This value may be used as input in order to obtain
    the default message order supported by the provider.

Also added unit tests.


prov/efa/src/efa_user_info.c
```diff
@@ -408,13 +408,32 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
...
 	if (hints) {
 		if (hints->tx_attr) {
+			/* efa device doesn't have ordering (EFA_MSG_ORDER == FI_ORDER_NONE).
+			 * if apps request an ordering that is relaxed than
+			 * what provider supports, we should respect that.
+			 * This is specially true for FI_ORDER_NONE:
+			 * No ordering is specified. This value may be used as input in order to obtain
+			 * the default message order supported by the provider.
+			 */
+			info->tx_attr->msg_order = hints->tx_attr->msg_order;
```
darrylabbate, line 418: I think it would make sense to mask `info->[tr]x_attr->msg_order` with `hints->[tr]x_attr->msg_order` instead to guard against unsupported message orderings requested by the application. This would effectively fall-back to `FI_ORDER_NONE`.

```suggestion
			info->tx_attr->msg_order &= hints->tx_attr->msg_order;
```

shijin-aws: >guard against unsupported message orderings requested by the application

This cannot happen because there was a `ofi_check_tx_attr` earlier.  It may not hurt to do that though

shijin-aws: Updated

darrylabbate (approved):

shijin-aws:
AWS CI failure is real, will fix

shijin-aws:
Need to rebase on https://github.com/ofiwg/libfabric/pull/9989

PR #9990, opened 2024-04-12
prov/efa: Add logic for non-contiguous IOV entries in prefix mode

darrylabbate:


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -44,25 +46,43 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
...
+
+	/*
+	 * In prefix mode with a single IOV, assume the prefix space is included
+	 * in the IOV buffer
+	 */
+	if (ep->user_info->mode & FI_MSG_PREFIX && msg->iov_count == 1) {
+		assert(txe->iov->iov_len >= ep->msg_prefix_size);
+		txe->iov->iov_base = (char *) txe->iov->iov_base + ep->msg_prefix_size;
+		txe->iov->iov_len -= ep->msg_prefix_size;
+	}
```
rajachan, line 70: With proper iovec walking logic, you shouldn't need to special-case on iov_count to decide where the data buffer starts. Needing to do that here tells me the rest of the iovec walking logic isn't written properly.

shijin-aws: You can use ofi_iov_locate with an offset of the prefix size to determine the real data's iov index

shijin-aws: Actually, this can be even better because it covers the iov count/len decrement as well: https://github.com/ofiwg/libfabric/blob/main/src/iov.c#L101-L104

darrylabbate: From a code quality perspective, I agree with using `ofi_consume_iov()` or `ofi_iov_locate()` - I clearly didn't check for any useful utility functions like these. But it should be noted that both functions also employ special logic when `iov_count` is 1. Is there more that needs to be dug into or are we happy with the basic logic of walking the iovec for `FI_MSG_PREFIX`?

shijin-aws: >Is there more that needs to be dug into or are we happy with the basic logic of walking the iovec for FI_MSG_PREFIX?

My vote is to just reuse them to walk the iov for FI_MSG_PREFIX case. Getting rid of 1 iov in those util functions can be a follow-up work

rajachan: Yeah, I was just pointing out that the prefix handling shouldn't need it... but I can see why the util has to work with iov_count.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -44,10 +44,10 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 	txe->efa_outstanding_tx_ops = 0;
 	dlist_init(&txe->queued_pkts);
 
-	memcpy(txe->iov, msg->msg_iov, sizeof(struct iovec) * msg->iov_count);
-	memset(txe->mr, 0, sizeof(*txe->mr) * msg->iov_count);
+	memcpy(txe->iov, msg->msg_iov, sizeof txe->iov * txe->iov_count);
```
shijin-aws, line 47: Should it be `sizeof txe->iov[0]`?

shijin-aws (approved):

PR #9999, opened 2024-04-17
prov/efa: Improve various `iov_count` assertions

darrylabbate:


prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -211,7 +211,7 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
 	struct efa_mr *mr;
 	int err;
 
-	assert(rxe->iov_count == 1);
+	assert(rxe->iov_count > 0 && rxe->iov_count <= ep->rx_iov_limit);
```
shijin-aws, line 214: I think the current logic in this function relies on iov_count == 1. If we want to touch this, we need to fix the other logic in this function as well

shijin-aws: better add a TODO here and fix it in a follow-up PR that fix the whole function

darrylabbate: So remove the `assert()` altogether?

shijin-aws: I suggest adding a TODO and leave this `assert()` in this function unchanged for now

shijin-aws (approved):

darrylabbate:
@shijin-aws So we need to fix `efa_rdm_ep_post_user_recv_buf()` itself s.t. the packet entry points to the correct IOV offset/entry? Similar to the TX entry construct fix in #9990? I didn't realize the core logic was broken.

shijin-aws:
> @shijin-aws So we need to fix `efa_rdm_ep_post_user_recv_buf()` itself s.t. the packet entry points to the correct IOV offset/entry? Similar to the TX entry construct fix in #9990? I didn't realize the core logic was broken.

Correct, the logic is currently broken if application posts recv with 2 iov. The efa_rdm_pke_recvv assumes there is 1 iov and only fill sge_list[0]. I don't think you can consume iov for rxe similar to what we did for txe in this case. You need to separate the hdr and payload iov and set them as the sge_list[0]. sge_list[1] for ibv_post_recv

PR #10000, opened 2024-04-18
prov/efa: A series of fixes on the zero-copy receive code path

darrylabbate:


rajachan:
We need "I was here for the libfabric 10000th Pull Request" merch.

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -238,11 +236,29 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
+			EFA_WARN(FI_LOG_CQ, "ofi_iov_locate failure: %s (%d)\n", fi_strerror(-err), -err);
+			return err;
+		}
+
+		assert(rx_iov_index < rxe->iov_count);
+		assert(rx_iov_offset < rxe->iov[rx_iov_index].iov_len);
+
+		pkt_entry->payload = (char *) rxe->iov[rx_iov_index].iov_base + rx_iov_offset;
+		pkt_entry->payload_mr = rxe->desc[rx_iov_index];
+		ofi_consume_iov(rxe->iov, &rxe->iov_count, ep->msg_prefix_size);
```
shijin-aws, line 255: I realize we can use ofi_consume_iov_desc https://github.com/ofiwg/libfabric/blob/main/src/iov.c#L63  so we can shift iov and desc together. Then we don't have to use ofi_iov_locate to get the rx_iov_index and rx_iov_offset

shijin-aws: Accordingly, for this line we also need to use ofi_consume_iov_desc https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ope.c#L60-L62

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -231,18 +228,28 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
 	/*
 	 * The actual receiving buffer size (pkt_size) is
 	 *    rxe->total_len - sizeof(struct efa_rdm_pke)
 	 * because the first part of user buffer was used to
 	 * construct pkt_entry. The actual receiving buffer
 	 * posted to device starts from pkt_entry->wiredata.
 	 */
-	pkt_entry->pkt_size = rxe->iov[0].iov_len - sizeof(struct efa_rdm_pke);
-
+	pkt_entry->pkt_size = rxe->total_len - sizeof *pkt_entry;
```
shijin-aws, line 239: During test I find `rxe->total_len` is not initialized before this function is called ....let's use `ofi_total_iov_len(rxe->iov, rxe->iov_count)`

darrylabbate: Shouldn't we initialize it instead?

shijin-aws: I remember it is expected to initialize that field during the progress engine. Because the total len aims to reflect the total length of data that efa received, it may be small than the recv buffer posted.

darrylabbate: Actually, I think we can just move the `FI_MSG_PREFIX` logic above this line. At that point, `rxe->total_len` would be set. WDYT?

darrylabbate: I moved the code block up. Let me know if you think it's insufficient.

shijin-aws: I think we just shouldn't update rxe->total_len here. It should be set as the received data size in the progress engine.

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -231,18 +228,27 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
 	 * posted to device starts from pkt_entry->wiredata.
 	 */
-	pkt_entry->pkt_size = rxe->iov[0].iov_len - sizeof(struct efa_rdm_pke);
-
+	pkt_entry->pkt_size = ofi_total_iov_len(rxe->iov, rxe->iov_count) - sizeof *pkt_entry;
 	pkt_entry->ope = rxe;
 	rxe->state = EFA_RDM_RXE_MATCHED;
 
+	if (ep->user_info->mode & FI_MSG_PREFIX) {
+		ofi_consume_iov_desc(rxe->iov, rxe->desc, &rxe->iov_count, ep->msg_prefix_size);
```
shijin-aws, line 244: I missed another potential bug. It seems efa_rdm_ep_post_user_recv_buf can be called by the receiver multiple times . To that end I think we shouldn't modify the original rxe->iov and rxe->desc. We need to  make a copy https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke_cmd.c#L969

darrylabbate: So, copies of `iov`, `desc` and `iov_count`?

shijin-aws: yep

shijin-aws:
I tested this PR with Amedeo's branch and find more issues. One major issue is the on the sender path it doesn't use the `(prefix_size - sizeof(struct efa_rdm_pke))` as the length of the first sgl. Instead it uses the eager msg header which doesn't have cq data hdr which is counted in the prefix.... For 2 iov send/recv, it will cause wrong iov target delivery when receiver posted two iovs  that first iov has size `(prefix_size - sizeof(struct efa_rdm_pke))`. To fix that, we need to make the sender always use the constant hdr length that include cq data hdr when zcpy is turned on

shijin-aws:
> I tested this PR with Amedeo's branch and find more issues. One major issue is the on the sender path it doesn't use the `(prefix_size - sizeof(struct efa_rdm_pke))` as the length of the first sgl. Instead it uses the eager msg header which doesn't have cq data hdr which is counted in the prefix.... For 2 iov send/recv, it will cause wrong iov target delivery when receiver posted two iovs that first iov has size `(prefix_size - sizeof(struct efa_rdm_pke))`. To fix that, we need to make the sender always use the constant hdr length that include cq data hdr when zcpy is turned on

It can be fixed by this patch https://github.com/shijin-aws/libfabric/commit/5429485d4247e56665cac9ec07bcd7a11970d892

prov/efa/src/rdm/efa_rdm_util.h
```diff
@@ -8,7 +8,11 @@
 #include "efa_rdm_protocol.h"
 #include "efa_rdm_pke.h"
 
-#define EFA_RDM_MSG_PREFIX_SIZE (sizeof(struct efa_rdm_pke) + sizeof(struct efa_rdm_eager_msgrtm_hdr) + EFA_RDM_REQ_OPT_RAW_ADDR_HDR_SIZE)
+#define EFA_RDM_MSG_PREFIX_SIZE (			\
```
shijin-aws, line 11: We need to update this protocol doc https://github.com/ofiwg/libfabric/blob/main/prov/efa/docs/efa_rdm_protocol_v4.md#43-keep-packet-header-length-constant-constant-header-length-and-zero-copy-receive for this.

shijin-aws: I am slightly concerned whether this will break backward compatibility. But I think it's fine as long as the sender & receiver use the same prefix size. Using a smaller prefix size on receiver will break, but we should be able to have a way to prevent this from happening.

shijin-aws: I mean, this is NOT a protocol breaking change.

darrylabbate:
Note that the latest pushed commits do not address all pending comments

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -58,7 +58,7 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 	txe->cq_entry.buf = OFI_LIKELY(txe->cq_entry.len > 0) ? txe->iov[0].iov_base : NULL;
 
 	if (ep->user_info->mode & FI_MSG_PREFIX) {
-		ofi_consume_iov(txe->iov, &txe->iov_count, ep->msg_prefix_size);
+		ofi_consume_iov_desc(txe->iov, txe->desc, &txe->iov_count, ep->msg_prefix_size);
```
shijin-aws, line 61: A further improvement on this should be constructing the pkt entry just from txe->iov[0].iov_base. Then we don't need to walk the iov here. Can be a follow-up PR

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -231,18 +228,27 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
 	pkt_entry->alloc_type = EFA_RDM_PKE_FROM_USER_BUFFER;
 	pkt_entry->flags = EFA_RDM_PKE_IN_USE;
 	pkt_entry->next = NULL;
+	pkt_entry->ep = ep;
 	/*
 	 * The actual receiving buffer size (pkt_size) is
 	 *    rxe->total_len - sizeof(struct efa_rdm_pke)
```
shijin-aws, line 234: Update this comment as we are not using `rxe->total_len`

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -231,18 +229,33 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
+			EFA_WARN(FI_LOG_CQ, "ofi_iov_locate failure: %s (%d)\n", fi_strerror(-err), -err);
+			return err;
+		}
+
+		assert(rx_iov_index < rxe->iov_count);
+		assert(rx_iov_offset < rxe->iov[rx_iov_index].iov_len);
+
+		pkt_entry->payload = (char *) rxe->iov[rx_iov_index].iov_base + rx_iov_offset;
+		pkt_entry->payload_mr = rxe->desc[rx_iov_index];
+		pkt_entry->payload_size = ofi_total_iov_len(rxe->iov, rxe->iov_count) - rx_iov_offset;
```
shijin-aws, line 256: This can fail for the case that rx_iov_index is > 0, in this case you need to use the total iov len starting from rx_iov_index - the rx_iov_offset

shijin-aws, line 256: This can fail for the case that rx_iov_index is > 0, in this case you need to use the total iov len starting from rx_iov_index - the rx_iov_offset

shijin-aws (changes requested):

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -231,18 +229,33 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
 	 * construct pkt_entry. The actual receiving buffer
 	 * posted to device starts from pkt_entry->wiredata.
 	 */
-	pkt_entry->pkt_size = rxe->iov[0].iov_len - sizeof(struct efa_rdm_pke);
-
+	pkt_entry->pkt_size = ofi_total_iov_len(rxe->iov, rxe->iov_count) - sizeof *pkt_entry;
 	pkt_entry->ope = rxe;
 	rxe->state = EFA_RDM_RXE_MATCHED;
 
+	if (ep->user_info->mode & FI_MSG_PREFIX) {
```
shijin-aws, line 244: For prefix mode, we only need to walk the iov. The `pkt_entry->payload` (and below) assignments should be general for non-prefix mode as well

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -209,12 +209,10 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
 	struct efa_mr *mr;
-	int err;
+	size_t rx_iov_offset;
+	int rx_iov_index, err;
 
-	/*
-	 * TODO remove/change assert expression when function logic is fixed
-	 */
-	assert(rxe->iov_count == 1);
+	assert(rxe->iov_count > 0  && rxe->iov_count <= ep->rx_iov_limit);
```
shijin-aws, line 215: This still doesn't make us safe if application tries to post a recv with iov count 4 (the current rx_iov_limit). I think a better fix should be reducing the rx iov limit to 2 when FI_MSG_PREFIX is used.

shijin-aws: Ignore this comment, I don't think we can naively require 2 iov limit for FI_MSG_PREFIX as user may not want to do zcpy recv.

prov/efa/src/rdm/efa_rdm_ep_utils.c
darrylabbate: Maybe, but you at least avoid calling `ofi_total_iov_len()` (linear time) for the non-prefix case

shijin-aws (approved):

PR #10003, opened 2024-04-23
prov/efa: Do not assert tx and rx msg order must equal

shijin-aws:
The tx_attr at the sender should align with the rx_attr at the receiver. But that's an application level responsibility to handle. The tx_attr and rx_attr of a single endpoint are likely to be consistent, but that's not required. Efa doesn't have to generate an error when an inconsitency is detected, but can generate a warning for it.


prov/efa/src/efa_user_info.c
```diff
@@ -434,6 +434,11 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
 			info->rx_attr->msg_order &= hints->rx_attr->msg_order;
 		}
 
+		if (info->tx_attr->msg_order != info->rx_attr->msg_order)
+			EFA_INFO(FI_LOG_EP_CTRL, "Inconsistent tx/rx msg order. Tx msg order: %lu, Rx msg order: %lu. "
+						 "Libfabric can proceed but it is recommended to align the tx and rx msg order. \n",
```
darrylabbate, line 439: Trim whitespace

```suggestion
						 "Libfabric can proceed but it is recommended to align the tx and rx msg order.\n",
```

prov/efa/src/rdm/efa_rdm_ep.h
```diff
@@ -244,7 +241,7 @@ static inline size_t efa_rdm_ep_get_tx_pool_size(struct efa_rdm_ep *ep)
 
 static inline int efa_rdm_ep_need_sas(struct efa_rdm_ep *ep)
 {
-	return ep->msg_order & FI_ORDER_SAS;
+	return ep->user_info->tx_attr->msg_order & FI_ORDER_SAS;
```
shijin-aws, line 244: probably we want to use `(ep->user_info->tx_attr->msg_order & FI_ORDER_SAS) || (ep->user_info->rx_attr->msg_order & FI_ORDER_SAS)`

shijin-aws: But AFAICT some applications like MPICH only set SAS for tx_attr.

darrylabbate (approved):

prov/efa/src/efa_user_info.c
shijin-aws: Done

prov/efa/src/rdm/efa_rdm_ep.h
shijin-aws: Updated

PR #10011, opened 2024-04-30
prov/efa: Correct the order of arguments in declaration

shijin-aws:
The order of payload_offset and segment_offset is flipped in the declaration of efa_rdm_pke_init_payload_from_ope. It didn't introduce trouble today because these two arguments have the same data type.


darrylabbate (approved):

shijin-aws:
bot:aws:retest

PR #10020, opened 2024-05-03
prov/tcp: Add fi_set_val support for active/passive port range

chien-intel:
There is a need (e.g. firewall rule) to limit TCP ports used for active connections.  Add FI_TCP_EGRESS_PORT_LOW_RANGE and FI_TCP_EGRESS_PORT_HIGH_RANGE environment variables to specify the port range for active connections.  If a port is specified for the active connection, it will be used over the specified range. This is the same behavior as passive connection.

I don't understand the check !ofi_is_any_addr(info->src_addr) in the original code.  It seems there is a possibility of INADDR_ANY with a port number and continue to call bind.  Anyhow, this is my best interpretation.

### DIFF
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -81,6 +81,7 @@ enum {
 enum {
 	/* null terminated string, <low port>-<high port> */
 	FI_TCP_DOMAIN_ACTIVE_PORT_RANGE = -FI_PROV_SPECIFIC_TCP,
+	FI_TCP_FABRIC_PASSIVE_PORT_RANGE,
 };
 
 struct fi_fid_export {
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -78,6 +78,11 @@ enum {
 	FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 };
 
+enum {
+	/* null terminated string, <low port>-<high port> */
+	FI_TCP_DOMAIN_ACTIVE_PORT_RANGE = -FI_PROV_SPECIFIC_TCP,
+};
+
 struct fi_fid_export {
 	struct fid **fid;
 	uint64_t flags;
--- a/man/fi_tcp.7.md
+++ b/man/fi_tcp.7.md
@@ -90,6 +90,17 @@ show all environment variables defined for the tcp provider.
   through the standard socket APIs (i.e. connect, accept, send, recv).
   Default: disabled.
 
+The tcp provider also supports domain parameter via fi_set_val.
+
+*FI_TCP_DOMAIN_ACTIVE_PORT_RANGE*
+: Limit active connections to a TCP port range for a domain.  Value is
+  struct xnet_port_range *.  To disable, set hight and low to 0.
+  If an active connection's source address has a port number, it will
+  be used instead of the port range.
+  Sample:
+    struct xnet_port_range range = { .high = 5000, .low = 6000 };
+    fi_set_val(&domain->fid, FI_TCP_DOMAIN_ACTIVE_PORT_RANGE, &range);
+
 # NOTES
 
 The tcp provider supports both msg and rdm endpoints directly.  Support
--- a/man/fi_tcp.7.md
+++ b/man/fi_tcp.7.md
@@ -90,7 +90,14 @@ show all environment variables defined for the tcp provider.
   through the standard socket APIs (i.e. connect, accept, send, recv).
   Default: disabled.
 
-The tcp provider also supports domain parameter via fi_set_val.
+The tcp provider also supports setting parameters via fi_set_val.
+
+*FI_TCP_FABRIC_PASSIVE_PORT_RANGE*
+: Set values corresponding to FI_TCP_PORT_LOW_RANGE and FI_TCP_PORT_HIGH_RANGE.
+  Use fabric object with struct xnet_port_range * set with desired values.
+  Sample:
+    struct xnet_port_range range = { .low = 9000, .high = 9100 };
+    fi_set_val(&fabric->fid, FI_TCP_FABRIC_PASSIVE_PORT_RANGE, &range);
 
 *FI_TCP_DOMAIN_ACTIVE_PORT_RANGE*
 : Limit active connections to a TCP port range for a domain.  Value is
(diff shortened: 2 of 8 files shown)


j-xiong:
It is valid to call bind with a port number over INADDR_ANY.

Also could you update fi_tcp.7.md accordingly?

shefty:
The existing high/low port range applies to listens and accepted sockets.  The proposed new range applies to sockets passed to connect.  Do we need separate port ranges for sockets configured using accept() versus connect()?  A single range makes life easier on the user.

The original code is performing this check:

If we have a source IP address or source port (or both), then the socket must be bound to that address.  The expected case is to have a source IP address only (wildcard port).  However, the code supports all options.

The changes in this PR break the combination of wildcard IP with a specific port, as bind() no longer gets called.

j-xiong:
@shefty We had a discussion prior to the PR about whether to use the existing variables or a new set.  The argument for new variables was that firewalls could configure different ranges of allowed ports for incoming connection requests and outgoing connection requests.

shefty:
I would change the variable names and text to better reflect their use.  The difference between "port range high" and "egress port range high" is unclear.  Maybe "high value for listening port" or "high value for connected socket port", etc.?

j-xiong:
I agree the variable names are not intuitive. Part of the reason is that we want to keep the original variable intact.

One possibility is that we deprecate the original variables and recommend using the following new variables:
`FI_TCP_EP_PORT_RANGE=<low>:<high>`
`FI_TCP_PEP_PORT_RANGE=<low>:<high>`

Does that sound good?

shefty:
A socket that's created using accept() is restricted to the listen/PEP socket range.  But such a socket still ends up as an 'EP' to the libfabric user.  So, I'm hesitant to refer to PEP vs EP ranges.  But I do like the idea of specifying the range using a single variable, rather than setting 2 separate ones, even if parsing the result takes more work.  It also opens the possibility of conflicting variables.

The current variable suggests that it applies to all sockets.  If you want to keep the same behavior and name, at least update the help text to describe what it actually controls.

chien-intel:
User requires finer control over TCP port range at domain level.   Use fi_set_val to set low/high port range in TCP domain object.  More details in commit message.

I moved the bind check from xnet_endpoint to xnet_bind_port as it was nested too deep.  The check is the same as before with additional condition on high port number to bind.  The binding order is: 1 - port number specified in src_address, 2 - port range if defined, 3 - no port

chien-intel:
Add another commit to use fi_set_val to variables corresponding to FI_TCP_PORT_LOW_RANGE and FI_TCP_PORT_HIGH_RANGE

shefty (changes requested):
This is better than another environment variable.

prov/tcp/src/xnet_ep.c
```diff
@@ -783,20 +812,10 @@ int xnet_endpoint(struct fid_domain *domain, struct fi_info *info,
...
-			ret = bind(ep->bsock.sock, info->src_addr,
-				(socklen_t) info->src_addrlen);
-			if (ret) {
-				FI_WARN(&xnet_prov, FI_LOG_EP_CTRL, "bind failed\n");
-				ret = -ofi_sockerr();
-				goto err3;
-			}
+		if (xnet_bind_port(domain, info, ep)) {
+			FI_WARN(&xnet_prov, FI_LOG_EP_CTRL, "bind failed\n");
+			ret = -ofi_sockerr();
```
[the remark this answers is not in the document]
chien-intel: It is a bit gnarly.  The return code from xnet_bind_to_port_range in xnet_bind_port (or new name xnet_bind_addr) is the same as bind except for the case where we exhaust port range and return -FI_EADDRNOTAVAIL that will leave errno from last bind call which would be EADDRINUSE.  Not quite correct.  Unfortunately, in theory xnet_bind_port can return -FI_EADDRNOTAVAIL and no way for caller to tell what happened.

Same problem existed with xnet_pep_sock_create calling into the xnet_bind_to_port_range.  It is just using the return code and not doing another call with -ofi_sockerr.

let me know what you want me to do for the port exhaustion case as I don't have a "correct" errno to use.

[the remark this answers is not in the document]
shefty: I would move the errno conversion down as far into the call stack as possible.  If there's some other failure down the stack, pick the most suitable errno to use.  Pushing the warnings down the call stack will help with debugging where exactly something went wrong.

chien-intel:
> This is better than another environment variable.

I agree.  We should consider moving more control knobs/capabilities under domain.

include/rdma/fi_ext.h
```diff
@@ -78,6 +78,12 @@ enum {
 	FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 };
 
+enum {
+	/* null terminated string, <low port>-<high port> */
```
shefty, line 82: update comment

man/fi_tcp.7.md
```diff
@@ -90,6 +90,24 @@ show all environment variables defined for the tcp provider.
...
   Default: disabled.
 
+The tcp provider also supports setting parameters via fi_set_val.
+
+*FI_TCP_FABRIC_PASSIVE_PORT_RANGE*
+: Set values corresponding to FI_TCP_PORT_LOW_RANGE and FI_TCP_PORT_HIGH_RANGE.
+  Use fabric object with struct xnet_port_range * set with desired values.
+  Sample:
+    struct xnet_port_range range = { .low = 9000, .high = 9100 };
+    fi_set_val(&fabric->fid, FI_TCP_FABRIC_PASSIVE_PORT_RANGE, &range);
```
shefty, line 100: The structure should be defined in a public header file available to the user.  The standard practice for this would be to define prov/tcp/include/rdma/fi_ext_tcp.h, and define a structure there.  The structure name should probably be something like fi_tcp_port_range { }.  The new header should be installed if the tcp provider is built.

As an example, you can look at the gni provider.  See prov/gni/include/fi_ext_gni.h.  (I think they should have placed the file under include/rdma.)

shefty: Btw, if the structure is defined as:
```
struct fi_tcp_port_range {
    uint16_t low;
    uint16_t high;
};
```
then you can remove the checks for the port being out of range.

PR #10021, opened 2024-05-03
contrib/aws: Modify Windows Test AMI

a-szegel:
Switch the Windows Test to use an AMI owned by the account running Jenkins

This will need to be backported to all branches we wish to run Windows tests against, or they will fail in ~90 days.


shijin-aws:
Can we make the ami id recorded in the code so we don't have to modify it in the jenkinsfile when we migrate?

a-szegel:
We can cancel this change and do it in our integration test framework instead.

PR #10027, opened 2024-05-09
02 of 10 LNX series - Minor fix to ofi_check_fabric_attr()

amirshehataornl:
util: Consider provider exclusion

When checking fabric attributes with ofi_check_fabric_attr() make sure to
consider provider exclusion.

When checking to see if a provider name is given, only consider ones which
are not excluded using the '^' character.


j-xiong:
`^` is only used in `FI_PROVIDER` to create provider filters. It should not appear in `fabric_attr->prov_name`.

j-xiong:
Close this series. Will post new PR with revised version.

PR #10028, opened 2024-05-09
04 of 10 LNX series - update get_msg()/get_tag() to take fi_peer_match structure

amirshehataornl:
Proposed updates the SRX peer callbacks


j-xiong:
See the comment for #10039. These two need to swap order, or merge into a single one.

PR #10036, opened 2024-05-13
prov/efa: Support unsolicited RDMA write with immediate receive

shijin-aws:
This patch onboards EFA provider with the rdma-core feature that supports unsolicited RDMA write with immediate receive

https://github.com/linux-rdma/rdma-core/pull/1459

. When a rdma-write with imm is unsolicited, libfabric won't release the rx pkt and bump the rx pkt counters.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -71,6 +71,30 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+
+#if HAVE_CAPS_UNSOLICITED_WRITE_RECV
+/**
+ * @brief Check whether a completion consumes recv buffer
+ *
+ * @param ibv_cq_ex extended ibv cq
+ * @return true the wc consumes a recv buffer
+ * @return false the wc doesn't consume a recv buffer
+ */
+static inline
+bool efa_rdm_cq_wc_is_unsolicited(struct ibv_cq_ex *ibv_cq_ex)
+{
+	return efadv_wc_is_unsolicited(efadv_cq_from_ibv_cq_ex(ibv_cq_ex));
+}
+
+#else
+
+static inline
+bool efa_rdm_cq_wc_is_unsolicited(struct ibv_cq_ex *ibv_cq_ex)
+{
+	return false;
+}
+
+#endif
 /**
  * @brief handle rdma-core CQ completion resulted from IBV_WRITE_WITH_IMM
  *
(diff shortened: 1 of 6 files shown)


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -120,11 +144,15 @@ void efa_rdm_cq_proc_ibv_recv_rdma_with_imm_completion(
 
 	efa_cntr_report_rx_completion(&ep->base_ep.util_ep, flags);
 
-	/* Recv with immediate will consume a pkt_entry, but the pkt is not
-	   filled, so free the pkt_entry and record we have one less posted
-	   packet now. */
-	ep->efa_rx_pkts_posted--;
-	efa_rdm_pke_release_rx(pkt_entry);
+	if (!efa_rdm_cq_wc_is_unsolicited(ibv_cq_ex)) {
```
shijin-aws, line 147: `efa_rdm_cq_proc_ibv_recv_rdma_with_imm_completion` currently needs pkt_entry as an input to get the ep where the completion is targeting. That requires we always have a valid wr_id for the wc. Are we sure it will be the case? @mrgolin

shijin-aws:
Need to rebase on https://github.com/ofiwg/libfabric/pull/10042

shijin-aws:
Need to add unit tests

shijin-aws:
@mrgolin can you review it, thanks

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -120,12 +144,20 @@ void efa_rdm_cq_proc_ibv_recv_rdma_with_imm_completion(
...
+		/**
+		 * Recv with immediate will consume a pkt_entry, but the pkt is not
+		 * filled, so free the pkt_entry and record we have one less posted
+		 * packet now.
+		 */
+		assert(pkt_entry);
+		ep->efa_rx_pkts_posted--;
+		efa_rdm_pke_release_rx(pkt_entry);
+	}
```
darrylabbate, line 160: Is this block supposed to be guarded with a directive? E.g. `#if HAVE_CAPS_UNSOLICITED_WRITE_RECV`

shijin-aws: efa_rdm_cq_wc_is_unsolicited is defined for both `HAVE_CAPS_UNSOLICITED_WRITE_RECV=1` and `HAVE_CAPS_UNSOLICITED_WRITE_RECV=0`. For the latter one, it will just return false

darrylabbate (approved):

PR #10037, opened 2024-05-15
MAINT: fix gdr_unpin_buffer warning

tylerjereddy:
* Fix the warning message when `gdr_unpin_buffer` fails, so that the message is not conflated with
the prior `gdr_unmap` call.

I believe there may be deeper issues in the code here per https://github.com/NVIDIA/gdrcopy/issues/296#issuecomment-2111473932, but just cleaning this up to start.


tylerjereddy:
done (not sure if you have `ci skip` type flags, a bit wasteful to flush it again perhaps)

j-xiong (approved):

PR #10038, opened 2024-05-15
prov/efa: Various zero-copy receive bugfixes

darrylabbate:
This includes the CQ data header size for the zero-copy receive use-case

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -57,7 +57,7 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 
 	dlist_init(&txe->queued_pkts);
 
-	if (ep->user_info->mode & FI_MSG_PREFIX)
+	if (txe->iov_count > 0 && ep->user_info->mode & FI_MSG_PREFIX)
 		ofi_consume_iov_desc(txe->iov, txe->desc, &txe->iov_count, ep->msg_prefix_size);
 	txe->total_len = ofi_total_iov_len(txe->iov, txe->iov_count);
 
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -20,56 +20,47 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 {
 	uint64_t tx_op_flags;
 
-	txe->ep = ep;
-	txe->type = EFA_RDM_TXE;
-	txe->op = op;
-	txe->tx_id = ofi_buf_index(txe);
-	txe->state = EFA_RDM_TXE_REQ;
-	txe->addr = msg->addr;
-	txe->peer = peer;
-	/* peer would be NULL for local read operation */
-	if (txe->peer) {
-		dlist_insert_tail(&txe->peer_entry, &txe->peer->txe_list);
-	}
-
-	txe->internal_flags = 0;
-	txe->bytes_received = 0;
-	txe->bytes_copied = 0;
-	txe->bytes_acked = 0;
-	txe->bytes_sent = 0;
-	txe->window = 0;
-	txe->iov_count = msg->iov_count;
-	txe->rma_iov_count = 0;
-	txe->msg_id = 0;
-	txe->efa_outstanding_tx_ops = 0;
-	dlist_init(&txe->queued_pkts);
-
-	memcpy(txe->iov, msg->msg_iov, sizeof(struct iovec) * msg->iov_count);
-	memset(txe->mr, 0, sizeof(*txe->mr) * msg->iov_count);
-	if (msg->desc)
-		memcpy(txe->desc, msg->desc, sizeof(*msg->desc) * msg->iov_count);
-	else
-		memset(txe->desc, 0, sizeof(txe->desc));
-
-	/* cq_entry on completion */
-	txe->cq_entry.op_context = msg->context;
-	txe->cq_entry.data = msg->data;
-	txe->cq_entry.len = ofi_total_iov_len(txe->iov, txe->iov_count);
-	txe->cq_entry.buf = OFI_LIKELY(txe->cq_entry.len > 0) ? txe->iov[0].iov_base : NULL;
-
-	if (ep->user_info->mode & FI_MSG_PREFIX) {
-		ofi_consume_iov_desc(txe->iov, txe->desc, &txe->iov_count, ep->msg_prefix_size);
-	}
-	txe->total_len = ofi_total_iov_len(txe->iov, txe->iov_count);
-
 	/* set flags */
 	assert(ep->base_ep.util_ep.tx_msg_flags == 0 ||
 	       ep->base_ep.util_ep.tx_msg_flags == FI_COMPLETION);
 	tx_op_flags = ep->base_ep.util_ep.tx_op_flags;
 	if (ep->base_ep.util_ep.tx_msg_flags == 0)
 		tx_op_flags &= ~FI_COMPLETION;
-	txe->fi_flags = flags | tx_op_flags;
-	txe->bytes_runt = 0;
+
+	*txe = (struct efa_rdm_ope) {
+		.type		= EFA_RDM_TXE,
+		.ep		= ep,
+		.addr		= msg->addr,
+		.peer		= peer,
+		.tx_id		= ofi_buf_index(txe),
+		.op		= op,
+		.state		= EFA_RDM_TXE_REQ,
+		.fi_flags	= flags | tx_op_flags,
+		.iov_count	= msg->iov_count,
+		.cq_entry	= (struct fi_cq_tagged_entry) {
+			.op_context	= msg->context,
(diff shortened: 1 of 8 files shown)


prov/efa/src/rdm/efa_rdm_pke_req.c
```diff
@@ -148,8 +148,9 @@ uint32_t *efa_rdm_pke_get_req_connid_ptr(struct efa_rdm_pke *pkt_entry)
 		return &raw_addr->qkey;
 	}
 
-	if (base_hdr->flags & EFA_RDM_REQ_OPT_CQ_DATA_HDR)
+	if (base_hdr->flags & EFA_RDM_REQ_OPT_CQ_DATA_HDR || pkt_entry->ep->use_zcpy_rx) {
 		opt_hdr += sizeof(struct efa_rdm_req_opt_cq_data_hdr);
```
shijin-aws, line 152: libfabric convention is to not have `{` for single line `if`

darrylabbate:
Latest push doesn't include fixes for every known bug re: CQ data header or zero-copy receive path

darrylabbate:
Latest push should fix hanging unit tests. Still testing changes related to missing logic when handling the receive completion - should be the last known bug re: zero-copy.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -20,56 +20,47 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
...
-	txe->addr = msg->addr;
-	txe->peer = peer;
+	/* set flags */
+	assert(ep->base_ep.util_ep.tx_msg_flags == 0 ||
+	       ep->base_ep.util_ep.tx_msg_flags == FI_COMPLETION);
+	tx_op_flags = ep->base_ep.util_ep.tx_op_flags;
+	if (ep->base_ep.util_ep.tx_msg_flags == 0)
+		tx_op_flags &= ~FI_COMPLETION;
+
+	*txe = (struct efa_rdm_ope) {
```
bwbarrett, line 30: yikes, this is a huge structure that we're now initializing to zero.  I'm not really sure I understand why the structure is so large, but it's concerning that every operation starts with initializing something that big.  This is not reasonable for a zero-copy path in particular.

prov/efa/src/rdm/efa_rdm_pke_rtm.c
```diff
@@ -649,36 +662,11 @@ ssize_t efa_rdm_pke_proc_matched_eager_rtm(struct efa_rdm_pke *pkt_entry)
...
-	hdr_size = pkt_entry->payload - pkt_entry->wiredata;
-	if (hdr_size != pkt_entry->ep->msg_prefix_size - sizeof(struct efa_rdm_pke)) {
-		/* if header size is wrong, the data in user buffer is not useful.
-		 * setting rxe->cq_entry.len here will cause an error cq entry
-		 * to be written to application.
-		 */
-		rxe->cq_entry.len = 0;
 	} else {
-		rxe->cq_entry.len = pkt_entry->pkt_size + sizeof(struct efa_rdm_pke);
+		efa_rdm_pke_rtm_handle_zcpy_recv_completion(pkt_entry, rxe);
```
shijin-aws, line 666: why do you still need to handle zcpy recv in this function?

prov/efa/src/rdm/efa_rdm_pke_rtm.c
```diff
@@ -635,11 +651,8 @@ void efa_rdm_pke_handle_eager_rtm_send_completion(struct efa_rdm_pke *pkt_entry)
...
 {
-	int err;
-	int hdr_size;
-	struct efa_rdm_ope *rxe;
-
-	rxe = pkt_entry->ope;
+	int err = 0;
+	struct efa_rdm_ope *rxe = pkt_entry->ope;
 
 	if (pkt_entry->alloc_type != EFA_RDM_PKE_FROM_USER_BUFFER) {
```
shijin-aws, line 657: We shouldn't expect a user buffer allocated pkt if we handle zcpy recv completion outside `efa_rdm_pke_proc_received`

bwbarrett (commented):
I'm a little worried that we're continuing down an approach where we try to thread the zero copy path through existing state machine in a really fragile way.  I know this is trying to move quickly, but we really need to rethink the approach.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -57,7 +57,7 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 
 	dlist_init(&txe->queued_pkts);
 
-	if (ep->user_info->mode & FI_MSG_PREFIX)
+	if (txe->iov_count > 0 && ep->user_info->mode & FI_MSG_PREFIX)
```
shijin-aws, line 60: I am worried that this is a hacky workaround. I would audit the code to figure out whether we need to use this function to construct ope for the handshake packet.

prov/efa/src/rdm/efa_rdm_pke_rtm.c
darrylabbate: I figure the function is just poorly named. Is it safe to remove the logic completely from `efa_rdm_pke_proc_matched_eager_rtm()`?

prov/efa/src/rdm/efa_rdm_ope.c
darrylabbate: Zero-initializing was the point of the refactor here. I find the alternative prone to bugs and less maintainable.

To be clear: I'm guessing the storage duration/stack utilization is the primary concern here, correct? I hadn't really considered that - likely explains the existing `memset()` usage.

bwbarrett: memset(0) on a big region is also a problem.  The big problem is you have a large structure that you're carrying around per operation.  In the zero copy case, you shouldn't need anything more than some way of getting back to the context to return to the user.  You've basically made it so that the fast path is limited by the slow path.

Readability / maintainability is important, but performance is the real goal.

prov/efa/src/rdm/efa_rdm_ope.c
darrylabbate: Are you referring to the change that adds `txe->iov_count > 0` to the condition or the condition as a whole? It's not currently valid to call `ofi_consume_iov_desc()` with an empty IOV - so that change is necessary.

shijin-aws: I don't want handshake efa_rdm_post_handshake call `efa_rdm_txe_construct`, so you don't have to make this change

shijin-aws: efa_rdm_txe_construct should be called with a `msg` passed by application. For handshake, we don't have any msg.

prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -961,7 +961,13 @@ void efa_rdm_pke_handle_recv_completion(struct efa_rdm_pke *pkt_entry)
 		zcpy_rxe = pkt_entry->ope;
 	}
 
-	efa_rdm_pke_proc_received(pkt_entry);
+	if (ep->use_zcpy_rx) {
```
shijin-aws, line 964: This code block should be moved to the `if (pkt_entry->alloc_type == EFA_RDM_PKE_FROM_USER_BUFFER) {` above. When pkt entry is allocated from user buffer, we do the zcpy recv handling, otherwise we do the non-zcpy recv handling via `efa_rdm_pke_proc_received`.

But Brian is doubting if this shortcut is the best approach, we can discuss that

prov/efa/src/rdm/efa_rdm_pke_rtm.c
shijin-aws: I mean you already make a branch in `efa_rdm_pke_handle_recv_completion` that the zcpy received pkts should not enter `efa_rdm_proc_received` , who calls `efa_rdm_pke_proc_matched_eager_rtm`. Then we can remove this logic safely here, right?

But Brian is doubting if this shortcut is the best approach, we can discuss that.

shijin-aws:
> I'm a little worried that we're continuing down an approach where we try to thread the zero copy path through existing 
state machine in a really fragile way. I know this is trying to move quickly, but we really need to rethink the approach.

@bwbarrett understood. I was proposing this approach earlier because the common code - big state machine `efa_rdm_proc_recv` - has too many layers in the stack that handles the ordering, tag matching, and data copy that 99.9% are not related to the zcpy receive mode where the data is already delivered when we get a completion from rdma-core. We can harmonize these two use cases as part of our efa provider rewriting. But for the short term and performance sake, I don't have a better idea now instead of making an early shortcut here. WDYT?

darrylabbate:
IIUC, an "early shortcut" would be the suggested short-term solution. Probably best to engineer a high-level fast path for zero-copy and worry about making the overall provider more clean and generic later - after performance is already obtained.

darrylabbate:
FWIW, I didn't intend for this PR to complete our zero-copy work; just fix a few bugs that were discovered while testing zero-copy receive.

PR #10039, opened 2024-05-16
03 of 10 LNX series - Add address matching callback to fi_ops_srx_peer

amirshehataornl:
include/rdma: Add address matching call back to peer_srx

It is not efficient to do a reverse lookup on the AV table when a message
is received. Some providers do not store the fi_addr_t associated with the
peer in the header passed on the wire. And it is not practical to require
providers to add that to wire header, as it would break backwards
compatibility.

In order to handle this case, an address matching callback is added to the
peer_srx.peer_ops structure. This allows the provider receiving the
message to register an address matching callback. This callback is called
by the owner provider to match an fi_addr_t with provider specific address
in the message received.

The callback allows the receiving provider to do an O(1) index into the AV
table to lookup the address of the peer, and then compare that with the
source address in the received message.

As part of this change provider specific address information needs to be
passed to the owner provider, which the owner will need to give back to the
receiving provider, when it attempts to do address matching.

Update the SHM and LINKx providers to conform with the API changes

### DIFF
--- a/include/rdma/providers/fi_peer.h
+++ b/include/rdma/providers/fi_peer.h
@@ -197,6 +197,7 @@ struct fi_ops_srx_peer {
 	int	(*start_tag)(struct fi_peer_rx_entry *entry);
 	int	(*discard_msg)(struct fi_peer_rx_entry *entry);
 	int	(*discard_tag)(struct fi_peer_rx_entry *entry);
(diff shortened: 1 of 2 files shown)


j-xiong:
`struct fi_peer_match` is undefined at this moment.

PR #10042, opened 2024-05-17
prov/efa: Get efa_rdm_ep from wc's qpn

shijin-aws:
Currently, we get ep pointer from the pkt_entry (wc's wr_id), which may be not reliable because some wc can be unsolicited (there is no pkts that the wc consumes thus no valid wr_id). This patch changes this mechanism to be deriving ep from the wc's qpn, which is always available and reliable. This is also what we are doing for dgram cq which is closest to rdma-core interface.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -78,25 +78,25 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
  * remote endpoint.  These do not have a packet context, nor do they have a
  * connid available.
  *
- * @param[in]		flags		flags (such as FI_REMOTE_CQ_DATA)
- * @param[in]		pkt_entry	packet entry
  * @param[in]		ibv_cq_ex	extended ibv cq
+ * @param[in]		flags		flags (such as FI_REMOTE_CQ_DATA)
+ * @param[in]		ep	        efa_rdm_ep
+ * @param[in]		pkt_entry	packet entry
  */
 static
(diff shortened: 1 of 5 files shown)


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -300,15 +301,20 @@ void efa_rdm_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
...
 
 	/* Call ibv_start_poll only once */
 	err = ibv_start_poll(ibv_cq->ibv_cq_ex, &poll_cq_attr);
 	should_end_poll = !err;
 
 	while (!err) {
 		pkt_entry = (void *)(uintptr_t)ibv_cq->ibv_cq_ex->wr_id;
-		ep = pkt_entry->ep;
-		assert(ep);
+		qp = efa_domain->qp_table[ibv_wc_read_qp_num(ibv_cq->ibv_cq_ex) & efa_domain->qp_table_sz_m1];
```
wenduwan, line 316: Could you remind me why we need to `... & qp_table_sz_m1`?

shijin-aws: I believe it is just for safe, we have the qp table of size `qp_table_size = roundup_power_of_two(efa_domain->device->ibv_attr.max_qp);`, and qp_num shouldn't exceeds `qp_num < ibv_attr.max_qp < qp_table_size`.

wenduwan (approved):

darrylabbate (approved):

PR #10046, opened 2024-05-23
fabtests/multinode: update arguments

### LINKED ISSUE #9330: fabtests/multinode: Multinode command line parameters are not documented
fi_multinode has test specific parameters which are not documented in either the test output (e.g. using -? option) or the man pages.  To make it worse, the test re-uses some common command line parameters, so the commont output from -? is wrong.

alex-mckinley:
Update arguments to avoid conflicts with other tests.

Will fix #9330

### DIFF
--- a/fabtests/multinode/src/harness.c
+++ b/fabtests/multinode/src/harness.c
@@ -321,17 +321,20 @@ int main(int argc, char **argv)
 	if (!hints)
 		return EXIT_FAILURE;
 
-	while ((c = getopt(argc, argv, "n:C:z:Th" CS_OPTS INFO_OPTS)) != -1) {
+	while ((c = getopt(argc, argv, "n:x:z:Ths:I:" INFO_OPTS)) != -1) {
 		switch (c) {
 		default:
 			ft_parse_addr_opts(c, optarg, &opts);
 			ft_parseinfo(c, optarg, hints, &opts);
-			ft_parsecsopts(c, optarg, &opts);
+			break;
+		case 'I':
+			opts.options |= FT_OPT_ITER;
+			opts.iterations = atoi(optarg);
 			break;
 		case 'n':
 			pm_job.num_ranks = atoi(optarg);
 			break;
-		case 'C':
+		case 'x':
 			pm_job.transfer_method = parse_caps(optarg);
 			break;
 		case 'T':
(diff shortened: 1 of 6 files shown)


alex-mckinley:
```
Usage:
    fi_multinode -s SERVER_NAME -n NUM_RANKS [OPTIONS]
    repeat command for each client
    reccomend using fabtests/scripts/runmultinode.sh

fi_multinode specific options: 

-n <num_ranks>                 Number of ranks to expect
-x <xfer_mode>                 msg or rma message mode
-I <iters>                     number of iterations
-T                             pass to enable performance timing mode
-z <pattern>                   full_mesh, ring, gather, or broadcast pattern. Default: All

General Fabtests options: 

-f <fabric>                    fabric name
-d <domain>                    domain name
-p <provider>                  specific provider name eg sockets, verbs
-B <src_port>                  non default source port number
-P <dst_port>                  non default destination port number
-s <address>                   source address
-b[=<oob_port>]                enable out-of-band address exchange and synchronization over the, optional, port
-E[=<oob_port>]                enable out-of-band address exchange only over the, optional, port
-C <number>                    simultaneous connections to server
-O <addr>                      use the provided addr for out of band
-F <addr_format>               Address format (default:FI_FORMAT_UNSPEC)
-D <device_iface>              Specify device interface: e.g. cuda, ze, neuron (default: None). Automatically enables FI_HMEM (-H)
-i <device_id>                 Specify which device to use (default: 0)
-H                             Enable provider FI_HMEM support
-R                             Register HMEM memory with fi_mr_dmabuf API
```


New fi_multinode opts/fi_multinode -h output

j-xiong:
There are a few more places that need to be updated with the new option (`-C` --> `-x`):
`fabtests/pytest/default/test_multinode.py`,
`fabtests/test_configs/tcp/io_uring.exclude`.

fabtests/multinode/src/harness.c
```diff
@@ -321,17 +321,20 @@ int main(int argc, char **argv)
 	if (!hints)
 		return EXIT_FAILURE;
 
-	while ((c = getopt(argc, argv, "n:C:z:Th" CS_OPTS INFO_OPTS)) != -1) {
+	while ((c = getopt(argc, argv, "n:x:z:Ths:" INFO_OPTS)) != -1) {
```
j-xiong, line 324: Here a new option "-s" is added, but not used later.

alex-mckinley: nvm, you are correct that -I isn't in INFO_OPTS. That also needs to be added manually.

j-xiong (approved):

PR #10049, opened 2024-05-28
Revert "prov/efa: Include CQ data header in prefix size"

shijin-aws:
We are moving to a different direction that doesn't require CQ data to be part of the hdr. Revert this change for now.

This reverts commit 90cd7d5f2d948b1042f20635536ac0eafa485437.


darrylabbate (commented):
Sign-off is missing in the commit message

shijin-aws:
Thanks, added

darrylabbate (approved):

Issue #9614, opened 2023-11-28
pthread_cond_timedwait does not use CLOCK_MONOTONIC


skypexu:
**Describe the bug**
function ofi_wait_cond uses ofi_gettime_ms() to get current time, and ofi_gettime_ms uses CLOCK_MONOTONIC,
but pthread_cond_timedwait default uses CLOCK_REALTIME, the two clocks are different.

**To Reproduce**

**Expected behavior**
ofi_gettime_ms should use CLOCK_REALTIME or make another API to return time with CLOCK_REALTIME

**Output**

**Environment:**
Linux

**Additional context**

j-xiong:
Fixed.

PR #10050, opened 2024-05-28
Added ofi_gettime interfaces to support CLOCK_REALTIME

nikhilnanal:
[src/common]: added ofi_get_realtime intefaces to support pthread_cond_timedwait which use CLOCK_REALTIME.In general CLOCK_MONOTONIC is preferred and in use by several providers, therefore new interfaces are added to support the special case of pthread apis.(ref #9614 )

[prov/sockets]: changes made to use ofi_get_realtime instead of ofi_get_time before ofi_wait_cond.

Signed-of-by: Nikhil Nanal <nikhil.nanal@intel.com>

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -304,6 +304,24 @@ uint32_t ofi_generate_seed(void)
 	return rand_seed;
 }
 
+uint64_t ofi_get_realtime_ns(void)
+{
+	struct timespec now;
+
+	clock_gettime(CLOCK_REALTIME, &now);
+	return now.tv_sec * 1000000000 + now.tv_nsec;
+}
+
+uint64_t ofi_get_realtime_us(void)
+{
+	return ofi_get_realtime_ns() / 1000;
+}
+
+uint64_t ofi_get_realtime_ms(void)
+{
+	return ofi_get_realtime_ns() / 1000000;
+}
+
 uint64_t ofi_gettime_ns(void)
 {
 	struct timespec now;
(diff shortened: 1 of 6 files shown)


src/common.c
```diff
@@ -303,7 +303,23 @@ uint32_t ofi_generate_seed(void)
 
 	return rand_seed;
 }
+uint64_t ofi_get_realtime_ns(void)
```
j-xiong, line 306: Add a line before this.

src/common.c
```diff
@@ -303,7 +303,23 @@ uint32_t ofi_generate_seed(void)
...
 
+uint64_t ofi_get_realtime_us(void)
+{
+	return ofi_get_realtime_ns() / 1000;
+}
+
+uint64_t ofi_get_realtime_ms(void)
+{
+	return ofi_get_realtime_ns() / 1000000;
+}
```
j-xiong, line 322: Add a line after this.

j-xiong (commented):
Please reformat the commit message so that:
(1) The first line (title line) is not too long (<= 72);
(2) Remove the brackets from `[src/common.c]:`, or shorten it to `core:`.

Also the prototypes of the new functions need to be added to `include/ofi.h`.

j-xiong:
We don't need the brackets (`[` and `]`) in the commit message title. For example:
```
core: blah blah
```

j-xiong:
`CLOCK_REALTIME` is undefined for Windows. Please add a definition to `include/windows/osd.h` (right before `CLOCK_MONOTONIC`). It can be defined as 0 (the value is not used anyway).

j-xiong (approved):

PR #10055, opened 2024-05-31
Intel/CI: Update DMABUF stage to use torchic partition

Juee14Desai:
DMABUF tests require to use new node which is available with torchic partition.
There is no need for 2 node tests hence the two node tests have been removed.
Log file name has been changed and number of nodes have been removed.


zachdworkin (approved):

zachdworkin:
Intel CI failure was a random failure that was not effected by this PR's changes. We do not need to wait for AWS CI since this only changes Intel CI's Jenkinsfile

PR #10057, opened 2024-06-03
[v1.20.x] Intel/CI: Cherry-pick oneccl gpu stage and dmabuf stage commits

Juee14Desai:
OneCCL GPU tests require to use new node which is available with torchic partition.
DMABUF tests require to use new node which is available with
torchic partition.
There is no need for 2 node tests hence the two node tests have
been removed.


zachdworkin (approved):

Juee14Desai:
This commit only has Intel CI changes hence AWS CI failure can be overlooked.

PR #10058, opened 2024-06-03
V1.21.x Intel/CI: Cherry-pick oneccl gpu stage and dmabuf stage commits

Juee14Desai:
OneCCL GPU tests require to use new node which is available with torchic partition.
DMABUF tests require to use new node which is available with
torchic partition.
There is no need for 2 node tests hence the two node tests have
been removed.


zachdworkin:
Intel CI failed a test because of a known race condition. These changes do not effect that test and can be merged despite its failure. No need to wait for AWS CI to finish either since these changes only effect Intel CI's Jenkinsfile

PR #10061, opened 2024-06-03
prov/coll: fix empty AV set initialization

### LINKED ISSUE #9738: fi_av_set/coll_av_set: error when trying to create an empty set
I am trying to create an empty av set of collective operations, where I will later add remote addresses (which are more complex than `(start_addr..end_addr).step_by(stride)` )

I am setting an av_set_attr and trying to create the av set as follows:
```
struct fi_av_set_attr av_set_attr = {0};
av_set_attr.count = num_remote_nodes;
av_set_attr.start_addr = FI_ADDR_NOTAVAIL;
av_set_attr.end_addr = FI_ADDR_NOTAVAIL;
av_set_attr.stride = 0;
av_set_attr.flags = 0;

struct fid_av_set *av_set;
int ret = fi_av_set(av, &av_set_attr, &av_set, NULL); //av initialized else where
if (ret) {
  printf("error: %s\n",fi_strerror(ret));
}

for (int i=0;i < num_nodes; i++){
  ret = fi_av_set_insert(av_set,remote_addrs[node_map[i]]); //node_maps takes a "collective" node id and maps it to its world id
  ...
}
```
This returns the following error: `fi_av_st failed: Cannot allocate memory (-12)`

Digging into this a little deeper it appears that in https://github.com/ofiwg/libfabric/blob/39e35c094769fe2df078866747b0c67a34bd5633/prov/coll/src/coll_av_set.c#L238

The for loop is incorrectly entered because `start_addr = end_addr = FI_ADDR_NOTAVAIL` , and then never exits the loop because `stride=0`

A simple check before the loop:
```
if (attr->start_addr != FI_ADDR_NOTAVAIL && attr->end_addr != FI_ADDR_NOTAVAIL) {
        for (i = attr->start_addr; i <= attr->end_addr; i += attr->stride) {
            if (av_set->fi_addr_count >= av_set->max_array_size)
                goto err3;

            av_set->fi_addr_array[av_set->fi_addr_count++] = i;
        }
    }
```
fixed this for me.

Alternatively for anyone looking for a workaround at the application layer, you can simple create the set with the address of the root node then add in the other addresses after
```
struct fi_av_set_attr av_set_attr = {0};
av_set_attr.count = num_remote_nodes;
av_set_attr.start_addr = remote_addrs[node_map[0]];
av_set_attr.end_addr = remote_addrs[node_map[0]];
av_set_attr.stride = 1;
av_set_attr.flags = 0;

struct fid_av_set *av_set;
int ret = fi_av_set(av, &av_set_attr, &av_set, NULL); //av initialized else where
if (ret) {
  printf("error: %s\n",fi_strerror(ret));
}

for (int i=1;i < num_nodes; i++){ //Notice we now start at 1 instead of 0
  ret = fi_av_set_insert(av_set,remote_addrs[node_map[i]]); //node_maps takes a "collective" node id and maps it to its world id
  ...
}
```

While this isn't necessarily a blocking bug, it does seem to be at odds with what the documentation says for `fi_av_set` (or I am misinterpreting the documentation) namely:

[fi_av_set](https://ofiwg.github.io/libfabric/v1.20.0/man/fi_av_set.3.html)
start_addr / end_addr
The starting and ending addresses, inclusive, to include as part of the AV set. The use of start and end address require that the associated AV have been created as type FI_AV_TABLE. Valid addresses in the AV which fall within the specified range and which meet other requirements (such as stride) will be added as initial members to the AV set. **_The start_addr and end_addr must be set to FI_ADDR_NOTAVAIL if creating an empty AV set_,** a communication key is being provided, or the AV is of type FI_AV_MAP.
The number of addresses between start_addr and end_addr must be less than or equal to the specified count value.

stride
The number of entries between successive addresses included in the AV set. The AV set will include all addresses from start_addr + stride x i, for increasing, non-negative, integer values of i, up to end_addr. A stride of 1 indicates that all addresses between start_addr and end_addr should be added to the AV set. **_Stride should be set to 0 unless the start_addr and end_addr fields are valid_**

Based on the emphasized portion of the docs above is what lead to my initial attempt.

Thanks!
-Ryan

**Environment:**
Linux, verbs;ofi_rxm, v1.20

aingerson:
An empty AV set can be created by setting start_addr and end_addr to FI_ADDR_NOTAVAIL. The for loop condition doesn't handle this condition and will cause a failure

Rework error paths a bit to return more helpful error messages.

Fixes #9738

### DIFF
--- a/prov/coll/src/coll_av_set.c
+++ b/prov/coll/src/coll_av_set.c
@@ -226,20 +226,38 @@ int coll_av_set(struct fid_av *av_fid, struct fi_av_set_attr *attr,
 	if (!av_set)
 		return -FI_ENOMEM;
 
-	if (ofi_mutex_init(&av_set->lock))
+	ret = ofi_mutex_init(&av_set->lock);
+	if (ret)
 		goto err1;
 
 	av_set->max_array_size = attr->count ? attr->count : av_attr.count;
 	av_set->fi_addr_array = calloc(av_set->max_array_size,
 				       sizeof(*av_set->fi_addr_array));
-	if (!av_set->fi_addr_array)
+	if (!av_set->fi_addr_array) {
+		ret = -FI_ENOMEM;
 		goto err2;
+	}
 
-	for (i = attr->start_addr; i <= attr->end_addr; i += attr->stride) {
-		if (av_set->fi_addr_count >= av_set->max_array_size)
-			goto err3;
+	if (attr->start_addr != FI_ADDR_NOTAVAIL &&
+	    attr->end_addr != FI_ADDR_NOTAVAIL) {
+		for (i = attr->start_addr; i <= attr->end_addr;
+		     i += attr->stride) {
+			if (av_set->fi_addr_count >= av_set->max_array_size) {
+				FI_WARN(av_set->av->prov, FI_LOG_AV,
+					"AV set size (%zu) not large enough\n",
+					av_set->max_array_size);
+				ret = -FI_EINVAL;
+				goto err3;
+			}
 
-		av_set->fi_addr_array[av_set->fi_addr_count++] = i;
+			av_set->fi_addr_array[av_set->fi_addr_count++] = i;
+		}
+	} else if (attr->start_addr != attr->end_addr) {
+		FI_WARN(av_set->av->prov, FI_LOG_AV,
+			"AV set start and end addr must both be set to"
+			"FI_ADDR_NOTAVAIL when creating empty AV set\n");
+		ret = -FI_EINVAL;
+		goto err3;
 	}
 
 	ofi_atomic_initialize32(&av_set->ref, 0);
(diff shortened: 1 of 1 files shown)


aingerson:
@rdfriese Sorry for the delayed fix for your issue but thank you again for reporting! Your proposed fix was pretty spot on - I just adapted a bit to add some error checking and handle the case where only one was set to FI_ADDR_NOTAVAIL to fully catch that misuse of the API.
If you happen to have a free moment, please take a look and verify that this looks correct to you and that it fixes your issue.
Thank!

j-xiong (approved):
Looks good, thanks for the fix!

rdfriese:
Works great, thanks for the update!

j-xiong:
There is a known issue with the DCO checking https://github.com/dcoapp/app/issues/211

j-xiong:
@aingerson The DCO issue has been fixed but doesn't trigger the status change automatically. Could you make a trivia change and do a force push?

PR #10062, opened 2024-06-04
Support for VA to PA changes

iziemba:
The following series of commits updates documentation and defines new fields to enable use-cases which may result in VA to PA changes with memory being registered.

### DIFF
--- a/include/rdma/fi_domain.h
+++ b/include/rdma/fi_domain.h
@@ -175,6 +175,7 @@ struct fi_mr_attr {
 		int		synapseai;
 	} device;
 	void			*hmem_data;
+	size_t			page_size;
 };
 
 struct fi_mr_modify {
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -244,7 +244,31 @@ The following apply to memory registration.
 
 *FI_MR_ALLOCATED*
 : When set, all registered memory regions must be backed by physical
-  memory pages at the time the registration call is made.
+  memory pages at the time the registration call is made. In addition,
+  applications must not perform operations which may result in the underlying
+  virtual address to physical page mapping to change (e.g. calling free()
+  against an allocated MR). Failing to adhere to this may result in the
+  virtual address pointing to one set of physical pages while the MR points
+  to another set of physical pages.
+
+  When unset, registered memory regions need not be backed by physical
+  memory pages at the time the registration call is made. In addition, the
+  underlying virtual address to physical page mapping is allowed to change,
+  and the provider will ensure the corresponding MR is updated accordingly.
+  This behavior enables application use-cases where memory may be frequently
+  freed and reallocated or system memory migrating to/from device memory.
+
+  When unset, the application is responsible for ensuring that a registered
+  memory region references valid physical pages while a data transfer is
+  active against it, or the data transfer may fail. Application changes to
+  the virtual address range must be coordinated with network traffic to or
+  from that range.
+
+  If unset and FI_HMEM is supported, the ability for the virtual address to
+  physical address mapping to change extends to HMEM interfaces as well. If a
+  provider cannot support a virtual address to physical address changing for
+  a given HMEM interface, the provider should support a reasonable fallback
+  or the operation should fail.
 
 *FI_MR_PROV_KEY*
 : This memory region mode indicates that the provider does not support
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -593,6 +593,7 @@ struct fi_mr_attr {
         int            synapseai;
     } device;
     void               *hmem_data;
+    size_t             page_size;
 };
 
 struct fi_mr_auth_key {
@@ -787,6 +788,26 @@ field is determined by the value specified through iface.
 
 ## hmem_data
 The hmem_data field is reserved for future use and must be null.
+
+## page_size
+Page size allows applications to optionally provide a hint at what the
+optimal page size is for the an MR allocation. Typically, providers can
+select the optimal page size. In cases where VA range has zero pages
+backing it, which is supported with FI_MR_ALLOCATED unset, the provider
+may not know the optimal page size during registration. Rather than use
+a less efficient page size, this attribute allows applications to specify
+the page size to be used.
+
+If page size is zero, provider will select the page size.
+
+If non-zero, page size must be supported by OS. If a specific page size
+is specified for a memory region during creation, all pages later
+associated with the region must be of the given size. Attaching a memory
+page of a different size to a region may result in failed transfers to
+or from the region.
+
+Providers may choose to ignore page size. This will result in a provider
+selected page size always being used.
 
 ## fi_hmem_ze_device
 
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -200,12 +200,14 @@ The following apply to memory registration.
 : When the FI_MR_LOCAL mode bit is set, applications must register all
   data buffers that will be accessed by the local hardware and provide
   a valid desc parameter into applicable data transfer operations.
-  When FI_MR_LOCAL is zero, applications are not required to register
+
+  When FI_MR_LOCAL is unset, applications are not required to register
   data buffers before using them for local operations (e.g. send and
-  receive data buffers).  The desc parameter into data transfer
-  operations will be ignored in this case, unless otherwise required
-  (e.g. se  FI_MR_HMEM).  It is recommended that applications pass in
-  NULL for desc when not required.
+  receive data buffers). Prior to libfabric 1.22, the desc parameter was
+  ignored. In libfabric 1.22 and later, the desc parameter must be either
+  valid or NULL. This behavior allows applications to optionally pass in a
+  valid desc parameter. If the desc parameter is NULL, any required local
+  memory registration will be handled by the provider.
 
   A provider may hide local registration requirements from applications
   by making use of an internal registration cache or similar mechanisms.
@@ -214,7 +216,7 @@ The following apply to memory registration.
   In order to support as broad range of applications as possible,
   without unduly affecting their performance, applications that wish to
   manage their own local memory registrations may do so by using the
-  memory registration calls.
+  memory registration calls and passing in a valid desc parameter.
 
   Note: the FI_MR_LOCAL mr_mode bit replaces the FI_LOCAL_MR fi_info mode bit.
   When FI_MR_LOCAL is set, FI_LOCAL_MR is ignored.
@@ -303,6 +305,10 @@ The following apply to memory registration.
   discover the appropriate device memory registration attributes, if
   applicable.
 
+  Similar to if FI_MR_LOCAL is unset, if FI_MR_HMEM is unset, applications may
+  optionally pass in a valid desc parameter. If the desc parameter is NULL, any
+  required local memory registration will be handled by the provider.
+
   If FI_MR_HMEM is set, but FI_MR_LOCAL is unset, only device buffers
   must be registered when used locally.  In this case, the desc parameter
   passed into data transfer operations must either be valid or NULL.
@@ -310,11 +316,17 @@ The following apply to memory registration.
   parameter must either be valid or NULL.
 
 *FI_MR_COLLECTIVE*
-: This bit is associated with the FI_COLLECTIVE capability.  When set,
-  the provider requires that memory regions used in collection operations
-  must explicitly be registered for use with collective calls.  This
-  requires registering regions passed to collective calls using the
+: This bit is associated with the FI_COLLECTIVE capability.
+
+  If FI_MR_COLLECTIVE is set, the provider requires that memory regions used in
+  collection operations must explicitly be registered for use with collective
+  calls. This requires registering regions passed to collective calls using the
   FI_COLLECTIVE flag.
+
+  If FI_MR_COLLECTIVE is unset, memory registration for collection operations is
+  optional. Applications may optionally pass in a valid desc parameter. If the
+  desc parameter is NULL, any required local memory registration will be handled
+  by the provider.
 
 *Basic Memory Registration*
 : Basic memory registration was deprecated in libfabric version 1.5, but
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -526,20 +526,32 @@ When binding the memory region to an endpoint, flags should be 0.
 
 ## fi_mr_refresh
 
-The use of this call is required to notify the provider of any change
-to the physical pages backing a registered memory region if the
-FI_MR_MMU_NOTIFY mode bit has been set.  This call informs the provider
-that the page table entries associated with the region may have been
-modified, and the provider should verify and update the registered
-region accordingly.  The iov parameter is optional and may be used
-to specify which portions of the registered region requires updating.
-Providers are only guaranteed to update the specified address ranges.
+The use of this call is to notify the provider of any change to the physical
+pages backing a registered memory region. This call must be supported by
+providers requiring FI_MR_MMU_NOTIFY and may optionally be supported by
+providers not requiring FI_MR_ALLOCATED.
 
-The refresh operation has the effect of disabling and re-enabling
-access to the registered region.  Any operations from peers that attempt
-to access the region will fail while the refresh is occurring.
-Additionally, attempts to access the region by the local process
+This call informs the provider that the page table entries associated with
+the region may have been modified, and the provider should verify and update
+the registered region accordingly. The iov parameter is optional and may be
+used to specify which portions of the registered region requires updating.
+
+Providers are only guaranteed to update the specified address ranges. Failing
+to update a range will result in an error being returned.
+
+When FI_MR_MMU_NOTIFY is set, the refresh operation has the effect of
+disabling and re-enabling access to the registered region. Any operations
+from peers that attempt to access the region will fail while the refresh is
+occurring. Additionally, attempts to access the region by the local process
 through libfabric APIs may result in a page fault or other fatal operation.
+
+When FI_MR_ALLOCATED is unset, -FI_ENOSYS will be returned if a provider
+does not support fi_mr_refresh. If supported, the provider will atomically
+update physical pages of the MR associated with the user specified address
+ranges. The MR will remain enabled during this time.
+
+Note: FI_MR_MMU_NOTIFY set behavior takes precedence over FI_MR_ALLOCATED unset
+behavior.
 
 The fi_mr_refresh call is only needed if the physical pages might have
 been updated after the memory region was created.
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -289,12 +289,19 @@ The following apply to memory registration.
 
 *FI_MR_HMEM*
 : This mode bit is associated with the FI_HMEM capability.
+
   If FI_MR_HMEM is set, the application must register buffers that
   were allocated using a device call and provide a valid desc
   parameter into applicable data transfer operations even if they are
   only used for local operations (e.g. send and receive data buffers).
   Device memory must be registered using the fi_mr_regattr call, with
   the iface and device fields filled out.
+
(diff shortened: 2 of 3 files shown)


iziemba:
This PR is not ready to land. At this point, I am looking for feedback on the proposed changes.

include/rdma/fi_domain.h
```diff
@@ -175,6 +175,7 @@ struct fi_mr_attr {
 		int		synapseai;
 	} device;
 	void			*hmem_data;
+	size_t			page_size_hint;
```
iziemba: This should be page_size

man/fi_mr.3.md
```diff
@@ -514,21 +514,33 @@ When binding the memory region to an endpoint, flags should be 0.
...
-modified, and the provider should verify and update the registered
-region accordingly.  The iov parameter is optional and may be used
-to specify which portions of the registered region requires updating.
-Providers are only guaranteed to update the specified address ranges.
-
-The refresh operation has the effect of disabling and re-enabling
-access to the registered region.  Any operations from peers that attempt
-to access the region will fail while the refresh is occurring.
-Additionally, attempts to access the region by the local process
+The use of this call is notify the provider of any change to the physical
```
chuckfossen, line 517: is to notify

man/fi_mr.3.md
```diff
@@ -287,27 +264,64 @@ The following apply to memory registration.
...
+
+  If FI_MR_LOCAL is set, applications must register all system memory buffers
+  and provide a valid desc parameter into applicable local data transfer
+  operations.
+
+  Note: The FI_MR_LOCAL mr_mode bit replaces the FI_LOCAL_MR fi_info mode bit.
+  When FI_MR_LOCAL is set, FI_LOCAL_MR is ignored.
+
+  FI_MR_HMEM is associated with FI_HMEM capability. If FI_MR_HMEM is set,
+  applications must register all device buffers provide a valid desc parameter
```
chuckfossen, line 299: s/provide/and provide/

man/fi_mr.3.md
```diff
@@ -287,27 +264,64 @@ The following apply to memory registration.
...
-  the provider requires that memory regions used in collection operations
-  must explicitly be registered for use with collective calls.  This
-  requires registering regions passed to collective calls using the
-  FI_COLLECTIVE flag.
+*Explicit Memory Registration Requirements*
+: Depending on memory and/or operation type, a provider may have different
+  requirements around explicit memory registration. Explicit memory
+  registration means the application must register data buffers being used for
+  the local operation (e.g. send and receive). Three different MR modes bits
+  are defined for this: FI_MR_LOCAL, FI_MR_HMEM, and FI_MR_COLLECTIVE.
```
shefty, line 289: FI_MR_HMEM and FI_MR_COLLECTIVE are independent from FI_MR_LOCAL.

FI_MR_COLLECTIVE is **not** restricted to local operations only.  It indicates that the buffers must be registered specifically for use with any collective operation.  The registration call must include the FI_COLLECTIVE flag.

FI_MR_HMEM is not quite a local memory registration requirement for device memory.  Even if the buffer is only access by an RDMA write, for example, FI_MR_HMEM requires that the buffer be registered using the hmem registration semantics.  This is true even in cases of unified addressing, where device memory shares the same VA space as system memory.

Please don't group these 3 bits together under the same definition.

man/fi_mr.3.md
```diff
@@ -219,7 +219,24 @@ The following apply to memory registration.
...
+  applications must not perform operations which may result in the underlying
+  virtual address to physical page mapping to change (e.g. calling free()
+  against an allocated MR). Failing to adhere to this may result in the
+  virtual address pointing to one set of physical pages while the MR points
+  to another set of physical pages.
+
+  Note: The operating system deciding to migrate application physical pages
+  may result in the virtual address pointing to one set of physical pages while
+  the MR points to another set of physical pages. Operating system features
+  which may do this should be disabled.
```
shefty, line 232: This note needs to be reworded or dropped.  The entire point of virtual addressing is to allow the physical pages behind them to change.  You can't 'disable' this feature without turning off the OS.

man/fi_mr.3.md
```diff
@@ -219,7 +219,24 @@ The following apply to memory registration.
...
+
+  When unset, all registered memory regions need not be backed by physical
```
shefty, line 234: remove "all"

man/fi_mr.3.md
```diff
@@ -219,7 +219,24 @@ The following apply to memory registration.
...
+  memory pages at the time the registration call is made. In addition, the
+  underlying virtual address to physical page mapping is allowed to change,
+  and the provider will ensure the corresponding MR is updated accordingly.
+  This behavior enables application use-cases where memory may be frequently
+  freed and reallocated or system memory migrating to/from device memory.
```
shefty, line 239: I'd add another paragraph stating something like:

```
The application is responsible for ensuring that a registered memory region references valid physical pages while a data transfer is active against it, or the data transfer may fail.  Application changes to the virtual address range must be coordinated with network traffic to or from that range.
 ```

File: include/rdma/fi_domain.h (hunk 1 above)
Comment (shefty) [reply include/rdma/fi_domain.h]: This breaks the ABI and requires some sort of versioning checks in all the providers and relevant updates to the man page.

File: include/rdma/fi_domain.h [hunk 8]
```diff
@@ -175,7 +175,7 @@ struct fi_mr_attr {
 		int		synapseai;
 	} device;
 	void			*hmem_data;
-	size_t			page_size_hint;
+	size_t			page_size;
```
shefty, line 178: Okay - the previous commit should be merged with this one

man/fi_mr.3.md
```diff
@@ -776,6 +777,27 @@ field is determined by the value specified through iface.
...
 
+## page_size
+Page size allows applications to notify the provider the page size to be
+used for an MR allocation. Typically, providers can select the optimal
+page size. In cases where a VA range has zero pages backing it, the
+provider may not know the optimal page size during registration. Rather
+than always use a potential inferior page size, allow apps to specify
+the page size to be used.
+
+Page size must be power of 2. If the provider supports the page size
```
shefty, line 788: Can drop the power of 2 requirement.  The page size must come from the OS supported page sizes...

man/fi_mr.3.md
```diff
@@ -776,6 +777,27 @@ field is determined by the value specified through iface.
 ## hmem_data
 The hmem_data field is reserved for future use and must be null.
 
+## page_size
+Page size allows applications to notify the provider the page size to be
```
shefty, line 781: Indicate up front that this field is optional and may be set to 0.

man/fi_mr.3.md
```diff
@@ -776,6 +777,27 @@ field is determined by the value specified through iface.
...
+
+Page size must be power of 2. If the provider supports the page size
+argument and later determines that this page size cannot be used, RDMA
+operations may fail. For example, suppose the user sets page size to 1
+GiB but the corresponding pages are only 2 MiB. The MR may be configured
+for 1 GiB and the provider cannot change MR page size without
+reconfiguring the MR.
+
+Providers will treat page size as a hint. Meaning that if a provider
+does not support user defined page size, it will be ignored.
```
shefty, line 796: This seems wrong.  If the provider cannot support the size, the call should fail.  The provider support may be to fall back to a less efficient size, but it must be supported.

man/fi_mr.3.md
```diff
@@ -776,6 +777,27 @@ field is determined by the value specified through iface.
 ## hmem_data
 The hmem_data field is reserved for future use and must be null.
 
+## page_size
+Page size allows applications to notify the provider the page size to be
+used for an MR allocation. Typically, providers can select the optimal
+page size. In cases where a VA range has zero pages backing it, the
+provider may not know the optimal page size during registration. Rather
+than always use a potential inferior page size, allow apps to specify
```
shefty, line 785: Replace "potential inferior" with "less efficient"

man/fi_mr.3.md
```diff
@@ -776,6 +777,27 @@ field is determined by the value specified through iface.
...
+the page size to be used.
+
+Page size must be power of 2. If the provider supports the page size
+argument and later determines that this page size cannot be used, RDMA
+operations may fail. For example, suppose the user sets page size to 1
+GiB but the corresponding pages are only 2 MiB. The MR may be configured
+for 1 GiB and the provider cannot change MR page size without
+reconfiguring the MR.
```
shefty, line 793: I'd replace this example with a specific requirement.  "If a specific page_size is specified for a memory region during creation, all pages later associated with the region must be of the given size.  Attaching a memory page of a different size to a region may result in failed transfers to or from the region."

man/fi_mr.3.md
iziemba: Without ODP, RDMA requires this does not change otherwise you risk potential data corruption. In Linux, with HPC at least, there are knobs to disable common things (THP and NUMA balancing) to prevent memory migration.

If we take"disable" this feature meaning turn off the OS, at face value, this means the MR cache + memhooks should never be used since memhooks would not be able to capture this.

Would it be better to state which Linux features may cause this and suggest disabling?

man/fi_mr.3.md
iziemba: Will fix.

man/fi_mr.3.md
iziemba: Sounds good.

include/rdma/fi_domain.h
iziemba: Yep. I'll add in the ABI compat stuff once we agree on API.

man/fi_mr.3.md
iziemba: I see where you are coming from. The problem this is trying to solve is how do apps know if a provider supports this field? Or do apps have to be updated with provider specific information knowing which providers may support this?

man/fi_mr.3.md
shefty: You need to be clear what you mean by migrate.  Virtual memory exists to allow the physical pages to change.

man/fi_mr.3.md
shefty: The change is updating the API.  The providers need to support the call being made with and without this field present, and it may be present, but uninitialized by the app.  The providers can handle this based on the API version selected by the app when opening the domain.

Apps should expect all providers to support this field.  The behavior describe by this change is that this value allows the provider to optimize for a specific page size.  Without that knowledge, the provider works the same as it does today and may be less efficient.

I don't think any check is needed by the app to see if the provider implements support.  If a provider doesn't it simply ignores that the field exists.

man/fi_mr.3.md
iziemba: RDMA without ODP requires physical pages to not change while registered.

I ended up just deleting this documentation.

iziemba:
Ready for round 2.

man/fi_mr.3.md
```diff
@@ -200,12 +200,12 @@ The following apply to memory registration.
...
+
+  When FI_MR_LOCAL is unset, applications are not required to register
   data buffers before using them for local operations (e.g. send and
-  receive data buffers).  The desc parameter into data transfer
-  operations will be ignored in this case, unless otherwise required
-  (e.g. se  FI_MR_HMEM).  It is recommended that applications pass in
-  NULL for desc when not required.
+  receive data buffers). Applications may optionally pass in a valid
+  desc parameter. If the desc parameter is NULL, any required local
+  memory registration will be handled by the provider.
```
shefty, line 208: I'm going to guess that in practical use cases, this change will have no effect on existing apps, but it still technically breaks the ABI and needs to be treated as such.  This change may be better going into the 2.0 update, where version checks in the data path APIs wouldn't be needed.

A possible version check in the provider could be:
```
if (desc && (version >= FI_VERSION(1,?)) {
    /* the desc is valid */
} 
```

Providers that don't perform the version check are technically okay, since the field just continues to be ignored by them.

The comment should be updated to include what libfabric version the desc parameter is no longer treated as optional and must be NULL.

prov/cxi/src/cxip_mr.c
```diff
@@ -1386,6 +1386,10 @@ static int cxip_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 
 	dom = container_of(fid, struct cxip_domain, util_domain.domain_fid);
 
+	if (FI_VERSION_GE(dom->fab->util_fabric.fabric_fid.api_version, FI_VERSION(1, 22)) &&
+	    attr->page_size)
+		return -FI_ENOSYS;
+
```
shefty, line 1392: I don't think you need to update any of the providers.  If page_size is set, the provider will just ignore it (treat as 0), and the behavior will be the same as it is today.

I would keep the update to the util code, however.

shefty (commented):
thanks

prov/cxi/src/cxip_mr.c
iziemba: Having providers ignore page_size means page_size is a hint. Currently documentation essentially states that if page_size is set, providers must support it or -FI_ENOSYS will be returned.

I'm ok with making this a hint. Thoughts?

shefty: Wouldn't the providers naturally support any page size already?  It may not be optimal, but it should work.

If FI_MR_ALLOCATED is set, then registrations are backed by pages, and the provider should be discovering the size.  Hopefully the discovered size matches the app value that is passed in.  This is less efficient than the provider just using the passed in value, but it should work.  Most providers fall into this category.

If FI_MR_ALLOCATED is NOT set, the situation isn't much different.  In this case, the provider is likely assuming some small page size, e.g. 4k.  That size also works for any given page size.

man/fi_mr.3.md
shefty: FI_MR_HYBRID doesn't help, as it doesn't provide any additional value beyond a version check.  If the provider requires FI_MR_HYBRID, that's the same as requiring that the app request version 1.22 (or whatever version this lands in).  In either case, the provider either needs a runtime check in the data path or it only runs with updated apps (existing apps will fail).  The runtime check to support existing apps just changes to something like:
```
if (desc && ep->domain->mr_mode & FI_MR_HYBRID)
```
For the version check, it would need to be documented everywhere 'desc' is documented to be ignored that it instead must be either valid or NULL.  My guess is that is what apps are doing anyway.

iziemba: The benefit of an new bit would be apps have to opt into this new desc behavior. ABI is not broken in this case since apps can use API 1.22 and get the existing ignore desc behavior.

prov/cxi/src/cxip_mr.c
iziemba: Right. So page size should be a hint providers can optionally use or ignore?

shefty: I think so.

iziemba:
Ready for round 3.

man/fi_mr.3.md
```diff
@@ -520,21 +520,33 @@ When binding the memory region to an endpoint, flags should be 0.
...
+providers requiring FI_MR_MMU_NOTIFY and may optionally be supported by
+providers not requiring FI_MR_ALLOCATED.
+
+This call informs the provider that the page table entries associated with
+the region may have been modified, and the provider should verify and update
+the registered region accordingly. The iov parameter is optional and may be
+used to specify which portions of the registered region requires updating.
+
+Providers are only guaranteed to update the specified address ranges. Failing
+to update a range will result -FI_EFAULT being returned.
```
chuckfossen, line 534: I believe there may be other errors encountered by the provider besides -FI_EFAULT. Is the provider required to return -FI_EFAULT even if another error is encountered?

swelch (approved):
This looks like a good solution for ODP.

man/fi_mr.3.md
```diff
@@ -244,7 +244,25 @@ The following apply to memory registration.
 
 *FI_MR_ALLOCATED*
 : When set, all registered memory regions must be backed by physical
-  memory pages at the time the registration call is made.
+  memory pages at the time the registration call is made. In addition,
```
chuckfossen, line 247: We need to update FI_MR_ALLOCATED in fi_domain.3.md as well.

shefty: I think we're okay here.  The descriptions in fi_domain.md aren't intended to be comprehensive, since they reference back to this page for full details.

shefty (approved):
thanks!  changes look good to me

man/fi_mr.3.md
```diff
@@ -520,21 +520,33 @@ When binding the memory region to an endpoint, flags should be 0.
...
+to update a range will result -FI_EFAULT being returned.
+
+When FI_MR_MMU_NOTIFY is set, the refresh operation has the effect of
+disabling and re-enabling access to the registered region. Any operations
+from peers that attempt to access the region will fail while the refresh is
+occurring. Additionally, attempts to access the region by the local process
 through libfabric APIs may result in a page fault or other fatal operation.
 
+When FI_MR_ALLOCATED is unset, -FI_ENOSYS will be returned if a provider
+does not support fi_mr_refresh. If supported, the provider will atomically
```
aingerson, line 543: Coming in late here just to clarify - "atomically"? or "automatically"?

shefty: The list of backing pages is being updated, and the update is "atomic".  An access to the region either references the old list of pages, or the new list, and not some combination of both.

man/fi_mr.3.md
j-xiong: @iziemba Any reason we need a specific error number here? Would `will result in an error code being returned` work just as fine?

chuckfossen: I agree, we should just state an error will be returned.

man/fi_mr.3.md
chuckfossen: ok

iziemba:
I'll have all comments address mid this week.

iziemba:
One question I have is do we need to address whether a provider can support FI_MR_ALLOCATED = 0 by HMEM iface types?

shefty:
We could restrict FI_MR_ALLOCATED to memory allocated using system calls only.  FI_MR_HMEM overrides FI_MR_LOCAL for HMEM.  It could also override the FI_MR_ALLOCATED bit for HMEM.

iziemba:
So FI_MR_HMEM = 1 and FI_MR_ALLOCATED = 0 results in VA <-> PA migration for FI_HMEM_SYSTEM only? The issue with this is that it now forces other FI_MR_HMEM behavior like explicit memory registration for local others, right?

Thinking out loud... CUDA, ROCR, and ZE device drivers all support a invalidation mechanism. This opens the door for implementing ODP for even device memory. We could have a bitmask to denote which HMEM ifaces support VA <-> PA changes with FI_MR_ALLOCATED = 0?

shefty:
In general, to keep things reasonable for a user, if a provider can't support all features for all types of memory that an app can use, my vote is for the provider to fallback and disable a feature entirely.  The permutations simply become untenable.

I don't want apps to deal with: "I can support X with system memory.  And that GPU is okay.  But that GPU isn't; it's too old.  And that NIC memory, I can do, but only if you touch the memory using system calls.  And NVMe is okay, but only if it's off the PCI bus, not attached to CXL..."

iziemba:
I agree the permutations become untenable.

How about updating the FI_MR_ALLOCATED = 0 with something like the following:

_If FI_MR_ALLOCATED = 0 is and FI_HMEM is supported, the ability for the VA to PA mapping to change extends to to HMEM interfaces as well. If a provider cannot support VA to PA changing for a given HMEM iface, the provider should support a reasonable fallback or the operation should fail._

shefty:
I'm good with that update.

iziemba:
Comments addressed and the following added:

If FI_MR_ALLOCATED = 0 is and FI_HMEM is supported, the ability for the VA to PA mapping to change extends to to HMEM interfaces as well. If a provider cannot support VA to PA changing for a given HMEM iface, the provider should support a reasonable fallback or the operation should fail.

j-xiong (approved):

PR #9847, opened 2024-03-01
Fabtests: Run rma_pingpong in AWS CI

a-szegel:

PR #10064, opened 2024-06-06
Fabtest: test_rma_pingpong bugfix

jiaxiyan:
1. The fixture memory_type was missing in test_rma_pingpong_range_no_inject
2. The RMA pingpong test should not run with 0-byte messages as zero-sized transfers are not supported. We should use different message sizes for rma pingpong, starting from non-zero values.

### DIFF
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -158,18 +158,20 @@ int pingpong_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 	if (ft_check_opts(FT_OPT_VERIFY_DATA))
 		inject_size = 0;
 
-	ret = ft_sync();
-	if (ret)
-		return ret;
-
 	if (opts.transfer_size == 0) {
 		FT_ERR("Zero-sized transfers not supported");
 		return EXIT_FAILURE;
 	}
 
-	/* Init rx_buf with invalid iteration number */
+	/* Init rx_buf with invalid iteration number.
+	 * This must be done before the sender sends any data.
+	 */
 	if (rma_op == FT_RMA_WRITE)
 		*(rx_buf + opts.transfer_size - 1) = (char)-1;
+
+	ret = ft_sync();
+	if (ret)
+		return ret;
 
 	if (opts.dst_addr) {
 		for (i = 0; i < opts.iterations + opts.warmup_iterations; i++) {
--- a/fabtests/pytest/efa/test_rma_pingpong.py
+++ b/fabtests/pytest/efa/test_rma_pingpong.py
@@ -10,33 +10,27 @@ def rma_pingpong_message_size(request):
     return request.param
 
 
-@pytest.mark.parametrize("operation_type", ["writedata", "write"])
+@pytest.mark.parametrize("operation_type", ["writedata"])
 @pytest.mark.parametrize("iteration_type",
                          [pytest.param("short", marks=pytest.mark.short),
                           pytest.param("standard", marks=pytest.mark.standard)])
 def test_rma_pingpong(cmdline_args, iteration_type, operation_type, completion_semantic, memory_type):
-    if memory_type != "host_to_host" and operation_type == "write":
-        pytest.skip("no hmem memory support for pingpong_rma write test")
     command = "fi_rma_pingpong -e rdm"
     command = command + " -o " + operation_type + " " + perf_progress_model_cli
     efa_run_client_server_test(cmdline_args, command, iteration_type, completion_semantic, memory_type, "all")
 
 
 @pytest.mark.functional
-@pytest.mark.parametrize("operation_type", ["writedata", "write"])
+@pytest.mark.parametrize("operation_type", ["writedata"])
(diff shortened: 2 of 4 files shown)


fabtests/pytest/efa/test_rma_pingpong.py
```diff
@@ -31,7 +36,7 @@ def test_rma_pingpong_range(cmdline_args, operation_type, completion_semantic, m
 
 @pytest.mark.functional
 @pytest.mark.parametrize("operation_type", ["writedata", "write"])
-def test_rma_pingpong_range_no_inject(cmdline_args, operation_type, completion_semantic, inject_message_size):
+def test_rma_pingpong_range_no_inject(cmdline_args, operation_type, completion_semantic, inject_message_size, memory_type):
```
wenduwan, line 39: Does this test work for 0 B message?

fabtests/pytest/efa/test_rma_pingpong.py
```diff
@@ -19,6 +19,11 @@ def test_rma_pingpong(cmdline_args, iteration_type, operation_type, completion_s
 
 @pytest.mark.functional
 @pytest.mark.parametrize("operation_type", ["writedata", "write"])
+@pytest.mark.parametrize("message_size", ["r:1,4,65",
+                                          "r:4048,4,4148",
+                                          "r:8000,4,9000",
+                                          "r:17000,4,18000",
+                                          "r:1,1024,1048577"])
```
wenduwan, line 26: I'm not sure if we want `r:1,4,65` - it expands to `1,5,9,13,...` which does not look right.

Overall I think it's probably easier to skip the test if message size starts from 0.

jiaxiyan: so we just test "r:4048,4,4148",
                          "r:8000,4,9000",
                          "r:17000,4,18000"?

wenduwan (approved):

shijin-aws:
Can you add this test to AWS CI so we can test the change? @a-szegel tried to add it earlier but hit similar issue that this PR is going to address https://github.com/ofiwg/libfabric/pull/9847

shijin-aws:
Please check AWS CI failure

jiaxiyan:
test_rma_pingpong_range  error message is 

```
 Client timed out
 libfabric:172391:1717730028::efa:ep_ctrl:efa_rdm_ep_destroy_buffer_pools():713<warn> Closing ep with unreleased rxe
 Data check error (D !=) at byte 17771 for buffer size 17775
```

Not sure how this change affects test_fabtests_functional[non-dmabuf]

wenduwan:
@jiaxiyan Were you able to find the root cause of the timeout?

wenduwan:
To run the test we also need to add the pytest file to https://github.com/ofiwg/libfabric/blob/main/fabtests/Makefile.am

fabtests/pytest/efa/test_rma_pingpong.py
```diff
@@ -3,39 +3,40 @@
...
                          [pytest.param("short", marks=pytest.mark.short),
                           pytest.param("standard", marks=pytest.mark.standard)])
 def test_rma_pingpong(cmdline_args, iteration_type, operation_type, completion_semantic, memory_type):
-    if memory_type != "host_to_host" and operation_type == "write":
-        pytest.skip("no hmem memory support for pingpong_rma write test")
     command = "fi_rma_pingpong -e rdm"
     command = command + " -o " + operation_type + " " + perf_progress_model_cli
     # rma_pingpong test with data verification takes longer to finish
-    timeout = max(540, cmdline_args.timeout)
+    timeout = max(1020, cmdline_args.timeout)
```
wenduwan, line 21: Based on my test this timeout is too long. Suggest removing the max.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -158,19 +158,21 @@ int pingpong_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 	if (ft_check_opts(FT_OPT_VERIFY_DATA))
 		inject_size = 0;
 
-	ret = ft_sync();
-	if (ret)
-		return ret;
-
```
shijin-aws, line 164: I missed the context of this change. Why do we need such commit? @wenduwan

shijin-aws: Are you saying that fi_write can happen before that overwrite is done on receiver side, if we don't make the overwrite before sync?

wenduwan: Yes. For example, consider client(sender) and server(receiver):
- Receiver posts receive(RX1)
- Sender sends msg with delivery complete(TX1)
- Receiver gets receive completion(RX1)
- Receiver sends reply with delivery complete(TX2)
- Sender gets send completion(TX1)
- Sender posts receive(RX2)
- Sender gets receive completion(RX2)
- Sender starts fi_write
- Receive RX buffer modified by fi_write
- Receiver gets send completion(TX2)
- Receiver modifies RX buffer

shijin-aws:
Intel CI doesn't run fabtests'pytest. Ignoring.

PR #10066, opened 2024-06-07
fabtests: support testing inject with cq data

shijin-aws:
Currently, ft_post_inject_buf doesn't support posting fi_injectdata and fi_tinjectdata when given a cq data. This patch adds such feature without changing the current tests' behavior


j-xiong (approved):
Looks fine. Thanks!

PR #10067, opened 2024-06-07
prov/efa: do not do copy for inline send

shijin-aws:
Efa can directly access memory if the buffer is host memory and the pkt size (hdr + data) is within inline buf size. Libfabric doesn't need to memcpy in this case.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke_utils.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_utils.c
@@ -41,7 +41,7 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
 					  size_t data_size)
 {
 	int tx_iov_index, ret;
-	bool p2p_available;
+	bool efa_can_access_memory;
 	size_t tx_iov_offset, copied;
 	struct efa_mr *iov_mr;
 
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -71,9 +71,16 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
...
 		if (ret < 0)
 			return ret;
-		p2p_available = ret;
+		efa_can_access_memory = ret;
+	/**
+	 * efa can directly access memory if the buffer is host memory
+	 * and the pkt size (hdr + data) is within inline buf size.
+	 */
+	} else if (!efa_mr_is_hmem(iov_mr) &&
+	           payload_offset + data_size <= efa_rdm_ep_domain(pke->ep)->device->efa_attr.inline_buf_size){
```
wenduwan, line 80: nit missing space between ) and {

wenduwan: Also, consider reordering the if else checks:
```
if (!efa_mr_is_hmem(iov_mr) &&
	           payload_offset + data_size <= efa_rdm_ep_domain(pke->ep)->device->efa_attr.inline_buf_size) {
    ...
} else if (iov_mr) {
    ...
} else {
    ...
}
```

The current logic does not work for `FI_HMEM_SYSTEM`

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_pke_utils.c
shijin-aws: Updated!

PR #10077, opened 2024-06-11
man: Fix inject endpoint opt names

darrylabbate:


j-xiong (approved):
Thanks! I have no idea what happened ;)

shijin-aws (approved):

PR #10078, opened 2024-06-11
prov/efa: Add some missing newlines to log statements

darrylabbate:


prov/efa/src/efa_env.c
```diff
@@ -98,7 +98,7 @@ void efa_env_param_get(void)
 			"FI_EFA_TX_MIN_CREDITS was set to %d, which is <= 0.\n"
 			"This value will cause EFA communication to deadlock.\n"
 			"Please unset the environment variable or set it to a positive number.\n"
-			"Your application will now abort.",
+			"Your application will now abort.\n",
```
darrylabbate, line 101: This one (and another like it elsewhere) has always been an odd format, but I'll save that battle for another day.

shijin-aws (approved):

PR #10082, opened 2024-06-12
prov/efa: Distinguish max RMA size from max message transfer size

darrylabbate:
This also adds getopt/setopt logic for the new supported endpoint options, as well as some validation logic for the `len` arguments to applicable operations.


shijin-aws (commented):
Mostly LGTM, I am not sure whether we have to validate the size every time during transmission calls.

shijin-aws (approved):

darrylabbate:
> I am not sure whether we have to validate the size every time during transmission calls.

I'll take another look at this - could be superfluous

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1346,6 +1347,84 @@ static int efa_rdm_ep_set_shared_memory_permitted(struct efa_rdm_ep *ep, bool sh
...
+ * @param[in,out]	ep		EFA RDM endpoint
+ * @param[in]		max_msg_size	Requested maximum msg size
+ *
+ * @return		0 on success, -FI_EINVAL otherwise
+ *
+ * @sa #FI_OPT_MAX_MSG_SIZE
+ */
+static int efa_rdm_ep_set_max_msg_size(struct efa_rdm_ep *ep, size_t max_msg_size)
+{
+	if (max_msg_size > ep->max_msg_size) {
```
darrylabbate, line 1365: This needs to be modified s.t. `max_msg_size` is compared against the EFA provider's static limit. Otherwise, the application would be unable to increase the max size via `fi_setopt()` after reducing it.

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -557,10 +553,6 @@ ssize_t efa_rdm_msg_tinject(struct fid_ep *ep_fid, const void *buf, size_t len,
 	struct efa_rdm_peer *peer;
 
 	efa_rdm_ep = container_of(ep_fid, struct efa_rdm_ep, base_ep.util_ep.ep_fid.fid);
-	if (len > efa_rdm_ep->inject_size) {
-		EFA_WARN(FI_LOG_CQ, "invalid message size %ld for inject.\n", len);
-		return -FI_EINVAL;
-	}
```
shijin-aws, line 556: you need an assert here right?

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -587,10 +579,6 @@ ssize_t efa_rdm_msg_tinjectdata(struct fid_ep *ep_fid, const void *buf, size_t l
...
```
shijin-aws, line 582: do you need an assert

prov/efa/src/rdm/efa_rdm_msg.c
darrylabbate: I omitted these since the documentation for `len` in fi_tagged(3) doesn't have the same qualifier re: valid lengths, as in fi_msg(3) and fi_rma(3). This could easily be an oversight since the documentation was writted ~10 years ago and never changed. WDYT?

shijin-aws: I think it's clearly an oversight in the documentation (cc @j-xiong ). You can add assert and update the man page

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -978,6 +966,7 @@ ssize_t efa_rdm_msg_recv(struct fid_ep *ep_fid, void *buf, size_t len,
 	struct efa_rdm_ep *ep;
 
 	ep = container_of(ep_fid, struct efa_rdm_ep, base_ep.util_ep.ep_fid.fid);
+	assert(len <= ep->max_msg_size);
```
shijin-aws, line 969: I don't think we need to apply max_msg_size for recv. Though the documentation doesn't say that, I think it's totally fine that user post a bigger recv buffer than the incoming msg size, i.e. the max_msg_size only applies to transmission.

prov/efa/src/rdm/efa_rdm_msg.c
j-xiong: Agree.

shijin-aws:
I think we also need to revert https://github.com/ofiwg/libfabric/commit/642fc3a22cc73d170c32e84eac29d6a97439bc6e

prov/efa/src/rdm/efa_rdm_msg.c
darrylabbate: Likewise for RMA read?

shijin-aws: RMA read is still a transmission, so the limit applies.

shijin-aws (approved):

darrylabbate:
Unit test failure after (under) reverting 642fc3a22cc73d170c32e84eac29d6a97439bc6e. Likely due to other unit tests utilizing the ability to specify `max_msg_size` via hints in 2be8604de759c2dfe3fa3e2b2109247f4c761b59.

@shijin-aws Are we still happy with the revert of 642fc3a22cc73d170c32e84eac29d6a97439bc6e? Reading through #9388, it seems like it was not hastily merged. If so, I can probably alter the unit test logic to use `fi_setopt()` instead.

shijin-aws:
Ah yes, I think it is caused by these tests https://github.com/ofiwg/libfabric/blob/main/prov/efa/test/efa_unit_test_ep.c#L646-L709. In order to fix it, you need to make it use fi_setopt to set max_msg_size, and alter the zcpy_recv set procedure during ep creation. I am ok with you make it as a separate PR

PR #10090, opened 2024-06-14
Opx upstream June 2024

charlesshereda:
Changes to opx provider path include:

-Further enhancements to support NVIDIA GPUs, included CUDA-allocated bounce buffers and in-provider support for GDRCopy
-Enhancements to enable support for CN5000 hardware
-Better checking for TID support
-General TID enhancements
-Pkey error handling
-Send work queue splitting
-Support for OPX tracer for profiling purposes
-Coverity scan fixes
-Fixes and enhancements to logging and debug messages
-Bug fixes including: 
-- intranode RMA read fixes
-- compile issues
-- shared memory segment index creation bug


j-xiong:
The ubuntu build error is real:
```
./prov/opx/include/rdma/opx/fi_opx_hfi1_packet.h:111:9: error: array index 256 is past the end of the array (which contains 10 elements) [-Werror,-Warray-bounds]
        return FI_OPX_HFI_BTH_HIGH_OPCODE_STRINGS[FI_OPX_HFI_BTH_OPCODE_TAG_RZV_RTS+1]; /* INVALID */
```

charlesshereda:
Thanks, yes, it's hidden under an #ifdef OPX_VERBOSE_TRIGGER and Bob and I are discussing. I'll change to draft.

charlesshereda:
Ready for review.

prov/opx/include/rdma/opx/fi_opx_hmem.h
```diff
@@ -107,6 +120,74 @@ enum fi_hmem_iface fi_opx_hmem_get_iface(const void *ptr,
...
+	if (hmem_handle != 0 && len <= threshold) {
+		OPX_TRACER_TRACE(OPX_TRACER_BEGIN, "GDRCOPY-TO-DEV");
+		cuda_gdrcopy_to_dev(hmem_handle, dest, src, len);
+		OPX_TRACER_TRACE(OPX_TRACER_END_SUCCESS, "GDRCOPY-TO-DEV");
+		ret = 0;
+	} else {
+		OPX_TRACER_TRACE(OPX_TRACER_BEGIN, "CUDAMEMCPY-TO-HMEM");
+		ret = (int) cudaMemcpy(dest, src, len, cudaMemcpyHostToDevice);
+		OPX_TRACER_TRACE(OPX_TRACER_END_SUCCESS, "CUDAMEMCPY-TO-HMEM");
+	}
```
j-xiong, line 147: Does that mean opx only support CUDA when CUDA is enabled?

j-xiong (commented):
I roughly went through the commits and didn't any obvious issue. Please confirm if the question below is true by design.

prov/opx/include/rdma/opx/fi_opx_hmem.h
charlesshereda: Not exactly sure what you're asking since it's a little unclear. If you're asking, will OPX only support CUDA if it's enabled through HAVE_CUDA, the answer is, we believe so. There might be a situation in which this falls through to `ofi_copy_from_hmem` and CUDA is still used, but from what we've seen, the CUDA calls are wrapped in `#if HAVE_CUDA` in `hmem_cuda.c` as well...

If instead you're asking, can we only support CUDA at present when CUDA is enabled (and not also alternatives like ROCR at the same time), then in this PR, that's correct; however, we do have an upcoming PR that changes this to be slightly more general with an addition of `#if HAVE_ROCR` logic in addition to the `#if HAVE_CUDA` case. As we add support for additional GPUs, we can add more of that type of logic.

I intend to open that other PR before the next release, btw.

j-xiong: @charlesshereda Your second paragraph answered my question. I was asking because the code block just assumes the memory is CUDA allocated.

j-xiong (approved):

Issue #10052, opened 2024-05-30
hmem: compilation fails with incompatible pointer types on macOS with gcc 14


pmrsbot:
**Describe the bug**
A basic configuration of libfabric on macOS with gcc 14 fails to build.

**To Reproduce**
```sh
brew install gcc
git clone https://github.com/ofiwg/libfabric
cd libfabric
./autogen.sh
./configure
make -j
```
**Expected behavior**
Build succeeds.

**Output**
```
src/hmem.c: In function 'ofi_async_copy_hmem_iov_buf':
src/hmem.c:330:45: error: passing argument 3 of 'ofi_iov_bytes_to_copy' from incompatible pointer type [-Wincompatible-pointer-types]
  330 |                                             &hmem_iov_offset, &hmem_buf);
      |                                             ^~~~~~~~~~~~~~~~
      |                                             |
      |                                             uint64_t * {aka long long unsigned int *}
In file included from src/hmem.c:41:
./include/ofi_iov.h:87:47: note: expected 'size_t *' {aka 'long unsigned int *'} but argument is of type 'uint64_t *' {aka 'long long unsigned int *'}
   87 |                         size_t *size, size_t *offset, char **copy_buf)
      |                                       ~~~~~~~~^~~~~~
src/hmem.c: In function 'ofi_copy_hmem_iov_buf':
src/hmem.c:366:45: error: passing argument 3 of 'ofi_iov_bytes_to_copy' from incompatible pointer type [-Wincompatible-pointer-types]
  366 |                                             &hmem_iov_offset, &hmem_buf);
      |                                             ^~~~~~~~~~~~~~~~
      |                                             |
      |                                             uint64_t * {aka long long unsigned int *}
./include/ofi_iov.h:87:47: note: expected 'size_t *' {aka 'long unsigned int *'} but argument is of type 'uint64_t *' {aka 'long long unsigned int *'}
   87 |                         size_t *size, size_t *offset, char **copy_buf)
      |                                       ~~~~~~~~^~~~~~
src/hmem.c: In function 'ofi_dev_reg_copy_hmem_iov_buf':
src/hmem.c:398:45: error: passing argument 3 of 'ofi_iov_bytes_to_copy' from incompatible pointer type [-Wincompatible-pointer-types]
  398 |                                             &hmem_iov_offset, &hmem_buf);
      |                                             ^~~~~~~~~~~~~~~~
      |                                             |
      |                                             uint64_t * {aka long long unsigned int *}
[...]

PR #10096, opened 2024-06-17
Fix incompatible pointer warnings on macOS with GCC 14

raffenet:
Fixes compilation issue reported in #10052.

### DIFF
--- a/src/iov.c
+++ b/src/iov.c
@@ -37,10 +37,10 @@
 #include <ofi.h>
 #include <ofi_iov.h>
 
-uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, uint64_t iov_offset,
-			  void *buf, uint64_t bufsize, int dir)
+size_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, size_t iov_offset,
+			void *buf, size_t bufsize, int dir)
 {
-	uint64_t done = 0, len;
+	size_t done = 0, len;
 	char *iov_buf;
 	size_t i;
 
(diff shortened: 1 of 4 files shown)


src/iov.c
```diff
@@ -37,8 +37,8 @@
 #include <ofi.h>
 #include <ofi_iov.h>
 
-uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, uint64_t iov_offset,
-			  void *buf, uint64_t bufsize, int dir)
+uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, size_t iov_offset,
```
darrylabbate, line 40: Probably a good idea to change the return type as well. I wasn't able to find any context wrt why `uint64_t` was chosen over `size_t`, but it's likely an oversight considering the compatibility.

raffenet: Makes sense to me. Let me give it a try.

darrylabbate (approved):

src/iov.c
```diff
@@ -37,8 +37,8 @@
 #include <ofi.h>
 #include <ofi_iov.h>
 
-uint64_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, uint64_t iov_offset,
-			  void *buf, uint64_t bufsize, int dir)
+size_t ofi_copy_iov_buf(const struct iovec *iov, size_t iov_count, size_t iov_offset,
+                        void *buf, size_t bufsize, int dir)
 {
 	uint64_t done = 0, len;
```
darrylabbate, line 43: Change to `size_t` (sorry; missed this).

raffenet: No worries. New version coming.

raffenet: @darrylabbate I added one more commit to convert the remaining `uint64_t` in `src/iov.c`. I can squash before merge if it looks OK.

darrylabbate: I think this LGTM. Much appreciated.

raffenet: 👍 should I squash the first and third commits? Anything else to do before merging?

darrylabbate: You can squash the first and third commits. If no one else has comments, we can merge after CI passes.

PR #10099, opened 2024-06-18
prov/efa: Fix efa_rdm_cq_wc_is_unsolicited check

shijin-aws:
efadv_wc_is_unsolicited() can only be called when driver/FW supports unsolicited write recv, in which we create IBV CQ with EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV. Otherwise the function pointer will be a NULL and cause segfault.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -83,7 +83,7 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
 static inline
 bool efa_rdm_cq_wc_is_unsolicited(struct ibv_cq_ex *ibv_cq_ex)
 {
-	return efadv_wc_is_unsolicited(efadv_cq_from_ibv_cq_ex(ibv_cq_ex));
+	return efa_device_support_unsolicited_write_recv() ? efadv_wc_is_unsolicited(efadv_cq_from_ibv_cq_ex(ibv_cq_ex)) : false;
 }
 
 #else


shijin-aws:
@mrgolin @ygoldamzn

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -83,7 +83,7 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
 static inline
 bool efa_rdm_cq_wc_is_unsolicited(struct ibv_cq_ex *ibv_cq_ex)
 {
-	return efadv_wc_is_unsolicited(efadv_cq_from_ibv_cq_ex(ibv_cq_ex));
+	return efa_device_support_unsolicited_write_recv() ? efadv_wc_is_unsolicited(efadv_cq_from_ibv_cq_ex(ibv_cq_ex)) : false;
```
mrgolin, line 86: Would be nice to somehow link it to CQ creation flags so that if we change the logic there it won't break this code.
In addition I think you can save the if inside ```efa_device_support_unsolicited_write_recv()```.

shijin-aws: Can you query CQ creation flags after CQ is created?

shijin-aws:
Tested on p5 with rdma-core master branch, confirmed the PR can fix the segfault

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: > In addition I think you can save the if inside efa_device_support_unsolicited_write_recv().

Yes this can be removed

mrgolin: >Can you query CQ creation flags after CQ is created?

There is no such API in IB verbs. I think that holding some flag in efa_rdm_cq is a good idea.

shijin-aws: Can I | EFADV_WC_EX_WITH_IS_UNSOLICITED to cq even if the driver/FW doesn't support? Then I don't need to maintain a separate wc_flags in libfabric's efa cq. Similar thing for the QP flags. I just think maintaining another wc_flags inside libfabric does make the code a bit messy.

shijin-aws: Just tested it, I cannot add QP flags when driver doesn't support, adding WC flags unconditionally works.

shijin-aws: Just tested, I still cannot call this function when the driver doesn't support, even if I add the flags in cq creation attribute.

shijin-aws: Then I think it's not safe to link efadv_wc_is_unsolicited() with the CQ creation flags.

mrgolin: If the feature isn't supported and you try to create CQ with EFADV_WC_EX_WITH_IS_UNSOLICITED, CQ creation should fail. You're saying this isn't the case?

shijin-aws: I can still create a CQ with this flag on hpc6a/p5 without such feature... as long as I build with rdma-core master

shijin-aws: I will merge this change first and revisit if we have a better plan here.

PR #10108, opened 2024-06-19
fabtests: Synchronize on Initialization

zachdworkin:
fabtests: Synchronize on Initialization
fabtests/functional: Add manual init sync to fi_rdm_multiclient

Some providers (verbs ud) might require the server to be fully initialized before the client process calls getinfo with the server address. This causes a No Data Available error due to the fi_info call failing during initialization by not being able to find the name on the server. This is seen most often in cases where a socket (usually oob [out of band]) is initialized before getinfo in the ft_init_fabric sequence. Adding a sync only if an oob socket has been initialized to order the initialization correctly will prevent this from happening.

The syncronization is for client to start getinfo only after the server is done initializing everything.

fi_rdm_multiclient test needs its client startup to have a manual sync since it does not follow the normal codepath of going through ft_init_fabric like the server and other tests do. This sync will only happen on the first client that connects because it is only necessary to give the server enough time to spin up all the resources. It is not necessary for future clients because the server has already started all of those resources.

### DIFF
--- a/fabtests/functional/rdm.c
+++ b/fabtests/functional/rdm.c
@@ -49,9 +49,13 @@ static int run(void)
 
 	while (nconn && !ret) {
 		ret = ft_send_recv_greeting(ep);
+		if (ret)
+			return ret;
 
 		if (--nconn && !ret) {
 			ret = ft_accept_next_client();
+			if (ret)
+				return ret;
 		}
 	}
 
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -987,7 +987,11 @@ int ft_accept_next_client() {
 		ret = ft_reset_oob();
 		if (ret)
 			return ret;
-	}
+
+		ret = ft_sock_sync(oob_sock, 0);
+		if (ret)
+			return ret;
+}
 	return ft_init_av();
 }
 
@@ -1305,6 +1309,12 @@ int ft_init_fabric(void)
 	if (ret)
 		return ret;
 
+	if (oob_sock >= 0 && opts.dst_addr) {
+		ret = ft_sock_sync(oob_sock, 0);
+		if (ret)
+			return ret;
+	}
+
 	ret = ft_getinfo(hints, &fi);
 	if (ret)
 		return ret;
@@ -1320,6 +1330,12 @@ int ft_init_fabric(void)
 	ret = ft_enable_ep_recv();
 	if (ret)
 		return ret;
+
+	if (oob_sock >= 0 && !opts.dst_addr) {
+		ret = ft_sock_sync(oob_sock, 0);
+		if (ret)
+			return ret;
+	}
 
 	ret = ft_init_av();
 	if (ret)
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -3916,31 +3916,31 @@ int ft_sock_recv(int fd, void *msg, size_t len)
 	return err ? err: 0;
 }
 
-int ft_sock_sync(int value)
+int ft_sock_sync(int fd, int value)
 {
 	int result = -FI_EOTHER;
 	int ret;
 
 	if (listen_sock < 0) {
-		ret = ft_sock_send(sock, &value,  sizeof value);
+		ret = ft_sock_send(fd, &value,  sizeof value);
 		if (ret) {
 			FT_PRINTERR("ft_sock_send", ret);
 			return ret;
 		}
 
-		ret = ft_sock_recv(sock, &result, sizeof result);
+		ret = ft_sock_recv(fd, &result, sizeof result);
 		if (ret) {
 			FT_PRINTERR("ft_sock_recv", ret);
 			return ret;
 		}
 	} else {
-		ret = ft_sock_recv(sock, &result, sizeof result);
+		ret = ft_sock_recv(fd, &result, sizeof result);
 		if (ret) {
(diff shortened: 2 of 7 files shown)


zachdworkin:
bot:aws:retest

zachdworkin:
@shijin-aws can you please tell me why the first run of AWS CI failed?

shijin-aws:
We have a multi-node test that involves fi_rdm failed

```
client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.8.166 'timeout 360 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10108/install/fabtests/bin/fi_rdm -p efa -E=9228 172.31.13.34'"'"''
client_stdout:
libfabric:69633:1718843593::efa:ep_ctrl:efa_rdm_peer_destruct():56<warn> Closing EP with unacked CONNREQs in flight
Sending message...
Send completion received

client returncode: 0

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.13.34 'timeout 360 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10108/install/fabtests/bin/fi_rdm -p efa -E=9228 172.31.13.34'"'"''
client_stdout:

client returncode: 107

server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.13.34 'timeout 360 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10108/install/fabtests/bin/fi_rdm -C 2 -p efa -E=9228'"'"''
server_stdout:

client_0_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.8.166 'timeout 360 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10108/install/fabtests/bin/fi_rdm -p efa -E=9228 172.31.13.34'"'"''
client_0_stdout:
libfabric:69633:1718843593::efa:ep_ctrl:efa_rdm_peer_destruct():56<warn> Closing EP with unacked CONNREQs in flight
Sending message...
Send completion received

client_1_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.13.34 'timeout 360 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn FI_EFA_USE_DEVICE_RDMA=0 /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10108/install/fabtests/bin/fi_rdm -p efa -E=9228 172.31.13.34'"'"''
client_1_stdout:
```

shijin-aws:
I think the failure is related. Need to look into further.

zachdworkin:
I agree that these changes broke the test. I think it might need some manual progression like how I had to add it to fi_rdm_multi_client. It looks like this test is just doing fi_rdm which goes through the main codepath and will have the second client hang while waiting to recv the oob socket from the server (the server only sends it once). Do you have any ideas for how to solve this problem?

j-xiong:
OOB for multiple clients is handled in `ft_accept_next_cleint`, which calls `ft_reset_oob()` which in turns calls `ft_init_oob()`. You can try adding a send inside `ft_accept_next_client` after `ft_reset_oob` is called.

zachdworkin:
@j-xiong how is this?

fabtests/functional/rdm.c
```diff
@@ -53,6 +53,11 @@ static int run(void)
 		if (--nconn && !ret) {
 			ret = ft_accept_next_client();
 		}
+		if (oob_sock >= 0 && !opts.dst_addr) {
+			ret = ft_sock_sync(oob_sock, 1);
+			if (ret)
+				return ret;
+		}
```
j-xiong, line 60: This should be moved into the `if` block above, otherwise the server would call extra sync for the first client.

j-xiong:
Looks good now. Let's see if CI would find any issue.

shijin-aws:
The same test failed again, looking at it now

shijin-aws:
So the test is run like this

server expects to get messages from 2 clients

```
fi_rdm -C 2 -p efa -E
```

client 1 and 2 comes and left and does ping-pong with server
The sever does OK when the first client comes and left

But when the second client comes it starts to get the error

```
ft_sock_recv(): common/shared.c:3951, ret=-107 (Transport endpoint is not connected)
```

shijin-aws:
I can reproduce the error with TCP provider as well

```
ubuntu@ip-172-31-39-234:~/PortaFiducia/build/libraries/libfabric/main/source/libfabric/fabtests$ fi_rdm -C 2 -p tcp -E
Waiting for message from client...
Data check OK
Received data from client: Hello from Client!
ft_sock_recv(): common/shared.c:3951, ret=-107 (Transport endpoint is not connected
```

Can you look into it?

j-xiong (approved):

PR #10113, opened 2024-06-21
prov/efa: Introduce user recv QP feature and RDM v4 protocol bump

shijin-aws:
This PR contains a series of commits that introduce the user recv QP feature used for zero copy receive. See the updated RDM v4 protocol for details.

It worth noticing that this PR doesn't really use this feature for zero-copy recv, which will be done in follow up PRs


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1202,26 +1220,26 @@ static int efa_rdm_ep_ctrl(struct fid *fid, int command, void *arg)
 		 * shared memory region.
 		 */
 		if (ep->shm_ep) {
-                        peer_srx_context.srx = util_get_peer_srx(ep->peer_srx_ep);
-                        peer_srx_attr.op_flags |= FI_PEER;
-                        ret = fi_srx_context(efa_rdm_ep_domain(ep)->shm_domain,
-					     &peer_srx_attr, &peer_srx_ep, &peer_srx_context);
-                        if (ret)
-                                goto err_unlock_and_destroy_qp;
+			peer_srx_context.srx = util_get_peer_srx(ep->peer_srx_ep);
```
shijin-aws, line 1222: This fixes some weird indentation in this blocks

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -24,6 +24,12 @@ OFI_DECL_RECVWIN_BUF(struct efa_rdm_pke*, efa_rdm_robuf, uint32_t);
  */
 #define EFA_RDM_PEER_HANDSHAKE_QUEUED      BIT_ULL(5)
 
+struct efa_rdm_peer_user_recv_qp
+{
+	uint32_t qpn;
+	uint32_t qkey;
+};
```
darrylabbate, line 31: Nit: I think I'd like the definition to be something more generic like `struct efa_qp_attr` defined in efa_base_ep.h. Then `efa_qp.{qp_num,qkey}` can be replaced with the same struct; making the (de)serialization a little simpler and future-proof. Just an idea.

darrylabbate (approved):
I think everything looks fine functionally.

As expected, the  handshake protocol is starting to become unruly. Eventually I'd like to see us implement something like a [TLV encoding scheme](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value) for handshake packets, but that's out of scope here.

prov/efa/src/rdm/efa_rdm_peer.h
shijin-aws: It's good point. I have made a solid mind whether we should name {qp_num, qkey} as efa_qp_attr because there may be some other attribute that peer doesn't need here. Can be addressed in a separate PR.

shijin-aws:
Thanks for reviewing the document, updated

PR #10115, opened 2024-06-24
prov/efa: Queue txes when handshake is enforced but not made

shijin-aws:
Currently, when a handshake is required but not made for a request to a given peer, we always return EAGAIN, which caused nasty user experience during the startup stage. This PR contains a series of commits that allows queuing the requests (when it's not for inject), and return 0 in such situation.


darrylabbate (approved):

shijin-aws:
Latest push fixed unit test failure and added new unit tests to cover the changes

zachdworkin:
@shijin-aws PR had the ucx rdm_tagged_peek test failure which was removed from #10124. If you rebase and re-push it will fix it.

PR #10116, opened 2024-06-25

wenduwan:
This patch removes the trial device memory registration with EFA device during domain initialization; instead, we query the sysfs to retrieve the P2P provider information emitted by the kernel module.

As a result, we have to delay the dmabuf support status check to the 1st application fi_mr_reg* call:
- We always register the region via ibv_reg_dmabuf_mr if the application requests FI_MR_DMABUF
- We will disable dmabuf if the hmem interface does not have P2P support
- For the 1st fi_mr_reg* call, we will try ibv_reg_dmabuf_mr
- If the 1st fi_mr_reg* call failed via ibv_reg_dmabuf_mr, we will NOT try ibv_reg_dmabuf_mr again, but always fallback to ibv_reg_mr API for that hmem interface
- Otherwise we will always use ibv_reg_dmabuf_mr for that hmem interface


wenduwan:
TODO:
- Unit test
- Neuron test
- Container test

PR #10117, opened 2024-06-25
prov/verbs: Optimize search for device max inline size

zachdworkin:
Let the user set the FI_VERBS_INLINE_SIZE environment variable to decide the maximum size of an inline transfer. The getinfo call will check if the device is capable of this size. If it is capable it will use it. If it not capable then the size is too big and we can goto the backwards search to find the largest inline size that is smaller than the FI_VERBS_INLINE_SIZE variable.

If a user wants the maximum inject size that a device can do then they should set this FI_VERBS_INLINE_SIZE variable to 0 so that the device can find its largest inject size.

### DIFF
--- a/man/fi_verbs.7.md
+++ b/man/fi_verbs.7.md
@@ -177,8 +177,10 @@ The verbs provider checks for the following environment variables.
   deprecated) EP type supports only 1
 
 *FI_VERBS_INLINE_SIZE*
-: Default maximum inline size. Actual inject size returned in fi_info
-  may be greater (default: 64)
+: Maximum inline size for the verbs device. Actual inline size returned may be
+  different depending on device capability. This value will be returned by
+  fi_info as the inject size for the application to use. Set to 0 for the
+  maximum device inline size to be used. (default: 256).
 
 *FI_VERBS_MIN_RNR_TIMER*
 : Set min_rnr_timer QP attribute (0 - 31) (default: 12)
--- a/prov/verbs/src/verbs_init.c
+++ b/prov/verbs/src/verbs_init.c
@@ -477,6 +477,7 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
 	struct ibv_cq *cq;
 	int max_inline = 2;
 	int rst = 0;
+	int pos, neg;
 	const char *dev_name = ibv_get_device_name(context->device);
 	uint8_t i;
 
@@ -501,6 +502,21 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
 		qp_attr.cap.max_recv_sge = 1;
 	}
 	qp_attr.sq_sig_all = 1;
+
+	if (vrb_gl_data.def_inline_size >= max_inline) {
+		qp_attr.cap.max_inline_data = vrb_gl_data.def_inline_size;
+		qp = ibv_create_qp(pd, &qp_attr);
+		if (qp) {
+			rst = qp_attr.cap.max_inline_data;
+			goto out;
+		}
+
+		/*
+		 * Truescale and iWarp will not reach here.
+		 */
+		max_inline = vrb_gl_data.def_inline_size;
+		goto backwards_query;
+	}
 
 	do {
 		if (qp)
@@ -527,7 +543,8 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
 	} while (qp && (max_inline < INT_MAX / 2) && (max_inline *= 2));
 
 	if (rst != 0) {
-		int pos = rst, neg = max_inline;
+backwards_query: ;
+		pos = rst, neg = max_inline;
 		do {
 			max_inline = pos + (neg - pos) / 2;
 			if (qp)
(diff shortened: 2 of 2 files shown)


prov/verbs/src/verbs_init.c
```diff
@@ -493,6 +493,18 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
 	}
 	qp_attr.sq_sig_all = 1;
 
+	if (vrb_gl_data.def_inline_size >= max_inline) {
+		qp_attr.cap.max_inline_data = vrb_gl_data.def_inline_size;
+		qp = ibv_create_qp(pd, &qp_attr);
+		if (qp) {
+			rst = vrb_gl_data.def_inline_size;
```
j-xiong, line 500: Need to handle the special case of truescale and iwarp here. Both would successfully create qp even if `qp_attr.cap.max_inline_size` is too big. Truescale would return 0 in `qp_attr.cap.max_inline_size`, iwarp would return the actual size there.

prov/verbs/src/verbs_init.c
```diff
@@ -493,6 +493,18 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
...
+			goto out;
+		}
+
+		rst = max_inline = vrb_gl_data.def_inline_size;
```
j-xiong, line 504: Need to start over from the next do loop instead. Use `vrb_gl_data.def_inline_size` in the condition of the matching `while`.

zachdworkin:
bot:aws:retest

prov/verbs/src/verbs_init.c
```diff
@@ -502,6 +502,26 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
...
+		if (qp) {
+			/*
+			 * truescale always sets max_inline_data to 0 with a
+			 * valid qp creation. So, update rst with qp_attr.
+			 *
+			 * iWarp is able to create qp with unsupported
+			 * max_inline so we can use the value directly.
+			*/
+
+			rst = qp_attr.cap.max_inline_data;
```
chien-intel, line 517: it is correct to use qp_attr.cap.max_inline_data regardless of the adapter.  This is from ibv_create_qp man page:

       **The function ibv_create_qp() will update the qp_init_attr->cap struct with the actual QP values of the QP that was created; the  values  will  be  greater than or equal to the values requested.**

chien-intel (approved):
looks good to me.

prov/verbs/src/verbs_init.c
zachdworkin: thank you! That was my logic exactly for 517.

Should I also be using it on 521 then? I thought accessing vrb_gl_data might be faster than accessing qp_attr at that point. I believe either is correct but do you have some insight that I don't?

chien-intel: you can use cap.max_inline_data when the QP was created successfully, so no.

shijin-aws:
Yes it is, you can ignore. BTW I have fixed the AWS CI now.

shijin-aws:
@zachdworkin yes please, AWS CI disabled verbs in the build anyway

prov/verbs/src/verbs_init.c
j-xiong: I think we can remove this comment (L509-L516) and add a comment before L521 saying that should not reach here for truescale and iWarp case.

prov/verbs/src/verbs_init.c
```diff
@@ -527,6 +547,7 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
 	} while (qp && (max_inline < INT_MAX / 2) && (max_inline *= 2));
 
 	if (rst != 0) {
+backwards_query: ;
```
j-xiong, line 550: Is the `;` redundant here?

zachdworkin: no, it is necessary to resolve build errors

prov/verbs/src/verbs_init.c
```diff
@@ -502,6 +502,21 @@ int vrb_find_max_inline(struct ibv_pd *pd, struct ibv_context *context,
...
+		qp_attr.cap.max_inline_data = vrb_gl_data.def_inline_size;
+		qp = ibv_create_qp(pd, &qp_attr);
+		if (qp) {
+			rst = qp_attr.cap.max_inline_data;
+			goto out;
+		}
+
+		/*
+		 * Truescale and iWarp will not reach here.
+		*/
```
j-xiong, line 515: Try keep the `*` aligned.

man/fi_verbs.7.md
```diff
@@ -177,8 +177,10 @@ The verbs provider checks for the following environment variables.
   deprecated) EP type supports only 1
 
 *FI_VERBS_INLINE_SIZE*
-: Default maximum inline size. Actual inject size returned in fi_info
-  may be greater (default: 64)
+: Maximum inline size for the verbs device. Actual inline size returned may be
+  lower if not supported by the device. This value will be returned by fi_info
```
j-xiong, line 181: Actual inline size can also be higher than the value. For example, on success, qp_attr.cap.max_inline_data can be greater than the input value.

j-xiong: could change to `Actual inline size returned may be different depending on the device capability`.

j-xiong (approved):

zachdworkin:
@j-xiong Can this be merged? Shijin said AWS CI doesnt test verbs.

PR #10129, opened 2024-06-26

wenduwan:
This patch removes the trial device memory registration with EFA device during domain initialization; instead, we query the sysfs to retrieve the P2P provider information emitted by the kernel module.

As a result, we have to delay the dmabuf support status check to the 1st application fi_mr_reg* call:
- We always register the region via ibv_reg_dmabuf_mr if the application requests FI_MR_DMABUF
- We will disable dmabuf if the hmem interface does not have P2P support
- For the 1st fi_mr_reg* call, we will try ibv_reg_dmabuf_mr
- If the 1st fi_mr_reg* call failed via ibv_reg_dmabuf_mr, we will NOT try ibv_reg_dmabuf_mr again, but always fallback to ibv_reg_mr API for that hmem interface
- Otherwise we will always use ibv_reg_dmabuf_mr for that hmem interface

### DIFF
--- a/prov/efa/src/efa_env.c
+++ b/prov/efa/src/efa_env.c
@@ -37,6 +37,7 @@ struct efa_env efa_env = {
 	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
 	.use_sm2 = false,
 	.huge_page_setting = EFA_ENV_HUGE_PAGE_UNSPEC,
+	.p2p_file_suffix= "/device/p2p",
 };
 
 /**
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -5,6 +5,8 @@
 #include "efa_hmem.h"
 #include "rdm/efa_rdm_pkt_type.h"
 
+#define P2P_PROV_MAX_LEN 32
+
 #if HAVE_CUDA || HAVE_NEURON
 static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_domain) {
 	int mtu_size;
@@ -31,6 +33,7 @@ static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_
 static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_domain, enum fi_hmem_iface iface)
 {
 	struct efa_hmem_info *info = &efa_domain->hmem_info[iface];
+	size_t tmp_value;
 
 	/* Fall back to FI_HMEM_SYSTEM initialization logic when p2p is unavailable */
 	if (!info->p2p_supported_by_device)
@@ -56,6 +59,12 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 		fi_param_get_size_t(&efa_prov, "runt_size", &info->runt_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &info->min_read_msg_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
+		if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
+			EFA_WARN(FI_LOG_DOMAIN,
+			         "The environment variable FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE was set, "
+			         "but EFA HMEM via Cuda API only supports eager and runting read protocols. "
+			         "The variable will not modify CUDA memory run config.\n");
+		}
 		break;
 	case FI_HMEM_NEURON:
 		info->runt_size = EFA_NEURON_RUNT_SIZE;
@@ -65,12 +74,31 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 		fi_param_get_size_t(&efa_prov, "runt_size", &info->runt_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &info->min_read_msg_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
+		if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
+			EFA_WARN(FI_LOG_DOMAIN,
+			         "The environment variable FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE was set, "
+			         "but EFA HMEM via Neuron API only supports eager and runting read protocols. "
+			         "The variable will not modify CUDA memory run config.\n");
+		}
 		break;
 	case FI_HMEM_SYNAPSEAI:
 		info->runt_size = 0;
 		info->max_medium_msg_size = 0;
 		info->min_read_msg_size = 1;
 		info->min_read_write_size = 1;
+		if (-FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_max_medium_message_size", &tmp_value) ||
+		    -FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &tmp_value) ||
+		    -FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &tmp_value) ||
+		    -FI_ENODATA != fi_param_get_size_t(&efa_prov, "runt_size", &tmp_value)) {
+			EFA_WARN(FI_LOG_DOMAIN,
+			        "One or more of the following environment variable(s) were set: ["
+			        "FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE, "
+			        "FI_EFA_INTER_MIN_READ_MESSAGE_SIZE, "
+			        "FI_EFA_INTER_MIN_READ_WRITE_SIZE, "
+			        "FI_EFA_RUNT_SIZE"
+			        "], but EFA HMEM via Synapse only supports long read protocol. "
+			        "The variable(s) will not modify Synapse memory run config.\n");
+		}
 		break;
 	default:
 		break;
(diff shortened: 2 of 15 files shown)


prov/efa/src/efa_hmem.c
```diff
@@ -78,6 +80,106 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
...
+		goto out;
+	}
+
+	if (strlen(p2p_prov) < strlen(p2p_prov_prefix) ||
+	    strncmp(p2p_prov, p2p_prov_prefix, strlen(p2p_prov_prefix))) {
+		EFA_INFO(FI_LOG_DOMAIN, "P2P provider does not support hmem interface: %d\n", iface);
+		ret = false;
+		goto out;
+	}
+out:
```
shijin-aws, line 179: do you really need an out here? shouldn't it just need `return ret`?

prov/efa/src/efa_hmem.c
```diff
@@ -78,6 +80,106 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
...
+ * @return	true if P2P is supported, otherwise false
+ */
+static bool efa_domain_hmem_support_p2p(struct efa_domain *efa_domain, enum fi_hmem_iface iface)
+{
+	bool ret = true;
+	char p2p_prov[P2P_PROV_MAX_LEN];
+	const char *p2p_prov_prefix;
+
+	if (iface == FI_HMEM_SYSTEM || iface == FI_HMEM_SYNAPSEAI) {
+		goto out;
```
shijin-aws, line 159: just return `true`

prov/efa/src/efa_hmem.c
```diff
@@ -109,30 +212,23 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
 {
 #if HAVE_CUDA
```
shijin-aws, line 213: Do you still need HAVE_CUDA ? Same for HAVE_NEURON. I don't see it uses any cuda API symbols anymore.

prov/efa/src/efa_hmem.c
```diff
@@ -320,7 +320,8 @@ static int efa_domain_hmem_info_init_synapseai(struct efa_domain *efa_domain)
 	info->p2p_disabled_by_user = false;
 	/* SynapseAI currently requires P2P */
 	info->p2p_required_by_impl = true;
-	info->p2p_supported_by_device = true;
+	info->p2p_supported_by_device = efa_domain_hmem_support_p2p(efa_domain, FI_HMEM_SYNAPSEAI);
+	info->dmabuf_support_status = EFA_DMABUF_SUPPORTED;
```
shijin-aws, line 324: I think if you have another function for `efa_domain_hmem_support_dmabuf` and make iface as an input parameter, we can get rid of these separate `efa_domain_hmem_info_init_<iface>` functions and just call one function with different ifaces in a loop

prov/efa/src/efa_hmem.c
```diff
@@ -206,88 +261,33 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
-	if (ret) {
-		EFA_WARN(FI_LOG_DOMAIN,
-			 "Failed to deregister Neuron buffer: %s\n",
-			 fi_strerror(-ret));
-		return ret;
-	}
-
-	info->p2p_supported_by_device = true;
 	efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_NEURON);
 	if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
```
shijin-aws, line 285: These warning can be moved to `efa_domain_hmem_info_init_protocol_thresholds` calls

prov/efa/src/efa_mr.c
```diff
@@ -511,76 +521,67 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
-		int ret;
-
-		ret = synapseai_get_dmabuf_fd(mr_attr->mr_iov->iov_base,
-						(uint64_t) mr_attr->mr_iov->iov_len,
-						&dmabuf_fd, &offset);
 		if (ret != FI_SUCCESS) {
-			EFA_WARN(FI_LOG_MR, "Unable to get dmabuf fd for Gaudi device buffer \n");
-			return NULL;
+			EFA_WARN(FI_LOG_MR,
+				 "Unable to get dmabuf fd for device buffer. "
```
shijin-aws, line 554: Add `iface`'s string  in the warning message

prov/efa/src/efa_mr.c
```diff
@@ -511,76 +521,67 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+
+	if (flags & FI_MR_DMABUF || *dmabuf_support == EFA_DMABUF_SUPPORTED) {
+		goto out;
 	}
-#endif
 
-	return ibv_reg_mr(efa_mr->domain->ibv_pd,
-			(void *)mr_attr->mr_iov->iov_base,
+	EFA_INFO(FI_LOG_MR, "Disable dmabuf support for HMEM interface.\n");
+no_dmabuf:
```
shijin-aws, line 579: These goto are confusing, I think just doing them explicitly in the if blocks are better. See comments above.

prov/efa/src/efa_mr.c
```diff
@@ -511,76 +521,67 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+	int ret, dmabuf_fd;
+	struct ibv_mr *mr = NULL;
+	uint64_t iova, offset;
+	size_t len;
+	enum efa_dmabuf_support_status *dmabuf_support;
+
+	dmabuf_support = &(efa_mr->domain->hmem_info[mr_attr->iface].dmabuf_support_status);
+
+	if (flags & FI_MR_DMABUF) {
+		goto dmabuf;
```
shijin-aws, line 533: This should just return the ibv_reg_dmabuf_mr call, we don't need any fallback here.

prov/efa/src/efa_mr.c
```diff
@@ -511,76 +521,67 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+	}
+
+	if (*dmabuf_support == EFA_DMABUF_NOT_SUPPORTED) {
+		goto no_dmabuf;
```
shijin-aws, line 537: just return `ibv_reg_mr` call, no fallback needed here.

prov/efa/src/efa_mr.c
```diff
@@ -511,76 +521,67 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
-		}
-		return NULL;
+	EFA_INFO(FI_LOG_MR,
+		 "Registering dmabuf mr with fd: %d, offset: %lu, len: %zu\n",
+		 dmabuf_fd, offset, len);
+
+	mr = efa_mr_reg_ibv_dmabuf_mr(efa_mr->domain->ibv_pd, offset, len, iova,
+				      dmabuf_fd, access);
+
+fallback:
```
shijin-aws, line 568: We need some more discussion here. I am not sure this fallback logic is the best practice.

prov/efa/src/efa_hmem.c
wenduwan: Cool please see update.

prov/efa/src/efa_hmem.c
wenduwan: Updated

prov/efa/src/efa_hmem.c
wenduwan: Good point. I prefer to do that separately. It needs some nice print for different iface. I need to change common code for that.

prov/efa/src/efa_hmem.c
wenduwan: Good call out. Updated.

prov/efa/src/efa_mr.c
wenduwan: Updated.

prov/efa/src/efa_mr.c
wenduwan: Yeah let's decide on the fallback logic first then I can simplify the logic here.

prov/efa/src/efa_mr.c
wenduwan: Yep let's have a discussion.

prov/efa/src/efa_hmem.c
wenduwan: Good callout. We can get rid of the guards after this change. Please see update.

prov/efa/src/efa_hmem.c
wenduwan: I figured there is `fi_tostr` util. Refactored this part in update.

prov/efa/src/efa_mr.c
wenduwan: For CUDA we are not sure about the correct fallback behavior if dmabuf fd/mr fails. Therefore I removed the fallback logic for CUDA in update. The application must request FI_MR_DMABUF with a valid fd in order to use dmabuf MR.

prov/efa/src/efa_mr.c
wenduwan: Aggregate response to the comments around the fallback logic.

I have experimented with a couple different implementations, trying to reduce `goto` statements and improve the readability. But after all I figured this logic is the most intuitive(pseudo code):
```
efa_mr_reg_ibv_mr(...) {
    if (require_dmabuf) {
        goto dmabuf;
    }
    if (no_dmabuf_support) {
        goto no_dmabuf;
    }
dmabuf:
    // Try dmabuf fd/mr
update_dmabuf_support:
    // Depending on whether dmabuf succeeded, update the support status
no_dmabuf:
    // Fallback ibv_reg_mr if allowed. No need to update dmabuf support in this case.
}
```

shijin-aws: The problem with your pseudo code is that the different blocks doesn't have unique dependency relationship. For example,  for synapseai and FI_MR_DMABUF, I don't think we have to reach `update_dmabuf_support` and `no_dmabuf` blocks no matter whether dmabuf block succeeds or not. That ends up with you have conditional return between `dmabuf` and subsequent blocks which is more confusing than making explicit code sections for different hmem ifaces.

shijin-aws: Thinking about it again, it may not be that bad, I will review the latest change before making more comments here

wenduwan: Refactored the logic after offline discussion. We will keep the current MR logic unchanged.

prov/efa/src/efa_env.c
```diff
@@ -37,6 +37,7 @@ struct efa_env efa_env = {
 	.host_id_file = "/sys/devices/virtual/dmi/id/board_asset_tag", /* Available on EC2 instances and containers */
 	.use_sm2 = false,
 	.huge_page_setting = EFA_ENV_HUGE_PAGE_UNSPEC,
+	.p2p_file_suffix= "/device/p2p",
```
shijin-aws, line 40: Did you update the man page to add this env?

prov/efa/src/efa_hmem.c
```diff
@@ -79,266 +107,162 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
...
+			info->p2p_required_by_impl = !hmem_ops[iface].initialized;
+		else
+			info->p2p_required_by_impl = true;
+		break;
+	case FI_HMEM_NEURON:
+		info->p2p_required_by_impl = true;
+		break;
+	case FI_HMEM_SYNAPSEAI:
+		if (!(g_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
+			EFA_WARN(FI_LOG_DOMAIN,
```
shijin-aws, line 250: I think we can remove this check now, because we already check this during FI_HMEM check in fi_getinfo, see https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_user_info.c#L312-L319 . If rdma-read is not enabled, we won't support FI_HMEM unless user request libfabric 1.18 API and it has cuda interface initialized

prov/efa/src/efa_hmem.c
```diff
@@ -79,266 +107,162 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
...
+ * @brief Initialize the efa_hmem_info state for iface
  *
  * @param[in,out]  efa_domain  Pointer to struct efa_domain
+ * @param[in]      iface       HMEM interface
  *
  * @return         0
  */
-static int efa_domain_hmem_info_init_synapseai(struct efa_domain *efa_domain)
+static int
+efa_domain_hmem_info_init_iface(struct efa_domain *efa_domain, enum fi_hmem_iface iface)
```
shijin-aws, line 216: Does this function return any non-zero? I haven't see it. If not, can we make it as void?

prov/efa/src/efa_hmem.c
```diff
@@ -407,44 +331,27 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
...
-
-	err = efa_domain_hmem_info_init_synapseai(efa_domain);
-	if (err) {
-		ret = err;
-		EFA_WARN(FI_LOG_DOMAIN, "Failed to populate the Synapseai hmem_info struct! err: %d\n",
-			 err);
+	for (int i = 0; i < sizeof(ifaces) / sizeof(enum fi_hmem_iface); ++i) {
+		ret = efa_domain_hmem_info_init_iface(efa_domain, ifaces[i]);
+		if (ret) {
+			EFA_WARN(FI_LOG_DOMAIN,
```
shijin-aws, line 349: similar comment as above, if `efa_domain_hmem_info_init_iface` doesn't return non-zero, we don't have to check return status.

shijin-aws (commented):
LGTM, some minor comments

prov/efa/src/efa_hmem.c
wenduwan: Makes sense. Will update.

prov/efa/src/efa_hmem.c
wenduwan: Makes sense. Will update.

prov/efa/src/efa_hmem.c
wenduwan: Addressed

prov/efa/src/efa_mr.c
```diff
@@ -499,88 +499,72 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+		         "Unable to get dmabuf fd for device buffer. "
+		         "iface: %s errno: %d, err_msg: %s\n",
+		         fi_tostr(&mr_attr->iface, FI_TYPE_HMEM_IFACE), ret,
+		         fi_strerror(-ret));
+		if (ret == -FI_EOPNOTSUPP && efa_mr_is_neuron(efa_mr)) {
+			return ibv_reg_mr(efa_mr->domain->ibv_pd,
+			                  (void *) mr_attr->mr_iov->iov_base,
+			                  mr_attr->mr_iov->iov_len, access);
 		}
 		return NULL;
```
jiaxiyan, line 557: Do we want a EFA_WARN here for failed Gaudi device mr?

wenduwan: The failure will be logged by the calling function such as `efa_mr_reg_impl` so we are covered.

shijin-aws (approved):

jiaxiyan (approved):

shijin-aws:
Pure efa provider change, Ignore Intel CI failure

PR #10130, opened 2024-06-26
prov/efa: Adjust the initial batch of internal rx pkt post

shijin-aws:
Move them from ep enable to the 1st fi_cq_read. It is observed that posting the initial 8192 rx pkts during ep enable will hurt performance if application open
some idle endpoints and never poll completions for them. Move these initial posts to the first cq read call before having a a long term fix.


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -502,11 +502,29 @@ static void efa_rdm_cq_progress(struct util_cq *cq)
...
+	 * some idle endpoints and never poll completions for them. Move these initial posts to
+	 * the first cq read call before having a long term fix.
+	 */
+	if (!efa_rdm_cq->initial_rx_to_all_eps_posted) {
+		dlist_foreach(&cq->ep_list, item) {
+			fid_entry = container_of(item, struct fid_list_entry, entry);
+			efa_rdm_ep = container_of(fid_entry->fid, struct efa_rdm_ep, base_ep.util_ep.ep_fid.fid);
+			efa_rdm_ep_post_internal_rx_pkts(efa_rdm_ep);
+		}
+		efa_rdm_cq->initial_rx_to_all_eps_posted = false;
```
darrylabbate, line 525: Supposed to be `true`?

shijin-aws: Sigh... I was modifying it back and forth and caused it. Good catch, I will add a test to cover it too

shijin-aws: Fixed, also added unit tests.

darrylabbate (approved):

PR #10139, opened 2024-07-01
prov/util, efa: discard outstanding operations in util_srx_close

shijin-aws:
Currently, util_srx_close call util_cancel_recv for the outstanding
untagged and tagged recv posts, which doesn't comply with the man
page for ep close that outstanding operations posted to the endpoint
when fi_close is called will be discarded.
Discarded operations will silently be dropped, with no completions reported.


shijin-aws:
After reading man page, I think even this fix is not correct

https://ofiwg.github.io/libfabric/v1.21.0/man/fi_endpoint.3.html

>Outstanding operations posted to the endpoint when fi_close is called will be discarded. Discarded operations will silently be dropped, with no completions reported. Additionally, a provider may discard previously completed operations from the associated completion queue(s). The behavior to discard completed operations is provider specific.

Shouldn't we just discard instead of canceling (generating cq err) here?

aingerson:
@shijin-aws I think you're right, we should just be able to discard it. We probably don't even need to iterate over the queues - as long as we free the memory in the bufpool itself, we can just ignore the dlist cleanup.
However, we might have a slight problem with the wording of the man page and the util_srx/peer api implementation:
```
Canceling an operation causes the fabric provider to search for the operation and, if it is still pending, complete it as having been canceled. An error queue entry will be available in the associated error queue with error code FI_ECANCELED. On the other hand, if the operation completed before the call to fi_cancel, then the completion status of that operation will be available in the associated completion queue.
```
The man page is a little unclear at what exactly "pending" means and suggest that there are two states of a receive - "pending" and "completed" which gets a little fuzzy with the asynchronous peer provider flow. Since the peer provider srx doesn't have access to the rx entry when the peer is processing the incoming command, there is a gray area here. It can't find a match so we assume it can't be canceled, but hasn't exactly "completed" yet and might not be available in the cq like the man page suggests.

We need to either add some language here to specify that the entry might not be completed but unable to be canceled OR add tracking in the peer srx implementation to keep a list of entries that are in peer space to be able to search that list as well. I think I fall more into the second category. Shouldn't be too hard - an extra slist and search and then call the srx->discard() function should do it. But we may need to add a specification that the owner can update the flags during the discard to specify that it should get written to the CQ with that flag. Could get a little spicy.

@j-xiong Do you have thoughts on this? Do you think we can treat an entry that has moved into peer space for processing as not being able to be canceled yet? Or do you think we should fix the util srx implementation to be able to update an entry that has moved to peer space?

j-xiong:
When the rx entry has been moved to the peer for processing, we could consider the rx entry being consumed and not being able to be cancelled any more.

shijin-aws:
>I think you're right, we should just be able to discard it. We probably don't even need to iterate over the queues - as long as we free the memory in the bufpool itself, we can just ignore the dlist cleanup.

But if your rx pool is not created with `OFI_BUFPOOL_NO_TRACK`, you will hit assertion error (use_cnt) during bufpool destroy. So explicity ofi_buf_free is still needed here.

aingerson:
@shijin-aws Yeah you'll still need the buf_free

shijin-aws:
Ideally it should be tested by a general fabtests, but for the time being I just added a unit test for EFA provider. @aingerson could you review the first commit, thx!

j-xiong:
how about `util_cleanup_queues` called a few lines above (L928-L929)?

shijin-aws:
> how about `util_cleanup_queues` called a few lines above (L928-L929)?

Good catch, will update

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -915,3 +915,25 @@ void test_efa_rdm_ep_user_zcpy_rx_unhappy_due_to_sas(struct efa_resource **state
...
+void test_efa_rdm_ep_close_discard_posted_recv(struct efa_resource **state)
+{
+	struct efa_resource *resource = *state;
+	char buf[16];
+	int err;
+
+	efa_unit_test_resource_construct(resource, FI_EP_RDM);
+
+	/* Post recv and then close ep */
+	fi_recv(resource->ep, (void *) buf, 16, NULL, FI_ADDR_UNSPEC, NULL);
```
aingerson, line 928: I don't know if we scan the efa tests for coverity but if we do I think this might get picked up - it doesn't like ignoring the return codes (here and on fi_close before). Technically should check this and only continue on FI_SUCCESS

aingerson (commented):
Just a small comment, otherwise looks good!

prov/efa/test/efa_unit_test_ep.c
shijin-aws: Thanks, updated

shijin-aws:
Is Intel CI failure related?

shijin-aws:
@j-xiong do you think the error can be related?

j-xiong:
Probably unrelated. The last push was efa test changes plus rebase and CI did pass for the push before that. I restarted the CI run anyway.

j-xiong (approved):

PR #10144, opened 2024-07-02
prov/efa: Post rx buffer in the first fi_cntr_read

shijin-aws:
It may be possible that application (currently fabtests)
only use fi_cntr_read to poll completions without calling fi_cq_read. In this case, the first fi_cntr_read should post rx buffers to all bounded eps as well.


shijin-aws:
related to https://github.com/ofiwg/libfabric/issues/10145

darrylabbate (approved):

shijin-aws:
bot:aws:retest

PR #10146, opened 2024-07-02
prov/efa: Create unit tests to validate efa provider meet the LL128 protocol requirement 

jiaxiyan:
Add the following unit tests:
1. Test the packet allocated from read_copy_pkt_pool is 128 byte aligned.
2. Test when using in order aligned send/recv, the copy method is always rdma read.
3. The data sent by runting read protocol should always have a size as 128 multiple. NCCL will always send a 128 multiple size for LL128 protocol, but runting read will only send a segment of the whole message via ibv_send. The test makes sure such segmented size must be 128 multiple.


prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -229,23 +228,62 @@ int efa_rdm_pke_queued_copy_payload_to_hmem(struct efa_rdm_pke *pke,
...
  *
- * localread copy, which is available only when p2p is supported by device, and device support read.
+ * @param[in]			ep, desc
+ * @param[in,out]		local_read_available, cuda_memcpy_available,
+ * gdrcopy_available
+ * @return		On success, return 0
+ * 				On failure, return libfabric error code
+ */
+int efa_rdm_pke_get_available_copy_methods(struct efa_rdm_ep *ep,
+					   struct efa_mr *desc,
```
shijin-aws, line 252: nit: Just use `efa_mr` for the name. We will do a cast from `desc` to `efa_mr` when call this function

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -275,23 +313,13 @@ int efa_rdm_pke_copy_payload_to_cuda(struct efa_rdm_pke *pke,
...
-	gdrcopy_available = desc->peer.flags & OFI_HMEM_DATA_DEV_REG_HANDLE;
+	segment_offset = efa_rdm_pke_get_segment_offset(pke);
 
-	/* For in-order aligned send/recv, only allow local read to be used to copy data */
-	if (ep->sendrecv_in_order_aligned_128_bytes) {
-		cuda_memcpy_available = false;
-		gdrcopy_available = false;
+	ret = efa_rdm_pke_get_available_copy_methods(
+		ep, desc, &local_read_available, &gdrcopy_available, &cuda_memcpy_available);
+	if (ret < 0) {
```
shijin-aws, line 320: nit: add a EFA_WARN here

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -309,7 +310,71 @@ void test_efa_rdm_ep_pkt_pool_page_alignment(struct efa_resource **state)
...
+	
+	/* p2p is available */
+	efa_rdm_ep_domain(efa_rdm_ep)->hmem_info[FI_HMEM_CUDA].p2p_supported_by_device = true;
+	efa_rdm_ep->hmem_p2p_opt = FI_HMEM_P2P_ENABLED;
+
+	/* RDMA read is supported */
+	efa_rdm_ep->use_device_rdma = true;
+	efa_rdm_ep_domain(efa_rdm_ep)->device->device_caps |= EFADV_DEVICE_ATTR_CAPS_RDMA_READ;
+
+	int ret = efa_rdm_pke_get_available_copy_methods(
```
shijin-aws, line 369: Move `int ret` to the beginning of the function, and do the assignment here.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -309,7 +310,71 @@ void test_efa_rdm_ep_pkt_pool_page_alignment(struct efa_resource **state)
...
+
+	fi_close(&ep->fid);
+}
+
+/**
+ * @brief When using LL128 protocol, the copy method is local read. 
+ *
+ * @param[in]	state		struct efa_resource that is managed by the framework
+ */
+void test_efa_rdm_pke_get_available_copy_methods(struct efa_resource **state)
```
shijin-aws, line 348: This only test the available copy methods when LL128 is enabled. You need to change the function name to indicate that

prov/efa/test/efa_unit_test_runt.c
```diff
@@ -127,6 +127,53 @@ void test_efa_rdm_peer_get_runt_size_cuda_memory_normal(struct efa_resource **st
 	test_efa_rdm_peer_get_runt_size_impl(resource, FI_HMEM_CUDA, peer_num_runt_bytes_in_flight, total_runt_size, msg_length, expected_runt_size);
 }
 
+/* When using LL128 protocol, the segmented size of runting read must be 128 multiple. */
```
shijin-aws, line 130: do you need some more test cases when the available runting bytes is bigger than the total len?

shijin-aws: Also need a test case when the available runting bytes is smaller than 128.

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -229,23 +228,62 @@ int efa_rdm_pke_queued_copy_payload_to_hmem(struct efa_rdm_pke *pke,
...
+ * @param[in,out]		local_read_available, cuda_memcpy_available,
+ * gdrcopy_available
+ * @return		On success, return 0
+ * 				On failure, return libfabric error code
+ */
+int efa_rdm_pke_get_available_copy_methods(struct efa_rdm_ep *ep,
+					   struct efa_mr *desc,
+					   bool *local_read_available,
+					   bool *cuda_memcpy_available,
+					   bool *gdrcopy_available)
```
darrylabbate, line 255: We can `restrict`-qualify the parameters here to avoid any aliasing quirks

```suggestion
					   bool *restrict local_read_available,
					   bool *restrict cuda_memcpy_available,
					   bool *restrict gdrcopy_available)
```

jiaxiyan:
CI failure is real. I will post the fix next week.

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -310,6 +311,85 @@ void test_efa_rdm_ep_pkt_pool_page_alignment(struct efa_resource **state)
...
+	/* Override global attribute */
+	iface = FI_HMEM_CUDA;
+	iface_initialized = hmem_ops[iface].initialized;
+	hmem_ops[iface].initialized = true;
+
+	/* rx_readcopy_pkt_pool is only created when application requested FI_HMEM capability */
+	resource->hints = efa_unit_test_alloc_hints(FI_EP_RDM);
+	assert_non_null(resource->hints);
+	resource->hints->caps |= FI_HMEM;
+	resource->hints->domain_attr->mr_mode |= FI_MR_HMEM | FI_MR_ALLOCATED;
```
wenduwan, line 337: I realized that `fi_getinfo(...)` with FI_HMEM would fail on non-HMEM platforms, because non of CUDA/NEURON/SYNAPSEAI is available at runtime. You might try configure without `--with-cuda`

It would be easier to initialize resources without FI_HMEM, and modify the attributes directly, which is sad :<

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -310,6 +311,85 @@ void test_efa_rdm_ep_pkt_pool_page_alignment(struct efa_resource **state)
...
+	efa_unit_test_resource_construct_with_hints(
```
shijin-aws, line 338: This involves the logic of HMEM checking during fi_getinfo, which makes this test depend on some other things that beyond the buffer pool creation logic.

You can do it in the following way:
1. Just call efa_unit_test_resource_construct(), which gives you a set of resources (including domain)
2. Modify the mr mode in efa domain to make it has FI_MR_HMEM
3. Create a new efa ep via `fi_endpoint`
4. alloc pke from this new EP.
5. do the check ...

prov/efa/test/efa_unit_test_ep.c
shijin-aws: See my comment below, I think this test doesn't have to do such cuda/hmem hack

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -275,21 +313,13 @@ int efa_rdm_pke_copy_payload_to_cuda(struct efa_rdm_pke *pke,
...
+	segment_offset = efa_rdm_pke_get_segment_offset(pke);
 
-	/* For in-order aligned send/recv, only allow local read to be used to copy data */
-	if (ep->sendrecv_in_order_aligned_128_bytes) {
-		cuda_memcpy_available = false;
-		gdrcopy_available = false;
+	ret = efa_rdm_pke_get_available_copy_methods(
+		ep, desc, &local_read_available, &gdrcopy_available, &cuda_memcpy_available);
+	if (ret < 0) {
+		EFA_WARN(FI_LOG_CQ, "p2p is not supported by device\n");
```
shijin-aws, line 321: This error message isn't correct, p2p_support return < 0 doesn't mean p2p is not supported. It means that there is unexpected error happens during p2p check. For here, just log as `Failed to get available copy method, ret %d\n`. Change FI_LOG_CQ to FI_LOG_EP_DATA

shijin-aws (changes requested):

shijin-aws (approved):

PR #10147, opened 2024-07-03
fabtests: Add benchmark option for `FI_OPT_MAX_MSG_SIZE`

darrylabbate:
This adds an option (`-X`) for benchmark tests; corresponding to the endpoint options `FI_OPT_MAX_MSG_SIZE`. `fi_setopt()` is called during EP enable with the value requested by the user.

This also includes a few related minor changes to the EFA provider.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -1427,6 +1427,14 @@ int ft_enable_ep(struct fid_ep *bind_ep, struct fid_eq *bind_eq, struct fid_av *
 		FT_EP_BIND(bind_ep, bind_rma_cntr, flags);
 	}
 
+	if (opts.max_msg_size) {
+		ret = fi_setopt(&bind_ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE, &opts.max_msg_size, sizeof opts.max_msg_size);
+		if (ret && ret != -FI_EOPNOTSUPP) {
+			FT_PRINTERR("fi_setopt(FI_OPT_MAX_MSG_SIZE)", ret);
+			return ret;
+		}
+	}
+
 	ret = fi_enable(bind_ep);
 	if (ret) {
 		FT_PRINTERR("fi_enable", ret);
@@ -4169,6 +4177,8 @@ void ft_longopts_usage()
 		"manual, or auto");
 	FT_PRINT_OPTS_USAGE("--control-progress <progress_model>",
 		"manual, auto, or unified");
+	FT_PRINT_OPTS_USAGE("--max-msg-size <size>",
+		"maximum untagged message size");
 }
 
 int debug_assert;
@@ -4180,6 +4190,7 @@ struct option long_opts[] = {
 	{"debug-assert", no_argument, &debug_assert, LONG_OPT_DEBUG_ASSERT},
 	{"data-progress", required_argument, NULL, LONG_OPT_DATA_PROGRESS},
 	{"control-progress", required_argument, NULL, LONG_OPT_CONTROL_PROGRESS},
+	{"max-msg-size", required_argument, NULL, LONG_OPT_MAX_MSG_SIZE},
 	{NULL, 0, NULL, 0},
 };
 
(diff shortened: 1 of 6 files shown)


fabtests/benchmarks/benchmark_shared.c
```diff
@@ -79,6 +82,7 @@ void ft_benchmark_usage(void)
 	FT_PRINT_OPTS_USAGE("-v", "enables data_integrity checks");
 	FT_PRINT_OPTS_USAGE("-k", "force prefix mode");
 	FT_PRINT_OPTS_USAGE("-j", "maximum inject message size");
+	FT_PRINT_OPTS_USAGE("-X", "maximum untagged message size");
```
shijin-aws, line 85: I think using a general `-X` only for untagged message size is not desired. We can introduce a long option `--max-msg-size` for it.

fabtests/common/shared.c
```diff
@@ -1427,6 +1427,14 @@ int ft_enable_ep(struct fid_ep *bind_ep, struct fid_eq *bind_eq, struct fid_av *
 		FT_EP_BIND(bind_ep, bind_rma_cntr, flags);
 	}
 
+	if (opts.max_msg_size) {
+		ret = fi_setopt(&bind_ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE, &opts.max_msg_size, sizeof opts.max_msg_size);
+		if (ret) {
```
shijin-aws, line 1432: What will happen if provider doesn't support such option? We should ignore such ENOSUPPORT error

shijin-aws (changes requested):

shijin-aws:
Do you need another commit to add such test to pytest?

darrylabbate:
> Do you need another commit to add such test to pytest?

Are you asking for a default pytest? I figure we can have a follow-up PR for some kind of EFA zero-copy fabtest.

shijin-aws:
> Are you asking for a default pytest? I figure we can have a follow-up PR for some kind of EFA zero-copy fabtest.

I prefer to just have it in this PR so we can make it tested by AWS CI. It should be a straightforward incremental change, like what we have in `test_rdm.py`

fabtests/common/shared.c
```diff
@@ -1427,6 +1427,14 @@ int ft_enable_ep(struct fid_ep *bind_ep, struct fid_eq *bind_eq, struct fid_av *
 		FT_EP_BIND(bind_ep, bind_rma_cntr, flags);
 	}
 
+	if (opts.max_msg_size) {
+		ret = fi_setopt(&bind_ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE, &opts.max_msg_size, sizeof opts.max_msg_size);
+		if (ret && ret != -FI_ENOPROTOOPT) {
```
shijin-aws, line 1432: I think it's a bug in the man page. In this case we should return `-FI_EOPNOTSUPP`

shijin-aws: Nvm, I think FI_ENOPROTOOPT is OK and that's what efa and shm returns today

fabtests/pytest/efa/test_rdm.py
```diff
@@ -109,3 +109,8 @@ def test_rdm_pingpong_1G(cmdline_args, completion_semantic):
     efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong -W 1", 2,
                                completion_semantic=completion_semantic, message_size=1073741824,
                                memory_type="host_to_host", warmup_iteration_type=0)
+
+@pytest.mark.functional
```
shijin-aws, line 113: can you add rdm_bw as well?

shijin-aws (approved):

shijin-aws:
@j-xiong can you review https://github.com/ofiwg/libfabric/pull/10147/commits/fa178bb1fe878de3be948304e3504a9b8da92ef8

j-xiong (approved):

fabtests/pytest/efa/test_rdm.py
```diff
@@ -109,3 +109,13 @@ def test_rdm_pingpong_1G(cmdline_args, completion_semantic):
...
+def test_rdm_pingpong_zcpy_recv(cmdline_args, memory_type, message_size):
+    efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong --max-msg-size 8192",
+                               "short", "transmit_complete", memory_type, message_size)
+
+@pytest.mark.functional
+def test_rdm_bw_zcpy_recv(cmdline_args, memory_type, message_size):
```
shijin-aws, line 119: The `message_size` fixture will have msg size larger than 8192, you need to customize it to make it smaller than 8192

shijin-aws:
I opened a new PR that reworked the last commit of this PR https://github.com/ofiwg/libfabric/pull/10168

PR #10157, opened 2024-07-09
prov/verbs: Flush CQ for SQ on no SQ credit

chien-intel:
This fixes issue 10150.

On post send, when we are out of SQ credit, attempt to gain SQ credit by polling CQ for SQ.

Looks like a copy-paste mistake.

### DIFF
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -182,19 +182,21 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
 		goto unlock;
 	}
 
-	if (!ep->sq_credits || !ep->peer_rq_credits) {
-		cq = container_of(ep->util_ep.rx_cq, struct vrb_cq, util_cq);
+	if (!ep->sq_credits) {
+		cq = container_of(ep->util_ep.tx_cq, struct vrb_cq, util_cq);
 		vrb_flush_cq(cq);
 
-		if (!ep->sq_credits || !ep->peer_rq_credits)
+		if (!ep->sq_credits)
 			goto freectx;
 	}
 
-	if (vrb_wr_consumes_recv(wr) && !--ep->peer_rq_credits &&
-	    !(flags & OFI_PRIORITY)) {
+	if (vrb_wr_consumes_recv(wr)) {
+		if  (!ep->peer_rq_credits || 
+		     (ep->peer_rq_credits == 1 && !(flags & OFI_PRIORITY)))
 		/* Last credit is reserved for credit update */
-		ep->peer_rq_credits++;
-		goto freectx;
+			goto freectx;
+
+		ep->peer_rq_credits--;
 	}
 
 	ep->sq_credits--;


j-xiong (approved):

prov/verbs/src/verbs_ep.c
```diff
@@ -183,7 +183,7 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
 	}
 
 	if (!ep->sq_credits || !ep->peer_rq_credits) {
-		cq = container_of(ep->util_ep.rx_cq, struct vrb_cq, util_cq);
+		cq = container_of(ep->util_ep.tx_cq, struct vrb_cq, util_cq);
```
shefty, line 186: I think the problem is slightly worse.  If sq_credits == 0, then we need to read transmit completions to free send credits.  If peer_rq_credits == 0, then we need to read receive completions to look for new credits sent by the peer.

Maybe the entire if statement needs to be split into 2?

chien-intel: >If peer_rq_credits == 0, then we need to read receive completions to look for new credits sent by the peer.
that's RXM though.  how would you do that without replicating RXM code from here?

shefty: Hmm... good point.  I think the way to fix this, so that the code reflects what's happening is:
```
if (!ep->peer_rq_credits)
    goto freectx;

if (!ep->sq_credits) {
    cq = ... tx_cq ...   // your fix
    vrb_flush_cq(cq)

    if (!ep->sq_credits)
        goto freectx;
}
```

chien-intel: yeah.  leave peer_rq_credits to normal progress call.

chien-intel: Side question, why does vrb_wr_consumes_recv macro include IBV_WR_RDMA_WRITE_WITH_IMM?  write with immediate should not consume a RX buffer, immediate data shows up in CQ not RX buffer.  What am I missing?

j-xiong: Verbs write with immediate consumes RX buffer. It's the OFI write w/imm that doesn't.

prov/verbs/src/verbs_ep.c
```diff
@@ -176,27 +176,26 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
...
-	if (vrb_wr_consumes_recv(wr) && !--ep->peer_rq_credits &&
-	    !(flags & OFI_PRIORITY)) {
-		/* Last credit is reserved for credit update */
-		ep->peer_rq_credits++;
-		goto freectx;
+	if (vrb_wr_consumes_recv(wr)) {
+		if ((ep->peer_rq_credits == 1) && !(flags & OFI_PRIORITY))
+			goto unlock;
+
+		ep->peer_rq_credits--;
```
shefty, line 192: This flow isn't quite right.  peer_rq_credits could be 0 on entry into this call (because we used the last credit to send a flow control message).  The results of the above flow would result in value wrapping.  This check should work:
```
if (!ep->peer_rq_credits || ((ep->peer_rq_credits == 1) && !(flags & OFI_PRIORITY)))
    goto unlock;
```

prov/verbs/src/verbs_ep.c
```diff
@@ -176,27 +176,26 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
...
 	}
 
+	ctx = vrb_alloc_ctx(vrb_ep2_progress(ep));
+	if (!ctx)
+		goto unlock;
+
```
j-xiong, line 198: should restore `peer_rq_credits` if it was decreased.

chien-intel: thanks.  good catch.

j-xiong (approved):

prov/verbs/src/verbs_ep.c
```diff
@@ -173,30 +173,32 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
 	struct ibv_send_wr *bad_wr;
 	struct vrb_cq *cq;
 	size_t credits_to_give = 0;
+	bool dec_peer_rq_credit = false;
```
shefty, line 176: This boolean was introduced as a result of moving the vrb_alloc_ctx() call.  If we assume that the common case is that we're usually able to send, we've added this assignment, possibly a second assignment to set it to true, and another if statement to the fast path.

The boolean has slight side effect on reading the code in the error case.  See the credit label.  There peer_rq_credits is updated based on vrb_wr_consumes_recv(), and not this variable, which throws off the code symmetry.

Basically, reverting the vrb_alloc_ctx() change should remove the need for this variable.

chien-intel: I see your point but there is assumption once there is no credit there won't be anymore calls to post_send?  In a heavily loaded system, once we are out of credit and returning -EAGAIN, wouldn't the caller try again and we are in alloc ctx and free ctx error path?

j-xiong: agree, we should prioritize the success pass.

j-xiong: to clarify, I agree with @shefty.

Having alloc_ctx and free_ctx in the error cases should be fine. The allocation is from a pre-allocated queue, not directly from OS.

shefty: If we return EAGAIN, the caller should try to progress the CQ first before trying to send again.  But even if the caller is spinning between reading the cq and trying to send, there's no harm in allocating and freeing the ctx.  The app is burning CPU cycles doing nothing.

chien-intel: Dropping the second commit.  Need to get this fix in.

j-xiong: How about adding the following line to the first commit as previously suggested by @shefty?
```
if (!ep->peer_rq_credits)
    goto freectx;
```

j-xiong: anyway, this is not important and doesn't really achieve anything more than saving a flush. This needs to be revisited later. Let's keep it as it is.

shefty: I don't completely agree.  Separating the peer_rq_credits should be part of this change.  The code flow through this function does not change peer_rq_credits -- and that's important to capture.  The reason is exactly why the bug being addressed occurred in the first place.  The wrong CQ was being read because peer_rq_credits is updated through the rx cq, not the tx cq.  But it's only updated through rxm.  I even missed this on the initial review.

The code strongly suggests that peer_rq_credits *might* be updated.  With the updated code, the right CQ is being read, but the resulting code still *looks* wrong.  It looks like the rx cq should be checked, but it's is not.  That's why peer_rq_credits should be removed entirely from the if statement.

chien-intel: This will allow post send that do not require RX buffer to go through.

prov/verbs/src/verbs_ep.c
```diff
@@ -182,19 +182,21 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
...
 
-	if (vrb_wr_consumes_recv(wr) && !--ep->peer_rq_credits &&
-	    !(flags & OFI_PRIORITY)) {
+	if (vrb_wr_consumes_recv(wr)) {
+		if  (!ep->peer_rq_credits || 
+		     (ep->peer_rq_credits == 1 && !(flags & OFI_PRIORITY)))
 		/* Last credit is reserved for credit update */
+			goto freectx;
+
 		ep->peer_rq_credits++;
```
shefty, line 199: Needs to change to -- now.  Otherwise, change looks good to me.

chien-intel: yeah...

shefty (approved):
thanks!

PR #10161, opened 2024-07-09
prov/sockets: Fix Coverity issue on integer overflow.

peinanz:
Fix coverity issues: CID 426839, 426829, 426817, 426813


prov/sockets/src/sock_ep_msg.c
```diff
@@ -183,36 +183,37 @@ static int sock_ep_cm_getpeer(struct fid_ep *ep, void *addr, size_t *addrlen)
...
-		if (ret < 0) {
-			if (OFI_SOCK_TRY_SND_RCV_AGAIN(ofi_sockerr()))
-				continue;
+		if (ret >= 0) {
+			done += ret;
+		} else if (!OFI_SOCK_TRY_SND_RCV_AGAIN(ofi_sockerr())) {
 			SOCK_LOG_ERROR("failed to write to fd: %s\n",
 				       strerror(ofi_sockerr()));
 			return -FI_EIO;
 		}
```
j-xiong, line 198: The convention here is to handle error in branches and let the normal path fall through. Can we keep the original flow?

prov/sockets/src/sock_ep_msg.c
```diff
@@ -183,36 +183,37 @@ static int sock_ep_cm_getpeer(struct fid_ep *ep, void *addr, size_t *addrlen)
...
-		if (ret <= 0) {
-			if (ret < 0 && (errno == EAGAIN || errno == EWOULDBLOCK))
-				continue;
+		if (ret > 0) {
+			done += ret;
+		} else if (!ret || (errno != EAGAIN && errno != EWOULDBLOCK)) {
 			SOCK_LOG_ERROR("failed to read from fd: %s\n",
 				       strerror(ofi_sockerr()));
 			return -FI_EIO;
 		}
```
j-xiong, line 216: same here.

prov/sockets/src/sock_ep_msg.c
peinanz: Ok, will revert the changes.

j-xiong (approved):

PR #10168, opened 2024-07-10
fabtests: Add benchmark option for FI_OPT_MAX_MSG_SIZE

shijin-aws:
Also added efa fabtests cases that uses this option

### DIFF
--- a/fabtests/pytest/efa/test_rdm.py
+++ b/fabtests/pytest/efa/test_rdm.py
@@ -109,3 +109,17 @@ def test_rdm_pingpong_1G(cmdline_args, completion_semantic):
     efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong -W 1", 2,
                                completion_semantic=completion_semantic, message_size=1073741824,
                                memory_type="host_to_host", warmup_iteration_type=0)
+
+@pytest.mark.functional
+def test_rdm_pingpong_zcpy_recv(cmdline_args, memory_type, zcpy_recv_max_msg_size, zcpy_recv_message_size):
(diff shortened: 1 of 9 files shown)


shijin-aws:
Reworked the last commit of https://github.com/ofiwg/libfabric/pull/10147

shijin-aws:
I don't think Intel CI is related, but would like to check @j-xiong

fabtests/pytest/efa/test_rdm.py
```diff
@@ -109,3 +109,13 @@ def test_rdm_pingpong_1G(cmdline_args, completion_semantic):
     efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong -W 1", 2,
                                completion_semantic=completion_semantic, message_size=1073741824,
                                memory_type="host_to_host", warmup_iteration_type=0)
+
+@pytest.mark.functional
+def test_rdm_pingpong_zcpy_recv(cmdline_args, memory_type, zcpy_recv_message_size):
+    efa_run_client_server_test(cmdline_args, "fi_rdm_pingpong --max-msg-size 8192",
```
wenduwan, line 115: Can we define `max_msg_size` as a fixture? IIUC later we will also have `max_rma_msg_size`, etc. It would be cleaner to have those fixtures in conftest.py side b side.

shijin-aws: Updated

j-xiong:
Intel CI has intermittent failures with impi tcp tests. Unrelated to this PR.

wenduwan (approved):

PR #10169, opened 2024-07-11
libfabric/core: Fix coverity issues 426807 and 426820

Juee14Desai:
This commit fixes integer overflow issues detected by coverity.

### DIFF
--- a/include/ofi_rbuf.h
+++ b/include/ofi_rbuf.h
@@ -116,6 +116,8 @@ struct ofi_ringbuf {
 
 static inline int ofi_rbinit(struct ofi_ringbuf *rb, size_t size)
 {
+	if (size == 0 || size > (SIZE_MAX / 2))
+		return -EINVAL;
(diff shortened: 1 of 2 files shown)


include/ofi_rbuf.h
```diff
@@ -116,7 +116,11 @@ struct ofi_ringbuf {
 
 static inline int ofi_rbinit(struct ofi_ringbuf *rb, size_t size)
 {
+	if (size == 0 || size > (SIZE_MAX / 2))
+		return -EINVAL;
 	rb->size = roundup_power_of_two(size);
+	if (rb->size == 0)
+		return -EINVAL;
```
j-xiong, line 123: This check seems to be redundant.

Juee14Desai: Ok, thanks for catching that. Removed from recent commit.

j-xiong (approved):

PR #10170, opened 2024-07-11
prov/shm: coverity fixes

aingerson:


j-xiong (commented):
Please check the places that use the value of `resp->status`. At least in `smr_progress_resp`, `-resp->status` is used to report the error number, which is expecting a positive value.  That need to be changed as well.

aingerson:
@j-xiong Good catch! I double checked all the usages of resp->status including where it is passed in and that was the only one. Changed it to pass in a positive. Thanks!

j-xiong (approved):

PR #10174, opened 2024-07-11
prov/tcp: optimize FI_THREAD_COMPLETION

ooststep:
In many multi-threaded workloads it's common for the application to manage many eps, each on their own threads, but under a single domain while setting their threading semantics as `FI_THREAD_COMPLETION`.  The tcp provider has been optimized for single threaded processes and doesn't do anything special for this use case.

Ideally, these applications would restructure their objects to follow `FI_THREAD_DOMAIN` semantics, with a domain per thread.

This set of changes introduces a path to handle `FI_THREAD_COMPLETION` semantics by multiplexing the application's domain over many 'subdomain' - one per ep - which each follow FI_THREAD_DOMAIN semantics.  Associated AVs and MRs will also be multiplexed.  CQs and CNTRs will be migrated to the appropriate subdomains.

### DIFF
--- a/include/ofi_mem.h
+++ b/include/ofi_mem.h
@@ -574,6 +574,33 @@ static inline void *ofi_ibuf_alloc(struct ofi_bufpool *pool)
 	return ofi_buf_data(buf_hdr);
 }
 
+static inline void *ofi_ibuf_alloc_at(struct ofi_bufpool *pool, size_t index)
+{
+	void *buf;
+	struct ofi_bufpool_hdr *buf_hdr;
+	struct ofi_bufpool_region *buf_region;
+	size_t region_index = index / pool->attr.chunk_cnt;
+
+	assert(pool->attr.flags & OFI_BUFPOOL_INDEXED);
+	while (region_index >= pool->region_cnt) {
+		if (ofi_bufpool_grow(pool))
+			return NULL;
+	}
+	buf_region = pool->region_table[region_index];
+	buf = buf_region->mem_region +
+		(index % pool->attr.chunk_cnt) * pool->entry_size;
+	buf_hdr = ofi_buf_hdr(buf);
+	assert(ofi_atomic_inc32(&buf_hdr->region->use_cnt));
+	assert(!ofi_buf_is_valid(buf));
+
+	dlist_remove(&buf_hdr->entry.dlist);
+	buf_hdr->entry.dlist.next = &buf_hdr->entry.dlist;
+
+	if (dlist_empty(&buf_region->free_list))
+		dlist_remove_init(&buf_region->entry);
+
+	return buf;
+}
 
 /*
  * Persistent memory support
--- a/include/ofi_mem.h
+++ b/include/ofi_mem.h
@@ -383,7 +383,6 @@ struct ofi_bufpool_hdr {
 	struct ofi_bufpool_region	*region;
 	size_t 				index;
 
-	OFI_DBG_VAR(bool, allocated)
 	OFI_DBG_VAR(struct ofi_bufpool_ftr *, ftr)
 	OFI_DBG_VAR(size_t,		magic)
 };
@@ -433,15 +432,32 @@ static inline struct ofi_bufpool *ofi_buf_pool(void *buf)
 	return ofi_buf_region(buf)->pool;
 }
 
+static inline bool ofi_buf_is_valid(void *buf)
+{
+	struct ofi_bufpool_hdr *buf_hdr = ofi_buf_hdr(buf);
+	return buf_hdr->entry.slist.next == &buf_hdr->entry.slist;
+}
+
+static inline bool ofi_bufpool_ibuf_is_valid(struct ofi_bufpool *pool, size_t index)
+{
+	void *buf;
+	size_t region_index = index / pool->attr.chunk_cnt;
+
+	assert(region_index < pool->region_cnt);
+
+	buf = pool->region_table[region_index]->mem_region +
+		(index % pool->attr.chunk_cnt) * pool->entry_size;
+
+	return ofi_buf_is_valid(buf);
+}
+
 static inline void ofi_buf_free(void *buf)
 {
 	assert(ofi_atomic_dec32(&ofi_buf_region(buf)->use_cnt) >= 0);
 	assert(!(ofi_buf_pool(buf)->attr.flags & OFI_BUFPOOL_INDEXED));
 	assert(ofi_buf_hdr(buf)->magic == OFI_MAGIC_SIZE_T);
 	assert(ofi_buf_hdr(buf)->ftr->magic == OFI_MAGIC_SIZE_T);
-	assert(ofi_buf_hdr(buf)->allocated == true);
-
-	OFI_DBG_SET(ofi_buf_hdr(buf)->allocated, false);
+	assert(ofi_buf_is_valid(buf));
 
 	slist_insert_head(&ofi_buf_hdr(buf)->entry.slist,
 			  &ofi_buf_pool(buf)->free_list.entries);
@@ -460,9 +476,7 @@ static inline void ofi_ibuf_free(void *buf)
 	assert(ofi_buf_pool(buf)->attr.flags & OFI_BUFPOOL_INDEXED);
 	assert(buf_hdr->magic == OFI_MAGIC_SIZE_T);
 	assert(buf_hdr->ftr->magic == OFI_MAGIC_SIZE_T);
-	assert(buf_hdr->allocated == true);
-
-	OFI_DBG_SET(buf_hdr->allocated, false);
+	assert(ofi_buf_is_valid(buf));
 
 	dlist_insert_order(&buf_hdr->region->free_list,
 			   ofi_ibuf_is_lower, &buf_hdr->entry.dlist);
@@ -487,8 +501,7 @@ static inline void *ofi_bufpool_get_ibuf(struct ofi_bufpool *pool, size_t index)
 
 	buf = pool->region_table[region_index]->mem_region +
 		(index % pool->attr.chunk_cnt) * pool->entry_size;
-
-	assert(ofi_buf_hdr(buf)->allocated);
+	assert(ofi_buf_is_valid(buf));
 
 	return buf;
 }
@@ -516,9 +529,9 @@ static inline void *ofi_buf_alloc(struct ofi_bufpool *pool)
 	slist_remove_head_container(&pool->free_list.entries,
 				struct ofi_bufpool_hdr, buf_hdr, entry.slist);
 	assert(ofi_atomic_inc32(&buf_hdr->region->use_cnt));
-	assert(buf_hdr->allocated == false);
+	assert(!ofi_buf_is_valid(ofi_buf_data(buf_hdr)));
 
-	OFI_DBG_SET(buf_hdr->allocated, true);
+	buf_hdr->entry.slist.next = &buf_hdr->entry.slist;
 
 	return ofi_buf_data(buf_hdr);
 }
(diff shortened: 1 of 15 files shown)


prov/tcp/src/xnet_rdm.c
```diff
@@ -695,13 +695,183 @@ static struct fi_ops_ep xnet_rdm_ep_ops = {
...
+				       &subdomain_fid->fid);
+		if (ret)
+			fi_close(&subdomain_fid->fid);
+
+		set_subdomain(rdm, domain, subdomain);
+
+		ofi_rbmap_foreach(domain->util_domain.mr_map.rbtree,
+				  domain->util_domain.mr_map.rbtree->root,
+				  xnet_reg_subdomain_mr, subdomain);
+	}
```
j-xiong, line 846: If I understand correctly, `find_subdomain()` handles the scenario that the cq(s)/cntr(s) bound to this rdm ep may have previously been bound to another rdm ep that has already been enabled (and thus already got a subdomain assigned), in which case the two rdm eps will share the same subdomain since they belong to the same completion group. However, there is the possibility that some of cq(s)/cntr(s) bound to this rdm ep may still be associated with the multiplex domain. So `set_subdomain()` needs also be called from the `else` branch (which is not present here).

prov/tcp/src/xnet_domain.c
```diff
@@ -128,6 +171,81 @@ xnet_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+	struct fid_list_entry *item;
+	struct fid_mr *sub_mr_fid;
+	struct ofi_mr *mr;
+	int ret;
+
+	domain = container_of(fid, struct xnet_domain,
+			      util_domain.domain_fid.fid);
+	ret = ofi_mr_regattr(fid, attr, flags, mr_fid);
+	if (ret)
+		goto out;
```
shefty, line 189: nit: can just return here

prov/tcp/src/xnet_domain.c
```diff
@@ -167,9 +278,45 @@ xnet_query_atomic(struct fid_domain *domain, enum fi_datatype datatype,
...
+{
+	struct xnet_domain *domain;
+	struct fid_list_entry *item;
+
+	domain = container_of(fid, struct xnet_domain, util_domain.domain_fid.fid);
+	while (!dlist_empty(&domain->subdomain_list)) {
+		dlist_pop_front(&domain->subdomain_list, struct fid_list_entry,	item, entry);
+		(void)fi_close(item->fid);
+		free(item);
+	}
```
shefty, line 309: nit: lock isn't needed, but holding one around the list may keep coverity quiet

prov/tcp/src/xnet_rdm.c
```diff
@@ -695,13 +695,183 @@ static struct fi_ops_ep xnet_rdm_ep_ops = {
...
+	ret = ofi_mr_map_insert(&subdomain->util_domain.mr_map, attr,
+				&attr->requested_key, attr->context,
+				((struct ofi_mr*)attr->context)->flags);
+	if (ret)
+		XNET_WARN_ERR(FI_LOG_MR, "ofi_mr_map_insert", ret);
+	else
+		ofi_atomic_inc32(&subdomain->util_domain.ref);
+}
+
+struct xnet_domain *find_subdomain(struct xnet_rdm *rdm)
```
shefty, line 757: Use xnet prefix here and below.

shefty (commented):
partial review -- I did not look at xnet_rdm.c yet.

prov/tcp/src/xnet_rdm.c
```diff
@@ -695,13 +695,183 @@ static struct fi_ops_ep xnet_rdm_ep_ops = {
...
+				       &domain->subdomain_list_lock,
+				       &subdomain_fid->fid);
+		if (ret)
+			fi_close(&subdomain_fid->fid);
+
+		set_subdomain(rdm, domain, subdomain);
+
+		ofi_rbmap_foreach(domain->util_domain.mr_map.rbtree,
+				  domain->util_domain.mr_map.rbtree->root,
+				  xnet_reg_subdomain_mr, subdomain);
```
shefty, line 845: How are failures in reg_subdomain_mr exposed upwards?  ofi_rbmap_foreach may need to return an error code and abort if the function to invoke fails.

prov/tcp/src/xnet_rdm.c
```diff
@@ -695,13 +695,183 @@ static struct fi_ops_ep xnet_rdm_ep_ops = {
...
+	}
+
+	ret = xnet_mplex_av_dup(&rdm->util_ep, domain, subdomain);
```
shefty, line 848: If the subdomain already exists, hasn't the AV already been duplicated?

prov/tcp/src/xnet_av.c
```diff
@@ -0,0 +1,235 @@
...
+static const char *
+xnet_mplex_av_straddr(struct fid_av *av_fid, const void *addr, char *buf, size_t *len)
+{
+	return ofi_ip_av_straddr(av_fid, addr, buf, len);
+}
+
+static int xnet_mplex_av_set(struct fid_av *av_fid, struct fi_av_set_attr *attr,
+			     struct fid_av_set **av_set_fid, void *context)
+{
+	return -FI_ENOSYS;
```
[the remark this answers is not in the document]
ooststep: I don't see one. Should I create one and use that?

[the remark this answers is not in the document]
shefty: I would do that.

prov/tcp/src/xnet_cq.c
```diff
@@ -315,7 +315,6 @@ int xnet_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 	return ret;
 }
 
-
```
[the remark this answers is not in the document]
ooststep: This was likely due to some other code that ended up not being needed and just not put back properly :P

prov/tcp/src/xnet_domain.c
```diff
@@ -128,6 +171,81 @@ xnet_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+		goto out;
+
+	mr = container_of(*mr_fid, struct ofi_mr, mr_fid.fid);
+	mr->mr_fid.fid.ops = &xnet_mplex_mr_fi_ops;
+
+	ofi_genlock_lock(&domain->subdomain_list_lock);
+	dlist_foreach_container(&domain->subdomain_list,
+				struct fid_list_entry, item, entry) {
+		subdomain = container_of(item->fid, struct fid_domain, fid);
+		ret = fi_mr_regattr(subdomain, attr, flags, &sub_mr_fid);
```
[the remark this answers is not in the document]
ooststep: We don't really have a great way to keep track of the duplicated mr fids or correlate them back to the mr fid the application has a reference to.  `xnet_subdomains_mr_close` removes and cleans based on the key instead.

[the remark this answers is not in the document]
ooststep: I suppose we're actually missing a `free` call on the mr that is found by the key.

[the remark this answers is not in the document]
shefty: fi_mr_regattr is an upper layer API and may be hooked.  I would call some lower-level, internal API directly, which makes it clearer that it's not intending to adhere to the API semantics

prov/tcp/src/xnet_rdm.c
ooststep: yes I suppose this should move up inside the if block

ooststep: So, I guess it's not as simple as moving this up.  The av may truly be shared amongst eps, and in that case we likely don't want to duplicate but just update the references.  Instead, we likely want to check that a matching subav exists for the subdomain and only duplicate if it doesn't.

In attempting to resolve this, I discovered a case that must have escaped prior testing.  When the main av is populated prior to creating subavs, we iterate the av entry pool to insert into the new subav, however if there are regions in the main av that are not allocated, we will fail assertions in the underlying `ofi_bufpool_get_ibuf`.  I believe the options here are to relax the assertion, promote the `ofi_bufpool_hdr:allocated` flag from debug only, or to have some fairly significant code duplication.  I'm trying to find a better option before committing to any of those.

shefty: Why would there be regions in the main AV that aren't allocated?  Isn't the AV created using AV_TABLE semantics, which would populate in order?

ooststep: we could overwrite the pool's chunk_cnt for the main av and set it to 1.  it would mean we have to call "ofi_bufpool_grow" more often, but no other code would need to change.

shefty: I don't think we should interact directly with bufpool.  The AV APIs, even internal calls, should be usable, or updated to be usable.  We can use ofi_av_size() to get the size of the AV.  ofi_av_get_addr() should return the address for a given index (fi_addr_t), or NULL, I guess, if one does not exist.

ooststep: we are using the ofi_av calls.  the problem is that these calls (`ofi_av_get_addr`) call into `ofi_bufpool_get_ibuf` where we hit the assert if the debug `allocated` flag is false.  This `allocated` flag is the only indication that the buffer is _actually_ used. `ofi_av_size` returns the number of entries in the av, but doesn't tell you the number of entries that are actually used (marked `allocated`).

With all that said, it seems you may be leading me toward promoting that `allocated` flag so that we can rely on it to tell us the entries that are used rather than just for debug purposes.  this means that all providers that use the `util_av` will pick up that overhead (1 byte per entry?).

prov/tcp/src/xnet_rdm.c
```diff
@@ -695,13 +695,183 @@ static struct fi_ops_ep xnet_rdm_ep_ops = {
...
+		goto out;
+	util_av = container_of(av_fid, struct util_av, av_fid);
+	for (i = 0; i < ep->av->av_entry_pool->entry_cnt; i++) {
+		ret = fi_av_lookup(&ep->av->av_fid, i, addr, &addr_size);
+		if (ret)
+			continue;
+		ret = ofi_av_insert_addr_at(util_av, addr, i);
+		if (ret)
+			goto out;
+	}
```
shefty, line 726: This loop goes into the AV implementation.  It should use AV calls directly, even if internal AV calls.  If the AV needs to adjust to handle an invalid fi_addr_t index value, it should do that.

ooststep:
addressed most current feedback - still working on addressing a couple things to wrap this up

include/ofi_mem.h
```diff
@@ -561,6 +561,30 @@ static inline void *ofi_ibuf_alloc(struct ofi_bufpool *pool)
...
+	buf_hdr = ofi_buf_hdr(buf);
+	assert(buf_hdr->allocated == false);
+	OFI_DBG_SET(buf_hdr->allocated, true);
+
+	dlist_remove(&buf_hdr->entry.dlist);
+	if (dlist_empty(&buf_region->free_list))
+		dlist_remove_init(&buf_region->entry);
+
+	return buf;
+}
```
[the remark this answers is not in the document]
ooststep: I don't see a great way to combine the two functions.  they are just different enough that trying to reduce duplication may lead to more complexity.  the issue is with how to decide is we need to grow and then how to capture the entry data pointer.

ooststep:
I believe I've addressed all current feedback.

I have been validating these flows with the changes proposed in #10249.  It _could_ be reasonable rebase this change set onto that one once it gets merged to ensure some extra validation.

prov/tcp/src/xnet_rdm.c
```diff
@@ -695,13 +695,222 @@ static struct fi_ops_ep xnet_rdm_ep_ops = {
...
+
+	ofi_genlock_lock(&xnet_av->lock);
+	dlist_foreach_container(&xnet_av->subav_list, struct fid_list_entry, item, entry) {
+		subav = container_of(item->fid, struct util_av, av_fid.fid);
+		if (subav->domain == &subdomain->util_domain)
+			goto move_av;
+	}
+	ret = xnet_mplex_av_dup(ep, xnet_av, subdomain, &subav_fid);
+	if (ret)
+		return ret;
```
shefty, line 753: returns holding a lock

j-xiong (approved):

j-xiong:
@shefty Do you have any more comments on this PR?

shefty:
All my comments were addressed.  Looks good.

ooststep:
Thank you!

PR #10186, opened 2024-07-15
[v1.22.x] prov/efa: do not do copy for inline send

shijin-aws:
Backport https://github.com/ofiwg/libfabric/pull/10176


shijin-aws:
CI failure is not related

PR #10188, opened 2024-07-16
prov/efa: Add more tracepoints

shijin-aws:
Add tracepoints for rx pkts releases and data copies


shijin-aws:
bot:aws:retest

darrylabbate (commented):
I don't want to block this, but we should use [`LTTNG_UST_TRACEPOINT_EVENT_CLASS()`](https://lttng.org/man/3/lttng-ust/v2.13/#doc-tracepoint-event-class) for tracepoints that share common fields ([example](https://github.com/ofiwg/libfabric/blob/8a4249790e07250a9e4a2d66c10560ce588ded38/prov/efa/src/rdm/efa_rdm_tracepoint_def.h#L19-L66))

> LTTng-UST creates one event serialization function per tracepoint class. Using `LTTNG_UST_TRACEPOINT_EVENT()` creates one tracepoint class per tracepoint definition, whereas using `LTTNG_UST_TRACEPOINT_EVENT_CLASS()` and `LTTNG_UST_TRACEPOINT_EVENT_INSTANCE()` creates one tracepoint class, and one or more tracepoint instances of this class. In other words, many tracepoints can reuse the same serialization code. Reusing the same code, when possible, can reduce cache pollution, thus improve performance.

shijin-aws:
> LTTng-UST creates one event serialization function per tracepoint class. Using `LTTNG_UST_TRACEPOINT_EVENT()` creates one tracepoint class per tracepoint definition, whereas using `LTTNG_UST_TRACEPOINT_EVENT_CLASS()` and `LTTNG_UST_TRACEPOINT_EVENT_INSTANCE()` creates one tracepoint class, and one or more tracepoint instances of this class. In other words, many tracepoints can reuse the same serialization code. Reusing the same code, when possible, can reduce cache pollution, thus improve performance.

That is a good idea, but is out of scope of this PR.  Also using this may require some changes in the babeltrace2 plugin for the postprocess I guess. AFAIK it just regards all TPs as events

darrylabbate:
> Also using this may require some changes in the babeltrace2 plugin for the postprocess I guess?

It shouldn't have any effect wrt tracepoint names or babeltrace. It's simply an optimization that only adds a few lines of code (and removes a few as well).

I don't agree that it's out of scope, but it's not critical enough to block merging.

darrylabbate (approved):

shijin-aws:
Updated the PR to use event class and event instance. I missed that we already use this feature for msg_context and x_entry classes

Issue #9443, opened 2023-10-16
prov/tcp, prov/verbs: fi_getinfo() does not seem to consider `hints->addr_format`


sydidelot:
**Describe the bug**
`fi_getinfo()` doesn't seem to correctly honor the address format requested by `hinst->addr_format` and returns `fi_info` structures from different address formats.

**To Reproduce/Expected behavior**
The issue can easily be reproduced with the tool `fi_info` and the `-a` argument.
With `-a <addr_format>`, I would expect only `fi_info` structures with the requested address format to be returned, which is not the case for at least two providers: TCP and Verbs. Here is a couple of examples:

For TCP with `FI_SOCKADDR_IN`: ipv6 addresses are returned, which is unexpected:
```
$ fi_info -p tcp -a FI_SOCKADDR_IN -vvv | grep sockaddr
    src_addr: fi_sockaddr_in://172.25.60.35:0
    src_addr: fi_sockaddr_in6://[fe80::966d:aeff:fe33:7a10]:0
    src_addr: fi_sockaddr_in://10.25.60.232:0
    src_addr: fi_sockaddr_in6://[fe80::a242:3fff:fe50:6143]:0
    src_addr: fi_sockaddr_in://127.0.0.1:0
    src_addr: fi_sockaddr_in6://[::1]:0
    src_addr: fi_sockaddr_in://172.25.60.35:0
    src_addr: fi_sockaddr_in6://[fe80::966d:aeff:fe33:7a10]:0
    src_addr: fi_sockaddr_in://10.25.60.232:0
    src_addr: fi_sockaddr_in6://[fe80::a242:3fff:fe50:6143]:0
    src_addr: fi_sockaddr_in://127.0.0.1:0
    src_addr: fi_sockaddr_in6://[::1]:0
    src_addr: fi_sockaddr_in://172.25.60.35:0
    src_addr: fi_sockaddr_in6://[fe80::966d:aeff:fe33:7a10]:0
    src_addr: fi_sockaddr_in://10.25.60.232:0
    src_addr: fi_sockaddr_in6://[fe80::a242:3fff:fe50:6143]:0
    src_addr: fi_sockaddr_in://127.0.0.1:0
    src_addr: fi_sockaddr_in6://[::1]:0
    src_addr: fi_sockaddr_in://172.25.60.35:0
    src_addr: fi_sockaddr_in6://[fe80::966d:aeff:fe33:7a10]:0
[...]

PR #10201, opened 2024-07-20
prov/verbs: Fix issue while displaying addresses with fi_info -a <addr_format>

Juee14Desai:
This commit addresses issue #9443 for verbs provider. Address format matching needs to happen for each info in the linked list which was not happening earlier. This check has been moved to the end of vrb_getinfo().

### DIFF
--- a/prov/verbs/src/verbs_info.c
+++ b/prov/verbs/src/verbs_info.c
@@ -1890,6 +1890,24 @@ void vrb_alter_info(const struct fi_info *hints, struct fi_info *info)
 	}
 }
 
+static void vrb_filter_info_by_addr_format(struct fi_info **info, int addr_format)
+{
+	struct fi_info *cur, *prev= NULL, *next = NULL;
+	for (cur = *info; cur; cur = next) {
+		next = cur->next;
+		if (!ofi_match_addr_format(cur->addr_format, addr_format)) {
+			if (prev)
+				prev->next = cur->next;
+			else
+				*info = cur->next;
+			cur->next = NULL;
+			fi_freeinfo(cur);
+		} else {
+			prev = cur;
+		}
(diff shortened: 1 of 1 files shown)


prov/verbs/src/verbs_info.c
```diff
@@ -243,9 +243,6 @@ static int vrb_check_hints(uint32_t version, const struct fi_info *hints,
 		return -FI_ENODATA;
 	}
 
-	if (!ofi_valid_addr_format(info->addr_format, hints->addr_format))
-		return -FI_ENODATA;
-
```
j-xiong, line 248: why remove this check?

Juee14Desai: This check is redundant since we moved it to vrb_filter_info_by_addr_format().

j-xiong: It's not redundant. User may pass an invalid addr_format. It should be check here instead of taking all the effort to set it and then remove it later.

Juee14Desai: Corrected in recent push.

prov/verbs/src/verbs_info.c
```diff
@@ -1910,6 +1928,9 @@ int vrb_getinfo(uint32_t version, const char *node, const char *service,
 	ofi_alter_info(*info, hints, version);
 
 	vrb_alter_info(hints, *info);
+
+	if (hints)
+			vrb_filter_info_by_addr_format(info, hints->addr_format);
```
j-xiong, line 1933: too much indentation. Remove one tab here.

Juee14Desai:
CI failure is unrelated to changes.

j-xiong (approved):

PR #10207, opened 2024-07-22
prov/efa: disable p2p support only if CUDA MR returns a fatal error

wenduwan:
registration failed and it's not required by the hmem iface.

This is undesirable for 2 reasons:
1. MR could fail for transient reasons such as ENOMEM, and can be safely retried.
2. P2P support can only be disabled for CUDA because we do not handle the fallback case for other hmem ifaces.

In this change we refine the MR error handling logic:
1. If CUDA p2p is not required, fallback to internal keygen; p2p is disdabled only if MR returns a fatal error other than ENOMEM.
2. Otherwise, return MR error but do not disable p2p support.


shijin-aws (changes requested):

prov/efa/src/efa_mr.c
```diff
@@ -854,30 +854,30 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *at
 		}
 	}
 
-	if (ret) {
+	if (ret && mr_attr.iface == FI_HMEM_CUDA && !efa_mr->domain->hmem_info[mr_attr.iface].p2p_required_by_impl) {
```
shijin-aws, line 857: I feel we make the code over complicated by checking `p2p_required_by_impl`

This is a bool useful for data transfer call only, not for fi_mr_reg

can we reorganize this block to 

```
if (ret) {
   if (iface == cuda and ret != -ENOMEM) {
    // disable p2p
   } else {
    // fail and return
   }
}

if (iface == cuda && p2p_is_disabled) {
 // generate mr_key manually
}
```

wenduwan: Please see update. I changed the branching slightly to avoid `if (iface == cuda && p2p_is_disabled)`.

Overall there are 3 possible error cases:
- CUDA && p2p is not required: disable p2p if ibv_reg_mr failed
- ibv_reg_mr(p2p) failed:  return error
- p2p is not supported(did not call ibv_reg_mr): return `FI_EOPNOTSUPP`

> This is a bool useful for data transfer call only, not for fi_mr_reg

However I don't understand this comment. Could you explain?

prov/efa/src/efa_mr.c
```diff
@@ -854,30 +854,30 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *at
...
-			 "Unable to register %s memory, disable P2P support.\n",
-			 fi_tostr(&mr_attr.iface, FI_TYPE_HMEM_IFACE));
-		efa_mr->domain->hmem_info[mr_attr.iface].p2p_supported_by_device = false;
-		ret = FI_SUCCESS;
-	}
-
-	/*
-	 * For FI_HMEM_CUDA iface when p2p is unavailable, skip ibv_reg_mr() and generate proprietary mr_fid key.
-	 */
-	if (mr_attr.iface == FI_HMEM_CUDA && !efa_mr->domain->hmem_info[mr_attr.iface].p2p_supported_by_device) {
```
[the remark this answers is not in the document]
wenduwan: Good catch! Fixed

prov/efa/src/efa_mr.c
shijin-aws: `p2p_required` is used during the transmission protocol selection in fi_send/fi_write, 

>FI_HMEM_P2P_REQUIRED: Peer to peer support must be used for transfers, transfers that cannot be performed using p2p will be reported as failing.

That being said, if user set this, and cuda p2p is disabled (while fi_mr_reg still succeeded), we should fail in the transmission call.

In other words, fi_mr_reg doesn't care this option, provider can choose to always do ibv_mr_reg or not internally. In our case, we don't require ibv_mr_reg in fi_mr_reg only for cuda. For any other hmem ifaces, we should always do ibv_mr_reg and fail if we can't.

In this regard, the `hmem_info[iface].p2p_required` shouldn't show in the reg call. We only do the p2p disabled fall back for cuda.

shijin-aws: During the transmission call, for non-cuda hmem, we don't accept any non-NULL desc, that means a successful ibv_mr_reg already happened earlier. Then the check of p2p_required is automatically satisfied.

For cuda hmem, since the transmission call only happen after a successful fi_mr_reg, we can check the p2p_required with the actual p2p status for cuda. If user required p2p while we disabled p2p, we will fail and return. This is also true for the current code base.

shijin-aws: Summarily, you can always keep `hmem_info[iface].p2p_supported` as `true` for non-cuda ifaces, as non-p2p mode doesn't exist for them. you only update `hmem_iface[FI_HMEM_CUDA].p2p_supported` selectively.

shijin-aws: >An important reason is that, p2p_supported_by_device is per hmem iface, which is independent from the EP. When the application sets FI_HMEM_P2P_REQUIRED(or other values) on the EP, we do not have any guarantee that the p2p_supported_by_device has been finalized,

That was my earlier concern, and then I think it's not an issue as the man page says this bool should be related to transmission. When application sets FI_HMEM_P2P_REQUIRED, we don't have to fail during fi_setopt. we will check with the `p2p_supported_by_device` in the transmission call, when `p2p_supported_by_device` has been finalized.

shijin-aws (approved):

shijin-aws:
Pure efa change, ignore Intel CI failure

PR #10210, opened 2024-07-23
prov/efa: introduce FI_EFA_IFACE to restrict visible NICs

wenduwan:
This patch introduces an environment variable to limit which NICs are visible to the application via fi_getinfo.


j-xiong:
Could it be named FI_EFA_IFACE to be consistent with other providers?

shijin-aws:
> Could it be named FI_EFA_IFACE to be consistent with other providers?

Could you share a link to the implementation of other providers?

j-xiong:
@shijin-aws See these man pages about **FI_TCP_IFACE** and **FI_VERBS_IFACE**:
https://ofiwg.github.io/libfabric/main/man/fi_tcp.7.html
https://ofiwg.github.io/libfabric/main/man/fi_verbs.7.html

wenduwan:
@j-xiong Thank you! PR is updated accordingly.

wenduwan:
@zachdworkin I imagine the intel CI failure is unrelated, correct?

shijin-aws:
Should we advertise this env in man page?

j-xiong:
Yes, I think so.

prov/efa/src/efa_env.h
```diff
@@ -92,7 +95,38 @@ static inline bool efa_env_has_use_device_rdma() {
...
+
+	if (!strncmp("all", efa_env.iface, 3))
+		return true;
+
+	while (match < end) {
+		match = strstr(match, name);
+		if (!match)
+			return false;
+
+		if ((match > efa_env.iface && *(match - 1) != ',') ||
```
shijin-aws, line 119: hmmm how do we compare two strings here?

wenduwan:
> Should we advertise this env in man page?

Good point. Added documentation.

prov/efa/src/efa_env.h
```diff
@@ -92,7 +95,38 @@ static inline bool efa_env_has_use_device_rdma() {
...
+		    ((match + strlen(name)) < end && *(match + strlen(name)) != ',')) {
+			/* Skip partial match */
+			match += strlen(name);
```
shijin-aws, line 122: Clarified with @wenduwan offline. But this logic is not intuitive to new readers.

shijin-aws (approved):

shijin-aws:
bot:aws:retest

PR #10211, opened 2024-07-23
v1.22.0

j-xiong:
For CI only. Do not merge. Will be pushed directly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,6 +35,7 @@ v1.22.0, Fri Jul 26, 2024
 ## EFA
 
 - General improvement and bug fixes
+- Handle recv cancel for zero copy recv
 - Avoid iterating EP list in CQ read
 - Add RDMA core errno for remote unknown peer
 - Map EFA errnos to Libfabric codes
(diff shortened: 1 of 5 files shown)


belynam:
Hi @j-xiong ,

Looking at the release candidate downloads [here](https://github.com/ofiwg/libfabric/releases/tag/v1.22.0rc2), specifically **libfabric-1.22.0rc2.tar.bz2**, I find that [prov/opx/provider_FABRIC_1.0.map](https://github.com/ofiwg/libfabric/blob/main/prov/opx/provider_FABRIC_1.0.map) is **not** included. This causes an error when running `configure` and specifying `--enable-direct=opx`. What would we need to do to make sure that file is included in official release tarballs?

j-xiong:
@belynam Have you ever been able to do it before? The same issue exists for opx, psm2 and sockets. I can fix it but want to double check that it's not a regression.

j-xiong:
@belynam I have included a fix here.

belynam:
> @belynam Have you ever been able to do it before? The same issue exists for opx, psm2 and sockets. I can fix it but want to double check that it's not a regression.

I don't think this is a regression, I see the same issue on previous release tarballs. Historically when testing, I've only downloaded the source itself or pulled down the branch to test the build. I'm not aware of any of our users running into this issue, so maybe most don't use that option.

> @belynam I have included a fix here.

Thank you!

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

PR #10220, opened 2024-07-25
hmem/cuda: Modify the logging for nvml dlopen

shijin-aws:
Move the libnvml.so -> libnvml.so.1 dlopen log from warning to info as they are informational only. Adjust the warning log when libnvlml.so.1 dlopen failure.


wenduwan (approved):

darrylabbate (approved):

shijin-aws:
@j-xiong could you review it, thx

PR #10221, opened 2024-07-25
prov/opx: Add GDRCopy logging and failure path

charlesshereda:
Added messaging and some failures when GDRCopy is/is not being used and is/is not available.

PR #10222, opened 2024-07-25
prov/opx: Add GDRCopy logging and failure path

charlesshereda:
Added messaging and some failures when GDRCopy is/is not being used and is/is not available.


j-xiong:
Same comments as #10221

j-xiong:
Please squash the two commits into one.

PR #10223, opened 2024-07-26
prov/verbs: Replace __BITS_PER_LONG with LONG_WIDTH

raffenet:
~The checks for the verbs provider can pass on a FreeBSD system, but will fail to compile because asm/types.h is not available.~

Updated description: 2024-08-01
The verbs provider uses asm/types.h for __BITS_PER_LONG, but that header
is limited to Linux systems. LONG_WIDTH is defined in C23 in
limits.h. Use it instead, and add logic to define it if not
available. This will allow the verbs provider to compile on FreeBSD
systems.

### DIFF
--- a/prov/verbs/configure.m4
+++ b/prov/verbs/configure.m4
@@ -44,6 +44,8 @@ AC_DEFUN([FI_VERBS_CONFIGURE],[
 
 	      ])
 
+	AC_CHECK_HEADERS([asm/types.h])
+
 	AS_IF([test $verbs_ibverbs_happy -eq 1 && \
 	       test $verbs_rdmacm_happy -eq 1], [$1], [$2])
 
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -1999,7 +1999,7 @@ check_datatype:
 	switch (datatype) {
 	case FI_INT64:
 	case FI_UINT64:
-#if __BITS_PER_LONG == 64
+#if LONG_WIDTH == 64
 	case FI_DOUBLE:
 	case FI_FLOAT:
 #endif
--- a/prov/verbs/src/verbs_ofi.h
+++ b/prov/verbs/src/verbs_ofi.h
@@ -39,7 +39,6 @@
 
 #include "config.h"
 
-#include <asm/types.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <arpa/inet.h>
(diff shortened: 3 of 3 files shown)


j-xiong:
The header is included in `prov/verbs/src/ofi_verbs.h`. Can verbs be built on BSD by commenting out the line? If so, a better fix would be using a conditional to guard the `#include` line.

raffenet:
It does compile if I comment out that header, but gives this warning about a missing define.
```c
--- prov/verbs/src/src_libfabric_la-verbs_ep.lo ---
prov/verbs/src/verbs_ep.c: In function 'vrb_query_atomic':
prov/verbs/src/verbs_ep.c:2002:5: warning: "__BITS_PER_LONG" is not defined, evaluates to 0 [-Wundef]
 2002 | #if __BITS_PER_LONG == 64
```

raffenet:
> Something like this should work:
> 
> ```
> #if HAVE_ASM_TYPES_H
> #include <asm/types.h>
> #else
> #if defined(__x86_64__) && !defined(__ILP32__)
> #define __BITS_PER_LONG 64
> #else
> #define __BITS_PER_LONG 32
> #endif
> #endif
> ```
> 
> where `HAVE_ASM_TYPES_H` needs to be defined by configure.

OK. I will work on it later this week.

raffenet:
> OK. I will work on it later this week.

BTW, it looks like C23 adds `LONG_WIDTH` in `limits.h` which we could also test for before using the fallback code? See https://en.cppreference.com/w/c/types/limits

j-xiong:
@raffenet I think that's something worth experimenting with (looks like it's already supported by gcc). Just need to keep the logic simple.

raffenet:
> @raffenet I think that's something worth experimenting with (looks like it's already supported by gcc). Just need to keep the logic simple.

What do you think about this? Looks like `LONG_BIT` is a non-standard define and `LONG_WIDTH` is the new standard. I also added support for `__aarch64__` so that it sets the right value on my laptop. It's kind of verbose, so maybe it belongs in an `osd.h` header rather than the provider?

```c
#include <limits.h>
#if !defined(LONG_WIDTH)
#ifdef LONG_BIT
#define LONG_WIDTH LONG_BIT
#elif defined(HAVE_ASM_TYPES_H)
#include <asm/types.h>
#define LONG_WIDTH __BITS_PER_LONG
#elif defined(__x86_64__) || defined(__aarch64__)
#ifdef !defined(__ILP32__)
#define LONG_WIDTH 64
#else
#define LONG_WIDTH 32
#endif
#endif
#endif
```

j-xiong:
@raffenet Thanks! This looks good to me. Let's keep it in the verbs provider for now since it's only used here.

raffenet:
Is the AWS CI failure real or no?

wenduwan:
@raffenet Likely noise. I'm restarting it.

wenduwan:
bot:aws:retest

j-xiong (approved):

PR #10245, opened 2024-07-30
prov/efa: fix typo in hmem unit test

wenduwan:
Fix a bug in ibv_reg_mr mocked function that did not have a return value.


shijin-aws (approved):
I am curious why this wasn't caught by CI (when building unit test with --enable-debug)

PR #10215, opened 2024-07-24
prov/efa: Handle recv cancel for zero-copy recv

shijin-aws:
Currently, posted recv is not tracked in zero-copy mode,
which breaks various ep operations including fi_cancel. This patch fixes this
issue by introducing a user_recv_rxe_list that tracks the posted user recv,
and implement fi_cancel operation for this list.

PR #10248, opened 2024-07-30
prov/efa: Make fi_cancel return ENOTSUPP for zero copy receive mode. 

shijin-aws:
A receive cannot be safely cancelled in zero copy receive mode
as we cannot cancel the recv in the HW level. Make fi_cancel return
ENOTSUPP instead of making hacks that don't really fake the behaviors.

This PR contains 2 commits

First commit is the revert of https://github.com/ofiwg/libfabric/pull/10215


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1367,9 +1313,12 @@ ssize_t efa_rdm_ep_cancel(fid_t fid_ep, void *context)
 	struct efa_rdm_ep *ep;
 
 	ep = container_of(fid_ep, struct efa_rdm_ep, base_ep.util_ep.ep_fid.fid);
+	if (ep->use_zcpy_rx) {
+		EFA_WARN(FI_LOG_EP_CTRL, "fi_cancel is not supported in zero-copy receive mode. \n");
```
darrylabbate, line 1317: nit
```suggestion
		EFA_WARN(FI_LOG_EP_CTRL, "fi_cancel is not supported in zero-copy receive mode.\n");
```

darrylabbate (approved):

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: Thanks, updated

darrylabbate (approved):

PR #10255, opened 2024-07-31
hmem/rocr: Fix dmabuf for amd gpu implementation - 1.22

chuckfossen:
Two symbols were added to check for and then get the dmabuf fd. They were not initialized.


chuckfossen:
@j-xiong - last one

j-xiong (approved):

PR #10256, opened 2024-07-31
prov/shm: Re-enable ze-ipc with new pidfd method

zachdworkin:
Refactor close_handle to add new field for ze pidfd method
Implement ze pidfd method for ze-ipc copies
Remove old method's (socket by path) code

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -204,24 +204,34 @@ int cuda_gdrcopy_dev_register(const void *buf, size_t len, uint64_t *handle);
 int cuda_gdrcopy_dev_unregister(uint64_t handle);
 int cuda_set_sync_memops(void *ptr);
 
+/*
+ * ze handle is just an fd so we need a wrapper to contain extra fields for ipc
+ * copies
+ * get_fd 	fd from handle_get
+ * open_fd	fd from pidfd_get_fd. Needed for closing extra fd and for
+ * 		opening with open_handle
+ * pid_fd	fd that references a pid used to open an fd across processes
+ */
+struct ze_pid_handle {
+	int get_fd;
+	int open_fd;
+	int pid_fd;
+};
+
 int ze_hmem_copy(uint64_t device, void *dst, const void *src, size_t size);
 int ze_hmem_init(void);
 int ze_hmem_cleanup(void);
 bool ze_hmem_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags);
 int ze_hmem_get_handle(void *dev_buf, size_t size, void **handle);
+void ze_set_pid_fd(void **handle, int pid_fd);
 int ze_hmem_open_handle(void **handle, size_t size, uint64_t device,
 			void **ipc_ptr);
-int ze_hmem_get_shared_handle(uint64_t device, void *dev_buf, int *ze_fd,
-			      void **handle);
-int ze_hmem_open_shared_handle(uint64_t device, int *peer_fds, void **handle,
-			       int *ze_fd, void **ipc_ptr);
 int ze_hmem_close_handle(void *ipc_ptr, void **handle);
 bool ze_hmem_p2p_enabled(void);
 int ze_hmem_get_ipc_handle_size(size_t *size);
 int ze_hmem_get_base_addr(const void *ptr, size_t len, void **base,
 			  size_t *size);
 int ze_hmem_get_id(const void *ptr, uint64_t *id);
-int *ze_hmem_get_dev_fds(int *nfds);
 int ze_hmem_host_register(void *ptr, size_t size);
 int ze_hmem_host_unregister(void *ptr);
 int ze_dev_register(const void *addr, size_t size, uint64_t *handle);
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -116,7 +116,7 @@ struct ofi_hmem_ops {
 	int (*get_handle)(void *base_addr, size_t base_length, void **handle);
 	int (*open_handle)(void **handle, size_t base_length, uint64_t device,
 			   void **mapped_addr);
-	int (*close_handle)(void *mapped_addr);
+	int (*close_handle)(void *mapped_addr, void **handle);
 	int (*host_register)(void *addr, size_t size);
 	int (*host_unregister)(void *addr);
 	int (*get_base_addr)(const void *addr, size_t len, void **base_addr,
@@ -149,7 +149,7 @@ int rocr_get_base_addr(const void *ptr, size_t len, void **base, size_t *size);
 int rocr_get_handle(void *dev_buf, size_t size, void **handle);
 int rocr_open_handle(void **handle, size_t size, uint64_t device,
 		     void **ipc_ptr);
-int rocr_close_handle(void *ipc_ptr);
+int rocr_close_handle(void *ipc_ptr, void **handle);
 bool rocr_is_ipc_enabled(void);
 int rocr_create_async_copy_event(uint64_t device,
 				 ofi_hmem_async_event_t *event);
@@ -184,7 +184,7 @@ int cuda_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
 int cuda_get_handle(void *dev_buf, size_t size, void **handle);
 int cuda_open_handle(void **handle, size_t size, uint64_t device,
 		     void **ipc_ptr);
-int cuda_close_handle(void *ipc_ptr);
+int cuda_close_handle(void *ipc_ptr, void **handle);
 int cuda_get_base_addr(const void *ptr, size_t len, void **base, size_t *size);
 
 bool cuda_is_ipc_enabled(void);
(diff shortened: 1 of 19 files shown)


prov/shm/src/smr_ep.c
```diff
@@ -347,15 +312,16 @@ static int smr_format_ipc(struct smr_cmd *cmd, void *ptr, size_t len,
...
+				&cmd->msg.data.ipc_info.base_length);
 	if (ret)
 		return ret;
 
 	ret = ofi_hmem_get_handle(cmd->msg.data.ipc_info.iface, base,
-				   cmd->msg.data.ipc_info.base_length,
-				   (void **)&cmd->msg.data.ipc_info.ipc_handle);
+			cmd->msg.data.ipc_info.base_length,
+			(void **)&cmd->msg.data.ipc_info.ipc_handle);
 	if (ret)
```
j-xiong, line 325: These format changes are unnecessary, and the original format was preferred.

prov/shm/src/smr_ep.c
```diff
@@ -740,15 +706,8 @@ static ssize_t smr_do_ipc(struct smr_ep *ep, struct smr_region *peer_smr, int64_
...
-		if (smr_ze_ipc_enabled(ep->region, peer_smr))
-			ret = smr_format_ze_ipc(ep, id, cmd, iov,
-					desc[0]->device, total_len, ep->region,
-					resp, pend);
-	} else {
-		ret = smr_format_ipc(cmd, iov[0].iov_base, total_len, ep->region,
-				     resp, desc[0]->iface, desc[0]->device);
-	}
+	ret = smr_format_ipc(cmd, iov[0].iov_base, total_len, ep->region,
+				resp, desc[0]->iface, desc[0]->device);
```
j-xiong, line 710: nitpick: try to align 'resp' with 'cmd'.

src/hmem_ze.c
```diff
@@ -1055,42 +915,65 @@ bool ze_hmem_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags)
...
+	fd = ofi_pidfd_getfd(pid_handle->pid_fd, pid_handle->get_fd, 0);
+	if (fd < 0) {
+		FI_WARN(&core_prov, FI_LOG_CORE,
+			"Unable find fd from peer pid\n");
+	}
+	pid_handle->open_fd = fd;
+
+	memcpy(&temp_handle, &pid_handle->open_fd, sizeof(pid_handle->open_fd));
+	ze_ret = ofi_zeMemOpenIpcHandle(context, devices[dev_id], temp_handle,
+					0, &buf);
```
j-xiong, line 959: Does it make sense to open ipc handle when open_fd is invalid?

prov/shm/src/smr_ep.c
j-xiong: This refers to the changes in the previous 10 lines. Somehow the whole range was not marked.

prov/shm/src/smr_ep.c
```diff
@@ -347,15 +312,16 @@ static int smr_format_ipc(struct smr_cmd *cmd, void *ptr, size_t len,
 	cmd->msg.hdr.size = len;
 	cmd->msg.data.ipc_info.iface = iface;
 	cmd->msg.data.ipc_info.device = device;
-	ret = ofi_hmem_get_base_addr(cmd->msg.data.ipc_info.iface, ptr, len,
-				     &base,
-				     &cmd->msg.data.ipc_info.base_length);
+
+	ret = ofi_hmem_get_base_addr(cmd->msg.data.ipc_info.iface, ptr,
+				len, &base,
+				&cmd->msg.data.ipc_info.base_length);
```
j-xiong, line 318: Please fix the alignment here.

src/hmem_ze.c
```diff
@@ -1055,42 +915,66 @@ bool ze_hmem_is_addr_valid(const void *addr, uint64_t *device, uint64_t *flags)
...
+	}
+
+	pid_handle->open_fd = fd;
+	memcpy(&temp_handle, &pid_handle->open_fd, sizeof(pid_handle->open_fd));
+	ze_ret = ofi_zeMemOpenIpcHandle(context, devices[dev_id], temp_handle,
+					0, &buf);
 	if (ze_ret) {
 		FI_WARN(&core_prov, FI_LOG_CORE,
 			"Unable to open memory handle\n");
+		close(fd);
```
j-xiong, line 964: Also need to reset pid_handle->open_fd to -1 since it is no longer valid.

j-xiong (commented):
Just one more comment.

j-xiong (approved):

zachdworkin:
I will re-run ci after updates from #10272 are done.

prov/shm/src/smr_ep.c
```diff
@@ -1189,80 +1040,6 @@ void smr_ep_exchange_fds(struct smr_ep *ep, int64_t id)
...
-	ep->sock_info->listen_sock = socket(AF_UNIX, SOCK_STREAM, 0);
-	if (ep->sock_info->listen_sock < 0)
-		goto free;
-
-	snprintf(smr_sock_name(ep->region), SMR_SOCK_NAME_MAX,
-		 "%ld:%d", (long) ep->region->pid, ep->ep_idx);
-
-	sockaddr.sun_family = AF_UNIX;
-	snprintf(sockaddr.sun_path, SMR_SOCK_NAME_MAX,
-		 "%s%s", SMR_ZE_SOCK_PATH, smr_sock_name(ep->region));
```
aingerson, line 1211: There are probably some defines you can remove - `SMR_SOCK_NAME_MAX` and `SMR_ZE_SOCK_PATH`

include/ofi_hmem.h
```diff
@@ -204,24 +204,33 @@ int cuda_gdrcopy_dev_register(const void *buf, size_t len, uint64_t *handle);
...
+ * copies
+ * get_fd 	fd from handle_get
+ * open_fd	fd from handle_open needed for closing extra fd
+ * pid_fd	fd that references a pid used to open an fd across processes
+ */
+struct ze_pid_handle {
+	int get_fd;
+	int open_fd;
+	int pid_fd;
+};
```
aingerson, line 218: Edit commit message to remove pid from here

prov/shm/src/smr_progress.c
```diff
@@ -906,6 +889,8 @@ static void smr_progress_connreq(struct smr_ep *ep, struct smr_cmd *cmd)
 		smr_map_to_region(&smr_prov, ep->region->map, idx);
 		peer_smr = smr_peer_region(ep->region, idx);
 	}
+
+	smr_set_ipc_invalid(ep->region, idx);
```
aingerson, line 893: I'd rename this to either `smr_set_ipc_valid` or `smr_set_ipc_support` or something like that. This kind of sounds like we're disabling ipc always

prov/shm/src/smr.h
```diff
@@ -320,6 +320,19 @@ static inline bool smr_vma_enabled(struct smr_ep *ep,
 			peer_smr->xpmem_cap_self == SMR_VMA_CAP_ON);
 }
 
+static inline void smr_set_ipc_invalid(struct smr_region *region, uint64_t id)
+{
+	if (region->map->peers[id].pid_fd == -1)
```
aingerson, line 325: Since the pid_fd gets initialized to -1 for non ZE cases, do we accidentally disable the other iface IPCs?

Do we need to add an if ZE check to conditionally set the validity?
```
static inline void smr_set_ipc_invalid(struct smr_region *region, uint64_t id)
{
	if (ofi_hmem_initialized(FI_HMEM_ZE) && region->map->peers[id].pid_fd == -1)
		smr_peer_data(region)[id].ipc_invalid = 1;
}
```

prov/shm/src/smr_util.h
```diff
@@ -177,7 +177,8 @@ struct smr_addr {
 struct smr_peer_data {
 	struct smr_addr		addr;
 	uint32_t		sar_status;
-	uint32_t		name_sent;
+	uint16_t		name_sent;
+	uint16_t		ipc_invalid;
```
aingerson, line 181: You'll need to add a version bump for these changes since you've touched the shm region and protocols

include/ofi_hmem.h
```diff
@@ -204,24 +204,33 @@ int cuda_gdrcopy_dev_register(const void *buf, size_t len, uint64_t *handle);
 int cuda_gdrcopy_dev_unregister(uint64_t handle);
 int cuda_set_sync_memops(void *ptr);
 
+/*
+ * ze handle is just an fd so we need a wrapper to contain extra fields for ipc
+ * copies
+ * get_fd 	fd from handle_get
+ * open_fd	fd from handle_open needed for closing extra fd
```
aingerson, line 211: I think the open_fd is actually the one returned from the pidfd_get_fd call, not the one from open handle

prov/shm/src/smr.h
```diff
@@ -320,6 +319,20 @@ static inline bool smr_vma_enabled(struct smr_ep *ep,
 			peer_smr->xpmem_cap_self == SMR_VMA_CAP_ON);
 }
 
+static inline void smr_set_ipc_valid(struct smr_region *region, uint64_t id)
+{
+	if (ofi_hmem_is_initialized(FI_HMEM_ZE) &&
+	    region->map->peers[id].pid_fd == -1)
+		smr_peer_data(region)[id].ipc_invalid = 1;
```
aingerson, line 326: I'm going to be really annoying and recommend that we switch the bool from ipc_invalid to ipc_valid. Originally it made sense to do it as invalid so that it would get initialized to 0 and we wouldn't have to modify the checks to handle other ifaces but now that we check for ZE when we initialize the value, I think it makes more sense to have ipc_valid and then have this check be
```
static inline void smr_set_ipc_valid(struct smr_region *region, uint64_t id)
{
	if (ofi_hmem_is_initialized(FI_HMEM_ZE) &&
	    region->map->peers[id].pid_fd == -1)
		smr_peer_data(region)[id].ipc_valid = 0;
        else
               smr_peer_data(region)[id].ipc_valid = 1;
}
```
and then the smr_ipc_valid check can be simplified to remove the !s
That change makes the bool more readable and removes a few instructions on the fast path to check whether we can do ipc.
Sorryyyy

aingerson (approved):

PR #10258, opened 2024-07-31
prov/efa: Require shm to be disabled for using zero-copy recv

shijin-aws:
zero-copy receive cannot work with mixed efa and shm traffic, this patch makes shm disabling as a requirement for using zero-copy recv.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_fiops.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_fiops.c
@@ -466,6 +466,12 @@ void efa_rdm_ep_set_use_zcpy_rx(struct efa_rdm_ep *ep)
 		goto out;
 	}
 
+	if (ep->shm_ep) {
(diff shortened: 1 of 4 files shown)


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -466,6 +466,12 @@ void efa_rdm_ep_set_use_zcpy_rx(struct efa_rdm_ep *ep)
 		goto out;
 	}
 
+	if (ep->shm_ep) {
+		EFA_INFO(FI_LOG_EP_CTRL, "Libfabric SHM is not turned off, zero-copy receive protocol will be disabled\n");
```
wenduwan, line 470: Shall we also document this limitation on man page?

wenduwan (approved):
Minor comment

darrylabbate (approved):

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: Yes, updated the rdm protocol doc

PR #10259, opened 2024-07-31
prov/efa: Register user recv buffer for zero-copy receive mode

shijin-aws:
When user doesn't register the recv buffer, libfabric should try to register it so it can be posted to device for receive. If the registration failed, an EAGAIN will be returned.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_utils.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_utils.c
@@ -207,7 +207,7 @@ struct efa_rdm_ope *efa_rdm_ep_alloc_rxe(struct efa_rdm_ep *ep, fi_addr_t addr, 
  */
 int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe, size_t flags)
 {
-	struct efa_rdm_pke *pkt_entry;
+	struct efa_rdm_pke *pkt_entry = NULL;
 	size_t rx_iov_offset = 0;
 	int err, rx_iov_index = 0;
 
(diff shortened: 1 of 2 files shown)


prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -225,29 +225,46 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
 	err = ofi_iov_locate(rxe->iov, rxe->iov_count, ep->msg_prefix_size, &rx_iov_index, &rx_iov_offset);
 	if (OFI_UNLIKELY(err)) {
 		EFA_WARN(FI_LOG_CQ, "ofi_iov_locate failure: %s (%d)\n", fi_strerror(-err), -err);
-		return err;
+		goto err_free;
```
shijin-aws, line 228: This is actually a bug fix, the old code doesn't free pkt entry before return in the err path

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -225,29 +225,46 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
...
 	}
 
 #if ENABLE_DEBUG
 	dlist_insert_tail(&pkt_entry->dbg_entry, &ep->rx_posted_buf_list);
 #endif
 	ep->user_rx_pkts_posted++;
 	return 0;
+
+err_free:
+	if (pkt_entry)
```
darrylabbate, line 265: Is it possible for `pkt_entry` to be `NULL` at this point?

shijin-aws: Not possible, just for safe in case someone messed the sequence of code in the future. Do you prefer to remove it?

darrylabbate (approved):

PR #10260, opened 2024-07-31
prov/efa: Update data types for various IOV operations

darrylabbate:
In accordance with various changes in #10096

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -248,11 +248,11 @@ int efa_domain_hmem_support_init_all(struct efa_domain *efa_domain)
  * @param[in]   iov_count     Number of IOV structures in IOV array
  * @return  number of bytes copied on success, or a negative error code
  */
-ssize_t efa_copy_from_hmem_iov(void **desc, char *buff, int buff_size,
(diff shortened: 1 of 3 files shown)


prov/efa/src/efa_hmem.c
```diff
@@ -283,9 +283,10 @@ ssize_t efa_copy_from_hmem_iov(void **desc, char *buff, int buff_size,
  * @return  number of bytes copied on success, or a negative error code
  */
 ssize_t efa_copy_to_hmem_iov(void **desc, struct iovec *hmem_iov,
-                             int iov_count, char *buff, int buff_size)
+                             size_t iov_count, char *buff, size_t buff_size)
 {
-	int i, ret, bytes_remaining = buff_size, size;
```
shijin-aws, line 288: how can it work before?

shijin-aws (approved):

prov/efa/src/efa_hmem.c
darrylabbate: I believe it's ultimately implementation-defined, but `int` and `size_t` are almost certainly the sizes you'd expect for our use-case. This PR simply minimizes potential risks wrt data type mismatch/integer promotion.

PR #10271, opened 2024-08-02

jiaxiyan:
Combine hmem init functions for different devices.
Avoid checking p2p support when p2p is disabled. FI_HMEM_DISABLE_P2P should take precedence if set.

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -31,9 +31,11 @@ static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_
 static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_domain, enum fi_hmem_iface iface)
 {
 	struct efa_hmem_info *info = &efa_domain->hmem_info[iface];
+	size_t tmp_value;
 
-	/* Fall back to FI_HMEM_SYSTEM initialization logic when p2p is unavailable */
-	if (!info->p2p_supported_by_device)
+	/* Fall back to FI_HMEM_SYSTEM initialization logic when p2p is
+	 * unavailable */
+	if (ofi_hmem_p2p_disabled() || !info->p2p_supported_by_device)
 		iface = FI_HMEM_SYSTEM;
 
 	switch (iface) {
@@ -56,6 +58,12 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 		fi_param_get_size_t(&efa_prov, "runt_size", &info->runt_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &info->min_read_msg_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
+		if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
+			EFA_WARN(FI_LOG_DOMAIN,
+			         "The environment variable FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE was set, "
+			         "but EFA HMEM via Cuda API only supports eager and runting read protocols. "
+			         "The variable will not modify CUDA memory run config.\n");
+		}
 		break;
 	case FI_HMEM_NEURON:
 		info->runt_size = EFA_NEURON_RUNT_SIZE;
@@ -65,12 +73,31 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 		fi_param_get_size_t(&efa_prov, "runt_size", &info->runt_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &info->min_read_msg_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
+		if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
+			EFA_WARN(FI_LOG_DOMAIN,
+			         "The environment variable FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE was set, "
+			         "but EFA HMEM via Neuron API only supports eager and runting read protocols. "
+			         "The variable will not modify CUDA memory run config.\n");
+		}
 		break;
 	case FI_HMEM_SYNAPSEAI:
 		info->runt_size = 0;
 		info->max_medium_msg_size = 0;
 		info->min_read_msg_size = 1;
 		info->min_read_write_size = 1;
+		if (-FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_max_medium_message_size", &tmp_value) ||
+		    -FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &tmp_value) ||
+		    -FI_ENODATA != fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &tmp_value) ||
+		    -FI_ENODATA != fi_param_get_size_t(&efa_prov, "runt_size", &tmp_value)) {
+			EFA_WARN(FI_LOG_DOMAIN,
+			        "One or more of the following environment variable(s) were set: ["
+			        "FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE, "
+			        "FI_EFA_INTER_MIN_READ_MESSAGE_SIZE, "
+			        "FI_EFA_INTER_MIN_READ_WRITE_SIZE, "
+			        "FI_EFA_RUNT_SIZE"
+			        "], but EFA HMEM via Synapse only supports long read protocol. "
+			        "The variable(s) will not modify Synapse memory run config.\n");
+		}
 		break;
 	default:
 		break;
@@ -78,70 +105,24 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 	return 0;
 }
 
-/**
- * @brief          Initialize the efa_hmem_info state for FI_HMEM_SYSTEM
- *
- * @param[in,out]  efa_domain  Pointer to struct efa_domain
- *
- * @return         0
- */
-static int efa_domain_hmem_info_init_system(struct efa_domain *efa_domain)
-{
-	struct efa_hmem_info *info = &efa_domain->hmem_info[FI_HMEM_SYSTEM];
-
-	info->initialized = true;
-	info->p2p_disabled_by_user = false;
-	info->p2p_required_by_impl = false;
-	info->p2p_supported_by_device = true;
-	efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_SYSTEM);
-	return 0;
-}
-
-/**
- * @brief          Initialize the efa_hmem_info state for FI_HMEM_CUDA
- *
- * @param[in,out]  efa_domain  Pointer to struct efa_domain
- *
- * @return         0 on success
- *                 negative libfabric error code on failure
- */
-static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
-{
+static inline void efa_domain_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *info) {
 #if HAVE_CUDA
-	struct efa_hmem_info *info = &efa_domain->hmem_info[FI_HMEM_CUDA];
 	cudaError_t cuda_ret;
 	void *ptr = NULL;
 	struct ibv_mr *ibv_mr;
 	int ibv_access = IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_READ;
-	size_t len = ofi_get_page_size() * 2, tmp_value;
+	size_t len = ofi_get_page_size() * 2;
 	int ret;
 	int dmabuf_fd;
 	uint64_t dmabuf_offset;
 
-	if (!ofi_hmem_is_initialized(FI_HMEM_CUDA)) {
-		EFA_INFO(FI_LOG_DOMAIN, "FI_HMEM_CUDA is not initialized\n");
-		return 0;
-	}
-
 	cuda_ret = ofi_cudaMalloc(&ptr, len);
 	if (cuda_ret != cudaSuccess) {
-		EFA_WARN(FI_LOG_DOMAIN,
-			 "Failed to allocate CUDA buffer: %s\n",
+		info->initialized = false;
+		EFA_WARN(FI_LOG_DOMAIN, "Failed to allocate CUDA buffer: %s\n",
 			 ofi_cudaGetErrorString(cuda_ret));
-		return 0;
+		return;
 	}
-
-	info->initialized = true;
-	info->p2p_disabled_by_user = false;
-
-	/* If user is using libfabric API 1.18 or later, by default EFA provider is permitted to
-	 * use CUDA library to support CUDA memory, therefore p2p is not required.
-	 */
-	if (FI_VERSION_GE(efa_domain->util_domain.fabric->fabric_fid.api_version, FI_VERSION(1,18)))
-		info->p2p_required_by_impl = !hmem_ops[FI_HMEM_CUDA].initialized;
-	else
-		info->p2p_required_by_impl = true;
-
 #if HAVE_EFA_DMABUF_MR
 	ret = cuda_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
 	if (ret == FI_SUCCESS) {
@@ -165,11 +146,10 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
 
 	if (!ibv_mr) {
 		info->p2p_supported_by_device = false;
-		efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_CUDA);
 		EFA_WARN(FI_LOG_DOMAIN,
 			 "Failed to register CUDA buffer with the EFA device, FI_HMEM transfers that require peer to peer support will fail.\n");
 		ofi_cudaFree(ptr);
-		return 0;
+		return;
 	}
 
 	ret = ibv_dereg_mr(ibv_mr);
@@ -178,54 +158,29 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
 		EFA_WARN(FI_LOG_DOMAIN,
 			 "Failed to deregister CUDA buffer: %s\n",
 			 fi_strerror(-ret));
-		return ret;
+		return;
 	}
 
 	info->p2p_supported_by_device = true;
-	efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_CUDA);
-	if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
-		EFA_WARN(FI_LOG_DOMAIN,
-		         "The environment variable FI_EFA_INTER_MAX_MEDIUM_MESSAGE_SIZE was set, "
-		         "but EFA HMEM via Cuda API only supports eager and runting read protocols. "
-				 "The variable will not modify Cuda memory run config.\n");
-	}
+	return;
 
 #endif
-	return 0;
+	return;
 }
 
-/**
- * @brief          Initialize the efa_hmem_info state for FI_HMEM_NEURON
- *
- * @param[in,out]  efa_domain  Pointer to struct efa_domain
- *
- * @return         0 on success
- *                 negative libfabric error code on failure
- */
-static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
-{
+static inline void efa_domain_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *info) {
 #if HAVE_NEURON
-	struct efa_hmem_info *info = &efa_domain->hmem_info[FI_HMEM_NEURON];
 	struct ibv_mr *ibv_mr = NULL;
 	int ibv_access = IBV_ACCESS_LOCAL_WRITE;
 	void *handle;
 	void *ptr = NULL;
-	size_t len = ofi_get_page_size() * 2, tmp_value;
+	size_t len = ofi_get_page_size() * 2;
 	int dmabuf_fd;
 	uint64_t offset;
 	int ret;
 
-	if (!ofi_hmem_is_initialized(FI_HMEM_NEURON)) {
-		EFA_INFO(FI_LOG_DOMAIN, "FI_HMEM_NEURON is not initialized\n");
-		return 0;
-	}
-
 	if (g_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ) {
 		ibv_access |= IBV_ACCESS_REMOTE_READ;
-	} else {
-		EFA_WARN(FI_LOG_DOMAIN,
-			 "No EFA RDMA read support, transfers using AWS Neuron will fail.\n");
-		return 0;
 	}
 
 	ptr = neuron_alloc(&handle, len);
@@ -235,14 +190,10 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
 	 * will move on and leave info->initialized as false.
 	 */
 	if (!ptr) {
+		info->initialized = false;
 		EFA_INFO(FI_LOG_DOMAIN, "Cannot allocate Neuron buffer\n");
-		return 0;
+		return;
 	}
-
-	info->initialized = true;
-	info->p2p_disabled_by_user = false;
-	/* Neuron currently requires P2P */
-	info->p2p_required_by_impl = true;
 
 #if HAVE_EFA_DMABUF_MR
 	ret = neuron_get_dmabuf_fd(ptr, (uint64_t)len, &dmabuf_fd, &offset);
@@ -267,7 +218,7 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
 		         "Failed to register Neuron buffer with the EFA device, "
 		         "FI_HMEM transfers that require peer to peer support will fail.\n");
 		neuron_free(&handle);
-		return 0;
+		return;
 	}
 
 	ret = ibv_dereg_mr(ibv_mr);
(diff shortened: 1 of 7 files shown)


prov/efa/src/efa_hmem.c
```diff
@@ -123,12 +123,14 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
...
-			 ofi_cudaGetErrorString(cuda_ret));
-		return 0;
+	if (!ofi_hmem_p2p_disabled()) {
+		cuda_ret = ofi_cudaMalloc(&ptr, len);
+		if (cuda_ret != cudaSuccess) {
+			EFA_WARN(FI_LOG_DOMAIN,
+				"Failed to allocate CUDA buffer: %s\n",
+				ofi_cudaGetErrorString(cuda_ret));
+			return 0;
+		}
```
shijin-aws, line 133: If p2p is disabled, shouldn't you skip all the subsequent ibv reg as well?

jiaxiyan: Do we just set info->p2p_supported_by_device to false if p2p is disabled? I see https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_hmem.c#L36-L37 fall back to FI_HMEM_SYSTEM initialization logic when p2p is unavailable

wenduwan: At the minimum, we should fix the obvious behavior on L137 - in this case the user explicitly disabled p2p should we should respect that.

Overall I suggest we audit CUDA/Neuron/Synapseai as a whole - if p2p is disabled by the user, we should skip the p2p checks entirely. CUDA is not special in this case.

darrylabbate:
We need to consider the endpoint option `FI_OPT_FI_HMEM_P2P`. The man page states the env var `FI_HMEM_DISABLE_P2P` should take precedence if set. So I'm thinking `efa_domain_hmem_validate_p2p_opt()` also needs an extra piece of logic to reflect this (this is called during `fi_setopt()`).

#8351 and #8529 have relevant discussions

prov/efa/src/efa_hmem.c
```diff
@@ -317,7 +335,11 @@ static int efa_domain_hmem_info_init_synapseai(struct efa_domain *efa_domain)
 	}
 
 	info->initialized = true;
-	info->p2p_disabled_by_user = false;
+	if (ofi_hmem_p2p_disabled()) {
+		info->p2p_disabled_by_user = true;
+	} else {
+		info->p2p_disabled_by_user = false;
+	}
```
darrylabbate, line 342: You can probably place this at the top of each function

```suggestion
	info->p2p_disabled_by_user = ofi_hmem_p2p_disabled();
```

prov/efa/src/efa_hmem.c
```diff
@@ -142,6 +134,23 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
 	else
 		info->p2p_required_by_impl = true;
 
+	if (ofi_hmem_p2p_disabled()) {
+		info->p2p_disabled_by_user = true;
+		info->p2p_supported_by_device = false;
```
darrylabbate, line 139: Doesn't make sense to `p2p_supported_by_device` to false just because the user disables P2P. IOW, the user doesn't dictate the device's P2P capabilities; only whether or not they wish to utilize them.

prov/efa/src/efa_hmem.c
```diff
@@ -142,6 +134,23 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
...
+		efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_CUDA);
```
darrylabbate, line 140: Doesn't really make sense to initialize the protocol thresholds based on the condition of disabled P2P - just move the call from further down the function above the `ofi_hmem_p2p_disabled()` and make the init unconditional.

prov/efa/src/efa_hmem.c
```diff
@@ -142,6 +134,23 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
...
+		return 0;
+	}
+
+	cuda_ret = ofi_cudaMalloc(&ptr, len);
+	if (cuda_ret != cudaSuccess) {
+		info->initialized = false;
+		info->p2p_required_by_impl = true;
```
darrylabbate, line 147: This also doesn't make sense. Failed allocation of CUDA memory doesn't imply any kind of P2P restriction. This feels like overly defensive coding. The P2P requirement for each iface is more of a fixed attribute.

prov/efa/src/efa_hmem.c
```diff
@@ -228,6 +237,17 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
 		return 0;
 	}
 
+	/* Neuron currently requires P2P */
+	info->p2p_required_by_impl = true;
+
+	if (ofi_hmem_p2p_disabled()) {
+		info->initialized = true;
```
darrylabbate, line 244: Likewise, `initialized` shouldn't depend on whether or not the user disabled P2P

prov/efa/src/efa_hmem.c
jiaxiyan: The reason I set these values for info is to do an early return when `ofi_hmem_p2p_disabled` so I can skip all the p2p checks, including ofi_cudaMalloc/neuron_alloc and ibv_reg_mr/ibv_reg_dmabuf_mr. 
Are you suggesting leaving `p2p_supported_by_device` unset?

darrylabbate: There's nothing wrong with an early exit when `ofi_hmem_p2p_disabled()` returns true. `p2p_supported_by_device` simply shouldn't depend on that condition. In the future, there may some other logic in the EFA provider that relies on `p2p_supported_by_device`; independent of the user's request to disable P2P HMEM operations.

prov/efa/src/efa_hmem.c
jiaxiyan: Without initializing `p2p_supported_by_device`, we will always fall back to FI_HMEM_SYSTEM initialization logic if we move `efa_domain_hmem_info_init_protocol_thresholds` above. That is not what we expect.

prov/efa/src/efa_hmem.c
```diff
@@ -377,11 +392,14 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
 	 *     PREFERED means a provider should prefer P2P if it is available.
 	 *
 	 * These options does not require that p2p is supported by device,
-	 * nor do they prohibit that p2p is reqruied by implementation. Therefore
-	 * they are always supported.
+	 * nor do they prohibit that p2p is required by implementation. 
+	 * Therefore FI_HMEM_P2P_PREFERRED is always supported. 
+	 * FI_HMEM_P2P_ENABLED is supported unless p2p is disabled.
```
darrylabbate, line 397: > Therefore FI_HMEM_P2P_PREFERRED is always supported.

I don't think this is completely accurate since we're returning `-FI_EOPNOTSUPP` when P2P is disabled

jiaxiyan: Do we want to add 
```
if (OFI_UNLIKELY(ofi_hmem_p2p_disabled()))
    return -FI_EOPNOTSUPP;
```
for FI_HMEM_P2P_PREFERRED too then?

prov/efa/src/efa_hmem.c
```diff
@@ -142,6 +133,23 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
 	else
 		info->p2p_required_by_impl = true;
 
+	info->p2p_disabled_by_user = ofi_hmem_p2p_disabled();
+
+	if (info->p2p_disabled_by_user) {
+		/* Don't need to check p2p support */
+		efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_CUDA);
```
wenduwan, line 140: Given this change, we should consider abstracting away the p2p check into separate function(s). With that out of the way, we can simplify the hmem info init logic:
```
info->p2p_required_by_impl = whatever;
info->p2p_disabled_by_user = IS_SYSTEM ? false : ofi_hmem_p2p_disabled();
info->p2p_supported_by_device = info->p2p_disabled_by_user ? false : is_p2p_supported(iface);
...
```

prov/efa/src/efa_hmem.c
```diff
@@ -142,6 +133,23 @@ static int efa_domain_hmem_info_init_cuda(struct efa_domain *efa_domain)
...
+		return 0;
+	}
+
+	cuda_ret = ofi_cudaMalloc(&ptr, len);
+	if (cuda_ret != cudaSuccess) {
```
wenduwan, line 145: If possible, I suggest adding unit tests to make sure:
- If user disables p2p, there is no `ibv_reg_mr` calls during domain open.
- FI_HMEM_SYSTEM always supports p2p and is never disabled.

prov/efa/src/efa_hmem.c
```diff
@@ -276,70 +227,81 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
+	case FI_HMEM_CUDA:
+		efa_domain_hmem_info_is_p2p_supported_by_cuda(info);
+		return info->p2p_supported_by_device;
+	case FI_HMEM_NEURON:
+		efa_domain_hmem_info_is_p2p_supported_by_neuron(info);
+		return info->p2p_supported_by_device;
+	case FI_HMEM_SYNAPSEAI:
+		return true;
+	default:
+		return false;
```
wenduwan, line 251: This does not look correct. FI_HMEM_SYSTEM should always support p2p

prov/efa/src/efa_hmem.c
```diff
@@ -276,70 +227,81 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
+		return;
 	}
 
 	info->initialized = true;
-	info->p2p_disabled_by_user = false;
-	/* SynapseAI currently requires P2P */
-	info->p2p_required_by_impl = true;
-	info->p2p_supported_by_device = true;
-	efa_domain_hmem_info_init_protocol_thresholds(efa_domain, FI_HMEM_SYNAPSEAI);
+	info->p2p_disabled_by_user = ofi_hmem_p2p_disabled();
```
wenduwan, line 281: p2p should never be disabled for FI_HMEM_SYSTEM

prov/efa/src/efa_hmem.c
```diff
@@ -276,70 +227,81 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
+	info->p2p_supported_by_device = info->p2p_disabled_by_user ?
+		false : efa_domain_hmem_info_is_p2p_supported_by_device(efa_domain, iface);
+	if (iface == FI_HMEM_SYNAPSEAI)
+		info->p2p_supported_by_device = true;
```
wenduwan, line 285: Should remove.

prov/efa/src/efa_hmem.c
```diff
@@ -276,70 +227,81 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
+	if (!info->p2p_supported_by_device)
```
wenduwan, line 286: We only need this warning if p2p is not disabled.

prov/efa/src/efa_hmem.c
```diff
@@ -31,9 +31,11 @@ static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_
...
 {
 	struct efa_hmem_info *info = &efa_domain->hmem_info[iface];
+	size_t tmp_value;
 
-	/* Fall back to FI_HMEM_SYSTEM initialization logic when p2p is unavailable */
-	if (!info->p2p_supported_by_device)
+	/* Fall back to FI_HMEM_SYSTEM initialization logic when p2p is
+	 * unavailable */
+	if (ofi_hmem_p2p_disabled() || !info->p2p_supported_by_device)
 		iface = FI_HMEM_SYSTEM;
```
wenduwan, line 39: We only need to check `p2p_supported_by_device` and return 0 if it's false.

prov/efa/src/efa_hmem.c
```diff
@@ -377,11 +339,14 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
...
-	 * nor do they prohibit that p2p is reqruied by implementation. Therefore
-	 * they are always supported.
+	 * nor do they prohibit that p2p is required by implementation. 
+	 * Therefore they are always supported unless p2p is disabled.
 	 */
 	case FI_HMEM_P2P_PREFERRED:
 	case FI_HMEM_P2P_ENABLED:
+		if (OFI_UNLIKELY(ofi_hmem_p2p_disabled()))
+			return -FI_EOPNOTSUPP;
+
```
wenduwan, line 349: How can we end up here?

prov/efa/src/efa_hmem.c
```diff
@@ -365,7 +327,7 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
 
 	switch (p2p_opt) {
 	case FI_HMEM_P2P_REQUIRED:
-		if (!info->p2p_supported_by_device)
+		if (OFI_UNLIKELY(ofi_hmem_p2p_disabled()) || !info->p2p_supported_by_device)
```
wenduwan, line 330: Shouldn't add ofi_hmem_p2p_disabled check.

FI_HMEM_SYSTEM should always support p2p.

prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -55,3 +55,68 @@ void test_efa_hmem_info_update_neuron()
...
+        assert_true(efa_domain->hmem_info[FI_HMEM_NEURON].initialized);
+        assert_false(efa_domain->hmem_info[FI_HMEM_NEURON].p2p_supported_by_device);
+#endif
+#if HAVE_CUDA
+        hmem_ops[FI_HMEM_CUDA].initialized = cuda_initialized_orig;
+
+        assert_true(efa_domain->hmem_info[FI_HMEM_CUDA].p2p_disabled_by_user);
+        assert_true(efa_domain->hmem_info[FI_HMEM_CUDA].initialized);
+        assert_false(efa_domain->hmem_info[FI_HMEM_CUDA].p2p_supported_by_device);
+#endif
```
wenduwan, line 120: We should split this into 2 tests for cuda and neuron respectively.

prov/efa/src/efa_hmem.c
jiaxiyan: > Fall back to FI_HMEM_SYSTEM initialization logic when p2p is unavailable

Is this not true any more?

prov/efa/src/efa_hmem.c
jiaxiyan: SYNAPSEAI always support p2p, so this should be true irrelevant of p2p disabled status.

prov/efa/src/efa_hmem.c
jiaxiyan: This function `efa_rdm_ep_set_fi_hmem_p2p_opt` is called with `EFA_HMEM_IFACE_FOREACH_NON_SYSTEM`.
This is checking the flag FI_HMEM_DISABLE_P2P.

prov/efa/src/efa_hmem.c
jiaxiyan: Uses could set both FI_HMEM_DISABLE_P2P and FI_HMEM_P2P_PREFERRED/FI_HMEM_P2P_ENABLED, which is not supported

prov/efa/src/efa_hmem.c
jiaxiyan: I reset the fields of FI_HMEM_SYSTEM at the end of this function, or I can add extra logic here.

prov/efa/src/efa_hmem.c
darrylabbate: Good catch. I still feel we need to honor this constraint upon `fi_setopt()`.

> The `FI_HMEM_DISABLE_P2P` environment variable discussed in `fi_mr(3)` takes precedence over this setopt option

prov/efa/src/efa_hmem.c
wenduwan: This function should only initialize 1 iface at a time. I don't see the value of the fallback logic.

prov/efa/src/efa_hmem.c
wenduwan: I think those options are mutually exclusive, isn't that the case?

prov/efa/src/efa_hmem.c
wenduwan: This is not necessary. If the user disables p2p on synapseai platform, then we should not attempt to use synapseai p2p, so it does not matter whether the device supports it or not.

This is the same for CUDA/Neuron - even if the device supports p2p, we still set p2p_supported_by_device = false it the user disables it.

prov/efa/src/efa_hmem.c
darrylabbate: `FI_HMEM_DISABLE_P2P` is also the name of the env var that controls `ofi_hmem_p2p_disabled()`. So the user can have it set (inadvertently or otherwise) while also passing whatever they want for `fi_setopt()`.

prov/efa/src/efa_hmem.c
darrylabbate: The function still initializes the single iface (`efa_domain->hmem_info[iface]` assigned to `info`); The `iface` parameter is simply overwritten here after for convenience to trigger a different iface's logic.

prov/efa/src/efa_hmem.c
wenduwan: So this function is on EP init path, after domain init - in other words,`info->p2p_supported_by_device` was already set at this point. We have 2 invariants:
- For non system HMEM, `info->p2p_supported_by_device` is always false if `ofi_hmem_p2p_disabled()` is true
- For system HMEM, `info->p2p_supported_by_device` is always true.

This means for non system HMEM `OFI_UNLIKELY(ofi_hmem_p2p_disabled()) || !info->p2p_supported_by_device` is equivalent to `!info->p2p_supported_by_device`

darrylabbate: Actually, Jessie's correct here. We simply return `-FI_EINVAL` for `FI_HMEM_SYSTEM` upon setopt. We need to find a way to avoid having to iterate through each iface for stuff like this in the future.

prov/efa/src/efa_hmem.c
```diff
@@ -377,11 +337,14 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
...
-	 * they are always supported.
+	 * nor do they prohibit that p2p is required by implementation. 
+	 * Therefore they are always supported unless p2p is disabled.
 	 */
 	case FI_HMEM_P2P_PREFERRED:
 	case FI_HMEM_P2P_ENABLED:
+		if (OFI_UNLIKELY(ofi_hmem_p2p_disabled()))
+			return -FI_EOPNOTSUPP;
+
 		info->p2p_disabled_by_user = false;
```
wenduwan, line 348: I don't think we need to handle such case regardless of info->p2p_disabled_by_peer.

>FI_HMEM_P2P_ENABLED: Peer to peer support may be used by the provider to handle FI_HMEM transfers, and which transfers are initiated using peer to peer is subject to the provider implementation.
FI_HMEM_P2P_PREFERRED: Peer to peer support should be used by the provider for all transfers if available, but the provider may choose to copy the data to initiate the transfer if peer to peer support is unavailable.


Based on the definition, the provider can decide which memcpy protocol should be used. In other words, if p2p is disabled by the user, EFA provider can decide to use a different protocol.

Also, we shouldn't `info->p2p_disabled_by_user = false;` here, as @darrylabbate already mentioned, the `FI_HMEM_DISABLE_P2P` option has a higher priority. It's much easier to change an environment variable than fi_setopt code.

prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -49,9 +49,109 @@ void test_efa_hmem_info_update_neuron(struct efa_resource **state)
...
         skip();
 }
+
+void test_efa_hmem_info_disable_p2p_neuron()
+{
+        skip();
+}
 #endif /* HAVE_NEURON */
+
+#if HAVE_CUDA
```
wenduwan, line 110: A topic for later - do we really need `HAVE_CUDA/HAVE_NEURON`? If we are calling `ofi_cudaMalloc/ofi_neuronMalloc`, then we don't need to check CUDA/Neuron availability again.

prov/efa/src/efa_hmem.c
wenduwan: `efa_domain_hmem_info_init_protocol_thresholds` is called for each iface - keeping this logic means FI_HMEM_SYSTEM could be initialized more than once. I don't think that is necessary.

wenduwan (approved):

prov/efa/src/efa_hmem.c
shijin-aws: I somehow agree with wenduo, I think we should only initialize system iface once. For non-system iface and p2p disabled/not supported, we can just return instead of doing things for system iface again

prov/efa/src/efa_hmem.c
```diff
@@ -276,70 +227,79 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
-				 "The variable will not modify Neuron memory run config.\n");
-	}
-
+	return;
 #endif
-	return 0;
+	return;
+}
+
+static inline bool efa_domain_hmem_info_is_p2p_supported_by_device(struct efa_domain *efa_domain, enum fi_hmem_iface iface) {
```
shijin-aws, line 239: This function name gives reader an impression that it can be called multiple times. But I do want it is a one-time call. So you either rename the function make it a light weight call that only return the `info->p2p_supported_by_device` related stuff, and do the heavy-weight check call separately.

prov/efa/src/efa_hmem.c
shijin-aws: `p2p_supported_by_device` still needs to be set for the p2p check later right? If we remove it here, do we have it somewhere else?

prov/efa/src/efa_hmem.c
darrylabbate: No. The `iface` overwrite has no effect on which iface is getting initialized since the `info` initialization at the top of the function is independent of it. `FI_HMEM_SYSTEM` would not be initialized again if e.g. the function is called with `FI_HMEM_CUDA` and P2P is unavailable.


```c
// Fetch intended hmem_info entry; i.e. the "iface" in this context
struct efa_hmem_info *info = &efa_domain->hmem_info[iface];

...

// Overwrite the iface var for the purpose of dispatching to FI_HMEM_SYSTEM
// logic in the switch-case construct below
if (ofi_hmem_p2p_disabled() || !info->p2p_supported_by_device)
	iface = FI_HMEM_SYSTEM;
```

shijin-aws: I see, you are right, it is actually initializing cuda's iface while using system's threshold.

wenduwan: I see. The confusion originates from the comment above
> // Overwrite the iface var for the purpose of dispatching to FI_HMEM_SYSTEM
// logic in the switch-case construct below

A more concise and less confusing logic will be
```
struct efa_hmem_info *info = &efa_domain->hmem_info[iface];

// Set default values
info->runt_size = 0;
info->max_medium_msg_size = EFA_DEFAULT_INTER_MAX_MEDIUM_MESSAGE_SIZE;
info->min_read_msg_size = EFA_DEFAULT_INTER_MIN_READ_MESSAGE_SIZE;
info->min_read_write_size = EFA_DEFAULT_INTER_MIN_READ_WRITE_SIZE;
fi_param_get_size_t(&efa_prov, "runt_size", &info->runt_size);
fi_param_get_size_t(&efa_prov, "inter_max_medium_message_size", &info->max_medium_msg_size);
		fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &info->min_read_msg_size);
fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);

if (!info->p2p_supported_by_device) return;

switch (iface) {
case ...:
        ... // override params and check illegal parameters
}

```

prov/efa/src/efa_hmem.c
```diff
@@ -56,6 +58,12 @@ static int efa_domain_hmem_info_init_protocol_thresholds(struct efa_domain *efa_
 		fi_param_get_size_t(&efa_prov, "runt_size", &info->runt_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_message_size", &info->min_read_msg_size);
 		fi_param_get_size_t(&efa_prov, "inter_min_read_write_size", &info->min_read_write_size);
+		if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &tmp_value)) {
```
darrylabbate, line 61: Just a tip (no need to change this here): you can avoid declaring a temp value for throwaway vars like these by passing a compound literal as an "anonymous" value.

```suggestion
		if (-FI_ENODATA != fi_param_get(&efa_prov, "inter_max_medium_message_size", &(size_t){0})) {
```

prov/efa/src/efa_hmem.c
```diff
@@ -409,42 +372,18 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
  */
 int efa_domain_hmem_info_init_all(struct efa_domain *efa_domain)
 {
-	int ret, err;
+	int ret = 0;
+	enum fi_hmem_iface ifaces[4] = {FI_HMEM_SYSTEM, FI_HMEM_CUDA,
+	                                FI_HMEM_NEURON, FI_HMEM_SYNAPSEAI};
```
darrylabbate, line 377: Use the predefined static array from `efa_hmem.h`

https://github.com/ofiwg/libfabric/blob/5b08723bf29a914d045554c0942e6d058a58397e/prov/efa/src/efa_hmem.h#L16-L22

prov/efa/src/efa_hmem.c
```diff
@@ -409,42 +372,18 @@ int efa_domain_hmem_validate_p2p_opt(struct efa_domain *efa_domain, enum fi_hmem
...
-	}
-
-	err = efa_domain_hmem_info_init_synapseai(efa_domain);
-	if (err) {
-		ret = err;
-		EFA_WARN(FI_LOG_DOMAIN, "Failed to populate the Synapseai hmem_info struct! err: %d\n",
-			 err);
+	for (int i = 0; i < sizeof(ifaces) / sizeof(enum fi_hmem_iface); ++i) {
+		efa_domain_hmem_info_init_iface(efa_domain, ifaces[i]);
 	}
```
darrylabbate, line 387: There's also a convenience macro for this in `efa_hmem.h`

https://github.com/ofiwg/libfabric/blob/5b08723bf29a914d045554c0942e6d058a58397e/prov/efa/src/efa_hmem.h#L10-L11

prov/efa/src/efa_hmem.c
jiaxiyan: Why would we call this functions multiple times? We would just get info->p2p_supported_by_device right?

prov/efa/src/efa_hmem.c
jiaxiyan: We set it in the function efa_domain_hmem_info_is_p2p_supported_by_device
```
info->p2p_supported_by_device = !info->p2p_disabled_by_user && efa_domain_hmem_info_is_p2p_supported_by_device(efa_domain, iface);
```

prov/efa/src/efa_hmem.c
shijin-aws: In convention, `is_p2p_supported_by_device`  should be a simple lightweight boolean get, and it is expected that we can call this function cheaply to get device support status. I don't want the heavy check code as part of it.

prov/efa/src/efa_hmem.c
```diff
@@ -276,70 +227,69 @@ static int efa_domain_hmem_info_init_neuron(struct efa_domain *efa_domain)
...
-		return 0;
+	if (!ofi_hmem_is_initialized(iface)) {
+		EFA_INFO(FI_LOG_DOMAIN, "%s is not initialized\n",
+		         fi_tostr(&iface, FI_TYPE_HMEM_IFACE));
+		return;
 	}
 
-	if (!(g_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
+	if ((iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_NEURON) &&
+	    !(g_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ)) {
```
shijin-aws, line 257: Nit: you can reuse https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_device.c#L243 `efa_device_support_rdma_read`

shijin-aws (approved):

prov/efa/test/efa_unit_test_mocks.h
```diff
@@ -103,6 +103,10 @@ struct efa_unit_test_mocks
 	void *(*neuron_alloc)(void **handle, size_t size);
 #endif
 
+#if HAVE_CUDA
+	cudaError_t *(*ofi_cudaMalloc)(void **ptr, size_t size);
```
shijin-aws, line 107: This doesn't seem correct, it shouldn't have `*` after `cudaErrot_t`

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -304,6 +307,23 @@ void *efa_mock_neuron_alloc_return_null(void **handle, size_t size)
...
+void *efa_mock_neuron_alloc_return_mock(void **handle, size_t size)
+{
+	return (void *) mock();
+}
+#endif
+
+#if HAVE_CUDA
+cudaError_t *__wrap_ofi_cudaMalloc(void **ptr, size_t size)
+{
+	return g_efa_unit_test_mocks.ofi_cudaMalloc(ptr, size);
```
shijin-aws, line 320: This is also wrong, no `*` before `__wrap_ofi_cudaMalloc`

prov/efa/test/efa_unit_test_mocks.c
```diff
@@ -304,6 +307,23 @@ void *efa_mock_neuron_alloc_return_null(void **handle, size_t size)
...
+}
+
+cudaError_t *efa_mock_ofi_cudaMalloc_return_mock(void **ptr, size_t size)
+{
+	return (cudaError_t *) mock();
```
shijin-aws, line 325: Why return `(cudaError_t *)` while your return type is `cudaError_t`

shijin-aws:
It seems you mixed the `*` function pointer and the real return type of the mocked function. Please fix it

prov/efa/test/efa_unit_test_hmem.c
```diff
@@ -49,9 +49,109 @@ void test_efa_hmem_info_update_neuron(struct efa_resource **state)
...
+        ret = fi_fabric(resource->info->fabric_attr, &resource->fabric, NULL);
+        assert_int_equal(ret, 0);
+
+        neuron_initialized_orig = hmem_ops[FI_HMEM_NEURON].initialized;
+        hmem_ops[FI_HMEM_NEURON].initialized = true;
+        efa_device_caps_orig = g_device_list[0].device_caps;
+        g_device_list[0].device_caps |= EFADV_DEVICE_ATTR_CAPS_RDMA_READ;
+        /* neuron_alloc should not be called when p2p is disabled */
+        g_efa_unit_test_mocks.neuron_alloc = &efa_mock_neuron_alloc_return_mock;
+
```
shijin-aws, line 82: for safe, add a `expect_function_call(efa_mock_neuron_alloc_return_mock);` here. Same thing for `ofi_cudaMalloc` mock

shijin-aws (changes requested):

prov/efa/test/efa_unit_test_hmem.c
jiaxiyan: Why adding `expect_function_call` when we expect `efa_mock_neuron_alloc_return_mock` not being called?

shijin-aws: Sorry, I missed it. Then how do we verify it is not called?

jiaxiyan: `efa_mock_neuron_alloc_return_mock` just returns `mock()`, which will fail when `neuron_alloc` is called

shijin-aws: Hmmm this seems odd - can you try if `expect_function_calls(efa_mock_neuron_alloc_return_mock, 0)` works here?

jiaxiyan: cmocka does not support 0 for expect_function_calls 
 https://github.com/clibs/cmocka/blob/master/src/cmocka.c#L1111

Right now efa_mock_neuron_alloc_return_mock will print 
 ```
No entries for symbol efa_mock_neuron_alloc_return_mock.
prov/efa/test/efa_unit_test_mocks.c:313: error: Could not get value to mock function efa_mock_neuron_alloc_return_mock
There were no previously returned mock values for this test.
```
if neuron_alloc is called. 

Or I can do a `assert_true(false)` in efa_mock_neuron_alloc_return_mock instead.

shijin-aws: I am ok with just adding some comment to explain why do you do a mock and why it can make sure the function is not called

shijin-aws (approved):

wenduwan (approved):

PR #10273, opened 2024-08-03
contrib/aws: add coverage for Amazon Linux 2023

wenduwan:
Amazon Linux 2023 has 6.x kernel which is not covered in CI.

As a result we will consume more hpc6a capacity. To avoid capacity issues, switch tcp/socket tests to a more abundant instance type, e.g. c6gn.16xlarge.

### DIFF
--- a/contrib/aws/Jenkinsfile
+++ b/contrib/aws/Jenkinsfile
@@ -194,11 +194,13 @@ pipeline {
 
                     // Single Node Tests - EFA
                     stages["1_g4dn_alinux2-efa"] = get_test_stage("1_g4dn_alinux2_efa", env.BUILD_TAG, "alinux2", "g4dn.8xlarge", 1, "us-east-1", "libfabric_pr_test.yaml", addl_args_pr)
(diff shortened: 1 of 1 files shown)


shijin-aws:
@wenduwan can we move tcp/sockets to non-efa instances? they don't need it, right?

wenduwan:
@shijin-aws I forgot but did we have a special reason to use hpc6a before? I don't think we need EFA either.

shijin-aws:
I don't think there is a special reason, probably because the yaml config enables EFA by default and if you use an non-efa instance you need to pass additional cmd args

wenduwan:
@shijin-aws Cool. Switched to c6g instead. Could you kick off CI? I don't have permission to run this change.

shijin-aws:
> I don't have permission to run this change.

I don't understand, shouldn't your PR trigger the new test as is?

wenduwan:
It does not.

shijin-aws:
Then how your earlier push go through AWS CI?

shijin-aws:
bot:aws:retest

wenduwan:
bot:aws:retest

wenduwan:
bot:aws:retest

Issue #6124, opened 2020-07-23
FI_HMEM support: ofi_cudaMemcpy() using default stream can cause dead lock.


wzamazon:
I encountered this issue when I was testing the FI_HMEM support of the EFA provider with NCCL, and I believe this problem will apply to any provider that use ```ofi_cudaMemcpy()``` to copy data to GPU during message transfer.

The core issue here is that ```ofi_cudaMemcpy()``` use ```cudaMemcpy()``` to deliver data, which is on default stream. Meanwhile, for efficiency, some application/library might launch a kernel that wait for libfabric to deliver data and process the data. If the kernel is also on default stream, there will be a dead lock.

Take NCCL as an example, 

When NCCL uses cuda buffer directly, it will allocate bounce buffer using GPU memory.
When a NCCL routine is called with default stream, it will launch a cuda kernel that is on that stream. 
That cuda kernel is waiting for data to be copied into bounce buffer, then copy the data from bounce buffer to target buffer and other GPU.
If an libfabric provider use ```ofi_cudaMemcpy```to deliver data,  because that function uses default stream, there will be  a dead lock.

Note the problem does not limit to NCCL. In principal, any message transfer application/library (MPI for example) that want to launch a kernel on a default stream that wait for libfabric to deliver data will have this issue. And they might want to do that for efficiency.

wzamazon:
To reproduce the problem, you need:

1. An application that uses NCCL. I used the example C program on NCCL website:
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/examples.html
The last one is what I used.

2. NCCL: I tried 2.4.8 and 2.5.6 both have the issue

3. A NCCL network plugin. that support cuda pointer

4. A libfabric that support FI_HMEM capability.

Note, one might need to adjust the environment variable  NCCL_NET_GDR_LEVEL to make NCCL use cuda buffer directly.

rajachan:
Cc: @iziemba. Did you run into this with any of your testing? Using the default stream for the copies seem to be an issue and we might have to change the ofi_copy_cuda_* implementation.

shefty:
It sounds like the API may need to be updated to give the app control over where copies occur for a given MR.

iziemba:
I have not seen deadlock issues. So is the following happening?

[...]

PR #10303, opened 2024-08-12
Add LPP provider

tstruk:
Add libfabric PCIe Provider (LPP) is a provider.
LPP is a a provider, which runs on top of Gigaio FabreX(™) fabric.


shijin-aws:
Should you rename your provider to reflect the NIC vendor? I think many providers like EFA utilize PCIe to realize Peer direct features.

tstruk:
> Should you rename your provider to reflect the NIC vendor? I think many providers like EFA utilize PCIe to realize Peer direct features.

We have been using this name for a long time, and it would require substantial effort to change it at this point.
Since there is no clash with any other provider name, or no conflict with naming conventions we would like to keep it that way.

tstruk:
Where can I see the results from AWS CI or Jenkins?
The builds passed fine for me in Github Actions https://github.com/tstruk/libfabric/actions/runs/10357960203/job/28671102705
Looks like the results from these tools are not available to the public, are they?

darrylabbate:
@tstruk For AWS CI:

```
Command make -j failed with error:
Makefile:3742: Extraneous text after `endif' directive
Makefile:3742: Extraneous text after `endif' directive
```

There may be more fatal errors (output is large), so I'll comment what I see

tstruk:
Thank you @darrylabbate I have pushed a new version. Let's wait and see if it will work any better.

prov/lpp/src/lpp_hmem.c
```diff
@@ -0,0 +1,97 @@
...
+		*iface = FI_HMEM_ROCR;
+	} else if (lpp_domain->enabled_iface == FI_HMEM_CUDA
+		&& cuda_is_addr_valid(ptr, NULL, NULL)) {
+
+		#if HAVE_CUDA
+		//TODO: not entirely clear if this is required for write-only MRs.
+		// Might wan to unset later as this can affect performance.
+		CUresult cures;
+		int flag = 1;
+		cures = ofi_cuPointerSetAttribute(&flag,
```
shijin-aws, line 69: We have problem with this sync memops set when running with NCCL >= 2.19. Basically it's not safe to set `CU_POINTER_ATTRIBUTE_SYNC_MEMOPS` for NCCL application

shijin-aws: I guess you set this attribute to enforce a synchronous cuda memcpy inside libfabric, while cuda memcpy is not allowed inside Libfabric either for NCCL application - https://github.com/ofiwg/libfabric/issues/6124

aingerson:
Throwing the Intel CI failures here as well - two different builds failed and had two different errors:
```
12:58:46  prov/lpp/src/hmem_cuda.o: In function `hmem_cuda_init':
12:58:46  .../libfabric/fabtests/prov/lpp/src/hmem_cuda.c:112: undefined reference to `count_of'
12:58:46  .../libfabric/fabtests/prov/lpp/src/hmem_cuda.c:113: undefined reference to `count_of'
12:58:46  .../libfabric/fabtests/prov/lpp/src/hmem_cuda.c:114: undefined reference to `count_of'
12:58:46  .../libfabric/fabtests/prov/lpp/src/hmem_cuda.c:115: undefined reference to `count_of'
12:58:46  .../libfabric/fabtests/prov/lpp/src/hmem_cuda.c:116: undefined reference to `count_of'
```

```
12:59:47  /usr/bin/ld: cannot find /usr/lib64/libatomic.so.1.2.0
12:59:47  collect2: error: ld returned 1 exit status
```

j-xiong:
`count_of` was removed from fabric.h recently (see https://github.com/ofiwg/libfabric/pull/10250/commits/6ac82423dd98c622f9210fb77738a5b5816e47bc#diff-6c89f68bf67f752f1c4034d054016bb948ce8f178cb6b3550b910b51cc4bc045). Please add the definition to your test code.

include/ofi_prov.h
```diff
@@ -304,6 +304,21 @@ UCX_INI ;
 #  define UCX_INIT NULL
 #endif
 
+#if defined(_WIN32) && (HAVE_EFA)
```
j-xiong, line 307: Is LPP somewhat related to EFA, or is it a typo?

prov/lpp/include/lpp_cntr.h
```diff
@@ -0,0 +1,72 @@
...
+	klpp_id_t		id;
+	int			valid;
+	int			index;
+	_Atomic uint64_t	*counter;
+	_Atomic	uint64_t	*errcounter;
+	ofi_atomic64_t		last_counter;
+	ofi_atomic64_t		last_errcounter;
+	struct dlist_entry	ep_list;
+	ofi_mutex_t		lock;
+} Lpp_cntr_t;
```
j-xiong, line 57: Please use struct directly instead of using typedef. Same for many other instances.

prov/lpp/include/lpp.h
```diff
@@ -0,0 +1,231 @@
...
+//
+// Capabilities may be grouped into two general categories: primary and
+// secondary. Primary capabilities must explicitly be requested by an
+// application, and a provider must enable support for only those primary
+// capabilities which were selected. Secondary capabilities may optionally be
+// requested by an application. If requested, a provider must support the
+// capability or fail the fi_getinfo request (FI_ENODATA). A provider may
+// optionally report non-selected secondary capabilities if doing so would not
+// compromise performance or security.
+//
```
j-xiong, line 111: Please use `/*` `*/` for multiline comments. Same for many other places.

prov/lpp/include/lpp_tagged.h
```diff
@@ -0,0 +1,58 @@
...
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef _LPP_TAGGED_H_
+#define _LPP_TAGGED_H_
+
+ssize_t lpp_fi_trecv(struct fid_ep *ep, void *buf, size_t len, void *desc,
+	fi_addr_t src_addr, uint64_t tag, uint64_t ignore, void *context);
```
j-xiong, line 37: Align the second line with the 'struct' in the first line. Same for other places.

prov/lpp/src/lpp_atomic.c
```diff
@@ -0,0 +1,875 @@
...
+static void *atomic_res_buf_alloc(size_t size)
+{
+	void *buf;
+
+	// TODO: buffer pool?
+	if (posix_memalign(&buf, KLPP_MQ_CELL_SIZE, size) != 0) {
+		return NULL;
+	} else {
+		return buf;
+	}
```
j-xiong, line 99: don't use `'{`  and `}` when the body has a single line. Same for other places.

prov/lpp/src/lpp_getinfo.c
```diff
@@ -0,0 +1,264 @@
...
+	for (size_t i = 0; i < globmatch.gl_pathc; i++) {
+		klpp_fd = open(globmatch.gl_pathv[i], O_RDWR);
+		if (klpp_fd < 0) {
+			FI_WARN(&lpp_prov, FI_LOG_FABRIC, "failed to open %s (errno %d: %s)\n",
+			    globmatch.gl_pathv[i], errno, strerror(errno));
+			continue;
+		}
+		if (klpp_getdevice(klpp_fd, &klpp_devinfo) < 0) {
+			FI_WARN(&lpp_prov, FI_LOG_FABRIC, "failed to query KLPP device %ld\n", i);
+			continue;
```
j-xiong, line 244: should klpp_fd be closed here?

prov/lpp/src/lpp_nosys.c
```diff
@@ -0,0 +1,382 @@
...
+#include <stdio.h>
+#include <rdma/fabric.h>
+#include <rdma/fi_errno.h>
+#include <ofi_enosys.h>
+#include <lpp.h>
+
+//
+// Common
+//
+int lpp_fi_no_bind(struct fid *fid, struct fid *bfid, uint64_t flags)
```
j-xiong, line 43: Can use the functions defined in `src/enosys.c` instead of defining here.

include/ofi_prov.h
tstruk: That's a typo.

prov/lpp/src/lpp_hmem.c
BKitor: Our device uses GPUDirectRDMA, which avoid the need for cudaMemcpy. We're setting the SYNC_MEMOPS attr out of an abundance of caution, as outlined in GpuDirectRDMA docs (https://docs.nvidia.com/cuda/gpudirect-rdma/#synchronization-and-memory-ordering).

Poking through the code, the EFA rdm calls `efa_rdm_attempt_to_sycm_memops` before a lot of transfers. How are you guys handing this issue then? Is there some nuance to when 'efa_mr->needs_sync` is set?

As for NCCL >=2.19, is this related to the changes with device memory allocation? I ran into a problem testing out newer NCCL version, but was able to mitigate it with the `NCCL_CUMEM_ENABLE=0` env var. 

Also, now that I'm noticing it, this could probably be reworked to use `cuda_set_sync_memops`...

darrylabbate:
Seeing the same error as before in AWS CI:

```
Command make -j failed with error:
Makefile:3742: Extraneous text after `endif' directive
Makefile:3742: Extraneous text after `endif' directive
```

prov/lpp/include/lpp_cntr.h
tstruk: done

prov/lpp/include/lpp.h
tstruk: done

prov/lpp/src/lpp_atomic.c
tstruk: done

prov/lpp/src/lpp_nosys.c
tstruk: done

tstruk:
@darrylabbate could you please check what AWS CI is not happy about?

prov/lpp/src/lpp_getinfo.c
tstruk: Yes. Added a close() before continue.

prov/lpp/include/lpp_tagged.h
tstruk: done

shijin-aws:
>@darrylabbate could you please check what AWS CI is not happy about?

It seems the latest push succeeded in compilation

tstruk:
> > @darrylabbate could you please check what AWS CI is not happy about?
> 
> It seems the latest push succeeded in compilation

It takes hours for the AWS CI to finish. I can see that it's still in progress. I don't have visibility into what stage it is at.

darrylabbate:
> 
> 
> 
> 
> 
> 
> 
> 
> > > @darrylabbate could you please check what AWS CI is not happy about?
> 
> > 
> 
> > It seems the latest push succeeded in compilation
> 
> 
> 
> It takes hours for the AWS CI to finish. I can see that it's still in progress. I don't have visibility into what stage it is at.

For security reasons, this is unfortunately a limitation of our current CI. We'll do our best to communicate CI failures that aren't caught by Actions, etc.

prov/lpp/src/lpp_hmem.c
shijin-aws: >As for NCCL >=2.19, is this related to the changes with device memory allocation? I ran into a problem testing out newer NCCL version, but was able to mitigate it with the NCCL_CUMEM_ENABLE=0 env var.

Yes, the problem can be mitigated by that env exactly. We have a boolean as `FI_OPT_CUDA_API_PERMITTED` to toggle whether CUDA API can be permitted to call inside Libfabric in data transfer. NCCL plugin set this boolean to false to disable cuda api usage in Libfabric.


https://ofiwg.github.io/libfabric/main/man/fi_endpoint.3.html

```

FI_OPT_CUDA_API_PERMITTED - bool
    This option only applies to the fi_setopt call. It is used to control endpoint’s behavior in making calls to CUDA API. By default, an endpoint is permitted to call CUDA API. If user wish to prohibit an endpoint from making such calls, user can achieve that by set this option to false. If an endpoint’s support of CUDA memory relies on making calls to CUDA API, it will return -FI_EOPNOTSUPP for the call to fi_setopt. If either CUDA library or CUDA device is not available, endpoint will return -FI_EINVAL. All providers that support FI_HMEM capability implement this option.


```

The problem is that boolean is set in the EP level via fi_setopt. while the MR reg can happen before that, so we cannot reuse that boolean so we can only do the toggle in the transmission call.

tstruk:
> > @darrylabbate could you please check what AWS CI is not happy about?
> 
> It seems the latest push succeeded in compilation

Eventually it failed again. Could you please check what was wrong?

darrylabbate:
It's still showing "in progress" on our end (and in GitHub). Where do you see it's failing?

tstruk:
> It's still showing "in progress" on our end (and in GitHub). Where do you see it's failing?

Because I pushed the latest updates, but it failed before that. So let's wait and see.

darrylabbate:
Looks like it failed to build on ARM-based EC2 instance types. I'll need to dig further to provide a root cause.

darrylabbate:
Error from previous build:

```
prov/lpp/src/lpp_memcpy.c:25:10: fatal error: immintrin.h: No such file or directory

 #include <immintrin.h>

          ^~~~~~~~~~~~~

compilation terminated.

make[1]: *** [prov/lpp/src/src_libfabric_la-lpp_memcpy.lo] Error 1

/tmp/ccOsPotZ.s: Assembler messages:

/tmp/ccOsPotZ.s:995: Error: unknown mnemonic `sfence' -- `sfence'

make[1]: *** [prov/lpp/src/src_libfabric_la-lpp_umc.lo] Error 1

make: *** [all] Error 2
```

tstruk:
> Error from previous build:

Thank you for digging into this. I guess I need to limit the build to x86 only for now.

tstruk:
> > Error from previous build:
> 
> Thank you for digging into this. I guess I need to limit the build to x86 only for now.

AWS CI still failing, even with x86_64 only build. Could please check what is wrong now?

darrylabbate:
Same error as earlier builds:

```
Command make -j failed with error:
Makefile:3742: Extraneous text after `endif' directive
Makefile:3742: Extraneous text after `endif' directive
```

Doesn't provide any more info than that, other than a bunch of deprecation warnings. We don't keep the generated files to inspect or anything.

shefty:
I did not see where the PR makes any changes to the libfabric core.  Those should be in a separate patch, but I expected to see new FI_ADDR and FI_PROTO values, at the very least, along with updates to the core handling for the new values (e.g. tostr functions).

fabtests/prov/lpp/src/error.h
```diff
@@ -0,0 +1,143 @@
...
+		debug_trace_push((ri), __LINE__, __func__, __FILE__, #func);   \
+		func;                                                          \
+		debug_trace_pop((ri));                                         \
+	} while (0)
+
+struct rank_info;
+void debug_trace_push(struct rank_info *ri, int line, const char *func,
+		      const char *file, const char *debugstr);
+void debug_trace_pop(struct rank_info *ri);
+void debug_dump_trace(struct rank_info *ri);
```
shefty, line 140: libfabric defines common debug print capabilities for all providers.  See fi_log.h.  This makes it easier on users, since all providers follow the same logging format and environment controls, and allows apps to redirect log messages as needed.

If you want to propose extending that functionality with trace push/pop functionality, I would move that into separate patches.

fabtests/prov/lpp/src/hmem_cuda.c
```diff
@@ -0,0 +1,341 @@
...
+	return 0;
+}
+
+#else
+inline int hmem_cuda_init(void) { return -ENOSYS; }
+void hmem_cuda_cleanup(void){}
+int hmem_cuda_alloc(void *uaddr, size_t len){return -ENOSYS;}
+void hmem_cuda_free(void *uaddr){}
+int hmem_cuda_memcpy_h2d(void *dst, void *src, size_t len){return -ENOSYS;}
+int hmem_cuda_memcpy_d2h(void *dst, void *src, size_t len){return -ENOSYS;}
```
shefty, line 340: fabtests has common hmem support functionality for cuda, rocr, level zero, neuron already

fabtests/prov/lpp/src/rcq_data.c
```diff
@@ -0,0 +1,138 @@
...
+
+		verify_buf_params.mr_idx = 0;
+		verify_buf_params.length = send_len;
+		verify_buf_params.expected_seed = seed_node_a;
+		TRACE(ri, util_verify_buf(ri, &verify_buf_params));
+	}
+
+	TRACE(ri, util_teardown(ri, pri));
+
+	return 0;
```
shefty, line 137: In general, what are these tests doing that the existing common fabtests aren't?  There's no specific objection to vendor tests, but why are they even needed?  If they extend testing, why can't the common test accommodate the changes?

wenduwan:
bot:aws:retest

j-xiong:
> > Should you rename your provider to reflect the NIC vendor? I think many providers like EFA utilize PCIe to realize Peer direct features.
> 
> We have been using this name for a long time, and it would require substantial effort to change it at this point. Since there is no clash with any other provider name, or no conflict with naming conventions we would like to keep it that way.

Keeping the name is fine, but I would prefer to de-emphasize what LPP actually stands for to avoid any confusion.

tstruk:
> I did not see where the PR makes any changes to the libfabric core. Those should be in a separate patch, but I expected to see new FI_ADDR and FI_PROTO values, at the very least, along with updates to the core handling for the new values (e.g. tostr functions).
> 
> Edited: It looks like the addressing is sockaddr. I didn't notice protocol details, but I didn't look closely into the provider implementation.

Yes, we use sockaddr, and that's why we didn't add new FI_ADDR and FI_PROTO values, and change the libfabric core.
We have tried that, but sockaddr works fine for us. It might change in the future, but for now we would like to keep it that way.

tstruk:
> bot:aws:retest

So what is the latest build failure in AWS?

shefty:
What is the actual protocol, though?  That should still be added.

tstruk:
> I did not see where the PR makes any changes to the libfabric core. Those should be in a separate patch, but I expected to see new FI_ADDR and FI_PROTO values, at the very least, along with updates to the core handling for the new values (e.g. tostr functions).
> 
> Edited: It looks like the addressing is sockaddr. I didn't notice protocol details, but I didn't look closely into the provider implementation.

Is it required to add the new FI_ADDR and FI_PROTO for LPP even if we use sockaddr addressing?

fabtests/prov/lpp/src/rcq_data.c
tstruk: The lpp specific fabtest (lpp_regression) is very useful for testing the LPP kernel driver (KLPP). We were not able to get that test coverage in the driver by using just the common fabtests.

fabtests/prov/lpp/src/error.h
tstruk: Yes, but this is in fabtests/prov/lpp/src for the lpp specific fabtest. Fabtests can not use FI_LOG macros.

darrylabbate:
From AWS CI:

```
prov/lpp/src/hmem_util.o: In function `hmem_memcpy_d2h':
/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:62: undefined reference to `hmem_rocm_memcpy_d2h'
prov/lpp/src/hmem_util.o: In function `hmem_memcpy_h2d':
/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:80: undefined reference to `hmem_rocm_memcpy_h2d'
prov/lpp/src/hmem_util.o: In function `hmem_alloc':
/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:98: undefined reference to `hmem_rocm_alloc'
prov/lpp/src/hmem_util.o: In function `hmem_free':
/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:114: undefined reference to `hmem_rocm_free'
collect2: error: ld returned 1 exit status
```

tstruk:
> From AWS CI:
> 
> ```
> prov/lpp/src/hmem_util.o: In function `hmem_memcpy_d2h':
> /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:62: undefined reference to `hmem_rocm_memcpy_d2h'
> prov/lpp/src/hmem_util.o: In function `hmem_memcpy_h2d':
> /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:80: undefined reference to `hmem_rocm_memcpy_h2d'
> prov/lpp/src/hmem_util.o: In function `hmem_alloc':
> /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:98: undefined reference to `hmem_rocm_alloc'
> prov/lpp/src/hmem_util.o: In function `hmem_free':
> /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10303-dso/source/libfabric/fabtests/prov/lpp/src/hmem_util.c:114: undefined reference to `hmem_rocm_free'
> collect2: error: ld returned 1 exit status
> ```

Thank you for checking. I've added a new AM_CONDITIONAL for HMEM. Let's see if it will help.

tstruk:
What does the `ci/cloudbees/stage/Summary` do?

tstruk:
@j-xiong could you please share the results of the two intel-ofi checks that failed?

j-xiong:
The errors are from fabtests over ucx due to the change in #10326. For example: 
```
    fi_getopt(FI_OPT_INJECT_MSG_SIZE)(): benchmarks/benchmark_shared.c:97, ret=-22 (Invalid argument) 
```
They should go away with #10341.

tstruk:
> They should go away with #10341.

Thank you for info.

tstruk:
@darrylabbate what's the latest status of AWS testing for this PR?

fabtests/prov/lpp/src/hmem_cuda.c
tstruk: > fabtests has common hmem support functionality for cuda, rocr, level zero, neuron already

yes, but this will require a bit of redesign of the lpp regression test.
If that's ok, can this be merge as is, and we will rework this later?

darrylabbate:
```
prov/lpp/src/test_util.c: In function ‘util_create_mr’:

prov/lpp/src/test_util.c:348:39: error: ‘PAGE_SIZE’ undeclared (first use in this function); did you mean ‘HAVE_ZE’?

   if (posix_memalign(&mr_info->uaddr, PAGE_SIZE,

                                       ^~~~~~~~~

                                       HAVE_ZE

prov/lpp/src/test_util.c:348:39: note: each undeclared identifier is reported only once for each function it appears in

make[1]: *** [prov/lpp/src/test_util.o] Error 1
```

tstruk:
@darrylabbate could you check the results now please.

shijin-aws:
@tstruk this time is ok, there are some efa test failure (we are fixing that) which is irrelevant to this PR

tstruk:
@shefty so based on the latest comments from @j-xiong and @shijin-aws this PR is ready to go.

shefty:
thanks!

j-xiong:
I should have put a do not merge label. I didn't because there is a work in progress label there.

shefty:
Sorry about that.  I checked for a do not merge label, but didn't see one, and I added the WIP one.  Hopefully that doesn't mess up the release.  If so, it can be reverted, then re-applied.

j-xiong:
@shefty No worry. I have update the v2.0.0alpha PR with the new changes included. Simpler to re-do the packaging and validation than un-doing the changes.

PR #10306, opened 2024-08-12
xpmem,shm: prefix public xpmem symbols with ofi

wenduwan:
libxpmem defines symbols such as xpmem_init(void) which conflict with ofi. Prefix public xpmem symbols in ofi to avoid interposition.


shijin-aws (approved):

aingerson (approved):
Thank you!

PR #10309, opened 2024-08-13
[v1.22.x] xpmem,shm: prefix public xpmem symbols with ofi

wenduwan:
libxpmem defines symbols such as xpmem_init(void) which conflict with ofi. Prefix public xpmem symbols in ofi to avoid interposition.

(cherry picked from commit 23973529c9e74b0562a0550d0fa3d863ee89d11a)


shijin-aws (approved):

aingerson (approved):
Intel CI is unrelated. Thanks!

PR #10314, opened 2024-08-15
include/rdma, prov/shm,efa,util: fi_peer API changes

aingerson:
Update srx flow to take in match attributes
Rename peer_rx_entry->size to msg_size and clarify usage in man pages
Update providers and implementation to match updates
Add FI_PEER as a primary capability which signals support for being used as a peer

### DIFF
--- a/man/fi_getinfo.3.md
+++ b/man/fi_getinfo.3.md
@@ -338,6 +338,11 @@ additional optimizations.
   allow an initiator to target (or name) a specific receive context as
   part of a data transfer operation.
 
+*FI_PEER*
+: Specifies that the provider must support being used as a peer provider in
+  the peer API flow. The provider must support importing owner_ops when opening
+  a CQ, counter, and shared receive queue.
+
 *FI_READ*
 : Indicates that the user requires an endpoint capable of initiating
   reads against remote memory regions.  This flag requires that FI_RMA
@@ -457,7 +462,8 @@ may optionally report non-selected secondary capabilities if doing so
 would not compromise performance or security.
 
 Primary capabilities: FI_MSG, FI_RMA, FI_TAGGED, FI_ATOMIC, FI_MULTICAST,
-FI_NAMED_RX_CTX, FI_DIRECTED_RECV, FI_HMEM, FI_COLLECTIVE, FI_XPU, FI_AV_USER_ID
+FI_NAMED_RX_CTX, FI_DIRECTED_RECV, FI_HMEM, FI_COLLECTIVE, FI_XPU,
+FI_AV_USER_ID, FI_PEER
 
 Primary modifiers: FI_READ, FI_WRITE, FI_RECV, FI_SEND,
 FI_REMOTE_READ, FI_REMOTE_WRITE
--- a/man/fi_peer.3.md
+++ b/man/fi_peer.3.md
@@ -416,7 +416,7 @@ struct fi_peer_rx_entry {
     struct fi_peer_rx_entry *prev;
     struct fi_peer_srx *srx;
     fi_addr_t addr;
-    size_t size;
+    size_t msg_size;
     uint64_t tag;
     uint64_t cq_data;
     uint64_t flags;
@@ -428,12 +428,20 @@ struct fi_peer_rx_entry {
     struct iovec *iov;
 };
 
+struct fi_peer_match_attr {
+    fi_addr_t addr;
+    size_t msg_size;
+    uint64_t tag;
+};
+
 struct fi_ops_srx_owner {
     size_t size;
-    int (*get_msg)(struct fid_peer_srx *srx, fi_addr_t addr,
-                   size_t size, struct fi_peer_rx_entry **entry);
-    int (*get_tag)(struct fid_peer_srx *srx, fi_addr_t addr,
-                   size_t size, uint64_t tag, struct fi_peer_rx_entry **entry);
+    int (*get_msg)(struct fid_peer_srx *srx,
+                   struct fi_peer_match_attr *attr,
+                   struct fi_peer_rx_entry **entry);
+    int (*get_tag)(struct fid_peer_srx *srx,
+                   struct fi_peer_match_attr *attr,
+                   uint64_t tag, struct fi_peer_rx_entry **entry);
     int (*queue_msg)(struct fi_peer_rx_entry *entry);
     int (*queue_tag)(struct fi_peer_rx_entry *entry);
     void (*foreach_unspec_addr)(struct fid_peer_srx *srx,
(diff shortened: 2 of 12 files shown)


man/man3/fi_peer.3
```diff
@@ -454,7 +454,7 @@ struct fi_peer_rx_entry {
     struct fi_peer_rx_entry *prev;
     struct fi_peer_srx *srx;
     fi_addr_t addr;
-    size_t size;
+    size_t msg_size;
```
j-xiong, line 457: Man page changes should be made to the markdown files. The actual man page is updated automatically.

aingerson: Oops! I think VS Code opened the wrong file for me... Updated. Thanks for catching!

j-xiong:
The build check error is real.  The `size` to `msg_size` conversion is missing somewhere:
```
prov/shm/src/smr_progress.c: In function ‘smr_alloc_cmd_ctx’:
prov/shm/src/smr_progress.c:919:17: error: ‘struct fi_peer_rx_entry’ has no member named ‘size’
  919 |         rx_entry->size = cmd->msg.hdr.size;
      |                 ^~
make[1]: *** [Makefile:29128: prov/shm/src/src_libfabric_la-smr_progress.lo] Error 1
make[1]: Leaving directory '/home/runner/work/libfabric/libfabric'
make: *** [Makefile:8872: all] Error 2
Error: Process completed with exit code 2.
```

aingerson:
Updated

shijin-aws:
AWS CI failure: failed to compile with ompi

```

*** Fault tolerance
checking if want fault tolerance... Enabled mpi (Specified auto)
checking if want checkpoint/restart enabled debugging option... Disabled

============================================================================
== Modular Component Architecture (MCA) setup
============================================================================
checking for subdir args...  '--prefix=/home/ec2-user/PortaFiducia/build/libraries/openmpi/v5.0.x/install' '--with-sge' '--without-verbs' '--with-libfabric=/home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10314/install/libfabric' '--disable-man-pages' '--enable-ipv6' 'LDFLAGS=-Wl,--as-needed' '--enable-prte-prefix-by-default' '--enable-mca-dso=all' '--with-cuda=/usr/local/cuda' '--with-cuda-libdir=/usr/local/cuda/lib64/stubs' '--with-xpmem=/home/ec2-user/PortaFiducia/build/libraries/xpmem/install' '--with-xpmem-libdir=/home/ec2-user/PortaFiducia/build/libraries/xpmem/install/lib'
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking which components should be disabled... 
checking which components should be direct-linked into the library... 
checking which components should be run-time loadable... all 
checking which components should be static... default
checking for projects containing MCA frameworks... opal, ompi, oshmem

*** Configuring MCA for opal
checking for frameworks for opal... common, accelerator, allocator, backtrace, btl, dl, hwloc, if, installdirs, memchecker, memcpy, memory, mpool, patcher, pmix, rcache, reachable, shmem, smsc, threads, timer

+++ Configuring MCA framework common
checking for no configure components in framework common... 
checking for m4 configure components in framework common... ofi, sm, ucx

--- MCA component common:ofi (m4 configuration macro)
checking for MCA component common:ofi compile mode... dso
checking for ofi pkg-config name... /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10314/install/libfabric/lib/pkgconfig/libfabric.pc
checking if ofi pkg-config module exists... no
checking for ofi header at /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10314/install/libfabric/include... not found

Traceback (most recent call last):
  File "/home/ec2-user/PortaFiducia/libraries/openmpi/install.py", line 295, in <module>
    main()
  File "/home/ec2-user/PortaFiducia/libraries/openmpi/install.py", line 291, in main
```

wenduwan:
bot:aws:retest

man/fi_peer.3.md
```diff
@@ -494,9 +502,17 @@ These calls are invoked by the peer provider to obtain the receive buffer(s)
...
+The msg_size field indicates the received message size.
+This field may be needed by the owner when handling FI_MULTI_RECV or FI_PEEK.
+The owner will set the peer_entry->msg_size field on get_msg/tag() for the owner
+and peer to use later, if needed. This field will be set on both the expected
+and unexpected paths.
+The returned rx_entry->iov returned from the owner refers to the full size of
+the posted receive passed to the peer. The peer provider is responsible for
+checking that an incoming message fits within the provided buffer space and
+generating truncation errors.
+The tag parameter is used for tagged messages.
```
shefty, line 515: I would consider documenting that tag must be set to 0 for non-tagged messages, so it's not left uninitialized.

man/fi_getinfo.3.md
```diff
@@ -338,6 +338,11 @@ additional optimizations.
   allow an initiator to target (or name) a specific receive context as
   part of a data transfer operation.
 
+*FI_PEER*
```
shefty, line 341: Are there changes which use this or needed updates to fabtests or the common flow?

man/fi_peer.3.md
aingerson: Good call. Will update.

man/fi_getinfo.3.md
aingerson: What do you mean?

shefty: This updates the man page only.  There's no code updated anywhere to use the new definition.

aingerson: Ah gotcha. Yes this is all targeted for the link provider as a way to query providers that it is able to link together. But other providers can use it as well if using a peer provider directly (like efa should TECHNICALLY add FI_PEER to its shm fi_getinfo call). I can add that.

shefty: It's okay if the link provider is going to set this on input.  But what provider is going to respond to the query?  How does setting this not just fail the call?

aingerson: I don't follow your question - the link provider needs to make sure that all providers it's going to link together support the peer API so it queries the provider with caps->FI_PEER. shm will succeed. cxi will succeed. rxm will succeed. But for example psm3 probably won't. So the fi_getinfo call in the link layer will not include psm3 as a possible link. It will only include rxm+shm, cxi+shm, etc. If a user requests link with psm3+shm, it would return ENODATA instead of attempting to run it and getting an ENOSYS error. Are you saying you prefer link to fail when trying to open everything instead of not including the bad combination in the list of possible links? Or are you saying I missed adding this in shm (as a supported capability) and efa (as a requested shm capability)?

shefty: E.g, in the shm provider:
```
#define SMR_TX_CAPS (OFI_TX_MSG_CAPS | FI_TAGGED | OFI_TX_RMA_CAPS | FI_ATOMICS)
#define SMR_RX_CAPS (FI_SOURCE | FI_RMA_EVENT | OFI_RX_MSG_CAPS | FI_TAGGED | \
		     OFI_RX_RMA_CAPS | FI_ATOMICS | FI_DIRECTED_RECV | \
		     FI_MULTI_RECV)

struct fi_info smr_info = {
	.caps = SMR_TX_CAPS | SMR_RX_CAPS | FI_MULTI_RECV | FI_LOCAL_COMM,
```

Where is FI_PEER being used as a cap anywhere?  I don't see any providers setting it.  It's not part of any common checks as a primary cap (e.g. included in ofi.h).  Fabtests getinfo_test doesn't include this.

aingerson: Ok so option number 2, ie you're grumpy I didn't add it. Got it. Will add.

shijin-aws (approved):
efa provider change LGTM, thanks!

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -850,7 +854,6 @@ struct efa_rdm_ope *efa_rdm_msg_alloc_rxe_for_tagrtm(struct efa_rdm_ep *ep,
 		}
 		(*pkt_entry_ptr)->ope = rxe;
 
-		peer_rxe->size = efa_rdm_pke_get_rtm_msg_length(*pkt_entry_ptr);
```
shijin-aws, line 853: I forget why I added this earlier... is it because tag matching also need size to support `fi_peek`? If so, is it safe to remove this?

shijin-aws: https://github.com/ofiwg/libfabric/commit/d865260e0b2c0d6db7e487a2a4747fb0cdbd6d40

aingerson:
Updated with following changes:
- Added wording to man pages to indicate tag needs to be set to 0 for non-tagged case
- Specified FI_PEER as a domain capability since all the resources belong in the domain
- Added FI_PEER to ofi capabilities for filtering
- Added FI_PEER and FI_AV_USER_ID to fi_tostr
- Added FI_PEER to shm capabilities
- Added FI_PEER to shm getinfo query in efa

j-xiong (approved):

PR #10316, opened 2024-08-16
fabtests: Add option for FI_MORE and fabtests with FI_MORE

jiaxiyan:
- Add the option `--use-fi-more` to allow tests with FI_MORE flag set
- Run rdm_tagged_bw with `fi_sendmsg/fi_tsendmsg`, run rma write/writedata with `fi_writemsg` to allow testing with customized flags
- Add efa pytests with FI_MORE for `fi_rma_bw` and `fi_rdm_tagged_bw`

### DIFF
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -274,9 +274,21 @@ static int rma_bw_rx_comp()
 	return ft_tx(ep, remote_fi_addr, FT_RMA_SYNC_MSG_BYTES, &tx_ctx);
 }
 
+static int set_fi_more_flag(int i, int j, int flags)
+{
+	if (j < opts.window_size - 1 && i >= opts.warmup_iterations &&
+	    i < opts.iterations + opts.warmup_iterations - 1) {
+		flags |= FI_MORE;
+	} else {
+		flags &= ~FI_MORE;
+	}
+	return flags;
+}
+
 int bandwidth(void)
 {
 	int ret, i, j, inject_size;
+	int flags = 0;
 
 	inject_size = inject_size_set ?
 			hints->tx_attr->inject_size : fi->tx_attr->inject_size;
@@ -310,6 +322,12 @@ int bandwidth(void)
 				ret = ft_post_inject_buf(ep, remote_fi_addr,
 						opts.transfer_size, NO_CQ_DATA,
 						tx_ctx_arr[j].buf, tx_seq);
+			} else if (opts.use_fi_more) {
+				flags = set_fi_more_flag(i, j, flags);
+				ret = ft_sendmsg(ep, remote_fi_addr,
+						tx_ctx_arr[j].buf,
+						opts.transfer_size,
+						&tx_ctx_arr[j].context, flags);
 			} else {
 				ret = ft_post_tx_buf(ep, remote_fi_addr,
 						opts.transfer_size, NO_CQ_DATA,
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -413,6 +413,7 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 {
 	int ret, i, j, inject_size;
 	size_t offset;
+	int flags = 0;
 
 	inject_size = inject_size_set ?
 			hints->tx_attr->inject_size: fi->tx_attr->inject_size;
@@ -455,6 +456,12 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 			if (opts.transfer_size <= inject_size) {
 				ret = ft_post_rma_inject(FT_RMA_WRITE, tx_buf + offset,
 						opts.transfer_size, remote);
+			} else if (opts.use_fi_more) {
+				flags = set_fi_more_flag(i, j, flags);
+				ret = ft_post_rma_writemsg(
+						tx_buf + offset,
+						opts.transfer_size, remote,
+						&tx_ctx_arr[j].context, flags);
 			} else {
 				ret = ft_post_rma(FT_RMA_WRITE, tx_buf + offset,
 						opts.transfer_size, remote,
@@ -478,6 +485,13 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 							tx_buf + offset,
 							opts.transfer_size,
 							remote);
+				} else if (opts.use_fi_more) {
+					flags = set_fi_more_flag(i, j, flags);
+					flags |= FI_REMOTE_CQ_DATA;
+					ret = ft_post_rma_writemsg(
+							tx_buf + offset,
+							opts.transfer_size, remote,
+							&tx_ctx_arr[j].context, flags);
 				} else {
 					ret = ft_post_rma(FT_RMA_WRITEDATA,
 							tx_buf + offset,
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -4183,6 +4183,8 @@ void ft_longopts_usage()
 		"manual, auto, or unified");
 	FT_PRINT_OPTS_USAGE("--max-msg-size <size>",
 		"maximum untagged message size");
+	FT_PRINT_OPTS_USAGE("--use-fi-more",
+		"Run tests with FI_MORE");
 }
 
 int debug_assert;
@@ -4195,6 +4197,7 @@ struct option long_opts[] = {
 	{"data-progress", required_argument, NULL, LONG_OPT_DATA_PROGRESS},
 	{"control-progress", required_argument, NULL, LONG_OPT_CONTROL_PROGRESS},
 	{"max-msg-size", required_argument, NULL, LONG_OPT_MAX_MSG_SIZE},
+	{"use-fi-more", no_argument, NULL, LONG_OPT_USE_FI_MORE},
 	{NULL, 0, NULL, 0},
 };
 
@@ -4234,6 +4237,9 @@ int ft_parse_long_opts(int op, char *optarg)
 		return 0;
 	case LONG_OPT_MAX_MSG_SIZE:
 		opts.max_msg_size = atoi(optarg);
+		return 0;
+	case LONG_OPT_USE_FI_MORE:
+		opts.use_fi_more = 1;
 		return 0;
 	default:
 		return EXIT_FAILURE;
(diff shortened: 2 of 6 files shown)


fabtests/benchmarks/benchmark_shared.c
```diff
@@ -306,16 +307,23 @@ int bandwidth(void)
...
+
+			if (opts.fi_more) {
+				if (j < opts.window_size - 1 &&
+				    i >= opts.warmup_iterations &&
+				    i < opts.iterations + opts.warmup_iterations - 1) {
+					flags |= FI_MORE;
+				} else {
+					flags &= ~FI_MORE;
+				}
+			}
```
shefty, line 319: This only applies to the else clause below.  This is starting to nest fairly deep, so a function for this check may be worthwhile.  Also, the check is repeated in the next patch, and the check itself is a little complicated in order to be correct.

fabtests/include/shared.h
```diff
@@ -607,6 +608,8 @@ int ft_recvmsg(struct fid_ep *ep, fi_addr_t fi_addr,
 		size_t size, void *ctx, int flags);
 int ft_sendmsg(struct fid_ep *ep, fi_addr_t fi_addr,
 	       void *buf, size_t size, void *ctx, int flags);
+int ft_writemsg(struct fid_ep *ep, fi_addr_t fi_addr,
+		void *buf, size_t size, void *ctx, struct fi_rma_iov *remote, int flags);
```
shefty, line 612: check line length

fabtests/common/shared.c
```diff
@@ -4183,6 +4183,8 @@ void ft_longopts_usage()
 		"manual, auto, or unified");
 	FT_PRINT_OPTS_USAGE("--max-msg-size <size>",
 		"maximum untagged message size");
+	FT_PRINT_OPTS_USAGE("--fi-more",
```
shijin-aws, line 4212: I prefer `--use-fi-more`

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -311,11 +325,11 @@ int bandwidth(void)
 						opts.transfer_size, NO_CQ_DATA,
 						tx_ctx_arr[j].buf, tx_seq);
 			} else {
-				ret = ft_post_tx_buf(ep, remote_fi_addr,
```
shijin-aws, line 314: I feel like we should keep the fi more test only an opt-in feature, that we should still keep the non-fi-more test using ft_post_tx_buf

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -452,9 +453,11 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 				ret = ft_post_rma_inject(FT_RMA_WRITE, tx_buf + offset,
 						opts.transfer_size, remote);
 			} else {
-				ret = ft_post_rma(FT_RMA_WRITE, tx_buf + offset,
-						opts.transfer_size, remote,
-						&tx_ctx_arr[j].context);
+				flags = set_fi_more_flag(i, j, flags);
+				ret = ft_post_rma(
```
shijin-aws, line 457: Similar comment as earlier, can we only use FT_RMA_WRITEMSG only when testing fi_more.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -310,12 +322,17 @@ int bandwidth(void)
...
+						 tx_ctx_arr[j].buf,
+						 opts.transfer_size,
+						 &tx_ctx_arr[j].context, flags);
 			} else {
-				ret = ft_post_tx_buf(ep, remote_fi_addr,
-						opts.transfer_size, NO_CQ_DATA,
-						&tx_ctx_arr[j].context,
-						tx_ctx_arr[j].buf,
-						mr_desc, tx_seq);
+				ret = ft_post_tx_buf(
```
shijin-aws, line 332: Nit: it seems you don't have to touch this code block? Leaving them unchanged will make it easier to review.

fabtests/pytest/efa/test_rma_bw.py
```diff
@@ -62,3 +62,13 @@ def test_rma_bw_1G(cmdline_args, operation_type, completion_semantic):
     efa_run_client_server_test(cmdline_args, command, 2,
                                completion_semantic=completion_semantic, message_size=1073741824,
                                memory_type="host_to_host", warmup_iteration_type=0, timeout=timeout)
+
+@pytest.mark.functional
+@pytest.mark.parametrize("operation_type", ["write"])
```
shijin-aws, line 67: can we do both `write` and `writedata`

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -460,17 +486,25 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
...
+							opts.transfer_size, 
+							remote, &tx_ctx_arr[j].context,
+							flags);
 				} else {
 					ret = ft_post_rma(FT_RMA_WRITEDATA,
 							tx_buf + offset,
 							opts.transfer_size,
-							remote,	&tx_ctx_arr[j].context);
+							remote,	&tx_ctx_arr[j].context,
+							flags);
```
shijin-aws, line 501: Hmmm.. do you expect a non-zero flags to be used here? Similar to other `ft_post_rma` that is used for non-fi-more.

jiaxiyan: I should add `flags |= FI_REMOTE_CQ_DATA` for writedata with fi_more.
Do you mean we should just pass 0 for the others?

shijin-aws: My asking is that before your change, `flag` is not an argument of `ft_post_rma`, if we only extend `ft_post_rma` to use flags for `FI_MORE` usage, should we keep flags as 0 for non-fi-more case?

jiaxiyan: Yeah that makes sense

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -460,17 +486,25 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 							tx_buf + offset,
 							opts.transfer_size,
 							remote);
+				} else if (opts.use_fi_more) {
```
shijin-aws, line 489: I think you need to `|= FI_REMOTE_CQ_DATA` to the flags here, right?

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -460,17 +486,25 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
...
+					flags = set_fi_more_flag(i, j, flags);
+					ret = ft_post_rma(FT_RMA_WRITEMSG, 
+							tx_buf + offset,
+							opts.transfer_size, 
+							remote, &tx_ctx_arr[j].context,
+							flags);
```
shijin-aws, line 495: I think you need to `| FI_REMOTE_CQ_DATA` to the flags here, right?

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -460,17 +486,25 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 							tx_buf + offset,
 							opts.transfer_size,
 							remote);
```
shijin-aws, line 488: Inject and fi_more are not excluded. I think the logic is a bit misleading here. `opts.use_fi_more` shouldn't be an `else if` to `inject`. Instead, it's better to make `if (opts.use_fi_more)` inside the `else` block.

fabtests/common/shared.c
```diff
@@ -2327,6 +2331,26 @@ ssize_t ft_post_rma(enum ft_rma_opcodes op, char *buf, size_t size,
 			remote_fi_addr, remote->addr + ft_remote_read_offset(buf),
 			remote->key, context);
 		break;
+	case FT_RMA_WRITEMSG:
```
shijin-aws, line 2334: I am concerned that, if we want to make `FT_RMA_WRITEMSG` as a separate rma op code, you would expect user to run `fi_rma_bw -o writemsg`, but your change doesn't seem to work for this case. A simpler change could be keep the current op codes, but implement an internal `ft_post_rma_writemsg` that is used for fi_more usage.

fabtests/benchmarks/benchmark_shared.c
j-xiong: I don't see there is a difference here between `if ... else if ... else ...` and `if ... else { if ... else ... }`. The current form is simpler.

fabtests/common/shared.c
jiaxiyan: I can add writemsg to ft_parse_api_opts if you prefer that.

shijin-aws: The problem is whether we want to test the api explicitly.... similar to inject, we don't make it as an op code, but we use it implicitly

j-xiong: That's a valid concern. I think the proposed `ft_post_rma_writemsg` is more preferrable than adding the new op code.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -460,6 +485,13 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
...
+				} else if (opts.use_fi_more) {
+					flags = set_fi_more_flag(i, j, flags);
+					flags |= FI_REMOTE_CQ_DATA;
```
shijin-aws, line 490: For safety,  better to just do `flags | FI_REMOTE_CQ_DATA` in `ft_post_rma_writemsg` instead of `|=` the `flags` globally.

shijin-aws: Hmmm nvm, I think this may not be better than what your have now.

shijin-aws (approved):

j-xiong (approved):

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

shijin-aws:
Is Intel CI failure related? @j-xiong

j-xiong:
@shijin-aws We have CI outage at this moment. Will update later.

j-xiong:
The re-run passed. However, the old, failed results can't be removed from the status.

shijin-aws:
@j-xiong Thanks, I am merging this then

PR #10338, opened 2024-08-27
prov/psm2,psm3: fix incorrect unlock function

aingerson:
Use optimized lock function to unlock on CQ write error instead of direct unlock which causes an assert failure if FI_THREAD_DOMAIN is specified


j-xiong:
Thanks. The psm2 provider probably has the same error, could you fix it as well?

aingerson:
@j-xiong Updated with psm2 fix as well. Thanks!

j-xiong (approved):

PR #10345, opened 2024-08-28
prov/efa: Zero the cq entry array in dgram ep progress

shijin-aws:
In efa_dgram_ep_progress_internal, cq_entry was declared without initializing as 0. It casts this array's pointer into smaller structs like `fi_cq_entry` during the cq->read_entry call, which doesn't have the flags member, and then cast it back to fi_cq_tagged_entry which has flags. This finally makes the temp_cq_entry struct has a flags from uninitialized values and can cause illegal flags to be used in the cq write. This patch fixes issue by initializing the cq entry array as 0.


darrylabbate (approved):

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

PR #10347, opened 2024-08-28
prov/efa:  Fix fi_sendmsg/fi_writemsg with FI_INJECT

jiaxiyan:
Currently EFA provider doesn't handle the *msg call with FI_INJECT flags correctly. It allows user to pass in a buffer with valid MR while doing inject - outbound buffer should be reusable immediately after function returns.
That being said, without the inline HW support, EFA provider should do a copy from the user buffer to the internal bounce buffer even if application provides a valid MR for the buffer.
The inject rdma write path assumes the outbound buffer is always FI_HMEM_SYSTEM, but it doesn't have to.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_pke_utils.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_utils.c
@@ -93,20 +93,8 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
 		return 0;
 	}
 
-	if (iov_mr && (iov_mr->peer.flags & OFI_HMEM_DATA_DEV_REG_HANDLE)) {
-		assert(iov_mr->peer.hmem_data);
-		copied = ofi_dev_reg_copy_from_hmem_iov(pke->wiredata + payload_offset,
-							data_size, iov_mr->peer.iface,
-							(uint64_t)iov_mr->peer.hmem_data,
-							ope->iov, ope->iov_count,
-							segment_offset);
-	} else {
-		copied = ofi_copy_from_hmem_iov(pke->wiredata + payload_offset,
-						data_size,
-		                                iov_mr ? iov_mr->peer.iface : FI_HMEM_SYSTEM,
-		                                iov_mr ? iov_mr->peer.device.reserved : 0,
-		                                ope->iov, ope->iov_count, segment_offset);
-	}
+	copied = efa_rdm_pke_copy_from_hmem_iov(
+		iov_mr, pke, ope, payload_offset, segment_offset, data_size);
 
 	assert(copied == data_size);
 	pke->payload = pke->wiredata + payload_offset;
--- a/prov/efa/src/rdm/efa_rdm_pke_utils.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_utils.c
@@ -22,7 +22,7 @@
  *
  * This function may or may not copy data from user's buffer to the packet's.
  * If copy, "payload" will be pointing to a location inside packet entry's wiredata.
- * Otherwise, "payload" will be pointing to a location insder user's buffer.
+ * Otherwise, "payload" will be pointing to a location inside user's buffer.
  *
  * @param[in,out]	pkt_entry	packet entry. Header must have been set when the function is called
  * @param[in]		ope		operation entry that has user buffer information.
@@ -41,7 +41,8 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
 					  size_t data_size)
 {
 	int tx_iov_index, ret;
-	bool efa_can_access_memory;
+	bool p2p_available;
+	bool use_inline_buf;
 	size_t tx_iov_offset, copied;
 	struct efa_mr *iov_mr;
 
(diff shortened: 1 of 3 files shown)


prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -82,13 +82,17 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
...
+	 * 2. data to be send is in 1 iov, because device only support 2 iov,
+	 * and we use 1st iov for header.
+	 * 3. FI_INJECT transfer flag is not set. FI_INJECT requires outbound
+	 * buffer to be reusable immediately after function returns, so without
+	 * the inline HW support, a copy is needed.
 	 */
 	if (efa_can_access_memory &&
-	    (tx_iov_offset + data_size <= ope->iov[tx_iov_index].iov_len)) {
+	    (tx_iov_offset + data_size <= ope->iov[tx_iov_index].iov_len) &&
+	    !(ope->fi_flags & FI_INJECT)) {
```
shijin-aws, line 95: I don't think this is correct.

FI_INJECT can still avoid copy when the buffer is on host and size is within inline. It seems we need some more refactor here.

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -100,6 +100,37 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
...
+ *
+ * @param[in]		iov_mr  memory descriptors of the user's data buffer.
+ * @param[in,out]	pke	    packet entry. Header must have been set when the function is called.
+ * @param[in]		ope		operation entry that has user buffer information.
+ * @param[in]		payload_offset	the data offset in reference to pkt_entry->wiredata.
+ * @param[in]		segment_offset	the data offset in reference to user's buffer
+ * @param[in]		data_size	length of the data to be set up.
+ * @return   		length of data copied.
+ */
+size_t efa_rdm_copy_from_hmem_iov(struct efa_mr *iov_mr,
```
shijin-aws, line 125: I prefer make it as a inline function in `efa_rdm_pke_utils.h`

shijin-aws: also name it as `efa_rdm_pke_copy_from_hmem_iov`

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -61,34 +62,36 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
...
 	/*
 	 * Copy can be avoid if the following 2 conditions are true:
-	 * 1. EFA can directly access the memory
-	 * 2. data to be send is in 1 iov, because device only support 2 iov, and we use
-	 *    1st iov for header.
+	 * 1. data to be send is in 1 iov, because device only support 2 iov,
+	 * and we use 1st iov for header.
+	 * 2. EFA can directly access the memory
+	 * Note that FI_INJECT requires outbound buffer to be reusable
+	 * immediately after function returns, so without the inline HW support,
```
shijin-aws, line 90: Can you rephrase the `without the inline HW support,` to `unless using inline`

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -61,34 +62,36 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
 	assert(tx_iov_index < ope->iov_count);
 	assert(tx_iov_offset < ope->iov[tx_iov_index].iov_len);
 	iov_mr = ope->desc[tx_iov_index];
-
+	p2p_available = false;
+	inline_send = false;
 	/* When using EFA device, EFA device can access memory that
```
shijin-aws, line 67: Since we are removing `efa_can_access_memory`, I think we can remove these comments, just explain how copy can be removed based on the p2p_available and inline_send status later.

shijin-aws (approved):
LGTM, some nits on the comments

prov/efa/src/rdm/efa_rdm_pke_utils.c
shijin-aws: Also can we change `EFA can directly access the memory` to `p2p is available or using inline` here

prov/efa/src/rdm/efa_rdm_pke_utils.h
```diff
@@ -58,6 +58,42 @@ size_t efa_rdm_pke_get_segment_offset(struct efa_rdm_pke *pke)
...
+ *
+ * @param[in]		iov_mr  memory descriptors of the user's data buffer.
+ * @param[in,out]	pke	    packet entry. Header must have been set when the function is called.
+ * @param[in]		ope		operation entry that has user buffer information.
+ * @param[in]		payload_offset	the data offset in reference to pkt_entry->wiredata.
+ * @param[in]		segment_offset	the data offset in reference to user's buffer
+ * @param[in]		data_size	length of the data to be set up.
+ * @return   		length of data copied.
+ */
+inline size_t
```
shijin-aws, line 72: I missed it earlier: it should be `static inline size_t`

jiaxiyan: Can I call it from efa_rdm_ope.c if it is static?

shijin-aws:
bot:aws:retest

PR #10366, opened 2024-09-09
fabtests: remove 2.0 deprecated features

aingerson:
Various calls and settings were deprecated as part of libfabric 2.0 alpha which cause lots of compile warnings.
There are so many warnings that it is becoming difficult to spot valid compile time warnings.
This removes all references to deprecated features in fabtests to clean up the warnings.
Specifically, the following calls and flags:
- fi_wait_open
- fi_wait
- fi_poll
- FI_MR_BASIC
- FI_MR_SCALABLE
- FI_LOCAL_MR


fabtests/common/check_hmem.c
```diff
@@ -44,7 +44,7 @@ int main(int argc, char** argv)
 		return EXIT_FAILURE;
 	hints->mode = ~0;
 	hints->domain_attr->mode = ~0;
-	hints->domain_attr->mr_mode = ~(FI_MR_BASIC | FI_MR_SCALABLE);
+	hints->domain_attr->mr_mode = FT_MR_ALL | FI_MR_HMEM;
```
shefty, line 47: It's interesting that "ALL" doesn't actually mean "ALL"

aingerson: It's interesting that leaving somehow made you *more* annoying

fabtests/include/shared.h
```diff
@@ -69,6 +69,12 @@ extern "C" {
 	((type *) ((char *)ptr - offsetof(type, field)))
 #endif
 
+/*
+ * Internal version of deprecated APIs.
+ * These are used internally to avoid compiler warnings.
+ */
+#define OFI_MR_BASIC		(1 << 0)
+#define OFI_MR_SCALABLE		(1 << 1)
```
shefty, line 77: As an alternative, consider:
```
#define OFI_MR_DEPRECATED_BITS (0x3)
```

aingerson: I was matching the strategy in libfabric to be clearer when reading but yeah we could do that for fabtests. I'm fine either way! You tell me.

shefty: I think OFI_MR_DEPRECATED would be more annoying for you, so I'm leaning towards it.  My actual reason is for someone reading the code 2 years from now, ```~DEPRECATED``` may convey more meaning than ```~(BASIC | SCALABLE)```

j-xiong: I prefer Sean's suggestion. The definitions are only used in the code to mask off unwanted bits so there is no need to keep them separate. 

This is different from the libfabric code, where the internal definition is still in place so that application backward compatibility can still be maintained.

aingerson: As you wish. Updated. I also removed FI_MR_LOCAL from the exclusion in the MR test because it didn't look necessary. I think maybe whoever wrote it was trying to exclude FI_LOCAL_MR? Unclear. You'll probably yell at me about it anyway so it'll be short lived :)

j-xiong (approved):

PR #10369, opened 2024-09-10
prov/efa: Improve the zero-copy recv error message.

shijin-aws:
Extend the error message when the receiver has zcpy recv turned on but get a rtm pkt that it cannot handle. The extended error message includes possible root causes and the potential mitigations.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -371,12 +371,13 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct 
 	 * QP and we cannot cancel that.
 	 */
 	if (OFI_UNLIKELY(ep->use_zcpy_rx && efa_rdm_pkt_type_is_rtm(pkt_type))) {
-		EFA_WARN(FI_LOG_CQ,
-			"Invalid pkt type %d! Peer %d doesn't respect the request from this EP that"
-			" RTM packets must be sent to the user recv QP.\n",
-			base_hdr->type, (int)pkt_entry->addr);
+		void *errbuf;
+		size_t errbuf_len;
 
-		efa_base_ep_write_eq_error(&ep->base_ep, FI_EINVAL, FI_EFA_ERR_INVALID_PKT_TYPE);
+		/* local & peer host-id & ep address will be logged by efa_rdm_write_error_msg */
+		if (!efa_rdm_write_error_msg(ep, pkt_entry->addr, FI_EFA_ERR_INVALID_PKT_TYPE_ZCPY_RX, &errbuf, &errbuf_len))
+			EFA_WARN(FI_LOG_CQ, "Error: %s\n", (const char *) errbuf);
+		efa_base_ep_write_eq_error(&ep->base_ep, FI_EINVAL, FI_EFA_ERR_INVALID_PKT_TYPE_ZCPY_RX);
 		efa_rdm_pke_release_rx(pkt_entry);
 		return;
 	}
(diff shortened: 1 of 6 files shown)


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -373,7 +373,10 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 	if (OFI_UNLIKELY(ep->use_zcpy_rx && efa_rdm_pkt_type_is_rtm(pkt_type))) {
 		EFA_WARN(FI_LOG_CQ,
 			"Invalid pkt type %d! Peer %d doesn't respect the request from this EP that"
-			" RTM packets must be sent to the user recv QP.\n",
+			" RTM packets must be sent to the user recv QP. One possible reason is that"
+			" the sender is using an old libfabric that doesn't support the user recv QP feature."
+			" Please consider upgrading the sender's libfabric version to >= 1.22.0, or disable"
+			" the zero-copy recv feature by setting FI_EFA_USE_ZCPY_RX=0 in the environment",
```
wenduwan, line 379: I think we can be more concise with the error message, e.g.

```
Packet type <type> from remote peer <instance-id>/<gid> is not supported when zero-copy recv mode is ON.
 Please ensure local and remote libfabric versions match, or turn off zero-copy recv via
 FI_EFA_USE_ZCPY_RX=0.
```

darrylabbate: We should relegate these long, passage-like messages to `efa_show_help()`

shijin-aws: If we use `efa_show_help()`, I am not sure what error code should I populate here? It seems a very specific corner case instead of a general libfabric error

shijin-aws: >Please ensure local and remote libfabric versions match

Libfabric was supposed to be backward compatibility (with some corner cases for sure, this is one of them). I do not think we should enforce users to have matched libfabric versions on local and remote

shijin-aws:
@wenduwan See latest update

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -371,10 +371,15 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
...
 	if (OFI_UNLIKELY(ep->use_zcpy_rx && efa_rdm_pkt_type_is_rtm(pkt_type))) {
+		char peer_host_id_str[EFA_HOST_ID_STRING_LENGTH + 1] = {0};
+		if (!peer->host_id || EFA_HOST_ID_STRING_LENGTH != snprintf(peer_host_id_str, EFA_HOST_ID_STRING_LENGTH + 1, "i-%017lx", peer->host_id))
+			strcpy(peer_host_id_str, "N/A");
+
 		EFA_WARN(FI_LOG_CQ,
-			"Invalid pkt type %d! Peer %d doesn't respect the request from this EP that"
-			" RTM packets must be sent to the user recv QP.\n",
-			base_hdr->type, (int)pkt_entry->addr);
+			"Packet type %d from remote peer %d (host id: %s) is not supported when zero-copy recv mode is ON."
```
wenduwan, line 379: GID would be better here - if host id is not available for some reason we can still trace down the peer with GID.

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: > I am not sure what error code should I populate here? It seems a very specific corner case instead of a general libfabric error

I'd suggest a new error code; similar to what we did for unresponsive receiver with/without a successful handshake.

If it can't really be detected in a similar way, it'd make a good backlog item.

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: I reused `efa_rdm_write_err_msg` in the latest revision, it should have host id and gid for local & remote

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: Done, see latest revision.

wenduwan (approved):

darrylabbate (approved):

PR #10371, opened 2024-09-11
prov/rxm : Corrected rxm to return an existing MR key in its MR Map.

nikhilnanal:
if an key/item already exists in its MR map rxm returns an error instead of reflecting the behaviour of the underlying core provider viz returning a key. This patch corrects the behaviour by returning a success to make sure it is consistent with the core provider's behaviour.

### DIFF
--- a/prov/rxm/src/rxm_domain.c
+++ b/prov/rxm/src/rxm_domain.c
@@ -372,9 +372,10 @@ static int rxm_mr_add_map_entry(struct util_domain *domain,
 	ofi_genlock_lock(&domain->lock);
 	ret = ofi_mr_map_insert(&domain->mr_map, msg_attr, &temp_key, rxm_mr, flags);
 	if (OFI_UNLIKELY(ret)) {
-		FI_WARN(&rxm_prov, FI_LOG_DOMAIN,
-			"MR map insert for atomic verification failed %d\n",
-			ret);
+		if (ret != -FI_ENOKEY)
+			FI_WARN(&rxm_prov, FI_LOG_DOMAIN,
+				"MR map insert for atomic verification failed %d\n",
+				ret);
 	} else {
 		assert(rxm_mr->mr_fid.key == temp_key);
 	}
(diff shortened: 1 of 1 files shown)


prov/rxm/src/rxm_domain.c
```diff
@@ -372,9 +372,10 @@ static int rxm_mr_add_map_entry(struct util_domain *domain,
 	ofi_genlock_lock(&domain->lock);
 	ret = ofi_mr_map_insert(&domain->mr_map, msg_attr, &temp_key, rxm_mr, flags);
 	if (OFI_UNLIKELY(ret)) {
-		FI_WARN(&rxm_prov, FI_LOG_DOMAIN,
-			"MR map insert for atomic verification failed %d\n",
-			ret);
+		if (ret != -FI_ENOKEY)
+			FI_WARN(&rxm_prov, FI_LOG_DOMAIN,
+				"MR map insert for atomic verification failed %d\n",
+				ret);
```
j-xiong, line 378: In the case of FI_ENOKEY, the key in the mr_map is still associated with an old rxm_mr, which apparently is still valid. We are now in a situation that a key is associated with multiple rxm_mr's, but only one is present in the mr_map.  

In `rxm_handle_atomic_req`,  `rxm_mr_get_map_entry` is called to retrieve the rxm_mr associated with the key. In this case, the old rxm_mr is going to be referenced. That may or may not work as expected. If one close the old rxm_mr. The key will be removed from the mr_map. The atomic req processing would fail even though it should.

I don't have a good solution for this. It's basically a behavior confliction of cached MR at verbs level and non-cached behavior of the mr_map at rxm level.

nikhilnanal: would it make sense to do a lookup before inserting the rxm_mr and only insert if lookup fails?

j-xiong: That doesn't make any difference. The mr_map is mainly used to find rxm_mr with a key. There can be only one association in mr_map, but we have multiple rxm_mr's.

PR #10375, opened 2024-09-11
fabtests/efa: Avoid testing duplicate mixed memory type workload

shijin-aws:
Currently, we always test mixed memory type workload in two directions explicitly: like cuda-to-host, host-to-cuda.

However. pingpong and rma read/write test are all bi-directional, which means there is no difference for cuda-to-host and host-to-cuda. Running them twice in this situation is not necessary and causing longer test duration.

This patch improves this by having a reduced memory type list for bi-directional tests.

### DIFF
--- a/fabtests/pytest/efa/conftest.py
+++ b/fabtests/pytest/efa/conftest.py
@@ -1,15 +1,38 @@
 import pytest
 
+# The memory types for bi-directional tests.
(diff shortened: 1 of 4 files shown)


wenduwan (approved):

shijin-aws:
bot:aws:retest

PR #10376, opened 2024-09-11
util/av: Log AV insert with AV's specified address format

darrylabbate:
This prevents `(null)` from being logged when addr isn't compatible with `struct sockaddr`

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1053,19 +1053,19 @@ size_t ofi_mask_addr(struct sockaddr *maskaddr, const struct sockaddr *srcaddr,
 	return len;
 }
 
-void ofi_straddr_log_internal(const char *func, int line,
+void ofi_straddr_log_internal(const char *func, int line, uint32_t addr_format,
 			      const struct fi_provider *prov,
 			      enum fi_log_level level,
 			      enum fi_log_subsys subsys, char *log_str,
 			      const void *addr)
 {
 	char buf[OFI_ADDRSTRLEN];
-	uint32_t addr_format;
 	size_t len = sizeof(buf);
 
 	if (fi_log_enabled(prov, level, subsys)) {
 		if (addr) {
-			addr_format = ofi_translate_addr_format(ofi_sa_family(addr));
+			if (addr_format == FI_FORMAT_UNSPEC)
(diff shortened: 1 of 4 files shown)


src/common.c
```diff
@@ -1053,25 +1053,21 @@ size_t ofi_mask_addr(struct sockaddr *maskaddr, const struct sockaddr *srcaddr,
...
-		} else {
-			fi_log(prov, level, subsys, func, line, "%s: (null)\n", log_str);
-		}
-	}
+	if (addr && addr_format == FI_FORMAT_UNSPEC)
+		addr_format = ofi_translate_addr_format(ofi_sa_family(addr));
+
+	if (fi_log_enabled(prov, level, subsys))
+		fi_log(prov, level, subsys, func, line, "%s: %s\n", log_str,
+		       addr ? ofi_straddr(buf, &len, addr_format, addr) : "(null)");
```
j-xiong, line 1070: The translation of `addr_format` is only needed when log is enabled and when addr is not empty. I would keep the original code block and just add `if (addr_format == FI_FORMAT_UNSPEC)` before the addr_format translation.

j-xiong (approved):

shijin-aws:
Can we enforce ofi_straddr_log to have addr_format as an argument explicitly?

shijin-aws:
`addr_format` is associated with `fi_info` and inherited by `domain` and so on, I don't think making an av specific log function is the best idea.

darrylabbate:
> `addr_format` is associated with `fi_info` and inherited by `domain` and so on, I don't think making an av specific log function is the best idea.

I was trying to avoid redundant params to `ofi_straddr_log()` for providers that don't utilize a proprietary address format like EFA (which seems to be most/all direct callers of it). No strong feelings here. Ideally I'd use something like `__VA_OPT__` to make it variadic and allow the format to be omitted.

shijin-aws:
It does require considerable change to enforce `addr_format` to be an argument of `ofi_straddr_log`. No strong opinion here.

shijin-aws (approved):

darrylabbate:
bot:aws:retest

PR #10377, opened 2024-09-11
man/fi_peer, prov/shm,efa: update peer context usage and providers using old usage

aingerson:
The man pages had no wording describing who owned the shared peer resource when using the peer API. Because of this, the util cq and counter implementation were using owner-allocated fid_peer_* resources and the shm srx implementation was using a peer-allocated resource. This patch set updates the man pages to add consistency to the resource allocation flow, specifying that the owner is responsible for allocating the resource and making sure peers use different fids. It also updates the shm implementation to adhere to this flow and updates efa's use of shm as a peer in this way as well.


aingerson:
@shijin-aws I updated efa here, but let me know if I missed something. I can also squash the second and third patches into on because technically the shm change will break efa until that third patch is added. I kept it separate for easier review. I can squash when you confirm it looks good and passes

shijin-aws:
AWS CI failure is related, seems all EFA single node test failed. No detailed logs yet, will try to reproduce

shijin-aws:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1304,10 +1309,22 @@ static int efa_rdm_ep_ctrl(struct fid *fid, int command, void *arg)
 		 * shared memory region.
 		 */
 		if (ep->shm_ep) {
-			peer_srx_context.srx = util_get_peer_srx(ep->peer_srx_ep);
```
shijin-aws, line 1307: Does `shm_peer_srx` does the same thing as `ep->peer_srx_ep`? We created `ep->peer_srx_ep` via `util_ep_srx_context` which allocates the owner resources for srx. I need to read your man page change more carefully, but it doesn't make sense to me that we have 2 owner srx resources here

aingerson: There are 2 resources here - the peer srx (allocated by efa and imported to shm) and then shm has a fid_ep that it uses to track any shm resources (it doesn't need any but it can). I think that second one I actually implemented incorrectly because shm doesn't return a fid anymore. It's not necessary for the shm case but it's part of the definition so that's definitely a miss and maybe why it's failing (because the fi_close at the end will actually segfault)

aingerson: @shijin-aws I fixed the issues I knew would cause a failure but it looks like it's still grumpy. I'm guessing I missed something in the message flow - can you share the failure?

shijin-aws:
@aingerson I get a segfault when running `fi_rdm_pingpong -p efa -E` on single node, the backtrace is 

```
Core was generated by `fi_rdm_pingpong -p efa -E'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fb4ea3b03ca in smr_av_insert (av_fid=0x555b85edd9d0, addr=0x7ffd3979cd10, count=1, fi_addr=0x7fb482e39258, flags=2199023255552, context=0x0)
    at prov/shm/src/smr_av.c:175
175                             smr_ep->srx->owner_ops->foreach_unspec_addr(smr_ep->srx,
(gdb) 
(gdb) bt
#0  0x00007fb4ea3b03ca in smr_av_insert (av_fid=0x555b85edd9d0, addr=0x7ffd3979cd10, count=1, fi_addr=0x7fb482e39258, flags=2199023255552, context=0x0)
    at prov/shm/src/smr_av.c:175
#1  0x00007fb4ea315ff6 in fi_av_insert (av=0x555b85edd9d0, addr=0x7ffd3979cd10, count=1, fi_addr=0x7fb482e39258, flags=2199023255552, context=0x0)
    at ./include/rdma/fi_domain.h:509
#2  0x00007fb4ea31970f in efa_conn_rdm_init (av=0x555b85eda2a0, conn=0x7fb482e39220, insert_shm_av=true) at prov/efa/src/efa_av.c:319
#3  0x00007fb4ea31bc3e in efa_conn_alloc (av=0x555b85eda2a0, raw_addr=0x7ffd3979d020, flags=0, context=0x0, insert_shm_av=true) at prov/efa/src/efa_av.c:487
#4  0x00007fb4ea31d4f4 in efa_av_insert_one (av=0x555b85eda2a0, addr=0x7ffd3979d020, fi_addr=0x7ffd3979cfa8, flags=0, context=0x0, insert_shm_av=true)
    at prov/efa/src/efa_av.c:625
#5  0x00007fb4ea31d6ed in efa_av_insert (av_fid=0x555b85eda2e8, addr=0x7ffd3979d020, count=1, fi_addr=0x555b85c8f3d0 <remote_fi_addr>, flags=0, context=0x0)
    at prov/efa/src/efa_av.c:667
#6  0x0000555b85c7bef8 in fi_av_insert (context=0x0, flags=0, fi_addr=0x555b85c8f3d0 <remote_fi_addr>, count=1, addr=0x7ffd3979d020, av=0x555b85eda2e8)
    at /home/ubuntu/PortaFiducia/build/libraries/libfabric/main/install/libfabric/include/rdma/fi_domain.h:509
#7  ft_av_insert (av=av@entry=0x555b85eda2e8, addr=addr@entry=0x7ffd3979d020, count=count@entry=1, fi_addr=fi_addr@entry=0x555b85c8f3d0 <remote_fi_addr>, 
    flags=flags@entry=0, context=context@entry=0x0) at common/shared.c:1516
#8  0x0000555b85c84bef in ft_exchange_addresses_oob (av_ptr=0x555b85eda2e8, ep_ptr=<optimized out>, remote_addr=0x555b85c8f3d0 <remote_fi_addr>)
    at common/shared.c:1555
#9  0x0000555b85c84da5 in ft_init_av_dst_addr (av_ptr=0x555b85eda2e8, ep_ptr=0x555b85eda520, remote_addr=remote_addr@entry=0x555b85c8f3d0 <remote_fi_addr>)
    at common/shared.c:1574
#10 0x0000555b85c84f74 in ft_init_av () at common/shared.c:1531
#11 0x0000555b85c85859 in ft_init_fabric () at common/shared.c:1336
#12 0x0000555b85c789b7 in run () at benchmarks/rdm_pingpong.c:109
#13 main (argc=4, argv=0x7ffd3979da08) at benchmarks/rdm_pingpong.c:109
```

shijin-aws:
@aingerson the latest push fixed the earlier segfaults, but we have some unit test failure, I trying to get to the bottom

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -987,6 +987,13 @@ static int efa_rdm_ep_close(struct fid *fid)
 	}
 
 	if (efa_rdm_ep->shm_ep) {
+		ret = fi_close(&efa_rdm_ep->shm_srx->fid);
```
shijin-aws, line 990: We have other places that can close shm resources, including https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L1048-L1059

shijin-aws: And our unit tests have tons of places that replies on closing shm ep hackily as `fi_close`: https://github.com/ofiwg/libfabric/blob/main/prov/efa/test/efa_unit_test_ep.c#L126-L131. These tests are all failed by this change. I think we need to modify our unit tests that doesn't close shm ep explicitly in the test.

aingerson: @shijin-aws Oh that is funky. I didn't realize that. I added a ref count in the shm srx import to make sure it gets closed properly even though there isn't really anything to be closed for shm so I think dropping this close will cause failures in the regular case. I think you probably at minimum need a wrapper around that hack to close both the shm ep and the shm srx. Do you think that would be enough?

shijin-aws: @shijin-aws I am preparing a patch to fix that. We can use `fi_setopt(.. FI_OPT_SHARED_MEMORY_PERMMITED, false)` to disable shm in the test.

aingerson: Sweet, thank you!

shijin-aws: But generally it is annoying that provider needs to manage more resources for peer providers

aingerson: I agree, but I think we need to match all the peer resources to the same flow - the peer cq and peer counter are the same (efa needs to close the shm cq and counters). I think it needs to happen and this one was just missed so it's annoying to add it in later. That would be a perk of the link provider, to remove that handling from efa

shijin-aws: I opened a PR to clean up our tests: https://github.com/ofiwg/libfabric/pull/10406. we need to rebase your PR after https://github.com/ofiwg/libfabric/pull/10406 is merged and then make sure shm resources are cleaned correctly in https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L1048-L1059 as well.

prov/efa/src/rdm/efa_rdm_ep.h
```diff
@@ -54,6 +54,10 @@ struct efa_rdm_ep {
 
 	/* shm provider fid */
 	struct fid_ep *shm_ep;
+	/* shm srx fid (shm-owned) */
+	struct fid_ep *shm_srx;
+	/* shm peer_srx (efa-owned) */
+	struct fid_peer_srx *shm_peer_srx;
```
shijin-aws, line 60: I am still not convinced that we need to have 3 peer srx related resources in efa ep, adding these 2, there are

- peer_srx_ep
- shm_srx
- shm_peer_srx

If we want to make an analogy to peer cq and cntr

efa_cq has only 1 peer cq related field: shm_cq,
it opens the shm_cq via 

```

shm_cq_attr.flags |= FI_PEER;
peer_cq_context.size = sizeof(peer_cq_context);
peer_cq_context.cq = cq->util_cq.peer_cq;
fi_cq_open(efa_domain->shm_domain, &shm_cq_attr,
				 &cq->shm_cq, &peer_cq_context);
```



I believe `shm_cq` plays a role similar to `shm_srx` here, `util_cq.peer_cq` plays the role of `shm_peer_srx` here.

If we make an analogy between `util_cq` and `util_srx`, shouldn't we make `shm_peer_srx` to be part of `util_srx`, as `util_srx.peer_srx` ?

aingerson: @shijin-aws I think it depends on the use case of efa (I don't really know how efa uses the srx on the efa side). The issue (and difference between the CQ use case) is the peer_ops. shm needs to be able to set the peer_srx->peer_ops to its own ops. Every peer that has a separate set of peer_ops needs a separate peer_srx for dedicated use for that peer. If in efa you access all the util_srx functions directly and not through the owner_ops (as if you were a peer) then you're right, you don't need a separate peer_srx - you could put it in the util_srx and have that one be dedicated for shm. However, if in efa you are calling get_msg(), etc, you have to pass in a unique peer_srx that has the efa peer_ops for the util_srx to call later. I think you likely need two (one for efa and one for shm) because I think you just go directly through the util_generic_recvmsg call which directly initiates the start_msg() flow.
For the CQ case, there are no peer_ops, so the peers can technically share the same peer_cq because the only purpose is to call the owner_ops which won't change from peer to peer.
So for the three resources you're talking about, I think each one is needed for a slightly different purpose:
- peer_srx_ep -> peer_srx for the efa-owned srx, for use with efa (owner_ops = util_srx_ops, peer_ops = efa_peer_ops)
- shm_srx -> shm fid for srx (shm doesn't need it but in case it did need anything extra, this is to track it)
- shm_peer_srx -> peer_srx for the efa-owned srx, for use with shm (owner_ops = util_srx_ops, peer_ops = shm_peer_ops)

shijin-aws:
@aingerson still a lot of unit test failure in AWS CI, will get back you later today

shijin-aws:
I hit a segmentation fault around this line inside `efa_rdm_ep_close`

```
if (efa_rdm_ep->shm_ep) {
		ret = fi_close(&efa_rdm_ep->shm_srx->fid);
```

shijin-aws:
I figure it out. EFA provider create shm_ep during fi_endpoint, but only create srx resources during fi_enable. Our unit tests have tests that only create/close ep without enable. Your `fi_close` shm_srx and shm_peer_srx should explicitly check `efa_rdm_ep->shm_srx` and `efa_rdm_ep->shm_peer_srx` before doing it.

shijin-aws:
I reworked your efa change: https://github.com/shijin-aws/libfabric/commit/57749f69f593f596c14054b22893f2e1f754d865 please cherry-pick

aingerson:
@shijin-aws Thanks for fixing, Shi! Looks like everything is passing. I think this is finally good to go know. Feel free to merge if that looks ok to you

shijin-aws (approved):

PR #10382, opened 2024-09-14
prov/efa: Some cleanups

shijin-aws:
This PR contains 2 commits that cleans up efa_rdm_ep and efa_rdm_peer structs.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_protocol.h
+++ b/prov/efa/src/rdm/efa_rdm_protocol.h
@@ -41,7 +41,12 @@ struct efa_ep_addr {
 #define EFA_RDM_EXTRA_FEATURE_READ_NACK		BIT_ULL(6)
 #define EFA_RDM_EXTRA_FEATURE_REQUEST_USER_RECV_QP	BIT_ULL(7)
 #define EFA_RDM_NUM_EXTRA_FEATURE_OR_REQUEST		8
-#define EFA_RDM_MAX_NUM_EXINFO				(256)
+/*
+ * The length of 64-bit extra_info array used in efa_rdm_ep
+ * and efa_rdm_peer
+ * 4 means 64*4=256 bits of extra features or requests
+ */
+#define EFA_RDM_MAX_NUM_EXINFO				(4)
 
 /*
  * Packet type ID of each packet type (section 1.3)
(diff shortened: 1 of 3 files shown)


prov/efa/src/rdm/efa_rdm_protocol.h
```diff
@@ -41,7 +41,12 @@ struct efa_ep_addr {
 #define EFA_RDM_EXTRA_FEATURE_READ_NACK		BIT_ULL(6)
 #define EFA_RDM_EXTRA_FEATURE_REQUEST_USER_RECV_QP	BIT_ULL(7)
 #define EFA_RDM_NUM_EXTRA_FEATURE_OR_REQUEST		8
-#define EFA_RDM_MAX_NUM_EXINFO				(256)
+/*
+ * The length of 64-bit extra_info array used in efa_rdm_ep
+ * and efa_rdm_peer
+ * 4 means 64*4=256 bits of extra features or requests
+ */
+#define EFA_RDM_MAX_NUM_EXINFO				(4)
```
wenduwan, line 49: Hmmm did this bug cause any issue?

wenduwan (approved):

prov/efa/src/rdm/efa_rdm_protocol.h
shijin-aws: The code was actually correct: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke_nonreq.c#L40 that it divided by 64 when calculating the number of extra_info. But I have no idea why we make it 256 as the array length. Likely an overlook

darrylabbate: The arrays were originally of length 1 ($4-4+1$): https://github.com/ofiwg/libfabric/commit/aa5fec1790d50b6f4efb8036ee03610f1c0176b0

shijin-aws: It was planned that RDM v4 will support only 64 extra features, v5 will support 128, and so on. So it was reasonable to have it as length 1 at that time. I think this PR aimed to make it support 256/64 = 4 so it can support 256 features.

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

PR #10385, opened 2024-09-16
core: 1.8 ABI compat

jiaxiyan:
ABI version is updated to 1.8 to accommodate fi_fabric2() API introduced in https://github.com/ofiwg/libfabric/pull/10279

### DIFF
--- a/include/ofi_abi.h
+++ b/include/ofi_abi.h
@@ -111,7 +111,7 @@ extern "C" {
  * name appended with the ABI version that it is compatible with.
  */
 
-#define CURRENT_ABI "FABRIC_1.7"
+#define CURRENT_ABI "FABRIC_1.8"
 
 #if  HAVE_ALIAS_ATTRIBUTE == 1
 #define DEFAULT_SYMVER_PRE(a) a##_
--- a/man/fabric.7.md
+++ b/man/fabric.7.md
@@ -447,6 +447,14 @@ attributes:
 *fi_domain_attr*
 : Added max_ep_auth_key
 
+## ABI 1.8
+
+ABI version starting with libfabric 2.0. Added new fi_fabric2 API call.
+Added new fields to the following attributes:
+
+*fi_domain_attr*
+: Added max_group_id
+
 # SEE ALSO
 
 [`fi_info`(1)](fi_info.1.html),


include/ofi_abi.h
```diff
@@ -111,7 +111,7 @@ extern "C" {
  * name appended with the ABI version that it is compatible with.
  */
 
-#define CURRENT_ABI "FABRIC_1.7"
+#define CURRENT_ABI "FABRIC_1.8"
```
shefty, line 114: @j-xiong 

This change indicates that the fi_fabric2() call is not aligned with the documentation in this file.  Checking the code, it's using DEFAULT_SYMVER(... FABRIC_1.8) directly instead of CURRENT_SYMVER().  There were also changes to fi_getinfo / fi_freeinfo / fi_dupinfo as part of ABI 1.8 changes.  The changes to those functions should be captured in the man page below as well.

If CURRENT_SYMVER() will no longer be used, that macro should be removed and the man page updated to reflect direct use of DEFAULT_SYMVER instead.

shijin-aws: @shefty @j-xiong If we are going to implement changes to `fi_info` today. Is it better make it part of FABRIC_1.8 (since 2.0 is not released yet?), or we need to bump it as FABRIC_1.9?

man/fabric.7.md
```diff
@@ -447,6 +447,10 @@ attributes:
 *fi_domain_attr*
 : Added max_ep_auth_key
 
+## ABI 1.8
+
+ABI version starting with libfabric 2.0. Added new fi_fabric2() API call.
+
```
j-xiong, line 453: Please add the following:
```
Added new fields to the following attributes:

*fi_domain_attr*
: Added  max_group_id
```

include/ofi_abi.h
j-xiong: We will keep it 1.8 for the 2.0 release.

j-xiong: @shefty Using DEFAULT_SYMVER() here for the new function instead of CURRENT_SYMVER() has the advantage of showing the exact ABI version at the definition and not needing to change it again when the ABI version is bumped. I think we can remove CURRENT_SYMVER() definition and usage info from the header. If no objection, I will make that change in a separate PR.

shefty: I think it's fine to remove CURRENT_SYMVER

shijin-aws (approved):

j-xiong (approved):

shefty (approved):

j-xiong:
Intel CI error is unrelated.

PR #10386, opened 2024-09-16
core: Remove CURRENT_SYMVER() macro

j-xiong:
CURRENT_SYMVER() is just a shortcut for DEFAULT_SYMVER() with version set to CURRENT_ABI. Using DEFAULT_SYMVER() directly makes the version info more visible from the code and eliminate the need to change from CURRENT_SYMVER() to DEFAULT_SYMVER() when the ABI version is bumped.

Remove the definition of CURRENT_SYMVER() from the header. Update the documentation accordingly.

### DIFF
--- a/include/ofi_abi.h
+++ b/include/ofi_abi.h
@@ -46,31 +46,16 @@ extern "C" {
 /*
  * ABI version support definitions.
  *
- * CURRENT_ABI:
- * This defines the current ABI version.  The ABI version is separate from
- * the packaging or interface versions.  Whenever a change is
- * added to the interfaces that breaks the ABI, this definition should be
- * updated.  If you don't know if a change breaks the ABI, then you shouldn't
- * be modifying the header files under include/rdma!  :P
- *
  * DEFAULT_SYMVER_PRE:
  * This macro appends an underscore to a function name.  It should be used
  * around any function that is exported from the library as the default call
  * that applications invoke.
  *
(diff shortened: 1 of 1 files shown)


j-xiong:
@shefty @shijin-aws With this update, the definition of `CURRENT_ABI` can also be removed. Any reason we should keep it other than having a reference in the source code? We can always find the latest ABI version in `libfabric.map.in`.

shefty:
I'd remove CURRENT_ABI as well.  Fewer items to update makes it less likely to miss something.

include/ofi_abi.h
```diff
@@ -97,7 +80,7 @@ extern "C" {
  * {
  *    ...
  * }
- * CURRENT_SYMVER(foo_, foo);
+ * DEFAULT_SYMVER(foo_, foo, MYLIB_1.1);
```
shefty, line 83: While we're updating this area, let's change "MYLIB" to "FABRIC"

shefty (approved):

j-xiong:
Rebased.

j-xiong:
Intel CI failures:
All shm tests that open endpoint and with dsa disabled:
```
fi_enable(): common/shared.c:1451, ret=-13 (Permission denied)
```
Unlikely related to this PR, but worth some investigation.

j-xiong:
A left-over shm file was not cleaned up. Retesting.

PR #10387, opened 2024-09-17
core: Add endpoint option for traffic class

shijin-aws:
Today, traffic class (tclass) is a global configuration per fi_info, which doesn't allow application to
configure it for certain endpoints. This patch removes such restriction by introducing a new endpoint option for traffic class, with allowed values inherited from the current tclass values.

A typical application will call fi_setopt() to set the tclass for a given endpoint, and call fi_getopt() to retrieve the tclass value.

Providers that don't support this option will return -FI_ENOPROTOOPT. In that case, `tx_attr->tclass` should be used.

### DIFF
--- a/include/rdma/fi_endpoint.h
+++ b/include/rdma/fi_endpoint.h
@@ -78,6 +78,7 @@ enum {
 	FI_OPT_INJECT_TAGGED_SIZE,	/* size_t */
 	FI_OPT_INJECT_RMA_SIZE,		/* size_t */
 	FI_OPT_INJECT_ATOMIC_SIZE,	/* size_t */
+	FI_OPT_TRAFFIC_CLASS,		/* int */
 };
 
 /*
--- a/man/fi_endpoint.3.md
+++ b/man/fi_endpoint.3.md
@@ -606,6 +606,12 @@ The following option levels and option names and parameters are defined.
   Providers that don't support this option will return -FI_ENOPROTOOPT. In that
   case, `tx_attr->inject_size` should be used.
 
+- *FI_OPT_TRAFFIC_CLASS - int *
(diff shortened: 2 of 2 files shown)


shefty:
fi_info isn't a global setting.  The user can change the attributes per endpoint.

shijin-aws:
>fi_info isn't a global setting. The user can change the attributes per endpoint.

Does that mean you need to modify `fi_info` in the fi_endpoint call?

```
int fi_endpoint(struct fid_domain *domain, struct fi_info *info,
    struct fid_ep **ep, void *context);

```

shijin-aws:
I have an impression that application is not allowed to modify `fi_info` from `fi_getinfo`. It is not the case?

shefty:
The app can modify it, and it may need to in order to change the attributes from the defaults.  This would be expected by apps using scalable endpoints, for example.  It just isn't supposed to allocate memory or try to manually construct one.

Having a setopt() to set the ep tclass isn't necessarily a problem, but the intent is to have the tclass specified at ep creation time, in case the provider must configure HW then.

shijin-aws:
>The app can modify it, and it may need to in order to change the attributes from the defaults. This would be expected by apps using scalable endpoints, for example. It just isn't supposed to allocate memory or try to manually construct one.

I see, thank you!

> Having a setopt() to set the ep tclass isn't necessarily a problem, but the intent is to have the tclass specified at ep creation time, in case the provider must configure HW then.

I only need to configure tclass before the ep enable. no matter ep open or setopt. Yeah, provider must configure HW to achieve that. EFA provider currently doesn't create QP until EP enable, so it is not an issue.

shijin-aws:
I will close this PR then

PR #10392, opened 2024-09-17
prov/efa: always use p2p for system memory

jiaxiyan:
For system memory, it should use p2p even with unregistered buffer. Otherwise it will affect performance.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep.h
+++ b/prov/efa/src/rdm/efa_rdm_ep.h
@@ -304,13 +304,11 @@ bool efa_rdm_ep_should_write_rnr_completion(struct efa_rdm_ep *ep)
 static inline
 int efa_rdm_ep_use_p2p(struct efa_rdm_ep *efa_rdm_ep, struct efa_mr *efa_mr)
 {
-	if (!efa_mr)
-		return 0;
-
 	/*
-	 * always send from host buffers if we have a descriptor
+	 * P2P is always available for host memory (Unregistered buffer will be
+	 * regarded as host memory as EFA provider requires FI_MR_HMEM)
 	 */
-	if (efa_mr->peer.iface == FI_HMEM_SYSTEM)
+	if (!efa_mr || efa_mr->peer.iface == FI_HMEM_SYSTEM)
 		return 1;
 
 	if (efa_rdm_ep_domain(efa_rdm_ep)->hmem_info[efa_mr->peer.iface].p2p_supported_by_device)
--- a/prov/efa/src/rdm/efa_rdm_pke_utils.c
+++ b/prov/efa/src/rdm/efa_rdm_pke_utils.c
@@ -41,7 +41,7 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
 					  size_t data_size)
 {
 	int tx_iov_index, ret;
-	bool p2p_available;
+	bool mr_p2p_available;
 	bool use_inline_buf;
 	size_t tx_iov_offset, copied;
 	struct efa_mr *iov_mr;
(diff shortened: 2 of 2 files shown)


shijin-aws:
I am concerned that https://github.com/ofiwg/libfabric/commit/deaa841d887a2a5231b20228c01a69450706c19c may also be impacted by this issue. You checked 

```
efa_rdm_ep_use_p2p(ep, rxe->desc[0]);
```

to determine if we can do p2p for the receive buffer. If receive buffer is unregistered and on host, we should allow rdma read as well.

I think instead of adding such iface check everywhere, we should investigate whether we can just make `efa_rdm_ep_use_p2p` return true for unregistered buffer. You need to audit whether that change can cause any behavior change. My most concern is here https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke_utils.c#L69. But it looks it is called when we have a valid `iov_mr`, so it shouldn't cause a problem if we make `efa_rdm_ep_use_p2p` return true for NULL mr.

prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -250,15 +250,15 @@ int efa_rdm_pke_get_available_copy_methods(struct efa_rdm_ep *ep,
 					   bool *restrict gdrcopy_available)
 {
 	int ret;
-	bool p2p_available;
+	bool mr_p2p_available;
```
shijin-aws, line 254: Better to add a `assert(efa_mr)` here

prov/efa/src/rdm/efa_rdm_ep.h
```diff
@@ -304,11 +304,8 @@ bool efa_rdm_ep_should_write_rnr_completion(struct efa_rdm_ep *ep)
 static inline
 int efa_rdm_ep_use_p2p(struct efa_rdm_ep *efa_rdm_ep, struct efa_mr *efa_mr)
 {
-	if (!efa_mr)
-		return 0;
-
 	/*
-	 * always send from host buffers if we have a descriptor
+	 * always send from host buffers no matter efa_mr is registered or not
```
shijin-aws, line 308: This comment is not correct. We are not always sending from host buffers if efa_mr is NULL. For eager send, we will do a copy from host buffer to registered bounce buffer. 

We can just change it to

```
P2p is always available for host memory (Unregistered buffer will be regarded as host memory as EFA provider requires FI_MR_HMEM)
```

shijin-aws (approved):

shijin-aws:
Need to backport

darrylabbate:
I'm a little dubious of this change and can see it easily becoming a future bug. If the [null `txe->desc[0]` being passed here](https://github.com/ofiwg/libfabric/blob/8f45c7a2513be7a86a948ae44d951f2dcd262f38/prov/efa/src/rdm/efa_rdm_msg.c#L132) was netting an undesired result, wouldn't it limit the blast radius to simply pass true for `use_p2p` in `efa_rdm_msg_select_rtm()` when `txe->desc[0]` is null?

I see that we've walked through every caller of `efa_rdm_ep_use_p2p()`, but changing the logic for the null parameter just strikes me as something that will surprise (and not delight) us later.

shijin-aws:
> I'm a little dubious of this change and can see it easily becoming a future bug. If the [null `txe->desc[0]` being passed here](https://github.com/ofiwg/libfabric/blob/8f45c7a2513be7a86a948ae44d951f2dcd262f38/prov/efa/src/rdm/efa_rdm_msg.c#L132) was netting an undesired result, wouldn't it limit the blast radius to simply pass true for `use_p2p` in `efa_rdm_msg_select_rtm()` when `txe->desc[0]` is null?
> 
> I see that we've walked through every caller of `efa_rdm_ep_use_p2p()`, but changing the logic for the null parameter just strikes me as something that will surprise (and not delight) us later.

I don't think it can become a bug, `efa_rdm_ep_use_p2p` should be interpreted as whether p2p can be used for a given MR. It is a valid deduction that NULL desc should be host memory, and we should always support p2p for host memory, I don't think it is wrong to make this function return true for NULL desc.

PR #10401, opened 2024-09-19
fabtests/prov/lpp: change invalid condition in fi_tsenddata

tstruk:
Change invalid condition in parameter check in fi_tsenddata test.

### DIFF
--- a/fabtests/prov/lpp/src/rcq_data.c
+++ b/fabtests/prov/lpp/src/rcq_data.c
@@ -1,6 +1,7 @@
 #include "test_util.h"
 
 static const uint64_t context = 0xabcd;
+#define BUF_NUM 4
 
 int run_fi_tsenddata(struct rank_info *ri){
 	struct wait_tx_cq_params wait_tx_cq_params = { 0 };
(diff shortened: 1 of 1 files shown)


fabtests/prov/lpp/src/rcq_data.c
```diff
@@ -16,7 +16,7 @@ int run_fi_tsenddata(struct rank_info *ri){
 	size_t ndata = sizeof(rcq_data)/sizeof(*rcq_data);
 	size_t nbufflens = sizeof(buff_lens)/sizeof(*buff_lens);
 
-	if (ndata == nbufflens)
+	if (ndata != nbufflens)
 		return -EINVAL;
```
j-xiong, line 20: Why having this check at all? The two values are always equal based on how the two arrays are initialized.

tstruk: That was an assert probably left there as the code evolved.
I have reworked the code to use buffers with static size and removed to check.

tstruk: PR updated.

j-xiong:
@tstruk Please rebase to resolve conflicts.

j-xiong (approved):

PR #10417, opened 2024-09-26
prov/efa: Use tclass to prioritize the messages from an ep

jiaxiyan:
To prioritize the messages from a given endpoint, user can specify ` fi_info->tx_attr->tclass = FI_TC_LOW_LATENCY` in the fi_endpoint() call to set the service level in rdma-core. All other tclass values will be ignored.

### DIFF
--- a/prov/efa/src/efa_base_ep.c
+++ b/prov/efa/src/efa_base_ep.c
@@ -167,7 +167,7 @@ static int efa_base_ep_modify_qp_rst2rts(struct efa_base_ep *base_ep,
  * @param init_attr_ex ibv_qp_init_attr_ex
  * @return int 0 on success, negative integer on failure
  */
-int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex)
+int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex, uint32_t tclass)
 {
 	struct efadv_qp_init_attr efa_attr = { 0 };
 
@@ -185,6 +185,10 @@ int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex)
 			efa_attr.flags |= EFADV_QP_FLAGS_UNSOLICITED_WRITE_RECV;
 #endif
 		efa_attr.driver_qp_type = EFADV_QP_DRIVER_TYPE_SRD;
+#if HAVE_EFADV_SL
+		if (tclass == FI_TC_LOW_LATENCY)
+			efa_attr.sl = EFA_QP_DEFAULT_SERVICE_LEVEL;
+#endif
 		(*qp)->ibv_qp = efadv_create_qp_ex(
 			init_attr_ex->pd->context, init_attr_ex, &efa_attr,
 			sizeof(struct efadv_qp_init_attr));
@@ -206,7 +210,7 @@ int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
 {
 	int ret;
 
-	ret = efa_qp_create(&base_ep->qp, init_attr_ex);
+	ret = efa_qp_create(&base_ep->qp, init_attr_ex, base_ep->info->tx_attr->tclass);
 	if (ret)
 		return ret;
 
--- a/prov/efa/src/efa_base_ep.h
+++ b/prov/efa/src/efa_base_ep.h
@@ -12,6 +12,8 @@
 #include "ofi_util.h"
 #include "rdm/efa_rdm_protocol.h"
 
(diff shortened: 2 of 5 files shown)


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1646,7 +1646,7 @@ int efa_rdm_ep_check_qp_in_order_aligned_128_bytes(struct efa_rdm_ep *ep,
 	/* Create a dummy qp for query only */
 	efa_rdm_ep_construct_ibv_qp_init_attr_ex(ep, &attr_ex, ibv_cq_ex, ibv_cq_ex);
 
-	ret = efa_qp_create(&qp, &attr_ex);
+	ret = efa_qp_create(&qp, &attr_ex, ep->base_ep.info->tx_attr->tclass);
```
shijin-aws, line 1649: this is only a trial QP that query the in order aligned 128 bytes capability, no need to specify tclass here. Can use FI_TC_UNSPEC.

prov/efa/src/efa_base_ep.c
```diff
@@ -185,6 +185,10 @@ int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex)
 			efa_attr.flags |= EFADV_QP_FLAGS_UNSOLICITED_WRITE_RECV;
 #endif
 		efa_attr.driver_qp_type = EFADV_QP_DRIVER_TYPE_SRD;
+#if HAVE_EFADV_SL
+		if (tclass == FI_TC_LOW_LATENCY)
+			efa_attr.sl = tclass;
```
shijin-aws, line 190: This is not correct, `tclass` is not same with sl. @mrgolin suggested using 8 as a start point.

jiaxiyan: Do they define a name for it? I want to avoid using magic number.

shijin-aws: It's a good callout.... I don't think there are defined macro for it. can we have that? @mrgolin

shijin-aws: If we are not going to have a macro from rdma-core. We can start from defining a macro inside Libfabric and reference it here.

shijin-aws:
pending on @mrgolin 's feedback

shijin-aws (approved):

prov/efa/src/efa_base_ep.c
```diff
@@ -185,6 +185,10 @@ int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex)
 			efa_attr.flags |= EFADV_QP_FLAGS_UNSOLICITED_WRITE_RECV;
 #endif
 		efa_attr.driver_qp_type = EFADV_QP_DRIVER_TYPE_SRD;
+#if HAVE_EFADV_SL
+		if (tclass == FI_TC_LOW_LATENCY)
+			efa_attr.sl = SL_8;
```
mrgolin, line 190: Notice that if service level isn't supported by the device, ```-EINVAL``` will be returned by ```efadv_create_qp_ex``` for any non zero value.

shijin-aws: Do you have any comments on that service level value macro? Are we having something similar in rdma-core?

prov/efa/src/efa_base_ep.h
```diff
@@ -12,6 +12,13 @@
 #include "ofi_util.h"
 #include "rdm/efa_rdm_protocol.h"
 
+#if HAVE_EFADV_SL
+enum efa_qp_service_level {
+    SL_0 = 0,
+    SL_8 = 8,
+};
```
darrylabbate, line 19: This is pretty superfluous unless `SL_n` can ever be anything other than `n`. We should just define a macro that represents whatever default value we want to use. E.g. `EFA_QP_DEFAULT_SERVICE_LEVEL`

shijin-aws: Agreed

shijin-aws (approved):

prov/efa/src/efa_base_ep.h
```diff
@@ -12,6 +12,10 @@
 #include "ofi_util.h"
 #include "rdm/efa_rdm_protocol.h"
 
+#if HAVE_EFADV_SL
+#define EFA_QP_DEFAULT_SERVICE_LEVEL 8
+#endif
```
darrylabbate, line 17: Nit: no need to make this conditional
```suggestion
#define EFA_QP_DEFAULT_SERVICE_LEVEL 8
```

darrylabbate (approved):

shijin-aws:
bot:aws:retest

PR #10420, opened 2024-09-27
fabtests/configure, efa: call provider specific configure

shijin-aws:
Create efa provider's own configure.m4 and make configure.ac call it.
Build efa_rdma_checker conditionally when all symbols are available.

### DIFF
--- a/fabtests/configure.ac
+++ b/fabtests/configure.ac
@@ -80,6 +80,9 @@ AS_IF([test -z "$CFLAGS"],
 # <3), but it is necessary in AM 1.12.x.
 m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 
+dnl Call the provider's CONFIGURE and CONDITIONALS macros
+m4_include([prov/efa/configure.m4])
+
 AM_PROG_LIBTOOL
 
 AC_ARG_WITH([valgrind],
--- a/fabtests/prov/efa/Makefile.include
+++ b/fabtests/prov/efa/Makefile.include
@@ -34,8 +34,10 @@ bin_PROGRAMS += prov/efa/src/fi_efa_rnr_read_cq_error \
 		prov/efa/src/fi_efa_rnr_queue_resend \
 		prov/efa/src/fi_efa_info_test
 if HAVE_VERBS_DEVEL
-bin_PROGRAMS += prov/efa/src/fi_efa_exhaust_mr_reg_rdm_pingpong \
-		prov/efa/src/fi_efa_rdma_checker
+bin_PROGRAMS += prov/efa/src/fi_efa_exhaust_mr_reg_rdm_pingpong
+if BUILD_EFA_RDMA_CHECKER
+bin_PROGRAMS += prov/efa/src/fi_efa_rdma_checker
+endif BUILD_EFA_RDMA_CHECKER
 endif HAVE_VERBS_DEVEL
 
 efa_rnr_srcs = \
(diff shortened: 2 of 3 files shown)


shijin-aws:
This should fix Intel CI issue that cannot build efa specific fabtests

fabtests/configure.ac
```diff
@@ -80,6 +80,10 @@ AS_IF([test -z "$CFLAGS"],
 # <3), but it is necessary in AM 1.12.x.
 m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 
+dnl Call the provider's CONFIGURE and CONDITIONALS macros
+m4_ifnblank(m4_esyscmd(ls [prov/efa/configure.m4] 2> /dev/null),
```
j-xiong, line 84: Since configure.m4 is added at the same time, any reason the existence check is needed?

shijin-aws: nope, removed

fabtests/prov/efa/Makefile.include
```diff
@@ -67,9 +69,10 @@ prov_efa_src_fi_efa_exhaust_mr_reg_rdm_pingpong_SOURCES = \
 	$(benchmarks_srcs)
 prov_efa_src_fi_efa_exhaust_mr_reg_rdm_pingpong_LDADD = libfabtests.la
 
+if BUILD_EFA_RDMA_CHECKER
 prov_efa_src_fi_efa_rdma_checker_SOURCES = \
 	prov/efa/src/efa_rdma_checker.c
 prov_efa_src_fi_efa_rdma_checker_LDADD = libfabtests.la
-prov_efa_src_fi_efa_rdma_checker_LDFLAGS = -lefa
```
j-xiong, line 73: I would keep this one and avoid changing LIBS.

fabtests/prov/efa/configure.m4
```diff
@@ -0,0 +1,33 @@
+dnl
+dnl SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only
+dnl SPDX-FileCopyrightText: Copyright Amazon.com, Inc. or its affiliates. All rights reserved.
+dnl
+dnl Configury specific to the fabtests Amazon EFA provider
+
+
```
j-xiong, line 7: Suggest save LIBS here and restore to the saved value at the end.  This can limit "-lefa" to only those places needing it.

shijin-aws: If I keep the LDFLAGS change this shouldn't be needed right? This I won't touch LIBS

fabtests/prov/efa/Makefile.include
shijin-aws: Thx, updated

zachdworkin:
The current Intel CI run looks okay with just this change applied. I will test it in combination with the other one now

zachdworkin:
> The current Intel CI run looks okay with just this change applied. I will test it in combination with the other one now

the --enable-efa=no option and these changes both allow fabtests to be built on our node which is missing the package

zachdworkin:
The build that is failing now is SHMEM middleware. Its not finding stdlib.h... im going to re-run

fabtests/prov/efa/configure.m4
j-xiong: yes, this shouldn't be needed any more. I was thinking it might be needed by the following AC_CHECK's, but apparently those only check for headers.

fabtests/prov/efa/configure.m4
```diff
@@ -0,0 +1,32 @@
+dnl
+dnl SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only
+dnl SPDX-FileCopyrightText: Copyright Amazon.com, Inc. or its affiliates. All rights reserved.
+dnl
+dnl Configury specific to the fabtests Amazon EFA provider
```
darrylabbate, line 5: nit: `Configury`

shijin-aws: fixed

aingerson (approved):
Fixes the issue I'm seeing - thanks!

j-xiong (approved):

zachdworkin (approved):

PR #10421, opened 2024-09-27
prov/efa: Fill the gap in Read NACK protocol 

jiaxiyan:
1. Correctly handle fallback longcts-rtw send completion
2. When FI_DELIVERY_COMPLETE is requested, also remove NACK packet from rxe map after recv completed

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -1066,6 +1066,12 @@ void efa_rdm_ope_handle_recv_completed(struct efa_rdm_ope *ope)
 		efa_rdm_rxe_report_completion(rxe);
 	}
 
+	if (ope->internal_flags & EFA_RDM_OPE_READ_NACK) {
+		assert(ope->type == EFA_RDM_RXE);
+		/* Apply to both DC and non-DC */
+		efa_rdm_rxe_map_remove(&ope->ep->rxe_map, ope->msg_id, ope->peer->efa_fiaddr, ope);
+	}
+
 	/* As can be seen, this function does not release rxe when
 	 * efa_rdm_ope_post_send_or_queue() was successful.
 	 *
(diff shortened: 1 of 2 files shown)


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1088,7 +1088,8 @@ void efa_rdm_ope_handle_recv_completed(struct efa_rdm_ope *ope)
 			efa_rdm_rxe_handle_error(rxe, -err, FI_EFA_ERR_PKT_POST);
 			efa_rdm_rxe_release(rxe);
 		}
-
+		if (rxe->internal_flags & EFA_RDM_OPE_READ_NACK)
```
shijin-aws, line 1091: Why not just move the `efa_rdm_rxe_map_remove` in L1110 before the DC flag check?

jiaxiyan: I wasn't sure if it is safe. I guess it is fine since rxe is not touched.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1066,6 +1066,10 @@ void efa_rdm_ope_handle_recv_completed(struct efa_rdm_ope *ope)
 		efa_rdm_rxe_report_completion(rxe);
 	}
 
+	if (ope->internal_flags & EFA_RDM_OPE_READ_NACK)
+		/* Apply to both DC and non-DC */
```
shijin-aws, line 1070: Nit: Better to add `assert(ope->type == EFA_RDM_TXE)` here as a sanity check

shijin-aws (approved):

prov/efa/src/rdm/efa_rdm_ope.c
shijin-aws: typo, TXE->RXE in earlier comment

PR #10433, opened 2024-10-02
core: Introduce new inject calls that take buffer descriptor

j-xiong:
Existing inject calls (for msg, tagged, rma, and atomic ops) don't provide a way to supply the memory descriptor of the data buffer. This presents difficulty when FI_HMEM is enabled. Providers either have to detect the HMEM interface the buffer is allocated with, or declare the supported inject_size be zero. Both may result in sub-optimal application performance.

Here a set of new inject calls are introduced. The difference of the new calls vs the corresponding existing calls is the addition of a new parameter "desc", like what have already existed in the send calls. With the new calls, the provider is able to know the iface of the data buffer and handle it properly.

### DIFF
--- a/include/rdma/fi_atomic.h
+++ b/include/rdma/fi_atomic.h
@@ -138,6 +138,10 @@ struct fi_ops_atomic {
 			enum fi_datatype datatype, enum fi_op op, size_t *count);
 	int	(*compwritevalid)(struct fid_ep *ep,
 			enum fi_datatype datatype, enum fi_op op, size_t *count);
+
+	ssize_t	(*inject2)(struct fid_ep *ep, const void *buf, size_t count,
+			void *desc, fi_addr_t dest_addr, uint64_t addr,
+			uint64_t key, enum fi_datatype datatype, enum fi_op op);
 };
 
 #ifdef FABRIC_DIRECT
@@ -182,6 +186,17 @@ fi_inject_atomic(struct fid_ep *ep, const void *buf, size_t count,
 {
 	return ep->atomic->inject(ep, buf, count, dest_addr, addr,
 			key, datatype, op);
+}
+
+static inline ssize_t
+fi_inject_atomic2(struct fid_ep *ep, const void *buf, size_t count, void *desc,
+		  fi_addr_t dest_addr, uint64_t addr, uint64_t key,
+		  enum fi_datatype datatype, enum fi_op op)
+{
+	return FI_CHECK_OP(ep->atomic, struct fi_ops_atomic, inject2) ?
+		ep->atomic->inject2(ep, buf, count, desc, dest_addr, addr, key,
+				    datatype, op) :
+		-FI_ENOSYS;
 }
 
 static inline ssize_t
--- a/include/rdma/fi_endpoint.h
+++ b/include/rdma/fi_endpoint.h
@@ -128,6 +128,10 @@ struct fi_ops_msg {
 			uint64_t data, fi_addr_t dest_addr, void *context);
 	ssize_t	(*injectdata)(struct fid_ep *ep, const void *buf, size_t len,
 			uint64_t data, fi_addr_t dest_addr);
+	ssize_t	(*inject2)(struct fid_ep *ep, const void *buf, size_t len,
+			void *desc, fi_addr_t dest_addr);
+	ssize_t	(*injectdata2)(struct fid_ep *ep, const void *buf, size_t len,
+			void *desc, uint64_t data, fi_addr_t dest_addr);
 };
 
 struct fi_ops_cm;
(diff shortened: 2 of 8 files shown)


shefty:
A purpose of the inject call is that the buffer doesn't have to be registered.  If it needs to be registered, the user can go through any of the other send operations.  It seems the gap here is to disallow hmem buffers from being passed into the inject call if FI_MR_HMEM is set.

j-xiong:
The actual usage of the inject calls can deviate. There are many cases that the buffer is pre-registered anyway (e.g. MPI eager buffer), and the user just wants the "not holding the buffer and no completion" semantics of the inject calls. Although that can be achieved with regular send with FI_INJECT flag and selective completion, but it is more cumbersome that way.

shefty:
The normal send call is usable, however.  Here, the app has already registered the buffer.  But the app wants the provider to copy the GPU memory somewhere, so that when the call returns the app can own the buffer again.  I'm struggling to see the benefit of introducing a new call or to follow the application use case.  Copying off GPU memory isn't exactly fast, which is what the inject call targets.

j-xiong:
The difficulty comes from handling the inject calls over the LNX provider when the underlying providers have different inject size and FI_MR_HMEM settings. For example, the CXI provider doesn't need FI_MR_HMEM mode and can always return positive inject size; but the SHM provider requires FI_MR_HMEM mode and has to return 0 inject size when FI_HMEM is enabled. If the LNX provider takes the conservative way by setting inject size to 0 when FI_HMEM is enabled, the application can't use inject calls at all even if the traffic goes through CXI. If the LNX provider setting the inject size to the value returned by CXI, it needs to handle device buffer manually if the traffic chooses SHM -- either register + copy + inject, or register + send.  The registration here can be redundant if the buffer has already been registered by the application.  Application can call these new inject functions if the buffer has already been registered, saving the provider from a duplicated registration.

shefty:
The problem is the current inject calls are not usable.  That needs to be indicated somehow, e.g. by restricting inject to host buffers only by definition, or have separate inject sizes for host vs hmem buffers or updating FI_MR_HMEM definition somehow.  That needs to be fixed, otherwise, we're simply introducing magic developer required knowledge as to why one call needs to be used over another.

In this case, new API calls are for convenience.  They aren't needed and likely have minimal performance benefit over a common path.

If LNX can return an inject size for transfers that go over SHM, then SHM should be able to support the same inject size directly.

amirshehataornl:
When you say: "A purpose of the inject call is that the buffer doesn't have to be registered. " What line of reasoning was used to come to that determination?

shefty:
Because there's no damn place to provide the memory descriptor.  Seems pretty damn obvious.

amirshehataornl:
Don't understand the tone of the response. It does not answer my question.

The lack for the descriptor in the original APIs might be due to a historical reason? That was my question and your answer says that the reason the descriptor is not there, is because it's not there.

As far as I understand the inject call is used to inject the data in the message in order not to incur the performance penalty of a more complex protocol. With the introduction of HMEM there is now a reason to add the descriptor, because the data to be injected resides in HMEM. The way CXI got around this is to do a lookup on the address and cache the result. At least in the context of CXI the inject path actually adds some significant performance gains.

aingerson:
I think what grumpy Sean is saying is that the behavior of the inject (in terms of dropping the completion and returning the buffer immediately) can be replicated with the sendmsg call. So the only additional thing the call is adding is that you don't need to register the buffer, hence the original reasoning behind the call. If we add fi_inject2 with the descriptor, it conflicts with what that call is providing (lack of registration). So what we should be doing is specifying that the inject calls can only be called with host memory (or memory does not have to be registered) and then move the middlewares to using fi_sendmsg() to replicate the inject behavior that provides the optimization.

shijin-aws:
I agree with Sean here, I prefer to keep fi_*inject* call without MR descriptor to indicate an MR is not needed & the send buffer is host type (when FI_MR_HMEM is required) so we have a clear logic.

shefty:
@amirshehataornl My 'damn response' was not personally directed and was because I read into the question a tone that you didn't intend.  I didn't read it as asking a legitimate question.  Sorry.

The inject call is a highly optimized API for low latency transfers involving small messages.  The intent of the call is to enable passing the user's data directly as part of a work request.  This allows HW to perform a single PCI transaction to get both the command and the associated data.

Semantically, from the viewpoint of the application, the buffer is owned by the application on return from the call.  The provider is required to copy the data into some network accessible buffer.  The user data may be declared on the stack, for example.  A memory descriptor is not part of the API, as it is not used or needed by the provider.  In the best case, the provider would ignore the parameter, but the app would still incur the cost of passing it to the function.  In the worst case, the provider reads the descriptor and performs some branch based on the result.

Even this proposal does not need a memory descriptor.  The descriptor is used as an indirect method to identify if a buffer resides on the GPU.  A more direct method, such as passing in the HMEM type and device ID, would allow the same check and work with non-registered buffers.

Passing a GPU buffer into inject requires copying the data that's located on the GPU into a network accessible buffer.  In order to support this, inject needs additional parameters.  The provider needs to branch on that data.  And the data must be copied over PCI into the network buffer.  Finally, the WR can be submitted.  Copying data from the GPU is slow, which makes it questionable why support inject from GPU buffers at all.  I'm claiming that inject should not support GPU buffers, or, more specifically, FI_MR_HMEM means that HMEM buffers cannot be passed into inject.

Note that a provider does not need to copy the inject data into the WR.  Semantically, copying it into a registered bounce buffer would suffice; however, the performance may be slower.  HW would fetch the WR, decode it, then need to fetch the data across PCI.  This is internal to the provider.  By NOT having a memory descriptor be part of the API, it is a signal to the provider implementation that the buffer is owned by the caller on return.

amirshehataornl:
@shefty, no problem. Thanks for the explanation.

j-xiong:
Whatever direction we take, I agree that we need to update the man pages indicating that current fi_inject* calls can only take system memory buffer when FI_MR_HMEM mode is set.

@amirshehataornl With the semantics clarification and w/o new inject calls, the shm provider can return non-zero inject size and set FI_MR_HMEM mode. The lnx provider has two options: (1) set FI_MR_HMEM mode, application will only call fi_inject with system memory buffer; (2) doesn't set FI_MR_HMEM mode, application can call fi_inject with any buffer, but lnx needs to detect non-memory type and register and call send instead if the shm path is chosen. Do you think current performance issue can be eliminated with either of these?

PR #10435, opened 2024-10-03
Fabtests: Added inband sync to ft_init_fabric_cm.

nikhilnanal:
Added condition to make preposted rx buffers based on FT_OPT_NO_PRE_POSTED_RX in case of msg endpoint based init code in fabtests. Changes similar to https://github.com/ofiwg/libfabric/pull/10394


j-xiong (commented):
The patch adds missing inband sync to ft_init_fabric_cm to properly handle the case that rx buffers are not pre-posted. Similar changes have already been made to ft_init_fabric. Please revise the title and the description to reflect what is really being done here.

j-xiong (approved):

PR #10443, opened 2024-10-07
fabtests: Multinode update

amirshehataornl:
Allow the -E parameter in the harness code. It enables out of band address exchange which is needed by CXI.

Adjust the runmultinode.py script to use the correct command line options.

### DIFF
--- a/fabtests/multinode/src/harness.c
+++ b/fabtests/multinode/src/harness.c
@@ -355,7 +355,7 @@ int main(int argc, char **argv)
 
 	opts = INIT_OPTS;
 	opts.options |= FT_OPT_SIZE | FT_OPT_OOB_ADDR_EXCH |
-			FT_OPT_DISABLE_TAG_VALIDATION;
+			FT_OPT_ADDR_IS_OOB | FT_OPT_DISABLE_TAG_VALIDATION;
 
 	pm_job.clients = NULL;
 	pm_job.pattern = -1;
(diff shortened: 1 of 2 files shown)


aingerson (commented):
This seems like 2 changes (the -C -> -x change and the -E change). In addition, OOB address exchange change is not needed. The multinode test always does an OOB address exchange
See https://github.com/ofiwg/libfabric/blob/main/fabtests/multinode/src/harness.c#L357

amirshehataornl:
I think my comment is not entirely clear. The issue really is when  you specify the -E and you don't specify the oob_addr then you set FT_OPT_ADDR_IS_OOB. 
```
3483 »·······case 'E':
3484 »·······»·······opts->options |= FT_OPT_OOB_ADDR_EXCH;
3485 »·······»·······if (optarg && strlen(optarg) > 1)
3486 »·······»·······»·······opts->oob_port = optarg + 1;
3487 »·······»·······else
3488 »·······»·······»·······opts->oob_port = default_oob_port;
3489 »·······»·······if (!opts->oob_addr)
3490 »·······»·······»·······opts->options |= FT_OPT_ADDR_IS_OOB;
3491 »·······»·······break;
```
When FT_OPT_ADDR_IS_OOB is set then when you call fi_getinfo you don't specify the service and the node, which is what CXI needs.

Is there are reason the code sets FT_OPT_OOB_ADDR_EXCH but not FT_OPT_ADDR_IS_OOB.

Also I'm not entirely clear on your comment. Do you want two separate commits for the -x and the -E change?

aingerson:
@amirshehataornl Oh ok, that makes more sense. I think it makes sense to add the FT_OPT_ADDR_IS_OOB flag to the multinode test since the address is always expected to be the OOB address of the server. It is probably like that because the flag was just added last year and not incorporated into the multinode test to fully support a provider like cxi

aingerson:
Oh and yes, I think they should be 2 different patches

aingerson (approved):

PR #10447, opened 2024-10-08
prov/efa: Only do dmabuf reg when FI_MR_DMABUF is set

shijin-aws:
Application should use FI_MR_DMABUF API to pass the dmabuf fd and offset to make Libfabric register the mr via dmabuf. The only exception is for synapseai, beacuse dmabuf is the only way to register Gaudi device buffer and it was implemented before the FI_MR_DMABUF API. Keep this behavior unchanged for compatibility.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -529,61 +529,32 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
 		);
 
 	/*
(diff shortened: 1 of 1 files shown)


jiaxiyan (approved):

shijin-aws:
@j-xiong it seems Intel CI is taking forever now. Just want to check if it is caused by anything related by efa changes (unlikely)

j-xiong:
There was an infrastructure maintenance shutdown for the past weekend. Many of the delayed are related to that.

PR #10448, opened 2024-10-09
Fix application + uffd hangs due to not handling page faults

iziemba:
In order to receive unmap events, uffd uses 'mode missing' when registering memory regions. This implies getting page fault events as well. So handle them by returning a zero-filled page.

### DIFF
--- a/prov/util/src/util_mem_monitor.c
+++ b/prov/util/src/util_mem_monitor.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2017-2021 Intel Inc. All rights reserved.
  * Copyright (c) 2019-2021 Amazon.com, Inc. or its affiliates.
  *                         All rights reserved.
- * (C) Copyright 2020 Hewlett Packard Enterprise Development LP
+ * (C) Copyright 2020,2024 Hewlett Packard Enterprise Development LP
  * Copyright (C) 2024 Cornelis Networks. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


prov/util/src/util_mem_monitor.c
```diff
@@ -606,6 +612,80 @@ static void *ofi_uffd_handler(void *arg)
...
+				/* setup a zeropage reqest for this pagesize */
+				zp.range.start = (uint64_t) (uintptr_t)
+					ofi_get_page_start(address, page_sizes[i]);
+				zp.range.len = (uint64_t) page_sizes[i];
+				zp.mode = 0;
+				zp.zeropage = 0;
+
+				ret = ioctl(uffd.fd, UFFDIO_ZEROPAGE, &zp);
+
+				if (0 == ret) {		/* success */
```
j-xiong: Please put the constant at the right side of the comparison.  Not about which way is better, but to keep the convention consistent. Same for other similar instances.

iziemba: Will do. Also, I think this PR can be squashed into a single commit.

j-xiong: Please also fix those similar places that this comment applies to.

iziemba: Done.

a-szegel:
bot:aws:retest

PR #10455, opened 2024-10-13
use new synapse api

imasuari:
synapse-ai API now expose the base address we need directly, making the code more clean and simplified

### DIFF
--- a/fabtests/common/hmem_synapseai.c
+++ b/fabtests/common/hmem_synapseai.c
@@ -41,31 +41,6 @@
 #include "habanalabs/synapse_api.h"
 #include "habanalabs/hlthunk.h"
 
-#define SCAL_SUCCESS 0
-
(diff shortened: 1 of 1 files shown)


imasuari:
@nbeer-habana

j-xiong (commented):
The changes look good to me. There is a question though, the implementation of the get_dmabuf_fd method is different from the one in libfabirc (src/hmem_synapseai.c). What is the difference between the two?

zachdworkin:
@imasuari it looks like one of our nodes was stuck in a bad state. I am rebooting it and will re-run your PR.

shijin-aws:
> The changes look good to me. There is a question though, the implementation of the get_dmabuf_fd method is different from the one in libfabirc (src/hmem_synapseai.c). What is the difference between the two?

I was asking the same question last time. But they intend to do that to keep backward compatibility?

j-xiong (approved):

PR #10460, opened 2024-10-14
Cornelis Networks upstream 10/14/2024

charlesshereda:
Cornelis Networks upstream for the OPX provider and a few other things for October 2024.

This PR includes the following:

- Ensure that only one completion is posted for RZV truncation receives
- Removal of FI_CONTEXT2 requirement
- Env var changes
- OPX-specific configure.m4 changes to support ROCR
- A number of Github Actions workflow changes
- CN5000-related changes to the scb
- Link bounce support for OPA100
- Documentation for FI_OPX_RELIABILITY_MAX_UNCONGESTED_PINGS and FI_OPX_RELIABILITY_MAX_CONGESTED_PINGS
- SDMA support for 16B headers for CN5000
- Several fixes for the enablement of 16B header support for CN5000
- Removal of an unnecessary/problematic polling call
- Fix to credit return
- OPX Tracer additions to RMA code paths
- Code cleanup and function simplification
- Initialization of nic info
- Immediate block offset calculation fix
- Addition of FI_OPX_TID_MIN_PAYLOAD_BYTES env var
- Fix to a replay payload copy
- 16B header eager and multipacket eager fixes
- Fix to the last_bytes field for an SDMA replay
- Update of fi_info -e description for FI_OPX_UUID
- Bug fix for an ignored context creation error
- Reduced amount of immediate data in RTS packet to improve rzv performance
- Change to conditionally set FI_REMOTE_CQ_DATA
- Debug check for zero-length packets
- Set the _immediate acknowledgement requested_ bit when sending the last packet of an RMA PUT


man/man7/fi_opx.7
```diff
@@ -163,6 +163,22 @@ inclusive.
...
+Default setting is 128. Range of valid values is 1-65535.
+.TP
+\f[I]FI_OPX_RELIABILITY_MAX_CONGESTED_PINGS\f[R]
+Integer.
+This setting controls how many PING requests the reliability/replay
+function will issue per iteration of FI_OPX_RELIABILITY_SERVICE_USEC_MAX 
+in situations with more contending, outgoing traffic from the HFI.
+.PP
+Default setting is 4. Range of valid values is 1-65535.
+.TP
```
j-xiong, line 181: FYI, man pages are updated automatically from the MarkDown files and there is no need to include the change here.

prov/opx/src/fi_opx_init.c
```diff
@@ -396,6 +396,9 @@ static int fi_opx_fillinfo(struct fi_info *fi, const char *node,
 			fi->ep_attr->rx_ctx_cnt = hints->ep_attr->rx_ctx_cnt;	/* TODO - check */
 	}
 
+	fi->nic = ofi_nic_dup(NULL);
+	fi->nic->bus_attr->bus_type = FI_BUS_PCI;
```
j-xiong, line 400: Are the pci attributes set anywhere? otherwise may just leave the bus type as unknown.

charlesshereda: @lsavers Can you answer this question?

man/man7/fi_opx.7
charlesshereda: @j-xiong If you want, I'll push an updated PR without the changes to fi_opx.7. Or we can just leave it and it will be written over. Any preference?

j-xiong: It's fine to leave it as it is.

prov/opx/src/fi_opx_init.c
charlesshereda: In discussing with Lindsay, we needed to initialize the NIC in order to get (we believe) NCCL to work; but we aren't sure whether we really needed to set bus type or just call ofi_nic_dup(NULL). Lindsay is going to retest just without setting bus type today, but we think at least line 399 is necessary to get past the issue she was seeing.

lsavers: I was able to verify that both lines are needed to get past a NCCL initialization error for OFI_NCCL_PROTOCOL=RDMA (https://github.com/aws/aws-ofi-nccl/blob/ecbf29867f0f772aea2767447dd126ebdbc45546/src/nccl_ofi_topo.c#L123).  We do have a future work item to add code to set the additional attribute values; however, this work has not yet been completed.

j-xiong: thanks for the verification.

j-xiong (approved):

PR #10471, opened 2024-10-18
fabtests/pytest/efa: Loose assertion for read request counters

shijin-aws:
The current assertion check for send read wrs and bytes are based on the exact calculation of benchmark data size + fabtests control message size, which is too strict. Fabtests may change control message sizes which should not impact this test.

This patch looses the assertion for read request counters which make it have enough coverage for the benchmark data size.

### DIFF
--- a/fabtests/pytest/efa/test_runt.py
+++ b/fabtests/pytest/efa/test_runt.py
@@ -74,16 +74,15 @@ def test_runt_read_functional(cmdline_args, memory_type, copy_method):
     if copy_method == "localread":
         # when local read copy is used, server issue RDMA requests to copy received data
         #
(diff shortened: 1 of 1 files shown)


shijin-aws:
Fix the fabtest failure caused by https://github.com/ofiwg/libfabric/commit/0c635e7c2f10aa476005afac2e87e252480a764c (cc @aingerson )

aingerson:
@shijin-aws Oops, sorry! How did it pass the CI for my PR?

shijin-aws:
@aingerson no worries, this test is only run on GPU platform which our public CI doesn't run due to capacity

jiaxiyan (approved):

PR #10480, opened 2024-10-22
Multinode test script update

nikhilnanal:
Improvements to runmultinode.sh script

### DIFF
--- a/fabtests/scripts/runmultinode.sh
+++ b/fabtests/scripts/runmultinode.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 Options=$(getopt --options h:,n:,p:,I:,-x:,z: \
-		  		--longoptions hosts:,processes-per-node:,provider:,xfer-method:,iterations:,ci:,cleanup,help \
+				--longoptions hosts:,processes-per-node:,provider:,xfer-method:,iterations:,ci:,cleanup,help \
 				-- "$@")
 
 eval set -- "$Options"
@@ -10,7 +10,7 @@ hosts=[]
 ppn=1
 iterations=1
 pattern=""
-xfer-method="msg"
+xfer_method="msg"
 cleanup=false
 help=false
 ci=""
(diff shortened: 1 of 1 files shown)


fabtests/scripts/runmultinode.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 Options=$(getopt --options h:,n:,p:,I:,-x:,z: \
-		  		--longoptions hosts:,processes-per-node:,provider:,xfer-method:,iterations:,ci:,cleanup,help \
+				--longoptions hosts:,processes-per-node:,provider:,xfer_method:,iterations:,ci:,cleanup,help \
```
j-xiong: Why would `--xfer-method` cause problem?  Did you use `-` instead of `--` (notice that `-x` is a valid option).

nikhilnanal: The script gets interpreted as  "..... -p 'verbs;ofi_rxm' -x -method -z full_mesh -I 1 -T....". I guess i misunderstood the error - The '-' is actually causing a misinterpretation of the of the bash variable xfer-method. this can be fixed using curly braces. However, "-" is not recommended in variable names to avoid such errors so i think using "_" is still better. I will correct the reason in the commit message

fabtests/scripts/runmultinode.sh
```diff
@@ -31,6 +31,16 @@ while true; do
...
 			xfer_method="$2"; shift 2 ;;
+		-E|--env)
+			delimiter="="
+			value=${2#*$delimiter}
+			var=${2:0:$(( ${#2} - ${#value} - ${#delimiter} ))}EXPORT_STRING="export $var=\"$value\""
+			if [[ -z $EXPORT_ENV ]] ; then
+				EXPORT_ENV="$EXPORT_STRING ;"
+			else
+				EXPORT_ENV="$EXPORT_ENV $EXPORT_STRING ;"
+			fi
```
j-xiong, line 42: No need to use conditional here. Just use the statement in the `else` branch directly.

j-xiong (approved):

PR #10482, opened 2024-10-23
fabtest: Fix compilation error with C99

j-xiong:
CMPLX is introduced in C11, it may be undefined in C99. Compilation error was observed with Coverity scan.

### DIFF
--- a/fabtests/include/unix/osd.h
+++ b/fabtests/include/unix/osd.h
@@ -125,7 +125,7 @@ static inline void OFI_COMPLEX_OP(name, set)(OFI_COMPLEX(name) *v1, OFI_COMPLEX(
 }												      \
 static inline void OFI_COMPLEX_OP(name, fill)(OFI_COMPLEX(name) *v1, name v2)			      \
 {												      \
-	*v1 = CMPLX(v2, v2);							  		      \
+	*v1 = (OFI_COMPLEX(name))((name)(v2) + I * (name)(v2));			  		      \
 }
 
 OFI_COMPLEX_OPS(float)


fabtests/include/unix/osd.h
```diff
@@ -94,6 +94,11 @@ OFI_COMPLEX_TYPE_DECL(float, float)
 OFI_COMPLEX_TYPE_DECL(double, double)
 OFI_COMPLEX_TYPE_DECL(long_double, long double)
 
+/* CMPLX may be undefined before C11 */
+#ifndef CMPLX
+#define CMPLX(r, i) ((double complex)((double)(r) + _Imaginary_I * (double)(i)))
```
aingerson, line 99: Oops, sorry!
What is the method used to define an atomic before C11? Does this work for other types (like long double)?

j-xiong: The macro here is what is defined in the C11 header.

j-xiong: So this should be equivalent to the original code. There is CMPLXF and CMPLXL for float and long complex numbers.

j-xiong:
```
#define CMPLX(x, y) ((double complex)((double)(x) + _Imaginary_I * (double)(y)))
#define CMPLXF(x, y) ((float complex)((float)(x) + _Imaginary_I * (float)(y)))
#define CMPLXL(x, y) ((long double complex)((long double)(x) + \
                      _Imaginary_I * (long double)(y)))
```

j-xiong: `_Imaginary_I` is defined since C99.

j-xiong: Updated. Should handle other types better now.

fabtests/include/unix/osd.h
```diff
@@ -94,6 +94,8 @@ OFI_COMPLEX_TYPE_DECL(float, float)
 OFI_COMPLEX_TYPE_DECL(double, double)
 OFI_COMPLEX_TYPE_DECL(long_double, long double)
 
+typedef long double long_double;
```
aingerson, line 97: This is a duplicate of line 86

aingerson:
Looks like OSX didn't like `_Imaginary_I` :(
```
./include/unix/osd.h:135:1: error: use of undeclared identifier '_Imaginary_I'
./include/unix/osd.h:130:39: note: expanded from macro 'OFI_COMPLEX_OPS'
        *v1 = OFI_COMPLEX(name)((name)(v2) + _Imaginary_I * (name)(v2));                              \
                                             ^
```

j-xiong:
Updated to use `I` instead, which should be defined as long as `complex.h` is present.

PR #10488, opened 2024-10-24
prov/efa: Set max rma order size correctly

shijin-aws:
EFA device doesn't support any ordering for
RMA and atomic operations. RDM endpoint supports
emulated ordered atomic protocol with max order
size as the max atomic size. EFA provider should
only return these non-zero for rdm ep type when
ordered atomic is requested: FI_ORDER_ATOMIC_*.


prov/efa/src/efa_user_info.c
```diff
@@ -418,11 +416,14 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
...
 			info->tx_attr->msg_order &= hints->tx_attr->msg_order;
-			atomic_ordering = FI_ORDER_ATOMIC_RAR | FI_ORDER_ATOMIC_RAW |
-					  FI_ORDER_ATOMIC_WAR | FI_ORDER_ATOMIC_WAW;
-			if (!(hints->tx_attr->msg_order & atomic_ordering)) {
+
+			/* If no atomic ordering is requested, set the max_order_*_size as 0 */
+			if (!(hints->tx_attr->msg_order & FI_ORDER_ATOMIC_RAW))
 				info->ep_attr->max_order_raw_size = 0;
-			}
+			if (!(hints->tx_attr->msg_order & FI_ORDER_ATOMIC_WAR))
```
shijin-aws, line 423: We were setting three `max_order_*size` as max_atomic_size before this commit https://github.com/ofiwg/libfabric/commit/0abaf5c2ea513f2613ca981189e13d07c603aec6#diff-3040bb85252dac4eb9962226fe26a0f3bc420994e7a26fccbf18156b24f03c6cR401, which only set `max_order_raw_size` when atomic ordering is requested. I think it was a miss in that PR. This PR fixes it.

shijin-aws: MPICH/Intel MPI requested these 3 ordering semantics https://github.com/search?q=repo%3Apmodels%2Fmpich%20max_order_&type=code, we should be able to test this by CI.

prov/efa/src/efa_prov_info.c
```diff
@@ -145,7 +145,9 @@ const struct fi_ep_attr efa_ep_attr = {
 	.protocol		= FI_PROTO_EFA,
 	.protocol_version	= 1,
 	.msg_prefix_size	= 0,
+	.max_order_raw_size	= 0,
 	.max_order_war_size	= 0,
+	.max_order_waw_size	= 0,
```
darrylabbate, line 150: These would be implicitly zero-initialized. NBD.

darrylabbate (approved):

PR #10492, opened 2024-10-25
prov/efa: Make efa_hmem_info a global variable

jiaxiyan:
Currently efa_hmem_info is part of efa_domain and created for every efa domain. hmem_info init involves several operations like device memory allocation / free, and trial ibv reg mr, which is expensive and can potentially cause more memory usage. Make efa_hmem_info a global variable and call it only once per process. Remove p2p_disabled_by_user from efa_hmem_info since it is only used for ep level operations, not global.

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -5,16 +5,18 @@
 #include "efa_hmem.h"
 #include "rdm/efa_rdm_pkt_type.h"
 
+struct efa_hmem_info g_efa_hmem_info[OFI_HMEM_MAX];
+
 #if HAVE_CUDA || HAVE_NEURON
-static size_t efa_max_eager_msg_size_with_largest_header(struct efa_domain *efa_domain) {
+static size_t efa_max_eager_msg_size_with_largest_header() {
 	int mtu_size;
 
-	mtu_size = efa_domain->device->rdm_info->ep_attr->max_msg_size;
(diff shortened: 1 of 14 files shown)


prov/efa/src/efa_hmem.c
```diff
@@ -283,20 +282,18 @@ efa_domain_hmem_info_init_iface(struct efa_domain *efa_domain, enum fi_hmem_ifac
 	* memory, therefore p2p is not required.
 	*/
 	if (iface == FI_HMEM_CUDA &&
-	    FI_VERSION_GE(efa_domain->util_domain.fabric->fabric_fid.api_version, FI_VERSION(1, 18)))
+	    FI_VERSION_GE(g_device_list[0].rdm_info->fabric_attr->api_version, FI_VERSION(1, 18)))
```
shijin-aws, line 285: Before your change, the `api_version` is the one passed from `fi_getinfo(FI_VERSION(...))`, after your change, I think it is the api version of the current libfabric library (please double check on this). They are different so I think you'd need to move `p2p_required_by_impl` from this struct too.

shijin-aws: We only have conditional p2p_required_by_impl for cuda, I think it's easy to just define this within the p2p check in ep.

shijin-aws (approved):
LGTM, thanks!

shijin-aws:
bot:aws:retest

PR #10493, opened 2024-10-25
prov/udp: detect and use interface MTU to support large messages over UDP

mpokorny:
This commit supports a maximum UDP message size that depends on the MTU size of the interface. The use case is UDP messaging on an Ethernet network configured to use jumbo frames.

For each interface detected by the udp provider, determine the MTU of the interface, and use that value to set the `max_msg_size` field of the `fi_ep_attr` and `fi_tx_attr `values of the fi_info element. When the MTU cannot be determined, the MTU value assumed by previous code versions (1500) is used.

### DIFF
--- a/include/freebsd/osd.h
+++ b/include/freebsd/osd.h
@@ -76,6 +76,11 @@ static inline size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
 	return 0;
 }
 
+static inline int ofi_ifaddr_get_mtu(const struct ifaddrs *ifa)
+{
+	return -1;
+}
+
 static inline ssize_t ofi_process_vm_readv(pid_t pid,
 			const struct iovec *local_iov,
 			unsigned long liovcnt,
@@ -185,5 +190,3 @@ ofi_recvv_socket(SOCKET fd, const struct iovec *iov, size_t cnt, int flags)
 }
 
 #endif /* _FREEBSD_OSD_H_ */
-
-
--- a/include/ofi_net.h
+++ b/include/ofi_net.h
@@ -655,6 +655,7 @@ struct ofi_addr_list_entry {
 	char			ipstr[INET6_ADDRSTRLEN];
 	union ofi_sock_ip	ipaddr;
 	size_t			speed;
+	int			mtu;
 	char			net_name[OFI_ADDRSTRLEN];
 	char			ifa_name[OFI_ADDRSTRLEN];
 	uint64_t		comm_caps;
--- a/man/fi_udp.7.md
+++ b/man/fi_udp.7.md
@@ -41,9 +41,10 @@ receiving datagram messages over an unreliable endpoint.
 
 # LIMITATIONS
 
-The UDP provider has hard-coded maximums for supported queue sizes and data
-transfers.  These values are reflected in the related fabric attribute
-structures
+The UDP provider has a hard-coded maximum for supported queue sizes,
+which is reflected in the related fabric attribute structures. Maximum
+size of data transfers is limited by the MTU size of an interface, and
+is also reflected in the related fabric attribute structures.
 
 EPs must be bound to both RX and TX CQs.
 
@@ -53,7 +54,10 @@ No support for counters.
 
 # RUNTIME PARAMETERS
 
-No runtime parameters are currently defined.
+The UDP provider checks for the following environment variables -
+
+*FI_UDP_IFACE*
+: An string value that specifies the name of the interface.
 
 # SEE ALSO
 
--- a/prov/udp/src/udpx.h
+++ b/prov/udp/src/udpx.h
@@ -63,11 +63,9 @@
 #ifndef _UDPX_H_
 #define _UDPX_H_
 
-
 extern struct fi_provider udpx_prov;
 extern struct util_prov udpx_util_prov;
 extern struct fi_info udpx_info;
-
 
 int udpx_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric,
 		void *context);
@@ -75,10 +73,12 @@ int udpx_domain_open(struct fid_fabric *fabric, struct fi_info *info,
 		struct fid_domain **dom, void *context);
 int udpx_eq_open(struct fid_fabric *fabric, struct fi_eq_attr *attr,
 		struct fid_eq **eq, void *context);
-
+void udpx_util_prov_init(uint32_t version);
+void udpx_util_prov_fini();
 
 #define UDPX_FLAG_MULTI_RECV	1
 #define UDPX_IOV_LIMIT		4
+#define UDPX_MTU		1500
 
 struct udpx_ep_entry {
 	void			*context;
@@ -87,6 +87,10 @@ struct udpx_ep_entry {
 	uint8_t			flags;
 	uint8_t			resv[sizeof(size_t) - 2];
 };
+
+#define UDPX_UDP_HEADER_SIZE	8
+#define UDPX_IP_HEADER_SIZE	20
+#define UDPX_MAX_MSG_SIZE(mtu) ((mtu) - (UDPX_UDP_HEADER_SIZE + UDPX_IP_HEADER_SIZE))
 
 OFI_DECLARE_CIRQUE(struct udpx_ep_entry, udpx_rx_cirq);
 
--- a/prov/udp/src/udpx_attr.c
+++ b/prov/udp/src/udpx_attr.c
@@ -31,6 +31,7 @@
  */
 
 #include "udpx.h"
+#include "ofi_osd.h"
 
 #define UDPX_TX_CAPS (OFI_TX_MSG_CAPS | FI_MULTICAST)
 #define UDPX_RX_CAPS (FI_SOURCE | OFI_RX_MSG_CAPS)
@@ -38,7 +39,7 @@
 
 struct fi_tx_attr udpx_tx_attr = {
 	.caps = UDPX_TX_CAPS,
-	.inject_size = 1472,
+	.inject_size = UDPX_MAX_MSG_SIZE(UDPX_MTU),
 	.size = 1024,
 	.iov_limit = UDPX_IOV_LIMIT
 };
@@ -53,7 +54,7 @@ struct fi_ep_attr udpx_ep_attr = {
 	.type = FI_EP_DGRAM,
 	.protocol = FI_PROTO_UDP,
 	.protocol_version = 0,
-	.max_msg_size = 1472,
+	.max_msg_size = UDPX_MAX_MSG_SIZE(UDPX_MTU),
 	.tx_ctx_cnt = 1,
 	.rx_ctx_cnt = 1
 };
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
 
 struct util_prov udpx_util_prov = {
 	.prov = &udpx_prov,
-	.info = &udpx_info,
-	.flags = 0,
+	.info = NULL,
+        .flags = 0,
 };
+
+
+static int match_interface(struct slist_entry *entry, const void *infop)
+{
+	struct ofi_addr_list_entry *addr_entry;
+	const struct fi_info* info = infop;
+
+	addr_entry = container_of(entry, struct ofi_addr_list_entry, entry);
+	return strcmp(addr_entry->net_name, info->fabric_attr->name) == 0 &&
+	       strcmp(addr_entry->ifa_name, info->domain_attr->name) == 0;
+}
+
+static void set_mtu_from_addr_list(struct fi_info* info,
+				   struct slist *addr_list)
+{
+	struct ofi_addr_list_entry *addr_entry;
+	struct slist_entry *entry;
+	int max_msg_size;
+
+	entry = slist_find_first_match(addr_list, match_interface, info);
+	if (entry) {
+		addr_entry = container_of(entry,
+					  struct ofi_addr_list_entry,
+					  entry);
+		max_msg_size = UDPX_MAX_MSG_SIZE(addr_entry->mtu);
+		if (max_msg_size > 0) {
+			info->tx_attr->inject_size = max_msg_size;
+			info->ep_attr->max_msg_size = max_msg_size;
+		}
+	} else {
+		FI_DBG(&udpx_prov, FI_LOG_CORE,
+		       "Failed to match interface (%s, %s) to "
+		       "address for MTU size\n",
+		       info->fabric_attr->name, info->domain_attr->name);
+	}
+}
+
+void udpx_util_prov_init(uint32_t version)
+{
+
+	struct slist addr_list;
+	struct fi_info* cur;
+	struct fi_info* info;
+
+        if (udpx_util_prov.info == NULL) {
+		udpx_util_prov.info = &udpx_info;
+		info = fi_allocinfo();
+		ofi_ip_getinfo(&udpx_util_prov, version, NULL, NULL, 0, NULL,
+			       &info);
+		slist_init(&addr_list);
+		ofi_get_list_of_addr(&udpx_prov, "iface", &addr_list);
+		for (cur = info; cur; cur = cur->next)
+			set_mtu_from_addr_list(cur, &addr_list);
+		*(struct fi_info**)&udpx_util_prov.info = info;
+		ofi_free_list_of_addr(&addr_list);
+	}
+}
+
+void udpx_util_prov_fini()
+{
+	if (udpx_util_prov.info != NULL)
+		fi_freeinfo((struct fi_info*)udpx_util_prov.info);
+}
--- a/prov/udp/src/udpx_init.c
+++ b/prov/udp/src/udpx_init.c
@@ -37,19 +37,24 @@
 
 #include <sys/types.h>
 
+static ofi_mutex_t init_lock;
 
 static int udpx_getinfo(uint32_t version, const char *node, const char *service,
 			uint64_t flags, const struct fi_info *hints,
 			struct fi_info **info)
 {
-	return ofi_ip_getinfo(&udpx_util_prov, version, node, service, flags,
-			      hints, info);
+	ofi_mutex_lock(&init_lock);
+	udpx_util_prov_init(version);
+	ofi_mutex_unlock(&init_lock);
+	return util_getinfo(&udpx_util_prov, version, node, service, flags,
+			    hints, info);
 }
 
 static void udpx_fini(void)
 {
-	/* yawn */
+	udpx_util_prov_fini();
 }
+
 
 struct fi_provider udpx_prov = {
 	.name = "udp",
@@ -65,5 +70,6 @@ UDP_INI
 	fi_param_define(&udpx_prov, "iface", FI_PARAM_STRING,
 			"Specify interface name");
 
+	ofi_mutex_init(&init_lock);
 	return &udpx_prov;
 }
--- a/src/common.c
+++ b/src/common.c
@@ -1942,7 +1942,8 @@ void ofi_free_list_of_addr(struct slist *addr_list)
 }
 
 static inline
-void ofi_insert_loopback_addr(const struct fi_provider *prov, struct slist *addr_list)
+void ofi_insert_loopback_addr(const struct fi_provider *prov,
+			      struct slist *addr_list, int mtu)
 {
 	struct ofi_addr_list_entry *addr_entry;
 
@@ -1953,6 +1954,7 @@ void ofi_insert_loopback_addr(const struct fi_provider *prov, struct slist *addr
 	addr_entry->comm_caps = FI_LOCAL_COMM;
 	addr_entry->ipaddr.sin.sin_family = AF_INET;
 	addr_entry->ipaddr.sin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+	addr_entry->mtu = mtu;
 	ofi_straddr_log(prov, FI_LOG_INFO, FI_LOG_CORE,
 			"available addr: ", &addr_entry->ipaddr);
 
@@ -1968,6 +1970,7 @@ void ofi_insert_loopback_addr(const struct fi_provider *prov, struct slist *addr
 	addr_entry->comm_caps = FI_LOCAL_COMM;
 	addr_entry->ipaddr.sin6.sin6_family = AF_INET6;
 	addr_entry->ipaddr.sin6.sin6_addr = in6addr_loopback;
+	addr_entry->mtu = mtu;
 	ofi_straddr_log(prov, FI_LOG_INFO, FI_LOG_CORE,
 			"available addr: ", &addr_entry->ipaddr);
 
@@ -2062,7 +2065,7 @@ void ofi_set_netmask_str(char *netstr, size_t len, struct ifaddrs *ifa)
 void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 			  struct slist *addr_list)
 {
-	int ret;
+	int ret, mtu = -1;
 	char *iface = NULL;
 	struct ofi_addr_list_entry *addr_entry;
 	struct ifaddrs *ifaddrs, *ifa;
@@ -2089,10 +2092,13 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 		if (ifa->ifa_addr == NULL ||
 			!(ifa->ifa_flags & IFF_UP) ||
 			!(ifa->ifa_flags & IFF_RUNNING) ||
-			(ifa->ifa_flags & IFF_LOOPBACK) ||
 			((ifa->ifa_addr->sa_family != AF_INET) &&
 			(ifa->ifa_addr->sa_family != AF_INET6)))
 			continue;
+		if (ifa->ifa_flags & IFF_LOOPBACK) {
+			mtu = ofi_ifaddr_get_mtu(ifa);
+			continue;
+		}
 		if (iface && strncmp(iface, ifa->ifa_name, strlen(iface) + 1)) {
 			FI_DBG(prov, FI_LOG_CORE,
 				"Skip (%s) interface\n", ifa->ifa_name);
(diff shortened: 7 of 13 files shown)


mpokorny:
I have a small command-line test application, but I'm not sure of the best way to modify it into a unit test and integrate it into the code base.
```c
#include <rdma/fabric.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

int main(int argc, char** argv) {
  int msg_size = 1000;
  if (argc > 1) {
    if (sscanf(argv[1], "%d", &msg_size) != 1 || msg_size <= 0) {
      printf("failed to convert argument '%s' to a positive integer\n", argv[1]);
      return 1;
    }
  }

  struct fi_fabric_attr* fabric_attr = calloc(1, sizeof(struct fi_fabric_attr));
  fabric_attr->prov_name = strdup("udp");

  struct fi_ep_attr* ep_attr = calloc(1, sizeof(struct fi_ep_attr));
  ep_attr->max_msg_size = msg_size;

  struct fi_info* hints = fi_allocinfo();
  hints->caps = FI_SEND | FI_RECV;
  hints->mode = FI_CONTEXT | FI_CONTEXT2;
  hints->addr_format = FI_SOCKADDR_IN;
  hints->ep_attr = ep_attr;
  hints->fabric_attr = fabric_attr;

  struct fi_info* info = NULL;
  struct fi_info* cur;
  fi_getinfo(
    FI_VERSION(FI_MAJOR_VERSION, FI_MINOR_VERSION),
    NULL,
    NULL,
    0,
    hints,
    &info);
  for (cur = info; cur; cur = cur->next)
    printf("domain %s, max_msg_size %d\n",
           cur->domain_attr->name,
           cur->ep_attr ? cur->ep_attr->max_msg_size : 0);
  fi_freeinfo(info);
  fi_freeinfo(hints);
  return 0;
}
```

prov/udp/src/udpx_init.c
```diff
@@ -37,20 +37,26 @@
...
 			uint64_t flags, const struct fi_info *hints,
 			struct fi_info **info)
 {
-	return ofi_ip_getinfo(&udpx_util_prov, version, node, service, flags,
-			      hints, info);
+	ofi_mutex_lock(&init_lock);
+	udpx_util_prov_init(version, node, service, flags);
+	ofi_mutex_unlock(&init_lock);
+	return util_getinfo(&udpx_util_prov, version, node, service, flags,
+			    hints, info);
```
j-xiong, line 50: udpx_util_prov is initialized once with the "node", "service" and "flags" from the first fi_getinfo() call.  That likely won't work properly for later fi_getinfo calls with different parameters.

src/linux/osd.c
```diff
@@ -257,3 +257,49 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+	char *line = NULL;
+	size_t len = 0;
+	char *mtu_filename_prefix = "/sys/class/net/";
+	char *mtu_filename_suffix = "/mtu";
+	char *mtu_filename;
+	int mtu;
+	/* IF_NAMESIZE includes NULL-terminated symbol */
+	size_t filename_len = strlen(mtu_filename_prefix) +
+				     strlen(mtu_filename_prefix) +
+				     IF_NAMESIZE;
```
j-xiong, line 273: Please align the 2nd and 3rd lines with the 'strlen' of the first line.

src/linux/osd.c
```diff
@@ -257,3 +257,49 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+	snprintf(mtu_filename, filename_len, "%s%s%s",
+		 mtu_filename_prefix, ifa->ifa_name, mtu_filename_suffix);
+
+	fd = fopen(mtu_filename, "r");
+	if (!fd)
+		goto err1;
+
+	if (getline(&line, &len, fd) == -1) {
+		goto err2;
+	}
```
j-xiong, line 288: Remove the braces here.

src/linux/osd.c
```diff
@@ -257,3 +257,49 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
 }
 
 #endif /* HAVE_ETHTOOL */
+
+int ofi_ifaddr_get_mtu(const struct ifaddrs *ifa)
+{
+	FILE *fd;
```
j-xiong, line 263: Use `fp` instead. `fd` is usually used for the integer file descriptor returned from `open`.

src/linux/osd.c
```diff
@@ -257,3 +257,49 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+
+	fd = fopen(mtu_filename, "r");
+	if (!fd)
+		goto err1;
+
+	if (getline(&line, &len, fd) == -1) {
+		goto err2;
+	}
+
+	if (sscanf(line, "%d", &mtu) != 1)
```
j-xiong, line 290: can use `fscanf` instead. that eliminates the need of calling `getline`.

src/linux/osd.c
```diff
@@ -257,3 +257,49 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+	size_t filename_len = strlen(mtu_filename_prefix) +
+				     strlen(mtu_filename_prefix) +
+				     IF_NAMESIZE;
+
+	mtu_filename = calloc(1, filename_len);
+	if (!mtu_filename)
+		return 0;
+
+	snprintf(mtu_filename, filename_len, "%s%s%s",
+		 mtu_filename_prefix, ifa->ifa_name, mtu_filename_suffix);
```
j-xiong, line 280: Can use `asprintf` instead. That way, you don't need to calculate the size of the string, and thus don't need to define the variable for the prefix and suffix.

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,39 @@ struct fi_info udpx_info = {
...
+
+static int detect_mtu(const struct fi_info* info) {
+
+	struct ifaddrs ifaddrs;
+	ifaddrs.ifa_next = NULL;
+	ifaddrs.ifa_flags = 0;
+	ifaddrs.ifa_netmask = NULL;
+	ifaddrs.ifa_name = info->domain_attr->name;
+	ifaddrs.ifa_addr = info->src_addr;
+	return ofi_ifaddr_get_mtu(&ifaddrs);
```
j-xiong, line 109: The window version of `ofi_ifaddr_get_mtu` returns `ifaddrs.mtu` which is undefined here.

prov/udp/src/udpx_init.c
mpokorny: Of course. Please check the updated code, which should address this.

prov/udp/src/udpx_attr.c
mpokorny: Please see the updated version.

src/linux/osd.c
mpokorny: Thanks for the suggestions. All applied in the update.

man/fi_udp.7.md
```diff
@@ -53,7 +52,10 @@ No support for counters.
 
 # RUNTIME PARAMETERS
 
-No runtime parameters are currently defined.
+The UDP provider checks for the following environment variables -
```
aingerson, line 55: Please move the man pages into a separate patch as they seem unrelated

prov/udp/src/udpx.h
```diff
@@ -88,6 +87,8 @@ struct udpx_ep_entry {
 	uint8_t			resv[sizeof(size_t) - 2];
 };
 
+#define UDPX_MAX_MSG_SIZE(mtu) ((mtu) - 28)
```
aingerson, line 90: What's the 28 here?

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,62 @@ struct fi_info udpx_info = {
...
+		}
+	} else {
+		FI_DBG(&udpx_prov, FI_LOG_CORE,
+		       "Failed to match interface (%s, %s) to "
+		       "address for MTU size\n",
+		       info->fabric_attr->name, info->domain_attr->name);
+	}
+}
+
+void udpx_util_prov_init(uint32_t version) {
```
aingerson, line 137: put the bracket on the next line

src/linux/osd.c
```diff
@@ -257,3 +257,31 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+
+int ofi_ifaddr_get_mtu(const struct ifaddrs *ifa)
+{
+	FILE *fp;
+        char *mtu_filename;
+        int mtu;
+
+	if (asprintf(&mtu_filename, "/sys/class/net/%s/mtu",
+		     ifa->ifa_name) == -1)
+	  return 0;
```
aingerson, line 269: check indent

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,62 @@ struct fi_info udpx_info = {
...
+static void set_mtu_from_addr_list(struct fi_info* info,
+				   struct slist *addr_list)
+{
+	struct ofi_addr_list_entry *addr_entry;
+	struct slist_entry *entry;
+	int max_msg_size;
+
+	entry = slist_find_first_match(addr_list, match_interface, info);
+	assert(entry);
+	if (entry) {
```
aingerson, line 120: I would keep either the assert and drop the else or remove the assert (I'd probably drop the assert). I could be convinced otherwise, though

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,62 @@ struct fi_info udpx_info = {
...
-	.flags = 0,
+	.info = NULL,
+        .flags = 0,
 };
+
+
+static int match_interface(struct slist_entry *entry, const void *infop)
+{
+	struct ofi_addr_list_entry *addr_entry;
+	const struct fi_info* info = infop;
```
aingerson, line 105: Add a line after the declarations for readability

aingerson (commented):
Just a few small things, nothing glaring. Thanks for the patch!
As far as your unit test, I would forgo putting it into our testing as it's not really testing anything but rather just printing out the values as far as I can tell.

man/fi_udp.7.md
mpokorny: Sure.

prov/udp/src/udpx.h
mpokorny: Combined size of UDP and IP headers. It's a step forward from the previous magic value of 1472 for the max message size, but I don't know of a convenient constant value or values I could otherwise use. I could make some local macro value definitions for clarity, I suppose.

prov/udp/src/udpx_attr.c
mpokorny: Dropped the assert.

prov/udp/src/udpx.h
aingerson: Where do you get that size from? Are you talking about the size of msghdr?

mpokorny: Well, I inferred that the value of `max_msg_size` needed to be smaller than the MTU because of the default, fixed value that `max_msg_size` had previously, namely 1472. Since 1472 is the maximum data size for a UDP packet in normally-sized Ethernet frames, I took a leap and figured that the relation between MTU size and `max_msg_size` was due to the UDP and IP headers. If there's more to it than that, I'd be happy to work out a better approach with some guidance.

aingerson: That makes sense, thanks! I couldn't find a better way to do it, you can just leave as is, but I do appreciate the local macro to make it clearer, thanks!

aingerson (dismissed):
Looks good to me, thanks!
You can include the man page updates in this PR btw, just have it in a separate patch. But you can also do it in a separate PR - up to you!
Our CI failed but it was unrelated to your changes. Rerunning for fun but don't consider it a blocker. Not sure what the AWS failure is though.

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+
+void udpx_util_prov_init(uint32_t version)
+{
+
+	struct slist addr_list;
+	struct fi_info* cur;
+	struct fi_info* info;
+
+        if (udpx_util_prov.info == NULL) {
+		udpx_util_prov.info = &udpx_info;
```
j-xiong, line 145: This line is redundant since it is overwritten by L153.

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+		info = fi_allocinfo();
+		ofi_ip_getinfo(&udpx_util_prov, version, NULL, NULL, 0, NULL,
+			       &info);
+		slist_init(&addr_list);
+		ofi_get_list_of_addr(&udpx_prov, "iface", &addr_list);
+		for (cur = info; cur; cur = cur->next)
+			set_mtu_from_addr_list(cur, &addr_list);
+		*(struct fi_info**)&udpx_util_prov.info = info;
```
j-xiong, line 153: Can be simplified to `udpx_util_prov.info = info;`

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+		ofi_free_list_of_addr(&addr_list);
+	}
+}
+
+void udpx_util_prov_fin()
```
j-xiong, line 158: rename to `udpx_util_prov_fini` to be consistent with other similar cases in the repo.

src/common.c
```diff
@@ -2182,7 +2190,7 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 	}
```
j-xiong, line 2190: `addr_entry->mtu` is unitialized in this case

mpokorny:
I pushed a commit to update the man page.

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+	struct ofi_addr_list_entry *addr_entry;
+	struct slist_entry *entry;
+	int max_msg_size;
+
+	entry = slist_find_first_match(addr_list, match_interface, info);
+	if (entry) {
+		addr_entry = container_of(entry,
+					  struct ofi_addr_list_entry,
+					  entry);
+		max_msg_size = UDPX_MAX_MSG_SIZE(addr_entry->mtu);
```
j-xiong, line 124: Need to handle the case that `addr_entry->mtu == -1`.

prov/udp/src/udpx_attr.c
mpokorny: But `udpx_util_prov` is used in the call to `ofi_ip_getinfo()`, which won't work correctly otherwise.

prov/udp/src/udpx_attr.c
mpokorny: The `info` field of `struct util_prov` is declared `const`, and the compiler complains without the cast.

prov/udp/src/udpx_attr.c
mpokorny: done with next commit

src/common.c
mpokorny: I'm fairly unfamiliar with Windows, which is where this code is used. Given what I see in `include/windows/config.h`, it's not clear to me that this code will actually be compiled, but I'm probably missing something. In any case, initializing `addr_entry->mtu` to -1 suffices to make the program correct, although the standard default MTU will then be assumed.

prov/udp/src/udpx_attr.c
mpokorny: That case is handled in line 125. (The value of UDPX_MAX_MSG_SIZE() is strictly less than its argument value.)

j-xiong (dismissed):

src/windows/osd.c
```diff
@@ -497,6 +498,11 @@ size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
 	return ifa->speed;
 }
 
+int ofi_ifaddr_get_mtu(const struct ifaddrs *ifa)
+{
+  return ifa->mtu;
```
shefty, line 503: Indentation is off.

include/freebsd/osd.h
```diff
@@ -76,6 +76,11 @@ static inline size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
 	return 0;
 }
 
+static inline int ofi_ifaddr_get_mtu(const struct ifaddrs *ifa)
+{
+	return -1;
```
shefty, line 81: Why not just return 1500 as the default?  Setting mtu = -1 as a default looks like a recipe for problems and missed error checking

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
 
 struct util_prov udpx_util_prov = {
 	.prov = &udpx_prov,
-	.info = &udpx_info,
-	.flags = 0,
+	.info = NULL,
+        .flags = 0,
```
shefty, line 98: replaced tab with spaces here - revert change

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+		info = fi_allocinfo();
+		ofi_ip_getinfo(&udpx_util_prov, version, NULL, NULL, 0, NULL,
+			       &info);
+		slist_init(&addr_list);
+		ofi_get_list_of_addr(&udpx_prov, "iface", &addr_list);
+		for (cur = info; cur; cur = cur->next)
+			set_mtu_from_addr_list(cur, &addr_list);
+		*(struct fi_info**)&udpx_util_prov.info = info;
+		ofi_free_list_of_addr(&addr_list);
+	}
```
shefty, line 155: Don't embed the entire function inside an if statement.  Revert the if check and return, then outdent the rest of the code.  Please add some spaces between lines.

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+void udpx_util_prov_init(uint32_t version)
+{
+
+	struct slist addr_list;
+	struct fi_info* cur;
+	struct fi_info* info;
+
+        if (udpx_util_prov.info == NULL) {
+		udpx_util_prov.info = &udpx_info;
+		info = fi_allocinfo();
```
shefty, line 146: Must check for failure

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+	entry = slist_find_first_match(addr_list, match_interface, info);
+	if (entry) {
+		addr_entry = container_of(entry,
+					  struct ofi_addr_list_entry,
+					  entry);
+		max_msg_size = UDPX_MAX_MSG_SIZE(addr_entry->mtu);
+		if (max_msg_size > 0) {
+			info->tx_attr->inject_size = max_msg_size;
+			info->ep_attr->max_msg_size = max_msg_size;
+		}
```
shefty, line 128: max_msg_size > 0 should be an assert.  We shouldn't leave these sizes unset.

src/common.c
```diff
@@ -2089,10 +2092,13 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 		if (ifa->ifa_addr == NULL ||
 			!(ifa->ifa_flags & IFF_UP) ||
 			!(ifa->ifa_flags & IFF_RUNNING) ||
-			(ifa->ifa_flags & IFF_LOOPBACK) ||
 			((ifa->ifa_addr->sa_family != AF_INET) &&
 			(ifa->ifa_addr->sa_family != AF_INET6)))
 			continue;
+		if (ifa->ifa_flags & IFF_LOOPBACK) {
+			mtu = ofi_ifaddr_get_mtu(ifa);
```
shefty, line 2099: rename to loopback_mtu to make it clear how this is used.  On first glance, this value looks like it's ignored.

src/common.c
```diff
@@ -2174,6 +2182,7 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 			addr_entry->ipaddr.sin.sin_family = AF_INET;
 			addr_entry->ipaddr.sin.sin_addr.s_addr =
 						iptbl->table[i].dwAddr;
+			addr_entry->mtu = -1;
```
shefty, line 2185: everywhere mtu = -1 should probably use a better default than an unusable value

include/freebsd/osd.h
mpokorny: Thanks for your feedback. One reason for using the value -1 is that it removes the need to put default MTU size values in a bunch more files. If there is there a good, single place to put such a definition, I must have missed it. I'm open to suggestions.

aingerson: I actually disagree with Sean for once. Returning -1 is similar to an enosys indicating that finding the MTU size is not supported/possible. Returning 1500 seems like it's falsely stating that it was able to determine that to be the MTU size. I think it's more appropriate for the default to be in the caller where it defaults to 1500 if were unable to find it.

shefty: Then you need to add checks everywhere that the mtu size is valid prior to using it.  There are several places where mtu == -1, but that value is used anyway, with, maybe, checks coming later that the calculation is valid.  If the mtu can't be determined, then remove that interface from the list.  I can't tell that these changes work at all on windows.

include/ofi_net.h
```diff
@@ -655,6 +655,7 @@ struct ofi_addr_list_entry {
 	char			ipstr[INET6_ADDRSTRLEN];
 	union ofi_sock_ip	ipaddr;
 	size_t			speed;
+	int			mtu;
```
shefty, line 658: Please separate out adding / setting the mtu to ofi_addr_list_entry into its own patch.  That will make it clearer which changes are generic, versus tied specifically to udp.

prov/udp/src/udpx_attr.c
shefty: The mtu valid check should be direct, not hidden behind a macro which uses an incorrect value that's then checked.  The code is assuming too much about an invalid mtu value and requires too much of the reader to figure out what's happening.

prov/udp/src/udpx_attr.c
```diff
@@ -93,6 +94,69 @@ struct fi_info udpx_info = {
...
+
+	struct slist addr_list;
+	struct fi_info* cur;
+	struct fi_info* info;
+
+        if (udpx_util_prov.info == NULL) {
+		udpx_util_prov.info = &udpx_info;
+		info = fi_allocinfo();
+		ofi_ip_getinfo(&udpx_util_prov, version, NULL, NULL, 0, NULL,
+			       &info);
```
shefty, line 148: I think the way to handle this is to add a callback to ofi_ip_getinfo().  That function is already iterating over the interfaces.  The only part that's missing is the ability of the provider to update the fi_info based on the specific interface.  For most callers, the callback will be null.  For UDP, the callback will receive the fi_info and ofi_addr_list_entry, so that it can update the attributes based on the mtu.  The callback may need to be passed through to subfunctions.

include/freebsd/osd.h
aingerson: Aaaand back to agreeing with Sean. Yes, proper error checking is absolutely necessary for that to work.

mpokorny: Wouldn't removing an interface whose mtu can't be determined potentially break existing client code? Doesn't the code in the main branch implicitly assume mtu=1500 for any interface supporting udp to determine the endpoint `max_msg_size` and `inject_size`? Falling back to that behavior seems like it might be safer (from that perspective) than removing the interface from the list. From a correctness point of view, dropping the interface seems preferable. For my use case, either choice is fine, so I could use some advice about how to proceed.

include/ofi_net.h
mpokorny: Makes sense. I'll do that with the next update.

prov/udp/src/udpx_attr.c
mpokorny: I will do that. It seems like a more extensive change to the code, but if you're looking over my shoulder, I'll give it a try.

shefty:
@mpokorny - There are other places where callbacks are used within the getinfo() path.  See ```struct util_prov::alter_defaults```.  I would see about adding a new callback to ```struct util_prov```, maybe alter_dev_defaults?  If that call is not null, call it for each fi_info created for each network interface.  The udp provider would just need to define the function and set it.

mpokorny:
I have a question about a possible alternative implementation to take. I noticed that `fi_link_attr` has an `mtu` field already, and was wondering whether that might be filled in by the udp provider for subsequent use by the `util_prov::alter_defaults` callback. That callback would then set the attribute values that depend on the MTU value (in `fi_tx_attr` and `fi_ep_attr`). Would such an implementation be correct, and, if so, how might it compare with what I've done so far?

shefty:
I'm not entirely certain, but reading the code, I don't think the fi_link_attr (via fi_info::nic) is being set for the tcp/udp providers.  If that were available, it would be great to make use of it.  Unfortunately, the current alter_defaults() callback occurs too early in the getinfo flow to be used.

ofi_ip_getinfo() -> util_getinfo()

This flow checks against the provider's template fi_info list.  For each fi_info that matches, the fi_info is dup'ed, then possibly altered by the provider.  This alter flow allows custom changes to the template fi_info based on the input hints.

The NIC interface list is checked after this flow (back in ofi_ip_getinfo()).  For every fi_info returned from the above flow, a new fi_info is created, one per interface.  (The exception is when a src_addr limits us to a single NIC.)  The places to add nic information would be adjacent to the calls to:

util_set_netif_names()

Sooo... here's what I would consider:

* Rename ofi_addr_list_entry to something like ofi_netif.  Rename related functions and variables.  This is just to clarify that it's capturing more than simple address information.
* Extend ofi_netif with mtu, plus additional device information which might be available.
* Update util_set_netif_names() -> util_set_netif().  In addition to changing the addressing, allocate and attach a fid_nic structure, plus sub-attributes, to the fi_info.  Note that it's okay if most of the data is left 0/null.  A future patch fill out more information later.  The fi_link_attr structure should be filled out and returned.
* The udp provider can iterate over the returned fi_info list from ofi_ip_getinfo() to update the msg_size based on the link mtu.

PR #10500, opened 2024-10-29
fabtests: Fix compiler warning about unitialized variable

j-xiong:
The compiler can't properly follow the def-use chain guarded by the same condition and complains about the variable could be used w/o initialization:

    if (atomic == FI_ATOMIC_COMPARE) {
        check_cmp =  ...;
    }
    ......
    if (atomic == FI_ATOMIC_COMPARE) {
        use check_cmp in macro
    }


aingerson (approved):
Silly compiler. Thanks!

PR #10502, opened 2024-10-29
hmem/cuda: Add env variable to enable/disable CUDA DMABUF

chuckfossen:
Add FI_HMEM_CUDA_USE_DMABUF environment variable and default to enabled.

There are applications that can run out of GPU BAR space when using dmabuf and gdrcopy both register memory.

### DIFF
--- a/src/hmem_cuda.c
+++ b/src/hmem_cuda.c
@@ -763,6 +763,9 @@ int cuda_hmem_init(void)
 			"If libfabric is not compiled with gdrcopy support, "
 			"this variable is not checked. (default: true)");
 
+	fi_param_define(NULL, "hmem_cuda_use_dmabuf", FI_PARAM_BOOL,
+			"Use dma-buf for sharing buffer with hardware. (default:true)");
+
 	ret = cuda_hmem_dl_init();
 	if (ret != FI_SUCCESS)
 		return ret;
@@ -936,7 +939,11 @@ bool cuda_is_gdrcopy_enabled(void)
 
 bool cuda_is_dmabuf_supported(void)
 {
-	return cuda_attr.dmabuf_supported;
+	int use_dmabuf = 1;
+
+	fi_param_get_bool(NULL, "hmem_cuda_use_dmabuf", &use_dmabuf);
+
+	return use_dmabuf && cuda_attr.dmabuf_supported;
 }
 
 #else


aws-nslick:
Can you provide more details of what's happening? I don't have a huge issue with a debug tunable to disable dmabuf, but I don't think it's reasonable to disable it by default, especially with nvidia now defaulting to the open driver variant so that dmabuf export is available in most cases.

aws-nslick:
See also: https://github.com/ofiwg/libfabric/pull/10445

shijin-aws:
I don't think we can make it false by default. EFA provider relies on this check to detect NIC-HBM p2p support status. And we likely will add this check when user call `fi_mr_regattr` with FI_MR_DMABUF. If we make this false by default, user need to export it to true manually to make it work, which is broken.

I think we have seen the similar issue you are concerned in https://github.com/ofiwg/libfabric/pull/10445/files, skipping gdrcopy for cuda dmabuf registration is a better mitigation.

iziemba:
@shijin-aws @aws-nslick : I agree. We will fix this PR to have it enabled by default.

I'll take a look at https://github.com/ofiwg/libfabric/pull/10445.

iziemba:
>The redundant mappings become costly for the driver to maintain.

The redundant mappings of the GPU bar is the issue we are seeing with DMA buf and GDR copy. Unfortunately, the solution in https://github.com/ofiwg/libfabric/pull/10445 rely doesn't apply because our apps requires the libfabric provider discover GPU buffer type (i.e. FI_MR_DMABUF and FI_MR_HMEM is not used). Our solution is to default to DMA buf being disabled with CUDA and users having to opt into DMA buf. We can make the change to disable DMA buf for CUDA by default in CXI provider.

shijin-aws:
@iziemba I see you have https://github.com/ofiwg/libfabric/pull/10502 to disable dmabuf for CXI only. Any reason you want to make it generalized now?

iziemba:
Our rational for generalizing it now is that ROCR already has FI_HMEM_ROCR_USE_DMABUF. For simplicity, rather than have CXI provider specific env vars, it seemed best to define a matching FI_HMEM_CUDA_USE_DMABUF.

CXI provider memory registration does the following:
# Calls ofi_get_hmem_iface() to determine buffer type
# Calls ofi_hmem_get_dmabuf_fd() to determine if DMA buf can be used
# Calls into kernel do do memory registration
# Calls ofi_hmem_dev_register() to enable load/store access

aws-nslick:
> Unfortunately, the solution in https://github.com/ofiwg/libfabric/pull/10445 rely doesn't apply because our apps requires the libfabric provider discover GPU buffer type (i.e. FI_MR_DMABUF and FI_MR_HMEM is not used).

The specifics aren't super clear to me, so this might not be viable, but I think you can use [`cuMemRetainAllocationHandle`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html#group__CUDA__VA_1g1ddca5437c502782155f95bf98e775c6) to solve this. It should be possible to pass the resulting allocation handle to [`cuMemGetAllocationPropertiesFromHandle`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html#group__CUDA__VA_1gc1c4c812caba5a21401c2cb4ab4512b1) and then check whether the posix fd requirements are met. If they are, then you can get a reference to the dmabuf and move forward, otherwise, go with gdr.

shijin-aws:
>Calls ofi_hmem_get_dmabuf_fd() to determine if DMA buf can be used
Calls into kernel to do memory registration
Calls ofi_hmem_dev_register() to enable load/store access

Then you already know it's a cuda buffer and whether you have done dmabuf reg before you are doing dev register, right? Then https://github.com/ofiwg/libfabric/pull/10445 can be slightly modified to skip dev reg when you have done dmabuf reg...But it's totally up to you. We choose to skip gdrcopy in this case as it has less value against making dmabuf on.

j-xiong (approved):
The updated PR looks fine to me.

iziemba:
>Then you already know it's a cuda buffer and whether you have done dmabuf reg before you are doing dev register, right? Then https://github.com/ofiwg/libfabric/pull/10445 can be slightly modified to skip dev reg when you have done dmabuf reg...But it's totally up to you. We choose to skip gdrcopy in this case as it has less value against keeping dmabuf on.

We have a functional requirement requiring both GDR copy and memory registration to work at the same time. The functional requirement driven by the CXI provider may resort to memcpy in the data path and CUDA memcpy APIs cannot be relied on in all causes due to deadlock. If memory registration is being done via DMA buf, we still need GDR copy for load/store support to avoid CUDA APIs..

chuckfossen:
The continuous-integration/intel/jenkins build is failing. The Details link doesn't seem to work for me.

shijin-aws:
>We have a functional requirement requiring both GDR copy and memory registration to work at the same time. The functional requirement driven by the CXI provider may resort to memcpy in the data path and CUDA memcpy APIs cannot be relied on in all causes due to deadlock. If memory registration is being done via DMA buf, we still need GDR copy for load/store support to avoid CUDA APIs..

Fair enough, we know that deadlock issue for a long time that you cannot use cuda API within NCCL for memcpy. The workaround is to use gdrcopy or using NIC local rdma read. We now prefer the latter one.

shijin-aws (approved):

PR #10508, opened 2024-10-31
contrib/intel/jenkins: Separate tcp & tcp;ofi_rxm testing

zachdworkin:
separate tcp and tcp;ofi_rxm testing so that both providers are covered. Previously tcp was testing everything and picking up rxm as needed. Instead it is better practice to separate them and just test both usecases entirely. Update both tcp ubertest lists to correctly run only supported tests.


zachdworkin:
I will rebase this to pick up #10510's change to the ofi_rxm ubertest config file

PR #10509, opened 2024-10-31
prov/efa: Initialize efa fork support in EFA_INI

jiaxiyan:
Move efa_fork_support_enable_if_requested before initializing efa_hmem_info in EFA_INI because ibv_fork_init needs to be called before ibv_reg_mr. Fix the wrong return value of efa_fork_support_is_enabled.
Also update efa_hmem log from FI_LOG_DOMAIN to FI_LOG_CORE.


shijin-aws (approved):

shijin-aws:
Thanks!

shijin-aws:
I guess we can catch this error earlier by running the fork fabtests on g4dn in PR test.

prov/efa/src/efa_user_info.c
```diff
@@ -610,6 +610,16 @@ int efa_getinfo(uint32_t version, const char *node,
 	struct fi_info *dgram_info_list, *rdm_info_list;
 	int err;
 
+#ifndef _WIN32
+	err = efa_fork_support_install_fork_handler();
```
shijin-aws, line 614: Nit: since we are not installing it during provider init, let's add back that TODO here like

```
TODO:
It'd be better to install this during provider init (since that's
only invoked once) but fork() is currently called by nvml_init in hmem_hook provider's init
which is called after efa provider init. This can move to the provider init after we get rid of
that fork() in other provider's init.
```

shijin-aws (approved):

shijin-aws (approved):

PR #10514, opened 2024-11-01
prov/hook: Fix the preprocessor

shijin-aws:
ifdef should be if here, as HAVE_HOOK_HMEM_DL is defined anyway.

### DIFF
--- a/prov/hook/hook_hmem/src/hook_hmem.c
+++ b/prov/hook/hook_hmem/src/hook_hmem.c
@@ -1914,7 +1914,7 @@ out:
 
 HOOK_HMEM_INI
 {
-#ifdef HAVE_HOOK_HMEM_DL
+#if HAVE_HOOK_HMEM_DL
 	ofi_hmem_init();
 #endif
 	hook_hmem_fabric_ops = hook_fabric_ops;


j-xiong:
@shijin-aws `fabric.c` is not part of DL providers. The call of `ofi_hmem_init` from `fi_ini` only applies to the libfabric core and the DL providers has to make their own calls.

j-xiong (approved):

shijin-aws:
Then if I build like efa provider as dl, I see `ofi_hmem_init` will be called twice, it will get worse for multiple dl providers, as some hmem init like cuda is not trivial. I thought `hmem_ops[iface].initialized` should make them finally called once, but this doesn't happen.

j-xiong:
The DL provider and the libfabric core have different name spaces. The DL provider has its own `hmem_ops` structure and that needs to be initialized by the DL provider itself.

shijin-aws:
Yeah, one thing bothers me is that `cuda_hmem_init` will call `nvml_init` which will call `fork()`  implicitly. If I want to register a fork handler (to catch application's fork behavior) inside efa provider's prov_ini it will catch that fork from other dl provider's init. I either need to make that fork handler installed later, or get rid of that nvml_init call in the future.

PR #10517, opened 2024-11-05
core: Remove redundant windows.h

weypro:
The windows.h header is not necessary for ssize_t definition on Windows platform since BaseTsd.h already provides the required SSIZE_T type definition. Remove this redundant include to reduce compilation dependencies.

### DIFF
--- a/include/rdma/fabric.h
+++ b/include/rdma/fabric.h
@@ -65,7 +65,6 @@
 
 #if defined(_WIN32)
 #include <BaseTsd.h>
-#include <windows.h>
 typedef SSIZE_T ssize_t;
 #endif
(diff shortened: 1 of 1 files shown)


aingerson:
@weypro Seems like a fine change, but you'll need to add a signoff line to your commit to let the DCO check pass. I'm guessing the AWS failure is unrelated but it will get re-triggered when you add the signoff and repush. Thanks!

j-xiong:
Also please add the text of the PR description to the commit message. That help people to understand the rationale of the change when looking at the log.

shijin-aws:
bot:aws:retest

PR #10520, opened 2024-11-06
prov/efa: Fall back to zero sl when non-zero sl qp creation failed

shijin-aws:
When driver doesn't support non-zero sl, the qp creation will fail. In this case, fallback to zero sl qp.

### DIFF
--- a/prov/efa/src/efa_base_ep.c
+++ b/prov/efa/src/efa_base_ep.c
@@ -186,13 +186,24 @@ int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex, 
 #endif
 		efa_attr.driver_qp_type = EFADV_QP_DRIVER_TYPE_SRD;
 #if HAVE_EFADV_SL
+		efa_attr.sl = EFA_QP_DEFAULT_SERVICE_LEVEL;
 		if (tclass == FI_TC_LOW_LATENCY)
-			efa_attr.sl = EFA_QP_DEFAULT_SERVICE_LEVEL;
+			efa_attr.sl = EFA_QP_LOW_LATENCY_SERVICE_LEVEL;
 #endif
 		(*qp)->ibv_qp = efadv_create_qp_ex(
 			init_attr_ex->pd->context, init_attr_ex, &efa_attr,
 			sizeof(struct efadv_qp_init_attr));
 	}
+
+#if HAVE_EFADV_SL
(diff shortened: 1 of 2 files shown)


shijin-aws:
@mrgolin can you take a look

prov/efa/src/efa_base_ep.c
```diff
@@ -186,12 +186,29 @@ int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex,
...
+			}
+		} else {
+			(*qp)->ibv_qp = efadv_create_qp_ex(
+				init_attr_ex->pd->context, init_attr_ex, &efa_attr,
+				sizeof(struct efadv_qp_init_attr));
+		}
+#else
 		(*qp)->ibv_qp = efadv_create_qp_ex(
 			init_attr_ex->pd->context, init_attr_ex, &efa_attr,
 			sizeof(struct efadv_qp_init_attr));
```
darrylabbate, line 210: Can we refactor this s.t. we're not repeating the same thing 4 times here (`(*qp)->ibv_qp = efadv_create_qp_ex(...)`)?

shijin-aws: I thought about it, but the current version is the what I think the most intuitive one. Any suggestions?

shijin-aws: We may move the retry logic to https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_base_ep.c#L197. It can reduce the number of `efadv_create_qp_ex` calls

shijin-aws: But it will involve more `HAVE_EFADV_SL` macros ... which I don''t think makes the code cleaner.

jiaxiyan: Like
```
#if HAVE_EFADV_SL
	efa_attr.sl = (tclass == FI_TC_LOW_LATENCY) 
		? EFA_QP_LOW_LATENCY_SERVICE_LEVEL 
		: EFA_QP_DEFAULT_SERVICE_LEVEL;
#endif
(*qp)->ibv_qp = efadv_create_qp_ex()

if (!(*qp)->ibv_qp) {
#if HAVE_EFADV_SL
        if (tclass == FI_TC_LOW_LATENCY) {
                // retry
        }
#endif
...
}
```

shijin-aws: Updated

jiaxiyan (approved):

PR #10527, opened 2024-11-07
prov/efa: Implement efa_msg interface

jiaxiyan:
Rename efa_dgram_msg.c to efa_msg.c as a common MSG interface for both rdm and dgram ep type.
Add support for FI_INJECT and FI_REMOTE_CQ_DATA.
Replace ibv_post_send with ibv_wr* calls and remove xmit_more_wr_head and xmit_more_wr_tail.
Replace recv_more_wr_head and recv_more_wr_tail with efa_recv_wr_vec.


prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+{
+	struct efa_base_ep *base_ep = container_of(ep_fid, struct efa_base_ep, util_ep.ep_fid);
+	struct fi_msg msg;
+
+	EFA_SETUP_MSG(msg, iov, desc, count, src_addr, context, 0);
+
+	return efa_post_recv(base_ep, &msg, 0);
+}
+
+static inline ssize_t efa_post_send_validate(struct efa_base_ep *base_ep, const struct fi_msg *msg,
```
shijin-aws, line 179: I prefer to make all of these validation as assertion so we don't have to check them explicitly for non-debug build

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+
+	return 0;
+}
+
+static inline bool efa_msg_has_hmem_mr(const struct fi_msg *msg)
+{
+	/* the device only support send up 2 iov, so iov_count cannot be > 2 */
+	assert(msg->iov_count == 1 || msg->iov_count == 2);
+	/* first iov is always on host memory, because it must contain packet header */
+	assert(!efa_mr_is_hmem(msg->desc[0]));
```
shijin-aws, line 210: This was wrong, we should shift the iov by prefix size and only check the payload

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+			     &iov_offset);
+	if (OFI_UNLIKELY(ret)) {
+		EFA_WARN(FI_LOG_CQ, "ofi_iov_locate failed! err: %d\n", ret);
+		return ret;
+	}
+	assert(iov_idx < msg->iov_count);
+	assert(iov_offset < msg->msg_iov[iov_idx].iov_len);
+
+	if (len <= base_ep->domain->device->efa_attr.inline_buf_size &&
+	    !efa_msg_has_hmem_mr(msg)) {
```
shijin-aws, line 255: use `efa_mr_is_hmem` with argument as the msg's desc

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+		for (i = 0; i < msg->iov_count; i++) {
+			inline_data_list[i].addr = msg->msg_iov[i].iov_base;
+			inline_data_list[i].length = msg->msg_iov[i].iov_len;
+		}
+		inline_data_list[iov_idx].addr = (char *) msg->msg_iov[iov_idx].iov_base + iov_offset;
```
shijin-aws, line 260: Is this correct? Should the loop start from `iov_index`? what about the case where iov_count is 2 but the first iov is all prefix? Then should you drop the first iov entirely?

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+		inline_data_list[iov_idx].length -= iov_offset;
+
+		ibv_wr_set_inline_data_list(qp->ibv_qp_ex, msg->iov_count, inline_data_list);
+	} else {
+		for (i = 0; i < msg->iov_count; i++) {
+			sg_list[i].addr = (uintptr_t)msg->msg_iov[i].iov_base;
```
shijin-aws, line 266: Similar comment as earlier

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	qp->ibv_qp_ex->wr_id = (uintptr_t)msg->context;
+	if (flags & FI_REMOTE_CQ_DATA) {
+		ibv_wr_send_imm(qp->ibv_qp_ex, msg->data);
+	} else {
+		ibv_wr_send(qp->ibv_qp_ex);
+	}
+
+	ret = ofi_iov_locate((struct iovec *) msg->msg_iov, msg->iov_count,
```
shijin-aws, line 244: Thinking about it again, I think it's easier to use 

```
void ofi_consume_iov_desc(struct iovec *iov, void **desc,
			  size_t *iov_count, size_t to_consume)
```
to walk the iov, so you will start from the payload as index 0 anyway and simplify your code

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+	if (OFI_UNLIKELY(err))
+		goto out_err;
+
+	wr = &base_ep->efa_recv_wr_vec[wr_index].wr;
+	wr->wr_id = (uintptr_t)msg->context;
+	wr->num_sge = msg->iov_count;
+	wr->sg_list = base_ep->efa_recv_wr_vec[wr_index].sge;
+
+	for (i = 0; i < msg->iov_count; i++) {
+		addr = (uintptr_t)msg->msg_iov[i].iov_base;
```
shijin-aws, line 103: similar comments for the procedure to walk iov by prefix

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,408 @@
...
+#define EFA_SETUP_IOV(iov, buf, len)           \
+	do {                                   \
+		iov.iov_base = (void *)buf;    \
+		iov.iov_len = (size_t)len;     \
+	} while (0)
+
+#define EFA_SETUP_MSG(msg, iov, _desc, count, _addr, _context, _data)    \
+	do {                                                             \
+		msg.msg_iov = (const struct iovec *)iov;                 \
+		msg.desc = (void **)_desc;                               \
```
shijin-aws, line 27: So this is not a copy of the user request (iov, desc), then and then if we use `ofi_consume_iov` , we will modify user's iov and desc, which is wrong.

So we either need to do a memcpy here, or keep using `ofi_iov_locate`

prov/efa/src/efa_msg.c
shijin-aws: I realize EFA_SETUP_MSG doesn't copy the iov and desc from user request, so we cannot use `ofi_consume_iov_desc` unless we do a copy there.

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,326 @@
...
+
+	qp->ibv_qp_ex->wr_id = (uintptr_t)msg->context;
+	if (flags & FI_REMOTE_CQ_DATA) {
+		ibv_wr_send_imm(qp->ibv_qp_ex, msg->data);
+	} else {
+		ibv_wr_send(qp->ibv_qp_ex);
+	}
+
+	if (len <= base_ep->domain->device->efa_attr.inline_buf_size &&
+	    !efa_mr_is_hmem(msg->desc[0])) {
```
shijin-aws, line 179: There is a corner case here, is that `desc` is NULL, which will be regarded as host memory as well

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,326 @@
...
+	len = ofi_total_iov_len(msg->msg_iov, msg->iov_count);
+	assert(len <= base_ep->info->ep_attr->max_msg_size);
+
+	if (!base_ep->is_wr_started) {
+		ibv_wr_start(qp->ibv_qp_ex);
+		base_ep->is_wr_started = true;
+	}
+
+	qp->ibv_qp_ex->wr_id = (uintptr_t)msg->context;
+	if (flags & FI_REMOTE_CQ_DATA) {
```
shijin-aws, line 172: Since we are here, I think we should update this https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_prov_info.c#L78 to the efa device imm data size, which is 4

https://github.com/ofiwg/libfabric/commit/ead305192bd78d2395c64b99b02f79b2f16be8b2

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,326 @@
...
+		ibv_wr_send_imm(qp->ibv_qp_ex, msg->data);
+	} else {
+		ibv_wr_send(qp->ibv_qp_ex);
+	}
+
+	if (len <= base_ep->domain->device->efa_attr.inline_buf_size &&
+	    msg->desc && msg->desc[0] && !efa_mr_is_hmem(msg->desc[0])) {
```
shijin-aws, line 179: This still doesn't look right... we want to do inline when it's host memory, should it be 

```
	if (len <= base_ep->domain->device->efa_attr.inline_buf_size &&
	    (!msg->desc || !efa_mr_is_hmem(msg->desc[0]))) {
```

`efa_mr_is_hmem` will return false already when `msg->desc[0]` is NULL

prov/efa/src/efa_prov_info.c
```diff
@@ -75,7 +75,7 @@ const struct fi_domain_attr efa_domain_attr = {
 	.resource_mgmt		= FI_RM_DISABLED,
 	.mr_mode		= OFI_MR_BASIC_MAP | FI_MR_LOCAL | OFI_MR_BASIC,
 	.mr_key_size		= sizeof_field(struct ibv_sge, lkey),
-	.cq_data_size		= 0,
+	.cq_data_size		= 4,
```
shijin-aws, line 78: This should be a general macro as `EFA_CQ_DATA_SIZE` (, and we should remove this https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_prov_info.c#L557-L559 and https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_cq.h#L18-L21 after having it by default

prov/efa/src/efa_msg.c
jiaxiyan: I misread your earlier comment. So we do inline send when `msg->desc` or `msg->desc[0]` is NULL too

shijin-aws: Correct

shijin-aws:
Do you need to patch efa/pytest to make dgram run without prefix (`-k`) ?

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,325 @@
...
+		efa_mr = (struct efa_mr *)msg->desc[i];
+		wr->sg_list[i].lkey = efa_mr->ibv_mr->lkey;
+		wr->sg_list[i].addr = addr;
+	}
+
+	base_ep->efa_recv_wr_vec[wr_index].wr.next = NULL;
+	if (wr_index > 0)
+		base_ep->efa_recv_wr_vec[wr_index - 1].wr.next = &base_ep->efa_recv_wr_vec[wr_index].wr;
+
+	wr_index++;
```
shijin-aws, line 96: Do you need this increment?

shijin-aws: It seems `wr_index` is not used in the subsequent lines

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,325 @@
...
+	wr->wr_id = (uintptr_t)msg->context;
+	wr->num_sge = msg->iov_count;
+	wr->sg_list = base_ep->efa_recv_wr_vec[wr_index].sge;
+
+	for (i = 0; i < msg->iov_count; i++) {
+		addr = (uintptr_t)msg->msg_iov[i].iov_base;
+
+		/* Set RX buffer desc from SGE */
+		wr->sg_list[i].length = msg->msg_iov[i].iov_len;
+		assert(msg->desc[i]);
```
shijin-aws, line 86: better to have `assert(msg->desc && msg->desc[i])` here as well

jiaxiyan:
I will rebase after https://github.com/ofiwg/libfabric/pull/10532 is merged because it also adds efa_recv_wr_index

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,369 @@
...
+
+	if (len <= base_ep->domain->device->efa_attr.inline_buf_size &&
+	    (!msg->desc || !efa_mr_is_hmem(msg->desc[0]))) {
+		for (i = 0; i < msg->iov_count; i++) {
+			inline_data_list[i].addr = msg->msg_iov[i].iov_base;
+			inline_data_list[i].length = msg->msg_iov[i].iov_len;
+
+			/* Whole prefix must be on the first sgl for dgram */
+			if (!i && qp->ibv_qp->qp_type == IBV_QPT_UD) {
+				inline_data_list[i].addr = (char*)inline_data_list[i].addr + base_ep->info->ep_attr->msg_prefix_size;
```
shijin-aws, line 217: I remember we used to  skip the first 1 iov if its length is the same with the prefix size...But IIRC having a 0 length sg_list[0] should be fine as driver will skip them

shijin-aws:
bot:aws:retest

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,369 @@
...
+static inline ssize_t efa_post_recv(struct efa_base_ep *base_ep, const struct fi_msg *msg, uint64_t flags)
+{
+	struct efa_mr *efa_mr;
+	struct efa_qp *qp = base_ep->qp;
+	struct ibv_recv_wr *bad_wr;
+	struct ibv_recv_wr *wr;
+	uintptr_t addr;
+	ssize_t err, post_recv_err;
+	size_t i, wr_index = base_ep->efa_recv_wr_index;
+
```
shijin-aws, line 71: do we need a check here, if `recv_wr_index` exceeds the rx limit, we should return EAGAIN

shijin-aws: my concern is that `efa_recv_wr_vec` is a fixed array of max_rx_size, it will cause overflow if we exceed the limit.

prov/efa/src/efa_msg.c
```diff
@@ -0,0 +1,378 @@
...
+	struct efa_mr *efa_mr;
+	struct efa_qp *qp = base_ep->qp;
+	struct ibv_recv_wr *bad_wr;
+	struct ibv_recv_wr *wr;
+	uintptr_t addr;
+	ssize_t err, post_recv_err;
+	size_t i, wr_index = base_ep->recv_wr_index;
+
+	if (wr_index >= base_ep->info->rx_attr->size) {
+		EFA_WARN(FI_LOG_EP_DATA,
```
shijin-aws, line 73: Make the log in info level. Also you can just `return -FI_EAGAIN` without assigning it to `err`.

shijin-aws: Correct my last comment. I think in this case we need to flush the recv, we should goto `out_err`

PR #10534, opened 2024-11-13
prov/tcp: introduce TCP_NO_CONNECT flag

ooststep:
There are some specific use cases where we may not want one side of communication to initiate connections, namely when we know that one side of our configuration is being heavily restricted by a firewall.  To prevent indefinite hangs with certain operations, such as RMA reads and writes, introduce a provider specific flag to trigger an error if there is not already an established connection.  In this case, the application can force the connection from the other direction.

### DIFF
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -78,6 +78,11 @@ enum {
 	FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 };
 
+/* provider specific op flags range between 60-63 */
+enum {
+	FI_TCP_NO_CONNECT = (1ULL << 60)
+};
+
 struct fi_fid_export {
 	struct fid **fid;
 	uint64_t flags;
--- a/man/fi_tcp.7.md
+++ b/man/fi_tcp.7.md
@@ -32,6 +32,15 @@ The following features are supported
 *Shared Rx Context*
 : The tcp provider supports shared receive context
 
+# PROVIDER SPECIFIC OPERATION FLAGS
+: The tcp provider supports the following op flags
+
+*FI_TCP_NO_CONNECT*
+: This flag indicates that operations should fail if there is no
+  existing connection to the remote peer.  In such case, an FI_ENOTCONN
+  error should be expected.
+
+
 # RUNTIME PARAMETERS
 
 The tcp provider may be configured using several environment variables.  A
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -98,6 +98,9 @@ typedef void  xnet_profile_t;
 #define XNET_MIN_MULTI_RECV	16384
 #define XNET_PORT_MAX_RANGE	(USHRT_MAX)
 
+/* provider specific op flags */
+#define TCP_NO_CONNECT FI_TCP_NO_CONNECT
+
 extern struct fi_provider	xnet_prov;
 extern struct util_prov		xnet_util_prov;
 void xnet_init_infos(void);
@@ -304,8 +307,8 @@ struct xnet_rdm {
 
 int xnet_rdm_ep(struct fid_domain *domain, struct fi_info *info,
 		struct fid_ep **ep_fid, void *context);
-ssize_t xnet_get_conn(struct xnet_rdm *rdm, fi_addr_t dest_addr,
-		      struct xnet_conn **conn);
+ssize_t xnet_get_conn(struct xnet_rdm *rdm, fi_addr_t addr,
+		      struct xnet_conn **conn, uint64_t flags);
 struct xnet_ep *xnet_get_rx_ep(struct xnet_rdm *rdm, fi_addr_t addr);
 void xnet_freeall_conns(struct xnet_rdm *rdm);
 
--- a/prov/tcp/src/xnet_rdm_cm.c
+++ b/prov/tcp/src/xnet_rdm_cm.c
@@ -317,28 +317,31 @@ xnet_alloc_conn(struct xnet_rdm *rdm, struct util_peer_addr *peer)
 	return conn;
 }
 
-static struct xnet_conn *
-xnet_add_conn(struct xnet_rdm *rdm, struct util_peer_addr *peer)
+static ssize_t
+xnet_add_conn(struct xnet_rdm *rdm, struct util_peer_addr *peer,
+	      struct xnet_conn **conn, uint64_t flags)
 {
-	struct xnet_conn *conn;
 
 	assert(xnet_progress_locked(xnet_rdm2_progress(rdm)));
-	conn = ofi_idm_lookup(&rdm->conn_idx_map, peer->index);
-	if (conn)
-		return conn;
+	*conn = ofi_idm_lookup(&rdm->conn_idx_map, peer->index);
+	if (*conn)
+		return 0;
 
-	conn = xnet_alloc_conn(rdm, peer);
-	if (!conn)
-		return NULL;
+	if (flags & TCP_NO_CONNECT)
+		return -FI_ENOTCONN;
 
-	if (ofi_idm_set(&rdm->conn_idx_map, peer->index, conn) < 0) {
-		xnet_free_conn(conn);
+	*conn = xnet_alloc_conn(rdm, peer);
+	if (!(*conn))
+		return -FI_ENOMEM;
+
+	if (ofi_idm_set(&rdm->conn_idx_map, peer->index, *conn) < 0) {
+		xnet_free_conn(*conn);
 		XNET_WARN_ERR(FI_LOG_EP_CTRL, "ofi_idm_set", -FI_ENOMEM);
-		return NULL;
+		return -FI_ENOMEM;
 	}
 
-	conn->flags |= XNET_CONN_INDEXED;
-	return conn;
+	(*conn)->flags |= XNET_CONN_INDEXED;
+	return 0;
 }
 
 /* The returned conn is only valid if the function returns success.
@@ -346,16 +349,16 @@ xnet_add_conn(struct xnet_rdm *rdm, struct util_peer_addr *peer)
  * we return that rather than int.
  */
 ssize_t xnet_get_conn(struct xnet_rdm *rdm, fi_addr_t addr,
-		      struct xnet_conn **conn)
+		      struct xnet_conn **conn, uint64_t flags)
 {
 	struct util_peer_addr **peer;
 	ssize_t ret;
 
 	assert(xnet_progress_locked(xnet_rdm2_progress(rdm)));
 	peer = ofi_av_addr_context(rdm->util_ep.av, addr);
-	*conn = xnet_add_conn(rdm, *peer);
-	if (!*conn)
-		return -FI_ENOMEM;
+	ret = xnet_add_conn(rdm, *peer, conn, flags);
+	if (ret)
+		return ret;
 
 	if (!(*conn)->ep) {
 		ret = xnet_rdm_connect(*conn);
@@ -444,8 +447,8 @@ static void xnet_process_connreq(struct fi_eq_cm_entry *cm_entry)
 		goto reject;
 	}
 
-	conn = xnet_add_conn(rdm, peer);
-	if (!conn)
+	ret = xnet_add_conn(rdm, peer, &conn, 0);
+	if (ret)
 		goto put;
 
 	FI_INFO(&xnet_prov, FI_LOG_EP_CTRL, "connreq for %p\n", conn);
--- a/prov/tcp/src/xnet_rdm.c
+++ b/prov/tcp/src/xnet_rdm.c
@@ -77,7 +77,7 @@ xnet_rdm_send(struct fid_ep *ep_fid, const void *buf, size_t len,
 
 	rdm = container_of(ep_fid, struct xnet_rdm, util_ep.ep_fid);
 	ofi_genlock_lock(&xnet_rdm2_progress(rdm)->rdm_lock);
-	ret = xnet_get_conn(rdm, dest_addr, &conn);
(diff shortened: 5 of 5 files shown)


prov/tcp/src/xnet_rdm_cm.c
```diff
@@ -317,45 +317,48 @@ xnet_alloc_conn(struct xnet_rdm *rdm, struct util_peer_addr *peer)
...
 	struct util_peer_addr **peer;
 	ssize_t ret;
 
 	assert(xnet_progress_locked(xnet_rdm2_progress(rdm)));
 	peer = ofi_av_addr_context(rdm->util_ep.av, addr);
-	*conn = xnet_add_conn(rdm, *peer);
-	if (!*conn)
-		return -FI_ENOMEM;
+	ret = xnet_add_conn(rdm, *peer, conn, flags);
+	if (!ret)
```
aingerson, line 360: Should this be `if (!ret || ret == -FI_ENOTCONN)` ?

prov/tcp/src/xnet.h
```diff
@@ -98,6 +98,9 @@ typedef void  xnet_profile_t;
 #define XNET_MIN_MULTI_RECV	16384
 #define XNET_PORT_MAX_RANGE	(USHRT_MAX)
 
+/* operation flags */
+#define TCP_NO_CONNECT  (1ULL << 60)
```
soumagne, line 102: I was actually expecting a public FI_ flag similar to FI_COMPLETION etc. How does this work currently ? is it just always enabled ? We need this flag not to be set on standard fi_send/fi_recv but only on fi_read / fi_write.

prov/tcp/src/xnet_attr.c
```diff
@@ -50,9 +50,10 @@
 
 #define XNET_TX_OP_FLAGS \
 	(FI_INJECT | FI_INJECT_COMPLETE | FI_TRANSMIT_COMPLETE | \
-	 FI_DELIVERY_COMPLETE | FI_COMMIT_COMPLETE | FI_COMPLETION)
+	 FI_DELIVERY_COMPLETE | FI_COMMIT_COMPLETE | FI_COMPLETION | \
+	 TCP_NO_CONNECT)
 
-#define XNET_RX_OP_FLAGS (FI_COMPLETION)
+#define XNET_RX_OP_FLAGS (FI_COMPLETION | TCP_NO_CONNECT)
```
aingerson, line 56: It doesn't look like there's any connection look up or handling of this flag on the rx side. I think this is probably just a tx flag

prov/tcp/src/xnet.h
ooststep: It's not a feature that will be supported beyond TCP provider so I don't think it makes much sense as a core feature.

As implemented, it can apply to the entire ep via default op_flag or to individual operations that support flags, such as fi_readmsg/fi_writemsg (and don't set as an op_flag)

aingerson (commented):
Add to the man page as well

prov/tcp/src/xnet_rdm_cm.c
ooststep: oops. yep!

prov/tcp/src/xnet.h
soumagne: I don't understand, this is a private provider flag. How can an app be using that flag without any header defining its value ?

ooststep: You're right - I misunderstood your original point. this needs to move.

prov/tcp/src/xnet_attr.c
```diff
@@ -50,9 +50,10 @@
 
 #define XNET_TX_OP_FLAGS \
 	(FI_INJECT | FI_INJECT_COMPLETE | FI_TRANSMIT_COMPLETE | \
-	 FI_DELIVERY_COMPLETE | FI_COMMIT_COMPLETE | FI_COMPLETION)
+	 FI_DELIVERY_COMPLETE | FI_COMMIT_COMPLETE | FI_COMPLETION | \
+	 TCP_NO_CONNECT)
```
soumagne, line 54: Also unless I'm misunderstanding the code here, I don't think we want this to be in the default flags ? The default should be the current behavior, which is to attempt connection?

man/fi_tcp.7.md
```diff
@@ -32,6 +32,16 @@ The following features are supported
 *Shared Rx Context*
 : The tcp provider supports shared receive context
 
+# PROVIDER SPECIFIC OPERATION FLAGS
+: The tcp provider supports the following op flags
+
+*FI_FLAG_TCP_NO_CONNECT*
+: This flag indicates that operations should fail if there is not an
```
j-xiong, line 39: `not an` --> `no`

man/fi_tcp.7.md
```diff
@@ -32,6 +32,16 @@ The following features are supported
...
+  existing connection to the remote peer.  Instead, if there is not
+  an established connection, an FI_ENOTCONN error should be expected.
```
j-xiong, line 41: `Instead ..... should be expected.` --> `In such case, an FI_ENOTCONN error should be returned.`

man/fi_tcp.7.md
```diff
@@ -32,6 +32,16 @@ The following features are supported
...
+  If there is an existing connection, the operation can complete normally.
```
j-xiong, line 42: The last sentence can be removed.

include/rdma/fi_ext.h
```diff
@@ -78,6 +78,11 @@ enum {
 	FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 };
 
+/* provider specific op flags range between 60-63 */
+enum {
+	FI_FLAG_TCP_NO_CONNECT = (1ULL << 60)
```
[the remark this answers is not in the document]
aingerson: Personally, I prefer the flag to be clear about it, similar to how the EP options start with FI_OPT_{PROV}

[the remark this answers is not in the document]
j-xiong: `FI_OPT_` is a common prefix for options passed to fi_setopt/fi_getopt, but we don't use `FI_FLAG_` for other flags.

prov/tcp/src/xnet.h
```diff
@@ -98,6 +98,9 @@ typedef void  xnet_profile_t;
 #define XNET_MIN_MULTI_RECV	16384
 #define XNET_PORT_MAX_RANGE	(USHRT_MAX)
 
+/* provider specific op flags */
+#define TCP_NO_CONNECT FI_TCP_NO_CONNECT
```
j-xiong, line 102: Any reason not to use `FI_TCP_NO_CONNECT` directly?

ooststep: The provider specific flags are typically defined within the provider header files without and FI prefixing and used internally.  I wanted to have a definition here as a placeholder so that it's obvious there are existing provider op flags that need to be accounted for.

man/fi_tcp.7.md
```diff
@@ -32,6 +32,15 @@ The following features are supported
 *Shared Rx Context*
 : The tcp provider supports shared receive context
 
+# PROVIDER SPECIFIC OPERATION FLAGS
+: The tcp provider supports the following op flags
+
+*FI_TCP_NO_CONNECT*
+: This flag indicates that operations should fail if there is no
+  existing connection to the remote peer.  In such case, an FI_ENOTCONN
+  error should be expected.
```
shefty, line 41: I would not make this a flag that's checked on every operation (i.e. it's okay to connect if it's a send, but not a write?).  It makes more sense either applied to the entire rdm endpoint or to a specific peer.

prov/tcp/src/xnet.h
shefty: @ooststep This is true when the API flag differs from the provider flag, or the scope of the flags are different.  E.g. the API flags cover a 64-bit range, but the provider flag maps to some 16-bit range in the protocol.

man/fi_tcp.7.md
ooststep: in the target problematic scenario, the rdm is used for some eps where this flag is needed and some eps where it would not be, so applying to the entire rdm was undesirable.

soumagne: I think I also agree with the first part (ie. exposing low-level connection is a bad idea) though providing 2 types of endpoints is also difficult for the user. To provide some context as to what problem we were trying to solve, in a client-server configuration where the client is behind a firewall, we have cases where for example client A would send a message to server A, and server A would then attempt to do an emulated tcp RMA to client B through an fi_write() call (without prior connection established from client B to server A). In that case, it seems that with the tcp provider server A remains stuck attempting to establish a connection and doing an fi_cancel on the RMA does not seem to be able to complete, as it's not supported currently by the tcp provider. So what we wanted was a way of having some completion and error being returned when server A is not able to reach client B. I'm opened to other means but having to manage 2 endpoints seems also cumbersome, I would also be happy though if we don't have to expose any connection logic.

jolivier23: Just to add a bit of context, this use case came from Parallelstore (Google's DAOS service).  We control only the server side of the equation and rely on the user to do client configuration (as we don't control their VMs, network configuration, and processes).  Opening the firewall to server to client connections is not common for services and requires them to do it explicitly or their writes will simply hang.  We want to remove this requirement as it is becoming a major scale issue for onboarding new customers.

shefty: Client A is telling a server to issue an RMA write to client B?  A knows the memory information on client B, and B is expecting this transfer?  Is this the model used by the app?  Because if so, weird.

In any case, the entire RDM ep is behind a firewall, but connections can only go one direction?  If this is the case, then I think some generic EP setting might be able to convey this, such that the implementation can adjust.  The server could probably just fail the operation (asynchronously makes the most sense) in this situation.

I assume A is somehow notified what happened between the server and B?

soumagne: oh yes sorry I was trying to keep it simple and in the end that didn't make much sense, the real model is client A tells server A, which tells server B about its memory information, which then issues an RMA to client A.

soumagne: > In any case, the entire RDM ep is behind a firewall, but connections can only go one direction? If this is the case, then I think some generic EP setting might be able to convey this, such that the implementation can adjust. The server could probably just fail the operation (asynchronously makes the most sense) in this situation.

yeah I think adding a setting to the server EP would probably be fine in this case, although we'd still want servers to connect to each other using the same endpoint and not use a separate endpoint for intra-communication.

shefty: I think we need to figure out the model for how firewalls should be handled, ignoring the implementation.  Maybe this is related to NAT as well.

An EP is assigned a single address.  Can traffic to that address go through the firewall?  Can traffic from that address go through the firewall?

Assume the implementation is using UDP (or some other connectionless transport).  Does that change anything about how the firewall must be configured?

If I'm understanding this now, what's being requested from the API is: Traffic is allowed to reach an EP (through a firewall), but traffic from that EP cannot reach the target ... unless it's already received some message from the target.  This assumes the EP is responsible for remembering which target addresses it has received data from, and that those addresses still reference the correct target process.

From the viewpoint of the API, this still seems weird.  I think the weirdness comes from the firewall acting on the lower-level connections, which the EP hides.

This is some per peer setting.  But if the application is responsible for providing the value of this setting, then it has an idea how to manage this.  In this example, server B could see if it knows A.  If not, fail sending to it.

One alternative is to configure the EP with firewall data only, and let the provider determine what to do with it.  In this case, a request to send to X checks the firewall config to see if it's possible.  Maybe X is another server within the firewall.  If so, connect and send.  If X is outside the firewall, fail the send.  I don't know what the firewall data would look like but providing it via some configuration file seems reasonable.  I.e. don't make this an application problem.

jolivier23: That alternative is a bit is tricky.   The client may be older than the server and we want old clients to continue behaving as they do now (e..g if the user has already configured their firewall correctly and the client doesn't support getting back an error and retrying, we want it to just continue working).

jolivier23: I think the problem is really on the client.  The client endpoint is the one behind the firewall and can't be reached (unless the server has already connected to it).   Can we encode something into the client side URI that would indicate it can't be reached?   The advantage to that approach from our end is that then the client could have complete control over telling the server whether or not it can handle the error (e.g. can support getting back an error indicating that the server couldn't connect and handle it appropriately)

shefty: It's the server's behavior that should change.

From the viewpoint of the client, everything works.  It wants to talk to server A and can do so.  That server A wants to pass off the response to some other system that the client doesn't know about is related to the storage architecture, not the client SW.

I don't think pushing this detail into the apps is the best option.  But you can work-around this in the client by having the client send some sort of 'hello' message to every storage server during initialization -- to poke holes through the firewall.  That pushes the burden onto every client app that might want to use DAOS.

Alternatively, you can configure the server SW to be firewall aware, so that it avoids forwarding requests to servers not already communicating with the client.

Or, change the protocol around handling firewalls.  Have server A tell the client to retry its request with server B, rather than forwarding it internally.

There are likely other options for this.  But I would avoid picking one which encoded these details in the SW API.

shefty: It would probably be better if this thread were copied into an issue for continued discussion, rather than attaching it to this PR.

j-xiong: #10637

jolivier23:
@ooststep @j-xiong How involved would it be to update this patch according to the proposed change in https://github.com/ofiwg/libfabric/issues/10637. Is it relatively simple or will take some effort and need to be scoped and planned?  Is it something someone on the core team could fit into a schedule or is it something we should try to contribute on our end?

ooststep:
replacing this PR with #10724

PR #10538, opened 2024-11-14
prov/efa: Fix the error path of zero copy recv

shijin-aws:
Fix the resource leak of rxe when the user rx pkt post fails. Make the user rx pkt pool fixed size (rx size) and doesn't allow further grow (for better traffic control).


aws-nslick:
for me, this is consistently spilling on init with:

```
Libfabric EFA provider has encountered an internal error:

Libfabric error: (-12) Cannot allocate memory
EFA internal error: (4112) Failure to post internal receive buffers
```

zachdworkin:
@shijin-aws I made libfabric CI stop. It should be hitting the opt-out stage to not run if all the changes are in prov/efa. I will look into fixing it for the future.

jiaxiyan (approved):

PR #10540, opened 2024-11-14
prov/efa: Post the initial batch of internal rx pkt post during ep enable

shijin-aws:
We cannot only iterate eps and post initial batch of internal rx pkt once, as there can be more eps joining later after the cq read call.

This reverts commit f2a059235a61f16d333778b504e3a22a83104333. and 2740faaa66c61c20981c2b6223d9b47bd2e98163


shijin-aws:
Replaced by https://github.com/ofiwg/libfabric/pull/10543

PR #10543, opened 2024-11-14
prov/efa: Fix the ep list scan in cq/cntr read

shijin-aws:
We cannot only iterate eps and post initial batch of internal rx pkt once, as there can be more eps joining later after the cq read call. This patch fixes by introducing a bit in cq/ctnr that indicates whether a ep list scan is needed. This bit is set as true when a new ep is bind to the cq, and will be set as false every time when a scan is done.


jiaxiyan (approved):

aws-nslick:
tested and confirmed that this fixed an issue with nccl-net-ofi

shijin-aws:
bot:aws:retest

PR #10550, opened 2024-11-18
prov/lnx: Fix various issues with initial commit

amirshehataornl:
1. must pass an ep fid for fi_srx_context to comply with API symantics
2. Don't overwrite attribute flags passed in fi_av_open() by application
3. Enforce the use of FI_AV_TABLE type as the other types have been deprecated
4. Check for NULL entry before calling ofi_mr_cache_delete()

### DIFF
--- a/prov/lnx/src/lnx_av.c
+++ b/prov/lnx/src/lnx_av.c
@@ -630,11 +630,7 @@ int lnx_av_open(struct fid_domain *domain, struct fi_av_attr *attr,
 	if (attr->name)
 		return -FI_ENOSYS;
 
-	if (attr->type != FI_AV_UNSPEC &&
-	    attr->type != FI_AV_TABLE)
-		return -FI_ENOSYS;
-
-	if (attr->type == FI_AV_UNSPEC)
+	if (attr->type != FI_AV_TABLE)
 		attr->type = FI_AV_TABLE;
 
 	peer_tbl = calloc(sizeof(*peer_tbl), 1);
--- a/prov/lnx/include/lnx.h
+++ b/prov/lnx/include/lnx.h
@@ -81,6 +81,7 @@ struct local_prov_ep {
 	struct fid_ep **lpe_txc;
 	struct fid_ep **lpe_rxc;
 	struct fid_av *lpe_av;
+	struct fid_ep *lpe_srx_ep;
 	struct lnx_peer_cq lpe_cq;
 	struct fi_info *lpe_fi_info;
 	struct fid_peer_srx lpe_srx;
--- a/prov/lnx/src/lnx_cq.c
+++ b/prov/lnx/src/lnx_cq.c
@@ -160,11 +160,12 @@ static int lnx_cq_open_core_prov(struct lnx_cq *cq, struct fi_cq_attr *attr)
 	int rc;
 	struct local_prov_ep *ep;
 	struct local_prov *entry;
+	struct fi_cq_attr peer_attr = {0};
 	struct dlist_entry *prov_table =
 		&cq->lnx_domain->ld_fabric->local_prov_table;
 
 	/* tell the core providers to import my CQ */
-	attr->flags |= FI_PEER;
+	peer_attr.flags |= FI_PEER;
 
 	/* create all the core provider completion queues */
 	dlist_foreach_container(prov_table, struct local_prov,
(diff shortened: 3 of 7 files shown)


j-xiong (approved):

prov/lnx/src/lnx_av.c
```diff
@@ -630,11 +630,7 @@ int lnx_av_open(struct fid_domain *domain, struct fi_av_attr *attr,
 	if (attr->name)
 		return -FI_ENOSYS;
 
-	if (attr->type != FI_AV_UNSPEC &&
-	    attr->type != FI_AV_TABLE)
-		return -FI_ENOSYS;
-
-	if (attr->type == FI_AV_UNSPEC)
+	if (attr->type != FI_AV_TABLE)
```
aingerson, line 633: I don't think this is needed. I'm ok with this change for now to get it to work if it is needed but I think the issue is higher up in link not returning the correct av type to begin with. The issue was that the app was passing in FI_AV_MAP because that was what was returned by link so it doesn't make sense for the fix to be to override whatever is passed through. We should return ENOSYS if MAP was requested and have UNSPEC change to TABLE. It looks like the later change with the form_info call setting the av type is probably correct but we should take a closer look.
I think the link getinfo path needs to be reworked in general later anyway so we can put that on the list of things to do for the restructuring.

amirshehataornl: The issue is that the core provider is returning FI_AV_MAP. LINKx returns that to the application in the fi_info and then the fi_info makes it back to LINKx. The main issue here is that LINKx needs to return some information to the application which aids the application in selecting the correct interface to use. This information includes the PCI information. This info allows the application to select the correct interface to use, for example the interface closest to the CPU or GPU the process is bound to. Currently because LINKx supports linking shm + another inter-node provider, I always take the internode provider information (ex: CXI or rxm) and return that back to the application. I have a patch which updates the av type in that path. However, this issue is more complicated, because once we move to binding two interfaces which are both used for inter-node communication, then the question arises which interface information do you bounce back the application. If the application requires the PCI information to decide which interface to use, now we have two sets of information, but one place to put it. One possible solution is to not send anything back and let LINKx decide which interface is best to use based on the message meta data. We still need to think about this more. However, currently, the way the fi_info is being formed is sound. The inter-node core provider fi_info is used with some modifications.

aingerson: Yeah, we should chat about what the returned fi_info looks like for communicating information like what you say.
The AV table implementation here I think is a bandaid for the issue. rxm is returning AV_MAP because it was requested from the app (OMPI defaults to requesting FI_AV_MAP in hints). Overriding it to AV_TABLE is incorrect. It looks like there is a variable that toggles the map vs table request but I'm unclear how to use it (I haven't looked too deeply into it though). Regardless, FI_AV_MAP is getting deprecated in 2.0 to OMPI should remove this toggling and default to FI_AV_TABLE for its implementation. 
```
diff --git a/ompi/mca/mtl/ofi/mtl_ofi_component.c b/ompi/mca/mtl/ofi/mtl_ofi_component.c
index 4fb44ecdf4..c6b1a068da 100644
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -209,10 +209,10 @@ ompi_mtl_ofi_component_register(void)
         return ret;
     }
 
-    av_type = MTL_OFI_AV_MAP;
+    av_type = MTL_OFI_AV_TABLE;
     mca_base_component_var_register (&mca_mtl_ofi_component.super.mtl_version,
                                      "av",
-                                     "Specify AV type to use (default: map). Set to table for FI_AV_TAB>
+                                     "Specify AV type to use (default: table). Set to map for FI_AV_MAP>
                                      MCA_BASE_VAR_TYPE_INT, new_enum, 0, 0,
                                      OPAL_INFO_LVL_3,
                                      MCA_BASE_VAR_SCOPE_READONLY,
```
Part of reworking lnx will be to identify what are things we want to pass back or through and what are things that we request from the core providers. For example, I think lnx will probably require FI_AV_TABLE from its implementation so we can add that as a requirement. 
Like I said, I'm fine with the bandaid for now because lnx is experimental and in development but eventually, OMPI should get changed to request the correct av type and then lnx should have something like:
```
diff --git a/prov/lnx/src/lnx_av.c b/prov/lnx/src/lnx_av.c
index f0b8d09fb8..9b628bfb99 100644
--- a/prov/lnx/src/lnx_av.c
+++ b/prov/lnx/src/lnx_av.c
@@ -631,8 +631,10 @@ int lnx_av_open(struct fid_domain *domain, struct fi_av_attr *attr,
        if (attr->name)
                return -FI_ENOSYS;
 
-       if (attr->type != FI_AV_TABLE)
-               attr->type = FI_AV_TABLE;
+       if (attr->type != FI_AV_TABLE && attr->type != FI_AV_UNSPEC)
+               return -FI_ENOSYS;
+
+       attr->type = FI_AV_TABLE;
 
        peer_tbl = calloc(sizeof(*peer_tbl), 1);
        if (!peer_tbl)
```
To make sure that we don't silently switch the AV type and instead error out properly as expected

PR #10553, opened 2024-11-18
prov/efa: Implement the rma interface

jiaxiyan:
Rename efa_dgram_rma.c to efa_rma.c and move it to prov/efa/src as a common RMA interface for both rdm and dgram ep type. Implement rdma write and inject.
Support inline rdma write.


prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,415 @@
...
+ *     msg: read operation information
+ *     flags: flags passed
+ *     self_comm: indicate whether the write is toward
+ *                the end point itself. If self_comm is true,
+ *                caller must set msg->addr to FI_ADDR_NOTAVAIL.
+ * @return On success return 0, otherwise return a negative libfabric error code.
+ */
+static inline ssize_t efa_rma_post_write(struct efa_base_ep *base_ep,
+					 const struct fi_msg_rma *msg,
+					 uint64_t flags, bool self_comm)
```
shijin-aws, line 196: i think self_comm needs to be defined within this function.

prov/efa/src/dgram/efa_dgram_ep.c
```diff
@@ -444,16 +442,6 @@ int efa_dgram_ep_open(struct fid_domain *domain_fid, struct fi_info *user_info,
 	 */
 	assert(user_info->tx_attr->iov_limit <= 2);
 
-	ret = ofi_bufpool_create(&ep->send_wr_pool,
```
shijin-aws, line 447: I thought this should be part of your earlier change that implements the msg interface? Am I wrong?

prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,417 @@
...
+			sge_list[i].addr = (uint64_t)msg->msg_iov[i].iov_base;
+			sge_list[i].length = msg->msg_iov[i].iov_len;
+			assert(msg->desc && msg->desc[i]);
+			sge_list[i].lkey = ((struct efa_mr *)msg->desc[i])->ibv_mr->lkey;
+		}
+		ibv_wr_set_sge_list(qp->ibv_qp_ex, msg->iov_count, sge_list);
+	}
+
+	if (msg->addr == FI_ADDR_NOTAVAIL) {
+		self_comm = true;
```
shijin-aws, line 245: I believe we used to allow a `FI_ADDR_NOTAVAIL` to support local read. But per libfabric standard we shouldn't allow a NOTAVAIL addr for a user call. So in that regard I think we should drop the L244-246.

prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,417 @@
...
+	} else {
+		conn = efa_av_addr_to_conn(base_ep->av, msg->addr);
+		self_comm = !(conn && conn->ep_addr);
```
shijin-aws, line 248: Is this correct? `self_comm` should be true when the `conn->ep_addr` is identical to the local base_ep's addr.

jiaxiyan: I was referring to https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke.c#L561 where 
 self_comm is true when peer is null and [efa_rdm_ep_get_peer](https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_utils.c#L95-L107) returns NULL when av_entry->conn.ep_addr is NULL.

prov/efa/src/dgram/efa_dgram_ep.c
jiaxiyan: Yes I realized I didn't include it in the last PR

jiaxiyan: I did. Do you mean separate PRs?

shijin-aws: Sorry, I missed that it's already in separate commit. No bother then

prov/efa/src/efa_rma.c
shijin-aws: I see, I think the pke write function was copied from the pke_read function, which needed to handle the local read path in the rdm code base where the peer is set as NULL.

I think for efa-raw, we don't need to handle this case. We should always require a valid `conn`

prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,399 @@
...
+	EFA_SETUP_IOV(iov, buf, len);
+	EFA_SETUP_RMA_IOV(rma_iov, addr, len, key);
+	EFA_SETUP_MSG_RMA(msg, &iov, NULL, 1, dest_addr, &rma_iov, 1, NULL, data);
+
+	return efa_rma_writemsg(ep_fid, &msg, FI_INJECT | FI_REMOTE_CQ_DATA);
+}
+
+struct fi_ops_rma efa_dgram_ep_rma_ops = {
+	.size = sizeof(struct fi_ops_rma),
+	.read = efa_rma_read,
```
shijin-aws, line 377: dgram doesn't support FI_RMA, we should just make ops as `no*`

shijin-aws: Like https://github.com/ofiwg/libfabric/blob/main/src/enosys.c#L520-L570

prov/efa/src/efa_msg.c
```diff
@@ -346,7 +335,7 @@ static ssize_t efa_ep_msg_injectdata(struct fid_ep *ep_fid, const void *buf,
 	EFA_SETUP_IOV(iov, buf, len);
 	EFA_SETUP_MSG(msg, &iov, NULL, 1, dest_addr, NULL, data);
 
-	flags = base_ep->info->tx_attr->op_flags | FI_REMOTE_CQ_DATA | FI_INJECT;
+	flags = FI_REMOTE_CQ_DATA | FI_INJECT;
```
shijin-aws, line 338: why not use `FI_REMOTE_CQ_DATA | FI_INJECT` directly as well?

prov/efa/src/efa_base_ep.h
```diff
@@ -15,6 +15,9 @@
 #define EFA_QP_DEFAULT_SERVICE_LEVEL 0
 #define EFA_QP_LOW_LATENCY_SERVICE_LEVEL 8
 
+#define efa_rx_flags(efa_base_ep) ((efa_base_ep)->util_ep.rx_op_flags)
```
shijin-aws, line 18: nit: I don't see these macros actually simplifies code... maybe just use the original code it represents

jiaxiyan: I was trying to be consistent with efa_rdm_rx_flags, efa_rdm_tx_flags, and smr_ep_rx_flags, smr_ep_tx_flags

prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,399 @@
...
+	       msg->rma_iov_count <= base_ep->domain->info->tx_attr->rma_iov_limit);
+	assert(ofi_total_iov_len(msg->msg_iov, msg->iov_count) <=
+	       base_ep->domain->device->max_rdma_size);
+
+	qp = base_ep->qp;
+	if (!base_ep->is_wr_started) {
+		ibv_wr_start(qp->ibv_qp_ex);
+		base_ep->is_wr_started = true;
+	}
+	qp->ibv_qp_ex->wr_id = (uintptr_t)msg->context;
```
shijin-aws, line 88: I think a future change will be like we will enforce `FI_CONTEXT`, and if flags doesn't have `FI_COMPLETION`, we will ignore `context` and make `wr_id` as NULL. Then in the cq read, we will treat NULL wr_id as the indicator of no-completion. This can be a future PR though.

shijin-aws (approved):

prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,399 @@
...
+	int err;
+
+	base_ep = container_of(ep_fid, struct efa_base_ep, util_ep.ep_fid);
+	assert(len <= base_ep->max_rma_size);
+	err = efa_rma_check_cap(base_ep);
+	if (err)
+		return err;
+
+	EFA_SETUP_IOV(iov, buf, len);
+	return efa_rma_writev(ep_fid, &iov, &desc, 1, dest_addr, addr, key, context);
```
shijin-aws, line 304: Missed this earlier ... Can we call efa_rma_post_write directly here.

prov/efa/src/efa_rma.c
```diff
@@ -0,0 +1,399 @@
...
+	int err;
+
+	base_ep = container_of(ep_fid, struct efa_base_ep, util_ep.ep_fid);
+	assert(len <= base_ep->max_rma_size);
+	err = efa_rma_check_cap(base_ep);
+	if (err)
+		return err;
+
+	EFA_SETUP_IOV(iov, buf, len);
+	return efa_rma_readv(ep_fid, &iov, &desc, 1, src_addr, addr, key, context);
```
shijin-aws, line 172: same comment, can we call efa_rma_post_read directly by setting up the MSG

PR #10560, opened 2024-11-19
prov/efa: Disable resource management when user set FI_OPT_EFA_RNR_RETRY

jiaxiyan:
If a user is explicitly asking for a retry count that means they also want to manage the resources themselves and the EFA provider should disable resource management to prevent implicit retries.

Also remove the unused function.


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -1697,6 +1697,7 @@ static int efa_rdm_ep_setopt(fid_t fid, int level, int optname,
 			return -FI_ENOSYS;
 		}
 		efa_rdm_ep->base_ep.rnr_retry = *(size_t *)optval;
+		efa_rdm_ep->handle_resource_management = FI_RM_DISABLED;
```
shijin-aws, line 1700: Please add a comment here why we disable RM when user set rnr_retry explicitly.

shijin-aws:
Changes LGTM, would u mind adding a unit test?

shijin-aws (approved):

fabtests/prov/efa/src/efa_rnr_shared.c
```diff
@@ -77,9 +77,13 @@ int ft_efa_rnr_init_fabric()
...
 	fprintf(stdout, "Setting RNR retry count to %zu ...\n", rnr_retry);
-	ret = fi_setopt(&ep->fid, FI_OPT_ENDPOINT, FI_OPT_EFA_RNR_RETRY, &rnr_retry, sizeof(rnr_retry));
+	/* 
+	 * Cannot use setopt FI_OPT_EFA_RNR_RETRY here because it sets FI_RM_DISABLED, 
+	 * which conflicts with the queue/resend test that requires FI_RM_ENABLED.
+	 */
+	ret = setenv("FI_EFA_RNR_RETRY", "0", 1);
 	if (ret) {
-		FT_PRINTERR("fi_setopt", -ret);
+		FT_PRINTERR("setenv", -ret);
```
shijin-aws, line 86: I don't think setenv works here because `fi_param_get` happens at fi_getinfo where we already have a efa_env.rnr_retry set as 3.

Also, this is shared among the rnr_cq_read_error and queue_resend test. The former intends to turn off RM, the latter one intends to turn on RM. It's totally fine for rnr_cq_read_error to call fi_setopt, but not queue_resend.

So we should have a bool somewhere to call fi_setopt only for cq_read_error test. And make queue_resend_test setenv before initing any resources

shijin-aws: Looking at the code again, Nah. rnr_retry i not really environment variable even though it is in `efa_env` so setenv won't work at all.

shijin-aws:
We likely won't prioritize working on this PR in the near future.

PR #10570, opened 2024-11-21
Make util_ns fileds: run and is_initialized atomics

dsciebu:
Values of util_ns fields can be read/write in parallel from multiple threads. Making them atomics secures these kind of access.

### DIFF
--- a/prov/util/src/util_ns.c
+++ b/prov/util/src/util_ns.c
@@ -301,7 +301,7 @@ static void *util_ns_accept_handler(void *args)
 	struct util_ns_cmd cmd;
 	int ret;
 
-	while (ns->run) {
+	while (ofi_atomic_get32(&ns->run)) {
 		conn_sock = accept(ns->listen_sock, NULL, 0);
 		if (conn_sock == INVALID_SOCKET)
 			break;
@@ -367,7 +367,7 @@ int ofi_ns_add_local_name(struct util_ns *ns, void *service, void *name)
 		.op = OFI_UTIL_NS_ADD,
 	};
 
-	if (!ns->is_initialized) {
+	if (!ofi_atomic_get32(&ns->is_initialized)) {
 		FI_WARN(&core_prov, FI_LOG_CORE,
 			"Cannot add local name - name server uninitialized\n");
 		return -FI_EINVAL;
@@ -415,7 +415,7 @@ int ofi_ns_del_local_name(struct util_ns *ns, void *service, void *name)
 		.op = OFI_UTIL_NS_DEL
 	};
 
-	if (!ns->is_initialized)
+	if (!ofi_atomic_get32(&ns->is_initialized))
 		return -FI_EINVAL;
 
 	write_buf = calloc(cmd_len + ns->service_len + ns->name_len, 1);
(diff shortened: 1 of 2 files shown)


prov/util/src/util_ns.c
```diff
@@ -592,12 +592,12 @@ void ofi_ns_init(struct util_ns *ns)
...
-	if (ns->is_initialized)
+	if (ofi_atomic_get32(&ns->is_initialized))
 		return;
 
 	ofi_atomic_initialize32(&ns->ref, 0);
 	ns->listen_sock = INVALID_SOCKET;
 	if (!ns->hostname)
 		ns->hostname = OFI_NS_DEFAULT_HOSTNAME;
-	ns->is_initialized = 1;
+	ofi_atomic_set32(&ns->is_initialized, 1);
```
shefty, line 602: The name service is basically a simple way to support running fabtests.  It's not something we expect to have enabled as a general purpose feature.

If it were, the changes here are not sufficient to support multiple threads.  Locks would be needed.  It's ultimately the responsibility of the caller to have appropriate serialization to the name service calls.

dsciebu: Thanks for the tip about the name service. I found that it is by default enabled for verbs provider, which, in case of a multithreaded application, leads to data races triggered by e.g. parallel fi_close and util_ns_accept_handler execution. Don't you think, that by as you said: "not expecting this option to be enabled as a general purpose feature" you should default this option to OFF?

shefty: Can you describe what sort of endpoints the app is using?  It looks like the name service is related to dgram ep's, and only started if a verbs domain is opened for dgram ep's.  Note that this can occur through the rxd provider for rdm ep's.

@j-xiong - I see that the name server is enabled by default, but disabled if specific, unrelated environment variables are set (OMPI_COMM_WORLD_RANK or PMI_RANK).  Should the default be to disable it, and use an environment variable to enable instead?

I'm unsure how frequently the verbs dgram provider is used in practice.

j-xiong: @shefty The name server is needed for verbs dgram to resolve "node" into dst_addr in fi_getinfo().  By default this is a feature an application may use. MPIs (OpenMPI, MPICH, Intel MPI) don't use this feature so the name server can be safely disabled.

dsciebu: My experiments that revealed the issue were made on a Mellanox RDMA card. Using that hardware and setting up the hints the way that rdm fabtests do, ended up in getting an RXD protocol 'based' endpoint. 
When checking my libfabric based app's test, I noticed that Thread Sanitizer reports data race in libfabric and ended up here. Turning off the 'name service' with the proper env variable solves the problem completely.
So, IMO it should be:
 a) turned off by default if not necessary or 
 b) made thread safe.

shefty: @dsciebu - I agree.  Can you open an issue to track this?  The name service should be made thread safe, which I'm guessing won't be that difficult.  Likely just needs a lock around most of the functions you identified.

zachdworkin:
@dsciebu please rebase to pickup Intel CI changes

zachdworkin:
debug build of verbs-rxd is failing most fabtests with this error: "./include/ofi_atom.h:420: ofi_atomic_get32: Assertion `atomic->is_initialized' failed".
debug build of verbs-rxm is failing: "fi_getinfo_test: ./include/ofi_atom.h:420: ofi_atomic_get32: Assertion `atomic->is_initialized' failed."

PR #10572, opened 2024-11-21
prov/lnx: fix av strncpy

aingerson:
Fix compiler error with strncpy:

prov/lnx/src/lnx_av.c:272:4: warning: ‘strncpy’ specified bound 64 equals destination size [-Wstringop-truncation]
    strncpy(peer_prov->lpp_prov_name, prov_name, FI_NAME_MAX);

Both dest and srouce are FI_NAME_MAX but strncpy is complaining that it doesn't have space for the NULL terminator if it needs to be truncated. The terminator should already be in the prov_name passed in. Just turn this into a memcpy to make the compiler happy


amirshehataornl (approved):

j-xiong:
Probably good enough for suppressing the compiler warning. However, throughout the provider this constant is used inconsistently in strncpy, strncasecmp, memcpy, etc, with or without subtracting one. Sooner or later this needs to be cleaned up. Furthermore, we should stay away from using FI_NAME_MAX and use a provider private definition instead.

PR #9974, opened 2024-04-03
prov/efa: Map EFA errnos to Libfabric codes

darrylabbate:
This adds a rudimentary function to map proprietary EFA status codes to common Libfabric status codes. This is useful when reporting errors to the application for operations that rely solely on ibverbs or RDMA Core, such as CQ polling.

PR #10576, opened 2024-11-25
prov/efa: Adjust the error code for flushed receive

shijin-aws:
Currently, efa provider propagate flushed qp error as FI_EHOSTDOWN, which is wrong as flushed QP error means the completions are canceled due to QP was
aborted earlier. Using FI_ECANCELED is more reasonable here.


jiaxiyan (approved):

darrylabbate:
> flushed QP error means the completions are canceled due to QP was
> aborted earlier

Double checking; you originally requested `EHOSTDOWN` for this same reason: https://github.com/ofiwg/libfabric/pull/9974#discussion_r1550234385

shijin-aws:
> > flushed QP error means the completions are canceled due to QP was
> > aborted earlier
> 
> Double checking; you originally requested `EHOSTDOWN` for this same reason: [#9974 (comment)](https://github.com/ofiwg/libfabric/pull/9974#discussion_r1550234385)

Yeah, I think we need to use a different one.

PR #10582, opened 2024-11-26
prov/cxi: synchronous fi_close on collective multicast

bulbul-hpe:
The fi_close() operation manages its internal state and return FI_SUCCESS, or a fatal error code on error.

### DIFF
--- a/prov/cxi/src/cxip_coll.c
+++ b/prov/cxi/src/cxip_coll.c
@@ -2704,13 +2704,17 @@ static void _curl_delete_mc_obj(struct cxip_coll_mc *mc_obj);
 static void _cxip_delete_mcast_cb(struct cxip_curl_handle *handle);
 
 /* Close multicast collective object */
-static void _close_mc(struct cxip_coll_mc *mc_obj, bool delete)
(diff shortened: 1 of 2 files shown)


prov/cxi/src/cxip_coll.c
```diff
@@ -3073,6 +3089,7 @@ static void _curl_delete_mc_obj(struct cxip_coll_mc *mc_obj)
 quit:
 	free(url);
 	if (ret < 0) {
+		mc_obj->close_state = ret;
```
swelch, line 3092: Here and below where you set mc_obj->close_state = ret and the call free(mc_obj). Why is this being done?

bulbul-hpe: Accessing memory after free. Works by chance but is undefined behavior. Fixing it.

heemstra (approved):

heemstra (changes requested):
Removing approval to prevent accidental merge.

swelch (approved):
Thanks for the updates.

heemstra (approved):

PR #10571, opened 2024-11-21
prov/verbs: Fix data race vrb_open_ep function

piotrchmiel:
Race Fix: Resolved a data race in the vrb_open_ep function of the verbs provider caused by concurrent modifications to the global variable vrb_ep_ops. This issue violated the FI_THREAD_SAFE threading model, leading to unpredictable behavior when creating endpoints from multiple threads.

Detailed race description: https://github.com/ofiwg/libfabric/issues/10569

Signed-off-by: Piotr Chmiel  <piotr.chmiel@hotmail.com>

Issue #10569, opened 2024-11-21
prov/verbs: Data race in vrb_open_ep function (Libfabric v1.22.0)


piotrchmiel:
**Describe the bug**
A data race occurs in the vrb_open_ep function in the verbs provider when creating multiple endpoints using multiple threads. The race is detected by Thread Sanitizer and involves concurrent modifications of the global variable vrb_ep_ops. The issue causes unpredictable behavior when using the FI_THREAD_SAFE threading model, which should guarantee safe concurrent operations.

**To Reproduce**

1. Set up an environment with Libfabric 1.22.0, Ubuntu 22.04, and the verbs provider.
2. Use the FI_THREAD_SAFE threading mode.
3. Create multiple endpoints (fi_endpoint) from multiple threads simultaneously.
4. Observe the data race using Thread Sanitizer.

**Expected behavior**
The FI_THREAD_SAFE threading mode should ensure thread safety, allowing multiple threads to create endpoints concurrently without encountering data races.

**Output**
Thread Sanitizer detects the following data race:

```
WARNING: ThreadSanitizer: data race (pid=76726)
  Write of size 8 at 0x7ff67d69a3b8 by thread T2:
    #0 vrb_open_ep /path/to/libfabric/prov/verbs/src/verbs_ep.c:1397:31 (libfabric.so.1+0x100f27)
    #1 fi_endpoint(fid_domain*, fi_info*, fid_ep**, void*) /path/to/libfabric/include/rdma/fi_endpoint.h:187:9 (application+0x6658b0)

 Previous write of size 8 at 0x7ff67d69a3b8 by thread T4:
    #0 vrb_open_ep /path/to/libfabric/prov/verbs/src/verbs_ep.c:1397:31 (libfabric.so.1+0x100f27)
    #1 fi_endpoint(fid_domain*, fi_info*, fid_ep**, void*) /path/to/libfabric/include/rdma/fi_endpoint.h:187:9 (application+0x6658b0)
 ```
 
The data race occurs at: https://github.com/ofiwg/libfabric/blob/v1.22.0/prov/verbs/src/verbs_ep.c#L1397

This line modifies the global variable vrb_ep_ops:

```
(*ep_fid)->fid.ops->ops_open = vrb_ep_ops_open;
```

The global variable vrb_ep_ops is defined here: https://github.com/ofiwg/libfabric/blob/v1.22.0/prov/verbs/src/verbs_ep.c#L1159

[...]

PR #10588, opened 2024-11-28
core/logging: Fix race in writing to vrb_ep_ops

dsciebu:
'vrb_ep_ops' is a static object which field 'ops_open' gets overriden in 'vrb_open_ep'. If the operation is performed by multiple threads at once, which is totally possible, we end up with a data race. Switching it to thread local fixes it.

### DIFF
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -191,7 +191,7 @@ ssize_t vrb_post_send(struct vrb_ep *ep, struct ibv_send_wr *wr, uint64_t flags)
 	}
 
 	if (vrb_wr_consumes_recv(wr)) {
-		if  (!ep->peer_rq_credits || 
+		if  (!ep->peer_rq_credits ||
 		     (ep->peer_rq_credits == 1 && !(flags & OFI_PRIORITY)))
 		/* Last credit is reserved for credit update */
 			goto freectx;
@@ -1156,7 +1156,7 @@ static int vrb_dgram_ep_getname(fid_t ep_fid, void *addr, size_t *addrlen)
 	return FI_SUCCESS;
 }
 
-static struct fi_ops vrb_ep_ops = {
+_Thread_local static struct fi_ops vrb_ep_ops = {
 	.size = sizeof(struct fi_ops),
 	.close = vrb_ep_close,
 	.bind = vrb_ep_bind,


dsciebu:
There is a bunch of static variables spread all over the place. IMHO they all should be marked as _Thread_local, as, even if they are not overridden somewhere now, they can be in future leading to another race. If you agree, just let know - I can modify my PR. Otherwise, let's stick to this ver.

piotrchmiel:
Isn't it duplicate of https://github.com/ofiwg/libfabric/pull/10571/ and https://github.com/ofiwg/libfabric/issues/10569 ?

dsciebu:
> Isn't it duplicate of #10571 and #10569 ?

It is in a sense that it generally solves the same issue. However what I am promoting here is a general rule of marking static globals with _Thread_local as probably there are still used concurrently in several places and will probably be used that way in the future. IMO the best solution would be to rethink the architecture and refactor the existing code, but as a quick fix adding thread locality to variables like the one mentioned may help fix a lot of UBs for pretty much 'free'.

dsciebu:
I accidentally found another spot with an identical bug, this time in xnet_ep.c. The line:
`	(*ep_fid)->fid.ops->ops_open = xnet_ep_ops_open;`
overwrites the global assigned to (*ep_fid)->fid.ops assigned in the following line:
`	(*ep_fid)->fid.ops = &xnet_ep_fi_ops;`
I will push a fix for it in a separate PR, but please treat it as an argument for my idea of marking the globals with thread_local.

dsciebu:
> I accidentally found another spot with an identical bug, this time in xnet_ep.c. The line: ` (*ep_fid)->fid.ops->ops_open = xnet_ep_ops_open;` overwrites the global assigned to (*ep_fid)->fid.ops assigned in the following line: ` (*ep_fid)->fid.ops = &xnet_ep_fi_ops;` I will push a fix for it in a separate PR, but please treat it as an argument for my idea of marking the globals with thread_local. Edit: The fix mentioned is #10590.

@shefty - counting on your voice!

shefty:
Correct fix is https://github.com/ofiwg/libfabric/pull/10571.

The static structs referencing function pointers is desirable.  Consider an application which allocates 10k endpoints which are nearly identical.  Each endpoint has a single pointer to a static struct, where the function pointers are stored.  When an app calls fi_send(), fi_write(), etc. those functions are inline wrappers which access the correct underlying function.

The benefit is that the size of each endpoint only increases by 1 pointer, rather than the size of the entire struct.  It also allows adding new functions to the end of the struct while maintaining backwards compatibility.  That wouldn't be possible if the structs were embedded directly into the user visible object.  The cost is that 2 pointers must be dereferenced when making a function call, rather than only 1.

shefty:
Fixed by #10571

PR #10594, opened 2024-12-02
prov/efa: Skip rx pkt refill under certain threshold

shijin-aws:
Libfabric currently refill the rx pkt pool in every cq read when there are >0 pkts to post, which makes it have chance to post ibv_recv 1-by-1 if there is only 1 pkt to post per cq read. Such 1-by-1 post is less performant than having a batch post once.

This patch improves this strategy by introducing a threshold for the refilling. When the number of internal rx pkts to post is lower than this threshold, the refill will be skipped.

Also introduced FI_EFA_INTERNAL_RX_REFILL_THRESHOLD that allows tuning this parameter.


shijin-aws:
bot:aws:retest

jiaxiyan (approved):

PR #10597, opened 2024-12-03
prov/lnx: Convert peer table to use buffer pools

amirshehataornl:
Convert peer table to use buffer pools in order to utilize the built-in capabilities of expanding the table as more peers are added dynamically.

The peer table is protected by a read/write lock. Locking/unlocking callbacks are set only if needed.

### DIFF
--- a/prov/lnx/include/lnx.h
+++ b/prov/lnx/include/lnx.h
@@ -33,7 +33,6 @@
 #ifndef LNX_H
 #define LNX_H
 
-#define LNX_DEF_AV_SIZE 1024
 #define LNX_MAX_LOCAL_EPS 16
 #define LNX_IOV_LIMIT 4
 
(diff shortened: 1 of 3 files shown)


prov/lnx/include/lnx.h
```diff
@@ -200,10 +200,13 @@ struct lnx_peer {
...
 	struct util_av lpt_av;
 	int lpt_max_count;
-	int lpt_count;
 	struct lnx_domain *lpt_domain;
-	/* an array of peer entries */
-	struct lnx_peer **lpt_entries;
+	pthread_rwlock_t lpt_lock;
+	void (*lpt_rlock)(pthread_rwlock_t *l);
+	void (*lpt_wlock)(pthread_rwlock_t *l);
+	void (*lpt_unlock)(pthread_rwlock_t *l);
```
aingerson, line 207: I think the fine tuning of the locks here is unnecessary and we can just use the lock in the util_av (which we can turn into a mutex to optimize out). If an application is interested in performant threading, then they should be using separate domains (and thus AVs) for their threads.

aingerson (approved):
Rerunning through CI because of issues after switching clusters. Will merge and cherry pick after

PR #10618, opened 2024-12-09
prov/shm,rxm,efa, fabtests: fix min multi recv size setting

### LINKED ISSUE #10591: prov/shm: fi_setopt causes a segmentation fault when setting FI_OPT_MIN_MULTI_RECV
**Describe the bug**
`fi_setopt` causes a segmentation fault when setting `FI_OPT_MIN_MULTI_RECV` before enabling the endpoint. 
According to `fi_endpoint` man page, it is recommended to set `FI_OPT_MIN_MULTI_RECV` before enabling the endpoint.
Doing so, causes a segmentation fault with the following backtrace:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2613126==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x7f128262fc1d bp 0x7f11fab08410 sp 0x7f11fab082d0 T2)
==2613126==The signal is caused by a READ memory access.
==2613126==Hint: address points to the zero page.
    #0 0x7f128262fc1d in smr_ep_setopt (/lib/libfabric.so.1+0xabc1d) (BuildId: 3f0a904075fdd37c2719c171f33bd661bbe68aeb)
    #1 0x7f128280d103 in fi_setopt /include/rdma/fi_endpoint.h:232:9
    #2 0x7f128280d103 in mstro_ep_build_from_ofi maestro-core/maestro/ofi.c:1198:14
```

However, setting `FI_OPT_MIN_MULTI_RECV` after enabling the endpoint works for the `shm` provider (unlike other providers and opposite to what is recommended in the man pages).

**To Reproduce**
set `FI_OPT_MIN_MULTI_RECV` before enabling the endpoint.

**Expected behavior**
The code should work.

**Output**
The application fails with a seg fault as above

**Environment:**
OS (if not Linux), provider, endpoint type, etc.
libfabric 1.21.0
shm provider
icx compiler 2023.2.0

aingerson:
Fixes #10591

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -128,14 +128,12 @@ int smr_ep_setopt(fid_t fid, int level, int optname, const void *optval,
 {
 	struct smr_ep *smr_ep =
 		container_of(fid, struct smr_ep, util_ep.ep_fid);
-	struct util_srx_ctx *srx;
 
 	if (level != FI_OPT_ENDPOINT)
 		return -FI_ENOPROTOOPT;
 
 	if (optname == FI_OPT_MIN_MULTI_RECV) {
-		srx = smr_ep->srx->ep_fid.fid.context;
-		srx->min_multi_recv_size = *(size_t *)optval;
+		smr_ep->min_multi_recv_size = *(size_t *)optval;
 		return FI_SUCCESS;
 	}
 
@@ -1146,7 +1144,7 @@ static int smr_ep_ctrl(struct fid *fid, int command, void *arg)
 					      util_domain.domain_fid);
 			ret = util_ep_srx_context(&domain->util_domain,
 					ep->rx_size, SMR_IOV_LIMIT,
-					SMR_INJECT_SIZE, &smr_update,
+					ep->min_multi_recv_size, &smr_update,
 					&ep->util_ep.lock, &srx);
 			if (ret)
 				return ret;
@@ -1307,6 +1305,8 @@ int smr_endpoint(struct fid_domain *domain, struct fi_info *info,
 
 	dlist_init(&ep->sar_list);
 	dlist_init(&ep->ipc_cpy_pend_list);
+
+	ep->min_multi_recv_size = SMR_INJECT_SIZE;
 
 	ep->util_ep.ep_fid.fid.ops = &smr_ep_fi_ops;
 	ep->util_ep.ep_fid.ops = &smr_ep_ops;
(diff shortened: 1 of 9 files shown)


prov/shm/src/smr_ep.c
```diff
@@ -128,14 +128,12 @@ int smr_ep_setopt(fid_t fid, int level, int optname, const void *optval,
...
 		container_of(fid, struct smr_ep, util_ep.ep_fid);
-	struct util_srx_ctx *srx;
 
 	if (level != FI_OPT_ENDPOINT)
 		return -FI_ENOPROTOOPT;
 
 	if (optname == FI_OPT_MIN_MULTI_RECV) {
-		srx = smr_ep->srx->ep_fid.fid.context;
-		srx->min_multi_recv_size = *(size_t *)optval;
+		smr_ep->min_multi_recv_size = *(size_t *)optval;
```
shijin-aws, line 136: I think EFA provider has the same bug: I was copying it from shm :(

https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L1666-L1667

aingerson: @shijin-aws Oops! Sorry!! I will add the fix to efa as well

shijin-aws: Reading the EFA code, it seems enough to remove L1666 and L1667

j-xiong (dismissed):

shijin-aws:
@aingerson Would you mind replacing your efa commit by https://github.com/shijin-aws/libfabric/commit/9eb3c796a707f8b74544d849e1524debc626c82d. I fixed a bug in our unit test which makes this bug not caught earlier

j-xiong:
Intel CI failed with fi_multi_recv. The reason is that the test calls fi_setopt() after enabling the EP, which won't be effective with the new code.

shijin-aws:
> Intel CI failed with fi_multi_recv. The reason is that the test calls fi_setopt() after enabling the EP, which won't be effective with the new code.

Yikes, that explained why the bug was not discovered earlier.

shijin-aws:
AWS CI failure should be fixed after ingesting https://github.com/shijin-aws/libfabric/commit/9eb3c796a707f8b74544d849e1524debc626c82d

aingerson:
Man, that was just one after another issue. This new version should fix it now - I picked your path @shijin-aws and I also added a patch in fabtests to call setopt before enable instead of after

shijin-aws (approved):

j-xiong (approved):

PR #8119, opened 2022-10-17
fabtests/common: close mr before closing ep for FI_MR_ENDPOINT only

shijin-aws:
In commit 1b12e5caa, mr close is moved before the ep close unconditionally. For non-FI_MR_ENDPOINT case, it is not safe to close mr before closing ep as the mr may still be used during ep finalization. This patch makes ft_free_res close mr before closing ep only when mr mode is FI_MR_ENDPOINT.

Signed-off-by: Shi Jin <sjina@amazon.com>

PR #10640, opened 2024-12-17
util/pingpong: close mr after ep close

shijin-aws:
pingpong doesn't support FI_MR_ENDPOINT today,
so the mr is associated with domain instead of ep. It is unsafe to close mr before closing ep because it can cause an EBUSY error when there are outstanding recvs of the mr posted to the ep/qp. This patch fixes this issue by moving the mr close after the ep close.

### DIFF
--- a/util/pingpong.c
+++ b/util/pingpong.c
@@ -1875,12 +1875,12 @@ static void pp_free_res(struct ct_pingpong *ct)
 {
 	PP_DEBUG("Freeing resources of test suite\n");
 
-	if (ct->mr != &(ct->no_mr))
-		PP_CLOSE_FID(ct->mr);
 	PP_CLOSE_FID(ct->ep);
 	PP_CLOSE_FID(ct->pep);
 	PP_CLOSE_FID(ct->rxcq);
 	PP_CLOSE_FID(ct->txcq);
+	if (ct->mr != &(ct->no_mr))
+		PP_CLOSE_FID(ct->mr);
 	PP_CLOSE_FID(ct->av);
 	PP_CLOSE_FID(ct->eq);
 	PP_CLOSE_FID(ct->domain);


shijin-aws:
I can implement a more comprehensive change like https://github.com/ofiwg/libfabric/pull/8119 for this, but to my knowledge we likely won't support FI_MR_ENDPOINT for fi_pingpong, so I just make this simple change.

aingerson:
I don't necessarily have an objection to this change but I'm trying to understand what outstanding operations this addresses - the finalize calls a send with FI_TRANSMIT_COMPLETE to consume the last rx and waits for the matching receive from the other side. Shouldn't all sends and receives be done with the MRs at the point of closing them?

shijin-aws:
@aingerson there is 1 extra fi_recv posted in this test: https://github.com/ofiwg/libfabric/blob/main/util/pingpong.c#L2216-L2222, so there is always 1 extra fi_recv than the sends it will actually get. There is already 1 fi_recv after enable which matches the finalize one.

I don't fully understand why dgram pingpong has to be implemented in this way, especially fabtests fi_dgram_pingpong has such extra recv as well. I tried to remove such extra fi_recv, but then I will frequently get a cq read timeout.

shijin-aws:
Is Intel CI failure related?

aingerson:
@shijin-aws No, they were weird cluster issues. We don't test fi_pingpong in our CI. You can merge whenever you want, thanks!

aingerson (approved):

shijin-aws:
@shefty @j-xiong do you have any background for that 1 extra recv for dgram pingpong?

shijin-aws:
@shefty Thanks! I will keep such extra recv as is for now, and adjust the sequence of mr close in pingpong to be consistent with dgram_pingpong for non-`FI_MR_ENDPOINT` case.

PR #10642, opened 2024-12-17
contrib/intel/jenkins: Uplevel pre-build, always cleanup, update slurm partition names

zachdworkin:
Uplevel pre-build so that it isn't copied over for summarizing
Always cleanup at the end of a pipeline
Update slurm partition names for new cluster
Put mpichtestsuite stage as first test to execute because it is the slowest. This allows other tests to complete in parallel on the other nodes while mpichtestsuite is running.


zachdworkin:
No need to wait for AWS CI since these are all Intel CI changes

PR #10643, opened 2024-12-17
fabtests: corrected flags argument type in ft_sendmsg/ft_recvmsg

nikhilnanal:
fixed fabtests send and recv functions to use flags argument type as uint64_t instead of int as the underlying fi calls use uint64_t. removed declaration of unused function ft_writemsg from shared.h

### DIFF
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -425,7 +425,7 @@ static int rma_bw_rx_comp()
 	return ft_tx(ep, remote_fi_addr, FT_RMA_SYNC_MSG_BYTES, &tx_ctx);
 }
 
-static int set_fi_more_flag(int i, int j, int flags)
+static uint64_t set_fi_more_flag(int i, int j, uint64_t flags)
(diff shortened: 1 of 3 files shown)


aingerson (commented):
Can you double check all places where these functions are called and make sure the correct type is passed in as well? For example [here](https://github.com/ofiwg/libfabric/blob/main/fabtests/benchmarks/benchmark_shared.c#L484)

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -425,7 +425,7 @@ static int rma_bw_rx_comp()
 	return ft_tx(ep, remote_fi_addr, FT_RMA_SYNC_MSG_BYTES, &tx_ctx);
 }
 
-static int set_fi_more_flag(int i, int j, int flags)
+static int set_fi_more_flag(int i, int j, uint64_t flags)
```
aingerson, line 428: the flags are returned here - should be uint64_t as well

aingerson (approved):
Thanks!

PR #10649, opened 2024-12-18
prov/efa: Decouple AV entry from endpoint

sunkuamzn:
Before this change, the EFA AV entry contained a reference to efa_rdm_peer which is specific to a given endpoint. This member also prevented binding a single AV to multiple endpoints.

This change removes efa_rdm_peer from AV entry  by adding a hashmap to the endpoint that maps fi_addr to efa_rdm_peer. And it also enables multiple EFA endpoints to bind to the same AV.

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -243,9 +243,6 @@ void efa_ah_release(struct efa_av *av, struct efa_ah *ah)
 	}
 }
 
-static
-void efa_conn_release(struct efa_av *av, struct efa_conn *conn);
-
 /**
  * @brief initialize the rdm related resources of an efa_conn object
  *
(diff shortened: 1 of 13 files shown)


prov/efa/src/rdm/efa_rdm_ep.h
```diff
@@ -187,6 +191,9 @@ struct efa_rdm_ep {
 	struct dlist_entry entry;
 	/* the count of opes queued before handshake is made with their peers */
 	size_t ope_queued_before_handshake_cnt;
+
+	struct ofi_bufpool *peer_map_entry_pool; 	/* bufpool to hold fi_addr->efa_rdm_peer key-value pairs */
+	struct efa_rdm_peer_map peer_map; 		/* Hashmap to find efa_rdm_peer given fi_addr */
```
jiaxiyan, line 196: The name peer_map feels a little generic.

sunkuamzn: What would you prefer instead? `fi_add_to_peer_map`?

jiaxiyan: Yeah that's better

prov/efa/src/efa_av.c
```diff
@@ -266,18 +263,12 @@ int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn, bool insert_shm_
...
 	char smr_name[EFA_SHM_NAME_MAX];
 	size_t smr_name_len;
-	struct efa_rdm_ep *efa_rdm_ep;
-	struct efa_rdm_peer *peer;
 
 	assert(av->ep_type == FI_EP_RDM);
 	assert(conn->ep_addr);
 
-	/* currently multiple EP bind to same av is not supported */
 	assert(!dlist_empty(&av->util_av.ep_list));
```
shijin-aws, line 270: do you need such assert now since av and peer is decoupled?

sunkuamzn: That means applications can do av insert before the AV is bound to any endpoints. Is that legal? Can't figure out from the documentation.

shijin-aws:
Can you also add some unit tests on the av insert / av removal path.. ?
We should also add `shared av` test in the multi_ep test https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/efa/test_multi_ep.py#L3-L11, shared av is enabled via `-A` for that test

prov/efa/src/efa_av.c
shijin-aws: Yeah, I don't know why it is not legal, and docs doesn't say it's illegal. You should be able to insert an address to the av as long as it's a valid ep name

shijin-aws (approved):

jiaxiyan (approved):

PR #10651, opened 2024-12-19
prov/lpp: add check for atomics

aingerson:
lpp includes stdatomic.h but does not include a check for it in the configure so can cause a build to fail on a system without it


aingerson:
@tstruk noticed this config issue, please take a look and see if that's ok with you or if you want to add a more appropriate check/implementation that doesn't require atomics

tstruk:
@aingerson LGTM thanks

aingerson:
@shijin-aws Can you confirm that the AWS error is unrelated?

shijin-aws:
@aingerson unrelated infrastructure error. Can ignore

PR #10659, opened 2024-12-23
prov/efa: Implement the cq progress

jiaxiyan:
Rename efa_dgram_cq.c to efa_cq.c and move it to prov/efa/src as a common CQ interface for efa-raw’s rdm and dgram ep type.
Create efa_cq_progress and poll ibv cq directly instead of using efa_dgram_ep_progress.
Remove rcq and scq, write completion to util_ep.rx_cq and tx_cq.
Construct the error message for cq_err_entry.err_data.


prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+
+	/* wr_id = NULL means no completion */
+	if (!ibv_cq_ex->wr_id)
+		return;
+
+	cq_flags = (base_ep->util_ep.rx_msg_flags == FI_COMPLETION) ?
+			   0 :
+			   FI_SELECTIVE_COMPLETION;
+
+	if (ofi_need_completion(cq_flags, cq_entry->flags)) {
```
shijin-aws, line 215: same comment as earlier, you should know whether you need completion when posting the recv (setting wr_id).

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+	err_entry.buf = NULL;
+	err_entry.data = 0;
+	err_entry.tag = 0;
+
+	if (ibv_wc_read_wc_flags(ibv_cq_ex) & IBV_WC_WITH_IMM)
+		err_entry.data = ibv_wc_read_imm_data(ibv_cq_ex);
+
+	addr = efa_av_reverse_lookup(base_ep->av,
+				     ibv_wc_read_slid(ibv_cq_ex),
+				     ibv_wc_read_src_qp(ibv_cq_ex));
```
shijin-aws, line 117: This is wrong, `ibv_wc_read_src_qp` only applied to the src addr, which means it only applied to the recv call. We need to determine the destination qpn for a send call. If there is an rdma-core API that can does it, we should use it. If it doesn't exist, we need a plan to how to get such addr.

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+			EFA_WARN(FI_LOG_CQ,
+				"Unable to write send completion: %s\n",
+				fi_strerror(-ret));
+			efa_cq_handle_error(base_ep, ibv_cq_ex, tx_cq, -ret,
+					    FI_EFA_ERR_WRITE_SEND_COMP);
+			return;
+		}
+	}
+
+	if (cq_entry->flags != 0)
```
shijin-aws, line 186: IIRC cntr should always be updated for any completion. In what cases you will have a `cq_entry->flags` as 0?

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+			break;
+		}
+
+		cq->read_entry(cq->ibv_cq.ibv_cq_ex, &cq_entry);
+
+		switch (opcode) {
+		case IBV_WC_SEND: /* fall through */
+		case IBV_WC_RDMA_WRITE: /* fall through */
+		case IBV_WC_RDMA_READ:
+			efa_cq_handle_send_completion(
```
shijin-aws, line 362: Then you need a general name than `send` here. Using `tx` is better

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+				base_ep, cq->ibv_cq.ibv_cq_ex, &cq_entry);
+			break;
+		case IBV_WC_RECV:
+			efa_cq_handle_recv_completion(
```
shijin-aws, line 366: What about `recv`->`rx` here?

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+	.readerr = ofi_cq_readerr,
+	.sread = fi_no_cq_sread,
+	.sreadfrom = fi_no_cq_sreadfrom,
+	.signal = fi_no_cq_signal,
+	.strerror = efa_cq_strerror
+};
+
+void efa_cq_progress(struct util_cq *cq)
+{
+	efa_cq_poll_ibv_cq(efa_env.efa_cq_read_size, cq);
```
shijin-aws, line 431: As I mentioned earlier, we may need a domain level lock here to protect the `qp_table` which is also accessed by the `ep_enable` and `ep_close` calls. But I am ok with making it as part of a follow up change when we implement the `ep` interface for efa raw.

shijin-aws (commented):
Some initial comments, will take a closer look later

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+				 "Unable to write recv completion: %s\n",
+				 fi_strerror(-ret));
+			efa_cq_handle_error(base_ep, ibv_cq_ex, rx_cq, -ret,
+					    FI_EFA_ERR_WRITE_SEND_COMP);
+			return;
+		}
+	}
+
+	if (cq_entry->flags != 0)
+		efa_cntr_report_rx_completion(&base_ep->util_ep, cq_entry->flags);
```
shijin-aws, line 242: Same comments, IIRC cntr should always be updated no matter a cq entry is written or not. (Double check man page please)

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+				EFA_WARN(FI_LOG_EP_CTRL, "Unhandled op code %d\n", opcode);
+				assert(0 && "Unhandled op code");
+			}
+			break;
+		}
+
+		cq->read_entry(cq->ibv_cq.ibv_cq_ex, &cq_entry);
+
+		switch (opcode) {
+		case IBV_WC_SEND: /* fall through */
```
shijin-aws, line 359: Cntr should be always updated. So we can generally call cntr_report_completion with the `cq_entry->flags` here.

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+		case IBV_WC_RDMA_READ:
+			efa_cq_handle_send_completion(
+				base_ep, cq->ibv_cq.ibv_cq_ex, &cq_entry);
+			break;
+		case IBV_WC_RECV:
+			efa_cq_handle_recv_completion(
+				base_ep, cq->ibv_cq.ibv_cq_ex, &cq_entry);
+			break;
+		case IBV_WC_RECV_RDMA_WITH_IMM:
+			efa_cq_proc_ibv_recv_rdma_with_imm_completion(
```
shijin-aws, line 370: I think you can use a general function to cover all
`efa_cq_handle_send_completion`, `efa_cq_handle_send_completion`, and `efa_cq_proc_ibv_recv_rdma_with_imm_completion `, if you can use a simple logic to tell whether to write completion based on wr_id.

prov/efa/src/efa_cq.c
jiaxiyan: fabtests use FI_CQ_FORMAT_CONTEXT, which doesn't have flags(see `efa_cq_read_context_entry`)

prov/efa/src/efa_msg.c
```diff
@@ -214,7 +214,12 @@ static inline ssize_t efa_post_send(struct efa_base_ep *base_ep, const struct fi
 		base_ep->is_wr_started = true;
 	}
 
-	qp->ibv_qp_ex->wr_id = (uintptr_t)msg->context;
+	if (!(flags & FI_COMPLETION)) {
```
shijin-aws, line 217: prefer flipping the if else here for better read.

shijin-aws: Don't you need such if for other ops (recv, rma) as well

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+		return FI_REMOTE_CQ_DATA | FI_RMA | FI_REMOTE_WRITE;
+	case IBV_WC_RDMA_READ:
+		return FI_RMA | FI_READ;
+	default:
+		assert(0);
+		return 0;
+	}
+}
+
+static void efa_cq_read_context_entry(struct ibv_cq_ex *ibv_cqx, void *buf)
```
shijin-aws, line 34: I am not sure we have to distinguish the format here. We can use an entry that can read all the formats, like the `data_entry`. We will not need tagged format here as efa device doesn't support tagged

shijin-aws (changes requested):

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,535 @@
...
+static inline uint64_t efa_cq_opcode_to_fi_flags(enum ibv_wc_opcode opcode) {
+	switch (opcode) {
+	case IBV_WC_SEND:
+		return FI_SEND | FI_MSG;
+	case IBV_WC_RECV:
+		return FI_RECV | FI_MSG;
+	case IBV_WC_RDMA_WRITE:
+		return FI_RMA | FI_WRITE;
+	case IBV_WC_RECV_RDMA_WITH_IMM:
+		return FI_REMOTE_CQ_DATA | FI_RMA | FI_REMOTE_WRITE;
```
shijin-aws, line 25: If you have IMM for recv op code, do you need `FI_REMOTE_CQ_DATA` as well?

prov/efa/src/efa_cq.c
shijin-aws: that flags can be derived from op code, it doesn't depend on format.

prov/efa/src/efa_cq.c
shijin-aws: Think about it again, probably worth a special function form recv rdma with imm, as you will always write a cq entry for it.

prov/efa/src/efa_cq.c
shijin-aws: efa_rdm's ope used `tagged_entry` as `cq_entry` because it supports tagged

shijin-aws: Think about it again, using `tagged_entry` is good as you have used it in t he cq poll code, and if in the future efa device supports tagged, we don't need to change this format

prov/efa/src/efa_cq.c
jiaxiyan: I was referring to https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_cq.c#L531

prov/efa/src/efa_cq.c
jiaxiyan: Man page says `Counters record the number of requested operations that have completed.` Does that mean update cntr for no completion(null wr_id) too?

prov/efa/src/efa_cq.c
jiaxiyan: I can't find one API for this. Can we save the msg->addr to be part of fi_context?

prov/efa/src/efa_cq.c
jiaxiyan: The rdm code https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ope.c#L915-L924 returns without updating cntr if unable to write a cq entry

prov/efa/src/efa_cq.c
shijin-aws: Yeah, then for IBV_WC_RECV, when will you add FI_REMOTE_CQ_DATA?

prov/efa/src/efa_cq.c
shijin-aws: we can, but that doesn't work for inject calls where you don't have a fi_context.

prov/efa/src/efa_cq.c
shijin-aws: Yes, look at the rdm code, it update cntr for inject calls as well where we don't write cqe

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,437 @@
...
+	struct util_cq *cq = is_tx ? base_ep->util_ep.tx_cq : base_ep->util_ep.rx_cq;
+	fi_addr_t src_addr;
+	int ret = 0;
+
+	/* NULL wr_id means no completion */
+	if (!ibv_cq_ex->wr_id)
+		return;
+
+	if (!is_tx)
+		src_addr = efa_av_reverse_lookup(base_ep->av,
```
shijin-aws, line 139: we only need src_addr when FI_SOURCE

shijin-aws: Sorry about that, but now I think leaving tx/rx_handle_completion makes the code cleaner as you won't have too much `if` inside them

prov/efa/src/efa_cq.c
jiaxiyan: Looks like rdm does not update cntr if ofi_cq_write fails

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,437 @@
...
+	struct fi_cq_tagged_entry *entry = buf;
+
+	entry->op_context = (void *)ibv_cqx->wr_id;
+	entry->flags = efa_cq_opcode_to_fi_flags(ibv_wc_read_opcode(ibv_cqx));
+	entry->len = ibv_wc_read_byte_len(ibv_cqx);
+	entry->buf = NULL;
+	entry->data = 0;
+	entry->tag = 0;
+
+	if (ibv_wc_read_wc_flags(ibv_cqx) & IBV_WC_WITH_IMM)
```
shijin-aws, line 45: Do you need to | FI_REMOTE_CQ_DATA for flags here?

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,437 @@
...
+		return FI_REMOTE_CQ_DATA | FI_RMA | FI_REMOTE_WRITE;
+	case IBV_WC_RDMA_READ:
+		return FI_RMA | FI_READ;
+	default:
+		assert(0);
+		return 0;
+	}
+}
+
+static void efa_cq_read_tagged_entry(struct ibv_cq_ex *ibv_cqx, void *buf)
```
shijin-aws, line 34: How about renaming it a `efa_cq_construct_cq_entry`, and make `buf` as `fi_cq_tagged_entry *entry` in the argument directly?

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,478 @@
...
+			: efa_strerror(err_entry.prov_errno),
+		err_entry.prov_errno);
+
+	efa_show_help(err_entry.prov_errno);
+
+	efa_cntr_report_error(&base_ep->util_ep, err_entry.flags);
+	write_cq_err = ofi_cq_write_error(util_cq, &err_entry);
+	if (write_cq_err) {
+		EFA_WARN(FI_LOG_CQ,
+			"Error writing error cq entry when handling TX error\n");
```
shijin-aws, line 114: This function is called for both tx/rx error, right?

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,478 @@
...
+	err_entry.data = 0;
+	err_entry.tag = 0;
+
+	if (ibv_wc_read_wc_flags(ibv_cq_ex) & IBV_WC_WITH_IMM) {
+		err_entry.flags |= FI_REMOTE_CQ_DATA;
+		err_entry.data = ibv_wc_read_imm_data(ibv_cq_ex);
+	}
+
+	// TODO: get correct peer addr for tx
+	addr = efa_av_reverse_lookup(base_ep->av,
```
shijin-aws, line 92: Before we get it correct, we should make this addr retrieve only for rx operation. For tx operation, make it temporarily as `FI_ADDR_NOTAVAIL`

prov/efa/src/efa_cq.c
shijin-aws: That is a bug then

prov/efa/src/efa_cq.h
```diff
@@ -176,3 +184,107 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
...
+ * @param[in]    ep          EFA base endpoint
+ * @param[in]    addr        Remote peer fi_addr_t
+ * @param[in]    prov_errno  EFA provider * error code(must be positive)
+ * @param[out]   buf         Pointer to the address of error data written by
+ * this function
+ * @param[out]   buflen      Pointer to the returned error data size
+ * @return       A status code. 0 if the error data was written successfully,
+ * otherwise a negative FI error code.
+ */
+static inline int efa_write_error_msg(struct efa_base_ep *ep, fi_addr_t addr,
```
shijin-aws, line 241: I am trying to refactor this function path and make the error msg construction a separate function https://github.com/ofiwg/libfabric/pull/10657. Can you rebase your change and reuse that?

shijin-aws: Or you can make it general for efa-rdm and efa-raw so I will rebase on your change

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,478 @@
...
+				struct util_cq *util_cq, int err,
+				int prov_errno)
+{
+	struct fi_cq_err_entry err_entry;
+	fi_addr_t addr;
+	int write_cq_err;
+
+	memset(&err_entry, 0, sizeof(err_entry));
+
+	err_entry.err = err;
```
shijin-aws, line 78: This looks can reuse the cq_entry_construct function, as `fi_cq_err_entry` and `fi_cq_tagged_entry` are castable.

prov/efa/src/efa_cq.h
jiaxiyan: I am not sure how to do this cleanly since I can't get the peer from base_ep

shijin-aws: I see, no bother then

prov/efa/src/efa_cq.h
```diff
@@ -176,3 +184,111 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
...
+
+static inline
+bool efa_cq_wc_is_unsolicited(struct ibv_cq_ex *ibv_cq_ex)
+{
+	return false;
+}
+
+#endif
+
+static inline const char *efa_ep_raw_addr_str(struct efa_base_ep *base_ep, char *buf, size_t *buflen)
```
shijin-aws, line 217: this doesn't seem to be cq related, should be ep related header file, e.g. efa_base_ep.h

prov/efa/src/efa_cq.h
```diff
@@ -2,6 +2,8 @@
 /* SPDX-FileCopyrightText: Copyright Amazon.com, Inc. or its affiliates. All rights reserved. */
 
 #include "efa.h"
+#include "efa_av.h"
+#include "efa_cntr.h"
```
shijin-aws, line 6: why do you need cntr in this header file? If you need cntr in efa_cq.c, you should include it explicitly in it.

prov/efa/src/efa_cq.h
shijin-aws, line 217: prefer to break the line between `static inline` and `const`. Same for other places.

shijin-aws, line 217: This is duplicated to https://github.com/ofiwg/libfabric/blob/c3f9e2134bfc8524ac80a1306835bbfdc8b4dcb3/prov/efa/src/rdm/efa_rdm_ep_utils.c#L23-L31, please combine them

prov/efa/src/efa_cq.h
```diff
@@ -176,3 +184,111 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
...
+}
+
+#endif
+
+static inline const char *efa_ep_raw_addr_str(struct efa_base_ep *base_ep, char *buf, size_t *buflen)
+{
+	return ofi_straddr(buf, buflen, FI_ADDR_EFA, &base_ep->src_addr);
+}
+
+static inline const char *efa_ep_get_peer_raw_addr_str(struct efa_base_ep *base_ep, fi_addr_t addr, char *buf, size_t *buflen)
```
shijin-aws, line 222: Same, this is duplicate to https://github.com/ofiwg/libfabric/blob/c3f9e2134bfc8524ac80a1306835bbfdc8b4dcb3/prov/efa/src/rdm/efa_rdm_ep_utils.c#L43C21-L51, please combine them

prov/efa/src/efa_cq.h
```diff
@@ -176,3 +184,111 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
...
+{
+	struct efa_conn *efa_conn;
+	
+	if (OFI_UNLIKELY(addr == FI_ADDR_NOTAVAIL))
```
shijin-aws, line 226: efa_av_addr_to_conn already returns NULL for `FI_ADDR_UNSPEC` https://github.com/ofiwg/libfabric/blob/c3f9e2134bfc8524ac80a1306835bbfdc8b4dcb3/prov/efa/src/efa_av.c#L56-L57, you should make it return NULL for `FI_ADDR_NOTAVAIL` as well so you can drop this if condition here.

prov/efa/src/efa_cq.h
```diff
@@ -176,3 +184,111 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
...
+
+	if (ret < 0 || ret > EFA_ERROR_MSG_BUFFER_LENGTH - 1) {
+		return -FI_EINVAL;
+	}
+
+	if (strlen(err_msg) >= EFA_ERROR_MSG_BUFFER_LENGTH) {
+		return -FI_ENOBUFS;
+	}
+
+	*buf = err_msg;
```
shijin-aws, line 290: Are you sure this will work? Your `err_msg` will be released when the function returns as it is declared within the stack.

shijin-aws (changes requested):

prov/efa/src/efa_cq.h
jiaxiyan: I guess it might cause undefined behavior. I will move err_msg to be a field of base_ep then.

shijin-aws: Do you have to use a persistent `err_msg` in `base_ep` ? If you read the `ofi_cq_write_err` code, you will find it will `mem_dup` the `err_data` you filled in `entry`. In other words, the `err_data` you passed can be safely released after `ofi_cq_write_err` returns. Just use a temporary char array in your code block that write the error cq.

prov/efa/src/efa_base_ep.h
```diff
@@ -65,6 +66,7 @@ struct efa_base_ep {
 	/* Only used by RDM ep type */
 	struct efa_qp *user_recv_qp; /* Separate qp to receive pkts posted by users */
 	struct efa_recv_wr *user_recv_wr_vec;
+	char err_msg[EFA_ERROR_MSG_BUFFER_LENGTH]; /* A large enough buffer to store CQ/EQ error data used by e.g. fi_cq_readerr */
```
shijin-aws, line 69: Paste my earlier comment here again

Do you have to use a persistent `err_msg` in `base_ep` ? If you read the `ofi_cq_write_err` code, you will find it will `mem_dup` the `err_data` you filled in `entry`. In other words, the `err_data` you passed can be safely released after `ofi_cq_write_err` returns. Just use a temporary char array in your code block that write the error cq.

jiaxiyan: Do you prefer to remove err_msg from rdm_ep too?

shijin-aws: yes

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,466 @@
...
+		ret = ofi_cq_write(rx_cq, NULL, flags, len, NULL, imm_data, 0);
+	}
+
+	if (OFI_UNLIKELY(ret)) {
+		EFA_WARN(FI_LOG_CQ,
+			 "Unable to write a cq entry for remote for RECV_RDMA "
+			 "operation: %s\n",
+			 fi_strerror(-ret));
+		efa_base_ep_write_eq_error(base_ep, -ret,
+					   FI_EFA_ERR_WRITE_SHM_CQ_ENTRY);
```
shijin-aws, line 232: This error code doesn't seem correct :)

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -348,11 +348,11 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 	 * QP and we cannot cancel that.
 	 */
 	if (OFI_UNLIKELY(ep->use_zcpy_rx && efa_rdm_pkt_type_is_rtm(pkt_type))) {
-		void *errbuf;
+		void *errbuf = {0};
```
shijin-aws, line 351: Shouldn't it be a new `err_msg[EFA_ERROR_MSG_BUFFER_LENGTH]` here?

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -695,8 +699,10 @@ void efa_rdm_txe_handle_error(struct efa_rdm_ope *txe, int err, int prov_errno)
 	err_entry.data = txe->cq_entry.data;
 	err_entry.tag = txe->cq_entry.tag;
 	if (OFI_UNLIKELY(efa_rdm_write_error_msg(ep, txe->addr, prov_errno,
-	                                         &err_entry.err_data, &err_entry.err_data_size))) {
+	                                         err_msg, &err_entry.err_data_size))) {
 		err_entry.err_data_size = 0;
+	} else {
```
shijin-aws, line 704: Better to `err_entry.err_data = err_msg;` before you call `efa_rdm_write_error_msg`, so you have a valid `err_data` for ofi_cq_write_error anyway.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -603,8 +604,10 @@ void efa_rdm_rxe_handle_error(struct efa_rdm_ope *rxe, int err, int prov_errno)
 	err_entry.data = rxe->cq_entry.data;
 	err_entry.tag = rxe->cq_entry.tag;
 	if (OFI_UNLIKELY(efa_rdm_write_error_msg(ep, rxe->addr, prov_errno,
-	                                         &err_entry.err_data, &err_entry.err_data_size))) {
+	                                         err_msg, &err_entry.err_data_size))) {
 		err_entry.err_data_size = 0;
+	} else {
```
shijin-aws, line 609: same comment as the `txe` one.

prov/efa/src/rdm/efa_rdm_ope.c
shijin-aws: Sorry, I think your current one may be better, just see  your err_entry.err_data is already NULL if the `efa_rdm_write_error_msg` returns non-zero.

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,470 @@
...
+
+	if (attr->wait_obj != FI_WAIT_NONE)
+		return -FI_ENOSYS;
+
+	cq = calloc(1, sizeof(*cq));
+	if (!cq)
+		return -FI_ENOMEM;
+
+	err = ofi_cq_init(&efa_prov, domain_fid, attr, &cq->util_cq,
+			  &efa_cq_progress, context);
```
shijin-aws, line 438: You need a similar change for efa_cntr: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_cntr.c#L200-L202. Since you make dgram_ep_progress as a no op, `ofi_cntr_progress` will be a no-op too as it call ep progress in the same way as `ofi_cq_progress` does. You need a `efa_cntr_progress` that polls ibv_cq as well

prov/efa/src/efa_cq.c
```diff
@@ -0,0 +1,470 @@
...
+		return;
+
+	/* TX completions should not send peer address to util_cq */
+	if (base_ep->util_ep.caps & FI_SOURCE)
+		ret = ofi_cq_write_src(tx_cq, cq_entry->op_context,
+				       cq_entry->flags, cq_entry->len,
+				       cq_entry->buf, cq_entry->data,
+				       cq_entry->tag, FI_ADDR_NOTAVAIL);
+	else
+		ret = ofi_cq_write(tx_cq, cq_entry->op_context, cq_entry->flags,
```
sunkuamzn, line 146: Looks like there are still two copies one after the other in the CQ read path. Can we eliminate one of the copies?

The CQ read path is (correct me if I'm wrong)
* Application calls `fi_cq_read` which we set to `ofi_cq_read`
*  `ofi_cq_read` -> `ofi_cq_readfrom`
    * `ofi_cq_readfrom` first calls `cq->progress(cq)` which we set to `efa_cq_progress` -> `efa_cq_poll_ibv_cq` -> poll IBV CQ -> `ofi_cq_write` and we copy the IBV CQ completion into the Libfabric CQ circular queue
    * `ofi_cq_readfrom` then calls `ofi_cq_read_entries` and we copy the CQ entry from Libfabric CQ circular queue to the application CQ entry buffer

If the progress was more complicated (progress other CQs or EPs for protocol reasons), then we need the circular queue. But in this case, we really don't.

prov/efa/src/efa_msg.c
```diff
@@ -97,9 +97,9 @@ static inline ssize_t efa_post_recv(struct efa_base_ep *base_ep, const struct fi
 	}
 
 	wr = &base_ep->efa_recv_wr_vec[wr_index].wr;
-	wr->wr_id = (uintptr_t)msg->context;
 	wr->num_sge = msg->iov_count;
 	wr->sg_list = base_ep->efa_recv_wr_vec[wr_index].sge;
+	wr->wr_id = (uintptr_t) ((flags & FI_COMPLETION) ? msg->context : NULL);
```
sunkuamzn, line 102: I discussed with Jessie. According to the docs for FI_CONTEXT

>This structure should be treated as opaque to the application. For performance reasons, this structure must be allocated by the user, but may be used by the fabric provider to track the operation.

So we can use this buffer however we want. I would suggest setting it up as a bitmask and have FI_COMPLETION be one of the bits. That way, we can use the other bits for something else in the future.

prov/efa/src/efa_av.c
```diff
@@ -53,7 +53,7 @@ struct efa_conn *efa_av_addr_to_conn(struct efa_av *av, fi_addr_t fi_addr)
 	struct util_av_entry *util_av_entry;
 	struct efa_av_entry *efa_av_entry;
 
-	if (OFI_UNLIKELY(fi_addr == FI_ADDR_UNSPEC))
+	if (OFI_UNLIKELY(fi_addr == FI_ADDR_UNSPEC || fi_addr == FI_ADDR_NOTAVAIL))
```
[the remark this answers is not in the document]
jiaxiyan: This function is called in`efa_base_ep_get_peer_raw_addr`. We temporarily return FI_ADDR_NOTAVAIL as peer addr for TX operation in `efa_cq_handle_error`.

prov/efa/src/efa_cq.c
sunkuamzn: Application calls `fi_cq_read` with a `count`. We can stop polling after we get `count` completions. rdma-core is already buffering the CQ entries, so we can just take advantage of that.

shijin-aws: >If the progress was more complicated (progress other CQs or EPs for protocol reasons), then we need the circular queue. But in this case, we really don't.

I don't think it matters with how complicated the protocol is. For protocol case, we only write ofi cq when an operation has completed, which may involve multiple ibv cqes. You can always customize a function that propagate the completion to the `buf` in the fi_cq_read directly to achieve best performance.

shijin-aws: It seems the current dgram code was kind of designed in this direction, see https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/dgram/efa_dgram_cq.c#L161. But then it rewrite the cq entries in `buf` back to ofi cq https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/dgram/efa_dgram_ep.c#L319-L338. We may reuse the dgram code pattern but remove that `ofi_cq_write*` phase.

shijin-aws: But then we still need to handle the error cq write, which is doesn't have a 1-to-1 relationship with the `buf`... I prefer we have a careful design on this if we really want to do that.

shijin-aws: One problem I think of is that, if you want to get rid of using libfabric's cirque, there is no way for you to make a progress engine while staging the cqes. IIRC `fi_cntr_read` needs to progress the ep for manual progress model, and that progress function needs to poll ibv cq to update the cntr while staging the completions for another `fi_cq_read` to read.

shijin-aws: See https://ofiwg.github.io/libfabric/main/man/fi_domain.3.html

```
FI_PROGRESS_MANUAL
    This progress model indicates that the provider requires the use of an application thread to complete an asynchronous request. When manual progress is set, the provider will attempt to advance an asynchronous operation forward when the application attempts to wait on or read an event queue, completion queue, or counter where the completed operation will be reported. Progress also occurs when the application processes a poll or wait set that has been associated with the event or completion queue. 
```

We need to support the case where application only use `fi_cntr_read` to progress the operations.

prov/efa/src/efa_cntr.c
```diff
@@ -178,6 +178,24 @@ static void efa_rdm_cntr_progress(struct util_cntr *cntr)
...
 }
 
+static void efa_cntr_progress(struct util_cntr *cntr)
+{
+	struct util_ep *ep;
+	struct fid_list_entry *fid_entry;
+	struct dlist_entry *item;
+
+	ofi_genlock_lock(&cntr->ep_list_lock);
+	dlist_foreach(&cntr->ep_list, item) {
```
shijin-aws, line 188: We should avoid iterating `ep_list`, but iterating the ibv cq list as the rdm code path does. This is a pain for cntr as it doesn't have a 1-to-1 relationship with ibv cq.

shijin-aws: But wait, you cannot bind 1 cntr to multiple eps, so that ep_list is just a 1 element array. Then we don't need to bother changing the logic here.

prov/efa/src/efa_cntr.c
```diff
@@ -178,6 +178,24 @@ static void efa_rdm_cntr_progress(struct util_cntr *cntr)
...
+		fid_entry = container_of(item, struct fid_list_entry, entry);
+		ep = container_of(fid_entry->fid, struct util_ep, ep_fid.fid);
+		if (ep->tx_cq)
+			efa_cq_progress(ep->tx_cq);
+		if (ep->rx_cq)
```
shijin-aws, line 193: If tx and rx cq are the same, we don't need to progress both of them

shijin-aws (approved):
Tentatively approve this PR. I think @sunkuamzn 's suggestion on writing completions to user buffer directly is reasonable. Though we need a plan to handle the counter case, I suggest refactoring the code a little bit to restore the earlier version that has separate `cq->read_entry` function table to handle writing cq buffer with different cq format. Then it will give us flexibility to move to the direct write path in the future when we have a plan.

prov/efa/src/efa_cntr.c
shijin-aws: Ignore the last comment, you can bind 1 cntr to multiple eps and what you read will be the aggregation of completions from multiple eps.

PR #10660, opened 2024-12-27
fabtests: Bugfixes for neuron

sunkuamzn:
This commit fixes the following bugs in neuron fabtests
1. The neuron accelerator detection is broken on some OSs because the full path of the executable `neuron-ls` was not used

2. Before this commit, each pytest worker was assigned a single  neuron core. This works on multi node tests but fails on single node tests because a neuron core can only be opened by a single process. This commit assigns two different neuron cores to each pytest worker for client-server tests: one for the server and one for the client. Trn1 has 2 cores per neuron device and Trn2 has 8 cores per neuron device, so
  this assignment works for both.

3. When running in serial mode, the env var PYTEST_XDIST_WORKER is not set, so the NEURON_RT_VISIBLE_CORES env var is also not set. This causes the server to occupy all neuron cores and the client fails. So this commit assigns device 0 to the server and client when running with one worker.

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -68,7 +68,7 @@ def num_cuda_devices(ip):
 @functools.lru_cache(10)
 @retry(retry_on_exception=is_ssh_connection_error, stop_max_attempt_number=3, wait_fixed=5000)
 def num_neuron_devices(ip):
-    proc = run("ssh {} neuron-ls -j".format(ip), shell=True,
+    proc = run("ssh {} /opt/aws/neuron/bin/neuron-ls -j".format(ip), shell=True,
                stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                timeout=60, encoding="utf-8")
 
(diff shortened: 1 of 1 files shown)


jiaxiyan (approved):

fabtests/pytest/common.py
```diff
@@ -455,19 +455,26 @@ def prepare_base_command(self, command_type, executable,
...
+        else:
+            hmem_device_id = 0
+
+        if host_memory_type == "cuda":
+            command += " -i {}".format(hmem_device_id)
+        else:
+            assert host_memory_type == "neuron"
+            num_cores = num_neuron_cores_on_device(host_ip, hmem_device_id)
+            if command_type == "server":
                 additional_environment = "NEURON_RT_VISIBLE_CORES={}".format(
```
shijin-aws, line 467: I remember you can specify multiple cores for single process. Does it you work if you make both client and server use something like `NEURON_RT_VISIBLE_CORES=hmem_device_id * num_cores ,hmem_device_id * num_cores + 1`?

sunkuamzn: No. The server takes both cores.

shijin-aws (approved):

shijin-aws:
need to backport to v2.0.x and v1.22.x

PR #10663, opened 2024-12-30
[v2.0.x] fabtests: Bugfixes for neuron

sunkuamzn:
This commit fixes the following bugs in neuron fabtests
1. The neuron accelerator detection is broken on some OSs because the full path of the executable `neuron-ls` was not used

2. Before this commit, each pytest worker was assigned a single  neuron core. This works on multi node tests but fails on single node tests because a neuron core can only be opened by a single process. This commit assigns two different neuron cores to each pytest worker for client-server tests: one for the server and one for the client. Trn1 has 2 cores per neuron device and Trn2 has 8 cores per neuron device, so
  this assignment works for both.

3. When running in serial mode, the env var PYTEST_XDIST_WORKER is not set, so the NEURON_RT_VISIBLE_CORES env var is also not set. This causes the server to occupy all neuron cores and the client fails. So this commit assigns device 0 to the server and client when running with one worker.

(cherry picked from commit f893f5f88eb5cf0f4cf4e0154edc91e0b6f8b7bf)

### DIFF
--- a/fabtests/pytest/common.py
+++ b/fabtests/pytest/common.py
@@ -68,7 +68,7 @@ def num_cuda_devices(ip):
 @functools.lru_cache(10)
 @retry(retry_on_exception=is_ssh_connection_error, stop_max_attempt_number=3, wait_fixed=5000)
 def num_neuron_devices(ip):
-    proc = run("ssh {} neuron-ls -j".format(ip), shell=True,
+    proc = run("ssh {} /opt/aws/neuron/bin/neuron-ls -j".format(ip), shell=True,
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

sunkuamzn:
@j-xiong @aingerson could you please check if the Intel CI failure is relevant for this PR and https://github.com/ofiwg/libfabric/pull/10662?

j-xiong:
The failures were with oneCCL tests over tcp. I think they are unrelated to this PR.

PR #10671, opened 2025-01-02
prov/cxi: Allow fi_close() to complete without error

bulbul-hpe:
Modified libfabric behavior to ensure that fi_close() calls on an endpoint complete successfully, even if an MCAST tree is still allocated. This change prevents errors in cases where the endpoint is being closed while retaining the MCAST tree.

### DIFF
--- a/prov/util/src/util_buf.c
+++ b/prov/util/src/util_buf.c
@@ -282,12 +282,18 @@ void ofi_bufpool_destroy(struct ofi_bufpool *pool)
 {
 	struct ofi_bufpool_region *buf_region;
 	size_t i;
+	int count;
 
 	for (i = 0; i < pool->region_cnt; i++) {
 		buf_region = pool->region_table[i];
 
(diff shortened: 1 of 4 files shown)


prov/util/src/util_buf.c
```diff
@@ -282,12 +282,18 @@ void ofi_bufpool_destroy(struct ofi_bufpool *pool)
...
+	int count;
 
 	for (i = 0; i < pool->region_cnt; i++) {
 		buf_region = pool->region_table[i];
 
-		assert((pool->attr.flags & OFI_BUFPOOL_NO_TRACK) ||
-			!ofi_atomic_get32(&buf_region->use_cnt));
+		if (!(pool->attr.flags & OFI_BUFPOOL_NO_TRACK)) {
+			count = ofi_atomic_get32(&buf_region->use_cnt);
+			if (count) {
```
shijin-aws, line 292: If you want to suppress the error, shouldn't you create the buffer pool with `OFI_BUFPOOL_NO_TRACK`, or manually warn and clean them in your ep close?

PR #10679, opened 2025-01-06
prov/efa: Remove inline write logic for efa_rma inject

jiaxiyan:
We previously set the inject size to 0 to prevent using inline write, but application can use inject rma for 0 size message. Make rma inject temporarily return FI_ENOSYS before firmware supports inline write.

### DIFF
--- a/prov/efa/src/efa_rma.c
+++ b/prov/efa/src/efa_rma.c
@@ -200,7 +200,6 @@ static inline ssize_t efa_rma_post_write(struct efa_base_ep *base_ep,
 	struct efa_conn *conn;
 #ifndef _WIN32
 	struct ibv_sge sge_list[msg->iov_count];
-	struct ibv_data_buf inline_data_list[msg->iov_count];
 #else
 	/* MSVC compiler does not support array declarations with runtime size, so hardcode
 	 * the expected iov_limit/max_sq_sge from the lower-level efa provider.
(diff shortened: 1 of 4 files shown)


prov/efa/src/efa_rma.c
```diff
@@ -200,15 +200,13 @@ static inline ssize_t efa_rma_post_write(struct efa_base_ep *base_ep,
...
-	struct ibv_data_buf inline_data_list[msg->iov_count];
 #else
 	/* MSVC compiler does not support array declarations with runtime size, so hardcode
 	 * the expected iov_limit/max_sq_sge from the lower-level efa provider.
 	 */
 	struct ibv_sge sge_list[EFA_DEV_ATTR_MAX_WR_SGE];
 	struct ibv_data_buf inline_data_list[EFA_DEV_ATTR_MAX_WR_SGE];
 #endif
-	size_t len;
 	int i, err = 0;
```
shijin-aws, line 211: You also need to raise error when `flags` has `FI_INJECT`

shijin-aws: Also add a unit test for it.

shijin-aws (dismissed):

shijin-aws:
need to resolve the merge conflict

shijin-aws (approved):

PR #10690, opened 2025-01-08
prov/cxi: cxi EQ do not support wait objects

swelch:
Make sure wait objects are not allowed for cxi EQs.


swelch:
appveyor/pr debug version failure appears unrelated.

iziemba (approved):

PR #10697, opened 2025-01-13
prov/efa: Make efa_rdm_cq use efa_cq

shijin-aws:
The structs efa_rdm_cq and efa_cq share the
same members util_cq and ibv_cq. This patch makes efa_rdm_cq use efa_cq as a subset so we can convert between efa_rdm_cq and efa_cq via container_of, like efa_rdm_ep and efa_base_ep.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_fiops.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_fiops.c
@@ -37,13 +37,13 @@ void efa_rdm_ep_construct_ibv_qp_init_attr_ex(struct efa_rdm_ep *ep,
 static inline
 struct efa_rdm_cq *efa_rdm_ep_get_tx_rdm_cq(struct efa_rdm_ep *ep)
 {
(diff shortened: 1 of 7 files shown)


sunkuamzn (approved):

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -37,13 +37,13 @@ void efa_rdm_ep_construct_ibv_qp_init_attr_ex(struct efa_rdm_ep *ep,
 static inline
 struct efa_rdm_cq *efa_rdm_ep_get_tx_rdm_cq(struct efa_rdm_ep *ep)
```
jiaxiyan, line 38: Do you still need these two functions?

shijin-aws: I thought about removing them, but then I need to check whether `tx_cq` / `rx_cq` is NULL before I can derive `tx_rdm_cq` and `rx_rdm_cq` via `container_of`.. so I tend to keep these two functions

PR #10702, opened 2025-01-15
prov/efa: Move struct efa_ep_addr to efa_base_ep

a-szegel:
struct efa_ep_addr is used in both the DGRAM and RDM providers, so the structure definition should be in a common file, and not an RDM only file.

### DIFF
--- a/prov/efa/src/efa_base_ep.h
+++ b/prov/efa/src/efa_base_ep.h
@@ -27,6 +27,18 @@ struct efa_qp {
 	uint32_t qkey;
 };
 
+#define EFA_GID_LEN	16
+
+struct efa_ep_addr {
+	uint8_t			raw[EFA_GID_LEN];
+	uint16_t		qpn;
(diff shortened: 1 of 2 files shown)


a-szegel:
I want bigger changes than this, closing.

a-szegel:
reopening, I think I do actually want this... baby steps

prov/efa/src/efa_base_ep.h
```diff
@@ -27,6 +27,19 @@ struct efa_qp {
 	uint32_t qkey;
 };
 
+/* raw address format. (section 1.4) */
```
shijin-aws, line 30: since you moved the code, you need to update the comment as now reader doesn't related `section 1.4` to the protocol doc.

a-szegel: removed the outdated comment.

shijin-aws (approved):

PR #10706, opened 2025-01-16
prov/efa: Make DGRAM provider use new av_entry struct

a-szegel:
Update the DGRAM provider to use the new efa_base_av_entry structure. This change splits the AV logic for the DGRAM and the RDM providers. Splitting the av logic for DGRAM makes DGRAM's AV logic simpler and faster in a few ways.

1. This change removes the efa_con struct from the av_entry because the DGRAM protocol does not maintain an idea of a connection to the endpoint receiving the messages. It does not matter if the remote endpoint we were previously communicating with exits, and a new QP opens with the same AH (NIC), and QPN (Endpoint). As long as someone is there to post recv buffers for our messages, DGRAM provider is happy.

2. This changes gets rid of the current/prev av reverse lookup map (ahn/qpn -> av_entry), and creates a single reverse look up map because the dgram provider does not care about old connections.

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -65,6 +65,29 @@ struct efa_conn *efa_av_addr_to_conn(struct efa_av *av, fi_addr_t fi_addr)
 	return efa_av_entry->conn.ep_addr ? &efa_av_entry->conn : NULL;
 }
 
+/*
+ * @brief find efa_base_av_entry struct using fi_addr
+ *
+ * @param[in]	av	efa av
+ * @param[in]	fi_addr	index into av
+ * @return	if address is valid, return pointer to efa_base_av_entry struct
+ * 		otherwise, return NULL
+ */
+struct efa_base_av_entry *efa_av_addr_to_base_av_entry(struct efa_av *av, fi_addr_t fi_addr)
+{
+	struct util_av_entry *util_av_entry;
+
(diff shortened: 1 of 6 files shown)


a-szegel:
This needs unit testing, but it should be ready to review.

shijin-aws:
A high level comment is that I want to get rid of `dgram`. I prefer to use `efa_av` prefix for the code path that doesn't involve the rdm protocols. And use `efa_rdm_av` prefix for the protocol path.

prov/efa/src/efa_av.c
```diff
@@ -76,15 +99,15 @@ struct efa_conn *efa_av_addr_to_conn(struct efa_av *av, fi_addr_t fi_addr)
  */
 fi_addr_t efa_av_reverse_lookup(struct efa_av *av, uint16_t ahn, uint16_t qpn)
```
shijin-aws, line 100: This is already the function used for dgram code path. rdm code path uses `efa_av_reserve_lookup_rdm`.

shijin-aws:
I am wondering if it is cleaner to decouple rdm_peer with efa_conn. As far as I read the code you only access peer from coon during the av insert and remove path. You don't need rdm peer for reverse look up. It should be much less changes if you do in this way

sunkuamzn:
Closing for now. Will re-open later.

PR #10707, opened 2025-01-17
prov/efa: Implement FI_CONTEXT2 in EFA Direct

jiaxiyan:
Store the completion flags and peer address in FI_CONTEXT2 and retrieve later when writing cq.

### DIFF
--- a/prov/efa/src/efa.h
+++ b/prov/efa/src/efa.h
@@ -107,6 +107,41 @@ struct efa_fabric {
 #endif
 };
 
+struct efa_context {
+	uint64_t completion_flags;
+	fi_addr_t addr;
+};
+
+#if defined(static_assert)
+static_assert(sizeof(struct efa_context) <= sizeof(struct fi_context2),
+	      "efa_context must not be larger than fi_context2");
+#endif
+
+/**
+ * Prepare and return a pointer to an EFA context structure.
+ *
+ * @param context           Pointer to the msg context.
+ * @param addr              Peer address associated with the operation.
+ * @param flags             Operation flags (e.g., FI_COMPLETION).
+ * @param completion_flags  Completion flags reported in the cq entry.
+ * @return A pointer to an initialized EFA context structure,
+ *  or NULL if context is invalid or FI_COMPLETION is not set.
+ */
+static inline struct efa_context *efa_fill_context(const void *context,
+						   fi_addr_t addr,
+						   uint64_t flags,
+						   uint64_t completion_flags)
+{
+	if (!context || !(flags & FI_COMPLETION))
+		return NULL;
+
+	struct efa_context *efa_context = (struct efa_context *) context;
+	efa_context->completion_flags = completion_flags;
+	efa_context->addr = addr;
+
+	return efa_context;
+}
+
 static inline
 int efa_str_to_ep_addr(const char *node, const char *service, struct efa_ep_addr *addr)
 {
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -36,7 +36,10 @@ static void efa_cq_construct_cq_entry(struct ibv_cq_ex *ibv_cqx,
 				      struct fi_cq_tagged_entry *entry)
 {
 	entry->op_context = (void *)ibv_cqx->wr_id;
-	entry->flags = efa_cq_opcode_to_fi_flags(ibv_wc_read_opcode(ibv_cqx));
+	if (ibv_cqx->wr_id)
+		entry->flags = ((struct efa_context *) ibv_cqx->wr_id)->completion_flags;
+	else
+		entry->flags = efa_cq_opcode_to_fi_flags(ibv_wc_read_opcode(ibv_cqx));
 	entry->len = ibv_wc_read_byte_len(ibv_cqx);
 	entry->buf = NULL;
 	entry->data = 0;
@@ -81,8 +84,7 @@ static void efa_cq_handle_error(struct efa_base_ep *base_ep,
 	err_entry.prov_errno = prov_errno;
 
 	if (is_tx)
-		// TODO: get correct peer addr for TX operation
-		addr = FI_ADDR_NOTAVAIL;
+		addr = ibv_cq_ex->wr_id ? ((struct efa_context *)ibv_cq_ex->wr_id)->addr : FI_ADDR_NOTAVAIL;
 	else
 		addr = efa_av_reverse_lookup(base_ep->av,
 					     ibv_wc_read_slid(ibv_cq_ex),
--- a/prov/efa/src/efa_rma.c
+++ b/prov/efa/src/efa_rma.c
@@ -90,7 +90,9 @@ static inline ssize_t efa_rma_post_read(struct efa_base_ep *base_ep,
 		ibv_wr_start(qp->ibv_qp_ex);
 		base_ep->is_wr_started = true;
 	}
-	qp->ibv_qp_ex->wr_id = (uintptr_t) ((flags & FI_COMPLETION) ? msg->context : NULL);
+
+	qp->ibv_qp_ex->wr_id = (uintptr_t) efa_fill_context(
+		msg->context, msg->addr, flags, FI_RMA | FI_READ);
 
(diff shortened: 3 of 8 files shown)


prov/efa/src/efa.h
```diff
@@ -107,6 +107,41 @@ struct efa_fabric {
 #endif
 };
 
+struct efa_context {
+	uint64_t completion_flags;
+	fi_addr_t addr;
+};
+
+#if defined(static_assert) && defined(__x86_64__)
```
a-szegel, line 115: Why do you have the x86 restriction here?

jiaxiyan: I see all static_assert in efa provider has this restriction so I add it too

a-szegel: This program on hpc7g:
```
#include <assert.h>

struct efa_context {
	int a;
	int b;
};

static_assert(sizeof(struct efa_context) <= 1 * sizeof(int),
	      "Should Fail");

int main() {
	return 0;
}
```
fails to compile with:
```
gcc main.c 
In file included from main.c:1:
main.c:8:1: error: static assertion failed: "Should Fail"
    8 | static_assert(sizeof(struct efa_context) <= 1 * sizeof(int),
      | ^~~~~~~~~~~~~
```

and this program
```
#include <assert.h>

struct efa_context {
	int a;
	int b;
};

static_assert(sizeof(struct efa_context) <= 2 * sizeof(int),
	      "Should Pass");

int main() {
	return 0;
}
```
passes and runs fine on hpc7g
```
> gcc main.c 
> ./a.out 
>
```

I am using ub24.04 with gcc `gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0`... maybe it is due to older OS's and gcc versions?

bwbarrett: There's no reason to have those checks only for x86.  I think there was a hack not understanding the issues previously that got copied around and should be fixed.

jiaxiyan: Ok I will fix them

shijin-aws: >I am using ub24.04 with gcc gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0... maybe it is due to older OS's and gcc versions?

static_assert is only available since C11, some old gcc doesn't support it.

prov/efa/src/efa_rma.c
```diff
@@ -225,7 +227,12 @@ static inline ssize_t efa_rma_post_write(struct efa_base_ep *base_ep,
 		ibv_wr_start(qp->ibv_qp_ex);
 		base_ep->is_wr_started = true;
 	}
-	qp->ibv_qp_ex->wr_id = (uintptr_t) ((flags & FI_COMPLETION) ? msg->context : NULL);
+
+	qp->ibv_qp_ex->wr_id = (uintptr_t) efa_fill_context(
+		msg->context, msg->addr, flags,
+		flags & FI_REMOTE_CQ_DATA ?
+			FI_REMOTE_CQ_DATA | FI_RMA | FI_REMOTE_WRITE :
```
shijin-aws, line 234: This is not right, `FI_REMOTE_WRITE` for the target of the fi_writedata.

`FI_REMOTE_CQ_DATA` as a completion flag is also for the target of `fi_*data` operation, You shouldn't have `FI_REMOTE_CQ_DATA` in the initiator of a `data` operation.

For this change, you don't need any check on `flags` for your completion flags, it should just be `FI_RMA | FI_WRITE`

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -974,16 +985,19 @@ void test_efa_cq_read_send_failure(struct efa_resource **state)
 	struct efa_resource *resource = *state;
 	struct efa_unit_test_buff send_buff;
 	struct fi_cq_data_entry cq_entry;
+	struct fi_context2 *ctx;
 	fi_addr_t addr;
 	int ret;
 
+	ctx = malloc(sizeof(struct fi_context2));
```
shijin-aws, line 992: why not just `struct fi_context 2 ctx`.... then you don't need to free it explicitly

shijin-aws (changes requested):

prov/efa/test/efa_unit_test_cq.c
jiaxiyan: I see the man page says fi_context should not be allocated on the stack.

bwbarrett: You can define a context on the stack, you just have to guarantee that the lifetime of the stack variable is at least as long as the time from starting the transaction to getting the completion event.  In this particular test, that would be true.  But it's really easy to accidentally screw up variable lifespans on the stack and those cases are impossible to debug, so the man page is a bit overly conservative.  Outside of tests, though, I'd be pretty strongly against putting a context on the stack, so I don't disagree with the manpage.

shijin-aws: ++1 for what Brian said. stack/non-stack suggestion is based on the lifecycle of the context. Your test is only using the context locally in the function so it doesn't require a malloc.

prov/efa/src/efa_cq.c
```diff
@@ -36,7 +36,10 @@ static void efa_cq_construct_cq_entry(struct ibv_cq_ex *ibv_cqx,
 				      struct fi_cq_tagged_entry *entry)
 {
 	entry->op_context = (void *)ibv_cqx->wr_id;
-	entry->flags = efa_cq_opcode_to_fi_flags(ibv_wc_read_opcode(ibv_cqx));
+	if (!ibv_cqx->wr_id)
```
shijin-aws, line 39: Nit: why not use `if(ibv_cqx->wr_id)` and flip the cases

shijin-aws (approved):

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -974,16 +981,17 @@ void test_efa_cq_read_send_failure(struct efa_resource **state)
...
 	test_efa_cq_read(resource, &addr, IBV_WC_SEND, IBV_WC_GENERAL_ERR,
-			 EFA_IO_COMP_STATUS_LOCAL_ERROR_UNRESP_REMOTE);
+			 EFA_IO_COMP_STATUS_LOCAL_ERROR_UNRESP_REMOTE, (struct efa_context *) &ctx);
 	efa_unit_test_buff_construct(&send_buff, resource, 4096 /* buff_size */);
 
 	assert_int_equal(g_ibv_submitted_wr_id_cnt, 0);
 	ret = fi_send(resource->ep, send_buff.buff, send_buff.size,
-		      fi_mr_desc(send_buff.mr), addr, (void *) 12345);
+		      fi_mr_desc(send_buff.mr), addr, &ctx);
 	assert_int_equal(ret, 0);
```
shijin-aws, line 995: Can you add test cases (asserts) here to check whether your ctx is assigned correctly? similar to other tests

prov/efa/test/efa_unit_test_cq.c
```diff
@@ -894,16 +897,18 @@ void test_efa_cq_read_send_success(struct efa_resource **state)
...
+	test_efa_cq_read(resource, &addr, IBV_WC_SEND, IBV_WC_SUCCESS, 0,
+			 (struct efa_context *) &ctx);
 	efa_unit_test_buff_construct(&send_buff, resource, 4096 /* buff_size */);
 
 	assert_int_equal(g_ibv_submitted_wr_id_cnt, 0);
 	ret = fi_send(resource->ep, send_buff.buff, send_buff.size,
-		      fi_mr_desc(send_buff.mr), addr, (void *) 12345);
+		      fi_mr_desc(send_buff.mr), addr, &ctx);
 	assert_int_equal(ret, 0);
 	assert_int_equal(g_ibv_submitted_wr_id_cnt, 1);
```
shijin-aws, line 914: Not related to this PR, but generally it worth checking the completion flags in the cq entry to be correct.

sunkuamzn (approved):

shijin-aws (approved):

shijin-aws:
bot:aws:retest

PR #10708, opened 2025-01-17
prov/efa: fix leak of dmabuf fd in cuda p2p probe

aws-nslick:
prior to this patch, when efa_hmem_info_check_p2p_support_cuda elected to attempt dmabuf for p2p, we previously leaked the file descriptor returned by cuMemGetHandleForAddressRange in all cases. This ultimately meant the dmabuf stuck around for the lifetime of the process, even after dereg and after releasing the memory back to the device mempool.

All calls to cuda_get_dmabuf_fd need a corresponding close call.

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -129,6 +129,7 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
 	if (ret == FI_SUCCESS) {
 		ibv_mr = ibv_reg_dmabuf_mr(g_device_list[0].ibv_pd, dmabuf_offset,
 					   len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+		(void)close(dmabuf_fd);
 		if (!ibv_mr) {
 			EFA_INFO(FI_LOG_CORE,
 				"Unable to register CUDA device buffer via dmabuf: %s. "


prov/efa/src/efa_hmem.c
```diff
@@ -129,6 +129,7 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
 	if (ret == FI_SUCCESS) {
 		ibv_mr = ibv_reg_dmabuf_mr(g_device_list[0].ibv_pd, dmabuf_offset,
 					   len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+		(void)close(dmabuf_fd);
```
shijin-aws, line 132: I was told by Nvidia folks at that time that I don't need to close the fd explicitly if I free the cuda buffer. But I am not against doing it if it's safer.

shijin-aws (approved):

prov/efa/src/efa_hmem.c
aws-nslick: I don't think it's a matter of one being safer than the other -- it's currently bugged. 

it's trivial to confirm that what that you're suggesting isn't happening:

```
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ sudo cat /sys/kernel/debug/dma_buf/bufinfo | grep -B1 'Attached' | grep -vi attached | sed 's/--//g' | sort -u | awk '{ printf "%s,%s,%s,%s\n", $1, $2, $3, $4 }' | sort -u
,,,
00524288,00000002,00080007,00000005
04194304,00000002,00080007,00000005
04915200,00000002,00080007,00000005
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ # with patch ^
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ # revert patch
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ sudo cat /sys/kernel/debug/dma_buf/bufinfo | grep -B1 'Attached' | grep -vi attached | sed 's/--//g' | sort -u | awk '{ printf "%s,%s,%s,%s\n", $1, $2, $3, $4 }' | sort -u
,,,
00008192,00000002,00080007,00000001
00524288,00000002,00080007,00000005
04194304,00000002,00080007,00000005
04915200,00000002,00080007,00000005
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ # 8k? where'd that come from?!
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ sudo cat /sys/kernel/debug/dma_buf/bufinfo | grep -B1 'Attached' | grep -vi attached | sed 's/--//g' | sort -u | awk '{ printf "%s,%s,%s,%s\n", $1, $2, $3, $4 }' | head
,,,
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00008192,00000002,00080007,00000001
00524288,00000002,00080007,00000005
nslick@ect-p5-test-queue-dy-p548xlarge-4:~$ # exactly 8 of them, one for each process
```

bwbarrett: Ok, but if Nvidia said one thing is happening, and we're seeing another thing, the right answer is to find out what's happening that there's a difference.  The right answer is not to paper over it without understanding.

shijin-aws: > e other -- it's currently bugged.

Yeah, I should have checked the behavior before doing it. Thanks for the fix!

shijin-aws:
AWS CI failure is likely an infra issue. Will fix it and retest

prov/efa/src/efa_hmem.c
aws-nslick: I don't see this patch as papering over it, although I tend to agree that I'd prefer this to create a distinct pool, to use the cuda APIs end-to-end, and to model itself after either:

The cuda samples: [link](https://github.com/NVIDIA/cuda-samples/blob/9c688d7ff78455ed42e345124d1495aad6bf66de/Samples/2_Concepts_and_Techniques/streamOrderedAllocationIPC/streamOrderedAllocationIPC.cu#L256-L379)

```c++
    cudaMemPoolProps poolProps;
    memset(&poolProps, 0, sizeof(cudaMemPoolProps));
    poolProps.allocType = cudaMemAllocationTypePinned;
    poolProps.handleTypes = cudaMemHandleTypePosixFileDescriptor;

    poolProps.location.type = cudaMemLocationTypeDevice;
    poolProps.location.id = shm->devices[i];

    checkCudaErrors(cudaMemPoolCreate(&pools[i], &poolProps));

    // Query the shareable handle for the pool
    cudaMemAllocationHandleType handleType =
        cudaMemHandleTypePosixFileDescriptor;
    // Allocate memory in a stream from the pool just created
    checkCudaErrors(cudaMallocAsync(&ptr, DATA_SIZE, pools[i], streams[i]));

    checkCudaErrors(cudaMemPoolExportToShareableHandle(
        &shareableHandles[i], pools[i], handleType, 0));
[...]
  // Close the shareable handles as they are not needed anymore.
  for (int i = 0; i < shm->nprocesses; i++) {
    checkIpcErrors(ipcCloseShareableHandle(shareableHandles[i]));
  }
```

Or nccl: [link](https://github.com/NVIDIA/nccl/blob/master/src/init.cc#L2340-L2401)

```c++
    int requestedHandleTypes = CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
    // Query device to see if FABRIC handle support is available
    flag = 0;
    (void) CUPFN(cuDeviceGetAttribute(&flag, CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED, currentDev));
    if (flag) requestedHandleTypes |= CU_MEM_HANDLE_TYPE_FABRIC;
    memprop.type = CU_MEM_ALLOCATION_TYPE_PINNED;
    memprop.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
    memprop.requestedHandleTypes = (CUmemAllocationHandleType) requestedHandleTypes;
    memprop.location.id = currentDev;
    // Query device to see if RDMA support is available
    flag = 0;
    CUCHECK(cuDeviceGetAttribute(&flag, CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_WITH_CUDA_VMM_SUPPORTED, currentDev));
    if (flag) memprop.allocFlags.gpuDirectRDMACapable = 1;
    CUCHECK(cuMemGetAllocationGranularity(&memGran, &memprop, CU_MEM_ALLOC_GRANULARITY_RECOMMENDED));
    CUDACHECK(cudaGetDeviceCount(&dcnt));

    CUCHECK(cuMemCreate(&handle, size, &memprop, 0));
    /* Reserve a virtual address range */
    CUCHECK(cuMemAddressReserve((CUdeviceptr*)ptr, size, memGran, 0, 0));
    /* Map the virtual address range to the physical allocation */
    CUCHECK(cuMemMap((CUdeviceptr)*ptr, size, 0, handle, 0));
    /* Now allow RW access to the newly mapped memory */
    for (int i = 0; i < dcnt; ++i) {
      int p2p = 0;
      if (i == cudaDev || ((cudaDeviceCanAccessPeer(&p2p, cudaDev, i) == cudaSuccess) && p2p)) {
        accessDesc.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
        accessDesc.location.id = i;
        accessDesc.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
        CUCHECK(cuMemSetAccess((CUdeviceptr)*ptr, size, &accessDesc, 1));
      }
      if (0 == p2p && i != cudaDev) INFO(NCCL_ALLOC, "P2P not supported between GPU%d and GPU%d", cudaDev, i);
    }
    goto exit;
  }
```

aws-nslick: Our code does not remotely resemble either of the above, and I am a little bit frustrated that nobody else seems to have a problem with this. It's like 20 lines of code, and I'm not aware of any other piece of software in this space that is exporting handles the way we are, outside of test code. But this code runs on every single prov/efa initialization when cuda is available.

Why are we leaving it up to chance that this isn't silently causing issues? We agreed on #10445 that this was going to be reworked.

> I know there have been patchsets previously that looked to unify some of these flows, and I believe you're looking at reposting them. [...] So I'm going to repost this patch with just the minimal checks required to elide the gdr calls when dmabuf arguments are provided to fi_mr_regattr, with the understanding that the two error cases covered here are going to be irrelevant at some point in the near future.

Is this still going to happen?

shijin-aws: It depends on our conversation with EFA team on how we can detect whether NIC can access HBM on the PCIe level. Currently we are not aware of any approach without malloc/registering/deregistering

shijin-aws:
bot:aws:retest

iziemba:
In https://github.com/ofiwg/libfabric/pull/10716/, I have defined ofi_hmem_put_dmabuf_fd() for CUDA which could be used to resolve this issue.

PR #10711, opened 2025-01-20
prov/cxi: Add FI_CXI_CURL_LIB_PATH #define from autoconf 

### LINKED ISSUE #10710: prov/cxi: hardcoded path to libcurl breaks use on some setups
**Describe the bug**
although configure for the CXI provider has a `with-curl` option and the config tests check that the curl library has the correct init function. Inside the cxi provider itself, the code calls a hardcoded path to `dlopen` which causes problem on machines where another curl has been used due to deletion of a pointer at shutdown that is allocated in another TU/object file
https://github.com/ofiwg/libfabric/blob/950fb1b4f015044bbffde1ea7bf1d3feb83f35c9/prov/cxi/src/cxip_curl.c#L194-L204
Since we use spack to build our toolchain, and use curl as a dependency on libfabric/cxi we expect the libcurl we have compiled to be used

The problem comes at shutdown during finalize https://github.com/ofiwg/libfabric/blob/950fb1b4f015044bbffde1ea7bf1d3feb83f35c9/prov/cxi/src/cxip_curl.c#L271-L278 where the cleanup occurs and triggers a segfault as can be seen ...
```
fi_info -p cxi
provider: cxi
    fabric: cxi
    domain: cxi0
    version: 0.1
    type: FI_EP_RDM
    protocol: FI_PROTO_CXI
...
Segmentation fault (core dumped)
```

as an experiment, I added a new ENV var that I could set and got the following
```
export FI_CXI_CURL_LIBRARY_PATH=$(spack location -i curl)/lib/libcurl.so
```
and then
```
fi_info -p cxi
provider: cxi
    fabric: cxi
    domain: cxi0
    version: 0.1
    type: FI_EP_RDM
    protocol: FI_PROTO_CXI
...
```
The segfault is gone.

I would like submit a PR where we pass the library used during `--with-curl=` at configure and use that in place of the hardcoded `"/usr/lib64/libcurl.so.%d"` and thereby solve the problem

**To Reproduce**
Build libfavbric with CXI and a non-standard libcurl (eg spack installed curl)

**Expected behavior**
no segfaults

**Output**
see above

**Environment:**
cray/hpe machine with slingshot and spack installed toolchain

biddisco:
fix #10710 

ensures libcurl dlopen path is correct

If the user passes '--with-curl=<path>' to configure, then the dlopen of libcurl should honor that selection and use the file path passed in

### DIFF
--- a/prov/cxi/configure.m4
+++ b/prov/cxi/configure.m4
@@ -35,8 +35,10 @@ AC_DEFUN([FI_CXI_CONFIGURE],[
 		[CPPFLAGS="-I$with_cxi_uapi_headers/include $CPPFLAGS"])
 
 	# Support non-standard install path for curl. This is needed by CXI provider.
+	# Add #define of the path to the curl library for use in the code
 	AC_ARG_WITH([curl],
-		[AS_HELP_STRING([--with-curl=DIR], [Install directory for curl])])
+		[AS_HELP_STRING([--with-curl=DIR], [Install directory for curl])],
+		[AC_DEFINE_UNQUOTED([FI_CXI_CURL_LIB_PATH], ["$with_curl"], [Path to the curl install root])])
 
 	# Support non-standard install path for json-c. This is needed by CXI provider.
 	AC_ARG_WITH([json-c],
--- a/prov/cxi/src/cxip_curl.c
+++ b/prov/cxi/src/cxip_curl.c
@@ -189,28 +189,44 @@ int cxip_curl_load_symbols(void)
 	if (cxip_curlhandle)
 		return 0;
 
+	char *curl_libpath = NULL;
+	#ifdef FI_CXI_CURL_LIB_PATH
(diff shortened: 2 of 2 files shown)


iziemba:
Thanks John for opening this. I'll have someone review this and provide feedback.

heemstra (changes requested):
Thanks for the contribution, John.  I've tested this locally and it checks out.  The code looks good - the only nit I have is that we're not checking whether `curl_libpath` exceeds the size of `libfile`.

heemstra (approved):

iziemba (approved):

j-xiong:
@biddisco Intel CI is blocked on authentication step. Have you received the invitation to join ofiwg I sent you two days ago?

biddisco:
> @biddisco Intel CI is blocked on authentication step. Have you received the invitation to join ofiwg I sent you two days ago?

Thanks. Just found it and accepted invite.

PR #10715, opened 2025-01-21
prov/efa: Adjust the logging level for unreleased rxe

shijin-aws:
Closing an ep with outstanding receives is a legal use case. There shouldn't be warning for this.


sunkuamzn (approved):

shijin-aws:
bot:aws:retest

PR #10716, opened 2025-01-22
hmem: Define ofi_hmem_put_dmabuf_fd

iziemba:
For some HMEM ifaces, ofi_hmem_get_dmabuf_fd() may result in a new FD being allocated. Define ofi_hmem_put_dmabuf_fd() to close FD.

Support ofi_hmem_get_dmabuf_fd() with ROCR.

Update CXI provider accordingly.

### DIFF
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -168,6 +168,7 @@ int rocr_dev_reg_copy_from_hmem(uint64_t handle, void *dest, const void *src,
 				size_t size);
 int rocr_hmem_get_dmabuf_fd(const void *addr, uint64_t size, int *dmabuf_fd,
 			    uint64_t *offset);
+int rocr_hmem_put_dmabuf_fd(int fd);
 
 int cuda_copy_to_dev(uint64_t device, void *dev, const void *host, size_t size);
 int cuda_copy_from_dev(uint64_t device, void *host, const void *dev, size_t size);
--- a/include/ofi_hmem.h
+++ b/include/ofi_hmem.h
@@ -131,6 +131,7 @@ struct ofi_hmem_ops {
 				      const void *src, size_t size);
 	int (*get_dmabuf_fd)(const void *addr, uint64_t size, int *fd,
 			     uint64_t *offset);
+	int (*put_dmabuf_fd)(int fd);
 };
 
 extern struct ofi_hmem_ops hmem_ops[];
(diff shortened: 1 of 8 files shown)


shijin-aws:
would u mind implementing this for cuda as well? it should be simply a `free(fd)` and we may want to use it in https://github.com/ofiwg/libfabric/pull/10708

include/ofi_hmem.h
```diff
@@ -131,6 +131,7 @@ struct ofi_hmem_ops {
 				      const void *src, size_t size);
 	int (*get_dmabuf_fd)(const void *addr, uint64_t size, int *fd,
 			     uint64_t *offset);
+	int (*put_dmabuf_fd)(int fd);
```
shijin-aws, line 134: Why not just name it as `close_dmabuf_fd` ?

iziemba: I chose put_dmabuf_fd to semantically align with get_dmabuf_fd.

iziemba:
@shijin-aws : I added CUDA support. If you could review, that would be great.

@swelch : If you could review CXI prov changes, that would be great.

Thanks!

swelch (approved):

j-xiong (approved):

shijin-aws (approved):
Thanks!

shijin-aws:
@j-xiong is intel ci failure related?

j-xiong:
@shijin-aws We are having slurm issues right now. The tests didn't actually run.

j-xiong:
There are two intermittent failures with oneCCL over tcp provider. They are unrelated to this PR.

include/ofi_hmem.h
aws-nslick: is there any hmem type where a `get_dmabuf_fd` does not implicitly increase the open count on the file? if not, I'd recommend going the other way and renaming `get` to `open` so that it's clear that it's not simply resolving an already-open file descriptor, but is actually opening a resource which must be eventually closed.

j-xiong: yes, the ZE heme type always return the same fd.

prov/cxi/src/cxip_iomm.c
```diff
@@ -575,6 +589,9 @@ static void cxip_unmap_nocache(struct cxip_md *md)
 {
 	int ret;
 
+	if (md->dmabuf_fd_valid)
+		ofi_hmem_put_dmabuf_fd(md->info.iface, md->dmabuf_fd);
```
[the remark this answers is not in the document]
j-xiong: I assume the kernel driver would reference count properly so that closing right after mapping should work just fine. I don't see the drawback of doing either way.

PR #10718, opened 2025-01-22
cleanup mrecx entry if current rx processing failed

wangdi1:
mrecx entry should be removed if the current rx processing failed, otherwise the following process from the upper layer may pick up this stale rx.

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -1114,6 +1114,8 @@ static void xnet_complete_rx(struct xnet_ep *ep, ssize_t ret)
 cq_error:
 	FI_WARN(&xnet_prov, FI_LOG_EP_DATA,
 		"msg recv failed ret = %zd (%s)\n", ret, fi_strerror((int)-ret));
+	if (rx_entry->ctrl_flags & XNET_MULTI_RECV)
+		xnet_srx_cancel_rx(ep->srx, &ep->srx->rx_queue, rx_entry->context);
 	xnet_cntr_incerr(rx_entry);
 	xnet_report_error(rx_entry, (int) -ret);
 	xnet_free_xfer(xnet_ep2_progress(ep), rx_entry);
--- a/prov/tcp/src/xnet.h
+++ b/prov/tcp/src/xnet.h
@@ -228,6 +228,8 @@ struct xnet_srx {
 	xnet_profile_t *profile;
 };
 
+bool
+xnet_srx_cancel_rx(struct xnet_srx *srx, struct slist *queue, void *context);
 int xnet_srx_context(struct fid_domain *domain, struct fi_rx_attr *attr,
 		     struct fid_ep **rx_ep, void *context);
 
--- a/prov/tcp/src/xnet_srx.c
+++ b/prov/tcp/src/xnet_srx.c
@@ -712,7 +712,7 @@ found:
 	return rx_entry;
 }
 
-static bool
+bool
 xnet_srx_cancel_rx(struct xnet_srx *srx, struct slist *queue, void *context)
 {
 	struct slist_entry *cur, *prev;


wangdi1:
@shefty Could you please have a look? Thanks.

prov/tcp/src/xnet_progress.c
```diff
@@ -1114,6 +1114,8 @@ static void xnet_complete_rx(struct xnet_ep *ep, ssize_t ret)
 cq_error:
 	FI_WARN(&xnet_prov, FI_LOG_EP_DATA,
 		"msg recv failed ret = %zd (%s)\n", ret, fi_strerror((int)-ret));
+	if (rx_entry->ctrl_flags & XNET_MULTI_RECV)
+		xnet_srx_cancel_rx(ep->srx, &ep->srx->rx_queue, rx_entry->context);
```
shefty, line 1118: On the surface, this doesn't look right.  xnet_srx_cancel_rx() does:
```
	slist_foreach(queue, cur, prev) {
		xfer_entry = container_of(cur, struct xnet_xfer_entry, entry);
		if (xfer_entry->context == context) {
			slist_remove(queue, cur, prev);
			xnet_report_error(xfer_entry, FI_ECANCELED);
			xnet_free_xfer(xnet_srx2_progress(srx), xfer_entry);
			return true;
		}
	}
```
We don't need to search for the entry, since we already have it, and the code below calls xnet_report_error() and xnet_free_xfer().  So, this should end up with a double free, plus the error completion generated below will follow the cancel.

I think we just need the slist_remove() here.

shefty: Btw, I'm not even sure about the slist_remove().  That may not be needed anyway.  Can you describe the problem you're seeing?

wangdi1: Hmm, this is not for the current xfer_entry, but for the entry added in xnet_start_recv->xnet_alter_mrecv

https://github.com/ofiwg/libfabric/blob/main/prov/tcp/src/xnet_progress.c#L675

So we need use context to the search the entry from the list?

wangdi1: The issue we are seeing is that the stale rx_entry is being left in the rx_queue, with stale context, which is added by xnet_start_recv-> xnet_alter_mrecv().  But if the rx entry receiving failed, then only the current entry is removed, but not the mrecv entry, which can be picked up by the following progress, so causing the problem.

shefty: It's a multi-receive buffer.  Why does the remaining buffer space need to be removed?

wangdi1: The new entry might be added in xent_alter_mrecv if the buffer is not enough, if I understand correctly,  which will share the same context with the current rx, and the the context needs to be refreshed or even freed if multiple receive rx is completed, which will not cleanup these created mrecv entry. Do I understand correctly?

shefty: I didn't follow that.  The multi-recv buffer may be paired with many incoming messages.  They will all report the same user context, which corresponds with the posted receive.

When a new message is received, a portion of the multi-recv buffer is carved off for it.  The remaining buffer is left for additional messages. xnet_alter_mrecv() is reposting the remaining buffer.

wangdi1: Ah in current mercury implementation will cleanup context if any rx failed, i.e. completion callback of the multiple receive will be called immediately and the context are freed, but any rx entries with the context needs to be removed/freed. Unfortunately.

soumagne: One assumption that we had been making was that if the current multi-recv operation reports an error from the fi cq then we mark that operation as completed, meaning we assumed that no further receives could be made even if there is still space in the buffer. Is that behavior documented somewhere ?
What you are saying is that if the current multi-recv operation reports a failure then we should just keep proceeding regardless until the FI_MULTI_RECV flag gets set, marking the completion of the operation ?

shefty: And, yes, I think you would continue processing until the FI_MULTI_RECV flag is set.

soumagne: ok thanks, that was useful to know. Looks like it was not always reported though with tcp so I've opened https://github.com/ofiwg/libfabric/pull/10721

soumagne:
@j-xiong this PR can be closed

PR #10720, opened 2025-01-22
fabtests/efa: add rdma check for unsolicited write recv

jiaxiyan:
~~This includes 2 commits:~~
1. add rdma check for unsolicited write recv
~~2. Fix rx buf size to not exceed max_msg_size
When the option max_msg_size is set, the rx buf size should not exceed this limit.~~

### DIFF
--- a/fabtests/prov/efa/configure.m4
+++ b/fabtests/prov/efa/configure.m4
@@ -28,5 +28,10 @@ AS_IF([test x"$have_efadv" = x"1"], [
             [],
             [efa_rdma_checker_happy=0],
             [[#include <infiniband/efadv.h>]])
+
+        AC_CHECK_DECL(EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV,
+            [],
+            [efa_rdma_checker_happy=0],
+            [[#include <infiniband/efadv.h>]])
 ])
 AM_CONDITIONAL([BUILD_EFA_RDMA_CHECKER], [test $efa_rdma_checker_happy -eq 1])
--- a/fabtests/prov/efa/src/efa_rdma_checker.c
+++ b/fabtests/prov/efa/src/efa_rdma_checker.c
@@ -14,6 +14,7 @@
 enum rdma_op {
 	READ,
 	WRITE,
+	UNSOLICITED_WRITE_RECV,
 };
 
 /*
@@ -37,8 +38,10 @@ int main(int argc, char *argv[])
 				op = READ;
 			} else if (!strcasecmp(optarg, "write")) {
 				op = WRITE;
+			} else if (!strcasecmp(optarg, "writedata")) {
+				op = UNSOLICITED_WRITE_RECV;
 			} else {
-				fprintf(stderr, "Unknown operation '%s. Allowed: read | write'\n", optarg);
+				fprintf(stderr, "Unknown operation '%s. Allowed: read | write | writedata '\n", optarg);
 				return EXIT_FAILURE;
 			}
 			break;
@@ -46,7 +49,7 @@ int main(int argc, char *argv[])
 		case 'h':
 		default:
 			fprintf(stderr, "Usage:\n");
-			FT_PRINT_OPTS_USAGE("fi_efa_rdma_checker -o <op>", "rdma operation type: read|write");
+			FT_PRINT_OPTS_USAGE("fi_efa_rdma_checker -o <op>", "rdma operation type: read | write | writedata");
 			return EXIT_FAILURE;
         }
 	}
@@ -86,11 +89,25 @@ int main(int argc, char *argv[])
 	if (op == READ)
 		goto out;
 
-	if (efadv_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE) {
-		fprintf(stdout, "rdma write is enabled \n");
-	} else {
-		fprintf(stderr, "rdma write is NOT enabled \n");
-		err = op == WRITE ? 1 : 0;
+	if (op == WRITE) {
+		if (efadv_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE) {
+			fprintf(stdout, "rdma write is enabled \n");
+		} else {
+			fprintf(stderr, "rdma write is NOT enabled \n");
+			err = 1;
+		}
+		goto out;
+	}
+
+	if (op == UNSOLICITED_WRITE_RECV) {
+		if (efadv_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV) {
+			fprintf(stdout,
+				"rdma unsolicited write recv is enabled \n");
+		} else {
+			fprintf(stderr, "rdma unsolicited write recv is NOT "
+					"enabled \n");
+			err = 1;
+		}
 	}
 
 out:
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -74,7 +74,7 @@ def has_rdma(cmdline_args, operation):
     operation: rdma operation name, allowed values are read and write
     return: a boolean
     """
-    assert operation in ["read", "write"]
+    assert operation in ["read", "write", "writedata"]
     binpath = cmdline_args.binpath or ""
     cmd = "timeout " + str(cmdline_args.timeout) \
           + " " + os.path.join(binpath, f"fi_efa_rdma_checker -o {operation}")


shijin-aws (dismissed):

shijin-aws:
@j-xiong could you review https://github.com/ofiwg/libfabric/pull/10720/commits/8634657860b4a2c5febd3ea9586dc899e3aae77a

fabtests/common/shared.c
```diff
@@ -2565,6 +2565,8 @@ ssize_t ft_post_rx_buf(struct fid_ep *ep, size_t size, void *ctx,
 		       void *op_buf, void *op_mr_desc, uint64_t op_tag)
 {
 	size = MAX(size, FT_MAX_CTRL_MSG) + ft_rx_prefix_size();
+	if (opts.max_msg_size)
+		size = MIN(size, opts.max_msg_size);
```
j-xiong, line 2569: This is not the best place to adjust the size. For example, `ft_rx()` calls `ft_post_rx()` which in turns calls `ft_post_rx_buf()`. The size passed to `ft_rx()` is the global variable `rx_size`.  The value of `rx_size` has been adjusted by `ft_set_tx_rx_sizes()` based on `ep_attr->max_msg_size`, which is in turn set according to `opt.max_msg_size` via fi_setopt`.

If the test is calling ft_rx(), nothing needs to be done here. If the test is calling `ft_post_rx` or `ft_post_rx_buf()` directly, it's better to adjust the size at the call site.

jiaxiyan: Make sense. Updated.

shijin-aws: >The size passed to ft_rx() is the global variable rx_size. The value of rx_size has been adjusted by ft_set_tx_rx_sizes() based on ep_attr->max_msg_size, which is in turn set according to opt.max_msg_size via fi_setopt`.

I don't see rx_size is getting set per `opt.max_msg_size`. `opt.max_msg_size` can be smaller than `ep_attr->max_msg_size`. In that regard `ft_rx()` will finally still use the bigger `ep_attr->max_msg_size` to post rx buffer?

shijin-aws: I think we either need to do the override in `ft_post_rx_buf` as the earlier revision of this PR, or have a way to modify `rx_size` globally after the `fi_setopt` call.

shijin-aws: Overriding the recv size inside `ft_post_rx_buf` is the best change I can think of, as `rx_size` is used to allocate receive buffer for RMA operations as well which can have larger size than `opts.max_msg_size`

j-xiong: The posted buffer size may need to be larger than opts.max_msg_size due to prefix buffer requirements.

fabtests/common/shared.c
```diff
@@ -1519,7 +1520,10 @@ int ft_enable_ep_recv(void)
 	if (!ft_check_opts(FT_OPT_SKIP_MSG_ALLOC) &&
 	    (fi->caps & (FI_MSG | FI_TAGGED))) {
 		/* Initial receive will get remote address for unconnected EPs */
-		ret = ft_post_rx(ep, MAX(rx_size, FT_MAX_CTRL_MSG), &rx_ctx);
+		size = MAX(rx_size, FT_MAX_CTRL_MSG);
+		if (opts.max_msg_size)
+			size = MIN(size, opts.max_msg_size);
+		ret = ft_post_rx(ep, size, &rx_ctx);
 		if (ret)
```
j-xiong, line 1527: I don't think you need to change the code that originally used `rx_size`. As I said previously, `rx_size` has already been adjusted based on `ep_attr->max_msg_size` (see `ft_set_tx_rx_size` at L539, and L600 inside `ft_alloc_msgs`), which is in turn adjusted based on `opts.max_msg_size` (see L1465, inside function `ft_enable_ep`).  

The resulting size may be larger than `opts.max_msg_size`, but that is necessary since the buffer need to be able to handle control messages (thus >= `FT_MAX_CTRL_MSG`) and prefix buffer requirement imposed by the provider.

shijin-aws: >which is in turn adjusted based on opts.max_msg_size (see L1465, inside function ft_enable_ep).

@j-xiong where does `rx_size` get adjusted based on `opts.max_msg_size`? I didn't see L1465 does it.

j-xiong:
```
	if (opts.max_msg_size) {
		ret = fi_setopt(&bind_ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE,
				&opts.max_msg_size, sizeof opts.max_msg_size);
		if (ret && ret != -FI_EOPNOTSUPP) {
			FT_PRINTERR("fi_setopt(FI_OPT_MAX_MSG_SIZE)", ret);
			return ret;
		}
	}
```
Not every provider handles the option (at least efa does), but that is where it is set.

j-xiong: ok, I take it back, it sets ep internal value, not ep_attr.

shijin-aws: Yeah. They can be different, `ep_attr->max_msg_size` should still be the largest supported size per man page.

j-xiong: Anyway, the best place to adjust is inside `ft_set_tx_rx_size`, after checking ep_attr.

shijin-aws: and `ep_attr->max_msg_size` doesn't distinguish on operation type. ep internal `max_msg_size` is for msg operation only

shijin-aws: > Anyway, the best place to adjust is inside `ft_set_tx_rx_size`, after checking ep_attr.

The problem is that `rx_size` is also used to allocate rx buffer for RMA operations today, which has different max size with `opts.max_msg_size`

shijin-aws: We can distinguish `rx_msg_size` and `rx_rma_size` though. Then it needs larger changes.

shijin-aws: We can modify `rx_size` after allocating the buffer? It sounds too hacky to me. Maybe we can split rx_size to `rx_msg_size` and `rx_rma_size`.. it will be larger change though

j-xiong: We need to think a little bit more about this. Is this commit needed for this PR? Can it be dropped?

jiaxiyan: Yes it can be a follow up PR.

shijin-aws (dismissed):

j-xiong:
Intel CI failure：
```
prov/efa/src/efa_rdma_checker.c: In function ‘main’:
prov/efa/src/efa_rdma_checker.c:103:46: error: ‘EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV’ undeclared (first use in this function)
  103 |                 if (efadv_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV) {
      |                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
prov/efa/src/efa_rdma_checker.c:103:46: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [Makefile:2640: prov/efa/src/efa_rdma_checker.o] Error 1
make: *** [Makefile:1756: all] Error 2
```
Does it require new efa headers?

shijin-aws:
> Intel CI failure：
> 
> ```
> prov/efa/src/efa_rdma_checker.c: In function ‘main’:
> prov/efa/src/efa_rdma_checker.c:103:46: error: ‘EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV’ undeclared (first use in this function)
>   103 |                 if (efadv_attr.device_caps & EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV) {
>       |                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> prov/efa/src/efa_rdma_checker.c:103:46: note: each undeclared identifier is reported only once for each function it appears in
> make[1]: *** [Makefile:2640: prov/efa/src/efa_rdma_checker.o] Error 1
> make[1]: *** Waiting for unfinished jobs....
> make: *** [Makefile:1756: all] Error 2
> ```
> 
> Does it require new efa headers?

Yes, we missed a check in configure

shijin-aws:
@jiaxiyan you need to add a check for `EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV` to this https://github.com/ofiwg/libfabric/blob/main/fabtests/prov/efa/configure.m4#L27

shijin-aws (approved):

PR #10721, opened 2025-01-23
prov/tcp: Fix FI_MULTI_RECV not set on cancel

soumagne:
When multiple multi-recv buffers were posted, FI_MULTI_RECV would have been set only if there was an mrecv entry already created, meaning the buffer would have already been in-use. If the buffer had not been used yet, the cancel error would have been reported to the error queue without FI_MULTI_RECV set, indicating no completion.


j-xiong:
Although this is first observed with the cancel operations, the issue and the fix also apply to other errors. Please update the commit message to be more precise. In addition, the wording "indicating no completion" is somewhat confusing. The flag indicates the multi-recv buffer has been fully consumed. Without the flag, the application may think the buffer is still in use.

j-xiong (approved):

PR #10724, opened 2025-01-24
prov/tcp: implement FI_NO_CONNECT for AV inserts

### LINKED ISSUE #10637: Handling TCP traffic restricted by a firewall
Comments copied from #10534:

There are some specific use cases where we may not want one side of communication to initiate connections, namely when we know that one side of our configuration is being heavily restricted by a firewall. To prevent indefinite hangs with certain operations, such as RMA reads and writes, introduce a provider specific flag to trigger an error if there is not already an established connection. In this case, the application can force the connection from the other direction.

[shefty](https://github.com/shefty) reviewed [2 weeks ago](https://github.com/ofiwg/libfabric/pull/10534#pullrequestreview-2465651594)
[man/fi_tcp.7.md](https://github.com/ofiwg/libfabric/pull/10534/files/95ae4b7a59f0b1aa7f318809a5ebbc5e4854a489#diff-7e68187ebb0bbb8e3e1e0878b0ab70f996c852b7c47ee1cedaa6751cd3bf8638)
*FI_TCP_NO_CONNECT*
: This flag indicates that operations should fail if there is no
  existing connection to the remote peer.  In such case, an FI_ENOTCONN
  error should be expected.
Member
@[shefty](https://github.com/shefty) shefty 2 weeks ago
I would not make this a flag that's checked on every operation (i.e. it's okay to connect if it's a send, but not a write?). It makes more sense either applied to the entire rdm endpoint or to a specific peer.

Member
Author
@[ooststep](https://github.com/ooststep) ooststep 2 days ago • 
in the target problematic scenario, the rdm is used for some peers where this flag is needed and some peers where it would not be, so applying to the entire rdm was undesirable.

Member
@[shefty](https://github.com/shefty) shefty 2 days ago
This doesn't make sense. An RDM endpoint is unconnected. Exposing low-level connection implementation details is not desirable. There could be multiple connections to the same peer. The connection might be in another process (e.g. Pony Express or SNAP or whatever it's called). It could be in the kernel (RDS).

This still isn't a per operation flag. At best it's per peer, but even that use case seems questionable. It's like putting half of an RDM endpoint behind a firewall, but the other half ignores it. Apply firewall semantics to the entire RDM endpoint. If some peers are outside the firewall, but some are inside, require 2 endpoints with some sort of per EP configuration.

Member
@[soumagne](https://github.com/soumagne) soumagne 2 days ago
I think I also agree with the first part (ie. exposing low-level connection is a bad idea) though providing 2 types of endpoints is also difficult for the user. To provide some context as to what problem we were trying to solve, in a client-server configuration where the client is behind a firewall, we have cases where for example client A would send a message to server A, and server A would then attempt to do an emulated tcp RMA to client B through an fi_write() call (without prior connection established from client B to server A). In that case, it seems that with the tcp provider server A remains stuck attempting to establish a connection and doing an fi_cancel on the RMA does not seem to be able to complete, as it's not supported currently by the tcp provider. So what we wanted was a way of having some completion and error being returned when server A is not able to reach client B. I'm opened to other means but having to manage 2 endpoints seems also cumbersome, I would also be happy though if we don't have to expose any connection logic.

@[jolivier23](https://github.com/jolivier23) jolivier23 2 days ago • 
@shefty Just to add a bit of context beyond what Jerome said, this use case came from Parallelstore (Google's DAOS service). We control only the server side of the equation and rely on the user to do client configuration (as we don't control their VMs, network configuration, and processes). Opening the firewall to server to client connections is not common for services and requires them to do it explicitly or their writes will simply hang. We want to remove this requirement as it is becoming a major scale issue for onboarding new customers because users don't read documentation.

Member
@[shefty](https://github.com/shefty) shefty 2 days ago
Client A is telling a server to issue an RMA write to client B? A knows the memory information on client B, and B is expecting this transfer? Is this the model used by the app? Because if so, weird.

In any case, the entire RDM ep is behind a firewall, but connections can only go one direction? If this is the case, then I think some generic EP setting might be able to convey this, such that the implementation can adjust. The server could probably just fail the operation (asynchronously makes the most sense) in this situation.

I assume A is somehow notified what happened between the server and B?

Member
@[soumagne](https://github.com/soumagne) soumagne 2 days ago
oh yes sorry I was trying to keep it simple and in the end that didn't make much sense, the real model is client A tells server A, which tells server B about its memory information, which then issues an RMA to client A.

Member
@[soumagne](https://github.com/soumagne) soumagne 2 days ago
In any case, the entire RDM ep is behind a firewall, but connections can only go one direction? If this is the case, then I think some generic EP setting might be able to convey this, such that the implementation can adjust. The server could probably just fail the operation (asynchronously makes the most sense) in this situation.

yeah I think adding a setting to the server EP would probably be fine in this case, although we'd still want servers to connect to each other using the same endpoint and not use a separate endpoint for intra-communication.

Member
@[shefty](https://github.com/shefty) shefty 2 days ago
I think we need to figure out the model for how firewalls should be handled, ignoring the implementation. Maybe this is related to NAT as well.

An EP is assigned a single address. Can traffic to that address go through the firewall? Can traffic from that address go through the firewall?

Assume the implementation is using UDP (or some other connectionless transport). Does that change anything about how the firewall must be configured?

If I'm understanding this now, what's being requested from the API is: Traffic is allowed to reach an EP (through a firewall), but traffic from that EP cannot reach the target ... unless it's already received some message from the target. This assumes the EP is responsible for remembering which target addresses it has received data from, and that those addresses still reference the correct target process.

From the viewpoint of the API, this still seems weird. I think the weirdness comes from the firewall acting on the lower-level connections, which the EP hides.

This is some per peer setting. But if the application is responsible for providing the value of this setting, then it has an idea how to manage this. In this example, server B could see if it knows A. If not, fail sending to it.

One alternative is to configure the EP with firewall data only, and let the provider determine what to do with it. In this case, a request to send to X checks the firewall config to see if it's possible. Maybe X is another server within the firewall. If so, connect and send. If X is outside the firewall, fail the send. I don't know what the firewall data would look like but providing it via some configuration file seems reasonable. I.e. don't make this an application problem.

@[jolivier23](https://github.com/jolivier23) jolivier23 2 days ago
That alternative is a bit is tricky. The client may be older than the server and we want old clients to continue behaving as they do now (e..g if the user has already configured their firewall correctly and the client doesn't support getting back an error and retrying, we want it to just continue working).

Member
@[shefty](https://github.com/shefty) shefty yesterday
For the alternative, I would only configure the EPs at the server, as that's where the actual problem occurs. The client behavior is unchanged.

@[jolivier23](https://github.com/jolivier23) jolivier23 11 hours ago
I think the problem is really on the client. The client endpoint is the one behind the firewall and can't be reached (unless the server has already connected to it). Can we encode something into the client side URI that would indicate it can't be reached? The advantage to that approach from our end is that then the client could have complete control over telling the server whether or not it can handle the error (e.g. can support getting back an error indicating that the server couldn't connect and handle it appropriately)

Member
@[shefty](https://github.com/shefty) shefty 1 hour ago
It's the server's behavior that should change.

From the viewpoint of the client, everything works. It wants to talk to server A and can do so. That server A wants to pass off the response to some other system that the client doesn't know about is related to the storage architecture, not the client SW.

I don't think pushing this detail into the apps is the best option. But you can work-around this in the client by having the client send some sort of 'hello' message to every storage server during initialization -- to poke holes through the firewall. That pushes the burden onto every client app that might want to use DAOS.

Alternatively, you can configure the server SW to be firewall aware, so that it avoids forwarding requests to servers not already communicating with the client.

Or, change the protocol around handling firewalls. Have server A tell the client to retry its request with server B, rather than forwarding it internally.

There are likely other options for this. But I would avoid picking one which encoded these details in the SW API.

Member
@[shefty](https://github.com/shefty) shefty 1 hour ago
It would probably be better if this thread were copied into an issue for continued discussion, rather than attaching it to this PR.

ooststep:
This PR introduces a new `fi_av_insert` flag, `FI_NO_CONNECT`, which indicates that an av insert should fail if there is not already an established connection to the peer.  The PR also implements the functionality for the TCP provider (and RXM provider, though this is not the primary goal).

I do still need to add a fabtest for the feature but wanted to push this up for initial review in the meantime.

resolves #10637
replaces #10534


include/rdma/fabric.h
```diff
@@ -159,6 +159,7 @@ typedef struct fid *fid_t;
 #define FI_PEER_TRANSFER	(1ULL << 36)
 /* #define FI_MR_DMABUF		(1ULL << 40) */
 #define FI_AV_USER_ID		(1ULL << 41)
+#define FI_NO_CONNECT		(1ULL << 42)
```
shefty, line 162: Rather than trying to control the behavior of the implementation, it may make more sense to define some property of the addresses being inserted.  For example, FI_FIREWALL_ADDR can indicate that the addresses being inserted are behind a firewall.  The provider can then decide what to do, or whether it has the ability to access the address behind the firewall.  In the future, there may be some sort of firewall configuration that may allow communicating with the target address, such as going through a proxy.

prov/util/src/rxm_av.c
```diff
@@ -98,23 +98,32 @@ static void rxm_free_peer(struct util_peer_addr *peer)
...
+			*peer = NULL;
+			ret = -FI_EHOSTUNREACH;
+			goto unlock;
+		}
 
+		*peer = rxm_alloc_peer(av, addr);
+		if (!*peer)
+			ret = -FI_ENOMEM;
+	}
+unlock:
```
shefty, line 124: nit: This can be restructured as:
```
if (node) {
} else if (flags & ...) {
} else {
}
```

man/fi_av.3.md
```diff
@@ -436,6 +436,11 @@ fi_av_set_user_id.
 
   See the user ID section below.
 
+- *FI_NO_CONNECT*
+: This flag indicates that AV insertion operations should fail if there
+  is not an existing connection to the peer. In such a case, an
+  FI_EHOSTUNREACH should be expected.
```
shefty, line 442: See above comment.  I would clarify that if multiple addresses are inserted at once, the flag applies to all of them.

It's also worth noting that the same error code may be returned if the target address becomes unreachable in the future

prov/util/src/rxm_av.c
```diff
@@ -98,23 +98,32 @@ static void rxm_free_peer(struct util_peer_addr *peer)
...
 	struct ofi_rbnode *node;
+	int ret;
 
 	ofi_mutex_lock(&av->util_av.lock);
 	node = ofi_rbmap_find(&av->addr_map, (void *) addr);
 	if (node) {
-		peer = node->data;
-		peer->refcnt++;
+		*peer = node->data;
+		(*peer)->refcnt++;
```
shefty, line 112: It's possible that a peer may be reachable at the time that the address is inserted, but if the connection is broken, it becomes unreachable.  That results in an unusable address in the AV.  The flag needs to be stored in the AV with the address.  Additional checks are needed to avoid trying to re-connect in such cases.

shefty: No.  There may be usable connection at the time the connection is inserted.  However, any error on the connection which occurs later will tear it down.  The next send will then try to reconnect.  That send needs to report an EHOSTUNREACH error.

prov/util/src/rxm_av.c
shefty: I would switch to if-else if-else.  This removes the goto and clearly shows the order of the checks.

prov/util/src/rxm_av.c
ooststep: Got it - thanks

prov/util/src/rxm_av.c
```diff
@@ -98,23 +98,29 @@ static void rxm_free_peer(struct util_peer_addr *peer)
...
```
shefty, line 112: Set ret = FI_SUCCESS here to avoid an extra branch on return.

prov/util/src/rxm_av.c
```diff
@@ -98,23 +98,29 @@ static void rxm_free_peer(struct util_peer_addr *peer)
...
+	} else if (flags & FI_FIREWALL_ADDR) {
+		*peer = NULL;
+		ret = -FI_EHOSTUNREACH;
 	} else {
-		peer = rxm_alloc_peer(av, addr);
+		*peer = rxm_alloc_peer(av, addr);
+		if (!*peer)
+			ret = -FI_ENOMEM;
+		(*peer)->av_flags = flags;
```
shefty, line 120: Needs to be part of an else, or this results in a crash.

prov/util/src/rxm_av.c
```diff
@@ -98,23 +98,29 @@ static void rxm_free_peer(struct util_peer_addr *peer)
 	ofi_ibuf_free(peer);
 }
 
-struct util_peer_addr *
-util_get_peer(struct rxm_av *av, const void *addr)
+
+int util_get_peer(struct rxm_av *av, const void *addr, struct util_peer_addr **peer,
```
shefty, line 102: nit: check line length

prov/util/src/rxm_av.c
shefty: Something is off in the flag settings.  A peer might be inserted in response to a connection request (xnet_process_connreq).  In that case, peer->av_flags = 0.  When an address is inserted with FI_FIREWALL_ADDR, the peer will be found, but peer->av_flags won't be adjusted.

I think you need ```(*peer)->av_flags |= flags;``` here.

shefty:
The latest changes look reasonable to me.

prov/rxm/src/rxm_conn.c
```diff
@@ -657,9 +660,9 @@ rxm_process_connreq(struct rxm_ep *ep, struct rxm_eq_cm_entry *cm_entry)
...
 
 	av = container_of(ep->util_ep.av, struct rxm_av, util_av);
-	peer = util_get_peer(av, &peer_addr);
-	if (!peer) {
-		RXM_WARN_ERR(FI_LOG_EP_CTRL, "util_get_peer", -FI_ENOMEM);
+	ret = util_get_peer(av, &peer_addr, &peer, 0);
+	if (ret) {
+		RXM_WARN_ERR(FI_LOG_EP_CTRL, "util_get_peer", ret);
 		goto reject;
 	}
```
jxiong, line 667: @j-xiong In my experiment, I just invoked `fi_av_insert()` to add the source address into av as follows:
```
ret = fi_av_insert(&av->util_av.av_fid, &peer_addr, 1, &peer->fi_addr, 0, NULL);
```
and then I'm planning to revise `xnet_cm_msg{}` to carry bind firewall information, and insert the address with `FI_FIREWALL_ADDR` if that is the case.

It seems to work well but I don't know if there is any side effect though

j-xiong: I believe you are referring to a similar code segment in `xnet_rdm_cm.c`.
 
`struct xnet_rdm_cm` has a one-byte reserved field that can be used to carry a flag indicating if the connection request is coming from behind a firewall. The passive side can then pass `FI_FIREWALL_ADDR` flag to the `util_get_peer()` call here accordingly.  That will set the flag for the peer regardless of whether it has been inserted into the address vector or not.

jxiong: yeah - if there is no objection, I can start a prototype and create a PR for feedback. @shefty @ooststep FYI

shefty: It sounds like a reasonable approach

jxiong:
@ooststep do you mind if I hijacked this PR for my new change? I could create a new one but I reused some of your work.

jxiong:
> > @ooststep do you mind if I hijacked this PR for my new change? I could create a new one but I reused some of your work.
> 
> I don't have issue with that. Works for me

Nevermind, I don't have permission to write your fork. I pushed two PRs here:
https://github.com/ofiwg/libfabric/pull/10912
https://github.com/jxiong/libfabric/pull/2

ooststep:
Ok, I would recommend doing this slightly differently. pull my branch into your working copy, then add your firewall addr changes as a separate commit on top of mine. Would merging this PR first make it easier?

jxiong:
it works for me either way. How about merging your first commit and I can stack my changes based on that? This way we can make it a cleaner git history?

PR #10725, opened 2025-01-25
prov/rxm: fix rxm multi recv getopt segfault

soumagne:
this prevents a segfault when the application calls fi_getopt of FI_OPT_MIN_MULTI_RECV before the endpoint gets enabled.

### DIFF
--- a/prov/rxm/src/rxm_ep.c
+++ b/prov/rxm/src/rxm_ep.c
@@ -270,8 +270,17 @@ static int rxm_ep_getopt(fid_t fid, int level, int optname, void *optval,
 
 	switch (optname) {
 	case FI_OPT_MIN_MULTI_RECV:
-		return rxm_ep->srx->ep_fid.ops->getopt(&rxm_ep->srx->ep_fid.fid,
-						level, optname, optval, optlen);
+		if (rxm_ep->srx) {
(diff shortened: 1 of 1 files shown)


soumagne:
@aingerson tagging you since this fix seems related to some of your previous patches.

prov/rxm/src/rxm_ep.c
```diff
@@ -270,8 +270,10 @@ static int rxm_ep_getopt(fid_t fid, int level, int optname, void *optval,
 
 	switch (optname) {
 	case FI_OPT_MIN_MULTI_RECV:
-		return rxm_ep->srx->ep_fid.ops->getopt(&rxm_ep->srx->ep_fid.fid,
-						level, optname, optval, optlen);
+		assert(sizeof(rxm_ep->min_multi_recv_size) == sizeof(size_t));
+		*(size_t *)optval = rxm_ep->min_multi_recv_size;
+		*optlen = sizeof(size_t);
+		break;
```
j-xiong, line 276: Both situations need to be handled, for example:
```
        if (rxm_ep->srx) {
                /* the old code here */
        } else {
                /* the new code here */
        }
```

soumagne: ok fixed

j-xiong (approved):

PR #10734, opened 2025-01-27
prov/util: fix FI_MULTI_RECV not set on FI_ECANCELED

soumagne:
When canceling a multi-recv entry, FI_MULTI_RECV was never set as part of the error entry flags. Fix so that the user can determine if the buffer is still in use.

### DIFF
--- a/prov/util/src/util_srx.c
+++ b/prov/util/src/util_srx.c
@@ -915,6 +915,10 @@ static int util_cancel_entry(struct util_srx_ctx *srx, uint64_t flags,
 
 	err_entry.op_context = rx_entry->peer_entry.context;
 	err_entry.flags = flags;
+	if (rx_entry->peer_entry.flags & FI_MULTI_RECV) {
+		assert(rx_entry->multi_recv_ref == 0);
+		err_entry.flags |= FI_MULTI_RECV;
+	}
 	err_entry.tag = rx_entry->peer_entry.tag;
 	err_entry.err = FI_ECANCELED;
 	err_entry.prov_errno = -FI_ECANCELED;
@@ -1027,8 +1031,8 @@ static struct fi_ops util_srx_fid_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
-static bool util_cancel_recv(struct util_srx_ctx *srx, struct slist *queue,
-			     uint64_t flags, void *context)
+static int util_cancel_recv(struct util_srx_ctx *srx, struct slist *queue,
+			    uint64_t flags, void *context)
 {
 	struct slist_entry *item, *prev;
 	struct util_rx_entry *rx_entry;
@@ -1037,12 +1041,19 @@ static bool util_cancel_recv(struct util_srx_ctx *srx, struct slist *queue,
 	slist_foreach(queue, item, prev) {
 		rx_entry = container_of(item, struct util_rx_entry, peer_entry);
 		if (rx_entry->peer_entry.context == context) {
+			/* With multi-recv, cancellation can only be processed
+			 * if the multi-recv buffer is no longer in use. If the
+			 * buffer is still in use, return EAGAIN back for user
+			 * to retry cancel request. */
+			if (rx_entry->peer_entry.flags & FI_MULTI_RECV &&
+			    rx_entry->multi_recv_ref)
+				return -FI_EAGAIN;
 			slist_remove(queue, item, prev);
 			util_cancel_entry(srx, flags, rx_entry);
-			return true;
+			return FI_SUCCESS;
 		}
 	}
-	return false;
+	return -FI_ENOENT;
 }
 
 static int util_cancel_src(struct ofi_dyn_arr *arr, void *list, void *context)
@@ -1056,31 +1067,36 @@ static int util_cancel_src(struct ofi_dyn_arr *arr, void *list, void *context)
 	flags = arr == &srx->src_trecv_queues ?
 		FI_TAGGED | FI_RECV : FI_MSG | FI_RECV;
 
-	return (int) util_cancel_recv(srx, queue, flags, context);
+	return (int) (util_cancel_recv(srx, queue, flags, context) == FI_SUCCESS);
 }
 
 static ssize_t util_srx_cancel(fid_t ep_fid, void *context)
 {
 	struct util_srx_ctx *srx;
+	ssize_t ret;
 
 	srx = container_of(ep_fid, struct util_srx_ctx, peer_srx.ep_fid);
 
 	ofi_genlock_lock(srx->lock);
-	if (util_cancel_recv(srx, &srx->tag_queue, FI_TAGGED | FI_RECV,
-			     context))
+	ret = util_cancel_recv(srx, &srx->tag_queue, FI_TAGGED | FI_RECV,
+			       context);
+	if (ret != -FI_ENOENT)
 		goto out;
 
-	if (util_cancel_recv(srx, &srx->msg_queue, FI_MSG | FI_RECV, context))
+	ret = util_cancel_recv(srx, &srx->msg_queue, FI_MSG | FI_RECV, context);
+	if (ret != -FI_ENOENT)
 		goto out;
 
-	if (ofi_array_iter(&srx->src_trecv_queues, context, util_cancel_src))
-		goto out;
+	if (ofi_array_iter(&srx->src_trecv_queues, context, util_cancel_src) ||
+	    ofi_array_iter(&srx->src_recv_queues, context, util_cancel_src)) {
+		/* nothing to do, always return success */
+	}
 
-	(void) ofi_array_iter(&srx->src_recv_queues, context, util_cancel_src);
+	ret = FI_SUCCESS;
 
 out:
 	ofi_genlock_unlock(srx->lock);
-	return FI_SUCCESS;
+	return ret;
 }
 
 static int util_srx_getopt(fid_t fid, int level, int optname,
--- a/man/fi_endpoint.3.md
+++ b/man/fi_endpoint.3.md
@@ -1704,7 +1704,9 @@ required by the application.
 
 Returns 0 on success.  On error, a negative value corresponding to
 fabric errno is returned.  For fi_cancel, a return value of 0
-indicates that the cancel request was submitted for processing.
+indicates that the cancel request was submitted for processing,
+a return value of -FI_EAGAIN indicates that the request could not be
+submitted and that it should be retried once progress has been made.
 For fi_setopt/fi_getopt, a return value of -FI_ENOPROTOOPT
 indicates the provider does not support the requested option.
 


j-xiong:
This is fixing the utility provider, please change the title to start with `prov/util:`.

soumagne:
ah right thanks fixed.

prov/util/src/util_srx.c
```diff
@@ -915,6 +915,10 @@ static int util_cancel_entry(struct util_srx_ctx *srx, uint64_t flags,
 
 	err_entry.op_context = rx_entry->peer_entry.context;
 	err_entry.flags = flags;
+	if (rx_entry->peer_entry.flags & FI_MULTI_RECV) {
+		assert(rx_entry->multi_recv_ref == 0);
```
aingerson, line 919: Do we want to assert this? What if we try to cancel an mrecv that has already matched? That shouldn't be a failure - the cancel API allows already matched entries to complete normally. We can either treat pending mrecvs attached to the owner as making it uncancellable or we can mark it as cancelled and let the already matched receives to complete and then the last one can fully cancel the owner entry. @j-xiong what do you think?

j-xiong: I agree the handling here is over simplified. An example of proper handling is in the function `util_free_entry`. We could either replace the `ofi_buf_free` call at L923 with `util_free_entry` or duplicate similar logic here. The difference is  that with the former a separate completion entry is generated with the FI_MULTI_RECV flag set.

aingerson: I don't think we can replace it with the `util_free_entry` call since that logic deals with the segment entries, not the owner entry. The owner entry that we're canceling here will never have the owner_entry field set (only the sub-segments do) so it won't trigger the dereferencing and will essentially just be a call to `ofi_buf_free`. 
Instead, I think we need to only call `ofi_buf_free` here for the non mrecv case and in the mrecv case set a flag that we check in `util_free_entry` so that we free the owner buffer in that case as well as when the length is done but then in the cancel case, don't complete again since we complete in the cancel path. Basically, something like this:
```
static void util_free_entry(struct fi_peer_rx_entry *entry)
{
	struct util_srx_ctx *srx;
	struct util_rx_entry *util_entry, *owner_entry;

	srx = (struct util_srx_ctx *) entry->srx->ep_fid.fid.context;
	assert(ofi_genlock_held(srx->lock));

	util_entry = container_of(entry, struct util_rx_entry, peer_entry);
	if (entry->owner_context) {
		owner_entry = (struct util_rx_entry *) entry->owner_context;
		if (!--owner_entry->multi_recv_ref) {
			if (owner_entry->flags & FI_ECANCELED) {
				ofi_buf_free(owner_entry);
			} else if (owner_entry->peer_entry.msg_size < srx->min_multi_recv_size) {
				if (ofi_peer_cq_write(srx->cq,
						owner_entry->peer_entry.context,
						FI_MULTI_RECV, 0, NULL, 0, 0,
						FI_ADDR_NOTAVAIL)) {
					FI_WARN(&core_prov, FI_LOG_EP_CTRL,
						"cannot write MULTI_RECV completion\n");
				}
				ofi_buf_free(owner_entry);
			}
		}
	}
	ofi_buf_free(util_entry);
}
```

shefty: Conceptually, cancel can be viewed as handling a receive that fills the rest of the multi-recv buffer, but with an error status.  Can the existing code path for that be reused?

aingerson: I considered just setting the size to 0 to mark that the buffer is done and should be freed but that would result in the common path generating a successful completion entry for the end of the multi recv. We will need to add another field to capture the error status to either not complete (because we completed in error already) or we could move the error completion generation to the common path. I think either is fine, but some change to the existing code is needed to avoid the successful completion

shefty: Could you check if the refcnt == 0.  If so, basically fall through like the proposed changes.  If refcnt > 0, set the size to 0, so the last completion reported will indicate the multi-recv is done?  Could the app tell any difference?

aingerson: We could do that, but that wouldn't generate the cancel completion entry. It would generate a successful completion instead. So I think it comes down to what the app would expect in the cq for the cancel. I would assume that if we set the size to 0, that would be equivalent to a successful cancel, ie and error entry should be generated for that rx, and not a successful cq entry.

shefty: Ignoring multi-recv, canceling a recv can fail because the buffer was already taken by a message and will complete successfully.  The goal is to return the buffer to the user.

I'm not suggesting to generate a 0-byte receive completion.  If the refcnt == 0 when cancel is called, generate a canceled completion with the FI_MULTI_RECV flag.  If refcnt > 0, mark the buffer as done, so when the last active message completes it includes the FI_MULTI_RECV flag.

It's challenging to have canceling a multi-recv make sense.  Reporting a CANCELED completion before any of the active messages ends up confusing.  The buffer is still in use.  What does the CANCELED completion mean?  The CANCELED completion needs to come last.  But if we have an active message, why not just let it return the buffer?

aingerson: Yes, I agree. Canceling with multi recv doesn't make sense and is too confusing. My hang up with the generating the extra completion is that there are 2 ways to generate the mrecv done completion - one where you add it to the last completion entry and one where you generate a second completion entry just to signal the end of the buffer. Because of the nature of the peer API, the peer srx has to take the second approach where you generate a second completion for the buffer as a whole once it's done. So with cancel I think it gets confusing because then you're generating 2 completion entries for the same buffer - one for the cancel and one to mark that the mrecv is done and I think that is confusing and could cause issues with the app reposting the same mrecv because it thinks it was both canceled and completed successfully. 
This is why I think if you're going to allow canceling of mrecvs that are partially used/in use, you have to modify the free entry call to suppress the successful completion. The alternative would be to just say that if you try to cancel an mrecv that is in use, you can't and just treat it as an already matched recv. Basically modify the cancel search to check for the mrecv and ref cnt:
```
	slist_foreach(queue, item, prev) {
		rx_entry = container_of(item, struct util_rx_entry, peer_entry);
		if (rx_entry->peer_entry.context == context) {
			if (rx_entry->peer_entry.flags & FI_MULTI_RECV &&
			    rx_entry->multi_recv_ref)
				return false;
			slist_remove(queue, item, prev);
			util_cancel_entry(srx, flags, rx_entry);
			return true;
		}
	}
	return false;
```

shefty: I agree there can only be one completion with FI_MULTI_RECV set.  Maybe we should ask why an app is wanting to cancel a multi-recv.

Your proposal to fail the cancel if the buffer is in use is okay.  It's nice because it's simple.  It basically returns EAGAIN to the cancel request and requires the app to try again.  Such a semantic might be hard to support in hardware, but then hardware could actually cancel the buffer.

aingerson: Since this fix is just for the util_srx, I don't think there's any issue with disallowing the canceling.
@soumagne Can you update the PR to take the new approach to not allow cancellation of an in use mrecv buffer? (see above). Do you see any issues with your use case with this approach?

soumagne: >  I agree there can only be one completion with FI_MULTI_RECV set. Maybe we should ask why an app is wanting to cancel a multi-recv.

We pre-post a few multi-recv buffers to handle incoming RPC request messages. When we clean up, we cancel those posted recvs and free the buffers. From my perspective we just want to be able to get a guarantee that libfabric is no longer holding onto the multi-recv buffer so that we can free it. (Same as a standard recv from this perspective)

I think I'm fine with failing the cancel if the buffer is in use, ie. returning EAGAIN.

aingerson: Sweet, that sounds reasonable to me on both ends. Thanks!

shefty: Is the app keeping the endpoint that can access the buffers around?

soumagne: In this case we're not keeping the endpoint. We close it after.

soumagne:
@aingerson I updated the fix yesterday, I've let cancel return `-FI_ENOENT` if the entry is not found (it used to return FI_SUCCESS before), let me know if you prefer to keep the previous behavior.

shefty:
Cancel is an async operation.  We intentionally did not want cancel to return ENOENT and for apps to rely on that return value.  That error is only possible for SW based cancel.

aingerson:
@soumagne Yeah, what Sean said. FI_SUCCESS means the operation was submitted and can either result in a cancel error entry if it was cancelled or it would basically be ignored because the operation can't be canceled. In your app, you'd have to read the cq to see if an entry is available and basically retry the cancel until either you get the cancel entry for that recv or get the rx completion for it.
Also, could you squash these two commits together? They are needed together since the first one asserts the ref cnt is 0 and then the second one actually handles the non-zero ref cnt.
Thanks!

soumagne:
ok that sounds good, I fixed it to always return SUCCESS instead of ENOENT and I squashed the commits.

prov/util/src/util_srx.c
```diff
@@ -1037,12 +1041,15 @@ static bool util_cancel_recv(struct util_srx_ctx *srx, struct slist *queue,
 	slist_foreach(queue, item, prev) {
 		rx_entry = container_of(item, struct util_rx_entry, peer_entry);
 		if (rx_entry->peer_entry.context == context) {
+			if (rx_entry->peer_entry.flags & FI_MULTI_RECV &&
+			    rx_entry->multi_recv_ref)
+				return -FI_EAGAIN;
```
aingerson, line 1046: I'm still confused about this EAGAIN return. If we want to define the EAGAIN return as valid for a cancel, we should add it to the man page to clarify what it means for the application. As it stands right now in the definition, it feels like this should be an FI_SUCCESS because it was found and we dealt with it accordingly indicating to the application that there should be an entry on the CQ eventually (either the rx that is pending or the cancel entry) and once that is processed, the application can try to cancel again if it didn't get the completion for the entry. I'm fine with either, I just think we should clarify the expectation here.

shefty: My intent was that EAGAIN would be returned immediately to the user.  EAGAIN is a valid return code for any data transfer or async operation, so the app is already expected to deal with that.  EAGAIN does not indicate anything about whether the entry was found or its state.  From the app's viewpoint, it just says we can't deal with the cancel right now, and it needs to be retried.

EAGAIN should not be an error code returned through a completion.

soumagne: I also agree, in that particular case, unless I'm missing something I don't think we've dealt with it, we just found it and returned EAGAIN back to the caller of fi_cancel(), so it would never get canceled until it's retried.

Let me know if you'd like me to add any clarification anywhere. I can also add a comment in the code.

aingerson: Yeah, I understand the benefit of the EAGAIN and I'm fine with allowing it for this case. I just read FI_SUCCESS as essentially telling the app to read the CQ because something will be available. FI_SUCCESS just says I looked at your request and I did what I thought correct with it which for the mrecv case is not to cancel it and to let the pending receives complete. Ie the app would have to read the mrecv completion and decide to try to cancel it again because the buffer is not done. I think returning EAGAIN is simpler and that's totally fine. I would just like the man page (fi_endpoint) to reflect that return code and explain what that means for the application.
Right now it says
```
Returns 0 on success. On error, a negative value corresponding to fabric errno is returned. For fi_cancel, a return value of 0 indicates that the cancel request was submitted for processing.
```
So just adding something there about EAGAIN is totally fine

soumagne: ok I updated the man page and added a comment.

aingerson (approved):
Thank you for the updates!

j-xiong (approved):

j-xiong:
bot:aws:retest

j-xiong:
The Intel CI failure is caused by issue with HW used for dmabuf tests, which is unrelated to this PR.

@shijin-aws Is the AWS CI failure related?

shijin-aws:
```
2025-02-03 19:42:49,475 - INFO - test_runner_windows - Efawin compilation finished successfully.

Usage:

  C:/10734_1738611233\libfabric\x64\Release-v142\fi_pingpong.exe [OPTIONS]		start server

  C:/10734_1738611233\libfabric\x64\Release-v142\fi_pingpong.exe [OPTIONS] <srv_addr>	connect to server


Ping pong client and server


Options:

 -B <src_port>        source control port number (server: 47592, client: auto)

 -P <dst_port>        destination control port number (client: 47592)

 -d <domain>          domain name

 -s <source address>  source address associated with domain name

 -p <provider>        specific provider name eg sockets, verbs

 -e <ep_type>         endpoint type: msg|rdm|dgram (dgram)

 -I <number>          number of iterations (10)

 -S <size>            specific transfer size or 'all' (all)

 -c                   enables data_integrity checks

 -m <transmit mode>   transmit mode type: msg|tagged (msg)

 -h                   display this help output

 -v                   enable debugging output

 -6                   use IPv6 address

Usage:

  C:/10734_1738611233\libfabric\x64\Release-v142\fi_pingpong.exe [OPTIONS]		start server

  C:/10734_1738611233\libfabric\x64\Release-v142\fi_pingpong.exe [OPTIONS] <srv_addr>	connect to server


Ping pong client and server


Options:

 -B <src_port>        source control port number (server: 47592, client: auto)

 -P <dst_port>        destination control port number (client: 47592)

 -d <domain>          domain name

 -s <source address>  source address associated with domain name

 -p <provider>        specific provider name eg sockets, verbs

 -e <ep_type>         endpoint type: msg|rdm|dgram (dgram)

 -I <number>          number of iterations (10)

 -S <size>            specific transfer size or 'all' (all)

 -c                   enables data_integrity checks

 -m <transmit mode>   transmit mode type: msg|tagged (msg)

 -h                   display this help output

 -v                   enable debugging output

 -6                   use IPv6 address

2025-02-03 19:42:59,612 - INFO - test_runner_windows - Server process failed with return code 1


...
```

@soumagne can you rebase your PR to the latest main branch that has https://github.com/ofiwg/libfabric/commit/67932605d4a58cea495b50504d57eaa73543256e

PR #10735, opened 2025-01-28
prov/efa: Add new efa-direct fi_info objects

sunkuamzn:
efa-direct is a new provider that provides direct access to the EFA
device with minimal overhead and features. It does not support SAS
ordering, tagged messaging, atomics. It does not have  copy based
protocols and cannot support sending/receiving large message sizes
and HMEM devices that don't have p2p support.

The functionality provided by the efa-direct provider is a subset of the
efa rdm provider. When applications do not require  capabilities only
supported by the efa rdm provider, it is more performant for them to use
the efa-direct provider. Accordingly, the efa-direct fi_info objects are
returned before the efa rdm fi_info objects in the fi_getinfo call.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -28,7 +28,7 @@ static struct fi_ops efa_ops_domain_fid = {
 	.ops_open = efa_domain_ops_open,
 };
 
-static struct fi_ops_domain efa_ops_domain_dgram = {
+static struct fi_ops_domain efa_domain_ops = {
 	.size = sizeof(struct fi_ops_domain),
 	.av_open = efa_av_open,
 	.cq_open = efa_cq_open,
@@ -42,7 +42,7 @@ static struct fi_ops_domain efa_ops_domain_dgram = {
 	.query_collective = fi_no_query_collective,
 };
 
-static struct fi_ops_domain efa_ops_domain_rdm = {
+static struct fi_ops_domain efa_domain_ops_rdm = {
 	.size = sizeof(struct fi_ops_domain),
 	.av_open = efa_av_open,
 	.cq_open = efa_rdm_cq_open,
@@ -230,8 +230,8 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	}
 
 	efa_domain->mr_local = ofi_mr_local(info);
-	if (EFA_EP_TYPE_IS_DGRAM(info) && !efa_domain->mr_local) {
-		EFA_WARN(FI_LOG_EP_DATA, "dgram require FI_MR_LOCAL, but application does not support it\n");
+	if ((EFA_INFO_TYPE_IS_DGRAM(info) || EFA_INFO_TYPE_IS_DIRECT(info)) && !efa_domain->mr_local) {
+		EFA_WARN(FI_LOG_EP_DATA, "EFA direct and dgram require FI_MR_LOCAL, but application does not support it\n");
 		ret = -FI_ENODATA;
 		goto err_free;
 	}
@@ -274,7 +274,7 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	}
 
 	efa_domain->util_domain.domain_fid.fid.ops = &efa_ops_domain_fid;
-	if (EFA_EP_TYPE_IS_RDM(info)) {
+	if (EFA_INFO_TYPE_IS_RDM(info)) {
 		err = efa_domain_init_rdm(efa_domain, info);
 		if (err) {
 			EFA_WARN(FI_LOG_DOMAIN,
@@ -282,10 +282,10 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 				 -err);
 			goto err_free;
 		}
-		efa_domain->util_domain.domain_fid.ops = &efa_ops_domain_rdm;
+		efa_domain->util_domain.domain_fid.ops = &efa_domain_ops_rdm;
 	} else {
-		assert(EFA_EP_TYPE_IS_DGRAM(info));
-		efa_domain->util_domain.domain_fid.ops = &efa_ops_domain_dgram;
+		assert(EFA_INFO_TYPE_IS_DGRAM(info) || EFA_INFO_TYPE_IS_DIRECT(info));
+		efa_domain->util_domain.domain_fid.ops = &efa_domain_ops;
 	}
 
 #ifndef _WIN32
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -11,7 +11,7 @@ struct efa_hmem_info g_efa_hmem_info[OFI_HMEM_MAX];
 static size_t efa_max_eager_msg_size_with_largest_header() {
 	int mtu_size;
 
-	mtu_size = g_device_list[0].rdm_info->ep_attr->max_msg_size;
+	mtu_size = g_device_list[0].ibv_port_attr.max_msg_sz;
 
 	return mtu_size - efa_rdm_pkt_type_get_max_hdr_size();
 }
--- a/prov/efa/src/efa_prov.c
+++ b/prov/efa/src/efa_prov.c
@@ -3,6 +3,7 @@
 
 #include <ofi_prov.h>
 #include "efa.h"
+#include "efa_prov.h"
 #include "efa_prov_info.h"
 #include "efa_env.h"
 
@@ -67,7 +68,6 @@ struct fi_provider efa_prov = {
 
 struct util_prov efa_util_prov = {
 	.prov = &efa_prov,
-	.flags = 0,
 };
 
 /**
@@ -79,10 +79,35 @@ struct util_prov efa_util_prov = {
 static int efa_util_prov_initialize()
 {
 	int i, err;
-	struct fi_info *head, *tail, *prov_info_rdm, *prov_info_dgram;
+	struct fi_info *head, *tail, *prov_info_rdm, *prov_info_dgram, *prov_info_direct;
 
 	head = NULL;
 	tail = NULL;
+
+	/*
+	* EFA direct provider is more performant if the application can use it
+	* Therefore, the efa-direct info objects should be returned _before_ efa rdm or dgram
+	* So we populate the efa-direct info objects first
+	*/
+	for (i = 0; i < g_device_cnt; ++i) {
+		prov_info_direct = fi_dupinfo(g_device_list[i].rdm_info);
+		if (!prov_info_direct) {
+			EFA_WARN(FI_LOG_DOMAIN, "Failed to allocate prov_info for EFA direct\n");
+			continue;
+		}
+
+		efa_prov_info_set_fabric_name(prov_info_direct, EFA_DIRECT_FABRIC_NAME);
+
+		if (!head) {
+			head = prov_info_direct;
+		} else {
+			assert(tail);
+			tail->next = prov_info_direct;
+		}
+
+		tail = prov_info_direct;
+	}
+
 	for (i = 0; i < g_device_cnt; ++i) {
 		err = efa_prov_info_alloc_for_rdm(&prov_info_rdm, &g_device_list[i]);
 		if (err) {
@@ -90,6 +115,8 @@ static int efa_util_prov_initialize()
 				 err);
 			continue;
 		}
+
+		efa_prov_info_set_fabric_name(prov_info_rdm, EFA_FABRIC_NAME);
 
 		if (!head) {
 			head = prov_info_rdm;
@@ -107,6 +134,8 @@ static int efa_util_prov_initialize()
 			EFA_WARN(FI_LOG_DOMAIN, "Failed to allocate prov_info for dgram\n");
 			continue;
 		}
+
+		efa_prov_info_set_fabric_name(prov_info_dgram, EFA_FABRIC_NAME);
 
 		if (!head) {
 			head = prov_info_dgram;
@@ -160,6 +189,14 @@ EFA_INI
 	 */
 	efa_env_initialize();
 
+	err = efa_hmem_info_initialize();
+	if (err)
+		goto err_free;
+
+	/*
+	* efa_util_prov_initialize uses g_efa_hmem_info, so it
+	* must be called after efa_hmem_info_initialize
+	*/
 	err = efa_util_prov_initialize();
 	if (err)
 		goto err_free;
@@ -168,10 +205,6 @@ EFA_INI
 	if (err) {
 		goto err_free;
 	}
-
-	err = efa_hmem_info_initialize();
-	if (err)
-		goto err_free;
 
 	dlist_init(&g_efa_domain_list);
 
@@ -202,4 +235,3 @@ static void efa_prov_finalize(void)
 	ofi_mem_fini();
 #endif
 }
-
--- a/prov/efa/src/efa_prov_info.c
+++ b/prov/efa/src/efa_prov_info.c
@@ -602,7 +602,7 @@ int efa_prov_info_alloc_for_rdm(struct fi_info **prov_info_rdm_ptr,
 		/*
 		 * EFA RDM endpoint implemented emulated atomic, hence set atomic size
 		 */
-		max_atomic_size = device->rdm_info->ep_attr->max_msg_size
+		max_atomic_size = device->ibv_port_attr.max_msg_sz
 					- sizeof(struct efa_rdm_rta_hdr)
 					- device->rdm_info->src_addrlen
 					- EFA_RDM_IOV_LIMIT * sizeof(struct fi_rma_iov);
@@ -637,10 +637,10 @@ int efa_prov_info_alloc_for_rdm(struct fi_info **prov_info_rdm_ptr,
 		 *    pkt_entry_size - maximum_header_size.
 		 */
 		if (efa_env.enable_shm_transfer)
-			min_pkt_size = MIN(device->rdm_info->ep_attr->max_msg_size - efa_rdm_pkt_type_get_max_hdr_size(),
+			min_pkt_size = MIN(device->ibv_port_attr.max_msg_sz - efa_rdm_pkt_type_get_max_hdr_size(),
 					   SHM_MAX_INJECT_SIZE);
 		else
-			min_pkt_size = device->rdm_info->ep_attr->max_msg_size - efa_rdm_pkt_type_get_max_hdr_size();
+			min_pkt_size = device->ibv_port_attr.max_msg_sz - efa_rdm_pkt_type_get_max_hdr_size();
 
 		if (min_pkt_size < efa_rdm_pkt_type_get_max_hdr_size()) {
 			prov_info_rdm->tx_attr->inject_size = 0;
--- a/prov/efa/src/efa_prov_info.c
+++ b/prov/efa/src/efa_prov_info.c
@@ -9,22 +9,19 @@
 #include "rdm/efa_rdm_pkt_type.h"
 
 #define EFA_FABRIC_PREFIX "EFA-"
-#define EFA_FABRIC_NAME "efa"
 #define EFA_DOMAIN_CAPS (FI_LOCAL_COMM | FI_REMOTE_COMM)
 
 #define EFA_RDM_TX_CAPS (OFI_TX_MSG_CAPS)
 #define EFA_RDM_RX_CAPS (OFI_RX_MSG_CAPS | FI_SOURCE)
+#define EFA_RDM_CAPS (EFA_RDM_TX_CAPS | EFA_RDM_RX_CAPS | EFA_DOMAIN_CAPS)
+
 #define EFA_DGRM_TX_CAPS (OFI_TX_MSG_CAPS)
 #define EFA_DGRM_RX_CAPS (OFI_RX_MSG_CAPS | FI_SOURCE)
-#define EFA_RDM_CAPS (EFA_RDM_TX_CAPS | EFA_RDM_RX_CAPS | EFA_DOMAIN_CAPS)
 #define EFA_DGRM_CAPS (EFA_DGRM_TX_CAPS | EFA_DGRM_RX_CAPS | EFA_DOMAIN_CAPS)
 
 #define EFA_TX_OP_FLAGS (FI_TRANSMIT_COMPLETE)
-
-#define EFA_RX_MODE (0)
-
-#define EFA_RX_RDM_OP_FLAGS (0)
-#define EFA_RX_DGRM_OP_FLAGS (0)
+#define EFA_RDM_TX_OP_FLAGS (FI_INJECT | FI_COMPLETION | FI_TRANSMIT_COMPLETE | \
+							FI_DELIVERY_COMPLETE)
 
 #define EFA_MSG_ORDER (0)
 
@@ -39,30 +36,6 @@ const struct fi_fabric_attr efa_fabric_attr = {
 	.prov_name	= NULL,
 	.prov_version	= OFI_VERSION_DEF_PROV,
 };
-
-/**
- * @brief set the fabric_attr field of a prov_info
- *
- * @param 	prov_info[out]		prov_info object
- * @param	device[in]		pointer to an efa_device struct, which contains device attributes
- * @return	0 on sucessess
- * 		-FI_ENOMEM if memory allocation failed
- */
-static
-int efa_prov_info_set_fabric_attr(struct fi_info *prov_info, struct efa_device *device)
-{
-	size_t name_len = strlen(EFA_FABRIC_NAME);
-
-	*prov_info->fabric_attr	= efa_fabric_attr;
-
-	name_len = strlen(EFA_FABRIC_NAME);
-	prov_info->fabric_attr->name = calloc(1, name_len + 1);
-	if (!prov_info->fabric_attr->name)
-		return -FI_ENOMEM;
-
-	strcpy(prov_info->fabric_attr->name, EFA_FABRIC_NAME);
-	return 0;
-}
 
 /**
  * @brief default domain_attr for prov_info
@@ -122,6 +95,15 @@ int efa_prov_info_set_domain_attr(struct fi_info *prov_info,
 	prov_info->domain_attr->max_ep_rx_ctx	= 1;
 	prov_info->domain_attr->resource_mgmt	= FI_RM_DISABLED;
 	prov_info->domain_attr->mr_cnt		= device->ibv_attr.max_mr;
+
+	if (ep_type == FI_EP_RDM) {
+		/* EFA direct path is thread safe */
+		prov_info->domain_attr->threading = FI_THREAD_SAFE;
+
+		/* EFA direct path retries indefinitely when Receiver Not Ready (RNR) */
+		prov_info->domain_attr->resource_mgmt = FI_RM_ENABLED;
+	}
+
 	EFA_DBG(FI_LOG_DOMAIN, "Domain attribute :\n"
 				"\t prov_info->domain_attr->cq_cnt		= %zu\n"
 				"\t prov_info->domain_attr->ep_cnt		= %zu\n"
@@ -167,31 +149,18 @@ void efa_prov_info_set_ep_attr(struct fi_info *prov_info,
 {
 
 	*prov_info->ep_attr = efa_ep_attr;
-	if (ep_type == FI_EP_DGRAM) {
-		prov_info->mode |= FI_MSG_PREFIX;
+	prov_info->ep_attr->protocol	= FI_PROTO_EFA;
+	prov_info->ep_attr->max_msg_size = device->ibv_port_attr.max_msg_sz;
+	prov_info->ep_attr->type = ep_type;
+
+	if (ep_type == FI_EP_RDM) {
+		/* ep_attr->max_msg_size is the maximum of both MSG and RMA operations */
+		if (prov_info->caps & FI_RMA)
+			prov_info->ep_attr->max_msg_size = MAX(device->ibv_port_attr.max_msg_sz, device->max_rdma_size);
+	} else {
+		assert(ep_type == FI_EP_DGRAM);
 		prov_info->ep_attr->msg_prefix_size = 40;
 	}
-
-	prov_info->ep_attr->protocol	= FI_PROTO_EFA;
-	prov_info->ep_attr->type	= ep_type;
-
-	if (prov_info->ep_attr->type == FI_EP_RDM) {
-		prov_info->tx_attr->inject_size = device->efa_attr.inline_buf_size;
-	} else {
-		assert(prov_info->ep_attr->type == FI_EP_DGRAM);
-                /*
-		 * Currently, there is no mechanism for device to discard
-		 * a completion, therefore there is no way for dgram endpoint
-		 * to implement FI_INJECT. Because FI_INJECT is not an optional
-		 * feature, we had to set inject_size to 0.
-		 * 
-		 * TODO:
-		 * Remove this after implementing cq read for efa-raw
-                 */
-		prov_info->tx_attr->inject_size = 0;
-	}
-
-	prov_info->ep_attr->max_msg_size		= device->ibv_port_attr.max_msg_sz;
 }
 
 /**
@@ -211,8 +180,7 @@ const struct fi_tx_attr efa_dgrm_tx_attr = {
  */
 const struct fi_rx_attr efa_dgrm_rx_attr = {
 	.caps			= EFA_DGRM_RX_CAPS,
-	.mode			= FI_MSG_PREFIX | EFA_RX_MODE,
-	.op_flags		= EFA_RX_DGRM_OP_FLAGS,
+	.mode			= FI_MSG_PREFIX,
 	.msg_order		= EFA_MSG_ORDER,
 	.iov_limit		= 1
 };
@@ -234,8 +202,6 @@ const struct fi_tx_attr efa_rdm_tx_attr = {
  */
 const struct fi_rx_attr efa_rdm_rx_attr = {
 	.caps			= EFA_RDM_RX_CAPS,
-	.mode			= EFA_RX_MODE,
-	.op_flags		= EFA_RX_RDM_OP_FLAGS,
 	.msg_order		= EFA_MSG_ORDER,
 	.iov_limit		= 1
 };
@@ -254,12 +220,26 @@ void efa_prov_info_set_tx_rx_attr(struct fi_info *prov_info,
 	if (ep_type == FI_EP_RDM) {
 		*prov_info->tx_attr	= efa_rdm_tx_attr;
 		*prov_info->rx_attr	= efa_rdm_rx_attr;
+
+		prov_info->tx_attr->op_flags |= EFA_RDM_TX_OP_FLAGS;
+		prov_info->rx_attr->op_flags |= FI_COMPLETION;
+
+		/* If the device supports FI_RMA, also set tx_attr and rx_attr */
+		if (prov_info->caps & FI_RMA) {
+			prov_info->tx_attr->caps |= OFI_TX_RMA_CAPS;
+			prov_info->rx_attr->caps |= OFI_RX_RMA_CAPS;
+		}
 	} else {
 		assert(ep_type == FI_EP_DGRAM);
 		*prov_info->tx_attr	= efa_dgrm_tx_attr;
 		*prov_info->rx_attr	= efa_dgrm_rx_attr;
 	}
 
+	/* efa-direct and DGRAM paths require FI_CONTEXT2 */
+	prov_info->tx_attr->mode |= FI_CONTEXT2;
+	prov_info->rx_attr->mode |= FI_CONTEXT2;
+
+	prov_info->tx_attr->inject_size = device->efa_attr.inline_buf_size;
 	prov_info->tx_attr->iov_limit = device->efa_attr.max_sq_sge;
 	prov_info->tx_attr->size = rounddown_power_of_two(device->efa_attr.max_sq_wr);
 	prov_info->rx_attr->iov_limit = device->efa_attr.max_rq_sge;
@@ -375,7 +355,7 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
 		goto err_free;
 	}
 
-	link_attr->mtu = device->ibv_port_attr.max_msg_sz - efa_rdm_pkt_type_get_max_hdr_size();
+	link_attr->mtu = device->ibv_port_attr.max_msg_sz;
 	link_attr->speed = ofi_vrb_speed(device->ibv_port_attr.active_speed,
 	                                 device->ibv_port_attr.active_width);
 
@@ -409,12 +389,46 @@ err_free:
 }
 
 #if HAVE_CUDA || HAVE_NEURON || HAVE_SYNAPSEAI
-void efa_prov_info_set_hmem_flags(struct fi_info *prov_info)
+void efa_prov_info_set_hmem_flags(struct fi_info *prov_info, enum fi_ep_type ep_type)
 {
-	if (prov_info->ep_attr->type == FI_EP_RDM &&
-	    (ofi_hmem_is_initialized(FI_HMEM_CUDA) ||
+	int i;
+	enum fi_hmem_iface iface;
+	struct efa_hmem_info *hmem_info;
+	bool enable_hmem = false;
+
+	if (ep_type != FI_EP_RDM)
+		return;
+
+	/* EFA direct only supports HMEM when p2p support is available */
+	if ((ofi_hmem_is_initialized(FI_HMEM_CUDA) ||
 	     ofi_hmem_is_initialized(FI_HMEM_NEURON) ||
 	     ofi_hmem_is_initialized(FI_HMEM_SYNAPSEAI))) {
+
+		/* Check if FI_HMEM_P2P_DISABLED is set */
+		if (ofi_hmem_p2p_disabled()) {
+			EFA_WARN(FI_LOG_CORE,
+				"FI_HMEM capability requires peer to peer "
+				"support, which is disabled because "
+				"FI_HMEM_P2P_DISABLED was set to 1/on/true.\n");
+			return;
+		}
+
+		EFA_HMEM_IFACE_FOREACH(i) {
+			iface = efa_hmem_ifaces[i];
+			hmem_info = &g_efa_hmem_info[iface];
+			if (hmem_info->initialized && !hmem_info->p2p_supported_by_device) {
+				EFA_INFO(FI_LOG_CORE,
+					"EFA direct provider was compiled with support for %s HMEM interface "
+					"but the interface does not support p2p transfers. "
+					"EFA direct provider does not support HMEM transfers without p2p support. "
+					"HMEM support will be disabled.\n", fi_tostr(&iface, FI_TYPE_HMEM_IFACE));
+					return;
+			}
+		}
+		enable_hmem = true;
+	}
+
+	if (enable_hmem) {
 		prov_info->caps			|= FI_HMEM;
 		prov_info->tx_attr->caps		|= FI_HMEM;
 		prov_info->rx_attr->caps		|= FI_HMEM;
@@ -422,7 +436,7 @@ void efa_prov_info_set_hmem_flags(struct fi_info *prov_info)
 	}
 }
 #else
-void efa_prov_info_set_hmem_flags(struct fi_info *prov_info)
+void efa_prov_info_set_hmem_flags(struct fi_info *prov_info, enum fi_ep_type ep_type)
 {
 }
 #endif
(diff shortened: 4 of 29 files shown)


prov/efa/src/efa_base_ep.h
```diff
@@ -39,6 +39,28 @@
 #define efa_rx_flags(efa_base_ep) ((efa_base_ep)->util_ep.rx_op_flags)
 #define efa_tx_flags(efa_base_ep) ((efa_base_ep)->util_ep.tx_op_flags)
 
+#define EFA_PROV_NAME "efa"
+#define EFA_DIRECT_PROV_NAME "efa-direct"
+
+/* TODO - CHECK THESE */
+#define EFA_EP_TYPE_IS_RDM(_info) \
+	(_info && _info->ep_attr && (_info->ep_attr->type == FI_EP_RDM) && !strcasecmp(_info->fabric_attr->name, EFA_PROV_NAME))
```
shijin-aws, line 47: Both efa and efa-direct has rdm ep type. Making this macro for efa only seems goofy to me.

prov/efa/src/efa_base_ep.h
```diff
@@ -39,6 +39,28 @@
...
+	(_info && _info->ep_attr && (_info->ep_attr->type == FI_EP_RDM) && !strcasecmp(_info->fabric_attr->name, EFA_DIRECT_PROV_NAME))
+
+/* TODO - CHECK THESE */
+#define EFA_EP_TYPE_IS_DGRAM(_info) \
+	(_info && _info->ep_attr && (_info->ep_attr->type == FI_EP_DGRAM))
+
+enum efa_ep_type {
+	EFA_EP_RDM = 0,
+	EFA_EP_DGRAM,
+	EFA_EP_DIRECT,
```
shijin-aws, line 59: efa direct shouldn't be an ep type.

prov/efa/src/efa_device.h
```diff
@@ -21,6 +21,7 @@ struct efa_device {
 	struct ibv_pd		*ibv_pd;
 	struct fi_info		*rdm_info;
 	struct fi_info		*dgram_info;
+	struct fi_info     *direct_info;
```
shijin-aws, line 24: indent, also I don't think direct_info should be a separate field in `efa_device`.

prov/efa/src/efa_domain.c
```diff
@@ -230,19 +233,31 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
...
+	err = efa_domain_init_device_and_pd(efa_domain, info->domain_attr->name, efa_ep_type);
 	if (err) {
 		ret = err;
 		goto err_free;
 	}
 
-	efa_domain->info = fi_dupinfo(EFA_EP_TYPE_IS_RDM(info) ? efa_domain->device->rdm_info : efa_domain->device->dgram_info);
+	switch (efa_ep_type) {
+		case EFA_EP_RDM:
+			efa_domain->info = fi_dupinfo(efa_domain->device->rdm_info);
```
shijin-aws, line 250: there shouldn't be a difference between `device->rdm_info` and `device->direct_info`. If we have anything today in `device->rdm_info` that doesn't reflect device capabilities, it is wrong and we need to fix.

prov/efa/src/efa_domain.h
```diff
@@ -71,10 +72,18 @@ static inline bool efa_is_cache_available(struct efa_domain *efa_domain)
...
  */
 static inline
-const char *efa_domain_name_suffix(enum fi_ep_type ep_type)
+const char *efa_domain_name_suffix(enum efa_ep_type ep_type)
 {
-	assert(ep_type == FI_EP_RDM || ep_type == FI_EP_DGRAM);
-	return (ep_type == FI_EP_RDM) ? "-rdm" : "-dgrm";
+	switch (ep_type) {
+		case EFA_EP_RDM:
+			return "-rdm";
```
shijin-aws, line 79: It is a bit odd to me, I thought both direct and non-direct domain can have the same domain name.

prov/efa/src/efa_prov.c
```diff
@@ -175,6 +228,32 @@ EFA_INI
...
+}
+
+EFA_INI
+{
+	int err;
+
+	if (g_device_cnt <= 0)
+		return NULL;
+
+	// The device list and HMEM interfaces are initialized
```
shijin-aws, line 245: Multi-line comments should use `/**/`. similar for other places.

prov/efa/src/efa_prov_info.c
```diff
@@ -187,8 +186,24 @@ void efa_prov_info_set_ep_attr(struct fi_info *prov_info,
...
+		case EFA_EP_DIRECT:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			prov_info->tx_attr->inject_size = device->efa_attr.inline_buf_size;
+			prov_info->mode |= FI_CONTEXT2; 	/* EFA direct provider requires FI_CONTEXT2 mode */
+			break;
+		case EFA_EP_DGRAM:
+			prov_info->ep_attr->type = FI_EP_DGRAM;
+			prov_info->mode |= FI_MSG_PREFIX;
+			prov_info->ep_attr->msg_prefix_size = 40;
+			prov_info->tx_attr->inject_size = 0;
```
shijin-aws, line 203: With recent work, we can have the same inject size for dgram now

prov/efa/src/efa_hmem.c
```diff
@@ -238,6 +238,27 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
 	return;
 }
 
+static inline void
+efa_hmem_info_popuplate_iface_string(enum fi_hmem_iface iface, struct efa_hmem_info *info) {
```
shijin-aws, line 242: there is already an `fi_tostr` for hmem interface: `fi_tostr(&iface, FI_TYPE_HMEM_IFACE)`

prov/efa/src/efa_hmem.c
```diff
@@ -238,6 +238,27 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
...
+	switch (iface) {
+		case FI_HMEM_SYSTEM:
+			info->iface_name = "system";
```
shijin-aws, line 245: And you cannot assign the string barely here because it will be freed outside the function. For that you need to use `strdup` to alloc a new mem and free them later. For all of the complexity I don't think you need this function

prov/efa/src/efa_hmem.h
```diff
@@ -22,6 +22,7 @@ static const enum fi_hmem_iface efa_hmem_ifaces[] = {
 };
 
 struct efa_hmem_info {
+	char* iface_name;
```
shijin-aws, line 25: same comment, I don't think you need this field

prov/efa/src/efa_prov_info.c
```diff
@@ -455,20 +455,43 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+void efa_prov_info_set_hmem_flags(struct fi_info *prov_info, enum efa_ep_type ep_type)
 {
-	if (prov_info->ep_attr->type == FI_EP_RDM &&
-	    (ofi_hmem_is_initialized(FI_HMEM_CUDA) ||
+	bool enable_hmem = false;
+
+	if ((ofi_hmem_is_initialized(FI_HMEM_CUDA) ||
 	     ofi_hmem_is_initialized(FI_HMEM_NEURON) ||
 	     ofi_hmem_is_initialized(FI_HMEM_SYNAPSEAI))) {
+			if ep_type == EFA_EP_RDM
```
shijin-aws, line 465: don't you need a parenthesis here?

prov/efa/src/efa_prov_info.c
```diff
@@ -455,20 +455,43 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+				enable_hmem = true;
```
shijin-aws, line 466: do you need an `else` here?

prov/efa/src/efa_prov_info.c
```diff
@@ -455,20 +455,43 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+
+			if ep_type == EFA_EP_DIRECT {
```
shijin-aws, line 468: do you need parenthesis?

prov/efa/src/efa_prov_info.c
```diff
@@ -455,20 +455,43 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+				/* EFA direct only supports HMEM when p2p support is available */
+				EFA_HMEM_IFACE_FOREACH(i) {
+					iface = efa_hmem_ifaces[i];
+					hmem_info = &g_efa_hmem_info[iface];
+					if (hmem_info->initialized && !hmem_info->p2p_supported_by_device) {
+						EFA_WARN(FI_LOG_CORE,
+							"EFA direct provider was compiled with support for %s HMEM interface "
+							"but the interface does not support p2p transfers. "
+							"EFA direct provider does support HMEM transfers without p2p support. "
+							"HMEM support will be disabled.\n", hmem_info->iface_name);
```
shijin-aws, line 478: do you need a break here??

prov/efa/src/efa_user_info.c
```diff
@@ -332,7 +332,99 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
...
+		 * Same goes for prefix mode, where the protocol does not
+		 * absolutely need a prefix before receive buffers, but it can
+		 * use it when available to optimize transfers with endpoints
+		 * having the following profile:
+		 *	- Requires FI_MSG and not FI_TAGGED/FI_ATOMIC/FI_RMA
+		 *	- Can handle registrations (FI_MR_LOCAL)
+		 *	- No need for FI_DIRECTED_RECV
+		 *	- Guaranteed to send msgs smaller than info->nic->link_attr->mtu
+		 */
+		if (hints->mode & FI_MSG_PREFIX) {
```
shijin-aws, line 399: I am not sure on this. efa-direct rdm doesn't require prefix. If application supports FI_MSG_PREFIX, we can just ignore it and return an zero prefix. Also the current efa_msg.c code doesn't actually handle prefix for rdm ep type.

shijin-aws (changes requested):
My high level comment is that I don't think efa direct should be a separate ep type. I understand you want to use this to simplify your branching. But the branching should be on prov_name instead of ep type. I thought our plan is to make efa direct cover both dgram and rdm ep type. But that will break existing dgram user which use efa as prov_name. I don't know if this user actually exists, but to avoid the risk I am ok to make dgram ep type stay in `efa` and make `efa-direct` only cover `rdm`.

prov/efa/src/efa_user_info.c
sunkuamzn: fixed

prov/efa/src/efa_domain.c
sunkuamzn: `fabric_attr->name` is different. `prov_info->mode` for efa-direct requires `FI_CONTEXT2`. `prov_info->tx_attr->msg_order` and `prov_info->rx_attr->msg_order` are different. HMEM capabilities are also different. `device->rdm_info` will support HMEM even without p2p.

We can do all of this in the user_info step but that's not how it is done now.

prov/efa/src/efa_base_ep.h
sunkuamzn: Discussed offline. Will change this enum to `efa_info_type`

prov/efa/src/efa_domain.c
sunkuamzn: Discussed offline. Refactoring to move logic to user_info will be done later.

prov/efa/src/efa_device.h
sunkuamzn: Discussed offline. Refactoring to move logic to user_info will be done later.

prov/efa/src/efa_domain.h
sunkuamzn: Will change domain name suffix for efa-direct to `-rdm`

prov/efa/src/efa_base_ep.h
sunkuamzn: Changed to `EFA_INFO_TYPE_IS_RDM`

prov/efa/src/efa_prov_info.c
```diff
@@ -158,37 +164,33 @@ const struct fi_ep_attr efa_ep_attr = {
...
+
+	switch (info_type) {
+		case EFA_INFO_RDM:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			break;
+		case EFA_INFO_DIRECT:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			prov_info->mode |= FI_CONTEXT2; 	/* EFA direct provider requires FI_CONTEXT2 mode */
+			break;
+		case EFA_INFO_DGRAM:
```
shijin-aws, line 187: dgram should require FI_CONTEXT2 now as well. as we have the same efa-direct code for dgram

prov/efa/src/efa_prov_info.c
```diff
@@ -158,37 +164,33 @@ const struct fi_ep_attr efa_ep_attr = {
...
-		 * Currently, there is no mechanism for device to discard
-		 * a completion, therefore there is no way for dgram endpoint
-		 * to implement FI_INJECT. Because FI_INJECT is not an optional
-		 * feature, we had to set inject_size to 0.
-		 * 
-		 * TODO:
-		 * Remove this after implementing cq read for efa-raw
-                 */
-		prov_info->tx_attr->inject_size = 0;
+	prov_info->tx_attr->inject_size = device->efa_attr.inline_buf_size;
```
shijin-aws, line 177: your function is setting ep attr but you are setting tx attr here?

prov/efa/src/efa_prov_info.c
```diff
@@ -158,37 +164,33 @@ const struct fi_ep_attr efa_ep_attr = {
...
+		case EFA_INFO_RDM:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			break;
+		case EFA_INFO_DIRECT:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			prov_info->mode |= FI_CONTEXT2; 	/* EFA direct provider requires FI_CONTEXT2 mode */
+			break;
+		case EFA_INFO_DGRAM:
+			prov_info->ep_attr->type = FI_EP_DGRAM;
+			prov_info->mode |= FI_MSG_PREFIX;
```
shijin-aws, line 189: this is not ep attribute either

prov/efa/src/efa_prov_info.c
shijin-aws: bump comment

prov/efa/src/efa_prov_info.c
```diff
@@ -442,20 +442,47 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+				EFA_HMEM_IFACE_FOREACH(i) {
+					iface = efa_hmem_ifaces[i];
+					hmem_info = &g_efa_hmem_info[iface];
+					if (hmem_info->initialized && !hmem_info->p2p_supported_by_device) {
+						EFA_WARN(FI_LOG_CORE,
+							"EFA direct provider was compiled with support for %s HMEM interface "
+							"but the interface does not support p2p transfers. "
+							"EFA direct provider does not support HMEM transfers without p2p support. "
+							"HMEM support will be disabled.\n", fi_tostr(&iface, FI_TYPE_HMEM_IFACE));
+							break;
```
shijin-aws, line 469: indent should be aligned with EFA_WARN.

prov/efa/src/efa_prov.c
```diff
@@ -125,16 +140,54 @@ static int efa_util_prov_initialize()
...
+	efa_direct_util_prov.info = head;
+	return 0;
+}
+
+static void efa_util_prov_finalize(struct util_prov *util_prov)
 {
 	struct fi_info *prov_info;
 
-	prov_info = (struct fi_info *)efa_util_prov.info;
+	prov_info = (struct fi_info *)(*util_prov).info;
```
darrylabbate, line 185: nit
```suggestion
	prov_info = (struct fi_info *) util_prov->info;
```

prov/efa/src/efa_prov.c
```diff
@@ -125,16 +140,54 @@ static int efa_util_prov_initialize()
...
 	fi_freeinfo(prov_info);
-	efa_util_prov.info = NULL;
+	(*util_prov).info = NULL;
```
darrylabbate, line 187: likewise
```suggestion
	util_prov->info = NULL;
```

prov/efa/src/efa_prov.c
```diff
@@ -70,6 +71,20 @@ struct util_prov efa_util_prov = {
...
+	.version = OFI_VERSION_DEF_PROV,
+	.fi_version = OFI_VERSION_LATEST,
+	.getinfo = efa_direct_getinfo,
+	.fabric = efa_fabric,
+	.cleanup = efa_direct_prov_finalize
+};
+
+struct util_prov efa_direct_util_prov = {
+	.prov = &efa_direct_prov,
+	.flags = 0,
```
darrylabbate, line 85: nit: superfluous (implicit zero initialization)

prov/efa/src/efa_fabric.c
```diff
@@ -82,13 +82,19 @@ int efa_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric_fid,
...
 	if (!efa_fabric)
 		return -FI_ENOMEM;
 
-	for (info = efa_util_prov.info; info; info = info->next) {
+	if (!strcmp(attr->name, EFA_DIRECT_PROV_NAME))
+		util_prov = &efa_direct_util_prov;
+	else
+		util_prov = &efa_util_prov;
+
+	for (info = (*util_prov).info; info; info = info->next) {
```
darrylabbate, line 97:
```suggestion
	for (info = util_prov->info; info; info = info->next) {
```

prov/efa/src/efa_base_ep.h
```diff
@@ -39,6 +39,26 @@
 #define efa_rx_flags(efa_base_ep) ((efa_base_ep)->util_ep.rx_op_flags)
 #define efa_tx_flags(efa_base_ep) ((efa_base_ep)->util_ep.tx_op_flags)
 
+#define EFA_PROV_NAME "efa"
+#define EFA_DIRECT_PROV_NAME "efa-direct"
+
+#define EFA_INFO_TYPE_IS_RDM(_info) \
+	(_info && _info->ep_attr && (_info->ep_attr->type == FI_EP_RDM) && !strcasecmp(_info->fabric_attr->name, EFA_PROV_NAME))
```
darrylabbate, line 46: Are the linear string comparisons necessary? These are already pretty gnarly chained conditionals.

If the `fi_info.fabric_attr.name` is truly the only way to disambiguate efa-rdm from efa-direct (I.e. we can't add some other enum field elsewhere), I'd like to explore employing constant-time comparison schemes.

prov/efa/src/efa_prov_info.c
```diff
@@ -235,7 +237,30 @@ const struct fi_tx_attr efa_rdm_tx_attr = {
...
+
+/**
+ * @brief default TX attributes for efa-direct end point
+ */
+const struct fi_tx_attr efa_direct_tx_attr = {
+	.caps			= EFA_RDM_TX_CAPS,
+	.mode			= 0,
+	.op_flags		= EFA_TX_OP_FLAGS,
+	.msg_order		= EFA_MSG_ORDER,
+	.inject_size		= 0,
```
shijin-aws, line 253: If we cannot assign inject_size to a valid value here, I would just drop it.

prov/efa/src/efa_prov_info.c
sunkuamzn: done

prov/efa/src/efa_base_ep.h
sunkuamzn: Is `strncmp` with the length of the longer string good enough? These macros are only used in the `fi_getinfo` and `fi_domain` calls, so they are not performance critical.

prov/efa/src/efa_fabric.c
```diff
@@ -82,13 +82,15 @@ int efa_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric_fid,
...
+	struct util_prov *util_prov;
 	int ret = 0;
 
 	efa_fabric = calloc(1, sizeof(*efa_fabric));
 	if (!efa_fabric)
 		return -FI_ENOMEM;
 
-	for (info = efa_util_prov.info; info; info = info->next) {
+	util_prov = &efa_util_prov;
+	for (info = util_prov->info; info; info = info->next) {
```
shijin-aws, line 93: unnecessary change?

prov/efa/src/efa_prov_info.c
```diff
@@ -49,18 +50,23 @@ const struct fi_fabric_attr efa_fabric_attr = {
...
  */
 static
-int efa_prov_info_set_fabric_attr(struct fi_info *prov_info, struct efa_device *device)
+int efa_prov_info_set_fabric_attr(struct fi_info *prov_info, struct efa_device *device, enum efa_info_type info_type)
 {
-	size_t name_len = strlen(EFA_FABRIC_NAME);
+	char *fabric_name;
+
+	if (info_type == EFA_INFO_DIRECT)
+		fabric_name = EFA_DIRECT_PROV_NAME;
```
shijin-aws: same comment, should be FABRIC_NAME instead of PROV_NAME

prov/efa/src/efa_user_info.c
```diff
@@ -332,7 +332,88 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
...
+
+	/* Print a warning and use FI_AV_TABLE if the app requests FI_AV_MAP */
+	if (hints && hints->domain_attr && hints->domain_attr->av_type == FI_AV_MAP)
+		EFA_WARN(FI_LOG_CORE, "FI_AV_MAP is deprecated in Libfabric 2.x. Please use FI_AV_TABLE. "
+					"EFA direct provider will now switch to using FI_AV_TABLE.\n");
+	info->domain_attr->av_type = FI_AV_TABLE;
+
+	if (!hints || !hints->domain_attr ||
+	    hints->domain_attr->resource_mgmt == FI_RM_UNSPEC)
+		info->domain_attr->resource_mgmt = FI_RM_ENABLED;
```
shijin-aws: There is no resource management for direct.

prov/efa/src/efa_base_ep.h
```diff
@@ -39,14 +39,14 @@
 #define efa_rx_flags(efa_base_ep) ((efa_base_ep)->util_ep.rx_op_flags)
 #define efa_tx_flags(efa_base_ep) ((efa_base_ep)->util_ep.tx_op_flags)
 
-#define EFA_PROV_NAME "efa"
-#define EFA_DIRECT_PROV_NAME "efa-direct"
+#define EFA_FABRIC_NAME "efa"
```
shijin-aws, line 42: This should have belonged to earlier commit

prov/efa/src/efa_base_ep.h
```diff
@@ -39,6 +39,26 @@
 #define efa_rx_flags(efa_base_ep) ((efa_base_ep)->util_ep.rx_op_flags)
 #define efa_tx_flags(efa_base_ep) ((efa_base_ep)->util_ep.tx_op_flags)
 
+#define EFA_FABRIC_NAME "efa"
+#define EFA_DIRECT_FABRIC_NAME "efa-direct"
```
shijin-aws, line 43: Can we make these macros defined in efa.h? It's odd to have these info level macros in a base_ep header file

prov/efa/src/efa_prov_info.c
```diff
@@ -455,13 +488,25 @@ int efa_prov_info_alloc(struct fi_info **prov_info_ptr,
...
+		case EFA_INFO_DIRECT:
+			prov_info->caps	= EFA_DIRECT_CAPS;
+			prov_info->mode |= FI_CONTEXT2; 	/* EFA direct path requires FI_CONTEXT2 mode */
+			break;
+		case EFA_INFO_RDM:
+			prov_info->caps	= EFA_RDM_CAPS;
+			break;
+		case EFA_INFO_DGRAM:
+			prov_info->caps	= EFA_DGRM_CAPS;
+			prov_info->mode |= FI_CONTEXT2; 	/* EFA direct path requires FI_CONTEXT2 mode */
```
shijin-aws, line 529: Nit: `direct` -> `dgram`

prov/efa/test/efa_unit_test_common.c
```diff
@@ -52,37 +52,42 @@ void efa_unit_test_construct_msg_rma(struct fi_msg_rma *msg, struct iovec *iov,
...
+struct fi_info *efa_unit_test_alloc_hints(enum fi_ep_type ep_type, char *fabric_name)
 {
 	struct fi_info *hints;
 
 	hints = fi_allocinfo();
 	if (!hints)
 		return NULL;
 
-	hints->fabric_attr->prov_name = strdup(prov_name);
+	if (fabric_name) {
```
shijin-aws, line 63: nit: no bracket for single line

shijin-aws (approved):
Looks great to me, thank you! Only a few nits.

prov/efa/src/efa_base_ep.h
sunkuamzn: I initially put them in `efa.h` but I need them in `efa_domain.h` for `efa_domain_name_suffix`

Will fix alignment

prov/efa/src/efa_prov_info.c
```diff
@@ -158,40 +164,35 @@ const struct fi_ep_attr efa_ep_attr = {
...
+
+	switch (info_type) {
+		case EFA_INFO_RDM:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			/* max_msg_size for RDM path is set in efa_prov_info_alloc_for_rdm */
+			break;
+		case EFA_INFO_DIRECT:
+			prov_info->ep_attr->type = FI_EP_RDM;
+			/* max_msg_size in ep_attr should be the maximum of send/recv and RMA sizes */
+			prov_info->ep_attr->max_msg_size = MAX(device->ibv_port_attr.max_msg_sz, device->max_rdma_size);
```
shijin-aws, line 187: I missed one thing earlier. We should only do the `MAX(device->ibv_port_attr.max_msg_sz, device->max_rdma_size)` when hints request FI_RMA and the efa-direct can supports it, which means the device needs to support read, write, and unsolicited write recv

shijin-aws: And you need to add `FI_RMA` cap into efa-direct's prov_info when device supports it. Today if I run with efa-direct for your PR I will get a failed fi_getinfo for `fi_rma_bw` which requests `FI_RMA`

```
ce/libfabric/fabtests$ FI_LOG_LEVEL=warn fi_rma_bw -p efa -f efa-direct -E -S 1024
fi_getinfo(): common/shared.c:1071, ret=-61 (No data available)
(env) ubuntu@ip-172-31-40-58:~/PortaFiducia/build/libraries/libfabric/main/source/libfabric/fabtests$ 
```

prov/efa/src/efa_prov_info.c
```diff
@@ -88,29 +94,29 @@ const struct fi_domain_attr efa_domain_attr = {
...
-	name_len = strlen(device->ibv_ctx->device->name) + strlen(efa_domain_name_suffix(ep_type));
+	name_len = strlen(device->ibv_ctx->device->name) + strlen(efa_domain_name_suffix(info_type));
 	prov_info->domain_attr->name = malloc(name_len + 1);
 	if (!prov_info->domain_attr->name) {
 		return -FI_ENOMEM;
 	}
 
 	snprintf(prov_info->domain_attr->name, name_len + 1, "%s%s",
-		 device->ibv_ctx->device->name, efa_domain_name_suffix(ep_type));
+		 device->ibv_ctx->device->name, efa_domain_name_suffix(info_type));
```
shijin-aws, line 119: For `	prov_info->domain_attr->resource_mgmt	= FI_RM_DISABLED;` , I think we need to make it as `FI_RM_ENABLED` for efa-direct, as we set infinite rnr retry for the qp which is some kind of resource management.. The current fi_rma_bw requires `FI_RM_ENABLED` so if we leave it as disabled it will make fi_getinfo failed.

shijin-aws:
I still hit bad qpn error when I only test rdm_pingpong with 1K... which means there is something wrong. Need to address this before can merge the PR

```
libfabric/fabtests$ FI_LOG_LEVEL=warn fi_rma_bw -p efa -f efa-direct -E -S 1024 lo
calhost
libfabric:1266354:1738371468::efa:cq:efa_cq_handle_error():100<warn> err: 107, message: Bad queue pair (QP) number (QP does not exist or is in error state) My EFA addr: fi_addr_efa://[fe80::86f:e0ff:fe87:ae7b]:1:574304512 My host id: i-0be65d40744bbb17b Peer EFA addr: fi_addr_efa://[fe80::86f:e0ff:fe87:ae7b]:0:0 Peer host id: N/A (9)
libfabric:1266354:1738371468::efa:cq:efa_show_help():100<warn> This error is detected remotely; typically encountered when the peer process is no longer present
[error] fabtests:common/shared.c:3043: cq_readerr 107 (Transport endpoint is not connected), provider errno: 9 (Bad queue pair (QP) number (QP does not exist or is in error state) My EFA addr: fi_addr_efa://[fe80::86f:e0ff:fe87:ae7b]:1:574304512 My host id: i-0be65d40744bbb17b Peer EFA addr: fi_addr_efa://[fe80::86f:e0ff:fe87:ae7b]:0:0 Peer host id: N/A)
```

prov/efa/src/efa_prov_info.c
```diff
@@ -235,7 +236,30 @@ const struct fi_tx_attr efa_rdm_tx_attr = {
...
+	.iov_limit		= 1
+};
+
+/**
+ * @brief default TX attributes for efa-direct end point
+ */
+const struct fi_tx_attr efa_direct_tx_attr = {
+	.caps			= EFA_RDM_TX_CAPS,
+	.mode			= 0,
+	.op_flags		= EFA_TX_OP_FLAGS,
```
shijin-aws, line 250: direct should support the same op flags as rdm support today

```
FI_INJECT | FI_COMPLETION | FI_TRANSMIT_COMPLETE |
						   FI_DELIVERY_COMPLETE;
```

prov/efa/src/efa_prov_info.c
```diff
@@ -235,7 +236,30 @@ const struct fi_tx_attr efa_rdm_tx_attr = {
...
+	.rma_iov_limit		= 1,
+};
+
+/**
+ * @brief default RX attributes for rdm end point
+ */
+const struct fi_rx_attr efa_direct_rx_attr = {
+	.caps			= EFA_RDM_RX_CAPS,
+	.mode			= EFA_RX_MODE,
+	.op_flags		= EFA_RX_OP_FLAGS,
```
shijin-aws, line 262: It should support `FI_COMPLETION`

shijin-aws:
I would carefully review `int efa_prov_info_alloc_for_rdm(struct fi_info **prov_info_rdm_ptr,
				struct efa_device *device)` to see whether we miss any configuration that should be supported for efa-direct as well.

prov/efa/src/efa_av.c
```diff
@@ -904,7 +904,7 @@ int efa_av_open(struct fid_domain *domain_fid, struct fi_av_attr *attr,
 	if (ret)
 		goto err;
 
-	if (EFA_EP_TYPE_IS_RDM(efa_domain->info)) {
+	if (EFA_INFO_TYPE_IS_RDM(efa_domain->info)) {
 		av->ep_type = FI_EP_RDM;
```
shijin-aws, line 908: This makes efa_av_open only assign ep_type as RDM when info type as rdm. Direct info will make it assigned as dgram....

prov/efa/src/efa_prov.c
```diff
@@ -91,6 +121,13 @@ static int efa_util_prov_initialize()
 			continue;
 		}
 
+		/* Set fabric name */
```
shijin-aws, line 124: You can still keep `efa_prov_info_set_fabric_attr` buit name `fabric_name` as an argument of the function?

prov/efa/src/efa_prov_info.c
```diff
@@ -167,31 +149,21 @@ void efa_prov_info_set_ep_attr(struct fi_info *prov_info,
...
-
 	prov_info->ep_attr->protocol	= FI_PROTO_EFA;
-	prov_info->ep_attr->type	= ep_type;
+	prov_info->ep_attr->max_msg_size = device->ibv_port_attr.max_msg_sz;
 
-	if (prov_info->ep_attr->type == FI_EP_RDM) {
-		prov_info->tx_attr->inject_size = device->efa_attr.inline_buf_size;
+	if (ep_type == FI_EP_RDM) {
+		prov_info->ep_attr->type = FI_EP_RDM;
+		prov_info->ep_attr->max_msg_size = device->ibv_port_attr.max_msg_sz;
```
shijin-aws, line 157: This line can be removed?

prov/efa/src/efa_prov_info.c
```diff
@@ -167,31 +149,21 @@ void efa_prov_info_set_ep_attr(struct fi_info *prov_info,
 {
 
 	*prov_info->ep_attr = efa_ep_attr;
-	if (ep_type == FI_EP_DGRAM) {
-		prov_info->mode |= FI_MSG_PREFIX;
-		prov_info->ep_attr->msg_prefix_size = 40;
-	}
-
 	prov_info->ep_attr->protocol	= FI_PROTO_EFA;
-	prov_info->ep_attr->type	= ep_type;
```
shijin-aws, line 176: This line can be kept, so you don't need to assign it in the if again?

prov/efa/src/efa_prov_info.c
```diff
@@ -40,30 +37,6 @@ const struct fi_fabric_attr efa_fabric_attr = {
...
-/**
- * @brief set the fabric_attr field of a prov_info
- *
- * @param 	prov_info[out]		prov_info object
- * @param	device[in]		pointer to an efa_device struct, which contains device attributes
- * @return	0 on sucessess
- * 		-FI_ENOMEM if memory allocation failed
- */
-static
-int efa_prov_info_set_fabric_attr(struct fi_info *prov_info, struct efa_device *device)
```
shijin-aws, line 52: This function can be reused by introducing fabric_name as an input? It doesn't seem you need `device` here.

prov/efa/src/efa_prov_info.c
```diff
@@ -455,13 +434,25 @@ int efa_prov_info_alloc(struct fi_info **prov_info_ptr,
 	if (!prov_info)
 		return -FI_ENOMEM;
 
-	if (ep_type != FI_EP_RDM && ep_type != FI_EP_DGRAM) {
-		EFA_WARN(FI_LOG_DOMAIN, "Unsupported endpoint type: %d\n",
-			 ep_type);
-		return -FI_ENODATA;
+	prov_info->mode |= FI_CONTEXT2; 	/* EFA direct path requires FI_CONTEXT2 mode */
```
shijin-aws, line 461: I am not sure whether we should assign FI_CONTEXT2 here as it is not a device attribute. But I don't have a strong opinion as long as we removed it for rdm

prov/efa/src/efa_user_info.c
```diff
@@ -332,7 +332,82 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
...
+ */
+static
+int efa_user_info_alter_direct(int version, struct fi_info *info, const struct fi_info *hints)
+{
+	if (hints && (hints->caps & FI_HMEM)) {
+		/*
+		 * FI_HMEM is a primary capability, therefore only check
+		 * (and cliam) its support when user explicitly requested it.
+		 */
+		if (!efa_user_info_should_support_hmem(version)) {
```
shijin-aws, line 355: I don't think we need to call this function, as it has some API version check & device rdma check which is only useful for efa-rdm

prov/efa/src/efa_prov_info.c
```diff
@@ -388,20 +388,44 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
 }
 
 #if HAVE_CUDA || HAVE_NEURON || HAVE_SYNAPSEAI
-void efa_prov_info_set_hmem_flags(struct fi_info *prov_info)
+void efa_prov_info_set_hmem_flags(struct fi_info *prov_info, enum fi_ep_type ep_type)
```
shijin-aws, line 391: I'd like to have this commit as part of your efa-direct info commit

prov/efa/src/efa_prov_info.c
shijin-aws: For safety, I think we need to add this to tx/rx_attr->mode as well.

prov/efa/src/efa_user_info.c
sunkuamzn: Discussed offline. Will keep `ofi_hmem_p2p_disabled` check but move it to `efa_prov_info`

prov/efa/src/efa_hmem.c
```diff
@@ -11,7 +11,7 @@ struct efa_hmem_info g_efa_hmem_info[OFI_HMEM_MAX];
 static size_t efa_max_eager_msg_size_with_largest_header() {
 	int mtu_size;
 
-	mtu_size = g_device_list[0].rdm_info->ep_attr->max_msg_size;
```
shijin-aws, line 14: Need to make sure the old rdm_info->ep_attr->max_msg_size doesn't cut the header size, otherwise the protocol threshold calculation could be wrong by this change.

prov/efa/src/efa_prov_info.c
```diff
@@ -254,12 +220,26 @@ void efa_prov_info_set_tx_rx_attr(struct fi_info *prov_info,
...
+		prov_info->rx_attr->op_flags |= FI_COMPLETION;
+
+		/* If the device supports FI_RMA, also set tx_attr and rx_attr */
+		if (prov_info->caps & FI_RMA) {
+			prov_info->tx_attr->caps |= OFI_TX_RMA_CAPS;
+			prov_info->rx_attr->caps |= OFI_RX_RMA_CAPS;
+		}
+
+		/* efa-direct path requires FI_CONTEXT2 */
+		prov_info->tx_attr->mode |= FI_CONTEXT2;
```
shijin-aws, line 234: Both dgram and direct requires FI_CONTEXT2.

prov/efa/src/efa_hmem.c
sunkuamzn: It does not https://github.com/ofiwg/libfabric/blob/c3420a7402a10f5fe3f475ac5eb39b3f2837f8d1/prov/efa/src/efa_prov_info.c#L194

prov/efa/src/efa_prov_info.c
sunkuamzn: fixed

shijin-aws (approved):

prov/efa/src/efa_prov_info.c
```diff
@@ -251,15 +217,29 @@ void efa_prov_info_set_tx_rx_attr(struct fi_info *prov_info,
 				  struct efa_device *device,
 				  enum fi_ep_type ep_type)
 {
+	/* efa-direct and DGRAM paths require FI_CONTEXT2 */
+	prov_info->tx_attr->mode |= FI_CONTEXT2;
```
shijin-aws, line 221: This mode will be washed by the change in L225....

shijin-aws: Need to move to the place after the `if` block

shijin-aws:
bot:aws:retest

shijin-aws (approved):

prov/efa/src/efa_prov_info.c
```diff
@@ -409,20 +389,54 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+			if (hmem_info->initialized && !hmem_info->p2p_supported_by_device) {
+				EFA_INFO(FI_LOG_CORE,
+					"EFA direct provider was compiled with support for %s HMEM interface "
+					"but the interface does not support p2p transfers. "
+					"EFA direct provider does not support HMEM transfers without p2p support. "
+					"HMEM support will be disabled.\n", fi_tostr(&iface, FI_TYPE_HMEM_IFACE));
+					return;
+			}
+		}
+		enable_hmem = true;
```
jiaxiyan, line 428: Do you need this variable `enable_hmem`?

prov/efa/test/efa_unit_test_info.c
```diff
@@ -44,6 +45,129 @@ void test_info_open_ep_with_wrong_info()
...
+ */
+void test_info_direct_attributes()
+{
+	struct fi_info *hints, *info = NULL, *info_head = NULL;
+	int err;
+
+	hints = efa_unit_test_alloc_hints(FI_EP_RDM, EFA_DIRECT_FABRIC_NAME);
+	assert_non_null(hints);
+
+	err = fi_getinfo(FI_VERSION(1,6), NULL, NULL, 0, hints, &info);
```
jiaxiyan, line 105: Should we test `FI_VERSION(2,0)`?

prov/efa/src/efa_prov_info.c
sunkuamzn: Yes. To catch the case when `HAVE_CUDA` is defined but `ofi_hmem_is_initialized(FI_HMEM_CUDA)` is false. That happens when you compile with CUDA but run without it.

prov/efa/test/efa_unit_test_info.c
sunkuamzn: It doesn't matter much

There is code in `fi_getinfo` in `fabric.c` that does not returns ENOSYS if the request version is older than the version Libfabric is compiled with. If the requested version is older, there is no impact.
https://github.com/ofiwg/libfabric/blob/c3420a7402a10f5fe3f475ac5eb39b3f2837f8d1/src/fabric.c#L1332-L1336

And there is some code in `ofi_check_info` for versions below 1.5. Other than that I don't see the version being used anywhere else.
https://github.com/ofiwg/libfabric/blob/c3420a7402a10f5fe3f475ac5eb39b3f2837f8d1/prov/util/src/util_attr.c#L1089-L1092

prov/efa/src/efa_prov_info.c
jiaxiyan: It seems better to do an early return for`if !ofi_hmem_is_initialized`, then you do not need this enable_hmem.

prov/efa/src/efa_prov_info.c
```diff
@@ -409,20 +389,54 @@ static int efa_prov_info_set_nic_attr(struct fi_info *prov_info, struct efa_devi
...
+		/* Check if FI_HMEM_P2P_DISABLED is set */
+		if (ofi_hmem_p2p_disabled()) {
+			EFA_WARN(FI_LOG_CORE,
+				"FI_HMEM capability requires peer to peer "
+				"support, which is disabled because "
+				"FI_HMEM_P2P_DISABLED was set to 1/on/true.\n");
+			return;
+		}
+
+		EFA_HMEM_IFACE_FOREACH(i) {
```
jiaxiyan, line 416: Do we need to check for system here?

sunkuamzn: No but FI_HMEM_SYSTEM will never fail the check below, so I don't think it's worth skipping FI_HMEM_SYSTEM explicitly
https://github.com/ofiwg/libfabric/blob/c3420a7402a10f5fe3f475ac5eb39b3f2837f8d1/prov/efa/src/efa_hmem.c#L265-L268

jiaxiyan: There is a `EFA_HMEM_IFACE_FOREACH_NON_SYSTEM`

sunkuamzn:
bot:aws:retest

prov/efa/test/efa_unit_test_info.c
aws-nslick, line 105: > returns ENOSYS if the request version is older than the version Libfabric is compiled with.

@sunkuamzn I think you accidentally swapped it in your head -- it errors in the case that the hardcoded config.h version (via `fi_version()`) is less than the requested version. So if an app is holding their version at some API level with the expectation that it isolates them from the user upgrading to a newer version, they will be surprised by this. 

eg: aws-ofi-nccl right now has the following: 

```cpp
	get_hints(hints);
	api_version = nccl_ofi_dmabuf_viable() ? FI_VERSION(1, 20) : FI_VERSION(1, 18);
	ret = nccl_ofi_ofiutils_get_providers(provider_filter, api_version, hints,
					      &provider_list, &num_providers);
	if (ret == 0) {
		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Using Libfabric %u.%u API, with %s support",
			       FI_MAJOR(api_version),
			       FI_MINOR(api_version),
			       FI_VERSION_GE(api_version, FI_VERSION(1, 20)) ? "DMA-BUF" : "GPUDirect RDMA");
		/* The 1.18 API allows providers to use CUDA to
		 * support HMEM pointers, so just having HMEM doesn't
		 * tell us anything about the usability of CUDA
		 * pointers with NCCL.  So leave the state unknown
		 * until we create an endpoint and try to disable
		 * CUDA
		 */
		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
			       "Using Libfabric 1.18 API, with GPUDirect RDMA support");
		support_gdr = GDR_UNKNOWN;
```

I did not test, but I'd not expect to hit the early return you linked if I ran with HEAD.

sunkuamzn: Yes I got that backwards. If the requested version is newer, then Libfabric will return ENOSYS. Updated my comment.

PR #10741, opened 2025-01-30
prov/util: Fix flag initialization for generic receive of unexpected entry

sdashevsky-ai:
The flags that are passed to util_init_rx_entry when called from util_srx_generic_trecv are not OR'ed with FI_TAGGED | FI_RECV. The function util_srx_generic_trecv is also called in a different flow from util_get_recv_entry with the flags OR'ed (working flow). The same issue is present in the util_srx_generic_recv function but instead of FI_TAGGED it should be FI_MSG.


aingerson (approved):

j-xiong (approved):

j-xiong:
bot:aws:retest

shijin-aws:
aws ci failure was a transient issue and not irrelevant

PR #10753, opened 2025-01-31
github/actions: Change to manual merge for the man page converter

j-xiong:
Merge cannot be done automatically due to the new review approval requirement on the main and stable branches.

Revert previous change about adding an approval review since self-approval is not allowed.

Add a random number to the temporary branch to avoid conflict among multiple action runs since now the temporary branches need to be deleted manually.

### DIFF
--- a/.github/workflows/nroff-elves.sh
+++ b/.github/workflows/nroff-elves.sh
@@ -51,7 +51,7 @@ done
 git config --global user.name "OFIWG Bot"
 git config --global user.email "ofiwg@lists.openfabrics.org"
 
-branch_name=pr/update-nroff-generated-man-pages-$BASE_REF
+branch_name=pr/update-nroff-generated-man-pages-$BASE_REF-`date +%s`
 git checkout -b $branch_name
 
 set +e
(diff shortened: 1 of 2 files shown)


.github/workflows/nroff-elves.sh
```diff
@@ -70,6 +70,10 @@ git push --set-upstream origin $branch_name
 url=`gh pr create --base $BASE_REF --title 'Update nroff-generated man pages' --body ''`
 pr_num=`echo $url | cut -d/ -f7`
 
+# skip the remaining steps for now since merge now requires
+# appproval which cannot be done automatically.
+exit 0
```
aingerson, line 75: Is there a benefit in exiting instead of just removing the below code?

.github/workflows/nroff-elves.sh
```diff
@@ -51,7 +51,7 @@ done
 git config --global user.name "OFIWG Bot"
 git config --global user.email "ofiwg@lists.openfabrics.org"
 
-branch_name=pr/update-nroff-generated-man-pages-$BASE_REF
+branch_name=pr/update-nroff-generated-man-pages-$BASE_REF-`date +%s`
```
aingerson, line 54: Does the %s just pick up something random?

zachdworkin:
CI failure will get fixed with a rebase. It is unrelated to this PR

.github/workflows/nroff-elves.sh
j-xiong: keep those code for future reference in case the repo policy is changed again or a new way of merging become available.

.github/workflows/nroff-elves.sh
j-xiong: the command would print "seconds since the Epoch".

aingerson (approved):

PR #10770, opened 2025-02-06
fabtests: Reorganize the FI_RX_CQ_DATA set.

shijin-aws:
FI_RX_CQ_DATA mode shouldn't be set per operation. It should be set based on whether the code supports it in the implementation. This patch move the FI_RX_CQ_DATA set to the hints setup phase among the code that supports FI_RX_CQ_DATA in implementation, like FI_CONTEXT and FI_CONTEXT2.


fabtests/common/shared.c
```diff
@@ -3711,13 +3711,11 @@ int ft_parse_api_opts(int op, char *optarg, struct fi_info *hints,
 			opts->rma_op = FT_RMA_READ;
 		} else if (!strcasecmp(optarg, "writedata")) {
 			hints->caps |= FI_WRITE | FI_REMOTE_WRITE;
-			hints->mode |= FI_RX_CQ_DATA;
 			hints->domain_attr->cq_data_size = 4;
 			opts->rma_op = FT_RMA_WRITEDATA;
 			opts->cqdata_op = FT_CQDATA_WRITEDATA;
 			cq_attr.format = FI_CQ_FORMAT_DATA;
 		} else if (!strcasecmp(optarg, "senddata")) {
-			hints->mode |= FI_RX_CQ_DATA;
```
shijin-aws, line 3720: Thinking about it more, I think I would keep these to avoid breaking the tests for providers that require FI_RX_CQ_DATA.

shefty:
FI_RX_CQ_DATA only needs to be set if the app is sending CQ data.  This looks possible for rma_pingpong.  I can't tell for rdm_bw, but it doesn't look like it.

The code in ft_parse_api_opts() looks like it updates the hints correctly, which should work for rma_pingpong.

shijin-aws:
@shefty 
>FI_RX_CQ_DATA only needs to be set if the app is sending CQ data. This looks possible for rma_pingpong. I can't tell for rdm_bw, but it doesn't look like it.

If a provider requires FI_RX_CQ_DATA for FI_RMA, having the program `| FI_RX_CQ_DATA` only for writedata operation will make the write/read test fail in the fi_getinfo due to missing mode. I think it should be added as long as the application can support this mode, no matter whether it is actually sending cq data

j-xiong:
Yes, I think it is safe to add the mode bit to hints as long as the application can handle it even if the situation may not actually occur.

j-xiong (approved):

shefty:
It's safe to add to the hints, but the absence of it stresses the correctness of the provider checks.  That is, if the test is failing, that's likely a problem being exposed in the provider, not the test.

If the mode is not set and the cq_data_size = 0, the provider should accept the request and return cq_data_size = 0 in the response.

shijin-aws:
>If the mode is not set and the cq_data_size = 0, the provider should accept the request and return cq_data_size = 0 in the response.

Thanks! I think that is a better response. I will close this PR then

shijin-aws:
The risk of silently returning cq_data_size is that it apply to both MSG and RMA. If provider can support cq data on MSG but not RMA, there is no way to tell it from this size. We may want to do such operation based cq data size split via `fi_getopt`, similar to `FI_OPT_MAX_MSG_SIZE` and `FI_OPT_MAX_RMA_SIZE`. But it will make the API more and more complicated and is something I want to avoid

shefty:
Each fi_info references a single endpoint type.  The cq_data_size is relative to that endpoint type.

If the user hints set cq_data_size = 0, the provider can deal with that in whatever way makes sense.  If that means there's 1 fi_info for msg_ep with cq_data_size=8 and 1 fi_info for rdm_ep with cq_data_size=0, that's perfectly fine.

The application should not need to set a mode bit for a feature that it isn't going to use at all, so the provider doesn't incorrectly reject the request.

shijin-aws:
But application can use the same endpoint for both MSG and RMA operations. I think it's legal for it to not specify a cq data size in hints and rely on whatever returned from in_info. Then it cannot tell from the return that whether they cannot do cq data for only RMA, or both.

shefty:
I mixed transfer types with ep types.  The API does not distinguish between msg vs rma cq data.  CQ data can be transferred with a separate transport message, as long as the data ordering and visibility requirements are met.  This is easier to add for rma operations, than msg sends.  I.e. write with cq data can be split into write + 0-length send with cq data.  I think this is what iwarp ends up with.

I don't see how this relates to the proposed changes to fabtests.

shijin-aws:
@shefty I am replying on your earlier response of

>If the mode is not set and the cq_data_size = 0, the provider should accept the request and return cq_data_size = 0 in the response.

I am talking about the risk of silently returning cq_data_size as 0 to the application when `FI_RX_CQ_DATA` is not set while application doesn't specify cq_data_size in hints either.

shefty:
But that's acceptable behavior.  The app didn't request cq data and didn't indicate that it can support the provider's requirements for exposing cq data.

shijin-aws:
>I don't see how this relates to the proposed changes to fabtests.

The motivation of the proposed changes in this PR is that I want to explicitly fail FI_RMA request when FI_RX_CQ_DATA mode is not set, no matter of operation type or cq data size. Then I don't use current `fi_rma_pingpong` or `fi_rma_bw` with write or read operation type to test this mode

shijin-aws:
>But that's acceptable behavior. The app didn't request cq data and didn't indicate that it can support the provider's requirements for exposing cq data.

All understood. I will evaluate it.

PR #10772, opened 2025-02-07
prov/shm: remove prefix from map inserts

aingerson:
EP names have the format fi_shm://<shm_name> to be consistent with other provider formats. Since shared memory names cannot have slashes in them, the shm provider skips this prefix when creating the shared memory region. However, it does not do this when inserting the names into the map which causes an issue when comparing names in the map against region names.
This removes the prefix for map additions as well so that names match in both data structures.

### DIFF
--- a/prov/shm/src/smr_util.c
+++ b/prov/shm/src/smr_util.c
@@ -558,10 +558,11 @@ int smr_map_add(const struct fi_provider *prov, struct smr_map *map,
 		const char *name, int64_t *id)
 {
 	struct ofi_rbnode *node;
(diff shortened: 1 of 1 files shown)


zachdworkin (approved):

j-xiong:
There were two failed tests with udp and 1 with tcp_io_uring. I don't think they are related to this PR.

zachdworkin:
You are right, they are not. It is rerunning anyways to get the green check.

shijin-aws (approved):

PR #10776, opened 2025-02-08
prov/efa: Bug fix in the RDM path with FI_MSG_PREFIX mode

sunkuamzn:
efa_rdm_txe_construct is called to construct a txe for the handshake packet and to construct txes for send, RMA and atomic operations.

iov_count is 0 only in the handshake path. When efa_rdm_txe_construct is called as part of an fi_send or fi_read or fi_write call from the application, iov_count is never 0.

The handshake packet is posted internally by the EFA provider, so it does not make sense to allocate and consume prefix space. Instead just check for iov_count before consuming the prefix space.


shijin-aws (approved):

shijin-aws:
Ideally we may want to stop using txe for handshake. It caused more troubles than the performance benefits I could bring. But we can discuss it later.

PR #10779, opened 2025-02-10
prov/tcp: Only disable ep if the failure can not be retried

wangdi1:
Only disable ep if the failure can not be retried.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1421,6 +1421,9 @@ int ofi_bsock_async_done(const struct fi_provider *prov,
 	msg.msg_controllen = sizeof(ctrl);
 	ret = recvmsg(bsock->sock, &msg, MSG_ERRQUEUE);
 	if (ret < 0) {
+		if (OFI_SOCK_TRY_SND_RCV_AGAIN(errno))
+			return 0;
+
 		FI_WARN(prov, FI_LOG_EP_DATA,
 			"Error reading MSG_ERRQUEUE (%s)\n", strerror(errno));
 		return -errno;


wangdi1:
@shefty  are you ok with the patch?  We saw some failures on our site

"ofi_bsock_async_done() Error reading MSG_ERRQUEUE (Resource temporarily unavailable)"

If this means EAGAIN, then ep should not be disabled?

shefty (changes requested):
Please move the check into ofi_bsock_async_done() after recvmsg().  This should have been trapped there and the error discarded (no warning message).

wangdi1:
> Please move the check into ofi_bsock_async_done() after recvmsg(). This should have been trapped there and the error discarded (no warning message).

ah, sure. thanks! @shefty

src/common.c
```diff
@@ -1421,6 +1421,9 @@ int ofi_bsock_async_done(const struct fi_provider *prov,
 	msg.msg_controllen = sizeof(ctrl);
 	ret = recvmsg(bsock->sock, &msg, MSG_ERRQUEUE);
 	if (ret < 0) {
+		if (OFI_SOCK_TRY_SND_RCV_AGAIN(ret))
```
shefty, line 1424: This is a direct call to recvmsg, so you want to use ```errno``` instead of ret.  This ends up being linux specific code.

wangdi1: ah, yes. Thanks!

shefty:
Final changes look okay, but please squash the commits

wangdi1:
> Final changes look okay, but please squash the commits

sure. thanks

shefty (approved):
looks good -- thanks

PR #10788, opened 2025-02-13
prov/efa: Do GDRCopy registrations only in the EFA RDM path

sunkuamzn:
The efa-direct and dgram paths do not need GDRCopy because they don't
support copy based protocols. GDRCopy registrations can cause problems
in some cases, so skip them.


prov/efa/src/efa_domain.h
```diff
@@ -33,6 +33,7 @@ struct efa_domain {
 	size_t ibv_mr_reg_sz;
 
 	/* Only valid for RDM EP type */
+	bool			rdm_ep;		/* Set to true for RDM domain. False otherwise. */
```
sunkuamzn, line 36: A new field seemed better than comparing `fabric_attr->name` for every `fi_mr_reg` call

shijin-aws: Yeah, that was I thinking of too. I don't want to compare fabric name every time since `fi_mr_reg` is on the hot path for rdm code path

prov/efa/test/efa_unit_test_mr.c
```diff
@@ -3,33 +3,150 @@
 
 #include "efa_unit_tests.h"
 
-void test_efa_mr_reg_counters(struct efa_resource **state)
+static void test_efa_mr_counters(struct efa_domain *efa_domain, struct fid_mr *mr, int mr_reg_count, int mr_reg_size, bool gdrcopy_flag)
```
shijin-aws, line 6: It's not testing "counters" only any more. we need a better name for this function, e.g. `test_efa_mr_impl`

prov/efa/test/efa_unit_test_mr.c
```diff
@@ -3,33 +3,150 @@
...
+    size_t mr_size = 64;
+    void *buf;
+    struct fid_mr *mr = NULL;
+    struct fi_mr_attr mr_reg_attr = {0};
+    struct iovec iovec;
+    int err;
+
+    if (g_efa_hmem_info[FI_HMEM_CUDA].initialized && g_efa_hmem_info[FI_HMEM_CUDA].p2p_supported_by_device) {
+	    resource->hints = efa_unit_test_alloc_hints_hmem(FI_EP_RDM, EFA_DIRECT_FABRIC_NAME);
+		efa_unit_test_resource_construct_with_hints(resource, FI_EP_RDM, FI_VERSION(2, 0),
```
shijin-aws, line 117: indent is off.

prov/efa/test/efa_unit_test_mr.c
```diff
@@ -3,33 +3,150 @@
...
+    size_t mr_size = 64;
+    void *buf;
+    struct fid_mr *mr = NULL;
+    struct fi_mr_attr mr_reg_attr = {0};
+    struct iovec iovec;
+    int err;
+
+    if (hmem_ops[FI_HMEM_CUDA].initialized && g_efa_hmem_info[FI_HMEM_CUDA].p2p_supported_by_device) {
+	    resource->hints = efa_unit_test_alloc_hints_hmem(FI_EP_RDM, EFA_FABRIC_NAME);
+		efa_unit_test_resource_construct_with_hints(resource, FI_EP_RDM, FI_VERSION(2, 0),
```
shijin-aws, line 66: indent is off

prov/efa/src/efa_domain.h
shijin-aws: Not related to this PR, but I find `rdm_mode` can be safely removed as it is not used.

shijin-aws (approved):

sunkuamzn:
Intel CI seems to be hanging. This PR only has EFA provider changes, so ignoring the hang.

PR #10789, opened 2025-02-13
prov/cxi: Fix bug in constrained LE test cases in test.sh and test_sw.sh

rconnaughty:
This commit resolves a bug in certain CXI criterion test cases where the le_pools[] max_alloc value was being decreased for all LE pool entries.  This resulted in unintended LE append failures + NMIs + inbound wait timeouts related to the CXI eth driver.  Now, the test only lowers the max_alloc value for LE pools associated with the default CXI service, which is used during the criterion tests.  Additionally, instances of csrutil have been replaced with cxiutil.


swelch (approved):

swelch:
@shijin-aws - this should have no impact on AWS CI, are you ok for me to merge?

shijin-aws:
@swelch we have some infra issue right now. Please ignore

PR #10790, opened 2025-02-14
prov/efa: Do not write cq error for ope from internal operations

shijin-aws:
We have OPEs created for internal usage instead, like the RTW pkt used to trigger a handshake, and handshake pkt itself. When handling completion error associated with such op entries, we shouldn't write cq error otherwise application will see garbage cq err entries that don't associate with any operations. We should write eq error in this case instead.


prov/efa/src/rdm/efa_rdm_pke_rtr.c
```diff
@@ -66,6 +66,40 @@ ssize_t efa_rdm_pke_init_longcts_rtr(struct efa_rdm_pke *pkt_entry,
...
+	struct efa_rdm_ope *rxe;
+	struct efa_rdm_rtr_hdr *rtr_hdr;
+
+	rxe = efa_rdm_ep_alloc_rxe(ep, pkt_entry->addr, ofi_op_read_rsp);
+	if (OFI_UNLIKELY(!rxe))
+		return NULL;
+
+	rxe->addr = pkt_entry->addr;
+	rxe->bytes_received = 0;
+	rxe->bytes_copied = 0;
```
darrylabbate, line 92: These are already be initialized via `efa_rdm_ep_alloc_rxe()`

shijin-aws: Thanks, removed

prov/efa/src/rdm/efa_rdm_pke_rtr.c
```diff
@@ -66,6 +66,38 @@ ssize_t efa_rdm_pke_init_longcts_rtr(struct efa_rdm_pke *pkt_entry,
...
+{
+	struct efa_rdm_ep *ep = pkt_entry->ep;
+	struct efa_rdm_ope *rxe;
+	struct efa_rdm_rtr_hdr *rtr_hdr;
+
+	rxe = efa_rdm_ep_alloc_rxe(ep, pkt_entry->addr, ofi_op_read_rsp);
+	if (OFI_UNLIKELY(!rxe))
+		return NULL;
+
+	rxe->addr = pkt_entry->addr;
```
darrylabbate, line 90: This is also redundant, right? https://github.com/ofiwg/libfabric/blob/e08d3161b8038ebfa50a4eb5bf76132e4eaeb137/prov/efa/src/rdm/efa_rdm_ep_utils.c#L100

darrylabbate (approved):

prov/efa/src/rdm/efa_rdm_pke_rtr.c
shijin-aws: Yep, I cleaned it for all `rta`, `rtr` and `rtw` in the new push

prov/efa/src/rdm/efa_rdm_pke_rta.c
```diff
@@ -415,10 +417,16 @@ int efa_rdm_pke_proc_fetch_rta(struct efa_rdm_pke *pkt_entry)
 	}
 
 	err = efa_rdm_ope_post_send_or_queue(rxe, EFA_RDM_ATOMRSP_PKT);
-	if (OFI_UNLIKELY(err))
-		efa_rdm_rxe_handle_error(rxe, -err, FI_EFA_ERR_PKT_POST);
 
 	efa_rdm_pke_release_rx(pkt_entry);
+
+	if (OFI_UNLIKELY(err)) {
```
sunkuamzn, line 423: Why make this change?

shijin-aws: This is an internal operation, so it should generate an eq error and release rxe. The old code is wrong here. `efa_rdm_rxe_handle_error` doesn't release the rxe and may write cq error (which this PR will fix as well)

sunkuamzn: But you fixed the behavior of `efa_rdm_rxe_handle_error`. And the rxe here is allocated by `efa_rdm_pke_alloc_rta_rxe` which does set the `EFA_RDM_OPE_INTERNAL` flag

shijin-aws: I slightly prefer to use a shortcut here to avoid involving unexpected dependencies in `rxe_handle_error`. And it is consistent with the similar code paths in `rtr`, `rtw` . If we want to use `efa_rdm_rxe_handle_error`, we should change them all to be consistent, which can be a follow-up PR.

prov/efa/src/rdm/efa_rdm_pke_rta.c
```diff
@@ -301,7 +301,9 @@ int efa_rdm_pke_proc_dc_write_rta(struct efa_rdm_pke *pkt_entry)
 		EFA_WARN(FI_LOG_CQ,
 			"Posting of receipt packet failed! err=%s\n",
 			fi_strerror(err));
-		efa_rdm_rxe_handle_error(rxe, -err, FI_EFA_ERR_PKT_POST);
+		efa_base_ep_write_eq_error(&pkt_entry->ep->base_ep, err, FI_EFA_ERR_PKT_POST);
+		efa_rdm_rxe_release(rxe);
+		efa_rdm_pke_release_rx(pkt_entry);
```
shijin-aws, line 306: I have a bug here: `pkt_entry` should have been released by `efa_rdm_pke_proc_write_rta` already. Need to move this change after L295

prov/efa/src/rdm/efa_rdm_pke_rta.c
shijin-aws: OK, I think there are many resource leak of `pkt_entry` in the rta path. So I'd rather make all such cleanup in a follow up work. I will revert to use `efa_rdm_rxe_handle_error` here but still release the `rxe` because it was a leak anyway

sunkuamzn (dismissed):

jiaxiyan (approved):

Issue #10757, opened 2025-02-02
prov/efa: failure to enumerate unconditionally when FI_PROV_ATTR_ONLY is present.


aws-nslick:
**Describe the bug**

Docs for `fi_getinfo` state the following, emphasis mine:

> ### FLAGS
> #### FI_PROV_ATTR_ONLY
> >    Indicates that the caller is only querying for what providers are potentially available. **All providers will return exactly one fi_info struct, regardless of whether that provider is usable on the current platform or not.** The returned fi_info struct will contain default values for all members, with the exception of fabric_attr. The fabric_attr member will have the prov_name and prov_version values filled in. 

EFA fails to appear in the results unconditionally.

**To Reproduce**

Test program:

```c++
// g++ -std=c++2a -DFMT_HEADER_ONLY=1 -L/opt/amazon/efa/lib -isystem/opt/amazon/efa/include test.cc -o test -lfabric
#include <cstdint>
#include <fmt/core.h>
#include <fmt/format.h>
#include <rdma/fabric.h>
#include <stdexcept>
#include <string>
#include <unordered_map>

int main() {
  static constexpr auto get = [](std::uint32_t major_version = FI_MAJOR_VERSION,
                                 std::uint32_t minor_version =
                                     FI_MINOR_VERSION) {
    fi_info *provider_list{nullptr};
    int ret = fi_getinfo(FI_VERSION(major_version, minor_version), nullptr,
                         nullptr, FI_PROV_ATTR_ONLY, nullptr, &provider_list);

    if (ret < 0 || !provider_list) {
      throw std::runtime_error("fi_getinfo(FI_PROV_ATTR_ONLY) failed");
    }
    struct version_pair {
      std::uint32_t major_version{};
      std::uint32_t minor_version{};
    };
    std::unordered_map<std::string, version_pair> providers;
    for (auto *current = provider_list; current != nullptr;
         current = current->next)
      providers.emplace(
          current->fabric_attr->prov_name,
          version_pair{FI_MAJOR(current->fabric_attr->prov_version) / 100,
                       FI_MAJOR(current->fabric_attr->prov_version) % 100});

    fi_freeinfo(provider_list);
    return providers;
  };

  for (auto &&[name, version] : get()) {
    auto &[maj, min] = version;
    fmt::print(stdout, "prov/{}: {}.{}\n", name, maj, min);
  }

  return 0;
};
```

Example output on a host where no EFA devices are present:

```shell
$ ./asdftest
[...]

PR #10794, opened 2025-02-17
prov/efa: Always return efa_prov in EFA_INI

shijin-aws:
Currently EFA_INI only return efa_prov
when there is efa device and the device info
setup correctly. This makes fi_getinfo
with FI_PROV_ATTR_ONLY not return efa
on non-efa instances. This patch
fixes it.

### DIFF
--- a/prov/efa/src/efa_prov.c
+++ b/prov/efa/src/efa_prov.c
@@ -190,10 +190,7 @@ EFA_INI
 
 	err = efa_device_list_initialize();
 	if (err)
-		return NULL;
-
-	if (g_device_cnt <= 0)
-		return NULL;
+		return &efa_prov;
 
 	/*
 	 * efa_env_initialize uses g_efa_device_list
@@ -236,7 +233,7 @@ EFA_INI
 
 err_free:
 	efa_prov_finalize();
-	return NULL;
+	return &efa_prov;
 }
 
 /**


shijin-aws:
Fixes https://github.com/ofiwg/libfabric/issues/10757

prov/efa/src/efa_prov.c
```diff
@@ -190,10 +190,7 @@ EFA_INI
 
 	err = efa_device_list_initialize();
 	if (err)
-		return NULL;
-
-	if (g_device_cnt <= 0)
-		return NULL;
+		return &efa_prov;;
```
shijin-aws, line 193: This should be ok since the device info allocate will be skipped anyway so there won't be unexpected info leaks

shijin-aws:
I tested on an instance without efa card and it works as expected:

- fi_info -p efa will return -61
- fi_info -l has efa

jiaxiyan (approved):

aws-nslick:
thank you!

prov/efa/src/efa_prov.c
aws-nslick: nit: clang complains for me when lines have extra semicolons, ymmv.

shijin-aws: Fish, fixed now!

shijin-aws:
@j-xiong is Intel CI hanging? I know this is a efa provider change, but it is in the `EFA_INI` part so it will be called anyway unless you disabled efa in the build

j-xiong:
@shijin-aws It's in the queue waiting to be started. There is a long-running job blocking the CI runs. You don't need to wait for the result.

PR #10802, opened 2025-02-20
prov/efa: Fix the max_msg_size reporting for efa-direct

shijin-aws:
Currently, efa-direct always return the max of rdma size and mtu size when device rdma is supported. This is wrong because when FI_RMA is not requested, only send/recv will be used, and the max_msg_size should be returned as the mtu size only. This patch fixes this issue by moving the max_msg_size change to alter_direct_info based on hints->caps.

Also adjust and introduces more unit tests.


prov/efa/test/efa_unit_test_info.c
```diff
@@ -94,24 +94,62 @@ void test_info_dgram_attributes()
...
+	assert_non_null(info_head);
 	for (info = info_head; info; info = info->next) {
-		assert_true(!strcmp(info->fabric_attr->name, EFA_DIRECT_FABRIC_NAME));
+		assert_true(!strcmp(info->fabric_attr->name,
+				    EFA_DIRECT_FABRIC_NAME));
 		assert_true(strstr(info->domain_attr->name, "rdm"));
 		assert_false(info->caps & (FI_ATOMIC | FI_TAGGED));
 		assert_false(info->tx_attr->msg_order & FI_ORDER_SAS);
-		assert_int_equal(info->ep_attr->max_msg_size, g_device_list[0].max_rdma_size);
+		assert_int_equal(
```
sunkuamzn, line 115: Could you also add
```
assert_int_equal(info->domain_attr->progress, FI_PROGRESS_AUTO);
assert_int_equal(info->domain_attr->control_progress, FI_PROGRESS_AUTO);
```

similar to https://github.com/ofiwg/libfabric/commit/103f94f24529ec3022716eb889f9a387aa9de359

sunkuamzn (commented):
Thanks! I had an almost similar commit that I was going to push today

prov/efa/test/efa_unit_test_info.c
shijin-aws: Done

sunkuamzn (approved):

jiaxiyan (approved):

PR #10803, opened 2025-02-20
include, prov: Change util_av lock to genlock

sunkuamzn:
The util_av lock will be held in the fast path in the EFA provider. With the correct threading and progress models
(FI_THREAD_DOMAIN/FI_THREAD_COMPLETION and FI_PROGRESS_CONTROL_UNIFIED), this lock should also be a no-op. Similar to the ep_list_lock.

This commit changes the mutex to a genlock that becomes a no-op for the correct threading and progress models and a mutex otherwise.

### DIFF
--- a/prov/util/src/util_av.c
+++ b/prov/util/src/util_av.c
@@ -275,7 +275,7 @@ int ofi_av_insert_addr_at(struct util_av *av, const void *addr, fi_addr_t fi_add
 {
 	struct util_av_entry *entry = NULL;
 
-	assert(ofi_mutex_held(&av->lock));
+	assert(ofi_genlock_held(&av->lock));
 	ofi_av_straddr_log(av, FI_LOG_INFO, "inserting addr", addr);
 	HASH_FIND(hh, av->hash, addr, av->addrlen, entry);
 	if (entry) {
@@ -302,7 +302,7 @@ int ofi_av_insert_addr(struct util_av *av, const void *addr, fi_addr_t *fi_addr)
 {
 	struct util_av_entry *entry = NULL;
 
-	assert(ofi_mutex_held(&av->lock));
+	assert(ofi_genlock_held(&av->lock));
 	ofi_av_straddr_log(av, FI_LOG_INFO, "inserting addr", addr);
 	HASH_FIND(hh, av->hash, addr, av->addrlen, entry);
 	if (entry) {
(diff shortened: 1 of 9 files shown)


prov/util/src/util_av.c
```diff
@@ -535,11 +534,16 @@ int ofi_av_init_lightweight(struct util_domain *domain, const struct fi_av_attr
 	 */
 	av->context = context;
 	av->domain = domain;
+	av->prov = domain->prov;
 
-
-	ep_list_lock_type = ofi_progress_lock_type(av->domain->threading,
+	lock_type = ofi_progress_lock_type(av->domain->threading,
 						   av->domain->control_progress);
```
j-xiong, line 540: Please fix the alignment caused by the change.

aingerson: Also I think the lock selection here is actually wrong. The progress lock type is more for the completion devices and the AV doesn't fall into that category. Here's the definition of the progress_lock_type:
```
static inline enum ofi_lock_type
ofi_progress_lock_type(enum fi_threading threading, enum fi_progress control)
{
	return (threading == FI_THREAD_DOMAIN || threading == FI_THREAD_COMPLETION) &&
		control == FI_PROGRESS_CONTROL_UNIFIED ? OFI_LOCK_NOOP : OFI_LOCK_MUTEX;
}
```
The AV still should get locked for thread completion. It can only be optimized out for thread domain.

sunkuamzn: @aingerson thanks, I misunderstood FI_THREAD_COMPLETION. I will fix

Is this the reason for Intel CI failure?

aingerson: hard to say for sure but maybe. the failure is a hang on MPICH verbs-rxm on a multi threaded test. MPICH I think uses FI_THREAD_COMPLETION so definitely a chance there's a race because it optimizes the lock out even though it needs it

prov/util/src/util_av.c
```diff
@@ -535,10 +534,21 @@ int ofi_av_init_lightweight(struct util_domain *domain, const struct fi_av_attr
 	 */
 	av->context = context;
 	av->domain = domain;
+	av->prov = domain->prov;
+
+	av_lock_type = domain->threading == FI_THREAD_DOMAIN &&
```
sunkuamzn, line 539: This is by clang-format 🤷

j-xiong (approved):

shijin-aws (approved):

sunkuamzn:
@aingerson do you have more comments?

aingerson (approved):
Looks good, thanks!

PR #10805, opened 2025-02-20
prov/efa: Set g_device_list to NULL after free

zachdworkin:
fi_info has a double free issue. The efa g_device_list isn't being set to NULL after being freed so the next time efa_device_list_finalize is called it will be freed again (double free). Setting it to NULL removes this.

### DIFF
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -230,6 +230,7 @@ void efa_device_list_finalize(void)
 			efa_device_destruct(&g_device_list[i]);
 
 		free(g_device_list);
+		g_device_list = NULL;
 	}
 
 	g_device_cnt = 0;


shijin-aws:
@zachdworkin Is it a new issue after https://github.com/ofiwg/libfabric/commit/b7b9dd69747b01f99653aa6266ff60c605d7e092? which test triggers this ?

shijin-aws:
@zachdworkin Thanks for catching it ! Let me run with ASAN to reproduce

zachdworkin:
@shijin-aws no problem! Seems simple enough. What does ASAN stand for?

shijin-aws:
ASAN is address sanitizer that we used to catch illegal memory access and double free error. But I couldn't reproduce the double free report: `fi_info -p tcp` just runs ok... how do you get that double free report?

shijin-aws:
I think I can see where the bug is introduced:

Before https://github.com/ofiwg/libfabric/commit/b7b9dd69747b01f99653aa6266ff60c605d7e092, `efa_prov` is never returned in the error path, so `EFA_FIN` is never called . But now `efa_prov` is returned anyway and `EFA_FIN` will be called in the process reaper again. So if `g_device_list` is not initialized and reset as NULL, `EFA_FIN` can see a garbage `efa_device_list` and double free it.

zachdworkin:
steps to reproduce
git reset --hard upstream/main
make distclean
./autogen.sh
./configure --prefix=install/location
make -j install

(check HAVE_EFA 1 in config.log)

install/location/bin/fi_info -p verbs
double free error

shijin-aws:
I think this PR is only a bandaid as we have more clean up issue here - The EFA_INI code block needs to move the initialization code to the getinfo. I will open a PR to fix it thoroughly

shijin-aws:
@zachdworkin can you try if https://github.com/ofiwg/libfabric/pull/10807 fixes the issue for you?

zachdworkin:
@shijin-aws #10807 fixes it. Should I close this and let that PR be the one to fix it?

shijin-aws:
Yes, you can close this one

PR #10807, opened 2025-02-20
prov/efa: Fix the clean up issue for efa_util_prov

shijin-aws:
Commit b7b9dd6 makes efa_prov always returned
by EFA_INI, even when efa device is not available. Then in `fi_fini` the efa_prov_finialize` will be
triggered on non-EFA platform. However, the current efa_prov_finalize doesn't handle the case where
the resources were not created, and it will cause
a double free error.

This patch fixex this issue.

### DIFF
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -162,8 +162,8 @@ static void efa_device_destruct(struct efa_device *device)
 	device->ibv_ctx = NULL;
 }
 
-struct efa_device *g_device_list;
-int g_device_cnt;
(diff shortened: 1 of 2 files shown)


zachdworkin (approved):

zachdworkin:
@shijin-aws I am forcing intel ci to run this. The X will go away when our controller picks up the job

shijin-aws:
@zachdworkin fixed a bug in earlier push, should be good now.

zachdworkin:
@shijin-aws thanks, our CI passed the second re-run

jiaxiyan (approved):

PR #10810, opened 2025-02-21
fabtests/prov/efa: Add remote exit early test

jiaxiyan:
Add tests to make sure when sender exits early in the middle of send, receiver polls cq and handle errors gracefully.
Add options for receiver posts receive, sender post tagged send, and sender posts fi_writedata.

### DIFF
--- a/fabtests/prov/efa/src/rdm_remote_exit_early.c
+++ b/fabtests/prov/efa/src/rdm_remote_exit_early.c
@@ -1,0 +1,220 @@
+/*
+ * Copyright (c) 2025, Amazon.com, Inc.  All rights reserved.
+ *
+ * This software is available to you under the BSD license
+ * below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * This test check the error handling path for receiver when sender exit early
+ * in the middle of send.
+ */
+
+#include <getopt.h>
+#include <shared.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <rdma/fi_tagged.h>
+
+static bool post_rx = false;
+
+enum {
+	LONG_OPT_POST_RX,
+};
+
+static int run()
+{
+	int ret;
+
+	ret = ft_init_fabric();
+	if (ret) {
+		FT_PRINTERR("ft_init_fabric", -ret);
+		return ret;
+	}
+
+	/*
+	 * The handshake procedure between server and client will happen in
+	 * either ft_exchange_keys() or ft_sync()
+	 */
+	if (opts.rma_op == FT_RMA_WRITEDATA) {
+		/* ft_exchange_keys finally calls ft_sync(), 
+		 * which will call ft_sync_inband(true), so it will cause the 
+		 * receiver to always post an outstanding recv buffer. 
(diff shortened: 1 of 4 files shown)


fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,221 @@
...
+
+	hints = fi_allocinfo();
+	if (!hints)
+		return EXIT_FAILURE;
+
+	timeout = 5;
+	int lopt_idx = 0;
+	struct option long_opts[] = {
+		{"post-rx", no_argument, NULL, LONG_OPT_POST_RX},
+		{"post-tagged", no_argument, NULL, LONG_OPT_POST_TAGGED},
```
shijin-aws, line 171: post_writedata and post_tagged can be implemented like an option code, see how `-o` is implemented in the fi_rma_bw test.

fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,221 @@
...
+			if (post_tagged)
+				ret = fi_trecv(ep, rx_buf, rx_size, mr_desc,
+				      FI_ADDR_UNSPEC, ft_tag, 0x0, &rx_ctx);
+			else
+				ret = fi_recv(ep, rx_buf, rx_size, mr_desc,
+				      FI_ADDR_UNSPEC, &rx_ctx);
+			printf("server posts recv\n");
+		}
+
+		ft_start();
```
shijin-aws, line 95: why do you need `ft_start()` and `ft_stop()`?

jiaxiyan: to track the time in between

fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,225 @@
...
+	if (!hints)
+		return EXIT_FAILURE;
+
+	timeout = 5;
+	int lopt_idx = 0;
+	struct option long_opts[] = {
+		{"post-rx", no_argument, NULL, LONG_OPT_POST_RX},
+		{0, 0, 0, 0}
+	};
+	while ((op = getopt_long(argc, argv, "T" ADDR_OPTS INFO_OPTS CS_OPTS API_OPTS,
```
shijin-aws, line 179: This is still not what I proposed .... You already have `-o` that covers `send`, `tagged` and `writedata` in 

https://github.com/ofiwg/libfabric/blob/main/fabtests/common/shared.c#L3704-L3731
, you shouldn't need to introduce `-T` again for tagged.

after the parse, you should be able to tell which op it is using based on the hints

fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if (optind < argc)
+		opts.dst_addr = argv[optind];
+	
+	if (opts.rma_op == FT_RMA_WRITEDATA)
+		post_writedata = true;
+	
+	hints->ep_attr->type = FI_EP_RDM;
+	hints->caps |= FI_MSG | FI_RMA;
```
shijin-aws, line 216: This should be already handled by `-o`

fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,225 @@
...
+	hints->mode |= FI_CONTEXT | FI_CONTEXT2;
```
shijin-aws, line 217: Do you really need this mode?

fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,225 @@
...
+					printf("server polls cq and exits\n");
+					ret = 0;
+				}
+				goto out;
+			}
+
+			if (ret == -FI_EAGAIN) {
+				continue;
+			} else if (ret == -FI_EAVAIL) {
+				ret = fi_cq_readerr(rxcq, &comp_err, FI_SEND);
```
shijin-aws, line 136: I think we shouldn't need this flag `FI_SEND` anymore.. It can be a 0.

fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -0,0 +1,225 @@
...
+		FT_PRINTERR("ft_init_fabric", -ret);
+		return ret;
+	}
+
+	/*
+	 * The handshake procedure between server and client will happen in
+	 * either ft_exchange_keys() or ft_sync()
+	 */
+	if (post_writedata) {
+		ret = ft_exchange_keys(&remote);
```
shijin-aws, line 62: Notice: `ft_exchange_keys` finally calls `ft_sync()`, which will call `ft_sync_inband(true)`, so it will cause the receiver to always post an outstanding recv buffer. This is ok for RMA test because the rx buffer post shouldn't matter here, unless we will support `FI_RX_CQ_DATA`

fabtests/prov/efa/src/rdm_remote_exit_early.c
jiaxiyan: I don't see this being set in ft_parse_api_opts

shijin-aws (approved):

fabtests/pytest/efa/test_remote_exit_early.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+from common import ClientServerTest
+
+
+@pytest.mark.functional
+def test_remote_exit_early_post_send(cmdline_args):
+    for message_size in [65536, 131072, 1048576]:
```
shijin-aws, line 7: nit: use a fixture as parameters

shijin-aws (approved):

shijin-aws:
This should be pure EFA change and don't need to check Intel CI failure.

PR #10800, opened 2025-02-19
fabtests/efa: Add fabtests for efa-direct

jiaxiyan:
Run fabtests for efa-direct up to device max msg size(8K) where FI_RMA is supported.
RMA tests currently post a recv with the max transfer size. It could run up to max rdma size after fixing fabtests to post the recv within device max msg size.
Also skip 0 byte for rma because fabtests use inject for messages smaller than inject_size, but efa-direct does not support it until firmware supports inline write.

PR #10811, opened 2025-02-21
prov/efa: More fixes for efa-direct

shijin-aws:
A series of fixes for efa-direct:

Tested by https://github.com/ofiwg/libfabric/pull/10800 on p5


prov/efa/src/efa_ep.c
```diff
@@ -227,12 +227,6 @@ static int efa_ep_bind(struct fid *fid, struct fid *bfid, uint64_t flags)
 
 	switch (bfid->fclass) {
 	case FI_CLASS_CQ:
-		if (flags & FI_SELECTIVE_COMPLETION) {
```
sunkuamzn, line 230: Why did we have this in the first place? Because dgram doesn't support FI_SELECTIVE_COMPLETION?

sunkuamzn (approved):

prov/efa/src/efa_ep.c
shijin-aws: Yes

shijin-aws: And the efa_ep PR was merged before the FI_CONTEXT2 change, so I didn't remove it.

Issue #10797, opened 2025-02-19
src/hmem_rocr.c: bug when --enable-rocr-dlopen is enabled


mflough-at-hpe:
**Describe the bug**

When libfabric is configured with "--with-rocr=$ROCM_PATH --enable-rocr-dlopen", three functions ("hsa_ops.hsa_signal_store_screlease", "hsa_ops.hsa_signal_load_scacquire", "hsa_ops.hsa_amd_agents_allow_access") are not mapped to their associated symbols from "libhsa-runtime64.so" in the initialization routine "rocr_hmem_dl_init()". This can lead to a segmentation violation and application crash.

**To Reproduce**

The issue was encountered when building libfabric v2.0.x using the LNX (LINKx) provider with the CXI SlingShot provider and using the associated libfabric library in a build of Open MPI 5.0.6 while running the OSU microbenchmarks.

```
git clone https://github.com/ofiwg/libfabric.git
cd libfabric
git switch v2.0.x
./autogen.sh
./configure --disable-sockets --disable-udp --disable-verbs --disable-rxm --disable-mrail --disable-rxd --disable-tcp --disable-usnic --disa
ble-rstream --disable-efa --disable-psm2 --disable-psm3 --disable-opx --with-json-c=${installJsonCDir} --with-criterion=${TOPDIR}/openmpi5x/libf
abric_work/criterion-2.4.2 --prefix=${TOPDIR}/extra_sw/libfabric-2.0.0 --with-cassini-headers=/usr --with-cxi-uapi-headers=/usr --with-curl=${TO
PDIR}/openmpi5x/libfabric_work/curl --enable-cxi=/usr --enable-shm --enable-xpmem=${XPMEM_ROOT} --with-rocr=${ROCM_PATH} --enable-rocr-dlopen
make
make install
```

The OSU test used with "osu_bibw -r rocm D D" while run on two processes intranode when run on AMD GPU accelerated nodes (MI250X and MI300A both show the same behavior).

**Expected behavior**
If needed, a clear and concise description of what you expected to happen.

The tests run properly if libfabric is configured with "--with-rocr=$ROCM_PATH" but without "--enable-rocr-dlopen".

Additionally, the tests run properly if libfabric is configured with "--with-rocr=$ROCM_PATH --enable-rocr-dlopen" and "src/hmem_rocr.c" is patched and with the following:

```
[loughm@portage-login1 openmpi5x]$ more patches/libfabric_enable_rocr_dlopen.patch 
[...]

PR #10812, opened 2025-02-21
core/util: Fix missing rocr dlopen function assignments

swelch:
Resolves issue #10797

### DIFF
--- a/src/hmem_rocr.c
+++ b/src/hmem_rocr.c
@@ -830,6 +830,30 @@ static int rocr_hmem_dl_init(void)
 		goto err;
 	}
 
+	hsa_ops.hsa_signal_store_screlease = dlsym(hsa_handle,
+						   "hsa_signal_store_screlease");
+	if (!hsa_ops.hsa_signal_store_screlease) {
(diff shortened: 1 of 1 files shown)


j-xiong:
Please change the prefix of the commit title from `core/util: ` to `core/hmem` or just `hmem`.

swelch:
> Thanks for the update. The issue label just puts `core` and `util` into the same category which is different from the commit title convention.
> 
> The updated commit message has the title line repeated in the body. That is not necessary. What I meant was to use the following description from the issue:
> 
> ```
> When libfabric is configured with "--with-rocr=$ROCM_PATH --enable-rocr-dlopen",
> three functions ("hsa_ops.hsa_signal_store_screlease", "hsa_ops.hsa_signal_load_scacquire", 
> "hsa_ops.hsa_amd_agents_allow_access") are not mapped to their associated symbols 
> from "libhsa-runtime64.so" in the initialization routine "rocr_hmem_dl_init()". This can 
> lead to a segmentation violation and application crash.
> ```

IMO this is obvious from the patch and adds no real value to the 1-line; but I'll add the requested text.

j-xiong (approved):

PR #10813, opened 2025-02-21
prov/tcp: Process EPOLLERR properly

jxiong:
The original code just clears MSG_ERRQUEUE but not handle SO_ERROR, which makes the socket always have something to report from epoll.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1417,6 +1417,20 @@ int ofi_bsock_async_done(const struct fi_provider *prov,
 	uint8_t ctrl[CMSG_SPACE(sizeof(*serr) * 2)];
 	int ret;
 
+	int val = 0;
+	socklen_t len = sizeof(val);
+	ret = getsockopt(bsock->sock, SOL_SOCKET, SO_ERROR, &val, &len);
+	if (ret < 0) {
+		FI_WARN(prov, FI_LOG_EP_DATA,
+			"Error reading socket error (%s)\n", strerror(errno));
+		return -errno;
+	}
+	if (val != 0) {
+		FI_WARN(prov, FI_LOG_EP_DATA,
+			"Socket error (%s)\n", strerror(val));
+		return -val;
+	}
+
 	msg.msg_control = &ctrl;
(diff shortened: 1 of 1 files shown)


src/common.c
```diff
@@ -1417,6 +1417,20 @@ int ofi_bsock_async_done(const struct fi_provider *prov,
...
+	socklen_t len = sizeof(val);
+	ret = getsockopt(bsock->sock, SOL_SOCKET, SO_ERROR, &val, &len);
+	if (ret < 0) {
+		FI_WARN(prov, FI_LOG_EP_DATA,
+			"Error reading socket error (%s)\n", strerror(errno));
+		return -errno;
+	}
+	if (val != 0) {
+		FI_WARN(prov, FI_LOG_EP_DATA,
+			"Socket error (%s)\n", strerror(val));
```
jxiong, line 1430: It would be nice if we could fetch remote address here in the log message, something like this:
```
ssize_t get_remote_addr(char *buf, size_t buflen, int sockfd) {
        struct sockaddr_in remote_addr;
        socklen_t addr_len = sizeof(remote_addr);

        if (getpeername(sockfd, (struct sockaddr *)&remote_addr, &addr_len) == -1) {
    strerror_r(errno, buf, buflen);
                return 0;
  }

        char remote_ip[INET_ADDRSTRLEN];
        if (inet_ntop(AF_INET, &(remote_addr.sin_addr), remote_ip, INET_ADDRSTRLEN) == NULL) {
    strerror_r(errno, buf, buflen);
                return 0;
  }

        int remote_port = ntohs(remote_addr.sin_port); // Network to host short

        return snprintf(buf, buflen, "%s:%d", remote_ip, remote_port);
}
```
but I don't know how to make more generic in the context of libfabric

jxiong: @j-xiong @shefty I think this PR is worth landing as well since POLLERR will be set for either error queue or sock error

j-xiong:
@jxiong Could you rebase? There have been bunch of CI changes since the PR was created.

j-xiong (approved):

PR #10817, opened 2025-02-21
prov/shm: new shm architecture

aingerson:
Reopening to pickup new CI changes. This PR should be close and hopefully pass all tests. Ready for review!

### DIFF
--- a/man/fi_shm.7.md
+++ b/man/fi_shm.7.md
@@ -170,6 +170,14 @@ The *shm* provider checks for the following environment variables:
     chunks. This environment variable is provided to fine tune performance
     on different systems. Default 262144
 
+*FI_SHM_BUFFER_THRESHOLD*
+ :  When to start requesting forced unexpected messaging buffering. When this
+    threshold is reached, the sender will notify the receiver to force buffering
+    of the entire message if it is unexpected. If the message is matched when
+    received, it has no effect. Requesting unexpected message buffering allows
+    shm to support unlimited unexpected messaging (memory permitting).
+    Default: 1
+
 # SEE ALSO
 
 [`fabric`(7)](fabric.7.html),
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -147,6 +147,7 @@ struct smr_env {
 	int	use_dsa_sar;
 	size_t	max_gdrcopy_size;
 	int	use_xpmem;
+	size_t	buffer_threshold;
 };
 
 extern struct smr_env smr_env;
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -468,6 +468,10 @@ static ssize_t smr_do_inject(struct smr_ep *ep, struct smr_region *peer_smr,
 	smr_generic_format(cmd, tx_id, rx_id, op, tag, data, op_flags);
 	smr_format_inject(ep, cmd, pend);
 
+	if (smr_freestack_avail(smr_cmd_stack(ep->region)) <=
+	    smr_env.buffer_threshold)
+		cmd->hdr.op_flags |= SMR_BUFFER_RECV;
+
 	return FI_SUCCESS;
 }
 
@@ -488,6 +492,10 @@ static ssize_t smr_do_iov(struct smr_ep *ep, struct smr_region *peer_smr,
 
 	smr_generic_format(cmd, tx_id, rx_id, op, tag, data, op_flags);
 	smr_format_iov(cmd, pend);
+
+	if (smr_freestack_avail(smr_cmd_stack(ep->region)) <=
+	    smr_env.buffer_threshold)
+		cmd->hdr.op_flags |= SMR_BUFFER_RECV;
 
 	return FI_SUCCESS;
 }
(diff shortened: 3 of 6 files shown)


include/ofi_atomic_queue.h
```diff
@@ -135,9 +135,10 @@ static inline void name ## _init(struct name *aq, size_t size)	\
 static inline struct name * name ## _create(size_t size)	\
 {								\
 	struct name *aq;					\
-	aq = (struct name*) calloc(1, sizeof(*aq) +		\
-		sizeof(struct name ## _entry) *			\
-		(roundup_power_of_two(size)));			\
+	aq = (struct name *) aligned_alloc(			\
+			OFI_CACHE_LINE_SIZE, sizeof(*aq) +	\
+			sizeof(struct name ## _entry) *		\
+			(roundup_power_of_two(size)));		\
```
j-xiong, line 141: Now the allocation is uninitialized.  The `init` macro would initialize most of the fields but leave `entry.noop` and `entry.buf` uninitialized. Would that be an issue? especially the `noop` field.

aingerson: The buf doesn't need to be initialized but you're right, the noop should get initialized to false. Good catch, thanks!

aingerson:
@sunkuamzn Could you share the AWS failure? You definitely test more shm than we do so I'm sure I'm just missing a case

sunkuamzn:
@aingerson at the fabtests level, I see the fi_unexpected_msg test timing out

```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.45.3 'timeout 1800 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10817/install/fabtests/bin/fi_unexpected_msg -e rdm -M 2048 -I 5 -v -S 512 -p shm -E=9235'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.45.3 'timeout 1800 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10817/install/fabtests/bin/fi_unexpected_msg -e rdm -M 2048 -I 5 -v -S 512 -p shm -E=9235 172.31.45.3'"'"''
client_stdout:

client returncode: 124
server_stdout:

server returncode: 124
```

At the MPI level, I see some collectives timing out on a single node. Which would make sense if unexpected messages are not being processed correctly.

aingerson:
@sunkuamzn Could you share the AWS failure please? I think I resolved the MPI races (there was one). The unexpected test failure I think is an incorrect use case that just happens to be working upstream. The new implementation doesn't buffer unexpected inject message data anymore and just holds onto the command which means when we have unexpected inject messages, we are still held to the tx size (the maximum number of sends we can have pending at once). Trying to have more than tx size (1024) number of pending sends at the same time is an incorrect use of the API. If you really want to force more messages, you can set FI_SHM_TX_SIZE=2048 which allows the test to pass but I think this test case isn't valid. 
I chatted with @shijin-aws about it and he said there was an application that was hitting this issue but we weren't sure if this was coming from the application or OMPI so he was going to investigate. We can buffer the inject data but this will hurt performance and likely will not find the root of the problem and will just enable the app to have much too many unexpected messages.

shijin-aws:
[main_new_shm_compare_c7gn.txt](https://github.com/user-attachments/files/20372968/main_new_shm_compare_c7gn.txt)
[main_new_shm_compare_c6i.txt](https://github.com/user-attachments/files/20372969/main_new_shm_compare_c6i.txt)
[main_new_shm_compare_hpc6a.txt](https://github.com/user-attachments/files/20372970/main_new_shm_compare_hpc6a.txt)

I re-benchmarked this PR on AMD (hpc6a), Intel (c6i) and ARM (c7gn) CPUs. For the inline sized protocol performance, new shm is almost identical to the main branch on Intel and ARM. It still has a ~ 20 ns shift on AMD

shijin-aws:
bot:aws:retest

shijin-aws:
For the inject protocol (256B - 4KB), Intel and ARM both show improvement on the bandwidth (tagged_bw). AMD is mixed 

```
15                  256                1302.980445               2774.116416                           1471.135971               112.905453      0.389102     0.508273
16                  384                1975.831559               2184.256267                            208.424708                10.548708      0.334450     0.302226
17                  512                2647.785359               2815.213744                            167.428386                 6.323337      1.049448     0.704887
18                  768                3933.024732               4569.192320                            636.167589                16.175021      0.645372     0.314787
19                 1024                5290.717703               5508.609283                            217.891580                 4.118375      0.520046     1.846573
20                 1536                8267.891931               7545.785170                           -722.106761                -8.733868      1.547265     2.353081
21                 2048               10108.174565               9447.355196                           -660.819369                -6.537475      0.788226     2.049158
22                 3072               13394.672777              12228.229398                          -1166.443379                -8.708263      0.577448     1.466927
23                 4096               15594.015821              13690.708081                          -1903.307740               -12.205373      1.027649     1.615210
```

shijin-aws:
bot:aws:retest

shijin-aws:
Paste my offline discussion with @aingerson here. We narrow down the issue to be something in new shm's atomics calls (fi_atomic, fi_fetch_atomic, fi_compare_atomic). If I make efa skip shm in these calls , there is no hang. Need to further narrow down which exact op is causing problem

PR #10819, opened 2025-02-22
fabtests/common: Set the min of tx/rx_mr_size

shijin-aws:
In FT_OPT_ALLOC_MULT_MR mode, currently tx/rx_mr_size is set as the opts.transfer_size, and the subsequent ft_alloc_ctx_array will use this size to allocate
tx/rx buffers in the ctx array.

However, when test call ft_post_rx_buf, it will
finally post a size as MAX(opts.transfer_size, FT_MAX_CTRL_MSG). When opts.transfer_size is smaller than FT_MAX_CTRL_MSG, it will cause test call fi_recv of size larger than the actual buffer (and MR) size and cause error when the provider offload the recv directly to hardware.

This patch fixes this issue by making the tx/rx_mr_size has a min value as the FT_MAX_CTRL_MSG.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -540,6 +540,7 @@ static void ft_set_tx_rx_sizes(size_t *set_tx, size_t *set_rx)
 {
 	*set_tx = opts.options & FT_OPT_SIZE ?
 		  opts.transfer_size : test_size[TEST_CNT - 1].size;
+	*set_tx = MAX(*set_tx, FT_MAX_CTRL_MSG);
 	if (*set_tx > fi->ep_attr->max_msg_size)
 		*set_tx = fi->ep_attr->max_msg_size;
 	*set_rx = *set_tx + ft_rx_prefix_size();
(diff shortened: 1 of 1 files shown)


fabtests/common/shared.c
```diff
@@ -592,6 +592,8 @@ int ft_alloc_msgs(void)
 
 	if (opts.options & FT_OPT_ALLOC_MULT_MR) {
 		ft_set_tx_rx_sizes(&tx_mr_size, &rx_mr_size);
+		tx_mr_size = MAX(FT_MAX_CTRL_MSG, tx_mr_size);
+		rx_mr_size = MAX(FT_MAX_CTRL_MSG, rx_mr_size);
```
aingerson, line 596: I think it makes more sense to set these inside the ft_set_tx_rx_sizes function. doesn't make sense to call a function that is supposed to set it and then we modify them. that we can remove the max from lines 605/606 and just multiply it in that line. I think that makes more sense

shijin-aws: Makes sense, updated

aingerson (approved):
Thanks, Shi!

aingerson:
@shijin-aws Please hold on merging. CI failed fabtests stages for multiple providers. Our cluster is having issues so I need to figure out if it had anything to do with the changes or if it was the cluster was in a bad state.

shijin-aws:
@aingerson @zachdworkin any update on the Intel CI failure?

PR #10820, opened 2025-02-23
prov/cxi: Fix peer CQ support

amirshehataornl:
cxi_cq->util_cq.cq_fid.ops is being overwritten after call to ofi_cq_init() which sets the ops to the peer callbacks. Update the code to only override this structure in the non-peer case.

### DIFF
--- a/prov/cxi/src/cxip_cq.c
+++ b/prov/cxi/src/cxip_cq.c
@@ -395,6 +395,10 @@ static int cxip_cq_verify_attr(struct fi_cq_attr *attr)
 	if (!attr)
 		return FI_SUCCESS;
 
+	if (attr->flags & FI_PEER &&
+	    attr->wait_obj != FI_WAIT_NONE)
+		return -FI_ENOSYS;
+
 	switch (attr->format) {
 	case FI_CQ_FORMAT_CONTEXT:
 	case FI_CQ_FORMAT_MSG:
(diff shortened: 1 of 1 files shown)


amirshehataornl:
@iziemba, can you please review this patch?

prov/cxi/src/cxip_cq.c
```diff
@@ -511,7 +511,8 @@ int cxip_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
 	}
 
 	cxi_cq->util_cq.cq_fid.fid.ops = &cxip_cq_fi_ops;
-	cxi_cq->util_cq.cq_fid.ops = &cxip_cq_ops;
+	if (!(attr->flags & FI_PEER))
+		cxi_cq->util_cq.cq_fid.ops = &cxip_cq_ops;
```
swelch, line 515: The change looks good. However, could add a check in cxip_cq_verify_attr() to return -FI_ENOSYS if the FI_PEER flag is set and the attr->wait_obj is not FI_WAIT_NONE. Our WAIT_FD wait object is reliant on the cxi specific CQ sread/sreadfrom. This might avoid a future incompatibility/error.

iziemba: @amirshehataornl - I'm good with this patch once Steve's comment is addressed. Thanks!

amirshehataornl: @swelch, let me know if the changes look alright to you.

iziemba (approved):

prov/cxi/src/cxip_cq.c
swelch: Thank you

swelch:
@j-xiong looks like your labeling and my merging where simultaneous more or less. Why did you not want to merge this one?

j-xiong:
@swelch I wanted to avoid re-doing the rc1 process. But it's fine since I have to rebase anyway.

Issue #10804, opened 2025-02-20
Unsafe reads of the av_entry_pool because the util_av implementation does not require a lock to be held in some cases


sunkuamzn:
The `util_av` implementation requires a mutex (will be a genlock after https://github.com/ofiwg/libfabric/pull/10803) to be held for when writing to the `av_entry_pool`
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/util/src/util_av.c#L305
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/util/src/util_av.c#L278
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/util/src/util_av.c#L337

but not when reading from the `av_entry_pool`
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/util/src/util_av.c#L243-L249
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/util/src/util_av.c#L419-L424

except for `ofi_av_lookup_fi_addr`
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/util/src/util_av.c#L362-L364

SHM, SM2 (and maybe other providers) are performing unsafe reads of the `av_entry_pool`
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/shm/src/smr_ep.c#L215
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/shm/src/smr_av.c#L242
https://github.com/ofiwg/libfabric/blob/a07fe39451f6897c122cf8350a70a105b354eee4/prov/sm2/src/sm2_ep.c#L169

PR #10823, opened 2025-02-24
prov/efa: Control plane AV operation locking fix

sunkuamzn:
#### prov/efa: Grab srx lock for control plane AV operations
The srx lock is already used in the data path in both tx and rx
operations. Using the same lock for control plane AV operations
will prevent unsafe access from the control plane.

The util_av code still requires the util_av->lock to be held for some
util_av calls. So grab that lock as well for those calls. This
additional lock will not introduce any extra contention because the srx
lock is already serializing the control plane AV calls.

#### prov/efa: Rename domain->srx_lock to domain->progress_lock
In addition to the SRX, this lock now protects read and write access to
the AV. Rename it to progress_lock.

This PR fixes some of the unsafe reads in https://github.com/ofiwg/libfabric/issues/10804 for the EFA provider

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -427,7 +427,7 @@ int efa_av_update_reverse_av(struct efa_av *av, struct efa_ep_addr *raw_addr,
 
 /**
  * @brief allocate an efa_conn object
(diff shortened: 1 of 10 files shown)


sunkuamzn:
Needs more internal discussion

PR #10825, opened 2025-02-25
contrib/aws: Add trn1 fabtests with efa direct

jiaxiyan:
Add coverage for efa-direct tests on trn1, which has unsolicited write recv capability to test rma.


jiaxiyan:
need to add lockable resource for trn first

sunkuamzn (approved):

PR #10827, opened 2025-02-26
prov/efa: Remove efa_av->ep_type in favor of efa_domain->info_type

sunkuamzn:
### prov/efa: Remove efa_av->ep_type in favor of efa_domain->info_type

efa_av->ep_type was used to distinguish between the RDM path and the
DGRAM path. The efa-direct path has FI_EP_RDM ep type but it does not
require a efa_conn->rdm_peer. But the efa-direct path still uses unique connid
based on timestamp.

Use efa_domain->info_type to distinguish the three code paths.

### prov/efa: Replace domain->rdm_ep with domain->info_type

domain->rdm_ep can't distinguish between efa-direct and dgram code paths

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -265,7 +265,7 @@ int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn, bool insert_shm_
 	struct efa_rdm_ep *efa_rdm_ep;
 	struct efa_rdm_peer *peer;
 
-	assert(av->ep_type == FI_EP_RDM);
+	assert(av->domain->info_type == EFA_INFO_RDM);
 	assert(conn->ep_addr);
 
 	/* currently multiple EP bind to same av is not supported */
@@ -347,7 +347,7 @@ void efa_conn_rdm_deinit(struct efa_av *av, struct efa_conn *conn)
 	struct efa_rdm_peer *peer;
 	struct efa_rdm_ep *ep;
 
-	assert(av->ep_type == FI_EP_RDM);
+	assert(av->domain->info_type == EFA_INFO_RDM);
 
 	peer = &conn->rdm_peer;
 	if (peer->is_local && av->shm_rdm_av) {
@@ -408,7 +408,7 @@ int efa_av_update_reverse_av(struct efa_av *av, struct efa_ep_addr *raw_addr,
 	/* We used a static connid for all dgram endpoints, therefore cur_entry should always be NULL,
 	 * and only RDM endpoint can reach here. hence the following assertion
 	 */
-	assert(av->ep_type == FI_EP_RDM);
+	assert(av->domain->info_type == EFA_INFO_RDM);
 	prv_entry = malloc(sizeof(*prv_entry));
 	if (!prv_entry) {
 		EFA_WARN(FI_LOG_AV, "Cannot allocate memory for prv_reverse_av entry\n");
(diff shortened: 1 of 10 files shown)


fabtests/functional/av_xfer.c
```diff
@@ -233,8 +233,7 @@ int main(int argc, char **argv)
 	if (optind < argc)
 		opts.dst_addr = argv[optind];
 
-	hints->caps = hints->ep_attr->type == FI_EP_RDM ?
-		      FI_TAGGED : FI_MSG;
+	hints->caps = FI_MSG;
```
shijin-aws, line 236: This should be a separate PR, at least we shouldn't mention efa-direct as the motivation of this common change

shijin-aws: Even if the test doesn't require tag matching, it's better to allow `FI_MSG` and `FI_TAGGED` toggling via `-o`

sunkuamzn: moved the fabtests commit to https://github.com/ofiwg/libfabric/pull/10828

prov/efa/test/efa_unit_test_av.c
```diff
@@ -4,38 +4,6 @@
 #include "efa_unit_tests.h"
 #include "efa_av.h"
 
-/**
- * @brief Verify the ep type in struct efa_av for efa RDM path
- *
- * @param[in]	state		struct efa_resource that is managed by the framework
- */
-void test_av_ep_type_efa_rdm(struct efa_resource **state)
```
shijin-aws, line 12: This change should belong to your next commit?

prov/efa/src/efa_av.c
```diff
@@ -272,8 +272,13 @@ int efa_conn_rdm_init(struct efa_av *av, struct efa_conn *conn, bool insert_shm_
 	assert(!dlist_empty(&av->util_av.ep_list));
 	efa_rdm_ep = container_of(av->util_av.ep_list.next, struct efa_rdm_ep, base_ep.util_ep.av_entry);
 
-	peer = &conn->rdm_peer;
-	efa_rdm_peer_construct(peer, efa_rdm_ep, conn);
+	peer = (struct efa_rdm_peer *)calloc(1, sizeof(struct efa_rdm_peer));
```
shijin-aws, line 275: peer is a big object, should we use bufferpool to optimize the allocation?

sunkuamzn: Tests are timing out now, so bufpool is not optional

prov/efa/src/efa_av.c
```diff
@@ -900,13 +904,32 @@ int efa_av_open(struct fid_domain *domain_fid, struct fi_av_attr *attr,
...
+				.size = sizeof(struct efa_rdm_peer),
+				.alignment = 16,
+				.chunk_cnt = attr->count,
+				.max_cnt = 0,
+				/* Don't track buffer use because user can close
+				 * the AV without removing addresses */
+				.flags = OFI_BUFPOOL_NO_TRACK,
+			};
+
+			ret = ofi_bufpool_create_attr(&rdm_peer_pool_attr, &av->rdm_peer_pool);
```
shijin-aws, line 929: Why it is only created when we have shm? Should it bemoved to L916?

sunkuamzn:
bot:aws:retest

jiaxiyan (approved):

PR #10829, opened 2025-02-26
prov/cxi: Added collectives logical operators

SeanP-2023:
Added or/and/xor logical support


swelch (approved):
Thanks, will merge when testing completes

PR #10830, opened 2025-02-28
prov/efa: released matched rxe before destroying the srx rx_pool

jiaxiyan:
If the sender exits early, there may still be unreleased rxe in the srx->rx_pool during util_srx_close, which will cause an
assertion error when the pool is destroyed. Release the matched rxe before calling util_srx_close.

fabtests/efa: Add remote exit early test with post recv
When rdma read is available and message size >= 1M, long read or runt read protocol is used and server is expected to get a cq entry or cq error.
Otherwise, if long CTS is used and sender exits before sending CTS data, receiver is expected to timeout after sending the CTS packet without getting a cq entry or cq error.

### DIFF
--- a/fabtests/prov/efa/src/rdm_remote_exit_early.c
+++ b/fabtests/prov/efa/src/rdm_remote_exit_early.c
@@ -34,7 +34,9 @@
 #include <shared.h>
 #include <stdio.h>
 #include <stdlib.h>
+
 #include <rdma/fi_tagged.h>
+#include <rdma/fi_ext.h>
 
 static bool post_rx = false;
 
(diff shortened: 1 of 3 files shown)


fabtests/prov/efa/src/rdm_remote_exit_early.c
```diff
@@ -203,6 +218,9 @@ int main(int argc, char **argv)
 		case LONG_OPT_POST_RX:
 			post_rx = true;
 			break;
+		case LONG_OPT_USE_RDMA_READ:
```
shijin-aws, line 221: You don't have to use such argument. You can query whether rdma read is available by calling `fi_getopt(... FI_OPT_EFA_EMULATED_READ, ...)`. If it is false, it means rdma read is available.

shijin-aws (approved):

PR #10838, opened 2025-03-03
prov/efa: Make rxe_map part of the efa_rdm_peer struct

sunkuamzn:
This change is necessary to implement an implicit AV.
With an implicit AV, using the fi_addr is no longer an option for
identifying different peers. All unknown peers will have fi_addr set
to -1. And the index in the implicit AV can be re-used, so it's not
guaranteed to be unique.

This commit makes the rxe_map part of the peer struct and only uses the
msg_id as the hash key

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_rxe_map.c
+++ b/prov/efa/src/rdm/efa_rdm_rxe_map.c
@@ -7,23 +7,19 @@
 #include "efa_rdm_pke_rtm.h"
 
 /**
- * @brief find an RX entry for a received RTM packet entry's sender address and msg_id
+ * @brief find an RX entry for a received RTM packet entry's msg_id
  *
  * @param[in]		rxe_map		RX entry map
- * @param[in]		pkt_entry	received packet entry
+ * @param[in]		msg_id		message id of the received packet
  * @returns
  * pointer to an RX entry. If such RX entry does not exist, return NULL
 */
 struct efa_rdm_ope *efa_rdm_rxe_map_lookup(struct efa_rdm_rxe_map *rxe_map,
-					   uint64_t msg_id, fi_addr_t addr)
+					   uint64_t msg_id)
 {
 	struct efa_rdm_rxe_map_entry *entry = NULL;
-	struct efa_rdm_rxe_map_key key;
 
-	memset(&key, 0, sizeof(key));
-	key.msg_id = msg_id;
-	key.addr = addr;
-	HASH_FIND(hh, rxe_map->head, &key, sizeof(struct efa_rdm_rxe_map_key), entry);
+	HASH_FIND(hh, rxe_map->head, &msg_id, sizeof(msg_id), entry);
 	return entry ? entry->rxe : NULL;
 }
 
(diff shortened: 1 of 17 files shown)


sunkuamzn:
bot:aws:retest

sunkuamzn:
bot:aws:retest

sunkuamzn:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_rxe_map.c
```diff
@@ -7,40 +7,35 @@
 #include "efa_rdm_pke_rtm.h"
 
 /**
- * @brief find an RX entry for a received RTM packet entry's sender address and msg_id
+ * @brief find an RX entry for a received RTM packet entry's msg_id
  *
  * @param[in]		rxe_map		RX entry map
- * @param[in]		pkt_entry	received packet entry
+ * @param[in]		msg_id	message id of the received packet
```
a-szegel: nit: spacing here looks off

prov/efa/src/rdm/efa_rdm_rxe_map.c
```diff
@@ -7,40 +7,35 @@
...
  * @brief insert an RX entry into an RX entry map
  *
  * @details
- * the insertion will use the combination of packet entry sender address and msg_id as key.
+ * the insertion will use the msg_id as key.
  * Caller is responsible to make sure the key does not exist in the map.
  *
  * @param[in,out]	rxe_map		RX entry map
- * @param[in]		pkt_entry	received RTM packet
+ * @param[in]		msg_id	message id of the received packet
```
a-szegel: nit: spacing here looks off

prov/efa/src/rdm/efa_rdm_rxe_map.c
```diff
@@ -52,50 +47,41 @@ void efa_rdm_rxe_map_insert(struct efa_rdm_rxe_map *rxe_map,
...
 	{
 		struct efa_rdm_rxe_map_entry *existing_entry = NULL;
 
-		HASH_FIND(hh, rxe_map->head, &entry->key, sizeof(struct efa_rdm_rxe_map_key), existing_entry);
+		HASH_FIND(hh, rxe_map->head, &msg_id, sizeof(msg_id), existing_entry);
 		assert(!existing_entry);
 	}
 #endif
 
+	entry->msg_id = msg_id;
```
a-szegel, line 59: It is a little weird to have the key of the map entry be part of the contents of the entry.  Is this an artifact of the hash map library?  Or do we do this so we can pass the entry through functions without the key (save a function parameter)?

a-szegel (dismissed):
looks good, 2 small nits (optional) and a question.

prov/efa/src/rdm/efa_rdm_rxe_map.c
sunkuamzn: fixed

prov/efa/src/rdm/efa_rdm_rxe_map.c
sunkuamzn: fixed

prov/efa/src/rdm/efa_rdm_rxe_map.c
sunkuamzn: That's how uthash works. The hashmap is made up of one struct with both the key and the value as members https://troydhanson.github.io/uthash/

jiaxiyan (approved):

PR #10844, opened 2025-03-03
prov/cxi: Bump internal CXI version to support 2.1

swelch:
CXI requires update of internal define to support 2.1


swelch:
@j-xiong - Can this fix get into libfabric 2.1?

j-xiong:
yes.

j-xiong (approved):

iziemba (approved):

PR #10845, opened 2025-03-03
prov/shm: free all pending srx entries on ep close

aingerson:
When an EP is closed, we may have rx_entries from the owner srx that still need to be freed. If we don't, the owner srx close may return an error since not all memory has been freed.

Matched inline, inject, CMA, and IPC transfers will all be freed immediately after matching but matched SAR messages that need multiple round trips need to be afreed asynchronously. Iterate through the sar_list and free the rx_entry with the owner srx if needed.

Unexpected messages are still assigned an rx_entry through the owner srx that needs to be freed. Add a list for all cmd_ctxs and free all outstanding unmatched rx_entries before closing.


aingerson:
@shijin-aws Let me know if this isn't what you had in mind or if I missed a case!

prov/shm/src/smr_ep.c
```diff
@@ -805,6 +807,21 @@ static int smr_ep_close(struct fid *fid)
...
+				entry);
+		if (pend->rx_entry)
+			ep->srx->owner_ops->free_entry(pend->rx_entry);
+		ofi_buf_free(pend);
+	}
+
+	while (!dlist_empty(&ep->unexp_cmd_list)) {
+		dlist_pop_front(&ep->unexp_cmd_list, struct smr_cmd_ctx,
+				cmd_ctx, entry);
+		ep->srx->owner_ops->free_entry(pend->rx_entry);
```
j-xiong, line 821: `pend` --> `cmd_ctx`

aingerson: oops thanks!

aingerson:
@shijin-aws Can you share the AWS failure? it looks ok on our side

sunkuamzn:
@aingerson our CI had problems. This commit failed on a test that doesn't use SHM, so the commit is good.

sunkuamzn:
bot:aws:retest

j-xiong (approved):

PR #10846, opened 2025-03-04
util/bufpool: Add new flag to skip zeroing new memory allocations

belynam:
When a buf pool allocates a new chunk of memory, the entire memory
chunk is zeroed out via a call to memset. This can be quite costly,
occur at inconvenient times, and end up being wasteful when buf pool
users initialize their own entries.
    
This change adds a new OFI_BUFPOOL_NO_ZERO flag and will skip
the memset when that flag is set.

### DIFF
--- a/include/ofi_mem.h
+++ b/include/ofi_mem.h
@@ -327,6 +327,7 @@ enum {
 	OFI_BUFPOOL_NO_TRACK		= 1 << 2,
 	OFI_BUFPOOL_HUGEPAGES		= 1 << 3,
 	OFI_BUFPOOL_NONSHARED		= 1 << 4,
+	OFI_BUFPOOL_NO_ZERO		= 1 << 5,
 };
 
 struct ofi_bufpool_region;
--- a/prov/util/src/util_buf.c
+++ b/prov/util/src/util_buf.c
@@ -142,7 +142,7 @@ int ofi_bufpool_grow(struct ofi_bufpool *pool)
 	size_t mem_allocated = 0;
 
 	FI_DBG(&core_prov, FI_LOG_CORE, "%s pool %p  size %ld region_size %zu "
-	       "entry_cnt %d chuck_cnt %d\n",
+	       "entry_cnt %d chunk_cnt %d\n",
 		__func__, pool, pool->alloc_size, pool->region_size,
 		(int)(pool->entry_cnt+pool->attr.chunk_cnt),
 		(int)pool->attr.chunk_cnt);
@@ -167,7 +167,8 @@ int ofi_bufpool_grow(struct ofi_bufpool *pool)
 
 	mem_allocated += buf_region->pool->alloc_size;
 
-	memset(buf_region->alloc_region, 0, pool->alloc_size);
+	if (!(pool->attr.flags & OFI_BUFPOOL_NO_ZERO))
+		memset(buf_region->alloc_region, 0, pool->alloc_size);
 	buf_region->mem_region = buf_region->alloc_region + pool->entry_size;
 	if (pool->attr.alloc_fn) {
 		ret = pool->attr.alloc_fn(buf_region);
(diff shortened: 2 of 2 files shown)


prov/util/src/util_buf.c
```diff
@@ -269,8 +270,7 @@ int ofi_bufpool_create_attr(struct ofi_bufpool_attr *attr,
 	pool->region_size = pool->alloc_size - pool->entry_size;
 
 	FI_DBG(&core_prov, FI_LOG_CORE,
-		"%s alloc_size %zu region_size %zu align_entry %zu "
-		"entry_size %zu chuck_cnt %ld pool %p  (%p)\n",
+		"%s alloc_size %zu region_size %zu align_entry %zu entry_size %zu chunk_cnt %ld pool %p  (%p)\n",
```
shefty, line 273: line length is too long, please revert change

include/ofi_mem.h
```diff
@@ -602,6 +602,15 @@ static inline void *ofi_ibuf_alloc_at(struct ofi_bufpool *pool, size_t index)
...
 
+/*
+ * Empty buf pool init function that can be used to avoid initializing
+ * newly allocated buf pool entries.
+ */
+static inline void ofi_bufpool_init_none(struct ofi_bufpool_region *region, void *buf)
+{
+	/* This function is purposely empty */
+}
+
```
shefty, line 613: This won't be a static inline function.  The function is passed by referenced, so actually needs to live somewhere.  I'd call this 'noop' instead of 'none'.

prov/util/src/util_buf.c
shefty: Yes, I know the kernel guidelines.  The print includes the file and line number anyway, so grep isn't needed.  Besides, it's not like someone can take the print output and toss it into grep anyway.  They have to know how and where to convert portions of the output into a regular expression.

Please keep the typo fix, but not make me need to scroll left/right in my editor.  My eyeballs hate that.

j-xiong:
Please squash the commits.

j-xiong:
Appveyor tests are failing. They are likely related.

prov/util/src/util_buf.c
```diff
@@ -167,7 +167,8 @@ int ofi_bufpool_grow(struct ofi_bufpool *pool)
 
 	mem_allocated += buf_region->pool->alloc_size;
 
-	memset(buf_region->alloc_region, 0, pool->alloc_size);
+	if (!pool->attr.init_fn)
+		memset(buf_region->alloc_region, 0, pool->alloc_size);
```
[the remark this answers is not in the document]
belynam: > This changes the existing behavior where the caller may be expecting that the region has already been initialized to zero. To preserve existing behavior, you'd want:
> 
> ```
> if (pool->attr.init_fn != ofi_bufpool_init_noop)
>     memset(..., 0, ...)
> ```
> 
> The alternative is to check and update all init functions to zero the memory, if needed.

I originally went this route of basing the memset off the lack of an init function so that users without an init function wouldn't break, and those who had one would automatically gain a performance benefit by not initializing their memory twice. Hinging the memset off the init function being ofi_bufpool_init_noop specifically means users with their own init function are still going to get some of their memory initialized twice. But to your point, we can't assume the presence of an init function means they aren't still relying on the memset being run first.

The alternative approach I considered that's looking more attractive to me now is to add a new flag and hinge the memset off that rather than the presence of an init function.

Another revision of this PR forthcoming...

prov/util/src/util_buf.c
```diff
@@ -317,3 +318,8 @@ int ofi_ibufpool_region_is_lower(struct dlist_entry *item, const void *arg)
 
 	return reg1->index < reg2->index;
 }
+
+void ofi_bufpool_init_noop(struct ofi_bufpool_region *region, void *buf)
+{
+	/* This function is purposely empty */
+}
```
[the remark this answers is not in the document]
belynam: > Is there a subsequent change that will use this function?

There was going to be, but N/A now given the different approach I'll be pursuing.

j-xiong (approved):

shefty (approved):

PR #10861, opened 2025-03-11
prov/cxi: fix ss_plugin_auth_key_priority test

chuckfossen:
The cxi-driver no longer allows services with overlapping VNIs so use a different VNI than the default.


swelch (approved):

swelch:
@j-xiong - If you end up needing to do another RC for 2.1 then we would like to pick this up; otherwise we can cherry-pick into 2.1.x following the release (it is a cxi unit test). @iziemba

PR #10884, opened 2025-03-18
prov/util: Only allocate MR key when needed

### LINKED ISSUE #10882: prov/tcp: fi_mr_reg fails when requested_key is unchanged even when the buffer is not used for RMA operations
**Describe the bug**
According to Sean here https://github.com/ofiwg/libfabric/pull/8350#issuecomment-1349946033
> A key is only needed and used if the memory will be the target of an RMA operation.

So if an endpoint does not have FI_RMA and FI_ATOMIC capabilities or if the buffer access flags are only FI_SEND and FI_RECV, then the key is not relevant or useful. Yet today, an `fi_mr_reg` call to the TCP provider will fail unless the `attr->requested_key` is unique for each call

The failure happens here https://github.com/ofiwg/libfabric/blob/996a6c1911ca91a8cb9367da854c09086963303b/prov/util/src/util_mr_map.c#L82-L90

What should be the behavior of `fi_mr_reg` and the remote key when the buffer is not meant for RMA or atomic usage? If the key doesn't matter unless the buffer is used for RMA or atomic operations, shouldn't the TCP provider check that and not fail `fi_mr_reg` in such cases?

**To Reproduce**
Call `fi_mr_regattr` without changing `attr->requested_key`

**Expected behavior**
`fi_mr_regattr` succeeds

**Additional context**
Similar to https://github.com/ofiwg/libfabric/issues/10821

j-xiong:
MR key is only needed for remote access. If remote access rights are not asked
for, MR registration should succeed even if key allocation fails. To prevent
key confliction with a future registration that needs the key, don't even try
to allocation the key if it is not needed.

A special key, FI_KEY_NOTAVAIL, is assigned for such case.

Fixes #10882

### DIFF
--- a/prov/util/src/util_mr_map.c
+++ b/prov/util/src/util_mr_map.c
@@ -222,16 +222,19 @@ void ofi_mr_map_close(struct ofi_mr_map *map)
 int ofi_mr_close(struct fid *fid)
 {
 	struct ofi_mr *mr;
-	int ret;
+	int ret = 0;
 
 	mr = container_of(fid, struct ofi_mr, mr_fid.fid);
+
+	if (mr->key == FI_KEY_NOTAVAIL)
+		goto free_mr;
 
 	ofi_genlock_lock(&mr->domain->lock);
 	ret = ofi_mr_map_remove(&mr->domain->mr_map, mr->key);
 	ofi_genlock_unlock(&mr->domain->lock);
 	if (ret)
 		return ret;
-
+free_mr:
 	ofi_atomic_dec32(&mr->domain->ref);
 	free(mr);
 	return 0;
@@ -343,10 +346,14 @@ int ofi_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	mr->device = cur_abi_attr.device.reserved;
 	mr->hmem_data = cur_abi_attr.hmem_data;
 
-	ret = ofi_mr_map_insert(&domain->mr_map, &cur_abi_attr, &key, mr, flags);
-	if (ret) {
-		free(mr);
-		goto out;
+	if (cur_abi_attr.access & (FI_REMOTE_READ | FI_REMOTE_WRITE)) {
+		ret = ofi_mr_map_insert(&domain->mr_map, &cur_abi_attr, &key, mr, flags);
+		if (ret) {
+			free(mr);
+			goto out;
+		}
+	} else {
+		key = FI_KEY_NOTAVAIL;
 	}
 
 	mr->mr_fid.key = mr->key = key;
--- a/man/fi_mr.3.md
+++ b/man/fi_mr.3.md
@@ -119,6 +119,8 @@ int fi_hmem_ze_device(int driver_index, int device_index);
 *requested_key*
 : Requested remote key associated with registered buffers.  Parameter
   is ignored if FI_MR_PROV_KEY flag is set in the domain mr_mode bits.
+  Parameter may be ignored if remote access permissions are not asked
+  for.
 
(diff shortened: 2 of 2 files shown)


prov/util/src/util_mr_map.c
```diff
@@ -82,6 +83,8 @@ int ofi_mr_map_insert(struct ofi_mr_map *map, const struct fi_mr_attr *attr,
 	if (map->mode & FI_MR_PROV_KEY)
 		item->requested_key = map->key++;
 
+	key_is_optional = !(attr->access & (FI_REMOTE_READ | FI_REMOTE_WRITE));
```
sunkuamzn, line 86: Should `FI_READ` and `FI_WRITE` also be included?

prov/util/src/util_mr_map.c
```diff
@@ -82,6 +83,8 @@ int ofi_mr_map_insert(struct ofi_mr_map *map, const struct fi_mr_attr *attr,
...
+
 	ret = ofi_rbmap_insert(map->rbtree, &item->requested_key, item, NULL);
```
sunkuamzn, line 88: The first insertion will succeed because the key is unique. Meaning `fi_mr_key` will return the `requested_key` from mr_attr for the first region and `FI_KEY_NOTAVAIL` for the others.

Also, if there is a future `fi_mr_reg` call for the same region with different access flags, they will overlap in the rb tree.

I think we should not insert into the rb tree if `key_is_optional` is true.

sunkuamzn:
Our CI failure is unrelated. We are trying to fix.

prov/util/src/util_mr_map.c
j-xiong: FI_READ and FI_WRITE are for local access only (the result buffer for fi_read, and the source buffer for fi_write).

prov/util/src/util_mr_map.c
j-xiong: Updated.

sunkuamzn (approved):

shefty:
What is the purpose of registering memory with the tcp provider if not for remote access?  Nothing will touch the region.  The tcp provider could either treat such calls as no-ops (return some static struct) or just fail with EWHATISTHEUSERDOING

j-xiong:
If we add FI_HMEM support to the tcp provider, the local desc will still be useful.

This PR though, is not limited to the tcp provider. The utility code is also used by efa, rxm, and ucx, which can potentially encounter the same issue.

j-xiong:
@shefty Note added to the man page.

sunkuamzn:
bot:aws:retest

sunkuamzn:
@j-xiong the EFA provider actually needs to insert all buffers into the MR map because (1) we implicitly use RDMA protocols for some send/receive operations and (2) we're working around the CUDA memory monitor issue in https://github.com/ofiwg/libfabric/issues/8081

Because these are specific to the EFA provider, we added a fix in EFA provider code in https://github.com/ofiwg/libfabric/pull/10892 Please rebase on that commit and our CI should pass

shefty (approved):

j-xiong:
@shefty @sunkuamzn Update the PR to handle the key in `ofi_mr_regattr` instead of `ofi_mr_map_insert`. This should make #10892 unnecessary.

shefty (approved):
I like this approach better.

sunkuamzn:
@j-xiong this fix is working for us, thanks!

Can we backport this to v2.1.x branch?

j-xiong:
@sunkuamzn Sure.

PR #10886, opened 2025-03-19
prov/verbs: Reload the list of interfaces on each call to fi_getinfo()

### LINKED ISSUE #10881: prov/verbs: Reload the list of verbs devices on each call to fi_getinfo()
The verbs provider loads the list of verbs devices available on the node only once, on the first call to `fi_getinfo()`. If a network device is added (hot-plugged) after this initial call to `fi_getinfo()`, it won't be visible in libfabric. 
A subsidiary problem is that `fi_getinfo()` only returns network adapters with active links:
[[		if (!ifa->ifa_addr || !(ifa->ifa_flags & IFF_UP) || ](https://github.com/ofiwg/libfabric/blob/996a6c1911ca91a8cb9367da854c09086963303b/prov/verbs/src/verbs_info.c#L1152)](https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L1152)
If the link is initially inactive and becomes active after the first call to `fi_getinfo()`, this interface will not be visible in libfabric.

This is a particularly a problem for long-running services where restarting a process to discover newly added network devices is not an option.

I would like the list of verbs devices to be updated on each call to `fi_getinfo()`

**Describe alternatives you've considered**
Restarting the process is the only work-around I know.

**Additional context**

sydidelot:
The verbs provider loads the list of verbs devices available on the node
only once, on the first call to fi_getinfo().
If a network device is added (hot-plugged) after this initial call to
fi_getinfo(), it won't be visible in libfabric.

A subsidiary problem is that fi_getinfo() only returns network
adapters with active links.
If the link is initially inactive and becomes active after the first call to
fi_getinfo(), this interface will not be visible in libfabric.

This is a particularly a problem for long-running services where restarting
a process to discover newly added network devices is not an option.

With this patch, the list of verbs interfaces is reloaded on each call
to fi_getinfo().

Fixes #10881

### DIFF
--- a/prov/rxd/src/rxd_init.c
+++ b/prov/rxd/src/rxd_init.c
@@ -43,6 +43,7 @@ struct rxd_env rxd_env = {
 	.retry		= 1,
 	.max_peers	= 1024,
 	.max_unacked	= 128,
+	.rescan		= -1,
 };
 
 char *rxd_pkt_type_str[] = {
@@ -55,6 +56,7 @@ static void rxd_init_env(void)
 	fi_param_get_bool(&rxd_prov, "retry", &rxd_env.retry);
 	fi_param_get_int(&rxd_prov, "max_peers", &rxd_env.max_peers);
 	fi_param_get_int(&rxd_prov, "max_unacked", &rxd_env.max_unacked);
+	fi_param_get_bool(&rxd_prov, "rescan", &rxd_env.rescan);
 }
 
 void rxd_info_to_core_mr_modes(uint32_t version, const struct fi_info *hints,
@@ -119,6 +121,10 @@ static int rxd_getinfo(uint32_t version, const char *node, const char *service,
 			uint64_t flags, const struct fi_info *hints,
 			struct fi_info **info)
 {
+	if (rxd_env.rescan > 0) /* Explicitly enabled */
+		flags |= FI_RESCAN;
+	else if (!rxd_env.rescan) /* Explicitly disabled */
+		flags &= ~FI_RESCAN;
 	return ofix_getinfo(version, node, service, flags, &rxd_util_prov,
 			    hints, rxd_info_to_core, rxd_info_to_rxd, info);
 }
@@ -147,6 +153,10 @@ RXD_INI
 			"Maximum number of peers to track (default: 1024)");
 	fi_param_define(&rxd_prov, "max_unacked", FI_PARAM_INT,
 			"Maximum number of packets to send at once (default: 128)");
+	fi_param_define(&rxd_prov, "rescan", FI_PARAM_BOOL,
+			"Force or disable rescanning for network interface changes. "
+			"Setting this to true will force rescanning on each fi_getinfo() invocation; "
+			"setting it to false will disable rescanning. (default: unset)");
 
 	rxd_init_env();
 
--- a/prov/rxm/src/rxm_init.c
+++ b/prov/rxm/src/rxm_init.c
@@ -59,6 +59,7 @@ int rxm_passthru = 0; /* disable by default, need to analyze performance */
 int force_auto_progress;
 int rxm_use_write_rndv;
 int rxm_detect_hmem_iface;
+int rxm_rescan = -1;
 enum fi_wait_obj def_wait_obj = FI_WAIT_FD, def_tcp_wait_obj = FI_WAIT_UNSPEC;
 
 char *rxm_proto_state_str[] = {
@@ -552,6 +553,10 @@ static int rxm_getinfo(uint32_t version, const char *node, const char *service,
 	if (ret)
 		return ret;
 
+	if (rxm_rescan > 0) /* Explicitly enabled */
+		flags |= FI_RESCAN;
+	else if (!rxm_rescan) /* Explicitly disabled */
+		flags &= ~FI_RESCAN;
 	ret = ofix_getinfo(version, node, service, flags, &rxm_util_prov, hints,
 			   rxm_info_to_core, rxm_info_to_rxm, info);
 	if (ret)
@@ -703,6 +708,11 @@ RXM_INI
 			"in. This allows such buffers be copied or registered "
 			"internally by RxM. (default: false).");
 
+	fi_param_define(&rxm_prov, "rescan", FI_PARAM_BOOL,
+			"Force or disable rescanning for network interface changes. "
+			"Setting this to true will force rescanning on each fi_getinfo() invocation; "
+			"setting it to false will disable rescanning. (default: unset)");
+
 	/* passthru supported disabled - to re-enable would need to fix call to
 	 * fi_cq_read to pass in the correct data structure.  However, passthru
 	 * will not be needed at all with in-work tcp changes.
@@ -729,6 +739,7 @@ RXM_INI
 			"level would be set to FI_THREAD_SAFE\n");
 
 	fi_param_get_bool(&rxm_prov, "detect_hmem_iface", &rxm_detect_hmem_iface);
+	fi_param_get_bool(&rxm_prov, "rescan", &rxm_rescan);
 
 #if HAVE_RXM_DL
 	ofi_mem_init();
--- a/prov/verbs/src/verbs_domain.c
+++ b/prov/verbs/src/verbs_domain.c
@@ -313,10 +313,14 @@ vrb_get_verbs_info(const char *domain_name)
 {
 	const struct fi_info *fi;
 
+	ofi_mutex_lock(&vrb_info_mutex);
 	for (fi = vrb_util_prov.info; fi; fi = fi->next) {
-		if (!strcmp(fi->domain_attr->name, domain_name))
+		if (!strcmp(fi->domain_attr->name, domain_name)) {
+			ofi_mutex_unlock(&vrb_info_mutex);
 			return fi_dupinfo(fi);
+		}
 	}
+	ofi_mutex_unlock(&vrb_info_mutex);
 
 	return NULL;
 }
@@ -523,13 +527,15 @@ int vrb_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric,
 		  void *context)
 {
 	struct vrb_fabric *fab;
-	const struct fi_info *cur, *info = vrb_util_prov.info;
+	const struct fi_info *cur, *info;
 	int ret = FI_SUCCESS;
 
 	fab = calloc(1, sizeof(*fab));
 	if (!fab)
 		return -FI_ENOMEM;
 
+	ofi_mutex_lock(&vrb_info_mutex);
+	info = vrb_util_prov.info;
 	for (cur = info; cur; cur = info->next) {
 		ret = ofi_fabric_init(&vrb_prov, cur->fabric_attr, attr,
 				      &fab->util_fabric, context);
@@ -537,11 +543,13 @@ int vrb_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric,
 			break;
 	}
 	if (ret) {
+		ofi_mutex_unlock(&vrb_info_mutex);
 		free(fab);
 		return ret;
 	}
 
 	fab->info = fi_dupinfo(cur);
+	ofi_mutex_unlock(&vrb_info_mutex);
 
 	*fabric = &fab->util_fabric.fabric_fid;
 	(*fabric)->fid.fclass = FI_CLASS_FABRIC;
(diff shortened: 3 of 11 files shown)


shefty:
Re-scanning the interfaces is expensive.  Utility providers will force scanning multiple times.

sydidelot:
@shefty would it be possible to give the application control over when the interfaces should be re-scanned? 
For example using an environment variable or a flag to fi_getinfo()?

shefty:
@sydidelot - Giving control to the app is the best idea I could come up with to handle this.  Going that route, I'd define a flag to pass into fi_getinfo().  An alternative I could think of that wouldn't involve application changes is to make refresh timer based.  I'm not thrilled with that option.  An environment variable to set the fi_getinfo() flag seems better.  I think that could allow the util providers to unset the flag after it first passed to the core provider.

shefty:
I was suggesting defining a flag that can be passed into fi_getinfo().  When set, a new scan is performed.

The utility providers can always clear that flag, assuming that the core provider is called directly first, before any utility providers process fi_getinfo.  (This is the case).

In addition, an environment variable could be set which the libfabric core can check.  If set, the libfabric core would add this flag when calling the core provider's getinfo handler.

sydidelot:
Thanks for the clarification @shefty! I didn't know that the core provider was called first.
I will incorporate these changes to the PR.

prov/verbs/src/verbs_domain.c
```diff
@@ -325,19 +325,23 @@ vrb_domain(struct fid_fabric *fabric, struct fi_info *info,
...
 			      util_fabric.fabric_fid);
-	const struct fi_info *fi = vrb_get_verbs_info(vrb_util_prov.info,
-							 info->domain_attr->name);
-	if (!fi)
-		return -FI_EINVAL;
+	struct fi_info *fi = vrb_get_verbs_info(vrb_util_prov.info,
+						info->domain_attr->name);
+	if (!fi) {
+		ret = -FI_EINVAL;
+		goto err0;
```
shefty, line 334: If fi == null, we can just return an error here, rather than call fi_freeinfo() on the null value.

prov/verbs/src/verbs_info.c
```diff
@@ -1916,14 +1914,18 @@ int vrb_getinfo(uint32_t version, const char *node, const char *service,
 {
 	int ret;
 
+	ofi_mutex_lock(&vrb_info_mutex);
```
shefty, line 1918: Can you move this after vrb_prof_func_start() to avoid including that call within the mutex?  I want to avoid someone thinking the mutex is needed for it.

prov/verbs/src/verbs_domain.c
```diff
@@ -521,18 +523,21 @@ int vrb_fabric(struct fi_fabric_attr *attr, struct fid_fabric **fabric,
 	if (!fab)
 		return -FI_ENOMEM;
 
+	ofi_mutex_lock(&vrb_info_mutex);
```
shefty, line 526: This assignment ```*info = vrb_util_prov.info;``` was made in the variable declaration section outside of holding the mutex.  It needs to move inside the lock now.

prov/verbs/src/verbs_domain.c
```diff
@@ -325,8 +325,10 @@ vrb_domain(struct fid_fabric *fabric, struct fi_info *info,
 	struct vrb_fabric *fab =
 		 container_of(fabric, struct vrb_fabric,
 			      util_fabric.fabric_fid);
+	ofi_mutex_lock(&vrb_info_mutex);
 	struct fi_info *fi = vrb_get_verbs_info(vrb_util_prov.info,
 						info->domain_attr->name);
+	ofi_mutex_unlock(&vrb_info_mutex);
```
shefty, line 331: vrb_get_verbs_info() is only called from here.  Can you move the function into this file (or integrate the functionality inline), only pass in the domain name, and take the mutex within the function?

prov/rxd/src/rxd_init.c
```diff
@@ -119,6 +121,10 @@ static int rxd_getinfo(uint32_t version, const char *node, const char *service,
 			uint64_t flags, const struct fi_info *hints,
 			struct fi_info **info)
 {
+	if (rxd_env.always_rescan)
+		flags |= FI_RESCAN;
+	else
+		flags &= ~FI_RESCAN;
```
shefty, line 127: If the user doesn't set any environment variable, this has the impact of removing the FI_RESCAN flag from their request.  That's probably not what we want.  I would leave the flags unchanged in this case.

prov/rxm/src/rxm_init.c
```diff
@@ -552,6 +553,10 @@ static int rxm_getinfo(uint32_t version, const char *node, const char *service,
 	if (ret)
 		return ret;
 
+	if (rxm_always_rescan)
+		flags |= FI_RESCAN;
+	else
+		flags &= ~FI_RESCAN;
```
shefty, line 559: Same comment as with rxd

prov/verbs/src/verbs_info.c
```diff
@@ -1349,40 +1349,20 @@ static int vrb_device_has_ipoib_addr(const char *dev_name)
...
-	static bool initialized = false;
 
 	vrb_prof_func_start(__func__);
 
-	if (initialized)
-		goto done;
-
-	initialized = true;
+	vrb_devs_free(verbs_devs);
+	fi_freeinfo((void *)*all_infos);
```
shefty, line 1363: Cast shouldn't be needed.  Please update the function signature to remove the const.

shefty (commented):
I did not review the last patch.  I'll need more time for that.  See comments so far.

prov/rxd/src/rxd_init.c
sydidelot: I updated the code and `always_rescan` is now an integer that can have the following values:
- `-1`: The user didn't set the environment variable
- `0`: rescan explicitly disabled by the environment variable
- `1`: rescan explicitly enabled by the environment variable

Note that the value is now retrieved with `fi_param_get_bool()` but the underlying value is effectively an integer (and not a boolean).

prov/rxm/src/rxm_init.c
sydidelot: Fixed. Thanks!

sydidelot (commented):
@shefty Thanks for the review! I have addressed all your review comments.

prov/verbs/src/verbs_info.c
sydidelot: If we want to remove the "const" in this function, we will also have to change the definition of `util_prov::info` and remove the "const" from there. Here is how the structure is defined right now:
```
struct util_prov {
       const struct fi_provider        *prov;
       const struct fi_info            *info;    <--- here.
       ofi_alter_info_t                alter_defaults;
       const int                       flags;
};
```

Is it what you want?

shefty: Hmm... when that was defined, I think the assumption was that the info list was a static list of structures.  But verbs needs to set this dynamically, including freeing it.  Maybe that const should be removed.  Can you tell if removing it causes an issue in a bunch of other places?

prov/rxd/src/rxd_init.c
shefty: Maybe convert to an enum?  It should be okay to use get_bool, as that will leave the value alone if the variable isn't defined.  I'd rename to just 'rescan' if false means never.

It seems that this would be a common flow (unset, false, true), but I don't see anything similar in tree based on a quick search.

sydidelot:
@j-xiong May I ask you to check why the Intel integration tests are failing? Are the failures related to the PR?
Thank you!

prov/verbs/src/verbs_info.c
sydidelot: Removing the "const" only causes an issue to the verbs provider. I added a commit to remove it as a consequence.

prov/rxd/src/rxd_init.c
sydidelot: I renamed the parameter to just 'rescan' as you suggested but I kept the boolean. Converting it to an enum will require some extra work and I think it is beyond the scope of this PR.

j-xiong:
@sydidelot Please rebase to pick up github action changes.

sydidelot:
@j-xiong Rebased but it's still failing in Intel CI pipeline. Can you please let me know what tests are failing? I assume they are related to my changes.

j-xiong:
There were a lot of failures, probably issue with CI itself. Restarted the tests.

sydidelot:
@j-xiong My PR is still failing and I have no visibility into the tests results.

j-xiong:
Fabtests over rxm/verbs failed with the following error:
```
./include/ofi_lock.h:257: ofi_mutex_unlock: Assertion `lock->in_use' failed.
```

prov/verbs/src/verbs_info.c
```diff
@@ -1911,21 +1894,49 @@ static void vrb_filter_info_by_addr_format(struct fi_info **info, int addr_forma
...
-	ofi_mutex_lock(&vrb_info_mutex);
-	ret = vrb_init_info(&vrb_util_prov.info);
-	if (ret) {
-		ofi_mutex_unlock(&vrb_info_mutex);
-		goto out;
+	if (!init_done || flags & FI_RESCAN) {
+		if (!init_done) {
+			ret = vrb_init();
+			if (ret) {
+				ofi_mutex_unlock(&vrb_info_mutex);
```
j-xiong, line 1926: The mutex unlock does not have a matching lock.

prov/verbs/src/verbs_info.c
```diff
@@ -1911,21 +1894,49 @@ static void vrb_filter_info_by_addr_format(struct fi_info **info, int addr_forma
...
+				goto out;
+			}
+			init_done = true;
+		}
+
+		ret = vrb_init_info(&vrb_devs, &vrb_util_prov.info);
+		if (ret) {
+			ofi_mutex_unlock(&vrb_info_mutex);
```
j-xiong, line 1934: Same here.

prov/verbs/src/verbs_info.c
sydidelot: Fixed. Thanks

prov/verbs/src/verbs_info.c
sydidelot: Fixed. Thanks

sydidelot:
@j-xiong Thanks for catching this stupid bug! I fixed the PR a few days ago and it now passes the CI pipeline. 
@j-xiong @shefty Can I request a new review from you? Also, should we add a few tests in `fabtests/unit/getinfo_test.c` for the new `FI_RESCAN` flag to getinfo? Thank you!

j-xiong (dismissed):

j-xiong:
@sydidelot Adding tests for the new flag is a good idea. The question though, is how to create a scenario that the flag can show a difference.

j-xiong:
@shefty Do you have any more concerns with this PR?

prov/rxd/src/rxd_init.c
```diff
@@ -147,6 +153,10 @@ RXD_INI
 			"Maximum number of peers to track (default: 1024)");
 	fi_param_define(&rxd_prov, "max_unacked", FI_PARAM_INT,
 			"Maximum number of packets to send at once (default: 128)");
+	fi_param_define(&rxd_prov, "rescan", FI_PARAM_BOOL,
+			"Enables dynamic network interface detection by rescanning "
+			"system interfaces on each fi_getinfo() invocation. "
+			"(default: -1)");
```
shefty, line 159: I'd modify the text to something like: "Force or disable rescanning for network interface changes... (default: unset)".  Replace -1 with unset, since -1 doesn't make much sense to the user reading this as a boolean value.  Or simply drop the default text completely.  I would make the text clear that setting this to true will force rescanning, while setting it to false will disable rescanning.

shefty (changes requested):
Only change request is to clean up the text describing the env var.  Thanks!

prov/rxd/src/rxd_init.c
sydidelot: Thanks @shefty! Here is how the text describing the env var now looks like:
```
fi_param_define(&rxd_prov, "rescan", FI_PARAM_BOOL,
                "Force or disable rescanning for network interface changes. "
                "Setting this to true will force rescanning on each fi_getinfo() invocation; "
                "setting it to false will disable rescanning. (default: unset)");
```

shefty: thanks -- this was my last concern

j-xiong (approved):

j-xiong:
@shefty Could you update your review status?

shefty (approved):

shefty:
done

PR #10891, opened 2025-03-20
fabtests/efa: Do not set multinode marker for test_efa_shm_addr

sunkuamzn:
The test only uses 2 nodes, so it's not a multinode test


jiaxiyan (approved):

sunkuamzn:
bot:aws:retest

PR #10892, opened 2025-03-21
prov/efa: Add efa_mr_map_insert

jiaxiyan:
ofi_mr_map_insert ignores the mr if remote access permissions are
not asked for, but EFA provider needs to add all memory regions
to the util MR map because read based message protocols might be
used for local operations, where application does not set
FI_REMOTE_READ in mr_attr->access.
We also need it for cuda memory so that we can recover from the
FI_ENOKEY error after cudaFree.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -557,6 +557,43 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
 			mr_attr->mr_iov->iov_len, access);
 }
 
+/* 
+ * efa_mr_map_insert inserts all buffers into the rb tree.
+ * ofi_mr_map_insert only inserts buffers with FI_REMOTE_READ 
+ * and FI_REMOTE_WRITE access flags.
+ */
+static int efa_mr_map_insert(struct ofi_mr_map *map,
+			     const struct fi_mr_attr *attr, uint64_t *key,
+			     void *context, uint64_t flags)
+{
+	struct fi_mr_attr *item;
+	int ret;
+
+	item = ofi_mr_dup_attr(attr, flags);
+	if (!item)
+		return -FI_ENOMEM;
+
+	if (!(map->mode & FI_MR_VIRT_ADDR))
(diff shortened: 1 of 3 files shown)


sunkuamzn (dismissed):

prov/efa/src/efa_mr.c
```diff
@@ -589,6 +589,66 @@ int efa_mr_is_cuda_memory_freed(struct efa_mr *efa_mr, bool *freed)
...
+		ofi_mr_get_iov_from_dmabuf((struct iovec *)dup_attr->mr_iov,
+			attr->dmabuf, attr->iov_count);
+	else
+		memcpy((void *) dup_attr->mr_iov, attr->mr_iov,
+			sizeof(*attr->mr_iov) * attr->iov_count);
+
+	return dup_attr;
+}
+
+static int efa_mr_map_insert(struct ofi_mr_map *map,
```
sunkuamzn, line 620: Could you add a comment on the difference between `efa_mr_map_insert` and `ofi_mr_map_insert`?

sunkuamzn (dismissed):

j-xiong:
You could simply add `FI_REMOTE_READ` to `flags` when calling `ofi_mr_map_insert`.

sunkuamzn:
@j-xiong but that would be overriding the access flags that the user has set. If we add `FI_REMOTE_READ`, then an `fi_read` call to a buffer that was registered with `FI_SEND` access flag would succeed.

prov/efa/src/efa_mr.c
```diff
@@ -557,6 +557,71 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
 			mr_attr->mr_iov->iov_len, access);
 }
 
+static struct fi_mr_attr *
+dup_mr_attr(const struct fi_mr_attr *attr, uint64_t flags)
```
shefty, line 561: Please make the current implementation available for re-use, rather than copying it here.  When exposing it, add the ofi prefix.

sunkuamzn:
@j-xiong yes, that would also work. EFA provider doesn't use `ofi_mr_regattr`

j-xiong:
No longer needed with updated #10884

PR #10900, opened 2025-03-24
prov/tcp: Coverity Issues

alex-mckinley:
Fix Coverity 454432 (unchecked return value) & 457970 (resource leak)

### DIFF
--- a/prov/tcp/src/xnet_cq.c
+++ b/prov/tcp/src/xnet_cq.c
@@ -129,6 +129,7 @@ void xnet_report_success(struct xnet_xfer_entry *xfer_entry)
 	struct util_cq *cq;
 	uint64_t flags, data, tag;
 	size_t len;
+	int ret;
 
 	if (xfer_entry->ctrl_flags & (XNET_INTERNAL_XFER | XNET_SAVED_XFER))
 		return;
@@ -172,14 +173,18 @@ void xnet_report_success(struct xnet_xfer_entry *xfer_entry)
 	}
 
 	if (cq->src) {
-		ofi_cq_write_src(cq, xfer_entry->context, flags, len,
-				 xfer_entry->user_buf, data, tag,
-				 xfer_entry->src_addr);
+		ret = ofi_cq_write_src(cq, xfer_entry->context, flags, len,
+				       xfer_entry->user_buf, data, tag,
+				       xfer_entry->src_addr);
 	} else {
-		ofi_cq_write(cq, xfer_entry->context, flags, len,
-			     xfer_entry->user_buf, data, tag);
+		ret = ofi_cq_write(cq, xfer_entry->context, flags, len,
+				   xfer_entry->user_buf, data, tag);
 	}
-	if (cq->wait)
+
+	if (ret) 
+		FI_WARN(&xnet_prov, FI_LOG_CQ, "cq write failed (%s)\n",
+			fi_strerror(ret));
+	else if (cq->wait)
 		cq->wait->signal(cq->wait);
 }
 


prov/tcp/src/xnet_domain.c
```diff
@@ -190,6 +190,17 @@ xnet_mplex_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+	/*
+	 * registering onto subdomains is only important for rma and
+	 * we rely on the fid being in the mr map for all close paths, which
+	 * will only be triggered on the rma case.
+	 * xnet does not require any special mr modes, so local access to
+	 * mr fids is not needed. If xnet is expanded to require local
+	 * registration (including FI_MR_HMEM), this must be changed. 
+	 */
+	if (!(attr->access & (FI_REMOTE_READ | FI_REMOTE_WRITE)))
+		return ret;
```
sydidelot, line 202: I think `return 0` would be better for code clarity.

j-xiong: I agree.

j-xiong (commented):
Generally looks fine to me. Please make change according to the existing comment, and add more descriptive text to the end of the title line of the first commit, e.g. `(unchecked return value)`.

prov/tcp/src/xnet_cq.c
```diff
@@ -172,14 +173,18 @@ void xnet_report_success(struct xnet_xfer_entry *xfer_entry)
...
-		ofi_cq_write(cq, xfer_entry->context, flags, len,
-			     xfer_entry->user_buf, data, tag);
+		ret = ofi_cq_write(cq, xfer_entry->context, flags, len,
+				   xfer_entry->user_buf, data, tag);
 	}
-	if (cq->wait)
+
+	if (ret) 
+		FI_WARN(&xnet_prov, FI_LOG_CQ, "cq write failed (%s)\n",
+			fi_strerror(ret));
```
shefty, line 186: Honestly, you could probably just cast away the return values from ofi_cq_write.  The app is the living dead at this point.  Please update the commit message to remove the reference to some coverity id and state the actual problem.  No one reading git logs is going to know what coverity 454432 means.

Are you even sure this is the correct fix for that issue?  "Value returned from a function is not checked for errors before being used"  No values being returned from ofi_cq_write were being used anyway.  Is this referring to the len value returned from xnet_msg_len()?

prov/tcp/src/xnet_domain.c
shefty: What is the actual memory leak here, and how does this fix it?

xnet_mr_regattr() returns a fid_mr (sub_mr_fid), which is discarded.  Ignore RMA, how does the memory referenced by sub_mr_fid get freed?

aingerson: You're right, the mr fid isn't getting freed. We're relying on the rbmap remove (by RMA key) to find the node and free it but this is only freeing the mr_attr and not the ofi_mr itself. A more correct (but still a bit hacky) solution would be to bypass the registration call entirely and only insert into the map so therefore the insert and remove are happening in parallel instead of ofi_mr_reg, mr_map_remove. Something like
```
static int xnet_mr_map_insert(struct fid *fid, const struct fi_mr_attr *attr)
{
	struct util_domain *domain;
	struct fi_mr_attr cur_abi_attr;
	uint64_t key;
	int ret = 0;

	domain = container_of(fid, struct util_domain, domain_fid.fid);

	ofi_mr_update_attr(domain->fabric->fabric_fid.api_version,
			   domain->info_domain_caps, attr, &cur_abi_attr, 0);

	ofi_genlock_lock(&domain->lock);

	ret = ofi_mr_map_insert(&domain->mr_map, &cur_abi_attr, &key, NULL, 0);

	ofi_genlock_unlock(&domain->lock);
	return ret;
}

static int
xnet_mplex_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
		uint64_t flags, struct fid_mr **mr_fid)
{
	struct xnet_domain *domain;
	struct fid_list_entry *item;
	struct fid_mr *sub_mr_fid;
	struct ofi_mr *mr;
	int ret;

	domain = container_of(fid, struct xnet_domain,
			      util_domain.domain_fid.fid);
	ret = ofi_mr_regattr(fid, attr, flags, mr_fid);
	if (ret)
		return ret;

	mr = container_of(*mr_fid, struct ofi_mr, mr_fid.fid);
	mr->mr_fid.fid.ops = &xnet_mplex_mr_fi_ops;

	if (!(attr->access & (FI_REMOTE_READ | FI_REMOTE_WRITE)))
		return ret;

	ofi_genlock_lock(&domain->subdomain_list_lock);
	dlist_foreach_container(&domain->subdomain_list,
				struct fid_list_entry, item, entry) {
		ret = xnet_mr_map_insert(item->fid, attr);
		if (ret) {
			FI_WARN(&xnet_prov, FI_LOG_MR,
				"Failed to reg mr (%ld) from subdomain (%p)\n",
				mr->key, item->fid);

			xnet_subdomains_mr_close(domain, mr->key);
			(void) ofi_mr_close(&(*mr_fid)->fid);
			break;
		}
	}

	ofi_genlock_unlock(&domain->subdomain_list_lock);
	return ret;
}
```

prov/tcp/src/xnet_cq.c
alex-mckinley: It's probably a non-issue overall, but coverity flags it because it's checked every other place that that function is called. We discussed squashing the issue, but ultimately decided that the warn logging was fine.

prov/tcp/src/xnet_domain.c
alex-mckinley: When the map insert was changed to be conditional (https://github.com/ofiwg/libfabric/commit/dbec7405f17249f80bf1c35b6ae0be61bfd86f3b), the memory referenced by sub_mr_fid doesn't always make it into the map, but is always allocated. Tcp cleans up by iterating through its map of registered memory. sub_mr_fid ends up allocated but not registered and is not found by the provider to be cleaned up during shut down

prov/tcp/src/xnet_cq.c
shefty: I'm focused on the wording "before being used" in "Value returned from a function is not checked for errors before being used".  Either the message is wrong or there's some other problem being referenced.  If the message stopped at 'errors', this change makes more sense.

Adding ```(void) ofi_cq_write(...)``` cast would indicate to coverity that the return value is intentionally being discarded.  I honestly don't care if the warning is kept instead.  My concern is whether the correct problem is being addressed.

prov/tcp/src/xnet_domain.c
```diff
@@ -190,6 +190,17 @@ xnet_mplex_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+	/*
+	 * registering onto subdomains is only important for rma and
+	 * we rely on the fid being in the mr map for all close paths, which
+	 * will only be triggered on the rma case.
+	 * xnet does not require any special mr modes, so local access to
+	 * mr fids is not needed. If xnet is expanded to require local
+	 * registration (including FI_MR_HMEM), this must be changed. 
+	 */
+	if (!(attr->access & (FI_REMOTE_READ | FI_REMOTE_WRITE)))
+		return 0;
```
shefty, line 202: I still don't like this fix, as it is masking the problem with a cryptic and fragile patch.  No one reading this code is going to connect this if statement with the memory leak below that still looks like a memory leak.

alex-mckinley:
@shefty With regards to the unchecked return value, the full coverity log is 
> Calling ofi_cq_write_src without checking return value (as is done elsewhere 6 out of 7 times).

I'm happy to change the patch to a discard. I think the static analysis just wants us to be intentional with what we do about the return. Whether that's logging a warning or just discarding, I'm not sure that matters.

shefty:
The warning is fine, just update the commit message to reflect the wording in the coverity message.  That makes it clear that the fix is correct.

j-xiong:
Please rebase to pick up github action changes.

prov/tcp/src/xnet_domain.c
```diff
@@ -190,10 +208,13 @@ xnet_mplex_mr_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	mr = container_of(*mr_fid, struct ofi_mr, mr_fid.fid);
 	mr->mr_fid.fid.ops = &xnet_mplex_mr_fi_ops;
 
+	if (!(attr->access & (FI_REMOTE_READ | FI_REMOTE_WRITE)))
+		return ret;
```
aingerson: Please update to address the original comment in this thread (return 0)

PR #10907, opened 2025-03-26
prov/shm: new shm architecture v2

aingerson:


sunkuamzn:
@aingerson the unexpected message fabtest is still timing out

```
server_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.84.23 'timeout 1800 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10907-debug/install/fabtests/bin/fi_unexpected_msg -e rdm -M 2048 -I 5 -v -S 512 -p shm -E=9234'"'"''

client_command: ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=30 -o BatchMode=yes 172.31.84.23 'timeout 1800 /bin/bash --login -c '"'"'FI_LOG_LEVEL=warn /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr10907-debug/install/fabtests/bin/fi_unexpected_msg -e rdm -M 2048 -I 5 -v -S 512 -p shm -E=9234 172.31.84.23'"'"''
client_stdout:

client returncode: 124
server_stdout:

server returncode: 124
```

aingerson:
@sunkuamzn Thanks! Is that the only failure? That error is expected, see my comment in #10817

shijin-aws:
The timing out is because the new shm removed the buffering for inject protocol, so it cannot handle unexpected messages of number larger than the tx size. I need to discuss with my team to decide whether it is acceptable for our customers.

aingerson:
@shijin-aws Yeah, it would be good to figure out why OMPI/your customers are having this issue. Sending over tx_size unexpected messages is a misuse of the API. Other providers don't support it as well and supporting it here would have a performance impact for other use cases so I'm hesitant about adding that buffering. There's likely a weird use of the API in OMPI that we need to investigate. If we can have setting FI_SHM_TX_SIZE as a workaround for any apps that see this issue until we can dig more into OMPI to root cause, I think that would be better.  Let me know what you decide with your team (and feel free to message me directly to discuss)

shijin-aws:
[main_new_shm_compare_c7gn_u2204.txt](https://github.com/user-attachments/files/20179232/main_new_shm_compare_c7gn_u2204.txt)
[main_new_shm_compare_hpc6a_u2204.txt](https://github.com/user-attachments/files/20179233/main_new_shm_compare_hpc6a_u2204.txt)
[main_new_shm_compare_hpc6id_u2204.txt](https://github.com/user-attachments/files/20179234/main_new_shm_compare_hpc6id_u2204.txt)

Ran fabtests with shm provider against main branch and PR 10693 on 3 CPU types: hpc6a (AMD), hpc6id (intel), c7gn (graviton) on ubuntu 2204. The comparisons are recorded in the attachments.

In the inline data size protocol ( <=256 bytes)
For AMD there is a systematic 20ns regression for the new shm on ping pong (tagged, non-tagged, rma)
For Graviton, the regression is almost negligible.
For Intel, the regression is worse as 50 ns.

Alexia said there should be improvement on the inject size (256 to 4K) bandwidth. I do see such improvement (20% - 50%) for all message sizes from 256 Bytes to 4KB on hpc6id and c7gn. But for hpc6a the bandwidth get worse for new shm between 1536 bytes to 4096 bytes.

PR #10912, opened 2025-03-31
av: introduce FI_FIREWALL_ADDR flag for insert operations

jxiong:
FI_FIREWALL_ADDR indicates to the provider that the inserted addresses may be restricted by a firewall.


shefty (approved):
looks reasonable to me

j-xiong (approved):

PR #10913, opened 2025-03-31
prov/verbs: Enable logging of ibv_async_events

### LINKED ISSUE #10852: prov/verbs: add logging of ibv_async_events
No specific issue, but it can be very helpful to debug certain issues when we're able to see specific ibv events as they happen.

The verbs provider already tracks the ibv_context which we can use to monitor the async_fd for async events, I'm looking at how to add this to the relevant progress paths for both active and passive ep.

ooststep:
This change enhances the verbs EQ to poll for ibv async events, which
notify the user of device state changes and errors.

As this is primarily focused on enhancing the debugging experience,
for now, events are only logged as they are read during EQ progress
and no action is taken.

(edit: PR summary updated to match commit message)

implements #10852

### DIFF
--- a/prov/verbs/src/verbs_domain.c
+++ b/prov/verbs/src/verbs_domain.c
@@ -159,8 +159,8 @@ static int vrb_domain_bind(struct fid *fid, struct fid *bfid, uint64_t flags)
 		switch (domain->ep_type) {
 		case FI_EP_MSG:
 			eq = container_of(bfid, struct vrb_eq, eq_fid);
-			domain->eq = eq;
 			domain->eq_flags = flags;
+			vrb_eq_attach_domain(eq, domain);
 			break;
 		case FI_EP_DGRAM:
 			return -FI_EINVAL;
@@ -222,6 +222,12 @@ static int vrb_domain_close(fid_t fid)
 		if (ret)
 			return -ret;
 		domain->pd = NULL;
+	}
+
+	if (domain->eq) {
+		ret = vrb_eq_detach_domain(domain);
+		if (ret)
+			return -ret;
 	}
 
 	ret = ofi_domain_close(&domain->util_domain);
@@ -430,6 +436,10 @@ vrb_domain(struct fid_fabric *fabric, struct fi_info *info,
 		ret = -FI_EINVAL;
 		goto err4;
 	}
+
+	ret = fi_fd_nonblock(_domain->verbs->async_fd);
+	if (ret)
+		goto err4;
 
 	ret = vrb_init_progress(&_domain->progress, _domain->info);
 	if (ret)
--- a/prov/verbs/src/verbs_eq.c
+++ b/prov/verbs/src/verbs_eq.c
@@ -1292,9 +1292,32 @@ out:
 	return ret;
 }
 
+static void vrb_eq_process_async_events(struct vrb_eq *eq)
+{
+	int ret;
+	struct vrb_domain *domain;
+	struct ibv_async_event async_event;
+
+	ofi_mutex_lock(&eq->fab->util_fabric.lock);
+	dlist_foreach_container(&eq->fab->util_fabric.domain_list,
+				struct vrb_domain, domain,
+				util_domain.list_entry) {
+		do {
+			ret = ibv_get_async_event(domain->verbs, &async_event);
+			if (!ret) {
+				VRB_WARN(FI_LOG_DOMAIN, "Async event for %s: %s\n",
+					 eq->fab->info->domain_attr->name,
+					 ibv_event_type_str(async_event.event_type));
+				ibv_ack_async_event(&async_event);
+			}
+		} while (!ret);
+	}
+	ofi_mutex_unlock(&eq->fab->util_fabric.lock);
+}
+
 static ssize_t
 vrb_eq_read(struct fid_eq *eq_fid, uint32_t *event,
-	       void *buf, size_t len, uint64_t flags)
+	    void *buf, size_t len, uint64_t flags)
 {
 	struct vrb_eq *eq;
 	struct rdma_cm_event *cma_event;
@@ -1320,7 +1343,8 @@ vrb_eq_read(struct fid_eq *eq_fid, uint32_t *event,
 		vrb_prof_func_end("rdma_get_cm_event");
 		if (ret) {
 			ofi_mutex_unlock(&eq->event_lock);
-			return -errno;
+			ret = -errno;
+			goto out;
 		}
 		vrb_prof_func_start("vrb_eq_cm_process_event");
 		ret = vrb_eq_cm_process_event(eq, cma_event, event, buf, len);
@@ -1331,7 +1355,9 @@ vrb_eq_read(struct fid_eq *eq_fid, uint32_t *event,
 
 	if (ret > 0 && flags & FI_PEEK)
 		ret = vrb_eq_write_event(eq, *event, buf, ret);
-
+out:
+	if (ret <= 0)
+		vrb_eq_process_async_events(eq);
 	vrb_prof_func_end(__func__);
 	return ret;
 }
@@ -1425,7 +1451,9 @@ static int vrb_eq_close(fid_t fid)
 	struct vrb_eq_entry *entry;
 
 	eq = container_of(fid, struct vrb_eq, eq_fid.fid);
-	/* TODO: use util code, if possible, and add ref counting */
+	/* TODO: use util code, if possible */
+	if (ofi_atomic_get32(&eq->ref))
+		return -FI_EBUSY;
 
 	if (!ofi_rbmap_empty(&eq->xrc.sidr_conn_rbmap))
 		VRB_WARN(FI_LOG_EP_CTRL, "SIDR connection RBmap not empty\n");
@@ -1538,6 +1566,8 @@ int vrb_eq_open(struct fid_fabric *fabric, struct fi_eq_attr *attr,
 	_eq->eq_fid.fid.ops = &vrb_eq_fi_ops;
 	_eq->eq_fid.ops = &vrb_eq_ops;
 
+	ofi_atomic_initialize32(&_eq->ref, 0);
+
 	*eq = &_eq->eq_fid;
 	return 0;
 err4:
@@ -1556,3 +1586,23 @@ err0:
 	return ret;
 }
 
+int vrb_eq_attach_domain(struct vrb_eq *eq, struct vrb_domain *domain)
+{
+	if (ofi_epoll_add(eq->epollfd, domain->verbs->async_fd,
+			  OFI_EPOLL_IN, domain))
+		return -errno;
+
+	domain->eq = eq;
+	ofi_atomic_inc32(&eq->ref);
+	return 0;
+}
+
+int vrb_eq_detach_domain(struct vrb_domain *domain)
+{
+	if (ofi_epoll_del(domain->eq->epollfd, domain->verbs->async_fd))
+		return -errno;
+
+	ofi_atomic_dec32(&domain->eq->ref);
+	domain->eq = NULL;
+	return 0;
+}
--- a/prov/verbs/src/verbs_ofi.h
+++ b/prov/verbs/src/verbs_ofi.h
@@ -329,6 +329,8 @@ struct vrb_eq {
 
 	ofi_epoll_t		epollfd;
 	enum fi_wait_obj	wait_obj;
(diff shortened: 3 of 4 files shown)


prov/verbs/src/verbs_cq.c
```diff
@@ -479,9 +479,20 @@ static struct fi_ops vrb_cq_fi_ops = {
...
 	cq = container_of(util_cq, struct vrb_cq, util_cq);
 	vrb_flush_cq(cq);
+
+	domain = container_of(util_cq->domain, struct vrb_domain, util_domain);
+	if (!domain->epoll_fd)
+		return;
+
+	nfds = ofi_epoll_wait(domain->epoll_fd, &event, 1, 0);
+	if (nfds > 0)
+		vrb_domain_process_async_events(domain, nfds);
```
shefty, line 495: This adds a kernel wait call into the cq progress path.  It is very unclear why this is needed for logging.

prov/verbs/src/verbs_domain.c
```diff
@@ -355,6 +362,34 @@ vrb_domain(struct fid_fabric *fabric, struct fi_info *info,
...
+		goto skip_async;
+
+	ret = fi_fd_nonblock(_domain->verbs->async_fd);
+	if(ret) {
+		VRB_INFO(FI_LOG_DOMAIN, "Failed to enable async events for %s: %s\n",
+				_domain->info->domain_attr->name, strerror(-errno));
+		goto skip_async;
+	}
+
+	ret = ofi_epoll_create(&_domain->epoll_fd);
```
shefty, line 375: Why is logging adding epoll overhead?

shefty (commented):
I don't get what this patch is trying to do, or why logging needs to introduce extra fd's.

prov/verbs/src/verbs_cq.c
ooststep: we need to poll the async_fd to know that an event is ready to retrieve, otherwise `ibv_get_async_event` is synchronous and may block until an event becomes available.

prov/verbs/src/verbs_domain.c
ooststep: I can switch to poll for getting these events on the cq, which will remove the epoll structures, but a poll will still be required. The only existing poll is in the cq_sread path and I don't think that's enough.

prov/verbs/src/verbs_cq.c
shefty: I still don't get it.  If the intent is to log when an async event is retrieved, why isn't that just done where the existing code reads an async event?  These changes actively poll for async events in different places than the previous code, so there's a lot more going on here than simply trying to log an event.

ooststep: Unless I'm severely mistaken, the existing code is polling for cq events, not async events, and only during cq_sread (talking specifically about polling an fd with poll, not considering ibv_poll_cq in the cq_read path).  I did not think adding async events to that path would be sufficient as it's only happening during cq_sread.  If you believe that would be sufficient it's simple enough to move the polling there.

shefty: I don't understand the intent.  If we just want to log async events, why not simply add the log where the code reads them?  This change is introducing wait support and async events in places where the previous code would never have generated an async event.  Why are you trying to *add* async events?  Why is *logging* trying to change how or where the code waits for events and where it polls for them?  Enabling this logging is significantly changing the behavior beyond simply writing output.

ooststep: I'm not *adding* async events. I'm *adding* **retrieving and logging** of async events generated by the kernel.  These events are not captured in today's implementation. They are retrievable only with synchronous APIs that will block until events are available, and not done today. The alternative to synchronously getting the events is polling the async_fd to know there are events waiting, making the synchronous calls return immediately.

Understanding that this could impact performance and as it is primarily focused on aiding debugging, it's optional.  This poll will only happen when requested by the added environment variable.

ooststep: I can update the commit message to make the intent more clear and explain why we need to add a poll but I don't see a way to add support for this feature without adding a poll to the progress path as that is the path taken with fi_cq_read.

j-xiong:
A few functions are missing from the windows build:  
```
verbs_domain.obj : error LNK2019: unresolved external symbol ibv_get_async_event
verbs_domain.obj : error LNK2019: unresolved external symbol ibv_ack_async_event 
verbs_domain.obj : error LNK2019: unresolved external symbol ibv_event_type_str 
```

prov/verbs/src/verbs_cq.c
```diff
@@ -137,16 +137,26 @@ vrb_cq_readerr(struct fid_cq *cq_fid, struct fi_cq_err_entry *entry,
...
 vrb_poll_events(struct vrb_cq *_cq, int timeout)
 {
-	int ret, rc;
+	int ret, rc, nfds;
 	void *context;
-	struct pollfd fds[2];
+	struct pollfd fds[3];
+	struct vrb_domain *domain = NULL;
 
+	nfds=2;
```
aingerson, line 145: space around operators

prov/verbs/src/verbs_cq.c
```diff
@@ -137,16 +137,26 @@ vrb_cq_readerr(struct fid_cq *cq_fid, struct fi_cq_err_entry *entry,
...
 
 	fds[0].events = fds[1].events = POLLIN;
 
-	rc = poll(fds, 2, timeout);
+	if (vrb_gl_data.log_async_events) {
+		domain = container_of(_cq->util_cq.domain, struct vrb_domain,
+				      util_domain);
+		fds[2].fd = domain->verbs->async_fd;
+		fds[2].events = POLLIN;
+		nfds+=1;
```
aingerson, line 156: same here

prov/verbs/src/verbs_domain.c
```diff
@@ -309,6 +319,18 @@ static struct fi_ops_domain vrb_dgram_domain_ops = {
 };
 
 
+static void vrb_domain_enable_async_events(struct vrb_domain *domain)
+{
+	assert(domain->verbs);
+	int ret = fi_fd_nonblock(domain->verbs->async_fd);
```
aingerson, line 325: I would change this to be
```
int ret;

assert(domain->verbs);
ret = fi_fd_nonblock(domain->verbs->async_fd);
```
which matches the declaration at the top, new line, then code format

prov/verbs/src/verbs_domain.c
```diff
@@ -309,6 +319,18 @@ static struct fi_ops_domain vrb_dgram_domain_ops = {
...
+	if(ret) {
```
aingerson, line 326: space after if

aingerson (commented):
A few nitpicky things but otherwise looks good!

shefty:
ibv_async_events are device level events.  Those should be read as part of EQ handling.

ibv_get_cq_event() are related to events generated in response to blocking CQ calls.

Don't process async events from the CQ calls.  If you want to check for async events, hook into the EQ path and read them when checking for CM events.  Only setup blocking fd support if the EQ was configured in blocking mode.  Otherwise, just call ibv_get_async_event() from the EQ read call.

ooststep:
> ibv_async_events are device level events. Those should be read as part of EQ handling.
> 
> ibv_get_cq_event() are related to events generated in response to blocking CQ calls.
> 
> Don't process async events from the CQ calls. If you want to check for async events, hook into the EQ path and read them when checking for CM events. Only setup blocking fd support if the EQ was configured in blocking mode. Otherwise, just call ibv_get_async_event() from the EQ read call.

I agree that it makes the most sense for these async device events to be polled solely by the EQ. However, existing apps that may not be modifiable would not be able to leverage the feature. For example, from what I can tell many MPI implementations do not appear to read the OFI EQ. It would be fairly difficult for MPI based applications to be modified to utilize the feature.

shefty:
Connection establishment should already be part of any flow.

ooststep:
> Connection establishment should already be part of any flow.

You're correct, I was mistaken, and I'll fix up my PR to align with that.

While MPI doesn't directly utilize the EQ since it's using RDM endpoints, RxM handles connection management and will poll the verb's EQ.

ooststep:
bot:aws:retest

ooststep:
@shefty I updated to align with your suggestions, though we still need a poll to get the related ep's fid to access the ibv_context.  Any additional feedback/suggestion?

prov/verbs/src/verbs_ep.c
```diff
@@ -1441,6 +1441,23 @@ static int vrb_pep_bind(fid_t fid, struct fid *bfid, uint64_t flags)
...
+				    OFI_EPOLL_IN, &pep->pep_fid.fid);
+		if (ret) {
+async_err:
+			VRB_WARN(FI_LOG_DOMAIN, "Failed to enable async events for %s: %s\n",
+				 pep->info->domain_attr->name, strerror(ret));
+			pep->log_async_events = false;
+		} else {
+			pep->log_async_events = true;
+		}
+	}
```
shefty, line 1459: If the app wants blocking support, just add the async_fd to the epollfd, so that waiting on the EQ will be triggered if any async event occurs.  Logging should just always be enabled if blocking support is desired.  If the EQ doesn't have blocking support, there's nothing to do here.  See below.

prov/verbs/src/verbs_eq.c
```diff
@@ -1346,7 +1388,7 @@ vrb_eq_sread(struct fid_eq *eq_fid, uint32_t *event,
 	eq = container_of(eq_fid, struct vrb_eq, eq_fid.fid);
 
 	while (1) {
-		ret = vrb_eq_read(eq_fid, event, buf, len, flags);
+		ret = vrb_eq_read_internal(eq_fid, timeout, event, buf, len, flags);
```
shefty, line 1391: The timeout should not be passed through.  Waiting will occur on the eq->epollfd, which has the async_fd attached.  If an async event occurs, the epollfd will be signaled.

shefty (commented):
I think it can be simplified.  See comments.

Measure performance with and without checking for async events and see if there's a noticeable difference.  If not, this can be always enabled.

ooststep:
I measured various mpi based workloads and synthetic benchmarks and did not see meaningful change in performance, though there were no events being sent during any of the performance focused testing.

shefty (commented):
Minor comments.  I wouldn't expect events to occur in most situations, so it seems like it's possible to always have this feature available.

ooststep:
appveyor error seems to be unrelated and possible a system issue. failed tests are TCP (untouched in thie commit) and looks to be due lack of hmem resources

ooststep:
Intel CI looks to be an issue with a specific set of nodes used for testing and is now passing.

@shefty @j-xiong any additional feedback?

j-xiong (approved):

prov/verbs/src/verbs_eq.c
```diff
@@ -1292,9 +1292,32 @@ static size_t vrb_eq_read_event(struct vrb_eq *eq, uint32_t *event,
...
+	ofi_mutex_lock(&eq->fab->util_fabric.lock);
+	dlist_foreach_container(&eq->fab->util_fabric.domain_list,
+				struct vrb_domain, domain,
+				util_domain.list_entry) {
+		do {
+			ret = ibv_get_async_event(domain->verbs, &async_event);
+			if (!ret) {
+				VRB_WARN(FI_LOG_DOMAIN, "Async event for %s: %s\n",
+					 eq->fab->info->domain_attr->name,
+					 ibv_event_type_str(async_event.event_type));
```
sydidelot, line 1310: I don’t think all events should be treated as "warnings." For example, establishing a connection triggers the `IBV_EVENT_COMM_EST` event, which could quickly flood the logs for large scale applications or services that rely on many short-lived connections.
Why don't we just post the event to the EQ and let the application handle it?

shefty: The events aren't defined by the libfabric API, so they shouldn't be pass through.  Existing apps aren't prepared to handle them.

I don't think the verbs provider will trigger IBV_EVENT_COMM_EST.  The rdma cm flow doesn't remain the RTR state (which is when COMM_EST is generated).  It moves to RTS prior to sending a CM REP or RTU.

sydidelot:
Hi, I've identified several issues introduced by this PR. Could you please take a look at https://github.com/ofiwg/libfabric/issues/12181? Thanks!

PR #10921, opened 2025-04-01
prov/efa: Require FI_MR_ALLOCATED in domain_attr

jiaxiyan:
efa provider requires that memory registration occurs on allocated data buffers, and physical pages must back all virtual addresses being registered.


j-xiong:
@jiaxiyan Could you rebase to pick up github action changes?

sunkuamzn (approved):

Issue #10918, opened 2025-04-01
.github/workflows RDMA core version failure


zachdworkin:
pr-ci.yml fails with v34.1 of rdma-core because cmake version is not supported. Should we update the rdma-core version or add the -DCMAKE_POLICY_VERSION_MINIMUM=3.5 flag the script is asking for?

Output:
+ git clone --depth 1 -b v34.1 https://github.com/linux-rdma/rdma-core.git
Cloning into 'rdma-core'...
warning: refs/tags/v34.1 84fedbc92bece64d62462e5937b1532e3a6f893e is not a commit!
Note: switching to '0974fbb2ec58536afb0[18](https://github.com/ofiwg/libfabric/actions/runs/14205992981/job/39804247893?pr=10883#step:4:19)6426e33886d18ebcca9'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

+ pushd rdma-core
~/work/libfabric/libfabric/rdma-core ~/work/libfabric/libfabric
+ bash build.sh
CMake Error at CMakeLists.txt:56 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

-- Configuring incomplete, errors occurred!

zachdworkin:
@j-xiong @sunkuamzn

j-xiong:
Let's try a newer rdma-core first. The latest one is v56.

sunkuamzn:
Upgrading rdma-core is fine with us

zachdworkin:
Upgrading to v56.0 has abi issues. 

We can use a lower one without the issues,
Fix them or,
add this build arg -Wgnu-variable-sized-type-not-at-end to suppress them.

At first glance fixing them would require a large change to usnic code because of variable struct sizes within the structs with the error.

ooststep:
[...]

PR #10922, opened 2025-04-01
prov/tcp,rxm: add firewall support for RDM

jxiong:
This PR allows the server and client to exchange feature flags, and firewall is introduced as the first feature in the features.

In order to mark a node is behind firewall, the node will set an environmental variable `FI_TCP_FIREWALL_ADDR`, and in that case, peer should never try to initiate a connection back to this node.

### DIFF
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -928,9 +928,12 @@ struct util_peer_addr {
 	int index;
 	int refcnt;
 	union ofi_sock_ip addr;
+	char str_addr[OFI_ADDRSTRLEN];
+	bool firewall_addr;
 };
 
-struct util_peer_addr *util_get_peer(struct rxm_av *av, const void *addr);
+struct util_peer_addr *util_get_peer(struct rxm_av *av, const void *addr,
+				     uint64_t flags);
 void util_put_peer(struct util_peer_addr *peer);
 
 /* All peer addresses, whether they've been inserted into the AV
--- a/prov/util/src/rxm_av.c
+++ b/prov/util/src/rxm_av.c
@@ -33,7 +33,6 @@
 #include "ofi_util.h"
 #include "uthash.h"
 
-
 size_t rxm_av_max_peers(struct rxm_av *av)
 {
 	size_t cnt;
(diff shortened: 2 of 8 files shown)


jxiong:
@j-xiong I saw this error from the build, and it doesn't seem to be relevant to my PR:
```
+ bash build.sh
CMake Error at CMakeLists.txt:56 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```
Can you take a look?

j-xiong:
see #10918

j-xiong:
@jxiong Could you rebase to pick up github action changes?

include/ofi_util.h
```diff
@@ -928,9 +928,12 @@ struct util_peer_addr {
 	int index;
 	int refcnt;
 	union ofi_sock_ip addr;
+	char str_addr[OFI_ADDRSTRLEN];
+	bool firewall_addr;
 };
 
-struct util_peer_addr *util_get_peer(struct rxm_av *av, const void *addr);
+struct util_peer_addr *util_get_peer(struct rxm_av *av, const void *addr,
+		  uint64_t flags);
```
j-xiong, line 936: alignment is off.

prov/util/src/rxm_av.c
```diff
@@ -113,6 +118,10 @@ util_get_peer(struct rxm_av *av, const void *addr)
 		peer = rxm_alloc_peer(av, addr);
 	}
 
+	if (peer) {
+		peer->firewall_addr |= !!(flags & FI_FIREWALL_ADDR);
+	}
```
j-xiong, line 123: don't need the braces here.

j-xiong (commented):
In addition, since the rxm provider is also changed, please change the prefix of the title line to `prov/tcp,rxm: `

j-xiong:
Please squash the commits.

j-xiong (approved):

PR #10923, opened 2025-04-01
prov/verbs: fix atomic work request fields

aingerson:
Since verbs only supports a single element for atomic operations, the work request takes in the value to be compared/swapped/operated, not the pointer to the buffer with the values. Update the fields to dereference the pointer and submit the value itself. This also fixes the incorrect msg parameter being inputted in various places. The msg->addr was being used (fi_addr_t) instead of the msg->msg_iov->addr which has the pointer to the data buffer.

This also changes query atomic query return to be -FI_ENOSYS instead of -FI_EINVAL when querying unsupported op/datatype combinations to match with the man pages and the rest of the returns.

### DIFF
--- a/prov/verbs/src/verbs_ep.c
+++ b/prov/verbs/src/verbs_ep.c
@@ -2004,7 +2004,7 @@ check_datatype:
 #endif
 		break;
 	default:
-		return -FI_EINVAL;
+		return -FI_ENOSYS;
 	}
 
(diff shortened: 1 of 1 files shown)


j-xiong:
@aingerson Could you rebase to pick up github action changes?

j-xiong (approved):

j-xiong:
bot:aws:retest

PR #10924, opened 2025-04-02
.github/workflows: Update rdma-core version to v35.0

zachdworkin:
v34.1 cannot build anymore because cmake version is too low. Update to latest v56.0 to fix this.


j-xiong:
Please remove `latest` from the commit message since v35.0 is not the latest.

j-xiong:
No need to wait for AWS CI to complete since this is a change to github actions.

PR #10926, opened 2025-04-02
prov/tcp: handle the error correctly

jxiong:
If ofi_recv_socket() returns 0, it indicates the socket has been closed. This is not an error condition therefore it shouldn't use errno since this is just a leftover from previous error.

To make things worse, if the leftover error is EAGAIN, the caller
 xnet_req_done() will handle that error specially that will leave the
 endpoint in a weird state.

### DIFF
--- a/prov/tcp/src/xnet_cm.c
+++ b/prov/tcp/src/xnet_cm.c
@@ -63,7 +63,7 @@ xnet_recv_cm_msg(SOCKET sock, struct xnet_cm_msg *msg)
 			"Failed to read cm header, ret: %zd, sockerr: %d\n",
 			ret, ofi_sockerr());
 		msg->hdr.seg_size = 0;
-		return ofi_sockerr() ? -ofi_sockerr() : -FI_EIO;
+		return ret >= 0 ? -FI_EIO : -ofi_sockerr();
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


jxiong:
@j-xiong that's the issue I talked during the meeting. @shefty FYI

shefty (approved):
thanks!

prov/tcp/src/xnet_cm.c
```diff
@@ -63,7 +63,7 @@ xnet_recv_cm_msg(SOCKET sock, struct xnet_cm_msg *msg)
 			"Failed to read cm header, ret: %zd, sockerr: %d\n",
 			ret, ofi_sockerr());
 		msg->hdr.seg_size = 0;
-		return ofi_sockerr() ? -ofi_sockerr() : -FI_EIO;
+		return ret == 0 ? -FI_EIO : -ofi_sockerr();
```
j-xiong, line 66: `0 < ret < len` is also not an error condition.

shefty: You can change the check to: ```ret >= 0 ? -FI_EIO ...```.  CM messages are small and the first data exchanged on the socket.  The code expects it to be delivered whole (doesn't support partial delivery of the CM message).  The tcp stack allocates at least an mtu size buffer for the socket, so the chances of 0 < ret < len occurring is zero in practice.

jxiong: Yeah I would think so, but I'm convinced by the comment:
```
/* The underlying socket has the POLLIN event set.  The entire
 * CM message should be readable, as it fits within a single MTU
 * and is the first data transferred over the socket.
 */
 ```
@shefty yeah I can make that change.

j-xiong (approved):

j-xiong:
@jxiong Could you rebase to pick up the github action changes?

j-xiong:
bot:aws:retest

PR #10930, opened 2025-04-03
prov/cxi: added new optional feature to enable accelerated collectives when building libfabric

SeanP-2023:
Disabled by default: This feature was needed to prevent curl lib dlopen/dlsym failures during collectives init and runtime errors when building a multicast address in the switch. If we do not have curl symbols to support accelerated collectives, the feature will be disabled.

### DIFF
--- a/prov/cxi/configure.m4
+++ b/prov/cxi/configure.m4
@@ -44,6 +44,18 @@ AC_DEFUN([FI_CXI_CONFIGURE],[
 	AC_ARG_WITH([json-c],
 		[AS_HELP_STRING([--with-json-c=DIR], [Install directory for json-c])])
 
+	# Support for collectives dlopen/dlsym of curl libs.
+	coll_dlopen=0
+	AC_ARG_ENABLE([coll-dlopen],
+        [AS_HELP_STRING([--enable-coll-dlopen], [Enable collectives and dlopen of required curl libraries @<:@default=no@:>@])],
+        [
+               # Fail if --with_dlopen is not set
+               AS_IF([test "$with_dlopen" = "no"], [AC_MSG_ERROR([dlopen not found.  libfabric requires libdl.])])
+	       AS_IF([test "$enable_coll_dlopen" != "no"], [coll_dlopen=1])
+        ])
+        # define and set ENABLE_COLL_DLOPEN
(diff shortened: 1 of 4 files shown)


prov/cxi/src/cxip_curl.c
```diff
@@ -263,6 +264,12 @@ int cxip_curl_load_symbols(void)
 	}
 	/* record handle to prevent reloading */
 	cxip_curlhandle = h;
+#else	/* Collectives are disabled, log it */ 
+	TRACE_CURL("Accelerated collectives cannot be enabled, libcurl not supported\n");
+	CXIP_WARN("Accelerated collectives cannot be enabled, libcurl not supported\n");
+	CXIP_WARN("Accelerated collectives cannot be enabled, libcurl not supported\n");
```
j-xiong, line 270: duplicated line.

j-xiong (commented):
The `enable-xxx-dlopen` options are usually used to provide an alternative to compile time linking. Here the curl lib can only be dlopened.

I would just name the option `--enable-coll` or '--enable-coll-offload`.

prov/cxi/src/cxip_curl.c
SeanP-2023: good catch!

prov/cxi/configure.m4
```diff
@@ -44,6 +44,18 @@ AC_DEFUN([FI_CXI_CONFIGURE],[
 	AC_ARG_WITH([json-c],
 		[AS_HELP_STRING([--with-json-c=DIR], [Install directory for json-c])])
 
+	# Support for collectives dlopen/dlsym of curl libs.
+	coll_dlopen=0
+	AC_ARG_ENABLE([coll-dlopen],
+        [AS_HELP_STRING([--enable-coll-dlopen], [Enable collectives and dlopen of required curl libraries @<:@default=no@:>@])],
+        [
+               # Fail if --with_dlopen is not set
```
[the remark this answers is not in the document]
SeanP-2023: fixed

prov/cxi/src/cxip_curl.c
SeanP-2023: fixed

swelch:
please rebase

PR #10931, opened 2025-04-03
prov/cxi: added new optional feature to enable accelerated collectives when building libfabric

SeanP-2023:
Disabled by default: This feature was needed to prevent curl lib dlopen/dlsym failures during collectives init and runtime errors when building a multicast address in the switch. If we do not have curl symbols to support accelerated collectives, the feature will be disabled.


prov/cxi/configure.m4
```diff
@@ -44,6 +44,16 @@ AC_DEFUN([FI_CXI_CONFIGURE],[
 	AC_ARG_WITH([json-c],
 		[AS_HELP_STRING([--with-json-c=DIR], [Install directory for json-c])])
 
+        # Support for collectives dlopen/dlsym of curl libs.
+        coll_enable=0
+        AC_ARG_ENABLE([collectives],
+        [AS_HELP_STRING([--enable-collectives], [Enable collectives and dlopen of required curl libraries @<:@default=no@:>@])],
+        [
+               AS_IF([test "$enable_collectives" != "no"], [coll_enable=1])
```
j-xiong, line 52: You still need to check if dlopen is enabled before setting col_enable to 1.

j-xiong:
For future reference, you don't need to create a new PR to make change to an existing PR. Just make the change and do a force push.

SeanP-2023:
Thanks, I created a new branch with these changes so I thought it needed a
new PR. I'll re-add the dlopen

prov/cxi/configure.m4
SeanP-2023: done

prov/cxi/configure.m4
```diff
@@ -44,6 +44,17 @@ AC_DEFUN([FI_CXI_CONFIGURE],[
 	AC_ARG_WITH([json-c],
 		[AS_HELP_STRING([--with-json-c=DIR], [Install directory for json-c])])
 
+        # Support for collectives dlopen/dlsym of curl libs.
+        coll_enable=0
+        AC_ARG_ENABLE([collectives],
```
j-xiong, line 49: Since it's a cxi only option, renaming it to `cxi-collectives` can prevent future naming conflict.

SeanP-2023: Meaning a different provider collectives flag conflict?

SeanP-2023: I can change the name if you feel that conflict could happen.

SeanP-2023: ok will do

SeanP-2023: all set

j-xiong (approved):

PR #10936, opened 2025-04-04
[v2.1.x] contrib/intel/jenkins: Pick modify env var withEnv instead of =

zachdworkin:
Jenkins doesn't let environment variables change with the '='. Instead we must use the withEnv block to update an environment variable for specific stages.


aingerson (approved):

zachdworkin:
no need to wait for AWS CI since changes are to libfabric jenkinsfile

PR #10960, opened 2025-04-11
Support read only memory registrations

iziemba:
By default, for cached local memory registrations, the CXI provider would always set access control to read/write. This was resulting in read-only memory registration to fail.

This issue is resolved by having each libfabric RDMA operation implementation defining exactly the access control it requires. For example, fi_send() requires read and fi_recv() requires write. This information is then passed into the memory registration function. If caching is disable, the access control are passed into the kernel unmodified. If caching is enabled, the following happens during memory registration:
1. Evict any overlapping entries with access control mismatch.
2. Register memory as read/write.
3. If read/write memory registration fails, attempt read-only.

### DIFF
--- a/prov/cxi/include/cxip.h
+++ b/prov/cxi/include/cxip.h
@@ -2477,7 +2477,7 @@ struct cxip_ep_obj {
 };
 
 int cxip_ep_obj_map(struct cxip_ep_obj *ep, const void *buf, unsigned long len,
-		    uint64_t flags, struct cxip_md **md);
+		    uint64_t access, uint64_t flags, struct cxip_md **md);
 
 static inline void
 cxip_ep_obj_copy_to_md(struct cxip_ep_obj *ep, struct cxip_md *md, void *dest,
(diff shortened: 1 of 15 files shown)


swelch (approved):

roblatham00:
My MPI-ony  CXI test case passes but still gives some chatter:

```
$ mpiexec -envall  -np 2 -ppn 1 ./cxi-testcase
[1 of 2] successfuly recv Hello string type
[1 of 2] successfuly recv Hello stack
[1 of 2] successfully recv hello
[0 of 2] succesfully sent from string type
cxil_map: write error
cxil_map: write error
[0 of 2] successfully sent from stack
[0 of 2] successfully sent string literal
```

the message shows up just once for either the "sent from stack" or "sent string literal" case (I commented out different tests to see what would happen).

roblatham00:
Oh... that comes from libcxi https://github.com/HewlettPackard/shs-libcxi/blob/main/src/libcxi.c#L953 guess libfabric is off the hook...

PR #10962, opened 2025-04-11
prov/efa: do copy for inline send

sunkuamzn:
This commit functionally reverts the change in
https://github.com/ofiwg/libfabric/commit/8a4249790e07250a9e4a2d66c10560ce588ded38

The EFA provider in rdma-core should be able to handle inline sends without a copy to Libfabric buffer but certain workloads are failing with a truncation error when the copy is avoided

Revert the change until we can find a longer term fix


prov/efa/src/rdm/efa_rdm_pke_utils.c
```diff
@@ -84,8 +79,10 @@ ssize_t efa_rdm_pke_init_payload_from_ope(struct efa_rdm_pke *pke,
 	 * immediately after function returns, so unless using inline,
```
jiaxiyan, line 79: Need to update the comment above

PR #10975, opened 2025-04-17
prov/rxm,verbs: fix the flow control enabling checking

peinanz:
verbs provider enabled the flow control by checking qp type (IBV_QPT_RC), but the qp might not be created when the checking happened. Change to check ep type to ensure flow control is enabled for FI_EP_MSG.

### DIFF
--- a/prov/rxm/src/rxm_conn.c
+++ b/prov/rxm/src/rxm_conn.c
@@ -399,6 +399,8 @@ rxm_alloc_conn(struct rxm_ep *ep, struct util_peer_addr *peer)
 	conn->state = RXM_CM_IDLE;
 	conn->remote_index = -1;
 	conn->flags = 0;
+	conn->flow_ctrl = false;
+	conn->peer_flow_ctrl = false;
 	dlist_init(&conn->deferred_entry);
 	dlist_init(&conn->deferred_tx_queue);
 	dlist_init(&conn->deferred_sar_msgs);
@@ -484,11 +486,11 @@ static void rxm_set_peer_flow_ctrl(struct rxm_conn *conn, int cm_flow_ctrl_flag)
 		break;
 
 	case RXM_CM_FLOW_CTRL_PEER_ON:
-		conn->peer_flow_ctrl = 1;
+		conn->peer_flow_ctrl = true;
 		break;
 
 	case RXM_CM_FLOW_CTRL_PEER_OFF:
-		conn->peer_flow_ctrl = 0;
+		conn->peer_flow_ctrl = false;
 		break;
 	}
 }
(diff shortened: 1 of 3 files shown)


prov/rxm/src/rxm_conn.c
```diff
@@ -511,7 +511,7 @@ void rxm_process_connect(struct rxm_eq_cm_entry *cm_entry)
 		rxm_set_peer_flow_ctrl(conn, cm_entry->data.accept.flow_ctrl);
 	}
 
-	if (conn->flow_ctrl & conn->peer_flow_ctrl) {
```
shefty, line 514: Can we change the definitions for flow_ctrl and peer_flow_ctrl to bool, and update all places they are accessed accordingly?

peinanz: Yes, will make the change. This will make the codes more readable and consistent with the flow_ctrl ops (available) return type.

shefty (approved):
thanks

j-xiong:
Intel CI error with **mpichtestsuite** over **rxm/verbs**:
```
		Load tests in ./pt2pt
		Running tests in ./pt2pt [00:00:00]
		Program sendrecv3 exited without No Errors
		Unexpected output in sendrecv3:
		    Isends for 100 messages of size 100 took too long (1.000144 seconds)
		     Found 1 errors
		Program sendrecv3 exited without No Errors
		Unexpected output in sendrecv3:
		    Isends for 100 messages of size 100 took too long (1.000171 seconds)
		     Found 1 errors
		2 tests failed out of 70 (total runtime: 0 min 58 sec)
```

This could be related.

j-xiong:
The MPI test failure is related to the flow control. This specific test performs a few iterations of the following steps:

* send 100 messages
* post 100 receives
* wait for all the send/recvs to completes
 
The Tx and Rx size are both set to 128. The first iteration finishes quickly because there are enough peer Rx credits. In the second iteration, both sides don't have enough credits to send all the 100 messages and get into a "progress & retry" loop.  When this happens, it takes a long time for one side to receive new packets (and in the end the needed credit updates) from the peer, but the other side doesn't suffer from the same delay.

A simple OFI test ([fi_sendrecv.c.gz](https://github.com/user-attachments/files/20195772/fi_sendrecv.c.gz)) with similar logic doesn't expose the same issue, so it might have something to do with the MPI implementation.

The simplest workaround / fix is to set the credit update threshold, the Tx size, and the Rx size to a ratio of 1:2:3.

j-xiong:
**Update:** The cause of the long delay at one side of the `sendrecv3` test is in the test itself. One side would sleep for 1 second before each round while the other side doesn't. The intention is to test if one side can submit the all the 100 sends w/o any help from the peer, which is impossible when flow control is enabled with only 128 rx credits -- only 64 credits are sent back out of the 100 used in the first round when the 2nd round starts.

Will update the CI to run this test with larger rx size.

j-xiong (approved):

PR #10976, opened 2025-04-17
hmem_cuda: Add runtime fallback for unsupported dmabuf flag

jiaxiyan:
The CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE symbol requires newer CUDA driver support. While the symbol may exist during compile time in cuda.h, drivers with lower CUDA versions cannot support it. Fallback to no flags to ensure backward compatibility.

### DIFF
--- a/fabtests/common/hmem_cuda.c
+++ b/fabtests/common/hmem_cuda.c
@@ -451,16 +451,26 @@ int ft_cuda_get_dmabuf_fd(void *buf, size_t len,
 	aligned_size = (uintptr_t) ft_get_page_end((void *) ((uintptr_t) base_addr + total_size - 1),
 						    host_page_size) - (uintptr_t) aligned_ptr + 1;
 
-# if HAVE_CUDA_DMABUF_MAPPING_TYPE_PCIE
-	flags = CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE;
-# else
 	flags = 0;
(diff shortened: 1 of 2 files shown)


j-xiong (approved):

fabtests/common/hmem_cuda.c
```diff
@@ -451,16 +451,26 @@ int ft_cuda_get_dmabuf_fd(void *buf, size_t len,
...
 	cuda_ret = cuda_ops.cuMemGetHandleForAddressRange(
 						(void *)dmabuf_fd,
 						aligned_ptr, aligned_size,
 						CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD,
 						flags);
+
+	if (cuda_ret == CUDA_ERROR_INVALID_VALUE && flags != 0) {
+		FT_WARN("cuMemGetHandleForAddressRange failed with flags: %llu, "
+		       "invalid argument. Retrying with no flags.\n", flags);
+		cuda_ret = cuda_ops.cuMemGetHandleForAddressRange(
```
a-szegel, line 467: Why is it advantageous for us to retry without the flags in this case?  Same comment for below.

jiaxiyan: We don't want to fail dmabuf completely when runtime driver is too old to support this flag.

a-szegel (approved):

PR #10979, opened 2025-04-21
fabtests/scripts: use yaml safe_load

jiaxiyan:
Suppress warnings from Amazon security scan.
similar to https://github.com/ofiwg/libfabric/pull/9339

### DIFF
--- a/fabtests/scripts/runmultinode.py
+++ b/fabtests/scripts/runmultinode.py
@@ -24,7 +24,7 @@ def parse_args():
 
 def parse_config(config):
 	with open(config, "r") as f:
-		yconf = yaml.load(f, Loader=yaml.FullLoader)
+		yconf = yaml.safe_load(f)
 	return yconf
 
 def mpi_env(config):


fabtests/scripts/runmultinode.py
```diff
@@ -24,7 +24,7 @@ def parse_args():
 
 def parse_config(config):
 	with open(config, "r") as f:
-		yconf = yaml.load(f, Loader=yaml.FullLoader)
+		yconf = yaml.safe_load(f, Loader=yaml.FullLoader)
```
j-xiong, line 27: Does it make any difference?  The loader is explicitly specified here.

jiaxiyan: You are right. FullLoader is already safe. It is just Amazon security scans the source code and prefers `safe_load`.

j-xiong: So what the change does is to suppress warnings from Amazon security scan.  Please update the commit message to reflect that.

jiaxiyan: Thanks. Updated.

fabtests/scripts/runmultinode.py
```diff
@@ -24,7 +24,7 @@ def parse_args():
 
 def parse_config(config):
 	with open(config, "r") as f:
-		yconf = yaml.load(f, Loader=yaml.FullLoader)
+		yconf = yaml.safe_load(f)
```
j-xiong, line 27: Have you verified that the default loader for the `safe_load` method (`SafeLoader`) work fine here?  There might be some reason behind the use of `FullLoader` in the original code.

j-xiong: No. I just want to make sure that the capability SafeLloader is sufficient for the configuration files used for fabtests.

I think we can go with the default safe loader for now and change to FullLoader if that is discovered to be necessary.

j-xiong (approved):

PR #10981, opened 2025-04-22
[v2.1.x]prov/efa: do copy for inline send

jiaxiyan:
This commit functionally reverts the change in
https://github.com/ofiwg/libfabric/commit/8a4249790e07250a9e4a2d66c10560ce588ded38

The EFA provider in rdma-core should be able to handle inline sends without a copy to Libfabric buffer but certain workloads are failing with a truncation error when the copy is avoided

Revert the change until we can find a longer term fix

(cherry picked from commit 807bdd6d59f6725c81839ee87f94a13dc1af238a)


a-szegel (approved):

sunkuamzn (approved):

jiaxiyan:
CI failed with trn1 ICE. This is not on efa-direct path. Safe to merge

PR #10990, opened 2025-04-24
prov/efa: Only initialize devices that match FI_EFA_IFACE filter and avoid initializing non-EFA devices

sunkuamzn:
### prov/efa: Only initialize devices that match FI_EFA_IFACE filter

FI_EFA_IFACE is used to restrict the EFA devices that are visible to the
application. There is no need to initialize devices that the application
does not want to use.

However, we still want to keep track of all GIDs on the instance even if
they are not selected in FI_EFA_IFACE because we use that list to
determine if a peer is local and create SHM provider resources for such
local peers.

So this commit splits efa_device_construct into two parts: the first
part initializes efa_device fields up to and including the GID. The
second part initializes the protection domain and prov_info structs. The
second part is only executed for EFA devices that match the FI_EFA_IFACE
filter.

And it introduces g_efa_ibv_gid_list and g_efa_ibv_gid_cnt to keep track
of all EFA device GIDs on the instance irrespective of FI_EFA_IFACE

### prov/efa: Do not try to initialize non-EFA devices

On some instances, there can be non-EFA Infiniband-like devices that show
up in ibv_devinfo. We do not want to initialize those devices.

### DIFF
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -61,7 +61,11 @@ int efa_device_construct_gid(struct efa_device *efa_device,
 				 sizeof(efa_device->efa_attr));
 	if (err) {
 		err = -err;
-		EFA_INFO_ERRNO(FI_LOG_FABRIC, "efadv_query_device", err);
+		if (err == -EOPNOTSUPP) {
+			EFA_INFO(FI_LOG_FABRIC, "Not an EFA device. Will not initialize.\n");
+		} else {
+			EFA_INFO_ERRNO(FI_LOG_FABRIC, "efadv_query_device", err);
+		}
 		goto err_close;
 	}
 
@@ -253,6 +257,11 @@ int efa_device_list_initialize(void)
 					   ibv_device_list[device_idx]);
 
 		if (err) {
+
+			/* efa_device_construct returns -EOPNOTSUPP for non-EFA devices */
+			if (err == -EOPNOTSUPP)
+				continue;
+
 			ret = err;
 			goto err_free;
 		}
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -30,19 +30,15 @@
  * @brief initialize data members of a struct of efa_device
  *
  * @param	efa_device[in,out]	pointer to a struct efa_device
- * @param	device_idx[in]		device index
  * @param	ibv_device[in]		pointer to a struct ibv_device, which is used
  * 					to query attributes of the EFA device
  * @return	0 on success
  * 		a negative libfabric error code on failure.
  */
 int efa_device_construct(struct efa_device *efa_device,
-			 int device_idx,
 			 struct ibv_device *ibv_device)
 {
 	int err;
-
-	efa_device->device_idx = device_idx;
 
 	efa_device->ibv_ctx = ibv_open_device(ibv_device);
 	if (!efa_device->ibv_ctx) {
@@ -198,7 +194,7 @@ int efa_device_list_initialize(void)
 	}
 
 	for (device_idx = 0; device_idx < g_device_cnt; device_idx++) {
-		err = efa_device_construct(&g_device_list[device_idx], device_idx, ibv_device_list[device_idx]);
+		err = efa_device_construct(&g_device_list[device_idx], ibv_device_list[device_idx]);
 		if (err) {
 			ret = err;
 			goto err_free;
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -17,6 +17,7 @@
 #include <errno.h>
 
 #include <rdma/fi_errno.h>
+#include <infiniband/verbs.h>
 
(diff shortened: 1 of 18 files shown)


prov/efa/src/efa_env.c
```diff
@@ -41,29 +41,6 @@ struct efa_env efa_env = {
...
- *
- * This function fetches the default value of using EFA device's
- * RDMA capability and defines it as a configuration parameter.
- */
-void efa_env_define_use_device_rdma()
-{
-	char *str = "";
-
-	/* Specify the help info about the usage of RDMA in the device. */
-	if (!efa_device_support_rdma_read()) {
```
bwbarrett, line 55: it seems like you can still run this check, just not as part of `efa_env_define()`.

prov/efa/src/efa_device.c
```diff
@@ -178,34 +178,46 @@ int efa_device_list_initialize(void)
 
 	initialized = true;
 
-	ibv_device_list = ibv_get_device_list(&g_device_cnt);
+	ibv_device_list = ibv_get_device_list(&total_device_cnt);
```
bwbarrett, line 181: I probably would have left this as `g_device_cnt` and had a `selected_device_count` or something for the `efa_device_construct()` call, but I guess this works too.

prov/efa/src/efa_prov_info.c
```diff
@@ -471,12 +471,18 @@ int efa_prov_info_alloc(struct fi_info **prov_info_ptr,
...
 		prov_info->caps	= EFA_RDM_CAPS;
+
 		/* Claim RMA support in the efa-direct path only if read, write
-		 *  and unsolicited write are all available */
-		if (efa_device_support_rdma_read() &&
-			efa_device_support_rdma_write() &&
-			efa_device_support_unsolicited_write_recv())
+		 *  and unsolicited write are all available
+		 */
+#if HAVE_CAPS_UNSOLICITED_WRITE_RECV
```
bwbarrett, line 478: I'm not sure I understand why this change is necessary.  Could really use some documentation in the comment.

prov/efa/src/efa_env.c
sunkuamzn: Yes. We have a check during endpoint open https://github.com/ofiwg/libfabric/blob/5dd902300d2c5cdf27f436ffe430d4217c093f5c/prov/efa/src/rdm/efa_rdm_util.c#L63-L70

prov/efa/src/efa_prov_info.c
sunkuamzn: `EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV` is not defined in older versions of rdma-core and that causes compilation failures. Will add comment.

prov/efa/src/efa_device.c
```diff
@@ -178,34 +178,46 @@ int efa_device_list_initialize(void)
...
+	for (device_idx = 0; device_idx < total_device_cnt; device_idx++) {
+		if (!efa_env_allows_nic(ibv_device_list[device_idx]->name))
+			continue;
+
+		err = efa_device_construct(&g_device_list[g_device_cnt], ibv_device_list[device_idx]);
 		if (err) {
 			ret = err;
 			goto err_free;
 		}
+		g_device_cnt++;
```
a-szegel, line 209: Are we sure this is the right functional change?  We are changing g_device_cnt to be the max number of NIC's on the platform to the max number of NIC's libfabric can use.  Have you looked at all the other places g_device_cnt is used and confirmed we want this behavior?

prov/efa/src/efa_device.c
sunkuamzn: Updated. We need both the total EFA device count and the selected device count.

prov/efa/src/efa_device.c
sunkuamzn: Thank you for forcing me to double check. I missed the usage in `efa_av.c` which requires all EFA devices. I've updated the PR.

prov/efa/src/efa_domain.c
```diff
@@ -76,19 +76,19 @@ static int efa_domain_init_device_and_pd(struct efa_domain *efa_domain,
 	if (!domain_name)
 		return -FI_EINVAL;
 
-	for (i = 0; i < g_device_cnt; i++) {
-		device_name = g_device_list[i].ibv_ctx->device->name;
+	for (i = 0; i < g_efa_selected_device_cnt; i++) {
```
bwbarrett, line 79: Have to say, I don't understand this change.  Why would you ever care about the discovered device count length?  and with this change, you've made the patch way bigger.

sunkuamzn: We need the count to iterate over the discovered devices

I renamed the variables because I felt the names `g_device_cnt` and `g_device_list` were ambiguous re whether they were all IB devices or EFA devices or the selected EFA devices that match FI_EFA_IFACE

a-szegel (approved):

PR #10991, opened 2025-04-24
Applied patch from later version to prevent errors returned when using fi_close

mhhen:
[Patch](https://github.com/ofiwg/libfabric/blame/5dd902300d2c5cdf27f436ffe430d4217c093f5c/util/pingpong.c#L1882) from later version. [Commit](https://github.com/ofiwg/libfabric/commit/17d9cf2c5a168bc22b6e98d2220cd7f802861233) related to the patch. Description of the patch:

pingpong doesn't support FI_MR_ENDPOINT today, So the mr is associated with domain instead of ep. It is unsafe to close mr before closing ep because it can cause an EBUSY error when there are outstanding recvs of the mr posted to the ep/qp. This patch fixes this issue by moving the mr close after.

### DIFF
--- a/util/pingpong.c
+++ b/util/pingpong.c
@@ -1868,12 +1868,12 @@ static void pp_free_res(struct ct_pingpong *ct)
 {
 	PP_DEBUG("Freeing resources of test suite\n");
 
-	if (ct->mr != &(ct->no_mr))
-		PP_CLOSE_FID(ct->mr);
 	PP_CLOSE_FID(ct->ep);
 	PP_CLOSE_FID(ct->pep);
 	PP_CLOSE_FID(ct->rxcq);
 	PP_CLOSE_FID(ct->txcq);
+	if (ct->mr != &(ct->no_mr))
+		PP_CLOSE_FID(ct->mr);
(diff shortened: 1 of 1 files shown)


j-xiong:
Why not cherry-picking the original commit to this branch?
```
git cherry-pick -x 17d9cf2
```

mhhen:
No specific reason. If necessary, I can update this pull-request to cherry-pick the change.

j-xiong:
Yes, please do the cherry-picking. That's the preferred way. Thanks.

mhhen:
I was just about to update with revert and cherry-pick, but see that shijin-aws is working on it.

mhhen:
I used:
```
git reset --hard 3a14237802afb28c7d7a0d404a55dbddaa87147e
git cherry-pick -x 17d9cf2
git push origin --force
```

mhhen:
@j-xiong What needs to be done to proceed?

j-xiong (approved):

j-xiong:
The actions on the 1.15 branch is out-of-date. Ignoring the failures.

PR #10999, opened 2025-04-28
prov/efa: Remove duplicate filtering on FI_EFA_IFACE

sunkuamzn:
In
https://github.com/ofiwg/libfabric/commit/d635b09af83f4c06ae08328200f94d7a898be4af the filtering on FI_EFA_IFACE was moved to the device initialization section in EFA_INI. Remove the filter from the fi_getinfo path


jiaxiyan (approved):

bwbarrett (dismissed):
You have a word wrapping fail in your commit message.

j-xiong:
bot:aws:retest

shijin-aws:
Need to look at the unit test failure

prov/efa/test/efa_unit_test_info.c
```diff
@@ -876,34 +876,6 @@ void test_efa_nic_select_first_device_with_surrounding_comma_matches() {
 	free(nic_name);
 }
 
-/**
```
shijin-aws, line 879: I am wondering why the other `test_efa_nic*` unit tests are still working. Is it false positive?

shijin-aws (approved):

prov/efa/test/efa_unit_test_info.c
shijin-aws: Then the tests actually lose their value now?

sunkuamzn: Good point. Let me move the `all` and multiple NIC selection tests to fabtests as well

PR #11003, opened 2025-04-30
prov/tcp: Add FI_GET_FD support to tcp provider

besnardjb:
Adds the ability to retrieve the underlying file descriptor associated with an active endpoint in the tcp provider. This functionality is exposed through `fi_control()` using the new `FI_GET_FD` command.

The primary motivation is to allow applications to tune socket options that are not exposed through the libfabric API, such as:
 - `SO_SNDBUF`: Control send buffer size
 - `SO_RCVBUF`: Control receive buffer size
 - `SO_BUSY_POLL`: Enable busy polling for lower latency
 - ...


j-xiong (approved):
Thanks!

Issue #10974, opened 2025-04-16
prov/efa: use-after-free when using multiple RDM EPs in a single domain


k0zmo:
Commit f58958062 moved ope bookkeeping (for example, `ope_longcts_send_list`) from `efa_rdm_ep` to `efa_domain`. This makes all the endpoints from a single domain to share this list and insert entries into it. Each entry - `efa_rdm_ope`, is created from a bufpool, which lifetime is bound to the endpoint. If we close and destroy an endpoint, leaving the rest of EPs alive, subsequent polling of the CQ (bound to the other EP) calls `efa_domain_progress_rdm_peers_and_queues` which iterate over that list and will dereference already freed `efa_rdm_ope` - leading to a segfault. 

This makes it impossible to reliably handle multiple RDM EPs using EFA provider using a single domain. A workaround is to create at most one EP for a given domain.

sunkuamzn:
@k0zmo thank you for reporting. We will investigate.

shijin-aws:
It seems the efa ep close doesn't grab a lock around the buffer pool destroy (ope release) https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L958. And it's a bug.

shijin-aws:
We can `dlist_remove(txe->entry)` when releasing txe, which will remove the txe from `ope_longcts_send_list`

shijin-aws:
@k0zmo I can open a PR for you to try, does that sound good to you?

k0zmo:
Sure, I can give it a try tomorrow

shijin-aws:
@k0zmo can you try https://github.com/ofiwg/libfabric/pull/11008

k0zmo:
https://github.com/ofiwg/libfabric/pull/11008 fixes the issue for me.

shijin-aws:
PR is fixed

PR #11008, opened 2025-05-05
prov/efa: Fix the ope list removal

shijin-aws:
This PR fixes 2 issues:

1. Currently, an ope is not removed from the domain level
list ope_longcts_list during release. This can cause
an use-after-free error when application polls the CQ after closing EPs
that have outstanding long cts opes.

2. efa_rdm_ep_destroy_buffer_pools does not grab
a lock before iterating the dlists in the ep/domain
that are supposed to be protected by the srx
lock. This patch fixes it.


shijin-aws:
Fixes https://github.com/ofiwg/libfabric/issues/10974

j-xiong:
bot:aws:retest

k0zmo:
One comment regarding the description of this PR. You don't necessarily need to use a single CQ for multiple EP to crash it. I'm getting use-after-free when using 1-to-1 CQ and EP. This is because calling `fi_cq_read` (for the still not closed EP) calls `efa_rdm_cq_progress` which at the end calls `efa_domain_progress_rdm_peers_and_queues(domain)` and that function traverses through `ope_longcts_send_list`.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -167,6 +175,14 @@ void efa_rdm_rxe_release_internal(struct efa_rdm_ope *rxe)
 
 	dlist_remove(&rxe->ep_entry);
 
+	/**
+	 * Make sure the entry is removed
+	 * from ope_longcts_list when the ope
+	 * is released for whatever reasons.
+	 */
+	if (rxe->state == EFA_RDM_TXE_SEND)
```
a-szegel, line 183: Is rxe->state really equal to `EFA_RDM_TXE_SEND`?  Why?

shijin-aws: Yes, the macro is a bit misleading, because we can post a send from an rxe as well https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke_nonreq.c#L411. Lemme rename it to `EFA_RDM_OPE_SEND`

a-szegel (approved):

PR #11012, opened 2025-05-07
prov/efa: Fix unit test build warning

shijin-aws:
assert is only compiled for debug build,
which will cause a unused warning for the efa_domain variable in non-debug build. Also, we should use the cmocka routine
assert_true to check the conditions anyway.


sunkuamzn (approved):

shijin-aws:
bot:aws:retest

a-szegel:
bot:aws:retest

PR #11013, opened 2025-05-07
prov/efa: Create 1:1 relationship between domain and PD, move address handle to domain level

shijin-aws:
This PR has 2 changes:

1. Make 1:1 relationship between efa domain and pd: Today, protection domain (PD) is created in efa_device discovery, when
    multiple domains are created against the same efa_device in the
    same process, there will share the same PD. Such implementation breaks
    the boundary of domain that resources created in different
    domains can finally interfere with each other, such as MR and AH.
    
    This patch fixes this issue by making PD created/destroyed by domain
    open and close, so there is a 1:1 relationship between domain
    and PD.

2. Move address handle to domain level: Today, the efa address handle (AH) resources are
    created and tracked in the AV level. This is wrong because for multiple AVs created in the same domain, they share the same PD and the created AHs will not be independent. When one AV gets closed and destroyed its AHs, WQEs posted via other AHs that share the same PD and GID will get aborted.
    
    This patch fixes this issue by doing the following:
    
    1. Move ah_map (hash map, key is GID, value is AH) from efa_av to efa_domain which now has a 1:1 map to PD. The map is protected by the domain lock.
    2. When fi_av_insert decide to create a new av entry, look up the AH in the device ah_map (in the locking), if it's not found, create AH and initialize its refcnt as 1 and insert to hashmap. If it's found, return the AH and increment its refcnt.
    3. In fi_av_remove, do the look up again in ah_map, decrement the refcnt of the returned AH. If the refcnt is 0, destroy the AH and delete it from the hashmap
    4. In efa_domain close, iterate the ah map again and destroy all uncleaned AHs (with warning).

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -91,8 +91,13 @@ static int efa_domain_init_device_and_pd(struct efa_domain *efa_domain,
 	if (i == g_efa_selected_device_cnt)
 		return -FI_ENODEV;
 
+	efa_domain->ibv_pd = ibv_alloc_pd(efa_domain->device->ibv_ctx);
+	if (!efa_domain->ibv_pd) {
+		EFA_WARN(FI_LOG_DOMAIN, "Failed to allocated ibv_pd: %d\n", errno);
+		return -FI_ENOMEM;
+	}
+
 	EFA_INFO(FI_LOG_DOMAIN, "Domain %s selected device %s\n", domain_name, device_name);
-	efa_domain->ibv_pd = efa_domain->device->ibv_pd;
 	return 0;
 }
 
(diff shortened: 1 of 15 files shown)


shijin-aws:
adding unit tests now, open the PR to let CI run first.

shijin-aws:
bot:aws:retest

prov/efa/src/efa_domain.c
```diff
@@ -351,6 +354,20 @@ static int efa_domain_close(fid_t fid)
 		efa_domain->cache = NULL;
 	}
 
+	/* Clean up ah_map if any entries remain */
+	ofi_genlock_lock(&efa_domain->util_domain.lock);
+	if (efa_domain->ah_map) {
+		EFA_WARN(FI_LOG_DOMAIN, "AH map not empty during domain close! Cleaning up ...\n");
```
a-szegel, line 360:
```
ret = ofi_domain_close(&efa_domain->util_domain);
```
Should return -FI_EBUSY if users have domain resources open (AV, EP....).  We should move deleting the util domain to the top of this function.  Also, this warning should only be possible to hit if we have a bug in our code.  Warnings, are for users, asserts() are for us, and this should be an assert.  The rest of the cleanup should also be removed from here. I recommend a follow up PR that fixes our  domain cleanup logic.

a-szegel (approved):
looks good, tyvm.  I am fine with follow up PR for my comment.

PR #11014, opened 2025-05-07
prov/efa: Use Mutex when removing from g_efa_domain_list

a-szegel:
Commit 633d2919 added a mutex lock around g_efa_domain_list list insertion in order to keep the EFA control interface thread safe, but forgot to add the mutex lock around the list item removal.  This commit fixes that mistake.


shijin-aws (approved):

a-szegel:
bot:aws:retest

PR #11027, opened 2025-05-15
prov/hook: Add monitor hook provider and sampler

philippfriese:
This PR introduces a new hooking provider `ofi_hook_monitor`, which extends the `ofi_hook_profile` provider.
API calls and processed data volume are gathered similar to the profile provider and accumulated per CQ, MR, and data operation calls. 
The hooking provider exposes this data via a communication file for continuous readout via a sampler. A simple sampler is provided by this PR.

Using the `ofi_hook_monitor` provider, libfabric applications can be monitored for their network usage. This is useful for per-process monitoring of userspace-only networking, such as RDMA, as well as for non-MPI applications without an integrated profiling infrastructure like PMPI (such as OpenFAM or pure libfabric applications). 

#### Data Export
The sampler periodically reads the communication file and extracts the stored monitoring data. During sampling, the sampler sets the data request bit in this file, which indicates that the sampler has read the currently exposed data and requests new data. Every N intercepted API calls, where N is configurable by the user, the provider will check whether the data request bit is set. If so: (1) the counter data accumulated by the provider is copied to the file, (2) the provider counters are reset, and (3) the data request bit is reset in order to indicate that new data is available.

The communication file is expected to be stored on a tmpfs. If available and unless otherwise specified by the user, files will be created under the tmpfs `/dev/shm`. One communication file is created per core provider.

Usage example of this provider and the sampler:

```bash
# launch monitored application & second pingpong pair
FI_HOOK=monitor fi_pingpong -I 100000 & ; fi_pingpong -I 100000 localhost

# launch sampler
fi_mon_sampler -o $HOME -w 1000 -f csv /dev/shm/ofi/$UID/$HOSTNAME
```
> Note: Assumes being launched in separate tty's or via backgrounding.

The sampler stores data in a CSV file, one column per monitored function, size bucket, and per API count (`_c`) and sum of processed data (`_s`). 
Example output of CSV file (first four columns & rows):
```
mon_recv_0_64_c,mon_recv_0_64_s,mon_recv_64_512_c,mon_recv_64_512_s
12069,0,0,0
95354,0,0,0
95092,0,0,0
98491,0,0,0
```

#### Commits
- `prov/hook: Add monitoring hook provider` contains the `ofi_hook_monitor` provider.
- `util: Add fi_mon_sampler` contains a simple sampler which reads data at a configurable frequency and exports it to a CSV file.

The hooking provider is presented in the paper "Application-focused Network Monitoring", published at ISC-HPC 2025.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -153,7 +153,8 @@ linkback = src/libfabric.la
 bin_PROGRAMS = \
 	util/fi_info \
 	util/fi_strerror \
-	util/fi_pingpong
+	util/fi_pingpong \
+	util/fi_mon_sampler
 
 bin_SCRIPTS =
 
@@ -168,6 +169,10 @@ util_fi_strerror_LDADD = $(linkback)
 util_fi_pingpong_SOURCES = \
 	util/pingpong.c
 util_fi_pingpong_LDADD = $(linkback)
+
+util_fi_mon_sampler_SOURCES = \
+	util/mon_sampler.c
+util_fi_mon_sampler_LDADD = $(linkback)
 
 nodist_src_libfabric_la_SOURCES =
 src_libfabric_la_SOURCES =			\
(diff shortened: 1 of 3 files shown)


philippfriese:
@j-xiong I see, thanks for the info! I've added the dummy page and pushed an updated commit.

philippfriese:
Pushed another set of updated commits addressing build failures on macOS.

j-xiong (approved):

PR #11031, opened 2025-05-15
[v2.1.x] prov/cxi: decouple existence CXI_MAP_IOVA_ALLOC for build

swelch:
Allow libfabric to build with both new and older drivers passing CXI_MAP_IOVA_ALLOC on event queue allocation if defined. This is to facilitate upstream builds independent of SHS release.

(cherry picked from commit d56969009907c35842011708ae55373a85cce8c3)


swelch:
@jswaro - could you review this OFIWG PR since John is out?

jswaro (approved):

PR #11032, opened 2025-05-15
fabtests/efa: Add FI_EFA_IFACE tests

shijin-aws:
Test efa device selection with FI_EFA_IFACE
as well as the domain name approach.

This change requires introducing different
server and client environment to cmdline_args


sunkuamzn (approved):
Thank you!

PR #11041, opened 2025-05-19
prov/verbs: Fix memory leak of fi_info structure in vrb_domain()

sydidelot:
The recent commit d15a7701d ("prov/verbs: Make a copy of the provider info in the fabric object) introduced a regression: the pointer returned by vrb_get_verbs_info() is never freed.

Rather than adding a missing call to fi_freeinfo(), I believe it's cleaner to move the domain attribute check directly into the function itself, while holding the lock.

For such a simple check, duplicating the fi_info structure is unnecessary.

Note: Leak found by ASAN.

### DIFF
--- a/prov/verbs/src/verbs_domain.c
+++ b/prov/verbs/src/verbs_domain.c
@@ -308,21 +308,23 @@ static struct fi_ops_domain vrb_dgram_domain_ops = {
 	.query_collective = fi_no_query_collective,
 };
 
-static struct fi_info *
-vrb_get_verbs_info(const char *domain_name)
+static int
+vrb_check_verbs_info(struct fid_fabric *fabric, struct fi_info *info)
 {
 	const struct fi_info *fi;
+	int ret = -FI_ENODATA;
 
 	ofi_mutex_lock(&vrb_info_mutex);
 	for (fi = vrb_util_prov.info; fi; fi = fi->next) {
-		if (!strcmp(fi->domain_attr->name, domain_name)) {
-			ofi_mutex_unlock(&vrb_info_mutex);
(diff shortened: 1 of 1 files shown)


sydidelot:
@shefty @j-xiong Could I ask you to review this memory leak I accidentally introduced in my patch series for FI_RESCAN?
I thought I had run all my tests with ASAN, but that apparently wasn’t the case. Sorry about that.

prov/verbs/src/verbs_domain.c
```diff
@@ -308,21 +308,23 @@ static struct fi_ops_domain vrb_dgram_domain_ops = {
...
+	int ret = -FI_ENODATA;
 
 	ofi_mutex_lock(&vrb_info_mutex);
 	for (fi = vrb_util_prov.info; fi; fi = fi->next) {
-		if (!strcmp(fi->domain_attr->name, domain_name)) {
-			ofi_mutex_unlock(&vrb_info_mutex);
-			return fi_dupinfo(fi);
+		if (!strcmp(fi->domain_attr->name, info->domain_attr->name)) {
+			ret = ofi_check_domain_attr(&vrb_prov, fabric->api_version,
+					fi->domain_attr, info);
```
j-xiong, line 321: please align the arguments to keep it consistent with other places in the same file.

j-xiong (commented):
Looks good to me, just some nitpicks.

prov/verbs/src/verbs_domain.c
sydidelot: Thanks for the quick review! This is fixed now.

j-xiong (approved):

PR #11045, opened 2025-05-20
prov/efa: Extend domain ops to allow querying of QP and CQ attributes

jiaxiyan:
Following rdma core changes, enable querying of QP work queue and completion queue's attributes in domain operations, which can be used for accelerator driven datapath.
Enable creation of CQs on top of pre-allocated memory buffers. The memory can be passed by supplying
the struct fi_efa_cq_init_attr with the flag FI_EFA_CQ_INIT_FLAGS_EXT_MEM_DMABUF.

### DIFF
--- a/man/fi_efa.7.md
+++ b/man/fi_efa.7.md
@@ -201,6 +201,10 @@ struct fi_efa_ops_domain {
 			  uint16_t *remote_qpn, uint32_t *remote_qkey);
 	int (*query_qp_wqs)(struct fid_ep *ep_fid, struct fi_efa_wq_attr *sq_attr, struct fi_efa_wq_attr *rq_attr);
 	int (*query_cq)(struct fid_cq *cq_fid, struct fi_efa_cq_attr *cq_attr);
+	int (*cq_open_ext)(struct fid_domain *domain_fid,
+			   struct fi_cq_attr *attr,
+			   struct fi_efa_cq_init_attr *efa_cq_init_attr,
+			   struct fid_cq **cq_fid, void *context);
 };
 ```
 
@@ -314,6 +318,49 @@ struct fi_efa_cq_attr {
 
 #### Return value
 **query_cq()** returns 0 on success, or the value of errno on failure
+(which indicates the failure reason).
+
+### cq_open_ext
+This op creates a completion queue with external memory provided via dmabuf.
+The memory can be passed by supplying the following struct.
+
+```c
+struct fi_efa_cq_init_attr {
+	uint64_t flags;
+	struct {
+		uint8_t *buffer;
+		uint64_t length;
+		uint64_t offset;
+		uint32_t fd;
+	} ext_mem_dmabuf;
+};
+```
+
+*flags*
+:	A bitwise OR of the various values described below.
+
+	FI_EFA_CQ_INIT_FLAGS_EXT_MEM_DMABUF:
+		create CQ with external memory provided via dmabuf.
+
+*ext_mem_dmabuf*
+:	Structure containing information about external memory when using
+	FI_EFA_CQ_INIT_FLAGS_EXT_MEM_DMABUF flag.
+
+	*buffer*
+	:	Pointer to the memory mapped in the process's virtual address space. 
+		The field is optional, but if not provided, the use of CQ poll interfaces should be avoided.
+
+	*length*
+	:	Length of the memory region to use.
+
+	*offset*
+	:	Offset within the dmabuf.
+
+	*fd*
+	:	File descriptor of the dmabuf.
+
+#### Return value
+**cq_open_ext()** returns 0 on success, or the value of errno on failure
 (which indicates the failure reason).
 
 # Traffic Class (tclass) in EFA
--- a/man/fi_efa.7.md
+++ b/man/fi_efa.7.md
@@ -199,13 +199,15 @@ struct fi_efa_ops_domain {
 	int (*query_mr)(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr);
 	int (*query_addr)(struct fid_ep *ep_fid, fi_addr_t addr, uint16_t *ahn,
 			  uint16_t *remote_qpn, uint32_t *remote_qkey);
+	int (*query_qp_wqs)(struct fid_ep *ep_fid, struct fi_efa_wq_attr *sq_attr, struct fi_efa_wq_attr *rq_attr);
+	int (*query_cq)(struct fid_cq *cq_fid, struct fi_efa_cq_attr *cq_attr);
 };
 ```
 
 It contains the following operations
 
 ### query_mr
-This op query an existing memory registration as input, and outputs the efa
+This op queries an existing memory registration as input, and outputs the efa
 specific mr attribute which is defined as follows
 
 ```c
@@ -256,6 +258,63 @@ This op queries the following address information for a given endpoint and desti
 
 #### Return value
 **query_addr()** returns FI_SUCCESS on success, or -FI_EINVAL on failure.
+
+### query_qp_wqs
+This op queries EFA specific Queue Pair work queue attributes for a given endpoint.
+It retrieves the send queue attributes in sq_attr and receive queue attributes in rq_attr, which is defined as follows.
+
+```c
+struct fi_efa_wq_attr {
+    uint8_t *buffer;
+    uint32_t entry_size;
+    uint32_t num_entries;
+    uint32_t *doorbell;
+    uint32_t max_batch;
+};
+```
+
+*buffer*
+:	Queue buffer.
+
+*entry_size*
+:	Size of each entry in the queue.
+
+*num_entries*
+:	Maximal number of entries in the queue.
+
+*doorbell*
+:	Queue doorbell.
+
+*max_batch*
+:	Maximum batch size for queue submissions.
+
+#### Return value
+**query_qp_wqs()** returns 0 on success, or the value of errno on failure
+(which indicates the failure reason).
+
+### query_cq
+This op queries EFA specific Completion Queue attributes for a given cq.
+
+```c
+struct fi_efa_cq_attr {
+    uint8_t *buffer;
+    uint32_t entry_size;
+    uint32_t num_entries;
+};
+```
+
+*buffer*
+:	Completion queue buffer.
+
+*entry_size*
+:	Size of each completion queue entry.
+
+*num_entries*
+:	Maximal number of entries in the completion queue.
+
+#### Return value
+**query_cq()** returns 0 on success, or the value of errno on failure
+(which indicates the failure reason).
 
 # Traffic Class (tclass) in EFA
 To prioritize the messages from a given endpoint, user can specify `fi_info->tx_attr->tclass = FI_TC_LOW_LATENCY` in the fi_endpoint() call to set the service level in rdma-core. All other tclass values will be ignored.
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -72,6 +72,7 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	have_caps_rnr_retry=0
 	have_caps_rdma_write=0
 	have_caps_unsolicited_write_recv=0
+	have_caps_cq_with_ext_mem_dmabuf=0
 	have_ibv_is_fork_initialized=0
 	efa_support_data_in_order_aligned_128_byte=0
 	efadv_support_extended_cq=0
@@ -103,6 +104,11 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 		AC_CHECK_DECL(EFADV_DEVICE_ATTR_CAPS_UNSOLICITED_WRITE_RECV,
 			[have_caps_unsolicited_write_recv=1],
 			[have_caps_unsolicited_write_recv=0],
+			[[#include <infiniband/efadv.h>]])
+
+		AC_CHECK_DECL(EFADV_DEVICE_ATTR_CAPS_CQ_WITH_EXT_MEM_DMABUF,
+			[have_caps_cq_with_ext_mem_dmabuf=1],
+			[have_caps_cq_with_ext_mem_dmabuf=0],
 			[[#include <infiniband/efadv.h>]])
 
 		AC_CHECK_DECL([ibv_is_fork_initialized],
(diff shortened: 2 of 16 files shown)


shijin-aws:
@mrgolin can you take a look as well

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,98 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+	int ret;
+
+	memset(sq_attr, 0, sizeof(*sq_attr));
+	memset(rq_attr, 0, sizeof(*rq_attr));
+
+	base_ep = container_of(ep_fid, struct efa_base_ep, util_ep.ep_fid);
+	ret = efadv_query_qp_wqs(base_ep->qp->ibv_qp, &qp_sq_attr, &qp_rq_attr, sizeof(qp_sq_attr));
+	if (ret) {
+		EFA_WARN(FI_LOG_DOMAIN, "efadv_query_qp_wqs failed. err: %d\n", ret);
+		return ret;
```
shijin-aws, line 481: Do you need to translate this `ret` to FI_* ? e.g. can `ret` be positive? If so, we need to convert it to a negative integer

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,98 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+{
+	struct efa_cq *efa_cq = container_of(cq_fid, struct efa_cq, util_cq.cq_fid);
+	struct efadv_cq_attr attr = {0};
+	int ret;
+
+	memset(cq_attr, 0, sizeof(*cq_attr));
+
+	ret = efadv_query_cq(ibv_cq_ex_to_cq(efa_cq->ibv_cq.ibv_cq_ex), &attr, sizeof(attr));
+	if (ret) {
+		EFA_WARN(FI_LOG_DOMAIN, "efadv_query_cq failed. err: %d\n", ret);
```
shijin-aws, line 526: Similar comment on the error code. I think returning `-FI_EINVAL` could be good

shijin-aws: You can still keep `ret` in the log, but the return code has to be compliant with libfabric semantic

man/fi_efa.7.md
```diff
@@ -240,6 +242,62 @@ struct fi_efa_mr_attr {
 **query_mr()** returns 0 on success, or the value of errno on failure
 (which indicates the failure reason).
 
+### query_qp_wqs
+This op queries EFA specific Queue Pair work queue attributes.
```
shijin-aws, line 246: Do you need to explain what sq and rq means in your input?

shijin-aws (dismissed):

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,98 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+/**
+ * @brief Query EFA specific Completion Queue attributes
+ *
+ * @param cq_fid pointer to completion queue fid
+ * @param cq_attr pointer to fi_efa_cq_attr
+ * @return 0 on success, negative integer on failure
+ */
+static int efa_domain_query_cq(struct fid_cq *cq_fid, struct fi_efa_cq_attr *cq_attr)
+{
+	struct efa_cq_ext *cq_ext = container_of(cq_fid, struct efa_cq_ext, cq_fid);
```
shijin-aws, line 518: What will happen if you call this function with a cq that is not created via `efa_cq_open_ext`?

prov/efa/src/efa_cq.h
```diff
@@ -148,11 +152,24 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
...
+
 	*ibv_cq_ex = efadv_create_cq(ibv_ctx, &init_attr_ex,
 				     &efadv_cq_init_attr,
 				     sizeof(efadv_cq_init_attr));
 
 	if (!*ibv_cq_ex) {
+#if HAVE_CAPS_CQ_WITH_EXT_MEM_DMABUF
+		if (efa_device_support_cq_with_ext_mem_dmabuf() &&
+		    (flags & EFADV_CQ_INIT_FLAGS_EXT_MEM_DMABUF))
+			return -FI_EINVAL;
```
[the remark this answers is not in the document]
jiaxiyan: Where is the errno returned?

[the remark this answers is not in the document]
shijin-aws: `errno` is not returned by function, it's set by process. We can log the `errno` in the warning, but I don't think we should return it as is because it may not be a recognized Libfabric error.

man/fi_efa.7.md
```diff
@@ -299,6 +303,36 @@ struct fi_efa_cq_attr {
 **query_cq()** returns 0 on success, or the value of errno on failure
 (which indicates the failure reason).
 
+### cq_open_ext
+This op creates a completion queue with external memory provided via dmabuf.
+The memory can be passed by supplying the following struct with EFADV_CQ_INIT_FLAGS_EXT_MEM_DMABUF flag.
+
+```c
+struct fi_efa_ext_mem_dmabuf {
+    uint8_t  *buffer;
```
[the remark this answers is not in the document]
jiaxiyan: I copied this from rdma-core

prov/efa/src/efa_cq.h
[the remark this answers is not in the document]
jiaxiyan: updated

prov/efa/configure.m4
```diff
@@ -165,6 +173,26 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 			[have_efadv_sl=1],
 			[have_efadv_sl=0],
 			[[#include <infiniband/efadv.h>]])
+		
+		have_efadv_query_qp_wqs=1
+		AC_CHECK_DECL([efadv_query_qp_wqs],
+			[],
+			[have_efadv_query_qp_wqs=0],
+			[[#include <infiniband/efadv.h>]])
+		AC_CHECK_MEMBER([struct efadv_wq_attr.max_batch],
```
mrgolin, line 182: This check seems redundant.

prov/efa/src/efa_cq.h
```diff
@@ -148,11 +152,31 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
 		efadv_cq_init_attr.wc_flags |= EFADV_WC_EX_WITH_IS_UNSOLICITED;
 #endif
 
+#if HAVE_CAPS_CQ_WITH_EXT_MEM_DMABUF
+	if (efa_device_support_cq_with_ext_mem_dmabuf() &&
```
mrgolin, line 156: Shouldn't we return some error when external memory was provided but not supported? (check requested vs. supported flags)

prov/efa/src/efa_device.c
```diff
@@ -360,6 +360,26 @@ bool efa_device_support_unsolicited_write_recv(void)
...
+ *
+ * @return a boolean indicating that creating CQs with external memory buffers
+ * by passing dmabuf is supported.
+ */
+#if HAVE_CAPS_CQ_WITH_EXT_MEM_DMABUF
+bool efa_device_support_cq_with_ext_mem_dmabuf(void)
+{
+	assert(g_efa_selected_device_cnt > 0);
+
+	return g_efa_selected_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_CQ_WITH_EXT_MEM_DMABUF;
```
mrgolin, line 374: nit: I would add !! to make it 0/1.

prov/efa/src/efa_domain.c
```diff
@@ -421,7 +421,7 @@ efa_domain_query_mr(struct fid_mr *mr_fid, struct fi_efa_mr_attr *mr_attr)
 	ret = efadv_query_mr(efa_mr->ibv_mr, &attr, sizeof(attr));
 	if (ret) {
 		EFA_WARN(FI_LOG_DOMAIN, "efadv_query_mr failed. err: %d\n", ret);
-		return ret;
+		return (ret == EOPNOTSUPP) ? -FI_EOPNOTSUPP : -FI_EINVAL;
```
mrgolin, line 424: I think this is true for old kernels that don't have uverbs ioctl support. With most kernels will will get EPROTONOSUPPORT when the driver doesn't have query_mr implemented. I would also avoid pushing it in this PR.

prov/efa/src/efa_cq.h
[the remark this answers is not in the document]
mrgolin: I think you should have an error translation from errno to FI_*, we shouldn't turn everything to FI_EINVAL.

prov/efa/src/efa_cq.h
mrgolin: Also it's very strange that the dmabuf info is passed as an arg to the function but the decision whether to use is comes from another place.

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,223 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+	struct efadv_wq_attr qp_sq_attr = {0};
+	struct efadv_wq_attr qp_rq_attr = {0};
+	int ret;
+
+	memset(sq_attr, 0, sizeof(*sq_attr));
+	memset(rq_attr, 0, sizeof(*rq_attr));
+
+	base_ep = container_of(ep_fid, struct efa_base_ep, util_ep.ep_fid);
+	ret = efadv_query_qp_wqs(base_ep->qp->ibv_qp, &qp_sq_attr, &qp_rq_attr, sizeof(qp_sq_attr));
+	if (ret) {
```
mrgolin, line 479: Need to also check ```qp_sq_attr.comp_mask``` and ```qp_rq_attr.comp_mask``` are zero.

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,223 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+static int efa_domain_query_cq(struct fid_cq *cq_fid, struct fi_efa_cq_attr *cq_attr)
+{
+	struct efa_cq *efa_cq = container_of(cq_fid, struct efa_cq, util_cq.cq_fid);
+	struct efadv_cq_attr attr = {0};
+	int ret;
+
+	memset(cq_attr, 0, sizeof(*cq_attr));
+
+	ret = efadv_query_cq(ibv_cq_ex_to_cq(efa_cq->ibv_cq.ibv_cq_ex), &attr, sizeof(attr));
+	if (ret) {
```
mrgolin, line 525: Please check ```attr.comp_mask``` is zero.

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,223 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+	.strerror = efa_cq_strerror,
+};
+
+static void efa_cq_ext_progress_no_op(struct util_cq *cq)
+{
+	return;
+}
+
+/**
+ * @brief Create a completion queue with external memory provided via dmabuf.
```
mrgolin, line 581: Having a dedicated function that allows supplying another param doesn't look as a scalable approach. Create CQ might have multiple inputs and support various permutations of them. How do you plan to handle it?

prov/efa/src/efa_domain.c
```diff
@@ -453,8 +453,223 @@ efa_domain_query_mr(struct fid_mr *mr, struct fi_efa_mr_attr *mr_attr)
...
+#else
+static int efa_domain_query_cq(struct fid_cq *cq_fid, struct fi_efa_cq_attr *cq_attr)
+{
+	return -FI_ENOSYS;
+}
+#endif /* HAVE_EFADV_QUERY_CQ */
+
+
+#if HAVE_CAPS_CQ_WITH_EXT_MEM_DMABUF && HAVE_EFADV_CQ_EX
+static struct fi_ops efa_cq_ext_fi_ops = {
```
mrgolin, line 545: Why not use existing ```efa_cq_fi_ops```?

prov/efa/src/efa_domain.c
jiaxiyan: I update the function to pass a struct `fi_efa_cq_init_attr` similar to `efadv_cq_init_attr`. Is this better?

jiaxiyan:
bot:aws:retest

prov/efa/src/efa_domain.c
shijin-aws: For ABI compatibility, we may need some check in Libfabric's `fi_efa_wq_attr` in the future as well when we introduce more fields. New fields have to be introduced to the end of the structure, and we need to check Libfabric API version before we access the new fields from the sq_attr ptr from the application - because it may compile with an older Libfabric that do not allocate memory for these new fields

shijin-aws (approved):

sunkuamzn:
bot:aws:retest

jiaxiyan:
bot:aws:retest

prov/efa/src/efa_cq.h
```diff
@@ -148,11 +153,27 @@ static inline int efa_cq_ibv_cq_ex_open(struct fi_cq_attr *attr,
 		efadv_cq_init_attr.wc_flags |= EFADV_WC_EX_WITH_IS_UNSOLICITED;
 #endif
 
+#if HAVE_CAPS_CQ_WITH_EXT_MEM_DMABUF
+	efadv_cq_init_attr.flags = efa_cq_init_attr->flags;
```
mrgolin: What ensures that Libfabric and rdma-core flags have same values?

jiaxiyan: Libfabric will pass flags like `FI_EFA_CQ_INIT_FLAGS_EXT_MEM_DMABUF` defined in `fi_ext_efa.h` to rdma-core.  What kind of check are you suggesting? Only set the flag `FI_EFA_CQ_INIT_FLAGS_EXT_MEM_DMABUF` in `efa_domain_cq_open_ext` for now?

jiaxiyan:
bot:aws:retest

prov/efa/src/efa_cq.h
mrgolin: I suggest setting EFADV_CQ_INIT_FLAGS_EXT_MEM_DMABUF explicitly when FI_EFA_CQ_INIT_FLAGS_EXT_MEM_DMABUF is set rather than relying on same layout.

jiaxiyan: updated

shijin-aws:
bot:aws:retest

shijin-aws (approved):

PR #11048, opened 2025-05-21
Fabtests: Add New Benchmark rma_tx_completion

a-szegel:
This commit adds a new benchmark rma_tx_completion, which measures the time for a TX completion to occur on the writer. The test does not do any data validation.  The test sets xfers_per_transfer to 1 to not have show_perf divide the latency result by 2 (the test is uni-directional already).


sunkuamzn:
bot:aws:retest

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -381,6 +381,75 @@ int pingpong_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
...
+	if (ft_check_opts(FT_OPT_ENABLE_HMEM))
+		inject_size = 0;
+
+	if (ft_check_opts(FT_OPT_VERIFY_DATA)){
+		FT_ERR("Verify Data Is Not Supported");
+		return EXIT_FAILURE;
+	}
+
+	if (opts.transfer_size == 0) {
+		FT_ERR("Zero-sized transfers not supported");
```
sunkuamzn, line 408: why?

a-szegel:
This is broken for write-data currently.  On the first iteration, both the writer/writee make it to ft_sync and hang.  Need to debug when I have time.

a-szegel:
Fixed all the bugs

fabtests/benchmarks/rma_tx_completion.c
```diff
@@ -0,0 +1,140 @@
...
+			ft_csusage(argv[0], "rma_tx_completion test using RMA operations.");
+			ft_benchmark_usage();
+			FT_PRINT_OPTS_USAGE("-o <op>", "rma op type: write|writedata (default: write)\n");
+			ft_longopts_usage();
+			return EXIT_FAILURE;
+		}
+	}
+
+	/* rma_tx_completion test does not support data validation */
+	if (opts.rma_op != FT_RMA_WRITEDATA && ft_check_opts(FT_OPT_VERIFY_DATA)) {
```
sunkuamzn: `ft_tx_rma` and `ft_rx_rma` automatically handle buffer verification but it's fine to not allow them in this test if you haven't tested it

You still need to make this check consistent with the check in `rma_tx_completion` which doesn't support verification for all operations

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -381,6 +381,78 @@ int pingpong_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
...
+				opts.argc, opts.argv);
+		else
+			show_perf(NULL, opts.transfer_size, opts.iterations, &start, &end, 1);
+
+		/* Inform RMA target that the test has ended */
+		ret = ft_sync();
+		if (ret)
+			return ret;
+	} else {
+		if (rma_op == FT_RMA_WRITEDATA) {
```
sunkuamzn, line 439: I think you can remove this check

`ft_rx_rma` will automatically poll the last byte for `fi_write`
https://github.com/ofiwg/libfabric/blob/4cadc72f49e0648b6cab6d050edcd3ed36570e0c/fabtests/common/shared.c#L2617-L2619

a-szegel: I rely on the ft_sync at the end to let the receiver know the sender is complete.  I didn't want that check (even though it may or may not be harmful).

```
		/* Wait for writer to complete */
		ret = ft_sync();
```

fabtests/benchmarks/rma_tx_completion.c
a-szegel: Done, I made the entire test not support buffer verification (aligned in both places).

fabtests/benchmarks/benchmark_shared.c
a-szegel: The only reason I call `ft_rx_rma()` I kept trying to short-circuit that method, but I kept running into issues handling completions... so I figured it would be easiest to just stick it there.

sunkuamzn: > I kept running into issues handling completions

Likely because the next expected completion count is a global variable

Would you mind adding a comment along these lines?
> Drain fi_writedata CQEs on the receiver side

sunkuamzn (dismissed):

sunkuamzn:
@j-xiong @aingerson is the Intel CI failure related?

j-xiong:
The issue was `disk quote exceeded` which I think is unrelated.

a-szegel:
Added the comment and rebased on upstream.

fabtests/benchmarks/benchmark_shared.c
a-szegel: done

PR #11049, opened 2025-05-21
fabtests/efa: Check rdma capability for both sides

shijin-aws:
The goal of has_rdma is to skip tests on platforms that cannot use RDMA, which requires both sides have consistent capabilities. In that regard has_rdma
should check both server and client capabilities.


sunkuamzn:
bot:aws:retest

sunkuamzn:
AWS CI failure is unrelated

@j-xiong @aingerson can you please check the Intel CI?

sunkuamzn (approved):

PR #11057, opened 2025-05-22
contrib/intel/jenkins: Make stages turn red on fail

zachdworkin:
If return status is true then errors will not get propogated up to the controller to fail a build. Switching it to false will allow the stages to properly fail.


j-xiong:
yes, red should be able to turn red ;)

zachdworkin:
changes are only to intel ci, no need to let aws ci pass

PR #11059, opened 2025-05-22
[v2.1.x] contrib/intel/jenkins: Pick make stages turn red on fail from main

zachdworkin:
If return status is true then errors will not get propogated up to the controller to fail a build. Switching it to false will allow the stages to properly fail.


aingerson (approved):

zachdworkin:
no need to wait for aws ci since changes are all to intel jenkinsfile

PR #11061, opened 2025-05-22
prov/efa: Track cloned rx pkts

shijin-aws:
ooo, unexp, and readcopy rx pkts are cloned
from the original rx pkts allocated in the rx pkt pool, but they are not inserted to the rx pkt list when
they are created. This caused non-processed pkts
not cleaned up before creating the buffer pools
and cause assertion errors. This patch
fixes this issue by inserting them into
rx pkt list when they are cloned. So they
will be all cleaned during the scan of
rx_pkt_list during ep close.

### DIFF
--- a/prov/efa/test/efa_unit_tests.h
+++ b/prov/efa/test/efa_unit_tests.h
@@ -221,6 +221,8 @@ void test_efa_srx_lock();
 void test_efa_srx_unexp_pkt();
 void test_efa_rnr_queue_and_resend_msg();
 void test_efa_rnr_queue_and_resend_tagged();
+
+/* begin of efa_unit_test_ope.c */
 void test_efa_rdm_ope_prepare_to_post_send_with_no_enough_tx_pkts();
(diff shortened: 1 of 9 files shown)


prov/efa/test/efa_unit_tests.h
```diff
@@ -267,18 +270,26 @@ void test_efa_rdm_cq_post_initial_rx_pkts();
...
+/* end of efa_unit_test_rdm_peer.c */
+
+/* begin of efa_unit_test_pke.c */
 void test_efa_rdm_pke_handle_longcts_rtm_send_completion();
 void test_efa_rdm_pke_release_rx_list();
 void test_efa_rdm_pke_alloc_rta_rxe();
 void test_efa_rdm_pke_alloc_rtw_rxe();
 void test_efa_rdm_pke_alloc_rtr_rxe();
+void test_efa_rdm_pke_get_unexp();
+/* begin of efa_unit_test_pke.c */
```
sunkuamzn, line 291:
```suggestion
/* end of efa_unit_test_pke.c */
```

PR #11071, opened 2025-05-28
prov/verbs: Fix (yet another) unprotected access to vrb_util_prov.info

sydidelot:
Access to vrb_util_prov.info is now guarded by the vrb_info_mutex mutex, as the list it references may change if the application triggers an interface rescan using the FI_RESCAN flag.

This issue was discovered during internal testing at DDN with FI_RESCAN enabled.

### DIFF
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -1019,19 +1019,25 @@ int ofi_prov_check_info(const struct util_prov *util_prov,
 			uint32_t api_version,
 			const struct fi_info *user_info)
 {
-	const struct fi_info *prov_info = util_prov->info;
+	const struct fi_info *prov_info;
 	size_t success_info = 0;
 	int ret;
 
 	if (!user_info)
 		return FI_SUCCESS;
 
-	for ( ; prov_info; prov_info = prov_info->next) {
+	if (util_prov->info_lock)
+	    ofi_mutex_lock(util_prov->info_lock);
+
+	for (prov_info = util_prov->info; prov_info; prov_info = prov_info->next) {
 		ret = ofi_check_info(util_prov, prov_info,
 				     api_version, user_info);
 		if (!ret)
 			success_info++;
 	}
+
+	if (util_prov->info_lock)
+	    ofi_mutex_unlock(util_prov->info_lock);
 
 	return (!success_info ? -FI_ENODATA : FI_SUCCESS);
 }
@@ -1043,7 +1049,7 @@ int ofi_prov_check_dup_info(const struct util_prov *util_prov,
 			    const struct fi_info *user_info,
 			    struct fi_info **info)
 {
-	const struct fi_info *prov_info = util_prov->info;
+	const struct fi_info *prov_info;
 	const struct fi_provider *prov = util_prov->prov;
 	struct fi_info *fi, *tail;
 	int ret;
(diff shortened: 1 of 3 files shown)


sydidelot:
@shefty @j-xiong I would appreciate your feedback on this one.
We need to protect the call to `ofi_endpoint_init()` with `vrb_info_mutex`, since it calls `ofi_prov_check_info()`, which iterates over `vrb_util_prov.info`. The lock is only necessary around the `ofi_prov_check_info()` call; the rest of `ofi_endpoint_init()` doesn’t require locking. 
I haven’t done any performance measurements yet, but I’m concerned about how frequently `vrb_info_mutex` is now being acquired.

prov/verbs/src/verbs_ep.c
```diff
@@ -464,10 +464,11 @@ vrb_alloc_init_ep(struct fi_info *info, struct vrb_domain *domain,
 		if (vrb_alloc_wrs(ep))
 			goto err1;
 	}
-
+	ofi_mutex_lock(&vrb_info_mutex);
```
shefty, line 467: The reason for needing this lock is obscure given the code.  Is there a better location for this lock, or better way to restructure the code, so that the locking is closer to what it actually protects?

I'm not overly concerned about acquiring this lock multiple times, since it's in the control paths.  It seems like it's the price to pay for handling device removal at any point in the code.

j-xiong: The code actually accessing the protected resource (vrb_util_prov.info) is in the util code, where the lock is not always needed (the lock is only needed for verbs). 

I think adding a comment here should be sufficient.

sydidelot: I agree with you @shefty -- this lock is a bit obscure. 
We could make the util code aware of this lock by adding a pointer to it in `struct util_prov`:
```
struct util_prov {
        const struct fi_provider        *prov;
        const struct fi_info            *info;
        ofi_mutex_t                     *info_lock;  /* address of the lock to protect the info */
        ofi_alter_info_t                alter_defaults;
        const int                       flags;
};
```
Then the util code can perform the locking if `info_lock` is valid. What do you think?

shefty: I think that would be better without much complexity.

sydidelot: Thanks to both of you for your feedback! I've updated the PR with the suggested changes.

prov/util/src/util_attr.c
```diff
@@ -1019,20 +1019,26 @@ int ofi_prov_check_info(const struct util_prov *util_prov,
...
+	const struct fi_info *prov_info;
 	size_t success_info = 0;
 	int ret;
 
 	if (!user_info)
 		return FI_SUCCESS;
 
-	for ( ; prov_info; prov_info = prov_info->next) {
+	if (util_prov->info_lock)
+	    ofi_mutex_lock(util_prov->info_lock);
```
shefty, line 1030: We could make this ofi_genlock, which allows NONE as an option.

sydidelot: I considered using `ofi_genlock` as well, but it requires explicit initialization via `ofi_genlock_init()` (since there is no static constructor), and there is no function available to initialize the `struct util_prov`. As a result, each provider would need to handle genlock initialization individually.

zachdworkin:
Our CI failure is mpich 4.2.3 + mpichtestsuite 4.2.3

 Running tests in ./threads/pt2pt [00:00:00]
 Program mt_improbe_sendrecv_huge exited without No Errors
 Unexpected output in mt_improbe_sendrecv_huge:
[mpiexec@node] APPLICATION TIMED OUT, TIMEOUT = 180s

j-xiong:
bot:aws:retest

j-xiong:
bot:aws:retest

PR #11072, opened 2025-05-28
fabtests/efa: Make FI_EFA_IFACE tests check the Libfabric build

sunkuamzn:
Calling `/opt/amazon/efa/bin/fi_info` checks the fi_info from the EFA
installer. This fabtest is supposed to check the getinfo path in the
current Libfabric build. This commit checks the current build by
printing the output from fi_efa_info_test

### DIFF
--- a/fabtests/prov/efa/src/efa_info_test.c
+++ b/fabtests/prov/efa/src/efa_info_test.c
@@ -35,6 +35,22 @@
 #include <getopt.h>
 #include <shared.h>
 
+static int print_short_info(struct fi_info *info)
+{
+	struct fi_info *cur;
+
+	for (cur = info; cur; cur = cur->next) {
+		printf("provider: %s\n", cur->fabric_attr->prov_name);
+		printf("    fabric: %s\n", cur->fabric_attr->name),
+		printf("    domain: %s\n", cur->domain_attr->name),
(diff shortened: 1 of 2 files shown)


fabtests/prov/efa/src/efa_info_test.c
```diff
@@ -37,13 +37,47 @@
...
 	hints = fi_allocinfo();
 	if (!hints)
 		return EXIT_FAILURE;
-	hints->fabric_attr->prov_name=strdup("efa");
+
+
+	while ((op = getopt(argc, argv, "f:p:")) != -1) {
+		switch (op) {
+		case 'f':
+			if (!hints->fabric_attr) {
```
shijin-aws, line 67: I thought `fi_allocinfo()` already allocated `fabric_attr`, see https://github.com/ofiwg/libfabric/blob/main/src/fabric.c#L1425-L1446

shijin-aws: Yes, I think the fabtests code were written to handle the case that application alloc hints manually via malloc. OFI 2.0 requires that all hints must be allocated via `fi_dupinfo(NULL)` (which is the same as `fi_allocinfo`).

sunkuamzn:
@aingerson no problem, thanks for checking!

PR #11082, opened 2025-05-30
prov/efa: post initial rx pkt when qp is enabled.

shijin-aws:
Currently, cq and cntr read don't check the qp status of the bound ep before posting the initial rx pkts. This is a bug and will cause a crash when a cq/cntr read is called before the bound ep is enabled.


shijin-aws:
bot:aws:retest

sunkuamzn (approved):

PR #11083, opened 2025-05-30
prov/shm: don't run smr progress if region isn't initialized

aingerson:
If fi_cq_read is called before the EP has been enabled, the region has not been initialized and will segfault. Make sure the region is valid before progressing commands


shijin-aws (approved):

zachdworkin (approved):

shijin-aws:
bot:aws:retest

PR #11084, opened 2025-05-31
prov/rxm: Fix potential data error in the rndv_write protocol

j-xiong:
The rndv write protocol uses a state in the tx_buf to track the send side progress and determine the next step, as shown below:

	(1) send(req):
		state --> RXM_RDNV_TX

	(2) send complete:
		state --> RXM_RNDV_WRITE_DATA_WAIT

	(3) received(resp):
		state --> RXM_RNDV_WRITE
		write(data)

	(4) write complete:
		state --> RXM_RNDV_WRITE_DONE_SENT
		send(done)

	(5) send complete:
		done

This works fine in this order. However, there is the possibility that step (2) and (3) happen in reverse order, i.e., the response is handled before the completion of the first send. In such case, the send completion would be misinterpreted as the write completion in step (4) since the operation is determined by the state, which is RXM_RNDV_WRITE set by step (3). This can cause a "done" message being sent prematurelly. As a result, the data may have not fully arrived at the receive buffer when the receive operation completes.

Fix by introducing a new state RXM_RNDV_WRITE_TX_WAIT, which expects an extra send complete before sending a "done" message.

**Update:**

Two more fixed are added. They fix different issues in the rndv_write protocol. All these issues can potentially lead to premature completion at the receive side and thus data error when the receive buffers are used when the data are not fully ready.


j-xiong:
bot:aws:retest

ooststep (dismissed):

shijin-aws:
AWS CI failure was an infra issue and can be ignored

zachdworkin (dismissed):

j-xiong:
@shijin-aws Could you check the AWS CI failure details?

Update: never mind, I assume the situation is the same as when you checked before.

j-xiong:
@ooststep @zachdworkin I need review from either of you in order to merge.

zachdworkin (approved):

PR #11087, opened 2025-06-02
fabtests: fix leak of dmabuf fd

jiaxiyan:
All calls to ft_hmem_get_dmabuf_fd need a corresponding close call. Or the dmabuf will stick around for the lifetime of the process, even after dereg and releasing the memory back to the device mempool.
similar to https://github.com/ofiwg/libfabric/pull/10708

### DIFF
--- a/fabtests/benchmarks/rdm_bw_mt.c
+++ b/fabtests/benchmarks/rdm_bw_mt.c
@@ -295,6 +295,8 @@ static int mt_reg_mr(struct fi_info *fi, void *buf, size_t size,
 	ft_fill_mr_attr(&iov, &dmabuf, 1, access, key, iface, device, &attr,
 			flags);
 	ret = fi_mr_regattr(dom, &attr, flags, mr);
+	if (opts.options & FT_OPT_REG_DMABUF_MR)
+		ft_hmem_put_dmabuf_fd(iface, dmabuf_fd);
 	if (ret)
 		return ret;
 
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -474,6 +474,8 @@ int ft_reg_mr(struct fi_info *fi, void *buf, size_t size, uint64_t access,
 
 	ft_fill_mr_attr(&iov, &dmabuf, 1, access, key, iface, device, &attr, flags);
 	ret = fi_mr_regattr(domain, &attr, flags, mr);
+	if (opts.options & FT_OPT_REG_DMABUF_MR)
+		ft_hmem_put_dmabuf_fd(iface, dmabuf_fd);
 	if (ret)
 		return ret;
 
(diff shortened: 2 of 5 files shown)


fabtests/common/shared.c
```diff
@@ -474,6 +474,8 @@ int ft_reg_mr(struct fi_info *fi, void *buf, size_t size, uint64_t access,
 
 	ft_fill_mr_attr(&iov, &dmabuf, 1, access, key, iface, device, &attr, flags);
 	ret = fi_mr_regattr(domain, &attr, flags, mr);
+	if (opts.options & FT_OPT_REG_DMABUF_MR)
+		(void)close(dmabuf_fd);
```
shijin-aws: I think you need to introduce a general op function to handle the fd close, like https://github.com/ofiwg/libfabric/commit/5357ceecd708d7eedd72e2df292a7a6b55a9df58. It is possible that different hmem iface has different API to make it.

fabtests/benchmarks/rdm_bw_mt.c
```diff
@@ -295,6 +295,8 @@ static int mt_reg_mr(struct fi_info *fi, void *buf, size_t size,
 	ft_fill_mr_attr(&iov, &dmabuf, 1, access, key, iface, device, &attr,
 			flags);
 	ret = fi_mr_regattr(dom, &attr, flags, mr);
+	if (opts.options & FT_OPT_REG_DMABUF_MR)
+		ft_hmem_put_dmabuf_fd(iface, dmabuf_fd);
```
shijin-aws, line 299: I am not sure whether we can close the fd before we are done with the MR. I think we should move this close before the mr close.

jiaxiyan: I copied the same logic from https://github.com/ofiwg/libfabric/pull/10708

shijin-aws: They are different, we close the fd in domain open right after the ibv_reg_dmabuf_mr because we are not using this MR for any real data transfer. I am not sure whether we can do the same thing for fabtests which use the MR for data transfer later. Maybe it's ok, I am just not sure.

shijin-aws: cc @mrgolin

mrgolin: I think this is fine as long you no longer need it in userspace.

sunkuamzn:
bot:aws:retest

fabtests/benchmarks/rdm_bw_mt.c
shijin-aws: Thanks for clarification

shijin-aws (approved):

j-xiong:
@shijin-aws Is the aws ci failure related?

jiaxiyan:
@j-xiong No, we have some testing infrastructure issues

jiaxiyan:
bot:aws:retest

jiaxiyan:
bot:aws:retest

PR #11090, opened 2025-06-02
fabtests/efa: Introduce multi_ep_mt test

shijin-aws:
Introduce a new multi-ep multi-threaded test for efa provider. The client creates N+1 (specified by -c) threads. The first N threads have their own EP/AV created, do
sends with given number of iterations (-I)
and size (-S), and destroy EP/AV without
waiting for completion. Each thread can join and exit in randome timeline, realized
by the random nanosleep at the beginning
and the end of the thread function.

The last (N+1) thread polls a persistent
CQ that are bound to the EPs in the first
N threads, in the given timeout (-T)

The server has 1 thread that posts recv
and posts CQ, in the given timeout (-T)


shijin-aws:
Opening for CI purpose only today.

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -0,0 +1,572 @@
...
+				    (void **) &recv_bufs[i],
+				    alloc_size);
+		if (ret)
+			return ret;
+	}
+
+	return 0;
+}
+
+static int ep_post_rx()
```
sunkuamzn: It might be better to re-use `FT_POST` for `ep_post_rx` and `ep_post_tx` so that they both have timeouts. Otherwise, the test can hang here until the CI times out

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -0,0 +1,572 @@
...
+			      remote_fiaddr[idx], &send_ctx[idx]);
+		if (ret == -FI_EAGAIN)
+			(void) fi_cq_read(txcq, NULL, 0);
+
+	} while (ret == -FI_EAGAIN);
+
+	return ret;
+}
+
+static int get_one_comp(struct fid_cq *cq)
```
sunkuamzn, line 162: Similar comment. I see you have timeouts in the functions that call `get_one_comp` but I think you can reuse `ft_spin_for_comp`

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -0,0 +1,572 @@
...
+	return FI_SUCCESS;
+}
+
+static void *post_sends(void *context)
+{
+	int idx, ret, i;
+	size_t len;
+	// the range of the sleep time (in nanoseconds)
+	int min = 0;
+	int max = 999999999;
```
sunkuamzn: Could you make these macros at the top of the file?

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -0,0 +1,572 @@
...
+		if (ret)
+			printf("Client: thread %d post_sends create failed: "
+			       "%d\n",
+			       i, ret);
+	}
+
+	pthread_join(context_cq.thread, NULL);
+
+	for (i = 0; i < num_eps; i++)
+		pthread_join(contexts_ep[i].thread, NULL);
```
sunkuamzn:
```suggestion
	for (i = 0; i < num_eps; i++)
		pthread_join(contexts_ep[i].thread, NULL);
		
	pthread_join(context_cq.thread, NULL);
```

Shouldn't the send threads finish before the CQ thread?

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -0,0 +1,572 @@
...
+
+	ret = ft_getinfo(hints, &fi);
+	if (ret)
+		return ret;
+
+	ret = ft_open_fabric_res();
+	if (ret)
+		return ret;
+
+	ret = ft_alloc_active_res(fi);
```
sunkuamzn, line 441: I see that you're using the `ep`, `rx_buf` etc maintained by the common code on the server side. That's fine when the server is a single thread always just receiving but it won't work when you want multiple endpoints on both the server and the client. Can be fixed later.

fabtests/pytest/efa/test_multi_ep.py
```diff
@@ -1,15 +1,24 @@
...
     if shared_av:
         cmd += " -A"
     test = ClientServerTest(cmdline_args, cmd, message_size=256, fabric=rma_fabric)
     test.run()
+
+
+# TODO: expand to different meory types
+@pytest.mark.functional
+@pytest.mark.parametrize("msg_size", [16, 8192, 16384, 1048576])
+def test_multi_ep_mt(cmdline_args, msg_size):
```
sunkuamzn, line 21: Prefer to merge it first without adding to fabtests

fabtests/prov/efa/src/multi_ep_mt.c
shijin-aws: Yeah, the current test just has 1 persistent receiver. I will change it in a future work

fabtests/prov/efa/src/multi_ep_mt.c
shijin-aws: They are concurrent anyway. I can modify the sequence

fabtests/prov/efa/src/multi_ep_mt.c
shijin-aws: The problem with `ft_spin_for_comp` is that it will return immediately after getting cq error, which is not what I want. I'd prefer to keep using `get_one_comp` for flexibility

fabtests/prov/efa/src/multi_ep_mt.c
shijin-aws: Done

fabtests/prov/efa/src/multi_ep_mt.c
shijin-aws: I will make it use `ft_post_tx/rx_buf` which use `FT_POST`. But if I read the code correctly `timeout` in those functions are actually not useful, am I wrong? 

`FT_POST` https://github.com/ofiwg/libfabric/blob/main/fabtests/common/shared.c#L2196-L2222 uses `ft_progress` as the `progress_fn`, but ft_progress did not finally use the `timeout` or `timeout_save` internally https://github.com/ofiwg/libfabric/blob/main/fabtests/common/shared.c#L2175-L2194

fabtests/prov/efa/src/multi_ep_mt.c
shijin-aws: Done

fabtests/pytest/efa/test_multi_ep.py
shijin-aws: Let me run it in CI for more times. Currently it is passing on my local setup

shijin-aws: I made it as a separate commit, I will drop it if it is still not stable after several runs.

shijin-aws: OK, now I make this test with a marker `unstable` and will not be run by default. It can be run explicitly by using `runfabtests.py .... -t unstable`. Please see the latest revision

shijin-aws:
bot:aws:retest

shijin-aws:
@j-xiong is Intel CI hanging? The change is efa only but I suspect Intel CI may compile the changes so I want to ask

fabtests/prov/efa/src/multi_ep_mt.c
sunkuamzn: Since you have timeouts on functions calling `get_one_comp`, I'm OK

fabtests/prov/efa/src/multi_ep_mt.c
sunkuamzn: You're right. That's probably unintentional.

j-xiong:
@shijin-aws There was a HW issue with the CI head node. It's being worked on. At this moment CI runs will just keep waiting.

PR #11105, opened 2025-06-06
prov/shm: acquire ep lock when freeing entries

### LINKED ISSUE #11098: prov/shm: 🐛 Assertion failure on `ofi_genlock_held(srx->lock)` during `fi_close` with shm provider
**Describe the bug**
My application occasionally crashes nondeterministically during teardown, with an assertion failure inside util_free_entry():
```
Assertion `ofi_genlock_held(srx->lock)` failed.
```
Unfortunately, I cannot share a minimal reproducible example, as the crash occurs randomly and is part of a larger multi-threaded and asynchronous application.

**To Reproduce**
Due to the nondeterministic nature of the bug, I cannot provide minimal reproduction steps. However, it happens during shutdown/teardown, after fi_close() is called on an endpoint.

**Expected behavior**
The library should either handle teardown gracefully or document expectations on locking. This assertion should never fail in correct user code. I would appreciate help understanding where the `srx->lock` is expected to be held at the moment util_free_entry() is called.

**Output**
Backtrace:
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=139932185970240) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=139932185970240) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=139932185970240, signo=6) at ./nptl/pthread_kill.c:89
#3  __GI_raise (sig=6) at ../sysdeps/posix/raise.c:26
#4  __GI_abort () at ./stdlib/abort.c:79
#5  __assert_fail_base (...) at ./assert/assert.c:94
#6  __GI___assert_fail ("ofi_genlock_held(srx->lock)", "prov/util/src/util_srx.c", 385, "util_free_entry") at ./assert/assert.c:103
#7  util_free_entry (entry=0x7f44700d3000) at prov/util/src/util_srx.c:385
#8  smr_ep_close (fid=0x7f44700608e0) at prov/shm/src/smr_ep.c:821
#9  fi_close (fid=0x7f44700608e0) at rdma/fabric.h:642
```

fi_info output:
```
caps: [ FI_MSG, FI_RMA, FI_TAGGED, FI_READ, FI_WRITE, FI_RECV, FI_SEND, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_LOCAL_COMM, FI_RMA_EVENT, FI_SOURCE, FI_DIRECTED_RECV ]
mode: [  ]
addr_format: FI_ADDR_STR
src_addr: fi_shm://75350
fi_tx_attr:
  caps: [ FI_MSG, FI_RMA, FI_TAGGED, FI_READ, FI_WRITE, FI_SEND ]
  op_flags: [ FI_DELIVERY_COMPLETE ]
  inject_size: 4096
  size: 1024
fi_rx_attr:
  caps: [ FI_MSG, FI_RMA, FI_TAGGED, FI_RECV, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_RMA_EVENT, FI_SOURCE, FI_DIRECTED_RECV ]
  msg_order: [ FI_ORDER_RAR, FI_ORDER_RAW, ..., FI_ORDER_SAS, ... ]
fi_ep_attr:
  type: FI_EP_RDM
  protocol: FI_PROTO_SHM
  max_msg_size: 18446744073709551615
fi_domain_attr:
  name: shm
  threading: FI_THREAD_SAFE
  progress: FI_PROGRESS_MANUAL
  mr_mode: [ FI_MR_VIRT_ADDR ]
fi_fabric_attr:
  name: shm
  prov_name: shm
  prov_version: 201.0
  api_version: 2.1
```

**Environment:**

- OS: Ubuntu 22.04
- Provider: shm
- Endpoint type: FI_EP_RDM
- libfabric version: 2.1.0

**Additional context**
The issue occurs nondeterministically in a multithreaded environment.

I suspect the assertion fails because srx->lock is not properly acquired before util_free_entry() is called, but I'm not sure where this is expected to happen.

If there's any way to validate the correct locking pattern, or if you have suggestions for debugging this further, I’d really appreciate it.

Is there any additional information I could provide to help diagnose this issue?

Thanks in advance for your support.

This issue started occurring after the following commit:

https://github.com/ofiwg/libfabric/commit/164204ef67fe4e12619d748b9c797e86e9ea4cbc

@aingerson

aingerson:
When closing the shm ep, we have to free entries with the owner srx if there are any outstanding unexpected messages. When shm is the owner, that means we have to lock the srx (uses the ep lock) to avoid lock debug errors. When used as a peer it is expected that the owner lock the srx a peer EP is bound to when closing the endpoint

Fixes #11098

### DIFF
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -807,6 +807,7 @@ static int smr_ep_close(struct fid *fid)
 		smr_free_sock_info(ep);
 	}
 
+	ofi_genlock_lock(&ep->util_ep.lock);
 	while (!dlist_empty(&ep->sar_list)) {
 		dlist_pop_front(&ep->sar_list, struct smr_pend_entry, pend,
 				entry);
@@ -821,6 +822,7 @@ static int smr_ep_close(struct fid *fid)
 		ep->srx->owner_ops->free_entry(cmd_ctx->rx_entry);
 		ofi_buf_free(cmd_ctx);
 	}
+	ofi_genlock_unlock(&ep->util_ep.lock);
 
 	if (ep->srx) {
 		/* shm is an owner provider */


j-xiong (commented):
When shm is the owner, the ep lock is the same as the sex lock. When shm is not the owner, the locking here doesn't help.

What exactly is the issue with locking the srx lock directly? It should be able to do that even it's not the owner.

shijin-aws:
why cannot shm just acquire the owner lock? It's violating the peer API ?

j-xiong:
ok, the issue is that there is no reference from `srx` to the lock, so we can't access the srx lock from the peer.

aingerson:
When shm is not the owner, shm is totally unaware of what the srx is doing so it can't lock the srx directly. It needs the owner to acquire whatever lock it relies on before closing the EP. So when the owner is not shm, the lock acquire is outside of shm and this addition doesn't protect against anything.
When shm is the owner, it defines which lock to use for the srx which it decides it the ep lock.
So this fix is only for the case where shm is the owner. It doesn't solve the case for when efa is the owner for example. Any owner using a peer will have to lock the srx lock before closing the peer EP that it is bound to

aingerson:
@shijin-aws I think is handling the efa case and @amirshehataornl lnx will likely need one as well

j-xiong:
should we add `lock` and `unlock` to the `owner_ops` of fid_peer_srx? that may open a can of new worm.

aingerson:
yeah I don't know, I think it will get complicated... I think this is ok for now and we can revisit after the release?

shijin-aws:
OK,  I was planning to lock the whole efa ep close under the srx lock (except for the srx context close). Today we grab the same lock inside efa's fi_cq_read. I think that should fix this issue

j-xiong (approved):

shijin-aws:
FWIW, I fixed the missing lock issue in efa provider as part of https://github.com/ofiwg/libfabric/pull/11151

PR #11112, opened 2025-06-10
man/fi_mr: Clarify fi_close behavior

sunkuamzn:


man/man3/fi_mr.3
```diff
@@ -503,18 +503,23 @@ fi_mr_attr (defined below).
...
 error at the peer.
 The state of the local endpoint will be unaffected.
 .PP
-When closing the MR, there must be no opened endpoints or counters
-associated with the MR.
+For MRs that are associated with an endpoint (FI_MR_ENDPOINT flag
+is set), the MR must be closed before the endpoint.
+For MRs that are associated with the domain (FI_MR_ENDPOINT flag
+is not set), all opened endpoints or counters associated with the
+MR must be closed before closing the MR.
```
shefty: For MRs associated with a domain, there's not a direct way to associate an EP with the MR.  So, I would remove the reference to EP.

For counters, the flow is that fi_mr_bind() is used to associate the MR to a counter.  Since the binding is MR -> counter, the reference is also MR -> counter, which would make destruction follow the same order: MR before counter.  That is, a counter may be referenced by multiple MRs.

You can think of it as how the objects must be created.  The counter is created prior to creating and configuring the MR.  (Think of fi_mr_reg + fi_mr_bind as a combined create+init step.)

sunkuamzn: OK that makes sense

In that case, should we just remove this entire line?

> For MRs that are associated with the domain (FI_MR_ENDPOINT flag
is not set), all opened endpoints or counters associated with the
MR must be closed before closing the MR.

shefty: Removing the entire line is probably best.  It makes more sense to have those statements in the counter and EP pages.  (They may already be there, and if not, could be added separately.)

man/man3/fi_mr.3
```diff
@@ -503,18 +503,23 @@ fi_mr_attr (defined below).
...
 Fi_close is used to release all resources associated with a registering
 a memory region.
-Once unregistered, further access to the registered memory is not
+Once deregistered, further access to the registered memory is not
 guaranteed.
 Active or queued operations that reference a memory region being closed
-may fail or result in accesses to invalid memory.
+may attempt to access an invalid memory region and fail.
+After an MR is closed, any new operations targeting the closed MR will
+also fail.
```
soumagne, line 511: @shefty is this really true / guaranteed ? default behavior with verbs / cxi is that operations targeting MRs that remain exposed in the MR cache may just succeed.

shefty: The MR cache is a separate problem that breaks the security model.  I'd rather any impact of the cache be kept to a cache specific discussion.

soumagne: Sure but if we're adding that line I would then expect providers have to guarantee that, though those semantics are currently usually broken because of MR caching.

bwbarrett: This is the behavior that customers need.  I'm surprised that you're putting explicitly created MRs in the registration cache, for exactly the security reason.  Implicitly created MRs makes complete sense to put in the cache, but I don't understand how you can have a transparent cache and allow customers their security model.

shefty: MR caching is always used by MPI and CCL.  And, really for a parallel app, it's likely a non-issue.  The cache is controllable through environment variables and is described further down in this man page.  Providers are free to enable/disable caching by default based on their whims or customer needs.  It looks like EFA also defaults to caching enabled.

```
#define EFA_DEF_MR_CACHE_ENABLE 1
int efa_mr_cache_enable	= EFA_DEF_MR_CACHE_ENABLE;
```
Storage probably never wants caching enabled.  Even hidden registration (e.g. when converting tag matching to rendezvous) isn't safe.

When layering providers, the underlying provider will see everything as an explicit registration.  There's an internal flag which can be used to disable caching.

bwbarrett: Actually, for the OFI NCCL Plugin, we tried pretty hard not to let Libfabric have things in a cache.  And more than one group is messing with fault tolerance under NCCL, which really assumes killing an MR kills communication.  Hence the path that lead us to these questions.

I'm apparently going to have to go discuss things with the EFA team again.  What is supposed to happen is that MRs created implicitly through send/recv (the MPI case) go through the cache, but that when a customer calls fi_close() on an MR, that MR is actually closed in the NIC.  If that's not happening, that's a bug in EFA that will be fixed shortly.

soumagne: I agree this is also what I would want  as a user so I can rely on those semantics but this has not been the default so if this is the new default, providers should update their code accordingly at the detriment of performance or maybe there should be a note in that paragraph that explains why this may not be guaranteed for performance reasons.

soumagne: > Storage probably never wants caching enabled.

@shefty This is not quite true. On storage servers yes we can easily preallocate MRs and have all memory allocated/registered beforehand but on clients, because typically client applications / middleware just pass down their own allocated buffers we have to register them on the fly (and we can't memcpy).

man/man3/fi_mr.3
sunkuamzn: @shefty I removed that line. Should the following line should also be removed?

> If resources are still associated with the MR when attempting to close,
the call will return -FI_EBUSY.

If the destruction order is MR followed by the counter, what other resources could be associated with the MR when attempting to close it?

man/man3/fi_mr.3
sunkuamzn: @bwbarrett I fixed EFA provider's behavior in https://github.com/ofiwg/libfabric/pull/11122

@soumagne do you have a preferred wording?

man/man3/fi_mr.3
shefty: A MR can be referenced as a 'base MR' by a sub-MR. I would leave that line.  There may be other cases where a reference is taken, but I can't think of any.

man/man3/fi_mr.3
shefty: I would use the MR cache section to describe any behavior changes to the API.  The most I would add here is to see that section.  The cache section does not clearly indicate how it impacts API behavior or really the pitfalls of using it.

soumagne: I like the current wording that you are proposing. The only concern I was raising was that some providers are currently not conforming to this behavior and that can be misleading for application developers. So my question was more, if we are adding that line, are we expecting providers to be updated to reflect that ? If not then there should be an additional note that warns developers.

sunkuamzn: @shefty I added a line referencing the memory registration cache section

@soumagne when the EFA provider does not meet specifications in the man pages, we list them in the Limitations section here https://ofiwg.github.io/libfabric/main/man/fi_efa.7.html

I'm not sure if that's the best solution here. If you want to add such a note to either this section or the MR cache section, would you like to suggest the wording?

@j-xiong do you have any thoughts on how to best handle this?

shefty (approved):
This is an improvement over what's there.

soumagne (approved):
The note that you added seems fine also with me. Thanks!

PR #11113, opened 2025-06-10
prov/efa: Move peer bufpool to endpoint 

sunkuamzn:
The peers are specific to each endpoint, so it's better for the peer
bufpool to also be associated with the endpoint

The initial size of the peer bufpool is defined by
EFA_RDM_EP_MIN_PEER_POOL_SIZE which is set to 1024


prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -102,25 +89,25 @@ struct efa_rdm_peer *efa_rdm_ep_get_peer(struct efa_rdm_ep *ep, fi_addr_t addr)
...
-		if (!peer) {
-			EFA_WARN(FI_LOG_EP_DATA, "Failed to allocate peer struct\n");
-			return NULL;
-		}
+	map_entry->addr = addr;
 
-		efa_rdm_peer_construct(peer, ep, &av_entry->conn);
+	conn = efa_av_addr_to_conn(ep->base_ep.av, addr);
+	if (conn->ep_addr) {
+		efa_rdm_peer_construct(&map_entry->peer, ep, conn);
```
shijin-aws: Should you move the `map_entry` alloc inside this block?

shijin-aws: Or even better, make it part of the map insertion call

sunkuamzn: I removed the check altogether

PR #11115, opened 2025-06-10
prov/verbs: Avoid holding vrb_info_mutex when reloading interfaces

sydidelot:
Problem:
Currently, interfaces are reloaded while holding `vrb_info_mutex`.
Since reloading interfaces can take a significant amount of time,
this may block other threads attempting to acquire `vrb_info_mutex`,
leading to potential performance issues.

Solution:
This patch addresses the issue by moving the call to `vrb_init_info()`
(which reloads the list of interfaces) outside the critical section
protected by `vrb_info_mutex`. Once the new interface lists are prepared,
the lock is briefly acquired to safely swap `vrb_devs` and `vrb_util_prov.info`.

### DIFF
--- a/prov/verbs/src/verbs_init.c
+++ b/prov/verbs/src/verbs_init.c
@@ -87,6 +87,7 @@ struct fi_provider vrb_prov = {
 	.cleanup = vrb_fini
 };
 
+/* mutex for guarding the initialization of vrb_util_prov.info */
 ofi_mutex_t vrb_info_mutex;
 
 struct util_prov vrb_util_prov = {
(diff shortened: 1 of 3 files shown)


prov/verbs/src/verbs_init.c
```diff
@@ -98,7 +99,8 @@ struct util_prov vrb_util_prov = {
 	.flags = 0,
 };
 
-/* mutex for guarding the initialization of vrb_util_prov.info */
+/* mutex for guarding concurrent calls to protect provider initialization */
```
shefty, line 102: Please add a comment that the lock ordering is init_mutex then info_mutex.

PR #11121, opened 2025-06-12
prov/cxi: Adjust cxi environment variable defaults

swelch:
Adjust many of the cxi environment variables to default to values commonly set by middle-ware. This should cover most applications by default that do not exhaust hardware resources.


heemstra (dismissed):

swelch:
I will refresh with unit required unit test change.

heemstra (approved):

jswaro (approved):

PR #11122, opened 2025-06-12
prov/efa: Do not add explicit MRs from application to MR cache

sunkuamzn:
This commit ensures that the MR cache is only used for internal MRs registered by the EFA provider. The domain's fi_ops_mr is always set to never use the MR cache. The fi_ops_mr with the cache is removed.

This commit adds a unit test that verifies that the MR cache is not used for explicit MRs when FI_MR_LOCAL is unset and a fabtest to run fi_rdm_pingpong without FI_MR_LOCAL

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -349,10 +349,25 @@ static int efa_mr_cache_regattr(struct fid *fid, const struct fi_mr_attr *attr,
 	return 0;
 }
 
-static int efa_mr_cache_regv(struct fid *fid, const struct iovec *iov,
-			     size_t count, uint64_t access, uint64_t offset,
-			     uint64_t requested_key, uint64_t flags,
-			     struct fid_mr **mr_fid, void *context)
+/**
+ * @brief Register a memory region and add it to the MR cache
+ *
+ * This function is only used internally by the EFA provider to register
+ * a memory region containing application memory and add it to the MR cache
+ * It is only used when FI_MR_LOCAL is not set and the MR cache is enabled
+ *
+ * @param[in]	fid		domain fid
+ * @param[in]	iov		iovec to be registered
+ * @param[in]	access		access flags for MR
+ * @param[in]	flags		MR flags
+ * @param[out]	mr_fid		MR descriptor fids
+ *
+ * @return FI_SUCCESS or negative FI error code
+ */
+int efa_mr_cache_regv(struct fid_domain *domain_fid, const struct iovec *iov,
+		      size_t count, uint64_t access, uint64_t offset,
+		      uint64_t requested_key, uint64_t flags,
+		      struct fid_mr **mr, void *context)
 {
 	struct fi_mr_attr attr = {0};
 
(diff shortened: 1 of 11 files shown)


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -301,10 +302,16 @@ void efa_rdm_ope_try_fill_desc(struct efa_rdm_ope *ope, int mr_iov_start, uint64
 				 "buf: %p len: %ld access: %#lx\n",
 				 ope->iov[i].iov_base, ope->iov[i].iov_len, access);
 
-		err = fi_mr_regv(
-			&efa_rdm_ep_domain(ope->ep)->util_domain.domain_fid,
-			ope->iov + i, 1, access, 0, 0, 0, &ope->mr[i],
-			NULL);
+		domain = efa_rdm_ep_domain(ope->ep);
+		if (OFI_LIKELY(efa_mr_cache_enable)) {
```
shijin-aws: How about we define a function pointer at the domain open for the internal mr_reg call, so you don't have to check this every time

shijin-aws: Also, I think there are other places that we call `fi_mr_reg` internally, see https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L37

shijin-aws: Also this mr close can take the mr cache path: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ope.c#L125. Now your change makes `fi_close` always non-cache, which introduce an inconsistency here.

sunkuamzn: Thanks, I will fix

sunkuamzn: The `fi_close` is not inconsistent because in `efa_mr_cache_entry_reg` (which is eventually called by `efa_mr_cache_regattr` through `ofi_mr_cache_search` when adding a new entry to cache), the `close` function pointer is set to `efa_mr_cache_close`

https://github.com/ofiwg/libfabric/blob/184cdde87aa725d35aedc168aeaa8cdf08d4933f/prov/efa/src/efa_mr.c#L266
https://github.com/ofiwg/libfabric/blob/184cdde87aa725d35aedc168aeaa8cdf08d4933f/prov/efa/src/efa_mr.c#L133
https://github.com/ofiwg/libfabric/blob/184cdde87aa725d35aedc168aeaa8cdf08d4933f/prov/util/src/util_mr_cache.c#L293

shijin-aws: Oh, my point was that the old behavior is using cache to close when it's available. But now your `fi_close` will always be non-cache.

sunkuamzn: Talked offline and addressed all comments

I added a function pointer to the domain

We agreed that putting the packet pool MRs in the cache is not useful. I added a comment to reflect that.

sunkuamzn:
@j-xiong would like to get your approval on the fabtests commit

j-xiong (commented):
fabtests changes look good to me.

prov/efa/src/efa_mr.c
```diff
@@ -349,45 +349,40 @@ static int efa_mr_cache_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+int efa_mr_cache_regv(struct fid_domain *domain_fid, const struct iovec *iov,
+		      size_t count, uint64_t access, uint64_t offset,
+		      uint64_t requested_key, uint64_t flags,
+		      struct fid_mr **mr, void *context)
 {
 	struct fi_mr_attr attr = {0};
 
 	attr.mr_iov = iov;
-	attr.iov_count = count;
+	attr.iov_count = 1;
```
shijin-aws: Is this correct? why overwrite the count and context?

prov/efa/src/efa_mr.c
```diff
@@ -349,45 +349,40 @@ static int efa_mr_cache_regattr(struct fid *fid, const struct fi_mr_attr *attr,
...
+ *
+ * @param[in]	fid		domain fid
+ * @param[in]	iov		iovec to be registered
+ * @param[in]	access		access flags for MR
+ * @param[in]	flags		MR flags
+ * @param[out]	mr_fid		MR descriptor fids
+ *
+ * @return FI_SUCCESS or negative FI error code
+ */
+int efa_mr_cache_regv(struct fid_domain *domain_fid, const struct iovec *iov,
```
shijin-aws, line 367: why do you need this header change?

shijin-aws: Nvm, it's just an `fi_*` and mr ops interface inconsistency

prov/efa/src/efa_mr.c
sunkuamzn: Fixed, thanks!

PR #11123, opened 2025-06-12
prov/cxi: Fix regression which could cause FI_THREAD_SAFE deadlock

swelch:
The following community upstream commit: fb928f48f27dc1591781101aa55d7cb8b026b693 locking changes opens a window for FI_THREAD_SAFE locking to deadlock should portals flow control occur and a receive buffer is posted. Fix buy progressing without taking the lock now held.


swelch:
This will need to cherry-picked into 2.2.x as well.

heemstra (approved):

jswaro (approved):

PR #11133, opened 2025-06-13
src/Makefile: Add version and commit startup log

a-szegel:
This commit modifies the build system, as well as src/fabric.c and src/fi_version_gen.h.in to create a info log which outputs the version of libfabric that is running as well as the commit used to build.  The log message looks like:

libfabric:906569:1749843568::core:core:fi_ini():1014<info> Initialized \ libfabric 2.2.0rc1 - built from commit 7ef321d3

This log message will be helpful when trying to prove which version/commit of libfabric is being run. This is helpful when there are multiple versions installed on the system, and application startup is non-trivial (docker/enroot/slurm/etc...).

The src/fi_version_gen.h file gets generated on every make command (including make dist) if there is a .git folder in the source directory. The file is compiled into the library so it works for both .a/.so libs. This has implications for the development workflow and distribution users. When developers create a new commit, and rebuild it, the log message will be updated. When developers bump the libfabric version in configure.ac and configure/make, the log message will be updated. When a developer wishes to release, and they call ./autogen.sh && ./configure && make dist, they will get a src distribution tarball.  They can then go into that tarball and call configure/make and they will always get the version/commit that was used to make the tarball.

When the user calls make clean in a git repo src dir, the generated src/fi_version_gen.h gets deleted.  When the user calls make clean in a dist src dir (no .git), the src/fi_version_gen.h file does not get deleted.  This is required because the project will not build without src/fi_version_gen.h, and in a dist src dir, the file does not get generated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,8 +61,11 @@ config/test-driver
 util/fi_info
 util/fi_strerror
 util/fi_pingpong
+util/fi_mon_sampler
 
 prov/*/*.spec
+
+src/fi_version_gen.h
 
 .vs
 fabtests.spec
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,6 +30,33 @@ pkglib_LTLIBRARIES = $(DL_PROVIDERS)
 ACLOCAL_AMFLAGS = -I config
 AM_CFLAGS = -Wall
 
+if NEED_VERSION_GEN
+BUILT_SOURCES = src/fi_version_gen.h
+# Force regeneration every time
+.PHONY: FORCE
+FORCE:
+
+src/fi_version_gen.h: $(srcdir)/src/fi_version_gen.h.in FORCE
+	@echo "Generating fi_version_gen.h..."
+	@{ \
+		COMMIT_HASH=$$(cd $(srcdir) && git rev-parse HEAD 2>/dev/null || echo "unknown"); \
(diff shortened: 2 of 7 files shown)


j-xiong:
Both windows and linux builds are complaining about `fi_version_gen.h bnot found` when compiling `src/fabric.c`.

For windows, the file needs to be added to the project files.

For linux, the case seems to be trying to build from a git tree on a machine without git command. It's totally valid, especially on clusters with shared file system that git command only exists on the head node while the build is done on a compute node. This scenario needs to be supported.

j-xiong:
Intel CI is still failing:
```
src/fabric.c:56:10: fatal error: fi_version_gen.h: No such file or directory
   56 | #include "fi_version_gen.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.

```

j-xiong:
Not sure what exactly went wrong, but it would help if `fi_version_gen.h` is aways generated if it doesn't exist at all.

a-szegel:
Is this windows?

j-xiong:
No, it's Linux.

a-szegel:
I will add a clause that if the file does not exist, to generate it in order to have a successful build.

a-szegel:
meh broke it somehow... will look when I get time

PR #11143, opened 2025-06-17
log: Fix buffer overrun when accessing the 'log_levels' array

### LINKED ISSUE #11116: prov/xxx: fi_convert_log_str out-of-bounds error
**Describe the bug**

Stop condition of the loop from fi_convert_log_str (https://github.com/ofiwg/libfabric/blob/main/src/log.c#L103) works by checking the value stored in log_levels array (https://github.com/ofiwg/libfabric/blob/main/src/log.c#L67) Null end marker was removed by https://github.com/ofiwg/libfabric/commit/e05eae92e768d4f67ae4d0ce6b16c014bdbb0a51 what results in out of bounds error.

j-xiong:
The array is no longer terminated with a NULL pointer. Check the array boundary instead.

Fixes #11116

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -100,7 +100,7 @@ static int fi_convert_log_str(const char *value)
 	if (!value)
 		return -1;
 
-	for (i = 0; log_levels[i]; i++) {
+	for (i = 0; i < OFI_LOG_MAX; i++) {
 		if (!strcasecmp(value, log_levels[i]))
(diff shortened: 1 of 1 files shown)


j-xiong:
bot:aws:retest

j-xiong:
bot:aws:retest

aingerson (approved):

PR #11144, opened 2025-06-17
prov/efa: Fix the wait_send procedure

darrylabbate:
This silently drops applicable outstanding operations for a closing EP, instead of reporting completions. Exceptions are RECEIPT and EOR packets when `FI_RM_ENABLED` is set - these packets will be reposted when RNR is encountered.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -256,30 +256,21 @@ efa_cq_proc_ibv_recv_rdma_with_imm_completion(struct efa_base_ep *base_ep,
  */
 int efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 {
-	bool should_end_poll = false;
 	struct efa_base_ep *base_ep;
 	struct efa_cq *cq;
 	struct efa_domain *efa_domain;
 	struct fi_cq_tagged_entry cq_entry = {0};
-	struct fi_cq_err_entry err_entry;
 	int err = 0;
 	size_t num_cqe = 0; /* Count of read entries */
 	int prov_errno, opcode;
-
-	/* Initialize an empty ibv_poll_cq_attr struct for ibv_start_poll.
-	 * EFA expects .comp_mask = 0, or otherwise returns EINVAL.
-	 */
-	struct ibv_poll_cq_attr poll_cq_attr = {.comp_mask = 0};
 
 	cq = container_of(ibv_cq, struct efa_cq, ibv_cq);
 	efa_domain = container_of(cq->util_cq.domain, struct efa_domain, util_domain);
 
 	/* Call ibv_start_poll only once */
-	err = efa_ibv_cq_start_poll(ibv_cq, &poll_cq_attr);
+	efa_cq_start_poll(ibv_cq);
 
-	should_end_poll = !err;
-
-	while (!err) {
+	while (efa_cq_wc_available(ibv_cq)) {
 		base_ep = efa_domain->qp_table[efa_ibv_cq_wc_read_qp_num(ibv_cq) & efa_domain->qp_table_sz_m1]->base_ep;
 		opcode = efa_ibv_cq_wc_read_opcode(ibv_cq);
 		if (ibv_cq->ibv_cq_ex->status) {
@@ -337,29 +328,10 @@ int efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 			break;
 		}
 
-		err = efa_ibv_cq_next_poll(ibv_cq);
+		efa_cq_next_poll(ibv_cq);
 	}
-
-	if (err && err != ENOENT) {
-		err = err > 0 ? err : -err;
-		prov_errno = efa_ibv_cq_wc_read_vendor_err(ibv_cq);
-		EFA_WARN(FI_LOG_CQ,
-			 "Unexpected error when polling ibv cq, err: %s (%d) "
-			 "prov_errno: %s (%d)\n",
-			 fi_strerror(err), err, efa_strerror(prov_errno),
-			 prov_errno);
-		efa_show_help(prov_errno);
-		err_entry = (struct fi_cq_err_entry) {
-			.err = err,
-			.prov_errno = prov_errno,
-			.op_context = NULL,
-		};
-		ofi_cq_write_error(&cq->util_cq, &err_entry);
-	}
-
-	if (should_end_poll)
-		efa_ibv_cq_end_poll(ibv_cq);
-
+	err = ibv_cq->poll_err;
+	efa_cq_end_poll(ibv_cq);
 	return err;
 }
 
--- a/prov/efa/src/efa_cq.h
+++ b/prov/efa/src/efa_cq.h
@@ -16,10 +16,11 @@ struct efa_ibv_cq {
 	struct ibv_cq_ex *ibv_cq_ex;
 	enum ibv_cq_ex_type ibv_cq_ex_type;
 	bool data_path_direct_enabled;
+	bool poll_active;
+	int poll_err;
 #if HAVE_EFADV_QUERY_CQ
 	struct efa_data_path_direct_cq data_path_direct;
 #endif
-
 };
 
 struct efa_ibv_cq_poll_list_entry {
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -557,73 +557,129 @@ static int efa_rdm_cq_match_ep(struct dlist_entry *item, const void *ep)
 	return (container_of(item, struct efa_rdm_ep, entry) == ep) ;
 }
 
-/**
- * @brief poll rdma-core cq and process the cq entry
- *
- * @param[in]	ep_poll	the RDM endpoint that polls ibv cq. Note this polling endpoint can be different
- * from the endpoint that the completed packet entry was posted from (pkt_entry->ep).
- * @param[in]	cqe_to_process	Max number of cq entry to poll and process. A negative number means to poll until cq empty
- */
-int efa_rdm_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
+static inline struct efa_rdm_ep *efa_rdm_cq_get_rdm_ep(struct efa_ibv_cq *cq, struct efa_domain *efa_domain)
 {
-	struct efa_rdm_pke *pkt_entry;
-	int err;
-	int opcode;
-	size_t i = 0;
+	struct efa_base_ep *base_ep = efa_domain->qp_table[efa_ibv_cq_wc_read_qp_num(cq) & efa_domain->qp_table_sz_m1]->base_ep;
+	return container_of(base_ep, struct efa_rdm_ep, base_ep);
+}
+
+/**
+ * @brief Process work completions for a closing endpoint
+ *
+ * This is a lighter-weight counterpart to #efa_rdm_cq_process_wc(); avoiding
+ * unnecessary overhead for processing completions for an endpoint that's
+ * closing anyway by simply releasing the packet entries. Exceptions include
+ * RECEIPT and EOR packets when a completion fails due to RNR and resource
+ * management is enabled by the user (FI_RM_ENABLED). In this case, packets are
+ * queued to be reposted.
+ *
+ * @param[in]	cq	IBV CQ
+ * @param[in]	ep	EFA RDM endpoint (to be closed)
+ * @return	Status code for the work completion
+ */
+static inline
+enum ibv_wc_status efa_rdm_cq_process_wc_closing_ep(struct efa_ibv_cq *cq, struct efa_rdm_ep *ep)
+{
+	uint64_t wr_id = cq->ibv_cq_ex->wr_id;
+	enum ibv_wc_status status = cq->ibv_cq_ex->status;
+	enum ibv_wc_opcode opcode = efa_ibv_cq_wc_read_opcode(cq);
+	struct efa_rdm_pke *pkt_entry = (struct efa_rdm_pke *) wr_id;
 	int prov_errno;
-	struct efa_rdm_ep *ep = NULL;
-	struct efa_cq *efa_cq;
-	struct efa_domain *efa_domain;
-	struct efa_qp *qp;
-	struct efa_rdm_peer *peer = NULL;
-	struct dlist_entry rx_progressed_ep_list, *tmp;
 
-	efa_cq = container_of(ibv_cq, struct efa_cq, ibv_cq);
-	efa_domain = container_of(efa_cq->util_cq.domain, struct efa_domain, util_domain);
-	dlist_init(&rx_progressed_ep_list);
-
-	/* Call ibv_start_poll only once */
-	efa_cq_start_poll(ibv_cq);
-
-	while (efa_cq_wc_available(ibv_cq)) {
-		pkt_entry = (void *)(uintptr_t)ibv_cq->ibv_cq_ex->wr_id;
-		qp = efa_domain->qp_table[efa_ibv_cq_wc_read_qp_num(ibv_cq) & efa_domain->qp_table_sz_m1];
-		ep = container_of(qp->base_ep, struct efa_rdm_ep, base_ep);
 #if HAVE_LTTNG
-		efa_rdm_tracepoint(poll_cq, (size_t) ibv_cq->ibv_cq_ex->wr_id);
-		if (pkt_entry && pkt_entry->ope)
-			efa_rdm_tracepoint(poll_cq_ope, pkt_entry->ope->msg_id,
-					   (size_t) pkt_entry->ope->cq_entry.op_context,
-					   pkt_entry->ope->total_len, pkt_entry->ope->cq_entry.tag,
-					   pkt_entry->ope->peer ? pkt_entry->ope->peer->conn->fi_addr : FI_ADDR_NOTAVAIL);
+	efa_rdm_tracepoint(poll_cq, (size_t) wr_id);
+	if (pkt_entry && pkt_entry->ope)
+		efa_rdm_tracepoint(poll_cq_ope, pkt_entry->ope->msg_id,
+				   (size_t) pkt_entry->ope->cq_entry.op_context,
+				   pkt_entry->ope->total_len, pkt_entry->ope->cq_entry.tag,
+				   pkt_entry->ope->peer ? pkt_entry->ope->peer->conn->fi_addr : FI_ADDR_NOTAVAIL);
 #endif
-		opcode = efa_ibv_cq_wc_read_opcode(ibv_cq);
-		if (ibv_cq->ibv_cq_ex->status) {
-			if (pkt_entry)
-				peer = pkt_entry->peer;
-			prov_errno = efa_rdm_cq_get_prov_errno(ibv_cq, peer);
-			switch (opcode) {
-			case IBV_WC_SEND: /* fall through */
-			case IBV_WC_RDMA_WRITE: /* fall through */
-			case IBV_WC_RDMA_READ:
-				efa_rdm_pke_handle_tx_error(pkt_entry, prov_errno);
-				break;
-			case IBV_WC_RECV: /* fall through */
-			case IBV_WC_RECV_RDMA_WITH_IMM:
-				if (efa_cq_wc_is_unsolicited(ibv_cq)) {
-					EFA_WARN(FI_LOG_CQ, "Receive error %s (%d) for unsolicited write recv",
-						efa_strerror(prov_errno), prov_errno);
-					efa_base_ep_write_eq_error(&ep->base_ep, to_fi_errno(prov_errno), prov_errno);
+
+	if (!efa_cq_wc_is_unsolicited(cq)) {
+		if (OFI_UNLIKELY(status != IBV_WC_SUCCESS)) {
+			prov_errno = efa_rdm_cq_get_prov_errno(cq, pkt_entry->peer);
+			if (prov_errno == EFA_IO_COMP_STATUS_REMOTE_ERROR_RNR &&
+				ep->handle_resource_management == FI_RM_ENABLED) {
+				switch(efa_rdm_pkt_type_of(pkt_entry)) {
+				case EFA_RDM_RECEIPT_PKT:
+				case EFA_RDM_EOR_PKT:
+					efa_rdm_ep_record_tx_op_completed(ep, pkt_entry);
+					efa_rdm_ep_queue_rnr_pkt(ep, pkt_entry);
+					return status;
+				default:
 					break;
 				}
-				efa_rdm_pke_handle_rx_error(pkt_entry, prov_errno);
-				break;
-			default:
-				EFA_WARN(FI_LOG_EP_CTRL, "Unhandled op code %d\n", opcode);
-				assert(0 && "Unhandled op code");
 			}
-			break;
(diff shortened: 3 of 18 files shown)


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -808,13 +808,16 @@ bool efa_rdm_ep_has_unfinished_send(struct efa_rdm_ep *efa_rdm_ep)
...
+	if (efa_rdm_ep->efa_outstanding_tx_ops > 0) {
+		EFA_WARN(FI_LOG_EP_CTRL, "efa_rdm_ep->efa_outstanding_tx_ops: %zu\n", efa_rdm_ep->efa_outstanding_tx_ops);
 		return true;
+	}
 
 	dlist_foreach_safe(&efa_rdm_ep_domain(efa_rdm_ep)->ope_queued_list, entry, tmp) {
 		ope = container_of(entry, struct efa_rdm_ope,
 					queued_entry);
 		if (ope->ep == efa_rdm_ep && (ope->internal_flags & queued_ope_flags)) {
+			EFA_WARN(FI_LOG_EP_CTRL, "efa_rdm_ep has queued packets \n");
```
shijin-aws: Remove my smuggles :slightly_smiling_face:

shijin-aws: same for L812

prov/efa/src/efa_cq.h
```diff
@@ -94,23 +94,23 @@ void efa_ibv_cq_poll_list_remove(struct dlist_entry *poll_list, struct ofi_genlo
...
+		.ibv_cq_ex_type = IBV_CQ,
+	};
 
-	if (!*ibv_cq_ex) {
+	if (!ibv_cq->ibv_cq_ex) {
 		EFA_WARN(FI_LOG_CQ, "Unable to create extended CQ: %s\n", strerror(errno));
 		return -FI_EINVAL;
 	}
 
-	*ibv_cq_ex_type = IBV_CQ;
```
shijin-aws: This is a behavior change, that makes `ibv_cq_ex_type` always set as `IBV_CQ`.

shijin-aws: Looking at it again, it seems not hurt though as it should not be used on the error path

darrylabbate: Yeah this shouldn't change anything, but the whole commit can be dropped regardless

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -844,9 +844,9 @@ void efa_rdm_ep_wait_send(struct efa_rdm_ep *efa_rdm_ep)
 	while (efa_rdm_ep_has_unfinished_send(efa_rdm_ep)) {
 		/* poll cq until empty */
 		if (tx_cq)
-			efa_rdm_cq_poll_ibv_cq(-1, &tx_cq->ibv_cq);
+			efa_rdm_cq_poll_ibv_cq_closing_ep(&tx_cq->ibv_cq, efa_rdm_ep);
 		if (rx_cq)
-			efa_rdm_cq_poll_ibv_cq(-1, &rx_cq->ibv_cq);
+			efa_rdm_cq_poll_ibv_cq_closing_ep(&rx_cq->ibv_cq, efa_rdm_ep);
```
shijin-aws, line 887: `efa_domain_progress_rdm_peers_and_queues` can repost all queued packets, including the closing ep. I do not think this is the behavior we want

shijin-aws: Since we still need to post the RNRed EOR/RECEIPT pkts queued earlier, that likely means we need to have another `closing_ep` version of `efa_domain_progress_rdm_peers_and_queues`

darrylabbate: To be clear, we _only_ want to repost the RNR'd RECEIPT/EOR packets after exhausting the CQs in the wait_send procedure, correct?

darrylabbate:
AWS CI failure should be fixed by #11157

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -822,6 +822,67 @@ bool efa_rdm_ep_has_unfinished_send(struct efa_rdm_ep *efa_rdm_ep)
...
+			peer->flags &= ~EFA_RDM_PEER_IN_BACKOFF;
+			dlist_remove(&peer->rnr_backoff_entry);
+		}
+	}
+
+	dlist_foreach_container_safe(&domain->ope_queued_list,
+			struct efa_rdm_ope, ope, queued_entry, tmp) {
+		if (ope->ep == ep) {
+			if (ope->internal_flags & EFA_RDM_OPE_QUEUED_RNR) {
+				assert(!dlist_empty(&ope->queued_pkts));
```
shijin-aws: these posts all pkts, can we only post for the two pkt types we proposed earlier

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -822,6 +822,67 @@ bool efa_rdm_ep_has_unfinished_send(struct efa_rdm_ep *efa_rdm_ep)
...
+						efa_rdm_txe_handle_error(ope, -ret, FI_EFA_ERR_PKT_SEND);
+					continue;
+				}
+
+				dlist_remove(&ope->queued_entry);
+				ope->internal_flags &= ~EFA_RDM_OPE_QUEUED_RNR;
+			}
+
+			if (ope->internal_flags & EFA_RDM_OPE_QUEUED_CTRL) {
+				ret = efa_rdm_ope_post_send(ope, ope->queued_ctrl_type);
```
shijin-aws: Same comment as earlier

darrylabbate:
Updating this but also looking into previous failures on AL2023

prov/efa/src/efa_cq.c
```diff
@@ -328,28 +319,9 @@ void efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
...
-			 fi_strerror(err), err, efa_strerror(prov_errno),
-			 prov_errno);
-		efa_show_help(prov_errno);
-		err_entry = (struct fi_cq_err_entry) {
-			.err = err,
-			.prov_errno = prov_errno,
-			.op_context = NULL,
-		};
-		ofi_cq_write_error(&cq->util_cq, &err_entry);
+		efa_cq_next_poll(ibv_cq);
```
bwbarrett, line 331: ew, you just made an already too deep call stack even deeper.  This can't be the sane answer.

darrylabbate: I'll admit I'm trusting these calls to be inlined: https://github.com/ofiwg/libfabric/blob/83717e25a0129ba0f7a1e904bf5e19b2862682d8/prov/efa/src/efa_cq.h#L348

but I don't really see why these functions wouldn't always be inlined in practice. FWIW, these definitions don't trigger any warnings in the Actions build workflows with `-Winline`: https://github.com/darrylabbate/libfabric/commit/47cf5ca59d8e47c0a917112d5a086c5e935731bc ([ubuntu22/clang](https://github.com/darrylabbate/libfabric/actions/runs/16010332033/job/45166529975), [ubuntu22/gcc](https://github.com/darrylabbate/libfabric/actions/runs/16010332033/job/45166529980))

darrylabbate: In any event, it's easy enough to swap the inline function definitions with function-like macros and call it a day.

prov/efa/src/rdm/efa_rdm_pke_utils.h
```diff
@@ -23,6 +23,29 @@ struct efa_rdm_base_hdr *efa_rdm_pke_get_base_hdr(struct efa_rdm_pke *pke)
...
+static inline
+int efa_rdm_pkt_type_of_base_hdr(struct efa_rdm_base_hdr *base_hdr)
+{
+	return base_hdr->type;
+}
+
+static inline
+int efa_rdm_pkt_type_of_pke(struct efa_rdm_pke *pke)
+{
+	return efa_rdm_pkt_type_of_base_hdr(efa_rdm_pke_get_base_hdr(pke));
```
shijin-aws: U need a check that whether the pke really has headers.

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -324,51 +324,117 @@ static int efa_rdm_cq_match_ep(struct dlist_entry *item, const void *ep)
...
-	/* Call ibv_start_poll only once */
-	efa_cq_start_poll(ibv_cq);
+	if (!efa_cq_wc_is_unsolicited(cq->ibv_cq_ex)) {
+		peer = efa_rdm_ep_get_peer(ep, pkt_entry->addr);
+		if (OFI_UNLIKELY(status != IBV_WC_SUCCESS)) {
+			prov_errno = efa_rdm_cq_get_prov_errno(cq->ibv_cq_ex, peer);
+			if (!(pkt_entry->flags & EFA_RDM_PKE_SEND_TO_USER_RECV_QP) &&
+				prov_errno == EFA_IO_COMP_STATUS_REMOTE_ERROR_RNR &&
+				ep->handle_resource_management == FI_RM_ENABLED) {
+				switch(efa_rdm_pkt_type_of(pkt_entry)) {
```
shijin-aws, line 603: Similar comment as before.. If we can move the `pkt_entry->flags & EFA_RDM_PKE_SEND_TO_USER_RECV_QP`  check to `efa_rdm_pkt_type_of`, we can simplify this `if`

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -818,6 +819,71 @@ bool efa_rdm_ep_has_unfinished_send(struct efa_rdm_ep *efa_rdm_ep)
...
+					peer->rnr_backoff_wait_time) {
+			peer->flags &= ~EFA_RDM_PEER_IN_BACKOFF;
+			dlist_remove(&peer->rnr_backoff_entry);
+		}
+	}
+
+	dlist_foreach_container_safe(&domain->ope_queued_list,
+			struct efa_rdm_ope, ope, queued_entry, tmp) {
+		if (ope->ep == ep) {
+			switch (efa_rdm_pkt_type_of(ope)) {
```
shijin-aws, line 843: similar comment on the type check

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -818,6 +819,71 @@ bool efa_rdm_ep_has_unfinished_send(struct efa_rdm_ep *efa_rdm_ep)
...
+			case EFA_RDM_RECEIPT_PKT:
+			case EFA_RDM_EOR_PKT:
+				if (ope->internal_flags & EFA_RDM_OPE_QUEUED_RNR) {
+					assert(!dlist_empty(&ope->queued_pkts));
```
shijin-aws: Can we have some inline function to reduce the code duplication between this one and https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_domain.c#L802-L839

darrylabbate:
Updating with rebased branch to keep updated, but I don't expect this to pass CI before encountering timeouts

@shijin-aws just now saw your latest comments

prov/efa/src/rdm/efa_rdm_ep_fiops.c
darrylabbate: Still need to address this

prov/efa/src/rdm/efa_rdm_ep_fiops.c
darrylabbate: This should be taken care of by the added packet entry check

prov/efa/src/rdm/efa_rdm_pke_utils.h
darrylabbate: Added some extra logic for this

prov/efa/src/rdm/efa_rdm_pke_utils.h
```diff
@@ -11,6 +11,13 @@
 #include "efa_rdm_pke_rtm.h"
 #include "efa_mr.h"
 
+static inline
+bool efa_rdm_pke_has_base_hdr(struct efa_rdm_pke *pke)
+{
+	return pke->alloc_type != EFA_RDM_PKE_FROM_READ_COPY_POOL &&
```
shijin-aws: Should be EFA_RDM_PKE_FROM_USER_RX_POOL

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -324,51 +324,117 @@ static int efa_rdm_cq_match_ep(struct dlist_entry *item, const void *ep)
...
+				   pkt_entry->ope->peer->conn->fi_addr);
+#endif
 
-	/* Call ibv_start_poll only once */
-	efa_cq_start_poll(ibv_cq);
+	if (!efa_cq_wc_is_unsolicited(cq->ibv_cq_ex)) {
+		peer = pkt_entry->peer;
+		if (OFI_UNLIKELY(status != IBV_WC_SUCCESS)) {
+			prov_errno = efa_rdm_cq_get_prov_errno(cq->ibv_cq_ex, peer);
+			if (!(pkt_entry->flags & EFA_RDM_PKE_SEND_TO_USER_RECV_QP) &&
```
shijin-aws: You can now use `efa_rdm_pke_has_hdr` ?

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: Right

prov/efa/src/rdm/efa_rdm_pke_utils.h
darrylabbate: You mean additionally? This logic indicates `EFA_RDM_PKE_FROM_READ_COPY_POOL` has no base header as well.

https://github.com/ofiwg/libfabric/blob/c8feaf9eba3dd3d1eb325cfe8fb6bc5e4e3ab0bf/prov/efa/src/rdm/efa_rdm_pke_utils.c#L504-L509

shijin-aws: Ah, I missed that, you are right, need to be additionally. I forgot that we already make readcopy pool header-less

shijin-aws: Think about it more, it's not scaled to have so many bits that indicate a pkt_entry is header-less. Can we just define a dedicated flag for this so we can check it with one flag

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: True, we can actually remove `!(pkt_entry->flags & EFA_RDM_PKE_SEND_TO_USER_RECV_QP)` entirely since `efa_rdm_pkt_type_of()` will safely check the packet type, or return 0.

shijin-aws: yep

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -72,6 +72,15 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 	pkt_entry->payload_size = 0;
 	pkt_entry->payload_mr = NULL;
 	pkt_entry->peer = NULL;
+
+	switch (alloc_type) {
+	case EFA_RDM_PKE_FROM_USER_RX_POOL:
+	case EFA_RDM_PKE_FROM_READ_COPY_POOL:
+		pkt_entry->flags |= EFA_RDM_PKE_HAS_NO_BASE_HDR;
```
shijin-aws, line 80: Thanks for doing that

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1909,3 +1909,48 @@ ssize_t efa_rdm_ope_repost_ope_queued_before_handshake(struct efa_rdm_ope *ope)
...
+		break;
+	}
+
+	if (OFI_UNLIKELY(ret)) {
+		if (ret == -FI_EAGAIN)
+			return ret;
+
+		assert(ope->type == EFA_RDM_TXE || ope->type == EFA_RDM_RXE);
+		if (ope->type == EFA_RDM_TXE)
+			efa_rdm_txe_handle_error(ope, -ret, flag == EFA_RDM_OPE_QUEUED_RNR ? FI_EFA_ERR_PKT_SEND : FI_EFA_ERR_PKT_POST);
```
shijin-aws: I am not sure we need to have these differentiation. IMO `FI_EFA_ERR_PKT_POST` should cover `FI_EFA_ERR_PKT_SEND` and `FI_EFA_ERR_READ_POST`. The old code even has 3 error code, `FI_EFA_ERR_PKT_SEND`, `FI_EFA_ERR_PKT_POST`, `FI_EFA_ERR_READ_POST`. 

I cannot tell the difference between `FI_EFA_ERR_PKT_POST` and `FI_EFA_ERR_PKT_SEND` at least.

For simplicity I would just use `FI_EFA_ERR_PKT_POST` for all post failure in this function

prov/efa/src/efa_cq.h
```diff
@@ -305,6 +305,11 @@ static inline int efa_write_error_msg(struct efa_base_ep *ep, fi_addr_t addr,
 	return 0;
 }
 
+static inline uint32_t efa_cq_wc_read_qp_num(struct efa_ibv_cq *cq)
+{
+	return ibv_wc_read_qp_num(cq->ibv_cq_ex);
```
shijin-aws: Can you drop this inline function and use ibv directly? That will make my life easier because the PR https://github.com/ofiwg/libfabric/pull/11217 is introducing wrapper that does the same thing

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -555,51 +555,116 @@ static int efa_rdm_cq_match_ep(struct dlist_entry *item, const void *ep)
...
+	efa_rdm_tracepoint(poll_cq, (size_t) wr_id);
+	if (pkt_entry && pkt_entry->ope)
+		efa_rdm_tracepoint(poll_cq_ope, pkt_entry->ope->msg_id,
+				   (size_t) pkt_entry->ope->cq_entry.op_context,
+				   pkt_entry->ope->total_len, pkt_entry->ope->cq_entry.tag,
+				   pkt_entry->ope->peer->conn->fi_addr);
+#endif
+
+	if (OFI_UNLIKELY(status != IBV_WC_SUCCESS)) {
+		if (efa_cq_wc_is_unsolicited(cq->ibv_cq_ex)) {
```
shijin-aws: We shouldn't need to check this for non-recv op code. Prefer to still keep them in the IBV_WC_RECV/IBV_WC_RECV_RDMA_WITH_IMM case

prov/efa/src/rdm/efa_rdm_ope.c
darrylabbate: Sure; makes sense. I couldn't find any relevant context, so I just left it as-is.

prov/efa/src/efa_cq.h
darrylabbate: Ack

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: The rearrangement here is mainly to alleviate a recent Coverity scan finding: https://scan4.scan.coverity.com/#/project-view/62117/10344?selectedIssue=468930

This trades a `NULL` check of `pkt_entry` for a `efa_cq_wc_is_unsolocited()` check. Alternatively, we would add extra `NULL` checks within `efa_rdm_pke_handle_{r,t}x_error()`.

shijin-aws: I see, is it possible to silence that warning by adding `assert` ?

darrylabbate: Reverted this s.t. `efa_cq_wc_is_unsolicited()` is only called for `IBV_WC_RECV{,_RDMA_WITH_IMM}`

darrylabbate:
Latest AWS CI run looked good. c6gn tests were finishing up and the tests that previously timed out had passed. Needed to restart since there was an unrelated infra issue.

shijin-aws (dismissed):

prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -573,16 +573,11 @@ void efa_rdm_pke_handle_send_completion(struct efa_rdm_pke *pkt_entry)
...
-		efa_rdm_pke_handle_eager_rtm_send_completion(pkt_entry);
-		efa_rdm_ep_record_tx_op_completed(ep, pkt_entry);
-		efa_rdm_pke_release_tx(pkt_entry);
-		return;
-	}
-
 	/* Start handling pkts with hdrs */
-	switch (efa_rdm_pke_get_base_hdr(pkt_entry)->type) {
+	switch (efa_rdm_pkt_type_of(pkt_entry)) {
+	case EFA_RDM_HEADERLESS_PKT:
```
shijin-aws: The comment in L576 is `/* Start handling pkts with hdrs */` :) 

It may be better to recover L 576-583 and drop the `case EFA_RDM_HEADERLESS_PKT` ?
It is more intuitive to distinguish the user recv qp send pkts and the local read pkt by flags explicitly in the send completion handler, IMO

darrylabbate: Sure. The comment sounds like it was simply trying to handle headerless packets; and it cleaned up nicely as well.

shijin-aws (approved):

darrylabbate:
bot:aws:retest

shijin-aws (approved):

shijin-aws:
@j-xiong [continuous-integration/appveyor/pr](https://ci.appveyor.com/project/ofiwg/libfabric/builds/52558781) failure looks unrelated


https://ci.appveyor.com/project/ofiwg/libfabric/builds/52558781/job/fg8vxeog4xupl0vx

```
ossible loss of data (compiling source file prov\tcp\src\xnet_profile.c) [C:\projects\libfabric\libfabric.vcxproj]
Error sending AppVeyor compilation message: The remote server returned an error: (500) Internal Server Error.
  Message: 'return': conversion from 'ssize_t' to 'int', possible loss of data (compiling source file prov\tcp\src\xnet_progress.c)
  Category: Warning
  Code: C4244
  File: C:\projects\libfabric\include\ofi_util.h
  Line: 701

```

Can u confirm that?

j-xiong:
I believe so. restarted the failed run anyway.

shijin-aws:
CI passed. @darrylabbate can you backport it to 2.2.x

darrylabbate:
> CI passed. @darrylabbate can you backport it to 2.2.x

That'll require #11217 as well. Did you plan on backporting those changes?

shijin-aws:
@darrylabbate Can you do a backport without https://github.com/ofiwg/libfabric/pull/11217 , I think you still have a good version https://github.com/ofiwg/libfabric/commit/9fd088cc2b88adee3f1177dabb86dbab8af08d5a before the rebase?

darrylabbate:
> @darrylabbate Can you do a backport without #11217 , I think you still have a good version [9fd088c](https://github.com/ofiwg/libfabric/commit/9fd088cc2b88adee3f1177dabb86dbab8af08d5a) before the rebase?

I need `efa_data_path_ops.h`. I can't non-trivially apply 1ab60e211 without it.

PR #11146, opened 2025-06-18
prov/efa: Track mr status

shijin-aws:
Currently, there is no track for the reg/dereg status of
an MR in efa provider, because `fi_close(mr)` will simply
free all the `efa_mr` object. This will cause crash or other
undeterministic when an application close a mr when there
are outstanding operations still reference it.

This patch resolves this problem by introducing two fields
in efa_mr structure: active and ref.

ref is the count of operations (op entries) that an mr
is referenced. It is initialized as 0 during mr registration,
, incremented when the mr is referenced by an operation,
and decremented when the operation is resolved.

active is a simple boolean that is toggled as true when
mr is registered and toggled as false when mr is deregistered
by fi_close.

The lifecycle or an efa_mr object behaves as:

1. Allocated during fi_mr_reg*, initialize active as 1, ref as 0
2. op 1 referenced the mr, ref++
3. op 2 referenced the mr, ref--
4. fi_close dereg the mr, active is set as 0, while the object is
not freed.
5. op 1 completes, ref--
6. op 2 completes, ref--, now ref is 0 and active is 0, object is
freed.

The sequence of 4, 5, 6 can be any order, and the object will be
freed only when active is 0 && ref is 0

Also introduce a check in the operation calls that any access to
inactive mr will fail with -FI_EINVAL.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -460,13 +460,19 @@ static int efa_mr_dereg_impl(struct efa_mr *efa_mr)
 static int efa_mr_close(fid_t fid)
 {
 	struct efa_mr *efa_mr;
-	int ret;
+	int ret = 0;
 
 	efa_mr = container_of(fid, struct efa_mr, mr_fid.fid);
-	ret = efa_mr_dereg_impl(efa_mr);
-	if (ret)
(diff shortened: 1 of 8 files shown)


shijin-aws:
Will rework it, close it for now

PR #11147, opened 2025-06-18
prov/shm: replace map retry with rename and retry

aingerson:
The shm provider does its best to randomize an EP name and make it unique by defaulting to "pid:domain_id:ep_id"
If there already exists a region with that name, the provider wil try to map to the existing region and see if the pid is alive and remap if the pid is dead.
This method doesn't work well with docker containers as the pids live within the container. This means you can have a docker pid 100 that coincides with a host system pid 100 running a root service OR you could have two docker containers running simultaneously that are trying to use the same pid (and therefore shm region name). This can result in a failure to create the shm region or failure to detect whether the pid is actually alive. To fix this, we can fallback to a retry of the shm create and append a random seed number to the shm region name to make sure it is unique in case a shm region already exists with that name.
There is a scenario where the app overrides the default shm name with a specific name. If that's the case, we shouldn't override the user-selected name and, instead, fail if it's not available.
The repercussion of this change is that the provider can no longer cleanup old regions from dead pids. The signal handlers should handle most of these issues but can't handle uninterruptable signals. These regions will persist and may cause memory issues but they should be rare (for example gdb sessions). It should be the user's responsibility to periodically clean up the shm regions from debugging sessions


aingerson:
@shijin-aws we saw some issues running with docker because of duplicate pids. Let me know if you have any concerns about this change for the remap process. It has some implications regarding stale region cleanup (explained in the commit message) so I want to make sure that is ok for you.

*Edit: I removed the limitation with the stale region cleanup and realized there was no reason to remove that. We still try to check and cleanup if possible and only fall back if absolutely necessary

aingerson:
@shijin-aws What's the AWS failure?

sunkuamzn:
@aingerson failure is unrelated to the PR

sunkuamzn:
bot:aws:retest

aingerson:
@shijin-aws Would appreciate a review from an AWS member whenever you have a minute. Thanks!

aingerson:
@sunkuamzn Is Shi OOO? Can you give a review?

a-szegel (approved):
EFA Provider is untouched by this change b/c they always set the SHM name.

PR #11148, opened 2025-06-18
man/fi_endpoint: Clarify rx_attr->caps usage

a-szegel:


man/fi_endpoint.3.md
```diff
@@ -1336,8 +1336,8 @@ struct fi_rx_attr {
 ## caps - Capabilities
 
 The requested capabilities of the context.  The capabilities must be
-a subset of those requested of the associated endpoint.  See the
-CAPABILITIES section if fi_getinfo(3) for capability details.  If
+a subset of those requested of the associated domain or scalable endpoint.
```
j-xiong: domain caps are different from endpoint caps. rx caps (and tx caps) are subset of endpoint caps, not domain caps. 

note that endpoint caps are expressed via `fi_info->caps`, although not all caps there are for endpoints.

a-szegel: Thanks, I modified the statement to make that more clear.

a-szegel: Do we need to differentiate domain vs ep capabilities instead of calling them generic capabilities?  The reason I ask is it might take several resources to implement FI_SOURCE (AV, CQ, EP), and calling it an EP only capability could be a misnomer.

man/fi_endpoint.3.md
```diff
@@ -1336,8 +1336,8 @@ struct fi_rx_attr {
 ## caps - Capabilities
 
 The requested capabilities of the context.  The capabilities must be
-a subset of those requested of the associated endpoint.  See the
-CAPABILITIES section if fi_getinfo(3) for capability details.  If
+a subset of those requested of the associated endpoint in fi_info->caps.
+See the CAPABILITIES section if fi_getinfo(3) for capability details.  If
```
j-xiong: `if` -> `of`

man/fi_endpoint.3.md
j-xiong: Domain level caps are described in the `fi_domain` man page. 

`FI_SOURCE` is an endpoint level capability. It doesn't matter how it is implemented; it's associated with an endpoint as perceived by the user.

man/fi_endpoint.3.md
a-szegel: good catch, fixed

PR #11149, opened 2025-06-18
prov/shm: fix srx entry cleanup

### LINKED ISSUE #11111: prov/shm: 🐛 Assertion failure on ofi_atomic_dec32(&ofi_buf_region(buf)->use_cnt) >= 0 during fi_close with shm provider bug
**Describe the bug**
After applying the fix for the previously reported ofi_genlock_held(srx->lock) assertion failure (see: https://github.com/ofiwg/libfabric/issues/11098), a new nondeterministic crash occurs during application shutdown. The crash is triggered by an assertion failure in ofi_buf_free():

```
ofi_buf_free: Assertion `ofi_atomic_dec32(&ofi_buf_region(buf)->use_cnt) >= 0` failed.
```
The issue seems to be related to reference counting in shared memory regions, possibly due to premature or duplicate release of resources.

**To Reproduce**
Unfortunately, this issue is nondeterministic and part of a larger multi-threaded application. It consistently appears during teardown, after calling fi_close() on a shared memory endpoint (FI_EP_RDM).

I cannot provide a minimal reproducible example, but the crash occurs only after applying the previous fix, and in test configurations using the shm provider.

**Output**
```
./include/ofi_mem.h:456: ofi_buf_free: Assertion `ofi_atomic_dec32(&ofi_buf_region(buf)->use_cnt) >= 0' failed.
Aborted (core dumped)
```

```
#7  0x00007ffff7ec21bc in ofi_buf_free (buf=0x7fffe4003010) at ./include/ofi_mem.h:456
#8  0x00007ffff7ec5f94 in smr_discard (rx_entry=0x7fffe4142330) at prov/shm/src/smr_ep.c:1064
#9  0x00007ffff7e4d40b in util_srx_close (fid=0x7fffe40dd080) at prov/util/src/util_srx.c:1007
#10 0x00007ffff7ec53e4 in smr_ep_close (fid=0x7fffe41202b0) at prov/shm/src/smr_ep.c:830
```

**Environment:**
Same as in (see: https://github.com/ofiwg/libfabric/issues/11098).

**Additional context**
This is a follow-up to my previous issue:

"Application crashes nondeterministically during shutdown with assertion failure in util_free_entry() (ofi_genlock_held)."
That issue was resolved with PR [11105](https://github.com/ofiwg/libfabric/pull/11105), but this new crash now occurs in the same shutdown path.

**Help Appreciated**
Please advise if further diagnostics, environment flags, or debug builds could help clarify this issue. I’d be happy to provide any other info or run additional tests to isolate the cause.

Thanks again for your ongoing support.

aingerson:
When cleaning up the srx (util_srx_close), the srx iterates through the list of unexpected messages and calls discard_msg/tag on each entry which means shm shouldn't have to iterate through the list of unexpected messages and free each one. The unexp_cmd_list was only there for cleanup so it can get removed.
The special case is unexpected SAR entries which also live in the sar list so the pend entry can get cleaned up. In this case, we can't free the srx entry but need to free the pend entry. This case is handled in the sar list cleanup so only the expected messages are freed. For the unexpected case, this will get triggered on the srx close

Fixes #11111

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -141,7 +141,6 @@ struct smr_pend_entry {
 };
 
 struct smr_cmd_ctx {
-	struct dlist_entry entry;
 	struct fi_peer_rx_entry *rx_entry;
 	struct smr_ep *ep;
 	struct smr_cmd cmd;
@@ -230,7 +229,6 @@ struct smr_ep {
 	struct smr_tx_fs	*tx_fs;
 	struct dlist_entry	sar_list;
 	struct dlist_entry	ipc_cpy_pend_list;
-	struct dlist_entry	unexp_cmd_list;
 	size_t			min_multi_recv_size;
 
 	int			ep_idx;
--- a/prov/shm/src/smr_ep.c
+++ b/prov/shm/src/smr_ep.c
@@ -791,7 +791,6 @@ static int smr_ep_close(struct fid *fid)
 {
 	struct smr_ep *ep;
 	struct smr_pend_entry *pend;
-	struct smr_cmd_ctx *cmd_ctx;
 
 	ep = container_of(fid, struct smr_ep, util_ep.ep_fid.fid);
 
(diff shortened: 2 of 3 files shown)


j-xiong (approved):

j-xiong:
bot:aws:retest

PR #11150, opened 2025-06-19
prov/efa: Set IBV_ACCESS_LOCAL_WRITE for FI_REMOTE_WRITE

sunkuamzn:
When emulating an fi_write with RDMA read, the destination buffer needs
IBV_ACCESS_LOCAL_WRITE permission even though it is registered only with
FI_REMOTE_WRITE

This commit also adds a fabtest that covers this code path


prov/efa/src/rdm/efa_rdm_ep.h
```diff
@@ -508,14 +508,26 @@ static inline int efa_rdm_attempt_to_sync_memops(struct efa_rdm_ep *ep, void *bu
...
+	struct efa_mr *efa_mr;
 
 	if (!desc)
 		return err;
 
-	for (i = 0; i < num_desc; i++) {
-		err = efa_rdm_attempt_to_sync_memops(ep, iov[i].iov_base, (struct efa_mr *) desc[i]);
-		if (err)
-			return err;
+	if (OFI_UNLIKELY(ep->cuda_api_permitted)) {
```
shijin-aws, line 499: Can we move this cuda_api_permitted check out of `efa_rdm_attempt_to_sync_memops`, so `efa_rdm_attempt_to_sync_memops_iov` and `efa_rdm_attempt_to_sync_memops_ioc` can still reuse `efa_rdm_attempt_to_sync_memops`

prov/efa/src/efa_mr.c
```diff
@@ -784,12 +784,17 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *at
...
 
-	/* Local read access to an MR is enabled by default in verbs */
-	if (core_access & FI_RECV)
+	/* Local read access to an MR is enabled by default in verbs
+	 *
+	 * We need IBV_ACCESS_LOCAL_WRITE for two emulated cases
+	 * 1. When emulating fi_send with RMA read - ofi_access is set to FI_RECV
+	 * 2. When emulating fi_write with RMA read - ofi_access is set to FI_REMOTE_WRITE
+	 */
+	if (ofi_access & (FI_RECV | FI_REMOTE_WRITE))
```
shijin-aws, line 797: According to https://www.rdmamojo.com/2012/09/07/ibv_reg_mr/, `IBV_ACCESS_LOCAL_WRITE` means 

>Enable Local Write Access: Memory Region can be used in Receive Requests or in IBV_WR_ATOMIC_CMP_AND_SWP or
IBV_WR_ATOMIC_FETCH_AND_ADD to write locally the remote content value

Does it exactly fit our model here that the mr is used to get data via rdma read?

I know we have been doing that even before your PR, that an `FI_RECV` access can also involve rdma read in read protocol. I am just curious

sunkuamzn: > Does it exactly fit our model here that the mr is used to get data via rdma read?

I think so. I think the destination buffer of a RDMA read needs to have IBV_ACCESS_LOCAL_WRITE

prov/efa/src/rdm/efa_rdm_ep.h
shijin-aws: You can do something like

```
if (ep->cuda_api_permitted && efa_rdm_attempt_to_sync_memops_iov(efa_rdm_ep, (struct iovec *)msg->msg_iov, msg->desc, msg->iov_count))
		return ret;
```

sunkuamzn: I removed `efa_rdm_attempt_to_sync_memops` instead. We setup the IOVs anyway in the TX path.

shijin-aws (approved):

shijin-aws (approved):

PR #11151, opened 2025-06-19
prov/efa: Drain cq after qp destroy

shijin-aws:
After QP is destroyed, there can be completions in the CQ belong to this QP tha are not polled yet. These stale completions should be drained (as EINVAL from ibv_start_poll/ibv_next_poll) otherwise it can be polled as active completions after a new QP with the same QPN is created.

This patch implements this cq drain after the QP destruction as part of the EP close in libfabric.


sunkuamzn (dismissed):

shijin-aws:
Need to implement the same change for efa-direct ep, will update the PR shortly

jiaxiyan (dismissed):

sunkuamzn (approved):

PR #11155, opened 2025-06-19
prov/util_cq: Fix bug in fi_cq_readfrom with no FI_SOURCE

a-szegel:
The libfabric man pages state that fi_cq_readfrom() will return FI_ADDR_NOTAVAIL for endpoints that are not configured with FI_SOURCE. If the user wishes to use FI_SOURCE, they must pass it as part of fi_info->caps so that domain resources involved in the FI_SOURCE feature (EP, AV, CQ) can all configure themselves correctly. The user is still able to call the fi_cq_readfrom API with or without FI_SOURCE configured in fi_info->caps. This change fixes a bug in util_cq by setting addr = FI_ADDR_NOTAVAIl if the user chooses to call fi_cq_readfrom() without setting fi_info->caps |= FI_SOURCE.


shijin-aws:
>The user is still able to call the fi_cq_readfrom API with or without FI_SOURCE configured in fi_info->caps. 

Can you share which part of man page says that? In my understanding, if users don't configure FI_SOURCE in domain caps, why do they want to call `fi_cq_readfrom`?

a-szegel:
https://ofiwg.github.io/libfabric/main/man/fi_cq.3.html


> If fi_cq_readfrom is called on an endpoint for which source addressing data is not available, the source address will be set to FI_ADDR_NOTAVAIL. The number of input src_addr entries must be the same as the count parameter.

Technically, I do not need the src_addr check for non-null in the IF statement.  I am going to remove that for speed.

a-szegel:
I guess we do need the NULL check to get the CI passing... even though the docs state that we don't.

j-xiong:
Did you see anything suspicious from AWS CI? Intel CI runs have been flaky lately, often with one UDP test timing out.

a-szegel:
pushed to rebase against main, failures from the last round look real on aws-ci.  Will need to dive into them when I get time.

prov/util/src/util_cq.c
```diff
@@ -267,14 +267,24 @@ ssize_t ofi_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 
 	cq = container_of(cq_fid, struct util_cq, cq_fid);
 
+	if (OFI_UNLIKELY(src_addr && !(cq->domain->info_domain_caps & FI_SOURCE)))
+		for (int i = 0; i < count; i++)
+			src_addr[i] = FI_ADDR_NOTAVAIL;
+
```
shefty: These changes don't look right.  A CQ may be used with a mix of endpoints, not all of which may have source address available.  Consider a CQ mixed with msg and rdm EPs.

The general flow is for the provider that's using the util code to insert FI_ADDR_NOTAVAIL as the source address if one is not available.

aingerson: Ignoring the peer case, shouldn't all the EPs be created on the same domain as the CQ which means they should be using the same fi_info/caps? If the CQ wasn't configured with FI_SOURCE then the EPs can't either.
With the peer case, the owner/app's fi_info needs to be configured with FI_SOURCE in order for the main CQ to have source addressing available so it would be redundant to ask for source addressing from the peers.
What EP source address configuration are you talking about? If the source addressing if configured/toggled on an EP level, then it seems like the implementation of the source/non source write functions is wrong since that's happening in the CQ right now.

prov/util/src/util_cq.c
```diff
@@ -267,14 +267,24 @@ ssize_t ofi_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t count,
...
 ssize_t ofi_cq_read(struct fid_cq *cq_fid, void *buf, size_t count)
 {
-	return fi_cq_readfrom(cq_fid, buf, count, NULL);
+	struct util_cq *cq;
+
+	cq = container_of(cq_fid, struct util_cq, cq_fid);
+
+	cq->progress(cq);
+
+	return ofi_cq_read_entries(cq, buf, count, NULL);
```
shefty: This change disallows the provider from using a custom cq_readfrom() call, which is exactly what the tcp provider does, breaking that provider.

prov/util/src/util_cq.c
shefty: MSG EP's do not have source addressing.  Those can share the same domain and CQ as RDM EP's which do.

Even if an RDM EP has source addressing enabled, there's no guarantee that a specific receive can map to a valid source address.  The flow is that the provider would simply insert FI_ADDR_NOTAVAIL if one was not available.

shefty: Try applying a fix to ofi_cq_read_entries()

shefty: Specifically, try changing flows like this:
```
if (src_addr && cq->src)
    src_addr[i] = cq->src[ofi_cirque_rindex(cq->cirq)];
```
to something like
```
if (src_addr)
    src_addr[i] = cq->src ? cq->src[ofi_cirque_rindex(cq->cirq)] : FI_ADDR_NOTAVAIL;
```

shijin-aws:
> https://ofiwg.github.io/libfabric/main/man/fi_cq.3.html
> 
> > If fi_cq_readfrom is called on an endpoint for which source addressing data is not available, the source address will be set to FI_ADDR_NOTAVAIL. The number of input src_addr entries must be the same as the count parameter.
> 
> Technically, I do not need the src_addr check for non-null in the IF statement. I am going to remove that for speed.

I was saying that the following semantic doesn't seem legitimate to me:
1. fi_cq_open W/O  `FI_SOURCE` in fi_info->domain_attr->caps
2. opening endpoints with and without `FI_SOURCE` in `fi_info->caps`, bind them to the cq
3. calling fi_cq_readfrom.

Even with this PR, the current util cq write will crash https://github.com/ofiwg/libfabric/blob/main/include/ofi_util.h#L640, because the `cq->src` is only allocated when `FI_SOURCE` is configured during cq open. Adding a check to `cq->src` will help, but it hurts performance in such fast path.

@shefty WDYT?

shefty:
The check for cq->src is already there.  See ofi_cq_read_entries().  It also checks for non-null src_addr.  The checks just need to nest.  Checking cq->src will be faster than fetching the domain attributes and performing a bit check.

Yes, you could create multiple copies of ofi_cq_read_entries(), optimized for having / not having source addressing, but that's a lot of duplicated code and probably doesn't provide any real benefit.  It's goofy for an app to call fi_cq_readfrom() if it doesn't want source addressing.  It's also goofy to call it but pass in src_addr = null.  But the latter may be done to drive progress, so there's some logic there, I guess.

shijin-aws:
> The check for cq->src is already there. See ofi_cq_read_entries(). It also checks for non-null src_addr. The checks just need to nest. Checking cq->src will be faster than fetching the domain attributes and performing a bit check.

I was saying there is no check for `cq->src` in `ofi_cq_write_src` which happens in the progress engine for many providers. Since we agree that this behavior is goofy anyway so it doesn't worth a check for `cq->src` again.

prov/util/src/util_cq.c
a-szegel: This change is just making it so that `ofi_cq_read` doesn't need to go through the extra check I added.

a-szegel: oh I see... I would need to make them use a custom ofi_cq_read.

prov/util/src/util_cq.c
a-szegel: This patch is supposed to handle the case where the user doesn't configure the CQ with `FI_SOURCE`, and calls `fi_cq_readfrom()`.

The EFA Provider only calls `ofi_cq_write_src` if the endpoint was configured with `FI_SOURCE` to avoid the seg fault b/c cq->src doesn't exist. 

https://github.com/ofiwg/libfabric/blob/main/include/ofi_util.h#L635-L642

If the preferred approach is to always call `ofi_cq_write_src` in the endpoint, then we should always alloc cq->src.

shefty: The util code has 2 write calls, one for src and one without.  But it implements a single read call, which handles all cases, because handling overflow is hard.  The read call already has checks in place for no source, it just needs to update those checks to handle the case to write back ADDR_NOTAVAIL when the user called readfrom with a non-null src_addr.

You should only need to change ofi_cq_read_entries() as I outlined above.  It should not introduce any additional overhead over what's there today.

a-szegel: I understand now, that makes sense.  Thank you!

a-szegel:
I have updated the PR with @shefty's suggested changes.  Thanks everyone!

include/ofi_util.h
```diff
@@ -580,8 +580,8 @@ ssize_t ofi_cq_read_entries(struct util_cq *cq, void *buf, size_t count,
 	for (i = 0; i < (ssize_t) count; i++) {
 		entry = ofi_cirque_head(cq->cirq);
 		if (!(entry->flags & UTIL_FLAG_AUX)) {
-			if (src_addr && cq->src)
-				src_addr[i] = cq->src[ofi_cirque_rindex(cq->cirq)];
+			if (src_addr)
+				src_addr[i] = cq->src ? cq->src[ofi_cirque_rindex(cq->cirq)] : FI_ADDR_NOTAVAIL;
```
shefty: line length is getting long.  Also, see line 599, which I think needs a similar (but not exactly the same) change.

a-szegel: fixed.

shijin-aws (approved):

a-szegel:
bot:aws:retest

a-szegel:
@j-xiong Is the intel CI build failure harmless?

a-szegel:
bot:aws:retest ... looks like random failures

PR #11173, opened 2025-06-24
github/workflows: Keep all-in-one man page in sync with GH man pages

j-xiong:


j-xiong:
No need to wait for the CI results.

PR #11175, opened 2025-06-25
github/workflows: Tighten the token permission settings

j-xiong:
Follow the principle of least privilege by setting the top level permissions to "none" and only granting needed permissions at the "run" level (i.e. for individual job).

Hopefully this can improve the OpenSFF scorecard number.


j-xiong:
CI failures can be ignored.

aingerson (approved):

PR #11178, opened 2025-06-26
contrib/aws: Use ami with pre-installed EFA Installer

a-szegel:
This will reduce PR CI run time by ~3 min and improve stability.


a-szegel:
Some of the pre-built AMI's that we need to do this do not exist yet.

PR #11184, opened 2025-06-27
prov/efa: Remove fallback logic for determining peer address

sunkuamzn:
The underlying bug has been fixed. We no longer need this logic.


shijin-aws:
This code can be useful in the future if we want to replace raw address to connid header completely: for unknown peer (AH is not inserted), you can detect the source gid by using the efadv API, then we never need the raw address. But that is a future work anyway

shijin-aws (approved):

shijin-aws (approved):

PR #11193, opened 2025-07-02
fabtests/pytest: Fix the additional environment return

shijin-aws:
prepare_base_command failed to return the default additional_env for ubertest and host test. This patch fixes it.


sunkuamzn:
bot:aws:retest

PR #11194, opened 2025-07-02
prov/efa: Add lkey to query_mr

jiaxiyan:
Expose lkey to enable applications to construct the TX/RX buffer descriptor in the wqe for send/recv operations.


prov/efa/src/efa_domain.c
```diff
@@ -443,6 +443,8 @@ efa_domain_query_mr(struct fid_mr *mr_fid, struct fi_efa_mr_attr *mr_attr)
 		mr_attr->ic_id_validity |= FI_EFA_MR_ATTR_RDMA_RECV_IC_ID;
 	}
 
+	mr_attr->lkey = efa_mr->ibv_mr->lkey;
```
shijin-aws, line 446: One risk I can see is that if an application is compiled with an old header that doesn't have `lkey`, this will crash, because the input `mr_attr` object doesn't have the memory space for such `lkey`

PR #11195, opened 2025-07-02
util/mon_sampler: Disable the sampler together with the monitor provider

j-xiong:
The monitor sampler can now be disabled together with the monitor provider by running configure with `--disable-monitor` option. This can also serves as a workaround for old gcc compilers (e.g. 4.8.5) that don't support the `_Atomic` keyword.


j-xiong:
bot:aws:retest

aingerson (approved):

PR #11200, opened 2025-07-03
[v2.1.x] Backport several commits

sunkuamzn:


darrylabbate:
@j-xiong Are Intel CI failures real?

j-xiong:
the failures are with rxd/verbs which happen intermittently. Thay can be ignored for this PR.

PR #11205, opened 2025-07-08
prov/util,rxm,shm,sm2: Get srx lock in the caller of util_foreach_unspec

sunkuamzn:
util_queue_[msg,tag] and util_get_[msg,tag] expect the caller to acquire the SRX lock. This commit requires the callers of util_foreach_unspec to also do the same.


sunkuamzn:
bot:aws:retest

shijin-aws (approved):

PR #11206, opened 2025-07-09
prov/efa: Improve the handshake tx error handling

shijin-aws:
Currently we only ignore handshake tx error when the prov_errno is EFA_IO_COMP_STATUS_REMOTE_ERROR_BAD_DEST_QPN, which doesn't cover all errno that indicate the peer is down. This patch expands this errno list to all three prov_errno that local can get when the peer is destroyed.


shijin-aws:
bot:aws:retest

jiaxiyan (approved):

PR #11207, opened 2025-07-09
include/fi_peer.h, prov/util,lnx: remove fi_peer_rx_entry dlist fields

aingerson:
fi_peer_rx_entry has two internal fi_peer_rx_entry fields so the struct could be castable to a dlist_entry or slist_entry. This suggests a dependency between the external headers and the internal headers which if difficult to ensure. In addition, there is no real reason for the owner and peer to share an entry for tracking the peer entry. The owner and peer each have their own structs and mechanisms for tracking the entry when the resource is on their side so the extra fields can be removed completely.

Since there is a requirement for all peers/owners to use the same version of libfabric, there is no issue with backwards compatibility

### DIFF
--- a/include/ofi_util.h
+++ b/include/ofi_util.h
@@ -1286,6 +1286,10 @@ struct ofi_ops_flow_ctrl {
 };
 
 struct util_rx_entry {
+	union {
+		struct dlist_entry	d_entry;
+		struct slist_entry	s_entry;
+	};
 	struct fi_peer_rx_entry	peer_entry;
 	uint64_t		seq_no;
 	uint64_t		ignore;
--- a/prov/util/src/util_srx.c
+++ b/prov/util/src/util_srx.c
@@ -175,8 +175,7 @@ static int util_match_msg(struct fid_peer_srx *srx,
 		ret = -FI_ENOENT;
 	} else {
 		head = srx_ctx->msg_queue.head;
-		util_entry = container_of(head, struct util_rx_entry,
-					  peer_entry);
+		util_entry = container_of(head, struct util_rx_entry, s_entry);
 		if (util_entry->peer_entry.flags & FI_MULTI_RECV) {
 			util_entry = util_process_multi_recv(srx_ctx,
 				&srx_ctx->msg_queue, attr, util_entry);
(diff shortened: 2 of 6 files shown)


aingerson:
I forgot to update the man page so I will do that but will wait for CI to finish and a review or two to catch any other issues.

amirshehataornl (approved):

sunkuamzn (approved):
Thanks!

shefty (dismissed):
I think these are good changes to have, but I'm also unsure that it helps the debug goals.  I'll comment on the other PR.

include/ofi_util.h
```diff
@@ -1286,6 +1286,7 @@ struct ofi_ops_flow_ctrl {
 };
 
 struct util_rx_entry {
+	struct dlist_entry	entry;
```
shefty: If you make this:
```
union {
    struct dlist_entry dentry;
    struct slist_entry sentry;
 }
```
I think it'll work in the debug case without abusing the fields.

prov/util/src/util_srx.c
```diff
@@ -332,18 +328,20 @@ static int util_get_tag(struct fid_peer_srx *srx,
...
 	if (rx_entry->addr == FI_ADDR_UNSPEC) {
-		dlist_insert_tail((struct dlist_entry *) rx_entry,
+		dlist_insert_tail(&util_entry->entry,
 				  &srx_ctx->unspec_unexp_msg_queue);
 	} else {
 		unexp_peer = ofi_array_at(&srx_ctx->src_unexp_peers,
 					  rx_entry->addr);
 		assert(unexp_peer);
-		slist_insert_tail((struct slist_entry *) rx_entry,
+		slist_insert_tail((struct slist_entry *) &util_entry->entry,
```
shefty: With the above, you can drop the casts when using as an slist.

shefty (commented):
Because dlist_entry > slist_entry, the changes should work.  But now that this is internal, the code can explicit indicate that the fields are being used as a union and accessed correctly.

include/ofi_util.h
aingerson: Yeah I considered that and it'll help clean up the code to remove the casts but I don't see why that will have a functional difference?

shefty: There's 'functional' difference if you print the structure in a debugger in terms of being able to read the output when the structure is on an slist.  The behavior in the code at runtime is the same.  In terms of code readability, this makes it clear that dlist and slist entries overlap and the larger struct may be stored on both.

sunkuamzn:
AWS CI is failing on this fabtest on a single node and timing out on others. It might be an EFA provider issue. I will debug.

```
server_command: FI_LOG_LEVEL=warn /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr11207/install/fabtests/bin/fi_recv_cancel -e rdm -V -p efa -E=9228

client_command: FI_LOG_LEVEL=warn /home/ec2-user/PortaFiducia/build/libraries/libfabric/pr11207/install/fabtests/bin/fi_recv_cancel -e rdm -V -p efa -E=9228 172.31.49.147
client_stdout:

client returncode: 124
server_stdout:
ERROR: no error CQ entry in cq_read deteceted(): functional/recv_cancel.c:112, ret=-256 (Unspecified error)

server returncode: 1
```

aingerson:
@sunkuamzn no need to debug. I missed an incorrect container and all providers (using the util srx) were failing the recv_cancel. Repushed and should be good now.

sunkuamzn:
bot:aws:retest

sunkuamzn:
AWS CI failure is unrelated

bot:aws:retest

sunkuamzn (approved):
Thanks!

PR #11209, opened 2025-07-11
configure: Fix clang checking

### LINKED ISSUE #11138: autoconf: incorrect linker detection
**Describe the bug**
In configure.ac, there is logic that detects whether we are linking with clang:
```
dnl patching ${archive_cmds} to affect generation of file "libtool" to fix linking with clang
dnl allows for building with clang lto and fast linking with non gnu ld
dnl --> checking if the linker (clang) is GNU ld... no
AS_CASE(["$LD"],
    [*clang*],
        [AS_CASE(["${host_os}"],
            [*linux*],
            [archive_cmds='$CC -shared $pic_flag $libobjs $deplibs $compiler_flags $wl-soname $wl$soname $wl-version-script $wl$libname.map -o $lib'])])
```

This logic falsely detects a clang linker if we are actually linking with a binutils (gcc) linker that is placed in a path that contains the string "clang".  This is very common in build environments that allow building packages with different compilers, like the Spack package manager. 

For example, when the gcc+binutils package itself has been built with clang compiler, it is placed in the following directories:
```
/spack/opt/spack/linux-rocky9-sapphirerapids/clang-19.1.7/binutils-2.43.1-oe5enuq4taykhjxfpmu5rnwsprmgzd5g
/spack/opt/spack/linux-rocky9-sapphirerapids/clang-19.1.7/gcc-13.3.0-zoi45onqsdek22ic7gs5fduygy7gi4py
```

Because these directory names contain the name "clang", we are detecting a clang linker, when in fact we are using the gnu linker that was built by clang. 

This breaks the build, because the gnu linker reports duplicate version tags as an error. The above script adds the argument ` -Wl,-soname -Wl,libfabric.so.1 -Wl,-version-script -Wl,libfabric.map` to the link/archive flags, although we already add them in Makefile.am. Passing the version scripts twice to the linker invocation results in duplicate version. I guess the clang linker is fine with that and only prints a warning since [#be82757](https://reviews.llvm.org/D64549?id=209163).

**To Reproduce**
Put the linker/compiler in a path containing the text "clang".

**Expected behavior**
Successfull build.

**Output**
Unsuccessfull build.

**Environment:**
Spack package manager, Rocky 9.

j-xiong:
When checking if clang is being used for linking, the pattern `*clang*` may incorrectly match a non-clang loader that is located under a path containing `clang`.

Fix by making sure the pattern only matches the filename part of a path.

Fix #11138

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1117,7 +1117,7 @@ dnl patching ${archive_cmds} to affect generation of file "libtool" to fix linki
 dnl allows for building with clang lto and fast linking with non gnu ld
 dnl --> checking if the linker (clang) is GNU ld... no
 AS_CASE(["$LD"],
-    [*clang*],
+    [*clang[^/]*],
         [AS_CASE(["${host_os}"],
             [*linux*],
             [archive_cmds='$CC -shared $pic_flag $libobjs $deplibs $compiler_flags $wl-soname $wl$soname $wl-version-script $wl$libname.map -o $lib'])])


j-xiong:
bot:aws:retest

j-xiong:
CI failure is unrelated.

PR #11216, opened 2025-07-14
prov/efa: Use dmabuf by default for neuron mr

shijin-aws:
This patch reverts an earlier change that aims to do dmabuf reg only when FI_MR_DMABUF flag is set.
We still want to use dmabuf by default even when this flag is not set.

It also adds the clean up of the dmabuf fd after the ibv_reg_dmabuf_fd to fix the fd leak.

This reverts commit 043e968534491319d1b9afc22e482783d5779677.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -509,6 +509,11 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
 static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr *mr_attr,
 					int access, const uint64_t flags)
 {
+	int dmabuf_fd;
+	uint64_t offset;
+	int ret;
+	struct ibv_mr *ibv_mr;
+
 	if (flags & FI_MR_DMABUF)
 		return efa_mr_reg_ibv_dmabuf_mr(
 			efa_mr->domain->ibv_pd,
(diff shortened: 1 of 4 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -519,32 +524,56 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
 		if (ret != FI_SUCCESS) {
 			EFA_WARN(FI_LOG_MR, "Unable to get dmabuf fd for Gaudi device buffer \n");
 			return NULL;
 		}
+		ibv_mr = efa_mr_reg_ibv_dmabuf_mr(
+				efa_mr->domain->ibv_pd, offset,
+				mr_attr->mr_iov->iov_len,
+				(uint64_t)mr_attr->mr_iov->iov_base,
+				dmabuf_fd, access);
+		(void) ofi_hmem_put_dmabuf_fd(efa_mr->peer.iface, dmabuf_fd);
```
jiaxiyan: Do you need to implement `synapseai_put_dmabuf_fd`?

prov/efa/src/efa_mr.c
```diff
@@ -519,32 +524,56 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+			/* Success => invoke ibv_reg_dmabuf_mr */
+			ibv_mr = efa_mr_reg_ibv_dmabuf_mr(
+					efa_mr->domain->ibv_pd, 0,
 					mr_attr->mr_iov->iov_len,
 					(uint64_t)mr_attr->mr_iov->iov_base,
 					dmabuf_fd, access);
+			(void) ofi_hmem_put_dmabuf_fd(efa_mr->peer.iface, dmabuf_fd);
+			return ibv_mr;
+		} else if (ret == -FI_EOPNOTSUPP) {
+			/* Protocol not availabe => fallback */
```
jiaxiyan: typo `availabe`

prov/efa/src/efa_mr.c
shijin-aws: `synapseai` uses `hcclLookupDMABuff` which I doubt is really opening a new file fd like cuda and neuron. @nbeer-habana do you have context here?

I would make it as a no-op now to avoid breaking things

shijin-aws: I think I should drop the `put` logic for synapseai to not change the behavior

prov/efa/src/efa_mr.c
shijin-aws: Fixed

prov/efa/src/efa_mr.c
shijin-aws: Updated

shijin-aws:
@j-xiong can you review https://github.com/ofiwg/libfabric/pull/11216/commits/884fdfa6e2874b5942abf82f3f11f9dbe4c6d987

j-xiong (approved):

shijin-aws:
bot:aws:retest

shijin-aws:

j-xiong:
There seem to some configure issue with the CI nodes, restarting the tests.

shijin-aws:
@j-xiong Is the test still failing? I'd merge it now if the failure is unrelated?

j-xiong:
The failure is with `fi_multinode -x rma -n 3 -p "verbs;ofi_rxm"`, which is an intermittent known issue. Can be ignored.

PR #11217, opened 2025-07-14
prov/efa: Bypass rdma-core in data path.

shijin-aws:
This patch allows libfabric to bypass rdma-core API in data path to reduce layering overhead and cache misses. It starts from making libfabric call internal ops that is the equivalent of the rdma-core API. Further optimization is needed to get rid of the rdma-core API workflow restriction and implement the logic directly for the transmission and cq poll calls.


prov/efa/src/efa_cqdirect_internal.h
```diff
@@ -161,14 +161,14 @@ MAYBE_INLINE void efa_cqdirect_process_ex_cqe(struct efa_ibv_cq *ibv_cq, struct
 
 MAYBE_INLINE void efa_wq_put_wrid_idx(struct efa_cqdirect_wq *wq, uint32_t wrid_idx)
 {
-	// pthread_spin_lock(&wq->wqlock);
+	ofi_genlock_lock(wq->wqlock);
```
yehuday: looks like your'e missing a lock at `efa_wq_get_next_wrid_idx()`

shijin-aws: It's already locked in libfabric's send/recv/ call as a ep lock, no need to grab a wqlock again in rdma-core. Such ep lock was missing in libfabric 's cq poll, so I only need to add that wqlock in the cq end poll

shijin-aws: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_msg.c#L63

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

prov/efa/src/efa_base_ep.c
```diff
@@ -7,6 +7,22 @@
 #include "efa_cq.h"
 #include "efa_cntr.h"
 #include "rdm/efa_rdm_protocol.h"
+#include "efa_cqdirect.h"
+
+static void efa_qp_fill_ibv_pfns(struct efa_qp *qp)
```
sunkuamzn: What does `pfn` mean?

man/fi_efa.7.md
```diff
@@ -521,6 +521,11 @@ Setting this environment variable to 0 can disable this feature.
 When the number of internal rx pkts to post is lower than this threshold,
 the refill will be skipped.
 
+*FI_EFA_DIRECT_CQ_OPS*
+
+: When true (Default: False), avoid using rdma-core function calls during CQ submit and
+completion polling, and instead directly operate on the CQ memory.
```
sunkuamzn: Also mention TX and RX submissions and send/receive queue buffers

include/windows/config.h
```diff
@@ -43,6 +43,9 @@
 /* Indicates if ibv_reg_dmabuf_mr verbs is available */
 #define HAVE_EFA_DMABUF_MR 0
 
+/* Indicates if efa cq direct is available */
+#define HAVE_EFA_CQ_DIRECT 0
```
sunkuamzn: "cq direct" doesn't capture the change in the QP behavior

I asked Claude and it came up with efa-bypass, efa-raw, efa-native, efa-core, efa-fast, efa-mmio and others

efa-bypass and efa-mmio sound best to me

prov/efa/src/efa_cqdirect_structs.h
```diff
@@ -0,0 +1,117 @@
...
+	int phase;
+	int qmask;
+	uint16_t consumed_cnt;
+};
+
+struct efa_cqdirect_rq {
+	/* see efa_rq in rdma-core/providers/efa/efa.h */
+	struct efa_cqdirect_wq wq;
+	uint8_t *buf;
+	// size_t buf_size;
```
sunkuamzn: Is this member required?

prov/efa/src/efa_cqdirect_structs.h
```diff
@@ -0,0 +1,117 @@
...
+#ifndef _EFA_CQDIRECT_STRUCTS_H
+#define _EFA_CQDIRECT_STRUCTS_H
+
+#include "config.h"
+
+#if HAVE_EFA_CQ_DIRECT
+
+#include "efa_cqdirect_efa_io_defs.h"
+#include <asm/types.h>
+// #include <infiniband/efadv.h>
```
sunkuamzn:
```suggestion
```

prov/efa/src/efa_cqdirect_structs.h
```diff
@@ -0,0 +1,117 @@
...
+
+/*
+ * The contents of this file only make sense if we can query rdma-core for QP
+ * and CQ information.
+ */
+
+/*** TODO: adjust CQ_INLINE_MODE and then refactor to remove
```
sunkuamzn: Can you do this now? Remove `MAYBE_INLINE`, `ENTRY_FUN`, `CQ_INLINE_MODE` and combine `efa_cqdirect_internal.h` with this file?

prov/efa/src/efa_cqdirect_internal.h
```diff
@@ -0,0 +1,386 @@
...
+}
+
+MAYBE_INLINE int efa_cqe_is_pending(struct efa_io_cdesc_common *cqe_common,
+				    int phase)
+{
+	return EFA_GET(&cqe_common->flags, EFA_IO_CDESC_COMMON_PHASE) == phase;
+}
+
+MAYBE_INLINE struct efa_io_cdesc_common *
+efa_sub_cq_get_cqe(struct efa_cqdirect_cq *cqd, int entry)
```
sunkuamzn: Could we remove the term `sub_cq`? To avoid confusion with the actual sub CQs that are only present on older hardware which we don't intend to support in this code path

prov/efa/src/efa_cqdirect_structs.h
```diff
@@ -0,0 +1,117 @@
...
+
+#ifndef _EFA_CQDIRECT_STRUCTS_H
+#define _EFA_CQDIRECT_STRUCTS_H
+
+#include "config.h"
+
+#if HAVE_EFA_CQ_DIRECT
+
+#include "efa_cqdirect_efa_io_defs.h"
+#include <asm/types.h>
```
sunkuamzn: Is this include actually needed? I don't see those types being used

prov/efa/src/efa_cqdirect_structs.h
```diff
@@ -0,0 +1,117 @@
...
+#endif
+
+struct efa_cqdirect_wq {
+	/* see `struct efa_wq` in rdma-core/providers/efa/efa.h */
+
+	uint64_t *wrid;
+	/* wrid_idx_pool: Pool of free indexes in the wrid array, used to select
+	 * the wrid entry to be used to hold the next tx packet's context. At
+	 * init time, entry N will hold value N, as OOO tx-completions arrive,
+	 * the value stored in a given entry might not equal the entry's index.
```
sunkuamzn:
```suggestion
	 * init time, entry N will hold value N. As OOO tx-completions arrive,
	 * the value stored in a given entry might not equal the entry's index.
```

prov/efa/src/efa_cqdirect_entry.h
```diff
@@ -0,0 +1,413 @@
...
+{
+	/* see rdma-core/providers/efa/verbs.c: efa_post_recv */
+	uintptr_t addr;
+	struct efa_io_rx_desc *rx_buf;
+	struct efa_cqdirect_wq *wq = &qp->cqdirect_qp.rq.wq;
+	uint32_t rq_desc_offset;
+	uint32_t i;
+	int err = 0;
+
+	// TODO: unlike rdma-core, this method directly writes to the
```
sunkuamzn: remove comment?

prov/efa/src/efa_cqdirect_internal.h
```diff
@@ -0,0 +1,386 @@
...
+		return NULL;
+
+	// TODO: how is ibvqpx->wr_flags ever set?
+	err = efa_post_send_validate(qp, ibvqpx->wr_flags);
+	if (OFI_UNLIKELY(err)) {
+		cqd_qp->wr_session_err = err;
+		return NULL;
+	}
+
+	/* MODIFIED: if any are pending, copy out the previous one first: */
```
sunkuamzn: what does this mean?

prov/efa/src/efa_cqdirect.c
```diff
@@ -0,0 +1,139 @@
...
+	direct_qp->sq.wq.db = sq_attr.doorbell;
+	direct_qp->sq.num_wqe_pending = 0;
+
+	direct_qp->sq.wq.wqe_size = sq_attr.entry_size;
+	efa_cqdirect_wq_initialize(&direct_qp->sq.wq, sq_attr.num_entries,
+				   &base_ep->util_ep.lock);
+
+	/* see efa_qp_init_indices */
+
+	efa_qp->cqdirect_enabled = 1;
```
sunkuamzn:
```suggestion
	efa_qp->cqdirect_enabled = true;
```

prov/efa/src/efa_cq.c
```diff
@@ -459,6 +580,10 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
 	(*cq_fid)->fid.ops = &efa_cq_fi_ops;
 	(*cq_fid)->ops = &efa_cq_ops;
 
+#if HAVE_EFA_CQ_DIRECT
+	efa_cqdirect_cq_initialize( cq );
```
sunkuamzn:
```suggestion
	efa_cqdirect_cq_initialize(cq);
```

prov/efa/src/efa_cqdirect.c
```diff
@@ -0,0 +1,139 @@
...
+	efa_cqdirect_wq_finalize(&direct_qp->rq.wq);
+}
+
+int efa_cqdirect_cq_initialize(struct efa_cq *efa_cq)
+{
+	struct efadv_cq_attr attr = {0};
+	struct efa_cqdirect_cq *cqdirect = &efa_cq->ibv_cq.cqdirect;
+	int ret;
+
+	memset(cqdirect, 0, sizeof(*cqdirect));
```
sunkuamzn: move the `memset` after the env and hardware check?

prov/efa/src/efa_cqdirect_efa_io_defs.h
```diff
@@ -0,0 +1,543 @@
...
+   Currently we lack writel macros that universally guarantee MMIO
+   writes happen in order, like the kernel does. Even worse many
+   providers haphazardly open code writes to MMIO memory omitting even
+   volatile.
+
+   Until this can be fixed with a proper writel macro, this barrier
+   is a stand in to indicate places where MMIO writes should be switched
+   to some future writel.
+*/
+#define mmio_ordered_writes_hack() mmio_flush_writes()
```
sunkuamzn: `mmio_ordered_writes_hack`, `mmio_wc_spinlock` and `mmio_wc_spinunlock` appear to be unused. Should we commit them?

prov/efa/src/efa_cqdirect_efa_io_defs.h
```diff
@@ -0,0 +1,543 @@
...
+#define EFA_GET(ptr, mask) FIELD_GET(mask##_MASK, *(ptr))
+
+#define EFA_SET(ptr, mask, value)                       \
+	({                                              \
+		typeof(ptr) _ptr = ptr;                 \
+		*_ptr = (*_ptr & ~(mask##_MASK)) |      \
+			FIELD_PREP(mask##_MASK, value); \
+	})
+
+#define min(a, b) ((a) < (b) ? (a) : (b));
```
sunkuamzn: Libfabric already defines min https://github.com/ofiwg/libfabric/blob/2148f26f4974934619de56013f7f761e7d455210/include/ofi.h#L206-L211

prov/efa/src/efa_cqdirect_efa_io_defs.h
```diff
@@ -0,0 +1,543 @@
...
+#include "config.h"
+
+#if HAVE_EFA_CQ_DIRECT
+
+#define EFA_IO_TX_DESC_NUM_BUFS	       2
+#define EFA_IO_TX_DESC_NUM_RDMA_BUFS   1
+#define EFA_IO_TX_DESC_INLINE_MAX_SIZE 32
+#define EFA_IO_TX_DESC_IMM_DATA_SIZE   4
+
+enum efa_io_queue_type {
```
sunkuamzn: These definitions are copied from rdma-core. How can we make sure that the definitions don't drift apart over time?

prov/efa/src/efa_cqdirect_structs.h
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	uint32_t *wrid_idx_pool;
+	uint32_t wqe_cnt;
+	uint32_t wqe_size;
+	uint32_t wqe_posted;
+	uint32_t wqe_completed;
+	uint16_t pc; /* Producer counter */
```
sunkuamzn: What is the producer counter? Could you add a bit more to the comment?

prov/efa/src/efa_cqdirect_entry.h
```diff
@@ -0,0 +1,413 @@
...
+					   uint32_t remote_qpn,
+					   uint32_t remote_qkey)
+{
+	/* TODO: This is terrible abstraction breakage to get efa_ah using
+	 * container_of!!!!*/
+	/* Fixing this requires further refactor on the send path's abstraction
+	 * to move the branching level from rdma-core API to the whole wqe
+	 * preparation level, so we do not need to derive ah from ibvah
+	 */
+	struct efa_ah {
```
sunkuamzn: We already have a `struct efa_ah` in `efa_av.c`. Could we rename?

prov/efa/src/efa_cqdirect_entry.h
```diff
@@ -0,0 +1,413 @@
...
+		efa_post_send_sgl(rdma_req->local_mem, sg_list, num_sge);
+		break;
+	default:
+		return;
+	}
+
+	tx_wqe->meta.length = num_sge;
+}
+
+ENTRY_FUN void efa_cqdirect_wr_set_ud_addr(struct efa_qp *efaqp,
```
sunkuamzn: Appears unused. Function is also using some rdma-core internal structs, so we should probably rewrite it.

prov/efa/src/efa_cqdirect_internal.h
```diff
@@ -0,0 +1,386 @@
...
+	}
+}
+
+MAYBE_INLINE int efa_post_send_validate(struct efa_qp *qp,
+					unsigned int wr_flags)
+{
+	if (OFI_UNLIKELY(qp->cqdirect_qp.sq.wq.wqe_posted -
+				 qp->cqdirect_qp.sq.wq.wqe_completed ==
+			 qp->cqdirect_qp.sq.wq.wqe_cnt)) {
+		EFA_DBG(FI_LOG_EP_DATA,
```
sunkuamzn:
```suggestion
		EFA_WARN(FI_LOG_EP_DATA,
```

prov/efa/src/efa_cqdirect_internal.h
```diff
@@ -0,0 +1,386 @@
...
+{
+	struct efa_io_tx_meta_desc *meta_desc;
+
+	meta_desc = &tx_wqe->meta;
+	meta_desc->immediate_data = be32toh(imm_data);
+	EFA_SET(&meta_desc->ctrl1, EFA_IO_TX_META_DESC_HAS_IMM, 1);
+}
+
+MAYBE_INLINE size_t
+efa_buf_list_total_bytes(const struct ibv_data_buf *buf_list, size_t num_buf)
```
sunkuamzn: Unused

prov/efa/src/efa_env.c
```diff
@@ -39,6 +39,7 @@ struct efa_env efa_env = {
 	.huge_page_setting = EFA_ENV_HUGE_PAGE_UNSPEC,
 	.use_unsolicited_write_recv = 1,
 	.internal_rx_refill_threshold = 8,
+	.use_direct_cq_ops = 1,
```
sunkuamzn:
```suggestion
	.use_direct_cq_ops = true,
```

prov/efa/src/efa_base_ep.h
```diff
@@ -46,8 +47,76 @@ struct efa_qp {
 	struct efa_base_ep *base_ep;
 	uint32_t qp_num;
 	uint32_t qkey;
+
+	bool cqdirect_enabled;
+#if HAVE_EFA_CQ_DIRECT
```
sunkuamzn: I'm wondering if having different memory layouts would make it harder to debug memory corruption or performance issues. We don't have structs like this right now.

The function pointers are also unnecessary if `HAVE_EFA_CQ_DIRECT` is false.

prov/efa/test/efa_unit_tests.c
```diff
@@ -113,6 +113,7 @@ int main(void)
 		cmocka_unit_test_setup_teardown(test_av_implicit_to_explicit, efa_unit_test_mocks_setup, efa_unit_test_mocks_teardown),
 		/* end efa_unit_test_av.c */
 
+		/* end efa_unit_test_ep.c */
```
sunkuamzn:
```suggestion
		/* begin efa_unit_test_ep.c */
```

prov/efa/src/efa_base_ep.h
darrylabbate: Adding onto this, I would say these conditional members should be placed at the tail of the struct

prov/efa/src/efa_base_ep.c
shijin-aws: pointer of functions... I can rename it

prov/efa/src/efa_base_ep.h
shijin-aws: >The function pointers are also unnecessary if HAVE_EFA_CQ_DIRECT is false.

If I don't use function table, I would need to either introduce a wrapper for every ibv call and have `if` `else` in runtime, or I introduce the `#if` branching everywhere of the callers. The former can introduce extra instructions in fast path. the latter one makes the code messy and hard to maintain & unit-testing (mocking will be more troublesome)

sunkuamzn: I understand the need for function pointers when `HAVE_EFA_CQ_DIRECT` is true

What I'm trying to say is: if `HAVE_EFA_CQ_DIRECT` is false, then the function pointer can only point to a single function. So they're not particularly useful.

include/windows/config.h
YonatanNachum: maybe efa_data_path_direct/ efa_data_path_bypass since its not just CQs, its also QPs

prov/efa/src/efa_cqdirect_entry.h
```diff
@@ -0,0 +1,413 @@
...
+	efa_domain = container_of(efa_cq->util_cq.domain, struct efa_domain,
+				  util_domain);
+
+	cqdirect->cur_cqe =
+		efa_cqdirect_next_sub_cqe_get(cqdirect);
+	if (!cqdirect->cur_cqe)
+		return ENOENT;
+
+	qpn = cqdirect->cur_cqe->qp_num;
+	cqdirect->cur_qp =
```
YonatanNachum: how do you synchronize the qp_table ? how do you know there is no destroy QP in progress removing the same QP in parallel ?

include/windows/config.h
YonatanNachum: and in general the cqdirect prefix to all functions (even for bypassed QPs) feels off in my opinion.

prov/efa/src/efa_env.c
```diff
@@ -215,6 +218,12 @@ void efa_env_define()
 			"Use device's unsolicited write recv functionality when it's available. (Default: true)");
 	fi_param_define(&efa_prov, "internal_rx_refill_threshold", FI_PARAM_SIZE_T,
 			"The threshold that EFA provider will refill the internal rx pkt pool. (Default: %zu)", efa_env.internal_rx_refill_threshold);
+	fi_param_define(
+		&efa_prov, "use_direct_cq_ops", FI_PARAM_BOOL,
+		"Use the direct CQ implementation that bypasses rdma-core on "
+		"data path, when it's available. Setting this variable as 0 "
+		"will disable this feature(Default: %d)",
```
YonatanNachum: I think you need to explain the synchronization module you enforce for things to work, you can't post WR and poll CQ a the same time, you can't destroy QP while polling CQ, those things can't be implicit they must be explicit so user can choose this option knowingly and not handle weird errors from race conditions.

darrylabbate:
Wrt the function pointer approach, have we considered simply having the objects point to a table of function pointers (similar to e.g. `struct fid_cq`), rather than each object keeping individual pointers for every single function? Do we need that level of granularity otherwise?

prov/efa/src/efa_cqdirect_entry.h
shijin-aws: It is protected by the ep_list_lock outside the efa_cq_poll_ibv_cq

prov/efa/src/efa_env.c
shijin-aws: See the last comment, this PR already integrate the rdma-core code with libfabric's threading model which helps sure the CQ and WR are accessed in serial. See the threading section in https://ofiwg.github.io/libfabric/main/man/fi_domain.3.html.

 For FI_THREAD_SAFE/FI_THREAD_COMPLETION that application can doing cq and wr in multiple threading, libfabric has a lock to protect them. For FI_THREAD_DOMAIN mode, it's application's responsibility to make sure the CQ and WR post are accessed in serial.

shijin-aws:
> Wrt the function pointer approach, have we considered simply having the objects point to a table of function pointers (similar to e.g. `struct fid_cq`), rather than each object keeping individual pointers for every single function? Do we need that level of granularity otherwise?

That is a very good question. I had that implementation in one revision, then it introduced some headache doing unit test: If you define a global function table like `struct cq_ops` to store the function pointers, if you modify any function pointer in that function table (for mocking purpose), you have to make sure they are recovered after each unit test. Otherwise the global variable will be permanently modified

Also, I am going to move away from the current function pointer approach as rdtsc told me it introduces at least 2 or 3 more CPU cycles every time when libfabric access the real function

include/windows/config.h
shijin-aws: will use `data_path_direct`

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: will remove

prov/efa/src/efa_cqdirect_internal.h
shijin-aws: This is EAGAIN which should not be warning.

prov/efa/src/efa_base_ep.c
shijin-aws: Removed these functions as we are not using function pointers

include/windows/config.h
shijin-aws: Updated

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: removed

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: removed

prov/efa/src/efa_base_ep.h
shijin-aws: This is not relevant as I removed function pointers

man/fi_efa.7.md
shijin-aws: Done

shijin-aws:
@sunkuamzn I missed some of your earlier comments as they were hidden by github... Just saw them now and I will address them tmr with another update

prov/efa/src/efa_cqdirect_entry.h
shijin-aws: I fixed this hacking in the latest push

prov/efa/src/efa_cq.c
shijin-aws: fixed

prov/efa/src/efa_cqdirect.c
shijin-aws: fixed

prov/efa/src/efa_cqdirect_internal.h
shijin-aws: That means the wqe is NOT copied to the device memory until either next ibv_wr_send or ibv_wr_complete.. This is silly and should be fixed after we are not restricted by ibv wr API workflow in the send calls

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: AI helped me rephrased this comments and the issue has been fixed.

prov/efa/src/efa_cqdirect_entry.h
shijin-aws: I will remove it. This is the implementation we want anyway.

prov/efa/src/efa_cqdirect.c
shijin-aws: The caller of this function did not zero this part of memory. So I think even in the error handling path, we want that zero as well. So I tend to keep this behavior

prov/efa/src/efa_cqdirect_efa_io_defs.h
shijin-aws: Just checked the rdma-core source, `mmio_ordered_writes_hack` was used by another nic provider, mthca, only

`mmio_wc_spinlock` and `mmio_wc_spinunlock` are used by efa to lock the WQ, which is already removed in this PR as the same lock was acquired by libfabric's EP lock. I will remove all of them

prov/efa/src/efa_cqdirect_entry.h
shijin-aws: `efa_*_wr_set_ud_addr` should be used ... the internal rdma-core structs issue has been addressed (efa_ah)

prov/efa/src/efa_cqdirect_internal.h
shijin-aws: removed

prov/efa/src/efa_env.c
shijin-aws: Done

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: I changed MAYBE_INLINE to `ALWAYS_INLINE` to reflect what it really means. I deleted `ENTRY_FUN` and declare the functions with `static inline` directly, I removed `CQ_INLINE_MODE`.

I still keep the the structs.h as a separate file as a well defined boundary between structs and function implementation. Some code that include structs.h only need the structs

prov/efa/src/efa_cqdirect_internal.h
shijin-aws: I rename it as `device_cq` . I have to keep the `efa_device_use_sub_cq` name unchanged because this is what the check is: we do not want to do bypass for platform that still use sub cq

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: It is the indexing of WQE in the device memory. Added the comment

prov/efa/test/efa_unit_tests.c
shijin-aws: I finally drop the reorganization of unit-test items in this PR, will do it separately later

prov/efa/src/efa_cqdirect_efa_io_defs.h
shijin-aws: good catch, removed and use `MIN`

prov/efa/src/efa_cqdirect_efa_io_defs.h
shijin-aws: I removed these macro, as we are not using them in the code any more. These were used to invalidate the WQE that exceeds the length limit, which is already done by libfabric on checking the inline data size and cqdata size

prov/efa/src/efa_cqdirect_structs.h
shijin-aws: it's not needed, removed

prov/efa/configure.m4
```diff
@@ -226,6 +226,12 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	AC_DEFINE_UNQUOTED([HAVE_EFADV_QUERY_CQ],
 		[$have_efadv_query_cq],
 		[Indicates if efadv_query_cq is available])
+	AS_IF([test "$have_efadv_query_qp_wqs" = "1" -a "$have_efadv_query_cq" = "1"],
+		[have_efa_data_path_direct=1],
+		[have_efa_data_path_direct=0])
+	AC_DEFINE_UNQUOTED([HAVE_EFA_DATA_PATH_DIRECT],
+		[$have_efa_data_path_direct],
+		[Indicates if CQ Direct is available (requires both QUERY_QP_WQS and QUERY_CQ)])
```
sunkuamzn: nit: data path direct

prov/efa/src/efa_data_path_ops.h
```diff
@@ -0,0 +1,330 @@
...
+static inline int efa_qp_wr_complete(struct efa_qp *efaqp)
+{
+#if HAVE_EFA_DATA_PATH_DIRECT
+	if (efaqp->data_path_direct_enabled)
+		return efa_data_path_direct_wr_complete(efaqp);
+#endif
+	return ibv_wr_complete(efaqp->ibv_qp_ex);
+}
+
+static inline void efa_qp_wr_rdma_read(struct efa_qp *efaqp, uint32_t rkey, uint64_t remote_addr)
```
sunkuamzn, line 83: Why not force inline these functions and `efa_data_path_direct_*` too? They are in the critical path

prov/efa/src/efa_cq.h
```diff
@@ -14,6 +15,12 @@ enum ibv_cq_ex_type {
 struct efa_ibv_cq {
 	struct ibv_cq_ex *ibv_cq_ex;
 	enum ibv_cq_ex_type ibv_cq_ex_type;
+	struct efa_ibv_cq_ops *ops;
```
sunkuamzn: Couldn't figure out where this struct is defined or what this member is used for

prov/efa/src/efa_data_path_direct.c
```diff
@@ -0,0 +1,206 @@
...
+
+	direct_qp->sq.wq.wqe_size = sq_attr.entry_size; /* Entry size */
+	/* Initialize send work queue management structures */
+	efa_data_path_direct_wq_initialize(&direct_qp->sq.wq, sq_attr.num_entries,
+			   &base_ep->util_ep.lock);
+
+	/* Mirror efa_qp_init_indices functionality - indices already initialized */
+
+	/* Mark direct data path operations as enabled for this queue pair */
+	efa_qp->data_path_direct_enabled = 1;
```
sunkuamzn: nit: I think we use `true` and `false` for bools

shijin-aws:
bot:aws:retest

prov/efa/src/efa_data_path_direct_entry.h
```diff
@@ -0,0 +1,557 @@
...
+			addr = wr->sg_list[i].addr;
+
+			/* Set RX buffer desc from SGE */
+			rx_buf->length = MIN(wr->sg_list[i].length, UINT16_MAX);
+			EFA_SET(&rx_buf->lkey_ctrl, EFA_IO_RX_DESC_LKEY,
+			wr->sg_list[i].lkey);
+			rx_buf->buf_addr_lo = addr;
+			rx_buf->buf_addr_hi = (uint64_t)addr >> 32;
+
+			/* reset descriptor for next iov */
```
sunkuamzn: remove this comment?

prov/efa/src/efa_env.c
```diff
@@ -136,6 +137,8 @@ void efa_env_param_get(void)
 	if (fi_param_get_bool(&efa_prov, "use_huge_page", &use_huge_page) ==0) {
 		efa_env.huge_page_setting = use_huge_page ? EFA_ENV_HUGE_PAGE_ENABLED : EFA_ENV_HUGE_PAGE_DISABLED;
 	}
+	fi_param_get_size_t(&efa_prov, "internal_rx_refill_threshold", &efa_env.internal_rx_refill_threshold);
```
sunkuamzn:
```suggestion
```

copy of line 113

prov/efa/src/efa_data_path_direct_internal.h
```diff
@@ -0,0 +1,632 @@
...
+	wq->wqe_posted++; /* Increment posted work request count */
+	wq->pc++; /* Advance producer counter */
+
+	/* Check for queue wraparound and advance phase if needed */
+	if (!(wq->pc & wq->desc_mask))
+		wq->phase++;
+}
+
+EFA_ALWAYS_INLINE void
+efa_data_path_direct_send_wr_set_imm_data(struct efa_io_tx_wqe *tx_wqe,
```
sunkuamzn: unused. Copy of `efa_send_wr_set_imm_data` below

sunkuamzn (dismissed):

sunkuamzn:
bot:aws:retest

prov/efa/src/efa_cq.h
shijin-aws: I failed to clean it up, thanks for catching it

prov/efa/src/efa_env.c
shijin-aws: removed

prov/efa/src/efa_data_path_direct.c
shijin-aws: fixed

prov/efa/src/efa_data_path_direct_internal.h
shijin-aws: God, removed.

prov/efa/src/efa_data_path_ops.h
shijin-aws: Good call, need some more profiling to show whether forcing inline here is better, can be done in a separate PR.

prov/efa/src/efa_data_path_direct_entry.h
shijin-aws: removed

prov/efa/configure.m4
shijin-aws: fixed

shijin-aws:
I rebased on the https://github.com/ofiwg/libfabric/commit/d22d5aede73e27835abec091fd85fade3e1a8e7d which is the last commit before the broken one https://github.com/ofiwg/libfabric/commit/42801061399075ff9ff4542c750b5c1dbf033d48 (to be reverted by https://github.com/ofiwg/libfabric/pull/11288).

The merge conflict should be resolved after https://github.com/ofiwg/libfabric/pull/11288 is merged

sunkuamzn (dismissed):

man/fi_efa.7.md
```diff
@@ -522,6 +522,11 @@ Setting this environment variable to 0 can disable this feature.
 When the number of internal rx pkts to post is lower than this threshold,
 the refill will be skipped.
 
+*FI_EFA_USE_DATA_PATH_DIRECT*
+
+: Use the direct data path implementation that bypasses rdma-core on data path, including the CQ polling and TX/RX submissions, when it's available.
+Setting this variable as 0 will disable this feature (Default: false).
```
jiaxiyan: This should be updated to 
`Set this variable as 1 to enable this feature ` if the default is false now.

prov/efa/src/efa_cq.c
```diff
@@ -373,6 +382,102 @@ struct fi_ops_cq efa_cq_ops = {
...
+		.wc_flags = EFADV_WC_EX_WITH_SGID,
+	};
+
+#if HAVE_CAPS_UNSOLICITED_WRITE_RECV
+	if (efa_use_unsolicited_write_recv())
+		efadv_cq_init_attr.wc_flags |= EFADV_WC_EX_WITH_IS_UNSOLICITED;
+#endif
+
+#if HAVE_CAPS_CQ_WITH_EXT_MEM_DMABUF
+	efadv_cq_init_attr.flags = efa_cq_init_attr->flags;
```
jiaxiyan: This looks like it is from my old commit. Can you make it up to date with main to set EFADV_CQ_INIT_FLAGS_EXT_MEM_DMABUF?

prov/efa/src/efa_data_path_direct_internal.h
```diff
@@ -0,0 +1,621 @@
...
+	struct ibv_qp_ex *ibvqpx = qp->ibv_qp_ex;
+	struct efa_data_path_direct_qp *dpd_qp = &qp->data_path_direct_qp;
+	struct efa_data_path_direct_sq *sq = &qp->data_path_direct_qp.sq;
+	struct efa_io_tx_meta_desc *meta_desc;
+	int err;
+
+	if (OFI_UNLIKELY(dpd_qp->wr_session_err))
+		return NULL;
+
+	// TODO: how is ibvqpx->wr_flags ever set?
```
jiaxiyan: Is this done?

prov/efa/src/efa_data_path_direct.c
```diff
@@ -0,0 +1,206 @@
...
+	/**
+	 * Direct CQ operations are disabled if:
+	 * 1. Environment variable disables direct CQ operations
+	 * 2. Hardware is still using sub-completion queues (older firmware)
+	 *
+	 * In these cases, fall back to standard completion processing.
+	 */
+	if (!efa_env.use_data_path_direct || efa_device_use_sub_cq()) {
+		/* Direct CQ not enabled - use standard completion processing */
+		return FI_SUCCESS;
```
jiaxiyan, line 180: Why is FI_SUCCESS returned for fallback?

prov/efa/src/efa_data_path_ops.h
```diff
@@ -0,0 +1,330 @@
...
+	if (ibv_cq->data_path_direct_enabled)
+		return efa_data_path_direct_wc_is_unsolicited(ibv_cq);
+#endif
+#if HAVE_CAPS_UNSOLICITED_WRITE_RECV
+	return efadv_wc_is_unsolicited(efadv_cq_from_ibv_cq_ex(ibv_cq->ibv_cq_ex));
+#else
+	return false;
+#endif
+}
+
```
jiaxiyan, line 297: too many empty lines

prov/efa/src/efa_data_path_ops.h
```diff
@@ -0,0 +1,330 @@
...
+#endif /* EFA_UNIT_TEST */
+
+/**
+ * @brief Check whether a completion consumes recv buffer
+ *
+ * @param ibv_cq efa ibv cq
+ * @return true the wc consumes a recv buffer
+ * @return false the wc doesn't consume a recv buffer
+ */
+static inline bool efa_cq_wc_is_unsolicited(struct efa_ibv_cq *ibv_cq)
```
jiaxiyan, line 321: Why is this function moved here?

prov/efa/src/efa_domain.c
```diff
@@ -492,7 +492,7 @@ static int efa_domain_query_addr(struct fid_ep *ep_fid, fi_addr_t addr,
  * @param rq_attr pointer to receive queue attributes
  * @return 0 on success, negative integer on failure
  */
-static int efa_domain_query_qp_wqs(struct fid_ep *ep_fid,
+int efa_domain_query_qp_wqs(struct fid_ep *ep_fid,
```
jiaxiyan: Why remove the static here?

prov/efa/src/efa_data_path_ops.h
shijin-aws: Because I introduced the wrapper `efa_ibv_cq_wc_is_unsolicited` in this function. If I keep `efa_cq_wc_is_unsolicited` in `efa_cq.h`, I need to include `efa_data_path_ops.h` in `efa_cq.h`, which will cause a cyclic dependency "efa_data_path_ops.h" -> "efa_cq.h" -> "efa_data_path_ops.h",

prov/efa/src/efa_domain.c
shijin-aws: Because `efa_data_path_direct.c` calls this function

shijin-aws: My bad, this was an accident change by luke, he didn't really call this function directly, I will fix it

prov/efa/src/efa_data_path_direct.c
shijin-aws: Because it is intended? When user disable it or device doesn't support it, we shouldn't fail right?

prov/efa/src/efa_data_path_direct_internal.h
shijin-aws: Good catch, it is unused and I will drop this wr_flags in the function call

prov/efa/src/efa_cq.c
shijin-aws: Hmm, not sure how I can make this slipped. I will make a separate commit for this function refactor for you to review again

man/fi_efa.7.md
shijin-aws: fixed

prov/efa/src/efa_data_path_ops.h
shijin-aws: fixed

prov/efa/src/efa_cq.c
shijin-aws: I moved this function back to efa_cq.h for better comparison in https://github.com/ofiwg/libfabric/pull/11217/commits/aef88fb26efaed1d42990dffe678dae8783456ca. I forgot why I moved it to efa_cq.c earlier.

shijin-aws:
rebase only

PR #11218, opened 2025-07-15
configure.ac: Move cuda cppflag set before DMABUF check

zachdworkin:
This check needs to happen before the DMABUF check so that the user doesn't need to set CFLAGS="-I/cuda_install_location" before running configure. This will let the AC_CHECK_DECL properly update CPPFLAGS and LDFLAGS before the DMABUF checks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -755,6 +755,9 @@ AS_IF([test x"$with_cuda" != x"no"],
 				 [have_cuda=1])
 	      ])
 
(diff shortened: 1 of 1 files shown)


zachdworkin:
@shijin-aws @jiaxiyan Can you please share the AWS CI failure?

jiaxiyan:
bot:aws:retest

jiaxiyan:
It is unrelated. I kicked off another run.

zachdworkin:
bot:aws:retest

PR #11237, opened 2025-07-23
prov/efa: Fix the hmem flags setting

shijin-aws:
Currently, efa_prov_info_set_hmem_flags is called as part of device list initialization, which is before efa_hmem_info_initialize() is called. This will make the efa_prov_info_set_hmem_flags not work because the metadata in `g_efa_hmem_info` is not initialized yet.

This patch fixes this issue by moving efa_prov_info_set_hmem_flags to efa_util_prov_initialize()'s efa-direct prov_info construction, which happens after the efa_hmem_info_initialize() call.

### DIFF
--- a/prov/efa/src/efa_prov.c
+++ b/prov/efa/src/efa_prov.c
@@ -109,6 +109,9 @@ static int efa_util_prov_initialize()
 			continue;
 		}
 
+		/* This function must be called after efa_hmem_info_initialize() */
+		efa_prov_info_direct_set_hmem_flags(prov_info_direct);
+
 		if (!head) {
 			head = prov_info_direct;
 		} else {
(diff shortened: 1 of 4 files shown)


darrylabbate (dismissed):

prov/efa/src/efa_prov.c
```diff
@@ -109,6 +109,9 @@ static int efa_util_prov_initialize()
 			continue;
 		}
 
+		/* This function must be called after efa_hmem_info_initialize() */
+		efa_prov_info_set_hmem_flags(prov_info_direct, FI_EP_RDM);
```
sunkuamzn: probably nit: If we're only going to call this function here, can we rename it to `efa_prov_info_direct_set_hmem_flags` and remove the `ep_type` argument?

shijin-aws: Updated.

sunkuamzn (approved):

shijin-aws:
bot:aws:retest

PR #11241, opened 2025-07-24
core: Add address format FI_SOCKADDR_IP

### LINKED ISSUE #10472: prov/verbs: establishing verbs connection using the GID doesn't work
**Describe the bug**
Attempting to use NICs with IPoIB (via a call to `fi_domain`) disabled with the verbs provider using @sydidelot's feature (#5605) with RxM enabled doesn't work, likely due to RxM's assumption of `dest_addr = FI_SOCKADDR` rather than `FI_SOCKADDR_IB`

**To Reproduce**
 - Use MPICH on a NIC where IPoIB is disabled (with `FI_PROVIDER=verbs,ofi_rxm`)
 - Unfortunately I don't really have a simple reproducer, but the specific call failing is attempting to call `fi_domain` on a NIC using its GID rather than via IPoIB

**Expected behavior**
Replacing [this line](https://github.com/ofiwg/libfabric/blob/0f7cc83b5e4449afa4115b2a402c3ae58fcd099e/prov/rxm/src/rxm_attr.c#L248) with `.addr_format = FI_SOCKADDR_IB` makes everything working as expected (ie, the call to `fi_domain` on a NIC with IPoIB disabled (thus using its GID instead) succeeds.

**Output**
OFI fails with `ENODATA`
```
[3] libfabric:1347797:1729070846:ofi_rxm:verbs:fabric:vrb_get_rai_id():301<warn> rdma_resolve_addr: Invalid argument (22)
[3] libfabric:1347797:1729070846:ofi_rxm:verbs:fabric:vrb_get_rai_id():303<info> src addr: fi_sockaddr_ib://[fe80::88e9:a4ff:ff1c:5860]:0xffff:0x13f:0x0
[3] libfabric:1347797:1729070846:ofi_rxm:verbs:fabric:vrb_get_rai_id():305<info> dst addr: (null)
[3] libfabric:1347797:1729070846:ofi_rxm:verbs:fabric:vrb_get_match_infos():1825<info> handling of the socket address fails - -22
```

**Environment:**
`fi_info`'s relevant output:
```
fi_info:
    caps: [ FI_MULTI_RECV, FI_LOCAL_COMM, FI_REMOTE_COMM, FI_HMEM ]
    mode: [ FI_BUFFERED_RECV ]
    addr_format: FI_SOCKADDR_IB
    src_addrlen: 48
    dest_addrlen: 0
    src_addr: fi_sockaddr_ib://[fe80::88e9:a4ff:ff4a:997c]:0xffff:0x13f:0x0
    dest_addr: (null)
    handle: (nil)
[...]
```

### LINKED ISSUE #11131: prov/rxm: RXM:verbs address format mismatch
**Describe the bug**
It happened in the [fi-mr-reg](https://github.com/ofiwg/libfabric/blob/main/fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c) fabtest with ofi_rxm:verbs + RDM (Though I was rewriting and running this on AMD platform, the failing snippet is basic ofi init before GPU runtime initializes). Ultimately `fi_endpoint()` would fail with ENODATA due to user info and provider info address format mismatch at [`ofi_check_info()`](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_attr.c#L1123)

The sequence of event I observed is as following:

- `fi_getinfo()` would iterate through all providers and their getinfo functions, reaching verbs first before rxm. 
- `vrb_getinfo()`->`vrb_init_info()` initializes `vrb_util_prov.info`. During this process, `ep_type` for [`vrb_alloc_info()`](https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L764) is evaluated to be MSG, and address format of the allocated info is SOCKADDR_IB. 
- verbs would have `vrb_util_prov.info` initialized, but not return any info. Afterwards, the provider loop will continue and ultimately reaches rxm. `rxm_getinfo()` would invoke `ofix_getinfo()`, which starts yet another loop over underlying providers. 
- Inside `ofix_getinfo()`, verbs now will return the `vrb_util_prov.info` as [core_info](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_attr.c#L353), which in turn becomes util_info through `ofi_info_to_util()`, which ultimately becomes the info returned.
- With this info returned from `fi_getinfo()` as user_info, down the line, `fi_endpoint()` calls `rxm_endpoint()` -> `ofi_endpoint_init()` -> `ofi_prov_check_info()` -> `ofi_check_info(..., prov_info, ..., user_info)`. There, [ofi_valid_addr_format](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_attr.c#L1137) returns false because user_format is `FI_SOCKADDR_IB`, and endpoint would end with ENODATA.

Along this path, there are a few factors leading to this, 
1. Inside verbs_info seems only MSG(XRC) and DGRAM are accepted and considered, any info initialized should only have either of them as ep type. 
2. rxm after obtaining the info from verbs, inside `rxm_getinfo()`->`ofix_getinfo()`, before the info is bound to util_info and returned, [`ofi_alter_info()`](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_attr.c#L371) is supposed to copy values from hint to the util_info's field. However, the routine only changes ep/domain attribute, rx/tx, etc., and does not change address format according to the ep_type. I saw [this comment](https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_info.c#L1810) before verbs version of matching info routine, so not sure if changing address format is also a thing of consideration.
3. Lastly, I see it's mentioned in  #10472 that FI_SOCKADDR should support sockaddr_ib, but address format validation [routine](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_attr.c#L57) prevents that atm, with sockaddr_ib being the largest.

I'm not sure which one from above is expected and which one should be fixed. Please let me know if anyone has any thoughts, I'd be happy to assist.

**To Reproduce**
It's like the fi-mr-reg test, except we call [`init_ofi()`](https://github.com/ofiwg/libfabric/blob/main/fabtests/component/dmabuf-rdma/fi-mr-reg-xe.c#L96) first in main, so basically just 
```
...
fi_getinfo(version, NULL, NULL, 0, hints, &fi);
...
fi_fabric(fi->fabric_attr, &fabric, NULL));
fi_domain(fabric, fi, &domain, NULL);
fi_endpoint(domain, fi, &ep, NULL); <- fi_endpoint(domain, fi, &ep, NULL) returned error -61
```

**Output**
If needed I can also provide the full FI_LOG, as well as any additional details/context

**Environment:**
OS: Ubuntu 22.04.2,
provider: rxm:verbs, 
endpoint type: RDM,
domain: mellanox connectx-7 (OFED-5.8-1.0.1)

**Additional context**
The same exact code and executable used to work months ago on the very platform where it crashed, everything other than OFED driver was the same _to my knowledge_, so there might be additional factors I overlooked. 
But regardless I believe the procedures I described above is still true and relevant, which could lead to problem. Please correct me if I made any mistake anywhere.

j-xiong:
Currently `FI_SOCKADDR` is interpreted inconsistently. By definition, it covers any address format derived form `sockaddr` which should include `sockaddr_ib`. In provider implementation and capability checks it is used for the union of IPv4 and IPv6 and in that sense `sockaddr_ib` is excluded. This has caused confusion to both provider writers and end users and also led to bugs such as IB address cannot be used with the rxm/verbs provider. 

The first patch introduces a new address format `FI_SOCKADDR_IP` which covers IPv4 and IPv6 and keeps `FI_SOCKADDR` true to its definition.

The second patch changes affected providers to use / handle the new address format.

Fixes #11131 
Fixes #10472


prov/util/src/util_attr.c
```diff
@@ -60,17 +64,15 @@ int ofi_valid_addr_format(uint32_t prov_format, uint32_t user_format)
...
 	case FI_SOCKADDR_IN:
 		/* Provider supports INET only */
 		return user_format <= FI_SOCKADDR_IN;
 	case FI_SOCKADDR_IN6:
 		/* Provider supports INET6 only */
 		return user_format <= FI_SOCKADDR_IN6;
 	case FI_SOCKADDR_IB:
-		/* Provider must support IB, INET, and INET6 */
+	case FI_SOCKADDR:
+		/* Provider supports IB, INET, and INET6 */
```
shefty: There's a difference in how FI_SOCKADDR is treated by the API, versus internally with providers.  The comments are the key here.

I think this breaks udp, tcp, and sockets providers.  When the provider is reporting FI_SOCKADDR, they support IP and IPv6.  This change will now pass IB addressing to udp/tcp/socket providers.

The problem is there's no easy way to indicate support for IP addressing only, so FI_SOCKADDR covers that.  One option could be to add FI_SOCKADDR_IP to cover IN and IN6.  That value could be kept internal only (OFI_SOCKADDR_IP) or exposed.  Either way may work.

While here, the check for FI_SOCKADDR_IN6 is incorrect.  The check would also accept FI_SOCKADDR_IN, but the comment indicates that only IPv6 should be supported.  The correct check would be ```user_format == FI_SOCKADDR || user_format == FI_SOCKADDR_IN6```.  This path isn't ever taken, but might be worth fixing.

j-xiong: I am totally fine with leaving FI_SOCKADDR for only covering INET and INET6 and instead making the single line change to the rxm provider as described in #10472. That should fix both #10472 and #11131 as well. We can explicitly document in the man page for that.

Will fix the checking for FI_SOCKADDR_IN6.

shefty: The more I think about this, I wonder if we shouldn't go ahead and add FI_SOCKADDR_IP.  The API defines FI_SOCKADDR to explicitly cover all sockaddr formats.  We're using a different definition internally, which ends up looking wrong and tripped up both me and you.  (My comments in #10472 and your changes here.)

Then SOCKADDR_IN would mean IPv4 only, SOCKADDR_IN6 IPv6 only, and SOCKADDR_IB IB only, which is consistent.  Today, there's more confusion because SOCKADDR_IN6 means IPv6 only, but SOCKADDR_IB means IP, IPv6, and IB.

j-xiong: Having a clear and consistent definition of the sock address formats is definitely a good thing.  Let me work out a patch for that.

j-xiong: @shefty I have updated the PR with FI_SOCKADDR_IP definition and usage.

src/common.c
```diff
@@ -381,6 +382,7 @@ const char *ofi_straddr(char *buf, size_t *len,
 
 	switch (addr_format) {
 	case FI_SOCKADDR:
+	case FI_SOCKADDR_IP:
```
shefty, line 397: This flow follows the current model where SOCKADDR excludes SOCKADDR_IB.  The SOCKADDR case should be expanded to check for IB

prov/ucx/src/ucx_init.c
```diff
@@ -332,7 +332,8 @@ static int ucx_getinfo(uint32_t version, const char *node,
 		    hints->addr_format == FI_ADDR_UCX ||
 		    hints->addr_format == FI_FORMAT_UNSPEC) {
 			ucx_info.addr_format = FI_ADDR_UCX;
-		} else if (hints->addr_format <= FI_SOCKADDR_IN) {
+		} else if (hints->addr_format <= FI_SOCKADDR_IN ||
+			   hints->addr_format == FI_SOCKADDR_IP) {
```
shefty, line 336: Use is_ipv4addr() ?

prov/util/src/util_attr.c
```diff
@@ -40,15 +40,52 @@
...
 
 	switch (user_format) {
 	case FI_SOCKADDR:
-		/* Provider supports INET and INET6 */
-		return if_format <= FI_SOCKADDR_IN6;
+		/* matches INET, INET6, and IB */
+		return is_sockaddr(if_format);
+	case FI_SOCKADDR_IP:
+		/* matches INET and INET6 */
+		return is_ipaddr(if_format);
```
shefty, line 86: nit: I'd drop the comments now that the semantics are consistent and you added the helper functions which make the checks clear.

Same comment on change below.

src/common.c
j-xiong: good catch, will fix.

prov/ucx/src/ucx_init.c
j-xiong: That requires moving the definitions to `ofi.h`. Is it worth the trouble?

j-xiong: I pushed an update that addressed all comments except this one. If you think we should move the help functions to ofi.h I can do that, otherwise I will leave it as it is.

shefty (dismissed):
thanks

j-xiong:
Added a missing change for the verbs provider. The address format set here needs to cover IPv4/IPv6/IB.
```
diff --git a/prov/verbs/src/verbs_info.c b/prov/verbs/src/verbs_info.c
index 68ad7826d..0e87b359f 100644
--- a/prov/verbs/src/verbs_info.c
+++ b/prov/verbs/src/verbs_info.c
@@ -768,7 +768,7 @@ static int vrb_alloc_info(struct ibv_context *ctx, struct fi_info **info,
                fi->caps = VERBS_MSG_CAPS;
                *(fi->tx_attr) = verbs_tx_attr;
                *(fi->rx_attr) = verbs_rx_attr;
-               fi->addr_format = FI_SOCKADDR_IB;
+               fi->addr_format = FI_SOCKADDR;
                break;
        case FI_EP_DGRAM:
                fi->caps = VERBS_DGRAM_CAPS;
```

j-xiong:
bot:aws:retest

j-xiong:
bot:aws:retest

j-xiong:
@shijin-aws Any details on the AWS failure?

j-xiong:
bot:aws:retest

shijin-aws:
@j-xiong AWS CI was having infra issue, it should be fixed now. just restarted it

shijin-aws:
bot:aws:retest

j-xiong:
bot:aws:retest

j-xiong:
@shijin-aws Still failing. Any more details?

shijin-aws:
@j-xiong one certain test stage failed due to insufficient hardware capacity... Safe to ignore and merge.. will look into why it fails persistently

PR #11243, opened 2025-07-24
examples: add example directory with first example

aingerson:
Add a new examples/ directory which is intended to have standalone, single-file examples that show how to use libfabric and its features. A single Makefile is provided for simplicity in building. Each test will be completely contained in a single file with no cross depdencies and will be thoroughly commented to explain each step in the program.
Each test executable will have the prefix fi_example_

This commit comes with the first example - fi_example_rdm which outlines a simple server-client and send-recv

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,30 @@
+# Libfabric Examples
+
+The examples in this directory provide standalone libfabric examples for
+developers as starting points for writing OFI programs.  Each example includes
+detailed comments that explain every OFI call and its purpose.
+
+A single Makefile is provided for simplicity. Set the environment variable
+`LIBFABRIC_PREFIX` to specify where to find the libfabric library and headers.
+
+## Examples
+
+### fi_example_rdm (rdm.c)
+
+This example uses `FI_EP_RDM` (reliable, unconnected) with a simple server and
+client to execute a one-directional send and receive.
+
+#### Server
+```bash
+FI_PROVIDER=<prov> ./fi_example_rdm
+```
+
+#### Client
+```bash
+FI_PROVIDER=<prov> ./fi_example_rdm <server_addr>
+```
+
+> **Note:** The server and client must use the same provider.  The server
+address provided to the client must match the provider/interface being used.
+Specifying `FI_PROVIDER` is optional; libfabric will select the fastest
+interface available.
--- a/examples/rdm.c
+++ b/examples/rdm.c
@@ -1,0 +1,434 @@
+/*
+ * Copyright (c) Intel Corporation.  All rights reserved.
+ *
+ * This software is available to you under the BSD license
+ * below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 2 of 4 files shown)


j-xiong:
I like the idea.

One suggestion is to use Markdown for the README file which can make it easy to read in a browser. This is especially helpful when more examples are added.

aingerson:
@j-xiong Good call. I've updated it.

examples/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+A single Makefile is provided for simplicity. Set the environment variable
+`LIBFABRIC_PREFIX` to specify where to find the libfabric library and headers.
+
+## Examples
+
+### fi_example_rdm (rdm.c)
+
+This example uses `FI_EP_RDM` (reliable, unconnected) with a simple server and
+client to execute a one-sided send and receive.
```
j-xiong: use `one-directional` or `one-way`.  `one-sided send/receive` is a confusing term.

examples/rdm.c
```diff
@@ -0,0 +1,431 @@
...
+	if (!hints)
+		return EXIT_FAILURE;
+
+	/*
+	 * Server run with no args, client has server's address as an argument
+	 */
+	dst_addr = argv[optind];
+
+
+/* Set anything in hints that the application needs */
```
j-xiong, line 398: Is it intentional or missing a tab?

examples/rdm.c
```diff
@@ -0,0 +1,431 @@
...
+	 * not get generated until our message arrives at the destination.
+	 * Otherwise, the client might get a completion and exit before the
+	 * server receives the message. This is to make the test simpler and
+	 * is not required. Delivery complete may have a performance impact and
+	 * can alternatively be set per data transfer instead of globally like
+	 * we do here.
+	 */
+	hints->tx_attr->op_flags = FI_DELIVERY_COMPLETE;
+
+/* Done setting hints */
```
j-xiong, line 424: same here.

examples/rdm.c
```diff
@@ -0,0 +1,431 @@
...
+	ret = fi_close(&domain->fid);
+	if (ret)
+		printf("warning: error closing domain (%d)\n", ret);
+
+	ret = fi_close(&fabric->fid);
+	if (ret)
+		printf("warning: error closing fabric (%d)\n", ret);
+
+	if (info)
+		fi_freeinfo(info);
```
j-xiong, line 244: hints also needs to be freed.

examples/rdm.c
```diff
@@ -0,0 +1,431 @@
...
+		return ret;
+	}
+
+	/*
+	 * Initialize our address vector.
+	 * Address vectors are used to map higher level addresses, which may
+	 * be more natural for an application to use, into fabric specific
+	 * addresses. An AV_TABLE av will map these addresses to indexed
+	 * addresses, starting with fi_addr 0. These addresses are used in data
+	 * transfer calls to specific which peer to send to/recv from.
```
ooststep: "to specify which peer"

examples/rdm.c
```diff
@@ -0,0 +1,431 @@
...
+	 * Endpoints are transport level communication portals which are used to
+	 * initiate and drive communication.
+	 * There are three main types of endpoints:
+	 * FI_EP_MSG - connected, reliable
+	 * FI_EP_RDM - unconnected, reliable
+	 * FI_EP_DGRAM - unconnected, unreliable
+	 * The type of endpoint will be requested in hints/fi_getinfo.
+	 * Different providers support different types of endpoints.
+	 * In this application, we requested FI_EP_RDM. Both tcp and verbs use a
+	 * connected-oriented protocol underneath but have built-in abstraction
```
ooststep, line 116: tcp does, but verbs needs rxm layering.  too pedantic?

ooststep (commented):
looks good to me!

j-xiong:
bot:aws:retest

aingerson:
These tests aren't being built by default (do we want them to be?) so they shouldn't be relevant to the CIs

j-xiong:
Yes, they are not relevant to the CI failure, and we don't want them to be built by default. Just tried to make the result looks pretty :)

j-xiong (dismissed):

examples/rdm.c
aingerson: I intentionally left it not tabbed so it would stick to highlight where hints are being set as that's an important part of the code. But I can indent it if you prefer?

examples/rdm.c
aingerson: I left it out because it seemed like a little too much to include in the proof of concept test

examples/rdm.c
j-xiong: It's fine. Just want to make sure it is not by mistake.

j-xiong (approved):

PR #11244, opened 2025-07-24
fabtests: fix type mismatch build warning

aingerson:
Fix build error:
argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]

Change num_ep type to size_t to suppress warning

### DIFF
--- a/fabtests/benchmarks/rdm_bw_mt.c
+++ b/fabtests/benchmarks/rdm_bw_mt.c
@@ -74,7 +74,7 @@
 #define BUFFER_SIZE 1024
 static char oob_buffer[BUFFER_SIZE];
 
-static int num_eps = 1;
+static size_t num_eps = 1;
 static bool bidir = false;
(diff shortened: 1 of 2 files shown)


shijin-aws:
@aingerson Thanks Alexia! I encountered that warning as well and I couldn't figure out t hat time. I was copying from https://github.com/ofiwg/libfabric/blob/main/fabtests/functional/multi_ep.c#L63 and I am not sure why that test doesn't raise such warning

aingerson:
@shijin-aws Yeah I'm not sure either. I looked at the rdm_bw_bt to compare usage and didn't see any differences 🤷

j-xiong:
This seems to be a compiler bug based on various discussions searchable over the web. Similar calls (`calloc(num_eps, ...);`) are present in the same file yet only this specific instance (L322) is generating the warning.

PR #11255, opened 2025-07-29
fabtests/efa: Add pytest marker and fixture to GDA fabtest

jiaxiyan:
Add testset short and fabric so it will be picked up by our CI.

### DIFF
--- a/fabtests/pytest/efa/test_gda.py
+++ b/fabtests/pytest/efa/test_gda.py
@@ -3,13 +3,12 @@ import pytest
 from common import ClientServerTest, has_cuda
 
 
+@pytest.mark.short
 @pytest.mark.functional
 @pytest.mark.cuda_memory
-def test_gda_send_recv(cmdline_args, direct_message_size):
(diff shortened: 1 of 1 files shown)


shijin-aws:
> Add testset short and fabric so it will be picked up by our CI.

fabtests require specific flag to build the gda test, right? Not see a reason you have to marker it as short to avoid it picked up

jiaxiyan:
> 
> fabtests require specific flag to build the gda test, right? Not see a reason you have to marker it as short to avoid it picked up

No our CI only runs `short`. I need to add it so this test gets picked.

shijin-aws:
>No our CI only runs short. I need to add it so this test gets picked.

Sorry I misunderstood, but making it marked as short is weird as your test is not supposed to be "short" only right? If you only want to get it picked by the CI I would prefer mark it as `functional`

jiaxiyan:
bot:aws:retest

jiaxiyan:
bot:aws:retest

PR #11261, opened 2025-07-31
prov/efa: Move protection domain back to device level, move address handle to device level

shijin-aws:
The earlier change that makes 1:1 relationship between PD and domain break the PD limit on platform that has high core count and limited max PD. This change moves the pd back to device level to lower the number of PDs libfabric can allocate.
Also moves the AH map resources to the device level from domain level, as AH is a PD level resources.

### DIFF
--- a/prov/efa/src/efa_device.c
+++ b/prov/efa/src/efa_device.c
@@ -22,6 +22,7 @@
 #include "efa.h"
 #include "efa_device.h"
 #include "efa_prov_info.h"
+#include "efa_av.h"
 
 #ifdef _WIN32
 #include "efawin.h"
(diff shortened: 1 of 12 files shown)


prov/efa/src/efa_device.c
```diff
@@ -182,6 +185,9 @@ static void efa_device_destruct(struct efa_device *device)
 		if (err)
 			EFA_INFO_ERRNO(FI_LOG_DOMAIN, "ibv_dealloc_pd",
 			               err);
+
```
sunkuamzn, line 188: I think you need to destroy any open AHs and clean up the AH map here

shijin-aws: I had that before, but I found a problem is it will break the fork test... If a child process exit and clean up the AH, the parent process will be impacted as AH is gone. We can have some follow up on this

shijin-aws: I think here the model is the same as MR.. they are both PD resources, and application needs to clean up them by fi_close. We don't clean all MRs during device destruct either

shijin-aws:
Close this PR as we are not moving in this direction

PR #11264, opened 2025-07-31
prov/cxi: Add multi-thread wait_fd support for RNR protocol

swelch:
When RNR messaging is used, a process should not block on a FD if a send is in retry time wait,
as manual progress is used. In addition, internal provider initiated progress should not disable
interrupts, e.g. RNR retry (to close race in multi-threaded environments).

### DIFF
--- a/man/fi_cxi.7.md
+++ b/man/fi_cxi.7.md
@@ -79,10 +79,20 @@ The CXI provider supports FI_THREAD_SAFE and FI_THREAD_DOMAIN threading models.
 ## Wait Objects
 
 The CXI provider supports FI_WAIT_FD and FI_WAIT_POLLFD CQ wait object types.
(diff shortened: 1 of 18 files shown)


swelch:
@j-xiong - the appveyor/pr non-debug CI failure appears unrelated:
```
Error sending AppVeyor compilation message: The remote server returned an error: (500) Internal Server Error.
  Message: '=': conversion from 'SOCKET' to 'int', possible loss of data
  Category: Warning
  Code: C4244
  File: C:\projects\libfabric\include\ofi_net.h
  Line: 218
  Column: 35
  Project name: libfabric
  Project file name: C:\projects\libfabric\libfabric.vcxproj
```

j-xiong:
@swelch I believe it's unrelated. I re-started the failed run anyway.

PR #11276, opened 2025-08-01
prov/efa: Optimizie the locking for efa cq

shijin-aws:
Currently, efa_cq_progress always acquire the ep_list_lock
    outside efa_cq_poll_ibv_cq. However, this lock is only needed
    when accessing the qp table which is not hit in the empty cq poll.
    
    This patch moves this locking to fruitful cq poll to optimize
    the empty cq poll case.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -269,6 +269,9 @@ int efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 	/* Call ibv_start_poll only once */
 	err = ibv_start_poll(cq->ibv_cq.ibv_cq_ex, &poll_cq_attr);
 	should_end_poll = !err;
+	/* Acquire the lock to prevent race conditions when qp_table is being updated */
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_cq.c
```diff
@@ -269,6 +269,9 @@ int efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 	/* Call ibv_start_poll only once */
 	err = ibv_start_poll(cq->ibv_cq.ibv_cq_ex, &poll_cq_attr);
 	should_end_poll = !err;
+	/* Acquire the lock to prevent race conditions when qp_table is being updated */
```
jiaxiyan, line 272: Can we do the same thing for efa_rdm_cq?

shijin-aws: We currently cannot do that for rdm cq as it is already iterating over some lists https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_cq.c#L757-L768 that are protected by this lock.

PR #11279, opened 2025-08-04
prov/cxi: Fix dlopen using default value for json-c library

Fruchix:
The current version of Libfabric does not consider a custom installation of the json-c dependency.

When providing the option `--with-json-c=DIR`, the selected json-c library is the system default and not the one located at DIR.

This is because `prov/cxi/src/cxip_curl.c` is using `lib_name` (with default value "libjson-c.so") instead of `json_libpath`.

### DIFF
--- a/prov/cxi/src/cxip_curl.c
+++ b/prov/cxi/src/cxip_curl.c
@@ -747,9 +747,9 @@ static int cxip_json_dl_init(void)
         json_libpath = strdup(lib_name);
         TRACE_CURL("libjson-c path set to %s\n", json_libpath);
 #endif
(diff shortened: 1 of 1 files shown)


heemstra (approved):

swelch (approved):

swelch:
@j-xiong the appveyor/pr failure looks to be unrelated and for the tcp provider.

swelch:
@shijin-aws - This is cxi internal, are you ok with us merging in spite of AWS CI error?

swelch:
@Fruchix - thanks for the patch, we will merge later today.

PR #6268, opened 2020-09-22
prov/efa: Switch to memhooks monitor

rwespetal:
The uffd monitor we are using currently defers the de-registration
until after the free, which is erroneous according to the rdma-core
interface. (e.g., it causes the RBTree in rdma-core update incorrectly
if fork support is enabled).

Switch to memhooks monitor to avoid the deferring.

Signed-off-by: Jie Zhang <zhngaj@amazon.com>

PR #6241, opened 2020-09-10
util_mem_monitor: Add flag to indicate whether to defer free

wckzhang:
Depending on monitor behavior, some monitors require calls that may
result in a free operation (for example memory deregistration), to be
deferred until outside of the monitor. This patch adds a flag and only
requests free deferral if necessary.

Signed-off-by: William Zhang <wilzhang@amazon.com>

Issue #5662, opened 2020-02-24
MR_CACHE userfaultfd leads to connection rejects on verbs.


vschhabra:
When using Verbs provider, we are seeing connection rejects leading to a failure of communication.

Client:
libfabric:172083:ofi_rxm:ep_ctrl:rxm_conn_handle_reject():1143<warn>connection reject:
 no reject error data (cm_data) was found (data length expected: 2
 found: 0)

On Server we see:

libfabric:180424:core:mr:ofi_monitor_subscribe():205<warn>
Failed (ret = -14) to monitor addr=0x7fb204041780 len=408

libfabric:180424:ofi_rxm:ep_ctrl:rxm_conn_handle_notify():1096<info>
notify event 1

libfabric:180424:verbs:domain:vrb_ep_close():360<info> EP
0x7fb2cc000a30 is being closed

FI_MR_CACHE_MAX_COUNT=0
(Turning MR CACHE off) makes is a workaround with reduced performance.

shefty:
Email communication indicated that using the memhooks cache monitor also worked.

shefty:
pull request #5666 changes the default memory monitor option until the usefaultfd problem can be resolved.

rwespetal:
Does turning on `FI_LOG_LEVEL=warn` give any useful information? I see a print in `ofi_uffd_register` that should give the errno for the failed userfaultfd registration.

Can we have providers override the default if the caching mechanism is not explicitly set? We'd like to keep userfaultfd enabled for EFA.

shefty:
I assume the errno is -14.  The monitor can still be set through the environment variable.  Does that work for you?  I'm also okay letting a provider change the default.

rwespetal:
Looking at `ofi_uffd_register` it returns -14 for all errors - seems like that's a side effect of having to try all the page sizes. I can open a PR to make this better.

I think we'd prefer to change the default in code for EFA instead of an environment variable. I didn't realize that the glibc memory hooks were replaced with syscall intercepts instead so this change might be okay.

shefty:
Yeah, the glibc memory hooks don't work.  What we have was pulled from OpenMPI, which they've used for quite some time now.

swelch:
@shefty, @vschhabra  - Is there a reproducer defined for this error when userfaultfd? userfaultfd has been working well for me, so I'd like to see what other issues there might be.

shefty:
[...]

PR #11282, opened 2025-08-05
prov/util: Change the order of setting default system memory monitor

### LINKED ISSUE #10943: memhook: Memhook monitor usage may lead to segfaults/illegal errors
Describe the bug:

memhook_monitor intercepts a set of glibc functions like mmap, unmap, madvise, etc. This involves patching the code memory at runtime, where the original assembler from glibc is replaced with custom libfabric runtime-generated assembler. This custom assembler runs some libfabric caching maintenance code before executing the actual syscall. The interception is global and affects glibc function calls made by every function in the running process. When the memhook is released (e.g., on domain destruction), another unrelated thread might be executing the assembler while it is being reverted to its original state by libfabric. This rare situation can lead to segmentation faults or illegal operations at runtime.

To Reproduce: 

Clone the reproducer repository: https://github.com/dsciebu/memhook_monitor.git
Follow the instructions in the repository to set up and run the reproducer.
Observe the behavior when the memhook is released.

Expected behavior 

The glibc functions should execute without causing segmentation faults or illegal operations, even when the memhook is released.

Environment:

OS: Ubuntu 22.04/24.04 (should irrelevant though)
Provider: Irrelevant (but the reproducer demands verbs)
Endpoint type: Irrelevant (RDM is used in the reproducer)
Additional context Add any other context about the problem here.

j-xiong:
Default system memory monitor is set to the first available one based on a predetermined order. Previously the order was: memhooks, uffd, kdreg2. The new order is: uffd, kdreg2, memhooks.

The reason behind the change is that the code patching mechanism in the memhooks monitor has inherent thread safety issues and other security concerns. In addition, userfault fd support has become more common with wider adption of newer kernels. Leaving memhooks monitor as the last choice reduces the chance of multithreaded applications hitting the thread safety issue with default configuration.

Fixes #10943

### DIFF
--- a/prov/util/src/util_mem_monitor.c
+++ b/prov/util/src/util_mem_monitor.c
@@ -298,12 +298,12 @@ void ofi_monitors_init(void)
 	if (!default_monitor) {
 #if defined(HAVE_MR_CACHE_MONITOR_DEFAULT)
 		set_default_monitor(HAVE_MR_CACHE_MONITOR_DEFAULT);
-#elif HAVE_MEMHOOKS_MONITOR
-		default_monitor = memhooks_monitor;
 #elif HAVE_UFFD_MONITOR
 		default_monitor = uffd_monitor;
 #elif HAVE_KDREG2_MONITOR
 		default_monitor = kdreg2_monitor;
+#elif HAVE_MEMHOOKS_MONITOR
+		default_monitor = memhooks_monitor;
 #else
 		default_monitor = NULL;
 #endif


shijin-aws:
I remember we had trouble with uffd monitor as it will cause dereg after free issue in mr cache. Let me find the relevant context

a-szegel:
bot:aws:retest

shijin-aws:
Found some historical PR on efa provider https://github.com/ofiwg/libfabric/pull/6268

shijin-aws:
Some more background:

The Userfault File Descriptor works by monitoring specific regions of memory. Whenever these regions have a page fault event, the thread that caused the page fault is put to sleep until a monitoring thread reads the faulting event off of an event list. Due to this specific behavior, several exceptions are present in this monitor.

  - Since it catches a page fault event, this means that there is no way to avoid a de-registration after a free for this monitor.
  - Since it puts a faulting thread to sleep, any free/deregistration event cannot occur in the handling/monitoring thread. Otherwise the monitoring thread can be put to sleep causing a permanent hang as there are no other threads to wake up the monitor thread. Thus, frees and deregistrations must be deferred.
   - Since reading the event wakes up the sleeping thread, the mm_lock must be locked before the read occurs to prevent access to this registration in the tree and cannot be unlocked until the registration is removed from the tree. This necessitates that the locking of the mm_lock occurs outside of the "ofi_mr_cache_notify" function.

Meanwhile, the dereg after free behavior in MR monitor caused problem for IBV MRs in fork conditions:

- ibv_dereg_mr() call ibv_dofork_range() to update the rbtree in rdma-core().

- ibv_dofork_range() call ibv_madvise_range(MADV_DOFORK) to do that.

- ibv_madvise_range(MADV_DOFORK) will call madvise(MADV_DOFORK), which will fail because the memory has already been returned to the OS, the error code is ENOMEM.

More context is available in this PR https://github.com/ofiwg/libfabric/pull/6241

j-xiong:
Also see #5662 and #5666

j-xiong:
Replaced by #11645

PR #11295, opened 2025-08-08
contrib/aws: Fix bug in kill_all_clusters() logic

a-szegel:
The get_cluster_name() function only takes 10 characters of the instance type. The kill_all_clusters() method was not previously working because it was REGEX string search was failing due to providing the entire instance type, instead of 10 characters.

### DIFF
--- a/contrib/aws/Jenkinsfile
+++ b/contrib/aws/Jenkinsfile
@@ -45,7 +45,7 @@ def install_porta_fiducia() {
 
 def kill_all_clusters(instance_type, region) {
     def instance_type_without_period = sh(
-        script: "echo ${instance_type} | tr -d '.\\n'",
+        script: "echo ${instance_type.take(10)} | tr -d '.\\n'",
         returnStdout: true
     )
     sh ". venv/bin/activate; ./PortaFiducia/scripts/delete_manual_cluster.py --cluster-name \'*${instance_type_without_period}*\' --region ${region} || true"


contrib/aws/Jenkinsfile
```diff
@@ -45,7 +45,7 @@ def install_porta_fiducia() {
 
 def kill_all_clusters(instance_type, region) {
     def instance_type_without_period = sh(
-        script: "echo ${instance_type} | tr -d '.\\n'",
+        script: "echo ${instance_type.take(10)} | tr -d '.\\n'",
```
sunkuamzn, line 48: why 10?

a-szegel: because 10 is what is used in the cluster name logic:
```

def get_cluster_name(build_tag, os, instance_type) {
    /*
     * Compose the cluster name. Pcluster requires a cluster name under 60 characters.
     * cluster name cannot have ".".
     * Jenkins does not allow groovy to use the replace() method
     * of string. Therefore we used shell command sed to replace "." with ""
     */
    def cluster_name = sh(
                        script: "echo '${get_cluster_name_prefix(build_tag)}-${os.take(10)}-${instance_type.take(10)}-'${get_random_string(8)} | tr -d '.\\n'",
                        returnStdout: true
                     )
    return cluster_name
}
```

a-szegel:
bot:aws:retest

shijin-aws (approved):

PR #11304, opened 2025-08-13
prov/efa: Support blocking cq read in efa-direct

jiaxiyan:
Implement fi_cq_sread and fi_cq_signal to support blocking completion queue operations with timeout and signaling capabilities.

Add test_rdm_pingpong and test_rma_bw with fi_cq_sread in efa-direct.


prov/efa/src/efa_cq.c
```diff
@@ -371,14 +371,212 @@ const char *efa_cq_strerror(struct fid_cq *cq_fid, int prov_errno,
...
+	ret = efa_cq_trywait(cq);
+	if (ret == FI_SUCCESS) {
+		/* CQ is empty, wait for events */
+		ret = efa_poll_events(cq, timeout);
+		if (ret)
+			return ret;
+	} else if (ret != -FI_EAGAIN) {
+		return ret;
+	}
+
```
shijin-aws, line 538: This assume you will always get a libfabric completion when device gets a completion event? What will happen for inject?

jiaxiyan: It doesn't make sense to use fi_cq_sread to wait for completions for fi_inject so I think it is safe to ignore. No cq entry will be generated for inject.

prov/efa/src/efa_cq.c
```diff
@@ -371,14 +371,212 @@ const char *efa_cq_strerror(struct fid_cq *cq_fid, int prov_errno,
...
+	ret = 0;
+	if (fds[0].revents & POLLIN) {
+		/* Data is available for reading on completion channel */
+		ibv_cq = ibv_cq_ex_to_cq(cq->ibv_cq.ibv_cq_ex);
+		ret = ibv_get_cq_event(cq->ibv_cq.channel, &ibv_cq, &context);
+		if (ret) {
+			EFA_WARN(FI_LOG_CQ, "ibv_get_cq_event failed with ret=%d\n", ret);
+			return ret;
```
alekswn, line 443: Is there possibility of getting the fds[1] event as well on this branch?
If so is it OK we have consumed it without calling `fd_signal_reset()`?

shijin-aws: AFAICT the fds[1] will be reset by L471 block even if this branch happens. The only early return happens when we hit an error, which I think it's fair.

prov/efa/src/efa_cq.c
shijin-aws: My concern is when user make both inject and non-inject calls and poll the cq for non-inject operations, this can potentially give them false positive. Can we follow verbs cq's implementation https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_cq.c#L194-L210 to loop until the actual completions reaches the threshold (1 for now) ?

jiaxiyan: It looks like a call to efadv_query_cq is needed to get the cq entry size https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_cq.c#L203
Can we skip this part for threshold 1 or we should just implement it for FI_CQ_COND_THRESHOLD now?

shijin-aws: that is only libfabric's cq entry size, you can follow https://github.com/ofiwg/libfabric/blob/main/prov/lpp/src/lpp_cq.c#L155-L172 to implement it

prov/efa/src/efa_cq.c
```diff
@@ -343,14 +343,227 @@ const char *efa_cq_strerror(struct fid_cq *cq_fid, int prov_errno,
...
+ * is signaled and no data is available to be read from the completion queue,
+ * returns -FI_EAGAIN. If the topmost completion is for a failed transfer
+ * (an error entry), returns -FI_EAVAIL.
+ */
+static ssize_t efa_cq_sreadfrom(struct fid_cq *cq_fid, void *buf, size_t count,
+								fi_addr_t *src_addr, const void *cond,
+								int timeout)
+{
+	struct efa_cq *cq;
+	ssize_t ret, num_completions = 0;
```
shijin-aws:
```
 prov/efa/src/efa_cq.c: In function ‘efa_cq_sreadfrom’:
prov/efa/src/efa_cq.c:487:17: error: ‘ret’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  487 |         ssize_t ret, num_completions = 0;
      |                 ^~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:30650: prov/efa/src/src_libfabric_la-efa_cq.lo] Error 1
```

shijin-aws: It looks your ret should always be set before return, unless the threshold is 0... I think set it as 0 here is safe.

fabtests/unit/cq_test.c
```diff
@@ -171,6 +171,8 @@ cq_signal()
 		sprintf(err_buf, "fi_cq_open(1, 0, FI_CQ_FORMAT_UNSPEC, "
 				"FI_WAIT_UNSPEC) = %d, %s",
 				ret, fi_strerror(-ret));
+		if (ret == -FI_EOPNOTSUPP)
```
shijin-aws: Not sure why we need this change, before your PR, EFA cq won't support WAIT_UNSPEC anyway right?

jiaxiyan: Before it returns -FI_ENOSYS which is handled by 
```
#define TEST_RET_VAL(_ret, _testret) \
	(_ret == -FI_ENOSYS || _ret == -FI_ENODATA) ? SKIPPED : (_testret)
```

shijin-aws: I think we should use the same errno for device unsupported case as well

shijin-aws: >FI_EOPNOTSUPP
    Operation not supported on transport endpoint

Which is not exactly what we have

jiaxiyan: change to FI_ENOSYS

jiaxiyan:
bot:aws:retest

shijin-aws (approved):

shijin-aws:
bot:aws:retest

PR #11306, opened 2025-08-14
contrib/intel/jenkins: Update opt-out files

zachdworkin:
Fix fabtests/pytests to be fabtests/pytest so that PRs with changes only in files fabtests/pytest/* get properly skipped.


zachdworkin:
no need to wait for appveyor or aws ci since changes are only to intel jenkinsfile

PR #11314, opened 2025-08-15
fabtests/efa: Don't use hardcoded `fi_info` path

darrylabbate:
This otherwise assumes the EFA installer is installed on the host.

### DIFF
--- a/fabtests/pytest/efa/efa_common.py
+++ b/fabtests/pytest/efa/efa_common.py
@@ -123,7 +123,7 @@ def get_efa_domain_names(server_id):
     process_timed_out = False
 
     # This command returns a list of EFA domain names and its related info
-    command = "ssh {} fi_info -p efa".format(server_id)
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

shijin-aws:
It seems this change break CI

fabtests/pytest/efa/efa_common.py
```diff
@@ -123,7 +123,7 @@ def get_efa_domain_names(server_id):
     process_timed_out = False
 
     # This command returns a list of EFA domain names and its related info
-    command = "ssh {} fi_info -p efa".format(server_id)
+    command = "ssh {} 'fi_info -p efa || /opt/amazon/efa/bin/fi_info -p efa'".format(server_id)
```
shijin-aws, line 126: I still think it is better to populate the environment variables from cmdline_args, which can include $PATH. The caller should have a correct $PATH in case the shell doesn't do it

PR #11317, opened 2025-08-15
prov/efa: Rework the efa cq locking optimization

shijin-aws:
The earlier https://github.com/ofiwg/libfabric/pull/11276 got reverted as it has a bug and caused deadlock
when polling cq and closing ep concurrently.

This patch fixed that bug and re-enable this optimization.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -270,6 +270,10 @@ int efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 	/* Call ibv_start_poll only once */
 	efa_cq_start_poll(ibv_cq);
 
+	/* Acquire the lock to prevent race conditions when qp_table is being updated */
+	if(ibv_cq->poll_active)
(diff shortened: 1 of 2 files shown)


shijin-aws:
Will rebase this PR on top of https://github.com/ofiwg/libfabric/pull/11315 after its merged

darrylabbate (approved):

prov/efa/src/efa_cq.c
```diff
@@ -331,6 +335,10 @@ int efa_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 		efa_cq_next_poll(ibv_cq);
 	}
 	err = ibv_cq->poll_err;
+
+	if(ibv_cq->poll_active)
```
shijin-aws, line 339: I need to update this.. This lock should be released after the end poll

shijin-aws:
I realize this change can be thread unsafe as we are accessing the qp table in `start_poll` as well in the rdma-core bypass path... Need some more thinking

shijin-aws:
Close for now

PR #11320, opened 2025-08-18
core/common: Make common runtime parameters working for DL providers

### LINKED ISSUE #11312: prov/rxm: `FI_UNIVERSE_SIZE` environment variable not working with RXM provider due to duplicate `ofi_universe_size` symbols
**Describe the bug**
The `FI_UNIVERSE_SIZE` environment variable does not take effect when using the RXM provider because the `ofi_universe_size` global variable exists as separate instances in `libfabric.so` and `librxm-fi.so`, causing the provider to use a different variable than the one updated by the environment variable parsing.

The issue occurs because `src/common.c` (which contains the definition of `ofi_universe_size`) is included in both:
- The main `libfabric.so` library 
- Provider libraries like `librxm-fi.so` when built as dynamic libraries

This creates duplicate symbol definitions. When the environment variable is processed, only the instance in `libfabric.so` is updated, while the provider libraries continue to use their own unmodified instances.

**To Reproduce**
1. Build libfabric with dynamic RXM provider support
2. Set environment variable: `export FI_UNIVERSE_SIZE=8`
3. Run an application using RXM provider with debug logging enabled
4. Observe that the RXM provider still uses the default value (1024) instead of the environment variable value (8)

**Expected behavior**
When `FI_UNIVERSE_SIZE=8` is set, both libfabric core and RXM provider should use the same value (8).

The libfabric core reads and updates the environment variable correctly, but the RXM provider continues to use the default value because it has its own separate instance of the variable.

**Output**
With `FI_LOG_LEVEL=Debug`, the following logs clearly show the problem:

```
# libfabric core - reads environment variable correctly
libfabric:565646:1755241641::core:core:fi_ini():953<info> ofi_universe_size: 1024
libfabric:565646:1755241641::core:core:fi_ini():955<info> ofi_universe_size pointer: 0x7f49c1a611a0
libfabric:565646:1755241641::core:core:fi_param_get_():405<info> read long var universe_size=8

# rxm provider - uses different instance with default value
libfabric:565646:1755241641::ofi_rxm:core:rxm_ep_msg_cq_open():1371<info> universe size: 1024
libfabric:565646:1755241641::ofi_rxm:core:rxm_ep_msg_cq_open():1373<info> universe size pointer: 0x7f49bee7ea20
```

Notice the different memory addresses (`0x7f49c1a611a0` vs `0x7f49bee7ea20`), confirming these are separate variable instances.

```bash
# Check for duplicate symbols
$ nm build/lib/libfabric.so | grep ofi_universe_size
000000000005a140 d ofi_universe_size
$ nm build/lib/libfabric/librxm-fi.so | grep ofi_universe_size
0000000000065a00 d ofi_universe_size
```

**Environment:**

- libfabric version: main(be64c6983626c3891a1e8cadebf7edb585b6dd5f)
- Build configuration: Dynamic provider libraries enabled (`--enable-rxm=dl`)
- Affected providers: All providers that use `$(common_srcs)` (rxm, verbs, sockets, etc.)
- OS: Linux

j-xiong:
Runtime parameters decalared in common.c weren't initialized for DL providers because the initialization was done in core-only code.

Move the initialization to a common function `ofi_params_init` which can be called from DL providers.

Fixes #11312

### DIFF
--- a/prov/ucx/src/ucx_init.c
+++ b/prov/ucx/src/ucx_init.c
@@ -399,7 +399,8 @@ struct fi_provider ucx_prov = {
 UCX_INI
 {
 #if HAVE_UCX_DL
-        ofi_hmem_init();
+	ofi_hmem_init();
+	ofi_params_init();
 #endif
 
 	ucx_init_errcodes();
--- a/src/common.c
+++ b/src/common.c
@@ -85,10 +85,20 @@ struct ofi_common_locks common_locks = {
 	.util_fabric_lock = PTHREAD_MUTEX_INITIALIZER,
 };
 
+int ofi_fork_unsafe;
 size_t ofi_universe_size = 1024;
 int ofi_av_remove_cleanup;
 char *ofi_offload_coll_prov_name = NULL;
 
+
+void ofi_params_init(void)
+{
+	fi_param_get_bool(NULL, "fork_unsafe", &ofi_fork_unsafe);
+	fi_param_get_size_t(NULL, "universe_size", &ofi_universe_size);
+	fi_param_get_bool(NULL, "av_remove_cleanup", &ofi_av_remove_cleanup);
+	fi_param_get_str(NULL, "offload_coll_provider",
+			 &ofi_offload_coll_prov_name);
+}
 
 int ofi_genlock_init(struct ofi_genlock *lock,
 		     enum ofi_lock_type lock_type)
--- a/include/ofi.h
+++ b/include/ofi.h
@@ -336,6 +336,7 @@ void fi_param_fini(void);
 void fi_param_undefine(const struct fi_provider *provider);
 void ofi_remove_comma(char *buffer);
 void ofi_dump_sysconfig(void);
+void ofi_params_init(void);
 
 const char *ofi_hex_str(const uint8_t *data, size_t len);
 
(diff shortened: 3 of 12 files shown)


shijin-aws:
bot:aws:retest

prov/ucx/src/ucx_init.c
```diff
@@ -400,6 +400,7 @@ UCX_INI
 {
 #if HAVE_UCX_DL
         ofi_hmem_init();
```
shijin-aws: L402 indent is off

j-xiong: Thanks. The old code was using spaces instead of tab. Fixed now.

shijin-aws (approved):

src/common.c
```diff
@@ -85,11 +85,46 @@ struct ofi_common_locks common_locks = {
...
 char *ofi_offload_coll_prov_name = NULL;
 
 
+void ofi_params_init(void)
+{
+	fi_param_define(NULL, "fork_unsafe", FI_PARAM_BOOL,
+			"Whether use of fork() may be unsafe for some providers "
+			"(default: no). Setting this to yes could improve "
+			"performance at the expense of making fork() potentially "
+			"unsafe");
```
shefty: Doesn't this result in the same parameter being defined multiple times?  The provider param is null here, so the core and all DL providers will define the same parameter.

I think the DL providers only need to call param_get to read the value, with the assumption that libfabric core has already defined the param.

j-xiong: Good call. Fixed.

PR #11326, opened 2025-08-19
prov/efa: Initialize nevents in efa_domain_cq_open_ext

jiaxiyan:
GPU cannot do a blocking wait on CQ entries because system FDs are only accessible to CPU.
Always initialize nevents to prevent accessing uninitialized memory during efa_cq_close.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -631,6 +631,8 @@ static int efa_domain_cq_open_ext(struct fid_domain *domain_fid,
 	struct efa_domain *efa_domain;
 	int err, retv;
 
+	/* GPU cannot do a blocking wait on CQ entries because 
+	 * system FDs are only accessible to CPU. */
 	if (attr->wait_obj != FI_WAIT_NONE)
 		return -FI_ENOSYS;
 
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):

shijin-aws:
I think we may disable rdma-core bypass when wait is requested as well (before the cq doorbell is implemented).. Not in this PR

prov/efa/src/efa_domain.c
```diff
@@ -678,11 +689,13 @@ static int efa_domain_cq_open_ext(struct fid_domain *domain_fid,
...
 	}
 
+	ofi_atomic_initialize32(&cq->nevents, 0);
+
 	*cq_fid = &cq->util_cq.cq_fid;
 	(*cq_fid)->fid.fclass = FI_CLASS_CQ;
 	(*cq_fid)->fid.context = context;
 	(*cq_fid)->fid.ops = &efa_cq_fi_ops;
-	(*cq_fid)->ops = efa_cq_init_attr->ext_mem_dmabuf.buffer ? &efa_cq_ops : &efa_cq_ext_ops;
+	(*cq_fid)->ops = efa_cq_init_attr->ext_mem_dmabuf.buffer ? &efa_cq_no_sread_ops : &efa_cq_ext_ops;
```
shijin-aws: Why not just fail when WAIT is requested?

jiaxiyan: We have that logic on line 645. I can remove this if you think it's unnecessary.

shijin-aws: Yeah it's fine. `fi_cq_sread` should return ENOSYS if channel is not created so we don't need a change on the cq ops anyway

jiaxiyan:
bot:aws:retest

shijin-aws (approved):

jiaxiyan:
CI failure is unrelated

PR #11327, opened 2025-08-20
prov/efa: Fix the unsolicited write recv detection

shijin-aws:
Currently, efa provider uses the device & env to check the unsolicited write recv. This check is wrong
and can cause false positive when the efadv_cq creation failed. This patch fixes this issue by tracking the unsolicited
write recv status in both efa_ibv_cq and efa_qp.


prov/efa/src/efa_base_ep.c
```diff
@@ -257,17 +261,92 @@ int efa_qp_create(struct efa_qp **qp, struct ibv_qp_init_attr_ex *init_attr_ex,
...
+				  struct efa_ibv_cq *tx_cq,
+				  struct efa_ibv_cq *rx_cq,
+				  bool create_user_recv_qp)
 {
 	int ret;
+	struct ibv_qp_init_attr_ex attr_ex = { 0 };
+
+	efa_base_ep_construct_ibv_qp_init_attr_ex(base_ep, &attr_ex, tx_cq->ibv_cq_ex, rx_cq->ibv_cq_ex);
+
+	assert(tx_cq->unsolicited_write_recv_enabled == rx_cq->unsolicited_write_recv_enabled);
```
jiaxiyan, line 320: what happens if these don't match? Should we disable unsolicited_write_recv?

shijin-aws: It is unlikely they don't match as they share the same setup and are in the same process. Adding the assert is to catch such really unexpected incident

PR #11337, opened 2025-08-22
prov/efa: Increase default efa-direct ibvcq size

a-szegel:
Increase default ibv_cq size on efa-direct from 1k to 16k.  Having too small of a cq can lead to cq overflow, which will cause completions to be dropped. Users can choose to specify cq size in cq_attr which will override this number.

### DIFF
--- a/prov/efa/src/efa.h
+++ b/prov/efa/src/efa.h
@@ -81,7 +81,8 @@
 #define EFA_DEF_POOL_ALIGNMENT (8)
 #define EFA_MEM_ALIGNMENT (64)
 
-#define EFA_DEF_CQ_SIZE 1024
+/* 4k tx_attr.size + 8k rx_attr.size */
+#define EFA_DEF_CQ_SIZE 12288
 
 
 #define EFA_DEFAULT_RUNT_SIZE (307200)


shijin-aws:
I think we should default as 12k (tx + rx queue size) per earlier discussion.

@bwbarrett I remember you had some comments on the default cq size behavior for efa-direct, can u share your feedback here?

a-szegel:
Write with immediate gets a spot on the CQ as well, which is why I made it a bit bigger.

bwbarrett:
That makes the minimum CQ size 512 KiB, which isn't exactly tiny.  I'd start with ignoring unsolicited write w/ immediate and set the default to the default TX submission size + default RX submission size.  That should be considerably smaller than 16K entries?

a-szegel:
@bwbarrett  8k RX, and 4k TX, so 4k smaller

prov/efa/src/efa.h
```diff
@@ -81,7 +81,7 @@
 #define EFA_DEF_POOL_ALIGNMENT (8)
 #define EFA_MEM_ALIGNMENT (64)
 
-#define EFA_DEF_CQ_SIZE 1024
+#define EFA_DEF_CQ_SIZE 12288
```
shijin-aws, line 85: Nit: add a comment here to explain why 12K

a-szegel: done

PR #11341, opened 2025-08-25
prov/cxi: Improve multithreaded performance of bound counter operations

ronawho:
Add an opt-in feature to use a dedicated triggered command queue for bound counters instead of using the domain's. This can improve multithreaded performance of counters for FI_THREAD_SAFE and FI_THREAD_COMPLETION threading models.


shijin-aws:
bot:aws:retest

PR #11325, opened 2025-08-19
fabtests/pingpong: EFA competition traces and a race condition fix

alekswn:
These changes should make AWS CI failures more informative and fix occasional client-side timeouts in 32 byte pingpong with efa-direct.

- **EFA RX/TX completions**
Currently we have traces for posting send and receive buffers with EFA provider. This change adds completion traces as well.

~~- **Unexpected completion check in `ft_finalize()`**~~
~~We have seen issues caused by duplicate EFA completions, adding check for unexpected ones to `ft_finalize()`.~~

~~- **Use `ft_sync()` instead of `ft_finalize_ep(ep)` if `ft_finalize()`**.~~
~~- **Call ft_sync() before ft_finalize()**~~
- **Follow `ft_sync_inband(false)` pattern in `ft_finalize()`**

There are two effects of this change:
1. If out-of-band sync was enabled it's going to be used for the final sync as well.
2. In case of inband sync server will wait for client to process all recv completions before it can exit.

PR #11342, opened 2025-08-25
prov/efa and fabtests: add completion traces and fixes ep finalize semantics

shijin-aws:
reworked https://github.com/ofiwg/libfabric/pull/11325


shijin-aws:
bot:aws:retest

PR #11352, opened 2025-08-27
fabtests/efa: Add volatile to prevent compiler optimization of CQE flag

jiaxiyan:
The CUDA compiler was incorrectly optimizing away memory reads of the completion queue entry (CQE) flags field, assuming the memory content was constant. However, the CQE flags are updated asynchronously by the EFA device hardware, and repeated reads must fetch the current value from memory rather than using a cached register value. Without this fix, the efa_cqe_is_pending() function could return stale data, causing completion polling to hang or miss valid completions.

### DIFF
--- a/fabtests/prov/efa/src/efagda/cuda_kernel.cu
+++ b/fabtests/prov/efa/src/efagda/cuda_kernel.cu
@@ -39,7 +39,8 @@ __device__ uint32_t efa_sub_cq_get_current_index(const efa_sub_cq *sub_cq)
 __device__ int efa_cqe_is_pending(const efa_io_cdesc_common *cqe_common,
 				  int phase)
 {
-	return EFA_GET(&cqe_common->flags, EFA_IO_CDESC_COMMON_PHASE) == phase;
+	volatile uint8_t *cqe_flag = (volatile uint8_t *)(&cqe_common->flags);
+	return EFA_GET(cqe_flag, EFA_IO_CDESC_COMMON_PHASE) == phase;
 }
 
 __device__ efa_io_cdesc_common *efa_sub_cq_get_cqe(efa_sub_cq *sub_cq,


a-szegel (approved):

shijin-aws (commented):
@mrgolin please take a look to see whether we need this fix for efagda library

bwbarrett (approved):
@a-szegel had asked why this wasn't causing problems in the fabtests, so I thought I'd answer here.  Technically, this change isn't needed for the test as written, but we should still make it to keep all the codes consistent (and keep someone from stubbing their toe later).

The problem in Seth's code is that `efa_cqe_is_pending()` was being called in a loop in the cuda kernel.  The compiler inlined the callstack inside that loop down to and including `efa_cqe_is_pending()`, and then the optimizer noticed that it could save a whole lot of effort by not reading `cqe_flag` every time, and the bug was born.

But in this test code, the cuda kernel calls poll once and then effectively exits, and the polling loop is actually in the host code.  The cuda compiler never sees the loop, because it's effectively split, and the host compiler can't perform the same optimizations because all the cuda calling semantics have the right hints to keep the compiler from doing the wrong thing.  The key is that the polling cuda kernel doesn't actually poll in a loop.  It reads the CQ once and exits, so there's never any opportunity for the compiler to optimize out a necessary read.

I would suggest a follow-up change, which is to also write a fabtest that drives multiple wqe submissions and the complete polling loop all from a single CUDA kernel, to help find other places we might have overlooked this kind of synchronization problem.

Also, we should really have a READ_ONCE() macro instead of all the extra casting, just to make it more clear what we're doing.

mrgolin:
> @mrgolin please take a look to see whether we need this fix for efagda library

Yes, I fixed it there too. That's one of the reasons why we should try to have a single implementation.

PR #11355, opened 2025-08-28
configure: Improve restricted-dl help text

jswaro:
The original text as not as clear as it could be why this option is useful. This commit amends the text to make it clearer. 

I've tested this by running autogen.sh to generate the configuration files from a clean build, and run ./configure to ensure everything looks fine, including the help text.


j-xiong (approved):

j-xiong:
The CI failure is unrelated.

PR #11356, opened 2025-08-28
Introduce a lightweight Rust bindings for Libfabric, using bindgen.

kyle-yh-kim:
### Motivation
Increasing number of HPC networking code is being written in Rust. Naturally, to support Libfabric usage in Rust, there needs a proper Rust library that wraps Libfabric APIs written in C. This practice is commonly referred as a Rust binding / FFI (foreign function interface).

This library builds a lightweight Rust binding via bindgen. Lightweight, meaning there's no additional abstraction on top of the automaticallhy generated code via bindgen, aside from the `wrapper.c/h` which is strictly used to support `static inline` functions to be properly binded, by introducing a new translation unit upon compilation.

### Build
```
// Clean the existing build.
cargo clean

// Build, using the Libfabric binary that is compiled on-the-fly.
cargo build --features vendored

// Build, using the already installed Libfabric.
// You may be required to set `PKG_CONFIG_PATH` env variable to point towards `libfabric.pkg` file.
cargo build

// Unit-tests.
cargo test

// Unit-tests with ASAN enabled.
cargo test --features asan
```

### How to use the library

Add the crate dependency under your Rust application's `Cargo.toml` file. Then;

```rust
use bindings as ffi;

fn test_get_info() {
    unsafe {
        // Configure hints.
        let hints = ffi::fi_allocinfo();
        assert_eq!(hints.is_null(), false);

        (*hints).caps = ffi::FI_MSG as u64;
        (*hints).mode = ff::FI_CONTEXT;
        (*(*hints).ep_attr).type_ = ffi::fi_ep_type_FI_EP_RDM;
        (*(*hints).domain_attr).mr_mode = ffi::FI_MR_LOCAL as i32;
        let prov_name = CString::new("efa").unwrap();
        (*(*hints).fabric_attr).prov_name = prov_name.into_raw() as *mut i8;

        // Get Fabric info based on the hints.
        let mut info_ptr = ptr::null_mut();
        let version = ffi::fi_version();
        let ret = ffi::fi_getinfo(
            version,
            ptr::null_mut(),
            ptr::null_mut(),
            0,
            hints,
            &mut info_ptr,
        );

        assert_eq!(ret, 0);

        // Free the info structure returned by fi_getinfo.
        if !info_ptr.is_null() {
            ffi::fi_freeinfo(info_ptr);
        }

        // Free the hints structure we allocated.
        ffi::fi_freeinfo(hints);
    }
}
```

### Files
- `build.rs`: The actual build script for the bindgen.
- `src/lib.rs`: The generated binding is copy-pasted programmatically and publicly exported under `bindings` namespace.
- `wrapper.c/h`: Wrapper source files that simply calls the static inline functions. This way, an isolated translation unit for each static inline function is made, for which the Rust bindgen is able to link against it.
- `tests/unit_test.rs`: Unit tests.

### DIFF
--- a/bindings/rust/README.md
+++ b/bindings/rust/README.md
@@ -1,0 +1,87 @@
+### Motivation
+
+Increasing number of HPC networking code is being written in Rust. Naturally, to
+support Libfabric usage in Rust, there needs a proper Rust library that wraps
+Libfabric APIs written in C. This practice is commonly referred as a Rust
+binding / FFI (foreign function interface).
+
+This library builds a lightweight Rust binding via bindgen. Lightweight, meaning
+there's no additional abstraction on top of the automatically generated code via
+bindgen, aside from the `wrapper.[ch]` which is strictly used to support
+`static inline` functions to be properly bound, by introducing a new translation
+unit upon compilation.
+
+### Build
+
+```
+// Clean the existing build.
+cargo clean
+
+// Build, using the Libfabric binary that is compiled on-the-fly.
+cargo build --features vendored
+
+// Build, using the already installed Libfabric.
+// You may be required to set `PKG_CONFIG_PATH` env variable to point towards `libfabric.pkg` file.
+cargo build
+
+// Unit-tests.
+cargo test
+
+// Unit-tests with ASAN enabled.
+cargo test --features asan
+```
+
+### How to use the library
+
+Add the crate dependency under your Rust application's `Cargo.toml` file. Then;
+
+```rust
+use bindings as ffi;
+
+fn test_get_info() {
+    unsafe {
+        // Configure hints.
+        let hints = ffi::fi_allocinfo();
+        assert_eq!(hints.is_null(), false);
+
+        (*hints).caps = ffi::FI_MSG as u64;
+        (*hints).mode = ff::FI_CONTEXT;
+        (*(*hints).ep_attr).type_ = ffi::fi_ep_type_FI_EP_RDM;
+        (*(*hints).domain_attr).mr_mode = ffi::FI_MR_LOCAL as i32;
+        let prov_name = CString::new("efa").unwrap();
+        (*(*hints).fabric_attr).prov_name = prov_name.into_raw() as *mut i8;
+
+        // Get Fabric info based on the hints.
+        let mut info_ptr = ptr::null_mut();
+        let version = ffi::fi_version();
+        let ret = ffi::fi_getinfo(
+            version,
+            ptr::null_mut(),
+            ptr::null_mut(),
+            0,
+            hints,
+            &mut info_ptr,
+        );
+
+        assert_eq!(ret, 0);
+
+        // Free the info structure returned by fi_getinfo.
+        if !info_ptr.is_null() {
+            ffi::fi_freeinfo(info_ptr);
+        }
+
+        // Free the hints structure we allocated.
+        ffi::fi_freeinfo(hints);
+    }
+}
+```
+
+### Files
+
+- `build.rs`: The actual build script for the bindgen.
+- `src/lib.rs`: The generated binding is copy-pasted programmatically and
+  publicly exported under `bindings` namespace.
+- `wrapper.[ch]`: Wrapper source files that simply calls the static inline
+  functions. This way, an isolated translation unit for each static inline
+  function is made, for which the Rust bindgen is able to link against it.
+- `tests/unit_test.rs`: Unit tests.
--- a/bindings/rust/wrapper.c
+++ b/bindings/rust/wrapper.c
@@ -1,0 +1,1098 @@
+#include "wrapper.h"
+
+/* Proprietary helper function declarations. */
+
+fid_t get_fid_ptr(void *ptr)
+{
+	return (fid_t) ptr;
+}
+
+/* Static inline function declarations from fi_prov.h */
+
+int wrap_fi_param_get_str(struct fi_provider *provider, const char *param_name,
+			  char **value)
+{
+	return fi_param_get_str(provider, param_name, value);
+}
+
+int wrap_fi_param_get_int(struct fi_provider *provider, const char *param_name,
+			  int *value)
+{
+	return fi_param_get_int(provider, param_name, value);
+}
+
+int wrap_fi_param_get_bool(struct fi_provider *provider, const char *param_name,
+			   int *value)
+{
+	return fi_param_get_bool(provider, param_name, value);
+}
+
+int wrap_fi_param_get_size_t(struct fi_provider *provider,
+			     const char *param_name, size_t *value)
+{
+	return fi_param_get_size_t(provider, param_name, value);
+}
+
+/* Static inline function declarations from fabric.h */
+
+struct fi_info *wrap_fi_allocinfo(void)
+{
+	return fi_allocinfo();
+}
+
+int wrap_fi_close(struct fid *fid)
+{
+	return fi_close(fid);
+}
+
+int wrap_fi_control(struct fid *fid, int command, void *arg)
+{
+	return fi_control(fid, command, arg);
+}
+
+int wrap_fi_alias(struct fid *fid, struct fid **alias_fid, uint64_t flags)
+{
+	return fi_alias(fid, alias_fid, flags);
+}
+
+int wrap_fi_get_val(struct fid *fid, int name, void *val)
+{
+	return fi_get_val(fid, name, val);
+}
+
+int wrap_fi_set_val(struct fid *fid, int name, void *val)
+{
+	return fi_set_val(fid, name, val);
+}
+
+int wrap_fi_open_ops(struct fid *fid, const char *name, uint64_t flags,
+		     void **ops, void *context)
+{
+	return fi_open_ops(fid, name, flags, ops, context);
+}
+
+int wrap_fi_set_ops(struct fid *fid, const char *name, uint64_t flags,
+		    void *ops, void *context)
+{
+	return fi_set_ops(fid, name, flags, ops, context);
+}
+
+/* Static inline function declarations from fi_endpoint.h */
+
+int wrap_fi_passive_ep(struct fid_fabric *fabric, struct fi_info *info,
+		       struct fid_pep **pep, void *context)
+{
+	return fi_passive_ep(fabric, info, pep, context);
+}
+
+int wrap_fi_endpoint(struct fid_domain *domain, struct fi_info *info,
+		     struct fid_ep **ep, void *context)
+{
+	return fi_endpoint(domain, info, ep, context);
+}
+
+int wrap_fi_endpoint2(struct fid_domain *domain, struct fi_info *info,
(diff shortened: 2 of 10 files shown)


src/bindings/rust/wrapper.c
```diff
@@ -0,0 +1,1030 @@
...
+}
+
+ssize_t _fi_fetch_atomicv(struct fid_ep *ep,
+						  const struct fi_ioc *iov, void **desc, size_t count,
+						  struct fi_ioc *resultv, void **result_desc, size_t result_count,
+						  fi_addr_t dest_addr,
+						  uint64_t addr, uint64_t key,
+						  enum fi_datatype datatype, enum fi_op op, void *context)
+{
+	return fi_fetch_atomicv(ep, iov, desc, count, resultv, result_desc, result_count, dest_addr, addr, key, datatype, op, context);
```
j-xiong: Lines like this are too long.

j-xiong:
I would put this under `bindings` instead of `src/bindings`. In addition, if it is to be included in a release tarball, these files need to be added to `EXTRA_DIST` in `Makefile.am`.

shefty:
wrapper.c and wrapper.h look almost completely generic. Those two files may be better off living within the main source tree somewhere.  I'd prefer to use a different prefix, such as 'wrap', rather than just an underscore.

Ideally, we'd come up with some macro magic or maybe even a script to generate those files from the public headers.  Otherwise, updating them will be missed when the public headers are changed.  I doubt is possible; however.

j-xiong:
I do prefer keeping the `wrapper.[ch]` files in current location. They serve no purpose other than making the static inline functions usable to rust.

kyle-yh-kim:
Thanks for your reviews thus far. Responding to the comments;

@j-xiong
> I would put this under bindings instead of src/bindings.
- Agreed. I benchmarked the location based on NIXL. But we don't have to follow their project structure.

> In addition, if it is to be included in a release tarball, these files need to be added to EXTRA_DIST in Makefile.am.
- `bindings.rs` is a file generated on-the-fly via `cargo build`. So this does not need to be included in a release tarball.

@shefty
> I'd prefer to use a different prefix, such as 'wrap', rather than just an underscore.

Agreed with the `wrap_` prefix. This is a stronger contract versus a generic `_`. For example; `wrap_fi_info()` versus `_fi_info()`.

> wrapper.c and wrapper.h look almost completely generic. Those two files may be better off living within the main source tree somewhere.

In terms of the file location, I prefer to keep it under the Rust bindings project directory. As Jianxin mentioned above, only under the Rust bindings project, they serve a specific purpose of exposing a translation unit for each `static inline` function.

> Ideally, we'd come up with some macro magic or maybe even a script to generate those files from the public headers. Otherwise, updating them will be missed when the public headers are changed. I doubt is possible; however.

Good point, and I've looked into this for sometime. I unfortunately couldn't find a good solution around it. Thankfully, the Rust bindings source code is still housed under `ofiwg/libfabric` repository, and not under a separate repository - and so at least, there's a tight coupling between Libfabric source code to its appropriate Rust binding code.

shefty:
The code ends up similar to what was done in all the hooking providers.  And it's worth noting that the hooking providers have not been updated to support new API calls (e.g. fi_endpoint2).  Even if there's no good way to generate the wrapping functions (that would be even harder for the hook providers), maybe there's some script that could identify missing functions which could be incorporated into the CI.  Maybe we can get some fancy AI to write one for us.

shefty:
Btw, I don't think we block progress on this PR while figuring out how to keep wrapper functions in sync.  We can just open an issue to track it for the next 365 days until the issue becomes stale and self-closes.

j-xiong:
Please squash the two commits and do a force push. There is no need to fix a prior commit with a new commit in the same PR.

src/bindings/rust/wrapper.c
kyle-yh-kim: Understood. No strong opinions on this one. Would you prefer to shorten the argument name, or to use new lines, or both?

j-xiong: use new lines.

bindings/rust/README.md
```diff
@@ -0,0 +1,74 @@
+### Motivation
+Increasing number of HPC networking code is being written in Rust. Naturally, to support Libfabric usage in Rust, there needs a proper Rust library that wraps Libfabric APIs written in C. This practice is commonly referred as a Rust binding / FFI (foreign function interface).
+
+This library builds a lightweight Rust binding via bindgen. Lightweight, meaning there's no additional abstraction on top of the automaticallhy generated code via bindgen, aside from the `wrapper.c/h` which is strictly used to support `static inline` functions to be properly binded, by introducing a new translation unit upon compilation.
```
j-xiong: `automatically`
`wrapper.[ch]`
`binded` --> `bound`

bindings/rust/README.md
```diff
@@ -0,0 +1,74 @@
...
+        // Free the hints structure we allocated.
+        ffi::fi_freeinfo(hints);
+    }
+}
+```
+
+### Files
+- `build.rs`: The actual build script for the bindgen.
+- `src/lib.rs`: The generated binding is copy-pasted programmatically and publicly exported under `bindings` namespace.
+- `wrapper.c/h`: Wrapper source files that simply calls the static inline functions. This way, an isolated translation unit for each static inline function is made, for which the Rust bindgen is able to link against it.
```
j-xiong: `wrapper.[ch]`

bindings/rust/README.md
```diff
@@ -0,0 +1,74 @@
+### Motivation
+Increasing number of HPC networking code is being written in Rust. Naturally, to support Libfabric usage in Rust, there needs a proper Rust library that wraps Libfabric APIs written in C. This practice is commonly referred as a Rust binding / FFI (foreign function interface).
```
j-xiong: Break the long lines by adding newline where appropriate. Same for other places.

bindings/rust/wrapper.h
```diff
@@ -0,0 +1,367 @@
+#ifndef container_of
```
j-xiong, line 4: As common practice, enclose the entire header file with conditionals to avoid duplicated inclusion, e.g.:
```
#ifndef __WRAPPER_H__
#define __WRAPPER_H__
......
#endif /* __WRAPPER_H */
```

bindings/rust/wrapper.h
```diff
@@ -0,0 +1,367 @@
...
+
+/* Static inline function declarations from fabric.h */
+struct fi_info *wrap_fi_allocinfo(void);
+int wrap_fi_close(struct fid *fid);
+int wrap_fi_control(struct fid *fid, int command, void *arg);
+int wrap_fi_alias(struct fid *fid, struct fid **alias_fid, uint64_t flags);
+int wrap_fi_get_val(struct fid *fid, int name, void *val);
+int wrap_fi_set_val(struct fid *fid, int name, void *val);
+int wrap_fi_open_ops(struct fid *fid, const char *name, uint64_t flags,
+                 void **ops, void *context);
```
j-xiong: The alignment of the lines is off.  When breaking a long line of function prototype or function calls, the most common practice here is to align with the first argument after `(`. e.g.
```
ssize_t wrap_fi_writedata(struct fid_ep *ep, const void *buf, size_t len, void *desc,
						  uint64_t data, fi_addr_t dest_addr, uint64_t addr, uint64_t key,
						  void *context);

ssize_t wrap_fi_inject_writedata(struct fid_ep *ep, const void *buf, size_t len,
								 uint64_t data, fi_addr_t dest_addr, uint64_t addr,
								 uint64_t key);
```
Use tabs as many as possible and only use spaces to fill up the last gaps for the alignment.

Alternatively, use fixed number of tabs so that all lines except the first one are aligned. e.g.
```
ssize_t wrap_fi_writedata(struct fid_ep *ep, const void *buf, size_t len, void *desc,
						uint64_t data, fi_addr_t dest_addr, uint64_t addr, uint64_t key,
						void *context);

ssize_t wrap_fi_inject_writedata(struct fid_ep *ep, const void *buf, size_t len,
						uint64_t data, fi_addr_t dest_addr, uint64_t addr, uint64_t key);
```

Since these are new files, feel free to choose one of the styles, but please make it consistent. 

This comment applies to both `wrapper.c` and `wrapper.h`, entirely.

kyle-yh-kim: Understood. I'll use the following `.clang-format`;
```
BasedOnStyle: Google
IndentWidth: 4
ColumnLimit: 80
```

This rule abides the first option listed above. Sample result has been attached below.
```c
ssize_t wrap_fi_writedata(struct fid_ep* ep, const void* buf, size_t len,
                          void* desc, uint64_t data, fi_addr_t dest_addr,
                          uint64_t addr, uint64_t key, void* context) {
    return fi_writedata(ep, buf, len, desc, data, dest_addr, addr, key,
                        context);
}

ssize_t wrap_fi_inject_writedata(struct fid_ep* ep, const void* buf, size_t len,
                                 uint64_t data, fi_addr_t dest_addr,
                                 uint64_t addr, uint64_t key) {
    return fi_inject_writedata(ep, buf, len, data, dest_addr, addr, key);
}
```

j-xiong: that works for me.

bindings/rust/wrapper.c
```diff
@@ -0,0 +1,898 @@
...
+/* Proprietary helper function declarations. */
+
+fid_t get_fid_ptr(void* ptr) { return (fid_t)ptr; }
+
+/* Static inline function declarations from fi_prov.h */
+
+int wrap_fi_param_get_str(struct fi_provider* provider, const char* param_name,
+                          char** value) {
+    return fi_param_get_str(provider, param_name, value);
+}
```
j-xiong: This not the right style for function definitions. We follow the [Linux kernel coding style](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst) , function definitions look like this:
```
int function(int x)
{
        body of function
}
```

kyle-yh-kim: Understood. I'll just go ahead and run the existing `.clang-format`.

```
find ./bindings/rust/ -iname '*.h' -o -iname '*.c' | clang-format --style=file -i --files=/dev/stdin
```

Result;
```c
int wrap_fi_param_get_str(struct fi_provider *provider, const char *param_name,
			  char **value)
{
	return fi_param_get_str(provider, param_name, value);
}
```

j-xiong:
bot:aws:retest

j-xiong:
@shijin-aws I don't expect this PR to break AWS CI in any way. But to be on the safe side, could you double check that the AWS CI failure is unrelated?

kyle-yh-kim:
The AWS CI failure is unrelated to the PR.
I will wait for a second confirmation by @shijin-aws.

shijin-aws:
It was an infra issue, I restarted the test

shijin-aws:
bot:aws:retest

shijin-aws:
AWS CI is having issue now and , the failures are unrelated

PR #11361, opened 2025-08-29
prov/efa: Fix the rx sgl in data path direct

shijin-aws:
The current efa_data_path_direct_entry
doesn't shift the rx buf when walking the SGEs in the sge_list. This patch fixes it.

This is following rdma-core pattern while avoiding memset and memcpy


shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

PR #11367, opened 2025-09-02
prov/lnx, fabtests: add support for lnx in fabtests

aingerson:
A few things were missing to support running lnx with fabtests by default:
- fabtests was by default requesting AV_MAP if the provider returned UNSPEC which lnx wasn't handling properly (we should fix the handling in lnx too but fabtests shouldn't be using AV_MAP anymore because it was deprecated)
- lnx was missing a wait object implementation for internal util eq implementation
- lnx was incorrectly erroring out for multiple iovs even though it does support it now
- lnx was passing in the wrong protocol to the core providers and causing an initialization error
- lnx doesn't support a lot of features - added exclude file to be able to run it with runfabtests with proper exclusion of invalid tests and documentation of limitations and issues
- added lnx config file to run ubertest

### DIFF
--- a/fabtests/Makefile.am
+++ b/fabtests/Makefile.am
@@ -146,6 +146,7 @@ nobase_dist_config_DATA = \
 	test_configs/ucx/quick.test \
 	test_configs/ucx/ucx.exclude \
 	test_configs/lnx/lnx.exclude \
+	test_configs/lnx/all.test \
 	pytest/pytest.ini \
 	pytest/options.yaml \
 	pytest/conftest.py \
--- a/fabtests/test_configs/lnx/all.test
+++ b/fabtests/test_configs/lnx/all.test
@@ -1,0 +1,48 @@
+{
+	prov_name: lnx,
+	test_type: [
+		FT_TEST_LATENCY,
+		FT_TEST_BANDWIDTH,
+	],
+	class_function: [
+		FT_FUNC_SEND,
+		FT_FUNC_SENDV,
+		FT_FUNC_SENDMSG,
+		FT_FUNC_SENDDATA,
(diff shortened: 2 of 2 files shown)


aingerson:
Verified these changes work when enabling lnx fabtests in our CI which will come later.
All of this can wait until after the release

aingerson:
@j-xiong Can you review? Only failure is the udp rma bw. I'm going to add it to the CI once it's upstream

amirshehataornl (approved):

fabtests/test_configs/lnx/lnx.exclude
```diff
@@ -0,0 +1,40 @@
...
+rdm_pingpong
+multi_recv
+cntr_pingpong
+multi_client
+flood
+multi_ep
+cq_data
+av_xfer
+shared_av
+fi_rdm\b
```
j-xiong, line 29: what is the `\b` here?

j-xiong:
The udp failure can be eliminated with a rebase.

fabtests/test_configs/lnx/lnx.exclude
aingerson: \b matches a word boundary (includes spaces and end of string) so that matches to any test that references just the fi_rdm test ("fi_rdm" and "fi_rdm -U") which requires FI_MSG which lnx currently does not support. Without the \b it would match anything with "fi_rdm" in it which includes things like "fi_rdm_tagged_pingpong" which we don't want to exclude

aingerson:
@j-xiong You ok with merging without rerunning? Or do you want me to rebase and retest? If it's ok to go can you approve?

j-xiong (approved):

PR #11371, opened 2025-09-03
fabtests/prov/efa: Finalize the EP for RNR queue resend test

darrylabbate:
Otherwise, a race condition can cause either the client or server to hang without invoking the EP close procedure.


shijin-aws:
Fixed by https://github.com/ofiwg/libfabric/pull/11376

PR #11372, opened 2025-09-03
prov/efa: Shift the src_addr correctly in sreadfrom

shijin-aws:
currently, efa_cq_sreadfrom doesn't shift the src_addr in the same way as buffer in the consecutive cq polls, which is wrong. This patch fixes it.


jiaxiyan (approved):

shijin-aws:
Pure efa change, ignoring intel ci failure

PR #11373, opened 2025-09-04
prov/efa: Bypass util cq in the data path for efa-direct

shijin-aws:
A series of commits to make efa-direct bypass util cq in the data path


prov/efa/src/efa_base_ep.c
```diff
@@ -700,6 +700,20 @@ int efa_base_ep_check_qp_in_order_aligned_128_bytes(struct efa_base_ep *ep,
...
+
+
+/**
+ * @brief Set CQ ops for counter binding
+ *
+ * @param ep efa_base_ep
+ * @param cq efa_cq to configure
+ */
+static inline void efa_base_ep_set_cq_ops_for_cntr(struct efa_base_ep *ep, struct efa_cq *cq)
+{
```
shijin-aws, line 712: This overriding is not necessary if I make efa_cq_readfrom also have the cntr bump logics, but I currently choose to not do that to keep the this path as fast as possible.

prov/efa/src/efa_cq.c
```diff
@@ -745,16 +971,20 @@ int efa_cq_open(struct fid_domain *domain_fid, struct fi_cq_attr *attr,
...
 		break;
 	case FI_CQ_FORMAT_MSG:
 		cq->entry_size = sizeof(struct fi_cq_msg_entry);
+		cq->read_entry = efa_cq_read_msg_entry;
 		break;
 	case FI_CQ_FORMAT_DATA:
 		cq->entry_size = sizeof(struct fi_cq_data_entry);
+		cq->read_entry = efa_cq_read_data_entry;
 		break;
 	case FI_CQ_FORMAT_TAGGED:
```
jiaxiyan: Do you want to remove FI_CQ_FORMAT_TAGGED too?

prov/efa/src/efa_cq.c
```diff
@@ -569,6 +569,232 @@ int efa_cq_signal(struct fid_cq *cq_fid)
...
+	}
+	entry->len = efa_ibv_cq_wc_read_byte_len(cq);
+}
+
+static void efa_cq_read_data_entry(struct efa_ibv_cq *cq, void *buf, int opcode)
+{
+	struct fi_cq_data_entry *entry = buf;
+	struct ibv_cq_ex *ibv_cqx = cq->ibv_cq_ex;
+
+	if (!efa_cq_wc_is_unsolicited(cq) && ibv_cqx->wr_id) {
```
jiaxiyan: Can you make the duplicate code a helper function? Does that hurt performance?

prov/efa/src/efa_cq.c
```diff
@@ -569,6 +569,232 @@ int efa_cq_signal(struct fid_cq *cq_fid)
...
+			fi_strerror(err), err, efa_strerror(prov_errno), prov_errno);
+	efa_show_help(prov_errno);
+
+	if (buf->err_data_size > 0 && FI_VERSION_GE(efa_cq->util_cq.domain->fabric->fabric_fid.api_version, FI_VERSION(1, 5))) {
+		err_msg = (char *)buf->err_data;
+	} else {
+		err_msg = efa_cq->err_buf;
+	}
+
+	if (efa_write_error_msg(base_ep, addr, prov_errno, err_msg, &buf->err_data_size) != 0) {
```
jiaxiyan, line 648: Do you need OFI_UNLIKELY here?

prov/efa/src/efa_cq.c
```diff
@@ -569,6 +569,232 @@ int efa_cq_signal(struct fid_cq *cq_fid)
...
+	int err = to_fi_errno(prov_errno);
+
+	EFA_WARN(FI_LOG_CQ, "Encountered error during CQ polling. err: %s (%d), prov_errno: %s (%d)\n",
+			fi_strerror(err), err, efa_strerror(prov_errno), prov_errno);
+	efa_show_help(prov_errno);
+
+	if (buf->err_data_size > 0 && FI_VERSION_GE(efa_cq->util_cq.domain->fabric->fabric_fid.api_version, FI_VERSION(1, 5))) {
+		err_msg = (char *)buf->err_data;
+	} else {
+		err_msg = efa_cq->err_buf;
```
jiaxiyan, line 645: Where is `err_buf` set?

prov/efa/src/efa_cq.c
```diff
@@ -569,6 +569,232 @@ int efa_cq_signal(struct fid_cq *cq_fid)
...
+	if (ret > 0)
+		goto out;
+
+	/* No error, return EAGAIN */
+	if(!ibv_cq->ibv_cq_ex->status) {
+		ret = -FI_EAGAIN;
+		goto out;
+	}
+	efa_cq_fill_err_entry(ibv_cq, buf);
+	efa_cq_end_poll(ibv_cq);
```
jiaxiyan, line 777: Why is `efa_cq_end_poll` needed here?

shijin-aws: because we need to end the poll if it was active? Otherwise the subsequent cq_read will not call ibv_start_poll and get the next cqe

prov/efa/src/efa_cq.c
shijin-aws: sure

prov/efa/src/efa_cq.c
```diff
@@ -569,6 +569,232 @@ int efa_cq_signal(struct fid_cq *cq_fid)
...
+
+		opcode = efa_ibv_cq_wc_read_opcode(ibv_cq);
+		if (!efa_cq_wc_is_unsolicited(ibv_cq)) {
+			if (ibv_cq->ibv_cq_ex->wr_id) {
+				efa_cq->read_entry(ibv_cq, (void *)((uintptr_t) buf + num_cqe * efa_cq->entry_size), opcode);
+				if (src_addr)
+					src_addr[num_cqe] = efa_cq_get_src_addr(ibv_cq, opcode);
+				num_cqe++;
+			}
+		} else {
```
jiaxiyan: Can you improve this if else so you don't duplicate code?

prov/efa/src/efa_cq.c
jiaxiyan: Should it be under `out:` then?

shijin-aws: The expected workflow is 

- user call fi_cq_read, efa_cq_readfrom polls cqes and find a cqe have non-zero status, efa_cq_readfrom won't call end poll but just return. The return code can be either -FI_EAVAIL (when such failed cqe is the first one) or the number of successful completions. In whatever cases, the pointer of the current cqe was not shifted because it expects a cq_readerr call to process it.
- user call fi_cq_readerr, efa_cq_readerr read the current cqe polled in the last cq_readfrom call, it should expect a non-zero ibv_cq_ex->status here, and then process the failed cqe and end the poll, so the next cq_readfrom can call ibv_start_poll that polls the next cqe from device

However, there can be a use case that user call fi_cq_readerr for whatever reason before fi_cq_read returns an EINVAL, in this case the status will be 0, it should just do nothing and return EAGAIN without messing the device cq. That efa_cq_end_poll will likely be a no-op in this case as the cq should not be in a poll_active=true status

shijin-aws: Think a bit more, I will change the 

```
if(!ibv_cq->ibv_cq_ex->status) {
```
into 
```
if(!ibv_cq->poll_active) {
```
plus a `assert(ibv_cq->ibv_cq_ex->status)`
to make it more intuitive.

prov/efa/src/efa_cq.c
shijin-aws: it is now part of efa_cq struct created during the cq open

prov/efa/src/efa_cq.c
shijin-aws: sure

shijin-aws: Think about it more, it doesn't make sense to add OFI_UNLIKELY here as you want `efa_write_error_msg` to get called anyway. if you make that function call returned to a intermediate variable `ret`, you can do `OFI_UNLIKELY(ret)` in the if.

prov/efa/src/efa_cq.c
shijin-aws: Sure

prov/efa/src/efa_cq.c
shijin-aws: Sure, I can make the common code static inline

prov/efa/src/efa_cq.c
shijin-aws: Updated

prov/efa/src/efa_cq.c
shijin-aws: Fixed

prov/efa/src/efa_cq.c
shijin-aws: Updated

prov/efa/src/efa_cq.c
shijin-aws: Updated

jiaxiyan (approved):

PR #11375, opened 2025-09-04
prov/ucx: change ucx from FI_AV_MAP to FI_AV_TABLE

aingerson:
FI_AV_MAP is deprecated but ucx is coded to require it However, the hard requirement is not necessary as underneath it uses an array to manage peer addresses so we can easily support FI_AV_TABLE

Return the array index to the user through the fi_addr and change the lookup to support indexing instead of casting the fi_addr


prov/ucx/src/ucx.h
```diff
@@ -176,7 +176,9 @@ struct ucx_av {
 	struct ofi_dyn_arr addr_array;
 };
 
-#define UCX_GET_UCP_EP(EP, ADDR) (((struct ucx_ave*)(ADDR))->uep)
+#define UCX_GET_UCP_EP(EP, ADDR)					       \
+	(((struct ucx_ave *) ofi_array_at(&EP->av->ave_array, ADDR))->uep);
```
j-xiong: Missing parentheses around EP and ADDR.

j-xiong (commented):
There are least two other places that `fi_addr` is casted to `struct ucx_ave *`:
* in `fi_av_remove()` (`ucx_av.c`)
* in `ucx_get_rkey()` (`ucx_rma.c`)

aingerson:
@j-xiong Thanks! Updated.

aingerson:
@j-xiong the current code (in main) doesn't check for that and could segfault if used incorrectly. Do you want me to add that error checking as part of this patch?

j-xiong:
The existing code doesn't really have a good way to check the validity of fi_addr (other than NULL) since it is a pointer not an index. By moving to FI_AV_TABLE it is a good opportunity to do so.  Yes, I would prefer the checking added to this patch if possible.

shijin-aws:
AWS CI is broken now and we are prioritizing fix it

aingerson:
@j-xiong Ok updated again with error checking

include/ofi_indexer.h
```diff
@@ -306,7 +306,8 @@ ofi_array_item(struct ofi_dyn_arr *arr, char *chunk, int offset)
 
 static inline void *ofi_array_at(struct ofi_dyn_arr *arr, int index)
 {
-	assert(index <= OFI_IDX_MAX_INDEX);
+	if (index > OFI_IDX_MAX_INDEX)
+		return NULL;
```
j-xiong: This is not what I was talking about. A general ofi_array doesn't have a boundary and is only limited by OFI_IDX_MAX_INDEX. It's fine to use `assert` here.

For address vector, insertion doesn't have boundary, but query has (i.e., `av->count`).  For query, the boundary check can be done before calling `ofi_array_at`. Or optionally, we can have the boundary check inside a new function `ofi_array_at_max` which has a user passed in `max` paramter.

prov/ucx/src/ucx.h
```diff
@@ -176,7 +176,17 @@ struct ucx_av {
...
 };
 
-#define UCX_GET_UCP_EP(EP, ADDR) (((struct ucx_ave*)(ADDR))->uep)
+
+static inline ucp_ep_h ucx_get_ucp_ep(struct ucx_ep *ep, fi_addr_t addr)
+{
+	struct ucx_ave *ave;
+
+	ave = ofi_array_at(&ep->av->ave_array, addr);
+	if (addr >= ep->av->count || !ave)
```
j-xiong: boundary check should be done before calling ofi_array_at to avoid unnecessary array expansion. that's why a new function `ofi_array_at_max` may work better. Same for other places.

aingerson:
@j-xiong Ok I updated again to use a new function array_at_max so it never grows if we expect it to be in there. Let me know if that's more what you're thinking.
I know there's multiple ways to implement it but I went with tracking the maximum index within the array so we don't have to pass one in from the user but let me know if you prefer it to be passed in.

j-xiong:
I still prefer the parameter be passed in. Two reasons: (1) no extra overhead to existing `ofi_array_at` calls; (2) more flexible choice for the max value.

aingerson:
@j-xiong Ok updated again to pass in the max

prov/ucx/src/ucx.h
```diff
@@ -176,7 +176,10 @@ struct ucx_av {
 	struct ofi_dyn_arr addr_array;
 };
 
-#define UCX_GET_UCP_EP(EP, ADDR) (((struct ucx_ave*)(ADDR))->uep)
+#define UCX_GET_UCP_EP(EP, ADDR) 					\
+		((struct ucx_ave *)					\
+		ofi_array_at_max(&(EP)->av->ave_array, (ADDR),		\
+				 (EP)->av->count))->uep;
```
j-xiong: No `;` at the end. Instead, add parentheses around the entire block L180-182.

j-xiong (commented):
One last comment below. It's not cosmetic so needs to be fixed.

Also need to rebase to pick up AWS CI fixes.

aingerson:
Updated

zachdworkin:
current intel ci failure

fi_av_xfer -e rdm -p "ucx"   -E
 		  server_stdout: |
 		    fi_av_remove(): functional/av_xfer.c:91, ret=-22 (Invalid argument)
 		    AV address removal: FAIL

j-xiong:
> current intel ci failure
> 
> fi_av_xfer -e rdm -p "ucx" -E 
> server_stdout: | 
> fi_av_remove(): functional/av_xfer.c:91, ret=-22 (Invalid argument) 
> AV address removal: FAIL

This error is real. It's related to the code change in `ucx_av_remove`.

aingerson:
@j-xiong I think this should be ready to go now!

PR #11376, opened 2025-09-04
prov/efa: Fix the derivation from ope to pke

shijin-aws:
We cannot derive ope to pke via container_of,
as ope is a pointer in pke struct. This patch fixes this issue by accessing the pkt as the head of
the ope->queued_pkts dlist.

In that regard, it is generally not safe to derive pkt type from an ope. So efa_rdm_pkt_type_of
macro is getting removed by this patch as well and replaced by the direct calls to efa_rdm_pkt_type_of_pke


prov/efa/src/rdm/efa_rdm_pke_utils.h
```diff
@@ -29,11 +29,6 @@ struct efa_rdm_base_hdr *efa_rdm_pke_get_base_hdr(struct efa_rdm_pke *pke)
 	return (struct efa_rdm_base_hdr *)pke->wiredata;
 }
 
-#define efa_rdm_pkt_type_of(obj) _Generic((obj), \
-		struct efa_rdm_pke *: efa_rdm_pkt_type_of_pke, \
-		struct efa_rdm_ope *: efa_rdm_pkt_type_of_ope, \
```
darrylabbate, line 34: All you need to do is simply remove the single association and the corresponding function. No need to throw out the whole macro.

shijin-aws: As I mentioned above, I was not sure whether it is valuable to keep this macro... I can keep it for future benefits

shijin-aws:
The current fix is still not complete, it only fixes the RNR case, it did not fix the the queue due to local resource exhaustion. Will update later

jiaxiyan (dismissed):

shijin-aws:
Rebased upon https://github.com/ofiwg/libfabric/pull/11377

Both of these two PRs are needed to make AWS CI green.

shijin-aws:
AWS CI actually passed. The failure is due to cluster destruction failed

shijin-aws:
bot:aws:retest

sunkuamzn (approved):

PR #11380, opened 2025-09-05
prov/util: configure control_progress with the user provided progress mode

snadampal:
Domain control_progress attribute was added earlier to provide a lockless statemachine in libfabric with FI_THREAD_DOMAIN threading and FI_PROGRESS_CONTROL_UNIFIED control_progress mode. However there was no way for a user to provide the required control_progress attribute. This patch is adding the support for allowing user to override the control_progress along with the other domain attributes.

### DIFF
--- a/prov/util/src/util_attr.c
+++ b/prov/util/src/util_attr.c
@@ -1264,8 +1264,10 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
 
 	if (hints->threading)
 		attr->threading = hints->threading;
-	if (hints->progress)
+	if (hints->progress) {
 		attr->progress = hints->progress;
+		attr->control_progress = hints->progress;
+	}
 	if (hints->av_type)
 		attr->av_type = hints->av_type;
 	if (hints->max_ep_auth_key)


shijin-aws:
I am surprised we need this change as we have been testing with such progress mode. @a-szegel can you take a look

a-szegel:
I don't think this hurts, but the other utils where it is used get it from the domain:
https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_cq.c#L740-L741

```
 grep -rI -A 2 ofi_progress_lock_type
./prov/util/src/util_av.c:	ep_list_lock_type = ofi_progress_lock_type(domain->threading,
./prov/util/src/util_av.c-						   domain->control_progress);
./prov/util/src/util_av.c-
--
./prov/util/src/util_cq.c:	ep_list_lock_type = ofi_progress_lock_type(cq->domain->threading,
./prov/util/src/util_cq.c-						   cq->domain->control_progress);
./prov/util/src/util_cq.c-	ret = ofi_genlock_init(&cq->ep_list_lock, ep_list_lock_type);
--
./prov/util/src/util_cntr.c:	ep_list_lock_type = ofi_progress_lock_type(cntr->domain->threading,
./prov/util/src/util_cntr.c-						   cntr->domain->control_progress);
./prov/util/src/util_cntr.c-	ret = ofi_genlock_init(&cntr->ep_list_lock, ep_list_lock_type);
```

Which gets set here: https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_domain.c#L119

snadampal:
thanks for the quick review @shijin-aws and @a-szegel . Without this user alter support, the domain attribute is configured to the default `FI_PROGRESS_UNSPEC`. So, during the `ep_list_lock_type` check the domain is always returning the unspecified and hence enforcing locking for CQ handling.

At least this is what is happening with the nccl ofi plugin.

a-szegel (approved):

shijin-aws:
Tested with main branch, indeed, without this PR, the lock is a mutex (0), progress is FI_PROGRESS_AUTO
```
(env) sjina@p5en-odcr-queue-dy-p5en48xlarge-21:~/PortaFiducia/build/libraries/libfabric/direct_cq/source/libfabric$ fi_rdm_pingpong --data-progress manual --control-progress unified -E -p efa
ep_list_lock_type: 0, domain control_progress: 1
ep_list_lock_type: 0, domain control_progress: 1
```

j-xiong:
libfabric 2.x has moved away from separate control and data progress. Now a single `progress` field is used. The `control_progress` fiedd is there for backward compatibility only.

`FI_PROGRESS_CONTROL_UNIFIED` is a progress mode that capture both data progress and control progress.  I think what really needs to do here is either change how `ofi_progress_lock_type` is called or update `control_progress` based on `progress` setting (at least for FI_PROGRESS_CONTROL_UNIFIED).

a-szegel:
We added `FI_PROGRESS_CONTROL_UNIFIED` instead of checking whether `data_progress == control_progress` in order to remove the "control interfaces are always thread safe regardless of threading model promise".  Now that we deprecated control_progress... the name doesn't make as much sense, but idk what we can do at this point.  We still need a way to break the "control interface are thread safe" promise.

This PR fixes an implementation bug that I made, and I think we do want this.

j-xiong:
`FI_PROGRESS_CONTROL_UNIFIED` is not added for replacing the `data_progress == control_progress` check. It is added to provide an exception for "control interface is always thread safe" and the implication is stronger than the above check. 

For current providers (2.x), the `control_progress` from the hints should be ignored. Inside the provider, whether locking is needed should be based on `domain_attr->threading` and `domain_attr->progress`.

snadampal:
Hi @j-xiong , 

>For current providers (2.x), the control_progress from the hints should be ignored.

Then I'm curious how it can be configured from the application.

> whether locking is needed should be based on domain_attr->threading and domain_attr->progress.

Today I see that `efa` and `psm3` providers are using the `control_progress` for decision.  If you all agree that the locking decision should be based on `progress` instead of `control_progress`, I can fix it in the efa provider. 

Hi @a-szegel , from high level understanding of the efa/rdma provider, I think `progress` attribute is not being used in any decision making, so,  we can safely let applications configure `progress` = `FI_PROGRESS_CONTROL_UNIFIED` to achieve lockless state machine without affecting any other functionality. What do you think?

j-xiong:
> Hi @j-xiong ,
> 
> > For current providers (2.x), the control_progress from the hints should be ignored.
> 
> Then I'm curious how it can be configured from the application.

Application only configures a single `progress`, which can be `FI_PROGRESS_AUTO`, `FI_PROGRESS_MANUAL` or `FI_PROGRESS_CONTROL_UNIFIED`.

snadampal:
Hi @j-xiong , looking into the `data_progress`, `control_progress` and `progress` attributes further i see that only at the domain attributes level they were unified partially 
~~~
struct fi_domain_attr {
        struct fid_domain       *domain;
        char                    *name;
        enum fi_threading       threading;
        enum fi_progress        control_progress; --> This is still left as a separate attribute
        union {
                enum fi_progress data_progress;
                enum fi_progress progress;
        };
~~~

as part of this commit

~~~
commit b51cb0f7e9a3f1a441f2f47f55f8bd30b5fbc34e
Author: Sean Hefty <sean.hefty@intel.com>
Date:   Mon Sep 25 16:01:41 2023 -0700

    core: Simplify progress definition
    
    Combine data and control progress into one progress option.
~~~

However, the `util_domain` still doesn't have this unified. so, we can't just yet switch to `progress` without updating the `util_domain` and everywhere else...

~~~
struct util_domain {
        struct fid_domain       domain_fid;
        struct dlist_entry      list_entry;
        struct util_fabric      *fabric;
...
        enum fi_threading       threading;
        enum fi_progress        data_progress;
        enum fi_progress        control_progress;
};        
~~~

Given this state, I think the best option right now is having application configure only `progress` but we internally populate both `data_progress` and `control_progress` attributes with the user provided configuration and continue to use them as they are today., something like following. Please let me know if this is acceptable.

~~~
@@ -1266,6 +1266,7 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
                attr->threading = hints->threading;
        if (hints->progress)
                attr->progress = hints->progress;
+              attr->control_progress = hints->progress;
        if (hints->av_type)
                attr->av_type = hints->av_type;
        if (hints->max_ep_auth_key)
~~~
and user sets 
~~~
hints->domain_attr->progress = FI_PROGRESS_CONTROL_UNIFIED; 
~~~

j-xiong:
Yes, I think this approach is fine. BTW, the `control_progress` field left in `domain_attr` is to avoid breaking the ABI.

snadampal:
thanks, @j-xiong , I have updated the PR, please let me know if this looks good now.

shijin-aws:
@snadampal can you rebase your PR on top of the latest main. There are several fixes merged today to make AWS CI green

snadampal:
@shijin-aws , done!

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

shijin-aws:

j-xiong:
One failure in mpichtestsuite over tcp:
```
Running tests in ./threads/init [00:02:06]
Program mult_session exited without No Errors
Unexpected output in mult_session:
    I_MPI_JOB_TIMEOUT = 180 second(s): job ending due to application timeout
```

This happened before, unlikely related to this PR.

I will give it one more try to see if anything changes.

shijin-aws:
@j-xiong any update on the testing ^^ ?

j-xiong:
@shijin-aws There was a different failure (still part of mpichtestsuite) over tcp due to job launcher issue on one node. I rebooted the node and restarted the test.

shijin-aws:
CI passed. @j-xiong good to go?

PR #11394, opened 2025-09-10
prov/efa: Add unit-tests for lock types

shijin-aws:
This is a follow up for the fix in

https://github.com/ofiwg/libfabric/commit/396cd4a38b680d0083536d85c3edae794316a97e

### DIFF
--- a/prov/efa/test/efa_unit_test_ep.c
+++ b/prov/efa/test/efa_unit_test_ep.c
@@ -1826,3 +1826,39 @@ void test_efa_rdm_ep_data_path_direct_disabled(struct efa_resource **state)
 
 	assert_false(efa_ep->qp->data_path_direct_enabled);
 }
+
+/**
+ * @brief Verify endpoint lock uses no-op locking with FI_THREAD_COMPLETION
+ *
+ * @param[in] state struct efa_resource managed by the framework
+ */
+void test_efa_ep_lock_type_no_op(struct efa_resource **state)
+{
+	struct efa_resource *resource = *state;
+	struct efa_base_ep *efa_base_ep;
+	resource->hints = efa_unit_test_alloc_hints(FI_EP_RDM, EFA_DIRECT_FABRIC_NAME);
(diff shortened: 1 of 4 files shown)


jiaxiyan (approved):

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -1826,3 +1826,39 @@ void test_efa_rdm_ep_data_path_direct_disabled(struct efa_resource **state)
...
+ *
+ * @param[in] state struct efa_resource managed by the framework
+ */
+void test_efa_ep_lock_type_no_op(struct efa_resource **state)
+{
+	struct efa_resource *resource = *state;
+	struct efa_base_ep *efa_base_ep;
+	resource->hints = efa_unit_test_alloc_hints(FI_EP_RDM, EFA_DIRECT_FABRIC_NAME);
+	assert_non_null(resource->hints);
+	resource->hints->domain_attr->threading = FI_THREAD_COMPLETION;
```
snadampal, line 1841: the progress hint is missing here
resource->hints->domain_attr->progress = FI_PROGRESS_CONTROL_UNIFIED;

the default one is AUTO I think and doesn't make the lock no-op.

snadampal (changes requested):
please check the review comment

prov/efa/test/efa_unit_test_ep.c
shijin-aws: That is intended, because the current util_ep.lock doesn't depend on the progress model

shijin-aws: So this part is not broken before your PR, I just add these tests together

snadampal (approved):

prov/efa/test/efa_unit_test_ep.c
snadampal: looks good to me then, I have approved it now.

PR #11401, opened 2025-09-11
prov/efa: Make implicit AV unbounded by default

sunkuamzn:
Implicit AV size of 1024 is too small for certain MPI applications.
Make it unbounded by default.

### DIFF
--- a/prov/efa/src/efa_env.c
+++ b/prov/efa/src/efa_env.c
@@ -40,7 +40,7 @@ struct efa_env efa_env = {
 	.use_unsolicited_write_recv = 1,
 	.internal_rx_refill_threshold = 8,
 	.use_data_path_direct = true,
-	.implicit_av_size = 1024,
+	.implicit_av_size = 0,
 };
 
 /* @brief Read and store the FI_EFA_* environment variables.
@@ -228,7 +228,10 @@ void efa_env_define()
 	fi_param_define(&efa_prov, "implicit_av_size", FI_PARAM_SIZE_T,
 			"The maximum size of the implicit AV used to store AV "
 			"entries of peers that were not explicitly inserted "
-			"into the AV by the application",
+			"into the AV by the application. Setting this variable "
+			"to a positive value will enforce the the maximum "
+			"size. Setting this value to 0 will allow unbounded "
+			"growth of the implicit AV. (Default: 0)",
 			efa_env.implicit_av_size);
 }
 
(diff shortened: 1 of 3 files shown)


j-xiong:
Can the AV size be increased dynamically (using ofi_array)? Not a request for this PR, just a suggestion for future improvement.

sunkuamzn:
@j-xiong thank you for the suggestion. The implicit AV is based on the util_av which uses a bufpool and does allow for dynamic growth. After more discussion, we also though this is the best solution. I've updated the PR.

prov/efa/src/efa_env.c
```diff
@@ -225,10 +225,12 @@ void efa_env_define()
 		"the CQ polling and TX/RX submissions, when it's available. Setting this variable as 0"
 		"will disable this feature (Default: %d)",
 		efa_env.use_data_path_direct);
-	fi_param_define(&efa_prov, "implicit_av_size", FI_PARAM_SIZE_T,
+	fi_param_define(&efa_prov, "implicit_av_size", FI_PARAM_INT,
 			"The maximum size of the implicit AV used to store AV "
 			"entries of peers that were not explicitly inserted "
-			"into the AV by the application",
+			"into the AV by the application. Setting this value to "
+			"-1 will allow unbounded growth of the implicit AV. "
```
shijin-aws: Shouldn't it be something like "setting this value to a positive value will enforce a maximum size ... by default it is unlimited "

prov/efa/src/efa_env.c
```diff
@@ -112,7 +112,7 @@ void efa_env_param_get(void)
 	fi_param_get_size_t(&efa_prov, "tx_size", &efa_env.tx_size);
 	fi_param_get_size_t(&efa_prov, "rx_size", &efa_env.rx_size);
 	fi_param_get_size_t(&efa_prov, "internal_rx_refill_threshold", &efa_env.internal_rx_refill_threshold);
-	fi_param_get_size_t(&efa_prov, "implicit_av_size", &efa_env.implicit_av_size);
+	fi_param_get_int(&efa_prov, "implicit_av_size", &efa_env.implicit_av_size);
```
shijin-aws: I am not sure if it's better to keep it as size_t and make default as 0 which means unlimited size, to be consistent with the buffer pool behavior ?

sunkuamzn: I thought a value of 0 might be a legitimate use case for some applications to disable the implicit AV

shijin-aws: I do not think we should expect such use case to be legitimate today considering efa provider doesn't support FI_SOURCE_ERR. Today disabling implicit av will cause any unknown peer to be silently dropped which is not friendly to applications

sunkuamzn: Updated

PR #11402, opened 2025-09-11
prov/util: Remove unmatched rx entry from unspec unexpected queue

sunkuamzn:
The EFA provider AV keeps track of peers that any endpoint bound to the AV has received messages from but were not explicitly inserted into the AV by the application. The provider limits the number of such peers to prevent runaway memory usage.

When the EFA provider has to remove such an implicitly inserted peer, it also needs to remove any unmatched entries in the SRX queues associated with the peer. The EFA provider calls free_entry to remove the rx entry. But free_entry does not remove the rx entry from the unspec_unexp_[msg,tag]_queue dlist which corrupts the dlist.

This commit adds logic to remove any entry that is about to be freed from the unspec_unexp_[msg,tag]_queue.

### DIFF
--- a/prov/util/src/util_srx.c
+++ b/prov/util/src/util_srx.c
@@ -390,6 +390,9 @@ static int util_queue_tag(struct fi_peer_rx_entry *rx_entry)
 static void util_free_entry(struct fi_peer_rx_entry *entry)
 {
 	struct util_srx_ctx *srx;
+	struct util_unexp_peer *unexp_peer;
+	struct slist *queue;
+	struct slist_entry *item, *prev;
 	struct util_rx_entry *util_entry, *owner_entry;
 
 	srx = (struct util_srx_ctx *) entry->srx->ep_fid.fid.context;
@@ -410,6 +413,30 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
 			ofi_buf_free(owner_entry);
 		}
 	}
+
+	if (util_entry->status == RX_ENTRY_UNEXP) {
+		if (util_entry->peer_entry.addr == FI_ADDR_UNSPEC) {
+			dlist_remove(&util_entry->d_entry);
+		} else {
+			unexp_peer = ofi_array_at(&srx->src_unexp_peers,
+						  util_entry->peer_entry.addr);
+			queue = util_entry->peer_entry.flags & FI_TAGGED ?
+					&unexp_peer->tag_queue :
+					&unexp_peer->msg_queue;
+
+			slist_foreach(queue, item, prev) {
+				if ((struct util_rx_entry *) item == util_entry)
+					slist_remove(queue, item, prev);
+			}
+
+			if (!--unexp_peer->cnt) {
+				assert(slist_empty(&unexp_peer->msg_queue) &&
+				       slist_empty(&unexp_peer->tag_queue));
+				dlist_remove(&unexp_peer->entry);
+			}
+		}
+	}
+
 	ofi_buf_free(util_entry);
 }
 


prov/util/src/util_srx.c
```diff
@@ -397,6 +397,14 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
 			ofi_buf_free(owner_entry);
 		}
 	}
+
+	/* If free_entry was called but the addr is still FI_ADDR_UNSPEC, the rx
+	 * entry must be in the unspec_unexp_[msg,tag]_queue. Remove it so that
+	 * the queue does not get corrupted. */
+	if (entry->addr == FI_ADDR_UNSPEC) {
+		dlist_remove(&util_entry->d_entry);
+	}
```
j-xiong, line 438: The braces can be got rid of here.

prov/util/src/util_srx.c
```diff
@@ -397,6 +397,14 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
 			ofi_buf_free(owner_entry);
 		}
 	}
+
+	/* If free_entry was called but the addr is still FI_ADDR_UNSPEC, the rx
+	 * entry must be in the unspec_unexp_[msg,tag]_queue. Remove it so that
+	 * the queue does not get corrupted. */
+	if (entry->addr == FI_ADDR_UNSPEC) {
```
shijin-aws: if an fi_trecv is called with FI_ADDR_UNSPEC, the util_rx_entry will be created with addr as FI_ADDR_UNSPEC as well and put in the expected rx queue. Then is this logic still correct? I forgot when `free_entry` will be called, it's before matching or after, or both

sunkuamzn:
Discussed offline with @aingerson. This PR also needs to handle the case when the rx entry is matched to a packet. This is related to @shijin-aws's comment. There doesn't seem to be an easy fix that covers all cases.

My previous PR (https://github.com/ofiwg/libfabric/pull/11401) is enough to stabilize our tests on the v2.3.x branch. So I'm removing the v2.3.x label from this PR to unblock the v2.3 release

fabtests/common/shared.c
```diff
@@ -1611,6 +1611,43 @@ int ft_exchange_addresses_oob(struct fid_av *av_ptr, struct fid_ep *ep_ptr,
...
+
+	ret = ft_sock_send(oob_sock, buf, FT_MAX_CTRL_MSG);
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
+/* Receive and insert peer address out-of-band */
+int ft_recv_addr_oob(struct fid_av *av_ptr, fi_addr_t *remote_addr)
```
shijin-aws, line 1615: Can these two functions used by the oob exchange code in this shared.c?

include/ofi_util.h
```diff
@@ -1287,6 +1287,12 @@ struct ofi_ops_flow_ctrl {
 			ssize_t (*send_handler)(struct fid_ep *ep, uint64_t credits));
 };
 
+enum util_rx_entry_status {
+	RX_ENTRY_UNMATCHED = 0,
+	RX_ENTRY_UNMATCHED_UNEXP,
```
shijin-aws: some comments will be helpful here. Also, I think the `unmatched` actually maps to `unexpected`, `unexp` maps `unspec` according to the current convention

prov/util/src/util_srx.c
```diff
@@ -408,6 +408,12 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
 			ofi_buf_free(owner_entry);
 		}
 	}
+
+	if (OFI_UNLIKELY(util_entry->status == RX_ENTRY_UNMATCHED_UNEXP &&
+			 util_entry->peer_entry.addr == FI_ADDR_UNSPEC)) {
```
shijin-aws: is the second condition still necessary?

sunkuamzn: Yes. An RX entry in `RX_ENTRY_UNMATCHED_UNEXP` state can also be in the peer-specific list which is an `slist`

shijin-aws: Huh, I thought `RX_ENTRY_UNMATCHED_UNEXP` means it is in the unspec queue , after moving to the peer-specific list, both status and addr should be updated?

sunkuamzn: `RX_ENTRY_UNMATCHED` - RX entry corresponding to a receive posted by the application that is not yet matched to a packet. It's in either `srx->msg_queue` or one of the peer-specific `srx->src_recv_queues`

`RX_ENTRY_UNMATCHED_UNEXP` - RX entry corresponding to a received packet that is not yet matched to any application posted receive. The RX entry is in `srx->unspec_unexp_msg_queue` or one of the peer specific `unexp_peer->msg_queue`

`RX_ENTRY_MATCHED` - RX entry corresponding to a matched pair of application posted receive and packet. It is not in any list tracked by SRX. It is freed by the owner after the receive operation is completed.

I will add comments explaining what each status represents

fabtests/common/shared.c
sunkuamzn: fixed

include/ofi_util.h
sunkuamzn: @aingerson do you have any thoughts on the names?

prov/util/src/util_srx.c
shijin-aws: >RX_ENTRY_UNMATCHED_UNEXP - RX entry corresponding to a received packet that is not yet matched to any application posted receive. The RX entry is in srx->unspec_unexp_msg_queue or one of the peer specific unexp_peer->msg_queue

This looks foggy to me. I would prefer we have two states for `unspec_unexp_msg_queue` and `unexp_peer->msg_queue`

include/ofi_util.h
shijin-aws: EFA's rxe state has the following names

```
enum efa_rdm_ope_state {
...
	EFA_RDM_RXE_UNEXP,	/**< rxe unexp msg waiting for post recv */
	EFA_RDM_RXE_MATCHED,	/**< rxe matched with RTM */
...
};
```

`UNEXP` means it is in unexpected queue (but peer is known)
`MATCHED` means it is not in any queue 

Maybe we can have `UNSPEC` for unspec queue, and `EXP` for expected queue

aingerson: I was confused when I first read it but I think it makes sense with the current naming. Sai isn't trying to differentiate between unspec or not unspec because that's done with the fi_addr. The distinction he's making with the states is to figure out where in the matching flow it is - unmatched (posted rx), unmatched (unexp), matched (not unexpected, being processed). Maybe the use of the "unmatched" in the unexpected case is the confusing part? Would this be better?
RX_ENTRY_POSTED,
RX_ENTRY_MATCHED,
RX_ENTRY_UNEXP

prov/util/src/util_srx.c
sunkuamzn: > I would prefer we have two states for unspec_unexp_msg_queue and unexp_peer->msg_queue

Why? The state of the RX entry is the same. The difference is whether the source peer is inserted into the receiver's AV or not. We already track that in the `addr` field of the `peer_rx_entry`.

prov/util/src/util_srx.c
```diff
@@ -408,6 +411,27 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
...
+						  util_entry->peer_entry.addr);
+			queue = &unexp_peer->msg_queue;
+
+			slist_foreach(queue, item, prev) {
+				if ((struct util_rx_entry *) item == util_entry)
+					slist_remove(queue, item, prev);
+			}
+
+			if (slist_empty(queue)) {
+				assert(!unexp_peer->cnt);
```
shijin-aws: This is great, after this change, I think efa provider doesn't have to do such extra cleanup in ep close https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L995-L1019. It can just release all rxe and call util_srx_close, as all peer rxe should be removed from the dlist or slist in the efa rxe releases as it calls free_entry

Can be in a separate PR

prov/util/src/util_srx.c
shijin-aws: If we are sure peer specific unexp queue won't have addr  FI_ADDR_UNSPEC, I am good with such plan

prov/util/src/util_srx.c
```diff
@@ -408,6 +411,27 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
 			ofi_buf_free(owner_entry);
 		}
 	}
+
+	if (OFI_UNLIKELY(util_entry->status == RX_ENTRY_UNMATCHED_UNEXP)) {
```
aingerson: I generally don't like using OFI_UNLIKELY because it builds bad habits for not understanding code and generally doesn't help that much with modern compilers. Please remove unless you can justify it with data.

prov/util/src/util_srx.c
sunkuamzn: SRX just sets `peer_rx_entry->addr` to whatever the owner provider passes in when calling `get_msg`
https://github.com/ofiwg/libfabric/blob/8a3de5a14609b34ae93bf2ba9ce23c36a7c20f7b/prov/util/src/util_srx.c#L108

So, as long as the owner provider provides the correct addr, it will work

prov/util/src/util_srx.c
aingerson: I think the logic here is off. You removed the entry but didn't decrement it so I don't think the assert will pass. I think it should be something more like 
```
if (!--unexp_peer->cnt) {
    assert(slist_empty(queue));
    dlist_remove(&unexp_peer->entry);
}
```

include/ofi_util.h
shijin-aws: I am good with this naming

sunkuamzn: Updated the names

prov/util/src/util_srx.c
sunkuamzn: Fixed

prov/util/src/util_srx.c
sunkuamzn: Fixed

prov/util/src/util_srx.c
```diff
@@ -410,6 +413,27 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
...
 		}
 	}
+
+	if (util_entry->status == RX_ENTRY_UNEXP) {
+		if (util_entry->peer_entry.addr == FI_ADDR_UNSPEC) {
+			dlist_remove(&util_entry->d_entry);
+		} else {
+			unexp_peer = ofi_array_at(&srx->src_unexp_peers,
+						  util_entry->peer_entry.addr);
+			queue = &unexp_peer->msg_queue;
```
aingerson: eek realizing now that you'll have to check both the msg and tag queues to find it... 😬 
or you can check the queue based on the flags which *should* be set (but you should double check)
ie
```
queue = util_entry->peer_entry.flags & FI_TAGGED ? unexp_peer->tag_queue : unexp_peer->msg_queue
```

prov/util/src/util_srx.c
```diff
@@ -410,6 +413,27 @@ static void util_free_entry(struct fi_peer_rx_entry *entry)
...
+
+			slist_foreach (queue, item, prev) {
+				if ((struct util_rx_entry *) item == util_entry)
+					slist_remove(queue, item, prev);
+			}
+
+			if (!--unexp_peer->cnt) {
+				assert(slist_empty(queue));
```
aingerson: and then this can expand to
`assert(slist_empty(&unexp_peer->msg_queue) && slist_empty(&unexp_peer->tag_queue);`

prov/util/src/util_srx.c
sunkuamzn: done

prov/util/src/util_srx.c
sunkuamzn: done

aingerson:
Intel CI failure was just the recent udp/rxd rma failure. Everything else passed so this PR should be good!

PR #11405, opened 2025-09-11
prov/efa: Ignore stale cqes from destroyed qps

shijin-aws:
Currently, efa provider always write cq errors when ibv_start_poll/ibv_next_poll returns EINVAL, which means the cqes are from destroyed QPs (not found in the qp table). This behavior is not complying with libfabric man page that

said the outstanding operations posted to the endpoint when fi_close is called will be silently dropped without generating completions.


shijin-aws:
bot:aws:retest

shijin-aws:
@j-xiong I have seen the appveyor/pr CI frequently failed at tcp provider fabtests. Maybe we should take a look to make it stable.

Issue #11362, opened 2025-08-29
prov/efa,tcp and others: Properly return an existing domain/fabric when calling fi_getinfo() with an existing fabric / domain name


shijin-aws:
**Describe the bug**
Multiple providers are not complying with the man page for the domain/fabric returned by fi_getinfo

https://ofiwg.github.io/libfabric/main/man/fi_domain.3.html

>On input to fi_getinfo, a user may set this to an opened domain instance to restrict output to the given domain. On output from fi_getinfo, if no domain was specified, but the user has an opened instance of the named domain, this will reference the first opened instance. If no instance has been opened, this field will be NULL.
The domain instance returned by fi_getinfo should only be considered valid if the application does not close any domain instances from another thread while fi_getinfo is being processed.

https://ofiwg.github.io/libfabric/main/man/fi_fabric.3.html

>fabric
On input to fi_getinfo, a user may set this to an opened fabric instance to restrict output to the given fabric. On output from fi_getinfo, if no fabric was specified, but the user has an opened instance of the named fabric, this will reference the first opened instance. If no instance has been opened, this field will be NULL.
The fabric instance returned by fi_getinfo should only be considered valid if the application does not close any fabric instances from another thread while fi_getinfo is being processed.

It seems multiple providers, including efa and tcp are not complying with this requirement, that `fi_getinfo` always return a NULL fabric/domain in domain_attr/fabric_attr

**To Reproduce**
See attached program

```
tar -xf single-fabric-getinfo.tar.gz
cd single-fabric-getinfo
make
./single-fabric-getinfo


(env) sjina@ip-10-0-2-245:~/single-fabric-getinfo$ ./single-fabric-getinfo 
mtl provider: tcp
mtl fabric name: 10.0.2.0/24
mtl domain name: ens130
btl fabric hint: 10.0.2.0/24
btl provider: tcp
btl fabric name: 10.0.2.0/24
btl domain name: ens130
btl fabric: (nil) mtl_fabric: 0x64e560adf6c0
btl domain: (nil) mtl_domain: 0x64e560ae44f0
```

**Expected behavior**
It should return an opened domain/fabric object in the output fi_info, like 

```
(env) sjina@ip-10-0-2-245:~/single-fabric-getinfo$ ./single-fabric-getinfo 
mtl provider: tcp
mtl fabric name: 10.0.2.0/24
mtl domain name: ens130
btl fabric hint: 10.0.2.0/24
btl provider: tcp
[...]

PR #11408, opened 2025-09-12
prov/util, efa: Return matching fabric/domain in fi_getinfo

jiaxiyan:
On input to fi_getinfo, user may set fabric/domain to an opened instance or passed the name. Return existing fabric/domain if they match.


jiaxiyan:
This fixes  https://github.com/ofiwg/libfabric/issues/11362

prov/efa/src/efa_user_info.c
```diff
@@ -477,6 +503,35 @@ int efa_get_user_info(uint32_t version, const char *node,
...
 		ofi_alter_info(dupinfo, hints, version);
 
+		/* On input to fi_getinfo, a user may set this to an opened
+		 * fabric instance to restrict output to the given fabric. */
+		if (hints && hints->fabric_attr)
+			dupinfo->fabric_attr->fabric = hints->fabric_attr->fabric;
+		if (hints && hints->domain_attr)
+			dupinfo->domain_attr->domain = hints->domain_attr->domain;
+
+		if (!dupinfo->fabric_attr->fabric && !dupinfo->domain_attr->domain)
```
shijin-aws, line 606: If application only specify domain_attr->name and fabric_attr->name, will `util_lookup_existing_fabric_domain` give us demanded behavior for efa-provider ?

prov/efa/src/efa_user_info.c
```diff
@@ -477,6 +503,35 @@ int efa_get_user_info(uint32_t version, const char *node,
 
 		ofi_alter_info(dupinfo, hints, version);
 
+		/* On input to fi_getinfo, a user may set this to an opened
+		 * fabric instance to restrict output to the given fabric. */
+		if (hints && hints->fabric_attr)
+			dupinfo->fabric_attr->fabric = hints->fabric_attr->fabric;
```
shijin-aws: So if application pass the ptr of the opened fabric and domain name, we will just use it without doing any check?

prov/efa/src/efa_user_info.c
jiaxiyan: It always returns the fabric, not necessarily the domain, so we check domain again with efa_find_domain

prov/efa/src/efa_user_info.c
jiaxiyan: > On input to fi_getinfo, a user may set this to an opened domain instance to restrict output to the given domain. On output from fi_getinfo, if no domain was specified, but the user has an opened instance of the named domain, this will reference the first opened instance. If no instance has been opened, this field will be NULL.

I don't know if I read the man page right. Maybe it only supports passing by name because I don't see other providers do this.

shijin-aws: Let's open a thread to get a clarification from the community

jiaxiyan:
7b9bcb6b21bd09752331d7465ed8515af2c6f4c8 fixes the TCP provider by comparing to hints fabric name when it is passed.

prov/efa/src/efa_user_info.c
shefty: If the user passed in a specific instance of the fabric or domain, it must be used.  However, other checks still apply.  If the opened instance cannot be used because it can't support some endpoint attribute, for instance, the fi_getinfo() call should fail.

It's not expected that the app would request a specific instance of the fabric or domain, versus leaving it unset, and letting the provider return if an existing instance is usable.  But in theory it should work.

prov/util/src/util_main.c
```diff
@@ -165,6 +165,44 @@ static int util_find_domain(struct dlist_entry *item, const void *arg)
...
+					const struct fi_info *hints,
+					struct fi_info **info)
+{
+	struct util_fabric *fabric;
+	struct util_domain *domain;
+	struct dlist_entry *item;
+	struct util_fabric_info fabric_info;
+
+	fabric_info.name = (hints && hints->fabric_attr && hints->fabric_attr->name) ?
+			    hints->fabric_attr->name : (*info)->fabric_attr->name;
```
shefty: nit: The line length looks long.

Are you saying that the check for the tcp provider is wrong because the name stored in the opened instance differs from the more generic name assigned to an uninitialized, hard-coded fi_info being used as a base?

There are 2 changes being combined here.  Please apply the fix to the existing code (so it can be backported more easily), then move the code into a separate function.

If the check against the fabric name is wrong, is the check against the domain name also wrong for tcp?

jiaxiyan: Yes. I will separate commits and update the domain name check too.

prov/efa/src/efa_user_info.c
jiaxiyan: Thanks for the clarification! I will update it accordingly.

jiaxiyan:
@shefty Can you give the prov/util commits another review?

shefty (dismissed):
The util changes look good to me.  Thanks!

j-xiong (approved):

prov/efa/src/efa_user_info.c
```diff
@@ -477,6 +547,38 @@ int efa_get_user_info(uint32_t version, const char *node,
 
 		ofi_alter_info(dupinfo, hints, version);
 
+		/* On input to fi_getinfo, a user may set this to an opened
+		 * fabric/domain instance to restrict output to the given fabric/domain. */
+		ret = efa_user_info_check_fabric_object(hints, dupinfo);
```
shijin-aws: Here we just use the opened fabric/domain, did we do the check as Sean mentioned earlier

> If the user passed in a specific instance of the fabric or domain, it must be used. However, other checks still apply. If the opened instance cannot be used because it can't support some endpoint attribute, for instance, the fi_getinfo() call should fail.

jiaxiyan: I thought it was handled by `ofi_check_info`, but looking closer I need to check the fabric_attr of `hints->fabric_attr->fabric`, not just `hints->fabric_attr`

shijin-aws:
bot:aws:retest

prov/efa/src/efa_user_info.c
```diff
@@ -477,6 +590,38 @@ int efa_get_user_info(uint32_t version, const char *node,
...
+			goto free_info;
+
+		ret = efa_user_info_check_domain_object(hints, dupinfo);
+		if (ret)
+			goto free_info;
+
+		if (!dupinfo->fabric_attr->fabric && !dupinfo->domain_attr->domain)
+			util_lookup_existing_fabric_domain(&efa_util_prov, hints, &dupinfo);
+
+		if (hints && hints->domain_attr && hints->domain_attr->name &&
```
shijin-aws, line 609: let's say you have 4 NICs on the host, there will be 4*3 (dgram, rdm, direct) info constructed in prov_info. can you tell me how many info will be returns if user pass an existing domain instance or name in the hints?

shijin-aws: I think it should return 3 infos, but reading the code it seems the loop will still return me 12 and each info will have domain_attr->domain as the existing domain. Am I wrong here?

jiaxiyan: If passing domain by name, it is already filtered by `efa_prov_info_compare_domain_name`. 
If passing domain by pointer, I add the name check in `efa_user_info_check_domain_object` so it always returns one info now.

prov/efa/src/efa_user_info.c
```diff
@@ -477,6 +594,42 @@ int efa_get_user_info(uint32_t version, const char *node,
 
 		ofi_alter_info(dupinfo, hints, version);
 
+		/* On input to fi_getinfo, a user may set this to an opened
+		 * fabric/domain instance to restrict output to the given fabric/domain. */
+		ret = efa_user_info_check_fabric_object(hints, dupinfo, prov_info);
```
shijin-aws, line 598: If ret is non-zero, what is the downside to continue? This check 

```
	ret = ofi_check_fabric_attr(efa_util_prov.prov, 
				    prov_info->fabric_attr, 
				    &fabric_attr);
``` 

Is prov_info dependent, so we should still move to the next info and run the check?

prov/efa/src/efa_user_info.c
```diff
@@ -477,6 +594,42 @@ int efa_get_user_info(uint32_t version, const char *node,
...
+		if (ret)
+			goto free_info;
+
+		ret = efa_user_info_check_domain_object(hints, dupinfo);
+		if (ret != 0) {
+			if (ret > 0) {
```
shijin-aws: Similar comment as earlier, why can't we always continue? Then you do not need to have an addition rc 1 for special handling

prov/efa/src/efa_user_info.c
jiaxiyan: If `ofi_check_fabric_attr` fails we should continue, but if it is not a valid fabric object, there is no point in continuing.

shijin-aws: Yeah, but it doesn't hurts right? It will finally skip all info.. Just want to call out that it can make the code cleaner

jiaxiyan: For code readability, I will just make it continue because this is not on the fast path.

shijin-aws (approved):

shijin-aws:
@j-xiong any ideas on Intel CI is still running?

j-xiong:
@shijin-aws The CI nodes have not come back yet. See #11424

shijin-aws:
@j-xiong I can see intel CI is passing for newer PRs. Any insights why this PR (and some older PRs) are still in progress?

j-xiong:
There are still some certificate related issues being debugged, at the same time some PRs are manually picked to run on alternative nodes.

PR #11425, opened 2025-09-17
prov/cxi: fix performance issue with close_mc()

SeanP-2023:
Reduce the amount of times we alloc/free root rdma buffers and take wait penalty only once


swelch:
AppVeyor failure is related to tcp provider. This change is restricted to cxi collectives.

PR #11429, opened 2025-09-19
prov/efa: Fail early in rma when > 1 iov is passed

shijin-aws:
efa-direct offloads the rma request to rdma wqes in efa device directly. But today efa device doesn't support > 1 iov. This patch makes efa_post_read and efa_post_write fail early with clear error message for this scenario to avoid unexpected crashing.


prov/efa/src/efa_rma.c
```diff
@@ -56,8 +56,10 @@ static inline ssize_t efa_rma_post_read(struct efa_base_ep *base_ep,
 		ofi_total_iov_len(msg->msg_iov, msg->iov_count),
 		msg->addr, (size_t) msg->context, flags);
 
-	assert(msg->iov_count > 0 &&
-	       msg->iov_count <= base_ep->domain->info->tx_attr->iov_limit);
+	if (OFI_UNLIKELY(msg->iov_count > 1)) {
```
amitrad-aws: There is "max_sge_rd" inside efa_device.ibv_attr that should hold this value.
We should use it to be future compatible

shijin-aws: Thank you! I will use it

shijin-aws: Updated

shijin-aws:
The latest change broke unit test on platforms that don't support rdma, will update

Zhenye-Na:
bot:aws:retest

shijin-aws:
bot:aws:retest

prov/efa/src/efa_rma.c
shijin-aws: Think about it more, the current data path direct code is already hard coded to only support 1 sge https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_io_defs.h#L125-L129. Though it is the same code as rdma-core and the firmware code today, if rdma-core and firmware change in the future and make max_sge_rd returned by rdma-core increase in the future, we will need to accommodate in libfabric to follow up, otherwise we will get unexpected crash. In that regard maybe having a hard-coded 1 limit is better. 

WDYT @amitrad-aws @mrgolin

prov/efa/src/efa_rma.c
```diff
@@ -56,8 +56,10 @@ static inline ssize_t efa_rma_post_read(struct efa_base_ep *base_ep,
 		ofi_total_iov_len(msg->msg_iov, msg->iov_count),
 		msg->addr, (size_t) msg->context, flags);
 
-	assert(msg->iov_count > 0 &&
-	       msg->iov_count <= base_ep->domain->info->tx_attr->iov_limit);
+	if (OFI_UNLIKELY(msg->iov_count > base_ep->domain->device->ibv_attr.max_sge_rd)) {
```
sunkuamzn: This check happens even for `fi_read` calls which are guaranteed to only provide a single IOV. If we don't want to iterate and want the check, it should be in `efa_rma_readv`, `efa_rma_readmsg` and the equivalent functions for write and RDM protocol paths

sunkuamzn (commented):
`fi_read` and `fi_write` can only provide a single buffer. `fi_readv` and `fi_readmsg` can provide multiple IOVs

So I think we should just implement our `readv` and `readmsg` to iterate over all IOV and post multiple read/write requests

prov/efa/src/efa_rma.c
shijin-aws: RDM protocol path doesn't have such issue, it currently only construct 1 WQE per IOV

prov/efa/src/efa_rma.c
shijin-aws: Per offline discussion, we will introduce a defined macro for libfabric internally and use the min of the macro and the rdma-core return value

prov/efa/src/efa_rma.c
sunkuamzn: The RDM path looks messy

We have asserts like this
https://github.com/ofiwg/libfabric/blob/223d0b604f146bfa73be2cf0ac9867e7307228dc/prov/efa/src/rdm/efa_rdm_ope.c#L1478

But we also copy the `iov_cnt`
https://github.com/ofiwg/libfabric/blob/223d0b604f146bfa73be2cf0ac9867e7307228dc/prov/efa/src/rdm/efa_rdm_rma.c#L66
https://github.com/ofiwg/libfabric/blob/223d0b604f146bfa73be2cf0ac9867e7307228dc/prov/efa/src/rdm/efa_rdm_ope.c#L42

shijin-aws: But if you read the logic below in the while loop

https://github.com/ofiwg/libfabric/blob/223d0b604f146bfa73be2cf0ac9867e7307228dc/prov/efa/src/rdm/efa_rdm_ope.c#L1568

It actually only run 1 iov per efa_rdm_pke_write, the code is really messy.

shijin-aws: The txe and msg can have multi iov, it will finally aggregate all the pkes from it

sunkuamzn: Yeah I saw that. Because we iterate and post a single IOV at a time, we should remove the asserts.

shijin-aws: But the asserts you referred to are the ope/msg ones, which should still be restricted by the tx iov limit...

prov/efa/test/efa_unit_test_rma.c
```diff
@@ -19,6 +19,8 @@ static void test_efa_rma_prep(struct efa_resource *resource, fi_addr_t *addr)
 	base_ep = container_of(resource->ep, struct efa_base_ep, util_ep.ep_fid);
 	/* Add rma caps explicitly to ep->info to allow local test */
 	base_ep->info->caps |= FI_RMA;
+	/* Mock the max rdma sge to allow local test */
+	base_ep->domain->device->ibv_attr.max_sge_rd = EFA_DEVICE_MAX_RDMA_SGE;
```
amitrad-aws: If you really want to test your minimum you shuold put a larger value

prov/efa/src/efa_rma.c
mrgolin: Please do this only when using Libfabric datapath implementation, otherwise it should be transparent.

shijin-aws: It will make non-datapath-direct not fail early with the same warning message, though rdma-core itself is checking it and return EINVAL, but it's not in libfabric's logging system and users won't know what happened immediately

prov/efa/src/efa_rma.c
sunkuamzn: Should they? I'm not sure why we should have asserts that are not enforced by the code that follows. That just leads to different behavior with debug and production builds

If we want to restrict the number of IOVs to `tx_iov_limit`, then we need a check in production builds too.

shijin-aws: I think they should be restricted as it is what we returned for the tx iov limit in fi_getinfo. We can add a check for sure

prov/efa/src/efa_rma.c
```diff
@@ -124,6 +140,10 @@ ssize_t efa_rma_readmsg(struct fid_ep *ep_fid, const struct fi_msg_rma *msg, uin
 	if (err)
 		return err;
 
+	err = efa_rma_check_iov_count(base_ep, msg->iov_count);
```
mrgolin: I think this should be called from ```efa_data_path_direct_wr_set_sge_list``` or adjacent to the calls to this function since 1) it's directly related to WQE format, 2) we should avoid enforcing Libfabric hardcoded limits when using rdma-core data path.

prov/efa/test/efa_unit_test_rma.c
shijin-aws: I removed this mock as we are not doing the minimal logic any more. Updated the unit test for the set_sge_list check only

prov/efa/src/efa_rma.c
shijin-aws: Update.

prov/efa/src/efa_rma.c
shijin-aws: I decided to make the assertion removal in a separate PR. Focusing on adding the rdma iov limit check in the data path direct code only

prov/efa/src/efa_rma.c
shijin-aws: The latest revision should do what @mrgolin is proposing. resolving this conversation

prov/efa/src/efa_data_path_direct_entry.h
```diff
@@ -318,6 +318,11 @@ static inline void efa_data_path_direct_wr_set_sge_list(struct efa_qp *efa_qp,
 		break;
 	case EFA_IO_RDMA_READ:
 	case EFA_IO_RDMA_WRITE:
+		if (OFI_UNLIKELY(num_sge > EFA_DEVICE_MAX_RDMA_SGE)) {
```
sunkuamzn, line 321: This check adds latency to some libfabric calls that are guaranteed to only provide a single IOV

shijin-aws: I would regard it as a cheap call to put it in the common code instead of having more complicated checks in the readv/writev/readmsg/writemsg everywhere.

prov/efa/src/efa.h
```diff
@@ -112,6 +112,15 @@
  */
 #define EFA_RDM_BUFPOOL_ALIGNMENT	(64)
 
+
+/**
+ * The hard-coded rdma sge num limit
+ * due to the structs layout rdma wqes.
+ * See `struct efa_io_rdma_req` in efa_io_defs.h
+ */
+#define EFA_DEVICE_MAX_RDMA_SGE 1
```
mrgolin, line 121: After moving the check you can now use ```EFA_IO_TX_DESC_NUM_RDMA_BUFS``` from efa_io_defs.h

shijin-aws: Thanks, opened a follow-up PR https://github.com/ofiwg/libfabric/pull/11489

PR #11431, opened 2025-09-23
prov/psm3: Fix integer overflow in psm3_ips_scbctrl_init()

tatarintsevsv:
Setting PSM3_NUM_SEND_BUFFERS to a large enough value results in an integer overflow and subsequent segfault.


tatarintsevsv:
> It's sufficient to just cast `bufsize` though.

I agree, that would be enough. Is there a need to make such changes?

j-xiong:
@tatarintsevsv No need to change.

PR #11432, opened 2025-09-23
prov/shm: make smr buffer pool allocations cache line aligned

snadampal:
Fixed the smr buffer pool alignment to match cache line size on x86 and aarch64 cpus. This has improved the p99 latencies and also throughput for nccl collective tests.

### DIFF
--- a/prov/shm/src/smr.h
+++ b/prov/shm/src/smr.h
@@ -72,6 +72,11 @@
 #ifndef _SMR_H_
 #define _SMR_H_
 
+/*
+ * Set alignment to aarch64 and x86 cache line size.
+ */
+#define SHM_SMR_BUFPOOL_ALIGNMENT       (64)
+
 struct smr_env {
 	size_t sar_threshold;
 	int disable_cma;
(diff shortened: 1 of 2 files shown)


aingerson:
@snadampal Can you share some benchmarks and numbers your ran?

prov/shm/src/smr.h
```diff
@@ -72,6 +72,11 @@
 #ifndef _SMR_H_
 #define _SMR_H_
 
+/*
+ * Set alignment to aarch64 and x86 cache line size.
+ */
+#define SHM_SMR_BUFPOOL_ALIGNMENT       (64)
```
aingerson, line 78: remove shm prefix

snadampal: @aingerson thanks for the prompt review. I'm running some internal benchmarks, will try to collect data that I can share here. 

>remove shm prefix

:) I started with `SMR_BUFPOOL_ALIGNMENT` name and in the last minute I've added SHM prefix to make it provider specific. sure, I will remove the prefix it to match the other definitions related to smr.

j-xiong: @snadampal SMR is the prefix for the shm provider. It is used instead of SHM to avoid any potential conflict with UNIX/Linux IPC "shm".

aingerson:
@snadampal Any update on this? Note that there has been a shm rearchitecture since this was opened resulting in conflicts. Shouldn't be too hard to resolve though

PR #11433, opened 2025-09-23
prov/efa: Allocate cq err_buf on demand

shijin-aws:
efa_cq->err_buf is only needed by efa-direct cq which can store the err data on the util cq bypass path. Make it allocated on demand during the efa-direct
cq open to save memory.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -860,6 +860,9 @@ int efa_cq_close(fid_t fid)
 	}
 #endif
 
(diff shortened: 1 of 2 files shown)


a-szegel:
What is the benefit of reducing the size of the structure?

shijin-aws:
> What is the benefit of reducing the size of the structure?

because we shouldn't allocate more memory when we do not need?

jiaxiyan (approved):

a-szegel (approved):

PR #11436, opened 2025-09-23
[v2.1.x] prov/efa and fabtests: add completion traces and fixes ep finalize semantics

alekswn:
backport https://github.com/ofiwg/libfabric/pull/11342


shijin-aws:
@j-xiong is Intel CI still down?

j-xiong:
@shijin-aws Unfortunately yes, but we are getting close to get it back.

PR #11438, opened 2025-09-23
prov/verbs: Return PCI bus information with fi_getinfo()

j-xiong:


shijin-aws (approved):

prov/verbs/src/verbs_info.c
```diff
@@ -533,6 +533,56 @@ static const char *vrb_link_layer_str(uint8_t link_layer)
...
+	pci++; /* skip '/' */
+
+	n = sscanf(pci, "%x:%x:%x.%x", &a, &b, &c, &d);
+	if (n != 4) {
+		VRB_WARN(FI_LOG_FABRIC,
+			 "Failed to parse pci bus address: %s\n", pci);
+		goto free_real_path;
+	}
+
+	bus_attr->attr.pci.domain_id = a;
```
shijin-aws, line 573: Since you are here, would u mind fixing efa provider's code https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_device.c#L554-L559 on using intermediate a/b/c/d so we will not have garbage value returned when sscanf failed

shijin-aws: Nvm, I see efa code will free this attr anyway in the error path

j-xiong:
Added a commit to fix windows build failure.

shefty (approved):

j-xiong:
@shefty Thanks!

PR #11437, opened 2025-09-23
fabtests/efa: cuda dmabuf validation logic

nmazzilli3:
test: Adding cuda dmabuf validation logic

Problem:
  - Users was submitting fabtests without the --do-dmabuf-reg-for-hmem flag

Solution:
  - Added hmem cuda logic changes based on nvidia manual [https://docs.nvidia.com/cuda/gpudirect-rdma/ here]
  - Added check_dmabuf to init cuda internals and get dmabuf support information
  - Added conftest to validate these checks if a user specifies cuda command in fabtests
  - Updated fabtest logger statements to include timestamp

Testing:
  - Validated tests with --do-dmabuf-reg-for-hmem flag and without it on p6-gb200 cluster
  - All tests skipped with cuda command
  - make -j install && python3 install/bin/runfabtests.py --expression "cuda" -vvv -p /home/nmazzill/libfabric/fabtests/install/bin/ --junit-xml ft_`git branch --show-current`_`git rev-parse HEAD`_all_junit.xml --nworkers 16 -b efa 10.0.123.149 10.0.121.190 | tee ft_`git branch --show-current`_`git rev-parse HEAD`_all_stdout
  - All tests passed with this cuda command
  - make -j install && python3 install/bin/runfabtests.py --expression "cuda" --do-dmabuf-reg-for-hmem -vvv -p /home/nmazzill/libfabric/fabtests/install/bin/ --junit-xml ft_`git branch --show-current`_`git rev-parse HEAD`_all_junit.xml --nworkers 16 -b efa 10.0.123.149 10.0.121.190 | tee ft_`git branch --show-current`_`git rev-parse HEAD`_all_stdout

PR #11443, opened 2025-09-25
fabtests/efa: cuda dmabuf validation logic

nmazzilli3:
Problem:
  - Users was submitting fabtests without the --do-dmabuf-reg-for-hmem flag

Solution:
  - Added hmem cuda logic changes based on nvidia manual [https://docs.nvidia.com/cuda/gpudirect-rdma/ here]
  - Added check_dmabuf to init cuda internals and get dmabuf support information
  - Added conftest to validate these checks if a user specifies cuda command in fabtests
  - Updated fabtest logger statements to include timestamp

Testing:
  - Validated tests with --do-dmabuf-reg-for-hmem flag and without it on p6-gb200 cluster
  - All tests skipped with cuda command
  - make -j install && python3 install/bin/runfabtests.py --expression "cuda" -vvv -p /home/nmazzill/libfabric/fabtests/install/bin/ --junit-xml ft_`git branch --show-current`_`git rev-parse HEAD`_all_junit.xml --nworkers 16 -b efa 10.0.123.149 10.0.121.190 | tee ft_`git branch --show-current`_`git rev-parse HEAD`_all_stdout
  - All tests passed with this cuda command
  - make -j install && python3 install/bin/runfabtests.py --expression "cuda" --do-dmabuf-reg-for-hmem -vvv -p /home/nmazzill/libfabric/fabtests/install/bin/ --junit-xml ft_`git branch --show-current`_`git rev-parse HEAD`_all_junit.xml --nworkers 16 -b efa 10.0.123.149 10.0.121.190 | tee ft_`git branch --show-current`_`git rev-parse HEAD`_all_stdout

Sim Issue:
  - N/A


fabtests/common/hmem_cuda.c
```diff
@@ -113,43 +117,134 @@ static int ft_cuda_pointer_set_attribute(void *buf)
...
+ *    indicates whether cudaHostAlloc() memory can be exported as dma-buf.
+ *  - On compute capability >= 10 (Blackwell), we force GPU_DIRECT_RDMA=0
+ *    regardless of attribute value to align with the deprecation notice.
+ *
+ * @return FI_SUCCESS on success
+ *        -FI_EIO on CUDA API error
+ *        Sets global flags dmabuf_supported, p2p_supported, and cuda_memory_support.
  */
-static int ft_cuda_hmem_detect_dmabuf_support(void)
+static int ft_cuda_detect_memory_support(void)
```
darrylabbate, line 170: This seems really well thought-out, and I think we should consider adding something like this to the core HMEM code. Especially if GPUDirect RDMA is on a deprecation path.

fabtests/include/shared.h
```diff
@@ -369,8 +369,9 @@ static inline int ft_use_size(int index, int enable_flags)
 #define FT_LOG(level, fmt, ...)						\
 	do {								\
 		int saved_errno = errno;				\
-		fprintf(stderr, "[%s] fabtests:%s:%d: " fmt "\n",	\
-			level, __FILE__, __LINE__, ##__VA_ARGS__);	\
+		int64_t ns = ft_gettime_ns();				\
+		fprintf(stderr, "[%s] %lu.%06lu fabtests:%s:%d: " fmt "\n",	\
+			level, ns/1000000000UL, (ns%1000000000UL)/1000UL, __FILE__, __LINE__, ##__VA_ARGS__);	\
```
darrylabbate: This should be a separate commit

darrylabbate:
Piggybacking on @jiaxiyan's comments here (https://github.com/ofiwg/libfabric/pull/11437#issuecomment-3330972244): we don't necessarily need to split commits based on which parts of the codebase are touched, but we do want to keep commits as atomic as possible.

fabtests/include/shared.h
nmazzilli3: ack will remove this outside of this pr

fabtests/common/hmem_cuda.c
```diff
@@ -113,43 +117,134 @@ static int ft_cuda_pointer_set_attribute(void *buf)
...
+    cuda_ret = cuda_ops.cuDeviceGetAttribute(&dma_buf_attr,
+        CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED, dev);
+    if (cuda_ret != CUDA_SUCCESS) {
+        ft_cuda_driver_api_print_error(cuda_ret, "cuDeviceGetAttribute(DMA_BUF_SUPPORTED)");
+        cuda_memory_support = CUDA_MEMORY_SUPPORT__UNKNOWN;
+        return -FI_EIO;
+    }
+
+    cuda_ret = cuda_ops.cuDeviceGetAttribute(&p2p_attr,
+        CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_SUPPORTED, dev);
```
darrylabbate: Let's avoid mixing `p2p` with GPUDirect RDMA, since p2p (sometimes GPUDirect P2P) typically refers to intra-node GPU to GPU communication in Nvidia nomenclature.

fabtests/include/shared.h
darrylabbate: You're free to include it in this PR. It simply needs to be its own commit.

fabtests/common/hmem_cuda.c
nmazzilli3: could you add a sim ticket since this would be another piece of atomic work touching something different

fabtests/common/hmem_cuda.c
nmazzilli3: What is the request here or recommended change?

fabtests/include/shared.h
nmazzilli3: Is the goal to have single atomic changes in a single PR or not? I have removed it for now

fabtests/common/hmem_cuda.c
darrylabbate: The variable name here (`p2p_attr`) as well as other vars/enums like `CUDA_MEMORY_SUPPORT__DMA_P2P_BOTH` should change s.t. we're not conflating p2p with GPUDirect RDMA. E.g. `gdr_attr` and `CUDA_MEMORY_SUPPORT__DMA_GDR_BOTH`

nmazzilli3: These names were from whiteboarding with @sunkuamzn. Sai are you onboard with these name changes I don't have all the context either way

fabtests/common/check_cuda_dmabuf.c
```diff
@@ -0,0 +1,67 @@
+/*
```
sunkuamzn, line 1:
```
/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
```

is shorter and also acceptable

fabtests/common/check_cuda_dmabuf.c
```diff
@@ -0,0 +1,67 @@
...
+ *
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <getopt.h>
+#include <string.h>
+#include <shared.h>
+#include <hmem.h>
+
+static cuda_memory_support_e dmabuf_viable_and_supported( void )
```
sunkuamzn:
```suggestion
static cuda_memory_support_e dmabuf_viable_and_supported(void)
```

fabtests/include/shared.h
sunkuamzn: Looks like it's still here

fabtests/pytest/efa/conftest.py
```diff
@@ -99,6 +105,77 @@ def rma_fabric(cmdline_args, fabric):
...
+            break
+        print(f"[retry] cuda_memory_support NOT_INITIALIZED (attempt {attempt+1}/{max_attempts})")
+        time.sleep(2)
+    
+
+    if cuda_support == CudaMemorySupport.NOT_INITIALIZED:
+        pytest.fail("CUDA memory support never initialized")
+    
+    do_dmabuf = cmdline_args.do_dmabuf_reg_for_hmem
+    if ( do_dmabuf is None and
```
sunkuamzn:
```suggestion
    if (do_dmabuf is None and 
```

fabtests/include/hmem.h
```diff
@@ -35,6 +35,14 @@
 #include <rdma/fi_domain.h>
 #include <rdma/fi_errno.h>
 
+typedef enum {
+	CUDA_MEMORY_SUPPORT__NOT_INITIALIZED = -1,
+    CUDA_MEMORY_SUPPORT__UNKNOWN = 0,
+    CUDA_MEMORY_SUPPORT__DMA_BUF_ONLY = 1,
+    CUDA_MEMORY_SUPPORT__P2P_ONLY= 2, 
+    CUDA_MEMORY_SUPPORT__DMA_P2P_BOTH = 3,
```
sunkuamzn:
```suggestion
    CUDA_MEMORY_SUPPORT__DMABUF_P2P_BOTH = 3,
```

DMA is overloaded, so using dmabuf is better

fabtests/common/hmem_cuda.c
sunkuamzn: I'm OK with `GDR` instead of `P2P`

fabtests/common/hmem_cuda.c
```diff
@@ -113,43 +117,134 @@ static int ft_cuda_pointer_set_attribute(void *buf)
...
+    }
+
+    FT_INFO("Compute capability %d.%d", cc_major, cc_minor);
+    FT_INFO("dmabuf_supported=%s", dmabuf_supported ? "true" : "false");
+    FT_INFO("GPU_DIRECT_RDMA_SUPPORTED raw=%d -> p2p_supported=%s",
+            p2p_attr, p2p_supported ? "true" : "false");
+
+    // Final truth table
+    if (!p2p_supported && !dmabuf_supported) {
+        cuda_memory_support = CUDA_MEMORY_SUPPORT__UNKNOWN;
```
sunkuamzn: Do you need an enum for no GDR support and no DMABUF support?

fabtests/pytest/efa/efa_common.py
```diff
@@ -1,9 +1,67 @@
...
 from common import SshConnectionError, is_ssh_connection_error, has_ssh_connection_err_msg, ClientServerTest
 from retrying import retry
 
+
+class CudaMemorySupport(IntEnum):
+    NOT_INITIALIZED = -1
+    UNKNOWN = 0
+    DMA_BUF_ONLY = 1
+    P2P_ONLY = 2
+    DMA_P2P_BOTH = 3
```
sunkuamzn:
```suggestion
    DMABUF_P2P_BOTH = 3
```

fabtests/pytest/efa/conftest.py
```diff
@@ -99,6 +105,77 @@ def rma_fabric(cmdline_args, fabric):
...
+    """Auto-run CUDA validation if CUDA tests are present."""
+    # Get the -k expression that runfabtests.py passes to pytest
+    debug = False
+    if debug:
+        print(f"Running the cuda_validation_fixture w requests {request}")
+        print(f"cmdline_args.__dict__: {cmdline_args.__dict__}")
+    expression = request.config.getoption('-k', default=None)
+    if debug:
+        print(f"Expression from -k: {expression}")
+    if expression and 'cuda' in expression.lower():
```
sunkuamzn: You need run this check in the memory type fixtures. We automatically parametrize over all memory types, so we can still run CUDA fabtests even if `cuda` is not in the test selection string

https://github.com/ofiwg/libfabric/blob/7d60749967a425fff75dd36df03757ab871b0b4e/fabtests/pytest/efa/conftest.py#L19-L25

fabtests/pytest/efa/conftest.py
```diff
@@ -99,6 +105,77 @@ def rma_fabric(cmdline_args, fabric):
...
+    return
+
+
+@pytest.fixture(scope="function", autouse=True)
+def cuda_validation_fixture(request, cmdline_args):
+    """Auto-run CUDA validation if CUDA tests are present."""
+    # Get the -k expression that runfabtests.py passes to pytest
+    debug = False
+    if debug:
+        print(f"Running the cuda_validation_fixture w requests {request}")
```
sunkuamzn: I would also remove these debug prints and the boolean `debug`

fabtests/pytest/efa/conftest.py
```diff
@@ -99,6 +105,77 @@ def rma_fabric(cmdline_args, fabric):
         pytest.skip("FI_RMA is not supported. Skip rma tests on efa-direct.")
     return fabric
 
+
+def cuda_memory_type_validation(cmdline_args, run_cuda_val_b: bool = False, max_attempts: int = 3):
```
sunkuamzn: Why do you need `run_cuda_val_b`? I don't think you need `max_attempts` and the retry logic either. The retries can be at the pytest level like you've added

fabtests/common/hmem_cuda.c
```diff
@@ -113,43 +117,134 @@ static int ft_cuda_pointer_set_attribute(void *buf)
...
+ *   - https://developer.nvidia.com/blog/cuda-toolkit-12-8-delivers-nvidia-blackwell-support/
+ *
+ * NVIDIA deprecated GPUDirect RDMA (nv-p2p APIs) starting with Blackwell
+ * (compute capability >= 10). Applications must migrate to dma-buf.
+ *
+ * Truth table for effective support (device memory only):
+ *
+ *  DMA_BUF   GPU_DIRECT_RDMA   Result
+ *  -------   ----------------  ------------------------
+ *     0            0           UNKNOWN
```
sunkuamzn: If you know that both dmabuf and GDR are not supported, the result shouldn't be `UNKNOWN`

sunkuamzn (changes requested):

fabtests/common/check_cuda_dmabuf.c
nmazzilli3: will change this

fabtests/common/check_cuda_dmabuf.c
nmazzilli3: will remove space

fabtests/pytest/efa/conftest.py
nmazzilli3: will remove whitespace

fabtests/include/hmem.h
nmazzilli3: Ok will change name

fabtests/common/hmem_cuda.c
nmazzilli3: what should it be?

fabtests/common/hmem_cuda.c
nmazzilli3: ok ill change all variables of p2p to gdr per comments

fabtests/common/hmem_cuda.c
nmazzilli3: From a test perspective this could definitely happen, from a device perspective based on your comment im assuming this is not feasible.

fabtests/pytest/efa/conftest.py
nmazzilli3: sure

fabtests/pytest/efa/efa_common.py
nmazzilli3: ok will changhe enum

fabtests/pytest/efa/conftest.py
nmazzilli3: Are you recommending we remove all retry logic and the input variable to skip this check altogether?

fabtests/pytest/efa/conftest.py
nmazzilli3: this is good to know what would this code fix make sense 

`@pytest.fixture(scope="function", autouse=True)
def cuda_validation_fixture(request, cmdline_args):
    """Auto-run CUDA validation if CUDA tests are present."""
    # Check if the current test has cuda_memory mark
    has_cuda_mark = any(mark.name == 'cuda_memory' for mark in request.node.iter_markers())`

fabtests/pytest/efa/conftest.py
sunkuamzn: yes

fabtests/common/hmem_cuda.c
sunkuamzn: `NOT_SUPPORTED` would be good

nmazzilli3: ok made this change

fabtests/include/shared.h
nmazzilli3: its no longer there

fabtests/pytest/efa/conftest.py
nmazzilli3: retested on p6-gb200 cluster and this will work

fabtests/pytest/efa/conftest.py
nmazzilli3: done

fabtests/common/check_cuda_dmabuf.c
```diff
@@ -0,0 +1,37 @@
...
+#include <getopt.h>
+#include <string.h>
+#include <shared.h>
+#include <hmem.h>
+
+static cuda_memory_support_e dmabuf_viable_and_supported(void)
+{
+	cuda_memory_support_e cuda_memory_support = ft_cuda_memory_support();
+
+	return cuda_memory_support;
```
shijin-aws: Why not just use `ft_cuda_memory_support` ?

shijin-aws:
Can you rephrase your commit title to be consistent with the PR title

fabtests/common/check_cuda_dmabuf.c
nmazzilli3: can fix the return to just call the ft_cuda_memory_support() func. Kept it there for an easier debug point

nmazzilli3:
> Can you rephrase your commit title to be consistent with the PR title
To ensure I'm understanding this correctly, you'd like me to add 

fabtests/efa: cuda dmabuf validation logic to the top of the commit message and then keep everything else below it?

shijin-aws:
> > Can you rephrase your commit title to be consistent with the PR title
> > To ensure I'm understanding this correctly, you'd like me to add
> 
> fabtests/efa: cuda dmabuf validation logic to the top of the commit message and then keep everything else below it?

yes

shijin-aws (approved):

shijin-aws:
@j-xiong can u review the fabtests common code change, thanks

fabtests/pytest/efa/conftest.py
```diff
@@ -99,6 +105,57 @@ def rma_fabric(cmdline_args, fabric):
...
+
+
+@pytest.fixture(scope="function", autouse=True)
+def cuda_validation_fixture(request, cmdline_args):
+    """Auto-run CUDA validation if CUDA tests are present."""
+    # Check if the current test has cuda_memory mark
+    has_cuda_mark = any(mark.name == 'cuda_memory' for mark in request.node.iter_markers())
+    
+    if has_cuda_mark:
+        print("Running CUDA validation")
```
jiaxiyan: Can you remove the `print` in this function? It will show up in all the test outputs, which is too verbose.

fabtests/common/check_cuda_dmabuf.c
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-2-Clause OR GPL-2.0-only */
+/* SPDX-FileCopyrightText: Copyright Amazon.com, Inc. or its affiliates. All rights reserved. */
```
j-xiong: Please follow the convention of multiline comments here.

fabtests/common/check_cuda_dmabuf.c
```diff
@@ -0,0 +1,35 @@
...
+#include <stdio.h>
```
j-xiong, line 6: Add an empty line before this.

fabtests/include/hmem.h
```diff
@@ -35,6 +35,14 @@
 #include <rdma/fi_domain.h>
 #include <rdma/fi_errno.h>
 
+typedef enum {
+	CUDA_MEMORY_SUPPORT__NOT_INITIALIZED = -1,
+    CUDA_MEMORY_SUPPORT__NOT_SUPPORTED = 0,
+    CUDA_MEMORY_SUPPORT__DMA_BUF_ONLY = 1,
+    CUDA_MEMORY_SUPPORT__GDR_ONLY= 2, 
+    CUDA_MEMORY_SUPPORT__DMABUF_GDR_BOTH  = 3,
```
j-xiong: Alignment is off, most like due to space being used instead of tabs.

fabtests/include/hmem.h
```diff
@@ -35,6 +35,14 @@
...
+} cuda_memory_support_e;
```
j-xiong: Don't use typedef here. Just define the enum, and use `enum enum_name var_name` to define variables.

I would use `FT_CUDA_xxx` instead of `CUDA_xxx` to avoid name space confusion. So for the lowercase names.

fabtests/common/check_cuda_dmabuf.c
```diff
@@ -0,0 +1,35 @@
...
+    opts.device = 0;   /* cuda device 0 */
+
+    /* Initialize CUDA side only; avoid ft_init_fabric() */
+    ret = ft_cuda_init();
+    if (ret != FI_SUCCESS) {
+        FT_ERR("ft_cuda_init failed: %d", ret);
+        return CUDA_MEMORY_SUPPORT__NOT_SUPPORTED;
+    }
+
+    cuda_memory_support_e cuda_memory_support = dmabuf_viable_and_supported();
```
j-xiong: Why not call `ft_cuda_memory_support()` directly?

fabtests/common/hmem_cuda.c
```diff
@@ -113,43 +133,134 @@ static int ft_cuda_pointer_set_attribute(void *buf)
...
+    // Final truth table
+    if (!gdr_supported && !dmabuf_supported) {
+        cuda_memory_support = CUDA_MEMORY_SUPPORT__NOT_SUPPORTED;
+    } else if (gdr_supported && dmabuf_supported) {
+        cuda_memory_support = CUDA_MEMORY_SUPPORT__DMABUF_GDR_BOTH;
+    } else if (dmabuf_supported) {
+        cuda_memory_support = CUDA_MEMORY_SUPPORT__DMA_BUF_ONLY;
+    } else {
+        cuda_memory_support = CUDA_MEMORY_SUPPORT__GDR_ONLY;
+    }
```
j-xiong: Remove all the braces here.

j-xiong (commented):
In addition to the comments below, I would suggest breaking this into two commits -- one for the common code changes and one for the efa test changes.

fabtests/pytest/efa/conftest.py
nmazzilli3: sure

nmazzilli3:
> In addition to the comments below, I would suggest breaking this into two commits -- one for the common code changes and one for the efa test changes.

Thanks for the suggestion however, this is a single atomic change, as the common code updates and the EFA test modifications are interdependent and need to be merged together to maintain consistency and functionality.

j-xiong:
> > In addition to the comments below, I would suggest breaking this into two commits -- one for the common code changes and one for the efa test changes.
> 
> Thanks for the suggestion however, this is a single atomic change, as the common code updates and the EFA test modifications are interdependent and need to be merged together to maintain consistency and functionality.

The efa test changes depend on the common code changes, but not the other way around. Any issue do you see with the common code changes come before the efa change? 

It would be a more serious issue if the common code depends on the efa changes.

nmazzilli3:
> > > In addition to the comments below, I would suggest breaking this into two commits -- one for the common code changes and one for the efa test changes.
> > 
> > 
> > Thanks for the suggestion however, this is a single atomic change, as the common code updates and the EFA test modifications are interdependent and need to be merged together to maintain consistency and functionality.
> 
> The efa test changes depend on the common code changes, but not the other way around. Any issue do you see with the common code changes come before the efa change?
> 
> It would be a more serious issue if the common code depends on the efa changes.

Thanks for the suggestion, I understand the intent behind keeping commits narrowly scoped. In this case, though, the change is small and self-contained, and splitting it across two commits would add unnecessary noise for something that’s functionally atomic.

The updates in fabtests/common, hmem_cuda.c, and the pytest/efa components are directly tied the test changes simply exercise the new logic added to the common code. Keeping them together ensures everything builds and runs correctly in a single, consistent change, and avoids additional rebase overhead.

If you strongly prefer separate commits, I can rebase and split them. However, since one commit would only introduce common code changes without any functional impact beyond enabling another file to compile, separating them would add process without improving clarity or maintainability.

j-xiong:
The fact those changes are put in the common code indicates they can be reused in the future in other places. Separating the common code does improve clarity. The commits are still in the same PR so they still get validated as a whole.  Rebasing as part of the review-fix cycle should not be a concern.

nmazzilli3:
> The fact those changes are put in the common code indicates they can be reused in the future in other places. Separating the common code does improve clarity. The commits are still in the same PR so they still get validated as a whole. Rebasing as part of the review-fix cycle should not be a concern.

Okay i have separated out the commits. Let me know if there is anything else you need here

fabtests/pytest/efa/conftest.py
```diff
@@ -99,6 +105,56 @@ def rma_fabric(cmdline_args, fabric):
...
+        print(f"CUDA validation failed: {error}")
+        pytest.skip(error)
+    
+    print(f"Correctly defined dma buf mode {do_dmabuf} and return {cuda_support}!")
+    
+    return
+
+
+@pytest.fixture(scope="function", autouse=True)
+def cuda_validation_fixture(request, cmdline_args):
```
sunkuamzn, line 147: This fixture still needs to be updated. I think the memory type fixtures should have this fixture as a dependency

https://github.com/ofiwg/libfabric/blob/502dd95d666c917dc9a031d18d248bbef034dc0f/fabtests/pytest/efa/conftest.py#L19-L25

sunkuamzn: Never mind, this logic is checking the pytest mark and we always have mark for CUDA memory tests

sunkuamzn:
@j-xiong do you have more comments?

PR #11450, opened 2025-09-30
prov/efa: Use efa_rdm_pke_release_rx_list when pke could be linked

jiaxiyan:
Replace efa_rdm_pke_release_rx() with efa_rdm_pke_release_rx_list() where packet entries could be linked in medium/runting protocol.

This ensures packets are unlinked before releasing and prevents memory leak.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -455,7 +455,7 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct 
 			", packet flags: %x\n",
 			efa_rdm_pke_get_base_hdr(pkt_entry)->type,
 			efa_rdm_pke_get_base_hdr(pkt_entry)->flags);
-		efa_rdm_pke_release_rx(pkt_entry);
+		efa_rdm_pke_release_rx_list(pkt_entry);
 		return;
 	}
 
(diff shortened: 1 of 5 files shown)


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -455,7 +455,7 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 			", packet flags: %x\n",
 			efa_rdm_pke_get_base_hdr(pkt_entry)->type,
 			efa_rdm_pke_get_base_hdr(pkt_entry)->flags);
-		efa_rdm_pke_release_rx(pkt_entry);
+		efa_rdm_pke_release_rx_list(pkt_entry);
```
shijin-aws, line 458: I am not sure we can just do this unconditionally. Looking at https://github.com/ofiwg/libfabric/pull/10784/files, this method is releasing all the pkes linked with the current pke. If you got a completion for the first pke, you cannot release the subsequent pkes before you get completions for them and process them.

jiaxiyan: I thought `efa_rdm_pke_release_rx` already assumed all subsequent pkes are null

shijin-aws: Alright, maybe I was misremembering. How do we release the rx pkts in a multi-req (medium/runting read) protocol today?

jiaxiyan: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_pke_rtm.c#L923-L926 We release it in `efa_rdm_pke_handle_data_copied` in a while loop

shijin-aws: I was reading wrong, it seems this release is in an error handling for unknown peer. It cannot have linked rx pkts at all. All good then

shijin-aws (approved):

jiaxiyan:
Not the right fix. Our reproducer shows the assertion fails in efa_rdm_ep_destroy_buffer_pools.

PR #11452, opened 2025-09-30
prov/efa: Improve emulated 1 sided protocol error case behavior

a-szegel:
There is still some work to do, but this makes our story better.  The biggest thing this PR is missing is UT's which I didn't have time to add yet.

    prov/efa: Stop 1-sided emulated protocols from abort on RX error
    
    Emulated write, read and atomic are single sided operations. They should
    never cause an abort on an RX side error because the RX side is "should
    not" be involved in the protocol that we are emulating.  We emulate due
    to the NIC not supporting the protocols, and we need to emulate the
    functionality as if the NIC did support the protocols.
    
    This change modifies the emulated write protocols, emulated read
    protocols, and emulated atomics to not abort when an error happens on
    the RX side. Error CQ's should only be written to the target CQ if write
    w/imm is used, otherwise, the operation should fail silently on the RX
    side and a NACK should be delivered to the TX side to indicate a failure
    of the operation depending on the TX's requested completion level.
    
    prov/efa: Slide recv-win on RTM/RTA error
    
    In order to continue processing future packets correctly, we need to
    progress the recv window when a RTM or RTA packet has been successfully
    delivered. If we fail processing a packet, we need to write the error to
    the CQ (when applicable), and continue to do work. Returning early here
    makes our recv window index be off by 1.
    
### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -649,9 +649,25 @@ void efa_rdm_rxe_handle_error(struct efa_rdm_ope *rxe, int err, int prov_errno)
 	//efa_rdm_rxe_release(rxe);
 
 	if (rxe->internal_flags & EFA_RDM_OPE_INTERNAL) {
-		EFA_WARN(FI_LOG_CQ,
-			"Writing eq error for rxe from internal operations\n");
-		efa_base_ep_write_eq_error(&ep->base_ep, err, prov_errno, true);
+		/*
+		 * Catch all errors from emulated one-sided operations, and ignore the
+		 * error on the RX side, which shouldn't be involved in a one sided
+		 * operation anyways.
+		 *
+		 * Emulated eager write
+		 * Emulated long-CTS write
+		 * Emulated long-read write
+		 * Emulated DC eager write
+		 * Emulated DC long-CTS write
+		 * Emulated short read
+		 * Emulated long-CTS read
+		 * Emulated atomic (for DC)
+		 * Emulated fetch atomic
+		 * Emulated compare atomic
+		 */
+		EFA_WARN(FI_LOG_CQ, "Emulated one-sided operation error on RX side\n");
+		efa_base_ep_write_eq_error(&ep->base_ep, err, prov_errno, false);
+		/* TODO Add NACK here */
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


prov/efa/src/efa_base_ep.c
```diff
@@ -586,11 +586,12 @@ bool efa_qp_support_op_in_order_aligned_128_bytes(struct efa_qp *qp, enum ibv_wr
...
  * this function will print the error message to console, then abort()
  *
- * @param[in,out] ep	base endpoint
- * @param[in] err	OFI error code
+ * @param[in,out] ep		base endpoint
+ * @param[in] err			OFI error code
  * @param[in] prov_errno	provider error code
+ * @param[in] should_abort 	If true, aborts when no EQ is available
  */
-void efa_base_ep_write_eq_error(struct efa_base_ep *ep, ssize_t err, ssize_t prov_errno)
```
shijin-aws, line 593: I do not think we need this commit at all, we shouldn't even call this function when the tx error can be ignored

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -649,9 +649,25 @@ void efa_rdm_rxe_handle_error(struct efa_rdm_ope *rxe, int err, int prov_errno)
 	//efa_rdm_rxe_release(rxe);
 
 	if (rxe->internal_flags & EFA_RDM_OPE_INTERNAL) {
-		EFA_WARN(FI_LOG_CQ,
-			"Writing eq error for rxe from internal operations\n");
-		efa_base_ep_write_eq_error(&ep->base_ep, err, prov_errno, true);
+		/*
```
shijin-aws, line 652: Like the last commit said, we shouldn't even call efa_base_ep_write_eq_error

prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -713,7 +713,13 @@ void efa_rdm_pke_handle_rx_error(struct efa_rdm_pke *pkt_entry, int prov_errno)
 			"Packet receive error from non TX/RX packet.  Our address: %s\n",
 			ep_addr_str);
 
-		efa_base_ep_write_eq_error(&ep->base_ep, err, prov_errno, true);
+		if (efa_rdm_pkt_type_of(pkt_entry) == EFA_RDM_WRITE_RTA_PKT ||
```
shijin-aws, line 716: Any rx error are unexpected, and I doubt whether you can really derive that pkt type here since the rx is failed.

a-szegel: We need the RX error to be handled for 1 sided emulated atomics.

PR #11453, opened 2025-09-30
prov/efa: remove duplicated efa_cq_construct_cq_entry

shijin-aws:
efa_cq_construct_cq_entry can be replaced by the read_entries op to avoid duplicated code.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -33,8 +33,15 @@ static inline uint64_t efa_cq_opcode_to_fi_flags(enum ibv_wc_opcode opcode) {
 	}
 }
 
-static void efa_cq_construct_cq_entry(struct efa_ibv_cq* cq,
-				      struct fi_cq_tagged_entry *entry, int opcode)
+static void efa_cq_read_context_entry(struct efa_ibv_cq *ibv_cq, void *buf, int opcode)
+{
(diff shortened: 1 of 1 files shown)


shijin-aws:
AWS CI failure is real, checking

prov/efa/src/efa_cq.c
```diff
@@ -45,12 +52,21 @@ static void efa_cq_construct_cq_entry(struct efa_ibv_cq* cq,
 		entry->op_context = NULL;
 		entry->flags = efa_cq_opcode_to_fi_flags(opcode);
 	}
-
 	entry->len = efa_ibv_cq_wc_read_byte_len(cq);
+}
+
+static void efa_cq_read_msg_entry(struct efa_ibv_cq *cq, void *buf, int opcode)
```
jiaxiyan, line 58: Is it better to remove `efa_cq_read_entry_common` and call `efa_cq_read_msg_entry` directly to reduce the call stack?

shijin-aws: efa_cq_read_entry_common is static inline

efa_cq_read_msg_entry is not a static inline, so we should make efa_cq_read_msg_entry or efa_cq_read_data_entry to call the static inline instead of calling the non-inlined function inside again otherwise it will increase the stack.

jiaxiyan (approved):

PR #11457, opened 2025-10-01
prov/efa: Grab srx lock when releasing rx entries in AV close

sunkuamzn:
The util srx code now requires the caller to grab the srx lock for freeing rx entries

The other lists in this path can also be updated in the TX, RX and CQ read paths

### DIFF
--- a/prov/efa/src/efa_av.c
+++ b/prov/efa/src/efa_av.c
@@ -16,6 +16,9 @@
 static void efa_conn_release(struct efa_av *av, struct efa_conn *conn,
 			     bool release_from_implicit_av);
 
+static void efa_conn_release_unsafe(struct efa_av *av, struct efa_conn *conn,
+			     bool release_from_implicit_av);
+
 /*
  * Local/remote peer detection by comparing peer GID with stored local GIDs
  */
@@ -390,15 +393,18 @@ int efa_conn_rdm_insert_shm_av(struct efa_av *av, struct efa_conn *conn)
 }
 
 /**
- * @brief release the rdm related resources of an efa_conn object
+ * @brief release the rdm related resources of an efa_conn object. This function
+ * requires the caller to take the SRX lock because this function modifies the
+ * peer map and destroys peers which are accessed and modified in the CQ read
+ * path.
  *
  * this function release the shm av entry and rdm peer;
  *
  * @param[in]	av	address vector
  * @param[in]	conn	efa_conn object
+ * peer
  */
-static
-void efa_conn_rdm_deinit(struct efa_av *av, struct efa_conn *conn)
+static void efa_conn_rdm_deinit(struct efa_av *av, struct efa_conn *conn)
 {
 	int err;
 	struct efa_rdm_peer *peer;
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_peer.c
```diff
@@ -98,10 +98,14 @@ void efa_rdm_peer_destruct(struct efa_rdm_peer *peer, struct efa_rdm_ep *ep)
 		efa_rdm_txe_release(txe);
 	}
 
-	dlist_foreach_container_safe(&peer->rxe_list,
-				     struct efa_rdm_ope,
-				     rxe, peer_entry, tmp) {
-		efa_rdm_rxe_release(rxe);
+	if (!dlist_empty(&peer->rxe_list)) {
```
shijin-aws: should we just lock the whole peer destruct?

sunkuamzn: Updated

shijin-aws:
Is it possible that we can destroy a peer during a cq read (which grab srx lock) ? I guess with the recent implicit AV eviction logic, we can run into this situation? Then probably we should refactor our locking ?

prov/efa/src/efa_av.c
```diff
@@ -437,7 +439,13 @@ void efa_conn_rdm_deinit(struct efa_av *av, struct efa_conn *conn)
 
 		peer = efa_rdm_ep_peer_map_lookup(peer_map, fi_addr);
```
shijin-aws, line 448: I think we need to lock this entire `dlist_foreach_safe(&av->util_av.ep_list, entry, tmp)`  when as the ep close is now hash deleting this map inside an srx lock

prov/efa/src/efa_av.c
```diff
@@ -739,9 +748,10 @@ struct efa_conn *efa_conn_alloc(struct efa_av *av, struct efa_ep_addr *raw_addr,
  * @param[in]	av	address vector
  * @param[in]	conn	efa_conn object pointer
  * @param[in]	release_from_implicit_av		whether to release conn from implicit AV
+ * @param[in]	grab_srx_lock		whether to get the SRX lock before destroying the peer struct
  */
 static
-void efa_conn_release(struct efa_av *av, struct efa_conn *conn, bool release_from_implicit_av)
+void efa_conn_release(struct efa_av *av, struct efa_conn *conn, bool release_from_implicit_av, bool grab_srx_lock)
```
shijin-aws: I would define an `efa_conn_release_unsafe` that has no lock, and define a `efa_conn_release` which acquires the lock and call `efa_conn_release_unsafe` so we do not need to have such extra bool

shijin-aws: And the implicit av insert can call `efa_conn_release_unsafe`

sunkuamzn: I'm trying to minimize the amount of time the srx lock is taken. If I take it for all of `conn_release`, it will be held for reverse AV and util AV updates

prov/efa/src/efa_av.c
sunkuamzn: updated

prov/efa/src/rdm/efa_rdm_peer.c
```diff
@@ -39,6 +39,30 @@ void efa_rdm_peer_construct(struct efa_rdm_peer *peer, struct efa_rdm_ep *ep, st
...
+ * efa_rdm_peer_destruct_srx_lock which grabs the SRX lock before destroying the
+ * peer and the CQ read path directly calls efa_rdm_peer_destruct
+ *
+ * release reorder buffer, txe list and rxe list of a peer after taking the SRX
+ * lock
+ *
+ * @param[in,out]	peer 	rdm peer
+ * @relates efa_rdm_peer
+ */
+void efa_rdm_peer_destruct_srx_lock(struct efa_rdm_peer *peer, struct efa_rdm_ep *ep)
```
shijin-aws: Do u still use this function ?

PR #11459, opened 2025-10-02
prov/efa: Take domain lock to protect concurrent access to domain fields

jiaxiyan:
Fix Data race undermines locking  (LOCK_EVASION)
The thread 1 sets "domain" to a new value. Now the two threads have an inconsistent view of "domain" and updates to fields of "domain" or fields correlated with "domain" may be lost.


shijin-aws:
Fixing CID 490430 https://scan4.scan.coverity.com/#/project-view/62099/10344?selectedIssue=490430

PR #11465, opened 2025-10-03
prov/efa: dmabuf try / fallback logic

nmazzilli3:
feat: Adding default dmabuf attempt and fallback logic for all efa_hmem_ifaces

Problem:
  - How make dmabuf usage default going forward and have fallback mechanism if dmabuf not supported

Solution:
  - Modified initial PR from @jiaxiyan at https://github.com/ofiwg/libfabric/commit/6aa6708f99234ddef233182d656ec25eb1c5159b#diff-9b57a9410ed94ed1f1aea837412e68bbe9b49582edce813ba352fffb37dcc007
  - Added dmabuf_supported_by_device_b flag in efa_hmem_info structure in prov/efa/efa_hmem.h
  - Updated dmabuf_supported_by_device_b in each fi_hmem_iface type p2p_support fcn in prov/efa/efa_hmem.c
  - Removed per fi_hmem_iface type checks in prov/efa/efa_mr_reg_ibv_mr.c -

Testing:
  - Ran mpi perf tests on 2 nodes on p5en with dmabuf and fallback option hard set
  - Ran mpi perf tests on 16 nodes on p5en with dmabuf and fallback option hard set

Sim Issue:
  - N/A

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -118,6 +118,8 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
 	int dmabuf_fd;
 	uint64_t dmabuf_offset;
 
+	info->dmabuf_supported_by_device_b = false;
+
 	cuda_ret = ofi_cudaMalloc(&ptr, len);
 	if (cuda_ret != cudaSuccess) {
 		info->initialized = false;
@@ -145,6 +147,7 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
 				"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
 			ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
 		}
+		info->dmabuf_supported_by_device_b = true;
 	} else {
 		EFA_INFO(FI_LOG_CORE,
 			"Unable to retrieve dmabuf fd of CUDA device buffer: %d. "
@@ -193,6 +196,8 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
 	uint64_t offset;
 	int ret;
 
+	info->dmabuf_supported_by_device_b = false;
+
 	if (g_efa_selected_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ) {
 		ibv_access |= IBV_ACCESS_REMOTE_READ;
 	}
@@ -222,6 +227,7 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
 		ibv_mr = ibv_reg_dmabuf_mr(
 					ibv_pd, offset,
 					len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+		info->dmabuf_supported_by_device_b = true;
 	} else if (ret == -FI_EOPNOTSUPP) {
 		EFA_INFO(FI_LOG_MR,
 			"Unable to retrieve dmabuf fd of Neuron device buffer, "
@@ -284,9 +290,14 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
 	}
 
 	info->initialized = true;
+	info->max_medium_msg_size = 0;
+	info->runt_size = 0;
+	info->min_read_msg_size = 0;
+	info->min_read_write_size = 0;
 
 	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
 		info->p2p_supported_by_device = true;
+		info->dmabuf_supported_by_device_b = true;
 	} else if (ofi_hmem_p2p_disabled()) {
 		info->p2p_supported_by_device = false;
 	} else {
--- a/prov/efa/src/efa_hmem.h
+++ b/prov/efa/src/efa_hmem.h
@@ -24,6 +24,7 @@ static const enum fi_hmem_iface efa_hmem_ifaces[] = {
 struct efa_hmem_info {
 	bool initialized; 	/* do we support it at all */
 	bool p2p_supported_by_device;	/* do we support p2p with this device */
+	bool dmabuf_supported_by_device_b;	/* do we support dmabuf with this device */
 
 	size_t max_medium_msg_size;
 	size_t runt_size;
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -506,77 +506,82 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
  * @param flags flags in fi_mr_reg/fi_mr_regattr
  * @return struct ibv_mr* the ptr to the registered MR
  */
-static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr *mr_attr,
-					int access, const uint64_t flags)
+static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr,
+                                        struct fi_mr_attr *mr_attr,
+                                        int access, const uint64_t flags)
 {
-	int dmabuf_fd;
-	uint64_t offset;
-	int ret;
-	struct ibv_mr *ibv_mr;
+    int dmabuf_fd;
+    uint64_t offset;
+    int ret;
 
-	if (flags & FI_MR_DMABUF)
-		return efa_mr_reg_ibv_dmabuf_mr(
-			efa_mr->domain->ibv_pd,
-			mr_attr->dmabuf->offset,
-			mr_attr->dmabuf->len,
-			(uintptr_t) mr_attr->dmabuf->base_addr + mr_attr->dmabuf->offset,
-			mr_attr->dmabuf->fd,
-			access
-		);
+    /* Explicit dmabuf registration */
+    if (flags & FI_MR_DMABUF) {
+        if (!mr_attr->dmabuf) {
+            EFA_WARN(FI_LOG_MR, "FI_MR_DMABUF set but mr_attr->dmabuf == NULL\n");
+            return NULL;
+        }
+        if (!g_efa_hmem_info[mr_attr->iface].dmabuf_supported_by_device_b) {
+            EFA_WARN(FI_LOG_MR,
+                     "Requested FI_MR_DMABUF, but dmabuf not supported for iface=%d\n",
+                     mr_attr->iface);
+            return NULL;
+        }
 
-	if (efa_mr_is_synapseai(efa_mr)) {
-		ret = ofi_hmem_get_dmabuf_fd(efa_mr->peer.iface,
-					     mr_attr->mr_iov->iov_base,
-					     (uint64_t) mr_attr->mr_iov->iov_len,
-					     &dmabuf_fd, &offset);
-		if (ret != FI_SUCCESS) {
-			EFA_WARN(FI_LOG_MR, "Unable to get dmabuf fd for Gaudi device buffer \n");
-			return NULL;
-		}
-		return efa_mr_reg_ibv_dmabuf_mr(
-				efa_mr->domain->ibv_pd, offset,
-				mr_attr->mr_iov->iov_len,
-				(uint64_t)mr_attr->mr_iov->iov_base,
-				dmabuf_fd, access);
-	}
+        EFA_INFO(FI_LOG_MR,
+                 "FI_MR_DMABUF: fd=%d offset=%lu len=%zu\n",
+                 mr_attr->dmabuf->fd, mr_attr->dmabuf->offset,
+                 mr_attr->dmabuf->len);
 
-	/*
-	 * TODO: need such fallback for cuda as well when
-	 * FI_CUDA_API_PERMITTED is true
-	 */
-	if (efa_mr_is_neuron(efa_mr)) {
-		ret = ofi_hmem_get_dmabuf_fd(
-				efa_mr->peer.iface,
-				mr_attr->mr_iov->iov_base,
-				mr_attr->mr_iov->iov_len,
-				&dmabuf_fd,
-				&offset);
+        return efa_mr_reg_ibv_dmabuf_mr(
+            efa_mr->domain->ibv_pd,
+            mr_attr->dmabuf->offset,
+            mr_attr->dmabuf->len,
+            (uintptr_t) mr_attr->dmabuf->base_addr + mr_attr->dmabuf->offset,
+            mr_attr->dmabuf->fd,
+            access);
+    }
 
-		if (ret == FI_SUCCESS) {
-			/* Success => invoke ibv_reg_dmabuf_mr */
-			ibv_mr = efa_mr_reg_ibv_dmabuf_mr(
-					efa_mr->domain->ibv_pd, 0,
-					mr_attr->mr_iov->iov_len,
-					(uint64_t)mr_attr->mr_iov->iov_base,
-					dmabuf_fd, access);
-			(void) ofi_hmem_put_dmabuf_fd(efa_mr->peer.iface, dmabuf_fd);
-			return ibv_mr;
-		} else if (ret == -FI_EOPNOTSUPP) {
-			/* Protocol not available => fallback */
-			EFA_INFO(FI_LOG_MR,
-				"Unable to get dmabuf fd for Neuron device buffer, "
-				"Fall back to ibv_reg_mr\n");
-			return ibv_reg_mr(
-				efa_mr->domain->ibv_pd,
-				(void *)mr_attr->mr_iov->iov_base,
-				mr_attr->mr_iov->iov_len, access);
-		}
-		return NULL;
-	}
+    /* Implicit VA path with dmabuf-first */
+    if (g_efa_hmem_info[mr_attr->iface].dmabuf_supported_by_device_b) {
+        ret = ofi_hmem_get_dmabuf_fd(
+                mr_attr->iface,
+                mr_attr->mr_iov->iov_base,
+                (uint64_t) mr_attr->mr_iov->iov_len,
+                &dmabuf_fd, &offset);
 
-	return ibv_reg_mr(efa_mr->domain->ibv_pd,
-			(void *)mr_attr->mr_iov->iov_base,
-			mr_attr->mr_iov->iov_len, access);
+        if (ret == FI_SUCCESS) {
+            EFA_INFO(FI_LOG_MR,
+                     "Registering dmabuf MR: fd=%d offset=%lu len=%zu\n",
+                     dmabuf_fd, offset, mr_attr->mr_iov->iov_len);
+
+            return efa_mr_reg_ibv_dmabuf_mr(
+                efa_mr->domain->ibv_pd, offset,
+                mr_attr->mr_iov->iov_len,
+                (uint64_t)mr_attr->mr_iov->iov_base,
+                dmabuf_fd, access);
+        }
+
+        if (ret == -FI_EOPNOTSUPP || ret == -FI_ENOSYS) {
+            EFA_WARN(FI_LOG_MR,
+                     "dmabuf not supported at runtime for iface=%d, disabling\n",
+                     mr_attr->iface);
+			g_efa_hmem_info[mr_attr->iface].dmabuf_supported_by_device_b = false;
+        } else {
+            EFA_WARN(FI_LOG_MR,
+                     "ofi_hmem_get_dmabuf_fd failed: ret=%d (%s)\n",
+                     ret, fi_strerror(-ret));
+        }
+        /* fall through to ibv_reg_mr */
+    }
+
+    /* Fallback: plain ibv_reg_mr */
+    EFA_WARN(FI_LOG_MR,
+             "Fallback ibv_reg_mr: addr=%p len=%zu\n",
+             mr_attr->mr_iov->iov_base, mr_attr->mr_iov->iov_len);
+
+    return ibv_reg_mr(efa_mr->domain->ibv_pd,
+                      (void *)mr_attr->mr_iov->iov_base,
+                      mr_attr->mr_iov->iov_len, access);
 }
 
 #if HAVE_CUDA


prov/efa/src/efa_hmem.c
```diff
@@ -284,9 +290,14 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
...
 
 	info->initialized = true;
+	info->max_medium_msg_size = 0;
+	info->runt_size = 0;
+	info->min_read_msg_size = 0;
+	info->min_read_write_size = 0;
 
 	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
 		info->p2p_supported_by_device = true;
+		info->dmabuf_supported_by_device_b = true;
```
shijin-aws: Is dmabuf working for system memory?

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,82 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+            mr_attr->dmabuf->offset,
+            mr_attr->dmabuf->len,
+            (uintptr_t) mr_attr->dmabuf->base_addr + mr_attr->dmabuf->offset,
+            mr_attr->dmabuf->fd,
+            access);
+    }
+
+    /* Implicit VA path with dmabuf-first */
+    if (g_efa_hmem_info[mr_attr->iface].dmabuf_supported_by_device_b) {
+        ret = ofi_hmem_get_dmabuf_fd(
```
shijin-aws, line 546: if it is system memory, does it work?

shijin-aws: system memory is ofi_hmem_no_get_dmabuf_fd will fail always, so we shouldn't need to detect it.

prov/efa/src/efa_hmem.c
nmazzilli3: This is most likely a misunderstanding on my end. It looked like dmabuf was attempted in every path for SYNAPSEAI. What should this be instead

prov/efa/src/efa_mr.c
nmazzilli3: Would it matter since we fallback anyways? What instance types / os combos could i run this against to get this data?

prov/efa/src/efa_hmem.c
shijin-aws: It was attempted in every path for synapseai. Though this is not idea, but when the code was introduced, there was no get_dmabuf_fd API for synapseai, and I didn't test whether that get_dmabuf_fd API work on Gaudi 1 (the synapseai GPU on AWS platform today). So I would keep the code as is today.

For system memory I think we should just make it as false as it doesn't make much sense to try and fallback additionally

prov/efa/src/efa_mr.c
shijin-aws: We shouldn't fallback, we should just do what it is expected to do. Btw, I have seen tons of such warning in the log which is not what we expect?

```
libfabric:162433:1759529671::efa:mr:efa_mr_reg_ibv_mr():578<warn> Fallback ibv_reg_mr: addr=0x6136f29d1170 len=1024
libfabric:162433:1759529671::efa:mr:efa_mr_reg_ibv_mr():578<warn> Fallback ibv_reg_mr: addr=0x6136f29d1820 len=1024
libfabric:162433:1759529671::efa:mr:efa_mr_reg_ibv_mr():578<warn> Fallback ibv_reg_mr: addr=0x6136f29d1ed0 len=1024
libfabric:162433:1759529671::efa:mr:efa_mr_reg_ibv_mr():578<warn> Fallback ibv_reg_mr: addr=0x6136f29d2580 len=1024
libfabric:162433:1759529671::efa:mr:efa_mr_reg_ibv_mr():578<warn> Fallback ibv_reg_mr: addr=0x6136f29d2c30 len=1024

```

prov/efa/src/efa_hmem.c
```diff
@@ -145,6 +147,7 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
 				"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
 			ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
 		}
+		info->dmabuf_supported_by_device_b = true;
```
jiaxiyan: This should be in an `else` statement.

prov/efa/src/efa_hmem.c
```diff
@@ -222,6 +227,7 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
 		ibv_mr = ibv_reg_dmabuf_mr(
 					ibv_pd, offset,
 					len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+		info->dmabuf_supported_by_device_b = true;
```
jiaxiyan: This needs to be set as false `if (!ibv_mr)`.

prov/efa/src/efa_hmem.h
```diff
@@ -24,6 +24,7 @@ static const enum fi_hmem_iface efa_hmem_ifaces[] = {
 struct efa_hmem_info {
 	bool initialized; 	/* do we support it at all */
 	bool p2p_supported_by_device;	/* do we support p2p with this device */
+	bool dmabuf_supported_by_device_b;	/* do we support dmabuf with this device */
```
jiaxiyan, line 27: what does `_b` stand for?

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,82 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+        if (ret == FI_SUCCESS) {
+            EFA_INFO(FI_LOG_MR,
+                     "Registering dmabuf MR: fd=%d offset=%lu len=%zu\n",
+                     dmabuf_fd, offset, mr_attr->mr_iov->iov_len);
+
+            return efa_mr_reg_ibv_dmabuf_mr(
+                efa_mr->domain->ibv_pd, offset,
+                mr_attr->mr_iov->iov_len,
+                (uint64_t)mr_attr->mr_iov->iov_base,
+                dmabuf_fd, access);
```
jiaxiyan, line 562: `ofi_hmem_put_dmabuf_fd` is needed here to close the fd.

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,82 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+        } else {
+            EFA_WARN(FI_LOG_MR,
+                     "ofi_hmem_get_dmabuf_fd failed: ret=%d (%s)\n",
+                     ret, fi_strerror(-ret));
+        }
+        /* fall through to ibv_reg_mr */
+    }
+
+    /* Fallback: plain ibv_reg_mr */
+    EFA_WARN(FI_LOG_MR,
```
jiaxiyan: This should be moved to the previous block.

prov/efa/src/efa_hmem.h
nmazzilli3: help user know its a boolean value when interfacing with the structure or reading its usage in code.

prov/efa/src/efa_hmem.c
nmazzilli3: why is that?

jiaxiyan: `dmabuf_supported_by_device_b` is true when `ibv_reg_dmabuf_mr` succeeded

prov/efa/src/efa_hmem.h
jiaxiyan: Prefer to follow the current naming convention

nmazzilli3: where is the naming convention defined?

jiaxiyan: https://github.com/ofiwg/libfabric/blob/main/prov/efa/docs/naming_convention.md
Not mentioned here but we never did `_b`

nmazzilli3: There are no rules for variable names outside a global declaration.

prov/efa/src/efa_hmem.c
nmazzilli3: Ok, I set FI_HMEM_SYSTEM to be true for p2p and false for dma. Let me know if i interpreted that incorrectly

prov/efa/src/efa_mr.c
nmazzilli3: Removed fallback logic

prov/efa/src/efa_env.c
```diff
@@ -184,6 +199,15 @@ void efa_env_define()
 			"Set the maximum number of receive operations before the provider returns -FI_EAGAIN.");
 	fi_param_define(&efa_prov, "rx_copy_unexp", FI_PARAM_BOOL,
 			"Enables the use of a separate pool of bounce-buffers to copy unexpected messages out of the pre-posted receive buffers. (Default: 1)");
+
+	/* HMEM DMABUF environment variables */
+	fi_param_define(&efa_prov, "hmem_system_use_dmabuf", FI_PARAM_BOOL,
```
shijin-aws: I would define them in the common hmem_* code

prov/efa/src/efa_hmem.c
```diff
@@ -7,6 +7,32 @@
...
+ * EFA-specific and core libfabric environment variables.
+ *
+ * @param[in] iface The HMEM interface to check
+ * @return true if DMABUF is enabled for the interface, false otherwise
+ */
+bool efa_hmem_is_dmabuf_env_var_enabled(enum fi_hmem_iface iface)
+{
+	switch (iface) {
+	case FI_HMEM_SYSTEM:
+		return efa_env.hmem_system_use_dmabuf;
```
shijin-aws: I am still not sure it is valuable to have such env for system. It can be good for the generalization of the interface, but you can return false for system here anyway

prov/efa/src/efa_hmem.c
```diff
@@ -134,21 +162,28 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
...
 
 #if HAVE_EFA_DMABUF_MR
-	ret = cuda_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
-	if (ret == FI_SUCCESS) {
-		ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
-					   len, (uint64_t)ptr, dmabuf_fd, ibv_access);
-		(void)cuda_put_dmabuf_fd(dmabuf_fd);
-		if (!ibv_mr) {
+	if (efa_hmem_is_dmabuf_env_var_enabled(FI_HMEM_CUDA)) {
+		ret = cuda_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
```
shijin-aws: Use the `ofi_hmem_get/put_dmabuf_fd` with `FI_HMEM_CUDA` here to reduce the cuda symbol exposure

prov/efa/src/efa_hmem.c
```diff
@@ -217,15 +254,28 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
...
-	if (ret == FI_SUCCESS) {
-		ibv_mr = ibv_reg_dmabuf_mr(
-					ibv_pd, offset,
-					len, (uint64_t)ptr, dmabuf_fd, ibv_access);
-	} else if (ret == -FI_EOPNOTSUPP) {
-		EFA_INFO(FI_LOG_MR,
-			"Unable to retrieve dmabuf fd of Neuron device buffer, "
-			"Fall back to ibv_reg_mr\n");
+	if (efa_hmem_is_dmabuf_env_var_enabled(FI_HMEM_NEURON)) {
+		ret = neuron_get_dmabuf_fd(ptr, (uint64_t)len, &dmabuf_fd, &offset);
```
shijin-aws: Similar comment, use the ofi_hmem wrappers for dmabuf ops

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,88 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+            (void) ofi_hmem_put_dmabuf_fd(mr_attr->iface, dmabuf_fd);
+            
+            if (dmabuf_mr) {
+                return dmabuf_mr;
+            }
+            
+            EFA_INFO(FI_LOG_MR, "DMABUF registration failed, returning NULL\n");
+        }
+
+        if (ret == -FI_EOPNOTSUPP || ret == -FI_ENOSYS) {
```
shijin-aws: This makes the logic a bit too complicated, for these two returns, the warning should already happens in their related function calls (if there are not, we should add them). No need to handle these again in the caller side

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,88 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+                dmabuf_fd, access);
+            
+            /* Close the dmabuf file descriptor - it's no longer needed after registration */
+            (void) ofi_hmem_put_dmabuf_fd(mr_attr->iface, dmabuf_fd);
+            
+            if (dmabuf_mr) {
+                return dmabuf_mr;
+            }
+            
+            EFA_INFO(FI_LOG_MR, "DMABUF registration failed, returning NULL\n");
```
shijin-aws: I would just return dmabuf_mr here, the upper layer will check the returned mr is NULL or not

prov/efa/src/efa_hmem.c
nmazzilli3: default return false

prov/efa/src/efa_env.c
nmazzilli3: Define them in hmem.c 

like https://github.com/nmazzilli3/libfabric/blob/main/src/hmem.c#L678

prov/efa/src/efa_mr.c
nmazzilli3: remove if (dmabuf_mr) check

prov/efa/src/efa_mr.c
nmazzilli3: will remove check for single message

src/hmem.c
```diff
@@ -679,10 +687,36 @@ void ofi_hmem_init(void)
 			"Disable peer to peer support between device memory and"
 			" network devices. (default: no).");
 
+	/* HMEM DMABUF environment variables */
+	fi_param_define(NULL, "hmem_system_use_dmabuf", FI_PARAM_BOOL,
+			"Use DMABUF for system memory registration. (Default: 0)");
+	fi_param_define(NULL, "hmem_cuda_use_dmabuf", FI_PARAM_BOOL,
```
jiaxiyan: `hmem_cuda_use_dmabuf` and `hmem_rocr_use_dmabuf` are already defined in `src/hmem_cuda.c` and `hmem_rocr.c`. Follow the same pattern as https://github.com/ofiwg/libfabric/pull/10502/files and add a function instead of using extern bool.
Always separate the common code change and efa change into different commits for the community to review.

nmazzilli3: That makes sense, I implemented what @shijin-aws said in a comment above which basically told me to do this. Could you read the comments above and then we can decide on best path forward.

jiaxiyan: He means moving to hmem common code, not specifically `hmem.c`

shijin-aws: Yes, that is what I meant

nmazzilli3: ack, synced with @jiaxiyan and will remove these from hmem.c and reference all env variables from hmem_*interface* files and create associated header files.

src/hmem_cuda.c
```diff
@@ -959,13 +959,16 @@ bool cuda_is_gdrcopy_enabled(void)
 	return cuda_attr.use_gdrcopy;
 }
 
-bool cuda_is_dmabuf_supported(void)
+bool cuda_is_dmabuf_requested(void)
 {
 	int use_dmabuf = 1;
-
 	fi_param_get_bool(NULL, "hmem_cuda_use_dmabuf", &use_dmabuf);
+	return use_dmabuf != 0;
```
jiaxiyan, line 966: Just `return use_dmabuf`. Same for the other places.

prov/efa/src/efa_hmem.c
```diff
@@ -284,9 +337,30 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
...
+		if (efa_hmem_is_dmabuf_env_var_enabled(FI_HMEM_SYNAPSEAI)) {
+			info->dmabuf_supported_by_device_b = true;
+		} else {
+			EFA_INFO(FI_LOG_CORE, "FI_HMEM_SYNAPSEAI DMABUF disabled by environment variable\n");
+			info->dmabuf_supported_by_device_b = false;
+		}
+	} else if(iface == FI_HMEM_SYSTEM) {
 		info->p2p_supported_by_device = true;
+		info->dmabuf_supported_by_device_b = false;
+	} else if (iface == FI_HMEM_ROCR) {
```
jiaxiyan: ROCR is not supported yet and it shouldn't be added here without verification. see https://github.com/ofiwg/libfabric/pull/11573/files#diff-087fdb0c5afcad21fd1eb64dc049a1f293b620b7259ae928f4674335027fe52a

prov/efa/src/efa_hmem.c
```diff
@@ -193,6 +230,8 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
 	uint64_t offset;
 	int ret;
 
+	info->dmabuf_supported_by_device_b = false;
```
jiaxiyan: move this to efa_hmem_info_init_iface

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,77 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+    uint64_t offset;
+    int ret;
+
+    /* Explicit dmabuf registration */
+    if (flags & FI_MR_DMABUF) {
+        if (!mr_attr->dmabuf) {
+            EFA_WARN(FI_LOG_MR, "FI_MR_DMABUF set but mr_attr->dmabuf == NULL\n");
+            return NULL;
+        }
+        if (!g_efa_hmem_info[mr_attr->iface].dmabuf_supported_by_device_b) {
```
jiaxiyan: `mr_attr->iface` is not always the same as `efa_mr->peer.iface`

src/hmem.c
jiaxiyan: bump my comment to 
> Always separate the common code change and efa change into different commits for the community to review.

src/hmem_cuda.c
nmazzilli3: The current code with return use_dmabuf != 0; is actually doing the right thing by explicitly converting the int to a boolean value and makes the boolean intent clear.

prov/efa/src/efa_hmem.c
nmazzilli3: would defaulting to false be reasonable for now 

```
case FI_HMEM_ROCR:
		return false;
```

prov/efa/src/efa_hmem.c
nmazzilli3: im not sure i follow this is just setting it to false so it doesn't have to deal with all the nested if statements if getting the dma buff fails?

prov/efa/src/efa_mr.c
nmazzilli3: What do you recommend? I'm also not seeing peer in that struct 

```
struct fi_mr_attr {
	union {
		const struct iovec *mr_iov;
		const struct fi_mr_dmabuf *dmabuf;
	};
	size_t			iov_count;
	uint64_t		access;
	uint64_t		offset;
	uint64_t		requested_key;
	void			*context;
	size_t			auth_key_size;
	uint8_t			*auth_key;
	enum fi_hmem_iface	iface;
	union {
		uint64_t	reserved;
		int		cuda;
		int		ze;
		int		neuron;
		int		synapseai;
	} device;
	void			*hmem_data;
	size_t			page_size;
	const struct fid_mr	*base_mr;
	size_t			sub_mr_cnt;
};

struct fi_mr_modify {
	uint64_t		flags;
	struct fi_mr_attr	attr;
};
```

prov/efa/src/efa_hmem.c
jiaxiyan: I wouldn't add ROCR in this PR.

prov/efa/src/efa_hmem.c
jiaxiyan: move it to line 344 so you don't need to initialize it in both `efa_hmem_info_check_p2p_support_cuda` and `efa_hmem_info_check_p2p_support_neuron`

prov/efa/src/efa_hmem.c
```diff
@@ -217,15 +256,29 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
...
-			"Unable to retrieve dmabuf fd of Neuron device buffer, "
-			"Fall back to ibv_reg_mr\n");
+	if (efa_hmem_is_dmabuf_env_var_enabled(FI_HMEM_NEURON)) {
+		ret = ofi_hmem_get_dmabuf_fd(FI_HMEM_NEURON, ptr, (uint64_t)len, &dmabuf_fd, &offset);
+		if (ret == FI_SUCCESS) {
+			ibv_mr = ibv_reg_dmabuf_mr(
+						ibv_pd, offset,
+						len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+			(void)ofi_hmem_put_dmabuf_fd(FI_HMEM_NEURON, dmabuf_fd);
+			if (ibv_mr) {
```
jiaxiyan: make this consistent with `efa_hmem_info_check_p2p_support_cuda` and do `if (!ibv_mr) {` first

prov/efa/src/efa_mr.c
jiaxiyan: It is part of `struct efa_mr`. I suggest keeping it as `efa_mr->peer.iface` both here and in line 545 to 565.

prov/efa/src/efa_hmem.c
nmazzilli3: sure

prov/efa/src/efa_mr.c
nmazzilli3: good call out, thank you.

prov/efa/src/efa_hmem.c
nmazzilli3: sure will remove it.

prov/efa/src/efa_hmem.c
nmazzilli3: ok

prov/efa/src/efa_mr.c
nmazzilli3: Made change in a few more places please resolve if you are satisfied with the changes

src/hmem.c
nmazzilli3: done

prov/efa/src/efa_mr.c
```diff
@@ -506,77 +506,77 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+                &dmabuf_fd, &offset);
+
+        if (ret == FI_SUCCESS) {
+            struct ibv_mr *dmabuf_mr;
+            EFA_INFO(FI_LOG_MR,
+                     "Registering dmabuf MR: fd=%d offset=%lu len=%zu\n",
+                     dmabuf_fd, offset, mr_attr->mr_iov->iov_len);
+
+            dmabuf_mr = efa_mr_reg_ibv_dmabuf_mr(
+                efa_mr->domain->ibv_pd, 0,
```
nmazzilli3, line 559: NOTE: I did remove the offset here to match what we previously had.

prov/efa/src/efa_hmem.c
```diff
@@ -284,9 +333,23 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
 	}
 
 	info->initialized = true;
+	info->max_medium_msg_size = 0;
+	info->runt_size = 0;
+	info->min_read_msg_size = 0;
+	info->min_read_write_size = 0;
```
jiaxiyan, line 339: These are default to 0 and set in efa_domain_hmem_info_init_protocol_thresholds.

prov/efa/src/efa_hmem.c
```diff
@@ -284,9 +333,23 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
...
+	info->dmabuf_supported_by_device_b = false;
 
-	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
+	if (iface == FI_HMEM_SYNAPSEAI) {
+		info->p2p_supported_by_device = true;
+		if (efa_hmem_is_dmabuf_env_var_enabled(FI_HMEM_SYNAPSEAI)) {
+			info->dmabuf_supported_by_device_b = true;
+		} else {
+			EFA_INFO(FI_LOG_CORE, "FI_HMEM_SYNAPSEAI DMABUF disabled by environment variable\n");
+			info->dmabuf_supported_by_device_b = false;
```
jiaxiyan, line 348: You don't need to set it again here

jiaxiyan:
@j-xiong Can you review 80a80f357c04f2aed99e7982a0472b4c20daf13d?

src/hmem_cuda.c
j-xiong: I agree with @jiaxiyan here even though the explicit type conversion might be more syntactically correct.  Implicit conversion from int to bool is quite common and the naming of the variable doesn't leave room for confusion.

prov/efa/src/efa_hmem.c
```diff
@@ -7,6 +7,34 @@
...
+ * @brief Check if DMABUF is enabled for a specific HMEM interface
+ *
+ * This function checks the environment variables to determine if DMABUF
+ * should be used for the specified HMEM interface. It respects both
+ * EFA-specific and core libfabric environment variables.
+ *
+ * @param[in] iface The HMEM interface to check
+ * @return true if DMABUF is enabled for the interface, false otherwise
+ */
+bool efa_hmem_is_dmabuf_env_var_enabled(enum fi_hmem_iface iface)
```
j-xiong, line 20: Maybe we should add a method to the common hmem interface and define an `ofi_hmem_is_dmabuf_requested` function in `src/hmem.c`.  That make the code reusable for other providers. Also need to add the function for `FI_HMEM_ZE` is going this route.

prov/efa/src/efa_hmem.c
charlesstoll: Fixed. New PR at https://github.com/ofiwg/libfabric/pull/11804

prov/efa/src/efa_hmem.c
charlesstoll: Fixed. New PR at https://github.com/ofiwg/libfabric/pull/11804

prov/efa/src/efa_hmem.c
charlesstoll: Fixed. New PR at https://github.com/ofiwg/libfabric/pull/11804

I also modified the first patch to add dmabuf support detection to ZE, please take a look.

charlesstoll:
Ownership of this feature has change to me. New PR at https://github.com/ofiwg/libfabric/pull/11804

src/hmem_cuda.c
charlesstoll: Fixed. New PR at https://github.com/ofiwg/libfabric/pull/11804

PR #11467, opened 2025-10-03
[v2.2.x] prov/efa: Fix the memleak for efa_rdm_peer

shijin-aws:
backport https://github.com/ofiwg/libfabric/pull/11462


shijin-aws:
The patch doesn't compile, will reopen when needed

PR #11468, opened 2025-10-04
prov/efa: allocate peer map entry pool during the rdm ep create

snadampal:
the commit "4016fd5" has moved the peer map entry pool from av to endpoint but deferred the pool memory allocation till the remote requests for a peer. So, the acutal memory pool allocation was happening during the connection establishment request and hence adding latency overhead to the communication.

This commit is fixing the latency issue by eagerly allocating the pool during the endpoint create.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_fiops.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_fiops.c
@@ -242,6 +242,10 @@ int efa_rdm_ep_create_buffer_pools(struct efa_rdm_ep *ep)
 	if (ret)
 		goto err_free;
 
+	ret = ofi_bufpool_grow(ep->peer_map_entry_pool);
+	if (ret)
+		goto err_free;
+
 	return 0;
 
 err_free:


snadampal:
the CI failure doesn't seem to be due to the PR change, @shijin-aws , can you please check and comment.

shijin-aws:
bot:aws:retest

shijin-aws:
Something is wrong with nccl installation, need to check further

a-szegel:
bot:aws:retest

snadampal:
Do you know why all those pools are lazy initialized?
It makes sense to delay the pools that get used only during overflows, unexpected or error message handling but not for others.

shijin-aws:
No I don't know. But I can see it is not good to lazy initialize some pools that must be allocated, otherwise the first transmission will be bad delayed

shijin-aws (approved):

sunkuamzn:
peer_map_entry_pool is also used during CQ read, so the delay could even be in the RX path

shijin-aws:
The peer map entry pool is special because the earlier commit 4016fd5 changed the behavior

so the peer entry pool was in AV, and it will be growed in the first av insertion aka the slow path

commit 4016fd5 makes it to the fast path call

so this change is moving it back to slow path, which I think it's a reasonable revert IMO.

Merging

PR #11469, opened 2025-10-05
prov/ucx: Fix ops structs initialization

tatarintsevsv:
Use of non initialized operations lead to a segmentation fault


j-xiong:
The CI failure is with udp and unrelated.

j-xiong (approved):

PR #11473, opened 2025-10-06
Fix peer robuf overhead issue

snadampal:
This PR adds (1) support for fi circular buffer and recwindows from a pre-allocated pool
(2) use that feature to allocate peer reorder buffers to avoid latency in the critical path


include/ofi_rbuf.h
```diff
@@ -83,6 +83,21 @@ static inline void name ## _free(struct name *cq)		\
...
+                                                                \
+static inline struct name * name ## _create_from_pool(size_t size, \
+					struct ofi_bufpool *pool) \
+{                                                               \
+        struct name *cq;                                        \
+        cq = (struct name*) ofi_buf_alloc(pool);		\
+        if (cq)                                                 \
+                name ##_init(cq, roundup_power_of_two(size));   \
+        return cq;                                              \
+}                                                               \
```
shefty: This is mixing separate abstractions together in a non-intuitive way.  The user is expected to create an appropriate buffer pool, but then use macros to alloc/free elements from it?

It seems that there should at least be some wrapper around creating the buffer pool for this purpose.

snadampal: Thanks for review, @shefty . could you please elaborate on this statement? I'm failing to understand the rational for creating the pool here for this purpose.
>It seems that there should at least be some wrapper around creating the buffer pool for this purpose.

My idea was to reuse the already existing pools. hence used the two wrappers that ofi already supports.
(1) one for ofi_buffer_pool_create/grow which we do mostly during the initialization and
(2) use those pre-allocated pools even for the circular queues to get both the features.

shefty: I understand the intent.

Start with the buffer pool abstraction.  A buffer pool manages a set of *fixed-size* objects.  All objects must be the same size.  This is why the alloc() call does not take a size.  Your abstraction here takes a size as input, but that size must always be the same and must always match the size that was specified when the buffer pool was created.

You want an abstraction that defines a buffer pool of circular queues.  Each queue will be the same size.  The call to alloc() an object from the pool should just return a queue, with free() returning it to the pool.

I don't know that you need to reuse the existing static inline macros, since they aren't called.  It's not great, but maybe just duplicate this struct?

struct name {							\
	size_t		size;					\
	size_t		size_mask;				\
	size_t		rcnt;					\
	size_t		wcnt;					\
	entrytype	buf[];					\
};

so the existing cirque defines are usable?  (Maybe duplicate the init() call as well?)  Then the existing bufpool calls should be usable directly.

snadampal: yeah, on the circular queue duplicating the `struct` and `init` might be sufficient. but I'm also using the `ofi_recvwindow` abstraction and that has 6 macros and I have to duplicate all of them :(

If you think, though it's redundant, a separate implementation is preferred then I will implement all the required macros internal to efa/rdm.

snadampal: Hi @shefty , I've implemented the required macros locally to efa/rdm. Please review this version and let's know if this looks good to you.

shefty: No, I don't think we should duplicate all of that.  Hopefully we can avoid any duplicate code.

What if we have OFI_DECLARE_CIRQUE() and OFI_DECLARE_CIRQUE_STRUCT().  The latter defines the struct and init.  The first calls STRUCT, then defines create / free.

The recvwindow macro only needs to call OFI_DECLARE_CIRQUE_STRUCT().

The recvwindow macro could also declare a 'named' bufpool, which is referenced directly by the buf_alloc/free calls.  Can the bufpool be stored as a field in 'struct name' of the recvwindow (e.g. next to exp_msg_id)?  If so, it doesn't need to be named.  There would need to be 'create/destroy' calls, which would create the actual bufpool referenced by the recvwindow code.  The create call would be simplified versus bufpool_create.

Does this all make sense?

snadampal: I see that `recvwindow` needs everything from circular buffer.

what if i just use the existing DECL as they are and implement the new functions locally instead of inside the macro template?
something like this??

~~~
OFI_DECL_RECVWIN_BUF(struct efa_rdm_pke*, efa_rdm_robuf, uint32_t); --> 

static inline
struct recvwin_cirq * recvwin_cirq_create_from_pool(size_t size,
                                        struct ofi_bufpool *pool)
{
        struct recvwin_cirq *cq;
        cq = (struct recvwin_cirq*) ofi_buf_alloc(pool);
        if (cq)
                recvwin_cirq_init(cq, roundup_power_of_two(size));
        return cq;
}

static inline
void recvwin_cirq_free_to_pool(struct recvwin_cirq *cq)
{
        ofi_buf_free(cq);
}

static inline
struct efa_rdm_robuf *ofi_recvwin_buf_alloc_from_pool(struct efa_rdm_robuf *recvq,
                unsigned int size,  struct ofi_bufpool *pool)
{
        assert(size == roundup_power_of_two(size));
        recvq->exp_msg_id = 0;
        recvq->win_size = size;
        recvq->pending  = recvwin_cirq_create_from_pool(recvq->win_size, pool);
        return recvq;
}

static inline
void ofi_recvwin_free_to_pool(struct efa_rdm_robuf *recvq)
{
        recvwin_cirq_free_to_pool(recvq->pending);
}
~~~

regarding having a named buffer pool in the recvwindow, since we don't want to allocate the pool during the circular buffer create, and the pool is owned by some other component, it may not be food to store it in circular buffer...

shefty: Yes, I think that's closer to what we want.  There's a level of abstraction being built over the definitions.  (Some undefined 'recvwin manager' type thing, which isn't formalized and is left to the provider to define).

I was thinking of adjusting the lower-level definitions, sort of what's like what's shown below (not even compile tested), then your changes layered on top.

recvwin does not use circque alloc/free calls, so don't define them.  Remove the recvwin alloc/free calls, to force using custom calls.

The custom calls (e.g. ofi_recvwin_buf_alloc_from_pool()) can use the recvwin init call below, plus assert that the size is <= the size of the object returned from the bufpool.

```
diff --git a/include/ofi_rbuf.h b/include/ofi_rbuf.h
index e38eee6ac..21e28c5f9 100644
--- a/include/ofi_rbuf.h
+++ b/include/ofi_rbuf.h
@@ -51,7 +51,7 @@
 /*
  * Circular queue/array template
  */
-#define OFI_DECLARE_CIRQUE(entrytype, name)                     \
+#define OFI_DECLARE_CIRQUE_STRUCT(entrytype, name)             \
 struct name {                                                  \
        size_t          size;                                   \
        size_t          size_mask;                              \
@@ -68,7 +68,10 @@ static inline void name ## _init(struct name *cq, size_t size)>
        cq->rcnt = 0;                                           \
        cq->wcnt = 0;                                           \
 }                                                              \
-                                                               \
+void dummy ## name (void) /* work-around global ; scope */
+
+#define OFI_DECLARE_CIRQUE(entrytype, name)                    \
+OFI_DECLARE_CIRQUE_STRUCT(entrytype, name)                     \
 static inline struct name * name ## _create(size_t size)       \
 {                                                              \
        struct name *cq;                                        \
@@ -82,8 +85,7 @@ static inline struct name * name ## _create(size_t size)      \
 static inline void name ## _free(struct name *cq)              \
 {                                                              \
        free(cq);                                               \
-}                                                              \
-void dummy ## name (void) /* work-around global ; scope */
+}
 
 #define ofi_cirque_isempty(cq)         ((cq)->wcnt == (cq)->rcnt)
 #define ofi_cirque_usedcnt(cq)         ((cq)->wcnt - (cq)->rcnt)
diff --git a/include/ofi_recvwin.h b/include/ofi_recvwin.h
index 2d1073a8a..0e7bf0ffd 100644
--- a/include/ofi_recvwin.h
+++ b/include/ofi_recvwin.h
@@ -50,29 +50,22 @@
 #include <ofi_rbuf.h>
 
 #define OFI_DECL_RECVWIN_BUF(entrytype, name, id_type)                 \
-OFI_DECLARE_CIRQUE(entrytype, recvwin_cirq);                           \
+OFI_DECLARE_CIRQUE_STRUCT(entrytype, recvwin_cirq);                    \
 struct name {                                                          \
        id_type exp_msg_id;                                             \
        id_type win_size;                                               \
        struct recvwin_cirq *pending;                                   \
 };                                                                     \
                                                                        \
-static inline struct name *                                            \
-ofi_recvwin_buf_alloc(struct name *recvq, unsigned int size)           \
+static inline void                                                     \
+ofi_recvwin_init(struct name *recvq, unsigned int size)                        \
 {                                                                      \
        assert(size == roundup_power_of_two(size));                     \
        recvq->exp_msg_id = 0;                                          \
        recvq->win_size = size;                                         \
-       recvq->pending  = recvwin_cirq_create(recvq->win_size);         \
        return recvq;                                                   \
 }                                                                      \
                                                                        \
-static inline void                                                     \
-ofi_recvwin_free(struct name *recvq)                                   \
-{                                                                      \
-       recvwin_cirq_free(recvq->pending);                              \
-}                                                                      \
-                                                                       \
```

snadampal: thanks, @shefty . just to be sure, are you suggesting a new `OFI_DECL_RECVWIN_BUF` macro for the pool support? 
otherwise, if we remove the alloc/free from the existing recvwind macro inside `ofi_recvwindow.h`, it might break other providers who are using it, right?

snadampal: Hi @shefty thanks for the discussion. I have changed the recvwindow macro as it's currently being used only by EFA provider. and as we aligned, i moved the allocation part to provider.
please let me know if this looks good now.

include/ofi_rbuf.h
```diff
@@ -51,24 +51,26 @@
...
+};                                                              \
+                                                                \
+static inline void name ## _init(struct name *cq, size_t size)  \
+{                                                               \
+        assert(size == roundup_power_of_two(size));             \
+        cq->size = size;                                        \
+        cq->size_mask = cq->size - 1;                           \
+        cq->rcnt = 0;                                           \
+        cq->wcnt = 0;                                           \
+}                                                               \
```
shefty, line 70: The macro is continuing here, which is unexpected.  Also, there's a line:
```
void dummy ## name (void) /* work-around global ; scope */
```
That probably should be part of this first macro, rather than the second one. Which means you'd want to continue with the macro, but copy the void dummy line at the end.

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -26,6 +24,56 @@ OFI_DECL_RECVWIN_BUF(struct efa_rdm_pke*, efa_rdm_robuf, uint32_t);
...
+}
+
+static inline
+void ofi_recvwin_free(struct efa_rdm_robuf *recvq, bool free_to_bufpool)
+{
+	if (free_to_bufpool)
+		ofi_buf_free(recvq->pending);
+	else
+		free(recvq->pending);
+}
```
shefty, line 75: If you're going to place both the above functions into the efa provider, use the efa function prefix, not the generic ofi one.  Since the functions reference efa structures directly, I would keep them here and rename the prefix.

include/ofi_rbuf.h
snadampal: thanks, I have fixed it now.

prov/efa/src/rdm/efa_rdm_peer.h
snadampal: done.

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -246,6 +246,21 @@ int efa_rdm_ep_create_buffer_pools(struct efa_rdm_ep *ep)
 	if (ret)
 		goto err_free;
 
+	ret = ofi_bufpool_create(&ep->peer_robuf_pool,
+				(sizeof(struct efa_rdm_pke*) * (roundup_power_of_two(efa_env.recvwin_size)) +
+				sizeof(struct recvwin_cirq)),
+				EFA_RDM_BUFPOOL_ALIGNMENT, 0, /* no limit to max_cnt */
+				EFA_RDM_EP_MIN_PEER_POOL_SIZE,
+				/* Don't track usage, because endpoint can be closed without removing entries from AV */
```
shijin-aws: this comment needs update. Also, I think we should track usage. right?

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -282,6 +297,9 @@ int efa_rdm_ep_create_buffer_pools(struct efa_rdm_ep *ep)
 	if (ep->peer_map_entry_pool)
 		ofi_bufpool_destroy(ep->peer_map_entry_pool);
 
+	if (ep->peer_robuf_pool)
+                ofi_bufpool_destroy(ep->peer_robuf_pool);
```
shijin-aws: indent is off.

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -808,6 +826,9 @@ static void efa_rdm_ep_destroy_buffer_pools(struct efa_rdm_ep *efa_rdm_ep)
 
 	if (efa_rdm_ep->peer_map_entry_pool)
 		ofi_bufpool_destroy(efa_rdm_ep->peer_map_entry_pool);
+
+	if (efa_rdm_ep->peer_robuf_pool)
+                ofi_bufpool_destroy(efa_rdm_ep->peer_robuf_pool);
```
shijin-aws: indent

prov/efa/src/rdm/efa_rdm_peer.h
```diff
@@ -26,6 +24,56 @@ OFI_DECL_RECVWIN_BUF(struct efa_rdm_pke*, efa_rdm_robuf, uint32_t);
...
+ *
+ */
+static inline
+int efa_recvwin_buf_alloc(struct efa_rdm_robuf *recvq,
+                unsigned int size,  bool alloc_from_bufpool, struct ofi_bufpool *pool)
+{
+	assert(size == roundup_power_of_two(size));
+	recvq->exp_msg_id = 0;
+	recvq->win_size = size;
+	if (alloc_from_bufpool) {
```
shijin-aws, line 52: Where do we toggle this boolean? It seems this PR always make it as true

prov/efa/src/rdm/efa_rdm_ep_fiops.c
shijin-aws: We are destroying all peers during ep close now: https://github.com/ofiwg/libfabric/commit/6e4645de31c998fa0df12d5a60101949ba8abdf7 so we should make this flags as 0 so we can track it

prov/efa/src/rdm/efa_rdm_peer.h
snadampal: This is not a toggle, it's a config option to choose how the recvwindow has to allocate the buffers.
Earlier the recvwindow used to allocate from calloc. I have extended that to allocate from either calloc or pre-allocated buffer pool. For the peer robuf we wanted to allocate from bufferpool so, we are calling this with `alloc_from_bufpool=true`.
But in the future, if any other component wants to use the recvwindow interface but can't use bufferpools, it can call the same alloc with `alloc_from_bufpool=false`.

shijin-aws: Thanks, makes sense

prov/efa/src/rdm/efa_rdm_ep_fiops.c
snadampal: thanks, I have fixed it.

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -246,6 +246,20 @@ int efa_rdm_ep_create_buffer_pools(struct efa_rdm_ep *ep)
...
+				(sizeof(struct efa_rdm_pke*) * (roundup_power_of_two(efa_env.recvwin_size)) +
+				sizeof(struct recvwin_cirq)),
+				EFA_RDM_BUFPOOL_ALIGNMENT, 0, /* no limit to max_cnt */
+				EFA_RDM_EP_MIN_PEER_POOL_SIZE,
+				0);
+
+	if (ret)
+		goto err_free;
+
+	ret = ofi_bufpool_grow(ep->peer_robuf_pool);
```
shijin-aws, line 259: Not for this PR, but we may consider moving all pool grow into the ep enable stage as that is the last call before the real transfer can happen.

shijin-aws: Can do that in a separate PR after we resolve the whole lazy initialization problem

shefty: I would add this feature directly to the bufpool (in a separate PR) via a flag passed into create.

snadampal:
there is a memory leak with the recvwindow. at this point not sure if it was present in the calloc path and carried here or got newly introduced. I'm looking into it.

snadampal:
Ok, the memory leak is coming from the unit test, here is the PR to fix it:
https://github.com/ofiwg/libfabric/pull/11487
after the unit test PR is merged, I will rebase the current PR and that should make the issue go away.

snadampal:
the unit test is fixed, so, I have rebased the PR to main latest.

PR #11475, opened 2025-10-06
prov/efa: Slide recv-win on RTM/RTA error

a-szegel:
In order to continue processing future packets correctly, we need to progress the recv window when a RTM or RTA packet has been successfully delivered. If we fail processing a packet, we need to write the error to the CQ (when applicable), and continue to do work. Returning early here makes our recv window index be off by 1.


sunkuamzn (approved):

a-szegel:
bot:aws:retest

PR #11477, opened 2025-10-07
prov/opx: Use getpid() instead of gettid() for POSIX compliancy.

belynam:
The 'gettid()' function is not POSIX compliant, and older versions of glibc may not have that function available in unistd.h, causing a compile failure. This changes the call to 'getpid()', which is POSIX compliant and should always be present in glibc.


belynam:
@j-xiong Once merged, can you please also add this into the v2.3.x branch so it makes the 2.3.1 RC? Or should I post a separate PR for that?

j-xiong:
@belynam I have added the `for-2.3.x` label and will make sure it gets into 2.3.1. Thanks!

PR #11478, opened 2025-10-07
efa/cq: Add traces for completions bypassing util-cq

alekswn:
This patch adds LTTng tracepoint for util cq bypass path introduced in commit https://github.com/ofiwg/libfabric/commit/b3abf8a40e185f7db239aeab2cb37251e183106f

This tracepoint allows to trace latency of efa-direct operations.


shijin-aws (approved):
thanks!

PR #11487, opened 2025-10-10
prov/efa: fix memory leak from `efa_srx_unexp_pkt` unit test

snadampal:
efa_srx_unexp_pkt test is creating a fake peer explicitly to emualte an unexpeted rxe scenario. But it wasn't destroying the fake peer at the end, so, leaking the buffers and causing memory faults when the endpoint is closed.

### DIFF
--- a/prov/efa/test/efa_unit_test_srx.c
+++ b/prov/efa/test/efa_unit_test_srx.c
@@ -131,4 +131,7 @@ void test_efa_srx_unexp_pkt(struct efa_resource **state)
         efa_rdm_pke_release_rx(pke);
         efa_rdm_rxe_release(rxe);
         ofi_genlock_unlock(srx_ctx->lock);
+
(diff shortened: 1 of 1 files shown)


prov/efa/test/efa_unit_test_srx.c
```diff
@@ -131,4 +131,7 @@ void test_efa_srx_unexp_pkt(struct efa_resource **state)
         efa_rdm_pke_release_rx(pke);
         efa_rdm_rxe_release(rxe);
         ofi_genlock_unlock(srx_ctx->lock);
+
+	/* Destroy the fake peer constructed above */
+	efa_rdm_peer_destruct(&peer, efa_rdm_ep);
```
sunkuamzn: nit: indent

Can be fixed later

snadampal: thanks, fixed.

PR #11488, opened 2025-10-10
fabtests/build: fix build issues for MacOS

alekswn:
Fabtests fail to build at my laptop (M1 Pro, macOS Sequoia 15.7.1, Apple clang version 17.0.0 (clang-1700.3.19.1)).

The build error is caused by mismatch in definition of `brk()` standard library function. It returns `int` in GLIBC, but in BSD libraries it is defined with `void *` return type. Clang treats implicit cast from `void *` to `int` as an error.

```
unit/mr_cache_evict.c:195:6: error: incompatible pointer to integer conversion assigning to 'int' from 'void *' [-Wint-conversion]
  195 |         ret = brk(rewind_brk);
``` 

Also fixed warnings from build log:
- Non-portable scanf format string at rdm_stress.c:654
- unused function  `print_timer()`
- deprecated stdlib functions `brk` and `sbrk`

### DIFF
--- a/fabtests/multinode/src/timing.c
+++ b/fabtests/multinode/src/timing.c
@@ -54,11 +54,14 @@ void multi_timer_stop(struct multi_timer *timer)
 		timer->end = ft_gettime_ns();
 }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
 static inline void print_timer(struct multi_timer timer, char* info)
 {
 	PRINTF("rank: %i, start: %ld, end: %ld, %s\n",
 		timer.rank, timer.start, timer.end, info);
 }
+#pragma GCC diagnostic pop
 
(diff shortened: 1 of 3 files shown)


fabtests/multinode/src/timing.c
```diff
@@ -54,6 +54,7 @@ void multi_timer_stop(struct multi_timer *timer)
 		timer->end = ft_gettime_ns();
 }
 
+__attribute__((unused))
```
j-xiong: Windows build doesn't like this line. We have dummy definition of `__attribute__(x)` in libfabric `include/windows/osd.h`, but it is not present in fabtests headers.

alekswn: Thank you for clarification!

I've replaced `__attribute__(x)` with `#pragma GCC diagnostics`

j-xiong:
CI failure is unrelated.

PR #11492, opened 2025-10-11
prov/efa: Remove rx_msg_flags from efa_rdm_msg_recv/efa_rdm_msg_recvv

jiaxiyan:
rx_msg_flags only apply to fi_recvmsg/fi_trecvmsg. For fi_recv, use the default operational flags rx_op_flags.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_msg.c
+++ b/prov/efa/src/rdm/efa_rdm_msg.c
@@ -974,8 +974,7 @@ ssize_t efa_rdm_msg_recv(struct fid_ep *ep_fid, void *buf, size_t len,
 	iov.iov_len = len;
 
 	efa_rdm_msg_construct(&msg, &iov, &desc, 1, src_addr, context, 0);
-	return efa_rdm_msg_generic_recv(ep, &msg, 0, 0, ofi_op_msg,
-					efa_rdm_rx_flags(ep) | ep->base_ep.util_ep.rx_msg_flags);
+	return efa_rdm_msg_generic_recv(ep, &msg, 0, 0, ofi_op_msg, efa_rdm_rx_flags(ep));
 }
 
 static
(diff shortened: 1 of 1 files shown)


shijin-aws:
Now I am not sure whether this change is correct, see shm's example: https://github.com/ofiwg/libfabric/blob/main/prov/shm/src/smr_msg.c#L284-L298, which is the same with the current efa provider's behavior

shijin-aws:
I think we even should not `|= util_ep.rx_msg_flags` for efa_rdm_recv

shijin-aws:
https://github.com/ofiwg/libfabric/commit/978ec0261d0a29b0e60a0de9ebb7b48f528e69ee actually didn't cleanup `util_ep.rx_msg_flags` from recv and recvv, which seems wrong to me. 


If application requests FI_COMPLETION in hints.rx_attr.flags, the efa_rdm_ep_rx_flags should already have FI_COMPLETION. when application bind cq with ep as FI_SELECTIVE_COMPLETION, rx_msg_flags should have no FI_COMPLETION, in that regard it shouldn't cause any behavior change through your PR. .
Application can use this way to selectively generate completion by pass FI_COMPLETION or not via `fi_trecvmsg`/`fi_recvmsg`, while the default `fi_trecv`/`fi_recv` can still generate completion by default

Even If application doesn't request FI_COMPLETION in hints.rx_attr.flags, util_ep will add FI_COMPLETION to tx_op_flags and tx_msg_flags if CQ is not bound with FI_SELECTIVE_COMPLETION.  In that regard, this PR won't change the behavior either.

Therefore It is still unclear to me what issue this PR is trying to solve. Missing completions or unexpected completions?

jiaxiyan:
bot:aws:retest

shijin-aws (approved):

PR #11493, opened 2025-10-12
prov/ucx: Fix completion entries insertion for RMA operations.

tatarintsevsv:
- Ignore the absence of the FI_COMPLETION flag if CQ configured without FI_SELECTIVE_COMPLETION flag.
- Error completions must generated for all operations, including those for which a completion was not requested

### DIFF
--- a/prov/ucx/src/ucx_msg.c
+++ b/prov/ucx/src/ucx_msg.c
@@ -120,6 +120,9 @@ static ssize_t ucx_mrecvmsg(struct fid_ep *ep, const struct fi_msg *msg,
 static ssize_t ucx_recvmsg(struct fid_ep *ep, const struct fi_msg *msg,
 			   uint64_t flags)
 {
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
+	flags |= u_ep->ep.rx_msg_flags;
 	if (flags & (FI_REMOTE_CQ_DATA | FI_PEEK | FI_CLAIM))
 		return -FI_EBADFLAGS;
 
@@ -227,11 +230,12 @@ ssize_t ucx_sendmsg(struct fid_ep *ep, const struct fi_msg *msg,
 		.context = msg->context,
 		.data = 0,
 	};
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
 
 	if(flags & FI_REMOTE_CQ_DATA)
 		return -FI_EBADFLAGS;
 
-	return ucx_do_sendmsg(ep, &tmsg, flags, UCX_MSG);
+	return ucx_do_sendmsg(ep, &tmsg, flags | u_ep->ep.tx_msg_flags, UCX_MSG);
 }
 
 
--- a/prov/ucx/src/ucx_rma.c
+++ b/prov/ucx/src/ucx_rma.c
@@ -158,14 +158,18 @@ static ssize_t ucx_writemsg(struct fid_ep *ep,
 			    const struct fi_msg_rma *msg,
 			    uint64_t flags)
 {
-	return ucx_proc_rma_msg(ep, msg, flags, UCX_DO_WRITE);
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
+	return ucx_proc_rma_msg(ep, msg, flags | u_ep->ep.tx_msg_flags, UCX_DO_WRITE);
 }
 
 static ssize_t ucx_readmsg(struct fid_ep *ep,
 			   const struct fi_msg_rma *msg,
 			   uint64_t flags)
 {
-	return ucx_proc_rma_msg(ep, msg, flags, UCX_DO_READ);
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
+	return ucx_proc_rma_msg(ep, msg, flags | u_ep->ep.tx_msg_flags, UCX_DO_READ);
 }
 
 void ucx_rma_callback(void *request, ucs_status_t status)
@@ -193,14 +197,12 @@ void ucx_rma_callback(void *request, ucs_status_t status)
 		if (cntr)
 			cntr->cntr_fid.ops->adderr(&cntr->cntr_fid, 1);
 
-		if (ucx_req->completion.flags & FI_COMPLETION) {
-			ucx_write_error_completion(ucx_req->ep->ep.tx_cq,
-						   ucx_req->completion.op_context,
-						   ucx_req->completion.flags,
-						   status,
-						   -ucx_translate_errcode(status),
-						   0, ucx_req->completion.tag);
-		}
+		ucx_write_error_completion(ucx_req->ep->ep.tx_cq,
+					   ucx_req->completion.op_context,
+					   ucx_req->completion.flags,
+					   status,
+					   -ucx_translate_errcode(status),
+					   0, ucx_req->completion.tag);
 	}
 	ucx_req_release(request);
 }
--- a/prov/ucx/src/ucx_tagged.c
+++ b/prov/ucx/src/ucx_tagged.c
@@ -81,6 +81,9 @@ static ssize_t ucx_tagged_recvmsg(struct fid_ep *ep,
 				  const struct fi_msg_tagged *msg,
 				  uint64_t flags)
 {
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
+	flags |= u_ep->ep.rx_msg_flags;
 	if (flags & FI_REMOTE_CQ_DATA)
 		return -FI_EBADFLAGS;
 
@@ -101,6 +104,8 @@ static ssize_t ucx_tagged_sendmsg(struct fid_ep *ep,
 				  const struct fi_msg_tagged *msg,
 				  uint64_t flags)
 {
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
 	if(flags & FI_REMOTE_CQ_DATA)
 		return -FI_EBADFLAGS;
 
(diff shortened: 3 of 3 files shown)


prov/ucx/src/ucx_rma.c
```diff
@@ -305,7 +304,7 @@ static ssize_t ucx_proc_rma_msg(struct fid_ep *ep,
 		else
 			ofi_ep_cntr_inc(&u_ep->ep, CNTR_RD);
 
-		if (flags & FI_COMPLETION)
+		if (flags & FI_COMPLETION || u_ep->ep.tx_op_flags & FI_COMPLETION)
```
j-xiong: `tx_op_flags` is already considered when `ucx_proc_rma_msg` is called.

prov/ucx/src/ucx_rma.c
```diff
@@ -178,7 +178,8 @@ void ucx_rma_callback(void *request, ucs_status_t status)
 		else
 			ofi_ep_cntr_inc(&(ucx_req->ep->ep), CNTR_RD);
 
-		if (ucx_req->completion.flags & FI_COMPLETION)
+		if (ucx_req->ep->ep.tx_op_flags & FI_COMPLETION ||
+		    ucx_req->completion.flags & FI_COMPLETION)
```
j-xiong: Same here. 

Should use `!(u_ep->ep.tx_op_flags & FI_SELECTIVE_COMPLETION)` to check for the case that selective completion is not asked for.

aingerson: I think you should just be able to check the completion.flags which are set when initializing the request. The op flags are set to force include `FI_COMPLETION` when `FI_SELECTIVE_COMPLETION` is not requested so it should be ok. See [here](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_ep.c#L51)

I think the missing thing in the current code is the OR'ing in of the msg flags on the *msg APIs to make sure `FI_COMPLETION` is set in the req->completion.flags for the non `FI_SELECTIVE_COMPLETION` case.

@tatarintsevsv Can you explain the selective completion path that is incorrectly writing a completion when it shouldn't be? I don't immediately see it.

tatarintsevsv: > I think you should just be able to check the completion.flags which are set when initializing the request. The op flags are set to force include `FI_COMPLETION` when `FI_SELECTIVE_COMPLETION` is not requested so it should be ok. See [here](https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_ep.c#L51)
> 
> I think the missing thing in the current code is the OR'ing in of the msg flags on the *msg APIs to make sure `FI_COMPLETION` is set in the req->completion.flags for the non `FI_SELECTIVE_COMPLETION` case.
> 
> @tatarintsevsv Can you explain the selective completion path that is incorrectly writing a completion when it shouldn't be? I don't immediately see it.

According to [manual](https://github.com/ofiwg/libfabric/blob/e912b06e9a8eaf2e6186f313cf72a6f515c63496/man/fi_endpoint.3.md?plain=1#L308), in this this case cq records must be generated

```
fi_ep_bind(ep, &cq->fid, FI_TRANSMIT|FI_RECV);
fi_writemsg(ep, &msg, 0);
```
but it's so only if we pass `FI_COMPLETION` to fi_writemsg flags.

As far as I understand, the correct usage of `FI_COMPLETION` / `FI_SELECTIVE_COMPLETION` is
```
fi_ep_bind(ep, &cq->fid, FI_TRANSMIT|FI_RECV|FI_SELECTIVE_COMPLETION);
fi_writemsg(ep, &msg, FI_COMPLETION); // completion must be generated
fi_writemsg(ep, &msg, 0); // completion doesn't matter, will not be generated
```
and
```
fi_ep_bind(ep, &cq->fid, FI_TRANSMIT|FI_RECV);
fi_writemsg(ep, &msg, FI_COMPLETION); // completion must be generated
fi_writemsg(ep, &msg, 0); // completion must be generated anyway
```

shijin-aws: U just need to use `flags | util_ep->tx_msg_flags` as the internal flags when you implement `fi_*msg`. tx_msg_flags will be 0 when FI_SELECTIVE_COMPLETION is used when ep bind, it will be `FI_COMPLETION` when there is no FI_SELECTIVE_COMPLETION

shijin-aws: Then you check whether that internal flags have `FI_COMPLETION` to determine whether you want to generate completions

tatarintsevsv:
After looking more closely, it seems this error only occurs with ucx_writemsg and ucx_readmsg. I'll fix the PR.

tatarintsevsv:
can anyone review last changes?

prov/ucx/src/ucx_rma.c
```diff
@@ -158,14 +158,16 @@ static ssize_t ucx_writemsg(struct fid_ep *ep,
 			    const struct fi_msg_rma *msg,
 			    uint64_t flags)
 {
-	return ucx_proc_rma_msg(ep, msg, flags, UCX_DO_WRITE);
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+	return ucx_proc_rma_msg(ep, msg, flags | u_ep->ep.tx_op_flags, UCX_DO_WRITE);
```
j-xiong: This is incorrect. The `flags` passed to the `msg` version override the default op flags, no adds to.

prov/ucx/src/ucx_rma.c
```diff
@@ -158,14 +158,16 @@ static ssize_t ucx_writemsg(struct fid_ep *ep,
...
 }
 
 static ssize_t ucx_readmsg(struct fid_ep *ep,
 			   const struct fi_msg_rma *msg,
 			   uint64_t flags)
 {
-	return ucx_proc_rma_msg(ep, msg, flags, UCX_DO_READ);
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+	return ucx_proc_rma_msg(ep, msg, flags | u_ep->ep.rx_op_flags, UCX_DO_READ);
```
j-xiong: same here.

zachdworkin:
libfabric ci is failing udp fabtests. Can you please rebase to pickup the changes from #11501 to fix this issue?

prov/ucx/src/ucx_rma.c
tatarintsevsv: Agreed. I've fixed it so that only FI_COMPLETION is taken into account in the internal flags.

prov/ucx/src/ucx_rma.c
tatarintsevsv: as it pointed by @j-xiong this is not correct - we must adds flags but not override.
fixed with only FI_COMPLETION is taken into account

prov/ucx/src/ucx_rma.c
```diff
@@ -158,13 +158,23 @@ static ssize_t ucx_writemsg(struct fid_ep *ep,
 			    const struct fi_msg_rma *msg,
 			    uint64_t flags)
 {
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
+	/* Completion required if FI_SELECTIVE_COMPLETION was not set */
+	if (u_ep->ep.tx_op_flags & FI_COMPLETION)
+		flags |= FI_COMPLETION;
```
aingerson: This is still incorrect. 

The *msg APIs don't take into account the op_flags. They are specifically to override the op_flags so you shouldn't be using the op_flags on the *msg paths. You should only use the u_ep->tx_msg_flags here which only has FI_COMPLETION for the non FI_SELECTIVE_COMPLETION case and are 0 if FI_SELECTIVE_COMPLETION is being used. You should be able to always OR in the tx_msg_flags for the msg APIs

tatarintsevsv: thanks for explaining. looks it just `flags|msg_flags`

prov/ucx/src/ucx_rma.c
aingerson: Should be `return ucx_proc_rma_msg(ep, msg, flags | u_ep->ep.tx_msg_flags, UCX_DO_WRITE);`

prov/ucx/src/ucx_rma.c
aingerson: Should be `return ucx_proc_rma_msg(ep, msg, flags | u_ep->ep.tx_msg_flags, UCX_DO_READ);`
(A read is still a transmitted operation)

tatarintsevsv: Fixed. Also fixed recvmsg for msg and tagged

prov/ucx/src/ucx_msg.c
```diff
@@ -120,6 +120,9 @@ static ssize_t ucx_mrecvmsg(struct fid_ep *ep, const struct fi_msg *msg,
 static ssize_t ucx_recvmsg(struct fid_ep *ep, const struct fi_msg *msg,
 			   uint64_t flags)
 {
+	struct ucx_ep *u_ep = container_of(ep, struct ucx_ep, ep.ep_fid);
+
+	flags |= u_ep->ep.tx_msg_flags;
```
aingerson: rx_msg_flags

prov/ucx/src/ucx_tagged.c
```diff
@@ -94,13 +96,15 @@ static ssize_t ucx_tagged_recvmsg(struct fid_ep *ep,
 	if (flags & FI_PEEK)
 		return ucx_tagged_peek(ep, msg, flags);
 
-	return ucx_do_recvmsg(ep, msg, flags, UCX_TAGGED);
+	return ucx_do_recvmsg(ep, msg, flags | u_ep->ep.tx_msg_flags, UCX_TAGGED);
```
aingerson: rx_msg_flags

prov/ucx/src/ucx_msg.c
tatarintsevsv: fixed

prov/ucx/src/ucx_tagged.c
tatarintsevsv: think ucx_tagged_peek() flags must be OR'ed with rx_msg_flags too

aingerson: I agree! Good catch!

tatarintsevsv:
rebased. waiting for review

j-xiong (approved):

j-xiong:
The CI failure is unrelated.

PR #11496, opened 2025-10-13
prov/efa: Fix the queued ope progress in ep close

shijin-aws:
The current "fall-through" in the switch is wrong: when the repost of the queued ope succeeds, we shouldn't release the ope. The ope release should only happen for non-EOR/RECEIPT opes


shijin-aws:
Please hold on merging this as I saw some timeout in my local testing... Will clear it after confirming it is not caused by this PR

shijin-aws:
> Please hold on merging this as I saw some timeout in my local testing... Will clear it after confirming it is not caused by this PR

All cleared, unrelated infra issue, I am able to run test with this PR in a loop successfully over the night

PR #11498, opened 2025-10-13
prov/efa: Update shared domain caps and modes

jiaxiyan:
When the application tries to reuse domain in fi_getinfo, add additional requested capabilities and modes to the domain as long as the provider supports them.

### DIFF
--- a/prov/efa/src/efa_user_info.c
+++ b/prov/efa/src/efa_user_info.c
@@ -167,6 +167,11 @@ int efa_user_info_check_domain_object(const struct fi_info *hints,
 	}
 
 	dupinfo->domain_attr->domain = hints->domain_attr->domain;
+
+	util_domain->info_domain_caps |= dupinfo->caps | dupinfo->domain_attr->caps;
+	util_domain->info_domain_mode |= dupinfo->mode | dupinfo->domain_attr->mode;
+	util_domain->mr_mode |= dupinfo->domain_attr->mr_mode;
+
 	return 0;
 }
 
--- a/prov/efa/test/efa_unit_test_info.c
+++ b/prov/efa/test/efa_unit_test_info.c
@@ -811,17 +811,18 @@ static void test_info_reuse_fabric_domain(setup_hints_func_t setup_func,
 				   bool expect_fabric_reuse, 
 				   bool expect_domain_reuse)
 {
-	struct fi_info *hints, *info1, *info2;
+	struct fi_info *hints1, *hints2, *info1, *info2;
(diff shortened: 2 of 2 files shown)


prov/efa/src/efa_user_info.c
```diff
@@ -167,6 +167,11 @@ int efa_user_info_check_domain_object(const struct fi_info *hints,
 	}
 
 	dupinfo->domain_attr->domain = hints->domain_attr->domain;
+
+	util_domain->info_domain_caps |= dupinfo->caps | dupinfo->domain_attr->caps;
+	util_domain->info_domain_mode |= dupinfo->mode | dupinfo->domain_attr->mode;
```
shijin-aws, line 172: domain caps/mode are not the same with info->caps/mode. Do we intend to mix them here?

jiaxiyan: Yes it is defined here https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_domain.c#L112-L114
and I am copying https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_user_info.c#L623-L625

shijin-aws: Thanks. `info_domain_caps` means info + domain caps I got it now

shijin-aws: nit: can u add a simple unit test on it

shijin-aws: It seems your existing unit test can already cover it, just need to add some assertions

jiaxiyan: updated

prov/efa/test/efa_unit_test_info.c
```diff
@@ -853,6 +854,11 @@ static void test_info_reuse_fabric_domain(setup_hints_func_t setup_func,
 	
 	if (expect_domain_reuse) {
 		assert_ptr_equal(info2->domain_attr->domain, domain);
+		util_domain = container_of(domain, struct util_domain, domain_fid);
+		assert_true(util_domain->info_domain_caps & (FI_MSG | FI_TAGGED | FI_LOCAL_COMM |
```
shijin-aws: Nit: you can avoid those duplicated macros via making your hints 1 and 2 as separate hints1 and hints2, and then assert can be changed as `assert_true((util_domain->info_domain_caps & (hints1->caps | hints2->caps)) == (hints1->caps | hints2->caps))`. Similar for other fields

PR #11502, opened 2025-10-14
src/hmem_cuda_gdrcopy: Adding more robust libgdrapi libpaths

nmazzilli3:
feat: Creating multiple libgdrapi libpath open attempts

Problem:
Intermittent failure on CI for P5en failing on gdrcopy open failure.

Solution:
  - Add array of lib_paths and required_symbols to validate dlopen against
  - Refactored error checks for each required_symbols into cuda_gdrcopy_load_symbol() -

Testing:
  - Validated cuda_gdrcopy_load_symbol() with valid and invalid symbols
  - libfabric:1356524:1760465932::core:core:cuda_gdrcopy_load_symbol():89&lt;warn&gt; Failed to find nonexistent_symbol: /lib64/libgdrapi.so.2: undefined symbol: nonexistent_symbol
  - Validated cuda_gdrcopy_dl_hmem_init() with libgdrapi.so.2 and libgdrapi.so path

### DIFF
--- a/src/hmem_cuda_gdrcopy.c
+++ b/src/hmem_cuda_gdrcopy.c
@@ -81,75 +81,57 @@ static pthread_spinlock_t global_gdr_lock;
 static void *gdrapi_handle;
 static struct gdrcopy_ops global_gdrcopy_ops;
 
-static int cuda_gdrcopy_dl_hmem_init(void)
+static int cuda_gdrcopy_load_symbol(void **func_ptr, const char *symbol_name)
 {
-	gdrapi_handle = dlopen("libgdrapi.so", RTLD_NOW);
-	if (!gdrapi_handle) {
-		FI_INFO(&core_prov, FI_LOG_CORE,
-			"Failed to dlopen libgdrapi.so\n");
-		return -FI_ENOSYS;
+	dlerror();
+	*func_ptr = dlsym(gdrapi_handle, symbol_name);
+	if (!*func_ptr) {
+		FI_WARN(&core_prov, FI_LOG_CORE, "Failed to find %s: %s\n", symbol_name, dlerror());
+		return -FI_ENODATA;
 	}
-
-	global_gdrcopy_ops.gdr_open = dlsym(gdrapi_handle, "gdr_open");
-	if (!global_gdrcopy_ops.gdr_open) {
-		FI_WARN(&core_prov, FI_LOG_CORE, "Failed to find gdr_open\n");
-		goto err_dlclose_gdrapi;
-	}
-
-	global_gdrcopy_ops.gdr_close = dlsym(gdrapi_handle, "gdr_close");
-	if (!global_gdrcopy_ops.gdr_close) {
-		FI_WARN(&core_prov, FI_LOG_CORE, "Failed to find gdr_close\n");
-		goto err_dlclose_gdrapi;
-	}
-
-	global_gdrcopy_ops.gdr_pin_buffer = dlsym(gdrapi_handle, "gdr_pin_buffer");
-	if (!global_gdrcopy_ops.gdr_pin_buffer) {
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find gdr_pin_buffer\n");
-		goto err_dlclose_gdrapi;
-	}
-
-	global_gdrcopy_ops.gdr_unpin_buffer = dlsym(gdrapi_handle, "gdr_unpin_buffer");
-	if (!global_gdrcopy_ops.gdr_unpin_buffer) {
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find gdr_unpin_buffer\n");
-		goto err_dlclose_gdrapi;
-	}
-
-	global_gdrcopy_ops.gdr_map = dlsym(gdrapi_handle, "gdr_map");
-	if (!global_gdrcopy_ops.gdr_map) {
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find gdr_map\n");
-		goto err_dlclose_gdrapi;
-	}
-
-	global_gdrcopy_ops.gdr_unmap = dlsym(gdrapi_handle, "gdr_unmap");
-	if (!global_gdrcopy_ops.gdr_unmap) {
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find gdr_unmap\n");
-		goto err_dlclose_gdrapi;
-	}
-
-	global_gdrcopy_ops.gdr_copy_to_mapping = dlsym(gdrapi_handle, "gdr_copy_to_mapping");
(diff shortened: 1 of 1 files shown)


src/hmem_cuda_gdrcopy.c
```diff
@@ -81,68 +81,87 @@ static pthread_spinlock_t global_gdr_lock;
...
-		goto err_dlclose_gdrapi;
+static int cuda_gdrcopy_dl_hmem_init(void)
+{	
+	const char *lib_paths[] = {
+		"libgdrapi.so.2", // BEST: stable major version
+		"libgdrapi.so", // Fallback: development version
+		"/lib64/libgdrapi.so.2",
+		"/lib64/libgdrapi.so",
+		"/usr/lib64/libgdrapi.so.2",
+		"/usr/lib64/libgdrapi.so",
```
j-xiong: `/lib64` and `/usr/lib64` are already in the default search path for `dlopen`.

`libgdrapi.so` is usually a symbolic link to the latest version installed (e.g. `libgdrapi.so.2`). It's used for building libraries / applications.   For `dlopen` the common practice is to use the `so` with the major version included.

So all needed here is to use `libgdrapi.so.2` instead of `libgdrapi.so`. The rest are unnecessary.

nmazzilli3: Ack, we previously only had libgdrapi.so so i'll keep that and remove the others.

src/hmem_cuda_gdrcopy.c
```diff
@@ -81,68 +81,83 @@ static pthread_spinlock_t global_gdr_lock;
...
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find gdr_copy_to_mapping\n");
+	if (cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_open, "gdr_open") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_close, "gdr_close") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_pin_buffer, "gdr_pin_buffer") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_unpin_buffer, "gdr_unpin_buffer") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_map, "gdr_map") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_unmap, "gdr_unmap") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_copy_to_mapping, "gdr_copy_to_mapping") ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_copy_from_mapping, "gdr_copy_from_mapping")) {
```
j-xiong: Can reuse `required_symbols` here.

src/hmem_cuda_gdrcopy.c
```diff
@@ -81,68 +81,83 @@ static pthread_spinlock_t global_gdr_lock;
...
-			"Failed to find gdr_map\n");
-		goto err_dlclose_gdrapi;
+static int cuda_gdrcopy_dl_hmem_init(void)
+{	
+	const char *lib_paths[] = {
+		"libgdrapi.so.2", // BEST: stable major version
+		"libgdrapi.so", // Fallback: development version
+		NULL
+	};
+
```
j-xiong: Remove this empty line.

src/hmem_cuda_gdrcopy.c
```diff
@@ -81,68 +81,83 @@ static pthread_spinlock_t global_gdr_lock;
...
+	const char *required_symbols[] = {
+		"gdr_open", "gdr_close", "gdr_pin_buffer", "gdr_unpin_buffer",
+		"gdr_map", "gdr_unmap", "gdr_copy_to_mapping", "gdr_copy_from_mapping",
+		NULL
+	};
+	Dl_info info;
+
+	for (int idx_1 = 0; lib_paths[idx_1]; idx_1++) {
```
j-xiong: use `i`, `j` instead of `idx_1`, `idx_2` for simplicity.

src/hmem_cuda_gdrcopy.c
```diff
@@ -81,68 +81,83 @@ static pthread_spinlock_t global_gdr_lock;
...
+			gdrapi_handle = handle;
+			FI_INFO(&core_prov, FI_LOG_CORE, "About to call dladdr\n");
+			void *gdr_open_ptr = dlsym(handle, "gdr_open");
+			if (gdr_open_ptr && dladdr(gdr_open_ptr, &info)) {
+				FI_INFO(&core_prov, FI_LOG_CORE,
+					"Loaded GDRCopy library: %s\n", info.dli_fname);
+			}
+			break;
+		}
+		else {
```
j-xiong: put `else` in the same line as `}`.

src/hmem_cuda_gdrcopy.c
```diff
@@ -81,68 +81,83 @@ static pthread_spinlock_t global_gdr_lock;
...
-		FI_WARN(&core_prov, FI_LOG_CORE,
-			"Failed to find gdr_copy_to_mapping\n");
+	if (cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_open, required_symbols[0]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_close, required_symbols[1]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_pin_buffer, required_symbols[2]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_unpin_buffer, required_symbols[3]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_map, required_symbols[4]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_unmap, required_symbols[5]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_copy_to_mapping, required_symbols[6]) ||
+	    cuda_gdrcopy_load_symbol((void**)&global_gdrcopy_ops.gdr_copy_from_mapping, required_symbols[7])) {
```
j-xiong: Upon further look, I think it's unnecessary to call `dlsym` twice. It would be simpler to replace L116~L123 with previous version of this block and remove the `required_symbols` array all together. And by doing this, the `dlsym` call at L128 also becomes redundant.

nmazzilli3: I have made the modification and verified it on my end.

j-xiong (approved):

PR #11507, opened 2025-10-15
fabtests: implement pausable timer

alekswn:
Replace global timespec variables with a flexible ft_timer union that supports pause/resume functionality. This allows for more accurate timing measurements in tests by excluding client-server synchronization time.

In particular, this change fixes issue with `fi_rma_bw -o writedata -p efa -f efa-direct`.
Currently bandwidth numbers  produced by that benchmark are wrong. In case of `writedata` client sends a stream of write messages to server and server replies with a send after each window. The purpose of  the benchmark is to measure bandwidth of writedata operation, not the synchronization time. However in some cases synchronization time can take up to 30% of overall timing, therefore bandwidth calculation is off by the same 30%.

Key changes:
- Add ft_timer union with pause/resume capabilities
- Replace timespec usage across all test files with new timer API
- Update timing functions in benchmark and test utilities
- `bandwidth_rma -o writedata` excludes reverse sync time from total timing
- ~~`bandwidth_rma` starts timer only after all completions from warm-up iterations had been received~~
- ~~Enable payload validation for warm-up iterations~~
- ~~Increase default warmup iterations from 10 to 64 to mach default window size~~
- Fix a bug in `uni_bandwidth` due to dafaulting to a global var `targs`
- Fixed a bug with timing in `efa_gda` fabtest

This improves timing precision in fabtests by allowing tests to exclude non-measurement periods from their timing calculations.

*Edit: Crossed out changes caused increase in BW wall clock timings causing CI timeouts. Removing for now*

### DIFF
--- a/fabtests/benchmarks/benchmark_shared.c
+++ b/fabtests/benchmarks/benchmark_shared.c
@@ -94,15 +94,14 @@ void ft_benchmark_usage(void)
 }
 
 /* Pingpong latency test with pre-posted receive buffers. */
-static int pingpong_pre_posted_rx(size_t inject_size)
+static int pingpong_pre_posted_rx(size_t inject_size, union ft_timer *timer)
 {
 	int ret, i;
 
 	if (opts.dst_addr) {
 		for (i = 0; i < opts.iterations + opts.warmup_iterations; i++) {
 			if (i == opts.warmup_iterations)
-				ft_start();
-
+				ft_start(timer);
 			if (opts.transfer_size <= inject_size)
 				ret = ft_inject(ep, remote_fi_addr,
 						opts.transfer_size);
@@ -119,7 +118,7 @@ static int pingpong_pre_posted_rx(size_t inject_size)
 	} else {
 		for (i = 0; i < opts.iterations + opts.warmup_iterations; i++) {
 			if (i == opts.warmup_iterations)
-				ft_start();
+				ft_start(timer);
 
 			ret = ft_rx(ep, opts.transfer_size);
 			if (ret)
@@ -135,20 +134,20 @@ static int pingpong_pre_posted_rx(size_t inject_size)
 				return ret;
 		}
 	}
-	ft_stop();
+	ft_stop(timer);
 
 	return FI_SUCCESS;
 }
 
 /* Pingpong latency test without pre-posted receive buffers. */
-static int pingpong_no_pre_posted_rx(size_t inject_size)
+static int pingpong_no_pre_posted_rx(size_t inject_size, union ft_timer *timer)
 {
 	int ret, i;
 
 	if (opts.dst_addr) {
 		for (i = 0; i < opts.iterations + opts.warmup_iterations; i++) {
 			if (i == opts.warmup_iterations)
-				ft_start();
+				ft_start(timer);
 
 			if (opts.transfer_size <= inject_size)
 				ret = ft_inject(ep, remote_fi_addr,
@@ -170,7 +169,7 @@ static int pingpong_no_pre_posted_rx(size_t inject_size)
 	} else {
 		for (i = 0; i < opts.iterations + opts.warmup_iterations; i++) {
 			if (i == opts.warmup_iterations)
-				ft_start();
+				ft_start(timer);
 
 			ret = ft_post_rx(ep, opts.transfer_size, &rx_ctx);
 			if (ret)
@@ -197,7 +196,7 @@ static int pingpong_no_pre_posted_rx(size_t inject_size)
 				return ret;
 		}
 	}
-	ft_stop();
+	ft_stop(timer);
 
 	return FI_SUCCESS;
 }
@@ -205,6 +204,7 @@ static int pingpong_no_pre_posted_rx(size_t inject_size)
 int pingpong(void)
 {
 	int ret;
+	union ft_timer timer = {};
 	size_t inject_size = fi->tx_attr->inject_size;
 
 	ret = fi_getopt(&ep->fid, FI_OPT_ENDPOINT, FI_OPT_INJECT_MSG_SIZE,
@@ -236,7 +236,7 @@ int pingpong(void)
 				return ret;
 		}
 
-		ret = pingpong_no_pre_posted_rx(inject_size);
+		ret = pingpong_no_pre_posted_rx(inject_size, &timer);
 		if (ret)
 			return ret;
 	} else {
@@ -244,16 +244,16 @@ int pingpong(void)
 		if (ret)
 			return ret;
 
-		ret = pingpong_pre_posted_rx(inject_size);
+		ret = pingpong_pre_posted_rx(inject_size, &timer);
 		if (ret)
 			return ret;
 	}
 
 	if (opts.machr)
-		show_perf_mr(opts.transfer_size, opts.iterations, &start, &end, 2,
+		show_perf_mr(opts.transfer_size, opts.iterations, timer, 2,
 				opts.argc, opts.argv);
 	else
-		show_perf(NULL, opts.transfer_size, opts.iterations, &start, &end, 2);
+		show_perf(NULL, opts.transfer_size, opts.iterations, timer, 2);
 
 	return 0;
 }
@@ -296,6 +296,7 @@ int run_pingpong(void)
 int pingpong_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 {
 	int ret, i;
+	union ft_timer timer = {};
 	size_t inject_size = fi->tx_attr->inject_size;
 
 	ret = fi_getopt(&ep->fid, FI_OPT_ENDPOINT, FI_OPT_INJECT_RMA_SIZE,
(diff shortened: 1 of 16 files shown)


fabtests/include/shared.h
```diff
@@ -520,14 +525,43 @@ static inline uint64_t ft_gettime_ms(void)
...
+	timer->start = ft_gettime_ns();
+}
+static inline void ft_timer_pause(union ft_timer *timer)
+{
+	timer->elapsed = ft_gettime_ns() - timer->start;
+}
+static inline void ft_timer_resume(union ft_timer *timer)
+{
+	timer->start = ft_gettime_ns() - timer->elapsed;
+}
```
aingerson, line 539: Can we simplify this to just a start and stop where stop is
elapsed = gettime - start

and start is
start = gettime - elapsed

and elapsed is initialized to 0?


or else you could consider a model where
start is 
start = gettime

and stop is
elapsed += gettime - start

either way, I don't think you need the pause/resume to be separate. I would rather keep it as simple as possible

fabtests/include/shared.h
```diff
@@ -520,14 +525,43 @@ static inline uint64_t ft_gettime_ms(void)
...
+	ft_timer_pause(timer);
+}
+static inline bool ft_timer_is_elapsed(union ft_timer timer, uint64_t timeout, enum precision p)
+{
+	// assuming timer is running
+	return ft_gettime_ns() - timer.start >= timeout * p;
+}
+static inline uint64_t ft_timer_get_elapsed(union ft_timer timer, enum precision p)
+{
+	// assiming timer is stopped
```
aingerson, line 551: typo

fabtests/include/shared.h
```diff
@@ -520,14 +525,43 @@ static inline uint64_t ft_gettime_ms(void)
...
+{
+	timer->start = ft_gettime_ns() - timer->elapsed;
+}
+static inline void ft_timer_stop(union ft_timer *timer)
+{
+	ft_timer_pause(timer);
+}
+static inline bool ft_timer_is_elapsed(union ft_timer timer, uint64_t timeout, enum precision p)
+{
+	// assuming timer is running
```
aingerson, line 546: typo, also couldn't hurt to add some asserts in the timer instead of having to comment this

fabtests/ubertest/cq.c
```diff
@@ -291,26 +291,18 @@ static inline size_t ft_update_credits(void *buf,
...
-				start = 1;
-			}
-
 			ft_cq_read(fi_cq_read, cq, buf, ft_x->cq_format, ft_x->credits,
 				   completions, x_str, ret, verify);
-
-			clock_gettime(CLOCK_MONOTONIC, &e);
-			poll_time = get_elapsed(&s, &e, MILLI);
-		} while (ret == -FI_EAGAIN && poll_time < timeout &&
+		} while (ret == -FI_EAGAIN && !ft_timer_is_elapsed(timer, timeout, MILLI) &&
```
aingerson, line 305: I think this is going to break the timeout for calls to this before the test as begun/when the timer is not running. For example, if we call this function during initialization before the test has begun/started timing, this is going to return (time - 0 >= timeout) which is going to probably break too early. So I think you have to keep this with the poll time updated for this loop, not globabally

fabtests/include/shared.h
```diff
@@ -520,14 +525,43 @@ static inline uint64_t ft_gettime_ms(void)
 	return ft_gettime_ns() / 1000000;
 }
 
-static inline void ft_start(void)
+static inline void ft_timer_start(union ft_timer *timer)
```
aingerson, line 528: Is there a specific reason you changed this to pass in a timer instead of using a global? I'm not necessarily against it if you have a reason you want to update it like if you have a test you want to add that has multiple timers. It's just not necessary right now and makes for a lot of code changes without a clear reason.

aingerson (commented):
Fun changes! Thank you!

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -461,14 +463,19 @@ int rma_tx_completion(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
...
 	ret = ft_get_tx_comp(tx_seq);
 	if (ret)
 		return ret;
-	return ft_rx(ep, FT_RMA_SYNC_MSG_BYTES);
+	if (timer)
+		ft_timer_pause(timer);
+	ret = ft_rx(ep, FT_RMA_SYNC_MSG_BYTES);
+	if (timer)
+		ft_timer_resume(timer);
+	return ret;
```
aingerson, line 478: Could this give you false performance numbers? Depending on your completion level, the send may have not been completely sent so you could be reporting higher bandwidth. Same on the receive side for the RMA ops. The RMA ops aren't going to generate completions so if you stop the timer right after get_rx_comp you're not actually waiting for all the messages to come in. Unfortunately, for RMA ops you kind of have to wait for the sync message to ensure you're stopping the timer when all the data has arrived.

alekswn: Makes sense. The rationale here was to exclude server -> client send latency from time calculation. For writedata benchmark we observed synchronization overhead of the same order of magnitude as overall timing due to connection setup delay for EFA. 

Connection setup delay is only seen for a first few synchronizations, but we see 30% error in bandwidth measurements attributed to connection setup latency. The delay is negligible after first 5 synchronizations.

The reason the synchronization delay affects writedata test so much is asymmetry of the test. Client makes a lot of writes in the loop, but server replies with just a single send at the end of each window. We are warming up client->server path with the warm-up iterations, but do not warm up server->client connection.

If we want keep synchronization timing included, I think an alternative approach would be to warm up server->client path as well. Does it sound like a better approach?

fabtests/include/shared.h
alekswn: Yeah, we can have just  two state transition functions (`start` and `stop`). Do we want some initialization function to set elapsed to zero or should let caller to just zero initialize the timer ?
The later is not not going to work if we returned to global variable model, so the caller would have to reset the global  timer before calling `start`.


Would not it be more convenient to reset timer in `start()` and still have `resume()` function which would just update `elapsed`? 

In that case callers which do not need pause functionality can just call `start()` and `stop()` as before.

fabtests/include/shared.h
alekswn: I initially implemented `ft_timer` as `struct` and added asserts here, but converted it to a `union` then.
There is no way to add asserts having it as a union. Probably that was a premature optimization and a bad idea.
I'll revert it back.

fabtests/ubertest/cq.c
alekswn: `timer` is a local variable here, which is declared and initialized in this function a few lines above, so should not be a problem.

fabtests/benchmarks/benchmark_shared.c
aingerson: Gotcha, thanks for the context. I think it's ok to warmup both sides. In my opinion, it's a little weird that the writedata function has such a different model than the other functions. I understand why (because of the rx completion) but I think it would be ok to have all ops behave in the same pattern. Can we change writedata to be two sided? Would that help as well? Or are you explicitly looking to have it remain a one-sided test?

fabtests/include/shared.h
aingerson: Ah I missed the union! A fun optimization! But I think I would prefer the struct, especially if that adds the option of an assert to catch misuses of it. I wouldn't want performance numbers to get reported incorrectly if someone forgot to stop/resume in the correct order or something.

fabtests/benchmarks/benchmark_shared.c
alekswn: Thanks for suggestion!
Will try to make it symmetric.

fabtests/include/shared.h
aingerson: I'm ok with start/stop/resume with an init in start

fabtests/include/shared.h
alekswn: The reason for that is to make it useful for cases like timeout in your previous comment. There are few places like that in fabtests.
Also, I have noticed global vars `s` and `e` are shadowed at lots of places across fabtests code. `show_perf()` function does function takes `s` and `e` as arguments and does not use globals. So, in my opinion, keeping it global might lead to tricky bugs in the future. 
I'm fixing a similar shadowing bug with `targs` variable  in the present PR [rdm_bw_mt.c:444](https://github.com/ofiwg/libfabric/pull/11507/commits/ebe3a5f1496ef3f5403f76d3f1961a9257085d03#diff-c94742681fb243fe054e6ba9f2fb5b94325f07c1472bf839013551d96b8345af).

aingerson: Sounds reasonable. Thanks!

alekswn:
> Fun changes! Thank you!

Thanks for reviewing this PR and detailed feedback, Alexia!

If we don't want exclude synchronization time from `bandwidth_rma` timing introducing a new timer is not making much sense. 
This PR also fixes a couple of timer-related bugs in `efa_gda.c` and `multy_ep_mt.c`. I think I'll cut separate PRs with fixes using the old timer.

PR #11508, opened 2025-10-15
prov/efa: Set compiler and linker flags for --enable-efa

alekswn:
This commit eliminates need to set environment variables manually at build-time and runtime then the path had been specified with ./configure key --enable-efa.

When you run ./configure --enable-efa=/path/to/rdma-core, the build system now:

- Finds headers in /path/to/rdma-core/include
- Finds libraries in /path/to/rdma-core/lib or /path/to/rdma-core/lib64
- Adds -L/path/to/rdma-core/lib for compile-time linking Automatically adds -Wl,-rpath,/path/to/rdma-core/lib for runtime linking

### DIFF
--- a/prov/efa/configure.m4
+++ b/prov/efa/configure.m4
@@ -188,13 +188,13 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 			[have_efadv_sl=1],
 			[have_efadv_sl=0],
 			[[#include <infiniband/efadv.h>]])
-		
+
 		have_efadv_query_qp_wqs=1
 		AC_CHECK_DECL([efadv_query_qp_wqs],
 			[],
 			[have_efadv_query_qp_wqs=0],
 			[[#include <infiniband/efadv.h>]])
-	
+
 		have_efadv_query_cq=1
 		AC_CHECK_DECL([efadv_query_cq],
 			[],
@@ -281,6 +281,13 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	efa_CPPFLAGS="$efa_ibverbs_CPPFLAGS $efadv_CPPFLAGS"
 	efa_LDFLAGS="$efa_ibverbs_LDFLAGS $efadv_LDFLAGS"
 	efa_LIBS="$efa_ibverbs_LIBS $efadv_LIBS"
+
+	dnl Add RPATH for custom efa_LIBDIR if specified
+	AS_IF([test x"$efa_LIBDIR" != x""],
+	[
+		AC_MSG_NOTICE([Adding RUNPATH for EFA libraries: $efa_LIBDIR])
+		efa_LDFLAGS+=" -L$efa_LIBDIR -Wl,--enable-new-dtags,-rpath,$efa_LIBDIR"
+	])
 	cmocka_rpath=""
 	AC_ARG_ENABLE([efa-unit-test],
 		[


a-szegel:
Can we rename this to --with-rdma-core=<path>

shijin-aws:
I am not an autotool wizard but I think we should avoid using RPATH as it will make the dependency not customizable

alekswn:
> Can we rename this to --with-rdma-core=

Will do

>  I am not an autotool wizard but I think we should avoid using RPATH as it will make the dependency not customizable

I've checked RPATH is not set in the binary, but RUNPATH is set
```
nalexey@p5en-odcr-queue-dy-p5en48xlarge-13 ~/libfabric/fabtests/install/bin
 % objdump -p fi_rma_pingpong | grep -E "(RPATH|RUNPATH)"
  RUNPATH              /home/nalexey/libfabric/install/lib
```
 I think all modern linkers resolve `-rpath` argument to `RUNPATH` variable in the binary file. 
 
RUNPATH is used after LD_LIBARY_PATH but before system libraries. So one can still customize rdma-core library using LD_LIBARY_PATH. It's just going to use rdma-core library path provided on compile time by default if neither LD_PRELOAD no LD_LIBARY_PATH env variables were set. 

Is not that exactly what we want compiling libfabric with a custom rdma-core?

darrylabbate:
Yeah I'd be concerned about `RPATH`'s precedence over `LD_LIBRARY_PATH`. If we have a test that ensures expected behavior wrt hot-swapping with `LD_LIBRARY_PATH`, this change should be ok.

New dynamic tags (`LDFLAGS+=--enable-new-dtags`) should also work since it sets `RUNPATH` instead of `RPATH`, if the linker supports it.

> I think all modern linkers resolve -rpath argument to RUNPATH variable in the binary file.

We'd need to confirm this is true on all platforms we support

alekswn:
I will add `LDFLAGS+=--enable-new-dtags` then, seems to be the safest bet

a-szegel:
@alekswn showed me that this is standard for multiple providers https://github.com/ofiwg/libfabric/blob/main/.github/workflows/pr-ci.yml#L28C16-L38 ... ignore my comment about --with-rdma-core=

alekswn:
bot:aws:retest

prov/efa/configure.m4
```diff
@@ -281,6 +281,13 @@ AC_DEFUN([FI_EFA_CONFIGURE],[
 	efa_CPPFLAGS="$efa_ibverbs_CPPFLAGS $efadv_CPPFLAGS"
 	efa_LDFLAGS="$efa_ibverbs_LDFLAGS $efadv_LDFLAGS"
 	efa_LIBS="$efa_ibverbs_LIBS $efadv_LIBS"
+
+	dnl Add RPATH for custom efa_LIBDIR if specified
```
shijin-aws, line 285: where is `efa_LIBDIR` specified?

alekswn: It's output of FI_CHECK_PACKGE:
https://github.com/ofiwg/libfabric/blob/f38de7d70d930eb9b275bdf1d3ab170aba8b6615/config/fi_check_package.m4#L140-L215. 

Assigned here:
https://github.com/ofiwg/libfabric/blob/main/prov/efa/configure.m4#L54

shijin-aws (approved):

PR #11515, opened 2025-10-16
prov/efa: Fix data path direct for FI_MORE

shijin-aws:
The device queue can only stage upt to max_batch cnt of wqe
before ringging the doorbell. When applications post more
than max_batch cnt of WQEs with FI_MORE, it should flush
the device queue and ring the doorbell before continue constructing new WQEs

### DIFF
--- a/fabtests/pytest/efa/test_rma_bw.py
+++ b/fabtests/pytest/efa/test_rma_bw.py
@@ -71,12 +71,16 @@ def test_rma_bw_1G(cmdline_args, operation_type, rma_bw_completion_semantic):
 
 @pytest.mark.functional
 @pytest.mark.parametrize("operation_type", ["writedata", "write"])
-def test_rma_bw_use_fi_more(cmdline_args, operation_type, rma_bw_completion_semantic, inject_message_size, direct_rma_size, rma_fabric):
+@pytest.mark.parametrize("iteration_type",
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_data_path_direct_internal.h
```diff
@@ -599,7 +600,9 @@ efa_data_path_direct_send_wr_common(struct efa_qp *qp,
 
 	/* MODIFIED: if any are pending, copy out the previous one first: */
 	if (sq->num_wqe_pending) {
-		efa_data_path_direct_send_wr_post_working(sq, false);
+		/* when reaching the sq max_batch, ring the db */
+		ring_db = sq->num_wqe_pending == sq->wq.max_batch ? true : false;
+		efa_data_path_direct_send_wr_post_working(sq, ring_db);
```
jiaxiyan: why not just `sq->num_wqe_pending == sq->wq.max_batch` here? you don't need `? true : false;`

shijin-aws: Thank you, fixed.

jiaxiyan (dismissed):

shijin-aws:
bot:aws:retest

sunkuamzn (approved):

PR #11516, opened 2025-10-16
fabtests/efa: Add AV removal to multi_ep_mt test 

sunkuamzn:
fabtests/efa: Add AV removal to multi_ep_mt test
Add AV removal to the client threads with random
sleeps before and after

fabtests: Add ft_av_remove

### DIFF
--- a/fabtests/prov/efa/src/multi_ep_mt.c
+++ b/fabtests/prov/efa/src/multi_ep_mt.c
@@ -197,21 +197,30 @@ static int get_one_comp(struct fid_cq *cq)
 	return FI_SUCCESS;
 }
 
+static inline void random_sleep()
+{
+	int sleep_time;
+	struct timespec ts = {0};
+
+	sleep_time = (rand() % (RANDOM_MAX - RANDOM_MIN + 1)) + RANDOM_MIN;
+	ts.tv_nsec = sleep_time;
+	nanosleep(&ts, NULL);
+}
+
 static void *post_sends(void *context)
 {
 	int idx, ret, i, j;
 	size_t len;
-	// the range of the sleep time (in nanoseconds)
-	int sleep_time;
-	struct timespec ts = {0};
 	int num_transient_eps = 10;
+	int av_idx = 0;
+
+	idx = ((struct thread_context *) context)->idx;
+	av_idx = shared_av ? 0 : idx;
 
 	srand(time(NULL));
-	sleep_time = (rand() % (RANDOM_MAX - RANDOM_MIN + 1)) + RANDOM_MIN;
-	idx = ((struct thread_context *) context)->idx;
-	ts.tv_nsec = sleep_time;
 
-	nanosleep(&ts, NULL);
+	random_sleep();
+
 	len = opts.transfer_size;
 	for (j = 0; j < num_transient_eps; j++) {
 		printf("Thread %d: opening client \n", idx);
@@ -230,9 +239,15 @@ static void *post_sends(void *context)
 			}
 		}
 
-		sleep_time = (rand() % (RANDOM_MAX - RANDOM_MIN + 1)) + RANDOM_MIN;
-		ts.tv_nsec = sleep_time;
-		nanosleep(&ts, NULL);
+		random_sleep();
+
+		ret = fi_av_remove(avs[av_idx], &remote_fiaddr[idx], 1, 0);
+		if (ret) {
+			FT_PRINTERR("fi_av_remove", ret);
+			return NULL;
+		}
+
+		random_sleep();
 
 		// exit
 		printf("Thread %d: closing client\n", idx);


shijin-aws (dismissed):

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -230,9 +238,17 @@ static void *post_sends(void *context)
...
 		}
 
-		sleep_time = (rand() % (RANDOM_MAX - RANDOM_MIN + 1)) + RANDOM_MIN;
-		ts.tv_nsec = sleep_time;
-		nanosleep(&ts, NULL);
+		random_sleep();
+
+		if (shared_av) {
+			av_idx = 0;
+			ret = ft_av_remove(avs[av_idx], 1, &remote_fiaddr[idx], 1, 0);
```
shijin-aws:
```
nt ft_av_remove(struct fid_av *av, size_t count, fi_addr_t *fi_addr,
		uint64_t flags, void *context)
```

what does 1 means for `flags` ? context should be `NULL` maybe

shijin-aws: I think this is using `fi_av_remove`'s signature instead of `ft_av_remove`'s

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -230,9 +238,16 @@ static void *post_sends(void *context)
...
-		ts.tv_nsec = sleep_time;
-		nanosleep(&ts, NULL);
+		random_sleep();
+
+		if (shared_av)
+			av_idx = 0;
+		else
+			av_idx = idx;
+
+		ret = ft_av_remove(avs[av_idx], 1, &remote_fiaddr[idx], 1, 0);
```
aingerson: I prefer
```
av_idx = shared_av ? 0 : idx;
```
but I don't super care about it.

I also don't think the ft_av_remove call really needs to be in the common code because you're just adding it in one place and it's basically just a pass through of the OFI call. I don't have a strong opinion about it if you have a vision for using it in the future though.
Also you're not checking the return code here so coverity will probably complain

j-xiong: Agree.

I have a stronger preference that the `ft_av_remove` call should not be added.  We should call `fi_av_remove` directly and check the return result and print error message accordingly. See the `fi_av_open` call in the same file for an example.

fabtests/prov/efa/src/multi_ep_mt.c
sunkuamzn: Sometimes, my thinking is faster than my typing..

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -230,9 +238,17 @@ static void *post_sends(void *context)
 			}
 		}
 
-		sleep_time = (rand() % (RANDOM_MAX - RANDOM_MIN + 1)) + RANDOM_MIN;
-		ts.tv_nsec = sleep_time;
-		nanosleep(&ts, NULL);
+		random_sleep();
+
+		av_idx = shared_av ? 0 : idx;
```
aingerson: Sorry one more that I missed - should this be j instead of idx? and then same thing with the close_client call at the end? I'm not super familiar with this test so I'm not positive. It's just weird that we're setting the av_idx every iteration when idx isn't changing in this loop.
And then in the shared AV case would you be removing the same address every time? Should line 245 also be j?

sunkuamzn: you're right, I can move the assignment outside the loop

fabtests/prov/efa/src/multi_ep_mt.c
```diff
@@ -230,9 +239,15 @@ static void *post_sends(void *context)
 			}
 		}
 
-		sleep_time = (rand() % (RANDOM_MAX - RANDOM_MIN + 1)) + RANDOM_MIN;
-		ts.tv_nsec = sleep_time;
-		nanosleep(&ts, NULL);
+		random_sleep();
+
+		ret = fi_av_remove(avs[av_idx], &remote_fiaddr[idx], 1, 0);
```
aingerson, line 244: Should the remote fi addr index here still be j though? Or maybe on shared AV you break instead of iterate? On the shared AV case wouldn't this result in removing the same address num_transient_eps times?
Sorry for the questions, like I said I'm not familiar with the format of the test

sunkuamzn: That is OK because `open_client` further up inserts the peer into the AV for each iteration of the loop https://github.com/ofiwg/libfabric/blob/ff97cc7efcfc303185b8bcb782eb8e3e79be3f2b/fabtests/prov/efa/src/multi_ep_mt.c#L400

The test has one server thread and many client threads. Each client thread creates a new endpoint, AV/AV entry, sends a bunch of messages to the server, closes the endpoint and AV/AV entry. The number of times the client thread repeats this whole process is `num_transient_eps`.

aingerson (approved):
@sunkuamzn Thank you for the explanation!

PR #8563, opened 2023-02-23
prov/efa: Make shm_info created on the fly during domain open 

shijin-aws:
Currently, shm info is initialized as a global variable `g_shm_info` once. When applications call `fi_getinfo()` multiple times, `g_shm_info` is only created by the first `fi_getinfo()` and won't be changed by subsequent `fi_getinfo()` calls.
    
This PR addresses this issue by making shm_info created on the fly during efa domain open, based on application info created earlier. A unit test is added to validate the change.

PR #11517, opened 2025-10-16
prov/efa: Create shm info on demand for efa_rdm_ep_open

jiaxiyan:
Application can create endpoints with the same domain but different info. Remove shm_info from efa_domain since we shouldn't reuse it in endpoint. Use the info passed in the fi_endpoint and create ep's own shm info.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -116,24 +116,26 @@ static int efa_domain_init_qp_table(struct efa_domain *efa_domain)
 
 static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *info)
 {
+	struct fi_info *shm_info = NULL;
 	int err;
 
 	assert(EFA_INFO_TYPE_IS_RDM(info));
 
-	efa_domain->shm_info = NULL;
-	efa_shm_info_create(info, &efa_domain->shm_info);
-	if (efa_domain->shm_info) {
-		err = fi_fabric(efa_domain->shm_info->fabric_attr,
+	efa_shm_info_create(info, &shm_info);
+	if (shm_info && !efa_domain->fabric->shm_fabric) {
+		err = fi_fabric(shm_info->fabric_attr,
 				&efa_domain->fabric->shm_fabric,
 				efa_domain->fabric->util_fabric.fabric_fid.fid.context);
-		if (err)
+		if (err) {
+			EFA_WARN(FI_LOG_DOMAIN, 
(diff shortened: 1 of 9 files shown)


prov/efa/src/efa_domain.c
```diff
@@ -136,14 +137,13 @@ static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *in
...
 	if (enable_shm)
-		efa_shm_info_create(info, &efa_domain->shm_info);
+		efa_shm_info_create(info, &shm_info);
 	else
 		EFA_INFO(FI_LOG_CORE, "EFA will not use SHM for intranode communication because FI_EFA_ENABLE_SHM_TRANSFER=0\n");
 
-	if (efa_domain->shm_info) {
-		err = fi_fabric(efa_domain->shm_info->fabric_attr,
+	if (shm_info) {
+		err = fi_fabric(shm_info->fabric_attr,
```
shijin-aws, line 126: I think now we should create shm fabric inside efa_fabric() call for consistency

prov/efa/src/efa_domain.c
```diff
@@ -136,14 +137,13 @@ static int efa_domain_init_rdm(struct efa_domain *efa_domain, struct fi_info *in
 		enable_shm = false;
 	}
 
-	efa_domain->shm_info = NULL;
 	if (enable_shm)
-		efa_shm_info_create(info, &efa_domain->shm_info);
+		efa_shm_info_create(info, &shm_info);
```
shijin-aws: We should move the enable_shm logic inside `efa_shm_info_create`, so we can have such check everywhere

prov/efa/src/efa_domain.c
jiaxiyan: I read the comment in https://github.com/ofiwg/libfabric/pull/8563 and you said

> We want shm info to be created on fly in domain open, if shm fi_getinfo() returns non-zero, we should not open fabric for shm. I am not sure how we can handle this logic in efa_fabric()? fi_fabric() does not have an argument for info though.

shijin-aws: I see, good catch. Then we should detect whether there is already shm_fabric created in efa_fabric, if we create multiple domains in a single fabric, this is creating stale fabric resources for shm

prov/efa/src/efa_domain.c
jiaxiyan: good idea

shijin-aws:
bot:aws:retest

PR #11518, opened 2025-10-17
prov/efa: Fix deadlocks in AV insert/remove/close and CQ read paths

sunkuamzn:
Acquring locks in different order in different code paths can cause
deadlocks. The AV insert/remove/close paths and the CQ read paths need
the SRX lock and the util AV locks.

The CQ read path always acquires the SRX lock but only acquires the util
AV lock when inserting into the implicit AV. So the AV
insert/remove/close paths must also acquire the locks in the same order.


sunkuamzn:
bot:aws:retest

sunkuamzn:
CI failed on TCP provider which this PR does not touch

PR #11533, opened 2025-10-21
prov/cxi: add FI_EBUSY debug messages

SeanP-2023:
Add debug messages when we close objects and
detect ref counts

### DIFF
--- a/prov/cxi/src/cxip_ep.c
+++ b/prov/cxi/src/cxip_ep.c
@@ -800,7 +800,7 @@ int cxip_free_endpoint(struct cxip_ep *ep)
 	 */
 	count = ofi_atomic_get32(&ep_obj->ref);
 	if (count) {
-		CXIP_WARN("EP refcount non-zero: %d\n", count);
+		CXIP_WARN("EP refcount non-zero:%d\n", count);
 		return -FI_EBUSY;
 	}
 
--- a/prov/cxi/src/cxip_eq.c
+++ b/prov/cxi/src/cxip_eq.c
@@ -30,16 +30,20 @@
 #include "cxip.h"
 
 #define CXIP_WARN(...) _CXIP_WARN(FI_LOG_EQ, __VA_ARGS__)
+#define CXIP_DBG(...) _CXIP_DBG(FI_LOG_EQ, __VA_ARGS__)
 
 static int cxip_eq_close(struct fid *fid)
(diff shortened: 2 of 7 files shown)


prov/cxi/src/cxip_eq.c
```diff
@@ -30,16 +30,20 @@
 #include "cxip.h"
 
 #define CXIP_WARN(...) _CXIP_WARN(FI_LOG_EQ, __VA_ARGS__)
+#define CXIP_DBG(...) _CXIP_DBG(FI_LOG_CQ, __VA_ARGS__)
```
swelch: I suppose this should probably be a FI_LOG_EQ instead of FI_LOG_CQ.

prov/cxi/src/cxip_fabric.c
```diff
@@ -19,6 +19,7 @@
 
 #include "cxip.h"
 
+#define CXIP_DBG(...) _CXIP_DBG(FI_LOG_CQ, __VA_ARGS__)
```
swelch: Probably something other than FI_LOG_CQ here, you can use FI_LOG_FABRIC if you want.

swelch (changes requested):

prov/cxi/src/cxip_ep.c
```diff
@@ -800,7 +800,8 @@ int cxip_free_endpoint(struct cxip_ep *ep)
 	 */
 	count = ofi_atomic_get32(&ep_obj->ref);
 	if (count) {
-		CXIP_WARN("EP refcount non-zero: %d\n", count);
+		CXIP_WARN("EP refcount non-zero:%d\n", count);
+		CXIP_DBG("EP refcount non-zero:%d returning FI_EBUSY\n", count);
```
swelch: If you have the WARN you don't need the DBG, you can delete this line or delete the WARN if you prefer.

SeanP-2023: Good catch done

prov/cxi/src/cxip_eq.c
SeanP-2023: done

prov/cxi/src/cxip_fabric.c
SeanP-2023: done

PR #11535, opened 2025-10-21
prov/efa: Cqe warn messages

nmazzilli3:
prov/efa: Log warnings only for internal OPE failures or if CQ error entry not written

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -648,7 +648,7 @@ void efa_rdm_rxe_handle_error(struct efa_rdm_ope *rxe, int err, int prov_errno)
 		err_entry.err_data = err_msg;
 	}
 
-	EFA_WARN(FI_LOG_CQ, "err: %d, message: %s (%d)\n",
+	EFA_INFO(FI_LOG_CQ, "err: %d, message: %s (%d)\n",
 		err_entry.err,
 		err_entry.err_data
 			? (const char *) err_entry.err_data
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -663,6 +663,14 @@ void efa_rdm_rxe_handle_error(struct efa_rdm_ope *rxe, int err, int prov_errno)
...
 
 	if (rxe->internal_flags & EFA_RDM_OPE_INTERNAL) {
+		EFA_WARN(FI_LOG_CQ, "EFA provider internal rxe failure err: %d, message: %s (%d)\n",
+			err_entry.err,
+			err_entry.err_data
+				? (const char *) err_entry.err_data
+				: efa_strerror(err_entry.prov_errno),
+		 	err_entry.prov_errno);
+		efa_show_help(err_entry.prov_errno);
+
```
darrylabbate, line 673: This exact code fragment is already duplicated all over:
- https://github.com/ofiwg/libfabric/blob/cd6be4c64b9ece3b30e02cd9a30cecbdf6f9337e/prov/efa/src/efa_cq.c#L125-L132
- https://github.com/ofiwg/libfabric/blob/cd6be4c64b9ece3b30e02cd9a30cecbdf6f9337e/prov/efa/src/rdm/efa_rdm_ope.c#L651-L657

Can we wrap this in a reusable utility function and replace existing instances as well?

Also we can remove the `"EFA provider internal failure ..."` prefix in the error string since the log provider already implies it's an EFA error

nmazzilli3: Making a utility to replace this duplicate code makes sense to me.

nmazzilli3: I believe it's important to keep the operation type identifier (rxe/txe) in the message for debugging clarity, log filtering and consistency.

nmazzilli3: Per discussion with @darrylabbate we will backlog an item to optimize a util function for log messages at a later date.

sunkuamzn (approved):

PR #11536, opened 2025-10-22
prov/efa: Evict from implicit AV when AH limit is reached 

sunkuamzn:
This PR implements logic to evict AH entries when the device AH
limit is reached. It iterates over the AH LRU list and evicts
the first AH entry that only has implicit AV entries.


prov/efa/src/efa_domain.c
```diff
@@ -181,7 +181,6 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	if (!efa_domain)
 		return -FI_ENOMEM;
 
-	dlist_init(&efa_domain->list_entry);
```
jiaxiyan: Why is this not needed?

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -165,7 +167,7 @@ struct efa_rdm_peer *efa_rdm_ep_get_peer_implicit(struct efa_rdm_ep *ep, fi_addr
 out:
 	assert(peer);
 	/* Move to the front of the LRU list */
```
jiaxiyan, line 169: Move to the end?

prov/efa/src/efa_conn.c
```diff
@@ -0,0 +1,468 @@
...
+ * @param[in]	release_from_implicit_av		whether to release conn
+ * from implicit AV
+ * @param[in]	grab_srx_lock		whether to get the SRX lock before
+ * destroying the peer struct
+ */
+void efa_conn_release(struct efa_av *av, struct efa_conn *conn,
+		      bool release_from_implicit_av)
+{
+	if (av->domain->info_type == EFA_INFO_RDM)
+		assert(ofi_genlock_held(&av->domain->srx_lock));
```
darrylabbate:
```suggestion
	assert(av->domain->info_type != EFA_INFO_RDM ||
		ofi_genlock_held(&av->domain->srx_lock));
```

prov/efa/src/efa_conn.c
```diff
@@ -0,0 +1,468 @@
...
+ * @param[in]	release_from_implicit_av		whether to release conn
+ * from implicit AV
+ * @param[in]	grab_srx_lock		whether to get the SRX lock before
+ * destroying the peer struct
+ */
+void efa_conn_release_ah_unsafe(struct efa_av *av, struct efa_conn *conn,
+				bool release_from_implicit_av)
+{
+	if (av->domain->info_type == EFA_INFO_RDM)
+		assert(ofi_genlock_held(&av->domain->srx_lock));
```
darrylabbate:
```suggestion
	assert(av->domain->info_type != EFA_INFO_RDM ||
		ofi_genlock_held(&av->domain->srx_lock));
```

prov/efa/src/efa_conn.h
```diff
@@ -0,0 +1,45 @@
...
+				 bool release_from_implicit_av);
+
+void efa_conn_release_util_av(struct efa_av *av, struct efa_conn *conn,
+			      bool release_from_implicit_av);
+
+void efa_conn_release(struct efa_av *av, struct efa_conn *conn,
+		      bool release_from_implicit_av);
+
+void efa_conn_release_ah_unsafe(struct efa_av *av, struct efa_conn *conn,
+				bool release_from_implicit_av);
```
darrylabbate, line 37: Duplicated declarations

prov/efa/src/efa_conn.c
```diff
@@ -0,0 +1,468 @@
...
+ *
+ * @param[in]	av	address vector
+ * @param[in]	conn	efa_conn object pointer
+ * @param[in]	release_from_implicit_av		whether to release conn
+ * from implicit AV
+ * @param[in]	grab_srx_lock		whether to get the SRX lock before
+ * destroying the peer struct
+ */
+void efa_conn_release(struct efa_av *av, struct efa_conn *conn,
+		      bool release_from_implicit_av)
```
darrylabbate, line 414: I'm just now noticing the boolean parameter for switching between implicit/explicit AV here and in a few other functions. AFAICT, all callers to these are hardcoding (or passing a param from a caller who hardcoded) their implicit/explicit boolean argument.

I know these weren't implemented for this PR, but they need to eventually be refactored into implicit/explicit-specific variants. Some of them are relatively small functions and might currently satisfy compiler heuristics for inlining, but it's not guaranteed, and we can't rely on the unused branches being optimized away.

prov/efa/src/efa_ah.h
```diff
@@ -0,0 +1,42 @@
...
+			    bool insert_implicit_av);
+
+void efa_ah_release(struct efa_domain *domain, struct efa_ah *ah,
+		    bool release_from_implicit_av);
+
+void efa_ah_release_unsafe(struct efa_domain *domain, struct efa_ah *ah,
+			   bool release_from_implicit_av);
+
+void efa_ah_release(struct efa_domain *domain, struct efa_ah *ah,
+		    bool release_from_implicit_av);
```
darrylabbate: Duplicate declaration

prov/efa/src/efa_conn.c
sunkuamzn: I've made sure to avoid branching in the hot path. The conditionals are only in the control path (AV insertion/removal) and the implicit AV insertion path during CQ read.

I can optimize the remaining branches in a future PR

prov/efa/src/rdm/efa_rdm_ep_utils.c
sunkuamzn: It depends on what you call front or back. This entry was just used, so I'm thinking of it as the front of the list and the last to get evicted.

jiaxiyan: > * @brief Move the conn to the end of the LRU list to indicate that it is the most recently used entry

just keep it consistent with your function comment

sunkuamzn: fixed, thanks!

prov/efa/src/efa_conn.h
sunkuamzn: fixed, thanks!

prov/efa/src/efa_ah.h
sunkuamzn: fixed, thanks!

prov/efa/src/efa_conn.c
sunkuamzn: fixed, thanks!

prov/efa/src/efa_conn.c
sunkuamzn: fixed, thanks!

darrylabbate (approved):

PR #11537, opened 2025-10-22
prov/efa: Remove efa_conn_release unsafe

sunkuamzn:
Commit 8a244dc5d9b5b09ed0f89bbacb9d37e4af78d524 acquires the SRX lock in the AV removal path. So efa_conn_release already expects the SRX lock to be held and efa_conn_release_unsafe is no longer necessary


sunkuamzn:
Included in https://github.com/ofiwg/libfabric/pull/11536

PR #11538, opened 2025-10-22
fabtests/efa: BUG: efa_gda: recv buffer verification is skipped silently

alekswn:
Problem:

efa_gda fabtest had a logical bug. When verifying rx_buf contents, we checked if both `FT_OPT_VERIFY_DATA` and `FT_OPT_ACTIVE` options were set. However, `FT_OPT_ACTIVE` is unset by `ft_stop()`, which is called in the main loop to check for timeout.

Fix:

Check only for `FT_OPT_VERIFY_DATA` option.

### DIFF
--- a/fabtests/prov/efa/src/efa_gda.c
+++ b/fabtests/prov/efa/src/efa_gda.c
@@ -227,7 +227,7 @@ static int run()
 			return ret;
 		}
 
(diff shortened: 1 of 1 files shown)


a-szegel:
sorry for the late review, missed this.

alekswn:
bot:aws:retest

alekswn:
rebased on current main

alekswn:
bot:aws:retest

PR #11539, opened 2025-10-22
prov/efa: Require FI_RX_CQ_DATA on devices without unsolicited write recv

jiaxiyan:
When unsolicited write recv is not supported, RDMA write with immediate data
operations consume an RX buffer, requiring the FI_RX_CQ_DATA mode bit to be
set for FI_RMA.

cq_data_size shouldn't be used to indicate support for fi_writedata
because there are reasonable use cases where the cq_data_size is 0 and
the completion needs to be generated.


bwbarrett (commented):
I don't think you should ever get to the point where this is necessary.  FI_RX_CQ_DATA is a mode bit.  If the provider requires FI_RX_CQ_DATA, the hints didn't set FI_RX_CQ_DATA, and FI_RMA is requested, then the provider should just not be selected.  That doesn't seem to be what this patch does?

jiaxiyan:
>  If the provider requires FI_RX_CQ_DATA, the hints didn't set FI_RX_CQ_DATA, and FI_RMA is requested, then the provider should just not be selected. That doesn't seem to be what this patch does?

The provider doesn't require FI_RX_CQ_DATA.

bwbarrett:
> The provider doesn't require FI_RX_CQ_DATA.

If the hardware doesn't support unsolicited write, it definitely does.

prov/efa/src/efa_user_info.c
```diff
@@ -474,6 +474,10 @@ int efa_user_info_alter_direct(int version, struct fi_info *info, const struct f
 	if ((hints && (hints->caps & FI_RMA)) &&
 	    (!efa_device_support_unsolicited_write_recv() &&
 	     !(info->mode & FI_RX_CQ_DATA))) {
+		/**
+		 * fi_writedata requires unsolicited write recv or FI_RX_CQ_DATA
+		 */
+		info->caps &= ~FI_RMA;
```
rauteric: In this case, what will happen if the application requested `FI_RMA`? Will efa-direct eventually return -ENODATA?

bwbarrett: The provider just shouldn't be returned from fi_getinfo().  Whether or not fi_getinfo() returns -ENODATA is not up to any particular provider / fabric.

darrylabbate (approved):

PR #11553, opened 2025-10-28
prov/efa: Do not fail fi_getinfo for the wrong fabric 

jiaxiyan:
When user requests fabric name, only alter the info that match the name. Continue to check the next prov_info if altering the info fails instead of failing the fi_getinfo call.

prov/efa: Clear FI_RMA when user doesn't request it and unsupported
https://github.com/ofiwg/libfabric/commit/25b1fa8c339fa507f21c9a5e8fc5c9c89ce49a26 set FI_RMA in efa direct prov info even when unsolicited write recv is not supported to advertise efa direct CAN support it so that it doesn't fail ofi_check_info.
When efa device does not support unsolicited write recv and the application does not request FI_RMA with FI_RX_CQ_DATA, we should clear FI_RMA.


prov/efa/src/efa_user_info.c
```diff
@@ -471,19 +471,25 @@ int efa_user_info_alter_direct(int version, struct fi_info *info, const struct f
...
+			 * FI_RX_CQ_DATA is required when application requests
+			 * FI_RMA.
+			 */
+			EFA_INFO(FI_LOG_CORE,
+				 "FI_RX_CQ_DATA is required for FI_RMA when "
+				 "unsolicited write recv is not supported.\n");
+			return -FI_ENODATA;
+		}
+		/** Clear FI_RMA if user doesn't request it */
+		info->caps &= ~(OFI_TX_RMA_CAPS | OFI_RX_RMA_CAPS);
```
shijin-aws, line 498: `info` is already the duped `prov_info` , which should be responsible for cutting unneeded capabilities anyway.  `FI_RMA` is a primary capability so we shouldn't add the caps if user doesn't request it, https://github.com/ofiwg/libfabric/commit/25b1fa8c339fa507f21c9a5e8fc5c9c89ce49a26 doesn't change this behavior though. Is this behavior already wrong before that commit? If it is, then I do not think it matters with unsolicited write recv or RX CQ data

jiaxiyan: If hints is null, ofi_alter_info won't trim it. It will just return prov_info with FI_RMA.

jiaxiyan: It wasn't wrong before because when application requests null hints to get the default prov_info, we only returned FI_RMA with unsolicited write recv.

sunkuamzn: I think Jessie's change is in the right direction. Consider the situation when calling fi_geinfo with NULL hints on an instance that doesn't support unsolicited write recv

Previously:
efa-direct prov_info did not set FI_RMA. So we never returned efa-direct fi_info object

Now:
efa-direct prov_info always sets FI_RMA no matter the hints. If the hints set both FI_RMA and FI_RX_CQ_DATA, we want to return the efa-direct info object.

The docs say there is a difference between NULL hints and hints with mode bits not set. With NULL hints, we are supposed to assume that all capabilities are requested and return which mode bits the provider supports
> If capabilities in the hint parameter are set to 0, the underlying provider will return the set of capabilities which are supported.

> If a NULL hints structure is provided, then the provider’s supported set of modes will be returned in the info structure(s).

So I think the new behavior should be:
1. For NULL hints, we return efa-direct info object with FI_RMA and FI_RX_CQ_DATA
2. For not-null hints that request FI_RMA but don't set FI_RX_CQ_DATA, we don't return efa-direct

For the not-null case without FI_RX_CQ_DATA, because prov_info now has FI_RMA cap set, we are returning the efa-direct info object with FI_RMA cap even though the hints did not set FI_RX_CQ_DATA. To stop that (which this PR is trying to do), we need to unset FI_RMA when the hints is not NULL but do not set FI_RX_CQ_DATA. Because we process the hints in user_info, we have to unset FI_RMA in user_info

So @jiaxiyan you need to add (1) a `if (hints)` check before unsetting FI_RMA and (2) set FI_RX_CQ_DATA mode bit for NULL hints (this may already be happening)

jiaxiyan: Added FI_RX_CQ_DATA for NULL hints when unsolicited write recv is not supported so FI_RMA won't be unset. Also added unit tests.

prov/efa/src/efa_user_info.c
```diff
@@ -466,24 +466,37 @@ int efa_user_info_alter_direct(int version, struct fi_info *info, const struct f
 	 * will respect it, and will finally disable unsolicited
 	 * write recv during the QP creation.
 	 */
-	if (hints && hints->mode & FI_RX_CQ_DATA) {
+	if ((hints && hints->mode & FI_RX_CQ_DATA) ||
+	    (!hints && !efa_device_support_unsolicited_write_recv())) {
```
sunkuamzn: I think we should always return FI_RX_CQ_DATA for NULL hints. The man page says that, for NULL hints, provider should return all mode bits that it supports
> If a NULL hints structure is provided, then the provider’s supported set of modes will be returned in the info structure(s).

```suggestion
	    (!hints) {
```

prov/efa/src/efa_user_info.c
sunkuamzn: I didn't realize that unsetting FI_RMA and returning ENODATA have to happen on the same path, so we can remove one of them

sunkuamzn (approved):

PR #11564, opened 2025-10-30
fabtests/getinfo: Remove test_caps_regression

jiaxiyan:
The man page requires that
"If capabilities in the hint parameter are set to 0, the underlying provider will return the set of capabilities which are supported." test_caps_regression prohibits provider from adding a capability that requires a mode, which contradicts with the man page. Newer MPI do not have this issue any more. Remove this test so providers can remain compliant with the man page.

### DIFF
--- a/man/fi_getinfo.3.md
+++ b/man/fi_getinfo.3.md
@@ -252,13 +252,17 @@ struct fi_info {
 # CAPABILITIES
 
 Interface capabilities are obtained by OR-ing the following flags
-together.  If capabilities in the hint parameter are set to 0, the
-underlying provider will return the set of capabilities which are
-supported.  Otherwise, providers will return data matching the
-specified set of capabilities.  Providers may indicate support for
-additional capabilities beyond those requested when the use of
-expanded capabilities will not adversely affect performance or expose
-the application to communication beyond that which was requested.
+together. If a NULL hints structure is provided, the underlying
+provider will return the set of capabilities which are supported.
+If a hints structure is provided but the capabilities in the hints
+structure are set to 0, the underlying provider will only return the
+set of capabilities that the provider can support with the mode bits
+set by the application in the hints structure. Otherwise, if
+capabilities are specified with non-zero values, providers will return
+data matching the specified set of capabilities. Providers may indicate
+support for additional capabilities beyond those requested when the use
+of expanded capabilities will not adversely affect performance or
+expose the application to communication beyond that which was requested.
 Applications may use this feature to request a minimal set of
 requirements, then check the returned capabilities to enable
 additional optimizations.


shefty:
We have to assume there are other apps beyond specific MPI implementations which made the same assumption regarding how the caps/mode bits work.

I suggest updating the man page to reflect what the actual behavior was / should be.

If caps and mode are both 0, allow the provider to return the set of capabilities that are supported.  This may result in mode bits being set on output.  If caps is 0, but mode is non-zero, the provider returns the caps that are supported within the constraints of the mode bits that are set.  That is, the mode bits cannot be ignored.

sunkuamzn:
@shefty how about this?

Current man page
>  If capabilities in the hint parameter are set to 0, the underlying provider will return the set of capabilities which are supported. 

Suggestion
>  If a NULL hints structure is provided, the underlying provider will return the set of capabilities which are supported. If a hints structure is provided but the capabilities in the hints structure are set to 0, the underlying provider will only return the set of capabilities that the provider can support with the mode bits set by the application in the hints structure.

j-xiong:
I agree more on @sunkuamzn 's version.

If `hints` is provided, and `mode` is 0, it's improper to add capabilities that requires non-zero mode. If the application just wants to check all the possibilities, it can set mode to -1.

jiaxiyan:
How does this test align with the suggested man page change? It is requesting hints->caps = 0 and hints->mode = ~0, and then requesting with the returned caps and FI_CONTEXT.

j-xiong:
> How does this test align with the suggested man page change? It is requesting hints->caps = 0 and hints->mode = ~0, and then requesting with the returned caps and FI_CONTEXT.

Anything wrong with that? `mode = ~0` is equivalent to `mode = -1`.

shefty:
The man page does have this text:
```
 With the exception of mode bits, hints that are set to zero are treated as a wildcard.
A zeroed hint value results in providers either returning a default value or a value that
works best for their implementation. Mode bits that are set to zero indicate the
application does not support any modes.
 ```
 So, mode of 0 is NOT treated as a wildcard, as my previous suggestion indicated.
 
"It is requesting hints->caps = 0 and hints->mode = ~0, and then requesting with the returned caps and FI_CONTEXT."
My expectation is that the second request must return a subset of the original request.

jiaxiyan:
The first fi_getinfo returns the caps that are supported by all the mode bits, including the cap that requires a specific mode bit. The second fi_getinfo request all caps but only with FI_CONTEXT, so it can fail with -FI_ENODATA. Is this expected?

shefty:
The second call failing with ENODATA should be considered possible, but the second call _could_ also pass.

sunkuamzn:
If the second call is allowed to either pass or fail, the test is not useful and we need to remove it anyway

@shefty if we update the man page and remove the test, would that be OK?

shefty:
Yes - I'm unsure what this specific test is trying to check, but on the surface it looks wrong to me.

j-xiong:
I think the scenario the test is trying to check is an incorrect application assumption. It is obsolete and can be removed.

jiaxiyan:
@j-xiong Is the intel CI failure related?

j-xiong:

PR #11568, opened 2025-10-31
prov/efa: Add support for blocking CQ read operations for RDM

darrylabbate:
This adds support for `fi_cq_sread{,from}`, `fi_control` and `fi_signal` for the `efa` fabric. This also expands EFA's `fi_trywait()` to handle wait objects for Util CQs.

Since the endpoint needs to progress in order to generate completions to read from, an exponential backoff strategy is employed to poll by driving progress periodically without excessively spinning the CPU for higher latency scenarios. The `timeout` specified by the user is used to cap the total time spent waiting for completions.

### DIFF
--- a/fabtests/pytest/efa/test_rdm.py
+++ b/fabtests/pytest/efa/test_rdm.py
@@ -162,12 +162,15 @@ def test_rdm_bw_zcpy_recv_use_fi_more(cmdline_args, memory_type, zcpy_recv_max_m
 
 @pytest.mark.functional
 @pytest.mark.parametrize("comp_method", ["sread", "fd"])
-def test_rdm_pingpong_sread(cmdline_args, completion_semantic, memory_type_bi_dir, direct_message_size, support_sread, comp_method):
+def test_rdm_pingpong_sread(cmdline_args, completion_semantic, memory_type_bi_dir,
+                            direct_message_size, support_sread, comp_method, fabric):
     if not support_sread:
         pytest.skip("sread not supported by efa device.")
+    if fabric == "efa" and comp_method == "sread" and completion_semantic == "delivery_complete":
+        pytest.skip("Skip delivery-complete with sread to avoid manual-progress deadlock.")
     efa_run_client_server_test(cmdline_args, f"fi_rdm_pingpong -c {comp_method}", "short",
                                completion_semantic, memory_type_bi_dir,
-                               direct_message_size, fabric="efa-direct")
+                               direct_message_size if fabric == "efa-direct" else "all", fabric=fabric)
 
 
 # These tests skip efa-direct because efa-direct does not
--- a/fabtests/pytest/efa/test_rma_bw.py
+++ b/fabtests/pytest/efa/test_rma_bw.py
@@ -91,12 +91,12 @@ def test_rma_bw_sread(cmdline_args, rma_operation_type, rma_bw_completion_semant
                       direct_rma_size, rma_bw_memory_type, support_sread, comp_method, rma_fabric):
     if not support_sread:
         pytest.skip("sread not supported by efa device.")
-    if rma_fabric == "efa":
-        pytest.skip("sread not implemented in efa fabric yet.")
+    if rma_fabric == "efa" and comp_method == "sread" and rma_bw_completion_semantic == "delivery_complete":
+        pytest.skip("Skip delivery-complete with sread to avoid manual-progress deadlock.")
     command = f"fi_rma_bw -e rdm -c {comp_method}"
     command = command + " -o " + rma_operation_type
     # rma_bw test with data verification takes longer to finish
     timeout = max(1080, cmdline_args.timeout)
     efa_run_client_server_test(cmdline_args, command, "short", rma_bw_completion_semantic,
-                               rma_bw_memory_type, direct_rma_size,
+                               rma_bw_memory_type, direct_rma_size if rma_fabric == "efa-direct" else "all",
                                timeout=timeout, fabric=rma_fabric)


darrylabbate:
CCing @j-xiong for the `fi_tostr` datatypes additions

shijin-aws:
can u you expand https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/efa/test_rdm.py#L165 and https://github.com/ofiwg/libfabric/blob/main/fabtests/pytest/efa/test_rma_bw.py#L90 to test sread/fd with efa fabric?

prov/efa/src/efa_fabric.c
```diff
@@ -63,16 +64,28 @@ static int efa_fabric_close(fid_t fid)
...
 	for (i = 0; i < count; i++) {
 		if (fids[i]->fclass == FI_CLASS_CQ) {
-			/* Use EFA-specific CQ trywait */
-			efa_cq = container_of(fids[i], struct efa_cq, util_cq.cq_fid.fid);
-			ret = efa_cq_trywait(efa_cq);
-			if (ret)
-				return ret;
+			util_cq = container_of(fids[i], struct util_cq, cq_fid.fid);
+			
+			/* RDM CQs use util wait objects, not hardware CQ events */
```
shijin-aws, line 76: what are the benefits of using util wait instead of efa device interrupt to implement wait? Is it because there can be too many device level completions that is only for 1 libfabric completion?

shijin-aws (commented):
May I know if this PR is following some existing code examples in other providers? The logic looks reasonable to me but I am actually not super familiar with the ofi_wait functions yet

darrylabbate:
> May I know if this PR is following some existing code examples in other providers? The logic looks reasonable to me but I am actually not super familiar with the ofi_wait functions yet

I referenced util, CXI and a few others. I also originally added some defensive logic for a wait object for the SHM CQ FD [here](https://github.com/darrylabbate/libfabric/commit/6ab6cc8a22baf6f4aa3b761e2e921d812b1b267a), but I opted to exclude it from the PR since it's purely for future proofing - SHM doesn't support `FI_GETWAIT` on the CQ

prov/efa/src/efa_fabric.c
shijin-aws: The major requirement of the blocking cq read is to avoid burning the CPU which was the busy loop that application will do when sread is not available. Are we confident this approach is meeting this requirement? cc @bwbarrett

darrylabbate: My understanding is we'd need SHM to expose some kind of wait object (e.g. `FI_WAIT_FD`) in order to implement any kind of multiplexing with EFA device interrupts. I'd be more than happy to engineer a solution but I figured that would be out of scope given the context/urgency of the feature. Could we consider this a follow-up/future improvement?

Short of that, don't we need to periodically drive progress or risk missing completions? I figured the exponential backoff strategy with a minimum interval of 1ms was appropriate since it matches backoff strategies we employ elsewhere. The `ofi_wait()` calls block on the util CQ FD between retries, so it's not "spinning," per se.

prov/efa/src/efa_fabric.c
```diff
@@ -63,16 +64,28 @@ static int efa_fabric_close(fid_t fid)
...
+			if (util_cq->wait) {
+				rdm_cq = container_of(util_cq, struct efa_rdm_cq, efa_cq.util_cq);
+				ret = efa_rdm_cq_trywait(rdm_cq);
```
shijin-aws, line 79: How will it work on single node when shm is on? AFAICT it is only calling efa_rdm_cq_progress which doesn't progress shm cq.

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -854,14 +880,105 @@ static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t coun
...
+			return num_completions ? (ssize_t) num_completions : -FI_EAGAIN;
+
+		/* Check for wakeup signal */
+		if (OFI_UNLIKELY(ofi_atomic_get32(&util_cq->wakeup))) {
+			ofi_atomic_set32(&util_cq->wakeup, 0);
+			return num_completions ? (ssize_t) num_completions : -FI_EAGAIN;
+		}
+
+		/* Progress then wait */
+		efa_rdm_cq_progress(util_cq);
```
shijin-aws: This only progress efa, not shm.

darrylabbate: This was a miss. I'll add it here.

darrylabbate:
Added https://github.com/ofiwg/libfabric/pull/11568/commits/65c9de6788cb4f4e51176437e7f6e19c11d660ab after the previous changes exposed a potential use-after-free issue upon SHM CQ init failure. More of a "nice-to-have" since the Libfabric spec doesn't make guarantees about the value of the CQ FID upon cq_open failure

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: If you change it to something like `fi_cq_read(efa, NULL, 0)` it should progress both efa and shm

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -903,6 +1020,69 @@ static void efa_rdm_cq_progress(struct util_cq *cq)
...
+ * @param[in]	cq	RDM CQ to check
+ * @return	FI_SUCCESS if ready to wait, -FI_EAGAIN if completions available
+ */
+int efa_rdm_cq_trywait(struct efa_rdm_cq *cq)
+{
+	struct efa_domain *domain;
+
+	domain = container_of(cq->efa_cq.util_cq.domain, struct efa_domain, util_domain);
+
+	if (cq->shm_cq) {
```
shijin-aws: why not just do fi_cq_read(NULL, 0) for efa, it should progress both efa and shm

darrylabbate:
Currently, any fabtests that explicitly test `fd` completions (`-c fd`) will hang since they unconditionally block on the provided FD before calling `fi_cq_read()`. Probably need to modify the shared code to ~~call `fi_cq_sread()`~~ or spawn a progress thread when the provider advertises `FI_PROGRESS_MANUAL`.

shijin-aws:
> Currently, any fabtests that explicitly test `fd` completions (`-c fd`) will hang since they unconditionally block on the provided FD before calling `fi_cq_read()`. Probably need to modify the shared code to ~call `fi_cq_sread()`~ or spawn a progress thread when the provider advertises `FI_PROGRESS_MANUAL`.

If we do not need shm support for efa rdm blocking cq read right now, would it be easier to turn off shm when wait is requested in cq open?

darrylabbate:
> If we do not need shm support for efa rdm blocking cq read right now, would it be easier to turn off shm when wait is requested in cq open?

Sure, let's try that

darrylabbate:
As I suspected, I don't think the SHM CQ is the issue here. Disabling SHM has no effect.

The actual issue is that AFAICT, `FI_WAIT_FD` is only possible for manual progress providers with a separate application thread driving progress. `ft_fdwait_comp` doesn't currently account for this. I ran into race condition issues with the CQ polling procedure trying to add this logic (surprise!), but I'll dig into it further.

The latest commits I've pushed disables fabtests in the EFA pytest suite with `fd` completion method and `efa` fabric. The fabtest code will need to be updated before this is merged.

darrylabbate:
Looks like my changes to skip fabtests with `-c fd -f efa` didn't get picked up, so those failures are expected for now.

ETA: that'd be because I didn't add the logic for the RMA BW tests. I thought I implemented the skip for those but _not_ RDM pingpong. Doesn't matter.

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: Ack, replaced.

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: Ack, replaced

prov/efa/src/efa_fabric.c
darrylabbate: `efa_rdm_cq_trywait()` should now be progressing the SHM CQ via `fi_cq_read()`

darrylabbate:
CC @j-xiong and/or @aingerson for the fabtests change: ~~https://github.com/ofiwg/libfabric/pull/11568/commits/9e7eb4cc6ddaa68767af0a587bed8885b8ca188c~~ https://github.com/ofiwg/libfabric/pull/11568/commits/1610850413c7f5dfe716bb748b825f3e1c57ebda

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: Is it updated yet?

fabtests/common/shared.c
```diff
@@ -2745,10 +2778,27 @@ static int ft_wait_for_comp(struct fid_cq *cq, uint64_t *cur,
 			    uint64_t tag)
 {
 	struct fi_cq_err_entry comp;
-	int ret;
+	int ret, remaining = timeout;
 
 	while (total - *cur > 0) {
-		ret = fi_cq_sread(cq, &comp, 1, NULL, timeout);
+		if (ft_fdwait_need_progress)
+			ret = fi_cq_read(cq, &comp, 1);
```
shijin-aws: I am not sure why we need this ... cq_sread should progress the completions for a manual_progress provider anyway. also, sread is designed for wait obj FI_WAIT_UNSPEC, we shouldn't have a fdwait leak here even if provider may use wait_fd to implement sread

prov/efa/src/efa_fabric.c
shijin-aws: Now this seems to be contradicting to your earlier change on fabtests that adding the progress in the waitfd code path, if we think that calling a progress function outside `fi_trywait` for manual_progress provider is required, do we still need a progress function inside `fi_trywait` ?

fabtests/common/shared.c
darrylabbate: This is solely a workaround for `-U` when the provider advertises manual progress. But now I'm thinking `-U` may not be valid for manual progress.

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: Sorry, I thought I was commenting on a different line. The `efa_rdm_cq_readfrom()` call at the top of the loop is already progressing and flushing the SHM completions into the util CQ. Then we progress the EFA CQ to generate completions.

shijin-aws: yes, but you may want to be consistent, efa_rdm_cq_progress doesn't really progress the whole cq

prov/efa/src/efa_fabric.c
darrylabbate: `fi_trywait` won't process SHM completions and move them into the util CQ. My comment in `efa_rdm_cq_trywait` is probably misleading.

fabtests/common/shared.c
shijin-aws: Not sure I follow, `-U` is `run fabtests with FI_DELIVERY_COMPLETE set`, is it relevant?

darrylabbate: Delivery-complete sends only complete after the peer consumes them. If both EPs post sends and immediately block on their TX CQ via `fi_cq_sread()`, no thread ever calls into the RX CQ, so neither side makes progress. The periodic progress forces each side to re-enter libfabric, drain the RX CQ, and let the delivery-complete TX CQE appear. Under transmit-complete, the TX CQE is generated locally, so sread eventually returns without any extra help.

prov/efa/src/efa_fabric.c
shijin-aws: >fi_trywait won't process SHM completions and move them into the util CQ. 

Are u saying fi_trywait for shm cq (no efa) will not progress shm completions, so you need that logic in fabtests? With your PR, fi_trywait for efa cq (+shm as peer) will progress both, right?

fabtests/common/shared.c
shijin-aws: >If both EPs post sends and immediately block on their TX CQ via fi_cq_sread(), no thread ever calls into the RX CQ, so neither side makes progress. The periodic progress forces each side to re-enter libfabric, drain the RX CQ, and let the delivery-complete TX CQE appear. Under transmit-complete, the TX CQE is generated locally, so sread eventually returns without any extra help.

How to implement DC is provider specific, application doesn't have to poll its RX CQ to make the TX CQ on peer side to get the delivery completion ack. EFA provider fill this gap by making the Libfabric TX CQ poll also poll the rx device CQ internally. So the behavior you described should never happen?

shijin-aws: But you raised a good question. Today, the receiver needs to poll CQ so the sender can get completions for DC. I haven't checked in man page whether it is a legal behavior for manual progress.

darrylabbate: The deadlock is technically compliant from the provider's perspective.

I see what happened here. I removed the unconditional exponential backoff when the wait object is `FI_WAIT_FD` after your initial comments re: spinning the CPU. For both `fd` and `sread` completion methods, the wait object is set to `FI_WAIT_FD`, so given an infinite timeout, it will hang forever after the initial manual progress. ~~Perhaps for an infinite timeout, we fallback to the exponential backoff?~~ On second thought, I think the deadlock may ultimately be more "correct" here. As you mentioned earlier, the app probably has a good reason to utilize the blocking CQ read for resource efficiency, so this seems like a necessary tradeoff to that end.

man/fi_efa.7.md
```diff
@@ -65,7 +65,16 @@ The following features are supported:
 *Completion events*
 : The provider supports *FI_CQ_FORMAT_CONTEXT*, *FI_CQ_FORMAT_MSG*, and
   *FI_CQ_FORMAT_DATA*. *FI_CQ_FORMAT_TAGGED* is supported on the `efa` fabric
-  of RDM endpoint. Wait objects are not currently supported.
+  of RDM endpoint.
+  
+  The `efa` fabric of RDM endpoint supports *FI_WAIT_FD* and *FI_WAIT_NONE* wait
+  objects for blocking CQ operations (*fi_cq_sread*). Applications should use
```
shijin-aws: Also `FI_WAIT_UNSPEC`?

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: Bump this... efa_rdm_cq_progress only progress efa endpoints. if we want to progress the whole state machine here, it should be a fi_cq_read(NULL, 0), if we do not need a progress here (because there is a cq_readfrom earlier already).. we can just remove it

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -854,14 +858,112 @@ static ssize_t efa_rdm_cq_readfrom(struct fid_cq *cq_fid, void *buf, size_t coun
...
+		if (OFI_UNLIKELY(ofi_atomic_get32(&util_cq->wakeup))) {
+			ofi_atomic_set32(&util_cq->wakeup, 0);
+			return num_completions ? (ssize_t) num_completions : -FI_EAGAIN;
+		}
+
+		/* Progress then wait */
+		efa_rdm_cq_progress(util_cq);
+		
+		/* Check if progress produced completions before blocking */
+		if (OFI_LIKELY(!ofi_cirque_isempty(util_cq->cirq)))
```
shijin-aws: Ugh, so you have an is empty check here ... then a progress is required before this?

man/fi_efa.7.md
darrylabbate: Added

prov/efa/src/rdm/efa_rdm_cq.c
darrylabbate: Removed this and the preceding progress call

fabtests/common/shared.c
```diff
@@ -1883,6 +1900,20 @@ static void ft_cleanup_mr_array(struct ft_context *ctx_arr, char **mr_bufs)
...
+	int interval;
+
+	if (!ft_fdwait_need_progress)
+		return remaining;
+
+	if (remaining < 0)
+		return FT_FDWAIT_PROGRESS_INTERVAL_MS;
+
+	interval = MIN(remaining, FT_FDWAIT_PROGRESS_INTERVAL_MS);
+	return interval > 0 ? interval : FT_FDWAIT_PROGRESS_INTERVAL_MS;
```
j-xiong: Can be simplified as:
```
	if (remaining <= 0)
		return FT_FDWAIT_PROGRESS_INTERVAL_MS;

	return MIN(remaining, FT_FDWAIT_INTERVAL_MS);
```
And the `interval` variable can be removed,

fabtests/common/shared.c
```diff
@@ -2770,17 +2802,36 @@ static int ft_fdwait_for_comp(struct fid_cq *cq, uint64_t *cur,
...
 	fd = cq == txcq ? tx_fd : rx_fd;
 	fids[0] = &cq->fid;
 
 	while (total - *cur > 0) {
 		ret = fi_trywait(fabric, fids, 1);
 		if (ret == FI_SUCCESS) {
-			ret = ft_poll_fd(fd, timeout);
+			int wait_timeout = ft_fdwait_need_progress
+				? ft_fdwait_poll_timeout(remaining)
+				: remaining;
```
j-xiong: the same condition is checked inside `ft_fdwait_poll_timeout()`.  can eliminate this variable and use `ft_fdwait_poll_timeout(remaining)` directly.

fabtests/common/shared.c
```diff
@@ -2791,9 +2842,10 @@ static int ft_fdwait_for_comp(struct fid_cq *cq, uint64_t *cur,
 		} else if (ret < 0 && ret != -FI_EAGAIN) {
 			return ret;
 		}
+		ret = 0;
 	}
 
-	return 0;
+	return ret;
```
j-xiong: There is no error path reaching here, can just return 0.

fabtests/common/shared.c
```diff
@@ -2791,9 +2842,10 @@ static int ft_fdwait_for_comp(struct fid_cq *cq, uint64_t *cur,
 		} else if (ret < 0 && ret != -FI_EAGAIN) {
 			return ret;
 		}
+		ret = 0;
```
j-xiong: This becomes unnecessary as well.

darrylabbate:
AWS CI is only failing due to an unrelated infrastructure issue

darrylabbate:
aws:bot:retest

darrylabbate:
aws:bot:retest

darrylabbate:
bot:aws:retest

man/fi_efa.7.md
```diff
@@ -65,7 +65,16 @@ The following features are supported:
 *Completion events*
 : The provider supports *FI_CQ_FORMAT_CONTEXT*, *FI_CQ_FORMAT_MSG*, and
   *FI_CQ_FORMAT_DATA*. *FI_CQ_FORMAT_TAGGED* is supported on the `efa` fabric
-  of RDM endpoint. Wait objects are not currently supported.
+  of RDM endpoint.
+  
+  The `efa` fabric of RDM endpoint supports *FI_WAIT_FD*, *FI_WAIT_UNSPEC* and
```
shijin-aws: Nit: is `efa-direct` support documented anywhere?

fabtests/common/shared.c
```diff
@@ -768,6 +773,16 @@ int ft_open_fabric_res(void)
 	return ft_open_domain_res();
 }
 
+static inline void ft_update_fdwait_progress_requirement(struct fi_info *info)
+{
+	bool manual_progress = info && info->domain_attr &&
+		(info->domain_attr->data_progress == FI_PROGRESS_MANUAL ||
+		 info->domain_attr->control_progress == FI_PROGRESS_MANUAL);
+
+	ft_fdwait_need_progress = manual_progress &&
```
jiaxiyan, line 782: Why do you use a static bool instead of returning bool?

fabtests/pytest/efa/test_rdm.py
```diff
@@ -162,12 +162,16 @@ def test_rdm_bw_zcpy_recv_use_fi_more(cmdline_args, memory_type, zcpy_recv_max_m
 
 @pytest.mark.functional
 @pytest.mark.parametrize("comp_method", ["sread", "fd"])
-def test_rdm_pingpong_sread(cmdline_args, completion_semantic, memory_type_bi_dir, direct_message_size, support_sread, comp_method):
+@pytest.mark.parametrize("fabric", ["efa", "efa-direct"])
```
jiaxiyan: `fabric` was defined in conftest.py

prov/efa/src/efa_fabric.c
shijin-aws: >My understanding is we'd need SHM to expose some kind of wait object (e.g. FI_WAIT_FD) in order to implement any kind of multiplexing with EFA device interrupts. I'd be more than happy to engineer a solution but I figured that would be out of scope given the context/urgency of the feature. Could we consider this a follow-up/future improvement?

Considering the following scenario: when application have 2 threads, 1 thread poll the fd, 1 thread progress the cq via fi_cq_read(NULL, 0). When we only have intra node (shm traffic), will thread 1 be signaled ? We know shm doesn't have any signals when writing the cq entries because it doesn't support wait

man/fi_efa.7.md
darrylabbate: I'm not sure, I only went through the man page with the `efa` fabric in mind. LEt me know what you want to do here or if it can be a follow-up task.

fabtests/common/shared.c
darrylabbate: LLM moment. No opinion here. I can clean up if I need to make other changes.

man/fi_efa.7.md
shijin-aws: Since u are here, can you add that these completion events are supported for both efa and efa-direct cc @jiaxiyan

jiaxiyan: I forgot to add it in https://github.com/ofiwg/libfabric/pull/11304/files

prov/efa/src/efa_fabric.c
darrylabbate: No. The CQ(s) are only signaled upon processing device CQEs. We could technically signal when we flush SHM completions into the util CQ, but then we're superseding SHM semantics while introducing FD writes/locks. SHM completions arrive too quickly for any real benefit anyway.

Do you want to document this explicitly? I feel this is implied via the SHM provider itself.

fabtests/pytest/efa/test_rdm.py
darrylabbate: Will fix

man/fi_efa.7.md
darrylabbate: ok

fabtests/common/shared.c
darrylabbate: I will leave it as-is since it does serve as a simple cache instead of chasing all these (mutable? immutable? who knows?) pointers between many loops iterations.

man/fi_efa.7.md
```diff
@@ -93,8 +103,10 @@ The following features are supported:
 # LIMITATIONS
 
 ## Completion events
-- Synchronous CQ read is not supported.
-- Wait objects are not currently supported.
+- DGRAM endpoints do not support synchronous CQ reads (*fi_cq_sread*) or wait objects.
+- For RDM endpoints with *FI_WAIT_FD*, applications must use *fi_cq_sread()* for
+  blocking reads. Direct polling of the wait fd without calling *fi_cq_read()* or
+  *fi_cq_sread()* will not work due to manual progress requirements.
```
jiaxiyan, line 109: Is this better rephrased as `We support fi_cq_sread or direct polling of the wait fd followed by fi_cq_read`?

PR #11573, opened 2025-11-01
prov/efa: Enable ROCr HMEM support

darrylabbate:
Marking as a draft since this hasn't been tested, but there are plenty of changes made here that can undergo CI testing and review.

### DIFF
--- a/fabtests/pytest/efa/conftest.py
+++ b/fabtests/pytest/efa/conftest.py
@@ -14,12 +14,15 @@ memory_type_list_bi_dir = [
     pytest.param("cuda_to_cuda", marks=pytest.mark.cuda_memory),
     pytest.param("host_to_neuron", marks=pytest.mark.neuron_memory),
     pytest.param("neuron_to_neuron", marks=pytest.mark.neuron_memory),
+    pytest.param("host_to_rocr", marks=pytest.mark.rocr_memory),
+    pytest.param("rocr_to_rocr", marks=pytest.mark.rocr_memory),
 ]
 
 # Add more memory types that are useful for uni-directional tests.
 memory_type_list_all = memory_type_list_bi_dir + [
     pytest.param("cuda_to_host", marks=pytest.mark.cuda_memory),
     pytest.param("neuron_to_host", marks=pytest.mark.neuron_memory),
+    pytest.param("rocr_to_host", marks=pytest.mark.rocr_memory),
 ]
 
 @pytest.fixture(scope="module", params=memory_type_list_all)


prov/efa/src/efa_hmem.c
```diff
@@ -181,6 +188,79 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
...
+		rocr_free(ptr);
+		return;
+	}
+
+#if HAVE_EFA_DMABUF_MR
+	ret = rocr_hmem_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
+	if (ret == FI_SUCCESS) {
+		ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
+					   len, (uint64_t) ptr, dmabuf_fd, ibv_access);
+		(void) rocr_hmem_put_dmabuf_fd(dmabuf_fd);
```
shijin-aws, line 223: I didn't see dmabuf ops in rocr_hmem.c. Rocr doesn't support it? That means you will always try the side channel, and that likely won't work unless the efa driver implemented it for rocr

bwbarrett: the only path the efa driver is going to support with ROCm is dmabuf.

darrylabbate: > I didn't see dmabuf ops in rocr_hmem.c

Help me out here: what's missing from the current `rocr_hmem_{get,put}_dmabuf_fd()` functions? It's currently gated behind `FI_HMEM_ROCR_USE_DMABUF`, which we can discuss changing the default there. But it should otherwise hit `hsa_amd_portable_export_dmabuf()` as long as the ROCr headers expose it at compile-time.

Were you talking about the fabtests code? Because yeah I think that's missing and I'll add that functionality.

shijin-aws: Sorry for the confusion. I can see rocr_hmem_{get,put}_dmabuf_fd() is implemented in src/hmem_rocr.c, (and not in fabtests code yet), but this PR doesn't seem to go this code path for the fi_mr_regattr via a VA (no FI_MR_DMABUF) as it finally calls ibv_reg_mr

The expected workflow would be 

1. efa_mr_reg_ibv_mr should retrieve the dmabuf fd internally and reg via ibv_reg_dmabuf_mr, when dmabuf is available
2. fabtests implement that dmabuf code as well and it will call fi_mr_regattr with FI_MR_DMABUF flags and fd directly

darrylabbate: Ack, I'll add this

prov/efa/src/efa_mr.c
```diff
@@ -574,6 +577,38 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
 		return NULL;
 	}
 
+	/*
+	 * Try to retrieve a dmabuf handle for ROCr buffers so we can register
+	 * with ibv_reg_dmabuf_mr. If dmabuf is not supported, fall back to
+	 * plain ibv_reg_mr on the VA.
+	 */
+	if (efa_mr->peer.iface == FI_HMEM_ROCR) {
```
shijin-aws: why not just change l550 to `if (efa_mr_is_neuron(efa_mr) || efa_mr_is_rocm(efa_mr)) {`

shijin-aws: there is some changes for the string err message needed to make it general

darrylabbate: Fixed

darrylabbate:
After rebasing, something is causing the LTTng builds to fail. I suspect it's something with the data path direct trace points that were added

darrylabbate:
@j-xiong Is the Intel CI failure real?

j-xiong:
The Intel CI failures are known intermittent issue. They are unrelated.

include/rdma/fi_domain.h
```diff
@@ -179,6 +179,7 @@ struct fi_mr_attr {
 		int		ze;
 		int		neuron;
 		int		synapseai;
+		int		rocr;
```
shijin-aws, line 182: This is a core API change so need @j-xiong to review... I am surprised that we didn't have that earlier

prov/efa/src/efa_hmem.c
```diff
@@ -386,6 +388,72 @@ int efa_hmem_info_initialize()
...
+/**
+ * @brief Copy data from a hmem device to a system buffer
+ *
+ * @param[in]    desc            Pointer to a memory registration descriptor
+ * @param[out]   dest            Destination system memory buffer
+ * @param[in]    src             Source hmem device memory
+ * @param[in]    size            Data size in bytes to copy
+ * @return       FI_SUCCESS status code on success, or an error code.
+ */
+int efa_copy_from_hmem(void *desc, void *dest, const void *src, size_t size)
```
shijin-aws, line 484: This is a fast path function, can we keep them static inline?

include/rdma/fi_domain.h
j-xiong: This one is fine.

prov/efa/src/efa_hmem.c
darrylabbate: Inlining works across TU boundaries with LTO. Even without LTO, this should have no effect wrt a compiler's inlining heuristics given how small the functions are. Furthermore, the callers themselves aren't `inline`d, so it's silly to draw the line at these 2 header-inlined functions and claim we're suddenly adding a new call boundary.

This was necessary because our include graph has become a tangled clusterfuck, and it's impossible to include tracepoints in these functions without moving them to a source file. Otherwise `#include efa_tp.h` in `efa_hmem.h` completes a circular include.

shijin-aws (approved):

PR #11577, opened 2025-11-03
prov/efa: Remove the incorrect check for device max_msg_size

jiaxiyan:
Device rdm_info is different from the info returned by fi_getinfo. Its max_msg_size is determined by g_efa_selected_device_list[0].rdm_info->caps so we shouldn't check it with info->caps here.

### DIFF
--- a/prov/efa/test/efa_unit_test_info.c
+++ b/prov/efa/test/efa_unit_test_info.c
@@ -117,11 +117,6 @@ static void test_info_direct_attributes_impl(struct fi_info *hints,
 		assert_int_equal(info->domain_attr->progress, FI_PROGRESS_AUTO);
 		assert_int_equal(info->domain_attr->control_progress, FI_PROGRESS_AUTO);
 		assert_int_equal(
-			g_efa_selected_device_list[0].rdm_info->ep_attr->max_msg_size,
-			(info->caps & FI_RMA) ?
-				g_efa_selected_device_list[0].max_rdma_size :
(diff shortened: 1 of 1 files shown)


prov/efa/test/efa_unit_test_info.c
```diff
@@ -116,11 +116,6 @@ static void test_info_direct_attributes_impl(struct fi_info *hints,
 		assert_false(info->tx_attr->msg_order & FI_ORDER_SAS);
 		assert_int_equal(info->domain_attr->progress, FI_PROGRESS_AUTO);
 		assert_int_equal(info->domain_attr->control_progress, FI_PROGRESS_AUTO);
-		assert_int_equal(
-			g_efa_selected_device_list[0].rdm_info->ep_attr->max_msg_size,
-			(info->caps & FI_RMA) ?
-				g_efa_selected_device_list[0].max_rdma_size :
-				g_efa_selected_device_list[0].ibv_port_attr.max_msg_sz);
```
shijin-aws, line 123: the commit message doesn't sound clear to me, can you elaborate in what scenario this check will fail?

jiaxiyan: When the application doesn't request FI_RMA, `info->caps & FI_RMA` is false but device max_msg_size is max_rdma_size https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_prov_info.c#L158-L159

shijin-aws: I see, then I agree that the check in L119-L124 should be sufficient

PR #11583, opened 2025-11-04
fabtests: add OOB finalize sync option

aingerson:
ft_finalize_ep is called in all tests and guarantees that no peer exits without confirming that the other peer is done as well. This is normally done with an in band send/recv with the FI_TRANSMIT_COMPLETE flag set which guarantees the send completion won't be generated until the message has completely be sent. This patch adds an out of band finalize option using the out of band socket if the user requested out of band syncs.
Typically an in band finalize is sufficient but may be problematic with certain scenarios. For example, if the provider doesn't support FI_TRANSMIT_COMPLETE (rare) and for providers using an unreliable protocol underneath (rxd) where the last underlying acknowledgement may be dropped, which can result in one side exiting before the other waiting for an acknowldgement retry that will never come

### DIFF
--- a/fabtests/functional/multi_ep.c
+++ b/fabtests/functional/multi_ep.c
@@ -72,6 +72,9 @@ static void free_ep_res()
 	int i;
 
 	for (i = 0; i < num_eps; i++) {
+		if (!fi)
+			continue;
+
 		if (fi->domain_attr->mr_mode & FI_MR_RAW)
 			(void) fi_mr_unmap_key(domain, peer_iovs[i].key);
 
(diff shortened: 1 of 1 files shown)


zachdworkin (approved):

aingerson:
@j-xiong Got all the tests to pass with a multi_ep test modification. Should be ok to merge I think, if it looks ok to you.
All fabtests are passing - CI failure is mpich test suite

fabtests/functional/multi_ep.c
```diff
@@ -473,7 +476,7 @@ static int setup_av_ep(int idx)
 	hints->src_addr = NULL;
 	hints->src_addrlen = 0;
 
-	ret = fi_getinfo(FT_FIVERSION, opts.src_addr, NULL, 0, hints, &fi);
+	ret = fi_getinfo(FT_FIVERSION, NULL, NULL, 0, hints, &fi);
```
j-xiong, line 479: The original intention could be honoring the user specified source address (`-s` option) but ignoring the port (the `-B` option).  How about keeping the `opts.src_addr` parameter but adding the `FI_SOURCE` flag?

aingerson: I tried that as well but udp didn't like that either. The other providers were ok with it though so it might just be a UDP  issue. I can try to make that option work instead. This was just an easy way to get it working

j-xiong: Hmm, then just keep it this way. The domain name in hints should also match the desired nic at this point.

j-xiong (approved):

PR #11586, opened 2025-11-04
fabtests: Add server-client sync during RMA writedata warmup

alekswn:
Warm up the server->client communication path during RMA writedata benchmark warmup iterations to mitigate connection setup latency impact on bandwidth measurements.

Changes:
- Wait for all pending completions after warmup
- Add sync message posts during writedata warmup iterations
- Increase timeout for test_rma_bw_range with SHM provider

This ensures accurate bandwidth measurements by excluding connection setup overhead from the timed portion of the benchmark.

### DIFF
--- a/fabtests/pytest/efa/conftest.py
+++ b/fabtests/pytest/efa/conftest.py
@@ -59,7 +59,7 @@ def rma_bw_completion_semantic(cmdline_args, completion_semantic, rma_operation_
                                         "r:4048,4,4148",
                                         "r:8000,4,9000",
                                         "r:17000,4,18000",
-                                        "r:0,1024,1048576"])
+                                        "r:0,4096,1048576"])
 def message_size(request):
     return request.param
 
--- a/fabtests/pytest/shm/conftest.py
+++ b/fabtests/pytest/shm/conftest.py
@@ -13,6 +13,6 @@ def memory_type(request):
                                         "r:4048,4,4148",
                                         "r:8000,4,9000",
                                         "r:17000,4,18000",
-                                        "r:0,1024,1048576"])
+                                        "r:0,4096,1048576"])
 def message_size(request):
     return request.param


shijin-aws:
My concern are two:

1. if we are changing measurement due to a specific behavior for efa device, such change should be an opt-in feature configured by cli
2. do we need sync for every message sizes?

alekswn:
> if we are changing measurement due to a specific behavior for efa device, such change should be an opt-in feature configured by cli

We are not changing measurements, but rather closing gap in warm-up behavior. Currently there is no warm-up for server->client connection.
We can do that under a new cmdline argument if you have strong opinion about it. However the argument will apply only to the specific `fi_rma_bw -o writedata` case. In other cases we're warming up both directions already.

> do we need sync for every message sizes?

 Did you want to say "warm-up for every message size"? I think we do not need it, but it what we're doing historically.

alekswn:
According to @aingerson failures in the lates Intel CI run are unrelated to fabtests.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -725,6 +730,23 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 			}
 			break;
 		case FT_RMA_WRITEDATA:
+			if (i < opts.warmup_iterations) {
+				if (opts.dst_addr)
+					ret = ft_post_rx(
+						ep,
+						FT_RMA_SYNC_MSG_BYTES,
+						&rx_ctx);
```
aingerson: You'll need to pass in separate contexts for each post (write and rx). See other calls using the ctx array

alekswn: Thank you for detailed review!
I've fixed this problem.

fabtests/benchmarks/benchmark_shared.c
```diff
@@ -689,8 +689,13 @@ int bandwidth_rma(enum ft_rma_opcodes rma_op, struct fi_rma_iov *remote)
 	offset_rma_start = FT_RMA_SYNC_MSG_BYTES +
 			   MAX(ft_tx_prefix_size(), ft_rx_prefix_size());
 	for (i = j = 0; i < opts.iterations + opts.warmup_iterations; i++) {
-		if (i == opts.warmup_iterations)
+		if (i == opts.warmup_iterations) {
```
shijin-aws, line 692: This is a separate change than`Add sync message posts during writedata warmup iterations`. You either need to make it a separate commit or rephrase your commit message title

fabtests/pytest/shm/test_rma_bw.py
```diff
@@ -11,7 +11,7 @@ def test_rma_bw(cmdline_args, iteration_type, operation_type, completion_semanti
     command = "fi_rma_bw -e rdm"
     command = command + " -o " + operation_type + " " + perf_progress_model_cli
     # rma_bw test with data verification takes longer to finish
-    timeout = max(540, cmdline_args.timeout)
```
shijin-aws, line 14: similar comment: This is a separate change than `Add sync message posts during writedata warmup iterations`. You either need to make it a separate commit or rephrase your commit message title

shijin-aws:
AWS CI still times out on one efa rma bw test

alekswn:
> AWS CI still times out on one efa rma bw test

This is a new timeout, previously it was SHM provider. Will have to increase the timeout for EFA provider as well.

aingerson:
Intel CI failure is unrelated. Please ignore

shijin-aws:
@alekswn Now AWS CI takes > 5hr. Can you try reducing the number of iterations to cut the total duration ? It used to be 3 - 4 hrs

alekswn:
> @alekswn Now AWS CI takes > 5hr. Can you try reducing the number of iterations to cut the total duration ? It used to be 3 - 4 hrs

We're doing 10 warm-up and 5 normal for each payload size in `test_rma_bw_range` (the default value for warm-up iterations). This is the most time consuming test. Should we just disable warm-up (pass `-w 0` in pytest)?
I think it would allow revert timeout increase and would make AWS CI complete even faster than before.

shijin-aws:
> > @alekswn Now AWS CI takes > 5hr. Can you try reducing the number of iterations to cut the total duration ? It used to be 3 - 4 hrs
> 
> We're doing 10 warm-up and 5 normal for each payload size in `test_rma_bw_range` (the default value for warm-up iterations). This is the most time consuming test. Should we just disable warm-up (pass `-w 0` in pytest)? I think it would allow revert timeout increase and would make AWS CI complete even faster than before.

Then try reducing the number of message sizes we are testing. It is crazy that we are testing a huge range with a step 4

alekswn:
> Then try reducing the number of message sizes we are testing. It is crazy that we are testing a huge range with a step 4

0-1MB range is the bottleneck. Reverted timeout increase and increased step to 4KB for this range.

alekswn:
AWS CI has completed in 2:42. Intel CI has failed once again.

alekswn:
@aingerson Could you check what is the Intel CI failure please?

PR #11588, opened 2025-11-05
fabtests/common: disable FI_RX_CQ_DATA correctly

shijin-aws:
The commit https://github.com/ofiwg/libfabric/commit/0b2034bfe86abbdd948eafbdd886caec898dcb93 introduces the long op LONG_OPT_NO_RX_CQ_DATA to not | FI_RX_CQ_DATA in fi_parse_api_opts when it is set. However, currently this logic doesn't get run because the ft_parse_long_opts is called after ft_parse_api_opts. This patch fixes this issue by clearing the FI_RX_CQ_DATA from hints->mode when the long opt is specified, so it will be always working for whatever sequence the function call is.

### DIFF
--- a/fabtests/common/shared.c
+++ b/fabtests/common/shared.c
@@ -1068,6 +1068,10 @@ int ft_getinfo(struct fi_info *hints, struct fi_info **info)
 		hints->domain_attr->mr_mode |= FI_MR_HMEM;
 	}
 
+	/* ft_cqdata_opcodes enum start from 1, 0 means no cq data */
+	if (opts.cqdata_op && allow_rx_cq_data)
+		hints->mode |= FI_RX_CQ_DATA;
+
 	hints->domain_attr->threading = opts.threading;
 
 	ret = fi_getinfo(FT_FIVERSION, node, service, flags, hints, info);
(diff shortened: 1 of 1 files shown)


shijin-aws:

aingerson:
@shijin-aws Sorry no it's not. We've been having a lot of udp/tcp failures lately. We are trying to get to the bottom of it. I've retriggered it just to be sure because there is a new failure which I haven't seen before but it looks the same as the previous errors with tcp.

fabtests/common/shared.c
```diff
@@ -4606,6 +4606,7 @@ int ft_parse_long_opts(int op, char *optarg)
 		return 0;
 	case LONG_OPT_NO_RX_CQ_DATA:
 		allow_rx_cq_data = false;
+		hints->mode &= ~FI_RX_CQ_DATA;
```
aingerson: Is this basically trying to undo the setting of it in the -o writedata and -o senddata paths?
ie would it make sense to remove the setting of it on those paths and just OR it in before getinfo based on the allow_rx_cq_data flag like the other caps/settings?
Just thinking towards the future to standardize the flow and make it easier to add other ops for example.

aingerson (commented):
I don't have a big issue with the change, it looks fine to me. Just clarifying

fabtests/common/shared.c
shijin-aws: >Is this basically trying to undo the setting of it in the -o writedata and -o senddata paths?
ie would it make sense to remove the setting of it on those paths and just OR it in before getinfo based on the allow_rx_cq_data flag like the other caps/settings?

Are you saying I should move the this bit OR/remove from the opts parsing to the location right before `fi_getinfo`? I can do that

aingerson: Yeah exactly. Just thinking it would be easier to change it so that you set it correctly in one place instead of having to set it in lots of places and then remove it when you want to force not use it

shijin-aws: Thanks, updated. It does make the logic cleaner.

aingerson: I think that is cleaner, thanks! Can you remove the OR'ing in of the mode in the op code parsing path as well please? I don't think it's needed anymore there, right?

shijin-aws: I think I already removed ?

aingerson: Oh sorry I don't know how I missed it lol. Thanks!

aingerson:
Intel CI failure is unrelated

PR #11591, opened 2025-11-05
prov/efa: Acquire CQ's `ep_list_lock` during counter progress

darrylabbate:
This prevents a race condition if a counter-driven thread tries to re-enter the polling procedure with an actively-polling CQ.

### DIFF
--- a/prov/efa/src/efa_cntr.c
+++ b/prov/efa/src/efa_cntr.c
@@ -146,6 +146,7 @@ static void efa_rdm_cntr_progress(struct util_cntr *cntr)
 	struct dlist_entry *item;
 	struct efa_cntr *efa_cntr;
 	struct efa_domain *efa_domain;
+	struct efa_cq *efa_cq;
 	struct efa_ibv_cq_poll_list_entry *poll_list_entry;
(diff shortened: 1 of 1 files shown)


shijin-aws (approved):
Thanks!

prov/efa/src/efa_data_path_direct_entry.h
```diff
@@ -428,6 +428,7 @@ static inline int efa_data_path_direct_start_poll(struct efa_ibv_cq *ibv_cq,
 
 static inline uint32_t efa_data_path_direct_wc_read_qp_num(struct efa_ibv_cq *ibv_cq)
 {
+	assert(ibv_cq->data_path_direct.cur_cqe != NULL);
```
shijin-aws: nit: why not just `assert(ibv_cq->data_path_direct.cur_cqe)`, similar to other asserts

shijin-aws (dismissed):

prov/efa/src/efa_data_path_direct_entry.h
darrylabbate: Changed. I don't feel strongly here.

Issue #11572, opened 2025-11-01
Coverity scan is failing


j-xiong:
Nightly Coverity scan has been failing in the build stage since Oct 16, 2025. 
```
+ make install
  GEN      prov/psm3/src/psm3_src_chksum.h
make  install-am
make[1]: Entering directory '/home/runner/work/libfabric/libfabric'
  CC       util/info.o
  CC       src/libfabric_la-fabric.lo
  CC       src/libfabric_la-fi_tostr.lo
  CC       src/libfabric_la-perf.lo
  CC       src/libfabric_la-log.lo
  CC       src/libfabric_la-var.lo
  CC       src/libfabric_la-abi_1_0.lo
......
  CCLD     prov/psm3/psm3/libhal_sockets.la
  CCLD     prov/psm3/psm3/libpsm3i.la
  CCLD     libpsm3.la
  CCLD     src/libfabric.la
./libtool: line 7563: cd: $'rdma-core/build/lib\302\240-Lrdma-core/build/lib': No such file or directory
libtool:   error: cannot determine absolute directory name of 'rdma-core/build/lib -Lrdma-core/build/lib'
make[1]: *** [Makefile:16063: src/libfabric.la] Error 1
make[1]: Leaving directory '/home/runner/work/libfabric/libfabric'
make: *** [Makefile:34240: install] Error 2
Error: Process completed with exit code 2.
```

Status of recent runs:
https://github.com/ofiwg/libfabric/actions/workflows/coverity.yml

aingerson:
@alekswn I think your recent change ff97cc7efcfc303185b8bcb782eb8e3e79be3f2b triggered the failure. Could you take a look please?

Coverity builds with the rdma-core directory in the efa arg. Here's the configure it uses:
```./configure --prefix=/home/runner/work/libfabric/libfabric/install --enable-efa=rdma-core/build --enable-mrail --enable-psm3=rdma-core/build --enable-rxd --enable-rxm --enable-shm --enable-tcp --enable-udp --enable-usnic --enable-verbs=rdma-core/build --enable-sm2 --enable-lpp CC=clang```

PR #11596, opened 2025-11-06
prov/efa: configure: replace no-break space with regular space character

alekswn:
Build with clang fails due to no-break space accidentally added to configure script. Replacing it with a regular space.

Issue: https://github.com/ofiwg/libfabric/issues/11572


shijin-aws:
what is "no-brake space" ?

alekswn:
A unicode character: https://unicode-explorer.com/c/00A0
Likely added by LLM

darrylabbate:
Do we want to add some kind of small Actions linter job to catch this going forward?

alekswn:
Would be nice to have I think.

PR #11604, opened 2025-11-10
prov/rxd, fabtests: Allow dg_addr to be 0 & test fi_av_lookup

zachdworkin:
dg_addr was previously not allowed to be 0. This caused failures when calling fi_av_lookup on fi_addr 0. rxd was returning an error instead of the valid address.

There is no fi_av_lookup test in fabtests so included in this bugfix is a test for this path. It only tests the "good cases" where a user is inserting X addresses into an AV, then checking the fi_addrs of the entire range of the AV to verify the expected behavior. These checks include verifying that the number of inserted addresses is correct, the translation from fi_addr to address from fi_av_lookup is correct and maps back to the original address, and to verify that (range - X) fi_addrs are correctly found as FI_ADDR_NOTAVAIL.

There will be a future test to handle the "bad case", where a user looks up an fi_addr outside of the av max range, but expected behavior must be defined in the man pages first. See follow up PR #11605.

### DIFF
--- a/fabtests/unit/av_test.c
+++ b/fabtests/unit/av_test.c
@@ -614,11 +614,109 @@ fail:
 	return TEST_RET_VAL(ret, testret);
 }
 
+static int
+av_lookup_good(void)
+{
+	int testret;
+	int ret;
+	int i, j, found, offset;
+	struct fid_av *av;
+	struct fi_av_attr attr;
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	uint8_t lookup_buf[4096];
+	size_t lookup_len;
+	fi_addr_t *fi_addr;
+
+	testret = FAIL;
+
+	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
+			 sizeof(fi_addr_t));
+	if (!fi_addr) {
+		sprintf(err_buf, "malloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
+	}
+
+	memset(&attr, 0, sizeof(attr));
+	attr.type = av_type;
+	attr.count = num_good_addr;
+
+	av = NULL;
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret != 0) {
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+			fi_tostr(&av_type, FI_TYPE_AV_TYPE), ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+
+	buflen = sizeof(addrbuf);
+	ret = av_create_address_list(good_address, 0, num_good_addr, addrbuf, 0,
+				     buflen);
+	if (ret < 0)
+		goto fail;
+
+	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
+	if (ret != num_good_addr) {
+		sprintf(err_buf, "fi_av_insert ret=%d, %s", ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_av_insert failed ret=%d, %s", ret,
+				fi_strerror(-ret));
+		}
+	}
+
+	found = offset = 0;
+	for (i = 0; i < attr.count; i++) {
+		memset(lookup_buf, 0, sizeof(lookup_buf));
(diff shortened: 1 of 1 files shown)


prov/rxd/src/rxd_av.c
```diff
@@ -108,18 +108,12 @@ static int rxd_av_set_addrlen(struct rxd_av *av, const void *addr)
...
 {
-	fi_addr_t dg_addr;
 	fi_addr_t rxd_addr = (intptr_t) ofi_idx_lookup(&av->fi_addr_idx,
 					     RXD_IDX_OFFSET((int)fi_addr));
 	if (!rxd_addr)
-		goto err;
-	dg_addr = (intptr_t) ofi_idx_lookup(&av->rxdaddr_dg_idx, (int)rxd_addr);
-	if (!dg_addr)
-		goto err;
+		return FI_ADDR_UNSPEC;
```
j-xiong, line 114: Isn't 0 a valid value for `rxd_addr`?

fabtests/unit/av_test.c
```diff
@@ -614,11 +614,114 @@ av_insert_stages(void)
...
+	if (ret != 0) {
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+				fi_tostr(&av_type, FI_TYPE_AV_TYPE),
+				ret, fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+	}
```
j-xiong: Remove the braces.

fabtests/unit/av_test.c
```diff
@@ -614,11 +614,114 @@ av_insert_stages(void)
...
+		}
+	}
+
+	found = offset = 0;
+	for (i = 0; i < attr.count; i++) {
+		memset(lookup_buf, 0, sizeof(lookup_buf));
+		lookup_len = sizeof(lookup_buf);
+		ret = fi_av_lookup(av, fi_addr[i], &lookup_buf, &lookup_len);
+		if (ret != 0) {
+			if (i >= num_good_addr && ret == -FI_ADDR_NOTAVAIL)
```
j-xiong: `ret` is not supposed to be `-FI_ADDR_NOTAVAIL` on error. 

`attr.count` was set to `num_good_addr` at L643, so this condition can't be true anyway.

fabtests/unit/av_test.c
```diff
@@ -614,11 +614,114 @@ av_insert_stages(void)
...
+	memset(&attr, 0, sizeof(attr));
+	attr.type = av_type;
+	attr.count = num_good_addr;
+
+	av = NULL;
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret != 0) {
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+				fi_tostr(&av_type, FI_TYPE_AV_TYPE),
+				ret, fi_strerror(-ret));
```
j-xiong: check alignment.

fabtests/unit/av_test.c
```diff
@@ -614,11 +614,114 @@ av_insert_stages(void)
...
+static int
+av_lookup_good(void)
+{
+	int testret;
+	int ret;
+	int i, j, found, offset;
+	struct fid_av *av;
+	struct fi_av_attr attr;
+	uint8_t addrbuf[4096];
+	int buflen;
```
j-xiong: use `size_t`

prov/rxd/src/rxd_av.c
zachdworkin: 0 is not a valid value for rxd_addr. This is because it is reserved for RXD_ADDR_INVALID. See rxd.h:397

j-xiong: I saw the commit message has been updated. It's good now.

j-xiong (approved):

fabtests/unit/av_test.c
```diff
@@ -614,11 +614,109 @@ av_insert_stages(void)
...
+	struct fi_av_attr attr;
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	uint8_t lookup_buf[4096];
+	size_t lookup_len;
+	fi_addr_t *fi_addr;
+
+	testret = FAIL;
+
+	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
```
aingerson, line 633: this might be clear later but why do you need to allocate an fi_addr if you have no addrs?
also do `sizeof(fi_addr)` instead of `sizeof(fi_addr_t)`

j-xiong: An array is allocated to be used in the `fi_av_insert` call. I don't see what the problem is here. And the element size for `calloc` is correct.

zachdworkin: I did this because the compiler was upset about the variable potentially not being used. It adds complexity but removes the warning

PR #11605, opened 2025-11-10
man/fi_av, prov/util, verbs, rxd, shm, sockets, fabtests/unit: Define out of range lookup behavior & add negative lookup test

zachdworkin:
Attempting fi_av_lookup with a fi_addr larger than the max entry number of the AV will result in undefined behavior. It is expected by an application to not request lookup of an out of bounds fi_addr.

### DIFF
--- a/fabtests/unit/av_test.c
+++ b/fabtests/unit/av_test.c
@@ -619,32 +619,24 @@ fail:
 static int
 av_lookup_good(void)
 {
-	int testret;
-	int ret;
-	int i, j, found, offset;
-	struct fid_av *av;
-	struct fi_av_attr attr;
+	int testret, ret, i, j, found, offset;
+	struct fid_av *av = NULL;
+	struct fi_av_attr attr = {0};
 	uint8_t addrbuf[4096];
-	size_t buflen;
 	uint8_t lookup_buf[4096];
-	size_t lookup_len;
-	fi_addr_t *fi_addr;
+	size_t buflen, lookup_len;
+	fi_addr_t *fi_addr = NULL;
 
 	testret = FAIL;
-
-	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
-			 sizeof(fi_addr_t));
+	fi_addr = calloc(num_good_addr, sizeof(*fi_addr));
 	if (!fi_addr) {
-		sprintf(err_buf, "malloc fi_addr failed");
+		sprintf(err_buf, "calloc fi_addr failed");
 		ret = -FI_ENOMEM;
 		goto fail;
 	}
 
-	memset(&attr, 0, sizeof(attr));
 	attr.type = av_type;
 	attr.count = num_good_addr;
-
-	av = NULL;
 	ret = fi_av_open(domain, &attr, &av, NULL);
 	if (ret != 0) {
 		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
@@ -664,15 +656,16 @@ av_lookup_good(void)
 
 	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
 	if (ret != num_good_addr) {
-		sprintf(err_buf, "fi_av_insert ret=%d, %s", ret,
-			fi_strerror(-ret));
+		sprintf(err_buf, "fi_av_insert ret=%d, expected %d", ret,
+			num_good_addr);
 		goto fail;
 	}
 
 	for (i = 0; i < num_good_addr; i++) {
 		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
-			sprintf(err_buf, "fi_av_insert failed ret=%d, %s", ret,
-				fi_strerror(-ret));
+			sprintf(err_buf, "fi_addr[%d]=%ld, expected %d", i,
+				fi_addr[i], i);
+			goto fail;
 		}
 	}
 
@@ -681,7 +674,7 @@ av_lookup_good(void)
 		memset(lookup_buf, 0, sizeof(lookup_buf));
 		lookup_len = sizeof(lookup_buf);
 		ret = fi_av_lookup(av, fi_addr[i], &lookup_buf, &lookup_len);
-		if (ret != 0) {
+		if (ret) {
 			sprintf(err_buf, "fi_av_lookup ret=%d, %s", ret,
 				fi_strerror(-ret));
 			goto fail;
@@ -707,6 +700,7 @@ av_lookup_good(void)
 	}
 
 	testret = PASS;
+	ret = FI_SUCCESS;
 fail:
 	FT_CLOSE_FID(av);
 	free(fi_addr);


zachdworkin:
DISCUSSION:
Behavior of fi_av_lookup is undefined for is an application attempts to lookup an address outside of the AV range.
Example: Create an AV table with 32 entries and lookup fi_addr 33.

Current behavior for this case is provider specific. Most use the util av code which does not check for this error and will segfault (if on non-debug build) in ofi_mem.h:ofi_bufpool_get_ibuf:501 when attempting to access a region in the table that is out of bounds. Providers like ucx specifically check for this case and return -FI_EINVAL if the requested fi_addr is out of bounds.

The next common error case is supported where 4 addresses are inserted into an AV with max 32 entries, fi_addr 3 is removed, and then fi_addr 3 is looked up. This is handled and will return -FI_ADDR_NOTAVAIL.

The out of bounds lookup case must either be defined as unsupported in the man page, or a common error such as -FI_EINVAL or -FI_ADDR_NOTAVAIL must be returned to let an application continue if the lookup "fails". I believe returning -FI_ADDR_NOTAVAIL makes the most sense for both of these error cases so the application can handle them both similarly.

No decision will be made until @j-xiong returns from vacation but please post thoughts here.

j-xiong:
`FI_ADDR_NOTAVAIL` is not an error code. It is a special value for `fi_addr_t`.

aingerson:
I don't think there needs to be a man page distinction between "out of bounds" and not a valid fiaddr. I think they should both return the same error that indicates the fiaddr is not valid.
The application doesn't know whether something is "out of bounds" because the implementation is provider-specific. For example if an address gets removed, that index can be reused or not. That implementation is determined by the provider so the application doesn't know whether a value is within or outside of the bounds of the array.
Providers should be able to easily determine whether the value is within bounds or not for their own implementation so I don't think it's a huge ask. 
I agree that the man page needs to be expanded to include the return the application should expect but I wouldn't say it should be left as "undefined." Providers can and should handle those in the same way.

shijin-aws:
Looking at EFA code, it looks compliant for the proposal? it has the following check before accessing the av entry

```
if (OFI_LIKELY(ofi_bufpool_ibuf_is_valid(util_av->av_entry_pool, fi_addr)))
		util_av_entry = ofi_bufpool_get_ibuf(util_av->av_entry_pool, fi_addr);
	else
		return NULL;
```

and the `NULL` will finally propagate to a -FI_EINVAL return codee in efa_av_lookup

zachdworkin:
> Looking at EFA code, it looks compliant for the proposal, it has the following check before accessing the av entry
> 
> ```
> if (OFI_LIKELY(ofi_bufpool_ibuf_is_valid(util_av->av_entry_pool, fi_addr)))
> 		util_av_entry = ofi_bufpool_get_ibuf(util_av->av_entry_pool, fi_addr);
> 	else
> 		return NULL;
> ```
> 
> and the `NULL` will finally propagate to a -FI_EINVAL return codee in efa_av_lookup

Great! I'll tag you when I open the PR that adds the negative test so you can run by hand

shijin-aws:
Maybe I am wrong, I am not sure whether `ofi_bufpool_ibuf_is_valid` is enough, it seems this call has an assertion

```
static inline bool ofi_bufpool_ibuf_is_valid(struct ofi_bufpool *pool, size_t index)
{
	void *buf;
	size_t region_index = index / pool->attr.chunk_cnt;

	assert(region_index < pool->region_cnt);

	buf = pool->region_table[region_index]->mem_region +
		(index % pool->attr.chunk_cnt) * pool->entry_size;

	return ofi_buf_is_valid(buf);
}

```

And I think that assertion will fail if the index is too large


It seems UCX checked it as 

```
	ave = (struct ucx_ave *) ofi_array_at_max(&mav->addr_array, fi_addr,
						  mav->count);
	if (!ave)
		return -FI_EINVAL;

```

shijin-aws:
Maybe we should just fix ofi_bufpool_ibuf_is_valid? I do not see a strong reason we have to assert here, if `region_index >= pool->region_cnt`, we can just return false?

@aingerson @j-xiong  WDYT?

zachdworkin:
> Maybe we should just fix ofi_bufpool_ibuf_is_valid? I do not see a strong reason we have to assert here, if `region_index >= pool->region_cnt`, we can just return false?
> 
> @aingerson @j-xiong WDYT?

Changing that function is part of the changeset of the PR that adds the negative test. Still fixing other edge cases before opening it.

aingerson:
@shijin-aws Yeah I think it might be best to remove the assert from the is valid and let the caller assert it if needed depending on the use case

j-xiong:
I agree. Assertion is used for conditions that shouldn't happen. Here the value can come from user input. We should handle invalid input instead of making the assertion.

shijin-aws:
bot:aws:retest

zachdworkin:
@shijin-aws this PR isn't ready for ci to test yet. I will repush with the full changeset

man/fi_av.3.md
```diff
@@ -514,11 +514,12 @@ set when removing an authorization key fi_addr_t.
 
 This call returns the address stored in the address vector that
 corresponds to the given fi_addr.  The returned address is the same
-format as those stored by the AV.  On input, the addrlen parameter
-should indicate the size of the addr buffer.  If the actual address is
-larger than what can fit into the buffer, it will be truncated.  On
-output, addrlen is set to the size of the buffer needed to store the
-address, which may be larger than the input value.
+format as those stored by the AV.  Requesting lookup of a fi_addr that is unset
+or is outside of the provider's address vector bounds will result in EINVAL.
```
j-xiong: `will result in -FI_EINVAL being returned`

prov/util/src/util_av.c
```diff
@@ -370,8 +373,13 @@ fi_addr_t ofi_av_lookup_fi_addr(struct util_av *av, const void *addr)
...
 ofi_av_lookup_addr(struct util_av *av, fi_addr_t fi_addr, size_t *addrlen)
 {
+	void *addr = ofi_av_get_addr(av, fi_addr);
+
+	if (!addr)
+		return NULL;
+
 	*addrlen = av->addrlen;
-	return ofi_av_get_addr(av, fi_addr);
+	return addr;
```
j-xiong: This change is unnecessary and introduces extra branch. It's fine to have `addrlen` set redundantly.

prov/rxd/src/rxd_av.c
```diff
@@ -320,6 +320,9 @@ static int rxd_av_lookup(struct fid_av *av, fi_addr_t fi_addr, void *addr,
 	fi_addr_t dg_fiaddr;
 
 	rxd_av = container_of(av, struct rxd_av, util_av.av_fid);
+	if (fi_addr > ofi_av_size(&rxd_av->util_av) - 1)
+		return -FI_EINVAL;
+
 	dg_fiaddr = rxd_av_dg_addr(rxd_av, fi_addr);
 	if (dg_fiaddr == FI_ADDR_UNSPEC)
 		return -FI_ENODATA;
```
j-xiong: Should here return `-FI_EINVAL` as well?

prov/sockets/src/sock_av.c
```diff
@@ -309,15 +309,20 @@ static int sock_av_lookup(struct fid_av *av, fi_addr_t fi_addr, void *addr,
...
+	if (!av_addr->valid)
+		goto inval;
+
 	memcpy(addr, &av_addr->addr, MIN(*addrlen, (size_t)_av->addrlen));
 	ofi_mutex_unlock(&_av->table_lock);
 	*addrlen = _av->addrlen;
 	return 0;
+inval:
+	ofi_mutex_unlock(&_av->table_lock);
+	return -EINVAL;
```
j-xiong: Let's take this chance to change it to `-FI_EINVAL` here.

j-xiong:
Please add `prov/rxd: ` to the title line of commit `60fea38edebed`

include/ofi_mem.h
```diff
@@ -502,7 +503,8 @@ static inline void *ofi_bufpool_get_ibuf(struct ofi_bufpool *pool, size_t index)
 
 	buf = pool->region_table[region_index]->mem_region +
 		(index % pool->attr.chunk_cnt) * pool->entry_size;
-	assert(ofi_buf_is_valid(buf));
+	if (!ofi_buf_is_valid(buf))
```
shijin-aws: I slightly prefer we keep this assert so the logic is consistent with L502. If we want a if instead of assert for invalid index, you need such if in L502 as well. I think it is better for the caller of `ofi_bufpool_get_ibuf` to validate the buffer via `ofi_bufpool_ibuf_is_valid` before calling this function, and we have those if already in that function

zachdworkin: I changed it for this path so that it is checked but there are several other places where ofi_bufpool_get_ibuf is called without ofi_bufpool_ibuf_is_valid being called before it. Several of them are on the fi_av_remove path. Similar to the purpose of this PR there is no positive or negative testing of fi_av_remove in the av_test.c file. I think a follow up PR that defines the fi_av_remove behavior and fixes these issues is necessary. WDYT? @j-xiong @shijin-aws

prov/shm/src/smr_av.c
```diff
@@ -238,6 +238,9 @@ static int smr_av_lookup(struct fid_av *av, fi_addr_t fi_addr, void *addr,
 	char *name;
 
 	util_av = container_of(av, struct util_av, av_fid);
+	if (fi_addr > ofi_av_size(util_av) - 1)
```
aingerson: Please change all instances of this similar check in the PR's changeset to be
```
if (fi_addr >= ofi_av_size(util_av))
		return -FI_EINVAL;
```
to match logic in other places

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
 	return TEST_RET_VAL(ret, testret);
 }
 
+static int
+av_lookup_bad(void)
+{
+	int testret;
+	int ret;
+	int i;
```
aingerson: merge these

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	struct fi_av_attr attr;
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	uint8_t lookup_buf[4096];
+	size_t lookup_len;
+	fi_addr_t *fi_addr;
+
+	testret = FAIL;
+
+	fi_addr = calloc(num_good_addr, sizeof(fi_addr_t));
```
aingerson: `sizeof(fi_addr)`

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	if (!fi_addr) {
+		sprintf(err_buf, "malloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
+	}
+
+	memset(&attr, 0, sizeof(attr));
```
aingerson: move initialization to declaration ie `{0}`

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	attr.type = av_type;
+	attr.count = num_good_addr + 1;
+
+	av = NULL;
```
aingerson: remove NULL init (already initialized to NULL)

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret != 0) {
```
aingerson: `if (ret)` to match the reset of fabtests - there are multiple places in the rest of this test, please update all

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+			fi_tostr(&av_type, FI_TYPE_AV_TYPE), ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr + 1; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
```
aingerson: does this test work after changing the calloc to just num_good_addr?

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	if (ret != num_good_addr) {
+		sprintf(err_buf, "fi_av_insert ret=%d, %s", ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_av_insert failed ret=%d, %s", ret,
+				fi_strerror(-ret));
```
aingerson: should there be a return here or do you keep going on purpose?

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+		}
+	}
+
+	/* case 1: lookup addr of non-inserted location in av open range */
+	memset(lookup_buf, 0, sizeof(lookup_buf));
+	lookup_len = sizeof(lookup_buf);
+	ret = fi_av_lookup(av, (fi_addr_t) num_good_addr, &lookup_buf,
+			   &lookup_len);
+	if (ret >= 0) {
```
aingerson: should this be `if (ret != -FI_EINVAL)` to catch the correct errno?

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+			"ret=%d, should be < 0", ret);
+		goto fail;
+	}
+
+	/* case 2: lookup addr outside of av open range */
+	memset(lookup_buf, 0, sizeof(lookup_buf));
+	lookup_len = sizeof(lookup_buf);
+	ret = fi_av_lookup(av, (fi_addr_t) (num_good_addr + 32), &lookup_buf,
+			   &lookup_len);
+	if (ret >= 0) {
```
aingerson: same here

fabtests/unit/av_test.c
```diff
@@ -711,6 +713,121 @@ av_lookup_good(void)
...
+	for (i = 0; i < num_good_addr + 1; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+
+	buflen = sizeof(addrbuf);
+	ret = av_create_address_list(good_address, 0, num_good_addr, addrbuf, 0,
+				     buflen);
+	if (ret < 0)
+		goto fail;
+
+	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
```
ooststep, line 748: fi_av_insert can be partially successful, and it's return value can be greater than 0 but less than count (num_good_addr). failure is determined by ret != count rather than an error code. I think it's probably OK for this test to treat any success count (ret) less than count as failure, but it should be considered. You likely won't get a good fi_strerror from ret (and the following loop likely won't get hits).

fabtests/unit/av_test.c
zachdworkin: for some reason yes it does... ill fix it though

prov/rxd/src/rxd_av.c
```diff
@@ -320,9 +320,12 @@ static int rxd_av_lookup(struct fid_av *av, fi_addr_t fi_addr, void *addr,
 	fi_addr_t dg_fiaddr;
 
 	rxd_av = container_of(av, struct rxd_av, util_av.av_fid);
+	if (fi_addr >= ofi_av_size(&rxd_av->util_av))
```
aingerson: I don't think this does anything. It doesn't hurt but I don't think it's needed and still has an edge case at index 0 when nothing has been inserted

prov/shm/src/smr_av.c
```diff
@@ -238,6 +238,9 @@ static int smr_av_lookup(struct fid_av *av, fi_addr_t fi_addr, void *addr,
 	char *name;
 
 	util_av = container_of(av, struct util_av, av_fid);
+	if (fi_addr >= ofi_av_size(util_av))
```
aingerson: same here

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,116 @@ av_lookup_good(void)
...
+	uint8_t lookup_buf[4096];
+	size_t buflen, lookup_len;
+	fi_addr_t *fi_addr;
+
+	testret = FAIL;
+	fi_addr = calloc(num_good_addr, sizeof(*fi_addr));
+	if (!fi_addr) {
+		sprintf(err_buf, "malloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
```
aingerson, line 726: init fi_addr to NULL to make this err case work

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,116 @@ av_lookup_good(void)
...
+	attr.count = num_good_addr + 1;
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret) {
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+			fi_tostr(&av_type, FI_TYPE_AV_TYPE), ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr + 1; i++)
```
aingerson: this still isn't fixed but it's marked as resolved

fabtests/unit/av_test.c
```diff
@@ -619,32 +619,24 @@ av_insert_stages(void)
...
 
 	testret = FAIL;
-
-	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
-			 sizeof(fi_addr_t));
+	fi_addr = calloc(num_good_addr, sizeof(*fi_addr));
 	if (!fi_addr) {
 		sprintf(err_buf, "malloc fi_addr failed");
 		ret = -FI_ENOMEM;
 		goto fail;
```
aingerson, line 635: same here - init to NULL

prov/util/src/util_av.c
```diff
@@ -916,10 +919,15 @@ bool ofi_ip_av_is_valid(struct fid_av *av_fid, fi_addr_t fi_addr)
 int ofi_ip_av_lookup(struct fid_av *av_fid, fi_addr_t fi_addr,
 		     void *addr, size_t *addrlen)
 {
-	struct util_av *av =
-		container_of(av_fid, struct util_av, av_fid);
+	struct util_av *av = container_of(av_fid, struct util_av, av_fid);
 	size_t av_addrlen;
+
+	if (fi_addr >= ofi_av_size(av))
```
aingerson: this one too

ooststep (approved):

zachdworkin:
@swelch @belynam is this okay to merge?

belynam:
> @swelch @belynam is this okay to merge?

Okay by me, approved!

zachdworkin:
bot:aws:retest

zachdworkin:
@shijin-aws what is the aws ci failure? Is it related?

sunkuamzn:
bot:aws:retest

PR #11607, opened 2025-11-11
Update Cargo.toml in preparation for crates.io publishing.

kyle-yh-kim:


Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[workspace]
+members = [
+    "bindings/rust"
+]
+resolver = "3"
+
+[workspace.package]
+version = "0.1.0"
+edition = "2024"
```
aingerson, line 17: 2025?

aingerson (commented):
Just a quick question but otherwise looks good to me! Thanks!

Cargo.toml
j-xiong: This is the edition of Rust and is set to 2024 by default with the `cargo new` command.

PR #11615, opened 2025-11-12
fabtests/efa: Use oob_port in multi_ep_stress 

jiaxiyan:
When multiple tests run in parallel, bind could fail because they
use the same hardcoded port number. Use the oob_port option to avoid
the port conflict.


alekswn (approved):

shijin-aws:
can we use the `-E=<port_id>` for this test as well?

Issue #11499, opened 2025-10-14
prov/efa: Potential atomicity issue with ofi_copy_to_iov() for NCCL_PROTO=LL128 protocol support


zhou-yukun:
In the context of NCCL_PROTO=LL128, which requires atomicity guarantees for received data within 128-byte boundaries, I've identified a potential issue in the current implementation.

In the function efa_rdm_pke_copy_payload_to_ope(), when handling non-CUDA and non-HMEM memory types (specifically system memory), the code uses ofi_copy_to_iov for data copying:

**bytes_copied = ofi_copy_to_iov(ope->iov, ope->iov_count, segment_offset + ep->msg_prefix_size, pke->payload, pke->payload_size);**

The ofi_copy_to_iov function ultimately calls memcpy, which may not guarantee atomicity for 128-byte data transfers. This could potentially violate the atomicity requirements of the LL128 protocol.

Questions:
1. Is this a genuine concern for LL128 protocol compliance?
2. Should system memory copies also have atomicity guarantees for 128-byte boundaries?
3. If this is indeed an issue, what would be the recommended approach to ensure atomicity for system memory copies in LL128 scenarios?

Thank you for your attention to this issue.

shijin-aws:
@zhou-yukun Thanks for reporting the issue. Today [ofi nccl plugin](https://github.com/aws/aws-ofi-nccl/blob/0cd2841263e8b24f4ecb35f92db13a7470e81ad6/src/platform-aws.cpp#L772-L774) (where libfabric is accessed by NCCL) is using FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES/FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES to enforce 128 bytes atomicity from libfabric. Today EFA provider only apply this option to cuda memory transfer (by copying the data from internal bounce buffer to the dest cuda memory via a local rdma read which guarnateed that 128 bytes atomicity), but according to https://github.com/ofiwg/libfabric/blob/5b39be88b1730bec37ba0aaaa7ebd80803276b65/man/fi_efa.7.md?plain=1#L154-L160, this option shouldn't be restricted to cuda. So we need to fix it for other memory types

zhou-yukun:
您好!您的邮件已收到!我查阅后会尽快回复您!

zhou-yukun:
@shijin-aws Thank you for your reply. I apologize for the delayed response due to recent personal matters.

[...]

PR #11616, opened 2025-11-12
prov/efa: Temporarily disable FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES support for efa protocol

shijin-aws:
Currently for efa protocol path, FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES option is only guaranteeing
    the data copy atomicity from the host buffer to cuda as destination buffer. This atomicity is not guaranteed
    for other dst memory types like host and neuron, which uses memcpy or nrt_memcpy to copy data.

Temporarily disable this option before the support is fully fixed.


shijin-aws:
related: https://github.com/ofiwg/libfabric/issues/11499

bwbarrett (approved):

PR #11620, opened 2025-11-13
Revert "prov/efa: Remove unneeded dlist init"

sunkuamzn:
This reverts commit d6d4a210eb9b45958e3c73c5a621b9b91354881a.


shijin-aws:
can u add the reason of the revert? it's because dlist_init is required for a equipotential dlist_remove happened in the error path.

PR #11621, opened 2025-11-14
prov/efa: Determine whether using device rdma based on p2p

shijin-aws:
When p2p is not available, we cannot use rdma write, as the buffer is not really registered to the device and we need to use emulation protocols in this case.


jiaxiyan:
Typo in your commit `Determine`

shijin-aws:
> Typo in your commit `Determine`

fixed

shijin-aws:
@jiaxiyan Fixed some unit-test issue, can u review again

jiaxiyan (approved):

PR #11624, opened 2025-11-14
prov/efa: Add tracepoints for handshake

jiaxiyan:
Record trigger_handshake_begin in efa_rdm_ep_enforce_handshake_for_txe, rtw_send_completion in efa_rdm_pke_handle_eager_rtw_send_completion, handshake_recv_completion in efa_rdm_pke_handle_handshake_recv, rtw_recv_completion in efa_rdm_cq_handle_recv_completion, post_handshake_begin in efa_rdm_ep_post_handshake, handshake_send_completion in efa_rdm_pke_handle_send_completion.


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -476,6 +476,12 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 		pkt_entry->peer->is_local = 0;
 	}
 
+
+	base_hdr = efa_rdm_pke_get_base_hdr(pkt_entry);
+	pkt_type = base_hdr->type;
+	if (pkt_type == EFA_RDM_EAGER_RTW_PKT)
```
shijin-aws: It is possible that pkt may not have hdr, use `efa_rdm_pkt_type_of_pke`

shijin-aws: Also we can make the whole if in `#if LTTNG` to save instructions for release build

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -476,6 +476,11 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 		pkt_entry->peer->is_local = 0;
 	}
 
+#if HAVE_LTTNG
+	if (efa_rdm_pkt_type_of_pke(pkt_entry) == EFA_RDM_EAGER_RTW_PKT)
+		efa_rdm_tracepoint(rtw_recv_completion, (size_t) pkt_entry);
```
shijin-aws: Thinking about it more, it is not scaled to have such pkt type specific TP. We should trace it for general purpose.  I would make this part of the event that we poll from the device cq

shijin-aws: It's good to keep handshake_trigger and other handshake* named TPs as we want to keep this mechanism for a long time

jiaxiyan: you mean tracking it with `efa_rdm_tracepoint(poll_cq, (size_t) wr_id);`?

shijin-aws: yes

shijin-aws: Is it implemented in this PR yet?

jiaxiyan: We already have `efa_rdm_tracepoint(poll_cq, (size_t) wr_id);`  so I removed the tracepoints for rtw pkt

shijin-aws: You can rely on wr id to measure the delay from the trigger_handshake, to the rtw pkt post to rtw pkt completion?

shijin-aws: I thought you need pkt type in the TPs otherwise you don't know which is which

PR #11629, opened 2025-11-18
prov/efa: Grow efa_tx_pkt_pool and ope_pool during rdm ep creation

jiaxiyan:
The memory pool allocation of efa_tx_pkt_pool and ope_pool are happening during the first handshake and add 6ms latency overhead to the communication. Fix it by eagerly allocating the pool during ep creation.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep_fiops.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_fiops.c
@@ -127,6 +127,9 @@ int efa_rdm_ep_create_pke_pool(struct efa_rdm_ep *ep,
 }
 
 /** @brief initializes the various buffer pools of EFA RDM endpoint.
+ * Grow the pools to avoid memory allocations during the first communication
+ * that add latency overhead to the fast path. RX pools growth is delayed until
+ * the first fi_cq_read call in efa_rdm_ep_grow_rx_pools.
  *
  * called by efa_rdm_ep_open()
  *
@@ -151,6 +154,10 @@ int efa_rdm_ep_create_buffer_pools(struct efa_rdm_ep *ep)
 		EFA_RDM_BUFPOOL_ALIGNMENT,
 		0,
 		&ep->efa_tx_pkt_pool);
+	if (ret)
+		goto err_free;
+
+	ret = ofi_bufpool_grow(ep->efa_tx_pkt_pool);
 	if (ret)
 		goto err_free;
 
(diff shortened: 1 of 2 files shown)


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -154,6 +154,10 @@ int efa_rdm_ep_create_buffer_pools(struct efa_rdm_ep *ep)
 	if (ret)
 		goto err_free;
 
+	ret = ofi_bufpool_grow(ep->efa_tx_pkt_pool);
```
sunkuamzn, line 160: Need a comment to explain why only these specific pools are grown during endpoint creation

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -127,6 +127,9 @@ int efa_rdm_ep_create_pke_pool(struct efa_rdm_ep *ep,
 }
 
 /** @brief initializes the various buffer pools of EFA RDM endpoint.
+ * Grow the pools to avoid memory allocations during the first communication that
+ * add latency overhead to the fast path. 
+ * RX pools growth are delayed to efa_rdm_ep_grow_rx_pools.
```
shijin-aws: Nit: since you are here, also mention efa_rdm_ep_grow_rx_pools happens in the 1st fi_cq_read

jiaxiyan:
I will rebase on https://github.com/ofiwg/libfabric/pull/11631 after it is merged to fix the fabtests failure

shijin-aws:
> I will rebase on #11631 after it is merged to fix the fabtests failure

how does the fabtests change relate to this PR?

jiaxiyan:
> how does the fabtests change relate to this PR?

The packets are sent faster now and it reveals a race condition in the fabtests where receiver evict the implicit AV entry and send completions never arrive.

sunkuamzn:
> > how does the fabtests change relate to this PR?
> 
> The packets are sent faster now and it reveals a race condition in the fabtests where receiver evict the implicit AV entry and ~~send~~ receive completions never arrive.

It's the receive completions that are not generated

Before https://github.com/ofiwg/libfabric/pull/11631, multiple endpoints would send at the same time. For medium protocol, it's possible that the packets are re-ordered in such a way that the implicit AV entry created for the earlier packet is evicted before the remaining packets arrive. Waiting for the send completion makes sure the receiver has finished processing all of the packets.

PR #11635, opened 2025-11-19
util: Handle signal interrupts in uffd memory monitor

### LINKED ISSUE #11627: prov/uffd: The uffd memory cache monitor cannot handle signal intterupts
If using the uffd memory cache monitor, it appears as though sending a returning signal, such as SIGPROF, to the thread that is running `ofi_uffd_handler` will cause it to exit early.

In practice, this issue has been seen to cause NCCL programs run with profilers which rely on signals to deadlock/hang/deadlock when running with `FI_MR_CACHE_MONITOR=userfaultfd` across multiple nodes.

This is likely due to the handling of system calls such as `poll` and `read`, which have a chance to break early:
```
ret = poll(fds, 2, -1);
if (ret < 0 || fds[1].revents)
	break;
...
ret = read(uffd.fd, &msg, sizeof(msg));
if (ret != sizeof(msg)) {
	pthread_rwlock_unlock(&mm_list_rwlock);
	if (errno != EAGAIN)
		break;
```
These calls might need expanding to handle the `EINTR` case.

**To Reproduce**
[thread_profiler.tar.gz](https://github.com/user-attachments/files/23586300/thread_profiler.tar.gz)

Compile the provided `thread_profiler` library by running `make -f thread-profiler.makefile`, this will create a library which intercepts `pthread_create` calls to setup signal timers which send SIGPROF for those spawned threads.

If you run a program with libfabric and preloading this library via `LD_PRELOAD=libthreadprofiler.so`, you will be able to see that the thread which spawns `ofi_uffd_handler` will exit early. This can be observed by observing the threads in `GDB` or running an `strace` on the application.

**Expected behavior**
The thread is able to gracefully handle signal interrupts.

**Output**
Here is the `strace` output after a SIGPROF signal is sent:
```
86483 14:52:11.041305 <... ppoll resumed>) = ? ERESTARTNOHAND (To be restarted if no handler) <0.205866>
86483 14:52:11.041328 --- SIGPROF {si_signo=SIGPROF, si_code=SI_TIMER, si_timerid=0, si_overrun=0, si_int=9046592, si_ptr=0x8a0a40} ---
86483 14:52:11.041363 rt_sigreturn({mask=[]}) = -1 EINTR (Interrupted system call) <0.000008>
86483 14:52:11.041413 madvise(0x4000412f0000, 1966080, MADV_DONTNEED) = 0 <0.000012>
86483 14:52:11.041457 exit(0)           = ?
```

**Environment:**
Seen occurring on a Linux HPE Slingshot system with both `MPICH` and `CrayMPICH`. Confirmed to occur with Libfabric version `2.3.1` and `2.2.0`.

j-xiong:
If unhandled, a signal such as SIGPROF sent to the thread that is running ofi_uffd_handler will cause it to exit early.

Fix #11627

### DIFF
--- a/prov/util/src/uffd_mem_monitor.c
+++ b/prov/util/src/uffd_mem_monitor.c
@@ -108,6 +108,9 @@ static void *ofi_uffd_handler(void *arg)
 
 	for (;;) {
 		ret = poll(fds, 2, -1);
+		if (ret < 0 && errno == EINTR)
+			continue;
+
 		if (ret < 0 || fds[1].revents)
 			break;
 
@@ -117,7 +120,7 @@ static void *ofi_uffd_handler(void *arg)
 		if (ret != sizeof(msg)) {
 			pthread_mutex_unlock(&mm_lock);
 			pthread_rwlock_unlock(&mm_list_rwlock);
-			if (errno != EAGAIN)
+			if (errno != EAGAIN && errno != EINTR)
 				break;
 			continue;
 		}


j-xiong:
@swelch Could you or someone else from HPE review this PR? The authors that made most recent changes to this file were from HPE but they have moved or retired.

prov/util/src/uffd_mem_monitor.c
```diff
@@ -117,7 +120,7 @@ static void *ofi_uffd_handler(void *arg)
 		if (ret != sizeof(msg)) {
 			pthread_mutex_unlock(&mm_lock);
 			pthread_rwlock_unlock(&mm_list_rwlock);
-			if (errno != EAGAIN)
+			if (errno != EAGAIN && errno != EINTR)
```
joscot-linaro, line 123: This is **not** a review but thought it was something worth raising. This is from the `man` page of `read`:

> On success, the number of bytes read is returned (zero indicates end of file), and the file position is advanced by this  number.
       It  is not an error if this number is smaller than the number of bytes requested; this may happen for example because fewer bytes
       are actually available right now (maybe because we were close to end-of-file, or because we are reading from a pipe,  or  from  a
       terminal), or because read() was interrupted by a signal.

As far as I can tell, there a chance that the size/amount read from the syscall might not exactly equal if interrupted by a signal. I do not know much about `libfabric` to know how this might affect this case but could there be a chance that a signal sent here could allow for only a partial read? Would a partial read here potential affect future reads in the loop?

j-xiong: That can be an issue but according to the man page:
```
       If a read() is interrupted by a signal before it reads any data, it shall return −1 with errno set to [EINTR].

       If a read() is interrupted by a signal after it has successfully read some data, it shall return the number of bytes read.
```
Partial read would not set errno to EINTR so the thread would exit early as before.

Let me see if I can find a good way to handle this situation.

j-xiong: According to the userfaultfd man page https://www.man7.org/linux/man-pages/man2/userfaultfd.2.html, reading from the fd returns 1 or more msg structures based on available events and the buffer size. Passing a buffer smaller than the msg structure would return error. This implies that the partial reading scenario is not going to happen (otherwise application has no way to recover).

j-xiong:
Intel CI failure is an unrelated known issue (Bad file descriptor).

PR #11644, opened 2025-11-20
prov/efa: Adjust logging level for txe releases

shijin-aws:
It is legitimate for applications to close endpoint before resolving all tx completions. Change the txe releasing log from warning to info level


shijin-aws:
bot:aws:retest

PR #11645, opened 2025-11-20
util: Use kdreg2 instead of memhooks as default when available

j-xiong:
The kdreg2 memory monitor has less issue than memhooks. Use kdreg2 by default if it is available.


shijin-aws:
does kdgreg2 require opt-in today? i.e. explicity build like `--with-kdgreg2`

j-xiong:
@shijin-aws No, it is opt-out with `--disable-kdreg2`.

shijin-aws:
It seems today it is only available on HPE slingshot platform. So in practice it still needs "onboarding" as well?

j-xiong:
In practice, the behavior will remain unchanged for non-Slingshot platform.

j-xiong:
Intel CI failure is a known issue (udp) and is unrelated.

shijin-aws (approved):

PR #11664, opened 2025-11-25
prov/ucx: Fix completion flags for RMA operations

### LINKED ISSUE #11650: prov/ucx: remote read completions has FI_COMPLETION instead of FI_RMA
**Describe the bug**
When doing an `fi_read` call, and checking the completion queue (`fi_cq_read`), we receive a completion with `flags = 0x1000000`, i.e., `FI_COMPLETION` instead of the expected `FI_RMA` and maybe `FI_REMOTE_READ` as per the documentation here https://ofiwg.github.io/libfabric/v2.3.1/man/fi_cq.3.html 

**To Reproduce**
1. Do a remote read call `fi_read`
2. Check and read completion queue `fi_cq_read`
3. Check the flags of the received completion 

**Expected behavior**
Flags should be `FI_RMA` and maybe `FI_REMOTE_READ` is also set.

j-xiong:
Add missing FI_RMA, FI_WRITE, FI_READ flags. Remove FI_COMPLETION.

Fix #11650

### DIFF
--- a/prov/ucx/src/ucx_rma.c
+++ b/prov/ucx/src/ucx_rma.c
@@ -64,8 +64,9 @@ static ssize_t ucx_write(struct fid_ep *ep, const void *buf, size_t len,
 		.context = context,
 		.data = 0,
 	};
+	uint64_t flags = u_ep->ep.tx_op_flags | FI_RMA | FI_WRITE;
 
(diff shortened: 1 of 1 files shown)


j-xiong:
Intel CI failure is a known issue that is unrelated (tcp `Bad file descriptor`).

PR #11665, opened 2025-11-25
prov/ucx: Fix multi-threading support

### LINKED ISSUE #11651: prov/ucx: Crashes when multiple threads do fabric calls even with  FI_THREAD_SAFE
**Describe the bug**
Even with `fi->domain_attr->threading = FI_THREAD_SAFE`, whenever another thread makes a fabric call, the application crashes with 

` ucp_ep.c:1808 Assertion `ucs_async_check_owner_thread(&(worker)->async)' failed`

**To Reproduce**
1. Create a ucx endpoint using one thread.
2. Create another pthread,
3. Try to make any fabric call from the other thread, e.g., `fi_av_remove` or `fi_getname`

**Expected behavior**
With FI_THREAD_SAFE set, calls should go through from multiple threads as libfabric should be doing locking properly. 

**Output**
Crash with 
` ucp_ep.c:1808 Assertion `ucs_async_check_owner_thread(&(worker)->async)' failed`

libfabric 2.3.1, ucx 1.19

j-xiong:
The threading mode was set to `FI_THREAD_SAFE`, but the UCX threading mode was set to `UCS_THREAD_MODE_SINGLE` which only allowed a single thread to access UCX objects. Trying to access an EP (which hosts the UCX worker) from multiple threads may cause the following error:

`ucp_ep.c:1808 Assertion 'ucs_async_check_owner_thread(&(worker)->async)' failed`

Changes made:

(1) Allow threading mode be set based on application hints.

(2) Use proper UCX threading mode based on OFI threading mode.

(3) Add a runtime parameter to force single thread mode. This is for the
    purpose of evaluating the performance impact of supporting multiple
    threads since the default setting has changed.

Fix #11651

### DIFF
--- a/man/fi_ucx.7.md
+++ b/man/fi_ucx.7.md
@@ -56,6 +56,9 @@ Threading
 *FI_UCX_CHECK_REQ_LEAK*
 : Check request leak (default: disabled).
 
+*FI_UCX_SINGLE_THREAD*
+: Only use a single thread for all UCX operations (default: disabled).
+
 # SEE ALSO
 
 [`fabric`(7)](fabric.7.html),
--- a/prov/ucx/src/ucx.h
+++ b/prov/ucx/src/ucx.h
@@ -104,6 +104,7 @@ struct ucx_global_descriptor{
 	int ep_flush;
 	int enable_spawn;
 	int check_req_leak;
+	int single_thread;
 };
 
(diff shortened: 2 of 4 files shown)


j-xiong:
Intel CI failure is a known issue that is unrelated (tcp `Bad file descriptor`).

PR #11668, opened 2025-11-26
prov/efa: Fail efa_rdm_ep_open when user passes FI_RX_CQ_DATA

jiaxiyan:
efa protocol does not support FI_RX_CQ_DATA.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ep.h
+++ b/prov/efa/src/rdm/efa_rdm_ep.h
@@ -273,27 +273,6 @@ void efa_rdm_ep_post_internal_rx_pkts(struct efa_rdm_ep *ep);
 
 int efa_rdm_ep_bulk_post_internal_rx_pkts(struct efa_rdm_ep *ep);
 
-/**
- * @brief return whether this endpoint should write error cq entry for RNR.
- *
- * For an endpoint to write RNR completion, two conditions must be met:
- *
- * First, the end point must be able to receive RNR completion from rdma-core,
- * which means rnr_etry must be less then EFA_RNR_INFINITE_RETRY.
- *
- * Second, the app need to request this feature when opening endpoint
- * (by setting info->domain_attr->resource_mgmt to FI_RM_DISABLED).
- * The setting was saved as efa_rdm_ep->handle_resource_management.
- *
- * @param[in]	ep	endpoint
- */
-static inline
-bool efa_rdm_ep_should_write_rnr_completion(struct efa_rdm_ep *ep)
-{
-	return (ep->base_ep.rnr_retry < EFA_RNR_INFINITE_RETRY) &&
-		(ep->handle_resource_management == FI_RM_DISABLED);
-}
-
 /*
  * @brief: check whether we should use p2p for this transaction
  *


prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -513,6 +513,11 @@ int efa_rdm_ep_open(struct fid_domain *domain, struct fi_info *info,
 	int ret, retv, i;
 	enum fi_hmem_iface iface;
 
+	if (info->mode & FI_RX_CQ_DATA) {
```
shijin-aws, line 516: We shouldn't fail, we should ignore it

prov/efa/src/efa_base_ep.c
```diff
@@ -341,8 +341,14 @@ static int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
 
 	assert(tx_cq->unsolicited_write_recv_enabled == rx_cq->unsolicited_write_recv_enabled);
 
-	/* If user intend to post rx buffer for cq data, we shouldn't enable unsolicited write recv */
-	use_unsolicited_write_recv = tx_cq->unsolicited_write_recv_enabled && !(base_ep->info->mode & FI_RX_CQ_DATA);
+	if (EFA_INFO_TYPE_IS_DIRECT(base_ep->info)) {
+		/* If user intend to post rx buffer for cq data, we shouldn't enable unsolicited write recv */
+		use_unsolicited_write_recv = tx_cq->unsolicited_write_recv_enabled && !(base_ep->info->mode & FI_RX_CQ_DATA);
```
shijin-aws: I think the logic would be: unsolicited write recv should always be turned off when RM is enabled, and for efa-direct, it has to work with FI_RX_CQ_DATA together. So if RM_ENABLED is on while RX_CQ_DATA is off, we should fail

prov/efa/src/rdm/efa_rdm_ep_fiops.c
bwbarrett: Are you sure we should ignore it?  We should definitely clear it on return from getinfo().  But at ep_open() point, it seems like the upper layer needs to know one way or the other and silently ignoring it seems wrong.  But the man page is unhelpfully unclear here.

prov/efa/src/efa_user_info.c
```diff
@@ -412,9 +412,8 @@ int efa_user_info_alter_rdm(int version, struct fi_info *info, const struct fi_i
 					"EFA provider will now switch to using FI_AV_TABLE.\n");
 	info->domain_attr->av_type = FI_AV_TABLE;
 
-	if (!hints || !hints->domain_attr ||
-	    hints->domain_attr->resource_mgmt == FI_RM_UNSPEC)
-		info->domain_attr->resource_mgmt = FI_RM_ENABLED;
+	if (!hints || !hints->domain_attr)
+		info->domain_attr->resource_mgmt = FI_RM_UNSPEC;
```
bwbarrett: I don't think this is legal.  UNSPEC is a valid input value, but does not appear to be a valid output value.  I think the previous conditional was correct, but we need to default to `FI_RM_DISABLED`.

prov/efa/src/efa_base_ep.c
```diff
@@ -341,8 +341,14 @@ static int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
```
bwbarrett, line 341: As a nit on the `Disable unsolicited write recv with FI_RM_ENABLED` commit message, while disabling unsolicited write is necessary for preventing CQ overruns, it is not sufficient for doing so.  It's just one piece of a larger puzzle, and that should be called out in the commit message.

prov/efa/src/efa_base_ep.c
bwbarrett: +1

prov/efa/src/efa_user_info.c
shijin-aws: There are some code in protocol path today that uses this bit to determine whether we retry when hitting RNR. If it is disabled, it will generate cq error instead of retry. Today when application doesn't specify it, it retries. I do not think we want to change this behavior.  It looks both MPICH and Ompi explicitly request RM_ENABLED, but ofi-nccl plugin doesn't.

shijin-aws: I agree that using UNSPEC is not a good output, just want to evaluate the impact if we default to DISABLED

bwbarrett: Ok, then return ENABLED (as much as I dislike that option).  But returning UNSPEC does not appear to be a valid choice.

PR #11675, opened 2025-11-27
prov/efa: Introduce ep option FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV

jiaxiyan:
This option only applies to the fi_setopt() call.
It is used to disable unsolicited write recv for this endpoint, which can reduce
the likelihood of CQ overflow. The default value is true.
For efa-direct, FI_RX_CQ_DATA is required when FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV
is false, or it will return -FI_EOPNOTSUPP for the call to fi_setopt().

### DIFF
--- a/include/rdma/fi_ext.h
+++ b/include/rdma/fi_ext.h
@@ -77,6 +77,7 @@ enum {
 	FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 	FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 	FI_OPT_EFA_HOMOGENEOUS_PEERS,   /* bool */
+	FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV,     /* bool */
 };
 
 struct fi_fid_export {
--- a/prov/efa/src/efa_base_ep.c
+++ b/prov/efa/src/efa_base_ep.c
@@ -341,8 +341,17 @@ static int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
 
 	assert(tx_cq->unsolicited_write_recv_enabled == rx_cq->unsolicited_write_recv_enabled);
 
-	/* If user intend to post rx buffer for cq data, we shouldn't enable unsolicited write recv */
-	use_unsolicited_write_recv = tx_cq->unsolicited_write_recv_enabled && !(base_ep->info->mode & FI_RX_CQ_DATA);
+	if (EFA_INFO_TYPE_IS_DIRECT(base_ep->info)) {
+		/* If user intend to post rx buffer for cq data, we shouldn't
+		 * enable unsolicited write recv */
+		use_unsolicited_write_recv =
+			tx_cq->unsolicited_write_recv_enabled && !(base_ep->info->mode & FI_RX_CQ_DATA);
+	} else {
+		/* RDM full protocol doesn't support FI_RX_CQ_DATA. 
+		 * Set FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV to false to disable unsolicited write recv. */
+		use_unsolicited_write_recv =
+			tx_cq->unsolicited_write_recv_enabled && base_ep->use_unsolicited_write_recv;
+	}
 	EFA_INFO(FI_LOG_EP_CTRL, "creating QP with unsolicited write recv status: %d\n", use_unsolicited_write_recv);
 	ret = efa_qp_create(&base_ep->qp, &attr_ex, base_ep->info->tx_attr->tclass,
 			    use_unsolicited_write_recv);
@@ -517,6 +526,7 @@ int efa_base_ep_construct(struct efa_base_ep *base_ep,
 	/* TODO: update inject_rma_size to inline size after firmware
 	 * supports inline rdma write */
 	base_ep->inject_rma_size = 0;
+	base_ep->use_unsolicited_write_recv = true;
 	return 0;
 }
 
--- a/prov/efa/src/efa_ep.c
+++ b/prov/efa/src/efa_ep.c
@@ -171,6 +171,17 @@ static int efa_ep_setopt(fid_t fid, int level, int optname, const void *optval, 
 	/* no op as efa direct ep will not handshake with peers */
 	case FI_OPT_EFA_HOMOGENEOUS_PEERS:
 		break;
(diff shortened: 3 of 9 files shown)


include/rdma/fi_ext.h
```diff
@@ -77,6 +77,7 @@ enum {
 	FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 	FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES, /* bool */
 	FI_OPT_EFA_HOMOGENEOUS_PEERS,   /* bool */
+	FI_OPT_EFA_CQ_FLOW_CONTROL,     /* bool */
```
shijin-aws: This naming is too general and may not do what it really means... turning off unsolicited write recv can be a way for cq flow control, but it doesn't solve the overflow itself. I would just name it as FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV for its exact meaning

prov/efa/src/efa_base_ep.c
```diff
@@ -341,8 +341,17 @@ static int efa_base_ep_create_qp(struct efa_base_ep *base_ep,
 
 	assert(tx_cq->unsolicited_write_recv_enabled == rx_cq->unsolicited_write_recv_enabled);
 
-	/* If user intend to post rx buffer for cq data, we shouldn't enable unsolicited write recv */
-	use_unsolicited_write_recv = tx_cq->unsolicited_write_recv_enabled && !(base_ep->info->mode & FI_RX_CQ_DATA);
+	if (EFA_INFO_TYPE_IS_DIRECT(base_ep->info)) {
```
shijin-aws, line 344: The comment in https://github.com/ofiwg/libfabric/pull/11668#discussion_r2562278418 still applies here. 

unsolicited write recv should always be turned off when the setopt is set to false, and for efa-direct, it has to work with FI_RX_CQ_DATA together. So if RM_ENABLED is on while RX_CQ_DATA is off, we should fail

shijin-aws: OK nvm, I see you implemented such logic in setopt already

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -297,6 +297,18 @@ int efa_rdm_ep_post_user_recv_buf(struct efa_rdm_ep *ep, struct efa_rdm_ope *rxe
 
 	assert(rxe->iov_count > 0  && rxe->iov_count <= ep->base_ep.info->rx_attr->iov_limit);
 	assert(rxe->iov[0].iov_len >= ep->msg_prefix_size);
+
+	if (ep->user_rx_pkts_posted >= ep->efa_max_outstanding_rx_ops) {
```
shijin-aws: every user_rx_pkt will consume a user_rx pkt from the pool, right ? shouldn't it be protected by that pke alloc already?

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -858,7 +870,17 @@ int efa_rdm_ep_bulk_post_internal_rx_pkts(struct efa_rdm_ep *ep)
 	if (ep->efa_rx_pkts_to_post < MIN(efa_env.internal_rx_refill_threshold, efa_rdm_ep_get_rx_pool_size(ep)))
 		return 0;
 
-	assert(ep->efa_rx_pkts_to_post + ep->efa_rx_pkts_posted <= ep->efa_max_outstanding_rx_ops);
+	if (ep->efa_rx_pkts_to_post + ep->efa_rx_pkts_posted > ep->efa_max_outstanding_rx_ops) {
+		EFA_WARN(FI_LOG_EP_CTRL,
```
shijin-aws: If we assert this won't happen, that means we expect to have protection somewhere else?

prov/efa/src/rdm/efa_rdm_msg.c
```diff
@@ -937,6 +937,17 @@ ssize_t efa_rdm_msg_generic_recv(struct efa_rdm_ep *ep, const struct fi_msg *msg
 
 	efa_rdm_tracepoint(recv_begin_msg_context, (size_t) msg->context, (size_t) msg->addr);
 
+	if (ep->efa_rx_pkts_to_post + ep->efa_rx_pkts_posted > ep->efa_max_outstanding_rx_ops) {
```
shijin-aws: This call doesn't necessarily post pkts to efa device so it doesn't make sense to check.  If it is not a zcpy receive, it is only adding the rx buffer to the util rx queue, the device level receive is posted by efa_rdm_ep_bulk_post_internal_rx_pkts

prov/efa/src/efa_ep.c
```diff
@@ -171,6 +171,17 @@ static int efa_ep_setopt(fid_t fid, int level, int optname, const void *optval,
 	/* no op as efa direct ep will not handshake with peers */
 	case FI_OPT_EFA_HOMOGENEOUS_PEERS:
 		break;
+	case FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV:
+		if (optlen != sizeof(bool))
+			return -FI_EINVAL;
+		ep->use_unsolicited_write_recv = *(bool *)optval;
```
amitrad-aws: You should set this after the check.
Othereise, you set it even though there was an error

prov/efa/test/efa_unit_test_ep.c
```diff
@@ -1877,3 +1877,69 @@ void test_efa_base_ep_disable_unsolicited_write_recv_with_rx_cq_data(struct efa_
...
+{
+	struct efa_resource *resource = *state;
+	bool optval = false;
+	
+	efa_unit_test_resource_construct_ep_not_enabled(resource, FI_EP_RDM, EFA_DIRECT_FABRIC_NAME);
+	assert_int_equal(fi_setopt(&resource->ep->fid, FI_OPT_ENDPOINT,
+				   FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV, &optval,
+				   sizeof(optval)),
+			 -FI_EOPNOTSUPP);
+}
```
amitrad-aws, line 1920: You should berify here it is still enabled. That would catch the bug in the comment above

shijin-aws (approved):

shijin-aws:
bot:aws:retest

jiaxiyan:
CI failure is unrelated infra issue

jiaxiyan:
bot:aws:retest

PR #11686, opened 2025-12-02
gh-pages: Use “macOS” terminology for Apple’s operating system

matt-dbs:
At time of writing, this seems to be what Apple call the operating system formerly known as “OS X”.


j-xiong (approved):
Thanks!

PR #11689, opened 2025-12-04
man/fi_av, prov/util, prov/rxd, prov/sockets, fabtests/unit: Define fi_av_remove behavior and add unit testing for it

zachdworkin:
Note: Depends on #11605. They are based on top of those changes and I will rebase to remove them once #11605 is merged.

Define what happens when requesting removal of an address not set or out of bounds of the opened address vector to return -FI_EINVAL. Add unit testing to fabtests to test the good removal case and the bad removal cases and update several providers to enforce the behavior.

Positive Test:
Open an av with num_good_addr. Insert num_good_addr into av. Check all addr are valid. Remove num_good_addr from av. Lookup num_good_addr from av and check that they are all invalid.

Negative Test:
Setup: Open an av with num_good_addr + 1. Insert num_good_addr into av. Check that all num_good_addr are valid.

Case 1: Remove num_good_addr + 1
Expected Result: -FI_EINVAL

Case 2: Remove fi_addr UINT64_MAX
Expected Result: -FI_EINVAL

Case 3: Remove num_good_addr + 32
Expected Result: -FI_EINVAL

Case 4: Remove fi_addr 0 twice
Expected Result: -FI_EINVAL

Case 5: Remove 0 through num_good_addr where 0 was already removed.
Expected Result: -FI_EINVAL

### DIFF
--- a/fabtests/unit/av_test.c
+++ b/fabtests/unit/av_test.c
@@ -817,12 +817,196 @@ fail:
 	return TEST_RET_VAL(ret, testret);
 }
 
+static int
+av_remove_good(void)
+{
+	int testret, ret, i;
+	struct fid_av *av = NULL;
+	struct fi_av_attr attr = {0};
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	fi_addr_t *fi_addr = NULL;
+
+	testret = FAIL;
+
+	fi_addr = calloc(num_good_addr, sizeof(*fi_addr));
+	if (!fi_addr) {
+		sprintf(err_buf, "calloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
+	}
+
+	attr.type = av_type;
+	attr.count = num_good_addr;
+
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret) {
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+			fi_tostr(&av_type, FI_TYPE_AV_TYPE), ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+
+	buflen = sizeof(addrbuf);
+	ret = av_create_address_list(good_address, 0, num_good_addr, addrbuf, 0,
+				     buflen);
+	if (ret < 0)
+		goto fail;
+
+	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
+	if (ret != num_good_addr) {
+		sprintf(err_buf, "fi_av_insert ret=%d, expected %d", ret,
+			num_good_addr);
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_addr[%d] expected %d, actual %ld",
+				i, i, fi_addr[i]);
+			goto fail;
+		}
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		ret = fi_av_remove(av, &fi_addr[i], 1, 0);
+		if (ret) {
+			sprintf(err_buf, "fi_av_remove ret=%d, %s", ret,
+				fi_strerror(-ret));
+			goto fail;
+		}
+	}
+
+	testret = PASS;
+	ret = FI_SUCCESS;
+fail:
+	FT_CLOSE_FID(av);
+	free(fi_addr);
+	return TEST_RET_VAL(ret, testret);
+}
+
+static int av_remove_bad(void)
+{
+	int testret, ret, i;
+	struct fid_av *av = NULL;
+	struct fi_av_attr attr = {0};
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	fi_addr_t *fi_addr = NULL;
+	fi_addr_t remove_addr;
+
+	testret = FAIL;
+	fi_addr = calloc(num_good_addr, sizeof(*fi_addr));
+	if (!fi_addr) {
+		sprintf(err_buf, "calloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
+	}
+
+	attr.type = av_type;
+	attr.count = num_good_addr + 1;
+
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret) {
+		sprintf(err_buf, "fi_av_open(%s) = %d, %s",
+			fi_tostr(&av_type, FI_TYPE_AV_TYPE), ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+
+	buflen = sizeof(addrbuf);
+	ret = av_create_address_list(good_address, 0, num_good_addr, addrbuf, 0,
+				     buflen);
+	if (ret < 0)
+		goto fail;
+
+	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
+	if (ret != num_good_addr) {
+		sprintf(err_buf, "fi_av_insert ret=%d, exepected %d", ret,
+			num_good_addr);
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_addr[%d] expected %d, actual %ld",
+				i, i, fi_addr[i]);
+			goto fail;
+		}
+	}
+
+	/* Case 1: Remove num_good_addr + 1 */
+	remove_addr = (fi_addr_t) (num_good_addr + 1);
+	ret = fi_av_remove(av, &remove_addr, 1, 0);
+	if (ret != -FI_EINVAL) {
+		sprintf(err_buf, "fi_av_remove num_good_addr + 1 ret=%d, "
+			"expected %d", ret, -FI_EINVAL);
+		goto fail;
+	}
+
+	/* Case 2: Remove UINT64_MAX */
+	remove_addr = (fi_addr_t) UINT64_MAX;
+	ret = fi_av_remove(av, &remove_addr, 1, 0);
+	if (ret != -FI_EINVAL) {
+		sprintf(err_buf, "fi_av_remove UINT_MAX ret=%d, expected %d",
+			ret, -FI_EINVAL);
+		goto fail;
+	}
+
+	/* Case 3: Remove num_good_addr + 32 */
+	remove_addr = (fi_addr_t) (num_good_addr + 32);
+	ret = fi_av_remove(av, &remove_addr, 1, 0);
+	if (ret != -FI_EINVAL) {
+		sprintf(err_buf, "fi_av_remove num_good_addr + 32 ret=%d, "
+			"expected %d", ret, -FI_EINVAL);
+		goto fail;
+	}
+
+	/* Case 4: Remove fi_addr 0 twice */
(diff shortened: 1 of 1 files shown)


fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	for (i = 0; i < num_good_addr + 1; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+
+	buflen = sizeof(addrbuf);
+	ret = av_create_address_list(good_address, 0, num_good_addr, addrbuf, 0,
+				     buflen);
+	if (ret < 0)
+		goto fail;
+
+	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
```
ooststep, line 748: fi_av_insert can be partially successful, and it's return value can be greater than 0 but less than count (num_good_addr). failure is determined by ret != count rather than an error code. I think it's probably OK for this test to treat any success count (ret) less than count as failure, but it should be considered.  You likely won't get a good fi_strerror from ret.

fabtests/unit/av_test.c
```diff
@@ -713,6 +713,121 @@ av_lookup_good(void)
...
+	if (ret != num_good_addr) {
+		sprintf(err_buf, "fi_av_insert ret=%d, %s", ret,
+			fi_strerror(-ret));
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
```
ooststep, line 755: I'm not sure we get a good error code for insertion failures. ret here will be the success count, but we won't get to this loop anyway if any of the insertions fail because how fi_av_insert return code is handled.

fabtests/unit/av_test.c
```diff
@@ -711,12 +713,323 @@ av_lookup_good(void)
...
+	for (i = 0; i < num_good_addr; i++)
+		fi_addr[i] = FI_ADDR_NOTAVAIL;
+
+	buflen = sizeof(addrbuf);
+	ret = av_create_address_list(good_address, 0, num_good_addr, addrbuf, 0,
+				     buflen);
+	if (ret < 0)
+		goto fail;
+
+	ret = fi_av_insert(av, addrbuf, num_good_addr, fi_addr, 0, NULL);
```
ooststep, line 859: same as above

fabtests/unit/av_test.c
```diff
@@ -711,12 +713,323 @@ av_lookup_good(void)
...
```
ooststep, line 929: again, here

fabtests/unit/av_test.c
zachdworkin: I agree that we won't get to this point if the insert failed but I also think its a good idea to explicitly check to ensure that the negative tests are properly setup. Even though we test the inserts in earlier tests, I would prefer to leave this so we don't get a false positive.

zachdworkin:
@shijin-aws @swelch @belynam These tests I am introducing enforce the -FI_EINVAL error code for bad fi_av_removes. Please make sure your providers can pass these new tests. Note: This PR is a follow up to #11605 and I will be rebasing to remove the duplicate patches once #11605 is merged

prov/util/src/util_av.c
```diff
@@ -905,7 +905,7 @@ int ofi_ip_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
 				fi_addr[i]);
 		}
 	}
-	return 0;
+	return ret;
```
aingerson: This function will return success as long as the last address was removed correctly, regardless of the rest of the addresses

fabtests/unit/av_test.c
ooststep: I have no issues with leaving this loop, while it is likely not going to get hits, it doesn't hurt. The concern was more about the use of the return code.

prov/util/src/rxm_av.c
```diff
@@ -241,11 +242,14 @@ static int rxm_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
 	for (i = count - 1; i >= 0; i--) {
 		FI_INFO(av->util_av.prov, FI_LOG_AV, "fi_addr %" PRIu64 "\n",
 			fi_addr[i]);
-		av_entry = ofi_bufpool_get_ibuf(av->util_av.av_entry_pool,
-						fi_addr[i]);
-		if (!av_entry)
+		if (!ofi_bufpool_ibuf_is_valid(av->util_av.av_entry_pool,
+					       fi_addr[i])) {
+			ret = -FI_EINVAL;
```
aingerson, line 247: Update the commit message to better reflect what this change is doing - we are not returning early on a failure

prov/sockets/src/sock_av.c
```diff
@@ -452,6 +454,8 @@ static int sock_av_remove(struct fid_av *av, fi_addr_t *fi_addr, size_t count,
 				if (conn->av_index == idx)
 					conn->av_index = FI_ADDR_NOTAVAIL;
 				ofi_idm_clear(&sock_ep->attr->av_idm, idx);
+			} else {
+				ret = -FI_EINVAL;
```
aingerson: I don't think this is quite right. This is treating an existing conn that is in use as invalid. EINVAL should only be set if the conn doesn't exist ie `!conn`, not the second condition

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+
+	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
+			 sizeof(fi_addr));
+	if (!fi_addr) {
+		sprintf(err_buf, "malloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
+	}
+
+	memset(&attr, 0, sizeof(attr));
```
aingerson: init on declaration ie `{0}`

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+	int testret, ret, i;
+	struct fid_av *av = NULL;
+	struct fi_av_attr attr;
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	fi_addr_t *fi_addr;
+
+	testret = FAIL;
+
+	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
```
aingerson: I know you probably just copied this from somewhere else but num_good_addr should always be greater than 0. See op parsing/error checking in main

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] != FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_av_remove failed fi_addr[%d] "
+				"expected %ld, actual %ld", i, FI_ADDR_NOTAVAIL,
+				fi_addr[i]);
+			goto fail;
+		}
+	}
```
aingerson: What is this checking? Does this pass any provider? fi_av_remove doesn't set the inputted fiaddr to FI_ADDR_NOTAVAIL. They will be the same value, just not valid anymore

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+{
+	int testret, ret, i;
+	struct fid_av *av = NULL;
+	struct fi_av_attr attr;
+	uint8_t addrbuf[4096];
+	size_t buflen;
+	fi_addr_t *fi_addr, remove_addr;
+
+	testret = FAIL;
+	fi_addr = calloc(num_good_addr > 0 ? num_good_addr : 1,
```
aingerson: same here

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+			 sizeof(fi_addr));
+	if (!fi_addr) {
+		sprintf(err_buf, "malloc fi_addr failed");
+		ret = -FI_ENOMEM;
+		goto fail;
+	}
+
+	memset(&attr, 0, sizeof(attr));
```
aingerson: same here

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+	attr.type = av_type;
+	attr.count = num_good_addr + 1;
+
+	ret = fi_av_open(domain, &attr, &av, NULL);
+	if (ret != 0) {
```
aingerson: update ret checks to `if (ret)`

fabtests/unit/av_test.c
```diff
@@ -825,12 +825,206 @@ av_lookup_bad(void)
...
+	if (ret != num_good_addr) {
+		sprintf(err_buf, "fi_av_insert ret=%d, exepected %d", ret,
+			num_good_addr);
+		goto fail;
+	}
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_addr[%d] expected %ld, actual %ld",
+				i, FI_ADDR_NOTAVAIL, fi_addr[i]);
```
aingerson: expected value after insert should be `i` not `FI_ADDR_NOTAVAIL`

prov/sockets/src/sock_av.c
```diff
@@ -445,6 +447,10 @@ static int sock_av_remove(struct fid_av *av, fi_addr_t *fi_addr, size_t count,
 		for (i = 0; i < count; i++) {
 			idx = (uint16_t)(fi_addr[i] & sock_ep->attr->av->mask);
 			conn = ofi_idm_lookup(&sock_ep->attr->av_idm, idx);
+			if (!conn) {
+				ret = -FI_EINVAL;
+				continue;
+			}
 			if (conn && conn != SOCK_CM_CONN_IN_PROGRESS) {
```
j-xiong: The `conn && ` part can be removed.

shijin-aws:
bot:aws:retest

zachdworkin:
@shijin-aws @belynam @swelch @j-xiong The dependency for this PR has been merged. I have restarted testing of these changes and they are ready for review

prov/util/src/rxm_av.c
```diff
@@ -225,6 +225,7 @@ static int rxm_av_remove(struct fid_av *av_fid, fi_addr_t *fi_addr,
 	struct dlist_entry *item;
 	struct rxm_av *av;
 	ssize_t i;
+	int ret = FI_SUCCESS;
```
shijin-aws, line 228: Not related to this change but it's a bit odd that we have rxm code in prov/util. Why?

fabtests/unit/av_test.c
```diff
@@ -817,12 +817,196 @@ av_lookup_bad(void)
...
+
+	for (i = 0; i < num_good_addr; i++) {
+		if (fi_addr[i] == FI_ADDR_NOTAVAIL) {
+			sprintf(err_buf, "fi_addr[%d] expected %d, actual %ld",
+				i, i, fi_addr[i]);
+			goto fail;
+		}
+	}
+
+	/* Case 1: Remove num_good_addr + 1 */
```
shijin-aws, line 944: I think it's better to make 5 cases ran as separate testing functions which share the same setup. Having dependencies between independent tests is anti-pattern IMO

prov/util/src/rxm_av.c
j-xiong: @shijin-aws This file was moved here from `prov/rxm/src` due to a few util functions (for peer provider support) added to the file. Those functions are tightly coupled with other functions in the file.

fabtests/unit/av_test.c
j-xiong: I think it's fine to have the 5 cases ordered this way. Case 1, 2, 3 are independent but none of them update the AV and they should not impact case 4 and 5.  Case 5 actually depends on case 4.

j-xiong:
@ooststep Are you ready to update your review?

ooststep (approved):

PR #11690, opened 2025-12-04
prov/shm, tcp, fabtests: fix shm completion flags on writedata and add ubertest and cq_data test verification to catch it

aingerson:
writedata operations were missing FI_REMOTE_WRITE flag in the completion
This adds operation flags on RX CQ completions.

ubertest verification mode was missing full flag verification.
This PR also adds ubertest CQ flag verification on verification mode to make sure all correct flags are set

With fabtest change, without shm change: ubertest fails shm
With fabtest change, with shm change: ubertest now passes

### DIFF
--- a/prov/tcp/src/xnet_progress.c
+++ b/prov/tcp/src/xnet_progress.c
@@ -950,7 +950,7 @@ static int xnet_handle_write(struct xnet_ep *ep)
 		return -FI_ENOMEM;
 
 	if (ep->cur_rx.hdr.base_hdr.flags & XNET_REMOTE_CQ_DATA) {
-		rx_entry->cq_flags = (FI_COMPLETION | FI_REMOTE_WRITE |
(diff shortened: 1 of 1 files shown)


shijin-aws:
Can you help expand the cq_data.c test to check the completion flags for the fi_writedata on rx side https://github.com/ofiwg/libfabric/blob/de5dd714516ffb8e3e5d46ef029b2040c1b09460/fabtests/functional/cq_data.c#L93-L116

Today it is only checking FI_REMOTE_CQ_DATA flag

efa provider today is not running uber test (we plan to add that)

aingerson:
@shijin-aws Updated to add fi_cq_data flag testing. Both tests properly caught the error and are resolved after the shm update

j-xiong (approved):
Looks good. Thanks!

shijin-aws:
@aingerson can u backport it to v2.3.x as well? Thx!

PR #11696, opened 2025-12-05
prov/efa: Check fabric name in fi_getinfo for existing fabric/domain

jiaxiyan:
Check fabric names match if fabric/domain object is passed so only one info is returned.

### DIFF
--- a/prov/efa/src/efa_user_info.c
+++ b/prov/efa/src/efa_user_info.c
@@ -599,6 +599,10 @@ int efa_get_user_info(uint32_t version, const char *node,
 		if (ret)
 			continue;
 
+		ret = efa_prov_info_compare_fabric_name(hints, prov_info);
+		if (ret)
+			continue;
+
 		ret = efa_prov_info_compare_domain_name(hints, prov_info);
 		if (ret)
 			continue;
@@ -620,10 +624,6 @@ int efa_get_user_info(uint32_t version, const char *node,
 		dupinfo->fabric_attr->api_version = version;
 
 		if (EFA_INFO_TYPE_IS_RDM(prov_info)) {
-			if (hints && hints->fabric_attr && hints->fabric_attr->name &&
-			    strcasecmp(hints->fabric_attr->name, EFA_FABRIC_NAME))
-				continue;
-
 			ret = efa_user_info_alter_rdm(version, dupinfo, hints);
 			if (ret) {
 				fi_freeinfo(dupinfo);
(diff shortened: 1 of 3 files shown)


prov/efa/src/efa_user_info.c
```diff
@@ -110,6 +110,9 @@ int efa_user_info_check_fabric_object(const struct fi_info *hints,
 	fabric_attr.prov_version = util_fabric->prov->version;
 	fabric_attr.api_version = hints->fabric_attr->fabric->api_version;
 
+	if (strcmp(fabric_attr.name, prov_info->fabric_attr->name))
```
shijin-aws: is this  already checked as part of ofi_check_fabric_attr?

jiaxiyan: no this was a miss https://github.com/ofiwg/libfabric/blob/c271730ad56282396121b0db5b187f0ceac46eb7/prov/util/src/util_attr.c#L457

shijin-aws: Then should we fix ofi_check_fabric_attr ?

shijin-aws (dismissed):

prov/efa/src/efa_user_info.c
shijin-aws: Now I am not sure how setting fabric_attr->name in hints was working ..... we know running fabtests with `-f <fabric_name>` will work. Why do we need such change now?

jiaxiyan: > Then should we fix ofi_check_fabric_attr ?

Move this to ofi_check_fabric_attr but tcp provider fails with ` Requesting fabric 10.0.48.0/20, skipping tcp` because tcp provider's default fabric name is "tcp" but it later changes in util_set_netif_names.

jiaxiyan: > we know running fabtests with -f <fabric_name> will work.

It had issue and I opened https://github.com/ofiwg/libfabric/commit/a0d90ebb76909850c5a5f8ca8d107708cb57cd06

shijin-aws: So we shouldn't fail in the check stage, but should do such filter in the alter stage?

jiaxiyan: Do you mean checking it in ofi_alter_info? I don't think that function drops any fi_info. It just trims the returned fi_info.

shijin-aws: I am ok to fix efa provider now without making changes in common place that fail other providers

PR #11703, opened 2025-12-05
prov/tcp: move keepalive to connect done

wangdi1:
before sending the reply.


wangdi1:
@shefty @j-xiong Could you please take a look? Thanks. Somehow my previous patch put keepalive on the server side, which is wrong, and can cause some performance issue.

wangdi1:
@shefty @j-xiong  if you are ok with the patch. is it possible to back port the fix to v2.2.0?

j-xiong:
@wangdi1 Yes, it can be backported.

wangdi1:
>@wangdi1 Yes, it can be backported.

Thanks! @j-xiong. It would be great if this can be backport to v2.2.0 and later, thanks!

PR #11712, opened 2025-12-08
[v2.2.x] prov/tcp: move keepalive to connect done

j-xiong:
before sending the reply.

(cherry picked from commit a2496c1797f4629f0a5495f5519f7493313183d4)

### DIFF
--- a/fabtests/functional/multi_ep.c
+++ b/fabtests/functional/multi_ep.c
@@ -72,6 +72,9 @@ static void free_ep_res()
 	int i;
 
 	for (i = 0; i < num_eps; i++) {
+		if (!fi)
+			continue;
+
 		if (fi->domain_attr->mr_mode & FI_MR_RAW)
 			(void) fi_mr_unmap_key(domain, peer_iovs[i].key);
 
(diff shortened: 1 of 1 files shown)


j-xiong:
bot:aws:retest

aingerson (dismissed):

j-xiong:
bot:aws:retest

j-xiong:
had to cherry-pick the multi-ep fix otherwise it fails constantly.

j-xiong:
bot:aws:retest

linsword13:
Hi @j-xiong, is there plan to release a new v2.2.x version that includes this patch (same question for v2.3.x as well)? Thank you!

aingerson:
@linsword13 Jianxin is on vacation so he can comment on that when he gets back next week.
We don't usually do bug fix releases unless required. So we weren't planning on it. Do you have a limitation here that would require you to need a 2.2.x version or 2.3.x version instead of updating to 2.4.x?

j-xiong:
@linsword13 We don't have plan for a new 2.2.x or 2.3.x release just for including this fix.

PR #11714, opened 2025-12-08
[v2.4.x] prov/tcp: move keepalive to connect done

j-xiong:
before sending the reply.

(cherry picked from commit a2496c1797f4629f0a5495f5519f7493313183d4)


j-xiong:
bot:aws:retest

j-xiong:
bot:aws:retest

j-xiong:
bot:aws:retest

PR #11717, opened 2025-12-09
prov/tcp: move keepalive to connect done

wangdi1:
before sending the reply.


wangdi1:
@j-xiong The patch for 2.3.0

j-xiong:
This is duplicate of #11713

PR #11721, opened 2025-12-10
prov/util, include, src: Update getinfo logic

zachdworkin:
Set addr_format on discovery of src_addr. This will enable providers like tcp to have a properly set addr_format after an info is selected for creating an rxm_av and needing to lookup the addr_format (which previously would have been unchanged from FI_SOCKADDR_IP).

Do not allow selection of an alternate iface if the user requested a specific one. This will ensure that the user gets the iface that they requested, and prevents iface mismatches if one side selects the user specific one and the other side cannot due to driver error, network instability, etc.

Check each ifaddr entry for iface before evaluating addr, flags, and sa_family. This will skip unwanted entries faster when an application specifically requests the iface it wants.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -2119,18 +2119,17 @@ void ofi_get_list_of_addr(const struct fi_provider *prov, const char *env_name,
 		goto insert_lo;
 
 	for (ifa = ifaddrs; ifa != NULL; ifa = ifa->ifa_next) {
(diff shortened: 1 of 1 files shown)


j-xiong (approved):
LGTM, thanks!

zachdworkin:
bot:aws:retest

sunkuamzn:
bot:aws:retest

PR #11730, opened 2025-12-13
prov/opx: Don't fail configure when OPX unhappy

belynam:


prov/opx/configure.m4
```diff
@@ -307,8 +307,6 @@ AC_DEFUN([FI_OPX_CONFIGURE],[
 		])
 	])
 	CPPFLAGS=$save_CPPFLAGS
-	AS_IF([text x"$enable_opx" = x"yes" && test $opx_happy -eq 0 ],
-		[AC_MSG_ERROR([OPX failed to configure.])])
```
j-xiong, line 311: Specifying `--enable-opx=yes` indicates that `opx` provider is required and configure is expected to fail if opx cannot be built.

In order to allow opx check to fail, use `--enable-opx=auto` or just `--enable-opx` or don't specify it at all since `auto` is the default option.

belynam: Sorry, I should add some context for this change: We didn't have this check in our configure.m4 at all until our last upstream, where I saw it failing the build checks and made [this change](https://github.com/ofiwg/libfabric/pull/11681/changes/dca43d7c5351c9ece14edf09274af284a7740229) to add the condition of `$enable_opx=yes` to the if. However, there is a typo there, `text` instead of `test`, which results in the configure script printing out an error for the line, and never failing the configure when `$opx_happy=0`. I was going to fix the typo, but then thought it might be safer to just remove the check altogether since it hasn't been present in any previous release and doesn't seem to be an issue to not have it. I've gotten approval from the dev who added the check to remove it.

In its current state, the check will never fail the configure, it will just print that there's an error in the configure script on line xxxxx and move on, effectively being the same as not having the check there at all. This change just gets rid of the configure error.

If you think we should have this check in our `configure.m4`, I can fix the typo instead, but was  concerned if it would start to break other providers during their build if they didn't explicitly disable OPX in the configure options. And at this point for the v2.4.x release, I wanted to err on the side of caution.

j-xiong: Thanks for the explanation. I'll take this as it is.

PR #11743, opened 2025-12-15
prov/efa: Log ibv device open / close calls

shijin-aws:


prov/efa/src/efa_device.c
```diff
@@ -166,8 +169,9 @@ static void efa_device_destruct(struct efa_device *device)
 	if (device->ibv_ctx) {
 		err = ibv_close_device(device->ibv_ctx);
 		if (err)
-			EFA_INFO_ERRNO(FI_LOG_DOMAIN, "ibv_dealloc_pd",
+			EFA_INFO_ERRNO(FI_LOG_DOMAIN, "ibv_close_device",
```
shijin-aws, line 172: This was a bug

prov/efa/src/efa_device.c
```diff
@@ -89,6 +90,7 @@ int efa_device_construct_gid(struct efa_device *efa_device,
 
 err_close:
 	ibv_close_device(efa_device->ibv_ctx);
+	EFA_INFO(FI_LOG_CORE, "Closed ibv device for ibv_ctx %p\n", efa_device->ibv_ctx);
```
jiaxiyan: Is the context still valid after closing?

shijin-aws: The pointer value should not be overwritten by the close call, I can move the logging before the close call though

shijin-aws:
bot:aws:retest

PR #11744, opened 2025-12-16
prov/efa: Various bugfixes for EFA-direct's blocking CQ read

darrylabbate:
This smooths a few edges, but also fixes a couple critical bugs:
- Using `cond` directly for the threshold instead of dereferencing it (effectively ignores `cond` unless it's null)
- Not adjusting the timeout before polling the comp channel in each iteration.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -516,6 +516,7 @@ static ssize_t efa_cq_sreadfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 {
 	struct efa_cq *cq;
 	ssize_t ret = 0;
+	uint64_t endtime = ofi_timeout_time(timeout);
 	ssize_t threshold = 1, num_completions;
 	uint8_t *buffer;
 
@@ -537,6 +538,8 @@ static ssize_t efa_cq_sreadfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 
 	for (num_completions = 0; num_completions < threshold; ) {
 		if (efa_cq_trywait(cq) == FI_SUCCESS) {
+			if (ofi_adjust_timeout(endtime, &timeout))
+				return num_completions ? num_completions : -FI_EAGAIN;
 			/* CQ is empty, wait for events */
 			ret = efa_poll_events(cq, timeout);
 			if (ret)


prov/efa/src/efa_cq.c
```diff
@@ -519,6 +519,9 @@ static ssize_t efa_cq_sreadfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 	ssize_t threshold = 1, num_completions;
 	uint8_t *buffer;
 
+	if (OFI_UNLIKELY(!buf || !count))
```
shijin-aws: Is it documented in the man page that null buf and count is not legal for sread? EINVAL is not a recognized return value for sread today per man page

prov/efa/src/efa_cq.c
```diff
@@ -540,6 +541,8 @@ static ssize_t efa_cq_sreadfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 
 	for (num_completions = 0; num_completions < threshold; ) {
 		if (efa_cq_trywait(cq) == FI_SUCCESS) {
+			if (ofi_adjust_timeout(endtime, &timeout))
```
shijin-aws, line 541: Good catch, the code was referencing verbs provider https://github.com/ofiwg/libfabric/blob/main/prov/verbs/src/verbs_cq.c#L194 which likely has the same bug

prov/efa/src/efa_cq.c
shijin-aws: But it doesn't make sense for application to call sread with NULL and 0 as they at least wait for 1 completion

shijin-aws: I think it can just be EAGAIN, similar to the cq read call for efa-direct: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_cq.c#L676-L677

shijin-aws: I can see we return EINVAL in L529 to L536. I think we need to talk to community about whether this should be a valid rc

darrylabbate: EINVAL is the most correct in this case if we agree it's not valid to call sread with null buf or count of 0. I'm not under the impression we're violating any API contract by validating the input and returning the most logical error code.

darrylabbate: I don't want the other fixes being blocked by this, so I may drop this commit from the PR

darrylabbate: Opened #11748 for this

darrylabbate:
bot:aws:retest

darrylabbate:
will backport

PR #11748, opened 2025-12-16
prov/efa: Validate `buf` and `count` inputs in CQ sreadfrom

darrylabbate:
Moved to dedicated PR from #11744 to capture discussion re: `-FI_EINVAL`.

### DIFF
--- a/prov/efa/src/efa_cq.c
+++ b/prov/efa/src/efa_cq.c
@@ -519,6 +519,9 @@ static ssize_t efa_cq_sreadfrom(struct fid_cq *cq_fid, void *buf, size_t count,
 	ssize_t threshold, num_completions;
 	uint8_t *buffer;
 
+	if (OFI_UNLIKELY(!buf || !count))
+		return -FI_EINVAL;
+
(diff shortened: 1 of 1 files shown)


shijin-aws:
@shefty @j-xiong I can see verbs provider return ~~EINVAL~~ ENOSYS for invalid call to sread/sreadfrom (when wait object and comp channel is not available) as well, but I don't see it as a recognized return value in the man page. Do you have any feedback here

shefty:
If wait support is not available, the calls should return -ENOSYS

shijin-aws:
> If wait support is not available, the calls should return -ENOSYS

Ack, what about NULL buf/count ? We probably need to add such return codes to the man pages for sread/sreadfrom

j-xiong:
`sread/sreadfrom` should allow `count == 0` and NULL buffer (if count is 0). The behavior could be either:
(1) similar to `read/sread`; or
(2) block until something happens, but don't consume the CQE.
which way to go can be discussed.

darrylabbate:
Ack, thanks

PR #11751, opened 2025-12-17
prov/efa: Generation counter for release build

alekswn:
The purpose of this commit is to  move generation counter out of debug build mode. The commit also adds a static assertion to ensure packet enty structure would not overflow to a new cache line

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -73,6 +73,11 @@ static struct fi_ops efa_rdm_cq_fi_ops = {
 	.ops_open = fi_no_ops_open,
 };
 
+static inline void efa_rdm_cq_increment_pkt_entry_gen(struct efa_rdm_pke *pkt_entry)
+{
+	pkt_entry->gen++;
+	pkt_entry->gen &= EFA_RDM_PACKET_GEN_MASK;
+}
 
 /**
  * @brief handle rdma-core CQ completion resulted from IBV_WRITE_WITH_IMM
@@ -133,6 +138,7 @@ static void efa_rdm_cq_proc_ibv_recv_rdma_with_imm_completion(
 		 */
 		assert(pkt_entry);
 		ep->efa_rx_pkts_posted--;
+		efa_rdm_cq_increment_pkt_entry_gen(pkt_entry);
 		efa_rdm_pke_release_rx(pkt_entry);
 	}
 }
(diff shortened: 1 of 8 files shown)


alekswn:
AWS CI is failing doe to issues with wr_id to address conversion.

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -489,11 +481,7 @@ ssize_t efa_rdm_pke_sendv(struct efa_rdm_pke **pkt_entry_vec,
 		if (pkt_idx != pkt_entry_cnt - 1)
 			flags_in_loop |= FI_MORE;
 
-		wr_id = (uintptr_t) pkt_entry;
-#if ENABLE_DEBUG
-		efa_rdm_pke_increment_gen(pkt_entry);
-		wr_id += pkt_entry->gen;
-#endif
+		wr_id = (uintptr_t) pkt_entry + pkt_entry->gen;
```
sunkuamzn: `pkt_entry->gen` is not masked to 5 bits. So this addition can make `wr_id` point to a different `pkt_entry`

alekswn: Fixed that, according to my local testing this is going to fix CI failure.

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -577,6 +577,14 @@ static inline struct efa_rdm_ep *efa_rdm_cq_get_rdm_ep(struct efa_ibv_cq *cq, st
 	return container_of(base_ep, struct efa_rdm_ep, base_ep);
 }
 
+static inline void efa_rdm_cq_increment_pkt_entry_gen(struct efa_rdm_pke *pkt_entry)
+{
+	if (pkt_entry) {
```
sunkuamzn: Because this is in the fast path, we should try to limit the branches. `pkt_entry` can only be NULL for unsolicited write recv. We already have branching for different types of WCs in `efa_rdm_cq_process_wc` and we can use that branching to increment the gen counter for all WCs except for `IBV_WC_RECV_RDMA_WITH_IMM`.

For `IBV_WC_RECV_RDMA_WITH_IMM`, there is already an unsolicited write recv check [here](https://github.com/alekswn/libfabric/blob/68890131373f08eaa91bf002f21c4f4e72c46109/prov/efa/src/rdm/efa_rdm_cq.c#L128), so let's re-use that check.

prov/efa/src/rdm/efa_rdm_cq.h
```diff
@@ -21,16 +21,28 @@ int efa_rdm_cq_open(struct fid_domain *domain, struct fi_cq_attr *attr,
...
+		assert(pkt_entry->gen == gen);
+	}
+
 	return pkt_entry;
 }
-#endif
+
+static inline struct efa_rdm_pke *efa_rdm_cq_get_pke_from_wr_id(struct efa_ibv_cq *ibv_cq, uint64_t wr_id)
+{
+	if (efa_cq_wc_is_unsolicited(ibv_cq)) {
```
sunkuamzn, line 41: We're now calling `efa_cq_wc_is_unsolicited` twice - once here and once here https://github.com/alekswn/libfabric/blob/68890131373f08eaa91bf002f21c4f4e72c46109/prov/efa/src/rdm/efa_rdm_cq.c#L128

With data path direct, it is reading a flag from the CQ in host memory, so I think it won't hurt performance much

alekswn:
bot:aws:retest

alekswn:
Assertion `(uint64_t)pkt_entry == (uint64_t)pkt_entry & ~((uint64_t)EFA_RDM_PACKET_GEN_MASK)` is failing in unit-tests and pingpong.

prov/efa/src/rdm/efa_rdm_cq.c
alekswn: Changed that, will wait until CI passed to make sure I have not missed anything.

alekswn:
bot:aws:retest

PR #11752, opened 2025-12-18
prov/efa: Fix unittest build warnings

alekswn:
EFA unittests use depricated functions from CMocka. That causes a lot of build warnings.
This commit updates function according to up-to-date recommendations.


alekswn:
bot:aws:retest

PR #11756, opened 2025-12-19
prov/hook: Harden hook monitor provider initalization

alekswn:
We randomly see warning message about tick being negative from hook monitor. Valgrind detects branch depending on uninitialized value `signed_tick_max` in `monitor_env_init()`.

Looking closely at `monitor_env_init()` I noticed `signed_tick_max` is going to contain garbage if env variable `tick_max` was not defined. The garbage is going to be coped over to global variable `mon_env.tick_max`.

This PR adds handling for return values of `fi_param_*` functions and better logging for monitor hook initialization.

### DIFF
--- a/prov/hook/monitor/src/hook_monitor.c
+++ b/prov/hook/monitor/src/hook_monitor.c
@@ -1151,12 +1151,13 @@ static int monitor_env_init(void)
 			"Number of API calls before synchronisation files are checked for data request. (default: %d)",
 			mon_env.tick_max);
 
-	fi_param_get_int(prov, "tick_max", &signed_tick_max);
-	if (signed_tick_max < 0) {
-		FI_WARN(prov, FI_LOG_CORE, "Tick is negative!\n");
-		return -EOVERFLOW;
+	if (fi_param_get_int(prov, "tick_max", &signed_tick_max) == FI_SUCCESS) {
+		if (signed_tick_max < 0) {
+			FI_WARN(prov, FI_LOG_CORE, "Tick is negative!\n");
+			return -EOVERFLOW;
+		}
+		mon_env.tick_max = (unsigned int)signed_tick_max;
 	}
-	mon_env.tick_max = (unsigned int)signed_tick_max;
 
 	fi_param_define(prov, "file_mode", FI_PARAM_INT,
 			"POSIX mode/permission for synchronisation files. (default: %04o)",


prov/hook/monitor/src/hook_monitor.c
```diff
@@ -1140,49 +1140,71 @@ static int monitor_env_init(void)
...
-	char *basepath = NULL;
+	char *basepath;
+	int ret;
 
-	fi_param_define(prov, "linger", FI_PARAM_BOOL,
+	ret = fi_param_define(prov, "linger", FI_PARAM_BOOL,
 			"Whether synchronisation files should linger after termination. (default: %d)",
 			mon_env.linger);
-	fi_param_get_bool(prov, "linger", &mon_env.linger);
+	if (ret != FI_SUCCESS) return ret;
```
aingerson: Please update all these one line returns to put the return on a separate line

prov/hook/monitor/src/hook_monitor.c
```diff
@@ -1140,49 +1140,71 @@ static int monitor_env_init(void)
...
 	return 0;
 }
 
 HOOK_MONITOR_INI
 {
-	if (monitor_env_init())
+	struct fi_provider *prov = &hook_monitor_ctx.prov;
+	int ret = monitor_env_init();
+	if (ret) {
+		FI_WARN(prov, FI_LOG_CORE, "Monitor hook has failed initialization: %s\n", strerror(-ret));
```
aingerson: can we just pass in `&hook_monitor_ctx.prov` here? is there another reason for needing the prov?

aingerson (commented):
Just a couple of notes. There's also a typo in the commit message ("coped" to "copied"). Otherwise looks good. Thanks!

prov/hook/monitor/src/hook_monitor.c
alekswn: changed that

prov/hook/monitor/src/hook_monitor.c
alekswn: fixed

prov/hook/monitor/src/hook_monitor.c
```diff
@@ -1140,49 +1140,81 @@ static int monitor_env_init(void)
...
-	}
-	mon_env.tick_max = (unsigned int)signed_tick_max;
+	ret = fi_param_get_int(prov, "tick_max", &signed_tick_max);
+	if (ret == FI_SUCCESS) {
+		if (signed_tick_max < 0) {
+			FI_WARN(prov, FI_LOG_CORE, "Tick is negative!\n");
+			return -EOVERFLOW;
+		}
+		mon_env.tick_max = (unsigned int)signed_tick_max;
+	} else if (ret != -FI_ENODATA)
```
aingerson: Add parents for this clause (an else if with a corresponding multiline if should also have brackets even if is only a single line)

prov/hook/monitor/src/hook_monitor.c
```diff
@@ -1140,49 +1140,81 @@ static int monitor_env_init(void)
 {
 	struct fi_provider *prov = &hook_monitor_ctx.prov;
 	int signed_tick_max;
-	char *basepath = NULL;
+	char *basepath;
+	int ret;
 
-	fi_param_define(prov, "linger", FI_PARAM_BOOL,
+	ret = fi_param_define(prov, "linger", FI_PARAM_BOOL,
 			"Whether synchronisation files should linger after termination. (default: %d)",
```
aingerson, line 1146: Can you also change the alignment of this and the other similar lines to match the code style, please?

prov/hook/monitor/src/hook_monitor.c
```diff
@@ -1140,49 +1140,81 @@ static int monitor_env_init(void)
...
+	if (ret == FI_SUCCESS) {
+		ret = snprintf(mon_env.basepath, sizeof(mon_env.basepath), "%s", basepath);
+		if (ret < 0)
+			return ret;
+		if (ret >= sizeof(mon_env.basepath)) {
+			FI_WARN(prov, FI_LOG_CORE, "basepath is too long\n");
+			return -EOVERFLOW;
+		}
+	} else if (ret != -FI_ENODATA)
+		return ret;
```
aingerson: same here

alekswn:
> No, we don't want to return early on error from `fi_param_define` or `fi_param_get`. Error on one parameter should not stop the processing of other parameters. We just need an error output which is already been given inside these functions.
> 
> To fix the issue mentioned in the commit message, you only need to initialize `signed_tick_max` to a reasonable value (e.g. `MON_TICK_MAX_DEFAULT`) before calling `fi_param_get_int`.

I think we should leave the snippet below from this PR then:

```
	ret = fi_param_get_int(prov, "tick_max", &signed_tick_max);
	if (ret == FI_SUCCESS) {
		if (signed_tick_max < 0) {
			FI_WARN(prov, FI_LOG_CORE, "Tick is negative!\n");
			return -EOVERFLOW;
		}
		mon_env.tick_max = (unsigned int)signed_tick_max;
	} 
```

`mon_env.tick_max` is initialized to the default value already.

PR #11757, opened 2025-12-22
prov/lnx: Add CPPFLAGS to build as DSO 

Madeeks:
When attempting to build the LNX provider as DSO (i.e. using the `--enable-lnx=dl` configure flag), the build fails with errors related to missing headers and GNU macros.

Setting `CPPFLAGS` for the `liblnx-fi` object to expose the missing resources during compilation makes the build succeed.

### DIFF
--- a/prov/lnx/Makefile.include
+++ b/prov/lnx/Makefile.include
@@ -47,7 +47,7 @@ _lnx_headers = \
 if HAVE_LNX_DL
 pkglib_LTLIBRARIES += liblnx-fi.la
 liblnx_fi_la_SOURCES = $(_lnx_files) $(_lnx_headers)
-liblnx_fi_la_CPPFLAGS = -I$(top_srcdir)/prov/lnx/include -I$(top_srcdir)/include
+liblnx_fi_la_CPPFLAGS = -I$(top_srcdir)/prov/lnx/include -I$(top_srcdir)/include -D_GNU_SOURCE
 liblnx_fi_la_LIBADD = $(linkback) $(lnx_LIBS)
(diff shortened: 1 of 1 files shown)


prov/lnx/Makefile.include
```diff
@@ -47,7 +47,7 @@ _lnx_headers = \
 if HAVE_LNX_DL
 pkglib_LTLIBRARIES += liblnx-fi.la
 liblnx_fi_la_SOURCES = $(_lnx_files) $(_lnx_headers)
-liblnx_fi_la_CPPFLAGS = -I$(top_srcdir)/prov/lnx/include -I$(top_srcdir)/include
+liblnx_fi_la_CPPFLAGS = -I$(top_srcdir)/prov/lnx/include -I$(top_srcdir)/include -D_GNU_SOURCE
```
sielicki, line 50: if enabling GNU extensions is truly necessary to get a functional build, IMO the only acceptable way to do it is to add `AC_USE_SYSTEM_EXTENSIONS` at the top level. I'm opposed to mixing code with/without gnu extensions in the same project on a by-target basis. (Either this project is portable extensionless C99, or it isn't, but enabling extensions only on a select group of translation units doesn't change that reality, it just causes compatibility bugs when `struct time_t` varies in size unexpectedly between translation units... just as one example of how these small differences can bite you. Debugging issues of this kind is not easy or fun.)

Madeeks: Thanks for your feedback!
I realize the current implementation is suboptimal. I will convert this PR to Draft until I can find a more appropriate and elegant solution.

PR #11759, opened 2025-12-22
[v2.4.x] prov/efa: Use endpoint->peer hashmap at the AV level

sunkuamzn:
An fi_addr->peer  hashmap at the endpoint level grows with the number of peers. An endpoint->peer hashmap grows with the number of endpoints - which would be better for applications that create many peers but few endpoints per AV e.g. MPI applications which
only create a single endpoint per AV.

(cherry picked from commit 865ec28aef3d8fa9f37e45361e91989a30af4290)


shijin-aws:
bot:aws:retest

shijin-aws (approved):

PR #11760, opened 2025-12-22
.github/workflows: Add PR character safety check workflow

alekswn:
This PR adds a GitHub actions workflow to block PRs with non-printable unicode characters.
Tested bash scripts on a past commit with a non-printable character.

```
Running PR character safety check...
Checking for files converted from text to binary between ff97cc7efcfc303185b8bcb782eb8e3e79be3f2b~1 and ff97cc7efcfc303185b8bcb782eb8e3e79be3f2b...
Checking for non-printable characters...
Non-printable or problematic characters detected:
8:+		efa_LDFLAGS+=" -L$efa_LIBDIR -Wl,--enable-new-dtags,-rpath,$efa_LIBDIR"
00000000  2b 09 09 65 66 61 5f 4c  44 46 4c 41 47 53 2b 3d  |+..efa_LDFLAGS+=|
00000010  22 c2 a0 2d 4c 24 65 66  61 5f 4c 49 42 44 49 52  |"..-L$efa_LIBDIR|
00000020  20 2d 57 6c 2c 2d 2d 65  6e 61 62 6c 65 2d 6e 65  | -Wl,--enable-ne|
00000030  77 2d 64 74 61 67 73 2c  2d 72 70 61 74 68 2c 24  |w-dtags,-rpath,$|
00000040  65 66 61 5f 4c 49 42 44  49 52 22 0a              |efa_LIBDIR".|
0000004c

❌ PR character safety check failed
```

### DIFF
--- a/.github/workflows/scripts/check_nonprintable.sh
+++ b/.github/workflows/scripts/check_nonprintable.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+BASE_SHA="$1"
+HEAD_SHA="$2"
+
+if [ -z "$BASE_SHA" ] || [ -z "$HEAD_SHA" ]; then
+    echo "Usage: $0 <base_sha> <head_sha>"
+    exit 1
+fi
+
+echo "Checking for non-printable characters..."
+
+# Store current directory and change to git root for proper file paths
+SCRIPT_DIR="$(pwd)"
+cd "$(git rev-parse --show-toplevel)"
+
+# Get added lines from text files only
+if [ -s "$SCRIPT_DIR/text_files.txt" ]; then
+    # Build file list manually for compatibility
+    files=""
+    while IFS= read -r file; do
+        files="$files $file"
+    done < "$SCRIPT_DIR/text_files.txt"
+
+    if [ -n "$files" ]; then
+        git diff --no-color --unified=0 "$BASE_SHA..$HEAD_SHA" -- $files | \
+          grep '^+' | grep -v '^+++' > "$SCRIPT_DIR/pr_added_lines.txt" || true
+    else
+        touch "$SCRIPT_DIR/pr_added_lines.txt"
+    fi
+else
+    touch "$SCRIPT_DIR/pr_added_lines.txt"
+fi
+
+# Change back to script directory for the rest of the script
+cd "$SCRIPT_DIR"
+
+if [ ! -s pr_added_lines.txt ]; then
+    echo "No added lines to scan."
+    exit 0
+fi
+
+# Check for control characters and problematic Unicode whitespace
+# Control chars, no-break space, zero-width space, etc.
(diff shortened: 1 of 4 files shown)


.github/workflows/scripts/check_nonprintable.sh
```diff
@@ -0,0 +1,31 @@
...
+if [ ! -s pr_added_lines.txt ]; then
+    echo "No added lines to scan."
+    exit 0
+fi
+
+# Check for control characters and problematic Unicode whitespace
+# Control chars, no-break space, zero-width space, etc.
+if LC_ALL=C grep -nP '[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]|\xC2\xA0|\xE2\x80\x8B' pr_added_lines.txt > hits.txt 2>/dev/null; then
+    echo "Non-printable or problematic characters detected:"
+    cat hits.txt
```
darrylabbate: Perhaps `hexdump`/`xxd` instead of (or in addition to) `cat` here?

alekswn: Added hexdump for each hit

alekswn:
AWS CI failures are irrelevant to GitHub actions workflow

alekswn:
bot:aws:retest

alekswn:
bot:aws:retest

sunkuamzn:
@j-xiong @aingerson could you take a look? Thank you!

aingerson (approved):
I don't have a problem with it but I do think the proper fix is to make sure we're testing the paths that we're changing. Like the issue that came up earlier with the rdma-core non breaking space was merged because we never tested that path which I think was a miss. This will definitely help diagnose them better so it's fine, but we need to do our due diligence as developers and make sure we're not opening PRs for untested paths.

 I'm going to assume that all the characters you're checking for are ones that you're *positive* we would never need?

alekswn:
The workflow would ignore any binary files and detect following characters in text files:
`[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]` - ASCII control characters:
- `\x00-\x08`: NULL, SOH, STX, ETX, EOT, ENQ, ACK, BEL, BS
- `\x0B`: Vertical Tab (VT)
- `\x0C`: Form Feed (FF)
- `\x0E-\x1F`: SO, SI, DLE, DC1-4, NAK, SYN, ETB, CAN, EM, SUB, ESC, FS, GS, RS, US
- `\x7F`: DEL (Delete)

UTF-8 invisible characters:
- `U+00A0`: NBSP Non-breaking space
- `U+200B`: ZWSP Zero-width space

None of these characters currently present in the codebase and I do not think there is any good reason to insert any of them.

We still allow following ASCII control characters:
- \x09 - Tab - '\t'
- \x0A - Line Feed - '\n'
- \x0D - Carriage Return - '\r'

Most UTF-8 invisible characters are still allowed.

PR #11765, opened 2025-12-26
prov/efa: Fix the IBV access flags for registering MR

jiaxiyan:
We should set the IBV access flags based on user provided mr access flags and the device support for RDMA.

### DIFF
--- a/fabtests/prov/efa/src/efa_mmap_test.c
+++ b/fabtests/prov/efa/src/efa_mmap_test.c
@@ -1,0 +1,243 @@
+/*
+ * Copyright (c) 2025, Amazon.com, Inc.  All rights reserved.
+ *
+ * This software is available to you under the BSD license
+ * below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <getopt.h>
+#include <sys/mman.h>
+#include <string.h>
+#include <errno.h>
+#include <shared.h>
+#include <rdma/fi_ext.h>
+
+static void *mmap_buf;
+static struct fid_mr *mmap_mr;
+static bool use_emulated_read;
+
+struct test_case {
+	const char *name;
+	int mmap_prot;
+	uint64_t mr_access;
+	bool should_pass;
+};
+
+static struct test_case test_cases[] = {
+	{"PROT_READ for FI_WRITE", PROT_READ, FI_WRITE, true},
+	{"PROT_READ for FI_REMOTE_WRITE", PROT_READ, FI_REMOTE_WRITE, false},
+	{"PROT_READ|PROT_WRITE for FI_REMOTE_WRITE", PROT_READ | PROT_WRITE, FI_REMOTE_WRITE, true},
+	{"PROT_READ for FI_RECV", PROT_READ, FI_RECV, false},
+	{"PROT_READ|PROT_WRITE for FI_RECV", PROT_READ | PROT_WRITE, FI_RECV, true},
+	{"PROT_READ for FI_READ", PROT_READ, FI_READ, false},
+	{"PROT_READ|PROT_WRITE for FI_READ", PROT_READ | PROT_WRITE, FI_READ, true},
+	{"PROT_READ for FI_REMOTE_READ", PROT_READ, FI_REMOTE_READ, true},
+};
+
+#define NUM_TEST_CASES (sizeof(test_cases) / sizeof(test_cases[0]))
+
+static int setup_mmap_buffer(int prot, uint64_t access)
+{
+	int ret;
+	void *mmap_desc;
+
+	/* Create writable mapping first to initialize data */
+	mmap_buf = mmap(NULL, opts.transfer_size, PROT_READ | PROT_WRITE,
+			MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+	if (mmap_buf == MAP_FAILED) {
+		FT_PRINTERR("mmap", -errno);
+		return errno;
+	}
+
+	/* Initialize buffer with test data */
+	ft_fill_buf(mmap_buf, opts.transfer_size);
+
+	/* Set desired protection */
+	if (mprotect(mmap_buf, opts.transfer_size, prot) < 0) {
+		FT_PRINTERR("mprotect", -errno);
+		munmap(mmap_buf, opts.transfer_size);
+		return errno;
+	}
+
+	/* Register memory region */
+	ret = ft_reg_mr(fi, mmap_buf, opts.transfer_size, access,
+			FT_MR_KEY, FI_HMEM_SYSTEM, 0, &mmap_mr, &mmap_desc);
+	if (ret) {
+		FT_PRINTERR("ft_reg_mr", ret);
+		munmap(mmap_buf, opts.transfer_size);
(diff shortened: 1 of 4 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -784,14 +784,51 @@ static uint64_t efa_mr_non_p2p_keygen(void) {
...
+			     bool device_support_rdma_write)
+{
+	int ibv_access = 0;
+
+	/* To support Emulated RMA path, if the access is not supported
+	 * by EFA, modify it to FI_SEND | FI_RECV
+	 */
+	if (!ofi_access || (ofi_access & ~EFA_MR_SUPPORTED_PERMISSIONS) ||
+	    ((ofi_access & FI_READ) && !device_support_rdma_read) ||
+	    ((ofi_access & FI_WRITE) && !device_support_rdma_write))
```
sunkuamzn: If `FI_WRITE` is requested on an instance that supports RDMA write, we don't set `IBV_ACCESS_LOCAL_WRITE`. But if it is requested on an instance that doesn't support RDMA write, we set `IBV_ACCESS_LOCAL_WRITE`. Which doesn't seem right.

jiaxiyan: FI_WRITE doesn't need IBV_ACCESS_LOCAL_WRITE

sunkuamzn: Discussed offline. We shouldn't set `IBV_ACCESS_LOCAL_WRITE` when only `FI_WRITE` is requested.

prov/efa/src/efa_mr.c
```diff
@@ -784,14 +784,51 @@ static uint64_t efa_mr_non_p2p_keygen(void) {
...
+	    ((ofi_access & (FI_READ | FI_REMOTE_READ)) && !device_support_rdma_read) ||
+	    ((ofi_access & (FI_WRITE | FI_REMOTE_WRITE)) && !device_support_rdma_write))
+		ofi_access = FI_SEND | FI_RECV;
+
+	if (ofi_access & FI_RECV)
+		ibv_access |= IBV_ACCESS_LOCAL_WRITE;
+
+	if (device_support_rdma_read) {
+		/* IBV_ACCESS_LOCAL_WRITE is needed for emulating fi_write with RMA read */
+		if (ofi_access & (FI_READ | FI_REMOTE_WRITE))
```
shijin-aws: So basically we need IBV_ACCESS_LOCAL_WRITE if the buffer can be updated by rdma-read, right?

shijin-aws: The rdmamojo doesn't explain it clearly though

Enable Local Write Access: Memory Region can be used in Receive Requests or in IBV_WR_ATOMIC_CMP_AND_SWP or
IBV_WR_ATOMIC_FETCH_AND_ADD to write locally the remote content value

jiaxiyan: > So basically we need IBV_ACCESS_LOCAL_WRITE if the buffer can be updated by rdma-read, right?

Yes

fabtests/prov/efa/src/efa_mmap_test.c
```diff
@@ -0,0 +1,223 @@
...
+	ret = setup_mmap_buffer(test->mmap_prot, test->mr_access);
+
+	if (ret) {
+		if (test->should_pass) {
+			printf("FAIL: %s - mr reg failed unexpectedly: %s\n\n", 
+			       test->name, fi_strerror(-ret));
+			cleanup_mmap_buffer();
+			return -1;
+		} else {
+			printf("PASS: %s - mr reg failed as expected\n\n", test->name);
```
shijin-aws, line 137: what if the failure is from the mprotect and mmap? Should have a way to distinguish them?

prov/efa/src/efa_mr.c
sunkuamzn: > So basically we need IBV_ACCESS_LOCAL_WRITE if the buffer can be updated by rdma-read, right?

Or by RDMA write

shijin-aws: >Or by RDMA write

I thought that will be IBV_ACCESS_REMOTE_WRITE?

shijin-aws: Oh, but IBV_ACCESS_LOCAL_WRITE is also needed in that case, NVM

shijin-aws (dismissed):

prov/efa/src/efa_mr.c
```diff
@@ -811,6 +813,25 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *at
 	if (ret)
 		return ret;
 
+	device_support_rdma_read = efa_mr->domain->device->device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ;
+#if HAVE_CAPS_RDMA_WRITE
+	device_support_rdma_write = efa_mr->domain->device->device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_WRITE;
+#endif
+	/* For efa-direct, fail registration if RDMA operations are requested 
+	 * but hardware doesn't support them */
+	if (!strcmp(efa_mr->domain->util_domain.fabric->name, EFA_DIRECT_FABRIC_NAME)) {
```
sunkuamzn: You can also compare `efa_domain->info_type`

prov/efa/src/efa_mr.c
```diff
@@ -811,6 +813,25 @@ static int efa_mr_reg_impl(struct efa_mr *efa_mr, uint64_t flags, const void *at
...
+		if ((mr_attr.access & (FI_READ | FI_REMOTE_READ)) && !device_support_rdma_read) {
+			EFA_WARN(FI_LOG_MR,
+				 "FI_READ requested but hardware does not support RDMA read operations\n");
```
sunkuamzn: nit: log entry says `FI_READ` but it can also be printed when only `FI_REMOTE_READ` is requested - which would be confusing to the user

shijin-aws (dismissed):

prov/efa/src/efa_mr.c
```diff
@@ -784,14 +784,46 @@ static uint64_t efa_mr_non_p2p_keygen(void) {
...
+	 * by EFA, modify it to FI_SEND | FI_RECV
+	 */
+	if (!ofi_access || (ofi_access & ~EFA_MR_SUPPORTED_PERMISSIONS))
+		ofi_access = FI_SEND | FI_RECV;
+
+	if (ofi_access & (FI_RECV | FI_REMOTE_WRITE | FI_READ))
+		ibv_access |= IBV_ACCESS_LOCAL_WRITE;
+
+	if (device_support_rdma_read &&
+	    (ofi_access & (FI_REMOTE_READ | FI_RECV) ||
```
sunkuamzn: Setting IBV_ACCESS_REMOTE_READ for FI_RECV doesn't make sense to me

Here's what I think the flags should be
### Device does not support RDMA read or RDMA write

All FI_* flags: no need to set any extra IBV* flags. Local read is always allowed. All transfers into the application buffer are copies from bounce buffer. The EFA device never writes to the application buffer.

### Device supports RDMA read but not RDMA write

* FI_SEND: IBV_ACCESS_REMOTE_READ - required for long/runting read protocols
* FI_RECV: IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE - required for long/runting read protocols
* FI_READ: IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE
* FI_REMOTE_READ: IBV_ACCESS_REMOTE_READ
* FI_WRITE: IBV_ACCESS_REMOTE_READ - required when we emulate write with RDMA read
* FI_REMOTE_WRITE: IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE


### Device supports RDMA read and RDMA write
* FI_SEND, FI_RECV, FI_READ, FI_REMOTE_READ: Same as the case with RDMA read but without RDMA write
* FI_WRITE: I'm not sure if IBV_ACCESS_REMOTE_READ is required here or if LOCAL_READ is sufficient - need to check with a test
* FI_REMOTE_WRITE: IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE

jiaxiyan: > All FI_* flags: no need to set any extra IBV* flags.

nccl tests hang on g4d(Device does not support RDMA read or RDMA write) if we don't set IBV_ACCESS_LOCAL_WRITE for FI_RECV.
> FI_RECV: IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE - required for long/runting read protocols

Why do FI_RECV and FI_READ need IBV_ACCESS_REMOTE_WRITE for read protocols? 

> FI_WRITE: I'm not sure if IBV_ACCESS_REMOTE_READ is required here or if LOCAL_READ is sufficient - need to check with a test

It is not required because we use RDMA write when p2p is available. If p2p is unavailable, RDMA read won't be used either.

sunkuamzn: > Why do FI_RECV and FI_READ need IBV_ACCESS_REMOTE_WRITE for read protocols?

Because an RDMA read operation will write to the buffer? It could be similar to the case of `FI_WRITE` when device supports RDMA write and that only need LOCAL_WRITE is required.

jiaxiyan: https://www.rdmamojo.com/2012/09/07/ibv_reg_mr/
> IBV_ACCESS_REMOTE_WRITE Enable Remote Write Access: Memory Region can be access from remote context using IBV_WR_RDMA_WRITE or IBV_WR_RDMA_WRITE_WITH_IMM

I don't think we need IBV_ACCESS_REMOTE_WRITE for the source of RDMA read

jiaxiyan:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

jiaxiyan:
bot:aws:retest

shijin-aws (approved):

sunkuamzn (approved):

PR #11773, opened 2025-12-30
prov/efa: Stop setting qp->ibv_qp_ex->wr_id for data path direct

a-szegel:
The efa-data path direct code does not require qp->ibv_qp_ex->wr_id for userspace book keeping, and the efa-device is reading the wqe/db. We can get a micro optimization by not setting this value because it is not being used.


shijin-aws:
nit: rename the title to be stop setting qp->ibv_qp_ex->wr_id in efa data path direct

a-szegel:
bot:aws:retest

PR #11786, opened 2026-01-05
Fix efa_mr_reg_impl/fa_mr_dereg_impl race condition

alekswn:
- Converted `ibv_mr_reg_ct` and `ibv_mr_reg_sz` to `ofi_atomic64`
- Updated MR registration/deregistration code to use atomic operations
- Maintained existing logging functionality with thread-safe counter reads

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -201,8 +201,8 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 		goto err_free;
 	}
 
-	efa_domain->ibv_mr_reg_ct = 0;
-	efa_domain->ibv_mr_reg_sz = 0;
+	ofi_atomic_initialize64(&efa_domain->ibv_mr_reg_ct, 0);
+	ofi_atomic_initialize64(&efa_domain->ibv_mr_reg_sz, 0);
 
 	efa_domain->ah_map = NULL;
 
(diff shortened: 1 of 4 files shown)


include/ofi_atom.h
```diff
@@ -187,14 +192,55 @@ typedef atomic_long	ofi_atomic_int64_t;
...
+	static inline											\
+	size_t ofi_atomic_sub_counter(ofi_atomic_counter *atomic, size_t val)				\
+	{												\
+		ATOMIC_IS_INITIALIZED(atomic);								\
+		return (size_t)atomic_fetch_sub_explicit(&atomic->val, val,				\
+								 memory_order_relaxed) - val;		\
+	}												\
+	static inline											\
+	size_t ofi_atomic_inc_counter(ofi_atomic_counter *atomic)					\
+	{												\
```
shijin-aws: Can it be merged with `ofi_atomic_inc/add/set/dec` functions?

alekswn: I think that can be done, but not so trivial. The way OFI_ATOMIC_DEFINE macro is implemented it does not have options for unsigned integer and memory model (following Windows model I guess).  Therefore the macro definition is going to become extremely complex.

alekswn: I'm excluding the new define from this PR. Going to cut a new PR with better macros.
For the time being we can use signed integer `ofi_atomic64_t`.

alekswn:
bot:aws:retest

alekswn:
bot:aws:retest

alekswn:
bot:aws:retest

PR #11788, opened 2026-01-05
prov/efa: Fix the access flags in efa_mr_cache_entry_reg

jiaxiyan:
Update it to be all supported access modes.

### DIFF
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -254,12 +254,11 @@ int efa_mr_cache_entry_reg(struct ofi_mr_cache *cache,
 	int ret = 0;
 	/* TODO
 	 * Since, access is not passed as a parameter to efa_mr_cache_entry_reg,
-	 * for now we will set access to all supported access modes i.e.
-	 * FI_SEND | FI_RECV | FI_REMOTE_READ | FI_REMOTE_WRITE. Once access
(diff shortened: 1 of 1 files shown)


prov/efa/src/efa_mr.c
```diff
@@ -254,12 +254,11 @@ int efa_mr_cache_entry_reg(struct ofi_mr_cache *cache,
...
 	/* TODO
 	 * Since, access is not passed as a parameter to efa_mr_cache_entry_reg,
-	 * for now we will set access to all supported access modes i.e.
-	 * FI_SEND | FI_RECV | FI_REMOTE_READ | FI_REMOTE_WRITE. Once access
+	 * for now we will set access to all supported access modes. Once access
 	 * information is available this can be removed.
 	 * Issue: https://github.com/ofiwg/libfabric/issues/5677
 	 */
-	uint64_t access = FI_SEND | FI_RECV | FI_REMOTE_READ | FI_REMOTE_WRITE;
+	uint64_t access = EFA_MR_SUPPORTED_PERMISSIONS;
```
shijin-aws, line 261: ~~Nit: what mode was missing in the old access ? This change is ok but I'd like to call out that mr cache reg is only for internal registration which doesn't involve rdma write?~~

Nvm, I can see the code can try to fill the desc for all operations

jiaxiyan:
CI failure is unrelated

PR #11794, opened 2026-01-07
prov/efa: Verify QP number on CQE process

dkkranz:
When accessing the QP table with the masked QP number from the CQE, need to make sure the QP number still matches. Otherwise, the QP for which the work request had been completed was destroyed, and its table entry was subsequently re-occupied by another QP.


shijin-aws:
bot:aws:retest

PR #11795, opened 2026-01-07
prov/efa: Add UT for error completion handling in efa_rdm_ep_wait_send

a-szegel:
Add unit test to verify that efa_outstanding_tx_ops counter is properly decremented when error completions are processed during endpoint closing, preventing resource leaks during shutdown.


sunkuamzn (approved):
Thanks!

PR #11796, opened 2026-01-08
prov/lnx: Add Multi-Rail selection policy

amirshehataornl:
Add two Multi-Rail selection policies:

1. PER_MSG: This is what was originally implemented. This method round robins over the local endpoints and peer addresses on every sent message. It doesn't support send after send requirements.
2. PER_PEER: In this method the local end points are round robined over per peer. This way all local endpoints are used across the peers, however, each peer is assigned exactly one local endpoint which sends to one of its addresses. This ensures send after send is satisfied.

A new environment variable, FI_LNX_MULTI_RAIL_SELECTION, can be set to either PER_MSG or PER_PEER. If the environment variable is not set it defaults to PER_PEER.

### DIFF
--- a/prov/lnx/include/lnx.h
+++ b/prov/lnx/include/lnx.h
@@ -39,6 +39,9 @@
 #define LNX_MAX_PRIMARY_ID	((1ULL << 56) - 1)
 #define LNX_MAX_SUB_ID 		((1ULL << 8) - 1)
 
+#define LNX_PER_MSG_SELECTION_STR	"PER_MSG"
+#define LNX_PER_PEER_SELECTION_STR	"PER_PEER"
+
 #define lnx_ep_rx_flags(lnx_ep) ((lnx_ep)->le_ep.rx_op_flags)
 
 struct lnx_match_attr {
@@ -134,6 +137,11 @@ struct lnx_peer {
 	ofi_atomic32_t lp_ep_rr;
 	struct lnx_core_av **lp_avs;
 	struct lnx_peer_ep_map *lp_src_eps;
+	/* Both fields below are used to lock the peer to a specific local
+	 * endpoint and remote peer address
+	 */
+	fi_addr_t lp_locked_core_addr;
+	struct lnx_core_ep *lp_locked_cep;
 };
 
 struct lnx_av {
@@ -159,6 +167,12 @@ struct lnx_domain {
 	int ld_ep_idx;
 };
 
+enum lnx_multirail_selection {
+	LNX_MR_SELECTION_PER_MSG = 0,
+	LNX_MR_SELECTION_PER_PEER,
+	LNX_MR_SELECTION_MAX,
+};
+
 struct lnx_ep {
 	int le_idx;
 	struct util_ep le_ep;
@@ -169,6 +183,9 @@ struct lnx_ep {
 	size_t le_fclass;
 	struct lnx_peer_srq le_srq;
 	struct lnx_av *le_lav;
+	/* global round robin index */
+	ofi_atomic32_t le_rr;
+	enum lnx_multirail_selection le_mr;
 };
 
 struct lnx_cq {
@@ -289,13 +306,12 @@ lnx_get_core_addr(struct lnx_core_ep *cep, fi_addr_t addr)
 	return map_addr->map_addrs[idx];
 }
 
-/* TODO:
- * Make this function a callback. Intent is to be able to have different
- * ways to select the endpoints. This is useful to try different methods
- * of selecting the endpoint <-> peer addr pairing.
+/*
+ * Round robin over the end points and peer addresses per message
+ * This does not take into consideration send after send requirements.
  */
 static inline int
-lnx_select_send_endpoints(struct lnx_ep *lep, fi_addr_t lnx_addr,
+lnx_select_send_endpoints_msg(struct lnx_ep *lep, fi_addr_t lnx_addr,
 		struct lnx_core_ep **cep_out, fi_addr_t *core_addr)
 {
 	int idx, rr;
@@ -309,8 +325,7 @@ lnx_select_send_endpoints(struct lnx_ep *lep, fi_addr_t lnx_addr,
 		return -FI_ENOSYS;
 
 	/* round robin over the endpoints which can reach this peer */
-	rr = ofi_atomic_get32(&lp->lp_ep_rr);
-	ofi_atomic_inc32(&lp->lp_ep_rr);
+	rr = ofi_atomic_inc32(&lp->lp_ep_rr) - 1;
 	ep_map = &lp->lp_src_eps[lep->le_idx];
 	idx = rr % ep_map->pem_num_eps;
 	cep = ep_map->pem_eps[idx];
@@ -318,8 +333,7 @@ lnx_select_send_endpoints(struct lnx_ep *lep, fi_addr_t lnx_addr,
 	map_addr = ofi_bufpool_get_ibuf(cep->cep_cav->cav_map, lp->lp_addr);
 
 	/* round robin over available peer addresses */
-	rr = ofi_atomic_get32(&map_addr->map_rr);
-	ofi_atomic_inc32(&map_addr->map_rr);
+	rr = ofi_atomic_inc32(&map_addr->map_rr) - 1;
 	idx = rr % (map_addr->map_count);
 
 	*core_addr = map_addr->map_addrs[idx];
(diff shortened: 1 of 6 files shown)


amirshehataornl:
Any chance someone can take a look at this patch? @aingerson, @j-xiong

prov/lnx/include/lnx.h
```diff
@@ -329,4 +347,98 @@ lnx_select_send_endpoints(struct lnx_ep *lep, fi_addr_t lnx_addr,
...
+		*core_addr = lp->lp_locked_core_addr;
+		*cep_out = lp->lp_locked_cep;
+		return FI_SUCCESS;
+	}
+
+	ep_map = &lp->lp_src_eps[lep->le_idx];
+
+find_again:
+	rr = ofi_atomic_get32(&lep->le_rr);
+	ofi_atomic_inc32(&lep->le_rr);
```
aingerson: atomic_inc returns the new value of the atomic so you can do `rr = ofi_atomic_inc32() - 1` here if you want

prov/lnx/include/lnx.h
```diff
@@ -329,4 +347,98 @@ lnx_select_send_endpoints(struct lnx_ep *lep, fi_addr_t lnx_addr,
...
+
+	if (lp->lp_selection_done) {
+		*core_addr = lp->lp_locked_core_addr;
+		*cep_out = lp->lp_locked_cep;
+		return FI_SUCCESS;
+	}
+
+	ep_map = &lp->lp_src_eps[lep->le_idx];
+
+find_again:
```
aingerson: This is a little hard to read - maybe make it a while loop? Is there a way to get stuck in this loop? What if we don't find it?

prov/lnx/include/lnx.h
```diff
@@ -329,4 +347,98 @@ lnx_select_send_endpoints(struct lnx_ep *lep, fi_addr_t lnx_addr,
...
+	rr = ofi_atomic_get32(&lep->le_rr);
+	ofi_atomic_inc32(&lep->le_rr);
+	idx = rr % lep->le_domain->ld_num_doms;
+	cep = &lep->le_core_eps[idx];
+
+	for (i = 0; i < ep_map->pem_num_eps; i++) {
+		if (cep == ep_map->pem_eps[i]) {
+			found = true;
+			break;
+		}
```
aingerson: I'm confused what we're checking for here and with the overall selection logic. Don't we create one EP per rail/domain? Can't we just round robin over those EPs and then lock it in? What's with this checking that it's in the map and then the second randomization of the address?

amirshehataornl: Ok, I'm getting on the bandwagon and trying some AI explanation from deepwiki. It's a little verbose, but I think it covers what I was trying to do:

# DeepWiki Q&A with Code Context for Repository: amirshehataornl/libfabric


## Q1

can you explain lnx_select_send_endpoints_peer in details


### Answer

# Answer

You're asking about the `lnx_select_send_endpoints_peer` function, which is a static inline function in the LNX (LINKx) provider for libfabric libfabric:355-357 </cite>.

## Purpose

`lnx_select_send_endpoints_peer` implements the PER_PEER multi-rail selection policy libfabric:350-354 </cite>. This policy ensures that each peer is assigned a specific local endpoint and peer address, guaranteeing that messages are received in the same order they are sent</cite>.

## How It Works

The function takes three parameters: the LNX endpoint (`lep`), the destination address (`lnx_addr`), and outputs the selected core endpoint (`cep_out`) and core address (`core_addr`) libfabric:356-357 </cite>.

### Fast Path - Already Selected

If a peer has already been assigned an endpoint/address pair (`lp->lp_selection_done` is true), the function immediately returns the cached values libfabric:371-375 </cite>.

### Selection Process

When selecting for the first time:

1. **Lookup the peer**: Retrieves the peer structure from the address vector libfabric:367-369 </cite>

2. **Round-robin endpoint selection**: Uses a global round-robin counter (`le_rr`) to select a local endpoint index libfabric:380-383 </cite>

3. **Validate endpoint**: Checks if the selected endpoint can actually reach this peer by searching the peer's endpoint map libfabric:385-392 </cite>. If not found, it tries again with the next round-robin value</cite>

4. **Randomize peer address**: Instead of round-robin, it uses a randomized selection based on current time to choose from available peer addresses libfabric:396-405 </cite>. This spreads traffic across peer addresses when multiple processes communicate with the same peer</cite>

5. **Lock the selection**: Marks the selection as done and caches both the core address and endpoint for future sends to this peer libfabric:407-409 </cite>

## Usage

This function is invoked through a function pointer array when the multi-rail selection mode is set to `LNX_MR_SELECTION_PER_PEER` libfabric:54-60 </cite>. For example, in `lnx_tsend`, it's called as `lnx_select_send_endpoints[lep->le_mr](lep, dest_addr, &cep, &core_addr)` libfabric:549-551 </cite>.

The selection mode defaults to PER_PEER if not explicitly configured libfabric:648-653 </cite>, and can be set via the `FI_LNX_MULTI_RAIL_SELECTION` environment variable</cite>.

## Notes

The PER_PEER policy differs from PER_MSG (`lnx_select_send_endpoints_msg`) which round-robins on every message libfabric:315-318 </cite>. PER_PEER is the safer default as it maintains message ordering guarantees</cite>.

There's also a companion function `lnx_set_send_pair_peer` that handles the receive side, ensuring symmetry by locking the send pair when receiving the first message from a peer libfabric:425-442 </cite>.


## Q2

can you expand on:
Validate endpoint: Checks if the selected endpoint can actually reach this peer by searching the peer's endpoint map lnx.h:385-392 . If not found, it tries again with the next round-robin value

How is validation exactly done.

Explain the mapping


### Answer

# Answer

You're asking about the endpoint validation logic in `lnx_select_send_endpoints_peer`, specifically how the function verifies that a selected core endpoint can actually reach the target peer.

## The Validation Process

The validation happens in a loop that checks if the selected core endpoint (`cep`) exists in the peer's endpoint map libfabric:385-392 . Here's how it works:

1. **Get the peer's endpoint map**: The function retrieves `ep_map` from `lp->lp_src_eps[lep->le_idx]` libfabric:377-377 . This gives the list of core endpoints that can reach this peer from the current LNX endpoint</cite>.

2. **Linear search**: The code iterates through `ep_map->pem_eps[]` array, comparing each entry against the selected `cep` libfabric:385-390 . If a match is found, `found` is set to true</cite>.

3. **Retry on failure**: If the endpoint isn't found in the map (`!found`), the code jumps back to `find_again` libfabric:391-392 , which increments the round-robin counter and tries the next endpoint</cite>.

## The Mapping Structure

The mapping is a two-level structure that tracks connectivity between local and remote endpoints:

### Peer Structure

Each peer (`struct lnx_peer`) contains `lp_src_eps`, which is an array of `lnx_peer_ep_map` structures libfabric:133-147 . There's one entry per local LNX endpoint (`lp_ep_count` entries total)</cite>.

### Endpoint Map Structure

Each `lnx_peer_ep_map` contains:
- `pem_eps`: An array of pointers to `lnx_core_ep` structures libfabric:128-131 
- `pem_num_eps`: The count of endpoints in the array</cite>

This map is populated during address insertion by `lnx_setup_ep_mapping` libfabric:218-262 , which builds the connectivity matrix by examining which core endpoints can reach which peer addresses</cite>.

### Why This Validation Is Needed

Not all core endpoints can reach all peers. In a multi-rail setup, you might have:
- Multiple network interfaces (e.g., InfiniBand, Ethernet)
- Different providers (e.g., verbs, tcp, shm)
- Network topology constraints

The round-robin selection at line 382 picks from all available core endpoints (`lep->le_core_eps[idx]`) libfabric:380-383 , but not all of them may have connectivity to the target peer</cite>. The validation ensures we only use endpoints that can actually deliver messages to this specific peer.

## Notes

The PER_MSG selection mode (`lnx_select_send_endpoints_msg`) uses a different approach - it round-robins directly over the peer's endpoint map (`lp->lp_src_eps[lep->le_idx]`) libfabric:329-334 , so it never needs validation since it only considers endpoints known to reach the peer</cite>. The PER_PEER mode needs validation because it tries to distribute peers evenly across all local endpoints using the global counter.

## Q3

can you explain why Randomize peer address: Instead of round-robin, it uses a randomized selection based on current time to choose from available peer addresses lnx.h:396-405 . This spreads traffic across peer addresses when multiple processes communicate with the same peer

is needed. Explain via an example


### Answer

# Answer

You're asking why the PER_PEER selection mode uses randomization instead of round-robin when choosing peer addresses in `lnx_select_send_endpoints_peer` libfabric:396-405 .

## The Problem with Round-Robin in PER_PEER Mode

The key issue is that **PER_PEER locks the endpoint/address pair for the lifetime of communication with that peer** libfabric:407-409 . Once selected, this choice is cached and reused for all future messages to that peer libfabric:371-375 .

The comment explains: "there is no effective way to enforce round robin across independent peers" libfabric:400-401 . This is because each process makes its selection independently, and there's no coordination between processes.

## Example Scenario

Consider a multi-rail setup where:
- **Peer Node A** has 2 network interfaces: `cxi0_addr` and `cxi1_addr`
- **Multiple sender processes** (Process 1, Process 2, Process 3, ...) all want to send to Peer Node A

### What Would Happen with Round-Robin

If we used round-robin (like PER_MSG does libfabric:338-341 ), all processes would independently start at index 0:

- **Process 1**: Selects `cxi0_addr` (index 0), locks it forever
- **Process 2**: Selects `cxi0_addr` (index 0), locks it forever  
- **Process 3**: Selects `cxi0_addr` (index 0), locks it forever
- ...

**Result**: All processes would send to the same peer address (`cxi0_addr`), completely saturating one interface while leaving `cxi1_addr` idle. This defeats the purpose of multi-rail.

### What Happens with Randomization

With time-based randomization libfabric:403-405 :

- **Process 1** (starts at time T1): Random seed → selects `cxi1_addr`
- **Process 2** (starts at time T2): Different random seed → selects `cxi0_addr`
- **Process 3** (starts at time T3): Different random seed → selects `cxi1_addr`
- ...

**Result**: Traffic is probabilistically distributed across both `cxi0_addr` and `cxi1_addr`, utilizing both interfaces and avoiding hotspots.

## Why Not Use PER_MSG Instead?

PER_MSG does use round-robin over peer addresses libfabric:338-341  and would distribute traffic perfectly. However, PER_MSG doesn't guarantee message ordering libfabric:311-313  because different messages can take different paths. PER_PEER is the default libfabric:648-653  specifically because it maintains ordering guarantees libfabric:350-354 .

## Notes

The randomization uses `ofi_gettime_ns()` as the seed libfabric:403-403 , which ensures different processes starting at different times will get different seeds and thus different selections. This is a pragmatic compromise: it doesn't achieve perfect load balancing like PER_MSG's round-robin, but it spreads traffic reasonably well while maintaining the ordering guarantees that PER_PEER provides.

The environment variable `FI_LNX_MULTI_RAIL_SELECTION` allows users to choose between these policies libfabric:158-171  based on whether they need ordering (PER_PEER) or maximum throughput (PER_MSG).

prov/lnx/include/lnx.h
amirshehataornl: done

prov/lnx/include/lnx.h
amirshehataornl: done

amirshehataornl:
@aingerson, do the update and comments satisfy your concerns?

aingerson:
@amirshehataornl Yes! Thank you that was very helpful! I will give it another review

prov/lnx/include/lnx.h
```diff
@@ -134,6 +137,13 @@ struct lnx_peer {
 	ofi_atomic32_t lp_ep_rr;
 	struct lnx_core_av **lp_avs;
 	struct lnx_peer_ep_map *lp_src_eps;
+	/* Both fields below are used to lock the peer to a specific local
+	 * endpoint and remote peer address
+	 */
+	fi_addr_t lp_locked_core_addr;
+	struct lnx_core_ep *lp_locked_cep;
+	/* set to true once the selection is complete */
+	bool lp_selection_done;
```
aingerson: Can you drop this field and just check if `lp_locked_cep` is NULL or not?

prov/lnx/src/lnx_init.c
```diff
@@ -761,6 +761,10 @@ LNX_INI
 	struct ofi_bufpool_attr bp_attrs = {};
 	int ret;
 
+	fi_param_define(&lnx_prov, "multi_rail_selection", FI_PARAM_STRING,
+			"Specify which Multi-Rail endpoint selection "
+			"algorithm to use. One of: PER_MSG, PER_PEER");
```
aingerson: Maybe add in the default value?

prov/lnx/src/lnx_ops.c
```diff
@@ -52,6 +51,21 @@
...
 
+static int (*lnx_select_send_endpoints[LNX_MR_SELECTION_MAX])
+	(struct lnx_ep *lep, fi_addr_t lnx_addr,
+	 struct lnx_core_ep **cep_out, fi_addr_t *core_addr) =
+{
+	lnx_select_send_endpoints_msg,
+	lnx_select_send_endpoints_peer,
+};
+
+static void (*lnx_set_send_pair[LNX_MR_SELECTION_MAX])
```
aingerson, line 63: Can you define this and the above functions with the enums they refer to so it's a bit more readable?
like this
```
static void (*lnx_set_send_pair[LNX_MR_SELECTION_MAX])
	(struct lnx_ep *lep, struct lnx_core_ep *cep, fi_addr_t addr) =
{
	[LNX_MR_SELECTION_MSG] = lnx_set_send_pair_noop,
	[LNX_MR_SELECTION_PEER] = lnx_set_send_pair_peer,
};

aingerson (commented):
Just a few things to cleanup/add. Otherwise looks good! Thanks, Amir!

aingerson:
@amirshehataornl Would love to get this into the release this weekend. Can you update to address the last two things and then I will approve and merge?

amirshehataornl:
sure. Will get to it today... sorry about the delay

prov/lnx/include/lnx.h
amirshehataornl: done

prov/lnx/src/lnx_init.c
amirshehataornl: done

prov/lnx/src/lnx_ops.c
amirshehataornl: done

PR #11798, opened 2026-01-08
bindings(rust): add macos support

camshaft:
Currently the Rust bindings fail to compile on any other platforms outside of Linux. However, this limitation is artificial for macOS, where the bindings are able to compile with a single change to use the macOS C ABI with a `_` prefix for functions.

For this change, I've enabled builds to succeed on macOS and add checks in GitHub Actions to ensure it continues to work in the future.

Additionally, I've cleaned up the build script a bit to use the `libc` crate instead of generating new structs for things like `iovec`.

### DIFF
--- a/bindings/rust/wrapper.c
+++ b/bindings/rust/wrapper.c
@@ -1,12 +1,5 @@
 #include "wrapper.h"
 
-/* Proprietary helper function declarations. */
-
-fid_t get_fid_ptr(void *ptr)
-{
-	return (fid_t) ptr;
-}
-
 /* Static inline function declarations from fi_prov.h */
 
 int wrap_fi_param_get_str(struct fi_provider *provider, const char *param_name,
--- a/bindings/rust/wrapper.h
+++ b/bindings/rust/wrapper.h
@@ -21,9 +21,6 @@
 #include "fi_tagged.h"
 #include "fi_trigger.h"
 
-/* Proprietary helper function declarations. */
-fid_t get_fid_ptr(void *ptr);
-
 /* Static inline function declarations from fi_prov.h */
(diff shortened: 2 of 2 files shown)


sielicki:
As someone with an interest in these bindings, can you give your impression on whether building with [cmake](https://github.com/ofiwg/libfabric/pull/11789/files#diff-fa3f9d3323572470deb657e1fb6833ae444c5747ad46e0d61f42332d2d5af7d0), per my WIP pr [here](https://github.com/ofiwg/libfabric/pull/11789), makes anything easier in terms of crate distribution or maintenance?

camshaft:
Yeah I think CMake is generally more widely installed than autotools. I think _ideally_ the rust bindings wouldn't require either cmake or autotools and the build logic were replicated in a build script. We've done that in [s2n-tls](https://github.com/aws/s2n-tls/blob/main/bindings/rust/extended/s2n-tls-sys/build.rs) and [aws-lc-rs](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/cc_builder.rs) ([pr](https://github.com/aws/aws-lc-rs/pull/911)). But I also know it's a bit annoying to have to maintain build logic across systems so it's understandable if libfabric doesn't want to go that direction.

sielicki:
I've increasingly seen projects using cargo as their sole c/cxx build tool, and it looks awesome and I'm jealous, but I agree that it's probably out of scope for libfabric. Thanks for the input.

camshaft:
@j-xiong the macos bindings are passing with my change now :)

j-xiong:
Some of the changes seem to be unrelated to MacOS support (e.g. moving the location of tests). Could you put them in a separate commit? They can remain under this PR.

PR #11799, opened 2026-01-08
fabtests/efa/multi_ep_stress: Fix race conditions detected by TSAN

alekswn:
## Summary

This PR addresses several race conditions in the multi_ep_stress test that could cause intermittent failures.

### Race Condition Fixes
- **CQ/AV attribute initialization**: Move attribute initialization to local stack variables to prevent concurrent access to shared global attributes during endpoint setup
- **Worker status tracking**: Add proper synchronization for `ctx->status.active` field access to prevent races between worker threads and main thread
- **Random generator**: Replace non-thread-safe `rand()` calls with a new thread-safe random number generation API
- **Random skipping polling CQ**: Re-enable random skipping of CQ polling

### Code Quality
- **Dead code removal**: Clean up unused code including obsolete variables, redundant checks, and unused function calls

### DIFF
--- a/fabtests/prov/efa/src/multi_ep_stress.c
+++ b/fabtests/prov/efa/src/multi_ep_stress.c
@@ -864,7 +864,7 @@ static void *run_receiver_worker(void *arg)
 				"for cycle %d: %d\n",
 				ctx->worker_id, cycle + 1, ret);
 			cleanup_endpoint(&ctx->common);
-			continue;
+			goto out;
 		}
 
 		// sleep random time up to 100ms to emulate the real workload
@@ -961,7 +961,7 @@ static void *run_receiver_worker(void *arg)
 			usleep(1000);
 		}
 	}
-
+out:
 	printf("Receiver %d: Completed %d EP cycles\n", ctx->worker_id, cycle);
 	return (void *) (intptr_t) ret;
 }


alekswn:
bot:aws:retest

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -121,9 +117,6 @@ struct ep_message {
 struct worker_status {
 	pthread_mutex_t mutex;
 	bool active;
-	enum ep_status ep_status;
-	uint64_t ep_generation;
-	bool poll_cq;
```
shijin-aws, line 126: I'd like to keep the poll_cq, as we want to have an option that sender/receiver can skip cq poll before closing ep

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -244,7 +245,7 @@ void *notification_handler(void *arg)
 {
 	struct sender_context *ctx = (struct sender_context *) arg;
 
-	while (ctx->status.active) {
+	while (atomic_flag_test_and_set(ctx->status.active)) {
```
shijin-aws, line 241: I thought about it, I can see ctx->status.active is only set to false after all threads except for the notification threads are finished. In that regard, does it hurt anything today? I agree we should fix this race anyway

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: The actual skip is commented out for some reason, so it is a dead brach as of right now. I will revive that branch instead of deleting.

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: Thread might not see the update of variable up until next memory fence. That means sender thread might think it has cleared the flag, but notification thread will thing the flag is still set. I don't see how it would cause any trouble  in this case, but is still not correct.

alekswn:
Converting to draft to fix rest of threading issues in the test

fabtests/include/ft_random.h
```diff
@@ -0,0 +1,125 @@
...
+	} while (unlikely(result >= rejection_limit));
+	return result % modulo;
+}
+
+static inline int32_t ft_random_sleep_ms(struct random_data *restrict random_data, int max_sleep_time_ms)
+{
+	const int32_t sleep_time_ms = ft_random_get_int32_in_range(random_data, 0, max_sleep_time_ms);
+	struct timespec rem, ts = {
+		.tv_sec = sleep_time_ms / 1000,
+		.tv_nsec = sleep_time_ms % 1000
```
alekswn, line 111: Need to multiply by 1000000 here to get ns from ms

PR #11802, opened 2026-01-09
Fix fabric type duplication in provider enumeration

yinliaws:
## Problem

The EFA provider currently returns both 'efa' and 'efa-direct' fabric types simultaneously when enumerating providers via `fi_getinfo()`. This causes applications to see duplicate entries for the same physical EFA hardware:

- 16 entries with `fabric_attr->name = "efa"` (DGRAM endpoints)
- 16 entries with `fabric_attr->name = "efa-direct"` (RDM endpoints)
- Total: 32 entries representing the same 16 EFA devices

This duplication breaks applications that expect consistent provider enumeration, notably causing the AWS OFI NCCL GIN plugin to fail with: "Unexpectedly got more than one GIN provider"

## Root Cause

Applications like the GIN plugin count provider entries and expect a single, consistent fabric type per hardware device. When they see both fabric types, they interpret this as multiple distinct providers rather than different interfaces to the same hardware.

## Solution

This patch adds fabric type deduplication logic in `efa_getinfo()`:

1. Count fabric types** in the provider list
2. When both types are present, remove all 'efa' entries
3. Keep only 'efa-direct' entries (the preferred RDM interface)
4. Preserve backward compatibility by only removing duplicates when both exist

## Testing

- Verified with AWS OFI NCCL GIN plugin - error resolved
- Provider count reduced from 32 to 16 entries (eliminates duplicates)

### DIFF
--- a/prov/efa/src/efa_user_info.c
+++ b/prov/efa/src/efa_user_info.c
@@ -736,6 +736,41 @@ int efa_getinfo(uint32_t version, const char *node,
 		return err;
 	}
 
+	/* filter out duplicate fabric types */
+	if (info_list) {
+		struct fi_info *cur, *prev, *next;
+		int efa_count = 0, efa_direct_count = 0;
+		
+		/* Count fabric types first */
+		for (cur = info_list; cur; cur = cur->next) {
+			if (strcmp(cur->fabric_attr->name, "efa") == 0) efa_count++;
+			else if (strcmp(cur->fabric_attr->name, "efa-direct") == 0) efa_direct_count++;
+		}
+		
+		/* Remove efa entries if efa-direct is present */
+		if (efa_direct_count > 0 && efa_count > 0) {
+			prev = NULL;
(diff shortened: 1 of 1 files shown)


shijin-aws:
Please sign-off your commit via `git commit -s` otherwise DCO check will fail

shijin-aws:
>This patch adds fabric type deduplication logic in efa_getinfo():
    Count fabric types** in the provider list
    When both types are present, remove all 'efa' entries
    Keep only 'efa-direct' entries (the preferred RDM interface)
    Preserve backward compatibility by only removing duplicates when both exist


I think the requirement is, when application pass a fabric/domain ptr, provider should only return the info of the fabric from where the ptrs are created. It shouldn't matter whether it is `efa` or `efa-direct`

j-xiong:
Please add `prov/efa: ` prefix to the title line.

jiaxiyan:
Not the right fix

PR #11804, opened 2026-01-12
prov/efa: dmabuf try/fallback logic

charlesstoll:
This is a continuation of https://github.com/ofiwg/libfabric/pull/11465

Updates since then:
-Fixes based on comments
-New patch to modify flow for Neuron. Separated into a separate patch since the concept it implements felt separate enough from the previous one, despite both modifying the same functions

### DIFF
--- a/prov/efa/src/efa_hmem.c
+++ b/prov/efa/src/efa_hmem.c
@@ -143,21 +143,29 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
 	}
 
 #if HAVE_EFA_DMABUF_MR
-	ret = cuda_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
-	if (ret == FI_SUCCESS) {
-		ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
-					   len, (uint64_t)ptr, dmabuf_fd, ibv_access);
-		(void)cuda_put_dmabuf_fd(dmabuf_fd);
-		if (!ibv_mr) {
+	if (ofi_hmem_is_dmabuf_env_var_enabled(FI_HMEM_CUDA)) {
+		ret = ofi_hmem_get_dmabuf_fd(FI_HMEM_CUDA, ptr, len, &dmabuf_fd, &dmabuf_offset);
+		if (ret == FI_SUCCESS) {
+			ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
+						   len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+			(void)ofi_hmem_put_dmabuf_fd(FI_HMEM_CUDA, dmabuf_fd);
+			if (!ibv_mr) {
+				EFA_INFO(FI_LOG_CORE,
+					"Unable to register CUDA device buffer via dmabuf: %s. "
+					"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
+				ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
+			} else {
+				info->dmabuf_supported_by_device = EFA_DMABUF_SUPPORTED;
+			}
+		} else {
 			EFA_INFO(FI_LOG_CORE,
-				"Unable to register CUDA device buffer via dmabuf: %s. "
-				"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
+				"Unable to retrieve dmabuf fd of CUDA device buffer: %d. "
+				"Fall back to ibv_reg_mr\n", ret);
 			ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
+			info->dmabuf_supported_by_device = EFA_DMABUF_NOT_SUPPORTED;
 		}
 	} else {
-		EFA_INFO(FI_LOG_CORE,
-			"Unable to retrieve dmabuf fd of CUDA device buffer: %d. "
-			"Fall back to ibv_reg_mr\n", ret);
+		EFA_INFO(FI_LOG_CORE, "FI_HMEM_CUDA_USE_DMABUF set to false. Not using DMABUF for CUDA.\n");
 		ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
 	}
 #else
@@ -216,21 +224,29 @@ static inline void efa_hmem_info_check_p2p_support_rocr(struct efa_hmem_info *in
 	}
 
 #if HAVE_EFA_DMABUF_MR
-	ret = rocr_hmem_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
-	if (ret == FI_SUCCESS) {
-		ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
-					   len, (uint64_t) ptr, dmabuf_fd, ibv_access);
-		(void) rocr_hmem_put_dmabuf_fd(dmabuf_fd);
-		if (!ibv_mr) {
+	if (ofi_hmem_is_dmabuf_env_var_enabled(FI_HMEM_ROCR)) {
+		ret = rocr_hmem_get_dmabuf_fd(ptr, len, &dmabuf_fd, &dmabuf_offset);
+		if (ret == FI_SUCCESS) {
+			ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
+						   len, (uint64_t) ptr, dmabuf_fd, ibv_access);
+			(void) rocr_hmem_put_dmabuf_fd(dmabuf_fd);
+			if (!ibv_mr) {
+				EFA_INFO(FI_LOG_CORE,
+					"Unable to register ROCr device buffer via dmabuf: %s. "
+					"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
+				ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
+			} else {
+				info->dmabuf_supported_by_device = EFA_DMABUF_SUPPORTED;
+			}
+		} else {
 			EFA_INFO(FI_LOG_CORE,
-				"Unable to register ROCr device buffer via dmabuf: %s. "
-				"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
+				"Unable to retrieve dmabuf fd of ROCr device buffer: %d. "
+				"Fall back to ibv_reg_mr\n", ret);
 			ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
+			info->dmabuf_supported_by_device = EFA_DMABUF_NOT_SUPPORTED;
 		}
 	} else {
-		EFA_INFO(FI_LOG_CORE,
-			"Unable to retrieve dmabuf fd of ROCr device buffer: %d. "
-			"Fall back to ibv_reg_mr\n", ret);
+		EFA_INFO(FI_LOG_CORE, "FI_HMEM_ROCR_USE_DMABUF set to false. Not using DMABUF for ROCr.\n");
 		ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
 	}
 #else
@@ -263,84 +279,6 @@ static inline void efa_hmem_info_check_p2p_support_rocr(struct efa_hmem_info *in
 	return;
 }
 
-static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *info) {
-#if HAVE_NEURON
-	struct ibv_mr *ibv_mr = NULL;
-	struct ibv_pd *ibv_pd;
-	int ibv_access = IBV_ACCESS_LOCAL_WRITE;
-	void *handle;
-	void *ptr = NULL;
-	size_t len = ofi_get_page_size() * 2;
-	int dmabuf_fd;
-	uint64_t offset;
-	int ret;
-
-	if (g_efa_selected_device_list[0].device_caps & EFADV_DEVICE_ATTR_CAPS_RDMA_READ) {
-		ibv_access |= IBV_ACCESS_REMOTE_READ;
-	}
-
-	ptr = neuron_alloc(&handle, len);
-	/*
-	 * neuron_alloc will fail if application did not call nrt_init,
-	 * which is ok if it's not running neuron workloads. libfabric
-	 * will move on and leave info->initialized as false.
-	 */
-	if (!ptr) {
-		info->initialized = false;
-		EFA_INFO(FI_LOG_CORE, "Cannot allocate Neuron buffer\n");
-		return;
-	}
-
-	ibv_pd = ibv_alloc_pd(g_efa_selected_device_list[0].ibv_ctx);
-	if (!ibv_pd) {
-		EFA_WARN(FI_LOG_CORE, "failed to allocate ibv_pd: %d", errno);
-		neuron_free(&handle);
-		return;
-	}
-
-#if HAVE_EFA_DMABUF_MR
-	ret = neuron_get_dmabuf_fd(ptr, (uint64_t)len, &dmabuf_fd, &offset);
-	if (ret == FI_SUCCESS) {
-		ibv_mr = ibv_reg_dmabuf_mr(
-					ibv_pd, offset,
-					len, (uint64_t)ptr, dmabuf_fd, ibv_access);
-	} else if (ret == -FI_EOPNOTSUPP) {
-		EFA_INFO(FI_LOG_MR,
-			"Unable to retrieve dmabuf fd of Neuron device buffer, "
-			"Fall back to ibv_reg_mr\n");
-		ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
-	}
-#else
-	ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
-#endif
-
-	if (!ibv_mr) {
-		info->p2p_supported_by_device = false;
-		/* We do not expect to support Neuron on non p2p systems */
-		EFA_WARN(FI_LOG_CORE,
-		         "Failed to register Neuron buffer with the EFA device, "
-		         "FI_HMEM transfers that require peer to peer support will fail.\n");
-		neuron_free(&handle);
-		(void) ibv_dealloc_pd(ibv_pd);
-		return;
-	}
-
-	ret = ibv_dereg_mr(ibv_mr);
-	neuron_free(&handle);
-	(void) ibv_dealloc_pd(ibv_pd);
-	if (ret) {
-		EFA_WARN(FI_LOG_CORE,
-			 "Failed to deregister Neuron buffer: %s\n",
-			 fi_strerror(-ret));
-		return;
-	}
-
-	info->p2p_supported_by_device = true;
-	return;
-#endif
-	return;
-}
-
 /**
  * @brief Initialize the efa_hmem_info state for iface
  *
@@ -366,9 +304,26 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
 	}
 
 	info->initialized = true;
+	info->dmabuf_supported_by_device = EFA_DMABUF_ASSUMED;
+	info->dmabuf_fallback_enabled = false;
 
-	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
+	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM ||
+	    iface == FI_HMEM_NEURON) {
+		/* It is not recommended to allocate neuron buffs this
+		 * early in initialization, so we must skip the explicit
+		 * check to see if p2p will work. Instead, assume it works.
+		 * and set fallback to true
+		 */
+		if (iface == FI_HMEM_NEURON)
+			info->dmabuf_fallback_enabled = true;
+
 		info->p2p_supported_by_device = true;
+
+		if (!ofi_hmem_is_dmabuf_env_var_enabled(iface)) {
+			info->dmabuf_supported_by_device = EFA_DMABUF_NOT_SUPPORTED;
+			EFA_INFO(FI_LOG_CORE, "%s DMABUF disabled by environment variable\n",
+				 fi_tostr(&iface, FI_TYPE_HMEM_IFACE));
+		}
 	} else if (ofi_hmem_p2p_disabled()) {
 		info->p2p_supported_by_device = false;
 	} else {
@@ -378,9 +333,6 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
 			break;
 		case FI_HMEM_ROCR:
 			efa_hmem_info_check_p2p_support_rocr(info);
-			break;
-		case FI_HMEM_NEURON:
-			efa_hmem_info_check_p2p_support_neuron(info);
 			break;
 		default:
 			break;
--- a/prov/efa/src/efa_hmem.h
+++ b/prov/efa/src/efa_hmem.h
@@ -31,15 +31,33 @@ static const enum fi_hmem_iface efa_hmem_ifaces[] = {
 	FI_HMEM_SYNAPSEAI
 };
 
+enum efa_dmabuf_support {
+	EFA_DMABUF_NOT_SUPPORTED,
+	EFA_DMABUF_SUPPORTED,
+	EFA_DMABUF_ASSUMED
+};
+
 struct efa_hmem_info {
 	bool initialized; 	/* do we support it at all */
 	bool p2p_supported_by_device;	/* do we support p2p with this device */
+	bool dmabuf_fallback_enabled;
+	enum efa_dmabuf_support dmabuf_supported_by_device;	/* do we support dmabuf with this device */
 
 	size_t max_medium_msg_size;
 	size_t runt_size;
 	size_t min_read_msg_size;
 	size_t min_read_write_size;
 };
+
+#define DMABUF_IS_SUPPORTED(info) \
+	((info)->dmabuf_supported_by_device == EFA_DMABUF_SUPPORTED || \
+	 (info)->dmabuf_supported_by_device == EFA_DMABUF_ASSUMED)
+
+#define DMABUF_IS_NOT_SUPPORTED(info) \
+	((info)->dmabuf_supported_by_device == EFA_DMABUF_NOT_SUPPORTED)
+
+#define DMABUF_IS_ASSUMED(info) \
+	((info)->dmabuf_supported_by_device == EFA_DMABUF_ASSUMED)
 
 extern struct efa_hmem_info	g_efa_hmem_info[OFI_HMEM_MAX];
 
--- a/prov/efa/src/efa_mr.c
+++ b/prov/efa/src/efa_mr.c
@@ -500,6 +500,23 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
 }
 
 #endif
+
+static inline void efa_mark_dmabuf_fail(struct efa_hmem_info *i)
+{
+	if (DMABUF_IS_NOT_SUPPORTED(i)) {
+		return;
+	} else if (DMABUF_IS_ASSUMED(i)) {
+		EFA_INFO(FI_LOG_MR,
+			 "assumed dmabuf support for %s was incorrect\n",
+			 fi_tostr(i, FI_TYPE_HMEM_IFACE));
+	} else {
+		EFA_WARN(FI_LOG_MR,
+			 "Explicit dmabuf support for %s was incorrect\n",
+			 fi_tostr(i, FI_TYPE_HMEM_IFACE));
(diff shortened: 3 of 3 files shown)


src/hmem.c
```diff
@@ -831,3 +831,34 @@ int ofi_hmem_put_dmabuf_fd(enum fi_hmem_iface iface, int fd)
...
+ * @brief Check if DMABUF is enabled for a specific HMEM interface
+ *
+ * This function checks the environment variables to determine if DMABUF
+ * should be used for the specified HMEM interface. It respects both
+ * EFA-specific and core libfabric environment variables.
+ *
+ * @param[in] iface The HMEM interface to check
+ * @return true if DMABUF is enabled for the interface, false otherwise
+ */
+bool efa_hmem_is_dmabuf_env_var_enabled(enum fi_hmem_iface iface)
```
jiaxiyan: All the functions in this file start with `ofi_hmem` since this is a common file.

src/hmem_cuda.c
```diff
@@ -959,13 +959,16 @@ bool cuda_is_gdrcopy_enabled(void)
 	return cuda_attr.use_gdrcopy;
 }
 
-bool cuda_is_dmabuf_supported(void)
+bool cuda_is_dmabuf_requested(void)
 {
 	int use_dmabuf = 1;
-
 	fi_param_get_bool(NULL, "hmem_cuda_use_dmabuf", &use_dmabuf);
+	return use_dmabuf != 0;
```
jiaxiyan: Can you help fix the unresolved comment in https://github.com/ofiwg/libfabric/pull/11465? Just return `use_dmabuf` here.

include/ofi_hmem.h
```diff
@@ -467,5 +467,6 @@ int ofi_hmem_dev_reg_copy_from_hmem(enum fi_hmem_iface iface, uint64_t handle,
 int ofi_hmem_get_dmabuf_fd(enum fi_hmem_iface, const void *addr, uint64_t size,
 			   int *fd, uint64_t *offset);
 int ofi_hmem_put_dmabuf_fd(enum fi_hmem_iface iface, int fd);
+bool efa_hmem_is_dmabuf_env_var_enabled(enum fi_hmem_iface iface);
```
jiaxiyan: All the functions in this file should start with `ofi_hmem` since it is a common file.

prov/efa/src/efa_hmem.c
```diff
@@ -299,15 +306,29 @@ static inline void efa_hmem_info_check_p2p_support_neuron(struct efa_hmem_info *
...
-	ret = neuron_get_dmabuf_fd(ptr, (uint64_t)len, &dmabuf_fd, &offset);
-	if (ret == FI_SUCCESS) {
-		ibv_mr = ibv_reg_dmabuf_mr(
-					ibv_pd, offset,
-					len, (uint64_t)ptr, dmabuf_fd, ibv_access);
-	} else if (ret == -FI_EOPNOTSUPP) {
-		EFA_INFO(FI_LOG_MR,
-			"Unable to retrieve dmabuf fd of Neuron device buffer, "
-			"Fall back to ibv_reg_mr\n");
+	if (efa_hmem_is_dmabuf_env_var_enabled(FI_HMEM_NEURON)) {
```
jiaxiyan: This change can be removed from this commit now.

charlesstoll: Squashed the last 2 patches together. Please take another look

include/ofi_hmem.h
charlesstoll: good catch. Done

src/hmem_cuda.c
charlesstoll: Fixed.

src/hmem.c
```diff
@@ -831,3 +831,34 @@ int ofi_hmem_put_dmabuf_fd(enum fi_hmem_iface iface, int fd)
...
+ * @brief Check if DMABUF is enabled for a specific HMEM interface
+ *
+ * This function checks the environment variables to determine if DMABUF
+ * should be used for the specified HMEM interface. It respects both
+ * EFA-specific and core libfabric environment variables.
+ *
+ * @param[in] iface The HMEM interface to check
+ * @return true if DMABUF is enabled for the interface, false otherwise
+ */
+bool ofi_hmem_is_dmabuf_env_var_enabled(enum fi_hmem_iface iface)
```
jiaxiyan, line 844: This common code change should be in the previous commit.

prov/efa/src/efa_hmem.c
```diff
@@ -366,8 +295,26 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
...
 
-	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
+	if (iface == FI_HMEM_SYNAPSEAI) {
+		info->p2p_supported_by_device = true;
+		if (ofi_hmem_is_dmabuf_env_var_enabled(FI_HMEM_SYNAPSEAI)) {
+			info->dmabuf_supported_by_device_b = true;
+		} else {
+			EFA_INFO(FI_LOG_CORE, "FI_HMEM_SYNAPSEAI DMABUF disabled by environment variable\n");
+		}
+	} else if (iface == FI_HMEM_NEURON) {
```
jiaxiyan: Can you combine them to `if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_NEURON)` since they share the same logic?

prov/efa/src/efa_mr.c
```diff
@@ -509,62 +509,69 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+
+    /* Explicit dmabuf registration */
+    if (flags & FI_MR_DMABUF) {
+        if (!mr_attr->dmabuf) {
+            EFA_WARN(FI_LOG_MR, "FI_MR_DMABUF set but mr_attr->dmabuf == NULL\n");
+            return NULL;
+        }
+        if (!g_efa_hmem_info[efa_mr->peer.iface].dmabuf_supported_by_device_b) {
+            EFA_WARN(FI_LOG_MR,
+                     "Requested FI_MR_DMABUF, but dmabuf not supported for iface=%d\n",
```
jiaxiyan: Use `fi_tostr(&efa_mr->peer.iface, FI_TYPE_HMEM_IFACE));` instead of integer

prov/efa/src/efa_mr.c
```diff
@@ -575,12 +582,17 @@ static struct ibv_mr *efa_mr_reg_ibv_mr(struct efa_mr *efa_mr, struct fi_mr_attr
...
+    } else {
+		return ibv_reg_mr(efa_mr->domain->ibv_pd,
+                      (void *)mr_attr->mr_iov->iov_base,
+                      mr_attr->mr_iov->iov_len, access);
 	}
 
-	return ibv_reg_mr(efa_mr->domain->ibv_pd,
-			(void *)mr_attr->mr_iov->iov_base,
-			mr_attr->mr_iov->iov_len, access);
+	return NULL;
```
jiaxiyan: I would move this to line 589

src/hmem.c
jiaxiyan: Btw you can add yourself as a co-author in the commit message

prov/efa/src/efa_mr.c
```diff
@@ -509,62 +509,69 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+                dmabuf_fd, access);
+            
+            /* Close the dmabuf file descriptor - it's no longer needed after registration */
+            (void) ofi_hmem_put_dmabuf_fd(efa_mr->peer.iface, dmabuf_fd);
+            
+            return dmabuf_mr;
+
+		/* TODO: need such fallback for cuda as well when FI_CUDA_API_PERMITTED is true */
+        } else if (ret == -FI_EOPNOTSUPP &&
+		   (efa_mr_is_neuron(efa_mr) || efa_mr_is_rocr(efa_mr))) {
```
shijin-aws: I do not think we want to fallback every time. We should stick to one mode and fail crisply when it is broken. Neuron is special here as we are suggested to move the initial dmabuf detection to the 1st fi_mr_regattr. So you need to have a enum instead of bool for your `dmabuf_supported_by_device_b` field: `supported`, `not_supported`, `unspec`.  And the fi_mr_reg call should look like 

```
if (g_efa_hmem_info[efa_mr->peer.iface].dmabuf_supported_by_device_b != `not_supported`) {
       ofi_hmem_get_dmabuf_fd 
       // ... the rest dmabuf reg code
      // if dmabuf_supported_by_device_b == unspec {
      //   dmabuf_supported_by_device_b = supported when reg ^^ succeeds
      //   dmabuf_supported_by_device_b = non_supported when reg ^^ fails
      // }
      //  
      // 
else  {
    // do non-dmabuf reg
}
return NULL;
```

prov/efa/src/efa_hmem.c
charlesstoll: Done

prov/efa/src/efa_mr.c
charlesstoll: Done

prov/efa/src/efa_mr.c
charlesstoll: Done

src/hmem.c
charlesstoll: Done

prov/efa/src/efa_mr.c
charlesstoll: Done. I used supported, not supported, and assumed--let me know if you would like me to use unspec instead. main thing to note here is that DMABUF_IS_SUPPORTED will return true for both supported or assumed, meaning we get treat assumed as the same as supported except for in cases where we want to explicitly check assumed. Also made the change to mark this if the main dmabuf api path fails, let me know if we should remove that but I didn't like the idea of leaving the flag as assumed even after we have confirmed it working/not working.

updated the git commit in accordance with this change

prov/efa/src/efa_mr.c
```diff
@@ -500,6 +500,17 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
 }
 
 #endif
+
+static inline void efa_mark_dmabuf_fail(struct efa_hmem_info *i)
+{
+	if (!DMABUF_IS_ASSUMED(i))
+		return;
+	EFA_INFO(FI_LOG_MR,
+		 "assumed dmabuf support for %s was incorrec\n",
```
shijin-aws: `incorrec` -> `incorrect`

src/hmem.c
charlesstoll: changed nick to co-author so that if someone does a git blame, they will bug me instead of him :)

prov/efa/src/efa_hmem.c
```diff
@@ -366,8 +295,21 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
 	}
 
 	info->initialized = true;
+	info->dmabuf_supported_by_device = EFA_DMABUF_NOT_SUPPORTED;
 
-	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
+	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_NEURON) {
+		/* Neuron does not allow allocating dmabufs this early on
```
shijin-aws: Let's rephrase it like `It's not recommended to allocate neuron buffer in the initialization`

prov/efa/src/efa_hmem.h
```diff
@@ -31,16 +31,33 @@ static const enum fi_hmem_iface efa_hmem_ifaces[] = {
...
 
 	size_t max_medium_msg_size;
 	size_t runt_size;
 	size_t min_read_msg_size;
 	size_t min_read_write_size;
 };
 
+#define DMABUF_IS_SUPPORTED(info) \
+	((info)->dmabuf_supported_by_device == EFA_DMABUF_SUPPORTED || \
+	 (info)->dmabuf_supported_by_device == EFA_DMABUF_ASSUMED)
```
shijin-aws, line 54: This is just `!DMABUF_IS_NOT_SUPPORTED(info)`. We probably do not have to have both since they are mutually exclusive

prov/efa/src/efa_mr.c
```diff
@@ -509,78 +520,111 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+			} else {
+				p_info->dmabuf_supported_by_device = EFA_DMABUF_SUPPORTED;
+			}
+
+			return dmabuf_mr;
+
+		/* TODO: need such fallback for cuda as well when
+		 * FI_CUDA_API_PERMITTED is true
+		 */
+		} else if (ret == -FI_EOPNOTSUPP &&
```
shijin-aws: This comment was saying we should do dmabuf for cuda first and then fallback to p2p approach. Before your change, we even do not do dmabuf for cuda in the VA path. 

Based on what you do in the PR, I think you need to remove this comment to make the cuda join the same story of neuron and rocr.

shijin-aws: Nvm, I think I was wrong, you already have a dmabuf detection for cuda at the domain init anyway and it should be covered by L623 anyway.

prov/efa/src/efa_mr.c
charlesstoll: Done

prov/efa/src/efa_hmem.c
charlesstoll: Done.

prov/efa/src/efa_hmem.h
charlesstoll: I'm not a fan of !DMABUF_IS_NOT_SUPPORTED() because of the double negative, gets difficult to read in my opinion. Also, if we later on add another enum value, all our !DMABUF_IS_NOT_SUPPORTED() calls will be incorrect.

I'll change it if you want though, not a big deal to me.

prov/efa/src/efa_hmem.c
```diff
@@ -143,21 +143,28 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
...
-				"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
+				"Unable to retrieve dmabuf fd of CUDA device buffer: %d. "
+				"Fall back to ibv_reg_mr\n", ret);
 			ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
 		}
 	} else {
-		EFA_INFO(FI_LOG_CORE,
-			"Unable to retrieve dmabuf fd of CUDA device buffer: %d. "
-			"Fall back to ibv_reg_mr\n", ret);
+		EFA_INFO(FI_LOG_CORE, "FI_HMEM_CUDA DMABUF disabled by environment variable\n");
```
sunkuamzn: nit: I would change this to either
"DMABUF for CUDA disabled by environment variable"
or
"FI_HMEM_CUDA_USE_DMABUF set to false, so not using DMABUF on CUDA"

prov/efa/src/efa_hmem.c
```diff
@@ -366,8 +295,21 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
...
 
-	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
+	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_NEURON) {
+		/* It is not recommended to allocate neuron buffs this
+		 * early in initialization, so we must skip the explicit
+		 * check to see if p2p will work. Instead, assume it works.
+		 */
+		info->p2p_supported_by_device = true;
+		if (ofi_hmem_is_dmabuf_env_var_enabled(iface)) {
+			info->dmabuf_supported_by_device = EFA_DMABUF_ASSUMED;
```
sunkuamzn: I think it's better to set the default value of `info->dmabuf_supported_by_device` to `EFA_DMABUF_ASSUMED` in `efa_hmem_info_init_iface` so that all ifaces are initialized to assumed by default

prov/efa/src/efa_mr.c
```diff
@@ -509,78 +520,108 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
 
-	if (flags & FI_MR_DMABUF)
-		return efa_mr_reg_ibv_dmabuf_mr(
+	/* Explicit dmabuf registration */
+	if (flags & FI_MR_DMABUF) {
+		if (!mr_attr->dmabuf) {
+			EFA_WARN(FI_LOG_MR, "FI_MR_DMABUF set but mr_attr->dmabuf == NULL\n");
+			return NULL;
+		}
+		if (DMABUF_IS_NOT_SUPPORTED(p_info)) {
```
sunkuamzn, line 545: If we don't initialize `efa_hmem_info->dmabuf_supported_by_device`, then we could fail here with ROCm even if ROCm supports dmabuf

jiaxiyan:
bot:aws:retest

prov/efa/src/efa_hmem.c
charlesstoll: printing out the actual environment variable name is a good call

charlesstoll: Done

prov/efa/src/efa_hmem.c
charlesstoll: Done

prov/efa/src/efa_mr.c
charlesstoll: Should be fixed with defaulting to ASSUMED.

prov/efa/src/efa_hmem.c
sunkuamzn: this line can be removed now

sunkuamzn (dismissed):

sunkuamzn:
@j-xiong @aingerson could you review the changes to the common code?

prov/efa/src/efa_mr.c
```diff
@@ -509,78 +520,108 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
 			(void) ofi_hmem_put_dmabuf_fd(efa_mr->peer.iface, dmabuf_fd);
-			return ibv_mr;
-		} else if (ret == -FI_EOPNOTSUPP) {
+			if (!dmabuf_mr)  {
+				efa_mark_dmabuf_fail(p_info);
+			} else {
+				p_info->dmabuf_supported_by_device = EFA_DMABUF_SUPPORTED;
+			}
+
+			return dmabuf_mr;
```
shijin-aws, line 609: So for neuron, for the 1st fi_mr_reg, we are not falling back to non-dmabuf reg, when `ofi_hmem_get_dmabuf_fd` succeeds while `efa_mr_reg_ibv_dmabuf_mr` failed?

prov/efa/src/efa_mr.c
```diff
@@ -509,78 +520,108 @@ struct ibv_mr *efa_mr_reg_ibv_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
...
+
+		} else if (ret == -FI_EOPNOTSUPP &&
+			   (efa_mr_is_neuron(efa_mr) || efa_mr_is_rocr(efa_mr))) {
```
shijin-aws: I am not sure whether we need such fallback for rocr since we can detect its dmabuf status during domain init?

prov/efa/src/efa_hmem.c
```diff
@@ -367,7 +367,7 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
 
 	info->initialized = true;
 
-	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_SYSTEM) {
+	if (iface == FI_HMEM_SYNAPSEAI || iface == FI_HMEM_NEURON) {
```
j-xiong: This doesn't belong to this commit.

src/hmem_ze.c
```diff
@@ -1088,6 +1091,14 @@ int ze_hmem_get_dmabuf_fd(const void *addr, uint64_t size, int *fd,
 	return 0;
 }
 
+bool ze_is_dmabuf_requested(void)
+{
+	int use_dmabuf = 0;
```
j-xiong: for ze the default should be 1.

sielicki:
Slightly related, I posted https://lkml.org/lkml/2026/1/7/1950 last week, motivated by roughly the same problems that are described in the commit message.

```
hmem: Add dmabuf support detection infrastructure

Problem:
  - Need infrastructure to detect and query dmabuf support per HMEM
	interface
  - Providers need to know if dmabuf is available before attempting
	registration
```

The kernel already tracks whether a module is importing the dmabuf namespace or not, and enforces that namespaces are consistent when modules are loaded (or refuses to load them, if not), but doesn't actually make it possible for userspace to query that retroactively after modules have been loaded.

A better fictional way to do this, IMHO, is for the kernel to mark which kernel modules are dmabuf namespace importers (already done), and for libfabric to simply check both ends through sysfs (possible after above patch), and if both the hmem kernel module and the provider kernel module import the dmabuf namespace, you can safely assume support on both ends without needing to actually attempt registration.

src/hmem_ze.c
charlesstoll: Done.

prov/efa/src/efa_mr.c
charlesstoll: I changed the code so that all failing calls will have a fallback after checking a new interface flag dmabuf_fallback_enabled

prov/efa/src/efa_mr.c
charlesstoll: explicit iface checks have been removed. we can now just use the initialized values which are all default to false except neuron which we overwrite to true

prov/efa/src/efa_hmem.c
charlesstoll: This was a rebase error, sorry. I have fixed it in my latest patch set.

prov/efa/src/efa_mr.c
charlesstoll: changed FI_HMEM_SYSTEM to use fallback path as well

prov/efa/src/efa_hmem.c
jiaxiyan: `EFA_DMABUF_ASSUMED` is already assigned above.

prov/efa/src/efa_mr.c
jiaxiyan: Why not set `dmabuf_supported = false` for FI_HMEM_SYSTEM?

prov/efa/src/efa_hmem.c
charlesstoll: Ok, I had thought about this earlier and had a question which I think I deleted for some reason. If we are getting rid of the ofi_hmem_is_dmabuf_env_var_enabled() call, then the only other call to this function is when it is just hard-coded to check for cuda support. So my question is, do we actually need the first patch? We are just assuming dmabuf support for everyone and are doing fallback for those we can't explicitly check. In this case, doesn't it make more sense to either remove the dmabuf requested environment variables altogether or to change them to dmabuf_fallback_enable?

jiaxiyan: Then you need to set `dmabuf_supported_by_device = EFA_DMABUF_NOT_SUPPORTED` when `!ofi_hmem_is_dmabuf_env_var_enabled`

charlesstoll: Ah, gotcha. thought you meant remove the whole if/else statement. so we would essentially be turning it into a disable flag. Do you think we should rename the environment variables then to better reflect that? Not sure what conventions are around stuff like this

prov/efa/src/efa_mr.c
charlesstoll: Oh, thats definitely better. Done.

src/hmem.c
```diff
@@ -831,3 +831,33 @@ int ofi_hmem_put_dmabuf_fd(enum fi_hmem_iface iface, int fd)
 {
 	return hmem_ops[iface].put_dmabuf_fd(fd);
 }
+
+/**
+ * @brief Check if DMABUF is enabled for a specific HMEM interface
+ *
+ * This function checks the environment variables to determine if DMABUF
+ * should be used for the specified HMEM interface. It respects both
+ * EFA-specific and core libfabric environment variables.
```
jiaxiyan: remove `EFA-specific`

prov/efa/src/efa_hmem.c
```diff
@@ -143,21 +143,28 @@ static inline void efa_hmem_info_check_p2p_support_cuda(struct efa_hmem_info *in
...
+			ibv_mr = ibv_reg_dmabuf_mr(ibv_pd, dmabuf_offset,
+						   len, (uint64_t)ptr, dmabuf_fd, ibv_access);
+			(void)ofi_hmem_put_dmabuf_fd(FI_HMEM_CUDA, dmabuf_fd);
+			if (!ibv_mr) {
+				EFA_INFO(FI_LOG_CORE,
+					"Unable to register CUDA device buffer via dmabuf: %s. "
+					"Fall back to ibv_reg_mr\n", fi_strerror(-errno));
+				ibv_mr = ibv_reg_mr(ibv_pd, ptr, len, ibv_access);
+			} else {
+				info->dmabuf_supported_by_device = EFA_DMABUF_SUPPORTED;
```
jiaxiyan, line 158: Do you need similar logic to set dmabuf_supported_by_device in `efa_hmem_info_check_p2p_support_rocr`?

prov/efa/src/efa_hmem.c
jiaxiyan: I wouldn't rename the environment variables because there could be applications using it today

charlesstoll: Oh, didn't realize that these were parameters used elsewhere and that we are just adding them to use elsewhere. Makes sense.

charlesstoll: Done.

src/hmem.c
charlesstoll: done.

prov/efa/src/efa_hmem.c
```diff
@@ -366,9 +295,28 @@ efa_hmem_info_init_iface(enum fi_hmem_iface iface)
...
+	    iface == FI_HMEM_NEURON) {
+		/* It is not recommended to allocate neuron buffs this
+		 * early in initialization, so we must skip the explicit
+		 * check to see if p2p will work. Instead, assume it works.
+		 * and set fallback to true
+		 */
+		if (iface == FI_HMEM_NEURON)
+			info->dmabuf_fallback_enabled = true;
+		if (iface == FI_HMEM_SYSTEM)
+			info->dmabuf_supported_by_device = EFA_DMABUF_NOT_SUPPORTED;
```
jiaxiyan: I think FI_HMEM_SYSTEM is already covered by `ofi_hmem_is_dmabuf_env_var_enabled` below

prov/efa/src/efa_hmem.c
charlesstoll: Done.

prov/efa/src/efa_hmem.c
charlesstoll: Good call. Done.

prov/efa/src/efa_hmem.c
jiaxiyan: You also need to check `ofi_hmem_is_dmabuf_env_var_enabled` for ROCR. 
Since `dmabuf_supported_by_device` is default to ASSUMED in `efa_hmem_info_init_iface`, CUDA and ROCR need to set `dmabuf_supported_by_device` to UNSUPPORTED if `ibv_reg_dmabuf_mr` failed here.

jiaxiyan: It would be nice to combine efa_hmem_info_check_p2p_support for CUDA and ROCR using the generic ofi_hmem functions. Something to consider as a follow up, not a blocker for this PR.

charlesstoll: Makes sense. although above comment about maybe being able to skip these alloc/dealloc tests in favor of looking at the kernel modules directly and if they include DMABUF might be a cleaner way to go, and then we would be getting rid of these functions anyways. Maybe something we should talk about as a way to move in the future for feature detection

src/hmem_rocr.c
```diff
@@ -1081,8 +1081,8 @@ int rocr_hmem_init(void)
 		"Threshold for switching to hsa memcpy for device-to-host"
 		"  copies. (Default 16384");
 
-	fi_param_define(NULL, "hmem_rocr_use_dmabuf", FI_PARAM_INT,
-			"Use dma-buf for sharing buffer with hardware. (default:0)");
+	fi_param_define(NULL, "hmem_rocr_use_dmabuf", FI_PARAM_BOOL,
+			"Use dma-buf for sharing buffer with hardware. (default: true)");
```
jiaxiyan: It seems ROCR is default as false. I am not sure which provider set this. Maybe we shouldn't check the env for ROCR as we want dmabuf to be the default in efa provider.

jiaxiyan (dismissed):

src/hmem_rocr.c
j-xiong: The default was original set with commit 43e92bf82a0bfd22718f73a2f26467bceead7dab. I don't see any issue changing the default to true.

jiaxiyan (approved):

PR #11808, opened 2026-01-14
prov/efa, fabtests: Fix double-free pkt-entry at efa_rdm_pke_proc_matched_longread_rtm

alekswn:
Skipping releasing pkt_entry if efa_rdm_pke_post_remote_read_or_nack returned error. The entry is going to be released outside the function.

### Testing

Before the change following fabtest would segfault:
Server:
```
 ./fi_rdm_tagged_pingpong -p efa -E -U -S 2000000
```

Client:
```
./fi_rdm_tagged_pingpong -p efa -E -S 2000001 -U  p5en-odcr-queue-dy-p5en48xlarge-9
```

After the change server detects CQ error:
```
[error] fabtests:common/shared.c:3064: cq_readerr 265 (Truncation error), provider errno: -265 (Unknown error)
```

*Update:*

I'm also adding a new fabtest cmdline argument `--expect-error` to facilitate testing of unhappy paths. Therefore we can run test in CI as:

Server:
```
./fi_rdm_tagged_pingpong -p efa -E -U -S 2000000 --expect-error 265
```

Client:
```
./fi_rdm_tagged_pingpong -p efa -E -U -S 2000001 --timeout 1 --expect-error 61 <server IP>
```

### DIFF
--- a/fabtests/pytest/efa/test_rdm.py
+++ b/fabtests/pytest/efa/test_rdm.py
@@ -302,3 +302,19 @@ def test_implicit_av_limited_av_size_unexpected_path(cmdline_args, msg_size):
     efa_run_client_server_test(cmdline_args, test_cmd, "short",
                                "transmit_complete", "host_to_host", "all", fabric="efa",
                                additional_env=additional_env, timeout=15)
+
+@pytest.mark.functional
+def test_rmd_tagged_pingpong_truncate_error(cmdline_args, completion_semantic):
+    if cmdline_args.server_id == cmdline_args.client_id:
+        pytest.skip("client hits an assertion in debug mode in case of intra-node")
+    executable = {"server": "fi_rdm_tagged_pingpong --timeout 1",
+                  "client": "fi_rdm_tagged_pingpong --expect-error 265"}
+    message_size = {"server": 2000001,
+                    "client": 2000000}
+    might_fail = {"server": True,
+                  "client": False}
+
+    efa_run_client_server_test(cmdline_args, executable, "short",
+                               completion_semantic, "host_to_host",
+                               message_size=message_size, fabric="efa",
+                               timeout=15, might_fail=might_fail)


jiaxiyan:
Can you add the fabtest so we can test it as part of this PR?

shijin-aws:
@j-xiong can u review https://github.com/ofiwg/libfabric/pull/11808/commits/bb2facea48d13ce55208f773c4fb29b4ecfda638

fabtests/pytest/common.py
```diff
@@ -611,7 +613,7 @@ def run(self):
                               self._cmdline_args.strict_fabtests_mode)
 
 
-class MultinodeTest(ClientServerTest):
+class MultinodeTest:
```
jiaxiyan: Did you test you can remove this? It is using `_run_client_command` from the parent class

fabtests/pytest/efa/test_rdm.py
```diff
@@ -302,3 +302,15 @@ def test_implicit_av_limited_av_size_unexpected_path(cmdline_args, msg_size):
     efa_run_client_server_test(cmdline_args, test_cmd, "short",
                                "transmit_complete", "host_to_host", "all", fabric="efa",
                                additional_env=additional_env, timeout=15)
+
+@pytest.mark.functional
+def test_rmd_tagged_pingpong_truncate_error(cmdline_args):
+    from common import ClientServerTest
+
+    executable = {"server": "fi_rdm_tagged_pingpong --expect-error 265",
```
jiaxiyan: Do you need the `delivery_complete`?

fabtests/pytest/common.py
alekswn: Good catch! It does, will bring it back.

fabtests/pytest/efa/test_rdm.py
alekswn: Seems like does not make any difference in this case.

alekswn: Test works both ways

jiaxiyan: Then we should add the fixture `completion_semantic` so we test both long read and runt read. Right now it is default to `completion_semantic="transmit_complete"`, which doesn't gurantee calling `efa_rdm_pke_proc_matched_longread_rtm`

j-xiong (commented):
The common code changes look good to me.

alekswn:
@j-xiong Do you happen to know if there is any particular reason to have timeout in  `ft_get_cq_comp()` for  rx completions, but not on tx completions in fabtests?

We have to let tx side time-out for negative test on rx side. I have added that change in the new revision.

fabtests/common/shared.c
```diff
@@ -2915,9 +2915,9 @@ int ft_get_tx_comp(uint64_t total)
 	int ret;
 
 	if (opts.options & FT_OPT_TX_CQ) {
-		ret = ft_get_cq_comp(txcq, &tx_cq_cntr, total, -1);
+		ret = ft_get_cq_comp(txcq, &tx_cq_cntr, total, timeout);
```
jiaxiyan, line 2918: Should this be a separate commit?

fabtests/common/shared.c
```diff
@@ -2915,9 +2915,9 @@ int ft_get_tx_comp(uint64_t total)
 	int ret;
```
jiaxiyan, line 2916: nit: you have an extra `fix` at the end of this commit message.

prov/efa/src/rdm/efa_rdm_pke_rtm.c
```diff
@@ -1207,8 +1207,9 @@ ssize_t efa_rdm_pke_proc_matched_longread_rtm(struct efa_rdm_pke *pkt_entry)
 		    (size_t) rxe->cq_entry.op_context, rxe->total_len);
 
 	err = efa_rdm_pke_post_remote_read_or_nack(ep, pkt_entry, rxe);
-
-	efa_rdm_pke_release_rx(pkt_entry);
+	if (OFI_LIKELY(!err)) {
```
jiaxiyan, line 1210: Is this really unlikely? I see many cases where this will return err.

alekswn: I had the impression we usually ask compiler to optimize performance for happy path maybe at the cost of additional latency if processing errors. This is exactly what is happening here.

fabtests/common/shared.c
alekswn: Can make it separate commit also it's optional. Would just make test timeout faster.

fabtests/common/shared.c
alekswn: Oops, will update that

j-xiong:
> @j-xiong Do you happen to know if there is any particular reason to have timeout in `ft_get_cq_comp()` for rx completions, but not on tx completions in fabtests?
> 
> We have to let tx side time-out for negative test on rx side. I have added that change in the new revision.

My guess is that tx is unlikely to timeout. But I don't think adding timeout to tx completion polling would hurt anything.

j-xiong:
There was one failed test with mpichtestsuite over rxm-verbs:
```
		Program comm_idup exited without No Errors
		Unexpected output in comm_idup:
		    [mpiexec] APPLICATION TIMED OUT, TIMEOUT = 180s
```

alekswn:
> There was one failed test with mpichtestsuite over rxm-verbs:
> 
> ```
> 		Program comm_idup exited without No Errors
> 		Unexpected output in comm_idup:
> 		    [mpiexec] APPLICATION TIMED OUT, TIMEOUT = 180s
> ```

Is there a way to check if it had failed for previous commit [096faf2](https://github.com/ofiwg/libfabric/commit/096faf2283417b7e78cbfdb058dc4f461c8fdfec) 
I don't remember seeing it failed. The difference between commits is in efa-specific  pytest code only.

alekswn:
Actually there is a tests report on the commit showing Intel CI has passed for that one.

PR #11816, opened 2026-01-20
 prov/efa: Add debug_info_vec for duplicate completion detection 

yinliaws:
## Add duplicate completion detection for EFA RDM provider

   ### Problem
   Production workloads encounter "Packet already processed" errors indicating
   duplicate completions from rdma-core or packet reuse bugs in libfabric.

   ### Solution
   Add debug instrumentation to track packet lifecycle:
   - 3-event circular buffer per packet (POST/COMPLETION history)
   - Records: generation, event type, QP number, qkey
   - Detects duplicate completions and logs full diagnostic info
   - Identifies bug source (rdma-core vs libfabric) via wr_id tracking

   ### Changes
   - efa_rdm_pke.h: Add debug_info structure and circular buffer
   - efa_rdm_pke.c: Record POST events with qkey query
   - efa_rdm_cq.c: Record COMPLETION events, detect duplicates

   ### Testing
   Validated on EC2 c5n.18xlarge with EFA hardware using fi_rdm_pingpong.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_cq.c
+++ b/prov/efa/src/rdm/efa_rdm_cq.c
@@ -230,6 +230,14 @@ static void efa_rdm_cq_proc_ibv_recv_rdma_with_imm_completion(
 		 */
 		assert(pkt_entry);
 		ep->efa_rx_pkts_posted--;
+#if ENABLE_DEBUG
+		/* Record RECV_RDMA_WITH_IMM event */
+		efa_rdm_pke_record_debug_info(pkt_entry,
+		                               ep->base_ep.qp->qp_num,
+		                               ep->base_ep.qp->qkey,
+		                               pkt_entry->gen,
+		                               EFA_RDM_PKE_DEBUG_EVENT_RECV_RDMA_WITH_IMM);
+#endif
 		efa_rdm_cq_increment_pkt_entry_gen(pkt_entry);
 		efa_rdm_pke_release_rx(pkt_entry);
 	}
@@ -523,6 +531,15 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct 
 	struct efa_rdm_base_hdr *base_hdr;
 	uint32_t imm_data = 0;
 	bool has_imm_data = false;
+
+#if ENABLE_DEBUG
+	/* Record RECV_COMPLETION event */
+	efa_rdm_pke_record_debug_info(pkt_entry,
+	                               ep->base_ep.qp->qp_num,
+	                               ep->base_ep.qp->qkey,
+	                               pkt_entry->gen,
+	                               EFA_RDM_PKE_DEBUG_EVENT_RECV_COMPLETION);
+#endif
 
 	EFA_DBG(FI_LOG_CQ, "Processing receive completion for packet %p\n", pkt_entry);
 
--- a/prov/efa/src/rdm/efa_rdm_ep_utils.c
+++ b/prov/efa/src/rdm/efa_rdm_ep_utils.c
@@ -381,6 +381,24 @@ void efa_rdm_ep_record_tx_op_submitted(struct efa_rdm_ep *ep, struct efa_rdm_pke
 	}
 #if ENABLE_DEBUG
 	ep->efa_total_posted_tx_ops++;
+	/* Record post event based on operation type */
+	int event_type;
+	switch (ope->op) {
+	case ofi_op_read_req:
+		event_type = EFA_RDM_PKE_DEBUG_EVENT_READ_POST;
+		break;
+	case ofi_op_write:
+		event_type = EFA_RDM_PKE_DEBUG_EVENT_WRITE_POST;
+		break;
+	default:
+		event_type = EFA_RDM_PKE_DEBUG_EVENT_SEND_POST;
+		break;
+	}
+	efa_rdm_pke_record_debug_info(pkt_entry,
+	                               ep->base_ep.qp->qp_num,
+	                               ep->base_ep.qp->qkey,
+	                               pkt_entry->gen,
+	                               event_type);
 #endif
 }
 
@@ -420,6 +438,31 @@ void efa_rdm_ep_record_tx_op_submitted(struct efa_rdm_ep *ep, struct efa_rdm_pke
 void efa_rdm_ep_record_tx_op_completed(struct efa_rdm_ep *ep, struct efa_rdm_pke *pkt_entry)
 {
 	struct efa_rdm_ope *ope = NULL;
+
+#if ENABLE_DEBUG
+	/*
+	 * Record completion event based on operation type.
+	 * Note: qp_num is truncated to 10 bits, gen to 6 bits.
+	 */
+	int event_type = EFA_RDM_PKE_DEBUG_EVENT_SEND_COMPLETION;
+	if (pkt_entry->ope) {
+		switch (pkt_entry->ope->op) {
+		case ofi_op_read_req:
+			event_type = EFA_RDM_PKE_DEBUG_EVENT_READ_COMPLETION;
+			break;
+		case ofi_op_write:
+			event_type = EFA_RDM_PKE_DEBUG_EVENT_WRITE_COMPLETION;
+			break;
+		default:
+			break;
+		}
+	}
+	efa_rdm_pke_record_debug_info(pkt_entry,
+	                               ep->base_ep.qp->qp_num,
+	                               ep->base_ep.qp->qkey,
+	                               pkt_entry->gen,
+	                               event_type);
+#endif
 
 	ope = pkt_entry->ope;
 	/*
--- a/prov/efa/src/rdm/efa_rdm_pke.c
+++ b/prov/efa/src/rdm/efa_rdm_pke.c
@@ -45,11 +45,49 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 	if (!pkt_entry)
 		return NULL;
 
+#ifdef ENABLE_EFA_POISONING
+	/* Preserve gen across poisoning */
+	uint8_t gen = pkt_entry->gen;
+#if ENABLE_DEBUG
+	/* Preserve debug_info pointer across poisoning to maintain packet history.
+	 * On first allocation from a freshly poisoned region, this will be 0xdeadbeef.
+	 * On reuse, this preserves the existing buffer. */
+	struct efa_rdm_pke_debug_info_buffer *debug_info = pkt_entry->debug_info;
+	/* If debug_info contains poison pattern, treat as NULL (uninitialized).
+	 * This happens when packets are allocated from freshly poisoned bufpool regions. */
+	if (((uintptr_t)debug_info & 0xffffffffUL) == 0xdeadbeefUL) {
+		debug_info = NULL;
+	}
+#endif
+	efa_rdm_poison_mem_region(pkt_entry, pkt_pool->attr.size);
+	pkt_entry->gen = gen;
+#if ENABLE_DEBUG
+	pkt_entry->debug_info = debug_info;
+#endif
+#endif
+	/* Without poisoning, debug_info pointer is naturally preserved in memory. */
+
 	pkt_entry->gen &= EFA_RDM_PACKET_GEN_MASK;
 	dlist_init(&pkt_entry->entry);
 
 #if ENABLE_DEBUG
 	dlist_init(&pkt_entry->dbg_entry);
+	/* Allocate debug info if not already allocated */
+	if (!pkt_entry->debug_info) {
+		pkt_entry->debug_info = ofi_buf_alloc(ep->pke_debug_info_pool);
+		if (!pkt_entry->debug_info) {
+			/* Debug info allocation failed from unlimited pool - indicates heap exhaustion.
+			 * Write EQ error since retrying won't help (debug_info is never released). */
+			EFA_WARN(FI_LOG_EP_CTRL,
+				"Failed to allocate debug_info buffer from unlimited pool - heap exhaustion likely\n");
+			efa_base_ep_write_eq_error(&ep->base_ep, FI_ENOMEM, FI_EFA_ERR_OOM);
+			efa_rdm_pke_release(pkt_entry);
+			return NULL;
+		}
+		pkt_entry->debug_info->counter = 0;
+		memset(pkt_entry->debug_info->entries, 0, 
+		       sizeof(pkt_entry->debug_info->entries));
+	}
 #endif
 	/* Initialize necessary fields in pkt_entry.
 	 * The memory region allocated by ofi_buf_alloc_ex is not initalized.
@@ -93,12 +131,18 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 void efa_rdm_pke_release(struct efa_rdm_pke *pkt_entry)
 {
 #ifdef ENABLE_EFA_POISONING
-	/* Restore pkt_entry->gen after poisoning. Otherwise, all pkts will have
-	 * the same gen when poisoning is enabled. */
+	/* Preserve gen and debug_info pointer across poisoning to maintain packet history */
 	uint8_t gen = pkt_entry->gen;
+#if ENABLE_DEBUG
+	struct efa_rdm_pke_debug_info_buffer *debug_info = pkt_entry->debug_info;
+#endif
 	efa_rdm_poison_mem_region(pkt_entry, ofi_buf_pool(pkt_entry)->attr.size);
 	pkt_entry->gen = gen;
+#if ENABLE_DEBUG
+	pkt_entry->debug_info = debug_info;
 #endif
+#endif
+	/* Without poisoning, debug_info pointer is naturally preserved in memory. */
 	pkt_entry->flags = 0;
 	ofi_buf_free(pkt_entry);
 }
@@ -708,6 +752,15 @@ ssize_t efa_rdm_pke_recvv(struct efa_rdm_pke **pke_vec,
 		recv_wr = &ep->base_ep.efa_recv_wr_vec[i];
 		recv_wr->wr.wr_id = efa_rdm_pke_get_wr_id(pke_vec[i]);
 
+#if ENABLE_DEBUG
+		/* Record RECV_POST event */
+		efa_rdm_pke_record_debug_info(pke_vec[i],
+		                               ep->base_ep.qp->qp_num,
+		                               ep->base_ep.qp->qkey,
+		                               pke_vec[i]->gen,
+		                               EFA_RDM_PKE_DEBUG_EVENT_RECV_POST);
+#endif
+
 		recv_wr->wr.num_sge = 1;
 		recv_wr->wr.sg_list = recv_wr->sge;
 		recv_wr->wr.sg_list[0].length = pke_vec[i]->pkt_size;
@@ -787,3 +840,50 @@ ssize_t efa_rdm_pke_user_recvv(struct efa_rdm_pke **pke_vec,
 
 	return err;
 }
+
+
+#if ENABLE_DEBUG
+/* Compile-time assertion that debug_info gen field can hold all possible gen values */
+_Static_assert(EFA_RDM_PKE_DEBUG_GEN_MASK >= EFA_RDM_PACKET_GEN_MASK, 
+               "DEBUG_GEN_BITS insufficient to hold EFA_RDM_PACKET_GEN_MASK");
+
+/**
+ * @brief Print debug info history for packet entry
+ *
+ * @param pkt_entry Packet entry
+ */
+void efa_rdm_pke_print_debug_info(struct efa_rdm_pke *pkt_entry)
+{
+	static const char *event_str[] = {
+		"SEND_POST",
+		"SEND_COMPLETION",
+		"RECV_POST",
+		"RECV_COMPLETION",
+		"READ_POST",
+		"READ_COMPLETION",
+		"WRITE_POST",
+		"WRITE_COMPLETION",
+		"RECV_RDMA_WITH_IMM"
+	};
+	int i, count;
+	int start_idx;
+
+	if (!pkt_entry->debug_info)
+		return;
+
+	count = MIN(EFA_RDM_PKE_DEBUG_INFO_SIZE, pkt_entry->debug_info->counter);
+	start_idx = (pkt_entry->debug_info->counter >= EFA_RDM_PKE_DEBUG_INFO_SIZE) ?
+	                (pkt_entry->debug_info->counter % EFA_RDM_PKE_DEBUG_INFO_SIZE) : 0;
+
+	for (i = 0; i < count; i++) {
+		int idx = (start_idx + i) % EFA_RDM_PKE_DEBUG_INFO_SIZE;
+		struct efa_rdm_pke_debug_info *info = &pkt_entry->debug_info->entries[idx];
+		uint8_t event = EFA_RDM_PKE_DEBUG_INFO_GET_EVENT(info);
+		EFA_WARN(FI_LOG_EP_DATA,
+		         "    [%d] counter=%u gen=%u qpn=%u qkey=%u (%s)\n",
+		         i, info->counter, EFA_RDM_PKE_DEBUG_INFO_GET_GEN(info),
+		         EFA_RDM_PKE_DEBUG_INFO_GET_QPN(info), info->qkey,
+		         event < EFA_RDM_PKE_DEBUG_EVENT_TYPE_COUNT ? event_str[event] : "UNKNOWN");
+	}
+}
+#endif
--- a/prov/efa/src/rdm/efa_rdm_pke.h
+++ b/prov/efa/src/rdm/efa_rdm_pke.h
@@ -75,6 +75,69 @@ enum efa_rdm_pke_alloc_type {
  * is not registered (when it is unexpected or out-of-order). A new packet entry will be cloned
  * using endpoint's read_copy_pkt_pool, whose memory was registered.
  */
+
+#if ENABLE_DEBUG
+/**
+ * @brief Debug info for tracking packet lifecycle events
+ *
+ * Used to diagnose "Packet already processed" errors by recording
+ * packet lifecycle events in a circular buffer.
+ *
+ * Size: 8 bytes per entry (naturally aligned)
+ * 48 entries provide sufficient history for diagnosis while keeping
+ * the debug buffer at a reasonable size (392 bytes total).
+ */
+#define EFA_RDM_PKE_DEBUG_INFO_SIZE 48
+
+/* Event type definitions (4 bits, 9 types) */
+#define EFA_RDM_PKE_DEBUG_EVENT_SEND_POST       0  /**< Send packet posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_SEND_COMPLETION 1  /**< Send packet completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_RECV_POST       2  /**< Receive buffer posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_RECV_COMPLETION 3  /**< Receive completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_READ_POST       4  /**< RDMA read posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_READ_COMPLETION 5  /**< RDMA read completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_WRITE_POST      6  /**< RDMA write posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_WRITE_COMPLETION 7 /**< RDMA write completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_RECV_RDMA_WITH_IMM 8 /**< Recv RDMA with immediate */
+#define EFA_RDM_PKE_DEBUG_EVENT_TYPE_COUNT      9  /**< Number of event types */
+
+struct efa_rdm_pke_debug_info {
+	uint32_t qkey;        /**< Queue key */
+	uint16_t qpn_gen;     /**< Bits [15:6]=qpn (10 bits), [5:0]=gen (6 bits) */
+	uint8_t counter;      /**< Incremental counter */
+	uint8_t event;        /**< Event type (9 types: 0-8) */
+};
+
+/* Compile-time size check */
+_Static_assert(sizeof(struct efa_rdm_pke_debug_info) == 8, "debug_info must be 8 bytes");
+
+/**
+ * @brief Debug info buffer - includes counter and circular buffer
+ */
+struct efa_rdm_pke_debug_info_buffer {
+	uint8_t counter;      /**< Global counter (0-255) */
+	uint8_t padding[7];   /**< Padding for alignment */
+	struct efa_rdm_pke_debug_info entries[EFA_RDM_PKE_DEBUG_INFO_SIZE]; /**< Circular buffer */
+};
+
+/* Bit field layout constants for qpn_gen field */
+#define EFA_RDM_PKE_DEBUG_GEN_BITS 6
+#define EFA_RDM_PKE_DEBUG_QPN_BITS 10
+#define EFA_RDM_PKE_DEBUG_QPN_SHIFT EFA_RDM_PKE_DEBUG_GEN_BITS
+#define EFA_RDM_PKE_DEBUG_GEN_MASK ((1 << EFA_RDM_PKE_DEBUG_GEN_BITS) - 1)
+#define EFA_RDM_PKE_DEBUG_QPN_MASK ((1 << EFA_RDM_PKE_DEBUG_QPN_BITS) - 1)
+
+/* Accessor macros for bit-packed qpn_gen field */
+#define EFA_RDM_PKE_DEBUG_INFO_GET_QPN(info)       (((info)->qpn_gen >> EFA_RDM_PKE_DEBUG_QPN_SHIFT) & EFA_RDM_PKE_DEBUG_QPN_MASK)
+#define EFA_RDM_PKE_DEBUG_INFO_GET_GEN(info)       ((info)->qpn_gen & EFA_RDM_PKE_DEBUG_GEN_MASK)
+#define EFA_RDM_PKE_DEBUG_INFO_GET_EVENT(info)     ((info)->event)
+
+#define EFA_RDM_PKE_DEBUG_INFO_SET_QPN_GEN(info, _qpn, _gen) \
+	do { \
+		(info)->qpn_gen = (((_qpn) & EFA_RDM_PKE_DEBUG_QPN_MASK) << EFA_RDM_PKE_DEBUG_QPN_SHIFT) | ((_gen) & EFA_RDM_PKE_DEBUG_GEN_MASK); \
+	} while(0)
+#endif
+
 struct efa_rdm_pke {
 	/**
 	 * entry to the linked list of outstanding/queued packet entries
@@ -183,6 +246,10 @@ struct efa_rdm_pke {
 	/**@brief Generation counter. It is incremented every time the packet is posted to rdma-core */
 	uint8_t gen;
 
+#if ENABLE_DEBUG
+	struct efa_rdm_pke_debug_info_buffer *debug_info; /**< Pointer to debug info buffer */
+#endif
+
 	/** @brief buffer that contains data that is going over wire
 	 *
 	 * @details
@@ -205,8 +272,10 @@ struct efa_rdm_pke {
 
 #if defined(static_assert)
 static_assert(sizeof (struct efa_rdm_pke) % EFA_RDM_PKE_ALIGNMENT == 0, "efa_rdm_pke alignment check");
-/* Checks if packet entry structure fits into two x86 cache lines */
+#if !ENABLE_DEBUG
+/* In optimized builds, packet entry structure is designed to fit into two x86 cache lines */
 static_assert(sizeof (struct efa_rdm_pke) == EFA_RDM_PKE_ALIGNMENT, "efa_rdm_pke size check");
+#endif
 #endif
 
 struct efa_rdm_ep;
@@ -228,6 +297,30 @@ void efa_rdm_pke_release_rx(struct efa_rdm_pke *pkt_entry);
 void efa_rdm_pke_release_rx_list(struct efa_rdm_pke *pkt_entry);
 
 void efa_rdm_pke_release(struct efa_rdm_pke *pkt_entry);
+
+#if ENABLE_DEBUG
+/**
+ * @brief Record debug info event in packet entry (inline for performance)
+ */
+static inline void efa_rdm_pke_record_debug_info(struct efa_rdm_pke *pkt_entry,
+                                                   uint16_t qpn, uint32_t qkey,
+                                                   uint8_t gen, uint8_t event_type)
+{
+	if (!pkt_entry->debug_info)
+		return;
+
+	uint8_t idx = pkt_entry->debug_info->counter % EFA_RDM_PKE_DEBUG_INFO_SIZE;
+	struct efa_rdm_pke_debug_info *info = &pkt_entry->debug_info->entries[idx];
+
+	info->qkey = qkey;
+	EFA_RDM_PKE_DEBUG_INFO_SET_QPN_GEN(info, qpn, gen);
+	info->counter = pkt_entry->debug_info->counter;
+	info->event = event_type;
+	pkt_entry->debug_info->counter++;
+}
+
+void efa_rdm_pke_print_debug_info(struct efa_rdm_pke *pkt_entry);
+#endif
 
 void efa_rdm_pke_append(struct efa_rdm_pke *dst,
 			struct efa_rdm_pke *src);
--- a/prov/efa/src/rdm/efa_rdm_ep.h
+++ b/prov/efa/src/rdm/efa_rdm_ep.h
@@ -127,6 +127,8 @@ struct efa_rdm_ep {
 	struct ofi_bufpool *peer_robuf_pool;
 
 #if ENABLE_DEBUG
+	/* buffer pool for packet debug info */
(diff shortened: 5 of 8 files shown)


prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -521,6 +521,41 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
...
+	/* Record completion event: time=1 for COMPLETION */
+	uint8_t event_time = 1;
+	efa_rdm_pke_record_debug_info(pkt_entry,
+	                               pkt_entry->gen,
+	                               event_time,
+	                               efa_ibv_cq_wc_read_qp_num(ibv_cq),
+	                               0);  /* qkey not available in completion */
+	
+	/* Check for duplicate completion by examining debug_info_vec */
+	if (pkt_entry->debug_info_idx > 1) {
```
sunkuamzn: Iterating over the list of previous completions for every single completion will get very expensive. We don't need to use this list to detect duplicates. We just need to print the contents of this list in these two error paths

https://github.com/ofiwg/libfabric/blob/fb38504a7d971571e1760ac614a1cbb61ff4bd40/prov/efa/src/rdm/efa_rdm_peer.c#L162-L166

https://github.com/ofiwg/libfabric/blob/fb38504a7d971571e1760ac614a1cbb61ff4bd40/prov/efa/src/rdm/efa_rdm_pke_cmd.c#L524-L529

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -899,6 +934,11 @@ int efa_rdm_cq_poll_ibv_cq(ssize_t cqe_to_process, struct efa_ibv_cq *ibv_cq)
 	efa_cq_start_poll(ibv_cq);
 
 	while (efa_cq_wc_available(ibv_cq)) {
+#if ENABLE_DEBUG
+		static uint64_t rdm_poll_seq = 0;
+		fprintf(stderr, "[RDMA-CORE] poll_seq=%lu wr_id=%lu\n",
```
sunkuamzn: Adding a log entry for every completion will overwhelm the logging systems for large applications

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -711,6 +716,23 @@ ssize_t efa_rdm_pke_recvv(struct efa_rdm_pke **pke_vec,
 		recv_wr = &ep->base_ep.efa_recv_wr_vec[i];
 		recv_wr->wr.wr_id = efa_rdm_pke_get_wr_id(pke_vec[i]);
 
+#if ENABLE_DEBUG
+		// At POST time, get qkey from QP
```
sunkuamzn: QKEY is also available in `efa_rdm_ep->base_ep->qp`

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +75,24 @@ enum efa_rdm_pke_alloc_type {
...
+ * @brief Debug info for tracking packet lifecycle events
+ * 
+ * Used to diagnose "Packet already processed" errors by recording
+ * key events: post, completion, duplicate completion.
+ */
+#define DEBUG_INFO_SIZE 3
+
+struct debug_info {
+	uint8_t gen;      /**< Generation counter at this event */
+	uint8_t time;     /**< Event type: 0=post, 1=completion, 2=duplicate */
```
sunkuamzn: The gen counter is tracking similar logic. If `wr_id && GEN_COUNTER_MASK == pke->gen`, then the pke is posted. If `(wr_id && GEN_COUNTER_MASK) + 1 == pke->gen` then we have received the completion for that pke.

I think timestamp would be more useful. @shijin-aws what do you think?

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -19,7 +19,7 @@
 #define EFA_RDM_PKE_IN_PEER_OUTSTANDING_TX_PKTS	BIT_ULL(7) /**< this packet entry is in peer->outstanding_tx_pkts list */
 #define EFA_RDM_PKE_IN_OPE_QUEUED_PKTS	BIT_ULL(8) /**< this packet entry is in ope->queued_pkts list */
 
-#define EFA_RDM_PKE_ALIGNMENT		128
+#define EFA_RDM_PKE_ALIGNMENT		256
```
sunkuamzn: We should keep this macro at 128 because it affects the alignment for release builds as well

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -79,6 +79,11 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 	pkt_entry->payload_mr = NULL;
 	pkt_entry->peer = NULL;
 
+#if ENABLE_DEBUG
+	memset(pkt_entry->debug_info_vec, 0, sizeof(pkt_entry->debug_info_vec));
+	pkt_entry->debug_info_idx = 0;
```
sunkuamzn: This would erase the previously recorded completions every time the packet is allocated

We have two options
(1) be OK with uninitialized values for `debug_info_idx` and `debug_info_vec` and interpret them later 

(2) add this logic to the `alloc_fn` of the packet pool so that its only executed once but that function operates on the entire buffer and not on individual pkes

https://github.com/ofiwg/libfabric/blob/fb38504a7d971571e1760ac614a1cbb61ff4bd40/prov/util/src/util_buf.c#L173-L177

https://github.com/ofiwg/libfabric/blob/fb38504a7d971571e1760ac614a1cbb61ff4bd40/prov/efa/src/rdm/efa_rdm_ep_fiops.c#L119

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -711,6 +716,23 @@ ssize_t efa_rdm_pke_recvv(struct efa_rdm_pke **pke_vec,
 		recv_wr = &ep->base_ep.efa_recv_wr_vec[i];
 		recv_wr->wr.wr_id = efa_rdm_pke_get_wr_id(pke_vec[i]);
 
+#if ENABLE_DEBUG
```
sunkuamzn, line 755: Need similar logic in the send path
https://github.com/ofiwg/libfabric/blob/fb38504a7d971571e1760ac614a1cbb61ff4bd40/prov/efa/src/rdm/efa_rdm_pke.c#L414-L415

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +75,24 @@ enum efa_rdm_pke_alloc_type {
...
  */
+
+#if ENABLE_DEBUG
+/**
+ * @brief Debug info for tracking packet lifecycle events
+ * 
+ * Used to diagnose "Packet already processed" errors by recording
+ * key events: post, completion, duplicate completion.
+ */
+#define DEBUG_INFO_SIZE 3
```
sunkuamzn: Can you instead have a macro for the maximum size of the pke struct and update this macro to automatically calculate the size of the debug info list?

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -521,6 +521,41 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 	uint32_t imm_data = 0;
 	bool has_imm_data = false;
 
+#if ENABLE_DEBUG
+	/* Record completion event: time=1 for COMPLETION */
+	uint8_t event_time = 1;
+	efa_rdm_pke_record_debug_info(pkt_entry,
+	                               pkt_entry->gen,
```
shijin-aws, line 540: I thought we want to record such information for all completions, not only receive..

prov/efa/src/rdm/efa_rdm_cq.c
shijin-aws: I am also confused what is our goal here? Log the rdma-core poll seq for each wr_id? We already have a logging here https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/efa_data_path_ops.h#L499-L513 and you can tell from the log whether the same wr_id was polled twice or more, without a seq no.

prov/efa/src/rdm/efa_rdm_pke.c
shijin-aws: Be aware for the mem-poison build this debug_info_vec will be erased by garbage during the pkt release as well.

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -790,3 +812,50 @@ ssize_t efa_rdm_pke_user_recvv(struct efa_rdm_pke **pke_vec,
...
+ * @param pkt_entry Packet entry
+ */
+void efa_rdm_pke_print_debug_info(struct efa_rdm_pke *pkt_entry)
+{
+	int i;
+	const char *event_name;
+	
+	for (i = 0; i < DEBUG_INFO_SIZE && i < pkt_entry->debug_info_idx; i++) {
+		struct debug_info *info = &pkt_entry->debug_info_vec[i];
+		event_name = (info->time == 0) ? "POST" :
```
shijin-aws: 1. can we use an map (array) for the enums -> string mapping.
2. can we change the `time` into `event_name` for intuitive
3. I thought we will have another field for the timestamp?

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -790,3 +812,50 @@ ssize_t efa_rdm_pke_user_recvv(struct efa_rdm_pke **pke_vec,
...
+ * @brief Print debug info history for packet entry
+ * 
+ * @param pkt_entry Packet entry
+ */
+void efa_rdm_pke_print_debug_info(struct efa_rdm_pke *pkt_entry)
+{
+	int i;
+	const char *event_name;
+	
+	for (i = 0; i < DEBUG_INFO_SIZE && i < pkt_entry->debug_info_idx; i++) {
```
shijin-aws: u can do `i < MIN(DEBUG_INFO_SIZE, debug_info_idx)` ?

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -521,6 +521,41 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
...
+	                               event_time,
+	                               efa_ibv_cq_wc_read_qp_num(ibv_cq),
+	                               0);  /* qkey not available in completion */
```
shijin-aws: you can get it from the hdr

prov/efa/src/rdm/efa_rdm_pke.h
yinliaws: I think so.

prov/efa/src/rdm/efa_rdm_pke.c
shijin-aws: But reading the PR further, it seems it only record the debug info within the lifecycle of a single packet entry

prov/efa/src/rdm/efa_rdm_cq.c
yinliaws: Okay. Then I will just remove this one.

yinliaws:
> I'm not sure if recording the posting events is useful. Need to discuss.

Do we still want to record the posting events?

prov/efa/src/rdm/efa_rdm_cq.c
```diff
@@ -521,6 +521,17 @@ static void efa_rdm_cq_handle_recv_completion(struct efa_ibv_cq *ibv_cq, struct
 	uint32_t imm_data = 0;
 	bool has_imm_data = false;
 
+#if ENABLE_DEBUG
+	/* Record RECV_COMPLETION event */
+	if (ep->base_ep.qp) {
```
sunkuamzn: This `if` can be removed. The QP will always exist in this situation.

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -57,6 +57,7 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 
 #if ENABLE_DEBUG
 	dlist_init(&pkt_entry->dbg_entry);
+	pkt_entry->debug_info_counter = 0;
```
sunkuamzn: This resets the counter on every alloc

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -790,3 +813,56 @@ ssize_t efa_rdm_pke_user_recvv(struct efa_rdm_pke **pke_vec,
...
+/**
+ * @brief Record debug info event in packet entry
+ *
+ * @param pkt_entry Packet entry
+ * @param qpn Queue pair number
+ * @param qkey Queue key
+ * @param gen Generation counter
+ * @param event_type Event type (SEND_POST, SEND_COMPLETION, RECV_POST, RECV_COMPLETION)
+ */
+void efa_rdm_pke_record_debug_info(struct efa_rdm_pke *pkt_entry,
```
sunkuamzn: `inline`

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +79,46 @@ enum efa_rdm_pke_alloc_type {
...
+#if ENABLE_DEBUG
+/**
+ * @brief Debug info for tracking packet lifecycle events
+ *
+ * Bit-packed structure to minimize space while tracking all packet events.
+ * Used to diagnose "Packet already processed" errors.
+ *
+ * Total size: 7 bytes per entry, 100 entries = 700 bytes
+ */
+#define DEBUG_INFO_SIZE 100
```
sunkuamzn: Can we automate this so that we don't accidentally corrupt application data which is immediately after?
```
#define DEBUG_INFO_SIZE ((EFA_RDM_PKE_ALIGNMENT - offsetof(struct efa_rdm_pke, wiredata)) / sizeof(struct debug_info)
```

I'm also OK if we calculate `EFA_RDM_PKE_ALIGNMENT` from `DEBUG_INFO_SIZE`

alekswn: We do `% DEBUG_INFO_SIZE` a lot (including the hot path).
I think making it power of 2 (i.e. 128) would reduce perf impact.

yinliaws: 128 entries cannot be fit into a 1024-byte packet entry, for this concern, we can either increase the size to 2048 bytes, or use 64 entries.

prov/efa/src/rdm/efa_rdm_pke.h
alekswn: This comment will be wrong in debug mode.

yinliaws: I will update this comment as /* In optimized builds, packet entry structure is designed to fit into two x86 cache lines */

prov/efa/src/rdm/efa_rdm_pke.h
alekswn: Why do we want packet entry size to be multiple of 1024?
Also I think it's unlikely all ring buffer would end up in L1 cache if it spans over so many cache lines. So, probably it does not make sense to continue enforcing cache line alignment for packet entry in debug mode at all.

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +79,46 @@ enum efa_rdm_pke_alloc_type {
...
+/* Event type definitions */
+#define DEBUG_EVENT_SEND_POST       0  /**< Send packet posted */
+#define DEBUG_EVENT_SEND_COMPLETION 1  /**< Send packet completed */
+#define DEBUG_EVENT_RECV_POST       2  /**< Receive buffer posted */
+#define DEBUG_EVENT_RECV_COMPLETION 3  /**< Receive completed */
+
+struct debug_info {
+	uint32_t qkey;        /**< Queue key (4 bytes) */
+	uint16_t counter_qpn; /**< Bits [15:9]=counter_low (7 bits), [8:4]=qpn_high (5 bits), [3:0]=qpn_low (4 bits) */
+	uint8_t gen_event;    /**< Bits [7:2]=gen (6 bits), [1:0]=event_type (2 bits) */
```
sunkuamzn: The number of bits used by `gen` depends on `EFA_RDM_BUFPOOL_ALIGNMENT` https://github.com/ofiwg/libfabric/blob/f8262817c337d615a1acceea6cd4ecb526ce548b/prov/efa/src/efa.h#L118

I'm OK hardcoding the number of bits here but can you add an assert that `gen` doesn't use more bits? Otherwise, we can end up with data corruption

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +79,46 @@ enum efa_rdm_pke_alloc_type {
...
+ */
+#define DEBUG_INFO_SIZE 100
+
+/* Event type definitions */
+#define DEBUG_EVENT_SEND_POST       0  /**< Send packet posted */
+#define DEBUG_EVENT_SEND_COMPLETION 1  /**< Send packet completed */
+#define DEBUG_EVENT_RECV_POST       2  /**< Receive buffer posted */
+#define DEBUG_EVENT_RECV_COMPLETION 3  /**< Receive completed */
+
+struct debug_info {
```
sunkuamzn: Could we also add `efa_rdm_pke` prefix to `debug_info`?

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -790,3 +813,56 @@ ssize_t efa_rdm_pke_user_recvv(struct efa_rdm_pke **pke_vec,
...
+
+	for (i = 0; i < count; i++) {
+		int idx = (start_idx + i) % DEBUG_INFO_SIZE;
+		struct debug_info *info = &pkt_entry->debug_info_vec[idx];
+		uint8_t event = DEBUG_INFO_GET_EVENT(info);
+		EFA_WARN(FI_LOG_EP_DATA,
+		         "    [%d] counter=%u gen=%u qpn=%u qkey=%u (%s)\n",
+		         i, DEBUG_INFO_GET_COUNTER(info), DEBUG_INFO_GET_GEN(info),
+		         DEBUG_INFO_GET_QPN(info), info->qkey,
+		         event < 4 ? event_type_str[event] : "UNKNOWN");
```
shijin-aws: such const char list should be defined in the same header of `efa_rdm_pke.h`, also, let's define the size as a macro too (or derive from sizeof) and not use the magic number here

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -19,7 +19,11 @@
 #define EFA_RDM_PKE_IN_PEER_OUTSTANDING_TX_PKTS	BIT_ULL(7) /**< this packet entry is in peer->outstanding_tx_pkts list */
 #define EFA_RDM_PKE_IN_OPE_QUEUED_PKTS	BIT_ULL(8) /**< this packet entry is in ope->queued_pkts list */
 
+#if ENABLE_DEBUG
+#define EFA_RDM_PKE_ALIGNMENT		1024
```
sunkuamzn: Please add a comment explaining why this number is larger for debug build and how the empty space is used

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +79,46 @@ enum efa_rdm_pke_alloc_type {
...
+
+/* Event type definitions */
+#define DEBUG_EVENT_SEND_POST       0  /**< Send packet posted */
+#define DEBUG_EVENT_SEND_COMPLETION 1  /**< Send packet completed */
+#define DEBUG_EVENT_RECV_POST       2  /**< Receive buffer posted */
+#define DEBUG_EVENT_RECV_COMPLETION 3  /**< Receive completed */
+
+struct debug_info {
+	uint32_t qkey;        /**< Queue key (4 bytes) */
+	uint16_t counter_qpn; /**< Bits [15:9]=counter_low (7 bits), [8:4]=qpn_high (5 bits), [3:0]=qpn_low (4 bits) */
```
shijin-aws: is the counter here corresponding to the `debug_info_counter` in pkt_entry?

prov/efa/src/rdm/efa_rdm_pke.h
shijin-aws: This seems messy, can we define it like 

```
struct debug_info {
    uint32_t qkey;      // 4 bytes
    uint32_t counter : 7;
    uint32_t qpn     : 9; 
    uint32_t gen     : 6;
    uint32_t event_type : 2; 
};

```
and then we do not have to use those bit masks to access each of them

shijin-aws: Also I think its better to keep such debug_info 8 bytes for better alignment, that `counter` looks something we can use a larger size, 7 bits seems too small

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -19,7 +19,11 @@
 #define EFA_RDM_PKE_IN_PEER_OUTSTANDING_TX_PKTS	BIT_ULL(7) /**< this packet entry is in peer->outstanding_tx_pkts list */
 #define EFA_RDM_PKE_IN_OPE_QUEUED_PKTS	BIT_ULL(8) /**< this packet entry is in ope->queued_pkts list */
 
+#if ENABLE_DEBUG
+#define EFA_RDM_PKE_ALIGNMENT		512
+#else
```
darrylabbate: Why does the alignment requirement change when adding the debug info members to the PKE object? The existing alignment requirement works regardless of how large the added members are.

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +79,74 @@ enum efa_rdm_pke_alloc_type {
...
+ * packet lifecycle events in a circular buffer.
+ *
+ * Size: 8 bytes per entry (naturally aligned)
+ * Array size: Calculated to fit within 1024-byte debug alignment
+ * Debug build uses 1024-byte alignment to accommodate debug fields while
+ * production build uses 128-byte alignment for cache efficiency.
+ */
+
+/* Calculate maximum debug entries that fit in available space */
+#define DEBUG_INFO_SIZE 48  /* (512 - 120 base - 1 counter) / 8 bytes per entry */
```
darrylabbate: Why are we targeting a max object size of 512?

darrylabbate: Also echoing Sai's comment that we should namespace everything here with some kind of prefix for `DEBUG_INFO_*`, `DEBUG_EVENT_*`, etc

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -790,3 +799,34 @@ ssize_t efa_rdm_pke_user_recvv(struct efa_rdm_pke **pke_vec,
...
+               "DEBUG_GEN_BITS insufficient to hold EFA_RDM_PACKET_GEN_MASK");
+
+/**
+ * @brief Print debug info history for packet entry
+ *
+ * @param pkt_entry Packet entry
+ */
+void efa_rdm_pke_print_debug_info(struct efa_rdm_pke *pkt_entry)
+{
+	static const char *event_str[] = DEBUG_EVENT_TYPE_STRINGS;
```
darrylabbate: I would consider simply defining the string constants here instead of using the macro if they're only used here.

prov/efa/src/rdm/efa_rdm_pke.h
yinliaws: We don't need to change alignment,  the struct can remain 128-byte aligned. The debug fields (48 entries × 8 bytes = 384 bytes + 1 counter byte = 385 bytes) fit within the existing struct padding/space.

prov/efa/src/rdm/efa_rdm_pke.h
yinliaws: Because we want to have enough entries meanwhile not growing the packet entry too much.

yinliaws: Revised.

prov/efa/src/rdm/efa_rdm_pke.c
yinliaws: Revised.

prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -546,6 +546,31 @@ void efa_rdm_pke_handle_send_completion(struct efa_rdm_pke *pkt_entry)
 	struct efa_rdm_ep *ep;
 
 	ep = pkt_entry->ep;
+
+#if ENABLE_DEBUG
```
shijin-aws: rdma write completion will not hit this function, it is in 

https://github.com/ofiwg/libfabric/blob/f8262817c337d615a1acceea6cd4ecb526ce548b/prov/efa/src/rdm/efa_rdm_pke_nonreq.c#L575

shijin-aws: Actually, u can use this aggregated function to record all tx completions 

https://github.com/ofiwg/libfabric/blob/f8262817c337d615a1acceea6cd4ecb526ce548b/prov/efa/src/rdm/efa_rdm_ep_utils.c#L412

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +75,62 @@ enum efa_rdm_pke_alloc_type {
...
+#define EFA_RDM_PKE_DEBUG_EVENT_RECV_COMPLETION 3  /**< Receive completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_READ_POST       4  /**< RDMA read posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_READ_COMPLETION 5  /**< RDMA read completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_WRITE_POST      6  /**< RDMA write posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_WRITE_COMPLETION 7 /**< RDMA write completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_TYPE_COUNT      8  /**< Number of event types */
+
+struct efa_rdm_pke_debug_info {
+	uint32_t qkey;        /**< Queue key */
+	uint16_t qpn_gen;     /**< Bits [15:6]=qpn (9 bits), [5:0]=gen (6 bits) */
```
shijin-aws: Friendly bump this comment https://github.com/ofiwg/libfabric/pull/11816#discussion_r2714601806, though it is not a blocker for me

shijin-aws: I realize we need bit shift anyway so we can keep the current implementation. Though qpn should take at least 10 bits as we can have qpn > 512 today

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +75,62 @@ enum efa_rdm_pke_alloc_type {
...
+/* Event type definitions (3 bits, 8 types) */
+#define EFA_RDM_PKE_DEBUG_EVENT_SEND_POST       0  /**< Send packet posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_SEND_COMPLETION 1  /**< Send packet completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_RECV_POST       2  /**< Receive buffer posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_RECV_COMPLETION 3  /**< Receive completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_READ_POST       4  /**< RDMA read posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_READ_COMPLETION 5  /**< RDMA read completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_WRITE_POST      6  /**< RDMA write posted */
+#define EFA_RDM_PKE_DEBUG_EVENT_WRITE_COMPLETION 7 /**< RDMA write completed */
+#define EFA_RDM_PKE_DEBUG_EVENT_TYPE_COUNT      8  /**< Number of event types */
```
shijin-aws: There is one more completion type: the recv_rdma_with_imm: https://github.com/ofiwg/libfabric/blob/f8262817c337d615a1acceea6cd4ecb526ce548b/prov/efa/src/rdm/efa_rdm_cq.c#L234. We should record it here as well

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +75,62 @@ enum efa_rdm_pke_alloc_type {
...
+ * packet lifecycle events in a circular buffer.
+ *
+ * Size: 8 bytes per entry (naturally aligned)
+ * Target size: 512 bytes total (balance between debug history and memory overhead)
+ * - 48 entries provide sufficient history for diagnosis
+ * - Keeps struct at 4x cache lines (512 bytes) vs 8x (1024 bytes)
+ */
+
+/* Calculate maximum debug entries that fit in available space */
+#define EFA_RDM_PKE_DEBUG_INFO_SIZE 48  /* (512 - 120 base - 1 counter) / 8 bytes per entry */
```
darrylabbate: I'll expand on my previous question: how did we settle on 512 bytes as the ideal PKE object size when this debug info buffer is included? Are we simply picking a size? Is it based off some other property from which we can then derive the `EFA_RDM_PKE_DEBUG_INFO_SIZE` calculation? At a minimum, this hardcoded calculation result is too fragile. I know there was a previous revision to this PR that hoisted the base object definition so it can be used to calculate the appropriate size of the final object, but I feel this all indicates this is the wrong overall approach.

Why does this debug data need to inlined in the PKE object at all? Since this is for debug mode, have we considered adding a single pointer to the PKE object; pointing to an allocation outside the PKE object for the buffer? Then it can be as large as it needs to be and we don't need to rely on the implementation-defined world of object alignment.

yinliaws: The pointer approach would require heap allocation since PKE objects are pool-allocated and persist across multiple operations - stack allocation isn't viable given the PKE lifecycle.

However, since we're debugging potential memory corruption issues (like "Packet already processed"), heap allocation introduces concerns:

1. The debug buffer itself could be corrupted by the same bug we're trying to diagnose
2. Separate allocation changes memory layout, potentially masking or altering corruption patterns
3. Heap metadata corruption could add noise to the diagnosis

The 512-byte alignment is admittedly somewhat arbitrary, but it provides enough space for 48 entries which should capture sufficient history for diagnosis.

prov/efa/src/rdm/efa_rdm_pke_cmd.c
```diff
@@ -546,6 +546,34 @@ void efa_rdm_pke_handle_send_completion(struct efa_rdm_pke *pkt_entry)
...
+	/*
```
shijin-aws: It seems this is an old change... we already moved it to `..record_tx_completed` ?

shijin-aws (approved):
Please rebase to resolve merge conflicts

darrylabbate (dismissed):

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -373,6 +373,26 @@ void efa_rdm_ep_record_tx_op_submitted(struct efa_rdm_ep *ep, struct efa_rdm_pke
 	ope->efa_outstanding_tx_ops++;
 #if ENABLE_DEBUG
 	ep->efa_total_posted_tx_ops++;
+	/* Record post event based on operation type */
+	{
```
sunkuamzn: why have this brace?

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -413,6 +433,33 @@ void efa_rdm_ep_record_tx_op_completed(struct efa_rdm_ep *ep, struct efa_rdm_pke
 {
 	struct efa_rdm_ope *ope = NULL;
 
+#if ENABLE_DEBUG
+	/*
+	 * Record completion event based on operation type.
+	 * Note: qp_num is truncated to 10 bits, gen to 6 bits.
+	 */
+	{
```
sunkuamzn: why have this brace?

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -75,6 +75,72 @@ enum efa_rdm_pke_alloc_type {
...
```
sunkuamzn: This part no longer makes sense `/* (512 - 120 base - 1 counter) / 8 bytes per entry */`

prov/efa/src/rdm/efa_rdm_ep_utils.c
```diff
@@ -413,6 +433,33 @@ void efa_rdm_ep_record_tx_op_completed(struct efa_rdm_ep *ep, struct efa_rdm_pke
...
+		int event_type = EFA_RDM_PKE_DEBUG_EVENT_SEND_COMPLETION;
```
sunkuamzn: ~~`efa_rdm_ep_record_tx_op_completed` does not get called for send completions..~~

~~I think recording the debug info for completions should all be `efa_rdm_cq.c` to make it easy~~

ignore

sunkuamzn (commented):
Just realized that we are not covering the zero copy path but that's fine for now

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -57,6 +57,7 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
 	pkt_entry->debug_info = debug_info;
 #endif
 #endif
+
```
jiaxiyan, line 69: Did you push this commit by mistake?

yinliaws: No. This is intentional. The nested #endif structure:
- Outer block: #ifdef ENABLE_EFA_POISONING
- Inner block: #if ENABLE_DEBUG

jiaxiyan: I am referring to your second commit `Merge branch 'main' into debug-info-vec-instrumentation`

shijin-aws: Please do not make merge commit as separate one, every upstreamed commit should be self-contained and deployable

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -45,11 +45,52 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
...
+#if ENABLE_DEBUG
+	pkt_entry->debug_info = debug_info;
+#endif
+#endif
+	/* Without poisoning, debug_info pointer is naturally preserved in memory
+	 * (or NULL on first allocation from zeroed region). */
+
+#if ENABLE_DEBUG
+#ifdef __SANITIZE_ADDRESS__
+	/* ASAN poisons freed memory, so we can't preserve debug_info pointer.
```
shijin-aws: ASAN poisons freed memory, but the pkt entry release doesn't really free the memory. So if we have a pke alloc after a pke release, there shouldn't be poison from ASAN, right?

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -45,11 +45,52 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
...
 	pkt_entry->gen &= EFA_RDM_PACKET_GEN_MASK;
 	dlist_init(&pkt_entry->entry);
 
 #if ENABLE_DEBUG
 	dlist_init(&pkt_entry->dbg_entry);
+	/* Allocate debug info if not already allocated */
+	if (!pkt_entry->debug_info) {
+		pkt_entry->debug_info = ofi_buf_alloc(ep->pke_debug_info_pool);
+		if (!pkt_entry->debug_info) {
+			efa_rdm_pke_release(pkt_entry);
```
shijin-aws, line 84: Need a warning here, your debug info pool is already unlimited, if it alloc failed, that means we have trouble in the heap...

prov/efa/src/rdm/efa_rdm_pke.h
```diff
@@ -205,9 +272,11 @@ struct efa_rdm_pke {
 
 #if defined(static_assert)
 static_assert(sizeof (struct efa_rdm_pke) % EFA_RDM_PKE_ALIGNMENT == 0, "efa_rdm_pke alignment check");
-/* Checks if packet entry structure fits into two x86 cache lines */
+#if !ENABLE_DEBUG
+/* In optimized builds, packet entry structure is designed to fit into two x86 cache lines */
```
shijin-aws, line 276: I thought debug build is already > 128 bytes, we added that alignas so it will be padded to 256 bytes. Your add-on shouldn't change this story, do I miss anything here?

yinliaws: The size assertion is unchanged. Debug builds are already > 128 bytes (Added the debug_info pointer after the gen field), The assertion only checks sizeof == 128 for non-debug builds (#if !ENABLE_DEBUG).

prov/efa/src/rdm/efa_rdm_pke.c
yinliaws: Added EFA_WARN to alert when allocation fails from the unlimited pool, as this indicates heap exhaustion.

prov/efa/src/rdm/efa_rdm_pke.c
yinliaws: I saw crashes in ASAN builds, but now I realize the root cause was ofi_buf_free() being called incorrectly instead of the proper
efa_rdm_pke_release() function (which I fixed at line 87). Removed the unnecessary ASAN handling code.

prov/efa/src/rdm/efa_rdm_pke.h
shijin-aws: so before your change, the debug build is 128 bytes ? Not sure why we don't need such `!ENABLE_DEBUG` assert before your change

prov/efa/src/rdm/efa_rdm_pke.c
shijin-aws: I am a little bit not sure whether warning and returning NULL is the best behavior here - AFAICT debug_info is never released until the buffer pool is destroyed. So here we return a NULL will finally cause a EAGAIN error in upper layer and the application will try to poll cq and retry, but these retry will not work because no debug info is really released

shijin-aws: @sunkuamzn thoughts?

yinliaws: What about add a flag: if failed use the flag to indicate debug info is unavailable to avoid retry?

yinliaws: Using the existing flags

shijin-aws: I think in this case, writing an EQ error would be the best thing libfabric can do

yinliaws: Revised accordingly.

shijin-aws: Oh, I think that warning can still be kept.

prov/efa/src/rdm/efa_rdm_pke.h
shijin-aws: Bump this question

yinliaws: Before this change, the struct was 128 bytes in both debug and non-debug builds. This PR adds the debug_info pointer in debug builds, making them > 128 bytes. So it now only checks non-debug builds where the struct remains 128 bytes.

prov/efa/src/rdm/efa_rdm_pke.c
yinliaws: Added the warning back.

prov/efa/src/rdm/efa_rdm_pke.h
shijin-aws: ok, I can see it now

shijin-aws (dismissed):

prov/efa/src/rdm/efa_rdm_pke.c
sunkuamzn: Yeah this is an issue with having multiple pools. Writing to EQ or returning ENOMEM sounds good to me.

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -45,11 +45,50 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
...
+		debug_info = NULL;
+	}
+#endif
+	efa_rdm_poison_mem_region(pkt_entry, pkt_pool->attr.size);
+	pkt_entry->gen = gen;
+#if ENABLE_DEBUG
+	pkt_entry->debug_info = debug_info;
+#endif
+#endif
+	/* Without poisoning, debug_info pointer is naturally preserved in memory
```
sunkuamzn: These comments seem out of date

prov/efa/src/rdm/efa_rdm_pke.c
```diff
@@ -93,12 +132,19 @@ struct efa_rdm_pke *efa_rdm_pke_alloc(struct efa_rdm_ep *ep,
...
+	struct efa_rdm_pke_debug_info_buffer *debug_info = pkt_entry->debug_info;
+#endif
 	efa_rdm_poison_mem_region(pkt_entry, ofi_buf_pool(pkt_entry)->attr.size);
 	pkt_entry->gen = gen;
+#if ENABLE_DEBUG
+	pkt_entry->debug_info = debug_info;
+#endif
 #endif
+	/* Without poisoning, debug_info pointer is naturally preserved in memory.
+	 * ASAN builds will corrupt it, but we handle that in alloc by forcing NULL. */
```
sunkuamzn: These comments seem out of date

yinliaws: Thanks for catching these. I updated the comments.

shijin-aws (approved):

sunkuamzn:
bot:aws:retest

PR #11817, opened 2026-01-20
[v2.4.x]  prov/efa, fabtests: Fix double-free pkt-entry at efa_rdm_pke_proc_matched_longread_rtm

alekswn:
Backport PR #11808


a-szegel:
@j-xiong restarted the failed run, and it should finish in around 20m.

PR #11822, opened 2026-01-21
man: Clarify failed inject calls generate error cqs

a-szegel:


man/fi_cq.3.md
```diff
@@ -432,10 +432,10 @@ for debugging purposes.  See field details below for more information
 on the use of err_data and err_data_size.
 
 Note that error completions are generated for all operations, including
-those for which a completion was not requested (e.g. an endpoint
-is configured with FI_SELECTIVE_COMPLETION, but the request did not have
-the FI_COMPLETION flag set).  In such cases, providers will return as
-much information as made available by the underlying software and
+those for which a completion was not requested (e.g. inject transfers or
```
j-xiong, line 435: Add `when` after `or`?

man/fi_msg.3.md
```diff
@@ -163,7 +163,8 @@ struct fi_msg {
 The send inject call is an optimized version of fi_send with the
 following characteristics.  The data buffer is available for reuse
 immediately on return from the call, and no CQ entry will be written
-if the transfer completes successfully.
+if the transfer completes successfully. If the transfer fails, an
```
sunkuamzn, line 166: Are all providers implementing this now?

a-szegel: The man pages already specified this behavior, it just wasn't as explicit as I thought it should be.  I can double check that.

j-xiong:
@a-szegel Missed my comment about the commit message title line.

a-szegel:
@j-xiong fixed!

man/fi_msg.3.md
shijin-aws: If we do not force all providers to do this, we can change the `will` to `may` here?

j-xiong: The providers are supposed to do so. If not, the provider should be fixed.

a-szegel:
bot:aws:retest ... although doc-only change...

j-xiong:
Appveyor skips changes under `man/`. Unfortunately, it is reported as failed check instead of success.

PR #11826, opened 2026-01-21
prov/efa: Poison pke only on release and restore gen

sunkuamzn:
Poisoning on both alloc and release doesn't make sense. And the poisoning during release was not restoring the gen.


shijin-aws:
I think the value of having poison at alloc is to catch bug that uses uninitialized memory. If we only have poison at release, the 1st alloc will not be poisoned.

sunkuamzn:
> If we only have poison at release, the 1st alloc will not be poisoned.

I can add poisoning to the bufpool `alloc_fn` which currently only registers memory

shijin-aws:
> > If we only have poison at release, the 1st alloc will not be poisoned.
> 
> I can add poisoning to the bufpool `alloc_fn` which currently only registers memory

That looks good to me

sunkuamzn:
bot:aws:retest

PR #11827, opened 2026-01-22
prov/efa: Fix the error handling path in ah creation and qp enable

shijin-aws:
A series of commits that fix the error handling path in ah creation and qp enable and add unit tests


shijin-aws:
bot:aws:retest

PR #11831, opened 2026-01-22
prov/efa: Change address recovered log entry to info

sunkuamzn:
If a packet arrives soon after AH creation, the EFA device can return -1 for AHN. So we shouldn't print a warning.


shijin-aws (approved):

sunkuamzn:
bot:aws:retest

PR #11832, opened 2026-01-22
prov/efa: Separate the efa and shm mr registration

shijin-aws:
This PR includes a series of commits to address the issues that the efa and shm mr are always coupled today.

For explicit MR registration, shm mr should be registered.
For implicit (efa internal) MR registration, shm mr shouldn't be invoked.

Also fixes a couple of incorrect access code used for shm mr which is historical leftover before efa started uses shm as a peer provider.

After the fix, the internal efa memory pool will only be registered to efa device.


prov/efa/src/efa_mr.c
```diff
@@ -873,15 +866,60 @@
...
+	efa_mr->mr_fid.fid.context = attr.context;
+	efa_mr->mr_fid.fid.ops = &efa_mr_ops;
+
+	ret = efa_mr_reg_impl(efa_mr, flags, &attr);
+	if (ret) {
+		EFA_WARN(FI_LOG_MR, "Unable to register MR: %s\n",
+			fi_strerror(-ret));
+		free(efa_mr);
+	}
+	*mr = &efa_mr->mr_fid;
```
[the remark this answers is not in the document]
shijin-aws: Fixed

prov/efa/src/efa_mr.c
```diff
@@ -373,14 +373,7 @@ int efa_mr_cache_regv(struct fid_domain *domain_fid, const struct iovec *iov,
...
 
-	attr.mr_iov = iov;
-	attr.iov_count = count;
-	attr.access = access;
-	attr.offset = offset;
-	attr.requested_key = requested_key;
-	attr.context = context;
-	attr.iface = FI_HMEM_SYSTEM;
-	attr.hmem_data = NULL;
+	EFA_MR_ATTR_INIT_SYSTEM(attr, iov, count, access, offset, requested_key, context);
```
darrylabbate: The macro only really offers a convenience for setting the HMEM iface. Why not handle this in the declaration?

```c
struct fi_mr_attr attr = {
  .mr_iov = iov,
  .iov_count = count,
  .access = access,
  .offset = offset,
  .requested_key = requested_key,
  .context = context,
  .iface = FI_HMEM_SYSTEM,
  // Everything else is zero-initialized
};
```

shijin-aws: I do not have a strong opinion here. I just want to save the number of lines

shijin-aws: I made that assignment at the declaration, while keeping the macro to reduce code duplication, lemme know if the update looks good to u

prov/efa/src/rdm/efa_rdm_ep_fiops.c
```diff
@@ -34,16 +34,12 @@ int efa_rdm_pke_pool_mr_reg_handler(struct ofi_bufpool_region *region)
...
-	/* The fi_mr_reg call here bypasses the MR cache
-	 * While this memory is internal to the EFA provider and can be put in
-	 * the MR cache, there is no need to do that. The packet entry pools
-	 * that need MR are allocated once and the MR is stored in the bufpool
-	 * region's context. We don't register the same memory region twice. So
-	 * it does not help to put these MRs in the cache.
-	 */
-	ret = fi_mr_reg(&domain->util_domain.domain_fid, region->alloc_region,
-			region->pool->alloc_size, FI_SEND | FI_RECV, 0, 0, 0,
+	ret = domain->internal_buf_mr_regv(&domain->util_domain.domain_fid, &iov, 1, FI_SEND | FI_RECV, 0, 0, 0,
```
shijin-aws, line 42: I believe this change (it will make the buffer pool reg use mr cache when it's available) should be ok, but @sunkuamzn let me know if I miss any corner case here

shijin-aws:
bot:aws:retest

prov/efa/src/rdm/efa_rdm_ep_fiops.c
sunkuamzn: I think your change is good

I wonder if calling these MRs `implicit` is better than `internal`. When I first saw `internal`, I thought this function was for the bounce buffers

PR #11836, opened 2026-01-23
prov/efa/test: Initialize extra_info in handshake packet construction

sunkuamzn:
With memory poisoning enabled, the uninitialized extra_info field in handshake packets could contain garbage values. If the EFA_RDM_EXTRA_FEATURE_REQUEST_USER_RECV_QP bit happened to be set, efa_rdm_pke_handle_handshake_recv would call
efa_rdm_pke_get_handshake_opt_user_recv_qp_ptr, which asserts that EFA_RDM_HANDSHAKE_USER_RECV_QP_HDR flag is set in the header - causing test failures.

Fix by initializing extra_info to zero in efa_unit_test_handshake_pkt_construct.

### DIFF
--- a/prov/efa/test/efa_unit_test_cq.c
+++ b/prov/efa/test/efa_unit_test_cq.c
@@ -329,7 +329,7 @@ void test_rdm_cq_handshake_bad_send_status_impl(struct efa_resource **state, int
 	g_efa_unit_test_mocks.efa_ibv_cq_wc_read_qp_num = &efa_mock_efa_ibv_cq_wc_read_qp_num_return_mock;
 	g_efa_unit_test_mocks.efa_ibv_cq_wc_read_vendor_err = &efa_mock_efa_ibv_cq_wc_read_vendor_err_return_mock;
 	g_efa_unit_test_mocks.efa_ibv_cq_start_poll = &efa_mock_efa_ibv_cq_start_poll_return_mock;
-	ibv_cq->ibv_cq_ex->wr_id = (uintptr_t)pkt_entry;
+	ibv_cq->ibv_cq_ex->wr_id = (uint64_t)pkt_entry | (uint64_t)pkt_entry->gen;
 
 	/* Mock cq to simulate the send comp error */
 	expect_function_call(efa_mock_efa_ibv_cq_end_poll_check_mock);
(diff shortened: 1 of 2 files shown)


prov/efa/test/efa_unit_test_common.c
```diff
@@ -370,6 +370,7 @@ void efa_unit_test_handshake_pkt_construct(struct efa_rdm_pke *pkt_entry, struct
 	handshake_hdr->version = EFA_RDM_PROTOCOL_VERSION;
 	handshake_hdr->nextra_p3 = nex + 3;
 	handshake_hdr->flags = 0;
+	memset(handshake_hdr->extra_info, 0, nex * sizeof(uint64_t));
```
shijin-aws, line 374: should we use `efa_rdm_pke_init_handshake` to construct it? It initialize this extra info correctly as

```
memcpy(handshake_hdr->extra_info, pkt_entry->ep->extra_info, nex * sizeof(uint64_t));
```

sunkuamzn: We could use it but we need to create a fake peer

shijin-aws: yeah, I think we should try to avoid replacing the to-be-tested code by the customized test code. Faking a peer is ok in this case, but constructing the hdr again in the test looks odd to me

sunkuamzn: Actually, there are more issues. `efa_rdm_pke_init_handshake` is meant to initialize the handshake packet when sending the packet. `efa_unit_test_handshake_pkt_construct` is constructing the handshake packet to mock the receive of such a packet. So it's not the same use case.

`efa_rdm_pke_init_handshake` will always set the peer to the provided value - which is good for sending. But for receiving, we want the peer to be determined by the CQ read path. And it could even be NULL.

Similarly for `connid`. `efa_rdm_pke_init_handshake` always sets `connid` but we have a test that simulates receiving a packet without `connid`.

So I think we need two separate functions. I can rename `efa_unit_test_handshake_pkt_construct` to something like `efa_unit_test_construct_handshake_pkt_receive` if you like

shijin-aws: >efa_rdm_pke_init_handshake will always set the peer to the provided value - which is good for sending. But for receiving, we want the peer to be determined by the CQ read path. And it could even be NULL.

It depends on what we are testing, you can mock cq read to get those peer information, but if it is not the goal of the unit-test, we can fake a peer anyway

shijin-aws: >So I think we need two separate functions. I can rename efa_unit_test_handshake_pkt_construct to something like efa_unit_test_construct_handshake_pkt_receive if you like

Yeah I am good with that, as long as we clearly define the usage of these two functions

shijin-aws (approved):

PR #11839, opened 2026-01-26
prov/rxd: fix atomic fetch/compare path

aingerson:
Fix multiple issues in the atomic path:
- rx_entry was incorrectly added to both the tx_list and rx_list, corrupting the list. This was also happening on the rma read case and this patch fixes that as well by moving the dlist insert to each op individual op case
- The transmit completion for an atomic that is waiting for fetched data was completing too early since it was satisfying the condition that the bytes_done (sent data) is equal to the length of the tranfer. While all the data has been sent, we can't complete until all the fetched data has been received. This is fixed by adding a separate field to track the number of bytes received as well as sent and checking that the full message has been processed

### DIFF
--- a/prov/rxd/src/rxd_cq.c
+++ b/prov/rxd/src/rxd_cq.c
@@ -292,7 +292,9 @@ void rxd_progress_tx_list(struct rxd_ep *ep, struct rxd_peer *peer)
 				break;
 		}
 
-		if (tx_entry->bytes_done == tx_entry->cq_entry.len) {
+		if (tx_entry->bytes_done == tx_entry->cq_entry.len &&
+		    tx_entry->op != RXD_ATOMIC_FETCH &&
+		    tx_entry->op != RXD_ATOMIC_COMPARE) {
 			if (ofi_before(tx_entry->start_seq + (tx_entry->num_segs - 1),
 			    head_seq)) {
 				if (tx_entry->op == RXD_DATA_READ) {
@@ -539,12 +541,15 @@ static struct rxd_x_entry *rxd_match_rx(struct rxd_ep *ep,
 
 		dup_entry->start_seq = base->seq_no;
 		dlist_init(&dup_entry->entry);
-		return dup_entry;
+		rx_entry = dup_entry;
+		goto init;
 	}
 
 out:
 	dlist_remove(&rx_entry->entry);
+init:
 	rx_entry->cq_entry.len = MIN(rx_entry->cq_entry.len, total_size);
+	dlist_insert_tail(&rx_entry->entry, &(rxd_peer(ep, base->peer)->rx_list));
 	return rx_entry;
 }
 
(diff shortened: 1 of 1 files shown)


ooststep (dismissed):

zachdworkin (dismissed):

prov/rxd/src/rxd_cq.c
```diff
@@ -535,16 +539,17 @@ static struct rxd_x_entry *rxd_match_rx(struct rxd_ep *ep,
...
 		dup_entry = rxd_progress_multi_recv(ep, rx_entry, total_size);
 		if (!dup_entry)
-			goto out;
+			return NULL;
 
 		dup_entry->start_seq = base->seq_no;
 		dlist_init(&dup_entry->entry);
-		return dup_entry;
+		rx_entry = dup_entry;
+		goto out;
```
j-xiong: This `goto` is redundant now.

prov/rxd/src/rxd_cq.c
```diff
@@ -292,7 +293,10 @@ void rxd_progress_tx_list(struct rxd_ep *ep, struct rxd_peer *peer)
 				break;
 		}
 
-		if (tx_entry->bytes_done == tx_entry->cq_entry.len) {
+		if (tx_entry->bytes_done == tx_entry->cq_entry.len &&
+		    ((tx_entry->op != RXD_ATOMIC_FETCH &&
+		    tx_entry->op != RXD_ATOMIC_COMPARE) ||
+		    tx_entry->fetch_bytes == tx_entry->cq_entry.len)) {
```
j-xiong: I am wondering how this condition works for atomics that both read and write. In that case, `bytes_done` would count for all the bytes written and read, `fetch_bytes` would count for all bytes read, so they are different. the `op` would be RXD_ATOMIC_FETCH and this condition can never be true. Am I missing something?

aingerson: bytes_done is only the bytes written and then fetch_bytes is only the bytes read so for an op like ATOMIC_FETCH (or ATOMIC_COMPARE) that does both a read and a write, the bytes_done will be the length it sent. So in this condition,  we wait until all data has been sent (bytes_done == len) and, if we also have data to read, that we've received all the fetched data too (fetch_bytes == len). The whole if statement is very confusing... essentially it's
if (writes done && (no_reads || reads_done))

j-xiong: But on the recv path, `bytes_done` and `fetch_bytes` are updated at the same time.

Image

aingerson: Oh I see what you're saying... I am not sure. I will take a look next week. Adding DO NOT MERGE in the meantime!

aingerson: You're right, the condition couldn't be hit which was helping it pass coincidentally but it wasn't the right way to do it. Updated for a better fix and explanation

PR #11840, opened 2026-01-26
prov/efa: Fix asserts for efa-proto

a-szegel:
Allow user to pass in null buff/null descriptor for efa-protocol for read/write without hitting an assert.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -1381,7 +1381,11 @@ int efa_rdm_ope_post_read(struct efa_rdm_ope *ope)
 
 	ep = ope->ep;
 
-	assert(ope->iov_count > 0 && ope->iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->iov_limit);
+	/*
+	 * Allow local iov count to be equal to 0 b/c bounce buffer's pre-registered buff/desc
+	 * will be passed to rdma-core
+	 */
+	assert((ope->iov_count == 0 && ope->bytes_read_total_len == 0) || ope->iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->iov_limit);
 	assert(ope->rma_iov_count > 0 && ope->rma_iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->rma_iov_limit);
 
 	if (ope->bytes_read_total_len == 0) {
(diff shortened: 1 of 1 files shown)


prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1381,7 +1381,11 @@ int efa_rdm_ope_post_read(struct efa_rdm_ope *ope)
 
 	ep = ope->ep;
 
-	assert(ope->iov_count > 0 && ope->iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->iov_limit);
+	/*
+	 * Allow local iov count to be equal to 0 b/c bounce buffer's pre-registered buff/desc
+	 * will be passed to rdma-core
+	 */
+	assert(ope->iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->iov_limit);
```
sunkuamzn:
```suggestion
	assert(ope->iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->iov_limit || (ope->iov_count == 0 && ope->bytes_read_total_len == 0));
```

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -1526,7 +1530,11 @@ int efa_rdm_ope_post_remote_write(struct efa_rdm_ope *ope)
...
```
sunkuamzn:
```suggestion
	assert(ope->iov_count <= efa_rdm_ep_domain(ep)->info->tx_attr->iov_limit || (ope->iov_count == 0 && ope->bytes_write_total_len == 0));
```

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: The second part of the assert is not needed because if iov_count == 0, the left side will always be true, making the right side never trigger.

I think you were going for:
```
#if ENABLE_DEBUG
if (ope->iov_count == 0)
     assert(ope->bytes_read_total_len == 0)
#endif
```

prov/efa/src/rdm/efa_rdm_ope.c
a-szegel: The second part of the assert is not needed because if iov_count == 0, the left side will always be true, making the right side never trigger.

I think you were going for:
```
#if ENABLE_DEBUG
if (ope->iov_count == 0)
     assert(ope->bytes_write_total_len == 0)
#endif
```

shijin-aws: if we use ofi_total_iov_len to calculate `bytes_write_total_len`, iov_count == 0 already means `bytes_write_total_len=0`... I do not think it is more useful to add that additional asserts though

prov/efa/src/rdm/efa_rdm_ope.c
sunkuamzn: Yeah, you're right. Or you can flip the order so that the `iov_count == 0` gets checked first

I wanted to keep the assert that if iov_count is 0, the total length is also 0

prov/efa/src/efa_rma.c
```diff
@@ -210,6 +210,7 @@ static inline ssize_t efa_rma_post_write(struct efa_base_ep *base_ep,
 		msg->context, msg->addr, flags, FI_RMA | FI_WRITE);
 
 	/* Prepare SGE list */
+	assert(msg->iov_count > 0);
```
shijin-aws: It still doesn't address my initial concern.... that assert is for checking something you are sure to happen inside the code, but here iov_count is something parsed by application. I think we either need to make efa-direct support 0 iov_count, or return error.

a-szegel: sorry, I didn't see where this initial concern was commented.  Will fix.

a-szegel:
@shijin-aws, I simplified this PR to only be for efa-proto, and I will post a follow up PR for efa-direct.

PR #11843, opened 2026-01-27
fabtests/efa/multi_ep_stress: Fix thread safety and simplify architecture

alekswn:
The multi_ep_stress test had multiple race conditions when running with many worker threads and endpoint recycling. This commit redesigns the test architecture to eliminate these races while simplifying the code.

Thread Safety Fixes:

- Add thread-safe random number generation using ft_random.h with _Thread_local storage. The previous rand() calls caused races when multiple threads accessed the shared global state.

- Replace socket-based control channel with MPSC message queue using condition variables. The socket approach had complex error handling and race conditions during endpoint updates. The new queue provides clean thread-safe communication between workers.

- Introduce context_pool for pre-registered memory buffers and fi_context2 structures. This eliminates repeated memory registration overhead and ensures proper cleanup without double-free issues.

Architecture Simplification:

- Unify sender_context and receiver_context into single worker_context structure. Both sides now use the same code paths, reducing duplication and making the test easier to maintain.

- Remove --sender-addr command line option. The test now uses the existing OOB (out-of-band) mechanism for address exchange, eliminating the need for manual address specification and making pytest tests simpler.

### DIFF
--- a/fabtests/prov/efa/src/multi_ep_stress.c
+++ b/fabtests/prov/efa/src/multi_ep_stress.c
@@ -87,8 +87,8 @@ struct rma_info {
 
 // Endpoint metadata
 struct ep_info {
-	uint16_t worker_id;
-	uint16_t peer_idx;
+	uint16_t worker_id; // destination - worker ID on sender side
+	uint16_t peer_idx;  // source - index of receiver worker in sender's peer_ids table
 	char ep_addr[MAX_EP_ADDR_LEN];
 	struct rma_info rma;
 };
@@ -102,7 +102,7 @@ struct ep_message {
 	struct ep_info info;
 };
 
-// Multy producer single consumer thread-safe queue
+// Multi-producer single consumer thread-safe queue
 struct ep_message_queue {
 	struct ep_message *messages;
 	pthread_mutex_t mutex;
@@ -424,8 +424,7 @@ static int wait_for_comp(struct fid_cq *cq, int num_completions)
 		ret = pthread_mutex_timedlock(&shared_cq_lock, &b);
 		if (ret == ETIMEDOUT) {
 			fprintf(stderr,
-				"%ds timeout expired "
-				"while waiting for shared CQ lock\n",
+				"%ds timeout expired while waiting for shared CQ lock\n",
 				timeout);
 			return 0;
 		}
@@ -588,6 +587,10 @@ static void *run_sender_worker(void *arg)
 			|| ops_posted_for_peer[peer_idx] == topts.msgs_per_sender) {
 			if (++peer_idx == ctx->num_peers)
 				peer_idx = 0;
+			// Relinquish current CPU core to break busy-wait loop
+			// on the current thread. Main thread will be scheduled
+			// before current thread, therefore it would likely push
+			// an update to worker's control queue if the one was pending.
 			sched_yield();
 			continue;
 		}
@@ -692,6 +695,8 @@ static void *run_sender_worker(void *arg)
 	printf("Sender %d: All cycles completed, total ops: %lu\n",
 	       ctx->worker_id, total_ops);
 
+	// Some updates might have arrived too late to be applied.
+	// Draining such messages from control queue until terminator.
 	do {
 		ret = ep_message_queue_pop(ctx->control_queue, &msg); // blocks
 		if (ret) {
@@ -865,7 +870,7 @@ static void *run_receiver_worker(void *arg)
 	}
 out:
 	printf("Receiver %d: Completed %d EP cycles\n", ctx->worker_id, cycle);
-	const struct ep_message terminator = {0};
+	const struct ep_message terminator = { .type = EP_MESSAGE_TYPE_TERMINATOR };
 	ep_message_queue_push(ctx->control_queue, &terminator);
 
 	if (ret) {
@@ -1014,6 +1019,9 @@ static int run_sender(void)
 	// Dispatch control messages from OOB channel to worker's control channels
 	struct ep_message msg;
 	while (true) {
+		// ft_sock_recv blocks until the complete message recived.
+		// It returns 0 on suceess and -FI_ENOTCONN if socket
+		// had been closed by peer.
 		ret = ft_sock_recv(oob_sock, (void*)&msg, sizeof(msg));
 		if (ret) {
 			FT_PRINTERR("ft_sock_recv", ret);
@@ -1029,6 +1037,9 @@ static int run_sender(void)
 			goto out;
 		}
 	}
+	// On the happy path we exit this loop because of terminator message
+	// recived on OOB channnel. That means all reciver's workers has
+	// completed all cycles and no one is listening anymore.
 
 	// Terminate workers
 	for (int i = 0; i < topts.num_sender_workers; i++) {
@@ -1169,7 +1180,7 @@ static int run_receiver(void)
 			}
 		}
 	}
-	struct ep_message terminator = {0};
+	struct ep_message terminator = { .type = EP_MESSAGE_TYPE_TERMINATOR };
 	ret = ft_sock_send(oob_sock, &terminator, sizeof(terminator));
 	if (ret) {
 		FT_PRINTERR("ft_sock_send", ret);
--- a/fabtests/prov/efa/src/multi_ep_stress.md
+++ b/fabtests/prov/efa/src/multi_ep_stress.md
@@ -82,16 +82,16 @@ Thread-safe random functions allow to re-run tests with the same random choises.
 
 For each sender worker thread:
 
-2. **Enter EP cycle loop**:
+1. **Enter EP cycle loop**:
    - Create new endpoint with CQ/AV
-   - Insert all cached peer addresses into AV
+   - Insert all previously cached peer addresses into AV
    - Random sleep (0-100ms) to simulate real workload
-3. **Message posting loop**:
-   - Check for pending AV updates from receivers
+2. **Message posting loop**:
+   - Check for pending AV updates from receivers, apply and cache the updates
    - Post send/write operation to next peer (round-robin)
    - Handle `-FI_EAGAIN` by draining completions
    - Track operations posted/completed per cycle and per peer
-4. **Cycle completion**:
+3. **Cycle completion**:
    - Randomly decide whether to wait for all completions or proceed immediately
    - Destroy endpoint and start next cycle
 


fabtests/include/ft_random.h
```diff
@@ -0,0 +1,125 @@
...
+
+static inline int32_t ft_random_get_int32_in_range(struct random_data *restrict random_data, int32_t min_int32, int32_t max_int32)
+{
+	/* Max range is INT32_MAX - 1*/
+	const int32_t modulo = max_int32 - min_int32 + 1;
+	const int32_t rejection_limit = INT32_MAX - (INT32_MAX % modulo);
+
+	int32_t result;
+	do {
+		result = ft_random_get_int32(random_data);
```
alekswn: must be `result = min_int32 + ft_random_get_int32(random_data);`

fabtests/include/ft_random.h
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) 2026 Amazon.com, Inc. or its affiliates. All rights reserved.
+ *
```
shijin-aws, line 3: If we want this to be in the common header folder, can you make it as a separate commit for community to review

fabtests/include/ft_random.h
```diff
@@ -0,0 +1,125 @@
...
+#endif /* HAVE_CONFIG_H */
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <time.h>
+
+#include "shared.h"
+
+#ifdef __GNUC__
+    #define likely(x)   __builtin_expect(!!(x), 1)
```
shijin-aws: Prefer `FT_UNLIKELY` and `FT_LIKELY` similar to the `OFI_LIKELY` naming pattern

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -250,849 +282,679 @@ static void cleanup_endpoint(struct common_context *ctx)
...
+	return FI_SUCCESS;
+error:
+	fprintf(stderr,
+		"Failed to allocate peer_ids for worker %d\n",
+		current_worker_id);
+	return -FI_ENOMEM;
 }
 
-static int wait_for_comp(struct fid_cq *cq, int num_completions, int64_t timeout_ns)
+static int wait_for_comp(struct fid_cq *cq, int num_completions)
```
shijin-aws, line 412: Now we have a signature change of the function, can you add docstring to explain the newest behavior

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -250,849 +282,679 @@ static void cleanup_endpoint(struct common_context *ctx)
...
-free_ai:
-	freeaddrinfo(ai);
+error:
+	cleanup_endpoint(ctx);
 	return ret;
 }
 
-static int calculate_worker_distribution(int sender_id, int *num_peers,
-					 int *peer_ids)
+static int calculate_worker_distribution(uint16_t current_worker_id,
```
shijin-aws, line 366: Since you are here, add documentation?

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -250,849 +282,679 @@ static void cleanup_endpoint(struct common_context *ctx)
...
-			peers_ready = true;
-			for (int i = 0; i < ctx->num_peers; i++) {
-				if (ctx->peer_addrs[i] == FI_ADDR_UNSPEC) {
-					peers_ready = false;
-					break;
+		// Apply all pending AV updates
+		while (true) {
+			struct ep_message msg;
+			ret = ep_message_queue_try_pop(ctx->control_queue, &msg);
+			if (ret == 0) {
```
shijin-aws, line 541: did we remove av before this PR, I could not find it

fabtests/configure.ac
```diff
@@ -308,6 +308,11 @@ dnl Check for 128-bit integer support
 AC_CHECK_TYPE([__int128],
 	[AC_DEFINE(HAVE___INT128, 1, [Set to 1 to use 128-bit ints])])
 
+dnl Check for random_r function
```
shijin-aws, line 311: similar comment, common code change should be in a separate commit

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -89,152 +76,197 @@ static struct option test_long_opts[] = {
...
+	pthread_cond_destroy(&q->cond);
+	free(q->messages);
+}
+
+static int ep_message_queue_push(struct ep_message_queue *q, const struct ep_message *src) {
+	int ret = 0;
+	ret = pthread_mutex_lock(&q->mutex);
+	if (ret)
 		return ret;
+	while (q->size == EP_MESSAGE_QUEUE_CAPACITY) {
```
shijin-aws, line 145: It's not clear to me how such message queue works.... The goal is for receiver to send its new EP address to sender which is on a different node... How does the queue realize it?? does the message in the queue finally get sent over the wire?

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: No we did not, should not we?

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: There is one queue on receiver side and multiple queues on sender side.
Receiver workers push their messages to the common queue. Receiver's main thread pushes these messages to fabtests's OOB socket.
Sender's main thread pulls messages from the OOB socket and routes them to individual worker queues.
Sender threads pull messages from their queues and apply AV updates.

fabtests/prov/efa/src/multi_ep_stress.md
```diff
@@ -0,0 +1,176 @@
...
+
+### 1. Endpoint Recycling
+
+The core stress mechanism involves repeatedly destroying and recreating endpoints during active communication:
+
+- **Sender EP Cycles**: Controlled by `--sender-ep-cycles` (default: 1)
+- **Receiver EP Cycles**: Controlled by `--receiver-ep-cycles` (default: 1)
+
+Within each cycle:
+1. Endpoint is created with fresh CQ/AV resources
```
sunkuamzn: Test has shared AV and shared CQ options

fabtests/prov/efa/src/multi_ep_stress.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Worker Model
+
+Both senders and receivers use a multi-threaded worker pool architecture:
+
+- Each worker manages its own endpoint lifecycle
+- Workers are assigned specific peer connections based on distribution logic
+- Workers operate independently with thread-safe coordination
+
+## Key Features
```
sunkuamzn, line 26: Random sleeps with a configurable seed and thread safe per thread random number generation to make reproductions easy is a key feature

fabtests/prov/efa/src/multi_ep_stress.md
```diff
@@ -0,0 +1,176 @@
...
+1. Parse command-line options and configure test parameters
+2. Initialize libfabric fabric/domain resources
+3. Establish OOB socket connection for control messages
+4. Create shared resources (if enabled)
+
+### Sender Workflow
+
+For each sender worker thread:
+
+1. **Wait for peer addresses**: Block until all receiver endpoint addresses arrive via control queue
```
sunkuamzn: Can we have each sender thread start as soon as it receives the receiver address? Why block for all addresses? Each sender may not even talk to all of the receivers

That can help catch issues with overlapping control and data plane calls

fabtests/prov/efa/src/multi_ep_stress.md
```diff
@@ -0,0 +1,176 @@
...
+1. **Enter EP cycle loop**:
+   - Create new endpoint with CQ/AV
+   - Allocate RMA buffer (if using writedata operation)
+   - Notify all connected senders of new endpoint address
+   - Random sleep (0-100ms)
+2. **Message posting loop**:
+   - Post receive operations (for msg operations)
+   - For RMA writedata, skip posting (writes are one-sided)
+3. **Cycle completion**:
+   - For RMA: always wait for completions
```
sunkuamzn: Why?

fabtests/prov/efa/src/multi_ep_stress.c
sunkuamzn: I think making it optional with a cmd line arg would be good

fabtests/prov/efa/src/multi_ep_stress.c
sunkuamzn: Then a sender thread can start as soon as it receives one address. Why wait for all addresses?

fabtests/include/ft_random.h
shijin-aws: Also, I think it should be in shared.h for general

fabtests/prov/efa/src/multi_ep_stress.md
alekswn: It was a special condition I had introduced for writedata operation. Actually now I can see it is not necessary. Removing in next revision.

fabtests/prov/efa/src/multi_ep_stress.md
alekswn: Will be done in coming revision

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: Will add that in the coming revision

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -86,155 +72,201 @@ static struct option test_long_opts[] = {
...
+	enum {
+		EP_MESSAGE_TYPE_TERMINATOR = 0,
+		EP_MESSAGE_TYPE_UPDATE,
+	} type;
 	struct ep_info info;
 };
 
-// Worker status tracking
-struct worker_status {
+// Multy producer single consumer thread-safe queue
```
charlesstoll: Nit: spelling "multy producer-> multi-producer". No need to re-spin the patch just for this though

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -250,849 +282,708 @@ static void cleanup_endpoint(struct common_context *ctx)
...
 
-	if (timeout_ns >= 0)
-		clock_gettime(CLOCK_MONOTONIC, &a);
+	clock_gettime(CLOCK_MONOTONIC, &a);
+	if (topts.shared_cq) {
+		memcpy(&b, &a, sizeof(b));
+		b.tv_sec += timeout;
+		ret = pthread_mutex_timedlock(&shared_cq_lock, &b);
+		if (ret == ETIMEDOUT) {
+			fprintf(stderr,
```
charlesstoll, line 426: Nit: shouldn't this be one line so that we can search for "timeout expired while waiting" in the code base?

alekswn:
bot:aws:retest

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -250,849 +282,708 @@ static void cleanup_endpoint(struct common_context *ctx)
...
-	return (void *) (intptr_t) ret;
-}
-
-// Common function for buffer and MR setup
-static int setup_worker_resources(void **buf, struct fi_context2 **ctx,
-				  struct fid_mr **mr, uint64_t access,
-				  int buffer_multiplier)
-{
-	int ret;
+	const struct ep_message terminator = {0};
```
shijin-aws: This assumes EP_MESSAGE_TYPE_TERMINATOR is 0, can we make it assigned explicitly so it won't be broken accidentally when someone change EP_MESSAGE_TYPE_TERMINATOR to 1

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -1101,159 +992,92 @@ static int run_sender(void)
...
-		printf("\n");
-
-		// Allocate control sockets array
-		workers[i].control_socks = calloc(num_peers, sizeof(int));
-		if (!workers[i].control_socks) {
-			ret = -FI_ENOMEM;
+	// Dispatch control messages from OOB channel to worker's control channels
+	struct ep_message msg;
+	while (true) {
+		ret = ft_sock_recv(oob_sock, (void*)&msg, sizeof(msg));
```
shijin-aws, line 1025: When ret is non zero that means the receiver closed its socket? need some comments here

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -1266,167 +1090,111 @@ static int run_receiver(void)
...
-					"worker %d->%d: %d\n",
-					i, sender_idx, ret);
+				FT_PRINTERR("ft_sock_send", ret);
 				goto out;
 			}
-			sock_idx++;
 		}
 	}
+	struct ep_message terminator = {0};
+	ret = ft_sock_send(oob_sock, &terminator, sizeof(terminator));
```
shijin-aws, line 1184: similar comments on terminator explicit assignment

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -250,849 +282,708 @@ static void cleanup_endpoint(struct common_context *ctx)
...
-						strerror(errno));
-					return -errno;
-				}
-			}
-			sent += ret;
-		}
-
+	for (size_t i = 0; i < ctx->num_peers; i++) {
+		msg.info.worker_id = ctx->peer_ids[i];
+		msg.info.peer_idx = ctx->worker_id / topts.num_sender_workers;
```
shijin-aws, line 734: IIUC, worker_id here is the sender's ctx id (worker_id), peer_idx is the index of this receiver worker in sender worker's peer_ids array. Right? Some more documentation is appreciated, otherwise the reader won't know what they refer ti

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -1101,159 +992,102 @@ static int run_sender(void)
...
-		listen_socks[i] = -1;
 	}
 
-	// Create worker threads
+	// Terminate workers
 	for (int i = 0; i < topts.num_sender_workers; i++) {
-		ret = pthread_create(&threads[i], NULL, run_sender_worker,
-				     &workers[i]);
+		assert(msg.type == EP_MESSAGE_TYPE_TERMINATOR);
+		ret = ep_message_queue_push(workers[i].control_queue, &msg);
```
shijin-aws, line 1047: This means sender got a terminating signal from the receiver (the whole), right? Need a comment to explain what it means

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -612,6 +587,15 @@ static void *run_sender_worker(void *arg)
 			}
 		}
 
+		// Skip not ready yet and completed already peers
+		if (fi_addr[peer_idx] == FI_ADDR_UNSPEC
+			|| ops_posted_for_peer[peer_idx] == topts.msgs_per_sender) {
+			if (++peer_idx == ctx->num_peers)
```
shijin-aws, line 588: why not just use `peer_idx = (peer_idx +1) % ctx->num_peers`

fabtests/prov/efa/src/multi_ep_stress.md
```diff
@@ -82,16 +82,15 @@ Thread-safe random functions allow to re-run tests with the same random choises.
 
 For each sender worker thread:
 
-1. **Wait for peer addresses**: Block until all receiver endpoint addresses arrive via control queue
 2. **Enter EP cycle loop**:
    - Create new endpoint with CQ/AV
-   - Insert all peer addresses into AV
+   - Insert all cached peer addresses into AV
```
shijin-aws: what does cached mean here?

fabtests/prov/efa/src/multi_ep_stress.md
```diff
@@ -82,16 +82,15 @@ Thread-safe random functions allow to re-run tests with the same random choises.
 
 For each sender worker thread:
 
-1. **Wait for peer addresses**: Block until all receiver endpoint addresses arrive via control queue
 2. **Enter EP cycle loop**:
```
shijin-aws: since you remove #1, you need to change 2 to 1 here?

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -612,6 +587,15 @@ static void *run_sender_worker(void *arg)
...
+				peer_idx = 0;
+			sched_yield();
```
shijin-aws, line 594: what does sched_yield do here?

fabtests/prov/efa/src/multi_ep_stress.c
```diff
@@ -702,9 +707,17 @@ static void *run_sender_worker(void *arg)
 		}
 	}
 
-	printf("Sender %d: All cycles completed, total ops=%lu\n",
+	printf("Sender %d: All cycles completed, total ops: %lu\n",
 	       ctx->worker_id, total_ops);
 
+	do {
```
shijin-aws, line 700: what does this loop of pops do?

fabtests/prov/efa/src/multi_ep_stress.c
shijin-aws: I don't quite get the commit message. The change looks to me we are going to terminate all sender workers when we get a terminator message from the receiver via sockets, why it says

>There is a possability for sender workers to exit before applying all
updates if they skip waiting for completions.

fabtests/prov/efa/src/multi_ep_stress.c
shijin-aws, line 700: So the goal of the loop is to waiting for a terminator message from the main thread? Need a comment

alekswn: Yes, also there might be some late updates which need to be drained. That can happen if sender had been skipping completion waiting frequently, but receiver did wait for most completions.  

Added a comment

alekswn: See below

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: It's useful to allow main thread to make progress faster if we're running more workers than CPU cores.

Added a comment.

fabtests/prov/efa/src/multi_ep_stress.c
alekswn: Probably over-optimizing but `if` statement seems to be order of magnitude faster in term of CPU cycles.
I think we want the least overhead possible in the body of this loop to catch more race conditions through.

Modulo operation with arbitrary arguments  (`idiv` instruction in x86 assembly) takes 10 CPU cycles. 
On the other hand `if` statement translates to conditional move (1 cycle, no branching).

See: https://godbolt.org/z/cPz1qov1n

fabtests/prov/efa/src/multi_ep_stress.md
alekswn: I meant restoring AV vector addresses saved at `ep_addr` array. I think it's referenced as cache in code comments as well.
Added reference to "caching" AV updates to p.2

fabtests/prov/efa/src/multi_ep_stress.c
shijin-aws: >also there might be some late updates which need to be drained. That can happen if sender had been skipping completion waiting frequently, but receiver did wait for most completions.

Does sender care about these late updates (except for terminator) after it is done? It doesn't care, right?

alekswn: No, it does not care. 
 it's too late to apply these updates, so sender just skips them while waiting for terminator.

PR #11845, opened 2026-01-27
fabtests/efa: Fix test _rmd_tagged_pingpong_truncate_error timeout

alekswn:
- increase pytest timeout to 60 seconds
- fix typo in test name

### DIFF
--- a/fabtests/pytest/efa/test_rdm.py
+++ b/fabtests/pytest/efa/test_rdm.py
@@ -317,4 +317,4 @@ def test_rdm_tagged_pingpong_truncate_error(cmdline_args, completion_semantic):
     efa_run_client_server_test(cmdline_args, executable, "short",
                                completion_semantic, "host_to_host",
                                message_size=message_size, fabric="efa",
-                               timeout=15, might_fail=might_fail)
+                               timeout=60, might_fail=might_fail)


fabtests/pytest/efa/test_rdm.py
```diff
@@ -317,4 +317,4 @@ def test_rmd_tagged_pingpong_truncate_error(cmdline_args, completion_semantic):
     efa_run_client_server_test(cmdline_args, executable, "short",
                                completion_semantic, "host_to_host",
                                message_size=message_size, fabric="efa",
-                               timeout=15, might_fail=might_fail)
+                               timeout=60, might_fail=might_fail)
```
shijin-aws, line 320: what is the difference between this timeout and the `--timeout 1` ?

alekswn: ClientServerTest passes it's timeout value to `timeout` linux tool which wraps fabtest. The tool kills application and returns 124 if timeout expired before the application exited.  

`--timeout 1` passed to fabtest sets global variable `timeout` in fabtests. This variable is used in different places in the fabtest framework as a threshold to give up on a particular operation (for instance, waiting for completion: https://github.com/ofiwg/libfabric/blob/69e40aff93b2ebb6d7250290c60155479f09ecae/fabtests/common/shared.c#L2707-L2738) .

PR #11850, opened 2026-01-28
core: Add max_cntr_value and max_err_cntr_value to fi_domain_attr 

jiaxiyan:
fi_cntr_read/fi_cntr_readerr return uint64_t but the full range may not be supported due to hardware limitations.
Allow providers to specify limits on these values so applications can discover and handle the wraparound properly.

### DIFF
--- a/man/fi_info.1.md
+++ b/man/fi_info.1.md
@@ -195,6 +195,8 @@ fi_info:
         max_err_data: 0
         mr_cnt: 0
         tclass: 0x0
+        max_ep_auth_key: 0
+        max_group_id: 0
         max_cntr_value: 18446744073709551615
         max_err_cntr_value: 18446744073709551615
     fi_fabric_attr:
--- a/src/fi_tostr.c
+++ b/src/fi_tostr.c
@@ -541,6 +541,10 @@ ofi_tostr_domain_attr(char *buf, size_t len, const struct fi_domain_attr *attr,
 		     attr->max_err_data);
 	ofi_strncatf(buf, len, "%s%smr_cnt: %zu\n", prefix, TAB, attr->mr_cnt);
 	ofi_strncatf(buf, len, "%s%stclass: 0x%x\n", prefix, TAB, attr->tclass);
+	ofi_strncatf(buf, len, "%s%smax_ep_auth_key: %zu\n", prefix, TAB,
+		     attr->max_ep_auth_key);
+	ofi_strncatf(buf, len, "%s%smax_group_id: %u\n", prefix, TAB,
+		     attr->max_group_id);
 	ofi_strncatf(buf, len, "%s%smax_cntr_value: %lu\n", prefix, TAB,
 		     attr->max_cntr_value);
 	ofi_strncatf(buf, len, "%s%smax_err_cntr_value: %lu\n", prefix, TAB,


jiaxiyan:
@jswaro @j-xiong Can you review this?

prov/util/src/util_attr.c
```diff
@@ -727,6 +727,21 @@ int ofi_check_domain_attr(const struct fi_provider *prov, uint32_t api_version,
 		return -FI_ENODATA;
 	}
 
+	if (FI_VERSION_GE(api_version, FI_VERSION(2, 5))) {
+		if (user_attr->max_cntr_value > prov_attr->max_cntr_value) {
```
j-xiong: Not every provider set `max_cntr_value`, nor call `ofi_alter_info`. For best compatibility, we may want to consider `0` as a special case:  for provider attributes, it works the same way as UINT64_MAX; for hints, it means "don't care".

man/fi_cntr.3.md
```diff
@@ -229,11 +229,37 @@ control commands are usable with a counter:
...
 The fi_cntr_read call returns the current value of the counter.
 
+Applications must handle counter wraparound that occurs when counter values
+exceed the limit defined by the domain attribute `max_cntr_value`. This
+value, which defaults to UINT64_MAX but may be lower due to provider
+limitations, defines the maximum value returned by fi_cntr_read()
+before wraparound occurs.
+
+When a counter reaches its maximum value and additional operations complete,
+the counter will wrap around to zero and continue incrementing from that
```
shijin-aws: I am not sure whether other providers will also wrap around, I'd prefer to make it general that "when a counter reaches its maximal value and additional operations complete, the counter behavior will be undefined and result in incorrect value."

man/fi_cntr.3.md
```diff
@@ -229,11 +229,37 @@ control commands are usable with a counter:
...
+point. Applications that need to track the total number of operations
+beyond the provider's counter limits are responsible for detecting and
+handling this wraparound behavior by resetting counters before they reach
```
shijin-aws: I think we should mention to use `fi_cntr_set` to reset the counter

man/fi_cntr.3.md
```diff
@@ -229,11 +229,37 @@ control commands are usable with a counter:
...
 error and were unable to update the counter.
 
+Applications must handle counter wraparound that occurs when error counter
+values exceed the limit defined by the domain attribute `max_err_cntr_value`.
+This value, which defaults to UINT64_MAX but may be lower due to provider
+limitations, defines the maximum value returned by fi_cntr_readerr()
+before wraparound occurs.
+
+When an error counter reaches its maximum value and additional error operations
+complete, the counter will wrap around to zero and continue incrementing from
```
shijin-aws: Similar comment as the normal cntr

prov/util/src/util_attr.c
shijin-aws: Agree

man/fi_domain.3.md
```diff
@@ -815,6 +817,27 @@ support for peer groups by setting this to a non-zero value.  Providers that
 cannot meet the requested max_group_id will fail fi_getinfo().  On output,
 providers may return a value higher than that requested by the application.
 
+## Maximum Counter Value (max_cntr_value)
+
+The maximum value returned by fi_cntr_read before wraparound occurs.
```
shijin-aws: Do we also support this as input?

prov/util/src/util_attr.c
```diff
@@ -1274,6 +1274,12 @@ static void fi_alter_domain_attr(struct fi_domain_attr *attr,
 	}
 
 	attr->caps = ofi_get_caps(info_caps, hints ? hints->caps : 0, attr->caps);
+	
+	if (attr->max_cntr_value == 0)
```
shijin-aws, line 1280: if we want to support max value as input, this can be updated

man/fi_domain.3.md
```diff
@@ -815,6 +817,29 @@ support for peer groups by setting this to a non-zero value.  Providers that
...
 providers may return a value higher than that requested by the application.
 
+## Maximum Counter Value (max_cntr_value)
+
+The maximum value returned by fi_cntr_read. While counters use uint64_t,
+the full range may not be supported due to hardware limitations. This
+field allows providers to advertise their specific limits.
+The default value is UINT64_MAX or 0, which indicates the full counter range.
+
+Applications can set the max_cntr_value on input to fi_getinfo, in order to
```
j-xiong: `Applications can set a non-zero max_cntr_value on input ...`.

man/fi_domain.3.md
```diff
@@ -815,6 +817,29 @@ support for peer groups by setting this to a non-zero value.  Providers that
...
+indicate their counter requirements. Applications must manage counter values
```
j-xiong: Need more clarification such as: `The provider should return -FI_ENODATA if the supported counter range is smaller than the requirement.  The provider is allowed to return a larger value than the required one. `

man/fi_domain.3.md
```diff
@@ -815,6 +817,38 @@ support for peer groups by setting this to a non-zero value.  Providers that
 cannot meet the requested max_group_id will fail fi_getinfo().  On output,
 providers may return a value higher than that requested by the application.
 
+## Maximum Counter Value (max_cntr_value)
+
+The maximum value returned by fi_cntr_read. While counters use uint64_t,
+the full range may not be supported due to hardware limitations. This
+field allows providers to advertise their specific limits.
+The default value is UINT64_MAX or 0, which indicates the full counter range.
```
jswaro: So, if an application sets the value to 0, should a provider expect that the application will handle any value returned by the provider? 

Is it the case that 0 is 'Do not care', or 'give me your highest possible value', or something else?

shijin-aws: For the new API version, I think no matter what input value is, the application should check the return value and handle it. IMO 0 in the input means application do not care and provider return the default and optimized value of it, which can be smaller than UINT64_MAX

j-xiong: Correct, 0 as input means the application does not have any specific requirements (do not care).

man/fi_domain.3.md
```diff
@@ -815,6 +817,38 @@ support for peer groups by setting this to a non-zero value.  Providers that
...
+
+Applications can set a non-zero max_cntr_value on input to fi_getinfo, in
+order to indicate their counter requirements. The provider should return
+`-FI_ENODATA` if the supported counter range is smaller than the requirement.
+The provider is allowed to return a larger value than the required one.
```
shijin-aws: nit, `require` -> `request` , similar comment for other places that use `require` and `requirement`

man/fi_domain.3.md
shijin-aws: >The default value is UINT64_MAX or 0, which indicates the full counter range.

0 means provider doesn't set this value?

man/fi_domain.3.md
shijin-aws: I think we need to mention the API version, we are only failing when user uses API >= 2.5 right?

man/fi_domain.3.md
shijin-aws: I forget it was suggested by Jianxin... Just elaborate it here a little bit why 0 also means the full range...

man/fi_domain.3.md
jiaxiyan: Earlier versions didn't have this field so it doesn't apply anyway?

man/fi_domain.3.md
j-xiong: Yes, 0 is for the case that provider doesn't set the value. It is treated the same as UINT64_MAX for the best compatibility. How about rewording this sentence as:
```  
The default value is UINT64_MAX, which indicated the full counter range. A value 0 means the provider
doesn't set the limit and should be treated the same as the default value.
```

shijin-aws (approved):

man/fi_domain.3.md
shijin-aws: It is for API version, application can use a new libfabric but stick to an older API version of fi_getinfo, they may just find the new field in the header and use it, and we want to have a gate here.

j-xiong (approved):

a-szegel:
bot:aws:retest

a-szegel:
@j-xiong Can you re-run IntelCI here please?

shijin-aws:
bot:aws:retest

shijin-aws:
bot:aws:retest

shijin-aws:
> @j-xiong Can you re-run IntelCI here please?

Bumping this, @j-xiong mind taking a look?

shijin-aws:
@j-xiong good to merge?

jswaro:
@j-xiong : We had a question from the team about behavior on roll-over. If the counter hits max value, does libfabric expect truncation or roll-over? CXI truncates. 

If it is different from provider to provider, then perhaps we need one more field to flag to applicatons what the expected behavior is.

bwbarrett:
@jswaro I'm being dense this morning.  Assuming an unsigned integer, what's the behavior difference between truncation and roll-over in terms of user-visible behaviors?

jswaro:
In this context, would the counter stay at max, or roll over to 0 through hardware. In CXI, if it is at max, and another increment comes in. Then it stays at max. Effectively, the increment is discarded.

Do other providers have the same behavior, or do they roll over if an increment would increment past the max bits

shijin-aws:
> In this context, would the counter stay at max, or roll over to 0 through hardware. In CXI, if it is at max, and another increment comes in. Then it stays at max. Effectively, the increment is discarded.
> 
> Do other providers have the same behavior, or do they roll over if an increment would increment past the max bits

I had the exact same question in https://github.com/ofiwg/libfabric/pull/11850#discussion_r2752539607 . I think today the man page just describes the behavior as "undefined and can cause incorrect value". I am not sure what is the value to define such overflow behavior when application is responsible to reset the counter to avoid such overflow?

bwbarrett:
Oh, that's not what I thought you meant by truncation.  EFA's behavior will be to overflow, so we either need to be ok with Shi's suggestion of "undefined behavior", or define both options and let the provider say which one you get.

j-xiong:
I would suggest stay with current "undefined behavior".  Applications are expected to reset the counter in time so that truncation/overflow never happens.

j-xiong:
And it's up to the application to determine when to reset to be "in time".  With the "max value" on hand, it doesn't need to wait for the very end to reset.

PR #11853, opened 2026-01-29
prov/efa: Fix iterator invalidation in efa_domain_progress_rdm_peers_and_queues()

alekswn:
efa_domain_progress_rdm_peers_and_queues() calls efa_rdm_txe_handle_error() while iterating over  ope_longcts_send_list.
efa_rdm_txe_handle_error() can ~release~ remove current list entry, therefore we have to use safe iteration

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -828,8 +828,8 @@ void efa_domain_progress_rdm_peers_and_queues(struct efa_domain *domain)
 	/*
 	 * Send data packets until window or data queue is exhausted.
 	 */
-	dlist_foreach_container(&domain->ope_longcts_send_list, struct efa_rdm_ope,
-				ope, entry) {
+	dlist_foreach_container_safe(&domain->ope_longcts_send_list, struct efa_rdm_ope,
+				ope, entry, tmp) {
 		peer = ope->peer;
 		assert(peer);
 		if (peer->flags & EFA_RDM_PEER_IN_BACKOFF)


prov/efa/src/efa_domain.c
```diff
@@ -828,7 +828,7 @@ void efa_domain_progress_rdm_peers_and_queues(struct efa_domain *domain)
 	/*
 	 * Send data packets until window or data queue is exhausted.
 	 */
-	dlist_foreach_container(&domain->ope_longcts_send_list, struct efa_rdm_ope,
+	dlist_foreach_container_safe(&domain->ope_longcts_send_list, struct efa_rdm_ope,
```
shijin-aws, line 831: tx_handle_error doesn't release the entry https://github.com/ofiwg/libfabric/blob/4cd9317bb694db7697efbb02b7645ad3a27d59e4/prov/efa/src/rdm/efa_rdm_ope.c#L787

alekswn: We can call  `efa_rdm_txe_handle_error()` in this loop. 
Seems like it can remove current list entry, which would invalidate iterator.
It's done here:
https://github.com/ofiwg/libfabric/blob/4cd9317bb694db7697efbb02b7645ad3a27d59e4/prov/efa/src/rdm/efa_rdm_ope.c#L740

Am I wrong?

shijin-aws: dlist_remove won't change the current entry's prev and next ptr, so that iterator should not be impacted. the only reason to use safe is that the entry itself is getting freed

PR #11854, opened 2026-01-29
prov/efa: Open mr cache when it is enabled

shijin-aws:
MR cache should only be used for internal MR
registration which doesn't depend on application
behavior. And explicit fi_mr_reg is not using mr cache anyway.

### DIFF
--- a/prov/efa/src/efa_domain.c
+++ b/prov/efa/src/efa_domain.c
@@ -262,13 +262,12 @@ int efa_domain_open(struct fid_fabric *fabric_fid, struct fi_info *info,
 	}
 
 	/*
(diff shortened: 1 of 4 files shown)


shijin-aws:
EFA NIC require FI_MR_LOCAL, but if application doesn't want to manage their registration (they do not set FI_MR_LOCAL), the provider will manage the registration internally via mr cache. This is what efa protocol path does for MPI which pass NULL desc for every host memory send.

If application supports FI_MR_LOCAL, those internal registration of user buffer won't happen, so the mr cache is indeed less useful. It can be enabled and we will only have internal bounce buffer pool registered in the cache which are actually not helpful.

shijin-aws:
Based on this reasoning I am closing the PR

PR #11856, opened 2026-01-30
prov/efa: Fix uninitialized descriptor arrays causing segfault 

yinliaws:
This patch fixes two related bugs involving uninitialized descriptor arrays that cause segmentation faults when FI_EFA_ENABLE_SHM_TRANSFER=0 is set.

Bug 1: Incorrect sizeof() in efa_rdm_txe_construct() When msg->desc is NULL, the code attempted to zero the desc array with: memset(txe->desc, 0, sizeof(txe->desc));

This only zeros sizeof(void*) = 8 bytes instead of the entire array. The correct size should be sizeof(*txe->desc) * msg->iov_count to match the memcpy() branch and the parallel iov[] and mr[] arrays.

Bug 2: Dangling pointer in efa_rdm_atomic_ex
The compare_desc field was declared as void **compare_desc (pointer), which was set to point to the caller's stack memory. When the caller returned, this became a dangling pointer. Later access to compare_desc[i] would read garbage from reused stack memory.

The fix changes compare_desc to an array void *compare_desc[EFA_RDM_IOV_LIMIT] with its own persistent storage, and copies the data instead of storing a pointer.

Tested on c7i.48xlarge clusters with 96-192 ranks using IMB-EXT Accumulate benchmark in both debug and release builds.

### DIFF
--- a/prov/efa/src/rdm/efa_rdm_atomic.c
+++ b/prov/efa/src/rdm/efa_rdm_atomic.c
@@ -436,7 +436,7 @@ efa_rdm_atomic_compwritemsg(struct fid_ep *ep,
 	struct efa_rdm_atomic_ex atomic_ex = {
 		.resp_iov_count = result_count,
 		.comp_iov_count = compare_count,
-		.compare_desc = compare_desc,
+		
 	};
 	size_t datatype_size;
 	int err;
@@ -481,6 +481,9 @@ efa_rdm_atomic_compwritemsg(struct fid_ep *ep,
 
 	ofi_ioc_to_iov(resultv, atomic_ex.resp_iov, result_count, datatype_size);
 	ofi_ioc_to_iov(comparev, atomic_ex.comp_iov, compare_count, datatype_size);
+	memset(atomic_ex.compare_desc, 0, sizeof(atomic_ex.compare_desc));
+	if (compare_desc)
+		memcpy(atomic_ex.compare_desc, compare_desc, sizeof(void*) * compare_count);
 	memcpy(atomic_ex.result_desc, result_desc, sizeof(void*) * result_count);
 
 	return efa_rdm_atomic_generic_efa(efa_rdm_ep, msg, peer, &atomic_ex, ofi_op_atomic_compare, flags);
--- a/prov/efa/src/rdm/efa_rdm_ope.c
+++ b/prov/efa/src/rdm/efa_rdm_ope.c
@@ -46,10 +46,11 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 
 	memcpy(txe->iov, msg->msg_iov, sizeof(struct iovec) * msg->iov_count);
 	memset(txe->mr, 0, sizeof(*txe->mr) * msg->iov_count);
-	if (msg->desc)
+	if (msg->desc) {
 		memcpy(txe->desc, msg->desc, sizeof(*msg->desc) * msg->iov_count);
-	else
-		memset(txe->desc, 0, sizeof(txe->desc));
+	} else {
+		memset(txe->desc, 0, sizeof(*txe->desc) * msg->iov_count);
+	}
 
 	/* cq_entry on completion */
 	txe->cq_entry.op_context = msg->context;
--- a/prov/efa/test/efa_unit_test_ope.c
+++ b/prov/efa/test/efa_unit_test_ope.c
@@ -1067,3 +1067,112 @@ void test_efa_rdm_ope_eor_packet_tracking_unresponsive_wait_send(struct efa_reso
 {
 	test_efa_rdm_ope_ack_packet_tracking_unresponsive_wait_send_common(state, EFA_RDM_EOR_PKT);
 }
+
+/**
+ * @brief Test that atomic_ex.compare_desc array is properly copied
+ * and persists after the caller's stack frame is destroyed
+ *
+ * This test verifies the compare_desc fix where compare_desc was a
+ * pointer to stack memory that became dangling.
+ *
+ * @param[in]	state	struct efa_resource that is managed by the framework
+ */
+void test_efa_rdm_atomic_compare_desc_persistence(struct efa_resource **state)
+{
+	struct efa_resource *resource = *state;
+	struct efa_unit_test_buff send_buff, result_buff, compare_buff;
+	struct efa_ep_addr raw_addr;
+	size_t raw_addr_len = sizeof(raw_addr);
+	fi_addr_t addr;
+	uint64_t operand = 0x1234567890ABCDEF;
+	uint64_t compare = 0;
+	void *desc_array[1];
+	void *result_desc_array[1];
+	int ret;
+	struct fi_ioc ioc = {0};
+	struct fi_ioc compare_ioc = {0};
+	struct fi_ioc result_ioc = {0};
+	struct fi_rma_ioc rma_ioc = {0};
+	struct fi_msg_atomic msg = {0};
+	struct efa_rdm_ep *efa_rdm_ep;
+	struct efa_rdm_peer *peer;
+	struct efa_rdm_ope *txe;
+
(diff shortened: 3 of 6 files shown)


a-szegel:
Can you please make this 1 bug fix per commit?

prov/efa/src/rdm/efa_rdm_atomic.c
```diff
@@ -481,6 +481,9 @@ efa_rdm_atomic_compwritemsg(struct fid_ep *ep,
 
 	ofi_ioc_to_iov(resultv, atomic_ex.resp_iov, result_count, datatype_size);
 	ofi_ioc_to_iov(comparev, atomic_ex.comp_iov, compare_count, datatype_size);
+	memset(atomic_ex.compare_desc, 0, sizeof(atomic_ex.compare_desc));
```
a-szegel, line 484: > When the caller returned, this became
a dangling pointer. Later access to compare_desc[i] would read garbage from
reused stack memory.

Is this a problem because we are queue'ing the atomic packet instead of sending it our immediately?  Do we have this problem other places?  My understanding of the atomic protocol is that we would immediately send out an eager RTA, and we would never use this descriptor again after we return control back to the application.

shijin-aws: I think it is likely because we are send this via FI_DELIVERY_COMPLETE flags @yinliaws can u check? Then in that case we can queue the ope before a handshake is made before the local and peer.

shijin-aws: I thought we always copy the desc into the ope, we cannot just pass the **desc from the message, yeah we should check if we have such problem elsewhere in the protocol path? It is ok to pass **desc for efa-direct because we will never access it after function returns

shijin-aws: The mr needs to valid, but that desc is a void** which can be a temp array of the mr descs

yinliaws: The bug occurs when the atomic operation is queued rather than sent immediately. The flow is: 1. efa_rdm_atomic_compwritemsg() creates atomic_ex on stack; 2. In efa_rdm_atomic_alloc_txe: memcpy(&txe->atomic_ex, atomic_ex, sizeof(struct efa_rdm_atomic_ex));, this copies the entire struct including the compare_desc pointer. 3. In efa_rdm_atomic_post_atomic(), the txe is queued if FI_DELIVERY_COMPLETE flag is set and before a handshake is arrived. I checked compare_desc was the only field using void** pointer which points to external memory. All other descriptor arrays are declared as void*[].

yinliaws: I think the the actual MR descriptors in the array remain valid, but we are reading them through a dangling pointer.

shijin-aws:
Can u implement unit-test for your change, I think it should be in efa_unit_test_ope.c

prov/efa/src/rdm/efa_rdm_atomic.c
a-szegel: FI_DELIVERY_COMPLETE protocol capability was introduced in libfabric 1.12.  Do we really care about keeping libfabric 2.4 working with libfabric <=1.11?  I think it makes sense to drop this: https://github.com/ofiwg/libfabric/blob/main/prov/efa/src/rdm/efa_rdm_atomic.c#L107-L123

shijin-aws: it's like this

```
void foo(fid_mr *mr)
{
void *temp[2];
temp[1] = fi_mr_desc(mr);
temp[2] = fi_mr_desc(mr);
return fi_atmoic(..., temp);
}
```
when foo returns, mr is still valid, temp (void **) is not

yinliaws:
> Can u implement unit-test for your change, I think it should be in efa_unit_test_ope.c

A unit test was added.

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -49,7 +49,7 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 	if (msg->desc)
 		memcpy(txe->desc, msg->desc, sizeof(*msg->desc) * msg->iov_count);
 	else
-		memset(txe->desc, 0, sizeof(txe->desc));
+		memset(txe->desc, 0, sizeof(*txe->desc) * msg->iov_count);
```
shijin-aws: This is good catch, I am wondering whether we can get rid of memset by zeroing it during the declaration

prov/efa/test/efa_unit_test_ope.c
```diff
@@ -1066,4 +1066,48 @@ void test_efa_rdm_ope_eor_packet_failed_posting(struct efa_resource **state)
...
+		sentinel_values[i] = (void *)(uintptr_t)(0xDEADBEEF + i);
+		compare_desc_stack[i] = sentinel_values[i];
+	}
+
+	/* Initialize atomic_ex and copy compare_desc array */
+	memset(&atomic_ex, 0, sizeof(atomic_ex));
+	atomic_ex.comp_iov_count = 4;
+
+	/* This is what the fix does - copy the array contents */
+	memcpy(atomic_ex.compare_desc, compare_desc_stack,
```
shijin-aws: This is copying the fix, not testing the fix? You should be able to create a test that construct the efa_rdm_atomic_ex and verify the txe->desc is still valid after the fi_atomic returns, like what I described in https://github.com/ofiwg/libfabric/pull/11856/changes#r2748287428

shijin-aws: For this exact case, we know we will have a ope queued after fi_atomic returns when it is sent via delivery complete, you can retrieve the op entry from the queue list and check it

prov/efa/test/efa_unit_test_ope.c
```diff
@@ -1066,4 +1066,92 @@ void test_efa_rdm_ope_eor_packet_failed_posting(struct efa_resource **state)
...
+	 */
+	ret = fi_compare_atomic(resource->ep,
+				send_buff.buff, 1, fi_mr_desc(send_buff.mr),
+				compare_buff.buff, desc_array,
+				result_buff.buff, fi_mr_desc(result_buff.mr),
+				addr, 0, 0,
+				FI_UINT64, FI_CSWAP, NULL);
+
+	/* Operation should return -FI_EAGAIN (peer not ready) or success */
+	assert_true(ret == -FI_EAGAIN || ret == 0);
```
shijin-aws: First, EAGAIN doesn't necessarily mean peer not ready, it is a return code to indicate application needs to poll cq and call again. Second, I think we need to be very clear here. The ope entry is queued when a FI_DELIVERY_COMPLETE (DC) is requested while we do not have a handshake. The return code is 0 in this case. You can request a DC call via fi_atomic_compwritemsg with DC in the flag

prov/efa/src/rdm/efa_rdm_ope.c
```diff
@@ -46,10 +46,9 @@ void efa_rdm_txe_construct(struct efa_rdm_ope *txe,
 
 	memcpy(txe->iov, msg->msg_iov, sizeof(struct iovec) * msg->iov_count);
 	memset(txe->mr, 0, sizeof(*txe->mr) * msg->iov_count);
+	memset(txe->desc, 0, sizeof(*txe->desc) * msg->iov_count);
```
shijin-aws: This will cause extra memset when desc is not NULL, let's keep it the old way. Ideally, we should remove that memset entirely in the future

prov/efa/test/efa_unit_test_ope.c
yinliaws: Thanks! Revised accordingly.

shijin-aws:
bot:aws:retest

PR #11857, opened 2026-01-30
fabtests/efa: Add nvcc flag for Blackwell sm_100

jiaxiyan:
Add support for p6-b200 Blackwell architecture.

### DIFF
--- a/fabtests/prov/efa/src/efagda/Makefile
+++ b/fabtests/prov/efa/src/efagda/Makefile
@@ -1,6 +1,6 @@
 CC=gcc
 NVCC=nvcc
-NVCC_FLAGS=-G -arch=sm_90
+NVCC_FLAGS=-G -gencode arch=compute_90,code=sm_90 -gencode arch=compute_100,code=sm_100
 CFLAGS=-pthread
 CUDA_FLAGS=-Xcompiler -fPIC
 CUDA_INCLUDES=-I/usr/local/cuda/include


fabtests/prov/efa/src/efagda/Makefile
```diff
@@ -1,6 +1,6 @@
 CC=gcc
 NVCC=nvcc
-NVCC_FLAGS=-G -arch=sm_90
+NVCC_FLAGS=-G -gencode arch=compute_90,code=sm_90 -gencode arch=compute_100,code=sm_100
```
shijin-aws, line 3: what will happen if we do not specify the arch?

jiaxiyan: It will use a default old architecture with`nvcc warning : Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).`
PTX JIT will be used and it becomes a problem when we run with mismatched nvcc and driver versions `PTX was compiled with an unsupported toolchain`.

shijin-aws: My question is whether keeping adding new arch to this line is a scalable solution

jiaxiyan: It is the preferred way if we know which GPU it is. For future GPUs, it is not needed if we run with compatible driver and nvcc version but I don't want to explicitly check the versions.

PR #11858, opened 2026-02-02
log: add log_location parameter to allow log redirection to file

philippfriese:
This PR adds the parameter `log_location` / the environment variabel `FI_LOG_LOCATION` to the standard logging provider in to allow control of the output location of the libfabric log. Possible options are: _stderr_ (default), _stdout_, as well as a path to a file or to a directory. If a directory is provided, a log file with format `ofi_<pid>.log` is used. Log files are created with the file mode specified via `log_location_mode` / `FI_LOG_LOCATION_MODE`, defaulting to `0600`, and must not exist prior to launching libfabric[1].

The use-case of this parameter concerns multi-process application cases, such as MPI applications. By default, libfabric writes to _stderr_, which is usually unbuffered. Even when _stderr_ of an MPI application is redirected to per-process files, for example using Open MPIs `--output-filename` option, then the libfabric log still contains interleaved lines, resulting in one MPI process' log containing a log line from another MPI process. 
When modifying _stderr_ to be line-buffered, then libfabric log entries are no longer interleaved, but may be out-of-order, which may render certain provider outputs impossible to interpret correctly. An example is the output of the `ofi_hook_profile` provider, which relies on strict line ordering for its output sematic.

While libfabric allows the logging provider to be extended and overwritten programmatically, this approach is not feasible for applications using another communication library on-top of libfabric, such as MPI applications.
The parameter introduced in this PR addresses this.

[1] This is to prevent libfabric from overwriting/modifying files via an accidentally or intentionally mis-crafted log location path. 

### Usage Example

```console
$ FI_LOG_LEVEL=debug FI_LOG_LOCATION="./" fi_info -l
ofi_rxm:
    version: 201.0
ofi_rxd:
[...]
$ ls ./
ofi_545157.log
$ cat ofi_545157.log
libfabric:545157:1770045780::core:core:fi_param_get_():372<info> variable perf_cntr=<not set>
libfabric:545157:1770045780::core:core:fi_param_get_():372<info> variable hook=<not set>
[...]
```

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -36,6 +36,10 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <limits.h>
+#include <sys/stat.h>
+#include <string.h>
+#include <errno.h>
 
 #include <rdma/fi_errno.h>
 #include <rdma/fi_ext.h>
@@ -88,6 +92,7 @@ enum {
 
 static int log_interval = 2000;
 uint64_t log_mask;
+static FILE *log_location;
 struct ofi_filter prov_log_filter;
 extern struct ofi_common_locks common_locks;
 
@@ -107,11 +112,64 @@ static int fi_convert_log_str(const char *value)
 	return 0;
 }
 
+static void fi_log_location_init(char* locationstr, mode_t mode)
+{
+	struct stat st = {0};
+	char path_buffer[PATH_MAX];
+	int len;
+	int fd;
+
+	if (strcmp(locationstr, "stderr") == 0 || *locationstr == '\0') {
+		log_location = stderr;
+		return;
+	}
+	if (strcmp(locationstr, "stdout") == 0) {
+		log_location = stdout;
+		return;
+	}
+
+	if (stat(locationstr, &st) == -1 && errno != ENOENT)
+		goto error_log_location;
+
+	if (st.st_mode & S_IFDIR) {
+		len = snprintf(path_buffer, PATH_MAX, "%s/ofi_%u.log",
+			locationstr, pid);
+		if (len < 0)
+			goto error_log_location;
+		if (len >= PATH_MAX) {
+			errno = ENAMETOOLONG;
+			goto error_log_location;
+		}
+	} else {
+		strncpy(path_buffer, locationstr, PATH_MAX-1);
+	}
+
+	// do not open (and potentially overwrite) existing files
+	fd = ofi_open(path_buffer, O_RDWR|O_CREAT|O_EXCL, mode);
+	if (fd == -1)
+		goto error_log_location;
+	log_location = ofi_fdopen(fd, "r+");
+	if (log_location == NULL) {
+		ofi_close(fd);
+		goto error_log_location;
+	}
+	return;
+
+error_log_location:
+	log_location = stderr;
+	FI_WARN(&core_prov, FI_LOG_CORE,
+		"Could not initialize log location at '%s': %s. "
+		"Will write log to stderr.\n",
+		locationstr, strerror(errno));
+}
+
 void fi_log_init(void)
 {
 	struct ofi_filter subsys_filter;
 	int level, i;
-	char *levelstr = NULL, *provstr = NULL, *subsysstr = NULL;
+	char *levelstr = NULL, *provstr = NULL, *subsysstr = NULL,
+		*locationstr = NULL;
+	int location_mode = 0600;
 
 	fi_param_define(NULL, "log_interval", FI_PARAM_INT,
 			"Delay in ms between rate limited log messages "
@@ -140,6 +198,19 @@ void fi_log_init(void)
 	}
 	ofi_free_filter(&subsys_filter);
 	pid = getpid();
+
+	fi_param_define(NULL, "log_location", FI_PARAM_STRING,
+			"Specify log output: stderr, stdout, filepath, or dirpath "
+			"(default: stderr)");
+	fi_param_define(NULL, "log_location_mode", FI_PARAM_INT,
+			"Specify file mode for log output (default: 0600)");
+	if (fi_param_get_int(NULL, "log_location_mode", &location_mode) != FI_SUCCESS)
+		location_mode = 0600;
+	if (fi_param_get_str(NULL, "log_location", &locationstr) == FI_SUCCESS
+		&& locationstr != NULL)
+		fi_log_location_init(locationstr, location_mode);
+	else
+		log_location = stderr;
 }
 
 static int ofi_log_enabled(const struct fi_provider *prov,
@@ -149,14 +220,14 @@ static int ofi_log_enabled(const struct fi_provider *prov,
 	int prov_filtered = (flags & FI_LOG_PROV_FILTERED);
 
 	return ((FI_LOG_TAG(prov_filtered, level, subsys) & log_mask) ==
-		FI_LOG_TAG(prov_filtered, level, subsys));
+		FI_LOG_TAG(prov_filtered, level, subsys)) && log_location != NULL;
 }
 
 static void ofi_log(const struct fi_provider *prov, enum fi_log_level level,
 		    enum fi_log_subsys subsys, const char *func, int line,
 		    const char *msg)
 {
-	fprintf(stderr, "%s:%d:%ld:%s:%s:%s:%s():%d<%s> %s",
+	fprintf(log_location, "%s:%d:%ld:%s:%s:%s:%s():%d<%s> %s",
 		PACKAGE, pid, (unsigned long) time(NULL), log_prefix,
 		prov->name, log_subsys[subsys], func, line,
 		log_levels[level], msg);
@@ -318,6 +389,21 @@ void ofi_tostr_log_subsys(char *buf, size_t len, enum fi_log_subsys subsys)
 void fi_log_fini(void)
 {
 	ofi_free_filter(&prov_log_filter);
+	if (log_location != NULL && log_location != stderr
+		&& log_location != stdout) {
+		if (fflush(log_location) != 0) {
+			log_location = stderr;
+			FI_WARN(&core_prov, FI_LOG_CORE,
+				"Failed to flush log file: %s\n",
+				strerror(errno));
+		}
+		if (fclose(log_location) != 0) {
+			log_location = stderr;
+			FI_WARN(&core_prov, FI_LOG_CORE,
+				"Failed to close log file: %s\n",
+				strerror(errno));
+		}
+	}
 }
 
 __attribute__((visibility ("default"),EXTERNALLY_VISIBLE))
--- a/man/fabric.7.md
+++ b/man/fabric.7.md
@@ -171,8 +171,8 @@ interfaces are defined by libfabric.
 
 # LOGGING INTERFACE
 
-Logging can be controlled using the FI_LOG_LEVEL, FI_LOG_PROV, and
-FI_LOG_SUBSYS environment variables.
+Logging can be controlled using the FI_LOG_LEVEL, FI_LOG_LOCATION,
+FI_LOG_LOCATION_MODE, FI_LOG_PROV, and FI_LOG_SUBSYS environment variables.
 
 *FI_LOG_LEVEL*
 : FI_LOG_LEVEL controls the amount of logging data that is output.  The
@@ -193,6 +193,20 @@ FI_LOG_SUBSYS environment variables.
 : Debug is high traffic and is likely to impact application performance.
   Debug output is only available if the library has been compiled with
   debugging enabled.
+
+*FI_LOG_LOCATION*
+: FI_LOG_LOCATION controls the location of the log output. Possible options are
+  `stdout`, `stderr`, as well as a path to a file or a directory. The default
+  location is `stderr`.
+  If set to a file path, then the specified file will be created and used. The
+  file pointed to by the setting must not exist prior to launching libfabric.
+  If set to a directory path, then a new file `ofi_<pid>.log` in the specified
+  directory will be used.
+  Log files use the file mode specified via *FI_LOG_LOCATION_MODE*.
+
+*FI_LOG_LOCATION_MODE*
+: FI_LOG_LOCATION_MODE controls the mode of the log file if *FI_LOG_LOCATION* is
+  set to a file or directory. The default is `0600`.
 
 *FI_LOG_PROV*
 : The FI_LOG_PROV environment variable enables or disables logging from


shefty:
The line indentation is off -- using spaces instead of tabs.  A library should never print without explicitly being enabled to do so.  For all the library knows, stderr could have been closed and redirected to an application file.  Use only logging functions to print errors, not direct calls to fprintf.

philippfriese:
Thank you for the feedback @shefty. I have pushed an updated commit which moves the fprintf calls to the OFI logging macros.
I also moved the handling of `log_location{_mode}` to after the handling of the other `log_*` settings in `fi_log_init` and added a check for `log_location != NULL` to `ofi_log_enabled`. Without this check, a small amount of logging calls, emitted by the `fi_param_get/define` functions in `fi_log_init` after setting `log_mask` and before calling `fi_log_location_init`, would be logged to stderr even if `FI_LOG_LOCATION` has been set to a valid, non-stderr target, breaking the semantics of that setting. 
As a consequence, these logging calls do not get emitted at all. This matches behaviour of the calls to `fi_param*` before setting `log_mask`, which also do not emit logs, even prior to this PR.

As for the line indentation: I adjusted some indentations but overall could not find places where my changes use spaces instead of tabs. Is this an issue on my end or did I miss spots where I did in fact use spaces for indentation?

src/log.c
```diff
@@ -107,11 +113,69 @@ static int fi_convert_log_str(const char *value)
 	return 0;
 }
 
+static void fi_log_location_init(char* locationstr, mode_t mode)
+{
+	struct stat st;
+	char path_buffer[PATH_MAX];
+	int len;
+
+	if (strncmp(locationstr, "stderr", 6) == 0 || *locationstr == '\0') {
```
j-xiong: Better to use `strcmp` instead. Reason: (1) for exact match; (2) the second string is constant with known length so the comparison will stop after 6 characters any way. Same for L126.

src/log.c
```diff
@@ -107,11 +113,69 @@ static int fi_convert_log_str(const char *value)
...
+			errno = ENAMETOOLONG;
+			goto error_log_location;
+		}
+	} else {
+		// stat(locationstr) above already checks if len >= PATH_MAX,
+		//  so this strncpy cannot truncate
+		strncpy(path_buffer, locationstr, PATH_MAX-1);
+		if (stat(dirname(path_buffer), &st) == -1)
+			goto error_log_location;
+		strncpy(path_buffer, locationstr, PATH_MAX-1);
```
j-xiong, line 144: L148 is duplicate of L145.

src/log.c
philippfriese: Thanks! Changed in the updated commit.

src/log.c
philippfriese: This is in fact intentional: `dirname` may modify the input string, so the second strncpy fetches a fresh version. I have added a comment in the updated commit to document this intent.

j-xiong:
Windows build failure is real:
```
  C:\projects\libfabric\src\log.c(43,1): fatal error C1083: Cannot open include file: 'libgen.h': No such file or directory
```

j-xiong:
The CodeQL warning needs to be addressed as well.

philippfriese:
Thanks for the feedback @j-xiong! The updated commit moves away from `fopen` + `chmod` and to `open` with appropriate flags + `fdopen`, which resolves the CodeQL TOCTOU warnings. This also removed the requirement of `dirname`/`libgen.h`, clearing the previous Windows build issues. 
In addition, I added a brief `#ifdef _WIN32` which defines: `PATH_MAX` as `MAX_PATH`, `mode_t` as an `uint32_t`, as well as `{fd}open` as `_{fd}open`. These are not present/defined when using the Windows toolchain. Please let me know whether this is an appropriate way to address these compilation issues.

src/log.c
```diff
@@ -45,6 +49,12 @@
 #include "ofi_enosys.h"
 #include "ofi_util.h"
 
+#ifdef _WIN32
+#define PATH_MAX MAX_PATH
+typedef uint32_t mode_t;
+#define open _open
+#define fdopen _fdopen
+#endif
```
j-xiong: These changes should go into `include/windows/osd.h`

src/log.c
```diff
@@ -107,11 +118,62 @@ static int fi_convert_log_str(const char *value)
...
+		strncpy(path_buffer, locationstr, PATH_MAX-1);
+	}
+
+	// do not open (and potentially overwrite) existing files
+	fd = open(path_buffer, O_RDWR|O_CREAT|O_EXCL, mode);
+	if (fd == -1)
+		goto error_log_location;
+	log_location = fdopen(fd, "r+");
+	if (log_location == NULL)
+		goto error_log_location;
```
j-xiong, line 154: Should close the opened fd when `fdopen` fails.

philippfriese: Good catch, thanks! Addressed in the updated commit.

src/log.c
philippfriese: Thanks, moved in updated commit. Also added `_close` redefinition needed below.

philippfriese: Adding `#define close _close` to `include/windows/osd.h` breaks several OFI components which rely on the `close` keyword (see https://ci.appveyor.com/project/ofiwg/libfabric/builds/53592941). I presume something similar could happen with `open` as well.
Would it make sense to keep the `PATH_MAX` and `mode_t` definitions in `osd.h` and keep a separate `#ifdef _WIN32` in `log.c` for the respective file-descriptor calls? This would prevent this rather generic redefinition from spreading to undesired places.

j-xiong: Defining `open` and `close` as static inline functions should work.

include/unix/osd.h
```diff
@@ -117,6 +120,28 @@ static inline int ofi_memalign(void **memptr, size_t alignment, size_t size)
 	return posix_memalign(memptr, alignment, size);
 }
 
+static inline int ofi_osd_close(int fd)
```
j-xiong: Just use `ofi_close` and `ofi_open`.

philippfriese:
@j-xiong Thank you for the prompt feedback! I moved to `ofi_{open,close}`.
The implementation of `ofi_open` is somewhat more involved than expected: `open` takes a variadic argument `mode` if certain flags are set. However, it does not take a `va_list`. The added code checks, for each OS, whether `mode` is expected (for Linux: `O_CREAT` or `O_TMPFILE` are set, the rest: `O_CREAT` is set), and if so extracts it from the `va_list` and calls `open` with it.

include/freebsd/osd.h
```diff
@@ -77,6 +77,26 @@ static inline size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+		mode = (mode_t)va_arg(args, int);
+		fd = open(path, flags, mode);
+		va_end(args);
+	} else {
+		fd = open(path, flags);
+	}
+
+	return fd;
+}
+
```
j-xiong, line 94: Anything wrong with the following?
```
static inline int ofi_open(const char *path, int flags, ...)
{
        int ret; 
        va_list args;

        va_start(args, flags);
        ret = open(path, flags, args);
        va_end(args);

        return ret;
}
```
At least it works fine on Linux.

j-xiong:
bot:aws:retest

include/freebsd/osd.h
j-xiong: I am using this program to test the `ofi_open` implementation with va_list, please let me know what the errors you see on your platforms.

```
#include <stdio.h>
#include <stdarg.h>
#include <fcntl.h>
#include <unistd.h>

static inline int ofi_open(const char *path, int flags, ...)
{
        int ret;
        va_list args;

        va_start(args, flags);
        ret = open(path, flags, args);
        va_end(args);

        return ret;
}

int main()
{
        int fd;

        fd = ofi_open("test.txt", O_CREAT, 0666);
        if (fd < 0) {
                perror("open new");
                return 1;
        }
        printf("File created successfully with fd: %d\n", fd);
        close(fd);

        fd = ofi_open("test.txt", O_RDONLY);
        if (fd < 0) {
                perror("open existing");
                return 1;
        }
        printf("File opened successfully with fd: %d\n", fd);
        close(fd);

        return 0;
}
```

shijin-aws:
> bot:aws:retest

@j-xiong it is the same problem as the hook PR, this PR needs to be rebased to resolve AWS CI failure

j-xiong:
@shijin-aws Thanks!

include/freebsd/osd.h
philippfriese: On an arm64-based macOS 24, I can both compile and successfully execute this snippet, on an x86-based Ubuntu 24.04 I can compile, but not successfully execute due to a "Permission denied" on the second `ofi_open` call. In both cases however, the provided mode `0666` is invalid:

macOS arm64:
```console
% ./main.o
File created successfully with fd: 3
File opened successfully with fd: 3
% ls -lah test.txt
-rwx------  1 pfriese  wheel     0B 28 Feb 08:15 test.txt
```

Ubuntu x86:
```
% ./main.out
File created successfully with fd: 3
open existing: Permission denied
% ls -lah test.txt
---x--S---  1 friese tumuser    0 Feb 28 08:11 test.txt
```

Repeated invocation on Ubuntu yields random modes on the created file. I believe that the reason for this is that `open` expects a proper variadic argument, and not a `va_list`. 
If my understanding is correct, then the involved `va_arg` call in the `open` code path extracts the variable argument from the stack (or from registers containing the function arguments). In the above case (and my initial implementation), that contains the `va_list` structure, which contains pointers to where `mode` is, but not `mode` itself.
I assume the consistent (but still wrong) result on ARM/Darwin, as opposed to the seemingly random behaviour on x86/Linux, is due to different behaviour of `va_arg` / different function argument passing on these platforms.

I toyed with always extracting the `mode` parameter in `ofi_open` and always passing it on, but ultimately decided to use the commited implementation given that `va_arg` is UB if no variable argument is given:
> va_arg()
    [...] If there is no next argument [...], random errors will occur. [[man va_arg](https://www.man7.org/linux/man-pages/man3/va_arg.3.html)]

glibc handles this by only extracting `mode` if the `__OPEN_MEEDS_MODE` [macro succeeds](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/open.c;hb=HEAD), which [checks against](https://sourceware.org/git/?p=glibc.git;a=blob;f=io/fcntl.h;hb=HEAD#l39) `O_CREAT` and, if available, against `O_TMPFILE`. This way, "random errors" will only occur if the function has been called incorrectly. The man-page for `open` acknowledges this behaviour on incorrect usage:

> The mode argument must be supplied if O_CREAT or O_TMPFILE is specified in flags; if it is not supplied, some arbitrary bytes from the stack will be applied as the file mode. [[man 2 open](https://www.man7.org/linux/man-pages/man2/open.2.html)]

Checking the glib-implementation of `open`, I noticed that it always calls `open` with `mode`, which is either zero-initialised or populated via `va_arg`. I will push a new commit with an updated implementation that matches this approach.

j-xiong: Thanks for the explanation. 

Thinking it more, how useful is setting the permissions of the log files? If only taking the default permissions, we can get away with using `fopen` and `fclose` only, which don't need wrappers.

philippfriese: I also thought about doing that. The default permission is `0666 & umask`, which equates to `0644` on most platforms. I would assume that not everyone wants their libfabric logs to be globally readable by default, especially on multi-tenant systems.
Users can of course adjust their umask prior to launching libfabric, but given that we are a library, this would imply that users automatically change the behaviour of the entire application, which might be prohibitive or at least inconvenient in some cases.

I also thought about exposing control of the umask of just the libfabric log via something like `log_location_umask`, but that makes for a rather awkward interface and I would expect users to rightfully ask why they cannot set the mode directly.

One way to avoid the `open` wrappers is to statically set the umask to `0077` while creating the log file so that it gets created with a mode of `0600`. This would mean that users would have to manually `chmod` their log files in a second step if they want something else. The implementation would be simpler, but users would have less options. 
Let me know how to proceed.

j-xiong: Fair enough. 

Could you do a rebase so that AWS CI can pass?

philippfriese: Pushed commits with updated implementation of wrappers (now matches the glibc implementation linked above) and rebased against current main.

include/freebsd/osd.h
```diff
@@ -77,6 +77,23 @@ static inline size_t ofi_ifaddr_get_speed(struct ifaddrs *ifa)
...
+	// open takes a variadic argument 'mode_t mode', but not a va_list
+	// extract mode if required, as indicated by flag O_CREAT
+	if ((flags & O_CREAT) != 0) {
+		va_start(arg, flags);
+		mode = (mode_t)va_arg(arg, int);
+		va_end(arg);
+	}
+	fd = open(path, flags, mode);
+	return fd;
```
j-xiong: Can use `return open(path, flags, mode);` and eliminate `fd`. Same for other places.

philippfriese: Good catch, thanks! Changed in updated commit.

include/freebsd/osd.h
shefty: Why not have the user create and restrict a directory and not bother with file level permissions?

philippfriese: That is a good point. Users explicitly set `log_location` to a file or directory, so it may be reasonable to expect users to properly restrict the parent directory. If users do not set this parameter, then we do not write to disk, so there is no path where logs are written with open permissions without explicit user intent.

I would argue that restricted parent directories are a good defence-in-depth approach, however I would prefer users not having to read the man-page to find out that it is their responsibility to prevent global log read-access, and instead provide sensible defaults that do not necessitate further user action.
In addition, given that we are a library, I would imagine that users appreciate good configuration interfaces.

However, I do of course acknowledge that the wrappers are not as simple as desired. If upstreaming these is the cause for concern, then one alternative approach could be removing them and the `log_location_mode` parameter and instead writing the log-files with hard-coded mode `0600`. If users start voicing the desire to modify this at the libfabric level, then we can re-consider the wrappers and the `log_location_mode` setting.

j-xiong:
@shijin-aws @a-szegel @sunkuamzn AWS CI has been pending for a few days for this PR. Could you check what's going on?

sunkuamzn:
There was an infra issue. Triggering CI again bot:aws:retest

j-xiong:
bot:aws:retest

shijin-aws:
@j-xiong Sorry about the inconvenience, but this PR needs to be rebased again on top of https://github.com/ofiwg/libfabric/commit/29b8376e7aba4244f8e321f8af0d2672d75b24fd to make AWS CI pass

j-xiong:
@philippfriese Could you rebase again?

philippfriese:
@j-xiong rebased against the current main branch.

PR #11861, opened 2026-02-03
prov/cxi: Fix default mon start failure

SeanP-2023:
Properly handle a start failure when a user requests a monitor type that does not exist on the system via env FI_MR_CACHE_MONITOR during cxip_iomm_init


swelch (approved):

swelch:
appveyor failure is unrelated.

